PR #1764, opened 2016-07-08
Remove unnecessary reflection call

jskeet:
This is the only call to TypeExtensions.IsValueType, so we can remove
that method, making the whole type conditionally compiled out for .NET 3.5


anandolee:
LGTM

PR #1781, opened 2016-07-13
Update version number to 3.0.0-beta-4

xfxyjwf:
@pherl to review

### DIFF
--- a/ruby/pom.xml
+++ b/ruby/pom.xml
@@ -86,7 +86,7 @@
         <dependency>
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
-            <version>3.0.0-alpha-3</version>
+            <version>3.0.0-beta-3</version>
         </dependency>
     </dependencies>
 </project>
--- a/Protobuf.podspec
+++ b/Protobuf.podspec
@@ -5,7 +5,7 @@
 # dependent projects use the :git notation to refer to the library.
 Pod::Spec.new do |s|
   s.name     = 'Protobuf'
-  s.version  = '3.0.0-beta-3'
+  s.version  = '3.0.0-beta-4'
   s.summary  = 'Protocol Buffers v.3 runtime library for Objective-C.'
   s.homepage = 'https://github.com/google/protobuf'
   s.license  = 'New BSD'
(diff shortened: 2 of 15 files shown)


xfxyjwf:
@haberman There is another Gemfile.lock file in the ruby directory pointing to 3.0.0-alpha-5.0.5. Should we update that number or leave it as is?
https://github.com/google/protobuf/blob/master/ruby/Gemfile.lock#L4

pherl:
LGTM @jskeet  for C# numbers.

jskeet:
I'm still trying to get my PR for changing to use dotnet in... that PR will delete the nuspec file (and the most of AssemblyInfo, which would otherwise have to be changed).

If you want to fix this PR to do everything required for C#, edit csharp/src/Google.Protobuf/Properties/AssemblyInfo.cs as well. That's probably safest, just in case I can't actually get my PR in. (Travis is causing me grief...)

jskeet:
If my PR doesn't make Travis go green this time for the csharp build, I'll update the version number in my PR so we don't need another one. Given how long it takes to get anything through Travis, that should speed things up a bit.

xfxyjwf:
@jskeet I updated AssemblyInfo.cs in this PR. For tests, does appveyor have a good coverage on csharp? It finishes much quicker than travis.

jskeet:
@xfxyjwf: Yes, but it's the Travis setup part that's failing :( I think I understand it now, although it's slightly complicated. (Basically, we're now using Ubuntu 14.04 for Travis, only for C#. That changes how Mono is installed.)

I've updated the version number in my PR.

xfxyjwf:
Should I drop the C# version number change in this PR?

jskeet:
No, let's keep it in, just in case we have to drop my PR. Better to both do it than neither :)
(It may cause merge conflicts, of course.)

ruby/pom.xml
```diff
@@ -86,7 +86,7 @@
         <dependency>
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
-            <version>3.0.0-alpha-3</version>
+            <version>3.0.0-beta-3</version>
```
podsvirov: Realy 3.0.0-beta-3, not 3.0.0-beta-4?

I could be wrong, but it does not fit well with the changes in the headline.

xfxyjwf: This is not the version of the jruby runtime, but the Java version JRuby runtime depends on. Since beta-4 is not  deployed yet, jruby has to depend on an slightly older version of Java runtime to work.

PR #1787, opened 2016-07-14
Fix compatibility issues for the future GeneratedMessageV3 change.

xfxyjwf:
Currently some public API methods are defined in GenreatedMessage.java
and they have a generric return type:
```java
  class GeneratedMessage {
    class Builder<BuilderType extends Builder<BuilderType>> {
      public BuilderType setField(...);
      public BuilderType setExtension(...);
    }
  }
```
With these definitions, the compiled byte code of a callsite will have
a direct reference to GeneratedMessage. For example:
```java
  fooBuilder.setField(...);
```
becomes:
```
  ##: invokevirtual // Method Builder.setField:(...)LGeneratedMessage.Builder
  ##: checkcast     // class Builder
```

This will prevent us from updating generated classes to subclass a
different versioned GeneratedMessageV3 class in the future (we can't do
it in a binary compatible way).

This change addresses the problem by overriding these methods directly
in the generated class:
```java
  class Foo {
    class Builder extends GeneratedMessage.Builder<Builder> {
      public Builder setField(...) {
        return super.setField(...);
      }
    }
  }
```
After this, fooBuilder.setField(...) will be compiled to:
  ##: invokevirtual // Method Builder.setField:(...)LFoo.Builder

The callsites will no longer reference GeneratedMessage directly and we
can change Foo to subclass GeneratedMessageV3 without breaking binary
compatiblity.

The downside of this change is:
  1. It increases generated code size (though it saves some instructions
     on the callsites).
  2. We can never stop generating these overrides because doing that
     will break binary compatibility.

Change-Id: I879afbbc1325a66324a51565e017143489b06e97

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
+++ b/java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
@@ -1396,7 +1396,7 @@ public abstract class GeneratedMessage extends AbstractMessage
       return setExtension((ExtensionLite<MessageType, Type>) extension, value);
     }
     /** Set the value of an extension. */
-    public final <Type> BuilderType setExtension(
+    public <Type> BuilderType setExtension(
         final GeneratedExtension<MessageType, Type> extension, final Type value) {
       return setExtension((ExtensionLite<MessageType, Type>) extension, value);
     }
@@ -1407,7 +1407,7 @@ public abstract class GeneratedMessage extends AbstractMessage
       return setExtension((ExtensionLite<MessageType, List<Type>>) extension, index, value);
     }
     /** Set the value of one element of a repeated extension. */
-    public final <Type> BuilderType setExtension(
+    public <Type> BuilderType setExtension(
         final GeneratedExtension<MessageType, List<Type>> extension,
         final int index, final Type value) {
       return setExtension((ExtensionLite<MessageType, List<Type>>) extension, index, value);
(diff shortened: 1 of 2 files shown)


xfxyjwf:
@pherl to review

java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
```diff
@@ -1396,7 +1396,7 @@ private void verifyExtensionContainingType(
       return setExtension((ExtensionLite<MessageType, Type>) extension, value);
     }
     /** Set the value of an extension. */
-    public final <Type> BuilderType setExtension(
```
pherl: will removal of the final break the binary compatibility on callsites?

src/google/protobuf/compiler/java/java_message_builder.cc
```diff
@@ -438,6 +450,62 @@ GenerateCommonBuilderMethods(io::Printer* printer) {
     "\n",
     "classname", name_resolver_->GetImmutableClassName(descriptor_));
 
+  printer->Print(
```
pherl: these will increase code size. Are we planning to ifdef these out internally?

java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
xfxyjwf: No, removing final is binary compatible:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.17

src/google/protobuf/compiler/java/java_message_builder.cc
xfxyjwf: I will ifdef that out in internal code base, but I don't think code size is a big issue here. In this beta-4 release we will also generate equals() and hashCode() methods for all proto files. That has a much larger impact on code size compared to this change but yet we do not seem to have a problem with it in our internal roll-out.

pherl:
LGTM

PR #1790, opened 2016-07-15
Exposes the currently registered extensions for a message.

sergiocampama:


objectivec/GPBMessage.h
```diff
@@ -275,6 +275,9 @@ CF_EXTERN_C_END
 /// Return the descriptor for the message.
 - (GPBDescriptor *)descriptor;
 
+/// Returns an array with the currently set GPBExtensionDescriptors.
+- (NSArray *)extensionsCurrentlySet;
```
thomasvl: Do me a favor, kill sortedExtensionsInUse (here, the private header, and the use in Utilities).  Instead have utilities call this since you are exposing it and do the sort there.  We'll just tighten up the interfaces a little.

sergiocampama: done

thomasvl: I killed the previous travis run, and we'll confirm the second one passes.

PR #1791, opened 2016-07-15
Uses head version of rvm to avoid shell_update_session not found error

sergiocampama:
https://github.com/google/protobuf/issues/1786


tests.sh
```diff
@@ -238,6 +238,8 @@ build_objectivec_osx() {
 build_objectivec_cocoapods_integration() {
   # First, load the RVM environment in bash, needed to update ruby.
   source ~/.rvm/scripts/rvm
+  # Update rvm to the latest version.
```
thomasvl: Can you add why we need to do this?  (so we maybe can update it in the future to a fixed version)

sergiocampama: done

PR #1447, opened 2016-04-24
JavaScript: Make implicit defaults consistent with explicit defaults

seishun:
Currently, the JS implementation treats proto2 fields with specified defaults differently from fields without specified defaults.

Consider the following message definition:

```proto
message Test {
  optional string foo = 1;
  optional string bar = 2 [default = ""];
}
```

The difference can be observed in the following two cases:

1. If both `foo` and `bar` are unset, `getFoo()` returns `undefined`, but `getBar()` returns `""`.
2. If both `foo` and `bar` are unset, `toObject()` returns `{ foo: undefined, bar: "" }`.

However, `foo` and `bar` should be equivalent when unset per the [spec](https://developers.google.com/protocol-buffers/docs/proto#optional) (which would be consistent with the other implementations). I propose to unify their behavior in the following way:

1. If both `foo` and `bar` are unset, both `getFoo()` and `getBar()` return `""`.
2. If both `foo` and `bar` are unset, `toObject()` returns `{}`. This is for two reasons: 1) putting the default value in the object would cause loss of presence information; 2) properties with `undefined` values don't carry any useful information and just add noise to the object.

Coincidentally, this also fixes at least two bugs that I accidentally found while fixing tests:

1. `oneof` fields with specified defauls are always considered set when serializing. As a result, if the defaulted field comes after the field that was actually set, the wrong field is serialized.
2. Unset fields with a specified default are serialized as if the field were explicitly set to the default value.

One test fails ("debugTest testSimple1") because I'm not sure how `debug.dump` should work. Currently it includes unset fields in the returned object iff they have specified defaults. Since the presence of a specified default is no longer exposed to JS, `debug.dump` must either always include unset fields, or never include them.

P.S. Consider this a "feature request with proof-of-concept code attached", rather than a pull request proper.

PR #1794, opened 2016-07-15
JavaScript: Make implicit defaults consistent with explicit defaults

seishun:
Continuing from #1447, now targeting 3.0.0-beta-4. Also contains the following additional fixes:

* Added `has*` methods for message fields
* Added `clear*` methods for proto3 repeated fields
* Removed `has*` methods for proto2 repeated fields


xfxyjwf:
FYI. I cancelled all other travis tests except the javascript one for this PR.

haberman:
It looks like the tests aren't passing? https://travis-ci.org/google/protobuf/jobs/145128599

seishun:
```
\npm ERR! cb() never called!
```

Seems to be an npm issue. Perhaps try running `npm cache clean` or upgrading Node.js? 0.10.x will be EOL in 4 months.

js/message_test.js
```diff
@@ -184,14 +180,7 @@ describe('Message test suite', function() {
...
     // Test toObject
-    var expectedObject = {
-      stringField: defaultString,
-      boolField: true,
-      intField: 11,
-      enumField: 13,
-      emptyField: '',
-      bytesField: 'bW9v'
-    };
+    var expectedObject = {};
```
haberman: I don't think this is correct. toObject() should reflect default values.

js/message_test.js
```diff
@@ -145,11 +145,7 @@ describe('Message test suite', function() {
         undefined, undefined, undefined, undefined]);
     var result = foo.toObject();
     assertObjectEquals({
-      aString: undefined,
-      anOutOfOrderBool: undefined,
-      aNestedMessage: {
-        anInt: undefined
-      },
+      aNestedMessage: {},
```
haberman: I think toObject() should reflect the implicit defaults (0/""/false) just like the accessors would. I'm double-checking on this.

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -1259,9 +1258,10 @@ string GetPivot(const Descriptor* desc) {
 // value. See http://go/proto3#heading=h.kozewqqcqhuz for more information.
 bool HasFieldPresence(const FieldDescriptor* field) {
   return
-      (field->cpp_type() == FieldDescriptor::CPPTYPE_MESSAGE) ||
+      !field->is_repeated() &&
```
haberman: This conditional is getting a little complicated. Let's break it apart, like:

```c++
if (field->is_repeated()) {
  return false;
}

return /* ... */
```

js/message_test.js
seishun: That would result in a loss of presence information. Note that currently toObject does preserve presence information for fields without explicit defaults.

src/google/protobuf/compiler/js/js_generator.cc
seishun: Agreed.

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -2349,6 +2341,20 @@ void Generator::GenerateClassField(const GeneratorOptions& options,
         "clearedvalue", (field->is_repeated() ? "[]" : "undefined"),
         "returnvalue", JSReturnClause(field));
 
+    printer->Print(
```
haberman: If there is a clean way, let's try to avoid duplicating this code. At the least, factor it into a shared function.

seishun:
Simplified HasFieldPresence and deduplicated the has* method generation.

js/message_test.js
haberman: After chatting a bit internally, I think our best option for the moment would be to preserve existing semantics. So all assertions related to `toObject()` should be unchanged. We can consider changing this later.

seishun: The existing semantics are different for fields with and without explicit defaults. I assume you mean the former? Making sure because some assertions will need to be changed either way.

haberman: > Making sure because some assertions will need to be changed either way.

Why? I am suggesting to keep exactly the current semantics, where fields with and without explicit defaults behave differently. That shouldn't require changing any assertions.

haberman:
The beta-4 branch has just been merged to master, so `has*` methods are on master now. Could you rebase your work on `master` and update the PR to be based on master? Thanks.

js/message_test.js
seishun: Oh, I see. That will make this PR effectively incomplete, but sure. I hope to address this again later.

seishun:
Moving back to #1447.

PR #1796, opened 2016-07-15
Add python protoc package

anandolee:
Going to provide protoc package in pypi. Only python and c++ code gen are included. The package is currently under grpc tools: https://github.com/grpc/grpc/tree/master/tools/distrib/python/grpcio_tools

After install, user can call
$ python -m google.protoc.protoc -I$INCLUDE --python_out=$OUTPUT $PROTO_FILES

### DIFF
--- a/python/protoc/google/protoc/protoc.py
+++ b/python/protoc/google/protoc/protoc.py
@@ -1,0 +1,41 @@
+#! /usr/bin/env python
+#
+# Protocol Buffers - Google's data interchange format
+# Copyright 2008 Google Inc.  All rights reserved.
+# https://developers.google.com/protocol-buffers/
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+#     * Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above
+# copyright notice, this list of conditions and the following disclaimer
+# in the documentation and/or other materials provided with the
+# distribution.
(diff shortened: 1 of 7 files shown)


python/protoc/google/protoc/protoc.py
```diff
@@ -0,0 +1,42 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import pkg_resources
+import sys
+
+import google.protoc.protoc_compiler as protoc_compiler
+
+if __name__ == '__main__':
+  proto_include = pkg_resources.resource_filename('google.protoc', '')
+  print proto_include
```
benjigoldberg: wrap in parens for Python 3 support

benjigoldberg:
+1 would love to see this in PyPi

anandolee:
@haberman can you help to review?

haberman:
Sorry I missed this before.

Is there a way to link against libprotoc instead of duplicating the list of .cc files?

python/protoc/setup.py
```diff
@@ -0,0 +1,101 @@
...
+      language='c++',
+      define_macros=[('HAVE_PTHREAD', 1)],
+  )
+  return plugin_ext
+
+def maybe_cythonize(exts):
+  from Cython import Build
+  return Build.cythonize(exts)
+
+setuptools.setup(
```
pfreixes: Why not distibute the  `protoc` together with the protobuf library? This will automatically align the versions of both packages. Otherwise, it has to be done by the user in his requirements file.

pfreixes:
any news on that?

PR #1801, opened 2016-07-18
Use public methods to fetch oneofs in generated code.

thomasvl:
When building into frameworks, the generated code doesn't always have direct
access to the proto internals.  Instead of opening up the access, just use the
public method to fetch the correct oneof.

Fixes https://github.com/google/protobuf/issues/1789


sergiocampama:
lgtm

PR #1810, opened 2016-07-19
Versioning Java GeneratedMessage class

xfxyjwf:
@pherl to review

Should be the last required Java change before 3.0.0 GA.

Deleted (Single|Repeated)FieldBuilderTest.java because (Single|Repeated)FieldBuilder can no longer work with the generated code after this change. The versioned (Single|Repeated)FieldBuilderV3 classes are used in the generated code starting from beta-4.


src/google/protobuf/compiler/java/java_message.cc
```diff
@@ -205,9 +205,10 @@ GenerateFieldAccessorTable(io::Printer* printer, int* bytecode_estimate) {
   } else {
     vars["final"] = "";
   }
+  vars["ver"] = GeneratedCodeVersionSuffix();
   printer->Print(vars,
     "$private$static $final$\n"
-    "  com.google.protobuf.GeneratedMessage.FieldAccessorTable\n"
+    "  com.google.protobuf.GeneratedMessage$ver$.FieldAccessorTable\n"
```
pherl: will we have test to cover those cases for internal changes?

pherl:
LG

PR #1812, opened 2016-07-20
Use the dotnet-release package feed for Travis.

jskeet:
See https://github.com/dotnet/core/issues/227 for background.


xfxyjwf:
LGTM

xfxyjwf:
I noticed the error in protobuf travis tests as well, but it disappeared after a while.

jskeet:
It's possible that they reverted the package breakage somewhere, but using the new feed is probably a good thing anyway.

PR #1822, opened 2016-07-21
Update travis tests for Java.

xfxyjwf:
1. Set JAVA_HOME so mvn can pick up the correct java version.
2. Remove jdk6 tests. It has been broken for a while and remain undetected as
   mvn is actually using java 7 to build the code. Given that we have
   set -source and -target to 6 in the pom.xml and the built .jar
   should be usable by java 6, having a dedicated java 6 test doesn't
   seem necessary (assuming very few Java 6 users want to compile protobuf
   Java from source).

@pherl to review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,8 @@ env:
   - CONFIG=cpp
   - CONFIG=cpp_distcheck
   - CONFIG=golang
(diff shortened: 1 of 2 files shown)


pherl:
LGTM

PR #1829, opened 2016-07-22
Fix sign-comparison warnings and add a test for that.

xfxyjwf:
Fixes https://github.com/google/protobuf/issues/1813

Also added a test to ensure public header files and protobuf generated code doesn't have these warnings. This test doesn't cover protobuf's own implementation code. We may want to update all of protobuf's code to be free of this warning in the future but right now there are just too many of them and require more work to cleanup.


xfxyjwf:
@pherl to review

src/Makefile.am
```diff
@@ -122,7 +124,6 @@ nobase_include_HEADERS =                                        \
   google/protobuf/reflection.h                                  \
   google/protobuf/reflection_ops.h                              \
   google/protobuf/repeated_field.h                              \
-  google/protobuf/repeated_field_reflection.h                   \
```
pherl: why this is removed?

xfxyjwf: It's deleted internally, but the deletion is lost from internal to opensource integration. See http://cl/71942508

xfxyjwf: Ops, wrong cl number. Should be this one: http://cl/86669259

pherl:
LGTM

PR #1837, opened 2016-07-25
Ruby: generated foo.proto -> foo_pb.rb instead of foo.rb.

haberman:
This brings us more into line with other langauges, and makes it more
obvious when we are requiring protobuf generated code.


haberman:
Review to @acozzette.

I've been meaning to do this for a while. This change will be a little intrusive, but it seems worth it for the long-term.

acozzette:
LGTM

PR #1851, opened 2016-07-26
Fixes traivs cpp build.

xfxyjwf:
Remove the use of INT64_C/UINT64_C and add static_cast when necessary.

@pherl to review.

Turns out static_cast is not needed when assigning a xxULL value to int64_t in most places as the compiler is smart enough to figure out a GOOGLE_ULONGLONG(xxx) value actually can fit into a int64_t type.


pherl:
LGTM

PR #1861, opened 2016-07-27
Bring C#'s ToPascalCase method in line with C++.

jskeet:
I'd been hoping this would get our JSON conformance test failures down, but it doesn't.
At least we're now consistent with the C++ code.

This code is called in two cases:

- If we have a proto descriptor that doesn't have the json_name field for some reason
- When formatting a field mask

It's important that this is consistent with other platforms; I hadn't noticed when the C++ code it was ported from changed.


pherl:
LG

PR #1871, opened 2016-07-28
Fix build protoc script for windows

pherl:
Mingwn32 doesn't support multiple thread compiling (-j4).
Fix the duplicated.exe prefix on windows.


xfxyjwf:
LGTM

PR #1878, opened 2016-07-29
Ruby: added API support for well-known types.

haberman:
Review to @raggi.

### DIFF
--- a/ruby/lib/google/protobuf/well_known_types.rb
+++ b/ruby/lib/google/protobuf/well_known_types.rb
@@ -90,6 +90,8 @@ module Google
       end
     end
 
+    class UnexpectedStructType < Google::Protobuf::Error; end
+
     Value.class_eval do
       def to_ruby(recursive = false)
         case self.kind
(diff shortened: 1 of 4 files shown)


ruby/lib/google/protobuf/well_known_types.rb
```diff
@@ -0,0 +1,206 @@
...
+require 'google/protobuf/any_pb'
+require 'google/protobuf/duration_pb'
+require 'google/protobuf/field_mask_pb'
+require 'google/protobuf/struct_pb'
+require 'google/protobuf/timestamp_pb'
+
+module Google
+  module Protobuf
+
+    Any.class_eval do
```
raggi: You can just do `class Any` here, there's no need for the block I think.

ruby/lib/google/protobuf/well_known_types.rb
```diff
@@ -0,0 +1,206 @@
...
+        when :null_value
+          nil
+        when :number_value
+          self.number_value
+        when :string_value
+          self.string_value
+        when :bool_value
+          self.bool_value
+        else
+          raise "Value not set"
```
raggi: How rare is this case?

ruby/lib/google/protobuf/well_known_types.rb
haberman: True, I wrote it this way because I like the fact that it will error out if "Any" wasn't already defined. It makes it more explicit that it's adding methods to an existing class and not defining a new one.

ruby/lib/google/protobuf/well_known_types.rb
```diff
@@ -0,0 +1,206 @@
...
+        end
+      end
+
+      def from_ruby(value)
+        if value.nil?
```
raggi: You may, if you want to, reformat this like so:

``` ruby
case value
when NilClass
  ...
when Numeric
  ...
when String
  ...
...
else
  raise ...
end
```

raggi:
Overall, this looks good. The raise cases appear to be untested, and the exceptions are untyped. If they're rare enough (like "should never happen" cases) I'm ok with that, but if they're not essentially impossible we should add concrete types and tests for them.

Other comments are minor style things that you can essentially take or leave.

ruby/lib/google/protobuf/well_known_types.rb
raggi: Ok, sounds good!

ruby/lib/google/protobuf/well_known_types.rb
haberman: Done.

ruby/lib/google/protobuf/well_known_types.rb
haberman: Rare, but still should be tested. I added a test.

haberman:
Addressed your comments. I'll plan to submit when tests are green. Thanks!

PR #1882, opened 2016-07-29
Be consistent with the use of CHECK()/ GOOGLE_CHECK().

legrosbuffle:


xfxyjwf:
Thanks for the patch!

PR #1888, opened 2016-07-29
Fix the build-zip.sh to add .exe for win packages.

### LINKED ISSUE #1881: Missing binary suffix (.exe) on win32 build
The win32 build zip file includes the protoc executable, but is missing the .exe suffix on the filename so is not directly executable on Windows.

https://github.com/google/protobuf/releases/download/v3.0.0/protoc-3.0.0-win32.zip

protoc should be renamed to protoc.exe

pherl:
Fixes #1881


xfxyjwf:
LGTM

PR #1903, opened 2016-08-01
Use 64-bit protoc binaries in compatibility tests.

xfxyjwf:
Turns out travis by default doesn't have the needed dependencies to run a 32-bit
binary.


pherl:
LGTM

PR #1907, opened 2016-08-02
JS: Fix for high utf-8 codepoints.

tetek:
Now you can enjoy emojis on iOS. 👍

### DIFF
--- a/js/binary/decoder.js
+++ b/js/binary/decoder.js
@@ -941,7 +941,8 @@ jspb.BinaryDecoder.prototype.readString = function(length) {
       codeUnits.push(high, low)
     }
   }
-
+  // String.fromCharCode.apply is faster than manually appending characters on
+  // Chrome 25+, and generates no additional cons string garbage.
   var result = String.fromCharCode.apply(null, codeUnits);
   this.cursor_ = cursor;
   return result;
--- a/js/binary/decoder.js
+++ b/js/binary/decoder.js
@@ -895,11 +895,9 @@ jspb.BinaryDecoder.prototype.readEnum = function() {
 
 /**
  * Reads and parses a UTF-8 encoded unicode string from the stream.
- * The code is inspired by maps.vectortown.parse.StreamedDataViewReader, with
- * the exception that the implementation here does not get confused if it
- * encounters characters longer than three bytes. These characters are ignored
- * though, as they are extremely rare: three UTF-8 bytes cover virtually all
- * characters in common use (http://en.wikipedia.org/wiki/UTF-8).
+ * The code is inspired by maps.vectortown.parse.StreamedDataViewReader.
+ * Supports codepoints from U+0000 up to U+10FFFF. 
+ * (http://en.wikipedia.org/wiki/UTF-8).
  * @param {number} length The length of the string to read.
  * @return {string} The decoded string.
  */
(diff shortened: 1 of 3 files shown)


tetek:
I signed it!

haberman:
Thanks for the fix. Could you add a test?

Also do we need a fix in the decode path or does that work already?

tetek:
@haberman A fix for decoder was also needed.
I'm not familiar with how to run tests here.
I think I could add an test for both encoding and decoding to `decoder_test.js` since `encoder_test.js` doesn't exist. How do you run it then?

haberman:
Sorry for letting this sit for a while!

The tests should be easy to run -- see the README for more information: https://github.com/google/protobuf/tree/master/js#setup

The fixes look great, we just need a test to verify that they don't regress.

tetek:
@haberman test added.

js/binary/decoder.js
```diff
@@ -895,42 +895,57 @@ jspb.BinaryDecoder.prototype.readEnum = function() {
...
+      low += 56320;
+      codepoints.push(high);
+      codepoints.push(low);
     }
   }
 
-  // String.fromCharCode.apply is faster than manually appending characters on
-  // Chrome 25+, and generates no additional cons string garbage.
-  var result = String.fromCharCode.apply(null, chars);
+  var result = String.fromCodePoint.apply(null, codepoints);
```
haberman: This function does not seem to have wide support. According to this it is not available at all on IE or Safari.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint

If that's true we'll need to take a different approach. The polyfill on that page indicates that you can append the two surrogates individually to the string using `String.fromCharCode`. You are using the surrogates already, so changing this line to String.fromCodePoint() should not be required.

js/binary/decoder.js
```diff
@@ -895,42 +895,57 @@ jspb.BinaryDecoder.prototype.readEnum = function() {
...
+    } else if (c < 248) { // UTF-8 with 4 bytes.
+      var c2 = bytes[cursor++];
+      var c3 = bytes[cursor++];
+      var c4 = bytes[cursor++];
+      // Characters written on 4 bytes have 21 bits for a codepoint. 
+      // We can't fit that on 16bit characters, so we use surrogates.
+      var codepoint = ((c & 7) << 18) | ((c2 & 63) << 12) | ((c3 & 63) << 6) | (c4 & 63);
+      // Surrogates formula from wikipedia.
+      // 1. Subtract 0x10000 from codepoint
+      codepoint -= 65536;
```
haberman: If the comments are referencing hex constants, I think it makes sense to use hex constants in the code also.

js/binary/decoder.js
```diff
@@ -895,42 +895,57 @@ jspb.BinaryDecoder.prototype.readEnum = function() {
...
+      // 2. Split this into the high 10-bit value and the low 10-bit value
+      var low = codepoint & 1023;
+      var high = (codepoint >> 10) & 1023;
+      // 3. Add 0xD800 to the high value to form the high surrogate
+      high += 55296;
+      // 4. Add 0xDC00 to the low value to form the low surrogate:
+      low += 56320;
+      codepoints.push(high);
+      codepoints.push(low);
```
haberman: Both code units can be added in a single push call `push(high, low)`.

js/binary/decoder.js
```diff
@@ -895,42 +895,57 @@ jspb.BinaryDecoder.prototype.readEnum = function() {
...
+ * (http://en.wikipedia.org/wiki/UTF-8).
  * @param {number} length The length of the string to read.
  * @return {string} The decoded string.
  */
 jspb.BinaryDecoder.prototype.readString = function(length) {
   var bytes = this.bytes_;
   var cursor = this.cursor_;
   var end = cursor + length;
-  var chars = [];
+  var codepoints = [];
```
haberman: Since this array uses UTF-16 surrogates instead of actual code points, I think this would better be named `codeUnits`.

js/binary/encoder.js
```diff
@@ -412,16 +412,22 @@ jspb.BinaryEncoder.prototype.writeString = function(value) {
 
   // UTF16 to UTF8 conversion loop swiped from goog.crypt.stringToUtf8ByteArray.
   for (var i = 0; i < value.length; i++) {
-    var c = value.charCodeAt(i);
+    var c = value.codePointAt(i);
```
haberman: This appears to not be supported in Safari:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt

According to the polyfill on that page, you can use `charCodeAt` to get the individual surrogates. You'll need to add some code below to decode the surrogates.

tetek:
@haberman thanks for the review. All suggestions applied.

js/binary/decoder_test.js
```diff
@@ -209,7 +209,30 @@ describe('binaryDecoderTest', function() {
...
+  
+  /**
+   * Test encoding and decoding utf-8.
+   */
+   it('testUtf8', function() {
+    var encoder = new jspb.BinaryEncoder();
 
+    var ascii = "ASCII should work in 3, 2, 1..."
+    var utf8_two_bytes = "©";
+    var utf8_tree_bytes = "❄";
```
seishun: Typo here.

tetek:
fixed typo, squashed commits

tetek:
@haberman requested changes have been added.

haberman:
ok to test

js/binary/encoder.js
```diff
@@ -409,19 +409,34 @@ jspb.BinaryEncoder.prototype.writeFixedHash64 = function(hash) {
...
 jspb.BinaryEncoder.prototype.writeString = function(value) {
   var oldLength = this.buffer_.length;
-
-  // UTF16 to UTF8 conversion loop swiped from goog.crypt.stringToUtf8ByteArray.
+ 
   for (var i = 0; i < value.length; i++) {
+    
     var c = value.charCodeAt(i);
+    // Look for surrogates
+    if (c >= 0xD800 && c <= 0xDBFF && i + 1 < value.length) {
```
haberman: Please put this check inside the branch `else if (c < 65536)`. There is no reason to check this in the `c < 128` and `c < 2048` cases.

js/binary/decoder.js
```diff
@@ -895,42 +895,54 @@ jspb.BinaryDecoder.prototype.readEnum = function() {
...
+      // 2. Split this into the high 10-bit value and the low 10-bit value
+      // 3. Add 0xD800 to the high value to form the high surrogate
+      // 4. Add 0xDC00 to the low value to form the low surrogate:
+      var low = (codepoint & 1023) + 0xDC00;
+      var high = ((codepoint >> 10) & 1023) + 0xD800;
+      codeUnits.push(high, low)
     }
   }
 
-  // String.fromCharCode.apply is faster than manually appending characters on
```
haberman: Please re-add this comment; it provides some rationale about why we use String.fromCharCode.apply instead of appending characters one by one.

Issue #1035, opened 2015-12-09
Getting Compilation error in protobuf 3.0.0 beta-1


bijaya012:
I am getting below compilation error.
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I.. -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -O2 -g -DNDEBUG -MT google/protobuf/dynamic_message.lo -MD -MP -MF google/protobuf/.deps/dynamic_message.Tpo -c google/protobuf/dynamic_message.cc  -fPIC -DPIC -o google/protobuf/.libs/dynamic_message.o
./google/protobuf/stubs/hash.h: In instantiation of 'google::protobuf::hash<long long int>':
./google/protobuf/map.h:839:   instantiated from here
./google/protobuf/stubs/hash.h:269: error: invalid use of undefined type 'struct std::tr1::hash<long long int>'
/usr/lib/gcc/x86_64-redhat-linux/4.1.2/../../../../include/c++/4.1.2/tr1/functional:1101: error: declaration of 'struct std::tr1::hash<long long int>'
./google/protobuf/map.h: In member function 'size_t google::protobuf::hashgoogle::protobuf::MapKey::operator()(const google::protobuf::MapKey&) const':
./google/protobuf/map.h:839: error: no match for call to '(google::protobuf::hash<long long int>) (google::protobuf::int64)'
./google/protobuf/stubs/hash.h: At global scope:
./google/protobuf/stubs/hash.h: In instantiation of 'google::protobuf::hash<long long unsigned int>':
./google/protobuf/map.h:843:   instantiated from here
./google/protobuf/stubs/hash.h:269: error: invalid use of undefined type 'struct std::tr1::hash<long long unsigned int>'
/usr/lib/gcc/x86_64-redhat-linux/4.1.2/../../../../include/c++/4.1.2/tr1/functional:1101: error: declaration of 'struct std::tr1::hash<long long unsigned int>'
./google/protobuf/map.h: In member function 'size_t google::protobuf::hashgoogle::protobuf::MapKey::operator()(const google::protobuf::MapKey&) const':
./google/protobuf/map.h:843: error: no match for call to '(google::protobuf::hash<long long unsigned int>) (google::protobuf::uint64)'
[...]

PR #1913, opened 2016-08-03
Fix gcc 4.1 build (#1035)

matthauck:


matthauck:
This should fix #1035

matthauck:
_ping_ 

Is there anything missing here? Any changes requested?

bazel-io:

TeBoring:
Can you rebase this pull request on the current master?

matthauck:
Just finished doing that and making sure it still built. :)

TeBoring:
ok to test.

TeBoring:
LGTM

src/google/protobuf/map.h
```diff
@@ -1726,12 +1726,19 @@ struct hash<google::protobuf::MapKey> {
...
       case google::protobuf::FieldDescriptor::CPPTYPE_INT64:
         return hash< ::google::protobuf::int64>()(map_key.GetInt64Value());
-      case google::protobuf::FieldDescriptor::CPPTYPE_INT32:
-        return hash< ::google::protobuf::int32>()(map_key.GetInt32Value());
       case google::protobuf::FieldDescriptor::CPPTYPE_UINT64:
         return hash< ::google::protobuf::uint64>()(map_key.GetUInt64Value());
+#else
+      case google::protobuf::FieldDescriptor::CPPTYPE_INT64:
+      case google::protobuf::FieldDescriptor::CPPTYPE_UINT64:
+        GOOGLE_LOG(FATAL) << "Unsupported on this platform.";
```
xfxyjwf: This is not the right use of LOG(FATAL). Protobuf library should not crash an user's program given a valid input. If the platform can not be supported, it should be reported at compile time. Though, in this case it seems easy enough to write a custom hash function if it's not already available in the platform headers. @matthauck @TeBoring Do you plan to actually implement this hash for int64? I saw this change while working on integrating github changes to our internal repo and I'm going to leave this change out.

matthauck: Oh, whoops, sorry. 😞 

The thought did not cross my mind to implement this on these platforms. I'd be open to helping out if you could give me some pointers on where to start.

TeBoring: I guess just an empty implementation would be enough. It's better to trigger compile error if the hash for int64 is used.

TeBoring: https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/hash.h#L184
Here is an example. For int64, how about just trimming?

matthauck: Okay, opened up two PR's to fix this missing hash: #2906 and #2907. Take your pick which fix you like better. :)

Issue #1885, opened 2016-07-29
`jspb.Map` not exported


Triztian:
The `jspb.Map` object is not exported in the case of CommonJS modules. Should it be added: after here: https://github.com/google/protobuf/blob/master/js/commonjs/export.js#L13

haberman:
Why does the Map symbol need to be available to clients? It's not intended to be constructed by generated code or user code. It's only constructed inside `jspb.Message.getMapField()`.

Triztian:
It is called when trying to access map fields of any generated proto buf and even if `google-protobuf` package is added as a dependency of the npm project that hosts the generated protos the `getMapField` function is undefined.

haberman:
Can you paste the error that you are getting?

gegles:
I concur with @Triztian a 100%. I was having the same issue (when serializing/deserializing a message including a map<string, string>).

Here is the error I was getting (deserializing):

Cannot read property 'deserializeBinary' of undefined

The location of the error is this line:

jspb.Map.deserializeBinary(message, reader, jspb.BinaryReader.prototype.readString, jspb.BinaryReader.prototype.readString);

After adding the following to export.js:

```javascript
goog.require('jspb.Map');
exports.Map = jspb.Map;
```

Everything works fine.

Thanks for fixing this.

PS: I could submit the patch, but I am not knowledgeable enough to know whether it is the best way to solve this problem.

haberman:
Did the error have a stack trace? Or do you have a way to reproduce this?

I want to understand the error better so I can know the best way to fix it.

gegles:
It did, but it is all from within the generated pb code which defines my messages. Here is the relevant block:

```javascript
	proto.aspera.art.ResponseMessage.deserializeBinaryFromReader = function(msg, reader) {
	  while (reader.nextField()) {
	    if (reader.isEndGroup()) {
	      break;
	    }
	    var field = reader.getFieldNumber();
	    switch (field) {
	    case 1:
	      var value = /** @type {number} */ (reader.readUint32());
	      msg.setStatusCode(value);
	      break;
	    case 2:
	      var value = msg.getHeadersMap();
	      reader.readMessage(value, function(message, reader) {
	        jspb.Map.deserializeBinary(message, reader, jspb.BinaryReader.prototype.readString, jspb.BinaryReader.prototype.readString);
	         });
	      break;
	    default:
	      reader.skipField();
	      break;
	    }
	  }
	  return msg;
	};
```

[...]

PR #1920, opened 2016-08-04
Fix missing import of jspb.Map (#1885)

gegles:
Fix missing import of jspb.Map (#1885)

### DIFF
--- a/js/commonjs/export.js
+++ b/js/commonjs/export.js
@@ -11,7 +11,9 @@ goog.require('jspb.BinaryWriter');
 goog.require('jspb.ExtensionFieldBinaryInfo');
 goog.require('jspb.ExtensionFieldInfo');
 goog.require('jspb.Message');
+goog.require('jspb.Map');
 
+exports.Map = jspb.Map;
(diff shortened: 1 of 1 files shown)


gegles:
I signed it!

haberman:
Could you take a quick look and see if we have any tests that serialize/deserialize a map? If not, let's add one and hopefully it will exercise this code path.

gegles:
@haberman, I had looked for it already, but could not find any. I am pretty new to codebase, where would you want me to add one? Can you point me to the right file/section?

haberman:
Ah, I think I found the problem. Internally we have a `maps_test.js` that tests this, but I missed that file when we open-sourced the new maps code.

For now let's just add a quick test in `js/binary/proto_test.js`. You can use the `TestMapFields` message from `testbinary.proto`. Let's just make sure a message with some map data can round-trip to binary format and back.

Thanks and sorry for the bug!

haberman:
I'll merge this for now and we can worry about a test later.

PR #1922, opened 2016-08-04
JavaScript: Make implicit defaults consistent with explicit defaults in toObject

seishun:
This was originally part of #1447, but it was decided to postpone this part. The relevant parts of the description of the previous PR follow.

Currently, the JS implementation treats proto2 fields with specified defaults differently from fields without specified defaults in `toObject`.

Consider the following message definition:

``` proto
message Test {
  optional string foo = 1;
  optional string bar = 2 [default = ""];
}
```

If both `foo` and `bar` are unset, `toObject()` returns `{ foo: undefined, bar: "" }`.

However, `foo` and `bar` should be equivalent when unset per the [spec](https://developers.google.com/protocol-buffers/docs/proto#optional) (which would be consistent with the other implementations). I propose to unify their behavior in the following way:

If both `foo` and `bar` are unset, `toObject()` returns `{}`. This is for two reasons: 1) putting the default value in the object would cause loss of presence information; 2) properties with `undefined` values don't carry any useful information and just add noise to the object.

### DIFF
--- a/js/message_test.js
+++ b/js/message_test.js
@@ -145,11 +145,7 @@ describe('Message test suite', function() {
         undefined, undefined, undefined, undefined]);
     var result = foo.toObject();
     assertObjectEquals({
-      aString: undefined,
-      anOutOfOrderBool: undefined,
-      aNestedMessage: {
-        anInt: undefined
-      },
+      aNestedMessage: {},
       // Note: JsPb converts undefined repeated fields to empty arrays.
       aRepeatedMessageList: [],
       aRepeatedStringList: []
@@ -184,14 +180,7 @@ describe('Message test suite', function() {
     var response = new proto.jspb.test.DefaultValues();
 
     // Test toObject
-    var expectedObject = {
-      stringField: defaultString,
-      boolField: true,
-      intField: 11,
-      enumField: 13,
-      emptyField: '',
-      bytesField: 'bW9v'
-    };
+    var expectedObject = {};
     assertObjectEquals(expectedObject, response.toObject());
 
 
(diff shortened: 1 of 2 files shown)


seishun:
Any update on this?

seishun:
Rebased on master.

acozzette:
@seishun, I would worry that no longer including default values in the `toObject()` output could likely break a lot of code that depends on those defaults. Omitting properties that are just `undefined` sounds like it might be relatively safe, though.

seishun:
@acozzette what would you say about moving default values to the prototype? That way, `obj.hasOwnProperty('some_unset_field')` would be `false`, but code that just does `obj.some_unset_field` would still get the default value.

To be clear, `toObject` would do something like this

```js
var defaults = { field_1: 5, field_2: 6 };
var obj = Object.create(defaults);
// populate set fields only
return obj;
```

instead of

```js
var obj = {};
// populate all fields (current master) or just set fields (this PR)
return obj;
```

Following this approach, `obj.unset_field_without_specified_default` would be equal to the type-specific default instead of the current `undefined`. This seems less likely to break a lot of code though.

acozzette:
ok to test

anandolee:
@acozzette Is it ready to merge ?

acozzette:
@seishun Sorry I haven't replied on this pull request in a long time--are you still interested in making this change or something similar? I see there are now some merge conflicts that would have to be resolved first.

seishun:
I'm still interested, but I never got a response to my question (https://github.com/google/protobuf/pull/1922#issuecomment-257714324). If you're okay with the behavior I originally proposed, then I'll look into resolving the merge conflicts.

acozzette:
OK, I don't have a strong opinion on this because I haven't been working on JS for a while. @TeBoring, do you have any thoughts on this?

TeBoring:
Internally, we have a file option called ```use_broken_proto2_semantics```. Currently, it's true by default. Behavior in this PR should only be enabled when ```use_broken_proto2_semantics``` is false. I would suggest we ```opensource use_broken_proto2_semantics``` first.

acozzette:
@TeBoring Ok, maybe we should hold off on making changes to `toObject` for now then. I'm also not sure what we want the future of `toObject` and `fromObject` to be--those functions are a bit strange because they produce output similar to proto3 JSON but without quite matching that format, so perhaps we should try to eventually make it match proto3 JSON.

PR #1929, opened 2016-08-08
Update TextFormat to handle unknown Enum values

dmap:
Currently `TextFormat.java` (and presumably the equivalent implementation in other languages) does not handle unknown enum values at all well.  When printed they produce a string value `UNKNOWN_ENUM_VALUE_NestedEnum_XXXX` which cannot then be parsed again.  So a perfectly valid proto3 message with an unknown enum value cannot be round tripped to a string a back again using `TextFormat`.

I have updated the java `TextFormat` implementation to instead handle unknown Enum values in precisely the same way as they are handled by the `JsonFormat` implementation i.e. they are printed as simple number values and parsed successfully back to unknown Enum values for proto3 messages.

### DIFF
--- a/java/core/src/test/java/com/google/protobuf/UnknownEnumValueTest.java
+++ b/java/core/src/test/java/com/google/protobuf/UnknownEnumValueTest.java
@@ -229,7 +229,7 @@ public class UnknownEnumValueTest extends TestCase {
     assertFalse(message.equals(differentMessage));
   }
   
-  public void testUnknownEnumValuesInTextFormat() {
+  public void testUnknownEnumValuesInTextFormat() throws ParseException {
     TestAllTypes.Builder builder = TestAllTypes.newBuilder();
     builder.setOptionalNestedEnumValue(4321);
     builder.addRepeatedNestedEnumValue(5432);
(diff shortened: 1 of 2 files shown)


dmap:
I should mention, I haven't made the equivalent changes in the implementation of TextFormat for the other languages.

I also realise this isn't the only way of the fixing the problem.  If maintaining the funky `UNKNOWN_ENUM_VALUE_NestedEnum_XXXX` string is important then just adding support for parsing from that string is also acceptable to me.

dmap:
I signed it!

java/core/src/test/java/com/google/protobuf/UnknownEnumValueTest.java
```diff
@@ -239,17 +239,26 @@ public void testUnknownEnumValuesInTextFormat() {
...
     assertEquals(
-        "optional_nested_enum: UNKNOWN_ENUM_VALUE_NestedEnum_4321\n"
-        + "repeated_nested_enum: UNKNOWN_ENUM_VALUE_NestedEnum_5432\n"
-        + "packed_nested_enum: UNKNOWN_ENUM_VALUE_NestedEnum_6543\n", textData);
+        "optional_nested_enum: 4321\n"
+        + "repeated_nested_enum: 5432\n"
+        + "packed_nested_enum: 6543\n", textData);
     
     // Parsing unknown enum values will fail just like parsing other kinds of
     // unknown fields.
```
xfxyjwf: Remove this comment?

xfxyjwf:
Thanks for the patch! I don't remember why I made the text format behavior different from JsonFormat. The one in JsonFormat is more aligned with the proto3 unknown enum semantics and makes more sense to me now.

dmap:
Another note on my current implementation.  It differs slightly from `JsonFormat` in that it still does the `tokenizer.lookingAtInteger()` test to decide whether to attempt to resolve the enum value by name or by number.  `JsonFormat` always try to resolve by name first and then only attempts to resolve by number if that fails.  This means that if an enum happens to have an integer string (e.g. '1') as a name and those names don't have the same numeric value then the wrong value could get resolved.  

This is clearly an edge case but one that can be avoided by always attempting to resolve by name first (like `JsonFormat` does).

anandolee:
@xfxyjwf  and @dmap 
What's the status of the PR?

xfxyjwf:
Feel free to resend the PR.

PR #1950, opened 2016-08-12
Fixes extra whitespace on generated comments.

sergiocampama:


src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -868,7 +870,12 @@ string BuildCommentsString(const SourceLocation& location,
     // Decouple / from * to not have inline comments inside comments.
     line = StringReplace(line, "/*", "/\\*", true);
     line = StringReplace(line, "*/", "*\\/", true);
-    final_comments += prefix + line + suffix;
+    line = prefix + line;
+    StripWhitespace(&line);
+    // If not a one line, need to add the first space before *, as
+    // StripWhitespace would have removed it.
+    line = (is_one_line ? "" : " ") + line;
+    final_comments += line + suffix;
```
thomasvl: What is we drop the \n from both suffixes, then I think you can just do:

line = prefix + line + suffix
StripWhiteSpace(&line)
final_comments += line + "\n"

And there's isn't the special knowledge about oneline needing the space back.

thomasvl: Scratch that, it still eats the leading space.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -850,11 +850,13 @@ string BuildCommentsString(const SourceLocation& location,
...
 
-  if (prefer_single_line && lines.size() == 1) {
+  bool is_one_line = prefer_single_line && lines.size() == 1;
+
+  if (is_one_line) {
     prefix = "/** ";
     suffix = " */\n";
   } else {
-    prefix = " * ";
+    prefix = "* ";
```
thomasvl: ok, instead of overloading is_one_line, add bool add_leading_space and set it to true here and false in the oneline case, and test that where you do the prefix.  Then things are a little more clear.

sergiocampama: Done

PR #1951, opened 2016-08-12
JavaScript: Add protoc option to preserve field names in toObject

seishun:
Currently, `toObject` returns an object with field names converted to camelCase. However, in some cases the original field names are preferable. This adds a `protoc` option `preserve_proto_fieldnames` that keeps them as-is, and also doesn't append "List" or "Map".

I'll add tests if/when I get a go-ahead on the general idea (and perhaps some suggestion on how the test should work).

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -475,33 +475,43 @@ bool IgnoreOneof(const OneofDescriptor* oneof) {
 }
 
 string JSIdent(const GeneratorOptions& options, const FieldDescriptor* field,
-               bool is_upper_camel, bool is_map, bool drop_list) {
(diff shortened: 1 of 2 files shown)


seishun:
Any update on this?

seishun:
Rebased on master.

haberman:
Sorry I missed this for so long.

Generally we try to keep codegen options to a minimum. We only introduce them when there is a very compelling need. In particular, anything that changes the semantic behavior of the library is not a good candidate for a codegen option.

[proto3 JSON does support preserving original field names](https://developers.google.com/protocol-buffers/docs/proto3#json_options), but [JavaScript doesn't support true proto3 JSON yet](https://github.com/google/protobuf/issues/1904).

I think the best way forward for this is to add true proto3 JSON support, and then add a runtime option to the serializer to use original field names. Sorry I don't have a better answer for you.

PR #1952, opened 2016-08-12
Fixed wrong package being exported

lukewink:
The correct package is com.google.protobuf.nano


lukewink:
I signed it

anandolee:
nano is not supported here anymore

PR #1953, opened 2016-08-12
Update Swift design doc based on recent findings/discussion.

allevato:


allevato:
@thomasvl

allevato:
@kientzle

docs/swift/DesignDoc.md
```diff
@@ -73,6 +73,28 @@ Some remapping will still be required, though, to avoid collisions between
...
+For types nested inside packages, the message/enum name itself does not pose an
+issue, because it will always be prefixed by the package path
+(`Some_Package_Int`) or a custom prefix (`SPInt`). However, there is still the
+potential for collisions among nested messages/enums, messages/enums that are
+not in a package, or the very unfortunate case where a package-prefixed name
+might still collide with a Swift type.
+
+In these scenarios, we will append `Message` to message types and `Enum` to
+enum types if the package-prefixed name would collide with an existing Swift
+type.
```
thomasvl: Is it worth an example proto where it would happen?

docs/swift/DesignDoc.md
```diff
@@ -95,13 +117,13 @@ unique type symbol emitted in the generated binary.
...
-messages that replace the package name prefix with the provided string. For
-example, the following `.proto` file:
+level option, `swift_package_prefix`, that can be added to `.proto` files.
+When present, this will cause the package portion of the top-level type names
+to be replaced with the provided string. For example, the following `.proto`
+file:
 
 ```protobuf
-option swift_package_typealias = "FBP";
+option swift_package_prefix = "FBP";
```
thomasvl: Now that I look at this, _package_ is sorta overload between Swift and protos. Should we call this `swift_proto_package_prefix`?  The objc one is `objc_class_prefix` (and also applies to enums), but given the Swift class/struct differences, that doesn't seem like it helps too much.  `swift_prefix` might also work (if the protobuf team signs off on it).

docs/swift/DesignDoc.md
```diff
@@ -137,65 +152,117 @@ lets users treat messages polymorphically. Any shared method implementations
...
+using optionals (regular and implicitly unwrapped), but determined that they
+were unsuitable; protocol buffer users expect to receive default values, not
+`nil`, when retrieving fields, and this is more important in proto2 where fields
+can have explicit non-zero defaults. Implicitly unwrapped computed properties
+could mitigate the `clear` situation by treating assignment of `nil` as clearing
+the field, but we would still need to create a `has` method because the property
+getter would never return nil (it would always return the default value, and
+there would be no way to distinguish "not set" from "explicitly set to the
+default value"). In that sense, having the property be optional seems misleading
+to users.
```
thomasvl: Now that I think about it, in proto2 syntax, required also would be odd.  Would they be non optional?  If they default to the default, then you aren't really requiring developers to set them (which it seems like we should given then are "required"); and having "required" be optional seems equally confusing.

docs/swift/DesignDoc.md
```diff
@@ -294,19 +361,19 @@ would be represented in Swift as
...
+  public var bar: Bar {
     get { ... }
     set { ... }
   }
 }
 ```
 
-If the user explicitly sets `bar` to nil, or if it was never set when read from
-the wire, retrieving the value of `bar` would return a default, statically
+If the user explicitly clear `bar`, or if it was never set when read from
```
thomasvl: clears

docs/swift/DesignDoc.md
allevato: We could look at killing the parameterless initializer and replacing it with one that takes the required fields as arguments, but that probably makes usage a good deal more difficult, and it would prohibit the scenario where you create a "draft" message with some required fields left unset and use partial serialization to store it aside until you're ready to use full serialization to ship it off (and only at that point would you want the required check to kick in).

Just to make sure I'm clear on this, proto2 distinguishes between "a field is unset" and "a field has been explicitly set to a value that just happens to be equal to its default", right? In that case, we'd still be able to check the values at serialization time; if primitives are stored as optionals in the underlying CoW storage class, we can distinguish it that way and error out if the user tries to save an incomplete proto.

docs/swift/DesignDoc.md
allevato: Done.

docs/swift/DesignDoc.md
allevato: Updated with `swift_proto_package_prefix`, since I prefer being more explicit to less explicit, and `swift_prefix` feels a bit too general.

docs/swift/DesignDoc.md
allevato: Fixed.

docs/swift/DesignDoc.md
thomasvl: Proto2 - yes, unset is different that set to the default value.

docs/swift/DesignDoc.md
```diff
@@ -224,6 +245,16 @@ drafted or sent proposals for review that would benefit our designs:
...
 
+Lastly, we note some name transformations that we would apply to fields:
+
+* Repeated fields will have `Array` appended to their names. This is because
+  there is a trend in some messages to name repeated fields with the singular
+  version of the name (`descriptor.proto` in particular does this), and
+  `foo.name.append(bar)` looks potentially like appending a string to another
+  string whereas `foo.nameArray.append(bar)` is much clearer.
+* If a field name would collide with one of the properties or methods in
+  `ProtoMessage`'s public interface, we append `Field` to its name.
```
thomasvl: just appending Array and Field mean we could have collisions:

```
optional int32 foo = 1;
repeated int64 foo_array = 2;
optional int32 int32 = 3;
optional int32 int32_field;
```

So you might want to stick underscores before what you append.  Same goes for appending Message/Enum to those types.

kientzle: Simply including an underscore here would certainly fix the common case.  There are some corner cases that would need to be documented, though:

     optional int32 _field__1 = 7;

Leading underscores and multiple underscores within field names are handled inconsistently elsewhere in the protobuf codebase; documenting this would certainly be a service to the community.

docs/swift/DesignDoc.md
kientzle: I think this discussion is largely recreating the rationale for dropping "required" from proto3:
  * Other languages do not actually require that "required" fields be set
  * Other languages seem to treat "required" and "optional" pretty much the same (some implementations used to warn if a required field was missing, but I believe that practice has faded).

The issue here is that Swift offers a language-level notion of "optional" that seems very awkward when applied to "required" fields.  I haven't found any approach yet that I'm really happy with.

I've tried both of these approaches:
  * Treat "required" fields the same as "optional".  That's essentially what other language implementations have done, but it seems quite confusing to have a "required" field (in the proto) return an Optional<T> (in Swift).
  * In my current implementation, "required" fields are actually required in that they always have a value and are always serialized.  (However, I don't error if they're missing on deserialization; that just leaves the default behind.)  But this seems to scandalize a lot of folks, so I suppose I'll have to change this.

The other options I can see:
  * Treat them like proto3 singular:  They are not Swift optionals, always have a value, but are serialized only if they differ from the default.  This provides a nice clean Swift API but still differs from the C++ implementation, which seems to be a point of concern.
  * Treat them as Swift optionals internally, but not via the API.  They are serialized only if non-nil (internally).

This last option begs the question of whether the user needs to be able to explicitly unset them and/or explicitly test them for being set.  If so, then the most straightforward way to provide that is to expose the field as a Swift optional, which brings us back to the first idea of treating "optional" and "required" as synonyms.

Based on the discussion so far, it seems that treating "optional" as a synonym for "required" is the best match for what other language implementations do.

docs/swift/DesignDoc.md
kientzle: FWIW, I don't like adding "Array" to every repeated field.  That's a bit too meddlesome for my tastes.

allevato: (Doesn't the Objective-C implementation just append `Array` without an underscore?)

While inserting underscores would certainly disambiguate most cases, I *really* want to avoid adding underscores if I can avoid it--the resulting code would stick out like a sore among other Swift code.

While it would make the implementation more complex, I'd prefer to explore a phased approach where we detect collisions and only insert underscores where necessary. For example, if we had the following proto:

```protobuf
repeated int32 foo = 1;
optional int64 foo_array = 2;
```

We would process the singular fields first and collect their transformed names. In this case, we'd transform `foo_array` -> `fooArray`. Then, we'd process the repeated fields. We would try to transform `foo` -> `fooArray`, see the collision, and fallback `foo_Array` instead. That way, common cases still produce code that is as close to Swift conventions as possible, and only the uncommon cases suffer.

One potential problem is that a field's generated name could change by adding new fields to a proto: if you had repeated `foo` and later added singular `foo_array`, the new field would take the `fooArray` name and the previous one would change from `fooArray` to `foo_Array`. Is that something we should worry about?

With fields, we see the entire set of fields in the message so this should work--types, on the other hand, could come from multiple files and multiple `protoc` passes, so we probably have to unconditionally use underscores there.
Regarding leading/multiple (and trailing) underscores for Swift, I'd recommend retaining them--that seems like a situation where the proto author had a specific reason in mind for punctuating the field that way. Single inner underscores would be removed with the following letter uppercased.

thomasvl: Yea, when I updated things, I stuck with `Array` and not `_Array`.  Was just meaning to call out the issues.

It does detect something ending in _Array_ and adds the suffix so it hopefully won't appear like an array (since it isn't). [Code here.](https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc#L455)

Historically, my one concern with trying to wait until conflicts happen is that can mean a proto "works", people code to it, then they update the proto, and a generated property name changes, breaking their code.  When this happens, by definition, there is now a second method with a "similar" name, adding to the confusion on unbreaking the code.  So as much as possible, I'd tried to error on the behaviors not depending on the other fields in the message.

PR #1972, opened 2016-08-17
Add mechanism to allow runtime resolution of Any messages by their full type URL

dmap:
Add an extension mechanism to the JsonFormat TypeRegistry so that callers can provide their own implementation to resolve Any message descriptors using the full type URL.

This extension comes in the form of a new `TypeResolver` interface, implementations of which can be supplied to a `TypeRegistry` to resolve the message Descriptor using the full type URL if the descriptor is not found in the registry's types map.

This gives users of the JsonFormat class the chance to actually resolve Any messages that utilise custom URL prefixes and dynamic resolution mechanisms.

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -277,6 +277,21 @@ public class JsonFormat {
   }
 
   /**
+   * A TypeResolver is used by a {@link TypeRegistry} to resolve Any messages
+   * that cannot be resolved directly from its supplied list of types.
+   * This extension mechanism allows the dynamic discovery of Any message types
+   * at runtime.
+   */
+  public static interface TypeResolver {
+    /**
+     * Find a type by typeUrl. Returns null if it cannot be found.
+     * 
+     * @throws InvalidProtocolBufferException if the typeUrl is invalid
+     */
+    public Descriptor resolve(String typeUrl) throws InvalidProtocolBufferException;
+  }
+  
+  /**
    * A TypeRegistry is used to resolve Any messages in the JSON conversion.
    * You must provide a TypeRegistry containing all message types used in
    * Any message fields, or the JSON conversion will fail because data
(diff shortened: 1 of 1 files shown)


xfxyjwf:
The fully qualified name of proto types should be unique, and JsonFormat parser simply ignores the URL prefix. It seems you are trying to add support for types with same fully qualified name but different URL prefix. This is not a supported use case of Any though...

dmap:
Yes the fully qualified name has to be unique but it does not (alone) give you any indication as to how to resolve the name to a Descriptor.  

We are implementing a system that can accept and parse any protobuf message (using DynamicMessage) as long as the Descriptor is supplied as well.  The [documentation for the Any class](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/any) indicates that an HTTP GET issued to the type url should return a `google.protobuf.Type` in binary format, so we want to build an implementation that takes advantage of this mechanism to also try to resolve the Descriptor.

To do this we need access to the full type URL.

Regardless of exactly what we are trying to do, I would respectfully suggest that given that the Any messages have a full type URL attached to them it only seems sensible to pass this complete URL through to the client supplied callback that needs to resolve the message type, otherwise what is the point of having the full URL?

dmap:
By the way, my preference would have been to get rid of `TypeRegistry` completely (or turn it into a default implementation of `TypeResolver`) and just have the `TypeResolver` interface be the API for the client supplied callback to resolve Descriptors.  This would have been a clearer API.

However I implemented things in the way I did in order to keep them backwards compatible with the existing codebase.  This is perhaps why it isn't completely clear what my intention was.

xfxyjwf:
Close old issues/prs.

From what I can tell, most users will resolve types locally without needing a TypeResolver. Actually I'm not aware of any production service that needs to resolve types remotely, or any production service that  implements what's described about the type URL. Right now I have no idea what this remote resolving mechanism is good for and would rather not accept a public API change for it.

Note that I also don't think a single TypeResolver interface is adequate for your purpose. A http type service according to the spec returns a google.protobuf.Type, but there isn't any official support for converting it to proto descriptors. It may not even be possible since google.protobuf.Type has a lot less info than the descriptors.

PR #1977, opened 2016-08-17
Bump the version in prep for the 3.0.2 tag being cut

thomasvl:
Also pull the watchos support info from master.


thomasvl:
We'll need the 3.0.2 tag cut before the podspec can get pushed live.

pherl:
LG

PR #1987, opened 2016-08-20
Fix Win32 error message on Unicode build

### LINKED ISSUE #1986: Truncated error message on Unicode build on Windows
When the generator plugin has errors, the printed error message is not helpful.
This was discovered on a Windows Chromium [build] (https://build.chromium.org/p/tryserver.chromium.win/builders/win_chromium_rel_ng/builds/274054) from this [CL](https://codereview.chromium.org/2034373002/#ps220001).

The error message is:
> [4190/24683] ACTION //third_party/dom_distiller_js:dom_distiller_proto_gen(//build/toolchain/win:x86)
> FAILED: pyproto/third_party/dom_distiller_js/dom_distiller_pb2.py gen/third_party/dom_distiller_js/dom_distiller.pb.cc gen/third_party/dom_distiller_js/dom_distiller.pb.h gen/third_party/dom_distiller_js/dom_distiller_json_converter.h 
> E:/b/depot_tools/python276_bin/python.exe ../../tools/protoc_wrapper/protoc_wrapper.py --proto-in-dir ../../third_party/dom_distiller_js/dist/proto --proto-in-file dom_distiller.proto --use-system-protobuf=0 -- ./protoc --python_out pyproto/third_party/dom_distiller_js --cpp_out gen/third_party/dom_distiller_js --plugin=protoc-gen-json_converter=../../third_party/dom_distiller_js/protoc_plugins/json_values_converter.py --json_converter_out=output_dir=:gen/third_party/dom_distiller_js
> --json_converter_out: protoc-gen-json_converter: %

Note that the message is just a single character.

wychen:
This fixes #1986.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,7 @@ environment:
   matrix:
     - language: cpp
       BUILD_DLL: ON
+      UNICODE: ON
 
     - language: csharp
 
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -32,6 +32,7 @@ option(protobuf_WITH_ZLIB "Build with zlib support" ${protobuf_WITH_ZLIB_DEFAULT
 set(protobuf_DEBUG_POSTFIX "d"
   CACHE STRING "Default debug postfix")
 mark_as_advanced(protobuf_DEBUG_POSTFIX)
+option(UNICODE "Build with Unicode support" OFF)
 # User options
 include(protobuf-options.cmake)
 
@@ -158,6 +159,10 @@ else (MSVC)
   # the build tool.
   set(LIB_PREFIX)
 endif (MSVC)
+
+if (UNICODE)
+  add_definitions(-DUNICODE -D_UNICODE)
+endif (UNICODE)
 
 include(libprotobuf-lite.cmake)
 include(libprotobuf.cmake)
--- a/src/google/protobuf/compiler/command_line_interface_unittest.cc
+++ b/src/google/protobuf/compiler/command_line_interface_unittest.cc
@@ -719,6 +719,11 @@ TEST_F(CommandLineInterfaceTest, TrailingBackslash) {
   ExpectGenerated("test_generator", "", "foo.proto", "Foo");
 }
 
+TEST_F(CommandLineInterfaceTest, Win32ErrorMessage) {
+  EXPECT_EQ("The system cannot find the file specified.\r\n",
+    Subprocess::Win32ErrorMessage(ERROR_FILE_NOT_FOUND));
+}
+
 #endif  // defined(_WIN32) || defined(__CYGWIN__)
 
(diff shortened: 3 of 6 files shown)


src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -719,6 +719,11 @@ TEST_F(CommandLineInterfaceTest, TrailingBackslash) {
   ExpectGenerated("test_generator", "", "foo.proto", "Foo");
 }
 
+TEST_F(CommandLineInterfaceTest, Win32ErrorMessage) {
+  EXPECT_EQ("The system cannot find the file specified.\r\n",
```
wychen: Locales other than English might have problems, but this is fine on CI.

appveyor.yml
```diff
@@ -11,6 +11,7 @@ environment:
   matrix:
     - language: cpp
       BUILD_DLL: ON
+      UNICODE: ON
```
wychen: One config takes ~7 minutes. It seems we can afford more configurations. Travis-CI takes much longer (>30 minutes) anyway.
Is Unicode much more common than Ansi mode? If not, we might want to keep coverage of Ansi mode as well.
One data point: Chromium uses Unicode mode.

cmake/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(protobuf_WITH_ZLIB "Build with zlib support" ${protobuf_WITH_ZLIB_DEFAULT
 set(protobuf_DEBUG_POSTFIX "d"
   CACHE STRING "Default debug postfix")
 mark_as_advanced(protobuf_DEBUG_POSTFIX)
+option(UNICODE "Build with Unicode support" OFF)
```
xfxyjwf: Adding this option doesn't seem necessary. Can we avoid it?

src/google/protobuf/compiler/subprocess.cc
```diff
@@ -258,7 +258,7 @@ bool Subprocess::Communicate(const Message& input, Message* output,
 }
 
 string Subprocess::Win32ErrorMessage(DWORD error_code) {
-  char* message;
+  TCHAR* message = nullptr;
```
xfxyjwf: Instead of using TCHAR (and wstring), can you change the code to use the ASCII version of all win API functions? For example, instead of FormatMessage, use FormatMessageA.

cmake/CMakeLists.txt
wychen: Removed.

src/google/protobuf/compiler/subprocess.cc
wychen: Fixed this and file.cc.

wychen:
To make history cleaner, this is continued in another pull request:
https://github.com/google/protobuf/pull/2044

PR #1992, opened 2016-08-22
Add note about JSON tests maybe being wrong.

thomasvl:
- Add note about the JSON test maybe not being correct yet.
- Add test to checks the generated names for double underscores to be sure they
  are what is expected.


xfxyjwf:
LGTM

PR #1997, opened 2016-08-23
Move ImportWriter into helpers for sharing

thomasvl:


jcanizales:
Hey Thomas, I think it might be better to keep passing an Options object to the constructor. That way:
- If you add a new parameter in the future you can do it easily without breaking ImportWriter's interface.
- If we can also reuse your flag-parsing logic we're guaranteed to be in lockstep.

thomasvl:
Options parsing is pretty generic, so reuse there isn't a big deal.  Re-using would means you could _accept_ an option that has no meaning to the grpc generation (instead of kicking out an error about an unknown option).

As far as breaking the interface - yes and no. That only covers the ctor if everything is still an option, if we needed a flag about doing imports for generated headers vs. sources, then it wouldn't be in the options and things would break.   Likewise, if we had to add something to `AddFile` that would also break.

In moving this, the one things I'm already a little worried about is if the protos went in one framework and grpc into another. Say because the protos were needed by one library, and then some other library uses grpc and those same protos.  It isn't clear how that could be expressed to deal with it in the current system.

jcanizales:
Re the last part, one can always put the proto service in a separate file and import the rest.

thomasvl:
fyi - landing as is.  We can bring back the options ctor arg if it really makes sense, but for now it seems safer to error on this side of things.

jcanizales:
@TimBurks @makdharma This is the class that can be used to generate correct proto imports, so that the Speech API Swift sample can compile.

jcanizales:
I.e. to fix https://github.com/grpc/grpc/issues/7563

Issue #2002, opened 2016-08-24
3.0.0: FTBFS on [s390x,alpha]: undefined reference to NoBarrier_(AtomicIncrement|Store|AtomicExchange|Load)


onlyjob:
```
/bin/bash ../libtool  --tag=CXX   --mode=link g++ -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare  -g -O2 -fdebug-prefix-map=/«PKGBUILDDIR»=. -fstack-protector-strong -Wformat -Werror=format-security -pthread -Wl,-z,relro -o protoc google/protobuf/compiler/main.o -lpthread libprotobuf.la libprotoc.la -lz 
libtool: link: g++ -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare -g -O2 -fdebug-prefix-map=/«PKGBUILDDIR»=. -fstack-protector-strong -Wformat -Werror=format-security -pthread -Wl,-z -Wl,relro -o .libs/protoc google/protobuf/compiler/main.o  -lpthread ./.libs/libprotobuf.so ./.libs/libprotoc.so -lz -pthread
./.libs/libprotobuf.so: undefined reference to `google::protobuf::internal::NoBarrier_AtomicIncrement(long volatile*, long)'
./.libs/libprotobuf.so: undefined reference to `google::protobuf::internal::NoBarrier_Store(long volatile*, long)'
./.libs/libprotobuf.so: undefined reference to `google::protobuf::internal::NoBarrier_AtomicExchange(long volatile*, long)'
./.libs/libprotobuf.so: undefined reference to `google::protobuf::internal::NoBarrier_Load(long const volatile*)'
collect2: error: ld returned 1 exit status
```

See full build log: https://buildd.debian.org/status/fetch.php?pkg=protobuf&arch=s390x&ver=3.0.0-3&stamp=1472017457

onlyjob:
Affects _alpha_ as well: https://buildd.debian.org/status/fetch.php?pkg=protobuf&arch=alpha&ver=3.0.0-3&stamp=1472018795

sebastic:
The [patch for s390x](https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=835302;filename=s390x.patch;msg=5) is available, based on the [port of Google V8 JavaScript engine to z Systems](https://github.com/ibmruntimes/v8z):
```patch
[...]

PR #2006, opened 2016-08-24
Add support for s390x architecture.

sebastic:
Based on port of Google V8 JavaScript engine to z Systems:

https://github.com/ibmruntimes/v8z

See: #2002


sebastic:

Issue #2003, opened 2016-08-24
3.0.0: FTBFS on [kfreebsd-amd64,kfreebsd-i386]: "'EXPECT_DEATH' was not declared in this scope"


onlyjob:
```
google/protobuf/stubs/stringpiece_unittest.cc: In member function 'virtual void google::protobuf::{anonymous}::NonNegativeLenTest_NonNegativeLen_Test::TestBody()':
google/protobuf/stubs/stringpiece_unittest.cc:788:50: error: 'EXPECT_DEATH' was not declared in this scope
   EXPECT_DEATH(StringPiece("xyz", -1), "len >= 0");
                                                  ^
Makefile:5553: recipe for target 'google/protobuf/stubs/protobuf_test-stringpiece_unittest.o' failed
```

Build logs:
- https://buildd.debian.org/status/fetch.php?pkg=protobuf&arch=kfreebsd-amd64&ver=3.0.0-3&stamp=1472017109
- https://buildd.debian.org/status/fetch.php?pkg=protobuf&arch=kfreebsd-i386&ver=3.0.0-3&stamp=1472017425

sebastic:
I've added the following patch to fix this issue:
```patch
Description: Disable test if EXPECT_DEATH is not available.
 Caused FTBFS on kfreebsd-*.
Author: Bas Couwenberg <sebastic@debian.org>

--- a/src/google/protobuf/stubs/stringpiece_unittest.cc
+++ b/src/google/protobuf/stubs/stringpiece_unittest.cc
@@ -784,9 +784,11 @@ TEST(FindOneCharTest, EdgeCases) {
 }
 
 #ifndef NDEBUG
+#ifdef PROTOBUF_HAS_DEATH_TEST
 TEST(NonNegativeLenTest, NonNegativeLen) {
   EXPECT_DEATH(StringPiece("xyz", -1), "len >= 0");
 }
+#endif  // PROTOBUF_HAS_DEATH_TEST
 #endif  // ndef DEBUG
 
 }  // namespace
```

Unfortunately the testsuite then fails with:
```
[ FATAL ] /usr/include/gtest/internal/gtest-port.h:1343:: pthread_mutex_lock(&mutex_)failed with error 22
FAIL protobuf-test (exit status: 134)
```

onlyjob:
Affects _hurd_ as well: https://buildd.debian.org/status/fetch.php?pkg=protobuf&arch=hurd-i386&ver=3.0.0-4&stamp=1472091254

[...]

PR #2008, opened 2016-08-24
Disable test if EXPECT_DEATH is not available.

sebastic:
All other tests using `EXPECT_DEATH()` are guarded by the same `#ifdef`.

See also: #2003


sebastic:

Issue #1999, opened 2016-08-24
hard-coded "python2.4"


onlyjob:
As reported in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=600261
the following files have `python2.4` hard-coded in shebang:
- [python/mox.py](../blob/master/python/mox.py#L1)
- [python/stubout.py](../blob/master/python/stubout.py#L1)

PR #2009, opened 2016-08-24
Use /usr/bin/python instead of /usr/bin/python2.4

sebastic:
Using `/usr/bin/env python` like `python/setup.py` is an option too.

Fixes: #1999


sebastic:

PR #2011, opened 2016-08-25
Add an Erlang project, gpb, to third_party.md

tomas-abrahamsson:
I have yet another Erlang binding to the protocol buffers.

I noticed the name, gpb, clashes with that used in the Objective C binding. I'm sorry about that. I was not aware of that when I started work on my Erlang project long ago, but since it is different languages, I don't think there's any confusion.


tomas-abrahamsson:
I signed it

PR #2021, opened 2016-08-27
benchmarks: update readme.txt

zlim:
Remove references to google_speed.proto which was removed in
commit 2e83110230b7.

Also fix-up commands to build and run.


xfxyjwf:
Thanks for the fix!

PR #2024, opened 2016-08-29
Restore New*Callback into google::protobuf namespace …

### LINKED ISSUE #1966: NewCallback() functions should not be in ::internal namespace
Commit 7c14dc837b2fcd61ed244ced90debeb4b42d1119 moved some externally used functions (such as "Closure* NewCallback(..)"  to the google::protobuf::internal namespace (from the google::protobuf namespace). These NewCallback functions are needed in the stub code generated by the proto compiler, and it seems strange that they are in the "internal" namespace.

pstavirs:
… since these are used by the service stubs code

Fixes #1966


xfxyjwf:
ok to test

xfxyjwf:
Will merge after the test finishes.

pstavirs:
Any ETA for the merge?

The [Ostinato](http://ostinato.org/) package is at risk of being removed from Debian, if this isn't fixed - see pstavirs/ostinato#193

PR #2031, opened 2016-08-30
Speed up ObjC Generation with large dependency trees

thomasvl:
Don't create FileGenerators for each dep. FileGenerators will deeply create all
the message, enum, and field generators; but those aren't needed when doing
the imports for dependencies. Instead directly generate the imports off the
FileDescriptors so no extra objects are created. The only other use was when
chaining together the *Roots for the file extension registry, but that also
can be generate off the name of the FileDescriptor directly.


sergiocampama:
LG

PR #2033, opened 2016-08-30
update the OSGi metadata for the javanano package

frett:
the javanano OSGi metadata was referencing the wrong package for exporting classes in an OSGi environment


xfxyjwf:
Thanks for the patch!

PR #2035, opened 2016-08-30
Fixes static analyzer issues from xcode.

sergiocampama:


sergiocampama:
Fixes https://github.com/google/protobuf/issues/2034

sergiocampama:
@pherl  FYI

sergiocampama:
@xfxyjwf

xfxyjwf:
I don't think these are the right changes to make. For example,
```cpp
GOOGLE_CHECK(pointer != NULL);
if (pointer != NULL) {
  ..
}
```
it seems silly to add useless checks to please a code analyzer.

sergiocampama:
yeah, that case is what I'm trying to find a solution for...  I could replace it with assert... 

The fact that GOOGLE_CHECK is based off of the Logger is the weird part, otherwise I could add __attribute__((noreturn)) and have the analyzer understand GOOGLE_CHECK...

xfxyjwf:
If xcode can understand some annotations, adding annotations is a more acceptable solution. Otherwise people seeing the code in protobuf internal code base may attempt to remove the additional checks you are adding here. That will be a waste of engineering resources.

sergiocampama:
As I was telling @pherl I don't think adding the annotation to GOOGLE_CHECK is feasible, as it is done inside the logger... so we could add the noreturn attribute there, but it would be cheating, as it would return in the majority of the use cases... Another option would be to refactor GOOGLE_CHECK into something different to use an assert, for example

sergiocampama:
Ok, removed the method one for now, what about the other ones?

sergiocampama:
ping @xfxyjwf

src/google/protobuf/util/message_differencer.cc
```diff
@@ -875,6 +877,7 @@ bool MessageDifferencer::CompareRepeatedField(
 
   for (int i = 0; i < count1; ++i) {
     if (match_list1[i] != -1) continue;
+    if (reporter_ == NULL) continue;
```
xfxyjwf: Make this an assert as well?

sergiocampama: Done

xfxyjwf:
Thanks. Can you squash the commits into one? I'll merge after that.

sergiocampama:
done... thanks! I'll try to find another way for the missing one

sergiocampama:
the appveyor error seems to be a timeout

xfxyjwf:
It seems appveyor is timing out on downloading dependencies very often recently...

PR #2037, opened 2016-08-30
Fix #2036 (Ruby `hash` broken for Messages with repeated fields.)

### LINKED ISSUE #2036: Ruby `hash` broken for Messages with repeated fields.
## Expected

All Ruby Message objects return a `hash`.

## Actual

Calling `hash` on a Message with a repeated field will occasionally raise the following error:

```ruby
RangeError: bignum too big to convert into `long'
```

### Test Case

I have written a [gist with tests](https://gist.github.com/abscondment/d8254d4482c024ad06a7e683c690ff98) demonstrating the behavior.

### Details

This is because of a bug in [`repeated_field.c`](https://github.com/google/protobuf/blob/master/ruby/ext/google/protobuf_c/repeated_field.c#L448), wherein the hash function occasionally returns a `Bignum` instead of a `Fixnum`. `RepeatedField_hash` can return a Bignum because of the algorithm it uses for combining hashes: bitwise shift left, then xor. If the original hash is large enough, the `<<` will cause Ruby to automagically promote the value from a `Fixnum` to a `Bignum`.

Ruby indicates that [`hash` should return a Fixnum](https://ruby-doc.org/core-2.3.1/Object.html#method-i-hash), so this behavior is not correct.

Up the stack, `storage.c` [attempts to convert this Bignum](https://github.com/google/protobuf/blob/master/ruby/ext/google/protobuf_c/storage.c#L863) to a long, which produces the `RangeError`.

## Proposal

I *think* the correct behavior is to use `rb_hash_start` -> `rb_hash_uint` -> `rb_hash_end` -> `INT2FIX` as `storage.c` does. ~~I'd be happy to provide a patch for this if that sounds correct.~~ I've provided a PR.

abscondment:
#2036 spells out how the `hash` function is broken in many Ruby messages. This PR fixes that.

### DIFF
--- a/ruby/ext/google/protobuf_c/repeated_field.c
+++ b/ruby/ext/google/protobuf_c/repeated_field.c
@@ -447,9 +447,8 @@ VALUE RepeatedField_eq(VALUE _self, VALUE _other) {
  */
 VALUE RepeatedField_hash(VALUE _self) {
   RepeatedField* self = ruby_to_RepeatedField(_self);
-
-  VALUE hash = LL2NUM(0);
-
+  st_index_t h = rb_hash_start(0);
+  VALUE hash_sym = rb_intern("hash");
   upb_fieldtype_t field_type = self->field_type;
   VALUE field_type_class = self->field_type_class;
   size_t elem_size = native_slot_size(field_type);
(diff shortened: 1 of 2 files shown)


abscondment:
@haberman Are you the right person to review this? thx!

haberman:
Ack sorry for the delay on this!

haberman:
Oh, looks like this caused a breakage in JRuby: https://travis-ci.org/google/protobuf/jobs/160942507

abscondment:
@haberman Interesting. I think this should actually be considered a second bug in JRuby that this new test exposed. I.e. the hashes of two messages with identical values should be identical, but in this case it is not.

If you agree with that assessment, I'd be happy to open another issue and look into the fix.

haberman:
@abscondment that sounds like a reasonable assessment, thanks for taking a look!

PR #2044, opened 2016-09-01
Fix Win32 error messages on Unicode build

### LINKED ISSUE #1986: Truncated error message on Unicode build on Windows

wychen:
This fixes #1986.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -159,9 +159,9 @@ else (MSVC)
   set(LIB_PREFIX)
 endif (MSVC)
 
-if (UNICODE)
+if (protobuf_UNICODE)
   add_definitions(-DUNICODE -D_UNICODE)
-endif (UNICODE)
+endif (protobuf_UNICODE)
 
 include(libprotobuf-lite.cmake)
 include(libprotobuf.cmake)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -159,6 +159,10 @@ else (MSVC)
   set(LIB_PREFIX)
 endif (MSVC)
 
+if (UNICODE)
+  add_definitions(-DUNICODE -D_UNICODE)
+endif (UNICODE)
+
 include(libprotobuf-lite.cmake)
 include(libprotobuf.cmake)
 include(libprotoc.cmake)
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -10,7 +10,7 @@ goto :error
 echo Building C++
 mkdir build_msvc
 cd build_msvc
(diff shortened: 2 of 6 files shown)


cmake/CMakeLists.txt
```diff
@@ -159,6 +159,10 @@ else (MSVC)
   set(LIB_PREFIX)
 endif (MSVC)
 
+if (UNICODE)
+  add_definitions(-DUNICODE -D_UNICODE)
```
xfxyjwf: Drop this as it's no longer used?

wychen: It's still used. Without these lines, these macro won't be passed to the compiler, and defining UNICODE=ON has no effect.

xfxyjwf: Why do we need to pass UNICODE to the compiler (msvc in this case)? If the user want to enable unicode build, can they just enable it inside visual studio (the configuration page)?

wychen: It's for the test coverage on continuous integration. CommandLineInterfaceTest.Win32ErrorMessage can still pass on ANSI without the fix. In appveyor.yml, UNICODE=ON is added.

xfxyjwf: I see. Can you name the option the same way as other options? I.e., protobuf_UNICODE

wychen: Done.

wychen:
Does it look good now?

xfxyjwf:
LG. Thanks!

PR #2045, opened 2016-09-01
fix #1342 cause by ownership issues

### LINKED ISSUE #1342: C++: JSON serialization of bytes not (yet) working?
I have the use case, where I would like to serialize a message to a JSON string that contains also a bytes field. The documentation (language guide) states that the `bytes` field shall be encoded as a `base64` encoded string.
However, on Windows, depending on the content of the bytes array, the serialized JSON contains an empty string for the bytes `""`.

I have made a  minimal example to reproduce the issue:
First the proto file: (message.proto)
```
syntax = "proto2";

package pb;

message Person
{
  optional string name   = 1;
  optional bytes  binary = 2;
}

```
 and then the code (main.cpp):

```
#include <iostream>
#include <memory>

// protobuf includes
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable: 4100 4127 4244 4267 4512)
#endif
#include "messages.pb.h"
#include <google/protobuf/util/json_util.h>
#include <google/protobuf/util/type_resolver_util.h>
#ifdef _MSC_VER
#pragma warning(pop)
#endif

int main(int argc, char **argv)
{
  class pb::Person person;
  std::string binary_string;
  for (char i = 0; i < 10; i++)
  {
    binary_string.push_back(i);
  }
  person.set_name("Maria");
  person.set_binary(binary_string.c_str(), binary_string.length());
  std::string serialized = person.SerializeAsString();
  std::string json_string;

  google::protobuf::util::JsonOptions options;
  options.always_print_primitive_fields = true;
  std::shared_ptr<google::protobuf::util::TypeResolver> resolver_;
  resolver_.reset(google::protobuf::util::NewTypeResolverForDescriptorPool(
    "", google::protobuf::DescriptorPool::generated_pool()));


  auto status = google::protobuf::util::BinaryToJsonString(resolver_.get(), "/pb.Person", serialized, &json_string, options);
  std::cout << json_string << std::endl;

  std::string res;
  std::cin >> res;
  return(0);
}
```

This prints `{"name":"Maria","binary":""}` while I expect it to print `{"name":"Maria","binary":"AAECAwQFBgcICQ=="}`.

I have tried with beta2 and the current master (as of today), using either `syntax = "proto2";` and `syntax = "proto3";`  Else the JSON serialization works fine, except for the `bytes` field.

Can you please give me some hints if this is a bug, or not yet supported.

mike07026:
#1342 it caused by "StringPiece value" ownership issues,  we can fix it like RenderString implement.

### DIFF
--- a/src/google/protobuf/util/internal/default_value_objectwriter.cc
+++ b/src/google/protobuf/util/internal/default_value_objectwriter.cc
@@ -165,7 +165,10 @@ DefaultValueObjectWriter* DefaultValueObjectWriter::RenderBytes(
   if (current_ == NULL) {
     ow_->RenderBytes(name, value);
   } else {
-    RenderDataPiece(name, DataPiece(value, false, true));
+    // Since StringPiece is essentially a pointer, takes a copy of "value" to
+    // avoid ownership issues.
+    string_values_.push_back(new string(value.ToString()));
+    RenderDataPiece(name, DataPiece(*string_values_.back(), false, true));
   }
   return this;
 }
--- a/src/google/protobuf/util/json_util_test.cc
+++ b/src/google/protobuf/util/json_util_test.cc
@@ -133,29 +133,29 @@ TEST_F(JsonUtilTest, TestDefaultValues) {
   m.set_string_value("i am a test string value");
   m.set_bytes_value("i am a test bytes value");
   EXPECT_EQ(
-	  "{\"boolValue\":false,"
-	  "\"int32Value\":0,"
-	  "\"int64Value\":\"0\","
-	  "\"uint32Value\":0,"
-	  "\"uint64Value\":\"0\","
-	  "\"floatValue\":0,"
-	  "\"doubleValue\":0,"
-	  "\"stringValue\":\"i am a test string value\","
-	  "\"bytesValue\":\"aSBhbSBhIHRlc3QgYnl0ZXMgdmFsdWU=\","
-	  "\"enumValue\":\"FOO\","
-	  "\"repeatedBoolValue\":[],"
-	  "\"repeatedInt32Value\":[],"
-	  "\"repeatedInt64Value\":[],"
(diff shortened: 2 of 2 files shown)


mike07026:
@googlebot I signed it!

src/google/protobuf/util/internal/default_value_objectwriter.cc
```diff
@@ -165,7 +165,10 @@ DefaultValueObjectWriter* DefaultValueObjectWriter::RenderBytes(
   if (current_ == NULL) {
     ow_->RenderBytes(name, value);
   } else {
-    RenderDataPiece(name, DataPiece(value, false, true));
+    // Since StringPiece is essentially a pointer, takes a copy of "value" to
+    // avoid ownership issues.
+    string_values_.push_back(new string(value.ToString()));
+    RenderDataPiece(name, DataPiece(*string_values_.back(), false, true));
```
xfxyjwf: Does RenderDataPiece cache the StringPiece value somewhere? I thought RenderDataPiece will write directly to output and won't have ownership issues even when the StringPiece is a temporary.

mike07026: @xfxyjwf 

In default_value_objectwriter.cc , there is 2 step to generate json.
1. collect all necessary information into DefaultValueObjectWriter::root_
2. write json in one breath with infomation in DefaultValueObjectWriter::root_

In the first step, RenderDataPiece will be called, it just cache the StringPiece value into "root_" member(cause ownership issus here), and will not write direct to output.
In the second step, the actual write happens, but bytes field information in "root_" is "dead" already. So json will not be correct in bytes field .

Please refer to L150-L160 in default_value_objectwriter.cc, RenderString implement.

xfxyjwf: Thanks for the explanation!

PR #2051, opened 2016-09-01
Rm check on dependency in the C# reflection API

gturri:
This check adds a few constraints on the way to build a project when we have
a proto file which imports another one. In particular, on projects which
build both C# and Java, it's easy to end up with exceptions like

    Expected: included.proto but was src/main/protobuf/included.proto

A user may work around this issue, but it may add unnecessary constraints
on the layout of the project.

According to https://github.com/google/protobuf/commit/f3504cf3b1d456a843e8242fdee9ba0bf2991dc1#diff-ecb0b909ed572381a1c8d1994f09a948R309
it has already been considered to get rid of this check, for
similar considerations, and because it doesn't exist in the Java code

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
@@ -251,17 +251,6 @@ namespace Google.Protobuf.Reflection
                     "Dependencies passed to FileDescriptor.BuildFrom() don't match " +
                     "those listed in the FileDescriptorProto.");
             }
-            for (int i = 0; i < proto.Dependency.Count; i++)
-            {
-                if (dependencies[i].Name != proto.Dependency[i])
-                {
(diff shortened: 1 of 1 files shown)


gturri:
I've just joined https://groups.google.com/forum/#!forum/criteo-contributors which is the group for which my employer has signed a CLA. AFAIU, it should be good then.

jskeet:
I can see checking the exact match not being necessary, but the count part still seems worth keeping - I suspect code elsewhere is going to assume there *are* enough dependency descriptors, and the error if there aren't is going to be pretty obscure. Feel free to push back though...

gturri:
I've indeed been a bit too greedy.
Here, I just updated this PR.

jskeet:
Okay - this looks like it should be okay. Will merge when Travis goes green (for C# at least; this won't affect other languages).

PR #2052, opened 2016-09-01
Initial check in of PHP implementation

TeBoring:
This pull request includes two implementation: C extension and PHP
package. Both implementations support encode/decode of singular,
repeated and map fields. Oneof support will be added in future pull
requests.


TeBoring:
@xfxyjwf

xfxyjwf:
ok to test

composer.json
```diff
@@ -0,0 +1,24 @@
+{
```
xfxyjwf: What is this file and does it have to be in the top level?

xfxyjwf:
Can you update the link in the top level README.md file for php?

composer.json
TeBoring: It is needed for using pure PHP runtime. Basically it describes the PHP runtime package like Protobuf.spec for objc. It has to be on the top level to be correctly pulled by the central PHP repository.

pmbethe09:
This PR broke the bazel BUILD file's ability to run protoc 
-- still compiles, but then bazel-bin/protoc fails with:
dyld: Symbol not found: __ZTVN6google8protobuf8compiler3php9GeneratorE
  Referenced from: bin/external/com_google_protobuf/protoc
  Expected in: flat namespace

> git bisect start HEAD HEAD~100
> git bisect run bash -c "bazel build :protoc && bazel-bin/protoc --cpp_out=/tmp src/google/protobuf/any.proto || exit 1"

e0e54661f76183684dca66694967a60cbb10f04e is the first bad commit

Fix: just need to add 
src/google/protobuf/compiler/php/php_generator.cc
to the "protoc_lib" target

PR #2058, opened 2016-09-02
Add go_package to field_mask.proto.

### LINKED ISSUE #2057: file_mask.proto is missing a go_package
https://github.com/google/protobuf/blob/master/src/google/protobuf/field_mask.proto

This file probably needs

```
option go_package = "github.com/golang/protobuf/ptypes/fieldmask";
```

I have filed https://github.com/golang/protobuf/issues/226 to track including it in the Go protobuf package.

tswast:
Fixes #2057.

### DIFF
--- a/src/google/protobuf/field_mask.proto
+++ b/src/google/protobuf/field_mask.proto
@@ -33,6 +33,7 @@ syntax = "proto3";
 package google.protobuf;
 
 option csharp_namespace = "Google.Protobuf.WellKnownTypes";
+option go_package = "google.golang.org/genproto/protobuf";
 option java_package = "com.google.protobuf";
 option java_outer_classname = "FieldMaskProto";
 option java_multiple_files = true;


xfxyjwf:
Can you ask someone from golang/protobuf to LGTM it? I can help merge the change after that.

tswast:
@zombiezen Could you take a look? I have updated this based on your comment here: https://github.com/golang/protobuf/pull/227

@broady I think we should regenerate https://github.com/google/go-genproto Would this change affect the current script used?

tswast:
Ping. Sorry I lost track of this.

broady:
LGTM, but I think the canonical location for these proto files is in the internal Google repo.

Ross? @zombiezen

zombiezen:
@pongad, can you take a look at this?

pongad:
I have already made a similar change in the internal Google repo.
The difference is that the option is
`option go_package = "google.golang.org/genproto/protobuf/field_mask;field_mask";`
so that the package structure is similar to the one we have inside Google.

Currently, I'm waiting for another CL for `descriptor.proto` to land. Once that lands, I'll work on copying those change en masse to this repo.
go-genproto's regen.sh might have to be updated to read go_package option to get the new package structure though.

pongad:
This change has already landed. Should we close this?

tswast:
Thanks @pongad

tswast:
@broady Can you update genproto?

PR #2062, opened 2016-09-02
Bump version number to 3.0.2

pherl:


pherl:
@jskeet @haberman

xfxyjwf:
LGTM

jskeet:
LGTM, but it's not clear that we actually need a new C# build at all - the C# code hasn't changed, has it? We need a new Google.Protobuf.Tools build to catch the protoc changes, but I don't think we need a new Google.Protobuf package.

xfxyjwf:
It's probably better to update all languages (including those that haven't actually changed), because otherwise we will need to tell users that protoc 3.0.2 can work with C# runtime 3.0.0 but not Objective-C runtime 3.0.0, and it will very confusing with many different languages...

Issue #2049, opened 2016-09-01
c++ std::min problem when use proto map


molixiaoge:
When i use proto map type with c++ source code,the visual studio out put  show:
error C2589: “(”:“::”

So i check the code ,and use (std::min) instead of std::min,because windows.h have min macro conflict with std::min.

The filelist i found in 3.0ga show as the follow:

google/protobuf/map.h:1081
google/protobuf/wire_format_lite_inl.h:278
google/protobuf/wire_format_lite_inl.h:370

Look forward to your reply.

xfxyjwf:
Can you point to which line of the code has this problem?

asquithea:
Defining `NOMINMAX` may help avoid this one.

molixiaoge:
because windef.h stdlib.h contains min or max macro define,so it is conflict

dschepler:
In a quick search I also found std::max usage in google/protobuf/repeated_field.h lines 1262 and 1263 that might need the same (std::max)(...) treatment.

david0u0:
Can we consider reopen this issue? I don't think this problem is perfectly solved if we had to modify the codegened files. It's against protobuf's workflow in my humble opinion.

Issue #2022, opened 2016-08-28
descriptor.h dependency in "using namespace std"


shooshx:
src/google/protobuf/descriptor.h line 146 has:

```
  vector<string> leading_detached_comments;
```

unlike everywhere else in the code this line does not have the std:: qualifier 
this line compiles only due to 
include/google/protobuf/stubs/common.h line 220 which says

```
 using namespace std;  // Don't do this at home, kids.
```

I ran into this issue when I had this included from a file where another namespace (Ogre) added std into itself and also had a global "using namespace Ogre" . This caused an ambiguity of which vector to use google::protobuf::std::vector or Ogre::std::vector

khingblue:
Can we remove `using namespace std;` and add `std::` to code?

xfxyjwf:
Is adding std:: to vector<string> good enough to resolve the issue? If so feel free to send us a pull request to do the fix.

Replacing "using namespace std;" will be a more involving change and will likely require changes to our internal code base as well, so we'd better make that change in our internal code base instead.

shooshx:
After browsing some more of the headers here, I am no longer convinced this is the right fix. I've fixed my own code to work around the issue I had. Please close this issue, sorry for the bother.

PR #2065, opened 2016-09-03
Use global namespace to avoid ambiguity.

khingblue:
This aligns with other codes in ProtoBuf, for example:

```
::std::pair<bool, const uint8*>
const ::std::string& Any::type_url()
```

### DIFF
--- a/src/google/protobuf/lite_unittest.cc
+++ b/src/google/protobuf/lite_unittest.cc
@@ -46,7 +46,7 @@
 #include <google/protobuf/wire_format_lite_inl.h>
 #include <google/protobuf/stubs/strutil.h>
 
-using namespace std;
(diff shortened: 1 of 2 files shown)


khingblue:
This PR is to avoid ambiguity when using namespace std, while not introducing too many changes in internal code. One example of the ambiguity can be found in issue #2049.

@xfxyjwf PTAL, thanks.

xfxyjwf:
I don't see how this helps with https://github.com/google/protobuf/issues/2049. From my understanding, this pull request doesn't change anything at all (except for the case where someone has defined another std namespace somewhere). Can you elaborate a bit more?

khingblue:
@xfxyjwf Sorry the issue should be #2022.

PR #2079, opened 2016-09-07
Fix #2071 replacing /bin/sh with bash when generates descriptor

### LINKED ISSUE #2071: generate_descriptor_proto.sh: Syntax error: "(" unexpected
When using script to regenerate descriptors on Ubuntu, it fails with following error:
```
$ ./generate_descriptor_proto.sh
./generate_descriptor_proto.sh: 31: ./generate_descriptor_proto.sh: Syntax error: "(" unexpected
```

Because on Ubuntu, by default /bin/sh is dash, so if user executes this script directly, it will fail with this error.

There are 2 ways to fix it:

1. Change the first line of `generate_descriptor_proto.sh` to `#!/usr/bin/env bash`
1. Prompt user to use: `bash generate_descriptor_proto.sh`

If this is an issue that you'd like to get fixed, I'm happy to submit a PR with one of the solutions above which is preferable. Thanks.

khingblue:
On Ubuntu `/bin/sh` is `dash` by default, this can cause `generate_descriptor_proto.sh` fail to execute.

### DIFF
--- a/generate_descriptor_proto.sh
+++ b/generate_descriptor_proto.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # Run this script to regenerate descriptor.pb.{h,cc} after the protocol
 # compiler changes.  Since these files are compiled into the protocol compiler


xfxyjwf:
Thanks for the patch. The failing tests in travis seem unrelated.

PR #2084, opened 2016-09-07
Add conformance tests for Go.

tswast:
Also updates the tests for Go to use https://github.com/travis-ci/gimme
instead of `apt-get` to install Go. This will make it easier to test
against specific versions of Go.

This PR depends on https://github.com/google/protobuf/pull/2058

@broady @rakyll Could you take a look for the Go changes?


conformance/conformance_go.go
```diff
@@ -0,0 +1,159 @@
...
+// If EOF is encountered in an unexpected place, returns an error.
+func doTestIO() (bool, error) {
+	r := bufio.NewReader(os.Stdin)
+
+	// Get the length of the next message, which is encoded in the first 4 bytes
+	// as an unsigned integer. If there aren't 4 bytes here, the test is
+	// finished.
+	b, err := r.Peek(4)
+	if len(b) == 0 && err == io.EOF {
+		return false, nil
```
rakyll: Why don't you return an EOF here. Then you don't need the additional bool return. It is pretty unexpected to look another value in Go to determine success once err = nil.

conformance/conformance_go.go
```diff
@@ -0,0 +1,159 @@
...
+	case pb.WireFormat_UNSPECIFIED:
+		return resp, errors.New("unspecified output format")
+	case pb.WireFormat_PROTOBUF:
+		b, err := proto.Marshal(test)
+		if err != nil {
+			return resp, err
+		}
+		resp.Result = &pb.ConformanceResponse_ProtobufPayload{ProtobufPayload: b}
+	case pb.WireFormat_JSON:
+		marshal := &jsonpb.Marshaler{}
```
rakyll: I'd s/marshal/marshaler or s/marshal/m. marshal sounds like a func.

conformance/conformance_go.go
```diff
@@ -0,0 +1,159 @@
...
+	var n uint32
+	if err := binary.Read(r, binary.LittleEndian, &n); err != nil {
+		return false, err
+	}
+
+	// Read the next message, now that the size is in the variable n.
+	b = make([]byte, n, n)
+	nb, err := r.Read(b)
+	if nb != int(n) {
+		return false, fmt.Errorf("got %d bytes; expected %d", nb, n)
```
rakyll: got %d bytes; want %d

You can use %v rather than %d.

conformance/conformance_go.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	req := &pb.ConformanceRequest{}
+	if err := proto.Unmarshal(b, req); err != nil {
+		return false, err
```
rakyll: It would be nice to add some context to this error since there are multiple marshaling/unmarshaling happening.
e.g.
return false, fmt.Errorf("failed to unmarshal conformance req: %v", err)

conformance/conformance_go.go
```diff
@@ -0,0 +1,159 @@
...
+	if _, err := os.Stdout.Write(b); err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func main() {
+	var n int
+	for {
```
rakyll: for n := 0;; n++ {}

conformance/conformance_go.go
```diff
@@ -0,0 +1,159 @@
...
+		success, err := doTestIO()
+		if err != nil {
+			log.Fatalf("Error during test: %s", err)
```
rakyll: "Error using tests: %v"

You don't have to worry about the type.

rakyll:
I left a few high-level comments for the Go parts. I don't have much context about conformance and would delegate its review to someone else.

conformance/conformance_go.go
tswast: Because in this case it is a success. If we got 0 bytes it means that the test is finished. (If we got 1, 2, or 3 bytes it means something is wrong / unexpected)

Should I add a comment? This is the same behavior as the Python and Ruby conformance tests.

rakyll: Shouldn't just returning with nil be sufficient enough?

There are no bytes, finished, success.

tswast: Ah, I see what you mean. Makes sense to me. I'll update it to just return an error and check if error is nil for the success case.

tswast:
Thanks @rakyll That was good feedback. The doTestIO method does feel more Go-like now that it just returns and error.

xfxyjwf:
@zombiezen to take a look.

zombiezen:
I'm not 100% sure which repository should live in, since ideally the conformance tests should be running on every change to the Go repo.  I already open-sourced a conformance test runner here: https://github.com/golang/protobuf/tree/master/_conformance

tswast:
@zombiezen How did you deal with the well known types for your conformance test runner? I had to add a go_package to FieldMask for it to work. https://github.com/google/protobuf/pull/2058

I agree it should run on every change in the Go repo. The thing is the Go repo is not stand-alone, we still need the C++ code for the `protoc` command-line tool that lives in this repo. The changes here actually regenerate the code each time, whereas yours seems to use a pre-generated `conformance.pb.go` file.

zombiezen:
I don't think there were tests for FieldMask at the time I authored the runner.

tswast:
That's probably an argument for having the tests live in this repo, then. That way when the conformance test runner is updated, the Go tests can get updated, too.

tswast:
Closing since the Go repo has its own conformance tests.

(I'd still love to see the shared tests being run, but it seems that's not desired.)

PR #2091, opened 2016-09-08
Allow the JsonFormat.Parser to ignore unknown fields.

dprotaso:


dprotaso:
I signed it!

dprotaso:
Need to fix the committers.

PR #2092, opened 2016-09-08
Allow the JsonFormat.Parser to ignore unknown fields

dprotaso:
The default behaviour of throwing an exception remains

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -224,7 +224,7 @@ public class JsonFormat {
    * Creates a {@link Parser} with default configuration.
    */
   public static Parser parser() {
-    return new Parser(TypeRegistry.getEmptyTypeRegistry());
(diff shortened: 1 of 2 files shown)


java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -247,7 +249,16 @@ public Parser usingTypeRegistry(TypeRegistry registry) {
...
-      return new Parser(registry);
+      return new Parser(registry, this.ignoreUnknownFields);
+    }
+
+    /**
+     * Creates a new {@link Parser} configured to not throw an exception
+     * when an unknown field is encountered. The new Parser clones all other
+     * configurations from this Parser.
+     */
+    public Parser ignoreUnknownFields() {
```
xfxyjwf: Name it "ignoringUnknownFields" to be consistent with other config options.

xfxyjwf:
Thanks for the patch!

skebanga:
Seems this is Java only - did this get ported to the other languages too by any chance?

PR #2111, opened 2016-09-13
Fix the version number for 3.0.2

pherl:


src/google/protobuf/stubs/common.h
```diff
@@ -96,7 +96,7 @@ namespace internal {
 
 // The current version, represented as a single integer to make comparison
 // easier:  major * 10^6 + minor * 10^3 + micro
-#define GOOGLE_PROTOBUF_VERSION 3000000
+#define GOOGLE_PROTOBUF_VERSION 3000002
```
xfxyjwf: Change other 3000000s below as well?

pherl: As minor releases are for bug fixes, I guess we can provide source compatibility among 3.0.x releases. i.e. being compatible with 3.0.0 runtime

PR #2112, opened 2016-09-14
Merge 3.0.x into master

pherl:


pherl:
I think most of the changes are change log updates and cherry picks. The failing tests should be flaky. Merging it now

PR #2121, opened 2016-09-15
issue-2108 [protoc/Java] Excessive copying on buildPartial()

mazurkin:
Pull request for issue-2108: https://github.com/google/protobuf/issues/2108

The build() method call time is too high if a bean has too many fields and only some of these fields are set. 

The reason of that because filed copying are outside bit check condition which differs from merge() method which is reverse for build().

I see no any reason why someone really wants uninitialized fields to be copied on build()

### DIFF
--- a/src/google/protobuf/compiler/java/java_enum_field.cc
+++ b/src/google/protobuf/compiler/java/java_enum_field.cc
@@ -292,11 +292,15 @@ GenerateBuildingCode(io::Printer* printer) const {
   if (SupportFieldPresence(descriptor_->file())) {
     printer->Print(variables_,
       "if ($get_has_field_bit_from_local$) {\n"
+      "  result.$name$_ = $name$_;\n"
       "  $set_has_field_bit_to_local$;\n"
+      "} else {\n"
+      "  result.$name$_ = $default_number$;\n"
       "}\n");
+  } else {
+    printer->Print(variables_,
+      "result.$name$_ = $name$_;\n");
   }
-  printer->Print(variables_,
-    "result.$name$_ = $name$_;\n");
 }
 
 void ImmutableEnumFieldGenerator::
--- a/src/google/protobuf/compiler/java/java_lazy_message_field.cc
+++ b/src/google/protobuf/compiler/java/java_lazy_message_field.cc
@@ -233,12 +233,9 @@ void ImmutableLazyMessageFieldGenerator::
 GenerateBuildingCode(io::Printer* printer) const {
   printer->Print(variables_,
       "if ($get_has_field_bit_from_local$) {\n"
+      "  result.$name$_.set($name$_);\n"
       "  $set_has_field_bit_to_local$;\n"
       "}\n");
-
-  printer->Print(variables_,
-      "result.$name$_.set(\n"
-      "    $name$_);\n");
 }
 
 void ImmutableLazyMessageFieldGenerator::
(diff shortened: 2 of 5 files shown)


mazurkin:
I signed it!

xfxyjwf:
ok to test

src/google/protobuf/compiler/java/java_enum_field.cc
```diff
@@ -292,11 +292,10 @@ GenerateBuildingCode(io::Printer* printer) const {
   if (SupportFieldPresence(descriptor_->file())) {
     printer->Print(variables_,
       "if ($get_has_field_bit_from_local$) {\n"
+      "  result.$name$_ = $name$_;\n"
       "  $set_has_field_bit_to_local$;\n"
       "}\n");
   }
-  printer->Print(variables_,
```
xfxyjwf: I think we need to keep this for the "!SupportFieldPresence" case.

xfxyjwf:
@mazurkin This pull request causes the Java unit tests to fail:
https://grpc-testing.appspot.com/job/protobuf_pull_request/531/testReport/junit/(root)/java_jdk7/java_jdk7/

Can you investigate and fix that?

mazurkin:
Yes, I'll try. I need some time to get used to. 

These test says: "These should all fail; instead they are corrupting data."
```
  for (String s : invalidStrings) {
      // TODO(dweis): These should all fail; instead they are corrupting data.
      CodedOutputStream.computeStringSizeNoTag(s);
      outputWithStream.writeStringNoTag(s);
      outputWithArray.writeStringNoTag(s);
      outputWithByteBuffer.writeStringNoTag(s);
    }
```

I even hasn't changed java_string_field.cc because there is a problem that Bean.Builder initializes string fields with "" statically but Bean doesn't

mazurkin:
The task is more complex than I expected. I didn't take into account default values. Even if default values are not set the 'string' and 'bytes' types have default values anyway: "" and ByteString.EMPTY. So all these cases should be properly handled to pass all the tests (your tests are great!)

I need some time to setup projects and research. I managed to run Java test locally.

But it seems it impossible to avoid excessive copying for 'string' and 'bytes' as some default values should be set in any case. It better to these to be nulls but I totally understand that's impossible without breaking compatibility.

mazurkin:
I have 24μs -> 12μs improvement in my synthetic test on build() call. Unfortunately copying of `string` and `bytes` can't be completely avoided in build() as they have implicit default values ("" and ByteString#EMPTY) even if explicit defaul values are not set.

src/google/protobuf/compiler/java/java_enum_field.cc
mazurkin: Fixed

mazurkin:
JMH benchmark is added to https://github.com/mazurkin/protobuf-test

```
# org.test.benchmarks.ProtoBenchmark

# Run complete. Total time: 00:10:03
Benchmark                                  Mode  Cnt   Score   Error  Units
ProtoBenchmark.serializeFixedFastProto     avgt       11.343          us/op
ProtoBenchmark.serializeJackson            avgt        6.747          us/op
ProtoBenchmark.serializeNanoProto          avgt        1.464          us/op
ProtoBenchmark.serializeProtoStaff         avgt        4.223          us/op
ProtoBenchmark.serializeUpstreamSlowProto  avgt       26.588          us/op
```

xfxyjwf:
@mazurkin Sorry for not following up for so long. I'm running our kokoro tests for this PR. Feel free to comment if you don't get a response after the tests finish.

dsilvasc:
@xfxyjwf  In which version was this released?

PR #2122, opened 2016-09-15
Update the ObjC version checks to support a min and current version.

thomasvl:
- Capture the version used to generated, and the min version required at
  generation time.
- Check at compile time and runtime that generated code isn't from a newer
  version, also check that the min version required is also supported.
- Keep the old constants/macros/functions to special case the last version
  that was working so those generated sources still work until we decide
  otherwise.


thomasvl:
Dropping this, the logic isn't complete.

PR #2129, opened 2016-09-16
Removed java_generate_equals_and_hash option from .proto files

acozzette:
This option is now a no-op and so we can safely clean it up in the files
where it appears.


acozzette:
@xfxyjwf to review

xfxyjwf:
@TeBoring is working on an down-integration which should also include this change. I guess we can wait on the down-integration.

acozzette:
Ok, that works, I'll just delete this pull request then.

PR #2131, opened 2016-09-18
Added Dockerfile

vidbina:
The Dockerfile should provides an env that contains all deps. This should simplify setup for dev and test purposes.

Shall we setup an automated build for this on Docker hub?

[](https://microbadger.com/images/vidbina/protobuf)


vidbina:
I signed it!

bazel-io:

haberman:
@xfxyjwf does a Dockerfile seem like something we should add to the repo?

xfxyjwf:
A Docker file for protobuf alone doesn't seem very useful because it's rare for an user to only need to use protobuf and nothing else. It's probably helpful for users who are learning using protobuf though, but that should goes into the examples directory.

xfxyjwf:
This PR needs to be reworked in any case (merge conflicts, cla:no, and many unneeded commits).

PR #2132, opened 2016-09-18
Bazel build: fix gmock remote repository

damienmg:
With googlecode deprecation, the old url was no longer existing.

Tested with bazel test //... on a mac machine.


damienmg:
Pff mistake sorry

PR #2144, opened 2016-09-20
Fix hash computation for JRuby's RubyMessage

abscondment:
#2037 added a test that exposed a bug in the JRuby extension: `System.identityHashCode` returns a hash that does not consider a Message's values. This means two Messages with identical values will not have identical hashCodes.

This patch uses the pattern from RubyMap to combine the hashCodes from all values in a given message and produce a unique, consistent, value-based hash.


abscondment:
@haberman Here's the fix for the JRuby breakage - thanks!

haberman:
Thanks! I will merge it once the tests pass.

abscondment:
AppVeyor seems to be tripping over itself.

```
Build execution time has reached the maximum allowed time for your plan (60 minutes).
```

haberman:
Travis is good enough for me here. Thanks again for the fix!

PR #2145, opened 2016-09-20
Fix for maps_test.js in JavaScript.

haberman:
Had to strip out some JSPB-format test code,
but also added some .proto test messages that
had been improperly stripped out.


haberman:
Review to @teboring.

TeBoring:
LGTM

Issue #2117, opened 2016-09-15
Massive performance drop in deserializeBinary method in javascript


priyatham-aps:
Hi,

We have been using protobuf 3.0.0-beta-2 in our application and recently tried to upgrade to the latest protobuf-3.0.2 release and observed a massive performance drop when deserializing responses on our frontend.

On debugging and profiling this behavior, we have noticed that a lot of processing takes place around syncMapFields_, which is being called by setRepeatedWrapperField method on our list field, which is taking up the major chunk of time.

In order to benchmark this minimally, we made two messages.
```
message BenchmarkMsgTest1 {
    bytes uid = 1;
    string name = 2;
    repeated BenchmarkMsgTest2 msg = 3;
}

message BenchmarkMsgTest2 {
    bytes uid = 1;
    string name = 2;
}
```

And we ran the following test in protobuf-3.0.0-beta-2 and protobuf-3.0.2 versions (in our typescript world).
```
let m1 = new BenchmarkMsgTest1();
m1.setName("BenchmarkMsgTest1");

let arr = new Array<BenchmarkMsgTest2>(10000);
for (let i = 0; i < 10000; i++) {
	let m2 = new BenchmarkMsgTest2();
	m2.setName("" + i);
	arr[i] = m2;
}
m1.setMsgList(arr);

let ser = new Uint8Array(m1.serializeBinary());
console.log("Message length", ser.length);

let startTime = new Date().getTime();
let m = BenchmarkMsgTest1.deserializeBinary(ser);
let endTime = new Date().getTime();
console.log("deserializeBinary time", endTime - startTime);
```

We got the following results (time in ms):
**protobuf-3.0.0-beta-2**
Message length 78909
deserializeBinary time 37
**protobuf-3.0.2**
Message length 78909
deserializeBinary time 1408

As you can see, the amount of time taken for deserializing this message has gone up by **50** times. And it goes higher and higher for larger messages. We have noticed that response messages in our application which were taking a couple of seconds to deserialize earlier are now taking 5 - 10 minutes to deserialize.

Is anyone else facing this issue? Any kind of help appreciated if there is a way to circumvent this problem. Also, happy to help in further investigation if needed.

Thanks!

xfxyjwf:
@haberman @acozzette

acozzette:
[...]

PR #2146, opened 2016-09-20
Fixed quadratic behavior in JSPB deserialization of repeated fields (#2117)

acozzette:
Currently deserialization of a non-packed binary repeated field is quadratic in
the number of elements, because each time we parse a new element we copy over
all elements we have parsed so far. This CL fixes the performance problem by
having the generated deserialization code just call addX() instead of using
getX() and setX().


acozzette:
@TeBoring to review

TeBoring:
LGTM

PR #2152, opened 2016-09-21
Added new has_bits.h file to cmake/extract_includes.bat.in

acozzette:
This is necessary so that the CMake build knows to include this header
file in the installation include/ directory.


acozzette:
@TeBoring to review

TeBoring:
LGTM

PR #2158, opened 2016-09-22
Update README.md

spacetourist:
PHP specific links aren't up at this time


spacetourist:
I signed it!

haberman:
It looks like one of those links is fixed, but the other is still broken. Bo, do you know the planned timeline for fixing the other link? If not maybe it should be removed for now?

TeBoring:
I have created bug to track this internally. Please remove https://developers.google.com/protocol-buffers/docs/phptutorial for now

PR #2170, opened 2016-09-23
Fix default instance destructor and memory leaks

pherl:
@TeBoring


xfxyjwf:
LGTM

pherl:
Default instance are now allocated statically. 
Introduce a Shutdown() method in the ExplicitDefault to call the destructors when the default instance was initialized. The destructor is needed, because the default instance will allocate a map entry. We should remove the allocation if possible, in a future change.

It also adds special logic for default instance destructors to not delete sub message fields, which point to other default instances. Those default instances are statically allocated and should not be deleted.

PR #2171, opened 2016-09-23
Reduce test length to avoid stack overflow on VS.

xfxyjwf:
It seems gmock is chaining all expectations in a nested structure and adding too many expectations will cause stack overflow when the gmock object is destructed. This pull request makes the test case shorter but still have the same test coverage.

It fixes JsonStreamParserTest.ArrayValues on Visual Studio and also fixes most other failing test cases caused by a corrupted stack.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,16 +20,16 @@ environment:
 test: off
 
 install:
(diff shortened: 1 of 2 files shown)


xfxyjwf:
@pherl to review

pherl:
LGTM

PR #2172, opened 2016-09-23
Remove missing static analyzer warning.

sergiocampama:
Is there any real problem in switching GOOGLE_CHECK for assert?

### DIFF
--- a/src/google/protobuf/map_field.cc
+++ b/src/google/protobuf/map_field.cc
@@ -94,7 +94,7 @@ int MapFieldBase::SpaceUsedExcludingSelfNoLock() const {
 
 void MapFieldBase::InitMetadataOnce() const {
   GOOGLE_CHECK(entry_descriptor_ != NULL);
-  GOOGLE_CHECK(assign_descriptor_callback_ != NULL);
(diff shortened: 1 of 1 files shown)


sergiocampama:
@xfxyjwf fyi

sergiocampama:
ping?

xfxyjwf:
GOOGLE_CHECK (or CHECK in google3) is the standard way to do precondition check. It has nice logging/stack trace support and works in both debug/release build. assert() is just not the equivalent and I don't think it's a good idea to switch to assert(). It's especially troublesome to have both GOOGLE_CHECK and assert in the same function in adjacent lines. This inconsistency will lead to confusion and your "fix" won't last long before someone cleans it up and reverts everything back to GOOGLE_CHECK.

I think it's not worthwhile to fix these warnings unless there is a solution that doesn't conflict with google3 convention and can prevent future inclusion of the same type of warnings.

I probably already asked this question once, but is there no way to teach the static analyzer to understand GOOGLE_CHECK? In the worst case we can add a #ifdef for the static analyzer to #define GOOGLE_CHECK assert?

PR #2176, opened 2016-09-24
Disable a problematic test for MSVC for now.

xfxyjwf:


xfxyjwf:
Well, disabled one test, another one starts to fail with the same symptom. Not a good sign... Need more investigation.

PR #2196, opened 2016-09-28
Added alias getFieldProto3 as used by older generated code.

haberman:
Un-breaks users who have old generated code and upgrade
to the 3.1.0 release.

CC/review: @murgatroid99, @nicolasnoble.

I wasn't able to test this against old generated code. If you have any tests that would exercise this case, let me know. Otherwise I'll try to create an artificial test for it.

### DIFF
--- a/js/message.js
+++ b/js/message.js
@@ -755,6 +755,18 @@ jspb.Message.getFieldWithDefault = function(msg, fieldNumber, defaultValue) {
 
 
 /**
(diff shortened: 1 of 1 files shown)


murgatroid99:
The gRPC Node tests use at least one `_pb.js` file generated with protoc 3.0. Those tests fail with the published 3.1.0 package, but pass with this change. LGTM.

haberman:
Great news. I will merge it now.

Issue #1746, opened 2016-07-05
require in commonjs does not respect folder structure


itavy:
this one relates to [1684](https://github.com/google/protobuf/issues/1684) 

i have 3 files, each one on a different directory:
./Base.proto
./User/User.proto 
./User/Web/Web.proto 

last one include the other 2;

in generated code for Web.proto (`protoc User/Web/Web.proto --js_out=import_style=commonjs,binary:protos -I=. -I=User -I=User/Web`) and the second require statement is wrong:

``` javascript
$ head -13 protos/User/Web/Web_pb.js 
/**
 * @fileoverview
 * @enhanceable
 * @public
 */
// GENERATED CODE -- DO NOT EDIT!

var jspb = require('google-protobuf');
var goog = jspb;
var global = Function('return this')();

var Base_pb = require('../../Base_pb.js');
var User_pb = require('../../User_pb.js');
```

it should be:

``` javascript
var User_pb = require('../User_pb.js');
```

below are the 3 proto files:

``` javascript
$ cat Base.proto 
syntax = "proto3";

package itavy.Base;

enum AppVersion {
  V_0_0_1 = 0;
  V_0_0_2 = 1;
};
```

``` javascript
$ cat User/User.proto 
syntax = "proto3";

package itavy.User;

enum Type {
  REGULAR = 0;
  ADMIN = 1;
};
```

``` javascript
$ cat User/Web/Web.proto 
syntax = "proto3";

package itavy.User.Web;

import "Base.proto";
import "User.proto";

message User {
  itavy.Base.AppVersion version = 1;
  string name = 2;
  string email = 3;
  itavy.User.Type type = 4;
};
```

graywolf336:
Any chance we can get an update on this? Ran into an issue related to this..

anandolee:
@haberman any updates or should we close it?

haberman:
Sorry for the (very) slow response. This isn't a bug in protoc. For include paths to work correctly, the import paths should be relative to the base.

Web.proto should start like this:

```
syntax = "proto3";

package itavy.User.Web;

import "Base.proto";
import "User/User.proto";
```

Then you shouldn't need any `-I` flags, and all of the import paths should work properly.

PR #2198, opened 2016-09-28
Adds a new import-mapping option to the js_generator.

### LINKED ISSUE #1745: javascript generated code doesn't work when you use dot in a file name and include it
i wrote 2 proto files (itavy.pkg1.proto and itavy.pkg2.proto)

``` javascript
syntax = "proto3";

package itavy.pkg1;

enum AppVersion {
  V_0_0_1 = 0;
  V_0_0_2 = 1;
};
```

``` javascript
syntax = "proto3";

package itavy.pkg2;

import "itavy.pkg1.proto";

message User {
  itavy.pkg1.AppVersion version = 1;
  string name = 2;
  string email = 3;
};
```

after i generate the js code `protoc itavy.pkg2.proto --js_out=import_style=commonjs,binary:protos`
it will create file protos/itavy.pkg2_pb.js 

``` javascript
$ head -12 protos/itavy.pkg2_pb.js 
/**
 * @fileoverview
 * @enhanceable
 * @public
 */
// GENERATED CODE -- DO NOT EDIT!

var jspb = require('google-protobuf');
var goog = jspb;
var global = Function('return this')();

var itavy.pkg1_pb = require('./itavy.pkg1_pb.js');
```

that last line will break if you try to run the script:

```
$ node protos/itavy.pkg2_pb.js 
/home/itavy/tmp/gproto/protos/itavy.pkg2_pb.js:12
var itavy.pkg1_pb = require('./itavy.pkg1_pb.js');
         ^
SyntaxError: Unexpected token .
    at Object.exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:513:28)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Function.Module.runMain (module.js:575:10)
    at startup (node.js:160:18)
    at node.js:456:3

```

pcj:
This option is directly modeled after protoc-gen-go's ImportMap
'Mfoo=bar' option, adapted for use in a commonjs context.  It enables
the caller to provide override values for 'require()' statements in
generated code.  For example,
'--js_out=Mexample/foo.proto=example-foo:outdir/ would generate
`require('example-foo')` instead of `require('./example/foo_js.pb')`.
Multiple 'M' options are allowed.

The main use case is to permit the use of module-names rather than
filenames for required files, allowing users to use the more flexible
module loading schemes in lieu of hardcoded filenames.

It also fixes #1745 and provides a workaround for #1746.

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -191,6 +191,7 @@ string ModuleAlias(const string& filename) {
   // exposed to users so we can change it later if we need to.
   string basename = StripProto(filename);
   StripString(&basename, "-", '$');
+  StripString(&basename, ".", '$'); // fixes #1745.
   StripString(&basename, "/", '_');
   return basename + "_pb";
 }
@@ -3016,6 +3017,24 @@ bool GeneratorOptions::ParseFromOptions(
         *error = "Unknown import style " + options[i].second + ", expected " +
                  "one of: closure, commonjs, browser, es6.";
       }
+    } else if (options[i].first.at(0) == 'M') {
+      string key = options[i].first.substr(1);
+      string val = options[i].second;
+      if (key.empty()) {
+        *error = "Missing importmap filename value (lhs).";
+        return false;
+      }
+      if (key.size() == StripProto(key).size()) {
+        *error = "The importmap filename value (" + key +
+                 ") should name a file having a .proto file suffix";
+        return false;
+      }
+      if (val.empty()) {
+        *error = "Missing importmap replacement operand value (rhs).";
+        return false;
+      }
+      import_map->insert(std::make_pair(key, val));
+      return true;
     } else {
       // Assume any other option is an output directory, as long as it is a bare
       // `key` rather than a `key=value` option.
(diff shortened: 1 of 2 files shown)


pcj:
@xfxyjwf can you review or assign this?  Thanks!

haberman:
Yikes, sorry for the delay on this.

This request seems related to this comment in the code: https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/js/js_generator.cc#L164

I can see the need for this, but it seems pretty manual to have to list out all of the filename/module pairs explicitly. You would have to update your `protoc` invocation every time one of your dependencies adds a new file. Also putting this all in the `--js_out` argument seems pretty noisy and unfriendly.

pcj:
After long hiatus, I'd like to pick this back up again.  This would simplify protobuf generation for node.js support in the repo I maintain that assists with protobuf generation for the bazel build tool (node.js suport there still isn't fully correct with complex proto dependencies).  

@haberman I understand your concern about the js_out being noisy, but I don't think it's any less noisy than the M option in protoc-gen-go.

If you're OK with this, I'd rebase on latest and go from there.

Thanks!

xfxyjwf:
@pcj We generally do not want to add options to protobuf. The style that protoc-gen-go uses is a particularly bad style to follow IMO. Right now we are working on improving protobuf JS support with bazel and I feel that there should be a better way to support it without complications like this.

Assigned to @BSBandme who is looking into protobuf JS support in bazel/blaze.

anandolee:
Ping @BSBandme, any update?

haberman:
I think this needs more discussion before it is ready to go to the implementation stage. One possibility I could see is a file-level option to indicate the npm package. Then you wouldn't have to explicitly specify the package for your dependencies, it would come from your dependency's .proto file.

Issue #1594, opened 2016-05-24
Unable to install google-protobuf gem on JRuby


blowmage:
The glcoud-ruby project has an issue opened by a user (GoogleCloudPlatform/gcloud-ruby#712) that is unable to install the latest version due to the dependencies on the grpc and google-protobuf gems not installing on jruby. I am not able to get the google-protobuf gem to install using versions 9.1.1.0 or 9.0.0.0.

Here is the install on mac:

```
$ ruby -v
jruby 9.1.1.0 (2.3.0) 2016-05-19 fe84e89 Java HotSpot(TM) 64-Bit Server VM 25.91-b14 on 1.8.0_91-b14 +jit [darwin-x86_64]

$ gem install --ignore-dependencies google-protobuf -v 3.0.0.alpha.5.0.5.1
Building native extensions.  This could take a while...
ERROR:  Error installing google-protobuf:
    ERROR: Failed to build gem native extension.

    current directory: /Users/blowmage/.gem/repos/jruby-9.1.1.0-grpc/gems/google-protobuf-3.0.0.alpha.5.0.5.1/ext/google/protobuf_c
/Users/blowmage/.rubies/jruby-9.1.1.0/bin/jruby -r ./siteconf20160524-43044-1fuqmwb.rb extconf.rb
NotImplementedError: C extensions are not supported
    <top> at /Users/blowmage/.rubies/jruby-9.1.1.0/lib/ruby/stdlib/mkmf.rb:1
  require at org/jruby/RubyKernel.java:944
   (root) at /Users/blowmage/.rubies/jruby-9.1.1.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
    <top> at extconf.rb:3

extconf failed, exit code 1

Gem files will remain installed in /Users/blowmage/.gem/repos/jruby-9.1.1.0-grpc/gems/google-protobuf-3.0.0.alpha.5.0.5.1 for inspection.
Results logged to /Users/blowmage/.gem/repos/jruby-9.1.1.0-grpc/extensions/universal-java-1.8/2.3.0/google-protobuf-3.0.0.alpha.5.0.5.1/gem_make.out
```

Here is the install on linux:

```
$ ./bin/jruby -v
jruby 9.1.1.0 (2.3.0) 2016-05-19 fe84e89 OpenJDK 64-Bit Server VM 24.95-b01 on 1.7.0_95-b00 +jit [linux-x86_64]

[...]

PR #2199, opened 2016-09-29
Set platform to "java" under JRuby

JasonLunn:
Proposed fix for #1594


JasonLunn:
`Gemfile.lock` has been removed in this PR because it was leading to broken tests. It is worth noting that the breakage seen under in [travis](https://travis-ci.org/google/protobuf/jobs/163611664#L1469) before it was removed was not reproducible locally while I was working under JRuby 9.1.5.0 but was immediately reproducible when I changed back to 1.7.x.

Currently the `Gemfile` defers all dependency management to the gemspec, so if there are concerns about having unpinned development dependencies, we could solve that by adding additional parameters to the calls to `add_development_dependency` within `google-protobuf.gemspec`.

JasonLunn:
CC: @blowmage

blowmage:
Whoah! Does this work? Can the gem be installed on JRuby with this change?

JasonLunn:
@blowmage - Works for me locally, and passed the travis test. I am still waiting for an answer on what impact this would have, if any, on the gem publishing process. Also, RubyGems has gone to v3.1.0, so I expect there is some large upstream merge about to land here that could delay this getting merged.

JasonLunn:
CC: @apolcyn

apolcyn:
I tried this change locally and it to work for me too! That is, I'm building the gem as described in https://github.com/google/protobuf/tree/master/ruby, and the -java suffix does appear now.
cc @haberman I think this is what's needed for jruby publish to work, though I don't think I can approve/merge this.

JasonLunn:
@apolcyn - is there anything about the code quality or style of the PR that needs improvement from your perspective in order to be eligible to be merged, or are you blocked by an authorization / policy issue?

apolcyn:
Code wise, I would think we'd want to keep the Gemfile.lock checked in, to keep actual versions used.
Merging this is an authorization issue for me though.

JasonLunn:
@apolcyn - In your opinion is the enhancing the specificity of the calls to `add_development_dependency` as suggested above an adequate alternative to restoring `Gemfile.lock`?

apolcyn:
@JasonLunn  - thanks I had missed that. In the case I think that making development dependencies more specific would work. That said, it looked to me like the travis failure was due to not finding the actual google-protobuf gem on jruby, wondering if it has to do with the gem just not being built or available somehow, in that test?

anyways though, I don't have authorization for this - I haven't been working on this codebase. cc @haberman

JasonLunn:
Travis seems a little backed up at the moment, but I've added the requirements to the development dependencies. Note that this does not result in the exact same patch level versions as appeared in the `Gemfile.lock`. I could lock it down even further if that is perceived as a defect.

JasonLunn:
Checks all passed - any other feedback?

JasonLunn:
@haberman - any questions or concerns?

JasonLunn:
@apolcyn - Is there an internal discussion going on about this PR out of band?

apolcyn:
@JasonLunn From what I know, this change looks good; I think it should fix the jruby package issue. But I don't have further approval to make.

ruby/google-protobuf.gemspec
```diff
@@ -10,16 +10,17 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
   s.files       = Dir.glob('lib/**/*.rb')
   if RUBY_PLATFORM == "java"
+    s.platform  = "java"
     s.files     += ["lib/google/protobuf_java.jar"]
   else
     s.files     += Dir.glob('ext/**/*')
     s.extensions= ["ext/google/protobuf_c/extconf.rb"]
-    s.add_development_dependency "rake-compiler-dock"
+    s.add_development_dependency "rake-compiler-dock", "~> 0.5.1"
```
haberman: Is it necessary to specify versions? When/why will we update these versions?

JasonLunn: It was added in response to commentary above about the removal of `Gemfile.lock`. The tests pass with the latest versions of all the gems that have version requirements specified. I believe the concern is that to have completely unpinned development dependencies could introduce flakiness in the automated tests.

JasonLunn:
Thanks all - is there any other work need from my end to change the distribution process so that both gems for both platforms are published during subsequent releases?

JasonLunn:
@haberman - Is there anything keeping the 3.1.0 build of the gem being published for the JRuby platform?

TeBoring:
I think this PR breaks the jenkins test for ruby.
https://grpc-testing.appspot.com/job/protobuf_branch/206/

JasonLunn:
@TeBoring - It is hard to tell from the just the output, but since an essential part of this PR was to remove `Gemfile.lock` (because it cannot be right for both JRuby and non-JRuby platforms simultaneously), my guess is that `Gemfile.lock` is being created by `bundle install` under one Ruby and not being cleared out before `bundle install` is rerun under JRuby.

JasonLunn:
Travis appears to be configured to run the various ruby tests [individually](https://travis-ci.org/google/protobuf/builds/163850686), rather than invoking `ruby_all` in `test.sh`. We could solve the failure in Jenkins a couple of different ways I think -
 1. Change Jenkins to run each test separately and remove the broken `ruby_all` since its current implementation assumes that you can repeatedly run `bundle` install in the same directory with different Ruby interpreters without managing `Gemfile.lock`
 1. Fix `ruby_all` by explicitly removing `Gemfile.lock` in each of `build_ruby_21`, `build_ruby_22`, and `build_jruby`
 1. Fix `ruby_all` by running `git clean`

Preferences?

xfxyjwf:
I think we need a fix that allows us to run:
```bash
$ ./tests.sh ruby21
$ ./tests.sh ruby22
$ ./tests.sh jruby
```
without failing tests. Only (2) can satisfy this?

JasonLunn:
I think (3) should remove a superset of the files removed by (2) - I'm theorizing that `Gemfile.lock` is enough but I'm not sure. I'll put up a new PR and link it here.

JasonLunn:
#2254 is running through Travis now - are Jenkins builds kicked off automatically?

steakknife:
Use the Applicable to generate separate for gemfiles if necessary.

Remove the Gemfile.lock from the repo, deps should be specified in the gemspec and the Gemfile should only contain:

    source "https://rubygems.org"
    gemspec

Issue #2179, opened 2016-09-25
v3.1.0 and openrtb.proto


va5ja:
Hi there,

I'm getting some errors trying to generate PHP classes with `protoc --php_out=php openrtb.proto` using version 3.1.0 (default proto2 syntax):

```
openrtb.proto:2045:3: When enum name is stripped and label is PascalCased (Iab11), this value label conflicts with IAB1_1
openrtb.proto:2051:3: When enum name is stripped and label is PascalCased (Iab12), this value label conflicts with IAB1_2
openrtb.proto:2055:3: When enum name is stripped and label is PascalCased (Iab13), this value label conflicts with IAB1_3
openrtb.proto:2068:3: When enum name is stripped and label is PascalCased (Iab14), this value label conflicts with IAB1_4
openrtb.proto:2077:3: When enum name is stripped and label is PascalCased (Iab15), this value label conflicts with IAB1_5
openrtb.proto:2088:3: When enum name is stripped and label is PascalCased (Iab16), this value label conflicts with IAB1_6
openrtb.proto:2096:3: When enum name is stripped and label is PascalCased (Iab17), this value label conflicts with IAB1_7
openrtb.proto:2213:3: When enum name is stripped and label is PascalCased (Iab21), this value label conflicts with IAB2_1
openrtb.proto:2214:3: When enum name is stripped and label is PascalCased (Iab211), this value label conflicts with IAB2_11
openrtb.proto:2215:3: When enum name is stripped and label is PascalCased (Iab212), this value label conflicts with IAB2_12
openrtb.proto:2216:3: When enum name is stripped and label is PascalCased (Iab213), this value label conflicts with IAB2_13
openrtb.proto:2217:3: When enum name is stripped and label is PascalCased (Iab22), this value label conflicts with IAB2_2
openrtb.proto:2218:3: When enum name is stripped and label is PascalCased (Iab221), this value label conflicts with IAB2_21
openrtb.proto:2219:3: When enum name is stripped and label is PascalCased (Iab222), this value label conflicts with IAB2_22
openrtb.proto:2220:3: When enum name is stripped and label is PascalCased (Iab223), this value label conflicts with IAB2_23
openrtb.proto:2222:3: When enum name is stripped and label is PascalCased (Iab23), this value label conflicts with IAB2_3
openrtb.proto:2233:3: When enum name is stripped and label is PascalCased (Iab24), this value label conflicts with IAB2_4
[...]

PR #2204, opened 2016-09-29
Do strict enum name checking only for proto3

acozzette:
There seem to already be .proto files out there that have conflicting
enum names, which will not be able to build successfully for some
languages (like C#). To prevent this problem from spreading, let's make
it an error for proto3 but just issue a warning for proto2. This fixes
issue #2179.


acozzette:
@haberman to review

haberman:
LGTM.

acozzette:
retest this please.

PR #2212, opened 2016-09-30
fix BUILD :protoc rule by adding missing php .cc file missed in https://github.com/google/protobuf/pull/2052

pmbethe09:


pmbethe09:
already fixed in the 3.1.x branch

Issue #1823, opened 2016-07-21
JsonFormat: why is Int64 being converted to a string ?


darkcrawler01:
I am using the `JsonFormat.Printer` util class for converting Protos to JSON.
and I noticed that the Int64 is converted to `String` instead of keeping it a number/integer. 
https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L880-L883

I am sure there is a good reason but I am not able to figure it out.
can someone explain ?

xfxyjwf:
See: https://groups.google.com/forum/#!topic/protobuf/4-BY-k-Lk-g

Quoted Jon's comment:
"It's due to the way that all numbers in JSON are basically IEEE-754 64-bit doubles... which means that any 64-bit integers can lose information. Using a string instead, all 64-bit integers can be round-tripped with no loss of information."

mickeyreiss:
@xfxyjwf would you be open to a patch that makes it possible to override this behavior change? I'm having trouble upgrading to the latest protobuf library without making a breaking change for clients that consume JSON.

xfxyjwf:
@mickeyreiss I don't remember we have made a behavior change on that. As far as I know, the JsonFormat class prints int64 as strings from the beginning. Are you migrating from some other JSON implementation?

mickeyreiss:
Yes, we are migrating from [`JsonJacksonFormat`](https://github.com/bivas/protobuf-java-format/blob/master/src/main/java/com/googlecode/protobuf/format/JsonJacksonFormat.java#L217) from [com.googlecode.protobuf-java-format:protobuf-java-format:1.4](https://mvnrepository.com/artifact/com.googlecode.protobuf-java-format/protobuf-java-format/1.4).

PR #2214, opened 2016-09-30
Add lossy64BitNumberRepresentations option to JsonFormat.Printer

mickeyreiss:
Here's my WIP for addressing #1823. @xfxyjwf let me know if you'd be interested in accepting this or something similar, and I'd be happy to clean it up/fix tests, whatever it takes.

I don't love this change on principle, but it will help my organization out with backwards compatibility.

Thanks!

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -101,7 +101,7 @@ public class JsonFormat {
    * Creates a {@link Printer} with default configurations.
    */
   public static Printer printer() {
-    return new Printer(TypeRegistry.getEmptyTypeRegistry(), false, false, false);
+    return new Printer(TypeRegistry.getEmptyTypeRegistry(), false, false, false, false);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


xfxyjwf:
In principle we would like minimize the number of custom options on the JsonFormat class because the more options we have, the harder it is to make all implementations conform to the same proto3 JSON spec. In this particular case, it's not just one more option, but also an option that breaks interoperability between different proto3 JSON implementations. For this reason we probably won't accept any such changes.

I don't have a great answer to your migration problem though. You probably need to find some other workaround without direct support from protobuf JsonFormat class. Sorry for that...

mickeyreiss:
Thanks for taking a look. You make a great point about conformant JSON implementations.

Would you be open to opening up the `JsonFormat` class to extension via subclassing?

xfxyjwf:
This is a very late comment but we do not want to allow subclassing either. The idea is to force everyone to use a standard format that's easy to communicate across different implementations/languages. Giving users ability to customize the format will defeat this purpose (unless the customization is part of the spec and implemented everywhere).

jonny-improbable:
@xfxyjwf Thanks for commenting on this issue; it's regrettable that there is still no solution for preserving int64 values when using JavaScript (also see https://github.com/google/protobuf/pull/1832) - this is a big problem for a lot of teams as large numeric values are not preserved from client <> server.

Changing int64 to be represented by a `string` instead of a `number` would be a breaking change, but is the only way teams can have any confidence when using javascript clients.

Please could one of the core maintainers provide a path forward, or at least some input, on this issue?

Thanks

Issue #2228, opened 2016-10-06
parseDelimitedFrom behavior has changed between pb2.5 and 3.1


saintstack:
parseDelimitedFrom didn't check limit in 2.5. In 3.1 it does throwing InvalidProtocolBufferException if we exceed the hard-coded max of 64MB.

This is issue is a bit like "Remove proto size restriction when parsing protos. #836" but parseDelimitedFrom only takes InputStream so we can't set max size.

64MB is large but previous there was no limit. There is a sort of a workaround by doing parsePartialFrom and creating a CodedInputStream probably set to this method but means creating a new object every time (or reading to see if > 64MB and then making new CIS...)

Let us know how you think best to address -- we'd like to have the no-limit back if only for compatibility -- and we can put up a patch.

Thanks.

`org.apache.hadoop.hbase.shaded.com.google.protobuf.InvalidProtocolBufferException: Protocol message was too large.  May be malicious.  Use CodedInputStream.setSizeLimit() to increase the size limit.
at org.apache.hadoop.hbase.shaded.com.google.protobuf.InvalidProtocolBufferException.sizeLimitExceeded(InvalidProtocolBufferException.java:134)
at org.apache.hadoop.hbase.shaded.com.google.protobuf.CodedInputStream$StreamDecoder.readRawBytesSlowPath(CodedInputStream.java:2768)
at org.apache.hadoop.hbase.shaded.com.google.protobuf.CodedInputStream$StreamDecoder.readBytes(CodedInputStream.java:2335)
at org.apache.hadoop.hbase.shaded.protobuf.generated.ProcedureProtos$Procedure.<init>(ProcedureProtos.java:564)
at org.apache.hadoop.hbase.shaded.protobuf.generated.ProcedureProtos$Procedure.<init>(ProcedureProtos.java:423)
at org.apache.hadoop.hbase.shaded.protobuf.generated.ProcedureProtos$Procedure$1.parsePartialFrom(ProcedureProtos.java:2498)
at org.apache.hadoop.hbase.shaded.protobuf.generated.ProcedureProtos$Procedure$1.parsePartialFrom(ProcedureProtos.java:2493)
at org.apache.hadoop.hbase.shaded.com.google.protobuf.CodedInputStream$StreamDecoder.readMessage(CodedInputStream.java:2314)
at org.apache.hadoop.hbase.shaded.protobuf.generated.ProcedureProtos$ProcedureWALEntry.<init>(ProcedureProtos.java:6565)
[...]

PR #2231, opened 2016-10-07
Change CodedInputStream#DEFAULT_SIZE_LIMIT from 64MB to kint32max (0x7FFFFFF)

saintstack:
Fix for "parseDelimitedFrom behavior has changed between pb2.5 and 3.1 #2228"

Seems a bit tough adding a test for something so primitive.

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/CodedInputStream.java
+++ b/java/core/src/main/java/com/google/protobuf/CodedInputStream.java
@@ -885,7 +885,8 @@ public final class CodedInputStream {
   private int sizeLimit = DEFAULT_SIZE_LIMIT;
 
   private static final int DEFAULT_RECURSION_LIMIT = 100;
-  private static final int DEFAULT_SIZE_LIMIT = 64 << 20;  // 64MB
+  // Integer.MAX_VALUE == ./src/google/protobuf/stubs/port.h:static const int32 kint32max = 0x7FFFFFFF;
+  private static final int DEFAULT_SIZE_LIMIT = Integer.MAX_VALUE;
   private static final int BUFFER_SIZE = 4096;
 
   private CodedInputStream(


xfxyjwf:
Can you limit the change to parseDelimitedFrom only? If we are to change the default limit, I would  change all languages together after getting a consensus from other protobuf team members.

apollo-time:
If create a CodedInputStream that reads from the given flat array that size is bigger than 64MB, couldn't parse graph.
I think CodedInputStream(const uint8* buffer, int size) function set total_bytes_limit_ with size when size is bigger 64MB(kDefaultTotalBytesLimit).

saintstack:
On the CLA, "I've signed it!": https://cla.developers.google.com/clas/edit?id=107138090177356171248&kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE&redirect=1

xfxyjwf:
@saintstack You probably need to recreate the commit and make sure the commit is made using the email address associated with your github account.

saintstack:
@xfxyjwf  Thank you for the guidance. Fixed discrepancy between pullrequest email and what was in CLA.

Regards your suggestion of limiting change to parseDelimitedFrom only, do you mean in AbstractParser in here:

```java
   @Override
   public MessageType parsePartialDelimitedFrom(
       InputStream input, ExtensionRegistryLite extensionRegistry)
       throws InvalidProtocolBufferException {
     int size;
     try {
       int firstByte = input.read();
       if (firstByte == -1) {
         return null;
       }
       size = CodedInputStream.readRawVarint32(firstByte, input);
     } catch (IOException e) {
       throw new InvalidProtocolBufferException(e.getMessage());
     }
     InputStream limitedInput = new LimitedInputStream(input, size);
     return parsePartialFrom(limitedInput, extensionRegistry);
   }
```

Something like `if (size > DEFAULT_SIZE_LIMIT) throw InvalidProtocolBufferException`?

@apollo-time Please come again. I'm not sure I follow. Thanks (Looking in C++ code, seems like it defaults 64MB limit `   static const int kDefaultTotalBytesLimit = 64 << 20;  // 64MB`).

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
```diff
@@ -885,7 +885,8 @@ public static long decodeZigZag64(final long n) {
   private int sizeLimit = DEFAULT_SIZE_LIMIT;
 
   private static final int DEFAULT_RECURSION_LIMIT = 100;
-  private static final int DEFAULT_SIZE_LIMIT = 64 << 20;  // 64MB
+  // Integer.MAX_VALUE == ./src/google/protobuf/stubs/port.h:static const int32 kint32max = 0x7FFFFFFF;
```
xfxyjwf: Remove this comment? It says what everybody should already know.

xfxyjwf:
I have recently changed the default C++ limit to kint32max:
https://github.com/google/protobuf/blob/master/src/google/protobuf/io/coded_stream.h#L640

It should be fine to apply the same change to Java. Though could you help add two test cases to verify: 1. parsing can succeed for a message close to 2G; 2. parsing will fail if the message is larger than 2G? Basically the Java version of the test I added here:
https://github.com/google/protobuf/blob/master/src/google/protobuf/message_unittest.cc#L344

saintstack:
@xfxyjwf  Thanks for the input. I opened a new pull request #2431. Let me close this one.

PR #2245, opened 2016-10-11
Add 32-bit machine test on jenkins.

TeBoring:


TeBoring:
@xfxyjwf

TeBoring:
Intended failure on 32-bit machine.

jenkins/pull_request_in_docker32.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
xfxyjwf: This file is mostly the same as pull_request_in_docker.sh except for the actual test list. I think we should use the same file and use an ENV variable to control the test set.

jenkins/buildcmds/pull_request_32.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+#
+# This is the top-level script we give to Jenkins as the entry point for
+# running the "pull request 32" project:
+#
+#   https://grpc-testing.appspot.com/view/Protocol%20Buffers/job/protobuf_pull_request/
```
xfxyjwf: Update this link

jenkins/pull_request_in_docker32.sh
TeBoring: Done

jenkins/buildcmds/pull_request_32.sh
TeBoring: Done

xfxyjwf:
LGTM

PR #2253, opened 2016-10-13
Clear Gemfile.lock before jruby test.

TeBoring:


TeBoring:
Fix #2199

TeBoring:
The problem in #2199 is fixed in this PR. The jenkins test is broken due to ruby's own problem.

TeBoring:
@xfxyjwf

PR #2251, opened 2016-10-13
Merge 3.1.x branch into master.

TeBoring:

PR #2254, opened 2016-10-14
Use git clean before installing via bundler

JasonLunn:
Should fix problems in Jenkins introduce by #2199 not seen in Travis due to the use of `./test.sh ruby_all` which runs every platform in the same directory

### DIFF
--- a/ruby/travis-test.sh
+++ b/ruby/travis-test.sh
@@ -10,12 +10,14 @@ test_version() {
     bash --login -c \
       "rvm install $version && rvm use $version && rvm get head && \
        which ruby && \
+       git clean -f && \
        gem install bundler && bundle && \
        rake test"
   else
     bash --login -c \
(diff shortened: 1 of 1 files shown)


JasonLunn:
@xfxyjwf / @TeBoring - can either of you make Jenkins run on this PR to see if it solves the [JRuby issue](https://grpc-testing.appspot.com/job/protobuf_branch/206/) observed in #2251

xfxyjwf:
ok to test

xfxyjwf:
I think "git clean" will actually break the tests because all non-C++ tests depend on the protoc to be built. After git clean, the built protoc binary will be gone.

TeBoring:
#2253 should have fixed the problem.

JasonLunn:
Seems to have worked. I'm unclear why #2253 didn't pass in Jenkins

PR #2260, opened 2016-10-16
Update autogen.sh

mridulnagpal:
For the people who had installed gmock before had gmock named as googlemock-release-1.7.0 because they didn't rename it. Hence if now they have it they won't have to download gmock.zip again.


mridulnagpal:
I signed it

PR #2264, opened 2016-10-18
Place Python extensions correctly in Bazel build.

rshin:
Currently google/protobuf/pyext/_message.so and google/protobuf/internal/_api_implementation.so do not end up in the correct place in the Bazel build (i.e. they don't live with all the other Python files at google/protobuf/...), so they are never loaded as intended.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -577,7 +577,7 @@ py_library(
 )
 
 cc_binary(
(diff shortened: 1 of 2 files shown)


xfxyjwf:
Looks good to me. Thanks for the patch!

PR #2282, opened 2016-10-25
Fix pure php implementation for 32-bit machine.

TeBoring:


TeBoring:
@stanley-cheung

TeBoring:
The broken test for 32bit jenkins depends on head of protobuf repository. Once this PR is submitted, it will be fine.

PR #2284, opened 2016-10-26
Support extra parameters for plugins.

pherl:


src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -262,6 +262,10 @@ void AddDefaultProtoPaths(vector<pair<string, string> >* paths) {
     return;
   }
 }
+
+string PluginName(const string& plugin_prefix, const string& directive) {
```
thomasvl: Maybe document that this assumes "--" on the front and "_out"/"_opt" on the end?

pherl: Done.

PR #2288, opened 2016-10-26
Trigger update of docker for new changes in #2282.

TeBoring:


TeBoring:
@xfxyjwf

jenkins/docker/Dockerfile
```diff
@@ -130,7 +130,7 @@ ENV MAVEN_REPO /var/maven_local_repository
 ENV MVN mvn --batch-mode
 
 RUN cd /tmp && \
-  git clone https://github.com/google/protobuf.git && \
+  git clone https://github.com/google/protobuf.git &&  \
```
xfxyjwf: Does this really make a difference? How about changing "make -j4" to "make -j6"?

TeBoring: It has to be in the line of git clone. Otherwise, jenkins will keeping using the cache.

TeBoring: jenkins tests has passed.

TeBoring: Not every file contains -j4

jenkins/docker32/Dockerfile
```diff
@@ -62,7 +62,7 @@ RUN php composer-setup.php
 RUN mv composer.phar /usr/bin/composer
 RUN php -r "unlink('composer-setup.php');"
 RUN cd /tmp && \
-  git clone https://github.com/google/protobuf.git && \
+  git clone https://github.com/google/protobuf.git &&  \
```
TeBoring: The stuff composer installs depends on composer.json which is contained in protobuf repository.

xfxyjwf: Why we need to install it in this Dockerfile rather than in the tests.sh file? Can you run composer install as part of the actual test?

TeBoring: Why should we do that? Keep it here avoids download the composer stuff for each test.

TeBoring:
It has to be the line for git clone. Otherwise, Jenkins still use the cache.

PR #2296, opened 2016-10-27
JavaScript: support using Uint8Array in toObject

seishun:
Currently, `toObject` always outputs `bytes` fields as base64-encoded strings. This is understandable for proto3, which has a canonical JSON mapping, but isn't necessary for proto2. I propose using `Uint8Array` for `bytes` fields in proto2 when binary support is enabled.

My usecase is described [here](https://github.com/google/protobuf/issues/1591#issuecomment-239203004). In short: I want a way to convert Protobuf messages to plain JS objects without having to hard-code fields for each message I use. `toObject` works for me for the most part (pending #1951 and #1922), but I would prefer not to have to decode base64 back into a `Buffer` or `Uint8Array`.

Consider this a feature request with proof-of-concept implementation attached.

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -2162,8 +2162,11 @@ void Generator::GenerateClassFieldToObject(const GeneratorOptions& options,
     }
   } else if (field->type() == FieldDescriptor::TYPE_BYTES) {
     // For bytes fields we want to always return the B64 data.
+    // Unless it's proto2 and the user explicitly wants binary.
+    auto mode = field->file()->syntax() == FileDescriptor::SYNTAX_PROTO2 &&
+                options.binary ? BytesMode::BYTES_U8 : BytesMode::BYTES_B64;
     printer->Print("msg.get$getter$()",
-                   "getter", JSGetterName(options, field, BYTES_B64));
+                   "getter", JSGetterName(options, field, mode));
   } else {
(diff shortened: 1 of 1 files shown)


xfxyjwf:
@haberman @acozzette

acozzette:
Sorry about the delay--somehow this slipped through my inbox.

I had a couple initial thoughts:
 - We don't actually support proto3 JSON for Javascript yet. I think what `toObject()` returns might be similar to proto3 JSON but it's not actually the same. So to me it seems logical that the `toObject()` behavior here should be the same for proto2 and proto3.
 - I would worry that this might cause backward-compatibility problems with existing code that is expecting to get base64-encoded data.

@haberman, what do you think?

seishun:
>We don't actually support proto3 JSON for Javascript yet. I think what toObject() returns might be similar to proto3 JSON but it's not actually the same. So to me it seems logical that the toObject() behavior here should be the same for proto2 and proto3.

So you think it would be fine to use `Uint8Array` in proto3 as well?

>I would worry that this might cause backward-compatibility problems with existing code that is expecting to get base64-encoded data.

Right. We could use `BYTES_DEFAULT` instead of `BYTES_U8` here - that way, `toObject` would use the default getter instead of coercing to `Uint8Array`, so if the code expects base64 from `obj.getBytesField()`, then it would still get base64 in the `toObject` output. But if there is a lot of code that expects `Uint8Array` from the getter but base64 in the object, then perhaps would could make this optional.

PR #2297, opened 2016-10-27
Adds pushLimit: and popLimit: into GPBCodedInputStream

sergiocampama:


objectivec/GPBCodedInputStream.h
```diff
@@ -218,6 +218,26 @@ CF_EXTERN_C_END
 - (size_t)position;
 
 /**
+ * Pushes (rolls forward) the current limit by the specified amount of bytes.
```
thomasvl: Let's drop the "rolls forward" part, because push is just from a stack POV.  In most cases, the new limit is less than the old limit.

thomasvl: Probably the same tweak on pop.

sergiocampama: done

PR #2298, opened 2016-10-28
total_bytes_limit_ with size when size is bigger 64MB

apollo-time:
Can't parse GraphDef from the given flat array that size is bigger than
64MB.

### DIFF
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -1349,6 +1349,8 @@ inline CodedInputStream::CodedInputStream(const uint8* buffer, int size)
     extension_factory_(NULL) {
   // Note that setting current_limit_ == size is important to prevent some
   // code paths from trying to access input_ and segfaulting.
+  if(total_bytes_limit_ < size)
+  	  total_bytes_limit_ = size;
 }
 
 inline bool CodedInputStream::IsFlat() const {


apollo-time:
I signed it!

src/google/protobuf/io/coded_stream.h
```diff
@@ -1349,6 +1349,8 @@ inline CodedInputStream::CodedInputStream(const uint8* buffer, int size)
     extension_factory_(NULL) {
   // Note that setting current_limit_ == size is important to prevent some
   // code paths from trying to access input_ and segfaulting.
+  if(total_bytes_limit_ < size)
```
xfxyjwf: How about changing the line in the initializer list instead? I.e., change:
```cpp
total_bytes_limit(kDefaultTotalBytesLimit),
```
to
```cpp
total_bytes_limit(max(size, kDefaultTotalBytesLimit)),
```

apollo-time: It's good idea. Thanks.

apollo-time: And I think there is same problem for CodedInputStream(ZeroCopyInputStream* input) constructor.

xfxyjwf:
I have changed the default size limit in both C++/Java to INT_MAX so you should be able to parse >64MB data just fine without any change.

PR #2299, opened 2016-10-28
Fix for construction of messages in the C++ Python implementation.

peterebden:
Fixes the case where a message has a map field and the caller passes in a map container (e.g. from another message object) as a constructor argument. Currently the field is silently left empty.

### DIFF
--- a/python/google/protobuf/internal/message_test.py
+++ b/python/google/protobuf/internal/message_test.py
@@ -1766,12 +1766,6 @@ class Proto3Test(BaseTestCase):
     msg.map_int32_foreign_message[19].c = 128
     self.assertEqual(msg.ByteSize(), size + 1)
 
-  def testMapFieldConstruction(self):
-    msg1 = map_unittest_pb2.TestMap()
-    msg1.map_string_foreign_message['test'].c = 42
-    msg2 = map_unittest_pb2.TestMap(map_string_foreign_message=msg1.map_string_foreign_message)
-    self.assertEqual(42, msg2.map_string_foreign_message['test'].c)
-
   def testMergeFrom(self):
     msg = map_unittest_pb2.TestMap()
     msg.map_int32_int32[12] = 34
@@ -2122,6 +2116,19 @@ class Proto3Test(BaseTestCase):
     msg = map_unittest_pb2.TestMap(
         map_int32_foreign_message={3: unittest_pb2.ForeignMessage(c=5)})
     self.assertEqual(5, msg.map_int32_foreign_message[3].c)
+
+  def testMapScalarFieldConstruction(self):
+    msg1 = map_unittest_pb2.TestMap()
+    msg1.map_int32_int32[1] = 42
+    msg2 = map_unittest_pb2.TestMap(map_int32_int32=msg1.map_int32_int32)
+    self.assertEqual(42, msg2.map_int32_int32[1])
+
+  def testMapMessageFieldConstruction(self):
+    msg1 = map_unittest_pb2.TestMap()
+    msg1.map_string_foreign_message['test'].c = 42
+    msg2 = map_unittest_pb2.TestMap(
+      map_string_foreign_message=msg1.map_string_foreign_message)
+    self.assertEqual(42, msg2.map_string_foreign_message['test'].c)
 
   def testMapValidAfterFieldCleared(self):
     # Map needs to work even if field is cleared.
--- a/python/google/protobuf/internal/message_test.py
+++ b/python/google/protobuf/internal/message_test.py
@@ -1766,6 +1766,12 @@ class Proto3Test(BaseTestCase):
     msg.map_int32_foreign_message[19].c = 128
     self.assertEqual(msg.ByteSize(), size + 1)
 
+  def testMapFieldConstruction(self):
+    msg1 = map_unittest_pb2.TestMap()
+    msg1.map_string_foreign_message['test'].c = 42
+    msg2 = map_unittest_pb2.TestMap(map_string_foreign_message=msg1.map_string_foreign_message)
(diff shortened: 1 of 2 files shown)


peterebden:
I'm pretty certain the failure on Travis is unrelated to this change; please let me know if there's anything I should do to try to rerun it or whatever.

anandolee:
Sorry we missed your PR.
I think the c extension map has changed a lot.  Can you check the current version still has same problem?

peterebden:
Thanks, I've rebased this on master and the relevant section of code looks fairly unchanged, so I assume this fix is still relevant.

python/google/protobuf/internal/message_test.py
```diff
@@ -1638,6 +1638,12 @@ def testMapByteSize(self):
     msg.map_int32_foreign_message[19].c = 128
     self.assertEqual(msg.ByteSize(), size + 1)
 
+  def testMapFieldConstruction(self):
+    msg1 = map_unittest_pb2.TestMap()
+    msg1.map_string_foreign_message['test'].c = 42
+    msg2 = map_unittest_pb2.TestMap(map_string_foreign_message=msg1.map_string_foreign_message)
```
anandolee: Please print this into two lines. Our style guide ask for 80 charactors at most for each line.

python/google/protobuf/internal/message_test.py
```diff
@@ -1638,6 +1638,12 @@ def testMapByteSize(self):
     msg.map_int32_foreign_message[19].c = 128
     self.assertEqual(msg.ByteSize(), size + 1)
 
+  def testMapFieldConstruction(self):
```
anandolee: Put this test under testMapConstruction, and need to add tests for both message map and scalar map
https://github.com/google/protobuf/blob/ed4321d1cb33199984118d801956822842771e7e/python/google/protobuf/internal/message_test.py#L1963

python/google/protobuf/internal/message_test.py
anandolee: Do we have similar support for repeated field? I didn't see such test for repeated field
https://github.com/google/protobuf/blob/ed4321d1cb33199984118d801956822842771e7e/python/google/protobuf/internal/reflection_test.py#L150

python/google/protobuf/internal/message_test.py
```diff
@@ -1638,6 +1638,12 @@ def testMapByteSize(self):
...
+    msg1 = map_unittest_pb2.TestMap()
+    msg1.map_string_foreign_message['test'].c = 42
+    msg2 = map_unittest_pb2.TestMap(map_string_foreign_message=msg1.map_string_foreign_message)
+    self.assertEqual(42, msg2.map_string_foreign_message['test'].c)
```
anandolee: Does this test work well? Is it a behavior diffs between pure python and c extension?

anandolee:
friendly ping?

python/google/protobuf/internal/message_test.py
peterebden: We don't, map fields can't be repeated?

Sorry about the line length, our limit at work is 100 chars and I've gotten used to that :)

python/google/protobuf/internal/message_test.py
peterebden: Done

python/google/protobuf/internal/message_test.py
peterebden: Yes, this is a minimal equivalent to the case we found when trying to migrate our codebase to the C++ implementation - it was fine in the pure-Python version but the field was not initialised correctly with the C++ version.

python/google/protobuf/pyext/message.cc
```diff
@@ -1180,20 +1180,41 @@ int InitAttributes(CMessage* self, PyObject* args, PyObject* kwargs) {
...
+                function_return.get(), "MergeFrom", "O", map_value));
+            if (ok.get() == NULL) {
+              return -1;
+            }
           }
-          ScopedPyObjectPtr ok(PyObject_CallMethod(
-              function_return.get(), "MergeFrom", "O", map_value));
-          if (ok.get() == NULL) {
+        } else {
+          ScopedPyObjectPtr iter(PyObject_GetIter(value));
```
anandolee: Can PyDict reuse this code?

python/google/protobuf/internal/message_test.py
```diff
@@ -1969,6 +1969,19 @@ def testMapConstruction(self):
...
+    msg1.map_int32_int32[1] = 42
+    msg2 = map_unittest_pb2.TestMap(map_int32_int32=msg1.map_int32_int32)
+    self.assertEqual(42, msg2.map_int32_int32[1])
+
+  def testMapMessageFieldConstruction(self):
+    msg1 = map_unittest_pb2.TestMap()
+    msg1.map_string_foreign_message['test'].c = 42
+    msg2 = map_unittest_pb2.TestMap(
+      map_string_foreign_message=msg1.map_string_foreign_message)
+    self.assertEqual(42, msg2.map_string_foreign_message['test'].c)
```
anandolee: Also add test to check if the error messages are the same when assign a none iterable to map: 
msg2 = map_unittest_pb2.TestMap(
￼       map_string_foreign_message=1)

peterebden: Done!

python/google/protobuf/pyext/message.cc
peterebden: I think so; was trying to avoid changing existing behaviour too much, but it should be basically equivalent.

I suppose I should just trust the tests :)

peterebden:
Thanks for the review!

PR #2301, opened 2016-10-28
[arm/gcc] Don't rely on KUSER_HELPERS feature for atomics

jbrianceau:
ARM specific CONFIG_KUSER_HELPERS feature can be disabled in Linux kernel,
and in this case, we shouldn't crash. Use gcc built-in functions instead
of arm specific code if they are available.


xfxyjwf:
Thanks for the patch! Have you tested the change with both gcc compilers >= 4.7 and <4.7?

jbrianceau:
We have tested the change with gcc compiler >= 4.7 and it works fine, i.e. when **KUSER_HELPERS** kernel feature is disabled, we do not crash anymore.
The change fallbacks with the current implementation so we didn't check with gcc compiler < 4.7

jbrianceau:
Do you need further information about this pull request @xfxyjwf ?

xfxyjwf:
No, this PR seems good to me. Merging now.

PR #2302, opened 2016-10-28
generic atomicops: Use strong compare_exchange

jbrianceau:
Weak compare-and-exchange are allowed to fail spuriously, so we have
to use the strong variation here.


jbrianceau:
Please note that current **v8** master branch uses strong compare-and-exchange in its atomicops implementation : [src/base/atomicops_internals_portable.h](https://chromium.googlesource.com/v8/v8.git/+/f7f591ab6b0061abc302120ad3040e51e2c43852/src/base/atomicops_internals_portable.h)

pherl:
LGTM. Thanks for the change.

PR #2306, opened 2016-10-30
[PERF]Improved performance of RepeatedField deserialization

mkosieradzki:
Because in case of packed encoding we know the number of records we can resize array only once instead of log(count) times.

This has major impact on another perf work I intend to do (Allowing use of structs instead of classes - in that case array of struct reallocation is way more costly).

### DIFF
--- a/csharp/src/Google.Protobuf/Collections/RepeatedField.cs
+++ b/csharp/src/Google.Protobuf/Collections/RepeatedField.cs
@@ -101,10 +101,11 @@ namespace Google.Protobuf.Collections
                 int length = input.ReadLength();
                 if (length > 0)
                 {
+                    EnsureSize(count + length);
                     int oldLimit = input.PushLimit(length);
                     while (!input.ReachedLimit)
                     {
-                        Add(reader(input));
+                        array[count++] = reader(input);
                     }
                     input.PopLimit(oldLimit);
                 }


mkosieradzki:
I signed it!

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -101,10 +101,11 @@ public void AddEntriesFrom(CodedInputStream input, FieldCodec<T> codec)
                 int length = input.ReadLength();
                 if (length > 0)
                 {
+                    EnsureSize(count + length);
```
xfxyjwf: length doesn't mean the number of elements, but the total byte length of these elements. Depending on the field type, the number of elements can vary from between length/10 (a varint takes at most 10 bytes) to length. I believe this call here will allocate more memory than it needs to be most of the time (double or triple at least).

mkosieradzki: Thanks a lot for your review. You are definitely right. It was because of my misundesrtanding of the pack format. Any performance improvement at this point would require some guessing or approximation and does not seem to be feasible..

This also renders my idea of allowing using structs instead of classes useless because of high cost of reallocations. That's disapointing. I wanted to allocate a contiguous array of structs at this point what should be way more GC friendly in many common scenarios.

I will stick to the new no-GC region feature - if used correctly it should give comparable results.

However, I have different idea of optimization in scenarios where protobuf generated parser generates temp representation of data - hybrid approach allowing using streaming for some repeated fields (decorated with some options).

xfxyjwf: You can still do this optimization for fixed-width field types such as "fixed32", "fixed64", "float", "double". For those you know for sure the number of elements is length/4 or length/8. I don't really know much about the C# implementation yet, but if FieldCodec has some methods to indicate whether a field is fixed-length or not, it should be easy to add an optimized code path for fixed-length field types here.

And for varint, it might not hurt to make an guess on the average length of a single varint, and use it to call EnsureSize(). We need more data for this guess work though.

mkosieradzki: While I was thinking about adding "streaming support" I found a nice solution for this reallocation issue.
It requires modifying codegen for C# by enabling use of partial method (as i intended for streaming support).
```
        public void MergeFrom(pb::CodedInputStream input)
        {
            uint tag;
            while ((tag = input.ReadTag()) != 0)
            {
                switch (tag)
                {
                    default:
                        input.SkipLastField();
                        break;
[...]
                    case 18:
                        {
//  We can insert partial method call here -------->
                            objects_.AddEntriesFrom(input, _repeated_objects_codec);
                            break;
                        }
                }
            }
        }
```

I intended to use this partial method to handle objects decoded from stream. Deserialize straight into destination structures instead of temporal protobuf representation.

But I have just realized that when developer wants to do a small size-to-performance tradeoff. He/she can insert an additional int32 field for an array length hint (a bit redundant) but it would allow to preallocate proper size of buffer here. Of course for security reasons it should not be treated authoritatively (or have a builtin support!) however it can be a nice preallocation hint.

This field would be:
- optional
- will work with both packed and unpacked formats
- will not require any special support from proto3, even not declaring custom options
- could be utilized on per repeated INSTANCE (not even type) basis (e.g. only if there is a significant number of repetitions)

xfxyjwf: I know nothing about C# partial message, but the ability to customize parsing/serializing routines is what I would be afraid to offer. Such customization comes with a close tie with the implementation details and makes it hard or impossible for us to make changes later. It also causes interoperability problems because other implementations (official implementations in other languages and third-party ones) will not be able to recognize such payload bytes with custom content.

mkosieradzki: Keeping this in mind, **IF** I will have good benchmark results, on real life applications, I will try to create a complete proposal backed by proper analysis and test-cases. I've noticed there are some potential areas to improve, however I am not sure if those improvements would be significant enough.

PR #2307, opened 2016-10-31
Docs: Link to Apple's Swift plugin and runtime for protobufs

sinzianag:
The project is still in prerelease but it's worth knowing about and taking into consideration.


sinzianag:
I signed the CLA!

PR #2309, opened 2016-10-31
Add note about extension use and Coding support.

thomasvl:


objectivec/GPBMessage.h
```diff
@@ -65,9 +65,17 @@ CF_EXTERN_C_END
 
 /**
  * Base class that each generated message subclasses from.
+ *
+ * @note While the class support NSSecureCoding, if the message has any
+ * extensions, they will end up reloaded in the @c unknownFields as there is
```
sergiocampama: remove "the" before unknownFields

sergiocampama: also align the left spacing to the first letter after @note in the first line... we already do this in other files..

PR #2310, opened 2016-10-31
Use require_relative in Ruby for imported definitions

tcwitte:
A 'require' won't work because the file is not on the $LOADPATH.

For example, with these definitions in a.proto and b.proto:
syntax = "proto3";
import "b.proto";
message A {
	repeated B b = 1;
}

syntax = "proto3";
message B {
	string text = 1;
}

When did "load 'a_pb.rb'" I used to get:
a_pb.rb:6:in `require': cannot load such file -- b_pb (LoadError)

### DIFF
--- a/src/google/protobuf/compiler/ruby/ruby_generator.cc
+++ b/src/google/protobuf/compiler/ruby/ruby_generator.cc
@@ -429,7 +429,7 @@ bool MaybeEmitDependency(const FileDescriptor* import,
     return true;
   } else {
     printer->Print(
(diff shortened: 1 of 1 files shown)


tcwitte:
I signed the CLA.

xfxyjwf:
I can expand on it more later, but I think supporting relative import is not the right solution. Similar requests have been made in Python as well, but all such issues occur because the user does not know how to use protobuf imports and protoc -I flag properly (many users mistakenly think it's the same thing as a C++ include directory). This needs better documentation to address.

PR #2317, opened 2016-11-02
Low-level support for custom options in C#

### LINKED ISSUE #2143: Support custom options in C#
Currently there's no way of accessing options metadata in the C# API.

jskeet:
Not ready for submission yet - more tests required.
Creating as PR for feedback before going any further.

Will eventually resolve #2143.

// cc @gabm.


csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -112,5 +112,10 @@ public EnumValueDescriptor FindValueByName(string name)
         {
             return File.DescriptorPool.FindSymbol<EnumValueDescriptor>(FullName + "." + name);
         }
+
+        /// <summary>
+        /// The (possibly empty) set of custom options for this enum.
+        /// </summary>
+        public CustomOptions CustomOptions => Proto.Options?.CustomOptions ?? CustomOptions.Empty;
```
anandolee: Can we add Options() method in DescriptorBase, and use extensions to get the the custom options?

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -466,10 +466,24 @@ void MessageGenerator::GenerateMergingMethods(io::Printer* printer) {
     "  switch(tag) {\n");
   printer->Indent();
   printer->Indent();
-  printer->Print(
-    "default:\n"
-    "  input.SkipLastField();\n" // We're not storing the data, but we still need to consume it.
-    "  break;\n");
+  // Some descriptor messages need to store unknown fields so that options can be parsed later.
+  // We don't decide which messages need that within protoc - we use a partial method, and
```
anandolee: I think its better to decide which messages need to store unknown fields in protoc. It's a waste of code for descriptors other than options. Though only for descriptor.proto file, but worth to do so to make the code more clear.

jskeet: That would make sense. Will take a while to rework this - am at a conference at the moment. Will look next week.

csharp/src/Google.Protobuf/Reflection/DescriptorPartials.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    // Partial classes for implementing options. Using a base class would be nice, but that would only
+    // remove the duplication for the property declaration; the partial methods have to be in the actual classes.
+
+    internal partial class FieldOptions
+    {
+        internal CustomOptions CustomOptions { get; private set; } = CustomOptions.Empty;
+
+        partial void MaybeReadUnknownData(ref bool skipField, CodedInputStream input) =>
+            CustomOptions = CustomOptions.MaybeReadUnknownData(ref skipField, input);
```
anandolee: How about remove this file and put "CustomOptions = ..." in code gen?

jskeet: Yes, exactly - that's what I'll do next week. I expect the C# code in the support library to stay the same though, so it's still worth reviewing that.

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
```diff
@@ -0,0 +1,316 @@
...
+
+        private CustomOptions() { }
+
+        /// <summary>
+        /// Retrieves a signed 32-bit integer value for the specified option field.
+        /// </summary>
+        /// <param name="field">The field to fetch the value for.</param>
+        /// <param name="value">The output variable to populate.</param>
+        /// <returns><c>true</c> if a suitable value for the field was found; <c>false</c> otherwise.</returns>
+        public bool TryGetInt32(int field, out int value)
```
anandolee: Other languages use templates, so users can get the options by GetExtension(). Is it possible to use template too? We promised to users the API of protobuf will be stable after 3.0.0, so we should think the APIs carefully.

jskeet: If you mean using generics, I wouldn't want to here... there are really only certain types that are feasible. We have generics for the message type, but I think it makes sense to stick to specific types for other things - we don't want to allow `TryGet<Guid>` for example.

For singular options, I think this is fine - and we *may* write code-gen to generate extra code per option anyway, which would "know" about both the type and the field number. But that can be added without changing this API at all (and indeed I'd expect it to use this).

anandolee: Yes I mean generics, I know there's some difference between C# generics and C++ template so make sense to me if you think it is better to stick to specific types.

Generate extra code per option to know the type and field number sounds good, so users can use field name instead of field number.

jskeet: Right - but that can come later. This much unblocks users immediately, giving us more time to get the codegen right. We should think about repeated options, but I think that can come a bit later too.

jskeet:
Rewritten to generate the appropriate code just for options, rather than using partial methods. The `CustomOptions` method now skips inappropriate wire formats instead of just returning that they should be skipped. Public API is the same as previous iteration.

PTAL @anandolee

csharp/protos/unittest_custom_options_proto3.proto
```diff
@@ -0,0 +1,338 @@
...
+}
+
+extend google.protobuf.FieldOptions {
+  fixed64 field_opt1 = 7740936;
+}
+
+// See https://github.com/google/protobuf/issues/2316
+// extend google.protobuf.OneofOptions {
+//   int32 oneof_opt1 = 7740111;
+// }
```
anandolee: Just want to make sure if the test proto same with C++ except oneof?

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -475,10 +481,18 @@ void MessageGenerator::GenerateMergingMethods(io::Printer* printer) {
...
+  // Option messages need to store unknown fields so that options can be parsed later.
+  if (IsDescriptorOptionMessage(descriptor_)) {
+	printer->Print(
+      "default:\n"
+      "  CustomOptions = CustomOptions.ReadOrSkipUnknownField(input);\n"
+      "  break;\n");
+  } else {
+    printer->Print(
+      "default:\n"
+	  "  input.SkipLastField();\n" // We're not storing the data, but we still need to consume it.
```
anandolee: format

csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        private void AssertOption<T>(T expected, OptionFetcher<T> fetcher, int field)
+        {
+            T actual;
+            Assert.IsTrue(fetcher(field, out actual));
+            Assert.AreEqual(expected, actual);
+        }
+    }
+}
```
anandolee: Add more tests which has defined in the proto file? Like TestMessageWithCustomOptions etc.

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
anandolee: Sounds good!
Does c++ support repeated options? Should be supported in c++ because it is an extension, but I didn't see unittest for that.

src/google/protobuf/compiler/csharp/csharp_message.cc
jskeet: Done. (Tabs are annoying...)

csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
jskeet: Added a test for every possible option (other than oneof) and a couple for aggregate options.
I'm sure more tests *could* be added, but I don't think they'd test anything extra at this stage.
Adding the tests *has* found a bug though - I haven't yet fixed it; looking into it... (The value -9876543210 is being read as 19753086419 and I don't know why.)

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
jskeet: I don't see any tests for that either, but I protoc allows it IIRC. The lack of any examples is one reason I don't think it's too crucial to do right now :) (In particular, I don't think it works for messages, because if an option is specified multiple times that basically means "Merge each occurrence into a single message.")

jskeet: Have replied to comments, but not pushed the new version yet as I'm still tracking down a bug. I may well add tests for the min/max values as well...

jskeet:
Looks like the problem is with zigzag encoding, which I'm not currently handling properly. Will need to change the CustomOptions class to mirror CodedInputStream, with all the different protobuf types (SInt32 etc). Will probably do that Friday. Definitely shows the benefits of testing - thanks!

jskeet:
Okay, pushed a second commit with requested changes. I've left it as a second commit at the moment rather than amending the first so that the change is obvious - I'll squash before merging.

Good job you requested more tests, so I was able to find the bug which suggested the API change. We dodged a bullet there :) This has also given me thoughts about what the generated options should look like when we get there - I can share that in a "not for submission" PR later...

PTAL.

anandolee:
LGTM

jskeet:
Thanks. Found a rogue old file in Makefile.am - have pushed a new single commit for the whole PR; will merge on green.

PR #2318, opened 2016-11-02
Fix compiler warnings when running :protobuf_test

mavrukin:
When running >bazel test :protobuf_test a few compiler warnings came up.  They were centered around the usage of "int" as the loop counter where the value should have been a size_t.


mavrukin:
I signed it!

Issue #2278, opened 2016-10-23
Version 3.1(release) - Incompatible serialization: C# -> JSON -> C++ 


zhenkas:
C# default JSON name is camel style and set by **ToCamelCase()** function, however C++ default JSON name is set by **ToJsonName()**.

This broke the compatibility between C# and C++ JSON serialization. If field in proto is not in Camel style (for example: "KVStore"), then there will be an error in C++ deserialize.

jskeet:
The C# formatter already uses the json_name field from the proto descriptor if it's present. Can you give a short but complete example including the expected and actual output?

zhenkas:
The problem occurs when json name is not specified and it uses "default
json name". Try to serialize proto struct with field named "KVStore". In c#
it will be converted to camel style "kvStore" while in c++ it will stay
with right case.

On Nov 2, 2016 9:42 AM, "Jon Skeet" <notifications@github.com> wrote:

The C# formatter already uses the json_name field from the proto descriptor
if it's present. Can you give a short but complete example including the
expected and actual output?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/google/protobuf/issues/2278#issuecomment-257793512>,
or mute
the thread

jskeet:
I was under the impression that the json_name field was always filled in by protoc, even if it wasn't specified explicitly. (I think that was the intention.)

It sounds like there may be two issues here:

- json_name not being filled in in all cases
- The casing rules not being consistent between C++ and C#

Do you know what other languages are doing with this? (That may affect which codebase needs to change to minimize breakage...)

zhenkas:
Nice catch, I did not think about protoc at all!

I agree that json_ name always should be filled in by protoc. Because, once
the code was generated, nothing should affect its serialization rules.

Fixing the first issue will "eliminate" the second issue for all languages.

xfxyjwf:
Every JSON implementation must implement its own snake_case to/from camelCase conversion code for FieldMask, so having protoc always populate json_name doesn't solve all the problem. Though I do have a pending change that does it but there are hundreds of failing tests to fix and I don't think it's worthwhile to pursue for no benefit.

I have updated the snake_case to camelCase conversion code to be consistent in C++ and Java. Other languages will need to copy the same conversion code as in:
https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.cc#L201
[...]

PR #2322, opened 2016-11-03
Change JSON field name formatting

jskeet:
This affects cases with leading capital letters.

This breaks compatibility with previous C# releases, but
fixes compatibility with other implementations.

See #2278 for details.


jskeet:
Two [conformance tests](https://github.com/google/protobuf/blob/master/conformance/conformance_test.cc) - I believe they need updating as they no longer match the desired functionality:

- `_field_name3` is expected to be converted to `fieldName3`, when following the new code it would be `FieldName3`
-  `__field_name13` is expected to be converted to `fieldName13` when it would actually be `FieldName13`

Is there a plan to fix these conformance tests? I believe they've been missed because both Java and C++ have such a long list of ignored tests already.

xfxyjwf:
LGTM

Don't worry about the conformance tests. I have already fixed them in google3. The next sync will fix them on github.

PR #2323, opened 2016-11-03
More Ruby-eqsue interface

marcinwyszynski:
There are two changes in this PR.

First change implements `respond_to_missing?` which is expected by a ton of Ruby APIs, and is a [good thing](https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding) to do. The implementation shares a ton of code with `method_missing` and I briefly thought about factoring it out to a shared function but I decided against it since it would create an abstraction that is hard to follow.

Second change makes `==` a bit more lenient. The comparison will now return `false` instead of failing if a message is compared to something with a different type. This is how Ruby APIs generally work:

```
$ irb
irb(main):001:0> Time.now == nil
=> false
irb(main):002:0> 7 == nil
=> false
irb(main):003:0> 7 == "7"
=> false
irb(main):004:0> Time.now == "bacon"
=> false
```

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -178,6 +178,45 @@ VALUE Message_method_missing(int argc, VALUE* argv, VALUE _self) {
   }
 }
 
+VALUE Message_respond_to_missing(int argc, VALUE* argv, VALUE _self) {
+  MessageHeader* self;
+  VALUE method_name, method_str;
+  char* name;
+  size_t name_len;
+  bool setter;
+  const upb_oneofdef* o;
+  const upb_fielddef* f;
+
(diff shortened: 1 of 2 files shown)


haberman:
ok to test

haberman:
Sorry for the delay. This change looks good, but please fix this warning before I merge it:

```
../../../../ext/google/protobuf_c/message.c: In function 'Message_eq':
../../../../ext/google/protobuf_c/message.c:348:3: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
   MessageHeader* self;
   ^
```

marcinwyszynski:
@haberman - fair point, I missed that. Fixed now, PTAL.

marcinwyszynski:
@haberman any chance this change can be released on rubygems some time soon? A library I want to publish depends on this behavior.

Issue #2326, opened 2016-11-04
warning: unused parameter 'size' in  arena_free at include/google/protobuf/arena.h:82


zd-double:
my gcc version is 4.8.5, 'unused parameter size' warning is caused by   inline void arena_free(void* object, size_t size)

This is on git master branch

acozzette:
This is now fixed with pull request https://github.com/google/protobuf/pull/2329.

PR #2329, opened 2016-11-04
Fix unused parameter warnings in arena_free

acozzette:
The size parameter is only used in cases where the compiler supports
sized delete, so when that's not available we need to specifically cast
it to void to prevent a warning.


acozzette:
@zd-double This should fix your warning from issue #2326.

acozzette:
@xfxyjwf  Thanks, Feng!

PR #2332, opened 2016-11-04
Added explicit cast to avoid size warning on Win64.

Kwizatz:
The message_lite.h file is a public file that gets pulled in by the public protobuf interface, Visual Studio 2015 projects with the flag /WX ([Threat Warnings as Errors](https://msdn.microsoft.com/en-us/library/thxezb7y.aspx)) fail compilation because of the warning generated by this line.

While a boundary check should be in place regardless of platform, the intent of the code is obvious, all this change does is silence the warning.


Kwizatz:
I signed it!

acozzette:
Thanks, @Kwizatz!

PR #2333, opened 2016-11-04
Implement RepeatedFieldIter for c extension.

### LINKED ISSUE #2324: php c extension repeated field  foreach error
i use the c extension for protobuf,the repeated field use foreach will get nothing and the for function will get the value.this is my test code:
`$repeat = new \Google\Protobuf\Internal\RepeatedField(\Google\Protobuf\Internal\GPBType::UINT32);`
`$repeat[] = 1;`
`$repeat[] = 2;`
` $repeat[] = 3;`
`echo 'foreach:';`
        `foreach ($repeat as $val)`
      `  {`
        `    echo  $val;`
    `    }`
` echo 'for:';`
   `     $count = count($repeat);`
     `   for ($i=0;$i<$count;$i++)`
    `    {`
      `      echo $repeat[$i];`
`}`
the result is :  foreach: for: 123

TeBoring:
Fix #2324


TeBoring:
@stanley-cheung

TeBoring:
The failed csharp test is not related to this PR.

PR #2348, opened 2016-11-09
Fix php c extension on 32-bit machines.

TeBoring:


TeBoring:
@stanley-cheung All related tests have passed.

PR #2353, opened 2016-11-10
MessageOptions should handle both proto2 and google.protobuf package names.

guptasu:
This is done so that the same utility code can be used irrespective of whether MessageOptions belong to package name 'proto2' or 'google.protobuf'.


guptasu:
Hi Feng,
Do you know how to fix this error ?

xfxyjwf:
The C# error is not relevant.

PR #2355, opened 2016-11-11
Speed up JSON parsing.

xfxyjwf:
It turns out calling StringOutputStream::Next()/BackUp() repeatedly is
very costly in opensource protobuf because it keeps resize() the string
back and forth. The current JSON conversion API suffers this problem and
leads to ridiculously long parsing time:
https://github.com/google/protobuf/issues/2305#issuecomment-257785492

This change fixes the problem but caching the buffer of Next() and avoid
calling BackUp() as much as possible.


xfxyjwf:
@pherl to review

Issue #2356, opened 2016-11-11
Map causes java.lang.ExceptionInInitializerError: null in J9VM


wu-sheng:
Here is my proto:
```proto
message TransactionRecord {
……
    map<string, string> parameters = 19;
……
}
```

Then, I use following code to get a builder
```java
final TransactionRecord.Builder metaDataBuilder = requestBuilder.getMetaData().toBuilder();
```

This code run fine in many JVMs, but fail in J9VM

Exception as following:
```
java.lang.ExceptionInInitializerError: null
        at java.lang.J9VMInternals.initialize(J9VMInternals.java:222) ~[na:1.6.0]
        at com.blueware.deps.com.google.protobuf.MapField.mergeFrom(MapField.java:204) ~[oneapm.jar:2.0]
        at com.blueware.monitor.collector.grpc.TransactionRecord$Builder.mergeFrom(TransactionRecord.java:1323) ~[oneapm.jar:2.0]
        at com.blueware.monitor.collector.grpc.TransactionRecord.toBuilder(TransactionRecord.java:1062) ~[oneapm.jar:2.0]
        at com.blueware.monitor.transaction.TransactionDataCompressorFactoryService.compressMetaData(TransactionDataCompressorFactoryS
ervice.java:401) ~[oneapm.jar:2.0]
        at com.blueware.monitor.transaction.TransactionDataCompressorFactoryService.compress(TransactionDataCompressorFactoryService.j
        at weblogic.deploy.internal.targetserver.BasicDeployment.activateFromServerLifecycle(BasicDeployment.java:361) [weblogic.jar:1
0.3.6.0]
        at weblogic.management.deploy.internal.ConfiguredDeployments.deployPreStandbyInternalApps(ConfiguredDeployments.java:85) [webl
ogic.jar:10.3.6.0]
        at weblogic.management.deploy.internal.DeploymentServerService.deployPreStandbyInternalApps(DeploymentServerService.java:168)
[weblogic.jar:10.3.6.0]
        at weblogic.management.deploy.internal.DeploymentPreStandbyServerService.start(DeploymentPreStandbyServerService.java:27) [web
logic.jar:10.3.6.0]
        at weblogic.t3.srvr.SubsystemRequest.run(SubsystemRequest.java:64) [weblogic.jar:10.3.6.0]
        at weblogic.work.ExecuteThread.execute(ExecuteThread.java:256) [com.bea.core.weblogic.workmanager_1.11.0.0.jar:1.11.0.0]
[...]

PR #2357, opened 2016-11-11
#2356 : fix ExceptionInInitializerError on IBM J9

D3Hunter:
this patch fixs #2356

when static field `EMPTY_MAP_FIELD` in `MapFieldLite` is initilized on IBM J9, `putAll` which is overrided and calls `ensureMutable` is called in its super class `LinkedHashMap`, that causes `UnsupportedOperationException` is thrown as `isMutable` is false at that time. 

`LinkedHashMap` in Oracle JDK doesn't call `putAll`, so it's OK.

code below comes from JD of `java.util.LinkedHashMap` in `java.util.jar` of IBM J9
```
  public LinkedHashMap(Map<? extends K, ? extends V> paramMap)
  {
    this.accessOrder = false;
    this.head = null;
    this.tail = null;
    putAll(paramMap);
  }
```
IBM J9 version info:
```
$ java -version
java version "1.6.0"
Java(TM) SE Runtime Environment (build pap3260sr9fp1-20110208_03(SR9 FP1))
IBM J9 VM (build 2.4, JRE 1.6.0 IBM J9 2.4 AIX ppc-32 jvmap3260sr9-20110203_74623 (JIT enabled, AOT enabled)
J9VM - 20110203_074623
JIT  - r9_20101028_17488ifx3
GC   - 20101027_AA)
JCL  - 20110203_01
```

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/MapFieldLite.java
+++ b/java/core/src/main/java/com/google/protobuf/MapFieldLite.java
@@ -58,7 +58,7 @@ public final class MapFieldLite<K, V> extends LinkedHashMap<K, V> {
   }
 
   @SuppressWarnings({"rawtypes", "unchecked"})
-  private static final MapFieldLite EMPTY_MAP_FIELD = new MapFieldLite(Collections.emptyMap());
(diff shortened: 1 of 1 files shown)


D3Hunter:
I signed it!

PR #2361, opened 2016-11-14
Add go_package option

pongad:
This commit brings the go_package option up-to-date with Google internal
repo.


pongad:
cc @zombiezen

@TeBoring Please take a look. This commit unblocks work on https://github.com/google/go-genproto . The change is quite small so I don't think a full release is worth it. I can work off of master until protobuf releases the next time.

xfxyjwf:
@zombiezen Do we really want inconsistent go_package for protobuf well-known types? What's the final picture here?

pongad:
Sorry about this, let's hold off on this PR for now. There are multiple discussions that we need to solve first.

pongad:
@xfxyjwf Most of the changes here are already integrated. The exception being descriptor.proto, which was landed in google's internal repo a little after everything else. Do you pull in changes from google on regular schedule? If so, I'll just close this and wait for the merge.

zombiezen:
@xfxyjwf Yes.  We eventually want to move these all to genproto, but can't until Go type aliases are available.

xfxyjwf:
@pongad We sync at least once from google3 before each release. If you need the change to land in github before the next protobuf release (3.2.0 targeted to be the end of Dec.), feel free to send me a CL including the descriptor.proto change. If otherwise you could wait for the next release.

pongad:
@xfxyjwf Thank you for your response. We're good with waiting until 3.2.0. I'll close the PR.

PR #2362, opened 2016-11-14
Class is final but declares protected field

wujingchao:
This class is declared to be final, but declares fields to be protected. Since the class is final, it can not be derived from, and the use of protected is confusing.


wujingchao:
I signed it!

acozzette:
Thanks, @wujingchao!

PR #2366, opened 2016-11-14
Fix #include in js_generator.cc

jbrianceau:
Reland commit f2885f6 that has been trampled by commit d64a2d99.


acozzette:
@jbrianceau Sorry about that previous fix getting trampled; for legacy reasons we have a script that (among other things) rewrites include paths to transform our internal codebase into the version on GitHub, and it had a bug that was causing it to rewrite that particular include wrong. I have a fix for that script and so that should prevent this fix from getting overwritten again.

jbrianceau:
No problem @acozzette, I understand.
Also, thanks for the fix :+1:

PR #2374, opened 2016-11-15
Fix jenkins tests.

xfxyjwf:
1. Listed failing csharp conformance tests.
2. Updated the maven cache in the Docker image.


xfxyjwf:
@haberman Hmm, not sure why but after fixing java/csharp, the ruby tests don't fail on jenkins any more.

haberman:
LGTM

PR #2375, opened 2016-11-16
Rename Empty to GPBEmpty in php generated file.

TeBoring:
In php, class name cannot be "Empty". Modified code generator to generate GPBEmpty instead (for google.protobuf.Empty only). Also change runtime code to work with the new generated code accordingly.


TeBoring:
@michaelbausor

TeBoring:
The failed test is not related to this pull request.

michaelbausor:
LGTM - thanks for the fix!

PR #2381, opened 2016-11-17
Ruby: assert_false -> assert.

haberman:
assert_false fails sometimes for unknown reasons.

Review to @xfxyjwf


xfxyjwf:
This fixes one jruby test failure but there is still another one:

  1) Failure:
test_respond_to(BasicTest::MessageContainerTest) [/tmp/protobuf/protobuf/ruby/tests/basic.rb:1191]:
Failed assertion, no message given.

This jruby test is failing on travis as well. Should we just disable jruby on jenkins?

PR #2390, opened 2016-11-18
Fix travis tests failures

xfxyjwf:


xfxyjwf:
@acozzette 

This fixes ruby/csharp/cpp_distcheck tests on travis.

PR #2392, opened 2016-11-19
Restore jenkins files.

xfxyjwf:
The last up-integration from github to google3 was made at commit
e28286fa05d8327fd6c5aa70cfb3be558f0932b8 but somehow the change in
that commit is not actually in the up-integration CL and as a result
that change was reverted by the down-integration. Restoring them
should make the jenkins tests running again.

Tried to rerun the up-integration script with both repos pined to the old versions but this time the jenkins changes showed up correct in the CL. Not sure where I made the mistake...


xfxyjwf:
@acozzette

acozzette:
LGTM

PR #2400, opened 2016-11-21
Fixed JSPB test failures

acozzette:


acozzette:
@xfxyjwf

js/message_test.js
```diff
@@ -1049,13 +1049,7 @@ describe('Message test suite', function() {
     var nested = new proto.jspb.test.Deeply.Nested.Message();
     nested.setCount(5);
     msg.setDeeplyNestedMessage(nested);
-
-    // After a serialization-deserialization round trip we should get back the
```
acozzette: I found that we don't currently generate `fromObject` methods (the code for it exists in js_generator.cc but is never called), so it seemed simplest to just avoid doing the round trip.

xfxyjwf: Does it still have the same test coverage? I thought the original bug can only be revealed after a round-trip. Is it not the case?

acozzette: You're right, I went ahead and added the binary serialization round trip back in.

PR #2403, opened 2016-11-22

acozzette:


xfxyjwf:
LGTM

acozzette:
@xfxyjwf I added a few more commits so do you want to take another look before I merge this? Now I believe all the tests should be passing. The CLA bot is confused but I think that shouldn't be a problem since and you and I are the only authors of commits in this pull request.

xfxyjwf:
Thanks Adam. The changes all look good to me.

acozzette:
Thanks, Feng! I'll merge this now then.

Issue #2135, opened 2016-09-19
PHP generated code missing comments


tbillington:
When generating php code from compiling the latest protoc (https://github.com/google/protobuf/commit/c44ca26fe89ed8a81d3ee475a2ccc1797141dbce), the php generated code does not have the comments from proto.

TeBoring:
This is a known issue. We will add it in our next release.

PR #2406, opened 2016-11-22
Generate phpdoc in php generated files.

TeBoring:


TeBoring:
Fixes #2135

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -759,11 +754,174 @@ void GenerateFile(const google::protobuf::FileDescriptor* file,
...
+  string result = value;
+
+  string::size_type pos = result.find_first_of('\n');
+  if (pos != string::npos) {
+    result.erase(pos);
+  }
+
+  // If line ends in an opening brace, make it "{ ... }" so it looks nice.
+  if (!result.empty() && result[result.size() - 1] == '{') {
+    result.append(" ... }");
```
michaelbausor: Maybe update test.proto to test this feature, and some of the other escaping features as well.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -759,11 +754,174 @@ void GenerateFile(const google::protobuf::FileDescriptor* file,
...
+
+  return result;
+}
+
+void GenerateMessageDocComment(io::Printer* printer,
+                               const Descriptor* message) {
+  printer->Print("/**\n");
+  GenerateDocCommentBody(printer, message);
+  printer->Print(
+    " * Protobuf type {@code ^fullname^}\n"
```
TeBoring: Modified to <code>...</code>

src/google/protobuf/compiler/php/php_generator.cc
TeBoring: Updated test.proto.
Php only supports proto3 now. In proto3, there is no case to have '{' at the end of line.

michaelbausor:
Overall this looks good to me. One thing I am unsure about is the best way to handle the code blocks. From the PSR-5 doc, this section (https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#52-description) suggests that markdown is supported. So perhaps we should use ``` instead of \<code> tags? See also this discussion: https://groups.google.com/forum/#!topic/php-fig/5odnVz3NogM

TeBoring:
The mark-down is supported for Description.
However, most of our comments just have one line, which will be regarded as Summary.
In Summary, mark-down is not supported.
https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#51-summary

michaelbausor:
Good point, in that case using code tags seems best.

PR #2407, opened 2016-11-23
Export symbols used in inline functions

jbrianceau:
fixed_address_empty_string and <classname>_default_instance_ symbols
are used in inline functions. We have to export them properly to
avoid undefined reference link errors.

### DIFF
--- a/src/google/protobuf/generated_message_util.h
+++ b/src/google/protobuf/generated_message_util.h
@@ -125,7 +125,7 @@ class ExplicitlyConstructed {
 
 // Default empty string object. Don't use this directly. Instead, call
 // GetEmptyString() to get the reference.
-extern ExplicitlyConstructed< ::std::string> fixed_address_empty_string;
+LIBPROTOBUF_EXPORT extern ExplicitlyConstructed< ::std::string> fixed_address_empty_string;
 LIBPROTOBUF_EXPORT extern ProtobufOnceType empty_string_once_init_;
 LIBPROTOBUF_EXPORT void InitEmptyString();
 


jbrianceau:
Mmm, the change works well in Chromium, but it doesn't seem to work very well here :/

jbrianceau:
It's better after having regenerated descriptor proto, the windows build is still broken though.

Maybe I shound deinline functions instead of trying to export symbol ? What do you think ?

xfxyjwf:
deinline is usually not an acceptable solution because of its performance impact. Our appveyor already covers DLL build and it's passing without this pull request. Could it be a build setup problem of Chrome?

jbrianceau:
> deinline is usually not an acceptable solution because of its performance impact.

Ok.

> Our appveyor already covers DLL build and it's passing without this pull request. Could it be a build setup problem of Chrome?

That's what I do not understand yet : why Windows version builds fine without my patch and fails with it when we use protobuf as a shared library ? (whereas Linux is fine).

For instance : 

- we have in **src/google/protobuf/generated_message_util.h** : 

```
extern ExplicitlyConstructed< ::std::string> fixed_address_empty_string;

LIBPROTOBUF_EXPORT inline const ::std::string& GetEmptyStringAlreadyInited() {
  return fixed_address_empty_string.get();
}

```
- and in **src/google/protobuf/generated_message_util.cc** : 

```
ExplicitlyConstructed< ::std::string> fixed_address_empty_string;
```

So when protobuf is a shared library, the symbol **google::protobuf::internal::fixed_address_empty_string** is hidden in it : 

```
   115: 00000000000370a0    16 OBJECT  LOCAL  HIDDEN    27 _ZN6google8protobuf8internal26fixed_address_empty_stringE
```

So when a piece of code from another module includes **generated_message_util.h** and links with protobuf shared library, there is an undefined reference to **google::protobuf::internal::fixed_address_empty_string** as the symbol is hidden.

Am I missing something here ?

thirdknife:
I am unable to build version 3.1.0 on windows with MSVS CE 2015. Getting an unresolved external symbol on:


`LNK2001	 unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > > google::protobuf::internal::fixed_address_empty_string" (?fixed_address_empty_string@internal@protobuf@google@@3V?$ExplicitlyConstructed@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@123@A)`

and 

`LNK2001	unresolved external symbol "__int64 google::protobuf::internal::empty_string_once_init_" (?empty_string_once_init_@internal@protobuf@google@@3_JA)	`

Any hints where I can look into to fix the source?

jbrianceau:
Does this pull request fix your issue @thirdknife ? Especially this commit : https://github.com/google/protobuf/pull/2407/commits/8a07630f34585ff05dc131fd3a3313634ab4401a

thirdknife:
This fails the protobuf build with the following error: 

`google::protobuf::internal::fixed_address_empty_string': redefinition; different storage class	libprotobuf
`

jbrianceau:
I made the mistake of exporting the symbol definition in the source file instead of exporting its declaration in the header file.
@xfxyjwf, I have rebased the change and hopefully fixed the issue.

jbrianceau:
Follow-up : https://github.com/google/protobuf/pull/2462

PR #2421, opened 2016-11-28
Export fixed_address_empty_string

tsondergaard:
commit 98835fb8f8 broke use of protobuf as shared library on Windows. Example
error message:

  error LNK2001: unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > > google::protobuf::internal::fixed_address_empty_string"

### DIFF
--- a/src/google/protobuf/generated_message_util.h
+++ b/src/google/protobuf/generated_message_util.h
@@ -111,7 +111,7 @@ class ExplicitlyConstructed {
 
 // Default empty string object. Don't use this directly. Instead, call
 // GetEmptyString() to get the reference.
-extern ExplicitlyConstructed< ::std::string> fixed_address_empty_string;
+LIBPROTOBUF_EXPORT extern ExplicitlyConstructed< ::std::string> fixed_address_empty_string;
 LIBPROTOBUF_EXPORT extern ProtobufOnceType empty_string_once_init_;
 LIBPROTOBUF_EXPORT void InitEmptyString();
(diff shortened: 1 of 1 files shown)


tsondergaard:
I signed it!

tsondergaard:
I don't see how my tiny patch could lead to these errors. Is it possible to trigger a new test run?

acozzette:
I believe you can just say "retest this please" to trigger a new test run.

acozzette:
@tsondergaard I think you're right that the test failures are unrelated to your change, so I'll just go ahead and merge this. Thanks for the pull request!

thirdknife:
I still get the `error LNK2001: unresolved external symbol `. I am using on windows MSVS 2015 CE.

tsondergaard:
@thirdknife, have you remembered to define PROTOBUF_USE_DLLS when building the code that links with protobuf.dll?

thirdknife:
@tsondergaard That fixed. The define should be generated from CMake?

tsondergaard:
@thirdknife, if you want to use the same headers with both static and shared library, you have to use a define that the user sets such as PROTOBUF_USE_DLLS to only enable declspec(dllimport) when the client code intends to link with the shared library (DLL). However, I think that in most cases declspec(dllimport) is optional, but apparently not here. I haven't researched why.

PR #2429, opened 2016-11-30
Add a swift_prefix file option.

thomasvl:
Swift generators should default to CamelCasing the proto package and prefixing
symbols with that, but this option allows developers to override that behavior
with something custom if they desire.

Fixes https://github.com/google/protobuf/issues/1833


pherl:
Could you also change the descriptor.proto internally to avoid conflict of field numbers in descriptor.proto?

thomasvl:
Done

PR #2435, opened 2016-12-02
Make php generated code conform to PSR-4.

### LINKED ISSUE #2267: [php] Allow individual files per class
It seems that the current `php` implementation generates multiple classes per file. Please allow an option similar to `option java_multiple_files = true;` to split each generated class out into its own file. 

This will comply with http://www.php-fig.org/psr/psr-0/ and http://www.php-fig.org/psr/psr-4/ autoloading directive that:

> The terminating class name corresponds to a file name ending in .php. The file name MUST match the case of the terminating class name.

The previous conventional protobuf implementation allowed an `option (php.multifile) = true;` directive, documented at https://drslump.github.io/Protobuf-PHP/protobuf-php.5.html which helped with this. Without this, users of this library who desire compliant conventional classloading will be forced to continue using `v3.0.0` with a different library, such as https://github.com/stanley-cheung/Protobuf-PHP.

TeBoring:
1. Generate single file for each message.
2. Lazily initiate metadata.
Fixes #2267


TeBoring:
@stanley-cheung 
This PR depends on newly added php/composer.json. The broken jenkins test needs to wait for this PR being merged, so it can see the newly added php/composer.json during the setup phase.

php/composer.json
```diff
@@ -0,0 +1,27 @@
...
+  "require": {
+    "php": ">=5.5.0"
+  },
+  "require-dev": {
+    "phpunit/phpunit": ">=4.8.0"
+  },
+  "autoload": {
+    "psr-4": {
+      "Foo\\": "tests/generated/Foo",
+      "Bar\\": "tests/generated/Bar",
```
stanley-cheung: This is really cool

TeBoring:
Thanks!

PR #2439, opened 2016-12-03
Fix generation of extending nested messages in JavaScript

### LINKED ISSUE #2438: JavaScript Extension of Nested Messages Bug
The ordering of extensions that use nested messages and the nested messages themselves appears to be broken on `3.1.0` (and also latest `master` as of writing: `4280c27`).

Test proto:
```syntax = "proto2";

message ToBeExtended {
  extensions 1 to max;
}

message MyOuterMessage {
  message MyExtension {}
  extend ToBeExtended {
    optional MyExtension my_extension = 1;
  }
}
```

Generating the JavaScript and running it through node using:
`protoc --js_out=import_style=commonjs:./ my_outer_message.proto && node my_outer_message_pb.js`

Throws:
```
TypeError: Cannot read property 'toObject' of undefined
proto.MyOuterMessage.MyExtension.toObject
my_outer_message_pb.js:291
```

This is because `proto.MyOuterMessage.MyExtension` is only defined on line `341` in the generated file.

MarcusImprobable:
Fixes #2438

Moving of the nested `GenerateClass` and `GenerateEnum` messages before `GenerateClassRegistration` ensures that extensions defined in the outer message that use the nested messages work correctly now. Previously the extension definition proceeded the message definition, causing an exception.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -819,6 +819,7 @@ js_EXTRA_DIST=                              \
   js/test3.proto                            \
   js/test4.proto                            \
   js/test5.proto                            \
+  js/test8.proto                            \
   js/test_bootstrap.js                      \
   js/testbinary.proto                       \
   js/testempty.proto
--- a/js/message_test.js
+++ b/js/message_test.js
@@ -39,6 +39,9 @@ goog.require('goog.userAgent');
 // CommonJS-LoadFromFile: google-protobuf jspb
 goog.require('jspb.Message');
 
+// CommonJS-LoadFromFile: test8_pb proto.jspb.exttest.nested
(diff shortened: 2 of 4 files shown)


marcusimprobable:
I signed it!

haberman:
Thanks for the fix. Could you include a test to verify the fix and prevent regressions?

bazel-io:

cuongtv51:
Hi all, pls Admins verify this path.

acozzette:
ok to test (this should make the bots run the CI tests)

acozzette:
@haberman It looks like the test you requested has been added and so this PR might be ready for another look.

haberman:
Thanks for the fix!

PR #2441, opened 2016-12-04
Change StringOutputStream's interface for better performance.

na-ka-na:
This fixes major performance issue with JsonStringToMessage(). Current version
of StringOutputStream is really slow outside Google. This is because
STLStringResizeUninitialized() is un-optimized (it has to zero out the
re-sized bytes). And STLStringResizeUninitialized() is invoked
repeatedly everytime a token is written.

### DIFF
--- a/src/google/protobuf/io/zero_copy_stream_impl_lite.cc
+++ b/src/google/protobuf/io/zero_copy_stream_impl_lite.cc
@@ -150,22 +150,27 @@ int64 ArrayOutputStream::ByteCount() const {
 // ===================================================================
 
 StringOutputStream::StringOutputStream(string* target)
-  : target_(target) {
+  : target_(target), byte_count_((target_ != NULL) ? target_->size() : 0) {
 }
 
 StringOutputStream::~StringOutputStream() {
+  if (target_ != NULL) {
+    target_->resize(byte_count_);
+  }
 }
 
 bool StringOutputStream::Next(void** data, int* size) {
   GOOGLE_CHECK(target_ != NULL);
-  int old_size = target_->size();
+  int old_size, new_size;
+  old_size = new_size = target_->size();
 
   // Grow the string.
   if (old_size < target_->capacity()) {
     // Resize the string to match its capacity, since we can get away
     // without a memory allocation this way.
     STLStringResizeUninitialized(target_, target_->capacity());
-  } else {
+    new_size = target_->size();
+  } else if (byte_count_ == old_size) {
     // Size has reached capacity, try to double the size.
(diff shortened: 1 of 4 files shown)


na-ka-na:
I signed it!

xfxyjwf:
ok to test

na-ka-na:
any updates here

xfxyjwf:
Sorry, I made some comments a few days ago but just found that it's not actually posted...

bazel-io:

src/google/protobuf/io/zero_copy_stream_impl_lite.cc
```diff
@@ -150,22 +150,29 @@ int64 ArrayOutputStream::ByteCount() const {
 // ===================================================================
 
 StringOutputStream::StringOutputStream(string* target)
-  : target_(target) {
+  : target_(target), byte_count_((target_ != NULL) ? target_->size() : 0) {
```
na-ka-na: Trying to emulate the existing behavior. If target is "abc" with size 3, current behavior is to start writing after "abc".

src/google/protobuf/io/zero_copy_stream_impl_lite.cc
```diff
@@ -150,22 +150,29 @@ int64 ArrayOutputStream::ByteCount() const {
...
 }
 
 StringOutputStream::~StringOutputStream() {
+  if (target_ != NULL) {
```
na-ka-na: Some places are constructing the object with nullptr and destroying them. One particular example is LazyStringOutputStream in this file itself. In fact, I had the CHECK earlier, but some protobuf lite tests were failing, so I decided not to make assumption that target would be non-null always.

src/google/protobuf/io/zero_copy_stream_impl_lite.cc
```diff
@@ -150,22 +150,29 @@ int64 ArrayOutputStream::ByteCount() const {
...
+    if (target_->size() != byte_count_) {
```
na-ka-na: Done

na-ka-na:
Done, I think I replied to the comments. Made one change of always invoking resize.

xfxyjwf:
Thanks. Before I merge this PR, I'll need to patch the change to our internal code base and check whether any project will be broken by this change. I'll post back here after I run the CL through our internal global test queue.

xfxyjwf:
Closing because this change caused many internal test failures as mentioned in https://github.com/google/protobuf/issues/2305#issuecomment-269400663

A workaround in JSON conversion code is implemented in https://github.com/google/protobuf/pull/2355 instead.

PR #2454, opened 2016-12-07
update descriptor.proto's go_package

pongad:
This change has already been made in google's repo.
Replicating the change here to unblock work on genproto.


pongad:
CC https://github.com/google/protobuf/pull/2361

There seems to be an urgent need in Google. I'm sorry @xfxyjwf , could you take another look at this?

PR #2457, opened 2016-12-07
Issue#2456 : ByteInput-based CIS (Mirror of ByteOutput-based COS)

anoopsjohn:


BSBandme:
Would you please describe the pros of adding ByteInput data structure?
Close for now as it's not on our road map and conflicts

PR #2459, opened 2016-12-07
Send all protobuf logging to logcat by default on Android

acozzette:
Currently the default for protobuf on Android is to silently drop all log
messages. This makes debugging difficult because things like GOOGLE_LOG(FATAL)
will crash the process without actually logging a useful error.

This CL changes the logging so that by default we send messages for all log
levels to logcat (and stderr). Users can override this by setting
GOOGLE_PROTOBUF_MIN_LOG_LEVEL. Also, that option was not being respected for
non-Android platforms so I went ahead and fixed that to make things consistent.


acozzette:
@xfxyjwf to review

PR #2460, opened 2016-12-07
Include -std=c++11 when compiling protobuf if available.

sergiocampama:


sergiocampama:
fyi @thomasvl @xfxyjwf

xfxyjwf:
ok to test

m4/ax_cxx_compile_stdcxx.m4
```diff
@@ -0,0 +1,982 @@
+# ===========================================================================
```
thomasvl: Do you need to pull this in, or can the AC_PREREQ(2.59) line be raised to ensure this support is present?

sergiocampama: I'd love to, but I can't seem to find the version that makes it work, or the actual name of the macro...  :(

xfxyjwf:
LGTM

PR #2461, opened 2016-12-08
Add missing include in embed.cc

jbrianceau:
This changes fixes "use of undeclared identifier 'assert'"
compilation error.


jbrianceau:
Note; the CI error seems unrelated to my change

PR #2463, opened 2016-12-08
Adding toolchain deps for proto_gen rule

nlopezgi:
The proto_gen rule runs with incomplete dependencies when using a custom toolchain (via the use of Bazel's --crosstool_top flag). The rule currently executes the protoc binary without including toolchain dependencies. Since the protoc binary is dynamically linked, including the toolchain as a dependency to its execution is necessary so that it is listed explicitly in the action's inputs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -171,3 +171,8 @@ js/testproto_libs2.js
 
 # Ignore the bazel symlinks
 /bazel-*
+
+# ruby test output
+ruby/lib/
+ruby/tests/generated_code_pb.rb
+ruby/tests/test_import_pb.rb
--- a/Makefile.am
+++ b/Makefile.am
@@ -862,6 +862,7 @@ ruby_EXTRA_DIST=                                                             \
   ruby/src/main/java/google/ProtobufJavaService.java                         \
   ruby/src/main/sentinel.proto                                               \
   ruby/tests/basic.rb                                                        \
+  ruby/tests/gc_test.rb                                                      \
   ruby/tests/repeated_field_test.rb                                          \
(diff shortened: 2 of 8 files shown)


xfxyjwf:
@lberki Can you help review this change?

lberki:
(Sorry for the late reply; I've been travelling and I had a lot of interesting days with spotty network coverage)

@cgrushko is probably a better person to review this, since he's working with protos of all kinds.

bazel-io:

cgrushko:
@lberki this isn't specific to protos. IIUC, the proto-compiler (a C++ binary) is missing a runtime dependency (I think?). 
How do we usually solve this problem? I'd expect the rule that provides the C++ binary to have enough information to execute it?
In other words, I don't think a proto rule should care about how to execute the proto-compiler (which could be Java or Go).

WDYT?

lberki:
We usually do one of the following two things
 * We put the required libraries into the runfiles tree of the binary (in this case, `protoc`) and then the runfiles should be available during the execution of the `protoc` action. 
 * We just assume that the libraries are available on the host system 

My guess (without having seen the actual breakage) is that either the toolchain doesn't specify the right `dynamic_runtime_libs` attribute, the `CROSSTOOL` file doesn't have the `dynamic_runtimes_filegroup` field set to the right value and thus they are not put in the runfiles tree or that the toolchain flags are "wrong" and thus the `protoc` binary does not find the runtime libraries in the runfiles tree.

(The quotes are there because said libraries are usually found by setting `RPATH` to some appropriate value in the binary, which is arguably an awful way to do it)

PR #2466, opened 2016-12-08
Improve ObjC deprecated annotation support.

thomasvl:
- Check the parent file options for deprecation when deciding to tag Messages
  and Enums as deprecated.
- Within the generated source push/pop the warning for implementing deprecated
  things around a deprecated class implementation.
- Annotate the methods generated for extension fields as deprecated.
- Add a testing .proto file that covers deprecated fields, messages, enums,
  enum values and compile it into the unittests to confirm things compile
  cleanly.
- Add a testing .proto file that uses the file level option to make everything
  deprecated and compile it into the unittests to confirm things compile
  cleanly.


thomasvl:
retest this please

PR #2471, opened 2016-12-09
Fix #include in cc files

jbrianceau:


jbrianceau:
I'd like to push this change to make Chromium presubmit check happy (https://codereview.chromium.org/2495533002/)

jbrianceau:
Could you please take a look at this PR @xfxyjwf ?

bazel-io:

xfxyjwf:
I don't quite understand the reason for these changes. Could you elaborate? Why are some of them changed to brackets while others are still double-quoted?

jbrianceau:
I'd like to upgrade protobuf version in Chromium (https://codereview.chromium.org/2495533002/).
There is a presubmit rule in Chromium that does not allow quoted includes for google/* and hence I cannot upgrade to current master branch.
Maybe @acozzette would have more information about this, as this is similar to https://github.com/google/protobuf/pull/2366

xfxyjwf:
ok, so it seems most of protobuf code uses angle-brackets to include test .pb.h files. This change seems fine to me.

jbrianceau:
Thanks !

PR #1138, opened 2016-01-13
Use require_relative instead of require in Ruby code

kblanks:
Closes #1137

PR #2477, opened 2016-12-09
Use `require_relative` rather than `require`

marcinwyszynski:
Maintainers PTAL

Local imports should not work like gem imports. Otherwise you need to add the base directory with generated Ruby files to the `$LOAD_PATH` which is suboptimal.


marcinwyszynski:
Ah, looks like this was argued against in #1138. I'll follow the suggestion in that PR instead.

PR #2478, opened 2016-12-09
Optional require_relative in Ruby generated code

marcinwyszynski:
This change allows instructing the Ruby compiler to use `require_relative` instead of `require` for imports.


TeBoring:
What's the benefit to use ```require_relative```

TeBoring:
closed for now since I don't quite understand why we need require_relative. If you still want to move forward with this PR, please show some use cases.

allenwu1973:
https://github.com/ruby-protobuf/protobuf/issues/240
You can see some use case in-depth in above link.

In my case, I've set `ruby_package` to avoid conflict with another package (MyProto to MyApp:MyProto), 
l'd like to put the generated code under `my_app/my_proto`, but the generated code wants to requrie from `my_proto`, which menas I have to replace generated code every time.

Using relative require can solve this without messing with LOAD_PATH

maxforasteiro:
Because of the eager_load made by [AnnotateModels](https://github.com/ctran/annotate_models) in a 
Rails application (issue https://github.com/ctran/annotate_models/issues/461) we can't keep the `_pb.rb` in the `/lib`. Following this issue https://github.com/ruby-protobuf/protobuf/issues/381 solution, I've tried to move the `_pb.rb` files to a custom directory and add a initializer for it, but as the require is global and not relative, I couldn't make it work without changing the `_pb.rb` file to `require_relative`. if I have some kind of flag on the compiler to generate a file with or without a relative it will help me a lot.

jvmlet:
@TeBoring , another benefit - [logstash-codec-protobuf ](https://github.com/logstash-plugins/logstash-codec-protobuf)  will work out-of-the-box. Currently, there is the [issue](https://github.com/logstash-plugins/logstash-codec-protobuf/issues/39)  that prevents loading `require`-ed proto-imported `rb` files.

Please integrate this PR.

PR #2482, opened 2016-12-11
[Ruby] fix floating point accuracy problem in Timestamp#to_f

andreaseger:
Observing the following script:

```ruby
require "bundler/inline"

gemfile true do
  source "https://rubygems.org"
  gem "google-protobuf"
end
require "google/protobuf/well_known_types"

time = Time.now
p time.to_f
p time.usec

time_pb = Google::Protobuf::Timestamp.new.tap{ |e| e.from_time(time) }
p time_pb

time2 = time_pb.to_time
p time2.to_f
p time2.usec


p time == time2
__END__
Resolving dependencies...
Using google-protobuf 3.1.0
Using bundler 1.13.6
1481419172.3675368
367536
<Google::Protobuf::Timestamp: seconds: 1481419172, nanos: 367536969>
1481419172.367537
367537
false
```

While using the `Google::Protobuf::Timestamp` class some millisecond accuracy is lost due to floating point handling in ruby.
This can be fixed by using the [quo](https://ruby-doc.org/core-2.2.0/Numeric.html#method-i-quo) method instead of a less accurate floating point division.

I adapted the tests to demonstrate the problem and changed the `to_f` method to use `quo` instead.

### DIFF
--- a/ruby/tests/well_known_types_test.rb
+++ b/ruby/tests/well_known_types_test.rb
@@ -13,10 +13,15 @@ class TestWellKnownTypes < Test::Unit::TestCase
     assert_equal Time.at(12345), ts.to_time
     assert_equal 12345, ts.to_i
 
-    ts.from_time(Time.at(123456, 654321))
+    time = Time.at(123456, 654321)
+    ts.from_time(time)
     assert_equal 123456, ts.seconds
     assert_equal 654321000, ts.nanos
-    assert_equal Time.at(123456.654321), ts.to_time
+    assert_equal time, ts.to_time
+
+    time = Time.now
+    ts.from_time(time)
+    assert_equal time.to_f, ts.to_time.to_f
   end
 
   def test_duration
--- a/ruby/tests/well_known_types_test.rb
+++ b/ruby/tests/well_known_types_test.rb
@@ -13,15 +13,18 @@ class TestWellKnownTypes < Test::Unit::TestCase
     assert_equal Time.at(12345), ts.to_time
     assert_equal 12345, ts.to_i
 
+    # millisecond accuracy
     time = Time.at(123456, 654321)
     ts.from_time(time)
     assert_equal 123456, ts.seconds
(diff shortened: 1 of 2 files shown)


andreaseger:
signed the CLA

andreaseger:
Any chance someone will have a look at this?

bazel-io:

ruby/tests/well_known_types_test.rb
```diff
@@ -13,10 +13,15 @@ def test_timestamp
...
 
-    ts.from_time(Time.at(123456, 654321))
+    time = Time.at(123456, 654321)
+    ts.from_time(time)
     assert_equal 123456, ts.seconds
     assert_equal 654321000, ts.nanos
-    assert_equal Time.at(123456.654321), ts.to_time
+    assert_equal time, ts.to_time
+
+    time = Time.now
```
haberman: I'd rather not make the test depend on the current time. We should make it deterministic. If it's important to use a time other than the `123456`/`654321` one, let's hard-code a time that exercises what we are testing.

andreaseger: IIRC I simply wanted a timestamp more accurate than milliseconds without running in some other floating point errors (which one would encounter if using i.e. `Time.at(123456, 654321.321)`)

I'll update the PR with some other way to do that.

andreaseger:
I removed the use of the current time and found an other way to ensure nanosecond accuracy.

I could even remove the middle test case as the last one implicitly includes it. Let me know if I should do that.

acozzette:
ok to test (this phrase will cause the other CI bots to run the tests)

andreaseger:
anyone looking at this again?

ok to test

(not sure what magic phrase I need to trigger)

bazel-io:

acozzette:
ok to test

@andreaseger  I'm sorry for the long delay, let me take a look at this now.

acozzette:
@andreaseger Thanks for the pull request!

PR #2484, opened 2016-12-11
Add support for Windows ARM builds

ngg:
This is based on the not merged https://github.com/google/protobuf/pull/1506 pull request.
*Internal function names are used as suggested in a comment there.
It is possible to build protobuf to the Universal Windows Platform with this.

### DIFF
--- a/src/google/protobuf/stubs/atomicops.h
+++ b/src/google/protobuf/stubs/atomicops.h
@@ -123,8 +123,8 @@ Atomic32 Barrier_AtomicIncrement(volatile Atomic32* ptr,
 // ensure that no later memory access can be reordered ahead of the operation.
 // "Release" operations ensure that no previous memory access can be reordered
 // after the operation.  "Barrier" operations have both "Acquire" and "Release"
-// semantics.   A MemoryBarrier() has "Barrier" semantics, but does no memory
-// access.
+// semantics.   A MemoryBarrierInternal() has "Barrier" semantics, but does no
+// memory access.
 Atomic32 Acquire_CompareAndSwap(volatile Atomic32* ptr,
                                 Atomic32 old_value,
                                 Atomic32 new_value);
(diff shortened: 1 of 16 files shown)


src/google/protobuf/stubs/atomicops.h
```diff
@@ -123,19 +123,16 @@ Atomic32 Barrier_AtomicIncrement(volatile Atomic32* ptr,
...
                                 Atomic32 new_value);
 Atomic32 Release_CompareAndSwap(volatile Atomic32* ptr,
                                 Atomic32 old_value,
                                 Atomic32 new_value);
 
-#if defined(__MINGW32__) && defined(MemoryBarrier)
-#undef MemoryBarrier
-#endif
-void MemoryBarrier();
+void MemoryBarrierInternal();
```
xfxyjwf: This is kind of a public symbol name and I would prefer not changing it. Does the original #undef not work for windows arm build? Could you try:
```cpp
#if (defined(__MINGW32__) || defined(__YOUR_PLATFORM_MACRO)) && defined(MemoryBarrier)
#undef MemoryBarrier
#endif
```
?
Or maybe simply:
```cpp
#ifdef MemoryBarrier
#undef MemoryBarrier
#endif
```

ngg: The problem is that we actually need that define in the windows implementation of MemoryBarrierInternal.
Using this Internal function name was suggested in the previous PR, instead of redefining MemoryBarrier to something else and then undefining it.
But I can change it to that version if you would like it better.

xfxyjwf: Can you point me to the other PR you mentioned?

I don't like redefining a system macro either... If we actually have to use the MemoryBarrier macro, renaming the protobuf to MemoryBarrierInternal seems more acceptable. Could you add a comment in the code describing why we have to use MemoryBarrierInternal?

ngg: https://github.com/google/protobuf/pull/1506

bazel-io:

xfxyjwf:
Jenkins test this please.

xfxyjwf:
Jenkins test this please

xfxyjwf:
@damienmg I'm not able to trigger the bazel test for this PR. Any ideas?

ngg:
@xfxyjwf Can you have a look at https://github.com/ngg/protobuf/commit/4a23716b37ef8692baf1b04f6fe2a233ad1d1cdb?
That is a (not yet tested) version that doesn't rename MemoryBarrier.

damienmg:
Jenkins test this please

You need to make visible the fact that you are from the Google org. See https://github.com/orgs/google/people?query=xfxyjwf

xfxyjwf:
@damienmg Thanks, I didn't notice my membership was private.

xfxyjwf:
@ngg Does the renaming macro approach really work? I guess I am fine with either as long as they work. Neither of them seems perfect.

ngg:
Actually the macro renaming trick does NOT work. It would expand to recursively call the same MemoryBarrier function again. I have seen this trick on Stack Overflow but it seems it's not standard compliant and it's not even working on MSVC.
So I propose this PR should be merged.

xfxyjwf:
@ngg Thanks.

PR #2491, opened 2016-12-12
Added conformance testing for binary primitive types.

haberman:
This is basic and more tests will be added over time.

### DIFF
--- a/conformance/conformance_test.cc
+++ b/conformance/conformance_test.cc
@@ -132,8 +132,8 @@ string fixed32(void *data) { return string(static_cast<char*>(data), 4); }
 string fixed64(void *data) { return string(static_cast<char*>(data), 8); }
 
 string delim(const string& buf) { return cat(varint(buf.size()), buf); }
-string uint32(uint32_t u32) { return fixed32(&u32); }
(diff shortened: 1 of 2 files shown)


bazel-io:

haberman:
Review to @xfxyjwf . This first CL is basic, but should be a good start. More tests to come.

haberman:
All tests are passing @xfxyjwf .

conformance/conformance_test.cc
```diff
@@ -675,6 +715,86 @@ bool ConformanceTestSuite::RunSuite(ConformanceTestRunner* runner,
     TestPrematureEOFForType(static_cast<FieldDescriptor::Type>(i));
   }
 
+  int64 kInt64Min = -9223372036854775808ULL;
```
xfxyjwf: ULL => LL?

haberman: You would think so. However the `ULL` is applied before the `-` (the `-` is not actually part of the constant). And `ABS(kInt64Min)` is too large to fit in an int64, so this throws compiler warnings. Some authors get around this by writing it as: `-9223372036854775807LL - 1`.

PR #2508, opened 2016-12-16
add a key to ctx.action dict to prevent protoc losing the default env

yliu120:
This is a small change on the protobuf.bzl but very reasonable and useful.

In practice, a few protobuf client (like tensorflow, see below and etc) would call ```_proto_gen_impl(ctx)``` indirectly to generate proto implementations. However by the current definition of bazel's ctx.action rule (http://www.bazel.io/versions/master/docs/skylark/lib/ctx.html#action), the default user's custom shell env (PATH/LD_LIBRARY_PATH) is not passed to the compiled protoc binary because the user_default_shell_env is set to be ```False``` as default. This will invoke link error cuz users may link protoc with a customized library. There are a lot of reports on this issue.

In principle, when calling ```_proto_gen_impl(ctx)``` to generate proto impls, we should pass user's env to the protoc binary.

Another small change is to delete an extra space. (a bonus created by my vim editor).

Please see some issues related to this small CL:
https://github.com/tensorflow/tensorflow/issues/3261

This change will not affect any other reasonable implementations.

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -90,6 +90,7 @@ def _proto_gen_impl(ctx):
         arguments=args + import_flags + [s.path for s in srcs],
         executable=ctx.executable.protoc,
         mnemonic="ProtoCompile",
+        use_default_shell_env=True,
     )
 
   return struct(
@@ -141,7 +142,7 @@ Args:
     compiler.
   plugin_language: the language of the generated sources
   plugin_options: a list of options to be passed to the plugin
-  gen_cc: generates C++ sources in addition to the ones from the plugin. 
+  gen_cc: generates C++ sources in addition to the ones from the plugin.
   gen_py: generates Python sources in addition to the ones from the plugin.
   outs: a list of labels of the expected outputs from the protocol compiler.
 """


bazel-io:

yliu120:
I signed it!

mitar:
This really solves a problem compiling protobuf with tensorflow with environment variables.

htuch:
+1. Can we get a merge on this? This is very useful when building with non-default toolchains and setting LD_LIBRARY_PATH.

xfxyjwf:
@cgrushko Can you help review this PR?

xfxyjwf:
@cgrushko  The change seems good to me, but given that default value of that flag is False, I'm wondering if there are concerns about making it True. Maybe bazel team has other recommended solutions for overriding CC/LD_LIBRARY_PATH?

yliu120:
@xfxyjwf Thanks for LGTM and the comment. I guess Bazel team made that flag False as default just because Bazel usually will let the user define a customized compiler toolchain in a crosstool file, so it means Bazel could setup an internal version of customized environment and always pass them to the binary command right before it (it doesn't export it to create global side effects). When this is true, then it will certainly not use the default shell 's environment. 

However, the case here is:
Some package depends on protobuf. Like tensorflow or other protobuf clients, when users compile their code from source, Bazel will pull protobuf down first and compile protobuf using a user-customized environment. It will create a protoc binary and then use that binary to compile .proto source files. Bazel will use the gen_rule defined here to compile those proto files. Bazel may have no responsibility for passing its internal customized envs to proto gen rules since they have already been defined here. (I may not be correct). So for not dropping those envs, we really should add the customized env back here.

htuch:
@yliu120 is correctly interpreting the use of this attribute - it's intended to allow the user to override. 

@xfxyjwf Can we merge this? It's something we'd like for the Envoy Bazel build, to allow it to be used internally in Google where we point `CC` at crosstools when using the OSS version from https://github.com/lyft/envoy/.

PR #2517, opened 2016-12-19
Auto-generate well_known_types_embed.cc

acozzette:
Until now this file was just checked into the repo, but actually it
should be generated from any.js, struct.js, and timestamp.js. This
change updates the build system to make this happen. To make it work I
also had to remove some C++11 features from embed.cc.


bazel-io:

cmake/tests.cmake
```diff
@@ -214,4 +214,5 @@ target_link_libraries(lite-arena-test libprotobuf-lite gmock_main)
 
 add_custom_target(check
   COMMAND tests
+  DEPENDS tests test_plugin
```
acozzette: I was wondering about that actually. Adding `tests` as a dependency did not seem to be empirically necessary to make the tests pass, but it seems like it ought to be there. If we leave it out, is CMake smart enough to determine that it should be a dependency since it appears as part of `COMMAND`?

podsvirov: I thought a little more and probably agree with your decision.
I added this command as an alias for compatibility with similar check for autotools project, but I forgot to specify dependencies.

acozzette: Ah ok, thanks. I'll just leave in the `tests` dependency then; if it turns out to be unnecessary then it won't cause any harm anyway.

PR #2519, opened 2016-12-20
unwrap descriptor class before comparison of RepeatedField types

rxbynerd:
Given the protobuf file:

```proto
message OperatorPerformanceRecord {
  // ... 

  repeated OperatorServiceGroupPerformanceRecord operator_service_group_performance_records = 10;
  repeated TimebandRecord timeband_records = 11;
}

message OperatorServiceGroupPerformanceRecord {
  string name = 1;
  // ...
}

message TimebandRecord {
  int32 timeband = 1;

  // ...
}
```

On assignment of a `repeated` field from Ruby-land, the C function `check_repeated_field_type` checks the following:

1. the function has been called in the correct circumstances (the field you're assigning to is of type 'repeated')
2. the object you're trying to assign to your field is an instance of `Google::Protobuf::RepeatedField` 
3. the primitive field type of the `RepeatedField` object matches that of the field you're assigning to
4. if the field is type 'MESSAGE' or 'ENUM', it checks the class of the `RepeatedField` against that of the field

Whilst trying to use repeated message fields, as in the example above, the 4th check doesn't work out as intended.

Because `self->field_type_class` returns the Ruby class created by a Descriptor, but `get_def_obj` returns the Descriptor object used to generate the Ruby class, comparing the two objects will never be successful.

This patch passes the descriptor returned from `get_def_obj` into `Descriptor_msgclass`, so we compare `msgclass` to `msgclass`.

--- 

My method for debugging this, using this patch on master:

```
diff --git a/ruby/ext/google/protobuf_c/storage.c b/ruby/ext/google/protobuf_c/storage.c
index 3ff2bda..d1ca38f 100644
--- a/ruby/ext/google/protobuf_c/storage.c
+++ b/ruby/ext/google/protobuf_c/storage.c
@@ -599,6 +599,15 @@ static void check_repeated_field_type(VALUE val, const upb_fielddef* field) {
       self->field_type == UPB_TYPE_ENUM) {
     if (self->field_type_class !=
         get_def_obj(upb_fielddef_subdef(field))) {
+
+      printf("[PEACH-DEBUG] type invalid - error incoming shortly...\n");
+      printf("[PEACH-DEBUG] self->field_type_class:\n");
+      rb_p(self->field_type_class);
+      printf("[PEACH-DEBUG] get_def_obj(upb_fielddef_subdef(field)):\n");
+      rb_p(get_def_obj(upb_fielddef_subdef(field)));
+      printf("[PEACH-DEBUG] attempting to ascertain correct comparison variables...\n");
+      rb_p(Descriptor_msgclass(get_def_obj(upb_fielddef_subdef(field))));
+
       rb_raise(rb_eTypeError,
                "Repeated field array has wrong message/enum class");
     }
```

Which displays the following:

screen shot 2016-12-20 at 00 27 39

---

This PR is more for discussion rather than a direct unit of work to be merged - I'm not positive this is the best way to solve the problem. I don't have a lot of experience dealing with the C backend of the protobuf library. For instance, I'm calling the hook Ruby calls instead of any internal API.

That said, this works for me and allows me to continue working on my project in parallel of discussion of a correct solution. I would also like to contribute some documentation & regression tests as well as this patch - repeated fields in Ruby are a little "sink-or-swim" at the moment!

### DIFF
--- a/ruby/ext/google/protobuf_c/storage.c
+++ b/ruby/ext/google/protobuf_c/storage.c
@@ -598,7 +598,7 @@ static void check_repeated_field_type(VALUE val, const upb_fielddef* field) {
   if (self->field_type == UPB_TYPE_MESSAGE ||
       self->field_type == UPB_TYPE_ENUM) {
     if (self->field_type_class !=
-        get_def_obj(upb_fielddef_subdef(field))) {
+        Descriptor_msgclass(get_def_obj(upb_fielddef_subdef(field)))) {
       rb_raise(rb_eTypeError,
                "Repeated field array has wrong message/enum class");
     }
--- a/ruby/ext/google/protobuf_c/storage.c
+++ b/ruby/ext/google/protobuf_c/storage.c
@@ -595,12 +595,20 @@ static void check_repeated_field_type(VALUE val, const upb_fielddef* field) {
     rb_raise(rb_eTypeError, "Repeated field array has wrong element type");
   }
 
-  if (self->field_type == UPB_TYPE_MESSAGE ||
-      self->field_type == UPB_TYPE_ENUM) {
+  if (self->field_type == UPB_TYPE_MESSAGE) { 
     if (self->field_type_class !=
         Descriptor_msgclass(get_def_obj(upb_fielddef_subdef(field)))) {
       rb_raise(rb_eTypeError,
-               "Repeated field array has wrong message/enum class");
+               "Repeated field array has wrong message class");
+    }
+  }
+
+
+  if (self->field_type == UPB_TYPE_ENUM) {
+    if (self->field_type_class !=
+        EnumDescriptor_enummodule(get_def_obj(upb_fielddef_subdef(field)))) {
+      rb_raise(rb_eTypeError,
+               "Repeated field array has wrong enum class");
     }
   }
 }
--- a/ruby/tests/repeated_field_test.rb
+++ b/ruby/tests/repeated_field_test.rb
@@ -126,6 +126,12 @@ class RepeatedFieldTest < Test::Unit::TestCase
     assert_equal false, m.repeated_string.empty?
   end
 
+  def test_reassign
+    m = TestMessage.new
+    m.repeated_msg = Google::Protobuf::RepeatedField.new(:message, TestMessage2, [TestMessage2.new(:foo => 1)])
+    assert_equal m.repeated_msg.first, TestMessage2.new(:foo => 1)
+  end
+
   def test_array_accessor
     m = TestMessage.new
     reference_arr = %w(foo bar baz)


bazel-io:

rxbynerd:
Fixed this up to handle enum cases properly - there doesn't seem to be a common interface for getting the descriptor from a msgclass/enummodule, so I've broken them out into two separate cases with less generic error messaging.

rxbynerd:
@TeBoring are you currently maintaining protobuf-ruby? If so, please can you take a look at this?

TeBoring:
@haberman

rxbynerd:
@TeBoring @haberman is there anything I can assist with to get this pull request merged? It's been sitting for a week - I also attempted to post on [the mailing list](https://groups.google.com/forum/#!topic/protobuf/RI4WFohaL_w) but I didn't get a reply there either.

Apologies if this is caused by my unfamiliarity with the process here!

haberman:
Thanks for the PR and sorry for letting it sit! Sometimes GitHub notifications get lost in my inbox.

Are you saying that assigning repeated message/enum fields is completely broken right now? That seems surprising, and certainly a very bad bug if true!

Can you include a test that fails currently but is fixed by this change?

rxbynerd:
@haberman so sorry for missing this - thought i responded!

I've written up a test harness type thing here - https://github.com/rxbynerd/prototest - should be able to clone this, bundle install, then `bin/prototest testcase` to replicate

in that repo, the [proto file is in protos/](https://github.com/rxbynerd/prototest/blob/master/protos/test.proto), [saved in lib/prototest/protos](https://github.com/rxbynerd/prototest/blob/master/lib/prototest/protos/test_pb.rb), then used in [lib/prototest/cli.rb](https://github.com/rxbynerd/prototest/blob/master/lib/prototest/cli.rb). you can use `make` to regenerate `test_pb.rb` if you need to change it.

gonna spend my evening looking at the test suite for google/protobuf to see what existing specs we have on this, and to make sure i'm using the API in the correct manner.

bazel-io:

rxbynerd:
@TeBoring @haberman please can I get some movement on this?

haberman:
Sorry again for the delay. I should have been more specific. Can you include a test in our current set of unit tests? We want the new test to be part of our regression suite.

https://github.com/google/protobuf/blob/master/ruby/tests/repeated_field_test.rb

rxbynerd:
Apologies for the misunderstanding - before I convert my example to a test, please can you confirm if [this is the way](https://github.com/rxbynerd/prototest/blob/master/lib/prototest/cli.rb#L23L25) you'd expect repeated message field assignment to work?

haberman:
> can you confirm if this is the way you'd expect repeated message field assignment to work?

I believe so. Do you get the same results when you append the members after constructing the repeated field?

rxbynerd:
I can't construct a repeating field properly, as when I assign it to the field on the message, I get the `Repeated field array has wrong message/enum class (TypeError)` error, even if the third argument of the repeating field is an empty array.

It's happy to make the `RepeatedField` object, just not happy to assign it to the field on the message.

I'll write up a test case tomorrow night, for now I can append messages to the field that's initialised when the message is created, and that seems to work fine

rxbynerd:
@haberman sorry for the delay - added a test to illustrate what this patch fixes

Please can you take a look?

rxbynerd:
👋 @haberman and @TeBoring - please can you take a look at this? apologies for my delay, looking to get this closed before this PR reaches its birthday 🙈

rxbynerd:
@haberman and @TeBoring - please can we get some movement on this?

rxbynerd:
@haberman and @TeBoring - thanks for merging this - should this have gone into `3.5.x` instead of `master`? Would really love to see this get released

rxbynerd:
I'll re-open this against the current development branch - awkward that this is in master and not in the release branch, as this doesn't seem to be in the queue for release

PR #2524, opened 2016-12-21
maps_test.js: check whether Symbol is defined before using it

acozzette:
Symbol is not yet available on older versions of Node.js and so this
test fails with them. This change just directly checks whether Symbol is
available before we try to use it.


TeBoring:
LGTM

PR #2526, opened 2016-12-21
Generating service interface with PHP generator

dParadiz:
Service interface is now generated with PHP generator. This interface can than be implemented with ClientStub.

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -71,8 +71,7 @@ void GenerateMethodDocComment(io::Printer* printer, const MethodDescriptor * met
 void Indent(io::Printer* printer);
 void Outdent(io::Printer* printer);
 void GenerateMessageDocComment(io::Printer* printer, const Descriptor* message);
-void GenerateFieldDocComment(io::Printer* printer,
-                             const FieldDescriptor* field);
+void GenerateFieldDocComment(io::Printer* printer, const FieldDescriptor* field, const string& doc_option);
 void GenerateEnumDocComment(io::Printer* printer, const EnumDescriptor* enum_);
 void GenerateEnumValueDocComment(io::Printer* printer,
                                  const EnumValueDescriptor* value);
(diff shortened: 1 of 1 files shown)


bazel-io:

dParadiz:
CLA should be fine now if email check is not case sensitive.

TeBoring:
@stanley-cheung Does this change work with grpc?

TeBoring:
Could you add an file option "php_generic_services" to control whether to generate code for service? The default value should be false.

stanley-cheung:
This shouldn't conflict with gRPC, as that was a separate plugin.

However, I am not sure how this PR is going to be used. So this prints out the service definition as an interface. Can I see an example on how the interface is going to be implemented? What's the base class, dependencies, there? Is it related to gRPC or not?

TeBoring:
@dParadiz could you make this PR focused on adding generated code for service? You can send another PR to improve the comment.

dParadiz:
Interface only describes methods which are exposed by RPC service. This ensures that by integrating client into application you are provided with correct method signature. Which helps a lot by implementation.

Interface is also used to decouple this from any concrete implementation of RPC service. So one can write it's own wrapper for gRPC extension or use existing one. Making explicit implementation of RPC service will introduce coupling with libraries used in implementation (that means code adjustments when libs change, re-implementing if they are deprecated). Are there any concrete libraries that can be used in implementation?

An example of using service interface with grpc/grpc php code. Here you will have base stub implementation and rpc method outline. All you need now is is to implement rpc methods and decide what kind of calls you are making in them.

```php
class MyClient extends Grpc\BaseStub implements ServiceInterface
{
// implementation of methods defined by rpc service
}
```

dParadiz:
I see. For use with that interface would need to have optional  parameters `$metadata` and `$options` defined in method signature.

How do you tell the generator which type of call should be used in function? 
Do you need different .proto files for server and client service generation?

stanley-cheung:
Yes in the gRPC plugin we defined optional `$metadata` and `$options` parameters in the method signature to be consistent with other languages.

I am not sure about your questions. Can you clarify? 

In fact, I am not sure if this is needed. Are you looking to be able to implement a gRPC Plugin a second way (other than one provided by gRPC)? Yes an interface is nice, but I am not sure if it's needed at this stage as we, in gRPC, are not looking to provide alternate implementation of a protoc plugin.

TeBoring:
To be consistent with other languages, we allow users to control whether proto php code generator will generate code for service by defining a file level option "php_generic_services" in proto files. If the option is true, proto php code generator will generate an interface for service same as the one in this PR. Otherwise, the code generator just generate nothing. Other plugins (e.g., grpc plugin) are free to implement it or not.
@dParadiz Does that work for you?

anandolee:
I am closing it for no response more than one year. Feel free to reopen if still like to work on this

PR #2529, opened 2016-12-21
Class deprecation support

wackoisgod:
Added the support for class level deprecation which will in turn also deprecate any fields that are currently using that type.

It was requested that support for deprecating a class in the c# code generation was added. In this implementation we deprecate the class as well as any fields that are also using that class. 

I think this ensures that users don't continue to use the class in other places, maybe some food for thought that its another option to allow for class->field level deprecation.

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_field_base.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_field_base.cc
@@ -126,6 +126,16 @@ void FieldGeneratorBase::AddDeprecatedFlag(io::Printer* printer) {
   {
     printer->Print("[global::System.ObsoleteAttribute]\n");
   }
+  else 
+  {
+    if (descriptor_->type() == FieldDescriptor::TYPE_MESSAGE) 
+    {
+      if (descriptor_->message_type()->options().deprecated())
+      {
+        printer->Print("[global::System.ObsoleteAttribute]\n");
+      }
+    }
+  }
 }
 
 void FieldGeneratorBase::AddPublicMemberAttributes(io::Printer* printer) {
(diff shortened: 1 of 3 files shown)


bazel-io:

wackoisgod:
I signed it!

acozzette:
ok to test

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -98,12 +98,21 @@ const std::vector<const FieldDescriptor*>& MessageGenerator::fields_by_number()
   return fields_by_number_;
 }
 
+void MessageGenerator::AddDeprecatedFlag(io::Printer* printer) {
+  if (descriptor_->options().deprecated())
+  {
```
acozzette: One small thing: our usual C++ style is to put the opening curly brace on the same line as the `if (...)`.

src/google/protobuf/compiler/csharp/csharp_field_base.cc
```diff
@@ -126,6 +126,16 @@ void FieldGeneratorBase::AddDeprecatedFlag(io::Printer* printer) {
   {
     printer->Print("[global::System.ObsoleteAttribute]\n");
   }
+  else
```
acozzette: To cut down on the block nesting I would suggest rewriting it like this:

```
else if (descriptor_->type() == FieldDescriptor::TYPE_MESSAGE &&
         descriptor_->message_type()->options().deprecated()) {
  printer->Print("[global::System.ObsoleteAttribute]\n");
}
```

acozzette:
@wackoisgod I made a couple of small comments but it looks great otherwise, thanks for the pull request!

wackoisgod:
@acozzette updated based on your feedback

acozzette:
Thanks, @wackoisgod! I'll wait for the Travis run to finish just to be safe and then merge this. I see Jenkins has an error but it looks like it's unrelated to this.

PR #2542, opened 2016-12-27
Fix warning in compiler/js/embed.cc

jbrianceau:
embed.cc: In function ‘std::string CEscape(const string&)’:
embed.cc:51:32: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for (int i = 0; i < str.size(); ++i) {
                                ^


bazel-io:

PR #2555, opened 2017-01-01
update license to 2017

fishercoder1534:
Happy New Year!


bazel-io:

acozzette:
Sorry for not getting back to you sooner on this, but I'm afraid we don't need this change because we are not trying to keep the year completely up to date in the license file.

PR #2562, opened 2017-01-04
Add ByteString.FromStream and ByteString.FromStreamAsync in C#

### LINKED ISSUE #2088: [C#] Provide a way to construct a ByteString from a stream
Hi, 

The C# implementation of ByteString does not provide a way to construct the object directly from a System.IO.Stream. 
In scenarios where the message contains bytes for binary data (such as images from the file system), we need to copy the data twice (Stream -> byte[] -> ByteString).

Although the Stream length is not always known, there are still many cases where it is known (such as files, request body in .net MVC applications, MemoryStream ...). 
A method like `ByteString.ReadFrom(System.IO.Stream input)` would solve this problem.

Please note that the Java implementation does have a `static readFrom(InputStream streamToDrain)` method to construct a ByteString. 

Regards,
Amine.

jskeet:
Fixes #2088.


csharp/src/Google.Protobuf.Test/project.json
```diff
@@ -27,6 +27,7 @@
   "testRunner": "nunit",
 
   "frameworks": {
+    "net451": {},
```
jtattermusch: nit: Google.Protobuf/project.json targets `net45`

jskeet: Yes, but I seem to remember there was some reason why I needed to make the tests depend on net451. I can't remember what offhand, but I'll have a look and change it if I can.

jskeet: Ah yes, this was it:

> Package dotnet-test-nunit 3.4.0-alpha-2 is not compatible with net45

Will merge as-is.

PR #2565, opened 2017-01-05
Fixed cross compilations with the Autotools build

acozzette:
Pull request #2517 caused cross compilations to start failing, because
the js_embed binary was being built to run on the target platform
instead of on the build machine. This change updates the Autotools build
to use the AX_PROG_CXX_FOR_BUILD macro to find a suitable compiler for
the build machine and always use that when building js_embed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -60,6 +60,7 @@ AC_ARG_WITH([protoc],
 # Checks for programs.
 AC_PROG_CC
 AC_PROG_CXX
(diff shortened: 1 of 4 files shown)


xfxyjwf:
Adam, this appears to break cpp_distcheck test on travis. Can you take a look?
https://travis-ci.org/google/protobuf/jobs/189334745

acozzette:
Sorry about that, I just created pull request #2567 to fix this.

PR #2568, opened 2017-01-07
Change hint type to `const void*`

matthauck:
This is both more correct, and the build fails on AIX without it

### DIFF
--- a/src/google/protobuf/map.h
+++ b/src/google/protobuf/map.h
@@ -592,7 +592,7 @@ class Map {
     MapAllocator(const MapAllocator<X>& allocator)
         : arena_(allocator.arena()) {}
 
-    pointer allocate(size_type n, const_pointer hint = 0) {
+    pointer allocate(size_type n, const void* hint = 0) {
       // If arena is not given, malloc needs to be called which doesn't
       // construct element object.
       if (arena_ == NULL) {


bazel-io:

matthauck:
Hmm. Unsure why the build failed. I couldn't figure out what failed from the build output. `make check` works for me...

matthauck:
_ping_

Is there anything missing here?

xfxyjwf:
Can you rebase this pull request on the current master?

matthauck:
Sure, will do.

matthauck:
Updated.

xfxyjwf:
ok to test

xfxyjwf:
The change looks good to me. Assigned to @TeBoring as he has some questions about the change.

TeBoring:
http://en.cppreference.com/w/cpp/memory/allocator/allocate
I was looking at the page. Was the API with const void* deprecated?

matthauck:
Hmm. It does say it is deprecated. I believe `std::allocator<void>::const_pointer` essentially ends up being a longer way to say the same thing. `void*` is at least more correct than `std::allocator<T>::const_pointer`, as it is now.

matthauck:
Looks to me that c++17 is moving to get rid of that `hint` parameter altogether.

Issue #1197, opened 2016-01-29
<src/google/protobuf/stubs/strutil.h> deliberately not listed in <cmake/extract_inlcudes.bat.in>?


KerstinKeller:
So I am just wondering, if the `strutil.h` file is not distributed with a CMake install on purpose or by accident. 
I have been writing a protoc compiler plugin which uses the `SimpleItoa` function defined in `strutil.h`   to convert numerical values to strings, only to realize that the file is not included when create an install target with CMake.
The functions of choice are either inline functions or marked with `LIBPROTOBUF_EXPORT`, which makes me think that they might have been made to be exported e.g. available from outside the library.

haberman:
On one hand, I'd be inclined to think this is an oversight:

- other files in stubs/ are listed in `cmake/extract_includes.bat.in`
- the functions in `strutil.h` are not in the `internal` namespace
- some of the functions in `strutil.h` are marked as `LIBPROTOBUF_EXPORT`, as you note.

On the other hand, `strutil.h` is also omitted from a `make install` with the automake setup.

So I'm not sure. @xfxyjwf or @pherl, can you clarify?

pherl:
Yep, I believe it's an oversight. Will fix this in the next release.

xfxyjwf:
@pherl I think that file is left out intentionally. The extract_includes.bat script installs only protobuf public header files. Some header files under stubs are left out from there because they are not part of protobuf public API and are also not included by any other public header files. strutil.h is not the only one. stringprintf.h/time.h/statusor.h and a few others are also left out for this reason.

We have an explicit list of public header files in src/Makefile.am (extract_includes.bat.in is generated from this list):
https://github.com/google/protobuf/blob/master/src/Makefile.am#L58

Everything not in this list should be considered internal to protobuf.

haberman:
I thought internal-only symbols were in the `google::protobuf::internal` namespace. These are in `google::protobuf`.

xfxyjwf:
My guess is that when these dependencies (strutil/stringprint/logging etc.) were forked for opensourcing protobuf, the public API part of these dependencies were put in google::protobuf namespace (while their internal implementation were put in google::protobuf::internal). Even though they are not in the google::protobuf::internal namespace, they are not actually part of protobuf API. I think it's better to not expose them through protobuf library when we can.

PR #2576, opened 2017-01-10
python: do not include internal 'strutil.h' header

cristicbz:
I inlined the one function call from `strutil.h` into `pyext/message.cc`. The `strutil.h`  file doesn't get install, so if you're trying to build the python extension you won't have it in your include path (see #1197).

I think the same reasoning as e.g. https://github.com/grpc/grpc/pull/3133 applies here.

(also on a different note, why don't the pip releases include the python cpp implementation? is that on purpose or an oversight?)


bazel-io:

cristicbz:
I signed it!

cristicbz:
friendly ping @haberman (since you seem to have lgtm'd a python-related PR in the past; sorry if not correct)

mcobzarenco:
+1

PR #2609, opened 2017-01-19
Undef TYPE_BOOL to avoid conflict with iOS.

yixiang:
TYPE_BOOL is defined as a macro in <ConditionalMacros.h>, which gets implicitly included in almost all iOS source files.

This fixes complaints like http://go/soverflow/15759559

For some context, here is how TYPE_BOOL is defined in ConditionalMacros.h

  #ifdef __cplusplus
     #define TYPE_BOOL                1
  #else
     #define TYPE_BOOL                0
  #endif


bazel-io:

yixiang:
I signed it!

yixiang:
Is there anything else I need to do to get this merged?

xfxyjwf:
Sorry for the delay. The change looks good.

PR #2613, opened 2017-01-20
 optimization and quick workaround to memory leak

aausch:
incremental solution for https://github.com/google/protobuf/issues/156

the memory leak is caused by this call to copyreg:

https://github.com/google/protobuf/blob/39f9b43219bc5718b659ed72a2130a7b2ce66108/python/google/protobuf/internal/python_message.py#L182

the default comparator fails to detect duplicate entries, effectively filling the internal table at  `copyreg.__dict__['dispatch_table']` until the machine runs out of memory

can probably implement `__cmp__` at some point before that call as a technically more correct solution - a cache saves on cycles though, and is optimal for my use case, yada yada, hope this progresses a fix

### DIFF
--- a/python/google/protobuf/reflection.py
+++ b/python/google/protobuf/reflection.py
@@ -61,6 +61,8 @@ else:
 # Part of the public interface, but normally only used by message factories.
 GeneratedProtocolMessageType = message_impl.GeneratedProtocolMessageType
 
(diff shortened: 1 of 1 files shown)


bazel-io:

aausch:
I signed it!

PR #2618, opened 2017-01-23
Reverting deployment target to 7.0

davidair:
The Protobuf library doesn’t require the 7.1 deployment target so
reverting it back to 7.0


bazel-io:

TeBoring:
LGTM

PR #2619, opened 2017-01-23
Add python compatibility tests against v2.5.0

anandolee:
Note: Files under python/compatibility_tests/v2.5.0/protos/ and python/compatibility_tests/v2.5.0/tests/ are copied from v2.5.0


bazel-io:

haberman:
I'm curious: did you consider checking in old generated code, instead of checking in old .proto files and relying on an old compiler?

anandolee:
test.sh will download an old protoc according to the command, so I do not need to checking in the generated code for each version.

I plan to add commands  ./test.sh 2.5.0 and ./test.sh 3.0.0-beta-1 (or more) in google/protobuf/tests.sh.

anandolee:
Have added the tests into Travis and all tests passed

python/compatibility_tests/v2.5.0/test.sh
```diff
@@ -0,0 +1,111 @@
...
+
+# Test A.3:
+#   proto set 1: use old version
+#   proto set 2 which may import protos in set 1: use new version
+# Compatiblility test fail if the old verison is less than 3.0.0-alpha-1.
+# Because module name aliases was added in v3.0.0-alpha-1 instead of
+# fully-qualified module names to refer to dependencies: dot was replaced
+# with _dot_.
+if [ "$(printf "$OLD_VERSION\n3.0.0" | sort -V | head -n 1 )" = "3.0.0" ]; then
+  cp old_protoc protoc
```
haberman: I think this is supposed to be `protoc_1` instead of `protoc`?

python/compatibility_tests/v2.5.0/test.sh
```diff
@@ -0,0 +1,111 @@
...
+cp old_protoc protoc_1
+cp old_protoc protoc_2
+python setup.py build
+python setup.py test
+
+# Test A.2:
+#   proto set 1: use new version
+#   proto set 2 which may import protos in set 1: use old version
+cp ../../../src/protoc protoc_1
+python setup.py build
```
haberman: Copy `old_protoc` to `protoc_2`, so this doesn't depend on state from the previous test. That will help if anyone copies and pastes or rearranges the tests.

python/compatibility_tests/v2.5.0/test.sh
```diff
@@ -0,0 +1,111 @@
...
+  echo "[ERROR]: Please build protoc first."
+  exit 1
+}
+
+# Test source compatibility. In these tests we recompile everything against
+# the new runtime (including old version generated code).
+rm google -f -r
+mkdir google
+mkdir google/protobuf
+mkdir google/protobuf/internal
```
haberman: I think `mkdir -p google/protobuf/internal` will do the trick, instead of three separate commands.

anandolee: Done. Thanks!

python/compatibility_tests/v2.5.0/test.sh
anandolee: Done

python/compatibility_tests/v2.5.0/test.sh
anandolee: Done

PR #2622, opened 2017-01-25
Work with truncated tag numbers.

ckennelly:
This allows more compact comparisons (1 byte instead of 4 byte immediates on x86) for each possible wire/tag inside each field.


bazel-io:

acozzette:
ok to test

ckennelly:
Superceded by #2722

PR #2626, opened 2017-01-25
Adds nullability modifiers to resolve Xcode 8.3 warnings

sergiocampama:
Notice that the arrays themselves are __nullable, but the ObjectType pointers inside the arrays are __nonnullable.


sergiocampama:
fyi @thomasvl 
fixes: https://github.com/google/protobuf/issues/2625

thomasvl:
Looks good.  We'll let travis confirm it builds fine with the old Xcode also before landing.

PR #2631, opened 2017-01-25
Oneof accessor should return the field name that is actually set.

TeBoring:


bazel-io:

TeBoring:
@stanley-cheung

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -163,6 +163,14 @@ protected function writeOneof($number, $value)
         $oneof_field->setNumber($number);
     }
 
+    protected function whichOneof($oneof_name)
+    {
```
stanley-cheung: Sorry not familiar with the oneOf API in general. Is the `whichOneOf` naming consistent with other languages?

TeBoring: whichOneof is for implementation only. It's used to implement API in generated code for oneof.
For proto:
message Foo {
  oneof my_oneof {
    int32 a = 1;
    int32 b = 1;
  }
}
In generated code, message Foo will have the method "getMyOneof", which returns the name of the field that is currently set (e.g., a or b).

PR #2633, opened 2017-01-26
Allow OneofOptions to be extended in proto3.

### LINKED ISSUE #2316: OneofOptions cannot be extended in proto3
This piece of unittest_custom_options.proto gives an error in proto3 syntax:

```text
extend google.protobuf.OneofOptions {
  int32 oneof_opt1 = 7740111;
}
```

I would expect OneofOptions to be supported in the same way as any other options.

anandolee:
Fixes #2316


bazel-io:

Issue #2623, opened 2017-01-25
Lack of binary Python packages on PyPI


public:
There is a protobuf wheel on PyPI, but it is an "any" wheel. This is great, but it does mean that if you want your protobufs to be fast you need to compile your own wheels using `--cpp-implementation`.

https://github.com/pypa/manylinux wheels are now a thing that works for shipping binary wheels on Linux. Should this project ship fast binary wheels for Linux and other supported platforms?

TeBoring:
@anandolee Can we have separate wheel for python2 and python3 and upload binaries for them?

public:
@TeBoring Anything I can do to help keep this ticket moving?

ryan-lane:
Agreed with @public. We also need this. Any update here?

TeBoring:
We are working on this.

public:
@teboring I made a PR that will build the wheels for you. Is this unwanted?

TeBoring:
Uploaded: https://pypi.python.org/pypi/protobuf/3.2.0
I also have a PR: https://github.com/google/protobuf/pull/2693 Feel free to leave comments there.

ryan-lane:
Is there any way you can also upload a wheel for 3.0.0?

TeBoring:
Will do that. Could you help check 3.2.0 works?

public:
@TeBoring We are using the 3.2.0 wheels and they seem to work as expected. Thanks a lot. :+1:

ewalker11:
@TeBoring Just checking on the 3.0.0 wheels, would it be possible to upload those this week? Thanks for your responsiveness on this issue!

TeBoring:
I'll do that.

TeBoring:
done

tebriel:
@TeBoring this causes Protobuf to now have a different behavior for 3.0.0 than it did before. 

This new version changes the behavior of protobuf such that running SomeProtobuf.FromString(‘abc’) now results in a “TypeError: a bytes-like object is required, not 'str'” where before it was: “google.protobuf.message.DecodeError: Truncated message.”. I'm concerned about the behavior change here and what other potential issues may arise from this upload.

tebriel:
We cannot help but download this version, as the bdist wheel takes precedence over the any wheel. So even requirements.txt requiring protobuf==3.0.0 still causes subsequent builds to take the new wheel.

TeBoring:
[...]

PR #2635, opened 2017-01-26
Python manylinux wheel builds

public:
So far this PR just includes a shell script for running inside a https://github.com/pypa/manylinux CentOS5 build container. It will produce statically linked, `cpp-implementation` enabled, python2 and python3 compatible wheels for whatever arch of manylinux builder it's run inside of.

It seems protobuf requires a much newer version of autotools that is present in CentOS5 so this has to download and install newer versions of those first. I'm not sure if there's a better way to do this upgrade (yum backports somewhere?), I'm a bit new to the RedHat ecosystem.

Run this like

```
docker run -w /io/protobuf -v $(pwd):/io/protobuf --rm \
    quay.io/pypa/manylinux1_x86_64 python/manylinux-build.sh
```

And it will dump wheels into `python/dist` for you.

See also #2623

### DIFF
--- a/python/manylinux-build.sh
+++ b/python/manylinux-build.sh
@@ -1,0 +1,55 @@
+#!/bin/sh
+
+set -xe
+
+pushd /tmp
+
+curl -O https://ftp.gnu.org/gnu/autoconf/autoconf-2.69.tar.gz
+gunzip autoconf-2.69.tar.gz
+tar xvf autoconf-2.69.tar
+
+pushd autoconf-2.69
+
+./configure
+make
+make install
+
+popd
+
+curl -O https://ftp.gnu.org/gnu/automake/automake-1.15.tar.gz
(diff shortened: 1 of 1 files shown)


bazel-io:

python/manylinux-build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+export SED=$(which sed)
+
+./autogen.sh
+./configure --disable-shared "CFLAGS=-fPIC" "CXXFLAGS=-fPIC"
+make
+make check
+
+cd python
+PYTHON_VERSIONS="cp27-cp27m cp34-cp34m cp35-cp35m cp36-cp36m"
```
public: This list was chosen to match those uploaded to PyPI by grpc.

public:
Anything I can do to keep this moving?

xfxyjwf:
@TeBoring is working on something similar.

PR #2637, opened 2017-01-26
Add namespace for std::string in case ambiguous symbol error

linluxiang:
The string argument has no namespace. It will cause "ambiguous symbol" compiling error in the situation that another "string" is defined in some other header files.


bazel-io:

linluxiang:
I signed it!

BSBandme:
Closed for now as it has been fixed in other PR

PR #2645, opened 2017-01-30
C#: Implement IReadOnlyList<T> in RepeatedField<T>

### LINKED ISSUE #2041: Support IReadOnly* in C# collections
Our RepeatedField and MapField classes implement everything required for `IReadOnlyList<T>` (and implicitly `IReadOnlyCollection<T>`) and `IReadOnlyDictionary<T>` respectively, but don't implement those interfaces themselves, which can be annoying. We should implement them.

(Note that IReadOnly\* doesn't actually mean it's read-only; it just means the interface is read-only. A bit confusing...)

jskeet:
We explicitly don't do this when targeting .NET 3.5, where the
interface doesn't exist.

No implementation is required, as we're already implementing
everything we need for IList<T>.

Fixes #2041.


anandolee:
LGTM

PR #2649, opened 2017-01-31
Avoid checking for tag == 0.

ckennelly:
This moves a common case that occurs during every parse off the error path.


bazel-io:

PR #2658, opened 2017-02-02
Fixes cmake build failure for js_embed step

udnaan:
This fixes an issue where js_embed generation step would fail since it cannot find js_embed binary.

The binary is correctly generated but since it is in the ${CMAKE_CURRENT_BINARY_DIR}, the system would be unable to find it.

Perhaps it works for windows but it fails for *nix.


bazel-io:

udnaan:
I don't have a windows build machine atm to see why it would fail with windows build. This will have to wait unless someone else with a windows machine is willing to help.

Yangqing:
Note that this would probably still fail in cross compilation (like Android or iOS) right? CMake will create a binary that only runs on the target platform, and one cannot use it directly from the host.

pherl:
Could check the appveyor results? This seems to break windows build.

xfxyjwf:
I have removed the js_embed binary from our build:
https://github.com/google/protobuf/pull/4709

It should fix all issues related to this binary.

PR #2660, opened 2017-02-03
Bug fix: When encoding, negative int32 values should be padded to int64 in order to be wire compatible.

TeBoring:


bazel-io:

TeBoring:
@stanley-cheung @haberman

Issue #1606, opened 2016-05-26
Missing value/null support for scalar value types in proto 3


lostindark:
From the protobuf wire format we can tell whether a specific field exists or not. And in protobuf 2 generated code, all fields have a "HasXXX" method to tell whether the field exists or not in code. However in proto 3, we lost that ability. E.g. now we can't tell if a int32 field is missing, or has a value of 0 (default for int32 type).

In many scenario, we need the ability to differentiate missing vs default value (basically nullable support for scalar types).

Feng suggest workarounds:

> 1. Use a wrapper message, such as google.protobuf.Int32Value. In proto3, message fields still have has-bits.
> 2. Use an oneof. For example:
>    message Test1 {
>    oneof a_oneof {
>      int32 a = 1;
>    }
>    }
>    then you can check test.getAOneofCase().

However, this requires change the message definition, which is a no go if you need to keep the message compatible while upgrade from proto2 to proto 3.

We need to add back the support to detect whether a scalar type field exist or not.

More discussion here: https://groups.google.com/forum/#!topic/protobuf/6eJKPXXoJ88

ngbrown:
While proto3 mentions using proto2 Message Types

> It's possible to import proto2 message types and use them in your proto3 messages, and vice versa. However, proto2 enums cannot be used in proto3 syntax.

It might not be good application design to expect the .proto files and binary format to stay the same after switching to proto3.

Even though it appears that using the `wrappers.proto` is the way proto3 is addressing nullable fields, and the c#/.net library has convenient nullable formatting for them, they will end up consuming an extra byte per value because of the additional message layer.

xfxyjwf:
@lostindark, workaround (2) using an oneof is a wire compatible change.

lostindark:
> @lostindark, workaround (2) using an oneof is a wire compatible change.
The issue is it doesn't produce nice code, and it needs box/unbox when use (hurt performance).

xfxyjwf:
@lostindark what if we optimize that case (one primitive field in an oneof) to eliminate the box/unbox cost?

lostindark:
@xfxyjwf That seems a good solution with minimum change required. However, I do wish I don't need to name the oneof, like below:
`message Test1 { oneof { int32 a = 1; } }`

xfxyjwf:
I am thinking of a dedicated syntax for that. Something like:
```proto
message Foo {
  nullable int32 value = 1;
}
```
which is basicaly a syntax sugar for:
```proto
[...]

PR #2664, opened 2017-02-04
Implement nullable syntaxic sugar keyword

dopuskh3:
Implementation proposal for nullable keyword that convert a scalar field to a one-field oneof.

Issue #1606


bazel-io:

dopuskh3:
I signed it!

heyleke:
What is the status of this, I see Travis fails on a PHP issue, does that mean this pull request will not go to master? And it needs a fix?

xfxyjwf:
This PR implements a proposal that hasn't been reviewed/approved (the review is undergoing). I'll post back with status update of the review which is either approved or rejected.

dopuskh3:
Hi @xfxyjwf , 

After some dicussions with @acozzette, it looks like making this change accepted could require some time and work.

Alternatively, I'm currently working on adding insertion points for setters and getter that will allow us to implement nullable fields using a less intrusive fashion:

- Adding a message option that tells if a specific message should support for nullable fields.
- If it does, then make declaration of a map<int32, bool> using a reserved field id required for message and all its nested message declarations
- Inject code in setters and getters to modify the content of the map value for corresponding field id.
- Inject hasXXX for message fields. 

The proposal has been done here Issue #2685. 

In the comming hours/day I will modify my merge request to add 2 insertion points per setters so that we can inject at the beginning or end of the corresponding setters.

I almost have a plugin working and implementing this.

xfxyjwf:
As I posted at https://github.com/google/protobuf/issues/1606#issuecomment-281832148, the team decided to not move forward with this proposal.

PR #2673, opened 2017-02-06
Add fixed version to phpunit used in travis

TeBoring:
The latest phpunit only supports php7.0+. Added an old fixed version to work with php 5.6.


bazel-io:

pherl:
OK to test

PR #2676, opened 2017-02-06
JS compatibility tests

acozzette:
Added compatibility tests for JavaScript. This initial pull request just contains tests for verifying compatibility with 3.0.0, but it should be easy to extend to 3.1, 3.2, etc.


acozzette:
@haberman This PR will probably need some more work just to clean up some rough edges of the shell script `test.sh` but I thought I would send it out now to get some early feedback. I was tempted to write another `gulpfile.js` instead of a shell script, does that sound like a reasonable approach? Also I ended up having to create a new top-level `compatibility_tests` directory because I ran into issues with `calcdeps.py` complaining about conflicting symbols when I kept everything in subdirectories of `js/`. Maybe there's a cleaner way to handle this.

acozzette:
retest this please

acozzette:
@haberman I think this should now be about ready to go; do you want to take a look?

js/compatibility_tests/v3.0.0/test.sh
```diff
@@ -0,0 +1,89 @@
...
+npm install && npm test
+popd
+
+old_protoc=./protoc
+new_protoc=../../../src/protoc
+
+# The protos in group 2 have some dependencies on protos in group 1. The tests
+# will verify that the generated code for one group can be regenerated
+# independently of the other group in a compatible way.
+group1_protos="data.proto test3.proto test5.proto commonjs/test6/test6.proto testbinary.proto testempty.proto test.proto"
```
haberman: This logic seems to be duplicated from the gulpfile. Do we need to have this logic in both places?

acozzette: Good question and sorry for the delayed reply. I thought about this a bit and I am thinking that it probably does have to be duplicated actually, because over time we might want to add new test .proto files (or even remove old ones) and thus we can't in general assume that the .proto files used for a particular version's compatibility test match the ones used for the current version. Does that seem fair? On the other hand I guess we could try to factor out some of this logic from gulpfile.js so that for future versions we have a more streamlined way of generating the compatibility test directory (ideally we could have it entirely scripted).

haberman: Maybe put the logic in the gulpfile only and then snapshot the current gulpfile into a copy for the old code?

If this ends up being more complicated overall, feel free to keep what you've got, but make a note in comments about the duplicated logic.

acozzette: Ok, I thought about this a bit but couldn't think of a good way to share that information without creating a fair bit more complexity, so I just added a comment about it for now.

acozzette:
@haberman Do you have any other comments on this pull request or does it look ok to merge?

haberman:
LGTM.

PR #2683, opened 2017-02-07
Added a "protobuf_headers" target that includes only the headers.

keveman:
Projects building multiple .so's and loading them dynamically can
link against this target instead of including all of protobuf.
TensorFlow is an example of such a project.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -181,6 +181,13 @@ cc_library(
     deps = [":protobuf_lite"],
 )
 
+cc_library(
+    name = "protobuf_headers",
+    hdrs = glob(["src/**/*.h"]),
+    includes = ["src/"],
+    visibility = ["//visibility:public"],
+)
+
 objc_library(
     name = "protobuf_objc",
     hdrs = ["objectivec/GPBProtocolBuffers.h"],


bazel-io:

xfxyjwf:
As discussed in chat, I think requiring all libraries you depend on to have a special header-only target is not a manageable solution and it's better to use a skylark rule or get native support from bazel to work around it in your own project settings.

danbarnes333:
@keveman I am trying to build tensorflows shared library to completely dynamically link against protobuf and came across this pr and your tensorflow [commit](https://github.com/tensorflow/tensorflow/commit/e83a041dd51dcef9df357375b9f9ad7cc62039f2).

In the commit mentioned you have changed the BUILD files to dynamically link to protobuf just for the custom ops. We have tried for quite some time to get the whole of libtensorflow.so to dynamically link against protobuf but however we do it protobuf symbols are still pulled into the shared library.

I was wondering if you have tried this before / have any suggestions for doing so?

Thanks

xfxyjwf:
This PR is a duplicate. The change was already made in https://github.com/google/protobuf/commit/2d430f8c72e359393ed15dd456765d3154981311

PR #2689, opened 2017-02-08
Avoid aliasing CodedInputStream::buffer_ when parsing little endian integers

ckennelly:
This eliminates an unnecessary reload of `buffer_` that occurs (before this change) after the store to `*value`.


bazel-io:

acozzette:
retest this please

PR #2691, opened 2017-02-08
Bazel can build protobuf when it's not in the root

cgrushko:
That is, Bazel can now build protobuf when the latter resides in a subdirectory of a project.


cgrushko:
@xfxyjwf

PR #2692, opened 2017-02-08
Update load() statement to latest style

cgrushko:
The first argument is currently implicitly a .bzl file.
Change this to be explicit.


cgrushko:
@xfxyjwf

PR #2699, opened 2017-02-09
Switch to gcc atomic intrinsics for macOS and delete the file that uses the deprecated atomics

jhseu:
The static_assert on the sizeof std::atomic fails with nvcc. Use the gcc atomic intrinsics instead, which are supported in both gcc and clang.

Manually tested that this fixes the nvcc build.

@sergiocampama


jhseu:
Primarily needed to fix the GPU build for TensorFlow on macOS.

src/google/protobuf/stubs/atomicops.h
```diff
@@ -194,14 +194,6 @@ Atomic64 Release_Load(volatile const Atomic64* ptr);
 #elif defined(GOOGLE_PROTOBUF_OS_AIX)
 #include <google/protobuf/stubs/atomicops_internals_power.h>
 
-// Apple.
```
TeBoring: Does it still pass on older version of mac

jhseu: I don't have access to an older mac, but the intrinsics work on clang/llvm 3.4 from 2014 (tested on my desktop). Looking at commits, should be fine for versions from 2011 forward. How old do we want to support?

TeBoring: Could you try this: https://github.com/google/protobuf/blob/master/protoc-artifacts/build-protoc.sh#L204?

TeBoring: You may need to comment out the c++11 flag in configure.ac

jhseu: Tested with `make clean && ./configure CXXFLAGS="-mmacosx-version-min=10.7" && make` and it doesn't affect the versions the existing code builds at:
\>= 10.9 required with C++11 set
\>= 10.7 required when deleting it from configure.ac

TeBoring:
LGTM. Could you squash commits into a single one?

jhseu:
Done.

FYI, in TensorFlow, we default to squash when merging pull requests:
https://github.com/blog/2141-squash-your-commits

PR #2701, opened 2017-02-10
Add csharp compatibility tests against v3.0.0

anandolee:


bazel-io:

jskeet:
So this is an exact copy of the tests from 3.0.0, so we can run the same tests against the newer code to check for compatibility? I can see cases where we may have failing tests as a restriction (which was tested as a restriction) is then lifted... is the plan to handle that on a case-by-case basis?

Obviously I'm not going to manually review the 7000+ lines - is there anything you'd particularly like to draw my attention to for checking?

anandolee:
There's two commits: first one is just copy form v3.0.0 that do not need to review. Second commit is the scripts which run the tests and need to be reviewed

csharp/compatibility_tests/v3.0.0/test.sh
```diff
@@ -0,0 +1,102 @@
...
+
+  ./protoc_1 -Iprotos/csharp --csharp_out=src/Google.Protobuf.Test \
+    --csharp_opt=base_namespace=UnitTest.Issues \
+    protos/csharp/protos/unittest_issues.proto
+
+  ./protoc_2 -Iprotos/src --csharp_out=src/Google.Protobuf.Test \
+    --csharp_opt=base_namespace=Google.Protobuf \
+    protos/src/google/protobuf/unittest_proto3.proto
+
+  # Build and test.
```
jskeet: Not an issue for now, but something we'll need to work around at some point: the project.json format used for the current code is going away, to be replaced by csproj. At some point we'll want to update to that, and we'll need to do it for this old code too.

csharp/compatibility_tests/v3.0.0/test.sh
```diff
@@ -0,0 +1,102 @@
...
+[ -f ../../../src/protoc ] || {
+  echo "[ERROR]: Please build protoc first."
+  exit 1
+}
+
+# Download old version protoc compiler (for linux)
+wget $OLD_VERSION_PROTOC -O old_protoc
+chmod +x old_protoc
+
+# Test source compatibility. In these tests we recompile everything against
```
jskeet: So this is just testing that we can still compile, yes?

csharp/compatibility_tests/v3.0.0/test.sh
```diff
@@ -0,0 +1,102 @@
...
+    protos/src/google/protobuf/map_unittest_proto3.proto \
+    protos/src/google/protobuf/unittest_import_proto3.proto \
+    protos/src/google/protobuf/unittest_import_public_proto3.proto \
+    protos/src/google/protobuf/unittest_well_known_types.proto
+
+  ./protoc_1 -Iprotos/csharp --csharp_out=src/Google.Protobuf.Test \
+    --csharp_opt=base_namespace=UnitTest.Issues \
+    protos/csharp/protos/unittest_issues.proto
+
+  ./protoc_2 -Iprotos/src --csharp_out=src/Google.Protobuf.Test \
```
jskeet: It's not clear to me why just this one uses protoc_2, when the others use protoc_1. Is is because this is the only file that's changing? (I've read the comments on the tests themselves, and I'm none the wiser I'm afraid!)

tests.sh
```diff
@@ -119,6 +119,11 @@ build_csharp() {
   (cd csharp/src; dotnet restore)
   csharp/buildall.sh
   cd conformance && make test_csharp && cd ..
+
+  # Run csharp compatibility test.
+  cd csharp/compatibility_tests/v3.0.0
```
jskeet: Perhaps it would be cleaner if the test script used its own location automatically? Then we could just run

`csharp/compatibility_tests/v3.0.0/test.sh`

which would be simpler.

csharp/compatibility_tests/v3.0.0/test.sh
anandolee: The old version tests should pass too.

tests.sh
anandolee: Done. Thanks

csharp/compatibility_tests/v3.0.0/test.sh
anandolee: No. 
The test separate proto files into two sets. Set 2 proto files may import some of proto files in set 1.
3 cases are tested:
Case 1:
protobuf runtime: new runtime
both .proto files set 1 and set 2 use v3.0.0 protoc

Case 2:
protobuf runtime: new runtime
.proto files set 1 uses new version protoc
.proto files set 2 uses v3.0.0 protoc

Case 3:
protobuf runtime : new runtime
.proto files set 1 uses v3.0.0 protoc
.proto files set 2 uses new version protoc

csharp/compatibility_tests/v3.0.0/test.sh
anandolee: Acknowledged. We can update it when needed

csharp/compatibility_tests/v3.0.0/test.sh
jskeet: Okay - it's not entirely clear what the motivation is for the different sets, but I'm fine with it. (And the PR in general.)

PR #2714, opened 2017-02-13
Improve support for plugin parameters.

thomasvl:
--[name]_opt support depended on the plugin being register, and didn't support
working with just --[name]_out directive (where the plugin is found via the
users PATH. This extends the command line handing to allow --[name]_out to
be all it takes for the _opt directive to also be supported.

Fixes https://github.com/google/protobuf/issues/2712

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface_unittest.cc
+++ b/src/google/protobuf/compiler/command_line_interface_unittest.cc
@@ -689,10 +689,36 @@ TEST_F(CommandLineInterfaceTest, UnrecognizedExtraParameters) {
     "message Foo {}\n");
 
   Run("protocol_compiler --plug_out=TestParameter:$tmpdir "
+      "--unknown_plug_a_opt=Foo "
+      "--unknown_plug_b_opt=Bar "
+      "--proto_path=$tmpdir foo.proto");
+
+  ExpectErrorSubstring("Unknown flag: --unknown_plug_a_opt");
+  ExpectErrorSubstring("Unknown flag: --unknown_plug_b_opt");
+}
+
+TEST_F(CommandLineInterfaceTest, ExtraPluginParametersForOutParameters) {
+  // This doesn't rely on the plugin having been registred and instead that
+  // the existence of --[name]_out is enough to make the --[name]_opt valid.
+  // However, running out of process plugins found via the search path (i.e. -
+  // not pre registered with --plugin) isn't support in this test suite, so we
+  // list the options pre/post the _out directive, and then include _opt that
+  // will be unknown, and confirm the failure output is about the expected
+  // unknown directive, which means the other were accepted.
+  // NOTE: UnrecognizedExtraParameters confirms that if two unknown _opt
+  // directives appear, they both are reported.
+
+  CreateTempFile("foo.proto",
+    "syntax = \"proto2\";\n"
+    "message Foo {}\n");
+
+  Run("protocol_compiler --plug_out=TestParameter:$tmpdir "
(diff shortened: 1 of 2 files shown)


src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -689,10 +689,36 @@ TEST_F(CommandLineInterfaceTest, UnrecognizedExtraParameters) {
...
+    "message Foo {}\n");
+
+  Run("protocol_compiler --plug_out=TestParameter:$tmpdir "
+      "--xyz_opt=foo=bar --xyz_out=$tmpdir "
+      "--abc_out=$tmpdir --abc_opt=foo=bar "
       "--unknown_plug_opt=Foo "
       "--proto_path=$tmpdir foo.proto");
 
-  ExpectErrorSubstring("Unknown flag: --unknown_plug_opt");
+  ExpectErrorText("Unknown flag: --unknown_plug_opt\n");
```
pherl: Use Substring instead. There used be issues to run the tests on windows shells with extra error messages.

src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -689,10 +689,36 @@ TEST_F(CommandLineInterfaceTest, UnrecognizedExtraParameters) {
...
+  // unknown directive, which means the other were accepted.
+  // NOTE: UnrecognizedExtraParameters confirms that if two unknown _opt
+  // directives appear, they both are reported.
+
+  CreateTempFile("foo.proto",
+    "syntax = \"proto2\";\n"
+    "message Foo {}\n");
+
+  Run("protocol_compiler --plug_out=TestParameter:$tmpdir "
+      "--xyz_opt=foo=bar --xyz_out=$tmpdir "
```
pherl: Are there anyway we can test the parameters are actually passed into the xyz and abc plugins?

thomasvl: Not that I can see with the current testing setup, as the plugins don't exist and I'm using the failing unknown_plug to prevent it trying to run.  `ExtraPluginParameters` is what checks that they properly get collected/passed.

src/google/protobuf/compiler/command_line_interface_unittest.cc
thomasvl: That would invalidate the test as it would mean errors about xyz/abc aren't checked.  We could use SubString if we add a ExpectNoErrorSubstring() and confirm they don't show up, would that be better?

pherl: Let's wait and see if the continuous build pass. Should be good to go.

thomasvl: Instead of using the failures to prevent running, the option option would be to clone most of `CommandLineInterfaceTest::RunWithArgs()`, and make `CommandLineInterfaceTest::ParseArgs()` that calls `ParseArguments` instead of `Run`, and a few tests could be switched over to that instead since they are just checking errors without actually doing the full run of the generators.  That would be a larger change, might better cover things.  (It still wouldn't let this case confirm the arguments are collected for the plugin, but the other test should cover that.

thomasvl: Is the javanano_oracle7.javanano_oracle7 failure on grpc-jenkins a known problem?

pherl: I don't think that is related to your change.

PR #2727, opened 2017-02-16
Add PreferredAlias into OriginalNameAttribute

### LINKED ISSUE #2405: C# ToString() on a message fails with an exception
Hi

I am facing a problem when trying to invoke ToString() on a message. The following exception is thrown. I think this may be because my enum has allow_alias set to true, and the same numeric value occurs more than once.

```
An item with the same key has already been added. Key: Pln
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(Object key)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Google.Protobuf.JsonFormatter.OriginalEnumValueHelper.GetNameMapping(Type enumType)
   at Google.Protobuf.JsonFormatter.OriginalEnumValueHelper.GetOriginalName(Object value)
   at Google.Protobuf.JsonFormatter.WriteValue(TextWriter writer, Object value)
   at Google.Protobuf.JsonFormatter.WriteMessageFields(TextWriter writer, IMessage message, Boolean assumeFirstFieldWritten)
   at Google.Protobuf.JsonFormatter.WriteMessage(TextWriter writer, IMessage message)
   at Google.Protobuf.JsonFormatter.WriteMessageFields(TextWriter writer, IMessage message, Boolean assumeFirstFieldWritten)
   at Google.Protobuf.JsonFormatter.WriteMessage(TextWriter writer, IMessage message)
   at Google.Protobuf.JsonFormatter.WriteList(TextWriter writer, IList list)
   at Google.Protobuf.JsonFormatter.WriteMessageFields(TextWriter writer, IMessage message, Boolean assumeFirstFieldWritten)
   at Google.Protobuf.JsonFormatter.WriteMessage(TextWriter writer, IMessage message)
   at Google.Protobuf.JsonFormatter.WriteList(TextWriter writer, IList list)
   at Google.Protobuf.JsonFormatter.WriteMessageFields(TextWriter writer, IMessage message, Boolean assumeFirstFieldWritten)
   at Google.Protobuf.JsonFormatter.WriteMessage(TextWriter writer, IMessage message)
   at Google.Protobuf.JsonFormatter.Format(IMessage message)
```

anandolee:
There might be duplicated enum values when allow_alias is true. Add Distinct() for Enum Dictionary used in JsonFormatter to remove the duplication.

fix #2405

### DIFF
--- a/csharp/src/Google.Protobuf.Test/TestProtos/UnittestProto3.cs
+++ b/csharp/src/Google.Protobuf.Test/TestProtos/UnittestProto3.cs
@@ -85,70 +85,72 @@ namespace Google.Protobuf.TestProtos {
             "ZXN0TXV0dWFsUmVjdXJzaW9uQRIzCgJiYhgBIAEoCzInLnByb3RvYnVmX3Vu",
             "aXR0ZXN0LlRlc3RNdXR1YWxSZWN1cnNpb25CImIKFFRlc3RNdXR1YWxSZWN1",
             "cnNpb25CEjIKAWEYASABKAsyJy5wcm90b2J1Zl91bml0dGVzdC5UZXN0TXV0",
-            "dWFsUmVjdXJzaW9uQRIWCg5vcHRpb25hbF9pbnQzMhgCIAEoBSLrAgoXVGVz",
-            "dENhbWVsQ2FzZUZpZWxkTmFtZXMSFgoOUHJpbWl0aXZlRmllbGQYASABKAUS",
-            "EwoLU3RyaW5nRmllbGQYAiABKAkSMQoJRW51bUZpZWxkGAMgASgOMh4ucHJv",
-            "dG9idWZfdW5pdHRlc3QuRm9yZWlnbkVudW0SNwoMTWVzc2FnZUZpZWxkGAQg",
-            "ASgLMiEucHJvdG9idWZfdW5pdHRlc3QuRm9yZWlnbk1lc3NhZ2USHgoWUmVw",
-            "ZWF0ZWRQcmltaXRpdmVGaWVsZBgHIAMoBRIbChNSZXBlYXRlZFN0cmluZ0Zp",
-            "ZWxkGAggAygJEjkKEVJlcGVhdGVkRW51bUZpZWxkGAkgAygOMh4ucHJvdG9i",
-            "dWZfdW5pdHRlc3QuRm9yZWlnbkVudW0SPwoUUmVwZWF0ZWRNZXNzYWdlRmll",
-            "bGQYCiADKAsyIS5wcm90b2J1Zl91bml0dGVzdC5Gb3JlaWduTWVzc2FnZSLH",
-            "AQoSVGVzdEZpZWxkT3JkZXJpbmdzEhEKCW15X3N0cmluZxgLIAEoCRIOCgZt",
-            "eV9pbnQYASABKAMSEAoIbXlfZmxvYXQYZSABKAISUwoVc2luZ2xlX25lc3Rl",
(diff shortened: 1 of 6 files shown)


bazel-io:

anandolee:
Jon, do you know if Distinct() keeps the first item for duplication?

jskeet:
It keeps the first one it encounters, but there's no guarantee which one will be returned first by reflection, so it's unstable, basically.

An alternative option would be to make protoc emit an attribute for fields to ignore when building the dictionary. That would allow us to make the decision in a consistent way, but would require regenerating the C# code before it would be effective, of course.

We could keep the current approach *and* add the attribute, so that we'd never throw an error, but the results might be inconsistent until code was regenerated - what do you think about that solution?

csharp/src/Google.Protobuf.Test/TestProtos/UnittestProto3.cs
```diff
@@ -209,8 +209,8 @@ public enum TestEnumWithDupValue {
     [pbr::OriginalName("FOO1")] Foo1 = 1,
     [pbr::OriginalName("BAR1")] Bar1 = 2,
     [pbr::OriginalName("BAZ")] Baz = 3,
-    [pbr::OriginalName("FOO2")] Foo2 = 1,
-    [pbr::OriginalName("BAR2")] Bar2 = 2,
```
anandolee: Does Empty the OriginalNameAttribute for duplicate enum values make sense to you? So there's no need to add more attribute. We do not need the OriginalName for duplicate enum.

If so, I will go head change the other cods like protoc.

csharp/src/Google.Protobuf.Test/TestProtos/UnittestProto3.cs
```diff
@@ -209,8 +209,8 @@ public enum TestEnumWithDupValue {
...
+    [pbr::OriginalName("")] Foo2 = 1,
+    [pbr::OriginalName("")] Bar2 = 2,
   }
```
anandolee: Does Empty the OriginalNameAttribute for duplicate enum values make sense to you? So there's no need to add more attribute. We do not need the OriginalName for duplicate enum.

If so, I will go ahead change the rest code and protoc.

jskeet: Rather than losing the information (even if we're not using it at the moment) I think it would be better to add an extra `bool` property to the attribute, e.g.

```csharp
[pbr::OriginalName("FOO1")] Foo1 = 1,
[pbr::OriginalName("BAR1")] Bar1 = 2,
[pbr::OriginalName("FOO2", PreferredAlias = false )] Foo2 = 1,
[pbr::OriginalName("BAR2", PreferredAlias = false)] Bar2 = 2,
```

It feels cleaner in terms of not trying to make one bit of information represent two things. (It may also be useful in terms of people reading the code.)

anandolee: Have added the PreferredAlias into OriginalName

csharp/src/Google.Protobuf/Reflection/OriginalNameAttribute.cs
```diff
@@ -47,12 +47,19 @@ public class OriginalNameAttribute : Attribute
         public string Name { get; set; }
 
         /// <summary>
+        /// If the name is preffered in the .proto file.
```
jskeet: Nit: preffered => preferred

anandolee:
Thanks for review!
I think it's fine not deal with the already-generated protos. Squashed into one commit and added the change for codegen. Will submit after travis green

PR #813, opened 2015-09-14
Add GOOGLE_ATTRIBUTE_NOINLINE to GetArena() and GetMaybeArenaPointer(…

0xAAE:
…) methods. This is to avoid "unresolved link" errors in MSVC 2015 during Release build

PR #726, opened 2015-08-21
Define GOOGLE_ATTRIBUTE_NOINLINE for MSVC. Workaround for VS2015 Release build compiler bug

DouglasHeriot:
See issue #240 - MSVC in VS2015 seems to inline a function it shouldn't. My original workaround was to disable inlining for the whole file, but I found a way to do it on just this specific function using __declspec(noinline).
Unfortunately __declspec has to go at the start of the function declaration, while __attribute in GCC can go either before or after. I had to move lots of GOOGLE_ATTRIBUTE_NOLINE to make it compile. I have not yet tested this change with GCC.

Will there be other side effects of defining this, given it wasn't previously?

I also noticed a few functions marked with both the 'inline' keyword, and GOOGLE_ATTRIBUTE_NOINLINE - huh? Is there an explanation for this, or is it an oversight?

PR #2729, opened 2017-02-17
Fix unresolved symbols with MSVC12 and /Zc:inline

MarcelRaad:
In #726 and #813, linking errors with MSVC14 were resolved. The change in
MSVC12 leading to these errors was not a newly introduced bug, but that
/Zc:inline was made on by default. This option is also available with MSVC12,
so the workaround should be applied for it too.

### DIFF
--- a/src/google/protobuf/message.cc
+++ b/src/google/protobuf/message.cc
@@ -451,8 +451,8 @@ struct ShutdownRepeatedFieldRegister {
 
 namespace internal {
 template<>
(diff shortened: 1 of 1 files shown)


bazel-io:

MarcelRaad:
I signed it!

acozzette:
ok to test

acozzette:
@MarcelRaad Thanks for the pull request!

PR #2734, opened 2017-02-18
README.md: Make docs URL a link

msabramo:
Dedent URL so it renders as a link and not as code (fixed width font and not clickable).


bazel-io:

acozzette:
@msabramo Thanks for the pull request!

Issue #2627, opened 2017-01-25
Static analyzer complaining that dispatch_once cannot use an instance variable as predicate


sergiocampama:
This means that the readOnlySemaphoreCreationOnce_ instance variable needs to be converted into a static or global variable.

Most likely change that introduced this is: http://llvm.org/viewvc/llvm-project?view=revision&revision=285605

thomasvl:
Fixed in #2748

PR #2746, opened 2017-02-22
Disable deprecated implementation warnings for Tests

sergiocampama:


sergiocampama:
@thomasvl  new implicit warning passed on the new beta... disable for the tests about the unittest_deprecated.proto messages.

thomasvl:
So #2627 went away?  i.e. - we should close that?

What exactly warns in on the deprecated side of things?  We want the generated .m files to now warn for things they are defining (so the generated code always compiles cleanly).  So I'm not sure suppressing is what we want if we're using this to track that the code works correctly.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -217,14 +217,16 @@ void FileGenerator::GenerateHeader(io::Printer *printer) {
...
+  //  deprecated-declarations and deprecated-implementations suppression is 
+  //    only needed when any symbol in this proto file has something 
+  //    deprecated or if it references something from another file that is
+  //    deprecated.
   printer->Print(
       "// @@protoc_insertion_point(imports)\n"
       "\n"
       "#pragma clang diagnostic push\n"
       "#pragma clang diagnostic ignored \"-Wdeprecated-declarations\"\n"
+      "#pragma clang diagnostic ignored \"-Wdeprecated-implementations\"\n"
```
thomasvl: Why'd you do this in the header, isn't it only needed in the .m file?

sergiocampama:
Can't reproduce any more

PR #2751, opened 2017-02-23
Added a header only cc_library target for the protobuf library.

keveman:


bazel-io:

xfxyjwf:
In the email thread it's mentioned a bazel fix is being working on that can solve the issue for tensorflow. Ping me if that doesn't work out.

damienmg:
Hi there,

The Bazel fix might take a long time to arrive as it is not simple. The Bazel team recommend to merge this PR in the time being.

BUILD
```diff
@@ -181,6 +181,13 @@ cc_library(
     deps = [":protobuf_lite"],
 )
 
+cc_library(
+    name = "protobuf_headers",
+    hdrs = glob(["src/**/*.h"]),
+    includes = ["src/"],
+    visibility = ["//visibility:public"],
```
xfxyjwf: Please restrict the visibility to tensorflow only and add a TODO pointing to the missing feature in bazel.

keveman: @damienmg what is the best way to restrict visibility to tensorflow only here, such that it works for both the cases, i.e., TF compiled by itself, and TF as a submodule? I was thinking something like:

    visibility = ["@org_tensorflow//:internal"]

but not sure if that would work in both cases. Also, what is the succinct description of the Bazel problem I can add here? Is there a tracking issue that I can simply link?

keveman:
Gentle ping.

BUILD
xfxyjwf: Please add a comment to explain what this rule is for and advise users to stay away from it. Also add a TODO that we should remove this rule when bazel supports depending on a cc_library without linking.

keveman: Done.

xfxyjwf:
ok to test

keveman:
Gentle ping again :)

PR #2754, opened 2017-02-24
Eliminate distinction between scalar and repeated message field parsing.

ckennelly:
Since the elimination of ExpectTag (5a76e633ea9b), the generated code is different while having the same functional behavior.  This shifts the recursion limit checks into CodedInputStream::DecrementRecursionDepthAndPopLimit (which is not inlined).


bazel-io:

TeBoring:
@gerben-s

PR #2766, opened 2017-02-27
Add mergeFrom method on Message

TeBoring:
This method merges the contents of the specified message into the
current message. Singular fields that are set in the specified message
overwrite the corresponding fields in the current message.  Repeated
fields are appended. Map fields key-value pairs are overritten.
Singular/Oneof sub-messages are recursively merged. All overritten
sub-messages are deep-copied.


bazel-io:

TeBoring:
@stanley-cheung to review

PR #2771, opened 2017-02-28
Avoid aliasing RepeatedField<T>::Rep* during serialization.

ckennelly:
This removes a load per element when a repeated primitive field by avoiding a reload of RepeatedField<T>::rep_.  This is also a data dependency for the subsequent (and remaining) load.

For fixed-sized fields, this extends the changes to SerializeWithCachedSizes in 7f3e23707 to SerializeWithCachedSizesToArray.


bazel-io:

TeBoring:
I suspect this is obsolete. @gerben-s to take a look.

PR #2774, opened 2017-03-01
Use closurebuilder.py in favor of calcdeps.py for compiling JavaScript

acozzette:
There are two motivations for this:
1) calcdeps.py is deprecated and replaced by closurebuilder.py.
2) As part of this I was able to tweak things so that the Closure
compiler does not attempt to examine every .js file in the tree under
js/. This makes it possible to put compatibility tests and related files
in a subdirectory without them getting mixed up with the main .js files
we care about.

### DIFF
--- a/js/commonjs/export.js
+++ b/js/commonjs/export.js
@@ -5,6 +5,10 @@
  * the google-protobuf.js file that we build at distribution time.
  */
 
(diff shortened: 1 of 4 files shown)


acozzette:
ok to test

acozzette:
@haberman to review

haberman:
LGTM.

PR #2775, opened 2017-03-01
Undef major/minor if they are defined as macro.

xfxyjwf:
Fixes https://github.com/google/protobuf/issues/2695

We use major/minor as field names in plugin.proto to represent version numbers and these names are unfortunately defined as macros by some header files in some platforms. This PR adds the necessary #undef's to make the code compile on these platforms.


xfxyjwf:
Ping @pherl 

This is needed to fix protobuf build on freebsd as well:
https://github.com/google/protobuf/issues/2757

PR #2778, opened 2017-03-01
Patched a move-constructor into the generated code

os12:
- the functionality exists already, this patch just plugs it into
   C++11 move semantics
 - this way a message can be std::moved into a function (i.e. pass by value)


bazel-io:

xfxyjwf:
The proposal for adding move semantics to protobuf hasn't been approved yet (it's an Google internal process). Before it's approved we will not accept such PRs.

PR #2784, opened 2017-03-02
Return uint32 from Log2FloorNonZero64

acozzette:
A uint32 is big enough to hold any return value from that function, and
doing it this way prevents compiler warnings in coded_stream.h about
narrowing a uint64 to a uint32.


acozzette:
@teboring to review

TeBoring:
LGTM

PR #2786, opened 2017-03-02
Avoid self-swap check when swapping repeated fields.

ckennelly:
In generated code, we check for self-swap in Swap (and same Arena, for protos using Arenas) and then individually check each field for self-swap in `RepeatedField<T>::UnsafeArenaSwap`/`RepeatedPtrField<T>::UnsafeArenaSwap`.  If the first check has passed, the second (per-field) check will as well.

  `this != other` implies `&(this->some_field_) != &(other->some_field_)`

  for the same types of `this` and `other`.


bazel-io:

TeBoring:
This change has been adopted by other change.

Issue #1212, opened 2016-02-04
Detect descriptor.proto changes and fail if C# needs regenerating


jskeet:
If descriptor.proto has changed and the C++ hasn't been regenerated, we get a continuous build failure. The same should be true for C#.

Ideally, we should do the same for all our generated code (including wrapper types), but at least descriptor.proto.

This probably involves:
- Editing `generate_descriptor_protos.sh` to call the C# file, just like Objective C
- Detecting the failure in `travish.sh` or wherever the C++ code does it. (Will need to investigate further.)

Note that this does _not_ involve running any C# code - just protoc.

jskeet:
We should detect the conformance test file changing as well as the descriptor proto, and ideally the AddressBook sample too.

PR #2788, opened 2017-03-03
Detect if Descriptor.cs changes for csharp

anandolee:
related to #1212


bazel-io:

src/google/protobuf/compiler/csharp/csharp_bootstrap_unittest.cc
```diff
@@ -0,0 +1,142 @@
...
+
+  void ExpectFileMatches(const string& virtual_filename,
+                         const string& physical_filename) {
+    string* expected_contents = FindPtrOrNull(files_, virtual_filename);
+    ASSERT_TRUE(expected_contents != NULL)
+      << "Generator failed to generate file: " << virtual_filename;
+
+    string actual_contents;
+    GOOGLE_CHECK_OK(
+        File::GetContents(TestSourceDir() + "/" + physical_filename,
```
jskeet: Does this handle differences in line endings? I'm wondering what happens if the repo is cloned on a Windows box and we generate Unix line endings. It could end up depending on the repo settings. Maybe that isn't an issue?

anandolee: I guess that is not an issue. This is exactly same with what cpp does for descriptor.pb.h(cc)

anandolee:
@jskeet is this PR good enough to merge?

anandolee:
Yes, I've tested.

Issue #2605, opened 2017-01-18
JS extensions binary serialization


pgn69:
In JS, binary serialization of a proto object with an extension doesn't work if the value of extension evaluates to false, e.g. if it's an int32 with value zero.
In this case the extension is not serialized.

Problem seems to be in message.js function jspb.Message.serializeBinaryExtensions:

```
    var value = getExtensionFn.call(proto, fieldInfo);
    if (value) {
      .... serialize the extension value ....

```
The patch could be:

```
    var value = getExtensionFn.call(proto, fieldInfo);
    if (value !== undefined) {
      .... serialize the extension value ....

```

acozzette:
Are you using proto3? If so then this is working as intended, because proto3 does not make a distinction between a value being empty and being set to its default value.

pgn69:
I'm using protoc 3.1.0, but in my proto file I declared syntax = "proto2".

acozzette:
I see, I think you're right. Do you want to send a pull request with the fix?

pgn69:
No, I'm not so confident with this project and with proto3.
My fix seems to work with proto2 but I'm not sure it would work for proto3.

acozzette:
This should now be fixed, thanks again for reporting this @pgn69.

PR #2794, opened 2017-03-06
JS: ensure that extension values are serialized even if they're falsy

acozzette:
There was a bug where for JavaScript we would only serialize an
extension value if it evaluated as truthy, which meant that values like
0 would get silently dropped (even in proto2, where field presence is
significant). This fixes issue #2605, and takes care of the output of
toObject() in addition to the binary format.


acozzette:
@haberman to review

haberman:
LGTM.

PR #2795, opened 2017-03-07
Rename encode/decode to serializeToString/mergeFromString

TeBoring:
This better shows the semantic of the API. For already setted fields,
mergeFromString do replacement for singular fields and appending for
repeated fields.


bazel-io:

TeBoring:
@stanley-cheung for review.

TeBoring:
It will be in our next release, which is 3.3.0.

Issue #2783, opened 2017-03-02
Ruby gem 3.2.0 fails to load on Centos 6 and Debian 7 due to missing GLIBC_2.14


jacobvosmaer:
https://gitlab.com/gitlab-org/omnibus-gitlab/issues/2042

We get this error when trying to load the google-protobuf 3.2.0 gem on Debian 7 and Centos 6:

```
LoadError: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.14' not found (required by /opt/gitlab/embedded/lib/ruby/gems/2.3.0/gems/google-protobuf-3.2.0-x86_64-linux/lib/google/protobuf_c.so) - /opt/gitlab/embedded/lib/ruby/gems/2.3.0/gems/google-protobuf-3.2.0-x86_64-linux/lib/google/protobuf_c.so
```

I wonder if this is a reincarnation of #354 #781 #969.

jacobvosmaer:
As a workaround you can install the gem from source. I documented that process in https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9665/diffs

jacobvosmaer:
The gem for protobuf 3.1.0 also has the GLIBC_2.14 problem.

jacobvosmaer:
We found a simpler workaround.

```
# get a .gem without platform-specific binaries
gem fetch google-protobuf -v 3.2.0 --platform=
gem install google-protobuf-3.2.0.gem --local
```

So a working release _is_ being pushed to rubygems, but `gem install google-protobuf` will grab the linux-amd64 one even if it does not work (due to GLIBC).

acozzette:
@jacobvosmaer  It sounds like we might just need to build for an older glibc version; what version do you have?

jacobvosmaer:
We (GitLab) ship to Debian 7 and Centos 6. Those seem to use gblic 2.13 and 2.12, respectively.

So 2.12 should do it.

```
[jacobvosmaer@centos6-test ~]$ ldd --version
ldd (GNU libc) 2.12
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

```
jacobvosmaer@debian7-test:~$ ldd --version
ldd (Debian EGLIBC 2.13-38+deb7u10) 2.13
[...]

PR #2804, opened 2017-03-08
Ruby: wrap calls to memcpy so that gem is compatible with pre-2.14 glibc

acozzette:
This commit adds a __wrap_memcpy function and a linker flag to use that
in place of memcpy for our Ruby gem C extension. This allows us to
always use the 2.2.5 version of memcpy, making it possible to use the
gem on distributions with pre-2.14 versions of glibc.

Before this change:
$ objdump -T protobuf_c.so | grep memcpy
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.3.4 __memcpy_chk
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.14  memcpy

After:
$ objdump -T protobuf_c.so | grep memcpy
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.2.5 memcpy
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.3.4 __memcpy_chk
0000000000042450 g    DF .text  0000000000000005  Base __wrap_memcpy

This is based on gRPC's solution to a similar problem:
https://github.com/grpc/grpc/blob/5098508d2d41a116113f7e333c516cd9ef34a943/src/core/lib/support/wrap_memcpy.c
However, we have not so far run into any issues with callers using
memcpy when they should be using memmove, so in this commit I did not
add anything to redirect memcpy to memmove.

This fixes issue #2783.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -795,6 +795,7 @@ ruby_EXTRA_DIST=                                                             \
   ruby/ext/google/protobuf_c/storage.c                                       \
   ruby/ext/google/protobuf_c/upb.c                                           \
   ruby/ext/google/protobuf_c/upb.h                                           \
+  ruby/ext/google/protobuf_c/wrap_memcpy.c                                   \
   ruby/google-protobuf.gemspec                                               \
   ruby/lib/google/protobuf/message_exts.rb                                   \
   ruby/lib/google/protobuf/repeated_field.rb                                 \
--- a/ruby/ext/google/protobuf_c/extconf.rb
+++ b/ruby/ext/google/protobuf_c/extconf.rb
@@ -4,7 +4,14 @@ require 'mkmf'
 
 $CFLAGS += " -std=c99 -O3 -DNDEBUG"
 
+
+if RUBY_PLATFORM =~ /linux/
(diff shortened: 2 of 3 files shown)


acozzette:
@haberman @nicolasnoble Does this look ok?

nicolasnoble:
The point of the memmove call is to provide an alternative to the "x86_64" case; if someone builds that code on linux that isn't Intel 64 bits, the resolution of the _wrap_memcpy symbol won't happen and you'll get a linker error.

nicolasnoble:
Also, you should only add the LDFLAGS if the OS is Linux.

acozzette:
Thanks @nicolasnoble, I went ahead and made those fixes. I was thinking I would actually prefer to use `__real_memcpy` (which points to the "real" memcpy) in the Linux, non-x86-64 case, but I found that wrapping `__real_memcpy` mysteriously causes [this test case](https://github.com/google/protobuf/blob/7f3e23707122f31ebfa58b7280bd56cbe77cb44e/ruby/tests/basic.rb#L967) to segfault. I'd like to figure out why that is, but for now I just set it up to use `memmove`.

haberman:
LGTM.

PR #2811, opened 2017-03-09
Java: Mark sun.misc dependency as optional

Dunemaster:
Mark sun.misc dependency as optional (which it is).
The pull request makes protobuf 3 usage possible in osgi environments where sun.misc is not availabe (which is by default)

### DIFF
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -132,6 +132,7 @@
             <Bundle-DocURL>https://developers.google.com/protocol-buffers/</Bundle-DocURL>
             <Bundle-SymbolicName>com.google.protobuf</Bundle-SymbolicName>
             <Export-Package>com.google.protobuf;version=${project.version}</Export-Package>
+            <Import-Package>sun.misc;resolution:=optional,*</Import-Package>
           </instructions>
         </configuration>
       </plugin>


bazel-io:

Dunemaster:
I signed it!

Dunemaster:
The commit has been authored by me

acozzette:
ok to test (this will trigger the other CI bots to run)

acozzette:
@Dunemaster Sorry for the delay getting back to you on this. I asked a few other people within Google who are working on Java protobuf about this, and they told me that while we're not currently relying on sun.misc.Unsafe, we will probably eventually end up requiring it because of a change in the works that will make the Java implementation table-driven. Because of that we're a little reluctant to take a pull request that makes sun.misc optional, given that it will probably need to become mandatory in the future. But could you provide some more details on the situation so that we can try to work out a solution?

Dunemaster:
@acozzette, does it mean that android will not be supported by full version of protobuf in future?

xfxyjwf:
@Dunemaster I believe full version of protobuf already doesn't work well on Android because of its usage of Java reflection. We recommend Android users to use protobuf lite runtime and that's becoming mandatory inside Google.

xfxyjwf:
@Dunemaster Our internal discussion suggests that we may have to require sun.misc.Unsafe for both full runtime and lite runtime. Are you suggesting sun.misc.Unsafe is not available on Android? If so we will need to sort out a solution soon. @nmittler

Dunemaster:
@xfxyjwf , sun.misc is available on most android devices, but it is even more unsupported that in HotSpot.   Code, using sun.misc will not compile against android sdk, but in fact the class is available in runtime and can be access through reflection, for example

xfxyjwf:
Merging this change for now. Will need to revisit this if sun.misc becomes mandatory in Java full runtime.

PR #2815, opened 2017-03-09
Ruby version optionally emits default values in JSON encoding.

### LINKED ISSUE #1216: Ruby client should be able to be configured to emit default fields in JSON
Much like the Java client's recent ability to emit default fields in JSON (as of [v3.0.0-beta2](https://github.com/google/protobuf/releases/tag/v3.0.0-beta-2)), I would personally like to be able to do the same in ruby.  While decoding and encoding protobufs on their own can provide us with default values, sometimes its necessary for these default values to exist in JSON bodies.  

Perhaps the `encode_json` method could be altered such that it could accept an additional, optional, configuration hash that could indicate the desire to emit default values.  Maybe something like this?

``` ruby
MyWireObj.encode_json(msg, {include_default_values: true})
```

Thanks!

devwout:
Usage: `Message.encode_json(m, emit_defaults: true)`
Message fields that are nil will still not appear in the encoded JSON.

This should fix #1216 .

Behavior should be in line with the Golang and Python versions. Singular primitive fields, repeated fields and map fields will always be serialized. This option does not affect fields of type message.

Chose the `emit_defaults` name from the Golang implementation. Python's option `including_default_value_fields` it a lot longer and does not seem to describe it better.

### DIFF
--- a/ruby/tests/basic.rb
+++ b/ruby/tests/basic.rb
@@ -1,6 +1,7 @@
 #!/usr/bin/ruby
 
 require 'google/protobuf'
+require 'json'
 require 'test/unit'
 
 # ------------- generated code --------------
@@ -1179,9 +1180,33 @@ module BasicTest
       return if RUBY_PLATFORM == "java"
       m = TestMessage.new
 
-      expected = '{"optionalInt32":0,"optionalInt64":0,"optionalUint32":0,"optionalUint64":0,"optionalBool":false,"optionalFloat":0,"optionalDouble":0,"optionalString":"","optionalBytes":"","optionalEnum":"Default","repeatedInt32":[],"repeatedInt64":[],"repeatedUint32":[],"repeatedUint64":[],"repeatedBool":[],"repeatedFloat":[],"repeatedDouble":[],"repeatedString":[],"repeatedBytes":[],"repeatedMsg":[],"repeatedEnum":[]}'
+      expected = {
+        optionalInt32: 0,
+        optionalInt64: 0,
+        optionalUint32: 0,
+        optionalUint64: 0,
+        optionalBool: false,
+        optionalFloat: 0,
+        optionalDouble: 0,
+        optionalString: "",
+        optionalBytes: "",
+        optionalEnum: "Default",
+        repeatedInt32: [],
+        repeatedInt64: [],
+        repeatedUint32: [],
+        repeatedUint64: [],
+        repeatedBool: [],
+        repeatedFloat: [],
+        repeatedDouble: [],
+        repeatedString: [],
+        repeatedBytes: [],
+        repeatedMsg: [],
+        repeatedEnum: []
+      }
 
-      assert TestMessage.encode_json(m, :emit_defaults => true) == expected
+      actual = TestMessage.encode_json(m, :emit_defaults => true)
+
+      assert JSON.parse(actual, :symbolize_names => true) == expected
     end
 
     def test_json_emit_defaults_submsg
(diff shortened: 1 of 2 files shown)


bazel-io:

devwout:
I signed it

ruby/tests/basic.rb
```diff
@@ -1174,6 +1174,36 @@ def test_json
       Foo.encode_json(Foo.new(bar: bar, baz: [baz1, baz2]))
     end
 
+    def test_json_emit_defaults
+      # TODO: Fix JSON in JRuby version.
+      return if RUBY_PLATFORM == "java"
+      m = TestMessage.new
+
+      expected = '{"optionalInt32":0,"optionalInt64":0,"optionalUint32":0,"optionalUint64":0,"optionalBool":false,"optionalFloat":0,"optionalDouble":0,"optionalString":"","optionalBytes":"","optionalEnum":"Default","repeatedInt32":[],"repeatedInt64":[],"repeatedUint32":[],"repeatedUint64":[],"repeatedBool":[],"repeatedFloat":[],"repeatedDouble":[],"repeatedString":[],"repeatedBytes":[],"repeatedMsg":[],"repeatedEnum":[]}'
```
haberman: Please format this to match other large hash literals in the file.

ruby/tests/basic.rb
```diff
@@ -1174,6 +1174,36 @@ def test_json
...
+
+      assert TestMessage.encode_json(m, :emit_defaults => true) == expected
+    end
+
+    def test_json_emit_defaults_submsg
+      # TODO: Fix JSON in JRuby version.
+      return if RUBY_PLATFORM == "java"
+      m = TestMessage.new(optional_msg: TestMessage2.new)
+
+      expected = '{"optionalInt32":0,"optionalInt64":0,"optionalUint32":0,"optionalUint64":0,"optionalBool":false,"optionalFloat":0,"optionalDouble":0,"optionalString":"","optionalBytes":"","optionalMsg":{"foo":0},"optionalEnum":"Default","repeatedInt32":[],"repeatedInt64":[],"repeatedUint32":[],"repeatedUint64":[],"repeatedBool":[],"repeatedFloat":[],"repeatedDouble":[],"repeatedString":[],"repeatedBytes":[],"repeatedMsg":[],"repeatedEnum":[]}'
```
haberman: Ditto.

haberman:
Looks great! Just a few minor formatting things.

ruby/tests/basic.rb
devwout: This is a string, not a hash literal. There's not much freedom to format as it should match the output exactly. I don't really see any examples of large string literals in the file, except in `test_inspect`.

Should I...
a) Use `<<STR` and remove all whitespace.
b) Parse the JSON string and compare with a hash literal, which can be nicely formatted.

haberman: Ah, my bad, I didn't read closely enough. I don't think we should use a string literal for the comparison, because it will depend on the order of the fields, but this order could change. Perhaps use the Ruby JSON library to parse the output and then assert that it is equal to an expected Ruby object?

haberman:
ok to test

devwout:
I modified the JSON tests (including one that already existed) to compare parsed values instead of the raw JSON string. I used the more compact ruby hash representation and the `symbolize_names` option. If desired, I can switch to the more verbose `"key" => val` or `:key => val` with `symbolize_names`.

devwout:
Is there anything else I can do to get this approved? I see Jenkins is failing for a completely different reason...

devwout:
No problem, thanks for the quick response!

Issue #2686, opened 2017-02-07
PHP: Repeated field setter should accept a regular PHP array


pauln:
### The problem
Currently, in order to set a repeated field in a message, you seem to have two choices - neither of which is particularly nice, especially if the repeated field is itself a message.  It would be much cleaner if the generated PHP setter would accept a regular PHP array of objects of the relevant type (in this case, the applicable message type) and internally convert them to a `RepeatedField`.

The [docs for generated PHP code](https://developers.google.com/protocol-buffers/docs/reference/php-generated#repeated-fields) appear to suggest that it's possible to provide a regular array (`$m->setFoo($array);`), but this doesn't seem to be the case - at least with Protobuf v3.1.0-alpha-1 and a repeated message (perhaps other types of repeated field do indeed accept arrays - I haven't had cause to try).

### The current options:
As far as I can see, these are the only options at present - sorry if I've missed something.

- **Create a `RepeatedField` yourself**
```
$request = new \MyNS\MyMessage();
$customWidgets = new Google\Protobuf\Internal\RepeatedField(Google\Protobuf\Internal\GPBType::MESSAGE, \MyNS\MyMessage_Submessage::class);
$widget= new \MyNS\MyMessage_Submessage();
$widget->setMyProperty('foo');
$customWidgets[] = $widget;
$request->setCustomWidgets($customWidgets);
```
This is messy, requiring direct usage of "internal" GPB classes and providing the type and class (which are already known by the generated Message class - as evidenced by its setter's call to `GPBUtil::checkRepeatedField`, which uses these exact parameters); additionally, although the `RepeatedField` can be indexed like a regular array, its elements are immutable - which restricts you to fully composing your submessage and then appending it to the `RepeatedField` (which isn't necessarily a bad thing, but is definitely a trap to watch out for).
- **Use the getter to set your repeated fields**
```
$request = new \MyNS\MyMessage();
$customWidgets = $request->getCustomWidgets();
$widget= new \MyNS\MyMessage_Submessage();
$widget->setMyProperty('foo');
$customWidgets[] = $widget;
// No call to setter required!
```
This is just plain weird, as you're using a getter in order to retrieve a reference to an (initially, but not necessarily!) empty `RepeatedField`, then appending submessages to it - without even needing to call the setter (although you can if you really want to, it's not necessary due to the getter returning a reference to the **private** property, rather than a new `RepeatedField` object).

### Potential solutions
[...]

PR #2817, opened 2017-03-09
Repeated/Map field setter should accept a regular PHP array

TeBoring:
Accept regular PHP array for repeated/map setter. Existing map/repeated
field will be swapped by a clean map/repeated field. Then, elements in
the array will be added to the map/repeated field. All elements will be
type checked before adding.

See #2686 for detail.


bazel-io:

TeBoring:
@stanley-cheung to review

jessedobbelaere:
@TeBoring This is currently in the [3.3.x-dev](https://packagist.org/packages/google/protobuf#3.3.x-dev) release if I see it correctly? Is there an ETA? 
I'd like to use this in a production application so can't use the dev version.

PR #2820, opened 2017-03-09
fixed js_embed binary not found

udnaan:
The build fails on linux/ubuntu with the error js_embed not found. This fixes it


bazel-io:

PR #2822, opened 2017-03-09
Detect generated code of WKT, addressbook and conformance protos

### LINKED ISSUE #1212: Detect descriptor.proto changes and fail if C# needs regenerating

anandolee:
Add auto detect for generated code of WKT protos, addressbook.proto and conformance.proto

fixes #1212

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_bootstrap_unittest.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_bootstrap_unittest.cc
@@ -34,8 +34,8 @@
 // generated automatically at build time.
 //
 // If this test fails, run the script
-// "generate_descriptor_proto.sh" and add
-// csharp/src/Google.Protobuf/Reflection/Descriptor.cs to your changelist.
+// "generate_descriptor_proto.sh" and add the changed files under
+// csharp/src/ to your changelist.
 
 #include <map>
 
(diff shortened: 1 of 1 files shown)


bazel-io:

anandolee:
I'd prefer only add .proto files that are stable into the detect test. We'd better not add unittest .proto files into it which may both other languages too much. Keep these parameters in hard code makes the adding more serious and formal.

jskeet:
So long as people know to make the change, of course. I agree it should be relatively rare. Do we have a document somewhere saying "When you add a WKT or similar, you should change X files; when you modify an existing WKT you should rerun Y generation." ?

anandolee:
It is mentioned in the comments of this file:

// If this test fails, run the script
// "generate_descriptor_proto.sh" and add the changed files under
// csharp/src/ to your changelist.

Is it enough?

jskeet:
That's okay for knowing what to do if the test fails - it's more when the tests should be expanded. But that can definitely come later.

PR #2829, opened 2017-03-10
Hide unnecessary exported library symbols

afrantzis:
The protobuf libraries currently export many symbols not useful for protobuf consumers. Although this is not usually an issue, there are some use cases where the exported symbols cause problems.

Such a problematic use case occurs due to weak symbols from C++ templates that are exported by protobuf libraries. Other libraries (including libstdc++) can bind to such symbols and prevent protobuf from properly unloading (and subsequently reloading) when protobuf itself or a library linking to it is loaded/unloaded with dlopen/dlclose. An example of such a binding from a real world example is:

binding file /usr/lib/x86_64-linux-gnu/libstdc++.so.6 [0] to /usr/lib/x86_64-linux-gnu/libprotobuf-lite.so.10 [0]: normal symbol `_ZNKSt5ctypeIcE8do_widenEc' [GLIBCXX_3.4]

This PR uses version scripts (--Wl,--version-script) to hide unnecessary symbols on platforms that support it , therefore solving problems like the one described above. Plus it's good practice to export only required symbols from libraries.

Note that the -export-symbols(-regex) libtool flag cannot be used in this case, since, at least when using gnu ld, it is implemented with --Wl,--retain-symbols-file which does not affect all the types of symbols we need to hide.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -93,6 +93,25 @@ ACX_CHECK_SUNCC
 # to the link
 AC_PROG_LIBTOOL
 
+# Check whether the linker supports version scripts
+AC_MSG_CHECKING([whether the linker supports version scripts])
+save_LDFLAGS=$LDFLAGS
+LDFLAGS="$LDFLAGS -Wl,--version-script=conftest.map"
+cat > conftest.map <<EOF
+{
+  global:
+    main;
+  local:
+    *;
+};
+EOF
+AC_LINK_IFELSE(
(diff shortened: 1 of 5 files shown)


bazel-io:

afrantzis:
I update the commit to fix the CI failure.

acozzette:
ok to test (this will kick off the Jenkins run)

acozzette:
@afrantzis Thanks for the pull request!

PR #2831, opened 2017-03-10
Cherry-picked Ruby fix and incremented version number

acozzette:
This pull request cherry-picks the fix for Ruby in 9fa40314fcfd1 and increments the gem version number to 3.2.0.1 so that we can publish new gems with that fix. This is part of issue #2783.


acozzette:
@TeBoring to review

TeBoring:
LGTM

PR #2842, opened 2017-03-13
Add property to the C#-JsonParser to ignore unknown fields during parsing

### LINKED ISSUE #2838: JsonParser will not ignore unknown field when parsing - C#
Like the closed Issue #1610 for Java, the C# JsonParser doesn't ignore unknown fields during parsing.
If an unknown object is encountered, it throws an Exception: 

```
if (jsonFieldMap.TryGetValue(name, out field))
{
	[...]
}
else
{
	// TODO: Is this what we want to do? If not, we'll need to skip the value,
	// which may be an object or array. (We might want to put code in the tokenizer
	// to do that.)
	throw new InvalidProtocolBufferException("Unknown field: " + name);
}
```

It would be great if there would exists a similar option as got added to the Java Parser (`ignoringUnknownFields()`).

Genmutant:
First version to fix #2838 . How should the JsonReplayTokenizer be handled?

It should ignore the object if it is unknown. To achieve this, the name and value is skipped if it is a simple value. If it is an array or dict, the tokens are read and discarded until the object is finished -> it has a lower remaining object-stack.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
@@ -926,6 +926,62 @@ namespace Google.Protobuf
             Assert.Throws<InvalidProtocolBufferException>(() => TestAllTypes.Parser.ParseJson(json));
         }
 
+        [Test]
+        public void UnknownField_Integer()
+        {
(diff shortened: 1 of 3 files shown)


bazel-io:

Genmutant:
I signed it!

anandolee:
Fixed in #3790

PR #2848, opened 2017-03-14
Fix freebsd build.

xfxyjwf:
Fixes https://github.com/google/protobuf/issues/2757

It turns out system headers included by generated plugin.pb.h file already contains
major/minor macro definitions when built on FreeBSD and we need to add #undefs to
the generated header file.

This change also fixes another compile error regarding EXPECT_DEATH on FreeBSD.


xfxyjwf:
retest this please

xfxyjwf:
ping @pherl

PR #2850, opened 2017-03-15
Avoid aliasing from._has_bits_ in MergeFrom

ckennelly:
Self-`MergeFrom` has been forbidden since 40ee5517 via a `GOOGLE_CHECK`.  Since 5a76e633e, we use a `GOOGLE_DCHECK` to guard against it.  The compiler forces a reload of the `from._has_bits_` after every field is merged.


bazel-io:

PR #2852, opened 2017-03-15
Defer updating local _has_bits_ in MergeFrom.

ckennelly:
This can save us up to 7 load/or/stores per 8 fields, as the compiler may assume that `&from` and `this` alias each other, so the local `_has_bits_` need to be reloaded with each field.

This extends #2850.


bazel-io:

PR #2869, opened 2017-03-17
Update commit number in Docker to update composer dependency

TeBoring:


bazel-io:

TeBoring:
@stanley-cheung

PR #2884, opened 2017-03-21
Add FormatEnumsAsIntegers support for Json Format. 

### LINKED ISSUE #1948: C#: Enum serialization/deserialization issue
Enum values are serialized as string rather than numeric value. Is there a way to support numeric value.

anandolee:
#Add FormatEnumsAsIntegers support for Json Format. 
This change also scale JsonFormatter.Settings for multi options.

fixes #1948


bazel-io:

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -758,31 +764,42 @@ public sealed class Settings
...
+            /// Whether to format enums as ints. Defaults to false.
+            /// </summary>
+            public bool FormatEnumAsInt { get; set;}
 
             // TODO: Work out how we're going to scale this to multiple settings. "WithXyz" methods?
 
             /// <summary>
             /// Creates a new <see cref="Settings"/> object with the specified formatting of default values
-            /// and an empty type registry.
+            /// and an empty type registry. This method is deprecated.
```
jskeet: Please add the `[Obsolete]` attribute here and below, if we still deprecated these. (And call them constructors rather than methods.)

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -758,31 +764,42 @@ public sealed class Settings
...
             {
-                Default = new Settings(false);
             }
 
             /// <summary>
             /// Whether fields whose values are the default for the field type (e.g. 0 for integers)
             /// should be formatted (true) or omitted (false).
             /// </summary>
-            public bool FormatDefaultValues { get; }
+            public bool FormatDefaultValues { get; set; }
```
jskeet: I really don't like making these settings mutable. This means that what one library does with the default settings will be visible to everything else in the application. Aside from anything else, changing something from being immutable to being mutable feels like a breaking change to me: users can no longer rely on something they used to be able to rely on.

If we think the settings are getting too complicated to have lots of constructors, we can introduce a builder, but let's keep the settings themselves immutable.

My preference would be to just add a third constructor, but make all the parameters optional:

```csharp
public Settings(bool formatDefaultValues = false, TypeRegistry typeRegistry = null, formatEnumAsInteger = false)
{
  // Normal code, using typeRegistry ?? TypeRegistry.Empty so that null means default
}
```

I realize that allowing a null value here but not elsewhere is inconsistent. Maybe we should change the other constructor to allow null as well. (That way feels like it's not a breaking change.)

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -758,31 +764,42 @@ public sealed class Settings
...
+                get { return typeRegistry_ ; }
+                set {
+                    typeRegistry_ = ProtoPreconditions.CheckNotNull(value, nameof(value));
+                }
+            }
+
+            /// <summary>
+            /// Whether to format enums as ints. Defaults to false.
+            /// </summary>
+            public bool FormatEnumAsInt { get; set;}
```
jskeet: I'd expand `Int` to `Integer` here; in .NET APIs normally you'd refer to `Int32` if you mean that specified type, or `Integer` to refer to integers in general.

Also, to be consistent with `FormatDefaultValues`, I'd be tempted to use the plural again: `FormatEnumsAsIntegers`

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -375,14 +375,20 @@ public void WriteValue(TextWriter writer, object value)
             }
             else if (value is System.Enum)
             {
-                string name = OriginalEnumValueHelper.GetOriginalName(value);
-                if (name != null)
-                {
-                    WriteString(writer, name);
+                if (settings.FormatEnumAsInt) {
```
jskeet: Brace at start of line, consistent with the rest of the code.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -758,31 +764,42 @@ public sealed class Settings
...
             /// </summary>
-            public bool FormatDefaultValues { get; }
+            public bool FormatDefaultValues { get; set; }
 
+            private TypeRegistry typeRegistry_;
             /// <summary>
             /// The type registry used to format <see cref="Any"/> messages.
             /// </summary>
-            public TypeRegistry TypeRegistry { get; }
+            public TypeRegistry TypeRegistry {
```
jskeet: Brace formatting. (The "get" is fine as it is.) Although I'm hoping all of this goes away by it becoming immutable again :)

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: make sense to me

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: done

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -778,27 +785,50 @@ static Settings()
...
                 FormatDefaultValues = formatDefaultValues;
                 TypeRegistry = ProtoPreconditions.CheckNotNull(typeRegistry, nameof(typeRegistry));
+                FormatEnumsAsIntegers = false;
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified parameters.
+            /// </summary>
+            /// <param name="formatDefaultValues"><c>true</c> if default values (0, empty strings etc) should be formatted; <c>false</c> otherwise.</param>
+            /// <param name="typeRegistry">The <see cref="TypeRegistry"/> to use when formatting <see cref="Any"/> messages.</param>
```
jskeet: Possibly add a comment explaining what happens if this is null.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -778,27 +785,50 @@ static Settings()
...
             }
 
             /// <summary>
             /// Creates a new <see cref="Settings"/> object with the specified formatting of default values
-            /// and type registry.
+            /// and type registry. This constructor is deprecated.
             /// </summary>
             /// <param name="formatDefaultValues"><c>true</c> if default values (0, empty strings etc) should be formatted; <c>false</c> otherwise.</param>
             /// <param name="typeRegistry">The <see cref="TypeRegistry"/> to use when formatting <see cref="Any"/> messages.</param>
+            [System.Obsolete]
```
jskeet: Possibly make this constructor just chain to the other one, just as the one above chains to this?

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -778,27 +785,50 @@ static Settings()
...
+            /// Whether to format enums as ints. Defaults to false.
+            /// </summary>
+            public bool FormatEnumsAsIntegers { get; }
+
             // TODO: Work out how we're going to scale this to multiple settings. "WithXyz" methods?
 
             /// <summary>
             /// Creates a new <see cref="Settings"/> object with the specified formatting of default values
-            /// and an empty type registry.
+            /// and an empty type registry. This constructor is deprecated.
```
jskeet: I don't think we really need to deprecate these constructors, to be honest - there's nothing wrong with them, really. (I don't think we want to end up adding warnings for entirely valid code which will keep working.)

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
```diff
@@ -309,7 +329,8 @@ public void WrapperFormatting_IncludeNull()
         [Test]
         public void OutputIsInNumericFieldOrder_NoDefaults()
         {
-            var formatter = new JsonFormatter(new JsonFormatter.Settings(false));
+            var formatter = new JsonFormatter(new JsonFormatter.Settings(
+                formatDefaultValues: true));
```
jskeet: This should be false.

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
```diff
@@ -300,7 +319,8 @@ public void WrapperFormatting_IncludeNull()
         {
             // The actual JSON here is very large because there are lots of fields. Just test a couple of them.
             var message = new TestWellKnownTypes { Int32Field = 10 };
-            var formatter = new JsonFormatter(new JsonFormatter.Settings(true));
+            var formatter = new JsonFormatter(new JsonFormatter.Settings(
```
jskeet: Are you making these changes just for the sake of clarity, using a named argument? I think it's still using the obsolete constructor.

anandolee: I think we may want to add more parameters in the future. It will be good to make the parameters clear

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: done

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: We will add more parameters I guess and I don't want to make set default different from others. However, adding warnings for entirely valid code is not good. Maybe we can keep what they are...

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
jskeet: Yup, that's fine. I'd probably do it all on one line, but that's all.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -795,10 +807,23 @@ public Settings(bool formatDefaultValues) : this(formatDefaultValues, TypeRegist
...
+            {
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified parameters.
+            /// </summary>
+            /// <param name="formatDefaultValues"><c>true</c> if default values (0, empty strings etc) should be formatted; <c>false</c> otherwise.</param>
+            /// <param name="typeRegistry">The <see cref="TypeRegistry"/> to use when formatting <see cref="Any"/> messages. TypeRegistry.Empty will be used if it is null.</param>
+            /// <param name="formatEnumsAsIntegers"><c>true</c> format the enums as ints; <c>false</c> otherwise.</param>
+            public Settings(bool formatDefaultValues = false,
```
jskeet: I've been thinking about this, and we end up with a problem around optional parameters. Currently, if someone calls

```csharp
new Settings(formatEnumsAsIntegers: true)
```

then that's fine - but if we add another constructor, overload resolution will fail.

Basically, this is all slightly messy, but it might be worth not having defaults, but have "With" methods instead, so we have:

- 3x constructor (and more added as required), no optional parameters
- One "With" method per setting, e.g. WithFormatDefaultValues, WithTypeRegistry etc.

For the sake of avoiding dragging this PR out too much, if you *just* remove the defaults here, I can mail another PR to add the With methods. How does that sound?

anandolee: Sorry I don't quite understand.

Why we need to add another constructor if we use optional parameters? Shouldn't we just modify this constructor to Settings(bool formatDefaultValues = false, TypeRegistry typeRegistry = null, bool formatEnumsAsIntegers = false, bool theAddedOption = false)?

What is the "With" method? Are users only able the set one parameters if they want to use "With" method?

jskeet: If you change that constructor, that breaks binary compatibility.
If you add a constructor, it breaks source compatibility.

Each `With` method would return a new `Settings` object - so clients *could* still call the constructors, or they could use

```csharp
var settings = Settings.Default.WithFormatDefaultValues(true).WithTypeRegistry(registry);`
```

(for example)

anandolee: With methods sounds good to me.
How about we only add With methods instead of more constructors (delete this constructor)?

jskeet: Sure. We'll need a private constructor, but that can change signature over time of course.

anandolee: Can you make this change first or I can do it in this PR as well

jskeet: I suggest you do it in this PR - I can't add a `WithFormatEnumsAsIntegers` until there's such a constructor parameter, for example. I could do it for the other two, but it's really simple:

```csharp
// TODO: Documentation
public Settings WithFormatDefaultValues(bool formatDefaultValues) =>
    new Settings(formatDefaultValues, TypeRegistry);

public Settings WithTypeRegistry(TypeRegistry typeRegistry) =>
    new Settings(FormatDefaultValues, typeRegistry);
```

anandolee: Changed according to our discussion

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -778,6 +785,11 @@ static Settings()
             /// </summary>
             public TypeRegistry TypeRegistry { get; }
 
+            /// <summary>
+            /// Whether to format enums as ints. Defaults to false.
+            /// </summary>
+            public bool FormatEnumsAsIntegers { get; }
+
```
jskeet: We can remove the line below this now - I hadn't spotted it earlier. At least it means I'm consistent :)

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -795,11 +807,42 @@ public Settings(bool formatDefaultValues) : this(formatDefaultValues, TypeRegist
...
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified parameters.
+            /// </summary>
+            /// <param name="formatDefaultValues"><c>true</c> if default values (0, empty strings etc) should be formatted; <c>false</c> otherwise.</param>
+            /// <param name="typeRegistry">The <see cref="TypeRegistry"/> to use when formatting <see cref="Any"/> messages. TypeRegistry.Empty will be used if it is null.</param>
+            /// <param name="formatEnumsAsIntegers"><c>true</c> format the enums as integers; <c>false</c> format enums as enum names.</param>
+            private Settings(bool formatDefaultValues,
+                            TypeRegistry typeRegistry,
+                            bool formatEnumsAsIntegers)
```
jskeet: Change `ProtoPreconditions.CheckNotNull(typeRegistry, nameof(typeRegistry))` to `typeRegistry ?? TypeRegistry.Empty`. I think it's fine for us to make it easy to get to that (and it matches the comment).

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -795,11 +807,42 @@ public Settings(bool formatDefaultValues) : this(formatDefaultValues, TypeRegist
...
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified type registry and the current settings.
+            /// </summary>
+            /// <param name="typeRegistry">The <see cref="TypeRegistry"/> to use when formatting <see cref="Any"/> messages.</param>
+            public Settings WithTypeRegistry(TypeRegistry typeRegistry) => new Settings(FormatDefaultValues, typeRegistry, FormatEnumsAsIntegers);
+
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified formatting enums as integers and the current settings.
+            /// </summary>
+            /// <param name="formatEnumsAsIntegers"><c>true</c> format the enums as integers; <c>false</c> format enums as enum names.</param>
```
jskeet: I'd insert "to" after each of the literals:

```text
<c>true</c> to format enum values as integers; <c>false</c> to format enum values as names
```

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -795,11 +807,42 @@ public Settings(bool formatDefaultValues) : this(formatDefaultValues, TypeRegist
...
+            public Settings WithFormatDefaultValues(bool formatDefaultValues) => new Settings(formatDefaultValues, TypeRegistry, FormatEnumsAsIntegers);
+
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified type registry and the current settings.
+            /// </summary>
+            /// <param name="typeRegistry">The <see cref="TypeRegistry"/> to use when formatting <see cref="Any"/> messages.</param>
+            public Settings WithTypeRegistry(TypeRegistry typeRegistry) => new Settings(FormatDefaultValues, typeRegistry, FormatEnumsAsIntegers);
+
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified formatting enums as integers and the current settings.
```
jskeet: "specified enum formatting option" perhaps?

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

csharp/src/Google.Protobuf/JsonFormatter.cs
anandolee: Done

Issue #2889, opened 2017-03-21
Custom option c#, 'DescriptorReflection' is inaccessible due to its protection level


plemanach:
Hi,

I get a compilation error : 'DescriptorReflection' is inaccessible due to its protection level
when I try to build this proto 

`syntax = "proto3";

import "google/protobuf/descriptor.proto";

// We don't put this in a package within proto2 because we need to make sure
// that the generated code doesn't depend on being in the proto2 namespace.
package protobuf_unittest;
option csharp_namespace = "UnitTest.Issues.TestProtos";

// Some simple test custom options of various types.

extend google.protobuf.EnumOptions {
  sfixed32 enum_opt1 = 7753576;
}

extend google.protobuf.EnumValueOptions {
  int32 enum_value_opt1 = 1560678;
}

// A test message with custom options at all possible locations (and also some
// regular options, to make sure they interact nicely).
message TestMessageWithCustomOptions {
  
  enum AnEnum {
    option (enum_opt1) = -789;
    ANENUM_UNSPECIFIED = 0;
    ANENUM_VAL1 = 1;
    ANENUM_VAL2 = 2 [(enum_value_opt1) = 123];
  }
}

`

anandolee:
Looks like this proto is in our tests and can work well. Are you able to create an PR to reproduce this issue?

plemanach:
I am not sure why kind of PR I will create the generated code does not compile :

`// Generated by the protocol buffer compiler.  DO NOT EDIT!
// source: unittest_custom_options_proto3.proto
#pragma warning disable 1591, 0612, 3021
#region Designer generated code

using pb = global::Google.Protobuf;
using pbc = global::Google.Protobuf.Collections;
using pbr = global::Google.Protobuf.Reflection;
using scg = global::System.Collections.Generic;
namespace UnitTest.Issues.TestProtos {

  /// <summary>Holder for reflection information generated from unittest_custom_options_proto3.proto</summary>
  public static partial class UnittestCustomOptionsProto3Reflection {

    #region Descriptor
    /// <summary>File descriptor for unittest_custom_options_proto3.proto</summary>
    public static pbr::FileDescriptor Descriptor {
      get { return descriptor; }
    }
    private static pbr::FileDescriptor descriptor;

    static UnittestCustomOptionsProto3Reflection() {
      byte[] descriptorData = global::System.Convert.FromBase64String(
          string.Concat(
[...]

PR #2890, opened 2017-03-21
Update Descriptor.cs

plemanach:
it will fix this issue  : https://github.com/google/protobuf/issues/2889


bazel-io:

csharp/src/Google.Protobuf/Reflection/Descriptor.cs
```diff
@@ -10,7 +10,7 @@
 namespace Google.Protobuf.Reflection {
 
   /// <summary>Holder for reflection information generated from google/protobuf/descriptor.proto</summary>
-  internal static partial class DescriptorReflection {
+  public static partial class DescriptorReflection {
```
anandolee: You can't modify this file directly as it is generated from compiler.

anandolee:
Closing for #2889 is closed

PR #2897, opened 2017-03-22
Compile the Java proto runtime with Java 6

cgrushko:
in order to avoid errors related to generics when 
building user code in newer versions of Java.


xfxyjwf:
Thanks!

PR #2898, opened 2017-03-22
Avoid aliasing _has_bits_ during SerializeWithCachedSizes.

ckennelly:


bazel-io:

PR #2906, opened 2017-03-25
Fix missing std::tr1::hash on GCC 4.1

matthauck:
Rather than crashing on use (doh!) better to just implement the
missing std::tr1::hash structs.

Fixes regression introduced by #1913


bazel-io:

xfxyjwf:
ok to test

xfxyjwf:
Closing this in favor of https://github.com/google/protobuf/pull/2907 because the hash function in this PR seems too weak.

PR #2907, opened 2017-03-25
Fix missing std::tr1::hash on GCC 4.1

matthauck:
Rather than crashing on use (doh!) better to just declare this platform
is missing a proper hash_map/hash_set implementation and use
the std::map/std::set emulation.

Fixes regression introduced by #1913


bazel-io:

xfxyjwf:
Closing as this seems a duplicate of https://github.com/google/protobuf/pull/2906

matthauck:
Yes, this was an intentional duplicate (see comment on #1913). I wasn't sure which implementation would be preferred.

xfxyjwf:
I see. I actually prefer the fallback to <map> <set> approach in this PR. The other PR casts int64 to size_t for hash and that is too week a hash function that invites hash flooding for 32bit systems where size_t is only 32bit.

bazel-io:

xfxyjwf:
ok to test

matthauck:
looks like there are conflicts here now. :cry:

bazel-io:

matthauck:
ok. updated with master. Is there anything left pending on this PR to pull this in?

matthauck:
_ping_.

PR #2933, opened 2017-04-03
Adding default shell env to allow non-default compilers

mharrend:
I am adding default shell environment, so that protobuf compilation makes use of set
LD_LIBRARY_PATH and so on.
In this way, also non-default gcc installations (e.g. not in /usr/lib) can be used to compile protobuf.
This would fix the following issue:
https://github.com/bazelbuild/bazel/issues/2515


bazel-io:

mharrend:
I signed it!

acozzette:
ok to test (this phrase will trigger the other CI runs)

PR #2937, opened 2017-04-04
Fix duplicate fields test and Any test. The repeated version is passing because nu…

anuraaga:
…ll values in a repeated field are rejected and not testing what it wanted to. Also adds a oneof version that verifies the case of oneof fields of different names (currently only same name check seems to be tested).

Also fix test of Any parsing with unexpected type url. Currently, it is failing because of an unkown field on TestAllTypes, not an unknown type when handling Any,

Also fix spelling of a test.

Note, I tried running Google Java Format on this file and many other places were changed, so I guess this file isn't formatted with that tool, so I've hand-inserted newlines in certain areas.


bazel-io:

acozzette:
ok to test (this phrase will trigger the other CI runs)

acozzette:
@anuraaga Thanks for the pull request!

PR #2950, opened 2017-04-07
Allow unknown values for Map put*Value methods just like every other …

### LINKED ISSUE #2936: Java: Maps with enum value type should allow putting unknown values
Currently, the singular ```put*Value``` method of a map with enum value type does not accept values that are not valid enums. This seems to literally be the only place in generated code that checks the validity of an enum value when using the ```*Value variants``` of the methods (e.g. singular, repeated fields). Even the ```putAll*Value``` variant that takes a ```Map<Key, Integer>``` parameter allows unknown values. I think the ```*Value``` variants were added to allow working with unknown values, for backwards compatibility situations, and in particular, just having the check in one location seems highly inconsistent.

anuraaga:
…enum mutation method.

Fixes #2936


bazel-io:

xfxyjwf:
ok to test

Issue #2965, opened 2017-04-11
Inconsistent dll linkage when compiling with PROTOBUF_USE_DLLS on Windows


ngg:
Since https://github.com/google/protobuf/commit/b1295eeffbf2ae7da8a5c7d8601c3ace0663fa09 the [compiler/plugin.pb.h](https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/plugin.pb.h) file contains lines like this:
`LIBPROTOC_EXPORT extern DescriptorProtoDefaultTypeInternal _DescriptorProto_default_instance_;`
That file also includes [descriptor.pb.h](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.pb.h) which uses `LIBPROTOBUF_EXPORT` for the same instances.
(This latter seems to be the correct one as they reside in [descriptor.pb.cc](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.pb.cc))
This emits [inconsistent dll linkage warnings](https://msdn.microsoft.com/en-us/library/35bhkfb6.aspx).

I think the proper fix would be not to write these lines into plugin.pb.h at all as the included descriptor.pb.h contains them already correctly.

xfxyjwf:
Thanks for reporting the issue. The fix you proposed seems the right solution to me. Could you help send us a pull request?

ngg:
@xfxyjwf can you please take a look at the comments in the PR?

ngg:
bump

xfxyjwf:
Thanks for the fix!

PR #2968, opened 2017-04-12
C++: Do not forward-declare dependencies in generated .h files

ngg:
Those classes and default instances are already forward-declared in included header files.
It solves #2965 where multiple forward-declarations had different dll linkage inside the protobuf compiler.


bazel-io:

src/google/protobuf/compiler/cpp/cpp_file.cc
```diff
@@ -824,19 +824,11 @@ void FileGenerator::GenerateNamespaceClosers(io::Printer* printer) {
 
 void FileGenerator::GenerateForwardDeclarations(io::Printer* printer) {
```
xfxyjwf: Where is this GenerateForwardDeclarations() method called? Can you remove that callsite instead?

src/google/protobuf/any.pb.h
```diff
@@ -167,7 +167,7 @@ class LIBPROTOBUF_EXPORT Any : public ::google::protobuf::Message /* @@protoc_in
   ::google::protobuf::internal::ArenaStringPtr value_;
   mutable int _cached_size_;
   ::google::protobuf::internal::AnyMetadata _any_metadata_;
-  friend struct LIBPROTOBUF_EXPORT protobuf_google_2fprotobuf_2fany_2eproto::TableStruct;
+  friend struct protobuf_google_2fprotobuf_2fany_2eproto::TableStruct;
```
xfxyjwf: This change seems suspiciously. I don't see any related protoc change.

xfxyjwf:
ok to test

src/google/protobuf/any.pb.h
ngg: the suspicuous change is due to https://github.com/google/protobuf/commit/b4b855cde4197de605d463630af9f0ea89583bca
generate_descriptor_proto.sh was not run after this change

src/google/protobuf/compiler/cpp/cpp_file.cc
ngg: It is called from FileGenerator::GenerateHeader().
That call is still neccesary for the types that are declared in the main protobuf file.
Only the dependencies are unneccesary. This is consistent with the generated .pb.cc file where the *\_default\_instance\_ variables are not written for dependencies either.

amerry:
Is this something you would consider merging into the 3.3 branch? This is the sort of warning that it seems dangerous to just ignore.

bazel-io:

ngg:
rebased to current master

PR #2969, opened 2017-04-12
Windows: support long paths in open/mkdir/access

laszlocsomor:
This PR adds support for long paths on Windows in
open(2), mkdir(2), access(2).

Long paths are the ones longer than `MAX_PATH` (260
characters).

~~By including `io_win32.h` we redefine
{open,mkdir,access} to win32_{open,mkdir,access}.~~

> **Update 2017-05-04**: instead of redefining open/close/etc.
> I removed the `win32_` prefix from the new function names
> and simply import them with a `using` statement (if `_WIN32` 
> is defined). At the same I removed all inclusions of `<io.h>`
> so the code can only use the functions declared in `io_win32.h`
> and imported with `using`. This is a much safer approach
> than the undef-define one, because this allows other
> namespaces to declare functions with this name and the
> preprocessor won't change those names to win32_open/etc.

These custom functions transform the path argument
to a format expected by the Windows API, then call
_w{open,mkdir,access} which understand these paths
(apparently they wrap the native widechar Windows
API functions).

This PR upstreams the fix for
https://github.com/bazelbuild/bazel/issues/2634

Fixes https://github.com/google/protobuf/issues/2891

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -45,10 +45,7 @@
 #endif
 #include <sys/stat.h>
 #include <fcntl.h>
-#ifdef _MSC_VER
-#include <io.h>
-#include <direct.h>
-#else
+#ifndef _MSC_VER
 #include <unistd.h>
 #endif
 #include <errno.h>
@@ -80,6 +77,7 @@
 #include <google/protobuf/io/coded_stream.h>
 #include <google/protobuf/io/zero_copy_stream_impl.h>
 #include <google/protobuf/io/printer.h>
+#include <google/protobuf/stubs/io_win32.h>
 #include <google/protobuf/stubs/logging.h>
 #include <google/protobuf/stubs/strutil.h>
 #include <google/protobuf/stubs/substitute.h>
@@ -90,22 +88,6 @@
 namespace google {
 namespace protobuf {
 namespace compiler {
-
-#if defined(_WIN32)
-#define mkdir(name, mode) mkdir(name)
-#ifndef W_OK
-#define W_OK 02  // not defined by MSVC for whatever reason
-#endif
-#ifndef F_OK
-#define F_OK 00  // not defined by MSVC for whatever reason
-#endif
-#ifndef STDIN_FILENO
-#define STDIN_FILENO 0
-#endif
-#ifndef STDOUT_FILENO
-#define STDOUT_FILENO 1
-#endif
-#endif
 
 #ifndef O_BINARY
 #ifdef _O_BINARY
@@ -118,6 +100,14 @@ namespace compiler {
 namespace {
 #if defined(_WIN32) && !defined(__CYGWIN__)
 static const char* kPathSeparator = ";";
+// DO NOT include <io.h>, instead create functions in io_win32.{h,cc} and import
+// them like we do below.
+using google::protobuf::internal::win32::access;
+using google::protobuf::internal::win32::close;
+using google::protobuf::internal::win32::mkdir;
+using google::protobuf::internal::win32::open;
+using google::protobuf::internal::win32::setmode;
+using google::protobuf::internal::win32::write;
 #else
 static const char* kPathSeparator = ":";
 #endif
@@ -141,9 +131,9 @@ static bool IsWindowsAbsolutePath(const string& text) {
 
 void SetFdToTextMode(int fd) {
 #ifdef _WIN32
-  if (_setmode(fd, _O_TEXT) == -1) {
+  if (setmode(fd, _O_TEXT) == -1) {
     // This should never happen, I think.
-    GOOGLE_LOG(WARNING) << "_setmode(" << fd << ", _O_TEXT): " << strerror(errno);
+    GOOGLE_LOG(WARNING) << "setmode(" << fd << ", _O_TEXT): " << strerror(errno);
   }
 #endif
   // (Text and binary are the same on non-Windows platforms.)
@@ -151,9 +141,9 @@ void SetFdToTextMode(int fd) {
 
 void SetFdToBinaryMode(int fd) {
 #ifdef _WIN32
-  if (_setmode(fd, _O_BINARY) == -1) {
+  if (setmode(fd, _O_BINARY) == -1) {
     // This should never happen, I think.
-    GOOGLE_LOG(WARNING) << "_setmode(" << fd << ", _O_BINARY): " << strerror(errno);
+    GOOGLE_LOG(WARNING) << "setmode(" << fd << ", _O_BINARY): " << strerror(errno);
   }
 #endif
   // (Text and binary are the same on non-Windows platforms.)
--- a/src/google/protobuf/stubs/io_win32.cc
+++ b/src/google/protobuf/stubs/io_win32.cc
@@ -1,0 +1,362 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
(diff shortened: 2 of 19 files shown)


laszlocsomor:
@buchgr : FYI

src/google/protobuf/io/io_win32.cc
```diff
@@ -0,0 +1,281 @@
+// Protocol Buffers - Google's data interchange format
```
xfxyjwf: Please move this to src/google/protobuf/stubs/.... That's where we put all base (and platform related) dependencies.

laszlocsomor: Done.

laszlocsomor:
https://github.com/google/protobuf/pull/2969/commits/292fef71f9775c4cb55983a4d35faec771aac976 was also my commit, just using git-bash instead of my usual git-on-msys, and the author wasn't set up properly.

laszlocsomor:
Gentle ping.l

xfxyjwf:
The travis test and appveyor test are failing (seems related to this PR). Could you fix them?

laszlocsomor:
PTAL -- all checks pass and I'm confident there won't be any more changes necessary (unless reviewers request so).

src/google/protobuf/stubs/io_win32.cc
```diff
@@ -0,0 +1,352 @@
...
+  static bool is_alpha(char ch) { return isalpha(ch); }
+};
+
+template <>
+struct CharTraits<wchar_t> {
+  static bool is_alpha(wchar_t ch) { return iswalpha(ch); }
+};
+
+template <typename char_type>
+bool has_drive_letter(const char_type* ch) {
```
damienmg: This only work if checked after the long path prefix. I suggest documenting so if intentional.

src/google/protobuf/stubs/io_win32.cc
```diff
@@ -0,0 +1,352 @@
...
+
+template <typename char_type>
+bool has_longpath_prefix(const std::basic_string<char_type>& path) {
+  return path.size() > 4 && path[0] == '\\' && path[1] == '\\' &&
+         path[2] == '?' && path[3] == '\\';
+}
+
+template <typename char_type>
+bool is_path_absolute(const std::basic_string<char_type>& path) {
+  return (path.size() > 2 && path[1] == ':') || has_longpath_prefix(path);
```
damienmg: replace with has_drive_letter?

src/google/protobuf/stubs/io_win32.cc
```diff
@@ -0,0 +1,352 @@
...
+struct CharTraits<char> {
+  static bool is_alpha(char ch) { return isalpha(ch); }
+};
+
+template <>
+struct CharTraits<wchar_t> {
+  static bool is_alpha(wchar_t ch) { return iswalpha(ch); }
+};
+
+template <typename char_type>
```
damienmg: All those onliner should probably be declared inline (also my C++ is a bit rusty, this might already be the case for some magic).?

src/google/protobuf/stubs/io_win32.cc
laszlocsomor: Yes, intentional. Done.

src/google/protobuf/stubs/io_win32.cc
laszlocsomor: Done.

src/google/protobuf/stubs/io_win32.cc
laszlocsomor: I think the compiler would auto-inline them where it can. I added the inline declarations anyway but ran into linker errors that I couldn't resolve and didn't want to spend too much time trying.

laszlocsomor:
gentle ping

src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -118,6 +100,14 @@ namespace compiler {
 namespace {
 #if defined(_WIN32) && !defined(__CYGWIN__)
 static const char* kPathSeparator = ";";
+// DO NOT include <io.h>, instead create functions in io_win32.{h,cc} and import
+// them like we do below.
+using google::protobuf::stubs::access;
```
xfxyjwf: Can you use the google::protobuf::internal namespace? That's what we currently use for most other stubs headers.

laszlocsomor: Sorry for the long silence. Done, though I moved them to `google::protobuf::internal::win32`, because the method names are generic enough that they might collide with similarly named functions.

laszlocsomor: ...if they were in `google::protobuf::internal`, that is.

laszlocsomor:
Re: https://github.com/google/protobuf/pull/2969#pullrequestreview-42691215 : Done.

laszlocsomor:
And rebased on top of upstream/master.

laszlocsomor:
Unfortunately ci.bazel.io no longer has the logs from that run (@damienmg : FYI) so I'd just go ahead and merge this PR and fix Bazel breakages if there are any.

buchgr:
Congrats @laszlocsomor :+1: :fireworks: :champagne:

tri-bao:
Hello,

Does Anybody know in which release this fix was included? I try version 3.4.0 (latest version 3.4.1 doesn't have windows compiler binary) on Windows 10 64bit and still get this error. The error like:

D:\...\SomeFile.java: No such file or directory

The path is 261-characters long. Shortening the path makes the compiler work.

I checked the source code zip of version 3.4.0, it seems that it includes this patch. 

Anybody facing the same problem? What are your solutions apart from shortening the path :)

laszlocsomor:
I'm not on the Protobuf team, but I assume if the files are in the 3.4.0 zip then they should have been compiled into the release binary.

The files' contents and the import statements are surrounded by `#ifdef _MSV_VER` ... `#endif` statements, meaning they'd only be compiled in if built with Visual C++.

@tri-bao Have you tried building your own protoc binary?

A workaround for path lengths is to create a junction that points to the long path, but the junction itself has a short path. Since junctions are just directories with some extra metadata in the filesystem, and the resolution happens in the kernel (I believe), this workaround works both with Command Prompt and Windows Explorer, as well as with all tools that are not longpath-aware.

tri-bao:
thanks @laszlocsomor , compiling with Visual Studio works! I've created a ticket #3856  for asking why the public precompiled binary was not done with Visual Studio, which makes sense for Windows

pherl:
The precompiled protoc was built with mingw32 cross compiled under linux environment for various reasons: mostly for automated building process and eliminate msvc dll dependencies.

 Is there anyway to support the long path also under this non MSVC? The code used to be guarded by #ifdef _WIN32, and I changed it to _MSV_VER as it create redefinition on those POSIX system APIs -  linux-like OS for windows also implements those syscalls. I believe it can be fixed by re-organizing the code. i.e. instead of "using io_win32::mkdir" define our own mkdir wrapper and delegate it to the one in io_win32 when _WIN32 is defined.

laszlocsomor:
Hey folks, sorry for dropping the ball on this issue.

I opted for defining new implementations in a C++ namespace but using the POSIX function names for the following benefits:
- `using io_win32::mkdir` drop-in replaces `mkdir` in the whole compilation unit by just adding that line
- no need to replace `mkdir` with `mkdir_wrapper`
- no need to remember calling `mkdir_wrapper` in new code

@pherl : Does it solve the compilation problem with MinGW if you only `#include` linux system headers (such as `<sys/stat.h>` and `<sys/types.h>`) if `_WIN32` is not defined?

pherl:
But the downside would be having to #ifdef _WIN32 and add using directives for each new files requires those posix functions. Having the wrapper made it more explicit and avoid the breakage by including posix headers transitively that we cannot control.

Anyway this should have been fixed in 3.5.0, with changes like: https://github.com/google/protobuf/blob/3.5.x/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc#L64

PR #2976, opened 2017-04-16
Memory leak fix since argv array could leak in this block.

devnexen:


bazel-io:

src/google/protobuf/compiler/subprocess.cc
```diff
@@ -303,6 +303,7 @@ void Subprocess::Start(const string& program, SearchMode search_mode) {
 
   child_pid_ = fork();
   if (child_pid_ == -1) {
+    free(argv[0]);
     GOOGLE_LOG(FATAL) << "fork: " << strerror(errno);
```
xfxyjwf: GOOGLE_LOG(FATAL) will terminate the process immediately. I don't think we care about memory deallocation at this point.

PR #2981, opened 2017-04-17
Fix PHP generator backward incompatibility

kpayson64:
GBP::CheckRepeatedField does not return a value for older runtimes.


bazel-io:

TeBoring:
Talked offline. This is not needed. We don't guarantee new generated code will work with old runtime. Plus, we haven't officially released PHP yet, so no compatibility guarantee.

Issue #2446, opened 2016-12-06
PHP generated code: syntax error


0xFEEDC0DE64:
I dont have any PHP-package manager like composer.
I've had many problems using protobuf in php because of missing includes.

I tried generating a demo proto for php. Please see attached files.

`Parse error: syntax error, unexpected '\' (T_NS_SEPARATOR), expecting identifier (T_STRING) in /www/protobuf/test.pb.php on line 57`

```
<?php

spl_autoload_register(function ($class_name) {
    require 'protobuf/' . str_replace('\\', '/', $class_name) . '.php';
});
require 'protobuf/Google/Protobuf/descriptor.php';
require 'protobuf/Google/Protobuf/Internal/Type.php';
require 'protobuf/Google/Protobuf/descriptor_internal.pb.php';

require 'protobuf/test.pb.php';

$test = new Person;
echo $test->encode();
```

The proto file:
```
syntax = "proto3";

message Person {
  string name = 1;
  int32 id = 2;
  string email = 3;

  enum PhoneType {
    MOBILE = 0;
    HOME = 1;
    WORK = 2;
  }

  message PhoneNumber {
    string number = 1;
    PhoneType type = 2;
  }

  repeated PhoneNumber phone = 4;
}

message AddressBook {
  repeated Person person = 1;
}
```

Generated .pb.php file:
```
<?php
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: test.proto

use Google\Protobuf\Internal\DescriptorPool;
use Google\Protobuf\Internal\GPBType;
use Google\Protobuf\Internal\RepeatedField;
use Google\Protobuf\Internal\GPBUtil;

class Person extends \Google\Protobuf\Internal\Message
{
    private $name = '';
    private $id = 0;
    private $email = '';
    private $phone;

    public function getName()
    {
        return $this->name;
    }

    public function setName($var)
    {
        GPBUtil::checkString($var, True);
        $this->name = $var;
    }

    public function getId()
    {
        return $this->id;
    }

    public function setId($var)
    {
        GPBUtil::checkInt32($var);
        $this->id = $var;
    }

    public function getEmail()
    {
        return $this->email;
    }

    public function setEmail($var)
    {
        GPBUtil::checkString($var, True);
        $this->email = $var;
    }

    public function getPhone()
    {
        return $this->phone;
    }

    public function setPhone(&$var)
    {
[...]

PR #2988, opened 2017-04-20
Enum defined without package have incorrect class name.

TeBoring:
Fix the bug by sharing the code for generating class name for both
message and enum.


bazel-io:

TeBoring:
Fix #2446

Issue #2743, opened 2017-02-21
Undefined Index Notice - PHP


tbillington:
I'm getting PHP notices for a lot of my types, not sure why.

```
PHP Notice:  Undefined index: .foo.bar.baz in /Users/trent/src/github.com/sajari/phptest/vendor/google/protobuf/php/src/Google/Protobuf/Internal/DescriptorPool.php on line 125
```

TeBoring:
which version are u using?

tbillington:
php grpc library: `"grpc/grpc": "v1.1.2"`
protoc: `libprotoc 3.2.0`

Just got the grpc_php_plugin from master a few hours ago.

TeBoring:
Have you used protoc to generate code?

tbillington:
Yup

```
protoc -I$PROTO_SRC --php_out=$dest --grpc_out=$dest --plugin=protoc-gen-grpc=/Users/trent/grpc/bins/opt/grpc_php_plugin $dir/*.proto
```

TeBoring:
What's the protoc version?
And did you autoload your generated proto files?

tbillington:
Yes, I autoloaded. I posted just before your previous comment the protoc version (maybe harder to see in email): `libprotoc 3.2.0`

TeBoring:
Can I see your proto file and test code?

tbillington:
Seems this is only happening on enums.

TeBoring:
Maybe it can be fixed in you have a non-empty namespace. Let me know if it works.

tbillington:
Still had the issue in a non-empty namespace.

tbillington:
Here is some test code that demonstrates the issue. I removed the namespace because it had no impact.

Proto File:
```proto
syntax = "proto3";

package simpletest;

service SimpleTest {
  rpc Test (TestRequest) returns (TestResponse);
}

message TestRequest {
  enum TestEnum {
    ZERO = 0;
    ONE = 1;
  }

  string name = 1;

  TestEnum test_enum = 2;
}

message TestResponse {
  string reply = 1;
}
```

Test Code:
```php
<?php
use Grpc\ChannelCredentials;
require  __DIR__ . "/vendor/autoload.php";
include_once  __DIR__ . "/proto/SimpleTest/SimpleTestClient.php";
include_once  __DIR__ . "/proto/SimpleTest/TestRequest.php";
[...]

PR #2989, opened 2017-04-20
Add test for nested enum for php

TeBoring:
#2743


bazel-io:

Issue #2740, opened 2017-02-21
PHP Empty


tbillington:
I don't know whether this is a grpc error or proto, I posted it initially in grpc but I'll cross post it here too.

Basically, `Empty` is generated to `GPBEmpty` properly in the metadata and for the service clients, but not for the actual class definition.

https://github.com/grpc/grpc/issues/9796

This is probably related to https://github.com/google/protobuf/commit/99564c3347223b92e49bd40f37b8c0d6fc6a3971 ?

TeBoring:
Is it possible for you to rename Empty?

tbillington:
Just seems weird to go through the effort of rewriting to `GPBEmpty` for the metadata but not doing it for the actual definition. It's like the use case has been thought of but the implementation only half works?

TeBoring:
I want to add PB by default as prefix for reserved words. But I don't have time to do that. Could you help send a fix? Thanks!

PR #2990, opened 2017-04-20
Prepend "PB" to generated classes whose name are reserved words.

TeBoring:


bazel-io:

TeBoring:
Fixes #2740

PR #2995, opened 2017-04-20
Allow proto files to import descriptor.proto

### LINKED ISSUE #2901: php:  Class 'GPBMetadata\Google\Protobuf\Descriptor' not found
Please refer to grpc/grpc#10217.

Ｗhen generated `google/api/annotations.proto`, it calls `\GPBMetadata\Google\Protobuf\Descriptor::initOnce();`.
However, there is only class `GPBMetadata\Google\Protobuf\Internal\Descriptor`.

cc @stanley-cheung

TeBoring:
descriptor.proto uses proto2 syntax, which is not ready for external
usage. However, some proto3 files import descriptor.proto and cannot be
used. In this PR, all references (We cheated by only removing
extensions, which is enough for now. User should avoid using messages
defined in descriptor.proto as field type.) to content in
descriptor.proto are removed from generated files. Those that import
descriptor.proto can be used like other proto files.
Fix #2901.

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -674,6 +674,12 @@ void GenerateAddFileToPool(const FileDescriptor* file, bool is_descriptor,
   } else {
     for (int i = 0; i < file->dependency_count(); i++) {
       const std::string& name = file->dependency(i)->name();
+      // Currently, descriptor.proto is not ready for external usage. Skip to
+      // import it for now, so that its dependencies can still work as long as
+      // they don't use protos defined in descriptor.proto.
+      if (name == kDescriptorFile) {
+        continue;
+      }
       std::string dependency_filename =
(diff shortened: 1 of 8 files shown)


bazel-io:

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -673,7 +673,13 @@ void GenerateAddFileToPool(const FileDescriptor* file, bool is_descriptor,
         "$pool->finish();\n");
   } else {
     for (int i = 0; i < file->dependency_count(); i++) {
-      const std::string& name = file->dependency(i)->name();
+      std::string name = file->dependency(i)->name();
```
michaelbausor: Can `name` stay as `const`?

TeBoring: Done

michaelbausor:
LGTM once tests have run

PR #2999, opened 2017-04-21
Bugfix: Enum classes not escaped properly

mcos:
The php generator isn't escaping classes properly for calls to `GPBUtil::checkRepeatedField` and `GPBUtil::checkMapField`. The current generator generates the following code:

```
GPBUtil::checkRepeatedField($var, \Google\Protobuf\Internal\GPBType::ENUM, My\Namespace\Foo::class);
```

Which results in an error when trying to use the class outside of its package. This fix properly escapes the class import to generate the following.  

```
GPBUtil::checkRepeatedField($var, \Google\Protobuf\Internal\GPBType::ENUM, \My\Namespace\Foo::class);
```


bazel-io:

TeBoring:
Duplicate of #2988.
@mcos, Sorry that I forgot to notified you that I have fixed it.

mcos:
okay, thanks @TeBoring. will a release be cut with this fix soon?

TeBoring:
Yes, we will include the fix in 3.3.0 next week.

Issue #2572, opened 2017-01-10
PHP  Class Message decode() function missing 'return'.


zwniux:
/**
     * Parses a protocol buffer contained in a string.
     *
     * This function takes a string in the (non-human-readable) binary wire
     * format, matching the encoding output by encode().
     *
     * @param string $data Binary protobuf data.
     * **@return bool Return true on success.**
     */
    public function decode($data)
    {
        $input = new InputStream($data);
        **$this->parseFromStream($input);**
    }

Issue #2376, opened 2016-11-16
PHP Fatal error when passing invalid value into decode


kgosschalk:
If I pass an invalid protobuf string (for example the string, "erwgrewg") into a decode method, I get a fatal error in the parseFieldFromStream method within Message.php.

The error happens if $field is null (due to failed decryption) when it goes to do $field_type = $field->getType();

It would be much better to throw an exception or do checking to ensure it doesn't cause a fatal error, which cannot be caught.

PR #3000, opened 2017-04-21
Throw exception when parsing invalid data.

TeBoring:
#2376, #2572


bazel-io:

TeBoring:
@stanley-cheung to review

php/src/Google/Protobuf/Internal/GPBWire.php
```diff
@@ -172,7 +172,7 @@ public static function zigZagDecode64($uint64)
 
     public static function readInt32(&$input, &$value)
     {
-        return $input->readVarint32($value);
```
stanley-cheung: This seems unintentional?

PR #3004, opened 2017-04-22
updated mox.py

vschs007:
> removed unnecessary spaces
> more space efficient


bazel-io:

vschs007:
any issues with this update ?

anandolee:
I don't think this change is needed. 

And we also have python style guide for google projects like two blank lines between classes which against your change.

PR #3010, opened 2017-04-24
Add a test case for nested enum, which was missed previously.

TeBoring:


bazel-io:

TeBoring:
ok to test

PR #3018, opened 2017-04-26
Remove "using namespace std" from stubs/common.h

acozzette:
This prevents the contents of the std namespace from being effectively
pulled into the top-level namespace in all translation units that
include common.h. I left in individual using statements for a few common
things like std::set and std::map, because it did not seem worth going
through the churn of updating the whole codebase to fix those right now.

### DIFF
--- a/src/google/protobuf/stubs/common.h
+++ b/src/google/protobuf/stubs/common.h
@@ -35,7 +35,12 @@
 #ifndef GOOGLE_PROTOBUF_COMMON_H__
 #define GOOGLE_PROTOBUF_COMMON_H__
 
+#include <algorithm>
+#include <iostream>
+#include <map>
+#include <set>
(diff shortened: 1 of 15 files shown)


src/google/protobuf/stubs/common.h
```diff
@@ -220,7 +225,13 @@ class FatalException : public std::exception {
 
 // This is at the end of the file instead of the beginning to work around a bug
 // in some versions of MSVC.
-using namespace std;  // Don't do this at home, kids.
+using std::istream;
```
xfxyjwf: Add a TODO here? I think we want to get rid of these using statements in public header files as well.

acozzette: Done, I've added a TODO there.

Issue #1814, opened 2016-07-20
Issues with the objc `named_framework_to_proto_path_mappings_path` parameter


justinseanmartin:
FYI - @thomasvl (looks like you implemented it?)

We have a problem attempting to use the framework support in the ObjectiveC compiler. The issue is two-fold:
1. Local proto dependencies within the same framework use "path/to/The.pbobjc.h" dependencies, but the framework imports expect them to be flattened as <framework/The.pbobjc.h> (or paths to be searched recursively).
2. We may have multiple protos with the same name in the same module under different folder paths, which will conflict on being flattened. e.g. squareup/group1/Options.pbobjc.h & squareup/group2/Options.pbobjc.h (no naming conflict by having a unique prefix)

For this reason, I believe we should be maintaining the folder structure within the frameworks and the imports that are added should continue to use the explicit full path to the dependency. I believe the fix would look like:

```
diff --git a/src/google/protobuf/compiler/objectivec/objectivec_file.cc b/src/google/protobuf/compiler/objectivec/objectivec_file.cc
index cf8c34e..421d4e1 100644
--- a/src/google/protobuf/compiler/objectivec/objectivec_file.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_file.cc
@@ -110,14 +110,14 @@ void ImportWriter::AddFile(const FileGenerator* file) {
   if (proto_lookup != proto_file_to_framework_name_.end()) {
     other_framework_imports_.push_back(
         proto_lookup->second + "/" +
-        FilePathBasename(file_descriptor) + extension);
+        file->Path() + extension);
     return;
   }

   if (!options_.generate_for_named_framework.empty()) {
     other_framework_imports_.push_back(
         options_.generate_for_named_framework + "/" +
-        FilePathBasename(file_descriptor) + extension);
+        file->Path() + extension);
     return;
   }

```

This leads to imports looking like:

```
#import <Framework/path/to/The.pbobjc.h>
```

I'm not sure if there is something I'm missing though. I'd be happy to open a PR with this change if this is an agreeable solution. Thanks for your feedback.

CC: @dnkoutso

justinseanmartin:
There appears to be another problem with framework support. In the imports that are generated for source files, the headers from within the same module need to be imported using quoted strings and relative paths rather than as framework imports. I'm working on a fix for this now.

sergiocampama:
One thing you can do is add the path to the mappings file, as so:

```
[...]

Issue #864, opened 2015-10-08
ObjC: Don't mangle filenames


kientzle:
The Obj-C backend currently mangles the input filenames, converting e.g., foo_bar.proto to FooBar.pb.m.  This causes problems for build systems (including Xcode) that could otherwise automatically track dependencies and recompile proto files only as needed.

The offending code is in src/google/protobuf/compiler/objectivec/objectivec_helpers.cc.  The call to UnderscoresToCamelCase() here should be removed:

string FilePath(const FileDescriptor\* file) {
  string output;
  string basename;
  string directory;
  PathSplit(file->name(), &directory, &basename);
  if (directory.length() > 0) {
    output = directory + "/";
  }
  basename = StripProto(basename);

  // CamelCase to be more ObjC friendly.
  basename = UnderscoresToCamelCase(basename, true);

  output += basename;
  return output;
}

thomasvl:
I believe a few other languages do this also (java?  C#?).  Since this would be a break change for folks, I don't think we're likely to change the behavior at this point.

kientzle:
How do you compile your Obj-C programs?  If you're using Xcode, I'd love to see the Xcode build rules you're using to recompile your protos.  Personally, I would like to use something similar to this with make:

%.objc.h %.objc.m: %.proto
   protoc ...

But that requires that the base filenames match.

thomasvl:
Since a proto can reference another proto, you end up needing to also have the dependency change exposes to trigger rebuilds (just like source files).  So Xcode's build rule model already can't capture that.  I've seen script do it to deal with the dependency chain problem, but none of them are pretty.  Google uses http://bazel.io which knows about protos, so we don't directly have this issue.  We've talked about doing an Xcode plugin to make the build/rebuild process better, but haven't really put time into it.

thomasvl:
fyi - You might also want to look at the --dependency_out option.  You might be able to use that to help pick off the name transform and dependency info.

PR #3025, opened 2017-04-28
ObjectiveC: make file name same as class name when objc_class_prefix is set

axxapy:
There is no technical limitation but in general classes names and file names supposed to be the same in ObjC.

There is at least one case when it does not work without this fix:
if there are messages with same names but in different packages, you can not put generated .h files into framework since it has plain file structure (no directories) for headers. So each file must have unique name.


bazel-io:

axxapy:
I signed it!

thomasvl:
There are plenty of cases where .proto files have multiple classes in them and the file name is about the collection, not any specific message within it.  This would also be a breaking change as developers would have to do a lot of source updates to \#import the new headers, not to mention all the Xcode project files updates it would trigger.

If we did anything to change the names of the generated files, I'd likely suggest we revisit #864.  The mangling we do is a historical thing from what was done internally at Google, but as that issue points out, it makes integration into other build systems much more difficult.

thomasvl:
Also, the current transform does make automated builds hard, but atleast it can be done since the output name is only dependent on the input name.  Using the prefix would mean an automated build system has to be able to parse/extract from the file the prefix to figure out what the output name is.

axxapy:
Well, collisions are possible anyway
For instance, when there are two messages one with name "Boo" and objc prefix "Foo" and another with name "FooBoo" - in this case we will have two different files with same classes names.

According concern that it will force a lot of developers to redo their imports I totally agree.

What would you say if I add special option for enabling this behavior? By default it will be switched off.

thomasvl:
> Well, collisions are possible anyway
> For instance, when there are two messages one with name "Boo" and objc prefix "Foo" and another with name "FooBoo" - in this case we will have two different files with same classes names.

Sorry, not sure I'm following since I didn't bring up anything about collisions.  Yes, there could be class name collisions and/or file name collisions.  I'm not sure that what is being suggested here really changes that fact, it just introduces _another way_ file collisions could happen.  If anything, that could be an argument for not adding this as it is another way for filename collisions to happen.

> What would you say if I add special option for enabling this behavior? By default it will be switched off.

The catch with something like this is making it work as things chain.  For example, the `#import` directives to include the Well Know Types wouldn't use the prefix since that's how the files are distributed with this library.  This then ripples, if a developer makes a framework that exposes some protos, they will include the headers in their framework.  If a second developer then uses that framework and protos in their own, they need the imports to match the naming style used in the framework.  Trying to let each developer pick something can quickly ballon into something like [generator option `named_framework_to_proto_path_mappings_path`](https://github.com/google/protobuf/tree/master/objectivec#objective-c-generator-protoc-options), which I believe still isn't complete because of the issues raised in #1814.

axxapy:
Fair enough.

Still problem with framework imports exists:
lets say we have two proto messages:

    package.request.Signin
    package.response.Signin

They will be compiled to

    ./package/request/Signin.pbobjc.h
    ./package/request/Signin.pbobjc.m
    ./package/response/Signin.pbobjc.h
    ./package/response/Signin.pbobjc.m

which is totally fine until you put them into framework: xcode simply can not import both at the same time because it has plain file structure for frameworks which causes filename collision.

To make it work you have to have unique message names across all protobuf structure which seems very strange requirement.

One way to solve this problem is to add prefix to filename (implemented in this pull request).
But may be there is another already existent solution I don't see?

thomasvl:
While not common, you can actually use sub directories within Frameworks for headers.  Cocoapods exposes this with their podspec setting `header_mappings_dir`.

thomasvl:
Another option (and likely more complete, but not perfect) to deal with collisions would be to support using the path to generate the file names.  So _foo/bar/baz.proto_ would generate _foo\_bar\_baz.pbobjc.h_ and _foo\_bar\_baz.pbobjc.m_.

I'm still a little worried adding any options can cascade into making things harder to integrate.

PR #3080, opened 2017-05-10
Fix upb load descriptor when no messages defined in prorto.

TeBoring:


bazel-io:

php/ext/google/protobuf/upb.c
```diff
@@ -2292,7 +2292,7 @@ bool upb_symtab_addfile(upb_symtab *s, upb_filedef *file, upb_status *status) {
   n = upb_filedef_defcount(file);
   defs = upb_gmalloc(sizeof(*defs) * n);
 
-  if (defs == NULL) {
+  if (defs == NULL && n != 0) {
```
haberman: Seems simpler to just say `if (n == 0) { return true; }` before the `upb_gmalloc()`.

TeBoring: Done

PR #3082, opened 2017-05-11
There is no need to adjust UInt32 values when php supports 8 byte integers

rltvty:
And the conversion actually causes an exception when calling `serializeToString()`

```
Caught an error, errno: 1024, msg: Output stream doesn't have enough buffer., /var/www/vendor/google/protobuf/php/src/Google/Protobuf/Internal/OutputStream.php:92
```

Removing this line fixes the issue.


bazel-io:

TeBoring:
ok to test

bazel-io:

TeBoring:
Could you add a test case?

TeBoring:
I thine the original motivation is to make uint32 the same value on 64-bit arch and 32-bit arch.

PR #3094, opened 2017-05-15
control repeated_field memory increase on mobile platform

afpro:


bazel-io:

afpro:
I signed it!

xfxyjwf:
@afpro Can you explain what this PR is trying to improve? My understanding is that this only helps when the repeated size is between 1M and 2M. For smaller sizes it doesn't help, and for larger sizes it makes the performance worse.

afpro:
i'm using tensorflow with DT_HALF dtype, this dtype use int32 for 16 bit data.
while parsing proto file, repeated field size is indeterminate.
if it's size is 256+1, it will keep 512 elements memory.
my model is about 80mb, after loaded in memory, use 137mb memory, that's not right.
(RepeatedField don't have a strip method)

afpro:
CI already fucked me up, oops.
i'll close this pr, try it on my own project first.

PR #3100, opened 2017-05-16
control RepeatedField memory growth on mobile platform

afpro:
RepeatedField double memory usage while Reserve
but mobile platform is memory sensitive, and usually won't load
very large protobuf model
so limit RepeatedField memory growth up to 1mb is acceptable

### DIFF
--- a/src/google/protobuf/repeated_field.h
+++ b/src/google/protobuf/repeated_field.h
@@ -106,6 +106,32 @@ inline int CalculateReserve(Iter begin, Iter end) {
 }
 }  // namespace internal
 
+namespace RepeatedFieldInternal {
+  template <typename T, typename U>
+  struct SameType {
+    static bool value() {
+      return false;
+    }
+  };
+
+  template <typename T>
+  struct SameType<T, T> {
+    static bool value() {
+      return true;
+    }
+  };
+
+  template <typename T>
+  bool isPrimitive() {
+    return SameType<T, bool>::value()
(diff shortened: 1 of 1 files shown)


bazel-io:

xfxyjwf:
@afpro I don't think this is the right change to make. It will make the performance much worse for large payloads, and special-casing for a particular platform with a magic number like 1M is usually a bad solution longer term. The improvement this change can achieve is also very limited. If you can parse a proto of 32MB without this change, with it you can at best get to 64MB. Not anything fundamental. It's way better you can just avoid having such large protos on mobile platform all together.

afpro:
@xfxyjwf 32MB is not a big improvement for a proto, but 32MB memory is very import for a mobile platform app. with this, i can use a 32MB proto, and leave about (128-32)MB for me. without this, i have only (128-64)MB, this is a big difference. but you're right, magic number is not a good solution, maybe with a switch and configurable growth size?

xfxyjwf:
@afpro Can you avoid creating large protos? It's generally recommended to keep your protos small and it's especially true on mobile.

afpro:
@xfxyjwf i'm using tensorflow, export graph as a proto, can't be very small. (especially with DT_HALF, without DT_HALF everything is fine). if you insist this is not necessary for mobile platform, feel free for closing this PR. i have already gave up DT_HALF for model compression, quantization is better.

PR #3103, opened 2017-05-17
Adds serial and parallel parsing tests.

sergiocampama:
Checks if parallel parsing is faster than serial parsing. This should be the case unless locks are leaking into different instances.

### DIFF
--- a/objectivec/Tests/GPBPerfTests.m
+++ b/objectivec/Tests/GPBPerfTests.m
@@ -64,6 +64,112 @@ static const uint32_t kRepeatedCount = 100;
   }];
 }
 
+- (void)testMessageSerialParsingPerformance {
+  // This and the next test are meant to monitor that the parsing functionality of protos does not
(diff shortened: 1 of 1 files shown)


objectivec/Tests/GPBPerfTests.m
```diff
@@ -64,6 +64,57 @@ - (void)testMessagePerformance {
   }];
 }
 
+- (void)testMessageSerialParsingPerformance {
```
thomasvl: Actually, can you tweak the name/add a comment to say what this (and the next one) are for?  Call out that they are specific to *parsing* (the other is building, serializing, and parsing), and that it is for extension fields.

It might not hurt to also make a version that uses TestAllTypes that just covers fields and not extensions.

sergiocampama: Added a TestAllTypes test and more comments

sergiocampama:
Done, PTAL

PR #3109, opened 2017-05-17
changes composer autoload files to classmap

### LINKED ISSUE #3107: Improper implementation of PHP Autoloader
The [PHP Autoloader](https://github.com/google/protobuf/blob/master/composer.json#L20) is including the class files at the start of each request instead of using proper PSR standards. This means if the `protobuf.so` C extension is used in tandem with this library, a fatal error is thrown due to the classes being redeclared.

If the classes are split into files based on their classnames, and `Classmap` or `PSR-0` is used for the autoloading instead of `files`, this will be resolved.

bshaffer:
fixes #3107


bazel-io:

bshaffer:
This won't work, as the classes still throw fatal errors with classmap. They will have to be broken out into their own class files.

Issue #3118, opened 2017-05-22
Documentation of PHP generated classes


michaelbausor:
Getter and setter methods on generated PHP classes should include phpdoc in the method documentation. E.g. `@return` for getters and `@param` for setters.

mlosev:
+1 for phpdoc, it's very important for modern IDE and large codebases

ev4ern:
+1 for phpdoc

TeBoring:
Will add more detail in the phpdoc.

Yakoot:
+1

publiteer:
+1

Grey2k:
+1 for phpdoc

TeBoring:
FYI, currently, the phpdoc getter and setter only contains the field definition:
https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/DescriptorProto.php#L85
Would that be enough?

michaelbausor:
I think that we want the `@return` and `@param` annotations. These are used by IDEs like PHPStorm to provide autocompletion, and by documentation generators

PR #3131, opened 2017-05-24
adds PHPDoc @return and @param for getters and setters respectively

bshaffer:
see #3118


grpc-testing:
Thanks for your pull request. The automated tests will run as soon as one of the admins verifies this change is ok for us to run on our infrastructure.

grpc-testing:
Thanks for your pull request. The automated tests will run as soon as one of the admins verifies this change is ok for us to run on our infrastructure.

bazel-io:

php/tests/generated_class_test.php
```diff
@@ -877,4 +877,279 @@ public function testPrefixForReservedWords()
...
+
+    #########################################################
+    # Test phpdoc for getters and setters.
+    #########################################################
+
+    public function testPhpDocForClass()
+    {
+        $class = new ReflectionClass('Foo\TestMessage');
+        $doc = $class->getDocComment();
+        $this->assertContains('Foo&#92;TestMessage', $doc);
```
bshaffer: @teboring I see these escaped characters all over the phpdocs. I do not think we need them. If this is for HTML escaping, this should be done wherever the HTML is generated. Thoughts on removing it?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -297,6 +302,34 @@ std::string TypeName(const FieldDescriptor* field) {
   }
 }
 
+std::string PhpTypeName(const FieldDescriptor* field, bool is_descriptor) {
+  if (field->is_repeated()) {
+    return "array|\\Google\\Protobuf\\Internal\\RepeatedField";
```
TeBoring: Currently, RepeatedField is for internal use. I'd like to only mention array as parameter for now.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1101,16 +1137,24 @@ void GenerateFieldDocComment(io::Printer* printer,
   printer->Print(
     " * <code>^def^</code>\n",
     "def", EscapePhpdoc(FirstLineOf(field->DebugString())));
+  if (function_type == kFieldSetter) {
+    printer->Print(" * @param ^php_type^ $var\n",
+      "php_type", PhpTypeName(field, is_descriptor));
```
TeBoring: Is @return for fluent setter necessary? I feel it's quite redundent. But I'm open to add it if we need to be strict with the doc format.

php/tests/generated_class_test.php
```diff
@@ -877,4 +877,279 @@ public function testPrefixForReservedWords()
         $m = new \Foo\PBEmpty();
         $m = new \PrefixEmpty();
     }
+
+    #########################################################
```
TeBoring: I prefer to add a new test file for documentation. This is not our API and is open to change.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -297,6 +302,34 @@ std::string TypeName(const FieldDescriptor* field) {
...
+    case FieldDescriptor::TYPE_INT64:
+    case FieldDescriptor::TYPE_UINT32:
+    case FieldDescriptor::TYPE_UINT64:
+    case FieldDescriptor::TYPE_SINT32:
+    case FieldDescriptor::TYPE_SINT64:
+    case FieldDescriptor::TYPE_FIXED32:
+    case FieldDescriptor::TYPE_FIXED64:
+    case FieldDescriptor::TYPE_SFIXED32:
+    case FieldDescriptor::TYPE_SFIXED64:
+    case FieldDescriptor::TYPE_ENUM: return "int";
```
TeBoring: int64, uint64 are represented as numeric string on 32-bit machine. How shall we handle this?

php/tests/generated_class_test.php
TeBoring: This is copied from java generated file. Java interprets Unicode escape sequences anywhere (e.g., \t will be interpreted). Will php do that even in comment? If not, we can remove it.

bshaffer: PHP does not do this, no, so I will remove it. Thanks!

src/google/protobuf/compiler/php/php_generator.cc
bshaffer: We should make the type `int|string` as the functions can receive and return both.

TeBoring: How about other integer fields (int32, enum) which takes integer but allow type conversion via string?

src/google/protobuf/compiler/php/php_generator.cc
michaelbausor: I think that we do want this - the `@return` will be used by IDEs to provide autocompletion, and by documentation generation tools.

src/google/protobuf/compiler/php/php_generator.cc
michaelbausor: It seems correct to only have RepeatedField for getters if that is always returned. In the case of setters, I would argue that having both is useful, because users will potentially want to take a value from a getter and pass it into a setter.

TeBoring: SGTM

TeBoring:
Any update?

bazel-io:

src/google/protobuf/compiler/php/php_generator.cc
bshaffer: `int|string` would handle this use case as well, unless I am missing something.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1077,17 +1113,17 @@ static string FirstLineOf(const string& value) {
...
 void GenerateMessageDocComment(io::Printer* printer,
-                               const Descriptor* message) {
+                               const Descriptor* message, int is_descriptor) {
   printer->Print("/**\n");
   GenerateDocCommentBody(printer, message);
   printer->Print(
     " * Protobuf type <code>^fullname^</code>\n"
     " */\n",
-    "fullname", EscapePhpdoc(message->full_name()));
+    "fullname", EscapePhpdoc(PhpName(message->full_name(), is_descriptor)));
```
bshaffer: It's not necessary here no, I can remove it.

bshaffer: If I remove this though, there's nothing in the class PHPDoc comment

TeBoring: Prefer to keep the original class comment (the proto message definition)

php/src/Google/Protobuf/Internal/FileDescriptorProto.php
```diff
@@ -400,6 +418,7 @@ public function hasExtension()
 
     /**
      * <code>optional .google.protobuf.FileOptions options = 8;</code>
+     * @return Google\Protobuf\Internal\FileOptions
```
michaelbausor: The tags seem inconsistent about whether or not to include a leading `\` for fully qualified types. I think that we do want to include a leading `\` in all cases

michaelbausor:
@TeBoring @bshaffer What do you think about removing the `<pre>` tags from comments? (see https://github.com/google/protobuf/pull/3131/files#diff-cef3abbfd1e249d6102970e01118fdcfR1115). I am not sure what value they are adding versus just rendering the proto comments.

Also, what do you think about changing the `<code>` blocks on fields from this:
`<code>string project_name = 1;</code>`
to
`Generated from protobuf message field: <code>string project_name = 1;</code>`
or something similar, so it is clear what the code block represents?

michaelbausor:
I cannot submit a PR to your fork @bshaffer but I have created a commit with some proposed changes here: https://github.com/michaelbausor/protobuf/commit/b1c257f700d133da8b4cafb88cd27ecd29f41c2c

bshaffer:
@michaelbausor I merged your CL and updated the native class files to match.

bshaffer:
@TeBoring this is ready for another look!

TeBoring:
ok to test

TeBoring:
You may need to add the new added file to Makefile.am in root directory for cpp_distcheck.

TeBoring:
For <pre>, this tag allows formatted comment in proto files to be represented as it is in html page.

php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -12,66 +12,62 @@
 use Google\Protobuf\Internal\GPBUtil;
 
 /**
- * <pre>
  * Describes a message type.
- * </pre>
  *
- * Protobuf type <code>google.protobuf.DescriptorProto</code>
+ * Protobuf type <code>Google\Protobuf\Internal</code>
```
TeBoring: How about Generated from protobuf message <code>Google\Protobuf\Internal</code>

php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -132,15 +132,17 @@ public function hasField()
...
      */
     public function getExtension()
     {
         return $this->extension;
     }
 
     /**
-     * <code>repeated .google.protobuf.FieldDescriptorProto extension = 6;</code>
+     * Generated from protobuf field <code>repeated .google.protobuf.FieldDescriptorProto extension = 6;</code>
+     * @param array|\Google\Protobuf\Internal\RepeatedField $var
```
TeBoring: Do we need to add comment for setter? We now have fluent setters.

php/tests/generated_phpdoc_test.php
```diff
@@ -0,0 +1,300 @@
+<?php
+
+require_once('generated/NoNamespaceEnum.php');
+require_once('generated/NoNamespaceMessage.php');
+require_once('test_base.php');
+require_once('test_util.php');
+
+// use Google\Protobuf\Internal\RepeatedField;
```
TeBoring: Remove not used message?

php/tests/generated_phpdoc_test.php
```diff
@@ -0,0 +1,300 @@
...
+                    'getMapSint32Sint32',
+                    'getMapSint64Sint64',
+                    'getMapFixed32Fixed32',
+                    'getMapFixed64Fixed64',
+                    'getMapSfixed32Sfixed32',
+                    'getMapSfixed64Sfixed64',
+                    'getMapInt32Float',
+                    'getMapInt32Double',
+                    'getMapBoolBool',
+                    'getMapStringString',
```
TeBoring: Map field return value should be MapField instead of RepeatedField

michaelbausor:
@TeBoring Re `<pre>`, that makes sense. But, is it necessary to have the comments render correctly as html? The use cases where someone is reading these comments are likely to be:
- In an editor, where they will appear as-is
- On github, where they will appear as-is
- In a generated documentation site. In the case of the google-cloud-php doc site, the `<pre>` tags interfere with the docs we would generate and render. I'm not saying we should optimize protobuf just for google-cloud-php, but is there a specific doc-site generation that we are optimizing for? Or just the general case? And if it is just the general case, is it normal to include html tags in documentation that is going to be parsed into a documentation site?

php/tests/generated_phpdoc_test.php
```diff
@@ -0,0 +1,300 @@
...
+                    'setMapSint32Sint32',
+                    'setMapSint64Sint64',
+                    'setMapFixed32Fixed32',
+                    'setMapFixed64Fixed64',
+                    'setMapSfixed32Sfixed32',
+                    'setMapSfixed64Sfixed64',
+                    'setMapInt32Float',
+                    'setMapInt32Double',
+                    'setMapBoolBool',
+                    'setMapStringString',
```
TeBoring: MapField Instead

php/tests/generated_phpdoc_test.php
bshaffer: done

php/tests/generated_phpdoc_test.php
bshaffer: done

php/src/Google/Protobuf/Internal/DescriptorProto.php
bshaffer: added

php/src/Google/Protobuf/Internal/DescriptorProto.php
bshaffer: done

TeBoring: Could you remove the line for "Protobuf type < code>Google\Protobuf\Internal</ code>"

TeBoring:
@michaelbausor we can add it back later if needed.

php/src/Google/Protobuf/Internal/DescriptorProto.php
bshaffer: Done.

php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -307,25 +330,24 @@ public function hasReservedRange()
...
 
     /**
-     * <pre>
      * Reserved field names, which may not be used by fields in the same message.
      * A given name may only be reserved once.
-     * </pre>
      *
-     * <code>repeated string reserved_name = 10;</code>
+     * Generated from protobuf field <code>repeated string reserved_name = 10;</code>
+     * @param array|\Google\Protobuf\Internal\RepeatedField $var
```
michaelbausor: @bshaffer @TeBoring What do you think about replacing `array|\Google\Protobuf\Internal\RepeatedField` with `string[]|\Google\Protobuf\Internal\RepeatedField`? This is more descriptive, since it accepts specifically an array of strings.

Similarly for other types, we could generate `\Fully\Qualified\Type[]|\Google\Protobuf\DescriptorProto[]`

This is longer, but means that the type to be used is available

TeBoring: SGTM, as long as string[] is ok in php document.
Do you mean `\Fully\Qualified\Type[]|\Google\Protobuf\Internal\RepeatedField`?

michaelbausor: That is what I meant, thanks :-)

The `[]` notation is supported by phpdoc: https://www.phpdoc.org/docs/latest/guides/types.html#arrays

bshaffer: done!

bshaffer: Github's [experiencing some delays](https://twitter.com/githubstatus/status/879397164673368066), but this is added in 2f1c48ad0acb677dee9d7853de9e6b53406125c6

Issue #3139, opened 2017-05-24
PHP: serialization of Value type


michaelbausor:
The `Value` type generated from https://github.com/google/protobuf/blob/master/src/google/protobuf/struct.proto#L63 does not distinguish between having no values set and having the null value set. So this PHP code:
```
$v = new Value();
$v->setNullValue(NullValue::NULL_VALUE);
echo "null:" . bin2hex($v->serializeToString()) . "\n";
$v2 = new Value();
echo "empty:" . bin2hex($v2->serializeToString()) . "\n";
```
generates this output:
```
null:
empty:
```

Compare this to the following Java code:
```
Value v = Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build();
System.out.println("null:" + BaseEncoding.base16().encode(v.toByteArray()));
Value v2 = Value.newBuilder().build();
System.out.println("empty:" + BaseEncoding.base16().encode(v2.toByteArray()));
```
which generates:
```
null:0800
empty:
```

This is a problem if an API will accept a null value but reject a value with nothing set (this is currently the case for Spanner).

michaelbausor:
Update: this seems to happen on all oneofs. Specifically, if a oneof field is set to the default value, it is not sent. This seems to be a difference in behaviour between PHP protobuf and other languages.

TeBoring:
I think this is a bug for serializing oneof field. I'll fix it soon.

PR #3153, opened 2017-05-29
Oneof field should be serialized even it's equal to default.

TeBoring:
#3139


bazel-io:

TeBoring:
ok to test

Issue #3122, opened 2017-05-22
PHP: support proto option to specify namespace


michaelbausor:
Similar to other languages, it would be useful to specify the namespace of the generated protobuf classes using an option in the proto file. See for example: https://github.com/googleapis/googleapis/blob/master/google/logging/v2/logging.proto#L28 which specifies namespace/package options for C#, Go and Java.

TeBoring:
Fixed in #3162. Will be released in 3.4. ETA end of July.

PR #3162, opened 2017-05-31
Add new file option php_namespace.

TeBoring:
Use this option to change the namespace of php generated classes.
Default is empty. When this option is empty, the package name will be
used for determining the namespace.


bazel-io:

TeBoring:
#3122

TeBoring:
@michaelbausor

php/tests/array_test.php
```diff
@@ -798,153 +798,153 @@ public function testMessageSetStringFail()
...
-        $start = memory_get_usage();
-        unset($arr);
-
-        // Explicitly trigger garbage collection.
-        gc_collect_cycles();
-
-        $end = memory_get_usage();
-        $this->assertLessThan($start, $end);
-    }
+#     /**
```
michaelbausor: Are these tests supposed to be commented out?

TeBoring:
Thanks for the catch.

php/tests/array_test.php
TeBoring: Fixed.

michaelbausor:
Testing locally, I am finding that this does not change the namespace of the GPBMetadata files. Is that intentional? I do not know of strong arguments either for changing it or not.

TeBoring:
I don't think we need to change the namespace of GPBMetadata files. These files are transparent to users.

TeBoring:
BTW, the namespace of GPBMetadata files are decided by the path of proto file instead of the package name.

heyanlong:
@TeBoring What time php_namespace. protoc 3.4?

PR #3169, opened 2017-06-01
Optimize GPBDictionary.m codegen to reduce size of library

dmaclach:
This reduces the size of the protobuf library by about ~80k per architecture (compared by building the release .a files without debug symbols and comparing the sizes).


thomasvl:
ok to test

Issue #3166, opened 2017-06-01
[Proposal] Async version of CSharp protobuf


mkosieradzki:
I would like to implement support for async/await versions of all methods dealing with I/O.

## Rationale
Mixing async and sync code in C# is very inefective thread-wise. It is a good practise to have entire code top-down async, but the worst case-scenario is when synchronous code calls async code like in case of Stream.Read and Stream.Write.

Today the only way to efficiently use protobuf in async code is to copy buffers to the memory and then perform deserialization synchronously. What makes it virtually impossible to use protobuf in streaming scenarios with near-to-constant memory.

## Background
Lately there have been many improvements to the TPL to allow more efficient (allocation-wise) handling of many scenarious especially parser-like where most of methods return synchronously. ValueTask and Task.CompletedTask are the most important of those improvements.

## High-level plan
1. Duplicate all I/O-bound methods in the following manner:
ReturnType MethodName(Arguments) 
to
async ValueTask<ReturnType> MethodNameAsync(Arguments, CancellationToken cancellationToken)

void MethodName(Arguments)
to
async Task MethodNameAsync(Arguments, CancellationToken cancellationToken)

And all references in async methods to those methods with await CalledMethodAsync(..., cancellationToken)

This way we have perfectly correct implementation (but not optimal, especially allocation-wise).

Replace all calls to Stream.Read/Write with proper ReadAsync/WriteAsync

2. After testing this version start optimizations:
a. Remove constructs like async await = where possible
b. Try to use as much of ValueTask<T> and Task.CompletedTask as possible.
c. Try to inline some async calls to avoid generation of separate state-machines

3. Modify code-generator to generate async variants.

I would be happy to create PR for this proposal.

However I believe optimization will be a bit longer process, going far beyond this single PR.

jskeet:
I would strongly suggest not using `ValueTask` or any C# 7 features that require additional dependencies. Any dependencies we add - or additional requirements for more recent frameworks - will add a toll in terms of maintenance and may reduce the potential audience. Creating regular async methods for parsing is more reasonable, although it's tricky in terms of the *generated* code... if any code needs to be generated with async methods, that would probably need to be specified via an option, as otherwise it could very easily break existing users. (And it seems likely - without having looked at this at all yet in detail - that it really *would* require additions to generated code.

More worryingly, it may require additions to interfaces, which basically can't happen due backwards compatibility. You may well want to look at an `IAsyncMessage` interface or similar.

[...]

PR #3173, opened 2017-06-02
Support for async in C# version

mkosieradzki:
This is PR for #3166


bazel-io:

jskeet:
I've assigned this to myself to look at, but I'm currently on vacation, and when I get back from vacation I'm likely to have a lot of work for my main projects. If you ping when it's ready for review, I'll try to look at it when I can.

mkosieradzki:
@jskeet Thanks a lot, I appreciate it.

I am a bit behind, because I had to setup two C++ environments (VS2017 for development and for protobuf code generation - Bash on Ubuntu on Windows), it took me a bit as I haven't coded in C++ for years.

mkosieradzki:
@jskeet I have commited most of the planned code. I have done **zero** testing yet.
To enable the new feature you can use either:
```
option csharp_async = true;
```
or
`--csharp_opt=async`

Still this PR is quite HUGE.

bazel-io:

csharp/generate_protos.sh
```diff
@@ -28,6 +28,7 @@ fi
 # descriptor.proto and well-known types
 $PROTOC -Isrc --csharp_out=csharp/src/Google.Protobuf \
     --csharp_opt=base_namespace=Google.Protobuf \
+    --csharp_opt=async \
```
jskeet: I believe these can be comma-separated, e.g.

```text
--csharp_opt=base_namespace=Google.Protobuf,async
```

Although I'd be tempted to use "async=true" - I'll need to research what other options do for this.

csharp/src/Google.Protobuf.Test/AsyncOnlyStreamWrapper.cs
```diff
@@ -0,0 +1,48 @@
+﻿#if !PROTOBUF_NO_ASYNC
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Protobuf
```
jskeet: Was going to suggest we moved it to Google.Protobuf.Test, but apparently everything's in Google.Protobuf. Apologies for that :(

csharp/src/Google.Protobuf/FieldCodec.Async.cs
```diff
@@ -0,0 +1,164 @@
+﻿#region Copyright notice and license
+// Protocol Buffers - Google's data interchange format
+// Copyright 2015 Google Inc.  All rights reserved.
```
jskeet: Let's update this to 2017 :)

csharp/src/Google.Protobuf/IAsyncMessage.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// Extension interface for a Protocol Buffers message, supporting
+    /// basic asynchronous operations required for serialization.
+    /// </summary>
+    public interface IAsyncMessage : IMessage
+    {
+        /// <summary>
+        /// Merges the data from the specified coded input stream with the current message.
+        /// </summary>
+        /// <remarks>See the user guide for precise merge semantics.</remarks>
+        /// <param name="input"></param>
```
jskeet: Please add XML doc comments for these parameters (here and below)

csharp/src/Google.Protobuf.Test/CodedInputStreamExtensions.cs
```diff
@@ -49,5 +49,15 @@ public static T ReadMessage<T>(this CodedInputStream stream, MessageParser<T> pa
             stream.ReadMessage(message);
             return message;
         }
+
+#if !PROTOBUF_NO_ASYNC
+        public static T ReadMessage<T>(this CodedInputStream stream, AsyncMessageParser<T> parser)
```
jskeet: Was this intended to be the async one? If not, it's not clear to me why we need this. (I can vaguely imagine, but can't quite see it - whereas I can definitely see why we'd want ReadMessageAsync.)

mkosieradzki: This is no longer necessary due to our change in the class hierarchy. Because now AsyncMessageParser<T> inherits MessageParser<T> and the original is valid.. I will remove this.

mkosieradzki:
I can definitely try to split this commit. If you really think it is worth.
1. C++ only and regenerate proto C# files adding this one error pragma.
2. Add C# library changes (I hope they will compile despite the lack of async code in wellknown types) - but I am not sure here.
3. Enable async option in build scripts and C++ conformance tests and regenerate all C# proto files

Is this what you mean?

I understand it does not hurt very much if 2 does not compile.

Should I create 3 separate PRs? or single PR with 3 commits and you will just not squash it during the final merge?

jskeet:
Just one commit with 3 PRs. The point is to make it far, far simpler to review them one commit at a time. There's no point in me reviewing the autogenerated code, but isolating the generated from the non-generated code in the github review UI is tricky.

That's why commit 1 shouldn't include any regenerated protos - it should be *purely* the manual C++ code. Commit 2 still shouldn't include any regenerated protos - it should be *purely* the manual C# code. Commit 3 will include regenerated descriptor/well-known-type/test protos.

And yes, when the review is complete, I'll merge onto master with a rebase.
(I'll want the protobuf team to review the whole feature as well, once I've done the C#-specific pass. I expect it'll be fine, but I think it's worth getting an extra pair of eyes.)

mkosieradzki:
>Just one commit with 3 PRs

You mean one PR with 3 commits? Or 3 PRs with 1 commit each :)? I apologize but I am not VERY familiar with all the github workflows.

jskeet:
One PR, 3 commits.

mkosieradzki:
@jskeet I have isolated this to 6 verifiable and buildable commits:
1. Async option + pragma [should build, will not pass due to proto mistmatch (pragma)]
2. [autogenerated]Regenerated with pragma [should pass]
3. Added async support to C# library [sholud pass]
4. Enabled async option while generating protos for C# library [should build, will not pass]
5. [autogenerated]Regenerated with async enabled[should pass] - coming soon
6. Added tests for async in C# library[should pass] - coming soon (tests are last because they depend on async protos generated) - coming soon

jskeet:
Thanks - will try to review over the next few days.

mkosieradzki:
I have rebased all my commits to match the specified order. ;)

jskeet:
Just to say, I haven't forgotten about this. I've just had many other huge code reviews to do this week :(

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -109,24 +110,62 @@ void MessageGenerator::Generate(io::Printer* printer) {
   vars["class_name"] = class_name();
   vars["access_level"] = class_access_level();
 
+  //It is critical that async class is declared first because static members must be declared BEFORE nested types
```
jskeet: Tiny nit: add a space after `//`

I'd also expand the comment... the *compiler* doesn't force you to declare static members before nested types. Is the important part the initialization order? If so, I'd say that:

```csharp
// It is critical that the async class is declared first to ensure the parser is initialized first during
// type initialization.
```

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -109,24 +110,62 @@ void MessageGenerator::Generate(io::Printer* printer) {
...
   AddDeprecatedFlag(printer);
-  
   printer->Print(
     vars,
     "$access_level$ sealed partial class $class_name$ : pb::IMessage<$class_name$> {\n");
   printer->Indent();
 
   // All static fields and properties
+
+  //In async mode this field will be generated in Async-class part. Condition is added to avoid duplicate parsers.
```
jskeet: Again, please add a space after the `//`, just for consistency and readability.

src/google/protobuf/compiler/csharp/csharp_wrapper_field.cc
```diff
@@ -196,6 +219,15 @@ void WrapperOneofFieldGenerator::GenerateSerializationCode(io::Printer* printer)
     "}\n");
 }
 
+void WrapperOneofFieldGenerator::GenerateAsyncSerializationCode(io::Printer* printer) {
+  // TODO: I suspect this is wrong...
```
jskeet: It's fine, I believe. I should remove the comment elsewhere, but let's not include this one :)

csharp/src/Google.Protobuf/AsyncMessageExtensions.cs
```diff
@@ -0,0 +1,126 @@
...
+
+#if !PROTOBUF_NO_ASYNC
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Extension methods on <see cref="IMessage"/> and <see cref="IMessage{T}"/>.
```
jskeet: Nit: fix comment to talk about IAsyncMessage.

csharp/src/Google.Protobuf/AsyncMessageExtensions.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// </summary>
+    public static class AsyncMessageExtensions
+    {
+        /// <summary>
+        /// Merges data from the given byte array into an existing message.
+        /// </summary>
+        /// <param name="message">The message to merge the data into.</param>
+        /// <param name="data">The data to merge, which must be protobuf-encoded binary data.</param>
+        /// <param name="cancellationToken"></param>
```
jskeet: Nit - please document these cancellation tokens. Elsewhere I've used:

```csharp
/// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
```

You might want to *consider* defaulting to `default(CancellationToken)` (equivalent to `CancellationToken.None`).

csharp/src/Google.Protobuf/AsyncMessageExtensions.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// <param name="message">The message to merge the data into.</param>
+        /// <param name="input">Stream containing the data to merge, which must be protobuf-encoded binary data.</param>
+        /// <param name="cancellationToken"></param>
+        public static async Task MergeDelimitedFromAsync(this IAsyncMessage message, Stream input, CancellationToken cancellationToken)
+        {
+            ProtoPreconditions.CheckNotNull(message, "message");
+            ProtoPreconditions.CheckNotNull(input, "input");
+            int size = (int) await CodedInputStream.ReadRawVarint32Async(input, cancellationToken).ConfigureAwait(false);
+            Stream limitedStream = new LimitedInputStream(input, size);
+            await message.MergeFromAsync(limitedStream, cancellationToken);
```
jskeet: ConfigureAwait(false)

It's probably worth adding a checker for this... put this in the project file:

```xml
    <PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="1.0.0-beta4">
      <PrivateAssets>All</PrivateAssets>
    </PackageReference>
```

csharp/src/Google.Protobuf/AsyncMessageParser.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// A general async message parser, typically used by reflection-based code as all the methods
+    /// return simple <see cref="IAsyncMessage"/>.
+    /// </summary>
+    public interface IAsyncMessageParser
```
jskeet: Do we actually need this at all, out of interest? Perhaps we'll need it for reading JSON asynchronously?

csharp/src/Google.Protobuf/AsyncMessageParser.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <param name="data">The byte array containing the message. Must not be null.</param>
+        /// <returns>The newly parsed message.</returns>
+        IAsyncMessage ParseFrom(ByteString data);
+
+        /// <summary>
+        /// Parses a message from the given stream.
+        /// </summary>
+        /// <param name="input">The stream to parse.</param>
+        /// <returns>The parsed message.</returns>
+        IAsyncMessage ParseFrom(Stream input);
```
jskeet: Shouldn't this (and the next two methods) be asynchronous?

csharp/src/Google.Protobuf/AsyncMessageParser.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <param name="cancellationToken"></param>
+        /// <returns>The parsed message.</returns>
+        public async Task<T> ParseFromAsync(Stream input, CancellationToken cancellationToken)
+        {
+            T message = CreateTemplate();
+            await message.MergeFromAsync(input, cancellationToken).ConfigureAwait(false);
+            return message;
+        }
+
+        IAsyncMessage IAsyncMessageParser.ParseFrom(byte[] data) => ParseFrom(data);
```
jskeet: Let's add `/// <inheritdoc />` to all of these.

csharp/src/Google.Protobuf/ByteString.cs
```diff
@@ -382,6 +382,10 @@ internal void WriteRawBytesTo(CodedOutputStream outputStream)
             outputStream.WriteRawBytes(bytes, 0, bytes.Length);
         }
 
+#if !PROTOBUF_NO_ASYNC
+        internal Task WriteRawBytesToAsync(CodedOutputStream outputStream, CancellationToken cancellationToken) => outputStream.WriteRawBytesAsync(bytes, 0, bytes.Length, cancellationToken);
```
jskeet: Line break after the `=>`?

csharp/src/Google.Protobuf/AsyncMessageExtensions.cs
jskeet: If you know of a checker to make sure that every asynchronous call has a cancellation token, that would be good too...

csharp/src/Google.Protobuf.Test/CodedInputStreamAsyncTest.cs
```diff
@@ -0,0 +1,499 @@
...
+            byte[] bytes = new byte[bytesAsInts.Length];
+            for (int i = 0; i < bytesAsInts.Length; i++)
+            {
+                bytes[i] = (byte) bytesAsInts[i];
+            }
+            return bytes;
+        }
+
+        /// <summary>
+        /// Parses the given bytes using ReadRawVarint32() and ReadRawVarint64()
```
jskeet: Well, using ReadVarint32Async and ReadRawVarint64Async. Possibly just add a comment at the top of the file saying that they're the direct async equivalents of the sync tests?

csharp/src/Google.Protobuf.Test/CodedInputStreamAsyncTest.cs
```diff
@@ -0,0 +1,499 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Google.Protobuf.TestProtos;
+using NUnit.Framework;
+
+namespace Google.Protobuf
+{
+    public class CodedInputStreamAsyncTest
+    {
+        /// <summary>
```
jskeet: This shouldn't be necessary (in the sync or async code). Could you add a TODO(jonskeet) to remove these? I'll tidy up afterwards.

jskeet:
Right, I've looked through it all now, and I'm pretty happy. There are a few nits as I've commented on, and I'm not sure of the need of the non-generic message/parser interfaces at all, but I strongly suspect we'll want them (at least to some extent) when we implement asynchronous JSON parsing. We might as well be consistent with the synchronous code rather than trying to be absolutely minimal.

If you could fix up the bits I've mentioned, we can then talk to members of the "regular"protobuf team about any further concerns before merging. (I think it's worth doing, although it does definitely introduce more complexity.)

csharp/src/Google.Protobuf/AsyncMessageExtensions.cs
mkosieradzki: Thanks for the cancellationToken comment text ;)
I thought about using providing default value, but I am not very convinced. It is generally a good practice to use cancellation tokens, especially in .NET Core where there is no thread abort.

csharp/src/Google.Protobuf/AsyncMessageParser.cs
mkosieradzki: I think it is at least convenient to be able to have an untyped parser interface, so I would definitely like to keep this one.

mkosieradzki:
@jskeet Thanks a lot for you review and comments. I am applying appropriate changes to my code, however do you have any idea why builds started to fail? I think it might be unrelated to this PR.

jskeet:
I think the problem is that I merged another PR which changed the `CodedInputStream` constructor signature. I suggest you rebase with master and then you should be able to fix the problems easily.

mkosieradzki:
Thanks.

csharp/src/Google.Protobuf/AsyncMessageExtensions.cs
mkosieradzki: Thanks for this analyzer. It's very helpful.

csharp/src/Google.Protobuf/AsyncMessageParser.cs
mkosieradzki: Hmm.
There are many solutions here:
1. We can create both versions: ParseFrom and ParseFromAsync for stream etc. (but that made me thinking...)
2. We can decrease the parser interface surface and add all the Parse* convenience methods as extension methods. That would only require exposing message instantiation (factory) on the parser interface because the heavy lifting is ALWAYS done by the message.
Currently the true job of the MessageParser is really instantiating the message. So message parser is apparently nothing more than a factory ;).

Please let me know which approach should I take.

jskeet: Message instantiation is already present in MessageParser internally in the CreateTemplate method. But that can't be in an interface, of course.

I think it's probably simplest to keep this as close to the synchronous parser as possible, to be honest - while it may have been a design mistake to have all of these methods to start with, we can't change that now, so we should probably keep things as similar as possible (barring this now being an interface, of course).

mkosieradzki: OK. Thanks

csharp/src/Google.Protobuf.Test/CodedInputStreamAsyncTest.cs
mkosieradzki: I'm sorry, I don't understand what do you mean here.

jskeet: The `Bytes` method shouldn't be necessary at all. I don't know whether I added it a long time ago or not, but we should remove it. Please just add a comment of

```csharp
// TODO(jonskeet): Remove this and the equivalent in the synchronous tests
```

That should remind me to do it...

jskeet:
Hmm... I think the rebase has gone a bit wrong. We now have 27 commits. Are you comfortable fixing this up?

mkosieradzki:
Yep, I am dealing with it right now :).

mkosieradzki:
Ok. This went really wrong I will try to rebase this in couple of days. Have to do different stuff right now.

mkosieradzki:
@jskeet I think I have managed to repair this using cherry picks. I just cannot guarantee something haven't been lost during all those operations.

jskeet:
Loks like you've fixed it... let's see what CI does with it. Once that's green (or as green as we expect to get to) I'll get protobuf team members in to check as well. We'll get there - thanks for your patience!

mkosieradzki:
Unfortunately something went wrong. However I think I will be able to fix it a bit later. Now I have a plane to catch :).

mkosieradzki:
Fixed this on the plane ;). I think it should work now.

jskeet:
Green is good :) Right, will reach out to the proto team.

anandolee:
Thank mkosieradzki@ and jskeet@ working on this.
mkosieradzki@, it is a large change, are you able to write a API design doc? You can use google doc and shared with jskeet@ and me.
Protobuf has lots users that APIs should be very reasonable and stable

jskeet:
I'm happy to write at least the first draft of the design doc (knowing what Googlers are likely to be looking for :). Will link to it here, and @mkosieradzki can review and add as well.

mkosieradzki:
@jskeet Thank you very much, I really appreciate it, especially that I am not familiar with Google conventions for design docs (but I would be definitely happy to learn and contribute to the document).

BTW. I would definitely suggest not freezing async APIs in Coded(Input|Output)Stream yet as it might be a subject to change due to some perf-work (async-await often requires function reorganization to limit heap allocations much more than classical functions). We have also some pending discussions on ValueTasks and we would also definitely like to have even more emphasize on fast-path handling.

Rest of the APIs should be quite stable...

jskeet:
First pass at design doc:
https://docs.google.com/document/d/1SMoJRimQnWB3PfYqfg_dEwWlUYGOIs3DUwJWxngoCJY/edit#
PTAL.

mkosieradzki:
Thanks a lot. I really appreciate your further research in the ValueTask area! I will add some comments soon.

csharp/src/Google.Protobuf.Test/CodedInputStreamAsyncTest.cs
```diff
@@ -0,0 +1,502 @@
+#region Copyright notice and license
```
anandolee: Most tests are similar with the sync version. We'd better to make the them share the test code. So if
we want to change some code in future, the share test code can remind us to change both sync and async.

jskeet: How would you suggest doing that? I suspect that trying to reduce the duplication will increase the complexity overall. I'm all for it if you can think of a good way of doing it.

mkosieradzki: The main problem is that async code has completely different semantics (continuation) despite it looks very similar. This is the most important practical limitation of async-await: to make it efficient you need to rewrite everything from the top to the bottom using continuations. Whenever you are using an sync-async bridge like: GetAwaiter().GetResult() or using IO-bound sync operation from async code you are blocking a threadpool thread.

However if you have any ideas how to to this I would be happy to improve this code.

anandolee: I do not have a very good idea. How about some help methods which take an isAsync option to call CodedInputStream/CodedOutputStream APIs. For example :
uint ReadRawVarint32(CodedInputStream input, bool isAsync)
{
    return isAsync ? (await input.ReadRawVarint32Async(CancellationToken.None)) : input.ReadRawVarint32();
}

Does this work? 
For C++ we can use macro, I think C# does not support macro?

mkosieradzki: Not really... This would require `ReadRawVarint32` to be an `async` method and return `Task<uint>`.
`async` means: use continuations semantics (you can use `await` only in `async` methods).

mkosieradzki: If only C# supported macros, a macro/template would definitely work here. However c# does not support macros/templates - so basically it is impossible to introduce sync/async polymorphism.

anandolee: LGTM

anandolee:
The change is good to me. jskeet@ do you have more comments?

jskeet:
There are a few things I think we need to do:

- Decide once and for all about using ValueTask instead of Task. I'm leaning towards allowing the dependency, as it's a system-level dependency. (I'd be astonished if MS broke it.) If we go with ValueTask, that means several changes are required in the PR, although they should be simple to do.
- Go through my design doc to check that everything has been resolved
- Check for changes in `Coded*Stream` and tests since this PR was created - IIRC, I fixed something, so we'd need to propagate that
- At least write draft docs, possibly based on the design doc

I'm on holiday (mostly) this week, but I'll see what I can do next week.

csharp/src/Google.Protobuf.Test/CodedInputStreamAsyncTest.cs
fabian-f: As with all streams, the .net standard is to have the legacy blocking methods, and then have the same signature but async keyword and their named suffixed with "Async". There's no need for all this switching.

Groostav:
Bump, I'd really like to see this as:

1. it simplifies some icky thread hopping code I've got in a plugin for a winforms app
2. it would offer a precedence for things like using streams/Flow/rx/reactive/kotlin's-coroutines for the java ecosystem.

csharp/src/Google.Protobuf/AsyncMessageParser.cs
```diff
@@ -0,0 +1,219 @@
...
+        /// The stream is expected to contain a length and then the data. Only the amount of data
+        /// specified by the length will be consumed.
+        /// </remarks>
+        /// <param name="input">The stream to parse.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>The parsed message.</returns>
+        public async Task<T> ParseDelimitedFromAsync(Stream input, CancellationToken cancellationToken)
+        {
+            T message = CreateTemplate();
+            await message.MergeFromAsync(input, cancellationToken).ConfigureAwait(false);
```
mfarrell34: Shouldn't this use MergeDelimitedFromAsync instead?

mkosieradzki: Yup it should. Thanks for finding this out!

mkosieradzki:
@xfxyjwf My apologies, but I am not very enthusiastic to do a manual merge if the PR is not going to be accepted soon. I did this several times already.

Most of the conflicts are in the automatically generated code... So it basically means I will need to rerun the tools and do some interactive rebasing...

So let's wait for a "political decision" whether we should merge this one, or should we take another approach. As, despite being the author, I am not super-convinced that this is the way to go.

As discussed here: #3166 and here #3431 ...

I am currently leaning towards a different approach with either:
- Pipelines
- Prefetch (buffer the entire message) + `ReadOnlySequence` of `ReadOnlySpan<T>`

Just to remind some important points:
- `Span<T>` does not like Streams - this means we should redesign the way data is passed for parsing
- `Span<T>` does not like async - this means we cannot mix async and `Span<T>` (Span<T> is stackallocated)
- `async` is quite expensive alloc-wise - this means we should at least consider using ValueTask or go with trivial approach with prefetch.

Pipelines are much better if we want to push data into parser and do parsing as the data comes...
ReadOnlySequence with prefetch can be less prone to the DDoS-like attacks (as parsing is done only after the entire message is received) and should work much nicer with gRPC which has a very neat concept of streaming.

When .NET Core 2.1 is live (probably after Microsoft Build conference) I can create a new "stream-less" and "async-less" `Span<T>`-based parser prototype - with async and stream support based on the prefetch.

This will be a quite huge effort as it requires rewriting everything, but I predict that the resulting parser performance will be quite decent.

xfxyjwf:
@mkosieradzki Thanks for contributing to protobuf but from my understanding of the current status, it's very unlikely we will merge this PR soon. As I see it the current issues are:
1. Uncertainty of the technical direction. A lot unknowns:
  a. Does it provide enough performance gain that worths the effort and complexity?
  b. Do we want Task<T> or ValueTask<T>?
  c. What about other options? Pipelines/Spans/etc.
  It does not look to me there is a clear win with the proposal (or any of the alternatives) and there isn't a clear technical path either.
2. Lack of engineering resource from protobuf team. Changes as significant as this must be supported by protobuf-team but at the moment neither Jon nor Jie has enough time to work on this and I don't expect anyone else from protobuf team will work on this in the near future.

I'm sorry to say this but given these situations I believe this PR is not actionable at the moment. It doesn't mean the proposal is rejected though. We just don't have the resource to investigate and make the decision.

mkosieradzki:
@xfxyjwf That's fully understandable. I have started another fork to investigate the other options and explore the larger picture including gRPC integration. I will let you know when it's ready.

The path I am exploring is: ValueTask + Pipelines + Spans.

I think I have quite a clear vision right now. I think I will reuse some code from this PR.

xfxyjwf:
Closing for now. Feel free to continue the discussion in the issue thread. Before you send another pull request, please makes sure @anandolee has agreed on the design and has time to review the change.

Issue #3114, opened 2017-05-19
Java: Invalid code generated when referencing a file with no package statement


thesamet:
file1.proto:

```
syntax = "proto2";

message Foo { }
```

file2.proto:

```
syntax = "proto2";

import "file1.proto";

message Bar {
  optional Foo foo = 1;
}
```

In File2.java you'll find:
```
   com.google.protobuf.Descriptors.FileDescriptor
      .internalBuildGeneratedFileFrom(descriptorData,
        new com.google.protobuf.Descriptors.FileDescriptor[] {
          .File1.getDescriptor(),         // <-------------- ERROR HERE
        }, assigner);
```

Compiler error:

```
File2.java:621: error: illegal start of expression
          .File1.getDescriptor(),
          ^
```

acozzette:
This should now be fixed with pull request #3176.

PR #3176, opened 2017-06-02
Ensure that for Java, imports of .proto files with empty packages works

acozzette:
This fixes a compiler bug that caused a Java syntax error when one
.proto file would import another one with an empty package and
java_package. This fixes issue #3114.


acozzette:
retest this please

PR #3178, opened 2017-06-02
Reduce size of GPBDictionary by getting rid of class creation methods

dmaclach:
By my quick and dirty back of the envelope calculations, this reduces the size of the libraries by about 80k per architecture.

It removes +dictionary... methods which are all just syntactic sugar for alloc] init...

### DIFF
--- a/objectivec/GPBDictionary.h
+++ b/objectivec/GPBDictionary.h
@@ -60,55 +60,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, readonly) NSUInteger count;
 
 /**
- * @return A newly instanced and empty dictionary.
- **/
-+ (instancetype)dictionary;
-
-/**
- * Creates and initializes a dictionary with the single entry given.
- *
- * @param value     The value to be placed in the dictionary.
- * @param key       The key under which to store the value.
- *
- * @return A newly instanced dictionary with the key and value in it.
- **/
-+ (instancetype)dictionaryWithUInt32:(uint32_t)value
-                              forKey:(uint32_t)key;
-
-/**
- * Creates and initializes a dictionary with the entries given.
- *
- * @param values      The values to be placed in the dictionary.
- * @param keys        The keys under which to store the values.
- * @param count       The number of entries to store in the dictionary.
- *
- * @return A newly instanced dictionary with the keys and values in it.
- **/
-+ (instancetype)dictionaryWithUInt32s:(const uint32_t [__nullable])values
-                              forKeys:(const uint32_t [__nullable])keys
-                                count:(NSUInteger)count;
-
-/**
- * Creates and initializes a dictionary with the entries from the given.
- * dictionary.
- *
- * @param dictionary Dictionary containing the entries to add to the dictionary.
- *
- * @return A newly instanced dictionary with the entries from the given
- *         dictionary in it.
- **/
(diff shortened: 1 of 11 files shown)


thomasvl:
Ok to test

objectivec/GPBDictionary.h
```diff
@@ -60,55 +60,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, readonly) NSUInteger count;
 
 /**
- * @return A newly instanced and empty dictionary.
- **/
-+ (instancetype)dictionary;
```
thomasvl: I wonder if we should keep this one…

objectivec/GPBDictionary.h
```diff
@@ -60,55 +60,6 @@ NS_ASSUME_NONNULL_BEGIN
...
- *
- * @param values      The values to be placed in the dictionary.
- * @param keys        The keys under which to store the values.
- * @param count       The number of entries to store in the dictionary.
- *
- * @return A newly instanced dictionary with the keys and values in it.
- **/
-+ (instancetype)dictionaryWithUInt32s:(const uint32_t [__nullable])values
-                              forKeys:(const uint32_t [__nullable])keys
-                                count:(NSUInteger)count;
```
thomasvl: …and this one, as these two likely would be the most useful ones?

objectivec/GPBDictionary.h
dmaclach: any particular reason? alloc/init appears to be fine for most classes.

objectivec/GPBDictionary.h
dmaclach: we have  an alloc/init equivalent. I'd rather not have all clients pay for the binary size.

objectivec/GPBDictionary.h
dmaclach: alloc/init works for most other classes. I don't see a need for it. If people want real short hand they can use "new"

objectivec/GPBDictionary.h
dmaclach: we have an alloc/init equivalent. I'd prefer not to penalize everybody with class methods.

objectivec/GPBDictionary.h
thomasvl: Doing some simple searching, this one seemed somewhat common.  That's why I wasn't sure.

objectivec/GPBDictionary.h
thomasvl: I forget if this one came up in searches.  I guess it likely isn't as common.

objectivec/GPBDictionary.h
dmaclach: Your call... would you prefer we left it? I am happier going with consistency (all alloc init) as opposed to having the one "weird" class method.

objectivec/GPBDictionary.m
```diff
@@ -596,39 +596,6 @@ void GPBDictionaryReadEntry(id mapDictionary,
 //%  NSMutableDictionary *_dictionary;
```
thomasvl: You'll need to double check this by hand, I think it was Bool -> Enum that isn't generated by PDDM, so you likely have to hand edit it.  (it would be a single case, so it wasn't clear if doing macros for it was worth it.

PR #3186, opened 2017-06-05
Remove unused output_file variable from js_embed

gkelly:
The js_embed tool outputs to stdout, so the output_file variable is
unnecessary and unused.


jyane:
Any updates? Why is this PR not merged?

PR #3205, opened 2017-06-08
Make sure directory exists for js well known types

jynnantonix:
If the build directory is not the same as the source directory then
the build fails because the google/protobuf/compiler/js directory
doesn't exist.  Make sure that it does.


bazel-io:

PR #3215, opened 2017-06-12
Ensure leaveOpen is true when writing to a buffer

### LINKED ISSUE #3209: C#: CodedInputStream.leaveOpen is often false when it should be true
The field docs claim it will always be true unless a stream is provided, but the code doesn't appear to obey that.

jskeet:
Fixes #3209.


jskeet:
Hi @anandolee - sorry I hadn't fixed the failures before. (I thought I had, in exactly the same way - maybe I never pushed...) Ready for review now.

PR #3225, opened 2017-06-14
Switch to addEnumType to fix fatal error

RyanGordon:
This addresses issue https://github.com/google/protobuf/issues/3224


bazel-io:

RyanGordon:
I signed it!

TeBoring:
Could you resolve the conflict?
Recently, we split descriptor.php.
You can add your changes to splitted files.

TeBoring:
ok to test

TeBoring:
Seems not working for php7. See details in travis.

php/tests/generated_class_test.php
```diff
@@ -853,7 +882,9 @@ public function testEnumWithoutNamespace()
     {
         $m = new TestMessage();
         $m->setOptionalNoNamespaceEnum(NoNameSpaceEnum::VALUE_A);
-        $m->getRepeatedNoNamespaceEnum()[] = NoNameSpaceEnum::VALUE_A;
+        $repeatedNoNamespaceEnum = $m->getRepeatedNoNamespaceEnum();
+        $repeatedNoNamespaceEnum[] = NoNameSpaceEnum::VALUE_A;
+        $n->setRepeatedNoNamespaceEnum($repeatedNoNamespaceEnum);
```
TeBoring: $n seems not defined.

RyanGordon: Yep on it :) Thanks!

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -896,4 +895,24 @@ public function byteSize()
...
+    }
+
+    private function kvUpdateHelper($field, $update_key, $update_value)
+    {
+        $getter = $field->getGetter();
+        $setter = $field->getSetter();
+
+        $field_arr_value = $this->$getter();
+        $field_arr_value[$update_key] = $update_value;
+        $this->$setter($field_arr_value);
```
TeBoring: Is this needed?

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -896,4 +895,24 @@ public function byteSize()
...
     }
+
+    private function appendHelper($field, $append_value)
+    {
+        $getter = $field->getGetter();
+        $setter = $field->getSetter();
+
+        $field_arr_value = $this->$getter();
+        $field_arr_value[] = $append_value;
+        $this->$setter($field_arr_value);
```
TeBoring: Is this needed?

php/src/Google/Protobuf/Internal/Message.php
RyanGordon: Yes, on line 914, (internally in PHP) we get a reference to the array, then on line 915 PHP does a CoW and we update the local copy of the array, then on 916 we update the original source of the array with the updated local copy. This is likely what was happening also in the previous syntax; it's just not compatible in HHVM because the behavior with the old syntax is ambiguous with PHP's CoW system.

TeBoring: However, the $field_arr_value returned on line914 is a MapField object instead of array.

RyanGordon: Ah I did not realize that. Will everything that is returned from `$this->$getter()` be a MapField object?

TeBoring: getters of repeated and map fields will return protobuf's custom data structure

PR #3226, opened 2017-06-14
Add json encode/decode for php.

TeBoring:


bazel-io:

jbq:
Hi @TeBoring nice contribution!  I'm a user of the protobuf library, looking for JSON support in Protobuf PHP.  Tell me if I can lend a hand.  Seems like there is a conflict in `php/src/Google/Protobuf/Internal/Message.php`.

TeBoring:
Thanks @jbq for your help. The json support is almost done. Welcome to try it when it's merged.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -68,6 +68,10 @@ public function __construct($desc = NULL)
         // specific descriptor from the descriptor pool.
         if (get_class($this) === 'Google\Protobuf\Internal\MapEntry') {
             $this->desc = $desc;
+            foreach ($desc->getField() as $field) {
+                $setter = $field->getSetter();
+                $this->$setter($this->defaultValue($field));
```
bshaffer: The `MapEntry` field has pass-by-reference setters (I don't know why, this may be a bad idea). Because of that, the default value needs to be set to a variable before being passed in, to avoid throwing a Strict Standards error. Something like this will work:

```php
$setter = $field->getSetter();
$defaultValue = $this->defaultValue($field);
$this->$setter($defaultValue);
```

TeBoring: Thanks for the catch. Do you know how to trigger the Strict Standards error?
For normal messages, a primitive field's setter is passing-by-value while a message field's setter is passing-by-message.
MapEntry is a special message for internal implementation only. Its interface need to be general enough for whether primitive field or message field. However, I think you are right. Change the setter parameter to pass-by-value seems better to me.

php/src/Google/Protobuf/Internal/GPBUtil.php
```diff
@@ -70,15 +75,13 @@ public function divideInt64ToInt32($value, &$high, &$low, $trim = false)
     public static function checkString(&$var, $check_utf8)
     {
         if (is_array($var) || is_object($var)) {
-            trigger_error("Expect string.", E_USER_ERROR);
-            return;
+            throw new \Exception("Expect string.");
```
michaelbausor: These could perhaps be `InvalidArgumentException`? (http://www.php.net/manual/en/class.invalidargumentexception.php) Or possibly a custom exception type like `GPBDecodeException`

php/tests/encode_decode_test.php
```diff
@@ -95,6 +95,20 @@ public function testEncodeDecodeOneof()
         $n = new TestMessage();
         $n->mergeFromString($data);
         $this->assertSame("oneof_enum", $n->getMyOneof());
+
+        $m->setOneofString("");
+        $data = $m->serializeToString();
+        $n = new TestMessage();
+        $n->mergeFromString($data);
+        $this->assertSame("oneof_string", $n->getMyOneof());
```
michaelbausor: Is it possible to also have an assertion to test that the value is equal to `""` here? Or is that unnecessary?

php/tests/php_implementation_test.php
```diff
@@ -206,9 +206,8 @@ public function testZigZagEncodeDecode()
         $this->assertSame(2, GPBWire::zigZagEncode32(1));
         $this->assertSame(3, GPBWire::zigZagEncode32(-2));
         $this->assertSame(0x7FFFFFFE, GPBWire::zigZagEncode32(0x3FFFFFFF));
-        $this->assertSame(0x7FFFFFFF, GPBWire::zigZagEncode32(0xC0000000));
-        $this->assertSame(-2, GPBWire::zigZagEncode32(0x7FFFFFFF));
-        $this->assertSame(-1, GPBWire::zigZagEncode32(0x80000000));
+        # $this->assertSame(0x7FFFFFFF, GPBWire::zigZagEncode32(0xC0000000));
```
michaelbausor: Should this be removed or uncommented?

php/tests/undefined_test.php
```diff
@@ -0,0 +1,920 @@
...
+class UndefinedTest extends PHPUnit_Framework_TestCase
+{
+
+    /**
+     * @expectedException PHPUnit_Framework_Error
+     */
+    public function testInt32AppendStringFail()
+    {
+        $arr = new RepeatedField(GPBType::INT32);
+        $arr []= 'abc';
```
michaelbausor: Nit: remove space between `$arr` and `[]`, add space between `[]` and `=` (And in other places)

bjwatson:
Hi @TeBoring. Thanks for working on this. Does this implementation require C Extension, or does it also include a pure PHP implementation?

TeBoring:
json is supported by both php and c implementation.

bjwatson:
Great! Thanks for getting this done.

php/src/Google/Protobuf/Internal/GPBUtil.php
TeBoring: Prefer `InvalidArgumentException`, since this function may also be used in other situations besides decode.

php/tests/encode_decode_test.php
TeBoring: Done

php/tests/php_implementation_test.php
TeBoring: uncommented

php/tests/undefined_test.php
TeBoring: Fixed

bshaffer:
LGTM!

mcos:
Which release will this be available in?

TeBoring:
ETA is end of July in 3.4 release.

mcos:
Are you planning to push the July release out into August?

TeBoring:
ETA end of week.

PR #3232, opened 2017-06-15
Fix incorrect function call to getFullClassName

michaelbausor:
cc @TeBoring @bshaffer


bazel-io:

TeBoring:
ok to test

TeBoring:
The failing test is due to not updated Dockfile, which is a separate issue.

PR #3233, opened 2017-06-15
Fix php jenkins test

TeBoring:
Update commit id to upload latest composer.
Compile php with bc-math for future json support.


bazel-io:

TeBoring:
The broken test is for javanano.

PR #3246, opened 2017-06-20
Cherry-pick php fix changes from master to 3.3.x

TeBoring:


bazel-io:

TeBoring:
@michaelbausor

xfxyjwf:
Why is this cherry-picked to 3.3.x branch?

TeBoring:
These fixes need to be released soon for other cloud teams.

michaelbausor:
@TeBoring thanks, this looks great!

xfxyjwf:
The change in this PR is more significant than what a patch release is for. A patch release is for bug fixes but I believe you are introducing new features here. I don't think it's a good idea to put things into a patch release just because we need it soon. We should version our releases according to the significance of the change and according to semantic versioning, we should increment the minor version if new backwards-compatible functionality is added, i.e., these php changes should be released in 3.4.0 instead.

TeBoring:
@michaelbausor , are all cherry-picked required? We need to remove new features and only include bug fixes for this branch.

bshaffer:
0469dd5  and a4dcd8b are the two big ones. The first adds the namespace option, and the second fixes a fatal error in autoloading.

bshaffer:
a4dcd8b (autoload issue) is a bug fix. Minor backwards compatibility was affected when we moved functions into static methods, but this could be split from this patch, as it's not critical in fixing the fatal error.

0469dd5 (namespace option) is a feature and not a bug fix, but it is a very important feature for google-cloud-php. If this is not accepted for a patch version, we hope to get it into a feature version very soon.

PR #3258, opened 2017-06-21
Update version number to 3.3.2

TeBoring:


bazel-io:

scpeters:
I see that the [v3.3.2 tag](https://github.com/google/protobuf/releases/tag/v3.3.2) was created after this pull request, but it is not classified as a release on github. So downstream packaging projects like [pypi](https://pypi.python.org/pypi/protobuf/3.3.0) and [homebrew](https://github.com/homebrew/homebrew-core/blob/master/Formula/protobuf.rb#L4) are still using 3.3.0 (see discussion in https://github.com/Homebrew/homebrew-core/pull/15074#issuecomment-311735007 ). I apologize for the intrusion on your time, but if you can help clarify whether the v3.3.2 tag is intended as a proper release, I would appreciate it.

PR #3259, opened 2017-06-21
3.3.x

22abhay:
I am trying to compile protobuf library on HP-UX,IA64 Architecture, using aCC compiler to compile it but not able to get through. Can anyone please let know whether we support protobuf on HP-UX.

Any help or pointer with protobuf support on HPUX will be much appreciated.


bazel-io:

TeBoring:
Could you send us a pull request to fix the error instead?

Issue #3223, opened 2017-06-13
Fatal error C1057 in multiple cc files


Spenhouet:
### System information
- **OS Platform**: `Windows 10`
- **Protobuf version**: 3.2.0
- **TensorFlow version**: `v1.2.0-rc2`
- **Bazel version**: `git/master`
- **Python**: 3.6.1 x64
- **Exact command to reproduce**: `bazel build -c opt --cpu=x64_windows_msvc --host_cpu=x64_windows_msvc --copt=-w --host_copt=-w tensorflow/tools/pip_package:build_pip_package`

### Describe the problem

I try to compile TensorFlow on Windows with Bazel (without GPU support). But i run into a error (see down below).

### Error log
```
ERROR: C:/users/spenh/appdata/local/temp/_bazel_spen/icnq02mb/external/protobuf/BUILD:623:1: C++ compilation of rule '@protobuf//:python/google/protobuf/pyext/_message.so' failed: msvc_cl.bat failed: error executing command external/local_config_cc/wrapper/bin/msvc_cl.bat /DOS_WINDOWS=OS_WINDOWS /DCOMPILER_MSVC /DNOGDI /DNOMINMAX /DPRAGMA_SUPPORTED /D_WIN32_WINNT=0x0600 /D_CRT_SECURE_NO_DEPRECATE /D_CRT_SECURE_NO_WARNINGS ... (remaining 42 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 2.
C:\users\spenh\appdata\local\temp\_bazel_spen\icnq02mb\execroot\org_tensorflow\external\protobuf\python\google\protobuf\pyext\message.cc(1064): error C2121: "#": ungültiges Zeichen: möglicherweise das Ergebnis einer Makroerweiterung
C:\users\spenh\appdata\local\temp\_bazel_spen\icnq02mb\execroot\org_tensorflow\external\protobuf\python\google\protobuf\pyext\message.cc(1064): error C2059: Syntaxfehler: "if"
Target //tensorflow/tools/pip_package:build_pip_package failed to build
```
The error seems to be in line 1064:

```
1056   if (PySlice_Check(slice)) {
1057     from = to = step = slice_length = 0;
1058     PySlice_GetIndicesEx(
1059 #if PY_MAJOR_VERSION < 3
1060         reinterpret_cast<PySliceObject*>(slice),
1061 #else
1062         slice,
1063 #endif
[...]

PR #3262, opened 2017-06-22
Make it compatible with python 3.6.1

snnn:
#3223 

https://bugs.python.org/issue29943

### DIFF
--- a/python/google/protobuf/pyext/message.cc
+++ b/python/google/protobuf/pyext/message.cc
@@ -1055,13 +1055,15 @@ int InternalDeleteRepeatedField(
 
   if (PySlice_Check(slice)) {
     from = to = step = slice_length = 0;
-    PySlice_GetIndicesEx(
 #if PY_MAJOR_VERSION < 3
+    PySlice_GetIndicesEx(
         reinterpret_cast<PySliceObject*>(slice),
-#else
-        slice,
-#endif
         length, &from, &to, &step, &slice_length);
+#else
+    PySlice_GetIndicesEx(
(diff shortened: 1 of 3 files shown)


bazel-io:

snnn:
PySlice_GetIndicesEx is a function in py2 but a macro in py3.

snnn:
Seems that this is specific to python 3.6

python/google/protobuf/pyext/message.cc
```diff
@@ -1055,13 +1055,15 @@ int InternalDeleteRepeatedField(
 
   if (PySlice_Check(slice)) {
     from = to = step = slice_length = 0;
-    PySlice_GetIndicesEx(
 #if PY_MAJOR_VERSION < 3
+    PySlice_GetIndicesEx(
```
anandolee: Looks like the same with original file?

snnn: Yes. They should be the same, except the #if/#else/#endif are moved out.


#if xxxx
foo(XXX)
#else
foo(YYY)
#endif

Instead of
foo(
#if xxxx
XXX
#else
YYY
#endif
)

anandolee: So what's the behavior difference for python 3.6.1?

snnn: In some python releases, PySlice_GetIndicesEx is a macro.

anandolee: Thanks for the fix and explain

PR #3264, opened 2017-06-22
Enusre public header and generated code have no implicit converion.

TeBoring:


bazel-io:

PR #3269, opened 2017-06-25
Support PHP generic services

### LINKED ISSUE #2334: Compiler-generated PHP code is missing service definition
When compiling a simple helloworld.proto with a service definition a request and a response, the generated PHP code contains just two classes for the request and response respectively. There is no class for the service.

[helloworld.proto.txt](https://github.com/google/protobuf/files/572793/helloworld.proto.txt)
[helloworld.pb.php.txt](https://github.com/google/protobuf/files/572795/helloworld.pb.php.txt)

yuloh:
This PR adds support for generating generic services when using PHP output.

The first commit adds a file level `php_generic_services` option (default false).  I ran `generate_descriptor_proto.sh` so the commit has a lot of changes.

The second commit updates `php_generator.cc` to generate a service interface if `php_generic_services` is true.  The service generated for the grpc helloworld.proto example [looks like this](https://gist.github.com/yuloh/a28d9ff223887aef11aba40c098b71d9).

I tried my best to follow the code style but I'm not a C++ developer so let me know if I need to change anything.

Since we can't use grpc to build PHP servers yet we are looking to use the service interface with our own simple RPC system.  Once we can switch to grpc we are planning on writing an adapter (so grpc doesn't need to implement the interface directly).

There is a comment in `descriptor.proto` that says 'Generic services are now considered deprecated in favor of using plugins', but the proto3 docs say 'the protocol buffer compiler will generate service interface code and stubs in your chosen language.'  Should we just be using a plugin instead?

Closes #2334.

### DIFF
--- a/php/src/Google/Protobuf/Internal/FileOptions.php
+++ b/php/src/Google/Protobuf/Internal/FileOptions.php
@@ -110,6 +110,11 @@ class FileOptions extends \Google\Protobuf\Internal\Message
     private $py_generic_services = false;
     private $has_py_generic_services = false;
     /**
+     * Generated from protobuf field <code>optional bool php_generic_services = 19 [default = false];</code>
+     */
+    private $php_generic_services = false;
+    private $has_php_generic_services = false;
+    /**
      * Is this file deprecated?
      * Depending on the target platform, this can emit Deprecated annotations
      * for everything in the file, or it will be completely ignored; in the very
@@ -546,6 +551,34 @@ class FileOptions extends \Google\Protobuf\Internal\Message
     public function hasPyGenericServices()
     {
         return $this->has_py_generic_services;
+    }
+
+    /**
+     * Generated from protobuf field <code>optional bool php_generic_services = 19 [default = false];</code>
+     * @return bool
+     */
+    public function getPhpGenericServices()
+    {
+        return $this->php_generic_services;
+    }
+
+    /**
+     * Generated from protobuf field <code>optional bool php_generic_services = 19 [default = false];</code>
+     * @param bool $var
+     * @return $this
+     */
+    public function setPhpGenericServices($var)
+    {
+        GPBUtil::checkBool($var);
+        $this->php_generic_services = $var;
+        $this->has_php_generic_services = true;
+
+        return $this;
+    }
+
+    public function hasPhpGenericServices()
+    {
+        return $this->has_php_generic_services;
     }
 
     /**
--- a/php/tests/proto/test_service.proto
+++ b/php/tests/proto/test_service.proto
@@ -1,0 +1,18 @@
+syntax = "proto3";
+
+package foo;
+
+option php_generic_services = true;
+
+service Greeter {
+  rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc SayHelloAgain (HelloRequest) returns (HelloReply) {}
+}
+
+message HelloRequest {
+  string name = 1;
+}
+
+message HelloReply {
+  string message = 1;
+}
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -84,7 +84,7 @@ void GenerateEnumValueDocComment(io::Printer* printer,
                                  const EnumValueDescriptor* value);
 void GenerateServiceDocComment(io::Printer* printer,
                                const ServiceDescriptor* service);
-void GenerateMethodDocComment(io::Printer* printer,
+void GenerateServiceMethodDocComment(io::Printer* printer,
                               const MethodDescriptor* method);
 
 std::string RenameEmpty(const std::string& name) {
@@ -1143,7 +1143,7 @@ void GenerateServiceFile(const FileDescriptor* file,
 
   for (int i = 0; i < service->method_count(); i++) {
     const MethodDescriptor* method = service->method(i);
-    GenerateMethodDocComment(&printer, method);
+    GenerateServiceMethodDocComment(&printer, method);
     GenerateServiceMethod(method, &printer);
   }
 
@@ -1332,7 +1332,7 @@ void GenerateEnumValueDocComment(io::Printer* printer,
     "def", EscapePhpdoc(FirstLineOf(value->DebugString())));
 }
 
-void GenerateMethodDocComment(io::Printer* printer,
+void GenerateServiceMethodDocComment(io::Printer* printer,
                               const MethodDescriptor* method) {
   printer->Print("/**\n");
   GenerateDocCommentBody(printer, method);
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -288,7 +288,7 @@ std::string GeneratedEnumFileName(const EnumDescriptor* en,
 
 std::string GeneratedServiceFileName(const ServiceDescriptor* service,
                                     bool is_descriptor) {
-  std::string result = FullClassName(service, is_descriptor);
+  std::string result = FullClassName(service, is_descriptor) + "Interface";
   for (int i = 0; i < result.size(); i++) {
     if (result[i] == '\\') {
       result[i] = '/';
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -82,6 +82,10 @@ void GenerateEnumDocComment(io::Printer* printer, const EnumDescriptor* enum_,
                             int is_descriptor);
 void GenerateEnumValueDocComment(io::Printer* printer,
                                  const EnumValueDescriptor* value);
+void GenerateServiceDocComment(io::Printer* printer,
+                               const ServiceDescriptor* service);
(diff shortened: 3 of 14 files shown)


bazel-io:

yuloh:
RE: the CLA - I signed it!

TeBoring:
Thanks for the change! We do plan to let user specify whether to generated service in proto file.
Could you resolve conflicts in the change?

yuloh:
> We do plan to let user specify whether to generated service in proto file.

Awesome!  I just rebased and re-ran `generate_descriptor_proto.sh` to resolve the conflict.

TeBoring:
ok to test

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -267,6 +271,18 @@ std::string GeneratedEnumFileName(const EnumDescriptor* en,
   return result + ".php";
 }
 
+std::string GeneratedServiceFileName(const ServiceDescriptor* service,
```
TeBoring: Please reference GeneratedMessageFileName to implement this function.
Recently, we add php_namespace to specify custom namespace instead of using package.
We also have some logic to escape reserved name in php.

yuloh: I just pushed a commit to use `FullClassName` to generate the service file name.  I had to overload the method because `ServiceDescriptor` does not have a `containing_type` method.  I also extracted a `NamespacedName` function to allow re-using the logic.

yuloh:
The build failure looks unrelated to these changes.  Hopefully rebuilding will fix it.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project compatibility-more-protos: Execution default-compile of goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile failed: Plugin org.apache.maven.plugins:maven-compiler-plugin:3.3 or one of its dependencies could not be resolved: Could not transfer artifact log4j:log4j:jar:1.2.12 from/to central (https://repo.maven.apache.org/maven2): GET request of: log4j/log4j/1.2.12/log4j-1.2.12.jar from central failed: Connection reset -> [Help 1]
```

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1111,6 +1204,16 @@ void GenerateMessageDocComment(io::Printer* printer,
...
 }
 
+void GenerateServiceDocComment(io::Printer* printer,
+                               const ServiceDescriptor* service) {
+  printer->Print("/**\n");
+  GenerateDocCommentBody(printer, service);
+  printer->Print(
+    " * Protobuf type <code>^fullname^</code>\n"
+    " */\n",
+    "fullname", EscapePhpdoc(service->full_name()));
```
bshaffer: This will need to be added to https://github.com/google/protobuf/pull/3131, or vice-versa depending on which is merged first

TeBoring: I can send a fixing PR after this is merged.

yuloh:
I rebased and resolved the conflicts.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -82,6 +82,10 @@ void GenerateEnumDocComment(io::Printer* printer, const EnumDescriptor* enum_,
                             int is_descriptor);
 void GenerateEnumValueDocComment(io::Printer* printer,
                                  const EnumValueDescriptor* value);
+void GenerateServiceDocComment(io::Printer* printer,
+                               const ServiceDescriptor* service);
+void GenerateMethodDocComment(io::Printer* printer,
```
bshaffer: This should be `GenerateServiceMethodDocComment`, as just `method` is too ambiguous, as `GenerateFieldDocComment` also documents a method.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -663,6 +688,16 @@ void GenerateEnumToPool(const EnumDescriptor* en, io::Printer* printer) {
   Outdent(printer);
 }
 
+void GenerateServiceMethod(const MethodDescriptor* method,
+                           io::Printer* printer) {
+  printer->Print(
+        "public function ^camel_name^(^request_name^ $request);\n\n",
+        "camel_name", UnderscoresToCamelCase(method->name(), false),
+        "request_name", UnderscoresToCamelCase(
+          method->input_type()->name(), true)
```
bshaffer: Will this always generate the correct typehint? Will the `Request` class always be in the same namespace as the service method?

php/src/Google/Protobuf/Internal/FileOptions.php
```diff
@@ -549,6 +554,34 @@ public function hasPyGenericServices()
...
+        GPBUtil::checkBool($var);
+        $this->php_generic_services = $var;
+        $this->has_php_generic_services = true;
+
+        return $this;
+    }
+
+    public function hasPhpGenericServices()
+    {
+        return $this->has_php_generic_services;
```
bshaffer: Can you provide an example for when/how these functions are used?

yuloh: These are generated by `generate_descriptor_proto.sh` which has to be run after editing descriptor.proto or the code won't compile.

src/google/protobuf/compiler/php/php_generator.cc
yuloh: Good point.  I was able to break it by defining the request in a separate proto, importing it, and setting different namespaces on each.

php/src/Google/Protobuf/Internal/FileOptions.php
TeBoring: The generated code for descriptor.proto is currently for implementation usage only.
descriptor.proto is a proto2 file instead of proto3. That's why it has the hasX method.

yuloh:
Is there any documentation on getting the php tests to run?  Seems like you can't run the tests with PHP 7.1 or PHPUnit 6+ and you need valgrind installed?

TeBoring:
Are you testing locally? The fix for php7.1 hasn't been merged from 3.3.x branch.
You can check https://github.com/google/protobuf/blob/master/tests.sh and search for php5.5 and php5.5_c to see how to run the test.

bshaffer:
I test by doing the following:

```
# Build the changes in php_generator.cc
pushd src
make
popd

# Build the core PHP files
bash php/generate_descriptor_protos.sh

# Build the test proto files
pushd php/tests
rm -rf generated
mkdir generated
../../src/protoc --php_out=generated proto/test.proto proto/test_include.proto proto/test_no_namespace.proto proto/test_prefix.proto proto/test_php_namespace.proto proto/test_empty_php_namespace.proto
pushd ../../src
./protoc --php_out=../php/tests/generated google/protobuf/empty.proto
./protoc --php_out=../php/tests/generated -I../php/tests -I. ../php/tests/proto/test_import_descriptor_proto.proto
popd
popd

# Run PHPUnit tests
pushd php
phpunit
```

TeBoring:
@yuloh are you fixing any comment?

yuloh:
Hello,

I got the tests working.  I will hopefully have some free time soon to push some tests and address @bshaffer's points.

TeBoring:
@yuloh any update?

yuloh:
Hi,

I pushed tests for the generated class, changed the method to always use the fully qualified class name (in case the message is in a different namespace), fixed an error in the doc comment for the service method, and renamed `GenerateMethodDocComment` like @bshaffer suggested.  Sorry it took so long!

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1064,6 +1099,58 @@ void GenerateMessageFile(const FileDescriptor* file, const Descriptor* message,
...
+    printer.Print(
+        "namespace ^name^;\n\n",
+        "name", fullname.substr(0, lastindex));
+  }
+
+  GenerateServiceDocComment(&printer, service);
+
+  if (lastindex != string::npos) {
+      printer.Print(
+        "interface ^name^\n"
```
bshaffer: Idiomatically, interfaces are suffixed with `Interface`, i.e.  `FooInterface` (e.g. [PSR naming conventions](http://www.php-fig.org/bylaws/psr-naming-conventions/)).

php/tests/proto/test_service.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package foo;
+
+option php_generic_services = true;
```
bshaffer: Is `generic service` part of protobuf terminology? This term is not used in PHP, so it would make more sense from a PHP perspective to make this option `php_interfaces`

src/google/protobuf/compiler/php/php_generator.cc
yuloh: I don't use the suffix personally so I didn't think about it 😆 .  The suffix is probably more common though, I will push a commit to use it 👍 .

TeBoring:
php_generic_service is proto terminology. Csharp, objc have the same term.

yuloh:
The appveyor failure seems unrelated.  It's failing for the other PRs too.

TeBoring:
Thanks for the work!

Issue #2684, opened 2017-02-07
Support for setters and getters insertion points


dopuskh3:
Adding insertion points in various field modifier could help extending generated classes with additional logic in those.

I think about implementing nullable support by using these insertion points the following way:

I will create a protobuf plugin that does the following:
- when a specific option is present on a message (say 'allowNullableFields') then the plugin looks for a map[int]->bool map.
- If the map exists then insert code that set value to true for the corresponding field ID in the map in various setters scopes.
- If the map does not exists then the protobuf plugin fail compiling the corresponding schema.
- Also the plugin will generate a hasXXX method for each of the field that will simply look for the field id in the map an return true if the corresponding value is set to true.

Sounds like a less intrusive approach to solve issue #1606 but still requires additional insertion points.

iamneal:
@dopuskh3 What is the status of this plugin?  Will it be available for all proto3 supported languages?

xfxyjwf:
Note that we are beginning to limit and reduce the number of insertion points.

PR #3280, opened 2017-06-28
Add insertion points for field modifiers in C# and Java

gturri:

Said differently, this is how we solve #1606 for Java and C#


bazel-io:

gturri:
CLA should be fixed now

gturri:
new test for CLA

xfxyjwf:
Can you split this into two PRs, one with C# change and the other with Java change? Also please direct them to the master branch as only bug fixes should go into patch releases.

gturri:
Sure!
I just created #3291 and #3292 accordingly.

PR #3281, opened 2017-06-28
Proto2 test message support to conformance test

BSBandme:
Add changes to conformance test that support proto2 test message. Only add one proto2 message test against double. Not sure about whether the message set message in TestAllTypesProto2 is correct.


bazel-io:

conformance/conformance_python.py
```diff
@@ -56,14 +57,26 @@ def do_test(request):
...
   try:
     if request.WhichOneof('payload') == 'protobuf_payload':
-      try:
-        test_message.ParseFromString(request.protobuf_payload)
-      except message.DecodeError as e:
-        response.parse_error = str(e)
-        return response
+      if isProto3:
+        try:
+          test_message.ParseFromString(request.protobuf_payload)
```
haberman: It should be possible to share most of the code between proto2 and proto3. Something like:

```
  test_message = <choose based on proto2/proto3>
  if request.WhichOneof('payload') == 'protobuf_payload':
    test_message.ParseFromString(request.protobuf_payload)
```

conformance/conformance_test.cc
```diff
@@ -299,6 +314,7 @@ void ConformanceTestSuite::RunValidInputTest(
   RunTest(test_name, request, &response);
 
   TestAllTypes test_message;
+  TestAllTypesProto2 test_message_proto2;
```
haberman: Too much code duplication. There should be a `Message*` that points to the correct message. Then message->ParseFromString() will do the right thing.

conformance/conformance_test.cc
```diff
@@ -756,23 +811,29 @@ bool ConformanceTestSuite::RunSuite(ConformanceTestRunner* runner,
     {dbl(0.1), "0.1"},
     {dbl(1.7976931348623157e+308), "1.7976931348623157e+308"},
     {dbl(2.22507385850720138309e-308), "2.22507385850720138309e-308"}
-  });
+  }, true);
+  TestValidDataForType(FieldDescriptor::TYPE_DOUBLE, {
```
haberman: It seems that TestValidDataForType() could test both proto2 and proto3 internally. Then you wouldn't need to change any of these call sites.

src/google/protobuf/test_messages_proto2.proto
```diff
@@ -0,0 +1,216 @@
...
+  optional int32 FIELD_name12 = 412;
+  optional int32 __field_name13 = 413;
+  optional int32 __Field_name14 = 414;
+  optional int32 field__name15 = 415;
+  optional int32 field__Name16 = 416;
+  optional int32 field_name17__ = 417;
+  optional int32 Field_name18__ = 418;
+
+  // message_set test case.
+  message mset_correct {
```
haberman: Please follow protobuf naming conventions, messages should be like `MessageSetCorrect`.

haberman:
ok to test

TeBoring:
What's the plan to enforce proto2 conformance test? php doesn't support proto2 in near future.

conformance/ConformanceJava.java
```diff
@@ -219,16 +304,96 @@ private void writeLittleEndianIntToStdout(int val) throws Exception {
 
     return messages[0];
   }
+  
+  private TestMessagesProto2.TestAllTypesProto2 parseBinaryToProto2(ByteString bytes)
+      throws InvalidProtocolBufferException {
```
BSBandme: Done

conformance/conformance.proto
```diff
@@ -77,6 +77,9 @@ message ConformanceRequest {
 
   // Which format should the testee serialize its message to?
   WireFormat requested_output_format = 3;
+
+  // should be set to either "proto2" or "proto3"
+  string message_type = 4;
```
BSBandme: Done

conformance/conformance_python.py
BSBandme: Done

conformance/conformance_test.cc
BSBandme: Done

conformance/conformance_test.cc
BSBandme: Done

src/google/protobuf/test_messages_proto2.proto
BSBandme: Done

BSBandme:
We're asked to add some conformance tests about messageSet and enum, etc that proto3 doesn't have currently. I've add some codes to php code to skip proto2 testmessage.

conformance/Makefile.am
```diff
@@ -242,8 +250,9 @@ google-protobuf:
 if USE_EXTERNAL_PROTOC
 
 # Some implementations include pre-generated versions of well-known types.
-protoc_middleman: $(conformance_protoc_inputs) $(well_known_type_protoc_inputs) google-protobuf
+protoc_middleman: $(conformance_protoc_inputs) $(conformance_proto2_protoc_inputs) $(well_known_type_protoc_inputs) google-protobuf
 	$(PROTOC) -I$(srcdir) -I$(top_srcdir) --cpp_out=. --java_out=. --ruby_out=. --objc_out=. --python_out=. --php_out=. --js_out=import_style=commonjs,binary:. $(conformance_protoc_inputs)
+	$(PROTOC) -I$(srcdir) -I$(top_srcdir) --cpp_out=. --java_out=. --python_out=. $(conformance_proto2_protoc_inputs)
```
haberman: Can't you add these inputs to the same invocation of protoc? It doesn't seem like we should need to call protoc more times.

conformance/conformance_cpp.cc
```diff
@@ -87,17 +91,30 @@ void CheckedWrite(int fd, const void *buf, size_t len) {
 }
 
 void DoTest(const ConformanceRequest& request, ConformanceResponse* response) {
-  TestAllTypes test_message;
+  Message *test_message;
```
haberman:
```
descriptor = DescriptorPool::generated_pool()->FindMessageTypeByName(
    request.message_type());
if (!descriptor) {
  GOOGLE_LOG(FATAL) << "No such message type: " << request.message_type();
}
test_message = MessageFactory::generated_factory()->GetPrototype(descriptor)->New();
```

conformance/conformance_nodejs.js
```diff
@@ -49,14 +49,22 @@ function doTest(request) {
...
+        if (request.getMessageType() == "protobuf_test_messages.proto3.TestAllTypes") {
+          try {
+            testMessage = test_messages_proto3.TestAllTypes.deserializeBinary(
+                request.getProtobufPayload());
+          } catch (err) {
+            response.setParseError(err.toString());
+            return response;
+          }
+        } else if (request.getMessageType() == "protobuf_test_messages.proto2.TestAllTypesProto2"){
+          response.setSkipped("NodeJS doesn\'t support proto2");
```
haberman: It does support proto2 (or at least it should). It supports proto2 internally.

conformance/conformance_python.py
```diff
@@ -53,18 +54,25 @@ class ProtocolError(Exception):
   pass
 
 def do_test(request):
-  test_message = test_messages_proto3_pb2.TestAllTypes()
+  isProto3 = (request.message_type == "protobuf_test_messages.proto3.TestAllTypes")
```
haberman: You should be able to look this up by name also I think.

conformance/conformance_test.cc
```diff
@@ -571,43 +606,43 @@ void ConformanceTestSuite::TestPrematureEOFForType(FieldDescriptor::Type type) {
 
   ExpectParseFailureForProto(
       tag(field->number(), wire_type),
-      "PrematureEofBeforeKnownNonRepeatedValue" + type_name, REQUIRED);
+      "PrematureEofBeforeKnownNonRepeatedValue" + type_name, REQUIRED, true);
```
haberman: ExpectParseFailureForProto() should not take a boolean. It should test against both proto2 and proto3.

conformance/conformance_test.cc
```diff
@@ -627,46 +662,49 @@ void ConformanceTestSuite::TestPrematureEOFForType(FieldDescriptor::Type type) {
...
-                       field->name() + ": " + values[0].second);
-
-  string proto;
-  string text = field->name() + ": " + values.back().second;
-  for (size_t i = 0; i < values.size(); i++) {
-    proto += cat(tag(field->number(), wire_type), values[i].first);
-  }
-  RunValidProtobufTest("RepeatedScalarSelectsLast" + type_name, REQUIRED,
-                       proto, text);
+  for (int isProto3 = 0; isProto3 < 2; isProto3++) {
```
haberman: Please factor this into a function and write:

```
TestValidDataForTypeWithProtoVersion(type, values, PROTO2);
TestValidDataForTypeWithProtoVersion(type, values, PROTO3);
```

conformance/conformance_test.cc
```diff
@@ -1337,14 +1376,14 @@ bool ConformanceTestSuite::RunSuite(ConformanceTestRunner* runner,
         WireFormatLite::DecodeFloat(0x7FA12345));
     RunValidJsonTestWithProtobufInput(
         "FloatFieldNormalizeQuietNan", REQUIRED, message,
-        "optional_float: nan");
+        "optional_float: nan", true);
```
haberman: JSON does not exist for proto2, so there is no need to pass a bool indicating that this is for proto3. Anything JSON-related we can assume is for proto3.

src/google/protobuf/test_messages_proto2.proto
```diff
@@ -0,0 +1,216 @@
...
+
+enum ForeignEnum {
+  FOREIGN_FOO = 0;
+  FOREIGN_BAR = 1;
+  FOREIGN_BAZ = 2;
+}
+
+extend TestAllTypesProto2 {
+  optional int32 extension_int32 = 120;
+}
```
haberman: Please ensure this is a normal file with a newline at the end.

conformance/conformance_python.py
BSBandme: It seems that only python_cpp can use [generated descriptor_pool ](https://github.com/google/protobuf/blob/master/python/google/protobuf/descriptor_pool.py#L990).

conformance/Makefile.am
BSBandme: How can I merge this into one? I tried protoc with proto2 file as input with flag --ruby_out before, but it raised an error.

src/google/protobuf/test_messages_proto2.proto
BSBandme: Done

conformance/conformance_test.cc
BSBandme: Done

conformance/conformance_test.cc
BSBandme: Done

conformance/conformance_test.cc
BSBandme: Done

conformance/conformance_nodejs.js
BSBandme: Done

conformance/conformance_cpp.cc
BSBandme: Done

BSBandme:
Add one test that first set one field in an unset message, then set the field again. It seems that php will fail on this test

BSBandme:
It seems that the new added test only fails locally for php

conformance/Makefile.am
```diff
@@ -2,7 +2,11 @@
 
 conformance_protoc_inputs =                                    \
   conformance.proto                                            \
-  $(top_srcdir)/src/google/protobuf/test_messages_proto3.proto
+  $(top_srcdir)/src/google/protobuf/test_messages_proto3.proto 
+  
+conformance_proto2_protoc_inputs =                             \
```
thomasvl: why do you need a second set of input and to invoke ptotoc twice in all the generation cases?  You should be able to generate proto2 and proto3 syntax files all at once.  Since you're compiling them into a single binary, I'm assuming they have unique proto package/name combinations, and that isn't the problem.

BSBandme: I tried to generate the proto file in one command, but it seems that it can't use proto2 file as input with "--ruby_out --objc_out --php_out" flag. There will be an error : 
"--ruby_out: google/protobuf/test_messages_proto2.proto: Can only generate Ruby code for proto3 .proto files."

thomasvl: Ah, a per generator limit.  Likely makes sense to add a comment here then to explain it to save someone else from trying it and seeing it fail.

BSBandme: Done by added some comments before definition of conformance_proto2_protoc_inputs

conformance/Makefile.am
```diff
@@ -2,7 +2,12 @@
 
 conformance_protoc_inputs =                                    \
   conformance.proto                                            \
-  $(top_srcdir)/src/google/protobuf/test_messages_proto3.proto
+  $(top_srcdir)/src/google/protobuf/test_messages_proto3.proto 
+  
+# proto2 input files, should be separated with proto3, as we 
+# can't generate proto2 files for ruby, php and objc
```
thomasvl: The objc part isn't true.  We support proto2.  I can back fill the conformance support after this lands.

conformance/conformance.proto
thomasvl: Looks like the comment on this needs to be updated

conformance/Makefile.am
BSBandme: Thx!

PR #3288, opened 2017-06-29
Add build targets for Java lite runtime.

xfxyjwf:
Fixes https://github.com/google/protobuf/issues/2762


xfxyjwf:
@cgrushko

BUILD
```diff
@@ -539,13 +539,103 @@ internal_gen_well_known_protos_java(
     srcs = WELL_KNOWN_PROTOS,
 )
 
+JAVA_LITE_FILES = [
```
cgrushko: Why not inline this into the rule itself?
There's an internal doc about why this is harmful, I can send you a link.

xfxyjwf: Done.

BUILD
```diff
@@ -540,12 +540,100 @@ internal_gen_well_known_protos_java(
...
+        "java/core/src/main/java/com/google/protobuf/Service.java",
+        "java/core/src/main/java/com/google/protobuf/ServiceException.java",
+        "java/core/src/main/java/com/google/protobuf/SingleFieldBuilder.java",
+        "java/core/src/main/java/com/google/protobuf/SingleFieldBuilderV3.java",
+        "java/core/src/main/java/com/google/protobuf/TextFormat.java",
+        "java/core/src/main/java/com/google/protobuf/TextFormatParseInfoTree.java",
+        "java/core/src/main/java/com/google/protobuf/TextFormatParseLocation.java",
+        "java/core/src/main/java/com/google/protobuf/UnknownFieldSet.java",
+        "java/core/src/main/java/com/google/protobuf/UnsafeByteOperations.java",
+    ] + [
```
cgrushko: Delete "] + [" ?

BUILD
```diff
@@ -540,12 +540,100 @@ internal_gen_well_known_protos_java(
 )
 
 java_library(
+    name = "protobuf_java_lite",
```
cgrushko: if the entire code is in `java/core/src/main/java/com/google/protobuf/`, perhaps put a BUILD file there?

BUILD
```diff
@@ -540,12 +540,100 @@ internal_gen_well_known_protos_java(
...
+cc_binary(
+    name = "protoc_gen_javalite",
+    srcs = ["src/google/protobuf/compiler/java/java_lite_main.cc"],
+    linkopts = LINK_OPTS,
+    visibility = ["//visibility:public"],
+    deps = [":protoc_lib"],
+)
+
+java_library(
     name = "protobuf_java",
```
cgrushko: Why is this needed, btw? if we're in the branch that only does javalite.

BUILD
xfxyjwf: Done.

BUILD
xfxyjwf: This branch is forked from master branch, so there are all sorts of non-lite stuff here (e.g., all the full Java runtime source code + all other languages are still in this javalite branch). I would rather keep these as-is.

BUILD
xfxyjwf: "java/core/src/main/java/com/google/protobuf" is a bad build package containing completely unrelated stuff (the path is specially crafted for Maven). Putting the target in the root BUILD file is much more cleaner.

xfxyjwf:
No, I don't think cl.bazel.io covers javalite. We are setting up kokoro tests to replace travis/jenkins and bazel tests can be setup as part of it.

PR #3289, opened 2017-06-29
Add JsonFormat option to use unrecognized enum value when enum value is unknown

veerenmandalia:
Instead of throwing a parse exception when a unknown enum value is seen, use the protobuf UNRECOGNIZED enum value.

This behavior can be enabled via `Parser.usingUnrecognizedEnumValue()`, similar to what `Parser.ignoringUnknownFields()` did for unknown fields.
This allows different parties to add new enum values without all parties needing to know about the value.

[See Issue 3012](https://github.com/google/protobuf/issues/3012)

Signed-off-by: Veeren Mandalia <veeren@postmates.com>

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -285,7 +285,8 @@ public class JsonFormat {
    * Creates a {@link Parser} with default configuration.
    */
   public static Parser parser() {
-    return new Parser(TypeRegistry.getEmptyTypeRegistry(), false, Parser.DEFAULT_RECURSION_LIMIT);
+    return new Parser(TypeRegistry.getEmptyTypeRegistry(), false, false,
+            Parser.DEFAULT_RECURSION_LIMIT);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


bazel-io:

veerenmandalia:
I signed it!

xfxyjwf:
Thanks for the patch! We protobuf-team had a discussion about this and it was decided that instead of adding a new option, the existing ignoringUnknownFields() option should be able to make the Parser ignore unknown enum values (i.e., it's a bug that it doesn't before). Can you update this pull request accordingly?

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -1706,6 +1724,10 @@ private EnumValueDescriptor parseEnum(EnumDescriptor enumDescriptor, JsonElement
           // Fall through. This exception is about invalid int32 value we get from parseInt32() but
           // that's not the exception we want the user to see. Since result == null, we will throw
           // an exception later.
+
+          if (usingUnrecognizedEnumValue) {
+            result = enumDescriptor.createUnrecognizedValueDescriptor();
```
xfxyjwf: The UNRECOGNIZED value should not be used here. Only generated getters should ever return UNRECOGNIZED and the value should not be stored anywhere. Can you treat it as if the field is not present?

veerenmandalia: Sure. Will remove this, client using the default value works for me.

rdsubhas: Hi @veerenmandalia you had a chance to do this, or I'll be glad to pick it up...

vijay-bhatt: I needed the same thing so have created a PR https://github.com/google/protobuf/pull/4825/commits/c26cf191f62e617942f28111fa813f67535c6feb

Please review.

PR #3291, opened 2017-06-30
Add C# insertion points for field modifiers

gturri:

Said differently, this is how we solve #1606 for C#

### DIFF
--- a/csharp/src/AddressBook/Addressbook.cs
+++ b/csharp/src/AddressBook/Addressbook.cs
@@ -88,6 +88,7 @@ namespace Google.Protobuf.Examples.AddressBook {
       get { return name_; }
       set {
         name_ = pb::ProtoPreconditions.CheckNotNull(value, "value");
+        //@@protoc_insertion_point(field_modifier_scope_after:tutorial.Person.setName)
       }
     }
 
@@ -102,6 +103,7 @@ namespace Google.Protobuf.Examples.AddressBook {
       get { return id_; }
       set {
         id_ = value;
+        //@@protoc_insertion_point(field_modifier_scope_after:tutorial.Person.setId)
       }
     }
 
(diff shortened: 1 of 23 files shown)


bazel-io:

jskeet:
I definitely, definitely don't want to do this by default. I'm not really fond of the approach in the first place, to be honest - it feels like it's trying to turn protos (raw DTOs) into something with more customizable behavior - which means that it's going to be harder for users to predict what any particular proto class does.

If other proto owners are happy with it going into other languages, I'd be willing to make it an *optional* feature, but I want to keep the generated code at least *reasonably* idiomatic.

wackoisgod:
@jskeet I will admit the inflexibility of the c# codegen can be rather frustrating. 

One example that I am currently looking at is I want to know when fields are set, thus allowing me to track these changes for an event sourcing mechanism. Right now without the insertion points there is no clean way for me to do this.

I would support adding this as some sort of optional flag to turn these on. @gturri I would gladly help to contribute if you do not have the time.

jskeet:
I certainly don't have the time, and I'm not sure whether the protobuf team would have the time to review this. Yes, the C# codegen is relatively simple - but that also means it's maintainable. Given the number of proposals there are around adding async support, span support, all kinds of things, I fear we'll end up in a situation where everyone's added their "one little feature" - and the code is then a nightmare to maintain.

wackoisgod:
@jskeet I 100% agree that this should be kept simple and easy to maintain, I would say what @gturri is proposing is not that insane and goes a long way in allowing people more flexibility in the codegen and is in sync with some of the other languages such as C++/Python/Obj-C. As of right now C# doesn't support insertion points at all. 

I will go ahead and fork and re-do the work @gturri did with the ability to turn on and off the points, with them defaulting to off. 

I would also like to look at resolving #3487 which I think we be a big win!

jskeet:
Well you can certainly fork - but please be aware that the protobuf team may not accept the resulting pull request. (They may do, of course - you just need to be aware of the possibility that they won't.)

wackoisgod:
Got it, I hope they take in the changes. I don’t see that these changes are anything crazy.

xfxyjwf:
Note that we are moving away from insertion points in other languages as well because they are problematic for the build system (in particular bazel can't handle protoc plugins using insertion points).

PR #3292, opened 2017-06-30
Add Java insertion points for field modifiers

gturri:

Said differently, this is how we solve #1606 for Java

### DIFF
--- a/src/google/protobuf/compiler/java/java_map_field.cc
+++ b/src/google/protobuf/compiler/java/java_map_field.cc
@@ -84,6 +84,7 @@ void SetMessageVariables(const FieldDescriptor* descriptor,
   SetCommonFieldVariables(descriptor, info, variables);
   ClassNameResolver* name_resolver = context->GetNameResolver();
 
+  (*variables)["full_name"] = descriptor->full_name();
   (*variables)["type"] =
       name_resolver->GetImmutableClassName(descriptor->message_type());
   const FieldDescriptor* key = KeyField(descriptor);
@@ -369,6 +370,7 @@ GenerateBuilderMembers(io::Printer* printer) const {
       "public Builder ${$clear$capitalized_name$$}$() {\n"
       "  internalGetMutable$capitalized_name$().getMutableMap()\n"
       "      .clear();\n"
+      "// @@protoc_insertion_point(builder_field_clear:$full_name$)\n"
       "  return this;\n"
       "}\n");
   printer->Annotate("{", "}", descriptor_);
@@ -381,6 +383,7 @@ GenerateBuilderMembers(io::Printer* printer) const {
       "  $key_null_check$\n"
       "  internalGetMutable$capitalized_name$().getMutableMap()\n"
       "      .remove(key);\n"
+      "// @@protoc_insertion_point(builder_field_remove:$full_name$)\n"
       "  return this;\n"
       "}\n");
   printer->Annotate("{", "}", descriptor_);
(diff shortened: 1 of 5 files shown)


bazel-io:

src/google/protobuf/compiler/java/java_map_field.cc
```diff
@@ -354,6 +356,7 @@ GenerateBuilderMembers(io::Printer* printer) const {
                  "public Builder clear$capitalized_name$() {\n"
                  "  internalGetMutable$capitalized_name$().getMutableMap()\n"
                  "      .clear();\n"
+                 "// @@protoc_insertion_point(builder_field_modifier_scope_after:$containing_type_full_name$.clear$capitalized_name$)\n"
```
xfxyjwf: How about a simpler name:
@@protoc_insertion_point(builder_field_clear:$full_name$)
?
where full_name is descriptor->full_name().

src/google/protobuf/compiler/java/java_message_builder.cc
```diff
@@ -473,26 +473,32 @@ GenerateCommonBuilderMethods(io::Printer* printer) {
     "public Builder setField(\n"
     "    com.google.protobuf.Descriptors.FieldDescriptor field,\n"
     "    Object value) {\n"
+    "  //@@protoc_insertion_point(field_modifier_scope_after:$full_name$.setField)\n"
```
xfxyjwf: Please don't add insertion points here. These methods are not meant to be customized.

src/google/protobuf/compiler/java/java_map_field.cc
gturri: It seems a good idea, indeed. I'll make the change

src/google/protobuf/compiler/java/java_message_builder.cc
gturri: Ok, I'll try to find a way to do this patch without needing insertion points in this file.

src/google/protobuf/compiler/java/java_message_field.cc
```diff
@@ -299,6 +300,7 @@ GenerateBuilderMembers(io::Printer* printer) const {
     "$name$Builder_.setMessage(value);\n",
 
     "$set_has_field_bit_builder$\n"
+    "// @@protoc_insertion_point(builder_field_modifier_scope_after:$containing_type_full_name$.set$capitalized_name$)\n"
```
gturri: What about the insertion points in this file and the other below:  shall I try to rename them to something like @@protoc_insertion_point(builder_field_XXX:$full_name$) too?

xfxyjwf: Yes, please. @@protoc_insertion_point(accessor_name:$full_name$) is the way we name these insertion points and should be consistent through out.

src/google/protobuf/compiler/java/java_map_field.cc
ychuzevi: Sorry for the delay, but it is now done in trhe latest pull request

src/google/protobuf/compiler/java/java_message_builder.cc
ychuzevi: Indeed we actually didn't need to customize those methods. The insertion points are no longer present in latest pull request

src/google/protobuf/compiler/java/java_message_field.cc
ychuzevi: Done

ychuzevi:
I see that Travis job is still tagged as pending. All of the builds succeed except one (php7.0_mac) which time-outed (it doesn't seem related to my changes though). What can I do to move forward?

ychuzevi:
Rebased and added missing insertion points for repeated enums.
@xfxyjwf , would you be available to have a look?

xfxyjwf:
In recent discussions about protoc insertion points, several others expressed their concerned about allowing customizing generated code and some teams using these insertion points had to replace them when importing the code into Google's internal code base due to incompatibilities with our build system. Given these, our current consensus is that we should move away from insertion points and discourage its usage.

PR #3317, opened 2017-07-05
protoc-artifacts: Bump JDK to 8u131

ejona86:
The update 45 download process no longer works, as it requires a login
to access the old build.

CC @zhangkun83


bazel-io:

PR #3327, opened 2017-07-07
Use consistent hash across NDEBUG/!NDEBUG builds.

### LINKED ISSUE #3322: C++ protobufs can't be safely used in a mix of NDEBUG and !NDEBUG code
Due to code such as https://github.com/google/protobuf/blob/master/src/google/protobuf/map.h#L868, it's not in general safe to use a protobuf in one C++ module that is built with NDEBUG with code in another that is built without NDEBUG. E.g. passing an in-memory proto between the modules via a function call. This results in hard to discover corruptions of the mapped field hash table.

While this is acceptable in a monorepo, in general OSS where C++ dependencies are sourced from prebuilt libraries, this seems like a bad situation. Can we remove this in OSS protobuf?

htuch:
Fixes #3322.

### DIFF
--- a/src/google/protobuf/map.h
+++ b/src/google/protobuf/map.h
@@ -862,14 +862,7 @@ class Map {
     size_type BucketNumber(const Key& k) const {
       // We inherit from hasher, so one-arg operator() provides a hash function.
       size_type h = (*const_cast<InnerMap*>(this))(k);
-      // To help prevent people from making assumptions about the hash function,
-      // we use the seed differently depending on NDEBUG.  The default hash
-      // function, the seeding, etc., are all likely to change in the future.
-#ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


bazel-io:

xfxyjwf:
Can you send this as a CL instead? Better to have the original author of this code double check.

htuch:
@xfxyjwf Ack, I've filed a CL. I'll close this for now but keep the issue open until we've integrated the CL and pushed upstream.

htuch:
The CL has merged internally, as discussed offline, can we get this PR to merge as well? I think it would be useful to Envoy to remove the workaround we have in place as early as possible. Thanks.

bazel-io:

xfxyjwf:
Thanks for helping resolve this!

PR #3331, opened 2017-07-07
adds PHP typehints for message-type setters

bshaffer:
This adds typehints for all Message setters. A few other interesting things have happened as a result:

  1. Setters for Message types no longer pass-by-reference, since this is implicit for PHP objects
  1. Setters for Message types now accept an optional `null`, since the base `Message` class requires passing `null` to the public setters in order for the method `clear()` to work.

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -558,10 +558,18 @@ void GenerateFieldAccessor(const FieldDescriptor* field, bool is_descriptor,
 
   // Generate setter.
   GenerateFieldDocComment(printer, field, is_descriptor, kFieldSetter);
-  printer->Print(
-      "public function set^camel_name^($var)\n"
-      "{\n",
-      "camel_name", UnderscoresToCamelCase(field->name(), true));
+  if (field->type() == FieldDescriptor::TYPE_MESSAGE && !field->is_repeated()) {
+    printer->Print(
+        "public function set^camel_name^(^typehint^ $var = null)\n"
+        "{\n",
+        "camel_name", UnderscoresToCamelCase(field->name(), true),
+        "typehint", "\\" + FullClassName(field->message_type(), is_descriptor));
+  } else {
+    printer->Print(
+        "public function set^camel_name^($var)\n"
+        "{\n",
+        "camel_name", UnderscoresToCamelCase(field->name(), true));
+  }
 
   Indent(printer);
 
--- a/php/src/Google/Protobuf/Internal/DescriptorProto.php
+++ b/php/src/Google/Protobuf/Internal/DescriptorProto.php
@@ -286,7 +286,7 @@ class DescriptorProto extends \Google\Protobuf\Internal\Message
      * @param \Google\Protobuf\Internal\MessageOptions $var
      * @return $this
      */
-    public function setOptions($var)
+    public function setOptions(\Google\Protobuf\Internal\MessageOptions $var = null)
     {
         GPBUtil::checkMessage($var, \Google\Protobuf\Internal\MessageOptions::class);
         $this->options = $var;
--- a/php/src/Google/Protobuf/Internal/EnumDescriptorProto.php
+++ b/php/src/Google/Protobuf/Internal/EnumDescriptorProto.php
@@ -108,7 +108,7 @@ class EnumDescriptorProto extends \Google\Protobuf\Internal\Message
      * @param \Google\Protobuf\Internal\EnumOptions $var
      * @return $this
      */
-    public function setOptions($var)
+    public function setOptions(\Google\Protobuf\Internal\EnumOptions $var = null)
     {
         GPBUtil::checkMessage($var, \Google\Protobuf\Internal\EnumOptions::class);
         $this->options = $var;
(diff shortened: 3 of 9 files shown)


bazel-io:

TeBoring:
ok to test

bshaffer:
Looks like the build timed out. Is this a flakey error?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -533,13 +546,19 @@ void GenerateFieldAccessor(const FieldDescriptor* field, bool is_descriptor,
...
   // Generate setter.
   GenerateFieldDocComment(printer, field, is_descriptor, kFieldSetter);
-  printer->Print(
-      "public function set^camel_name^(^var^)\n"
-      "{\n",
-      "camel_name", UnderscoresToCamelCase(field->name(), true),
-      "var", (field->is_repeated() ||
-              field->cpp_type() == FieldDescriptor::CPPTYPE_MESSAGE) ?
-             "&$var": "$var");
+  if (field->type() == FieldDescriptor::TYPE_MESSAGE && !field->is_repeated()) {
```
michaelbausor: I think the logic here would be clearer if broken in to 3 cases:

```
if (field->is_repeated()) {
  // &$var
} else if (field->type() == FieldDescriptor::TYPE_MESSAGE) {
  // typehint $var = null
} else {
  // $var
}
```

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -139,31 +139,44 @@ std::string ClassNamePrefix(const string& classname,
...
-  if (desc->file()->options().has_php_namespace()) {
-    const string& php_namespace = desc->file()->options().php_namespace();
-    if (php_namespace != "") {
-      return php_namespace + '\\' + classname;
-    } else {
-      return classname;
-    }
-  }
+template <typename DescriptorType>
+std::string FullClassName(const DescriptorType* desc, bool is_descriptor) {
```
michaelbausor: This refactoring of FullClassName seems like it is from https://github.com/google/protobuf/pull/3338 but it isn't used in this PR

bshaffer: Good catch, thank you!

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -533,13 +533,23 @@ void GenerateFieldAccessor(const FieldDescriptor* field, bool is_descriptor,
...
-      "camel_name", UnderscoresToCamelCase(field->name(), true),
-      "var", (field->is_repeated() ||
-              field->cpp_type() == FieldDescriptor::CPPTYPE_MESSAGE) ?
-             "&$var": "$var");
+  if (field->is_repeated()) {
+    printer->Print(
+        "public function set^camel_name^(&$var)\n"
+        "{\n",
+        "camel_name", UnderscoresToCamelCase(field->name(), true));
+  } else if (field->type() == FieldDescriptor::TYPE_MESSAGE && !field->is_repeated()) {
```
michaelbausor: Remove `!field->is_repeated()`

TeBoring:
1. LGTM to only pass by value for all field setters.
2. LGTM to allow null for message setter. However, can we make it explicit (users should explicitly set null if they want to)?

TeBoring:
https://stackoverflow.com/questions/14584145/cannot-pass-null-argument-when-using-type-hinting
FYI, nullable typehint only supported since 7.1

TeBoring:
Is this still needed?

server-may-cry:
+1 typehint really helpful

TeBoring:
As talked offline, this is no longer needed because protobuf runtime has its own type checking.

TeBoring:
Reopen this because this may be a performance improvement.

bazel-io:

server-may-cry:
Also type hint help IDE and static analysis tools

bshaffer:
Rebased! Ready for another look.

bshaffer:
@TeBoring the typehints make the call to `GPBType::checkMessage` unnecessary. We can do okne of the following:

1. **Remove from generated classes** - This preserves BC for messages generated with a previous version, as `GPBType::checkMessage` would not change. However, if there is something we'd like to check in the future, we will not be able to do so.
1. **Leave it** - It's a redundant check, but we could leave it in anyway, and remove it in a later version.
1. **Leave it, but have it do nothing** - It's a useless function call, but this would allow us to add other checks someday.

Personally I like 1 and 2, in that order.

bshaffer:
Another issue to consider here is this library throws an exception of class `Exception` if the message does not match the appropriate type, whereas the PHP Typehint will throw a `Catchable fatal error` (PHP 5.6) or `TypeError`(PHP 7.x).

While it would be unlikely to find an app depending on this behavior (exceptions are exceptional after all), this is technically a backwards compatibility break. If we wanted to play it absolutely safe, this should wait for a major version release.

TeBoring:
This is a good point. However, I don't think it will break backward compatibility because setter type checking is an undefined behavior. We have never promised whether having type checking and what kind of type checking.

xfxyjwf:
Ping @TeBoring

xfxyjwf:
@TeBoring Can you advise what needs to be done here?

TeBoring:
Can we also send this to php-generated-dev branch?

TeBoring:
This is no longer needed.

PR #3334, opened 2017-07-07
imports GPBType instead of using complete classname

bshaffer:
`GPBType` is already imported in the message class, but the absolute classname is being used everywhere. Because we've imported it, we can make the generated code *much* more readable by using the shortened name.

### DIFF
--- a/php/src/GPBMetadata/Google/Protobuf/Internal/Descriptor.php
+++ b/php/src/GPBMetadata/Google/Protobuf/Internal/Descriptor.php
@@ -4,6 +4,8 @@
 
 namespace GPBMetadata\Google\Protobuf\Internal;
 
+use Google\Protobuf\Internal\GPBType;
+
 class Descriptor
 {
     public static $is_initialized = false;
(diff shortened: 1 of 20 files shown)


bazel-io:

TeBoring:
I think explicitly mention full class path is necessary in generated code.
Users can define their \OwnNamespace\GPBType and cause a conflict,

server-may-cry:
same in [3338](/google/protobuf/pull/3338)
@TeBoring , don't worry about conflict because of
```php
use Google\Protobuf\Internal\GPBType;
use Google\Protobuf\Internal\GPBWire;
use Google\Protobuf\Internal\RepeatedField;
use Google\Protobuf\Internal\InputStream;
```
so
```php
\OwnNamespace\GPBType
```
will not be used

xfxyjwf:
ping @TeBoring

anandolee:
ping @TeBoring any update?

TeBoring:
Close this for now. Feel free to reopen if there is still need for this feature.

PR #3335, opened 2017-07-08
changed the URL for google tests it was failing for protobuf version 2.5

adityaka:
autogen.sh needs update in the older version. Later version autogen.sh have the updated google test URLs. protobuf v2,5 isn't available easily on the latest linux builds,

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -19,8 +19,11 @@ fi
 # directory is set up as an SVN external.
 if test ! -e gtest; then
   echo "Google Test not present.  Fetching gtest-1.5.0 from the web..."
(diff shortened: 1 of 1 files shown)


bazel-io:

xfxyjwf:
@adityaka We don't do 2.x releases anymore. If you have an issue with autogen.sh file, you can download the release tar file we put up in the release page:
https://github.com/google/protobuf/releases/download/v2.5.0/protobuf-2.5.0.tar.gz

This tar file has gtest included already and you can build it with "./configure && make" without autogen.sh.

Issue #3300, opened 2017-06-30
MapEntry pass-by-reference error


bshaffer:
The `MapEntry` field has [pass-by-reference setters](https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/MapEntry.php#L42). Because of this, the result of `defaultValue` in `Message` being called [in the constructor](https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L73) needs to be set to a variable before being passed in, to avoid throwing a Strict Standards error. Something like this will work:

```php
$setter = $field->getSetter();
$defaultValue = $this->defaultValue($field);
$this->$setter($defaultValue);
```

Otherwise this error is thrown: 

```
Strict Standards: Only variables should be passed by reference
```

bshaffer:
@Teboring can we remove pass-by-reference for repeated field setters? In my own code, I get errors when I try something like this:

```php
$label = new Label();
$metric->setLabels([$label]);
```

This is a common use-case in PHP, and pass-by-reference is rarely used.

bshaffer:
How to duplicate:

1. Check out the master branch
2. Go to the `php` directory
3. Run `phpunit tests/encode_decode_test.php`

I get the following output:

```
PHPUnit 5.3.2 by Sebastian Bergmann and contributors.

EEE.........E...........................                          40 / 40 (100%)

Time: 404 ms, Memory: 16.75Mb

There were 4 errors:

1) EncodeDecodeTest::testEncode
Only variables should be passed by reference

/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:73
/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:983
/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:1003
/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:1028
/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:1066
/Library/WebServer/Code/protobuf/php/tests/encode_decode_test.php:24

2) EncodeDecodeTest::testDecode
Only variables should be passed by reference

/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:73
/Library/WebServer/Code/protobuf/php/src/Google/Protobuf/Internal/Message.php:341
[...]

PR #3337, opened 2017-07-10
fixes 'Only variables should be passed by reference' error for PHP

bshaffer:
See #3300


bazel-io:

bshaffer:
And since we will remove all pass-by-references (#3339), this one is obsolete.

PR #3338, opened 2017-07-10
Only use full PHP classnames when necessary

bshaffer:
/cc @michaelbausor

### DIFF
--- a/php/tests/generated_class_test.php
+++ b/php/tests/generated_class_test.php
@@ -13,6 +13,7 @@ use Foo\TestIncludeNamespaceMessage;
 use Foo\TestIncludePrefixMessage;
 use Foo\TestMessage;
 use Foo\TestMessage_Sub;
+use Foo\TestNamespaces;
 use Php\Test\TestNamespace;
 
 class GeneratedClassTest extends TestBase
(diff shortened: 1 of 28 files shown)


bazel-io:

TeBoring:
I think explicitly mention full class path is necessary in generated code.
Users can define their \OwnNamespace\DescriptorProto and cause a conflict.

bshaffer:
@TeBoring if we can safely implement coding standards, one of which is to only include the full class path when necessary, then why would we not want to do so?

bshaffer:
To be clear, if a custom namespace is used, then the full path to `DescriptorProto` would be used in the generated classes.

TeBoring:
What's the benefit of this PR?

bshaffer:
@TeBoring To make the generated classes more idiomatic and readable!

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -139,31 +139,55 @@ std::string ClassNamePrefix(const string& classname,
...
+  } else {
+    return php_classname;
   }
+}
 
-  if (desc->file()->package() == "") {
-    return classname;
+template <typename DescriptorType>
+std::string RelativeClassName(const FieldDescriptor* field, const DescriptorType* type, bool is_descriptor) {
+  if (PhpNamespace(field, is_descriptor) == PhpNamespace(type, is_descriptor)) {
```
bshaffer: Add a test case to ensure `A\Foo` would not conflict with `A\B\Foo`

bshaffer:
@TeBoring added tests for every permutation I could think of. The tests cover phpdoc and also call the getters and setters to ensure exceptions are not thrown.

TeBoring:
ok to test

php/tests/generated_class_test.php
```diff
@@ -679,6 +680,40 @@ public function testNamespaceMessage()
     }
 
     #########################################################
+    # Test across different packages and php namespaces
+    #########################################################
+
+    public function testNamespaces()
```
TeBoring: Please try \TestNamespaces

bshaffer: How should I test for this? If I do `class_exists('\TestNamespaces');`, this will return false. If I define `\TestNamespaces`, I will get an import error because I am importing that classname at the top of the test.

I could define a class `\TestNamespaces` in this file, and import `\Foo\TestNamespaces as FooTestNamespaces`. If this is what you're testing for, that's implemented on a language level, and PHP will not accept `\[classname]` for `[classname]` in a file with a namespace.

xfxyjwf:
@TeBoring Why is this removed from the project?

xfxyjwf:
@TeBoring What's the status of this PR? If we are not going to accept it, please explain why and close it. If you plan to accept it, please let @bshaffer know what needs to be done to get it merged.

TeBoring:
Close this for now. Feel free to reopen if there is still need for this feature.

PR #3343, opened 2017-07-11
Add HHVM Tests

RyanGordon:
This is an attempt to add HHVM testing. We are currently blocked on waiting on Debian 9 packaging to be released by HHVM team sometime in the next few months.


bazel-io:

TeBoring:
#3349 may have some effect on this PR.
Do you have time to add some automated test for HHVM? That will be very appreciated.

RyanGordon:
Sure, let me try adding hhvm to the tests

RyanGordon:
Okay I believe I added it. HHVM currently has experimental mac support but it needs to be recompiled each time (https://docs.hhvm.com/hhvm/installation/mac). Thoughts on adding the mac version to travis CI?

TeBoring:
Is there any third party source that has put their build on S3?

.travis.yml
```diff
@@ -49,6 +50,10 @@ matrix:
     # which doesn't work on OS X.
     - os: osx
       env: CONFIG=golang
+    # HHVM currently uses apt-get to install hhvm and doesn't have a stable
+    # OS X build yet
+    - os: osx
+      env: CONFIG=hhvm
   include:
```
TeBoring: I think you should add the test in include part instead of excluding the unused.

TeBoring:
ok to test

.travis.yml
RyanGordon: :+1: ee141a9

RyanGordon:
@TeBoring I looked around but I couldn't find anything. Does the linux test run in any of the automated test suites or is it local only?

TeBoring:
It's run in latest travis test now.
I am thinking, since we only test linux, can we set up test in jenkins instead of travis?
You can do that by modifying jenkins/dockers/Dockerfile

RyanGordon:
Sure! Let me look at that

TeBoring:
It seems that tests are still failing. Is there any coming fixes?

RyanGordon:
Yeah, I'm having trouble getting HHVM installed through apt-get. Do you know if I should be targeting Debian 7 / 8 / 9 or CentOS 6 / 7 for this?

TeBoring:
https://github.com/google/protobuf/blob/master/jenkins/docker/Dockerfile#L10
I think you should use debian. For hhvm install, is there difference among debian 7/8/9?

RyanGordon:
Yeah, they currently only offer packages for debian 7 and 8 which each have different dependencies. Currently the failure is dependency installation problems

TeBoring:
@xfxyjwf, shall we have a fixed version of debian for jenkins test?

TeBoring:
https://www.howtoinstall.co/en/debian/stretch/hhvm
It seems you can install hhvm on debian 9.

RyanGordon:
Debian 9 stretch was released on June 17th and it doesn't look like the docker file had been updated since June 15th so it appears to have been getting cached to Debian 8. When I bump it down to Debian 8, and fix one other thing I get closer. Currently stuck on "libgd3 : Depends: libjpeg8 (>= 8c) but it is not installable". It seems like debian switched to libjpeg62-turbo but I'm not sure why the libgd3 library still uses it.

TeBoring:
Can we install all potential dependencies we need?

RyanGordon:
It doesn't appear that libjpeg8 is available on Debian Jessie which is the main problem. Maybe there's some alternate libgd3 package we can use instead?

TeBoring:
https://docs.hhvm.com/hhvm/installation/linux#debian-8-jessie
Are you following this document?

RyanGordon:
Yep, that's exactly what I added to the dockerfile

TeBoring:
FYI, I could install hhvm with the following Dockerfile:
```
FROM debian:8

RUN apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0x5a16e7281be7a449
RUN echo deb http://dl.hhvm.com/debian jessie main | tee /etc/apt/sources.list.d/hhvm.list
RUN apt-get update
RUN apt-get install -y --force-yes hhvm
```

TeBoring:
It seems other repository has side-effect.

TeBoring:
I added a fix to install hhvm in #3415. Let me know if that works.

TeBoring:
@RyanGordon are you still working on this?

RyanGordon:
Hi @TeBoring, yes I ran it last night and had some failures. I'm running it again with your latest commit to see if that fixes it

RyanGordon:
Hi @TeBoring, I tested your Dockerfile out and it does indeed appear to work. However, how are we going to install the regular php versions / phpunit on top of it though? Did those cause conflicts with installing hhvm?

TeBoring:
I think it also install php from source.

TeBoring:
Did you use another email to commit the change?

RyanGordon:
Nope, I just merged in your changes to see if the Jenkins tests would pass here

RyanGordon:
@TeBoring Can I split off the HHVM code fixes from waiting for HHVM testing to work? I can open a separate PR with just the HHVM code fixes and once the Debian 9 package is ready, I can get the tests working here

TeBoring:
SGTM

RyanGordon:
:+1: Done here: https://github.com/google/protobuf/pull/3437

PR #3349, opened 2017-07-11
PHP array constructors for protobuf messages

bshaffer:
RFC for adding array construction for PHP messages, i.e.

```php
$m = new TestMessage([
    'int' => -42,
    'message' => ['a' => 33],
    'repeated_message' => [
        new TestMessage_Sub(['a' => 34]),
        ['a' => 35],
    ],
    'map_message' => [
        1 => new TestMessage_Sub(['a' => 36])
    ],
]);
```

### DIFF
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -862,6 +862,42 @@ class Message
                         "Bool field only accept bool value");
                 }
                 return $value;
+            case GPBType::INT64:
+            case GPBType::SINT64:
+            case GPBType::SFIXED64:
+                if (is_null($value)) {
+                    return $this->defaultValue($field);
+                }
+                if (!is_numeric($value)) {
+                   throw new GPBDecodeException(
+                       "Invalid data type for int64 field");
+                }
+                if (bccomp($value, "9223372036854775807") > 0) {
+                    throw new GPBDecodeException(
+                        "Int64 too large");
+                }
+                if (bccomp($value, "-9223372036854775808") < 0) {
+                    throw new GPBDecodeException(
+                        "Int64 too small");
+                }
+                return $value;
+            case GPBType::UINT64:
+            case GPBType::FIXED64:
+                if (is_null($value)) {
+                    return $this->defaultValue($field);
+                }
+                if (!is_numeric($value)) {
+                   throw new GPBDecodeException(
+                       "Invalid data type for int64 field");
+                }
+                if (bccomp($value, "18446744073709551615") > 0) {
+                    throw new GPBDecodeException(
+                        "Uint64 too large");
+                }
+                if (bccomp($value, "9223372036854775807") > 0) {
+                    $value = bcsub($value, "18446744073709551616");
+                }
+                return $value;
         }
         return $this->convertNativeValueToProtoValue(
             $value,
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -831,6 +831,9 @@ class Message
                 }
                 return $submsg;
             case GPBType::BYTES:
+                if (is_null($value)) {
+                    return $this->defaultValue($field);
+                }
                 if (!is_string($value)) {
                     throw new GPBDecodeException("Expect string");
                 }
@@ -841,6 +844,9 @@ class Message
                 }
                 return $proto_value;
             case GPBType::BOOL:
+                if (is_null($value)) {
+                    return $this->defaultValue($field);
+                }
                 if ($is_map_key) {
                     if ($value === "true") {
                         return true;
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -1012,6 +1012,9 @@ class Message
                     throw new GPBDecodeException(
                         "Int64 too small");
                 }
+                if (PHP_INT_SIZE == 4) {
+                    return (string) $value;
+                }
                 return $value;
             case GPBType::UINT64:
             case GPBType::FIXED64:
@@ -1028,6 +1031,9 @@ class Message
                 }
                 if (bccomp($value, "9223372036854775807") > 0) {
                     $value = bcsub($value, "18446744073709551616");
+                }
+                if (PHP_INT_SIZE == 4) {
+                    return (string) $value;
                 }
                 return $value;
             default:
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -66,19 +66,30 @@ class Message
     /**
      * @ignore
      */
-    public function __construct($desc = NULL)
+    public function __construct($data = NULL)
     {
         // MapEntry message is shared by all types of map fields, whose
         // descriptors are different from each other. Thus, we cannot find a
         // specific descriptor from the descriptor pool.
-        if (get_class($this) === 'Google\Protobuf\Internal\MapEntry') {
-            $this->desc = $desc;
-            foreach ($desc->getField() as $field) {
-                $setter = $field->getSetter();
-                $this->$setter($this->defaultValue($field));
+        if ($data instanceof Descriptor) {
+            $this->initWithDescriptor($data);
+        } else {
+            $this->initWithGeneratedPool();
+            if (is_array($data)) {
+                $this->mergeFromArray($data);
+            } else if (!empty($data)) {
+                throw new \InvalidArgumentException(
+                    'Message constructor must be an array, Descriptor, or null.'
+                );
             }
-            return;
         }
+    }
+
+    /**
+     * @ignore
+     */
+    private function initWithGeneratedPool()
+    {
         $pool = DescriptorPool::getGeneratedPool();
         $this->desc = $pool->getDescriptorByClassName(get_class($this));
         if (is_null($this->desc)) {
@@ -148,6 +159,19 @@ class Message
                         $this->$setter("0");
                 }
             }
+        }
+    }
+
+    /**
+     * @ignore
+     */
+    private function initWithDescriptor(Descriptor $desc)
+    {
+        $this->desc = $desc;
+        foreach ($desc->getField() as $field) {
+            $setter = $field->getSetter();
+            $defaultValue = $this->defaultValue($field);
+            $this->$setter($defaultValue);
         }
     }
 
@@ -737,6 +761,23 @@ class Message
         }
     }
 
+    private function convertToProtoValueImpl(
+        $is_json,
+        $value,
+        $field,
+        $is_map_key = false)
+    {
+        return $is_json
+            ? $this->convertJsonValueToProtoValue(
+                $value,
+                $field,
+                $is_map_key)
+            : $this->convertNativeValueToProtoValue(
+                $value,
+                $field,
+                $is_map_key);
+    }
+
     private function convertJsonValueToProtoValue(
         $value,
         $field,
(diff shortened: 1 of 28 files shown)


bazel-io:

TeBoring:
As we talked, can we add a optional config parameter and make php array an element of it?

TeBoring:
ok to test

bshaffer:
We discussed it at our PHP sync, and would like to keep the existing implementation for the following reasons:
  - It's intuitive that the data for the object is the first argument.
  - The second argument could be an `$options` array, if we end up needing that
  - Other languages use the first argument for message data (go, nodejs, python)
  - Apiary implemented it this way

I would also be ok with just exposing `fromArray` and using that in our samples, although it's somewhat less optimal. This change would look like this:

```php
$m = new TestMessage();
$m->fromArray([
    'int' => -42,
    'message' => ['a' => 33],
    'repeated_message' => [
        new TestMessage_Sub(['a' => 34]),
        ['a' => 35],
    ],
    'map_message' => [
        1 => new TestMessage_Sub(['a' => 36])
    ],
]);
```

TeBoring:
I would like to add the mergeFromArray (similar to fromArray with merge semantic) first, and see whether a special constructor is still needed.

bshaffer:
@TeBoring as in, add the `mergeFromArray` functionality for the next release, and then in a future release add the ability to pass arrays in the constructor?

TeBoring:
If `mergeFromArray` is implemented quickly enough, we can add that in this release.
Need to get more feedback from users to see whether we need the constructor change.

bshaffer:
@TeBoring I assume we need to implement it in the C-extension as well? What do you think of my implementation? If it looks good I can add it to the extension.

TeBoring:
I remember I have already implemented the private method mergeFromJsonArray on Message.
You can make it public and add a c extension implementation for that.

TeBoring:
@michaelbausor @tmatsuo Do you feel mergeFromJsonArray a good name?

michaelbausor:
I have a small preference for mergeFromArray over mergeFromJsonArray, just because it is shorter and I don't know if mergeFromJsonArray is more informative.

Whatever name we choose, it should be consistent between the c extension and the PHP implementation.

tmatsuo:
@TeBoring I slightly prefer `mergeFromArray`, but `mergeFromJsonArray` also works.

BTW, what makes you hesitate to change the constructor? In my understanding, there's any compatibility issues

TeBoring:
Let's add mergeFromArray then. A simple refactoring for php will be enough.
For the constructor, I don't know whether it's still necessary if we have mergeFromArray. In order to only add necessary features, I want to see whether there is still strong need to change the constructor after mergeFromArray is public.

bshaffer:
There is a difference between `mergeFromArray` and `mergeFromJsonArray` in this implementation. The differences are as follows:

1. `mergeFromArray` accepts actual objects for message fields. This logic could be allowed in JSON array also, as there's no compatibility here.
2. `mergeFromArray` expects integers `0` and `1` for Boolean map keys, where `mergeFromJsonArray` expects strings `"true"` and `"false"` as Boolean map keys. We could accept both here, and there'd be no compatibility issue.
3. `mergeFromArray` expects a raw byte string for bytes, whereas `mergeFromJsonArray` expects a base64 encoded byte string. This causes a compability issue here, and I'm not sure how we would resolve this.

TeBoring:
When doing the refactoring, can we add an implementation layer to specify the strictness of type checking.
For `mergeFromJsonArray`, we need strict type checking. For `mergeFromArray`, we can allow array=>message, 1/0=>true/false and raw bytes for bytes field.

bshaffer:
We can add `mergeFromArray`, and then our samples ([for Monitoring, for instance](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/monitoring/quickstart.php)) which currently look like this:

```php
$client = new MetricServiceClient();
$formattedProjectName = MetricServiceClient::formatProjectName($projectId);
$labels = [
    'instance_id' => $instanceId,
    'zone' => $zone,
];

$m = new Metric();
$m->setType('custom.googleapis.com/my_metric');

$r = new MonitoredResource();
$r->setType('gce_instance');
$r->setLabels($labels);

$value = new TypedValue();
$value->setDoubleValue(3.14);

$timestamp = new Timestamp();
$timestamp->setSeconds(time());

$interval = new TimeInterval();
$interval->setStartTime($timestamp);
$interval->setEndTime($timestamp);

$point = new Point();
$point->setValue($value);
$point->setInterval($interval);
$points = [$point];

$timeSeries = new TimeSeries();
$timeSeries->setMetric($m);
$timeSeries->setResource($r);
$timeSeries->setPoints($points);

$client->createTimeSeries($formattedProjectName, [$timeSeries]);
```

Would look something like this:

```php
$client = new MetricServiceClient();
$formattedProjectName = MetricServiceClient::formatProjectName($projectId);

$timeSeries = new TimeSeries()
$timeSeries->mergeFromArray([
    'metric' => [
        'type' => 'custom.googleapis.com/my_metric'
    ],
    'resource' => [
        'type' => 'gce_instance',
        'labels' => [
            'instance_id' => $instanceId,
            'zone' => $zone,
        ],
    ],
    'points' => [
        'value' => ['doubleValue' => '3.14'],
        'interval' => [
            'startTime' => ['seconds' => time()]
            'endTime' => ['seconds' => time()]
        ]
    ]
]);

$client->createTimeSeries($formattedProjectName, [$timeSeries]);
```

This works for me. If we can make `value` and `interval` accept native types, then this will look almost perfect ("almost" because passing the data in the constructor is still preferable).

I will submit a PR for just the `mergeFromArray` implementation, and another for passing the array in the constructor. That way we can track further discussion on using the array constructor there.

One thing we should note is that the first argument of the constructor is currently being used in the `Message` class. It seems like we should remove this, if we are wanting to keep the constructor neutral for now.

TeBoring:
LGTM for the overall approach.
I still have some questions.
First, the first element of Message currently is used for implementing MapEntry. As we talked before, can we wrap everything in a $config parameter and use it as the first parameter?
Second, do we need to allow Message value in the array (besides the array style of the Message)? Having to create the submessage first, put it into the array and pass the array to the containing message seems cumbersome to me.

bshaffer:
@TeBoring 
> First, the first element of Message currently is used for implementing MapEntry. As we talked before, can we wrap everything in a $config parameter and use it as the first parameter?

I can remove the constructor changes all together if you'd line, but I'd *really* prefer to not make a wrapping `$config` array. Here are the reasons (mentioned above as well)

- It's intuitive that the data for the object is the first argument.
- The second argument could be a `$config` array, if we end up needing that. Currently we don't have any "config" requirements, and I believe the `MapEntry` logic could be implemented without this.
- The other supported languages use the first argument for message data (go, nodejs, python)
- Apiary implemented it this way

@michaelbausor and @tmatsuo also agree that using `$data` for the first constructor argument is the way to go.

> Second, do we need to allow Message value in the array (besides the array style of the Message)? Having to create the submessage first, put it into the array and pass the array to the containing message seems cumbersome to me.

It is not required to pass messages in as submessages in the array - array syntax is also supported, so it seems like a win-win to me. There may be times when a user already has a configured sub-message, or they'd prefer to configure their message object in PHP, so it seems consistent to support this also.

TeBoring:
SGTM for making the first parameter for value array.
For submessage value, I suggest do not accept at the moment. Wait and see if there are actual use cases.

bshaffer:
@TeBoring I believe it is intuitive that if you're populating values from an array, you should be able to pass in a submessage object.

bshaffer:
@TeBoring my concern is that a user will try something like this, and get an error:

```php
$timeSeries = new TimeSeries([
    'metric' => new Metric(['type' => 'my_metric'])
]);
```

Would it be better if we checked for the specific class like this:

```php
$klass = $field->getMessageType()->getClass();
if ($value instanceof $klass) {
    return $value;
}
```

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -61,19 +61,27 @@ class Message
     /**
      * @ignore
      */
-    public function __construct($desc = NULL)
+    public function __construct($data = NULL)
```
TeBoring: I would prefer to have two parameters here to avoid confusion.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -61,19 +61,27 @@ class Message
...
-        // MapEntry message is shared by all types of map fields, whose
-        // descriptors are different from each other. Thus, we cannot find a
-        // specific descriptor from the descriptor pool.
-        if (get_class($this) === 'Google\Protobuf\Internal\MapEntry') {
-            $this->desc = $desc;
-            foreach ($desc->getField() as $field) {
-                $setter = $field->getSetter();
-                $this->$setter($this->defaultValue($field));
+        if ($data instanceof Descriptor) {
+            $this->fromDescriptor($data);
```
TeBoring: initWithDescriptor?

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -61,19 +61,27 @@ class Message
...
+        } else {
+            $this->fromDescriptorPool();
```
TeBoring: initWithGeneratedPool()?

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -61,19 +61,27 @@ class Message
     /**
      * @ignore
      */
-    public function __construct($desc = NULL)
+    public function __construct($data = NULL)
     {
-        // MapEntry message is shared by all types of map fields, whose
```
TeBoring: These comments should be kept somewhere.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -705,6 +744,62 @@ private function convertJsonValueToProtoValue(
...
+        if (is_null($value)) {
+            return $this->defaultValue($field);
+        }
+        switch ($field->getType()) {
+            case GPBType::MESSAGE:
+                $klass = $field->getMessageType()->getClass();
+                if ($value instanceof $klass) {
+                    return $value;
+                }
+                if (!is_object($value) && !is_array($value)) {
```
TeBoring: If $value is array, we also need to handle that.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -826,7 +916,7 @@ private function convertJsonValueToProtoValue(
         }
     }
 
-    private function mergeFromJsonArray($array)
+    private function mergeFromArray($array, $is_json = false)
```
TeBoring: Do we still plan to provide mergeFromArray? If yes, I would like to just have $array as parameter. We could have another mergeFromArrayImpl in that case.

php/src/Google/Protobuf/Internal/Message.php
bshaffer: Done. We handle this below when we call `mergeFromArray`.
Also I updated error message to "Expect message or array."

php/src/Google/Protobuf/Internal/Message.php
bshaffer: done

php/src/Google/Protobuf/Internal/Message.php
bshaffer: done

php/src/Google/Protobuf/Internal/Message.php
bshaffer: Added  comments back

php/src/Google/Protobuf/Internal/Message.php
bshaffer: I believe we came to an agreement to keep the data array as the first parameter.

php/src/Google/Protobuf/Internal/Message.php
bshaffer: Good call, I made `mergeFromArray` public and added a private `mergeFromArrayImpl`.

bshaffer:
@TeBoring I updated the code in response to all the comments and synced the PR. PTAL!

bshaffer:
Gentle push on this - Will we be able to get this change in, and if so when can we expect similar changes in the C extension?

Thank you!

TeBoring:
Sorry for the long delay. I understand this improves the code experience quite much. However, I am working on another issue of c extension right now and have no time working on this. Once I finish my current work, I will work on this.

TeBoring:
As we talked, can we send this PR to php-generated-dev

TeBoring:
This is no longer needed.

PR #3354, opened 2017-07-12
C#: Implement IReadOnlyDictionary<K,V> in MapField<K,V>

JRBANCEL:
IReadOnlyDictionary<K,V> was introduced in .Net 4.5, so it is excluded when targeting .Net 3.5.
This feature request was originally mentioned in #2041.

Very similar to #2645 by @jskeet.
In this case IReadOnlyDictionary<K,V>.Keys returns an IEnumerable<K> and IReadOnlyDictionary<K,V>.Values returns an IEnumerable<V>. Even if the already implemented IDictionary<K,V>.Keys returns an ICollection<K> which implements IEnumerable<K>, .Net requires the return type in the interface implementation to be exact, so I added the explicit implementation of IReadOnlyDictionary<K,V>.Keys and IReadOnlyDictionary<K,V>.Values.

There is also a minimal test.

[ReadOnlyDictionary<TKey, TValue>](https://msdn.microsoft.com/en-us/library/gg712875(v=vs.110).aspx) exists but it is less convenient and less efficient than having MapField implement IReadOnlyDictionary.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Collections/MapFieldTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Collections/MapFieldTest.cs
@@ -540,6 +540,22 @@ namespace Google.Protobuf.Collections
             Assert.Throws<ArgumentException>(() => map.ToString());
         }
 
+#if !NET35
+        [Test]
+        public void IDictionaryKeys_Equals_IReadOnlyDictionaryKeys()
+        {
+            var map = new MapField<string, string> { { "foo", "bar" }, { "x", "y" } };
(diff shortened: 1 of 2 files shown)


bazel-io:

jskeet:
Change looks good to me - once we've got the CLA out of the way, I think we can merge this...

JRBANCEL:
I signed it!

jskeet:
Great, thanks - will merge when it's gone through CI.

jskeet:
All required successes have completed - merging.

PR #3357, opened 2017-07-12
Fix build when using -Werror=undef

### LINKED ISSUE #3356: Build broken with -Werror=undef
The protobuf library fails to build when telling the compiler to treat  undefined macros as errors (-Werror=undef).

bklarson:
Correct a couple places where macros were being checked when they
might not exist in some cases.

Fixes #3356.


bazel-io:

bklarson:
I've (just now) signed the CLA.

acozzette:
ok to test (this will trigger the other CI runs)

acozzette:
@bklarson Thanks for the pull request!

PR #3363, opened 2017-07-13
Fix Implicit Return Types

Fleshgrinder:
Both `native_slot_merge` and `native_slot_merge_by_array` have no return type declared. GCC implicitly uses int as return type in such cases, however, these functions are actually void.


bazel-io:

Fleshgrinder:
@TeBoring this is for the PHP extension.

TeBoring:
ok to test

PR #3364, opened 2017-07-13
Added Missing Module Reference

Fleshgrinder:
PHP expects a global reference to the module for static compilation.

### DIFF
--- a/php/ext/google/protobuf/protobuf.h
+++ b/php/ext/google/protobuf/protobuf.h
@@ -31,9 +31,12 @@
 #ifndef __GOOGLE_PROTOBUF_PHP_PROTOBUF_H__
 #define __GOOGLE_PROTOBUF_PHP_PROTOBUF_H__
 
-#include <php.h>
+#include "php.h"
 
-// ubp.h has to be placed after php.h. Othwise, php.h will introduce NDEBUG.
+extern zend_module_entry protobuf_module_entry;
+#define phpext_protobuf_ptr &protobuf_module_entry
+
+// upb.h must come after php.h, otherwise, php.h introduces NDEBUG.
 #include "upb.h"
 
 #define PHP_PROTOBUF_EXTNAME "protobuf"


bazel-io:

Fleshgrinder:
@TeBoring this is for the PHP extension.

TeBoring:
Thanks for the PR. How will this class entry be used?

Fleshgrinder:
They are required in PHP 7 (see [extension skeleton](https://github.com/php/php-src/blob/master/ext/skeleton/php_skeleton.h)), not sure about PHP 5, and especially important if you want to build a static PHP extension (which is exactly what I am trying to achieve).

The reference is required because the startup requires them to retrieve the info about them at startup. Dynamic extensions apparently do not require this, because the can be de-/activated at runtime and at will. Hence, PHP must go through a few indirections before it can start them. I cannot tell you the exact inner workings of PHP here. I have a little knowledge, but not that deep.

TeBoring:
ok to test

Fleshgrinder:
Just FYI, I changed the `<php>` to `"php"` because it isn't a system header file that we are referring to, but rather one of our code base. I know that it seems like a system header if you only think of a dynamic extension, it's a different game if you think of a static extension. 😉

Fleshgrinder:
@TeBoring build failure of AppVeyor is not related to this PR.

TeBoring:
I am trying to understand why you change <php.h> to "php.h". Are you writing your own php.h to hide system php.h?

TeBoring:
Or you don't want to use the system header (pre-installed) for building the extension statically (e.g., php5.5 installed on system, but you are building php7.0 extension)?

TeBoring:
ping?

Fleshgrinder:
Sorry for the late reply, was on vacation. `<>` is for system headers, but PHP is not a system header (especially not if compiling a static extension as part of PHP). Hence, using `""` is the correct thing to do in an extension.

TeBoring:
In array.c, we have something like: #include <Zend/zend_interfaces.h>
Do we also need to change that?

Fleshgrinder:
This is not a hard requirement because most compilers will be able to resolve those references anyways. However, highest compatibility should be the goal and to achieve that, well, yes, it should be changed.

TeBoring:
Could you help update your PR?

Issue #3263, opened 2017-06-22
Mix int64 + string with some interleaved tags causes coredump in PHP


shagren:
It`s very strange but next proto:

```proto
message TestMessage {
    int64 tag13 = 150;
    string tag14 = 160;
}
```

Causes coredump with next php code:

```php
require_once 'TestMessage.php';
$t = new TestMessage();
print $t->serializeToString();
```

If I change tags for 1+2, or 150+151 or change both types to string or int64 it`s works as expected.

I test it on PHP 7.1.6 and protobuf 3.3.2
There is my Dockerfile for reproducing:
```dockerfile
FROM php:7-fpm-alpine
LABEL maintainer  "Ilya Gruzinov <igruzinov@yandex.ru>"
VOLUME /app
WORKDIR /app
RUN apk add --no-cache git openssh-client gd gettext-libs libintl icu-libs
RUN apk add --no-cache --no-cache --virtual .protobuf-dbuild-deps autoconf automake libtool curl make g++ unzip

#Required php extensions
RUN apk add --virtual .build-deps --no-cache libjpeg-turbo-dev libpng-dev curl-dev zlib-dev icu-dev gettext-dev cyrus-sasl-dev freetype-dev autoconf gcc libc-dev make && \
    docker-php-source extract && \
    pecl bundle -d /usr/src/php/ext xdebug && \
    pecl bundle -d /usr/src/php/ext redis && \
    docker-php-ext-configure gd --with-gd  --with-freetype-dir=/usr/include/ --with-png-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && \
    docker-php-ext-install -j "$(grep processor /proc/cpuinfo | wc -l)" gd mbstring gettext intl opcache xdebug zip redis

#Protobuff v3.3 from git, library+protoc+extension
RUN cd /tmp && \
    git clone https://github.com/google/protobuf.git -b v3.3.2 --depth 1 && \
    cd protobuf && \
    ./autogen.sh && \
    ./configure --prefix=/usr && \
    make -j "$(grep processor /proc/cpuinfo | wc -l)" && \
    make install && \
    cp -r php/ext/google/protobuf /usr/src/php/ext && \
    docker-php-ext-install -j "$(grep processor /proc/cpuinfo | wc -l)" protobuf

#Cleanup
RUN apk del .build-deps .protobuf-dbuild-deps && \
    docker-php-source delete 

```

PR #3367, opened 2017-07-13
Initial value in generated code cannot be used by c extension.

TeBoring:
In the generated code of previous versions, each php field is given an
initial value. In c extension, it was assumed that the field order in
the generated code is consistent with upb fields order, so that the
correct initial value can be bound to the correct upb field. However,
this may not be true. The order of fields in generated code is decided
by proto compiler, while the order of upb fields is decided by the hash
function used in c extension.
This PR fixes the issue by reset the initial value at runtime.


bazel-io:

TeBoring:
Fix #3263

PR #3371, opened 2017-07-14
includes the expected class in the exception, otherwise the error is …

bshaffer:
…harder to track down


bazel-io:

xfxyjwf:
@TeBoring What should we do here?

xfxyjwf:
Ping @TeBoring

PR #3374, opened 2017-07-15
Add option to omit insignificant whitespaces from json output.

gabrielhora:
Added a new parameter called omitting_insignificant_whitespace. This
parameter is responsible for defining if the json output should be
indented or not. If False (it's default value) the output will be
formatted with 2 spaces (as it was before), if True no indentation is
set and the output will be in a single line.

This is useful for streaming json output based on NDJSON
(http://ndjson.org/) or Server Side Events where each message should be
contained in a single line.

This is the same functionally that is already present in Java
(at google/protobuf/util/JsonFormat.java#L134).

### DIFF
--- a/python/google/protobuf/json_format.py
+++ b/python/google/protobuf/json_format.py
@@ -88,7 +88,8 @@ class ParseError(Error):
 
 def MessageToJson(message,
                   including_default_value_fields=False,
-                  preserving_proto_field_name=False):
+                  preserving_proto_field_name=False,
+                  omitting_insignificant_whitespace=False):
   """Converts protobuf message to JSON format.
 
   Args:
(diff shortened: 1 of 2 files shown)


bazel-io:

acozzette:
ok to test (this comment will kick off the other CI tests)

python/google/protobuf/json_format.py
```diff
@@ -88,7 +88,8 @@ class ParseError(Error):
 
 def MessageToJson(message,
                   including_default_value_fields=False,
-                  preserving_proto_field_name=False):
+                  preserving_proto_field_name=False,
+                  omitting_insignificant_whitespace=False):
```
anandolee: Maybe add indent option directly

linyatis: @anandolee: It would be nice, but I think it should have the same behavior as in languages where support already exists.
Java example: [constructor](https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L134) and [default usage (`JsonFormat.printer()`)](https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L106-L108)

gabrielhora: I wouldn't mind to change if all think it's better. 

On one hand It's good because it would serve "two purposes", indent or not and how many spaces to use for indentation, but maybe a it's a bit less explict (is it weird to right `MessageToJson(some_dict, indent=None)`? I don't think it is actually), and the point raised by @linyatis is also valid (same behavior between different implementations).

gabrielhora:
Hi there, can we merge this? Anything else I can do to help?

gabrielhora:
I guess someone else ended up doing something similar and got in before this PR https://github.com/google/protobuf/blob/master/python/google/protobuf/json_format.py#L95.

PR #3390, opened 2017-07-18
Add note about includes to README

danielgtaylor-isp:
Ran into an issue today where a machine had the `protoc` compiler but not the include files. Took a while to sort out, and this added note to the README included in every zip should help.


bazel-io:

danielgtaylor-isp:
I signed it!

xfxyjwf:
Thanks for the patch!

PR #3392, opened 2017-07-18
std::is_trivially_copy_assignable is not supported in all compilers even if c++11 is available.

TeBoring:


src/google/protobuf/generated_message_table_driven.cc
```diff
@@ -73,7 +73,7 @@ inline Arena* GetArena(MessageLite* msg, int64 arena_offset) {
 
 template <typename Type>
 inline Type* AddField(MessageLite* msg, int64 offset) {
-#if LANG_CXX11
+#if GOOGLE_HAVE_STD_IS_TRIVIALLY_ASSIGNABLE
```
xfxyjwf: I think you need to keep LANG_CXX11? GOOGLE_HAVE_STD_IS_TRIVIALLY_ASSIGNABLE may be true even when C++11 is disabled (I don't see any check for C++11 there).

TeBoring:
No longer needed, as there is google::protobuf::internal::has_trivial_copy.

PR #3399, opened 2017-07-21
Fix cycle dependency for repeated field not collected by gc

TeBoring:


bazel-io:

PR #3403, opened 2017-07-21
CMake: Install .pc Files

### LINKED ISSUE #1384: protobuf 3.0.0-beta2 cmake build does not install pkg-config files (which confuses grpc)
I'm using cmake to compile and install protobuf 3.0.0-beta2 on MinGW64 / MSYS2. Everything compiles, tests and works fine. However, the protobuf.pc file for pkg-config is not installed in <target>/lib/pkgconfig.

This confuses grpc compile into not finding protobuf, since grpc assumes that pkg-config must have protobuf.pc (if pkg-config is available).

ax3l:
Fix #1384: Adds pkg-config (`.pc`) files for CMake based installs.

### To Do

- [x] pthread libs not yet properly replaced in .pc files (`@PTHREAD_CFLAGS@`, `@PTHREAD_LIBS@`) -> just use `@CMAKE_THREAD_LIBS_INIT@`?

### DIFF
--- a/cmake/protobuf-lite.pc.cmake
+++ b/cmake/protobuf-lite.pc.cmake
@@ -1,0 +1,11 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=@CMAKE_INSTALL_PREFIX@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+
+Name: Protocol Buffers
+Description: Google's Data Interchange Format
+Version: @protobuf_VERSION@
+Libs: -L${libdir} -lprotobuf-lite @CMAKE_THREAD_LIBS_INIT@
+Cflags: -I${includedir} @CMAKE_THREAD_LIBS_INIT@
+Conflicts: protobuf
--- a/cmake/protobuf.pc.cmake
+++ b/cmake/protobuf.pc.cmake
@@ -1,0 +1,11 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=@CMAKE_INSTALL_PREFIX@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+
+Name: Protocol Buffers
+Description: Google's Data Interchange Format
+Version: @protobuf_VERSION@
+Libs: -L${libdir} -lprotobuf @CMAKE_THREAD_LIBS_INIT@
+Cflags: -I${includedir} @CMAKE_THREAD_LIBS_INIT@
+Conflicts: protobuf-lite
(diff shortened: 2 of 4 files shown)


bazel-io:

ax3l:
ping @acozzette @xfxyjwf :)

cmake/protobuf-lite.pc.cmake
```diff
@@ -0,0 +1,11 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=@CMAKE_INSTALL_PREFIX@
+libdir=@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_LIBDIR@
```
ax3l: this two lines still looks fishy.
Somehow, `@CMAKE_INSTALL_LIBDIR@` and `@CMAKE_INSTALL_INCLUDEDIR@` are still empty sometimes, maybe configure it later (in `install.cmake`?)

cmake/protobuf.pc.cmake
```diff
@@ -0,0 +1,12 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=@CMAKE_INSTALL_PREFIX@
+libdir=@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_LIBDIR@
+includedir=@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_INCLUDEDIR@
+
+Name: Protocol Buffers
+Description: Google's Data Interchange Format
+Version: @protobuf_VERSION@
+Libs: -L${libdir} -lprotobuf @CMAKE_THREAD_LIBS_INIT@
+Libs.private: @LIBS@
```
ax3l: this needs to be replaced, unknown var in CMake

ax3l: dunno what to put here, zlib? anyone? :)
otherwise we could also skip it for now ;)

xfxyjwf: Let's drop it for now

acozzette:
ok to test

cmake/protobuf.pc.cmake
ax3l: update: removed it :)

ax3l:
@mcheshkov @emmenlau just ping-ing you in case you want to test this feature: pkg-config files for protobuf's CMake-based install :)

ax3l:
`make dist`: should those two files be added or excluded from `EXTRA_DIST`? (I guess the latter?)

ax3l:
@acozzette all tested :)

acozzette:
@ax3l Thanks for the pull request!

ax3l:
@acozzette wuhu, thanks! :)
Can you tell me which release will first feature it? I could need it in all 3.X releases for [packing](https://github.com/LLNL/spack/pull/4846), but if it's the next major (3.4?) that would be fine, too.

acozzette:
@pherl is working on the 3.4 release now. We've already created a separate 3.4.x branch for that release, but if you want to include this change in 3.4 then I think you could just send another pull request cherry-picking this to the 3.4.x branch. @pherl, does that sound OK?

PR #3407, opened 2017-07-24
Add declaration for AddDescriptorsImpl()

bklarson:
AddDescriptorsImpl() was added without a proper declaration, causing
build errors in more restrictive build environments that use
Werror=missing-declarations.  Add this declaration to fix the build.

Add -Werror=missing-declarations to test builds to prevent this case
from happening in the future.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -925,7 +925,7 @@ no_warning_test_LDADD = $(PTHREAD_LIBS) libprotobuf.la      \
                         ../gmock/gtest/lib/libgtest_main.la
 no_warning_test_CPPFLAGS = -I$(srcdir)/../gmock/gtest/include
 no_warning_test_CXXFLAGS = $(PTHREAD_CFLAGS) $(PTHREAD_DEF) $(ZLIB_DEF) \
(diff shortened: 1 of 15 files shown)


bazel-io:

bklarson:
I put the 'test change' commit first, which will break the build, until the follow-up fix commit is merged as well.  This will result in a broken build if testing that single commit.  I'm happy to reverse the order of the commits if that is preferred - I know some groups want to very that the test works before putting in code that fixes the test.  Others want no broken builds across any commits.

acozzette:
Actually the AddDescriptorsImpl function should really never be called except by the corresponding AddDescriptors function, so I think the best solution for getting rid of the warning would be to put AddDescriptorsImpl into an anonymous namespace.

bklarson:
@acozzette - great call.  Disappointed I didn't think of that originally.  Fixed and tested.

acozzette:
@bklarson Thanks for the pull request! Let me just wait for the full tests to run just to be safe and then I'll merge it.

ok to test (this comment should kick off the other tests)

PR #3412, opened 2017-07-24
Use keys() instead of iterkeys() to be python3 compatbile.

anandolee:


bazel-io:

pherl:
ok to test

PR #3420, opened 2017-07-26
Objc proto2 conformance test support

thomasvl:


conformance/conformance_objc.m
```diff
@@ -63,28 +64,29 @@ static void Die(NSString *format, ...) {
...
 
     case ConformanceRequest_Payload_OneOfCase_ProtobufPayload: {
-      if ([request.messageType isEqualToString:@"protobuf_test_messages.proto3.TestAllTypesProto3"]) {
+      Class msgClass = nil;
+      if ([request.messageType isEqual:@"protobuf_test_messages.proto3.TestAllTypesProto3"]) {
+        msgClass = [Proto3TestAllTypesProto3 class];
+      } else if ([request.messageType isEqual:@"protobuf_test_messages.proto2.TestAllTypesProto2"]) {
+        msgClass = [TestAllTypesProto2 class];
+      } else {
+        Die(@"Protobuf request doesn't have specific payload type");
```
haberman: Error message seems misleading; should be `"unknown message type: " + messageType`

conformance/conformance_objc.m
```diff
@@ -63,28 +64,29 @@ static void Die(NSString *format, ...) {
...
+      }
+      if (msgClass) {
```
haberman: How can this be nil? You called `Die` in the only path that doesn't set this.

thomasvl: Correct, tweaked.

conformance/conformance_objc.m
thomasvl: Yup, that was the message from before.  Tweaked.

BSBandme:
Sorry that I just saw your request. So we need to update the error message for all the other language?

PR #3421, opened 2017-07-26
Update the comment on the message_type to cover what it should be.

thomasvl:


conformance/conformance.proto
```diff
@@ -78,7 +78,9 @@ message ConformanceRequest {
   // Which format should the testee serialize its message to?
   WireFormat requested_output_format = 3;
 
-  // should be set to either "proto2" or "proto3"
+  // The full name for the test message to use; either
```
haberman: Maybe better to say: "for the moment, either:"

We might have reason to add others in the future. Implementations that can lookup messages by name should just do that.

csharp/src/Google.Protobuf.Conformance/Conformance.cs
```diff
@@ -142,7 +142,9 @@ public sealed partial class ConformanceRequest : pb::IMessage<ConformanceRequest
     public const int MessageTypeFieldNumber = 4;
     private string messageType_ = "";
     /// <summary>
-    /// should be set to either "proto2" or "proto3"
+    /// The full name for the test message to use; either
```
haberman: ditto

conformance/conformance.proto
thomasvl: Commit amended.

csharp/src/Google.Protobuf.Conformance/Conformance.cs
thomasvl: regenerated to get it pulled here also.

Issue #1296, opened 2016-03-05
How should protobuf Python be packaged to coexist with other packages in the google namespace?


haberman:
I'm creating this as a master issue to track this long-standing problem. Several people have encountered this problem, but no one seems to know the correct way to solve it.

Here are several issues that say that installing the protobuf package makes other Python packages under the `google.` namespace inaccessible because protobuf has no `__init__.py`. All of these issues were reported against version `<= 3.0.0b2`: https://github.com/google/protobuf/issues/713 https://github.com/google/protobuf/issues/859 https://github.com/google/protobuf/issues/1153 https://github.com/google/protobuf/issues/1272

These issues were supposed to be fixed by this PR: https://github.com/google/protobuf/pull/1259  That PR was integrated into a new release on PyPI, 3.0.0b2.post1 (https://pypi.python.org/pypi/protobuf/3.0.0b2.post1).

However that PR caused a different bug: https://github.com/google/protobuf/issues/1294 https://github.com/GoogleCloudPlatform/gcloud-python/issues/1570

We need a solution to this problem, but finding a solution that works has proven to be elusive. Would very much appreciate anyone who can propose a fix they are confident will fix all of these problems.

awan1:
Currently I've hacked this to work locally by going into the `site-packages/google` folder in my virtualenv (introduced by `pip install protobuf==3.0.0b2.post1`) and introducing a simlink to the `appengine` folder in my GAE SDK install. This will work in general for any package that gets shadowed by the `google` top-level module.

This is very far from optimal and I hope we can work towards a solution!

jonparrott:
*Every* python package that wants to use the `google.*` namespace needs to do the following:

Add `google/__init__.py`:

```python
# this is a namespace package
try:
    import pkg_resources
    pkg_resources.declare_namespace(__name__)
except ImportError:
    import pkgutil
    __path__ = pkgutil.extend_path(__path__, __name__)
```

(`pkg_resources` is preferred because 

Add ``namespace_packages`` to ``setup.py``:

```python
setup(
    name='protobuf',
    namespace_packages=['google'],
    ...
)
```

If any one package doesn't do this and is installed alongside the others, it's very likely to completely break the installation.

We should get a list of all known packages and ensure they conform to this.

craigcitro:
With the other PR merged, we're now at least back to a reasonable working state. (Famous last words?) I think the pending things we should do before closing out this PR include:

[...]

PR #3428, opened 2017-07-31
Include __init__.py for google.* python namespacing in Bazel builds

benley:
Related to #1296: The issue seems to be fixed for consumers of the python protobuf package from pypi, but not for anyone getting it from here as a Bazel remote repository.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -641,6 +641,7 @@ py_library(
     name = "python_srcs",
     srcs = glob(
         [
+            "python/google/__init__.py",
             "python/google/protobuf/*.py",
             "python/google/protobuf/**/*.py",
         ],


bazel-io:

pherl:
OK to test

benley:
Any chance someone could take a look at this soon?  It's currently a bit difficult to use protobuf together with any other google libraries in a Python app because of the namespace collision.

benley:
(slight clarification: difficult to use protobuf and other google libraries in a python app _with bazel_)

raliste:
Also seeing this error, breaking our build when we start introducing google-cloud libs :(

benley:
as a workaround, you could try using a skylark rule to apply a tiny patch to the protobuf sources before building, using something like `patched_http_archive` from https://gist.github.com/benley/e9dfdfef516aa024d50816ec92b7bd9a

anandolee:
I am closing this PR as It has been added in BUILD file

PR #3433, opened 2017-07-31
Removes unnecessary pass-by-references in PHP internal classes

bshaffer:
Adds `Descriptor` typehints to ensure these are already passed by reference. 

The exceptions are 
 - `MapEntry::setKey` and `MapEntry::setValue` do not have typehints. 
 - `DescriptorPool::addEnumDescriptor` and `DescriptorPool::addDescriptor` could have typehints, but since these methods are public I didn't want to break BC by adding them. However, it would be better to have them.

Thoughts?


bazel-io:

TeBoring:
ok to test

PR #3436, opened 2017-08-01
Change divideInt64ToInt32 to static

microtony:
divideInt64ToInt32 is called statically from protobuf/php/src/Google/Protobuf/Internal/CodedOutputStream.php
(the only reference)
This causes fatal error in PHP 7.1 (32-bit only because 64-bit doesn't use this function)


bazel-io:

TeBoring:
ok to test

PR #3445, opened 2017-08-02
Fix signed integer overflow Parser::ParseExtensions

vitalybuka:
Detected by parser_unittest with clang and -fsanitize=signed-integer-overflow


bazel-io:

Issue #3446, opened 2017-08-03
PHP c extension setter error


michaelbausor:
I am seeing an error when using setters in the PHP c extension. It seems that the `binding` field is being set by the `setEtag` method in the example below. The following code uses the Policy class [available here](https://github.com/googleapis/gax-php/blob/master/src/generated/Google/Iam/V1/Policy.php).

```
$policy = new Policy();
$policy->setEtag("myetag");
var_dump($policy->getEtag());
var_dump($policy->getBindings());
```
produces this output:
```
/usr/local/google/home/michaelbausor/dev/veneer/testing/php-release-test/tests/system/Iam/IamTest.php:23:
string(6) "myetag"
/usr/local/google/home/michaelbausor/dev/veneer/testing/php-release-test/tests/system/Iam/IamTest.php:24:
string(6) "myetag"
```

When the c extension is disabled, the output is correct:
```
/usr/local/google/home/michaelbausor/dev/veneer/testing/php-release-test/tests/system/Iam/IamTest.php:23:
string(6) "myetag"
/usr/local/google/home/michaelbausor/dev/veneer/testing/php-release-test/tests/system/Iam/IamTest.php:24:
class Google\Protobuf\Internal\RepeatedField#244 (3) {
  private $container =>
  array(0) {
  }
  private $type =>
  int(11)
  private $klass =>
  string(21) "Google\Iam\V1\Binding"
}
```

TeBoring:
In master?

michaelbausor:
I am finding this issue with master, but also with 3.3.2 installed via pecl.

TeBoring:
https://github.com/google/protobuf/commit/c78dbd7c895b8c80cd2f2401d73acf890edf82c6
Similar bug has been fixed in this PR.
Could you show me the definition of proto?

michaelbausor:
The proto is defined here: https://github.com/googleapis/googleapis/blob/master/google/iam/v1/policy.proto#L60

Looks like the fields are "out of order", in that `bindings = 4` is before `etag = 3`

Should I try to regenerate using a protoc version with that fix?

PR #3455, opened 2017-08-04
Fix the bug in php c extension that setting one field can change another field's value.

TeBoring:
The reason is that previously, in c extension, it was assumed that the order that fields were declared in php is the same as the order of fields in upb. This is not true. Now, for every field in upb, we will look up the actual property that is corresponding to the upb field.

### DIFF
--- a/php/tests/encode_decode_test.php
+++ b/php/tests/encode_decode_test.php
@@ -208,49 +208,49 @@ class EncodeDecodeTest extends TestBase
         $this->expectFieldsMerged($m1);
     }
 
-  public function testDecodeFieldNonExist()
-  {
-      $data = hex2bin('c80501');
-      $m = new TestMessage();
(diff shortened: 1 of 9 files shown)


bazel-io:

TeBoring:
Fix #3446

php/tests/encode_decode_test.php
```diff
@@ -208,49 +208,49 @@ public function testDecodeToExistingMessage()
...
-
-    /**
-     * @expectedException Exception
-     */
-    public function testDecodeInvalidInt32()
-    {
-        $m = new TestMessage();
-        $m->mergeFromString(hex2bin('08'));
-    }
+  public function testDecodeFieldNonExist()
```
michaelbausor: Fix indentation

TeBoring: done

PR #3460, opened 2017-08-04
Merge master into 3.4.x

TeBoring:


bazel-io:

pherl:
LGTM as long as the presubmit is green.

TeBoring:
The test was not run because there was limited space on test machine. After cleaning up the test machine, rebuild was still failed. Created #3461 instead.

PR #3465, opened 2017-08-07
Make .bzl files compatible with future versions of Bazel

vladmos:
In future versions of Bazel += on lists will mutate the lhs list instead of copying it. That will fail if the list is frozen, in that case it has to be copied explicitly.


vladmos:
Jenkins, retest please.

PR #3475, opened 2017-08-09
Use the context ClassLoader where appropriate

### LINKED ISSUE #3474: Issues with Protobuf and Spark
Hi, I wanted to get your thoughts on an issue we've run into with a few of our Spark jobs when they try to serialize protobufs using Java serialization (and it seems like others have run into [this](https://stackoverflow.com/questions/40885430/unable-to-find-proto-buffer-class-in-spark/40904420#40904420) [same](http://apache-spark-user-list.1001560.n3.nabble.com/Unable-to-find-proto-buffer-class-error-with-RDD-lt-protobuf-gt-td14529.html) [issue](http://permalink.gmane.org/gmane.comp.lang.scala.spark.user/4936)). I put an example stack trace [here](https://gist.github.com/jhaber/198bbb1f3a91378cfe4271655021b54f). The error was confusing at first, but looking at the relevant line in GeneratedMessageLite I think I see the problem: https://github.com/google/protobuf/blob/774d630bde574f5fcbb6dae6eaa0f91f7bc12967/java/src/main/java/com/google/protobuf/GeneratedMessageLite.java#L768

The call to `Class.forName` doesn't include a `ClassLoader` argument, so it will attempt to use whichever `ClassLoader` was used to load  `GeneratedMessageLite`. Since Spark itself uses protobuf-java, this ends up being a `ClassLoader` further up the chain that doesn't know anything about our protobuf classes. To confirm this is the issue, we ran this code in the Spark executor:
```java
// fails with ClassNotFoundException
Class.forName("com.hubspot.Protos$ExampleProto", true, GeneratedMessageLite.class.getClassLoader());
// works fine
Class.forName("com.hubspot.Protos$ExampleProto", true, Thread.currentThread().getContextClassLoader());
```

There are a few workarounds in the Spark case, but it seems like this is a more general issue that could crop up outside of Spark. Thoughts on switching all usages of `Class.forName` to include an explicit `ClassLoader` argument? This way, preference could be given to the thread's context class loader, similar to the strategy that's used by the [Resources class in Guava](https://github.com/google/guava/blob/6290373588ef914111f0896b8d0c86a15b7f0f6b/guava/src/com/google/common/io/Resources.java#L194-L195)

I'm interested to hear your thoughts on this issue, thanks

jhaber:
Fix for #3474, thanks for the quick response @xfxyjwf 

I kept the existing behavior on Android, because grpc/grpc-java#1273 suggests that using the context `ClassLoader` may not be relied upon. I wasn't aware of any way to reuse the grpc code, so I just copied the relevant bits over.

I'm also not entirely sure about the change to `java_file.cc`, I don't have much experience with extensions but I think we'd want the same behavior there. 

I followed the instructions [here](https://github.com/google/protobuf/tree/master/src) to build and then I ran the Maven build and all the tests passed. Setting up a unit test to verify that it fixes #3474 would be a bit involved but I'm happy to do it if you think it's worthwhile.


bazel-io:

jhaber:
I signed it!

jhaber:
Hmm I'm the author of all the commits and the pull request submitter, I'm not sure what it's referring to

xfxyjwf:
ok to test

xfxyjwf:
Not sure what tripped the CLA bot... Did you use different email addresses in the commits? The bot is not very smart and often gets confused when multiple email addresses are used in different places.

Anyway, I manually flipped the CLA flag.

jhaber:
I didn't change anything between commits, but I initially signed the CLA about 3 years ago and a bunch of stuff may have changed since then so I'm guessing that's what confused it

jhaber:
I merged master since it seems like some conflicts had cropped up

jhaber:
I just wanted to make sure this PR didn't fall through the cracks. I think the current behavior is pretty clearly a bug, there isn't always an easy workaround available, and the user ends up with a vexing `ClassNotFoundException` when they hit this issue

jhaber:
I'm curious why this was labeled obsolete and closed? As far as I can tell #3474 is still a problem

xfxyjwf:
Sorry, we are doing a batch cleanup and closing issues that haven't been updated for a while or are not actionable for us. If you are still interested in getting it in, could you help update and resent the change? Feel free to ping me if you don't get a response in time. We have only few engineers working on this repo and are constantly losing track of things...

PR #3480, opened 2017-08-10
Clean up typedefs for Atomic32/Atomic64

bklarson:
The typedefs for Atomic32 and Atomic64 were sometimes causing
Atomic32 to be an int32 rather than an intptr_t on 32-bit platforms.
On some of these platforms (ARM/CortexM building with GCC 6 in one
case) int32 is a long int, while intptr_t is an int, which causes a
compiler error even though long int and int are both 4 bytes.

Having Atomic32 always be intptr_t on 32-bit platforms and Atomic64
always be intptr_t on 64-bit platforms should resolve any of these
types of errors.

### DIFF
--- a/src/google/protobuf/stubs/atomicops.h
+++ b/src/google/protobuf/stubs/atomicops.h
@@ -63,28 +63,21 @@ namespace google {
 namespace protobuf {
 namespace internal {
 
-#if defined(GOOGLE_PROTOBUF_ARCH_POWER)
-#if defined(_LP64) || defined(__LP64__)
-typedef int32 Atomic32;
-typedef intptr_t Atomic64;
+#ifdef GOOGLE_PROTOBUF_ARCH_32_BIT
+  typedef intptr_t Atomic32;
+  typedef int64 Atomic64;
 #else
-typedef intptr_t Atomic32;
-typedef int64 Atomic64;
-#endif
-#else
-typedef int32 Atomic32;
-#ifdef GOOGLE_PROTOBUF_ARCH_64_BIT
-// We need to be able to go between Atomic64 and AtomicWord implicitly.  This
-// means Atomic64 and AtomicWord should be the same type on 64-bit.
-#if defined(__ILP32__) || defined(GOOGLE_PROTOBUF_OS_NACL)
-// NaCl's intptr_t is not actually 64-bits on 64-bit!
-// http://code.google.com/p/nativeclient/issues/detail?id=1162
-// sparcv9's pointer type is 32bits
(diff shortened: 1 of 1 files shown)


bazel-io:

bklarson:
This fixes my build error in GCC 6 targeting ARM platforms, which was discussed at https://groups.google.com/forum/#!topic/protobuf/IOzjwQmlL0k.  It seems reasonable to me that it would continue to work for all other platforms and is a bit cleaner / easier to read.  However I don't have access to test on most of the platforms supported by GPB and I'm not sure how much of that coverage we get from the build bot.

src/google/protobuf/stubs/atomicops.h
```diff
@@ -63,28 +63,21 @@ namespace google {
 namespace protobuf {
 namespace internal {
 
-#if defined(GOOGLE_PROTOBUF_ARCH_POWER)
-#if defined(_LP64) || defined(__LP64__)
-typedef int32 Atomic32;
-typedef intptr_t Atomic64;
+#if defined(GOOGLE_PROTOBUF_ARCH_32_BIT) && GOOGLE_PROTOBUF_ARCH_32_BIT
```
xfxyjwf: Simply "#ifdef GOOGLE_PROTOBUF_ARCH_32_BIT"? We don't actually use the macro's value.

src/google/protobuf/stubs/atomicops.h
```diff
@@ -63,28 +63,21 @@ namespace google {
...
+    typedef intptr_t Atomic32;
```
xfxyjwf: Indent by 2 spaces.

src/google/protobuf/stubs/atomicops.h
bklarson: Done.

src/google/protobuf/stubs/atomicops.h
bklarson: Done.

pherl:
This actually breaks Mac on 32bit platform where we fixed the issue by defining AtomicWord APIs directly and delegate them to Atomic32 (https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/atomicops_internals_atomicword_compat.h). Changing the Atomic32 to be intptr_t (AtomicWord) caused the redefinition for 32bit mac.

It seems to me removing the atomicword_compat.h is the cleanest solution. Though the comments in that file suggests Atomic32 should be int32, I guess we are already using intptr_t for Atomic64, it's fine to use intptr_t for Atomic32 as well.

PR #3485, opened 2017-08-11
Fix build on MinGW32

pherl:
Restrict the io_win32 change and only enable that for MSVC, because:
1) lots of the _w functions are not available on MinGW32 (possibly on cygwin either)
2) those platforms include <io.h> already via other common headers; using the functions in internal::win32 creates conflicts.

This is a temporary fix to make the build pass. We can revisit this if support on MinGW32 is required. Note that the pre-built protoc artifacts are built under mingw32. With this change, they won't use the long path support code and would rely on the platform APIs for path support.

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -108,7 +108,7 @@ namespace compiler {
 #endif
 
 namespace {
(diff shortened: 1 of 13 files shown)


pherl:
This fixes https://github.com/google/protobuf/issues/3483
@laszlocsomor

PR #3486, opened 2017-08-11
Add any support in php runtime.

TeBoring:


bazel-io:

michaelbausor:
A question about the approach here: is it necessary to implement a version of Any in pure c? I ask because this is a lot of effort to do for all of the protobuf well known types, when it is relatively easy just to generate them from the proto compiler.

By not implementing in c, the downsides I can see are:
- You need the PHP composer package to get the well known types. This doesn't seem like too much of a downside to me, so long as the composer package is side-by-side compatible with the c extension.
- It will be slightly slower. But, I think that this will be negligible since the base message class is implemented in c, so the PHP class is a thin layer on top.

What do you think? I only did a very quick scan, but it seems like other languages don't implement the well known types in c code.

TeBoring:
In that way, we will need another repository on packgist for well known type, right?

michaelbausor:
No - they will be in the google/protobuf package. So users who need well known types will always need this package, even if they use the c extension.

TeBoring:
I think add well know type in c is a one time burden. By having that, we don't need to require php implementation.

michaelbausor:
That is true, although I assume there will be some ongoing maintenance costs (e.g. if there are changes in the code generator similar to adding fluent setters, they would need to be added to the c implementation for all of the well known types) and the disadvantage of requiring the php implementation seems fairly low.

But, I just wanted to suggest this as an idea, not a requirement.

TeBoring:
That's a good point. 
However, I have some small concern about it:
1) In that way, we need to use public API to implement well known types. There may be a situation that we only need one function to implement well known type and we are forced to make it public.
2) The current framework to test c extension needs to be changed. Have you tried to test c extension where php implementation is also available.

php/ext/google/protobuf/protobuf.h
```diff
@@ -292,6 +352,17 @@ static inline int php_proto_zend_hash_get_current_data_ex(HashTable* ht,
   zend_object std;                \
   };
 
+#define PHP_PROTO_INIT_SUBMSGCLASS_START(CLASSNAME, CAMELNAME, LOWWERNAME)   \
```
michaelbausor: Is this supposed to be redefined? (it is on line 119)

TeBoring: This has different definitions for php5 and php7.

php/ext/google/protobuf/message.c
```diff
@@ -342,3 +343,322 @@ PHP_METHOD(Message, whichOneof) {
...
+    return;
+  }
+
+  zval member;
+  zval value;
+  PHP_PROTO_ZVAL_STRING(&member, "type_url", 1);
+  PHP_PROTO_ZVAL_STRINGL(&value, type_url, type_url_len, 1);
+
+  zend_class_entry* old_scope = EG(scope);
+  EG(scope) = any_type;
```
stanley-cheung: Talked offline about this. Bo may need to update some code around this.

php/ext/google/protobuf/message.c
```diff
@@ -342,3 +343,322 @@ PHP_METHOD(Message, whichOneof) {
...
+  message_handlers->write_property(getThis(), &member, &data,
+                                   NULL PHP_PROTO_TSRMLS_CC);
+  EG(scope) = old_scope;
+
+  // Set type url.
+  Descriptor* desc =
+      UNBOX_HASHTABLE_VALUE(Descriptor, get_ce_obj(Z_OBJCE_P(val)));
+  const char* fully_qualified_name = upb_msgdef_fullname(desc->msgdef);
+  const char* url_prefix = "type.googleapis.com/";
```
stanley-cheung: Put this in a constant?

php/ext/google/protobuf/message.c
```diff
@@ -342,3 +343,322 @@ PHP_METHOD(Message, whichOneof) {
...
+  ZVAL_OBJ(return_value, klass->create_object(klass TSRMLS_CC));
+  MessageHeader* msg = UNBOX(MessageHeader, return_value);
+  custom_data_init(klass, msg PHP_PROTO_TSRMLS_CC);
+
+  // Get value.
+  zval value_member;
+  PHP_PROTO_ZVAL_STRING(&value_member, "value", 1);
+  old_scope = EG(scope);
+  EG(scope) = any_type;
+#if PHP_MAJOR_VERSION < 7
```
stanley-cheung: This pattern happens quite a lot. Perhaps worth it to put a macro in `protobuf.h`?

php/ext/google/protobuf/message.c
```diff
@@ -342,3 +343,322 @@ PHP_METHOD(Message, whichOneof) {
...
+  PHP_PROTO_ZVAL_STRING(&member, "value", 1);
+  PHP_PROTO_ZVAL_STRINGL(&value_to_set, value, value_len, 1);
+
+  zend_class_entry* old_scope = EG(scope);
+  EG(scope) = any_type;
+  message_handlers->write_property(getThis(), &member, &value_to_set,
+                                   NULL PHP_PROTO_TSRMLS_CC);
+  EG(scope) = old_scope;
+
+#if PHP_MAJOR_VERSION < 7
```
stanley-cheung: This pattern happens quite a lot. Perhaps worth it to put a macro in `protobuf.h`?

TeBoring:
All comments fixed.

PR #3496, opened 2017-08-15
Static link pthread for mingw64 protoc and enable cross compile

pherl:
@zhangkun83


zhangkun83:
Is mingw64 instead of cygwin64 used for building 64-bit binaries now? If it's the case, could you also update README?

pherl:
I think all linux, cygwin64, mingw64 support building the artifacts now. Updated the readme.

PR #3503, opened 2017-08-16
Work around a bug in clang's static analyzer

gburgessiv:
Due to https://bugs.llvm.org/show_bug.cgi?id=34198, clang's static
analyzer emits diagnostics about leaking `container`. Doing this
assignment in two steps works around this.

In particular, this code is making the static analyzer we run in The
Wonderful World of Android sad.


bazel-io:

gburgessiv:
Looks like github chose my not-gmail to make this commit with. Crossing my fingers that the bot doesn't care...

gburgessiv:
Ping :)

pherl:
ok to test

gburgessiv:
(Added the bug link to the comments in the patch)

gburgessiv:
Thank you!

Issue #40, opened 2014-10-02
Allow repeated custom option, or disallow it as in <=2.5


protobufel:
The recent change in 2.6.0 to kind of allow for repeated custom options, as seen in google/protobuf/unittest_custom_options.proto, feels and is, IMHO, as a hack for allowing repeated fields in custom options. 
1. This is inconsistent with the "inlined" field options, that don't allow for such repetition - protoc throws a parsing error.
2. This is confusing and requires a relax, "strange" handling - what before 2.6.0  was a single field now becomes a repeated field after options' extension resolution, aka re-parsing.
3. In most cases, such repetition is really an unintentional user error.
4. If repeated fields in options to be supported, then there are, as usual, two separate cases to handle - singular, and repeated fields. 

Therefore, the right thing to do is either disallow repeated option fields as before, or allow repeated LABEL and option LABEL along with it.

In case of both allowed, re-parsing of the FileDescriptorProto in FileDescriptor.buildFrom, should result in either repeated, or optional field/extension set on options.   

Here is the google/protobuf/unittest_custom_options.proto excerpt: 

// Note that we try various different ways of naming the same extension.
message VariousComplexOptions {
  option (.protobuf_unittest.complex_opt1).foo = 42;
  option (protobuf_unittest.complex_opt1).(.protobuf_unittest.quux) = 324;
  option (.protobuf_unittest.complex_opt1).(protobuf_unittest.corge).qux = 876;
  option (protobuf_unittest.complex_opt1).foo4 = 99;
  option (protobuf_unittest.complex_opt1).foo4 = 88;
  option (complex_opt2).baz = 987;
  option (complex_opt2).(grault) = 654;
  option (complex_opt2).bar.foo = 743;
  option (complex_opt2).bar.(quux) = 1999;
  option (complex_opt2).bar.(protobuf_unittest.corge).qux = 2008;
  option (complex_opt2).(garply).foo = 741;
  option (complex_opt2).(garply).(.protobuf_unittest.quux) = 1998;
  option (complex_opt2).(protobuf_unittest.garply).(corge).qux = 2121;
  option (ComplexOptionType2.ComplexOptionType4.complex_opt4).waldo = 1971;
  option (complex_opt2).fred.waldo = 321;
  option (complex_opt2).barney = { waldo: 101 };
  option (complex_opt2).barney = { waldo: 212 };
  option (protobuf_unittest.complex_opt3).qux = 9;
[...]

PR #3506, opened 2017-08-17
Updated for supporting arch ppc64le

anuphalarnkar:
The files namely "build-protoc.sh" and "build-zip.sh" in sub-folder "protoc-artifacts" has been updated to support architecture PPC64LE.
After this modification, I am able to successfully build for Power platform.

### DIFF
--- a/protoc-artifacts/build-protoc.sh
+++ b/protoc-artifacts/build-protoc.sh
@@ -73,6 +73,8 @@ checkArch ()
         assertEq $format "elf32-i386" $LINENO
       elif [[ "$ARCH" == x86_64 ]]; then
         assertEq $format "elf64-x86-64" $LINENO
+      elif [[ "$ARCH" == ppcle_64 ]]; then
+        assertEq $format "elf64-powerpcle" $LINENO
       else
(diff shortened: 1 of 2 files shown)


bazel-io:

pherl:
Which platform did you test the ppc build? ppc or linux cross compile?

anuphalarnkar:
The patch is for following system:
Arch: ppc64le
OS: Ubuntu 14.04.1

uname -a
Linux xxxxxxx 3.16.0-30-generic #40~14.04.1-Ubuntu SMP Thu Jan 15 17:42:36 UTC 2015 ppc64le ppc64le ppc64le GNU/Linux

anuphalarnkar:
Also, we need to update "os-maven-plugin" version from "1.2.3.Final" to "1.5.0.Final" as arch "ppc64_le" tag is not present in earlier version of the plugin. I did not modify this as I am not sure how this will affect x86 builds

pherl:
Thanks for the change. Could you try if there's cross compile available? We don't have ppc workstations, and would have to rely on cross compiling other platforms (e.g. windows) in linux.

PR #3507, opened 2017-08-17
import inspect

cclauss:
Inspect is used five times on line 65-74 without being imported.


bazel-io:

PR #3508, opened 2017-08-17
Python 2.6 or 2.7

cclauss:
Python 3 is _not_ supported and there will never be a Python 2.8 https://www.python.org/dev/peps/pep-0404/

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -30,7 +30,7 @@ join the Protocol Buffers discussion list and let us know!
 Installation
 ============
 
-1) Make sure you have Python 2.6 or newer.  If in doubt, run:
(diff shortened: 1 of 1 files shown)


bazel-io:

xfxyjwf:
Python 3 is not supported? Where did you get that?

cclauss:
@xfxyjwf https://travis-ci.org/cclauss/itinerant-tester/jobs/265518973#L487 is what convinced my that Python 3 was not yet supported.

xfxyjwf:
That's a failing test we need to fix... not that we need to drop python 3 support because of a failed unittest.

xfxyjwf:
Does your other change to add python 3.4, 3.5, 3.6 trigger the test failure on our travis test?

PR #3512, opened 2017-08-17
Drop Python 3.3 from testing & add Python 3.5, 3.6

cclauss:
Python 3.3 goes [EOL](https://docs.python.org/devguide/index.html#branchstatus) next month.

### DIFF
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{26,27,33,34}-{cpp,python}
+    py{26,27,34,35,36}-{cpp,python}
 
 [testenv]
 usedevelop=true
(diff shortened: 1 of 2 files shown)


bazel-io:

xfxyjwf:
FYI. I have canceled all travis tests other than the python test so the test can finish sooner.

anandolee:
protobuf/tests.sh

bazel-io:

anandolee:
ok to test

cclauss:
@anandolee What went wrong when the original PR was applied?

anandolee:
I forgot to trigger Jenkins tests so it need to be rollback and run again.
No worries, I will add it back once tests pass


And also we do not want to drop 3.3 right now. Python 2.6 which is alrady
EOL is in the tests as well

cclauss:
OK.  Thanks for the explanation.  Understood on retaining old versions, Py2.6 has been EOL for 1,390 days now I think you can let it go sometime soon but we will leave it in place.

PR #3516, opened 2017-08-17
Python 3 compatibility fixes: print(), long(), etc.

cclauss:
Define raw_input() for Python 3, replace file() with open(), etc. #3507, #3508, and #3512 together should fix most of the Python 3 incompatibilities.


bazel-io:

PR #3517, opened 2017-08-18
Define cmp() for Python 3

cclauss:
http://python-future.org/compatible_idioms.html?highlight=cmp#cmp


bazel-io:

PR #3527, opened 2017-08-18
Remove cmp for python3

anandolee:
It is a revert for #3517 because Jenknins show syntax errors:
https://grpc-testing.appspot.com/job/protobuf_pull_request/1479/testReport/junit/(root)/python/python/


bazel-io:

PR #3534, opened 2017-08-21
Fixing unused parameter warning under Clang.

drivehappy:
Marks the value parameter as unused.


bazel-io:

Issue #3533, opened 2017-08-21
3.4.0 not building against VS9 (requires c++11 features)


hajapy:
Over at conda-forge we noticed that builds of protobuf 3.4.0 are failing with VS9 (see https://github.com/conda-forge/protobuf-feedstock/pull/31) because of use of c++11 features. Here is the error: 

```
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(206) : error C2039: 'back' : is not a member of 'std::basic_string<_Elem,_Traits,_Ax>'
        with
        [
            _Elem=char,
            _Traits=std::char_traits<char>,
            _Ax=std::allocator<char>
        ]
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(212) : error C2039: 'unique_ptr' : is not a member of 'std'
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(212) : error C2143: syntax error : missing ';' before '<'
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(212) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(212) : error C2143: syntax error : missing ';' before '{'
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(212) : error C2447: '{' : missing function header (old-style formal list?)
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(221) : error C2039: 'unique_ptr' : is not a member of 'std'
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(221) : error C2059: syntax error : ']'
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(222) : error C2065: 'wbuf' : undeclared identifier
C:\bld\protobuf_1503290160041\work\protobuf-3.4.0\src\google\protobuf\stubs\io_win32.cc(222) : error C2228: left of '.get' must have class/struct/union
        type is ''unknown-type''
[...]

PR #3536, opened 2017-08-21
Remove C++11 features in io_win32.cc

pherl:
- Use scoped_array instead of unique_ptr.
- Do not use string::front() and string::back()


pherl:
Fixes https://github.com/google/protobuf/issues/3533

jrandalldemllo:
Getting VS10 build error for v3.4.1
Line 198 contains range based for loop syntax. Posted build error log in #3533

PR #3538, opened 2017-08-21
Remove unreferenced 'GPBMessageSignatureProtocol' class.

dmaclach:
Remove unreferenced 'GPBMessageSignatureProtocol' class that is just taking up
space in the Objective C runtime information by hanging the protocol it needs to
declare off of GPBRootObject instead.

Small binary size reduction, but more importantly a removal of a class that
appears to be unused when statically analyzed.


thomasvl:
ok to test

objectivec/GPBUtilities.m
```diff
@@ -1216,8 +1216,13 @@ void GPBSetMessageMapField(GPBMessage *self, GPBFieldDescriptor *field,
 const char *GPBMessageEncodingForSelector(SEL selector, BOOL instanceSel) {
   Protocol *protocol =
       objc_getProtocol(GPBStringifySymbol(GPBMessageSignatureProtocol));
+  NSCAssert(protocol, @"Missing GPBMessageSignatureProtocol");
   struct objc_method_description description =
       protocol_getMethodDescription(protocol, selector, NO, instanceSel);
+  NSCAssert(description.name != Nil,
```
thomasvl: Why both of these vs. just checking for a nil result?

thomasvl: Oops, returns a struct, how about a single assert.

PR #3560, opened 2017-08-24
Move parse frame array to the Map object

### LINKED ISSUE #3250: Segfault when ruby-protobuf decodes lots of messages with maps concurrently
When looking into a grpc-ruby crash described in https://github.com/grpc/grpc/issues/11331, it looks like the repro could be trimmed down to include only the ruby `google/protobuf` library.

What seems necessary to trigger the crash is:
* large or lots of messages (garbage collector pressure needed?).
* `map` types with <i>any</i> key type and a value of a <i>Message</i> type.
* trying to `decode` these messages
* at least some concurrency (it seems to need at least two threads to trigger).

The minimal repro seems to be:

`repro.proto`:

```
syntax = "proto3";

package repro;

message Outer {
        map<int32,Inner> items = 1;
}

message Inner {}
```

In the same directory, the `repro.rb` file looks like:

```
require_relative 'repro_pb'

thds = []

2.times do
  thds << Thread.new do
    o = Repro::Outer.new
    o.items[0] = Repro::Inner.new
    raw = Repro::Outer.encode(o)

    100000.times do
      Repro::Outer.decode(raw)
    end
  end
end

thds.each(&:join)
```

`repro_pb.rb` was generated with:

```
protoc -I. --ruby_out=. repro.proto
```

see [generated repro_pb.rb](https://gist.github.com/apolcyn/f7f68be8a54450637f8cb9769fc83abf)

After generating, running `ruby repro.rb` should segfault during a `decode` call (I'm seeing it repro about 9/10 tries - increasing the number of calls to `Repro::Outer.decode` seems to cause it more frequently):

[example stack trace](https://gist.github.com/apolcyn/cca44cbb9d38a77183d8b2ee76733974)

Environment details:

`protoc --version` gives `libprotoc 3.3.0` 
(protoc built from head of grpc repo with `make plugins`, and used from the `<grpc_repo>/bins/opt/protobuf/protoc`)

`gem list google-protobuf` gives `google-protobuf (3.3.0 x86_64-linux)`

ruby platform: `x86_64-linux`
ruby version: 2.4.0

tenderlove:
This makes the frame stack per-parser, and per-thread.  Fixes #3250

### DIFF
--- a/ruby/ext/google/protobuf_c/encode_decode.c
+++ b/ruby/ext/google/protobuf_c/encode_decode.c
@@ -280,11 +280,6 @@ rb_data_type_t MapParseFrame_type = {
   { MapParseFrame_mark, MapParseFrame_free, NULL },
 };
 
-// Array of Ruby objects wrapping map_parse_frame_t.
-// We don't allow multiple concurrent decodes, so we assume that this global
-// variable is specific to the "current" decode.
-VALUE map_parse_frames;
-
 static map_parse_frame_t* map_push_frame(VALUE map,
                                          const map_handlerdata_t* handlerdata) {
   map_parse_frame_t* frame = ALLOC(map_parse_frame_t);
@@ -293,14 +288,10 @@ static map_parse_frame_t* map_push_frame(VALUE map,
   native_slot_init(handlerdata->key_field_type, &frame->key_storage);
   native_slot_init(handlerdata->value_field_type, &frame->value_storage);
 
-  rb_ary_push(map_parse_frames,
+  Map_push_frame(map,
               TypedData_Wrap_Struct(rb_cObject, &MapParseFrame_type, frame));
 
   return frame;
-}
-
-static void map_pop_frame() {
-  rb_ary_pop(map_parse_frames);
 }
 
 // Handler to begin a map entry: allocates a temporary frame. This is the
@@ -336,7 +327,7 @@ static bool endmap_handler(void *closure, const void *hd, upb_status* s) {
       &frame->value_storage);
 
   Map_index_set(frame->map, key, value);
-  map_pop_frame();
+  Map_pop_frame(frame->map);
 
   return true;
 }
@@ -775,10 +766,6 @@ VALUE Message_decode(VALUE klass, VALUE data) {
   msg_rb = rb_class_new_instance(0, NULL, msgklass);
   TypedData_Get_Struct(msg_rb, MessageHeader, &Message_type, msg);
 
-  // We generally expect this to be clear already, but clear it in case parsing
-  // previously got interrupted somehow.
-  rb_ary_clear(map_parse_frames);
-
   {
     const upb_pbdecodermethod* method = msgdef_decodermethod(desc);
     const upb_handlers* h = upb_pbdecodermethod_desthandlers(method);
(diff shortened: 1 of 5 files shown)


bazel-io:

tenderlove:
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

I signed it!

pherl:
OK to test

bazel-io:

pherl:
ok to test

tenderlove:
I see that some builds failed, but I can't see the details.  Do I need to make sure that all of them are green before this can be applied?

pherl:
I believe it's just flaky. The tests result look good. @haberman could you take a look at the PR?

haberman:
Thanks very much for the fix! Can you help me understand what could cause concurrent decodes though? I know Ruby can have multiple threads, but I was under the impression that Ruby has a GVL that prevents Ruby code or C extensions from running concurrently (unless you use `rb_thread_call_without_gvl`): https://github.com/ruby/ruby/blob/trunk/doc/extension.rdoc#threading

tenderlove:
> Thanks very much for the fix! Can you help me understand what could cause concurrent decodes though?

Ruby can switch threads when the GC is invoked.  Array manipulations with `rb_ary_push` / `rb_ary_pop` may cause the GC to be invoked and allow threads to switch.  If the array was implemented with a non-Ruby data structure, then the VM would not be given an opportunity to switch threads.

> I know Ruby can have multiple threads, but I was under the impression that Ruby has a GVL that prevents Ruby code or C extensions from running concurrently

It sounds like you might be confusing concurrency and parallelism.  The GVL prevents parallelism (two Ruby threads running **at the same time**), but it does not prevent concurrency (two Ruby threads running **interleaved**).

The bug that we are hitting here is that we have two decodes running *interleaved*.  Interleaving thread execution happened because the calls to `rb_ary_*` allowed a different thread to be scheduled.  These interleaved executions both mutated a shared array, and that leads to a SEGV or ["strange values for a key in the ruby object"](https://github.com/google/protobuf/issues/3250#issuecomment-314696014).

I hope this helps.  If there's anything I can do to better explain, please let me know!

NOTE: It may not have been the `rb_ary_*` functions that allowed the thread to switch, they are just the ones that stuck out to me.  Many `rb_*` calls allow threads to switch (especially if they hit the GC or do IO operations).

ruby/ext/google/protobuf_c/encode_decode.c
```diff
@@ -293,16 +288,12 @@ static map_parse_frame_t* map_push_frame(VALUE map,
   native_slot_init(handlerdata->key_field_type, &frame->key_storage);
   native_slot_init(handlerdata->value_field_type, &frame->value_storage);
 
-  rb_ary_push(map_parse_frames,
+  Map_push_frame(map,
```
haberman: If you are storing this in the map directly, I don't think it needs to be an array with push and pop. I think it could just be a single value. The array was for when we had a single global stack that needed to be able to handle parsing of nested maps.

haberman:
Got it, thanks for the explanation. I hadn't realized that calls into Ruby APIs could release the GVL; I thought every call into a C extension routine was atomic unless you explicitly released the GVL.

ruby/ext/google/protobuf_c/encode_decode.c
tenderlove: Ok, I changed this to store only one value.

PR #3563, opened 2017-08-24
Correct PHP JSON_BIGINT_AS_STRING parameter and decode as an object

jackwakefield:
This fixes the placement of the `JSON_BIGINT_AS_STRING` option passed to `json_decode`, this is being passed as the `$assoc` parameter which decodes the JSON as an associative array.

I've changed the JSON to be decoded as as object rather as an array. This fixes an issue where the keys to a Protobuf map could be a numerical string, which then become integers when decoded as an associative array causing `expected string`

```php
$data = '{"1001": true, "foo": true}';

var_dump(json_decode($data, true));
// array(2) {
//   [1001]=>
//   bool(true)
//   ["foo"]=>
//   bool(true)
// }

var_dump(json_decode($data));
// object(stdClass)#1 (2) {
//   ["1001"]=>
//   bool(true)
//   ["foo"]=>
//   bool(true)
// }
```

### DIFF
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -718,7 +718,7 @@ class Message
                         throw new GPBDecodeException("Expect message.");
                     }
 
-                    $submsg->mergeFromJsonArray($value);
+                    $submsg->mergeFromJsonObject($value);
                 }
                 return $submsg;
             case GPBType::ENUM:
@@ -842,9 +842,9 @@ class Message
         }
     }
 
-    private function mergeFromJsonArray($array)
+    private function mergeFromJsonObject($object)
     {
-        foreach ($array as $key => $value) {
+        foreach ($object as $key => $value) {
             $field = $this->desc->getFieldByJsonName($key);
             if (is_null($field)) {
                 $field = $this->desc->getFieldByName($key);
@@ -913,13 +913,13 @@ class Message
      */
     public function parseFromJsonStream($input)
     {
-        $array = json_decode($input->getData(), JSON_BIGINT_AS_STRING);
-        if (is_null($array)) {
+        $object = json_decode($input->getData(), false, 512, JSON_BIGINT_AS_STRING);
+        if (is_null($object)) {
             throw new GPBDecodeException(
                 "Cannot decode json string.");
         }
         try {
-            $this->mergeFromJsonArray($array);
+            $this->mergeFromJsonObject($object);
         } catch (Exception $e) {
             throw new GPBDecodeException($e->getMessage());
         }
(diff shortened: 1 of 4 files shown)


bazel-io:

jackwakefield:
I signed it!

TeBoring:
ok to test

TeBoring:
Could you add a test to prevent regression?

jackwakefield:
I've added a simple test which will throw an exception when the JSON is being decoded as an associative array.

I've also changed the test script to run the tests against both the extension and PHP implementation as this was only reproducible with the latter. The PHP implementation tests are currently failing in `well_known_test.php` with the following, which I'll open an issue for.

```
Use of undefined constant url_prifix_len - assumed 'url_prifix_len'

php/src/Google/Protobuf/Any.php:212
php/tests/well_known_test.php:46
```

bazel-io:

TeBoring:
Could you rebase this change?

jackwakefield:
Done

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -850,9 +850,9 @@ private function convertJsonValueToProtoValue(
         }
     }
 
-    private function mergeFromJsonArray($array)
+    private function mergeFromJsonObject($object)
```
TeBoring: Is this rename intentional?

php/tests/test.sh
```diff
@@ -12,11 +12,17 @@ tests=( array_test.php encode_decode_test.php generated_class_test.php generated
...
-  echo "****************************"
-  echo "* $t"
-  echo "****************************"
+  echo "************************************"
+  echo "* $t - Extension"
+  echo "************************************"
   php -dextension=../ext/google/protobuf/modules/protobuf.so `which phpunit` --bootstrap autoload.php $t
   echo ""
+
+  echo "************************************"
```
TeBoring: This is duplicate of what we do to test php implementation.

php/src/Google/Protobuf/Internal/Message.php
TeBoring: The original mergeFromJsonArray is for implementation only. What's the benefit to use object instead of array?

jackwakefield: When the field is a map PHPs type system will convert a key with string `"1"` to the number `1` when decoding as an associative array, which will cause an exception with Protobuf's type checking. Decoding as an object preserving the key types.

php/tests/test.sh
jackwakefield: Could you point me in the right direction to run the existing tests? I wasn't able to get the tests to run against the PHP implementation (and the test to fail) without adding this.

TeBoring: cd php
composer install
./vendor/bin/phpunit

PR #3564, opened 2017-08-24
Add well-known timestamps to JSON for PHP

jackwakefield:
This adds PHP support for JSON serialization and deserialization of `google.protobuf.Timestamp` RFC 3339 timestamps.

I appreciate there may be a better way of doing this, especially to accommodate the other well-known types in the future.


bazel-io:

TeBoring:
Could you verify the conformance test for Timestamp has passed?
https://github.com/google/protobuf/blob/master/conformance/conformance_test.cc#L2040

TeBoring:
Here is the script to run conformance test: https://github.com/google/protobuf/blob/master/tests.sh#L400

TeBoring:
ok to test

jackwakefield:
Thanks for the heads up @TeBoring, there were a number of failures but the conformance tests are now passing.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -1260,7 +1273,13 @@ private function fieldDataOnlyJsonByteSize($field, $value)
                 $size += 2;  // size for \"\"
                 break;
             case GPBType::MESSAGE:
-                $size += $value->jsonByteSize();
+                if ($field->isTimestamp()) {
```
TeBoring: Can we merge this logic to jsonByteSize?

php/src/Google/Protobuf/Internal/GPBWellKnown.php
```diff
@@ -0,0 +1,99 @@
+<?php
```
TeBoring: Move this to GPBUtil?

TeBoring:
Broken travis test is for ruby. Should be unrelated.

php/src/Google/Protobuf/Internal/Message.php
jackwakefield: That's done, I've made the same change with `serializeToJsonStream`

php/src/Google/Protobuf/Internal/GPBWellKnown.php
jackwakefield: Done

jackwakefield:
Sorry, bad rebase. Those have now been fixed.

TeBoring:
Thanks for the help.

PR #3565, opened 2017-08-24
Add mising cmake files in dist

### LINKED ISSUE #3562: CMake 3.4.0 builds broken on Linux due to missing files in release package
For some reason, the "cpp" tar/zip releases for 3.4.0 are missing two key files:

* cmake/protobuf-lite.pc.cmake
* cmake/protobuf.pc.cmake

The absence of these files breaks the cmake build on Linux, as they are referred to from cmake/install.cmake.

These files are present in the full source code link, just not the cpp one. See for comparison:

* https://github.com/google/protobuf/releases/download/v3.4.0/protobuf-cpp-3.4.0.tar.gz
* https://github.com/google/protobuf/archive/v3.4.0.tar.gz

pherl:
Fixes #3562

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -982,8 +982,10 @@ EXTRA_DIST = $(@DIST_LANG@_EXTRA_DIST)   \
   cmake/libprotoc.cmake                  \
   cmake/protobuf-config-version.cmake.in \
   cmake/protobuf-config.cmake.in         \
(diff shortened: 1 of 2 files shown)


pherl:
introduced in #3403

NelsonBilber:
It's fixed.

PR #3576, opened 2017-08-30
change the field number of php_generic_service to fix the conflict with internal descriptor.proto

TeBoring:


bazel-io:

PR #3578, opened 2017-08-31
Add @Deprecated annotation support for proto file.

pherl:


src/google/protobuf/compiler/java/java_file.cc
```diff
@@ -248,9 +248,12 @@ void FileGenerator::Generate(io::Printer* printer) {
   PrintGeneratedAnnotation(
       printer, '$', options_.annotate_code ? classname_ + ".java.pb.meta" : "");
   printer->Print(
-      "public final class $classname$ {\n"
+      "$deprecation$public final class $classname$ {\n"
```
xfxyjwf: Is there a test to check this annotation?

pherl: I only added a proto to make sure it compiles. Manually checked the annotations were generated though.

I can add a test to verify those annotations in a later PR (when I get time...)

java/core/src/test/proto/com/google/protobuf/deprecated_file.proto
```diff
@@ -0,0 +1,35 @@
...
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto3";
+
+package deprecated_file;
+
```
xfxyjwf: Can you add a TODO in this file (that we need a test to check the annotation)?

pherl: Done.

PR #3583, opened 2017-09-01
Add native php support for Duration.

TeBoring:


bazel-io:

PR #3591, opened 2017-09-04
Update port.h

NanXiao:
Fix following build errors:  

```
error: this use of "defined" may not be portable [-Werror=expansion-to-defined]
 #if GOOGLE_PROTOBUF_USE_UNALIGNED
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/google/protobuf/stubs/port.h
+++ b/src/google/protobuf/stubs/port.h
@@ -263,7 +263,7 @@ static const uint64 kuint64max = GOOGLE_ULONGLONG(0xFFFFFFFFFFFFFFFF);
 # endif
 #endif
 
-#if GOOGLE_PROTOBUF_USE_UNALIGNED
+#ifdef GOOGLE_PROTOBUF_USE_UNALIGNED
 
 #define GOOGLE_UNALIGNED_LOAD16(_p) (*reinterpret_cast<const uint16 *>(_p))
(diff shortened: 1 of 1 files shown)


bazel-io:

src/google/protobuf/stubs/port.h
```diff
@@ -263,7 +263,7 @@ static const uint64 kuint64max = GOOGLE_ULONGLONG(0xFFFFFFFFFFFFFFFF);
 # endif
 #endif
 
-#if GOOGLE_PROTOBUF_USE_UNALIGNED
+#ifdef GOOGLE_PROTOBUF_USE_UNALIGNED
```
acozzette: I think this would actually change the behavior. When we get here this macro should have been assigned a value of either 0 or 1, so we need to check the value of the macro and not whether it was defined. Does the warning disappear if you do something like this instead?
`#if defined(GOOGLE_PROTOBUF_USE_UNALIGNED) && GOOGLE_PROTOBUF_USE_UNALIGNED`

xfxyjwf:
The reported warning is already fixed by https://github.com/google/protobuf/commit/a23e198ccaf53fededaa09b5988afece897fc3ac

NanXiao:
@xfxyjwf @acozzette Yes, this warning is already fixed, sorry for the inconvenience!

PR #3594, opened 2017-09-05
bazel: Add proto_library rules for well known types. Fixes #2763

### LINKED ISSUE #2763: Add a proto_library rule for well-known types and update the proto_lang_toolchain rules
Bazel now has built-in rules for building Java and C++ protos (`proto_libarry`, `java_proto_library`, `java_lite_proto_library` and `cc_proto_library`).

For users to import well-known types (e.g., `descriptor.proto`) from their `.proto` files, they must depend on a `proto_library` that `srcs` the former.

This FR is about adding such a `proto_library`.

https://github.com/google/protobuf/issues/2598 is a blocker.

buchgr:
Adds a proto_library rule for each well known type proto:

```
$ bazel query "filter(\".*_proto$\", \"...\")"
//:wrappers_proto
//:timestamp_proto
//:struct_proto
//:field_mask_proto
//:empty_proto
//:duration_proto
//:compiler_plugin_proto
//:descriptor_proto
//:api_proto
//:type_proto
//:source_context_proto
//:any_proto
```

Bazel users can reference these proto_library rules for their own
language specific rules i.e.

```
java_proto_library(
  name = "any_java_proto",
  deps = ["@com_google_protobuf//:any_proto"],
)
```

Also set the workspace name to `"com_google_protobuf"`, as proto_library
rules reference protobuf that way.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -224,21 +224,24 @@ objc_library(
     visibility = ["//visibility:public"],
 )
 
-RELATIVE_WELL_KNOWN_PROTOS = [
-    # AUTOGEN(well_known_protos)
-    "google/protobuf/any.proto",
-    "google/protobuf/api.proto",
-    "google/protobuf/compiler/plugin.proto",
-    "google/protobuf/descriptor.proto",
-    "google/protobuf/duration.proto",
-    "google/protobuf/empty.proto",
-    "google/protobuf/field_mask.proto",
-    "google/protobuf/source_context.proto",
-    "google/protobuf/struct.proto",
-    "google/protobuf/timestamp.proto",
-    "google/protobuf/type.proto",
-    "google/protobuf/wrappers.proto",
-]
+# Map of all well known protos.
+# name => (include path, imports)
+WELL_KNOWN_PROTO_MAP = {
+    "any" : ("google/protobuf/any.proto", []),
+    "api" : ("google/protobuf/api.proto", ["source_context", "type"]),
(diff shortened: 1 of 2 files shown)


BUILD
```diff
@@ -259,6 +262,25 @@ cc_proto_library(
...
+internal_copied_filegroup(
+    name = "_internal_wkt_protos",
+    srcs = WELL_KNOWN_PROTOS,
+    dest = "",
+    strip_prefix = "src",
+    visibility = ["//visibility:hidden"],
+)
+
+[proto_library(
+    name = "well_known_types_" + proto[0] + "_proto",
```
xfxyjwf: Is this prefix needed to avoid conflict? If not it seems better to just use the simpler name.

buchgr: We can also just use the simpler name. SGTM.

xfxyjwf: Let's use the simpler version then. Can you update this pull request?

buchgr:
Updated. Please take another look @xfxyjwf

BUILD
```diff
@@ -259,6 +262,25 @@ cc_proto_library(
...
+################################################################################
+
+internal_copied_filegroup(
+    name = "_internal_wkt_protos",
+    srcs = WELL_KNOWN_PROTOS,
+    dest = "",
+    strip_prefix = "src",
+    visibility = ["//visibility:hidden"],
+)
+
```
xfxyjwf: Can you add a comment to explain how to use these protos? An example should be good enough.

buchgr: Done

xfxyjwf:
Thanks!

PR #3597, opened 2017-09-05
Calling reflected c'tors no longer breaks MergeFrom.

zunger-humu:
Hey everyone,

I came across a nasty little bug with Python reflection, with a simple workaround. The problem can be triggered by a map whose value is a nested message.

*Problem description:* Both the constructors of protoc-generated messages and of reflection-generated messages (GeneratedProtocolMessageType.__init__) set descriptor._concrete_class to themselves. Unfortunately, if you access a class both directly and through reflection, they try to do this to the same descriptor, and hilarity ensues.

A way to manifest this problem is to print a protobuf (constructed normally) containing a map whose value is a nested (not foreign) type. When the printer reaches the map field, it will call value.GetEntryClass()(key=...) on the MapMessage object, which in turn calls descriptor._concrete_class(); the constructor then tries to MergeFrom on the output of that, merging the contents of the value field from the class being printed into what was made by the concrete class. But if the generated class has ever been constructed, descriptor._concrete_class() will now return one of *those*, and MergeFrom will throw a TypeError because it's trying to merge two different classes!

*A hypothetical hard fix:* would be to have multiple concrete classes available in the descriptor and switch through them based on something or other.

*A much easier fix:* There are really two bugs here: the fact that both classes are modifying the same descriptor, and the fact that MergeFrom is testing class match incorrectly. Specifically, MergeFrom should work fine if you are trying to merge a generated class into a compiled class or vice-versa. Therefore, a much better fix is to replace the isinstance() check on python_message.py:1212 with a check that also allows it to pass if the two classes are protos whose descriptor matches.

There do not appear to be any other points in the protobuf code that require such a change; the only other dubious uses of isinstance() are in the two equality checks (_AddEqualsMethod.__eq__ and _ExtensionDict.__eq__), where it's arguable that returning false when comparing generated and non-generated messages is reasonable. (Although you could argue the other way; proto equality has always been a contentious subject)

This pull request contains the fix and a unittest which reproduces the core issue.,

### DIFF
--- a/python/google/protobuf/internal/reflection_test.py
+++ b/python/google/protobuf/internal/reflection_test.py
@@ -47,6 +47,7 @@ try:
 except ImportError:
   import unittest
 
+from google.protobuf import map_unittest_pb2
 from google.protobuf import unittest_import_pb2
 from google.protobuf import unittest_mset_pb2
 from google.protobuf import unittest_pb2
@@ -1479,6 +1480,30 @@ class ReflectionTest(BaseTestCase):
     # Merge into message2.  This should not instantiate the field is message2.
     message2.MergeFrom(message1)
     self.assertFalse(message2.HasField('optional_nested_message'))
+
+  def testReflectionDoesntBreakMerging(self):
+    """Test that building a class from reflection doesn't break merging."""
+    # Instantiate a TestMap from the generated class. (We use this class
+    # because copying a MessageMap relies on the map value's descriptor's
+    # concrete class, unlike most other copies, and is especially bug-prone)
+    proto = map_unittest_pb2.TestMap()
+    proto.map_string_nested_message["one"].c = 1234
+
+    print "First: %s" % proto
+
+    # Instantiate one from reflection
+    generated = reflection.MakeClass(proto.DESCRIPTOR)()
+    generated.map_string_nested_message["two"].c = 2345
+
+    print "Second: %s" % generated
+    print "Third: %s" % proto
+
+    # The concrete class of their shared descriptor has changed! Make sure
+    # we can still assign copies.
+    proto2 = map_unittest_pb2.TestMap()
+    proto2.MergeFrom(proto)
+    self.assertEqual(proto.map_string_nested_message["one"],
+                     proto2.map_string_nested_message["one"])
 
   def testCopyFromSingularField(self):
     # Test copy with just a singular field.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -100,3 +100,5 @@ Patch contributors:
   Andrew Paprocki <andrew@ishiboo.com>
     * Fixed minor IBM xlC compiler build issues
     * Added atomicops for AIX (POWER)
+  Yonatan Zunger <zunger@humu.com>
+    * Fixed python proto issue with reflected classes breaking descriptors.
--- a/python/google/protobuf/internal/python_message.py
+++ b/python/google/protobuf/internal/python_message.py
@@ -1208,8 +1208,17 @@ def _AddMergeFromMethod(cls):
   LABEL_REPEATED = _FieldDescriptor.LABEL_REPEATED
   CPPTYPE_MESSAGE = _FieldDescriptor.CPPTYPE_MESSAGE
 
+  def __IsInstance(msg, cls):
+      """Check if msg is an instance of class, handling protos correctly."""
+      if isinstance(msg, cls):
+          return True
+      if hasattr(msg, "DESCRIPTOR") and hasattr(
+              cls, "DESCRIPTOR") and msg.DESCRIPTOR == cls.DESCRIPTOR:
+          return True
+      return False
+
   def MergeFrom(self, msg):
-    if not isinstance(msg, cls):
+    if not __IsInstance(msg, cls):
       raise TypeError(
           "Parameter to MergeFrom() must be instance of same class: "
           'expected %s got %s.' % (cls.__name__, msg.__class__.__name__))
--- a/python/google/protobuf/internal/python_message.py
+++ b/python/google/protobuf/internal/python_message.py
@@ -50,26 +50,22 @@ this file*.
 
 __author__ = 'robinson@google.com (Will Robinson)'
 
-from io import BytesIO
 import struct
 import sys
 import weakref
+from io import BytesIO
 
 import six
 
-# We use "as" to avoid name collisions with variables.
-from google.protobuf.internal import api_implementation
-from google.protobuf.internal import containers
-from google.protobuf.internal import decoder
-from google.protobuf.internal import encoder
-from google.protobuf.internal import enum_type_wrapper
-from google.protobuf.internal import message_listener as message_listener_mod
-from google.protobuf.internal import type_checkers
-from google.protobuf.internal import well_known_types
-from google.protobuf.internal import wire_format
 from google.protobuf import descriptor as descriptor_mod
 from google.protobuf import message as message_mod
 from google.protobuf import text_format
+from google.protobuf.internal import message_listener as message_listener_mod
+# We use "as" to avoid name collisions with variables.
+from google.protobuf.internal import (api_implementation, containers, decoder,
+                                      encoder, enum_type_wrapper,
+                                      type_checkers, well_known_types,
+                                      wire_format)
 
 _FieldDescriptor = descriptor_mod.FieldDescriptor
 _AnyFullTypeName = 'google.protobuf.Any'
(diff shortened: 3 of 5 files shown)


bazel-io:

zunger-humu:
I signed it!

zunger-humu:
Hello, @googlebot. I am attempting to ping you.

zunger-humu:
Hello, @googlebot. I am pinging you again because the system sent me an e-mail.

xfxyjwf:
@zunger-humu Can you double-check? Our internal database doesn't show you have signed the CLA.

zunger-humu:
@xfxyjwf Weird. When I go to the CLA site, it tells me that I'm covered by (and an administrator for) Humu's corporate CLA with Google. Could there have been some propagation lag?

zunger-humu:
@xfxyjwf Alternatively: Is it possible that setting my e-mail address to private was breaking it? I tried setting it to public; let's see if that fixes it.

zunger-humu:
Also, ping @anandolee -- could you take a look?

xfxyjwf:
@zunger-humu It still reports CLA not signed. Also the commit you created is associated with another github account @yonatanzunger .

I'll check with our internal team about whether public email address is required for the CLA bot.

zunger-humu:
@xfxyjwf Hah! It was the wrong name on the commit. Fixed now. Thank you!

@anandolee, still ready for review...

zunger-humu:
@anandolee, ping? It's been two weeks.

zunger-humu:
@xfxyjwf @anandolee Is there any chance of having this PR reviewed and merged? We're encountering this bug all the time in our production code (and working out of a branch to patch around it), and it's kind of silly not to have the fix actually go in.

pherl:
Will there be problem if you use the same descriptor to generate multiple classes? I feel like this use case is not supported - there might be other users besides the merge call. Could you serialize the other message and parse from the binary instead? That should be faster if you are using the C++ implementation.

zunger-humu:
This isn't triggered by explicit calls to MergeFrom, necessarily; the original case was triggered by text_format.MessageToString, for example. So working around the specific call sites doesn't work.

As far as using the same descriptor for multiple classes: Apart from the case of classes made from the generated and reflected classes, are there other situations where this could happen? If a user subclasses a proto to extend it (without overriding MergeFrom) and calls CopyFrom with the two different classes as args, I'd say that copying the proto fields is exactly expected behavior – and certainly less unexpected than having real MergeFrom's fail.

pherl:
I think the root cause is having multiple concrete classes on the same descriptor object, which is not supported; the protoc generated classes also use reflection to create message classes dynamically. You'd need to duplicate the descriptor before creating new concrete class reflectively. Those classes are considered distinct and not interoperable - except via the binary parsing/serialization. Subclassing of proto classes is also not recommended, and is explicitly forbidden in languages that support such feature. Composition should be used instead.

I'd like to understand a bit more why you need to dynamically generate classes that are already generated by protoc.

zunger-humu:
If I were the one making two concrete classes out of the same descriptor object, this bug would be no more than I deserved. :) Unfortunately, it's not me: it's the proto reflection library. Specifically, protobuf.reflection.MakeClass() returns a different class (something in the google.protobuf.internal namespace) than the constructor of the class in the generated code.

(The C++ proto impl at least used to do the same thing; I can't remember if it still does)

Unfortunately, these two classes share the same descriptor, and their joint superclass' `__init__` method sets descriptor._concrete_class = cls (python_message.py:164). As a result, various places where proto objects are _implicitly_ created get confused as well. For example, if you have a proto map whose value is a nested (not foreign) message type, and you call text_format.MessageToString() on your message, then when it's trying to grab an object for the value the printer calls value.GetEntryClass() (text_format.py:348), which ultimately uses descriptor._concrete_class to instantiate the object.

Then, since the generated class' first constructor always ran first (at import time), its concrete class has been overridden in the (global static) descriptor by the one for the reflected class, and so GetEntryClass() gets back a new object of the reflected type. It then tries to copy from the object you're trying to print into that – so if you're trying to print a non-reflected object, it's now trying to CopyFrom between a non-reflected and a reflected object, and (without this fix) that throws an exception.

The text_format case is just a specific situation where it shows up; I've found a bunch of other random ones as well, often when there's a proto somewhere in a dict or some other behavior which implicitly causes one thing to get copied from another.

But in all cases, there are only two classes which can share a message descriptor: the reflected and the generated one. And those _should_ have a working CopyFrom between them.

zunger-humu:
It's this PR's one-month birthday!

Ping for reviewers? :)

python/google/protobuf/internal/reflection_test.py
```diff
@@ -1480,6 +1481,30 @@ def testMergeFromBug(self):
...
+    # Instantiate a TestMap from the generated class. (We use this class
+    # because copying a MessageMap relies on the map value's descriptor's
+    # concrete class, unlike most other copies, and is especially bug-prone)
+    proto = map_unittest_pb2.TestMap()
+    proto.map_string_nested_message["one"].c = 1234
+
+    print "First: %s" % proto
+
+    # Instantiate one from reflection
+    generated = reflection.MakeClass(proto.DESCRIPTOR)()
```
pherl: The descriptor by design is only allowed to be bound with one concrete class e.g. we only have one concrete_class field. The generated message descriptor is already owned by the generated concrete class. As I mentioned in the previous comment, making a second class on the generated descriptor object is not supported. You would need to get the FileDescriptorProto out and build your own descriptor for the new class.

To really support making multiple classes on the same descriptor, we would have to redesign how descriptor and concrete class work. Otherwise it may simply break the python c++ extension based implementation, which relies the concrete class field. I don't think treating those two classes the same fixes the root problem.

I would suggest creating your own descriptors instead.

zunger-humu: Unfortunately, the Python protobuf code itself doesn't obey this invariant: as noted above, reflection.MakeClass() doesn't return an instance of the same class as is produced by the generated _pb2.py code. (They have the same class name but in different packages, and they share a descriptor)

The C++ implementation has a similar issue; there, reflection creates an object of some dedicated "reflected message" type, which is not the same as what shows up in the pb.cc's. We resolve it there the same way as we do here.

pherl: I think then it's a documentation issue. reflection.MakeClass() requires the descriptors are built by the users, rather than from the protoc generated code. It should throw exceptions when the descritpor is already bound.

zunger-humu: Whoa, what? If that's true, then how are you supposed to parse a proto given its name? I'm doing the classic thing: storing the proto name and data type in an underlying store, then fetching the descriptor from the symbol database and passing it to reflection. 

To hand-build the descriptors would require copying a descriptor from the symbol database, which would require a deep-copy operation for descriptors (doesn't exist), and explicitly registering those copies in the symbol database (exists but is undocumented). And then you couldn't copy between a reflected and default object, which is especially weird: it's very natural to want to convert between the two, especially in a library.

The C++ code handles this very naturally, by treating the descriptor object as the proto identity; all these problems go away. (I'm painfully aware of this after having to factor out codex's crazy proto logic into SourceFileDatabase a few years ago) What I'm proposing is basically the identical behavior for the python code, and it makes everything "just work" the way callers expect it.

pherl: I see. Sorry for the confusion. I thought you'd like to create your own classes. It seems like the reflection API was already broken by the concrete_class field. We need to either loosen the restriction by allowing multiple concrete classes (e.g. make that into a list), or introducing another API to retrieve the generated classes instead of creating new ones each time. I have forwarded this discussion to people working on protobuf python and they will chime in.

anandolee: Sorry for late reply. I looked into the code, does this solution sounds good for your usage? Add a check is the descriptor has '_concrete_class' already at the beginning of reflection.MakeClass():

  if hasattr(descriptor, '_concrete_class'):
    return descriptor._concrete_class

Your tests can pass with this solution.

zunger-humu: Done.

pherl: This will change the behavior though, which has potential risks of breaking other users. 

We are still discussing about the proper fix for those APIs.  Mostly important question is about the semantics: shall we allow multiple classes created on the same descriptor. The implementation assumed this is disallowed, however we have two public APIs violating it, namely reflection.MakeClass and MessageFactory.GetPrototype. Also we need to converge the two APIs which basically are the same.

zunger-humu: /nod. That makes good sense. 

If it's any help, the C++ code didn't make this assumption as strongly as people had thought. The Python code primarily seems to do it through the _concrete_class variable, which isn't as widely depended on as it seems to be; I could imagine restructuring things to no longer depend on it.

But one invariant that turned out to make a lot of sense was that descriptor matching indicated proto _compatibility,_ for things like MergeFrom and so on, and was almost always more useful inside the proto code than class matching.

anandolee: Updates. We are still discuss the right way to fix it. There are actually two issues:
1, reflection.MakeClass() is duplicate with message_factory.GetPrototype(). I am contacting the author who added reflection.MakeClass() to know the history and see if this can be removed.
2, One of our main contributor has tried multiple times to kill _concrete_class without success. We are planning to give it another try.

zunger-humu: Well, I'd be happy to see _concrete_class go away, but that did look like a monstrous PITA. :)

FWIW, a fix very similar to this did work in C++, and has one important protection against breaking existing users: It causes something which previously didn't work (merging a reflected and constructed object) to work, but not vice-versa. So it would take a pretty exotic user (someone who was using a failure of MergeFrom to detect a reflected object?!) to be broken by this.

anandolee: We have marked the  reflection.MakeClass() deprecated and ask users to use MessageFactory.GetPrototype().

Is your test works well with MessageFactory.GetPrototype() without fix?

anandolee:
reflection.ParseMessage() and reflection.MakeClass() are now marked as deprecated. Users are suggested to use message factory instead

anandolee:
I am closing this PR as reflection.MakeClass() are now marked as deprecated. Feel free to reopen if still want to go with the fix

PR #3603, opened 2017-09-06
Update switch's fallthrough for GCC 7

tiagofalcao:
Implicit fallthrough throws an warning on GCC 7 with -Wextra.

The current code already supports the explicit attribute using
GOOGLE_FALLTHROUGH_INTENDED at specific configuration with clang.

This commit supports the C++17 new attribute [[fallthrough]] and the GNU
version, [[gnu:fallthrough]].

I kept the current rules to enable the attribute when compiling with clang.


bazel-io:

tiagofalcao:
I signed it!

Issue #3545, opened 2017-08-22
PHP protobuf generation and reserved keywords


michaelbausor:
There are some keywords that are not correctly handled in protobuf generation:
- `DEFAULT` - see here: https://github.com/googleapis/gax-php/blob/master/src/generated/Google/Logging/Type/LogSeverity.php#L30
- `Namespace` - we have an upcoming API that has a protobuf message type called `Namespace`

In the first case, we have resolved the issue with a manual edit. In the second case, it is not possible to rename the class with a manual edit.

It would perhaps be appropriate to also attempt to determine all possible PHP keywords, and handle them correctly, rather than only addressing these cases. I can put together a test protobuf file that uses all of the keywords that require special case handling.

TeBoring:
Do you have a list of reserved keywords. I want to add them all at once.

michaelbausor:
I will create a list, and also an example proto file that uses them that we can put into tests

michaelbausor:
Here is a proto file that contains keywords: https://gist.github.com/michaelbausor/d38e76da78a0c6206ec31b5a09689816

The keywords were compiled from here: http://php.net/manual/en/reserved.php

I would recommend adding a test which runs the proto compiler on this proto file, and then parses and instantiates all of the classes generated using a test like this one: https://github.com/googleapis/proto-client-php/blob/master/tests/Unit/InstantiateClassesTest.php
(Note that this test also needs to handle the generated grpc stub classes, so it can be simplified for the protobuf-only case)

PR #3612, opened 2017-09-08
Add prefix to php reserved keywords.

TeBoring:


bazel-io:

TeBoring:
Fix #3545

php/src/Google/Protobuf/Internal/GPBUtil.php
```diff
@@ -252,9 +252,23 @@ public static function getClassNamePrefix(
             return $prefix;
         }
 
-        $reserved_words = array("Empty", "ECHO", "ARRAY");
+        $reserved_words = array(
```
michaelbausor: Perhaps this could be an associative array, and then the check could be done with a lookup instead of a loop?

TeBoring: Done

michaelbausor:
These changes LGTM - not sure if it is also desirable to use a lookup instead of a linear search in the c code, similar to my comment in the PHP code, other whether that is not worth the additional complexity.

I will try testing code generation with this change.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -49,8 +49,23 @@ const std::string kDescriptorMetadataFile =
     "GPBMetadata/Google/Protobuf/Internal/Descriptor.php";
 const std::string kDescriptorDirName = "Google/Protobuf/Internal";
 const std::string kDescriptorPackageName = "Google\\Protobuf\\Internal";
-const char* const kReservedNames[] = {"ARRAY", "Empty", "ECHO"};
-const int kReservedNamesSize = 3;
+const char* const kReservedNames[] = {
```
michaelbausor: There is a slightly different behavior between which classes are permitted as class names and which are permitted as constants. The names `int`, `float`, `bool`, `string`, `true`, `false`, `null`, `void` and `iterable` are valid as PHP constants, but not as class names.

We could support allowing these by supporting a reduced set of reserved names that would be checked here instead of the full set checked in ClassNamePrefix: https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/php/php_generator.cc#L984

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -161,6 +166,34 @@ std::string ClassNamePrefix(const string& classname,
   return "";
 }
 
+std::string ClassNamePrefix(const string& classname,
+                            const EnumValueDescriptor* desc) {
```
michaelbausor: The value of `desc` is not used at all in this function. Perhaps this could be removed, and the function renamed to something like `ConstantNamePrefix`, since it is only used in that context.

TeBoring: Done

michaelbausor:
LGTM

PR #3614, opened 2017-09-09
Add _C_DESCRIPTOR_CLASS to DescriptorBase

pfhayes:
Not necessarily advocating for this change, mostly asking as a question and thought this PR was perhaps the easiest way to answer

Why is `pyext._message.DescriptorBase` not a virtual subclass of `google.protobuf.descriptor.DescriptorBase` as the other descriptors are? I ran into this issue when migrating to the C++ runtime and observing that an `isinstance(d, DescriptorBase)` which previously passed no longer does


bazel-io:

pfhayes:
signed

anandolee:
MacOS Python C++ failed. Though it is marked as Allowed Failure, but as it is a python c++ change so it may not a flakey failure. Try another tests

pfhayes:
How can I see the test results? The "Linux C#" and the "MacOS Python C++ (Allowed Failure)" builds have failed, but the "Details" link takes me to some broken "sponge" link I cannot access. Is this a private link internal to the protobuf project?

anandolee:
I reported this issue and should be fixed now to point ResultStore URL instead of Sponge.

pfhayes:
@anandolee now that the build steps have been updated, the checks have all passed

anandolee:
kokoro will not run when you update the PR unless we add kokoro:run once we think is it ready to test. The default check only checks cla and a subset of the tests( I only see c++ and csharp from the continuous link)

Will add  kokoro:run label to trigger another run

anandolee:
Try add PyType_Ready in
https://github.com/google/protobuf/blob/master/python/google/protobuf/pyext/descriptor.cc#L1870

anandolee:
@amauryfa, do you have any idea why MacOS Python C++ can not pass on the PR?

anandolee:
Will close it for clean up. Feel free to reopen after resolve the conflicts if still need.

PR #3621, opened 2017-09-12
Google.Protobuf should target net45 (upport to upstream/master)

jtattermusch:
Upport https://github.com/google/protobuf/pull/3596 to upstream/master

(to make sure this change will land in the following releases too).


bazel-io:

Issue #3120, opened 2017-05-22
Ruby: Allow for the instantiation of nested message fields with ruby-hashes


landrito:
# What

Currently instantiation of a protobuf message can be done using a ruby-hash but does not support instantiating the nested messages using nested hashes. Being able to support the instantiation of messages including nested messages using a single hash would be a **big** usability win for [googleapis/toolkit](https://github.com/googleapis/toolkit).

# Example
### Current
```rb
message = MyMessage.new({
  submessage_field: SubMessage.new(:foo => 42)
})
```
### Requested
```rb
message = MyMessage.new({
  submessage_field: {foo: 42}
})
```

landrito:
cc @geigerj @lukesneeringer

acozzette:
CC @haberman 
That sounds like a useful feature; I'm not sure if anyone on the proto team has the cycles to work on this but feel free to send us a pull request if you want to try implementing it.

geigerj:
@acozzette We have limited bandwidth for the remainder of the quarter, but this is an important usability problem for the Ruby Cloud client libraries. Could you send us some code links so we can get some sense what changes would be needed to the protobuf codebase to enable this?

(If we're not able to make the fix in protobuf, we might work around by providing a utility method in the client that transforms hashes to protobuf messages. But if possible, I'd prefer to fix this here rather than in the client libraries.)

acozzette:
@geigerj I'm not actually that familiar with the protobuf implementation for Ruby. @haberman, maybe you could comment on this?

geigerj:
@haberman -- could you advise here?

geigerj:
@zanker @acozzette It looks like this was implemented in #3627, in which case this issue could be closed. But I'm having trouble getting the feature to work in protobuf `3.5.0`:

```
> require 'google/protobuf/struct_pb'
 => true 
> s = Google::Protobuf::Struct.new(fields: {'a' => {number_value: 0.0}})
TypeError: Invalid type Hash to assign to submessage field.
```

@zanker -- is this the intended usage of the feature you implemented in #3627?

zanker:
Hm, that's odd. That is the correct syntax, and the test for it https://github.com/google/protobuf/pull/3627/files#diff-bbb59c88129c0bf6537d0487634b282dR205 is the same as what you have, and it does look like the code made it into `3.5.0`.

Silly question, but are you sure you're pulling `3.5.0`? That was something I had running on Square code before I left, so the feature does work.

[...]

PR #3627, opened 2017-09-12
Allow initializing a chain of protos using only a hash in Ruby

zanker:
I saw #3120  also asked for this. We're in the process of trying to migrate people from a pure Ruby proto gem to the Google one, and the requirement to initialize all protos is a blocker. Sounds like there was interest in this, so figured I'd take a stab at implementing it in.

Taking the examples from #3120, this lets you do: `MyMessage.new(submessage_field: { foo: 42 })`and it will automatically turn the `{ foo: 42 }` into a `SubMessage.new(foo: 42)`.

Since a common scenario is also to be turning JSON into protos (for proto backed JSON APIs), I also changed the initialization to support symbol or string keys to prevent having to do a deep symbolization pass before initializing the protos.

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -217,20 +217,32 @@ VALUE Message_respond_to_missing(int argc, VALUE* argv, VALUE _self) {
   return Qtrue;
 }
 
+VALUE create_submsg_from_hash(const upb_fielddef *f, VALUE hash) {
+  const upb_def *d = upb_fielddef_subdef(f);
+  assert(d != NULL);
+
+  VALUE descriptor = get_def_obj(d);
+  VALUE msgclass = rb_funcall(descriptor, rb_intern("msgclass"), 0, NULL);
+
+  VALUE args[1] = { hash };
+  return rb_class_new_instance(1, args, msgclass);
+}
+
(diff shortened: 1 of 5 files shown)


bazel-io:

haberman:
ok to test

haberman:
Looks like this change failed Travis?

zanker:
@haberman Fixed the test that failed. It was one of the compatibility tests, which now fails with `RangeError` because it's now checking if `z` is an enum value, whereas previously it would fail with `TypeError` since it was a string.

Is the fact that the exception changed ok?

ruby/ext/google/protobuf_c/message.c
```diff
@@ -248,16 +260,26 @@ int Message_initialize_kwarg(VALUE key, VALUE val, VALUE _self) {
     Map_merge_into_self(map, val);
   } else if (upb_fielddef_label(f) == UPB_LABEL_REPEATED) {
     VALUE ary;
+    VALUE entry;
```
haberman: This is only used inside the for loop, so please declare it there.

zanker: Done!

TeBoring:
@zanker could you take a look for #4809?

PR #3639, opened 2017-09-14
Fixed to_h with repeated messages to return hashes in Ruby

zanker:
Calling `to_h` converts everything to a Ruby object representation, except if you have a repeated field with a message, then you get the message field object. This fixes it so it converts that to a hash too (and recursively does it).

Calling `to_h` in JRuby didn't work if you had a repeated field, which is fixed in this too (mostly incidentally). `RubyRepeatedField` includes `RubyEnumerable` which responds to `to_h`, so it was trying to call `to_h` on an array, which errored. `to_h` in JRuby with a map field is still broken, because it converts it to a key/value representation where value is the proto object not the hash representation. Didn't fix that here, because it was unrelated to fixing repeated messages.

I think it likely makes more sense if `RepeatedField_to_ary` and `RubyRepeatedField#toArray` did this, but I was opting to make less dramatic changes.

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -419,6 +419,13 @@ VALUE Message_to_h(VALUE _self) {
       msg_value = Map_to_h(msg_value);
     } else if (upb_fielddef_label(field) == UPB_LABEL_REPEATED) {
       msg_value = RepeatedField_to_ary(msg_value);
+
+      if (upb_fieldddef_type(field) == UPB_TYPE_MESSAGE) {
+        for (int i = 0; i < RARRAY_LEN(msg_value); i++) {
+          VALUE elem = rb_ary_entry(msg_value, i);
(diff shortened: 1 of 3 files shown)


bazel-io:

haberman:
ok to test

ruby/ext/google/protobuf_c/message.c
```diff
@@ -419,6 +419,13 @@ VALUE Message_to_h(VALUE _self) {
       msg_value = Map_to_h(msg_value);
     } else if (upb_fielddef_label(field) == UPB_LABEL_REPEATED) {
       msg_value = RepeatedField_to_ary(msg_value);
+
+      if (upb_fielddef_msgsubdef(field) != NULL) {
```
haberman: I think it would be a bit clearer to write:

```
if (upb_fieldddef_type(field) == UPB_TYPE_MESSAGE)
```

zanker: Makes sense! I fixed that.

PR #3643, opened 2017-09-14
for generated pb.cc files, avoid __FILE__ in VERIFY_VERSION macro call

packetzero:
Currently, whether generated .pb.cc files are built in release or debug mode, the source filenames are included as strings in the resulting binary.

This is because protoc-cpp adds a call to the GOOGLE_PROTOBUF_VERIFY_VERSION macro in the InitDefaultsImpl function of every generated C++ protobuf message file.  That macro contains the __FILE__ macro.  Therefore, if you have a source file named MySpecialMessage.pb.cc, runnings *strings* on the resulting .o or .a or executable will reveal all the source file names.  If you are using CMAKE for your build, it will include the absolute path of all .pb.cc files (e.g. "/Users/bob/dev/blah/generated/MySpecialMessage.pb.cc").

This change follows the behavior of the assert macro, which will omit __FILE__ when NDEBUG is defined.

### DIFF
--- a/src/google/protobuf/any.pb.cc
+++ b/src/google/protobuf/any.pb.cc
@@ -92,7 +92,7 @@ void TableStruct::Shutdown() {
 }
 
 void TableStruct::InitDefaultsImpl() {
-  GOOGLE_PROTOBUF_VERIFY_VERSION;
+  GOOGLE_PROTOBUF_VERIFY_VERSION_FOR("google/protobuf/any");
 
   ::google::protobuf::internal::InitProtobufDefaults();
   _Any_default_instance_.DefaultConstruct();
--- a/src/google/protobuf/api.pb.cc
+++ b/src/google/protobuf/api.pb.cc
@@ -130,7 +130,7 @@ void TableStruct::Shutdown() {
 }
 
 void TableStruct::InitDefaultsImpl() {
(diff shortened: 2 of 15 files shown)


bazel-io:

packetzero:
cla: I signed it!

xfxyjwf:
Unfortunately we cannot use NDEBUG in header files due to the reasons stated in: https://github.com/google/protobuf/issues/3322

Can you think of a solution that doesn't require the use of NDEBUG macro?

packetzero:
So the latest revision uses a relative path to proto file without ".proto" extension as the name ("google/proto/descriptor") and for most users simply "MyProtoMessage".  Personally, I would prefer DEBUG or NDEBUG to enable/disable __FILE__ usage.  The issues in #3322 that are a concern are only for changes that alter the behavior or message hash.  Etiher way, this pull request is ready for review.

xfxyjwf:
Sorry for the late response. If you only need to hide the prefix of the build path, maybe you can build it in a place where the prefix doesn't matter? For example, /home/build/.... The alternative is for protobuf code base to support a GOOGLE_PROTOBUF_NO_FILE_MACRO macro that you can define to switch off the use of __FILE__ in our code base.

If you are still interested in getting it in, feel free to reopen.

PR #3659, opened 2017-09-19
Reserve unknown fields in php

TeBoring:
1) For decoding, an unknownfields will be lazily created on message,
which contains bytes of unknown fields.
2) For encoding, if the unknownfields is present on message, all bytes
contained in it will be serialized.


bazel-io:

stanley-cheung:
PHP changes LGTM

PR #3663, opened 2017-09-19
Make sure pragma GCC diagnostic push / pop is only added for supported compilers

RoelBlomjousDevelopment:
`#pragma GCC diagnostic push `and `#pragma GCC diagnostic pop` were added in [gcc 4.6](https://gcc.gnu.org/gcc-4.6/changes.html).


bazel-io:

PR #3675, opened 2017-09-22
protobuf_generate: add EXPORT_MACRO option

hesmar:


bazel-io:

a-andre:
What's missing now is a way to include a header into the generated protobuf-header which defines the EXPORT_MACRO.

PR #3693, opened 2017-09-27
When initializing a message, skip a field if value is nil

zanker:
This is mostly in the same vein as https://github.com/google/protobuf/pull/3627 for Ruby.

It allows doing: `MyMessage.new(string_field: nil)` and `string_field` is treated as a default. It makes it easier to construct protos when you're building up the options across multiple files, or from JSON where a nil value isn't necessarily a bad thing.

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -249,6 +249,10 @@ int Message_initialize_kwarg(VALUE key, VALUE val, VALUE _self) {
              "Unknown field name '%s' in initialization map entry.", name);
   }
 
+  if (TYPE(val) == T_NIL) {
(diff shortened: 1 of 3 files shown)


bazel-io:

zanker:
I signed it

cheister:
@haberman can you take a look at this?  This would be helpful for us.

TeBoring:
Doesn't not passing these parameters already give you the default values?

zanker:
Yes, but it's not uncommon that you have a hash that contains `nil`s and you want them to be excluded, without having to recursively filter out `nil` values. The typical example is if I parse a blob of JSON data and pass it to a proto, it's fine for the values to be nil.

PR #3695, opened 2017-09-28
Prefer using compiler built-in endianness definitions.

EdSchouten:
In order to make Protobuf build and work on CloudABI, a sandboxed
UNIX-like runtime environment, we have to extend the endianness
detection code. Instead of adding code specific to CloudABI, let's
change this to simply use definitions that are documented by GCC and
Clang, so that any other operating system/runtime may benefit.

### DIFF
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -113,7 +113,9 @@
 #include <climits>
 #include <string>
 #include <utility>
-#ifdef _MSC_VER
+#if defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__)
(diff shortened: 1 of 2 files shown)


bazel-io:

src/google/protobuf/io/coded_stream.h
```diff
@@ -113,7 +113,9 @@
 #include <climits>
 #include <string>
 #include <utility>
-#ifdef _MSC_VER
+#if defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__)
+  #define PROTOBUF_LITTLE_ENDIAN (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
```
pherl: The callsite code actually checks for ```#if defined(PROTOBUF_LITTLE_ENDIAN)``` so should probably move this in to the line above.

PR #3698, opened 2017-09-29
protobuf_generate: create include path only for proto files

hesmar:


bazel-io:

pherl:
OK to test

PR #3701, opened 2017-10-01
configure gtest

toanju:
gtest is not configured by default which is fixed in this PR.


bazel-io:

toanju:
Closing since the proposed fix causes the distcheck tests to fail. The actual problem was that I have gtest installed. Thus gmock tries to use the local copy but protobuf tests force the use of the downloaded gtest.

PR #3707, opened 2017-10-02
Use constexpr more with VC++ 2017

randomascii:
Chrome's official builds have over 170 dynamic initializers for
variables of the form *::TableStruct::aux. Defining
PROTOBUF_CONSTEXPR_VAR to be constexpr for VS 2017 gets rid of all of
these and saves about 10 KB of binary size.

The Chromium bug crbug.com/341941 has more details.

### DIFF
--- a/src/google/protobuf/generated_message_table_driven.h
+++ b/src/google/protobuf/generated_message_table_driven.h
@@ -43,8 +43,8 @@
 
 // We require C++11 and Clang to use constexpr for variables, as GCC 4.8
 // requires constexpr to be consistent between declarations of variables
-// unnecessarily (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58541).
-#ifdef __clang__
+// VS 2017 Update 3 also supports this usage of constexpr.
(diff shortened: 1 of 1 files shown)


bazel-io:

randomascii:
I have made this test locally in a Chromium repo (which contains a local copy of google/protobuf) and confirmed that it works well with VS 2017, which is now our default compiler.

I think it would be best to land it here and then in Chromium, but I could do it in either order.

src/google/protobuf/generated_message_table_driven.h
```diff
@@ -43,8 +43,8 @@
 
 // We require C++11 and Clang to use constexpr for variables, as GCC 4.8
 // requires constexpr to be consistent between declarations of variables
-// unnecessarily (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58541).
```
xfxyjwf: keep this line?

xfxyjwf:
Thanks for the patch!

Issue #3649, opened 2017-09-15
Google::Protobuf::DescriptorPool.generated_pool.build SEGV at 0x00000000000090


tarcieri:
I've only seen it once while loading generated proto definitions, but it looks like a NULL struct dereference:

```
Segmentation fault at 0x00000000000090
```

```
-- Control frame information -----------------------------------------------
c:0138 p:---- s:0730 e:000729 CFUNC  :build
```

```
-- Machine register context ------------------------------------------------
 RIP: 0x00005566edca511f RBP: 0x00007f2530f2b188 RSP: 0x00007ffd8bd9fed0
 RAX: 0x0000000000000012 RBX: 0x00005566eebbbac0 RCX: 0x000000000000003d
 RDX: 0x000000000000013d RDI: 0x00005566eebbbac0 RSI: 0x0000000000000000
  R8: 0x00005566f1ef6ad0  R9: 0x0000000000000001 R10: 0x00005566f312d970
 R11: 0x00005566eebbbac0 R12: 0x00007f2530f28000 R13: 0x0000000000000004
 R14: 0x00007ffd8bda0010 R15: 0xcccccccccccccccd EFL: 0x0000000000010202
```

pherl:
Can you verify if this has been fixed in 2.4.1? There was a bug related to maps implementation in multi-thread.

tarcieri:
I assume you mean Ruby/MRI 2.4.1? I do happen to be using Ruby 2.4.0...

TeBoring:
3.4.1?

tarcieri:
Aah, this was on 3.4.1 which got yanked today. I upgraded to 3.4.1.1. Guess I'll see if it still has problems

TeBoring:
How can it be reproduced?

tarcieri:
I cannot reproduce it myself. It only occurred once.

apolcyn:
I think I am seeing the same thing.

I'm seeing this reproduce 100% of the time with the following repro:

starting with this proto file:

```
syntax = "proto3";
message Empty {}
```

and generating with:

```
$PROTOC -I . empty.proto --ruby_out=.
```

that generates this ruby-protobuf code:

```
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: empty.proto

require 'google/protobuf'

Google::Protobuf::DescriptorPool.generated_pool.build do
  add_message "Empty" do
  end
end

[...]

PR #3708, opened 2017-10-02
Fix ruby segment fault

TeBoring:
1) rb_ary_new cannot be called during allocate function. During allocate
fucntion, the containing object hasn't been marked and rb_ary_new may
invoke gc to collect containing object.
2) The global map should be marked before allocating it. Otherwise it
may be garbage collected.


bazel-io:

TeBoring:
Fix #3649

ruby/Rakefile
```diff
@@ -99,7 +99,13 @@ Gem::PackageTask.new(spec) do |pkg|
 end
 
 Rake::TestTask.new(:test => :build) do |t|
-  t.test_files = FileList["tests/*.rb"]
+  t.test_files = FileList["tests/*.rb"].exclude("tests/gc_test.rb")
+end
+
+# gc_test needs to be splited out to ensure the generated file hasn't been
```
haberman: "splited" -> "split"

TeBoring: done

joshrose:
Thanks for this @TeBoring . Any idea when this will be published as part of a new release? Is there somewhere we can look to track what's happening with assembling future releases? Thanks.

TeBoring:
We are working on 3.5.0 release. https://github.com/google/protobuf/milestone/21

PR #3722, opened 2017-10-06
Suppress VS2017 compiler/linker warnings

timou:
VS2017 warns about a few additional things, for example, files that
define no symbols (which is typically the result of a platform specific
ifdef not applying on Windows).

Suppress these warnings so the build is clean on VS2017.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -135,12 +135,15 @@ endif (protobuf_BUILD_SHARED_LIBS)
 if (MSVC)
   # Build with multiple processes
   add_definitions(/MP)
-  add_definitions(/wd4244 /wd4267 /wd4018 /wd4355 /wd4800 /wd4251 /wd4996 /wd4146 /wd4305 /wd4309)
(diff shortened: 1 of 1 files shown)


bazel-io:

timou:
I signed it!

cmake/CMakeLists.txt
```diff
@@ -135,12 +135,15 @@ endif (protobuf_BUILD_SHARED_LIBS)
 if (MSVC)
   # Build with multiple processes
   add_definitions(/MP)
-  add_definitions(/wd4244 /wd4267 /wd4018 /wd4355 /wd4800 /wd4251 /wd4996 /wd4146 /wd4305 /wd4309)
+  add_definitions(/wd4244 /wd4267 /wd4018 /wd4355 /wd4800 /wd4251 /wd4996 /wd4146 /wd4305 /wd4309 /wd4065 /wd4506 /wd4307 /wd4334)
```
pherl: could you help sort the whole list? it would be easier to add / lookup in the future.

timou: Good call; fixed. Also included descriptions of the suppressed warnings.

PR #3724, opened 2017-10-09
C# map merge exception bug-fix

freepvps:
Merging of two maps with intersected keys throw exception
I think, that it's not ok
Maybe "MergeFrom" method for Map<> is a good idea?

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_map_field.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_map_field.cc
@@ -91,7 +91,7 @@ void MapFieldGenerator::GenerateMembers(io::Printer* printer) {
 void MapFieldGenerator::GenerateMergingCode(io::Printer* printer) {
   printer->Print(
       variables_,
-      "$name$_.Add(other.$name$_);\n");
+      "foreach (var $name$Value_ in other.$name$_) {\n  if (!$name$_.ContainsKey($name$Value_.Key)) {\n    $name$_.Add($name$Value_.Key, $name$Value_.Value);\n  }\n}\n");
 }
 
 void MapFieldGenerator::GenerateParsingCode(io::Printer* printer) {


bazel-io:

freepvps:
I signed it!

src/google/protobuf/compiler/csharp/csharp_map_field.cc
```diff
@@ -91,7 +91,7 @@ void MapFieldGenerator::GenerateMembers(io::Printer* printer) {
 void MapFieldGenerator::GenerateMergingCode(io::Printer* printer) {
   printer->Print(
       variables_,
-      "$name$_.Add(other.$name$_);\n");
+      "foreach (var $name$Value_ in other.$name$_) {\n  if (!$name$_.ContainsKey($name$Value_.Key)) {\n    $name$_.Add($name$Value_.Key, $name$Value_.Value);\n  }\n}\n");
```
pherl: The merging semantics for maps is to override the existing value.

Also, usually we make a line wrap in the generator when you print '\n' for better readability.

There also needs to be a test to catch this behavior change.

jskeet:
Hmm. I think the C++ code is different already because the message fields aren't pointers, but I could be very wrong. (I have very little experience with C++ in general, and C++ protos in particular.)

I note that the Java code looks (a bit) like the C# code after this commit:

```java
void ImmutableMessageFieldGenerator::
GenerateMergingCode(io::Printer* printer) const {
  printer->Print(variables_,
    "if (other.has$capitalized_name$()) {\n"
    "  merge$capitalized_name$(other.get$capitalized_name$());\n"
    "}\n");
}
...
void ImmutableMessageOneofFieldGenerator::
GenerateMergingCode(io::Printer* printer) const {
  printer->Print(variables_,
    "merge$capitalized_name$(other.get$capitalized_name$());\n");
}
```

anandolee:
@freepvps  are you still working on fix it according to liujisi's comments?

freepvps:
@anandolee me sorry, I'm forget about this review, but I will continue fix it

PR #1511, opened 2016-05-10
python: force relative imports

little-dude:
fix https://github.com/google/protobuf/issues/1491

PR #3728, opened 2017-10-09
[protoc python] Relative import paths in generated python modules

cherezov:
In order to avoid module name conflict with modules in PYTHON_PATH relative package import is introduced to generated python files.

This means that all imports like ```from Module1.Module2 import ProtoPackage``` will now looks like
```
from . import ProtoPackage
from .. import Module2
```
This will allow us to safety have another module with name 'Module1' and 'Module2' in the python path.

### DIFF
--- a/src/google/protobuf/compiler/python/python_generator.cc
+++ b/src/google/protobuf/compiler/python/python_generator.cc
@@ -375,11 +375,12 @@ bool Generator::Generate(const FileDescriptor* file,
 
 // Prints Python imports for all modules imported by |file|.
 void Generator::PrintImports() const {
+  vector<string> modules_list;
   for (int i = 0; i < file_->dependency_count(); ++i) {
     const string& filename = file_->dependency(i)->name();
 
-    string module_name = ModuleName(filename);
-    string module_alias = ModuleAlias(filename);
+    const string module_name = ModuleName(filename);
+    const string module_alias = ModuleAlias(filename);
     if (ContainsPythonKeyword(module_name)) {
       // If the module path contains a Python keyword, we have to quote the
       // module name and import it using importlib. Otherwise the usual kind of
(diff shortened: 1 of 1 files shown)


cherezov:
I signed it!

bazel-io:

arthur-tacca:
(Note: I am not a Googler, or even a contributor) I think this is a dangerous pull request, because it masks mistakes in the use of `protoc`.

`protoc` creates directories containing `*_pb2.py` files to mirror the directories containing `*.proto` files. This is a bit confusing when you first see it, but is the Pythonic way to allow users to specify the packages that their `*_pb2.py` files should go in and import dependencies from, so long as they have correctly set the `--proto_path=` switch to `protoc`. For example, if you put all your proto files in a `proto/` directory but use `--proto_path=.`, you'll soon notice `from proto import dependancy_pb2`. This is a hint that you need to create an intermediate directory `./proto/package_name` and pass the switch `--proto_path=./proto`; then you end up with the desired `from package_name import dependancy_pb2`

Say you are getting an error from `from Package1.Package2 import ProtoModule` (I have switched your example to the proper Python terminology here) due to `Package1` being split over several directories in your `PYTHONPATH`. Then you should either (a) Make `Package1` a namespace package or (b) not have two places for `Package1`! The first option is best avoided because namespace packages are a bit of a tricky beast. The second option is usually pretty easy, because you just make the `--python_out=` parameter to `protoc` point to the location that also contains your non-generated source code for `Package1`. I suspect this point is the cause of your problems.

By the way, `from Package1.Package2 import ProtoModule` only creates a variable `ProtoModule`; it does not create a variable `Package2`, so there is no need for the second relative import in your example.

See also a similar pull request #1511 and related bug report #1491.

arthur-tacca:
I've realised I don't fully understand what this is for. If `package_name.high_level_pb2` tries to import `package_name.low_level_pb2`, but gets an error because there is another `package_name` earlier in the `PYTHON_PATH`, then relative imports won't help at all because you won't be able to import `package_name.high_level_pb2` in the first place the same reason! What situation would this actually help in?

pherl:
What if the dependency is in a different directory in the PYTHONPATH, and relative path wouldn't work? I guess you would have to reorganize your packages to avoid such conflicts.

tlyng:
Any progress on this issue? Currently it does not generate valid python3 code.

cherezov:
Actually no progress for now. Let's close my request at all.

PR #3733, opened 2017-10-10
cmake: Remove LIB_PREFIX and protobuf_DEBUG_POSTFIX

philix:
### LIB_PREFIX

CMake should decide the name of the output file of a static library. The right way to to this is to not have the `lib-` prefix in the CMake target names. So there's no need for the `LIB_PREFIX` variable and library target should preferably not have the `lib` prefix.

### protobuf_DEBUG_POSTFIX

The CMake files used to build `protobuf` are usually part of the CMake build of a project that uses `protobuf`. The debug suffix just makes is harder to find the static library artifacts from the main project.

The build files generated by CMake will have separate folders for different CMAKE_BUILD_TYPEs so there's no need to differentiate in the file names.


bazel-io:

philix:
@googlebot I signed it!

podsvirov:
-1

philix:
@podsvirov anything I can do to get a +1? Or is it a lost cause?

xfxyjwf:
I think the LIB_PREFIX is needed for windows msvc build where the compiler will just produce a protobuf.lib file instead of libprotobuf.lib. The "d" suffix is also the convention used by msvc build where libprotobuf.lib is the release build and libprotobufd.lib is a debug build.

podsvirov:
@philix, how long have you been using CMake and protobuf?
These projects have been developing for many years. Over the years, certain decisions have been made on naming output files and CMake targets. Also, support was provided in various build systems, including CMake.
In my opinion, renaming the names is not the best idea, from the point of view of backward compatibility.
Also note that your experience of using this project may not correspond to the experience of other users. In particular, the project can not only be built from source code, but also distributed in a precompiled form.
I still recommend to reject your changes.

philix:
> In my opinion, renaming the names is not the best idea, from the point of view of backward compatibility.

I agree. I made this change in my own fork and thought it would be good to follow more common CMake practices in upstream as well. I also felt the CMake build files were second-class citizens in the repo because they don't work as they come from the repo so I didn't expect backwards compatibility to be such an issue for the CMake build.

philix:
If I find time to keep the `OUTPUT_NAME` meddling and still have it working correctly when building in Xcode I will send another PR.

pherl:
Closing this for now due to compatibility.

PR #3735, opened 2017-10-10
Don't pass -lpthread and -lm on Windows

sgreenstein:
These flags are ignored by cl.exe and have no effect, but add noise to the build logs.


bazel-io:

PR #3736, opened 2017-10-11
Converting tags to immutable hashable types

jleni:
In a Raspberry Pi, tags are not hashable if data is passed as byterray, however, running in Ubuntu this is not a problem.

This PR addresses the problem as described in:
https://github.com/google/protobuf/issues/3734


bazel-io:

jleni:
I signed it!

pherl:
OK to test

jleni:
@pherl it seems that there is an unrelated problem with macos ruby. 
Is that OK?

PR #3737, opened 2017-10-11
Update version number for 3.4.1?

was4444:


bazel-io:

pherl:
We will bump this soon to 3.5.0 so probably doesn't worth the effort to change it now (you would have to re-generate descriptpors due to the version change).

PR #3739, opened 2017-10-11
Merge3.4

pherl:


Makefile.am
```diff
@@ -650,6 +650,9 @@ php_EXTRA_DIST=                                                       \
   php/src/Google/Protobuf/Internal/EnumOptions.php                    \
   php/src/Google/Protobuf/Internal/EnumValueDescriptorProto.php       \
   php/src/Google/Protobuf/Internal/EnumValueOptions.php               \
+  php/src/Google/Protobuf/Internal/ExtensionRangeOptions.php          \
+  php/src/Google/Protobuf/Internal/FieldDescriptorProto_Label.php     \
+  php/src/Google/Protobuf/Internal/FieldDescriptorProto.php           \
```
TeBoring: You probably also need to add new files after running generate_descriptor_proto

pherl: Good catch. The distcheck tests will probably catch other new files if there are any.

PR #3744, opened 2017-10-12
Move namespace closing brace inside the header guard block

fmarier:
This doesn't seem to cause problems when it's only included once, but it fails to build as part of the Firefox unified builds.


bazel-io:

PR #3745, opened 2017-10-12
Work around wingdi.h defining ERROR as 0

fmarier:
Without this fix, we get build errors like these on Windows when bundling protobuf in Firefox:

```
src/google/protobuf/stubs/strutil.cc(351): error C2039: 'LOGLEVEL_0': is not a member of 'google::protobuf'
src/google/protobuf/stubs/strutil.cc(60): note: see declaration of 'google::protobuf'
src/google/protobuf/stubs/strutil.cc(351): error C2065: 'LOGLEVEL_0': undeclared identifier
```

### DIFF
--- a/src/google/protobuf/stubs/logging.h
+++ b/src/google/protobuf/stubs/logging.h
@@ -59,6 +59,19 @@ enum LogLevel {
 #else
   LOGLEVEL_DFATAL = LOGLEVEL_FATAL
 #endif
+
+#ifdef _WIN32
+  // wingdi.h defines ERROR to be 0, so `GOOGLE_LOG(ERROR, ...)` expands
+  // into `GOOGLE_LOG(0, ...)` which then expands into
+  // `someGoogleLogging(LOGLEVEL_0, ...)`. This is not ideal, because the
+  // GOOGLE_LOG macro expects to expand itself into
+  // `someGoogleLogging(LOGLEVEL_ERROR, ...)` instead. The workaround to
+  // get everything building is to simply define LOGLEVEL_0 as
+  // LOGLEVEL_ERROR and also define ERROR the same way that the Windows
+  // SDK does for consistency.
+#define ERROR 0
+  , LOGLEVEL_0 = LOGLEVEL_ERROR
+#endif
 };
 
 class StringPiece;


bazel-io:

src/google/protobuf/stubs/logging.h
```diff
@@ -59,6 +59,19 @@ enum LogLevel {
...
+  // wingdi.h defines ERROR to be 0, so `GOOGLE_LOG(ERROR, ...)` expands
+  // into `GOOGLE_LOG(0, ...)` which then expands into
+  // `someGoogleLogging(LOGLEVEL_0, ...)`. This is not ideal, because the
+  // GOOGLE_LOG macro expects to expand itself into
+  // `someGoogleLogging(LOGLEVEL_ERROR, ...)` instead. The workaround to
+  // get everything building is to simply define LOGLEVEL_0 as
+  // LOGLEVEL_ERROR and also define ERROR the same way that the Windows
+  // SDK does for consistency.
+#define ERROR 0
+  , LOGLEVEL_0 = LOGLEVEL_ERROR
```
xfxyjwf: How does wingdi.h get included in protobuf code?

src/google/protobuf/stubs/logging.h
```diff
@@ -59,6 +59,19 @@ enum LogLevel {
...
+
+#ifdef _WIN32
+  // wingdi.h defines ERROR to be 0, so `GOOGLE_LOG(ERROR, ...)` expands
+  // into `GOOGLE_LOG(0, ...)` which then expands into
+  // `someGoogleLogging(LOGLEVEL_0, ...)`. This is not ideal, because the
+  // GOOGLE_LOG macro expects to expand itself into
+  // `someGoogleLogging(LOGLEVEL_ERROR, ...)` instead. The workaround to
+  // get everything building is to simply define LOGLEVEL_0 as
+  // LOGLEVEL_ERROR and also define ERROR the same way that the Windows
+  // SDK does for consistency.
```
pherl: Would you be able to wrap the wingdi include instead? Something like:
```c++
// safe_wingdi.h
#include <wingdi.h>
#undef ERROR
```
You can then redefine ERROR when all the headers are included in your translation unit.

src/google/protobuf/stubs/logging.h
fmarier: It gets included via `Windows.h`.

pherl: Would defining NOGDI work for your project?

fmarier: Thanks for that idea. I defined `NOGDI` in our build files and it now works fine without this patch.

That seems like a much better work-around. If the logging system is going to be replaced, then there's not much point in merging this pull request.

pherl:
Closing this for now.

PR #3747, opened 2017-10-12
Add #ifdef AtomicWord typedef to fix build on OpenBSD/i386

fmarier:
This is a patch that Firefox carries in order to successfully build protobuf on 32-bit OpenBSD.

It originally comes from <https://bugzilla.mozilla.org/show_bug.cgi?id=1192556>.

### DIFF
--- a/src/google/protobuf/stubs/atomicops.h
+++ b/src/google/protobuf/stubs/atomicops.h
@@ -82,7 +82,11 @@ namespace internal {
 
 // Use AtomicWord for a machine-sized pointer.  It will use the Atomic32 or
 // Atomic64 routines below, depending on your architecture.
+#if defined(__OpenBSD__) && !defined(GOOGLE_PROTOBUF_ARCH_64_BIT)
+typedef Atomic32 AtomicWord;
+#else
 typedef intptr_t AtomicWord;
+#endif
 
 // Atomically execute:
 //      result = *ptr;


bazel-io:

pherl:
This change looks like no-op to me. Could you verify if this is still needed after: https://github.com/google/protobuf/commit/1b423474fd8ec6beab8d51bbc73b7756960ce2d6

fmarier:
You're right, the patch is essentially a no-op. I hadn't noticed that this code had changed since the commit actually made it to master before the 3.4.1 release date but [didn't make it to the 3.4 branch](https://github.com/google/protobuf/compare/v3.4.1...master).

On 3.4.1, we have the following:

    // 32-bit
    typedef int32 Atomic32;
    typedef intptr_t AtomicWord;

    // 64-bit
    typedef int32 Atomic32;
    typedef intptr_t Atomic64;
    typedef intptr_t AtomicWord;

whereas on master, we have:

    // 32-bit
    typedef intptr_t Atomic32;
    typedef int64 Atomic64;
    typedef intptr_t AtomicWord;

    // 64-bit
    typedef int32 Atomic32;
    typedef intptr_t Atomic64;
    typedef intptr_t AtomicWord;

and the OpenBSD patch made 3.4.1 look like this:

    // 32-bit
    typedef int32 Atomic32;
    typedef Atomic32 AtomicWord;

    // 64-bit
    typedef int32 Atomic32;
    typedef intptr_t Atomic64;
    typedef intptr_t AtomicWord;

So essentially OpenBSD changed `AtomicWord` to be an alias for `int32` instead of `intptr_t` on 32-bit platforms.

On master, `AtomicWord` (on 32-bit platforms) is still an alias for `intptr_t` but the big difference is that `Atomic32` is also aliased to `intptr_t`. That's different from OpenBSD patch, but since `Atomic32` and `AtomicWord` map to the same type, it probably means that it will work fine.

Let's close this no-op PR (sorry about that) and I'll reopen a new one if this is still causing issues once https://github.com/google/protobuf/commit/1b423474fd8ec6beab8d51bbc73b7756960ce2d6 makes it to a release.

PR #3764, opened 2017-10-17
Add the proto-lens Haskell library to the 3rd party doc.

zearen:


bazel-io:

PR #3767, opened 2017-10-18
Documentation fix regarding release file names

kdubovikov:
There is a typo in the readme


bazel-io:

pherl:
The README text in that section did mean the protoc-xxx packages which contains an executable binary of the protocol compiler. We do have the protobuf-xxx packages contain the source code of the distribution though.

kdubovikov:
@pherl, I haven't found the "protoc" packages on releases page at the first time because release 3.4.1 does not contain them, but they are present for 3.4.0. Thanks!

PR #3768, opened 2017-10-18
Update io_win32_unittest.cc

ybsave:
Fix the same issue mentioned in #3533.


bazel-io:

ybsave:
@googlebot already signed

pherl:
There are also non C++11 usages that we need to remove as well. I have sent out https://github.com/google/protobuf/pull/3769

pherl:
It should have been fixed in #3769 Thanks for the patch though!

PR #3778, opened 2017-10-21
remove -lpthread for Android based on PLATFORM=android define

spinorx:
This is needed because not all Android builds use //external:android/crosstool
For example one can use --android_crosstool_top too.

Without this:
external/android_ndk/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lpthread
external/android_ndk/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lpthread
external/android_ndk/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lpthread
clang: error: linker command failed with exit code 1 (use -v to see invocation)

With this and passing --define PLATFORM=android,  protobuf compiles fine.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -44,9 +44,19 @@ config_setting(
     },
 )
 
+# Android build need not always use //external:android/crosstool
+# for example one can use --android_crosstool_top too.
+config_setting(
+    name = "android_platform",
+    values = {
+      "define": "PLATFORM=android",
+    },
+)
+
 # Android and Windows builds do not need to link in a separate pthread library.
 LINK_OPTS = select({
     ":android": [],
+    ":android_platform": [],
     ":windows": [],
     ":windows_msvc": [],
     "//conditions:default": ["-lpthread", "-lm"],


xfxyjwf:
This doesn't seem the right approach to take. I would prefer a standard config setting provided by bazel for android, rather than a protobuf-specific command line flag. For example, if some other library has this same issue, decides to use the same approach but just happens to choose a different flag name (e.g., "TARGET_PLATFORM=android"), are you going to ask people using both protobuf and that library to pass two flags to build for android?
```
$ bazel --define PLATFORM=android --define TARGET_PLATFORM=android ... ?
```

spinorx:
Bazel does not seem to have standard way to detect android.  The crosstool_top as I pointed out is incorrect.  I have asked https://github.com/bazelbuild/bazel/issues/3924.  I am happy with any solution that works.  Would be unfortunate if we block usage without any solution though.

spinorx:
Note I have similar android specific changes needed in grpc (For ares) and boringssl (which also passes -pthread).  So finding a consistent way for all these projects will be good.

xfxyjwf:
I think we need a solution that will be endorsed/publicized by the bazel team. I'm concerned that the "--define PLATFORM=android" solution contradicts bazel team's recommendation for android and results in a weird bazel user interface that only exists in protobuf.

Also I'm amendable to change the config_setting(name="android") to whatever form recommended by the bazel team, but having both config_setting(name="android") and config_setting(name="andorid_platform") is just confusing.

Please consult the bazel team for the right way to detect android build instead. I'll follow the linked bazel issue https://github.com/bazelbuild/bazel/issues/3924 as well.

spinorx:
So what is the soltution till bazel team comes up with full solution.  I am ok with it there is some solution that i overlooked.  But if it is going to take few months for perfect solution to come out what would be the way out?

xfxyjwf:
That's really a question you should be asking the bazel team. They should have a short-term solution and long-term solution.

PR #3785, opened 2017-10-23
Make Status::error_message() return const string&.

ondrasej:
This change should be safe for most users thanks to the automatic construction of StringPiece out of std::string&.


xfxyjwf:
What problem is this PR intended to solve?

ondrasej:
As discussed with @xfxyjwf in person - we'll keep the API as is and abandon the PR.

PR #3791, opened 2017-10-24
To be clear that we set a new variable

signalwerk:


signalwerk:
Yo Mr. Roboto I signed it!

acozzette:
@signalwerk Thanks for the pull request!

PR #3792, opened 2017-10-24
Fix merging with message-valued oneof

### LINKED ISSUE #3200: Merging oneof fields in C# replaces
As reported on the mailing list

If two message use the same oneof case, then when they're merged, the values of those oneofs should be merged too - currently the value being merged simply replaces the previous one.

jskeet:
If messages A and B have the same oneof case, which is a message
type, and we merge B into A, those sub-messages should be merged.

Fixes #3200.

Note that I haven't regenerated all the code, as some of the protos
have been changed, breaking generation.

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_message_field.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_message_field.cc
@@ -171,6 +171,14 @@ void MessageOneofFieldGenerator::GenerateMembers(io::Printer* printer) {
     "}\n");
 }
 
+void MessageOneofFieldGenerator::GenerateMergingCode(io::Printer* printer) {
+  printer->Print(variables_, 
+    "if ($property_name$ == null) {\n"
+    "  $property_name$ = new $type_name$();\n"
+    "}\n"
+    "$property_name$.MergeFrom(other.$property_name$);\n");
+}
+
 void MessageOneofFieldGenerator::GenerateParsingCode(io::Printer* printer) {
   // TODO(jonskeet): We may be able to do better than this
   printer->Print(
--- a/src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
@@ -196,6 +196,10 @@ void PrimitiveOneofFieldGenerator::GenerateMembers(io::Printer* printer) {
       "}\n");
 }
 
+void PrimitiveOneofFieldGenerator::GenerateMergingCode(io::Printer* printer) {
+  printer->Print(variables_, "$property_name$ = other.$property_name$;\n");
+}
+
 void PrimitiveOneofFieldGenerator::WriteToString(io::Printer* printer) {
   printer->Print(variables_,
     "PrintField(\"$descriptor_name$\", $has_property_check$, $oneof_name$_, writer);\n");
--- a/csharp/protos/unittest_issues.proto
+++ b/csharp/protos/unittest_issues.proto
@@ -124,3 +124,18 @@ message TestJsonName {
   string description = 2 [json_name = "desc"];
   string guid = 3 [json_name = "exid"];
 }
+
+// Issue 3200: When merging two messages which use the same
+// oneof case, which is itself a message type, the submessages should
+// be merged.
+message OneofMerging {
+  message Nested {
+    int32 x = 1;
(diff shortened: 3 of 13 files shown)


src/google/protobuf/compiler/csharp/csharp_message_field.cc
```diff
@@ -171,6 +171,14 @@ void MessageOneofFieldGenerator::GenerateMembers(io::Printer* printer) {
     "}\n");
 }
 
+void MessageOneofFieldGenerator::GenerateMergingCode(io::Printer* printer) {
```
anandolee: I think the original GenerateMergingCode method on line 19 is correct. This one has problems same with PrimitiveOneofFieldGenerator I mentioned bellow. I think the reason of the issue is somewhere else cleared the message field

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
```diff
@@ -196,6 +196,10 @@ void PrimitiveOneofFieldGenerator::GenerateMembers(io::Printer* printer) {
       "}\n");
 }
 
+void PrimitiveOneofFieldGenerator::GenerateMergingCode(io::Printer* printer) {
+  printer->Print(variables_, "$property_name$ = other.$property_name$;\n");
```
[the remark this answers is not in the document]
jskeet: No, because this code is only called if the oneof *is* of this case. It's in a switch statement.

src/google/protobuf/compiler/csharp/csharp_message_field.cc
jskeet: No, the original code checks `other` without any need - because again, with the change I've made, the oneof merge code is *only* called when `other` is already for the field in question.

The issue is because prior to this PR, the merge for a `oneof` field was just an assignment, with no merging. It's now a simple assignment in some cases, and a merge where appropriate.

anandolee: Oh you are right I missed your change of csharp_message.cc which is the real cause of the issue. For none oneof fields, can we change the check of other into csharp_meesge.cc too? Then oneof and normal fields can share the GernerateMergingCode  which is same with what we c++ code does.

jskeet: I'll have a look and see what I can do. I think I'll at least keep that as two separate commits - I'd like that step to be a pure refactoring that doesn't affect the generated code. (It's probably worth waiting until we've got #3798 in before this one, too, so that we can regenerate all the code at the same time.)

jskeet: Hmm. I think the C++ code is different already because the message fields aren't pointers, but I could be very wrong. (I have very little experience with C++ in general, and C++ protos in particular.)

I note that the Java code looks (a bit) like the C# code after this commit:

```java
void ImmutableMessageFieldGenerator::
GenerateMergingCode(io::Printer* printer) const {
  printer->Print(variables_,
    "if (other.has$capitalized_name$()) {\n"
    "  merge$capitalized_name$(other.get$capitalized_name$());\n"
    "}\n");
}
...
void ImmutableMessageOneofFieldGenerator::
GenerateMergingCode(io::Printer* printer) const {
  printer->Print(variables_,
    "merge$capitalized_name$(other.get$capitalized_name$());\n");
}
```

jskeet: (Ditto for primitive fields in Java.)

anandolee: LGTM as it is similar with Java

anandolee:
ok to test

jskeet:
@anandolee: I don't know enough about the process to know what the "ok to test" means - we currently have the tests having run and only the JRuby failure. Does that mean it's okay to merge?

We might also want to consider merging #3798 first, so we can regenerate *everything* rather than just the few test protos that were regenerated here.

anandolee:
'ok to test' triggers more tests. We only run part of the tests for PRs at the begging to save time. Once the PR looks good, add 'ok to tests' to trigger the whole tests.

Yes, I thinks this one is ok to merge as JRuby failure is not related

jskeet:
Okay, will merge this, and we can regen again once #3798 is in.

PR #3798, opened 2017-10-25
Fixing C# codegen scripts and test protos

jskeet:
Fixes the C# codegen issues I ran into yesterday. Importantly:

- Moves all the test protos into a C#-specific directory so they won't be changed without warning
- Runs the codegen as part of the build


jskeet:
Good - first commit generates this error on Travis, as expected:

```text
google/protobuf/map_unittest_proto3.proto: "ForeignMessage" is not defined.
google/protobuf/map_unittest_proto3.proto: "TestAllTypes" is not defined.
google/protobuf/map_unittest_proto3.proto: "ForeignMessage" is not defined.
google/protobuf/map_unittest_proto3.proto: warning: Import google/protobuf/unittest_proto3.proto but not used.
```

jskeet:
Yup, looks like everything that's expected to work still works - PTAL :)

jskeet:
Have rebased this to master and regenerated the code (as #3792 changed codegen as well).

@anandolee PTAL

anandolee:
I'd prefer not copy those files into C#. When those files are changed, it most likely adding new features. Runs the codegen as part of the build in tests.sh is good

jskeet:
Will take this up on the internal mail thread - there are definitely pros and cons to any solution, but I reckoned that detaching the C# tests from the C++ tests would be the best option.

jskeet:
Have restricted this change to only forking unittest_proto3.proto now. PTAL.

anandolee:
ok to test.

LGTM. Thanks for the fixing

csharp/protos/unittest_proto3.proto
```diff
@@ -0,0 +1,391 @@
...
+//
+// A proto file we will use for unit testing.
+
+syntax = "proto3";
+
+// Some generic_services option(s) added automatically.
+// See:  http://go/proto2-generic-services-default
+option cc_generic_services = true;     // auto-added
+option java_generic_services = true;   // auto-added
+option py_generic_services = true;     // auto-added
```
anandolee: Maybe remove those options which does not related? The original .proto file has removed them too

jskeet: Happy to do that, but could it be in a separate PR, to avoid to much churn on this? (Given that this is in multiple commits, each change is relatively time-consuming.)

jskeet:
The only unexpected CI failure now is in Ruby, which doesn't look like a problem I would have caused:

```text
File does not exist: google/protobuf
```

Okay to merge?

anandolee:
sound good to me to have a separate PR do the option removal .

I think it is OK to merge

jskeet:
Fantastic, thanks - merging now, and will create a new PR tomorrow (and rebase the one about double.NaN).

PR #3802, opened 2017-10-25
Perform bitwise floating point equality comparisons in C#

jskeet:
(This builds on and requires #3798; it will just be 3 commits once that's merged.)


jskeet:
Now rebased onto master, and ready to review. (It'll be easiest to review one commit at a time.) PTAL.

jskeet:
Rebased from master with all the test proto changes etc. PTAL.

jskeet:
@anandolee: Ping for review on this? I'd love to get it in before more changes happen...

csharp/src/Google.Protobuf/Collections/ProtobufEqualityComparers.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            public override bool Equals(double? x, double? y) =>
+                x == null && y == null ? true
+                : x == null || y == null ? false
+                : BitwiseDoubleEqualityComparer.Equals(x.Value, y.Value);
+
+            // The hash code for null is just a constant which is at least *unlikely* to be used
+            // elsewhere. (Compared with 0, say.)
+            public override int GetHashCode(double? obj) =>
+                obj == null ? 293864 : BitwiseDoubleEqualityComparer.GetHashCode(obj.Value);
```
anandolee: 293864 seems too odd to me and it has been used twice(line 114 &127). Do they need to be the same? If so, define a const?

jskeet: They don't need to be the same, but they don't need to be different. They're just arbitrary values. I could make them different if you want...

anandolee: No need

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -51,6 +51,7 @@ public sealed class RepeatedField<T> : IList<T>, IList, IDeepCloneable<RepeatedF
         , IReadOnlyList<T>
 #endif
     {
+        private static readonly EqualityComparer<T> EqualityComparer = ProtobufEqualityComparers.GetEqualityComparer<T>();
```
anandolee: I didn't see the whole file. But can't we use the name comparer here and remove line 438 and 458?

jskeet: The point of those lines is to copy the static field value into a local variable *once* rather than on each iteration.

anandolee:
ok to test

jskeet:
The Travis failures are all either allowed or ruby21 (which seems to have been failing for a while).

Okay to merge?

jskeet:
@anandolee: Ping?

anandolee:
Yes, it is good to merge

jskeet:
Rebased from master as github was reluctant to do so - will merge on "green enough" :)

PR #3812, opened 2017-10-27
Minor fixes in code improvements.

Carmot:
Unusued variables, variable assigned twice, conditions never met or always having the same value.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message_field.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message_field.cc
@@ -972,9 +972,6 @@ GenerateInlineAccessorDefinitions(io::Printer* printer) const {
       "  // @@protoc_insertion_point(field_add:$full_name$)\n"
       "  return $name$_.Add();\n"
       "}\n");
-  }
-
-  if (!dependent_field_) {
     printer->Print(variables_,
       "inline ::google::protobuf::RepeatedPtrField< $type$ >*\n"
       "$classname$::mutable_$name$() {\n"
--- a/src/google/protobuf/compiler/java/java_primitive_field_lite.cc
+++ b/src/google/protobuf/compiler/java/java_primitive_field_lite.cc
@@ -146,7 +146,7 @@ void SetPrimitiveVariables(const FieldDescriptor* descriptor,
 
     (*variables)["is_field_present_message"] = GenerateGetBit(messageBitIndex);
   } else {
-    (*variables)["set_has_field_bit_message"] = "";
+    (*variables)["get_has_field_bit_message"] = "";
     (*variables)["set_has_field_bit_message"] = "";
     (*variables)["clear_has_field_bit_message"] = "";
 
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -109,7 +109,7 @@ string UnderscoresToCamelCase(const string& input, bool first_capitalized) {
         current = "";
       }
       current += c;
-      last_char_was_number = last_char_was_lower = last_char_was_upper = false;
+      last_char_was_lower = last_char_was_upper = false;
       last_char_was_number = true;
     } else if (ascii_islower(c)) {
       // lowercase letter can follow a lowercase or uppercase letter
@@ -118,7 +118,7 @@ string UnderscoresToCamelCase(const string& input, bool first_capitalized) {
         current = "";
       }
       current += c;  // already lower
-      last_char_was_number = last_char_was_lower = last_char_was_upper = false;
+      last_char_was_number = last_char_was_upper = false;
       last_char_was_lower = true;
     } else if (ascii_isupper(c)) {
       if (!last_char_was_upper) {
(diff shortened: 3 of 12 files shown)


Carmot:
I signed it!

2017-10-27 11:29 GMT+02:00 googlebot <notifications@github.com>:

acozzette:
Ok to test (this comment will trigger the other CI runs)

src/google/protobuf/unknown_field_set.cc
```diff
@@ -205,23 +205,26 @@ void UnknownFieldSet::AddField(const UnknownField& field) {
...
-  }
-  // Pop off the # of deleted fields.
-  for (int i = 0; i < num; ++i) {
-    fields_->pop_back();
-  }
-  if (fields_ && fields_->size() == 0) {
-    // maintain invariant: never hold fields_ if empty.
-    delete fields_;
-    fields_ = NULL;
+  if (fields_ != NULL) {
```
acozzette: Actually I think you may be able to remove this null check altogether. We always call `fields_->size()`  on line 213 anyway, so we already seem to be assuming that `fields_` is non-null.

pherl:
I lot of this redundant code is because we stripped internal code during the opensource release. In the future, we plan to also opensource those part so the two repo will be more aligned, and it also makes the integration easier.

src/google/protobuf/compiler/subprocess.cc
```diff
@@ -335,11 +335,9 @@ void Subprocess::Start(const string& program, SearchMode search_mode) {
 
     // Write directly to STDERR_FILENO to avoid stdio code paths that may do
     // stuff that is unsafe here.
-    int ignored;
```
pherl: We have internal compilation tools to ensure the return value of this function must not be ignored, thus this redundant code.

src/google/protobuf/generated_message_reflection.cc
```diff
@@ -1124,9 +1124,6 @@ string GeneratedMessageReflection::GetString(
         return GetField<ArenaStringPtr>(message, field).Get();
       }
     }
-
```
pherl: same here for compilation checks.

src/google/protobuf/unknown_field_set.cc
```diff
@@ -222,6 +222,7 @@ void UnknownFieldSet::DeleteSubrange(int start, int num) {
     delete fields_;
     fields_ = NULL;
   }
+  return;
```
pherl: remove this one?

src/google/protobuf/util/internal/utility.cc
```diff
@@ -353,21 +353,15 @@ bool IsMap(const google::protobuf::Field& field,
              google::protobuf::Field_Cardinality_CARDINALITY_REPEATED &&
          (GetBoolOptionOrDefault(type.options(), "map_entry", false) ||
           GetBoolOptionOrDefault(type.options(),
-                                 "google.protobuf.MessageOptions.map_entry", false) ||
-          GetBoolOptionOrDefault(type.options(),
```
pherl: this was due to the different package name we use in google. One of the symbol is actually the internal code names in google, and we converted to google.protobuf duration opensource release. Please revert this.

src/google/protobuf/util/internal/utility.cc
```diff
@@ -353,21 +353,15 @@ bool IsMap(const google::protobuf::Field& field,
...
+                                 "google.protobuf.MessageOptions.map_entry", false));
 }
 
 bool IsMessageSetWireFormat(const google::protobuf::Type& type) {
   return GetBoolOptionOrDefault(type.options(), "message_set_wire_format",
                                 false) ||
          GetBoolOptionOrDefault(type.options(),
                                 "google.protobuf.MessageOptions.message_set_wire_format",
-                                false) ||
-         GetBoolOptionOrDefault(
```
pherl: dtto

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -118,15 +118,15 @@ string UnderscoresToCamelCase(const string& input, bool first_capitalized) {
         current = "";
       }
       current += c;  // already lower
-      last_char_was_number = last_char_was_lower = last_char_was_upper = false;
```
pherl: @thomasvl to take a look at the change to this file.

src/google/protobuf/compiler/java/java_primitive_field_lite.cc
```diff
@@ -146,7 +146,7 @@ void SetPrimitiveVariables(const FieldDescriptor* descriptor,
 
     (*variables)["is_field_present_message"] = GenerateGetBit(messageBitIndex);
   } else {
-    (*variables)["set_has_field_bit_message"] = "";
```
pherl: Could you also fix this in the "lite" branch?

src/google/protobuf/compiler/cpp/cpp_message_field.cc
```diff
@@ -972,9 +972,6 @@ GenerateInlineAccessorDefinitions(io::Printer* printer) const {
       "  // @@protoc_insertion_point(field_add:$full_name$)\n"
       "  return $name$_.Add();\n"
       "}\n");
-  }
-
-  if (!dependent_field_) {
```
pherl: This section has been changed dramatically in the internal branch. And this part no longer exist in the other repo. Could you revert this? It will make the conflict resolve easier.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: These are fine, the first line in each was clearing all state for consistency, but it is an ignored assignment because of the following lines.

PR #3824, opened 2017-10-31
[Java] Add a UTF-8 decoder that uses Unsafe to directly decode a byte buffer.

anuraaga:
While debugging performance of `UnsafeDirectNioDecoder` vs `ArrayDecoder`, I found that reading strings is a hotspot - profiles specifically showed `isValidUtf8` as a hotspot, and I think this is because the small string copies made while decoding a byte buffer cause poor utilization of the CPU cache compared to operating entirely on a contiguous heap array. The effect of this is that it is usually better to first copy an entire `ByteBuffer` into a `byte[]` and use `ArrayDecoder` for real-world protos that will have some strings - `UnsafeDirectNioDecoder` could pretty much never beat `ArrayDecoder` if there is any reasonable amount of string content in the proto.

This change adds a custom decoding path that works directly on byte buffers without copying onto the heap. The decoding code path is modeled after the current isValidUtf8 with actual codepoint calculation inserted. Performance goes up a lot and allows `UnsafeDirectNioDecoder` to be viable, especially when the strings are ASCII which is very common. This is nice for Netty users (i.e., gRPC), which usually work with direct buffers.

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/CodedInputStream.java
+++ b/java/core/src/main/java/com/google/protobuf/CodedInputStream.java
@@ -64,6 +64,14 @@ public abstract class CodedInputStream {
   // Integer.MAX_VALUE == 0x7FFFFFF == INT_MAX from limits.h
   private static final int DEFAULT_SIZE_LIMIT = Integer.MAX_VALUE;
 
+  /**
+   * Whether to enable our custom UTF-8 decode codepath which does not use {@link StringCoding}.
+   * Enabled by default, disable by setting
+   * {@code -Dcom.google.protobuf.enableCustomutf8Decode=false} in JVM args.
+   */
+  private static final boolean ENABLE_CUSTOM_UTF8_DECODE
+      = !"false".equals(System.getProperty("com.google.protobuf.enableCustomUtf8Decode"));
+
   /** Visible for subclasses. See setRecursionLimit() */
   int recursionDepth;
 
@@ -825,13 +833,19 @@ public abstract class CodedInputStream {
     public String readStringRequireUtf8() throws IOException {
       final int size = readRawVarint32();
       if (size > 0 && size <= (limit - pos)) {
-        // TODO(martinrb): We could save a pass by validating while decoding.
-        if (!Utf8.isValidUtf8(buffer, pos, pos + size)) {
-          throw InvalidProtocolBufferException.invalidUtf8();
+        if (ENABLE_CUSTOM_UTF8_DECODE) {
+          String result = Utf8.decodeUtf8(buffer, pos, size);
+          pos += size;
+          return result;
+        } else {
+          // TODO(martinrb): We could save a pass by validating while decoding.
+          if (!Utf8.isValidUtf8(buffer, pos, pos + size)) {
+            throw InvalidProtocolBufferException.invalidUtf8();
+          }
+          final int tempPos = pos;
+          pos += size;
+          return new String(buffer, tempPos, size, UTF_8);
         }
-        final int tempPos = pos;
-        pos += size;
-        return new String(buffer, tempPos, size, UTF_8);
       }
 
       if (size == 0) {
@@ -1524,6 +1538,8 @@ public abstract class CodedInputStream {
       final int size = readRawVarint32();
       if (size > 0 && size <= remaining()) {
         // TODO(nathanmittler): Is there a way to avoid this copy?
+        // TODO(anuraaga): It might be possible to share the optimized loop with
+        // readStringRequireUtf8 by implementing Java replacement logic there.
         // The same as readBytes' logic
         byte[] bytes = new byte[size];
         UnsafeUtil.copyMemory(pos, bytes, 0, size);
@@ -1544,19 +1560,26 @@ public abstract class CodedInputStream {
     @Override
     public String readStringRequireUtf8() throws IOException {
       final int size = readRawVarint32();
-      if (size >= 0 && size <= remaining()) {
-        // TODO(nathanmittler): Is there a way to avoid this copy?
-        // The same as readBytes' logic
-        byte[] bytes = new byte[size];
-        UnsafeUtil.copyMemory(pos, bytes, 0, size);
-        // TODO(martinrb): We could save a pass by validating while decoding.
-        if (!Utf8.isValidUtf8(bytes)) {
-          throw InvalidProtocolBufferException.invalidUtf8();
-        }
+      if (size > 0 && size <= remaining()) {
+        if (ENABLE_CUSTOM_UTF8_DECODE) {
+          final int bufferPos = bufferPos(pos);
+          String result = Utf8.decodeUtf8(buffer, bufferPos, size);
+          pos += size;
+          return result;
+        } else {
+          // TODO(nathanmittler): Is there a way to avoid this copy?
+          // The same as readBytes' logic
+          byte[] bytes = new byte[size];
+          UnsafeUtil.copyMemory(pos, bytes, 0, size);
+          // TODO(martinrb): We could save a pass by validating while decoding.
+          if (!Utf8.isValidUtf8(bytes)) {
+            throw InvalidProtocolBufferException.invalidUtf8();
+          }
 
-        String result = new String(bytes, UTF_8);
-        pos += size;
-        return result;
+          String result = new String(bytes, UTF_8);
+          pos += size;
+          return result;
+        }
       }
 
       if (size == 0) {
@@ -2324,11 +2347,15 @@ public abstract class CodedInputStream {
         bytes = readRawBytesSlowPath(size);
         tempPos = 0;
       }
-      // TODO(martinrb): We could save a pass by validating while decoding.
-      if (!Utf8.isValidUtf8(bytes, tempPos, tempPos + size)) {
-        throw InvalidProtocolBufferException.invalidUtf8();
+      if (ENABLE_CUSTOM_UTF8_DECODE) {
+        return Utf8.decodeUtf8(bytes, tempPos, size);
+      } else {
+        // TODO(martinrb): We could save a pass by validating while decoding.
+        if (!Utf8.isValidUtf8(bytes, tempPos, tempPos + size)) {
+          throw InvalidProtocolBufferException.invalidUtf8();
+        }
+        return new String(bytes, tempPos, size, UTF_8);
       }
-      return new String(bytes, tempPos, size, UTF_8);
     }
 
     @Override
@@ -3348,23 +3375,34 @@ public abstract class CodedInputStream {
     public String readStringRequireUtf8() throws IOException {
       final int size = readRawVarint32();
       if (size > 0 && size <= currentByteBufferLimit - currentByteBufferPos) {
-        byte[] bytes = new byte[size];
-        UnsafeUtil.copyMemory(currentByteBufferPos, bytes, 0, size);
-        if (!Utf8.isValidUtf8(bytes)) {
-          throw InvalidProtocolBufferException.invalidUtf8();
+        if (ENABLE_CUSTOM_UTF8_DECODE) {
+          final int bufferPos = (int) (currentByteBufferPos - currentByteBufferStartPos);
+          String result = Utf8.decodeUtf8(currentByteBuffer, bufferPos, size);
+          currentByteBufferPos += size;
+          return result;
+        } else {
+          byte[] bytes = new byte[size];
+          UnsafeUtil.copyMemory(currentByteBufferPos, bytes, 0, size);
+          if (!Utf8.isValidUtf8(bytes)) {
+            throw InvalidProtocolBufferException.invalidUtf8();
+          }
+          String result = new String(bytes, UTF_8);
+          currentByteBufferPos += size;
+          return result;
         }
-        String result = new String(bytes, UTF_8);
-        currentByteBufferPos += size;
-        return result;
       }
       if (size >= 0 && size <= remaining()) {
         byte[] bytes = new byte[size];
         readRawBytesTo(bytes, 0, size);
-        if (!Utf8.isValidUtf8(bytes)) {
-          throw InvalidProtocolBufferException.invalidUtf8();
+        if (ENABLE_CUSTOM_UTF8_DECODE) {
+          return Utf8.decodeUtf8(bytes, 0, size);
+        } else {
+          if (!Utf8.isValidUtf8(bytes)) {
+            throw InvalidProtocolBufferException.invalidUtf8();
+          }
+          String result = new String(bytes, UTF_8);
+          return result;
         }
-        String result = new String(bytes, UTF_8);
-        return result;
       }
 
       if (size == 0) {
--- a/java/core/src/main/java/com/google/protobuf/UnsafeUtil.java
+++ b/java/core/src/main/java/com/google/protobuf/UnsafeUtil.java
@@ -33,7 +33,6 @@ package com.google.protobuf;
 import java.lang.reflect.Field;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
 import java.util.logging.Level;
@@ -71,6 +70,8 @@ final class UnsafeUtil {
   private static final long OBJECT_ARRAY_INDEX_SCALE = arrayIndexScale(Object[].class);
 
   private static final long BUFFER_ADDRESS_OFFSET = fieldOffset(bufferAddressField());
+
+  private static final long STRING_VALUE_OFFSET = fieldOffset(stringValueField());
 
   private UnsafeUtil() {}
 
@@ -259,6 +260,26 @@ final class UnsafeUtil {
     return MEMORY_ACCESSOR.getLong(buffer, BUFFER_ADDRESS_OFFSET);
   }
 
+  /**
+   * Returns a new {@link String} backed by the given {@code chars}. The char array should not
+   * be mutated any more after calling this function.
+   */
+  static String moveToString(char[] chars) {
+    if (STRING_VALUE_OFFSET == -1) {
+      // In the off-chance that this JDK does not implement String as we'd expect, just do a copy.
+      return new String(chars);
+    }
+    final String str;
+    try {
+      str = (String) UNSAFE.allocateInstance(String.class);
+    } catch (InstantiationException e) {
+      // This should never happen, but return a copy as a fallback just in case.
+      return new String(chars);
+    }
+    putObject(str, STRING_VALUE_OFFSET, chars);
+    return str;
+  }
+
   static Object getStaticObject(Field field) {
     return MEMORY_ACCESSOR.getStaticObject(field);
   }
@@ -375,7 +396,12 @@ final class UnsafeUtil {
 
   /** Finds the address field within a direct {@link Buffer}. */
   private static Field bufferAddressField() {
-    return field(Buffer.class, "address");
+    return field(Buffer.class, "address", long.class);
+  }
+
+  /** Finds the value field within a {@link String}. */
+  private static Field stringValueField() {
+    return field(String.class, "value", char[].class);
   }
 
   /**
@@ -390,11 +416,14 @@ final class UnsafeUtil {
    * Gets the field with the given name within the class, or {@code null} if not found. If found,
    * the field is made accessible.
    */
-  private static Field field(Class<?> clazz, String fieldName) {
+  private static Field field(Class<?> clazz, String fieldName, Class<?> expectedType) {
     Field field;
     try {
       field = clazz.getDeclaredField(fieldName);
       field.setAccessible(true);
+      if (!field.getType().equals(expectedType)) {
+        return null;
+      }
     } catch (Throwable t) {
       // Failed to access the fields.
       field = null;
--- a/java/core/src/main/java/com/google/protobuf/Utf8.java
+++ b/java/core/src/main/java/com/google/protobuf/Utf8.java
@@ -34,11 +34,15 @@ import static com.google.protobuf.UnsafeUtil.addressOffset;
 import static com.google.protobuf.UnsafeUtil.hasUnsafeArrayOperations;
 import static com.google.protobuf.UnsafeUtil.hasUnsafeByteBufferOperations;
 import static java.lang.Character.MAX_SURROGATE;
+import static java.lang.Character.MIN_HIGH_SURROGATE;
+import static java.lang.Character.MIN_LOW_SURROGATE;
+import static java.lang.Character.MIN_SUPPLEMENTARY_CODE_POINT;
 import static java.lang.Character.MIN_SURROGATE;
 import static java.lang.Character.isSurrogatePair;
 import static java.lang.Character.toCodePoint;
 
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 
 /**
  * A set of low-level, high-performance static utility methods related
@@ -289,7 +293,7 @@ final class Utf8 {
         if (Character.MIN_SURROGATE <= c && c <= Character.MAX_SURROGATE) {
           // Check that we have a well-formed surrogate pair.
           int cp = Character.codePointAt(sequence, i);
-          if (cp < Character.MIN_SUPPLEMENTARY_CODE_POINT) {
+          if (cp < MIN_SUPPLEMENTARY_CODE_POINT) {
             throw new UnpairedSurrogateException(i, utf16Length);
           }
           i++;
@@ -328,6 +332,26 @@ final class Utf8 {
    */
   static int partialIsValidUtf8(int state, ByteBuffer buffer, int index, int limit) {
     return processor.partialIsValidUtf8(state, buffer, index, limit);
+  }
+
+  /**
+   * Decodes the given UTF-8 portion of the {@link ByteBuffer} into a {@link String}.
+   *
+   * @throws InvalidProtocolBufferException if the input is not valid UTF-8.
+   */
+  static String decodeUtf8(ByteBuffer buffer, int index, int size)
+      throws InvalidProtocolBufferException {
+    return processor.decodeUtf8(buffer, index, size);
+  }
+
+  /**
+   * Decodes the given UTF-8 encoded byte array slice into a {@link String}.
+   *
+   * @throws InvalidProtocolBufferException if the input is not valid UTF-8.
+   */
+  static String decodeUtf8(byte[] bytes, int index, int size)
+      throws InvalidProtocolBufferException {
+    return processor.decodeUtf8(bytes, index, size);
   }
 
   /**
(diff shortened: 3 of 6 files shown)


java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -1554,5 +1767,93 @@ private static int unsafeIncompleteStateFor(long address, final int byte1, int r
     }
   }
 
+  /**
+   * Adaptation of http://bjoern.hoehrmann.de/utf-8/decoder/dfa/
+   *
+   * Copyright (c) 2008-2009 Bjoern Hoehrmann <bjoern@hoehrmann.de>
```
xfxyjwf: I'll consult our internal team about this, but I think we can't accept this code piece unless the author 'Bjoern Hoehrmann' has signed google CLA and explicitly acknowledge the inclusion of his work in protobuf.

anuraaga:
@xfxyjwf Thanks for the quick initial review - was hoping that including the license would be sufficient but realize now that's not the case. I went ahead and reimplemented the logic by looking at Wikipedia + current isValidUtf8 code, so there isn't any licensed code anymore.

Fun enough, this new stateless version is faster than the other one even on non-ASCII - it seems to be faster in all cases so I went ahead and wired up the `byte[]` input streams too. Hopefully this can provide a good speedup for proto3 users across the board. Replicating JDK replacement logic is very hairy, so I've left that out, but doing so would also speedup proto2 users which could be pretty significant for large-scale legacy deployments.

Also removed JMH.

anuraaga:
@xfxyjwf Friendly ping - curious if the new approach looks reasonable

xfxyjwf:
@pherl @nmittler Can you guys also help take a look at this Java UTF-8 decoding changes? It looks good to me but it's a large change that I don't feel comfortable just merging without someone else also reviewing.

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -610,6 +634,98 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
...
+          if (offset >= limit || byte1 < (byte) 0xC2) {
+            throw InvalidProtocolBufferException.invalidUtf8();
+          }
+          byte byte2 = buffer.get(offset++);
+          DecodeUtil.handleTwoBytes(byte1, byte2, resultArr, resultPos++);
+        } else if (byte1 < (byte) 0xF0) {
+          if (offset >= limit - 1) {
+            throw InvalidProtocolBufferException.invalidUtf8();
+          }
+          byte byte2 = buffer.get(offset++);
```
nmittler: It might be better to avoid the store operations and pass `buffer.get(offset++)` direction into the method call.

same comment throughout.

nmittler:
@anuraaga WRT JMH, I just saw the comment from @xfxyjwf ... I guess you'll follow this up with another PR for benchmarks?

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
```diff
@@ -825,13 +825,9 @@ public String readString() throws IOException {
     public String readStringRequireUtf8() throws IOException {
       final int size = readRawVarint32();
       if (size > 0 && size <= (limit - pos)) {
-        // TODO(martinrb): We could save a pass by validating while decoding.
```
pherl: Can we guard this change by a static boolean? That way it would be easier to roll back if there are performance regressions.

pherl:
OK to test

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
anuraaga: Added a constant that reads a system property. It means javac won't remove the branches but JVM will easily do so, but let me know if you'd prefer a compiler constant.

anuraaga:
Ping @xfxyjwf @nmittler @pherl for another look if you have a chance. Looking forward to the speedup :)

nmittler:
@anuraaga can you re-run your benchmark to verify that the recent changes don't impact perf?

anuraaga:
@nmittler Yup, been running it on pretty much every small change since addressing comments and there's been no performance change while extracting the methods. Looked at the assembly with jitwatch and confirmed that it all inlines nicely.

nmittler:
@anuraaga Thanks! The changes all LGTM ... great work!

@xfxyjwf can you confirm that there is no perf regression on Android?

anuraaga:
Ping @xfxyjwf - any chance to check on Android? :) To be honest, wasn't thinking of Android at all so not sure what might happen there, but I guess AoT compiling should inline everything similarly.

anuraaga:
Hey @xfxyjwf @nmittler @pherl - any updates on this? Let me know if it would be smoother to split the flag into two, one for bytebuffers, defaulted `true`, and one for arrays, defaulted `false`.

- For bytebuffers, the performance advantage is massive and makes `UnsafeDirectNioDecoder` practical to use, so it's very important - IMO, fixing the bytebuffer performance issue is a bug fix more than an improvement. It's also the use case I'm personally invested in and which resulted in this PR.

- For arrays, the performance advantage also seems significant, but admittedly the potential effect is larger as it is the standard code path when using protobuf. While I'd like to apply it so people can benefit (I guess Google servers using proto3 would benefit :D ), I can see the need to take more caution.

nmittler:
@xfxyjwf It would be great to get this in.  I think we just need to make sure this doesn't have a negative impact on Android.  Can you verify that quickly?

xfxyjwf:
I'll run our internal microbenchmarks against this PR today.

nmittler:
Awesome! Thanks, @xfxyjwf!

xfxyjwf:
Benchmark results from our internal micro-benchmarks are neutral (both android ones and server ones). See: http://cl//176177299

anuraaga:
@xfxjwf Thanks! Just wondering are those proto3 benchmarks are a mix of proto2 and proto3? This change only affects proto3 (except for the rare case where someone turns on utf8 validation in proto2) so would want to look only at those for a change one way or another.

anuraaga:
Hello - another checkin :) I wish I could look at the benchmark as I'm curious as well, but I guess restricting to proto3, and also separating out ByteBuffer, is important to lessen noise. Is there anything I can do to help make progress on this? As always, no problem limiting the change to just UnsafeDirectNioDecoder if it helps.

anuraaga:
Hey @xfxyjwf @nmittler @pherl - checking in again (forgot it's Thanksgiving in the US last week, sorry for pinging on the holidays).

Was hoping that this would make it to 3.5 and a bit sad it didn't, but understand it's a big change. Any way we can move forward though (and possibly get in to 3.6 :D)?

In case there's some lack of enthusiasm since the internal benchmarks are neutral, I'm almost certain it's because they're mostly proto2 which is unaffected by this change. Throw a proto3 web search or maps suggest proto into there and I'm sure you'd see a big difference ;) I wouldn't be surprised if internal users are reluctant to switch to proto3 due to string decoding performance based on the numbers I'm seeing. But as always, I'm perfectly happy to restrict the change to direct byte buffers, which desperately needs this change, if it's easier to make progress.

FWIW, here are the numbers for the benchmark I've been using, on Java8 and Java9. Java9 doesn't get the 0-copy string move optimization, but even then it seems to be on-par with JDK for ASCII and much faster elsewhere.

The most important comparisons are
- StringsMessageBenchmark.byteBuffer / StringsMessageBenchmark.byteBuffer_copyFirst / StringsMessageBenchmark.byteBuffer_old
- ByteBufferBenchmark.decodeDirectly / DirectByteBufferBenchmark.copyToHeapFirst 
- ArrayBenchmark.array_custom / ArrayBenchmark.array_jdk

Source: https://gist.github.com/anuraaga/f94119780438bded4209f453a8f85169

```
Java 8

Benchmark                                                         (byteBufferType)   (value)   Mode  Cnt         Score        Error  Units
Utf8DecodeBenchmark.AllTypesArrayBenchmark.array_custom                        N/A       N/A  thrpt   20   1035675.946 ±  14019.777  ops/s
Utf8DecodeBenchmark.AllTypesArrayBenchmark.array_jdk                           N/A       N/A  thrpt   20    517516.528 ±  10718.457  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A   ENGLISH  thrpt   20  11344041.452 ± 140782.185  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A  JAPANESE  thrpt   20   4454176.927 ±  56869.938  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A     MIXED  thrpt   20   3146504.578 ±  45848.406  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A    HEBREW  thrpt   20   5882934.372 ± 100397.262  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A   CHINESE  thrpt   20   4475807.173 ±  66273.370  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A   ENGLISH  thrpt   20   8839746.140 ± 139353.873  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A  JAPANESE  thrpt   20   2661279.910 ±  32935.258  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A     MIXED  thrpt   20   1857703.502 ±  26763.854  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A    HEBREW  thrpt   20   3347970.823 ±  49345.380  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A   CHINESE  thrpt   20   3022370.706 ± 113706.644  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT   ENGLISH  thrpt   20   9631185.803 ± 159909.352  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT  JAPANESE  thrpt   20   4435904.935 ±  53276.172  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT     MIXED  thrpt   20   3297931.635 ±  65039.965  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT    HEBREW  thrpt   20   5600922.434 ±  95066.874  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT   CHINESE  thrpt   20   5140227.259 ±  87286.966  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP   ENGLISH  thrpt   20  11238772.624 ± 165250.902  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP  JAPANESE  thrpt   20   4432397.379 ±  44467.814  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP     MIXED  thrpt   20   3143543.713 ±  80910.752  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP    HEBREW  thrpt   20   5930799.888 ±  80753.351  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP   CHINESE  thrpt   20   4374172.489 ±  64001.325  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A   ENGLISH  thrpt   20   6673743.550 ± 109527.715  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A  JAPANESE  thrpt   20   2394081.304 ±  30274.810  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A     MIXED  thrpt   20   1620800.857 ±  32531.092  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A    HEBREW  thrpt   20   3106074.752 ±  41766.242  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A   CHINESE  thrpt   20   2758085.489 ±  44238.833  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT   ENGLISH  thrpt   20    416999.717 ±  16860.979  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT  JAPANESE  thrpt   20    152614.148 ±   3191.066  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT     MIXED  thrpt   20    105724.100 ±   1597.169  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT    HEBREW  thrpt   20    211967.029 ±   4796.024  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT   CHINESE  thrpt   20    150352.506 ±   1498.043  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT   ENGLISH  thrpt   20    368173.971 ±  13488.251  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT  JAPANESE  thrpt   20    163459.826 ±   4069.326  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT     MIXED  thrpt   20    115647.488 ±   2749.487  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT    HEBREW  thrpt   20    204758.891 ±   5319.577  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT   CHINESE  thrpt   20    169024.764 ±   5210.278  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP   ENGLISH  thrpt   20    459181.956 ±   5178.505  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP  JAPANESE  thrpt   20    158686.655 ±   2280.379  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP     MIXED  thrpt   20    108419.500 ±   2143.423  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP    HEBREW  thrpt   20    223132.476 ±   3852.461  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP   CHINESE  thrpt   20    159317.360 ±   2051.436  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT   ENGLISH  thrpt   20    381548.205 ±  14642.620  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT  JAPANESE  thrpt   20    140823.909 ±   3801.774  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT     MIXED  thrpt   20    101037.087 ±   2867.023  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT    HEBREW  thrpt   20    199506.587 ±   6117.641  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT   CHINESE  thrpt   20    142748.718 ±   3096.335  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP   ENGLISH  thrpt   20    424750.212 ±   7060.122  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP  JAPANESE  thrpt   20    145176.969 ±  10440.393  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP     MIXED  thrpt   20    104181.653 ±   1632.933  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP    HEBREW  thrpt   20    207804.030 ±   8557.704  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP   CHINESE  thrpt   20    155530.760 ±   2465.193  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT   ENGLISH  thrpt   20    273278.066 ±  11323.380  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT  JAPANESE  thrpt   20     84605.972 ±   1433.077  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT     MIXED  thrpt   20     59695.944 ±    858.012  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT    HEBREW  thrpt   20    109256.426 ±   2135.230  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT   CHINESE  thrpt   20    101906.669 ±   3213.892  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP   ENGLISH  thrpt   20    377401.924 ±   6666.030  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP  JAPANESE  thrpt   20     95649.830 ±   1185.386  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP     MIXED  thrpt   20     65389.237 ±    919.510  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP    HEBREW  thrpt   20    125759.850 ±   2539.549  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP   CHINESE  thrpt   20    114718.268 ±   1671.434  ops/s

Java 9

Benchmark                                                         (byteBufferType)   (value)   Mode  Cnt         Score        Error  Units
Utf8DecodeBenchmark.AllTypesArrayBenchmark.array_custom                        N/A       N/A  thrpt   20    979233.661 ±  10012.953  ops/s
Utf8DecodeBenchmark.AllTypesArrayBenchmark.array_jdk                           N/A       N/A  thrpt   20    541470.347 ±   9336.800  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A   ENGLISH  thrpt   20   9996077.522 ± 125613.379  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A  JAPANESE  thrpt   20   4209367.121 ±  55791.397  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A     MIXED  thrpt   20   2883566.824 ±  50564.694  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A    HEBREW  thrpt   20   5685211.678 ±  67744.823  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_custom                                N/A   CHINESE  thrpt   20   4378542.460 ± 111577.429  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A   ENGLISH  thrpt   20  10687035.954 ± 157215.156  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A  JAPANESE  thrpt   20   2826209.172 ±  39353.465  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A     MIXED  thrpt   20   2025476.087 ±  21770.999  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A    HEBREW  thrpt   20   3477542.374 ±  44940.345  ops/s
Utf8DecodeBenchmark.ArrayBenchmark.array_jdk                                   N/A   CHINESE  thrpt   20   3325769.121 ±  42774.892  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT   ENGLISH  thrpt   20   8822377.021 ± 207093.640  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT  JAPANESE  thrpt   20   3997758.356 ± 452676.034  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT     MIXED  thrpt   20   3210459.933 ± 120734.598  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT    HEBREW  thrpt   20   5115177.256 ±  73427.771  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                      DIRECT   CHINESE  thrpt   20   4364487.868 ±  76792.000  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP   ENGLISH  thrpt   20  10105268.129 ± 124063.026  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP  JAPANESE  thrpt   20   4202349.863 ±  70464.257  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP     MIXED  thrpt   20   2992099.443 ±  52333.824  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP    HEBREW  thrpt   20   5603021.148 ± 105407.183  ops/s
Utf8DecodeBenchmark.ByteBufferBenchmark.decodeDirectly                        HEAP   CHINESE  thrpt   20   4315432.381 ±  63811.573  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A   ENGLISH  thrpt   20   8015844.856 ± 155062.216  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A  JAPANESE  thrpt   20   2440636.535 ±  32144.760  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A     MIXED  thrpt   20   1656699.029 ±  29489.713  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A    HEBREW  thrpt   20   2941237.104 ±  86575.047  ops/s
Utf8DecodeBenchmark.DirectByteBufferBenchmark.copyToHeapFirst                  N/A   CHINESE  thrpt   20   2840008.086 ±  46602.012  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT   ENGLISH  thrpt   20    380879.953 ±  18276.071  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT  JAPANESE  thrpt   20    142730.085 ±   1912.097  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT     MIXED  thrpt   20     99962.482 ±   1479.362  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT    HEBREW  thrpt   20    188939.494 ±   4516.471  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.array                           DIRECT   CHINESE  thrpt   20    159535.449 ±   1936.158  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT   ENGLISH  thrpt   20    357865.915 ±   8309.979  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT  JAPANESE  thrpt   20    154177.425 ±   2456.391  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT     MIXED  thrpt   20    112304.389 ±   3628.963  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT    HEBREW  thrpt   20    186148.594 ±   2943.913  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                      DIRECT   CHINESE  thrpt   20    162703.751 ±   3820.713  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP   ENGLISH  thrpt   20    399438.689 ±   3683.416  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP  JAPANESE  thrpt   20    150696.624 ±   4219.753  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP     MIXED  thrpt   20    102825.498 ±   1643.060  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP    HEBREW  thrpt   20    211749.172 ±   4625.593  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer                        HEAP   CHINESE  thrpt   20    161054.281 ±   2183.351  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT   ENGLISH  thrpt   20    318984.779 ±   8135.769  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT  JAPANESE  thrpt   20    134186.007 ±   4598.491  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT     MIXED  thrpt   20     95816.538 ±   1283.384  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT    HEBREW  thrpt   20    178311.424 ±   7073.692  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst            DIRECT   CHINESE  thrpt   20    138070.212 ±   2569.738  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP   ENGLISH  thrpt   20    375249.554 ±   7202.198  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP  JAPANESE  thrpt   20    142648.957 ±   2582.319  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP     MIXED  thrpt   20     97951.935 ±   1311.572  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP    HEBREW  thrpt   20    198253.487 ±   2770.646  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_copyFirst              HEAP   CHINESE  thrpt   20    152498.568 ±   2756.298  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT   ENGLISH  thrpt   20    367151.701 ±   6440.305  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT  JAPANESE  thrpt   20     88343.758 ±   2391.535  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT     MIXED  thrpt   20     64460.688 ±   1533.509  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT    HEBREW  thrpt   20    108406.855 ±   2833.226  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                  DIRECT   CHINESE  thrpt   20    102713.102 ±   1617.851  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP   ENGLISH  thrpt   20    440636.287 ±   2489.340  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP  JAPANESE  thrpt   20     93893.509 ±   1821.500  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP     MIXED  thrpt   20     67751.734 ±   1045.973  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP    HEBREW  thrpt   20    122597.875 ±   1997.821  ops/s
Utf8DecodeBenchmark.StringsMessageBenchmark.byteBuffer_old                    HEAP   CHINESE  thrpt   20    115595.428 ±   1846.056  ops/s
```

xfxyjwf:
@anuraaga Sorry for the late response. Yes, our internal benchmarks are proto2 heavy. The PR should be fine as long as it doesn't negatively impact proto2 performance. We will need to add more benchmarks for proto3 protos but it's not a blocking issue for this PR.

I'll wait for @nmittler @pherl in case they have some more comments and if not I'll merge it tomorrow.

anuraaga:
Thanks for the reviews!

java/core/src/main/java/com/google/protobuf/UnsafeUtil.java
```diff
@@ -259,6 +260,26 @@ static long addressOffset(ByteBuffer buffer) {
...
+      return new String(chars);
+    }
+    final String str;
+    try {
+      str = (String) UNSAFE.allocateInstance(String.class);
+    } catch (InstantiationException e) {
+      // This should never happen, but return a copy as a fallback just in case.
+      return new String(chars);
+    }
+    putObject(str, STRING_VALUE_OFFSET, chars);
```
xfxyjwf: Copy-paste the comment from Luke:

Seems like this undermines the memory semantics of the value field.

The value field in string is marked final, but this is a non-volatile write.  So if this string is unsafely published (which is possible using the APIs this is exposed through), then people could get NPEs when calling methods on this string.

There is also the other problem about the other fields in string.  In Java7/8 it doesn't have any (besides hashCode), but in the past it has had a length and offset field and in the future it might have other fields.  How do we make sure that this code works on future JDK versions, or JDKs with different field semantics.  Or things like -XX:+UseCompressedStrings in jdk6 which changed the field to be a byte[] instead of a char[]

For example in jdk10 it looks like there is an extra byte field
http://hg.openjdk.java.net/jdk10/master/file/be620a591379/src/java.base/share/classes/java/lang/String.java


To fix the memory ordering problem you can probably just use putObjectVolatile, as for handling compatibility with multiple future jdk versions, im not sure.

anuraaga: I thought the memory ordering would work because we both allocate and populate the variable in this same function - it should not be possible to dereference `value` until the object is initialized (this code should be effectively the same as the code in the standard constructor). But perhaps inlining + escape analysis breaks that assumption - are you seeing that? Doing a volatile write would indeed fix that.

As for JDK9+, they don't use a `char[]` for `value` so don't execute this code (guarded by the first if). I figured this would keep it future proof for all practical purposes, and in the fairly inconveivable chance that `char[] value` is restored in a future version a change could be made then. The speedup on JRE8 is reasonable, but not a deal breaker so it would be ok to remove this function too.

lukesandberg: The memory semantics of final fields and constructors are fairly well documented in JLS
https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.4

the key thing is that final field writes force a memory barrier at the end of the constructor.  But since you aren't using the constructor this doesn't work.  I believe volatile will fix the problem in the current code in JDK8, but it can still cause other problems in the future. 

I think the function should be removed and the string constructor should simply be invoked.  The caller of this function actually performs an Arrays.copyOf to correctly size the array.  instead we could just call the String(char[], int, int) constructor and do both things at the same time.  At least in some cases it would be just as good as the current unsafe code

anuraaga: Ah yeah, I was thinking of the function return here as a memory fence for some reason, but obviously that's not correct.

I still think the chance of a problem in future Java versions is extremely low since the method is guarded by the presence of `char[] value`, but at the same time that's enough reason to avoid this  hacky optimization if it only applies to older (Java8) runtimes. Will send a PR

PR #3827, opened 2017-11-01
Adds Dockerfile to build Protobuf

abacaphiliac:
Adds Dockerfile which creates a container to build Protobuf. Rather than worrying about Windows, OSX, or various Linux distros, developers *should* be able to quickly build Protobuf in a consistent cross-platform fashion. The build container installs dependencies required to build Protobuf and runs the configuration and make scripts unless the Docker CMD is overridden.

I need some help:
I ran through the config and make scripts successfully on my Mac, but `make` is failing on Ubuntu 16 and 14 (LTS) with the following error:

```
google/protobuf/stubs/.libs/io_win32.o: file not recognized: File format not recognized
collect2: error: ld returned 1 exit status
Makefile:2425: recipe for target 'libprotobuf.la' failed
make[2]: *** [libprotobuf.la] Error 1
make[2]: Leaving directory '/protobuf/src'
Makefile:1545: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/protobuf'
Makefile:1452: recipe for target 'all' failed
make: *** [all] Error 2
```

Any idea why this has happened? 

Thank you for the consideration and assistance.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+# Building Protobuf via Docker
+# ----------------------------
+#
+# Create the build container:
+# ```bash
+# docker build -t protobuf .
+# ```
+#
+# Run the build container (mounts repo to `/protobuf`, which is the work dir):
+# ```bash
+# docker run -it --rm --name protobuf -v `pwd`:/protobuf -w /protobuf protobuf
+# ```
+
+FROM ubuntu:xenial
+
+RUN apt-get update && \
+    apt-get install -yq autoconf automake libtool curl make g++ unzip
+
+CMD ./autogen.sh && \
+    ./configure && \
+    make && \
+    make check && \
+    make install && \
+    ldconfig


abacaphiliac:
I signed it!

abacaphiliac:
I'd be happy to post the full output, or link to a file, if it would be helpful. Just realized that there might be an upstream error leading to the snippet I pasted in the description. Let me know : )

pherl:
I guess you can skip the protobuf building step and run those inside of the docker interactively. io_win32.o should be empty in non windows platforms. I guess I would check whether the object is empty and then whether the linker allows empty file input.

Take a step back, protobuf is usually used with other projects. I'm not quite sure if it would be super useful to have a docker to build the library. Typically you would need to have the include directory available, etc.

abacaphiliac:
My use case is that I've installed a stable version of Protobuf via brew on OSX. I'm doing some development which requires the latest build from source. I do not want to compromise my stable installation so I'm exploring other options.

Per the documentation, it seems that Ubuntu is the quickest path to success. Armed with an Ubuntu container, I think we can codify the documentation. The result should be a locally available Ubuntu container with the latest Protobuf version ready to compile proto files!

Next, I'll run my protobuf container with my proto folder volume-mounted. I can use protoc on my Ubuntu container to compile my proto files into that same (or another) mount, and then stop my container. Now I don't have to worry about corrupting my stable protoc install on OSX, and I don't have to bother with Windows, etc ; P

Long-term we might be able to push those builds to Docker Hub so that everyone can benefit, at least from the dependency installation step. Maybe we can move protoc compilation into the build as well when we want to share with everyone.

xfxyjwf:
@abacaphiliac I think your use of docker file with protoc is very interesting. We have other team members brought up similar ideas the other day about providing a docker file to provide protoc compilation functionality. I don't think the docker file should be included in the top level of this repository though. We are moving protobuf to its own organization (see: https://github.com/google/protobuf/issues/4796). After that I think we can have a protobuf-docker repository for host related docker files (not only for protoc but for other components around protobuf as well).

PR #3829, opened 2017-11-01
Update Composer config in PHP module

abacaphiliac:
* Prevents installation of `phpunit/phpunit:^6` because the test cases in this project require `\PHPUnit_Framework_TestCase` which was removed in v6.
* Moves generated PHP classes for tests into autoload-dev so that they are not referenced in the dist autoload. I realize the published config is in the project root (meaning the generated classes would never be available in dist), but test dependencies are test dependencies after all.
* Adds convenience Composer scripts to run protoc with and without grpc.


abacaphiliac:
I signed it!

php/composer.json
```diff
@@ -9,13 +9,21 @@
...
+      "GPBMetadata\\Google\\Protobuf\\": "src/GPBMetadata/Google/Protobuf"
+    }
+  },
+  "autoload-dev": {
+    "psr-4": {
       "": "tests/generated"
     }
+  },
+  "scripts": {
+    "protoc": "protoc --proto_path=tests/ --php_out=tests/generated/ tests/proto/*.proto",
```
TeBoring: https://github.com/google/protobuf/blob/master/tests.sh#L348

php/composer.json
```diff
@@ -9,13 +9,21 @@
...
+    "protoc-with-grpc": "protoc --proto_path=tests/ --php_out=tests/generated/ --grpc_out=tests/generated/ --plugin=protoc-gen-grpc=`which grpc_php_plugin` tests/proto/*.proto"
```
TeBoring: Why grpc_out?

php/composer.json
abacaphiliac: should i incorporate this test function somehow?

php/composer.json
TeBoring: please remove it

abacaphiliac:
e.g. `composer protoc` and `composer protoc-with-grpc`. i can add documentation to the readme, if you like : )

abacaphiliac:
i will also add a `test` helper that can be run as `composer test`, which will wrap `vendor/bin/phpunit`. such a wrapper seems to be a convention for php libraries.

we can also include a build-and-test wrapper, e.g. `build`, that runs `composer protoc && composer test`.

TeBoring:
The ```test``` sounds good to me. I think the ```test``` should depend on building the protoc and generating code. However, we don't need an exposed target of ```protoc```

TeBoring:
any update?

abacaphiliac:
hi @TeBoring , sorry about the delay. i think `composer test` is aligned with our conversation now. what do you think?

i am concerned of the duplication in composer.json and tests.sh. do you think it will be possible to use one within the other?

TeBoring:
I think the duplication is ok. composer is used for testing php implementation while test.sh is used for testing c extension.

PR #3851, opened 2017-11-08
add optional goJsonFlags (default=",omitempty") to FieldOptions

ilius:
Our usecase right now, is that we marshal the message to json in Go (using generated go struct), and unmarshal that json in a dynamic-typed language like Lua, and unmarshal it into a generic data structure (table in Lua), but we need the zero/empty values in the json. So in short, we want the "omitempty" json flag be removed from go structs, by setting `[goJsonFlags=""]` in proto file.

But this option can be also used for completely excluding this param from json: `[goJsonFlags="-"]`
Or changing the param name in json: `[goJsonFlags="anotherName,omitempty"]`

The default value for this param (`default=",omitempty"`) makes it backward-compatible.

### DIFF
--- a/src/google/protobuf/descriptor.proto
+++ b/src/google/protobuf/descriptor.proto
@@ -576,6 +576,8 @@ message FieldOptions {
   // For Google-internal migration only. Do not use.
   optional bool weak = 10 [default=false];
 
+  // For defining json flags in generated Go structs, either removing default "omitempty" flag, or adding new flags
+  optional string goJsonFlags = 11 [default=",omitempty"];
 
   // The parser stores options it doesn't recognize here. See above.
   repeated UninterpretedOption uninterpreted_option = 999;


xfxyjwf:
We generally don't want to introduce new options in proto files. In this case it seems to me it should be a Go protobuf library feature rather than a flag in the .proto file. For example, in C++ we have this always_print_primitive_fields option to print empty fields:
https://github.com/google/protobuf/blob/master/src/google/protobuf/util/json_util.h#L63

Could you instead make a proposal to the go protobuf repo?
https://github.com/golang/protobuf

ilius:
@xfxyjwf Looks like descriptor.proto in golang/protobuf is a copy of the file in google/protobuf
(`make regenerate` overrides it)
https://github.com/golang/protobuf/blob/master/protoc-gen-go/descriptor/Makefile

ilius:
@xfxyjwf Looks like descriptor.proto in golang/protobuf is a copy of this file
https://github.com/golang/protobuf/blob/master/protoc-gen-go/descriptor/Makefile
(`make regenerate` overrides it)

664485186:
Has this problem been solved   (  omitempty  )

PR #3858, opened 2017-11-09
C#: Allow message parsing from an array slice

jtattermusch:
In gRPC C# we would like to enable more efficient buffer management and it seems as a prerequisite
to be able to parse protobuf messages from array slices (when pooling, we cannot ensure that the size of the array will be the same as the length of the payload, and for some techniques, we can't even rely on the offset being 0).
Two reasons why I'd like to add the methods to protobuf directly:
- input.CheckReadEndOfStreamTag() is internal which makes it harder to check parsing went alright in external code.
- the code to serialize/deserialize protobuf is part of the generated gRPC stubs, so keeping this code concise would help (and we can't have helpers in Grpc.Core as that one in protobuf agnostic).

Context: https://github.com/grpc/grpc/issues/13236.

### DIFF
--- a/csharp/src/Google.Protobuf/MessageExtensions.cs
+++ b/csharp/src/Google.Protobuf/MessageExtensions.cs
@@ -54,6 +54,22 @@ namespace Google.Protobuf
         }
 
         /// <summary>
+        /// Merges data from the given byte array slice into an existing message.
+        /// </summary>
+        /// <param name="message">The message to merge the data into.</param>
+        /// <param name="data">The data containing the slice to merge, which must be protobuf-encoded binary data.</param>
+        /// <param name="offset">The offset of the slice to merge.</param>
+        /// <param name="length">The length of the slice to merge.</param>
+        public static void MergeFrom(this IMessage message, byte[] data, int offset, int length)
+        {
+            ProtoPreconditions.CheckNotNull(message, "message");
+            ProtoPreconditions.CheckNotNull(data, "data");
+            CodedInputStream input = new CodedInputStream(data, offset, length);
+            message.MergeFrom(input);
+            input.CheckReadEndOfStreamTag();
+        }
+
+        /// <summary>
         /// Merges data from the given byte string into an existing message.
         /// </summary>
         /// <param name="message">The message to merge the data into.</param>
--- a/csharp/src/Google.Protobuf/MessageParser.cs
+++ b/csharp/src/Google.Protobuf/MessageParser.cs
@@ -71,6 +71,21 @@ namespace Google.Protobuf
         }
 
         /// <summary>
+        /// Parses a message from a byte array slice.
+        /// </summary>
+        /// <param name="data">The byte array containing the message. Must not be null.</param>
+        /// <param name="offset">The offset of the slice to parse.</param>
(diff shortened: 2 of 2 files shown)


csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -54,6 +54,22 @@ public static void MergeFrom(this IMessage message, byte[] data)
...
         /// <summary>
+        /// Merges data from the given byte array slice into an existing message.
+        /// </summary>
+        /// <param name="message">The message to merge the data into.</param>
+        /// <param name="data">The data containing the slice to merge, which must be protobuf-encoded binary data.</param>
+        /// <param name="offset">The offset of the slice to merge.</param>
+        /// <param name="length">The length of the slice to merge.</param>
+        public static void MergeFrom(this IMessage message, byte[] data, int offset, int length)
+        {
+            ProtoPreconditions.CheckNotNull(message, "message");
```
jskeet: At some point I should do a nameof tidy-up... but I understand the point of sticking to the convention of the rest of the file here.

csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -71,6 +71,21 @@ public IMessage ParseFrom(byte[] data)
...
         /// <summary>
+        /// Parses a message from a byte array slice.
+        /// </summary>
+        /// <param name="data">The byte array containing the message. Must not be null.</param>
+        /// <param name="offset">The offset of the slice to parse.</param>
+        /// <param name="length">The length of the slice to parse.</param>
+        /// <returns>The newly parsed message.</returns>
+        public IMessage ParseFrom(byte[] data, int offset, int length)
+        {
+            ProtoPreconditions.CheckNotNull(data, "data");
```
jskeet: I'd ditch this given that it'll be validated in the call to message.MergeFrom.

(If you change the parameter names to buffer in both cases, you could rely on the CodedInputStream constructor validation, but I prefer data instead of buffer here...)

jtattermusch:
A few more corrections:
- remove redundant CheckNotNull check where already checked by MergeFrom
- add a generic version of ParseFrom as well.

PTAL.

jtattermusch:
@anandolee , what's the timeline for the next protobut 3.5 release? I'd like to make sure this change will be included.

jskeet:
I won't be able to look again until tomorrow - about to board a flight.

anandolee:
Jisi has already did a cut off and the change logs are ready. I assume the release will be in next week or two.
@pherl  are we able to include this change in 3.5?

csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -64,7 +64,6 @@ internal IMessage CreateTemplate()
         /// <returns>The newly parsed message.</returns>
         public IMessage ParseFrom(byte[] data)
         {
-            ProtoPreconditions.CheckNotNull(data, "data");
```
jskeet: The downside of removing this is that the parameter name will be wrong in the exception (buffer rather than data). But I'm not massively worried.

PR #3862, opened 2017-11-10
Fix out-of-tree build for cmake

lofcek:


lofcek:
Fix out-of-tre build for cmake

File well_known_types_embed.cc was generatoed to the same directory, where are sources. Therefore it was not possible to mount sources as read-only.

PR #3863, opened 2017-11-10
Fix out-of-tree build for cmake

lofcek:
File well_known_types_embed.cc is generated in the same place where are source code and this does not allow us make our sources read-only

### DIFF
--- a/cmake/libprotoc.cmake
+++ b/cmake/libprotoc.cmake
@@ -72,7 +72,6 @@ set(libprotoc_files
   ${protobuf_source_dir}/src/google/protobuf/compiler/javanano/javanano_message_field.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/javanano/javanano_primitive_field.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/js/js_generator.cc
-  ${protobuf_source_dir}/src/google/protobuf/compiler/js/well_known_types_embed.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_enum.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_enum_field.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_extension.cc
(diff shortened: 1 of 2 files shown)


cmake/libprotoc.cmake
```diff
@@ -92,6 +91,7 @@ set(libprotoc_files
   ${protobuf_source_dir}/src/google/protobuf/compiler/ruby/ruby_generator.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/subprocess.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/zip_writer.cc
+  ${protobuf_BINARY_DIR}/well_known_types_embed.cc		# This file is generated, therefore is in different directory
```
xfxyjwf: This file list here is generated by the update_file_lists.sh script:
https://github.com/google/protobuf/blob/master/update_file_lists.sh

So your change here won't stand unless you make a corresponding change to src/Makefile.am. Can you make a change to src/Makefile.am accordingly?

lofcek: I will try, although I have no experience with automake .. but for now I have

lofcek: something that could fix it for a few days

update_file_lists.sh
```diff
@@ -96,7 +96,12 @@ set_cmake_value() {
       print \$0;
       len = split(values, vlist, \" \");
       for (i = 1; i <= len; ++i) {
-        printf(\"  %s%s\\n\", prefix, vlist[i]);
+	    if (vlist[i] != \"google/protobuf/compiler/js/well_known_types_embed.cc\") {
```
xfxyjwf: Format these using two-spaces indentation.

update_file_lists.sh
```diff
@@ -96,7 +96,12 @@ set_cmake_value() {
...
+           printf(\"  %s%s\\n\", prefix, vlist[i]);
+		} else {
+           printf(\"  \${protobuf_BINARY_DIR}/well_known_types_embed.cc\t\t# This file is generated, therefore is in different directory\\n\");
```
xfxyjwf: Put the comment before the print line.

BSBandme:
Please fix the conflicts first.

xfxyjwf:
This should no longer be needed because I have deleted the problematic js_embed binary in our build.

PR #3876, opened 2017-11-13
3.4.x

lofcek:


pietern:
Could you consider #2658 for inclusion as well? I'm trying to bump protobuf to 3.4.1 for Caffe2 but see the exact same failure when there is no system protobuf installed. As is, the CMake build will use "js_embed" from $PATH instead of the one that is built just in time. If you have protobuf installed and js_embed in your $PATH, this error won't show up.

PR #3878, opened 2017-11-13
Allow one to omit building libprotoc and protoc binaries

Yangqing:
Basically, in Android/iOS, if one uses the cmake script to do cross compiling, the js_embed target fails making the whole build fail. What one usually need in Android cross compilation probably do not include the protoc compiler, so this PR provides an option to turn it off.

Locally tested with cmake -Dprotobuf_BUILD_PROTOC_BINARIES=off && make && make install.


Yangqing:
Appveyor is passing and I don't think the travis failures are due to this PR. Would someone like to take a look and kick off jenkins?

xfxyjwf:
I think we should fix the build issue of js_embed rather than disabling protoc. More configuration options mean more maintenance overhead.  Maybe we just need to tell cmake the js_embed binary is built for host someway?

Yangqing:
@xfxyjwf I also agree that reducing the number of options is in general good.

In this case, I am a bit torn mainly because in cross-compilation stages, it would be tricky to require a protoc build because the toolchain would also need a host compiler. For example, when one uses gradle to build Android apps which further uses cmake, it makes life harder if one has to build js_embed for host (for example, it requires one to install the whole XCode).

Mixing cross-compilation and host compilation is also not very well supported by CMake IIRC.

Another option is to pass a protobuf_JS_EMBED_BINARY_PATH pointing to a prebuilt one, but that involves adding an option anyway, so this one is actually more preferred - build what you use and not any more.

Yangqing:
For possible solutions, see this thread:

https://cmake.org/pipermail/cmake/2013-January/053253.html

xfxyjwf:
I'm surprised cmake is lacking such basic support for cross-compiling. Without proper support from cmake this PR seems a reasonable workaround.

PR #3880, opened 2017-11-14
Add Setter/Getter type verification.

dmaclach:
Add runtime asserts (that can be disabled in release) that verify
that the types being get/set for messages using the C Api match
the type in the descriptor for the field being get/set.

### DIFF
--- a/objectivec/GPBUtilities.m
+++ b/objectivec/GPBUtilities.m
@@ -355,12 +355,14 @@ void GPBMaybeClearOneof(GPBMessage *self, GPBOneofDescriptor *oneof,
 //%PDDM-DEFINE IVAR_POD_ACCESSORS_DEFN(NAME, TYPE)
 //%TYPE GPBGetMessage##NAME##Field(GPBMessage *self,
 //% TYPE$S            NAME$S       GPBFieldDescriptor *field) {
+//%#if defined(DEBUG) && DEBUG
 //%  NSCAssert(DataTypesEquivalent(GPBGetFieldDataType(field),
 //%                                GPBDataType##NAME),
 //%            @"Attempting to get value of TYPE from field %@ "
 //%            @"of %@ which is of type %@.",
 //%            [self class], field.name,
 //%            TypeToString(GPBGetFieldDataType(field)));
+//%#endif
 //%  if (GPBGetHasIvarField(self, field)) {
 //%    uint8_t *storage = (uint8_t *)self->messageStorage_;
 //%    TYPE *typePtr = (TYPE *)&storage[field->description_->offset];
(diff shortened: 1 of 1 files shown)


objectivec/GPBUtilities.m
```diff
@@ -1185,11 +1395,8 @@ void GPBSetMessageMapField(GPBMessage *self, GPBFieldDescriptor *field,
   if (dictionary) {
     GPBDataType keyDataType = field.mapKeyDataType;
     GPBDataType valueDataType = GPBGetFieldDataType(field);
-    NSString *keyStr = TypeToStr(keyDataType);
-    NSString *valueStr = TypeToStr(valueDataType);
-    if (keyDataType == GPBDataTypeString) {
```
thomasvl: Isn't this still needed since the type is GPBString*Dictionary?

objectivec/GPBUtilities.m
```diff
@@ -49,6 +49,19 @@ static void AppendTextFormatForMessage(GPBMessage *message,
                                        NSMutableString *toStr,
                                        NSString *lineIndent);
 
+// Are two datatypes the same basic type representation (ex Int32 and SInt32).
+// Marked unused because currently only called from asserts/debug.
+static BOOL GPBDataTypesEquivalent(GPBDataType type1,
```
thomasvl: Don't really need GPB on these since they are file private and not meant to ever be exposed.

objectivec/GPBUtilities.m
```diff
@@ -1137,8 +1316,40 @@ void GPBSetMessageRepeatedField(GPBMessage *self, GPBFieldDescriptor *field, id
   GPBSetObjectIvarWithField(self, field, array);
 }
 
-#if defined(DEBUG) && DEBUG
-static NSString *TypeToStr(GPBDataType dataType) {
+BOOL GPBDataTypesEquivalent(GPBDataType type1, GPBDataType type2) {
```
thomasvl: Repeat the static on these?

objectivec/GPBUtilities.m
```diff
@@ -1137,8 +1316,40 @@ void GPBSetMessageRepeatedField(GPBMessage *self, GPBFieldDescriptor *field, id
...
+  return GPBBaseDataType(type1) == GPBBaseDataType(type2);
+}
+
+GPBDataType GPBBaseDataType(GPBDataType type) {
+  switch(type) {
```
thomasvl: space after `switch`

objectivec/GPBUtilities.m
```diff
@@ -1137,8 +1316,40 @@ void GPBSetMessageRepeatedField(GPBMessage *self, GPBFieldDescriptor *field, id
...
+    case GPBDataTypeSFixed32:
+    case GPBDataTypeInt32:
+    case GPBDataTypeSInt32:
+    case GPBDataTypeEnum:
```
thomasvl: I guess mapping enum here is ok with the current uses, but I wonder if it should be ensure it is enum?

objectivec/GPBUtilities.m
dmaclach: Fixed

objectivec/GPBUtilities.m
dmaclach: Fixed

objectivec/GPBUtilities.m
dmaclach: I wasn't that concerned about it. Want me to fix it up?

objectivec/GPBUtilities.m
dmaclach: Oops.. thank you. not sure how it got removed.

objectivec/GPBUtilities.m
thomasvl: I guess we should be safe with current uses, with the naming helping indicate it is internal, it likely isn't as much of an issue.

dmaclach:
Added #if DEBUG checks

PR #3884, opened 2017-11-15
Simplify getter/setter method implementations

dmaclach:
Remove unnecessary retain/releases for GPBFieldDescriptors when
creating implementations with blocks. The descriptors do not need
to be memory managed because they are essentially global constructs
that are guaranteed to be around the lifetime of the app.

### DIFF
--- a/objectivec/GPBMessage.m
+++ b/objectivec/GPBMessage.m
@@ -2998,7 +2998,7 @@ typedef struct ResolveIvarAccessorMethodResult {
   SEL encodingSelector;
 } ResolveIvarAccessorMethodResult;
 
-static void ResolveIvarGet(GPBFieldDescriptor *field,
+static void ResolveIvarGet(__unsafe_unretained GPBFieldDescriptor *field,
                            ResolveIvarAccessorMethodResult *result) {
   GPBDataType fieldDataType = GPBGetFieldDataType(field);
   switch (fieldDataType) {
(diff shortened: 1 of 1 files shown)


objectivec/GPBMessage.m
```diff
@@ -3086,7 +3086,12 @@ + (BOOL)resolveInstanceMethod:(SEL)sel {
   // no need for checks here to see if has*/setHas* are allowed.
 
   ResolveIvarAccessorMethodResult result = {NULL, NULL};
-  for (GPBFieldDescriptor *field in descriptor->fields_) {
+
```
thomasvl: Extra blank line here can go away.

objectivec/GPBMessage.m
```diff
@@ -3086,7 +3086,12 @@ + (BOOL)resolveInstanceMethod:(SEL)sel {
...
+  
+  // |field| can be __unsafe_unretained because they are created at startup
+  // and are essentially global. No need to pay for retain/release when
+  // they are captured in blocks.
```
thomasvl: Can you dup this comment on the two other places that added `__unsafe_unretained`.  Or maybe a block comment before the two statics explain the use generally in this part of the code?

objectivec/GPBMessage.m
dmaclach: Done

objectivec/GPBMessage.m
dmaclach: Done

PR #3887, opened 2017-11-15
Fix up warnings from Xcode 9.1

dmaclach:
Removes a framework path that no longer exists in SDK
Cleans up icons and launchimages to match what Xcode expects (also updated icons to more hires versions 😄


objectivec/ProtocolBuffers_iOS.xcodeproj/project.pbxproj
```diff
@@ -880,7 +880,6 @@
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
 				FRAMEWORK_SEARCH_PATHS = (
-					"\"$(SDKROOT)/Developer/Library/Frameworks\"",
```
thomasvl: Will thinks still build with older Xcodes?

thomasvl:
The travis setup is current still on 8.x, I'm in the process of updating that also.  I haven't looked at what the kokoro setup for the mac builds is using.

thomasvl:
Xcode 8.3 is happy with the Xcode change, but you need to fix up the `Makefile.am`, see the `cpp_distcheck` failure.  That makes a tar/zip out of the sources for each language support, so the [files all have to be listed](https://github.com/google/protobuf/blob/master/Makefile.am#L561-L570).

dmaclach:
Makefile updated appropriately

PR #3900, opened 2017-11-16
build: isolate symbols of Python extensions on Linux

### LINKED ISSUE #3899: build[python]: avoid symbols conflicts with other loaded libprotobuf.so (Linux)
There are several libraries/tools that uses protobuf via linking with system-wide libprotobuf.so.
Protobuf version mismatch between system protobuf binary and protobuf Python's extension (from [PyPi](https://pypi.python.org/pypi/protobuf/3.4.0)) usually cause a fatal crashes.

Symbols conflict detection:
- Fedora 26 with protobuf-devel package installed (protobuf version is 3.2.0).
- Python virtualenv with installed protobuf package (`pip install protobuf`, protobuf version is 3.4.0)
- try to simulate some library/tool which loads system protobuf binary:
> $ python -c 'from ctypes import *; CDLL("/lib64/libprotobuf.so", mode=RTLD_GLOBAL); from google.protobuf import api_pb2'
> Segmentation fault (core dumped)
- `LD_PRELOAD=/lib64/libprotobuf.so` instead of `CDLL` works too

Crash stack trace shows protobuf symbols:
```
(gdb) bt
#0  std::_Hashtable<std::string, std::string, std::allocator<std::string>, std::__detail::_Identity, std::equal_to<std::string>, google::protobuf::hash<std::string>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, true, true> >::_M_deallocate_nodes (this=0x5555557b4178, __n=0x1)
    at /opt/rh/devtoolset-2/root/usr/include/c++/4.8.2/bits/hashtable.h:763
#1  std::_Hashtable<std::string, std::string, std::allocator<std::string>, std::__detail::_Identity, std::equal_to<std::string>, google::protobuf::hash<std::string>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, true, true> >::clear (this=0x5555557b4178)
    at /opt/rh/devtoolset-2/root/usr/include/c++/4.8.2/bits/hashtable.h:1641
#2  0x00007fffeb038116 in std::unordered_set<std::string, google::protobuf::hash<std::string>, std::equal_to<std::string>, std::allocator<std::string> >::clear (this=<optimized out>)
    at /opt/rh/devtoolset-2/root/usr/include/c++/4.8.2/bits/unordered_set.h:472
#3  google::protobuf::DescriptorPool::FindFileByName (this=0x55555578d3b0, name="")
    at google/protobuf/descriptor.cc:1327
#4  0x00007fffeaff1f98 in google::protobuf::python::cdescriptor_pool::AddSerializedFile (self=0x7fffeb9cc1b8, serialized_pb=0x555555815e30)
    at google/protobuf/pyext/descriptor_pool.cc:510
#5  0x00007ffff7b12342 in PyEval_EvalFrameEx () from /lib64/libpython2.7.so.1.0
```

But frames 0,1,2 are related to system libprotobuf.so (3.2.0, **/opt/rh/** marker or "info source" gdb command) and frames 3,4 are from protobuf Python extension (3.4.0).
Sure, these functions are not binary compatible and they can't call each other.

<details>
<summary>Output of LD_DEBUG=files,symbols</summary>

```
      1379:	file=/home/alalek/penv/lib/python2.7/site-packages/google/protobuf/pyext/_message.so [0];  dynamically loaded by /lib64/libpython2.7.so.1.0 [0]
      1379:	file=/home/alalek/penv/lib/python2.7/site-packages/google/protobuf/pyext/_message.so [0];  generating link map
      1379:	  dynamic: 0x00007f796de7bba8  base: 0x00007f796da87000   size: 0x00000000003ff648
      1379:	    entry: 0x00007f796db1d190  phdr: 0x00007f796da87040  phnum:                  7
      1379:	
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=python [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libpython2.7.so.1.0 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libpthread.so.0 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libdl.so.2 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libutil.so.1 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libm.so.6 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libc.so.6 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/ld-linux-x86-64.so.2 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libprotobuf.so [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libz.so.1 [0]
      1379:	symbol=_ZTVN10__cxxabiv120__si_class_type_infoE;  lookup in file=/lib64/libstdc++.so.6 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=python [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libpython2.7.so.1.0 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libpthread.so.0 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libdl.so.2 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libutil.so.1 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libm.so.6 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libc.so.6 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/ld-linux-x86-64.so.2 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libprotobuf.so [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libz.so.1 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libstdc++.so.6 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libgcc_s.so.1 [0]
      1379:	symbol=_ZTSN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/home/alalek/penv/lib/python2.7/site-packages/google/protobuf/pyext/_message.so [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=python [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libpython2.7.so.1.0 [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libpthread.so.0 [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libdl.so.2 [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libutil.so.1 [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libm.so.6 [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libc.so.6 [0]
      1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/ld-linux-x86-64.so.2 [0]
<OOPS> 1379:	symbol=_ZTIN6google8protobuf18DescriptorDatabaseE;  lookup in file=/lib64/libprotobuf.so [0]
      1379:	symbol=_ZTIN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=python [0]
      1379:	symbol=_ZTIN6google8protobuf6python20PyDescriptorDatabaseE;  lookup in file=/lib64/libpython2.7.so.1.0 [0]
... there are many OOPS cases ...
```
</details>

---

Related SO question: https://stackoverflow.com/questions/7201667/ld-magically-overrides-statically-linked-symbols

Possible solution is to isolate protobuf symbols in Python extension (assume gcc+ld on Linux):
1. link Python extensions with `-Bsymbolic` linker flag
2. don't export protobuf symbols by using `-fvisibility=hidden` compiler flag. But this breaks .so build.
3. don't export symbols from .a files: `-exclude-libs=ALL` linker option

There are many "magic" crashes related to conflicts of different protobuf versions, so lets try to resolve this issue.

Additional information:
- similar problem can be observed on Ubuntu
- Installing the same version of Python extension (`pip install protobuf==3.2.0`) doesn't help (still SIGSEGV)

alalek:
to avoid conflicts with other loaded libprotobuf.so libraries.

resolves #3899

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -200,12 +200,15 @@ if __name__ == '__main__':
         extra_compile_args.append('-std=c++11')
 
     if platform.system() == 'Linux':
+      linker_skip_symbols_isolation = get_option_from_sys_argv('--linker_skip_symbols_isolation')
       # isolate "local" symbols
-      extra_link_args.append('-Wl,-Bsymbolic')
+      if not linker_skip_symbols_isolation:
+        extra_link_args.append('-Wl,-Bsymbolic')
       if compile_static_ext:
         # don't export symbols from .a libraries
         # (there is no -fvisibility=hidden option during compilation of .a files)
-        extra_link_args.append('-Wl,--exclude-libs=ALL')
+        if not linker_skip_symbols_isolation:
+          extra_link_args.append('-Wl,--exclude-libs=ALL')
 
     if warnings_as_errors in sys.argv:
       extra_compile_args.append('-Werror')
--- a/python/setup.py
+++ b/python/setup.py
@@ -181,6 +181,7 @@ if __name__ == '__main__':
     extra_compile_args = ['-Wno-write-strings',
                           '-Wno-invalid-offsetof',
                           '-Wno-sign-compare']
+    extra_link_args = []
     libraries = ['protobuf']
     extra_objects = None
     if compile_static_ext:
@@ -198,6 +199,14 @@ if __name__ == '__main__':
       if v >= 10.12:
         extra_compile_args.append('-std=c++11')
 
+    if platform.system() == 'Linux':
+      # isolate "local" symbols
+      extra_link_args.append('-Wl,-Bsymbolic')
+      if compile_static_ext:
+        # don't export symbols from .a libraries
+        # (there is no -fvisibility=hidden option during compilation of .a files)
+        extra_link_args.append('-Wl,--exclude-libs=ALL')
+
     if warnings_as_errors in sys.argv:
       extra_compile_args.append('-Werror')
       sys.argv.remove(warnings_as_errors)
@@ -212,11 +221,13 @@ if __name__ == '__main__':
             extra_objects=extra_objects,
             library_dirs=['../src/.libs'],
             extra_compile_args=extra_compile_args,
+            extra_link_args=extra_link_args,
         ),
         Extension(
             "google.protobuf.internal._api_implementation",
             glob.glob('google/protobuf/internal/api_implementation.cc'),
             extra_compile_args=['-DPYTHON_PROTO2_CPP_IMPL_V2'],
+            extra_link_args=extra_link_args,
         ),
     ])
     os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'


haberman:
The risk of this approach is: if someone gets a `google::protobuf::Message*` from Python somehow, then tries to pass it to another protobuf API, we could get a crash then because the two protobuf library versions could have different internal layouts for objects.

Mixing libprotobuf.so versions in the same process seems inherently dangerous. If we go down this path, we should ensure as strictly as possible that you can't get a `google::protobuf::Message*`, `proto2::protobuf::Descriptor*`, etc. from the Python library.

alalek:
Thank you for taking a look on this!

Protobuf is very popular library (even more than we want really), so avoiding mixed usage of different libprotobuf versions is very hard sometimes:
- for example, using proto3 formats with system libprotobuf.so 2.x
- or Python extension case: it is build on another platform (manylinux) without any version of libprotobuf.so and after that this extension is reused on other multiple Linux systems (with or without own libprotobuf.so).

Main assumption here is that protobuf implementations are isolated via API interfaces (according this: https://github.com/google/protobuf/issues/435#issuecomment-122145694) and don't share libprotobuf entries at all.
Second assumption is that non-system protobuf library is linked statically (case of Python extension from Pypi).

I agree, that some detection of this on compile / link time should be better, but this is out of scope of this patch.

This patch is attempt to isolate/hide symbols of statically linked protobuf from Linux loader (to avoid linking them with external .so with similar symbols of libprotobuf including different version).

xfxyjwf:
Closing old issues.

PR #3901, opened 2017-11-16
Fix unused parameter warnings in generated code with arenas enabled

wijagels:
This only seems to affect code generated with `option cc_enable_arenas = true`.  All I did was update `cpp_message.cc` to exclude a parameter for `RegisterArenaDtor` if registration is a no-op.


pherl:
ok to test

PR #3907, opened 2017-11-16
Travis config cleanups and move ObjC to Xcode 9.1.

thomasvl:
Stop having travis build the test matrix and make it explicit so extra
settings are only on the test that need the, this helps make the travis
UI a little cleaner (linux hosted stuff doesn't list an Xcode version),
as well as avoid unneeded requirements limiting the travis pool something
can run on.

Small improvements to objectivec/DevTools/full_mac_build.sh to avoid some
repeated scripting and hopefully make things more reliable work with Xcode
9.x.

Don't mark the iOS tests as able to fail, hopefully with the new settings
they are stable and won't give false signals.

Simplify the case for xcode versions and fixup xcode 9.[0-1].


thomasvl:
Got one failure on the travis side, but it looks like a networking timeout in one of the tests (php7.0_mac), retrying that one config to confirm it.

thomasvl:
Retry of that failure worked fine, so it was a networking/timeout issue.  Landing.

PR #3918, opened 2017-11-18
Include .cc and .h files in Python source distribution

OEP:
If you try to build from the source distribution on PyPI and use the cpp implementation it currently fails with this error when you try to use it:

```
exceptions.ImportError, dynamic module does not define init function (init_message)
```

Part of the problem is the source distribution is missing the .cc and .h files for the Python extension. I'm able to build the extension from the source distribution after making this change.


OEP:
I signed it.

PR #3926, opened 2017-11-20
Sync benchmark changes from internal

BSBandme:


xfxyjwf:
Yilun, why does this PR contain several commits? Isn't it just a clean copy from our internal code base?

BSBandme:
Seems that MacOS c++ can't be built successfully because of dependency things, but locally it can built.
Still investigating.

xfxyjwf:
Can you split this pull request into 2 commits?
1. A clean copy from internal code base.
2. The change you need to make tests pass.

BSBandme:
Split done

benchmarks/Makefile.am
```diff
@@ -19,16 +19,17 @@ benchmarks_protoc_inputs_proto2 =                                          \
 	datasets/google_message4/benchmark_message4_2.proto                      \
 	datasets/google_message4/benchmark_message4_3.proto
 
-MAINTAINERCLEANFILES =                                                     \
-	Makefile.in
+make_tmp_dir:
+	if [ ! -d 'tmp' ]; then mkdir -p 'tmp'; fi
```
xfxyjwf: You can just do "mkdir -p tmp". No need to check whether it exists.

PR #3936, opened 2017-11-21
 Add unknown field support for csharp

anandolee:


csharp/src/Google.Protobuf.Test/UnknownFieldSetTest.cs
```diff
@@ -0,0 +1,127 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+using System;
+using Google.Protobuf.TestProtos;
+using NUnit.Framework;
+
+namespace Google.Protobuf
+{
```
jskeet: Just starting to review this - I'll be sending a full review later. But before I start, a general point: please reformat all the C# code to use spaces rather than tabs for consistency. (If you're using Visual Studio, it's easy to set this as the default, then use Ctrl-E, D to reformat a whole document.)

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,340 @@
...
+				}
+			}
+			return this;
+		}
+
+		/// <summary>
+		/// Adds a field to the unknown field set. If a field with the same
+		/// number already exists, the two are merged.
+		/// </summary>
+		public UnknownFieldSet MergeField(int number, UnknownField field)
```
jskeet: This doesn't looks like it needs to be public.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,340 @@
...
+		/// <summary>
+		/// Creates a new UnknownFieldSet.
+		/// </summary>
+		public UnknownFieldSet()
+		{
+			this.fields = new Dictionary<int, UnknownField>();
+		}
+
+		/// <summary>
+		/// Gets the default UnknownFieldSet instance.
```
jskeet: We can't sensibly do this while the type is mutable. Rather than talk about this in individual comments, I'll write an overall comment later.

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -1271,6 +1271,30 @@ private void SkipImpl(int amountToSkip)
             }
         }
 
+        private static volatile bool preserveUnknownFieldsDefault = false;
+        /// <summary>
+        /// Sets the preserveUnknownFieldsDefault true. It is added for
+        /// migration. Will be removed once the migration is done.
+        /// </summary>
+        public static void SetPreserveUnknownsDefaultTrue() {
```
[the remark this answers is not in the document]
anandolee: The C++ for example, introduced preserve unknown fields in 3.4.0 and the default is drop. 
3.5.0 changed the default to preserve. Release log ask users to use DiscardUnknownFields() if they want to drop it. I will add DiscardUnknownFields() API in C# as well

csharp/src/Google.Protobuf.Test/UnknownFieldSetTest.cs
anandolee: Done

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -1271,6 +1271,30 @@ private void SkipImpl(int amountToSkip)
...
+        /// migration. Will be removed once the migration is done.
+        /// </summary>
+        public static void SetPreserveUnknownsDefaultFalse() {
+            preserveUnknownFieldsDefault  = false;
+        }
+
+        /// <summary>
+        /// Gets the preserveUnknownFieldsDefault;
+        /// </summary>
+        public static bool GetPreserveUnknownsDefault() {
```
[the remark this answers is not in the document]
anandolee: Done

csharp/src/Google.Protobuf/Collections/Lists.cs
```diff
@@ -0,0 +1,84 @@
...
+			}
+			if (left == null || right == null)
+			{
+				return false;
+			}
+			if (left.Count != right.Count)
+			{
+				return false;
+			}
+			IEqualityComparer<T> comparer = EqualityComparer<T>.Default;
```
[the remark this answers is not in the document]
anandolee: ack

csharp/src/Google.Protobuf/Collections/Lists.cs
```diff
@@ -0,0 +1,84 @@
...
+
+		/// <summary>
+		/// Gets the list's hash code.
+		/// </summary>
+		public static int GetHashCode<T>(List<T> list)
+		{
+			int hash = 31;
+			foreach (T element in list)
+			{
+				hash = hash*29 + element.GetHashCode();
```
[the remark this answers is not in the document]
anandolee: Done

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -0,0 +1,269 @@
...
+		private List<uint> fixed32List;
+		private List<ulong> fixed64List;
+		private List<ByteString> lengthDelimitedList;
+
+		/// <summary>
+		/// Creates a new UnknownField.
+		/// </summary>
+		public UnknownField()
+		{
+			varintList = new List<ulong>();
```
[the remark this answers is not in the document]
anandolee: done

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -0,0 +1,269 @@
...
+				   && Lists.Equals(lengthDelimitedList, otherField.lengthDelimitedList);
+		}
+
+		/// <summary>
+		/// Get the hash code of the unknown field.
+		/// </summary>
+		public override int GetHashCode()
+		{
+			int hash = 43;
+			hash = hash*47 + Lists.GetHashCode(varintList);
```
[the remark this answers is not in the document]
anandolee: done

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -0,0 +1,269 @@
...
+			}
+			return current;
+		}
+
+		/// <summary>
+		/// Clears the contents of this builder.
+		/// </summary>
+		public UnknownField Clear()
+		{
+			varintList.Clear();
```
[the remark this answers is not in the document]
anandolee: done

csharp/src/Google.Protobuf/UnknownFieldSet.cs
anandolee: Have remove it and use null instead

csharp/src/Google.Protobuf/UnknownFieldSet.cs
anandolee: done

anandolee:
Yes, preserving unknown fields are happening in all other languages from 3.4.0.  C++ have already changed the default behavior to preserve in 3.5.0  You can see the release log for more info.

For the DefaultInstance issue, I've remove the one in UnknownFieldSet and use null instead. The UnknownField may still need a default one. I don't quit understand your option 1. Do you mean that add a check in each mutate method, just return if it is the DefaultInstance ?

jskeet:
I'd suggest leaving the field null by default even if unknown fields are enabled - I'd expect *most* parse operations not to end up finding any unknown fields, so let's avoid the extra allocation.

Instead, if you create a static method in UnknownFieldSet that accepts an existing field set, and creates a new one if necessary, then you can have:

```csharp
unknownFields = UnknownFields.MergeFieldFrom(unknownFields, input)
````

In equality, you can just use

```csharp
Equals(unknownFields, other.unknownFields)
```

... which will call Equals if and only if both values are non-null. Note that this relies on a field set never being empty - null is effectively *the* representation of an empty unknown field set.

I'd also suggest renaming the field to _unknownFields so that it won't conflict with a message field called "unknown_fields".

anandolee:
Have removed the DefaultInstance.
Changed unknownFields to _unknownFields in code gen. 
Also changed to use Equals(_unknownFields, other._unknownFields) in code gen.

For unknownFields = UnknownFields.MergeFieldFrom(unknownFields, input) suggestion, I think it is not needed. Unknown fields are most used in code gen which does not need this support

jskeet:
> For unknownFields = UnknownFields.MergeFieldFrom(unknownFields, input) suggestion, I think it is not needed. Unknown fields are most used in code gen which does not need this support

I'm not sure what you mean. The point is that that would allow you to keep `_unknownFields` null until it needs to be populated. Anyway, I'll try to review the new code very early next week.

anandolee:
Oh, that's what you mean. I've added those methods.
Thanks for all your feedback

jskeet:
I think you need to regenerate all the C# files - it won't build at the moment.
If we're using null for a missing field set, do we still need `Clear()` on either `UnknownFieldSet` or `UnknownField`?

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,371 @@
...
+
+        // Optimization:  We keep around the last field that was
+        // modified so that we can efficiently add to it multiple times in a
+        // row (important when parsing an unknown repeated field).
+        private int lastFieldNumber;
+        private UnknownField lastField;
+
+        private UnknownField GetField(int number)
+        {
+            if (lastField != null)
```
jskeet: Use && here?

```csharp
if (lastField != null && number == lastFieldNumber)
```

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,371 @@
...
+            if (number == 0)
+            {
+                return null;
+            }
+
+            lastField = new UnknownField();
+            UnknownField existing;
+            if (fields.TryGetValue(number, out existing))
+            {
+                lastField.MergeFrom(existing);
```
jskeet: Why are we merging an old one with a new one here? Again, this feels like it's left over from a builder mentality. I'd expect to return `existing` here, and only create a new instance if it's new.

I'd also rename the method to GetOrAdd, as otherwise it's odd that it mutates the set in some cases.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
anandolee: done

csharp/src/Google.Protobuf/UnknownFieldSet.cs
anandolee: done

jskeet:
Can I check, is this now ready for complete review as far as you're aware?

anandolee:
Yes, I think it is ready fro complete review

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+        public bool HasField(int field)
+        {
+            return fields.ContainsKey(field);
+        }
+
+        /// <summary>
+        /// Fetches a field by number, returning an empty field if not present.
+        /// Never returns null.
+        /// </summary>
+        public UnknownField this[int number]
```
jskeet: Creating a new field but not associating it with anything feels misleading to me:

`fieldSet[10].AddVarint(...)` looks like it's always going to modify the set, but it might not.

I'd suggest either removing the indexer entirely, or letting it just delegate to `fields[number]` so it'll throw a KeyNotFoundException if the field isn't present.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+    /// <summary>
+    /// Used to keep track of fields which were seen when parsing a protocol message
+    /// but whose field numbers or types are unrecognized. This most frequently
+    /// occurs when new fields are added to a message type and then messages containing
+    /// those fields are read by old software that was built before the new types were
+    /// added.
+    ///
+    /// Most users will never need to use this class directly.
+    /// </summary>
+    public sealed partial class UnknownFieldSet : IMessage
```
jskeet: Does this need to implement IMessage? Is there something that makes that particularly useful, or is it just that it happens to look similar?

Should we have some way of iterating over the fields? I can see that being useful.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// Gets the unknown field set's hash code.
+        /// </summary>
+        public override int GetHashCode()
+        {
+            int ret = 1;
+            foreach (KeyValuePair<int, UnknownField> field in fields)
```
jskeet: Presumably you're using ^ to make the order irrelevant? It might be worth explicitly stating that.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+        public MessageDescriptor Descriptor
+        {
+            get { return null; }
+        }
+
+        /// <summary>
+        /// Checks if the unknown field set is initialized. It is never used.
+        /// Returns true to implement the interface.
+        /// </summary>
+        public bool IsInitialized
```
jskeet: This isn't part of the C# API - let's remove it.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+
+        // Optimization:  We keep around the last field that was
+        // modified so that we can efficiently add to it multiple times in a
+        // row (important when parsing an unknown repeated field).
+        private int lastFieldNumber;
+        private UnknownField lastField;
+
+        private UnknownField GetOrAddField(int number)
+        {
+            if (lastField != null && number == lastFieldNumber)
```
jskeet: When would lastField be null, but lastFieldNumber be equal to number in an unexpected way?

(In other words, can we just remove the first condition?)

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+            AddField(number, lastField);
+            lastFieldNumber = number;
+            return lastField;
+        }
+
+        /// <summary>
+        /// Adds a field to the set. If a field with the same number already exists, it
+        /// is replaced.
+        /// </summary>
+        public UnknownFieldSet AddField(int number, UnknownField field)
```
jskeet: Calling this AddField is confusing when .NET dictionaries tend to throw if you try to add an existing key.

AddOrReplaceField perhaps?

Additionally, it looks like this could be internal (even then, only visible for testing.)

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+            fields[number] = field;
+            return this;
+        }
+
+        /// <summary>
+        /// Parse an entire message from <paramref name="input"/> and merge
+        /// its fields into this set.
+        /// </summary>
+        public void MergeFrom(CodedInputStream input)
```
jskeet: I believe this can be deleted, if we don't need this to implement IMessage. (And I don't think we do - I've locally removed that, and it looks okay.)

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Parse a single field from <paramref name="input"/> and merge it
+        /// into this set.
+        /// </summary>
+        /// <param name="input">The coded input stream containing the field</param>
+        /// <returns>false if the tag is an "end group" tag, true otherwise</returns>
+        public bool MergeFieldFrom(CodedInputStream input)
```
jskeet: Make this private? It doesn't look like it's used outside this class.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -0,0 +1,364 @@
...
+            unknownFields.MergeFieldFrom(input);
+            return unknownFields;
+        }
+
+        /// <summary>
+        /// Merges the fields from <paramref name="other"/> into this set.
+        /// If a field number exists in both sets, the values in <paramref name="other"/>
+        /// will be appended to the values in this set.
+        /// </summary>
+        public UnknownFieldSet MergeFrom(UnknownFieldSet other)
```
jskeet: This can be private too (with one test change)

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -0,0 +1,269 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Computes the number of bytes required to encode this field, including field
+        /// number.
+        /// </summary>
+        public int GetSerializedSize(int fieldNumber)
+        {
+            int result = 0;
```
jskeet: Compute the tag size once? No need to do it for each iteration. Then you can change each loop to something like:

```csharp
result += tagSize * varintList.Count;
foreach (ulong value in varintList)
{
    result += CodedOutputStream.ComputeUInt64Size(value);
}
```

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -0,0 +1,269 @@
...
+                    result += CodedOutputStream.ComputeTagSize(fieldNumber);
+                    result += CodedOutputStream.ComputeUInt64Size(value);
+                }
+            }
+            if (fixed32List != null)
+            {
+                foreach (uint value in fixed32List)
+                {
+                    result += CodedOutputStream.ComputeTagSize(fieldNumber);
+                    result += CodedOutputStream.ComputeFixed32Size(value);
```
jskeet: Surely Fixed32 and Fixed64 always have the same size, don't they? No need to loop at all here.

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -0,0 +1,269 @@
...
+                foreach (ByteString value in lengthDelimitedList)
+                {
+                    result += CodedOutputStream.ComputeTagSize(fieldNumber);
+                    result += CodedOutputStream.ComputeBytesSize(value);
+                }
+            }
+            return result;
+        }
+
+        /// <summary>
```
jskeet: Currently all of the methods below here can be internal, can't they? (I'm generally in favour of keeping things internal or private until we definitely need them to be public.)

csharp/src/Google.Protobuf/UnknownFieldSet.cs
anandolee: Done

anandolee:
It is not required to provide APIs to get Unknown Field. Some languages like c++ have this API, some(like python) do not provide it. We could not provide it at least now

anandolee:
Have changed the 3 methods to internal.

Should also clone the unknown fields. I've added the corresponding code and tests. Sorry I missed it.

We don't have a clear timeline for the migration, this is the java version:
https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L470

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -424,7 +424,10 @@ public void SkipLastField()
             }
         }
 
-        private void SkipGroup(uint startGroupTag)
+        /// <summary>
+        /// Skip a group.
+        /// </summary>
+        public void SkipGroup(uint startGroupTag)
```
jskeet: Does this need to be public now? We're skipping it within the library, aren't we? Maybe just internal?

anandolee: Done

anandolee:
For other languages, we have released a version that have PreserveUnknownsDefault (default false). Thus users can migrate their code. Then release a version with PreserveUnknownsDefault  true and the last step is remove it (None of them have removed PreserveUnknownsDefault).

So we are indeed going to release a version which have PreserveUnknownsDefault.

jskeet:
> So we are indeed going to release a version which have PreserveUnknownsDefault.

Then removing the property later will be a breaking change, meaning that the library should be revved to version 4.x to follow semantic versioning. I'm very, very keen on not breaking compatibility like this. Changing the default over time seems reasonable, but I'm nervous over it being part of the public API.

One alternative option would be to make it an internal property that could be controlled via an external mechanism, e.g. an environment variable. That's not _great_ either, but at least it doesn't completely violate compatibility.

anandolee:
Just discussed with Jisi. He thinks it is OK to change the behavior directly(always preserve) and add a DiscardUnknownFields(), as C# does not have to support internal usages. Does this make sense?

jskeet:
So remove the property, change codegen as if it were always true, and add the method? Makes sense to me.

Unfortunately we can't add the `DiscardUnknownFields` method to the interface itself (again, a breaking change) - that's the next piece of design work, I guess. Maybe make the "just always preserve them" change, get that merged, and then work on discarding?

anandolee:
Have removed the DiscardUnknownFields, so the behavior is changed to always preserve.
Sounds good to work on discarding after this merge.

anandolee:
ok to test

PR #3938, opened 2017-11-22
Squash CMake CMP0048 warnings when using add_subdirectory

hjmallon:
CMP0048 warns about project_VERSION variables not being set in the NEW way. This happens when using protobuf as a submodule in a larger build. GTest and RE2 have adopted
similar patches.
See https://cmake.org/cmake/help/v3.10/policy/CMP0048.html

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -8,34 +8,6 @@ endif()
 # CMake policies
 cmake_policy(SET CMP0022 NEW)
 
-# Project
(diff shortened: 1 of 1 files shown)


hjmallon:
I have rebased in the hope that the Travis dotnet fix will mean the tests pass.

hjmallon:
Fixed by
https://github.com/google/protobuf/commit/1ec9beb477a2c8c4e77b6e6c9046ea634019644d

PR #3939, opened 2017-11-22
Make cached hash value transient

mairbek:


xfxyjwf:
Sorry for the late response. I don't think it matters to add transient there though as protobuf has its own serialization methods.

PR #3959, opened 2017-11-27
Allow --dependency_out for multiple input files

kkm000:
Ref. https://github.com/google/protobuf/issues/3843#issuecomment-346173274

@xfxyjwf, @TeBoring, can you think of any downside of enabling this?

/cc @jtattermusch @jskeet

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -1287,12 +1287,6 @@ CommandLineInterface::ParseArguments(int argc, const char* const argv[]) {
               << std::endl;
     return PARSE_ARGUMENT_FAIL;
   }
-  if (!dependency_out_name_.empty() && input_files_.size() > 1) {
-    std::cerr
-        << "Can only process one input file when using --dependency_out=FILE."
-        << std::endl;
-    return PARSE_ARGUMENT_FAIL;
-  }
   if (imports_in_descriptor_set_ && descriptor_set_out_name_.empty()) {
     std::cerr << "--include_imports only makes sense when combined with "
                  "--descriptor_set_out." << std::endl;
--- a/src/google/protobuf/compiler/command_line_interface_unittest.cc
+++ b/src/google/protobuf/compiler/command_line_interface_unittest.cc
@@ -1491,21 +1491,33 @@ TEST_F(CommandLineInterfaceTest, WriteTransitiveDescriptorSetWithSourceInfo) {
 // TODO(teboring): Figure out how to write test on windows.
 #else
 TEST_F(CommandLineInterfaceTest, WriteDependencyManifestFileGivenTwoInputs) {
-  CreateTempFile("foo.proto",
+  CreateTempFile("common.proto",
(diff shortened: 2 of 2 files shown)


TeBoring:
Could you add a test: A depends on C, B depends on D.
On command line, you give A and B together. In the output, does A only depends on C and B only depends on D?

kkm000:
@TeBoring:
> On command line, you give A and B together. In the output, does A only depends on C and B only depends on D?

Exactly as I expected, everything depends on everything in a single set:
```
a.pb.cc \
a.pb.h \
b.pb.cc \
b.pb.h: c.proto\
 a.proto\
 d.proto\
 b.proto
```
Not the best way to handle it, naturally. Minimizing into multiple sets would be quite better.

Also, I just noticed the dep file does not have a trailing `'\n'`, which is considered kind of offensive under Un*x. Do you want me to add the final newline, or you think there might be unintended consequences?

Full test setup as I ran it:
```
kkm@yupana:~/work/protobuf/test[dep-allow-many-inputs]$ head *.proto
==> a.proto <==
syntax="proto3";
import "c.proto";
message A {
  C c = 1;
}

==> b.proto <==
syntax="proto3";
import "d.proto";
message B {
  D d = 1;
}

==> c.proto <==
syntax="proto3";
message C {}

==> d.proto <==
syntax="proto3";
message D {}
kkm@yupana:~/work/protobuf/test[dep-allow-many-inputs]$ ../src/protoc -I . --cpp_out=. --dependency_out=manifest a.proto b.proto
kkm@yupana:~/work/protobuf/test[dep-allow-many-inputs]$ cat manifest
a.pb.cc \
a.pb.h \
b.pb.cc \
b.pb.h: c.proto\
 a.proto\
 d.proto\
 b.protokkm@yupana:~/work/protobuf/test[dep-allow-many-inputs]$
```

TeBoring:
The output is not as I expected. For the example, it should be:
```
a.pb.cc a.pb.h: a.proto c.proto
b.pb.cc b.pb.h: b.proto d.proto
c.pb.cc c.pb.h: c.proto
d.pb.cc d.pb.h: d.proto
```
With your output, changes in c.proto will trigger rebuilding of b.pb.cc, which is not wanted.

kkm000:
@TeBoring: I entirely agree, the way I proposed is far from ideal. It guarantees listing all the real dependencies but does err on the side of overgeneralizing. From my understanding of how protoc works, there is a single set of resolved references maintained internally for a single invocation, and transitivity information is not maintained in it. What I mean by transitivity is additional bits of information, like that `C` is here because of `A` and, say, (hypothetical, not in our test), `E`) . Am I understanding this correctly? If that is correct, then changing that looks like a very big rework.

TeBoring:
Your understanding is correct. I think that's why we didn't support multiple invocation at the first time. I don't think we can allow the overgeneralizing version. Do you have time to implement the ideal version?

kkm000:
@TeBoring, I am really swamped now, sorry. I'll keep that change in mind that for later, when I have time, but cannot really even estimate now when that "later" is going to happen. :(

PR #3960, opened 2017-11-28
Added back in LIBPROTOC_EXPORT which was removed by mistake

acozzette:


jleni:
I see that you also hit the dotnet problem with your PR. I have submitted a fix here PR #3962 as it is affecting my other PR too.

PR #3963, opened 2017-11-28
Add "src/" for proto_library() of well known types

fahhem:
Otherwise `proto_library()` is looking in the wrong path for the well known types, so you can't depend on `@com_google_protobuf//:descriptor_proto`


fahhem:
Never mind, I was running an old version of bazel. TIme to open an issue in bazel to add a rule for checking the version, so you can specify a minimum bazel version...

PR #3968, opened 2017-11-28
Add a check_version() for use in WORKSPACEs

fahhem:


examples/WORKSPACE
```diff
@@ -5,6 +5,8 @@ http_archive(
     strip_prefix = "protobuf-master",
     urls = ["https://github.com/google/protobuf/archive/master.zip"],
 )
+load("@com_google_protobuf//:protobuf.bzl", "check_version")
```
xfxyjwf: Can this be done in the BUILD file instead?

protobuf.bzl
```diff
@@ -400,3 +400,10 @@ def internal_protobuf_py_tests(
         srcs=[s],
         main=s,
         **kargs)
+
+
+def check_version():
```
xfxyjwf: It seems better to specify the required minimum version number in the BUILD file instead of here. How about change this to:

```
def require_bazel_version(version):
  ...
```

And in the BUILD file do:

```
load("@com_google_protobuf//:protobuf.bzl", "require_bazel_version")
require_bazel_version("0.5.4")
```

examples/WORKSPACE
fahhem: No, `native.bazel_version` is only available in WORKSPACE-included skylark macros

protobuf.bzl
fahhem: I wish, but bazel doesn't expose its version in the BUILD rules, only in the WORKSPACE

xfxyjwf: Can you change this to `require_bazel_version()` and do `require_bazel_version("0.5.4")` in the toplevel WORKSPACE file?

fahhem: That WORKSPACE file doesn't affect users at all, right? Like if in my WORKSPACE I pull in com_google_protobuf, bazel never looks at your WORKSPACE file

xfxyjwf: Really? I think bazel has to look at protobuf's WORKSPACE file because otherwise it won't know what library protobuf depends on and won't be able to build protobuf.

fahhem: I wish, I have to have the `python_header` bind in my WORKSPACE too. bazel only looks at the top-level WORKSPACE and all loaded BUILD and .bzl files. Plenty of other repositories do something like this:

```py
git_repository(
    name = "io_bazel_rules_pex",
    remote = "https://github.com/fahhem/bazel_rules_pex.git",
    tag = "applied-1.0",
)
load("@io_bazel_rules_pex//pex:pex_rules.bzl", "pex_repositories")
pex_repositories()
```

That way they put all their dependencies into a bzl file, and include that in their own WORKSPACE and all users' WORKSPACEs

fahhem:
PTAL

protobuf.bzl
xfxyjwf: Can you change the name of this function to `check_protobuf_required_bazel_version`?

xfxyjwf: Also please add comment for this function to explain how this function is supposed to be used.

fahhem: Done, thanks! PTAL

xfxyjwf:
Thanks! The patch now looks good to me. Could you squash the commits into 1? Also please double check the commit's author email. It seems the CLA bot has detected a different email address used in your commits.

fahhem:
GitHub can squash into a single commit when merging, or would you rather I did it?

Yes, the commit author changes between my personal and work email, but they're all `fahhem@` and me. I'm authorized to submit this with my work email

xfxyjwf:
Can you do the squash? Make sure to update the commit message appropriately.

fahhem:
Squashed and ready to go!

xfxyjwf:
Thanks!

PR #3972, opened 2017-11-29
php: Added nanosecond support for Timestamp

lhecker:
This PR adds support for the nanoseconds field of the wellknown `Timestamp` type for PHP 5.6 and later.
The accuracy will be in the order of microseconds though, as that's the highest resolution PHP supports.

---

I hope this PR fullfills all necessary preconditions to be merged, but if it doesn't (e.g. because the `Timestamp` class is generated automatically), I'd be glad to help get this feature implemented whatever is necessary for that to happen. 🙂

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -30,6 +30,7 @@
 
 #include <php.h>
 #include <stdlib.h>
+#include <inttypes.h>
 
 #include "protobuf.h"
 #include "utf8.h"
@@ -1249,28 +1250,62 @@ PHP_METHOD(Timestamp, fromDateTime) {
     return;
   }
 
-  // Get timestamp from Datetime object.
-  zval retval;
-  zval function_name;
-  int64_t timestamp;
+  int64_t timestamp_seconds;
+  {
+    zval retval;
+    zval function_name;
 
 #if PHP_MAJOR_VERSION < 7
-  INIT_ZVAL(retval);
-  INIT_ZVAL(function_name);
+    INIT_ZVAL(retval);
+    INIT_ZVAL(function_name);
 #endif
 
-  PHP_PROTO_ZVAL_STRING(&function_name, "date_timestamp_get", 1);
+    PHP_PROTO_ZVAL_STRING(&function_name, "date_timestamp_get", 1);
 
-  if (call_user_function(EG(function_table), NULL, &function_name, &retval, 1,
-          ZVAL_PTR_TO_CACHED_PTR(datetime) TSRMLS_CC) == FAILURE) {
-    zend_error(E_ERROR, "Cannot get timestamp from DateTime.");
-    return;
+    if (call_user_function(EG(function_table), NULL, &function_name, &retval, 1,
+            ZVAL_PTR_TO_CACHED_PTR(datetime) TSRMLS_CC) == FAILURE) {
+      zend_error(E_ERROR, "Cannot get timestamp from DateTime.");
+      return;
+    }
+
+    protobuf_convert_to_int64(&retval, &timestamp_seconds);
+
+    zval_dtor(&retval);
+    zval_dtor(&function_name);
   }
 
-  protobuf_convert_to_int64(&retval, &timestamp);
+  int64_t timestamp_micros;
+  {
+    zval retval;
+    zval function_name;
+    zval format_string;
 
-  zval_dtor(&retval);
-  zval_dtor(&function_name);
+#if PHP_MAJOR_VERSION < 7
+    INIT_ZVAL(retval);
+    INIT_ZVAL(function_name);
+    INIT_ZVAL(format_string);
+#endif
+
+    PHP_PROTO_ZVAL_STRING(&function_name, "date_format", 1);
+    PHP_PROTO_ZVAL_STRING(&format_string, "u", 1);
+
+    CACHED_VALUE params[2] = {
+      ZVAL_PTR_TO_CACHED_VALUE(datetime),
+      ZVAL_TO_CACHED_VALUE(format_string),
+    };
+
+    if (call_user_function(EG(function_table), NULL, &function_name, &retval,
+            ARRAY_SIZE(params), params TSRMLS_CC) == FAILURE) {
+      zend_error(E_ERROR, "Cannot format DateTime.");
+      return;
+    }
+
+    protobuf_convert_to_int64(&retval, &timestamp_micros);
+
+    zval_dtor(&retval);
+    zval_dtor(&function_name);
+    zval_dtor(&format_string);
+  }
 
   // Set seconds
   MessageHeader* self = UNBOX(MessageHeader, getThis());
@@ -1278,13 +1313,13 @@ PHP_METHOD(Timestamp, fromDateTime) {
       upb_msgdef_ntofz(self->descriptor->msgdef, "seconds");
   void* storage = message_data(self);
   void* memory = slot_memory(self->descriptor->layout, storage, field);
-  *(int64_t*)memory = timestamp;
+  *(int64_t*)memory = timestamp_seconds;
 
   // Set nanos
   field = upb_msgdef_ntofz(self->descriptor->msgdef, "nanos");
   storage = message_data(self);
   memory = slot_memory(self->descriptor->layout, storage, field);
-  *(int32_t*)memory = 0;
+  *(int32_t*)memory = timestamp_micros * 1000;
 
   RETURN_NULL();
 }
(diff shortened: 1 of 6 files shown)


TeBoring:
Could you check the c extension also works? https://github.com/google/protobuf/tree/master/php/ext/google/protobuf

lhecker:
@TeBoring You're right: The `fromDateTime`/`toDateTime` functions are actually also implemented in the native PHP extension. Didn't realize that. 😐

Can you run kokoro on this again?

----

I tried my best converting my PHP logic into C and also improving the existing C code a little bit (e.g. by not relying on the flaky `gmtime()` function, or by employing the `ARRAY_SIZE` macro).

But if possible someone more experienced in PHP extensions has to look over this PR as I was basically only poking around in the dark here (I really don't have any clue how to write PHP extensions).

TeBoring:
```
Warning: assert(): assert(0 == $timestamp->getNanos()) failed in /home/teboring/protobuf-timestamp/php/tests/memory_leak_test.php on line 155
==14741== 
==14741== HEAP SUMMARY:
==14741==     in use at exit: 77,591 bytes in 42 blocks
==14741==   total heap usage: 42,875 allocs, 42,833 frees, 9,228,805 bytes allocated
==14741== 
==14741== 32 bytes in 1 blocks are definitely lost in loss record 15 of 42
==14741==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14741==    by 0x7039E8: __zend_malloc (zend_alloc.c:2820)
==14741==    by 0x8FEF208: zend_string_alloc (zend_string.h:122)
==14741==    by 0x8FEF208: zend_string_init (zend_string.h:158)
==14741==    by 0x8FEF208: zim_Timestamp_fromDateTime (message.c:1289)
==14741==    by 0x7D020F: ZEND_DO_FCALL_SPEC_RETVAL_UNUSED_HANDLER (zend_vm_execute.h:970)
==14741==    by 0x77753A: execute_ex (zend_vm_execute.h:429)
==14741==    by 0x7D21AF: zend_execute (zend_vm_execute.h:474)
==14741==    by 0x72DBF2: zend_execute_scripts (zend.c:1476)
==14741==    by 0x6CA37F: php_execute_script (main.c:2537)
==14741==    by 0x7D4435: do_cli (php_cli.c:993)
==14741==    by 0x432EFB: main (php_cli.c:1381)
```

php/ext/google/protobuf/message.c
```diff
@@ -1248,42 +1249,75 @@ PHP_METHOD(Timestamp, fromDateTime) {
...
-  zval_dtor(&retval);
-  zval_dtor(&function_name);
+#if PHP_MAJOR_VERSION < 7
+    INIT_ZVAL(retval);
+    INIT_ZVAL(function_name);
+    INIT_ZVAL(format_string);
+#endif
+
+    PHP_PROTO_ZVAL_STRING(&function_name, "date_format", 1);
+    PHP_PROTO_ZVAL_STRING(&format_string, "u", 1);
```
TeBoring: You forget to call ```zval_dtor(&format_string);``` which causes memory leak.

php/ext/google/protobuf/message.c
```diff
@@ -1248,42 +1249,75 @@ PHP_METHOD(Timestamp, fromDateTime) {
...
+
+    CACHED_VALUE params[2] = {
+      ZVAL_PTR_TO_CACHED_VALUE(datetime),
+      ZVAL_TO_CACHED_VALUE(format_string),
+    };
+
+    if (call_user_function(EG(function_table), NULL, &function_name, &retval,
+            ARRAY_SIZE(params), params TSRMLS_CC) == FAILURE) {
```
TeBoring: Instead of using ARRAY_SZIE, why not using actual number?

php/ext/google/protobuf/protobuf.h
```diff
@@ -42,6 +42,10 @@
 #define MAX_LENGTH_OF_INT64 20
 #define SIZEOF_INT64 8
 
+/* From Chromium. */
+#define ARRAY_SIZE(x) \
+    ((sizeof(x)/sizeof(0[x])) / ((size_t)(!(sizeof(x) % sizeof(0[x])))))
```
TeBoring: Why ```ARRAY_SIZE``` is like this? Is it related to implementation detail of php runtime?

lhecker: I've taken that macro from here: https://github.com/google/protobuf/blob/3beb19945b599d448c2524666effe1acdf29165f/php/ext/google/protobuf/upb.c#L8107-L8109

It's a classical macro to determine the size of constant-sized arrays in C.
For instance if you have a `char foo[25]` then `ARRAY_SIZE(foo)` will evaluate to `25`.

You can find an explanation here: https://stackoverflow.com/a/4415646

Since such a macro is much safer than hardcoding the size of arrays I've started replacing these hardcoded numbers in any PHP-extension code I touched, but I'd recommend using that macro in other places as well as soon as possible. It will probably reduce accidential mistakes in future PRs. 🙂

php/ext/google/protobuf/message.c
lhecker: See my explanation for `ARRAY_SIZE` below. 🙂
Should I change it back?

php/ext/google/protobuf/message.c
lhecker: Fixed.

php/ext/google/protobuf/message.c
TeBoring: Thanks for the explanation. ARRAY_SIZE looks good to me.

TeBoring:
```Warning: assert(): assert(0 == $timestamp->getNanos()) failed in /home/teboring/protobuf-timestamp/php/tests/memory_leak_test.php on line 155```

You probably also need to fix the memory_leak_test.

**Comment by TeBoring** [comment]
any update?

**Comment by lhecker** [comment]
@TeBoring Sorry I'm pretty busy at the moment and forgot about this PR for the last few days.
I "allocated" myself some time tomorrow though to finish up the remaining issues in this PR. 🙂
(It'd be nice if I can get valgrind running on macOS though. 🤔)

**Comment by TeBoring** [comment]
The broken test is not about memory anymore. You can just run it with php.

**Comment by lhecker** [comment]
@TeBoring I fixed this one remaining issue just now. Sorry that you had to wait that long. 🙂

**Comment by TeBoring** [comment]
For php 7.1, c extension test:
```
1) WellKnownTest::testTimestamp
Failed asserting that 12345000 is identical to 0.
/tmp/protobuf/protobuf/php/tests/protobuf/php/tests/well_known_test.php:315
```

lhecker:
@TeBoring You (or Kokoro) seem to have made a mistake when testing my PR:
You're using the **new** C extension (= from this branch) which correctly returns `12345000` for `getNanos()`, but you're using the **old**  phpunit test (= from master branch or older) for testing.

---

Proof for my "accusation":
The message
> `Failed asserting that <actual> is identical to <expected>.`

is generated by phpunit's `assertSame()` function.
But if you check the diff of this PR you'll notice that I replaced the `assertSame()` in `well_known_test.php:315`.
It uses `assertEqual()` now instead, which would print an error message like this:
> `Failed asserting that <actual> matches expected <expected>.`

If I'm correct this means that you're trying to test the changed C extension using the old test code, which is not possible in this case.

TeBoring:
Ah, I think you are right. That is our compatibility test.
https://github.com/google/protobuf/blob/master/php/tests/compatibility_test.sh#L123
Here we have some sed script stuff. Could you add one to change the broken test?

TeBoring:
I doubt it's our compatibility test.
https://github.com/google/protobuf/blob/master/php/tests/compatibility_test.sh#L123
Here we have some sed script to fix compatibility test. Could you help add one to fix the broken case?

lhecker:
@TeBoring My fork was out of date and would've resulted in a merge conflict if I touch that file.
I've thus taken the liberty to rebase my fork on the current master.

I sadly didn't quite understand your last comment.
So I went ahead and [wrote a `sed` command](https://github.com/google/protobuf/pull/3972/commits/423e0bb89807eb4aefc3313b29d44dd65d14f773), which simply removes the entire `testTimestamp` function from the test file.

I feel like this solution is _possibly_ a bit like shooting sparrows with cannons, but I didn't had any better idea. Do you happen to have one?

TeBoring:
@lhecker Thanks for the fix. There were some incorrect test. When we update the against compatibility tests, we will have the correct test. Before that, we just manually disable them. I think your approach is acceptable, because that test is truly not correct.

lhecker:
@TeBoring Thanks for merging this and helping me ironing out all these issues in my PR. 🙂🎉

PR #3976, opened 2017-11-30
Provide discardUnknonwnFields API in php

TeBoring:


php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -576,6 +576,15 @@ public function clear()
         }
     }
 
+    /**
+     * Clear all unknown fields previously parsed.
+     * @return null.
+     */
+    public function discardUnknownFields()
+    {
+        $this->unknown = "";
```
pherl: it should recursively clear all the unknown fields in the sub messages.

TeBoring: What about map fields?

TeBoring: does map entry itself preserve unknown?

PR #3981, opened 2017-12-01
Handle srcs in generated files by cd'ing in and out

### LINKED ISSUE #1313: py_proto_library gets confused when a filegroup is provided for srcs
Currently, it is not possible to do the following:
## Repository `foo`

`BUILD`:

``` python
filegroup(
    name = "foo_pb_srcs",
    srcs = ["foo.proto"],
)
```
## Repository `bar`

`WORKSPACE`:

``` python
git_repository(
    name = "foo"
    remote = "..."
   ...
)
```

`BUILD`:

``` python
load("@protobuf//:protobuf.bzl", "py_proto_library")

py_proto_library(
    name = "foo_pb",
    srcs = ["@foo//:foo_pb_srcs"],
)
```

I attempted to have this setup where I add a `filegroup` to `//src/main/protobuf` in the Bazel repository for `build.proto` and then have a `py_proto_library` target in a different repository that uses that filegroup (which is available in `@bazel_tools//`. This is the error that I get:

```
❯❯❯ bazel build //skydoc:build_pb
Extracting Bazel installation...
Sending SIGTERM to previous Bazel server (pid=10593)... done.
.......
ERROR: /Users/dzc/Sandbox/skydoc/skydoc/BUILD:5:1: //skydoc:build_pb_genproto: label '@bazel_tools//src/main/protobuf:build_p_pb2.py' is not in the current package.
ERROR: no such package 'skydoc': Package 'skydoc' contains errors.
INFO: Elapsed time: 1.493s
```

### LINKED ISSUE #1314: py_proto_library gets confused when source file is in genfiles
This is a similar issue to #1313. The main difference is that the source file is generated in genfiles:

``` python
load("@protobuf//:protobuf.bzl", "py_proto_library")

genrule(
    name = "foo_pb_srcs",
    srcs = ["@foo//:foo_pb_srcs"],  # this target is a filegroup
    outs = ["foo.proto"],
    cmd = "cp -f $< $@",
)

py_proto_library(
    name = "foo_pb",
    srcs = ["foo.proto"],
)
```

fahhem:
The change:
* Each source file is compiled separately now
* If a source file is generated, we cd into its folder so protoc can see it, then move it out

The downside is that a separate protoc call is made per file now, which is a small decrease in performance. An optimization can be made to combine all non-generated files into a single call.

I tried a few approaches over the past 2 days, all of them failed for some reason or another (mostly because bazel is super uptight about things):
* Explicitly create an intermediate file and move it (only for generated inputs): Fails because bazel refuses to let me create a `File()` that isn't relative to my package. protoc will output the file into `bazel-out/.../genfiles/bazel_out/.../genfiles/<path>/<filename>_pb2.py` but the closest I could get was a declared file (via `ctx.actions.declare_file()`) that was `bazel-out/.../genfiles/<path>/bazel_out/.../genfiles/<path>/...` and that's got <path> in there twice and you can't change the filename at all, ever, in any way, from Skylark.
* A single protoc call. Figuring out where things are output and need to be moved to is a mess and you're better off with the performance hit if you don't want to mess up the sandbox too much.

Fixes #1314 and fixes #1313

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -913,3 +913,30 @@ objc_library(
     non_arc_srcs = OBJC_SRCS,
     visibility = ["//visibility:public"],
 )
+
+################################################################################
+# Test generated proto support
+################################################################################
+
+genrule(
+    name = "generated_protos",
+    srcs = ["src/google/protobuf/unittest_import.proto"],
+    outs = ["unittest_gen.proto"],
+    cmd = "cat $(SRCS) | sed 's|google/|src/google/|' >  $(OUTS)"
+)
+
+proto_library(
+    name = "generated_protos_proto",
+    srcs = ["unittest_gen.proto"],
+)
+
+
+py_proto_library(
+    name = "generated_protos_py",
+    srcs = [
+        "unittest_gen.proto",
+        "src/google/protobuf/unittest_import_public.proto",
+    ],
+    default_runtime = "",
+    protoc = ":protoc",
+)
--- a/BUILD
+++ b/BUILD
@@ -927,9 +927,11 @@ genrule(
 
 proto_library(
     name = "generated_protos_proto",
-    srcs = ["unittest_gen.proto"],
+    srcs = [
+        "unittest_gen.proto",
+        "src/google/protobuf/unittest_import_public.proto",
+    ],
 )
-
 
 py_proto_library(
     name = "generated_protos_py",
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -72,7 +72,7 @@ def _proto_gen_impl(ctx):
   deps = []
   deps += ctx.files.srcs
   source_dir = _SourceDir(ctx)
-  gen_dir = _GenDir(ctx)
+  gen_dir = _GenDir(ctx).rstrip('/')
   if source_dir:
     import_flags = ["-I" + source_dir, "-I" + gen_dir]
   else:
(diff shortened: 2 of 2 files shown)


fahhem:
Current patch verified working using this release: https://github.com/fahhem/protobuf/releases/tag/v3.5.0-applied1

fahhem:
PTAL?

fahhem:
ping

fahhem:
Any chance of this getting in?

pherl:
OK to test

pherl:
Generated file means the generated `.proto` file? Could you add more details about what the PR changes in the PR description? Probably also add a test. Like a rule in the BUILD of protobuf package itself to exercise the new logic. 

For per .proto file protoc invocation, can you make sure that the files in the same srcs are still cross-importable?

pherl:
Note to self: We should probably create a bzl test package and have a couple of downstream integration tests to be more confident about merging bzl changes.

fahhem:
I've added the details and the test rules, though I'm not sure you want them in the root BUILD file.

fahhem:
@pherl Do you have time to comment on this?

fahhem:
It already does that, and even if I add it on, it doesn't work. It can't find the non-generated files because those aren't included correctly.

fahhem:
The main problem is that by default, it generates the wrong files than what bazel expects.

anandolee:
@xfxyjwf can you help to take a review

xfxyjwf:
@davidzchen David, since you filed https://github.com/google/protobuf/issues/1314 originally, could you also help review this .bzl change? I don't think anybody on protobuf-team has enough knowledge/experience to maintain this py_proto_library rule.

PR #3982, opened 2017-12-01
Recursively clear unknown fields in submessages.

TeBoring:


php/tests/encode_decode_test.php
```diff
@@ -473,6 +473,27 @@ public function testUnknown()
         $m->discardUnknownFields();
         $to = $m->serializeToString();
         $this->assertSame("", bin2hex($to));
+
+        $m = new TestMessage();
```
pherl: add some comments about what those test cases are.

php/tests/encode_decode_test.php
```diff
@@ -443,36 +443,65 @@ public function testDecodeInvalidPackedMessageLength()
...
 
+        // Test discard unknown in message.
         $m = new TestMessage();
         $from = hex2bin('F80601');
         $m->mergeFromString($from);
         $m->discardUnknownFields();
         $to = $m->serializeToString();
         $this->assertSame("", bin2hex($to));
+
+        // Test discard unknown for singular message field.
```
pherl: you mean the unknown fields inside a singular message field? I'm wondering if a simple util that can encode varint would make the test case much easer to read.  Probably leave this as a TODO.

PR #3985, opened 2017-12-01
BUILD,globs: ignore *.pb.{h,cc} in the source tree

laszlocsomor:
The globs were picking up such files and created
dependency cycles so I could no longer build
//:protoc on Windows.

This is what happened:

1. I built the tests with CMake. As part of that,
   CMake generated header files in the source
   tree, such as
   "src/google/protobuf/unittest_lite.pb.cc"

2. I attempted to build "//:protoc" with Bazel:

2.1. As part of loading the BUILD file, Bazel
     expanded glob() statements in the
     "protobuf" cc_library's hdrs attribute.

2.2. The result of glob() expansion is a list of
     labels, and the expanded list included
     "src/google/protobuf/unittest_lite.pb.cc"

2.3. After glob() expansion, Bazel parsed all rule
     definitions. It found the
     "cc_test_protos_genproto" rule which
     generates a file called
     "src/google/protobuf/unittest_lite.pb.cc"

2.4. Because Bazel can't tell from a label whether
     it refers to a source file or a generated
     file or a rule -- after all, srcs=["foo.cc"]
     could refer to either the source file
     "foo.cc", or the generated file "foo.cc", or
     the rule called "foo.cc" -- Bazel didn't
     realise that it should depend on the source
     file "unittest_lite.pb.cc", not on the output
     file "unittest_lite.pb.cc". (Or that it
     should error out because of the ambiguity.)

2.5. Building the generated file
     "unittest_lite.pb.cc" however involves
     running the protocol compiler, which involes
     building ":protoc", which depends on the
     ":protobuf" rule, thus we now have a
     dependency cycle.

3. It made debugging more difficult that the
   .gitignore file includes a pattern to match the
   CMake-generated headers, so I didn't see any
   new files in "git status", so I initially ruled
   out globs as the source of the problem.

Takeaway:
- Bazel should error out if labels are ambiguous.
- As a workaround, the glob() should use exclude
  statements. This commit adds that.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -111,7 +111,12 @@ cc_library(
         "src/google/protobuf/stubs/time.cc",
         "src/google/protobuf/wire_format_lite.cc",
     ],
-    hdrs = glob(["src/google/protobuf/**/*.h"]),
+    hdrs = glob(
+        ["src/google/protobuf/**/*.h"],
+        exclude = [
+            "src/google/protobuf/**/*.pb.h",
+        ],
+    ),
     copts = COPTS,
     includes = ["src/"],
     linkopts = LINK_OPTS,
@@ -178,7 +183,12 @@ cc_library(
         "src/google/protobuf/wire_format.cc",
         "src/google/protobuf/wrappers.pb.cc",
     ],
-    hdrs = glob(["src/**/*.h"]),
+    hdrs = glob(
+        ["src/**/*.h"],
+        exclude = [
+            "src/**/*.pb.h",
+        ],
+    ),
     copts = COPTS,
     includes = ["src/"],
     linkopts = LINK_OPTS,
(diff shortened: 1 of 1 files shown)


pherl:
What about those checked-in generated files like descriptor.proto and well known types?

Bazel build requires the source tree to be clean. Do you need to use cmake and bazel together?

laszlocsomor:
My change only affects globs that match `.cc` and `.h` files.

> Bazel build requires the source tree to be clean.

Not necessarily. This PR fixes a corner case by more precisely specifying what a glob should match.

> Do you need to use cmake and bazel together?

Normally you don't. I ran CMake because I wanted to build the tests to repro the Appveyor test failures.

laszlocsomor:
ping

xfxyjwf:
@laszlocsomor Sorry for the late response. Could you rebase this pull request against the current master (so the new bazel test I added not very long ago will be ran for this PR)?

laszlocsomor:
> Could you rebase this pull request against the current master 

Done.

xfxyjwf:
@laszlocsomor The bazel test is failing. Can you take a look?

xfxyjwf:
Closing as it will cause our bazel test to fail. Feel free to resent.

laszlocsomor:
Sorry I never got back to you. I tried running the tests but couldn't figure out what was wrong. Anyway, I have no intention to pick this up any time soon again so closing it is fine. Thanks!

PR #3990, opened 2017-12-02
Add discard unknown API in ruby.

TeBoring:


ruby/tests/encode_decode_test.rb
```diff
@@ -0,0 +1,50 @@
...
+require 'test/unit'
+
+def hex2bin(s)
+  s.scan(/../).map { |x| x.hex.chr }.join
+end
+
+class EncodeDecodeTest < Test::Unit::TestCase
+  def test_discard_unknown
+    # Test discard unknown in message.
+    from = hex2bin('F80601')
```
haberman: The hex strings aren't readable. Could you create the proto by using a different message which has that field defined? Then you could serialize it.

PR #3993, opened 2017-12-03
Include googletest as a submodule

### LINKED ISSUE #236: Add gtest as a git submodule.
from https://github.com/google/protobuf/blob/master/vsprojects/readme.txt:

"Check whether a gtest directory exists in the upper level directory. If
   you checkout the code from github via "git clone", this gtest directory
   won't exist and you won't be able to build the tests described below. To
   avoid this problem consider downloading one of the release tar balls which
   contains gtest already and copying the gest directory from there to your
   protobuf directory".

I would argue that adding gtest as a git submodule would be a much better practice here. 
Then user would be able to grab the desired version of gtest using just one git command.

See what gRPC is doing:
https://github.com/grpc/grpc/tree/master/third_party

coryan:
This fixes #236.  Currently it patches googletest to support autoconf / automake builds, because google/googletest#776.  I have submitted google/googletest#1341 to fix that problem, but until it is accepted the patches work around the underlying googletest issue.

The builds with cmake and bazel do not need such hacks.


coryan:
@xfxyjwf PTAL.  The changes in googletest are in, it works correctly with autoconf/automake now, and the builds are passing.

.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "third_party/benchmark"]
 	path = third_party/benchmark
 	url = https://github.com/google/benchmark.git
+[submodule "third_party/googletest"]
```
xfxyjwf: Can you pin this submodule to a specific googletest version (a tag or a commit)?

cmake/tests.cmake
```diff
@@ -1,24 +1,26 @@
...
+set(googlemock_source_dir "${protobuf_source_dir}/third_party/googletest/googlemock")
+set(googletest_source_dir "${protobuf_source_dir}/third_party/googletest/googletest")
 include_directories(
-  ${protobuf_source_dir}/gmock
-  ${protobuf_source_dir}/gmock/gtest
-  ${protobuf_source_dir}/gmock/gtest/include
-  ${protobuf_source_dir}/gmock/include
+  ${googlemock_source_dir}
+  ${googletest_source_dir}
+  ${googletest_source_dir}/include
```
xfxyjwf: Remove duplicated lines here.

.gitmodules
coryan: I believe it is already.  The SHA1 of the commit a submodule points to is stored directly in the git database:

https://stackoverflow.com/a/5033973

You can see that it already points at a SHA1 here:

https://github.com/coryan/protobuf/tree/fix-issue-236-gmock-as-submodule/third_party

cmake/tests.cmake
coryan: Sorry, I do not see the duplicates.  Line 12 points to google**mock**, while line 13 is google**test**.  Or maybe you wanted be to look elsewhere and I missed it?

coryan:
PTAL.  Ping me if I am just being dumb and do not see what is it you are asking me to do.

coryan:
Ping, is there something else I should be doing here?

PR #4000, opened 2017-12-05
Using binary one's complement to negate an unsigned int

Kwizatz:
This removes a Visual Studio warning:

warning C4146: unary minus operator applied to unsigned type, result
still unsigned.

### DIFF
--- a/src/google/protobuf/wire_format_lite.h
+++ b/src/google/protobuf/wire_format_lite.h
@@ -860,7 +860,7 @@ inline uint32 WireFormatLite::ZigZagEncode32(int32 n) {
 
 inline int32 WireFormatLite::ZigZagDecode32(uint32 n) {
   // Note:  Using unsigned types prevent undefined behavior
-  return static_cast<int32>((n >> 1) ^ -(n & 1));
+  return static_cast<int32>((n >> 1) ^ (~(n & 1) + 1));
 }
 
 inline uint64 WireFormatLite::ZigZagEncode64(int64 n) {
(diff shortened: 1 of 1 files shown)


acozzette:
ok to test (this comment will trigger the other test runs)

acozzette:
@Kwizatz Thanks for the pull request!

Kwizatz:
@acozzette Thanks for your prompt review!

andschwa:
@Kwizatz Thanks for fixing this warning! I was _really_ sick of seeing it.

Kwizatz:
@andschwa You're welcome 😄. I like to keep my code base compiling with -WX (warnings as errors in msvc), so whenever these pop up I do get quite annoeyd too.

andschwa:
@Kwizatz Haha, that's exactly what I'm trying to do with Mesos right now. Just a couple warnings to go. Started with four digits worth...

I grabbed another piece of a commit to fix one other warning.

```
commit 5a76e633e
Author: Adam Cozzette <acozzette@google.com>
Date:   Thu Nov 17 16:48:38 2016 -0800

--- c/src/google/protobuf/repeated_field.h
+++ i/src/google/protobuf/repeated_field.h
@@ -86,7 +86,7 @@ void LogIndexOutOfBounds(int index, int size);

 template <typename Iter>
 inline int CalculateReserve(Iter begin, Iter end, std::forward_iterator_tag) {
-  return std::distance(begin, end);
+  return static_cast<int>(std::distance(begin, end));
 }
```

Kwizatz:
@andschwa Wow! 4 digits? you sure are patient.  I once tried going through all of them, but for some the solution is too intrusive and/or time consuming (moving declspec(export) from a class declaration to specific class members of said class was the worst), some involve a lot of busy work.

So in the end I decided to let the internal warnings be, and disable the handful that come up on generated code via pragmas or -w switches, but the ones that pop up from hand writen code that #include protobuf headers, I do try to fix.

Anyway, best of luck, I for one really appreciate your effort 👍.

PR #4003, opened 2017-12-06
Php memory leak bug

TeBoring:


TeBoring:
Same change has been merged to 3.5.0.1 branch.

PR #4007, opened 2017-12-06
--pre is not necessary

graywolf:
--pre was not necessary to install google-protobuf when I tried it today (2017-12-06).


acozzette:
@graywolf Thanks for the pull request!

PR #4016, opened 2017-12-07
fix undefined behavior in C++03

### LINKED ISSUE #4015: normalize() in io_win32.cc uses undefined behavior in C++03
The `normalize(string path)` function in `io_win32.cc` relies on undefined behavior in C++03. The function loops through `path`, parsing the segments along the way. The loop breaks out when `path[i] == '\0'`. For example, see [line 163](https://github.com/jquesnelle/protobuf/blob/9021f623e1420f513268a01a5ad43a23618a84ba/src/google/protobuf/stubs/io_win32.cc#L163). `path` is a mutable `std::string`. According to C++03, `operator[]` with `size()` as the parameter (i.e. accessing where the null terminator is) only allowed for the `const` version of `operator[]`. 

[C++03](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1905.pdf), page 638:
>const_reference operator []( size_type pos ) const ;
>reference operator []( size_type pos );
> Returns: If pos < size() , returns *(begin() + pos ) . Otherwise, if pos == size() , the const version
>returns charT() . Otherwise, the behavior is undefined.

Since `path` is a mutable string, the compiler chooses the non-const version of `operator[]`, which leads to undefined behavior. I discovered this because Visual Studio 2010 explicitly checks for this condition if `_ITERATOR_DEBUG_LEVEL != 0` and asserts, which is case when making debug builds.

Note: [C++11](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf) changed the wording of this requirement. Page 651:

>const_reference operator[](size_type pos) const;
>reference operator[](size_type pos);
>Requires: pos <= size() .
>Returns: *(begin() + pos) if pos < size() . Otherwise, returns a reference to an object of type
charT with value charT() , where modifying the object leads to undefined behavior.

jquesnelle:
Fixes #4015.

This is a simple if inelegant fix -- force the use of `const operator[]`.

### DIFF
--- a/src/google/protobuf/stubs/io_win32.cc
+++ b/src/google/protobuf/stubs/io_win32.cc
@@ -155,12 +155,13 @@ string normalize(string path) {
 
   static const string dot(".");
   static const string dotdot("..");
+  const string& cpath = path;
 
   std::vector<string> segments;
   int segment_start = -1;
   // Find the path segments in `path` (separated by "/").
   for (int i = 0;; ++i) {
-    if (!is_separator(path[i]) && path[i] != '\0') {
+    if (!is_separator(cpath[i]) && cpath[i] != '\0') {
       // The current character does not end a segment, so start one unless it's
       // already started.
       if (segment_start < 0) {
@@ -169,7 +170,7 @@ string normalize(string path) {
     } else if (segment_start >= 0 && i > segment_start) {
       // The current character is "/" or "\0", so this ends a segment.
       // Add that to `segments` if there's anything to add; handle "." and "..".
-      string segment(path, segment_start, i - segment_start);
+      string segment(cpath, segment_start, i - segment_start);
       segment_start = -1;
       if (segment == dotdot) {
         if (!segments.empty() &&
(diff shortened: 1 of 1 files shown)


jquesnelle:
I signed it!

src/google/protobuf/stubs/io_win32.cc
```diff
@@ -155,12 +155,13 @@ string normalize(string path) {
 
   static const string dot(".");
   static const string dotdot("..");
+  const string& cpath = path;
```
pherl: how about using const char* p looping on path.c_str() instead? This avoids mixed use of cpath and path in the parsing code.

pherl:
@laszlocsomor to also take a look.

src/google/protobuf/stubs/io_win32.cc
jquesnelle: Yeah, that's probably preferable. Also may have better performance than calling `operator[]` every time

pherl:
OK to test.

PR #4021, opened 2017-12-08
Add CLI flag --schemas_out to turn descriptor set back to .protos

djudd-stripe:
This is a proposed feature, on which I'm looking for comment. I haven't implemented tests yet, but will if folks like the idea.

The main use case I have is generating .proto schemas from another, canonical schema definition. Rather than writing out .proto strings myself, it seems safer/nicer to build a descriptor set object using standard setters and then use google/protobuf C++ code to turn that into .proto strings, and I'm not working in C++ so would like to be able to shell out to protoc.

Evidence that I'm not the only person who has wanted this feature: https://stackoverflow.com/questions/19418655/restoring-proto-file-from-descriptor-string-possible

I should be covered by the Stripe CLA, although the bot doesn't always seem to recognize that.

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -189,6 +189,67 @@ bool TryCreateParentDirectory(const string& prefix, const string& filename) {
   return true;
 }
 
+bool WriteSingleFile(const string& prefix, const string& relative_filename,
+    const char* data, int size) {
+
+  if (!TryCreateParentDirectory(prefix, relative_filename)) {
+    return false;
+  }
+  string filename = prefix + relative_filename;
+
+  // Create the output file.
+  int file_descriptor;
+  do {
+    file_descriptor =
+      open(filename.c_str(), O_WRONLY | O_CREAT | O_TRUNC | O_BINARY, 0666);
+  } while (file_descriptor < 0 && errno == EINTR);
+
+  if (file_descriptor < 0) {
+    int error = errno;
+    std::cerr << filename << ": " << strerror(error);
+    return false;
+  }
+
+  // Write the file.
+  while (size > 0) {
+    int write_result;
+    do {
+      write_result = write(file_descriptor, data, size);
+    } while (write_result < 0 && errno == EINTR);
+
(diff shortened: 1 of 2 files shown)


xfxyjwf:
From what I can tell, this is not a common use case of protoc. The interaction between existing protoc command line flags are already complicated and I'm concerned about adding more.

Can you instead just use the DescriptorPool API in your schema generator? The implementation should be pretty straightforward as shown by Kenton on the stackoverflow question.

djudd-stripe:
I suspect it's not a common use case, true. It seems fitting as an inverse to the existing `--descriptor_set_out` flag, but if you want to minimize the interface, I understand.

I'm not sure how the DescriptorPool API alone would help with my use case. I use essentially Kenton's implementation in this PR, and it seems to require the C++-only DebugString method as well. Can I produce actual .proto schemas by that means without C++?

xfxyjwf:
Converting descriptors to proto syntax is only supported in C++ API. If you are using some other language, it should still be easy enough to write a dedicated C++ binary for this purpose.

Also I think you can consider using the descriptor set directly without generating intermediate .proto files since protoc (and proto runtime) can make use of the descriptor set without relying on actual .proto files.

djudd-stripe:
I have, in fact, written a dedicated C++ binary, but I thought it might be useful to others to have it in protoc. If not, ok.

xfxyjwf:
We are moving protobuf to its own organization (see https://github.com/google/protobuf/issues/4796) and after that we should be able to use a separate repo to build a more feature-rich command line tool instead of stretching protoc.

ruanspies:
> I have, in fact, written a dedicated C++ binary, but I thought it might be useful to others to have it in protoc. If not, ok.

Hey @djudd-stripe. It has been a while since this post but was wondering if you could perhaps provide the code for the C++ binary that you built? Even though I would have appreciated the functionality you suggested in this PR, providing the code can at least help people like me to find alternatives.

djudd-stripe:
@ruanspies I believe this was it: https://gist.github.com/djudd-stripe/721edf9fa42edeccd8a786faae4e365c

Issue #3929, opened 2017-11-21
[3.5.0] undefined symbol timelib_update_ts


remicollet:
Tyring to load the 3.5.0 extension
`PHP Warning:  PHP Startup: Unable to load dynamic library '/dev/shm/BUILDROOT/php-pecl-protobuf-3.5.0-1.fc25.remi.7.0.x86_64/usr/lib64/php/modules/protobuf.so' - /dev/shm/BUILDROOT/php-pecl-protobuf-3.5.0-1.fc25.remi.7.0.x86_64/usr/lib64/php/modules/protobuf.so: undefined symbol: timelib_update_ts in Unknown on line 0
`

This is NOT a public api, so cannot be used.

remicollet:
@derick, mostly FYI

BTW, does this make sense to install timelib.h header if functions are not public ?
Perhaps some may be made public, other move to private header ?

remicollet:
@derickr mostly FYI

BTW, does this make sense to install timelib.h header if functions are not public ?
Perhaps some may be made public, other move to private header ?

remicollet:
Notice, they are indeed public from timelib PoV, but private from PHP PoV

remicollet:
3.5.0.1 still broken

TeBoring:
@remicollet thanks for help. What's Pov by the way?

TeBoring:
I have a question. If user has installed the date extension, is timelib_update_ts guaranteed to be available?
If yes, I will just add a check in protobuf (throw error if user want to call Timestamp method without installing data extension). If not, I will implement using other API. @remicollet

TeBoring:
@remicollet Do u mind take a look at #4034?

remicollet:
> I have a question. If user has installed the date extension, is timelib_update_ts guaranteed to be available?

1/ the date extension is always there (cannot be shared or disabled)
2/ timelib_update_ts  is never there (private)

> Do u mind take a look at #4034?

I will try the patch later this week (busy with 7.2.1RC1)

btw, could make sense to add a dependency on date extension (using ZEND_MOD_REQUIRED)

remicollet:
> Do u mind take a look at #4034?

Tested 3.5.0.1 + pr #4034 
Build + load test OK with 5.5, 5.6, 7.0, 7.1 and 7.2

Thanks

remicollet:
[...]

PR #4034, opened 2017-12-12
Avoid calling method from php extension directly

TeBoring:


TeBoring:
#4023, #3929

php/ext/google/protobuf/message.c
```diff
@@ -1129,33 +1128,47 @@ PHP_METHOD(Timestamp, fromDateTime) {
     return;
```
derickr: You probably want to make this work with DateTimeImmutable as well, as that's how people *should* use it in PHP. I would recommend that instead of ``php_date_get_date_ce`` you use zpp as a val* and check with ``instanceof`` function whether datetime matches the ``php_date_get_date_ce()`` or ``php_date_get_immutable_ce`` return values.

Unfortunately, ``date_ce_interface`` is not exported through ``php_date_get_interface_ce()`` as that function doesn't exist.

php/tests/memory_leak_test.php
```diff
@@ -126,6 +126,18 @@
...
 
+$timestamp = new \Google\Protobuf\Timestamp();
+
+date_default_timezone_set('UTC');
+$from = new DateTime('2011-01-01T15:03:01.012345UTC');
+$timestamp->fromDateTime($from);
+assert($from->format('U') == $timestamp->getSeconds());
+assert(0 == $timestamp->getNanos());
+
+$to = $timestamp->toDateTime();
```
derickr: I would recommend you add a ``toDateTimeImmutable`` method as well.

TeBoring: We can do it in another PR if we decide to add new API. Could you help submit an issue for that?

php/ext/google/protobuf/message.c
TeBoring: Good catch. Thx

derickr: You don't need this code. You need to make this extension dependent on the "date" extension in your module's struct, similar to what we do here: https://github.com/derickr/mongo-php-driver/blob/master/php_phongo.c#L2872

The date extension is always available as @remicollet writes in https://github.com/google/protobuf/issues/3929#issuecomment-351288485 (where he also points to the module dependency). And with the date extension, ``php_date_get_date_ce()`` too.

What I was pointing at is that you really want to check for DateTimeInterface, which does *not* currently exports its CE. Which makes it not possible right now.

php/tests/memory_leak_test.php
derickr: Sorry, I don't have the time for that. I've plenty of things to work on for https://github.com/xdebug/xdebug for PHP 7.2 support.

php/ext/google/protobuf/message.c
TeBoring: Sorry for misunderstanding.
In the latest change, I made two thing:
1) Add optional dependency to date, because not every protobuf user needs fromDatetime/toDatetime.
2) Check DatetimeInterface class entry in fromDatetime. Now I use interface name to find class entry, should it work?

php/tests/memory_leak_test.php
TeBoring: Added #4050 to track it. We will provide support for DateTimeImmutable in next release.
How do u think about adding a boolean arg (is_immutable) to toDateTime instead of providing toDateTimeImmutable? As I see, fromDatetime already takes both DateTime and DateTimeImmutable, it's better to let fromDateTime provide both type of instance.

TeBoring:
I need to merge this now for release before Christmas

TeBoring:
Please leave comment if there is more change needed.

Issue #2054, opened 2016-09-01
[proto3] Deprecated method `valueOf()` called in generated code


biran0079:
Sample proto:

```
option java_multiple_files = true;
option java_package = "io.grpc.examples.helloworld";

enum TestEnum {
  VALUE = 0;
}

message TestMessage {
  TestEnum test_enum = 1;
}
```

Following code is generated:

``` java
  /**
   * <code>optional .TestEnum test_enum = 1;</code>
   */
  public io.grpc.examples.helloworld.TestEnum getTestEnum() {
    io.grpc.examples.helloworld.TestEnum result = io.grpc.examples.helloworld.TestEnum.valueOf(testEnum_);
    return result == null ? io.grpc.examples.helloworld.TestEnum.UNRECOGNIZED : result;
  }
```

which gives following warnings:

```
/io/grpc/examples/helloworld/TestMessage.java:87: mandatory_warning: valueOf(int) in io.grpc.examples.helloworld.TestEnum has been deprecated
    io.grpc.examples.helloworld.TestEnum result = io.grpc.examples.helloworld.TestEnum.valueOf(testEnum_);
                                                                                      ^
/io/grpc/examples/helloworld/TestMessage.java:378: mandatory_warning: valueOf(int) in io.grpc.examples.helloworld.TestEnum has been deprecated
      io.grpc.examples.helloworld.TestEnum result = io.grpc.examples.helloworld.TestEnum.valueOf(testEnum_);
                                                                                        ^
```

biran0079:
Would appreciate if this issue is fixed soon.
Our application is built with `-Werror` flag and this is blocking us from upgrading to Grpc 1.0.

jorgjanke:
+10 - same as warnings for us also prevent build

acozzette:
It looks like we need to have the generated code continue to call `valueOf()`, because we need to remain compatible with older generated code that does not yet have the recommended `forNumber()` method. So it seems to me like the right solution is to update the compiler to stop putting a deprecation annotation on `valueOf`. That should happen in these two places:
 - https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/java/java_enum.cc#L186
 - https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/java/java_enum_lite.cc#L150

Would anyone be interested in sending out a pull request to make this change?

jorgjanke:
... I used c last time a decade ago .. so my temp solution was

```
[...]

PR #4046, opened 2017-12-14
Java: generate an annotation to suppress deprecation warnings

acozzette:
The valueOf method on enums is deprecated, but for compatibility reasons
the generated code has to keep using it. We can use an annotation to
suppress the deprecation warnings.


acozzette:
This should hopefully fix #2054.

ravenspree:
Can someone tell me what's the fix here? Coz I still see this in proto3

PR #4047, opened 2017-12-14
Allow roundtrip of FLT_MIN/MAX in JSON convention.

xfxyjwf:
Fixes https://github.com/google/protobuf/issues/3615

### DIFF
--- a/src/google/protobuf/util/internal/datapiece.cc
+++ b/src/google/protobuf/util/internal/datapiece.cc
@@ -106,8 +106,22 @@ StatusOr<float> DoubleToFloat(double before) {
   } else if (!MathLimits<double>::IsFinite(before)) {
     // Converting a double +inf/-inf to float should just work.
     return static_cast<float>(before);
-  } else if (before > std::numeric_limits<float>::max() ||
-             before < -std::numeric_limits<float>::max()) {
+  } else if (before > std::numeric_limits<float>::max()) {
+    // We accept values that are slightly larger than FLT_MAX or slightly
+    // smaller than FLT_MIN because converting FLT_MAX/FLT_MIN to text may
+    // result in these values due to round-up/down.
+    if (before <= 3.4028235e+38) {
+      return std::numeric_limits<float>::max();
+    }
+    // Double value outside of the range of float.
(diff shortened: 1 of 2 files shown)


src/google/protobuf/util/internal/datapiece.cc
```diff
@@ -106,8 +106,22 @@ StatusOr<float> DoubleToFloat(double before) {
   } else if (!MathLimits<double>::IsFinite(before)) {
     // Converting a double +inf/-inf to float should just work.
     return static_cast<float>(before);
-  } else if (before > std::numeric_limits<float>::max() ||
-             before < -std::numeric_limits<float>::max()) {
+  } else if (before > std::numeric_limits<float>::max()) {
+    // We accept values that are slightly larger than FLT_MAX or slightly
+    // smaller than FLT_MIN because converting FLT_MAX/FLT_MIN to text may
+    // result in these values due to round-up/down.
+    if (before <= 3.4028235e+38) {
```
pherl: is there constant we can use instead of hard coding the value?

xfxyjwf: No idea. There are FLT_MIN/FLT_MAX, but there are different from the value here...

pherl: Is there any float precision documentation in our JSON spec? Is so, a cleaner solution could be print the FLT_MIN/MAX into the string representation and parse it back, then compare those rounded values. If we don't have any precision spec, we should probably change FLT_DIG to be roundtrip-safe in text representation.

xfxyjwf: There isn't detailed spec about float precision. I'm not sure what you mean by "print to string, part it back, then compare those rounded values". Are you suggesting we use a value calculated at runtime instead of a hard-coded 3.4028235e+38? What if the value ends up smaller than FLT_MAX (e.g. 3.402823e+38)? The printing code comes from system library so it could differ from platform to platform. Especially other languages are likely to choose a different precision than C++. 

Updating the printing code also wouldn't work because the parsing code will receive JSON input from other sources. Besides, we will replace the string formatting code with absl in the future. Any fix in that formatting code won't stay for long.

PR #4052, opened 2017-12-14
Ruby: make .to_json method play nice with builtin json module

epidemian:
Generated Ruby message objects implement a convenience `to_json` method that return a JSON-formatted string for that message.

However, that `to_json` method doesn't play nice with native Ruby's [`json` module](http://ruby-doc.org/stdlib-2.4.0/libdoc/json/rdoc/JSON.html), specially in the case of nesting a protobuf message into a data structure like an Array or a Hash. Take this example program:

```ruby
require 'json'
require 'google/protobuf'

  add_message 'MyMessage' do
    optional :str, :string, 1
  end
end

MyMessage = Google::Protobuf::DescriptorPool.generated_pool.lookup('MyMessage').msgclass

message = MyMessage.new(str: 'foobar')

puts(message.to_json) # => {"str":"foobar"}
puts([message].to_json) # ArgumentError: wrong number of arguments (given 1, expected 0)
puts({message: message}.to_json) # ArgumentError: wrong number of arguments (given 1, expected 0)
```

This PR simply adds an optional argument to the `to_json` method of these generated classes. This argument is completely ignored, but makes it possible to nest protobuf objects into Ruby data structures and then serialize everything with a `to_json` call.

With this patch, the above code runs and yields:

```
{"str":"foobar"}
[{"str":"foobar"}]
{"message":{"str":"foobar"}}
```

😸

### DIFF
--- a/ruby/lib/google/protobuf/message_exts.rb
+++ b/ruby/lib/google/protobuf/message_exts.rb
@@ -40,7 +40,7 @@ module Google
       module ClassMethods
       end
 
-      def to_json
+      def to_json(_opts=nil)
         self.class.encode_json(self)
       end
 


epidemian:
I signed it!

epidemian:
Hi! This change is pretty simple, and i wanted to have the PR opened quickly to have some early feedback about whether it would be accepted or not 🙂 

But let me know if i need to add something else, like some unit test(s) for this extension `to_json` method. Id that'd be the case, is [this](https://github.com/google/protobuf/blob/master/ruby/compatibility_tests/v3.0.0/tests/basic.rb) where such tests should go?

acozzette:
ok to test (this comment will trigger the other CI runs)

PR #4068, opened 2017-12-19
resolved issue 4032 and added a unit test

wsw2016:


wsw2016:
I signed it!

acozzette:
ok to test (this comment will start the other CI runs)

src/google/protobuf/util/json_util.cc
```diff
@@ -50,7 +50,8 @@ namespace util {
 
 namespace internal {
 ZeroCopyStreamByteSink::~ZeroCopyStreamByteSink() {
-  stream_->BackUp(buffer_size_);
+  if (buffer_size_ > 0)
```
acozzette: Could you put in curly braces around the body of the if statement, i.e.

if (buffer_size_ > 0 ) {
  ...
}

We usually go with that convention just to be safe.

src/google/protobuf/util/json_util_test.cc
```diff
@@ -457,6 +458,20 @@ TEST(ZeroCopyStreamByteSinkTest, TestAllInputOutputPatterns) {
   }
 }
 
+TEST_F(JsonUtilTest, TestWronJsonInput) {
+  const char json[] = "{\"unknown_field\":\"some_value\"}";
```
acozzette: Couple small things:
- Add line breaks to keep the lines under 80 characters long
- Our convention is to use `snake_case` for variables in C++ so `protoBuffer` should be `proto_buffer` and likewise for `messageType`, and `resultStatus`.

PR #4070, opened 2017-12-19
Update version number to 3.5.1

pherl:
For libtools version-info, bumping the revision number only.


xfxyjwf:
So the .so interface hasn't changed?

xfxyjwf:
And we probably should fix https://github.com/google/protobuf/issues/4064 in 3.5.1 as well.

pherl:
Will do.

PR #4081, opened 2017-12-21
Add DiscardUnknownFields support for csharp

anandolee:


jskeet:
I'm on vacation until January - can this wait until then?

anandolee:
Sure it's holiday season. Take your time

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -1274,5 +1274,12 @@ private void SkipImpl(int amountToSkip)
             }
         }
         #endregion
+
+        private bool discardUnknownFields = false;
```
jskeet: Change this to just

```csharp
internal bool DiscardUnknownFields { get; set; }
```

No need for an explicit field declaration etc.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -185,6 +185,10 @@ internal UnknownFieldSet AddOrReplaceField(int number, UnknownField field)
         /// <returns>false if the tag is an "end group" tag, true otherwise</returns>
         private void MergeFieldFrom(CodedInputStream input)
         {
+            if (input.DiscardUnknownFields) {
```
jskeet: Nit: move the `{` to the start of the next line.

csharp/src/Google.Protobuf/DiscardUnknownFieldsMessageParser.cs
```diff
@@ -0,0 +1,60 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Inherits a MessageParser to a new Parser that discards unknown fields during parsing.
+    /// </summary>
+    public class DiscardUnknownFieldsMessageParser<T>: MessageParser<T> where T: IMessage<T>
```
jskeet: Rather than creating a new class, you could just include an `DiscardUnknownFields` flag in `MessageParser`. We can then add instance methods of `WithDiscardUnknownFields(bool discard)` which return a new `MessageParser` or `MessageParser<T>`. I'd have to check how much of the current change then changes again.

That would allow us to keep `MessageParser<T>` sealed, which is handy from an API stability perspective.

anandolee:
Go ahead to create a new PR. 

Yes, discarding at parse time is OK. That's what we do in Java: https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/DiscardUnknownFieldsParser.java

jskeet:
Just worked out there's rather more to it than in this PR, too - a lot of MessageParser methods use the extension methods in MessageExtensions - we'll have to stop doing that. Will address that in my PR.

Issue #4069, opened 2017-12-19
Build error using Intel compiler in generated_message_table_driven.h


neundorf:
Hi, 
with the Intel C++ compiler (both versions 15.0 and 16.0.4) there is a compiler error in 
generated_message_table_driven.h:
generated_message_table_driven.h(186): error: static assertion failed with ""
      static_assert(std::is_pod<AuxillaryParseTableField>::value, "");

Commenting out that line makes it build, then the code looks like this:

```
if LANG_CXX11 && !defined(__NVCC__)
static_assert(sizeof(ParseTableField) <= 16, "ParseTableField is too large");
// The tables must be composed of POD components to ensure link-time
// initialization.
static_assert(std::is_pod<ParseTableField>::value, "");
//static_assert(std::is_pod<AuxillaryParseTableField>::value, "");     <---- this line is the problem
static_assert(std::is_pod<AuxillaryParseTableField::enum_aux>::value, "");
static_assert(std::is_pod<AuxillaryParseTableField::message_aux>::value, "");
static_assert(std::is_pod<AuxillaryParseTableField::string_aux>::value, "");
static_assert(std::is_pod<ParseTable>::value, "");
#endif

```

What's going wrong there ?

neundorf:
The static_assert() does not fail if I comment out the additional ctors of AuxillaryParseTableField:

```
union AuxillaryParseTableField {

...

#if LANG_CXX11
  AuxillaryParseTableField() = default;
#else
  AuxillaryParseTableField() { }
#endif
/*   Commented out:
  PROTOBUF_CONSTEXPR AuxillaryParseTableField(
      AuxillaryParseTableField::enum_aux e) : enums(e) {}
  PROTOBUF_CONSTEXPR AuxillaryParseTableField(
      AuxillaryParseTableField::message_aux m) : messages(m) {}
  PROTOBUF_CONSTEXPR AuxillaryParseTableField(
      AuxillaryParseTableField::string_aux s) : strings(s) {}
  PROTOBUF_CONSTEXPR AuxillaryParseTableField(
      AuxillaryParseTableField::map_aux m)
      : maps(m) {}*/
};
```

acozzette:
@neundorf It sounds like this is probably a bug with the Intel compiler, right? Do you want to send us a pull request that ifdefs out that static_assert when we are building with that compiler?

neundorf:
[...]

PR #4084, opened 2017-12-21
Fix build for Intel compilers.

neundorf:
The full discussion is in ticket #4069 .
The Intel compilers version 15 and 16 fail at this static_assert.
The warning 2196 "routine is both inline and noninline" makes it impossible to build make check, so disable this warning.


jgsogo:
Hi @neundorf! 

I really need this patch for a project (and a protobuf release), would you mind signing the CLA and going ahead with this?

Thanks!

PR #4087, opened 2017-12-22
Create CMakeLists.txt

maPaydar:
To add protobuf project from root using External_Project_Add in cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,4 @@
+# Minimum CMake required
+cmake_minimum_required(VERSION 2.8.12)
+
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/cmake)


maPaydar:
I signed it!

jtattermusch:
CC @jtattermusch

Mizux:
Not needed for using Add_ExternalProject, just add a `SOURCE_SUBDIR cmake`


> `SOURCE_SUBDIR <dir>`
When no `CONFIGURE_COMMAND` option is specified, the configure step assumes the external project has a CMakeLists.txt file at the top of its source tree (i.e. in `SOURCE_DIR`). The `SOURCE_SUBDIR` option can be used to point to an alternative directory within the source tree to use as the top of the CMake source tree instead. This must be a relative path and it will be interpreted as being relative to `SOURCE_DIR`.

source: [here](https://cmake.org/cmake/help/latest/module/ExternalProject.html)

Issue #4091, opened 2017-12-23
Improve generated code's API (C++)


os12:
### Problem statement

The C++ code generated by `protoc` has a peculiar asymmetry between the reading and writing cases.

**A reader:**
```
message.repeated_field(1).map_field()[0].value();
```

**A writer:**
```
(*message.mutable_repeated_field(1)->mutable_map_field())[0]->set_value(...);
```

As seen from the examples above there are two inconsistencies:

1.  The presence of the `mutable_` prefix for non-const accessors (while const ones have natural names)
    - This becomes rather annoying as the code for a mutation is a lot longer than that of a read
    - Most code bases have a hard limit on the maximum line length and so
      * the code author is forced to create local variables
      * the code reviewer is forced to deal with extra local variables in the scope
1. Mutable accessors return a pointer (while const ones return a `const &`)
    - This is annoying when modifying code
    - Natural assignment does not work (as one must dereference)
    - Map syntax is clumsy (as one must dereference before using `operator[]`)

### Proposal

In most cases there is no technical reason to give mutable accessors a distinct name as the C++ compiler selects the appropriate overload based on the "const" property of the given value. Also the mutable accessors can return a mutable reference as there is no technical reason to impose the dereference onto the caller.

**Proposed writer syntax:**
```
message.repeated_field(1).map_field()[0].set_value(...);
```

Note:
  - Such a thing will not compile when "message" is const, just like the existing `mutable_` syntax.
  - I realize that Google's C++ style guide bans mutable reference arguments, but let's leave that point aside.

### Discussion
I realize that a significant change in the generated code's syntax has breaking implications on the vast amount of existing code and would only benefit new projects/programmers adopting Protobuf. So, let's figure out a gentle way of landing such a feature:
  - Should it be a user-selectable option?
  - Should the new mutable accessors extend the existing API rather than modify it?
  - Should the new syntax imply a major version change (i.e. protobuf v4)?

### Proof of concept

I've implemented the proposal and updated the tests. I'll post a PR so that we can see the syntax improvement. My plan is to do the following:
  1. Eliminate the `mutable_` prefix wherever possible
  1. Move the mutable accessors to mutable C++ references

xfxyjwf:
Thanks for experimenting with the idea, but such API breaking change will likely never happen in protobuf. The cost of updating all existing users is just too high.

PR #4092, opened 2017-12-23
POC: improved the generated C++ code

os12:
This is a POC for #4091:
  - mutable accessors return a mutable C++ reference
  - repeated fields generate a mutable accessor with a natural name

The point here is unify reader and writer scenarios:

**A reader:**
```
message.repeated_field(1).map_field()[0].value();
```

**A proposed writer:**
```
message.repeated_field(1).map_field()[0].set_value(...);
```

All tests compile and pass on Linux x86-64 using GCC 4.8 (Ubuntu 14). I have not compiled in other environments...

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,5 @@
+---
+BasedOnStyle: Google
+Standard: Cpp11
+---
+Language: Cpp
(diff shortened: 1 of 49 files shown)


xfxyjwf:
Thanks for experimenting with the idea, but such API breaking change will likely never happen in protobuf.

Issue #3510, opened 2017-08-17
idx is an undefined name 3 times in pddm_tests.py


cclauss:
flake8 testing of https://github.com/google/protobuf on Python 2.7.13

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./objectivec/DevTools/pddm_tests.py:314:54: F821 undefined name 'idx'
      self.fail('Should throw exception, entry %d' % idx)
                                                     ^

./objectivec/DevTools/pddm_tests.py:486:54: F821 undefined name 'idx'
      self.fail('Should throw exception, entry %d' % idx)
                                                     ^

./objectivec/DevTools/pddm_tests.py:506:54: F821 undefined name 'idx'
      self.fail('Should throw exception, entry %d' % idx)
                                                     ^
```

PR #4108, opened 2017-12-27
Resolved issue #3510.

v-silverfin:
Strings in which undefined 'idx' was found were copy-pasted within the same file to multiple places but were not properly altered afterwards. It never showed as a problem because their use is to fail broken tests which never happened.


thomasvl:
Thanks!

PR #4112, opened 2017-12-29
Support Conversion between JSON and message in Objective C

larrytin:
Fixes google/protobuf#1982
Fixes google/protobuf#1675
Fixes google/protobuf#3935

### DIFF
--- a/objectivec/GPBMessage+JsonFormat.h
+++ b/objectivec/GPBMessage+JsonFormat.h
@@ -1,0 +1,39 @@
+//
+// Created by Larry Tin.
+//
+
+#import "GPBMessage.h"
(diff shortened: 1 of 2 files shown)


thomasvl:
The `Descriptor` object in the objc runtime do not carry the correct naming for JSON support.  In some cases it may might happen to be the same (or close) and appear to work, but since it isn't ensure to be accurate it can't be trusted to always work.  Ensuring the right data is is there is where the bloat could come from which is why ObjC does not support JSON.  Since most uses of the ObjC support are for mobile, the binary format is a better transmission format, and keeping the size of the binaries down is important to those clients.

PR #4114, opened 2017-12-29
check null for reference type object before writeto can avoid many nullpointer exception

code4ever:
…llpointer exception.

Signed-off-by: code4ever <jessee_hustcs@126.com>


xfxyjwf:
Closing due to conflicts.

PR #4117, opened 2017-12-29
Add __init__.py files to compiler and util subpackages

### LINKED ISSUE #2895: Compiler module missing from Python 2.7 wheels
The `google.protobuf.compiler` module is missing in the Python 2.7 wheels for releases 3.2.0 and 3.0.0 on PyPI:
https://pypi.python.org/pypi/protobuf/3.2.0
https://pypi.python.org/pypi/protobuf/3.0.0

This only seems to affect the wheels for Python 2.7, which are only available on PyPI for those two releases. The module is still present in the source distribution and Python 3.x wheels.

To reproduce missing module error:

`$ pip install protobuf==3.2.0`
`$ python`
Python 2.7.6 (default, Oct 26 2016, 20:30:19) 
[GCC 4.8.4] on linux2
`>>> import google.protobuf`
`>>> import google.protobuf.compiler`
  File "<stdin>", line 1, in <module>
ImportError: No module named compiler

thisch:
The compiler and util subpackages are created by the build_py class in
setup.py. This has caused an issue in the protobuf package in
conda-forge (https://github.com/conda-forge/protobuf-feedstock/issues/40). This
commit fixes this issue.

Fixes #2895

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,8 +58,6 @@ python/*.egg
 python/.eggs/
 python/.tox
 python/build/
-python/google/protobuf/compiler/
-python/google/protobuf/util/
 
 src/js_embed
 src/protoc
--- a/python/setup.py
+++ b/python/setup.py
@@ -119,9 +119,7 @@ class clean(_clean):
       for filename in filenames:
         filepath = os.path.join(dirpath, filename)
         if filepath.endswith("_pb2.py") or filepath.endswith(".pyc") or \
-          filepath.endswith(".so") or filepath.endswith(".o") or \
-          filepath.endswith('google/protobuf/compiler/__init__.py') or \
-          filepath.endswith('google/protobuf/util/__init__.py'):
+          filepath.endswith(".so") or filepath.endswith(".o"):
           os.remove(filepath)
     # _clean is an old-style class, so super() doesn't work.
     _clean.run(self)
@@ -143,12 +141,6 @@ class build_py(_build_py):
     generate_proto("../src/google/protobuf/wrappers.proto")
     GenerateUnittestProtos()
 
-    # Make sure google.protobuf/** are valid packages.
-    for path in ['', 'internal/', 'compiler/', 'pyext/', 'util/']:
-      try:
-        open('google/protobuf/%s__init__.py' % path, 'a').close()
-      except EnvironmentError:
-        pass
     # _build_py is an old-style class, so super() doesn't work.
     _build_py.run(self)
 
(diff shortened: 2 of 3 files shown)


thisch:
I signed it!

.gitignore
```diff
@@ -58,8 +58,6 @@ python/*.egg
 python/.eggs/
 python/.tox
 python/build/
-python/google/protobuf/compiler/
-python/google/protobuf/util/
```
jakirkham: Maybe we should keep these, but add things like `!python/google/protobuf/compiler/__init__.py`.

thisch: Note that the `compiler` and `util` directories only contain generated `*_pb2.py` files besides the `__init__.py` files. The `*_pb2.py` files are already ignored in L56. 

Why do you want to keep the two lines?

hajapy:
Build failure seems to be due to missing listing the `__init__.py` files here: https://github.com/google/protobuf/blob/master/Makefile.am#L773

thisch:
@hajapy thx for the hint (I tried to find the reason why this CI job was failing but wasn't aware of the file list in Makefile.am). This should be fixed now.

.gitignore
jakirkham: As I'm guessing there is a reason for them. Even though I'm not entirely sure what it is.

thisch: Me neither. git blame doesn't help much in this case ...

jakirkham:
@xfxyjwf, do you know of anyone who has time to review this?

thisch:
friendly ping

xfxyjwf:
@anandolee to take a look

thisch:
I've just found this [discussion](https://groups.google.com/forum/#!topic/protobuf/5Ywz8gwn9Kk) about the same issue this PR is fixing.

thisch:
Fixed #2895

balopat:
+1 this is a problem for us as well

thisch:
@anandolee gentle ping :)

python/setup.py
```diff
@@ -143,12 +141,6 @@ def run(self):
     generate_proto("../src/google/protobuf/wrappers.proto")
     GenerateUnittestProtos()
 
-    # Make sure google.protobuf/** are valid packages.
-    for path in ['', 'internal/', 'compiler/', 'pyext/', 'util/']:
```
anandolee: Why this check need to be removed?

.gitignore
anandolee: They are in here for generated directories. It is reasonable to remove them as now __init__ files are added

xfxyjwf:
@anandolee Jie, can you merge the pull request if it looks good to you?

jakirkham:
Thanks all. :)

Issue #4094, opened 2017-12-23
nullptr used in 3.5.1


jakirkham:
Appears that [`nullptr`]( http://en.cppreference.com/w/cpp/language/nullptr ) is used in 3.5.1 as evidenced by this [build failure]( https://ci.appveyor.com/project/conda-forge/protobuf-feedstock/build/1.0.115/job/c54b86hvqpv8pvix#L756 ). `nullptr` requires C++11 at least.

jakirkham:
For reference, `nullptr` shows up on [this line]( https://github.com/google/protobuf/blob/v3.5.1/src/google/protobuf/stubs/io_win32.cc#L94 ). This was introduced by PR ( https://github.com/google/protobuf/pull/4013 ). So cc-ing @laszlocsomor.

laszlocsomor:
Thanks for letting me know! That PR (https://github.com/google/protobuf/pull/4013) isn't yet merged to `master`, so the line doesn't exist on `master` yet.
@pherl , shall I send PR to merge it to `master`?

pherl:
Master is now c++11 only. We only need to fix it in the 3.5.x branch.

laszlocsomor:
@pherl : Thanks! So shall I create a PR to merge that commit to master, or will you (or someone else) merge all 3.5.x changes back to master?

pherl:
I will merge it back.

PR #4124, opened 2018-01-02
remove nullptr

pherl:


pherl:
Fixes #4094

jakirkham:
Thanks for doing this @pherl. Worked great.

Sorry if I already asked you this earlier, but are any of the CIs for this branch testing on a pre-C++11 compiler (e.g. CentOS 6 with gcc 4.4.7 or similar)?

PR #4127, opened 2018-01-03
Add DiscardUnknownFields support for C#

jskeet:
By default, unknown fields are preserved when parsing. To discard
them, use a parser configured to do so:

```csharp
var parser = MyMessage.Parser.WithDiscardUnknownFields(true);
// Unknown fields will be discarded
var message = parser.Parse(data);
```

This PR is an alternative to #4081.

### DIFF
--- a/csharp/src/Google.Protobuf/MessageParser.cs
+++ b/csharp/src/Google.Protobuf/MessageParser.cs
@@ -42,10 +42,13 @@ namespace Google.Protobuf
     public class MessageParser
     {
         private Func<IMessage> factory;
+        // TODO: When we use a C# 7.1 compiler, make this private protected.
+        internal bool DiscardUnknownFields { get; }
 
-        internal MessageParser(Func<IMessage> factory)
+        internal MessageParser(Func<IMessage> factory, bool discardUnknownFields)
         {
             this.factory = factory;
+            DiscardUnknownFields = discardUnknownFields;
         }
 
         /// <summary>
@@ -65,7 +68,7 @@ namespace Google.Protobuf
         public IMessage ParseFrom(byte[] data)
         {
             IMessage message = factory();
-            message.MergeFrom(data);
+            message.MergeFrom(data, DiscardUnknownFields);
             return message;
         }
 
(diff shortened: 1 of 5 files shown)


csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -204,7 +234,7 @@ internal new T CreateTemplate()
         public new T ParseFrom(byte[] data)
         {
             T message = factory();
-            message.MergeFrom(data);
+            message.MergeFrom(data, DiscardUnknownFields);
```
anandolee: Do we have test coverage for these template?

jskeet: Not sure what you mean by "for these template" but the new test calls all the ParseFrom overloads, and the existing extension methods are tested by the existing tests.

anandolee: Sorry, I commented on wrong place. See comments above

csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -65,7 +68,7 @@ internal IMessage CreateTemplate()
         public IMessage ParseFrom(byte[] data)
         {
             IMessage message = factory();
-            message.MergeFrom(data);
+            message.MergeFrom(data, DiscardUnknownFields);
```
anandolee: Actually I want to ask if these changes are tested. The comments say it is "A general message parser, typically used by reflection-based code as all the methods" but I can't find tests for the reflection based parse. I guess the tests you added can also be parse without change this general message parser?

jskeet: The tests I've provided call all the parse methods, because they use parsers declared as `MessageParser<T>` and `MessageParser`.

anandolee:
LGTM 
ok to test

aaronhudon-ts:
Is this behavior intended to work with the `ParseJson` API?  It is throwing `InvalidProtocolBufferException: 'Unknown field: xxxxxxx' when the JSON content has additional (unknown fields)

jskeet:
@aaronhudon-ts: As you've noted elsewhere, it's entirely separate from ignoring unknown fields when parsing JSON. It looks like you've now worked out how to do that. It's unfortunate that these two features have very different histories... we might have planned them to work together if they'd been developed at the same time.

PR #4132, opened 2018-01-03
Java caliper counter

BSBandme:
Make Java benchmark run for enough times so that can make the uncertain error from caliper measurements less than 1%


benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -43,18 +43,22 @@
...
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.RandomAccessFile;
+import java.nio.file.*;
```
xfxyjwf: Expend this. Don't use wildcards in imports.
https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -43,18 +43,22 @@
...
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Scanner;
 
 
 public class ProtoBench {
 
+
```
xfxyjwf: Remove this empty line

benchmarks/java/src/main/java/com/google/protobuf/ProtoBenchCaliper.java
```diff
@@ -95,7 +98,7 @@ Message getDefaultInstance() {
   private List<ByteArrayInputStream> inputStreamList;
   private List<ByteString> inputStringList;
   private List<Message> sampleMessageList;
-  private int counter;
+  private long counter, mCounter;
```
xfxyjwf: What is mCounter? Can you give them better names?

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -66,17 +70,26 @@ public static void main(String[] args) {
       System.exit(1);
     }
     boolean success = true;
+    int endIndex = args.length;
     for (int i = 0; i < args.length; i++) {
-      success &= runTest(args[i]);
+      if (args[i].equals("--")) {
```
xfxyjwf: What is this used for?

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -66,17 +70,26 @@ public static void main(String[] args) {
...
     System.exit(success ? 0 : 1);
   }
+  
  
   /**
    * Runs a single test with specific test data. Error messages are displayed to stderr, 
    * and the return value indicates general success/failure.
    */
-  public static boolean runTest(String file) {
+  public static boolean runTest(String[] args, int currentIndex, int endIndex) {
```
xfxyjwf: What's the meaning of these parameters?

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -94,19 +107,69 @@ public static boolean runTest(String file) {
...
     }
+    try {
+      double maxTimeIntervalScale = 0;
+      Scanner scanner = new Scanner(new String(readAllBytes("JavaBenchmarkWarning.txt")));
+      while (scanner.hasNext()) {
+        maxTimeIntervalScale = Math.max(maxTimeIntervalScale, scanner.nextDouble());
+      }
+      scanner.close();
+
+      String timingIntervalWithUnit = getFromArgs(caliperArgs,
```
xfxyjwf: Can you explain what the following code is doing?

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -94,19 +107,69 @@ public static boolean runTest(String file) {
...
+      } 
+      timingIntervalValue *= maxTimeIntervalScale * 1.2;
+      String measurementsString = getFromArgs(caliperArgs, 
+          "instrument.runtime.options.measurements");
+      if (measurementsString.equals("")) {
+        measurementsString = "9";
+      }
+      int measurements = (int) (Integer.parseInt(measurementsString) * 1.2 * maxTimeIntervalScale);
+      
+      System.out.println(
```
xfxyjwf: It seems this warning is printed unconditionally?

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -132,9 +199,9 @@ public static boolean runTest(String file) {
     temp.add("runtime");
     temp.add("-b");
     String benchmarkNames = "serializeToByteString,serializeToByteArray,serializeToMemoryStream"
-       + ",deserializeFromByteString,deserializeFromByteArray,deserializeFromMemoryStream";
+        + ",deserializeFromByteString,deserializeFromByteArray,deserializeFromMemoryStream";
     temp.add(benchmarkNames);
-    temp.add("-Cinstrument.runtime.options.timingInterval=3000ms");
+    temp.add("--print-config");
```
xfxyjwf: Remove this?

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: This is for user to add caliper options. All the strings after "--" will be considered to be caliper options.

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: The parameter is for split the caliper options and the input files. 
And renamed the parameters,

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: getFromArgs is to get the value string from the args array of a specific argument.

Add some comments for getFromArgs

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: If there's no file "JavaBenchmarkWarning.txt" then there will be IOException thrown inside try catch, and won't print the warning.

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoBenchCaliper.java
BSBandme: This is used for get rid of the warmup runs which shouldn't be counted.

There're only 3 kinds of runs for caliper benchmark. 
One is dry run, usually just run for once and not in subprocess. Then follows warmup run,  which is constrained by running time, which is run in subprocess. Then are measurements, which follows up with warmup, and are constrained by number of measurements and one mearsurements running time.

Rewriten the counter to use an array to restore all the last (number of measurements + 1) counters.

BSBandme:
Done changing back the counter's logic

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -71,10 +75,16 @@ public static void main(String[] args) {
     }
     System.exit(success ? 0 : 1);
   }
+  
  
   /**
    * Runs a single test with specific test data. Error messages are displayed to stderr, 
    * and the return value indicates general success/failure.
+   * 
+   * @param args The arguments from the main function
```
xfxyjwf: Revert this comment change.

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -97,16 +106,39 @@ public static boolean runTest(String file) {
     
     try {
       String args[] = new String[argsList.size()];
-      argsList.toArray(args);
-      CaliperMain.exitlessMain(args, 
+      argsList.toArray(args);
```
xfxyjwf: Remove trailing spaces.

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -97,16 +106,39 @@ public static boolean runTest(String file) {
...
+        + "run for more time to make the result more convincing. You may change the configure "
+        + "code in com.google.protobuf.ProtoBench.getCaliperOption() of benchmark " 
+        + benchmarkDataset.getMessageName()
+        + " to run for more time. e.g. Change the value of instrument.runtime.options.timingInterva"
+        + "l or value of instrument.runtime.options.measurements to be at least "
+        + Math.round(mininumScale * 10 + 1) / 10.0 
+        + " times of before, then build and run the benchmark again\n");
+      Files.deleteIfExists(Paths.get("JavaBenchmarkWarning.txt"));
+    } catch (IOException e) {
+
```
xfxyjwf: Add a comment here to explain why the exception is ignored.

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
```diff
@@ -97,16 +106,39 @@ public static boolean runTest(String file) {
...
+      }
+      scanner.close();
+      
+      System.out.println(
+         "WARNING: This benchmark's whole iterations are not enough, consider to config caliper to "
+        + "run for more time to make the result more convincing. You may change the configure "
+        + "code in com.google.protobuf.ProtoBench.getCaliperOption() of benchmark " 
+        + benchmarkDataset.getMessageName()
+        + " to run for more time. e.g. Change the value of instrument.runtime.options.timingInterva"
+        + "l or value of instrument.runtime.options.measurements to be at least "
```
xfxyjwf: Don't break a word into two lines.

BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoBench.java
BSBandme: Done

PR #4133, opened 2018-01-03
Force a copy when saving the NSData that came from another.

thomasvl:


objectivec/GPBMessage.m
```diff
@@ -2027,7 +2027,12 @@ - (void)parseMessageSet:(GPBCodedInputStream *)input
       [newInput release];
     } else {
       GPBUnknownFieldSet *unknownFields = GetOrMakeUnknownFields(self);
-      [unknownFields mergeMessageSetMessage:typeId data:rawBytes];
+      // rawBytes was created via a NoCopy, so it can be resusing a
```
dmaclach: reusing

dmaclach: Do we have a unit test for this?

thomasvl: I didn't see a clean way to test this one without trying to mock out some dummy NSData and craft a flow to hit this case and validate what was called.

Issue #3220, opened 2017-06-13
upgrade plugin.proto and protobuf/descriptor.proto to version 3


lvht:
PHP code generation has been introduced in V3.0.0. And it seems that protobuf only support generating php code for generating php code.

If we transform this proto to version 3, we could generate php code, and make it is possible to write plugin by pure php code.

thomasvl:
Why?

xfxyjwf:
This is not going to happen...

PR #4135, opened 2018-01-04
PHP: Add support for generating php code for plugin.proto

lvht:
Generate php file for the **plugin.proto**, so that we can use php to develop protoc-gen-plugin.

Also fix the issue #3220

And here is a demo,

```php
<?php
require './vendor/autoload.php';

$input = file_get_contents("php://stdin");
$request = new Google\Protobuf\Internal\CodeGeneratorRequest;
$response = new Google\Protobuf\Internal\CodeGeneratorResponse;
$request->mergeFromString($input);

/** @var \Google\Protobuf\Internal\FileDescriptorProto $fs */
$fs = $request->getProtoFile()[0];
$package = $fs->getPackage();

$f = new \Google\Protobuf\Internal\CodeGeneratorResponse_File();
$f->setName("foo.php");
$f->setContent("<?php // $package\n");

$response->setFile([$f]);

file_put_contents("php://stdout", $response->serializeToString());
```


lvht:
I signed it!

lvht:
Would you @TeBoring like to review this PR? Thank you.

lvht:
Is any body here?

TeBoring:
Sorry for late reply. What is the purpose of this PR?

TeBoring:
AFAIK, all protoc plugin are written in cpp. So u don't need php plugin.proto

lvht:
Thank you for reply @TeBoring 

> all protoc plugin are written in cpp. So u don't need php plugin.proto

Although the cpp is the first class language for protoc, it should not be the only language. 

Writing a plugin has no difference on reading binary data from `stdin`, decode by the `CodeGeneratorRequest`, fill some data into the `CodeGeneratorResponse`, encode and write the response to the `stdout`.

The architecture protoc plugin is awesome. You can use any language you favorite. And there is no need to make such a limit that all plugin should **only** be written in the **cpp**. 

Actually, as I cannot write plugin using PHP, I have used the **golang** to write the [protoc-gen-grpc-php](https://github.com/lvht/protoc-gen-grpc-php) for custom php grpc client and server code generation.

The protobuf is great for language independent data serialize and deserialize. So it is natural to use it in PHP. And it is natural to use the `.proto` file to describe our own RPC definition. However, both the message php code or the service interface php code make not meet our internal requirements, we need to write a plugin to generate code in our favorite way.

As a protoc developer, Using the **cpp** is normal for you. However, as a user, using **cpp** to write plugin will never be a normal or required skill for PHP developer. It will be easier and more natural to use PHP to write plugin, so that all the PHP developer can understand and modify the code.

As far as I know, we can use both python and golang to write plugin. So why cannot we use php? 

The current **protoc-php-gen** can only generate PHP code for the `google/protobuf/descriptor.proto`, and threat it in a internal way. If we made **protoc-php-gen** support generating code for `google/protobuf/compiler/plugin.proto`, we will get the php code of `CodeGeneratorRequest` and `CodeGeneratorResponse`. And then open the door for writing plugin use pure php code.

---

Maybe google not use PHP widely. The support for php is still limited. For example, the `protoc-php-gen` cannot generate code for `proto2`. However, the PHP is used widely all over the world and deserve to be invested more work. And I wish to make my own contribute to make protoc better and more convenient for PHP developer.

Thank you.

lvht:
@TeBoring would you like to offer any more comment?

lvht:
ping @TeBoring

lvht:
@TeBoring Happy New Year. But would you like to offer any feedback?

optiman:
@TeBoring Would you be so kind as to review this pull request or recommend someone else who could do it?
I think, this PR would be greatly beneficial for many developers and would accelerate the adoption of Protocol Buffers and gRPC throughout the PHP community.

xfxyjwf:
ping @TeBoring

TeBoring:
I understand that you have the need to develop code gen in php. Could you leave it in your own repository? I don't see the need to put it in our repository.

PR #4144, opened 2018-01-05
Add TextFormat::Parser AllowUnknownField

skebanga:
As in [#2092](https://github.com/google/protobuf/pull/2092) which added `AllowUnknownField` support for Java, this does the same for C++.


skebanga:
I signed it!

PR #4145, opened 2018-01-05
Add TextFormat::Parser AllowUnknownField

skebanga:
As in [#2092](https://github.com/google/protobuf/pull/2092) which added `AllowUnknownField` support for Java, this does the same for C++.

### DIFF
--- a/src/google/protobuf/text_format.h
+++ b/src/google/protobuf/text_format.h
@@ -502,6 +502,14 @@ class LIBPROTOBUF_EXPORT TextFormat {
       allow_case_insensitive_field_ = allow;
     }
 
+    // Allow unknown fields to be ignored.
+    // This is not advisable if the input is manually created and human errors
+    // (like typos) need to be caught.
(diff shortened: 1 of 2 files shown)


skebanga:
Is anyone going to review and merge this please?

xfxyjwf:
We intentionally disabled the ability to skip unknown fields in text format because text format is not designed to be a wire format. I suggest you switch to proto binary or JSON format instead.

skebanga:
@xfxyjwf What about this pull request, which added this feature for Java: https://github.com/google/protobuf/pull/2092

All this pull request does is to the same for C++

skebanga:
@xfxyjwf you mention I should use JSON format. How do I do this in C++ please? I can only see JSON format for Java?

xfxyjwf:
For C++ JSON conversion, see:
https://github.com/google/protobuf/blob/master/src/google/protobuf/util/json_util.h

skebanga:
Thanks!

PR #4147, opened 2018-01-05
Fix ValidateDateTime: check day instead month

datacompboy:
Found with PVS-Studio static analyser, see 
https://www.viva64.com/en/b/0550/


acozzette:
ok to test (this comment will trigger the other test runs)

PR #4158, opened 2018-01-09
Fix java benchmark to use parser, fix cpp benchmark new arena to use …

BSBandme:
Fix java benchmark to use parser, fix cpp benchmark new arena to use …

### DIFF
--- a/benchmarks/cpp_benchmark.cc
+++ b/benchmarks/cpp_benchmark.cc
@@ -121,9 +121,10 @@ class ParseNewArenaFixture : public Fixture {
   virtual void BenchmarkCase(benchmark::State& state) {
     WrappingCounter i(payloads_.size());
     size_t total = 0;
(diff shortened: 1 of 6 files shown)


benchmarks/cpp_benchmark.cc
```diff
@@ -121,9 +121,10 @@ class ParseNewArenaFixture : public Fixture {
   virtual void BenchmarkCase(benchmark::State& state) {
     WrappingCounter i(payloads_.size());
     size_t total = 0;
+    Arena arena;
 
     while (state.KeepRunning()) {
-      Arena arena;
+      arena.Reset();
```
xfxyjwf: Does this make any difference in the benchmark result?

BSBandme: Seems a little different for some messages.

PR #4160, opened 2018-01-09
Guards for older GCC versions

### LINKED ISSUE #4156: Old versions of gcc give warnings for pragma diagnostic push and pop
`#pragma GCC diagnostic push` and `#pragma GCC diagnostic pop` are only supported in gcc >= 4.6. 

A solution to this would be to change the guards from
```c++ 
#ifdef __GNUC__
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wstrict-aliasing"
#endif  // __GNUC__
```
and
```c++ 
#ifdef __GNUC__
  #pragma GCC diagnostic pop
#endif  // __GNUC__
```
To: 
```c++ 
#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wstrict-aliasing"
#endif  // __GNUC__
```
and
```c++ 
#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
  #pragma GCC diagnostic pop
#endif  // __GNUC__
```

There are approximately a dozen header files that do this, as well as the cpp_file.cc codegen.  This could be applied to the 3.?.x branches, but might not be needed on master....if master is now c++11, gcc versions before 4.6 might not work with it anyway...I guess it depends on how good various gcc's -std=c++0x support was back then. Might not hurt to apply to master as well.

teeks99:
Fixes #4156

### DIFF
--- a/src/google/protobuf/any.pb.h
+++ b/src/google/protobuf/any.pb.h
@@ -197,7 +197,7 @@ class LIBPROTOBUF_EXPORT Any : public ::google::protobuf::Message /* @@protoc_in
 
 // ===================================================================
 
-#ifdef __GNUC__
+#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
   #pragma GCC diagnostic push
   #pragma GCC diagnostic ignored "-Wstrict-aliasing"
 #endif  // __GNUC__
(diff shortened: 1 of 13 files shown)


xfxyjwf:
Can you help recreate this PR? It seems many of the tests are failing with this change. I'm not sure if it's this PR or our test setup.

PR #4168, opened 2018-01-12
Use versions module from Skylib for version checking 

jin:
Many Bazel repositories implement their own version comparison functions - one implementation is now in [bazelbuild/bazel-skylib](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/versions.bzl).

Further details: https://github.com/bazelbuild/bazel/issues/4433.


jin:
cc @xfxyjwf

PR #4191, opened 2018-01-19
PHP: disable unicode escape for json

lvht:
Add the `JSON_UNESCAPED_UNICODE` options so that characters like **中国** will not be escaped into **\u4e2d\u56fd** to improve serialize and transfer performance.

### DIFF
--- a/php/src/Google/Protobuf/Internal/GPBJsonWire.php
+++ b/php/src/Google/Protobuf/Internal/GPBJsonWire.php
@@ -226,7 +226,7 @@ class GPBJsonWire
                 $output->writeRaw("\"", 1);
                 break;
             case GPBType::STRING:
-                $value = json_encode($value);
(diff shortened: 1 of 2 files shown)


lvht:
@TeBoring Happy New Year. But would you like to offer any feedback?

TeBoring:
https://github.com/interagent/http-api-design/issues/46
It seems unicode + escaped control character is preferred.
Does ```JSON_UNESCAPED_UNICODE``` still escape control characters?

TeBoring:
BTW, you probably also need to check the c implementation behaves accordingly. https://github.com/google/protobuf/tree/master/php/ext/google/protobuf

lvht:
Yes, `JSON_UNESCAPED_UNICODE` escape control characters.

TeBoring:
If you use the c extension of php, it should already avoid unnecessary escaping.

PR #4195, opened 2018-01-19
Adds an option in C++ JSON parser to ignore unrecognized enum values

alexey-malov:
This PR adds `JsonParseOptions::ignore_unknown_enum_values` option to treat unknown enum values as 0.

For example, if we have an application compiled with the following protobuf scheme:
```
syntax = "proto3";
enum ShapeType
{
    DEFAULT = 0;
    RECTANGLE = 1;
    ELLIPSE = 2;
}
message Shape
{
    ShapeType type = 1;
}
```

It will fail to parse the following JSON created with the newer version of an application:
```json
{
  "type":"ROUNDED_RECTANGLE"
}
```
If you set the `ignore_unknown_enum_values` option to `true`, the application will be able to parse this JSON. The `Shape.type` field will be set to `DEFAULT` value (0). The application can handle this situation, for instance, by replacing a shape with one of the supported shape types.
If the `ignore_unknown_enum_values` option is `false`, the JSON parser will have existing behavior.

### DIFF
--- a/src/google/protobuf/util/internal/datapiece.cc
+++ b/src/google/protobuf/util/internal/datapiece.cc
@@ -272,7 +272,8 @@ StatusOr<string> DataPiece::ToBytes() const {
 }
 
 StatusOr<int> DataPiece::ToEnum(const google::protobuf::Enum* enum_type,
-                                bool use_lower_camel_for_enums) const {
+                                bool use_lower_camel_for_enums,
+                                bool ignore_unknown_enum_values) const {
   if (type_ == TYPE_NULL) return google::protobuf::NULL_VALUE;
 
   if (type_ == TYPE_STRING) {
@@ -305,6 +306,10 @@ StatusOr<int> DataPiece::ToEnum(const google::protobuf::Enum* enum_type,
       value = FindEnumValueByNameWithoutUnderscoreOrNull(enum_type, enum_name);
       if (value != NULL) return value->number();
     }
+
+    // If ignore_unknown_enum_values is true an unknown enum value is treated
+    // as the default
+    if (ignore_unknown_enum_values) return enum_type->enumvalue(0).number();
   } else {
     // We don't need to check whether the value is actually declared in the
     // enum because we preserve unknown enum values as well.
--- a/src/google/protobuf/util/internal/datapiece.h
+++ b/src/google/protobuf/util/internal/datapiece.h
@@ -164,7 +164,8 @@ class LIBPROTOBUF_EXPORT DataPiece {
   // If the value is not a string, attempts to convert to a 32-bit integer.
   // If none of these succeeds, returns a conversion error status.
   util::StatusOr<int> ToEnum(const google::protobuf::Enum* enum_type,
-                               bool use_lower_camel_for_enums) const;
+                               bool use_lower_camel_for_enums,
+                               bool ignore_unknown_enum_values) const;
 
  private:
   // Disallow implicit constructor.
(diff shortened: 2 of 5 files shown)


alexey-malov:
I have signed a CLA

alexey-malov:
I signed the CLA. What should I do to remove cla: no mark?

xfxyjwf:
It seems the commit is not associated with your github account. Can you update the email address you used in your commit?

alexey-malov:
I updated the commit email

alexey-malov:
I updated some comments and formatting in my pull request code

src/google/protobuf/util/json_util.h
```diff
@@ -48,8 +48,11 @@ namespace util {
 struct JsonParseOptions {
   // Whether to ignore unknown JSON fields during parsing
   bool ignore_unknown_fields;
+  // Whether to ignore unrecognized values in enum fields
+  bool ignore_unknown_enum_values;
```
xfxyjwf: Instead of adding a new flag, could you update the behavior of the existing ignore_unknown_fields behvaior to ignore unknown enum values as well?

alexey-malov: Sure, I'll do

alexey-malov: I updated my pull request. Now JsonParseOptions::ignore_unknown_files allows ignoring unknown enum values as well. No need for additional option.

src/google/protobuf/util/internal/protostream_objectwriter.h
```diff
@@ -94,10 +94,13 @@ class LIBPROTOBUF_EXPORT ProtoStreamObjectWriter : public ProtoWriter {
     // the field name.
     bool use_lower_camel_for_enums;
 
+	bool ignore_unknown_enum_values;
```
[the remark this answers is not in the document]
alexey-malov: I removed this option since ignore_unknown_fields option is used for ignoring enum values.

src/google/protobuf/util/internal/datapiece.cc
```diff
@@ -305,6 +306,9 @@ StatusOr<int> DataPiece::ToEnum(const google::protobuf::Enum* enum_type,
       value = FindEnumValueByNameWithoutUnderscoreOrNull(enum_type, enum_name);
       if (value != NULL) return value->number();
     }
+
+    // If ignore_unknown_enum_values is true an unknown enum value is treated as 0
+    if (ignore_unknown_enum_values) return 0;
```
xfxyjwf: I think we should return enum_type->enumvalue(0)->number() here. That's the default value you will get when an enum value is not set. In proto3, it's guaranteed to be 0, however, in proto2, it can be something else.

alexey-malov: fixed

alexey-malov:
I wonder if there is anything else I should do for this pull request.

src/google/protobuf/util/internal/datapiece.cc
```diff
@@ -305,6 +306,9 @@ StatusOr<int> DataPiece::ToEnum(const google::protobuf::Enum* enum_type,
       value = FindEnumValueByNameWithoutUnderscoreOrNull(enum_type, enum_name);
       if (value != NULL) return value->number();
     }
+
+    // If ignore_unknown_enum_values is true an unknown enum value is treated as 0
```
xfxyjwf: Please update the comment. It's treated as the default.

alexey-malov: I updated the commend and squashed my commits into one commit

alexey-malov:
Done

alexey-malov:
I wonder if this pull request can be merged into the library?

PR #4196, opened 2018-01-19
cmake: privately add sources to targets

mathstuf:
Public sources get added to dependencies of the target as well which can
cause duplicate symbols.


acozzette:
ok to test (this comment will trigger the other CI builds)

acozzette:
Hmm, Jenkins is showing an error but I can't tell if it's related to this or not. Let me rerun the tests just to be safe.

acozzette:
retest this please (this comment will trigger the CI runs to execute again)

PR #4204, opened 2018-01-21
remove bind for six

shariat:
- removing bind is recommended 
https://github.com/bazelbuild/bazel/issues/1952
and
https://docs.bazel.build/versions/master/best-practices.html#repository-rules

- it conflicts with other external repos that have `@six` like `rules_docker`

I recommend the rest of bind rules be removed as well.


shariat:
any hope merging this?

pherl:
How would this change the existing users' setup? They need to use load the repositories.bzl and call protobuf_dependencies()? Is there any migration path so we can roll out this change without breaking user build in the next release?

Also this approach forces protobuf to stick to a specific six version. What if there are multiple libraries using the six but specifying different versions? This makes it impossible for the end users to use those libraries together without causing symbol conflicts. While by using bind(), it's the end users' choice which version the dependency sticks to.

I don't see using the archive is a clear win over bind() approach. There are still open issues.

shariat:
yes, they have to call `protobuf_dependencies()`. The different version problem is kind of fixed by checking if the lib is already loaded or not. This is what other libraries are doing as well https://github.com/bazelbuild/rules_docker/blob/master/container/container.bzl#L84

shariat:
main discussion is here https://github.com/bazelbuild/bazel/issues/1952

pherl:
I commented on the thread.

shariat:
@pherl btw, users do not have to call `protobuf_dependencies()`. `protobuf_dependencies()` is for convenience. Users can define the dependencies as before.

pherl:
My understanding is users have to change bind() to a explicit repo named "six". This has to been done atomically when users upgrade their protobuf dependency. In this case, the change on the protobuf is a breaking change.

pherl:
Because we use it? https://github.com/google/protobuf/search?utf8=%E2%9C%93&q=six&type=

shariat:
yes, but `//external:six` is only used in `//:protobuf_python`. the output of that rule is `_pb2.py` files that do not import `six`.

pherl:
`//:protobuf_python` is the only public target for users to depend on. It transitively include `//:python_srcs` which requires the `six` dependency. I guess a cleaner fix is to move the the `six` dependency to `//:python_srcs` where the library is actually used.

PR #4207, opened 2018-01-22
Update minimal Python version to 2.7

mapx:
Python 2.6 is marked as dropped elsewhere.

### DIFF
--- a/python/google/protobuf/json_format.py
+++ b/python/google/protobuf/json_format.py
@@ -379,11 +379,7 @@ def Parse(text, message, ignore_unknown_fields=False):
   """
   if not isinstance(text, six.text_type): text = text.decode('utf-8')
   try:
-    if sys.version_info < (2, 7):
-      # object_pair_hook is not supported before python2.7
-      js = json.loads(text)
-    else:
-      js = json.loads(text, object_pairs_hook=_DuplicateChecker)
+    js = json.loads(text, object_pairs_hook=_DuplicateChecker)
   except ValueError as e:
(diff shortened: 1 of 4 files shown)


hugovk:
Some more 2.6 stuff to be removed:

https://github.com/google/protobuf/blob/master/python/setup.py#L226-L228

And this can be cleaned up:

https://github.com/google/protobuf/blob/master/python/setup.py#L158

And there's some other `version_info` things to be checked too.

3.5 - 3.6 can be added here:

https://github.com/google/protobuf/blob/master/python/setup.py#L246

acozzette:
ok to test (this comment will trigger the other CI test runs)

python/google/protobuf/json_format.py
```diff
@@ -379,11 +379,7 @@ def Parse(text, message, ignore_unknown_fields=False):
   """
   if not isinstance(text, six.text_type): text = text.decode('utf-8')
   try:
-    if sys.version_info < (2, 7):
-      # object_pair_hook is not supported before python2.7
-      js = json.loads(text)
-    else:
-      js = json.loads(text, object_pairs_hook=_DuplicateChecker)
+    js = json.loads(text, object_pairs_hook=_DuplicateChecker)
```
anandolee: Can you change the testDuplicateField in json_format_test.py as well? 
https://github.com/google/protobuf/blob/master/python/google/protobuf/internal/json_format_test.py#L793

mapx: Sure. The PR is amended.

PR #4179, opened 2018-01-17
enable source link for csharp

ctaggart:
@jskeet Thanks for adopting SourceLink.Create.CommandLine in the [Google Cloud .NET tools](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs#L83). SourceLink 2.7 helps add the portable pdb files to the nupkg as well.

PR #4208, opened 2018-01-22
Update .NET SDK to 2.0.3

jskeet:
This will allow SourceLink as per #4179, and mean that we can use C#
7.X language features in the library (but not in generated code).

This does not affect which platforms we're *targeting*, so end users
won't see any difference.

(Will assign a reviewer when CI is happy.)


gentlejack:
Files changed.

jskeet:
@gentlejack: I'm not sure what your comment is meant to indicate. Could you elaborate?

jskeet:
It looks like Travis is stalled here - @anandolee, do you know anything about this?

xfxyjwf:
@jskeet All travis tests should now be covered by corresponding kokoro test cases. For C#, I think it's the "Linux C#" one (which is passing for this PR). You can click the details link and there will be a sponge link you can view the test result:
https://sponge.corp.google.com/target?id=c5df01e1-3f4c-436e-8158-267952cbc0b0&target=protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fcsharp%2Fpresubmit&searchFor=

The plan is to replace both jenkins/travis with kokoro eventually. Let me know if there is something covered by travis but not by kokoro,

jskeet:
@xfxyjwf: Does kokoro use the `.travis.yml` file to work out what to install? If not, then a) I've probably missed something in this change, and b) we won't know whether Travis works without it running.

xfxyjwf:
oh, sorry, I haven't checked the actual change in this PR and it seems you are changing the test environment settings. Kokoro doesn't use .travis.yml, instead it uses scripts in the kokoro directory. The one for csharp is:
https://github.com/google/protobuf/blob/master/kokoro/linux/csharp/build.sh

The setup is done in this file:
https://github.com/google/protobuf/blob/master/kokoro/linux/prepare_build_linux_rc

I guess you can try to change this file and check the kokoro result of "Linux C#". It should finish much faster than travis.

jskeet:
Okay - will do that in the morning. I'd still like to see Travis pass or fail with this change though...

jskeet:
Added another commit for the kokoro environment - although it was already green, which surprises me. It's possible that something else was already installing .NET Core SDK 2.0, although that seems a tad unlikely.

jskeet:
Hmm... and the C++ tests *still* haven't started.

Do you think we're safe to merge this change? I suspect so, but I wouldn't want to go ahead without double checking.

xfxyjwf:
I think it's fine to merge as long as other tests are passing. Travis is too slow and I usually merge without it nowadays.

jskeet:
Thanks - done!

PR #4209, opened 2018-01-22
Removed using statements from common.h

acozzette:
These statements pulled a bunch of symbols from the std namespace into
the global namespace. This commit removes all of them except for
std::string, which is a bit trickier to remove.


acozzette:
@xfxyjwf Friendly ping, all the tests look good so I think this is ready to go.

PR #4223, opened 2018-01-25
Java TestUtil.readTextFromFile(): normalize CR-LF to LF

os72:
(allows TextFormatTest to pass on Windows too)


xfxyjwf:
Can you make sure the line-endings in the test txt files are not converted to CR-LF instead? You should be able to do that with some git configuration if you are using git clone.

PR #4226, opened 2018-01-25
Add Haskell implementations

themattchan:
Add Awake Security's Haskell Protobuf and GRPC implementations to third party impl list


themattchan:
I signed it

themattchan:
I signed it!

PR #4250, opened 2018-01-30
Adds better Ruby error messages for TypeErrors by including the field name

ebenoist:
Currently the TypeErrors thrown by the generated classes do not include the field name in the error description, which can make debugging these issues a matter of bisecting messages until the error is repeatable. Including the name of the field in the TypeErrors thrown by these classes helps immensely with their debug-ability.

I've added extra tests to demonstrate this new behavior, but wanted to get some feedback before tackling the field names for the RepeatedField and Map type containers, neither of which currently hold their field name. If the above approach for primitives seems sane, I'd love some guidance on how to add the field names to those type containers so they can be passed to the type checkers for richer error messages.

* TODO: Add field name for repeated fields and maps (anything using the typed field containers)
* TODO: jRuby support. I am happy to add the jRuby support here as well once we everyone is onboard with this change.

### DIFF
--- a/ruby/tests/basic.rb
+++ b/ruby/tests/basic.rb
@@ -261,6 +261,7 @@ module BasicTest
       e = assert_raise NoMethodError do
         TestMessage.new.hello = "world"
       end
+
       assert_match(/hello/, e.message)
     end
 
@@ -283,33 +284,45 @@ module BasicTest
 
     def test_type_errors
       m = TestMessage.new
-      assert_raise TypeError do
+      e = assert_raise TypeError do
         m.optional_int32 = "hello"
       end
-      assert_raise TypeError do
+      assert_equal e.message, "Expected number type for integral field: optional_int32"
+
+      e = assert_raise TypeError do
         m.optional_string = 42
       end
-      assert_raise TypeError do
+      assert_equal e.message, "Invalid argument for string field: optional_string"
+
+      e = assert_raise TypeError do
         m.optional_string = nil
       end
-      assert_raise TypeError do
+      assert_equal e.message, "Invalid argument for string field: optional_string"
+
+      e = assert_raise TypeError do
         m.optional_bool = 42
       end
-      assert_raise TypeError do
+      assert_equal e.message, "Invalid argument for boolean field: optional_bool"
+
+      e = assert_raise TypeError do
         m.optional_msg = TestMessage.new  # expects TestMessage2
       end
+      assert_equal e.message, "Invalid type BasicTest::TestMessage to assign to submessage field: optional_msg"
 
-      assert_raise TypeError do
+      e = assert_raise TypeError do
         m.repeated_int32 = []  # needs RepeatedField
       end
+      assert_equal e.message, "Expected repeated field array"
 
-      assert_raise TypeError do
+      e = assert_raise TypeError do
         m.repeated_int32.push "hello"
       end
+      assert_equal e.message, "Expected number type for integral field: foo" # TODO
 
-      assert_raise TypeError do
+      e = assert_raise TypeError do
         m.repeated_msg.push TestMessage.new
       end
+      assert_equal e.message, "Invalid type BasicTest::TestMessage to assign to submessage field: foo" # TODO
     end
 
     def test_string_encoding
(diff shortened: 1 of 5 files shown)


ebenoist:
I signed it!

ebenoist:
Debugging type errors in Ruby is very difficult because of the poor error messages here. What can be done to move this PR along?

anandolee:
ping @TeBoring

xfxyjwf:
ping @TeBoring

ruby/tests/basic.rb
```diff
@@ -283,33 +284,45 @@ def test_initialization_map_errors
...
+      e = assert_raise TypeError do
         m.repeated_int32 = []  # needs RepeatedField
       end
+      assert_equal e.message, "Expected repeated field array"
 
-      assert_raise TypeError do
+      e = assert_raise TypeError do
         m.repeated_int32.push "hello"
       end
+      assert_equal e.message, "Expected number type for integral field: foo" # TODO
```
TeBoring: There seems to be an error for this line.

ebenoist: Alternatively I can punt on repeated fields for now if this is enough of an improvement on a first pass.

TeBoring:
Without this change, when there is error, does the place of calling setter show in stack trace?

ruby/tests/basic.rb
ebenoist: I couldn't agree more :). Why don't I back out the repeated field stuff for now and we can focus on just the setters for primitives?

ebenoist:
@TeBoring Currently and with this change, yes, a TypeError is thrown and the stack trace should emit from that location.

TeBoring:
FMO, the current error message + stack track is already enough for debugging.
Besides, the native_slot_set was shared for map/repeated field. If you do detailed error for primitive field but do something else for map/repeated field, the logic will become unnecessarily complicate.
Sorry that for now I don't want this change.

ebenoist:
@TeBoring Sorry to hear that. This is particularly a problem not for individual setters, but when using the constructor interface to set a number of fields. When using the constructor you have have no idea which field doesn't currently meet the type expectations. This means literally having to bisect larger structs with many fields to debug, _which_ of the fields is throwing an error.

I'll give an example.

```Ruby
my_msg = MyMsg.new({
  field_one: "2",
  field_two: 1,
  field_three: {}
})
```

If all of those fields expect an integer, I'll get a stack track that emits from the construction of MyMsg. In the stack trace there is _no_ information about which field is currently emitting a TypeError.

ebenoist:
@TeBoring Here's a great practical example of the current issue:

```
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}

Reverb::Search::Service
  #core_completions
    performs a core completions search (FAILED - 1)

Failures:

  1) Reverb::Search::Service#core_completions performs a core completions search
     Failure/Error:
       Reverb::Search::SuggestOption.new({
         text: option.text,
         id: option.id,
         score: option.score,
         output: option.doc[:output].to_s,
         core_payload: present_core_payload(option.doc)
       })

     TypeError:
       Invalid argument for string field.
     # ./lib/service/handlers/core_completions_handler.rb:43:in `initialize'
     # ./lib/service/handlers/core_completions_handler.rb:43:in `new'
     # ./lib/service/handlers/core_completions_handler.rb:43:in `block in present_core_options'
     # ./lib/service/handlers/core_completions_handler.rb:42:in `map'
     # ./lib/service/handlers/core_completions_handler.rb:42:in `present_core_options'
     # ./lib/service/handlers/core_completions_handler.rb:36:in `block in present_core_completions'
     # ./lib/service/handlers/core_completions_handler.rb:33:in `map'
     # ./lib/service/handlers/core_completions_handler.rb:33:in `present_core_completions'
     # ./lib/service/handlers/core_completions_handler.rb:26:in `handle'
     # ./lib/service/service.rb:20:in `core_completions'
     # ./spec/service/handlers/core_completions_handler_spec.rb:60:in `block (3 levels) in <top (required)>'
```

With the current error I have _no_ idea which of the fields currently has an issue, as the error raises from line 43, where the constructor is called.

PR #4252, opened 2018-01-30
Allows the Ruby json marshaller to be passed json marshal options

ebenoist:
The JSON marshaller accepts both the preseve_field_names and emit_defaults options in the C bindings, but these options are not easily exposed to the Ruby encode/decode methods. This PR adds an options hash, empty by default, that can be used to pass options at runtime.

### DIFF
--- a/ruby/lib/google/protobuf.rb
+++ b/ruby/lib/google/protobuf.rb
@@ -60,8 +60,8 @@ module Google
       msg.to_proto
     end
 
-    def self.encode_json(msg)
-      msg.to_json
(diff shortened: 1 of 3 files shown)


ebenoist:
Fixes https://github.com/google/protobuf/pull/4052 as well

ebenoist:
This is a simple change that compiles better with Ruby's native `to_json` function and better exposes some of the marshal options that the Ruby proto library already supports.

anandolee:
ping @TeBoring

xfxyjwf:
ping @TeBoring

PR #4255, opened 2018-01-31
[Ruby] Adds a base class for all explicitly raised TypeErrors

ebenoist:
* This allows for ruby code to catch and handle Protobuf
    TypeErrors separately from the standard Ruby TypeError

  * Maintains backwards compatibility by having the new
    Google::Protobuf::TypeError inherit from the base
    TypeError. Any code that was catching TypeError should
    continue to work.

### DIFF
--- a/ruby/lib/google/protobuf.rb
+++ b/ruby/lib/google/protobuf.rb
@@ -37,6 +37,7 @@ module Google
   module Protobuf
     class Error < StandardError; end
     class ParseError < Error; end
+    class TypeError < ::TypeError; end
   end
 end
 
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -812,7 +812,7 @@ VALUE FieldDescriptor_submsg_name_set(VALUE _self, VALUE value) {
   upb_fielddef* mut_def = check_field_notfrozen(self->fielddef);
   const char* str = get_str(value);
   if (!upb_fielddef_hassubdef(self->fielddef)) {
-    rb_raise(rb_eTypeError, "FieldDescriptor does not have subdef.");
+    rb_raise(cTypeError, "FieldDescriptor does not have subdef.");
   }
   CHECK_UPB(upb_fielddef_setsubdefname(mut_def, str, &status),
             "Error setting submessage name");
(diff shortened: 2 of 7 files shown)


ruby/lib/google/protobuf.rb
```diff
@@ -37,6 +37,7 @@ module Google
   module Protobuf
     class Error < StandardError; end
     class ParseError < Error; end
+    class TypeError < ::TypeError; end
```
kylecrum: shouldn't this inherit from Google::Protobuf::Error instead of ruby's TypeError?

ebenoist: Possibly, but that would make this change a breaking one. Any code that is currently relying on this being a `TypeError` would start failing if the parent class is changed. By making this a subclass of the standard `TypeError` existing rescues should continue to work, but folks can opt in to catching the more specific class now.

Technically you could make TypeError the "base" class here (TypeError inherits from StandardError) instead of StandardError and it would then allow someone to rescue any Protobuf::Error and it should still be backwards compatible.

```Ruby
def foo; raise TypeError; rescue StandardError => e; puts "caught"; end
>> foo
caught
```

```Ruby
module Google
  module Protobuf
    class Error < ::TypeError; end
    class ParseError < Error; end
    class TypeError < Error; end
  end
end
```

TeBoring:
I think in test, you should still test against ::TypeError.
Please also add a TODO saying we will change ::TypeError to Googo::Protobuf::TypeError in test when we will it's safe to do.

ebenoist:
Will do!

ebenoist:
@TeBoring I already have a test that shows that `Google::Protobuf::TypeError` inherits from `::TypeError`

```Ruby
# Google::Protobuf::TypeError should inherit from TypeError for backwards compatibility
assert_true e.is_a?(::TypeError)
```

I can add a TODO to that line. I'm also happy to change the tests back to `::TypeError`, but this should eliminate having to make many changes when that change _can_ be safely made.

TeBoring:
Does ```assert_raise``` accept subclass? I see our conformance test are broken.

ebenoist:
@TeBoring -- it should take any class in the hierarchy. Is there a failure I can take a look at? Happy to help if there's an issue.

TeBoring:
You can check ruby/conformance

TeBoring:
https://github.com/google/protobuf/pull/4829

TeBoring:
See MacOS Ruby 2.1

haberman:
What is the rationale for this change, out of curiosity?

What is the use case for catching protobuf type errors separately from regular Ruby type errors?

PR #4259, opened 2018-01-31
Various CMake Update

Mizux:
1) Add alias so when using protobuf as subproject you can still use `protobuf::libprotobuf` like in the module Protobuf cf https://cmake.org/cmake/help/latest/module/FindProtobuf.html

2) Change MSVC default behaviour to be the same than unix -> let's discuss about  it !  
Personally for using protobuf for google/or-tools on windows, I didn't find difficult to provide zlib as well
(and protobuf cmake know how to deal if zlib is not found BTW)  
Concerning /MD as default even for static library I'm based on:
- gRPC default to /MD cf https://github.com/grpc/grpc/blob/master/cmake/msvc_static_runtime.cmake#L15
- Different modules (i.e. DLLs or EXEs) can exchange "memory ownership". For example, a memory allocated through new/malloc in one module can be reallocated/deleted/freed by another...


Mizux:
MacOS seems to have an error in `rake` installation
```
...
Installing rake 12.3.0
Gem::InstallError: rake requires Ruby version >= 2.0.0.
An error occurred while installing rake (12.3.0), and Bundler cannot continue.
Make sure that `gem install rake -v '12.3.0'` succeeds before bundling.

In Gemfile:
  rake-compiler was resolved to 0.9.9, which depends on
    rake
```

cmake/libprotobuf-lite.cmake
```diff
@@ -62,5 +62,7 @@ if(MSVC AND protobuf_BUILD_SHARED_LIBS)
     PRIVATE LIBPROTOBUF_EXPORTS)
 endif()
 set_target_properties(libprotobuf-lite PROPERTIES
+		VERSION ${protobuf_VERSION}
```
xfxyjwf: Please fix the indentation.

xfxyjwf:
Other changes seem fine to me, but I'm not sure about the /MD change. The flag is only for the static library build. If you are building protobuf as DLL, that flag is ignored.

cmake/libprotobuf-lite.cmake
Mizux: DONE

Mizux:
@xfxyjwf  Concerning /MD it's more using libprotobuf as static lib and incorporate it in a DLL (to have a standalone DLL on windows since rpath / package manager is still not the norm).
e.g. Having a FooBarApp which depend on Foo.dll and Bar.dll project which integrate the static libprotobuf.lib...

After that if you prefer to leave as it is now, well, I guess I have to leave with forcing`set(protobuf_MSVC_STATIC_RUNTIME OFF)` and remove this path from the PR...

Please, also notice that gRPC allways use /MD by default (I guess it is also the default cmake behaviour but i can't retrieve a source on this assumption)
and deactivate it on Protobuf https://github.com/grpc/grpc/blob/master/cmake/protobuf.cmake#L32

xfxyjwf:
@Mizux It's usually not a good idea to static-link protobuf into your dynamic library because it can cause conflicts (or just bizarre crash) when you have multiple dynamic libraries linking-in multiple copies of protobuf. I would prefer to keep the current default value of protobuf_MSVC_STATIC_RUNTIME. Except for that, this PR looks good to me.

Oh, could you sign Google CLA to pass the CLA check? if you are a member of Google organization, making your membership public should also pass the CLA check.

Mizux:
So I'll remove the commit concerning the static library stuff...

For the CLA need to figure out why is not working since I'm part of the google organization but only have write access to google/or-tools which I'm maintaining...
Otherwise as a temp ask to my manager how to sign the CLA which is weird since all my stuff already belongs to google....

willnorris:
@googlebot rescan

xfxyjwf:
@willnorris signcla still says we can't accept changes from github username 'Mizux'. Is the CLA bot using some other source? How can I manually check if the CLA bot is correct?

willnorris:
looks like we're treating GitHub usernames as case sensitive for some odd reason (http://signcla/Mizux fails, http://signcla/mizux passes).  I'll file a bug for that.

PR #4262, opened 2018-01-31
Only check filenames when end with .py in _CalledFromGeneratedFile()

### LINKED ISSUE #2896: Python generated file compiled with Cython causes error with protobuf 3.2.0
When a _pb2 is compiled with Cython, and the C++ backend of the Python library is turned on (as it now is on Linux for the PyPI deployed protobuf, as of 3.2.0), instantiating any message object gives an error. I am using Python 3.5. Here's an example of the error:

```
  File "bug.py", line 1, in <module>
    from addressbook_pb2 import Person
  File "addressbook_pb2.py", line 34, in init addressbook_pb2 (addressbook_pb2.c:1997)
    _descriptor.EnumValueDescriptor(
  File "/home/arthur/envs/default/lib/python3.5/site-packages/google/protobuf/descriptor.py", line 649, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors should not be created directly, but only retrieved from their parent.
```

The immediate cause of the error is [`_CheckCalledFromGeneratedFile` in `google/protobuf/pyext/message.cc`](https://github.com/google/protobuf/blob/v3.2.0/python/google/protobuf/pyext/message.cc#L2564), which calls [`_CalledFromGeneratedFile` in `google/protobuf/pyext/descriptor.cc`](https://github.com/google/protobuf/blob/v3.2.0/python/google/protobuf/pyext/descriptor.cc#L95). I haven't looked any further than this, but my suspicion is raised by the `strcmp` against `"_pb2.py"` at the end of this function.  I think that for Cython the filename would end with `"_pb2.c"`, but I'm not sure.

Just to be completely explicit, here are steps to reproduce:

1. Create a new Python virtual environment with the relevant packages (I am just using grpio-tools for easy access to protoc):

```
pip install grpcio-tools==1.1.0 protobuf==3.2.0 Cython==0.25.2
```

2. Generate a pb2:

```
wget https://raw.githubusercontent.com/google/protobuf/master/examples/addressbook.proto
python -m grpc_tools.protoc -I. addressbook.proto --python_out=.
```

3. Cythonise and compile this file. The easiest way to do this is with a little setup.py:

```
from distutils.core import setup
from Cython.Build import cythonize
setup(name="bug", ext_modules = cythonize("addressbook_pb2.py"))
```

Then run: `python setup.py build_ext --inplace`

4. Try to instantiate a message:

```
python -c 'from addressbook_pb2 import Person; p = Person()'
```

anandolee:
Cython is not using .py file. Only check filenames when end with .py in _CalledFromGeneratedFile()

Fixes #2896

### DIFF
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -115,7 +115,7 @@ bool _CalledFromGeneratedFile(int stacklevel) {
   }
   
   if ((filename_size < 3) || (strcmp(&filename[filename_size - 3], ".py") != 0)) {
-    // Cython is not using .py file and not at global module scope.
+    // Cython's stack does not have .py file name and is not at global module scope.
     return true;
   }
   
@@ -123,7 +123,6 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     // filename is too short.
     return false;
   }
-
   if (strcmp(&filename[filename_size - 7], "_pb2.py") != 0) {
     // Filename is not ending with _pb2.
     return false;
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -101,6 +101,35 @@ bool _CalledFromGeneratedFile(int stacklevel) {
   if (frame == NULL) {
     return false;
   }
+  
+  if (frame->f_code->co_filename == NULL) {
+    return false;
+  }
+  
+  char* filename;
+  Py_ssize_t filename_size;
+  if (PyString_AsStringAndSize(frame->f_code->co_filename,
+                               &filename, &filename_size) < 0) {
+    // filename is not a string.
+    PyErr_Clear();
+    return false;
+  }
+  
+  if (filename_size < 3 or strcmp(&filename[filename_size - 3], ".py") != 0) {
+    // Cython is not using .py file and not at global module scope.
+    return true;
+  }
+  
+  if (filename_size < 7) {
+    // filename is too short.
+    return false;
+  }
+
+  if (strcmp(&filename[filename_size - 7], "_pb2.py") != 0) {
+    // Filename is not ending with _pb2.
+    return false;
+  }
+  
   while (stacklevel-- > 0) {
     frame = frame->f_back;
     if (frame == NULL) {
@@ -110,29 +139,6 @@ bool _CalledFromGeneratedFile(int stacklevel) {
   if (frame->f_globals != frame->f_locals) {
     // Not at global module scope
     return false;
-  }
-
-  if (frame->f_code->co_filename == NULL) {
-    return false;
-  }
-  char* filename;
-  Py_ssize_t filename_size;
-  if (PyString_AsStringAndSize(frame->f_code->co_filename,
-                               &filename, &filename_size) < 0) {
-    // filename is not a string.
-    PyErr_Clear();
-    return false;
-  }
-  if (filename_size < 7) {
-    // filename is too short.
-    return false;
-  }
-  // Cython is not using .py file. Only check filenames when end with .py
-  if (strcmp(&filename[filename_size - 3], ".py") == 0) {
-    if (strcmp(&filename[filename_size - 7], "_pb2.py") != 0) {
-      // Filename is not ending with _pb2.
-      return false;
-    }
   }
 #endif
   return true;
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -105,7 +105,6 @@ bool _CalledFromGeneratedFile(int stacklevel) {
   if (frame->f_code->co_filename == NULL) {
     return false;
   }
-  
   char* filename;
   Py_ssize_t filename_size;
   if (PyString_AsStringAndSize(frame->f_code->co_filename,
@@ -115,7 +114,7 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     return false;
   }
   
-  if (filename_size < 3 or strcmp(&filename[filename_size - 3], ".py") != 0) {
+  if ((filename_size < 3) || (strcmp(&filename[filename_size - 3], ".py") != 0)) {
     // Cython is not using .py file and not at global module scope.
     return true;
   }
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -127,9 +127,12 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     // filename is too short.
     return false;
   }
-  if (strcmp(&filename[filename_size - 7], "_pb2.py") != 0) {
-    // Filename is not ending with _pb2.
-    return false;
+  // Cython is not using .py file. Only check filenames when end with .py
+  if (strcmp(&filename[filename_size - 3], ".py") == 0) {
+    if (strcmp(&filename[filename_size - 7], "_pb2.py") != 0) {
+      // Filename is not ending with _pb2.
+      return false;
+    }
   }
 #endif
   return true;
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -101,7 +101,13 @@ bool _CalledFromGeneratedFile(int stacklevel) {
   if (frame == NULL) {
     return false;
   }
-  
+  while (stacklevel-- > 0) {
+    frame = frame->f_back;
+    if (frame == NULL) {
+      return false;
+    }
+  }
+
   if (frame->f_code->co_filename == NULL) {
     return false;
   }
@@ -113,12 +119,10 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     PyErr_Clear();
     return false;
   }
-  
   if ((filename_size < 3) || (strcmp(&filename[filename_size - 3], ".py") != 0)) {
     // Cython's stack does not have .py file name and is not at global module scope.
     return true;
   }
-  
   if (filename_size < 7) {
     // filename is too short.
     return false;
@@ -128,12 +132,6 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     return false;
   }
   
-  while (stacklevel-- > 0) {
-    frame = frame->f_back;
-    if (frame == NULL) {
-      return false;
-    }
-  }
   if (frame->f_globals != frame->f_locals) {
     // Not at global module scope
     return false;


arthur-tacca:
Many thanks for creating this pull request. I have just tested it and unfortunately it doesn't fix the problem; I still get the error "Descriptors should not be created directly...". The problem is this piece of code earlier in that function:

    if (frame->f_globals != frame->f_locals) {
      // Not at global module scope
      return false;
    }

Unfortunately, when I was debugging this before it hadn't occured to me to print this out in the stack frame. Here is a typical stack with everything being tested for, with Cython (first is the topmost level of the stack):

    In _CalledFromGeneratedFile, stacklevel = 0...
       [0] isGlob: 1, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/mypkg/myprot_pb2.py
       [1] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [2] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [3] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [4] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [5] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [6] isGlob: 1, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/mypkg/foo.py
       [7] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [8] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [9] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [10] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [11] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [12] isGlob: 1, !NULL: 1, fn is str: 1, fn: run.py
    Checking filename: /home/arthur/envs/protocytest/lib/python3.5/site-packages/mypkg/badprot.py

It seems to me that this check serves two purposes, so here are similar stacks from those two situations but WITHOUT Cython (where the check stands a chance of working):

(1) The check finds where a `_pb2.py` file has been renamed. If the check is bypassed this usually works but is presumably unsupported and certainly would break imports between generated files. Here is an example stack trace:

    In _CalledFromGeneratedFile, stacklevel = 1...
       [0] isGlob: 0, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/protobuf-3.5.1-py3.5-linux-x86_64.egg/google/protobuf/descriptor.py
       [1] isGlob: 1, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/mypkg/badprot.py
       [2] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [3] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [4] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [5] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [6] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [7] isGlob: 1, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/mypkg/badfoo.py
       [8] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [9] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [10] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [11] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [12] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [13] isGlob: 1, !NULL: 1, fn is str: 1, fn: run.py
    Checking filename: /home/arthur/envs/protocytest/lib/python3.5/site-packages/mypkg/badprot.py


(2) The check stops people manually creating `Descriptor` objects, which seems to be the original intention. This seems like an odd check to me because the constructor takes so many arguments it is basically impossible to do by accident. Anyway, here is an example stack trace:


    In _CalledFromGeneratedFile, stacklevel = 1...
       [0] isGlob: 0, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/protobuf-3.5.1-py3.5-linux-x86_64.egg/google/protobuf/descriptor.py
       [1] isGlob: 0, !NULL: 1, fn is str: 1, fn: manual_create_desc.py
       [2] isGlob: 0, !NULL: 1, fn is str: 1, fn: manual_create_desc.py
       [3] isGlob: 1, !NULL: 1, fn is str: 1, fn: manual_create_desc.py
    Not at global scope

So we can NOT change the check to `return true` if we are not at global scope, because it will miss people manually creating `Descriptor` objects, which probably usually happens inside functions or methods.

So here are my suggestions:

  * Remove the check of global scope altogether, and rely on the filename check.
  * Change all the other return statements (e.g. filename not a string). in this function to return true . This indicates an exceptional situation, perhaps along the lines Cythoning.
  * EXCEPT you can't do this for the 7 characters check, because it misses e.g `pb.py`. I think this check needs to be a little more complicated:
    * If the filename length >= 3 characters and ends in `.py`:
    * AND if (the length <7 OR does not end with `_pb2.py`)
    * then return false (the only such return in the function)

Would that be reasonable?

Thanks again for your efforts on this and really sorry my suggestions aren't in the form of a pull request.

anandolee:
arthur-tacca@, Thanks for your investigate. But I am little confusing with what situations makes the Cython check fail.

The first stack example with Cython is checking myprot_pb2.py (stacklevel = 0) and it is global which should be passing.
The second WITHOUT Cython is checking badprot.py (stacklevel = 1) which should fail.
Third one WITHOUT Cython is  checking manual_create_desc.py (stacklevel = 1) which should fail.
They are all correct behaviors.


Are you able to give a stack example when Cython check fail? Stack frame for '_CalledFromGeneratedFile, stacklevel = 1' with Cython would help

arthur-tacca:
Ugh, I'm really sorry. The text in my comment was correct, but that first stack trace was wrong – I must have pasted in the wrong one. That stack trace is where *NOTHING* is Cythoned! There is a top-level script `foo.py`, which imports a `mypkg.foo` , which imports `mypkg.myprot_pb2`, all of which are pure Python (you can see `foo.py`, `mypkg/foo.py` and `mypkg/myprot_pb2.py` in the stack trace). This is a correct usage, and passes the check, as you said.

Here are stack traces from that should work but fail. They come from a program run with a single import of a `_pb2`, but there are actually three calls to `_CalledFromGeneratedFile` that fail; I forced the result to `true` so I could see the later calls. It is the same as the situation I just described for the stack trace in my original comment, except that the two modules in `mypkg` are Cythoned.


    In _CalledFromGeneratedFile, stacklevel = 1...
       [0] isGlob: 0, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/protobuf-3.5.1-py3.5-linux-x86_64.egg/google/protobuf/descriptor.py
       [1] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [2] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [3] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [4] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [5] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [6] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [7] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [8] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [9] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [10] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [11] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [12] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [13] isGlob: 1, !NULL: 1, fn is str: 1, fn: run.py
    Not at global scope
    Checking filename: <frozen importlib._bootstrap>

    In _CalledFromGeneratedFile, stacklevel = 1...
       [0] isGlob: 0, !NULL: 1, fn is str: 1, fn: /home/arthur/envs/protocytest/lib/python3.5/site-packages/protobuf-3.5.1-py3.5-linux-x86_64.egg/google/protobuf/descriptor.py
       [1] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [2] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [3] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [4] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [5] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [6] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [7] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [8] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [9] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [10] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [11] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [12] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [13] isGlob: 1, !NULL: 1, fn is str: 1, fn: run.py
    Not at global scope
    Checking filename: <frozen importlib._bootstrap>

    In _CalledFromGeneratedFile, stacklevel = 0...
       [0] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [1] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [2] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [3] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [4] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [5] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [6] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [7] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap_external>
       [8] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [9] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [10] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [11] isGlob: 0, !NULL: 1, fn is str: 1, fn: <frozen importlib._bootstrap>
       [12] isGlob: 1, !NULL: 1, fn is str: 1, fn: run.py
    Not at global scope
    Checking filename: <frozen importlib._bootstrap>

anandolee:
Thanks for the updated stack traces. I've changed the PR, can you help to test and review?

arthur-tacca:
I notice that instead of removing the global check, you have kept it in, but moved after the check that the filename ends in `.py`. Good idea. The code looks good to me.

I have tested your most recent commit (7b44b6d) with a Cythoned `_pb2.py` (same situation as my earlier comments). The check passes and the import works i.e. your changes fix the problem. Thanks!

anandolee:
Hi haberman@,
Can you help to review? arthur-tacca@ has tested it works in Cython.

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -123,6 +119,10 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     PyErr_Clear();
     return false;
   }
+  if ((filename_size < 3) || (strcmp(&filename[filename_size - 3], ".py") != 0)) {
```
haberman: Could we use C++ string manipulation instead? `strcmp()` scares me.

anandolee: The original code is using strcmp. I also searched the code base, strcmp has been used several times. If we want to use string manipulation instead, I'd prefer to in another change.

amauryfa:
LGTM. This check is only to prevent surprises, so that descriptors appear immutable.
I would not object to remove or disable this function completely, when code is compiled by Cython.

Issue #3038, opened 2017-05-02
Non-functional fallthrough annotation macro (macOS, C++)


aiusepsi:
When compiling the simple protocol buffer shown below, there's an implicit fallthrough warning generated even though the fallthrough is annotated with GOOGLE_FALLTHROUGH_INTENDED. This macro appears to be disabled on Apple platforms in ports.h.

macOS: 10.12.4 (16E195)
Clang: Apple LLVM version 8.1.0 (clang-802.0.42) Target: x86_64-apple-darwin16.5.0
Protobuf: 3.3.0

Minimal repro:

test.proto
```
syntax = "proto3";
message TestMsg {
  map<string, string> test = 1;
}
```

```bash
protoc --cpp_out=. test.proto
clang++ -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk  -I/usr/local/include -std=c++11 -Wimplicit-fallthrough  -c test.pb.cc -o test.pb.o -Werror
```

Output:
```
In file included from test.pb.cc:5:
In file included from ./test.pb.h:32:
In file included from /usr/local/include/google/protobuf/map_field_inl.h:40:
In file included from /usr/local/include/google/protobuf/map_field.h:40:
In file included from /usr/local/include/google/protobuf/map_entry.h:35:
/usr/local/include/google/protobuf/map_entry_lite.h:207:9: error: unannotated fall-through between switch labels [-Werror,-Wimplicit-fallthrough]
        case kValueTag:
        ^
test.pb.cc:140:29: note: in instantiation of member function 'google::protobuf::internal::MapEntryImpl<TestMsg::TestMsg_TestEntry, google::protobuf::Message, std::__1::basic_string<char>,
      std::__1::basic_string<char>, google::protobuf::internal::WireFormatLite::FieldType::TYPE_STRING, google::protobuf::internal::WireFormatLite::FieldType::TYPE_STRING, 0>::MergePartialFromCodedStream'
      requested here
TestMsg::TestMsg_TestEntry::TestMsg_TestEntry() {}
                            ^
/usr/local/include/google/protobuf/map_entry_lite.h:207:9: note: insert '[[clang::fallthrough]];' to silence this warning
        case kValueTag:
        ^
        [[clang::fallthrough]]; 
/usr/local/include/google/protobuf/map_entry_lite.h:207:9: note: insert 'break;' to avoid fall-through
        case kValueTag:
        ^
[...]

PR #4279, opened 2018-02-06
Re-enable GOOGLE_FALLTHROUGH_INTENDED attribute for macOS

alasher:
I was able to replicate the error in #3038 and confirmed that this patch fixes the issue. I guess [[clang::fallthrough]] does work for macOS now.


nico:
I have an imho better fix for this at #4288.

PR #4288, opened 2018-02-07
Remove use of GOOGLE_FALLTHROUGH_INTENDED from protobuf.

nico:
Chrome is running into two issues with the use of this macro
in open-source protobuf (https://crbug.com/809157):

1. GOOGLE_FALLTHROUGH_INTENDED is defined to nothing on `__APPLE__`
   platforms, which blocks us from enabling -Wimplicit-fallthrough
   on Mac and iOS. (We use a hermetic self-built modern clang,
   so whatever Xcode bug that exclusion might be for doesn't apply
   to us.)

2. It's in a public header file, and it's included in a public header file.
   When clang suggests adding [[clang::fallthrough]], it checks if it knows of
   a macro expanding to that and if so, suggests inserting that. Since lots of
   chrome code includes protobuf headers, it often suggests inserting
   GOOGLE_FALLTHROUGH_INTENDED (from protobuf) instead of the correct
   FALLTHROUGH (from chrome's base).

Since the fallthrough doens't do anyting useful, just remove it.
Long ago, this might have had perf impact, but d64a2d9941c36a7bc added a
parsing fast path that calls this switch as slow fallback, so it should
be off the hot path nowadays.

No intended behavior change.

This is the public version of internal change 184824132.

### DIFF
--- a/src/google/protobuf/map_entry_lite.h
+++ b/src/google/protobuf/map_entry_lite.h
@@ -201,8 +201,7 @@ class MapEntryImpl : public Base {
             return false;
           }
           set_has_key();
-          if (!input->ExpectTag(kValueTag)) break;
-          GOOGLE_FALLTHROUGH_INTENDED;
+          break;
 
         case kValueTag:
           if (!ValueTypeHandler::Read(input, mutable_value())) {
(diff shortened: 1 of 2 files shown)


alasher:
> This is the public version of internal change 184824132

Yikes, I guess this means my PR is toast 😄 I agree that this is better than mine, though - assuming there isn't a need to use an intentional fallthrough in the future.

xfxyjwf:
Please hold on this pull request until cl/184824132 is resolved.

nico:
The internal change has landed; please merge this one too :-)

xfxyjwf:
Merging this pull request. Note that protobuf team is not committed to not use fall-through in the future.

nico:
Thanks much!

PR #4289, opened 2018-02-07
Fix all warnings that `./test.sh cpp` prints on my Linux box.

nico:
Mostly -Wunused-function warnings that are mostly true positives,
but also one -Wunused-but-set-variable that pointed out that one
test wasn't testing what it intended to test.

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface_unittest.cc
+++ b/src/google/protobuf/compiler/command_line_interface_unittest.cc
@@ -145,10 +145,6 @@ class CommandLineInterfaceTest : public testing::Test {
   // substring.
   void ExpectErrorSubstring(const string& expected_substring);
 
-  // Like ExpectErrorSubstring, but checks that Run() returned zero.
-  void ExpectErrorSubstringWithZeroReturnCode(
-      const string& expected_substring);
-
   // Checks that the captured stdout is the same as the expected_text.
   void ExpectCapturedStdout(const string& expected_text);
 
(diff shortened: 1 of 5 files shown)


src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -157,9 +153,11 @@ class CommandLineInterfaceTest : public testing::Test {
   void ExpectCapturedStdoutSubstringWithZeroReturnCode(
       const string& expected_substring);
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
```
acozzette: The one thing I'm worried about is that it's getting tough to follow all the ifdefs.

For this one, could you perhaps put an ifdef inside `HasAlternateErrorSubstring()` that makes it keep its current behavior if `defined(_WIN32) && !defined(__CYGWIN__)` is true, but otherwise return false? That way all the ifdef logic is hidden inside the function and most of the other ifdefs can disappear.

src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -190,7 +188,9 @@ class CommandLineInterfaceTest : public testing::Test {
                                      const string& message_name);
   void CheckGeneratedAnnotations(const string& name, const string& file);
 
+#if defined(_WIN32)
   void ExpectNullCodeGeneratorCalled(const string& parameter);
```
acozzette: Since this member function is only used by two Windows-only tests, could you try moving it to a stand-alone function defined right before those Windows tests? I think that would also help simplify the ifdef logic.

acozzette:
@nico Thanks for the fix and cleanup! I made just a couple comments about `command_line_interface_unittest.cc`.

acozzette:
@nico Have you had a chance to look at this again yet?

acozzette:
I'm going to close this pull request since there hasn't been any activity on it in a long time. Feel free to reopen it or send a new one if you're interested, though.

nico:
(fwiw i'd argue the codebase would've been in better shape if this had gotten merged as-is instead of not merging it at all. I don't remember which computer this patch was on.)

PR #4296, opened 2018-02-08
Add field name to error message in proto_writer

snowp:
This should make it easier to debug invalid formats when writing out
serialized proto json by hand.

Another option would be to have the StatusErrorListener add the name of the field when constructing the `util::Status` object, but I don't really have a sense of which option is preferable. 

Also open for suggestions as to how to format the error message


xfxyjwf:
Sorry for the late response. Feel free to reopen after resolving the merge conflicts.

PR #4299, opened 2018-02-08
Feat: add import-style=commonjs_strict option to the compiler

hekike:
The `protoc` output pollutes the global scope by attaching the messages to the global object per namespaces. It works fine while the process doesn't require multiple version of the generated output.

To avoid conflicts but keep backward compatibility I added the `import_style=commonjs_strict` mode that doesn't use the global scope.

**Related issues:**

- https://github.com/google/protobuf/issues/4271

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -276,7 +276,8 @@ string GetEnumPath(const GeneratorOptions& options,
 string MaybeCrossFileRef(const GeneratorOptions& options,
                          const FileDescriptor* from_file,
                          const Descriptor* to_message) {
-  if (options.import_style == GeneratorOptions::kImportCommonJs &&
+  if ((options.import_style == GeneratorOptions::kImportCommonJs ||
+       options.import_style == GeneratorOptions::kImportCommonJsStrict) &&
       from_file != to_message->file()) {
     // Cross-file ref in CommonJS needs to use the module alias instead of
     // the global name.
@@ -1675,8 +1676,18 @@ void Generator::GenerateProvides(const GeneratorOptions& options,
       //
       //   // Later generated code expects foo.bar = {} to exist:
       //   foo.bar.Baz = function() { /* ... */ }
-      printer->Print("goog.exportSymbol('$name$', null, global);\n", "name",
-                     *it);
+
+      // Do not use global scope in strict mode
+      if (options.import_style == GeneratorOptions::kImportCommonJsStrict) {
+        string namespaceObject = *it;
+        // Remove "proto." from the namespace object
+        namespaceObject.erase(0, 6);
+        printer->Print("goog.exportSymbol('$name$', null, proto);\n", "name",
+                namespaceObject);
+      } else {
+        printer->Print("goog.exportSymbol('$name$', null, global);\n", "name",
+                *it);
+      }
     }
   }
 }
@@ -3321,6 +3332,8 @@ bool GeneratorOptions::ParseFromOptions(
         import_style = kImportClosure;
       } else if (options[i].second == "commonjs") {
         import_style = kImportCommonJs;
+      } else if (options[i].second == "commonjs_strict") {
+        import_style = kImportCommonJsStrict;
       } else if (options[i].second == "browser") {
         import_style = kImportBrowser;
       } else if (options[i].second == "es6") {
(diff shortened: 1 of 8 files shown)


pherl:
OK to test.

hekike:
@haberman @pherl when do you plan to test it?

pherl:
Could you add a test to exercise the option?

hekike:
@pherl sure, test added!

hekike:
@pherl @haberman I fixed tests, all CI passes, could you please review it?

pherl:
Thanks for the change. I will wait for Josh or Bo to take a look as well.

anandolee:
Looks like it is ready to merge, can any reviewers or the author merge it?

hekike:
@anandolee we discovered an issue when there are imports in a proto file. 
I'll send a commit soon to this PR that fixes it.

hekike:
@anandolee fix commit published!

hekike:
@anandolee @TeBoring when do you plan to proceed with this?

xfxyjwf:
@haberman Josh, can you have a final look at this PR? It has new changes than when you originally approved.

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -1675,8 +1676,18 @@ void Generator::GenerateProvides(const GeneratorOptions& options,
       //
       //   // Later generated code expects foo.bar = {} to exist:
       //   foo.bar.Baz = function() { /* ... */ }
-      printer->Print("goog.exportSymbol('$name$', null, global);\n", "name",
-                     *it);
+
+      // Do not use global scope in strict mode
+      if (options.import_style == GeneratorOptions::kImportCommonJsStrict) {
+        string namespaceObject = *it;
+        // Remove "proto." from the namespace object
```
haberman: We should verify (GOOGLE_CHECK()?) that the string we're removing is actually "proto.".

hekike: Added

xfxyjwf:
@haberman Josh, can you take another look?

xfxyjwf:
Ping @haberman

src/google/protobuf/compiler/js/js_generator.cc
haberman: I don't see anything added? Once this is added, I approve the change to go in.

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -1675,8 +1676,19 @@ void Generator::GenerateProvides(const GeneratorOptions& options,
...
+        namespaceObject.erase(0, 6);
+        GOOGLE_CHECK(namespaceObject.find('proto.'));
```
hekike: @haberman now?

haberman: I'm confused, it looks like you are doing the check after the removal? If we are asserting that we are removing "proto.", shouldn't we check *before* the removal?

How is the JavaScript test passing in Travis? Is this code-path being tested?

hekike: I moved the check before removal, please check it out.

anandolee:
@haberman can you take another look?

src/google/protobuf/compiler/js/js_generator.cc
haberman: In the most recent version of the code, I still see:

```
+        namespaceObject.erase(0, 6);
+        GOOGLE_CHECK(namespaceObject.find('proto.'));
```

Am I missing something?

I am still worried that we are not exercising this code-path, since the JavaScript test is passing despite this code being incorrect.

xfxyjwf:
Ping @haberman

src/google/protobuf/compiler/js/js_generator.cc
hekike: @haberman how about now?

TeBoring:
@hekike This seems to cause test failure: https://source.cloud.google.com/results/invocations/83093f07-674e-470c-8a04-b16e686cbc44/targets/protobuf%2Fgithub%2Fmaster%2Fmacos%2Fjavascript%2Fpresubmit/log

hekike:
@TeBoring I see you opened https://github.com/google/protobuf/pull/4905 Does it mean that the CI fail is not related to this PR?

Issue #2128, opened 2016-09-16
Parser erroneously allows whitespace in package declaration.


JakeWharton:
We had a case where `protoc` allowed a `package` declaration that had newlines in it but other tooling based on the spec failed to parse the file. I believe this should have failed to parse inside `protoc` as well.

Relevant parts of the spec:

```
package = "package" fullIdent ";"
fullIdent = ident { "." ident }
ident = letter { letter | decimalDigit | "_" }
letter = "A" … "Z" | "a" … "z"
decimalDigit = "0" … "9"
```

Example showing `protoc` allowing newlines:

```
$ echo "package example.

foo

;

message Foo {}" > test.proto

$ protoc --version
libprotoc 3.0.0

$ protoc --java_out=. test.proto
[libprotobuf WARNING google/protobuf/compiler/parser.cc:547] No syntax specified for the proto file: test.proto. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defaulted to proto2 syntax.)

$ tree
.
├── example
│   └── foo
│       └── Test.java
└── test.proto
```

haberman:
Yes this seems like a bug.

xfxyjwf:
@anandolee It's the syntax spec that needs a fix here. A package name spreading across multiple lines is needed when the name is very long. For example:
```proto
message Foo {
  optional a.very.very.long.name.that.can.not.fit.into.one.line.without.exceeding
      .80.char.limit.Bar value = 1;
}
```

haberman:
@xfxyjwf Is the syntax spec checked in anywhere? What actually needs to be fixed here?

xfxyjwf:
The syntax spec here:
https://developers.google.com/protocol-buffers/docs/reference/proto2-spec

The spec needs to specify where spaces are allowed and where not.

fowles:
I suspect this is still a bug...

jhump:
There is definitely code in the wild that expects qualified names (dot-separated sequence of identifiers, such as in type names, extension names in message literals and custom options) to _allow_ whitespace, including newlines, in between name components. (Admittedly, this might be less likely in package declarations.) But either way, this would not be a backwards compatible change.

[...]

PR #4307, opened 2018-02-14
Revise package name parsing to reject whitespace

alasher:
Fixes issue #2128. Hope this hasn't been fixed internally already, I assumed it hadn't since the issue was last updated almost a year ago. Feel free to review @haberman.

There are a few ways to fix this, but I decided to leave this whitespace handling to the parser instead of the tokenizer. I tried adding support for fullIdent tokens to the tokenizer, but realized I'd need to add additional support for option names as well, and felt this was stepping too far into what the parser should be doing. The solution I ended on here allows the tokenizer to mark whether or not whitespace was present before it, and the parser can use that information as it wishes. Right now I've only added this support to package parsing, but this support can also be extended to option names (where this bug is also present.)

Also, I had to modify the ParsePackageWithSpaces unit test, which was actually testing to make sure this bug existed! I think we should instead be checking for failure on this case, so I modified the unit test to reflect that.

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -319,6 +319,21 @@ bool Parser::ConsumeEndOfDeclaration(
   }
 }
 
+bool Parser::HasNoPrecedingWhitespace(const char *error) {
+  if(input_ == NULL) {
+    return false;
+  }
+  if(input_->current().follows_whitespace) {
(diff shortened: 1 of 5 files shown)


xfxyjwf:
There is a bug but it's in the spec. We need to update the spec to reflect what protoc accepts. Not the other way around.

PR #4310, opened 2018-02-14
Support using MSVC intrinsics in Log2FloorNonZero

KindDragon:


src/google/protobuf/stubs/port.h
```diff
@@ -394,12 +395,10 @@ class Bits {
...
   return 31 ^ static_cast<uint32>(__builtin_clz(n));
-#elif defined(COMPILER_MSVC) && defined(_M_IX86)
-  _asm {
-    bsr ebx, n
-    mov n, ebx
-  }
-  return n;
+#elif defined(_MSC_VER)
+  unsigned long where;
+  _BitScanReverse(&where, n);
```
xfxyjwf: Is this more efficient than the original assembly?

KindDragon: No, it's just as effective but works also for 64bit build (inline assembler can only be used in 32-bit compiler `_M_IX86`)

KindDragon:
Hi, @xfxyjwf. Can you review this merge request?

KindDragon:
Thanks

PR #4312, opened 2018-02-15
Update JsonParser.cs

beppemarazzi:
other numeric types are allowed to be quoted, but boolean not...

### DIFF
--- a/csharp/src/Google.Protobuf/JsonParser.cs
+++ b/csharp/src/Google.Protobuf/JsonParser.cs
@@ -680,6 +680,12 @@ namespace Google.Protobuf
         {
             switch (field.FieldType)
             {
+                case FieldType.Bool:
+                    if (text.ToLower() == "true" || text == "1")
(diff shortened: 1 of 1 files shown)


beppemarazzi:
I signed it!

anandolee:
According to Proto3 JSON Specification, "true" and 1 is not accept for bool. This is to copy of the doc:

Boolean Value

A boolean value is converted to true or false in JSON. These two values along with null are the only values accepted by a conformat proto3 JSON parser for a boolean field.

For example:
message Foo {
  bool value = 1;
}

Valid JSON representation:
{
  “value”: true
}

The following should be rejected:

// Invalid: Not a boolean value.
{
  “value”: 1
}

// Invalid: Not a boolean value.
{
  “value”: “true”
}

anandolee:
I am closing this PR as it against  Proto3 JSON Specification. Feel free to reopen if you want to discuss with the spec

PR #4313, opened 2018-02-15
Testing public Kokoro logs

matt-kwong:
Verifying Kokoro builds are public


xfxyjwf:
Thanks Matt. I think the kokoro results are publicly visible now.

PR #4318, opened 2018-02-19
make Timestamp implement IComparable (fixes #4267)

### LINKED ISSUE #4267: [C#] WellKnownTypes.Timestamp does not implement IComparable
This is a runtime exception:

```csharp
IEnumerable<MyMessage> myMessages = GetSome();
var inChronologicalOrder = myMessages.OrderBy(m => m.MyTimestampField).ToArray(); // At least one object must implement IComparable
```

I know that dealing with time in code is wrought with perilous traps and so I'm wondering if there's a reason that it is ill-advised to make this comparable.  I couldn't think of one.  If you convert a Timestamp to a DateTime, UTC is specified for you, and if you try to get a Timestamp from a DateTime structure that doesn't specify itself as UTC, it fails.  So therefore we don't think Timestamp should be storing anything that is non-UTC.  But if we think that Timestamp is UTC, then it should be safe to compare it to other Timestamps, right?

warrenfalk:
I've implemented ```IComparable<Timestamp>``` for the ```Timestamp``` class.

I have also implemented comparison operators ```<, >, <=, >=, ==, !=```.

This fixes #4267 and is intended to make usage much more natural:

```c#
if (thing.Happened > setting.Deadline)
{
    // This is late
}

var inChronologicalOrder = listOfEvents.OrderBy(e => e.WhenItHappened);
```


warrenfalk:
Oh yeah, forgot about this.  Yep give me a few minutes.

warrenfalk:
Thanks for the ping.  This has been updated to handle non-normalized Timestamps.

csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs
```diff
@@ -222,6 +222,93 @@ internal static string ToJson(long seconds, int nanoseconds, bool diagnosticOnly
...
+
+        /// <summary>
+        /// Returns whether two timestamps are equivalent
+        /// </summary>
+        /// <param name="a"></param>
+        /// <param name="b"></param>
+        /// <returns>true if the two timestamps refer to the same nanosecond</returns>
+        public static bool operator ==(Timestamp a, Timestamp b)
+        {
+            return ReferenceEquals(a, b) || (a is null ? (b is null ? true : false) : a.Equals(b));
```
anandolee: For a.Equals(b): Maybe I missed somewhere but didn't see Equals() implementation in this file. Shouldn't use a.CompareTo(b)==0 instead?   Maybe add a test to compare two equal timestamps which are not normalized

csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs
```diff
@@ -222,6 +222,93 @@ internal static string ToJson(long seconds, int nanoseconds, bool diagnosticOnly
...
+        /// <summary>
+        /// Given another timestamp, returns 0 if the timestamps are equivalent, -1 if this timestamp precedes the other, and 1 otherwise
+        /// </summary>
+        /// <param name="other">Timestamp to compare</param>
+        /// <returns>an integer indicating whether this timestamp precedes or follows the other</returns>
+        public int CompareTo(Timestamp other)
+        {
+            if (!IsNormalized(Seconds, Nanos))
+                return Normalize(Seconds, Nanos).CompareTo(other);
+            if (other == null)
```
anandolee: Better put null check in the first place (before normalize this timestamp). 
Why compare to null should return 1?  Is any other well known compare do like this that consider null as the smallest?

csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs
```diff
@@ -222,6 +222,93 @@ internal static string ToJson(long seconds, int nanoseconds, bool diagnosticOnly
             }
         }
 
+        /// <summary>
+        /// Given another timestamp, returns 0 if the timestamps are equivalent, -1 if this timestamp precedes the other, and 1 otherwise
+        /// </summary>
+        /// <param name="other">Timestamp to compare</param>
+        /// <returns>an integer indicating whether this timestamp precedes or follows the other</returns>
+        public int CompareTo(Timestamp other)
```
anandolee: Does CompareTo needs to be public? I'd prefer to make it private if possible and only expose < == operations etc. to users

csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs
warrenfalk: OK.  I think normalization at this level is a mistake and recommend reconsidering removing it.  Let me answer your questions and explain my reasoning, and you can decide what you want to do.

```Equals()``` is already implemented in the other, generated half of the ```partial``` class.  I can't change the implementation there, and I don't think we want to because ultimately ```Timestamp``` is a protocol buffer structure and it should follow protocol buffer semantics where "equals" means "structurally all fields are equal".

So if a user wants "time equality" instead of just "protocol buffer" equality then the user must (already, today) ensure that her structures are all normalized.

Keep in mind that equality and comparison are semantically separate things, which is why ```IEquatable``` and ```IComparable``` are separate interfaces.  "Comparison" deals with ordering/sorting.  So ```CompareTo(a) == 0``` does not mean the same thing as ```Equals(a) == true``` and they do not necessarily have to match.

Additionally the operators don't match each other either.  The ```< > <= >=``` operators are not necessarily related to ```==```.  So it cannot be asserted that ```!(a < b) && !(a > b) == (a == b)```.

But of course people expect them to match, so I think we should make them match.

If you want them to match, however, you have to give up on normalization because Equals ignores it and can't be changed.  So if you want "time comparison" then you have to pre-normalize, just like you have to already pre-normalize if you want "time equality".

So the bottom line is you can either have comparisons that normalize, or comparisons that match equality, but not both.  You have to choose.

File: csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs (hunk 2 above)
Comment (warrenfalk) [reply csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs]: The BCL treats null this way.  Consider: ```Debug.Assert(1 == "bcl".CompareTo(null));```

See my other comment on normalization and how I think we should reconsider not doing it, but if we keep it, here are my thoughts here:

Doing a null check first will do the null check twice.  Since the only Normalize() function we have creates an entirely new Timestamp instance, we just call that one's CompareTo() which will be doing its own null check.

I would understand if you wanted to do it anyway, considering un-normalized to be the exceptional case, and normalization to be the more expensive operation.  So let me know.  But I recommend we just not normalize in the first place.

File: csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs (hunk 3 above)
Comment (warrenfalk) [reply csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs]: Well, we are implementing ```IComparable```, so does it have to be public?  It does have to be exposed, yes.  We could make the implementation explicit, but it will always be public on the interface and therefore always exposed to users and so there's no way around exposing it.  Note that for reference, these are always public in the BCL, as far as I can tell.

File: csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs (hunk 1 above)
Comment (anandolee) [reply csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs]: Agree, we should make them match. I talked with our tech lead, we think it make sense to not normalize for all comparison. Add comments that if the timestamp is not normalized the result may not correct in the API documentation

File: csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs [hunk 4]
```diff
@@ -222,6 +222,109 @@ internal static string ToJson(long seconds, int nanoseconds, bool diagnosticOnly
             }
         }
 
+        /// <summary>
+        /// Given another timestamp, returns 0 if the timestamps are equivalent, -1 if this timestamp precedes the other, and 1 otherwise
+        /// </summary>
+        /// <remarks>
+        /// The result of comparing non-normalized timestamps is not specified and may give unexpected results
```
anandolee: "Make sure the timestamps are normalized.  Comparing non-normalized timestamps is not specified and may give unexpected results."

PR #4324, opened 2018-02-21
 Added instruction for existing ZLIB configuration

abdul-sami:
Using two configuration flags `DZLIB_INCLUDE_DIR` and `DZLIB_LIB` we can locate an existing ZLIB on the system


abdul-sami:
I signed it!

PR #4329, opened 2018-02-22
made RepeatedField<T> throw InvalidOperationException when collection…

andreabalducci:
… changes during iteration

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
@@ -304,6 +304,19 @@ namespace Google.Protobuf.Collections
         }
 
         [Test]
+        public void Throw_InvalidOperation_Exception_if_modified()
+        {
+            var list = new RepeatedField<string> { "first", "second" };
+            using (var enumerator = list.GetEnumerator())
+            {
+                list.RemoveAt(0);
+                var ex = Assert.Throws<InvalidOperationException>(() => { enumerator.MoveNext(); });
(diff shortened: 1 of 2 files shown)


andreabalducci:
I signed it!

xfxyjwf:
@anandolee

csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
```diff
@@ -303,6 +303,19 @@ public void Enumerator()
             }
         }
 
+        [Test]
+        public void Throw_InvalidOperation_Exception_if_modified()
+        {
+            var list = new RepeatedField<string> { "first", "second" };
+            using (var enumerator = list.GetEnumerator())
+            {
+                list.RemoveAt(0);
```
anandolee: What if remove one item and then add one item back during enumeration operation?

andreabalducci: it's possible but unlikely. I just covered the most common scenario to limit the footprint of the patch. 

A better approach would be to clone the internal array and enumerate the snapshot but will impact memory usage. 80% of the time you'll enumerate to consume repeated field.

An even safer approach would be adding checks on array mutations while an enumeration is in progress.

anandolee: Does "add checks on array mutations while enumeration is in progress" cost lot? If no can you change to this way?

anandolee:
Friendly ping, andreabalducci@ are you still working on this PR ?

anandolee:
I am closing it for no response about 4 months. Feel free to reopen if still like to go with the PR

PR #4342, opened 2018-02-28
source code info for interpreted options; fix source code info for extension range options

### LINKED ISSUE #3819: protoc produces SourceCodeInfo that references uninterpreted options, which are not present
When protoc produces `SourceCodeInfo` for a file descriptor, all details related to options are attributed (according to the `path` field) to uninterpreted options. There is no way, however, for client code to actually map uninterpreted options to the actual options. During linking, protoc updates the options message and then *discards* the uninterpreted versions. So the path refers to absent elements of the descriptor.

This makes it effectively impossible to extract, for example, comments associated with options. Another task that *seems* like it should be possible is to reconstruct a proto source file from a descriptor that includes source code info (with the understanding that certain things, like trailing comments, cannot be preserved in the descriptor). But the way options are handled in source code info further reduces the fidelity that is possible.

Possible approaches to solve:

1. During linking, the source code info could be updated so that the info for options is actually attributed to the options (e.g. references proper option field numbers, or extension numbers for custom options). For options that define values nested deep inside of options that are messages, the path could contain additional elements that pinpoint exactly the field that is being set. This would lose fidelity since the current representation allows attribution of source code location and comments to option names (even name parts, when the option type is itself a message) whereas this could not be preserved with the suggested transformation.

2. During linking, preserve uninterpreted options in the descriptor when source code info is being included. Furthermore, add fields to the uninterpreted option (or maybe just the name part message) that provide the resolved field/extension numbers. This allows client code to discover the link between uninterpreted option source code info and the actual option values in the descriptor.

jhump:
Fixes #3819 

I broke this up into multiple commits to make the changes easier to follow:
1. Compute paths (in the `SourceCodeInfo.Location` meaning of path) to options when building the file descriptor.
2. Use the computed paths from step 1 to construct a map from "path to uninterpreted option" to "path to interpreted option" during the option interpretation step.
3. After options are interpreted, apply the mapping constructed in step 2 to modify the file's source code info.

While testing this (with my own Go test harness that I also used to test my [protoparse](https://godoc.org/github.com/jhump/protoreflect/desc/protoparse) package), I discovered that protoc was also constructing the paths to extension range options incorrectly. The paths that ended up in the source code info were missing the extension range index. So I added a fourth commit so that things looked good:

4. Fix the way paths to extension range options are generated.

I have not yet added a test in this repo. I am a little uncertain of *how* I should test it. FWIW, in my custom Go test harness, I simply process a proto source file that has lots of options (normal and custom options) across various descriptor types (including on extension ranges). My harness can print a human-readable interpretation of the source code info which I visually inspect. I can then commit the results as *golden test files* and have a unit test just process the same source file and verifies that the output matches the "blessed" golden result.

I have looked through some of the existing test code, expecting/hoping to find similar tests (that match outputs against "golden test files") but couldn't quite find any. Since my changes impact both parsing (for the extension range options fix) and descriptors (for interpreting options and re-writing source code info), I think I probably need two tests: (1) a test in `compiler/parser_unittest.cc` that parses source with extension range options and verifies the resulting source code info; and (2) a test in `descriptor_unittest.cc` that builds a descriptor using a proto with uninterpreted options and corresponding source code info and then verifies that the source code info is modified correctly.

If someone can advise, maybe provide pointers to similar kinds of tests, I would greatly appreciate.

---
This change is [Reviewable](https://reviewable.io/reviews/google/protobuf/4342)

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -1521,15 +1521,13 @@ bool Parser::ParseExtensions(DescriptorProto* message,
         message->mutable_extension_range(old_range_size)->mutable_options();
 
     {
-      DO(Consume("["));
-
       LocationRecorder index_location(
-        extensions_location, 0 /* we fill this in w/ actual index below */,
-        &info);
-
+          extensions_location, 0 /* we fill this in w/ actual index below */,
+          &info);
       LocationRecorder location(
           index_location,
           DescriptorProto::ExtensionRange::kOptionsFieldNumber);
+      DO(Consume("["));
 
       do {
         DO(ParseOption(options, location, containing_file, OPTION_ASSIGNMENT));
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -337,31 +337,42 @@ void Parser::AddError(const string& error) {
 
 Parser::LocationRecorder::LocationRecorder(Parser* parser)
   : parser_(parser),
+    source_code_info_(parser->source_code_info_),
     location_(parser_->source_code_info_->add_location()) {
   location_->add_span(parser_->input_->current().line);
   location_->add_span(parser_->input_->current().column);
 }
 
 Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent) {
-  Init(parent);
+  Init(parent, parent.source_code_info_);
+}
+
+Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent,
+                                           int path1,
+                                           SourceCodeInfo* source_code_info) {
+  Init(parent, source_code_info);
+  AddPath(path1);
 }
 
 Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent,
                                            int path1) {
-  Init(parent);
+  Init(parent, parent.source_code_info_);
   AddPath(path1);
 }
 
 Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent,
                                            int path1, int path2) {
-  Init(parent);
+  Init(parent, parent.source_code_info_);
   AddPath(path1);
   AddPath(path2);
 }
 
-void Parser::LocationRecorder::Init(const LocationRecorder& parent) {
+void Parser::LocationRecorder::Init(const LocationRecorder& parent,
+                                    SourceCodeInfo* source_code_info) {
   parser_ = parent.parser_;
-  location_ = parser_->source_code_info_->add_location();
+  source_code_info_ = source_code_info;
+
+  location_ = source_code_info_->add_location();
   location_->mutable_path()->CopyFrom(parent.location_->path());
 
   location_->add_span(parser_->input_->current().line);
@@ -401,6 +412,10 @@ void Parser::LocationRecorder::RecordLegacyLocation(const Message* descriptor,
     parser_->source_location_table_->Add(
         descriptor, location, location_->span(0), location_->span(1));
   }
+}
+
+int Parser::LocationRecorder::CurrentPathSize() const {
+  return location_->path_size();
 }
 
 void Parser::LocationRecorder::AttachComments(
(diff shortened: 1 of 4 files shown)


jhump:
Ping, @xfxyjwf

jhump:
Ping @xfxyjwf 
Mind taking a look? Feedback appreciated.

xfxyjwf:
@jhump Sorry for delay. I probably won't have time to review this change until late this week though.

jhump:
> I probably won't have time to review this change until late this week though.

@xfxyjwf, no problem. I'm just trying to keep it from totally falling off the radar :)

jhump:
Ping?

xfxyjwf:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/google/protobuf/descriptor.cc, line 4158 at r1](https://reviewable.io:443/reviews/google/protobuf/4342#-L7CM0nymBfXMLjbabX0:-L7CM0nymBfXMLjbabX1:bmlxleg) ([raw file](https://github.com/google/protobuf/blob/3174b1e8d432690ea92480712fd77e0df0edc74d/src/google/protobuf/descriptor.cc#L4158)):*
> ```cpp
>     options_to_interpret_.push_back(
>         OptionsToInterpret(name_scope, element_name, element_path_, &orig_options, options));
>   }
> ```

Can we instead use DescriptorT::GetLocationPath() to reconstruct the path?
https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.h#L440

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/protobuf/4342)*

xfxyjwf:
I'm using reviewable for this PR as it's more convenient for large changes. Let me know if you have issue viewing the comment.

jhump:
*[src/google/protobuf/descriptor.cc, line 4158 at r1](https://reviewable.io:443/reviews/google/protobuf/4342#-L7CM0nymBfXMLjbabX0:-L7Gm_dDkZPyi7RhQ4fy:b18qn8i) ([raw file](https://github.com/google/protobuf/blob/3174b1e8d432690ea92480712fd77e0df0edc74d/src/google/protobuf/descriptor.cc#L4158)):*
<details><summary><i>Previously, xfxyjwf (Feng Xiao) wrote…</i></summary><blockquote>

Can we instead use DescriptorT::GetLocationPath() to reconstruct the path?
https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.h#L440

</blockquote></details>

I revised the code to do that. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/protobuf/4342)*

xfxyjwf:
Thanks. The code overall looks good to me. Could you add some test cases for these new locations in descriptor_unittest.cc?
https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor_unittest.cc#L6948

You probably need to use FileDescriptor::GetSourceLocation(path, location) with manually constructed paths in the test.

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/google/protobuf/descriptor.cc, line 3407 at r2](https://reviewable.io:443/reviews/google/protobuf/4342#-L7QfT8sUGKEb_WE8dCe:-L7QfT8sUGKEb_WE8dCf:bxpmv2b) ([raw file](https://github.com/google/protobuf/blob/11cb912f4cbd054b3a8e95710d919e077bbf1e1b/src/google/protobuf/descriptor.cc#L3407)):*
> ```cpp
>     // must be the mutable copy of the original options message to which
>     // uninterpreted_option_ belongs.
>     bool InterpretSingleOption(Message* options, std::vector<int>& src_path,
> ```

Please add some comments to explain what src_path/options_path are.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/protobuf/4342#-:-L7QktMYRc0KVyIjgSeP:b-79zoc6)*

xfxyjwf:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/google/protobuf/descriptor_unittest.cc, line 7088 at r3](https://reviewable.io:443/reviews/google/protobuf/4342#-L8F-TK4lPcpUiDWUPWr:-L8F-TK4lPcpUiDWUPWs:b-r40wwd) ([raw file](https://github.com/google/protobuf/blob/9427fdd95ec11178f4b236fa5e6e46224d5c6653/src/google/protobuf/descriptor_unittest.cc#L7088)):*
> ```cpp
>                    0};
> 
>     std::vector<int> vpath(path, path+2);
> ```

minor style issue: "+" should be surrounded with spaces.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/protobuf/4342)*

jhump:
@xfxyjwf I squashed into one commit and rewrote message.

However, you can actually do this yourself when you merge. In case you didn't know: the button in Github has a drop-down next to it that allows for three options (and you can actually change repo settings to not allow all of them):
1. Create a merge commit
2. Squash and merge
3. Rebase and merge
If you do "squash and merge", it gives you a chance to edit the commit message for the squashed commit that will be made to master. And since this branch is in my own fork, not your repo, you don't have to worry about the state of history on my PR branch.

That way, you don't need contributors to do any sort of history clean-up in their forks before accepting their contributions.

xfxyjwf:
@jhump I usually just click the "Merge pull request" button and haven't used the other options before. Thanks for explaining it to me. I'll try that next time.

jhump:
@xfxyjwf, I don't see any possible link between those MacOS Ruby build issues and this PR. So hopefully this branch is good to go?

jhump:
Whoops. I realized I didn't address your last style nit. And I also find a small bug in my change to `parser.cc` that misattributes the location of options for extension ranges. Another commit incoming soon.

jhump:
@xfxyjwf, I think this is ready to go now.

jhump:
Thanks for reviewing and merging @xfxyjwf!!

PR #4347, opened 2018-03-02
Revert "Move `compiler/plugin.pb.cc` to libprotobuf with the other WKT sources."

xfxyjwf:
This reverts commit 724f0be33d0eb42055ebe81ce0c8c7bf88ea666c, which is part of PR https://github.com/google/protobuf/pull/4334 to fix a bazel build issue.

The commit in question moves plugin.pb.cc from libprotoc to libprotobuf. However, because libprotoc uses different dllimport/export macros from libprotobuf, the move causes appveyor tests to fail:
https://ci.appveyor.com/project/protobuf/protobuf/build/1.0.5732/job/h436akiwyf78f3b5


xfxyjwf:
Note after reverting that commit, plugin_proto can no longer be depended by other proto_library targets. I suspect it won't be much of a problem because plugin_proto is rarely used compared to other protos like descriptor_proto. Given that Josh is working on cleaning up the bootstrap stuff internally, I think we can just wait util the internal cleanup is done and implement it the same way in opensource.

jmillikin:
Would it be practical to replace `LIBPROTOC_EXPORT` with `LIBPROTOBUF_EXPORT` in `src/google/protobuf/compiler/plugin.pb.h`, instead of making compiler plugins risk an ODR violation (and undefined behavior)?

Two other headers in the `compiler/` directory (`importer.h` and `parser.h`) already use the non-protoc exports.

xfxyjwf:
@jmillikin I think the question is more of "whether plugin.proto belongs to libprotobuf instead of libprotoc". The more I think about it the less convinced I am about this move. Logically plugin.proto should be part of libprotoc because it's used for protoc plugins and used only in protoc but not the protobuf runtime. Internally in Google, plugin.proto is part of libprotoc and I don't see an easy to move it to libprotobuf there. I

jmillikin:
> Logically plugin.proto should be part of libprotoc because it's used for protoc plugins and used only in protoc but not the protobuf runtime.

Protoc plugins dont need libprotoc, though. They're documented as interacting via stdin/stdout, and no libraries other than `libprotobuf` are required to parse/serialize the `plugin.proto` messages.

> Internally in Google, plugin.proto is part of libprotoc and I don't see an easy to move it to libprotobuf there.

Is it possible to have the Google-internal and the OSS BUILD files drift temporarily until the bootstrap work you mentioned earlier lands?

xfxyjwf:
It's an API breaking change if we remove plugin.proto from libprotobuf after adding it there. If we don't plan to move plugin.proto into libprotobuf, we shouldn't do it as a temporary workaround.

From what I can tell, you can just depend on libprotoc if you want to use plugin.proto. For languages other than C++, depending on plugin_proto should just works. It's not clear to me what use case requires plugin.proto to be part of libprotobuf.

PR #4348, opened 2018-03-03
Use LIBPROTOBUF_EXPORT for DLL import/export macros in plugin.pb.h

jmillikin:
This is an alternative to the partial revert proposed in <https://github.com/google/protobuf/pull/4347>. See <https://github.com/google/protobuf/pull/4334> for additional context.

Note: please make sure all CI builds (including AppVeyor) are green before merging.


xfxyjwf:
Thanks for the suggestion but we will use a different approach instead (created https://github.com/google/protobuf/issues/4402).

PR #4354, opened 2018-03-05
For encoding upb needs descriptor type instead of type.

TeBoring:


TeBoring:
The broken test is not due to this change.

PR #4360, opened 2018-03-06
Conversion from uint64 to uint32 didn't consider sign bit.

TeBoring:


haberman:
Already merged in another sync.

PR #4361, opened 2018-03-07
Add go benchmark

BSBandme:
Reorganize internal one to open source.


BSBandme:
How can I add @dsnet to reviewer?

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	data.count = len(data.Unmarshalled)
+
+	os.Exit(m.Run())
+}
+
+func BenchmarkUnmarshal(b *testing.B) {
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		payload := data.Marshalled[counter % data.count]
```
BSBandme: I changed to parse one message for one rep to accommodate other languages. 
Is there any way to make the benchmark testing time longer?

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
+package main
+
+import (
+	"errors"
```
xfxyjwf: Shouldn't it be intended by 2 spaces?

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+type Data struct {
+	// Marshalled is a slice of marshalled protocol
+	// buffers. 1:1 with Unmarshalled.
+	Marshalled [][]byte
+
+	// Unmarshalled is a slice of unmarshalled protocol
+	// buffers. 1:1 with Marshalled.
+	Unmarshalled []proto.Message
+
+	count int
```
xfxyjwf: Why is the other two using UpperCamelCase and this one using lower case?

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+func TestMain(m *testing.M) {
+	flag.Parse()
+	data = new(Data)
+	rawData, error := ioutil.ReadFile(flag.Arg(0))
+	if error != nil {
+		panic("Couldn't find file" + flag.Arg(0))
+	}
+	var dataset benchmarkWrapper.BenchmarkDataset
+
+	if err1 := proto.Unmarshal(rawData, &dataset); err1 != nil {
```
xfxyjwf: Can you use the same variable name "err" throughout?

benchmarks/go_benchmark_test.go
dsnet: This seems problematic. Go benchmarking tries to dynamically determine the optimal b.N to use in order to get a stable measurement of performance.

However, this approach breaks that assumption when N is less than `data.count` since the distribution of test messages is almost certainly not uniform in cost.

For that reason, the original benchmarks ran through all messages in the testdata for every iteration in N.

benchmarks/go_benchmark_test.go
BSBandme: This is just tab, which seems appears to be 6 spaces on github.
I think this is the same as [other go file](https://github.com/golang/protobuf/blob/master/descriptor/descriptor.go)

benchmarks/go_benchmark_test.go
BSBandme: Done by changed the first two to lower case for not exporting outside package

benchmarks/go_benchmark_test.go
BSBandme: Done.

benchmarks/README.md
```diff
@@ -36,6 +36,18 @@ $ sudo apt-get install python3-dev
 ```
 And you also need to make sure `pkg-config` is installed.
 
+### Go
+Golang protobuf is under [golang repo](https://github.com/golang/protobuf), you
+need to install golang and golang protobuf's proto and protoc-gen-go before
+running the benchmark. And we're using [testing package](https://golang.org/pkg/testing/)
+to test benchmark.
```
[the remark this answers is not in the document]
BSBandme: Suggestion applied

benchmarks/README.md
```diff
@@ -36,6 +36,18 @@ $ sudo apt-get install python3-dev
...
+Golang protobuf is under [golang repo](https://github.com/golang/protobuf), you
+need to install golang and golang protobuf's proto and protoc-gen-go before
+running the benchmark. And we're using [testing package](https://golang.org/pkg/testing/)
+to test benchmark. 
+
+To install golang protobuf proto and protoc-gen-go, you can 
+```
+$ go get -u github.com/golang/protobuf/protoc-gen-go
+$ go get -u github.com/golang/protobuf/proto
+```
```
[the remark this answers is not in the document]
BSBandme: Suggestion applied.

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"os"
+
+	benchmarkWrapper "./tmp"
+	proto "github.com/golang/protobuf/proto"
+	googleMessage1Proto3 "./tmp/datasets/google_message1/proto3"
+	googleMessage1Proto2 "./tmp/datasets/google_message1/proto2"
+	googleMessage2 "./tmp/datasets/google_message2"
+	googleMessage3 "./tmp/datasets/google_message3"
+	googleMessage4 "./tmp/datasets/google_message4"
+
```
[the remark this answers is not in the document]
BSBandme: Done.

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// Data is returned by the Load function.
+type Data struct {
+	// Marshalled is a slice of marshalled protocol
```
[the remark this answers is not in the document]
BSBandme: Done.

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// buffers. 1:1 with Unmarshalled.
+	Marshalled [][]byte
+
+	// Unmarshalled is a slice of unmarshalled protocol
```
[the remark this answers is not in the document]
BSBandme: Done.

benchmarks/go_benchmark_test.go
BSBandme: @xiaofeng, would you please explain more about this? In fact I'm also confusing why decode one message for one loop is better.

xfxyjwf: Let's just change it to parse the entire dataset for each iteration. BTW, my github user name is xfxyjwf, not xiaofeng.

benchmarks/go_benchmark_test.go
```diff
@@ -0,0 +1,133 @@
+package main
```
dsnet: This benchmark can be more simply written as the following:
```go
package main

type Dataset struct {
	name        string
	newMessage  func() proto.Message
	marshaled   [][]byte
	unmarshaled []proto.Message
}

var datasets []Dataset

func init() {
	for _, f := range flag.Args() {
		// Load the benchmark.
		b, err := ioutil.ReadFile(f)
		if err != nil {
			panic(err)
		}

		// Parse the benchmark.
		var dm benchmarkWrapper.BenchmarkDataset
		if err := proto.Unmarshal(b, &dm); err != nil {
			panic(err)
		}

		// Determine the concrete protobuf message type to use.
		var ds Dataset
		switch dm.MessageName {
		case "benchmarks.proto3.GoogleMessage1":
			ds.newMessage = func() proto.Message { return new(googleMessage1Proto3.GoogleMessage1) }
		case "benchmarks.proto2.GoogleMessage1":
			ds.newMessage = func() proto.Message { return new(googleMessage1Proto2.GoogleMessage1) }
		case "benchmarks.proto2.GoogleMessage2":
			ds.newMessage = func() proto.Message { return new(googleMessage2.GoogleMessage2) }
		case "benchmarks.google_message3.GoogleMessage3":
			ds.newMessage = func() proto.Message { return new(googleMessage3.GoogleMessage3) }
		case "benchmarks.google_message4.GoogleMessage4":
			ds.newMessage = func() proto.Message { return new(googleMessage4.GoogleMessage4) }
		default:
			panic("unknown message type: " + dm.MessageName)
		}

		// Unmarshal each test message.
		for _, payload := range dm.Payload {
			ds.marshaled = append(ds.marshaled, payload)
			m := ds.newMessage()
			if err := proto.Unmarshal(payload, m); err != nil {
				panic(err)	
			}
			ds.unmarshaled = append(ds.unmarshaled, m)
		}

		datasets = append(datasets, ds)
	}
}

func Benchmark(b *testing.B) {
	for _, ds := range datasets {
		b.Run(ds.name, func(b *testing.B) {
			b.Run("Unmarshal", func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					// ...
				}
			})
			b.Run("Marshal", func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					// ...
				}
			})
			b.Run("Size", func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					// ...
				}
			})
			b.Run("Clone", func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					// ...
				}
			})
			b.Run("Merge", func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					// ...
				}
			})
		})
	}
}
```

BSBandme: Thank you very much for this suggestion! This is much cleaner now.

Before I tried to use testing.M.run for several times, but it seems that it shouldn't be, and I didn't realize testing.B.run can run for several times.

I keep most of your code, except I still keep the function for generating the newMessage function as we may add some generator for generating benchmark code in the future so that it can use user defined schema. To write it in an function is easier for generator to change this part of code.

And I still left to deal with one message in one loop currently and I'll discuss this with Feng first.

BSBandme:
Changed to iterate over all messages in one loop which is the same as internal version

dsnet:
LGTM.

PR #3248, opened 2017-06-20
Use mb_str* functions instead str*

DenKoren:
This makes protobuf work with overloaded str* functions (when mbstring.func_overload PHP param is greater than 2)

PR #4362, opened 2018-03-07
Make protobuf work with mbstring.func_overload >= 2

DenKoren:
This is the same fix, as in #3248, but for modern version of protobuf (#3248 had conflicts with master and thus was frozen)

It makes protobuf work with overloaded str* functions (when mbstring.func_overload PHP param is greater than 2).

### DIFF
--- a/php/src/Google/Protobuf/Internal/CodedInputStream.php
+++ b/php/src/Google/Protobuf/Internal/CodedInputStream.php
@@ -36,7 +36,6 @@ use Google\Protobuf\Internal\Uint64;
 
 class CodedInputStream
 {
-
     private $buffer;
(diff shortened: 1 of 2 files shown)


DenKoren:
AppVeyor failed on CPP stage. Patch affects only clear PHP code and should not be the cause of failure.

TeBoring:
Can we pass mbstring.func_overload on php command line to make existing tests able to test the case?

DenKoren:
> Can we pass mbstring.func_overload on php command line to make existing tests able to test the case?

Sorry, was this action expected from me, @TeBoring ? Or this changes failed to pass the tests?

Issue #2911, opened 2017-03-28
Pylint segfault while trying to check generated file


i-feofilaktov:
Versions:
Protobuf:
```
$ sudo pip freeze | grep proto
protobuf==3.2.0
```
Protoc:
```
$ sudo /usr/local/bin/protoc --version
libprotoc 3.1.0
```
Pylint
```
$ sudo pip freeze | grep pylint
pylint==1.6.5
```
Trying to check generated file from helloworld grpc example: https://github.com/grpc/grpc/blob/master/examples/python/helloworld/helloworld_pb2.py
Here's gdb backtrace:
```
$ gdb python /tmp/cores/core.pylint.1763.centos-runtime.1490694728
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-94.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /usr/bin/python3.4...Reading symbols from /usr/bin/python3.4...(no debugging symbols found)...done.
(no debugging symbols found)...done.

warning: core file may not match specified executable file.
[New LWP 1763]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/usr/bin/python3.4 /usr/bin/pylint examples/python/helloworld/helloworld_pb2.py'.
Program terminated with signal 11, Segmentation fault.
#0  google::protobuf::python::message_meta::GetExtensionsByName (self=0x7fc525783ea0 <google::protobuf::python::CMessage_Type>, closure=<optimized out>) at google/protobuf/pyext/message.cc:351
351	google/protobuf/pyext/message.cc: No such file or directory.
[...]

PR #4365, opened 2018-03-08
message_meta: fix segfault in GetMessageDescriptor

zmedico:
This fixes a segmentation fault triggered by the following command, see the gdb backtrace in issue #2911:
    
`python -c 'from google.protobuf import duration_pb2; help(duration_pb2.Duration)'`

Orivej Desh [writes](https://github.com/google/protobuf/issues/2974#issuecomment-354191487):

    The crash in accessing
    google.protobuf.pyext._message.Message._extensions_by_name happens
    because metaclass getters [1] shadow class getters [2] and the intended
    dispatch from the latter to the former [3] does not run. This can be
    fixed by replacing message_meta::Getters with 0 at [1].

[1] https://github.com/google/protobuf/blob/099d99759101c295244c24d8954ec85b8ac65ce3/python/google/protobuf/pyext/message.cc#L443
[2] https://github.com/google/protobuf/blob/099d99759101c295244c24d8954ec85b8ac65ce3/python/google/protobuf/pyext/message.cc#L2849
[3] https://github.com/google/protobuf/blob/099d99759101c295244c24d8954ec85b8ac65ce3/python/google/protobuf/pyext/message.cc#L2615-L2616

See: https://github.com/google/protobuf/issues/2974#issuecomment-354191487
Fixes: https://github.com/google/protobuf/issues/2974

### DIFF
--- a/python/google/protobuf/pyext/message.cc
+++ b/python/google/protobuf/pyext/message.cc
@@ -402,12 +402,6 @@ static PyObject* GetExtensionsByNumber(CMessageClass *self, void *closure) {
   return result.release();
 }
 
-static PyGetSetDef Getters[] = {
-  {"_extensions_by_name", (getter)GetExtensionsByName, NULL},
-  {"_extensions_by_number", (getter)GetExtensionsByNumber, NULL},
-  {NULL}
-};
-
 }  // namespace message_meta
 
 PyTypeObject CMessageClass_Type = {
@@ -440,7 +434,7 @@ PyTypeObject CMessageClass_Type = {
   0,                                   // tp_iternext
   0,                                   // tp_methods
   0,                                   // tp_members
-  message_meta::Getters,               // tp_getset
+  0,                                   // tp_getset
   0,                                   // tp_base
   0,                                   // tp_dict
   0,                                   // tp_descr_get


zmedico:
I signed it!

orivej:
Thanks for making the PR out of that patch! You should probably change the target branch from `3.5.1.1` to `3.5.x`. (Or `master`, I'm not sure.)

anandolee:
Do you mean that message_meta::Getters is set, google.protobuf.pyext._message.Message._extensions_by_name is calling message_meta::Getters instead of cmessage::Getters?

zmedico:
> How about move reinterpret_cast to message_meta::GetExtensionsByName?

@anandolee, I've tried something like that but the `help()` test cases posted in https://github.com/google/protobuf/issues/2974 still segfault. Can you create a patch to show me what you mean?

> One more question:
> Is google.protobuf.pyext._message.Message._extensions_by_name calling message_meta::Getters instead of cmessage::Getters?

Yes, that's how I interpret the gdb backtrace posted in https://github.com/google/protobuf/issues/2911.

zmedico:
@amauryfa, PTAL

anandolee:
What I say may not work. 

This change looks good to me. But as it is removing an 'API', I will do a test inside google to make sure it does not break any internal usages. Will approval this change if everything goes well. The test takes 1-2 business days.

anandolee:
Sorry, I just noticed that this segfault is already fixed in internal by @amauryfa when doing internal test.
Expect to see the fix in next release. I will close this for clean up, feel free to reopen it if it is still an issue after next release.

manuel-koch:
I see the same problem on python 3.4.7 using latest protobuf 3.6.0 !?

I.e. trying the snippet from above
```
Python 3.4.7 (default, Dec 12 2017, 06:47:53)
[GCC 4.9.2] on linux
>>> import google.protobuf
>>> google.protobuf.__version__
'3.6.0'
>>> import google.protobuf.pyext._message
>>> google.protobuf.pyext._message.Message._extensions_by_name
Segmentation fault
```

PR #4366, opened 2018-03-08
Fix bug: whether always_print_enums_as_ints is true or false, it always print the default value of enums as strings

stone4774:
description of the problem:

call the function MessageToJsonString, and set both always_print_primitive_fields and always_print_enums_as_ints to true.  it print the default value of enums as strings,  does not as ints.

This submission resolved the issue


stone4774:
I signed it!

stone4774:
try again

xfxyjwf:
@stone4774 It seems the commits you made are not associated with your github account. Could you update these commits with the correct email address?

stone4774:
I submited a new PR:
https://github.com/google/protobuf/pull/4395

so turn off the present

PR #4368, opened 2018-03-08
add a setter to allow the support of UnknownFields in Text Mode

Thomas-Barbier-1A:
This is complementing another PR which is not yet merged in master https://github.com/google/protobuf/pull/4145


xfxyjwf:
Closing as we will not support this in TextFormat. See https://github.com/google/protobuf/pull/4145#issuecomment-375054364

PR #4370, opened 2018-03-08
Objective-C: Check return value on write of raw pointer

felixjendrusch:
I was recently debugging an unrelated bug that made me look into a few details of the protobuf code and came across [this]( https://github.com/google/protobuf/blob/master/objectivec/GPBCodedOutputStream.m#L945), a call to `-[NSOutputStream write:maxLength:]` that doesn't check the return value (bytes written). A few lines up, in `GPBRefreshBuffer`, the bytes written [are checked](https://github.com/google/protobuf/blob/a48d58df9643781947da57bdc13a23ac8d868346/objectivec/GPBCodedOutputStream.m#L70-L74). This pull requests adds a test case and implements the check.

### DIFF
--- a/objectivec/GPBCodedOutputStream.m
+++ b/objectivec/GPBCodedOutputStream.m
@@ -942,7 +942,10 @@ static void GPBWriteRawLittleEndian64(GPBOutputBufferState *state,
       state_.position = length;
     } else {
       // Write is very big.  Let's do it all at once.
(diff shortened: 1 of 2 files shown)


felixjendrusch:
I signed it!

thomasvl:
Sorry for the delay, some hiccups on our end checking the cla validation.

Anyway, thanks for the fix!

Issue #4374, opened 2018-03-09
py_proto_library fails with "no such target '//:protoc'"


jo2y:
ERROR: /path/to/proto/BUILD:34:1: no such target '//:protoc': target 'protoc' not declared in package '' defined by /path/to/BUILD and referenced by '//proto:foo_py_proto_genproto'

The py_proto_library has defaults for protoc and default_runtime that don't work outside of the protobuf repository. It's a simple fix and I'll send a PR. s#:protoc#@com_google_protobuf//:protoc#

PR #4375, opened 2018-03-09
Replace //:protoc and similar default macro arguments with

jo2y:
Replace //:protoc and similar default macro arguments with @com_google_protobuf prefixed versions. This allows them to work in 3rd party repositories.
Fix a bad visibility rule. :hidden does not exist, but :private does.

Fixes: #4374

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -254,7 +254,7 @@ internal_copied_filegroup(
     srcs = WELL_KNOWN_PROTOS,
     dest = "",
     strip_prefix = "src",
-    visibility = ["//visibility:hidden"],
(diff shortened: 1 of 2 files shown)


pherl:
How will this affect others without using `@com_google_protobuf` repo?

xfxyjwf:
@pherl It must be com_google_protobuf because that's part of our WORKSPACE specification:
https://github.com/google/protobuf/blob/master/WORKSPACE#L1

blaze is trying to enforce this on all repos (i.e., all repos must declare its name in the WORKSPACE file and everyone importing that repo must use the specified name).

PR #4381, opened 2018-03-12
fix json_decode call parameters

urfinjuezz:
This fixes the placement of the JSON_BIGINT_AS_STRING option passed to json_decode, this is being passed as the $assoc parameter which decodes the JSON as an associative array.


urfinjuezz:
I signed it!

TeBoring:
@xfxyjwf How to trigger cocoro tests?

xfxyjwf:
Apply the "kokoro_run" label and it will start.

xfxyjwf:
@TeBoring Is this PR ready to merge?

PR #4385, opened 2018-03-14
Allow setting repeated fields with the #{field_name}= method

kylecrum:
While you can do `message.repeated_field.replace([])`, this is not a very intuitive interface for a Ruby developer. This allows using the `#{field_name}=` method which essentially falls back to the replace method

### DIFF
--- a/ruby/ext/google/protobuf_c/storage.c
+++ b/ruby/ext/google/protobuf_c/storage.c
@@ -606,7 +606,7 @@ static void check_repeated_field_type(VALUE val, const upb_fielddef* field) {
     rb_raise(rb_eTypeError, "Repeated field array has wrong element type");
   }
 
-  if (self->field_type == UPB_TYPE_MESSAGE) { 
+  if (self->field_type == UPB_TYPE_MESSAGE) {
     if (self->field_type_class !=
         Descriptor_msgclass(get_def_obj(upb_fielddef_subdef(field)))) {
       rb_raise(rb_eTypeError,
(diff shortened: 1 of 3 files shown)


kylecrum:
I signed it!

kylecrum:
cc @ebenoist

haberman:
I really understand why someone would want this. We talked about this a lot when we were designing the API. The danger of allowing this usage is that it gives very surprising results if you do this:

```ruby
arr = []
foo.bar = arr

# Whoops, these don't actually affect foo.bar, because
# the assignment didn't actually assign.
arr << 1
arr << 2
arr << 3
```

The `replace()` syntax, despite being a bit more verbose, makes it really clear that the operation is not simple assignment.

If there is another way to be more ergonomic without having surprising edge cases like the above, let us know.

kylecrum:
@haberman interesting perspective and I see the argument for explicitness there.

For context, we're doing a lot with proto messages and so we're trying to make some of the edges not so sharp for developers at large. This is one of the edges that I see frustrating people when I see them build up messages from existing data.

From a developer experience point of view, I say that the case you submitted in your comment would be how I would expect it to work. Going out on a bit of a limb here, but I've been around ruby for a long time and most people assume that everything is passed by value. Modifying an array that has a side effect of changing an associated object would be a bit of an odd side effect since it stretches the boundaries of that object and does some weird things that I think rubyists would not expect.

For example:

```
arr = []
message_1 = Message.new
message_1.bar = arr

message_2 = Message.new
message_2.bar = arr

# .... sometime much later
# I would not expect this to change the values in 2 messages
arr << 2
```

So, IMHO, I think that the ergonomics of having `fieldname=` be the same as replace meets the expectations of a developer using that method. An object should hold it's own state and that state should stay within the boundaries of its object (i.e. we should not be able to modify that state without using a method on the object to do so)

kylecrum:
The other thing that is a little bit confusing is that it works slightly differently in the constructor than it does in other places.

e.g.
```
  # this works
  arr = []
  msg = Message.new(bar: arr)

  # this does not modify the foo attribute in the message
  arr << 'foo'

  # this currently does not work
  msg.bar = arr
```

For better or worse, this PR at least has parity with passing the same value in the constructor.

bobbytables:
This makes total sense to have and should be merged in IMO.

haberman:
I've been convinced on this point. We would now accept this PR, if it were rebased on the current code.

Sorry for holding this up for so long.

PR #4387, opened 2018-03-14

acozzette:
This pull request also deletes scoped_ptr and scoped_array, now that we have switched
to C++11 and can use std::unique_ptr.


pherl:
Thanks for the integration! Please make sure the presubmit check pass.

PR #4395, opened 2018-03-19
Fix bug: whether always_print_enums_as_ints is true or false, it alwa…

stone4774:
…ys print the default value of enums as strings

description of the problem:

call the function MessageToJsonString, and set both always_print_primitive_fields and always_print_enums_as_ints to true. it print the default value of enums as strings, does not as ints.

This submission resolved the issue


stone4774:
Has been submitted, please check again.  Thanks

xfxyjwf:
There is merge conflict with master. Could you update and resolve the conflict?

stone4774:
It has been merged.

stone4774:
Updated and the conflict resolved. Please take a look. Thanks.

PR #4413, opened 2018-03-22
Don't assume Windows builds use MSVC.

pmuetschard:
Switch to using the MSVC options, not based on the cpu, but based on the compiler name. This allows building on Windows with MSYS or MinGW's gcc.


acozzette:
@pmuetschard Thanks for the pull request!

PR #4424, opened 2018-03-25
Fix missing LIBPROTOC_EXPORT.

egorpugin:
This makes possible to build grpc_php_plugin with protobuf in shared mode.


egorpugin:
I signed it!

PR #4439, opened 2018-03-27
Removed atomicops.h stub

acozzette:
We can delete this file since it is no longer used.


.gitignore
```diff
@@ -19,7 +19,7 @@ m4/lt~obsolete.m4
 autom4te.cache
 
 # downloaded files
-./gmock
+/gmock
```
xfxyjwf: What's the difference here?

acozzette: The ./gmock syntax didn't seem to work for me--I think if you want to match only directories at the top level then you have to put a leading slash but no leading dot.

PR #4440, opened 2018-03-27
Write messages to backing field in generated C# cloning code

### LINKED ISSUE #4316: C# copy ctor inconsistently writes to property rather than backing field for non-repeated messages
For example, in the [copy ctor for the Subscription type in Pubsub.cs](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Pubsub.cs#L1967); `PushConfig` and `MessageRetentionDuration` write to the property because they refer to non-repeated message types, whereas all other writes are to the backing field ([from pubsub.proto here](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L405)).

Is there a reason these writes are not always consistently to the backing field?

Thanks :)

ObsidianMinor:
Resolves #4316

### DIFF
--- a/csharp/src/AddressBook/Addressbook.cs
+++ b/csharp/src/AddressBook/Addressbook.cs
@@ -78,7 +78,7 @@ namespace Google.Protobuf.Examples.AddressBook {
       id_ = other.id_;
       email_ = other.email_;
       phones_ = other.phones_.Clone();
-      LastUpdated = other.lastUpdated_ != null ? other.LastUpdated.Clone() : null;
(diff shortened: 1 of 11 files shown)


jskeet:
Fine by me, I *think*. In particular, this doesn't change the one-of handling, right? IIRC, it's important that it uses the property there.

ObsidianMinor:
It does not change one-of handling, it only changes single message fields

xfxyjwf:
Jie, is this PR ready to merge?

anandolee:
The failure is not related to C#

PR #4445, opened 2018-03-28
Add kokoro build status badges.

xfxyjwf:
Moved the badges to the language table as they are per-language based.
Some badges are not yet available due to stuck kokoro jobs.


xfxyjwf:
Also unlike the build result of individual invocations, the status page these badges link to is not visible to external users.

xfxyjwf:
All badges are available now.

PR #4447, opened 2018-03-29
Cleaner protoc artifacts

### LINKED ISSUE #4419: protoc-artifacts Dockerfile: `git clone` fails
The `git clone --depth 1 https://github.com/google/protobuf.git` command in `protoc-artifacts/Dockerfile` fails with a pretty generic error message and error code 128. Entering the docker container and running `GIT_CURL_VERBOSE=1 GIT_TRACE=1 git clone ...` shows that the error is caused by a failure in the TLS negotiation. 

I believe this was caused when Github dropped support for TLS1.0 and TLS1.1: https://githubengineering.com/crypto-removal-notice/

The Dockerfile uses CentOS 6.6. It appears that this version of CentOS does not include support for TLS1.2, at least not in the libcurl library used by git: out-of-the-box, `curl` cannot fetch a Github web page. I used yum in update libcurl, and then curl recognized the `--tlsv1.2` flag and was able to get a web page from Github, but doesn't change the failure with git clone.

(FWIW, I tried changing the dockerfile to use CentOS 7, which does support TLS1.2, and encountered an unrelated error with the command in the dockerfile's last line. Using CentOS 6.9 instead, everything built fine, but I'm not sure what reasons went into choosing 6.6 and if 6.9 may have other unwanted side-effects)

ejona86:
You'll want to look at each commit independently, including its description.
I'll want to submit the commits separately as well.

This does not fix the git clone TLS problem. But it allows you to workaround it
manually using wget instead.

The scl-enable-devtoolset.sh is a more powerful change than it may look
at first glance. It is taken from grpc-java (https://github.com/grpc/grpc-java/commit/9c5e96e376754cf922b1417af184fd8552f9feef) as part of
our work to improve our release process. Coupled with run_in_docker.sh
it allows an improved release and development process as the files are
stored on the host in a normal checkout and the computation is performed
in the docker environment.

run_in_docker.sh (in that referenced commit) is overly complex to
workaround Java issues when running as a uid that does not exist in
/etc/passwd. If not for Java, it would be a single "docker run" line notably
using -v to mount a host directory into the container and -u/-g to set
the user to run as so that created files have the correct permissions.

If one of the commits is unsettling, then I can split it out.

Fixes #4419

### DIFF
--- a/protoc-artifacts/Dockerfile
+++ b/protoc-artifacts/Dockerfile
@@ -11,7 +11,9 @@ RUN yum install -y git \
                    libtool \
                    glibc-static.i686 \
                    glibc-devel \
-                   glibc-devel.i686
+                   glibc-devel.i686 \
+                   && \
+    yum clean all
 
 # Install Java 8
 RUN wget -q --no-cookies --no-check-certificate \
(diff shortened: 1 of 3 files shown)


xfxyjwf:
Thanks for the pull request! Something to be aware of in case you don't already know: copying a file from grpc to protobuf means we are re-licensing the file from grpc's Apache license to protobuf's BSD license. We have consulted the opensource licensing team and they say it's OK but ask us to consult them each time for each repo.

ejona86:
@xfxyjwf, ah, good point. In this case I'm the sole author of the code, so it would be safe. But that is a good point to remember.

PR #4448, opened 2018-03-29
ObjC: IWYU and resolve header order problems.

thomasvl:
fyi - @sergiocampama @dmaclach


src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -84,6 +84,52 @@ bool MessageContainsExtensions(const Descriptor* message) {
...
+    if (MessageContainsGPBArrayField(message->nested_type(i))) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// Checks if the Message contains a field that requires the custom
+// GPBDictionary types. Checks at this message and sub messages.
+bool MessageContainsGPBDictField(const Descriptor* message) {
```
sergiocampama: MessageContainsMapField? the GPBDict naming is for objc protos, and technically we're dealing with proto descriptors here

objectivec/Tests/GPBWireFormatTests.m
```diff
@@ -30,9 +30,11 @@
 
 #import "GPBTestUtilities.h"
 
-#import "GPBCodedInputStream.h"
+#import "GPBProtocolBuffers.h"
```
sergiocampama: just curious, why import the umbrella one here instead of what you use?

src/google/protobuf/compiler/objectivec/objectivec_file.cc
thomasvl: Because some map<> fields just use NSMutableDictionary (String->Message, String->Data, String->String).  So this really is a test to see if they will need a GPBDict class.

objectivec/Tests/GPBWireFormatTests.m
thomasvl: As I started fixing these up, they started needing just about everything; so rather then add everything, it seemed cleaner to just use the public everything header and the required private ones.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
thomasvl: Same goes for the Array side, String/Data/Message all use NSMutableArray, so it was being specific to what gets the custom class.

thomasvl:
Dropping this for a simpler change will lest impact on 3rd part code.

PR #4449, opened 2018-03-29
Deliberately call simple code to avoid Unity linker pruning

jskeet:
The SampleEnumMethod method was previously only called via
reflection, so the Unity linker thought it could be removed.
This PR should avoid that issue, reducing the work needed by
customers to use Google.Protobuf from Unity.

@SetoKaiba could you try this change, removing the linker options in your Unity project? (It would be so nice if Unity developers could just use it without extra configuration.)

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/ReflectionUtil.cs
+++ b/csharp/src/Google.Protobuf/Reflection/ReflectionUtil.cs
@@ -163,6 +163,8 @@ namespace Google.Protobuf.Reflection
         {
             try
             {
+                PreventLinkerFailures();
+                // Try to do the conversion using reflection, so we can see whether it's supported.
                 MethodInfo method = typeof(ReflectionUtil).GetMethod(nameof(SampleEnumMethod));
                 // If this passes, we're in a reasonable runtime.
                 method.CreateDelegate(typeof(Func<int>));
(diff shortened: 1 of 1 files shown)


SetoKaiba:
@jskeet Not working without linker.xml to preserve the dll.

> System.TypeInitializationException: The type initializer for 'Co.Kaiba.Blueeyes.Protocol.BlueeyesBasicReflection' threw an exception. ---> System.MissingMethodException: Default constructor not found for type Google.Protobuf.Reflection.ReflectionUtil+ReflectionHelper`2[[Co.Kaiba.Blueeyes.Protocol.LoginRequest, Assembly-CSharp, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null],[System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]

jskeet:
@SetoKaiba: Thanks for testing. I've added a second commit which should fix this one - but if this fails as well, I think it's probably not worth chasing further. (I can imagine ending up writing more and more code just to avoid having a linker configuration - it gets a little silly.)

If you could test the latest version of this PR, that would be great - but please don't feel any pressure.

SetoKaiba:
@jskeet Ok. It's working now without the need to disable strip in link.xml.

jskeet:
@SetoKaiba: Brilliant - thanks so much. If this is all that's required, I think it's worth doing. I'll get it reviewed and merged.

jskeet:
@jtattermusch I've assigned you as you reviewed the previous PR - you're probably in a reasonable position to understand what I'm trying to do :)

jtattermusch:
I think we can merge this?

jskeet:
@jtattermusch: I was just waiting for @SetoKaiba to give an indication, but it does seem to be better than nothing. Will merge now.

PR #4464, opened 2018-04-02
Move ObjC WKTs to minimal imports to avoid import cycle.

thomasvl:


thomasvl:
fyi - @dmaclach @sergiocampama

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -1518,9 +1520,14 @@ void ImportWriter::AddFile(const FileDescriptor* file,
   const string file_path(FilePath(file));
 
   if (IsProtobufLibraryBundledProtoFile(file)) {
-    protobuf_framework_imports_.push_back(
-        FilePathBasename(file) + header_extension);
-    protobuf_non_framework_imports_.push_back(file_path + header_extension);
+    // The imports of the WKTs are only needed within the library itself,
+    // in other cases, they get skipped because the generated code already
+    // import GPBProtocolBuffers.h and hence proves them.
+    if (include_wkt_imports_) {
```
sergiocampama: tracing this back, this is the output of `IsProtobufLibraryBundledProtoFile`, but we're already inside that `if` case. do you expect these to be different at some point? or is the `IsProtobufLibraryBundledProtoFile` check here enough? making the `include_wkt_imports_` unnecessary

thomasvl: Lost me.  `include_wkt_imports_` is set based on of the file being generated was a wkt.  `IsProtobufLibraryBundledProtoFile` is checking the passed file to see if it is a bundle file (wkt).  So they are "speaking" about different file, no?

PR #4468, opened 2018-04-02
For windows, all python version should use /MT

TeBoring:


TeBoring:
The broken python test is due to tox cannot be downloaded. That issue is not related to this change.

PR #4470, opened 2018-04-03
Amalgamate C++ sources in protobuf Bazel build

jart:
According to the SQLite author, this might make things go between 5% and 10%
faster: https://www.sqlite.org/amalgamation.html

It reduced `du -sh bazel-bin/external/protobuf_archive` from to 20MB to 14MB.
It also reduced the latency on one of my builds by about 10 seconds. If I also
use the --distinct_host_configuration=false it reduced by 20 seconds.

Please note mine editor runs buildifier automatically.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,12 +21,12 @@ config_setting(
 
 MSVC_COPTS = [
     "/DHAVE_PTHREAD",
-    "/wd4018", # -Wno-sign-compare
-    "/wd4514", # -Wno-unused-function
+    "/wd4018",  # -Wno-sign-compare
+    "/wd4514",  # -Wno-unused-function
 ]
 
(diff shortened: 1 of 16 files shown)


xfxyjwf:
Thanks for sending us a patch but I don't think we can accept this. Concatenating all sources into one file only provides a benefit when the build is not parallelized, and can potentially hurt build performance when the giant file becomes a sticking point in the critical path. With bazel I think we are moving towards parallel build. It doesn't seem worthwhile to introduce complexity to improve single-threaded build. Also the additional constraint posed on protobuf developers with this PR is not very pleasant. Having to give symbols globally unique names even when they are in an anonymous namespace? That really is broken C++.

PR #4475, opened 2018-04-04
Fix problem: cmake build failed in c++11  by clang

chenchuanyin:
CMakeLists.txt adds c++11 flags for clang

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -15,6 +15,9 @@ endif()
 # Project
 project(protobuf C CXX)
 
+# Add c++11 flags for clang
+set(CMAKE_CXX_FLAGS "-std=c++11")
+
 # Options
 option(protobuf_BUILD_TESTS "Build tests" ON)
 option(protobuf_BUILD_EXAMPLES "Build examples" OFF)


xfxyjwf:
I'm not sure how this works with visual studio, but since our appveyor tests passed, it seems fine to merge.

chenchuanyin:
When I'm using clang in MacOS, clang don't compile cpp code using c++11.

rcane:
Simply hard-setting CMAKE_CXX_FLAGS has the (hopefully) unintended effect of clearing all flags that have been set by default. In case of Visual C++ this removes the following default flags:

    /DWIN32 /D_WINDOWS /W3 /GR /EHsc

Although the code seems to still compile without those flags, there are some warnings about the missing /EHsc (exception handling) now.

So the line CMakeLists.txt should rather be changed to the following in order to preserve those default flags:

    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")

Mizux:
Why don't using:
```cmake
target_compile_features(protobuf PUBLIC cxx_std_11)
```
it's more the Modern CMake way and customer consuming the target will got it for free !
ref: https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html?highlight=standard

rcane:
This would indeed be the better way to do it. But we would at least need CMake-3.1 for that. Right now the minimum required CMake version is set to 2.8.12. Personally I have no problem using a "newer" CMake (3.1 was released December 17, 2014) but others might be less modern than you and I. ;-)

xfxyjwf:
@Mizux Can you help send us a pull request to make the change?

Mizux:
Try to do it today evening, wanted to check kokoro env first !

FYI on kokoro workers:
- Ubuntu[GCP]: cmake 3.7.1-1
- Windows [GCP]: cmake 3.7.2
- MacOS: cmake 3.9.2
- MacOS External: cmake 3.9.5

PR #4476, opened 2018-04-04
New custom FieldValueParser C++ Feature.

Kwizatz:
Motivation:
Being able to human write byte arrays.

Supose a Byte Array field in the context of
the application consists of an array of
integers or floats, in order to write
a text format protobuf message
the string may have to look something like this:

Int8Array: "\000\001\002\003"

Keeping in mind scalar size (8bit,16bit,32bit,etc)
as well as escape character values makes this task
humanly complicated, being able to supply custom parsers
allows for this syntax:

Int8Array: "0,1,2,3"

or even

Matrix: "1,0,0,0,"
	"0,1,0,0,"
	"0,0,1,0,"
	"0,0,0,1"

The parser would then expect a CSV, handle white space,
and build the byte array properly.

In the case of regular strings the parser may be used
as a formater, removing whitespace, etc,
and for integer and float scalar values transformation
such as degrees to radians may be performed.

Detailed description of the introduced feature:

This commit implements a functionality
similar to TextFormat::(Fast)FieldValuePrinter
but on the parser side.

This allows the creation of custom parse
functions to preprocess strings,
integers and doubles, based on the
io::Tokeinzer::Parse* api.

The default introduced parser is just a wrapper
for the tokenizer functions,
so functionality should not change,
and no old should break.

On the public API side,
besides the FieldValueParser class,
two member functions are introduced:

TextFormat::Parser::RegisterFieldValueParser
TextFormat::Parser::SetDefaultFieldValueParser

Which follow the same structure as their paralells in
TextFormat::Printer.

The FieldValueParser class contains only four member functions:

virtual double ParseFloat(const string& text) const;
virtual void ParseString(const string& text, string* output) const;
virtual void ParseStringAppend(const string& text, string* output) const;
virtual bool ParseInteger(const string& text, uint64 max_value,uint64* output) const;

ParseFloat and ParseInteger have limited utility,
though perhaps they maybe useful to hook a degree to radians
conversion or similar situation.

ParseString is currently not being used but included for completeness,
while ParseStringAppend is likely the one to see most use.

### DIFF
--- a/src/google/protobuf/text_format.cc
+++ b/src/google/protobuf/text_format.cc
@@ -244,7 +244,9 @@ class TextFormat::Parser::ParserImpl {
              bool allow_unknown_enum,
              bool allow_field_number,
              bool allow_relaxed_whitespace,
-             bool allow_partial)
+             bool allow_partial,
+             const FieldValueParser* default_field_value_parser,
+             const CustomParserMap* custom_parsers)
     : error_collector_(error_collector),
       finder_(finder),
       parse_info_tree_(parse_info_tree),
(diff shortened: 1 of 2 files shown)


Kwizatz:
Will do in the near future,
thanks.

Kwizatz:
I have updated indentation, will write unit tests soon.

xfxyjwf:
Closing now. Feel free to reopen.

PR #4505, opened 2018-04-13
Added the ability to ignore unknown types in csharp's JsonParser

charlie430:
This feature request is similar to https://github.com/google/protobuf/issues/2838

Except instead of ignoring fields, we should be able to ignore an unknown type when parsing an Any message.

I added IgnoreUnknownTypes to JsonParser.Settings just like IgnoreUnknownFields to opt into this.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
@@ -848,6 +848,24 @@ namespace Google.Protobuf
         }
 
         [Test]
+        public void Any_IgnoreUnknownType_WithValue()
+        {
+            string json = "{ \"@type\": \"type.googleapis.com/bogus\", \"foo\": \"bar\" }";
+            var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownTypes(true));
+            var any = parser.Parse<Any>(json);
+            Assert.IsTrue(any.Value.IsEmpty);
+        }
+
+        [Test]
+        public void Any_IgnoreUnknownType_NoValue()
+        {
+            string json = "{ \"@type\": \"type.googleapis.com/bogus\" }";
+            var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownTypes(true));
+            var any = parser.Parse<Any>(json);
+            Assert.IsTrue(any.Value.IsEmpty);
+        }
+
+        [Test]
         public void Any_NoTypeUrl()
         {
             string json = "{ \"foo\": \"bar\" }";
--- a/csharp/src/Google.Protobuf/JsonParser.cs
+++ b/csharp/src/Google.Protobuf/JsonParser.cs
@@ -521,29 +521,46 @@ namespace Google.Protobuf
             string typeUrl = token.StringValue;
             string typeName = Any.GetTypeName(typeUrl);
 
+            message.Descriptor.Fields[Any.TypeUrlFieldNumber].Accessor.SetValue(message, typeUrl);
+
             MessageDescriptor descriptor = settings.TypeRegistry.Find(typeName);
+
             if (descriptor == null)
             {
-                throw new InvalidOperationException($"Type registry has no descriptor for type name '{typeName}'");
-            }
+                if (!this.settings.IgnoreUnknownTypes)
+                {
+                    throw new InvalidOperationException($"Type registry has no descriptor for type name '{typeName}'");
+                }
 
-            // Now replay the token stream we've already read and anything that remains of the object, just parsing it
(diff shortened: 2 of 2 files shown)


charlie430:
I signed it!

charlie430:
I signed it!

charlie430:
I understand that ignoring unknown types is not the ideal solution, instead we should preserve the information. I'd prefer to address issue https://github.com/google/protobuf/issues/4511.

However, ignoring unknown types could be a way to allow .Net consumers to deserialize messages with unknown types without throwing an exception until a more appropriate solution is submitted.

xfxyjwf:
When we were discussing the proto3 JSON spec and the API for the conversion library, it was explicitly requested that unknown Any types must fail the JSON conversion due to concerns of silent data losses. Inside Google, this has been strictly enforced and the proxy that does JSON to proto conversion must know about all types it's converting. I would suggest you do the same in your project instead.

charlie430:
@xfxyjwf Thank you for looking into this. I agree that protobuf should not silently lose data.

My project is too generic to know all the types, so it would be infeasible to create a proxy that converts proto JSON to proto binary. For instance, I need to be able to allow external clients (i.e. not in my organization) to define their own models, and pack them into the Any property. However, I would follow that recommendation for a more scoped project where I could know all the types.

It looks like you were also assigned to issue https://github.com/google/protobuf/issues/4511. Is it being considered? I assume this will need to be reviewed by the protobuf specification review team before being implemented, so I understand if you do not have an answer at this time.

xfxyjwf:
https://github.com/google/protobuf/issues/4511 is a proposal that had been considered in the initial design of the Any type but did not make it through. The decision was made that JSON and binary should not be mixed together. Reflected in the implementations, it means the google.protobuf.Any type's payload contains proto binary data only and should not be used to hold JSON data. Likewise, the JSON representation of Any should not contain base64 encoded proto binary data and should only contain the JSON representation of the payload message. This works well with the additional requirement that any JSON-proto converter must know all types it's converting. It's unlikely that we will consider any alternatives at the moment but it could change if we get enough requests.

charlie430:
@xfxyjwf Thank you for your response. Without lossless JSON serialization with unknown types, it may make it difficult for our organization to adopt protobuf for standard external data interchange, as our clients may prefer JSON for data exploration use cases. Consider usage in 3rd party applications, like Tableau and R Studio. I was hoping to propose protobuf for our organization's standard external data interchange format, but it won't be as easy now, since forcing our clients to always use the binary representation may prove difficult. The only thing I can think of is for our clients to register their protobuf models with us as well.

xfxyjwf:
To me the typical use case of the JSON-proto converter looks like this:
1. A client understands JSON only and sends JSON data to a service according to the service's JSON spec.
2. The service receives the JSON data and converts it to proto binary. After that everything in the backend talks in proto binary format.

In this scenario embedding a base64-encoded proto binary in JSON doesn't make sense because the client doesn't know how to construct such payload. Also since the service defines the protocol, it should already know all proto types it's dealing with. That's roughly how we use proto-JSON converter inside Google.

It seems your use case is different and both the client and server are using protobuf already but still want to talk in JSON. Is that the case? Could you help me understand why the client would want to use JSON when they are using protos on both ends already?

charlie430:
Hope this helps illustrate my situation:
1. Service defines a schema with an 'Any' field F
1. External client A uploads data D to service with protobuf packing type T into field F that only client A knows
1. External client B downloads data D from service with JSON
1. Service does not know type T that client A uploaded and therefore cannot serialize data D for client B
1. Assume Service could serialize types with unknown types in 'Any' fields, then client B would also need to know how to deserialize types with unknown types in 'Any' fields

I think this scenario presents itself in situations where the type being defined is generic. In my case, it is a generic table. We have multiple analytics engines that produce tables of data that have attributed metadata to the columns. The table model is generic and used by all of these services. The metadata is the unknown types, and you can think of them as settings for how a column was calculated in the service, like what risk model was used, or what prices were used. Since the metadata is domain specific, it would not make sense to include in the generic table model.

I could generalize the metadata, and my options as I see them are to force metadata to use primitive values, a struct, or represent the information as tables, but after considering these options, I'd prefer metadata to be defined as a protobuf type since this metadata is usually used to drive UI in the application that generated the settings for how the service should calculate the data.

xfxyjwf:
@charlie430 Thanks for the detailed explanation (and sorry for the late response..). In your case, it seems to me you shouldn't send the Any payload to B at all because B can't understand the payload. Can you perhaps filter the data before converting it to JSON? For example, loop through all Any fields and remove all those that the service doesn't know about. This should give you a finer-grained control over what to be discarded and what not.

xfxyjwf:
Closing for now.

PR #4508, opened 2018-04-13
Add proto_library() well_known_protos_proto.

g-easy:
This wraps us all the well known protos into one proto_library that
consumers can add to deps.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -226,6 +226,13 @@ cc_proto_library(
     visibility = ["//visibility:public"],
 )
 
+proto_library(
+    name = "well_known_protos_proto",
+    srcs = [],
+    deps = [dep + "_proto" for dep in WELL_KNOWN_PROTO_MAP],
+    visibility = ["//visibility:public"],
+)
+
 ################################################################################
 # Well Known Types Proto Library Rules
 #


g-easy:
I'd like to add this target, so that grpc can use it in its `cc_grpc_library()` rule like this: https://github.com/g-easy/grpc/commit/2fbf00934c3832b2e58697473f9e43de635c1f6d

So that gRPC users can build the `_${name}_only` target generated by `cc_grpc_library`. Currently it fails due to well known protos not being in the include path when protoc is called.

xfxyjwf:
Can you list explicitly every dependency you use in your BUILD file instead?

g-easy:
Sorry for the long delay. I can't get this to work with dependencies. Here's what I tried:

BUILD:
```
load("@com_github_grpc_grpc//bazel:cc_grpc_library.bzl", "cc_grpc_library")
cc_grpc_library(
    name = "foo_grpc",
    srcs = ["foo.proto"],
    proto_only = False,
    use_external = True,
    well_known_protos = False,
    deps = ["@com_google_protobuf//:timestamp_proto"],
)
```

foo.proto:
```
syntax = "proto3";
package foo;
import "google/protobuf/timestamp.proto";
message Interval {
  google.protobuf.Timestamp start_time = 1;
  google.protobuf.Timestamp end_time = 2;
}
```

WORKSPACE:
```
http_archive(
    name = "com_github_grpc_grpc",
    urls = ["https://github.com/grpc/grpc/archive/master.tar.gz"],
    strip_prefix = "grpc-master"
)
load("@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps")
grpc_deps()
```

Build fails:
```
$ bazel build :foo_grpc
ERROR: .../BUILD:2:1: no such target '@com_google_protobuf//:_timestamp_proto_only': target '_timestamp_proto_only' not declared in package '' defined by .../external/com_google_protobuf/BUILD and referenced by '//:_foo_grpc_only'
ERROR: Analysis of target '//:foo_grpc' failed; build aborted: Loading failed
```

The motivation for this isn't to be able to use `timestamp.proto` specifically, but to be able to use proto definitions from other packages generally, e.g. https://github.com/googleapis/googleapis

xfxyjwf:
@g-easy The error seems to indicate an issue in the cc_grpc_library rule. Can you create an issue against the grpc repo instead?

Issue #4488, opened 2018-04-06
Visual Studio 2017 CE error 


MikeGarrod:
After building a MSVC 2017 Solution file using

cmake -G "Visual Studio 15 2017 Win64" - DCMAKE_INSTALL_PREFIX=../../../../install ../..

I am getting a number of compilation errors in type_traits.cc at line 527, The offending code looks like

    _INLINE_VAR constexpr bool is_convertible_v = is_convertible<_From, _To>::value;

And the 6 errors all complain "A native array cannot contain this type"

Full error below...  (There are actually 6 similar errors 

Error	C2728	
'google::protobuf::compiler::cpp::FieldGenerator': a native array cannot contain this type (compiling source file D:\Dev\Libraries\protobuf\src\google\protobuf\compiler\cpp\cpp_field.cc)	

Project:    libprotoc	

   c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\type_traits	

Line:    525	

I am pretty new to this project and was hoping that someone could point out what I am missing.

Thanks

acozzette:
@MikeGarrod Which version or Git commit are you using? Does the full error message say something about what line in cpp_field.cc is causing the problem?

MikeGarrod:
My mistake, I pulled the latest Master, Release 3.5.1.1 does not have the same issue.

Thanks

acozzette:
@MikeGarrod No problem, I'm glad it's working for you now.

rcane:
The problematic code is the allocation of arrays of unique_ptr like in java_field.cc(214):

    template <>
    FieldGeneratorMap<ImmutableFieldGenerator>::FieldGeneratorMap(...)
      : descriptor_(descriptor),
        field_generators_(new std::unique_ptr<
          ImmutableFieldGenerator>[descriptor->field_count()]) {
    ...
    }

The code itself is not wrong but apparently triggers a bug in Visual C++ 2017 15.6 ([see here](https://developercommunity.visualstudio.com/content/problem/181975/make-unique-fails-on-array-of-unique-ptr.html)). Although the VS bug was supposedly fixed already the problem still persists with the current VS 2017 15.6.6.

acozzette:
@rcane Thanks for tracking down the problem! Do you think we should try to work around this or just wait for the compiler fix? I'm not really familiar with Visual Studio and so I'm not sure how their release process works.

rcane:
[...]

PR #4517, opened 2018-04-16
Fixed a Visual Studio 2017 build error. (#4488)

rcane:
The current 15.6.x versions of Visual Studio 2017 contain a bug that
prevent them from compiling the following construct under certain
conditions:

std::unique_ptr<std::unique_ptr<Foo> []> foos;

This will fail to compile if Foo is an abstract class. To work-around
the problem the whole construct was change into:

std::vector<std::unique_ptr<Foo>> foos;

This not only fixes the compiler error but is also more readable than
previous version.


rcane:
I signed it!

acozzette:
ok to test (this comment will trigger some other CI runs)

xfxyjwf:
@acozzette Jenkins has been turned down already so we no longer need to reply "ok to test" to trigger those tests. The kokoro tests use a different trigger: the "kokoro:run" label. You just need to apply the label and kokro will pick it up (and remove the label so you can lable it again to trigger another run if needed).

PR #4522, opened 2018-04-17
+ added an option: enum_value_name_style, it expected one of: full, s…

shrimpz:
…hort.

  where it equal ~short~, the name of enum's value liked C# or Java style.
  such as:
enum kGenderType
{
  kGenderType_None = 0;
  kGenderType_Male = 1;
  kGenderType_Female = 2;
};

will generated JS code like:
proto.kGenderType = {
  NONE : 0,
  MALE: 1,
  FEMALE: 2
}

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -391,6 +391,31 @@ string ToEnumCase(const string& input) {
   return result;
 }
 
+string ToEnumValueName(const string& enumName, const string& valueName) {
+
+    string result;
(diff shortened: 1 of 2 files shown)


xfxyjwf:
@shrimpz Thanks for contributing to protobuf code base but I don't think we can accept this change. We want to avoid new options in general and are particularly cautious about options that affect public APIs. Options like the one proposed in this PR alter public APIs and must be forced upon the entire transitive closure of protos because protos with different styles can't depend on each other. That effectively create two protobuf APIs that aren't compatible and will cause issues when people mix the two.

PR #4530, opened 2018-04-18
PHP array constructors for protobuf messages

bshaffer:
Same as #3349, but to the `php-generated-dev` branch


bshaffer:
These changes will result in the ability to do something like this (for example):

```php
$m = new TestMessage([
    'int' => -42,
    'message' => ['a' => 33],
    'repeated_message' => [
        new TestMessage_Sub(['a' => 34]),
        ['a' => 35],
    ],
    'map_message' => [
        1 => new TestMessage_Sub(['a' => 36])
    ],
]);
```

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -66,19 +66,30 @@ class Message
...
-                $this->$setter($this->defaultValue($field));
+        if ($data instanceof Descriptor) {
+            $this->initWithDescriptor($data);
+        } else {
+            $this->initWithGeneratedPool();
+            if (is_array($data)) {
+                $this->mergeFromArray($data);
+            } else if (!empty($data)) {
+                throw new \InvalidArgumentException(
+                    'Message constructor must be an array, Descriptor, or null.'
```
TeBoring: Don't mention Descriptor to users.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -66,19 +66,30 @@ class Message
     /**
      * @ignore
      */
-    public function __construct($desc = NULL)
+    public function __construct($data = NULL)
     {
         // MapEntry message is shared by all types of map fields, whose
         // descriptors are different from each other. Thus, we cannot find a
         // specific descriptor from the descriptor pool.
-        if (get_class($this) === 'Google\Protobuf\Internal\MapEntry') {
```
TeBoring: Where is MapEntry handled now?

bshaffer: This is handled in `initWithDescriptor`, which is called when `$desc` is a descriptor. I can only find two places `MapEntry` is created, [here](https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L355) and [here](https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L1188). In both cases, a descriptor is provided in the constructor.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -937,23 +1074,51 @@ private function convertJsonValueToProtoValue(
...
+     *          'endTime' => time(),
+     *     ]
+     * ]);
+     * ```
+     *
+     * @param array $array An array containing message properties and values.
+     * @return null.
+     * @throws Exception Invalid data.
+     */
+    public function mergeFromArray($array)
```
TeBoring: I don't want to make this public.

php/tests/php_implementation_test.php
```diff
@@ -513,4 +513,128 @@ public function testPackedByteSize()
...
+
+        $messageFromArray = new TestMessage($requestData);
+        $messageFromJson = new TestMessage();
+        $messageFromJson->mergeFromJsonString(json_encode($requestData));
+
+        foreach ($requestData as $key => $value) {
+            $getter = 'get' . str_replace(' ', '', ucwords(
+                str_replace('_', ' ', $key)
+            ));
+            $this->assertTrue(empty($messageFromArray->$getter()));
```
TeBoring: Primitive field default is not null. e.g., 0 for int32

php/src/Google/Protobuf/Internal/Message.php
TeBoring: ```mergeFrom(new TestMessage([...]))``` should be ok.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -151,6 +162,19 @@ public function __construct($desc = NULL)
         }
     }
 
+    /**
+     * @ignore
+     */
+    private function initWithDescriptor(Descriptor $desc)
+    {
+        $this->desc = $desc;
```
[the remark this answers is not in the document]
bshaffer: It's not needed because it's an internal class and the only times it's instantiated is with a descriptor, so this'll be executed. 

Essentially I took the behavior but made it generic. Checking the class in the constructor is not as clean, and this results in the same behavior.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -66,19 +66,30 @@ class Message
...
     {
         // MapEntry message is shared by all types of map fields, whose
         // descriptors are different from each other. Thus, we cannot find a
         // specific descriptor from the descriptor pool.
-        if (get_class($this) === 'Google\Protobuf\Internal\MapEntry') {
-            $this->desc = $desc;
-            foreach ($desc->getField() as $field) {
-                $setter = $field->getSetter();
-                $this->$setter($this->defaultValue($field));
+        if ($data instanceof Descriptor) {
```
TeBoring: I think you should check ```$this instanceof MapEntry```

bshaffer: We can do this, but it seemed better to make this not hardcoded for a specific class, don't you think?

php/src/Google/Protobuf/Internal/Message.php
bshaffer: Are you saying we should allow for `Message` objects to be passed in the constructor?

bshaffer: Oh I see, you want to use the existing `mergeFrom` method. Sounds good!

php/src/Google/Protobuf/Internal/Message.php
TeBoring: The original behavior was only do the work for MapEntry if arg is descriptor. Your refactoring change the behavior.

php/tests/php_implementation_test.php
bshaffer: The function `empty` evaluates to true for zero also:
```php
php > var_dump(empty(0));
bool(true)
```

php/tests/php_implementation_test.php
```diff
@@ -513,4 +513,128 @@ public function testPackedByteSize()
...
+            'optional_fixed64' => 47,
+            'optional_sfixed32' => -46,
+            'optional_sfixed64' => -47,
+            'optional_float' => 1.5,
+            'optional_double' => 1.6,
+            'optional_bool' => true,
+            'optional_string' => 'a',
+            'optional_bytes' => 'b',
+            'optional_enum' => TestEnum::ONE,
+            'optional_message' => [
```
TeBoring: Can we just allow submessage? ```'optional_message' => new TestMessage_Sub([...])``` will be the same.

bshaffer: It's not quite the same, as it requires the user to import the classes for all nested messages. Why not allow both? 

One of the main use cases we're trying to meet is when there are nested properties of messages, and importing / instantiating each class and object is unnecessarily tedious, e.g. 

```php
use Google\Cloud\Monitoring\V3\Point;
use Google\Cloud\Monitoring\V3\TypedValue;

$value = new TypedValue([
    'double_value' => new TypedValue(['value' => 123.45]),
]);
```

Could become

```php
use Google\Cloud\Monitoring\V3\Point;

$point = new Point(['double_value' => ['value' => 123.45]]);
```

php/tests/php_implementation_test.php
TeBoring: Setting fields with null is undefined behavior. Don't test it.

bshaffer: Do you want me to remove this test entirely then?

php/tests/php_implementation_test.php
TeBoring: Allowing both makes the syntax confusing and causing unnecessary maintenance for us. The tediousness seems trivial to me.

php/src/Google/Protobuf/Internal/Message.php
TeBoring: I think you misunderstood my point. I don't want to change the behavior of ```mergeFrom```. I want to make ```mergeFromArray``` protected instead of public

php/tests/php_implementation_test.php
bshaffer: As requested I've removed this in b80f9f9. However, I do think the flexibility to use arrays for message properties would be useful in some cases, e.g. repeated fields and messages with a single property (as above).

bshaffer:
Please note 11155a6 - as suggested by @michaelbausor I've added a doc block for the array constructor. I think this looks great and will be very useful for IDEs and users poking around in the generated code classes. Thanks Michael!

php/src/Google/Protobuf/Internal/Message.php
bshaffer: @TeBoring Okay, please check it now. I added it back and made it protected.

php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -71,9 +71,29 @@ class DescriptorProto extends \Google\Protobuf\Internal\Message
...
     private $has_reserved_name = false;
 
-    public function __construct() {
+    /**
+     * Constructor.
+     *
+     * @param array $data {
+     *     Optional. Data for populating the Message object.
+     *
+     *     @type string $name
```
TeBoring: Do we need info for all fields in doc? This seems too much for me. User should be able to figure out the correct through field accessors.

php/tests/php_implementation_test.php
TeBoring: I think you only need to test optional message

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -786,27 +827,99 @@ private function convertJsonValueToProtoValue(
...
+            $field,
+            $is_map_key);
+    }
+
+    private function convertNativeValueToProtoValue(
+        $value,
+        $field,
+        $is_map_key = false)
+    {
+        if (is_null($value)) {
```
TeBoring: For json, null is acceptable for primitive fields. However, for array constructor, null is not acceptable for primitive fields. You need to add check for this case.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -786,27 +827,99 @@ private function convertJsonValueToProtoValue(
...
+                }
+                return $value;
+        }
+        return $this->convertNativeValueToProtoValue(
+            $value,
+            $field,
+            $is_map_key);
+    }
+
+    private function convertNativeValueToProtoValue(
```
TeBoring: check fields are implemented in https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/GPBUtil.php
I think we should reuse them here for array constructor type check.

php/src/Google/Protobuf/Internal/DescriptorProto.php
michaelbausor: They can, but we want to help users as much as possible. Otherwise they will have to search the entire class looking at getters/setters to see what options are supported, and also they won't know necessarily to use `nested_type` vs. `nestedType`.

TeBoring: I prefer a general guide here, e.g., use the original field name defined in proto file.
If field amount is small, then there is small overhead to look at the file. If field amount is large, then looking at doc here will be the same as looking at the whole file. In our experience, it's quite common that user define hundreds of fields in a single message. To that case, the doc will be unreasonably huge.

TeBoring: Or shall we first check how many fields are there? and then decide whether to generate the doc?

TeBoring: Besides, I think only field name and type is needed. User should go to the doc of accessor for detailed info.

bshaffer: @TeBoring I think it is important to provide the constructor documentation accurately, even if that means a long list of fields. One thing we could do is, if the protobuf has say 100 fields, we could provide only one-line for it, and omit the description comment from the constructor, e.g.

```
## With < 100 fields 
@param int count
       The number of things that make up something. This is the description that exists
       in the proto file before the field, if specified.

## With > 100 fields 
@param int count
```

php/src/Google/Protobuf/Internal/Message.php
bshaffer: What should happen in the array constructor case? Do you want an exception thrown? e.g:

```php
new TestMessage(['intField' => null]);

// @throws InvalidArgumentException "field intField is an int and cannot be null"
```

If we do this, however, it's inconsistent, as it is possible to call `$testMessage->setIntField(null);`

bshaffer: Nevermind! I'm wrong. I will follow the existing behavior, which is the following:

```php
if (is_numeric($var)) {
    $var = intval($var);
} else {
    throw new \Exception("Expect integer.");
}
```

php/src/Google/Protobuf/Internal/DescriptorProto.php
TeBoring: ruby also supports array constructor. However, it doesn't provide any info about params. Besides, previously, we also don't provide detailed info at accessors. Can we postpone this to another PR?

TeBoring: For java, it's the same.

php/src/Google/Protobuf/Internal/Message.php
TeBoring: Please follow the ```checkInt32``` in GPBUtil for this case.

php/src/Google/Protobuf/Internal/DescriptorProto.php
michaelbausor: I think Java uses the builder pattern, so all of the constructors are private.
I am not enough of a Rubyist to interpret what is going on in this file - it looks like the protobuf classes are dynamic, and have no documentation whatsoever? https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-video_intelligence/lib/google/cloud/videointelligence/v1/video_intelligence_pb.rb
Python seems to generate something in its docs that lists the supported "attributes": https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/videointelligence/google/cloud/videointelligence_v1/proto/video_intelligence_pb2.py#L1132
Go uses structs, e.g.: https://github.com/google/go-genproto/blob/master/googleapis/cloud/videointelligence/v1/video_intelligence.pb.go#L169

To be honest I'm not really sure what the right comparison language is. Both python and ruby seem to support "dynamic" protobuf classes in some sense; Java uses builders; Go uses structs.

TeBoring: 1) My feeling to this change is that is relatively trivial. And it's also not clear to how many users will think it useful and how many users will think it bothering.
2) This is not a blocking change. We can also add it later without affecting any existing users.
3) In all other languages, protobuf is also exposed as API for other libraries. We haven't been asked for that feature yet. From this experience, I don't think it's very important to add it.

We can add it later if we find it is very important. Right now, if you want your library's proto generated file have that, could you add it in your own exposed file?

TeBoring: I am ok with adding these docs. Users are fully responsible for the doc content (e.g., if docs are too long, we are not responsible to make it shorter. Users need to control their proto files.)

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -1035,6 +1074,24 @@ protected function mergeFromJsonArray($array)
...
             return;
         }
+        if ($is_json) {
+            $this->mergeFromArrayJsonImpl($array);
+        } else {
+            // Just call the setters for the field names
+            foreach ($array as $key => $value) {
+                $field = $this->desc->getFieldByJsonName($key)
+                    ?: $this->desc->getFieldByName($key);
+                if (is_null($field)) {
```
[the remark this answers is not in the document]
bshaffer: This is if the `$field` object is null though. I am not sure in what case this could arise, but if it did we'd get a PHP Fatal error on the call to `$field->getSetter()` in the next line.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -1035,6 +1074,24 @@ protected function mergeFromJsonArray($array)
             }
             return;
         }
+        if ($is_json) {
+            $this->mergeFromArrayJsonImpl($array);
+        } else {
+            // Just call the setters for the field names
+            foreach ($array as $key => $value) {
+                $field = $this->desc->getFieldByJsonName($key)
```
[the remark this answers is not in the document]
bshaffer: I was wondering about this. The reason I kept it is I figured a user might see the response from an API and use those keys. I am not familiar with the differences enough to know if this is a valid use case, however. Is this the difference between `field_name` and `fieldName`?

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -1035,6 +1074,24 @@ protected function mergeFromJsonArray($array)
             }
             return;
         }
+        if ($is_json) {
```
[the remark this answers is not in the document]
bshaffer: I was wondering about this as well. My thoughts were that if we allow well-known types, our user could do the following:

```php
// Allowing well-known types for array construction
new TestMessage(['double_field' => 3.4]);
// not allowing well-known-types
new TestMessage(['double_field' => ['value' => 3.4]]);
```

php/src/Google/Protobuf/Internal/Message.php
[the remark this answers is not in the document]
bshaffer: Yes, I just looked into it. If we keep this, then array constructors also work with JSON-formatted names, such as
```php
new TestMessage([
    'someField' => 'foo',
    'someOtherFIeld' => 'bar',
]);
```

php/src/Google/Protobuf/Internal/Message.php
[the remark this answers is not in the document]
TeBoring: Well known types are also messages. For sub-message, we only allow setting via sub-message in the array constructor. To be consistent, I think we should only allow setting via sub-message for well known types. If needed, we can add special constructors to well known type messages. However, for now, I think what we already have is enough.

php/src/Google/Protobuf/Internal/Message.php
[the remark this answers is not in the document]
TeBoring: oh, sorry, I misunderstood previously. Shall we also throw an exception in this case?

php/src/Google/Protobuf/Internal/Message.php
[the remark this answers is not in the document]
TeBoring: Could you remove it for now? The array constructor currently doesn't take json. I don't want to give users the impression that they are interchangeable. It seems to me confusing that sometimes json features are allowed in the constructor but sometimes it's not.

php/src/Google/Protobuf/Internal/EnumDescriptorProto_EnumReservedRange.php
```diff
@@ -36,9 +36,21 @@ class EnumDescriptorProto_EnumReservedRange extends \Google\Protobuf\Internal\Me
...
 
-    public function __construct() {
+    /**
+     * Constructor.
+     *
+     * @param array $data {
+     *     Optional. Data for populating the Message object.
+     *
+     *     @type int $start
+     *           Inclusive.
```
bshaffer: Should we not print the comments after the fields? These don't really fit in this case, and I imagine in most cases where this is done. An example of this:

```proto
  message ReservedRange {
    optional int32 start = 1; // Inclusive.
    optional int32 end = 2;   // Exclusive.
  }
```

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -841,20 +845,28 @@ private function convertJsonValueToProtoValue(
...
-                    return -INF;
+                $enum_value = $field->getEnumType()->getValueByName($value);
+                if (!is_null($enum_value)) {
+                    return $enum_value->getNumber();
                 }
-                if ($value === "NaN") {
-                    return NAN;
+                throw new GPBDecodeException(
+                        "Enum field only accepts bool value");
+            case GPBType::STRING:
```
TeBoring: Could you move ENUM and STRING cases to their original places?

bshaffer: yep, good call

PR #4532, opened 2018-04-19
PHP string-to-int and int-to-string methods for enums

bshaffer:
In accordance with [Python](https://developers.google.com/protocol-buffers/docs/reference/python-generated#enum), this PR adds `name` and `value` methods to PHP Protobuf Enums, so that conversion is programmatically possible between the two.

```php
$this->assertEquals('VALUE_A', SomeEnum::name(SomeEnum::VALUE_A));
$this->assertEquals(5, SomeEnum::value('VALUE_B'))
```

cc @michaelbausor @dwsupplee @tmatsuo

### DIFF
--- a/php/src/Google/Protobuf/EnumTrait.php
+++ b/php/src/Google/Protobuf/EnumTrait.php
@@ -1,0 +1,62 @@
+<?php
+
+// Protocol Buffers - Google's data interchange format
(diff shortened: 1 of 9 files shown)


bshaffer:
Closing, as this was supposed to be to the `php-generated-dev` branch

PR #4533, opened 2018-04-19
PHP typehints for message setters

bshaffer:
same as #3331, but to the `php-generated-dev` branch

This adds typehints for all Message setters. A few other interesting things have happened as a result:

  1. Setters for Message types no longer pass-by-reference, since this is implicit for PHP objects
  1. Setters for Message types now accept an optional `null`, since the base `Message` class requires passing `null` to the public setters in order for the method `clear()` to work.


anandolee:
ping @TeBoring

xfxyjwf:
ping @TeBoring

xfxyjwf:
ping @TeBoring

PR #4534, opened 2018-04-19
PHP string-to-int and int-to-string methods to enums

bshaffer:
In accordance with [Python](https://developers.google.com/protocol-buffers/docs/reference/python-generated#enum), this PR adds `name` and `value` methods to PHP Protobuf Enums, so that conversion is programmatically possible between the two.

```php
$this->assertEquals('VALUE_A', SomeEnum::name(SomeEnum::VALUE_A));
$this->assertEquals(5, SomeEnum::value('VALUE_B'))
```

cc @michaelbausor @dwsupplee @tmatsuo

### DIFF
--- a/php/tests/generated_class_test.php
+++ b/php/tests/generated_class_test.php
@@ -8,6 +8,7 @@ require_once('test_util.php');
 use Google\Protobuf\Internal\RepeatedField;
 use Google\Protobuf\Internal\MapField;
 use Google\Protobuf\Internal\GPBType;
+use Google\Protobuf\Internal\EnumTrait;
 use Bar\TestLegacyMessage;
 use Bar\TestLegacyMessage_NestedEnum;
 use Bar\TestLegacyMessage_NestedMessage;
@@ -232,6 +233,28 @@ class GeneratedClassTest extends TestBase
         // Set string.
         $m->setOptionalEnum("1");
         $this->assertEquals(TestEnum::ONE, $m->getOptionalEnum());
+
+        // Test Enum methods
+        $this->assertEquals('ONE', TestEnum::name(1));
+        $this->assertEquals(1, TestEnum::value('ONE'));
+    }
+
+    /**
+     * @expectedException UnexpectedValueException
+     * @expectedExceptionMessage Enum Foo\TestEnum has no name defined for value -1
+     */
+    public function testInvalidEnumValueThrowsException()
+    {
+        TestEnum::name(-1);
+    }
+
+    /**
+     * @expectedException UnexpectedValueException
+     * @expectedExceptionMessage Enum Foo\TestEnum has no value defined for name DOES_NOT_EXIST
+     */
+    public function testInvalidEnumNameThrowsException()
+    {
+        TestEnum::value('DOES_NOT_EXIST');
     }
 
     public function testNestedEnum()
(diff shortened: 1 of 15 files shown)


php/src/Google/Protobuf/EnumTrait.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+// Protocol Buffers - Google's data interchange format
+// Copyright 2017 Google Inc.  All rights reserved.
```
michaelbausor: Nit: 2018

anandolee:
ping @TeBoring

xfxyjwf:
ping @TeBoring

xfxyjwf:
ping @TeBoring

php/src/Google/Protobuf/EnumTrait.php
```diff
@@ -0,0 +1,63 @@
+<?php
```
TeBoring: Could you put this into internal namespace?

php/tests/generated_class_test.php
```diff
@@ -226,6 +227,37 @@ public function testEnumField()
...
+        // Test Enum methods
+        $this->assertEquals('ONE', TestEnum::name(1));
+        $this->assertEquals(1, TestEnum::value('ONE'));
+    }
+
+    /**
+     * @expectedException LogicException
+     * @expectedExceptionMessage Classes implementing Google\Protobuf\EnumTrait must define the static property $valueToName to use this function
+     */
+    public function testBadEnumThrowsException()
```
TeBoring: why this test is needed? ```valueToName``` is implementation detail. That should not be part of API

php/src/Google/Protobuf/EnumTrait.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Google\Protobuf;
+
+use LogicException;
+use UnexpectedValueException;
+
+trait EnumTrait
+{
+    public static function name($value)
+    {
+        if (!property_exists(__CLASS__, 'valueToName')) {
```
TeBoring: This should be an internal error instead of an exception.

php/tests/generated_class_test.php
bshaffer: I just wanted to test `EnumTrait` responds in a sane way if implemented incorrectly. Since there's no way to prevent users from implementing it themselves. I can remove the test if you'd like.

php/src/Google/Protobuf/EnumTrait.php
bshaffer: What do you mean? If PHP encounters an uncaught exception in a web request, this will result in an Internal Error.

php/tests/generated_class_test.php
TeBoring: If you do need it, you can put it in the php_implementation_test

bshaffer: Done!

TeBoring:
Could you avoid unrelated commits in this change?

bshaffer:
@TeBoring I removed them... I accidentally rebased this branch using `master` instead of `php-generated-dev`.

TeBoring:
Any update?

michaelbausor:
@bshaffer Any chance of reviving this PR?

beccasaurus:
ping @bshaffer et al to breathe some life back into this PR 😄

TeBoring:
Fixed in #5342

bshaffer:
Sorry for not responding, I've been on leave for 2 months and just returned! Glad you guys got this taken care of!

PR #4536, opened 2018-04-19
PHP namespaces for nested messages and enums

bshaffer:
The proto

```
message TestMessage {
  NestedEnum nested_enum = 1;
  NestedMessage nested_message = 2;

  enum NestedEnum {
      ZERO = 0;
  }

  // Test nested messages
  message NestedMessage {
      int32 a = 1;
  }
}
```
Will result in the files
```
TestMessage.php
TestMessage/NestedEnum.php
TestMessage/NestedMessage.php
```

With namespaces corresponding to the classes, e.g.

```php
namespace TestMessage;

class NestedEnum
{
    // ...
}
```


anandolee:
ping @TeBoring

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -744,7 +744,7 @@ void GenerateMessageToPool(const string& name_prefix, const Descriptor* message,
     return;
   }
   string class_name = name_prefix.empty()?
-      message->name() : name_prefix + "_" + message->name();
+      message->name() : name_prefix + "\\" + message->name();
```
TeBoring: ClassNamePrefix?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -995,9 +995,9 @@ void GenerateEnumFile(const FileDescriptor* file, const EnumDescriptor* en,
     if (!php_namespace.empty()) {
       printer.Print(
           "namespace ^name^;\n\n",
-          "name", php_namespace);
+          "name", fullname.substr(0, lastindex));
     }
-  } else if (!file->package().empty()) {
+  } else if (!file->package().empty() || lastindex != string::npos) {
```
TeBoring: Shall we merge the above branch and here?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1054,9 +1054,9 @@ void GenerateMessageFile(const FileDescriptor* file, const Descriptor* message,
...
```
TeBoring: Same as above.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1414,7 +1414,7 @@ std::string GeneratedClassName(const Descriptor* desc) {
   std::string classname = desc->name();
   const Descriptor* containing = desc->containing_type();
   while (containing != NULL) {
-    classname = containing->name() + '_' + classname;
+    classname = containing->name() + '\\' + ClassNamePrefix(classname, desc) + classname;
```
TeBoring: Could you add a test in which containing message has reserved name? For example:
```
message Empty {
  message Nested {}
}
```

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1424,7 +1424,7 @@ std::string GeneratedClassName(const EnumDescriptor* desc) {
...
```
TeBoring: Same as above.

TeBoring:
Previously, we didn't take into account reserved name in namespace. Can we also fix it?

TeBoring:
In https://github.com/google/protobuf/blob/master/php/tests/compatibility_test.sh, we need a test to verify the new runtime can work with old generated code.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -993,14 +992,12 @@ void GenerateEnumFile(const FileDescriptor* file, const EnumDescriptor* en,
...
-    const string& php_namespace = file->options().php_namespace();
-    if (!php_namespace.empty()) {
-      printer.Print(
-          "namespace ^name^;\n\n",
-          "name", php_namespace);
-    }
-  } else if (!file->package().empty()) {
+  const string& php_namespace = file->options().php_namespace();
+  if (
+    !php_namespace.empty()
```
TeBoring: Can we do
```
if (!php_namespace.empty() ||
    (!file->options().has_php_namespace() && !file->package().empty()) ||
    lastindex != string::npos) {
...
}
```

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1052,14 +1049,12 @@ void GenerateMessageFile(const FileDescriptor* file, const Descriptor* message,
...
-  if (file->options().has_php_namespace()) {
-    const string& php_namespace = file->options().php_namespace();
-    if (!php_namespace.empty()) {
-      printer.Print(
-          "namespace ^name^;\n\n",
-          "name", php_namespace);
-    }
-  } else if (!file->package().empty()) {
+  const string& php_namespace = file->options().php_namespace();
+  if (
```
TeBoring: Same as above

src/google/protobuf/compiler/php/php_generator.cc
TeBoring: Actually, I think we just need to check lastindex != string::npos

xfxyjwf:
Ping @TeBoring

TeBoring:
I don't think you need to generate two different classes for classes under php/src/Google/Protobuf/Internal. Simply change them with be ok. (you still need two classes for classes under php/src/Google/Protobuf)

php/src/Google/Protobuf/Field_Cardinality.php
```diff
@@ -4,36 +4,7 @@
...
-    /**
-     * For repeated fields.
-     *
-     * Generated from protobuf enum <code>CARDINALITY_REPEATED = 3;</code>
-     */
-    const CARDINALITY_REPEATED = 3;
-}
+// This class has been renamed.
+// @see Google\Protobuf\Field\Cardinality
+class_exists(Field\Cardinality::class);
```
TeBoring: Can we also add a deprecate warning here?

php/src/Google/Protobuf/Timestamp.php
```diff
@@ -156,25 +175,5 @@ public function setNanos($var)
         return $this;
     }
 
-    /**
```
TeBoring: Please add these methods back.

php/tests/generated_class_test.php
```diff
@@ -12,10 +12,13 @@
 use Foo\TestIncludeNamespaceMessage;
 use Foo\TestIncludePrefixMessage;
 use Foo\TestMessage;
+use Foo\TestMessage\Sub;
 use Foo\TestMessage_Sub;
```
TeBoring: Do we need the old one?

php/src/Google/Protobuf/Timestamp.php
bshaffer: These were removed when I regenerated the protos. Do you know why they were removed?

php/src/Google/Protobuf/Field/Cardinality.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * For repeated fields.
+     *
+     * Generated from protobuf enum <code>CARDINALITY_REPEATED = 3;</code>
+     */
+    const CARDINALITY_REPEATED = 3;
+}
+
+// Adding a class alias for backwards compatibility with the previous class name.
+class_alias(Cardinality::class, \Google\Protobuf\Field_Cardinality::class);
```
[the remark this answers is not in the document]
bshaffer: There's no reason to.  When `class_exists` is called in `Field_Cardinality`, it will autoload this class file and the alias will be added.

php/src/Google/Protobuf/Any.php
```diff
@@ -100,30 +111,65 @@ class Any extends \Google\Protobuf\Internal\Message
      */
     private $value = '';
 
-    const TYPE_URL_PREFIX = 'type.googleapis.com/';
```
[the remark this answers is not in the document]
bshaffer: I can add it back no problem. But did I do something wrong in regenerating these classes? I just passed the protos to the newly compiled `protoc`.

php/src/Google/Protobuf/Field_Cardinality.php
bshaffer: Yeah, not a problem.

php/tests/generated_class_test.php
bshaffer: We run a test with it [here](https://github.com/google/protobuf/pull/4536/files#diff-71f0202385891cc441f357b417daa431R407) to verify compatibility. So while it's not needed (we could just reference it as `Foo\TestMessage_Sub` in the typehint), it's cleaner to import it.

php/src/Google/Protobuf/Field/Cardinality.php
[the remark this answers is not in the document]
TeBoring: If we put this in Field_Cardinality, all code for compatibility issue will be in the file that we want to delete later. If technically we can do this, I think it is easier for us to clean it.

php/src/Google/Protobuf/Timestamp.php
TeBoring: There methods were manually added on generated code. Please add them back.

php/tests/generated_class_test.php
TeBoring: I prefer to test it in the compatibility test. I will set it up. You don't need to add compatibility test in the code.

php/tests/generated_class_test.php
```diff
@@ -635,9 +688,6 @@ public function testMessageWithoutNamespace()
         $repeatedNoNamespaceMessage = $m->getRepeatedNoNamespaceMessage();
         $repeatedNoNamespaceMessage[] = new NoNameSpaceMessage();
         $m->setRepeatedNoNamespaceMessage($repeatedNoNamespaceMessage);
-
-        $n = new NoNamespaceMessage();
```
TeBoring: Why this test is removed?

php/tests/generated_class_test.php
```diff
@@ -666,19 +723,80 @@ public function testPrefixMessage()
...
+        $this->assertSame($n, $m->getOptionalMessage());
+        $this->assertSame(TestMessage\NestedEnum::ZERO, $m->getOptionalNestedEnum());
+    }
 
-        $n = new TestNamespace();
-        $n->setA(1);
-        $m->setNamespaceMessage($n);
-        $this->assertSame(1, $m->getNamespaceMessage()->getA());
+    public function testNestedMessagesAndEnumsWithPrefixes()
+    {
```
TeBoring: Shall we merge this with testPrefixMessage and call it testNestedMessagesAndEnumsWithPrefix?

php/tests/generated_class_test.php
```diff
@@ -649,6 +699,13 @@ public function testEnumWithoutNamespace()
         $m->setRepeatedNoNamespaceEnum($repeatedNoNamespaceEnum);
     }
 
+    public function testNestedMessageWithoutNamespace()
```
TeBoring: Rename to testMessageWithoutNamespace?

TeBoring:
Shall we add a proto/empty/echo.proto file?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -116,31 +116,23 @@ std::string RenameEmpty(const std::string& name) {
...
-    return envm->full_name();
+    return desc->full_name();
   }
 }
 
 template <typename DescriptorType>
 std::string ClassNamePrefix(const string& classname,
-                            const DescriptorType* desc) {
+                            const DescriptorType* desc,
+                            bool add_php_prefix) {
```
TeBoring: Why do we need add_php_prefix? Don't we have all info in desc?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1013,32 +1082,24 @@ void GenerateEnumFile(const FileDescriptor* file, const EnumDescriptor* en,
...
 
-  if (file->options().has_php_namespace()) {
-    const string& php_namespace = file->options().php_namespace();
-    if (!php_namespace.empty()) {
-      printer.Print(
-          "namespace ^name^;\n\n",
-          "name", php_namespace);
-    }
-  } else if (!file->package().empty()) {
+  if (!file->options().php_namespace().empty() ||
```
TeBoring: Can we just check lastindex?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1072,14 +1144,9 @@ void GenerateMessageFile(const FileDescriptor* file, const Descriptor* message,
...
+      (!file->options().has_php_namespace() && !file->package().empty()) ||
+      lastindex != string::npos) {
```
TeBoring: Same as above.

php/src/Google/Protobuf/Field/Cardinality.php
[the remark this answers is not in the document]
bshaffer: I tested for this, and we can't technically do it. The compatibility tests I added for typehints fails in [this case](https://github.com/google/protobuf/pull/4536/files#diff-71f0202385891cc441f357b417daa431R430). Essentially, passing a new class into an old typehint throws a PHP Fatal Error unless the `class_alias` exists in the new class.

php/tests/generated_class_test.php
bshaffer: This is a very important tests that I'd like to make sure gets integrated properly. I can move them to the compatibility test if you'd like, but I do think these tests need to be a part of this PR.

php/src/Google/Protobuf/Field_Cardinality.php
bshaffer: One thing to consider about using `trigger_error` with the deprecated warning is that many systems wrap `E_USER_DEPRECATED` warnings in an exception. So this will break backwards compatibility for some users. 

Symfony gets around this by suppressing the warning, e.g. `@trigger_error`. This allows the deprecated warning to be detected but it will not break the application. Here are some examples:

 - [Guzzle](https://github.com/guzzle/guzzle/blob/ec70becf65a682371e9e10c3156eb01dc5fb4d26/src/Exception/BadResponseException.php#L20)
 - [Symfony](https://github.com/symfony/symfony/blob/61da48799c2ce020a72915311d053ec4557d6923/src/Symfony/Component/DependencyInjection/TypedReference.php#L32)
 - [Twig](https://github.com/twigphp/Twig/blob/ed1ce4d0d11abd68dd6e3f97d78ceb6e5ca238e5/lib/Twig/Function.php#L38)

I will trigger a silenced warning for now. LMK what you think.

TeBoring: I was thinking just ```@deprecated```

php/tests/generated_class_test.php
TeBoring: ok, just leave it here for now.

src/google/protobuf/compiler/php/php_generator.cc
bshaffer: It's not about if the classname needs a prefix, it's about whether to add the prefix even if it's specified. This is because this function is used to escape PHP reserved names, and there are cases in the namespaces (e.g. `PBEmpty\PBEcho\PrefixTestPrefix`) where we want a prefix to prevent a reserved name, but we do not want the PHP prefix added.

This is not actually related to the nested namespace enhancement, and is actually a separate bug fix for escaping namespaces.

src/google/protobuf/compiler/php/php_generator.cc
bshaffer: Good catch!

php/src/Google/Protobuf/Timestamp.php
bshaffer: Added them back!

php/src/Google/Protobuf/Field_Cardinality.php
bshaffer: Okay, I will do that also. Let me know if you'd like me to remove the `@trigger_error`

TeBoring:
I think in php_generattor.cc, you need to replace ```FullClassName``` with ```LegacyFullClassName``` in several places in case depended message is still in old generated code.

php/tests/generated_class_test.php
```diff
@@ -243,19 +246,12 @@ public function testLegacyNestedEnum()
...
-     * @runInSeparateProcess
-     * @preserveGlobalState disabled
-     */
     public function testLegacyTypehintWithNestedEnums()
     {
-        if (extension_loaded('protobuf')) {
-            $this->markTestSkipped('Protobuf extension fails when phpunit runs in a separate process');
-        }
-        $this->legacyEnum(new NestedEnum);
+        $this->legacyEnum(new TestLegacyMessage\NestedEnum);
```
TeBoring: Why not simply set new TestLegacyMessage_NestedEnum?

php/tests/generated_class_test.php
```diff
@@ -421,19 +417,12 @@ public function testLegacyMessageField()
...
-     * @runInSeparateProcess
-     * @preserveGlobalState disabled
-     */
     public function testLegacyTypehintWithNestedMessages()
     {
-        if (extension_loaded('protobuf')) {
-            $this->markTestSkipped('Protobuf extension fails when phpunit runs in a separate process');
-        }
-        $this->legacyMessage(new Sub);
+        $this->legacyMessage(new TestLegacyMessage\NestedMessage);
```
TeBoring: Use TestLegacyMessage_NestedMessage

php/tests/generated_class_test.php
TeBoring: I think you just need ```new TestLegacyMessage_NestedEnum```

TeBoring:
I think in tests, we should also reference old name.
Otherwise, in compatibility test, we have no way to use old protoc to build test generated code and pass tests.

TeBoring:
Currently, it still generate ```GPBMetadata/Proto/Empty/Echo.php``` for empty/echo.proto

bshaffer:
[here is an example of the deprecated class file and message](https://gist.github.com/bshaffer/906dcfbf1e05cd44ac74f0279e185ad3). I would love to get @dwsupplee and @michaelbausor 's thoughts on this!

michaelbausor:
What is the purpose of this line: https://gist.github.com/bshaffer/906dcfbf1e05cd44ac74f0279e185ad3#file-testmessage_sub-php-L7
Is it just to make sure that class gets autoloaded?

bshaffer:
@michaelbausor Yes, it ensures that if someone creates the deprecated class, the new class is loaded and the call to `class_alias(new::class, old::class)` is executed. So it allows a user's code to continue using the old class name, but it issues a deprecated warning and will show up as @deprecated in the IDE

TeBoring:
@michaelbausor and @dwsupplee, if you have other issues, we can fix them in another PR.

PR #4537, opened 2018-04-19
use brew install instead of easy_install in OSX

anandolee:


Mizux:
Better to use:
```sh
brew upgrade python;
python -m pip install -q tox;
```
This will guaranty that tox is installed for the correct `python` cli interpreter

src: https://github.com/google/or-tools/blob/master/.travis/install.sh#L60
log: https://travis-ci.org/google/or-tools/jobs/369083296#L2351
log: https://travis-ci.org/google/or-tools/jobs/369083296#L2436

xfxyjwf:
Jie, can you comment on @Mizux's suggestion?

anandolee:
It is not working because OSX does not have pip:
https://travis-ci.org/google/protobuf/jobs/371294220

@Mizux any idea?

Mizux:
Hi,

my 2 cents,
First you `brew upgrade python` OK
log: https://travis-ci.org/google/protobuf/jobs/371294220#L1452

notice the caveat:
https://travis-ci.org/google/protobuf/jobs/371294220#L3826
> Unversioned symlinks `python`, `python-config`, `pip` etc. pointing to
`python3`, `python3-config`, `pip3` etc., respectively, have been installed into
  /usr/local/opt/python/libexec/bin

but your failure pip call use.... `/usr/bin/python: No module named pip`

Solutions:
* use `python3 -m pip install tox`
* add `/usr/local/opt/python/libexec/bin` to your `PATH`, then `python -m pip` should work
* use `pip3 install tox`

IMHO modify the path is the best solution so if later part of the script call "python" they will use the same version (i.e python3)...

ps: You could run `which python`, `which python3` and `echo $PATH` to debug also
pps: 
```sh
ls -l /usr/local/opt/python/libexec/bin
...
pip -> ../../bin/pip3
python -> ../../Frameworks/Python.framework/Versions/3.6/bin/python3.6
```

anandolee:
"IMHO modify the path is the best solution so if later part of the script call "python" they will use the same version (i.e python3)..." But our test should cover different python versions

tests.sh
```diff
@@ -226,10 +226,11 @@ internal_install_python_deps() {
   if [ "$TRAVIS" != "true" ]; then
     return;
   fi
-  # Install tox (OS X doesn't have pip).
+  # Install 
+  (OS X doesn't have pip).
```
Mizux: Missing "#" add the beginning of the line

ps: You should squash (i.e. fixup) all your commits to have atomic change and all commit which compile (so you can fast forward on master instead of merging if you want...)

anandolee: Fixed. The tests are now passing, thanks for your help!

Will squash into on commit

PR #4550, opened 2018-04-25
CMake: Update CXX Standard management

Mizux:
Add -std=c++11 using CMake default target properties...
It follows #4475 and #4509

/!\ THIS WILL BUMP MINIMUM CMAKE VERSION TO 3.1.3 /!\

ref:
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html
https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Minimum CMake required
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.1.3)
 
 if(protobuf_VERBOSE)
   message(STATUS "Protocol Buffers Configuring...")
@@ -15,8 +15,10 @@ endif()
 # Project
 project(protobuf C CXX)
 
-# Add c++11 flags for clang
-set(CMAKE_CXX_FLAGS "-std=c++11")
+# Add c++11 flags
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Options
 option(protobuf_BUILD_TESTS "Build tests" ON)


xfxyjwf:
Thanks for the patch!

jtattermusch:
@xfxyjwf 

I think that increasing the required cmake version just to get access to the CMAKE_CXX_STANDARD cmake feature is a pretty bad idea.
cmake 3.1.3 is pretty new and is not available in many widely linux distributions - so anyone using e.g. ubuntu 14.04 (cmake 2.8) and debian jessie (has cmake 3.0.2)  user won't be able to build using cmake (unless they compile cmake from source, which major friction). I expect similar issues in other major distros.

I think the right way to approach this is:
- don't bump the required cmake version
- detect version of cmake and only use the "CMAKE_CXX_STANDARD 11" when available.
- use a workaround for older cmake versions -  e.g. just doing https://github.com/grpc/grpc/blob/3d5de335223ccf5994603193c52a6e5d3323f4b0/CMakeLists.txt#L125 will work just fine for most users.

Mizux:
1) Ubuntu 14.04 has a [cmake 3.5.1 package](https://packages.ubuntu.com/trusty/cmake3). and (travis-CI which is based on 14.04 image has cmake 3.9...)

2) I would say barely any active/widely available distro has at least cmake 3.5.1
[Packaging status](https://repology.org/metapackage/cmake)

I would also add:
- android-sdk has cmake 3.6 embedded (fun fact: cmake 3.7 has builtin support for android toolchain ^^;) 
- for embedded world aka Yocto (de facto standard) you have cmake-native 3.3.1 since [jethro (Yocto project 2.0)](http://layers.openembedded.org/layerindex/branch/jethro/recipes/?q=cmake-native)

3) dubious: Do you have some use case where people have a old cmake but want protobuf from source (>= 3.5.2) using cmake based build ?
I mean they can still rely on autotools if they like living in the past (which is BTW what you describe in your [installation instruction](https://github.com/google/protobuf/blob/master/src/README.md))

note: also bumping the required version to `3.0 >=` will half fix #4620 by using `@rpath/` as install name prefix by default. (e.d. CMake default behaviour for policy is based on minimum version required, i.e. cmake 3.11 behave like cmake 3.1.3 if you set `cmake_minimum_required(VERSION 3.1.3)`)

ps: for a full-featured Modern CMake I would push for 3.2 (or 3.5) but it's a little bit too new... beside there is good call at [CppCon 2017](https://youtu.be/eC9-iRN2b04) and [C++Now 2017](https://youtu.be/bsXLMQ6WgIk) on Modern CMake, hope you'll "live at head" with OSS community for CMake usage...

PR #4553, opened 2018-04-25
Set ext.no_native = true for non mac platform

pherl:
From:
https://github.com/rake-compiler/rake-compiler/issues/146#issuecomment-368539245

### DIFF
--- a/ruby/Rakefile
+++ b/ruby/Rakefile
@@ -52,6 +52,12 @@ if RUBY_PLATFORM == "java"
   end
 else
   Rake::ExtensionTask.new("protobuf_c", spec) do |ext|
+    unless RUBY_PLATFORM =~ /darwin/
+      # TODO: also set "no_native to true" for mac if possible. As is,
+      # "no_native" can only be set if the RUBY_PLATFORM doing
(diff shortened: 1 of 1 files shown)


ruby/Rakefile
```diff
@@ -52,6 +52,12 @@ if RUBY_PLATFORM == "java"
   end
 else
   Rake::ExtensionTask.new("protobuf_c", spec) do |ext|
+    unless RUBY_PLATFORM =~ /darwin/
+      # TODO: also set "no_native to true" for mac if possible. As is,
+      # "no_native" can only be set if the RUBY_PLATFORM doing
+      # cross-compilation is contained in the "ext.cross_platform" array.
+      ext.no_native = true
+    end
```
apolcyn: I <i>think</i> this also needs

```
ext.cross_compiling do |spec|
  s.files       = Dir.glob('lib/**/*.rb')
end
```

to get keep the .rb files in the binary packages, without the .c files.

apolcyn: scratch my above comment. That is optional, and is only useful if you want to also want to remove source code from the binary packages.

Issue #3781, opened 2017-10-22
<Java> JDK 9 warning: Illegal reflective access by com.google.protobuf.UnsafeUtil to field java.nio.Buffer.address


xenji:
G'day!

While testing one of our projects against JDK 9, the message below was printed to stderr. As it the message suggests, I'm now going on your nerves ;).

I've collected a few system details below and also a link to a file in your repo where I suspect the candidate that it complains about. If you can point me in a direction of how to solve it, I'm happy to contribute a PR.

As far as I can judge, the error message refers to this line here:
 [java/core/src/main/java/com/google/protobuf/UnsafeUtil.java#L378](https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/UnsafeUtil.java#L378)

----

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/Users/mmueller/.gradle/caches/modules-2/files-2.1/com.google.protobuf/protobuf-java/3.4.0/b32aba0cbe737a4ca953f71688725972e3ee927c/protobuf-java-3.4.0.jar) to field java.nio.Buffer.address
WARNING: Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

<details>
<summary>System specs:</summary>

```
System:
Darwin Brick 17.0.0 Darwin Kernel Version 17.0.0: Thu Aug 24 21:48:19 PDT 2017; root:xnu-4570.1.46~2/RELEASE_X86_64 x86_6

Java:
java version "9.0.1"
Java(TM) SE Runtime Environment (build 9.0.1+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.1+11, mixed mode)

Gradle:
------------------------------------------------------------
Gradle 4.2.1
------------------------------------------------------------

Build time:   2017-10-02 15:36:21 UTC
Revision:     a88ebd6be7840c2e59ae4782eb0f27fbe3405ddf

Groovy:       2.4.12
Ant:          Apache Ant(TM) version 1.9.6 compiled on June 29 2015
JVM:          9.0.1 (Oracle Corporation 9.0.1+11)
OS:           Mac OS X 10.13 x86_64
```

</details>

pherl:
[...]

PR #4557, opened 2018-04-26
Make the use of Unsafe conditional

eed3si9n:
Ref #3781

This is an attempt to workaround the Unsafe warning on JDK 9+.


eed3si9n:
I signed it!

xfxyjwf:
@eed3si9n Thanks for the patch, but I don't think we can just disable unsafe like this as we have code paths relying on the use of unsafe. I'm working on integrating our internal Java changes to github right now and we should fix the Java 9 issue after all the changes are integrated.

xfxyjwf:
I'm closing this PR for now.

eed3si9n:
@xfxyjwf Could you update the status on #3781 please?
What's the expected schedule for the integration?

PR #4576, opened 2018-04-29
Install gcc 4.8 package on Travis build machine

jozefizso:
Newly created travis machines uses docker without **sudo**, so `tests.sh` will fail when installing gcc manualy. This ensures gcc 4.8 is installed on travis machine for testing **csharp** language.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,10 +65,12 @@ matrix:
       addons:
         apt:
           sources:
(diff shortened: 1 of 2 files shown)


xfxyjwf:
@jozefizso Thanks for the patch but we no longer use travis. I'll delete travis related configuration in our repo after a few weeks.

PR #4586, opened 2018-05-02
MinGW build failed

chronoxor:


src/google/protobuf/stubs/mathlimits.h
```diff
@@ -43,6 +43,11 @@
 #ifndef UTIL_MATH_MATHLIMITS_H__
 #define UTIL_MATH_MATHLIMITS_H__
 
+// MinGW build requires __STRICT_ANSI__ to be undefined to use _finite(), _isnan(), _fpclass() functions
```
xfxyjwf: Hmm, can you change _finite to std::isfinite() instead?

chronoxor: Fixed!

Issue #4462, opened 2018-04-02
Lots of issues with Timestamp's use of DateTime


atrauzzi:
Noticing quite a few problematic abstractions around how protobuf's PHP implementation works.

![image](https://user-images.githubusercontent.com/115199/38178785-c1c67148-35df-11e8-9a3a-c91700fd7b8f.png)

If I use the `DateTime` utility method to set the value of a `Timestamp`, most editors will complain because the docblock hint isn't fully qualified.

![image](https://user-images.githubusercontent.com/115199/38178811-38053092-35e0-11e8-9020-9e20e6dec35e.png)

Also, because this library performs a naive type check, my instance has to be a `\DateTime` and nothing else:

![image](https://user-images.githubusercontent.com/115199/38178922-12cb90bc-35e2-11e8-99d5-699da217ad6d.png)

_([`Carbon` subclasses the native `DateTime`](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php#L55), so it still technically qualifies.)_

I see two possible solutions.  The first - and best - is to add a type hint to the function signature in PHP code.  This would eliminate the need to do the naive check.
The second is to change `DateTime` to `\DateTime` in the docblock so that it's off the root namespace rather than relative to the current namespace.  But that doesn't help with any subclasses of `\DateTime`, so I don't consider it a worthwhile solution.

Also, I'd suggest renaming this method to something like `setFrom`, `setDateTime` or `setFromDateTime` as `fromDateTime` sounds more like a static factory method that would return a new instance, not mutate state.

Which might not be a bad idea to add as well (`Timestamp::fromDateTime`)...

TeBoring:
1) type hint of class is not supported in php5. We can add it after we drop support for php5
2) use instanceof instead of comparing class name?
3) We don't want to change the method name at this point.

atrauzzi:
1. [Not true.](https://secure.php.net/manual/it/language.oop5.typehinting.php).  It's been supported since `5`.  Moreover, please see the [supported versions page](https://secure.php.net/supported-versions.php). There's no point in supporting anything older than `5.6` at present and anyone who is not using at least `5.6` today is no longer getting security updates.

![image](https://user-images.githubusercontent.com/115199/38222233-200bf02e-36a9-11e8-81e4-7ac2ae973e2d.png)

2. Given the above, I suggest using the type hint for `DateTime` in the method signature.  This allows the language to naturally handle the type check for you.

[...]

PR #4601, opened 2018-05-08
Avoid direct check of class name

TeBoring:
So that subclasses can also be ok. Check #4462

### DIFF
--- a/php/src/Google/Protobuf/Timestamp.php
+++ b/php/src/Google/Protobuf/Timestamp.php
@@ -161,12 +161,8 @@ class Timestamp extends \Google\Protobuf\Internal\Message
      *
      * @param \DateTime $datetime
      */
-    public function fromDateTime($datetime)
+    public function fromDateTime(\DateTime $datetime)
     {
-        if ($datetime instanceOf \DateTime) {
-            trigger_error("Given parameter is not a DateTime.",
-                          E_USER_ERROR);
-        }
         $this->seconds = $datetime->format('U');
         $this->nanos = 0;
     }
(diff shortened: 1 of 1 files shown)


php/src/Google/Protobuf/Timestamp.php
```diff
@@ -159,11 +159,11 @@ public function setNanos($var)
     /**
      * Converts PHP DateTime to Timestamp.
      *
-     * @param DateTime $datetime
+     * @param \DateTime $datetime
      */
     public function fromDateTime($datetime)
     {
-        if (get_class($datetime) !== \DateTime::class) {
+        if ($datetime instanceOf \DateTime) {
```
atrauzzi: I realize it's not case sensitive, but if you want to stick with community convention, it's `instanceof`, not `instanceOf`.

php/src/Google/Protobuf/Timestamp.php
```diff
@@ -159,11 +159,11 @@ public function setNanos($var)
     /**
      * Converts PHP DateTime to Timestamp.
      *
-     * @param DateTime $datetime
+     * @param \DateTime $datetime
      */
     public function fromDateTime($datetime)
```
atrauzzi: Instead of doing the `if` below, you could just do:

```
public function fromDateTime(DateTime $dateTime)
```

?

php/src/Google/Protobuf/Timestamp.php
TeBoring: Done

php/src/Google/Protobuf/Timestamp.php
TeBoring: Done

PR #4604, opened 2018-05-09
Add C++ library target for including python_protobuf.h

ThomasColthurst:
Hi!  I'm Thomas Colthurst, and I work on the Nucleus project (github.com/google/nucleus).

For that project, I've written a custom Clif_PyObjAs converter that bypasses the expensive copies that Clif normally imposes when passing protocol buffers between C++ and Python.  But that function needs to call MutableCProtoInsidePyProto, and to do that, I need that function to be exported by some cc_library target.  Thus this pull request.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -793,6 +793,16 @@ py_library(
     ],
 )
 
+cc_library(
+    name = "python_protobuf",
+    hdrs = [
+        "python/google/protobuf/python_protobuf.h",
+    ],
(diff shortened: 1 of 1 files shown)


ThomasColthurst:
Ping?

xfxyjwf:
Please follow-up in cl/150482556 instead. We plan to delete this python_protobuf.h file.

@anandolee

PR #4608, opened 2018-05-10
Add "override" keyword for overridden virtual functions

khuzemap:
Add PROTOBUF_OVERRIDE macro, which is defined to be
"override" for c++11 and nothing for older versions.
PROTOBUF_OVERRIDE is added for overridden virtual
functions.
Please refer following issue for discussion on this.
https://github.com/google/protobuf/issues/67

### DIFF
--- a/src/google/protobuf/compiler/importer.h
+++ b/src/google/protobuf/compiler/importer.h
@@ -96,12 +96,13 @@ class LIBPROTOBUF_EXPORT SourceTreeDescriptorDatabase : public DescriptorDatabas
   }
 
   // implements DescriptorDatabase -----------------------------------
-  bool FindFileByName(const string& filename, FileDescriptorProto* output);
+  bool FindFileByName(const string& filename,
+                      FileDescriptorProto* output) override;
   bool FindFileContainingSymbol(const string& symbol_name,
-                                FileDescriptorProto* output);
+                                FileDescriptorProto*output) override;
   bool FindFileContainingExtension(const string& containing_type,
                                    int field_number,
-                                   FileDescriptorProto* output);
+                                   FileDescriptorProto* output) override;
 
  private:
   class SingleFileErrorCollector;
@@ -119,13 +120,13 @@ class LIBPROTOBUF_EXPORT SourceTreeDescriptorDatabase : public DescriptorDatabas
                   const string& element_name,
                   const Message* descriptor,
                   ErrorLocation location,
-                  const string& message);
+                  const string& message) override;
 
     virtual void AddWarning(const string& filename,
                             const string& element_name,
                             const Message* descriptor,
                             ErrorLocation location,
-                            const string& message);
+                            const string& message) override;
 
    private:
     SourceTreeDescriptorDatabase* owner_;
@@ -293,9 +294,9 @@ class LIBPROTOBUF_EXPORT DiskSourceTree : public SourceTree {
   bool VirtualFileToDiskFile(const string& virtual_file, string* disk_file);
 
   // implements SourceTree -------------------------------------------
-  virtual io::ZeroCopyInputStream* Open(const string& filename);
+  virtual io::ZeroCopyInputStream* Open(const string& filename) override;
 
-  virtual string GetLastErrorMessage();
+  virtual string GetLastErrorMessage() override;
 
  private:
   struct Mapping {
--- a/src/google/protobuf/io/zero_copy_stream_impl.h
+++ b/src/google/protobuf/io/zero_copy_stream_impl.h
@@ -87,24 +87,24 @@ class LIBPROTOBUF_EXPORT FileInputStream : public ZeroCopyInputStream {
   int GetErrno() { return copying_input_.GetErrno(); }
 
   // implements ZeroCopyInputStream ----------------------------------
-  bool Next(const void** data, int* size);
-  void BackUp(int count);
-  bool Skip(int count);
-  int64 ByteCount() const;
+  bool Next(const void** data, int* size) override;
+  void BackUp(int count) override;
+  bool Skip(int count) override;
+  int64 ByteCount() const override;
 
  private:
   class LIBPROTOBUF_EXPORT CopyingFileInputStream : public CopyingInputStream {
    public:
     CopyingFileInputStream(int file_descriptor);
-    ~CopyingFileInputStream();
+    ~CopyingFileInputStream() override;
 
     bool Close();
     void SetCloseOnDelete(bool value) { close_on_delete_ = value; }
     int GetErrno() { return errno_; }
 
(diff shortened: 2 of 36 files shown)


xfxyjwf:
@khuzemap I just commented on the issue: we are allowing C++11 features in protobuf master now. Could you update this PR accordingly?

src/google/protobuf/compiler/cpp/cpp_message.cc
```diff
@@ -1121,12 +1122,12 @@ GenerateClassDefinition(io::Printer* printer) {
   if (HasGeneratedMethods(descriptor_->file(), options_)) {
     if (HasDescriptorMethods(descriptor_->file(), options_)) {
       printer->Print(vars,
-        "void CopyFrom(const ::google::protobuf::Message& from) PROTOBUF_FINAL;\n"
-        "void MergeFrom(const ::google::protobuf::Message& from) PROTOBUF_FINAL;\n");
+        "void CopyFrom(const ::google::protobuf::Message& from) PROTOBUF_FINAL PROTOBUF_OVERRIDE;\n"
```
xfxyjwf: From what I understand you don't need "override" if you already have "final". Doesn't "final" imply "override" already?

khuzemap: No. "override" and "final" have independent behavior. "final" means the function cannot be overridden, however, "override" means the function is overriding a base class function. If a function has both "override" and "final" it means, the function is overriding a base class function but cannot be overriden.

xfxyjwf: I believe "final" has the exactly same semantics as "final override", i.e., it's overriding a virtual method in a base class and it can no longer be overridden.  In any case, please follow our C++ style guide:
https://google.github.io/styleguide/cppguide.html#Inheritance

Specifically:
```
Explicitly annotate overrides of virtual functions or virtual destructors with exactly one of an override or (less frequently) final specifier
```

khuzemap: we use gcc "Wsuggest-override" warning option for compile time enforcement and that doesn't take "final" as replacement of "override". clang has better option called "Winconsistent-missing-override", but unfortunately not available on gcc yet.

xfxyjwf: I think we should follow the style guide here. I just checked Google's internal code base and there is pretty much 0 usage of "final override" and "override final". Using it in protobuf code would require an exception granted by the style guide arbiters but I don't think it's justified for this case. Eventually gcc will have the same option, won't it?

xfxyjwf:
@khuzemap Do you plan to update the PR to conform to Google's style guide? (i.e., use only one of "final" and "override").

khuzemap:
@xfxyjwf , sorry for not pushing it early. have updated the change with only one of override or final. please review. Thanks!

src/google/protobuf/io/zero_copy_stream_impl.h
```diff
@@ -67,6 +67,7 @@ class LIBPROTOBUF_EXPORT FileInputStream : public ZeroCopyInputStream {
   // should be read and returned with each call to Next().  Otherwise,
   // a reasonable default is used.
   explicit FileInputStream(int file_descriptor, int block_size = -1);
+  ~FileInputStream() override = default;
```
xfxyjwf: Why is this needed?

src/google/protobuf/map_field.h
```diff
@@ -311,27 +312,28 @@ class LIBPROTOBUF_EXPORT DynamicMapField: public TypeDefinedMapFieldBase<MapKey,
...
  private:
   Map<MapKey, MapValueRef> map_;
   const Message* default_entry_;
 
   // Implements MapFieldBase
-  void SyncRepeatedFieldWithMapNoLock() const;
-  void SyncMapWithRepeatedFieldNoLock() const;
-  size_t SpaceUsedExcludingSelfNoLock() const;
-  void SetMapIteratorValue(MapIterator* map_iter) const;
+
```
xfxyjwf: Remove this empty line

src/google/protobuf/compiler/importer.h
```diff
@@ -119,13 +120,13 @@ class LIBPROTOBUF_EXPORT SourceTreeDescriptorDatabase : public DescriptorDatabas
...
                   ErrorLocation location,
-                  const string& message);
+                  const string& message) override;
 
     virtual void AddWarning(const string& filename,
-                            const string& element_name,
-                            const Message* descriptor,
-                            ErrorLocation location,
-                            const string& message);
+                    const string& element_name,
```
xfxyjwf: Revert this alignment change.

khuzemap: Done

khuzemap: Done

src/google/protobuf/map_field.h
khuzemap: Done

src/google/protobuf/io/zero_copy_stream_impl.h
khuzemap: removed.

xfxyjwf:
The MacOS C++ test is failing:
```
./google/protobuf/map_entry_lite.h:275:10: error: 'SpaceUsedLong' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  size_t SpaceUsedLong() const {
```

Can you fix that?

khuzemap:
MapEntryImpl is templated and instantiated with its Base class as Message and MessageLite. Message class defines SpaceUsedLong as virtual but MessageLite class don't. So enforcing 'override' here fails when MapEntryImpl is instantiated with MessageLite. Since there was enforcement of 'override', I didn't add any special handling. If "override" is enforced on code base, then for this function, it needs to be waived/ignored. In our code we use "-Wsuggest-override" to enforce "override" so I have added code to ignore that. For macOS build, it seems -Winconsistent-missing-override is used, so it needs to be ignored. of course other option is not to use any compile time switch as there are more than one ways of doing it.

xfxyjwf:
Can we move that SpaceUsedLong() method from MapEntryImpl to MapEntry?

khuzemap:
Was able to move. here is the diff.

xfxyjwf:
@khuzemap Can you update this pull request?

khuzemap:
yes, done.

xfxyjwf:
I'm rerunning the tests.

daquexian:
This "Down-integrate internal changes to github" commit https://github.com/protocolbuffers/protobuf/commit/624a40a387e1b2da4c6bbde44ad3d0a5efb9879b reverted this change :(

PR #4620, opened 2018-05-14
CMake OSX rpath  management

Mizux:
CMake has introduced [CMP0042](https://cmake.org/cmake/help/latest/policy/CMP0042.html) in CMake 3.0

Context:
I'm currently moving google/or-tools from static dependencies to dynamic one, unfortunately protobuf don't use `@rpath/` as prefix path contrary to gflags and glog so it is difficult for me to recreate our sdk since libprotobuf.dylib contains absolute (build) path...

On MacOS library need to find itself and by default they use an absolute path pointing to the build dir (i.e. everyhing blow up at install and it is no relocatable ;)). Like on Unix world, you can use rpath and $ORIGIN like.
- First use `@rpath/` as --install-name prefix (done by enabling CMP0042)
- then you could add some `@loader_path;@loader_path/../lib;etc` to LC_RPATH list (cf [working example](https://github.com/Mizux/cmake-cpp/blob/master/FooBarApp/CMakeLists.txt#L11) if you are curious..)
note: `@loader_path` is the equivalent to `$ORIGIN`
note2: contrary to UNIX it seems you need several calls to  -Wl,-rpath, i.e. one path at a time while on unix you can add several path using colon ':' (on CMake with RPATH_INSTALL you can use item separator semicolon ';' so cmake will add several calls (otools -l will show you all LC_RPATH entries))

**WARNING**: On MacOS `protoc` depends on `libprotoc.dylib` and `libprotobuf.dylib`,  so while it's work in build dir since cmake add an `LC_RPATH`, **BUT** it's not working if using `make install DESTDIR=foo` with:
foo/usr/local:
 + /bin
      - protoc
 + /lib
     - libprotoc.dylib
     - libprotobuf.dylib

i.e. cmake remove build dir rpath (good) but doesn't add a `@loader_path/..` entry (bad but expected), working on it...
EDIT: my last commit add the correct rpath -> everything seems to work and are relocatable
ToDo: check if this fix the python setup.py use of protoc cf https://github.com/google/protobuf/blame/master/python/README.md#L60

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -7,7 +7,10 @@ endif()
 
 # CMake policies
 cmake_policy(SET CMP0022 NEW)
-
+# On MacOS use @rpath/ for target's install name prefix path
+if (POLICY CMP0042)
+  cmake_policy(SET CMP0042 NEW)
+endif ()
 # Clear VERSION variables when no VERSION is given to project()
 if(POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -8,6 +8,7 @@ endif()
 # CMake policies
 cmake_policy(SET CMP0022 NEW)
 
+# Clear VERSION variables when no VERSION is given to project()
 if(POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif()
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -15,6 +15,13 @@ foreach(_library ${_protobuf_libraries})
     PROPERTY INTERFACE_INCLUDE_DIRECTORIES
     $<BUILD_INTERFACE:${protobuf_source_dir}/src>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+  if (UNIX AND NOT APPLE)
+    set_property(TARGET ${_library}
+      PROPERTY INSTALL_RPATH "$ORIGIN")
+  elseif (APPLE)
+    set_property(TARGET ${_library}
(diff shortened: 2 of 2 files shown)


Mizux:
Actually rpath management seems borken with cmake install rules...
[LINUX protocol]
Building
```sh
cmake -Hcmake -Bbuild -DBUILD_SHARED_LIBS=ON
cmake --build build
```
Analyzing libprotoc.so (which depends on libprotobuf.so)
```sh
ldd build/libprotoc.so
	[...]
	libprotobuf.so.3.5.2 => /usr/local/google/home/corentinl/work/protobuf/build/libprotobuf.so.3.5.2 (0x00007f8455ec2000)
objdump -p build/libprotoc.so
	[...]
	Dynamic Section:
	NEEDED               libprotobuf.so.3.5.2
	[...]
	RUNPATH              /usr/local/google/home/corentinl/work/protobuf/build:
```

Then lets check the install
```sh
cmake --build build --target install -- DESTDIR=install
ldd build/install/usr/local/lib/libprotoc.so
	[...]
	libprotobuf.so.3.5.2 => not found
objdump -p build/install/usr/local/lib/libprotoc.so
	[...]
	Dynamic Section:
	NEEDED               libprotobuf.so.3.5.2
	[...]
```
there is no RPATH specified, which is the correct behaviour since CMake remove build rpath during install to replace them install RPATH wich is currently empty

I'm working on it, hope to have something working in few hours !

xfxyjwf:
Let me know if this PR is ready to review.

Mizux:
Hi,

Sorry, rpath install has been fixed by the third patch.
Now you can install (and use) protobuf in an unconventional dir (e.g. workspace, python site package) without having to modify your LD_LIBRARY_PATH/DYLD_LBRARY_PATH

PR #4622, opened 2018-05-14
Adopt php_metadata_namespace in php code generator

TeBoring:
The php_metadata_namespace is corresponded to the relative directory of
the metadata file. e.g., previously, the metadata file of foo.proto was
GPBMetadata/Foo.php. If the php_metadata_namespace is "Metadata\\Bar",
the metadata file will be Metadata/Bar/Foo.php.


TeBoring:
The broken appveyor test is not related to this change.

PR #4624, opened 2018-05-14
Fix undefined behavior in hash function.

eughermann:
Remove signed integer overflow.
Signed integer overflow creates undefined behavior that may lead to unpredictable fails on different platforms.
One known example of the hardware where this code did fail is Apple A6 (32-bit Apple Swift CPU)

Hash function updated.
16777619, 16777499 - two prime numbers that typically used to get better dispersion.


eughermann:
Looks like I shouldn't need to sign CLA.

xfxyjwf:
@acozzette Adam, is this the issue you mentioned in today's team meeting?

acozzette:
@xfxyjwf Yes, this is the same one.

PR #4627, opened 2018-05-15
Adopt ruby_package in ruby generated code.

TeBoring:


TeBoring:
Broken appveyor test is not related to this change.

PR #4634, opened 2018-05-16
Fix java benchmark bug, fix python lib cache

BSBandme:


benchmarks/java/pom.xml
```diff
@@ -14,7 +14,10 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>3.5.0</version>
+	  <version>1.0</version>
```
xfxyjwf: This seems wrong?

benchmarks/java/src/main/java/com/google/protobuf/ProtoCaliperBenchmark.java
```diff
@@ -22,6 +23,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+@VmOptions("-XX:-TieredCompilation")
```
xfxyjwf: Why do we need this?

BSBandme: It seems a bug for caliper that it set CICompilerCount to 1, for TieredCompilation this value must be at least 2. So I just disable TieredCompilation.

benchmarks/java/pom.xml
BSBandme: Then what version number should I set?

BSBandme: I think the benchmark should use protobuf-java in the same repo not just for a specific version?

benchmarks/java/src/main/java/com/google/protobuf/ProtoCaliperBenchmark.java
xfxyjwf: So it fails on kokoro but not locally? Can we set some environment variable to make it also work on kokoro? I'm worried that disabling TieredCompilation would change the performance characteristics and make the benchmark result less meaningful.

benchmarks/java/pom.xml
xfxyjwf: In that case, can you use something like "<version>${protobuf.version}</version>" here and pass "-Dprotobuf.version=..." when invoking mvn in the Makefile?

benchmarks/java/src/main/java/com/google/protobuf/ProtoCaliperBenchmark.java
BSBandme: I think it fail on kokoro is the version of java is different from local one. In fact caliper's fix after 1.0-beta-2 for that is just [disabling TieredCompilation](https://github.com/google/caliper/blob/97a906bd6d91e9e0a30d413966b3a11469cb9ebf/caliper-runner/src/main/java/com/google/caliper/runner/target/Jvm.java#L39).  I think it's OK to do that

benchmarks/java/pom.xml
BSBandme: Done

benchmarks/java/pom.xml
```diff
@@ -14,7 +14,10 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>3.5.0</version>
+	  <version>${protobuf.version}</version>
```
xfxyjwf: Fix the indentation.

benchmarks/java/src/main/java/com/google/protobuf/ProtoCaliperBenchmark.java
xfxyjwf: Please add comments for why this is needed and link the caliper source as well.

benchmarks/java/pom.xml
BSBandme: Done

benchmarks/java/src/main/java/com/google/protobuf/ProtoCaliperBenchmark.java
BSBandme: Done

PR #4642, opened 2018-05-17
C# Proto2 feature : Field presence and default values

ObsidianMinor:
This adds field presence and default values to the C# protoc compiler and library. This also includes some fixes for existing hacks in reflection that were created due to only supporting proto3 at the time of creation.

This PR also includes breaking changes. For required fields to work properly, a method is added the IMessage interface and breaks backwards compatibility with previous versions.

This is the first PR in a set of four PRs implementing all the changes required to support proto2 in the C# library and compiler.

---
This change is [Reviewable](https://reviewable.io/reviews/google/protobuf/4642)


xfxyjwf:
Assigned to @anandolee to view the code. I also linked it to https://reviewable.io/reviews/google/protobuf/4642#- because the code review function of github is horrible.

xfxyjwf:
Ping @anandolee

ObsidianMinor:
Any updates @anandolee?

src/google/protobuf/compiler/csharp/csharp_field_base.cc
```diff
@@ -72,29 +75,43 @@ void FieldGeneratorBase::SetCommonFieldVariables(
 
   (*variables)["property_name"] = property_name();
   (*variables)["type_name"] = type_name();
+  (*variables)["nullable_type_name"] = nullable_type_name();
```
anandolee: Only define it for proto2

ObsidianMinor:
I was going to add tests after I make the PRs for the other features. Since everything has already been implemented, after this gets pulled in I'll open the PR for extension support. And once that gets pulled in I'll open the PR for group support and all the tests for the other features since then they'd be all there. This would save work since to add support for this PR I'd have to create a custom tests proto and slowly add onto it with each PR rather than use the existing tests protos once.

csharp/src/Google.Protobuf/WellKnownTypes/Struct.cs
```diff
@@ -552,8 +604,13 @@ public enum KindOneofCase {
           }
         }
       }
+      pb::ProtoPreconditions.CheckMergedRequiredFields(this);
```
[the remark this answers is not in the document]
ObsidianMinor: It could be done that way in C++ since the merge method that is generated is a partial merge and is properly named for that. However in C# the merge method is not named for that. I could possibly make the generated MergeFrom and WriteTo methods not check for required fields and add an extension method that does check once (something like "CompleteMergeFrom" and "CompleteWriteTo"), but I thought it might be confusing to developers so I left it out.

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
```diff
@@ -89,6 +120,27 @@ void PrimitiveFieldGenerator::GenerateMembers(io::Printer* printer) {
   printer->Print(
     "  }\n"
     "}\n");
+  if (descriptor_->file()->syntax() == FileDescriptor::SYNTAX_PROTO2) {
+    printer->Print(variables_, "/// <summary>Gets whether the \"$descriptor_name$\" field is set</summary>\n");
+    AddPublicMemberAttributes(printer);
+    printer->Print(
+      variables_, 
+      "$access_level$ bool Has$property_name$ {\n"
```
anandolee: Field presence does not mean we need to provide has methods in each message. As C# can have nullable type, users can check if the field is set by themselves. C++ provide has methods because there is no nullable type and is using has bits that user can't access for field presence. It is a waste to produce has methods for each field which seems "same code". I'd prefer to ask user check null directly. 

Same issue for clear methods.

csharp/src/Google.Protobuf/WellKnownTypes/Struct.cs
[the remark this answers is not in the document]
ObsidianMinor: So do you want me to make the extensions or keep the duplicate calls?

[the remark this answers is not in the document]
anandolee: Is it possible to do like this? : 
1, Change the generated MergeFrom() to MergePartialFrom()
2, Add an "abstract class MessageBase : IMessage" or something for MergeFrom() that calls generated MergePartialFrom() and do the check. 
Then it will be same with C++

Note: I am not an expert in C#, my review is more based on protobuf features and other language implementations.

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
ObsidianMinor: Now that I review that code, I realize I shouldn't have used nullables in the first place. Nullables aren't exactly efficient and messages should use bitfields like the Java, Python, and C++ versions.

So I'll retain the Has and Clear methods and reimplement the underlying field presence to use something that doesn't use add four bytes of memory to each field.

csharp/src/Google.Protobuf/WellKnownTypes/Struct.cs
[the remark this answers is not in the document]
ObsidianMinor: Not exactly, since that would require changing the name of the method in the IMessage interface and would be an even bigger breaking change than just the addition of the IsInitialized method.

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
anandolee: Change to an efficient way sounds good to me. But four bytes of memory to each field looks like not good enough.  C++'s  hasbit array is one bit for each field:
https://github.com/google/protobuf/blob/f81a3910c07daba7d1f2fd57e328692fc8c480ca/src/google/protobuf/compiler/cpp/cpp_message.cc#L502

csharp/src/Google.Protobuf/WellKnownTypes/Struct.cs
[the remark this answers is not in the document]
anandolee: I double checked the behaviors in other languages:
1, Serialized do not need to check required fields. (C++ is doing a debug check, not required)
2. Parse should check required fields, Merge should not check unless there a MergePaitialFrom method. 

So we can go like this: Not change the generated code's MergeFrom and WriteTo for check. Change the MessageParser to do the check.

Does this make sense to you?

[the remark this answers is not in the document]
ObsidianMinor: Yes, that's much easier

ObsidianMinor:
Alright @anandolee, the proto2 check has been reintroduced and required handling has been changed to only be used for parsers.

Field presence currently uses ints for bit fields, however if you want I can make it use bytes instead of ints.

anandolee:
C++ is using one hasbit array for a message not has bytes for each field. Are you able to do like that?

ObsidianMinor:
It is, however it has some cost. When you make a dynamically sized array in C#, it's considered another object on the heap which the garbage collector has to keep track of. In C++, that hasbit array is statically sized and is part of the memory of the original class. C# doesn't have features like these except in unsafe code, which we can't be sure a consumer will have as it's disabled by default. So to keep the code from allocating another array every time an object is made, I made the ints part of the class itself.

ObsidianMinor:
@anandolee

anandolee:
Make sense to me. "I can make it use bytes instead of ints", I think bytes are better because they are name _hasBits in the code.

Another question, have you considered reflection usages for has method?

ObsidianMinor:
I have considered reflection usages, @anandolee, but I left it out since was unsure how to implement it since it'd have to be implemented it as well for proto3 and I don't know what a has method returns for proto3 fields.

I also take back what I said about using bytes instead of ints since bytes actually don't have bitwise operations, only ints and longs have them (as well as their unsigned counterparts). So each operation would convert the byte field to an int, then you'd have to cast back to an int for any assignment operations so that's two extra casts for each property set.

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -123,6 +130,10 @@ void MessageGenerator::Generate(io::Printer* printer) {
   printer->Print(
       "private pb::UnknownFieldSet _unknownFields;\n");
 
+  for (int i = 0; i < has_bit_field_count_; i++) {
```
anandolee: Thanks for explain. Please add comments here to explain why use ints here instead of bitsarray or bits.

src/google/protobuf/compiler/csharp/csharp_wrapper_field.cc
```diff
@@ -81,6 +81,25 @@ void WrapperFieldGenerator::GenerateMembers(io::Printer* printer) {
     "  set {\n"
     "    $name$_ = value;\n"
     "  }\n"
+    "}\n\n");
+
+  printer->Print(
```
anandolee: No need to add has method for message field in proto3. Users can compare to null.  I also think we should not add clear methods for proto3 in this PR because it is to support proto2 field presence. One of the proto3's goal is to reduce the generated code size. As users can use proto3 well with out clear method we can keep the current generated code. If you really want to add those methods, they should be another PR and need discussion for the cost and gain for decision.

csharp/src/AddressBook/Addressbook.cs
```diff
@@ -89,17 +89,28 @@ public sealed partial class Person : pb::IMessage<Person> {
 
     /// <summary>Field number for the "name" field.</summary>
     public const int NameFieldNumber = 1;
-    private string name_ = "";
+    /// <summary>Default value for the "name" field</summary>
+    public const string NameDefaultValue = "";
```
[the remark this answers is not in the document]
ObsidianMinor: I can use static readonly for all default values. But I don't think we should change the field numbers to be static readonly. Some people might use them in switch cases and switch cases only accept constant expressions (which constants are, `static readonly`s are not).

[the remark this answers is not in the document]
anandolee: Thanks! Let's change the default value then.  We can reconsider field numbers in future which is not related to you PR.

src/google/protobuf/compiler/csharp/csharp_wrapper_field.cc
ObsidianMinor: Should I also remove the default value fields from proto3 as well?

anandolee: It will be nice to keep proto3 not changed except we have to. What is the current behavior way for users to clear the field? Set to default value from the descriptor I guess?

ObsidianMinor:
Default values have been made `private static readonly` and have been removed from proto3. Clear methods have also been removed and field accessors have been adjusted accordingly @anandolee

ObsidianMinor:
I'll fix the distcheck. I'll also change how default string values are resolved since it seems codecvt isn't available.

ObsidianMinor:
@anandolee

anandolee:
Compatibility tests requires old generated code can work with new runtime. Can we remove IsInitliazed from IMessage interface?

ObsidianMinor:
Unless we want to wait until default interface implementations (which could take years or never), no. It's necessary for checking initialization in map fields and repeated fields, which is in turn necessary for checking initialization in messages.

anandolee:
Can we add an abstract class in between IMessage and Message?  Add new API on IMessage interface will not only break compatibility tests I believe, current users who implement their own message based on IMessage will also break.

ObsidianMinor:
Not exactly, since that'd changed the API of message parsers thus breaking compatibility tests again. @anandolee

ObsidianMinor:
I added a new interface simply called `IMessage2` since making an abstract class and inheriting from that would break code that inherits from another class in a separate partial definition.

ObsidianMinor:
@anandolee

ObsidianMinor:
@xfxyjwf could we run kokoro for this?

anandolee:
Sorry for late reply.  
It is great to see compatibility tests pass now. Will review your change sometime tomorrow.

csharp/src/Google.Protobuf/IMessage.cs
```diff
@@ -69,6 +69,18 @@ public interface IMessage
         MessageDescriptor Descriptor { get; }
     }
 
+    /// <summary>
+    /// Interface for a Protocol Buffers message, supporting 
+    /// proto2 operations required for serialization
+    /// </summary>
+    public interface IMessage2 : IMessage
```
anandolee: What if we want to add other API for IMessage in future?

ObsidianMinor: If we really, *really* don't want to add new methods to the IMessage interface or make new interface versions, I could create a reflection based extension method that checks if a message is initialized that way.

anandolee: "making an abstract class and inheriting from that would break code that inherits from another class in a separate partial definition." I don't understand this clear.

@jtattermusch do you have any suggestions how to add a new API to IMessage interface without break compatibility tests?  I suggested to add an abstract in between IMessage and generated message, but seems like it has some issue. Add another interface in between can solve the problem but we still unable to extend IMessage APIs in future.

ObsidianMinor: To explain, the generated messages are partial. This allows for implementing OnConstruction in the generated code without modifying it. This also allows for users to make the generated code to extend upon a base class if they wish. Since classes in .NET can't inherit from two base classes, if you add an abstract class into the generated code this breaks any user's code that makes the generated code use another base class in a separate partial declaration.

For example:

Generated code file
```cs
public partial class Message : pb::GeneratedMessage
{
    // … generated code
}
```
User's code
```cs
public partial class Message : BaseClass // compiler error: Partial declarations of 'Message' must not specify different base classes
{
    // … some user code
}
```

jtattermusch: one important point: adding methods to an interface breaks backward compatibility (e.g. after adding a method to IMessage, the implementations of IMessage outside of this repository wont' compile anymore -> backward compatibility is broken), so adding a method to IMessage is not an option (unless we want to break the backward compatibility and release a new  major version, and I don't think we want that).

ad naming: "IMessage2" doesn't seem like a good name (it looks a bit like a hacked together offshoot of IMessage and it's not descriptive). Sth like "IMessageProto2" is at least descriptive.

anandolee: We should NOT break backward compatibility, thus can't add method into IMessage. Add a new interface is not a good solution, still unable to extend base methods in future. Add a new interface each time scares.   

"if you add an abstract class into the generated code this breaks any user's code that makes the generated code use another base class in a separate partial declaration." Other protobuf languages do not support such cases, from protobuf team's review, we can break such cases and document it at the release notes. However, we still want to know from C# devolopers' veiw. @jtattermusch and @jskeet, do you agree to break such cases?

jskeet: I would personally strongly prefer not to break users like this. Breaking users breaks trust, which can be very hard to regain later. It's possible that C# users have a lower tolerance to breaking than other developers, or it may be a personal bias.

It's not only breaking, but breaking in a way which is very hard to recover from. It's not like "You need to change method call X by adding another argument" or similar.

Default methods in C# 8 might help in the future, but it'll be a while before we can assume those...

anandolee: Agree to use field accessors for now before we can add the default methods

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -140,6 +143,75 @@ public static ByteString ToByteString(this IMessage message)
...
+            if (message.Descriptor.File.Proto.Syntax != "proto2")
+            {
+                return true;
+            }
+
+            return message.Descriptor
+                .Fields
+                .InDeclarationOrder()
+                .Where(f => f.IsRequired || f.IsRepeated || f.IsMap || f.MessageType != null)
+                .All(f =>
```
anandolee: I am not an expert for C#, does this return true when all checks bellow return true, and it return false when one or more checks bellow return false?

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -140,6 +143,75 @@ public static ByteString ToByteString(this IMessage message)
...
+                        }
+                    }
+                    else if (f.IsRepeated)
+                    {
+                        IList list = (IList)f.Accessor.GetValue(message);
+                        foreach (IMessage item in list)
+                        {
+                            if (item is null)
+                            {
+                                return true;
```
anandolee: It will stop the foreach (IMessage item in list) loop when one item is true?

ObsidianMinor: Yes, if the item in the list is not an IMessage, the `item` will be null. Since you can't add null values to repeated fields, this doesn't risk returning early for null messages

csharp/src/Google.Protobuf/MessageExtensions.cs
ObsidianMinor: Yes

csharp/src/Google.Protobuf/MessageExtensions.cs
anandolee: Do you mean when it is a repeated scalar field? I'd prefer to check the field descriptor before the for loop?

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -140,6 +143,75 @@ public static ByteString ToByteString(this IMessage message)
...
+                    if (f.MessageType != null)
+                    {
+                        if (f.Accessor.HasValue(message))
+                        {
+                            var field = (IMessage)f.Accessor.GetValue(message);
+                            return field.IsInitialized();
+                        }
+                        else
+                        {
+                            return false;
```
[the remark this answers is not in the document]
anandolee: Please fix it, it looks like an error

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -140,6 +143,75 @@ public static ByteString ToByteString(this IMessage message)
...
+                        return true;
+                    }
+                    else if (f.IsMap)
+                    {
+                        IDictionary list = (IDictionary)f.Accessor.GetValue(message);
+                        foreach (IMessage item in list.Values)
+                        {
+                            if (item is null)
+                            {
+                                return true;
```
anandolee: map field item is always MapEntry message. Same for scalar map

csharp/src/Google.Protobuf/MessageExtensions.cs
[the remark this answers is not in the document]
ObsidianMinor: if sub messages aren't required, then should *any* sub messages like repeated message fields or map fields have their messages checked?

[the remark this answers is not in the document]
ObsidianMinor: @anandolee I was following the generated C++ for this

anandolee:
Friendly ping @ObsidianMinor , are you still working on this PR?

ObsidianMinor:
I am, @anandolee, I'm still wondering about [your comments earlier](https://github.com/protocolbuffers/protobuf/pull/4642#discussion_r210130189) about how

> sub message may not required

since if this is true, then there would be no need to check *any* sub messages, including repeated fields and map fields.

csharp/src/Google.Protobuf/MessageExtensions.cs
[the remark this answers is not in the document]
anandolee: Sub messages should be checked. I mean if the message field is optional, should not return false here when not set.

For example
message foo {
optional sub_message A;  //If set, should check A.IsInitialized() which is correct in your code. If not set, should not return false here.
}

src/google/protobuf/compiler/csharp/csharp_field_base.cc
```diff
@@ -357,9 +342,15 @@ std::string FieldGeneratorBase::default_value(const FieldDescriptor* descriptor)
         return "false";
       }
     case FieldDescriptor::TYPE_STRING:
-      return GetStringDefaultValueInternal();
+      if (descriptor->default_value_string().empty())
```
anandolee: Put this into GetStringDefaultValueInternal()

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -61,20 +61,27 @@ bool CompareFieldNumbers(const FieldDescriptor* d1, const FieldDescriptor* d2) {
 MessageGenerator::MessageGenerator(const Descriptor* descriptor,
                                    const Options* options)
     : SourceGeneratorBase(descriptor->file(), options),
-      descriptor_(descriptor) {
-
-  // sorted field names
-  for (int i = 0; i < descriptor_->field_count(); i++) {
```
anandolee: If this can be removed, the field_nams_ definition in charp_message.h and the "const std::vector<std::string>& MessageGenerator::field_names() " below should also be removed.

ObsidianMinor:
Doesn't look like I caused the failed Java builds.

anandolee:
Thanks for the change. 
Just note: as proto2 is not fully supported, we still disable the proto2 codegen thus field presence is unable to test in this CL. The API and implementation might be change in future RPs.

Will merge the PR

csharp/src/Google.Protobuf/Reflection/IFieldAccessor.cs
```diff
@@ -51,6 +51,11 @@ public interface IFieldAccessor
         /// </summary>
         void Clear(IMessage message);
 
+        /// <summary>
+        /// Indicates whether the field in the specified message is set. For proto3 fields, this throws an <see cref="InvalidOperationException"/>
+        /// </summary>
+        bool HasValue(IMessage message);
```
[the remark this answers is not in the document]
ObsidianMinor: Yes, I can introduce another interface (something like `IFieldPresenceAccessor`) and add the method in there.

csharp/src/Google.Protobuf/ProtoPreconditions.cs
```diff
@@ -1,79 +1,79 @@
...
+{
+    /// <summary>
+    /// Helper methods for throwing exceptions when preconditions are not met.
+    /// </summary>
+    /// <remarks>
+    /// This class is used internally and by generated code; it is not particularly
+    /// expected to be used from application code, although nothing prevents it
+    /// from being used that way.
+    /// </remarks>
+    public static class ProtoPreconditions
```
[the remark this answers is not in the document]
ObsidianMinor: I don't believe so, I think there were white space changes I committed accidentally.

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -140,6 +143,53 @@ public static ByteString ToByteString(this IMessage message)
...
+        {
+            if (message.Descriptor.File.Proto.Syntax != "proto2")
+            {
+                return true;
+            }
+
+            return message.Descriptor
+                .Fields
+                .InDeclarationOrder()
+                .All(f =>
```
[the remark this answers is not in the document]
ObsidianMinor: It's slower than a normal implementation however I couldn't think of a non-breaking way to introduce a code-generated version. I don't believe it will be used too often, since the consumer will have to invoke it manually after parsing (the library doesn't on its own).

Theoretically yes. As far as I know, protobuf reflection should work on Unity so this should too. But I don't know exactly how I can test it on Unity so I don't know for sure.

csharp/src/Google.Protobuf/Reflection/IFieldAccessor.cs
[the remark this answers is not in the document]
ObsidianMinor: Actually, I just checked the public package and the HasValue method is already in the public API. Can we still change it and write it off as "it wasn't meant to be released yet"?

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -1,771 +1,771 @@
...
-                }
-                foreach (var item in this)
-                {
-                    array.SetValue(item, index++);
-                }
-            }
-        }
-    }
-}
+﻿#region Copyright notice and license
```
jtattermusch: what exactly are the changes in this file? Looks like most of the file is unchanged, but due to whitespace changes (line endings?), it looks as if the file was completely rewritten.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -1,595 +1,595 @@
...
-            if (!(value is T))
-            {
-                return;
-            }
-            Remove((T)value);
-        }
-        #endregion        
-    }
-}
+﻿#region Copyright notice and license
```
jtattermusch: same as above: what exactly are the changes in this file? Looks like most of the file is unchanged, but due to whitespace changes (line endings?), it looks as if the file was completely rewritten.

ObsidianMinor: A year ago I made the mistake of committing line ending changes multiple times as I moved back and forth between wls and Windows where I had different git configurations on how to handle line endings. This file has no changes besides line endings (however at one point during review it did have changes which have since been removed).

csharp/src/Google.Protobuf/Collections/MapField.cs
ObsidianMinor: Same as below, mistakenly committed line endings.

csharp/src/Google.Protobuf/Reflection/IFieldAccessor.cs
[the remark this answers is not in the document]
jtattermusch: See my comment on other PR - this change seems to be there for about a year, so let's just leave as is.

PR #4659, opened 2018-05-21
Declare 'google' namespace when importing Python protobuf via Bazel

### LINKED ISSUE #4658: //:python_srcs doesn't play nice with other Google libraries
`//:python_srcs` doesn't include [`python/google/__init__.py`](https://github.com/google/protobuf/blob/master/python/google/__init__.py) in its `srcs` glob. Without it, protobuf exclusively defines a top-level `google` namespace which breaks other packages like [google-auth](https://google-auth.readthedocs.io/en/latest/) that expect to share the `google` namespace.

calder:
This fixes #4658.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -640,6 +640,7 @@ py_library(
     name = "python_srcs",
     srcs = glob(
         [
+            "python/google/__init__.py",
             "python/google/protobuf/*.py",
             "python/google/protobuf/**/*.py",
         ],


xfxyjwf:
I'm not sure if this is the right solution here. @nathanielmanistaatgoogle, any suggestion on how to solve the problem mentioned in https://github.com/google/protobuf/issues/4658 ?

calder:
@nathanielmanistaatgoogle @xfxyjwf: Expanded commit description. Does this make more sense?

> `python/google/__init__.py` declares the top-level 'google' namespace so that
`google.protobuf` can be imported alongside other Google Python modules like
`google.auth`.
>
> This works well when installing protobuf via Pip, but the Bazel `//:python_srcs`
rule doesn't include this file in its `srcs`. Bazel implicitly creates an empty
`google/__init__.py`, which does *not* set up a namespace. The result is that
consuming Python protobuf via Bazel breaks all other Google Python libraries.

nathanielmanistaatgoogle:
@calder: yeah, I think that's great.

@xfxyjwf: I'm satisfied with this; any remaining concerns of yours?

xfxyjwf:
@calder @nathanielmanistaatgoogle Thanks for helping fix it!

PR #4665, opened 2018-05-23
Fix syntax error in LINK_OPTS select

katre:
This has been silently failing for a while, it looks like.


xfxyjwf:
This has been fixed in another pull request.

PR #4667, opened 2018-05-24
Fix array constructor in c extension for compatibility

TeBoring:
In old generated code, the constructor of message doesn't provide a
default null to parent's constructor. Previously, in c extesnion, this
case was not handled.


bshaffer:
Good catch!

PR #4669, opened 2018-05-24
Fix syntax error in BUILD file

jtattermusch:
Fixes https://github.com/google/protobuf/issues/4668.


jtattermusch:
I'd suggest adding a continuous test for "bazel build".

PR #4672, opened 2018-05-24
Generated Code Changes for PHP

TeBoring:
1) Support array constructor
2) Use namespace instead of '_' for nested messages/enums.
3) Fixed the issue that reserved name in package or proto file path is not skipped.


TeBoring:
All changes had been reviewed in php-generated-dev branch.

acozzette:
Do we need these changes for the 3.6 release? If so then maybe we should merge them into the 3.6.x branch instead of master.

TeBoring:
SGTM

acozzette:
LGTM

PR #4674, opened 2018-05-24
Add kokoro configs for bazel build.

xfxyjwf:


kokoro/linux/bazel/presubmit.cfg
```diff
@@ -0,0 +1,5 @@
+# Config file for running tests in Kokoro
+
+# Location of the build script in repository
+build_file: "protobuf/kokoro/linux/bazel/build.sh"
+timeout_mins: 1440
```
BSBandme: Have this test run before? 1440 seems to be too long, I think if some PR does something wrong that makes some test deadlock forever then the presubmit will block following PR's presubmit.

xfxyjwf: Changed to 15 instead.

xfxyjwf:
@BSBandme It seems to me kokoro will run tests right away despite there are unfinished tests on old commits. When did you find an unfinished job blocks the tests of other pull requests?

BSBandme:
I'm not sure about presubmit kokoro test, but continuous will be blocked. The continuous benchmark kokoro build will run in 1.5h, and any PR submit during the period will be postponed. You may compare the log for 
[cpp_distcheck continous](https://fusion.corp.google.com/projectanalysis/current/KOKORO/prod:protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fcpp_distcheck%2Fcontinuous), which only take 20 minutes, and [benchmark continuous](https://fusion.corp.google.com/projectanalysis/current/KOKORO/prod:protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fbenchmark%2Fcontinuous) for the time period began from 12:27 May 22 to now.

Issue #4683, opened 2018-05-25
-std=c++11 flag added on wrong macOS versions


ryandesign:
python/setup.py checks the macOS version incorrectly, performing a *floating point* comparison instead of a *version number* comparison:

```
    v, _, _ = platform.mac_ver()
    if v:
      v = float('.'.join(v.split('.')[:2]))
      if v >= 10.12:
        extra_compile_args.append('-std=c++11')
```

So the `-std=c++11` flag will be added for macOS versions with a *floating point* value greater than or equal to 10.12—e.g. 10.12, 10.13, 10.2, 10.3, 10.4, 10.5, 10.6, 10.7, 10.8, 10.9—but not for 10.10 or 10.11. Surely this isn't what was intended. The clang++ compiler in Xcode on macOS only started using libc++ (which supports C++11) as of 10.9, and newer versions of clang++ in newer versions of Xcode on newer versions of macOS support more of the standard. In 10.8 and earlier, the C++ compilers used libstdc++ 4.2.1 or earlier (which doesn't support C++11 at all).

This problem was introduced in the 3.2.x branch in 117f7716414bb0826ad4b78d9b27f30e6f8e2e24 and merged to master as part of 7f3e23707122f31ebfa58b7280bd56cbe77cb44e.

What was the intention? On which macOS versions did you mean to add the flag?

I suspect it is incorrect to add the flag on the basis of a macOS version. You should instead add the flag on the basis of compiler capability. (It is possible to use newer clang++ compilers and libc++ on older versions of macOS and thus give them C++11 capability.)

TeBoring:
You are correct. Could you send us a PR for that?

ryandesign:
I am not a Python programmer and don't know how to do a version comparison there. I have also never looked at protobuf code before and don't know what the fix should be. Should the flag be added on 10.12 or later, as it was written? Or should it be 10.9 or later? What will be done for the 3.6.x branch, where [as I understand it](https://github.com/google/protobuf/issues/2780#issuecomment-352557942) C++11 will be unconditionally required?

TeBoring:
Yeah, I think we should unconditionally add the flag now.

ryandesign:
Please reopen. That commit only changed master. What about the 3.5.x branch which, if I understand correctly, is intended not to require C++11 yet?

TeBoring:
Although we didn't enforce c++11 in 3.5.x, it can still work with c++11.

TeBoring:
[...]

PR #4684, opened 2018-05-25
Always add -std=c++11 for mac

TeBoring:


TeBoring:
See #4683

python/setup.py
```diff
@@ -203,11 +203,9 @@ def get_option_from_sys_argv(option_str):
 
     v, _, _ = platform.mac_ver()
     if v:
-      v = float('.'.join(v.split('.')[:2]))
-      if v >= 10.12:
-        extra_compile_args.append('-std=c++11')
+      extra_compile_args.append('-std=c++11')
```
ryandesign: Why does this need to be conditional on `v`, or on macOS for that matter? If you now need C++11, don't you *always* (or at least on all non-Windows systems; Windows might be different) want to use `-std=c++11`?

PR #4686, opened 2018-05-25
Add back GeneratedClassName to public

TeBoring:
These methods are used by grpc


jtattermusch:
This should help.
Note: these were originally removed in https://github.com/google/protobuf/pull/4536/files#diff-1fa0334dd52f38affd6a06da49b29b79.

TeBoring:
We combined ```GeneratedClassName(const google::protobuf::EnumDescriptor* desc);``` and ```GeneratedClassName(const google::protobuf::Descriptor* desc);``` to a macro in implementation.
If you have link issue, we can submit another PR.

PR #4694, opened 2018-05-29
Don't generate well_known_types_embed in tree during out-of-tree CMake build

benesch:
The well_known_types_embed.cc source file is generated during the build;
as a result, it should be generated in the build directory, not the
source directory, during an out-of-tree build. The autotools build does
this right, but the CMake build was doing it wrong.

### DIFF
--- a/cmake/libprotoc.cmake
+++ b/cmake/libprotoc.cmake
@@ -1,3 +1,7 @@
+set(js_well_known_types_embed
+  ${CMAKE_CURRENT_BINARY_DIR}/src/google/protobuf/compiler/js/well_known_types_embed.cc
+)
+
 set(libprotoc_files
   ${protobuf_source_dir}/src/google/protobuf/compiler/code_generator.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/command_line_interface.cc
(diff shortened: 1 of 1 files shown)


cmake/libprotoc.cmake
```diff
@@ -82,6 +85,7 @@ set(libprotoc_files
   ${protobuf_source_dir}/src/google/protobuf/compiler/ruby/ruby_generator.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/subprocess.cc
   ${protobuf_source_dir}/src/google/protobuf/compiler/zip_writer.cc
+  ${js_well_known_types_embed}
```
xfxyjwf: This list here will be overwritten when we sync the file lists and this change here will not last.

xfxyjwf:
@acozzette @haberman I think the existence of this special js_embed binary is costing us much more maintenance overhead than what could possibly be saved by having js well-known-types support defined in .js files. I would very much like to just check-in the well_known_types_embed.cc file and get rid of the js_embed binary from our automake/cmake/bazel build configs. Any objections?

acozzette:
@xfxyjwf That sounds good to me. It has been a hassle trying to keep this embed binary working in every build system.

benesch:
Sounds great to me. Anything I can do to help?

xfxyjwf:
@benesch I'll make the change in our internal code base. Could you help make a pull request to do the same (check-in well_known_types_embed.cc and remove js_embed) in our github repo?

xfxyjwf:
I'm removing the js_embed binary and checking in the well_known_types_embed.cc file in https://github.com/google/protobuf/pull/4709

benesch:
Awesome, thanks @xfxyjwf!

PR #4697, opened 2018-05-29
Move methods out of class

TeBoring:


src/google/protobuf/compiler/php/php_generator.h
```diff
@@ -48,19 +48,18 @@ class LIBPROTOC_EXPORT Generator
...
-      const google::protobuf::EnumDescriptor* desc);
-  LIBPROTOC_EXPORT std::string GeneratedClassName(
-      const google::protobuf::ServiceDescriptor* desc);
-
 };
 
+// To skip reserved keywords in php, some generated classname are prefixed.
+// Other code generators may need following API to figure out the actual
+// classname.
+LIBPROTOC_EXPORT std::string GeneratedClassName(
```
acozzette: I think a change is necessary in php_generator.cc as well. The thing is that the Descriptor and EnumDescriptor cases are handled by a template function in the .cc file, so that doesn't exactly match the function signatures here in the .h file.

Issue #4705, opened 2018-05-31
Building using "./autogen.sh; ./configure; make" with older compiler fails because -std=c++11 is not set


jtattermusch:
E.g. failures here (clang3.5, clang3.6, clang3.7, gcc4.8):

https://source.cloud.google.com/results/invocations/889e43b5-a2f7-4bc9-b12b-2a56bd3df701/targets/github%2Fgrpc%2Faggregate_tests/tests

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.9/../../../../include/c++/4.9/bits/c++0x_warning.h:32:2: error: This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
#error This file requires compiler and library support for the \
 ^
In file included from google/protobuf/any.pb.cc:4:
```

```
./autogen
./configure
make
```

jtattermusch:
FYI @acozzette  @TeBoring

acozzette:
I just sent out #4706 which should hopefully fix this.

jtattermusch:
I created https://github.com/google/protobuf/pull/4710 to upport to master, but otherwise this can be considered fixed.

PR #4706, opened 2018-05-31
Require C++11 and pass -std=c++11

acozzette:
This commit updates the Autotools build to require a C++11 compiler and
explicitly pass -std=c++11. This seems to be necessary for enabling
C++11 support in older compilers.

This should fix issue #4705.


acozzette:
I'm not entirely sure if this is the right change to make but I figure we can give it a try. I'm surprised that the `AX_CXX_COMPILE_STDCXX` macro doesn't automatically add `-std=c++11` on its own.

acozzette:
That seems like a good idea because I don't know much about autotools either. @xfxyjwf Does this change look right to you?

PR #4707, opened 2018-05-31
proto_writer.h: RenderBytes is an override. Marking it.

medinandres:


src/google/protobuf/util/internal/proto_writer.h
```diff
@@ -110,7 +110,7 @@ class LIBPROTOBUF_EXPORT ProtoWriter : public StructuredObjectWriter {
     return RenderDataPiece(name,
                            DataPiece(value, use_strict_base64_decoding()));
   }
-  virtual ProtoWriter* RenderBytes(StringPiece name, StringPiece value) {
+  virtual ProtoWriter* RenderBytes(StringPiece name, StringPiece value) override {
```
xfxyjwf: Remove 'virtual'?

medinandres: I'm not involved in what was the decision to keep this method virtual. Our compiled flow just flagged the override was missing. I do see other classes inheriting ProtoWriter, but I don't see them overriding RenderBytes.

xfxyjwf: Removing "virtual" does not make this method non-virtual. The presence of "override" already tells the compiler that this is a virtual method. Having both "virtual" and "override" is redundant.

PR #4720, opened 2018-06-03
Align Bazel on Windows with CMake's MSVC flags

rongjiecomputer:
Make Bazel on Windows build log less noisy.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -22,7 +22,6 @@ config_setting(
 MSVC_COPTS = [
     "/DHAVE_PTHREAD",
     "/wd4018", # -Wno-sign-compare
-    "/wd4065", # switch statement contains
     "/wd4065", # switch statement contains 'default' but no 'case' labels
     "/wd4146", # unary minus operator applied to unsigned type, result still unsigned
     "/wd4244", # 'conversion' conversion from 'type1' to 'type2', possible loss of data
--- a/BUILD
+++ b/BUILD
@@ -69,17 +69,6 @@ config_setting(
 LINK_OPTS = select({
     ":android": [],
     ":msvc": [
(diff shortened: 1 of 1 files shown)


rongjiecomputer:
`-WHOLEARCHIVE:setargv.obj` is moved from `LINKOPTS` to `cc_library(name = "protoc_lib")` as it is the only place that needs this. Adding this flag in `LINKOPTS` will cause LNK4042 warning (object specified more than once; extras ignored).

`-WHOLEARCHIVE:` is changed to `-DEFAULTLIB:` as `-WHOLEARHIVE` can only be used for `.lib` library since VS2017 (LNK1356 error).

rongjiecomputer:
If possible, I want this PR be included in 3.6.0 so that I can convince Tensorflow and gRPC to update dependency that contains this PR.

BUILD
```diff
@@ -22,7 +22,21 @@ config_setting(
 MSVC_COPTS = [
     "/DHAVE_PTHREAD",
     "/wd4018", # -Wno-sign-compare
+    "/wd4065", # switch statement contains
```
xfxyjwf: This one is duplicated.

xfxyjwf:
I think this PR is too late to make into the 3.6.0 release.

BUILD
rongjiecomputer: Done.

rongjiecomputer:
@xfxyjwf In that case, I will try to see if Tensorflow and gRPC are willing to pin to a commit near 3.6.0 after 3.6.0 is released. I thought this might still make it since 3.6.0 is still in rc2 and the changelog isn't updated yet #4713.

rongjiecomputer:
@xfxyjwf Ping.

PR #4729, opened 2018-06-05
Update code to work for Xcode 10b1

sergiocampama:


objectivec/GPBArray.m
```diff
@@ -2537,12 +2537,12 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
   return [_array countByEnumeratingWithState:state objects:buffer count:len];
 }
 
-- (void)enumerateObjectsUsingBlock:(void (^)(id obj, NSUInteger idx, BOOL *stop))block {
+- (void)enumerateObjectsUsingBlock:(void (NS_NOESCAPE ^)(id obj, NSUInteger idx, BOOL *stop))block {
```
thomasvl: Please update objectivec/README.md, i believe it says we just require Xcode 7.0, so this likely needs to be raised.

thomasvl: Actually, don't we also want the NO_ESCAPE in the header also so code calling this is correct?

thomasvl: And no we don't since these are just the hooked cases.  We probably should add NOESCAPE to the GPB ones, but that can happen later.

PR #4754, opened 2018-06-12
print() is a function in Python 3

cclauss:
This PR should enable us to run the benchmarks under both Python 2 and Python 3.

### DIFF
--- a/benchmarks/util/run_and_upload.py
+++ b/benchmarks/util/run_and_upload.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+from __future__ import absolute_import
 import argparse
 import os
 import re
@@ -5,7 +7,7 @@ import copy
 import uuid
 import calendar
 import time
-import big_query_utils
+from . import big_query_utils
 import datetime
 import json
 # This import depends on the automake rule protoc_middleman, please make sure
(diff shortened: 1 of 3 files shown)


BSBandme:
Thanks for the PR to support python3 in benchmark1
But I think we need to do some more work. e.g. change 'open(...)' to 'open(..., "rb")', change the python/python_benchmark_messages.cc to work with python3-dev. 

Would you please make python3 work well for python benchmark?

cclauss:
@BSBandme Let’s make the changes that you suggest under other PRs.  The focus of this PR was to fix the print statements, not to solve all Python 3 issues.

BSBandme:
Asking Jie for python style reviewing @anandolee

xfxyjwf:
@anandolee Jie, please take a look at this python change.

benchmarks/util/run_and_upload.py
```diff
@@ -1,11 +1,13 @@
...
+from __future__ import absolute_import
 import argparse
 import os
 import re
 import copy
 import uuid
 import calendar
 import time
-import big_query_utils
+from . import big_query_utils
```
anandolee: Why this change is related?

cclauss: Python 2 tolerates relative imports but Python 3 does not without the leading dot.

PR #4756, opened 2018-06-12
Run flake8 to find Python syntax errors and undefined names

cclauss:
http://flake8.pycqa.org tests can find Python syntax errors and undefined names on each pull request before they are reviewed.

Results on Python 2.7:
* 3     F821 undefined name

Results on Python 3.6:
* 10    E999 SyntaxError: invalid syntax
* 5     F821 undefined name

### DIFF
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -3,6 +3,7 @@ setlocal
 IF %platform%==MinGW GOTO build_mingw
 IF %language%==cpp GOTO build_cpp
 IF %language%==csharp GOTO build_csharp
+IF %language%==python GOTO build_python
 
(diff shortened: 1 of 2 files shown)


xfxyjwf:
Thanks for the PR, but I don't think we want to include python tests in appveyor though. All python tests we have now works with kokoro. Can you instead change the kokoro scripts?

cclauss:
Is there public documentation for Kokoro?

xfxyjwf:
kokoro is use this test script:
https://github.com/google/protobuf/blob/master/tests.sh#L280

I think you just need to add the additional tests there. You can run the test script locally with:
```
./tests.sh python
```

PR #4766, opened 2018-06-14
Added Kokoro Windows release build config for protoc

acozzette:
This should make it easy to automatically build 32-bit and 64-bit
Windows protoc binaries.


kokoro/windows/protoc/release.cfg
```diff
@@ -0,0 +1,8 @@
+# Configuration for Windows protoc release builds
+build_file: "protobuf/kokoro/windows/protoc/build.bat"
```
xfxyjwf: For the file path, how about a dedicated release directory? For example:
protobuf/kokoro/release/protoc/windows/...
protobuf/kokoro/release/python/...
protobuf/kokoro/release/java/...

acozzette: That sounds good to me. I just now rearranged it to match that directory structure.

PR #4777, opened 2018-06-18
Work around MSVC issue with std::atomic initialization

acozzette:
MSVC seems to have a bug where it does not use constant initialization
for std::atomic, which ends up causing crashes during initialization.
This change introduces a workaround by putting the std::atomic inside a
union, which causes the compiler to use constant initialization for it.


xfxyjwf:
Can you include https://github.com/google/protobuf/pull/4772 in this PR so we have the test coverage for this?

acozzette:
@xfxyjwf Sure, I went ahead and included that change. `appveyor.yml` is somewhat different between the master and 3.6.x branches, so I couldn't directly cherry-pick your change, but I manually added the same kind of test.

PR #4779, opened 2018-06-18
check provided size to determine whether to export

mwei0210:
https://github.com/google/protobuf/issues/4747


mwei0210:
I signed it!

PR #4780, opened 2018-06-18
check provided size to check whether to export

mwei0210:
https://github.com/google/protobuf/issues/4747

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -3477,7 +3477,8 @@ void Generator::GenerateFile(const GeneratorOptions& options,
     GenerateExtension(options, printer, *it);
   }
 
-  if (options.import_style == GeneratorOptions::kImportCommonJs) {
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -3477,7 +3477,8 @@ void Generator::GenerateFile(const GeneratorOptions& options,
     GenerateExtension(options, printer, *it);
   }
 
-  if (options.import_style == GeneratorOptions::kImportCommonJs) {
+  // if provided is empty, do not export anything
+  if (options.import_style == GeneratorOptions::kImportCommonJs && provided.size()) {
```
acozzette: Just to make it a little bit clearer, could you do `!provided.empty()` instead of `provided.size()`?

mwei0210: changed :)

acozzette:
Thanks for the pull request!

Issue #4721, opened 2018-06-04
Python 3 - Unicode errors reading various METADATA.pb files with Python


bootrino:
I'm cross posting this because I don't know if it is a problem with protobuf or a problem with Google fonts.

If anyone can help it would be much appreciated.

Here is the issue:

https://github.com/google/fonts/issues/1591

acozzette:
@bootrino Could you paste an example of one of these `.pb` files that's causing a parse failure?

bootrino:
https://github.com/google/fonts/blob/master/ofl/imfellgreatprimer/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/ptsansnarrow/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/boogaloo/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/nanumbrushscript/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/imfellenglish/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/imfellfrenchcanon/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/playfairdisplaysc/METADATA.pb
https://github.com/google/fonts/blob/master/ofl/lato/METADATA.pb


Plus about 30 more that I can't be bothered getting the URL for.

bootrino:
```
(venv35) ubuntu@ip-10-0-0-196:~/code/$ file /usr/share/fonts/truetype/imfellgreatprimer/METADATA.pb
/usr/share/fonts/truetype/imfellgreatprimer/METADATA.pb: UTF-8 Unicode text
(venv35) ubuntu@ip-10-0-0-196:~/code/$
```

Have you tried executing the code fragment in the link above?  Here it is again:

```
import os
from fonts_public_pb2 import FontProto, FamilyProto
from google.protobuf import text_format


for dirpath, dirnames, filenames in os.walk('/usr/share/fonts/truetype'):
    metadataProtoFile = os.path.join(dirpath, 'METADATA.pb')
    if not os.path.exists(metadataProtoFile):
        continue

    metadata = FamilyProto()
    with open(metadataProtoFile) as f:
        text_data = f.read()
    try:
        text_format.Merge(text_data, metadata)
    except Exception as e:
        print(metadataProtoFile)
        print(repr(e))
    continue
```

acozzette:
[...]

PR #4804, opened 2018-06-20
Make text_encoding.CUnescape return a result in UTF-8 format

acozzette:
This function seemed to have a bug where it would encode its result in
the "raw_unicode_escape" format, which is an unusual Python-specific
format. The docstring indicates it should return UTF-8 bytes, so this
commit changes it to do that. This should fix a bug that was causing
text format protos with non-ASCII characters to be unparseable (#4721).


acozzette:
I realized that this fix doesn't work (it breaks other functionality), so I'm going to give up on this for now and close this pull request. I summarized the problem on #4721 in case someone else wants to give this a try.

PR #4808, opened 2018-06-21
Increase conversion support for JSON format

rannger:
Increase conversion support for JSON format

### DIFF
--- a/objectivec/GPBMessage.h
+++ b/objectivec/GPBMessage.h
@@ -242,6 +242,18 @@ CF_EXTERN_C_END
                     extensionRegistry:(nullable GPBExtensionRegistry *)extensionRegistry
                                 error:(NSError **)errorPtr;
 
+
+/**
+ * Initializes an instance by parsing the data. This method should be sent to
+ * the generated message class that the data should be interpreted as. If
+ * there is an error the method returns nil and the error is returned in
+ * errorPtr (when provided).
+
+ * @param data The JSON data to parse.
+ * @return An initialized instance of the generated class.
+ */
+- (nullable instancetype)initWithJSONData:(NSData*)data;
+
 /**
  * Initializes an instance by parsing the data from the given input stream. This
  * method should be sent to the generated message class that the data should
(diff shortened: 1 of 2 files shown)


thomasvl:
This also came up in #4112, `Descriptor` objects on the ObjC Runtime don't currently capture everything to fully support the JSON support in the protobuf spec, so while what you have here might work for some of your cases, it won't be correct for all cases to properly support it from a library pov.

Most servers should also accept the binary encoding which should also be smaller, for some ObjC where most of the "client" are likely mobile devices, the smaller payload of the binary encoding is usually a better choice.

rannger:
@thomasvl But Apple's swift version of protobuf also supports JSON conversion, why does Objective-C not consider support?

thomasvl:
> @thomasvl But Apple's swift version of protobuf also supports JSON conversion, why does Objective-C not consider support?

ObjC is completely dynamic, so nothing really ever dead strips, so everything the library could do, every client ships, so we've tried to be more minimal about what we includes because binary size matters a lot for ObjC since the products tends to be things like iOS Applications.

Swift does have the ability to do more dead stripping, so the hope is unused things can be left off, and as the tooling there matures, it will continue to get better, so apps might not have to pay the cost.

Swift also seems more likely to be used server side than ObjC, so that's why we've support JSON and TextFormat parsing on in that library.  We felt there was a higher likely hood for them to be useful/needed.

Also, since we figure ObjC would be more used for mobile setting, the data size being sent/received is likely to matter, the binary format is smaller, so it is a better choice.  (not to mention, the binary is the only one that allows the messages to evolve over times, if you read the JSON spec for Proto, you'll see it should error for unknown fields, where as the binary form can relay them even though it doesn't know what they are.)

dreampiggy:
Hi. Our company use Protobuf 3 in the Client & Server API request and response. It works for minimal bandwidth. 

But however, we face the issue that our QA Tester or RD need to map local the server response to trigger the specify case to test, ensure the robust and correct behavior for different response. When using JSON, we can use some network debugger tools like [https://www.charlesproxy.com/](Charles) to map to a local modified JSON file. However, when using Protobuf, we can not do this and have to use our company's server mock, which dramatically increase the complexibility.

So we found that #4112, fix it a little and introduce to our Client develop pipeline. It works for us in all aspects (even there will be rare case we don't meet). We can use Client debug options to enable or switch the JSON / Protobuf request.

As a user for your awesome framework, I hope this feature is official available. Even if there are some limit or rare case, maybe we can point it out in the documentations. And for user who don't need this, we can even create a CocoaPods's subspec or something to avoid effect their binary size. Hope for your response.

PR #4811, opened 2018-06-21
json: include field name in InvalidName status messages.

htuch:
This is invaluable for chasing down the culprit in failed JSON -> proto
conversions. We often see in Envoy that a failed JSON (or indirectly via
YAML) configuration won't load but it's unclear why.

### DIFF
--- a/src/google/protobuf/util/json_util.cc
+++ b/src/google/protobuf/util/json_util.cc
@@ -127,7 +127,8 @@ class StatusErrorListener : public converter::ErrorListener {
   virtual void InvalidName(const converter::LocationTrackerInterface& loc,
                            StringPiece unknown_name, StringPiece message) {
     status_ = util::Status(util::error::INVALID_ARGUMENT,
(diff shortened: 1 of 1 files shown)


htuch:
@xfxyjwf this was useful in debugging https://github.com/google/protobuf/issues/4771 (and related issues), would be great to merge with those in-flight fixes. Thanks.

src/google/protobuf/util/json_util.cc
```diff
@@ -127,7 +127,8 @@ class StatusErrorListener : public converter::ErrorListener {
   virtual void InvalidName(const converter::LocationTrackerInterface& loc,
                            StringPiece unknown_name, StringPiece message) {
     status_ = util::Status(util::error::INVALID_ARGUMENT,
-                             loc.ToString() + ": " + string(message));
+                           loc.ToString() + ":invalid name " +
```
xfxyjwf: Add a space after ":"?

PR #4815, opened 2018-06-22
Fix parameters declaration

vervaekejonathan:
Fix declaration of SerializeInternal and SerializeInternalToArray.
The linker of arm-none-eabi-gcc cannot found these functions without this modification.


vervaekejonathan:
I've signed it!

acozzette:
@vervaekejonathan Thanks for fixing this!

PR #4816, opened 2018-06-22
Basic Proto2 support for Ruby gem

hrsht:
This PR is basically the same as that created by @zanker in [#3606](https://github.com/google/protobuf/pull/3606), plus a few bug fixes. Opening a new PR as suggested by @xfxyjwf 

I have left the commits as it is for now (for ease of review), but will squash them before merging.

The next steps for a full proto2 support are to support extensions and then options. I am currently working on adding extensions support in the UPB library, so that the Ruby gem can support extensions.

cc: @yurokle


hrsht:
@xfxyjwf I am in the process of checking the CLA through my company's POC. Once that is done I will let you know.

The first 4 commits are what @zanker had in #3606, so those are under his name and email address.

Also, I realized that my github account was linked to my personal email. I have changed that now to use my work email, which will be in the commits going forward.

hrsht:
I signed it!

hrsht:
@xfxyjwf CLA should be good now.

Sorry @zanker I had to change the author to myself for your commits in order to get the CLA check to pass.

Yurokle:
@TeBoring could you please take another look at this PR? 
Is there anything else that's blocking from merging it?

hrsht:
Pinging again for a review!

TeBoring:
Sorry for late reply. I am taking a look.

TeBoring:
Do you have access to test results? If you do, could you fix the broken tests?

hrsht:
Yea, going to take a look at them. Looks like these kokoro builds were recently added (they weren't there before when I created the PR)

hrsht:
@TeBoring the tests pass now.

hrsht:
@TeBoring pinging again for a review!

ruby/.gitignore
```diff
@@ -1,8 +1,21 @@
 *.bundle
 tags
 .idea/
+lib/google/protobuf/any_pb.rb
```
TeBoring: Are these well known types added by intention?

ruby/tests/basic.rb
```diff
@@ -999,7 +999,6 @@ def test_to_h
         :optional_float=>0.0,
         :optional_int32=>0,
         :optional_int64=>0,
-        :optional_msg=>nil,
```
TeBoring: why remove this line?

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,1261 @@
...
+
+    add_message "Bar" do
+      optional :msg, :string, 1
+    end
+
+    add_message "Baz" do
+      optional :msg, :string, 1
+    end
+
+    add_message "TestMessage" do
```
TeBoring: Instead of creating the message manually, could you use protoc to generate code and import the generated code in test?

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,1261 @@
...
+  TestNonZeroEnum = pool.lookup("TestNonZeroEnum").enummodule
+  BadFieldNames = pool.lookup("BadFieldNames").msgclass
+  MapMessageWireEquiv = pool.lookup("MapMessageWireEquiv").msgclass
+  MapMessageWireEquiv_entry1 =
+    pool.lookup("MapMessageWireEquiv_entry1").msgclass
+  MapMessageWireEquiv_entry2 =
+    pool.lookup("MapMessageWireEquiv_entry2").msgclass
+  OneofMessage = pool.lookup("OneofMessage").msgclass
+
+# ------------ test cases ---------------
```
TeBoring: There are a lot of tests same for proto3. Could you help refactor the existing test framework to make it work with both proto2 and proto3?

ruby/.gitignore
zanker: These were from my original PR if I recall. Those were auto generated files when you ran some subset of the tests locally.

TeBoring: I think these should be added to .gitignore if you don't intentionally add them.

TeBoring:
Ruby tests seem failing.

ruby/tests/basic_proto2.rb
hrsht: These tests for proto2 are based on the existing tests for proto3 in `tests/basic.rb`. Are you suggesting to change both to use protoc generated protos (instead of the hardcoded ones)? If the latter, I will prefer to do it in a separate PR.

ruby/tests/basic.rb
hrsht: This is because of the introduction of hasbit functionality, which mirrors the upb's implementation. A submessage is only considered to be present if not nil and the `to_h` method excludes everything which is not present.

If you want me to preserve the behavior for proto3, I can.

ruby/tests/basic_proto2.rb
TeBoring: Yes, I mean use generated code (instead of hardcoded) in test. Besides, the test itself should be general to take both prot2 and proto3 message.
Do you plan to have this basic_proto2.rb first, and later refactor it to merge it with test for proto3? I prefer to do the refactoring in this PR.

ruby/tests/basic.rb
TeBoring: http://ruby-doc.org/core-2.2.0/Array.html#method-i-to_h, from the doc, it seems that we only need to make sure same messages have same to_h result and different messages have different to_h result. So it has no restriction on whether we include unsetted fields.

I don't think the present behavior for message field has any change between proto2 and proto3. Message field always has presence check. So the to_h bevavior for message field should be consistent. I think not include unset message field in ```to_h``` is the right option.

@haberman, any thought? Can we still change this?

haberman: I often look to Ruby's built-in `Struct` class to decide how things like this should behave. With Struct I notice:

```
> Struct.new("Foo", :a, :b).new.to_h
=> {:a=>nil, :b=>nil}
```

So I would be inclined to keep the nil values here. It matches the existing proto3 behavior which also seems right.

ruby/.gitignore
hrsht: Done.

ruby/tests/basic.rb
hrsht: I am fine preserving the proto3 behavior.

But for proto2, I will suggest to not include fields in `to_h` which aren't set. This is because proto2 uses `hasbit` for every field, including the basic types. So setting a value explicitly for a field in proto2, even for the default value, should set the `hasbit` for that field. Which means if `to_h` includes the field even when not set, it becomes impossible to know the difference between when the field was explicitly set and when it wasn't.

TeBoring:
@hrsht Hi, are you still working on refactoring the test?

hrsht:
@TeBoring Yes, plan to do it this week. Sorry, work was busy last week, so wasn't able to get to this PR. Will ping you when things are ready for re-review. Thanks!

TeBoring:
@hrsht No worry. Thanks for your help!

hrsht:
@TeBoring just pushed commits to refactor the common tests between proto2 and proto3. I defined a module `tests/common_tests.rb` which has all the common tests. Please let me know of what you think of this methodology.

ruby/tests/common_tests.rb
```diff
@@ -0,0 +1,1110 @@
...
+    assert m.optional_int64 == -0x1_0000_0000
+    m.optional_uint32 = 0x9000_0000
+    assert m.optional_uint32 == 0x9000_0000
+    m.optional_uint64 = 0x9000_0000_0000_0000
+    assert m.optional_uint64 == 0x9000_0000_0000_0000
+    m.optional_bool = true
+    assert m.optional_bool == true
+    m.optional_float = 0.5
+    assert m.optional_float == 0.5
+    m.optional_double = 0.5
```
TeBoring: The original test missed assert. Could you help add it?

ruby/tests/common_tests.rb
```diff
@@ -0,0 +1,1110 @@
+# Defines tests which are common between proto2 and proto3 syntax.
+#
+# Requires that the proto messages are exactly the same in proto2 and proto3 syntax
+# and that the including class should define a 'proto_module' method which returns
+# the enclosing module of the proto message classes.
+module CommonTests
+  def test_defaults
+    m = proto_module::TestMessage.new
```
TeBoring: Is it possible to define proto_module::TestMessage in the begining so that you don't need to write so many proto_module::TestMessage?

ruby/tests/basic.rb
```diff
@@ -9,270 +14,90 @@
...
-      assert_match(/hello/, e.message)
+      m.a = "foo"
+      assert_equal "foo", m.a
+      assert_true m.has_my_oneof?
+      m.clear_a
+      assert_false m.has_my_oneof?
     end
 
+
     def test_initialization_map_errors
```
TeBoring: Why this hasn't been moved to common test?

ruby/tests/basic.rb
```diff
@@ -290,417 +115,6 @@ def test_initialization_map_errors
...
-      assert m["b"].object_id == m2["b"].object_id
-
-      m2 = Google::Protobuf.deep_copy(m)
-      assert m == m2
-      assert m.object_id != m2.object_id
-      assert m["a"].object_id != m2["a"].object_id
-      assert m["b"].object_id != m2["b"].object_id
-    end
-
     def test_map_field
```
TeBoring: ditto

ruby/tests/basic.rb
```diff
@@ -784,211 +198,6 @@ def test_map_encode_decode
...
-    def test_protobuf_encode_decode_json_helpers
-      m = TestMessage.new(:optional_string => 'foo', :repeated_string => ['bar1', 'bar2'])
-      encoded_msg = Google::Protobuf.encode_json(m)
-      assert_equal m.to_json, encoded_msg
-
-      decoded_msg = Google::Protobuf.decode_json(TestMessage, encoded_msg)
-      assert_equal TestMessage.decode_json(m.to_json), decoded_msg
-    end
-
     def test_to_h
```
TeBoring: ditto?

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,196 @@
...
+
+      m.optional_msg = TestMessage2.new
+      assert_true m.has_optional_msg?
+
+      m = OneofMessage.new
+      assert_false m.has_my_oneof?
+      m.a = "foo"
+      assert_true m.has_a?
+      assert_equal "foo", m.a
+      assert_true m.has_my_oneof?
```
TeBoring: Could you also check other oneof fields' has and getter method?

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,196 @@
...
+      assert_false m.has_optional_enum?
+    end
+
+    def test_set_clear_defaults
+      m = TestMessageDefaults.new
+
+      m.optional_int32 = -42
+      assert_equal -42, m.optional_int32
+      assert_true m.has_optional_int32?
+      m.clear_optional_int32
```
TeBoring: Is there a clear method for proto3?

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,196 @@
...
+      assert_equal TestMessage2.new(:foo => 42), m.optional_msg
+      assert_true m.has_optional_msg?
+
+      m.clear_optional_msg
+      assert_equal nil, m.optional_msg
+      assert_false m.has_optional_msg?
+
+      m.optional_msg = TestMessage2.new(:foo => 42)
+      assert_equal TestMessage2.new(:foo => 42), m.optional_msg
+      assert_true TestMessageDefaults.descriptor.lookup('optional_msg').has?(m)
```
TeBoring: This is also a new API?

ruby/tests/common_tests.rb
hrsht: Do you mean as a constant? The problem with that is that `proto_module` is a method and for different classes where `CommonTests` gets included, it can have a different value. And hence, a constant won't work.

The alternative is to define a method like `test_message_class` or so and use that, but I like having `proto_module::TestMessage.new` over `test_message_class.new`.

ruby/tests/basic.rb
hrsht: This PR doesn't add support for maps for proto2 (plan to do it in a followup). And since this test uses MapMessage, I have kept it separate for now. Will put it in `CommonTests` module once I add map support in proto2.

ruby/tests/basic.rb
hrsht: Same reason as above

ruby/tests/basic.rb
hrsht: `to_h` method for proto2 does not return fields which aren't set. And since the expectations are different between proto2 and proto3, this test remains separate.

ruby/tests/basic_proto2.rb
hrsht: Yes. See `test_set_clear_defaults` in `basic.rb`.

ruby/tests/basic_proto2.rb
hrsht: Yes.

hrsht:
@TeBoring ping on the review!

TeBoring:
I am waiting for a review for the API.
I think @haberman has contacted Yuriy Naidyon from square for that.
@hrsht Could you check with Yuriy Naidyon about the API review?

haberman:
This looks great overall.

Our main priority will be reviewing the new APIs. Implementation issues can be refined over time, but APIs are somewhat set in stone once introduced. Could you possibly write a brief summary of the new APIs in a Google Doc?

At first glance, the only suggestion I have on the new APIs is to make the FileDescriptor more consistent with the style of the builder. Something more like:

```ruby
  add_file("foo/bar.proto", syntax = :proto2) do |f|
    add_message "A.B.C.TestMessage" do
          optional :optional_int32, :int32, 1, default: 1
          # ...
    end
  end
end
```

haberman:
Thanks for the doc. Hope the work on the API is going well.

Just wanted to throw something out there: a much simpler approach would be to ditch the DSL and just put a serialized descriptor into the generated code. I know this isn't always a popular idea since the DSL is nice and readable. But tunneling every possible protobuf construct through a DSL is going to get more and more difficult, particularly when it comes to custom options. This gets especially hairy since custom options can describe themselves!

One reasonable middle ground is to put everything in the DSL except custom options themselves, which can be protobuf serialized strings.

The simple approach would be trivial to implement though: we could forget all the builders and just have a single method that passes a serialized file descriptor to the pool and builds it.

TeBoring:
@hrsht Could you set the default access to allow everyone to comment for [the doc](https://docs.google.com/document/d/1Gh_X3sxTjc9sFbgGSpIaj6T1oBtHDUOZFKPBFRVr4xk/edit#)?

hrsht:
@haberman are you suggesting to also get rid of the current `add_message` and `add_enum` builders and build everything from the serialized descriptors? Or just build the FileDescriptor from the serialized file descriptor protobuf?

TeBoring:
It seems I don't have any access (even viewing ...). Could you double check? @hrsht

hrsht:
Sorry, it should be fixed now. Also, gave you direct permissions.

hrsht:
@haberman changed the Builder DSL to add the `add_file` method for FileDescriptor and move `add_message` and `add_enum` builders inside the block passed to `add_file` (as you had suggested before).

For backwards compatibility, `add_message` and `add_enum` methods still exist for Builder, but with this change, any new file generated by protoc should have the new DSL.

hrsht:
@haberman Ping!

Do you think if the PR is at the point where it can be merged before end of the month? I am going on paternity leave for 3 months starting 1 Oct and won't be able to work on it during this time if it is still open.

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,253 @@
...
+        :optional_bool=>true,
+        :optional_double=>-10.100001,
+        :optional_string=>"foo",
+        :repeated_string=>["bar1", "bar2"],
+      }
+      assert_equal expected_result, m.to_h
+
+      m = OneofMessage.new(:a => "foo")
+      expected_result = {:a => "foo"}
+      assert_equal expected_result, m.to_h
```
TeBoring: Maybe a wrong question: do we only need to guarantee equal messages have same to_h result and non-equal messages have different to_h result? Do we need to actually make the content in to_h part of the API? @haberman

ruby/tests/common_tests.rb
```diff
@@ -0,0 +1,1111 @@
...
+    assert_equal 3, m.repeated_msg.first.sub_child.optional_int32
+  end
+
+  def test_inspect
+    m = proto_module::TestMessage.new(:optional_int32 => -42,
+                                      :optional_enum => :A,
+                                      :optional_msg => proto_module::TestMessage2.new,
+                                      :repeated_string => ["hello", "there", "world"])
+    expected = "<#{proto_module}::TestMessage: optional_int32: -42, optional_int64: 0, optional_uint32: 0, optional_uint64: 0, optional_bool: false, optional_float: 0.0, optional_double: 0.0, optional_string: \"\", optional_bytes: \"\", optional_msg: <#{proto_module}::TestMessage2: foo: 0>, optional_enum: :A, repeated_int32: [], repeated_int64: [], repeated_uint32: [], repeated_uint64: [], repeated_bool: [], repeated_float: [], repeated_double: [], repeated_string: [\"hello\", \"there\", \"world\"], repeated_bytes: [], repeated_msg: [], repeated_enum: []>"
+    assert_equal expected, m.inspect
```
TeBoring: map? oneof?

ruby/tests/common_tests.rb
```diff
@@ -0,0 +1,1111 @@
...
+  end
+
+  def test_hash
+    m1 = proto_module::TestMessage.new(:optional_int32 => 42)
```
TeBoring: map? oneof?

ruby/tests/common_tests.rb
TeBoring: Ignore me if this PR doesn't deal with inspect.

ruby/tests/common_tests.rb
TeBoring: Ignore me if this PR doesn't deal with to_h.

ruby/ext/google/protobuf_c/message.c
```diff
@@ -129,99 +183,81 @@ static VALUE which_oneof_field(MessageHeader* self, const upb_oneofdef* o) {
...
+    rb_raise(rb_eArgError, "Expected 1 argument, received %d", argc);
   }
 
+  // Return which of the oneof fields are set
   if (o != NULL) {
-    // This is a oneof -- return which field inside the oneof is set.
-    if (setter) {
-      rb_raise(rb_eRuntimeError, "Oneof accessors are read-only.");
+    if (accessor_type == METHOD_SETTER || accessor_type == METHOD_CLEAR) {
+      rb_raise(rb_eRuntimeError, "Oneoff accessors are read-only.");
```
haberman: Oneof fields support clearing in C++, I think this should be supported.

ruby/tests/common_tests.rb
```diff
@@ -0,0 +1,1111 @@
...
+    assert_equal :C, m.optional_enum
+    assert_equal [:A, :B], m.repeated_enum
+
+    m = proto_module::TestMessage.new(:optional_string => :foo, :repeated_string => [:foo, :bar])
+    assert_equal 'foo', m.optional_string
+    assert_equal ['foo', 'bar'], m.repeated_string
+  end
+
+  def test_ctor_nil_args
+    m = proto_module::TestMessage.new(:optional_enum => nil, :optional_int32 => nil, :optional_string => nil, :optional_msg => nil)
```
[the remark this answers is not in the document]
hrsht: Passing `nil` does not set the field, so has method will return false.

hrsht:
@haberman Apparently I can't still turn on conformance tests for proto2 until support for extensions is added :(

Will fix the >80 char lines

ruby/tests/common_tests.rb
hrsht: Added for oneof and map in proto3. Map isn't supported yet in proto2, so can't have the test in common_tests.rb file yet.

ruby/tests/common_tests.rb
hrsht: Sorry, this test is for hash values. The test for `to_h` exist in the respective test files for proto2 and proto3 (`basic_proto2.rb` and `basic.rb`)

hrsht:
@haberman @TeBoring Addressed your latest comments. PTAL?

ruby/tests/common_tests.rb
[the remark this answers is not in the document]
TeBoring: Can we test the has behavior in proto2 specific test?

ruby/tests/basic_proto2.rb
TeBoring: Talked offline that the content of to_h is also public.

hrsht:
@haberman ping!

ruby/tests/basic_proto2.rb
```diff
@@ -0,0 +1,265 @@
...
+    add_file "test_proto2.proto", syntax: :proto2 do
+      add_message "BadFieldNames" do
+        optional :dup, :int32, 1
+        optional :class, :int32, 2
+        optional :"a.b", :int32, 3
+      end
+    end
+  end
+
+  BadFieldNames = pool.lookup("BadFieldNames").msgclass
```
haberman: I looked over everything once more, and it all looks good. I just had one question about this test: what is it intending to test? Some of those field names look problematic for Ruby but are valid field names for protobuf ("dup") while some of them are just invalid names period ("a.b"). Shouldn't this be throwing an error? What is this test trying to test.

hrsht: This proto setup is actually copied over from the proto3 test in `basic.rb` https://github.com/protocolbuffers/protobuf/blob/master/ruby/tests/basic.rb#L79. We had just copied over the test setup for proto3 from `basic.rb` and repurposed it for proto2.

elliot42:
CONGRATS!!  🎉 🎉

Yurokle:
Huge, huge thank you @hrsht, @TeBoring and @haberman!!! 🎉 🦄 🎆

hrsht:
@haberman @TeBoring thanks a lot for your review on this PR which turned out to be pretty big, non-trivial and a tedious one.

TeBoring:
Thanks you for the work. Really appreciate that!

TeBoring:
I am afraid this change has broken our continuous test: https://source.cloud.google.com/results/invocations/0cf66438-ad5f-4fe1-b576-d371dfe5f33d/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fruby_all%2Fpresubmit/log

PR #4819, opened 2018-06-23
Build ruby gem on kokoro

TeBoring:
For mac, linux, windows


xfxyjwf:
@TeBoring Bo, can you change the layout of ruby release scripts to be consistent with the protoc directory? See: https://github.com/google/protobuf/issues/4884

PR #4825, opened 2018-06-25
Ignore unknown enum received in json when ignoreUnknownFields flag is set

vijay-bhatt:
Ignore unknown enum received in json when ignoreUnknownFields flag is set.

---
This change is [Reviewable](https://reviewable.io/reviews/google/protobuf/4825)

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -50,6 +50,7 @@ import com.google.protobuf.Descriptors.Descriptor;
 import com.google.protobuf.Descriptors.EnumDescriptor;
 import com.google.protobuf.Descriptors.EnumValueDescriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
+import com.google.protobuf.Descriptors.FieldDescriptor.Type;
 import com.google.protobuf.Descriptors.FileDescriptor;
 import com.google.protobuf.Descriptors.OneofDescriptor;
 import com.google.protobuf.DoubleValue;
@@ -1539,7 +1540,7 @@ public class JsonFormat {
         Object key = parseFieldValue(keyField, new JsonPrimitive(entry.getKey()), entryBuilder);
         Object value = parseFieldValue(valueField, entry.getValue(), entryBuilder);
         if (value == null) {
-          if(field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
+          if(valueField.getType() == Type.ENUM && field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
             continue;
           } else {
             throw new InvalidProtocolBufferException("Map value cannot be null.");
@@ -1561,7 +1562,7 @@ public class JsonFormat {
       for (int i = 0; i < array.size(); ++i) {
         Object value = parseFieldValue(field, array.get(i), builder);
         if (value == null) {
-          if(field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
+          if(field.getType() == Type.ENUM && field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
             continue;
           } else {
             throw new InvalidProtocolBufferException(
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -1746,6 +1746,10 @@ public class JsonFormat {
           // Fall through. This exception is about invalid int32 value we get from parseInt32() but
           // that's not the exception we want the user to see. Since result == null, we will throw
           // an exception later.
+            
+          if (ignoringUnknownFields) {
+            result = enumDescriptor.getValues().get(0);
+          }            
         }
 
         if (result == null) {
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -1747,12 +1747,12 @@ public class JsonFormat {
           // that's not the exception we want the user to see. Since result == null, we will throw
           // an exception later.
             
-          if (ignoringUnknownFields && !ignoringUnknownFields) {
+          if (ignoringUnknownFields) {
             result = enumDescriptor.getValues().get(0);
           }            
         }
 
-        if (result == null) {
+        if (result == null && !ignoringUnknownFields) {
           throw new InvalidProtocolBufferException(
               "Invalid enum value: " + value + " for enum type: " + enumDescriptor.getFullName());
         }
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -1747,7 +1747,7 @@ public class JsonFormat {
           // that's not the exception we want the user to see. Since result == null, we will throw
           // an exception later.
             
-          if (ignoringUnknownFields) {
+          if (ignoringUnknownFields && !ignoringUnknownFields) {
             result = enumDescriptor.getValues().get(0);
           }            
         }
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -1539,7 +1539,11 @@ public class JsonFormat {
         Object key = parseFieldValue(keyField, new JsonPrimitive(entry.getKey()), entryBuilder);
         Object value = parseFieldValue(valueField, entry.getValue(), entryBuilder);
         if (value == null) {
-          throw new InvalidProtocolBufferException("Map value cannot be null.");
+          if(field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
+            continue;
+          } else {
+            throw new InvalidProtocolBufferException("Map value cannot be null.");
+          }
         }
         entryBuilder.setField(keyField, key);
         entryBuilder.setField(valueField, value);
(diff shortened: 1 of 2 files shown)


java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -1746,6 +1746,10 @@ private EnumValueDescriptor parseEnum(EnumDescriptor enumDescriptor, JsonElement
           // Fall through. This exception is about invalid int32 value we get from parseInt32() but
           // that's not the exception we want the user to see. Since result == null, we will throw
           // an exception later.
+            
+          if (ignoringUnknownFields) {
+            result = enumDescriptor.getValues().get(0);
```
xfxyjwf: This will cause the field to have the default value, i.e., instead of ignoring unknown enum values, they are converted to the default value. It doesn't matter to proto3 because having the default value == not set, but for proto2 it means data corruption (users will get an incorrect enum value).

Could you find a way to ignore the field entirely instead?

vijay-bhatt: I actually based this on your comment dated Jun 30 2017 for the discussion on https://github.com/google/protobuf/pull/3289/commits/70049452c2ece7918b3bf8cb992b8178e6e5e1cf
for veerenmandalia's implementation

Having to "treat it as if the field is not present" would mean assigning it a default value.

xfxyjwf: "treat it as if the field is not present" does not mean assigning it a default value. In proto3, assigning default value == not present, but that's not the case in proto2. Here we need to find a way to skip the field entirely and not set any value into the field.

vijay-bhatt: Ok. So instead, if i return null from this method, the protobuf field will be not be set. In my test, that is also returning the default value as the builder enum int is preset to 0 and the proto generated class returns default if the enum int value is 0. Should returning null be the fix here? How will it behave with proto2 compiler?

xfxyjwf: I think the code needs some refactoring to be able to ignore unknown enum values. Returning null is one option and we need to update all callsites of the method to make sure they are handling the null return value properly.

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -1179,8 +1179,8 @@ public void testParserIgnoringUnknownEnums() throws Exception {
     TestAllTypes.Builder builder = TestAllTypes.newBuilder();
     String json = "{\n" + "  \"optionalNestedEnum\": \"XXX\"\n" + "}";
     JsonFormat.parser().ignoringUnknownFields().merge(json, builder);
-    assertEquals(NestedEnum.FOO, builder.getOptionalNestedEnum());
-  }  
+    assertEquals(0, builder.getOptionalNestedEnumValue());
```
xfxyjwf: Add a test for unknown enum value in a repeated enum field?

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
vijay-bhatt: I have made some refactoring change. I also saw the call sites where the changed method is called. In map, it is adding the null as value. As a result, it should return a default 0th enum. Accordingly I have added 2 more tests, one for repeated enum and other for enum in map value.

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -1173,7 +1173,41 @@ public void testParserIgnoringUnknownFields() throws Exception {
...
+
+  public void testUnknownEnumMap() throws Exception {
+    TestMap.Builder builder = TestMap.newBuilder();
+    JsonFormat.parser().ignoringUnknownFields().merge(
+        "{\n"
+          + "  \"int32ToEnumMap\": {1: XXX, 2: FOO}"
+          + "}",
+        builder);
 
+    assertEquals(Integer.valueOf(0), builder.getInt32ToEnumMapValueMap().get(1));
```
xfxyjwf: The entry should be ignored, i.e., no entry for 1 should exist in the map.

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -1173,7 +1173,41 @@ public void testParserIgnoringUnknownFields() throws Exception {
...
+  
+  public void testRepeatedUnknownEnum() throws Exception {
+    TestAllTypes.Builder builder = TestAllTypes.newBuilder();
+    JsonFormat.parser().ignoringUnknownFields().merge(
+        "{\n"
+          + "  \"repeatedNestedEnum\": [XXX, FOO, BAR, BAZ]"
+          + "}",
+        builder);
+
+        assertEquals(0, builder.getRepeatedNestedEnumValue(0));
```
xfxyjwf: Same here. I think the entry should be discarded instead of converted to a different value.

vijay-bhatt:
Please review. I have refactored the code so that unknown enum in a Map of enum value is ignored. Similarly for repeated unknown enum, it also is ignored. Changed the test cases accordingly.

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -1557,8 +1561,12 @@ private void mergeRepeatedField(
       for (int i = 0; i < array.size(); ++i) {
         Object value = parseFieldValue(field, array.get(i), builder);
         if (value == null) {
-          throw new InvalidProtocolBufferException(
-              "Repeated field elements cannot be null in field: " + field.getFullName());
+          if(field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
```
xfxyjwf: Same here.

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -659,7 +659,9 @@ public void testMapNullValueIsRejected() throws Exception {
     } catch (InvalidProtocolBufferException e) {
       // Exception expected.
     }
+  }
 
+  public void testMapNullValueIsIgnored() throws Exception {
```
xfxyjwf: We shouldn't introduce this behavior change for non-enum fields.

vijay-bhatt:
Please review

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -1539,7 +1540,11 @@ private void mergeMapField(FieldDescriptor field, JsonElement json, Message.Buil
         Object key = parseFieldValue(keyField, new JsonPrimitive(entry.getKey()), entryBuilder);
         Object value = parseFieldValue(valueField, entry.getValue(), entryBuilder);
         if (value == null) {
-          throw new InvalidProtocolBufferException("Map value cannot be null.");
+          if(valueField.getType() == Type.ENUM && field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
```
xfxyjwf: Can you change this to:
```
if (ignoringUnknownFields && valueField.getType() == Type.ENUM) {
  continue;
} else {
  throw ...
}
```

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -1557,8 +1562,12 @@ private void mergeRepeatedField(
       for (int i = 0; i < array.size(); ++i) {
         Object value = parseFieldValue(field, array.get(i), builder);
         if (value == null) {
-          throw new InvalidProtocolBufferException(
-              "Repeated field elements cannot be null in field: " + field.getFullName());
+          if(field.getType() == Type.ENUM && field.getFile().getSyntax() == FileDescriptor.Syntax.PROTO3) {
```
xfxyjwf: Change this to:
```
if (ignoringUnknownFields && field.getType() == Type.ENUM) {
} else {
}
```

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -659,7 +659,9 @@ public void testMapNullValueIsRejected() throws Exception {
     } catch (InvalidProtocolBufferException e) {
       // Exception expected.
     }
+  }
 
+  public void testMapEnumNullValueIsIgnored() throws Exception {
```
xfxyjwf: Please keep the old "testMapNullValueIsRejected()" method as-is, and add a new "testMapEnumNullValueIsIgnored()" method.

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -668,7 +670,9 @@ public void testMapNullValueIsRejected() throws Exception {
               + "  \"int32ToMessageMap\": {\"2\": null}\n"
               + "}",
           builder);
-      fail();
+      assertEquals(0, builder.getInt32ToEnumMapMap().size());
```
xfxyjwf: Please don't change this test for non-enum map values. As I suggested above, add a new method instead (do not split the code from the existing "testMapNullValueIsRejected()" method.

vijay-bhatt:
I have incorporated all your code review comments. Please review.

java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java
```diff
@@ -30,6 +30,18 @@
 
 package com.google.protobuf.util;
 
+import java.io.IOException;
```
xfxyjwf: Please restore the original imports order...

vijay-bhatt:
Done

vijay-bhatt:
Will these changes be pulled or do I need to merge / push?

xfxyjwf:
I just started the tests and will merge after the tests finish.

vijay-bhatt:
Thanks

PR #4829, opened 2018-06-26
Add continuous test for ruby 2.3, 2.4 and 2.5

TeBoring:


ebenoist:
@TeBoring RE: #4255

Is there some build output that I can look at? I can't fathom why this would be failing, but I'm happy to take a look.

Looks like only mac os failures?

TeBoring:
```
Failure: test_type_errors(BasicTest::MessageContainerTest)
/Volumes/BuildData/tmpfs/src/github/protobuf/ruby/compatibility_tests/v3.0.0/tests/basic.rb:225:in `test_type_errors'
     222: 
     223:     def test_type_errors
     224:       m = TestMessage.new
  => 225:       assert_raise TypeError do
     226:         m.optional_int32 = "hello"
     227:       end
     228:       assert_raise TypeError do

<TypeError> expected but was
<Google::Protobuf::TypeError(<Expected number type for integral field.>)>

diff:
? Google::Protobuf::TypeError(<Expected number type for integral field.>)
```

TeBoring:
Actually, same problem on linux

ebenoist:
@TeBoring I can take a closer look next week. I'm going to be out this one, but it does look like those tests may need to be updated if `assert_raise` doesn't accept a parent class. Note that `rescue` of course will catch parent class. Happy to pick this up next week when I'm back at work. This is why I updated the tests in the other file, and added a test to ensure that the error raised `is_a?(::TypeError)`

TeBoring:
Broken MacOS C++ and MacOS C++ Distcheck is not related to this change.

PR #4830, opened 2018-06-26
Add -Og to no-warning-test

### LINKED ISSUE #3711: Building no-warning-test fails with -Og in CXXFLAGS
When building with GCC, inclusion of `-Og` in `CXXFLAGS` causes `make check` to fail with errors such as:

```
In file included from google/protobuf/compiler/cpp/cpp_test_bad_identifiers.pb.cc:5:0:
./google/protobuf/compiler/cpp/cpp_test_bad_identifiers.pb.h: In member function ‘virtual bool protobuf_unittest::TestConflictingEnumNames::MergePartialFromCodedStream(google::protobuf::io::CodedInputStream*)’:
./google/protobuf/compiler/cpp/cpp_test_bad_identifiers.pb.h:3798:21: error: ‘value’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   conflicting_enum_ = value;
   ~~~~~~~~~~~~~~~~~~^~~~~~~
google/protobuf/compiler/cpp/cpp_test_bad_identifiers.pb.cc:6765:15: note: ‘value’ was declared here
           int value;
               ^~~~~
```

This appears to be related to [Bug 58455 - spurious "may be used uninitialized" warning with -Og](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58455) and [Bug 42145 - bogus "may be used uninitialized" (a || b converted to a|b)](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=42145).  As part of protobuf gets built with `-Werror`, such spurious warnings become fatal.

Tested with protobuf-3.3.0 using GCC-5.4.0 and GCC-6.4.0.

See also [Bug 618150 - dev-libs/protobuf-3.3.0 - error: 'value' may be used uninitialized in this function [-Werror=maybe-uninitialized]](https://bugs.gentoo.org/618150)

TeBoring:
Fixes #3711

### DIFF
--- a/kokoro/linux/bazel/build.sh
+++ b/kokoro/linux/bazel/build.sh
@@ -7,4 +7,3 @@ cd $(dirname $0)/../../..
 
 git submodule update --init --recursive
 bazel test :protobuf_test
-cat /tmpfs/tmp/bazel/execroot/com_google_protobuf/bazel-out/k8-fastbuild/testlogs/protobuf_test/test.log
--- a/kokoro/linux/bazel/build.sh
+++ b/kokoro/linux/bazel/build.sh
@@ -7,3 +7,4 @@ cd $(dirname $0)/../../..
 
 git submodule update --init --recursive
 bazel test :protobuf_test
+cat /tmpfs/tmp/bazel/execroot/com_google_protobuf/bazel-out/k8-fastbuild/testlogs/protobuf_test/test.log
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -876,7 +876,7 @@ no_warning_test.cc:
 
 no_warning_test_LDADD = $(PTHREAD_LIBS) libprotobuf.la libprotoc.la
 no_warning_test_CXXFLAGS = $(PTHREAD_CFLAGS) $(PTHREAD_DEF) $(ZLIB_DEF) \
-                           -Wall -Wextra -Werror -Wno-unused-parameter
+                           -Wall -Wextra -Werror -Wno-unused-parameter -Og
 nodist_no_warning_test_SOURCES = no_warning_test.cc $(protoc_outputs)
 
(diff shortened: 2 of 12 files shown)


src/Makefile.am
```diff
@@ -876,7 +876,7 @@ no_warning_test.cc:
 
 no_warning_test_LDADD = $(PTHREAD_LIBS) libprotobuf.la libprotoc.la
 no_warning_test_CXXFLAGS = $(PTHREAD_CFLAGS) $(PTHREAD_DEF) $(ZLIB_DEF) \
-                           -Wall -Wextra -Werror -Wno-unused-parameter
+                           -Wall -Wextra -Werror -Wno-unused-parameter -Og
```
xfxyjwf: What does -Og do?

TeBoring:
```
In GCC 4.8, a new general optimization level, -Og, has been introduced. It addresses the need for fast compilation and a superior debugging experience while providing a reasonable level of runtime performance. Overall experience for development should be better than the default optimization level -O0. Note that -Og does not imply -g, it simply disables optimizations that may interfere with debugging.
```

It turns out it can trigger more warning errors than the default mode. See #3711 for more detauls

kokoro/linux/bazel/build.sh
```diff
@@ -7,3 +7,4 @@ cd $(dirname $0)/../../..
 
 git submodule update --init --recursive
 bazel test :protobuf_test
+cat /tmpfs/tmp/bazel/execroot/com_google_protobuf/bazel-out/k8-fastbuild/testlogs/protobuf_test/test.log
```
xfxyjwf: Is this path stable? Can we use a path that doesn't depend on cpu or build type?

TeBoring: I will remove it before merge.

TeBoring:
ruby failure is not related to this change.

PR #4836, opened 2018-06-26
Fix capsule name.

ThomasColthurst:
I'm not entirely sure how the capsule name works, but this string empirically works (i.e., PyCapsule_Import(proto2::python::PyProtoAPICapsuleName(), 0) returns the right thing)
and the old one didn't.


ThomasColthurst:
Ping.

anandolee:
Internal convert file also need to be fixed to prevent future same error. I will look into it

PR #4839, opened 2018-06-28
Enable ignoring unknown in json parsing in php

TeBoring:
Also refactored conformance test to allow different option during json parsing.


TeBoring:
Broken tests are not related to this change.

PR #4843, opened 2018-06-28
Use the context ClassLoader where appropriate

### LINKED ISSUE #3474: Issues with Protobuf and Spark

jhaber:
Fix for #3474, thanks for the quick response @xfxyjwf 

I kept the existing behavior on Android, because grpc/grpc-java#1273 suggests that using the context `ClassLoader` may not be relied upon. I wasn't aware of any way to reuse the grpc code, so I just copied the relevant bits over.

I'm also not entirely sure about the change to `java_file.cc`, I don't have much experience with extensions but I think we'd want the same behavior there. 

I followed the instructions [here](https://github.com/google/protobuf/tree/master/src) to build and then I ran the Maven build and all the tests passed. Setting up a unit test to verify that it fixes #3474 would be a bit involved but I'm happy to do it if you think it's worthwhile.


jhaber:
I fixed the merge conflicts, but there are still some refactorings to do based on upstream changes

jhaber:
Interestingly, grpc-java has gone the opposite direction and removed usage of the context classloader: https://github.com/grpc/grpc-java/issues/2375

But it's still used by Guava [here](https://github.com/google/guava/blob/fd919e54a55ba169dc7d9f54b7b3485aa7fa0970/guava/src/com/google/common/io/Resources.java#L191-L194) (and in the android [version](https://github.com/google/guava/blob/fd919e54a55ba169dc7d9f54b7b3485aa7fa0970/android/guava/src/com/google/common/io/Resources.java#L191-L194)). 

If using the context classloader is a concern, some tweaks that come to mind:
- more narrowly scope the changes to the readResolve method 
- try with the context classloader first and retry without a classloader argument if a `ClassNotFoundException` is thrown

jhaber:
Going to close this PR based on the feedback [here](https://github.com/google/protobuf/issues/3474#issuecomment-401096085) and open a new one

PR #4846, opened 2018-06-28
Fix to allow a compilation under mips big endian with gcc

ckhardin:
The __MIPSEL__ is mips little endian, specific and we needed
to compile for a big endian mips target that also happened
to be a mips64, so replaced the __MIPSEL__ with a __mip__
which should preserve the build on little endian and allow a
build for a big endian setup.

	$ echo | mips64-linux-gcc -E -dM -
	... snip snip ...
	#define __mips__ 1
	#define __MIPSEB__ 1

Signed-off-by: Charles Hardin <ckhardin@gmail.com>


ckhardin:
I have signed the CLA - 3 times now, and get the error that the CLA is already signed. So, whatever that bot is checking is beyond me.

PR #4857, opened 2018-07-03
Ask to use local file if $TARGET not found in maven repo

nashimus:
Minor tweak to facilitate the creation of release zip where $TARGET is not found in the Maven repository.

E.g. 

[ERROR] Failed to download http://repo1.maven.org/maven2/com/google/protobuf/protoc/3.6.0/protoc-3.6.0-linux-ppcle_64.exe
Use local file (./target/protoc.exe)? [y/N] y
[INFO] Successfully created /home/nashimus/protobuf/protoc-artifacts/dist/protoc-3.6.0-linux-ppcle_64.zip

### DIFF
--- a/kokoro/release/linux/build_artifacts.sh
+++ b/kokoro/release/linux/build_artifacts.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -ex
+
+# change to repo root
+cd $(dirname $0)/../../..
+
+source kokoro/release/linux/prepare_build.sh
+
+# ruby environment
+source kokoro/release/linux/ruby/ruby_build_environment.sh
+
+# build artifacts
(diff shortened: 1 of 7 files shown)


nashimus:
I signed it!

xfxyjwf:
@nashimus This script is used by us to create release zips and I think it's better to avoid manual confirmation in the process. It looks like we can cross-compile for ppcle but haven't uploaded the binary to maven. A better solution is to update our release process to update the ppcle binary to maven as well. Could you help send us a pull request to update the instructions here instead?
https://github.com/google/protobuf/tree/master/protoc-artifacts#to-push-artifacts-to-maven-central

It hasn't mentioned ppcle build at the moment.

Issue #4086, opened 2017-12-22
//:protobuf_python with use_fast_cpp_protors does not compile with python3.7 due to python's breaking changes in C API


qzmfranklin:
How to reproduce:

```
git clone https://github.com/qzmfranklin/torrent
git submodule update --init --recursive
bazel build third_party/cc/protobuf:protobuf_python --define use_fast_cpp_protos=true
```

The `bazel -s` output:
```
ERROR: /home/zhongming/git/torrent/third_party/cc/protobuf/BUILD:668:1: C++ compilation of rule '//third_party/cc/protobuf:python/google/protobuf/pyext/_message.so' failed (Exit 1): clang failed: error executing command 
  (cd /home/zhongming/.cache/bazel/_bazel_zhongming/252308e7db9d6378926a9d84bb91921c/execroot/torrent && \
  exec env - \
    PWD=/proc/self/cwd \
[...]

PR #4862, opened 2018-07-03
Add Python 3.7 compatibility

benmwebb:
Compilation of Python wrappers fails with Python 3.7 because
the Python folks changed their C API such that
PyUnicode_AsUTF8AndSize() now returns a const char* rather
than a char*. Add a patch to work around. Relates #4086.

### DIFF
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -56,7 +56,7 @@
   #endif
   #define PyString_AsStringAndSize(ob, charpp, sizep) \
     (PyUnicode_Check(ob)? \
-       ((*(charpp) = PyUnicode_AsUTF8AndSize(ob, (sizep))) == NULL? -1: 0): \
+       ((*(charpp) = const_cast<char*>(PyUnicode_AsUTF8AndSize(ob, (sizep)))) == NULL? -1: 0): \
        PyBytes_AsStringAndSize(ob, (charpp), (sizep)))
 #endif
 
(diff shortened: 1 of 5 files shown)


benmwebb:
I signed it!

anandolee:
Is it the only ting that we need to fix for python3.7? Can you also add python3.7 in tests.sh as well?

benmwebb:
> Is it the only thing that we need to fix for python3.7?

To the best of my knowledge, yes.

> Can you also help add python3.7 in tests.sh as well

I don't know enough about your build system to confidently do that, but it looks like you run this on Travis, and I do know that making Python 3.7 work on Travis is a little involved because they require a newer OpenSSL than is present in Ubuntu Trusty - see travis-ci/travis-ci#9069.

anandolee:
Thanks, we can add the 3.7 tests in other RPs.

anandolee:
CLA is still complain.  Can you double check?

benmwebb:
> CLA is still complain. Can you double check?

Just checked at https://cla.developers.google.com/clas and it shows I have a Google Individual CLA signed Jul 03, 2018 09:57 PDT.

anandolee:
Are you using a different email address on your commit. It shows you are using ben@salilab.org for the commit, are you using this email to sign CLA and this email is set on your github account

benmwebb:
Right, ben@salilab.org is the commit email and is tied to my GitHub account. It is also an alternate email for my Google account, so all should be good there too. I just updated the CLA to add my GitHub username - maybe that will fix things at your end.

anandolee:
@benmwebb  I noticed that #4086 also mentioned about PyString_AsStringAndSize? Can you double check if this should be fixed as well?

acozzette:
@handlerbot We don't have a set timeline for when our next release will be, but we will most likely do a 3.7.0 release before the end of the year (we probably won't be doing any more 3.6 patch releases).

kkdd:
I look forward to python-3.7-compatible protobuf 3.6.x release, beause protobuf 3.6.1.1 or earlier is incompatible with it.

PR #4872, opened 2018-07-06
Fix SFINAE in 'RepeatedPtrFieldBase::Add'

demon386:
Without this the error of un-moveable GenericTypeHandler will pop until
'TypeHandler::New(arena_, std::move(value))' is called.


demon386:
I signed it!

acozzette:
@demon386 Thanks for the fix!

PR #4881, opened 2018-07-09
Add C++ Cocoapods support

codingrhythm:
This PR adds C++ support for Cocoapods.

### DIFF
--- a/Protobuf-C++.podspec
+++ b/Protobuf-C++.podspec
@@ -1,0 +1,60 @@
+# This file describes to Cocoapods how to integrate the C++ runtime into a dependent
+# project.
+# Despite this file being specific to C++, it needs to be on the root of the repository.
+# Otherwise, Cocoapods gives trouble like not picking up the license file correctly, or not letting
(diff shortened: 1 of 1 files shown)


codingrhythm:
I signed it!

TeBoring:
Could you explain more about why we need to support this support? Doesn't current ```./autogen.sh && ./configure && make``` work? Or shall we use homebrew?

codingrhythm:
@TeBoring thanks for your response. So one stupid question here, how can I use `./autogen.sh && ./configure && make` to build binary for mobile platforms (e.g. fat binary for iOS). I am new to this build system.

TeBoring:
http://blog.omega-prime.co.uk/2011/03/08/how-to-build-3264-bit-fat-universal-binaries/
You could define ```CXXFLAGS="-arch i386 -arch x86_64" CFLAGS="-arch i386 -arch x86_64" LDFLAGS="-arch i386 -arch x86_64"```

codingrhythm:
@TeBoring Thanks. This PR is not longer valid then.

PR #4878, opened 2018-07-06
Fix initialization with Visual Studio

acozzette:
It appears that Visual Studio does not work well with std::once_flag
because it has a bug causing it to initialize that during dynamic
initialization instead of constant initialization. This change works
around the problem by using function static initializers instead.

@gerben-s originally wrote this change for the Google-internal codebase
but I am just cherry-picking it here.

This fixes #4773.

PR #4882, opened 2018-07-09
Merge 3.6.x into master

### LINKED ISSUE #4773: global std::once_flag variables don't work with Visual Studio
Symptom: statically built protoc binary check-fails.
https://github.com/google/protobuf/issues/4409
```
[libprotobuf FATAL D:\Google-Protobuf\src\src\google\protobuf\generated_message_reflection.cc:2372] CHECK failed: file != NULL: 
```
Cause:
Visual studio compiler bug:
https://developercommunity.visualstudio.com/content/problem/262083/compiler-emits-dynamic-initializer-for-variable-wi.html

Similar bug: https://developercommunity.visualstudio.com/content/problem/1044/compiler-emits-dynamic-initializer-for-constexpr-v.html

The compiler generates dynamic-initialization code for global std::once_flag variables even though the constructor of once_flag is constexpr. As a result, the following code is broken:
https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.cc#L1343-L1362 
```
GOOGLE_PROTOBUF_DECLARE_ONCE(generated_pool_init_);
static void InitGeneratedPool() {
  generated_database_ = new EncodedDescriptorDatabase;
  generated_pool_ = new DescriptorPool(generated_database_);
  generated_pool_->InternalSetLazilyBuildDependencies();
}
void InitGeneratedPoolOnce() {
  ::google::protobuf::GoogleOnceInit(&generated_pool_init_, &InitGeneratedPool);
}
```

1. InitGeneratedPoolOnce() is called before generated_pool_init_ is dynamically initialized. It will call InitGeneratedPool() and create EncodeDescriptorDatabase/DescriptorPool for the first time.
2. dynamical initialization of generated_pool_init_ happens and it simply zeros the variable's content.
3. InitGeneratedPoolOnce() is called again and it calls InitGeneratedPool() again which creates new EncodedDescriptorDatabase/DescriptorPool for the second time.

Anything registered with the descriptordatabase/pool created in (1) will get lost after (3), and as a result descriptors that should already be in the global pool appear missing and this triggers the following check failure:
https://github.com/google/protobuf/blob/master/src/google/protobuf/generated_message_reflection.cc#L2370-L2372
```
  const ::google::protobuf::FileDescriptor* file =
      ::google::protobuf::DescriptorPool::generated_pool()->FindFileByName(filename);
  GOOGLE_CHECK(file != NULL);
```
We can workaround the problem by moving the global once_flag into a static variable inside the init function:
```
void InitGeneratedPoolOnce() {
  static ::google::protobuf::ProtobufOnceType generated_pool_init_;
  ::google::protobuf::GoogleOnceInit(&generated_pool_init_, &InitGeneratedPool);
}
```
This way, the variable is guaranteed to be initialized before GoogleOnceInit() and it appears to work with Visual Studio 2017.

acozzette:
We need to merge 3.6.x into master so that master picks up #4878, which fixes #4773 (MSVC initialization problems with debug builds).

### DIFF
--- a/kokoro/release/linux/build_artifacts.sh
+++ b/kokoro/release/linux/build_artifacts.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -ex
+
+# change to repo root
+cd $(dirname $0)/../../..
+
+source kokoro/release/linux/prepare_build.sh
+
+# ruby environment
+source kokoro/release/linux/ruby/ruby_build_environment.sh
+
+# build artifacts
+bash kokoro/release/linux/ruby/ruby_build.sh
--- a/kokoro/release/linux/prepare_build.sh
+++ b/kokoro/release/linux/prepare_build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Move docker's storage location to scratch disk so we don't run out of space.
+echo 'DOCKER_OPTS="${DOCKER_OPTS} --graph=/tmpfs/docker"' | sudo tee --append /etc/default/docker
+# Use container registry mirror for pulling docker images (should make downloads faster)
+# See https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring
+echo 'DOCKER_OPTS="${DOCKER_OPTS} --registry-mirror=https://mirror.gcr.io"' | sudo tee --append /etc/default/docker
+sudo service docker restart
+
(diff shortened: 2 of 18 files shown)


xfxyjwf:
Thanks Adam!

PR #4898, opened 2018-07-11
fix for API change in PHP 7.3

TeBoring:


TeBoring:
broken tests are not related to this change.

PR #4903, opened 2018-07-12
fix commonjs js provide GOOGLE_CHECK

BSBandme:
Additional fix for #4299

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -1681,7 +1681,7 @@ void Generator::GenerateProvides(const GeneratorOptions& options,
       if (options.import_style == GeneratorOptions::kImportCommonJsStrict) {
         string namespaceObject = *it;
         // Remove "proto." from the namespace object
-        GOOGLE_CHECK(namespaceObject.compare(0, 6, "proto."));
+        GOOGLE_CHECK(namespaceObject.compare(0, 6, "proto.") == 0);
         namespaceObject.erase(0, 6);
         printer->Print("goog.exportSymbol('$name$', null, proto);\n", "name",
                 namespaceObject);


src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -1681,7 +1681,7 @@ void Generator::GenerateProvides(const GeneratorOptions& options,
       if (options.import_style == GeneratorOptions::kImportCommonJsStrict) {
         string namespaceObject = *it;
         // Remove "proto." from the namespace object
-        GOOGLE_CHECK(namespaceObject.compare(0, 6, "proto."));
+        GOOGLE_CHECK(!namespaceObject.compare(0, 6, "proto."));
         namespaceObject.erase(0, 6);
```
TeBoring: If there is a check for that, do we still need to erase it explicitly?

BSBandme: Sorry but what do you mean by "erase it explicitly"? Do you mean change to GOOGLE_CHECK for erasing?

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -1681,7 +1681,7 @@ void Generator::GenerateProvides(const GeneratorOptions& options,
       if (options.import_style == GeneratorOptions::kImportCommonJsStrict) {
         string namespaceObject = *it;
         // Remove "proto." from the namespace object
-        GOOGLE_CHECK(namespaceObject.compare(0, 6, "proto."));
+        GOOGLE_CHECK(!namespaceObject.compare(0, 6, "proto."));
```
xfxyjwf: Please do:
```
GOOGLE_CHECK(namespaceObject.compare(0, 6, "proto.") == 0);
```

Treating the return value of compare as bool is apparently confusing and nobody caught the error in the old code.

xfxyjwf:
Thanks for the fix. Can you also help check why it didn't fail on Linux before? Such an error should be caught everywhere.

PR #4905, opened 2018-07-12
fix linux kokoro build in docker

BSBandme:


BSBandme:
The linux build is inside docker, the run docker shell is [kokoro/linux/build_and_run_docker.sh](https://github.com/google/protobuf/blob/master/kokoro/linux/build_and_run_docker.sh), this shell run the shell [kokoro/linux/pull_request_in_docker.sh](https://github.com/google/protobuf/blob/master/kokoro/linux/pull_request_in_docker.sh) to run building and testing. Previously pull_request_in_docker.sh will always return true even test failed. I just let it return false if there's test failed.

But it seems that there're too many test failed. Should we enable this?

xfxyjwf:
So many of our tests are failing without us noticing? That's pretty bad... Let's merge this PR first though. The failing tests can be fixed in subsequent PRs. Can you create an issue for each failing tests?

BSBandme:
@TeBoring, I think the two new fails are all about PHP, would you please take a look?

TeBoring:
I will have a fix for them.

BSBandme:
Thx! Do you want me to submit first or after your fix?

TeBoring:
I can fix it first. I’ll let you know when I finish.

TeBoring:
Check #4908

PR #4913, opened 2018-07-13
Update protoc build scripts.

xfxyjwf:
1. Changed maven script to only do artifact uploading and removed build
script invocation from it. We didn't use maven to invoke the build
script before (we built protoc manually and editted pom.xml to only do
uploading for previous releases), and will not use it in the future (we
will use kokoro to build artifacts).
2. Cleaned up build-protoc.sh and README.md: removed the part about
using maven to build and listed supported platforms explicitly.


xfxyjwf:
@acozzette  I'm submitting these changes to test it with kokoro. Please take a look when you have a chance. I can address issues in follow-up PRs.

PR #4940, opened 2018-07-19
Add Python 3.7 to testing

cclauss:
Given that #4862 has recently been merged, it is possible to start testing Python 3.7?  There might be issues related to __OpenSSL 1.0.2+__ similar to https://github.com/travis-ci/travis-ci/issues/9069

Probably need to step up to __Ubuntu xenial__ here:
* https://github.com/google/protobuf/blob/master/.travis.yml#L87


cclauss:
@benmwebb Your review please.

.travis.yml
```diff
@@ -84,7 +84,7 @@ matrix:
     - os: linux
       env: CONFIG=python_compatibility
       sudo: required
-      dist: trusty
+      dist: xenial  # Required for Python 3.7
```
Croydon: Xenial is not natively available on Travis. You would need to use a Docker image.

See https://docs.travis-ci.com/user/reference/overview/

cclauss: At the bottom of travis-ci/travis-ci#9069 you will see that a substantial number of significant Python projects are already relying on Travis CI's Xenial virtual machines.  This PR recommends that we follow that trend.  If Travis CI removes or disables their existing Xenial and Bionic virtual machines then we could adopt the Docker image approach as a workaround.

Croydon: I missed that Xenial seems to be available now as some kind of public beta. Thanks for the link!

anandolee:
Please resolve the conflict and I will give another kokoro run to see if tests can pass

TeBoring:
Any update to this change?

TeBoring:
Fixed by #5475

PR #4941, opened 2018-07-19
Make assertEquals work for message

TeBoring:
This change makes protobuf message checkable via assertEquals. Although
this change expose the array representation for assertEquals, we don't guarantee
the content of the array not changing. Currently, the assertEquals still
cannot check difference regarding to oneof, repeated and map fields.


TeBoring:
Actually, I would prefer another approach to provide an equal method on message.
The approach in this change exposes array representation, which not necessarily contains all information about a message. Would like to hear more feedback.

TeBoring:
See #4946

Issue #3319, opened 2017-07-06
protobuf release-v3.3.0 test fails


gogomogolo:
Setup Guide: https://github.com/google/protobuf/blob/master/cmake/README.md

OS: Windows 7
cmd: Developer Command Prompt for VS 2017

Testing Step at the Guide, protobuf is failed in 6 tests

![image](https://user-images.githubusercontent.com/29948637/27910194-c056ff78-625c-11e7-827a-5832d90444c7.png)

![image](https://user-images.githubusercontent.com/29948637/27910288-24daa512-625d-11e7-93ce-30d37923cb9f.png)

xfxyjwf:
Pretty sure these are caused by line-ending conversions. You can ignore them, or you can configure your git client to not do line-ending conversions and try again.

gogomogolo:
thanks for your reply.

gogomogolo:
https://github.com/google/protobuf/issues/547 
https://github.com/google/protobuf/issues/2548

solution : 
As a workaround, I just ended up doing: 

``` git config --global core.autocrlf false ```

while cloning on Windows and the tests passed.

Issue #2548, opened 2016-12-28
Unit tests fail on windows


koubaa:
See https://github.com/google/protobuf/issues/547

I reproduced these test failures on protobuf 3.1.0 using MSVC2012.  (the compiler actually does not matter)
[ FAILED ] 5 tests, listed below:
[ FAILED ] BootstrapTest.GeneratedDescriptorMatches
[ FAILED ] RubyGeneratorTest.GeneratorTest
[ FAILED ] TextFormatMapTest.Sorted
[ FAILED ] TextFormatTest.Basic
[ FAILED ] TextFormatExtensionsTest.Extensions

There are 9 files that need to be converted to linux line endings and the tests pass:
src/google/protobuf/testdata/map_test_data.txt
src/google/protobuf/testdata/text_format_unittest_data_oneof_implemented.txt
src/google/protobuf/testdata/text_format_unittest_extensions_data.txt
src/google/protobuf/descriptor.pb.h
src/google/protobuf/descriptor.pb.cc
src/google/protobuf/compiler/plugin.pb.h
src/google/protobuf/compiler/plugin.pb.cc
src/google/protobuf/compiler/ruby/ruby_generated_code_pb.rb
src/google/protobuf/compiler/ruby/ruby_generated_code.proto

I think the tests should not be sensitive to this since when checking out the git reposity on windows, the git default is to checkout windows-style.  A helper to do a line-by-line file diff could be used instead of comparing the entire file.

koubaa:
Python tests reveal an additional two files:

src/google/protobuf/testdata/text_format_unittest_extensions_data_pointy.txt
src/google/protobuf/testdata/text_format_unittest_data_pointy_oneof.txt

gerben-s:
This seems all related to CRLF

koubaa:
I'm curious, was this fixed on master or is this not a priority?

nehaljwani:
As a workaround, I just ended up doing: `git config --global  core.autocrlf false` while cloning on Windows and the tests passed.

Issue #547, opened 2015-06-26
Several protobuf unit tests fail when compiled with VS 2010 with x64


nslay:
See reproduction notes below.

Summary:
[  PASSED  ] 1477 tests.
[  FAILED  ] 16 tests, listed below:
[  FAILED  ] EncodeDecodeTest.Encode
[  FAILED  ] EncodeDecodeTest.Decode
[  FAILED  ] EncodeDecodeTest.Partial
[  FAILED  ] EncodeDecodeTest.DecodeRaw
[  FAILED  ] EncodeDecodeTest.UnknownType
[  FAILED  ] EncodeDecodeTest.ProtoParseError
[  FAILED  ] BootstrapTest.GeneratedDescriptorMatches
[  FAILED  ] GeneratedDescriptorTest.IdenticalDescriptors
[  FAILED  ] RubyGeneratorTest.GeneratorTest
[  FAILED  ] TextFormatMapTest.Sorted
[  FAILED  ] MessageTest.ParseFromFileDescriptor
[  FAILED  ] MessageTest.ParsePackedFromFileDescriptor
[  FAILED  ] TextFormatTest.Basic
[  FAILED  ] TextFormatTest.ParseBasic
[  FAILED  ] TextFormatExtensionsTest.Extensions
[  FAILED  ] TextFormatExtensionsTest.ParseExtensions

16 FAILED TESTS

Reproduce:
I used gmock-1.7.0 from an old Google Code page of gmock.

Making it build in VS 2010:
You'll want to replace the following in several places
isnan() with _isnan()
isfinite() with _finite()

VS 2010 acts dumb and complains about not initializing const members of a union. Make all const members of the union non-const (datapiece.h). While VS 2010 includes quite a bit of C++11, I'm not sure how well it works with that non-POD data member in the union.

VS 2010 complains about passing NULL to a std::pair<char *, int> constructor since it treats NULL as an integer. You'll need to cast NULL to (char *) around line 154 in json_util_test.cpp while initializing last_segment_.

There's also a scary lseek() #define somewhere. I forget where, but it doesn't actually perform a seek.

Other than that, it was pretty easy to setup, compile and run. Thanks for the convenient CMakelists.txt!

nslay:
I had another look at this today. While stepping through, I found that the executable expects to find files in the src directory. I ran it with protobuf source directory as the current working directory. It passes most of the tests that it failed before. Sadly, it still doesn't past all tests.

[...]

PR #4943, opened 2018-07-20
When running unit tests, read reference files in text mode

Stevie-O:
This should put an end to all of the reports of unit test failures on Windows machines: #547, #3319, #2548


Stevie-O:
I signed it!

acozzette:
@Stevie-O Thanks for the fix!

PR #4946, opened 2018-07-20
[WIP] Make assertEquals work for message

TeBoring:
This change makes protobuf message checkable via assertEquals. Although
this change expose the array representation for assertEquals, we don't guarantee
the content of the array not changing.

### DIFF
--- a/php/tests/generated_class_test.php
+++ b/php/tests/generated_class_test.php
@@ -1342,4 +1342,37 @@ class GeneratedClassTest extends TestBase
 
         TestUtil::assertTestMessage($m);
     }
+
+    #########################################################
+    # Test message equals.
+    #########################################################
+
+    public function testMessageEquals()
+    {
+        $m = new TestMessage();
+        TestUtil::setTestMessage($m);
+        $n = new TestMessage();
+        TestUtil::setTestMessage($n);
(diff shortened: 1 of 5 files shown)


php/tests/generated_class_test.php
```diff
@@ -1342,4 +1342,37 @@ public function testArrayConstructor()
...
+        $n->setOptionalString("b");
+        $this->assertNotEquals($m, $n);
+
+        $m = new TestMessage();
+        $subm = new Sub(['a' => 1]);
+        $m->setOptionalMessage($subm);
+        $n = new TestMessage();
+        $subn = new Sub(['a' => 2]);
+        $n->setOptionalMessage($subn);
+        $this->assertNotEquals($m, $n);
```
bshaffer: I think it would be better to have two tests, `testMessageEquals` and `testMessageNotEquals`, which would have all the same tests here (for int, string, and nested message), but test both cases.

tmatsuo: Can you add a test for comparing arrays of Messages? It goes a bit different code path, so I want to make sure it's also working fine.

bshaffer:
Why was this closed @TeBoring? We would like to revisit this feature.

PR #4947, opened 2018-07-20
Make assertEquals pass for message

TeBoring:
This change only makes assertEquals pass for message in c extension. However, it actually does nothing.
This is the same behavior before 3.6.0 release.


TeBoring:
I don't think broken tests are related to this change.

PR #4948, opened 2018-07-20
3.5.1.1

chenmingsf:


xfxyjwf:
Bo, why is there a 3.5.1.1 branch? Can we delete that branch? All 3.5.x releases should be made from the 3.5.x branch.

PR #4954, opened 2018-07-21
Add handshake for testing optional features in conformanc test

TeBoring:
Before sending a request for optional feature, test runner first send a handshake request to testee.
The handshake request has a string field "optional_test_name" which is the name of test to be tested.
If testee thinks it's ok to run the specified test, it should set the "optional_test_needed" in the
response. Test runner will run the actual test if found "optional_teset_needed" is set.

Some optional tests require additional changes in testee. Therefore, some testees may have simply
added these tests to their failure list instead of making the testee work. For these testees, they need
to remove those tests from their failure list after this change.

For other testee, this change won't break them.


conformance/conformance.proto
```diff
@@ -84,6 +84,14 @@ message ConformanceRequest {
...
 
   bool ignore_unknown_json = 5;
+
+  // The name of optional test, e.g.,
+  // Required.Proto3.JsonInput.IgnoreUnknownJsonFalse.ProtobufOutput. If this
+  // name is set in the ConformanceRequest, testee program should set true to
+  // the optional_test_needed in the returned ConformanceResponse if the test
+  // is needed. The test runner will only run the optional test if the
+  // optional_test_needed is true in the returned ConformanceResponse. 
+  string optional_test_name = 6;
```
xfxyjwf: Bo, can you start a discussion (share a doc) about this first?

PR #4956, opened 2018-07-22
Added support for Visual Studio 2017 in gitignore

mkosieradzki:
Visual studio keeps a lot of own state in .vs folder. This makes development much easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,3 +186,6 @@ cmake/cmake-build-debug/
 # Common build subdirectories.
 ./.build/
 ./_build/
+
+# Visual Studio 2017
+.vs


.gitignore
```diff
@@ -186,3 +186,6 @@ cmake/cmake-build-debug/
 # Common build subdirectories.
 ./.build/
 ./_build/
+
+# Visual Studio 2017
+.vs
```
xfxyjwf: How could this ".vs" directory end up in our repository root? I think we typically use a separate build tree for vs project files and all VS files will be put in a separate directory like "vs2017".

mkosieradzki: VS2017 uses "open folder" approach for entire repository, and has nice CMake integration today. You only switch to the VS Project system to build C# code.

PR #2847, opened 2017-03-14
Ruby: fixed Message#to_h for map fields.

haberman:
Fixes: https://github.com/google/protobuf/issues/1761

This integrates some of the code from this PR: https://github.com/google/protobuf/pull/396

Issue #1761, opened 2016-07-08
Ruby: message with Google::Protobuf::Map cannot be converted to Hash


mxmCherry:
Ruby protobuf message cannot be converted to hash if it uses `map<...>` fields.

Probably, because of https://github.com/google/protobuf/blob/master/ruby/ext/google/protobuf_c/message.c#L371

---

Gem version:

```
gem 'google-protobuf', '3.0.0.alpha.5.0.5.1'
```

Proto:

```
// compile:
//   protoc --ruby_out=. xxx.proto

syntax = "proto3";
package xxx;

message WithMap {
    map<string,string> map = 1;
}
```

Compiled proto:

```
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: xxx.proto

require 'google/protobuf'

Google::Protobuf::DescriptorPool.generated_pool.build do
  add_message "xxx.WithMap" do
    map :map, :string, :string, 1
  end
end

module Xxx
  WithMap = Google::Protobuf::DescriptorPool.generated_pool.lookup("xxx.WithMap").msgclass
end
```

This fails:

```
require_relative 'xxx'

p Xxx::WithMap.new.to_h
```

With:

```
debug.rb:3:in `to_h': wrong argument type Google::Protobuf::Map (expected Google::Protobuf::RepeatedField) (TypeError)
    from debug.rb:3:in `<main>'
```

mxmCherry:
Seems to be resolved in https://github.com/google/protobuf/pull/396 , but still not merged.

acoffman:
I am also experiencing this issue; is there any expectation that this will be fixed in the near future? I would offer to try and patch it myself, but it looks like there's an outstanding PR that fixes it already there. Is there something I can do to help move it along?

ebenoist:
Agreed-- what can be done here to get this fix in?

haberman:
Wow, sorry this bug went unfixed for so long. I've assigned it to myself now.

PR #396, opened 2015-05-16
ruby: Map to_h fixes

skippy:
map needed a #to_h method

Also:
- moving back to strings for hash keys (a change from the previous commit #338) to keep it consistent with type definitions in maps, such as defining 'string' as the key type.  
- the behavior of RepeatedFields#each_index was not correct.  The array and enumerable
  behaviors are different (who knew?!), so making this equivalent to the array version
- merge in upstream changes from master.  **WARNING** under jruby the expected protobuf binary test fails, meaning it generates a different protobuf than the exact same msg run under MRi.  This seems to be an issue upstream in the java library than in this code base.
- Fixing jruby errors when run under jruby-9000.pre1.  Their is a jruby AST parse error that required odd code to work around in a few of the tests.
- Found a bug with the `encode_json` method.  When run under MRI, it does not return any default values, while JRuby does.  It should probably return default values.

PR #4962, opened 2018-07-24
Fix nested Message objects in to_h

oggy:
For example, for this proto file:

```
syntax = "proto3";
import "google/protobuf/struct.proto";

message Parent {
  map<string, Child> children = 1;
}

message Child {
  string name = 1;
}
```

We used to have:

```
irb(main):001:0> Parent.new(children: {'a' => Child.new(name: 'a')}).to_h
=> {:children=>{"a"=>{:name=>"a"}}}
```

Now:

```
irb(main):001:0> Parent.new(children: {'a' => Child.new(name: 'a')}).to_h
=> {:children=>{"a"=><Child: name: "a">}}
```

It appears this changed in #2847 / #396, but it's unclear what the intent
was with these lines, so I'm assuming this was an unintended side-effect
(it's not needed to solve #1761).

### DIFF
--- a/ruby/ext/google/protobuf_c/map.c
+++ b/ruby/ext/google/protobuf_c/map.c
@@ -681,9 +681,6 @@ VALUE Map_to_h(VALUE _self) {
                                   self->value_type_class,
                                   mem);
 
-    if (self->value_type == UPB_TYPE_MESSAGE) {
-      value = Message_to_h(value);
-    }
     rb_hash_aset(hash, key, value);
   }
   return hash;
--- a/ruby/tests/basic.rb
+++ b/ruby/tests/basic.rb
@@ -1023,7 +1023,8 @@ module BasicTest
                             "b" => TestMessage2.new(:foo => 2)})
       expected_result = {
         :map_string_int32 => {"a" => 1, "b" => 2},
-        :map_string_msg => {"a" => {:foo => 1}, "b" => {:foo => 2}}
+        :map_string_msg => {"a" => TestMessage2.new(:foo => 1),
+                            "b" => TestMessage2.new(:foo => 2)}
       }
       assert_equal expected_result, m.to_h
     end


oggy:
CLA signed.

haberman:
The current behavior is intentional, there is even a test for it:

https://github.com/google/protobuf/blob/master/ruby/tests/basic.rb#L1026

What is the rationale for changing this? Is there prior art in Ruby that says this should be shallow instead of deep?

oggy:
Thanks for the quick responses!

@TeBoring We do have code that uses `to_h`, and expected the values of the hash to contain nested `Message` objects. We can rewrite it without too much difficulty, but I found the change in behavior surprising, and saw no mention of backwards-incompatible behavior in `CHANGES.txt`, so I assumed it wasn't intentional.

@haberman Thanks for the test pointer. I would say there is prior art. Ruby's built-in `Struct`, e.g., has a shallow `to_h`.

```
irb(main):001:0> Parent = Struct.new(:child)
=> Parent
irb(main):002:0> Child = Struct.new(:name)
=> Child
irb(main):003:0> parent = Parent.new(Child.new(name: 'a'))
=> #<struct Parent child=#<struct Child name={:name=>"a"}>>
irb(main):004:0> parent.to_h
=> {:child=>#<struct Child name={:name=>"a"}>}
```

If deep `to_h` is desired, I would have gone for either adding a deep `to_h` under a different name, or support an option like `to_h(deep: true)`. Happy to push code for either of these or something else -- let me know.

TeBoring:
@oggy Thanks for the explanation. I thought ```to_h``` is for usage like hash map element. So as long as we guarantee two messages have the same to_h result if and only if these two messages are the same. Does that make sense to you?

oggy:
@TeBoring This is *almost* true. `Hash`es compare elementwise, and a `Message#==` is defined that compares the type & fields, so any child `Message`s need to match according to `==`. The "almost" is because `to_h` discards the type of the receiving `Message`, so the "only if" part doesn't quite hold. But this is ok, right?

I've updated the test that @haberman mentioned.

haberman:
@oggy I like the idea of having a parameter to control whether it is deep or not.

For the parameter default, I would be in favor of keeping it as it is, unless there is an argument that the shallow behavior is clearly better. I hear your point that we changed the behavior of this a year ago. But that was an intentional change that was a part of a bugfix. The current behavior has been in place for over a year now, so I don't think we should change it now unless it is obviously wrong.

I hear your point about Struct. But to me a key difference is that Struct doesn't know when some of its members might also be Struct. Protobuf does know this. So the comparison doesn't seem exact.

TeBoring:
Closed this for now, since previous change was intended.

PR #4964, opened 2018-07-25
Fix conflict between MessageFactory and MergeFrom

zunger-humu:
This fixes a bug in the Python code where reflection generates classes structurally identical to, and with almost the same names as, but not actually equal to, the protoc-generated classes. This means that MergeFrom between two seemingly identical classes fails because of a type mismatch.

This breaks one of the most important ways to use MessageFactory (reflection), where someone stores both the bytes of a protobuf and the name of the proto class, and later on reads it back using the same: the resulting object created is of one of the reflected types, and any other code trying to interoperate will be using a generated type, and Hilarity Ensues.

This PR is a resurrection of a previous one (https://github.com/google/protobuf/pull/3597) which fixed this issue in `reflection.MakeClass`; that PR was closed when MakeClass was deprecated in favor of MessageFactory, but the bug persists. This PR fixes the bug in MessageFactory.


zunger-humu:
Fun note about this bug: Because MessageFactory is clever about caching, if you add the test case without also adding the fix, not only does that test case fail, but a bunch of other random test cases across protobuf start to fail for the same reason. :)

python/google/protobuf/internal/reflection_test.py
```diff
@@ -1505,6 +1508,24 @@ def testMergeFromBug(self):
     message2.MergeFrom(message1)
     self.assertFalse(message2.HasField('optional_nested_message'))
 
+  def testReflectionDoesntBreakMerging(self):
```
anandolee: Add the test to message_factory_test instead.

src/google/protobuf/map_unittest.proto
```diff
@@ -61,6 +61,11 @@ message TestMap {
   map<int32   , ForeignMessage> map_int32_foreign_message = 17;
   map<string  , ForeignMessage> map_string_foreign_message = 18;
   map<int32   , TestAllTypes> map_int32_all_types = 19;
+
```
anandolee: Are you able to reproduce the error by current .proto files? I guess google.protobuf.unittest_pb2.TestAllTypes should be good enough.

python/google/protobuf/message_factory.py
```diff
@@ -122,6 +115,32 @@ def GetMessages(self, files):
         extended_class.RegisterExtension(extension)
     return result
 
+  def ClearCache(self):
```
anandolee: Please do not add new APIs for fixing PR. 

If you want to add new APIs, we should have a API review (as a public library, we are very careful to add new APIs). New API should be supported in both pure python and cpp extension. Also need to have test coverage for it.

python/google/protobuf/message_factory.py
```diff
@@ -122,6 +115,32 @@ def GetMessages(self, files):
...
+    """Clear the MessageFactory's cache of which classes it knows about. This
+    is typically only something you need to do if you've been programmatically
+    importing _pb2.py files and want to make sure the factory understands them
+    correctly.
+    """
+    self._classes = {}
+
+  def _GetPrototypeClass(self, descriptor):
+    try:
```
anandolee: No need to add  _GetPrototypeClass(self, descriptor), just move the try/except to GetPrototype(self, descriptor) should be enough

python/google/protobuf/internal/reflection_test.py
zunger-humu: Done.

python/google/protobuf/message_factory.py
zunger-humu: Done.

python/google/protobuf/message_factory.py
zunger-humu: Done.

src/google/protobuf/map_unittest.proto
zunger-humu: Oops, good catch. The old PR needed this, the new one creates the bug just fine without it. :)

anandolee:
Thanks so much for your fix @zunger-humu, and your change is correct.

However, I just checked our internal code that someone has added a similar _concrete_class check in GeneratedProtocolMessageType which can also make your test pass. The change was added in May, you'd expect the see the fix in next release.

zunger-humu:
Sweet! But is this change in yet? I synced against git yesterday, and my
test definitely doesn't pass with this change commented out.

anandolee:
Not, the change is not in github yet. It is in google's internal code base, which will be integrated to github when preparing next release.

The integrated RP for 3.6.0 for example:
https://github.com/google/protobuf/pull/4387

anandolee:
Closing it as 3.7.0 is released

PR #4966, opened 2018-07-26
Add conformance test for php c back

TeBoring:
php c extension has different result for conformance test for different
php version and architecture. Try to add conformance back for php 7.1 c extension first.

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -521,7 +521,7 @@ build_php7.1_c() {
     pushd conformance
     make test_php_c
     popd
-  if
+  fi
 }
 
 build_php7.1_zts_c() {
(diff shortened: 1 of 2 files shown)


tests.sh
```diff
@@ -507,18 +507,21 @@ build_php7.1() {
...
   cd php/tests && /bin/bash ./test.sh 7.1 && cd ../..
-  pushd conformance
-  # make test_php_c
-  popd
+  if [ "$ENABLE_CONFORMANCE_TEST" = "true" ]
+  then
+    pushd conformance
+    make test_php_c
+    popd
+  if
```
xfxyjwf: if => fi

PR #4986, opened 2018-08-01
add missed header files for dependency issue

newle:
`make install` will add missed header files to appropriate path


newle:
I signed it!

PR #4987, opened 2018-08-01
add missed header files for dependency issue

newle:
`make install` will add missed header files to appropriate path

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -78,6 +78,8 @@ nobase_include_HEADERS =                                         \
   google/protobuf/stubs/status.h                                 \
   google/protobuf/stubs/stl_util.h                               \
   google/protobuf/stubs/stringpiece.h                            \
+  google/protobuf/stubs/stringprintf.h                           \
+  google/protobuf/stubs/strutil.h                                \
   google/protobuf/stubs/template_util.h                          \
   google/protobuf/any.pb.h                                       \
   google/protobuf/api.pb.h                                       \


src/Makefile.am
```diff
@@ -78,6 +78,8 @@ nobase_include_HEADERS =                                         \
   google/protobuf/stubs/status.h                                 \
   google/protobuf/stubs/stl_util.h                               \
   google/protobuf/stubs/stringpiece.h                            \
+  google/protobuf/stubs/stringprintf.h                           \
+  google/protobuf/stubs/strutil.h                                \
```
xfxyjwf: Are these two headers included by any other public headers?

newle: Not exactly! 

I had some problems when compile [apollo](https://github.com/ApolloAuto/apollo). And then I found that repository use this two header in its headers like [here](https://github.com/ApolloAuto/apollo/blob/master/modules/common/util/string_util.h).

So It would be nice If I can put these headers in property place when install.

xfxyjwf: Can you ask them to not use these header files? These helper utilities are not intended to be part of protobuf public API and we might change or delete them in the future. The better alternative is the absl library which we are considering migrating to:
https://github.com/abseil/abseil-cpp

Most of the code in google/protobuf/stubs are out-dated version of the same code in the abseil library...

newle: Got it!  Thanks very much.

PR #4988, opened 2018-08-01
Add spans support

mkosieradzki:
/cc: @jtattermusch 

This is the pull request for #3431  

It will consist of multiple commits introducing incremental, but reverse compatible changes.

1) [Breaking change] Changes compatibility from .NET Standard 1.0 to .NET Standard 1.1. `System.Memory` requires .NET Standard 1.1 and .NET Standard 1.0 has no real use (see https://docs.microsoft.com/en-us/dotnet/standard/net-standard). Keeping .NET Standard 1.0 compatibility will be costly in terms of maintenance a lot of `#if`'s are required.

… more to come

### DIFF
--- a/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -417,7 +417,7 @@ namespace Google.Protobuf
                 output.Flush();
 
                 ms.Position = 0;
-                CodedInputStream input = new CodedInputStream(ms, new byte[ms.Length / 2], 0, 0, false);
+                CodedInputStream input = new CodedInputStream(ms, new byte[ms.Length / 2], null, null, 0, 0, false);
 
                 uint tag = input.ReadTag();
                 Assert.AreEqual(1, WireFormat.GetTagFieldNumber(tag));
--- a/csharp/src/Google.Protobuf.Test/CodedOutputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedOutputStreamTest.cs
@@ -334,7 +334,7 @@ namespace Google.Protobuf
             }
             // Now test Input stream:
             {
-                CodedInputStream cin = new CodedInputStream(new MemoryStream(bytes), new byte[50], 0, 0, false);
+                CodedInputStream cin = new CodedInputStream(new MemoryStream(bytes), new byte[50], null, null, 0, 0, false);
                 Assert.AreEqual(0, cin.Position);
                 // Field 1:
                 uint tag = cin.ReadTag();
--- a/csharp/src/Google.Protobuf/ByteString.cs
+++ b/csharp/src/Google.Protobuf/ByteString.cs
@@ -34,6 +34,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
+using System.Security;
 using System.Text;
 #if !NET35
 using System.Threading;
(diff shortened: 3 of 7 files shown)


mkosieradzki:
@jtattermusch This is first real commit - I made sure it is fully reverse compatible. As agreed I have removed all unnecessary changes including aggressive inlining.

Let's work on this one - when it gets accepted let's continue our work.

ObsidianMinor:
I've been reading the discussion on #3431 and saw you were working on getting rid of the FieldCodecs. Since my proto2 PR for extensions will use FieldCodec, I was wondering (if your PR comes first) what do you plan to create to replace the FieldCodec?

mkosieradzki:
@ObsidianMinor If this PR will make into 3.x it must retain reverse compatibility. So FieldCodecs are here to stay. I will change the codegen to avoid using FieldCodecs, but I will not touch the FieldCodecs API.

What did I do instead of FieldCodecs? I basically inlined all the code for reading maps, repeated fields, wrapped fields and other scenarios where codes were being used.

jtattermusch:
@mkosieradzki  I thought you were going to create a PR against https://github.com/jtattermusch/protobuf/tree/span_support_experimental (which already has a few selected changes merged).

jtattermusch:
@mkosieradzki  no problem, this was the strategy I had in mind:
1. split the original PR in pieces so that it can be reviewed more comfortably without risking that we will break things.
2. merge the partial PRs one by one into an experimental branch (https://github.com/jtattermusch/protobuf/tree/span_support_experimental) so that we end up with something that a) is reviewed b) has some new usable functionality (individual PRs might not add any new functionality by themselves) c) and that we can package up and potentially run some performance benchmarks with.  With the whole functionality being split into individual PRs, it's going to be much easier to drive the process of actually adopting the code in upstream/master (because this is a bigger change, a consensus of more than owner will be needed - so I'm trying to be prepared in case we need to come up with a more formal proposal summarizing the pros and cons and providing rationale about why some of the changes were needed).
Does that make sense?

mkosieradzki:
@jtattermusch 
1. If we need to retain backwards compatibility - then I will need to make some adjustments to the original PR, but I have an idea how to deal with this issue - so that will be slightly more than splitting.
2. Unfortunately Travis has been removed from the CI/CD and I was counting to be able to induce kokoro builds using this specific repo and PRs - but I was wrong. I hope you will be able to run kokoro builds because there are tons of tests there - I have obviously no access.
3. It would be great if you could review my PRs one by one - your repo seems a great place (I went through the changes in your repo and I think it would be best if you created a clean branch from master and we started with current state of this repo and use this PR as the first one.

I am ready to proceed. Let's keep the PRs as clean as possible so we would be able to merge them back to the upstream.

jtattermusch:
@mkosieradzki   
ad 1:  sounds good.
ad 2: I certainly have the ability to run the full test suite on the span_support_experimental branch. That will be definitely needed at some point to check that we are on the right track, but given the nature of changes you're working on, just building the solution and running tests locally should be good enough for most of the changes planned.
ad 3. Yes, reviewing your PRs against the span_support_experimental branch is what i'm planning, but I think it would be better if you could rebase (or cherry-pick) your current PR on top of what's already merged in that branch (I consider the 3 PRs https://github.com/jtattermusch/protobuf/pull/3, https://github.com/jtattermusch/protobuf/pull/4, https://github.com/jtattermusch/protobuf/pull/6) as "already reviewed" and they seem compatible with what's in your current PR)

ObsidianMinor:
Another thing I'm wondering about is what's up with all the SecurityCritical markings everywhere?

jtattermusch:
Review will continue on this PR: https://github.com/jtattermusch/protobuf/pull/7

PR #4990, opened 2018-08-01
Added support for building the aarch64 and ppcle64 protoc binaries with Kokoro

acozzette:
This was fairly straightforward using the existing build-protoc.sh
script. The only problem I ran into was that the x86 Docker builds
create output directories owned by root, which caused some permission
issues. Fortunately it was easy to get around that just by doing those
Docker builds last.

### DIFF
--- a/protoc-artifacts/README.md
+++ b/protoc-artifacts/README.md
@@ -154,10 +154,10 @@ and upload these zip packages to the download section of the github
 release. For example:
 
 ```
-$ ./build-zip.sh 3.6.0
+$ ./build-zip.sh protoc 3.6.0
 ```
 
-The above command will create 5 zip files:
+The above command will create 7 zip files:
 
 ```
 dist/protoc-3.6.0-win32.zip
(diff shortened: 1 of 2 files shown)


protoc-artifacts/README.md
```diff
@@ -154,17 +154,19 @@ and upload these zip packages to the download section of the github
...
-dist/protoc-3.0.0-beta-4-osx-x86_64.zip
-dist/protoc-3.0.0-beta-4-linux-x86_32.zip
-dist/protoc-3.0.0-beta-4-linux-x86_64.zip
+dist/protoc-3.6.0-win32.zip
+dist/protoc-3.6.0-osx-x86_32.zip
+dist/protoc-3.6.0-osx-x86_64.zip
+dist/protoc-3.6.0-linux-x86_32.zip
+dist/protoc-3.6.0-linux-x86_64.zip
+dist/protoc-3.6.0-linux-aarch_64.zip
+dist/protoc-3.6.0-linux-ppcle_64.zip
```
xfxyjwf: Do we have kokoro job to build this binary? Should we omit it here?

acozzette: This script is kind of strange because it actually fetches the binaries from the Maven Central Repository and then repackages them in zip files for the GitHub releases page. But the binaries ultimately come from Kokoro since that is how they're built before they're uploaded to Maven. Maybe we should start building these zip files through Kokoro but for now it's still a separate step.

acozzette:
I went ahead and made another change to add support for ppcle64.

PR #4991, opened 2018-08-01
Merge 3.6.x branch into master

acozzette:


acozzette:
Thanks, Bo! We are good to go with respect to the CLA, because all these commits are from googlers with the exception of one commit from PR #4880, and that one PR was merged with a valid CLA.

PR #4994, opened 2018-08-02
Expand hdrs and add textual_hdrs to :protobuf_objc target.

ascherkus:
A Bazel project wishing to depend on this target, for, say, linking in the objc
runtime with protoc-generated objc files needs to have the hdrs/textual_hdrs
fully defined to avoid import/link errors.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -199,9 +199,10 @@ cc_library(
 
 objc_library(
     name = "protobuf_objc",
-    hdrs = ["objectivec/GPBProtocolBuffers.h"],
+    hdrs = glob(["objectivec/**/*.h"]),
     includes = ["objectivec"],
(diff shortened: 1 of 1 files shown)


BUILD
```diff
@@ -199,9 +199,10 @@ cc_library(
 
 objc_library(
```
thomasvl: I can't say I've looked at the build file before, but it seems like this is the same as the target `objectivec`, but building it differently.  I'd suggest making one an alias for the other to avoid breaking folks and using globs is likely better than listing all the files.  for non_arc_srcs, just use a glob that excludes `GPBProtocolBuffers.m`.

ascherkus: Thanks for the review -- I've determined an alternative way to fix this in my project so to avoid polluting the BUILD file I'll go ahead and close this PR.

PR #4995, opened 2018-08-03
Add DescriptorPool_clear function to reload protobuf definition 

### LINKED ISSUE #3814: [Ruby] Reload protobuf definition
I wan't to reload the protobuf definition in my ruby code when schema change.

But when I try to reload it I got this error:

`RuntimeError: Unable to add defs to DescriptorPool: Symtab already has a def named 'proto.example'`

To reproduce it just run twice this code:
```
  add_message "proto.example" do
    optional :test, :string, 1
  end
end
```

Is there a way to clean the `generated_pool` before reloading it ?

nishio-dens:
When developing with Ruby on Rails, I would like to reload the protobuf definition. 
If I can clean generate_pool, I can accomplish this purpose.

```
  add_message "proto.example" do
    optional :test, :string, 1
  end
end

Google::Protobuf::DescriptorPool.generated_pool.clear  # added

    add_message "proto.example" do
      optional :test, :int32, 1
    end
end
```

This fixes #3814.

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -124,6 +124,7 @@ void DescriptorPool_register(VALUE module) {
   rb_define_method(klass, "add", DescriptorPool_add, 1);
   rb_define_method(klass, "build", DescriptorPool_build, 0);
   rb_define_method(klass, "lookup", DescriptorPool_lookup, 1);
+  rb_define_method(klass, "clear", DescriptorPool_clear, 0);
   rb_define_singleton_method(klass, "generated_pool",
                              DescriptorPool_generated_pool, 0);
   rb_gc_register_address(&cDescriptorPool);
@@ -204,6 +205,21 @@ VALUE DescriptorPool_lookup(VALUE _self, VALUE name) {
     return Qnil;
   }
   return get_def_obj(def);
+}
+
+/*
+ * call-seq:
+ *     DescriptorPool.clear()
+ *
+ * Clear all registered Descriptor.
+ */
+VALUE DescriptorPool_clear(VALUE _self) {
+  DEFINE_SELF(DescriptorPool, self, _self);
+
+  upb_symtab_free(self->symtab);
+  self->symtab = upb_symtab_new();
+
+  return Qnil;
 }
 
 /*
--- a/ruby/ext/google/protobuf_c/protobuf.h
+++ b/ruby/ext/google/protobuf_c/protobuf.h
@@ -177,6 +177,7 @@ DescriptorPool* ruby_to_DescriptorPool(VALUE value);
 VALUE DescriptorPool_add(VALUE _self, VALUE def);
 VALUE DescriptorPool_build(VALUE _self);
 VALUE DescriptorPool_lookup(VALUE _self, VALUE name);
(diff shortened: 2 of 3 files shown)


nishio-dens:
@googlebot I signed it!

TeBoring:
I think the generated pool instance should be cleared automatically after each request is done. @haberman

TeBoring:
@nishio-dens Do you know whether ruby free global instance after handling each request?

nishio-dens:
@TeBoring 
What does "each request" mean?
I could not understand your suggestion correctly.

If I could call generated_pool.clear method directly, It is useful.
I'd like to use generated_pool.clear  method in the Development mode of Ruby on Rails.


For example, when I use Rails REPL(rails console), I want to do the following things.


1. I define the proto file

```
# proto/lib/proto_pb.rb

  add_message "proto.example" do
    optional :test, :string, 1
  end
end
module Proto
  Example = Google::Protobuf::DescriptorPool.generated_pool.lookup("proto.example").msgclass
end
```

2. start REPL(Rails console), and use Proto::Example class,

```
$ bundle exec rails console

Loading development environment (Rails 5.2.0)
[1] pry(main)> Proto::Example.new
=> <Proto::Example: test: "">
```

3. change the proto message definition like this,

```
# proto/lib/proto_pb.rb

  add_message "proto.example" do
    optional :test_change, :string, 1.    # <---- changed
  end
end
module Proto
  Example = Google::Protobuf::DescriptorPool.generated_pool.lookup("proto.example").msgclass
end
```

4. call reload! method in rails console (see: https://api.rubyonrails.org/classes/Rails/ConsoleMethods.html#method-i-reload-21)

Reload the environment. reload! method is super-helpful, because I do not need to restart the rails console.

```
# proto/lib/proto_pb.rb

$ be rails console                                                                                                                                                                                                                                                                                                     
Loading development environment (Rails 5.2.0)
[1] pry(main)> Proto::Example.new
=> <Proto::Example: test: "">
[2] pry(main)> reload!
Reloading...
=> true
[3] pry(main)> Proto::Example.new
=> <Proto::Example: test_change: "">
```

In order to realize this, I will hook the reload! method and call generated_pool.clear() method explicitly.

TeBoring:
Where do you plan to hook the ```generated_pool.clear()```? Could you also add it to this PR?

haberman:
I don't think we should add a `DescriptorPool#clear` method. That would be dangerous because any references to the freed descriptors would become invalid.

Instead I think we should allow setting `generated_pool` to a new, empty `DescriptorPool`.

nishio-dens:
@TeBoring 

I will plan to hook reload! method like this,

https://github.com/nishio-dens/rails_grpc/blob/master/lib/rails_grpc/railtie.rb#L17
https://github.com/nishio-dens/rails_grpc/blob/master/lib/rails_grpc/reloader.rb#L9


@haberman

> That would be dangerous because any references to the freed descriptors would become invalid.
> Instead I think we should allow setting generated_pool to a new, empty DescriptorPool.

You are right. Thank you for your advice.

I withdraw this PR, and I will PR again if I can fix it.

akimd:
@TeBoring  So what has become of this?  I see no related PR.

PR #5018, opened 2018-08-08
Disable warning about negating unsigned ints

matthauck:
> .../include/google/protobuf/wire_format_lite.h(863):
> warning C4146: unary minus operator applied to unsigned type, result
> still unsigned

### DIFF
--- a/src/google/protobuf/wire_format_lite.h
+++ b/src/google/protobuf/wire_format_lite.h
@@ -851,6 +851,13 @@ inline double WireFormatLite::DecodeDouble(uint64 value) {
 //        >> encode >>
 //        << decode <<
 
+
+#if defined(_MSC_VER) && _MSC_VER >= 1400 // VC++ 8.0
+// Disable warning about negating unsigned
+#pragma warning( push )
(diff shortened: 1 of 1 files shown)


matthauck:
Would be cool if this could make it to the 3.5.x branch too. Should I open up a second PR for that?

xfxyjwf:
Can you make the change CMakeLists.txt instead? We already have a list of disabled warnings there:
https://github.com/google/protobuf/blob/master/cmake/CMakeLists.txt#L154

xfxyjwf:
@matthauck If we don't want to ignore those warnings in general, could you help fix the source code so it doesn't trigger these warnings? It would require a bit more work but would help us keep the code cleaner.

xfxyjwf:
Thoughts on whether we can fix this warning in the code?

matthauck:
Sorry, I haven't had a chance to look yet. I'm not quite sure.

xfxyjwf:
Given how often users report about similar warnings, I think it's not a good idea to add preprocessors to disable them in the code (a lot more will show up later and I don't like the code littering with these preprocessors).

PR #5019, opened 2018-08-09
Down-integrate from google3.

xfxyjwf:


src/google/protobuf/message_lite.cc
```diff
@@ -398,13 +646,6 @@ void GenericTypeHandler<string>::Merge(const string& from,
   *to = from;
 }
 
-bool proto3_preserve_unknown_ = true;
-
-void SetProto3PreserveUnknownsDefault(bool preserve) {
-  proto3_preserve_unknown_ = preserve;
```
rchav: What was this function replaced with?

acozzette: We've removed that function and now proto3 messages always preserve unknown fields during parsing. However, you can call `message.DiscardUnknownFields()` if you want to drop them.

PR #5024, opened 2018-08-09
Fix TryCreateParentDirectory on Windows

sesmith177:
On Windows, both '/' and '\' are valid path separators. So when creating
the parent directories, split the filename on both

Signed-off-by: Akshat Gokhale <agokhale@pivotal.io>

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -168,7 +168,12 @@ bool VerifyDirectoryExists(const string& path) {
 // directories listed in |filename|.
 bool TryCreateParentDirectory(const string& prefix, const string& filename) {
   // Recursively create parent directories to the output file.
+#if defined(_WIN32)
+  // on Windows, both '/' and '\' are valid path separators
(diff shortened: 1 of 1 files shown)


sesmith177:
signed the CLA

xfxyjwf:
Can you provide an example where the current implementation doesn't work but can be fixed with this PR?

src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -168,7 +168,12 @@ bool VerifyDirectoryExists(const string& path) {
 // directories listed in |filename|.
 bool TryCreateParentDirectory(const string& prefix, const string& filename) {
   // Recursively create parent directories to the output file.
+#if defined(_WIN32)
+  // on Windows, both '/' and '\' are valid path separators
+  std::vector<string> parts = Split(filename, "/\\", true);
```
xfxyjwf: Can you use this code path always? I.e., splitting on "/\\" even on linux. I don't think we want to support "\" as part of the filename at all.

PR #5027, opened 2018-08-09
Cleanup kokoro scripts.

xfxyjwf:
Changes:
1. Remove stuff no longer needed. Lots of the heavy lifting were there
because we were running our own jenkins cluster and had to manage all
the test logs ourselves. Now they are useless.
2. Change "-j2" to "-j4" to speed up the test a little bit. Kokoro
machines have 4 logic CPUs according to their spec.


xfxyjwf:
This PR improves the running time of linux tests significantly because it has removed the C++ test that was there for every language. For example, linux Python:

Before: 22 mins
https://source.cloud.google.com/results/invocations/dd2181b6-af9c-4035-9c27-8f8695a7de18/targets

After: 10 mins
https://source.cloud.google.com/results/invocations/0f623ded-e412-41bc-bf6f-f9b0f25dfc47/targets

PR #5032, opened 2018-08-10
Fix json parsing of wrapper values

TeBoring:
Copy paste upb change into protobuf repository.


TeBoring:
Broken test is not due to this change.

PR #5034, opened 2018-08-11
Python wheel kokoro for windows

TeBoring:


kokoro/release/python/windows/build_python_env.bat
```diff
@@ -0,0 +1,10 @@
+REM Prepend newly installed Python to the PATH of this build (this cannot be
```
xfxyjwf: Why does this have to be in a separate script? Can you embed it in another existing script?

TeBoring: done

PR #5035, opened 2018-08-13
Option to print 64-bit integers as integers in JSON.

arunov:
Though this won't work in javascript, keeping this option will be useful for other languages.
Related issue: https://github.com/google/protobuf/issues/2679


xfxyjwf:
See discussion in: https://github.com/google/protobuf/issues/5015#issuecomment-411847550

PR #5040, opened 2018-08-13
Fix bazel test.

xfxyjwf:
* Fixes https://github.com/google/protobuf/issues/5031
* Fixes https://github.com/google/protobuf/issues/5037


xfxyjwf:
Note that the bazel test will only pass after this PR is merged because part of the test (bazel build examples) depends on the master branch.

PR #5042, opened 2018-08-13
Document wrapper non-uses

dsnet:
Document the cases where the wrapper well-known types are not useful.


xfxyjwf:
LGTM

Adding @acozzette to double check

acozzette:
LGTM

That sounds right to me.

PR #5044, opened 2018-08-14
Windows: expand path names on command line

laszlocsomor:
release notes: no

Fixes https://github.com/google/protobuf/issues/3957


xfxyjwf:
Is linking setargv.obj not working any more?

laszlocsomor:
@xfxyjwf : I thought it never worked so users needed workarounds like this one: https://github.com/google/protobuf/issues/3957#issuecomment-409954998

Did it actually work in newer releases?

laszlocsomor:
(FYI, I can't work in this again until 2018/08/27.)

xfxyjwf:
My understanding is that linking setargv.obj should work, but the protoc binary we release for windows is not built with bazel so it doesn't include that change. I'll test it a bit on windows and see if changing our protoc release script to also link setargv.obj can fix the issues.

laszlocsomor:
Thanks @xfxyjwf , I see.
If you would ping this PR once you're done testing, that'd be nice. Thanks!

acozzette:
@laszlocsomor Do you know what the current status of this pull request is? Feng left to join a startup, so I can take over reviewing this. I don't really have much context on the Windows path name issue here yet, though.

laszlocsomor:
@acozzette : Thanks for pinging! This PR was never merged. Protoc 3.6.1 still does not support wildcards on Windows, so I guess the current solution (linking the Microsoft-shipped `setargv.obj`) does not work, and this PR is still necessary to fix the bug.

src/google/protobuf/stubs/io_win32.cc
```diff
@@ -355,6 +360,92 @@ wstring testonly_utf8_to_winpath(const char* path) {
...
 
+bool expand_wildcards(
+    const string& path, std::function<void(const string&)> consume) {
+  if (path.find_first_of("*?") == string::npos) {
+    // There are no wildcards in the path, we don't need to expand it.
+    consume(path);
+    return ExpandWildcardsResult::kSuccess;
+  }
+
+#ifdef SUPPORT_LONGPATHS
```
acozzette: Is there a way to unify some of this logic between the `SUPPORT_LONGPATHS` and non-`SUPPORT_LONGPATHS` cases? It seems like the logic is mostly the same, so I wonder if templates would make it possible to unify the two cases.

src/google/protobuf/stubs/io_win32.h
```diff
@@ -71,6 +72,30 @@ LIBPROTOBUF_EXPORT int stat(const char* path, struct _stat* buffer);
...
+// to the current working directory, and it may contain wildcards ("*" and "?")
+// in the last path segment. This function passes all matching file names to
+// `consume`. The resulting paths may not be absolute nor normalized.
+//
+// The function returns true if the path did not contain any wildcards (in
+// which case the path may or may not exist), or the path did contain wildcards
+// and it matched at least one file.
+// The function returns false if the path contained wildcards but it did not
+// match any files.
+LIBPROTOBUF_EXPORT bool expand_wildcards(
```
acozzette: Shouldn't this return an `ExpandWildcardsResult` instead of `bool`?

laszlocsomor:
@acozzette : Thanks for the review! And sorry about the delay. Lots of other things to do. I'll get back to this thread as soon as I can.

xycui:
@laszlocsomor @acozzette any update on the PR?

laszlocsomor:
Can't trigger presubmit tests, "release notes: yes|no" label is missing.

laszlocsomor:
@xycui : Sorry for the long silence.

laszlocsomor:
@acozzette : Could you add "release notes: no" tag please?

src/google/protobuf/io/io_win32.h
```diff
@@ -75,6 +76,26 @@ PROTOBUF_EXPORT int stat(const char* path, struct _stat* buffer);
 PROTOBUF_EXPORT int write(int fd, const void* buffer, size_t size);
 PROTOBUF_EXPORT std::wstring testonly_utf8_to_winpath(const char* path);
 
+struct ExpandWildcardsResult {
+  enum {
```
acozzette: Since we're now using C++11, you could make this an `enum class` instead of just `enum`, and that will scope the enum without the need for an enclosing struct.

src/google/protobuf/io/io_win32.cc
```diff
@@ -357,6 +362,92 @@ wstring testonly_utf8_to_winpath(const char* path) {
...
 
+int ExpandWildcards(
+    const string& path, std::function<void(const string&)> consume) {
+  if (path.find_first_of("*?") == string::npos) {
+    // There are no wildcards in the path, we don't need to expand it.
+    consume(path);
+    return ExpandWildcardsResult::kSuccess;
+  }
+
+#ifdef SUPPORT_LONGPATHS
```
acozzette: Isn't `SUPPORT_LONGPATHS` always defined here since it's set on line 50?

src/google/protobuf/io/io_win32.h
laszlocsomor: Thanks! Done.

src/google/protobuf/io/io_win32.cc
acozzette: Great, I would say you can just delete a lot of this code since we only need to handle the `SUPPORT_LONGPATHS` case.

laszlocsomor: Done.

src/google/protobuf/io/io_win32.h
```diff
@@ -75,6 +76,24 @@ PROTOBUF_EXPORT int stat(const char* path, struct _stat* buffer);
 PROTOBUF_EXPORT int write(int fd, const void* buffer, size_t size);
 PROTOBUF_EXPORT std::wstring testonly_utf8_to_winpath(const char* path);
 
+enum ExpandWildcardsResult {
```
acozzette: I don't think this will work as it is--to get the scoping I believe it has to be written `enum class ExpandWildcardsResult`.

laszlocsomor: Done.

acozzette:
Thanks, @laszlocsomor!

laszlocsomor:
Thanks for following through!

PR #5045, opened 2018-08-14
Ported FieldMaskUtil from Java to C#

### LINKED ISSUE #4325: Add FieldMaskUtil support in C#
As of today, the C# implementation is missing support for working with `FieldMask`. The message itself exists in `Google.Protobuf.WellKnownTypes`, so it can be used for merging in other languages. But since the `FieldMaskUtil` is missing in C#, merging and working with `FieldMask`s in general has to be done manually.

If no one is working on it, I could also tackle it since I need it myself. Just wanted to make sure it's not done by anyone else right now or if there is a cause why it's not implemented.

Falco20019:
Fixes #4325 

I ported the functionality from Java to C#. Still a bit unsure if we should keep the classes seperate as `Utils` or just merging it all into `FieldMaskTreePartial` (would be getting a huge class...). All other `Utils` from Java have been ported into the `*Partial` classes, but those were mainly static classes anyway. The `FieldMaskUtil` has a complete `FieldMaskTree` helper class.

Right now, this would allow for easy portability by keeping the `Util` but also fit with the other classes by creating static helpers in `Partial`. Any preferences?

@anandolee offered in #4325 to do the PR


xfxyjwf:
Thanks for putting this together! Note that we are not so happy about the name "FieldMaskUtil" in Java and have a plan to rename it to "FieldMasks". I would recommend against using the name "FieldMaksUtil" in C# just because Java did.

Falco20019:
@xfxyjwf Thanks for the feedback. Would you prefer to include the FieldMaskTree in FieldMaskPartial (separate partial class), or maybe putting it in the root like JsonFormatter and other helper classes? I think it may be better to only have the stuff directly in FieldMasks and dropping the FieldMaskUtil class completely.

anandolee:
Can we remove FieldMaskUtil. I noticed FieldMaskPartial has most APIs except IsValid. FieldMaskPartial and FieldMaskUtil are kind of duplicated

Falco20019:
I plan to do so, was just waiting on a response to my question. Would you prefer having the FieldMaskTree in a partial or as separate class in root like JsonFormatter?

Falco20019:
@anandolee @xfxyjwf Still waiting for feedback on your preference how to separate `FieldMaskTree`. Either adding it to `FieldMaskPartial` which would blow it up a bit, or putting it into root like done with `JsonFormatter` but cluttering up the space there while it's only used by FieldMaskTree and completely internal?

Both are valid ways to proceed with their own up- and downsides. I would prefer the root `JsonFormatter`-like aproach.

anandolee:
I'd prefer to make FieldMaskTree in a separated file instead of blow FieldMaskPartial up. And yes FieldMaskTree should be internal.

Falco20019:
Thanks for the feedback. Just applied the changes.

csharp/src/Google.Protobuf/FieldMaskTree.cs
```diff
@@ -0,0 +1,373 @@
...
+                mask.Paths.AddRange(paths);
+            }
+
+            return mask;
+        }
+
+        /// <summary>
+        /// Gathers all field paths in a sub-tree.
+        /// </summary>
+        internal List<string> GetFieldPaths()
```
anandolee: Looks like GetFieldPaths() is only used in ToFieldMask(), I'd prefer to remove this internal API. And FieldMaskTreeTest should test ToFieldMask() instead of GetFieldPaths()

Falco20019: Will be removed in the next commit.

csharp/src/Google.Protobuf.Test/WellKnownTypes/FieldMaskTest.cs
```diff
@@ -58,5 +59,174 @@ public void ToString_Invalid_Precise()
...
+
+        [Test]
+        public void Union_UsingVarArgs()
+        {
+            FieldMask mask1 = FieldMask.FromString("foo");
+            FieldMask mask2 = FieldMask.FromString("foo.bar,bar.quz");
+            FieldMask mask3 = FieldMask.FromString("bar.quz");
+            FieldMask mask4 = FieldMask.FromString("bar");
+            FieldMask result = mask1.Union(mask2, mask3, mask4);
+            Assert.Contains("bar", result.Paths);
```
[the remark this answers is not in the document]
Falco20019: I tried avoiding to make assumptions about the order of paths in `FieldMask` and therefore prefer `Contains` over a `ToString` comparison. But I will add assertions that proof there are no elements than the expected ones.

Falco20019:
As a short summary:
I prefer `Contains` and checking the `Count` over `ToString` to avoid making assumptions on sorting and have it future proof. Also added assertions to proof specific members are not part of the created path list.

Due to the removal of `FieldMaskTree.GetFieldPaths`, `FieldMaskTreeTest` is now having asumptions based on `FieldMask` and `FieldMaskTree.ToFieldMask`.

anandolee:
Thanks for the change. It looks good to me, just wait the tests green

anandolee:
Oh for the newly added FieldMaskTreeTest.cs, you need to update Makefile.am to pass distcheck test:
https://github.com/protocolbuffers/protobuf/blob/master/Makefile.am

Falco20019:
@anandolee I hope the distcheck works now :)

anandolee:
csharp/src/Google.Protobuf/FieldMaskTree.cs is also newly added and need to be in Makefile.am

Falco20019:
Ok, commited. Is it only giving one error a time? This should be all of the newly added files. If there is anything else to check, just let me know :) (Currently I'm in the same timezone as you which should speed stuff up a bit).

PR #5050, opened 2018-08-15
Change deprecated Bazel single file attr param

jjudd:
This removes the need for --incompatible_disable_deprecated_attr_params when using Protobuf


jjudd:
I was updating our internal version of Bazel to 0.16  these errors showed up with `--all_incompatible_changes` enabled, so I fixed them.

xfxyjwf:
The amount of breaking changes in bazel are a bit annoying...

Issue #4907, opened 2018-07-13
repeated primitives in Java TextFormat


wangdf:
Dear all,

The Java TextFormat implementation doesn't seem to support the 'short form' in the c++ version for repeated primitives. 

It not only creates long-winded results in some cases, but this mismatch between languages also causes headache when try to parse messages authorized in one with another.

Any suggestions are welcome.

Best Regards,
Difu
  
For example,
data: [0, 1] in c++
vesus 
data: 0
data: 1
in java.

xfxyjwf:
We can probably add an option to Java TextFormat to do the same.

elharo:
sounds like this should simply be done unconditionally (though that risks breaking poorly written golden tests) or not done at all. I don't think an option is needed.

go/textformat-spec

shaod2:
This seems to be already [supported](https://github.com/protocolbuffers/protobuf/blob/a3c8e2deb05186334b7ee8c1174f44802e38b43d/java/core/src/test/java/com/google/protobuf/TextFormatTest.java#L1502-L1508).
Closing the issue as it's not reproducible.

PR #5064, opened 2018-08-21
print repeated primitives in Java TextFormat same as in c++

yonatankahana:
Hi, 
I saw the bug #4907 and decided to fix the printer to print the values of repeated primitives to be as in c++.
I didn't touch the parser, because the parser already knows to work with this prints already.


yonatankahana:
I signed it!

xfxyjwf:
The new behavior needs to be guarded by a new printer flag or the change will break many.

xfxyjwf:
@yonatankahana Sorry for the late response but I don't have a great idea either. It seems we are using overloads currently and we will need a new set of overloads for this flag. That doesn't look ideal though...

PR #5065, opened 2018-08-21
Add Timestamp timezone parsing error case

kastiglione:
Calling `Timestamp.FromJsonString()` with a string that contains no timezone will result in a Python `UnboundLocalError` exception. For example:

```python
Timestamp.FromJsonString('2018-08-21T18:56:51.707893')
```

This change causes a `ParseError` to be raised if a matched `-` comes from the initial `YYYY-MM-DD` date.

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -142,7 +142,8 @@ class Timestamp(object):
       timezone_offset = value.find('+')
     if timezone_offset == -1:
       timezone_offset = value.rfind('-')
-    if timezone_offset == -1:
+    # Don't match '-' in date (YYYY-MM-DD)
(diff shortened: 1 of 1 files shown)


xfxyjwf:
@googlebot

anandolee:
Thanks for the change. Please add a test for it

kastiglione:
Apologies for the silence. I'm not sure when I'll switch context to this. The development environment setup have enough steps and system state that they look like something I'd rather avoid if I could. I take it there's no CI for pull requests?

anandolee:
For this change, you can add the test without setup environment. github will run the test.

You can look at these tests for example:
https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/internal/well_known_types_test.py#L295

PR #5078, opened 2018-08-24
Added cmake build files for the conformance test runner.

haberman:
I want to use the conformance tests from upb, which is switching
to CMake.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -30,6 +30,7 @@ endif()
 
 # Options
 option(protobuf_BUILD_TESTS "Build tests" ON)
+option(protobuf_BUILD_CONFORMANCE "Build conformance tests" OFF)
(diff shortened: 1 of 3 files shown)


cmake/CMakeLists.txt
```diff
@@ -218,6 +219,10 @@ if (protobuf_BUILD_TESTS)
   include(tests.cmake)
 endif (protobuf_BUILD_TESTS)
 
+if (protobuf_BUILD_CONFORMANCE)
+  include(conformance.cmake)
```
xfxyjwf: This new file needs to be included in the toplevel Makefile.am file for the make dist test to pass.

haberman: Done.

haberman:
I think the Kokoro failure is unrelated.

PR #5085, opened 2018-08-28
Route "compiler" value of config_setting through @bazel_tools/tools/cpp:compiler.

scentini:


xfxyjwf:
Can you point me to some documentations about this change in bazel?

scentini:
Sorry for the delayed reply,

We are rerouting how the 'compiler' value is accessed in Bazel, which is a backwards incompatible change:
https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#disable-late-bound-option-defaults

I combined the old and the new way of dealing with 'compiler' in a compiler_config_setting.bzl file so it stays backwards compatible.

PR #5095, opened 2018-08-30
Fix well known type class not inheriting Message

### LINKED ISSUE #5090: PHP: well known types in C extension do not extend Message class in PHP 7.2
This makes it impossible to reliably use type hints in functions that accept a protobuf message, because passing a well known type to such a function with the protobuf extension enabled will cause a TypeError.

Version: composer version v3.6.1, c extension version 3.6.1
Language: PHP

Tested on Ubuntu Travis instance. Details:
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty

PHP 7.2.4, 5.6.37

**What did you do?**
Run the following script using PHP cli. See below for composer.json file.
```
<?php

use Google\Protobuf\FloatValue;
use Google\Protobuf\Internal\Message;

require __DIR__ . '/vendor/autoload.php';

$floatValue = new FloatValue();
echo get_parent_class($floatValue) . PHP_EOL;
echo (($floatValue instanceof Message) ? 'True' : 'False') . PHP_EOL;
```

```
# php test.php
Google\Protobuf\Internal\Message
True
```

**What did you see instead?**
Running under PHP 7.2.4, *without* protobuf c extension:
```
# php test.php
Google\Protobuf\Internal\Message
True
```
Running under PHP 7.2.4, *with* protobuf c extension:
```
# php test.php

False
```
Running under PHP 5.6.37, *without* protobuf c extension:
```
# php test.php
Google\Protobuf\Internal\Message
True
```
Running under PHP 5.6.37, *with* protobuf c extension:
```
# php test.php

True
```

Composer.json file used:
```
{
    "name": "root/test",
    "require": {
        "google/protobuf": "^3.6"
    }
}
```

cc @TeBoring @bshaffer @tmatsuo @dwsupplee @jdpedrie

TeBoring:
Previously, well known types only implement Message interface but not inhert it.
Fix the problem by calling zend_do_inheritance.

Fixes #5090

### DIFF
--- a/php/ext/google/protobuf/protobuf.h
+++ b/php/ext/google/protobuf/protobuf.h
@@ -138,7 +138,8 @@
     INIT_CLASS_ENTRY_EX(class_type, CLASSNAME, strlen(CLASSNAME),            \
                         LOWWERNAME##_methods);                               \
     LOWWERNAME##_type = zend_register_internal_class(&class_type TSRMLS_CC); \
-    LOWWERNAME##_type->create_object = message_create;
+    LOWWERNAME##_type->create_object = message_create;                       \
+    zend_do_inheritance(LOWWERNAME##_type, message_type TSRMLS_CC);
 #define PHP_PROTO_INIT_SUBMSGCLASS_END \
   }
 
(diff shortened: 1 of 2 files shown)


michaelbausor:
Change LGTM, but it looks like a lot of tests are commented out?

TeBoring:
Fixed

PR #5104, opened 2018-09-03
Fix typo in path docs.

tmc:


src/google/protobuf/descriptor.proto
```diff
@@ -785,7 +785,7 @@ message SourceCodeInfo {
     // This is because FileDescriptorProto.message_type has field number 4:
     //   repeated DescriptorProto message_type = 4;
     // and DescriptorProto.field has field number 2:
-    //   repeated FieldDescriptorProto field = 2;
+    //   repeated FieldDescriptorProto field = 7;
```
xfxyjwf: This field does have a field number 2. I think you are mistaking what "7" means in the example. "7" is the index of the entry inside the repeated field, not the field number of the repeated field.

PR #5122, opened 2018-09-07
update to Guava 20

### LINKED ISSUE #5116: Upgrade to Guava 20 for Java
Version: master/v3.6.1
Language: Java

Most of the stack is converging on/upgrading to Guava 20.0. Projects that import protobufs are likely already using 20.0 in their classpaths. 

This was tried before a year and half ago in #3086 but had to be rolled back then due to gRPC compatibility issues. However gRPC is now on Guava 20 itself so they should no longer block us.

elharo:
fix #5116

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -82,7 +82,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>19.0</version>
+        <version>20.0</version>
       </dependency>
     </dependencies>
   </dependencyManagement>


xfxyjwf:
@ejona86 @carl-mastrangelo Can you help confirm if this upgrade is OK for grpc?

ejona86:
Yes, it is fine. [grpc-java bumped to Guava 20 at end of March.](https://github.com/grpc/grpc-java/releases/tag/v1.11.0) We've not heard of any compatibility problems this time around.

PR #5125, opened 2018-09-07
Upport kokoro/release/python from 3.6.x branch

jtattermusch:
That directory is missing in upstream/master so the artifact builds are failing:

```
[09:11:59][ERROR] Failed to get build config
com.google.devtools.kokoro.config.ConfigException: Build config /tmp/workspace/workspace/protobuf/github/master/release/python/macos/release/src/github/protobuf/kokoro/release/python/macos/release.cfg does not exist
```

Copying the missing files from 3.6.x branch.

### DIFF
--- a/kokoro/release/python/linux/build_artifacts.sh
+++ b/kokoro/release/python/linux/build_artifacts.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+set -ex
+
+# change to repo root
+pushd $(dirname $0)/../../../..
+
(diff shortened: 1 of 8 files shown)


xfxyjwf:
@acozzette @TeBoring Should we do a merge from 3.6.x to master?

acozzette:
@xfxyjwf That sounds like a good idea, that way we will make sure to pick up any other fixes that might have been made on 3.6.x but which haven't made it to master yet.

PR #5126, opened 2018-09-07
PHP: Add support for primitive types in setters

michaelbausor:
This PR adds support for automatic conversion from PHP native types to protobuf wrapper types in message setters. Still pending:
- Support for repeated fields
- PHPDoc updates

@TeBoring PTAL and let me know if this is the approach you had in mind


src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -726,6 +726,11 @@ void GenerateFieldAccessor(const FieldDescriptor* field, bool is_descriptor,
       printer->Print(");\n");
     }
   } else if (field->cpp_type() == FieldDescriptor::CPPTYPE_MESSAGE) {
+    if (IsWrapperType(field)) {
```
TeBoring: In this way, you also need to implement GPBWrapperTypes in c extension.
I was thinking we could simply do the box/unbox implementation in the setter/getter.
```
public function setStringValue($val) {
  if (is_string($va)) {
    $val = new Google\Protobuf\StringValue(["value"=>$val]);
  }
  GPBUtil::checkMessage($val, Google\Protobuf\StringValue::class)
  $this->string_value_field = $val;
}
```

TeBoring: Since we have to support the same logic for repeated field, I would suggest changing checkMessage which is also shared by repeated field.

michaelbausor: Isn't that what I am doing? I am using IsWrapperType check in php_generator.cc, and using it to add the extra `"$var = GPBWrapperTypes::normalizeToMessageType($var, \\^class_name^::class);\n",` line into the generated setter.

Re `checkMessage`, if it was implemented that way, would I also need to update the c extension implementation? (here: https://github.com/protocolbuffers/protobuf/blob/master/php/ext/google/protobuf/type_check.c#L413). That was why I didn't take the approach of modifying `checkMessage`, and instead added code to the setters.

TeBoring: I can implement the c extension part. Thanks for writing the test part already.

michaelbausor: No worries! I have updated the PR to add the check in to checkMessage, and added a single test case for repeated fields. If this looks good, I will go ahead and update generated phpdocs and add complete test coverage.

php/src/Google/Protobuf/Internal/GPBUtil.php
```diff
@@ -597,4 +609,65 @@ public static function hasJsonValue($msg)
...
+        BytesValue::class,
+    ];
+
+    /**
+     * Determine if a fully qualified class name is a wrapper type.
+     *
+     * @param string $class the fully qualified class name
+     * @return bool true if this is a wrapper type
+     */
+    public static function isWrapperType($class)
```
michaelbausor: @TeBoring I would like to get your take on this method vs the existing `hasJsonValue` method. They check all the same classes, but have slightly different inputs (instance vs class name). I can reconcile into a single method, but wanted to get your input first on what you would prefer.

php/src/Google/Protobuf/Internal/GPBUtil.php
```diff
@@ -597,4 +609,65 @@ public static function hasJsonValue($msg)
...
+    {
+        return in_array($class, self::$wrapperTypes);
```
TeBoring: Will this linear search be a conformance concern? How about add a ```checkWrapperMessage``` which wrap ```checkMessage``` so that we only pay cost for wrapper messages?

php/tests/proto/test_wrapper_type_setters.proto
```diff
@@ -0,0 +1,19 @@
...
+  google.protobuf.FloatValue float_value = 2;
+  google.protobuf.Int64Value int64_value = 3;
+  google.protobuf.UInt64Value uint64_value = 4;
+  google.protobuf.Int32Value int32_value = 5;
+  google.protobuf.UInt32Value uint32_value = 6;
+  google.protobuf.BoolValue bool_value = 7;
+  google.protobuf.StringValue string_value = 8;
+  google.protobuf.BytesValue bytes_value = 9;
+
+  repeated google.protobuf.DoubleValue double_value_list = 10;
```
TeBoring: Test for map field?

php/tests/wrapper_type_setters_test.php
```diff
@@ -0,0 +1,180 @@
...
+            [TestWrapperSetters::class, "setDoubleValue", "abc"],
+            [TestWrapperSetters::class, "setDoubleValue", []],
+            [TestWrapperSetters::class, "setDoubleValue", new stdClass()],
+
+            [TestWrapperSetters::class, "setFloatValue", "abc"],
+            [TestWrapperSetters::class, "setFloatValue", []],
+            [TestWrapperSetters::class, "setFloatValue", new stdClass()],
+
+            [TestWrapperSetters::class, "setInt64Value", "abc"],
+//            [TestWrapperSetters::class, "setInt64Value", 5.5],
```
TeBoring: I think this is allowed in php API. Will convert to 5.

php/tests/wrapper_type_setters_test.php
```diff
@@ -0,0 +1,180 @@
...
+            [TestWrapperSetters::class, "setInt64Value", []],
+            [TestWrapperSetters::class, "setInt64Value", new stdClass()],
+
+            [TestWrapperSetters::class, "setUInt64Value", "abc"],
+//            [TestWrapperSetters::class, "setUInt64Value", 5.5],
+//            [TestWrapperSetters::class, "setUInt64Value", -7],
```
TeBoring: This is allowed.

php/tests/wrapper_type_setters_test.php
```diff
@@ -0,0 +1,180 @@
...
+            [TestWrapperSetters::class, "setUInt32Value", "abc"],
+//            [TestWrapperSetters::class, "setUInt32Value", 5.5],
+//            [TestWrapperSetters::class, "setUInt32Value", -7],
+            [TestWrapperSetters::class, "setUInt32Value", []],
+            [TestWrapperSetters::class, "setUInt32Value", new stdClass()],
+
+            [TestWrapperSetters::class, "setBoolValue", []],
+            [TestWrapperSetters::class, "setBoolValue", new stdClass()],
+
+//            [TestWrapperSetters::class, "setStringValue", 5],
```
TeBoring: I remember this was allowed. Convert to "5".

michaelbausor:
@TeBoring @fiboknacky PTAL
Regarding offline discussion of boxing/unboxing - I have put the boxing logic for setters into a method in a new class, GPBWrapperUtils. This would be my personal preference instead of generating this code inline in every setter, since it is not completely trivial like the unboxing logic. But let me know if you prefer to have this generated inline as well, and I can do that instead.

php/src/Google/Protobuf/Internal/GPBWrapperUtils.php
```diff
@@ -0,0 +1,73 @@
...
+            // existing type checking done by GPBUtil.
+            return;
+        } else {
+            // Try to instantiate $class and set the value
+            try {
+                $msg = new $class;
+                $msg->setValue($value);
+                $value = $msg;
+                return;
+            } catch (\Exception $exception) {
```
TeBoring: Do we need to rethrow?

TeBoring:
For setter, can we also create setXXXValue? So that the original setter doesn't need to be complicated. The new setter will also looks cleaner.

fiboknacky:
> @TeBoring @fiboknacky PTAL
> Regarding offline discussion of boxing/unboxing - I have put the boxing logic for setters into a method in a new class, GPBWrapperUtils. This would be my personal preference instead of generating this code inline in every setter, since it is not completely trivial like the unboxing logic. But let me know if you prefer to have this generated inline as well, and I can do that instead.

I'm not quite familiar with this part of the lib. Will it affect how we can use setXXXValue and getXXXValue?
If not, then SGTM.

michaelbausor:
> For setter, can we also create setXXXValue? So that the original setter doesn't need to be complicated. The new setter will also looks cleaner.

I am not necessarily opposed to a setXXXValue method, but on reflection I think there are some downsides we should consider: null handling, and constructor arguments.

For null handling, we have two options: either setXXXValue(null) can be an error, or setXXXValue(null) will be the same as calling setXXX(null).
I think having setXXXValue(null) be an error would be a really bad user experience, because then users have to know to switch back and forth between the setXXX method and setXXXValue method based on whether they have a null value or not.
The alternative of having setXXXValue(null) just call setXXX(null) is better, but might be confusing because it is subtly different from the case of a non-null value. For a non-null value, setXXXValue($x) will *construct a wrapper type and set its value to $x*, so users might assume that setXXXValue(null) will *construct a wrapper type and set its value to null*, but this is not true and in fact is impossible (which is why wrapper types exist in the first place).

For constructor arguments, we would like to allow this usage:
```php
// Old usage
$myMessage = new MyMessage(['double_value' => new DoubleValue(['value' => 1.1])]);
// New usage
$myMessage = new MyMessage(['double_value' => 1.1]);
```
With the current implementation allowing the setXXX method to accept both the wrapper message type and the primitive type, then passing primitive values into the constructor "just works" - I have added tests for this in the latest commit.

This is not an argument against adding the setXXXValue method, but if we don't support accepting the primitive value in the existing setXXX setter then we will also need to make other changes to support the usage in the constructor, e.g. in [Message::mergeFromArray](https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L989), and it is not clear how we could do that without extra complexity.

>I'm not quite familiar with this part of the lib. Will it affect how we can use setXXXValue and getXXXValue?
If not, then SGTM.

This should only affect the implementation, not how it is used.

TeBoring:
I got your point. I think having both setXXXValue and getXXXValue tells
more clearly what the API is doing.

TeBoring:
+1 for calling setXXX(null) in setXXXValue

TeBoring:
I think setXXX already report error when given null

michaelbausor:
If we do add a setXXXValue method, WDYT about whether or not to support accepting the primitive type in setXXX? If we don't, do you have a suggestion about how to implement the changes for the constructor argument? I haven't looked into what I think the best approach will be.

>I think setXXX already report error when given null

This is true for primitive types, but not for message types such as DoubleValue and other wrapper types, where null is a valid value.

TeBoring:
https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L990
I think we can add a try catch here. If an exception is thrown, we check whether the field to be set is a wrapper message. If so, we do the setXXXValue.

michaelbausor:
cc-ing @AnashOommen for comment

fiboknacky:
Is there any way to:
* Support setting primitive type in constructor argument `$myMessage = new MyMessage(['double_value' => 1.1]);`
* Add both setXXXValue and getXXXValue

I see your concerns about using null with setXXXValue now. IMHO, it would be better not to support using null with setXXXValue at all.
It'd be misleading if setXXX(null) is equal to setXXXValue(null).

As for supporting primitive value in setXXX(), I'm quite neutral to this with a bit inclination to *not support it*.
The reason is that it'd be clearer to separate the use of setXXX and setXXXValue.
(Needing to type one more word "Value" after the original method names doesn't seem bad to me)

If we sometimes allow some kinds of arguments in both methods, but in other times, disallow other kinds of arguments in one methods, it'd be pretty confusing.

php/tests/wrapper_type_setters_test.php
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public function testInvalidSetters($class, $setter, $value)
+    {
+        (new $class())->$setter($value);
+    }
+
+    public function invalidSettersDataProvider()
+    {
+        return [
+            [TestWrapperSetters::class, "setDoubleValue", "abc"],
```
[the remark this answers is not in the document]
michaelbausor: Yes, you're right - fixed.

php/tests/wrapper_type_setters_test.php
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * @dataProvider constructorWithWrapperTypeDataProvider
+     */
+    public function testConstructorWithWrapperType($class, $wrapperClass, $wrapperField, $value)
+    {
+        $actualInstance = new $class([$wrapperField => $value]);
+        $expectedInstance = new $class([$wrapperField => new $wrapperClass(['value' => $value])]);
+        $this->assertEquals($expectedInstance, $actualInstance);
```
TeBoring: Currently, assertEquals doesn't work for message in c extension. Basically, the assertEquals assert nothing in c extension. Could you actually test equal of value?

php/tests/wrapper_type_setters_test.php
```diff
@@ -0,0 +1,221 @@
...
+use Google\Protobuf\StringValue;
+use Google\Protobuf\UInt32Value;
+use Google\Protobuf\UInt64Value;
+
+class WrapperTypeSettersTest extends TestBase
+{
+    /**
+     * @dataProvider gettersAndSettersDataProvider
+     */
+    public function testGettersAndSetters($class, $wrapperClass, $setter, $valueSetter, $getter, $valueGetter, $sequence)
```
TeBoring: line is too long?

TeBoring:
I implemented c extension change in https://github.com/TeBoring/protobuf/tree/php-wrapper-setters. It's based on your branch. But I can't find your repo to send PR. Could you pull my change and merge it in your branch?

php/ext/google/protobuf/message.c
```diff
@@ -1024,7 +1035,15 @@ static void hex_to_binary(const char* hex, char** binary, int* binary_len) {
 PHP_METHOD(Any, __construct) {
   init_file_any(TSRMLS_C);
   MessageHeader* intern = UNBOX(MessageHeader, getThis());
-  custom_data_init(any_type, intern PHP_PROTO_TSRMLS_CC);
+
```
michaelbausor: @TeBoring All of these constructors are the same. Could they use a common function instead of repeating the same code?

michaelbausor:
I have pulled in the changes and made fixes, PTAL.

michaelbausor:
(Fixes in my code based on PR comments, not fixes in the C extension changes)

php/ext/google/protobuf/message.c
TeBoring: Updated in my branch. Please sync again.

michaelbausor:
Synced, PTAL

TeBoring:
Sorry, I forgot to add include path in tests.sh. I have send the fix. Please sync again.

michaelbausor:
Synced, PTAL

TeBoring:
I am really sorry that there were other problems. This time I have tested in https://github.com/protocolbuffers/protobuf/pull/5222. Could you please sync again?

michaelbausor:
No worries! Synced, PTAL.

TeBoring:
The broken javascript test should not be related to this change.

TeBoring:
@BSBandme could you help take a look of the broken javascript test?

Issue #5141, opened 2018-09-13
Protobuf fails to configure with SunCC


noloader:
**What version of protobuf and what language are you using?**

Version: Master

Language: C++

**What operating system and version?**

Solaris 11.3 x86

**What runtime / compiler are you using (e.g., python version or gcc version)**

SunCC 12.4 and above [support C++11](https://docs.oracle.com/cd/E37069_01/html/E37071/gndfg.html). SunCC 12.4 can be a bit C++11 anemic but 12.5 should have everything needed for Protobuf.

```
$ /opt/developerstudio12.5/bin/CC -V
CC: Studio 12.5 Sun C++ 5.14 SunOS_i386 2016/05/31
```

**Steps to reproduce the behavior:**
1. git clone ...
2. ./autogen
3. CC=/opt/developerstudio12.5/bin/cc CXX=/opt/developerstudio12.5/bin/CC ./configure

**Result**

```
$ cat config.log
...

configure:18987: WARNING: could not find an STL hash_map
configure:18997: checking for library containing sched_yield
configure:19028: /opt/developerstudio12.5/bin/CC -o conftest -g0 -xO3 -xlibmil -xdepend -xbuiltin -mt -compat=5 -library=stlport4 -library=Crun -template=no%extdef  -std=c++11 -DNDEBUG -m64   conftest.cpp  >&5
CC: -compat=5 can not be used with -std
CC: -library=Crun cannot be used with -std=c++11. To use this library you need to switch to -std=sun03
CC: -library=stlport4 cannot be used with -std=c++11. To use this library you need to switch to -std=sun03
configure:19028: $? = 1
```

**Anything else we should know about your project / environment**

Mostly a standard Solaris environment using Bash as the shell.

This Solaris machine is available for remote access. I need a SSH authorized_keys file. Email to *noloader, gmail* account.

[config.log.zip](https://github.com/protocolbuffers/protobuf/files/2377410/config.log.zip)

------

Note to self... After fixing the M4 macro Protobuf [crashes the compiler](https://www.youtube.com/watch?v=DFloZbT99oE) (the Assertion below).  SunCC with Sun Studio 12.5 crashes. Sun Studio 12.6 is OK and configure completes.

```
configure:19397: /opt/developerstudio12.5/bin/CC -c -g0 -xO3 -xlibmil -xdepend -xbuiltin -mt -template=no%extdef  -std=c++11 -DNDEBUG -m64  conftest.cpp >&5
[...]

PR #5142, opened 2018-09-13
Mostly fix Sun Studio configuration using SunCC compiler (GH #5141)

noloader:
Sun Studio 12.4 is required because it provides the first C++11 compiler from Sun. However, SunCC in 12.4 fails to configure because it is so buggy. The SunCC compiler in Sun Studio 12.5 crashes on the m4 macro test test_type_deduction (line 257). The m4 test can probably be reworked so Sun Studio 12.5 is supported.

The Sun Studio 12.6 compiler finishes configuring. At the moment it looks like Sun Studio 12.6 is the only one supported.


xfxyjwf:
@haberman is working on deprecating autotools support. Can you try cmake instead and update the PR to update the cmake files instead?

haberman:
Yes I concur with @xfxyjwf, is it possible to fix this in CMake instead?

haberman:
I'll merge this for now, but if/when we migrate to CMake this fix will be lost.

PR #5146, opened 2018-09-13
Provide simple access to descriptor declarations in C#

jskeet:
This is primarily for access to comments, which would be expected to be available in a protoc plugin.

The implementation has two fiddly aspects:

- We use a Lazy<T> to avoid building the map before cross-linking. An alternative would be to crosslink at the end of the constructor, and remove the calls to CrossLink elsewhere. This would be generally better IMO, but deviate from the Java code.
- The casts to IReadOnlyList<DescriptorBase> are unfortunate. They'll always work, because these lists are always ReadOnlyCollection<T> for a descriptor type... but we can't use IList<DescriptorBase> as that's not covariant, and it's annoyingly fiddly to change the field to be of type ReadOnlyCollection<T>.


csharp/src/Google.Protobuf.Test/Reflection/DescriptorDeclarationTest.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Reflection;
+
+namespace Google.Protobuf.Test.Reflection
+{
+    // In reality this isn't a test for DescriptorDeclaration so much as the way they're loaded.
+    public class DescriptorDeclarationTest
+    {
+        static readonly FileDescriptor unitTestProto3Descriptor = LoadProtos();
+
+        [Test]
```
anandolee: Test FileDesciptor comments as well?

jskeet: Ah, I didn't think FileDescriptors could have comments themselves. Will add that support in another commit, and let you know when I've done so.

anandolee: Might  unable to get the leading comments which are in "an unstable location" mentioned in this issue:
https://buganizer.corp.google.com/issues/18735548

We may only able to add comments for FileDescriptor's package, syntax etc...

jskeet: Given the comments on that bug, I'd suggest leaving `FileDescriptor` without a `Declaration` property at all for the moment. It will be easy enough to add later if we want to. As such, I think the PR should be okay in its current form. Let me know if you disagree though - I'd be happy to be convinced :)

anandolee: Agree, just add a comments for the FileDescriptor issue.

jskeet: I've added two commits - one with a comment for FileDescriptor, and one tidying up the other descriptor-based tests.

csharp/protos/unittest_proto3.proto
```diff
@@ -378,3 +380,31 @@ message BarRequest  {}
...
+  }
+  
+  // Leading nested enum comment
+  enum NestedCommentEnum {
+    // Zero value comment
+    ZERO_VALUE = 0;
+  }
+  
+  // Leading field comment
+  string text = 1; // Trailing field comment
```
anandolee: More trailing comments tests for MessageDescriptor EnumValueDescriptor etc.

anandolee:
Thanks for the PR! It looks good, just need more tests.
Please also fix the MacOS C++ Distcheck for the newly added files

jskeet:
Not sure why the tests have started failing - will look into that.

jskeet:
Looks like I'd run an old, locally-built version of protoc when I regenerated the test protos. Fixed and repushed...

PR #5150, opened 2018-09-14
Upgrade Guava to 26.0

kluever:


xfxyjwf:
@ejona86

java/pom.xml
```diff
@@ -82,7 +82,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>20.0</version>
+        <version>26.0</version>
```
ejona86: As I mentioned in the internal review, this version does not exist. The versions are`26.0-jre` and `26.0-android`. Since protobuf supports Java 7, it needs to use `26.0-android`.

elharo: Many things that depend on protobuf are on 20.0. This is likely to introduce convergency issues downstream. I wish we could do a multirepo PR like we can in google3.

haon4: I created https://github.com/protocolbuffers/protobuf/pull/5630 with 26.0-android instead. Please take a look.

BSBandme:
In #5630 we have updated to 26.0-android, will close this PR.

PR #5156, opened 2018-09-15
[C#] Make it possible to run tests under .NET 3.5

JohnHBrock:
For those who want to create custom builds of Google.Protobuf.dll targeting .NET 3.5, this change makes it possible to also run the tests under 3.5.

`dotnet test` is unable to run tests under 3.5, so I've changed Google.Protobuf.Test.csproj to produce an EXE instead of a DLL. One can run the tests simply by running Google.Protobuf.Test.exe.

This is undoing some of the changes made in commit e823897005b88b03088fe17d219ffaeeabb2961d. Before that commit, Google.Protobuf.Test.csproj was producing an EXE instead of a DLL.


JohnHBrock:
Closing this, just discovered some possible issues with `dotnet test` after this change...

PR #5157, opened 2018-09-15
[C#] Target .NET 4.5.2 instead of 4.5 or 4.5.1

JohnHBrock:
As of January 12, 2016, Microsoft no longer supports .NET 4.5 or .NET 4.5.1, so let's use 4.5.2 instead.

### DIFF
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -39,7 +39,7 @@ dotnet build -c %configuration% || goto error
 
 echo Testing C#
 dotnet test -c %configuration% -f netcoreapp1.0 Google.Protobuf.Test\Google.Protobuf.Test.csproj || goto error
-dotnet test -c %configuration% -f net451 Google.Protobuf.Test\Google.Protobuf.Test.csproj || goto error
+dotnet test -c %configuration% -f net452 Google.Protobuf.Test\Google.Protobuf.Test.csproj || goto error
 
 goto :EOF
 
--- a/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
+++ b/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net451;netcoreapp1.0</TargetFrameworks>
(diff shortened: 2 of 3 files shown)


jskeet:
I don't support this change: it would be a breaking change for anyone who *is* still using .NET 4.5.
We could emphasize that it's not a Microsoft-supported platform, but I don't like the idea of just starting to break people's builds. Continuing to target .NET 4.5 doesn't actually hurt us.

JohnHBrock:
I can live with leaving the library itself targeting 4.5, but the presence of unpatched security vulnerabilities in 4.5 makes me nervous. [Here's a list of vulnerabilities](https://www.cvedetails.com/vulnerability-list.php?vendor_id=26&product_id=2002&version_id=171205&page=1&hasexp=0&opdos=0&opec=0&opov=0&opcsrf=0&opgpriv=0&opsqli=0&opxss=0&opdirt=0&opmemc=0&ophttprs=0&opbyp=0&opfileinc=0&opginf=0&cvssscoremin=0&cvssscoremax=0&year=0&month=0&cweid=0&order=1&trc=36&sha=c34682d207dbd27b5e460260f4d9741a3a22e12d) that have been found in 4.5.2; I'm worried that a lot of the vulnerabilities found since January 2016 are also in .NET 4.5 and have gone unpatched.

At the very least, what about leaving the `net45` target for now, but also adding `net452`? We could also include a deprecation warning in the library if protobuf is a dependency, e.g.:
```csharp
#if NET45
#warning protobuf has deprecated support for .NET <=4.5.1, please target .NET 4.5.2 or higher
#endif
```

jskeet:
I see no need for this package to try to make a decision on behalf of its users. Sure, it would be great if all users kept up with security fixes and latest versions etc, but I don't think it's appropriate to use NuGet packages to force that.

(As an example of another very popular package, Newtonsoft.Json still includes a target for .NET 2.0.)

JohnHBrock:
@jskeet How do you feel about keeping the `net45` target, but adding a `net452` target?
And how would you feel about adding a `net35` target?

jskeet:
I see no benefit in adding net452. I'd like to add net35 at some point (as the code now works there) but there's some build work to do in order to get that to work.

JohnHBrock:
Closing this PR based on your feedback. I'll submit another PR that takes a pass at explicitly supporting .NET 3.5.

PR #5158, opened 2018-09-16
Add Elixir protobuf implementation

m13m:


xfxyjwf:
Hi @m13m , this README.md file lists Google official implementations only. For third-party ones, you can add it in this third-party page:
https://github.com/protocolbuffers/protobuf/blob/master/docs/third_party.md

Issue #5165, opened 2018-09-18
Python ParseFromString does not return length


joshkel:
**What version of protobuf and what language are you using?**
Version: 3.5.2post1
Language: Python

**What operating system (Linux, Windows, ...) and version?**

Ubuntu 16.04 32-bit

**What runtime / compiler are you using (e.g., python version or gcc version)**

Python 3.5.2

**What did you do?**

The Python API docs say that `ParseFromString` returns the number of bytes parsed, and the C++ implementation of `google.protobuf.message.Message.ParseFromString` returns the length, but the Python implementation does not.

([The protobuf wheel for Ubuntu 16.04 32-bit](https://files.pythonhosted.org/packages/bf/dc/42d35afffe88ce301c90457eb60d145dd64144f283ef4bf927c1187ef114/protobuf-3.5.2.post1-py2.py3-none-any.whl) is built without the C++ extension, so it's easy to encounter this issue on this platform.)

**What did you expect to see**

`message.ParseFromString` returns length.

**What did you see instead?**

`message.ParseFromString` returns None.

PR #5166, opened 2018-09-18
ParseFromString returns bytes parsed

joshkel:
ParseFromString is documented as returning the number of bytes parsed,
and the C++ implementation does this, so the Python implementation
should too.

See #5165.

### DIFF
--- a/python/google/protobuf/message.py
+++ b/python/google/protobuf/message.py
@@ -182,7 +182,7 @@ class Message(object):
     do not return the value that MergeFromString returns.
     """
     self.Clear()
-    self.MergeFromString(serialized)
+    return self.MergeFromString(serialized)
 
   def SerializeToString(self, **kwargs):
     """Serializes the protocol message to a binary string.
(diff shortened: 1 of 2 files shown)


python/google/protobuf/message.py
```diff
@@ -182,7 +182,7 @@ def ParseFromString(self, serialized):
     do not return the value that MergeFromString returns.
```
Code0x58: This docstring is now out of date

joshkel: Fixed. Thanks.

I see now that the Python API documentation at https://developers.google.com/protocol-buffers/docs/reference/python/google.protobuf.message.Message-class is out of date (from an older version of protobuf that didn't have this text), although I don't see anywhere to report or suggest updates to the documentation site.

anandolee:
Can you point me where is the document that said ParseFromString returns the number of bytes parsed?

Also need to add test for the change and make sure pure python have the same behavior with python cpp extension.

anandolee:
Your change is good to go and thanks for the fix.
Just note:  Unexpected end-group tag should be an error. We always want to change c extension to raise error instead of warning but it was blocked by some google internal projects.

You could see a TODO here:
https://github.com/protocolbuffers/protobuf/blob/f9d8138376765d229a32635c9209061e4e4aed8c/python/google/protobuf/pyext/message.cc#L2155

Issue #3965, opened 2017-11-28
[python] well known types lacks of rich comparison


cuonglm:
Take a look at [well_known_types.py](https://github.com/google/protobuf/blob/master/python/google/protobuf/internal/well_known_types.py), most of them lacks of rich comparison method. So writing it the idiom way in Python can cause confusion. Example:

```
In [1]: from google.protobuf import timestamp_pb2

In [2]: bool(timestamp_pb2.Timestamp())
Out[2]: True
```

always give `True`, causing this code:
```
t = timestamp_pb2.Timestamp()
if not t:
    print 't is empty message'
```

does not print as expected.

While:

```
In [1]: from google.protobuf import struct_pb2

In [2]: bool(struct_pb2.Struct())
Out[2]: False
```

always give `False` (`Struct` has its own [\_\_len\_\_\(\)](https://github.com/google/protobuf/blob/master/python/google/protobuf/internal/well_known_types.py#L748)).

And also:

```
In [1]: from google.protobuf import timestamp_pb2

In [2]: t = timestamp_pb2.Timestamp()

In [3]: now = timestamp_pb2.Timestamp()

In [4]: now.GetCurrentTime()

In [5]: s = timestamp_pb2.Timestamp()

In [6]: now > t
Out[6]: False

In [7]: now > s
Out[7]: True
```

is not consistent, because `Timestamp` lacks all of rich comparison methods.

anandolee:
Close it as the related PR #3965 is closed. Feel free to reopen if it is still needed

PR #5168, opened 2018-09-19
#3965 well known rich comparisons

Code0x58:
For #3965
- [x] add comparisons to Timestamp
  - [x] Timestamp
  - [x] ~~datetime~~ requires `timetuple` attribute for python2. Didn't add to avoid surprises
  - [x] test
- [x] add comparisons to Duration
  - [x] Duration
  - [x] ~~timedelta~~ not doable in python2 (without a big mess?)
  - [x] test
- [x] ~~add comparisons to FieldMask~~ moved to a separate [branch](https://github.com/Code0x58/protobuf/tree/fieldmask-total-ordering)

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -254,6 +254,31 @@ class Timestamp(object):
     self.seconds = calendar.timegm(dt.utctimetuple())
     self.nanos = dt.microsecond * _NANOS_PER_MICROSECOND
 
+  def _comparable_vector(self, obj):
+    # not dealing with datetime as didn't want to cause any surprises with 2/3
+    # differences. see https://github.com/protocolbuffers/protobuf/pull/5168
+    return (obj.seconds, obj.nanos)
+
+  # Make sure to normalize timestamps before using the following comparisons.
+  # Comparing non-normalized timestamps is not specified and may give unexpected results.
+  def __eq__(self, other):
+    return self._comparable_vector(self) == self._comparable_vector(other)
+
+  def __ne__(self, other):
+    return self._comparable_vector(self) != self._comparable_vector(other)
+
+  def __lt__(self, other):
+    return self._comparable_vector(self) < self._comparable_vector(other)
+
+  def __le__(self, other):
+    return self._comparable_vector(self) <= self._comparable_vector(other)
+
+  def __gt__(self, other):
+    return self._comparable_vector(self) > self._comparable_vector(other)
+
+  def __ge__(self, other):
+    return self._comparable_vector(self) >= self._comparable_vector(other)
+
 
 class Duration(object):
   """Class for Duration message type."""
@@ -383,6 +408,30 @@ class Duration(object):
       nanos -= _NANOS_PER_SECOND
     self.seconds = seconds
     self.nanos = nanos
+
+  def _comparable_vector(self, obj):
+    # comarison with timedelta not possible with python2 - see https://github.com/protocolbuffers/protobuf/pull/5168
+    return (obj.seconds, obj.nanos)
+
+  # Make sure to normalize timestamps before using the following comparisons.
+  # Comparing non-normalized timestamps is not specified and may give unexpected results.
+  def __eq__(self, other):
+    return self._comparable_vector(self) == self._comparable_vector(other)
+
+  def __ne__(self, other):
+    return self._comparable_vector(self) != self._comparable_vector(other)
+
+  def __lt__(self, other):
+    return self._comparable_vector(self) < self._comparable_vector(other)
+
+  def __le__(self, other):
+    return self._comparable_vector(self) <= self._comparable_vector(other)
+
+  def __gt__(self, other):
+    return self._comparable_vector(self) > self._comparable_vector(other)
+
+  def __ge__(self, other):
+    return self._comparable_vector(self) >= self._comparable_vector(other)
 
 
 def _CheckDurationValid(seconds, nanos):
--- a/python/google/protobuf/internal/well_known_types_test.py
+++ b/python/google/protobuf/internal/well_known_types_test.py
@@ -34,6 +34,7 @@
 
 __author__ = 'jieluo@google.com (Jie Luo)'
 
+import copy
 import datetime
 
 try:
@@ -60,6 +61,28 @@ from google.protobuf.internal import test_util
 from google.protobuf.internal import well_known_types
 from google.protobuf import descriptor
 from google.protobuf import text_format
+
+
+def total_ordering_test(self, lesser, greater):
+    self.assertTrue(lesser == copy.deepcopy(lesser))
+    self.assertTrue(not lesser == greater)
+
+    self.assertTrue(lesser != greater)
+    self.assertTrue(not lesser != lesser)
+
+    self.assertTrue(greater > lesser)
+    self.assertTrue(not lesser > lesser)
+
+    self.assertTrue(greater >= lesser)
+    self.assertTrue(lesser >= lesser)
+    self.assertTrue(not lesser >= greater)
+
+    self.assertTrue(lesser < greater)
+    self.assertTrue(not greater < lesser)
+
+    self.assertTrue(lesser <= greater)
+    self.assertTrue(lesser <= lesser)
+    self.assertTrue(not greater <= lesser)
 
 
 class TimeUtilTestBase(unittest.TestCase):
@@ -386,6 +409,24 @@ class TimeUtilTest(TimeUtilTestBase):
         ValueError,
         r'Duration is not valid\: Sign mismatch.',
         message.ToJsonString)
+
+  def testTimestampOrdering(self):
+    a, b = timestamp_pb2.Timestamp(), timestamp_pb2.Timestamp()
+    # normalized comparisons
+    a.seconds, a.nanos, b.seconds, b.nanos = 0, 1, 1, 0
+    total_ordering_test(self, a, b)
+    # non-normalised is not equal to normalised
+    a.seconds, a.nanos, b.seconds, b.nanos = 1, 0, 0, 1000000001
+    self.assertNotEqual(a, b)
+
+  def testDurationOrdering(self):
+    a, b = duration_pb2.Duration(), duration_pb2.Duration()
+    # normalized comparisons
+    a.seconds, a.nanos, b.seconds, b.nanos = 0, 1, 1, 0
+    total_ordering_test(self, a, b)
+    # non-normalised is not equal to normalised`
+    a.seconds, a.nanos, b.seconds, b.nanos = 1, 0, 0, 1000000001
+    self.assertNotEqual(a, b)
 
 
 class FieldMaskTest(unittest.TestCase):


Code0x58:
I signed it!

python/google/protobuf/internal/well_known_types_test.py
```diff
@@ -372,6 +372,56 @@ def testInvalidDuration(self):
         r'Duration is not valid\: Sign mismatch.',
         message.ToJsonString)
 
+  def testTimestampOrdering(self):
+    a, b = timestamp_pb2.Timestamp(), timestamp_pb2.Timestamp()
+    a.seconds, a.micros = 0, 1
+    b.seconds, b.micros = 1, 0
```
anandolee: Add test for Timestamp that is not normalized and add comments to indicate the result may not correct if Timestamp is not normalized.

anandolee:
Some tests will not be triggered unless protobuf-team has add a kokoro:run tag. Just added the tag

python/google/protobuf/internal/well_known_types_test.py
Code0x58: I replaced `_comparable_vector(…)` with `_comparable_scalar(…)` which returns `ToNanoseconds()` for non-datetimes so will work with non-normalized values, and added tests

Code0x58:
I guess FieldMask comparisons wouldn't be very common (I don't have a practical use case for these at the moment), but thought it was interesting so had a go. I was thinking of them as a sets where, where the fields matched by _x,a.b_ is a proper subset of _x,a_.

I can pull out the FieldMask comparisons for a separate PR so they can get more scrutiny

python/google/protobuf/internal/well_known_types_test.py
```diff
@@ -372,6 +372,54 @@ def testInvalidDuration(self):
         r'Duration is not valid\: Sign mismatch.',
         message.ToJsonString)
 
+  def testTimestampOrdering(self):
+    a, b = timestamp_pb2.Timestamp(), timestamp_pb2.Timestamp()
+    # normalized + non-normalized comparisons
+    for a.seconds, a.nanos, b.seconds, b.nanos in ((0, 1, 1, 0), (1, 0, 0, 1000000001)):
+        self.assertTrue(a == a)
```
Code0x58: These assertions could be factored out into a method that takes (a, b), then the two test methods can be DRY and take little extra work to add (datetime, Timestamp) comparison tests.

It could also have `a == deepcopy(a)`.

python/google/protobuf/internal/well_known_types.py
```diff
@@ -244,6 +244,34 @@ def FromDatetime(self, dt):
     self.seconds = calendar.timegm(dt.utctimetuple())
     self.nanos = dt.microsecond * _NANOS_PER_MICROSECOND
 
+  def _comparable_scalar(self, obj):
+    if isinstance(obj, datetime):
+      return calendar.timegm(obj.utctimetuple()) * _NANOS_PER_SECOND + obj.microsecond * _NANOS_PER_MICROSECOND
+    return obj.ToNanoseconds()
+
+  def timetuple(self):
+    """Only exists to allow comparisons to datetimes in Python 2.7"""
```
Code0x58: [[docs](https://docs.python.org/2/library/datetime.html#datetime.datetime.tzinfo)]
> **Note:** In order to stop comparison from falling back to the default scheme of comparing object addresses, datetime comparison normally raises TypeError if the other comparand isn’t also a datetime object. However, NotImplemented is returned instead if the other comparand has a timetuple() attribute. This hook gives other kinds of date objects a chance at implementing mixed-type comparison. If not, when a datetime object is compared to an object of a different type, TypeError is raised unless the comparison is == or !=. The latter cases return False or True, respectively.

Code0x58: Humm, it looks like this will only go down to seconds https://docs.python.org/3/library/time.html#time.struct_time - it may be better to drop this and say it isn't available in Python2

python/google/protobuf/internal/well_known_types.py
```diff
@@ -244,6 +244,33 @@ def FromDatetime(self, dt):
...
     self.nanos = dt.microsecond * _NANOS_PER_MICROSECOND
 
+  def _comparable_scalar(self, obj):
+    if isinstance(obj, datetime):
+      return calendar.timegm(obj.utctimetuple()) * _NANOS_PER_SECOND + obj.microsecond * _NANOS_PER_MICROSECOND
+    return obj.ToNanoseconds()
+
+  # Just so datetime's can be compared with as they'll raise NotImplemented and
+  # allow this class to do rich comparisons
+  timetuple = None
```
Code0x58: From reading py27's datetime source: https://github.com/python/cpython/blob/2.7/Modules/datetimemodule.c#L4268

python/google/protobuf/internal/well_known_types.py
```diff
@@ -372,6 +399,34 @@ def _NormalizeDuration(self, seconds, nanos):
     self.seconds = seconds
     self.nanos = nanos
 
+  def _comparable_scalar(self, obj):
+    if isinstance(obj, timedelta):
+      return (obj.seconds + obj.days * _SECONDS_PER_DAY) * _NANOS_PER_SECOND + obj.microseconds * _NANOS_PER_MICROSECOND
+    return obj.ToNanoseconds()
+
+  def __eq__(self, other):
```
Code0x58: These rich comparisons [just won't work in Python2](https://github.com/python/cpython/blob/2.7/Modules/datetimemodule.c#L1757).

python/google/protobuf/internal/well_known_types.py
```diff
@@ -372,6 +399,34 @@ def _NormalizeDuration(self, seconds, nanos):
     self.seconds = seconds
     self.nanos = nanos
 
+  def _comparable_scalar(self, obj):
+    if isinstance(obj, timedelta):
+      return (obj.seconds + obj.days * _SECONDS_PER_DAY) * _NANOS_PER_SECOND + obj.microseconds * _NANOS_PER_MICROSECOND
```
Code0x58: An issue with both this, and the comparison for datetimes (along with switching between types) is that their resolution is different (builtin milliseconds vs. protobuf nanoseconds). This is probably something to make sure that is documented, mostly for the comparisons.

python/google/protobuf/internal/well_known_types.py
anandolee: Isn't the comparisions implemented by yourself, why related to python2/python3? If there's difference for timedelta, you could check the sys.version_info to do different code for python2/python3"
https://github.com/protocolbuffers/protobuf/blob/fd4baa91e7c127a29ee6d5f6ecf0fa149f6d9cec/python/compatibility_tests/v2.5.0/setup.py#L10

Code0x58: Unfortunately the C code for python2 has some [oddities](https://github.com/protocolbuffers/protobuf/pull/5168#discussion_r220855631) that stop comparisons from simply working. The fix for `datetime` is simple and not too messy, but for [`timedelta`](https://github.com/python/cpython/blob/2.7/Modules/datetimemodule.c#L1757) I think they can only ever compare to objects that are `timedelta` instances, so the only way may be to subclass from `timedelta`.

anandolee: In that case, maybe we only provide Timestamp compare with Timestamp not datetime, Duration only compare with Duration not timedelta. Users should call ToDatetime/FromDatetime/ToTimedelta/FromTimedelta first if them want to compare with datetime/timedelta

Code0x58: It turns out that either `Duration`/`timedelta` comparisons either aren't possible in python2 (tried dropping in `timdelta` as `Duration`'s parent class), or would take involved changes and still leave a mess in the code. I doubt anyone likes the idea of introducing introducing python 2-3 incompatibilities, so that is gone.

`Timestamp`/`datetime` is something I was erring on on, as I think it would be nice to have, and the cost of it is 1 junk attribute, but then it'd be a bit surprising to users that built in type comparisons aren't all implemented, so seems like a fair reason to avoid that junk.

I updated the PR to drop the built in type comparisons, they would be good to add when python2 support is dropped though - is there an issue to drop 2 (even if support will go on for years)?

python/google/protobuf/internal/well_known_types.py
```diff
@@ -244,6 +244,29 @@ def FromDatetime(self, dt):
     self.seconds = calendar.timegm(dt.utctimetuple())
     self.nanos = dt.microsecond * _NANOS_PER_MICROSECOND
 
+  def _comparable_scalar(self, obj):
+    # not dealing with datetime as didn't want to cause any surprises with 2/3
+    # differences. see https://github.com/protocolbuffers/protobuf/pull/5168
+    return obj.ToNanoseconds()
```
anandolee: It comes to a question "do we want to do normalize before compare" . It was discussed in #4318:

Message has implemented an equal compare which compare if all fields are the same, thus there are two equals  "time equality"  and  "protocol buffer equality" for Timestamps. If we do normalize on time equality, the two will not match.  We think it make sense to not normalize for all comparison. Add comments that if the timestamp is not normalized the result may not correct in the API documentation

Code0x58: Thanks for the detail @anandolee! Where about would I make changes to get into the API documentation?

_p.s. I should probably update the documentation mention the change of #5234_

anandolee: Add comments in the well_known_types.py should be enough.
Before " def __eq__(self, other):" add something like "# Make sure to normalized timestamps before using the following comparisons. Comparing non-normalized timestamps is not specified and may give unexpected results."

C#  example:
https://github.com/protocolbuffers/protobuf/blob/52fee0e8f03a56931a38e63c2b2ac2ba91191742/csharp/src/Google.Protobuf/WellKnownTypes/TimestampPartial.cs#L229

Code0x58: Thanks, I just amended with the comment

python/google/protobuf/internal/well_known_types_test.py
```diff
@@ -372,6 +395,18 @@ def testInvalidDuration(self):
         r'Duration is not valid\: Sign mismatch.',
         message.ToJsonString)
 
+  def testTimestampOrdering(self):
+    a, b = timestamp_pb2.Timestamp(), timestamp_pb2.Timestamp()
+    # normalized + non-normalized comparisons
+    for a.seconds, a.nanos, b.seconds, b.nanos in ((0, 1, 1, 0), (1, 0, 0, 1000000001)):
```
Code0x58: **TODO:** assert that `norm(denorm) != denorm` for Timestamp and Duration

anandolee:
The change Looks good to me. Can you fix the conflict and then I will run tests.

python/google/protobuf/internal/well_known_types.py
```diff
@@ -244,6 +244,31 @@ def FromDatetime(self, dt):
     self.seconds = calendar.timegm(dt.utctimetuple())
     self.nanos = dt.microsecond * _NANOS_PER_MICROSECOND
 
+  def _comparable_vector(self, obj):
+    # not dealing with datetime as didn't want to cause any surprises with 2/3
+    # differences. see https://github.com/protocolbuffers/protobuf/pull/5168
+    return (obj.seconds, obj.nanos)
+
+  # Make sure to normalized timestamps before using the following comparisons.
```
amauryfa: typo: "to normalize"

python/google/protobuf/internal/well_known_types.py
```diff
@@ -372,6 +397,35 @@ def _NormalizeDuration(self, seconds, nanos):
...
+  def __le__(self, other):
+    return self._comparable_vector(self) <= self._comparable_vector(other)
+
+  def __gt__(self, other):
+    return self._comparable_vector(self) > self._comparable_vector(other)
+
+  def __ge__(self, other):
+    return self._comparable_vector(self) >= self._comparable_vector(other)
+
+  def __bool__(self):
```
amauryfa: This is not really a comparison, and the first time that a Message has a boolean value. Can't users just call `ToTimedelta()`?

Code0x58: Good point, I just amended this out

anandolee:
Looks like the tests can not pass with python c extension

Code0x58:
Thanks @anandolee! I just rebased onto master and the tests at least work for me locally

TeBoring:
Feel free to reopen if this is still needed (btw some python tests failed).

Issue #4794, opened 2018-06-19
Allow filenames in _CalledFromGeneratedFile to contain "/pb2/"


thisch:
**What language does this apply to?**
python (c++ implementation)

**Describe the problem you are trying to solve.**

The company I work with strips the _pb2 suffix from the filenames of the generated python modules. Instead a pb2 python subpackage (e.g. ``companyname.pb2``) is created, which denotes that the modules in this pb2 subpackage are generated by protoc (and postprocessed).

If the c++ implementation of the protobuf bindings is used, importing those generated modules does not work:

```
    652 
    653     def __new__(cls, name, index, number, type=None, options=None):
--> 654       _message.Message._CheckCalledFromGeneratedFile()
    655       # There is no way we can build a complete EnumValueDescriptor with the
    656       # given parameters (the name of the Enum is not known, for example).

TypeError: Descriptors should not be created directly, but only retrieved from their parent.
```

**Describe the solution you'd like**
Given that a similar bug/feature request was recently fixed (#4262), I would like that _CalledFromGeneratedFile returns true when the filename contains the string "/pb2/".

**Describe alternatives you've considered**

**Additional context**

thisch:
Yes, I know. It was me who created this `companyname.pb2.xxx` package hierarchy format for our compiled proto-files years ago.

I created this convention because I think that this hierarchy is more pythonic than having modules suffixed with _pb2.

thisch:
There is a comment in https://github.com/google/protobuf/blob/HEAD/python/google/protobuf/pyext/descriptor.cc#L94

suggesting to completely remove the (hacky) `_CalledFromGeneratedFile` function. Are there any updates on this?

thisch:
@anandolee what do you think about this feature request? Does it make sense if I come up with a PR?

anandolee:
I agree to remove this function, it is a behavior diffs between pure python and c extension as well. 

+ @amauryfa on the TODO for removal, do we still have something to be worried for the removal?

thisch:
FYI: I've started to work on the removal of _CalledFromGeneratedFile in https://github.com/thisch/protobuf/commit/e0ae30255f1292d8a7e100d012224e603654d671

anandolee:
Those tests are for test coverage and also mark the behavior diffs of pure python and cpp extension. Could remove the failure tests or change to new behavior.

anandolee:
[...]

PR #5171, opened 2018-09-19
WIP: Remove _CheckCalledFromGeneratedFile

thisch:
and related functions.

Closes: #4794


thisch:
Does anyone know why the setter functions that are called when attributes in the _pb2 files are set, only return the return value of CheckCalledFromGeneratedFile() and not change the pyobject (see e.g. https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/pyext/descriptor.cc#L572)? 

The reason why I'm asking is that removing those setter functions causes problems in the generated _pb2.py files:

```
  File "/Users/travis/build/thisch/protobuf/python/google/protobuf/descriptor_pb2.py", line 1752, in <module>
    _DESCRIPTORPROTO_EXTENSIONRANGE.containing_type = _DESCRIPTORPROTO
```

anandolee:
Because you have removed the (setter)SetContainingType. Can just return 0 in SetContainingType()

thisch:
Maybe it makes sense that protoc generates code, which checks if the cpp-"backend" is used or not (using .e.g. `google.protobuf.descriptor._USE_C_DESCRIPTORS`).

thisch:
I see two options for moving forward with this PR.

1) Make the protoc generated python code compatible with backend agnostic.
2) Introduce a new protoc plugin that outputs python code that can only be used with the cpp backend.

 (Note that I would prefer if somebody else that is more familiar with the protoc codebase could do this)

WDYT?

anandolee:
In my other work of python descriptor pool, I am also thinking of adding checks if the cpp backend is used. However for compatibility issue it is still unable to remove those APIs: old generated code should work with new runtime. 

I may add the checks to improve cpp performance and make it possible to remove those APIs in future(might a few years later). But for now, we definitely can't break compatibility

thisch:
Do you really think that we have to support the code generated by an older version of protoc? If yes, we can aim for changing the output format with the next major version of protobuf. WDYT?

What are the plans for the cpp backend? Will it replace the python backend in the future?

> I may add the checks to improve cpp performance

I don't understand what you mean by checks here. How can checks improve the performance?

anandolee:
Will close it for clean up. Feel free to reopen if still need

PR #5172, opened 2018-09-19
Smear comments

### LINKED ISSUE #1890: Leap second handling is different inside and outside Google.
`timestamp.proto` refers to leap second smearing, but this generally only happens inside Google. With APIs inside and outside Google potentially needing to communicate sub-second timestamps, we should have a conversation about how smearing and non-smearing systems should interoperate.

shields:
Update comments to discuss leap smearing in detail.

Fixes #1890.


acozzette:
@shields Thanks for the pull request!

PR #5181, opened 2018-09-23
DynamicMessage.newBuilder support set initialCapacity.

wenshao:
If the number of Fields is greater than 16, specifying initialCapacity can improve performance.


BSBandme:
Thanks for pull request! Would you please first resolve the CLA issues? Like after signing CLA just reply here "I sign it!"

dlj-NaN:
This PR looks quite stale. Without a CLA, we can't consider merging.

@wenshao , sorry for the short reply. However, if you do want to re-open this PR, please make sure to follow googlebot's CLA instructions.

PR #5186, opened 2018-09-25
Add node and php to benchmark dashboard

BSBandme:


benchmarks/util/result_uploader.py
```diff
@@ -59,13 +59,14 @@ def upload_result(result_list, metadata):
...
-
-    bq = big_query_utils.create_big_query()
-    row = big_query_utils.make_row(str(uuid.uuid4()), new_result)
-    if not big_query_utils.insert_rows(bq, _PROJECT_ID, _DATASET,
-                                       _TABLE + "$" + _NOW,
-                                       [row]):
-      print('Error when uploading result', new_result)
+    print(labels_string)
+# 
+#     bq = big_query_utils.create_big_query()
```
TeBoring: Is this intentionally commented out?

BSBandme: Sorry that I forgot to uncomment this after testing.
Done.

BSBandme:
[Dashboard](http://performance-dot-grpc-testing.appspot.com/explore?dashboard=5715350830186496&widget=596013819&container=581906293)

benchmarks/util/result_parser.py
```diff
@@ -252,5 +284,11 @@ def get_result_from_file(cpp_file="",
     __parse_go_result(go_file)
   if synthetic_file != "":
     __parse_synthetic_result(synthetic_file)
+  if node_file != "":
+    __parse_js_php_result(node_file, "node")
```
TeBoring: Why php and js share the same method?

BSBandme: Their results have the same json format (which are constructed by myself). For code reuse I just use the same method to parse the result.

TeBoring: Could you give it a better name? e.g., __parse_XX_format_result and let js and php wrap it?
Why other parse methods work with different json format?

BSBandme: Done. The other results are most like using third_party benchmark tools for benchmark, e.g. c++ is using google/benchmark, the json result format are defined by the tool. But js and php and just using language internal time library to get the consuming time.

benchmarks/util/result_parser.py
```diff
@@ -236,11 +234,51 @@ def __parse_go_result(filename):
...
+#   {
+#     "filename": string,
+#     "benchmarks": {
+#       behavior: results,
+#       ...
+#     },
+#   },
+#   ...
+# ]
+def __parse_js_php_result(filename, language):
```
TeBoring: Could you change a name? js_php tells nothing about the format.

TeBoring: __parse_custom_result?

PR #5195, opened 2018-09-27
fix(python): fix deprecated collections.abc usage

### LINKED ISSUE #5135: [python] deprecated collections usage
The `Python` module has a deprecated usage pattern for `collections.abc`. Here's the warning I'm getting:

```
/usr/local/lib/python3.7/site-packages/google/protobuf/internal/containers.py:340: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  collections.MutableSequence.register(BaseContainer)
```

Any references to `collections.MutableSequence` should instead become `collections.abc.MutableSequence` in any Python >= 3.3. The same thing applies to `collections.Mapping` (eg. `collections.Mapping` -> `collections.abc.Mapping`).

TheKevJames:
Many classes within collections were moved to collections.abc in Python
3.3 -- their usage as imported directly from collections is now
deprecated as of Python 3.7 and will be removed soon.

The fallback import added in this PR can be removed entirely when
support for versions of Python prior to 3.3 is dropped.

Fixes #5135 

@anandolee I believe you wanted to review this.

### DIFF
--- a/python/google/protobuf/internal/containers.py
+++ b/python/google/protobuf/internal/containers.py
@@ -41,12 +41,16 @@ are:
 
 __author__ = 'petar@google.com (Petar Petrov)'
 
-import collections
+try:
+    # This fallback applies for all versions of Python before 3.3
+    import collections.abc as collections_abc
+except ImportError:
+    import collections as collections_abc
 import sys
 
 if sys.version_info[0] < 3:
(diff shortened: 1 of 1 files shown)


anandolee:
PR looks good. Thanks for the change.
Ruby failure is not related.

PR #5198, opened 2018-09-28
Revert "protobuf_generate(): add relative path to output dir"

hesmar:
This reverts commit 93f6b67eb217812266fb29a00977046789863aca.

Protoc adds already the relative path to the output directory.
Therefore, we have to remove this again from the output directory
to prevent adding it twice.


haberman:
Is it possible to add a test that will validate this behavior?

hesmar:
I am not sure if it is possible to add a test for this because the protobuf-config.cmake is generated during the install step, I think. So, the test has to be run after installing protobuf. Is there any example for this? I didn't implement tests for protobuf before, therefore I need some help...

Issue #5200, opened 2018-09-28
string subscript out of range when parse command line argument


windpenguin:
**What version of protobuf and what language are you using?**
Version: v3.5.2
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Windows10

**What runtime / compiler are you using (e.g., python version or gcc version)**
vs2010, win32

**What did you do?**
when i use protoc to generate proto files, i pass the proto path: "E:/GitRepo/ThirdParty/protobuf-3.5.2/src", the length of the proto path is 40.

protoc call function normalize(wstring path) in io_win32.cc to normalize path. in the for loop in line 158, it visit pos 40, path[i], which is out of range of proto path.

in vs2010-win32, it abort and tells that: string subscript out of range.
but in vs2017-win32, in the same situation, it is ok.

it is right to visit path[40] ? what should i do with this problem?

![snipaste_2018-09-28_23-43-59](https://user-images.githubusercontent.com/6115575/46218958-964e3600-c378-11e8-950f-3d8ff7d859d3.png)

acozzette:
@windpenguin I am not yet familiar with this path handling code, but would you be interested in sending us a pull request if you can create a fix for the out-of-range issue?

acozzette:
@windpenguin I just realized you already sent us a pull request, sorry for not noticing that before.

windpenguin:
@acidtonic i google the c++ string, and find that string null-terminator is no guaranteed in c++03. ref [this](https://stackoverflow.com/questions/11752705/does-stdstring-contain-null-terminator). i think in version 3.5.x, subscript check should be added.

acidtonic:
@windpenguin I think I was mistakenly mentioned when you likely meant to mention @acozzette (we both start with ac but if you in fact meant to add me just let me know)

Good luck!

windpenguin:
sorry for that @acidtonic, i just meat to mention @acozzette

PR #5203, opened 2018-09-29
Fix string subscript out of range when parse command line argument

windpenguin:
this pull request try to solve the issue #5200

### DIFF
--- a/src/google/protobuf/stubs/io_win32.cc
+++ b/src/google/protobuf/stubs/io_win32.cc
@@ -151,12 +151,13 @@ wstring normalize(wstring path) {
 
   static const wstring dot(L".");
   static const wstring dotdot(L"..");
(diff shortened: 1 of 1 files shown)


windpenguin:
I signed it!

acozzette:
@windpenguin I noticed that we seem to have already fixed this bug on the master branch with #4016, but we never backported that fix to the 3.5.x branch. Do you want to try backporting that fix? I think that approach may be a bit simpler since calling `c_str()` gives you back a string that is guaranteed to have a null byte at the end.

windpenguin:
@acozzette i checked #4016, but for reason that type of path is wstring on branch 3.5.x and is string on branch master, i add const char* into 3.5.x manually.

acozzette:
Thanks @windpenguin!

PR #5208, opened 2018-10-01
Add `Serializable` attribute to generated C# classes as compiler option

### LINKED ISSUE #5148: Add `Serializable` attribute to generated C# classes

Just the generated C#.

Generated C# classes should be serializable by default. The official method of implementing this is to use the `Serializable` attribute on class declarations.

The specific use case I have is to make the instances of the generated classes visible in the inspector in Unity3d, which relies on the `Serializable` attribute.

The C# code generator should add the `[System.Serializable]` attribute before every class declaration. Putting this behavior behind a command-line option is also reasonable, but not necessary.

I considered an automated tool to rewrite the generated `.cs` files to add the attribute.

**Additional context**

[Documentation for the Serializable attribute.](https://docs.microsoft.com/en-us/dotnet/api/system.serializableattribute?view=netcore-2.1)

ObsidianMinor:
Adds option "`serializable`" for C# protoc compiler that adds a `[global::System.SerializableAttribute]` above all generated classes.

Resolves #5148

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_generator.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_generator.cc
@@ -81,6 +81,8 @@ bool Generator::Generate(
       cli_options.base_namespace_specified = true;
     } else if (options[i].first == "internal_access") {
       cli_options.internal_access = true;
+    } else if (options[i].first == "serializable") {
+      cli_options.serializable = true;
     } else {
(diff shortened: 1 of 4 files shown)


Danl2620:
@anandolee @xfxyjwf Is there anything I can do (testing etc...) to prioritize this into a release?

anandolee:
Sorry I was on vacation. The change looks good, just wait for kokoro test

PR #5211, opened 2018-10-01
Fix parsing empty Struct Values from Json

leon-barrett:
This fixes a bug. When parsing a struct from JSON like
```python
struct = json_format.Parse('{"k": {}}', Struct())
```
then the struct's "k" value would end up not initialized, and accessing
the value would raise an error.
```python
In[1]: struct['k']
ValueError: Value not set
```
That seems to be because the Struct field of the Value was not set.
```python
In[2]: struct
Out[2]:
fields {
  key: "k"
  value {
  }
}
```

This commit makes sure that the Value's Struct field is initialized even
if the Struct has no values itself.

This commit also extends a test to cover this case.

### DIFF
--- a/python/google/protobuf/json_format.py
+++ b/python/google/protobuf/json_format.py
@@ -614,6 +614,9 @@ class _Parser(object):
     if not isinstance(value, dict):
       raise ParseError(
           'Struct must be in a dict which is {0}.'.format(value))
+    # Clear will mark the struct as modified so it will be created even if
+    # there are no values.
+    message.Clear()
     for key in value:
       self._ConvertValueMessage(value[key], message.fields[key])
     return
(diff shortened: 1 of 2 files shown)


python/google/protobuf/json_format.py
```diff
@@ -614,6 +614,9 @@ def _ConvertStructMessage(self, value, message):
     if not isinstance(value, dict):
       raise ParseError(
           'Struct must be in a dict which is {0}.'.format(value))
+    # Clear will mark the struct as modified so it will be created even if
+    # there are no values.
+    message.Clear()
```
anandolee: The change looks good to me and thanks for the fix!
Just notice that ListValue also called clear on line 608, but didn't find empty ListValue test. Can you help to add such tests as well.

leon-barrett: done

arcana261: Hey guys! This change breaks behavior.  Prior to this, structs were deep-merged by parse. Now they are not merged anymore...

Issue #5209, opened 2018-10-01
Link error on Windows due to use of ::close


GregTho:
Version: master
Language: C++
Windows
MSVC

**What did you do?**
1. Build something that needs stubs\io_win32.cc's open/close pair
4. See link errors for _close being undefined

The problem is that close is defined as:

int close(int fd) { return ::close(fd); }

whereas it should be:

int close(int fd) { return ::_close(fd); }

since _close is the converse of _open (as per MSDN).

PR #5212, opened 2018-10-02
Use ::_close rather than ::close in Win32 stubs.

GregTho:
The former is the proper converse of _open, which is used in
::google::protobuf::internal::win32::open. This resolves issue #5209.


acozzette:
@GregTho Thanks for the pull request!

PR #5214, opened 2018-10-02
remove obsolete parent

### LINKED ISSUE #5213: Bad parent pom for Java
Not sure how deep this one goes or who owns it but let's start here. In java/pom.xml I noticed this:

```
  <parent>
    <groupId>com.google</groupId>
    <artifactId>google</artifactId>
    <version>1</version>
  </parent>
```

I never saw that artifact before so I looked it up in Maven central. There are several later versions of com.google:google but none later than 2010. All reference googlecode.com as scm which is now dead.

Since I haven't seen this artifact anywhere except here, I think the answer is that we should just delete this parent element in pom.xml. However, if there's something there we need, then we need to figure out who owns this and push a long-overdue update.

elharo:
fix #5213 If nothing else, this will be one less network connection to Maven central on build.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -3,11 +3,6 @@
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>com.google</groupId>
-    <artifactId>google</artifactId>
-    <version>1</version>
-  </parent>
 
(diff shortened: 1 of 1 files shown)


BSBandme:
Is there any policy restriction for removing \<parent\>?

In fact there are still some others use this parent, like [Guice](https://github.com/google/guice/blob/master/pom.xml#L20-L24)
You may found more by searching in internal piper maven repository central.

elharo:
Guava is four versions past this one, and still out of date. Let me see if  can get that removed too.

elharo:
I know of no such policy. There are many current projects that don't use this or any other parent POM, and thus inherit from the default maven parent POM.

acozzette:
@elharo Thanks for the pull request!

PR #5217, opened 2018-10-02
Various Ruby encoding changes

tenderlove:
These commits are just a little clean up around encoding handling in the Ruby extension.  Each commit has commits for why I made the change.  In some cases, we can avoid some runtime overhead, in addition to deleting some code.

I *think* we can eventually remove the `kRubyStringUtf8Encoding` encoding constants.  (It looks like `kRubyStringASCIIEncoding` isn't used at all and can be removed).

Thanks!

### DIFF
--- a/ruby/ext/google/protobuf_c/encode_decode.c
+++ b/ruby/ext/google/protobuf_c/encode_decode.c
@@ -193,8 +193,7 @@ static void* appendstr_handler(void *closure,
                                const void *hd,
                                size_t size_hint) {
   VALUE ary = (VALUE)closure;
-  VALUE str = rb_str_new2("");
-  rb_enc_associate(str, kRubyStringUtf8Encoding);
+  VALUE str = rb_utf8_str_new_cstr("");
   RepeatedField_push_native(ary, &str);
   return (void*)str;
 }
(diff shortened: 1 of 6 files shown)


tenderlove:
Hi.  It looks like the build was failing because it's testing against Ruby 2.1, and Ruby 2.1 doesn't have the utf8 string helper functions.  Ruby 2.2 and older are [EOL'd by the Ruby core team](https://www.ruby-lang.org/en/downloads/branches/), so maybe they don't need to be supported?

I've backported the functions in case these older Ruby versions are important to support.

Thanks!

haberman:
Thanks for the heads-up on this, we're working on removing Ruby 2.1 and 2.2 from our tests and adding newer Ruby versions instead.

That should allow you to remove the backported functions.

See: https://github.com/protocolbuffers/protobuf/pull/5256

haberman:
@tenderlove We have updated our tests to remove Ruby 2.1/2.2 and add Ruby 2.3/2.4/2.5 instead. So you should be able to remove the compatibility functions from this PR. Thanks!

TeBoring:
Could you close this PR and resend a new PR? Our continuous test may have cached the old result. Sorry about the inconvenience.

haberman:
I'm going to close this for now. We're eager to accept these fixes, but it sounds like our CI cached the old test result. Please open a new PR for this and we will accept it.

(Also you can remove the compatibility fixes since we don't support the old Ruby versions any more).

PR #5220, opened 2018-10-03
Revert "Basic Proto2 support for Ruby gem"

TeBoring:
Reverts protocolbuffers/protobuf#4816
This change has broken continuous test: https://source.cloud.google.com/results/invocations/0cf66438-ad5f-4fe1-b576-d371dfe5f33d/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fruby_all%2Fpresubmit/log

Please fix tests and submit again.


hrsht:
@TeBoring I am not able to reproduce it locally. Is there a way I can trigger these on a PR that I create in order to debug?

TeBoring:
The failure is in our kokoro environment. I guess you don’t have access to
run that. I’ll debug it first.

hrsht:
Ok. Thanks!

hrsht:
@TeBoring Looks like this bug has existed even before my change and my change is not the cause.

I was able to reproduce it with a commit before my proto2 commit and with the same docker image used in the presubmit test link you provided above.

Here are the steps to reproduce:

Run the Docker image in interactive mode:
```
docker run -e CCACHE_DIR=/tmp/protobuf-ccache -e KOKORO_BUILD_NUMBER=672 -e KOKORO_BUILD_ID=0cf66438-ad5f-4fe1-b576-d371dfe5f33d -e EXTERNAL_GIT_ROOT=/var/local/kokoro/protobuf -e TEST_SET=ruby_all -v <host_protobuf_dir>:/var/local/kokoro/protobuf:ro -v /tmp/protobuf-ccache:/tmp/protobuf-ccache -w /var/local/git/protobuf --name=build_and_run_docker_efc343e2-9782-484d-a5b5-be34eccd32e3 -it grpctesting/protobuf_70e53ae1868cce2192317507ba0f02618dfbb2ba
```
Replace `<host_protobuf_dir>` in the above command with the protobuf directory on your local machine.

Once inside Docker, perform the rest of the steps in there
```
git clone /var/local/kokoro/protobuf
cd protobuf
git checkout 4426cb573356c5ff710bff8ac8a1170baac7dcf0 # Commit before my proto2 change
./autogen.sh
./configure
make
cd ruby
```

Change `tests/gc_test.rb` to add another require: `require 'test_import_pb'` right before `GC.stress` is reverted:
```
old_gc = GC.stress
GC.stress = 0x01 | 0x04
require 'generated_code_pb'
require 'test_import_pb'
GC.stress = old_gc
```

Now run 
```
bash --login -c rvm use ruby-2.1 && gem install bundler && bundle && rake gc_test
```

It is possible that you may have to run the above command a few times to trigger the failure (some runs succeed).

Seems like the trigger for this bug is requiring another `_pb` file. Since my change added that another require statement, the bug got triggered.

Let me know if you need any help with this.

TeBoring:
Will work on a different PR to fix the ruby issue.

PR #5227, opened 2018-10-05
Link to libatomic when necessary (eg. on armv6l)

### LINKED ISSUE #5219: Linking fails on armv6l due to missing libatomic
Version: v3.6.1

Linux (NixOS)

gcc 7.3.0

**What did you do?**
When I attempted to build protobuf for armv6l, I ran into the following linker error:
```
./.libs/libprotobuf.so: undefined reference to `__atomic_fetch_add_8'
```

On armv6l, protobuf needs to be linked to `libatomic` to provide this symbol. Manually adding `-latomic` to `LDFLAGS` fixes the issue.

lopsided98:
This PR fixes #5219. It implements detection of whether `std::atomic` can be used without linking to `libatomic` and adds `libatomic` to the linker arguments if this fails.

It is likely mostly harmless to always link to `libatomic` (at least with gcc and clang), but this solution was chosen because many other projects have similar logic in their build systems. It is only implemented for the autotools based build, not bazel or cmake, but this was deemed adequate in #5219. I assume the bazel build has the same issue, but I have not tested it.

I tested this on Linux on x86_64, armv6l and armv7l, but I am not very familiar with autotools so this may need to be cleaned up/tweaked.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -165,6 +165,22 @@ AS_IF([test "$with_zlib" != no], [
 ])
 AM_CONDITIONAL([HAVE_ZLIB], [test $HAVE_ZLIB = 1])
 
+dnl On some platforms, std::atomic needs a helper library
+AC_MSG_CHECKING(whether -latomic is needed)
+AC_LINK_IFELSE([AC_LANG_SOURCE([[
+  #include <atomic>
+  #include <cstdint>
+  std::atomic<std::int64_t> v;
+  int main() {
+    return v;
+  }
+]])], STD_ATOMIC_NEED_LIBATOMIC=no, STD_ATOMIC_NEED_LIBATOMIC=yes)
+AC_MSG_RESULT($STD_ATOMIC_NEED_LIBATOMIC)
+if test "x$STD_ATOMIC_NEED_LIBATOMIC" = xyes; then
+  LIBATOMIC_LIBS="-latomic"
+fi
+AC_SUBST([LIBATOMIC_LIBS])
+
 AS_IF([test "$with_protoc" != "no"], [
   PROTOC=$with_protoc
   AS_IF([test "$with_protoc" = "yes"], [
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -166,7 +166,7 @@ nobase_include_HEADERS =                                         \
 
 lib_LTLIBRARIES = libprotobuf-lite.la libprotobuf.la libprotoc.la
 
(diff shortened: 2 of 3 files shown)


lopsided98:
Here is an example of how it could be done with CMake: https://github.com/llvm-mirror/llvm/blob/master/cmake/modules/CheckAtomic.cmake

I was considering adapting this, but I'm not sure how that would work with licensing.

acozzette:
It looks like one test is failing because -latomic is being passed in a case where the linker doesn't know what to do with it. This is happening in a Docker environment that deliberately uses an old toolchain for the purpose of building protoc binaries that are compatible with old versions of libc. I haven't had a chance to figure out exactly what's going wrong but you could try running it by cd'ing to `kokoro/linux/cpp_distcheck` and running `./build.sh`. That requires Docker to be installed, though.

lopsided98:
That failure seems somewhat strange, because gcc 4.8 should have libatomic AFAIK. There is something else wrong though, because x86_64 builds shouldn't need to link to libatomic, but the configure check is claiming that it is necessary anyway. The test program is probably failing to compile for some other reason. I'll run the test myself and see if I can figure it out.

lopsided98:
The libatomic configure test program is failing to compile because `-std=c++11` is not in `CXXFLAGS` in the configure script. By default, the configure script [adds `-std=c++11` to `CXXFLAGS`](https://github.com/protocolbuffers/protobuf/blob/master/configure.ac#L91), but the [test script overrides `CXXFLAGS`](https://github.com/protocolbuffers/protobuf/blob/master/tests.sh#L19) and does not include `-std=c++11`. The build normally still works because `-std=c++11` gets added somewhere else in the build system.

I think the simplest way to fix this is to always add `-std=c++11` to `CXXFLAGS`, even when the user supplies their own `CXXFLAGS`. I have modified this PR to include this fix and now the test passes.

Also, the reason linking to libatomic failed is probably that centos has a separate package for libatomic, which is likely not installed.

acozzette:
@lopsided98 Thanks for tracking down the problem. It seems to me that it's probably best to just update tests.sh to set `-std=c++11` in `CXX_FLAGS` instead of adding that in the configure script, because someone might for example want to use C++14, and we wouldn't want to downgrade to C++11 in that case.

lopsided98:
Ok, done.

acozzette:
Thanks @lopsided98!

PR #5234, opened 2018-10-07
Add EnumTypeWrapper.__getattr__ to access values

Code0x58:
This is so you can do `MyEnum.FOO` as an alternative to `MyEnum.Value("FOO")`, more like [Python's enums](https://docs.python.org/3/library/enum.html)

### DIFF
--- a/python/google/protobuf/internal/enum_type_wrapper.py
+++ b/python/google/protobuf/internal/enum_type_wrapper.py
@@ -87,3 +87,9 @@ class EnumTypeWrapper(object):
     """
     return [(value_descriptor.name, value_descriptor.number)
             for value_descriptor in self._enum_type.values]
+
+  def __getattr__(self, name):
+    """Returns the value coresponding to the given enum name."""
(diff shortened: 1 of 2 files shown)


thisch:
According to git blame/praise, the EnumTypeWrapper class was created 6 years ago. This was before the enum class was introduced in the stdlib (with python3.4, which was released in 2014 https://www.python.org/download/releases/3.4.0/) I think we should use the enum module (or enum34 for python2.7) from the stdlib instead of this EnumTypeWrapper class. (Maybe there are still good reasons for keeping the EnumTypeWrapper class, which I'm not aware of)

Code0x58:
I'm not sure the effort is worth the [reward](https://docs.python.org/3/library/enum.html#how-are-enums-different), as the python's enum really seems to be for constructing enums, which is what we get from protobuf.

Code0x58:
I feel like that segfault in the Ruby tests wasn't me, I can rebase this evening

PR #5238, opened 2018-10-08
Fix the compiling error for xcode 10

JerryShih:
The xcode 10 removes the deprecated libstdc++ library. Now, we should
use libc++ instead.

Here is the error message when I try to build the protobuf with cpp implementation.
> warning: include path for stdlibc++ headers not found; pass '-std=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]
> In file included from google/protobuf/pyext/extension_dict.cc:34:
> ./google/protobuf/pyext/extension_dict.h:39:10: fatal error: 'memory' file not found
> #include <memory>

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -183,7 +183,8 @@ if __name__ == '__main__':
         extra_compile_args.append('-std=c++11')
 
     if sys.platform == 'darwin':
-      extra_compile_args.append("-Wno-shorten-64-to-32");
+      os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.9'
+      extra_compile_args.append("-Wno-shorten-64-to-32")
 
     # https://github.com/Theano/Theano/issues/4926
     if sys.platform == 'win32':


JerryShih:
I signed it!

JerryShih:
@anandolee
Could you please review this change?

JerryShih:
@TeBoring 
How about this change?

TeBoring:
Do you known since which version of xcode does libc++ exist? We probably can add it by default.

JerryShih:
@TeBoring
Maybe start from xcode5. What's the minimum requirement to build protobuf at mac? Could I change to libc++ by default?

https://github.com/nodejs/nan/issues/606

TeBoring:
We provide support for up to osx 10.6 for now. It seems xcode 4 can still be run on osx 10.6.
https://stackoverflow.com/questions/13047630/what-is-the-latest-version-of-xcode-i-can-use-on-10-6-8
However, we do cross-compiling. If we don't use libc++, our binary won't work on machines with only xcode10 installed.
So the problem is that it's impossible to have a binary that supports both xcode4 and xcode10. Is that true?

TeBoring:
https://gitlab.kitware.com/cmake/cmake/issues/18396
https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes

> Libgcc is obsoleted. Xcode 10 can no longer build apps with deployment targets of macOS 10.4 and 10.5. (42818150, 38035243)

I feel since xcode10 still claims to support osx10.6, it should not deprecate libstdc++...

TeBoring:
In https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes, it suggests:

> C++ projects must now migrate to libc++ and are recommended to set a deployment target of macOS 10.9 or later, or iOS 7 or later

Perhaps the proper change is to set MACOSX_DEPLOYMENT_TARGET: https://github.com/grpc/grpc/blame/0a1d0d25a2392a0fa70d8ad54c572bdc726b62d3/tools/distrib/python/grpcio_tools/setup.py#L123

JerryShih:
@TeBoring 
It looks like the xcode 4.3 has libc++.
https://stackoverflow.com/questions/9345271/xcode-4-3-and-c11-include-paths/13621792

> In https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes, it suggests:
> 
> > C++ projects must now migrate to libc++ and are recommended to set a deployment target of macOS 10.9 or later, or iOS 7 or later
> 
> Perhaps the proper change is to set MACOSX_DEPLOYMENT_TARGET: https://github.com/grpc/grpc/blame/0a1d0d25a2392a0fa70d8ad54c572bdc726b62d3/tools/distrib/python/grpcio_tools/setup.py#L123

I can't catch the idea that we could change  MACOSX_DEPLOYMENT_TARGET.  You said that the protobuf could run at 10.6. How could we change MACOSX_DEPLOYMENT_TARGET to 10.9?

TeBoring:
We have to support the newest xcode. For that, we can drop the older
version of Mac OS.

JerryShih:
@TeBoring 
So, what should I do for the next step?
We could replace the libstdc++ with libc++. The xcode 4.3 has libc++.
Do you mean I could add "os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.9'" to setup.py?

TeBoring:
Yes

JerryShih:
@TeBoring 
Here is the patch with "os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.9'" only.
Thank you.

TeBoring:
Could you do me a favor to send another PR with the same change to ```python-wheel-dev``` branch?
I need to run a test on the branch.

TeBoring:
Fixed in #5406

PR #5246, opened 2018-10-11
csharp generate identifiers for extensions

jswigart:
the goal of these changes is to allow the csharp code generator to emit named constants for the extensions defined in the proto file so that fetching them in csharp code doesn't need to use literal number constants, which are error prone and not very readable

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
@@ -33,7 +33,6 @@
 using Google.Protobuf.Reflection;
 using Google.Protobuf.WellKnownTypes;
 using NUnit.Framework;
-using System;
 using System.IO;
 using System.Linq;
 using UnitTest.Issues.TestProtos;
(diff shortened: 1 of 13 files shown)


jswigart:
I signed it!

anandolee:
Not quit understand how this change will be used as no test has been added for the usage. It looks a temp solution which should indeed work in descriptors not generated APIs. We are very serious on generated APIs because once it is published, it will be very hard to change/delete for compatibility reason.

jswigart:
This change basically brings the c# generator up to parity with the c++ generator in terms of using your named extension identifiers to create field constants so that you can fetch them with a named variable rather than a literal integer. That make sense?

jswigart:
I would think that whatever extension work that is being performed, fetching them by a readable identifier that maps to the field id would be part of it.

ObsidianMinor:
It is a part of it. [Making it type safe is a bigger part of it.](https://github.com/ObsidianMinor/protobuf/blob/csharp/proto2-rebased/csharp/src/Google.Protobuf.Test/SampleMessages.cs#L103) This PR would be completely replaced by full support, so what's the point?

jswigart:
My first check in was also a type safe variation but I ultimately changed it to be simpler like c++ so as not to totally break the api for existing users. I didn't know about your work or I wouldn't have bothered.

Feel free to close it.

PR #5248, opened 2018-10-12
 Mark sun.misc dependency as optional

fgdrf:
backport of #2811 to 3.6.x branch


fgdrf:
I've choosen the wrong destination branch, going to fix it with another pull request

PR #5250, opened 2018-10-12
Add ruby2.3,2.4.2.5 test on Mac and disable ruby2.1,2.2 test on Linux

TeBoring:


tests.sh
```diff
@@ -575,6 +573,9 @@ Usage: $0 { cpp |
             python_compatibility |
             ruby21 |
```
haberman: Can't we disable Ruby 2.1 everywhere and delete the respective files?

PR #5252, opened 2018-10-12
Fix ruby tests

TeBoring:


TeBoring:
Seems like the commented gc test still fails in ruby 2.3.

PR #5256, opened 2018-10-14
 Add ruby 2.3, 2.4 and 2.5 test for linux.

TeBoring:


TeBoring:
@haberman, it seems new version of ruby 2.3 and ruby 2.5 has fixed our broken tests.

PR #5265, opened 2018-10-16
JsonFormat supports add customized fields

milton0825:
Allow users to add plugin functions that evaluate the protobuf object and inject customized fields in JSON.

For example, given a protobuf IDL like:
```
message TestWrappers {
  google.protobuf.Int32Value int32_value = 1;
  google.protobuf.UInt32Value uint32_value = 2;
  google.protobuf.Int64Value int64_value = 3;
  google.protobuf.UInt64Value uint64_value = 4;
  google.protobuf.FloatValue float_value = 5;
  google.protobuf.DoubleValue double_value = 6;
  google.protobuf.BoolValue bool_value = 7;
  google.protobuf.StringValue string_value = 8;
  google.protobuf.BytesValue bytes_value = 9;
}
```

user may want to add a tag in JSON that counts the number fields set in the protobuf object:
```
{
  "number_of_fields_set": 2,
  "bool_value": false,
  "string_value": ""
}
```

The user can add a plugin function that evaluate the structure of the protobuf
object and then inject custom fields in JSON.

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -106,7 +106,70 @@ public class JsonFormat {
   public static Printer printer() {
     return new Printer(
         TypeRegistry.getEmptyTypeRegistry(), false, Collections.<FieldDescriptor>emptySet(),
-        false, false, false);
+        false, false, false,
+        Collections.<AdditionalFieldRegistry>emptyList());
+  }
+
+  /**
+   * An interface to register additional fields to JSON.
+   */
+  interface AdditionalFieldRegistry {
+    /**
+     * Defines how you infer additional fields in JSON based on the protobuf
+     * message.
+     */
+    List<AdditionalField> add(MessageOrBuilder message);
+  }
+
+  /**
+   * Describes the additional field you want to register.
+   */
+  public static class AdditionalField {
+    private final String fieldName;
+    private final String fieldValue;
+    private final boolean withQuote;
+
+    /**
+     * Describes the additional field.
+     * With fieldName = "user_name", fieldValue = "Bob", withQuote=true
+     * It translates to:
+     * {
+     *   "user_name": "Bob"
+     * }
+     * in JSON.
+     *
+     * With fieldName = "age", fieldValue = "13", withQuote=false
+     * It translates to:
+     * {
+     *   "age": 13
+     * }
+     * in JSON.
+     *
+     * @param fieldName name of the field
+     * @param fieldValue value of the field
+     * @param withQuote add quote to the value of the field
+     */
+    public AdditionalField(String fieldName, String fieldValue,
+                           boolean withQuote) {
+      this.fieldName = fieldName;
+      this.fieldValue = fieldValue;
+      this.withQuote = withQuote;
+    }
+
+    private String addQuote(String s) {
+      return "\"" + s + "\"";
+    }
+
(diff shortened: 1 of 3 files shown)


TeBoring:
Could you add some description to this change? like why is this needed?

milton0825:
I signed it!

TeBoring:
In your example, could you replace EmptyMessage with google.protobuf.Struct?

milton0825:
@TeBoring I have updated the example so that it is much more clear and does not require modification to the `.proto`

TeBoring:
I mean using google.protobuf.Struct can solve your problem of adding custom fields without any code change.

milton0825:
@TeBoring say we have a protobuf IDL like:
```
message TestWrappers {
  google.protobuf.Int32Value int32_value = 1;
  google.protobuf.UInt32Value uint32_value = 2;
  google.protobuf.Int64Value int64_value = 3;
  google.protobuf.UInt64Value uint64_value = 4;
  google.protobuf.FloatValue float_value = 5;
  google.protobuf.DoubleValue double_value = 6;
  google.protobuf.BoolValue bool_value = 7;
  google.protobuf.StringValue string_value = 8;
  google.protobuf.BytesValue bytes_value = 9;
}
```

user may want to add a tag in JSON (but not in the proto definition) that counts the number fields set in the protobuf object:
```
{
  "number_of_fields_set": 2,
  "bool_value": false,
  "string_value": ""
}
```

TeBoring:
You can define ```google.protobuf.Struct extendable = 10;```
```
$m = new TestWrappers();
// set whatever
$m->getExtendable()["number_of_fields_set"] = 2;
```

milton0825:
Say you have some legacy protobuf IDL that does not have `google.protobuf.Struct` in it. You are required to add additional field, e.g. `google.protobuf.Struct` to every one of them to enable this. 

What I am proposing is to support a way to evaluate `number_of_fields_set` without changing any protobuf definition.

TeBoring:
The result with custom fields cannot be decoded by protobuf message. Who will decode that?

milton0825:
If you want to transform the JSON back to protobuf, you can set `ignoringUnknownFields=true` to ignore these tags.

milton0825:
@TeBoring let me know what you think

TeBoring:
Could you wrap around protobuf?
In your code, use some third party json library to decode our json output, add custom fields and use third party json library to encode json?

milton0825:
@TeBoring 
We will lose the type information once the data is transformed to JSON. If we want to do tagging based on the type information, we will still need to look at the protobuf definition. The benefit of doing tagging in `JsonFormat` is that we have all the type information.

TeBoring:
However, the custom fields you added in are not defined in proto file. You cannot get type information for these fields from protobuf in any case.

milton0825:
I mean if we want to generate tag based on the fields defined in protobuf message, we will need the type information of those fields.

TeBoring:
Where are these tag? In json data?

milton0825:
Yes, tags only exist in JSON space

TeBoring:
Are these tags added as custom fields but they depend on defined fields’
type information?

milton0825:
Yes.

milton0825:
@TeBoring ping

TeBoring:
Sorry for the long delay. I think you should make your custom encoder to support this feature. You can probably use reflection api to do that. @BSBandme know more about it.

milton0825:
sure I will close this PR then

PR #5275, opened 2018-10-19
Added safety checks when malloc returns nil

parveen-bhatia:
Added safety checks to prevent exceptions/crash when malloc returns nil in **GPBDescriptor**

### DIFF
--- a/objectivec/GPBDescriptor.m
+++ b/objectivec/GPBDescriptor.m
@@ -819,6 +819,7 @@ uint32_t GPBFieldAlternateTag(GPBFieldDescriptor *self) {
       return;
     }
     uint32_t *offsets = malloc(valueCount_ * sizeof(uint32_t));
+    if (!offsets) return;
     const char *scan = valueNames_;
     for (uint32_t i = 0; i < valueCount_; ++i) {
       offsets[i] = (uint32_t)(scan - valueNames_);
@@ -851,6 +852,7 @@ uint32_t GPBFieldAlternateTag(GPBFieldDescriptor *self) {
   nameAsCStr += prefixLen;
 
   if (nameOffsets_ == NULL) [self calcValueNameOffsets];
+  if (nameOffsets_ == NULL) return NO;
 
   // Find it.
   for (uint32_t i = 0; i < valueCount_; ++i) {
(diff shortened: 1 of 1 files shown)


objectivec/GPBDescriptor.m
```diff
@@ -851,31 +853,33 @@ - (BOOL)getValue:(int32_t *)outValue forEnumName:(NSString *)name {
...
 
   if (nameOffsets_ == NULL) [self calcValueNameOffsets];
-
-  // Find it.
-  for (uint32_t i = 0; i < valueCount_; ++i) {
-    const char *valueName = valueNames_ + nameOffsets_[i];
-    if (strcmp(nameAsCStr, valueName) == 0) {
-      if (outValue) {
-        *outValue = values_[i];
+  if (nameOffsets_) {
```
thomasvl: Lets go ahead an used the early return style in all of these like you did for `getEnumNameForIndex:` and `getEnumTextFormatNameForIndex:` that way is will hopefully be more clear going forward how to handle a failure with calc.

objectivec/GPBDescriptor.m
```diff
@@ -819,13 +819,15 @@ - (void)calcValueNameOffsets {
       return;
     }
     uint32_t *offsets = malloc(valueCount_ * sizeof(uint32_t));
-    const char *scan = valueNames_;
-    for (uint32_t i = 0; i < valueCount_; ++i) {
-      offsets[i] = (uint32_t)(scan - valueNames_);
-      while (*scan != '\0') ++scan;
-      ++scan;  // Step over the null.
+    if (offsets) {
```
thomasvl: Use an early exit here.

I almost wonder if we should do an `NSAssert` in `DEBUG` builds also, but I guess it isn't worth it since it really shouldn't happen.

parveen-bhatia: Yeah, it is not happening in real scenarios. This is just to silent Static Analyzer messages.

parveen-bhatia:
@thomasvl fixed PR comments. Please re-review.

parveen-bhatia:
@thomasvl can you please take care of merging this, i am not sure to whom i should reach.

thomasvl:
> @thomasvl can you please take care of merging this, i am not sure to whom i should reach.

I will, i'm just waiting for a CI run, but it looks like github status says they are having issues delivering notifications, so that seems to be slowing CI picking this up.

parveen-bhatia:
> > @thomasvl can you please take care of merging this, i am not sure to whom i should reach.
> 
> I will, i'm just waiting for a CI run, but it looks like github status says they are having issues delivering notifications, so that seems to be slowing CI picking this up.

Thanks.

PR #5229, opened 2018-10-05
Remove Debug Statement for Ruby on Windows / MinGW

micw523:
This pull request fixes #5025. It fixes the problem that the protobuf library is built under debug mode for Windows / MINGW environment.

Issue #5025, opened 2018-08-09
Ruby gem 3.6.1 fails on windows


taylanisikdemir:
Hi,

I am seeing issues with protobuf ruby gem 3.6.1 on windows. Here is my repro:

`ret = Google::Protobuf::Struct.new`
`val = Google::Protobuf::Value.new`
`val.string_value = "some string"`
`ret.fields[“key”] ||= val`

After executing above I get the error below:
Assertion failed!
Program: C:\Ruby24-x64\bin\ruby.exe
../../../../ext/google/protobuf_c/upb.c, Line 6331
Expression: val.ctype == t->ctype

My environment:
OS: Windows Server 2016 version 1607 (build 14393.2363)
Ruby: 2.4.4p296 (2018-03-28 revision 63013) [x64-mingw32]
Gem: google-protobuf (3.6.1 x64-mingw32)

I don't see any problem with below environment:
OS: Windows Server 2016 version 1607 (build 14393.2363)
Ruby: 2.3.3p222 (2016-11-21 revision 56859) [i386-mingw32]
Gem: google-protobuf (3.0.2 x86-mingw32, 3.0.0 x86-mingw32)

Can someone help me debug this further?

TeBoring:
Could you gdb to check what’s val.ctype and what’s t->ctype?

pbasl:
Hi I am having the same issue. Any update on this??

TeBoring:
@taylanisikdemir did you get any progress?
BTW, does the same problem exist in previous versions?

pbasl:
@TeBoring I have the same issue in my rails app in development. It's almost impossible for me to determine what action exactly causes the error. I explicitly used the 3.5.1.2 version of the gem and the error doesn't show up anymore.

taylanisikdemir:
I also switched to an older version for now . It would be great if you guys take a look and identify the issue because apparently there's a regression.

TeBoring:
@haberman, is there any suggestion for debugging on windows?

TeBoring:
help needed to debug on windows. Anyone has experience to gdb a ruby c extension on windows?

micw523:
Checked to see that before assert:
val.ctype is 6352080
t->ctype is 4 (UPB_CTYPE_UINT64)

haberman:
I have a suspicion that the Windows compile is mixing flags. It appears this source file was compiled in debug mode, but I suspect other parts of the code were built with `NDEBUG`.

[...]

PR #5282, opened 2018-10-23
Remove Debug Statement for Ruby on Windows / MinGW

micw523:
#5229 Reopened that addresses #5025 .


micw523:
@TeBoring I think the previous results are still being used.

TeBoring:
Did you try copying code to another fork instead of send PR directly from
the original repo?

PR #5289, opened 2018-10-25
Fix bugs in our keyword conversion support for objectivec

dmaclach:
We have code for converting C/C++/Objc keywords that appear in protos
to convert them so that they can be compiled.
One of the things we need to be careful of is accidentally overriding methods
that Apple declares in NSObject. It turns out that we have run into issues
where we conflict with "hidden" methods in NSObject or methods added by
categories. method_dump.sh collects all of the methods we care about for
macOS and iOS and dumps them into objectivec_nsobject_methods.h which
is then included in objectivec_helpers.cc as part of the build.

Added a pile of tests to verify that conversions are happening as expected.


src/google/protobuf/compiler/objectivec/method_dump.sh
```diff
@@ -0,0 +1,154 @@
...
+    for (unsigned int i = 0; i < propCount; ++i) {
+      NSString *propertyName = [NSString stringWithUTF8String:property_getName(props[i])];
+      [reservedNames addObject:propertyName];
+    }
+    free(props);
+
+    // Note that methods have 2 defaults args (_cmd and SEL) so a method "0 arg method"
+    // actually has 2.
+    unsigned int methodCount;
+    Method *methods = class_copyMethodList(cls, &methodCount);
```
sergiocampama: does this include methods from categories that are always added? For example, `accessibilityLabel`

src/google/protobuf/compiler/objectivec/method_dump.sh
```diff
@@ -0,0 +1,154 @@
...
+    for (unsigned int i = 0; i < methodCount; ++i) {
+      int argCount = method_getNumberOfArguments(methods[i]);
+      NSString *methodName =
+          [NSString stringWithUTF8String:sel_getName(method_getName(methods[i]))];
+      if (argCount == 2) {
+        [reservedNames addObject:methodName];
+      }
+      if (argCount == 3 && [methodName hasPrefix:@"set"] && methodName.length > 4) {
+        NSString *firstLetter = [methodName substringWithRange:NSMakeRange(3,1)];
```
thomasvl: Should this check if this character was upper case?  i.e. - "setting:" would collect "ting" as a property name.

src/google/protobuf/compiler/objectivec/method_dump.sh
```diff
@@ -0,0 +1,154 @@
...
+        methodName = [firstLetter stringByAppendingString:restOfString];
+        [reservedNames addObject:methodName];
+      }
+    }
+    free(methods);
+
+    SEL sortSelector = @selector(caseInsensitiveCompare:);
+    NSArray *array = [reservedNames.allObjects sortedArrayUsingSelector:sortSelector];
+    for (NSString *item in array) {
+      // Some items with _ in them get returned in quotes, so do not add more.
```
thomasvl: Quotes from where?  The runtime adds the quotes?  ugh.

src/google/protobuf/compiler/objectivec/method_dump.sh
```diff
@@ -0,0 +1,154 @@
...
+// Autogenerated by method_dump.sh. Do not edit by hand.
+// Date: %DATE%
+// macOS: %MACOS%
+// iOS: %IOS%
+
+END_HEADER
+)
+
+# Check to make sure we are updating the correct file.
+if ! [ -e objectivec_nsobject_methods.h ]; then
```
thomasvl: using bash so `[[` and `]]` instead of `[` and `]`.  Also move the `!` inside the test then.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -192,15 +194,20 @@ const char* const kReservedWordList[] = {
...
     "Category", "Ivar", "Method", "Protocol",
 
     // NSObject Methods
-    // new is covered by C++ keywords.
-    "description", "debugDescription", "finalize", "hash", "dealloc", "init",
-    "class", "superclass", "retain", "release", "autorelease", "retainCount",
-    "zone", "isProxy", "copy", "mutableCopy", "classForCoder",
+    // These are brought in from an external header that is generated using
+    // method_dump.sh.
+#include "objectivec_nsobject_methods.h"
```
thomasvl: This is likely going to be fragile for different build systems.  It it likely safer to have the header generate a real structures and then include it the same way all the other headers are imported in this file.

The update the code using this existing list to feel this local list and this external one into the single set for checking.

src/google/protobuf/compiler/objectivec/method_dump.sh
thomasvl: It isn't in the header in the PR, but I don't think it is because the categories aren't captures, I think it is because the test binaries are only linking Foundation.  We likely need to link in Cocoa/UIKit on each platform to ensure a more complete list of category additions are picked up.

src/google/protobuf/compiler/objectivec/objectivec_nsobject_methods.h
```diff
@@ -0,0 +1,90 @@
+// NSObject methods
```
thomasvl: This file needs to be listed in the root `Makefile.am`.

src/google/protobuf/compiler/objectivec/method_dump.sh
dmaclach: Done

src/google/protobuf/compiler/objectivec/method_dump.sh
dmaclach: Done

src/google/protobuf/compiler/objectivec/method_dump.sh
dmaclach: Yeah.. the runtime adds quotes occasionally. I was surprised as well.

src/google/protobuf/compiler/objectivec/method_dump.sh
dmaclach: Done

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dmaclach: Done

src/google/protobuf/compiler/objectivec/objectivec_nsobject_methods.h
dmaclach: Done

src/google/protobuf/compiler/objectivec/method_dump.sh
```diff
@@ -0,0 +1,180 @@
...
+        // -5 because 3 for set, 1 for the firstLetter and 1 for the colon on the end.
+        NSString *restOfString =
+            [methodName substringWithRange:NSMakeRange(4, methodName.length - 5)];
+        methodName = [lowerFirstLetter stringByAppendingString:restOfString];
+        [reservedNames addObject:methodName];
+      }
+    }
+    free(methods);
+
+		printf("\n");
```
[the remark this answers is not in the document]
dmaclach: Done.. the printf wasn't even supposed to be there :)

PR #5293, opened 2018-10-28
Fix RepeatedField#first in Ruby gem

tobyhs:
Given an argument, the previous implementation was off by one (`.first(2)` would return 3 elements) compared to the `Enumerable#first` method.


TeBoring:
Thanks for the fix!

PR #5297, opened 2018-10-29
Remove unused types

brian-peloton:
They're not doing anything, so might as well get rid of them.

I was running a static analysis tool over protobuf, looking for things to fix in local changes, and noticed these.


acozzette:
Thanks for the pull request!

brian-peloton:
Thanks for taking a look at it so quickly!

PR #5298, opened 2018-10-30
Import versions from @bazel_skylib//lib:versions.bzl

rogerhub:
@bazel_skylib//:lib is supposedly deprecated. The 'versions' struct in lib.bzl was committed to bazel-skylib at the same time as //lib:versions.bzl, so there should be no backward incompatibility.

### DIFF
--- a/examples/WORKSPACE
+++ b/examples/WORKSPACE
@@ -39,5 +39,5 @@ http_archive(
     urls = ["https://github.com/bazelbuild/bazel-skylib/archive/2169ae1c374aab4a09aa90e65efe1a3aad4e279b.tar.gz"],
 )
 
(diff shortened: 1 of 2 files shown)


JaredNeil:
Should also update [examples/WORKSPACE#L42](https://github.com/protocolbuffers/protobuf/blob/master/examples/WORKSPACE#L42)

rogerhub:
> Should also update [examples/WORKSPACE#L42](https://github.com/protocolbuffers/protobuf/blob/master/examples/WORKSPACE#L42)

Done. Please take a look.

acozzette:
Thanks for the PR! Sorry for the delay merging this.

PR #5300, opened 2018-10-30
Build updates

joshkel:
Make language instructions' documentation links clickable. (See 86208c5.)

Add a `git submodules` check to autogen.sh. Running `./autogen.sh && ./configure && make` happens to work without initializing git submodules - but autogen.sh writes to what is supposed to be a git submodules directory, so trying to later initialize git submodules fails. Adding a check makes building a bit more reliable for people who miss this step.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -26,6 +26,17 @@ __EOF__
   exit 1
 fi
 
+# Verify the third_party directory before we start writing to it below.
+if test ! -d third_party/googletest/googletest; then
+  cat >&2 << __EOF__
+Third party code is missing.  Please run the following command, if you have
+not yet done so:
+
+git submodule update --init --recursive
+__EOF__
+  exit 1
+fi
+
 set -ex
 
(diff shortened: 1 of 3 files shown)


TeBoring:
I think in autogen.sh, you should just create the m4 dir only if the ```third_party/googletest``` .

TeBoring:
Any update?

joshkel:
Sorry for the delayed reply.

If the [installation instructions](https://github.com/protocolbuffers/protobuf/tree/master/src#c-installation---unix) say to make sure that submodules are initialized, and the test suite requires that submodules be initialized, and autogen.sh had assumed that submodules were initialized, I thought it might be better to enforce that?

I'm happy to do whichever you prefer; just let me know. And thank you for the reply and follow-up.

TeBoring:
any update?

joshkel:
I've incorporated your feedback, @TeBoring. I'm sorry it's taken me so long to get back to you on this.

PR #5302, opened 2018-10-31
Fix potential linking errors for MSVC

peterjc123:
When the moving semantics is used on a Type in the statically built library, the following error will be thrown: `error LNK2019: unresolved external symbol "const NAMESPACE::Type::`vftable'"`. It can be solved by moving the implementation of the moving constructor to the source file.

### DIFF
--- a/src/google/protobuf/type.pb.cc
+++ b/src/google/protobuf/type.pb.cc
@@ -443,6 +443,10 @@ Type::Type(const Type& from)
   syntax_ = from.syntax_;
   // @@protoc_insertion_point(copy_constructor:google.protobuf.Type)
 }
+Type::Type(Type&& from) noexcept
+  : Type() {
+  *this = ::std::move(from);
+}
 
 void Type::SharedCtor() {
   ::google::protobuf::internal::InitSCC(
@@ -1089,6 +1093,10 @@ Field::Field(const Field& from)
     static_cast<size_t>(reinterpret_cast<char*>(&packed_) -
     reinterpret_cast<char*>(&kind_)) + sizeof(packed_));
   // @@protoc_insertion_point(copy_constructor:google.protobuf.Field)
+}
+Field::Field(Field&& from) noexcept
+  : Field() {
+  *this = ::std::move(from);
 }
 
 void Field::SharedCtor() {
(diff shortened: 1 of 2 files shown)


acozzette:
@peterjc123 All the `pb.h` and `pb.cc` files are generated by `protoc` (the proto compiler), so for this change to be permanent it has be made in `protoc`. Do you have any more details on why this is needed, though? Is this a workaround for an MSVC bug?

peterjc123:
@acozzette Could you tell you how to change in `protoc`? BTW, I have no idea why this error triggered and this one may be a workaround for a potential MSVC bug.

acozzette:
To make this change in `protoc` you would need to edit the code around [here](https://github.com/protocolbuffers/protobuf/blob/cd95c791497747043200755a4e45e1ead98e85b3/src/google/protobuf/compiler/cpp/cpp_message.cc#L998). But I think we would be hesitant to accept this change without knowing more about the linker error and how this solves it.

peterjc123:
@davidbrownellMS Could you please add some context here?

peterjc123:
cc. @davidbrownellWork @davidbrownell

davidbrownellWork:
I've been looking at this issue off and on and still don't have a clear understanding of why it is happening. I have 2 guesses, neither of which I have been able to confirm.

1) The most likely is that the header files are being included by object files compiled with settings that impact the generated ABI; when it comes time to perform the final linking, the ABIs are incompatible which is causing the linker error. I am working through some discrepancies between the use of /EHsc and /EHa that *could* have this kind of impact, although I don't think it is causing this particular error.

2) A less likely, but still possible scenario is that this is a compiler bug. CUDA forces us to use an older version of MSVC, and it support around noexcept and constexpr isn't the best (for example, here is a change that I needed to make to work around a similar limitation in the compiler (https://github.com/davidbrownellWork/pytorch/pull/1/files#diff-9aba44730805b771612ecbc074298fcbR32). I'm always hesitant to blame the compiler, because I usually turn out to be wrong when I have done that in the past :).

acozzette:
Let me close this PR since there hasn't been any activity on it for a while.

PR #5309, opened 2018-11-01
Support suppressing the objc package prefix checks on a list of files.

thomasvl:


src/google/protobuf/compiler/objectivec/objectivec_generator.cc
```diff
@@ -89,6 +89,12 @@ bool ObjectiveCGenerator::GenerateAll(const std::vector<const FileDescriptor*>&
       // There is no validation that the prefixes are good prefixes, it is
       // assumed that they are when you create the file.
       generation_options.expected_prefixes_path = options[i].second;
+    } else if (options[i].first == "expected_prefixes_suppressions") {
+      // A semicolon delimited string that lists the paths of .proto files to
+      // skip the package prefix validations (expected_prefixes_path). This is
+      // provided as an "out", to skip some files being becked.
```
dmaclach: What is becked?

A semicolon delimited string that lists the paths of .proto files to exclude from the package prefix validations...

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -1169,6 +1173,15 @@ bool ValidateObjCClassPrefixes(const std::vector<const FileDescriptor*>& files,
   }
 
   for (int i = 0; i < files.size(); i++) {
+    bool should_skip =
+      (std::find(generation_options.expected_prefixes_suppressions.begin(),
+                 generation_options.expected_prefixes_suppressions.end(),
+                 files[i]->name())
+          != generation_options.expected_prefixes_suppressions.end());
+    if (should_skip) {
+      return true;
```
dmaclach: Should this be 'return true', or should this be a 'continue'? Don't we want to validate all of the files?

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
thomasvl: Done

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: done.

PR #5313, opened 2018-11-04
Fix most of the unused variable warnings in Ruby tests

tobyhs:


ruby/tests/common_tests.rb
```diff
@@ -614,7 +614,6 @@ def test_oneof
     assert d2 == d
 
     encoded_field_a = proto_module::OneofMessage.encode(proto_module::OneofMessage.new(:a => "string"))
-    encoded_field_b = proto_module::OneofMessage.encode(proto_module::OneofMessage.new(:b => 1000))
```
TeBoring: Do you need to remove the whole line?

TeBoring: Should it be used some how later?

tobyhs: I added `encoded_field_b` back in the latest commit so it is used in 2 places after, but I'm not sure what the original author's intent was.

PR #5323, opened 2018-11-08
compiler/cpp: replace NULL with nullptr

aconchillo:
This avoids compiler errors/warnings if zero-as-null-pointer-constant is considered an error or warning.

Not sure this is something you want as is. May be an option would be to pass a flag to `protoc` to specify if you want code for c++11 or older, for example?

### DIFF
--- a/src/google/protobuf/any.pb.cc
+++ b/src/google/protobuf/any.pb.cc
@@ -134,13 +134,13 @@ const int Any::kValueFieldNumber;
 #endif  // !defined(_MSC_VER) || _MSC_VER >= 1900
 
 Any::Any()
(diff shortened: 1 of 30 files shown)


acozzette:
@aconchillo Thanks for the pull request! Sorry but I just merged a big change that seems to have caused a conflict--if you wouldn't mind just fixing the merge conflict then I'll merge this as soon as that's done.

aconchillo:
Hi @acozzette! Done. I also called `generate_descriptor_proto.sh` so I could run the unit tests, all good. I updated the commit with the autoegenerated changes. Let me know if there's anything else.

acozzette:
Thanks @aconchillo!

Issue #4489, opened 2018-04-08
need a Warning on the usage of pure Python implementation


jizhilong:
In some complexed circumstance like grpc services, a pure python implementation may dramatically take down the whole performance, if there was a message in the logs explicitly warning people this would be a bottleneck, the time costed for troubleshooting such issues would be decreased a lot.

anandolee:
Cpp extension is not the default implementation yet and the major google internal projects are still using pure python. Add this warning will produce annoying message for lot of projects.

We are still working on fixing the behavior difference of pure python and cpp extension and planning to migrate the default implementation to cpp extension. I'd expect to see this warning added after that

PR #5334, opened 2018-11-10
dump warning on usage of pure python implementaion on non-pypy platforms

jizhilong:
for closing #4489


jizhilong:
I signed it!

anandolee:
Pure python is not the default implementation yet and the major google internal projects are still using pure python. Add this warning will know produce annoying message for lot of projects.

We are still working on fixing the behavior difference of pure python and cpp extension and planning to migrate the default implementation to cpp extension. I'd expect to see this warning added after that

PR #5339, opened 2018-11-12
Increase C# default recursion limit to 100

jskeet:
This matches the Java and C++ defaults.


jskeet:
I suspect the failures are due to backward compatibility checks. Will need to dig in to check.

jskeet:
Ah, compatibility tests were a problem. Added a second commit.

jskeet:
The single failing test is not caused by this PR. PTAL.

PR #5342, opened 2018-11-12
PHP: Add Enum methods for converting to/from strings

michaelbausor:
This PR migrates the changes in https://github.com/protocolbuffers/protobuf/pull/4534 and rebases them onto master branch. cc @TeBoring @fiboknacky


TeBoring:
It seems some tests are broken. Could you help take a look?

michaelbausor:
@TeBoring updated to not use EnumTrait, PTAL

TeBoring:
It seems the test is still broken. Basically, GPBUtil should also be avoided in generated code.

TeBoring:
Doesn't simply expand the GPBUtil method in the generated code work?

michaelbausor:
Yes, it does. I thought that using GPBUtil would work, but it seems not (I'll look into why my local testing caught the earlier error with EnumTrait, but not this one GPBUtil).

michaelbausor:
Updated, PTAL

TeBoring:
Still some broken tests. Please take a look.

php/tests/well_known_test.php
```diff
@@ -392,4 +392,33 @@ public function testWrappers()
...
+            $this->assertSame($v, $class::value($k));
+        }
+    }
+
+    public function enumNameValueConversionDataProvider()
+    {
+        return [
+            ['\Google\Protobuf\Field\Cardinality'],
+            ['\Google\Protobuf\Field\Kind'],
+            ['\Google\Protobuf\Internal\FieldDescriptorProto\Label'],
```
TeBoring: These messages under Google\Protobuf\Internal are internal only and don't have implementation in c extension.
We don't need to test them.

php/src/Google/Protobuf/Syntax.php
```diff
@@ -4,6 +4,8 @@
 
 namespace Google\Protobuf;
 
+use UnexpectedValueException;
```
TeBoring: I got warning: The use statement with non-compound name 'UnexpectedValueException' has no effect
Remove that?

TeBoring:
Implemented c extension in the last commit.

php/src/Google/Protobuf/Syntax.php
michaelbausor: I found that I got this warning only for the no-namespace enum. Is that what you found? I have added a commit to remove the use statement from the no-namespace case.

TeBoring: yes. For NoNamespace enum

TeBoring:
@michaelbausor Do you want to take another look?

bshaffer:
Hooray! nice work guys!

beccasaurus:
@michaelbausor et al – any update on when this will make it into the PHP protobuf library?

Tried using this today and there is no method `::name` on my enum class, presuming I'm calling it correctly.

TeBoring:
3.7.0rc2 should already have this

PR #5348, opened 2018-11-13
Removed stubs/singleton.h

acozzette:
This header is no longer used, so we can delete it. This PR also tweaks Makefile.am
to fix the failing distcheck test.


acozzette:
@TeBoring Friendly ping, it would be good if we could merge this soon because it also fixes the failing distcheck test.

PR #5349, opened 2018-11-13
cpp_generator: Add generator options for --cpp_out.

Allen-Webb:
Previously --cpp_out:lite was available to force the optimization mode
to the lite runtime. This change generalizes that change to work for
--cpp_out=speed:... and --cpp_out=code_size:...
This is needed to override the file specific options in cases where
libprotobuf-mutator is used for fuzzing since it depends on reflection.

Signed-off-by: Allen Webb <allenwebb@google.com>


src/google/protobuf/compiler/cpp/cpp_options.h
```diff
@@ -42,14 +42,21 @@ class AccessInfoMap;
 
 namespace cpp {
 
+enum class EnforceOptimizeMode {
+  kNoEnforcement, // Use the runtime specified by the file specific options.
+  kSpeed, // This is the full runtime.
+  kCodeSize,
```
acozzette: Could you actually leave out the code-size mode for now? That mode seems to be only rarely useful, so I think we should avoid making it more visible. We can always add a flag for it in the future if someone has a good use case for it.

Allen-Webb: Done.

Allen-Webb:
I do not see how my change could have caused the failure:
java/bom/pom.xml

It looks like that same failure has shown up on other pull requests as well:
https://github.com/protocolbuffers/protobuf/pull/5342

When I checked the commit history it looks like there is a pull request that test passed for initially but failed after merge:
"compiler/cpp: replace NULL with nullptr"

Allen-Webb:
It looks like that failure might be fixed by: https://github.com/protocolbuffers/protobuf/commit/c0bc265b675ab20b9fec7e00a2051675b21306d3

PR #5350, opened 2018-11-14
C# Proto2 feature : Extensions

ObsidianMinor:
We're at the final stretch.

This PR adds support for extensions in the C# library and code generator.


src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
```diff
@@ -226,7 +226,18 @@ void PrimitiveFieldGenerator::GenerateCloningCode(io::Printer* printer) {
 void PrimitiveFieldGenerator::GenerateCodecCode(io::Printer* printer) {
   printer->Print(
     variables_,
-    "pb::FieldCodec.For$capitalized_type_name$($tag$)");
+    "pb::FieldCodec.For$capitalized_type_name$($tag$, $default_value$)");
```
anandolee: Looks like this PR has several changes related to default value support?

src/google/protobuf/compiler/csharp/csharp_field_base.cc
```diff
@@ -57,24 +57,41 @@ void FieldGeneratorBase::SetCommonFieldVariables(
   // repeated fields varies by wire format. The wire format is encoded in the bottom 3 bits, which
   // never effects the tag size.
   int tag_size = internal::WireFormat::TagSize(descriptor_->number(), descriptor_->type());
+  int part_tag_size = tag_size;
```
anandolee: No need to copy tag_size I guess

src/google/protobuf/compiler/csharp/csharp_field_base.cc
```diff
@@ -57,24 +57,41 @@ void FieldGeneratorBase::SetCommonFieldVariables(
...
+  for (int i = 1; i < part_tag_size; i++) {
     tag_bytes += ", " + SimpleItoa(tag_array[i]);
   }
 
-  (*variables)["access_level"] = "public";
   (*variables)["tag"] = SimpleItoa(tag);
   (*variables)["tag_size"] = SimpleItoa(tag_size);
   (*variables)["tag_bytes"] = tag_bytes;
 
+  if (descriptor_->type() == FieldDescriptor::Type::TYPE_GROUP) {
```
anandolee: This change is related to group support not extension?

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
ObsidianMinor: The changes made for default values are specifically changes to make default values work with extensions.

src/google/protobuf/compiler/csharp/csharp_field_base.cc
ObsidianMinor: Same as above, it's cleanup that I found necessary to implement the changes for extensions.

src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
```diff
@@ -60,7 +60,7 @@ void SourceGeneratorBase::WriteGeneratedCodeAttributes(io::Printer* printer) {
 }
 
 std::string SourceGeneratorBase::class_access_level() {
-  return (IsDescriptorProto(descriptor_) || this->options()->internal_access) ? "internal" : "public";
+  return this->options()->internal_access ? "internal" : "public";
```
anandolee: I don't think we can make the generated descriptor file to public now because proto2 features are not well tested. Can this change hold until the tests are added and proto2 gen is public?

Makefile.am
```diff
@@ -100,7 +100,6 @@ csharp_EXTRA_DIST=                                                           \
   csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs                       \
   csharp/src/Google.Protobuf.Test/JsonParserTest.cs                          \
   csharp/src/Google.Protobuf.Test/JsonTokenizerTest.cs                       \
-  csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs            \
```
anandolee: CustomOptionsTest.cs  is removed, does this change breaks current usages for custom options?

src/google/protobuf/compiler/csharp/csharp_message_field.cc
```diff
@@ -169,7 +154,7 @@ void MessageFieldGenerator::GenerateSerializedSizeCode(io::Printer* printer) {
     printer->Print(
       variables_,
       "if ($has_property_check$) {\n"
-      "  size += $tag_size$ + $tag_size$ + pb::CodedOutputStream.ComputeGroupSize($property_name$);\n"
+      "  size += $tag_size$ + pb::CodedOutputStream.ComputeGroupSize($property_name$);\n"
```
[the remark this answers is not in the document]
ObsidianMinor: Because the changes [over here](https://github.com/protocolbuffers/protobuf/pull/5350#discussion-diff-237694119R60) make tag_size the full tag size instead of half the tag size.

Makefile.am
ObsidianMinor: Yes. I realized it wasn't a good idea a few days ago and have been working on a workaround that lets CustomOptions work through the ExtensionSet.

src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
ObsidianMinor: Sure.

Makefile.am
ObsidianMinor: With [the most recent commits](https://github.com/protocolbuffers/protobuf/pull/5350/commits/9084c23dbfabb738c28fb445bc6d62df309898ea) I reimplemented CustomOptions through the ExtensionSet and removed tests that no longer apply to the current implementation (empty set equality and reading values tests).

src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
ObsidianMinor: I reintroduced the internal access level for the descriptor proto. There were changes I manually made to generated test code to make sure the access levels matched up. I also commented out the CustomOptions deprecation for all the descriptors. The commented descriptor GetOption methods will be uncommented with the next PR.

anandolee:
Can you also make sure this PR is work with C# 6.0?  We've got an issue for C# 6.0 compatible, and was working on it in #5445

ObsidianMinor:
Sure but after #5445 is pulled so i can rebase on top of it.

ObsidianMinor:
Now C# 6 compatible. Can we get another kokoro run, @anandolee?

csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
```diff
@@ -99,66 +99,6 @@ public class CustomOptionsTest
     {
         delegate bool OptionFetcher<T>(int field, out T value);
 
-        [Test]
```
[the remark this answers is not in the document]
ObsidianMinor: They were deleted since they tested a part of how CustomOptions were implemented before. Since those implementation details don't apply anymore, the tests should be deleted. Specifically:

* EmptyOptionsIsShared tested to make sure two empty CustomOptions instances are the same empty instance

* SimpleIntegerTest tested parsing fields, something CustomOptions no longer does

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -127,6 +128,26 @@ public EnumValueDescriptor FindValueByName(string name)
         /// <summary>
         /// The (possibly empty) set of custom options for this enum.
         /// </summary>
-        public CustomOptions CustomOptions => Proto.Options?.CustomOptions ?? CustomOptions.Empty;
+        //[Obsolete("CustomOptions are obsolete. Use GetOption")]
```
ObsidianMinor: > This change added extension and will Obsolete custom options. 

It does not actually obsolete custom options yet, @anandolee, since I can't add the necessary APIs to replace custom options until the descriptor protos have been made public. Originally it obsoleted custom options but now it does not due to [this review](https://github.com/protocolbuffers/protobuf/pull/5350#discussion_r239288615).

jskeet:
I'm afraid I wouldn't have time to look at this until mid January at the earliest

anandolee:
@jskeet 
Understand. Most people will be OOO for next two weeks. I will review this PR. Can you help the review the next (final) PR?  The next PR will enable proto2 generator, change the descriptor protos public, enable all the APIs Sydney has added and will add all the tests. I assume it will be late January or February

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
anandolee: Ack. Thanks for make it clear

csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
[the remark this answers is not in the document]
anandolee: Since they are only test internal implementations, they are safe to be deleted. Thanks

jskeet:
I'll see when the time comes, but hopefully I'll be able to help. It'll have to be somewhat time limited though.

lostindark:
This seems has been opened for quite some time. Any updates?
I'm eager to test proto2 support.

ObsidianMinor:
I believe we're still waiting for @anandolee to review @lostindark

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// Used for keeping track of extensions in messages. 
+    /// <see cref="IExtensionMessage{T}"/> methods route to this set.
+    /// 
+    /// Most users will not need to use this class directly
+    /// </summary>
+    /// <typeparam name="TTarget">The message type that extensions in this set target</typeparam>
+    public sealed class ExtensionSet<TTarget> where TTarget : IExtensionMessage<TTarget>
+    {
+        internal Dictionary<Extension, IExtensionValue> ValuesByIdentifier { get; } = new Dictionary<Extension, IExtensionValue>();
+        internal Dictionary<int, IExtensionValue> ValuesByNumber { get; } = new Dictionary<int, IExtensionValue>();
```
anandolee: I think only one Dictionary is needed?

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// 
+    /// Most users will not use this class directly
+    /// </summary>
+    public static class ExtensionSet
+    {
+        private static IExtensionValue GetValue<TTarget>(ref ExtensionSet<TTarget> set, Extension extension) where TTarget : IExtensionMessage<TTarget>
+        {
+            if (!set.ValuesByIdentifier.ContainsKey(extension))
+            {
+                Register(ref set, extension);
```
anandolee: To my understanding, get should not insert if not present. Maybe change the function name.

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,294 @@
...
+            }
+            return set.ValuesByIdentifier[extension];
+        }
+
+        /// <summary>
+        /// Registers the specified extension in this set
+        /// </summary>
+        public static void Register<TTarget>(ref ExtensionSet<TTarget> set, Extension extension) where TTarget : IExtensionMessage<TTarget>
```
anandolee: To not confusing with ExtensionRegistry, I'd prefer to use Add() for ExtensionSet and use Register() for ExtensionRegistry

csharp/src/Google.Protobuf/ExtensionRegistry.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <summary>
+        /// Returns whether the registry is readonly
+        /// </summary>
+        bool ICollection<Extension>.IsReadOnly => false;
+
+        /// <summary>
+        /// Adds the specified extension to the registry
+        /// </summary>
+        public void Add(Extension extension)
+        {
```
anandolee: To not confusing ExtensionSet with ExtensionRegistry, I'd prefer to use Add() for ExtensionSet and use Register() for ExtensionRegistry

csharp/src/Google.Protobuf/ExtensionSet.cs
ObsidianMinor: I used two since one is keyed to each value by reference (for accessing values by identifier) and the other is keyed to each value by tag (for accessing values by input stream tag). Thus instead of having to manually search one dictionary for each value by reference or tag, it's fast for both. The trade-off being slightly more memory usage.

anandolee: It is easy to get the field number from reference: extension.FieldNumber. Only keep ValuesByNumber will be good enough?

ObsidianMinor: Ah, I see what you're saying. You're right, one dictionary will work.

csharp/src/Google.Protobuf/ExtensionRegistry.cs
ObsidianMinor: `Add` comes from the implementation for `ICollection<T>.Add` so I can't rename it unless I make it an explicit implementation and add a `Register` method that just calls `Add`.

csharp/src/Google.Protobuf/ExtensionSet.cs
ObsidianMinor: Oh, in that case I'll remove the Register call.

csharp/src/Google.Protobuf/ExtensionSet.cs
ObsidianMinor: ExtensionSet is a generated code API. It shouldn't be used by consumers so I don't think it will be confusing.

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -584,6 +589,10 @@ public void ReadMessage(IMessage builder)
             }
             int oldLimit = PushLimit(length);
             ++recursionDepth;
+            if (builder is IExtensionMessage)
+            {
+                ExtensionRegistry?.RegisterExtensionsFor(builder as IExtensionMessage);
```
anandolee: Why need to register all the extensions to this message? Can ExtensionSet.ValuesByNumber only contain the extensions that are/were set/get?

ObsidianMinor: I believe the reason I did that (it's been nearly a year at this point) was because I needed a way to implement extensions for newly created messages. But now that you mention it, it does have the unintended side effect of adding extensions not added yet which may be good or bad depending on how you look at it.

If this is something we don't want ReadMessage to do, I know a way I can changed it so extensions can only be registered for newly created messages.

anandolee: Some of our internal .proto usages has a lot of extensions but only a few are used. It it a waste to make each new message has such a big dict. 
Have you checked how java implemented it?

csharp/src/Google.Protobuf/ExtensionRegistry.cs
anandolee: Make sense to keep the current name then

csharp/src/Google.Protobuf/CodedInputStream.cs
ObsidianMinor: You're right, I never considered many extension defined for a type where only a few are used. I checked Java's implementation and redesigned ExtensionRegistry and ExtensionSet to more closely match their functionality. Now extensions are only registered from a registry when encountered while reading from the input.

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,341 @@
...
+            else 
+            {
+                return GetDefaultValue<TValue>();
+            }
+        }
+
+        /// <summary>
+        /// Gets the value of the specified repeated extension
+        /// </summary>
+        public static RepeatedField<TValue> Get<TTarget, TValue>(ref ExtensionSet<TTarget> set, RepeatedExtension<TTarget, TValue> extension) where TTarget : IExtensionMessage<TTarget>
```
ObsidianMinor: I'm actually unsure about this one. I have a couple ways I could implement it and they each have pros and cons but I'm unsure which would be the best fit for the lib.

1. Current implementation
Pros: Matches with other repeated field APIs. Doesn't require null checks or separate "RegisterExtension" calls.
Cons: Simple calls to read members such as Length can allocate a extension set, extension value, and repeated field.

2. Return null on empty set or non-registered value
Pros: Calls to check properties like length don't allocate
Cons: Null checks required which don't match other APIs that use repeated fields. RegisterExtension call required to create and use field

3. Match the Java implementation
Pros: Doesn't require null checks and calls to length properties don't allocate.
Cons: Massive difference in API. I personally wouldn't chose it since it's not very "Google.Protobuf" in its API.

ObsidianMinor: Another alternative I've just considered is a `GetOrRegister` extension method. Get returns null as to not allocate and the extension method will register if Get returns null, thus not requiring a separate call.

anandolee: For proposal 2, how users set/add repeated field if return null?

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -250,6 +261,29 @@ void MessageGenerator::Generate(io::Printer* printer) {
   GenerateMessageSerializationMethods(printer);
   GenerateMergingMethods(printer);
 
+  if (has_extension_ranges_) {
```
anandolee: Is it possible to: Make IExtensionMessage a abstract class, put  _extensions and the following implementation into IExtensionMessage (IExtendableMessage)?  I think we used to have a similar discussion on IMessage, not sure if they have the same situation.

Then no need to generate the duplicate code. For other generated code that need to use _extensions, add a getExtensionSet() in IExtensionMessage or more methods.

csharp/src/Google.Protobuf/IExtensionMessage.cs
```diff
@@ -0,0 +1,78 @@
...
+#endregion
+
+using Google.Protobuf.Collections;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Interface for a Protocol Buffers message, supporting basic operations for serialization as well as registering extension field values
+    /// </summary>
+    public interface IExtensionMessage : IMessage
```
anandolee: rename to IExtendableMessage

csharp/src/Google.Protobuf/ExtensionSet.cs
ObsidianMinor: A call to RegisterExtension would be required to create and use the field

src/google/protobuf/compiler/csharp/csharp_message.cc
ObsidianMinor: I actually experimented with this at one point. I decided against it at the time since it was difficult to do and didn't really payoff, there were too many choices with how to implement each with their own costs. I'd prefer to keep it simple with interfaces even if it has a little code duplication.

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,341 @@
...
+
+        /// <summary>
+        /// Clears the value of the specified extension
+        /// </summary>
+        public static void Clear<TTarget, TValue>(ref ExtensionSet<TTarget> set, Extension<TTarget, TValue> extension) where TTarget : IExtensionMessage<TTarget>
+        {
+            IExtensionValue value;
+            if (GetValue(ref set, extension, out value))
+            {
+                ((ExtensionValue<TValue>)value).ClearValue();
```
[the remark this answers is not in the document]
ObsidianMinor: That's true, it'd probably save some memory as well.

csharp/src/Google.Protobuf/Extension.cs
```diff
@@ -0,0 +1,109 @@
...
+            return new ExtensionValue<TValue>(codec);
+        }
+    }
+
+    /// <summary>
+    /// Represents a type-safe extension identifier used for getting repeated extension values in <see cref="IExtensionMessage{T}"/> instances
+    /// </summary>
+    /// <typeparam name="TTarget">The message type this field applies to</typeparam>
+    /// <typeparam name="TValue">The repeated field value type of this extension</typeparam>
+    public sealed class RepeatedExtension<TTarget, TValue> : Extension where TTarget : IExtensionMessage<TTarget>
```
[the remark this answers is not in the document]
ObsidianMinor: I'm not sure if that'd be safe. I don't understand the full scope of your API but it sounds like it lacks the type safety that I had in mind with my implementation. 

I'd prefer to use the current implementation since it has a couple of advantages over extension identifiers in other language implementations; Specifically because the type itself is self documenting. 

You can't accidentally call the wrong method with it, you don't need to check the original proto file to see if it's repeated or not because you automatically know if it's repeated or not based on the type provided to the method.

If a consumer lacks type information (if you're using the base Extension type) we could add an IsRepeated method to Extension though it'd just be some simple runtime reflection.

```cs
public bool IsRepeated()
{
    return GetType().GetGenericTypeDefinition() == typeof(RepeatedField<,>);
}
```

csharp/src/Google.Protobuf/ExtensionSet.cs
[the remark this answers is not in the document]
ObsidianMinor: Should we also allow clearing repeated extensions to remove their values from sets?

csharp/src/Google.Protobuf/Extension.cs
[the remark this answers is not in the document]
anandolee: The usage of Extension is somehow like FieldDescriptor.  We do not separate into FieldDescriptor and RepeatedFieldDescriptor, thus better not to separate Extension as well. 

Plus C++ and Java  do not have RepeatedExtension, try best to keep different languages similar will benefit project that use multi languages for protobuf.
https://github.com/protocolbuffers/protobuf/blob/46a48e49aa8357bbeee8040819a35e59880e329a/java/core/src/main/java/com/google/protobuf/ExtensionLite.java#L48

csharp/src/Google.Protobuf/ExtensionSet.cs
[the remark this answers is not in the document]
anandolee: Yes, I think clean repeated should also be allowed.

src/google/protobuf/compiler/csharp/csharp_message.cc
anandolee: acknowledged

csharp/src/Google.Protobuf/ExtensionSet.cs
anandolee: Both 1 and 2 make sense to me. I prefer 1

ObsidianMinor: I'm actually starting to prefer option two with the "GetOrRegister" method, given how gRPC is getting integrated with ASP.NET Core and with their focus on low memory and speed, it'd make sense to try and keep allocations low.

csharp/src/Google.Protobuf/Extension.cs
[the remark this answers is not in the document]
ObsidianMinor: Right, but C++ and Java have language features that allow them to easily use one extension type safely, or they don't have their own APIs that can be reused effectively, *or* the extension APIs were designed before those features were created (they are old libraries after all). C# either doesn't have these features or has other APIs I could leverage for code reuse, making it easier to understand and maintain from a consumer side and developer side.
 
C# has FieldCodec and RepeatedField. If we want to expose a RepeatedField we have to use a FieldCodec to leverage the APIs in it to write, merge, and calculate the size of a field. FieldCodec is also capable of being used for single fields which means we can use it in single value fields. If we got rid of the RepeatedExtension type, I couldn't reuse FieldCodec and RepeatedField and the API would have to match C++ and Java which *doesn't match* other C# repeated field APIs.

Plus, docs for extensions in C++ and Java state

> (The exact implementation of extension identifiers is complicated and involves magical use of generics / templates – however, you don't need to worry about how extension identifiers work to use them.)

I'm still under the firm belief that having a separate type isn't going to matter that much since they're magic in other languages anyway and most consumers will exclusively be using them via the `IExtendableMessage` methods and the custom options accessors. So they probably won't realize they're separate types anyway.

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,357 @@
...
+            }
+            else
+            {
+                IExtensionValue extensionValue;
+                if (set.ValuesByNumber.TryGetValue(lastFieldNumber, out extensionValue))
+                {
+                    extensionValue.MergeFrom(stream);
+                    return true;
+                }
+                else if (stream.ExtensionRegistry != null && stream.ExtensionRegistry.ContainsInputField(stream, typeof(TTarget), out extension))
```
anandolee: Merge the code with line 195 to simplify

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
```diff
@@ -240,6 +243,11 @@ private static IList<FileDescriptor> DeterminePublicDependencies(FileDescriptor
         /// </value>
         public IList<ServiceDescriptor> Services { get; }
 
+        /// <summary>
+        /// Unmodifiable list of top-level extensions declared in this file.
+        /// </summary>
+        public ExtensionCollection Extensions { get; }
```
anandolee: Can't  use IList<FieldDescriptor>?

ObsidianMinor: We could, however with it fashioned in this way we can provide methods to get fields defined for specific messages in declaration or number order (which I believe will be a common use-case like in FieldDescriptor). If consumers want extensions in any order not separated by extended type, they can get `Extensions.UnorderedExtensions` (which we can rename to just `Unordered` if it's too repetitive).

ObsidianMinor:
Is there any way I can help speed up this review @anandolee? It's taking a really long time. So long in fact that at this point we've almost hit the one year anniversary of opening the first feature PR.

anandolee:
Oh sorry... there are two many emails from github which makes it easy to be missed. I want to speed up it as well. Are you happy to ping or email me at jieluo@google.com when there's new changes or anything need my attension.

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
anandolee: ExtesnionCollection make sense as we also have FieldCollection.

ObsidianMinor:
I think you also need to add a "release notes: no" label to pass Mergable's checks @anandolee

csharp/src/Google.Protobuf/FieldCodec.cs
```diff
@@ -59,8 +60,9 @@ public static FieldCodec<string> ForString(uint tag)
         /// Retrieves a codec suitable for a bytes field with the given tag.
         /// </summary>
         /// <param name="tag">The tag.</param>
+        /// <param name="defaultValue">The default value.</param>
         /// <returns>A codec for the given tag.</returns>
-        public static FieldCodec<ByteString> ForBytes(uint tag)
+        public static FieldCodec<ByteString> ForBytes(uint tag, ByteString defaultValue = null)
```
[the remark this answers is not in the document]
ObsidianMinor: Right, this will be transitively fixed by #6380

csharp/src/Google.Protobuf/Extension.cs
```diff
@@ -0,0 +1,111 @@
...
+#endregion
+
+using System;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Represents a non-generic extension definition
+    /// </summary>
+    public abstract class Extension
```
jtattermusch: it might be a good idea to mark the classes of Extension API (Extension, RepeatedExtension, IExtensionValue, ExtensionSet...) as experimental and subject to change (add note to summary) - we might have to do some adjustments once first users try it and they discover some usability/functionality issues.
We can then remove the experimental note in the future once things are settled (e.g. in 3.10 or 3.11 release)

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,341 @@
...
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Methods for managing <see cref="ExtensionSet{TTarget}"/>s with null checking.
+    /// 
+    /// Most users will not use this class directly
+    /// </summary>
+    public static class ExtensionSet
+    {
+        private static bool GetValue<TTarget>(ref ExtensionSet<TTarget> set, Extension extension, out IExtensionValue value) where TTarget : IExtendableMessage<TTarget>
```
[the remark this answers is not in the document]
ObsidianMinor: Each set in the generated message classes are null by default (like `UnknownFieldSet`). The methods are ref so that rather than returning a new `ExtensionSet` like the `UnknownFieldSet` methods, it just takes them by-ref and assigns a new instance if necessary. Some methods could be made to not take the set by-ref but I decided to make them all take it `by-ref` for consistency and in case by some outstanding reason we need to reassign the sets in get methods (though I'm doubtful that'll happen).

csharp/src/Google.Protobuf/ExtensionRegistry.cs
```diff
@@ -0,0 +1,175 @@
...
+            ProtoPreconditions.CheckNotNull(extension, nameof(extension));
+
+            extensions.Add(new ObjectIntPair<Type>(extension.TargetType, extension.FieldNumber), extension);
+        }
+
+        /// <summary>
+        /// Adds the specified extensions to the registry
+        /// </summary>
+        public void Add(params Extension[] newExtensions)
```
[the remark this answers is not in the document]
ObsidianMinor: It's just a convenience method so you can `Add(Extension1, Extension2, Extension3)` instead of `Add(new[] { Extension1, Extension2, Extension3 })`.

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,341 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Methods for managing <see cref="ExtensionSet{TTarget}"/>s with null checking.
+    /// 
+    /// Most users will not use this class directly
```
jtattermusch: Does this read "the class is intended to be used by the generated code, users should not use it directly"?

ObsidianMinor:
> (or is the API completely useless right now and it's going to be only start being usable with #5936?)

Correct, you can't generate any code that uses Extension or ExtensionSet with this PR.

> I'm still reviewing the Extension, ExtensionSet, ExtensionValue, ... APIs to make sure they are making sense. What is the best place to look at to see the intended usage of extensions in C#?

There's many tests you can browse to see how to use extensions (ExtensionSet and ExtensionValue are not exposed).

 * [required fields in extension values](https://github.com/ObsidianMinor/protobuf/blob/csharp/proto2-feature/finale/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs#L265)
 * [required fields in groups](https://github.com/ObsidianMinor/protobuf/blob/csharp/proto2-feature/finale/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs#L341)
 * [full instance of `TestAllExtensions`](https://github.com/ObsidianMinor/protobuf/blob/csharp/proto2-feature/finale/csharp/src/Google.Protobuf.Test/SampleMessages.cs#L155)
 * [`CustomOptions` tests were redone to make sure the behavior matched the previous implementation (since with this PR it now runs through `ExtensionSet`) and make sure the values match how you're supposed to access them with #5936 via the GetOption accessor in each *Descriptor class](https://github.com/ObsidianMinor/protobuf/blob/csharp/proto2-feature/finale/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs)

csharp/src/Google.Protobuf/ExtensionRegistry.cs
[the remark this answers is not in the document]
jtattermusch: I see it's a syntactic sugar, I'm just wondering how useful it is. What are some common situations where being able to call Add(ext1, ext2, ext3) would come handy?

csharp/src/Google.Protobuf/IExtendableMessage.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Gets the value of the specified repeated extension or null if the extension isn't registered in this set.
+        /// For a version of this method that never returns null, use <see cref="IExtendableMessage{T}.GetOrRegisterExtension{TValue}(RepeatedExtension{T, TValue})"/>
+        /// </summary>
+        RepeatedField<TValue> GetExtension<TValue>(RepeatedExtension<T, TValue> extension);
+
+        /// <summary>
+        /// Gets the value of the specified repeated extension, registering it if it isn't
+        /// </summary>
+        RepeatedField<TValue> GetOrRegisterExtension<TValue>(RepeatedExtension<T, TValue> extension);
```
jtattermusch: it seems odd that we have a GetOrRegisterExtension for RepeatedExtension but we don't have it for a non-repeated Extension. What is the reason?

csharp/src/Google.Protobuf/ExtensionSet.cs
jtattermusch: nit: the method has a "TryGetValue" semantics (returns bool and out parameter), so perhaps it should be called TryGetValue for better readability?

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,341 @@
...
+            else
+            {
+                if (!set.ValuesByNumber.TryGetValue(extension.FieldNumber, out value))
+                {
+                    value = extension.CreateValue();
+                    set.ValuesByNumber.Add(extension.FieldNumber, value);
+                }
+            }
+
+            return ((RepeatedExtensionValue<TValue>)value).GetValue();
```
jtattermusch: needing to do this cast doesn't seem right.  The ExtensionValue classes are internal, so you can have methods that will allow you to avoid dirty tricks like this:
E.g. you can have two virtual methods/properties:  `GetSingleValue` and `GetRepeatedValue`  where GetSingleValue just throws InvalidOperationException for repeated extension and vice versa. The cast you're doing is basically a dirty way of achieving the same effect (and you also need to cast in multiple places in ExtensionSet instead of defining the behavior just once).

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -0,0 +1,341 @@
...
+            if (first == null)
+            {
+                first = new ExtensionSet<TTarget>();
+            }
+            foreach (var pair in second.ValuesByNumber)
+            {
+                IExtensionValue value;
+                if (first.ValuesByNumber.TryGetValue(pair.Key, out value))
+                {
+                    value.MergeFrom(pair.Value);
```
jtattermusch: if the extension value happens to be of the wrong kind (ExtensionValue vs RepeatedExtensionValue), extensionValue.MergeFrom(stream) silently ignores the value (there's a "is" check). Is that intended?

csharp/src/Google.Protobuf/IExtendableMessage.cs
ObsidianMinor: `GetExtension` will get the default value of the field if it isn't present in the set and doesn't register the field. Repeated fields don't have a default, so they return `null`. But `null` isn't normally expected for a `RepeatedField` value, so you could make it so that `GetExtension` for repeated fields returns non-`null`, but could cause allocations if the value isn't used. So to allow for the most control, `GetExtension` won't allocate and returns `null`, `GetOrRegisterExtension` will allocate and returns a normal value, and `ClearExtension` clears the field, making `GetExtension` return `null` again.

csharp/src/Google.Protobuf/ExtensionRegistry.cs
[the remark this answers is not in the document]
ObsidianMinor: I suppose most wouldn't use it since most of the time people will build a registry via list initialization syntax and leave it at that. I'll remove it the overload.

csharp/src/Google.Protobuf/ExtensionSet.cs
ObsidianMinor: Yes, though I worded it like this to match up with UnknownFieldSet.

csharp/src/Google.Protobuf/ExtensionSet.cs
ObsidianMinor: I don't think I could do that with `IExtensionValue` since it's not generic. `GetSingleValue` and `GetRepeatedValue` on `IExtensionValue` would have to return an object instead of an actual generic value.

jtattermusch: Ok, got it. I think it would still be cleaner to have e.g. a `AsRepeatedExtension<TValue>()` method on IExtensionValue which would make the cast with the option to perform some extra checking. But after looking at the code, I cannot think of a failure scenario where the extension value would be of a wrong type (because we already know it's an extension targeting TTarget and we already checked the field number), so I think it's fine to leave the logic as is.
Thanks for the explanation.

csharp/src/Google.Protobuf/ExtensionSet.cs
jtattermusch: not sure. @anandolee would know better?

csharp/src/Google.Protobuf/IExtendableMessage.cs
jtattermusch: It the term "register" in the meaning of "set the repeated extension to the default empty value" a name that you came up with or do other languages use similar terminology? My understanding of "register" is slightly different so this method's name seems a bit confusing to me. 

Idea:
How about adding an `RepeatedField<TValue> GetExtension<TValue>(RepeatedExtension<T, TValue> extension, bool initializeValue)` overload instead ?  `initializeIfNotPresent = false` would give you the GetExtension behavior and `initializeIfNotPresent = true` corresponds to GetOrRegisterExtension?

csharp/src/Google.Protobuf/IExtendableMessage.cs
```diff
@@ -0,0 +1,78 @@
...
+#endregion
+
+using Google.Protobuf.Collections;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// Generic interface for a Protocol Buffers message containing one or more extensions, where the type parameter is expected to be the same type as the implementation class
+    /// </summary>
+    public interface IExtendableMessage<T> : IMessage<T> where T : IExtendableMessage<T>
```
jtattermusch: This interface should also be marked as experimental for now.
The tricky part about this interface is that even if we mark it as experimental, the generated code needs to implements the methods provided by this interface, so adding or removing a method could still be breaking for users.
It is therefore important to get this interface right. On the other hand, until we actually enable proto2 codegen, 
no code is really referencing this interface, so I guess we are good for now (until https://github.com/protocolbuffers/protobuf/pull/5936 gets merged)

csharp/src/Google.Protobuf/IExtendableMessage.cs
ObsidianMinor: Generally other languages use `Register` to mean they set up the extension to be tracked by either adding them to a registry or some other means. In C++ it adds them to the global registry, in Java you use it to add them to a registry. So I took it to mean that the extension was "tracked" and would exist in the set if you accessed it. But I guess since that definition isn't universal.

I'm open to renaming the method to something easier to understand from a brief glance. I don't think `GetExtension(RepeatedInt32Extension, true)` would be that easy to read from an outside perspective. Perhaps something like `Insert` or `Initialize` or some other word over `Register`.

jtattermusch: @ObsidianMinor  thanks for the explanation, but I think that transition from the original meaning of "Register" in other languages to the way you used it should be avoided to maintain consistency. In the context of this PR, the "Register" also sounds like it'd have to do with ExtensionRegistry ("registering the fact that an extension was declared for a message") while in fact it doesn't (it has to do with whether the value of an extension was set or not).

I think `GetOrInitializeExtension` is a definitely better than GetOrRegisterExtension, but it's still not perfect IMHO

PR #4304, opened 2018-02-11
Ruby: Allows decode_json to optionally ignore unknown fields

ebenoist:
Currently the json parser raises a ParseError in the case that an
unexpected field is present. To better preserve forwards and backwards
compatible changes, an option to ignore these unknown fields can be
passed to decode_json. This fixes a long standing TODO by @haberman 
and adopts functionality already present in the json parsers for Go, C++, 
and Java.

Fixes #4251

I've noticed that this same upb code is used for PHP and in a standalone library. Should I make changes to those projects in addition?

PR #5373, opened 2018-11-16
Enable the ignore_unknown_field option in the Ruby unmarshal options

ebenoist:
This synchronizes the latest changes in upb.c and upb.h to enable these options


ebenoist:
This should wrap up the changes discussed here: https://github.com/protocolbuffers/protobuf/pull/4304

ruby/tests/basic.rb
```diff
@@ -254,6 +254,15 @@ def test_map_encode_decode
                     "b" => TestMessage2.new(:foo => 2)}
     end
 
+    def test_protobuf_decode_json_ignore_unknown_fields
+      m = TestMessage.decode_json({ optional_string: "foo", not_in_message: "some_value" }.to_json, { ignore_unknown_fields: true })
+      assert_equal m.optional_string, "foo"
+      e = assert_raise Google::Protobuf::ParseError do
```
TeBoring: If ignore_unknown_fields is true, why is there still a parser error?
BTW, could you add tests for both ignore_unknown_fields is true and false?

TeBoring:
We recently added more changes to upb. Could you sync upb again?

ruby/tests/basic.rb
ebenoist: I'm not passing in the ignore_unknown_fields flag in this second assert. So this test covers both cases. Would it be better if that was split across two different tests?

ebenoist:
@TeBoring Going to the latest upb.h/upb.c breaks a number of things in the protobuf.c. Looks like there were some breaking changes made.

ebenoist:
@TeBoring How can I see the CI output? It keep telling me that the log was not found? I suspect I don't have access to something.

ebenoist:
Fixed up the error, everything passes locally now.

ebenoist:
Can we re-run CI for this one? It should be gtg

TeBoring:
The failure is due to the update to new upb has fixed some conformance tests. However, these tests haven't been removed from the failure list.
You could go to ```conformance``` dir and run ```make test_ruby``` and remove succeeded tests from ruby's failure list: https://github.com/protocolbuffers/protobuf/blob/master/conformance/failure_list_ruby.txt.

ebenoist:
I updated the list, but one was added to the list: `Required.Proto3.JsonInput.TimestampJsonInputTooSmall`. Is this a regression?

TeBoring:
You can just put that in the failure test. The conformance test previously may miss it.

TeBoring:
I see it passed in the continuous test. Maybe resync your repository?

ebenoist:
Rebased! I added it to the failure list, which I think is why it passed. It still passes after a rebase.

ebenoist:
Anything else I need to do here?

ruby/ext/google/protobuf_c/encode_decode.c
```diff
@@ -897,13 +897,29 @@ VALUE Message_decode(VALUE klass, VALUE data) {
  * format) under the interpretration given by this message class's definition
```
haberman: Could you document the new params here please? Also please add docs for the hash params to encode_json (this should have been done before).

TeBoring:
Some conformance tests seem to have regression.
Could you double check you have rebased and included latest upb changes?
If everything has been done, probably just add those regressed tests to ruby's conformance failure list.

ebenoist:
@TeBoring Can you tell me which conformance tests you're seeing that should be added to the list? This is _not_ the latest upb.c/upb.h, but rather what was being used for PHP in this repo. I could not get the latest to compile.

ruby/ext/google/protobuf_c/encode_decode.c
ebenoist: Absolutely

TeBoring:
Required.Proto3.JsonInput.TimestampJsonInputTooSmall
  Required.Proto3.JsonInput.TimestampMinValue.JsonOutput
  Required.Proto3.JsonInput.TimestampMinValue.ProtobufOutput
  Required.Proto3.JsonInput.TimestampRepeatedValue.JsonOutput
  Required.Proto3.JsonInput.TimestampRepeatedValue.ProtobufOutput

TeBoring:
What are the result you rerun locally?
It seems the conformance test ask you to remove these tests from failing list...

ruby/ext/google/protobuf_c/encode_decode.c
haberman: I wonder if it would be a little easier to follow if it was phrased more like this:

- ignore_unknown_fields: set true to ignore unknown fields (default is to raise an error)
- preserve_proto_fieldnames: set true to use original fieldnames (default is to camelCase)
- emit_defaults: set true to emit 0/false values (default is to omit them)

ebenoist:
```
make test_ruby
RUBYLIB=../ruby/lib:. ./conformance-test-runner --enforce_recommended --failure_list failure_list_ruby.txt ./conformance_ruby.rb
[libprotobuf ERROR google/protobuf/wire_format_lite.cc:612] String field 'protobuf_test_messages.proto3.TestAllTypesProto3.optional_string' contains invalid UTF-8 data when parsing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.

CONFORMANCE TEST BEGIN ====================================

CONFORMANCE SUITE PASSED: 536 successes, 194 skipped, 93 expected failures, 0 unexpected failures.
```

Happy to make that file do the right thing, but I'm not seeing this locally unfortunately, which is making it hard to debug.

ebenoist:
Originally these failed for me locally

```
Required.Proto3.JsonInput.TimestampJsonInputTooSmall
Required.Proto3.JsonInput.TimestampMinValue.JsonOutput
Required.Proto3.JsonInput.TimestampMinValue.ProtobufOutput
Required.Proto3.JsonInput.TimestampRepeatedValue.JsonOutput
Required.Proto3.JsonInput.TimestampRepeatedValue.ProtobufOutput
```

The script suggested adding them to the list. Should I remove them?

ebenoist:
```diff
diff --git a/conformance/failure_list_ruby.txt b/conformance/failure_list_ruby.txt
index 89270326..44f67a92 100644
--- a/conformance/failure_list_ruby.txt
+++ b/conformance/failure_list_ruby.txt
@@ -73,11 +73,6 @@ Required.Proto3.JsonInput.RepeatedUint32Wrapper.JsonOutput
 Required.Proto3.JsonInput.RepeatedUint64Wrapper.JsonOutput
 Required.Proto3.JsonInput.StringFieldSurrogatePair.JsonOutput
 Required.Proto3.JsonInput.StringFieldSurrogatePair.ProtobufOutput
-Required.Proto3.JsonInput.TimestampJsonInputTooSmall
-Required.Proto3.JsonInput.TimestampMinValue.JsonOutput
-Required.Proto3.JsonInput.TimestampMinValue.ProtobufOutput
-Required.Proto3.JsonInput.TimestampRepeatedValue.JsonOutput
-Required.Proto3.JsonInput.TimestampRepeatedValue.ProtobufOutput
 Required.Proto3.ProtobufInput.DoubleFieldNormalizeQuietNan.JsonOutput
 Required.Proto3.ProtobufInput.DoubleFieldNormalizeSignalingNan.JsonOutput
 Required.Proto3.ProtobufInput.FloatFieldNormalizeQuietNan.JsonOutput
```

Should I apply this patch?

ruby/ext/google/protobuf_c/encode_decode.c
ebenoist: totally works for me

ebenoist: Updated. Thanks for the feedback.

haberman:
I say we merge if the tests are passing in Kokoro. That's odd that you're getting different results locally. We should look into this further.

ebenoist:
How do we get this back in? I'm happy to help. Just need a little direction.

lucas-ibotta:
What is the status of this? It seems to have been merged, reverted and closed? Has this feature been released into a version yet? Thanks.

TeBoring:
I think it’s in 3.7.0

PR #5379, opened 2018-11-19
ObjC: small improvement to extension serialization.

thomasvl:
For messages that have multiple extension ranges, this will improve things
by avoiding repeated work.  For messages with a single range, it should
be a wash.

- Sort the list of set extensions once during serialization and reuse the list.
- Break out of the serialization loop as soon as the loop has moved pasted at
  accepted range for field ids.

### DIFF
--- a/objectivec/GPBMessage.m
+++ b/objectivec/GPBMessage.m
@@ -1283,16 +1283,20 @@ static GPBUnknownFieldSet *GetOrMakeUnknownFields(GPBMessage *self) {
   NSUInteger fieldCount = fieldsArray.count;
   const GPBExtensionRange *extensionRanges = descriptor.extensionRanges;
   NSUInteger extensionRangesCount = descriptor.extensionRangesCount;
+  NSArray *sortedExtensions =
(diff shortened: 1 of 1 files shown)


thomasvl:
@dmaclach @sergiocampama

objectivec/GPBMessage.m
```diff
@@ -1808,17 +1812,20 @@ - (NSArray *)extensionsCurrentlySet {
 }
 
 - (void)writeExtensionsToCodedOutputStream:(GPBCodedOutputStream *)output
-                                     range:(GPBExtensionRange)range {
-  NSArray *sortedExtensions = [[extensionMap_ allKeys]
-      sortedArrayUsingSelector:@selector(compareByFieldNumber:)];
+                                     range:(GPBExtensionRange)range
+                                extensions: sortedExtensions {
```
dmaclach: maybe the name should be "sortedExtensions", and you forgot to add a type (I'm guessing Objc is defaulting to id, although I had never realized this was potentially valid syntax).

thomasvl: Done

Issue #5345, opened 2018-11-13
Minor Typo in Message.php and maybe others


Allmoz:
Hi, where it says:
@throws Exception Invalid data.

it should be :
@throws \Exception Invalid data.

or perhaps:
@throws GPBDecodeException Invalid data.

:) keep the great work

TeBoring:
Could you help send a fix to us?

Allmoz:
ok, i will try

Allmoz:
Done :) i found some ErrorException too.
https://github.com/protocolbuffers/protobuf/pull/5381

PR #5381, opened 2018-11-20
Fix Issue #5345

Allmoz:
Added a few \ on the comments to match the exceptions scope when is referenced while in a namespace. So the IDEs can be happy.

I think i got them all.

Best regards!

### DIFF
--- a/php/src/Google/Protobuf/Any.php
+++ b/php/src/Google/Protobuf/Any.php
@@ -252,9 +252,9 @@ class Any extends \Google\Protobuf\Internal\Message
      * the method will create a message of the targeted type and fill it with
      * the decoded value in Any.
      * @return unpacked message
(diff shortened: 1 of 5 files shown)


Allmoz:
I signed it!

TeBoring:
Could you also update the code generator in src/google/protobuf/compiler/php

Allmoz:
you mean adding the "* @throws \Exception" to the setters?

TeBoring:
I mean, for whatever changes to make in php code, could you add the logic to our code generator? So that, next time, we won't have regression.

Allmoz:
Ok, but for the changes i made i did not find the compiler routines that made it (if it was made by it)

TeBoring:
Ah, you are right. Code you modified were made manually.

PR #5389, opened 2018-11-21
Add Bazel config for zlib support

schroederc:
Allow downstream Bazel clients to build protobuf library with zlib support.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -16,6 +16,19 @@ config_setting(
 )
 
 ################################################################################
+# ZLIB configuration
+################################################################################
+
+config_setting(
+    name = "zlib",
+    values = {"define": "protobuf_zlib=true"},
+)
+
+ZLIB_COPTS = ["-DHAVE_ZLIB"]
+
+ZLIB_DEPS = ["//external:zlib"]
+
+################################################################################
 # Protobuf Runtime Library
 ################################################################################
 
@@ -50,6 +63,9 @@ COPTS = select({
         "-Wno-writable-strings",
         "-Wno-write-strings",
     ],
+}) + select({
+    ":zlib": ZLIB_COPTS,
+    "//conditions:default": [],
 })
 
 load(":compiler_config_setting.bzl", "create_compiler_config_setting")
@@ -117,6 +133,11 @@ cc_library(
     visibility = ["//visibility:public"],
 )
 
+PROTOBUF_DEPS = select({
+    ":zlib": ZLIB_DEPS,
+    "//conditions:default": [],
+})
+
 cc_library(
     name = "protobuf",
     srcs = [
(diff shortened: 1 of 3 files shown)


acozzette:
@schroederc I tried this out locally but got a build error. When I do `bazel build --define protobuf_zlib=true //:protobuf` I get an error saying `no such target '//external:zlib': target 'zlib' not declared in package 'external'` Is that error to be expected?

schroederc:
It's expected on the protobuf side because it does not bind the `zlib` target in its `WORKSPACE`.  It works if you add zlib as a dependency in your WORKSPACE (https://github.com/kythe/kythe/blob/master/external.bzl#L18) and then bind that target (https://github.com/kythe/kythe/blob/master/external.bzl#L606).

It seemed like the protobuf build doesn't care about zlib support so I didn't add it, but downstream builds can enable it if they have zlib.

acozzette:
Ah OK, thanks for the explanation. Is there any way we can just unconditionally depend on `zlib`? I think if possible it would be nice to avoid the need for a new build flag here.

schroederc:
Yes, you'd need to add the zlib repo to the `WORKSPACE` file as mentioned above.  This will have the side-effect of requiring all downstream builds to also add that repository once they update protobuf.  I can do that, if that's wanted instead.

acozzette:
OK, I see. That's too bad that adding the dependency unconditionally would break downstream builds. Is there a best practice in Bazel for how to add a new dependency to a library without disrupting downstream projects? Your approach of using a flag should work, but still it would be nice if we could avoid adding a new flag.

schroederc:
I don't know of a way to add the dependency to all downstream projects automagically.  Many projects use a custom [Starlark function](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#go-rules-dependencies) that downstream can call to add necessary deps to their WORKSPACE.  The upstream project can then curate the list of deps without breaking builds.  However, current downstream projects using protobuf do not already call such a function.

acozzette:
@schroederc OK, thanks for the info. I'd say for now let's go with omitting the flag and making the dependency unconditional. Later on we can add a Starlark function like the one you described if we need to.

BUILD
```diff
@@ -15,6 +15,17 @@ config_setting(
     },
 )
 
+################################################################################
+# ZLIB configuration
+################################################################################
+
+config_setting(
```
acozzette: We no longer need this new config_setting, right?

schroederc: Right.

acozzette:
Thanks @schroederc!

ronshapiro:
Is `@bazel_tools//third_party/zlib` an option here? That way there's no additional work for users

schroederc:
Any Bazel repository depending on protobuf could put this in their WORKSPACE to have protobuf compile against the Bazel version of zlib:

```
bind(
    name = "zlib",
    actual = "@bazel_tools//third_party/zlib",
)
```

It's a different question to pin all builds of protobuf to the Bazel version of zlib (which is not under control of protobuf).

PR #5391, opened 2018-11-21
Make sure Objective C Proto compiler doesn't "duplicate" prefixes

dmaclach:
In some cases proto files that want/need to use the objc_class_prefix option have
types that already have the prefix on a subset of their names. In this case we don't
want to duplicate the prefix.

Added tests for this (and prefixes in general).

### DIFF
--- a/objectivec/Tests/unittest_objc_options.proto
+++ b/objectivec/Tests/unittest_objc_options.proto
@@ -1,0 +1,61 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto2";
+
+package protobuf_objc_unittest;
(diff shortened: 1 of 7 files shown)


src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -243,21 +243,29 @@ bool IsReservedCIdentifier(const string& input) {
...
   static const std::unordered_set<string> kReservedWords =
       MakeWordsMap(kReservedWordList, GOOGLE_ARRAYSIZE(kReservedWordList));
   static const std::unordered_set<string> kNSObjectMethods =
       MakeWordsMap(kNSObjectMethodsList, GOOGLE_ARRAYSIZE(kNSObjectMethodsList));
-  if (IsReservedCIdentifier(input) ||
-      (kReservedWords.count(input) > 0) ||
-      (kNSObjectMethods.count(input) > 0)) {
+  string sanitized;
+  // If the string already has a prefix, don't add it again.
+  if (prefix.length() && !HasPrefixString(input, prefix)) {
```
thomasvl: So there is some risk here because a prefix of "For" with match a message of "Form" even though it clearly isn't a prefix.  To be safe, you probably want to ensure the `prefix != input` and that `input[prefix.length()]` is uppercase.

I guess the on part to debate is `prefix != input`.  As that does lead to things like _FooFoo_, but that does seems like it might be safer compared to leaving potentially generic class names like "Foo" with no way to avoid the proto class colliding with application code.

src/google/protobuf/descriptor.proto
```diff
@@ -396,7 +396,9 @@ message FileOptions {
 
 
   // Sets the objective c class prefix which is prepended to all objective c
-  // generated classes from this .proto. There is no default.
+  // generated types from this .proto. Note that if the type already has
```
thomasvl: Not sure it is worth this tweak pending what the final decision is about prefix == input cases.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -416,12 +424,11 @@ string FilePathBasename(const FileDescriptor* file) {
 }
 
 string FileClassName(const FileDescriptor* file) {
-  string name = FileClassPrefix(file);
-  name += UnderscoresToCamelCase(StripProto(BaseFileName(file)), true);
-  name += "Root";
+  const string &prefix = FileClassPrefix(file);
```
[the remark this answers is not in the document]
dmaclach: removed

objectivec/Tests/unittest_objc_options.proto
```diff
@@ -0,0 +1,61 @@
+// Protocol Buffers - Google's data interchange format
```
thomasvl: Done forget to add this to the Makefile.am so it is known to the outer build system

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -243,21 +243,35 @@ bool IsReservedCIdentifier(const string& input) {
...
   static const std::unordered_set<string> kReservedWords =
       MakeWordsMap(kReservedWordList, GOOGLE_ARRAYSIZE(kReservedWordList));
   static const std::unordered_set<string> kNSObjectMethods =
       MakeWordsMap(kNSObjectMethodsList, GOOGLE_ARRAYSIZE(kNSObjectMethodsList));
-  if (IsReservedCIdentifier(input) ||
-      (kReservedWords.count(input) > 0) ||
-      (kNSObjectMethods.count(input) > 0)) {
+  string sanitized;
+  // If the string already has a prefix, don't add it again.
+  printf("HooHah!\n");
```
thomasvl: oops

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dmaclach: From the examples that I ran into it seemed a lot better to clean up the duplicates when prefix == input. It made the code more readable.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -243,21 +243,36 @@ bool IsReservedCIdentifier(const string& input) {
...
                            string* out_suffix_added) {
   static const std::unordered_set<string> kReservedWords =
       MakeWordsMap(kReservedWordList, GOOGLE_ARRAYSIZE(kReservedWordList));
   static const std::unordered_set<string> kNSObjectMethods =
       MakeWordsMap(kNSObjectMethodsList, GOOGLE_ARRAYSIZE(kNSObjectMethodsList));
-  if (IsReservedCIdentifier(input) ||
-      (kReservedWords.count(input) > 0) ||
-      (kNSObjectMethods.count(input) > 0)) {
+  string sanitized;
+  // If the string already has a prefix, don't add it again.
```
thomasvl: Can you fix up these comments since we do add it in some cases?

You can also collapse the first two into:

```
if (input.length() == prefix.length() || !ascii_isupper(input[prefix.length()])) {
  sanitized = prefix + input;
} else {
  sanitized = input;
}
```

unless documenting it is easier to keep them split.

PR #5397, opened 2018-11-26
Enable fromObject mathod in generated JS code

### LINKED ISSUE #1591: JS generator missing jspb.generate_from_object option
The comment for [jspb.Message.GENERATE_FROM_OBJECT](https://github.com/google/protobuf/blob/master/js/message.js#L151) states:

>  *     NOTE: By default no protos actually have a fromObject method. You need to
>  *     add the jspb.generate_from_object options to the proto definition to
>  *     activate the feature.

However, no such option is currently implemented in the open sourced version. I see the code which generates output for this option (`GenerateClassFromObject`) but it is invoked 0 times.

> ~/protobuf$ grep -r GenerateClassFromObject .
> ./src/google/protobuf/compiler/js/js_generator.cc:void Generator::GenerateClassFromObject(const GeneratorOptions& options,
> ./src/google/protobuf/compiler/js/js_generator.h:  void GenerateClassFromObject(const GeneratorOptions& options,

Can the jspb.generate_from_object option be open sourced?

shaxbee:
Fixes #1591


BSBandme:
For Bo, maybe need also to take care of [b/119763404](b/119763404)

TeBoring:
the toObject/fromObject have some issues. We are still talking whether they are good solutions internally.

yinzara:
Any update on this?

karel-3d:
this would be nice and fix us lot of trouble :/

sassman:
To: @TeBoring 

Can somebody just merge this one?? Seriously it's been around since 2018 and the issue it's fixing is from 2016.

acozzette:
The format returned by `toObject()` is kind of a weird, non-standard format, sort of like [proto3 JSON](https://developers.google.com/protocol-buffers/docs/proto3#json) but not quite. I'm not sure of the story behind it but it appears to be a historical mistake. The reason we hesitate to merge this is that it's not clear how we can maintain that as an additional wire format going forward. From what I understand it also causes code size bloat by forcing a lot of string names to be included when they would otherwise be unnecessary.

yinzara:
I can tell you that the lack of 'fromObject' is the reason we use ProtobufJS instead of this official library.  It's a really important feature and while I understand there are some issues,  not providing it even via an option isn't acceptable IMHO

LeeStephen:
Would really appreciate this being implemented. Recently switched over from protobufjs to this because of the generated typescript support, but having to manually write all the code to do what "fromObject" could be doing is wasting a lot of time. I was really hoping this would be the 'de-facto' approach, but I'm now reconsidering going back to protobufjs.

karelbilek:
what toObject emits is NOT proto3 json.

Anyway. If I needed this nowadays, I would just write my own protobuf compiler that emits code that converts from/to proto3 json. As Google is not interested to do so, obviously.

I have learned that writing protoc compilers/code generators is easier than it looks, once you understand FileDescriptorProto format and all that stuff. It takes about a month to learn, but then everyone looks at you like a wizard that can generate code out of protobuf :)

PR #5398, opened 2018-11-26
Avoid octal constants in strutil.cc

brian-peloton:
The code now matches what the comment above it actually generates.

Also tweak that code so it actually compiles for me, and to avoid
embedding /* and */ in comments because that's generally a reasonable
thing to trigger warnings.


acozzette:
Thanks @brian-peloton!

PR #4650, opened 2018-05-19
remove PACKAGE_NAME and REPOSITORY_NAME deprecated usage

ittaiz:
Hi,
We (rules_scala) use this repository as a bazel source repository.
We want to run with bazel 0.13.0 with `--all_incompatible_changes` and fail because of https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#package-name-is-a-function
This PR fixes it.
@dslomov should there be a process to make sure very root projects (like protobuf) support the above flag?
cc @johnynek

PR #5404, opened 2018-11-27
Cherry-picked fix from PR #4650 into 3.6.x branch

acozzette:


acozzette:
The CLA is OK because it was already verified when the commit was originally merged into master.

PR #5412, opened 2018-11-29
[objc] Provide more details/context in deprecation messages ⚠️

stephanemoore:
For deprecated fields, Include the full name of the deprecated field and the file containing the deprecated field in deprecation messages. For deprecated files, include the deprecated file in deprecation message. This additional context will help provide more details/context to help developers seek recommended alternatives to deprecated fields.

### DIFF
--- a/objectivec/GPBBootstrap.h
+++ b/objectivec/GPBBootstrap.h
@@ -92,10 +92,21 @@
 #endif
 
 /**
- * Attribute used for Objective-C proto interface deprecations.
+ * Attribute used for Objective-C proto interface deprecations without messages.
  **/
 #ifndef GPB_DEPRECATED
 #define GPB_DEPRECATED __attribute__((deprecated))
+#endif
+
+/**
+ * Attribute used for Objective-C proto interface deprecations with messages.
+ **/
+#ifndef GPB_DEPRECATED_MSG
+#if __has_extension(attribute_deprecated_with_message)
+#define GPB_DEPRECATED_MSG(msg) __attribute__((deprecated(msg)))
+#else
+#define GPB_DEPRECATED_MSG(msg) __attribute__((deprecated))
+#endif
 #endif
 
 // If property name starts with init we need to annotate it to get past ARC.
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.h
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.h
@@ -163,11 +163,22 @@ string GetOptionalDeprecatedAttribute(
   // The file is only passed when checking Messages & Enums, so those types
   // get tagged. At the moment, it doesn't seem to make sense to tag every
   // field or enum value with when the file is deprecated.
+  bool isFileLevelDeprecation = false;
   if (!isDeprecated && file) {
-    isDeprecated = file->options().deprecated();
+    isFileLevelDeprecation = file->options().deprecated();
+    isDeprecated = isFileLevelDeprecation;
   }
   if (isDeprecated) {
-    string result = "GPB_DEPRECATED";
+    string message;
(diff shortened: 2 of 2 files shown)


objectivec/GPBBootstrap.h
```diff
@@ -95,7 +95,11 @@
  * Attribute used for Objective-C proto interface deprecations.
  **/
 #ifndef GPB_DEPRECATED
-#define GPB_DEPRECATED __attribute__((deprecated))
```
thomasvl: Won't removing this be source breaking if someone had generated something with the pervious change that switched to it?  i.e. - I think you need to keep `GPB_DEPRECATED` and add a new `GPB_DEPRECATED_MSG(...)` and keep the old just for safety sake.

objectivec/GPBBootstrap.h
```diff
@@ -95,7 +95,11 @@
...
+#if __has_extension(attribute_deprecated_with_message)
+#define GPB_DEPRECATED(...) __attribute__((deprecated("" __VA_ARGS__)))
```
thomasvl: Why does this need to be varargs vs just take a single _msg_?

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
```diff
@@ -167,7 +167,15 @@ string GetOptionalDeprecatedAttribute(
     isDeprecated = file->options().deprecated();
   }
   if (isDeprecated) {
-    string result = "GPB_DEPRECATED";
+    string message = descriptor->full_name();
+    const FileDescriptor* sourceFile = descriptor->file();
+    if (sourceFile != nullptr) {
```
thomasvl: skip this, just build the string, if a descriptor really didn't have a file, the whole thing could come crashing down.  We wouldn't know the package, syntax version, etc.

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
```diff
@@ -167,7 +167,15 @@ string GetOptionalDeprecatedAttribute(
     isDeprecated = file->options().deprecated();
```
thomasvl: If you really want to point folks at the right thing, you could also track when the marking was found here (a file) vs. on a field, and adjust your messaging.

objectivec/GPBBootstrap.h
stephanemoore: Good catch (and sorry that I didn't think to support the message in my initial addition of `GPB_DEPRECATED`). Created a separate `GPB_DEPRECATED_MSG`.

objectivec/GPBBootstrap.h
thomasvl: I guess the catch would be if we ever wanted to support `[[gnu::deprecated]]` or `[[deprecated]]`.  Given it now is a `_MSG` form, lets go with one arg so no one were to think it was a printf style args…

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
stephanemoore: Good idea! Putting together the changes to make that happen.

stephanemoore: Completed and verified 👌

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
stephanemoore: Done.

objectivec/GPBBootstrap.h
stephanemoore: Done.

objectivec/GPBBootstrap.h
stephanemoore: Done.

PR #5406, opened 2018-11-27
Change MACOSX_DEPLOYMENT_TARGET to 10.9

TeBoring:
https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes

PR #5423, opened 2018-12-04
Cherry-pick fix for compatibility with Python 3.7

acozzette:
This cherry-picks the fix from PR #4862 so that we can include it in a 3.6.1.2 patch release.

### DIFF
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -56,7 +56,7 @@
   #endif
   #define PyString_AsStringAndSize(ob, charpp, sizep) \
     (PyUnicode_Check(ob)? \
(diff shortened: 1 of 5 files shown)


acozzette:
The CLA is OK because it was already verified when this fix was merged as part of PR #4862.

TeBoring:
Why #4940 still fail?

acozzette:
I'm not sure about #4940 but it looks like those tests are probably failing because of some kind of issue with the test setup and not necessarily a real problem with Python 3.7. The TensorFlow team requested this cherry-pick because they have users who want to upgrade to Python 3.7.

TeBoring:
To build python 3.7, you probably also needs #5406. Otherwise, the built wheel cannot be run on latest mac (usually python 3.7 is on the newest mac)

PR #5424, opened 2018-12-04
Cherry-picked fix from #5406

acozzette:


acozzette:
The CLA is OK because it was already verified when the change was originally merged as #5406.

PR #5433, opened 2018-12-06
Remove mention of Buffer in utils.js

davido:
This replays this [commit](f00e06c95bc117fb2ed0ca56c96041c93039f1fe) on 3.6.x branch.
Without this commit, tests are failing on closure compiler, because Buffer class is undefined.

### DIFF
--- a/js/binary/utils.js
+++ b/js/binary/utils.js
@@ -971,6 +971,10 @@ jspb.utils.byteSourceToUint8Array = function(data) {
     return /** @type {!Uint8Array} */(new Uint8Array(data));
   }
 
+  if (data.constructor === Buffer) {
+    return /** @type {!Uint8Array} */(new Uint8Array(data));
+  }
+
   if (data.constructor === Array) {
     data = /** @type {!Array<number>} */(data);
     return /** @type {!Uint8Array} */(new Uint8Array(data));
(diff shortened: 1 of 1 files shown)


js/binary/utils.js
```diff
@@ -971,10 +971,6 @@ jspb.utils.byteSourceToUint8Array = function(data) {
     return /** @type {!Uint8Array} */(new Uint8Array(data));
   }
 
-  if (data.constructor === Buffer) {
```
acozzette: I think we should go with the logic on master that looks like this:

```
  if (typeof Buffer != 'undefined' && data.constructor === Buffer) {
    return /** @type {!Uint8Array} */ (
        new Uint8Array(/** @type {?} */ (data)));
  }
```

That should do the right thing whether `Buffer` exists or not.

davido: Done.

davido:
@acozzette Thanks for merging. Can you help to tag 3.6.1.3 on the 3.6.x branch, so that we can consume it from rules_closure? Thanks!

acozzette:
@davido Sure, I went ahead and tagged v3.6.1.3 with this change.

davido:
@acozzette Thanks so much! Preparing upgrade of rules_closure with v.3.6.1.3.

PR #5434, opened 2018-12-06
Php 7.3 fix

### LINKED ISSUE #4933: Broken with PHP 7.3.0alpha4
Build ok since pr #4898 is merged but extension dosn't load (since 7.3.0alpha4, was OK with alpha3)

`$ php --no-php-ini --define extension=modules/protobuf.so --modules | grep protobuf`

Output is empty. No error message.

P.S. as I can't sign your CLA, I don't plan to work on this issue to propose a fix, so, do it yourself.

TeBoring:
#4933


TeBoring:
@haon4 could you help review the docker and continuous test changing.

TeBoring:
@michaelbausor could you help review the php change?

.gitignore
```diff
@@ -145,7 +145,9 @@ php/ext/google/protobuf/Makefile.objects
 php/ext/google/protobuf/acinclude.m4
 php/ext/google/protobuf/build/
 php/ext/google/protobuf/config.h
+php/ext/google/protobuf/config.h.in~
```
michaelbausor: The trailing `~` looks odd - is that meant to be there?

php/tests/encode_decode_test.php
```diff
@@ -31,6 +31,7 @@ public function testDecodeJsonSimple()
     {
         $m = new TestMessage();
         $m->mergeFromJsonString("{\"optionalInt32\":1}");
+        $this->assertTrue(true);
```
michaelbausor: Why is this necessary? What about asserting that `$m->getOptionalInt32()` equals `1`?

tests.sh
```diff
@@ -349,7 +340,7 @@ build_php5.5() {
 
 build_php5.5_c() {
   use_php 5.5
-  wget https://phar.phpunit.de/phpunit-4.8.0.phar -O /usr/bin/phpunit
+  ls /usr/local/php-5.5/bin
```
michaelbausor: Do we want to keep the `ls` call, or is this for debugging?

.gitignore
haon4: +1. Is this change expected?

kokoro/linux/32-bit/build.sh
```diff
@@ -10,8 +10,9 @@
...
 cd $(dirname $0)/../../..
 
-export DOCKERFILE_DIR=kokoro/linux/32-bit
+export DOCKERHUB_ORGANIZATION=protobuftesting
+export DOCKERFILE_DIR=kokoro/linux/dockerfile/test/php_32bit
 export DOCKER_RUN_SCRIPT=kokoro/linux/pull_request_in_docker.sh
 export OUTPUT_DIR=testoutput
 export TEST_SET="php_all_32"
-./kokoro/linux/build_and_run_docker.sh
+./kokoro/linux/build_and_run_docker2.sh
```
haon4: Is this necessary? How about we just update build_and_run_docker to set a default DOCKERHUB_ORGANIZATION only if it is not already set? Then we don't need two almost identical script.

kokoro/linux/dockerfile/test/php_32bit/Dockerfile
```diff
@@ -0,0 +1,224 @@
...
+  make \
+  parallel \
+  time \
+  wget \
+  && apt-get clean
+
+# Install php dependencies
+RUN apt-get clean && apt-get update && apt-get install -y --force-yes \
+  bison \
+  php5 \
```
haon4: should this be installed here if we can going to install php manually below?

kokoro/linux/dockerfile/test/php_32bit/Dockerfile
```diff
@@ -0,0 +1,224 @@
+FROM 32bit/debian:jessie
```
haon4: is it better to have all PHP versions setup in one file? or would it be preferable split them up into separate Dockerfiles?

.gitignore
TeBoring: Yes, I don't know why but compiling php7.3 code always have that.

kokoro/linux/dockerfile/test/php_32bit/Dockerfile
TeBoring: We have to install a php first in order to install composer later.

kokoro/linux/dockerfile/test/php_32bit/Dockerfile
TeBoring: As discussed on meeting, for php, we can share docker image as much as possible.
However, it's not possible to share docker image between 64 and 32 bit tests.

php/tests/encode_decode_test.php
TeBoring: Change this to assert equals 1.
For the other assertTrue, higher version of phpunit want to have at least one assertion. Otherwise, it will throw warnings.

tests.sh
TeBoring: Removed.

kibertoad:
I assume this is coming in 3.6.2?

PR #5442, opened 2018-12-10
fix: cmake install

ayounes-nviso:


ayounes-nviso:
I signed it!

acozzette:
Thanks @ayounes-nviso!

PR #5445, opened 2018-12-10
Make Protobuf compatible with C# 6

haon4:
Fix internal C# compilation issue.


ObsidianMinor:
You can also set the language version to C# 6 in each project (`<LangVersion>6</LangVersion>`) to prevent future instances of incompatibility.

haon4:
> You can also set the language version to C# 6 in each project (`<LangVersion>6</LangVersion>`) to prevent future instances of incompatibility.

Thanks for the suggestion! Done.

PR #5452, opened 2018-12-12
Updated Bazel workspace in examples/ to handle zlib dependency

acozzette:
To make this work, I created a symlink to the existing zlib.BUILD file.
Maybe there is a better way to do this, but this is at least a quick
workaround to fix the build.


acozzette:
OK, fixed the EXTRA_DIST issue.

PR #5453, opened 2018-12-12
Minimize amount of filepaths being copied into protos.

dmaclach:
By putting the NSAssert behind a DEBUG flag we aren't copying filepaths
into release builds.


thomasvl:
Hopefully folks don't normally do release builds with assertions on, but better safe than sorry.

PR #5454, opened 2018-12-12
Create a script to help update the version automatically

haon4:


src/google/protobuf/stubs/common.h
```diff
@@ -88,20 +88,20 @@ namespace internal {
 
 // The minimum library version which works with the current version of the
 // headers.
-#define GOOGLE_PROTOBUF_MIN_LIBRARY_VERSION 3006001
+#define GOOGLE_PROTOBUF_MIN_LIBRARY_VERSION GOOGLE_PROTOBUF_VERSION
```
TeBoring: https://github.com/protocolbuffers/protobuf/commit/050fc9a43717193ffbccbd8ca5d19e1e5648c815#diff-cbd31ea2a5761cd281396b4277c36887
I think they are not necessarily always the same.

src/Makefile.am
```diff
@@ -169,7 +171,7 @@ nobase_include_HEADERS =                                         \
 lib_LTLIBRARIES = libprotobuf-lite.la libprotobuf.la libprotoc.la
 
 libprotobuf_lite_la_LIBADD = $(PTHREAD_LIBS) $(LIBATOMIC_LIBS)
-libprotobuf_lite_la_LDFLAGS = -version-info 17:0:0 -export-dynamic -no-undefined
+libprotobuf_lite_la_LDFLAGS = -version-info $(PROTOBUF_VERSION):0:0 -export-dynamic -no-undefined
```
TeBoring: https://autotools.io/libtool/version.html

src/google/protobuf/stubs/common.h
TeBoring: For minor releases, we usually just update ```GOOGLE_PROTOBUF_VERSION```

src/Makefile.am
```diff
@@ -214,7 +216,7 @@ libprotobuf_lite_la_SOURCES =                                  \
   google/protobuf/io/zero_copy_stream_impl_lite.cc
 
 libprotobuf_la_LIBADD = $(PTHREAD_LIBS) $(LIBATOMIC_LIBS)
-libprotobuf_la_LDFLAGS = -version-info 17:0:0 -export-dynamic -no-undefined
+libprotobuf_la_LDFLAGS = -version-info $(PROTOBUF_VERSION):0:0 -export-dynamic -no-undefined
```
TeBoring: https://github.com/protocolbuffers/protobuf/commit/050fc9a43717193ffbccbd8ca5d19e1e5648c815#diff-480477e89f9b6ddafb30c4383dcdd705
For major release, we update the first (API change). For minor releases, we update the second (non-API change).

TeBoring: BTW, the last update to 17 in 3.6.1 release seems suspicious.

src/google/protobuf/stubs/common.h
haon4: The documentation seems to suggest that we should update all, but reverted this file, and let the script updates all of them by default. The author can change the version number manually if necessary.

src/Makefile.am
haon4: ok, I have updated the script to calculate this number from the given version. It should something like [MINOR_VERSION+OFFSET]:[MICRO_VERSION]:0, right?

src/google/protobuf/stubs/common.h
TeBoring: I think the document was written for major releases.
Is this script also used for major release only?

TeBoring: If it's minor release, I think we should keep things like ```GOOGLE_PROTOBUF_MIN_LIBRARY_VERSION``` unchanged. Because the minor release has no API change and the new library should work with the old header and the new header should also work with the old library.

haon4: The script simply calculate the version based on the given version number:
https://github.com/protocolbuffers/protobuf/pull/5454/commits/79748f318442c70cab077ac44da92c8bd84eb077. So if the given version is 3.6.1, it will generate 3006001.

Should I revise it such that only GOOGLE_PROTOBUF_VERSION would care about the micro version, and all other constants would still treat it like 3006000?

haon4: Done. Revised script to only update these constants on major release.

PR #5460, opened 2018-12-12
fix(php): fix php `serializeToJsonString` empty ListValue.

lightjiao:
PHP function `serializeToJsonString` when have an empty ListValue, it return an invalid json, just like this

# bug
```
{
    "name": "jack",
    "friends":,  // here should have an empty `[]`
    "age": 22
}
```

# after fix
```
{
    "name": "jack",
    "friends": [],
    "age": 22
}
```

### DIFF
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -1389,9 +1389,12 @@ class Message
             }
             $fields = $this->desc->getField();
             $first = true;
+            $empty = true;
             foreach ($fields as $field) {
-                if ($this->existField($field) ||
-                    GPBUtil::hasJsonValue($this)) {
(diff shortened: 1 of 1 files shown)


lightjiao:
Hi,
Is there any problem on this pr ?

TeBoring:
Could you check whether master branch has already fixed this problem?

TeBoring:
#5106 may have fixed this.

TeBoring:
If it's not fixed, could you add a test to encode_decode_test.php?

lightjiao:
ok,
thanks.

lightjiao:
I have test on master branch, the problem has already fixed, but there is no empty ListValue test case in `encode_decode_test.php`.  

I would add a test case.

thank you.

Issue #5448, opened 2018-12-11
version v3.6.1.3 fails to compile on Linux, version.rc issue


shardator:
**What version of protobuf and what language are you using?**
Version: v3.6.1.3
Language: C++

**What operating system (Linux, Windows, ...) and version?**

Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**

gcc version 5.4.0 20160609 (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.10)

**What did you do?**

Simply ran CMake build.

**What did you expect to see**

Build successful. :)

**What did you see instead?**

Build fail:

```
==> Processing plain cmake package: 'protobuf-dev'
==> Building with env: '/home/odroid/cscsatho/catkin_ws/install_isolated/env.sh'
==> cmake /home/odroid/cscsatho/catkin_ws/src/protobuf-dev -DCMAKE_INSTALL_PREFIX=/home/odroid/cscsatho/catkin_ws/install_isolated -DCMAKE_BUILD_TYPE=Release -G Ninja in '/home/odroid/cscsatho/catkin_ws/build_isolated/protobuf-dev/install'
-- Configuring done
CMake Error at cmake/protoc.cmake:9 (add_executable):
  Cannot find source file:

    /home/user/catkin_ws/build_isolated/protobuf-dev/install/cmake/version.rc

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
Call Stack (most recent call first):
  cmake/CMakeLists.txt:210 (include)


-- Generating done
-- Build files have been written to: /home/odroid/cscsatho/catkin_ws/build_isolated/protobuf-dev/install
<== Failed to process package 'protobuf-dev': 
  Command '['/home/odroid/cscsatho/catkin_ws/install_isolated/env.sh', 'cmake', '/home/odroid/cscsatho/catkin_ws/src/protobuf-dev', '-DCMAKE_INSTALL_PREFIX=/home/odroid/cscsatho/catkin_ws/install_isolated', '-DCMAKE_BUILD_TYPE=Release', '-G', 'Ninja']' returned non-zero exit status 1
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**NOTE**: I successfully fixed it by cherry-picking the commit `291d9a0ab0b452537d22a81b551eaea53fd157a8` (*Fix issue with version.rc*) from the `master` branch.

[...]

PR #5462, opened 2018-12-12
Fix issue with version.rc

shardator:
Based on commit for issue #5448:
```
commit 291d9a0ab0b452537d22a81b551eaea53fd157a8
Author: Ivan Shynkarenka <chronoxor@gmail.com>
Date:   Tue May 15 23:19:58 2018 +0300

    Fix issue with version.rc
```

### DIFF
--- a/cmake/libprotobuf-lite.cmake
+++ b/cmake/libprotobuf-lite.cmake
@@ -49,7 +49,7 @@ set(libprotobuf_lite_includes
 )
 
 if (MSVC)
-set(libprotoc_rc_files
+set(libprotobuf_lite_rc_files
   ${CMAKE_CURRENT_BINARY_DIR}/version.rc
 )
 endif()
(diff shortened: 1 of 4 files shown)


shardator:
Any chance that this can be accepted?

zchrissirhcz:
@acozzette Hi, I think there are still some people using protobuf 3.6 branch, and accept this PR to 3.6.x branch should be OK.

zchrissirhcz:
Actually, @shardator 's PR only change one thing: only if on MSVC platform, assign value to `protoc_rc_files` variable in `cmake/protoc.cmake`. This only-two-line fixing, however, takes me many minutes to figure it out. In my opinion, branch 3.6.x should at least be OK to be build.

shardator:
@acozzette I agree with @zchrissirhcz that even though 3.6.x will be "archived" it should be possible to build it on all platforms, that's why this patch is necessary. Since this change is non-functional, I think it is safe to accept it. Also if 3.6.x will be abandoned it cannot really cause too much harm.

PR #5470, opened 2018-12-14
Exclude command_line_interface_unittest.cc in MSVC build

meteorcloudy:
This will allow `bazel test //:all` to pass on Windows, thus we can enable Windows platform for protobuf on Bazel CI.
@acozzette Can you take a look?


acozzette:
Thanks, @meteorcloudy. Just for future reference, what is it about that test that doesn't work on Windows with MSVC?

meteorcloudy:
I didn't look carefully, but I believe it's just some test failures that could be fixed on Windows.

PR #5473, opened 2018-12-16
Support finding fields in UpperCamelCase when composing objects from JSON

### LINKED ISSUE #5414: JsonStringToMessage incorrectly rejects lowerCamelCase field names
Version: v3.6.0
Language: C++

Windows 10

Visual Studio 2017, x86 and x64 toolchains

**What did you do?**
1. Create a .proto file with an object with PascalCase fields
2. Build .proto file into C++
3. In C++ call JsonStringToMessage with a snippet of JSON which uses lowerCamelCase fields instead of PascalCase
4. Protobuf fails with "Cannot find field" error after failing to match the lowerCamelCase field in the JSON to the pascal case in the type descriptor.

The JSON mapping section on the protobuf guide [here](https://developers.google.com/protocol-buffers/docs/proto3#json) states **"Message field names are mapped to lowerCamelCase and become JSON object keys. ... Parsers accept both the lowerCamelCase name (or the one specified by the json_name option) and the original proto field name."**

**What did you see instead?**

The C++ parser does not accept the lowerCamelCase name in fields as it should.

YuryBandarchuk16:
Hello!

I have been trying to come up with an idea on how to resolve the #5414 issue. The main reason why this issue was created is because `ToJsonName` does not make the first letter lower comparing to `ToCamelCase` function which indeed makes the first letter lower.

Of course, one the solutions is just to change the documentation and change the status of this issue into "Won't Fix" :)

But I thought that it might be somehow useful to support find fields in UpperCamelCase when composing a proto object from JSON.

Since we are not allowed to break backwards compatibility and may not change the behaviour of `ToJsonName` function, I have came up with the following solution:

1. Add new option (`try_lower_camel_for_unknown_fields`) to `JsonParseOptions` which will be responsible for the needed check.
2. If the flag is set, we just need to check that the field name is UpperCamelCase of the name we are looking for (this logic is implemented in `FindLowerCamelFieldInTypeOrNull`).

Looking forward to hear your opinion! @acozzette @TeBoring @BSBandme 

Fix #5414

### DIFF
--- a/src/google/protobuf/util/internal/proto_writer.cc
+++ b/src/google/protobuf/util/internal/proto_writer.cc
@@ -69,6 +69,7 @@ ProtoWriter::ProtoWriter(TypeResolver* type_resolver,
       ignore_unknown_fields_(false),
       ignore_unknown_enum_values_(false),
       use_lower_camel_for_enums_(false),
+      try_lower_camel_for_unknown_fields_(false),
       element_(nullptr),
       size_insert_(),
       output_(output),
@@ -89,6 +90,7 @@ ProtoWriter::ProtoWriter(const TypeInfo* typeinfo,
       ignore_unknown_fields_(false),
       ignore_unknown_enum_values_(false),
       use_lower_camel_for_enums_(false),
+      try_lower_camel_for_unknown_fields_(false),
       element_(nullptr),
       size_insert_(),
       output_(output),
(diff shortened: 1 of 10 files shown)


wd40bomber7:
Hey @YuryBandarchuk16 thanks so much for addressing the issue I filed. I'd like to contribute but honestly I'm not familiar enough with how protobuf works to give meaningful feedback. Do you know if someone will be available to code review this PR at some point?

Thanks!

wd40bomber7:
Hey @acozzette it looks like you were the "pending reviewer" for this change. Is there still a way forward to get this merged? I'd really love if this fix was officially part of protobuf.

PR #5474, opened 2018-12-16
[Protoc] Show error on duplicate json names

### LINKED ISSUE #5063: protoc: does not verify that json_name does not conflict
Using `protoc v3.6.1`.

Compile this file:
```proto 
syntax = "proto3";
message M {
}
```

What I see: compile succeeds.

YuryBandarchuk16:
When having two fields with the same json name `protoc` did not fail with error neither showed a warning.

For example:
```proto
syntax = "proto3";
message Hi {
    int32 f1 = 1 [json_name = "fooBar"];
    int32 f2 = 2 [json_name = "fooBar"];
}
```

This MR shows warning when using `protoc` in the following format:
>> dup_json_test.proto: warning: Fields with names "Hi.f2" and "Hi.f1" have the same JSON name "fooBar". This may cause problems.

I decided to show warning instead of error in order to save backwards compatibility. It could be replaced with error later on.

cc: @xfxyjwf @acozzette @TeBoring @dsnet 

Fix #5063

### DIFF
--- a/src/google/protobuf/descriptor.cc
+++ b/src/google/protobuf/descriptor.cc
@@ -5885,13 +5885,28 @@ void DescriptorBuilder::ValidateProto3Message(
         message->field(i)->name());
     if (name_to_field.find(lowercase_name) != name_to_field.end()) {
       AddError(message->full_name(), proto,
-               DescriptorPool::ErrorCollector::OTHER,
+               DescriptorPool::ErrorCollector::JSON_NAME,
                "The JSON camel-case name of field \"" +
                message->field(i)->name() + "\" conflicts with field \"" +
                name_to_field[lowercase_name]->name() + "\". This is not " +
                "allowed in proto3.");
     } else {
       name_to_field[lowercase_name] = message->field(i);
+    }
+  }
+
+  HASH_MAP<string, const FieldDescriptor*> reserved_json_name_map;
+  for (int i = 0; i < message->field_count(); i++) {
+    const string &json_name = message->field(i)->json_name();
+    auto it = reserved_json_name_map.find(json_name);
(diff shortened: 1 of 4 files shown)


src/google/protobuf/descriptor.cc
```diff
@@ -5894,6 +5894,21 @@ void DescriptorBuilder::ValidateProto3Message(
       name_to_field[lowercase_name] = message->field(i);
     }
   }
+
+  HASH_MAP<string, const FieldDescriptor*> reserved_json_name_map;
```
dsnet: This problem affects proto2 as well, so I'm not sure the `ValidateProto3Message` function is the right place to add this.

dsnet:
No tests?

Sigh... I noted this issue a while ago. I would love to see this be an error and not a warning, but agree that this can break existing cases.

YuryBandarchuk16:
@dsnet Thank you for your comments! I have updated the tests which got broken after switching from warning to error and added more tests. Will appreciate if you take a look again!

wangkirin:
Hi all, I found this problem still exists in the master until now,  and your commit also conflicts with the latest. I would love to finish fixing this problem.

src/google/protobuf/descriptor.cc
```diff
@@ -4576,6 +4576,21 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
     }
   }
 
+  HASH_MAP<string, const FieldDescriptor*> reserved_json_name_map;
```
TeBoring, line 4579: Only proto3 message should have this check

TeBoring:
Feel free to reopen after sync the change.

PR #5475, opened 2018-12-17
Rebuild python docker image

TeBoring:


tests.sh
```diff
@@ -234,6 +282,66 @@ build_python_cpp() {
   cd ..
 }
 
+build_python27_cpp() {
+  internal_build_cpp
+  export LD_LIBRARY_PATH=../src/.libs # for Linux
+  export DYLD_LIBRARY_PATH=../src/.libs # for OS X
+  cd python
+    envlist=py27-cpp
```
haon4: why is this indented? Same comment below

tests.sh
```diff
@@ -220,6 +220,54 @@ build_python() {
   cd ..
 }
 
+build_python27() {
```
haon4: most of these functions are the same, can we call a helper function and just pass in the envlist as an argument?

TeBoring: done

tests.sh
TeBoring: Removed indent.

Issue #5192, opened 2018-09-27
[PHP] Error serializing empty Struct to JSON


jdpedrie:
**What version of protobuf and what language are you using?**
Version: v3.6.1
Language: PHP (PHP runtime, not PHP extension)

```php
use Google\Cloud\Dialogflow\V2\QueryResult;
use Google\Protobuf\Struct;

$queryResult = new QueryResult;
$queryResult->setParameters(new Struct);
echo $queryResult->serializeToJsonString();
```

Output:
```
{"parameters":}
```

This JSON string cannot be decoded.

When the Protobuf PHP extension is enabled, this issue is not present.

For context, see https://github.com/googleapis/google-cloud-php/issues/1320

This issue has been confirmed with different messages also containing Struct messages.

TeBoring:
Could you try the master branch? Recently, I have serval fixes for json encoding/decoding in c extension.

jdpedrie:
Hi @TeBoring, thanks for the reply. I tried the master branch, and get the same output as v3.6.1.

PR #5477, opened 2018-12-17
test(php):php json_encode empty ListValue test case.

lightjiao:
Add php test case.

### DIFF
--- a/php/tests/encode_decode_test.php
+++ b/php/tests/encode_decode_test.php
@@ -890,6 +890,16 @@ class EncodeDecodeTest extends TestBase
         $this->assertSame("[1.5]", $m->serializeToJsonString());
     }
 
+    public function testEncodeEmptyListValue()
+    {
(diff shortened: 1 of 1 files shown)


lightjiao:
I cannot run `./test.sh php7.2` or `cd php && composer install && composer test` command to test in local.

Any way I can test it in local ?

lightjiao:
I cp my last fix to this pr, it look good for me, but I don't know how to run unit test

by the way, I think my fix is not very good, I think should fix the `_deserializeResponse` function, when parse empty `ListValue`'s desc, but this is too hard for me, and would spend lot of my time. 

Could you fix this issue as soon as possible ?
If you think my fix is not good, I would to change this pr.
thx.🙂

lightjiao:
well, I only pr the test case.😂

TeBoring:
Fixed in #5192

PR #5481, opened 2018-12-18
Fix benchmark to use docker image on kokoro

BSBandme:
Currently still under testing in VM. Will add reviewers after testing passed


BSBandme:
Finished fixing some problems. @TeBoring PTAL. Thanks!

PR #5484, opened 2018-12-18
Do not add DEBUG_POSTFIX on non-Windows

XeCycle:
My projects builds protobuf together when building itself, using CMake ExternalProject; on Windows the CMake build works very well, but on Linux the file names are a little bit surprising.  UNIX-like systems typically do not ship separately release and debug libraries.

I am aware that I can use autotools on Linux, but the CMake build integrates better, e.g. parallel builds get controlled together.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -44,7 +44,12 @@ cmake_dependent_option(protobuf_MSVC_STATIC_RUNTIME "Link static runtime librari
   "NOT protobuf_BUILD_SHARED_LIBS" OFF)
 set(protobuf_WITH_ZLIB_DEFAULT ON)
 option(protobuf_WITH_ZLIB "Build with zlib support" ${protobuf_WITH_ZLIB_DEFAULT})
-set(protobuf_DEBUG_POSTFIX "d"
+if (WIN32)
+  set(protobuf_DEBUG_POSTFIX_DEFAULT "d")
+else (WIN32)
+  SET(protobuf_DEBUG_POSTFIX_DEFAULT "")
+endif (WIN32)
+set(protobuf_DEBUG_POSTFIX ${protobuf_DEBUG_POSTFIX_DEFAULT}
   CACHE STRING "Default debug postfix")
 mark_as_advanced(protobuf_DEBUG_POSTFIX)
 # User options


XeCycle:
Signed the CLA just now.

acozzette:
@XeCycle Thanks for the pull request, but I am a bit hesitant to accept this because I'm worried about adding extra complexity to the build configuration. I don't know that much about CMake but isn't there a way you could customize the value of `protobuf_DEBUG_POSTFIX` without editing the cmake files? That would hopefully allow you to set the postfix value to the empty string without us needing any new Windows-specific logic here.

XeCycle:
@acozzette Yes I understand, and indeed I simply set build type to Release so I am not actually affected.  However don't you think it better to make the default correct?  The Windows library naming thing is a fairly common problem and I have seen many projects keeping specific configurations.

Btw you reminded me of complexity, now I feel it may be better to tweak the default value of `protobuf_DEBUG_POSTFIX` instead of to dive down to each library target.  Will update.

aaron-bray:
I prefer the default to build the debug library with the 'd' suffix, its the more standard convention. More often than not, I see most libraries use the 'd' suffix convention, but its really up to your software package if you do it or not.

Visual Studio (not Windows) is an IDE that kind of lives in its own world (similar to XCODE) and does not use suffix's by default, as it can build multiple configurations in the same build directory by creating a debug dir and a release dir and it keeps all binaries separate. That is not the case with makefiles and build directories which I think is why the 'd' suffix came into existence. 

Using CMAKE you can hard code a value for the protobuf_DEBUG_POSTFIX in your ExternalProject_Add call to always be empty, if you want (but I don't recommend). But when using CMake, use the find_package(protobuf) in your code so that CMake understands what protobuf is and how it was built, it will configure your projects correctly for you. That way, you can simply add (into your project CMake): `target_link_libraries(YourProject libprotobuf)` and CMake will know to link to the right library to the right configuration, no matter what it's name is!

acozzette:
I had not heard of this 'd' suffix convention before, but it sounds like it's common among CMake projects in particular. I don't really have an opinion on whether we should use that convention, but I do think it's important to be consistent and avoid using different conventions on different platforms.

aaron-bray:
By default, when you build the release protobuf lib, you get libprotobuf.lib, when you build the debug library, you get libprotobufd.lib (same goes for .dll/.so extensions), at least with CMake, I am not sure if your other build tools use this naming convention. Do they not put the d on the end of the libprotobuf when you build debug?

acozzette:
Ah OK, we don't use this convention with our other build systems (autotools and Bazel). Bazel puts the debug and release build artifacts in different directories. I don't think the autotools build even makes a distinction, and it just requires you to fiddle with the CXXFLAGS manually if you care about the optimization level and debug info.

aaron-bray:
Sounds good. 

When you CMake on the source directory, you specify the CMAKE_BUILD_TYPE with DEBUG, RELEASE, RELWITHDEBINFO, or MINSIZEREL, and it will build that specific configuration and internally set all the CXXFLAGS to what they generally should be for such a build. (You can override them if you want)
The based on the what config is chosen you can program CMake to do special things, so the protobuf CMake does this naming convention deal, which is pretty standard in writing CMake.

But when using MSVC (mult-configuration IDE), CMake just ignores the build type, since you select it via the IDE, but you can still modify what flags are used for what if you don't like the defaults. 

It's great that protobuf does expose this CMake library suffix flag to be modified by the builder, that is not always done by CMake projects, but is really good practice.

XeCycle:
AFAICT the whole `DEBUG_POSTFIX` thing is only relevant on Windows.  ("as far as" = my experience on Linux, macOS, Solaris and Windows)

That Windows does this, is mainly because Debug and Release binaries link to different and ABI-incompatible msvcrt; therefore when distributing binary releases, the author is recommended to release both the Debug build and Release build, so that the user can build his own project in Debug or Release.  Visual Studio IDE projects build in Debug and Release subdirs and use the same file name, but that's just the build tree.  NMake and Ninja generators are single-config, and do not use those subdirs, but this is not the reason for different library names; it's just because they have to ship two builds, and they do not want the "install" dir to contain subdirs.  Some authors go another way, though.  i.e. some projects ship their package as
```
a-package/
    a.dll
    a.lib
    ad.dll
    ad.lib
```
and some other projects have
```
a-package/
    Debug/
        a.dll
        a.lib
    Release/
        a.dll
        a.lib
```

On *nix there's no such linking problems, we have a single "Release"-build libc, and every binary link to that, regardless of themselves being Debug or Release.  So many Linux-only devs never know about `DEBUG_POSTFIX`; they simply link to, say, `-lprotobuf` and never `-lprotobufd`.  When they want to debug protobuf itself they can create a Debug build and simply replace the Release file.  If we give the Debug binaries different names, they may end up with different SONAME; which, I'm afraid, is manually making an ABI breakage on *nix.  Having two coexistent libraries also means that, one can have parts of their app link to the Release binary, and othert parts to the Debug one; finally the whole app links in both binaries and may clash.

Btw there was even a (very old) [bug](https://cmake.org/pipermail/cmake/2004-October/005759.html) that `DEBUG_POSTFIX` was even used on Linux, implying that very few actually cares about this.

Anyway, I'm proposing to change the default to follow platform conventions, and the user can opt-in their choice.  I'm totally okay if you prefer to keep the current behavior and have users opt-out that postfix thing.

PR #5486, opened 2018-12-18
Use NoLocalStrtod to fix errors reading in decimal floats in Europe

aaron-bray:
I had issues with a colleague reading in json protobuf files with a decimal float separator rather than a comma.

This fixed his reading errors
I replaced all the std strtod calls with io::NoLocaleStrtod just to be safe.

This also was the only file still using the std strtod calls

### DIFF
--- a/src/google/protobuf/stubs/strutil.cc
+++ b/src/google/protobuf/stubs/strutil.cc
@@ -41,6 +41,7 @@
 #include <iterator>
 
 #include <google/protobuf/stubs/stl_util.h>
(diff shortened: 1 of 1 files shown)


aaron-bray:
I signed it!

acozzette:
@aaron-bray Thanks for the pull request. We're not doing anymore 3.6.x releases, but would you mind submitting this change against the master branch instead?

acozzette:
Let me close this PR since it has been superseded by #5489.

PR #5489, opened 2018-12-19
Use NoLocalStrtod for international compliance

aaron-bray:
Received errors parsing json files with decimal floats from European colleagues


acozzette:
@aaron-bray The CI is showing some build failures with the autotools build. To fix those I think you just have to update `src/Makefile.am` to move `strtod.cc` into the lite library so that `strutil.cc` can use it.

aaron-bray:
I also updated the CMake just to make sure these two build systems are consistent

acozzette:
@aaron-bray Thanks for the pull request!

PR #5494, opened 2018-12-19
Fix oneOf invalidation after MergeMessage

### LINKED ISSUE #4555: C++: FieldMask causes OneOf invalidation
This is similar to https://github.com/google/protobuf/issues/3115, which was indeed fixed by 3.4

We are using 3.5.1, and I'm observing what looks like `MergeMessageTo` changing which element in a oneof is set.

oneof_mask.proto:

```proto
syntax = "proto3";

message OneOfDemo {
  int64 id = 1;

  oneof demo {
    string foo = 2;
    SubMsg bar = 3;
  }
}

message SubMsg {
  string baz = 1;
}
```

oneof_mask_test.cc:

```c++
TEST(OneofMaskTest, OneofMaskValueStrippedTest) {
  OneOfDemo demo;
  demo.set_id(123);
  demo.set_foo("foo");

  google::protobuf::FieldMask mask;
  mask.add_paths("bar.baz");

  OneOfDemo dest;
  google::protobuf::util::FieldMaskUtil::MergeMessageTo(demo, mask, {}, &dest);

  EXPECT_EQ(dest.demo_case(), OneOfDemo::DEMO_NOT_SET);
  EXPECT_NE(dest.demo_case(), OneOfDemo::kBar);
  EXPECT_FALSE(dest.has_bar());
}
```

When running this test, both `demo_case` and `has_bar` are still set incorrectly. 

```
[ RUN      ] OneofMaskTest.OneofMaskValueStrippedTest
oneof_mask_test.cc:47: Failure
  dest.demo_case()
    Which is: 3
  OneOfDemo::DEMO_NOT_SET
    Which is: 0
oneof_mask_test.cc:48: Failure
Expected: (dest.demo_case()) != (OneOfDemo::kBar), actual: 3 vs 3
oneof_mask_test.cc:49: Failure
Value of: dest.has_bar()
  Actual: true
Expected: false
[  FAILED  ] OneofMaskTest.OneofMaskValueStrippedTest (0 ms)             
```

Interestingly, I don't see the same behavior when using `TrimMessage` - this test case passes:

```c++
TEST(OneofMaskTest, OneofMaskValueStrippedTestTrimMessage) {
  OneOfDemo demo;
  demo.set_id(123);
  demo.set_foo("foo");

  google::protobuf::FieldMask mask;
  mask.add_paths("bar.baz");

  google::protobuf::util::FieldMaskUtil::TrimMessage(mask, &demo);

  EXPECT_EQ(demo.demo_case(), OneOfDemo::DEMO_NOT_SET);
  EXPECT_NE(demo.demo_case(), OneOfDemo::kBar);
  EXPECT_FALSE(demo.has_bar());
}
```

YuryBandarchuk16:
In order to fix the invalidation of oneOf in `MergeMessage` firstly, we need to come up with the reason the invalidation occurs. 

Let's take a look at the proto example from the issue.

```proto
message OneOfDemo {
  int64 id = 1;

  oneof demo {
    string foo = 2;
    SubMsg bar = 3;
  }
}

message SubMsg {
  string baz = 1;
}
```

```cpp
OneOfDemo demo;
demo.set_id(123);
demo.set_foo("foo");

google::protobuf::FieldMask mask;
mask.add_paths("bar.baz");

OneOfDemo dest;
google::protobuf::util::FieldMaskUtil::MergeMessageTo(demo, mask, {}, &dest);
```

The reason for the invalidation of oneOf is because when merging `bar.baz`, we determine a *sub-message* `bar`, existence of which **is not** checked in the source of merging. 
Therefore, it is being created and then traversed to `baz`. Since `baz` is not set in the source, `ClearField` is called.  But the problem here is that we should have checked the existence of *sub-message* and cleared the appropriate field in merging destination. 

In the example above, once we figure out that `bar` does not exist in the source of merging, we must clear `bar` field in the destination object and skip all its children.  

RFC @acozzette @xfxyjwf @BSBandme 

Fix #4555


src/google/protobuf/util/field_mask_util.cc
```diff
@@ -469,8 +469,13 @@ void FieldMaskTree::MergeMessage(const Node* node, const Message& source,
                    << "have sub-fields.";
         continue;
       }
-      MergeMessage(child, source_reflection->GetMessage(source, field), options,
-                   destination_reflection->MutableMessage(destination, field));
+
+      if (!source_reflection->HasField(source, field)) {
+        destination_reflection->ClearField(destination, field);
```
acozzette: Actually in the case where the source message does not have the given field, I think the right thing to do is just leave the destination field alone instead of clearing it.

YuryBandarchuk16: The main reason for doing this is described on example in description for this PR. 

If shortly, it just copies the behaviour for primitives which are not specified in the FieldMask, since all of them are cleared in the if statement below. 

```cpp
if (!field->is_repeated()) {
       switch (field->cpp_type()) {
 #define COPY_VALUE(TYPE, Name)                                              \
   case FieldDescriptor::CPPTYPE_##TYPE: {                                   \
     if (source_reflection->HasField(source, field)) {                       \
       destination_reflection->Set##Name(                                    \
           destination, field, source_reflection->Get##Name(source, field)); \
     } else {                                                                \
       destination_reflection->ClearField(destination, field);               \
     }                                                                       \
     break;                                                                  \
   }
   ...
   ...
   ...
}
```

acozzette: OK, I think you are right. That does make sense that the behavior should be consistent with how it works for primitive types. I can see how one might want to use a field mask to clear a particular field. Is this covered by any tests, though? Maybe I am missing something but it looks like the new test doesn't cover the case of using a field mask to clear a field that already exists in the destination message.

src/google/protobuf/util/field_mask_util_test.cc
```diff
@@ -747,6 +749,36 @@ TEST(FieldMaskUtilTest, TrimMessage) {
...
+
+  TestOneOfMerging dest1;
+  google::protobuf::util::FieldMaskUtil::MergeMessageTo(demo, mask, {}, &dest1);
+
+  EXPECT_EQ(dest1.demo_case(), TestOneOfMerging::DEMO_NOT_SET);
+  EXPECT_NE(dest1.demo_case(), TestOneOfMerging::kBar);
+  EXPECT_FALSE(dest1.has_bar());
+  EXPECT_FALSE(dest1.has_foo());
+
+  OneOfSubMsg* sub_msg = new OneOfSubMsg();
```
acozzette: These three lines can be simplified to just one line: `demo.mutable_bar()->set_baz("baz_value");`

src/google/protobuf/util/field_mask_util.cc
YuryBandarchuk16: I believe I understood what you meant, but just want to double check it first :)

Am I right that you are looking for the test, which verifies the following:

We are having two messages like that 
```proto
message M {
    string first = 1;
    MM second = 2;
}

message MM {
     int64 value = 1;
}
```

* We have a source message `M` which **does not** have field *second*. 

* The field mask will be definitely having `second.value` as one of the paths.

* Destination message `M` which **has** something stored in field *second*.

As a result, after calling `MergeMessage` we should expect that destination will not have anything stored in field *second*.

Could you confirm, that this is the scenario we would like to cover?

acozzette: Right, that is exactly the scenario I was thinking of.

But actually, now that I have thought about it some more, I wonder if the logic still needs to change somewhat. Let's say the field mask is "a.b.c.d.e.f" and the source message does not have field "a". In that case the destination proto will have "a" completely cleared, even though the field mask is really just indicating a particular deeply nested field within "a". So I think the right answer is that we shouldn't clear any field until we get down to the very last element of the field mask path.

acozzette:
@YuryBandarchuk16 Thank you for the change but I am going to close this for now since there has been no activity for a while. I'm still wondering if this will work properly in the case I mentioned of a deeply nested field mask `a.b.c.d.e.f`. If you're still interested in working on this, though, feel free to comment and I can reopen it.

PR #5496, opened 2018-12-19
Fix Any json encoding/decoding in php

TeBoring:


TeBoring:
@michaelbausor Could you help take a look?

php/tests/encode_decode_test.php
```diff
@@ -915,4 +917,127 @@ public function testEncodeTopLevelStruct()
...
+        $packed->setSeconds(946684800);
+        $packed->setNanos(123456789);
+        $m = new Any();
+        $m->pack($packed);
+        $this->assertSame(
+            "{\"@type\":\"type.googleapis.com/google.protobuf.Timestamp\"," .
+            "\"value\":\"2000-01-01T00:00:00.123456789Z\"}",
+            $m->serializeToJsonString());
+    }
+
```
michaelbausor: What do we expect to happen if we try to deserialize a json Any when the packed type in unknown? Can we test for that? Ideally we would still get an Any object, but with unknown data inside it. But maybe that is not possible?

TeBoring: I think the behavior should be failure.

TeBoring: Added test to verify unknown message throws exception.

TeBoring:
I don't think the go and objc test is related to this change.

TeBoring:
@michaelbausor, fixed your comments. Any more comments?

PR #5498, opened 2018-12-20
PHP library CodedOutputStream.php  buffer 

yqbaa:
public function __construct($size)
    {
        $this->current = 0;
        $this->buffer_size = $size;
        $this->buffer = str_repeat(chr(0), $this->buffer_size);
    }
why you create this buffer , it weel course memory double.


TeBoring:
Why there are so many commits?

PR #5500, opened 2018-12-20
Optionally make arenastring.h safer for shared library use

ThomasColthurst:
If the C++ protocol buffer code is loaded into memory multiple times as a shared library (which happens quite often when multiple different Python modules use the "fast cpp" protobuf implementation), arenastring.h's IsDefault check isn't quite right, because the current default_value can be different than the default_value the protocol buffer was initialized with.  This pull request adds a #define named "SHARED_LIBRARY_SAFE" that, if enabled, will use a safer but sometimes more expensive check.

### DIFF
--- a/src/google/protobuf/arenastring.h
+++ b/src/google/protobuf/arenastring.h
@@ -68,7 +68,7 @@ class TaggedPtr {
 struct PROTOBUF_EXPORT ArenaStringPtr {
   inline void Set(const ::std::string* default_value,
                   const ::std::string& value, ::google::protobuf::Arena* arena) {
-    if (ptr_ == default_value) {
+    if (IsDefault(default_value)) {
       CreateInstance(arena, &value);
     } else {
       *ptr_ = value;
@@ -86,7 +86,7 @@ struct PROTOBUF_EXPORT ArenaStringPtr {
 
   inline ::std::string* Mutable(const ::std::string* default_value,
                            ::google::protobuf::Arena* arena) {
-    if (ptr_ == default_value) {
+    if (IsDefault(default_value)) {
       CreateInstance(arena, default_value);
     }
     return ptr_;
@@ -98,7 +98,7 @@ struct PROTOBUF_EXPORT ArenaStringPtr {
   // release_<field>() methods on generated classes.
   inline ::std::string* Release(const ::std::string* default_value,
                            ::google::protobuf::Arena* arena) {
-    if (ptr_ == default_value) {
+    if (IsDefault(default_value)) {
       return NULL;
     }
     return ReleaseNonDefault(default_value, arena);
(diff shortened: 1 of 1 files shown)


ThomasColthurst:
Friendly ping.

src/google/protobuf/arenastring.h
```diff
@@ -352,7 +352,11 @@ struct PROTOBUF_EXPORT ArenaStringPtr {
   }
 
   inline bool IsDefault(const ::std::string* default_value) const {
+#ifdef SHARED_LIBRARY_SAFE
+    return ptr_ == default_value || ptr_->empty();
```
acozzette: I suspect that this logic will not quite work, because the string might just happen to be empty even if it's not the default string.

ThomasColthurst: That's right -- this will sometimes return True for IsDefault, even when ptr_ has been allocated to a non-default value.  And that will lead to memory leaks, as CreateInstance is called and re-assigns to an already allocated ptr_.

But:  I think the alternatives are worse, in that they either involve memory corruption (which is what happens now without SHARED_LIBRARY_SAFE) or adding something like an explicit is_allocated_ bool to the class, which will likely cause much more memory consumption than the rare memory leaks.

I'd be very happy to consider other alternatives, though, if you can think of any.

acozzette: I think the core problem here is that dynamically loading protobuf more than once is not really a use case we have ever tried to support, so I suspect that will be a fragile setup unless we make a concerted effort to get that setup working well. Could you provide some more details on how you're running into this issue? If multiple Python modules are independently loading the same protobuf code, doesn't that result in ODR violations anyway? I am not very familiar with how that works in Python but it seems like there should be some way to have multiple modules using protobuf without loading the code multiple times.

ThomasColthurst: Well, all Python extension modules are dynamically loaded, so for example, _message.so defined at https://github.com/protocolbuffers/protobuf/blob/master/BUILD#L684 will be dynamically loaded.

Our code is at https://github.com/google/nucleus -- we use bazel and CLIF and protobuf with some custom bazel rules for CLIF at https://github.com/google/nucleus/blob/master/third_party/clif.bzl and a slightly modified protobuf BUILD file at
https://github.com/google/nucleus/blob/master/third_party/protobuf.BUILD

Each .clif rule creates a new extension module, and many of those extension modules include compiled protobuf code in order to use the PyProto_API.  (And I'll note that the PyProto_API very carefully uses a Python capsule interface described at https://docs.python.org/2/extending/extending.html#using-capsules which is explicitly for communicating across dynamically loaded moudles.)

And that works just fine as long as (1) we include our protocol buffer descriptors in the _message.so module [that's what the modified protobuf.BUILD file is doing] and (2) we only read [and not write] the "cross-module" protocol buffers.

We would like to get rid of that read-only restriction, and we've found that there are only a couple of blockers to doing so, one of which is addressed in this pull request.  If/when this pull request is accepted, we'll address the other blockers in a follow-up pull request.

acozzette:
If I remember right we decided to go in a different direction based on our meeting, so let me close this pull request for now.

PR #5509, opened 2018-12-26
Enable ADVANCED_OPTIMIZATIONS in Closure Compiler

### LINKED ISSUE #2303: JavaScript: Closure compilation at the advanced level (WIP)
Hey @haberman!

I've recently found your [TODO](https://github.com/google/protobuf/blob/104723fccf4c9f7bf2398073e6b04390dcd2298a/js/gulpfile.js#L63-L64) and started adding those proper `@export` etc. annotations.

I'm actually nearly done now: All required annotations are added, the gulp file as well as the tests have been modified & fixed. The problem I now have is that you guys are using `goog.asserts`, which gets stripped out on the advanced closure level and there is no CLI switch to disable that.

So... What should be the way forward now? Should I simply copy/paste the required parts from `goog.asserts` into this project, write a modified closure-compiler CLI or simply drop all asserts in the dist release? 🙂

cc: @xfxyjwf 

P.S.: The size of the gzipped dist release is about 60% smaller (28kB vs. 11kB) and permanently (!) about 30-50% (!) faster without the asserts @ advanced level. So maybe if we continue dropping all non-essential asserts we could even gain a non-trivial speed improvement (or release a "fast" version alternatively).

jjbubudi:
Fixes #2303

After optimization the bundle size went from ~156KB to ~48KB (uncompressed)


jjbubudi:
I signed it!

js/commonjs/export_testdeps.js
```diff
@@ -13,12 +13,5 @@ goog.provide('jspb.ExportTestDeps');
...
 goog.require('goog.testing.PropertyReplacer');
-goog.require('jspb.arith.Int64');
-goog.require('jspb.arith.UInt64');
-goog.require('jspb.BinaryEncoder');
-goog.require('jspb.BinaryDecoder');
-goog.require('jspb.BinaryWriter');
-goog.require('jspb.utils');
 
 exports.goog = goog;
-exports.jspb = jspb;
```
jjbubudi: `jspb` is available in the `google-protobuf` module. Exporting `jspb` here and subsequently using it in test will result in a lot of undefined symbol errors because the compiled `google-protobuf` is expecting obfuscated symbols.

js/commonjs/rewrite_tests_for_commonjs.js
```diff
@@ -67,7 +67,7 @@ console.log("var global = Function('return this')();");
 console.log("var googleProtobuf = require('google-protobuf');");
 console.log("var testdeps = require('testdeps_commonjs');");
 console.log("global.goog = testdeps.goog;");
-console.log("global.jspb = testdeps.jspb;");
+console.log("global.jspb = googleProtobuf.jspb;");
```
jjbubudi: See comment above. We should use `google-protobuf`'s jspb instead of the one we exported for tests.

skyjia:
any updates on this PR?

PR #5511, opened 2018-12-26
Revert "Enable the ignore_unknown_field option in the Ruby unmarshal options"

TeBoring:
Reverts protocolbuffers/protobuf#5373
Breaks continuous test.


ebenoist:
Can I help get this back in? What test did this break?

PR #5516, opened 2018-12-27
Clean up unnecessary compile warning with GCC

KeithMoyer:
The -Wno-writable-strings warning flag is Clang-specific. GCC's
equivalent is -Wno-write-strings, which Clang also supports as a
synonym. So, -Wno-write strings is more compatible and there is no need
to specify both.

https://clang.llvm.org/docs/DiagnosticsReference.html#wwrite-strings
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

Compiled with both GCC 8.2 and Clang trunk.


acozzette:
Thanks @KeithMoyer!

PR #5522, opened 2018-12-30
NVCC complains about static constexpr in dllexport class.

### LINKED ISSUE #5521: CUDA issue with static constexpr in dllexport class: kMaximumFlatCapacity may not be initialized
Version: master/v3.6.1.3
Language: C++

Win10

VS 2017 (15.9.4) + CUDA 10.0

**What did you do?**
Build as DLL for CUDA libraries (e.g. caffe2).

**What did you see instead?**

Error message like `member "XXX" may not be initialized`.

Seems NVCC do not allow static constexpr in dllexport class:
https://github.com/protocolbuffers/protobuf/blob/be1716a6d0346de7a990b60f4afc9aebe2e06f36/src/google/protobuf/extension_set.h#L659
https://github.com/protocolbuffers/protobuf/blob/66dc42d891a4fc8e9190c524fd67961688a37bbe/src/google/protobuf/compiler/cpp/cpp_message.cc#L950

xkszltl:
Resolve #5521

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -1078,7 +1078,7 @@ void MessageGenerator::GenerateClassDefinition(io::Printer* printer) {
       "  return reinterpret_cast<const $classname$*>(\n"
       "             &_$classname$_default_instance_);\n"
       "}\n"
-      "static constexpr int kIndexInFileMessages =\n"
+      "static const int kIndexInFileMessages =\n"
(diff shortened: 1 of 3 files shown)


xkszltl:
I signed it!

acozzette:
@xkszltl Sorry for taking so long to get back to this. I am a little bit hesitant to take this change because I am worried that there might be code depending on these constants to be `constexpr`. Do you have any more background on the issue with NVCC? Is this an NVCC bug that could be fixed by the NVCC developers?

PR #5528, opened 2019-01-02
Add ruby 2.6 test

TeBoring:


TeBoring:
Broken ruby tests should not be related.

PR #5532, opened 2019-01-03
Fix Empty ListValue/Struct json encoding

TeBoring:


TeBoring:
@michaelbausor could you help take a look?

TeBoring:
Fix #5192

TeBoring:
Broken ruby tests should not be related.

php/tests/encode_decode_test.php
```diff
@@ -898,6 +898,13 @@ public function testEncodeTopLevelListValue()
         $this->assertSame("[1.5]", $m->serializeToJsonString());
     }
 
+    public function testEncodeEmptyListValue()
+    {
+        $m = new TestMessage();
+        $m->setRepeatedInt32([]);
```
michaelbausor: This doesn't look like it tests ListValue?

TeBoring: Oh, I guess I removed the wrong test. Thanks for the catch.

TeBoring:
Thanks for the review!

PR #5536, opened 2019-01-04
Update raker compiler docker image for ruby2.6

TeBoring:


TeBoring:
I found this change doesn't work.
Will create another one...

Issue #4460, opened 2018-03-31
Segmentation Fault (Segfault) on the ruby gem


larribas:
Hello there,

The following segmentation fault pops up upon running

`ruby -e 'require "google/cloud/trace"'`

inside of a Docker container based on `ruby:2.5-alpine` (alpine 3.7.0)

## Gemfile.lock (partial)

```
    google-cloud-core (1.2.0)
      google-cloud-env (~> 1.0)
    google-cloud-env (1.0.1)
      faraday (~> 0.11)
    google-cloud-trace (0.31.0)
      google-cloud-core (~> 1.2)
      google-gax (~> 1.0)
      stackdriver-core (~> 1.3)
    google-gax (1.0.1)
      google-protobuf (~> 3.2)
      googleapis-common-protos (>= 1.3.5, < 2.0)
      googleauth (~> 0.6.2)
      grpc (>= 1.7.2, < 2.0)
      rly (~> 0.2.3)
    google-protobuf (3.5.1.2)
    googleapis-common-protos (1.3.7)
      google-protobuf (~> 3.0)
      googleapis-common-protos-types (~> 1.0)
      grpc (~> 1.0)
    googleapis-common-protos-types (1.0.1)
      google-protobuf (~> 3.0)
    googleauth (0.6.2)
      faraday (~> 0.12)
      jwt (>= 1.4, < 3.0)
      logging (~> 2.0)
      memoist (~> 0.12)
      multi_json (~> 1.11)
      os (~> 0.9)
      signet (~> 0.7)
    grpc (1.10.0)
      google-protobuf (~> 3.1)
      googleapis-common-protos-types (~> 1.0.0)
      googleauth (>= 0.5.1, < 0.7)
```

## Ruby Segfault Dump

```
/usr/local/bundle/gems/google-protobuf-3.5.1.2-x86_64-linux/lib/google/protobuf/any_pb.rb:15: [BUG] Segmentation fault at 0x0000000000016e16
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-musl]

-- Control frame information -----------------------------------------------
c:0039 p:---- s:0240 e:000239 CFUNC  :msgclass
c:0038 p:0022 s:0236 e:000235 CLASS  /usr/local/bundle/gems/google-protobuf-3.5.1.2-x86_64-linux/lib/google/protobuf/any_pb.rb:15
[...]

PR #5537, opened 2019-01-04
Add ruby 2.6 gem build for linux

TeBoring:


TeBoring:
Failed tests seem to be caused by the new mac image from kokoro.

TeBoring:
This change modifies the script to build ruby gems, which should not be related to all these failing tests.

Envek:
@oggy this works for me:
```ruby
# Gemfile
gem "google-protobuf", git: "https://github.com/protocolbuffers/protobuf"
```
Isn't it works for you?

aadityataparia:
@Envek this doesn't work,

oggy:
@Envek No. It bundles ok, but because there are ruby files that need to be generated before the gem is built, you can't, e.g., `require 'google/protobuf/struct_pb'`.

Envek:
@oggy, @aadityataparia, @romansklenar, then try to set `BUNDLE_FORCE_RUBY_PLATFORM` environment variable to `1` to make Bundler ignore precompiled binaries and thus build them from scratch.

See https://github.com/protocolbuffers/protobuf/issues/4460#issuecomment-434631296 for reference.

PR #5540, opened 2019-01-04
`Libs` should not include `@PTHREAD_CFLAGS@`

ruoso:
That should be only on the CFlags. When using the compiler
as the linker driver, one may want to also call pkg-config to get the
cflags, but when invoking the linker, there should be no compiler flags.

### DIFF
--- a/protobuf-lite.pc.in
+++ b/protobuf-lite.pc.in
@@ -6,6 +6,6 @@ includedir=@includedir@
 Name: Protocol Buffers
 Description: Google's Data Interchange Format
 Version: @VERSION@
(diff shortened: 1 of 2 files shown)


acozzette:
@ruoso Thank you for the pull request. We just have to get the CLA sorted out before I can merge this, but I sent you an email with some more details on how to do that.

ruoso:
I signed it!

ruoso:
Actually, just realized protobuf-lite.pc has the same issue... let me add another commit.

ruoso:
Ok, rebased and now fix both `protobuf.pc` and `protobuf-lite.pc`

Issue #4852, opened 2018-06-30
Compiled nested message definitions return wrong variable name


wlwanpan:
Version: master/v3.5.0.
Language: Ruby
OS: macOS
Compiler: gcc

1. Add a nested message definition
```
message parent_msg
{
   message child_msg {}
}
```
2. Compile with flag --ruby_out

3. Result
```
Google::Protobuf::DescriptorPool.generated_pool.build do
    add_message "parent_msg" do
    end

    add_message "parent_msg.child_msg" do
    end
end

Parent_msg = Google::Protobuf::DescriptorPool.generated_pool.lookup("parent_msg").msgclass
parent_msg::Child_msg = Google::Protobuf::DescriptorPool.generated_pool.lookup("parent_msg.child_msg").msgclass
```

4. Error
```
undefined local variable or method `parent_msg' for main:Object
```

Expected
```
Parent_msg::Child_msg = Google::Protobuf::DescriptorPool.generated_pool.lookup("parent_msg.child_msg").msgclass
```

Got
```
parent_msg::Child_msg = Google::Protobuf::DescriptorPool.generated_pool.lookup("parent_msg.child_msg").msgclass
```

TeBoring:
I'll have a fix for that.

fowles:
Closing out old issues.  Please reopen if this still occurs on recent releases.

PR #5564, opened 2019-01-09
Ruby nested msg caps

toc007:
#4852
Capitalize all prefix while printing out all nested messages.

### DIFF
--- a/ruby/tests/generated_code.proto
+++ b/ruby/tests/generated_code.proto
@@ -69,10 +69,8 @@ enum TestEnum {
   C = 3;
 }
 
-message TestNestedMsg {
-  message testLowercase {
-    message lowercase{}
-  }
+message testLowercaseNested {
+  message lowercase{}
 }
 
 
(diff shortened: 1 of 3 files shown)


toc007:
I signed it!

TeBoring:
Would you mind adding a test?

ruby/tests/generated_code.proto
```diff
@@ -53,6 +53,13 @@ message TestMessage {
   map<string, bool> map_string_bool = 70;
 
   message NestedMessage {
+    message Hello {
```
TeBoring: Does this cause problem without your fix? I thought a lower letter started name causes the problem.

ruby/tests/generated_code.proto
```diff
@@ -69,6 +69,12 @@ enum TestEnum {
   C = 3;
 }
 
+message TestNestedMsg {
```
TeBoring: Could you give it a better name? e.g.,
```
message testlowercase {
  message lowercase {}
}
message _test_underscore {
  message _underscore {}
}
```

ruby/tests/generated_code.proto
```diff
@@ -69,6 +69,13 @@ enum TestEnum {
...
```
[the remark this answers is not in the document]
toc007: Would you like me to have testLowercase as the top-level message, or get rid of TestNestedMsg entirely?

PR #5566, opened 2019-01-10
Down-integrate internal changes to github.

BSBandme:


BSBandme:
Should we make some comments on the Text format conformance test?

TeBoring:
SG. Google internal only. Opensource testees just skip it.

TeBoring:
There are some conflicts. Could you resolve?

conformance/conformance.proto
```diff
@@ -70,8 +71,14 @@ enum TestCategory {
   JSON_IGNORE_UNKNOWN_PARSING_TEST = 3;
   // Test jspb wire format. Google internal only. Opensource testees just skip it.
   JSPB_TEST = 4;
+  // Test text format. Google internal only currently. Opensource won't have any
```
thomasvl: This does not appear to be correct given the Java conformance test is trying to support them.  So it would appear some tests will be in opensource soon.

BSBandme: Done fixing by changing to "for now".

TeBoring: Probably mention only c++, java and python currently need to care about it.

BSBandme: Done

conformance/conformance.proto
```diff
@@ -70,8 +71,15 @@ enum TestCategory {
   JSON_IGNORE_UNKNOWN_PARSING_TEST = 3;
   // Test jspb wire format. Google internal only. Opensource testees just skip it.
   JSPB_TEST = 4;
+  // Test text format. Google internal only currently. Opensource doesn't 
+  // have any test with this category for now. For cpp java and python,
+  // testee can already dual with this type.
```
TeBoring: Do you mean deal?

TeBoring: Test text format. For cpp, java and python, testees can already deal with this type. Testees of other languages can simply skip it.

BSBandme: Done.

BSBandme:
Any updates for this?

PR #5281, opened 2018-10-22
Python compiler takes keywords into account then generating code for top level enums.

yoks:
At the moment protobuf compiler for Python is not taking into account Python keywords then generating top level enums e.g. following pb:
```protobuf
syntax = "proto3";

enum TestEnum {
    None = 0;
    Something = 1;
    for = 2;
    class = 3;
}
```
Will generate invalid python code:
```python
# Top of the file omitted
TestEnum = enum_type_wrapper.EnumTypeWrapper(_TESTENUM)
None = 0
Something = 1
for = 2
class = 3
```

This merge requests proposes change to generator so it will append underscore `_` character to reserved Python keywords as stated in [PEP8](https://www.python.org/dev/peps/pep-0008/)

> If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss. (Perhaps better is to avoid such clashes by using a synonym.)

Generating following code from proto file:
```python
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: tests.proto

import sys
_b=sys.version_info[0]<3 and (lambda x:x) or (lambda x:x.encode('latin1'))
from google.protobuf.internal import enum_type_wrapper
from google.protobuf import descriptor as _descriptor
from google.protobuf import message as _message
from google.protobuf import reflection as _reflection
from google.protobuf import symbol_database as _symbol_database
# @@protoc_insertion_point(imports)

_sym_db = _symbol_database.Default()




DESCRIPTOR = _descriptor.FileDescriptor(
  name='tests.proto',
  package='',
  syntax='proto3',
  serialized_options=None,
  serialized_pb=_b('\n\x0btests.proto*7\n\x08TestEnum\x12\x08\n\x04None\x10\x00\x12\r\n\tSomething\x10\x01\x12\x07\n\x03\x66or\x10\x02\x12\t\n\x05\x63lass\x10\x03\x62\x06proto3')
)

_TESTENUM = _descriptor.EnumDescriptor(
  name='TestEnum',
  full_name='TestEnum',
  filename=None,
  file=DESCRIPTOR,
  values=[
    _descriptor.EnumValueDescriptor(
      name='None', index=0, number=0,
      serialized_options=None,
      type=None),
    _descriptor.EnumValueDescriptor(
      name='Something', index=1, number=1,
      serialized_options=None,
      type=None),
    _descriptor.EnumValueDescriptor(
      name='for', index=2, number=2,
[...]

Issue #4833, opened 2018-06-26
Python reserved word 'None' is not considered in code generation


FlowerCode:
**What version of protobuf and what language are you using?**
Version: v3.6.0
Language: Python

**What operating system (Linux, Windows, ...) and version?**
Windows 10

**What runtime / compiler are you using (e.g., python version or gcc version)**
Python 3.6.5

**What did you do?**
1. Define Enum

```
enum NotificationType {
    None = 0;
    Type1 = 1;
    Type2 = 2;
    Type3 = 3;
}
```

2. Compile to Python

`protoc.exe -I=.\pb --python_out=.\pb_python .\pb\test.proto`

3. Run it in Python

**What did you expect to see**
No Error

**What did you see instead?**
```
  File "test.py", line 10
    None = 0
SyntaxError: cannot assign to None
```

anandolee:
Update:

None is considered in Python keywords. 
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/python/python_generator.cc#L113

The error is not only for None. Message, enum, enum value, extension all have issues with keywords conflict. There's an ongoing fix in internal which will make getattr() work with keyword conflict.

FlowerCode:
@anandolee 
That's great news. Thanks!

zhangskz:
Closing since 3.8.0 has long been released.

PR #5569, opened 2019-01-11
Add ContainsPythonKeyword check for message names

johnsbrew:
### In a nutshell
Make the generated python work even when message names are python keyworks

### Context
- This PR complets #5281 
- This PR helps fixing #4833
----

### Example of the issue

Source with python-reserved keyword "None" used as message name...
```proto
syntax = "proto3";
message None {
    string name = 1;
}
```

...compiles into invalid python:
```python
# ...
None = _reflection.GeneratedProtocolMessageType('None', (_message.Message,), dict(
  DESCRIPTOR = _NONE,
  __module__ = 'test_pb2'
  # @@protoc_insertion_point(class_scope:None)
  ))
_sym_db.RegisterMessage(None)
```

```bash
  File "test_pb2.py", line 61
    None = _reflection.GeneratedProtocolMessageType('None', (_message.Message,), dict(
SyntaxError: cannot assign to None
```

### Proposal: Escape message name that are Python keywords

Like in #5281 this PR proposes change the generator so it appends underscore _ character to reserved Python keywords as stated in [PEP8](https://www.python.org/dev/peps/pep-0008/)
> If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss. (Perhaps better is to avoid such clashes by using a synonym.)

### Result
Diff of the generated files for the above example

```bash
--- test_pb2.py.before
+++ test_pb2.py.after
@@ -58,12 +59,12 @@ _NONE = _descriptor.Descriptor(
 DESCRIPTOR.message_types_by_name['None'] = _NONE
 _sym_db.RegisterFileDescriptor(DESCRIPTOR)

-None = _reflection.GeneratedProtocolMessageType('None', (_message.Message,), dict(
+None_ = _reflection.GeneratedProtocolMessageType('None_', (_message.Message,), dict(
   DESCRIPTOR = _NONE,
   __module__ = 'test_pb2'
   # @@protoc_insertion_point(class_scope:None)
   ))
-_sym_db.RegisterMessage(None)
+_sym_db.RegisterMessage(None_)


 # @@protoc_insertion_point(module_scope)
```

### DIFF
--- a/src/google/protobuf/compiler/python/python_generator.cc
+++ b/src/google/protobuf/compiler/python/python_generator.cc
@@ -834,12 +834,14 @@ void Generator::PrintMessages() const {
 void Generator::PrintMessage(const Descriptor& message_descriptor,
                              const string& prefix,
                              std::vector<string>* to_register) const {
-  string qualified_name(prefix + message_descriptor.name());
+  string message_name = message_descriptor.name();
+  string safe_message_name = ContainsPythonKeyword(message_name) ? message_name + "_" : message_name;
+  string qualified_name(prefix + safe_message_name);
   to_register->push_back(qualified_name);
   printer_->Print(
       "$name$ = _reflection.GeneratedProtocolMessageType('$name$', "
       "(_message.Message,), dict(\n",
-      "name", message_descriptor.name());
+      "name", safe_message_name);
   printer_->Indent();
 
   PrintNestedMessages(message_descriptor, qualified_name + ".", to_register);


anandolee:
Thanks for raise it up. We are currently discussing the keywords problems for all protobuf languages. For python, not only message has this issue. Keywords not work on message, enum, enum value, extension field, service.  And python2 python3 have different keywords that we should consider.

There is an on going proposal for python in google internal and waiting for approval

anandolee:
Update: as keyword conflict on fields can work with getattr()/setattr(), append '_' to other descriptors may confuse users. Another proposal that make getattr() work with message, enum, enum value, extension is under review

anandolee:
The change is submitted in internal. Expect to see in 3.8.0 release

anandolee:
Closing it as the keyword support is already in master

Issue #4473, opened 2018-04-03
[PHP]: “Expect [type]” should be “Expected [type] for method [class::method]”


bshaffer:
Related to #3371 and #3331

Errors from `GPBUtil` are a little hard to track down. The error is `Expect int`, where it would be better as `Expected int for method Google\Cloud\StorageClient::getBuckets` (for instance), so as to make these errors easy to debug.

If we use typehints (#3331), we still would want to improve this for primitive types (`string`, `int`, `bool`), unless we decide to make this library only compatible for PHP 7.0 and above.

Issue #3330, opened 2017-07-07
PHP: Second argument to GPBUtil::checkEnum is not used


bshaffer:
The GPBUtil class has a function called [`checkEnum`](https://github.com/google/protobuf/blob/master/php/src/Google/Protobuf/Internal/GPBUtil.php#L8), which is an alias for `checkInt32`.

The generated enum fields [call this function](https://github.com/google/protobuf/blob/12acbc2678073c3439b427be0b713b97e2074bfb/php/src/Google/Protobuf/Internal/MethodOptions.php#L101) and pass in a second argument which is the class name of the enum field, e.g. `MyEnum::class`. 

So either this second field should be used to determine if the enum integer is valid, or it should not be supplied as an argument to `checkEnum`.

TeBoring:
This is implementation detail.

sam0737:
The protoc (v3.6.1) generate setter codes that call GPBUtil::checkEnum($var, \CLASSNAME)
But GPBUtil::checkEnum in the PHP source code is just expecteding 1 input parameter.

Although it works, but it flags as error with static code checker.

I am afraid this is not just implementation detail.

PR #5571, opened 2019-01-12
Php dev merge

TeBoring:


TeBoring:
@bshaffer

TeBoring:
ping @bshaffer

php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -307,7 +307,7 @@ public function getOptions()
      * @param \Google\Protobuf\Internal\MessageOptions $var
      * @return $this
      */
-    public function setOptions($var)
+    public function setOptions(\Google\Protobuf\Internal\MessageOptions $var = null)
```
dwsupplee: I'm excited to see this change, but should we consider this breaking? If users pass in anything other than `MessageOptions|null` they will get a different error now than the one provided by `GPBUtil::checkMessage`. 

Assuming we move forward, however, a few thoughts:

- With the typecheck in place, are there any objections to removing the `GPBUtil` check?
- There is a mixture of style in the class, so I understand not being strict about this - but if we are open to the idea, moving the needle towards using relative namespaces instead of absolute would be great :).

TeBoring: What do you mean by relative namespace? Could you give me an example?

dwsupplee: > Currently, whether to throw error or exception for getter/setter/constructor is undefined behavior.

I'm not sure I understand. Without this changeset it is currently possible for users to have code like the following:

```php
use Google\Protobuf\Internal\DescriptorProto;

try {
    (new DescriptorProto)
        ->setOptions(new \stdClass);
} catch (\Exception $ex) {
    // Handling the exception thrown by GPBUtil::checkMessage
    echo 'There was a problem.';
}
```

On PHP 7+, for example, with this changeset added the above code would break with the following `TypeError`:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to
Google\Protobuf\Internal\DescriptorProto::setOptions() must
be an instance of Google\Protobuf\Internal\MessageOptions
or null, instance of stdClass given
```

> GPBUtil also checks other primitive fields. Since php5 doesn't provide typehint for primitive fields and we are currently still supporting that. We cannot remove GPBUtil.

Yes, that definitely is fair for primitives. I didn't mean to suggest removing the check across the board, just for situations where we would already be relying on a typehint.

> What do you mean by relative namespace? Could you give me an example?

Sure!

```php
// Instead of..
public function setOptions(\Google\Protobuf\Internal\MessageOptions $var = null)
{ }

// Import the class and use the relative name
use Google\Protobuf\Internal\MessageOptions;

public function setOptions(MessageOptions $var = null)
{ }
```

TeBoring: Expecting setter to throw an exception is an undefined behavior. Users should not write code like that. So I think we are good to move forward.

TeBoring: I think it's ok to remove the message check part in GPBUtil.

TeBoring: What if there are same message name under different namespace? e.g., A\Foo and B\Foo

dwsupplee: I still don't understand what you mean by undefined behavior. We have GA level libraries with these generated classes living in our codebase over on google-cloud-php. We have no such notice published anywhere so I don't know how users would know they shouldn't catch the exception that is thrown by the validation utility.

> What if there are same message name under different namespace? e.g., A\Foo and B\Foo

Good question. In cases like that aliases can be used:

```php
use A\Foo as AFoo;
use B\Foo;
```

TeBoring: By "undefined behavior", I mean although the code works like this, but we have never said we will guarantee such behavior. Such usage of undefined behavior may be broken by future release. (I think it's a common situation for opensource library).

TeBoring: ummm... I feel it's a little bit intuitive from AFoo to figure out A\Foo.
What if both A\Foo and AFoo already exist?

TeBoring: To state all undefined behavior is impracticable. Unless there is unlimited test cases, we cannot know how many undefined behaviors there are. The only way we can provide is unless we explicitly say "You can rely on this", users should not do that.

dwsupplee: > To state all undefined behavior is impracticable.

Yes, agreed. That is why it should be an edge case.

Could you help me understand what process we use here to define behavior as "expected"?

In my experience, once we have a stable public API the "expected" behavior is the exact behavior that the code performs and users should be able to rely on that. The exception being bugs and/or when we document that there may be an outlier for a given reason (for example, perhaps we mark a publicly accessible method as internal and say it shouldn't be relied on by users).

TeBoring: https://developers.google.com/protocol-buffers/docs/reference/php-generated
We have public doc for the API. The expected behavior should be stated there.
Although I agree this doc may miss some API later added. We should make the doc complete.

dwsupplee: Even if we attempt to complete the document, there is a very high chance it will fall out of alignment as new features are introduced. It seems very risky to ask users to rely on that as the source of truth for how they can expect the library to function.

I do really like this change, but would rather we defer it to the next major version if possible.

TeBoring:
When we add new features, we make sure new features don’t break existing
feature or we provide roll out plan. In this way, it’s not risky to use API
doc as source of truth. Actually, if users always conform to API doc, they
are less risky to be broken. The way to assume API should work act exactly
the way it is is actually the source of risk.

dwsupplee:
If there are users out there which *are* catching the current exception and handling it in some way, we will break their applications with this change. All of our generated clients in [google-cloud-php](https://github.com/googleapis/google-cloud-php) would be vulnerable to this. I don't feel strongly that our users utilizing the generated proto classes living under our [versioning semantics](https://github.com/googleapis/google-cloud-php#versioning) would know to look at the aforementioned API doc to determine undefined behavior.

We have a guarantee to our users of backward compatibility and we are not in a position to bump the major versions. Is this trade-off really worth the value it provides currently?

TeBoring:
The guarantee you mentioned is for documented API, not for undefined
behavior.
But I am ok to leave it until next release.
However, what do you plan to do to roll out this? Does delaying rolling out
this change anything?

dwsupplee:
> However, what do you plan to do to roll out this? Does delaying rolling out
this change anything?

What do you think of doing an evaluation for any other potential breaking changes in the generated code and batching this change set along with those? We could time/plan this out for when we are ready to introduce breaking changes to our GA libraries on google-cloud-php.

TeBoring:
SGTM in general. Still some questions:
1) What is the trigger to say we are ready to introduce these breaking changes? Does google-cloud-php have a list of features of want to have? We are not ready until we have finished code for all features in the list?
2) @bshaffer Do you prefer keeping this PR open or still leaving the original change in the dev branch?

dwsupplee:
> What is the trigger to say we are ready to introduce these breaking changes? Does google-cloud-php have a list of features of want to have? We are not ready until we have finished code for all features in the list?

These are really great questions. The list isn't ready yet, but is something we are working on.

bshaffer:
I agree we should use a major version for this. Along with the following fixes:

 1. [Changing exception messages](https://github.com/protocolbuffers/protobuf/issues/4473) 
 1. [Proper signature for checkEnum](https://github.com/protocolbuffers/protobuf/issues/3330)
 1. Removing all `Message_SubMessage.php` class files and remove the call to `class_alias` from the end of every `Message\SubMessage.php` class file.
 1. (possibly) Native types instead of protobuf messages

TeBoring:
> 2\. Proper signature for checkEnum
I think I mentioned in the issue it's an implementation detail of php implementation. I don't think we need to fix that.

TeBoring:
Could you send PR to us to fix those?

PR #5589, opened 2019-01-17
Fixed compile warnings for PHP extension.

haberman:


haberman:
I fixed the broken tests. I think the remaining failures are unrelated.

PR #5591, opened 2019-01-17
Fix field mask for php in json

TeBoring:


TeBoring:
@stanley-cheung could you help take a look?

TeBoring:
Thanks!

Issue #5565, opened 2019-01-09
GetErrno() on streams is not const


ivanp2015:
**What version of protobuf and what language are you using?**
Version: 3.6.1
Language: C++

**What operating system (Linux, Windows, ...) and version?**
all OS types

**What runtime / compiler are you using (e.g., python version or gcc version)**
all C++ compilers

**What did you do?**
Try to call GetErrno() on the const FileInputStream& and  const FileOutputStream&

**What did you expect to see**
It compiles

**What did you see instead?**
It does not compiles. GetErrno() does not modify any data, so can and must be declared const, but it is not for some reason.

**Anything else we should know about your project / environment**
none

ivanp2015:
Submitted PR #5595 for this. Please review.

PR #5595, opened 2019-01-17
Fix GetErrno() to be const noexcept

ivanp2015:
Fix for the #5565


acozzette:
@ivanp2015 Thanks for the pull request and sorry for taking such a long time to get back to you. This looks good to me but could you make the methods const without the noexcept? We don't really use exceptions at all, so that makes it hard to reason about noexcept. I think for now we should avoid using noexcept unless we decide to make a real effort to make the codebase exception-friendly.

ivanp2015:
Hi @acozzette, removed it.

acozzette:
Thanks @ivanp2015!

PR #5599, opened 2019-01-18
Fix c extension doesn' allow message reference in array

### LINKED ISSUE #5581: [PHP] Fatal error: Given value is not message.
Version: 3.6.1
Language: PHP

Linux Ubuntu 16.04.5 LTS. Kernel: 4.15.0-43
php extension compiled with gcc 5.4.0
**What did you do?**
I'm trying to use google/cloud-datastore package. The package suggests ext-grpc & ext-protobuf for better performance. I installed both from pecl, but with protobuf.so loaded, I consistently get:
`PHP Fatal error: Given value is not message. in vendor/google/cloud-datastore/src/V1/LookupRequest.php on line 129`. Please note that when disabling protobuf.so extension, the error disappears and code works as expected.
1. composer require "google/cloud-datastore:1.6.0"
2. run the following snippet:
```php

require_once __DIR__ . '/vendor/autoload.php';
// Intialize Datastore client with valid credentials.
$dataStoreClient = new \Google\Cloud\Datastore\DatastoreClient();
// Get Datastore entity key
$key = $dataStoreClient->key('<kind>', '<entity-id-value>');

// This call here produces: "Fatal error: Given value is not message."
$dataStoreClient->lookup($key);
```
Code works as is the case when google/protobuf package is used instead of c extension

**What did you see instead?**
PHP Fatal error: Given value is not message. in vendor/google/cloud-datastore/src/V1/LookupRequest.php on line 129

TeBoring:
Fix #5581

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -312,6 +312,11 @@ void Message_construct(zval* msg, zval* array_wrapper) {
        zend_hash_move_forward_ex(array, &pointer)) {
     zend_hash_get_current_key_zval_ex(array, &key, &pointer);
     field = upb_msgdef_ntofz(intern->descriptor->msgdef, Z_STRVAL_P(&key));
+#if PHP_MAJOR_VERSION >= 7
+    if (Z_ISREF_P((CACHED_VALUE*)value)) {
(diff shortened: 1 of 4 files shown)


TeBoring:
Thanks for the review!

PR #5604, opened 2019-01-19
PHP updates for new upb APIs

haberman:


haberman:
This passes all tests now and is ready for review.

haberman:
Hmm, the problem seems to arise from the fact that the new upb expects dependencies to be loaded before a given descriptor is loaded. Right now `well_known_test.php` is attempting to load a descriptor for proto `proto/test_import_descriptor_proto.proto`, which depends on `google/protobuf/descriptor.proto`, but it doesn't load `descriptor.proto` first like it should.

What's the best way to make sure descriptors are loaded in dependency order in PHP?

TeBoring:
Previously, descriptors are loaded lazily when messages are first created.
If you check the constructor of generated messages, you can find how descriptors are loaded.
I don't fully get how this has been changed.

TeBoring:
Please resolve the conflict.

TeBoring:
We also have some fixes in 3.7.x branch. Probably also need to sync.

haberman:
The Ruby builds are failing due to a known bug. PHP builds are passing, so I am going to merge.

Issue #4734, opened 2018-06-06
PHP FieldMasks json serialization is inconsistent for empty object


michaelbausor:
An empty FieldMask is serialized to `""`, but this does not deserialize to an empty FieldMask object. This PHPUnit test fails:

```php
public function testJsonEquality()
{
  $fm = new FieldMask();
  $json = $fm->serializeToJsonString();
  $fm2 = new FieldMask();
  $fm2->mergeFromJsonString($json);
  $this->assertEquals($fm, $fm2);
}
```

TeBoring:
I'll fix it.

TeBoring:
There is an ambiguity in FieldMask: both {} and {"path":""} (in text format) have the json representation of "".
Because {"path":""} is actually an invalid FieldMask, we have decided the decoded result of "" should be {} (in text format).

PR #5605, opened 2019-01-19
Fix empty FieldMask json encoding/decoding

TeBoring:


TeBoring:
Fix #4734

TeBoring:
@stanley-cheung could help take a look?

PR #5606, opened 2019-01-19
Added constructor to ArenaOptions to allow for struct-style initializer

GMScribe:
Default constructor and private attributes prevented this previously.

This is to allow full configuration of public attributes from an initializer-list within a class declaration.

### DIFF
--- a/src/google/protobuf/arena.h
+++ b/src/google/protobuf/arena.h
@@ -153,6 +153,23 @@ struct ArenaOptions {
         on_arena_destruction(NULL),
         on_arena_allocation(NULL) {}
 
+  ArenaOptions(size_t start_block_size,
+               size_t max_block_size,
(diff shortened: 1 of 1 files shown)


src/google/protobuf/arena.h
```diff
@@ -153,6 +153,23 @@ struct ArenaOptions {
         on_arena_destruction(NULL),
         on_arena_allocation(NULL) {}
 
+  ArenaOptions(size_t start_block_size,
+               size_t max_block_size,
+               char* initial_block,
+               size_t initial_block_size,
+               void* (*block_alloc)(size_t),
+               void (*block_dealloc)(void*, size_t),
```
acozzette: I think instead of a comma there needs to be a closing parenthesis at the end of this line.

PR #5609, opened 2019-01-21
[Python] Allow json_format.ParseDict to parse messages with extensions containing scalar types.

psobot:
This PR fixes an issue with `json_format.ParseDict` in Python, in which certain Protobuf definitions that contain extensions fail to be parsed. See [`protobuf-python-bug`](https://github.com/psobot/protobuf-python-bug) for a direct example of the bug outside of the `protobuf` codebase itself, but the gist of it is that any attempt to use `ParseDict` to deserialize a dict representation of a message containing a non-repeating extension field of scalar type currently fails. (I believe that's the simplest and smallest reproducible case of this bug that I could find, but my understanding of the problem could be incomplete.)

### DIFF
--- a/python/google/protobuf/internal/json_format_test.py
+++ b/python/google/protobuf/internal/json_format_test.py
@@ -202,6 +202,17 @@ class JsonFormatTest(JsonFormatBase):
     json_format.ParseDict(message_dict, parsed_message)
     self.assertEqual(message, parsed_message)
 
+  def testExtensionToDictAndBackWithScalar(self):
+    message = unittest_pb2.TestAllExtensions()
+    ext1 = unittest_pb2.TestNestedExtension.test
+    message.Extensions[ext1] = 'data'
+    message_dict = json_format.MessageToDict(
+        message
+    )
+    parsed_message = unittest_pb2.TestAllExtensions()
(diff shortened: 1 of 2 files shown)


python/google/protobuf/internal/json_format_test.py
```diff
@@ -202,6 +202,17 @@ def testExtensionToDictAndBack(self):
     json_format.ParseDict(message_dict, parsed_message)
     self.assertEqual(message, parsed_message)
 
+  def testExtensionToDictAndBackWithScalar(self):
+    message = unittest_mset_pb2.TestEmptyMessageContainer()
+    ext1 = unittest_mset_pb2.TestDummyMessageWithExtension.scalar_extension
```
anandolee: Test show TestDummyMessageWithExtension is not defined.
You could use TestAllExtensions for test
https://github.com/protocolbuffers/protobuf/blob/d9ccd0c0e6bbda9bf4476088eeb46b02d7dcd327/src/google/protobuf/unittest.proto#L221

psobot: Whoops, forgot to push all my commits. Thanks for catching that! I've updated this PR to use the existing `unittest_pb2.TestAllExtensions` instead.

PR #5613, opened 2019-01-22
Fix more issues for reference values

### LINKED ISSUE #5091: [PHP] "Given value cannot be converted to string" when setting RepeatedField value.
Version: v3.6.1
Language: PHP 7.2 (observed in all 7-series of PHP)

PHP C extension

**What did you do?**

Created an array of strings, accessed the values by reference, then assigned it to a protocol buffer value.

Successful assignment

**What did you see instead?**

"PHP Fatal error:  Given value cannot be converted to string."

This error does not appear to affect the protobuf pure PHP implementation. The error also does not seem to affect fields of type `string`.

A sample to reproduce this error using Cloud Firestore:

```php
<?php

use Google\Cloud\Firestore\V1beta1\BatchGetDocumentsRequest;

$documents = [
    'a', 'b', 'c'
];

function doThing(&$val) {}

// Either of these two lines cause the error. If they are removed, the error goes away.
doThing($documents[2]);
array_walk($documents, function (&$document) {});

$b = new BatchGetDocumentsRequest;
$b->setDocuments($documents);
```

TeBoring:
Fix #5091

### DIFF
--- a/php/tests/map_field_test.php
+++ b/php/tests/map_field_test.php
@@ -443,6 +443,25 @@ class MapFieldTest extends \PHPUnit\Framework\TestCase {
     }
 
     #########################################################
+    # Test reference in map
+    #########################################################
+
+    public function testMapElementIsReference()
+    {
+        // String elements
+        $values = ['a' => 'a'];
+        array_walk($values, function (&$value) {});
+        $m = new TestMessage();
+        $m->setMapStringString($values);
+
+        // Message elements
+        $values = [1 => new Sub()];
+        array_walk($values, function (&$value) {});
+        $m = new TestMessage();
+        $m->setMapInt32Message($values);
+    }
+
+    #########################################################
     # Test memory leak
     #########################################################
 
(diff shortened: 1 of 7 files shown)


TeBoring:
@jdpedrie, could you take a look?

TeBoring:
Done

php/tests/map_field_test.php
```diff
@@ -448,6 +448,24 @@ public function testMessage() {
 
     public function testMapElementIsReference()
     {
+        // Bool elements
```
jdpedrie: As a small note, tests like this may be easier to manage using a data provider, like this:

```php
public function mapElementIsReference()
{
    return [
        [
            [true => true],
            'setMapBoolBool'
        ], [
            [1 => 1],
            'setMapInt32Int32'
        ], [
            [1 => 1.0],
            'setMapInt32Double'
        ], [
            ['a' => 'a'],
            'setMapStringString'
        ], [
            [1 => new Sub()],
            'setMapInt32Message'
        ]
    ];
}

/**
 * @dataProvider mapElementIsReference
 */
public function testMapElementIsReference($values, $method)
{
    array_walk($values, function ($value) {});
    $m = new TestMessage();
    call_user_func($m, $method, $values);
}
```

It looks perfectly fine as is, but I often find data providers allow for more concise tests against a set of inputs.

TeBoring: Thanks for the suggestion. I will try it next time.

PR #5622, opened 2019-01-23
Parse unknown enum values if target is proto3

vitalybuka:
This patch is similar to the one for Java #1929
proto3 should support unknown enum values. WireFormat already writes and parse them and it's covered with tests. TextFormat however writes them successfully but was unable to parse.


acozzette:
Thanks @vitalybuka!

PR #5626, opened 2019-01-24
Update version to 3.7.0

haon4:


haon4:
PTAL. Tests are passing now.

PR #5628, opened 2019-01-24
Updated googletest submodule to the most recent upstream commit

acozzette:


haon4:
The mac OS C++ test is failing though. Can you double check?

acozzette:
I sent a PR to googletest to fix that test failure: https://github.com/google/googletest/pull/2079 Once that gets merged I can update this PR.

acozzette:
@haon4 I was able to get all the tests to pass with that googletest fix. I also updated the Linux distcheck test because I noticed it was passing when it should have been failing.

Issue #5619, opened 2019-01-23
PHP: Deprecated error is improperly thrown 


bshaffer:
In the `protobuf.so` PHP extension, when using the correct namespaces for sub-messages, a warning is thrown. Steps to reproduce:

1. Install a library with sub-messages. E.g. `google/common-protos`:
     ```
     composer require google/common-protos
    ```

2. Create a sub-message using a properly namespaced submessage classname:
    ```php
    require_once __DIR__ . '/vendor/autoload.php';

    // set a custom error handler so that the deprecated warnings are no longer suppressed
    set_error_handler(function ($errno, $errstr) {
        // Keep existing error handling
        printf("%s: %s\n", $errno, $errstr);
    }, E_ALL);

    // This should NOT result in a warning (but it does when the extension is enabled)
    $link = new Google\Rpc\Help\Link;
    ```

The output is the following:

```
16384: Google\Rpc\Help_Link is deprecated and will be removed in the next major release. Use Google\Rpc\Help\Link instead
16384: Google\Rpc\QuotaFailure_Violation is deprecated and will be removed in the next major release. Use Google\Rpc\QuotaFailure\Violation instead
16384: Google\Rpc\PreconditionFailure_Violation is deprecated and will be removed in the next major release. Use Google\Rpc\PreconditionFailure\Violation instead
16384: Google\Rpc\BadRequest_FieldViolation is deprecated and will be removed in the next major release. Use Google\Rpc\BadRequest\FieldViolation instead
```

The output _should_ be nothing, since the proper namespaced class name is used.

Similarly, if the wrong name is used, the output should print a single deprecated warning:

```php
require_once __DIR__ . '/vendor/autoload.php';

// set a custom error handler so that the deprecated warnings are no longer suppressed
set_error_handler(function ($errno, $errstr) {
    // Keep existing error handling
    printf("%s: %s\n", $errno, $errstr);
}, E_ALL);

// This SHOULD result in a warning, and it does when the extension is enabled, but additional errors are printed as well.
$link = new Google\Rpc\Help_Link;
```

Without the extension enabled:
```
16384: Google\Rpc\Help_Link is deprecated and will be removed in the next major release. Use Google\Rpc\Help\Link instead
```

With the extension  enabled:

```
16384: Google\Rpc\Help_Link is deprecated and will be removed in the next major release. Use Google\Rpc\Help\Link instead
[...]

PR #5629, opened 2019-01-24
Allow internalAddGeneratedFile to depend on new nested message name

TeBoring:
Previously, internalAddGeneratedFile has to depend on old non-nested name for sub-messages.
This creates a hard dependency on old generated code for compatibility usage.
If user's code has custome error handler, the deprecation warning in the old generated code
will also be thrown (even though users haven't explicitly depended on the old message name).

To fix this problem, this change added an additional flag in the generated code to tell run
time that it's safe to use new message name. In this way, internalAddGeneratedFile can safely
depend on new name and don't trigger unnecessary deprecation warning.

After this change is released, users need to use the new protoc to re-generate code.

### DIFF
--- a/php/ext/google/protobuf/def.c
+++ b/php/ext/google/protobuf/def.c
@@ -789,47 +789,44 @@ static bool is_reserved(const char *segment, int length) {
   return result;
 }
 
-static char* fill_prefix(const char *segment, int length,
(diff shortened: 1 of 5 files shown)


TeBoring:
Fix #5619

PR #5632, opened 2019-01-24
Fix checkUint32() for PHP

simPod:
You can quickly run the test here https://3v4l.org/TLTnK

The issue was that method `GPBUtil::checkUint32` was converting high values to negative ones. Eg. `4200010096` was being converted to `-94957199`. Also it was not handling overflow properly.

This patch respect unsigned behaviour so on overflow the number resets to 0 (can be seen in test).


simPod:
I signed it!

TeBoring:
The behavior of ```GPBUtil::checkUint32``` is intended to make it consistent for 32-bit machines.

simPod:
@TeBoring I can fix that for 32-bit machines too. Or do you propose other solution? It's obviously broken on 64-bit machines and doesn't behave like uint on both with current state.

TeBoring:
How do you plan to fix 32-bit? Internally, php uses int to represent integer type. Thus, large uint32 can only be represented as negative numbers?

TeBoring:
There are already users depend on the old behavior. What's the plan to make them using the new behavior without breaking there existing code?

simPod:
I don't know, I hoped that I'd get some proposition from maintainers because I'm not familiar with protobuf release politics and how fixed bugs that break BC are released. And from the conversation so far I don't know whether the obvious bug is admitted or not (?)  
I suppose it can be introduced with new minor version though I'd prefer it in patch version ofc.

But simply: The proper behaviour for dummy getter+setter pair is that **if I set value via setter, I expect to get the same value via getter.** And current behaviour is unpredictable and unexpected. 

In my case it caused a bug. I relied on the fact that if I serialize data to profobuf, send it to kafka and deserialize it back on consumption, I get the same values. So when I tried to insert those values to database to column with UInt32 type, the insert failed because data before serialization in uint32 range was suddenly negative for no reason.

TBH it seems like nobody noticed that uint32 doesn't accept values between 2^31 and 2^32 which seems to me like pretty large set of values. Something like half actually, ugh. So I dare to say that nobody would notice this fix anyway...

TeBoring:
Probably, we can introduce two API for uint32. ```getXXXAsInt32``` (existing behavior) and ```getXXXAsUInt32``` (new behavior).
Could you write a google doc to summarize this idea? We need internal design review for this.

simPod:
So we would have had 3 getters `getXXX()`, `getXXXAsInt32()` and `getXXXAsUInt32()`?

`getXXX()` and `getXXXAsUInt32()` would have had the same behaviour whilst `getXXXAsInt32()` would have the previous one?

--------------

I was wondering that if we would fix current behaviour and leave only one getter, it would be BC but not that severe. Imagine there are 2 main groups of users now:

1) Those who don't know about broken behaviour so they receive negative values for values above 2^31. But the don't do anything about it and don't know about it so they would benefit from the fix.
2) Those who know about it so they handle the value themselves. When `value from getter < 0` they convert it to the proper value. With the new behaviour `value from getter < 0` won't ever occur and their systems will still work.
3) but of course there might be others who use it in ways I can't imagine

TeBoring:
Because this is an API breaking change. We need at least one release to support all behaviors and change the getXXX behavior in the next release.
In the first release, we will ask existing users of getXXX on 32-bit machines to migrate to getXXXAsInt32.
In the second release, we will change the behavior of getXXX

TeBoring:
Anyway, we need a design doc. You could put all options into the doc.

simPod:
I have put something together at: https://docs.google.com/document/d/1Yd5QliMi4MBye5lWZ8OB4D_PXA2Vw4DHnkKH7RHiCZQ/edit?usp=sharing

Copy paste:

# Proper implementation of UInt32 behaviour for PHP Protobuf
## Current state

Protobuf in PHP uses `GPBUtil::checkUint32` function for converting passed value into UInt32. As there is no such data type in PHP it’s represented through integer value. There are two strategies implemented and both have flaws. 

There is a strategy for 32-bit systems. As max integer on 32-bit systems is 2^31-1, the strategy for these takes passed value and subtracts 4294967296 if value is > 2^31-1 otherwise the value is unchanged. It’s converted to integer then.

```php
                if (bccomp($var, 0x7FFFFFFF) > 0) {	
                    $var = bcsub($var, "4294967296");	
                }	
                $var = (int) $var;
```

The problem is that for example: for input value `2^31` the output value is `-2^31`. That is not the same value so it causes unexpected result when getting the value afterwards.

For 64-bit systems it’s pretty much the same

`$var |= ((-(($var >> 31) & 0x1)) & ~0xFFFFFFFF)`

For every input value > 2^31 the output value is also negative.
## Proposed fix

For 64-bit systems the proposed fix is

`$var = ((int) $var) % 4294967296`

That results in proper UInt32 behaviour, not storing any negative values etc.

For 32-bit systems it’s proposed the same, however represented through string

`$var = bcmod((string) $var, "4294967296")`

1) PHP BC Math extension works with string arguments 
2) it can be expected that numeric but string input value is passed. That’s the only way to represent values > 2^31-1 on 32-bit systems while keeping the value intact.

## Backward Compatibility
For keeping backward compatibility in terms of Semver, the proposed steps for each version are
1. In next version deprecate current `getXXX()` getters for UInt32 fields (old behaviour) and introduce new `getXXXAsInt32()` (new behaviour)
2. Then change behaviour of `getXXX()` to new behaviour. Consider it dropping of deprecated method `getXXX()` and introducing new method `getXXX()`

TeBoring:
Thanks for the proposal! I would ask other engineers to review it internally.

simPod:
@TeBoring I thought you would ask the engineers but maybe you meant that I should ask them? Thanks!

fowles:
@haberman do you mind taking a look?

PR #5633, opened 2019-01-25
Python/C++: Add tests to ensure that AllowOversizeProtos also handles deeper levels of recursion.

dtanabe:
Add tests to ensure that AllowOversizeProtos also handles deeper levels of recursion.

EDIT: Was originally intending to try to fix a problem where `SetAllowOversizeProtos` was supposed to set the maximum Protobuf message size _and_ recursion depth arbitrarily high, but under most non-trivial circumstances the recursion depth setting didn't actually seem to be affected. This happened to be fixed as part of 3.7.0 and I erroneously thought my changes fixed the problem.

Included a test case which would have failed without the accompanying code change that intentionally creates a deeper message that the C++ parser implementation for Python would normally allow.

### DIFF
--- a/python/google/protobuf/internal/message_test.py
+++ b/python/google/protobuf/internal/message_test.py
@@ -81,6 +81,7 @@ from google.protobuf import text_format
 from google.protobuf.internal import api_implementation
 from google.protobuf.internal import encoder
 from google.protobuf.internal import more_extensions_pb2
+from google.protobuf.internal import more_messages_pb2
 from google.protobuf.internal import packed_field_test_pb2
 from google.protobuf.internal import test_util
 from google.protobuf.internal import testing_refleaks
@@ -2527,9 +2528,8 @@ class PackedFieldTest(unittest.TestCase):
     self.assertEqual(golden_data, message.SerializeToString())
 
 
-@unittest.skipIf(api_implementation.Type() != 'cpp' or
-                 sys.version_info < (2, 7),
-                 'explicit tests of the C++ implementation for PY27 and above')
+@unittest.skipIf(api_implementation.Type() != 'cpp',
+                 'explicit tests of the C++ implementation')
 @testing_refleaks.TestCase
 class OversizeProtosTest(unittest.TestCase):
 
@@ -2588,6 +2588,36 @@ class OversizeProtosTest(unittest.TestCase):
     q = self.proto_cls()
     q.ParseFromString(self.p_serialized)
     self.assertEqual(self.p.field.payload, q.field.payload)
+
+
+@unittest.skipIf(api_implementation.Type() != 'cpp',
+                 'explicit tests of the C++ implementation')
+class DeepNestedProtosTest(BaseTestCase):
+
+  def setUp(self):
+    # create a Tree whose depth exceeds the default allowed recursive depth
+    # when reading
+    root = more_messages_pb2.Node(value=1)
+    for x in range(199):
+        root = more_messages_pb2.Node(tree=more_messages_pb2.Tree(right=root))
+
+    self.p = more_messages_pb2.Tree(right=root)
+    self.p_serialized = self.p.SerializeToString()
+
+  def testAssertOversizeProto(self):
+    from google.protobuf.pyext._message import SetAllowOversizeProtos
+    SetAllowOversizeProtos(False)
+    q = more_messages_pb2.Tree()
+    try:
+      q.ParseFromString(self.p_serialized)
+    except message.DecodeError as e:
+      self.assertEqual(str(e), 'Error parsing message')
+
+  def testSucceedOversizeProto(self):
+    from google.protobuf.pyext._message import SetAllowOversizeProtos
+    SetAllowOversizeProtos(True)
+    q = more_messages_pb2.Tree()
+    q.ParseFromString(self.p_serialized)
 
 if __name__ == '__main__':
   unittest.main()
--- a/python/google/protobuf/internal/more_messages.proto
+++ b/python/google/protobuf/internal/more_messages.proto
@@ -358,3 +358,15 @@ message LotsNestedMessage {
   message B254 {}
   message B255 {}
 }
+
+message Tree {
+  optional Node left = 1;
+  optional Node right = 2;
+}
+
+message Node {
+  oneof Child {
+    Tree tree = 1;
+    sint64 value = 2;
+  }
+}


dtanabe:
I signed it!

dtanabe:
Hey…anything I can do to help speed this along? I could really use this fix in a released version; the performance difference between the C++ version and the native Python version is pretty sizable and this bug prevents me from using the native C++ version.

python/google/protobuf/internal/message_test.py
```diff
@@ -2441,5 +2442,36 @@ def testSucceedOversizeProto(self):
...
+    for x in range(199):
+        root = recursive_test_pb2.Node(tree=recursive_test_pb2.Tree(right=root))
+
+    self.p = recursive_test_pb2.Tree(right=root)
+    self.p_serialized = self.p.SerializeToString()
+
+  def testAssertOversizeProto(self):
+    from google.protobuf.pyext._message import SetAllowOversizeProtos
+    SetAllowOversizeProtos(False)
+    q = recursive_test_pb2.Tree
```
amauryfa: The test won't pass without parentheses:
```suggestion
    q = recursive_test_pb2.Tree()
```

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -196,6 +196,10 @@ bool Reparse(
   from.SerializeToString(&serialized);
   io::CodedInputStream input(
       reinterpret_cast<const uint8*>(serialized.c_str()), serialized.size());
+  if (cmessage::InternalGetAllowOversizeProtos()) {
+    input.SetTotalBytesLimit(INT_MAX, INT_MAX);
+    input.SetRecursionLimit(INT_MAX);
+  }
```
amauryfa: This code is executed when parsing *Descriptor options*, which are not usually that large because they appear in .proto source files. Why did you modify this part?

dtanabe:
@amauryfa I have a large .proto that is fairly deeply nested (can't change the .proto, unfortunately). The C++ implementation consistently crashes (even with calling `SetAllowOversizeProtos()`), whereas if I force it to use the Python implementation, I can successfully parse deeply nested objects (but only by increasing the recursion depth).

Definitely surprised to hear that the tests pass for you locally; FWIW I'm using Mac OS X 10.14 (not sure that's relevant or not).

python/google/protobuf/pyext/descriptor.cc
dtanabe: Instances of `CodedInputStream` only appear to be constructed in two places of the codebase. I didn't have a great way of getting a stack trace, but this was ultimately the fix that stopped my use case from crashing.

dtanabe:
I also had to hack around my local environment to get the relevant tests to even be _run_ on Python 3 on Mac (see https://github.com/protocolbuffers/protobuf/blob/master/tests.sh#L236). Would agree that it would be bizarre for this code change to affect the behavior of OS X only specifically, but on the other hand this is the only way I could get the test to pass on OS X (with parentheses of course 😅).

dtanabe:
Ok…on further review it looks like other changes on `master` must have landed that actually fixed the problem. I tried 3.7.0rc2 and it seems to work fine for me. When I was building locally, I was comparing 3.6.1 released with my changes to `master` _without_ also trying `master` unchanged.

If you think the test is useful, I'll clean up the PR to remove my changes and just leave the test in place. But otherwise thanks for fixing my bug before you knew I had it!

amauryfa:
Yes, we could keep the tests, I don't see anything that exercise deeply nested protos. @anandolee what do you think?

anandolee:
Keep the tests make sense to me

python/google/protobuf/internal/message_test.py
```diff
@@ -2441,5 +2442,36 @@ def testSucceedOversizeProto(self):
     q.ParseFromString(self.p_serialized)
     self.assertEqual(self.p.field.payload, q.field.payload)
 
+
+@unittest.skipIf(api_implementation.Type() != 'cpp' or
+                 sys.version_info < (2, 7),
```
anandolee: Protobuf python only support python 2.7 and above, no need to do the filter here

dtanabe: Ah…I copied this from https://github.com/protocolbuffers/protobuf/blob/d1e72c11e676ddb0aa852fbb3e3f968828a95381/python/google/protobuf/internal/message_test.py#L2385; should I also remove the explicit version check from the other test too?

python/google/protobuf/internal/message_test.py
```diff
@@ -2441,5 +2442,36 @@ def testSucceedOversizeProto(self):
     q.ParseFromString(self.p_serialized)
     self.assertEqual(self.p.field.payload, q.field.payload)
 
+
+@unittest.skipIf(api_implementation.Type() != 'cpp' or
```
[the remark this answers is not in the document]
dtanabe: In pure Python, modifying the recursion depth is strictly done with https://docs.python.org/3/library/sys.html#sys.setrecursionlimit; the C++ implementation's max recursion depth is set with `SetOversizeProtos`, ultimately modifying `CodedInputStream`.

The test here is really explicitly for, if you _want_ to use the C++ bindings, that deeply recursive messages are still faithfully processed properly. (This was the original problem I thought I was fixing with the original changes, and now realize was fixed with other people's changes.) The pure Python implementation definitely allows for deeper messages without any special allowances other than changing Python's maximum recursion limit, as that was the workaround I had employed before switching to 3.7.0rc2.

anandolee:
dtanabe@ are you still working on this?

dtanabe:
@anandolee Sorry for the long delay—PR updated to reflect all your comments; additionally I rebased the whole branch as I was quite a bit behind upstream.

python/google/protobuf/internal/message_test.py
```diff
@@ -2589,5 +2589,35 @@ def testSucceedOversizeProto(self):
     q.ParseFromString(self.p_serialized)
     self.assertEqual(self.p.field.payload, q.field.payload)
 
+
+@unittest.skipIf(api_implementation.Type() != 'cpp',
+                 'explicit tests of the C++ implementation for PY27 and above')
```
anandolee: 'explicit tests of the C++ implementation'

dtanabe: Amended.

acozzette:
The test runs are showing this error: `NameError: name 'BaseTestCase' is not defined`

acozzette:
I'm going to close this since there has been no activity in a while, but feel free to comment here if you would like to reopen it and keep working on it.

PR #5636, opened 2019-01-25
Fix python release build.

haon4:


kokoro/release/python/linux/build_artifacts.sh
```diff
@@ -7,11 +7,13 @@ pushd $(dirname $0)/../../../..
 
 export REPO_DIR=protobuf
 export BUILD_VERSION=`grep -i "version" python/google/protobuf/__init__.py | grep -o "'.*'" | tr -d "'"`
-export BUILD_COMMIT=v$BUILD_VERSION
+export BUILD_COMMIT=master
```
TeBoring: If we pin to master, we need to change it when we work on release branch?

kokoro/release/python/linux/config.sh
```diff
@@ -6,6 +6,12 @@ function pre_build {
     # Runs in the root directory of this repository.
     pushd protobuf
 
+    if [ -n "$IS_OSX" ]; then
```
TeBoring: We probably should file a bug to multibuild and kokoro to let them include these.

kokoro/release/python/linux/config.sh
```diff
@@ -6,6 +6,12 @@ function pre_build {
...
+      brew install libatomic
```
TeBoring: Shall we create a separate file in kokoro/release/python/macos?

kokoro/release/python/linux/config.sh
haon4: But libatomic is something that we include recently. I assume not everyone need it, right?

kokoro/release/python/macos/config.sh
```diff
@@ -0,0 +1,50 @@
...
+# Test for OSX with [ -n "$IS_OSX" ]
+
+function pre_build {
+    # Any stuff that you need to do before you start building the wheels
+    # Runs in the root directory of this repository.
+    pushd protobuf
+
+    export CC=gcc
+    export CXX=g++
+    # Build protoc
```
TeBoring: No need to brew install now?

kokoro/release/python/linux/build_artifacts.sh
haon4: Changed to extract from the KOKORO_JOB_NAME name instead.

kokoro/release/python/macos/config.sh
haon4: Still trying to verify that this actually works. It is still running...

kokoro/release/python/macos/build_artifacts.sh
```diff
@@ -7,17 +7,23 @@ pushd $(dirname $0)/../../../..
...
 export TRAVIS_OS_NAME="osx"
 
+rm -rf artifacts/
+rm -rf multibuild/
 mkdir artifacts
 export ARTIFACT_DIR=$(pwd)/artifacts
 
 git clone https://github.com/matthew-brett/multibuild.git
-cp kokoro/release/python/linux/config.sh config.sh
+cp kokoro/release/python/macos/config.sh config.sh
```
TeBoring: I think you don't need to copy. It's already in the place you want to move.

haon4: Actually no. On L6, we change the directory to the repo root, so we are copying it to the repo's root directory.

haon4:
Looks like setting the -std=c++14 flag on Mac OS resolve the issue.

PR #5644, opened 2019-01-28
Convert integer to string if field is string field in json

### LINKED ISSUE #4765: [PHP] mergeFromJsonString can not unserialize maps with numeric keys
With the following proto:

``` proto
syntax = "proto3";

package test;

message Hello {
    map<string, string> m = 1;
};
```

The PHP implementation of protobuf has troubles decoding the following JSON message:

```
{"m":{"foo":"bar","123":"42"}}
```

The error is:

```
PHP Fatal error:  Uncaught Google\Protobuf\Internal\GPBDecodeException: Error occurred during parsing: String field only accepts string value in php/src/Google/Protobuf/Internal/Message.php:836
Stack trace:
#0 php/src/Google/Protobuf/Internal/Message.php(1116): Google\Protobuf\Internal\Message->convertJsonValueToProtoValue(123, Object(Google\Protobuf\Internal\FieldDescriptor), true)
#1 php/src/Google/Protobuf/Internal/Message.php(1086): Google\Protobuf\Internal\Message->mergeFromArrayJsonImpl(Array)
#2 php/src/Google/Protobuf/Internal/Message.php(1168): Google\Protobuf\Internal\Message->mergeFromJsonArray(Array)
#3 php/src/Google/Protobuf/Internal/Message.php(742): Google\Protobuf\Internal\Message->parseFromJsonStream(Object(Google\Protobuf\Internal\RawInputStream))
#4 bug/test.php(17): Google\Protobuf\Internal\Message->mergeFromJsonString('{"m":{"foo":"ba...')
#5 {main}
  thrown in php/src/Google/Protobuf/Internal/Message.php on line 836
```

The cause is that we decode the [JSON as array](https://github.com/google/protobuf/blob/f1911f37f817fb2dddbb07478bdcda526488fc55/php/src/Google/Protobuf/Internal/Message.php#L1159) before parsing it, and that PHP arrays always store numeric keys as integers, and everything else as strings. So, json decoding `{"123":"42"}` results in `[123 => "42"]` rather than `["123" => "42"]` (which is impossible in PHP).

TeBoring:
json_decode automatically convert numeric string to integer, so that
we need to convert it back. However, this will suceed to parse invalid
json data with string field set as integer even though it should have failed.
Because, the failure case is less often than the succeeding case, we decided
to make this change. Users should make sure their data don't use integer for
string fields by themselves.

Fix #4765

### DIFF
--- a/conformance/failure_list_php.txt
+++ b/conformance/failure_list_php.txt
@@ -13,6 +13,8 @@ Required.Proto3.JsonInput.FloatFieldTooSmall
 Required.Proto3.JsonInput.DoubleFieldTooSmall
 Required.Proto3.JsonInput.Int32FieldNotInteger
 Required.Proto3.JsonInput.Int64FieldNotInteger
+Required.Proto3.JsonInput.RepeatedFieldWrongElementTypeExpectingStringsGotInt
+Required.Proto3.JsonInput.StringFieldNotAString
 Required.Proto3.JsonInput.Uint32FieldNotInteger
 Required.Proto3.JsonInput.Uint64FieldNotInteger
 Required.Proto3.JsonInput.Int32FieldLeadingSpace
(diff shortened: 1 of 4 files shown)


TeBoring:
@michaelbausor PTAL

Issue #5384, opened 2018-11-20
build protoc with Link error 2005 in vs2017


lingerer:
**What version of protobuf and what language are you using?**
Version: master/v3.61
**What operating system (Linux, Windows, ...) and version?**
windows 10
**What runtime / compiler are you using (e.g., python version or gcc version)**
vs 2017
**What did you do?**
using cmake to build,and the protoc will end with link error 2005
this will be solve by switch libprotoc and libprotobuf in link position.
So my suggestion is :
Change protoc.cmake line 6 
target_link_libraries(protoc libprotobuf libprotoc)
to 
target_link_libraries(protoc libprotoc libprotobuf)
**Anything else we should know about your project / environment**

acozzette:
@lingerer Do you want to send us a pull request with this fix?

lingerer:
sorry,github is block here.I can't send a fix through clone

acozzette:
This has now been fixed with #5648.

PR #5647, opened 2019-01-29
Swapped libprotoc and libprotobuf to fix LNK2005 in MSVC

MichaelMitchellM:
Fixed problem stated in [#5384](https://github.com/protocolbuffers/protobuf/issues/5384).

The ordering of libprotobuf and libprotoc caused LNK2005 errors in MSVC builds.
The solution seems to be swapping them in cmake/protoc.cmake

Prior to this change, using cmake to create a Visual Studio 2017 x86/x64 project will succeed, but when running `cmake --build` or building in VS, you get ~193 LNK 2005 errors complaining that a symbol has already been defined in libprotobuf.

### DIFF
--- a/cmake/protoc.cmake
+++ b/cmake/protoc.cmake
@@ -9,7 +9,7 @@ set(protoc_rc_files
 endif()
 
 add_executable(protoc ${protoc_files} ${protoc_rc_files})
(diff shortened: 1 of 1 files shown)


MichaelMitchellM:
I signed it!

PR #5648, opened 2019-01-29
Swapped libprotoc and libprotobuf

MichaelMitchellM:
libprotobuf being targeted before libprotoc was causing Visual Studio to fail in linking protoc.

This fix addresses the issue in [#5384](https://github.com/protocolbuffers/protobuf/issues/5384).


acozzette:
Thanks @MichaelMitchellM!

PR #5651, opened 2019-01-29
Fix version.rc for linux

gelldur:
Add missing if for MSVC


gelldur:
I signed it!

acozzette:
@gelldur Thank you for the pull request but at this time we're not accepting any more changes to the 3.6.x branch because we are not going to do any more 3.6 releases. We're almost ready to release 3.7.0 and it already includes this fix if I'm not mistaken.

PR #5652, opened 2019-01-30
Fixes warnings in https://github.com/google/oss-fuzz/issues/864#issue…

bshastry:
…comment-458596374

This PR fixes a bunch of "unused variable warnings" that were emitted by the compiler (clang) when another project that included protobuf headers had a stricter compiler setting (e.g., `-Werror`) than the compiler setting used for protobuf itself.

This fix ensures that projects that depend on protobuf AND have a strict compiler setting build cleanly :)

### DIFF
--- a/src/google/protobuf/util/field_comparator.h
+++ b/src/google/protobuf/util/field_comparator.h
@@ -182,7 +182,7 @@ class PROTOBUF_EXPORT DefaultFieldComparator : public FieldComparator {
   // basic types (instead of submessages). They return true on success. One
   // can use ResultFromBoolean() to convert that boolean to a ComparisonResult
   // value.
-  bool CompareBool(const FieldDescriptor& field, bool value_1, bool value_2) {
+  bool CompareBool(const FieldDescriptor& /* unused */, bool value_1, bool value_2) {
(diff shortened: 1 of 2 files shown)


src/google/protobuf/util/field_comparator.h
```diff
@@ -183,6 +183,7 @@ class PROTOBUF_EXPORT DefaultFieldComparator : public FieldComparator {
   // can use ResultFromBoolean() to convert that boolean to a ComparisonResult
   // value.
   bool CompareBool(const FieldDescriptor& field, bool value_1, bool value_2) {
+    (void)field;
```
acozzette: I think a slightly more succinct way of accomplishing this is just to avoid giving the parameter a name at all, so for example the function signature could like this:

`bool CompareBool(const FieldDescriptor& /* unused */, bool value_1, bool value_2) {`

Could you try doing it that way in this file? (I know in `arenastring.h` it would not work very well because of the `ifdef`).

bshastry: done :)

acozzette:
Thanks @bshastry!

Issue #5655, opened 2019-01-31
More descriptive error messages in Ruby


jbolinger:
**What language does this apply to?**
Ruby

**Describe the problem you are trying to solve.**
If a proto message type is created with the wrong arguments an error message is thrown, but it is vague. For example if a number is given for a string field the error is something like:
`'initialize': Invalid argument for string field. (TypeError)`

**Describe the solution you'd like**
It would be helpful if the error message mentioned the name of the field and the expected/given types. For example, something like `TypeError: my_field expects a string but was given 12345.`

**Describe alternatives you've considered**

**Additional context**

TeBoring:
Would you mind sending a PR to us?

jbolinger:
Sure, took a shot at it in #5659

jbolinger:
This was added in #5659, closing.

PR #5659, opened 2019-01-31
Add more descriptive error messages to init methods in Ruby

jbolinger:
This change adds more details (field name and type) to the error messages in Ruby when an object is created with invalid arguments (#5655). I think a stringified version of the value might also be helpful, but I didn't see an obvious way to do that.


TeBoring:
You need to add new file to src/Makefile.am

jbolinger:
Ah, thanks @TeBoring I've updated `Makefile.am`.

PR #5661, opened 2019-02-01
Add stack overflow protection for text format

vitalybuka:
It already exists for wire and json formats


vitalybuka:
Not sure if API to change it is needed

acozzette:
OK, that makes sense and sounds like a good use case. Thanks for fuzzing our code!

PR #5667, opened 2019-02-01
Add several fixes for python toolchain

TeBoring:


cmake/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ string(REGEX REPLACE        "${protobuf_AC_INIT_REGEX}" "\\2"
 string(REGEX REPLACE        "${protobuf_AC_INIT_REGEX}" "\\3"
     protobuf_CONTACT        "${protobuf_AC_INIT_LINE}")
 # Parse version tweaks
-set(protobuf_VERSION_REGEX "^([0-9]+)\\.([0-9]+)\\.([0-9]+)-?(.*)$")
+set(protobuf_VERSION_REGEX "^([0-9]+)\\.([0-9]+)\\.([0-9]+)(-rc)?(.*)$")
```
acozzette: If we add `(-rc)` here in parentheses then doesn't that mean `\\4` below will match `-rc`? It seems to me like we need to replace `\\4` below with `\\5`. Also I wonder if we need another dash (i.e. `(-rc-)` since the current version is 3.7.0-rc-2.

Issue #4598, opened 2018-05-07
Cannot compare enum value with enum definition


mainameiz:
Assuming we have the following proto:

```
syntax = "proto3";

message Resource {
  enum Type {
    UNDEFINED = 0;
    BOOK = 1;
  }
  Type type = 1;
}
```

The following ruby code does not work (but as it seems to me it should work)

```ruby
resource = Resource.new(type: Resource::Type::BOOK)
case resource.type # => :BOOK
when Resource::Type::BOOK # => 1
  ... do something
end
```

It does not work because `resource.type` has type of ruby Symbol (its value `:BOOK`) and `Resource::Type::BOOK` equals to `1`. So it is not possible to use this well known code-pattern with protobuf enums.

Can you fix this, please? Or is there any workaround?

blowmage:
FWIW, the workaround I've been using is `Resource::Type.get_const :BOOK`.

mainameiz:
First of all. I made a mistake in my example (it is fixed now).

Then
In ruby `Resource::Type.get_const :BOOK` is the same as `Resource::Type::BOOK`.

```ruby
[26] pry(main)> Resource::Type.const_get(:BOOK)
=> 1
[27] pry(main)> Resource::Type::BOOK
=> 1
```

So, it is not a solution.

blowmage:
```ruby
resource = Resource.new(type: Resource::Type::BOOK)

case Resource::Type.const_get(resource.type)
when Resource::Type::BOOK
  ... do something
end
```

mcloonan:
It's nice to have a workaround, but it's still pretty awkward. It would be nice to have a more discoverable/intuitive built-in solution.

jbolinger:
Would generating an extra property like `resource.type_const` be an option? It's not ideal but it's probably the only way it could be done without breaking the existing behavior.

mainameiz:
> Would generating an extra property like resource.type_const be an option?

Yes

jbolinger:
@TeBoring I tried to come up with something for this in #5670. It's just a POC, but what do you think about doing something along these lines?

haon4:
It seems like the fix has been merged. Closing this.

PR #5670, opened 2019-02-02
Generate extra enum method helpers for Ruby

jbolinger:
This is an example of a potential solution to #4598. It causes the generator to create an extra method for types with enum fields in the following form:

```ruby
class MyMessage
  def type_const
    Google::Example::Resource::Type.const_get(type)
  end
end
```

`type` is the field name of an enum type in the `MyMessage` proto message in this example.

I'm not sure if there's a better way to add something like this via the DSL (I couldn't find any docs).


TeBoring:
Tests still failing

jbolinger:
Opps, I missed those tests but I've updated them now.

The generated code could be simplified a bit here, but I wanted to get your feedback on the approach first.

TeBoring:
Could you add a test to use the new method?

jbolinger:
Certainly! I've added one.

jbolinger:
@TeBoring @haberman what do you think about the way these additional helper methods are generated? I was thinking of taking the same approach for the `*_as_value` methods that we've been discussing.

TeBoring:
The approach looks good to me. I am just not sure whether the name is idiomatic to ruby users.
I also want @haberman to take a look to make sure we want this API.

src/google/protobuf/compiler/ruby/ruby_generated_code_pb.rb
```diff
@@ -70,6 +70,17 @@ module B
     module C
       TestMessage = Google::Protobuf::DescriptorPool.generated_pool.lookup("A.B.C.TestMessage").msgclass
       TestMessage::NestedMessage = Google::Protobuf::DescriptorPool.generated_pool.lookup("A.B.C.TestMessage.NestedMessage").msgclass
+      class TestMessage
+        def optional_enum_const
```
blowmage: I am not a fan of adding these methods by altering the generated Ruby code. I would rather see these new methods added to the object that is returned when calling `lookup("A.B.C.TestMessage").msgclass`. That way, Ruby files that were previously generated will get these new methods by upgrading to the latest google-protobuf gem, without having to regenerate the the Ruby files.

jbolinger:
I updated the implementation per @blowmage's suggestion. The method functionality remains the same but the generated code is no longer changed. I also added some more tests for various kinds of field types.

ruby/tests/common_tests.rb
```diff
@@ -706,6 +706,50 @@ def test_enum_lookup
...
+    assert_equal proto_module::TestEnum.const_get(:C), m.optional_enum_const
+  end
+
+  def test_enum_getter
+    m = proto_module::Enumer.new(:optional_enum => :B, :repeated_enum => [:A, :C])
+
+    assert_equal :B, m.optional_enum
+    assert_equal 2, m.optional_enum_const
+    assert_equal [:A, :C], m.repeated_enum
+    assert_equal [1, 3], m.repeated_enum_const
```
blowmage: Is this the same as the following?

```ruby
assert_equal [proto_module::TestEnum::A, proto_module::TestEnum::C], m.repeated_enum_const
```

If so, would it be possible to show the constants as well as the constant's values?

jbolinger: Yes, they are the same and it would be clearer to show that. I'll add both to the tests.

blowmage: :heart:

ruby/ext/google/protobuf_c/message.c
```diff
@@ -231,13 +255,13 @@ VALUE Message_method_missing(int argc, VALUE* argv, VALUE _self) {
...
       if (oneof_field != NULL) {
-	layout_clear(self->descriptor->layout, Message_data(self), oneof_field);
+        layout_clear(self->descriptor->layout, Message_data(self), oneof_field);
       }
       return Qnil;
     } else {
       // METHOD_ACCESSOR
       return oneof_field == NULL ? Qnil :
-	ID2SYM(rb_intern(upb_fielddef_name(oneof_field)));
+       ID2SYM(rb_intern(upb_fielddef_name(oneof_field)));
```
TeBoring: Is the indent correct?

jbolinger: fixed.

ruby/ext/google/protobuf_c/message.c
```diff
@@ -153,9 +154,32 @@ static int extract_method_call(VALUE method_name, MessageHeader* self,
     accessor_type = METHOD_GETTER;
   }
 
+  bool has_field = upb_msgdef_lookupname(self->descriptor->msgdef, name, name_len,
+			                                   &test_f, &test_o);
+
+  // Look for enum accessor of the form <enum_name>_const
+  if (!has_field && accessor_type == METHOD_GETTER && name_len > 6) {
```
TeBoring: Is this checking too loose?
What if another field's name is ```foo_XXXXX```?

TeBoring: Please also test this case.

jbolinger: I think it's okay because `METHOD_UNKNOWN` will still be returned unless the `enum_name` is found in the innermost check. I added that as a test case - is that what you were concerned about?

TeBoring: I think it matters.
Suppose foo is the enum field, if you call m.foo_abcde, you don't return 
METHOD_UNKNOWN.

ruby/tests/common_tests.rb
```diff
@@ -722,6 +722,16 @@ def test_enum_const_get_helpers
     m = proto_module::TestMessage.new({optional_enum: proto_module::TestEnum::C})
     assert_equal proto_module::TestEnum::C, m.optional_enum_const
     assert_equal proto_module::TestEnum.const_get(:C), m.optional_enum_const
+
+    m = proto_module::TestMessage2.new({foo: 2})
```
TeBoring: Foo needs to be an enum field.

ruby/ext/google/protobuf_c/message.c
jbolinger: Oh, I see. Good catch. I've added an additional check for the `_const` suffix.

ruby/tests/common_tests.rb
jbolinger: Added a new test that uses an enum field.

TeBoring:
please sync

jbolinger:
@TeBoring sync'd. Thanks for the help!

PR #5698, opened 2019-02-08
Use cached value for type safety check during unpacking of Any message

yvanin:
This simple performance optimization allows to skip the overhead of reflection (in the `is()` method [here](https://github.com/protocolbuffers/protobuf/blob/de3be6c225cdb92548648cb899391ef5368051fa/src/google/protobuf/compiler/java/java_message.cc#L1449)) when the message has been already unpacked once.


src/google/protobuf/compiler/java/java_message.cc
```diff
@@ -1446,13 +1446,17 @@ void ImmutableMessageGenerator::GenerateAnyMethods(io::Printer* printer) {
     "public <T extends com.google.protobuf.Message> T unpack(\n"
     "    java.lang.Class<T> clazz)\n"
     "    throws com.google.protobuf.InvalidProtocolBufferException {\n"
-    "  if (!is(clazz)) {\n"
+    "  boolean invalidClazz = false;\n"
```
TeBoring: What if previously an instance of the same class but different value was unpacked?

yvanin: Old implementation would still return `cachedUnpackValue` in this hypothetical case, this implementation retains this behavior.

Issue #5318, opened 2018-11-06
[Javascript] Incorrectly generated import paths when using `google/api` or `google/type`


dominikeinkemmer:
/cc @zchee

**What version of protobuf and what language are you using?**
Version: v3.6.0
Language: Javascript

**What operating system (Linux, Windows, ...) and version?**
Docker Image: node:9

**What runtime / compiler are you using (e.g., python version or gcc version)**

**What did you do?**
When using imports for `google/api/..` or `google/type/...` the imports generated for javascript resolve to incorrect paths
Example:
```
import "google/api/annotations.proto";
import "google/type/date.proto";
```

**What did you expect to see**
Correctly resolved import paths

**What did you see instead?**
Wrongly resolved import paths:
generated output (these directories don't exist):
```javascript
var google_api_annotations_pb = require('../../../../../../google/api/annotations_pb.js');
var google_type_date_pb = require('../../../../../../google/type/date_pb.js');
```

We resolved these imports by generating these files ourselves (from https://github.com/googleapis/googleapis repository) and copying them to the correct folder but inside of `annotations_pb.js` another incorrect import is generated so it seems that this problem is carrying through to any imports down the road:

```javascript
// inside of annotations_pb.js 
var google_api_http_pb = require('../../../../../../google/api/http_pb.js');
```

We wondered how we could resolve this? Generating the files ourselves and copying them is totally ok but the wrongly generated imports inside these files are an issue we can not resolve on our end unfortunately, so any help is appreciated.

**Anything else we should know about your project / environment**
We use `grpc_tools_node_protoc` and `grpc_tools_node_protoc_ts` plugins to generate typescript and grpc code.

dominikeinkemmer:
I checked the code for the `js_generator.cc` regarding the import paths. I'm not familiar with `C` code but if I understood this part correctly then:
(https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/js/js_generator.cc#L174)

- it first checks for `google/protobuf` and resovles these imports to `google-protobuf/`
- it checks how many `/` are in the import string and resovles it to `./` for 0 and otherwise aggregates as many `../` as there are `/` in the import name.

As I said, I'm not familiar with `C`, but is there any chance that inside this method the import path could be compared with the path of the file that has this import?

[...]

PR #5703, opened 2019-02-09
fix import generation for javacsript

dominikeinkemmer:
As explained in this issue: https://github.com/protocolbuffers/protobuf/issues/5318 javascript imports are sometimes generated incorrectly and also inefficient (i.e. same directory imports could be way shorter than they currently are).

This PR aims to improve javascript imports. 

(I'm sorry if there is more efficient ways to do this, this was my very first time to write c++ code)

Remaining question:
* this PR works under the assumption, that if the first part of an import compared to the source file does not match it has to be a third-party module. This could be changed to:
   * Only assume this for `google/protobuf` as currently and for other cases return `../../ .....` like the current compiler is doing it.
* The way I currently implemented it has an issue with the following imports as they would be seen as third party npm modules but there are none for these types:
    * `import "google/api/annotations.proto";`
    * `import "google/type/date.proto";`

Maybe this PR can start a discussion on how this can be improved further. I also couldn't find where to add test code for this on the repo so please let me know where I could put tests 🙇 . For now I tested it with one of my own projects where I need to compile to nodejs.

### DIFF
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -163,22 +163,18 @@ string GetSnakeFilename(const string& filename) {
 }
 
 // Given a filename like foo/bar/baz.proto, returns the corresponding JavaScript
-// file foo/bar/baz.js.
+// file baz.js
+// The relative import path is extracted by GetRelativeImportPath 
+// so we only need the filename here
 string GetJSFilename(const GeneratorOptions& options, const string& filename) {
-  return StripProto(filename) + options.GetFileNameExtension();
+  std::vector<string> source_file_split = Split(filename, "/");
+  string filename_only = source_file_split.back();
+  return StripProto(filename_only) + options.GetFileNameExtension();
 }
 
 // Given a filename like foo/bar/baz.proto, returns the root directory
 // path ../../
 string GetRootPath(const string& from_filename, const string& to_filename) {
-  if (to_filename.find("google/protobuf") == 0) {
-    // Well-known types (.proto files in the google/protobuf directory) are
-    // assumed to come from the 'google-protobuf' npm package.  We may want to
-    // generalize this exception later by letting others put generated code in
-    // their own npm packages.
-    return "google-protobuf/";
-  }
-
   size_t slashes = std::count(from_filename.begin(), from_filename.end(), '/');
   if (slashes == 0) {
     return "./";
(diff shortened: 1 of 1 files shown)


dominikeinkemmer:
I signed it!

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -163,31 +163,85 @@ string GetSnakeFilename(const string& filename) {
...
+    return JoinStrings(import_file_split, "/") + "/";
+  } else if (source_file_split.size() > import_file_split.size()) {
+    // Import from higher directory
+    // e.g. source /foo/bar/baz/baz.proto
+    //      import /foo/bar.proto
+    // expected return -> ../../
+    std::vector<string> mapped_vector;
+    mapped_vector.resize(source_file_split.size());
+    // transform remaining path length to ../ 
+    std::transform(source_file_split.begin(), source_file_split.end(), mapped_vector.begin(), [](string d) -> string { return "../"; });
```
TeBoring: Wrap the line?

TeBoring:
Please fix broken js test.

TeBoring:
> * Only assume this for `google/protobuf` as currently and for other cases return `../../ .....` like the current compiler is doing it.

I think this should be correct.

dominikeinkemmer:
> Please fix broken js test.

@TeBoring sorry, let me try to find where the tests are located and fix it 🙇

dominikeinkemmer:
@TeBoring hmm I checked but it seems the `Linux Javascript` and `MacOS Javascript` tests don't fail because of my code 🤔 even if I run the tests on latest master they fail with the same errors for the closure tests.

dominikeinkemmer:
> > * Only assume this for `google/protobuf` as currently and for other cases return `../../ .....` like the current compiler is doing it.

> I think this should be correct.

I updated the PR to follow this pattern, ptal 🙇

dominikeinkemmer:
@TeBoring I checked the errors of the javascript test runs and it seems the issue is here:

image

However, that issues shouldn't be a result of my code changes, as this error refers to a missing file that is imported in `js/binary/proto_test.js`. Any way you can check this? I'm completely unfamiliar with the test setup (and also what `goog.require` is doing) so it's difficult to check what's going on.

image

EItanya:
Hey, any news on this PR. I am  very interested in this feature, and I am more than willing to take over/help out.

dominikeinkemmer:
@EItanya I'm not familiar with their testing environment at all so I can't proceed the PR without any help because right now I don't have the time to read up on all of it 😢 .

Also our team actually switched our service to golang so this is also not a priority in our team anymore so I for now I can't spend any time on this during work either :/

EItanya:
@dominikeinkemmer I totally understand. I think we found a way around it for now, and so we don't have time to work on it either. Too bad the testing env isn't more transparent.

dominikeinkemmer:
No time to pursue this at the moment so I'll close this PR for now.

PR #5709, opened 2019-02-11
Allow Java reserved keywords to be used in extensions

jbolinger:
Protoc can produce invalid code when a reserved name is used in extensions. For example this field, named `package`, results in code that will not compile: https://github.com/googleapis/api-common-protos/blob/3f5a16b2c94a60a1aeb5424d490e62e3a69485ba/google/api/client.proto#L89

This change checks for reserved words and adds a suffix `_` to the name if there is a match.

Note that this PR does not include all reserved names yet and I used `_` for no particular reason. I can add them or change the naming pattern, but wanted to get feedback on the approach first.


src/google/protobuf/compiler/java/java_helpers.cc
```diff
@@ -62,6 +62,8 @@ const char kThickSeparator[] =
 const char kThinSeparator[] =
   "// -------------------------------------------------------------------\n";
 
+const std::set<std::string> kReservedNames = { "default", "package" };
```
haon4: move this into the anonymous namespace.

haon4:
+anandolee who was looking into similar problem. There is a list of Java Language Keywords here: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/_keywords.html. We should probably add them all to the list. The keyword conflict occurs not only in the extension name, but also in Message name, enum name and value, RPC Service and Method, etc.

src/google/protobuf/compiler/java/java_helpers.cc
jbolinger: done

src/google/protobuf/compiler/java/java_helpers.cc
```diff
@@ -76,6 +76,16 @@ const char* kForbiddenWordList[] = {
   "class",
 };
 
+const std::set<std::string> kReservedNames = {
```
haon4: for consistency, can you use string instead of std::string here and below.

jbolinger: Sure, I also changed the `std::set` to `std::unordered_set` since you were already using it.

jbolinger:
Sure, @anandolee I added a test and I also updated the lite generator to use the same naming pattern. I only tested a few names, let me know if you think this is sufficient.

java/core/src/test/proto/com/google/protobuf/non_nested_extension.proto
```diff
@@ -45,4 +45,5 @@ message MyNonNestedExtension {}
 
 extend MessageToBeExtended {
   optional MyNonNestedExtension nonNestedExtension = 1;
+  optional string if = 2000;
```
anandolee: Protobuf have two types of extension: top level extension and non top level extension. They should have same behavior. Does non top level extension also break without append '_' ? Some languages have different generated code for them, both should be tested

The extension you tested is top level extension.

None top level extension is like:
```
message ExtensionMessage {
    extend MessageToBeExtended {
        optional ExtensionMessage ext_field = 100;
    }
}
```
For example:
https://github.com/protocolbuffers/protobuf/blob/bf32b36ab8aa9e56c031e00c0553db640d3627d4/src/google/protobuf/unittest.proto#L350

java/core/src/test/java/com/google/protobuf/ExtensionRegistryFactoryTest.java
```diff
@@ -156,6 +158,11 @@ public void testAdd_immutable() {
       } catch (IllegalArgumentException expected) {
       }
     }
+
+    @Override
+    public void testExtensionRenamesKeywords() {
+      assertTrue(NonNestedExtension.if_ instanceof GeneratedMessage.GeneratedExtension);
+    }
```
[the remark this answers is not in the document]
jbolinger: I've added a few extra checks for the getters and setters.

java/core/src/test/proto/com/google/protobuf/non_nested_extension.proto
jbolinger: Makes sense. I added a similar test for the nested extension in both cases. Let me know if you think we need anything more.

anandolee:
LGTM and thank for the PR. Let's wait for the tests

liya0412:
GPBExtensionDescriptor 如何生成这个子类的实例化对象

PR #5713, opened 2019-02-12
pyext: missing cast to constant char*

grindhold:
We need to cast to a constant charpointer here.

This has already been done in other files of the module, so I analogously applied the existing fix.

The error occured to me when building tensorflow against protobuf:

```
external/protobuf_archive/python/google/protobuf/pyext/message_factory.cc:69:45: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
        ((*(charpp) = PyUnicode_AsUTF8AndSize(ob, (sizep))) == NULL? -1: 0): \
                      ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
```

### DIFF
--- a/python/google/protobuf/pyext/message_factory.cc
+++ b/python/google/protobuf/pyext/message_factory.cc
@@ -43,9 +43,11 @@
     #error "Python 3.0 - 3.2 are not supported."
   #endif
   #define PyString_AsStringAndSize(ob, charpp, sizep) \
-    (PyUnicode_Check(ob)? \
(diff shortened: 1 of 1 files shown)


anandolee:
Have you signed CLA? If you have signed, please reply "I signed it!"

grindhold:
I signed it!

grindhold:
I don't have the slightest clue why my python-related fix breaks the js-build. I can't even find the file that it breaks on in the code base.

anandolee:
JS failure should not related. Will merge the PR.
Thanks for your fix

PR #5716, opened 2019-02-12
Support hashes for struct initializers

### LINKED ISSUE #3120: Ruby: Allow for the instantiation of nested message fields with ruby-hashes

jbolinger:
This change simplifies working with `Google::Protobuf::Struct` and `Google::Protobuf::Value` so that code like this:

```ruby
Google::Protobuf::Struct.new(fields: {
  'a' => Google::Protobuf::Value.new({number_value: 2.0})
})
```
can be written as:

```ruby
Google::Protobuf::Struct.new(fields: {
   'a' => {number_value: 2.0}
})
```

This would simplify struct usage in many practical cases, such as: https://github.com/googleapis/google-cloud-ruby/issues/2839

I believe this also resolves #3120.

### DIFF
--- a/ruby/ext/google/protobuf_c/map.c
+++ b/ruby/ext/google/protobuf_c/map.c
@@ -395,6 +395,11 @@ VALUE Map_index_set(VALUE _self, VALUE key, VALUE value) {
   void* mem;
   key = table_key(self, key, keybuf, &keyval, &length);
 
+  if (TYPE(value) == T_HASH) {
+    VALUE args[1] = { value };
+    value = rb_class_new_instance(1, args, self->value_type_class);
+  }
+
   mem = value_memory(&v);
   native_slot_set("", self->value_type, self->value_type_class, mem, value);
 
--- a/ruby/ext/google/protobuf_c/map.c
+++ b/ruby/ext/google/protobuf_c/map.c
@@ -71,6 +71,9 @@ static VALUE table_key(Map* self, VALUE key,
     case UPB_TYPE_BYTES:
     case UPB_TYPE_STRING:
       // Strings: use string content directly.
+      if (TYPE(key) == T_SYMBOL) {
+        key = rb_id2str(SYM2ID(key));
+      }
       Check_Type(key, T_STRING);
       key = native_slot_encode_and_freeze_string(self->key_type, key);
       *out_key = RSTRING_PTR(key);
--- a/ruby/tests/well_known_types_test.rb
+++ b/ruby/tests/well_known_types_test.rb
@@ -139,4 +139,31 @@ class TestWellKnownTypes < Test::Unit::TestCase
     assert any.is(Google::Protobuf::Timestamp)
     assert_equal ts, any.unpack(Google::Protobuf::Timestamp)
   end
+
+  def test_struct_assign
+    s = Google::Protobuf::Struct.new(fields: {'a' => Google::Protobuf::Value.new({number_value: 4.4})})
+    assert_equal 4.4, s['a']
+
+    s = Google::Protobuf::Struct.new(fields: {'a' => {number_value: 2.2}})
+    assert_equal 2.2, s['a']
+  end
+
+  def test_struct_nested_assign
+    s = Google::Protobuf::Struct.new(
+      fields: {
+        'a' => {string_value: 'A'},
+        'b' => {struct_value: {
+          fields: {
+            'x' => {list_value: {values: [{number_value: 1.0}, {string_value: "ok"}]}},
+            'y' => {bool_value: true}}}
+        },
+        'c' => {struct_value: {}}
+      }
+    )
+    assert_equal 'A', s['a']
+    expected_b_x = [Google::Protobuf::Value.new(number_value: 1.0), Google::Protobuf::Value.new(string_value: "ok")]
+    assert_equal expected_b_x, s['b']['x'].values
+    assert_equal true, s['b']['y']
+    assert_equal Google::Protobuf::Struct.new, s['c']
+  end
 end
--- a/ruby/tests/well_known_types_test.rb
+++ b/ruby/tests/well_known_types_test.rb
@@ -140,7 +140,7 @@ class TestWellKnownTypes < Test::Unit::TestCase
     assert_equal ts, any.unpack(Google::Protobuf::Timestamp)
   end
 
-  def test_struct_assign
+  def test_struct_init
     s = Google::Protobuf::Struct.new(fields: {'a' => Google::Protobuf::Value.new({number_value: 4.4})})
     assert_equal 4.4, s['a']
 
@@ -148,10 +148,10 @@ class TestWellKnownTypes < Test::Unit::TestCase
     assert_equal 2.2, s['a']
 
     s = Google::Protobuf::Struct.new(fields: {a: {number_value: 1.1}})
-    assert_equal 1.1, s['a']
+    assert_equal 1.1, s[:a]
   end
 
-  def test_struct_nested_assign
+  def test_struct_nested_init
     s = Google::Protobuf::Struct.new(
       fields: {
         'a' => {string_value: 'A'},
--- a/ruby/tests/well_known_types_test.rb
+++ b/ruby/tests/well_known_types_test.rb
@@ -164,10 +164,18 @@ class TestWellKnownTypes < Test::Unit::TestCase
       }
     )
     assert_equal 'A', s['a']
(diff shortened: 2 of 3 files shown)


ruby/tests/well_known_types_test.rb
```diff
@@ -139,4 +139,31 @@ def test_any
     assert any.is(Google::Protobuf::Timestamp)
     assert_equal ts, any.unpack(Google::Protobuf::Timestamp)
   end
+
+  def test_struct_assign
+    s = Google::Protobuf::Struct.new(fields: {'a' => Google::Protobuf::Value.new({number_value: 4.4})})
+    assert_equal 4.4, s['a']
+
+    s = Google::Protobuf::Struct.new(fields: {'a' => {number_value: 2.2}})
```
blowmage: i would love to see the `fields` Hash keys be converted to Strings, so we could also use this:

```ruby
s = Google::Protobuf::Struct.new(fields: {a: {number_value: 2.2}})
```

ruby/tests/well_known_types_test.rb
```diff
@@ -139,4 +139,31 @@ def test_any
...
+    assert_equal 2.2, s['a']
```
blowmage: i would also like to see `Google::Protobuf::Struct#[]` convert the method argument to a String, so this would also work:

```ruby
assert_equal 2.2, s[:a]
```

ruby/tests/well_known_types_test.rb
jbolinger: you got it

ruby/tests/well_known_types_test.rb
jbolinger: you got it. How's this look?

blowmage: Looks great!

ruby/tests/well_known_types_test.rb
```diff
@@ -165,5 +168,17 @@ def test_struct_nested_assign
     assert_equal expected_b_x, s['b']['x'].values
     assert_equal true, s['b']['y']
```
blowmage: I would like to see these access checks made with string, symbols, and a combination, to make it really clear they are checking the same thing.

```ruby
assert_equal true, s['b']['y']
assert_equal true, s[:b][:y]
assert_equal true, s[:b]['y']
assert_equal true, s['b'][:y]
```

jbolinger: Makes sense - added.

ruby/ext/google/protobuf_c/map.c
```diff
@@ -71,6 +71,9 @@ static VALUE table_key(Map* self, VALUE key,
     case UPB_TYPE_BYTES:
     case UPB_TYPE_STRING:
       // Strings: use string content directly.
+      if (TYPE(key) == T_SYMBOL) {
```
TeBoring: Why we have to this in order to support Struct?

jbolinger: I don't think so, and it didn't break any tests that I'm aware of. Can you think of a case where it would make a difference?

This would not throw type errors in some cases where it would have previously but I don't know if that should be considered additive vs a breaking change?

TeBoring: It doesn't break existing test cases. However, it allows symbol for string map, which is not allowed previously.

jbolinger: Right, is there any reason to prohibit that?

TeBoring: It makes map not consistent with optional, repeated fields.
Besides, how to compare symbol? identity compare? value compare? Is the semantic same as string?
Besides, is the feature quite important? Otherwise, we don't want to over-provide it, as these trivial features necessarily makes our library complicate.

TeBoring: It also breaks existing type checking. We should added tests for our type checking.

blowmage: It allows users to provide:

```ruby
Google::Protobuf::Struct.new(fields: {
   a: {number_value: 2.0}
})
```

instead of:

```ruby
Google::Protobuf::Struct.new(fields: {
   'a' => {number_value: 2.0}
})
```

I would argue that Converting `Symbol` key values to `String` in `Struct` is a beneficial affordance, as `Struct` will only every have `String` keys. `Map` can have any any value type as a key, so it wouldn't make as much sense.

That said, I do think it would be fine to accept `Symbol` values every time a `String` is expected. In Ruby, I would write it like this:

```ruby
class Google::Protobuf::Struct
  def [](key)
    inner_struct[String(key)]
  end
end
```

Basically, anytime we expect a `String` object, pass it to the `String()` method to perform the necessary conversions.

jbolinger:
@TeBoring what kind of type checking tests did you have in mind? I'm happy to add them but I'm not sure exactly what you're asking for.

TeBoring:
The check test I have in mind is something like:
```
// expect throw error
a.foo = 1  // foo is a string field
```
However, this could be a separate change.

TeBoring:
If ruby users do expect Symbol to be the same as string, I think we should make it consistent in all our APIs.
However, we need a API review to allow that. @jbolinger Could you write a doc for that?
My suggestion is allowing Symbol could be another change. We don't need to do it in this change.

jbolinger:
I'm not sure that the inconsistency is all that bad. Backing up a bit, the first version of this PR didn't include the symbol bit. I was trying to fix an existing issue that prevented using a hash to init a map value:

So, instead of this:

```ruby
Google::Protobuf::Struct.new(fields: {
  'a' => Google::Protobuf::Value.new({number_value: 2.0})
})
```

one could write a more concise version:

```ruby
Google::Protobuf::Struct.new(fields: {
   'a' => {number_value: 2.0}
})
```

I shouldn't speak for @blowmage, but he pointed out that getting rid of the hash rocket would look even better. It's shorter and likely the more commonly used syntax. So, I added the implicit string/symbol conversion to allow this:

```ruby
Google::Protobuf::Struct.new(fields: {
   a: {number_value: 2.0}
})
```

Sure, it's a little inconsistent but since a map key can't be symbol allowing either to be used in this context seems like a useful convenience. I'm not sure if I understand what you mean by making it consistent for optional fields or treating string/symbol as the same, etc. but I don't think it makes sense to support something like this:

```ruby
msg.a_string = :some_thing
```

That should probably require an explicit conversion and, even if we did support it, I'm not sure it would actually be useful in very many situations.

jbolinger:
@TeBoring any updates on this (and could you run the tests again)?

TeBoring:
I mean please make the Symbol a separate change.

jbolinger:
@TeBoring what other use cases / changes did you have in mind? It looks like Symbol behaves as a string in certain contexts already (see my previous example). What else should we support?

TeBoring:
I see your comment. So you only want to make string map key to accept
symbol?

TeBoring:
Only for the map in struct?

TeBoring:
If you could make the change only affect the map in struct but not other
maps, I am fine with that.

TeBoring:
Never mind. I think I can accept this change. I want to let @haberman to also take a look, since this affects API.

jbolinger:
Thanks @TeBoring! I see a build failure but I don't think it's related to this, but let me know if it is or if anything else needs to be done.

jbolinger:
Are you still interested in accepting this change? I think it's a nice simplification but it fell off my radar. Let me know if I can do anything else.

TeBoring:
It seems ruby 2.3 still failed?

jbolinger:
Hmm, that's sort of a strange error that looks like it might be related to the rvm or something unrelated in the environment (openssl?). I can try to reproduce it when I'm back in the office on Monday.

TeBoring:
Maybe sync and push?

TeBoring:
https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/

Since the support for ruby 2.3 has been ended, we will deprecate ruby 2.3.

PR #5723, opened 2019-02-13
PHP: Exclude repeated and map fields from normalization in constructor

michaelbausor:
This PR:
- Exclude repeated and map fields over wrapper types from trying to convert a primitive value to a wrapper type in the protobuf message constructor
- Adds tests

cc @TeBoring

### DIFF
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -987,7 +987,7 @@ class Message
                     'Invalid message property: ' . $key);
             }
             $setter = $field->getSetter();
-            if ($field->isWrapperType() && !($field->isRepeated() || $field->isMap())) {
+            if ($field->isWrapperType() && !$field->isRepeated()) {
                 self::normalizeToMessageType($value, $field->getMessageType()->getClass());
             }
             $this->$setter($value);
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -973,9 +973,12 @@ class Message
      * ]);
      * ```
      *
+     * This method will trigger an error if it is passed data that cannot
+     * be converted to the correct type. For example, a StringValue field
+     * must receive data that is either a string or a StringValue object.
+     *
      * @param array $array An array containing message properties and values.
      * @return null.
-     * @throws \Exception Invalid data.
      */
     protected function mergeFromArray(array $array)
     {
@@ -1013,7 +1016,6 @@ class Message
      *
      * @param mixed $value The array of values to normalize.
      * @param string $class The expected wrapper class name
-     * @throws \Exception If an element of $value cannot be converted
      */
     private static function normalizeArrayElementsToMessageType(&$value, $class)
     {
@@ -1038,9 +1040,11 @@ class Message
      * instance of $class and assign $value to it using the setValue method
      * shared by all wrapper types.
      *
+     * This method will raise an error if it receives a type that cannot be
+     * assigned to the wrapper type via setValue.
+     *
      * @param mixed $value The value to normalize.
      * @param string $class The expected wrapper class name
-     * @throws \Exception If $value cannot be converted to a wrapper type
      */
     private static function normalizeToMessageType(&$value, $class)
     {
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -987,10 +987,56 @@ class Message
                     'Invalid message property: ' . $key);
             }
             $setter = $field->getSetter();
-            if ($field->isWrapperType() && !$field->isRepeated()) {
-                self::normalizeToMessageType($value, $field->getMessageType()->getClass());
+            if ($field->isMap()) {
+                $valueField = $field->getMessageType()->getFieldByName('value');
+                if (!is_null($valueField) && $valueField->isWrapperType()) {
+                    self::normalizeArrayElementsToMessageType($value, $valueField->getMessageType()->getClass());
+                }
+            } elseif ($field->isWrapperType()) {
+                $class = $field->getMessageType()->getClass();
+                if ($field->isRepeated()) {
+                    self::normalizeArrayElementsToMessageType($value, $class);
+                } else {
+                    self::normalizeToMessageType($value, $class);
+                }
             }
             $this->$setter($value);
+        }
+    }
+
+    /**
+     * Tries to normalize the elements in $value into a provided protobuf
+     * wrapper type $class. If $value is any type other than array, we do
+     * not do any conversion, and instead rely on the existing protobuf
+     * type checking. If $value is an array, we process each element and
+     * try to convert it to an instance of $class.
+     *
+     * @param mixed $value The array of values to normalize.
+     * @param string $class The expected wrapper class name
+     * @throws \Exception If an element of $value cannot be converted
+     */
+    private static function normalizeArrayElementsToMessageType(&$value, $class)
+    {
+        if (!is_array($value)) {
+            // In the case that $value is not an array, we do not want to
+            // attempt any conversion. Note that this includes the cases
+            // when $value is a RepeatedField of MapField. In those cases,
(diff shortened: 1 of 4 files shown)


TeBoring:
Tests seem failing.

TeBoring:
wrapper_type_setters_test.php

michaelbausor:
Yes, that test is failing when running with the protobuf extension installed. It passes with the extension installed. I have implemented a fix only for the pure php portion, but we also need to fix in the c extension.

TeBoring:
I see. I'll work on it.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -987,7 +987,7 @@ protected function mergeFromArray(array $array)
                     'Invalid message property: ' . $key);
             }
             $setter = $field->getSetter();
-            if ($field->isWrapperType()) {
+            if ($field->isWrapperType() && !($field->isRepeated() || $field->isMap())) {
```
TeBoring: A wrapper type field cannot be map field. Only the value field of a map entry can be a wrapper type message.

php/tests/wrapper_type_setters_test.php
```diff
@@ -225,4 +225,43 @@ public function constructorWithWrapperTypeDataProvider()
...
+
+    public function constructorWithRepeatedWrapperTypeDataProvider()
+    {
+        return [
+            [TestWrapperSetters::class, 'repeated_string_value', 'getRepeatedStringValue', [new StringValue(['value' => 'eight'])]],
+            [TestWrapperSetters::class, 'map_string_value', 'getMapStringValue', ['key' => new StringValue(['value' => 'eight'])]],
+        ];
+    }
+
+    /**
```
TeBoring: Unused code?

php/tests/wrapper_type_setters_test.php
```diff
@@ -225,4 +225,43 @@ public function constructorWithWrapperTypeDataProvider()
             [TestWrapperSetters::class, BytesValue::class, 'bytes_value', 'getBytesValue', "nine"],
         ];
     }
+
+    /**
+     * @dataProvider invalidConstructorWithWrapperTypeDataProvider
+     * @expectedException \Exception
+     */
+    public function testInvalidConstructorWithWrapperType($class, $wrapperField, $value)
```
TeBoring: I think we need some discussion for the failure behavior. Do we want error or exception?
https://github.com/protocolbuffers/protobuf/pull/5571 This change will make wrong type as error. Shall we make all wrong type failure behavior the same?

michaelbausor: The intention of this change is not to introduce a new error, but just to preserve existing behaviour prior to adding the `normalizeToMessageType` functionality. Right now, if you pass an array of StringValue objects into the constructor, it will incorrectly throw an error because it tries to convert the array into a StringValue. With this change, this case will be handled correctly (as it was prior to the change adding `normalizeToMessageType`. There should be no new error cases covered.

This test is just to confirm the existing behaviour in previous releases.

php/src/Google/Protobuf/Internal/Message.php
michaelbausor: Ah ok, will remove check for isMap

php/tests/wrapper_type_setters_test.php
michaelbausor: Yes, removed.

php/tests/wrapper_type_setters_test.php
TeBoring: Shall we also make repeated/map wrapper types work instead?
If we find it's a repeated field of wrapper types or a map field whose value type is wrapper types, we can convert the array/map to another one with wrapper messages inside.

TeBoring: Shall we make repeated/map also work?

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -987,13 +987,59 @@ protected function mergeFromArray(array $array)
...
+    /**
+     * Tries to normalize the elements in $value into a provided protobuf
+     * wrapper type $class. If $value is any type other than array, we do
+     * not do any conversion, and instead rely on the existing protobuf
+     * type checking. If $value is an array, we process each element and
+     * try to convert it to an instance of $class.
+     *
+     * @param mixed $value The array of values to normalize.
+     * @param string $class The expected wrapper class name
+     * @throws \Exception If an element of $value cannot be converted
```
TeBoring: Can we make it an error?

TeBoring:
Sent you a fix.

php/tests/wrapper_type_setters_test.php
```diff
@@ -225,4 +225,113 @@ public function constructorWithWrapperTypeDataProvider()
             [TestWrapperSetters::class, BytesValue::class, 'bytes_value', 'getBytesValue', "nine"],
         ];
     }
+
+    // /**
+    //  * @dataProvider invalidConstructorWithWrapperTypeDataProvider
+    //  * @expectedException \Exception
+    //  */
+    // public function testInvalidConstructorWithWrapperType($class, $wrapperField, $value)
```
TeBoring: I don't want to make it an exception. I want to make it an error.
Our plan is make all type checking an error (which is consistent with native php type hint).
I know previously php implementation throws exception. However, c extension throws error.
(In my mind, this is actually an undefined behavior)

php/tests/wrapper_type_setters_test.php
```diff
@@ -225,4 +225,113 @@ public function constructorWithWrapperTypeDataProvider()
...
+    {
+        $sv7 = new StringValue(['value' => 'seven']);
+        $sv8 = new StringValue(['value' => 'eight']);
+
+        $testWrapperSetters = new TestWrapperSetters();
+        $testWrapperSetters->setRepeatedStringValue([$sv7, $sv8]);
+        $repeatedField = $testWrapperSetters->getRepeatedStringValue();
+
+        return [
+            [TestWrapperSetters::class, 'repeated_string_value', 'getRepeatedStringValue', []],
```
TeBoring: Could you wrap the line?

php/src/Google/Protobuf/Internal/Message.php
michaelbausor: Right now, in the normalizeToMessageType function, we catch any exception coming from `setValue()` and change the message to be more descriptive. Should I instead catch the exception and convert it to an error?

TeBoring:
SGTM

michaelbausor:
Updated, PTAL

php/tests/wrapper_type_setters_test.php
michaelbausor: Shortened line by removing `TestWrapperSetters::class`

php/tests/wrapper_type_setters_test.php
michaelbausor: Removed

php/src/Google/Protobuf/Internal/Message.php
michaelbausor: Done

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -987,13 +987,51 @@ protected function mergeFromArray(array $array)
...
```
TeBoring: Should this doc be updated?

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -987,13 +987,51 @@ protected function mergeFromArray(array $array)
                     'Invalid message property: ' . $key);
             }
             $setter = $field->getSetter();
-            if ($field->isWrapperType()) {
```
TeBoring: mergeFromArray throws error with invalid data. Update doc?

michaelbausor: Done

php/src/Google/Protobuf/Internal/Message.php
michaelbausor: Done

PR #5726, opened 2019-02-14
Add frozen checks in Ruby

### LINKED ISSUE #907: `freeze` should work in Ruby
In Ruby frozen objects cannot be modified. Unfortunately this is not the case with Ruby objects generated from protobuf IDL:

```
irb(main):020:0> loc = Codebeat::Analyser::Location.new
=> <Codebeat::Analyser::Location: path: "", start_line: 0, end_line: 0>
irb(main):021:0> loc.path = 'my_path'
=> "my_path"
irb(main):022:0> loc.freeze
=> <Codebeat::Analyser::Location: path: "my_path", start_line: 0, end_line: 0>
irb(main):023:0> loc.path = 'new_path'
=> "new_path"
irb(main):024:0> loc 
=> <Codebeat::Analyser::Location: path: "new_path", start_line: 0, end_line: 0>
irb(main):025:0> loc.frozen?
=> true
```

jbolinger:
An attempt to implement `.freeze` support for protobuf message types and resolve #907.


ruby/ext/google/protobuf_c/map.c
```diff
@@ -388,6 +388,10 @@ VALUE Map_index(VALUE _self, VALUE key) {
 VALUE Map_index_set(VALUE _self, VALUE key, VALUE value) {
   Map* self = ruby_to_Map(_self);
 
+  if (rb_obj_frozen_p(_self)) {
```
blowmage: I believe this should raise a `FrozenError`, not an `ArgumentError`. You might be able to replace this whole conditional with the following:

```c
rb_check_frozen(_self);
```

blowmage:
I’m okay with it.

ruby/ext/google/protobuf_c/map.c
```diff
@@ -388,6 +388,8 @@ VALUE Map_index(VALUE _self, VALUE key) {
 VALUE Map_index_set(VALUE _self, VALUE key, VALUE value) {
   Map* self = ruby_to_Map(_self);
```
blowmage: Should the frozen check happen before the conversion to Map?

jbolinger: I doubt it really makes a difference, but I'll move them up just in case.

TeBoring:
@blowmage have you signed CLA?

blowmage:
Yes, I have signed the CLA.

blowmage:
> The good news is that everyone that needs to sign a CLA (the pull request submitter and all commit authors) have done so. Everything is all good there.

jbolinger:
@TeBoring thanks for helping with this!

I just realized that this makes the unit tests require Ruby >= 2.5 because that's when `FrozenError` was added (previously it throws a `RuntimeError` instead). Want me to send another PR to check the error based on the version?

TeBoring:
https://github.com/protocolbuffers/protobuf/blob/master/ruby/tests/common_tests.rb#L212
Previously, we did in this way.

jbolinger:
Opened #5823 to take care of the version differences.

PR #5735, opened 2019-02-15
Fix Ruby module name generation when the ruby_package option is used

### LINKED ISSUE #5584: Option ruby_package causes inconsistent code generation
Version: 3.6.1
Language: Ruby

MacOS 10.14

Ruby 2.4.1

**What did you do?**
1. Compile the following proto: `protoc --proto_path=. --ruby_out=. example.proto`
```
syntax = "proto3";

package google.example;

option java_multiple_files = true;
//option ruby_package = "Google::Example";

service HelloService {
     rpc HiThere (HiRequest) returns (HiResponse);
}

message HiRequest {
    string query = 1;
}

message HiResponse {
    string result = 1;
}
```
2. Repeat step 1 after uncommenting the `ruby_package` line.

The generated Ruby code should be the same.

**What did you see instead?**
When the `ruby_package` options is specified the generated code is:
```ruby
module Google::Example
  HiRequest = Google::Protobuf::DescriptorPool.generated_pool.lookup("google.example.HiRequest").msgclass
  HiResponse = Google::Protobuf::DescriptorPool.generated_pool.lookup("google.example.HiResponse").msgclass
end
```

Without the `ruby_package` option it is:
```ruby
module Google
  module Example
    HiRequest = Google::Protobuf::DescriptorPool.generated_pool.lookup("google.example.HiRequest").msgclass
    HiResponse = Google::Protobuf::DescriptorPool.generated_pool.lookup("google.example.HiResponse").msgclass
  end
end
```

This inconsistency causes issues because the module must already be defined if the `ruby_package` option is used (example: https://github.com/googleapis/google-cloud-ruby/pull/2782).

jbolinger:
Resolves #5584 by ensure that the Ruby module name is always generated in the following format:

```ruby
module A
  module B
    # ....
```

instead of:

```ruby
module A::B
```

I also added some tests for the `ruby_package` option.


ruby/tests/test_ruby_package.proto
```diff
@@ -2,6 +2,6 @@ syntax = "proto3";
 
 package foo_bar;
 
-option ruby_package = "A.B";
+option ruby_package = "A::B";
```
TeBoring: Does this change the existing API?

jbolinger: Sort of, but I think the original implementation was incorrect. 

In all other languages, the package options follow the same format as the language and `::` is the correct delimiter in Ruby.

For example, this proto has a bunch of language package options defined: https://github.com/googleapis/googleapis/blob/1a4f0f12777dc2f8bf2c2ce84438329639c75e29/google/pubsub/v1/pubsub.proto#L33

TeBoring: Will this break existing users?
If so, can we support both for now?

TeBoring: Maybe in protoc, we can throw warning if "A.B" is used.

jbolinger: Supporting both options sounds like a good solution. I've updated it to work with both and issue a warning if the `A.B` form is used instead of `A::B`.

PR #5736, opened 2019-02-15
Allow repeated fields in Ruby to accept multiple arguments on push

### LINKED ISSUE #4278: Ruby: Repeated Field handling of custom types not working
I've found a bug that prevents the modification (or setting) of repeated fields for non-primitive types. I know that setting a repeated custom field at message initialization works, but that's not what we need to be doing. Below I've provided a minimal example showing what I've attempted.

I created a simple protobuf file with two repeated fields; one of a primtive type, and one of a custom type: repeated_test.proto
```
syntax = "proto3";             
  
message CustomField {          
  string name = 1;
  string value = 2;
}

message RepeatedTest {
  repeated string       repeated_primitive = 1;
  repeated CustomField  repeated_custom = 2;
}
```

Generated protobuf with protoc
```
$ protoc --version
libprotoc 3.5.1

$ protoc --ruby_out=generated/ *proto
```

generated/repeated_test_pb.rb
```
# source: repeated_test.proto

require 'google/protobuf'

  add_message "CustomField" do
    optional :name, :string, 1
    optional :value, :string, 2
  end
  add_message "RepeatedTest" do
    repeated :repeated_primitive, :string, 1
    repeated :repeated_custom, :message, 2, "CustomField"
  end
end

CustomField = Google::Protobuf::DescriptorPool.generated_pool.lookup("CustomField").msgclass
RepeatedTest = Google::Protobuf::DescriptorPool.generated_pool.lookup("RepeatedTest").msgclass
```

Then I tried to use it (Ruby version of 2.2.2, google-protobuf version of 3.5.1.2)
```
2.2.2 :001 > test = RepeatedTest.new()
 => <RepeatedTest: repeated_primitive: [], repeated_custom: []>
2.2.2 :002 > custom_value = CustomField.new(name: "foo", value: "bar")
 => <CustomField: name: "foo", value: "bar"> 

# Appending to the primtive repeated field worked as expected
2.2.2 :003 > test.repeated_primitive += ["new_string"]
 => ["new_string"] 
2.2.2 :004 > test
 => <RepeatedTest: repeated_primitive: ["new_string"], repeated_custom: []> 
 
# But appending to the custom repeated field throws an error
2.2.2 :005 > test.repeated_custom += [custom_value]
TypeError: Repeated field array has wrong message/enum class
	from (irb):5:in `method_missing'
	from (irb):5
	from /Users/cgross/.rbenv/versions/2.2.2/bin/irb:11:in `<main>'

# I also tried pulling out the pre-generated repeated field object, adding to it and putting it back...
2.2.2 :006 > separate_repeated_custom = test.repeated_custom
 => [] 
2.2.2 :008 > separate_repeated_custom.class
 => Google::Protobuf::RepeatedField 

# I can add to it just fine...
2.2.2 :010 > separate_repeated_custom += [custom_value]
 => [<CustomField: name: "foo", value: "bar">] 
2.2.2 :011 > separate_repeated_custom.class
 => Google::Protobuf::RepeatedField 

# But when I try to put it back in (with assignment or appending), it throws the same error as before
2.2.2 :012 > test.repeated_custom = separate_repeated_custom
TypeError: Repeated field array has wrong message/enum class
	from (irb):12:in `method_missing'
	from (irb):12
	from /Users/cgross/.rbenv/versions/2.2.2/bin/irb:11:in `<main>'
2.2.2 :013 > test.repeated_custom += separate_repeated_custom
TypeError: Repeated field array has wrong message/enum class
	from (irb):13:in `method_missing'
	from (irb):13
	from /Users/cgross/.rbenv/versions/2.2.2/bin/irb:11:in `<main>'

```

jbolinger:
Allow the `push` method on repeated fields to accept multiple values and resolve #4278.

### DIFF
--- a/ruby/tests/common_tests.rb
+++ b/ruby/tests/common_tests.rb
@@ -708,6 +708,28 @@ module CommonTests
     assert proto_module::TestEnum::resolve(:C) == 3
   end
 
+  def test_repeated_push
+    m = proto_module::TestMessage.new
+
+    m.repeated_string << "one"
+    m.repeated_string.push "two"
+    assert_equal ["one", "two"], m.repeated_string
+
+    m.repeated_string.push ["three", "four"]
+    assert_equal ["one", "two", "three", "four"], m.repeated_string
+
+    m.repeated_string << ["five"]
+    assert_equal ["one", "two", "three", "four", "five"], m.repeated_string
+
+    m = proto_module::TestMessage.new
+
+    m.repeated_msg << proto_module::TestMessage2.new(:foo => 1)
+    m.repeated_msg.push proto_module::TestMessage2.new(:foo => 2)
+    m.repeated_msg << [proto_module::TestMessage2.new(:foo => 3)]
+    m.repeated_msg.push [proto_module::TestMessage2.new(:foo => 4)]
+    assert_equal [1, 2, 3, 4], m.repeated_msg.map {|x| x.foo}
+  end
+
   def test_parse_serialize
     m = proto_module::TestMessage.new(:optional_int32 => 42,
                                       :optional_string => "hello world",
--- a/ruby/tests/common_tests.rb
+++ b/ruby/tests/common_tests.rb
@@ -711,22 +711,21 @@ module CommonTests
   def test_repeated_push
     m = proto_module::TestMessage.new
 
-    m.repeated_string << "one"
-    m.repeated_string.push "two"
-    assert_equal ["one", "two"], m.repeated_string
+    m.repeated_string += ['one']
+    m.repeated_string += %w[two three]
+    assert_equal %w[one two three], m.repeated_string
 
-    m.repeated_string.push ["three", "four"]
-    assert_equal ["one", "two", "three", "four"], m.repeated_string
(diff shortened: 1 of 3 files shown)


ruby/tests/common_tests.rb
```diff
@@ -708,6 +708,28 @@ def test_enum_lookup
...
   end
 
+  def test_repeated_push
+    m = proto_module::TestMessage.new
+
+    m.repeated_string << "one"
+    m.repeated_string.push "two"
+    assert_equal ["one", "two"], m.repeated_string
+
+    m.repeated_string.push ["three", "four"]
```
blowmage: This should only behave this way if instead of an array, there values were given as different arguments to `push`:

```ruby
m.repeated_string.push "three", "four"
# or
m.repeated_string.push *["three", "four"]
```

jbolinger: I reverted the change to the other operations and changed `push` to allow multiple arguments. Let me know if that looks right.

TeBoring:
@haberman PTAL

PR #5739, opened 2019-02-16
Add wrapper type helpers for Ruby

### LINKED ISSUE #5656: Better handling of wrapper types in Ruby
Ruby

Using wrapper types, such as `google.protobuf.StringValue` can be useful when field presence is important, but they are difficult to work with. For example, to assign a new value one has to write: 

`message.a_string = Google::Protobuf::StringValue.new(value: "some thing")`
instead of
`message.a_string = "some thing"`

Additional getters / setters that accept strings and transparently handle the wrapper types would be more concise. 

Helper methods for creating the wrapper types can help, but still aren't great. Example: https://github.com/googleads/google-ads-ruby/issues/5

**Additional context**

jbolinger:
Resolves #5656 by adding additional helper methods for fields of wrapper types (defined `wrappers.proto`). 

For example: https://github.com/protocolbuffers/protobuf/blob/3bf05b88eaf938526f7daee85ab6fb1efb0e809c/src/google/protobuf/wrappers.proto#L56-L59)

A message like this:
```proto
message Wrapper {
  google.protobuf.DoubleValue double = 1;
}
```

Will have getters and setters defined as follows:
```ruby
msg = Wrapper.new
msg.double_as_value = 2.2
msg.double
# => <Google::Protobuf::DoubleValue: value: 2.2>
msg.double_as_value
# => 2.2
```

This avoids having to use the wrapper types directly, such as:
```ruby
msg = Wrapper.new
msg.double = Google::Protobuf::DoubleValue.new value: 2.2
msg.double.value
# => 2.2
```

This change has a bit of overlap with #5670 and it would be best to handle that one first.


jbolinger:
@TeBoring any updates on this?

TeBoring:
Has the design doc been reviewed?

jbolinger:
Yes, AFAIK we are just waiting for you and/or @haberman to approve. Everyone on my side has reviewed & signed off on it.

ruby/ext/google/protobuf_c/message.c
```diff
@@ -157,6 +185,34 @@ static int extract_method_call(VALUE method_name, MessageHeader* self,
...
 			                                   &test_f, &test_o);
 
+  // Look for wrapper type accessor of the form <field_name>_as_value
+  if (!has_field &&
+      (accessor_type == METHOD_GETTER || accessor_type == METHOD_SETTER) &&
+      name_len > 9 && strncmp(name + name_len - 9, "_as_value", 9) == 0) {
+    // Find the field name
+    char wrapper_field_name[name_len - 8];
+    strncpy(wrapper_field_name, name, name_len - 9);
+    wrapper_field_name[name_len - 7] = '\0';
```
tenderlove: `wrapper_field_name` is only `name_len - 8` wide.  `name_len - 7` is greater than `name_len - 8`, so I think this is writing outside the buffer.

PR #5740, opened 2019-02-16
Simplify holding copies of strings in default_value_objectwriter

54markov:
If you don't mind, i propose to improve the implementation of `default_value_objectwriter` with smart pointer for `string_values_`.

### DIFF
--- a/src/google/protobuf/util/internal/default_value_objectwriter.cc
+++ b/src/google/protobuf/util/internal/default_value_objectwriter.cc
@@ -71,9 +71,6 @@ DefaultValueObjectWriter::DefaultValueObjectWriter(
       ow_(ow) {}
 
 DefaultValueObjectWriter::~DefaultValueObjectWriter() {
-  for (int i = 0; i < string_values_.size(); ++i) {
-    delete string_values_[i];
-  }
   if (own_typeinfo_) {
     delete typeinfo_;
   }
@@ -156,7 +153,7 @@ DefaultValueObjectWriter* DefaultValueObjectWriter::RenderString(
   } else {
     // Since StringPiece is essentially a pointer, takes a copy of "value" to
     // avoid ownership issues.
-    string_values_.push_back(new string(value));
+    string_values_.emplace_back(new string(value));
     RenderDataPiece(name, DataPiece(*string_values_.back(), true));
   }
   return this;
@@ -169,7 +166,7 @@ DefaultValueObjectWriter* DefaultValueObjectWriter::RenderBytes(
   } else {
     // Since StringPiece is essentially a pointer, takes a copy of "value" to
     // avoid ownership issues.
-    string_values_.push_back(new string(value.ToString()));
+    string_values_.emplace_back(new string(value));
     RenderDataPiece(name, DataPiece(*string_values_.back(), false, true));
   }
   return this;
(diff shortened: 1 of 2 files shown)


acozzette:
@54markov Thank you for cleaning this up. Could you actually try changing it to std::vector<std::string> instead of a vector of std::unique_ptr? I think that should simplify things even more. That also solves another problem, which is that we have to support C++11 for now and std::make_unique is only available starting from C++14.

54markov:
Ok, i will try.

54markov:
> @54markov Thank you for cleaning this up. Could you actually try changing it to std::vectorstd::string instead of a vector of std::unique_ptr? I think that should simplify things even more. That also solves another problem, which is that we have to support C++11 for now and std::make_unique is only available starting from C++14.

Fixed it.

src/google/protobuf/util/internal/default_value_objectwriter.cc
```diff
@@ -156,8 +153,10 @@ DefaultValueObjectWriter* DefaultValueObjectWriter::RenderString(
   } else {
     // Since StringPiece is essentially a pointer, takes a copy of "value" to
     // avoid ownership issues.
-    string_values_.push_back(new string(value));
-    RenderDataPiece(name, DataPiece(*string_values_.back(), true));
+    // To avoid unnecessary call of std::string() constructor, make sure that
+    // StringPiece data is valid and pass only pointer
+    string_values_.push_back(value.empty() ? value : value.data());
```
acozzette: I think you can just call `string_values_.emplace_back(value);` as a way to minimize the number of unnecessary string constructions.

54markov: Ok, and yeah, it's much simpler)

src/google/protobuf/util/internal/default_value_objectwriter.cc
```diff
@@ -156,8 +153,8 @@ DefaultValueObjectWriter* DefaultValueObjectWriter::RenderString(
   } else {
     // Since StringPiece is essentially a pointer, takes a copy of "value" to
     // avoid ownership issues.
-    string_values_.push_back(new string(value));
-    RenderDataPiece(name, DataPiece(*string_values_.back(), true));
+    string_values_.emplace_back(value);
+    RenderDataPiece(name, DataPiece(string_values_.back(), true));
```
acozzette: Sorry to keep going back and forth but I realized that I told you the wrong thing before. It turns out that `DataPiece` hangs onto a reference to the string data here instead of making a copy. That means that `string_values_` can't be a `vector<string>` because when the vector resizes it will move elements around and create dangling references.

I think we should go back to your original idea of using `std::vector<std::unique_ptr<std::string>>`. We have to avoid `std::make_unique` since we don't have that available in C++11, but something like `string_values_.emplace_back(new string(value));` should work if I'm not mistaken.

54markov: Ok, no problem to revert to original pull request's implementation=)
(guess in this time, everything going well, especially unit tests)

acozzette:
Thanks @54markov! I see some Ruby tests failing but it looks like those failures are unrelated so I'll merge this now.

PR #5749, opened 2019-02-19
Update version to 3.7.0

TeBoring:


haon4:
BTW, did we append the -rc2 suffix by hands?

TeBoring:
@acozzette How did you do that?
I think we have to do it manually currently. Different languages use different style.

haon4:
> @acozzette How did you do that?
> I think we have to do it manually currently. Different languages use different style.

https://github.com/protocolbuffers/protobuf/blob/master/update_version.py handles each language separately. I wasn't sure the style for each language before, but I can use this as an example.

acozzette:
I ran Hao's script to update the version to 3.8.0, and then I went through and replaced 3.8.0 with 3.7.0-rc2. I had to do it that way because the script doesn't know how to do rc releases yet, but I wanted to run it anyway to make sure I didn't miss any places that needed to be updated.

TeBoring:
@haon4 SG.

haon4:
https://github.com/protocolbuffers/protobuf/pull/5750 to support the rc suffix in update_version.

PR #5752, opened 2019-02-20
Raise error for JSON overflow encoding in Ruby

jbolinger:
JSON encoding in Ruby can segfault in cases such as:
https://github.com/googleapis/google-cloud-ruby/issues/2876 when deeply nested messages are given. For example, the added test `test_deep_json` will crash with a segmentation fault.

This change adds an extra check to raise an error instead of crashing.

cc\ @quartzmo

### DIFF
--- a/ruby/tests/common_tests.rb
+++ b/ruby/tests/common_tests.rb
@@ -1116,43 +1116,43 @@ module CommonTests
     assert JSON.parse(actual, :symbolize_names => true) == expected
   end
 
+  def value_from_ruby(value)
+    ret = Google::Protobuf::Value.new
+    case value
+    when String
+      ret.string_value = value
+    when Google::Protobuf::Struct
+      ret.struct_value = value
+    when Hash
+      ret.struct_value = struct_from_ruby(value)
+    when Google::Protobuf::ListValue
+      ret.list_value = value
+    when Array
+      ret.list_value = list_from_ruby(value)
+    else
+      @log.error "Unknown type: #{value.class}"
+      raise Google::Protobuf::Error, "Unknown type: #{value.class}"
+    end
+    ret
+  end
+
+  def list_from_ruby(arr)
+    ret = Google::Protobuf::ListValue.new
+    arr.each do |v|
+      ret.values << value_from_ruby(v)
+    end
+    ret
+  end
+
+  def struct_from_ruby(hash)
+    ret = Google::Protobuf::Struct.new
+    hash.each do |k, v|
+      ret.fields[k] ||= value_from_ruby(v)
+    end
+    ret
+  end
+
   def test_deep_json
-    def value_from_ruby(value)
-      ret = Google::Protobuf::Value.new
-      case value
-      when String
-        ret.string_value = value
-      when Google::Protobuf::Struct
-        ret.struct_value = value
-      when Hash
-        ret.struct_value = struct_from_ruby(value)
-      when Google::Protobuf::ListValue
-        ret.list_value = value
-      when Array
-        ret.list_value = list_from_ruby(value)
-      else
-        @log.error "Unknown type: #{value.class}"
-        raise Google::Protobuf::Error, "Unknown type: #{value.class}"
-      end
-      ret
-    end
-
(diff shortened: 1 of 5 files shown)


quartzmo:
@jbolinger Thank you so much for this fix!!

So it looks like `put_ruby_value` needs the same check that is [already present in `putmsg`](https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/encode_decode.c#L1251-L1256)? Is there any way to make this check in just one place, for example in initialization?

ruby/tests/common_tests.rb
```diff
@@ -1116,6 +1116,65 @@ def test_json_emit_defaults_repeated_submsg
...
+
+    struct = struct_from_ruby(JSON.parse(json))
+    assert_equal json, struct.to_json
+
+    encoded = proto_module::MyRepeatedStruct.encode(
+      proto_module::MyRepeatedStruct.new(structs: [proto_module::MyStruct.new(struct: struct)]))
+    assert_equal json, proto_module::MyRepeatedStruct.decode(encoded).structs[0].struct.to_json
+
+    # will overflow
+    json = '{"a":{"a":{"a":[{"a":{"a":[{"a":[{"a":{"a":[{"a":[{"a":{"a":[{"a":[{"a":{"a":{"a":[{"a":"a"}]}}}]}]}}]}]}}]}]}}]}}}'
```
TeBoring: Is it possible to wrap the line?

jbolinger: Sure, I broke it into two lines.

jbolinger:
@quartzmo it may be possible but I'm not sure how to consolidate the check to a single location. In this case, the crash happens because `putmap` can be called after the depth check that you pointed out and `putmap` can increase the depth (so it needs to be checked again).

quartzmo:
@jbolinger :+1 Sounds good, I was just wondering.

jbolinger:
@TeBoring do you know what's going on with all the Mac build failures? I don't think it has anything to do with this change, but let me know if I need to update something.

qingling128:
Hi, thanks for looking into this. Is `3.7.0` supposed to include this fix? (Based on the date, it looks like so.)

I tried the [Ruby snippet](https://github.com/googleapis/google-cloud-ruby/issues/2876#issue-405914583) again with the latest google-cloud-ruby and google-protobuf. Still getting the segfault.

Here are the versions I tried

```
$ gem list | grep "google-cloud-logging\|google-protobuf"
google-cloud-logging (1.6.2)
google-protobuf (3.7.0 x86_64-linux)
```

Any idea?

qingling128:
BTW, when we were trying to reproduce the segmentation fault, one observation is that:
The JSON has to be very specific to trigger it.

Works: (Recursion = 22)
'{"a":{"a":{"a":[{"a":{"a":[{"a":[{"a":{"a":[{"a":[{"a":{"a":[{"a":[{"a":{"a":{"a":"a"}}}]}]}}]}]}}]}]}}]}}}'

Works: (Recursion = 20)
'{"20": {"19": {"18": {"17": {"16": {"15": {"14": {"13": {"12": {"11": {"10": {"9": {"8": {"7": {"6": {"5": {"4": {"3": {"2": {"1": "finish"}}}}}}}}}}}}}}}}}}}}'

Doesn't Work: (Recursion = 21) but returns expected Error Code of max size exceeded
'{"21": {"20": {"19": {"18": {"17": {"16": {"15": {"14": {"13": {"12": {"11": {"10": {"9": {"8": {"7": {"6": {"5": {"4": {"3": {"2": {"1": "finish"}}}}}}}}}}}}}}}}}}}}}

Doesn't Work: (Recursion = 24) Crashes in protobuf'{"a":{"a":{"a":[{"a":{"a":[{"a":[{"a":{"a":[{"a":[{"a":{"a":[{"a":[{"a":{"a":{"a":[{"a":"a"}]}}}]}]}}]}]}}]}]}}]}}}'

Hope that helps.

qingling128:
Turned out that this fix did not get in `3.7.0`. Will need to wait for the next release.

Issue #5725, opened 2019-02-14
protobuf.pc.in does not reflect CXXFLAGS


dickmao:
Currently, downstream libraries such as RProtoBuf fail because `-std=c++11` is required via `configure.ac`, but RProtobuf has no way of knowing this by consulting `pkg-config --cflags protobuf`.

Perhaps `protobuf.pc.in` needs to mention `@CXXFLAGS@` in Cflags.

BSBandme:
Feel free to send a PR for updating the pkg-config input file.

PR #5755, opened 2019-02-20
protobuf.pc.in does not reflect CXXFLAGS

dickmao:
Currently, downstream libraries such as RProtoBuf fail because
`-std=c++11` is required via `configure.ac`, but RProtobuf has no way
of knowing this by consulting `pkg-config --cflags protobuf`.

Cf #5725


acozzette:
@dickmao Thanks for the pull request!

PR #5758, opened 2019-02-21
delete all duplicate empty blanks

daixiang0:
Signed-off-by: Xiang Dai <764524258@qq.com>

PR #5771, opened 2019-02-25
Revert "delete all duplicate empty blanks"

TeBoring:
Reverts protocolbuffers/protobuf#5758


BSBandme:
The fix should have been submitted in #5762.
I think current broken in mac is because of Kokoro updates and m4 isn't installed by default. 

I think we'd add m4 in prepare_build_macos_rc for fixing.

TeBoring:
Ah, I see. Could you help send a fix for that?

BSBandme:
I'll give a try, but may take some time as I'm not very familiar with mac's m4 dependency and how we can install it by `brew install`.

TeBoring:
Maybe check with kokoro team?

PR #5772, opened 2019-02-25
fix for mac build by adding m4

BSBandme:


TeBoring:
If you need to have m4 first in your PATH run:
  echo 'export PATH="/usr/local/opt/m4/bin:$PATH"' >> ~/.bash_profile

PR #5776, opened 2019-02-25
autogen test url not found bug fix

claud-kim:
gtest-1.5.0 url not found fix


claud-kim:
sorry, tag not test..

PR #5777, opened 2019-02-25
Date and time need to be before version

TeBoring:


TeBoring:
pecl (php release tool) checks that.

PR #5780, opened 2019-02-26
fix macos build for High Sierra

BSBandme:


BSBandme:
@anandolee, do you think we should fix the python c++ on mac to adding flag `-Wno-deprecated-register`?

BSBandme:
Submit for now. Maybe create follow up PR for fixing the macos python c++

Issue #5472, opened 2018-12-14
New zlib dependency breaks protobuf as an external bazel repository


mumbleskates:
**What version of protobuf and what language are you using?**
Version: master
Language: Bazel/any

**What operating system (Linux, Windows, ...) and version?**
Ubuntu Linux 18.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
bazel 0.20.0

**What did you do?**
Have the following rule in another repository's WORKSPACE file:

    local_repository(
        name = "com_google_protobuf"
        path = "path/to/protobuf"
    )

...then run `bazel build` on a target that depends upon anything in `com_google_protobuf//...`

**What did you expect to see**
Expected to see it work as it did before the inclusion of the zlib http_archive.

**What did you see instead?**
`ERROR: /home/mumbleskates/.cache/bazel/_bazel_mumbleskates/0123456789abcdef0123456789abcdef/external/com_google_protobuf/BUILD:132:1: no such target '//external:zlib': target 'zlib' not declared in package 'external' defined by /home/mumbleskates/repos/projectname/WORKSPACE and referenced by '@com_google_protobuf//:protobuf'`

Essentially, protobuf can no longer be used as an external dep e.g. as a submodule unless we make a fragile copy of the zlib http_archive repo in our own WORKSPACE file.

**Anything else we should know about your project / environment**
This was working at 81c3e0cf, for example, but now that protobuf has an external dependency on another bazel repo, rather than including the dependencies in its own files or cloned into a submodule, this becomes difficult and frustrating.

I feel that a good fix to this issue might be to clone zlib at a desired commit into a submodule and reference it as a package instead of a separate bazel repository. This avoids the requirement that everyone who references _this_ repository needs to know about, and maintain the correct version of in their own rules, every dependency so declared.

acozzette:
@mumbleskates Thanks for filing this issue. I was wondering if we would run into problems like this when we added the zlib dependency. I can see the rationale for cloning zlib into a submodule, but on the other hand that would make life harder for projects that use protobuf and also independently use zlib. Those projects would then have ODR violations from linking against two copies of zlib.

I have heard that some projects get around these issues by providing a Starlark function that returns all the necessary dependencies. Would it solve your problem if we provided a function like this that you could call?

mumbleskates:
Yes, that'd probably be an ideal solution, I'd appreciate it!

vgrippa:
Im hitting the same issue:

[...]

PR #5784, opened 2019-02-26
create importable function for defining zlib

mumbleskates:


mumbleskates:
#5472

acozzette:
The Bazel test is failing but that is only because the examples directory downloads and tests against master, not against the actual state of the repo in this pull request. As a result the Bazel tests should all pass once this is merged, so I'll go ahead and merge it. I believe the failing Mac OS tests are also due to a separate issue.

PR #5787, opened 2019-02-26
Cherry-pick fixing for mac tests after kokoro updates 

BSBandme:


BSBandme:
The macOS c++ distcheck failure is unrelated (Linux c++ distcheck in fact also failed because of the same error but didn't exit), will submit.

PR #5791, opened 2019-02-28
Revert "Fix Ruby module name generation when the ruby_package option is used"

TeBoring:
Reverts protocolbuffers/protobuf#5735

This change breaks cpp distcheck because it hasn't added new files to Makefile.am


TeBoring:
@jbolinger

jbolinger:
oh, sorry I'm not too familiar with that `Makefile.am` thing. I'll send a new PR.

PR #5794, opened 2019-02-28
Fix Ruby module name generation when the ruby_package option is used (again)

jbolinger:
Reverts #5791, to restore #5735 with the additional change of adding newly added files to `src/Makefile.am'


jbolinger:
@TeBoring let me know if I've missed any more of the relevant build files.

PR #5800, opened 2019-03-01
Remove TODO from FieldDescriptorProto.default_value

dsnet:
While the suggested change in behavior would have been nice,
it has gone un-addressed for 11 years. At this point in time,
switching the encoding to base64 is almost certainly not going
to happen without breaking many users who already had to implement
their own C string parser.

Rather than keeping the TODO, just delete it since it does appear
in the documentation of the generated proto for most languages.


haon4:
Kokoro tests are failing. I think you would have to update the golden generated files too.

haon4:
Closing due to no activity.

PR #5810, opened 2019-03-03
Use fixed size for bool, float and double field in codec

### LINKED ISSUE #5809: C# Serialization Performance - Part I
Version: master
Language: C#

Linux, Windows

MSBUILD 15

bool, double and float should be marked as fixed size. In FieldCodec.cs we create the codecs used by RepeatedField.cs. Having these not be fixed means that on large repeated fields for these fixed length primitives, we will call the size calculator each time, for every single data point (inside CalculatePackedDataSize). I've seen about 40% serialization improvement on repeated fields for these types when they are marked as fixed. 

Straight forward fix is to rename ComputeBoolSize to ComputeFixedBoolSize, ComputeFloatSize to ComputeFixedFloatSize and ComputeDoubleSize to ComputeFixedDoubleSize, then we should change ForBool, ForFloat and ForDouble to send in a dummy value to the computesize functions when creating the codec (we will set the FixedSize attribute to the result).

WilliamWhispell:
this commit fixes #5809 

Removes repeated calls to the size calculator inside CalculatePackedDataSize. This improves performance for large vectors of primitives. This should benefit the repeated fields.


WilliamWhispell:
I signed it!

WilliamWhispell:
I signed it!

WilliamWhispell:
I signed it!

csharp/src/Google.Protobuf/FieldCodec.cs
```diff
@@ -182,7 +182,7 @@ public static FieldCodec<ulong> ForUInt64(uint tag)
         /// <returns>A codec for the given tag.</returns>
         public static FieldCodec<float> ForFloat(uint tag)
         {
-            return new FieldCodec<float>(input => input.ReadFloat(), (output, value) => output.WriteFloat(value), CodedOutputStream.ComputeFloatSize, tag);
+            return new FieldCodec<float>(input => input.ReadFloat(), (output, value) => output.WriteFloat(value), CodedOutputStream.ComputeFixedFloatSize(0f), tag);
```
ObsidianMinor: If you know the size of the float will be 4, you should just write 4 like in ForFixed32 and ForSFixed32 (this applies to ForDouble and ForBool as well).

WilliamWhispell: I was thinking about this. I suppose I was struggling with  why the compute methods should take a value, if they do nothing with it. I assumed the design was to ensure  others didn't write code calling the wrong method, compiler error. So then I thought about hard coding the size here, and my thought is, then if someone changes the size of an encoded value down the road inside ComputeSize, they need to also be aware the value is hard coded in FieldCodec and make the change. As I've written it, a change to CodedOutputStream could be done without touching the FieldCodec logic. I don't like passing in the dummy value either, ideally the fixed size methods wouldn't take an argument. 

I will change this to be hard coded as you've asked.

ObsidianMinor: The "compute size" methods are very old and outdated. Back in [jskeet/protobuf-csharp-port](https://github.com/jskeet/protobuf-csharp-port/blob/master/src/ProtocolBuffers/CodedOutputStream.ComputeSize.cs) they included the size of a tag with the result of the calculation (as the summary implies). But the methods have been changed since then and no longer reflect that design. So it doesn't make sense to call them if they return a fixed value.

WilliamWhispell: ok sounds good

anandolee:
Thank both WilliamWhispell and ObsidianMinor.  
WilliamWhispell, please change the PR title to something like: use fixed size for bool, float and double field in codec. And mention the benefit for repeated field in description

WilliamWhispell:
@anandolee I pushed a change to the compatibility tests so the broken build should pass. Can you rerun them?

WilliamWhispell:
Anything else need from me to get this merged?

csharp/src/Google.Protobuf/FieldCodec.cs
```diff
@@ -72,7 +72,7 @@ public static FieldCodec<ByteString> ForBytes(uint tag)
         /// <returns>A codec for the given tag.</returns>
         public static FieldCodec<bool> ForBool(uint tag)
         {
-            return new FieldCodec<bool>(input => input.ReadBool(), (output, value) => output.WriteBool(value), CodedOutputStream.ComputeBoolSize, tag);
+            return new FieldCodec<bool>(input => input.ReadBool(), (output, value) => output.WriteBool(value), 1, tag);
```
jtattermusch: is the ComputeBoolSize method still needed? Also, I think we should define a constant in https://github.com/protocolbuffers/protobuf/blob/d9ccd0c0e6bbda9bf4476088eeb46b02d7dcd327/csharp/src/Google.Protobuf/CodedOutputStream.ComputeSize.cs#L40 rather than hardcoding the size values here.

same below

WilliamWhispell: @jtattermusch - I've pushed changes based on feedback.

WilliamWhispell:
@anandolee - Can you rerun the tests again following the latest changes?

anandolee:
I will merge it after the tests green

WilliamWhispell:
@anandolee - I assume the broken ruby tests are because my branch is behind master. Do you want me to rebase or is this okay to merge as is?

WilliamWhispell:
Merged master

anandolee:
Ruby failure is not related. Merged. 
Thanks again for the PR

PR #5751, opened 2019-02-19
Add native type setters for Timestamp and Duration in Ruby

jbolinger:
Resolves #3121 by adding additional setters for fields of type `Google::Protobuf::Timestamp` and `Google::Protobuf::Duration`.

This allows a message with a `Timestamp` field to be set via:
```ruby
my_message.timestamp = Time.at(1)
my_message.timestamp
# => <Google::Protobuf::Timestamp: seconds: 1, nanos: 0>
```

Similarly, for `Duration` fields:
```ruby
MyMessage.new(duration: 1.1)
# => MyMessage: duration: <Google::Protobuf::Duration: seconds: 1, nanos: 100000000>>
```

PR #5539, opened 2019-01-04
Updated upb from defcleanup branch and modified Ruby to use it

haberman:

PR #5823, opened 2019-03-04
Fix unit tests for older rubies

jbolinger:
#5726 and #5659 added references to `FrozenError` and `Integer`, which were added to Ruby in 2.5 and 2.4, and will cause test failures for older rubies. This updates the tests so that they are compatible.


TeBoring:
How did you find this problem?
I don't see any of our continuous test was broken.

jbolinger:
It was mostly by accident. I was double checking that the tests passed locally when I sync'd that other PR but didn't realize that I had switched to Ruby 2.4. So, I ran them on all rubies that I have installed on my machine and found these issues.

What versions of Ruby does kokoro test with?

jbolinger:
For example, #5751 should fail now because of this.

TeBoring:
https://github.com/protocolbuffers/protobuf
Our continuous test are still passing.

TeBoring:
https://github.com/protocolbuffers/protobuf/blob/master/tests.sh
Could you try adding ```set -ex``` in this file in another PR?
See whether the ruby test can fail.

jbolinger:
It looks like that file already has a `set -e` in it. I'm not sure if/how that interacts with `eval`, but I'm not sure what to change there.

Have you tried running the tests locally on 2.3?

jbolinger:
@TeBoring looks like someone found a cleaver way to speed up the build 😄 (see #5840)

TeBoring:
Could you sync?

jbolinger:
@TeBoring sure, sync'd.

haberman:
Could you sync again? Sorry, my earlier PR from today (https://github.com/protocolbuffers/protobuf/pull/5539) broke the build, but I didn't know it because a bug made it show as green. I reverted that PR so hopefully yours will pass now.

jbolinger:
@haberman no problem, sync'd. 

Hope this gets the build back in shape. After sending #5840 it looks like that false positive build status might have been going on for awhile?

PR #5835, opened 2019-03-05
C#: add System.Memory dependency

jtattermusch:
Also add useful Span-based methods for ByteString

System.Memory dependency is only added for netstandard2.0 target, netstandard1.0 and net45 stay unchanged (we actually could add System.Memory dependency for net45 too, but it seems that most users that will benefit from Span optimizations will already have access to netstandard2.0).

This PR is mostly preparation for future span-based optimizations, but to make at least some use of the new dependency, two utility ByteString methods are being added.

See https://github.com/grpc/grpc-dotnet/issues/30 for more details about the planned serialization/deserialization API.

### DIFF
--- a/csharp/src/Google.Protobuf/Google.Protobuf.csproj
+++ b/csharp/src/Google.Protobuf/Google.Protobuf.csproj
@@ -7,7 +7,7 @@
     <VersionPrefix>3.7.0</VersionPrefix>
     <LangVersion>6</LangVersion>
     <Authors>Google Inc.</Authors>
-    <TargetFrameworks>netstandard1.0;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;net45</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyOriginatorKeyFile>../../keys/Google.Protobuf.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
@@ -28,8 +28,12 @@
     - Visual Studio.
     -->
   <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
-    <TargetFrameworks>netstandard1.0</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.Memory" Version="4.5.2" />
+  </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All" />
--- a/csharp/src/Google.Protobuf/ByteString.cs
+++ b/csharp/src/Google.Protobuf/ByteString.cs
@@ -67,15 +67,6 @@ namespace Google.Protobuf
             {
                 return new ByteString(bytes);
             }
-
-            /// <summary>
-            /// Provides direct, unrestricted access to the bytes contained in this instance.
-            /// You must not modify or resize the byte array returned by this method.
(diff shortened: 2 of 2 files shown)


jtattermusch:
@jskeet  WDYT about changing the target from `netstandard1.0` to `netstandard1.1`?

CC @JunTaoLuo @davidfowl @JamesNK

jtattermusch:
Looks like this change is mutually exclusive with https://github.com/protocolbuffers/protobuf/pull/5445. I'll reach out to the internal user that requested downgrading LangVersion to see what we can do.

jskeet:
I think I'd prefer to *add* a netstandard2.0 target (which comes with a bunch of benefits for 2.0+ users) and only conditionally include System.Memory things.

(Jan, I know a bit more about the LangVersion change - ping me if you need to.)

csharp/src/Google.Protobuf/Google.Protobuf.csproj
```diff
@@ -7,7 +7,7 @@
     <VersionPrefix>3.7.0</VersionPrefix>
     <LangVersion>6</LangVersion>
     <Authors>Google Inc.</Authors>
-    <TargetFrameworks>netstandard1.0;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.1;net45</TargetFrameworks>
```
JunTaoLuo: This would be considered a breaking change no? Adding another target for netstandard2.0 (or netstandard 1.1 if you want) would probably be better here.

csharp/src/Google.Protobuf/Google.Protobuf.csproj
```diff
@@ -26,11 +26,12 @@
...
     -->
   <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
-    <TargetFrameworks>netstandard1.0</TargetFrameworks>
+    <TargetFrameworks>netstandard1.1</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SourceLink.Create.CommandLine" PrivateAssets="All" Version="2.7.6"/>
+    <PackageReference Include="SourceLink.Create.CommandLine" PrivateAssets="All" Version="2.7.6" />
+    <PackageReference Include="System.Memory" Version="4.5.2" />
```
JunTaoLuo: Adding a new dependency is also usually considered a "breaking change". Maybe make this conditional on the target framework?

csharp/src/Google.Protobuf/Google.Protobuf.csproj
JamesNK: +1

csharp/src/Google.Protobuf/Google.Protobuf.csproj
davidfowl: > Adding a new dependency is also usually considered a "breaking change". Maybe make this conditional on the target framework?

Adding a new dependency isn't a breaking change. Removing one is though. I think we should potentially bump the minor version here.

jtattermusch:
I'll wait for https://github.com/protocolbuffers/protobuf/pull/5838 to be merged first.

csharp/src/Google.Protobuf/Google.Protobuf.csproj
jtattermusch: Minor version will be bumped because this change won't be available before protobuf v3.8.x is out.

jtattermusch:
I added `netstandard2.0` target and added conditional dependency on System.Memory for netstandard2.0 (It would have been possible to add also for net45, but I'm not sure how useful that would be).

JamesNK:
FYI I'm halfway through a large PR to add span-based reader and writer for protobuf. Should be able to create a PR in the next day or two.

jtattermusch:
@anandolee  @jskeet can we review and merge?

anon17:
I think in the `FromStream` method you can use the "avoid a copy" branch on netstandard2.0, MemoryStream supports GetBuffer there.

PR #5840, opened 2019-03-06
Run Ruby tests

jbolinger:
Some of the Ruby tests are successfully terminating without actually running.


jbolinger:
It looks like the older Rubies are failing as expected (without #5823). Not so sure about the others - the failing conformance tests might be due to something that changed while the tests weren't being run? Any ideas?

PR #5852, opened 2019-03-07
Add file option pharo_class_prefix

zecke:
Pharo is a Smalltalk-80 derived programming environment and has
gained a protobuf implementation.


acozzette:
@zecke The way we usually handle new implementations is that we can just allocate an extension number (or numbers) for you to use with descriptor.proto, and then you can use that to extend the descriptor types with your own custom options. If you send us a pull request (or modify this one) to update [options.md](https://github.com/protocolbuffers/protobuf/blob/master/docs/options.md) then you can reserve an extension number that way. Here is some more documention on [custom options](https://developers.google.com/protocol-buffers/docs/proto#customoptions).

PR #5853, opened 2019-03-07
Fix ruby continuous tests

TeBoring:


haberman:
Looks like there are a few more failures?

TeBoring:
Failed ruby 2.3 test seems to be a random failure.

PR #5855, opened 2019-03-07
Add link to continuous test badges

TeBoring:


haon4:
Can you also add the link to dist_install test that I just added?

TeBoring:
Done

PR #5866, opened 2019-03-11
Roll forward Ruby upb changes now that protobuf Ruby build is fixed

haberman:
Updates the Ruby protobuf library to a new version of upb, the C library used for proto parsing. This removes some APIs that allowed setting fields directly on descriptor objects, but these were internal-only. Any code generated by the protocol compiler, and any code that uses the Ruby/Protobuf DSL should continue to work.


TeBoring:
Any update?

haberman:
There is a sporadic test failure in the parallelism test. It's very hard to reproduce locally, and I haven't figured out what is causing it yet.

haberman:
@TeBoring This CL should be ready with the exception of the macOS Timestamp parsing stuff.

TeBoring:
Could you also merge the fix for Timestamp in this change?

haberman:
All tests are passing, but this seems to still be [an intermittent, hard-to-reproduce error](https://source.cloud.google.com/results/invocations/9630ac98-2e0a-4979-9cf1-bebb0914bd40/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fruby23%2Fpresubmit/log):

```
..F
===============================================================================
Failure: test_concurrent_decoding(BasicTest::MessageContainerTest)
/tmp/protobuf/protobuf/ruby/tests/basic.rb:228:in `block (3 levels) in test_concurrent_decoding'
     225:       thds = 2.times.map do
     226:         Thread.new do
     227:           100000.times do
  => 228:             assert_equal o, Outer.decode(raw)
     229:           end
     230:         end
     231:       end
/tmp/protobuf/protobuf/ruby/tests/basic.rb:227:in `times'
/tmp/protobuf/protobuf/ruby/tests/basic.rb:227:in `block (2 levels) in test_concurrent_decoding'
<<BasicTest::Outer: items: {0=><BasicTest::Inner: >}>> expected but was
<<BasicTest::Outer: items: {}>>
diff:
? <BasicTest::Outer: items: {0=><BasicTest::Inner: >}>
```

I can't figure out if this is a new problem introduced by this change or if it existed previously.

I've tried compiling a custom Ruby that uses 1ms timeslices instead of 100ms, and I've tried increasing the number of map entries to 1,000, but neither change has managed to reproduce this problem reliably.

ruby/tests/basic.rb
```diff
@@ -17,7 +17,6 @@ module BasicTest
     add_message "BadFieldNames" do
       optional :dup, :int32, 1
       optional :class, :int32, 2
-      optional :"a.b", :int32, 3
```
TeBoring: Why this is not needed?

haberman: It's an illegal field name. This is checked more robustly now, so the `add_message` call itself will fail now.

PR #5872, opened 2019-03-11
Stop enabling more warnings for CXX compiles.

thomasvl:
For reasons I don't follow, when passing CXXFLAGS to configure, the build
ends up breaking because something along the way fails to detect C++11
support within the third_party/googletest part of the build.


thomasvl:
@haberman @acozzette - fyi - I have no idea why adding some things via CXXFLAGS suddenly stops working, but I had been doing this for a log time.  Is there a better way to add in extra flags past what configure will auto detect?

Issue #5815, opened 2019-03-04
3.7.0: make check fails  on OSX: invalid integral value 'g' in '-Og'


ergl:
Installing the latest stable version (3.7.0) from source (using Homebrew) errors with

```
echo "int main(int, char**) { return 0; }" >> no_warning_test.cc
clang++ -DHAVE_CONFIG_H -I. -I..    -Qunused-arguments -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare  -DNDEBUG -c -o google/protobuf/testing/zcgzip.o google/protobuf/testing/zcgzip.cc
clang++ -DHAVE_CONFIG_H -I. -I..    -Qunused-arguments -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare  -DNDEBUG -c -o google/protobuf/testing/zcgunzip.o google/protobuf/testing/zcgunzip.cc
clang++ -DHAVE_CONFIG_H -I. -I..  -I./../third_party/googletest/googletest/include  -Qunused-arguments -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare  -DNDEBUG -c -o google/protobuf/testing/test_plugin-file.o `test -f 'google/protobuf/testing/file.cc' || echo './'`google/protobuf/testing/file.cc
clang++ -DHAVE_CONFIG_H -I. -I..    -Qunused-arguments -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wextra -Werror -Wno-unused-parameter -Og -DNDEBUG -c -o no_warning_test-no_warning_test.o `test -f 'no_warning_test.cc' || echo './'`no_warning_test.cc
/bin/sh ../libtool  --tag=CXX   --mode=link clang++ -Qunused-arguments -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare  -DNDEBUG -Qunused-arguments -pthread  -o zcgzip google/protobuf/testing/zcgzip.o  libprotobuf.la -lz
error: invalid integral value 'g' in '-Og'
error: invalid integral value 'g' in '-Og'
make[2]: *** [no_warning_test-no_warning_test.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[...]

PR #5873, opened 2019-03-11
Avoid passing -Og flag to C++ compiler

acozzette:
It appears that some versions of Clang do not support this flag, and
this came up as a problem for Homebrew builds in issue #5815.


TeBoring:
Since when does clang support "-Og" and since when does clang support c++11?

acozzette:
Clang seems to support `-Og` starting from version 4.0 which was released in March 2017. It has had C++11 support since 3.3 (I'm not sure when that was released but I think some time after April 2013).

TeBoring:
Can we test with clang that doesn’t support -Og? Or can we have some script
test to make sure there is no -Og used in Makefile?

acozzette:
I think testing with an older version of Clang might be complicated but I went ahead and added a simple test to just make sure the `-Og` flag doesn't appear in the makefile.

PR #5838, opened 2019-03-05
Bump target frameworks of C# programs from netcoreapp1.0 to netcoreapp2.1

ObsidianMinor:
Moved global.json up to root of repo, changed SDK version to 2.2.100.
Changed .NET Core SDK in dockerfile for kokoro to version 2.2.100.

.NET Core 1.0 is reaching end of life, it's hard to install, and it's getting in the way of everyone's fun.

This should fix issues with #5802

PR #5876, opened 2019-03-12
C# testing improvements

jtattermusch:
- linux: switch tests to use netcoreapp2.1 (**currently some tests are being skipped due to the failure to update buildall.sh in #5838**)
- add kokoro testing scripts for windows (will be enabled on pull requests later)
- a bit more cleanup

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -57,6 +57,7 @@ csharp_EXTRA_DIST=                                                           \
   csharp/README.md                                                           \
   csharp/build_packages.bat                                                  \
   csharp/build_tools.sh                                                      \
+  csharp/buildall.bat                                                        \
   csharp/buildall.sh                                                         \
   csharp/generate_protos.sh                                                  \
   csharp/install_dotnet_sdk.ps1                                              \
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -38,7 +38,7 @@ dotnet restore
 dotnet build -c %configuration% || goto error
 
 echo Testing C#
-dotnet test -c %configuration% -f netcoreapp1.0 Google.Protobuf.Test\Google.Protobuf.Test.csproj || goto error
+dotnet test -c %configuration% -f netcoreapp2.1 Google.Protobuf.Test\Google.Protobuf.Test.csproj || goto error
 dotnet test -c %configuration% -f net451 Google.Protobuf.Test\Google.Protobuf.Test.csproj || goto error
 
 goto :EOF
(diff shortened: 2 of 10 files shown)


jtattermusch:
CC @jskeet

jtattermusch:
I am getting a failure on netcoreapp2.1, which we didn't catch in #5838. I will try to fix in this PR.

```
Failed   NaNValuesComparedBitwise
Error Message:
   Expected: not equal to -33554432
  But was:  -33554432

Stack Trace:
   at Google.Protobuf.EqualityTester.AssertInequality[T](T first, T second) in /usr/local/google/home/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/EqualityTester.cs:line 60
   at Google.Protobuf.Collections.RepeatedFieldTest.NaNValuesComparedBitwise() in /usr/local/google/home/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs:line 753
```

jtattermusch:
Turns out that `SampleNaNs.SignallingFlipped` has the same hashcode as `SampleNaNs.PayloadFlipped` under netcore2.1, which is why the test is failing. :-)

jskeet:
Ah, that'll be a fix to the issue I raised a long time ago :) Previously `a.Equals(b)` could return true, without `a.GetHashCode()` being equal to `b.GetHashCode()`.

What's relying on the hash codes being the same to determine equality? Is that a production code issue, or our test infrastructure?

jtattermusch:
It's just our testing code 
https://github.com/protocolbuffers/protobuf/blob/70dc39ed3d677c39c6b0abd3fa76bfa0db54ac14/csharp/src/Google.Protobuf.Test/EqualityTester.cs#L60  (it assumes that normally hash codes would be very likely to be different for non-equal objects).  It looks like all the NaNs have the same hashcode though.

jskeet:
Right. So we *could* just remove that part of the code. It's annoying, as in most cases that's a useful sanity check. Perhaps we should have an optional parameter for "check hash codes as well" that defaults to true, and we could explicitly pass in false for these tests, with a comment?

jtattermusch:
> Right. So we _could_ just remove that part of the code. It's annoying, as in most cases that's a useful sanity check. Perhaps we should have an optional parameter for "check hash codes as well" that defaults to true, and we could explicitly pass in false for these tests, with a comment?

Sg, I made the check for hashcodes being not equal optional.

jtattermusch:
This is now ready to review.

PR #5890, opened 2019-03-14
Update file list for protobuf_lite

schoppmp:
protobuf_lite now needs `strtod.cc`. The changes were obtained by running `./update_file_lists.sh`.


acozzette:
Thanks @schoppmp!

PR #5891, opened 2019-03-14
Fix source stepping by upgrading sourcelink

### LINKED ISSUE #5850: Update Source Link package
C#

Google.Protobuf is using `SourceLink.Create.CommandLine` for Source Link. That is an old package that has been replaced with `Microsoft.SourceLink.*` packages.

More info: https://github.com/ctaggart/SourceLink

This is the modern package: https://www.nuget.org/packages/Microsoft.SourceLink.GitHub

jtattermusch:
Fixes #5850 (according to my experiments)

### DIFF
--- a/csharp/build_packages.bat
+++ b/csharp/build_packages.bat
@@ -1,7 +1,7 @@
 @rem Builds Google.Protobuf NuGet packages
 
 dotnet restore src/Google.Protobuf.sln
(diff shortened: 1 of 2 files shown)


jtattermusch:
CC @JamesNK

jtattermusch:
This would potentially be a good candidate for backporting  to 3.7.x branch and releasing as part of the upcoming v3.7.1 release (if we can make it on time).

PR #5896, opened 2019-03-14
Ruby error messages in 3.7.x

jbolinger:
Cherry-pick of #5659.


TeBoring:
Seems ruby tests are failing.

jbolinger:
Oh, I think I missed one of `compatibility_tests` that you fixed in #5853. Let me try to reproduce this locally and if that's the case I'll have to cherry-pick an additional commit from that PR.

jbolinger:
@TeBoring I think that was it. Can you run the tests again?

PR #5901, opened 2019-03-15
Fix some UBSAN warnings.

htuch:
* external/com_google_protobuf/src/google/protobuf/stubs/strutil.cc:1122:9: runtime error: negation of -9223372036854775808 cannot be represented in type 'google::protobuf::int64' (aka 'long'); cast to an unsigned type to negate this value to itself

* Bad external/com_google_protobuf/src/google/protobuf/text_format.cc:1320:14: runtime error: null pointer passed as argument 1, which is declared to never be null
  /usr/include/string.h:62:62: note: nonnull attribute specified here


acozzette:
@htuch If you don't mind could you either rebase this onto master or merge master into your branch? The Linux Ruby 2.6 test shows a strange crash. I don't see how it could be related to this pull request but maybe it's a separate issue that has since been fixed. For the other test failure (MacOS Ruby 2.3), let's not worry about that because I have a separate fix that just needs to be merged.

htuch:
@aconchillo done

acozzette:
Thanks @htuch!

Issue #5053, opened 2018-08-16
Is the protoc binary available in the GitHub Releases zip files statically compiled? If not, can we get statically compiled binaries?


peter-edge:
See https://github.com/uber/prototool/issues/194 for the issue and https://github.com/uber/prototool/issues/193 for reproduction. If you do `make run` there, you get:

```
docker run -it uber/prototool-protoc-issue /protoc --version
standard_init_linux.go:190: exec user process caused "no such file or directory"
```

If you drop into a shell via `make shell`, you get:

```
/ # /protoc
/bin/sh: /protoc: not found
```

The binary is there and executable:

```
/ # ls -l /protoc
-rwxr-xr-x    1 root     root       4539800 Aug 16 08:22 /protoc
```

This appears to be similar to https://github.com/gin-gonic/gin/issues/1178 in some way, in that it appears protoc needs additional assets to run.

I searched the GitHub issues but did not immediately see something similar filed. There may be an easy answer to this, but at the least I thought we'd have a conversation about it - being able to easily download `protoc` to an Alpine image and run it would be something I could see many people using.

johanbrandhorst:
My `protoc` binary (3.6.1) has a number of shared library dependencies:
```
$ ldd /usr/bin/protoc
    linux-vdso.so.1 (0x00007ffe34fda000)
    libprotoc.so.16 => /usr/lib/libprotoc.so.16 (0x00007f8286e0b000)
    libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007f8286c7c000)
    libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f8286c62000)
    libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f8286c41000)
    libc.so.6 => /usr/lib/libc.so.6 (0x00007f8286a7d000)
    libprotobuf.so.16 => /usr/lib/libprotobuf.so.16 (0x00007f8286594000)
    libz.so.1 => /usr/lib/libz.so.1 (0x00007f828637b000)
    libm.so.6 => /usr/lib/libm.so.6 (0x00007f82861f6000)
    /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f828748d000)
```

peter-edge:
Well I guess I've been out of the linking game for a while, heh. There we go:

```
/ # ldd /protoc
[...]

PR #5904, opened 2019-03-16
Release statically linked protoc for Linux

### LINKED ISSUE #5875: Released protoc 3.7.0 require libatomic1 (not statically linked?)
Version: v3.7.0
Language: n/a

Linux
Docker image `debian:stretch-slim`

None

**What did you do?**
1. docker run -it --rm debian:stretch-slim
2. apt update && apt -y install wget unzip
3. wget https://github.com/protocolbuffers/protobuf/releases/download/v3.7.0/protoc-3.7.0-linux-x86_64.zip
4. unzip protoc-3.7.0-linux-x86_64.zip
5. ./bin/protoc

protoc runs and prints some help text

**What did you see instead?**

loader error

```
root@6a6b297121e5:/# ./bin/protoc
./bin/protoc: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
```
It runs if I install libatomic

`apt -y install libatomic1`

Is this an intended change? I guess it's supposed to be statically linked into the prebuilt binaries. v3.7.0 is the first version I see this issue with.

anuraaga:
There was some interest in a statically linked protoc in #5053. Statically linked binaries can prevent loader issues when running protoc such as in the linked issues while making the build process simpler, as not as much care has to be taken in using dependencies outside of system dependencies. This means that modern OS distributions can be used to build protoc instead of intentionally using an old one to link to old versions of libc.

## Pros of static link
- All Linux users will be able to run the binary, no matter what distribution
- No restrictions on build environment - an OS with modern GCC and libc can be used without issue since the resulting binaries are loadable anywhere

## Cons of static link
- Larger binary. Comparing 3.7.0, a statically linked one is 5.7MB vs 4.7MB for the released dynamically linked one

I didn't see any strong objection to statically linking in #5053 and it seemed to get closed since the immediate issue at hand was solved. But as we've seen with `libatomic`, these issues will continue to crop up as long as the binaries are dynamically linked. Hopefully this PR makes sense - if there is a reason to avoid statically linking though, feel free to close it. Merging and using it will fix the `libatomic` issue though :) The best part of it, IMO, is that the same process is used to build all the targets - currently, aarch64 and ppc64 are built on the Kokoro debian host and others built in a CentOS docker container. This discrepancy doesn't seem intuitive or ideal. I suspect the `libatomic` release issues happened in the first place since developers use debian, which includes `libatomic` automatically when installing a build chain, and not centos.

## Changes
- Switch build image from old centos to modern debian, which includes good support for cross-compiling all the target architectures
- Pass `--static` to `LDFLAGS` on linux, ensuring resulting binary is not a dynamic executable
- Update dependency checker to verify the linux binaries are statically linked
- Kokoro script uses the same docker image to build all Linux targets

## Testing

I ran all builds using the changed `Dockerfile` using something like

```
$ docker build -t protobuf/build protoc-artifacts
$ docker run -it --rm protobuf/build
$ apt-get install -y git
$ git clone https://github.com/anuraaga/protobuf
$ git fetch origin static-link
$ git checkout static-link
$ ./autogen.sh
$ cd protoc-artifacts
$ ./build-protoc.sh linux x86_64 protoc
$ ./build-protoc.sh linux x86_32 protoc
$ ./build-protoc.sh linux aarch_64 protoc
$ ./build-protoc.sh linux ppcle_64 protoc
$ file target/linux/aarch_64/protoc.exe
$ file target/linux/ppcle_64/protoc.exe
```

All the built binaries ran fine. The aarch / ppcle ones I think have the qemu runtime layer statically linked in, but the binaries seem to not be much larger as a result so hopefully this is fine.

Fixes #5875


anuraaga:
@acozzette Would be great if you could consider this as a solution for #5875

acozzette:
@anuraaga Thanks so much for working on this but I'm a bit hesitant to accept this change, mostly because I have heard that statically linking `libc` can be unreliable. Some people seem to say glibc in particular is not safe to link against statically. e.g. [here](https://github.com/cockroachdb/cockroach/issues/3392). Maybe this is something we should consider again in the future but now that I found a small fix for the `libatomic` issue in #5922 I think we should probably stick with that for now. Producing `protoc` binaries that work portably has definitely been a source of problems, so sorry about the trouble and thanks again for looking at this.

PR #5906, opened 2019-03-17
Replace strptime with custom implementation

### LINKED ISSUE #5884: [Ruby] Decoding JSON raises ParseError on Windows

Version: v3.7.0
Language: Ruby

Windows 10

Ruby 2.6.1

**What did you do?**

Install latest google-cloud-pubsub and google-protobuf gems, and run this code:

```ruby
require "google/cloud/pubsub/v1"

now = Time.now
now_ts = Google::Protobuf::Timestamp.new seconds: now.to_i, nanos: now.nsec
msg = Google::Cloud::PubSub::V1::PubsubMessage.new publish_time: now_ts

msg.to_json #=> "{\"attributes\":{},\"publishTime\":\"2019-03-13T15:40:43.729054800Z\"}"

decoded_msg = Google::Cloud::PubSub::V1::PubsubMessage.decode_json msg.to_json
#=> Traceback (most recent call last):
#           1: from pubsub.rb:9:in `<main>'
#   pubsub.rb:9:in `decode_json': Error occurred during parsing: error parsing timestamp: mingw doesn't support strptime (Google::Protobuf::ParseError)
```

The ability to create a new protobuf resource object by decoding JSON that contains a timestamp.

**What did you see instead?**

A `Google::Protobuf::ParseError` error.

This appears to only happen on Windows, and [it looks intentional](https://github.com/protocolbuffers/protobuf/blob/v3.7.0/ruby/ext/google/protobuf_c/upb.c#L14125-L14129). But Ruby will parse the timestamp string without issue, so I don't see why Windows couldn't use `Time.parse` instead of `strptime`.

```ruby
require "time"

time_obj = Time.parse "2019-03-13T15:40:43.729054800Z"
timestamp_obj = Google::Protobuf::Timestamp.new seconds: time_obj.to_i, nanos: time_obj.nsec
```

TeBoring:
Fix #5884

### DIFF
--- a/ruby/ext/google/protobuf_c/upb.c
+++ b/ruby/ext/google/protobuf_c/upb.c
@@ -14288,7 +14288,7 @@ static void start_timestamp_zone(upb_json_parser *p, const char *ptr) {
 static bool end_timestamp_zone(upb_json_parser *p, const char *ptr) {
   size_t len;
   const char *buf;
-  int hours;
+  int hours = 0;
   int64_t seconds;
   const char *seconds_membername = "seconds";
 
@@ -14308,12 +14308,11 @@ static bool end_timestamp_zone(upb_json_parser *p, const char *ptr) {
     if (buf[0] == '+') {
       hours = -hours;
     }
-
-    p->tm.tm_hour += hours;
   }
 
   /* Normalize tm */
-  seconds = mktime(&p->tm);
+  seconds = timegm(&p->tm);
+  seconds += 3600 * hours;
 
   /* Check timestamp boundary */
   if (seconds < -62135596800) {
--- a/ruby/ext/google/protobuf_c/upb.c
+++ b/ruby/ext/google/protobuf_c/upb.c
@@ -1,4 +1,5 @@
-// Amalgamated source file
+/* Amalgamated source file */
+#define _XOPEN_SOURCE 700
 #include "upb.h"
 
 #if UINTPTR_MAX == 0xffffffff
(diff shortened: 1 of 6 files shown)


ruby/ext/google/protobuf_c/upb.c
```diff
@@ -14106,49 +14128,100 @@ static bool end_duration_base(upb_json_parser *p, const char *ptr) {
...
+  const char *buf;
+  char *end;
+  int val;
+
+  /* atoi() and friends unfortunately do not support specifying the length of
+   * the input string, so we need to force a copy into a NULL-terminated buffer. */
+  multipart_text(p, "\0", 1, false);
+
+  buf = accumulate_getptr(p, &len);
+  val = atoi(buf);
```
haberman: please use strtol() instead so you can check for errors.

TeBoring: strtol cannot handle leading 0. It parses 08 as 0.

haberman: If our regex is already validating the input, then I'm ok with atoi().

ruby/ext/google/protobuf_c/upb.c
```diff
@@ -14311,7 +14311,7 @@ static bool end_timestamp_zone(upb_json_parser *p, const char *ptr) {
   }
 
   /* Normalize tm */
-  seconds = timegm(&p->tm);
+  seconds = mktime(&p->tm) - timezone;
```
knu: This does not work on *BSD platforms where `timezone` is not a long int variable but a [function](https://www.freebsd.org/cgi/man.cgi?query=timezone).

Is this change due to timegm() not available everywhere?

PR #5921, opened 2019-03-20
Down integrate to Github

haon4:


BSBandme:
Sorry but I didn't receive the log updating CL reviewing request...
I think the new script also should generate the log updating CL?

PR #5929, opened 2019-03-21
Fix "the the".

benjaminp:


acozzette:
@benjaminp Thanks for fixing these!

PR #5932, opened 2019-03-21
Revert "Remove unused f expression from jscode gen. (#5573)"

dankurka:
The removed variable here is actually used breaking downstream bazel
builds.


acozzette:
LGTM

PR #5953, opened 2019-03-26
Merge 3.7.x into master

acozzette:
The 3.7.x branch has a bunch of fixes for 3.7.1, so let's merge them into master.


acozzette:
The CLA is OK because this pull request is just merging commits that have already gone through the CLA process.

TeBoring:
upb for php and ruby is complicate to merge. I will send append a PR to this change to fix that.

TeBoring:
bazel test still fails. @acozzette could you take a look?

acozzette:
@TeBoring I made one more commit which I think should fix that

PR #5954, opened 2019-03-26
Ignore make.sh

songy23:


acozzette:
Where does this `make.sh` file come from? I can't remember ever seeing that file appear.

songy23:
After running the C++ installation instructions (https://github.com/protocolbuffers/protobuf/blob/master/src/README.md#c-installation---unix):
> $ ./configure
$ make
$ make check
$ sudo make install
$ sudo ldconfig # refresh shared library cache.

Then I saw
```bash
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	make.sh
```

acozzette:
That's interesting, I have never seen that file before and I can't find any documentation referencing it. I guess adding it to the `.gitignore` theoretically can't hurt but I am still curious to know what it is first. Could you post the contents of the file?

songy23:
Apologies, for some reason I cannot reproduce this when running the installation commands again (the original file was removed with `git clean`). I guess this is a temporary intermediate file generated by the commands. I'll come back update this PR if I saw this file again.

PR #5962, opened 2019-03-27
Fix preproc definition checks

mathstuf:
Fixes preprocessor definitions assumed to be `0` by being undefined.

~~The `mathlimits.cc` change is what worries me most, but the code is obviously wrong right now by using `__cpluscplus` instead of `__cplusplus`. If this change is correct, the `DEF_*` macros and their usages should probably just go away.~~ Resolved in https://github.com/protocolbuffers/protobuf/pull/5962#discussion_r269690058

### DIFF
--- a/src/google/protobuf/stubs/mathlimits.cc
+++ b/src/google/protobuf/stubs/mathlimits.cc
@@ -40,41 +40,10 @@
 namespace google {
 namespace protobuf {
 
-// MSVC++ 2005 and older compilers think the header declaration was a
-// definition, and erroneously flag these as a duplicate definition.
-#if defined(COMPILER_MSVC) || __cpluscplus < 201103L
-
 #define DEF_COMMON_LIMITS(Type)
 #define DEF_UNSIGNED_INT_LIMITS(Type)
 #define DEF_SIGNED_INT_LIMITS(Type)
 #define DEF_PRECISION_LIMITS(Type)
-
-#else
-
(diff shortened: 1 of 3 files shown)


src/google/protobuf/stubs/mathlimits.cc
```diff
@@ -40,42 +40,11 @@
 namespace google {
 namespace protobuf {
 
-// MSVC++ 2005 and older compilers think the header declaration was a
-// definition, and erroneously flag these as a duplicate definition.
-#if defined(COMPILER_MSVC) || __cpluscplus < 201103L
```
acozzette: Thanks for catching this. I think we can just delete the `#if` case and go with the `#else` case, though. Starting from version 3.6 we require C++11 support, so I don't think we need to worry about old versions of MSVC anymore.

mathstuf: Well, the `#else` side fails to compile on my machine (Linux/GCC 8.3.1) with redefinition errors. Since everything has always been using the `#if` side since 2015, I think that's the side that should be used.

acozzette: Ah OK, got it, somehow I was thinking the condition would always be false but I see now that it's always true.

mathstuf: OK, I've pushed up a change which removes the no-op expansions then.

PR #5966, opened 2019-03-28
Add a new conformance test for top level JSON null

haon4:


conformance/failure_list_ruby_mac.txt
```diff
@@ -70,3 +70,5 @@ Required.Proto3.JsonInput.IgnoreUnknownJsonObject.ProtobufOutput
 Required.Proto3.JsonInput.IgnoreUnknownJsonString.ProtobufOutput
 Required.Proto3.JsonInput.IgnoreUnknownJsonTrue.ProtobufOutput
 Recommended.Proto3.JsonInput.FieldMaskInvalidCharacter
+Required.Proto3.JsonInput.EmptyFieldMask.JsonOutput
+Required.Proto3.JsonInput.RejectTopLevelNull
```
TeBoring: Still needed?

haon4: Yes, it seems like parsing top level null only succeeds in PHP and Ruby, so these tests need to be whitelisted as they are not consistent with other languages.

TeBoring: Should top level null succeed with top-level NullValue?

haon4: Hmm. Should we leave that as undefined behavior? Parsing null into top-level NullValue doesn't seem to have any real use-case since NullValue cannot be used to represent anything other than null. Why would anyone want to deserialize anything into a top-level NullValue?

TeBoring: undefined looks fine to me. So we don't need a test for that, I guess?

TeBoring: Oh, you mean just top level null for top-level NullValue is undefined?
That's also fine for me.

haon4: Yea, I just meant that we shouldn't define top level null for top-level NullValue. I can't think of a good use-case for that. null for other top-level messages can happen though, and I think we should at least define what we want there.

PR #5981, opened 2019-04-02
Add missing declaration in protobuf_headers

meteorcloudy:
To prevent us from getting such error with sandbox or remote execution:
```
bazel-out/k8-opt/genfiles/tensorflow/core/lib/core/error_codes.pb.h:10:40: fatal error: google/protobuf/port_def.inc: No such file or directory
 #include <google/protobuf/port_def.inc>
                                        ^
```

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -203,7 +203,7 @@ cc_library(
 # TODO(keveman): Remove this target once the support gets added to Bazel.
 cc_library(
     name = "protobuf_headers",
-    hdrs = glob(["src/**/*.h"]),
+    hdrs = glob(["src/**/*.h", "src/**/*.inc"]),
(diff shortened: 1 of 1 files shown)


acozzette:
@meteorcloudy Could you make this pull request against the `3.7.x` branch? That will make things easier if we do another release with this fix in the 3.7 series.

meteorcloudy:
But we still need this change for master, right?

acozzette:
@meteorcloudy Yes but we will make sure to merge the 3.7.x branch into master again so that master picks up all the fixes we make for 3.7.

meteorcloudy:
I see, thanks for the explanation!

PR #5982, opened 2019-04-02
support specify C++ STL for android

yejw5:
As [bazel docs](https://docs.bazel.build/versions/0.21.0/android-ndk.html#configuring-the-stl) says, we can configure C++ STL by `crosstool_top`.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -69,9 +69,25 @@ config_setting(
     },
 )
 
+config_setting(
+    name = "android-libcpp",
+    values = {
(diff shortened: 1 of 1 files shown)


yejw5:
@acozzette Is there any schedule to release a new edition include this commit?

acozzette:
We don't have a precise schedule but we do plan to do at least one release this quarter.

yejw5:
@acozzette Many thanks. I'll keep an eye on this.

PR #5985, opened 2019-04-03
For 3.7.x: Add missing declaration in protobuf_headers

meteorcloudy:
To prevent us from getting such error with sandbox or remote execution:
```
bazel-out/k8-opt/genfiles/tensorflow/core/lib/core/error_codes.pb.h:10:40: fatal error: google/protobuf/port_def.inc: No such file or directory
 #include <google/protobuf/port_def.inc>
                                        ^
```

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -201,7 +201,7 @@ cc_library(
 # TODO(keveman): Remove this target once the support gets added to Bazel.
 cc_library(
     name = "protobuf_headers",
-    hdrs = glob(["src/**/*.h"]),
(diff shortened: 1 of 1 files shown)


meteorcloudy:
@acozzette

meteorcloudy:
When can we expect a patch release for 3.7.1?

acozzette:
This is the only  bug I know of in 3.7.1 so I am not sure if we will do a patch release just for this. We might wait and see if we get a few more bug reports that need to be fixed before doing a 3.7.2 release. As a workaround can you just point directly at this Git commit?

PR #5994, opened 2019-04-06
Down integrate to Github

haon4:


java/lite/pom.xml
```diff
@@ -0,0 +1,345 @@
...
+
+  <build>
+     <!-- Include core protos in the bundle as resources -->
+     <resources>
+      <resource>
+        <directory>${protobuf.source.dir}</directory>
+        <includes>
+          <include>google/protobuf/any.proto</include>
+          <include>google/protobuf/api.proto</include>
+          <include>google/protobuf/descriptor.proto</include>
```
BSBandme: Is descriptor.proto used in lite test?

haon4: I don't think so, otherwise the lite tests wouldn't build. I intentionally removed descriptor.proto and compiler plugin.proto to ensure that Lite cannot depend on them.

BSBandme: Then why not delete the descriptor.proto here?

BSBandme:
Maybe also make some changes to [this doc](https://github.com/protocolbuffers/protobuf/blob/master/java/lite.md)?

BSBandme:
Would you please authorize [Reviewable](https://reviewable.io) as an accepted third party application in the organization setting?

I can't open the first commit in github for reviewing as the commit has too many files. I just left some comments in Reviewable.

java/lite/pom.xml
haon4: Done.

PR #6000, opened 2019-04-08
Small patch to handle -Wfloat-equal warnings

senhalil:
To check whether a field is set, we make "field != 0" comparisons; however, this makes the generated *.pb.cc files to throw the following warning if float-equal flag is set:

```
file.pb.cc:494:34:  warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
   // float field_name = 10;
   if (this->field_name() != 0) {
                             ^
```

In case of a float, we can can avoid the float-equal warnings by using `<=` and `>=` to handle `==`.

https://groups.google.com/d/msg/protobuf/H8_D1hfg3p4/mSJKQB75AAAJ

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -178,6 +178,10 @@ bool EmitFieldNonDefaultCondition(io::Printer* printer,
     } else if (field->cpp_type() == FieldDescriptor::CPPTYPE_MESSAGE) {
       // Message fields still have has_$name$() methods.
       format("if ($prefix$has_$name$()) {\n");
+    } else if (field->cpp_type() == FieldDescriptor::CPPTYPE_DOUBLE 
+               || field->cpp_type() == FieldDescriptor::CPPTYPE_FLOAT) {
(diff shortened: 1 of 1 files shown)


senhalil:
I signed it!

senhalil:
The same result can be achieved by doing the following check (instead of two inequalities); however, this introduces `<cmath>` to every proto generated file and I chose the inequality method to keep it simpler.

```
 std::fpclassify(field) == FP_ZERO
```

senhalil:
Hi @acozzette I can see that there is an issue with a test and it asks me to include " release note: no " but I am not really sure where to include this message.

acozzette:
@senhalil You can add the label "release notes: yes" or "release notes: no" to control whether this PR will appear in the release notes for the next release. I added "yes" but you can change it if you want.

acozzette:
Thanks, @senhalil!

senhalil:
Thanks for the help @acozzette ! In fact I couldn't find how to add a label, still can't find it. I thought you need to be a maintainer to add/change labels. I would appreciate if you correct me on that if I am wrong.

acozzette:
Thanks for letting me know, we might need to change some permissions to allow all pull request authors to change those labels.

PR #6001, opened 2019-04-09
Down integrate to Github

haon4:
Down integration + include the changes to enable building JavaLite module


haon4:
> Maybe also make some change to [the doc](https://github.com/protocolbuffers/protobuf/blob/master/java/lite.md)?

Doing that in a separate PR since this one is already too big.

Issue #5194, opened 2018-09-27
are google.protobuf and google.protobuf.pyext supposed to be namespace packages?


benjaminp:
I'm trying to understand why 786f80fe6632902be74f90a577d75e8641494abd added the namespace package stanza to `google/protobuf/__init__.py` and `google/protobuf/pyext/__init__.py`. I see the `google` namespace has a fraught history, but is it actually expected that other packages will put code in `google.protobuf` or `google.protobuf.pyext`? Note also `google/protobuf/__init__.py` technically violates the requirement that a namespace packages's `__init__.py` must not include any other code and data aside from the namespace declaration.

I think either the namespace declaration code should be removed from `google/protobuf/__init__.py` and `google/protobuf/pyext/__init__.py`, or these namespace packages should be declared along with the `google` namespace in `setup.py`.

benjaminp:
#7902

PR #6002, opened 2019-04-09
Remove namespace package boilerplate from google.protobuf and google.protobuf.pyext.

benjaminp:
setup.py does not declare these packages to be namespace packages, so they shouldn't have the boilerplate.

Fixes https://github.com/protocolbuffers/protobuf/issues/5194.


benjaminp:
Thank you to 66e3562aafce093d30473555248f22b3fdc88aad.

PR #6003, opened 2019-04-09
Update documentation for JavaLite

haon4:


haon4:
I just created https://github.com/protocolbuffers/protobuf/pull/6004 to put the old Java Lite documentation in the old frozen javalite branch. Added a link from this doc to that old doc. We will probably never delete that branch and that lite.md doc any more, but we will likely need to keep updating this doc.

BSBandme:
Got it. Then let's check in both.

PR #6011, opened 2019-04-10
Remove unused constants

keith:
When compiling protobuf these produce warnings


keith:
It looks like Mergeable is failing because I need labels here, but I can't add those. It might be nice to add instructions about this to the contributing guidelines so folks know it's expected for their PRs to be red after submission

acozzette:
Sorry about the issues around Mergeable, that is something we just introduced and we're still working out the details.

keith:
Thanks!

PR #6014, opened 2019-04-10
Make cold_threshold_ look used to fix warning

keith:
This field isn't used yet, but it will be in the future so we don't want
to remove it. This fixes the warning in the meantime.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -521,7 +521,11 @@ class ColdChunkSkipper {
 // Tuning parameters for ColdChunkSkipper.
 const double kColdRatio = 0.005;
 
-bool ColdChunkSkipper::IsColdChunk(int chunk) { return false; }
+bool ColdChunkSkipper::IsColdChunk(int chunk) {
+  // Mark this variable as used until it is actually used
+  (void) cold_threshold_;
+  return false;
+}
 
 void ColdChunkSkipper::OnStartChunk(int chunk, int cached_has_bit_index,
                                     const std::string& from,


acozzette:
@keith Thank you for sending this cleanup, but unfortunately we have to keep this cold threshold code in place. While it does appear unused, we are using it within Google and just haven't been able to open source this functionality yet.

keith:
@acozzette do you have a recommendation on how we can silence this warning without removing this field?

acozzette:
Hmm, good question. Maybe in `ColdChunkSkipper::IsColdChunk()` we could add something like `(void) cold_threshold_;` just to reference that variable in a trivial way?

acozzette:
If you need to delete the `kColdRatio` variable as well then that is probably fine because it's easy to tweak our internal fork to retain that line.

src/google/protobuf/compiler/cpp/cpp_message.cc
```diff
@@ -512,15 +510,11 @@ class ColdChunkSkipper {
   const std::vector<std::vector<const FieldDescriptor*>>& chunks_;
   const std::vector<int>& has_bit_indices_;
   const AccessInfoMap* access_info_map_;
-  const double cold_threshold_;
```
kastiglione: @keith @acozzette two possible options:

Add `__attribute__((used))` to this field.

Or, wrap this in:

```c++
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wunused-whatever
const double cold_threshold_;
#pragma clang diagnostic pop
```

keith: > Add __attribute__((used)) to this field. I haven't tried this, not sure if it works in all places, specifically member fields.

Turns out it does not

> #pragma clang diagnostic push

Are these ok when this is being compiled with gcc?

acozzette:
Actually I just realized someone within Google is working on a fix for this problem so we might want to just wait for that fix. CC @haberman who might have more context on that.

src/google/protobuf/compiler/cpp/cpp_message.cc
kastiglione: `#pragma GCC diagnostic ...` would also have to be added to support gcc.

keith:
I updated this with the fake use suggestion in the meantime

acozzette:
OK, if that simple solution works then let's go with that. I will just wait for the CI tests to finish and then merge this.

keith:
Thanks!

Issue #6015, opened 2019-04-10
Enable -Werror when building protobuf itself


keith:
**What language does this apply to?**

None

**Describe the problem you are trying to solve.**

Currently building protobuf itself for bazel related tools can produce some warnings like the ones fixed here:

- https://github.com/protocolbuffers/protobuf/pull/6011
- https://github.com/protocolbuffers/protobuf/pull/6014

This produces noisy output for users, and makes it impossible to enable `--host_copt=-Werror` for other tools.

**Describe the solution you'd like**

To make sure new warnings aren't introduced, it would be great to enable `-Werror` on protobuf's CI.

**Describe alternatives you've considered**

Alternatively we can force all users to disable all warnings produced by protobuf (which also will work if https://github.com/protocolbuffers/protobuf/pull/6013 is merged) but that doesn't seem ideal.

**Additional context**

There are some other issues about using `-Werror` with proto generated code, but those are unrelated since this is about when building protobuf itself.

acozzette:
This is probably doable if we can find a good solution for #6014.

keith:
@acozzette is it feasible for me to make the changes for the CI setup for this, or is this something a Googler has to do?

acozzette:
@keith I think it might be easiest to just edit the BUILD file and then there's no need to worry about the CI config.

keith:
One worry I had about that is that it would apply to all consumers too, which could end up creating difficult issues to workaround like the ones caused by https://github.com/protocolbuffers/protobuf/pull/6013

acozzette:
Ah OK, I see. I guess if consumers pass stricter warning flags then those would turn into errors.

I think updating the CI should just be a matter of adding additional Bazel flags in this file: https://github.com/protocolbuffers/protobuf/blob/master/kokoro/linux/bazel/build.sh

keith:
I've submitted https://github.com/protocolbuffers/protobuf/pull/6022 for this

keith:
Here's another PR with some more warning fixes https://github.com/protocolbuffers/protobuf/pull/6025

keith:
Thanks for all the help @acozzette !

PR #6022, opened 2019-04-11
Add -Werror for CI builds

keith:
Fixes https://github.com/protocolbuffers/protobuf/issues/6015

### DIFF
--- a/kokoro/linux/bazel/build.sh
+++ b/kokoro/linux/bazel/build.sh
@@ -7,7 +7,7 @@ set -ex
 cd $(dirname $0)/../../..
 
 git submodule update --init --recursive
-bazel test :protobuf_test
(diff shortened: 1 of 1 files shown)


keith:
This will fail on CI until https://github.com/protocolbuffers/protobuf/pull/6011 and https://github.com/protocolbuffers/protobuf/pull/6014 are fixed

acozzette:
I kicked off the CI builds just now so we should see a failure in the Bazel tests.

keith:
I've updated this to not add `-Werror` for examples because there are a lot of warnings in there that we don't control

keith:
This also now depends on https://github.com/protocolbuffers/protobuf/pull/6025

keith:
@acozzette I _think_ this should be green now

PR #6025, opened 2019-04-12
Fix more warnings

keith:
These were found attempting to enable -Werror

### DIFF
--- a/src/google/protobuf/arena_unittest.cc
+++ b/src/google/protobuf/arena_unittest.cc
@@ -1345,10 +1345,8 @@ TEST(ArenaTest, GetArenaShouldReturnNullForNonArenaCompatibleTypes) {
 
   // Test that GetArena returns nullptr for types that have a GetArena alias.
   struct {
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-local-typedef"
     using GetArena = Arena*;
-#pragma GCC diagnostic pop
+    GetArena unused;
   } has_get_arena_alias;
   EXPECT_EQ(nullptr, Arena::GetArena(&has_get_arena_alias));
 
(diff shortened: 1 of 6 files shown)


src/google/protobuf/arena_unittest.cc
```diff
@@ -1345,7 +1345,10 @@ TEST(ArenaTest, GetArenaShouldReturnNullForNonArenaCompatibleTypes) {
 
   // Test that GetArena returns nullptr for types that have a GetArena alias.
   struct {
+#pragma GCC diagnostic push
```
acozzette: I'm a little reluctant to use these kinds of pragmas because I'm not sure how portable they are. Could we fix this one by doing something like `(void) has_get_arena_alias::GetArena`?

kastiglione: Which compilers does protobuf support?

logankaser: I've been compiling with clang.
It supports the same pragma but with `s/GCC/clang/g`.
I'm personally in favor of the `(void)var;` solution. I've been using that for years to silence unused warnings and works on at least gcc and clang.

kastiglione: Yep, should have mentioned GCC was used here to cover both gcc and clang.

keith: It looks like there's a decent amount of this style disable elsewhere in the codebase, but I'm also happy to change it to the other option. What do you think @acozzette ?

acozzette: Ideally we should be able to support any compiler that can handle C++11, but I guess the most common ones are Clang, GCC, and MSVC. I would say it's best if we can use the `(void) var;` style but if that doesn't work for some reason then we can fall back on the pragmas.

keith: @acozzette updated!

acozzette:
Thanks, @keith!

PR #6034, opened 2019-04-12
comply with Bazel recommendations

nicmcd:
This brings the BUILD file into compliance with Bazel recommendations to not use bind (see [here](https://docs.bazel.build/versions/master/be/workspace.html#bind)). 

My Bazel recommendation compliant code wouldn't work with protobuf because it forced me to use bind() to remap "@zlib//:zlib" to "//external:zlib".

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -19,7 +19,7 @@ config_setting(
 # ZLIB configuration
 ################################################################################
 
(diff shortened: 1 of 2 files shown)


acozzette:
Thanks @nicmcd!

PR #6039, opened 2019-04-15
Make use of RDTSC optional as it may not be available

letmaik:
`RDTSC` is unavailable when using [seccomp](https://en.wikipedia.org/wiki/Seccomp) while setting  `PR_SET_TSC` to `PR_TSC_SIGSEGV`. There may be other odd cases as well.

Since the value of `RDTSC` is only used conditionally anyway (`#if defined(__x86_64__) && defined(__GNUC__)`) I suggest simply removing the relevant code. It's used for making a random seed more random, but I doubt this is really necessary.

EDIT: See discussion below. `RDTSC` is disabled optionally by macro.

### DIFF
--- a/src/google/protobuf/map.h
+++ b/src/google/protobuf/map.h
@@ -937,7 +937,7 @@ class Map {
     // Return a randomish value.
     size_type Seed() const {
       size_type s = static_cast<size_type>(reinterpret_cast<uintptr_t>(this));
-#if defined(__x86_64__) && defined(__GNUC__)
+#if defined(__x86_64__) && defined(__GNUC__) && !defined(GOOGLE_PROTOBUF_NO_RDTSC)
       uint32 hi, lo;
       asm("rdtsc" : "=a"(lo), "=d"(hi));
       s += ((static_cast<uint64>(hi) << 32) | lo);


letmaik:
CLA signed.

acozzette:
@letmaik Sorry for taking so long to get back to this. The one thing I am worried about is that part of the motivation for using the RDTSC is to prevent hash flooding attacks, so I am not sure if we will have a sufficient amount of randomness without it. Maybe we could provide a way to optionally disable it with a macro but still keep it on by default?

letmaik:
As an alternative, how about instead trying to use the `rdrand` instruction of Intel CPUs if available and fall-back to `rdtsc` otherwise. This would be enough in my scenario as I'm not targeting old or non-Intel CPUs. Example code (needs `-mrdrnd` on gcc/clang to access `_rdrand64_step` intrinsic):
```c++
#include <iostream>

#include <cpuid.h>
#include <immintrin.h>

#define RDRAND_MASK 0x40000000

bool is_rdrand_supported() {
    static int supported = -1;
    if (supported == -1) {
        unsigned int eax, ebx, ecx, edx;
        __get_cpuid(1, &eax, &ebx, &ecx, &edx);
        supported = (ecx & RDRAND_MASK) == RDRAND_MASK;
    }
    return supported;
}

bool rdrand_64(unsigned long long* x) {
    int i = 10;
    while (i-- > 0) {
        if (_rdrand64_step(x)) {
            return true;
        }
    }
    return false;
}

int main (void) {
    if (is_rdrand_supported()) {
        unsigned long long x;
        if (rdrand_64(&x))
            std::cout << x << std::endl;
        else
            std::cout << "error calling rdrand" << std::endl;
    } else {
        std::cout << "rdrand not supported" << std::endl;
    }
}
```

If that's too much added complexity then I'm fine with your proposed solution as well (keep `rdtsc` enabled by default, disable on request, I guess via a CMake option).

acozzette:
@letmaik The `rdrand` idea is tempting but I think we should instead provide a way to disable the use of `rdtsc` just to keep things simple.

letmaik:
@acozzette Alright, that's perfectly fine with me. Are you guys OK with making the change yourselves? I feel if I start messing with the CMake scripts and preprocessor macros then I will likely not put things at the right place, or choose bad names. Do let me know if you prefer me to do it, my goal is to get it in for the next release.

acozzette:
@letmaik I think if you want to get it in soon then your best bet is to write the change yourself. Maybe a good macro name would be `GOOGLE_PROTOBUF_NO_RDTSC`. I am not very familiar with CMake, so I'm not sure of the best way to set that macro from CMake.

letmaik:
@acozzette I made the change, it's good to go from my side. I thought about CMake again but came to the conclusion that there's no need to expose it at that level. There are other macro definitions which are handled the same, e.g. `GOOGLE_PROTOBUF_NO_THREADLOCAL`. I will consume protobuf directly from a CMake build via `add_subdirectory` and can inject any preprocessor definitions from the outside, so that's not an issue.

acozzette:
Thanks @letmaik!

PR #6044, opened 2019-04-16
Replaced all instances of Simple{IDF}toa with StrCat.

haberman:
This makes for easier rewrites between internal and external.


haberman:
Ack, I'm sorry, we should roll this back. Since writing this I learned that StrCat() (at least the one inside Google) truncates floats and doubles. It's only safe as a replacement for `SimpleItoa()`, not SimpleDtoa()` or `SimpleFtoa()`.

haon4:
ah. Would you like me to prepare a rollback or do you already have a rollback PR?

PR #6046, opened 2019-04-17
Change "python" to "python_headers" in comment

sgreenstein:
Following the instructions as is does not work, the new_local_repository rule needs to be called "python_headers"

### DIFF
--- a/util/python/BUILD
+++ b/util/python/BUILD
@@ -9,7 +9,7 @@
 # includes the following local_repository rule that points to this directory:
 #
 # new_local_repository(
-#   name = "python",
+#   name = "python_headers",
 #   path = __workspace_dir__ + "/util/python",
 # )
 cc_library(


anandolee:
Does it produce any errors if use python?

sgreenstein:
In the comment it says "Projects that include protobuf using a Bazel external repository will need to add a workspace rule to their WORKSPACE files to add an external workspace that includes the Python headers. For example, the protobuf WORKSPACE file includes the following local_repository rule that points to this directory:"

First, the protobuf WORKSPACE file does not contain that local_repository rule, so the final sentence of that comment should probably be updated too. Second, if I put the rule in my WORKSPACE as is, there is this error:

com_google_protobuf/BUILD:685:1: no such target '//external:python_headers': target 'python_headers' not declared in package 'external' defined by WORKSPACE and referenced by '@com_google_protobuf//:python/google/protobuf/internal/_api_implementation.so'

It needs to be python_headers, as evidenced by this line: https://bazel.googlesource.com/bazel/+/master/WORKSPACE#9

PR #6048, opened 2019-04-17
Fix javalite's protobuf.bzl to be compatible with new Bazel releases

jin:
This allows the `javalite` branch to be used with latest Bazel versions.

Ref: https://github.com/googlesamples/android-testing/issues/239

cc @laurentlb

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -127,7 +127,7 @@ proto_gen = rule(
         "protoc": attr.label(
             cfg = "host",
             executable = True,
-            single_file = True,
+            allow_single_file = True,
             mandatory = True,
         ),
         "plugin": attr.label(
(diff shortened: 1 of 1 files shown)


haon4:
Thanks. We are also trying to release the new version of javalite in the main branch. That version will probably go out in the next Protobuf release. We will probably need to make some changes in bazel again then.

jin:
This will require a new javalite version, yes. When is the next protobuf release expected? The newest Bazel versions need these changes, and I don't think there are any more backward incompatible fixes to make.

haon4:
We are planning to release the next version of Protobuf with the new javalite version in the next few weeks. I am hoping to get something out in the first two weeks of May. Would it be possible to wait until then? We would like to completely freeze this old javalite branch if possible.

jin:
I see, if it's a few weeks then I think we can wait for that. I believe the changes in this PR is already in the main branch, so we can close this?

haon4:
SG. Thank you!

haon4:
Hi jin, 3.8.0-rc1 for javalite is released. Can you give that a try?
https://search.maven.org/artifact/com.google.protobuf/protobuf-javalite/3.8.0-rc-1/bundle

jin:
We don't depend on the artifact, but the protobuf source tree. I don't see a Javalite release in https://github.com/protocolbuffers/protobuf/releases?

haon4:
protoc now supports javalite in it. You just need to prepend lite: to the --java_out flag. See https://github.com/protocolbuffers/protobuf/blob/master/java/lite.md. Can you give that a try?

PR #6055, opened 2019-04-20
Replaced unnecessary include with struct definition

parnic:
It appears this file was including winsock2.h just to get the timeval struct definition. There are platforms, such as Xbox One, that don't use winsock but do compile with _MSC_VER set, so just drop in the struct definition we needed instead of including all of winsock. This fixes compilation of this file for Xbox One. Without this change, there were numerous `timeval` complaints such as
```
1>c:\dev\protobuf\src\google\protobuf\util\time_util.h(153): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int (compiling source file C:\dev\protobuf\src\google\protobuf\util\time_util.cc)
1>c:\dev\protobuf\src\google\protobuf\util\time_util.h(153): error C2143: syntax error: missing ',' before '&' (compiling source file C:\dev\protobuf\src\google\protobuf\util\time_util.cc)
```
etc.

### DIFF
--- a/src/google/protobuf/util/time_util.h
+++ b/src/google/protobuf/util/time_util.h
@@ -37,7 +37,14 @@
 #include <ostream>
 #include <string>
 #ifdef _MSC_VER
+#ifdef _XBOX_ONE
+struct timeval {
+	long    tv_sec;         /* seconds */
+	long    tv_usec;        /* and microseconds */
+};
+#else
 #include <winsock2.h>
+#endif // _XBOX_ONE
 #else
 #include <sys/time.h>
 #endif
--- a/src/google/protobuf/io/io_win32.cc
+++ b/src/google/protobuf/io/io_win32.cc
@@ -42,7 +42,7 @@
 //
 // This file is only used on Windows, it's empty on other platforms.
 
-#if defined(_WIN32)
+#if defined(_WIN32) && !defined(_XBOX_ONE)
 
 // Comment this out to fall back to using the ANSI versions (open, mkdir, ...)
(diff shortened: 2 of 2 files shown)


src/google/protobuf/util/time_util.h
```diff
@@ -37,7 +37,10 @@
 #include <ostream>
 #include <string>
 #ifdef _MSC_VER
-#include <winsock2.h>
+struct timeval {
```
acozzette: The thing that makes me hesitate about this is that if someone happens to include this header and also winsock2.h then there will be an error about the struct being redefined. Is there a way we could restrict this change so that it applies to XBox One only?

parnic: That's certainly a good point of concern.

Yes, we could guard it with `#ifdef _XBOX_ONE`, though I was hesitant to do that because there are other problems that need to be addressed to get pb compiling for Xbox One (such as all of io_win32.cc needing to be skipped, some filesystem calls like `access()` and friends not being supported, etc.). It felt like adding that include guard would send a signal that the platform was supported which isn't true even after this change.

Perhaps a `GOOGLE_PROTOBUF_` define could be added to control this instead, though that may also open a can of worms.

acozzette: Guarding it with `_XBOX_ONE` sounds good to me.

parnic: Replaced this commit with an `_XBOX_ONE` guard and added a second commit necessary to get the library compiling for Xbox One.

PR #6059, opened 2019-04-22
fix deep merge for message collection

drashnane:
RepeatedField<Message>.MergeFrom and MapField<Message>.MergeFrom should be deep merge, not just copy references


drashnane:
I signed it!

PR #6061, opened 2019-04-22
fix csharp : repeated messages should be deep merge, not just copy references

drashnane:
fix csharp : repeated messages should be deep merge, not just copy references


csharp/src/Google.Protobuf/Collections/CollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace Google.Protobuf.Collections
+{
+
+    /// <summary>
+    /// Extension Methods for Collections
+    /// </summary>
+    public static class CollectionExtensions
```
anandolee: Extension may be misunderstand as we are going to add proto2 support which has extensions. We will also add ExtensionCollection similar with FieldCollection.

ObsidianMinor:
Do other language implementations perform deep copies of elements vs reference copies?

anandolee:
Yes, python is deep copy:
https://github.com/protocolbuffers/protobuf/blob/bf32b36ab8aa9e56c031e00c0553db640d3627d4/python/google/protobuf/internal/containers.py#L411

C++:
https://github.com/protocolbuffers/protobuf/blob/5b232b8ecbce13286be09e703997e887ae0d464d/src/google/protobuf/repeated_field.h#L1663

csharp/src/Google.Protobuf/Collections/CollectionExtensions.cs
anandolee: And please add test for this PR

ObsidianMinor:
In that case, it'd be better to add `MergeFrom` methods to both `RepeatedField<T>` and `MapField<TKey, TValue>`, as well as use `IDeepCloneable<T>` which will do the cloning for us.

It wouldn't be good to change the behavior of `Add` like @anandolee [says](https://github.com/protocolbuffers/protobuf/pull/6061#discussion_r280532399) because adding owned messages normally would result in a *deep* clone for no reason.

```cs
public sealed class RepeatedField<T>
{
    public void MergeFrom(RepeatedField<T> other)
    {
        ProtoPreconditions.CheckNotNull(other, nameof(other));
        EnsureSize(count + other.count);
        for (int i = 0, j = count; i < other.count; i++, j++)
        {
            var element = other.array[i];
            IDeepCloneable<T> cloneable;
            if ((cloneable = element as IDeepCloneable<T>) != null)
            {
                array[j] = cloneable.Clone();
            }
            else
            {
                array[j] = element;
            }
        }
        count += other.count;
    }
}
```

This way we can ensure the size ahead of time and clone without having to use `MessageParser<T>`.

On the other end `MapField<TKey, TValue>` doesn't get the same size check optimzations, but can still use `IDeepCloneable<T>`

```cs
public sealed class MapField<T>
{
    public void MergeFrom(MapField<TKey, TValue> entries)
    {
        ProtoPreconditions.CheckNotNull(entries, nameof(entries));
        foreach (var pair in entries)
        {
            IDeepCloneable<TValue> cloneable;
            if ((cloneable = pair.Value as IDeepCloneable<TValue>) != null)
            {
                Add(pair.Key, cloneable.Clone());
            }
            else
            {
                Add(pair.Key, pair.Value);
            }
        }
    }
}
```

drashnane:
Hi, I've remove CollectionExtensions

drashnane:
Hi, I remove CollectionCollections and add tests for deep copy

csharp/src/Google.Protobuf/Reflection/Descriptor.cs
```diff
@@ -705,10 +705,10 @@ internal sealed partial class FileDescriptorProto : pb::IMessage<FileDescriptorP
       dependency_.Add(other.dependency_);
```
anandolee: Although there's no behavior difference, I sill prefer to change repeated_scalar_fields and repeated_enum_fields to call MergeFrom() as well

ObsidianMinor: A normal `Add` would probably be faster with scalar fields (since it'd just be an array copy) however that's an optimization that can be built into `MergeFrom`.

drashnane: I've fixed it. Only message type && not wrapper type use MergeFrom

jskeet:
Closing as this would be a very significant change in behavior, quite possibly breaking a lot of users.

PR #6068, opened 2019-04-23
Update version to 3.8.0

haon4:


BSBandme:
It seems java lite's version is not updated

haon4:
> It seems java lite's version is not updated

Fixed.

haon4:
Also send out https://github.com/protocolbuffers/protobuf/pull/6069 to fix the update_version script.

haon4:
Tests are green now. PTAL.

Issue #3925, opened 2017-11-20
Ruby Segmentation Fault


blowmage:
This morning the google-cloud-ruby project had an [issue opened for a segfault](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1843) using the most recent 3.5.0 gem. It appears the segfault is coming from google-protobuf's `msgclass` function.

System details:
 
* gRPC language and version: grpc 1.7.2 (x86_64-linux)
* Protobuf version: google-protobuf 3.5.0 (x86_64-linux)
* Operating System: Alpine Linux 3.6.2
* Runtime version: ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]

<details>
<summary>Stack trace:</summary>

```
/usr/local/bundle/gems/google-protobuf-3.5.0-x86_64-linux/lib/google/protobuf/any_pb.rb:15: [BUG] Segmentation fault at 0x00000000000170f6
ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0133 p:---- s:0780 e:000779 CFUNC  :msgclass
c:0132 p:0026 s:0776 e:000775 CLASS  /usr/local/bundle/gems/google-protobuf-3.5.0-x86_64-linux/lib/google/protobuf/any_pb.rb:15
c:0131 p:0011 s:0773 e:000772 CLASS  /usr/local/bundle/gems/google-protobuf-3.5.0-x86_64-linux/lib/google/protobuf/any_pb.rb:14
c:0130 p:0039 s:0770 E:000180 TOP    /usr/local/bundle/gems/google-protobuf-3.5.0-x86_64-linux/lib/google/protobuf/any_pb.rb:13 [FINISH]
c:0129 p:---- s:0767 e:000766 CFUNC  :require
c:0128 p:0012 s:0762 e:000761 BLOCK  /usr/local/bundle/gems/activesupport-5.1.1/lib/active_support/dependencies.rb:292
c:0127 p:0057 s:0759 e:000758 METHOD /usr/local/bundle/gems/activesupport-5.1.1/lib/active_support/dependencies.rb:258
c:0126 p:0017 s:0753 e:000752 METHOD /usr/local/bundle/gems/activesupport-5.1.1/lib/active_support/dependencies.rb:292
[...]

PR #6073, opened 2019-04-25
Register mark objects so Protobuf is compaction friendly

tenderlove:
This commit removes an unused reference and registers globals with the
GC so that they will never die.  Ruby is getting a compacting GC, and it
means that these references can move.  Registering them with
`rb_gc_register_mark_object` will ensure the constants don't move and
will not be collected.

### DIFF
--- a/ruby/ext/google/protobuf_c/protobuf.c
+++ b/ruby/ext/google/protobuf_c/protobuf.c
@@ -37,7 +37,7 @@
 #include "message.h"
 #include "repeated_field.h"
 
-VALUE cError;
+VALUE cParseError;
 VALUE cTypeError;
 
 const upb_fielddef* map_field_key(const upb_fielddef* field) {
@@ -368,8 +368,10 @@ void Init_protobuf_c() {
   Map_register(protobuf);
   Message_register(protobuf);
 
-  cError = rb_const_get(protobuf, rb_intern("Error"));
+  cParseError = rb_const_get(protobuf, rb_intern("ParseError"));
+  rb_gc_register_mark_object(cParseError);
   cTypeError = rb_const_get(protobuf, rb_intern("TypeError"));
+  rb_gc_register_mark_object(cTypeError);
 
   rb_define_singleton_method(protobuf, "discard_unknown",
                              Google_Protobuf_discard_unknown, 1);


ruby/ext/google/protobuf_c/protobuf.c
```diff
@@ -103,9 +102,10 @@ void Init_protobuf_c() {
   RepeatedField_register(protobuf);
   Map_register(protobuf);
 
-  cError = rb_const_get(protobuf, rb_intern("Error"));
   cParseError = rb_const_get(protobuf, rb_intern("ParseError"));
+  rb_gc_register_mark_object(cParseError);
```
TeBoring: https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/defs.c#L129
I think you should register the variable before assigning value to it.

tenderlove: @TeBoring I don't think it really matters.  This will also work. (BTW I wrote Ruby's GC Compactor so I'm familiar with GC internals)

TeBoring: Is it possible that cParseError got gc'ed before calling rb_gc_register_mark since cParseError is a global variable and has not been marked yet?
We had https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d

TeBoring: Any thought?

tenderlove: I don't think it's possible.  Something must be maintaining a reference to the `ParseError` constant, otherwise `rb_const_get(protobuf, rb_intern("ParseError"));` would return `Qnil` or something.  The only reason I added these to `rb_gc_register_mark_object` is to prevent movement during compaction.  It has nothing to do with liveness.

Looking at https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d, I'm not sure how it would fix anything.  You said you were able to reproduce this: https://github.com/protocolbuffers/protobuf/issues/3925  Do you have a c-level stack trace?  Possibly with debugging symbols enabled?

TeBoring: https://github.com/protocolbuffers/protobuf/blob/master/ruby/tests/gc_test.rb
The gc test was added to reproduce the bug.

haberman: @tenderlove It appears that `rb_gc_register_address()` itself calls `ALLOC()` and could therefore trigger a collection before the address is actually registered? https://github.com/ruby/ruby/blob/ca435ed04ae15607c598808473c59dd7c83debc1/gc.c#L6749

I am guessing that was the root cause of https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d?

That implies that the following code sequence is not safe:

```c
VALUE global;

void f() {
  global = create_new_object();
  // This could GC the new object before the address is registered.
  rb_gc_register_address(&global);
}
```

Which is why we switched to using this instead:

```c
VALUE global = Qnil;

void f() {
  rb_gc_register_address(&global);
  global = create_new_object();
}
```

But here you make the point that we do not have a *unique* reference to `cParseError`, since the object was just returned from `rb_const_get()`. So we don't need to worry here about whether we are properly gc-rooting this object. While that is true, it also seems error-prone to write different code depending on whether or not we have a unique reference to the object.

Would using `rb_gc_register_address()` as we do elsewhere in the code solve the problem of GC-compacting?

tenderlove: > I am guessing that was the root cause of [cf7c15e](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d)?

It's hard for me to say without seeing a stack trace or looking at a core file. cf7c15e31a456f634d4e3deaf4ef74a6bfad825d contains 12 changes (besides the top Qnil one). These 11 changes: [1](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L129), [2](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L302), [3](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L521), [4](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L919), [5](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L1040), [6](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L1205), [7](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L1494), [8](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L1572), [9](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L1648), [10](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-5ea74e72671b17df2672380bef7093b0L828), [11](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-4af94f9c1bdc9d60e1ad59527d5ade5cL629) all use `rb_define_class_under` to define their globals.  That function eventually [adds the constant to a mark array](https://github.com/ruby/ruby/blob/c19d2faa43d947e44743d08e1bd49ca76de89eee/class.c#L751) before returning it. In these cases, the constants are already kept alive so I think these 11 changes are not necessary.

The only remaining change is to [`generated_pool`](https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d#diff-b1a70082252aba3ab3b7fb55253d8b38L132).  As you know, Ruby's GC is a conservative stack scanning GC.  This object *may* be kept alive on the machine stack, but the C compiler could have done some optimizations that prevent the GC from seeing that pointer.

I noticed the original bug was reported as only happing on Ruby 2.4.  This leads me to believe that it could be either a bug in Ruby (were the constants not getting added to the global mark array in Ruby 2.4?), or the compiler optimized the code such that stack scanning didn't catch the "generated_pool" global.

There was no analysis of the underlying issue in the commit message or PR, which is why I asked about a core file or stack trace.

> Would using `rb_gc_register_address()` as we do elsewhere in the code solve the problem of GC-compacting?

Yes, either function will work.  It seems like passing the pointer as `rb_gc_register_mark_object` does, rather than a pointer to a pointer as `rb_gc_register_address` does, would be better for stack scanning.  But either will work fine. 😄 

Note: it's technically not necessary to root the constants, but I think it's good that we are.  I need to find the changelog entry, but I think we (ruby core) consider it technically a bug that a C extension creates a class but doesn't root it.  It's just so common that we made the C interface automatically root them.

haberman: > There was no analysis of the underlying issue in the commit message or PR, which is why I asked about a core file or stack trace.

Yes, it's true the fix should have been better documented. Here is a repro if you want to take a look (originally from @scottserok but I updated the Docker image to be something that matches the correct Ruby version): https://github.com/haberman/protobuf-segfault-example.  Here is a stack trace: https://github.com/haberman/protobuf-segfault-example/blob/master/backtrace.txt

But actually, I just noticed that the repro is using the newest versions of all gems (including protobuf), and yet still crashing, which leads me to believe the bug is not fixed at all. :( We had [one report that the bug was no longer occurring](https://github.com/protocolbuffers/protobuf/issues/3925#issuecomment-397014436) but then [two](https://github.com/protocolbuffers/protobuf/issues/3925#issuecomment-418206393) [more](https://github.com/protocolbuffers/protobuf/issues/3925#issuecomment-396853548) reports that the crash was still happening. The people who experienced the crash were requested to open new issues but it appears they didn't, and I'm not sure the bug was ever investigated further after that.

The crash is in our code. It seems like it might be attempting to operate on a data structure that was collected. Do you know if there is an easy way with the Gem toolchain to substitute a debug version of our library into the repro?

@TeBoring Do you remember the details of your investigation into this bug, and how you concluded that https://github.com/protocolbuffers/protobuf/commit/cf7c15e31a456f634d4e3deaf4ef74a6bfad825d was the correct fix?

TeBoring: We have gc_test. Without the fix, it was broken, and with the fix, the test could pass.
I remember the cause was that global variables will be gc'ed at any time unless they are registered. If some memory is assigned to a global variable before it's registered, the assigned memory could be gc'ed.

Do you have github issue ID that you saw the bug happened again after the fix? Probably due to different cause?

jhawthorn: > Is it possible that cParseError got gc'ed before calling rb_gc_register_mark since cParseError is a global variable and has not been marked yet? We had cf7c15e

I agree with Aaron that this can't be an issue when using `rb_gc_register_mark_address`, since the value we care about marking will be passed to that function we can be sure it will be found by stack scanning (if this didn't work, almost everything would crash 😅)

tenderlove:
Hi, just a friendly bump on this.  We need this so that protobuf will work with Ruby's compacting GC and the C globals don't go bad.  Thanks!

haberman:
Wow, sorry we let this sit so long. We should definitely merge it now. I'll kick off the tests once more.

Issue #1297, opened 2016-03-05
FR: Text proto format could support multi-line strings


riannucci:
In a text proto file like:

```
something: <
  description: "this\nwould\n naturally be a\n multiline string"
  ...
>
```

It would be awesome if we could e.g.:

```
something: <
  description: `
this
would
naturally be a
 multiline string
`
  ...
>
```

This could be used for e.g. attaching markdown documentation to configuration entries which could then be displayed nicely. I'm not 100% on backticks (implies quoting in the parser)... Other things like a bash style `<<EOM .... EOM` demarcation could also work and would be easier to parse.

Thoughts?

gwelymernans:
+1 this idea (whatever syntax it takes). Having the ability to write free-flowing text in the text protobuf is great for people who want to use protobufs but still have it be human readable when you have multiline text.

riannucci:
FWIW, we did actually implement this as a crappy pre-processor over text proto (using the `<<EOM ... EOM` syntax). The tests ("examples") are here: https://github.com/luci/luci-go/blob/master/common/proto/multiline_test.go

It also will do a python-style `textwrap.dedent` on it, so you can do:

```proto
something: <
  description: <<EOM
    this
    would
    naturally be a
      multiline string
  EOM
  ...
>
```

Which would be the equivalent of:

```proto
something: <
  description: "this\nwould\nnaturally be a\n  multiline string"
  ...
>
```

(note the _beautiful_ indentation and lack of weird prefix spaces in the translated version :wink:)

gwelymernans:
Hmm...it does appear to be [an option](https://github.com/google/protobuf/blob/d5fb408ddc281ffcadeb08699e65bb694656d0bd/src/google/protobuf/io/tokenizer.h#L239), but it's discouraged. I don't know the reasoning for it.

aselle:
This would be a great feature, we're looking at needing to do large blocks of text in a proto for api documentation, and it would be great to do this! In lieu of this we'll probably make a preprocessor as @riannucci is using, but that is much less elegant. Thanks!

laike9m:
+1 for this.

DonGar:
+1 again.

maguro:
I'm going to take a stab at this.  I'm going to use backticks.  Not sure when, or if, I'll get this done.  Feel free to ping me.

xfxyjwf:
Multi-line string value is already supported as far as I can tell, see:
[...]

PR #6078, opened 2019-04-28
Add support for support multiline strings

maguro:
Support for support multiline strings is provided in the following two
ways

* backtick delimited string literals - for "lite" multiline strings
* triple double quotes - for strings that might include a backtick,
  e.g. Markdown documentation

Strings are now categorized in two manors, each with their own lexical
token

* string names - strings that cannot be multiline, e.g. names
* string values - strings values, e.g. default and custom values

NOTE: There may be some tooling that has layered additional semantics
on top of the Protobuf spec that assumes strings are single line, e.g.
package paths.  Since these are layered on top of the Protobuf spec,
there is no way that the compiler can protect them.  This, however,
should not cause undue concern since providing multiline values for
such elements is no different than misspelling a path.


maguro:
This is for [FR: Text proto format could support multi-line strings](https://github.com/protocolbuffers/protobuf/issues/1297).

acozzette:
@maguro Could you sign the CLA at https://cla.developers.google.com/?

jvolkman:
I don't fully understand the split between names and values. Strings can already erroneously contain newlines today. E.g. `syntax = "proto\n3";` would be just as invalid as
```
syntax = """proto
3"""
```

And if the concern is about other tooling that parses aspects of .proto files, strings can already span multiple lines today. E.g., the following is valid today:

```
import "my/really/long/path/to/"
       "a/proto/file.proto"; 
```

Seems like the change would be a lot simpler without the addition of a new token type.

maguro:
re: CLA - it's slogging through our legal workflow... :(

jvolkman:
> Given the above example, the compiler will complain when there is a newline in the syntax value since it cannot be a multiline string.

Sure, but the compiler would complain even without the new token type because "proto\n3" is not a valid syntax level. My point is that it's already possible today to specify values that contain newlines at any part of the input that expects a string. If, semantically, a value must not contain a newline, then the compiler or some other part of the toolchain should already enforce that.

> The compiler will concatenate that into a single string with no newlines; not sure how that example is relevant. Multiline strings preserve the newlines.

I wasn't sure whether your intent was to protect other tooling that parses proto files. E.g., `egrep '^import' myproto.proto` would not handle concatenated strings spanning multiple lines today. It doesn't sound like this is what you're trying to address.

maguro:
I signed it!

acozzette:
@maguro It seems that something is still preventing the CLA bot from verifying your CLA. If you're using a corporate CLA signed by your company, could you make sure the author field on your Git commit contains your company email address?

maguro:
I'll fix the author field.

jierongjian:
you counld see whilch email was used before push

maguro:
Not sure what the the **Mergeable** issue is.

acozzette:
CC @jhump @pedgeio @dsnet who I think are all interested in the specification of the .proto file format

jhump:
The triple-quote form, combined with existing support for concatenating adjacent string literals, seem like an error-prone combination. For example, this is valid source today:
```
syntax = """proto3";
```
Is it valid with this PR? If not, that's a backwards compatibility issue.

If it is, then it feels like a sharp-edge that pasting text that contains a quote into a triple-quoted block could lead to unexpected error messages. A single quote in the block necessarily terminates the whole string and changes the interpretation from "multi-line string" to "empty string concatenated with non-empty string concatenated", which means any newlines in the second string are actually syntax errors (as will be anything after the quote).

Seems like this would be way cleaner and simpler to implement correctly if you only allow backticks as the indicator of multi-line string literals. Also, if you support backticks, does that mean you also need to add support for ```'\`'``` as an escape sequence, so you can include a backtick character inside of a backtick-quoted block?

pedgeio:
I second @jhump's view that this feels like it will be error-prone, likely have backwards-compatibility issues, and also likely result in issues in a myriad of secondary Protobuf libraries in the OSS community.

jhump:
Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the _text format_ for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.

src/google/protobuf/io/tokenizer.cc
```diff
@@ -413,8 +419,11 @@ void Tokenizer::ConsumeString(char delimiter) {
       }
 
       default: {
-        if (current_char_ == delimiter) {
-          NextChar();
+        if (delimiter == 0) {
+          if (TryConsume('"') && TryConsume('"') && TryConsume('"')) {
```
jhump: This seems wrong. If there is only one quote, that should also terminate a triple-quoted block and cause its contents to be reinterpreted as a regular quoted literal (since the first three quotes means it was an empty string concatenated with a normal string). So there should be some different action taken if the first `TryConsume` succeeds but the second fails.

Similarly, if there are two quotes (instead of three), that should be interpreted as ending regular quoted literal and then starting a new regular quoted literal (whose value will be concatenated to the previous one).

But this code just falls through if it only consumes one or two quotes, which means it calls `NextChar()` on line 429. I think that means it eats/loses a character of input.

maguro:
> Seems like this would be way cleaner and simpler to implement correctly if you only allow backticks as the indicator of multi-line string literals. 

I agree.

> Also, if you support backticks, does that mean you also need to add support for `` '\`' `` as an escape sequence, so you can include a backtick character inside of a backtick-quoted block?

Yes, it's in the PR.  I'll some additional testing for it specifically.

maguro:
> I second @jhump's view that this feels like it will be error-prone, likely have backwards-compatibility issues, and also likely result in issues in a myriad of secondary Protobuf libraries in the OSS community.

May I assume that you are specifically speaking of the triple-double-quote?

maguro:
> Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the _text format_ for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.

IIUC, the PR is merely a syntactic shortcut for multi-line strings which are supported in the compiler today.  I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.

src/google/protobuf/io/tokenizer.cc
maguro: You are correct and, iiuc, alluded to this in an earlier comment.  Limiting multi-line strings to backticks should solve this.

pedgeio:
> > I second @jhump's view that this feels like it will be error-prone, likely have backwards-compatibility issues, and also likely result in issues in a myriad of secondary Protobuf libraries in the OSS community.
> 
> @pedgeio, may I assume that you are specifically speaking of the triple-double-quote?

Yes, and really adding this support in general - is there really a need to add this to the Protocol Buffers spec? I think adding support for multiline strings with any new construct is going to have unintended consequences, I'm not sure if any gain here is worth the potential pitfalls.

pedgeio:
> > Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the _text format_ for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.
> 
> IIUC, the PR is merely a syntactic shortcut for multi-line strings which are supported in the compiler today. I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.

What is the need for multi-line strings specifically? I am speaking just as one user, so it's just a single point of data, but I've been using Protobuf for many years and have never found myself wanting this functionality (and I haven't been made aware of anyone else who wants it either) - obviously you do though :-) so I was hoping for a rationale for making this major change that could break a lot of downstream libraries (potentially).

jhump:
> > Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the text format for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.
>
> IIUC, the PR is merely a syntactic shortcut for multi-line strings which are supported in the compiler today. I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.

What I'm saying is that you are not _just_ touching the compiler though. You are also touching the protobuf text format (in file `src/google/protobuf/text_format.cc`) -- the compiler re-uses that code to process aggregate literals (e.g. custom options where the option type is a message). And changing the protobuf text format has consequences for _all_ of the other runtimes, which must also implement the text format. In other words, to do this properly, you must change at least four other implementations of the text format -- Java, native JS (for browser), Go, and Dart -- just to roll this out to the _officially supported_ languages. (I believe the official languages other than those link in and rely on this implementation in libproto.)

This will also have knock-on effects for community tools that parse protos (such as `clang`, which includes a proto formatter, Square's Wire library, various Go parser libraries including my own protoreflect) as well as community implementations that add support for other languages and must also implement the text format.

As @pedgeio said, I think changes to the syntax need to be _very_ seriously considered. And in this case, since it also touches the protobuf text format (whose impact extends much wider than just the compiler), there should be a very high bar to justify this sort of change.

jvolkman:
As far as I can tell, the backtick-only variant would be backwards-compatible with any existing, valid proto or textproto.

> I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.

As I've attempted to state before, I don't think this guarding is worth the extra complexity. Limit the change to the lexer and the parser shouldn't even care. The change would be pretty small in Google's own IntelliJ editor: https://github.com/jvolkman/intellij-protocol-buffer-editor/commit/53cca3466f5a2f6015714e613d9f5ceb4d09a915

maguro:
> > > Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the text format for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.
> > 
> > 
> > IIUC, the PR is merely a syntactic shortcut for multi-line strings which are supported in the compiler today. I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.
> 
> What I'm saying is that you are not _just_ touching the compiler though. You are also touching the protobuf text format (in file `src/google/protobuf/text_format.cc`) -- the compiler re-uses that code to process aggregate literals (e.g. custom options where the option type is a message). And changing the protobuf text format has consequences for _all_ of the other runtimes, which must also implement the text format. In other words, to do this properly, you must change at least four other implementations of the text format -- Java, native JS (for browser), Go, and Dart -- just to roll this out to the _officially supported_ languages. (I believe the official languages other than those link in and rely on this implementation in libproto.)
> 
> This will also have knock-on effects for community tools that parse protos (such as `clang`, which includes a proto formatter, Square's Wire library, various Go parser libraries including my own protoreflect) as well as community implementations that add support for other languages and must also implement the text format.
> 
> As @pedgeio said, I think changes to the syntax need to be _very_ seriously considered. And in this case, since it also touches the protobuf text format (whose impact extends much wider than just the compiler), there should be a very high bar to justify this sort of change.

Supporting backticked multiline strings in serialized text-format proto files is not a priority for me and am happy to undo the relevant changes.

pedgeio:
> > > > Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the text format for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.
> > > 
> > > 
> > > IIUC, the PR is merely a syntactic shortcut for multi-line strings which are supported in the compiler today. I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.
> > 
> > 
> > What I'm saying is that you are not _just_ touching the compiler though. You are also touching the protobuf text format (in file `src/google/protobuf/text_format.cc`) -- the compiler re-uses that code to process aggregate literals (e.g. custom options where the option type is a message). And changing the protobuf text format has consequences for _all_ of the other runtimes, which must also implement the text format. In other words, to do this properly, you must change at least four other implementations of the text format -- Java, native JS (for browser), Go, and Dart -- just to roll this out to the _officially supported_ languages. (I believe the official languages other than those link in and rely on this implementation in libproto.)
> > This will also have knock-on effects for community tools that parse protos (such as `clang`, which includes a proto formatter, Square's Wire library, various Go parser libraries including my own protoreflect) as well as community implementations that add support for other languages and must also implement the text format.
> > As @pedgeio said, I think changes to the syntax need to be _very_ seriously considered. And in this case, since it also touches the protobuf text format (whose impact extends much wider than just the compiler), there should be a very high bar to justify this sort of change.
> 
> Supporting backticked multiline strings in serialized text-format proto files is not a priority for me and am happy to undo the relevant changes.

What is the priority here? I asked this above, but just to ask again in case it was missed: can you explain what is the motivation/need for this PR?

jvolkman:
> > > > Also, in order to add support for multi-line strings to the code that parses aggregate option values, you are actually changing the code that handles the text format for the C/C++ (and derived) runtimes. That means it's not congruent with other runtimes that don't use the C/C++ impl, such as Java, Android, Go, and Dart.
> > > 
> > > 
> > > IIUC, the PR is merely a syntactic shortcut for multi-line strings which are supported in the compiler today. I've taken pains to ensure where multi-line strings are not allowed, backtick enclosed strings are not allowed.
> > 
> > 
> > What I'm saying is that you are not _just_ touching the compiler though. You are also touching the protobuf text format (in file `src/google/protobuf/text_format.cc`) -- the compiler re-uses that code to process aggregate literals (e.g. custom options where the option type is a message). And changing the protobuf text format has consequences for _all_ of the other runtimes, which must also implement the text format. In other words, to do this properly, you must change at least four other implementations of the text format -- Java, native JS (for browser), Go, and Dart -- just to roll this out to the _officially supported_ languages. (I believe the official languages other than those link in and rely on this implementation in libproto.)
> > This will also have knock-on effects for community tools that parse protos (such as `clang`, which includes a proto formatter, Square's Wire library, various Go parser libraries including my own protoreflect) as well as community implementations that add support for other languages and must also implement the text format.
> > As @pedgeio said, I think changes to the syntax need to be _very_ seriously considered. And in this case, since it also touches the protobuf text format (whose impact extends much wider than just the compiler), there should be a very high bar to justify this sort of change.
> 
> Supporting backticked multiline strings in serialized text-format proto files is not a priority for me and am happy to undo the relevant changes.

Not applying the change to text format would require more changes, not fewer. You'd need to add a parameter to the shared tokenizer that enables the backtick multiline feature, and only enable it in the protoc usage.

Funnily enough, the parameter already exists: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/io/tokenizer.cc#L210

PR #6117, opened 2019-05-09
Add use_nested to internalAddGeneratedFile to make php and c extension consistent

### LINKED ISSUE #5965: Protobuf 3.7.1 for PHP does not work with non-PECL version
Version: v3.7.1
Language: PHP

Not relevant

gcc version, 3.7.1

**What did you do?**
1. Write any protobuf structure and compile it to PHP with version 3.7.1
2. instantiate the PHP class in a repository that relies on `"google/protobuf": "3.7.1"` through composer/packagist.
3. Get a segfault.

No segmentation fault

**What did you see instead?**
Segmentation fault, caused by incorrect syntax.

This was first introduced in https://github.com/protocolbuffers/protobuf/commit/53a1d49633f1e4396c11d1fe593268793a118fc3. However, when that was added, the separate PHP code was not updated to accept the second argument: https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/DescriptorPool.php#L58-L78

TeBoring:
Fix #5965

### DIFF
--- a/php/src/Google/Protobuf/Internal/DescriptorPool.php
+++ b/php/src/Google/Protobuf/Internal/DescriptorPool.php
@@ -55,7 +55,7 @@ class DescriptorPool
         return self::$pool;
     }
 
-    public function internalAddGeneratedFile($data)
+    public function internalAddGeneratedFile($data, $use_nested = false)
     {
         $files = new FileDescriptorSet();
         $files->mergeFromString($data);


TeBoring:
@dwsupplee Could you help take a look?

TeBoring:
Although I also haven't reproduced it, I suspect some version of php may see this problem.
Seems to me this change has no harm.

haon4:
How does this actually fix anything?

dwsupplee:
I agree there isn't any harm (outside of code readability - since the variable has no use) - but agree with @haon4 it would be surprising if this fixed the reported issue.

TeBoring:
The segment fault may be due to some other issues (not sure how to reproduce).

PR #6121, opened 2019-05-09
Fix update_version script to use the appropriate rc style for C#

haon4:


haon4:
Ack. I think we normally merge all these release branch changes back to master after the release (and before the next release).

PR #6127, opened 2019-05-13
Down integrate to GitHub

haon4:


haon4:
@haberman seems like the recent reflection change may break some unit test. Can you suggest a fix?

PR #6128, opened 2019-05-13
Convert Google::Protobuf.deep_copy to pure Ruby

tenderlove:
In general, I think it will help us to debug issues if we have less C
code and more Ruby code.  This method can be implemented in pure Ruby,
so this commit converts it to pure Ruby.


ruby/ext/google/protobuf_c/message.c
```diff
@@ -512,7 +512,7 @@ VALUE Message_dup(VALUE _self) {
 }
 
 // Internal only; used by Google::Protobuf.deep_copy.
```
haberman: If we are exposing this method publicly we should remove this comment.

In many cases I would prefer to keep internal-only methods private. But in this case I don't see any benefit in hiding `Google::Protobuf::Map::deep_copy` while `Google::Protobuf::deep_copy` is public.

tenderlove: 👍 I'll remove the comments

tenderlove: I changed them to be `:nodoc:`.  We usually `:nodoc:` "publicly visible but private" methods in Rails, and I think it makes sense in this case too.

haberman:
Hmm, this appears to have broken our Ruby Kokoro tests: https://source.cloud.google.com/results/invocations/7f69f595-c945-4d75-92bb-673e2f486878/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fruby23%2Fpresubmit/log

Not sure why they didn't run when the PR was open. We should roll this back until we figure out the fix.

haberman:
Oh, I think I was supposed to add the "kokoro:run" label. Sorry about that.

PR #6153, opened 2019-05-18
Update performance.md

TechProofreader:
I corrected a few grammatical mistakes and increased the document's readability by easing the language flow.


TechProofreader:
I initially submitted this under an account I no longer use, which is why I resubmitted under this account. I'm unsure of what this should be labelled as, so if someone could help out with that, it would be appreciated.

TechProofreader:
Thanks for the help, I appreciate you merging my commit! I did just resubmit another though because I missed one word that was previously misspelled, but I corrected it in the second commit #6158

PR #6155, opened 2019-05-18
php: Fix formatting of Duration

lhecker:
This fixes the following issues for php's native Duration class:
* Missing nanoseconds. The nanoseconds where divided as a float and implicitly converted to a string before being passed to `bcadd` (Most bcmath functions accept their parameters as strings!). This can result in a float formatted in scientific/exponential notation, which bcmath doesn't understand.
  E.g. a Duration instance with 123s and 456ns can previously result in an invocation like
  ```php
  strval(bcadd('123', '4.56e-7', 9)) === '123.000000000'
  ```
  Using a regular `sprintf` seems like a far simpler solution to me.
* Durations are supposed to be formatted without trailing zeroes.
  I've added a simple call to `rtrim` for this.

### DIFF
--- a/conformance/failure_list_php.txt
+++ b/conformance/failure_list_php.txt
@@ -3,10 +3,6 @@ Recommended.FieldMaskPathsDontRoundTrip.JsonOutput
 Recommended.FieldMaskTooManyUnderscore.JsonOutput
 Recommended.Proto3.JsonInput.BytesFieldBase64Url.JsonOutput
 Recommended.Proto3.JsonInput.BytesFieldBase64Url.ProtobufOutput
-Recommended.Proto3.JsonInput.DurationHas3FractionalDigits.Validator
-Recommended.Proto3.JsonInput.DurationHas6FractionalDigits.Validator
-Recommended.Proto3.JsonInput.DurationHas9FractionalDigits.Validator
-Recommended.Proto3.JsonInput.DurationHasZeroFractionalDigit.Validator
 Recommended.Proto3.JsonInput.FieldMaskInvalidCharacter
 Required.Proto3.JsonInput.FloatFieldTooLarge
 Required.Proto3.JsonInput.FloatFieldTooSmall
--- a/php/src/Google/Protobuf/Internal/EnumDescriptor.php
+++ b/php/src/Google/Protobuf/Internal/EnumDescriptor.php
@@ -39,17 +39,26 @@ class EnumDescriptor
 
     public function getValueByNumber($number)
     {
-        return $this->value[$number];
+        if (isset($this->value[$number])) {
+            return $this->value[$number];
+        }
+        return null;
     }
 
     public function getValueByName($name)
     {
-        return $this->name_to_value[$name];
+        if (isset($this->name_to_value[$name])) {
+            return $this->name_to_value[$name];
+        }
+        return null;
     }
 
     public function getValueDescriptorByIndex($index)
     {
-        return $this->value_descriptor[$index];
+        if (isset($this->value_descriptor[$index])) {
+            return $this->value_descriptor[$index];
+        }
+        return null;
     }
 
     public function getValueCount()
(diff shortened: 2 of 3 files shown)


lhecker:
@TeBoring Mergeable is telling me to include a language label but I don't quite understand how I'm supposed to do that. And does this PR belong to the "Various bug fixes" banner that's already in the CHANGES.txt?

TechProofreader:
I'm having the same problem with my pull request too, so you're not alone. I submitted my pull request to one of the docs in protobuf, and it initially passed all "tests," but then it stopped and said I'm missing labels for languages and a ton of other things even though (1) there's no option for that, and (2) the doc I submitted a pull for contains multiple languages so selecting a single language is impossible.

TeBoring:
We have conformance test at https://github.com/protocolbuffers/protobuf/blob/master/conformance/binary_json_conformance_suite.cc
Would you mind also adding the case there?
If other languages cannot pass the test, you can whitelist the test for them in their failure list.

TeBoring:
To run conformance test, ```cd conformance; make test_php```

lhecker:
@TeBoring Thanks for your advice.
I had to rebase my PR because the initial attempt didn't work with negative nanos.

I tried running the conformance tests by checking out protobuf and running:
```
./autogen.sh
/configure
make
cd conformance
make test_php
```

but the runner gets stuck soon. The cause is a "Undefined offset: 123" notice in the stderr output.
It's caused by this line: https://github.com/protocolbuffers/protobuf/blob/b4f193788c9f0f05d7e0879ea96cd738630e5d51/php/src/Google/Protobuf/Internal/EnumDescriptor.php#L42
`getValueByNumber` is called in two different places and both assume that the return value is nullable, even though it's written in such a way that implies that `$number` is always a valid key.
I think I'm going to fix this as part of this PR as I'm unable to run the conformance tests otherwise.

lhecker:
@TeBoring This PR fixes the `DurationHasZeroFractionalDigit` conformance test and I removed it from the failure list.

I don't understand why it doesn't fix e.g. the `DurationHas3FractionalDigits` test though...
Why is there a trailing zero in the first place? https://github.com/protocolbuffers/protobuf/blob/b4f193788c9f0f05d7e0879ea96cd738630e5d51/conformance/binary_json_conformance_suite.cc#L1947
Even though there's supposed to be _no_ trailing zeroes when being formatted: https://github.com/protocolbuffers/protobuf/blob/b4f193788c9f0f05d7e0879ea96cd738630e5d51/src/google/protobuf/duration.proto#L98-L100
Do you know why that's the case? Is the `duration.proto` description incorrect?

TeBoring:
yntax error, unexpected '?' in /tmp/protobuf/protobuf/php/src/Google/Protobuf/Internal/EnumDescriptor.php on line 42

TeBoring:
In our spec for serialization, we require nanosec has 0/3/6/9 digits precision.
The example in duration.proto does follow our spec.

lhecker:
@TeBoring Ah I used the null-coalescing operator - I'm very sorry. 😔
I'll fix my PR until tomorrow and make sure to also properly test using PHP 5.5 this time.
I'll also try to make the duration formatting standard compliant if you don't mind.

lhecker:
@TeBoring I've updated this PR finally.
I've run the conformance suite with PHP 5.5 as well this time. (There were some errors like `Error parsing text-format protobuf_test_messages.proto3.TestAllTypesProto3: 1:18: Integer out of range (18446744073709551616)` though.)

This PR also fixes the 0/3/6/9 digit formatting. If you dislike that I can open a separate PR for that.

lhecker:
@TeBoring (Just in case you missed it...) I think you can trigger kokoro again.
Also:
> This PR also fixes the 0/3/6/9 digit formatting. If you dislike that I can open a separate PR for that.

TeBoring:
Thanks for the fix!

PR #6160, opened 2019-05-21
Update protobuf version to 3.8.0

haon4:


TeBoring:
https://github.com/protocolbuffers/protobuf/milestone/28
Two php issues still need to be fixed.

haon4:
Bo just merged his PHP fix. I think we can proceed now.

PR #6162, opened 2019-05-22
Move tool out of inputs in bazel config

keith:
This fixes compatibility with bazel when passing `--incompatible_no_support_tools_in_action_inputs` which will be flipped in an upcoming release.


keith:
@acozzette think you could review this one? 🙏

acozzette:
Thanks, @keith!

keith:
Thanks!

PR #338, opened 2015-05-03
ruby: Encode decode cleanup and behavior normalization

skippy:
Hey @cfallin 

This PR may be a bridge too far.  I am sending this PR as *DO NOT MERGE*; if you agree with the direction of the changs, I'll add some tests (in the old-style format) and documentation.  

This pr started as just wanting to add common ruby syntactic sugar, specifically `#to_json` and `#to_proto`.  Then I realized that the jruby `.encode_json` was broken ([gist](https://gist.github.com/skippy/89cafb280017d52a582c)).  But it also had the very nice and useful `#to_h` which the c-ext did not.

I tried consolidating the json encoding/decoding into pure ruby, using the stdlib JSON module, the JSON gem (using the c-extension), and Yajl.  This ripped out a lot of c code, and a bit of java code, and I liked the simplification.  Until I ran benchmarks.  tl;dr: the native json library (haberman/upb) in this lib was consistantly 10x faster ([gist](https://gist.github.com/skippy/93d6c51947960260ab63)).  Yowzers....

I don't want to reopen the conversation in #121, and definitely not the tenor.  However, I agree with the concerns that having two fully separate implementations is going to be hard to maintain.  I'm finding more inconsitancies than I expected, and I'm just starting to scratch the surface.  (e.g., another is that the jruby extension doesn't check STRING encoding during init. like the c-ext does).  *BUT* it is alpha :)

If the JSON performance was closer, I would have submitted the PR with consolidated logic and have tried to make the case that moving some code into ruby is a good thing from a consistancy and maintenance perspective.  However, the performance difference is so great I don't feel I should make that argument.

So here is a less-aggressive PR with fewer changes.  Again, if this is moving in an acceptable direction, I'll finish it up.

PR #6166, opened 2019-05-22
[Ruby] Remove to_hash methods

blowmage:
The Ruby implementation creates the explicit coercion `#to_h` method and the implicit coercion `#to_hash` method on Message and Map. Ruby treats objects that respond to implicit coercion methods differently. I do not think Protobuf intends that Message and Map are to be treated implicitly as hashes. In fact, I think this should actually be considered a bug.

For example, here are how the explicit and implicit methods such as `#to_h` and `#to_hash` are used:

```ruby
# assuming these primitive objects
ary = [1, 2, 3]
hash = { foo: :bar }
int = 42
str = "Hello world!"

# hash has an array representation it itself
hash.to_a #=> [[:foo, :bar]]
# but that doesn't mean it _is_ an array
hash.to_ary # NoMethodError (undefined method `to_ary' for {:foo=>:bar}:Hash)

# similarily, an array doesn't know how to be a hash
ary.to_h #TypeError (wrong element type Integer at 0 (expected array))
# but an array does have a string representation
ary.to_s #=> "[1, 2, 3]"
# and a hash has a string representation
hash.to_s #=> "{:foo=>:bar}"
# and an integer has a string representation
int.to_s #=> "42"
# but none of them _are_ strings
ary.to_str # NoMethodError (undefined method `to_str' for [1, 2, 3]:Array)
hash.to_str # NoMethodError (undefined method `to_str' for {:foo=>:bar}:Hash)
int.to_str # NoMethodError (undefined method `to_str' for 42:Integer)

# and a string doesn't know how to be an array, hash, or integer
str.to_a # NoMethodError (undefined method `to_a' for "Hello world!":String)
str.to_h # NoMethodError (undefined method `to_h' for "Hello world!":String)
str.to_i # NoMethodError (undefined method `to_i' for "Hello world!":String)
```

Similarily, Protobuf Messages and Maps are not Ruby Hashes. They can generate hash representations, but they are not themselves hashes. They should not be used in place of hashes.

What does the presence of the `#to_hash` method mean? It means Ruby will treat these objects as if they were hashes. Consider the following method:

```ruby
def do_a_thing req_resource, opt_resource = nil, show_warnings: true
  p req_resource
  p opt_resource
  p show_warnings
end
```

You would expect you could pass two Protobuf objects to the method, but you cannot.

```ruby
gem "googleapis-common-protos-types"
require "google/type/color_pb"

red = Google::Type::Color.new red: 1.0
blue = Google::Type::Color.new blue: 1.0

# call the method
do_a_thing red, blue # ArgumentError (unknown keywords: red, green, blue, alpha)
```

This is because Ruby sees that the `blue` object has defined `#to_hash` and treats it as a hash, and tries to match the optional named argument instead of the positional object.

Because of this, and because there is no documentation that states that this method is intended to be part of the public API, I think these methods can and should be removed.

### DIFF
--- a/ruby/ext/google/protobuf_c/map.c
+++ b/ruby/ext/google/protobuf_c/map.c
@@ -847,7 +847,6 @@ void Map_register(VALUE module) {
   rb_define_method(klass, "dup", Map_dup, 0);
   rb_define_method(klass, "==", Map_eq, 1);
   rb_define_method(klass, "hash", Map_hash, 0);
-  rb_define_method(klass, "to_hash", Map_to_h, 0);
   rb_define_method(klass, "to_h", Map_to_h, 0);
   rb_define_method(klass, "inspect", Map_inspect, 0);
   rb_define_method(klass, "merge", Map_merge, 1);
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -737,7 +737,6 @@ VALUE build_class_from_descriptor(Descriptor* desc) {
   rb_define_method(klass, "eql?", Message_eq, 1);
   rb_define_method(klass, "hash", Message_hash, 0);
   rb_define_method(klass, "to_h", Message_to_h, 0);
-  rb_define_method(klass, "to_hash", Message_to_h, 0);
   rb_define_method(klass, "inspect", Message_inspect, 0);
   rb_define_method(klass, "to_s", Message_inspect, 0);
   rb_define_method(klass, "[]", Message_index, 1);
--- a/ruby/src/main/java/com/google/protobuf/jruby/RubyMap.java
+++ b/ruby/src/main/java/com/google/protobuf/jruby/RubyMap.java
@@ -338,7 +338,7 @@ public class RubyMap extends RubyObject {
         return newMap;
     }
 
-    @JRubyMethod(name = {"to_h", "to_hash"})
+    @JRubyMethod(name = "to_h")
     public RubyHash toHash(ThreadContext context) {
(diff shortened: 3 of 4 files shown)


blowmage:
The PR (#338) and commit (d1b52a00e0) that added the implicit coercion methods only mentions the explicit use-case, and [this comment in particular](https://github.com/protocolbuffers/protobuf/pull/338/files#r29558567) makes me think the consequences of including the implicit methods was probably not fully understood and unintentional.

haberman:
Yes this appears to have been unintentional and not fully understood. I wasn't aware of this corner of Ruby and it's surprising to me that you can't pass two actual hashes to a function that takes optional named arguments.

I agree that it does seem better overall to not implement `#to_hash`. However I'm concerned about the possibility of breaking users who may possibly depend on the current behavior. Since Ruby classes are always open, could users who are broken by this add `#to_hash` to their classes manually? I want to avoid a scenario where users are blocked from upgrading to the new version because it breaks their code.

blowmage:
> Since Ruby classes are always open, could users who are broken by this add `#to_hash` to their classes manually?

Yes, users can add methods to the Protobuf resource classes to restore `#to_hash` if they are dependent on it.

blowmage:
> it's surprising to me that you can't pass two actual hashes to a function that takes optional named arguments.

Matz has stated that the way named arguments were implemented was a mistake and this will be changed in Ruby 3. I assume he means this issue of unrolling the last positional Hash object and matching the named arguments.

blowmage:
Ping.

derekargueta:
I was surprised to come across this because the documentation still states that `to_hash` is implemented. How can we get the documentation updated? @haberman

PR #6169, opened 2019-05-23
Append field number to accessors if there is conflict

TeBoring:
Previously, foo_value and foo (wrapper field) will both have get/setFooValue

### DIFF
--- a/php/tests/proto/test_wrapper_type_setters.proto
+++ b/php/tests/proto/test_wrapper_type_setters.proto
@@ -24,3 +24,14 @@ message TestWrapperSetters {
 
   map<string, google.protobuf.StringValue> map_string_value = 13;
 }
+
+message TestWrapperAccessorConflicts {
+  int32 normal_vs_wrapper_value = 1;
+  google.protobuf.Int32Value normal_vs_wrapper = 2;
+
+  int32 normal_vs_normal_value = 3;
+  int32 normal_vs_normal = 4;
+
+  google.protobuf.Int32Value wrapper_vs_wrapper_value = 5;
+  google.protobuf.Int32Value wrapper_vs_wrapper = 6;
+}
--- a/php/tests/wrapper_type_setters_test.php
+++ b/php/tests/wrapper_type_setters_test.php
@@ -16,6 +16,44 @@ use Google\Protobuf\UInt64Value;
 
 class WrapperTypeSettersTest extends TestBase
 {
+    public function testConflictNormalVsWrapper()
+    {
+        $m = new Foo\TestWrapperAccessorConflicts();
+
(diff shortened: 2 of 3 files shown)


php/tests/wrapper_type_setters_test.php
```diff
@@ -16,6 +16,44 @@
 
 class WrapperTypeSettersTest extends TestBase
 {
+    public function testConflictNormalVsWrapper()
```
michaelbausor: It looks like *both* functions have been renamed with a number on the end. Is that necessary? Why not just rename one of them, and keep the other unchanged?

php/tests/proto/test_wrapper_type_setters.proto
```diff
@@ -24,3 +24,14 @@ message TestWrapperSetters {
 
   map<string, google.protobuf.StringValue> map_string_value = 13;
 }
+
+message TestWrapperAccessorConflicts {
+  int32 normal_vs_wrapper_value = 1;
+  google.protobuf.Int32Value normal_vs_wrapper = 2;
```
michaelbausor: Appending the field number works for this example, but what if there was also a field named `normal_vs_wrapper_value_2`? I'm not suggesting that that is very likely, but WDYT about a more general solution to this problem that would guarantee no conflicts. For example, inside a class, create a container of names used for methods, and each time you generate a method, add it to the container, check for conflicts, and if there is a conflict, resolve it by appending an incrementing counter - keep incrementing until the name is unique. This would also cover any other possible cases of collision (I'm not sure if any other cases exist in PHP).

TeBoring: This will forces us to delay generating all accessors until we have traversed all fields of the same message. I feel this will be quite complicate.
For now, we used the same approach in java (which also has the problem of conflict after renaming)
In future, we would like to introducing some rules for naming a field, like no field can be named as a prefix of another field.

michaelbausor: Ok, sgtm.

TeBoring:
@fiboknacky please take a look

php/tests/wrapper_type_setters_test.php
haon4: I think that is necessary. One field may be added before the other, and we don't want to accidentally change which field user is calling.

Consider the case if `int32 normal_vs_wrapper_value = 1;` is added first, and user is calling getNormalVsWrapperValue(), then someone added `google.protobuf.Int32Value normal_vs_wrapper = 2;`. If we rename the field 1 but not field 2, that user is suddenly reading from field 2, which is a behavior change. Similar problem exist for the other ordering too. Not to mention that it is also not clear which field getNormalVsWrapperValue() is referring to.

PR #6176, opened 2019-05-24
Add extension number for protoc-gen-validate

akonradi:
This project is used by the [Envoy proxy](https://www.envoyproxy.io/) and several others.

The existing code is using the extension value [919191](https://github.com/envoyproxy/protoc-gen-validate/blob/54b08a686ff6548d8a248218664254c422fc4599/validate/validate.proto#L29) but since maintaining binary compatibility is not important, it shouldn't be a problem to change once this allocation is approved.


BSBandme:
Sorry but it seems the 1070 has been used by others. Would you please choose another one?

akonradi:
@BSBandme thanks, switched to 1071.

PR #6177, opened 2019-05-24
Add `protobuf_java_lite` Bazel target

cgrushko:
that builds the JavaLite runtime.

Contents based on java/lite/pom.xml.


cgrushko:
cc @haon4 
Mergeable says the PR isn't mergeable because there are missing labels, but I can add them myself.

PR #6183, opened 2019-05-28
avoid error when compiling with MSVC2010

amitport:
I'm getting `src/google/protobuf/map_field.cc(284): error C2899: typename cannot be used outside a template declaration`

removing `typename` fixes the issue


acozzette:
Thanks, @amitport!

PR #6184, opened 2019-05-28
Just reopening #6128

tenderlove:
I forgot to commit some stuff in #6128, this just adds it.


haberman:
Looks like tests are failing:

> /usr/local/rvm/rubies/ruby-2.3.8/bin/ruby: symbol lookup error: /tmp/protobuf/protobuf/ruby/lib/google/protobuf_c.so: undefined symbol: RepeatedField_deep_copy

PR #6203, opened 2019-05-31
JSON string output functions should clear() the destination first

boscosiu:
Make the JSON util string output functions match the Message::SerializeToString() behavior.  This is helpful when packing the output into string fields of reused protobuf messages.


boscosiu:
I signed it!

acozzette:
Thanks, @boscosiu!

PR #6206, opened 2019-06-03
Treat plugins as host tools.

benjaminp:
As a followup to ca3ead7745581b24d9e84471b5dee6921fcd369d, treat plugin executables as well as protoc as host tools.

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -117,6 +117,7 @@ def _proto_gen_impl(ctx):
 
         outs = [ctx.actions.declare_file(out, sibling = src) for out in outs]
         inputs = [src] + deps
+        tools = [ctx.executable.protoc]
         if ctx.executable.plugin:
             plugin = ctx.executable.plugin
             lang = ctx.attr.plugin_language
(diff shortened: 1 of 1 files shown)


meteorcloudy:
Thanks! @acozzette Can we merge this into 3.8.0 branch?

acozzette:
Thanks, @benjaminp!

@meteorcloudy I went ahead and merged this to master because that's what the pull request was based on, but we can cherry-pick it to 3.8 if we need to. Do you need a 3.8.1 release with this fix?

benjaminp:
Thanks for merging.

It would be nice to have this and ca3ead7745581b24d9e84471b5dee6921fcd369d in a release soon because otherwise `protobuf.bzl` is not compatible with Bazel 0.27 out of the box.

meteorcloudy:
@acozzette I totally agree with @benjaminp , TensorFlow cannot migrate to Bazel 0.27 if we don't have this fix in a release.

acozzette:
@haon4 Do you have any thoughts on when we should do the next release?

If this is just for compatibility with Bazel 0.27 then I'm not sure we can justify doing a point release for that alone. Protobuf releases are a fair amount of work since we support so many languages.

haon4:
I think we can cut the new release any time. We plan to do it twice this quarter, so it maybe a good time to cut a new release now.

meteorcloudy:
@acozzette I'm also fine with just cherry-pick this change into the 3.8.x branch or having a 3.8.0.1 tag like https://github.com/protocolbuffers/protobuf/releases/tag/v3.6.1.3 would also be nice. I don't think we need to do a full release only for this change since it won't affect any protobuf binaries.

acozzette:
OK, cherry-picking and tagging 3.8.0.1 is pretty easy to do. Do you want to send a pull request cherry-picking the necessary commits to 3.8.x and then I can merge it and create the new tag?

meteorcloudy:
@acozzette Thank you very much! It's here: https://github.com/protocolbuffers/protobuf/pull/6248

PR #6221, opened 2019-06-06
ObjC: avoid assign for object properties

ahunt:
This avoids hitting the following warning when including GPBDescriptor.h in ARC
projects with stricter warning settings:
error: 'assign' property of object type may become a dangling reference; consider using 'unsafe_unretained' [-Werror,-Wobjc-property-assign-on-object-type

We first hit this in XCode 11, it's unclear whether it simply got stricter about
enforcing this warning or whether the flags enabled with -Wall or -Weverything
have been expanded.

### DIFF
--- a/objectivec/GPBDescriptor.h
+++ b/objectivec/GPBDescriptor.h
@@ -75,7 +75,7 @@ typedef NS_ENUM(uint8_t, GPBFieldType) {
 /** Number of extension ranges declared for the message. */
 @property(nonatomic, readonly) uint32_t extensionRangesCount;
 /** Descriptor for the file where the message was defined. */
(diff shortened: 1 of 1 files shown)


thomasvl:
The CocoaPods Integration failure seems to be a networking hiccup.

jpoumailloux:
Should the following property declaration from GPBExtensionDescriptor be updated?
@property(nonatomic, readonly, assign) Class msgClass;

thomasvl:
> Should the following property declaration from GPBExtensionDescriptor be updated?
> \@property(nonatomic, readonly, assign) Class msgClass;

probably, feel free to send another PR.

Issue #6186, opened 2019-05-29
Python module does not compile, because google/protobuf/stubs/map_util.h is missing


moubctez:
**What version of protobuf and what language are you using?**
Version: v3.8.0
Language: C++/Python

**What operating system (Linux, Windows, ...) and version?**
Any

**What runtime / compiler are you using (e.g., python version or gcc version)**
Clang 8.0.0/Python 3.7.3

**What did you do?**
Steps to reproduce the behaviour:
1. Build and install protobuf-cpp 3.8.0
2. Build and install protobuf-python 3.8.0 (I used the source distribution from PyPI)

**What did you expect to see**
Building Python module succeeds.

**What did you see instead?**
Building fails with `google/protobuf/pyext/map_container.cc:46:10: fatal error: 'google/protobuf/stubs/map_util.h' file not found`

**Proposed fix**
Apply this patch
```
--- src/Makefile.am.orig	2019-05-29 10:28:11.000000000 +0000
+++ src/Makefile.am
@@ -72,6 +72,7 @@ nobase_include_HEADERS =                
   google/protobuf/stubs/hash.h                                   \
   google/protobuf/stubs/logging.h                                \
   google/protobuf/stubs/macros.h                                 \
+  google/protobuf/stubs/map_util.h                               \
   google/protobuf/stubs/mutex.h                                  \
   google/protobuf/stubs/once.h                                   \
   google/protobuf/stubs/platform_macros.h                        \
```

PR #6223, opened 2019-06-06
Added stubs/map_util.h to list of installed headers

acozzette:
This should fix issue #6186.


acozzette:
I'm not sure why the CLA bot was complaining, but this PR is fine because I am the only author.

PR #6226, opened 2019-06-07
Fix for PHP 7+

remicollet:


remicollet:
Sorry, but have to go away from this damned CLA

PR #6228, opened 2019-06-07
Add async overloads for ByteString.WriteTo()

erikma:
Related to #3166

### DIFF
--- a/csharp/src/Google.Protobuf/ByteString.cs
+++ b/csharp/src/Google.Protobuf/ByteString.cs
@@ -404,6 +404,7 @@ namespace Google.Protobuf
         /// <summary>
         /// Writes the entire byte array to the provided stream.
         /// </summary>
+        /// <param name="outputStream">The stream to which to write.</param>
         public void WriteTo(Stream outputStream)
         {
             outputStream.Write(bytes, 0, bytes.Length);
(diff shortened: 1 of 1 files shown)


acozzette:
CC @jskeet @anandolee 

@jtattermusch Would you be interested in reviewing this? @anandolee usually reviews C# changes but is on vacation for the next month.

erikma:
@acozzette Would you mind adding the "C#" tag here and in my other PR? The mergeable-bot fails on that missing label, and I can't add labels

jtattermusch:
Same as https://github.com/protocolbuffers/protobuf/pull/6230#issuecomment-500869085 can you describe the use case?
Also, on netstandard2.0 you can use the .Span property and write to an output stream (or to a pipewriter) in a more flexible fashion.

erikma:
Good point and I'll move toward that. In this case I'm working in a net472 codebase, the .Span property is not visible in this case, so I was optimizing for what was available to minimize thread waits.

erikma:
@jtattermusch Memory<T> only supported on netcore2.1+. I'd prefer to get this change in for downlevel net472 optimization.

jtattermusch:
Same response as on https://github.com/protocolbuffers/protobuf/pull/6230#issuecomment-506685273

erikma:
@jtattermusch This is different from #6230 as it simply adds the ability to write to a stream without holding a thread context. I closed that PR, but this one still seems useful by itself.

jtattermusch:
Once we add the System.Memory dependency for all targets (see https://github.com/protocolbuffers/protobuf/pull/6317), it seems that this will no longer be useful, because you can grab byteString.Span  and use that to write to the Stream?

PR #6229, opened 2019-06-07
Fix Alphabetization of BUILD

arashikou:
This was found while integrating Protobuf 3.8.0 into a downstream project. @SuperOxigen noticed that our build files, which are autogenerated from the BUILD file, were out of order. So that we don't have to keep manually fixing our build files, I am pushing this fix upstream.


arashikou:
Release notes: no
Cleanup

PR #6230, opened 2019-06-07
Add overload of ByteString.CopyTo that can copy a slice

erikma:

### DIFF
--- a/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
+++ b/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
@@ -233,5 +233,21 @@ namespace Google.Protobuf
             ByteString b2 = ByteString.CopyFrom(200, 1, 2, 3, 4);
             Assert.AreNotEqual(b1.GetHashCode(), b2.GetHashCode());
         }
+
+        [Test]
+        public void CopyTo()
+        {
+            byte[] sourceBytes = { 1, 2, 3, 4, 5 };
+            ByteString b = ByteString.CopyFrom(sourceBytes);
+
+            var dst = new byte[5];
+            b.CopyTo(dst, 0);
+            CollectionAssert.AreEqual(sourceBytes, dst);
+
+            dst = new byte[2];
+            b.CopyTo(3, dst, 0, 2);
+            Assert.AreEqual(4, dst[0]);
+            Assert.AreEqual(5, dst[1]);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


jtattermusch:
What is the use case?
Btw, if you are on netstandard2.0 (which has been around for a while now), you don't need this API. You can use ByteString.Span  method to gain readonly access to byteString's data directly (which is more efficient and flexible than the API you're proposing).
https://github.com/protocolbuffers/protobuf/blob/21e4b1c3366fe6831ed5c62ca974166e6e4d98c2/csharp/src/Google.Protobuf/ByteString.cs#L118

erikma:
I'll move toward that over time. In this case I'm working in a net472 codebase, the .Span property is not visible in this case, so I was optimizing for what was available to minimize thread waits.

erikma:
@jtattermusch Memory<T> only supported on netcore2.1+. I'd prefer to get this change in for downlevel net472 optimization.

jtattermusch:
the point about supporting optimizations for net472 is fair. I think we could add `CopyFrom(ReadOnlySpan<byte> bytes)` (see https://github.com/protocolbuffers/protobuf/pull/5835) to the net45 target too, which would make a cleaner solution than what this PR proposes.

@jskeet  @mgravell @JamesNK  WDYT?

mgravell:
I guess it comes down that "unity" discussion again; i.e. if "unity" is a supported target, what TFM does "unity" consume? if it is netstandard1.*, then: great - no harm, no foul; if it is net45 then it gets into that awkward spot of

1. is there an API that unity can't exploit - which to be clear **isn't a problem** - there are other APIs, but more importantly:
2. does the generated code now *use that API by default*, making it impossible to compile on the down-level unity compiler

jskeet:
Marc's given a better response than I would have done, basically :)
(And I suspect the answer to the question of Unity TFMs is going to be really complex, based on the version of Unity and whatever target the app is being built for.)

jtattermusch:
The current version of unity supports both net45 and netstandard2.0, so both TFM targets can be consumed by unity users and it depends on them what they choose.

- the generated code currently doesn't use any API that would require require Span<> so from that respect there should be no problem on Unity

erikma:
@jtattermusch Can you clarify the result of the thread above? Do I need to add #if for a net45 target, or is this PR acceptable as-is?

erikma:
Closing in deference to a possible re-do as expansion of #5835 to a net45 target.

jtattermusch:
I think the plan is to make Google.Protobuf depend on System.Memory for all targets (see https://github.com/protocolbuffers/protobuf/pull/6317 which is currently blocked on some nugets being released as stable) and with that everyone will be able to access the data with byteString.Span property (without needing an additional CopyTo method).

Issue #4107, opened 2017-12-27
PHP c extension doesn't work with mock well


bshaffer:
I am receiving a segmentation fault when using the current version of the protobuf C extension for PHP. This is duplicated by [this travis test for PHP 5.6](https://travis-ci.org/googleapis/gax-php/jobs/322270222#L2829) and [this travis test for PHP 7.2](https://travis-ci.org/googleapis/gax-php/jobs/322270226#L2797):

```
$ vendor/bin/phpunit --coverage-clover=coverage.xml
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.
...............................................................  63 / 276 ( 22%)
...../home/travis/.travis/job_stages: line 57: 30853 Segmentation fault      (core dumped) vendor/bin/phpunit --coverage-clover=coverage.xml

```

I can reproduce this on my local machine for PHP 5.6 and PHP 7.2 also.

This may be fixed by #4102, but strangely enough the 3.5.1 version of the protobuf extension has [already been released in PECL](https://pecl.php.net/package/protobuf/3.5.1). Can someone shed some light on this situation?

In the meantime I'll do some digging to see what the exact line is which causes the fault.

TeBoring:
Did u add new test to reproduce the bug? We didn't find segment fault in
our Travis test.

bshaffer:
Here is a test to reproduce the bug:

```
$ git clone -b json-transport git@github.com:googleapis/gax-php.git && cd gax-php && composer install && vendor/bin/phpunit
```

This only fails when the protobuf C extension is enabled.

bshaffer:
Okay I tracked down the issue and was able to reproduce it with the following code:

```php
<?php

class TestSegfault extends Google\Protobuf\Internal\Message
{
    private $foo;

    public function setFoo($foo)
    {
        $this->foo = $foo;
    }
}

$message = new TestSegfault();
$message->setFoo('any value');
```

Paste that into any file and execute it with PHP when the protobuf extension is enabled and the result will be a segfault.

Is this happening because there is no Metadata being initialized?

TeBoring:
Please try initialize metadata.

bshaffer:
@TeBoring Protobuf should be segfaulting because metadata hasn't been initialized...

bshaffer:
It should be mentioned that if the native `google/protobuf` implementation is used, the segmentation fault does not occur. It's only when using the protobuf C extension that this happens.

bshaffer:
[...]

PR #6256, opened 2019-06-13
Enable mock for proto message

TeBoring:
Previously mock cannot be used for proto message because it cannot be extended.
Now, it can be extended by doing:
1) Avoid freeing internal data, since it was not initialized for subclass
2) Resort to standard handlers for subclass's properties


TeBoring:
Fix #4107

haon4:
Do we still want this PR?

PR #6268, opened 2019-06-17
Update GSON version to 3.8.5

### LINKED ISSUE #3531: Update protobuf-java-util gson dep from v2.7 to v2.8.1
protobuf-java-util isn't using the latest gson. I had to add a gradle exclude rule. It'd be nice to update to the latest stable version.

haon4:
Fixes #3531

### DIFF
--- a/java/util/pom.xml
+++ b/java/util/pom.xml
@@ -35,7 +35,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.7</version>
+      <version>2.8.5</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>


TeBoring:
In #3531, there seems to be some concern about compatibility of gson.

haon4:
Sorry I missed your last comment. We upgraded the Gauva library in one of our previous releases and there wasn't an issue. I don't think GSON would be that different, right?

henryju:
This issue is marked as closed and the changelog indicates it was released in protobuf 3.9.0. But the pom on Maven central is still referencing gson 2.7:
http://repo1.maven.org/maven2/com/google/protobuf/protobuf-java-util/3.9.1/protobuf-java-util-3.9.1.pom

henryju:
And BTW the changelog and this PR title are wrong, the new version is 2.8.5, not 3.8.5.

PR #6269, opened 2019-06-17
Revert "JSON string output funtions should clear() the destination fi…

acozzette:
…rst"

We found that some code depended on the behavior of appending to the
string without clearing it first. The comments in json_util.h also make
it appear that the appending behavior is intentional, so I think we need
to stick with the old behavior.

This reverts commit 3cbe45a4b55952609dff0b3a9e2e7043679af93f.


acozzette:
@boscosiu Sorry for the bad news but I'm afraid we have to revert your change from #6203 because we found some code that depended on the old behavior.

PR #6277, opened 2019-06-18
Make php message class final to avoid mocking

TeBoring:


TeBoring:
@michaelbausor PTAL

dwsupplee:
@TeBoring This change is breaking, can we please revert it?

TeBoring:
This changed is intended. Do you know why the proto message needs to be
extended?

PR #6279, opened 2019-06-19
Add "src/**/*.inc" to //:protobuf_headers

meteorcloudy:
Otherwise, sandbox and remote execution will complain `google/protobuf/port_def.inc` doesn't exist.


meteorcloudy:
@acozzette

PR #6281, opened 2019-06-19
Merge 3.8.x into 3.9.x

acozzette:
There are a few fixes on the 3.8.x branch that we need to merge into 3.9.x.


acozzette:
This PR is fine with respect to the CLA because it is just merging commits that have already had their CLA verified.

PR #6282, opened 2019-06-19
Revert "Append field number to accessors if there is conflict"

TeBoring:
Reverts protocolbuffers/protobuf#6169
No longer needed.
@michaelbausor


haon4:
Why is this no longer needed? We may still run into conflict, right? (Just less likely)

TeBoring:
This is specifically added for xxxValue instead of xxxWrapper.
Because _wrapper is less common, this may not be necessary.
Instead some other long term solution may be needed to solve the conflict problem.
For now, I think we don't need to treat xxxWrapper differently.

PR #6288, opened 2019-06-20
Migrate protobuf for https://github.com/bazelbuild/bazel/issues/7152

hlopko:
This PR migrates protobuf.bzl for Bazel's incompatible change https://github.com/bazelbuild/bazel/issues/7152.

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -83,8 +83,8 @@ def _proto_gen_impl(ctx):
         import_flags = ["-I."]
 
     for dep in ctx.attr.deps:
-        import_flags += dep.proto.import_flags
-        deps += dep.proto.deps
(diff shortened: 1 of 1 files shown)


acozzette:
@hlopko If this is something we need to release soon then you might want to submit this change against the 3.9.x branch so that we include it in version 3.9.0.

hlopko:
Thank you Adam, I created https://github.com/protocolbuffers/protobuf/pull/6289. Not sure if we need both PRs or only one of them, feel free to close what's not needed.

Thank you!

acozzette:
Thanks, Marcel. Let me close this one because the other one should suffice on its own. We will merge 3.9.x into master later so that master picks up the various fixes we're making in preparation for the 3.9.0 release.

PR #6300, opened 2019-06-24
Merge 3.9.x into master

acozzette:


haon4:
I don't think we want to update the version number in master yet. We just need to fix the json_util, right?

acozzette:
Is there anything wrong with updating the version number now? To me that seems like the best workflow: important fixes get committed to the release branch and periodically merged into master. Otherwise every fix will have to get duplicated because we will have to commit it on one branch and cherry-pick it to the other.

haon4:
I don't think there is anything wrong. It is just different from what we used to do, but I am fine with this too.

PR #6304, opened 2019-06-26
Fix usage of deprecated single_file attribute

codesuki:
This removes the need to use --incompatible_disable_deprecated_attr_params when using protobuf.


BSBandme:
Close as it duplicates with #6301

PR #6307, opened 2019-06-26
Add an option to preserve proto names in JsonFormatter

### LINKED ISSUE #6299: C# JsonFormatter does not have a preservingProtoFieldNames method
If it's a proto syntax change, is it for proto2 or proto3?
If it's about generated code change, what programming language?
C#, Proto3

The `JsonFormatter` does not have a `preservingProtoFieldNames` method like Java; see [JsonFormat.Printer](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/util/JsonFormat.Printer#preservingProtoFieldNames--)

Adding the method, like in Java; simliar issue #2029

None

**Additional context**
Example:
```proto
message FullName {
  string first_name = 1;
  string last_name = 2;
}
```
```csharp
var name = new FullName {FirstName = "John", LastName = "Doe"};
var jsonName = JsonFormatter.Default.Format(name);
Console.Out.WriteLine(jsonName);
```

Output is:
```text
{ "firstName": "John", "lastName": "Doe" }
```

ObsidianMinor:
Fixes #6299


ObsidianMinor:
This has been sitting here for quite a bit. Any updates? @anandolee

ahmelsayed:
I was looking for this option as well. Any updates on merging this PR?

ObsidianMinor:
Could you also review this as well @jtattermusch?

ahmelsayed:
@anandolee Any updates on this PR?

anandolee (approved):

jtattermusch:
@jskeet  @amanda-tarafa this probably deserves to be looked at by a pair of C# expert's eyes - just to make sure nothing was missed (I'll try to take a look as well but can't guarantee).

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
```diff
@@ -1,698 +1,705 @@
...
-        /// to read.
-        /// </summary>
-        private static void AssertJson(string expectedJsonWithApostrophes, string actualJson)
-        {
-            var expectedJson = expectedJsonWithApostrophes.Replace("'", "\"");
-            Assert.AreEqual(expectedJson, actualJson);
-        }
-    }
-}
+﻿#region Copyright notice and license
```
jtattermusch, line 1: what has changes about this file that it shows up in the diff?
Unless I missed something, the entire file showing up as modified makes it impossible to see any actual changes to the file (which makes reviewing it very difficult).

Let's avoid "false diffs" like this in the future.

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
```diff
@@ -1,698 +1,705 @@
...
+
+        [Test]
+        public void WriteValue_Message()
+        {
+            var value = new TestAllTypes { SingleInt32 = 100, SingleInt64 = 3210987654321L };
+            AssertWriteValue(value, "{ 'singleInt32': 100, 'singleInt64': '3210987654321' }");
+        }
+
+        [Test]
+        public void WriteValue_Message_PreserveNames()
```
amanda-tarafa, line 666: I think this is the only test that has been added.

amanda-tarafa (commented):
I've done a quick review as well. It looks good.

But I'm wondering about the usefulness of this without the counterpart change in JsonParser. But there might be something I'm missing here.

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
jskeet: I believe it's whitespace-only - quite possibly line endings. I'd really like to see the whitespace-only changes here reverted before we merge; it makes it much hard to trace changes.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -233,7 +233,14 @@ private bool WriteMessageFields(TextWriter writer, IMessage message, bool assume
                     writer.Write(PropertySeparator);
                 }
 
-                WriteString(writer, accessor.Descriptor.JsonName);
+                if (settings.PreserveProtoFieldNames)
```
jskeet, line 236: I think I'd do this with a conditional operator instead:

```csharp
WriteString(writer, settings.PreserveProtoFieldNames ? accessor.Descriptor.Name : accessor.Descriptor.JsonName);
```

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -816,6 +823,11 @@ static Settings()
             /// </summary>
             public bool FormatEnumsAsIntegers { get; }
 
+            /// <summary>
+            /// Whether to use the original proto field names as defined in the .proto file. Defaults to false.
+            /// </summary>
+            public bool PreserveProtoFieldNames { get; }
+
```
jskeet, line 830: Nit: please remove redundant blank line.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -842,32 +854,41 @@ public Settings(bool formatDefaultValues, TypeRegistry typeRegistry) : this(form
...
             /// Creates a new <see cref="Settings"/> object with the specified enums formatting option and the current settings.
             /// </summary>
             /// <param name="formatEnumsAsIntegers"><c>true</c> to format the enums as integers; <c>false</c> to format enums as enum names.</param>
-            public Settings WithFormatEnumsAsIntegers(bool formatEnumsAsIntegers) => new Settings(FormatDefaultValues, TypeRegistry, formatEnumsAsIntegers);
+            public Settings WithFormatEnumsAsIntegers(bool formatEnumsAsIntegers) => new Settings(FormatDefaultValues, TypeRegistry, formatEnumsAsIntegers, PreserveProtoFieldNames);
+
+            /// <summary>
+            /// Creates a new <see cref="Settings"/> object with the specified field name formatting option and the current settings.
+            /// </summary>
+            /// <param name="preserveProtoFieldNames"><c>true</c> to preserve proto field names; <c>false</c> to convert them to lowerCamelCase.</param>
```
jskeet, line 890: Hmm... I *thought* we allowed users to specify a different JSON field name to the one that would be inferred by the compiler otherwise. Not sure - but arguably using "false to use the default JSON field name representation" would cover that either way.

jskeet (commented):
It looks like we don't need an option for this in the parser as we already handle using either the JSON field name or the original name (see https://github.com/protocolbuffers/protobuf/blob/b926a7d2097405fd3be4cc4009d61710562852d7/csharp/src/Google.Protobuf/Reflection/MessageDescriptor.cs#L121).

However, I'm concerned with per-language feature creep. #6299 observes that the option *does* exist in Java, but I don't know about any other languages. I think the Protobuf team should decide whether they're happy for this to be a feature in general. If it's approved in general, then this implementation looks broadly okay.

jskeet:
Eek - I hadn't seen that this had been merged as-is.

PR #6310, opened 2019-06-27
Fix usage of six in //:protobuf_python rule and add import

aaliddell:
Using `//external:six` works when protobuf is the root repo, but fails when protobuf is a third party dependency of another workspace. `@six//:six` works in both situations once the relevant workspace rule has been updated to match.

Also added six to the protobuf_deps.bzl file.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -871,7 +871,7 @@ py_proto_library(
     protoc = ":protoc",
     py_libs = [
         ":python_srcs",
-        "//external:six",
+        "@six//:six",
     ],
     py_extra_srcs = glob(["python/**/__init__.py"]),
     srcs_version = "PY2AND3",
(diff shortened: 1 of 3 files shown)


BSBandme:
Please fix the test errors.

aaliddell:
The four failed so far do not appear to be failed tests. Each has the error `[14:03:29][ERROR] Cannot check out non-mergeable pull request as if merged.`, yet Mergeable, GitHub and the other pipelines seem to think it's conflict free.

So I don't see what I can do to resolve this when the actual tests themselves run fine in the other pipelines? Do you want a rebase or master-merge?

BSBandme:
Alright... Seems it has some errors during last run.
Let's wait for the rerun result though.

aaliddell:
Looks OK on the second time through :thinking:

PR #6316, opened 2019-06-28
add type conversions to generated code javascript setters

### LINKED ISSUE #6086: JS Unexpected behavior when serializing/deserializing
Hello opening this issue because I've seen an unexpected behavior and want to discuss about it and see what's the best pattern

Given the following message:

```
message MyMessage {
    String pkid = 1;
}
```

If I set a number field into pkid, I'm able to retrieve it correctly. Once I serialize and then deserialize the message, the value gets coerced as an empty string:

```
> protoConfig.setPkid(123);
> protoConfig.getPkid();
123
> MyMessage.deserializeBinary((protoConfig.serializeBinary())).getPkid()
""
```

I wasn't expecting the field to be transformed silently once the message is serialized. What I would expect from order of preference:

1. Setting the field with `setPkid` to crash (or a warning) because the type is not what was expected
2. Serialization crashing (or a warning) because the type is not expected
4. Coercing the type using `toString` which would set it to '123'

I understand suggested behaviors may have performance implications but I'm not sure what's the reason of current behavior because this still forces the user to do type checks before setting fields in a protobuf message when using javascript? IMO silently changing the value of a field when serializing a message is dangerous and I would aim for correctness of data first.

r6min:
Fixes #6086

Adds type conversions to the generated javascript setters for primitives & repeated primitives:

First checks if the argument is of the correct type or null (which will clear the field).
If not, convert to the correct type using the appropriate wrapper function(Number(x), String(x) or Boolean(x)).

For repeated fields, first the type of the passed variable is an array. If not, throw an error. If it is an array, iterate through the array and perform the same conversion logic as above for each element.

Examples

protofile:

message MyMessage {
    String pkid = 1;
}
after compiling:

```
> protoConfig.setPkid(123);
> protoConfig.getPkid();
123
> MyMessage.deserializeBinary((protoConfig.serializeBinary())).getPkid()
123
```

protofile:

message MyMessage {
    repeated String pkid = 1;
}

```
> protoConfig.setPkidList([true, 123]);
> protoConfig.getPkidList();
[ 'true', '123' ]
> MyMessage.deserializeBinary((protoConfig.serializeBinary())).getPkidList()
[ 'true', '123' ]
```

### DIFF
--- a/js/message_test.js
+++ b/js/message_test.js
@@ -920,6 +920,37 @@ describe('Message test suite', function() {
     assertBooleanFieldFalse(message.getRepeatedBooleanFieldList()[1]);
   });
 
+  it('testFieldsOfDifferentTypesAreConverted', function() {
+    var assertNan = function(x) {
+        assertTrue(
+            'Expected ' + x + ' (' + goog.typeOf(x) + ') to be NaN.',
+            goog.isNumber(x) && isNaN(x));
+    };
+
+    var assertArrayEqualsWithNan = function(arr1, arr2) {
+        assertTrue('Arrays are not equal. Expected [' + arr2 + '] and got [' + arr1 + ']',
+            arr1.length === arr2.length && arr1.every((element, index) => {
+                return (element === arr2[index]) || (isNaN(element) && isNaN(arr2[index]));
+            })
+        );
+    }
+    
+    var message = new proto.jspb.test.FloatingPointFields;
+
+    message.setOptionalFloatField('123');
+    message.setRequiredFloatField('12a3');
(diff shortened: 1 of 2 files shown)


TeBoring:
I think the better approach is have type checking. Besides, we are working on some type checking internally.

r6min:
@TeBoring Yes, I thought about this initially, however there are some type conversions that are supported already (passing ints & converting to booleans), so it was unclear what to allow and what to prevent.

r6min:
@TeBoring if you have a solution internally for this already I can close this

PR #6320, opened 2019-06-30
1.try fix compile error on Android &UE4

RPG3D:
get error: undefined bswapXXX on Android & UE4; so add a macro for Android


RPG3D:
I signed it!

acozzette:
@RPG3D On Android I would expect these bswap functions to come from `byteswap.h`. That header should get included on line 85 of `port.h`. Do you know why that is not working?

RPG3D:
> that is not w

UE4 has an header ByteSwap.h, may cause some problem with android byteswap.h

acozzette:
What is the conflict between the two headers (`ByteSwap.h` and `byteswap.h`)? Do they declare conflicting symbols or something like that?

PR #6323, opened 2019-07-01
Use proto_library in proto_lang_toolchain.blacklisted_protos

cushon:
Support for using `proto_library` in this attribute was added in:
bazelbuild/bazel@a5ee2c4

Legacy support for using `.proto` files will be removed in a future
version of Bazel.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1012,7 +1012,7 @@ cc_library(
 
 proto_lang_toolchain(
     name = "cc_toolchain",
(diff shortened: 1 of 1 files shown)


BSBandme:
Would you please elaborate why change to this?

cushon:
@BSBandme there's some context in the PR description, and I cc'd you on a related bug.

BSBandme:
Got it.

PR #6324, opened 2019-07-01
Revert "Make php message class final to avoid mocking (#6277)"

TeBoring:
This reverts commit 7f84a943667358a53bd5d255ee9769ca83e75cd9.
This is just temporary. Eventually, we still want to roll forward this
change. Some users are complaining they need more time to clean up their
code.


TeBoring:
@dwsupplee

PR #6333, opened 2019-07-02
fix a UBSAN warnings.

asraa:
Fixing a UBSAN error reproduced from an OSS-Fuzz testcase

* external/com_google_protobuf/src/google/protobuf/stubs/strutil.cc:1071:9: runtime error: negation of -2147483648 cannot be represented in type 'google::protobuf::int32' (aka 'int'); cast to an unsigned type to negate this value to itself

Similar to #5901 for 32-bit integers.

Signed-off-by: Asra Ali <asraa@google.com>


acozzette:
Thanks, @asraa.

PR #6337, opened 2019-07-03
link libatomic on systems without adequate builtin atomics

boscosiu:
Some systems don't support builtin atomic intrinsics, or don't support the 8 byte variety used in [arena.cc](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/arena.cc).  In these situations [GCC falls back to a software implementation](https://gcc.gnu.org/wiki/Atomic/) provided by `libatomic` which we must link with.

For example, building on a RPi with the latest Raspbian image produces errors like `undefined reference to __atomic_fetch_add_8`.


acozzette:
Thanks, @boscosiu.

PR #6344, opened 2019-07-05
add test case for Generate() in the cpp_bootstrap_unittest.cc

wangkirin:
i found the function Generate() in cpp_generator.cc and there are no test cases to override the exception statement. i personally think that we can supplement the test cases of the corresponding scene.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
@@ -178,6 +178,19 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
   }
 }
 
+//test Generate in cpp_generator.cc
+TEST(BootstrapTest, OptionNotExist)
+{
+  cpp::CppGenerator generator;
+  DescriptorPool pool;
+  GeneratorContext *generator_context = nullptr;
+  std::string parameter = "aaa";
+  string error;
+  ASSERT_FALSE(generator.Generate(pool.FindFileByName("google/protobuf/descriptor.proto"),
+                                  parameter, generator_context, &error));
+  EXPECT_EQ(error, "Unknown generator option: " + parameter);
+}
+
 }  // namespace
 
 }  // namespace cpp


wangkirin:
Hi , @acozzette , I have already update the commit to fix the **FAIL**   Items in `kokoro` check

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -183,6 +183,69 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
   }
 }
 
+//test Generate in cpp_generator.cc
+TEST(BootstrapTest, optionNotExist) {
```
acozzette: @wangkirin Thanks for the pull request! I would say let's just keep this first test only. The second one seems to cover roughly the same case and I wouldn't worry about the 3rd one since it covers an internal-only option that we might remove in the future.

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -183,6 +183,69 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
...
+  CodeGenerator *generator=new cpp::CppGenerator();
```
acozzette: I think this line is causing a memory leak since the generator never gets deleted. You can just put the `CppGenerator` on the stack to solve that problem.

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
wangkirin: OK，I will keep the first case only

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
wangkirin: Thank you for reminding， I will fix this

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -183,6 +183,69 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
...
+  GeneratorContext *generator_context;
+  const std::string &parameter_="safe_boundary_check";
+  string error;
+  ASSERT_FALSE(generator->Generate(pool_.FindFileByName("google/protobuf/descriptor.proto"), 
+                                   parameter_, generator_context, &error));
+  EXPECT_EQ(error,"The safe_boundary_check option is not supported outside of Google.");
+}
+
+//test Generate in cpp_generator.cc
+TEST(BootstrapTest, optionReturnTrue) {
```
wangkirin: @acozzette  How about this case?  Maybe it is not duplicated with others

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
wangkirin: Done!

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
wangkirin: > @wangkirin Thanks for the pull request! I would say let's just keep this first test only. The second one seems to cover roughly the same case and I wouldn't worry about the 3rd one since it covers an internal-only option that we might remove in the future.

Done

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
acozzette: I think the options are already fairly well covered since we rely on them all the time and would notice pretty quickly if they stopped working. Also I hesitate to expand the scope of this test file too much, since its main purpose is just to check that the generated files are up to date.

wangkirin: > I think the options are already fairly well covered since we rely on them all the time and would notice pretty quickly if they stopped working. Also I hesitate to expand the scope of this test file too much, since its main purpose is just to check that the generated files are up to date.

OK, thank you for pointing out, I understand your concern. 
The commit has already updated.

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -178,6 +178,18 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
   }
 }
 
+//test Generate in cpp_generator.cc
+TEST(BootstrapTest, optionNotExist) {
```
acozzette: Capitalize the o in option: `optionNotExist` -> `OptionNotExist`

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -178,6 +178,18 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
...
+  cpp::CppGenerator generator;
+  DescriptorPool pool_;
```
acozzette: Based on our [style guide](https://google.github.io/styleguide/cppguide.html#Variable_Names) we only put an underscore at the end of a variable name if it's a class data member, so `pool_` and `parameter_` should not have the underscore.

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -178,6 +178,18 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
...
+  GeneratorContext *generator_context;
+  const std::string &parameter_="aaa";
```
acozzette: I think this should just be a `std::string` and not a const reference.

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
wangkirin: DONE

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
wangkirin: fixed
And thank you for providing the style guide :)

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -178,6 +178,18 @@ TEST(BootstrapTest, GeneratedFilesMatch) {
   }
 }
 
+//test Generate in cpp_generator.cc
+TEST(BootstrapTest, OptionNotExist) {
+  cpp::CppGenerator generator;
+  DescriptorPool pool;
+  GeneratorContext *generator_context;
```
acozzette: One more thing I forgot to mention: could you set generator_context to nullptr? I don't think C++ is guaranteed to initialize it, so we have to explicitly set it to nullptr if we want it to be defined.

wangkirin: OK ,thank you for reminding

src/google/protobuf/compiler/cpp/cpp_bootstrap_unittest.cc
```diff
@@ -60,32 +60,41 @@
...
-namespace google {
-namespace protobuf {
-namespace compiler {
-namespace cpp {
-
-namespace {
-
-class MockErrorCollector : public MultiFileErrorCollector {
- public:
+namespace google
```
acozzette: Can you undo the formatting changes? The existing formatting matches our coding style so we would prefer to keep the existing formatting.

wangkirin: Oh, I didn't notice this. I just run the `format ` in my local IDE :(
Fixed

PR #6348, opened 2019-07-08
Update version in README

elharo:
to 3.8.0


BSBandme:
Would you please also fix the [update_version.py](https://github.com/protocolbuffers/protobuf/blob/master/update_version.py#L190) to make it automatic update?
Thanks!

elharo:
My Python-fu is weak, but that aside I'm not sure you want to make this automatic. It appears the version update script runs well in advance of pushing a release--e.g. other version in the repo are now at 3.9.0-rc. The README might need to document the most current released version, not the upcoming version.

PR #6354, opened 2019-07-08
OpenJDK 7 docker file

### LINKED ISSUE #6353: Dockerfile for openjdk-7-jdk
Java

The docker file for Java build cannot run because of #6314

I'll create a Dockerfile for Java 7

suztomo:
Fixes #6353 

- [openjdk-7-jdk](https://packages.debian.org/search?keywords=openjdk-7-jdk) is available in Debian Jessie.
- Because of [Oracle Java (JDK) 8 Installer PPA (DISCONTINUED) via WebUpd8 team](https://launchpad.net/~webupd8team/+archive/ubuntu/java), not touching Oracle JDK 7 test image with this PR.

@TeBoring Would you publish the image to DockerHub?

### DIFF
--- a/kokoro/linux/dockerfile/test/java_jessie/Dockerfile
+++ b/kokoro/linux/dockerfile/test/java_jessie/Dockerfile
@@ -1,0 +1,34 @@
+FROM debian:jessie
+
+# Install dependencies.  We start with the basic ones require to build protoc
+# and the C++ build
+RUN apt-get update && apt-get install -y \
+  autoconf \
+  autotools-dev \
+  build-essential \
+  bzip2 \
+  ccache \
+  curl \
+  gcc \
+  git \
+  libc6 \
+  libc6-dbg \
+  libc6-dev \
+  libgtest-dev \
+  libtool \
+  make \
+  parallel \
+  time \
+  wget \
+  && apt-get clean
+
+# Apt source for Oracle Java 7 is no longer available in webupd8
+# https://launchpad.net/~webupd8team/+archive/ubuntu/java
+
+# Java dependencies
+RUN apt-get install -y \
+  # -- For all Java builds -- \
+  maven \
+  # -- For java_jdk7 -- \
+  openjdk-7-jdk \
+  && apt-get clean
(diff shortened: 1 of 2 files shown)


TeBoring:
The docker image has been pushed.
Seems there are some other problem: No implementation for org.codehaus.plexus.languages.java.jpms.LocationManager was bound.
  while locating org.apache.maven.plugin.surefire.SurefirePlugin

TeBoring:
BTW, the previous 64-bit dockerfile also build local maven repository.
That way helps speeds up test but may miss some coverage for pom file.
@haon4 do we want to build local maven repo in docker?

suztomo:
Looking into the error.

suztomo:
> No implementation for org.codehaus.plexus.languages.java.jpms.LocationManager

[This stackoverflow "Unable to load the mojo 'test' - org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M1:test"](https://stackoverflow.com/questions/53211941/unable-to-load-the-mojo-test-org-apache-maven-pluginsmaven-surefire-plugin) tells to upgrade Maven 3.6.0.

[Debian jessie's maven is 3.0.5](https://packages.debian.org/jessie/maven) (old).

Debian jessie's Maven cannot build the project.

@TeBoring 
I'm thinking to add 'curl https://www-eu.apache.org/dist/maven/maven-3/3.6.1/binaries/apache-maven-3.6.1-bin.tar.gz' etc. in Dockerfile. What do you think?

TeBoring:
Can we use older version of mojo?

suztomo:
It's technically feasible to use older version of Maven plugins to resolve this surefire plugin problem. However, in general, I think you don't want the deprecated Java 7 build environment to prevent catching up latest versions of plugins.

I believe the compatibility test is to confirm the binary compatibility of the resulting protobuf-java classes with Java 7 runtime, not for Kokoro's build environment (Debian Jessie and Maven 3.0.5).

Having said that, I don't own the repository. If you think downgrading is a good option, I'm happy to help downgrading the problematic mojo.

TeBoring:
I have worries that if we downgrade mojo, there will be problem for java_strecth, which may not provide older version of mojo.
@haon4, do we still want to support java7?

elharo:
Suspect this is obsolete. @suztomo?

suztomo:
Right. I don’t need this (Java 7) any more.

PR #6370, opened 2019-07-11
Fix null terminated string in Ruby wrapper type helper names

jbolinger:
Fix for an error with the string terminator placement in #5739.

Thanks for pointing it out @tenderlove!

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -192,7 +192,7 @@ static int extract_method_call(VALUE method_name, MessageHeader* self,
     // Find the field name
     char wrapper_field_name[name_len - 8];
     strncpy(wrapper_field_name, name, name_len - 9);
-    wrapper_field_name[name_len - 7] = '\0';
+    wrapper_field_name[name_len - 9] = '\0';
 
     // Check if field exists and is a wrapper type
     const upb_oneofdef* test_o_wrapper;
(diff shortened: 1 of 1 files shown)


TeBoring:
Is there any test for avoiding regression?

jbolinger:
No, I'm not sure what kind of test have in mind for this?

TeBoring:
Why previously this hasn't broken anything?

ruby/ext/google/protobuf_c/message.c
```diff
@@ -192,7 +192,7 @@ static int extract_method_call(VALUE method_name, MessageHeader* self,
     // Find the field name
     char wrapper_field_name[name_len - 8];
     strncpy(wrapper_field_name, name, name_len - 9);
-    wrapper_field_name[name_len - 7] = '\0';
+    wrapper_field_name[name_len - 9] = '\0';
```
TeBoring: https://github.com/protocolbuffers/protobuf/blob/8041dd4034dc837f14c92846acd669e4df01cae2/ruby/ext/google/protobuf_c/message.c#L295

Previously, it should be interpreted as METHOD_UNKNOWN. Then, the value will be set to the parent ruby object.
I guess, if you actually encode the message then, the value you set won't appear in the encoding result.

jbolinger:
The problem here is writing the `\0` at `wrapper_field_name[name_len - 7]` is out of bounds for the variable declared as `char wrapper_field_name[name_len - 8];` True, it doesn't seem to have caused problems in the past for whatever reason though.

TeBoring:
I see. Thought it was writing \0 before the last character.

PR #6371, opened 2019-07-12
Disable javadoc error

TeBoring:
Actual fixes of the javadoc will be followed up


haon4:
can you also make the maven-javadoc-plugin version the same?

PR #6374, opened 2019-07-12
Add comparison of Timestamp

Oshiumi:
Support comparison of `Google::Protobuf::Timestamp` in Ruby

For example, you write as follows below if you'd like to sort an array of message.

```ruby
messages.sort_by { |message| message.time.seconds }
```

You can write as follows if merge this feature.

```ruby
messages.sort_by(&:time)
```

### DIFF
--- a/ruby/lib/google/protobuf/well_known_types.rb
+++ b/ruby/lib/google/protobuf/well_known_types.rb
@@ -72,6 +72,8 @@ module Google
     end
 
     Timestamp.class_eval do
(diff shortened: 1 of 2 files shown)


TeBoring:
Would you mind creating a simple google doc explain the proposal?

CavHack:
I like this but it's just not practical for most distributed systems since there are no global clocks. Perhaps you're thinking of vector clocks.

Issue #6376, opened 2019-07-12
Python Struct value assignment raises ValueError


leon-barrett:
**What version of protobuf and what language are you using?**
Version: master
Language: Python

**What operating system (Linux, Windows, ...) and version?**
Linux
Ubuntu 18.04.2 LTS

**What runtime / compiler are you using (e.g., python version or gcc version)**
Python 3.7.0
(I see the same problem on Python 3.6.6 as well.)

**What did you do?**
Currently, if you access a ListValue or Struct from a Struct and attempt
to assign it into another Struct, you get an exception:

```python
> import google.protobuf.struct_pb2 as spb
> s1 = spb.Struct()
> s1['a'] = [1]
> s2 = spb.Struct()
> s2['a'] = s1['a']
ValueError: Unexpected type
> s1['b'] = {'x': 'y'}
> s2['b'] = s1['b']
ValueError: Unexpected type
```

**What did you expect to see**
I expected to be able to assign a value I got from one Struct into another Struct.

**What did you see instead?**
```
ValueError                                Traceback (most recent call last)
<ipython-input-12-8395286e45ea> in <module>
----> 1 s2['c'] = s1['c']

python/google/protobuf/internal/well_known_types.py in __setitem__(self, key, value)
    763
    764   def __setitem__(self, key, value):
--> 765     _SetStructValue(self.fields[key], value)
    766
    767   def __delitem__(self, key):

python/google/protobuf/internal/well_known_types.py in _SetStructValue(struct_value, value)
    730     struct_value.list_value.extend(value)
    731   else:
--> 732     raise ValueError('Unexpected type')
    733
    734

ValueError: Unexpected type
```

**Anything else we should know about your project / environment**
I have a fix and will send a pull request soon.

PR #6377, opened 2019-07-12
Fix assignment between Python protobuf Structs

leon-barrett:
This fixes bug https://github.com/protocolbuffers/protobuf/issues/6376

Currently, if you access a ListValue or Struct from a Struct and attempt
to assign it into another Struct, you get an exception:

```python
> import google.protobuf.struct_pb2 as spb
> s1 = spb.Struct()
> s1['a'] = [1]
> s2 = spb.Struct()
> s2['a'] = s1['a']
ValueError: Unexpected type
> s1['b'] = {'x': 'y'}
> s2['b'] = s1['b']
ValueError: Unexpected type
```

This fixes that case.

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -722,10 +722,10 @@ def _SetStructValue(struct_value, value):
     struct_value.string_value = value
   elif isinstance(value, _INT_OR_FLOAT):
     struct_value.number_value = value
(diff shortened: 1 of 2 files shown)


leon-barrett:
It looks like the "Mergeable" test fails because this PR should have the label "python", but it looks like I don't have the needed permissions to add that label myself.

anandolee:
Thanks for the PR

Issue #6379, opened 2019-07-13
Binary compatibility broken in Google.Protobuf by introduction of Proto2 features


jskeet:
This may be the case for other types as well, but *at least* FieldCodec introduced binary-incompatible changes in 3.9.0, in [this commit](https://github.com/protocolbuffers/protobuf/commit/9e89b6e70a4988a4b473a19a7647154bd07583ce#diff-ead1b5f72e34b49fa51ea24edd259428).

Each FieldCodec static factory method gained a default parameter:

    public static FieldCodec<string> ForString(uint tag)

became

    public static FieldCodec<string> ForString(uint tag, string defaultValue = "")

This is a binary breaking change. Code compiled against an earlier version of Google.Protobuf fails at execution time when using the new version, as it can't find the method. We should have added overloads instead of default parameters here.

ObsidianMinor:
This was my mistake, I'll fix it. To be 100% clear, reintroducing the previous method will make the next release binary-compatible, right?

jskeet:
I haven't looked over all the changes - I only found that one. Ideally, you should take a copy of the public API at an earlier version, then take a copy of the public API at a later version, and diff them. Introducing a new overload and then making the currently-optional parameter required will change the compile-time behaviour (in that it will resolve to the single-parameter method rather than the two-parameter method) but that's unlikely to break anyone.

The public API generator tool at https://github.com/JakeGinnivan/ApiApprover could help with this.

ObsidianMinor:
Alright I ran the tool and found the FieldCodec factories are the only breaking API changes (beyond methods introduced on interfaces that should only be implemented by internal types).

jskeet:
As a thought, I wonder whether we should delist 3.9.0 from NuGet for now, then release 3.9.1 with the fix when it's ready to release. That will reduce the chances of other people accidentally taking a dependency on 3.9.0 which would cause diamond dependency issues.

anandolee:
Delisted. Thanks

Sorry for late reply, just back from vacation

anandolee:
After a second thought, I have a question for the compatibility. Is the compatibility break that new generated code can not compile with old C# runtime?  Protobuf compatibility does not gurantee such case. We only consider old generated code not work with new runtime as breaking change

ObsidianMinor:
[...]

PR #5936, opened 2019-03-22
C# Proto2 feature : Finale

ObsidianMinor:
This wraps up all the previous feature PRs. Changes include:

 * Many tests for previous feature PRs
 * The proto3 constraint on C# code-gen has been removed
 * Conformance tests for proto2 are ran instead of skipped
 * GetOption APIs have been added to descriptors
 * CustomOptions are now deprecated
 * The generated descriptor code has been made public

When pulled, this PR will fix numerous proto2 feature issues, including: #5759, #4591, #4407, #3487, #2991

PR #6380, opened 2019-07-13
Fix binary compatibility in FieldCodec factory methods

ObsidianMinor:
Fix for #6379.

### DIFF
--- a/csharp/src/Google.Protobuf/FieldCodec.cs
+++ b/csharp/src/Google.Protobuf/FieldCodec.cs
@@ -219,7 +219,7 @@ namespace Google.Protobuf
         /// <param name="tag">The tag.</param>
         /// <param name="defaultValue">The default value.</param>
         /// <returns>A codec for the given tag.</returns>
-        public static FieldCodec<string> ForString(uint tag, string defaultValue = "")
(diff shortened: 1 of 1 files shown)


jskeet:
I'm confused about why we have these defaultValue parameters at all - they never seem to be used. My guess is that that's a separate bug, that the existing methods aren't calling the right constructor. (That should be confirmed with a breaking test before being fixed, and that should probably be in a different PR.)

I would make the overloads delegate to the methods accepting a default value though, unless there's a good reason not to.

ObsidianMinor:
> that should probably be in a different PR

Right. I was going to do it in this one but decided to fix it in #5936 when the different default values would be put to use (with extensions being finalized then).

TeBoring:
Why previously there were no test to catch it?

ObsidianMinor:
There was no code to test against. There was no way to generate extension identifiers at the time (before #5936) so rather than add tests for code that couldn't actually be used (unless the consumer decided to implement the interfaces and use the API manually) I decided to wait until the last PR. Then the code could be fully tested properly when unittest.proto could actually be generated in the test project.

jtattermusch:
Fine to merge (and backport to v3.9.x branch)?

PR #6384, opened 2019-07-13
Add option to print enums as ints in Ruby JSON encoder

### LINKED ISSUE #6178: Ruby JSON serialize enums as ints
Ruby proto3

The proto3 docs says that a JSON implementation may provide an option to:
> Emit enum values as integers instead of strings: The name of an enum value is used by default in JSON output. An option may be provided to use the numeric value of the enum value instead.

Currently this isn't implemented by the Ruby library. 

As far as I know of this feature is implemented by the [C++](https://github.com/protocolbuffers/protobuf/issues/2735) and [Go](https://godoc.org/github.com/golang/protobuf/jsonpb#Marshaler) libraries. I would like to request this so that Ruby has feature parity with the other supported languages.

**Additional context**
I believe the entry point to Ruby's encode_json method is here:
https://github.com/protocolbuffers/protobuf/blob/f425b9f059a57dc5f13a6e6a389fff1a6cc06749/ruby/ext/google/protobuf_c/encode_decode.c#L1477

r6min:
Fixes #6178

Adds the option to print enums as int values instead of the default strings for Ruby (similar to the option present for C++).

Example:

test.proto
```
enum TestEnum {
  Default = 0;
  Foo = 1;
}
enum TestMessage {
  TestEnum test_enum = 1;
}
```
code.rb
```
mymessage = TestMessage.new(:test_enum => 1)
puts TestMessage.encode(mymessage, {always_show_enums_as_ints: true})
```
will output:
```
{"testEnum":1}
```

### DIFF
--- a/ruby/ext/google/protobuf_c/encode_decode.c
+++ b/ruby/ext/google/protobuf_c/encode_decode.c
@@ -1151,7 +1151,8 @@ static void putmap(VALUE map, const upb_fielddef *f, upb_sink *sink,
 }
 
 static const upb_handlers* msgdef_json_serialize_handlers(
-    Descriptor* desc, bool preserve_proto_fieldnames);
+    Descriptor* desc, bool preserve_proto_fieldnames, 
+    bool always_show_enums_as_ints);
 
 static void putjsonany(VALUE msg_rb, const Descriptor* desc,
                        upb_sink* sink, int depth, bool emit_defaults) {
@@ -1227,7 +1228,7 @@ static void putjsonany(VALUE msg_rb, const Descriptor* desc,
       }
 
       subsink.handlers =
-          msgdef_json_serialize_handlers(payload_desc, true);
+          msgdef_json_serialize_handlers(payload_desc, true, true);
       subsink.closure = sink->closure;
       putmsg(payload_msg_rb, payload_desc, &subsink, depth, emit_defaults, true,
              is_wellknown);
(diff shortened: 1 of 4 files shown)


TeBoring:
Please send changes of upb to https://github.com/protocolbuffers/upb first.

TeBoring:
Any update?

r6min:
@TeBoring yes, it seems like this repo (the ruby sections) have deprecated interfaces compared to the upb repo, mentioned [here](https://github.com/protocolbuffers/upb#user-content-old-handlers-interfaces), doesn't seem like the changes here can be added to the upb repo (correct me if i'm wrong).

TeBoring:
There is some ongoing work to update upb in ruby: https://github.com/protocolbuffers/protobuf/pull/5866

Please send change to upb first. Once you are done, we probably have updated the upb in ruby.

TeBoring:
We have update upb in ruby. Could you try again?

acozzette:
@spiffyyeng I'm going to close this since there has been no activity in a long time, but feel free to comment here if you would still like to pursue this change.

danielmorrison:
I'd personally love to see it, but don't know the internals enough yet to attempt it myself.

haberman:
This would be an easy change.

We already support two options to the JSON encoder: https://github.com/protocolbuffers/protobuf/blob/7309e8882cb81a92cee4b0be708a61c8f4707224/ruby/ext/google/protobuf_c/message.c#L1146-L1155

We would just need to add support for the `upb_JsonEncode_FormatEnumsAsIntegers` option: https://github.com/protocolbuffers/upb/blob/ada907989e075658b72a6dcb3ec8335b43596cbc/upb/json_encode.h#L48-L49

akuita:
hi @haberman, I created a pull request based on your suggestion. Would love to have your feedback https://github.com/protocolbuffers/protobuf/pull/11673

MQuy:
hi @haberman, I created a pull request based on your suggestion. Would love to have your feedback https://github.com/protocolbuffers/protobuf/pull/11673

PR #6386, opened 2019-07-15
3.5.x

salugoju:
Could you please point me to the version that  builds with C99 and C03


acozzette:
The 3.5.x branch should still build with C++03.

Issue #4500, opened 2018-04-12
cant deserialize Map type with key =  0 in NodeJS


jet6666:
`syntax = "proto3" ;
  message MapTest {
		map<int32 ,MapTestItem> test=1 ; 
  }
  
  message MapTestItem {
  	int32 hp=1;
  	int32 hp2=2;
  }`

**Server-side：PHP (protobuf 3.5.0~3.5.1)**
`$p = new MapTest();
$index = 0;　// Index=0 

$item = new MapTestItem();
$item->setHp(99);
 
$p->setToChanged( array($index =>$item) );
$packed=$p->serializeToString();
  
echo "byteLen=",strlen($packed),PHP_EOL;
for($i=0;$i< strlen($packed);$i++) {
  echo bin2hex($packed{$i}),PHP_EOL;
}`

the result goes to :
byteLen=6
0a
04
12
02
08
63

---------------------
But cant deserializeBinary in JS:

`var buffBytes=Buffer.alloc(6);
buffBytes[0] =58;
buffBytes[1] =4;
buffBytes[2] =18;
buffBytes[3] =2;
buffBytes[4] =8;
buffBytes[5] =99;
 function buff2uint(buffBytes) {
	var a = new Uint8Array(buffBytes.length);
	for (var i = 0; i < buffBytes.length; i++) a[i] = buffBytes[i];
		return a ;
}
bytes = buff2uint(buffBytes);
var message2 = proto.MapTest.deserializeBinary(bytes);
`
 
**throw an error:**
`C:\Users\jet\AppData\Roaming\npm\node_modules\google-protobuf\google-protobuf.js:98
[...]

Issue #3351, opened 2017-07-11
php/js - map<string,string> can't decode value of ""


redaniel22:
php 7.0
google/protobuf installed via composer and is version 3.3.2
protobuf code generation done with version 3.3

protobuf
```protobuf
message example {
    map<string,string> custom_fields = 1;
}
```

php
```php
    $example->getCustomFields()["my_key"] = "";
```

Encoding this via php and decoding (uint8) via javascript causes an assertion error in javascript. Making sure there are no empty strings as values in the map resolves the assertion error.

TeBoring:
Could you try master? We had a lot of encoding fix recently.

redaniel22:
Unfortunately, neither 3.3.x-dev or dev-master 9ab7c73 on the php side have resolved the issue.

Still getting assertion error:

```
protobuf.b3a75a6054c5aefb7719.js:1 Uncaught goog.asserts.AssertionError {message: "Assertion failed", reportErrorToServer: true, messagePattern: "Assertion failed", stack: "AssertionError↵    at new goog.asserts.AssertionEr…com:8443/dist/app.6b0000b3cbadadfd39a8.js:1:8348)"}
goog.asserts.DEFAULT_ERROR_HANDLER @ protobuf.b3a75a6054c5aefb7719.js:1
goog.asserts.doAssertFailure_ @ protobuf.b3a75a6054c5aefb7719.js:1
goog.asserts.assert @ protobuf.b3a75a6054c5aefb7719.js:1
jspb.Map.deserializeBinary @ protobuf.b3a75a6054c5aefb7719.js:1
(anonymous) @ protobuf.b3a75a6054c5aefb7719.js:1
jspb.BinaryReader.readMessage @ protobuf.b3a75a6054c5aefb7719.js:1
proto.Proto.Marketplace.EventView.deserializeBinaryFromReader @ protobuf.b3a75a6054c5aefb7719.js:1
proto.Proto.Marketplace.EventView.deserializeBinary @ protobuf.b3a75a6054c5aefb7719.js:1
convert @ app.6b0000b3cbadadfd39a8.js:1
handleReadyStateChange @ app.6b0000b3cbadadfd39a8.js:1
```

Additionally, on dev-master, if I set the value to a non-empty string, it decodes without any errors, but the map comes through empty. A separate int64 field did decode correctly when the map had a non-empty string. This may be due to a notice I had to suppress to get things working properly on dev-master which was:

```
[...]

PR #6394, opened 2019-07-17
Fixed JS parsing of default map values

chemhack:
We recently recently detected JS protobuf library fail to deserialize any map<uint64,double> data where value is 0. 

There has been many cases being reported:

https://github.com/protocolbuffers/protobuf/issues/3351 
https://github.com/grpc/grpc-web/issues/533
https://github.com/protocolbuffers/protobuf/issues/4500

In binary format, default map values are not sent, however this is not probably handled in JS library. 

This fix is inspired by:

https://github.com/protocolbuffers/protobuf/pull/4687

### DIFF
--- a/js/map.js
+++ b/js/map.js
@@ -465,11 +465,15 @@ jspb.Map.prototype.serializeBinary = function(
  *    entries with unset keys is required for maps to be backwards compatible
  *    with the repeated message representation described here: goo.gl/zuoLAC
  *
+ * @param {V=} opt_defaultValue
+ *    The default value for the type of map values.
+ *
  */
 jspb.Map.deserializeBinary = function(map, reader, keyReaderFn, valueReaderFn,
-                                      opt_valueReaderCallback, opt_defaultKey) {
+                                      opt_valueReaderCallback, opt_defaultKey,
+                                      opt_defaultValue) {
   var key = opt_defaultKey;
-  var value = undefined;
+  var value = opt_defaultValue;
 
   while (reader.nextField()) {
     if (reader.isEndGroup()) {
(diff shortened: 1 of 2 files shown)


chemhack:
I signed it!

klaemo:
I signed it!

klaemo:
I consent.

chemhack:
@TeBoring the CLA bot doesn't seem to work reliably, could you please check this manually? thank you

TeBoring:
Is it possible to resend a PR with only one person to commit?

chemhack:
@TeBoring changed the commit

RXminuS:
Any indication when this could be merged, it's been a bug with far-reaching consequences in the community (grpc and a few other projects as well). And thank you @chemhack for creating a fix!

TeBoring:
Could you also add some test into maps_test.js?

chemhack:
@TeBoring it's actually already covered by the JS unit test, but the unit test doesn't catch the bug since the binary message is serialized by JS library as well, which is behaving differently than other language libs. 

IMHO the test should belong to conformance test rather in maps_test.js. But i only have very limited knowledge of the codebase and won't have time to dig into the test setup. Hope you can understand that.

TeBoring:
Ah, I see. Thank you for the fix.

TeBoring:
When the value of map is message, null is used as the fault, which could cause the problem. Could you add a test and fix the problem?

chemhack:
> When the value of map is message, null is used as the fault, which could cause the problem. Could you add a test and fix the problem?

What exact problem were you experiencing? If a value is a message, and not specified, null is the correct default. Are you expecting it to be an empty message instead?

Also, when the key of map is a message, that's exactly the same behavior.

TeBoring:
When value is message, jspb.BinaryReader.prototype.readMessage complains it cannot accept null.
Also, message cannot be key.

Issue #6395, opened 2019-07-17
AttributeError: 'Timestamp' object has no attribute 'seconds' (well_known_types.py, introduced v3.9.0)


dhendry:
**What version of protobuf and what language are you using?**
Version: v3.9.0
Language: Python

**What operating system (Linux, Windows, ...) and version?**

Mac 10.14.5

**What runtime / compiler are you using (e.g., python version or gcc version)**

Python 3.7.4

**What did you do?**

This code worked as of protobuf v3.8.0:

```
from datetime import datetime
from google.protobuf.internal.well_known_types import Timestamp

# Datetime str formatter used by google proto Timestamp library
DATETIME_FMT = "%Y-%m-%dT%H:%M:%S.%fZ"

def to_protobuf_timestamp_str(date):
    # The Timestamp does specialized padding on the microseconds when
    # serializeing to strings, so use it in order to match test values
    timestamp = Timestamp()
    timestamp.FromJsonString(date.strftime(DATETIME_FMT))
    return timestamp.ToJsonString()

print(to_protobuf_timestamp_str(datetime.now()))
```

**What did you expect to see**

Printed timestamp, ex: `2019-07-17T09:56:56.601774Z`

**What did you see instead?**

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 5, in to_protobuf_timestamp_str
  File "/Users/.../.local/share/virtualenvs/..../lib/python3.7/site-packages/google/protobuf/internal/well_known_types.py", line 190, in FromJsonString
    self.seconds = int(seconds)
AttributeError: 'Timestamp' object has no attribute 'seconds'
```

**Anything else we should know about your project / environment**

* After doing some digging it seems like this problem was caused by the introduction for the `__slots__` attribute (https://github.com/protocolbuffers/protobuf/commit/a1c886834425abb64a966231dd2c9dd84fb289b3#diff-79a29d050e84ff410293a63e1a1551d6R105)
* It seems like I can fix our above code by changing 
  * `from google.protobuf.internal.well_known_types import Timestamp` 
  * to `from google.protobuf.timestamp_pb2 import Timestamp`
* Arguably this is a bug in our code, **however I believe the protobuf bug here is the fact its not at all clear the classes in `well_known_types.py` are actually mixins**, this code worked before and no longer does now, and its not clear what the right thing to do is.
* At a minimum, I suggest: 
[...]

PR #6396, opened 2019-07-17
Commenting only: well_known_types.py contains mixins, not classes which should be used directly

dhendry:
This contains the minim set of my recommendations from https://github.com/protocolbuffers/protobuf/issues/6395 to help users understand that the classes in `well_known_types.py` are not intended for direct use and are in fact mixin classes.

A better (but possibly more disruptive) approach would be to explicitly rename these classes with `Mixin` (ex: `TimestampMixin`) to avoid confusion and prevent identically named classes, just in different packages.

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -65,7 +65,10 @@ _DURATION_SECONDS_MAX = 315576000000
 
 
 class Any(object):
-  """Class for Any Message type."""
+  """Mixin class for Any message type.
+  NOTE: This class is not intended to be used directly.
+  These methods get added to google.protobuf.any_pb2.Any type.
+  """
 
   __slots__ = ()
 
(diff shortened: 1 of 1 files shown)


anandolee:
Hi Dan, all modules under google.protobuf.internal are not supposed to be used directly by users. I don't think it is necessary to change those methods names and add comments for each method.

We also have documented the well known types usages and other generated code in: https://developers.google.com/protocol-buffers/docs/reference/python-generated

dhendry:
@anandolee Feels like adding these comments is an extremely simple way to help others avoid the pain I went through after updating to the most recent protobuf version. Due to the way python imports work, it was not clear from reading our code which broke we were importing the "wrong" thing. I find it unexpected and not user friendly that a library would have identically named classes and not highlight the fact one is not intended to be used directly. 

This problem, and my confusion, was exacerbated by the fact protobuf classes are generated dynamically (an unused pattern in our internal codebases, and a generally uncommon pattern in many popular python libraries). Its entirely unclear when looking at `timestamp_pb2.py` that additional methods get mixed into the class (particularly when this capability is explicitly unavailable for user generated messages).

anandolee:
How about add some comments on the top of file? Give some usage examples and guide people to https://developers.google.com/protocol-buffers/docs/reference/python-generated#wkt

PR #6404, opened 2019-07-18
Add podspec for C++

rmstar:


muxi:
@thomasvl - PTAL

Protobuf-cc.podspec
```diff
@@ -0,0 +1,107 @@
...
+                    'src/google/protobuf/util/internal/protostream_objectwriter_test.cc'
+
+  s.header_mappings_dir = 'src'
+  s.libraries = 'c++'
+
+  s.ios.deployment_target = '7.0'
+  s.osx.deployment_target = '10.9'
+  s.tvos.deployment_target = '9.0'
+  s.watchos.deployment_target = '2.0'
+  s.requires_arc = false
```
thomasvl: Why is this needed since it is all C++ code?

Protobuf-cc.podspec
```diff
@@ -0,0 +1,107 @@
...
+
+  s.pod_target_xcconfig = {
+    # Do not let src/google/protobuf/stubs/time.h override system API
+    'USE_HEADERMAP' => 'NO',
+    'ALWAYS_SEARCH_USER_PATHS' => 'NO',
+
+    # Configure tool is not being used for Xcode. When building, assume pthread is supported.
+    'GCC_PREPROCESSOR_DEFINITIONS' => '"$(inherited)" "COCOAPODS=1" "HAVE_PTHREAD=1"',
```
thomasvl: What is `"COCOAPODS=1"` needed for?

Is `"HAVE_PTHREAD=1"` needed for should a `configure` have already taken care of things?

Protobuf-cc.podspec
```diff
@@ -0,0 +1,107 @@
...
+                    'src/google/protobuf/util/message_differencer_unittest.cc',
+                    'src/google/protobuf/util/internal/json_stream_parser_test.cc',
+                    'src/google/protobuf/util/internal/default_value_objectwriter_test.cc',
+                    'src/google/protobuf/util/internal/json_objectwriter_test.cc',
+                    'src/google/protobuf/util/internal/protostream_objectsource_test.cc',
+                    'src/google/protobuf/util/internal/protostream_objectwriter_test.cc',
+                    'src/google/protobuf/util/internal/protostream_objectwriter_test.cc'
+
+  s.header_mappings_dir = 'src'
+  s.libraries = 'c++'
```
thomasvl: Is this needed or does Xcode just get things right when things get added to the projects/workspaces?

php/ext/google/protobuf/package.xml
```diff
@@ -10,8 +10,8 @@
   <email>protobuf-opensource@google.com</email>
   <active>yes</active>
  </lead>
- <date>2019-06-17</date>
- <time>09:34:50</time>
+ <date>2019-07-17</date>
```
thomasvl: Why is this file changing, seems unrelated?  If the one on _master_ is stale, I'd suggest doing a separate PR to fix that.

thomasvl:
@muxi - The protobuf team will also have to agree to maintain/updating this like they do the objc one with releases.

Protobuf-cc.podspec
```diff
@@ -0,0 +1,107 @@
...
+               :tag => "v#{s.version}" }
+
+  s.source_files = 'src/google/protobuf/*.{h,cc,inc}',
+                   'src/google/protobuf/stubs/*.{h,cc}',
+                   'src/google/protobuf/io/*.{h,cc}',
+                   'src/google/protobuf/util/*.{h,cc}',
+                   'src/google/protobuf/util/internal/*.{h,cc}'
+
+  # Excluding all the tests in the directories above
+  s.exclude_files = 'src/google/protobuf/any_test.cc',
```
thomasvl: Since `excludes_files` can also use globs, I wonder if it would be safer to do something like:

```
s.exclude_files = 'src/google/**/*_test.{h,cc,inc}',
                  'src/google/**/*_unittest.{h,cc}',
                  'src/google/protobuf/test_util*.{h,cc}',
```

That way it is less likely to accidentally include something that gets added in the future.

muxi:
Sounds good. We will ask them about it. Thanks!

Protobuf-cc.podspec
rmstar: Looks like it's not required, will remove.

Protobuf-cc.podspec
rmstar: Not required, removing

Protobuf-cc.podspec
rmstar: `"COCOAPODS=1"` is not needed, I'll remove it. `"HAVE_PTHREAD=1"` is required because `configure` is not run when we install using cocoapods.

php/ext/google/protobuf/package.xml
rmstar: This file changed when I ran `update_version.py` to verify that the version in `Protobuf-cc.podspec` is updated automatically. I guess this change shouldn't checked in - maybe it should only be run when someone cuts cuts a release?
I'll revert this change.

Protobuf-cc.podspec
rmstar: Yes, this seems safer than explicitly specifying `exclude_files`, but unfortunately it doesn't work.

Compiling src/google/protobuf/map_lite_test_util.h fails as it depends on codegen for a test proto.
`'google/protobuf/map_lite_unittest.pb.h' file not found`

thomasvl: Just add that to the exclusion:

```
s.exclude_files = 'src/google/**/*_test.{h,cc,inc}',
                  'src/google/**/*_unittest.{h,cc}',
                  'src/google/protobuf/test_util*.{h,cc}',
                  'src/google/protobuf/map_lite_test_util.{h,cc}'
```

thomasvl: You might need to do the same thing for `map_test_util*.{h,cc}`

rmstar: Thanks! fixed.

thomasvl:
@muxi this seems fine to me, so you'll need to get someone on the protobuf team for final sign off on name (i.e. - _Protobuf-cc_ or something else) and also that why want to maintain/publish this with releases.

muxi:
@thomasvl - we will get this done. Thank you for the help again Thomas.

update_version.py
```diff
@@ -250,6 +250,11 @@ def UpdateObjectiveC():
       r"^  s.version  = '.*'$",
       "  s.version  = '%s'" % GetFullVersion(rc_suffix = '-rc'),
       line))
+  RewriteTextFile('Protobuf-cc.podspec',
```
TeBoring: cc to c++

rmstar: Thanks for catching that. Fixed.

shali3:
Hi there,
I'm posting this here for others that try to use it.
I tried using it with:
```
  pod 'Protobuf-C++', :git => 'https://github.com/protocolbuffers/protobuf.git', :tag => 'v3.11.2'
```
and the Protobuf-C++ target doesn't compile.
Looks like a header path issue (I have a bunch of those):
```
...../Pods/Protobuf-C++/src/google/protobuf/any.cc:31:10: fatal error: 'google/protobuf/any.h' file not found
#include <google/protobuf/any.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
1 warning and 1 error generated.
```

After digging around looks like this pod does not support the `use_frameworks!` configuration.
After switching this off it looks like it's working :)

I have 2 questions for you:
1. Are you planning to publish this package to cocoapods anytime soon?
2. Are you planning to support `useframeworks!` option?

Regarding question 2, I managed to compile it by manually adding this search path to `Protobuf-C++` pod target.
```
${PODS_TARGET_SRCROOT}/src
```
You can do it in the podspec file like so:
```
  spec.pod_target_xcconfig = {
    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}/src"'
  }
```

But it crashes on startup with the following error:
```
dyld: Library not loaded: @rpath/Protobuf_C_.framework/Protobuf_C_
  Referenced from: /Users/myuser/Library/Developer/CoreSimulator/Devices/BC55661B-9EC5-4EF1-A071-BC69ABA420AE/data/Containers/Bundle/Application/DBFE7F24-6521-4639-A0FD-E14B07DEB389/ios-sample.app/ios-sample
  Reason: image not found
```
You can clone [this repo](https://github.com/shali3/ios-single-view-app) to reproduce the issue:
```
git clone https://github.com/shali3/ios-single-view-app.git
```

shali3:
This is a workaround you can do on your Podfile:
```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      if target.name == 'Protobuf-C++'
        # Pure C++ doesn't support framework imports. We need to help it find it's path.
        target.build_configurations.each do |config|
          config.build_settings['HEADER_SEARCH_PATHS'] = "${PODS_TARGET_SRCROOT}/src";
        end
      end
    end
  end
end
```

PR #6407, opened 2019-07-18
Add csharp compatibility tests between last released and the current version

anandolee:
#6379

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -148,8 +148,9 @@ build_csharp() {
   # Run csharp compatibility test between 3.0.0 and the current version.
   csharp/compatibility_tests/v3.0.0/test.sh 3.0.0
 
-  # Run csharp compatibility test between 3.7.0 and the curretn version.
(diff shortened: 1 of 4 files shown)


tests.sh
```diff
@@ -147,6 +147,9 @@ build_csharp() {
 
   # Run csharp compatibility test between 3.0.0 and the current version.
   csharp/compatibility_tests/v3.0.0/test.sh 3.0.0
+
+  # Run csharp compatibility test between 3.7.0 and the curretn version.
+  csharp/compatibility_tests/v3.0.0/test.sh 3.7.0
```
TeBoring: How about just update 3.0.0 on line 149 to 3.7.0?

anandolee: I'd prefer to keep two compatibility tests: one between a fixed very old version with current and another is last major version with current.

Have changed C# to use last major version instead of 3.7.0 here. Will try to change other languages later

PR #6409, opened 2019-07-18
Down integrate to GitHub

TeBoring:


src/google/protobuf/compiler/objectivec/objectivec_enum.h
```diff
@@ -28,6 +28,36 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+// Protocol Buffers - Google's data interchange format
```
acozzette: It seems like something went wrong here because the copyright notice is duplicated.

src/google/protobuf/compiler/objectivec/objectivec_enum.cc
```diff
@@ -28,13 +28,47 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+// Protocol Buffers - Google's data interchange format
```
thomasvl: Looks like some automation went wrong here

src/google/protobuf/compiler/objectivec/objectivec_enum.cc
```diff
@@ -28,13 +28,47 @@
...
 #include <string>
 
 #include <google/protobuf/compiler/objectivec/objectivec_enum.h>
 #include <google/protobuf/compiler/objectivec/objectivec_helpers.h>
 #include <google/protobuf/io/printer.h>
 #include <google/protobuf/stubs/strutil.h>
+
+
+
+
```
thomasvl: same

src/google/protobuf/compiler/objectivec/objectivec_enum.h
haon4: These files shouldn't be down integrated since their SOT is GitHub. I think we just need to update our file_list.bzl file a little. Will send out a fix for that. Then you will need to close this and run git_sync again.

src/google/protobuf/compiler/objectivec/objectivec_enum.cc
thomasvl: Oops, missed the other comments about the same thing.

PR #6411, opened 2019-07-19
Fix incorrect field type in python JsonObjects

aletheap:
google.protobuf.json_format.MessageToDict incorrectly casts ints to strs. This fixes that.

### DIFF
--- a/python/google/protobuf/json_format.py
+++ b/python/google/protobuf/json_format.py
@@ -292,7 +292,7 @@ class _Printer(object):
     elif field.cpp_type == descriptor.FieldDescriptor.CPPTYPE_BOOL:
       return bool(value)
     elif field.cpp_type in _INT64_TYPES:
-      return str(value)
+      return int(value)
     elif field.cpp_type in _FLOAT_TYPES:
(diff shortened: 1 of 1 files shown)


aletheap:
I can't figure out how to add a label to this PR, but it's for Python

python/google/protobuf/json_format.py
```diff
@@ -292,7 +292,7 @@ def _FieldToJsonObject(self, field, value):
     elif field.cpp_type == descriptor.FieldDescriptor.CPPTYPE_BOOL:
       return bool(value)
     elif field.cpp_type in _INT64_TYPES:
-      return str(value)
```
anandolee: https://developers.google.com/protocol-buffers/docs/proto3?hl=zh-CN#json

From our json mapping, int63 is present by string in Josn

aletheap: I see. Is the idea that some languages don't support 64bit numbers but all support strings?

aletheap: I see. I guess JSON doesn't support 64 bit ints?

PR #6412, opened 2019-07-19
Down integrate to GitHub

TeBoring:


acozzette:
Do you know why `csharp/src/Google.Protobuf.Test/testprotos.pb` is being modified? It seems like that file should not be getting changed by the sync.

src/google/protobuf/compiler/csharp/csharp_bootstrap_unittest.cc
```diff
@@ -46,6 +46,10 @@
 #include <google/protobuf/stubs/map_util.h>
 #include <google/protobuf/stubs/stl_util.h>
 #include <google/protobuf/stubs/strutil.h>
```
acozzette: I think we are getting blank lines here because the extraction script is removing internal-only headers. Do we actually need to update these C# files? If GitHub is the source-of-truth for them then maybe we don't even need to update them during the down-integrations.

TeBoring: Haven't figured out how to avoid syncing these files yet.
But the extract script has been fixed. Seems it can still help to replace some methods. e.g, Split.

PR #6431, opened 2019-07-23
Backport review comments for C# proto2 features (for 3.9.x)

jtattermusch:
- backporting some changes suggested in https://github.com/protocolbuffers/protobuf/pull/5350/files review.

Most importantly, the new extension APIs are marked as experimental so that we can do some minor adjustments in 3.10 if needed.


jtattermusch:
CC @ObsidianMinor

PR #6436, opened 2019-07-24
protoc: fix source code info location for missing label

### LINKED ISSUE #6378: protoc: invalid span in source location
Using `protoc v3.6.1`

Given the following .proto file:
```proto
// Copyright 2018 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

syntax = "proto3";

package goproto.protoc.comments;

option deprecated = true;
option go_package = "google.golang.org/protobuf/cmd/protoc-gen-go/testdata/comments";

message DeprecatedMessage {
  option deprecated = true;
  string deprecated_field = 1 [deprecated=true];
}

enum DeprecatedEnum {
  option deprecated = true;
  DEPRECATED = 0 [deprecated=true];
}
```

When compiling this, one of the `google.protobuf.SourceCodeInfo` provided by `protoc` looks like this:
```
location: {
	path: [4, 0, 2, 0, 4]
	span: [13, 2, 12, 27]
}
```
The span in this message makes no sense. It is saying that the location starts on line 13 and ends of line 12.

jhump:
Fixes #6378.

This was always emitting a location for the label, even if no label was present. This resulted in a malformed span -- since the label did not actually exist, the associated span had an end that was _before_ the start.

This patch makes it so that the location is simply omitted if there is no location in source (e.g. optional fields in a file with proto3 syntax).

### DIFF
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -2808,6 +2808,35 @@ TEST_F(SourceInfoTest, Fields) {
   EXPECT_TRUE(HasSpan(file_.message_type(0), "name"));
 }
 
+TEST_F(SourceInfoTest, Proto3Fields) {
+  EXPECT_TRUE(
+      Parse("syntax = \"proto3\";\n"
+            "message Foo {\n"
+            "  $a$int32$b$ $c$bar$d$ = $e$1$f$;$g$\n"
+            "  $h$repeated$i$ $j$X.Y$k$ $l$baz$m$ = $n$2$o$;$p$\n"
+            "}\n"));
+
+  const FieldDescriptorProto& field1 = file_.message_type(0).field(0);
+  const FieldDescriptorProto& field2 = file_.message_type(0).field(1);
+
+  EXPECT_TRUE(HasSpan('a', 'g', field1));
+  EXPECT_TRUE(HasSpan('a', 'b', field1, "type"));
+  EXPECT_TRUE(HasSpan('c', 'd', field1, "name"));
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -2808,6 +2808,34 @@ TEST_F(SourceInfoTest, Fields) {
   EXPECT_TRUE(HasSpan(file_.message_type(0), "name"));
 }
 
+TEST_F(SourceInfoTest, Fields) {
```
acozzette: I think the test is failing because this test case has the same name as the previous case.

jhump: Yep. I was having problems getting `make check` to run locally. I just got it running locally and verified that the test passes. (I also intentionally commented out a line below, hoping it would make the test fail, to ensure that these assertions would catch any location for the label that should _not_ be present.)

Anyhow, it should be good now.

jhump: BTW, where did you see test output with that error? The only things I see in GitHub are the CLA check and a "mergeable" check which was looking at PR labels that I can't set. Or did a third check appear on a prior commit which will eventually show up for the last commit I pushed?

acozzette: I think the test results must have disappeared after you pushed a new commit (since they would be invalidated in that case). I just added the tag to kick off a new test run though, so the results should start reappearing.

acozzette:
@jhump Thanks for fixing this!

PR #6432, opened 2019-07-23
[bazel] Add fixes for --incompatible_load_{cc,java,proto}_rules_from_bzl

Yannic:
Starting with Bazel 0.29.0, C++, Java, and Protobuf rules will require explicit load statements (https://github.com/bazelbuild/bazel/issues/8743, https://github.com/bazelbuild/bazel/issues/8746, https://github.com/bazelbuild/bazel/issues/8922).

PR #6391, opened 2019-07-16
Update six version to 1.12.0 and fix legacy_create_init issue

aaliddell:
When the @six//:six library is used on a target without the `legacy_create_init` flag disabled, the library will not be usable due to __init__.py being empty and the six code will be in six.py. This change forces six to occupy the __init__ name, preventing this file getting created regardless of the `legacy_create_init` setting.

See comments on a74c43bbd906a42b6bf70668c5d515c6f1d8d2cb for context.

PR #6438, opened 2019-07-24
Backport Bazel-incompatible fixes to 3.9.x (orginal: #6432)

Yannic:
@acozzette Thank you for your reviews! Can we cherry-pick this onto 3.9.x and possibly get a new release in the next 2-3 weeks before Bazel 0.29.0 drops?

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,5 +1,9 @@
 # Bazel (https://bazel.build/) BUILD file for Protobuf.
 
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test", "objc_library")
+load("@rules_java//java:defs.bzl", "java_library")
+load("@rules_proto//proto:defs.bzl", "proto_lang_toolchain", "proto_library")
+
 licenses(["notice"])
 
 exports_files(["LICENSE"])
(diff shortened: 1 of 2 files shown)


acozzette:
Thanks, @Yannic. I am guessing we probably won't do another 3.9.x release because we usually don't do patch releases unless there's a particularly important bug fix. But it can't hurt to cherrypick this just in case. Actually now that I think of it, we could potentially do a Bazel-only 3.9.1 release. The releases are a lot of work when we have to publish artifacts for every language, but if it's just for Bazel then that is easy.

Yannic:
@acozzette I see. Maybe let's wait then to see if there are other changes that justify a patch release. It's always possible to cut a new release if lots of people start complaining about a broken build.

jmillikin-stripe:
I think there was a typo in this PR. `build_file = "@//:six.BUILD",` is referencing a file in the top-level workspace -- you probably meant `build_file = "@com_google_protobuf//:six.BUILD",` instead.

Yannic:
Yes, your right. Voting for #6391 to fix it. @anandolee @acozzette Can you help to get the fix merged?

PR #6440, opened 2019-07-25
Add binary conformance test for default repeated fields

TeBoring:
1) Both packed and unpacked encoding should be accepted for parsing.
2) Encode should follow the default way for the syntax.


conformance/conformance_test.cc
```diff
@@ -54,6 +54,25 @@ using google::protobuf::util::MessageDifferencer;
 using google::protobuf::util::Status;
 using std::string;
 
+namespace {
+
+static string ToHexString(const string& binary_string) {
```
acozzette: Isn't this octal rather than hex?

conformance/conformance_test.cc
```diff
@@ -54,6 +54,25 @@ using google::protobuf::util::MessageDifferencer;
 using google::protobuf::util::Status;
 using std::string;
 
+namespace {
+
+static string ToOctString(const string& binary_string) {
+  string hex_string;
```
acozzette: Small nit: rename hex_string to oct_string

conformance/conformance_test.cc
```diff
@@ -270,8 +292,20 @@ void ConformanceTestSuite::VerifyResponse(
...
 
-  bool check;
-  check = differencer.Compare(*reference_message, *test_message);
+  bool check = false;
+
+  if (require_same_wire_format) {
+    GOOGLE_DCHECK_EQ(response.result_case(),
+                    ConformanceResponse::kProtobufPayload);
+    const string& protobuf_payload =  response.protobuf_payload();
+    check = (equivalent_wire_format.compare(
```
acozzette: I think instead of doing `equivalent_wire_format.compare(...) == 0` you can just do `equivalent_wire_format == protobuf_payload`.

PR #6447, opened 2019-07-26
Conformance packed

TeBoring:
Tests explicit packed/unpacked fields.


conformance/binary_json_conformance_suite.cc
```diff
@@ -151,18 +151,43 @@ string tag(uint32_t fieldnum, char wire_type) {
 
 #define UNKNOWN_FIELD 666
 
-const FieldDescriptor* GetFieldForType(FieldDescriptor::Type type,
-                                       bool repeated, bool is_proto3) {
-
+enum Packed {
```
acozzette: Small nit: make this an `enum class` to keep the enum values from going in the global namespace.

acozzette:
LGTM

PR #6451, opened 2019-07-29
[Python] struct_pb2.Struct class should behave like a native Python dictionary when using __getitem__ / dict()

Kami:
## Background

Currently, ``struct_pb2.Struct`` behaves somewhat like a Python dictionary, but not quite. It doesn't handle nested structs (aka complex types) correctly.

For example:

The following script (https://gist.github.com/Kami/e40375f5cae99f8ee28dd4780b26486a), will return this:

```python
{'x': True, 'a': 1.0, 'b': 2.0, 'c': 'foo bar'}
{'foo': 'bar', 'nested': fields {
  key: "a"
  value {
    list_value {
      values {
        number_value: 1.0
      }
      values {
        number_value: 2.0
      }
      values {
        number_value: 3.0
      }
    }
  }
}
fields {
  key: "c"
  value {
    string_value: "d"
  }
}
fields {
  key: "f"
  value {
    struct_value {
      fields {
        key: "g"
        value {
          struct_value {
            fields {
              key: "h"
              value {
                string_value: "j"
              }
            }
            fields {
              key: "m"
              value {
                list_value {
                  values {
                    struct_value {
                      fields {
                        key: "a"
                        value {
                          string_value: "1"
                        }
                      }
                      fields {
                        key: "d"
                        value {
                          bool_value: true
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
}
```

In the first example where struct only contains simple types, correct / expected value is returned. But in the second example, it won't recurse into complex / nested types and incorrect / unexpected value is returned.

This brakes Python ``dict()`` API and makes structs hard to work with since you need extra glue code to work around those limitations.

In reality, I would expect it to return something like this (as in my proposed solution below):

```python
{'x': True, 'a': 1.0, 'b': 2.0, 'c': 'foo bar'}
{'foo': 'bar', 'nested': {'f': {'g': {'h': 'j', 'm': [{'a': '1', 'd': True}]}}, 'a': [1.0, 2.0, 3.0], 'c': 'd'}}
```

## Proposed Solution

I believe calling ``dict()`` on Struct class instance should return a Python dictionary which only contains simple types (aka behaves like a regular Python dictionary).

This means we need to update the code to recurse into complex types.

In my proposed solution, I added new ``to_dict()`` method, but I think the best course of action would be to modify ``__getitem__`` to call ``_GetStructValue``. This way calling ``dict()`` on the struct instance will behave as expected.

The reason I didn't make that change is because it breaks one of the tests and it's not backward compatible - we need to decide how to handle backward compatibility.


Kami:
When you get a chance to review this, please let me know if you are fine with changing ``__getitem__`` behavior so it behaves as expected if you call ``dict()`` on the Struct class instance.

As mentioned above, this would potentially be a breaking / backward incompatible change. I think this is the right behavior though. We just need to document the change in release notes or similar.

anandolee:
Can you let me know why do you want to convert a Struct to dict? We have a lot of struct\list like object in python, but didn't provide convert APIs for them. If it is only for print, have you tried json_format.MessageToJson() ?

Kami:
@anandolee It's not for printing - it's so I can dynamically process data which is stored inside that struct without needed to know how to handle all the special struct member types (aka I just want to work with simple native Python types, I don't care about the struct "internals").

Right now, if I work directly with struct which contains nested complex types, I can't process it directly just by iterating over it - I need to know the struct type "internals" aka know how to handle list values and other types.

Let's take a concrete example:

```python
from google.protobuf import struct_pb2

data = {
    'parameters': {
        'param11': {
            'type': 'array',
            'default_value': [1, 2, 3, 4]
        },
        'param2': {
            'type': 'object',
            'default_value': {'a': 1}
        }
    }
}

struct = struct_pb2.Struct()
struct.update(data)

for param_name, param_opts in struct.to_dict()['parameters'].items():
    print('Param name: %s' % (param_name))
    print('Param opts: %s' % (param_opts))

print('')

for param_name, param_opts in dict(struct['parameters']).items():
    print(param_name)
    print(param_opts)
```

And the output:

```python
Param name: param2
Param opts: {'type': 'object', 'default_value': {'a': 1.0}}
Param name: param11
Param opts: {'default_value': [1.0, 2.0, 3.0, 4.0], 'type': 'array'}

param2
fields {
  key: "default_value"
  value {
    struct_value {
      fields {
        key: "a"
        value {
          number_value: 1.0
        }
      }
    }
  }
}
fields {
  key: "type"
  value {
    string_value: "object"
  }
}

param11
fields {
  key: "default_value"
  value {
    list_value {
      values {
        number_value: 1.0
      }
      values {
        number_value: 2.0
      }
      values {
        number_value: 3.0
      }
      values {
        number_value: 4.0
      }
    }
  }
}
fields {
  key: "type"
  value {
    string_value: "array"
  }
}
```

I know this is just a contrived example, but I hope you get the idea.

Basically, I just want to get back a version which looks the same as the input ``data`` dictionary (minus the differences between integer / double type).

anandolee:
Does this work for you:
json_string = json_format.MessageToJson(struct_message) 
converted_struct = json.loads(json_string)

Kami:
@anandolee Yes, that does work, but I still believe calling ``dict()`` on struct valye should return actual Python dictionary with native types.

Which as mentioned above, does happen, but only for non-nested structs.

AlekzanderDee:
Hello,

Do you plan on merging it any time soon? If I understand it right, it will fix [this](https://github.com/protocolbuffers/protobuf/issues/6537) issue.

Thank you!

Kami:
@anandolee Please let me know if there is anything else I can do to help move this along?

Thanks.

In the mean time, I'm using a utility function for converting Struct to dict with native types, but as mentioned in the PR description, I believe this should happen inside struct ``__getitem__`` and struct should behave like an actual Python dictionary when calling ``dict()`` / ``__getitem__()`` on it.

AlekzanderDee:
Any updates on this?

AlekzanderDee:
do you plan to fix the MacOS Ruby build and merge it? thank you!

adisunw:
status on this?

NikPaushkin:
Hi guys, when merge? 2 years gone

python/google/protobuf/internal/well_known_types.py
```diff
@@ -750,6 +750,23 @@ def _GetStructValue(struct_value):
     raise ValueError('Value not set')
 
 
+def _StructToDict(struct_value):
```
anandolee, line 753: Current implementation is a little confusing because struct_value can be a List or other type. From the _StructToDict() name, only Struct should call this function. The implement can be this:
```
def _StructToDict(struct_value):
    result = {}
    for key, value in struct_value.items():
       if isinstance(value, Struct):
            ...
       elif isinstance(value, ListValue):
            ...  # may need to add the to_list method
       elif getattr(struct_value, 'WhichOneof', None):
            ...
       else:
            ...
    return result
```

acozzette:
@Kami I'm going to close this since it's been a few weeks with no action, but feel free to comment here if you would like to reopen this.

PR #6453, opened 2019-07-29
Use value ctor in case of message values in map

chemhack:
Address https://github.com/protocolbuffers/protobuf/pull/6394#issuecomment-516090227

@TeBoring


TeBoring:
I had another fix: https://github.com/protocolbuffers/protobuf/pull/6454

PR #6454, opened 2019-07-29
Fix the bug for message value map

TeBoring:
Previously, jspb.BinaryReader.prototype.readMessage complains it cannot accept null.
Also added test to it.

### DIFF
--- a/js/testbinary.proto
+++ b/js/testbinary.proto
@@ -257,7 +257,7 @@ message MapEntryOptionalValuesBoolValue {
 
 message MapEntryOptionalValuesDoubleValue {
   optional string key = 1;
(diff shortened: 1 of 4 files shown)


js/testbinary.proto
```diff
@@ -232,6 +232,56 @@ message TestMapFieldsOptionalKeys {
...
+}
+
+message MapEntryOptionalValuesBoolValue {
+  optional string key = 1;
+  optional bool value = 2;
+}
+
+message MapEntryOptionalValuesDoubleValue {
+  optional string key = 1;
+  optional bool value = 2;
```
acozzette: Shouldn't this be a double instead of bool?

PR #6455, opened 2019-07-30
Fix a Python extension crash in descriptor's nested types container

gaeldelalleau:
This PR adds tests to check some basic behavior of a descriptor's nested types container.
It fixes a bug that appear in the C++ implementation (Python extension) when trying to check if an object with an invalid (non-descriptor) type is present in the nested_types container :

======================================================================
ERROR: testDescriptorNestedTypesContainer (google.protobuf.internal.descriptor_test.GeneratedDescriptorTest)
----------------------------------------------------------------------
TypeError: Not a BaseDescriptor

The above exception was the direct cause of the following exception:

  File "/protobuf/python/google/protobuf/internal/descriptor_test.py", line 657, in testDescriptorNestedTypesContainer
    self.assertFalse(None in message_descriptor.nested_types)
SystemError: PyEval_EvalFrameEx returned a result with an error set

----------------------------------------------------------------------

### DIFF
--- a/python/google/protobuf/internal/descriptor_test.py
+++ b/python/google/protobuf/internal/descriptor_test.py
@@ -650,6 +650,13 @@ class GeneratedDescriptorTest(unittest.TestCase):
     del enum
     self.assertEqual('FOO', next(values_iter).name)
 
+  def testDescriptorNestedTypesContainer(self):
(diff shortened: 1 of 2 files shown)


anandolee:
Thanks for the PR

Issue #6399, opened 2019-07-18
add SetStringByArray(..., const char *, size_t) method in Reflection


reed-lau:
**What language does this apply to?**
cxx and other binding languages such as python's cpp_implenentation

**Describe the problem you are trying to solve.**
the Reflection class in message.h only has `SetString(..., const std::string &value)`interface. in some cases, we just have raw buffer (char * and size), to call the SetString method, we have to construct/memcpy/deconstruct a std::string. But in some time-sensitive case, this consumes much, especially when the the bytes is large enough (eg. 8MB), which will cause large amount of page-faults.

**Describe the solution you'd like**
Add `SetStringByArray(..., const char *, size_t )` interface, and call it when necessorily.

**Describe alternatives you've considered**
even in protobuf, construct 8MB bytes(std::string) will cause large amount of page-fault, this will slow down much. is it possible to setup an allocator for the protobuf. I did not seen such method. 

**Additional context**

acozzette:
Unfortunately I don't think there's any easy way to do this. Internally messages store `string` and `bytes` fields as `std::string`, so ultimately we have to construct a real `std::string`. However, we could potentially add an overload that takes a `std::string&&` and then we could at least move the string into place without creating another copy.

reed-lau:
I'm sorry, I did not describe it clearly.
I uses the `contruct/memcpy/deconstruct` did not mean the protobuf's internal data storage but the extra std::string, which must be create to adapt to the interface.
like the following code, could we remove the `temp` std::string's `construct/memcpy/deconstruct`
```c
bool foo(const char *buf, size_t size,...) {
  ...
  std::string temp(buf, size); // actually, we don't need to create a temporary std::string here, if we have 
                               // some method like SetStringByArray(..., buf, size); 
  reflection->SetString(...., temp);
  ...
}
```
the original problem comes from the following code, and I measured it in a 8MB bytes case. the  `value_string`'s construct and memcpy take about 3ms( actually due to the page fault).
https://github.com/protocolbuffers/protobuf/blob/a03d332aca5d33c5d4b2cd25037c9e37d57eff02/python/google/protobuf/pyext/message.cc#L787

acozzette:
If the string is passed by value or by rvalue reference then we can eliminate one of the two copies. The user can create the string and then it can be moved directly into place.

t-rybarski:
[...]

PR #6477, opened 2019-08-05
C++ Add move constructor for Reflection's SetString

reed-lau:
Add move constructor for Reflection's SetString
add feature https://github.com/protocolbuffers/protobuf/issues/6399

### DIFF
--- a/src/google/protobuf/extension_set.h
+++ b/src/google/protobuf/extension_set.h
@@ -269,6 +269,7 @@ class PROTOBUF_EXPORT ExtensionSet {
   void SetBool(int number, FieldType type, bool value, desc);
   void SetEnum(int number, FieldType type, int value, desc);
   void SetString(int number, FieldType type, const std::string& value, desc);
+  void SetString(int number, FieldType type, std::string&& value, desc);
   std::string* MutableString(int number, FieldType type, desc);
   MessageLite* MutableMessage(int number, FieldType type,
                               const MessageLite& prototype, desc);
@@ -332,6 +333,7 @@ class PROTOBUF_EXPORT ExtensionSet {
   void SetRepeatedBool(int number, int index, bool value);
   void SetRepeatedEnum(int number, int index, int value);
   void SetRepeatedString(int number, int index, const std::string& value);
+  void SetRepeatedString(int number, int index, std::string&& value);
   std::string* MutableRepeatedString(int number, int index);
   MessageLite* MutableRepeatedMessage(int number, int index);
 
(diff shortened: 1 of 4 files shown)


reed-lau:
@googlebot I signed it!

src/google/protobuf/generated_message_reflection.cc
```diff
@@ -1223,6 +1223,38 @@ void Reflection::SetString(Message* message, const FieldDescriptor* field,
 }
 
 
+void Reflection::SetString(Message* message, const FieldDescriptor* field,
+                           const std::string&& value) const {
```
t-rybarski: @reed-lau rvalue references cannot be const as they cannot be moved. The overloaded functions rvalue parameter should be SetString(..,..,std::string&& value) without the const!

src/google/protobuf/generated_message_reflection.cc
```diff
@@ -1223,6 +1223,38 @@ void Reflection::SetString(Message* message, const FieldDescriptor* field,
...
+  USAGE_CHECK_ALL(SetString, SINGULAR, STRING);
+  if (field->is_extension()) {
+    return MutableExtensionSet(message)->SetString(field->number(),
+                                                   field->type(), value, field);
```
t-rybarski: Also here i believe the value parameter (i.e. the input string) should be moved, so:
`SetString(...,..., std::move(value), field)`
Additionaly, the string which you are creating in your code should also be moved into place:

`bool foo(const char *buf, size_t size,...) {
  ...
  std::string temp(buf, size); // actually, we don't need to create a temporary std::string here, if we have 
                               // some method like SetStringByArray(..., buf, size); 
  reflection->SetString(...., std::move(temp));
  ...
}`

reed-lau:
@t-rybarski  fixed it according to your comment.

reed-lau:
@acozzette @protobuf-kokoro 
what should I do to fix the Kokoro build failure for `Linux Golang`?

acozzette:
That Go error looks unrelated, so I wouldn't worry about it for now.

src/google/protobuf/extension_set.h
```diff
@@ -269,6 +269,7 @@ class PROTOBUF_EXPORT ExtensionSet {
   void SetBool(int number, FieldType type, bool value, desc);
   void SetEnum(int number, FieldType type, int value, desc);
   void SetString(int number, FieldType type, const std::string& value, desc);
+  void SetString(int number, FieldType type, std::string&& value, desc);
```
acozzette: I believe it's possible to enable this optimization without creating new overloads. Could you try taking the existing methods and changing them to take a `std::string` (by value) instead of `const std::string&`? I think that should be backward-compatible with existing callers and still allow us to optimize the implementation.

reed-lau:
after change the interface to `std::string`(by value), the following line may using the `std::move` form, and it could benefit from it, but I did change it at this PR.
https://github.com/protocolbuffers/protobuf/blob/d2d6ff51a8fb67391c44223c6b0b523110ad8d21/src/google/protobuf/generated_message_reflection.cc#L589

acozzette:
This looks great but we are having some issues with Docker in our CI setup. Hopefully we will get the test issues sorted out soon and then I plan to merge this.

PR #6445, opened 2019-07-26
[bazel] Add fixes for --incompatible_load_{cc,java,proto}_rules_from_bzl (Part 2)

Yannic:
This change adds the required loads to examples and zlib. Sorry for missing them in the first PR.
For full compatibility with --incompatible_load_{cc,java,proto}_rules_from_bzl,
we will need to roll gtest to a newer version.

PR #6481, opened 2019-08-06
[bazel] Load python rules from @rules_python

Yannic:
This should be the last change to make Protobuf compatible with `--incompatible_load_*_rules_from_bzl`. For full compatibility, we'll have to update third_party/googletest after https://github.com/google/googletest/pull/2364 has landed.


Yannic:
Test failure seems unrelated.

It seems like there is going to be a 3.9.1 release. Can we cherry-pick this one and #6445 into it or is it already too late?

acozzette:
3.9.1 was already published yesterday and today, so I'm afraid it will have to wait for 3.10.

acozzette:
Thanks, @Yannic.

PR #6483, opened 2019-08-07
No free when construct was not done

TeBoring:
This fix the segfault when some other error happens


TeBoring:
broken tests are due to docker hub. Not related to this change.

PR #6491, opened 2019-08-08
Accept string for int64 wrappers

TeBoring:


TeBoring:
upb fixed in https://github.com/protocolbuffers/upb/pull/202

TeBoring:
Broken test is being fixed in https://github.com/protocolbuffers/protobuf/issues/6532

PR #6495, opened 2019-08-10
add Map forward declaration in arena.h

DNKpp:
I'm currently using protobuf with godot and during compilation of godot the compiler fails because godot has a similar "Map" type.
This 2 lines of forward declaration fixes this.


DNKpp:
@googlebot I signed it!

PR #6499, opened 2019-08-13
Document proto2 feature implementations in Google.Protobuf

ObsidianMinor:
@jtattermusch

### DIFF
--- a/docs/csharp/proto2.md
+++ b/docs/csharp/proto2.md
@@ -1,4 +1,4 @@
-As part of the 3.10 release of Google.Protobuf, proto2 support has been reimplemented. This document outlines the new changes brought about by this experimental rerelease of proto2 support. This does not break existing proto3 support and users may continue to use proto3 features without changing their current code. Again the generated code and public API associated with proto2 is experimental and subject to change in the future. APIs for proto2 may be added, removed, or adjusted as feedback is received.
+As part of the 3.10 release of Google.Protobuf, experimental proto2 support has been released. This document outlines the new changes brought about to include proto2 support. This does not break existing proto3 support and users may continue to use proto3 features without changing their current code. Again the generated code and public API associated with proto2 is experimental and subject to change in the future. APIs for proto2 may be added, removed, or adjusted as feedback is received.
 Generated code for proto2 may also be modified by adding, removing, or adjusting APIs as feedback is received.
 
 ### Enabling proto2 features
--- a/docs/csharp/proto2.md
+++ b/docs/csharp/proto2.md
@@ -1,17 +1,19 @@
-Proto2 support in Google.Protobuf is finally here! This document outlines the new changes brought about 
-by this initial release of proto2 support. The generated code and public API associated with proto2 
-is experimental and subject to change in the future. APIs may be added, removed, or adjusted as feedback is received.
-Generated code may also be modified, adding, removing, or adjusting APIs as feedback is received.
+As part of the 3.10 release of Google.Protobuf, proto2 support has been reimplemented. This document outlines the new changes brought about by this experimental rerelease of proto2 support. This does not break existing proto3 support and users may continue to use proto3 features without changing their current code. Again the generated code and public API associated with proto2 is experimental and subject to change in the future. APIs for proto2 may be added, removed, or adjusted as feedback is received.
+Generated code for proto2 may also be modified by adding, removing, or adjusting APIs as feedback is received.
+
+### Enabling proto2 features
+
+For information about specific proto2 features, please read the [proto2 language guide](https://developers.google.com/protocol-buffers/docs/proto).
+
+Much like other languages, proto2 features are used with proto2 files with the syntax declaration `syntax = "proto2";`. However, please note, proto3 is still the recommended version of protobuf and proto2 support is meant for legacy system interop and advanced uses.
 
 # Generated code
 
 ### Messages
 
-Messages in proto2 files are very similar to their proto3 counterparts. However, they have some added properties
-and methods to handle field presence.
+Messages in proto2 files are very similar to their proto3 counterparts. They expose the usual property for getting and setting,  but they also include properties and methods to handle field presence.
 
-A normal single value proto2 fields will have a normal property for getting and setting, as well as a 
-`HasValue` property for checking presence, and a `Clear` method for clearing the value.
+For `optional`/`required` field XYZ, a `HasXYZ` property is included for checking presence and a `ClearXYZ` method is included for clearing the value.
 
 ```proto
 message Foo {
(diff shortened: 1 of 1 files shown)


docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
+Proto2 support in Google.Protobuf is finally here! This document outlines the new changes brought about
```
jtattermusch: nit: we should add info about starting from which release proto2 is available.

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
+Proto2 support in Google.Protobuf is finally here! This document outlines the new changes brought about 
+by this initial release of proto2 support. The generated code and public API associated with proto2
```
jtattermusch: nit: let's change the tone to sound less like an announcement of the initial release and more as documentation that's there to stay (because I assume we will keep this doc for a while).

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
+Proto2 support in Google.Protobuf is finally here! This document outlines the new changes brought about 
+by this initial release of proto2 support. The generated code and public API associated with proto2 
+is experimental and subject to change in the future. APIs may be added, removed, or adjusted as feedback is received.
```
jtattermusch: optional nit: change to "APIs as well as the generated code may ..."? 
internal note: of course even though proto2 has been declared experimental, we can only change the apis/ generated code so that it doesn't break existing proto3 users (who are relying on the API stability and the proto2 support being there is irrelevant to them).

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
...
+Generated code may also be modified, adding, removing, or adjusting APIs as feedback is received.
+
+# Generated code
+
+### Messages
+
+Messages in proto2 files are very similar to their proto3 counterparts. However, they have some added properties
+and methods to handle field presence.
+
+A normal single value proto2 fields will have a normal property for getting and setting, as well as a
```
jtattermusch: small nit: rephrase as "will expose the usual property for getting and setting"

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
...
+
+# Generated code
+
+### Messages
+
+Messages in proto2 files are very similar to their proto3 counterparts. However, they have some added properties
+and methods to handle field presence.
+
+A normal single value proto2 fields will have a normal property for getting and setting, as well as a 
+`HasValue` property for checking presence, and a `Clear` method for clearing the value.
```
jtattermusch: HasXYZ   and ClearXYZ  (XYZ is the name of the field?)

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
+Proto2 support in Google.Protobuf is finally here! This document outlines the new changes brought about 
+by this initial release of proto2 support. The generated code and public API associated with proto2 
+is experimental and subject to change in the future. APIs may be added, removed, or adjusted as feedback is received.
+Generated code may also be modified, adding, removing, or adjusting APIs as feedback is received.
+
+# Generated code
```
jtattermusch: it might be useful to add a small section "Enabling proto2 features"  that says that for proto3 users, nothing is changing and that to use proto2, one needs to have .proto files that specify syntax = proto2.
Also mention that proto3 is the recommended version of protobuf and everyone should use that unless they have special needs (perhaps link to proto3 docs). Proto2 support is there for more advanced use cases (e.g. working with proto descriptors that are historically proto2) or to interop with legacy systems that use proto2.

jtattermusch: Also link to proto2 docs?

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
...
+var foo = new Foo();
+foo.SetExtension(FooExt, new Baz());
+foo.GetOrInitializeExtension(RepeatedFooExt).Add(new Baz());
+```
+
+# APIs
+
+### Message initialization
+
+Checking message initialization is not handled automatically by the library, but can be done manually via the
```
jtattermusch: it might be useful to elaborate a bit more on what's meant by "message initialization". I think many users won't be familiar with what exactly that means.
we could also answer these questions: Also, what are the consequences of needing to do IsInitialized() manually? Is that the ways it behaves in other languages? 
If this is a limitation compared to other languages, it might be worth putting one sentence reasoning of that design decision.

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
...
+};
+var foo = Foo.Factory.WithExtensionRegistry(registry).ParseFrom(input);
+Assert.True(foo.HasExtension(Bas.Extensions.FooExt));
+var fooNoRegistry = Foo.Factory.ParseFrom(input);
+Assert.False(foo.HasExtension(Bas.Extensions.FooExt));
+```
+
+### Custom options
+
+The original `CustomOptions` APIs are now deprecated. Using the new generated extension identifiers,
```
jtattermusch: optional nit: maybe add one more sentence of detail why deprecating CustomOptions is a good thing?

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
...
+        {
+            Console.WriteLine($"    {method.Name}: no HTTP binding");
+        }
+    }
+}
+```
+
+### Reflection
+
+Reflection APIs have been included to access the new portions of the library.
```
jtattermusch: optional nit: rephrase as "Reflection APIs have been extended to enable accessing the newly added proto2 concepts " ?

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
...
+    * Returns whether a field's value is set. For proto3 fields, throws an InvalidOperationException.
+ * FileDescriptor.Syntax
+    * Gets the syntax of a file
+ * FileDescriptor.Extensions
+    * An immutable list of extensions defined in the file
+ * MessageDescriptor.Extensions
+    * An immutable list of extensions defined in the message
+
+```cs
+var extensions = Baz.Descriptor.Extensions.GetExtensionsInDeclarationOrder(Foo.Descriptor);
```
jtattermusch: idea: in the future, if this is a pattern that's used a lot, there could be an API (extension method) that allows you to create extension registry populated with all extensions from given descriptor?

ObsidianMinor: It seems easy enough to do manually via `AddRange` and `UnorderedExtensions`

```cs
var registry = new ExtensionRegistry();
registry.AddRange(Baz.Descriptor.Extensions.UnorderedExtensions);
```

ObsidianMinor:
I've addressed your review comments @jtattermusch

docs/csharp/proto2.md
```diff
@@ -0,0 +1,184 @@
+As part of the 3.10 release of Google.Protobuf, proto2 support has been reimplemented. This document outlines the new changes brought about by this experimental rerelease of proto2 support. This does not break existing proto3 support and users may continue to use proto3 features without changing their current code. Again the generated code and public API associated with proto2 is experimental and subject to change in the future. APIs for proto2 may be added, removed, or adjusted as feedback is received.
```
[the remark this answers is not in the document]
ObsidianMinor: Fixed. I had misinterpreted git history and the C# readme and thought the library originally supported proto2.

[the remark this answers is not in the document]
jtattermusch: I see the git history could have easily made you think that. The story is that we originally imported Jon Skeet's protobuf-csharp-port (which supported proto2) and we reworked it significantly (changes APIs, removed proto2, etc.) before we released the first version of Google.Protobuf. So Google.Protobuf itself never supported proto2.

Issue #6496, opened 2019-08-12
Ruby: decoding an empty protocol buffer allocates objects for every field in that protobuf's descriptor


penelopezone:
```
require 'google/protobuf'

Google::Protobuf::DescriptorPool.generated_pool.build do
  add_message "hi" do
    # vary this to see more or less allocations
    10.times do |i|
      optional :"resource_name_#{i}", :string, i+1
    end
  end
end

module Mod
  Hi = Google::Protobuf::DescriptorPool.generated_pool.lookup("hi").msgclass
end

h = Mod::Hi.new
data = Mod::Hi.encode(h)

# 4 GCs to force all allocated objects to either be removed
# or moved to the oldest generation
GC.start
GC.start
GC.start
GC.start

start_alloc = GC.stat
# uncomment either
#Mod::Hi.decode("")
#Mod::Hi.decode(data)
end_alloc = GC.stat

p end_alloc[:total_allocated_objects]-start_alloc[:total_allocated_objects]
```

samphippen:
(Just dropped this in here from a hangout with @haberman)

haberman:
Confirmed. This is specific to string fields. I have a fix for this here: https://github.com/protocolbuffers/protobuf/pull/6502

PR #6502, opened 2019-08-13
Optimized away the creation of empty string objects.

haberman:
Prior to this CL, creating an empty message object would create
two empty string objects for every declared field.  First we
created a unique string object for the field's default.  Then
we created yet another string object when we assigned the
default value into the message: we called #encode to ensure
that the string would have the correct encoding and be frozen.

I optimized these unnecessary objects away with two fixes:

1. Memoize the empty string so that we don't create a new empty
   string for every field's default.
2. If we are assigning a string to a message object, avoid creating
   a new string if the assigned string has the correct encoding and
   is already frozen.


haberman:
This PR fixes https://github.com/protocolbuffers/protobuf/issues/6496.

PR #6504, opened 2019-08-13
JS: add "int64_type" option to set default type of 64-bit integers

lqs:
so we can use `int64_type=string` in the protoc option and don't need to add `[ jstype = JS_STRING ]` on every field.


TeBoring:
Later we will flip the default.

seishun:
@TeBoring when is this going to happen?

edvardchen:
@TeBoring Any Update?

PR #6514, opened 2019-08-14
Fix compilation on clang-cl

m0hamed:
For RepeatedPtrField<> `explicit instantiation declaration should not be
'dllexport'` (you cant put dllexport in the declaration of templated types only the definition)

For WrappedMutex `__attribute__(())` must come before `__declspec()`


m0hamed:
closing this because of a bug, will upload another one once fixed

Issue #6532, opened 2019-08-19
Ruby concurrent map decoding sporadically has empty results


haberman:
The Ruby unit tests occasionally fail with:

```
Failure: test_concurrent_decoding(BasicTest::MessageContainerTest)
/tmp/protobuf/protobuf/ruby/tests/basic.rb:245:in `block (3 levels) in test_concurrent_decoding'
     242:       thds = 2.times.map do
     243:         Thread.new do
     244:           100000.times do
  => 245:             assert_equal o, Outer.decode(raw)
     246:           end
     247:         end
     248:       end
/tmp/protobuf/protobuf/ruby/tests/basic.rb:244:in `times'
/tmp/protobuf/protobuf/ruby/tests/basic.rb:244:in `block (2 levels) in test_concurrent_decoding'
<<BasicTest::Outer: items: {0=><BasicTest::Inner: >}>> expected but was
<<BasicTest::Outer: items: {}>>
diff:
? <BasicTest::Outer: items: {0=><BasicTest::Inner: >}>
```

Here is a [sample failure in Kokoro](https://source.cloud.google.com/results/invocations/e355b503-6ede-4c49-88d7-dc5bf4450b74/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fruby23%2Fpresubmit/log).

This was first observed on the PR https://github.com/protocolbuffers/protobuf/pull/5866, so is presumed to be related. However the bug is very sporadic and difficult to reproduce, so it's hard to say for sure when it was introduced.

Previously I tried the following strategies to make the bug more reproducible:

- I compiled a custom Ruby that uses 1ms timeslices instead of 100ms
- I've increased the number of map entries to 1,000
- I set GC.stress = 0x01 | 0x04 like gc_test.rb

haberman:
I've had the most luck reproducing this on Ruby 2.3, so that's what I've been using.

This doesn't appear to be GC-related. I can reproduce the error even if the test is wrapped in `GC.disable`, `GC.enable`.

I can get this modified version of the test to fail about 1 out of 15 times on the Ruby 2.3 docker image.

```ruby
    def test_concurrent_decoding
      GC.start
      GC.disable
      o = Outer.new
      o.items[0] = Inner.new
      raw = Outer.encode(o)

      thds = 2.times.map do |t|
        Thread.new do
          assert_equal o, Outer.decode(raw), "thread #{t}, raw=#{raw.length}"
        end
      end
[...]

PR #6533, opened 2019-08-19
Fix for GC of Ruby map frames.

haberman:
We were creating a map decoding frame when starting the *map*,
but clearing the GC slot when finishing each *map entry*.  This
means that the decoding frame could be collected in the meantime.

Hopefully this will fix https://github.com/protocolbuffers/protobuf/issues/6532 (needs to be tested/verified).

I have not verified that this fixes anything yet, but it does seem to correct a correctness error in the previous code.


TeBoring:
The existing gc test only tests map with one key/value pair. Can we add more key/value pair into the test?
https://github.com/protocolbuffers/protobuf/blob/master/ruby/tests/gc_test.rb#L39

Issue #6217, opened 2019-06-05
Proposal: C# RepeatedField.AsSpan() for more efficient access


prat0088:
**What language does this apply to?**

c# protobuf 3

**Describe the problem you are trying to solve.**

Efficient lookups and assignments of RepeatedField<> (10s of millions )

It currently performs bounds and null checks inside an indexer method.  Indexing into a span is essentially the same as a ref return and the JIT compiler can optimize that even further to the point it is almost identical to operating on the array itself.  

I haven't yet benchmarked this proposal, but I expect it to perform similarly to ref returns, which are 1.25-10x faster than RepeatedField's current indexer method.

**Describe the solution you'd like**

```
public Span<T> AsSpan(int start, int length) {
    // Bounds check...

    return this.array.AsSpan(start, length);
}

```

Some might say this bypasses the null checks of the indexer method, but the C++ implementation allows you to memcpy entire memory chunks into RepeatedFields from what I understand.  While I feel a good library prevents the developer from using it incorrectly, I also feel it should be possible to squeeze extra performance out of a library if the developer accepts the risks.

**Describe alternatives you've considered**

- Adding a preprocessor variable that exposes a ref readonly indexer to access the array directly, without bounds or null checks.  I saw a 1.25 - 10x speedup in reads and writes.  I don't like this because the developer can no longer simply use the copy of Nuget.  
- I'm using a fork of Protobuf now that exposes the array directly.  

**Additional context**

I'm serving up ~30MB messages that contain millions of integers in RepeatedField<>s.  I need to make it as fast and efficient as possible while still using Protobuf.

Code-wise it's a small addition.  I'm happy to submit a PR if the above looks ok.

prat0088:
@acozzette 

This requires netstandard 2.1.  It looks like Protobuf currently targets netstandard 1.0 and 2.0.  

Is it ok if I add a new target for netstandard2.1?

prat0088:
One more point to bring up now: Returning a Span will allow the user to potentially write null values to the RepeatedField because it bypasses the null value check.  It's possible to write null values to RepeatedField in C++, too.  This gives such a great boost to performance (1-10x).  I think it is sufficient to add a note to the documentation that the user is responsible for null checks when using this method.

ObsidianMinor:
We have a .NET Standard 2.0 target for System.Memory. You should use that.

Falco20019:
[...]

PR #6538, opened 2019-08-20
c# feature: new method AsSpan() on RepeatedField<T>

prat0088:
Provides very efficient access to the underlying data at the expense of foregoing null checks.

#6217

### DIFF
--- a/csharp/src/Google.Protobuf/Collections/RepeatedField.cs
+++ b/csharp/src/Google.Protobuf/Collections/RepeatedField.cs
@@ -543,6 +543,20 @@ namespace Google.Protobuf.Collections
             }
         }
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Creates a new span over the RepeatedField.  This is the most
+        /// efficient way to read and write values to the RepeatedField because
+        /// there are no internal null or bounds checks.  You are responsible
+        /// for ensuring no null values are written to the span.
+        /// </summary>
+        /// <returns>Span representation of the RepeatedField</returns>
+        public Span<T> AsSpan()
+        {
+            return array.AsSpan(0, count);
+        }
+#endif
+
         #region Explicit interface implementation for IList and ICollection.
         bool IList.IsFixedSize => false;
 
--- a/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
@@ -756,5 +756,33 @@ namespace Google.Protobuf.Collections
             Assert.True(list1.Contains(SampleNaNs.SignallingFlipped));
             Assert.False(list2.Contains(SampleNaNs.SignallingFlipped));
         }
+
+#if NETCOREAPP2_1
+        [Test]
+        public void AsSpan_BasicUsage()
+        {
+            var list = new RepeatedField<int> { 10, 11 };
+            var span = list.AsSpan();
+
+            // Span was constructed correctly
+            Assert.AreEqual(2, span.Length);
+            Assert.AreEqual(10, span[0]);
+            Assert.AreEqual(11, span[1]);
+
+            // Writing to span writes to RepeatedField
+            span[0] = 100;
+            Assert.AreEqual(100, list[0]);
+        }
+
+        [Test]
+        public void AsSpan_Empty()
+        {
+            var list = new RepeatedField<int>();
+            var span = list.AsSpan();
+
+            // Empty Span was constructed correctly
+            Assert.AreEqual(0, span.Length);
+        }
+#endif
     }
 }


prat0088:
@googlebot I signed it!

jtattermusch:
This seems as an overkill. What is the use case where using the indexer is not sufficient? I don't thing we should add an extra API (that will need to be supported forever) unless this is a common use case that will be used by many.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -543,6 +543,20 @@ public override string ToString()
             }
         }
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Creates a new span over the RepeatedField.  This is the most
+        /// efficient way to read and write values to the RepeatedField because
+        /// there are no internal null or bounds checks.  You are responsible
+        /// for ensuring no null values are written to the span.
```
jtattermusch: I think that the lack of null check is dangerous and relying on users to always to the right thing is dangerous.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -543,6 +543,20 @@ public override string ToString()
...
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Creates a new span over the RepeatedField.  This is the most
+        /// efficient way to read and write values to the RepeatedField because
+        /// there are no internal null or bounds checks.  You are responsible
+        /// for ensuring no null values are written to the span.
+        /// </summary>
+        /// <returns>Span representation of the RepeatedField</returns>
+        public Span<T> AsSpan()
```
jtattermusch: Let's say I wanted to populate the repeated field with 1 million values. How do I resize the empty RepeatedField to the right size using the public API?

jtattermusch:
> > This seems as an overkill. What is the use case where using the indexer is not sufficient?
> 
> Bulk operations. My use case is adding lots of primitives to repeated fields to construct a response.
> In the ballpark of 7.5 million values across 500 repeated fields. 15k per field.
> 
> RepeatedField's indexer is 12x slower than a raw array, and 4.5x slower than Span.
> 
> | Raw array | 1.3ms |
> | Span | 3.3ms |
> | RepeatedField Indexer | 15.5ms |
> 
> On a related note, I need to look further into whether or not this will actually work with C# 6. I ran into some errors creating the benchmark and needed to update the language to 7, which the maintainers aren't ready to do yet.

Would an optimized version of AddRange() help you then? (have you tried using AddRange()?)

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
prat0088: C++ allows you to memcpy into RepeatedField.  But that's C++.  I suppose C# users expect the null checks and such.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
prat0088: I have another PR for that, but I will need to modify it slightly.  It preallocates the array but doesn't increase count, so, as you point out, it isn't going to work.  It will still be nice to have for reads.

jtattermusch: It would be good to first know what it the full API you're proposing. Isolated like this, it doesn't make much sense.
Btw, we just merged https://github.com/protocolbuffers/protobuf/pull/6317, so this PR might be out of date.

Falco20019:
I would also like to have support for `Span<T>`. I sometimes have to get ranges which is currently not easily possible with `RepeatedField<T>`. I understand that it should not be possible to manipulate it, so I would suggest offering `ReadOnlySpan<T>` instead.

```csharp
public ReadOnlySpan<T> this[Range range]
{
	get
	{
		(int start, int length) = range.GetOffsetAndLength(array.Length);
		return new ReadOnlySpan<T>(array, start, length);
	}
}
```

jtattermusch:
This PR is out of date now that https://github.com/protocolbuffers/protobuf/pull/7351 has been merged.

jtattermusch:
After a long time, there's still some unsolved issues with this PR (e.g. it cannot ensure that the values won't become null, there's no way of setting size when filling a repeated field, should we only allow read only access and provide a different way for populating the contents, etc.).
For these reason, I'm going to close this PR.
I'm not in principle opposed to providing a high-performance bulk accessor, but all the design questions need to be answered before we can add one. We cannot add a half-baked API.

jtattermusch:
Btw, a RepeatedField has a `AddRange(IEnumerable ` method which can be used for efficiently bulk-filling  a repeated field (if you know the size in advance).

Also note that exposing the data as read-only (`ReadOnlySpan<T>`) addresses some concerns over the safety, but there are still some potential issues - e.g.  the returned span is bound to the current `array`, but if a repeated field gets resized, a new internal array is allocated, so the span will point to a dead array and all the changes will be lost.


One idea for safe span-based initialization:
We could use a similar trick that is used for Creating a string from a Span:
https://docs.microsoft.com/en-us/dotnet/api/system.string.create?view=netcore-3.1
Similar trick could also work for bulk reading the contents of a repeated field, but doing that safely. (but then again, if you're doing a bulk read, why can't you just enumerate over the values? - GetEnumerator should be really fast)

nietras:
@jtattermusch has this PR stranded completely? What's the status?

I have issues with `RepeatedField` API design and perf too, and couldn't understand why this was not supported.

PR #6542, opened 2019-08-20
Update to Truth 1.0

elharo:


elharo:
@rafi-kamal Can you force the two remaining builds?

rafi-kamal:
@elharo Sorry, which remaining builds do you mean? Looks like all the Kokoro builds are passing now.

PR #6544, opened 2019-08-21
[javalite] Fix bzl and BUILD files by running buildifier

Yannic:
This change was automated by
`buildifier -lint=fix -warnings=all WORKSPACE $(find . -name "*.bzl" -o -name "BUILD" -o -name "*.BUILD")`


TeBoring:
This change is sent to the javalite branch instead of master. Is that intended?

Yannic:
Yes, that's intended. Similar changes have already been merged into master.

PR #6546, opened 2019-08-21
Down integrate internal changes to GitHub

rafi-kamal:


Protobuf-C++.podspec
```diff
@@ -1,42 +0,0 @@
-Pod::Spec.new do |s|
```
TeBoring: This file should not be deleted.
I guess we need to add this to some internal list.

build_files_updated_unittest.sh
```diff
@@ -1,62 +0,0 @@
-#!/bin/bash
```
TeBoring: This should not be deleted

conformance/binary_json_conformance_suite.cc
```diff
@@ -922,107 +895,106 @@ void BinaryAndJsonConformanceSuite::RunSuiteImpl() {
...
+                           {delim("\xfb"), delim("\xfb")},
+                           {delim(""), delim("")},
+                       });
   TestValidDataForType(FieldDescriptor::TYPE_ENUM, {
-    {varint(0), varint(0)},
-    {varint(1), varint(1)},
-    {varint(2), varint(2)},
-    {varint(-1), varint(-1)},
-  });
+                                                       {varint(0), varint(0)},
```
TeBoring: why so many spaces?

conformance/failure_list_js.txt
```diff
@@ -1,113 +1,13 @@
-Recommended.Proto2.ProtobufInput.ValidDataRepeated.BOOL.PackedInput.DefaultOutput.ProtobufOutput
```
TeBoring: I think you should pick the github change for this file

update_compatibility_version.py
```diff
@@ -1,57 +0,0 @@
-#!/usr/bin/env python
```
TeBoring: You also need to keep this

conformance/binary_json_conformance_suite.cc
rafi-kamal: That also looks odd to me. But unfortunately even if I remove the spaces, g4 fix brings them back.

conformance/failure_list_js.txt
rafi-kamal: Done

src/Makefile.am
```diff
@@ -213,6 +213,7 @@ libprotobuf_lite_la_SOURCES =                                  \
   google/protobuf/repeated_field.cc                            \
   google/protobuf/wire_format_lite.cc                          \
   google/protobuf/io/coded_stream.cc                           \
+  google/protobuf/io/coded_stream_inl.h                        \
```
TeBoring: This line needs to be removed. Please double check it's removed when you do up-integration.

rafi-kamal: Done

PR #6547, opened 2019-08-22
Optimization for layout_init()

haberman:
This avoids most of the work of layout_init() by creating a template object we can `memcpy()` from when initializing a new object.

There is a bit of complication since we must eagerly allocate maps and repeated fields. So we have to `dup` these still in `layout_init()`.

In the benchmark on my machine this takes us from 8 seconds to 2.3 seconds, or a 70% reduction in runtime overall.


samphippen:
Initially this looks good for ads protos, just gathering some data on our large data set now!

haberman:
The crash that I was seeing earlier has been fixed in the latest commit. This is ready for review.

Issue #6182, opened 2019-05-28
repeated field functional improvement


guopenglun:
**What language does this apply to?**
proto3
c#

**Describe the problem you are trying to solve.**
repeated Functional improvement

**Describe the solution you'd like**
by Default generate
`

        private readonly pbc::RepeatedField<MList> mlist_ = new pbc::RepeatedField<MList>();

        public pbc::RepeatedField<MList> Mlist {
            get { return mlist_; }
        }

`
I expect
`

        private readonly pbc::RepeatedField<MList> mlist_ = new pbc::RepeatedField<MList>();

        public pbc::RepeatedField<MList> Mlist {
            get { return mlist_; }
            set {
                mlist_.Clear();
                if (value != null)
                    mlist_.AddRange(value);
            }

        }

`
**Describe alternatives you've considered**
Can set the generate set option
**Additional context**

jtattermusch:
Closing for this reason: https://github.com/protocolbuffers/protobuf/pull/6549#issuecomment-526198163

PR #6549, opened 2019-08-22
C#: repeated field functional improvement

893949088:
#6182


893949088:
@googlebot I signed it!

893949088:
@BSBandme

893949088:
@jtattermusch I hope get any one reply about this pull request.

jtattermusch:
I think repeated fields not being settable is working as intended. I think there are scenarios when enabling the setter can cause unexpected results and therefore it's safer to only allow the getter.
What is the motivation for this change?

Issue #6493, opened 2019-08-09
Add Automatic-Module-Name header to jar files' manifests


elharo:
https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-20.md

```
Manifest-Version: 1.0
Bnd-LastModified: 1565047417349
Build-Jdk: 1.8.0_181-google-v7
Built-By: jieluo
Bundle-Description: Utilities for Protocol Buffers
Bundle-DocURL: https://developers.google.com/protocol-buffers/
Bundle-License: https://opensource.org/licenses/BSD-3-Clause
Bundle-ManifestVersion: 2
Bundle-Name: Protocol Buffers [Util]
Bundle-SymbolicName: com.google.protobuf.util
Bundle-Version: 3.9.1
Created-By: Apache Maven Bundle Plugin
Export-Package: com.google.protobuf.util;version="3.9.1";uses:="com.goog
 le.protobuf"
Import-Package: com.google.common.base;version="[26.0,27)",com.google.co
 mmon.io;version="[26.0,27)",com.google.common.math;version="[26.0,27)",
 com.google.common.primitives;version="[26.0,27)",com.google.gson;versio
 n="[2.7,3)",com.google.gson.stream;version="[2.7,3)",com.google.protobu
 f;version="[3.9,4)"
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
Tool: Bnd-3.0.0.201509101326
```

**What language does this apply to?**
Java 9+

**Describe the problem you are trying to solve.**
Make jars work better with modular Java code

**Describe the solution you'd like**

elharo:
fixed in https://github.com/protocolbuffers/protobuf/pull/6568

PR #6568, opened 2019-08-26
Java: Add Automatic-Module-Name entries to the Manifest

### LINKED ISSUE #3903: Any plans to modularize (or add Automatic-Module-Name properties to) protobuf-java?
Hello.

In order to play nicely with programs/libraries that have been modularized, it would be good if `Automatic-Module-Name` properties could be added to the `protobuf-java` jar files. This would give a stable name that can be used in the `requires` directives of modules.

http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html

> Do not release to Maven Central a modular jar file that depends on an automatic module, unless the automatic module has an "Automatic-Module-Name" MANIFEST.MF entry.

I'd submit a PR, but I'm not sure what module names you'd want to use. Presumably something like `com.google.protobuf.core` and `com.google.protobuf.lite`.

ennerf:
This PR adds `Automatic-Module-Name` entries to the Manifest files. This allows Java9+ users to depend on protobuf, without having any impact for users of earlier versions.

The module names match the OSGI symbolic names as recommended in [JLBP-20](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-20.md)

This fixes #3903 and #6493

### DIFF
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -144,6 +144,7 @@
         <extensions>true</extensions>
         <configuration>
           <instructions>
+            <Automatic-Module-Name>com.google.protobuf</Automatic-Module-Name> <!-- Java9+ Jigsaw module name -->
             <Bundle-DocURL>https://developers.google.com/protocol-buffers/</Bundle-DocURL>
             <Bundle-SymbolicName>com.google.protobuf</Bundle-SymbolicName>
             <Export-Package>com.google.protobuf;version=${project.version}</Export-Package>
--- a/java/lite/pom.xml
+++ b/java/lite/pom.xml
@@ -324,6 +324,7 @@
         <extensions>true</extensions>
         <configuration>
           <instructions>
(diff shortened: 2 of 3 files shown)


ennerf:
At the moment the `lite` version is a [subset](https://github.com/protocolbuffers/protobuf/blob/3.9.x/java/lite/pom.xml#L86-L174) of the regular artifact. Considering that both artifacts use the same package and filenames they are guaranteed to **always** conflict, and users should really avoid having both of them in the same project.

For example, an older `lite` artifact may load a utility class like `Utf8` that is incompatible with a newer version of the regular artifact.

The worst case would be if a dependency transitively adds the `lite-runtime`, but in that case users could [manually exclude it](https://www.viralpatel.net/maven-ignore-transitive-dependency/). Any generated `lite-messages` would still run fine against the regular runtime.

It may actually be preferrable to get a compile time error rather than potentially running into obscure issues at runtime.

elharo:
FWIW in Java 9 and later you will get a compile time error if both of these appear in the classpath.

PR #6576, opened 2019-08-26
Updated issue templates to remove question form

acozzette:
This commit removes the question template. It also updates the bug report and feature request templates to direct questions to the mailing list.


wangkirin:
I think it's good to isolate question issues from bug reports and feature requests.
Maybe we can point out directly to guide users  in someplace such as `README.md`

PR #6585, opened 2019-08-29
bazel: Change zlib URL to GitHub

keith:
We've notice significant issues downloading from zlib.net. Since there
are already other archives coming from GitHub, this shouldn't negatively
affect reliability.


keith:
Thanks!

PR #6588, opened 2019-08-29
Fix issues for php map when parsing missing key/value

TeBoring:
* For missing message value, map should create a default message
instance in order to keep its invariable.
* On 32-bit platform, int64 map key should be string

Integration test implemented in #6560, which will catch bugs fixed in this issue.

### DIFF
--- a/php/src/Google/Protobuf/Internal/MapEntry.php
+++ b/php/src/Google/Protobuf/Internal/MapEntry.php
@@ -32,12 +32,26 @@
 
 namespace Google\Protobuf\Internal;
 
+use Google\Protobuf\Internal\GPBType;
 use Google\Protobuf\Internal\Message;
 
 class MapEntry extends Message
 {
     public $key;
     public $value;
+
+    public function __construct($desc) {
+        parent::__construct($desc);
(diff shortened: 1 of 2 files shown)


php/src/Google/Protobuf/Internal/MapEntry.php
```diff
@@ -32,13 +32,27 @@
...
 {
     public $key;
     public $value;
 
+    public function __construct($desc) {
+        parent::__construct($desc);
+        // For MapEntry, getValue should always return a valid value. Thus, we
+        // need to create a default instance value if the value type is
+        // message, in case no value is provided in data.
+        $value_field = $desc->getFieldByNumber(2);
```
stanley-cheung: why the number 2 here specifically?

TeBoring: Map entry only has key (1) and value (2) field.
Besides, only value field can be message type.

TeBoring:
Test implemented in https://github.com/protocolbuffers/protobuf/pull/6560.
However, I want separate the fix for php out.

stanley-cheung:
Thanks LGTM

PR #6589, opened 2019-08-29
Split libprotoc by generator language

Yannic:
In a future version of Bazel, all languages will require
a protoc plugin instead of having some of them built
directly into protoc [1]. This change prepares Protobuf
for that by splitting //:protoc_lib into language-specific
targets. For other build-systems, this is effectively a no-op.

By exporting all language-specific targets as //:protoc_lib,
this remains fully compatible with previous versions.

[1] https://docs.google.com/document/d/1u95vlQ1lWeQNR4bUw5T4cMeHTGJla2_e1dHHx7v4Dvg/edit

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -319,11 +319,27 @@ cc_proto_library(
 ################################################################################
 
 cc_library(
-    name = "protoc_lib",
+    name = "protoc_base",
     srcs = [
-        # AUTOGEN(protoc_lib_srcs)
+        # AUTOGEN(protoc_base_srcs)
         "src/google/protobuf/compiler/code_generator.cc",
         "src/google/protobuf/compiler/command_line_interface.cc",
+        "src/google/protobuf/compiler/plugin.cc",
+        "src/google/protobuf/compiler/plugin.pb.cc",
+        "src/google/protobuf/compiler/subprocess.cc",
+        "src/google/protobuf/compiler/zip_writer.cc",
+    ],
+    copts = COPTS,
+    includes = ["src/"],
+    linkopts = LINK_OPTS,
+    visibility = ["//visibility:public"],
+    deps = [":protobuf"],
+)
+
+cc_library(
+    name = "protoc_cpp",
+    srcs = [
+        # AUTOGEN(protoc_cpp_srcs)
         "src/google/protobuf/compiler/cpp/cpp_enum.cc",
         "src/google/protobuf/compiler/cpp/cpp_enum_field.cc",
         "src/google/protobuf/compiler/cpp/cpp_extension.cc",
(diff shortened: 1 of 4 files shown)


acozzette:
@Yannic Could you provide some more context on this change? I don't fully understand why each language will need its own plugin. It seems to me that we could move the LANG_proto_library() rules to Starlark without necessarily needing to break each language into its own plugin.

Yannic:
You're right that it's possible to implement `<lang>_proto_library` rules in Starlark without requiring that every language needs a plugin. However, this has at least two major disadvantages:
  - It makes the generator API proposed in the Google Doc linked above a lot more complicated to implement, but also to use. If we distinguish between built-in generators and plugins, there are lots of special cases we have to handle. If all generators are plugins, everything's nice and easy (tm).
  - I don't think it's right to assume that everyone uses the built-in generators. For example, there are lots of third_party alternatives for build-in languages listed in `/third_party.md`. Supporting these would require that every `<lang>_proto_library` rule also needs support for plugins, regardless of whether they'd normally use built-in generators or not. This forces all the special cases to be not only handled by the generator framework but also to every `<lang>_proto_library` rule, making them more complicated, hence more likely to contain bugs.

I do understand that making every generator a plugin has a maintenance cost for you, but given that it's going to be Bazel-only, I think the overhead will be very very close to zero. In addition to this change, all we'll do is to add `#ifndef` to `protoc_main.cc` to not include the generators in the binary and create `.../compiler/<lang>/plugin_main.cc` with nothing but a `main` that creates a generator object and calls `PluginMain`. Again, there's no additional artifact you'd have to build as part of the release process. Everyone else stays on the existing `protoc` binary.

acozzette:
@Yannic It seems like another simple alternative would be to build the Skylark logic with the assumption that every language has a plugin, but the "plugin" for a built-in language like C++ might be nothing but a thin wrapper around protoc. I'm not sure that's necessarily a good idea but would be one option to consider if we want to keep the built-in generators the way they are.

Would it be OK to put this PR on hold for now just to wait for more comments on your design doc? I think this change should probably be fine, especially since it only affects Bazel, but it would be good to just make sure everyone's onboard with the direction.

Yannic:
I'll have to think whether using thin whappers around protoc as plugins will work.

Putting this PR on hold for now is fine. Would you mind commenting on the relevant paragraph (page 4, right below the `proto_common.generate` API) that Protobuf is probably fine with this change, and maybe put some more Protobuf folks into the loop to make sure everyone's on board?

Yannic:
@acozzette Can we move forward with this? The design doc got approved now.

hlopko:
I think we should not require each language to be a separate plugin. That will make distributing protoc harder, many people want to use the protoc provided by their distribution etc. We currently build protoc with rules_proto but that should be configurable. Therefore I think this PR is not necessary.

I didn't realize that this PR is a requirement of the design doc. @Yannic, would you mind elaborating what complications are associated with some languages not requiring the plugin in the doc? Let's continue that thread there. Maybe I incorrectly assumed that `proto_lang_toolchain` having optional 'plugin' attribute is all there is to use a third party generator plugin for built-in languages. But let's continue on the design doc.

acozzette:
I'm going to go ahead and close this pull request since there has been no activity in the past year, but let me know if it should be reopened.

Issue #5124, opened 2018-09-07
We are considering deprecating autotools in favor of CMake


haberman:
We are considering deprecating autotools builds. CMake has wide platform support, and we need to support the CMake build for systems like Windows that don't support autotools.

Please let us know if this would cause major problems for you.

Before we deprecated autotools, there is a fair amount of logic we would need to port to CMake. For example:

- conformance tests
- benchmarks
- platform detection for Objective C (Objective C only builds on macOS)
- various scripts that call "make"

BSBandme:
OK, if we decide to do that, I'll put benchmark's cleaning into next q's okr

forskning:
How much of the respective audience would be utilising Windows native, MSVC or Git for Windows with MSYS2, as opposed to WSL?

Has ninja (optionally re2c), python3, and meson, been considered?

haberman:
@forskning  CMake works with MVSC, and I would expect that it works with UNIX-like environments on Windows like WSL or MSYS2 also?

CMake can compile to Ninja. That is how I usually use CMake.

I don't think we want to create a new build system in meson.  We already have a CMake build to support Windows. I want to deprecate autotools support in favor of CMake everywhere.

forskning:
I only mentioned ninja (optionally re2c), and python3, as meson dependencies.

https://github.com/python/cpython
Python version 3.8.0 alpha 0

FWIW, downloading cpython-master.zip, extracting it, and running
'grep -r %EXTERNAL_DIR% cpython-master' returns a result 
for cpython-master/PCbuild/find_python.bat, and running thereafter
'grep -r %EXTERNALS_DIR% cpython-master' returns results in both
the cpython-master/PCbuild and cpython-master/Tools directories.

https://github.com/python/cpython/commit/68d663cf85d1ac5eaf83482eed39c0a6f8093601#diff-60f6a0e6e49e5990c3ad465771fff9cd

committed on GitHub Jul 17, 2017

So I wonder was there a typo in the above commit, as "%EXTERNAL_DIR%" rather than "%EXTERNALS_DIR%"?

PCbuild/find_python.bat

lines 27-28

I'm not on the Python mailing list so if someone could clarify whether a typo occurred or not it would be appreciated.

http://nurmi-labs.blogspot.com/2017/03/p9p.html
http://nurmi-labs.blogspot.com/2017/04/xs.html

A couple of links regarding Win10 (Anniversary Update circa) WSL.

[...]

PR #6590, opened 2019-08-29
Update GoogleTest

Yannic:
This updates GoogleTest to include load statements for `@rules_cc`, which will be required by a future version of Bazel.


acozzette:
@Yannic It looks like this somehow causes `make distcheck` to stop working and that is causing some CI test failures.

Yannic:
Seems like `make` support for GoogleTest got removed in https://github.com/google/googletest/commit/6b8c13815496df183e2247a9375c8dfedefdddc2

acozzette:
That's interesting, we have also been thinking about removing our autotools build support (#5124).

acozzette:
Let me go ahead and close this. I appreciate the change and it would be great to upgrade GoogleTest, but unfortunately we can't do that until we figure out a solution for autotools support (which has been removed in recent GoogleTest versions).

PR #6610, opened 2019-09-04
Fix small style issues in the dart example, following Effective Dart

rkj:
- use final instead of obvious types for local variables
- add missing `void` return type
- make the comments into doccomments (///)


acozzette:
@rkj Thanks for the fixes.

Issue #6419, opened 2019-07-20
Compile warning C++ - Ubuntu


mohsenomidi:
**What version of protobuf and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Ubuntu 18.04.1 LTS (GNU/Linux 4.15.0-54-generic x86_64)

**What runtime / compiler are you using (e.g., python version or gcc version)**
gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04.1)

**What did you do?**
by compilation manual start compiling the project

**What did you expect to see**
Successful compilation
 
**What did you see instead?**
Compile Warning 

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

  CXX      google/protobuf/implicit_weak_message.lo
  CXX      google/protobuf/message_lite.lo
In file included from /usr/include/string.h:494:0,
                 from ./google/protobuf/stubs/port.h:38,
                 from ./google/protobuf/stubs/common.h:46,
                 from ./google/protobuf/message_lite.h:45,
                 from google/protobuf/message_lite.cc:36:
In function ‘void* memcpy(void*, const void*, size_t)’,
    inlined from ‘google::protobuf::uint8* google::protobuf::io::EpsCopyOutputStream::WriteRaw(const void*, int, google::protobuf::uint8*)’ at ./google/protobuf/io/coded_stream.h:697:16,
    inlined from ‘virtual google::protobuf::uint8* google::protobuf::internal::ImplicitWeakMessage::InternalSerializeWithCachedSizesToArray(google::protobuf::uint8*, google::protobuf::io::EpsCopyOutputStream*) const’ at ./google/protobuf/implicit_weak_message.h:92:35,
    inlined from ‘bool google::protobuf::MessageLite::SerializePartialToZeroCopyStream(google::protobuf::io::ZeroCopyOutputStream*) const’ at google/protobuf/message_lite.cc:423:51:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:34:71: warning: ‘void* __builtin___memcpy_chk(void*, const void*, long unsigned int, long unsigned int)’: specified size between 18446744071562067968 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
   return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));
                                                                       ^
  CXXLD    libprotobuf-lite.la
[...]

PR #6617, opened 2019-09-05
Fix compile warning

wangkirin:
This PR fix compile warning refer in issue #6419.

### DIFF
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -690,8 +690,8 @@ class PROTOBUF_EXPORT EpsCopyOutputStream {
     }
   }
 
-  uint8* WriteRaw(const void* data, int size, uint8* ptr) {
-    if (PROTOBUF_PREDICT_FALSE(end_ - ptr < size)) {
+  uint8* WriteRaw(const void* data, size_t size, uint8* ptr) {
+    if (PROTOBUF_PREDICT_FALSE(end_ - ptr < static_cast<long int>(size))) {
       return WriteRawFallback(data, size, ptr);
     }
     std::memcpy(ptr, data, size);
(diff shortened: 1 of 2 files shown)


acozzette:
@wangkirin Thank you for the fix but I'm a little bit hesitant to merge this because I am not very familiar with the details of this particular code. Let's wait for @gerben-s to to take a look after he gets back from vacation next week.

wangkirin:
> @wangkirin Thank you for the fix but I'm a little bit hesitant to merge this because I am not very familiar with the details of this particular code. Let's wait for @gerben-s to to take a look after he gets back from vacation next week.

No problem ,  I will wait for @gerben-s

src/google/protobuf/io/coded_stream.h
```diff
@@ -690,8 +690,8 @@ class PROTOBUF_EXPORT EpsCopyOutputStream {
     }
   }
 
-  uint8* WriteRaw(const void* data, int size, uint8* ptr) {
-    if (PROTOBUF_PREDICT_FALSE(end_ - ptr < size)) {
+  uint8* WriteRaw(const void* data, size_t size, uint8* ptr) {
+    if (PROTOBUF_PREDICT_FALSE(end_ - ptr < static_cast<long int>(size))) {
```
gerben-stavenga: ptrdiff_t

0x8000-0000:
Ping @acozzette  @gerben-s ?

acozzette:
Before we can merge this I think we just need Gerben's comment about ptrdiff_t to be addressed, and the merge conflict also needs to be fixed.

OSHW-Rico:
Do you need any more information? Just encountered the same warning.

ChisholmKyle:
I also encountered this warning.

ao2:
Hi, I see this has been closed without explanation. Was the change rejected?

acozzette:
I think this change would be OK to merge but as @gerben-s said, the type should be ptrdiff_t instead of long int.

Arfrever:
This patch fixes this particular compilation warning, but this change seems to be incomplete as proper solution:
`WriteRaw()` function would receive `size_t size` argument and pass it to `WriteRawFallback()` function, which still takes `int size` argument.
(On x86_64 Linux platform, `int` has size 4 bytes = 32 bits, while `size_t` has size 8 bytes = 64 bits.)

There are 34 instances of `int size` in `src/google/protobuf/io/coded_stream.h` file and maybe many or all of them should be changed to `size_t` (the same for definitions in `src/google/protobuf/io/coded_stream.cc`).

PR #6618, opened 2019-09-05
Cherry Pick (#6614)

TeBoring:
* Sync upb

https://github.com/protocolbuffers/upb/pull/209

* Update upb

https://github.com/protocolbuffers/upb/pull/210


rafi-kamal:
Should we also update the release notes?

TeBoring:
I think this is a internal bug fix. No need to update release note.

PR #6620, opened 2019-09-05
Update version number to 3.10.0 rc1. 

anandolee:
To keep master always have newer version to the released version, update the master version to next rc1.


TeBoring:
still needed?

rafi-kamal:
I thought we are just going to append SNAPSHOT suffix for Java versions?

PR #6644, opened 2019-09-13
Ignore RuntimeWarning from description_database.py in tests

Code0x58:
This cuts out what is currently 209 lines of ~junk spat out while the tests are running:
```
/protobuf/python/google/protobuf/descriptor_database.py:157: RuntimeWarning: Conflict register for file "other_file": protobuf_unittest.TestAllTypes is already defined in file "google/protobuf/unittest.proto"
```
The warning is already explicitly tested for and is unaffected by this change.

### DIFF
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -11,6 +11,7 @@ setenv =
     cpp: DYLD_LIBRARY_PATH={toxinidir}/../src/.libs
     cpp: PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=cpp
     python: PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
+    PYTHONWARNINGS=ignore::RuntimeWarning:google.protobuf.descriptor_database
 commands =
     python setup.py -q build_py
     python: python setup.py -q build


acozzette:
This looks like a good change to me, but I'm puzzled by the Python test failures:
```
======================================================================
ERROR: testSingleScalarTypeSafety_proto2 (google.protobuf.internal.reflection_test.ReflectionTest)
testSingleScalarTypeSafety_proto2(<module 'google.protobuf.unittest_pb2' from '/tmp/protobuf/protobuf/python/google/protobuf/unittest_pb2.py'>)
----------------------------------------------------------------------
  File "/tmp/protobuf/protobuf/python/google/protobuf/internal/_parameterized.py", line 229, in BoundParamTest
    test_method(self, *testcase_params)
  File "/tmp/protobuf/protobuf/python/google/protobuf/internal/reflection_test.py", line 384, in testSingleScalarTypeSafety
    proto.optional_bool = 1.1
TypeError: 1.1 has type float, but expected one of: int, long, bool
```
I don't understand why the test failures are happening only with the Python-C++ implementation for Python 3.8 and 3.9 (not 3.7 for example). 

If `proto.optional_bool = 1.1` is raising an error then that is actually a good thing because that's the desired behavior. But the test doesn't expect this to happen for Python versions before 3.10: https://github.com/protocolbuffers/protobuf/blob/8495372e1eb78a385f53bbc562c8935777107ce9/python/google/protobuf/internal/reflection_test.py#L384

@anandolee Do you have any ideas on what is going on? Maybe we should tweak the test so that it allows a `TypeError` for earlier Python versions?

acozzette:
Side note: the failing Java tests are not a problem. We deleted those test runs but there is no way to make the existing test results go away without creating a new PR.

anandolee:
Yes we can change the test to allow TypeError for earlier Python versions.

anandolee:
Have submitted a change in internal to catch the warnings in the test. Future releases will not have show those warnings

PR #6647, opened 2019-09-13
Update protobuf version to 3.10.0 stable

rafi-kamal:


rafi-kamal:
Linkage monitor seems to be broken because https://github.com/protocolbuffers/protobuf/pull/6628 wasn't merged in this branch. Shouldn't be an issue.

PR #6651, opened 2019-09-13
Add license

TeBoring:


acozzette:
I think it might be good to do some more research on the JsonCpp license issue. If there's a code similarity then it's hard to know for sure whether JsonCpp borrowed the code from Upcaste or Upcaste borrowed it from JsonCpp. Maybe it would be good to send a PR to the JsonCpp project and see if they accept it? That way our copy of JsonCpp would also match the upstream.

PR #6656, opened 2019-09-16
Simplify template exporting macros

m0hamed:


m0hamed:
This based on feedback from https://github.com/protocolbuffers/protobuf/pull/6535#issuecomment-530993531

PR #6661, opened 2019-09-18
Cast size_t type to int type explicitly

EFanZh:
Currently, the generated header file might cause warnings if user specifies `-Wconversion` compiler option. This commit should be able to fix it.

### DIFF
--- a/src/google/protobuf/any.pb.cc
+++ b/src/google/protobuf/any.pb.cc
@@ -221,7 +221,7 @@ failure:
   // string type_url = 1;
   if (this->type_url().size() > 0) {
     ::PROTOBUF_NAMESPACE_ID::internal::WireFormatLite::VerifyUtf8String(
-      this->_internal_type_url().data(), static_cast<int>(this->_internal_type_url().length()),
+      this->_internal_type_url().data(), this->_internal_type_url().length(),
       ::PROTOBUF_NAMESPACE_ID::internal::WireFormatLite::SERIALIZE,
       "google.protobuf.Any.type_url");
     target = stream->WriteStringMaybeAliased(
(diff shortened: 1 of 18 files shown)


acozzette:
I think it might be better to just change VerifyUtf8String to take a size_t instead of an int. Could you try making that change?

EFanZh:
I have observed that generally Protocol Buffers codes like to use `int` type to indicate the size of something, rather than `size_t`. I guess it is for performance reason. See the following codes:

https://github.com/protocolbuffers/protobuf/blob/763c3588740b97e8e80b1b1a1a2dc4f417647133/src/google/protobuf/message_lite.cc#L264-L331

and here:

https://github.com/protocolbuffers/protobuf/blob/763c3588740b97e8e80b1b1a1a2dc4f417647133/src/google/protobuf/map_field.h#L105

Is this some kind of internal convention in Protocol Buffers? Is is OK for me to change the signature?

acozzette:
Yeah, it's kind of a weird quirk of the Google C++ style, see [here](https://google.github.io/styleguide/cppguide.html#Integer_Types) for some more details. Theoretically signed integers can be faster in some cases because the compiler doesn't have to handle the possibility of overflow.

I don't think we want to go through the codebase and switch all sizes to `size_t`, but in this one case it is probably fine.

EFanZh:
I have found there are a lot of dependencies on the UTF-8 verification functions. Changing the function signatures leads to a lot of other changes. Nevertheless, I have updated my pull request. If this change is too much, I can revert my second commit.

acozzette:
@EFanZh Could you go back to your original PR with just the first commit? Sorry to change my mind but I didn't realize it would involve updating so much code.

PR #6664, opened 2019-09-18
Add imenu support for Emacs protobuf-mode

kaiwk:
Hello,

I found `protobuf-mode` doesn't utilize `imenu`. `imenu` is a framework for Emacs to index specific major mode buffer.

This patch will enable `protobuf-mode` to make use of  `imenu`.

屏幕快照 2019-09-18 下午8 59 59


kaiwk:
Close for #7620.

PR #6672, opened 2019-09-19
[ObjC] Don't use unions and instead use memcpy for the type swaps.

thomasvl:
The code in question hasn't change in a long time so there likely is a compiler
change that is causing this to suddenly be an issue. memcpy should be a better
code pattern in general.

Should address https://github.com/firebase/firebase-ios-sdk/issues/3851


paulb777:
This still crashes
Screen Shot 2019-09-19 at 9 29 20 AM

thomasvl:
@paulb777 which Xcode version(s) and can you share your repo I can poke at it more directly?

PR #6674, opened 2019-09-20
Cleanup various bits of Google.Protobuf

ObsidianMinor:
While adding nullable annotations throughout the lib, I found a few missing null-checks and some IsInitialized checks accidentally left in MessageParser. Additionally, ExtensionCollection.CrossLink did some unnecessary things so this simplifies that.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
+++ b/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
@@ -261,9 +261,11 @@ namespace Google.Protobuf
             Assert.True(message.IsInitialized());
         }
 
-        // Code was accidentally left in message parser that threw exceptions when missing required fields after parsing.
-        // We've decided to not throw exceptions on missing fields, instead leaving it up to the consumer how they
-        // want to check and handle missing fields.
+        /// <summary>
+        /// Code was accidentally left in message parser that threw exceptions when missing required fields after parsing.
+        /// We've decided to not throw exceptions on missing fields, instead leaving it up to the consumer how they
+        /// want to check and handle missing fields.
+        /// </summary>
         [Test]
         public void RequiredFieldsNoThrow()
         {
--- a/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
+++ b/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
@@ -261,11 +261,14 @@ namespace Google.Protobuf
             Assert.True(message.IsInitialized());
         }
 
+        // Code was accidentally left in message parser that threw exceptions when missing required fields after parsing.
+        // We've decided to not throw exceptions on missing fields, instead leaving it up to the consumer how they
+        // want to check and handle missing fields.
         [Test]
         public void RequiredFieldsNoThrow()
         {
-            TestRequired.Parser.ParseFrom(new byte[0]);
-            (TestRequired.Parser as MessageParser).ParseFrom(new byte[0]);
+            Assert.DoesNotThrow(() => TestRequired.Parser.ParseFrom(new byte[0]));
+            Assert.DoesNotThrow(() => (TestRequired.Parser as MessageParser).ParseFrom(new byte[0]));
         }
 
         [Test]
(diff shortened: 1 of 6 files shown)


ObsidianMinor:
Could you possibly review this @jtattermusch?

jtattermusch:
@ObsidianMinor  there's a conflict

ObsidianMinor:
Fixed

jtattermusch:
Is this purely cosmetic or does it actually fix any user-visible bugs?

csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -72,7 +72,6 @@ public IMessage ParseFrom(byte[] data)
         {
             IMessage message = factory();
             message.MergeFrom(data, DiscardUnknownFields, Extensions);
-            CheckMergedRequiredFields(message);
```
jtattermusch: what does this fix, is there a test that demonstrates the wrong behavior?

csharp/src/Google.Protobuf/Reflection/ExtensionCollection.cs
```diff
@@ -107,13 +107,11 @@ internal void CrossLink()
             {
                 descriptor.CrossLink();
 
-                IList<FieldDescriptor> _;
-                if (!declarationOrder.TryGetValue(descriptor.ExtendeeType, out _))
-                {
-                    declarationOrder.Add(descriptor.ExtendeeType, new List<FieldDescriptor>());
-                }
+                IList<FieldDescriptor> list;
+                if (!declarationOrder.TryGetValue(descriptor.ExtendeeType, out list))
```
jtattermusch: - in protobuf codebase, always use curly brackets for `if` statements (single statements ifs and loops are considered dangerous).
- I think 
```
list = new List<FieldDescriptor>();
declarationOrder.Add(......., list);
```
is more readable.
```

File: csharp/src/Google.Protobuf/MessageParser.cs (hunk 1 above)
Comment (ObsidianMinor) [reply csharp/src/Google.Protobuf/MessageParser.cs]: I can write one. The wrong behavior can only be seen if you parse using the non-generic methods.

```cs
var foo = Foo.Parser.ParseFrom(data); // doesn't throw if missing required
foo = (Foo.Parser as MessageParser).ParseFrom(data); will throw if missing required
```

csharp/src/Google.Protobuf/Reflection/ExtensionCollection.cs
ObsidianMinor: Could we add an editorconfig with `csharp_prefer_braces` enabled to prevent style issues like this and others in the future?

csharp/src/Google.Protobuf/MessageParser.cs
jtattermusch: Adding a test would be useful.

csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
```diff
@@ -261,6 +261,13 @@ public void RequiredFields()
             Assert.True(message.IsInitialized());
         }
 
+        [Test]
+        public void RequiredFieldsNoThrow()
+        {
+            TestRequired.Parser.ParseFrom(new byte[0]);
```
jtattermusch: without context, it's not obvious what this test does and what is being checked. 
can we add comment or improve clarity?
Also consider using `Assert.DoesNotThrow` to make it obvious what is it that you're testing.

csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
```diff
@@ -261,6 +261,16 @@ public void RequiredFields()
             Assert.True(message.IsInitialized());
         }
 
+        // Code was accidentally left in message parser that threw exceptions when missing required fields after parsing.
```
jtattermusch: thanks for adding the comment.
nit: Use `/// <summary>` block for comments outside the method body.

csharp/src/Google.Protobuf/Reflection/ExtensionCollection.cs
jtattermusch: At some point, I'd like to add a set of rules to enforce code style in the protobuf C# codebase. I'm not sure what the best approach is (there are multiple ways of doing this).

PR #6675, opened 2019-09-20
Remove win64 conversion warnings in generated code

chrisse74:
When compiling protoc generated code in 64bit windows, we get conversion warnings.

The reason being, that the size method of std::string returns a size_t, while VerifyUtf8String accepts an int.

```
"  static bool ValidateKey(std::string* s) {\n"
             "    return ::$proto_ns$::internal::WireFormatLite::"
-            "VerifyUtf8String(s->data(), s->size(), "
```

Motivated by _cpp_string_field.cc(call of GenerateUtf8CheckCodeForString)_ this PR solves the issue by static-casting the return value of the size method to int.

By approving this PR protobuf will get rid of these warnings:

```
File: XyzProto.pb.h, Line: xxx
Message: 'argument': conversion from 'size_t' to 'int', possible loss of data

```


chrisse74:
I just noticed, that #6661 already aims at solving this, therefore i will close this PR.

PR #6685, opened 2019-09-23
[Ruby] Add Google::Protobuf::Timestamp.from_time

DerekStride:
Previously, you would have to instantiate a timestamp class then call #from_time on that object to mutate its state. Instead provides a mechanism to instantiate the object with the correct state.

### DIFF
--- a/ruby/lib/google/protobuf/well_known_types.rb
+++ b/ruby/lib/google/protobuf/well_known_types.rb
@@ -82,6 +82,10 @@ module Google
         end
       end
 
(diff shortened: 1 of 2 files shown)


jufemaiz:
I can only assume that this was abandoned @DerekStride ?

DerekStride:
I closed it because of a lack of involvement from the maintainers but I still think this is a valuable change

jufemaiz:
Ok sweet. See #8254 (also expands to resolve similar issue with other well known types)

PR #6687, opened 2019-09-24
Fix clang-format warnings

tapaswenipathak:
Heyo folks! 

[clang-format](https://clang.llvm.org/docs/ClangFormat.html) is great tool for fixing the clang-format warnings and have been used for many open source projects. I ran clang-format across protobuf, hope you find my pr useful.


acozzette:
Thanks for sending us this PR but I'm afraid we're a bit picky about formatting. For pointers and references we like to put the `*` or `&` next to the type instead of the variable name, like `T* t` for example instead of `T *t`. We usually put two spaces before a single-line (`//`) comment.

tapaswenipathak:
np.

tapaswenipathak:
No problem. :)

PR #6688, opened 2019-09-24
Remove negative values from enumField grammer

tapaswenipathak:
Fixes https://github.com/protocolbuffers/protobuf/issues/4561 I think.


acozzette:
@tapaswenipathak Thanks for the pull request but I think in this case the code is right and the documentation is wrong. We need to update the documentation to clarify that negative enum values are OK.

tapaswenipathak:
oh ok.

Issue #6271, opened 2019-06-17
Test against current cuda version, to ensure c++ feature compatibility


Fleischner:
Currently cuda (10.1) compiler nvcc still has problems with constexpr. We get errors when building pytorch with the current protobuf master. Pytorch uses therefore an older version with some patches, but the version inconsistency generates other problems.

It would be great if protobuf could hide newer c++ construcs behind an #define until cuda nvcc compiler gets compatible.

Currently there are two places:

cpp_message.cc
static constexpr int kIndexInFileMessages =\n"
==> static int const kIndexInFileMessages =\n"

extension_set.h
static constexpr uint16 kMaximumFlatCapacity = 256;
==> static uint16 const kMaximumFlatCapacity = 256;

haon4:
Feel free to send us a PR for this. We don't have much experience with the cude compiler so it is difficult to verify our changes.

PR #6689, opened 2019-09-24
Test against current cuda version, to ensure c++ feature compatibility

tapaswenipathak:
Should fix https://github.com/protocolbuffers/protobuf/issues/6271.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -1220,7 +1220,12 @@ void MessageGenerator::GenerateClassDefinition(io::Printer* printer) {
       "  return reinterpret_cast<const $classname$*>(\n"
       "             &_$classname$_default_instance_);\n"
       "}\n"
+	  #ifdef __CUDACC__
+	  //nvcc
+	  "static constexpr int kIndexInFileMessages =\n"
+	  #else
       "static constexpr int kIndexInFileMessages =\n"
+	  #endif
       "  $1$;\n"
       "\n",
       index_in_file_messages_);
--- a/src/google/protobuf/extension_set.h
+++ b/src/google/protobuf/extension_set.h
@@ -664,7 +664,10 @@ class PROTOBUF_EXPORT ExtensionSet {
   // Grows the flat_capacity_.
   // If flat_capacity_ > kMaximumFlatCapacity, converts to LargeMap.
   void GrowCapacity(size_t minimum_new_capacity);
+  #ifndef __CUDACC__
   static constexpr uint16 kMaximumFlatCapacity = 256;
+  #warning expression not supported in CUDA
+  #endif
   bool is_large() const { return flat_capacity_ > kMaximumFlatCapacity; }
 
   // Removes a key from the ExtensionSet.


acozzette:
What's the problem with nvcc and constexpr? Does nvcc error out when it sees the constexpr keyword? Maybe in this case we could just replace constexpr with const.

tapaswenipathak:
@acozzette: Ref for errors: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#constexpr-variables.

\+ review ping.

src/google/protobuf/extension_set.h
```diff
@@ -664,7 +664,11 @@ class PROTOBUF_EXPORT ExtensionSet {
   // Grows the flat_capacity_.
   // If flat_capacity_ > kMaximumFlatCapacity, converts to LargeMap.
   void GrowCapacity(size_t minimum_new_capacity);
+  #ifndef __CUDACC__
```
acozzette: I think this condition is backwards, shouldn't it be ifdef instead of ifndef?

tapaswenipathak: yep.

src/google/protobuf/compiler/cpp/cpp_message.cc
```diff
@@ -1220,7 +1220,12 @@ void MessageGenerator::GenerateClassDefinition(io::Printer* printer) {
       "  return reinterpret_cast<const $classname$*>(\n"
       "             &_$classname$_default_instance_);\n"
       "}\n"
+	  #ifdef __CUDACC__
```
acozzette: Actually this might not work, because this ifdef will only affect protoc, which is just a plain binary unrelated to Cuda. For it to work it would probably have to go in the generated code.

acozzette:
I still don't quite understand what the error is with Cuda. I see the docs say "Device source code cannot contain a reference to V or take the address of V." Is that the problem? Maybe we can refer to these constexpr variables but we can't take a pointer or reference to them?

tapaswenipathak:
> I still don't quite understand what the error is with Cuda. I see the docs say "Device source code cannot contain a reference to V or take the address of V." Is that the problem? Maybe we can refer to these constexpr variables but we can't take a pointer or reference to them?

Constant variables in C++ programming language are allocated in data segment of Random Access Memory. 

The variable `kIndexInFileMessages` at line 
- https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/cpp/cpp_message.cc#L1424

The variable `kMaximumFlatCapacity` at line 
- https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/extension_set.h#L672
- https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/extension_set.h#L840-L846

I don't think I should spend time in explaining you for what these variables are. 

Possible errors and warnings:

- warning: "constexpr" is ignored here in Microsoft mode
- error: function ref must have a constant value in a constant expression

and more.

The ticket says the error is generated when you would run protobuf master in pytorch, which much strictly optimizes the code per compiler. 

Let's see the documentation which says -

> Let 'V' denote a namespace scope variable or a class static member variable that has const qualified type and does not have execution space annotations (e.g., __device__, __constant__, __shared__). V is considered to be a host code variable.

> The value of V may be directly used in device code, if:
> V has been initialized with a constant expression before the point of use, and
> it has one of the following types:
> builtin floating point type except when the Microsoft compiler is used as the host compiler,
> builtin integral type.

> Device source code cannot contain a reference to V or take the address of V.

When the code will be run w/ pytorch, the code will break at every point and not just the last line.

You might require a read of `--expt-relaxed-constexpr` in CUDA.  This flag allows the device code in CUDA to host constexpr functions. Unfortunately, the CUDA compiler (at least for CUDA 9.0, 9.1, 10.1) isn't compatible with many of the constexpr things for which the device code compiler crashes when that sees one of these host-only functions. The code only works when nvcc builds host code, but not when it builds device code and notices the use of constexpr functions from device code.

The best correction would be using `C10_HOST_CONSTEXPR`  instead of constexpr. This enables constexpr when compiled on the host and applies, constexpr would not be present anymore and should be enough to stop that from being called from the device code and fully hidden. There is no other fix.

The implementation fix I have suggested also works, but the implementation will fail in MSVC.

In simple terms, hiding the bad + unsupported is the best and only way.

acozzette:
@tapaswenipathak Thanks for all the info. I think if we want to fix this we should probably just replace constexpr with const in the two places, since otherwise the ifdefs add a fair amount of boilerplate (especially since some will have to in the generated code). Do you have any way to confirm that this fixes the Pytorch problem?

Issue #5984, opened 2019-04-02
Undefined behavior in UTF8GenericScanFastAscii


monate:
Hello,
On this line https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/stubs/structurally_valid.cc#L507 the computation of

` const uint8* srclimit8 = srclimit - 7;`

may cause a violation of the C++ standard as `srclimit - 7` is not a valid operation when `srclimit` points, for example, to the beginning of a string.
Hope this helps,
Benjamin

PR #6690, opened 2019-09-24
Fix: Undefined behavior in UTF8GenericScanFastAscii

tapaswenipathak:
Fixes https://github.com/protocolbuffers/protobuf/issues/5984.

Pl. let me know if you prefer different error handling.


acozzette:
@tapaswenipathak Thanks for the fix but we need a solution that doesn't use exceptions (we don't use exceptions in the C++ codebase). I think it's also valid for the string to be less than 7 bytes long, so we need to handle that case.

tapaswenipathak:
Yea, perfect. That's :fire: :slightly_smiling_face:

I will write the code fix.

tapaswenipathak:
Yea, right, ko; that's a unique item I missed, lol.

I will write the code fix.

tapaswenipathak:
@acozzette: 

There are few options
- Use https://en.cppreference.com/book/intro/smart_pointers.
- Append null characters in the string, make sure the string size atleast 7
- Check for dangling pointer
- Add a log statement
- Throw an exception

Do you have any notes/thoughts?

acozzette:
Actually I think there is an easy way to solve it. If there are fewer than 7 bytes in the array then srclimit8 should just point to the beginning of the array. I haven't tested this but I believe something like this would work:

`const uint8* srclimit8 = str_length < 7 ? isrc : srclimit - 7;`

tapaswenipathak:
Agreed.

acozzette:
Thanks, @tapaswenipathak!

PR #6692, opened 2019-09-24
Remove unused function and add more UTs

bencsikandrei:
When browsing around the strutil files I found a function
that was never referenced inside the code base (but might be used by others since it's exported):

`void StripString(string* s, const char* remove, char replacewith)`

The name was kind of **misleading** as well and it seems like
it's a carbon copy of:

`void ReplaceCharacters(string* s, const char* remove, char replacewith);`

(even the parameter names are the same, the code is 100% the same.., makes me think it might have been a merge conflict gone wrong?)

Is it intentional? Maybe for compatibility reasons? If so,
let's make it deprecated and use the ReplaceCharacters method inside
or the other way around?

Also, noticed there were no tests for "StripString" or "Replace".
Added some for both. This might help with refactoring.

In any case, please do tell if this is something interesting or we should leave it be.
I'll continue browsing through the strutils.

### DIFF
--- a/src/google/protobuf/stubs/strutil.cc
+++ b/src/google/protobuf/stubs/strutil.cc
@@ -81,21 +81,6 @@ inline bool isprint(char c) {
 }
 
 // ----------------------------------------------------------------------
-// StripString
-//    Replaces any occurrence of the character 'remove' (or the characters
-//    in 'remove') with the character 'replacewith'.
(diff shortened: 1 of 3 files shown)


acozzette:
Thanks, @bencsikandrei! I think you're right that we ended up with this function being duplicated by mistake.

PR #6693, opened 2019-09-24
Fixed leap year handling by reworking upb_mktime() -> upb_timegm().

haberman:
The new function name also better reflects the semantics of the
function.  Like timegm(), this function always converts to/from
UTC, not local time.


TeBoring:
Could you send a change to 3.10.x?

haberman:
I'm happy to add conformance tests later, but this is a data loss bug that I want to fix now. We can add conformance tests after.

haberman:
Here is a PR against 3.10.x: https://github.com/protocolbuffers/protobuf/pull/6695

TeBoring:
Since the other PR has been merged to 3.10.x which will be merged back to master, this is no longer necessary.

Issue #6102, opened 2019-05-05
[JavaScript] Add insertion points to the compiler


travikk:
**What language does this apply to?**
This feature request is for proto3 and JavaScript language generator.

**Describe the problem you are trying to solve.**

Problem I am trying to solve, is that when using gRPC (and generated proto files) in combination with create-react-app, the eslint complain about all the generated proto JS code with the following error reported all over the shop:
`  Line 29:   'proto' is not defined  no-undef`

**Describe the solution you'd like**

Since I already have my own plugin to help generate some other things along with JS, it'd be nice to have access to insertion points for the top and bottom of the file, so I could add the following myself:

`/* eslint-disable */` and `/* eslint-enable */` respectively

**Describe alternatives you've considered**

Ana lternative would be to have an option in the compiler that would add those for me, for instance:

```
    protoc 
      --js_out="import_style=commonjs,disable_eslint=true,binary:$OUTPUT_PATH"
      ${currentProto}
```

BSBandme:
@haon4 
There's one problem that internally js protobuf is implemented through plugin.

haon4:
@dankurka 
I feel like we should add /* eslint-disable */ and /* eslint-enable */ by default. Probably don't need to add insertion points for this.

dankurka:
Before we try to suppress anything we should take a look at whether or not we can make eslint happy.

Why does lint not see the declaration of proto, could you add some output here or a way to reproduce?

travikk:
@dankurka @haon4 here's a snipped that would produce eslint issues:

```
/**
 * @fileoverview
 * @enhanceable
 * @suppress {messageConventions} JS Compiler reports an error if a variable or
 *     field starts with 'MSG_' and isn't a translatable message.
 * @public
 */
// GENERATED CODE -- DO NOT EDIT!

var jspb = require('google-protobuf');
var goog = jspb;
var global = Function('return this')();

goog.exportSymbol('proto.com.PACKAGE_NAME_ERASED.AssetCategoryDto', null, global);

/**
 * Generated by JsPbCodeGenerator.
 * @param {Array=} opt_data Optional initial data array, typically from a
 * server response, or constructed directly in Javascript. The array is used
 * in place and becomes part of the constructed object. It is not cloned.
 * If no data is provided, the constructed object will be empty, but still
 * valid.
 * @extends {jspb.Message}
 * @constructor
 */
[...]

PR #6697, opened 2019-09-25
Add insertion points in js_generator.

ribrdb:
#6102


ribrdb:
Any chance I could get a code review?

acozzette:
@ribrdb Thank you for the pull request, but I'm afraid we don't want to add any more insertion points. While insertion points can be useful, we now largely consider them to be a mistake as a protobuf feature because they break encapsulation and allow too much direct access to protobuf internals.

PR #6705, opened 2019-09-26
Fix #6098

### LINKED ISSUE #6098: cmake file overwrites existing CMAKE_STATIC_LINKER_FLAGS
Version: 3.7.0
Language: C++

Windows 10

msvc-14.2 (Visual Studio 2019)

**What did you do?**
1. Configure with cmake using a non-default target arch, e.g. `cmake -A Win32`
2. Build with Visual Studio

Build is successful

**What did you see instead?**
```
warning LNK4068: /MACHINE not specified; defaulting to x64
fatal error LNK1112: module machine type 'X86' conflicts with target
machine type 'x64'
```

The offending code is:

https://github.com/protocolbuffers/protobuf/blob/b06fd483aebfdbf804fbec15df3ebbe01e62138c/cmake/CMakeLists.txt#L183

Which overwrites the value of CMAKE_STATIC_LINKER_FLAGS thus removing the `/MACHINE` parameter which was there. This line needs to append to the existing value rather than overwrite it.

snnn:
protocolbuffers/protobuf#6098

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -199,7 +199,7 @@ if (MSVC)
   configure_file(extract_includes.bat.in extract_includes.bat)
 
   # Suppress linker warnings about files with no symbols defined.
-  set(CMAKE_STATIC_LINKER_FLAGS /ignore:4221)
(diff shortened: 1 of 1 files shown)


acozzette:
Thanks, @snnn.

PR #6731, opened 2019-10-04
Added option --create_phony_dependencies

ronern:
Added option --create_phony_dependencies that generates phony target  dependencies for .d files like -MP in gcc #

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -779,6 +779,7 @@ CommandLineInterface::CommandLineInterface()
       direct_dependencies_explicitly_set_(false),
       direct_dependencies_violation_msg_(
           kDefaultDirectDependenciesViolationMsg),
+      generate_phony_target_set_(false),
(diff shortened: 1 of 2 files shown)


src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -1844,6 +1858,15 @@ void CommandLineInterface::PrintHelpText() {
          "                              expected by make. This writes the "
          "transitive\n"
          "                              set of input file paths to FILE\n"
+         "  --create_phony_dependencies Add a phony target for each dependency, "
+         "causing\n"
+         "                              each to depend on nothing.These dummy rules "
```
HeroCC: Looks like there's a space missing here.

acozzette:
@ronern Thank you for the pull request but I'm afraid we probably don't want to accept this one. I can see how this option would be useful in some situations, but we have a fairly high bar for adding new flags to `protoc` and I'm just not sure this one is important enough to justify being incorporated directly into `protoc`.

PR #6733, opened 2019-10-04
Use proper log channel for compiling issues

core-process:
`GOOGLE_LOG` had been used for reporting of a specific input file related warning. Use error collector instead and fix `SingleFileErrorCollector` to allow propagation of warnings.

### DIFF
--- a/src/google/protobuf/compiler/importer.cc
+++ b/src/google/protobuf/compiler/importer.cc
@@ -105,6 +105,12 @@ class SourceTreeDescriptorDatabase::SingleFileErrorCollector
     had_errors_ = true;
   }
 
+  void AddWarning(int line, int column, const std::string& message) override {
+    if (multi_file_error_collector_ != NULL) {
+      multi_file_error_collector_->AddWarning(filename_, line, column, message);
+    }
+  }
+
  private:
   std::string filename_;
   MultiFileErrorCollector* multi_file_error_collector_;
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -643,10 +643,11 @@ bool Parser::Parse(io::Tokenizer* input, FileDescriptorProto* file) {
       // Store the syntax into the file.
       if (file != NULL) file->set_syntax(syntax_identifier_);
     } else if (!stop_after_syntax_identifier_) {
-      GOOGLE_LOG(WARNING) << "No syntax specified for the proto file: " << file->name()
-                   << ". Please use 'syntax = \"proto2\";' "
-                   << "or 'syntax = \"proto3\";' to specify a syntax "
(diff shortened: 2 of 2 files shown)


core-process:
Not sure what you need from me to make it "mergeable"... Please let me know. From my point of view it is "mergeable" ;-).

core-process:
```
[  FAILED  ] 7 tests, listed below:
[  FAILED  ] ParserTest.WarnIfSyntaxIdentifierOmmitted
[  FAILED  ] CommandLineInterfaceTest.InsertWithAnnotationFixup
[  FAILED  ] CommandLineInterfaceTest.PrintFreeFieldNumbers
[  FAILED  ] FileDescriptorSetSource/EncodeDecodeTest.Encode/0, where GetParam() = 0
[  FAILED  ] FileDescriptorSetSource/EncodeDecodeTest.Decode/0, where GetParam() = 0
[  FAILED  ] FileDescriptorSetSource/EncodeDecodeTest.Partial/0, where GetParam() = 0
[  FAILED  ] FileDescriptorSetSource/EncodeDecodeTest.UnknownType/0, where GetParam() = 0
```

I did not look into it. But I guess these tests need to be adjusted to the new behavior. The first one is probably looking at the wrong channel or expects a prefix generated by `GOOGLE_LOG` to be printed. The other ones might be affected by some warnings showing up now.

How do we proceed from here? If you want I can look into it next week.

acozzette:
@core-process Sorry for taking so long to get back to you. Updating the tests should be easy enough but now I'm actually that worried that this change could break a lot of code that relies on the current behavior. I would say if you want to continue then feel free to update the tests and I can merge this, but there is a significant chance that the change could get reverted later if it ends up breaking many of Google's internal tests.

core-process:
@acozzette Thanks for your feedback. I get your point; still it is sad to hear. From my point of view, it is a design mistake to use `GOOGLE_LOG` for compilation warnings. We stumbled upon this behavior while developing a framework based on Google Protocol Buffers, which links to the protoc library directly. For us, it is not reasonable to hook into `GOOGLE_LOG` to catch input file specific errors, as many problems arise from it; e.g., which file produces this warning (well, at least it needs to be handled different, than other input file related messages)? How do we handle this properly in a multi-threaded environment? When using the error collector, all these questions disappear immediately. We will fix the unit test within the next days. In case the PR gets rejected or reverted, we will keep it as a patch (which we do already). Therefore the work is not in vain, even though we would be delighted in case our PR will be included in the official project. :-)

acozzette:
Actually let me try running our internal tests against your change. It will take at least a few hours but I'll let you tomorrow what the results are.

acozzette:
@core-process I only found one of our internal tests failed and it looks easy to fix, so actually it would be great to merge this PR if you can just fix the unit test failures.

PR #6747, opened 2019-10-08
Include cmath instead of math.h

jurajlutter:
This fixes build on, at least, NetBSD 8.1 with gcc 5.5 and SmartOS with gcc 7.


jurajlutter:
@googlebot I signed it!

PR #6761, opened 2019-10-14
Ruby - Fixes for Microsoft VC

cfis:
The two commits fix a MSCV compiler error, compiler warning and crash.

* Commit 63073d3 - MSCV does not support variable length arrays (VLAs), so instead use alloca when using MSVC.

* Commit 1783b9c - Usage of the VOIDPTR_AT macro results in compiler error (or warning) C4146, "unary minus operator applied to unsigned type, result still unsigned".  If you manually expand the macro, it becomes (for line 1162):

VOIDPTR_AT(msg, -sizeof(upb_msg_internal))
(void*)(char*)(msg, (int)-sizeof(upb_msg_internal));

size_of returns size_t, which in MSVC is defined as typedef unsigned __int64 size_t. So you are taking the negative value of an unsigned type.

This change simply stops using the macro removing the (int)-sizeof bit causing this issue.

In addition, before this change I would randomly (but frequently) get segmentation faults when loading ruby generated protobuf classes (so just requiring the ruby files, not processing any data). The error was always a pointer at 0xFFFFFFFF. This change fixed the error.

Note I did try and go back to replicate the original crash (I have been running with this patch for about a week), but failed to do.

### DIFF
--- a/js/experimental/runtime/internal/checks.js
+++ b/js/experimental/runtime/internal/checks.js
@@ -2,7 +2,7 @@
  * @fileoverview Proto internal runtime checks.
  *
  * Checks are grouped into different severity, see:
- * http://g3doc/javascript/protobuf/README.md#configurable-check-support-in-protocol-buffers
+ * http://g3doc/third_party/protobuf/javascript/README.md#configurable-check-support-in-protocol-buffers
  *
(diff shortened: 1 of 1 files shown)


cfis:
@googlebot I signed it!

PR #6767, opened 2019-10-15
c++ remove warning switch contains default but no case

chrisse74:
The generated code sometimes emits code, that issues the follwing warnings:

`C4065: switch statement contains 'default' but no 'case' labels
`

This PR solves it by suppressing "switch / case / default" when encountering empty ordered-fields lists.


acozzette:
Thanks, @chrisse74.

PR #6769, opened 2019-10-15
Hide all symbols from dynamic symbol table except Init_protobuf_c.

haberman:
cc @veblush @jtattermusch


acozzette:
I'm going to close this since it looks like we probably ended up not needing the change.

PR #6771, opened 2019-10-16
c++ avoid potential macro redifinition on windows

woegster:
In my build environment WIN32_LEAN_AND_MEAN is always defined
This gives a macro-redifition warning when compiling protobuf from source
Thus: define the macro only if it's not already defined


woegster:
@googlebot I signed it!

woegster:
I see, the problem seems to be that my git configured email address differs from the one i use at github
i will simply try again with a new pull request, where that is not the case

PR #6772, opened 2019-10-16
avoid potential c++ macro redefinition on windows [release notes: no] [c++]

woegster:
In my windows project, i build protobuf from source using the WIN32_LEAN_AND_MEAN define via cmake flags.
But: common.cc also defines this macro, causing MSVC to warn about macro redifinition
Solution: only define the macro when it's not already defined


woegster:
I fail to properly understand on how to apply labels :(
I guess "release notes: no" because it is not relevant for release notes IMO
and "c++" - but that is debateable, depending on "the core, which is written in c++" (which this PR affects) or "c++ code generated by protobuf" (which this is not)

acozzette:
Thanks, @woegster.

PR #6785, opened 2019-10-22
Update protobuf version to 3.10.1

rafi-kamal:
Preparing for the Ruby and C# 3.10.1 release.


rafi-kamal:
The linkage monitor breakage is a known issue and should be fixed by https://github.com/protocolbuffers/protobuf/pull/6783.

PR #6786, opened 2019-10-23
Add com.google.truth as test dependencies

donaldchai:
This goes hand-in-hand with other dependencies like junit, but seems to
only have been used from util/pom.xml.


rafi-kamal:
Thanks a lot!

PR #6795, opened 2019-10-24
Load @rules_python in six.BUILD

Yannic:
This PR adds a load statement for `py_library` to make `@six` compatible with https://github.com/bazelbuild/bazel/issues/9006.

//cc @aaliddell @gnossen

### DIFF
--- a/protobuf_deps.bzl
+++ b/protobuf_deps.bzl
@@ -27,6 +27,7 @@ def protobuf_deps():
             name = "six",
             build_file = "@com_google_protobuf//:third_party/six.BUILD",
             sha256 = "d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73",
+            strip_prefix = "six-1.12.0",
             urls = ["https://pypi.python.org/packages/source/s/six/six-1.12.0.tar.gz"],
         )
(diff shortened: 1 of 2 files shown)


aaliddell:
If I remember correctly, there was a reason it was done this way. I'll have a check

aaliddell:
See https://github.com/protocolbuffers/protobuf/pull/6391

Prior to that PR, it was done effectively as you are now in this PR (i.e the only file in the library is `six.py`). However, we can't know at the `@six//:six` `py_library` creation time what the `legacy_create_init` flag might be set to when anyone creates a `py_binary` that depends on six. Since a number of other repos depend on `@six`, we can't just fix all the `py_binary` definitions in protobuf. Therefore, by moving `six.py` to `__init__.py`, we force it to work regardless of the `legacy_create_init` flag, as it won't overwrite an existing `__init__.py` file.

At some point `legacy_create_init` is getting flipped and removed, at which point this bodge can be fixed.

Yannic:
@aaliddell Thank you! Also fixed this in my PR for grpc.

Updated the PR to add the missing load for `@rules_python`. @acozzette can you review?

acozzette:
Thanks, @Yannic.

PR #6231, opened 2019-06-08
Fix javadoc warnings in generated files

ST-DDT:
Currently the protoc generated java classes will cause javadoc compilation warnings due to missing `@param` and `@return` values.

I first thought this was an easy change, but I soon discovered that this takes quite some effort.
That's why I would like to get some feedback whether these changes have a chance to be merged at all / are appreciated.

example diff get

example diff set and clear

This PR is by no means complete, I just changed about 50/430 occurances of `WriteFieldDocComment`.

PR #6802, opened 2019-10-25
protoc-java: Fix broken javadoc generation

ST-DDT:
protoc-3.10.0 generates broken javadocs for repeated enum values:

````java
/**
 * <pre>
 * A list of types.
 * </pre>
 * <code>repeated .foo.bar.ItemType types = 1;</code>
 * @param value The types to add.
 */
public int getTypesCount() {
    return types_.size();
}
````

which causes an `error` during javadoc generation.

This PR fixes that issue by using the correct parameter.
The error was accidentally added by me in #6231

Maybe this fix could be added to a 3.10.1 release.


acozzette:
Thanks, @ST-DDT.

PR #6810, opened 2019-10-28
Update to 3.10.0 in docs

elharo:


rafi-kamal:
I think https://github.com/protocolbuffers/protobuf/blob/master/update_version.py should be able to handle the version update in docs. I'll try adding that when I have some free cycles.

PR #6812, opened 2019-10-28
fix javascript message pivot selection [javascript]

artiz:
I have found that protoc compiler incorrectly set suggested pivot for generated message wrappers in JavaScript. So if new message is created in JS runtime without reading of real binary data then it is created with incorrect pivot and it breaks `jspb.Message.setFieldIgnoringDefault_` and so some message properties setters calls lead to data array corruption or incorrect property value load.

I have prepared simple test case at https://github.com/artiz/google-protobuf-invalid-message-pivot-selection

### DIFF
--- a/js/message.js
+++ b/js/message.js
@@ -470,7 +470,7 @@ jspb.Message.initPivotAndExtensionObject_ = function(msg, suggestedPivot) {
     }
   }
 
-  if (suggestedPivot > -1) {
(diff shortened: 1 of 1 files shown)


artiz:
@googlebot I signed it!

artiz:
Closed because incorrect solution, fix should be applied to protoc sources

PR #6818, opened 2019-10-29
Add ocaml-protoc-plugin option extension

andersfugmann:
Reserve an option extension number for Ocaml protoc plugin, to be used for extending FileOptions and ServiceOptions

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -212,3 +212,8 @@ with info about your project (name and website) so we can add an entry for you.
 1. Dart port of protocol buffers
    * Website https://github.com/dart-lang/protobuf
    * Extensions: 1073
+   
+1. Ocaml-protoc-plugin
+   * Website https://github.com/issuu/ocaml-protoc-plugin
+   * Extensions: 1074
+   


andersfugmann:
@googlebot I signed it!

acozzette:
@andersfugmann It seems that something is wrong with the CLA. Did you try to sign it as an individual or as part of a company? If you're doing it as an individual then you might want to make sure the email address on your Git commits is the same as the email address you register with the CLA.

andersfugmann:
I don't have a google account associated with my email address, and it seems that google requires me to use a google account so sign the CLA. I accidently signed it with my work account. Should I create a ticket for the same information instead and close the PR?

acozzette:
Sure, you can just create a GitHub issue with the name and website you want an option number for and then I can add it. Sorry for the extra hassle but I guess to keep the lawyers happy it would be good to avoid the appearance of any code getting checked in without the CLA signed.

andersfugmann:
Sure - I'll create a ticket. We would not want to upset the lawyers.
Created ticket: https://github.com/protocolbuffers/protobuf/issues/6819

PR #6824, opened 2019-10-30
Fix reflection access when using old generated code

jskeet:
Fixes #6822.

PR #6826, opened 2019-10-30
Avoid NullReferenceException when accessing CustomOptions

jskeet:
Even though the property is deprecated, it still shouldn't throw.


jtattermusch:
Does this need to be backported to 3.10.x or is it only problem on HEAD?

jskeet:
I'll double check - I *think* it's okay in 3.10.1.

jskeet:
Nope, it turns out it *is* a problem in 3.10.x.

@jtattermusch: Should I merge here, then create a separate PR that cherry-picks to the 3.10.x branch? What about #6824?

jtattermusch:
@jskeet can you please create a cherry-pick for 3.10.x?

jtattermusch:
@rafi-kamal   there seems to be yet another problem with the 3.10.x release - what' the ETA for 3.11.x? (if it's not very soon, we should do another patch release).

rafi-kamal:
@dankurka 

I've seen an internal discussion about doing a release soon regarding CY, is this something that can go as part of that or does it need an immediate release? 

I'm traveling now, if needed I'll be happy to do a minor release for this once I'm back on Wednesday. @TeBoring if you need any immediate release related help.

PR #6828, opened 2019-10-31
Use 3 parameter Encoding.GetString for default string values

ObsidianMinor:
`Encoding.GetString(byte[])` is actually incompatible with .NET Standard 1.0.


ObsidianMinor:
The new .cs files from #6824 (OldExtensions2.cs and OldExtensions1.cs) didn't get added to Makefile.am.

ObsidianMinor:
Rebased on master, try rerunning tests @jtattermusch

jtattermusch:
FTR
https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netstandard-1.0#System_Text_Encoding_GetString_System_Byte___System_Int32_System_Int32_

and 

https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netstandard-2.0#System_Text_Encoding_GetString_System_Byte___System_Int32_System_Int32_

PR #6832, opened 2019-10-31
Move generated test code for Google.Protobuf.Test to a separate lib

ObsidianMinor:
Many of the recent issues with proto2 extensions were accidentally caused by tests passing when code-gen didn't actually compile outside of internal visibility which the Google.Protobuf.Test project has. To prevent mistakes like this in the future, this separates the code-gen from the test project to:

 * Keep the `InternalAccessVisibleTo` attribute out of the test code-gen
 * Allow us to use a separate minimum framework and standard version when compiling the code-gen

/cc: @jtattermusch


csharp/src/Google.Protobuf.Test.CodeGen/Google.Protobuf.Test.CodeGen.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net451;netstandard1.0</TargetFrameworks>
```
jtattermusch: Let's set the LangVersion really low?
According to https://github.com/protocolbuffers/protobuf/tree/master/csharp#supported-platforms ,  C# 3 should be enough to compile the generated code.

jtattermusch: I think we'd still want to include netstandard2.0 (but we can keep netstandard1.0)?

Can you also add a comment into the .csproj file to explain why the project with generated code is separate? There are several good reasons.

jtattermusch:
Test failure:
https://source.cloud.google.com/results/invocations/95078627-231e-41e9-bc4b-665e6e48a041/log
`Unittest.cs(891,162): error CS7036: There is no argument given that corresponds to the required formal parameter 'index' of 'Encoding.GetString(byte[], int, int)' [T:\src\github\protobuf\csharp\src\Google.Protobuf.Test.CodeGen\Google.Protobuf.Test.CodeGen.csproj]`

ObsidianMinor:
That's fixed with #6828

ObsidianMinor:
If you want I can move it to this PR

csharp/src/Google.Protobuf.Test.CodeGen/Google.Protobuf.Test.CodeGen.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--
+    This CodeGen project is kept seperate from the original test project for many reasons.
+    It allows us to make sure code can compile on a seperate compiler version, different frameworks,
+    and without the internal visibility from the test project (all of which have caused issues in the past).
+  -->
+  <PropertyGroup>
+    <TargetFrameworks>net451;netstandard1.0;netstandard2.0</TargetFrameworks>
```
jtattermusch: nit: just net45 instead of net451 should be enough?

csharp/src/Google.Protobuf.Test.CodeGen/Google.Protobuf.Test.CodeGen.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--
+    This CodeGen project is kept seperate from the original test project for many reasons.
```
jtattermusch: nit: How about Google.Protobuf.Test.GeneratedProtos or just Google.Protobuf.TestProtos
instead of Google.Protobuf.Test.CodeGen as it makes the purposes even clearer?
I don't feel strongly about this, so leaving up to you.

jtattermusch:
Looks like C#3 is nearly not enough. I think we can set the LangVersion to 6 in this PR, leave a TODO and take a followup action to make it compatible with C#3 again in a future PR?
```
Unittest.cs(5142,61): error CS8024: Feature 'expression-bodied property' is not available in C# 3. Please use language version 6 or greater. [/tmp/protobuf/protobuf/csharp/src/Google.Protobuf.Test.CodeGen/Google.Protobuf.Test.CodeGen.csproj]
```

csharp/src/Google.Protobuf.Test.CodeGen/Google.Protobuf.Test.CodeGen.csproj
ObsidianMinor: Could probably keep it as TestProtos, sure.

ObsidianMinor:
I could also change it in this PR, since it's a pretty simple change.

```cs
private ExtensionSet<T> _Extensions => _extensions;
```
to
```cs
private ExtensionSet<T> _Extensions { get { return _extensions; } }
```

csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--
+    This CodeGen project is kept seperate from the original test project for many reasons.
```
jtattermusch: nit: still says "CodeGen" after renaming to TestProtos

typo: s/seperate/separate/

jtattermusch:
> I could also change it in this PR (or another PR and rebase), since it's a pretty simple change.
> 
> ```cs
> private ExtensionSet<T> _Extensions => _extensions;
> ```
> 
> to
> 
> ```cs
> private ExtensionSet<T> _Extensions { get { return _extensions; } }
> ```

Can we make that change in a separate PR and only use this PR to verify the end result? - It makes things easier to review.
(let's investigate that this is the only problem with C# 3 incompatibility)

jtattermusch:
I merged https://github.com/protocolbuffers/protobuf/pull/6856, can you update this PR?

ObsidianMinor:
Rebased 👍 @jtattermusch

ObsidianMinor:
Missed a few things when renaming project. Should've gotten everything now.

jtattermusch:
Distcheck is failing, but otherwise looking good.
```
++ '[' '!' -z 'csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj ' ']'
++ echo 'Missing files in EXTRA_DIST: csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj '
Missing files in EXTRA_DIST: csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
++ exit 1
```

ObsidianMinor:
Done @jtattermusch

PR #6833, opened 2019-11-01
Lazily Create Singular Wrapper Message

TeBoring:
In order to mimic proto2 presence semantic in proto3, some users use lots of well known wrapper to replace primitive fields. This causes the problem of performance, because object creation is expensive.

This change lazily creates wrapper objects, which saves lots of time for unnecessary object creation/deletion.

In order to use the new feature, users needs to use the protoc in this change to update their generated code. However, old generated code is still compatible with the new runtime.


php/ext/google/protobuf/encode_decode.c
```diff
@@ -887,6 +978,78 @@ static void* oneofsubmsg_handler(void* closure, const void* hd) {
   return submsg;
 }
 
+// Sets a non-repeated wrapper submessage field in a message.
+static void* wrapper_submsg_handler(void* closure, const void* hd) {
+  MessageHeader* msg = closure;
+  const submsg_handlerdata_t* submsgdata = hd;
+  TSRMLS_FETCH();
```
stanley-cheung: This may not be needed if those TSRM macros are not needed for the rest of the function? Or maybe it's needed after some other macros got expanded? please confirm.

php/ext/google/protobuf/encode_decode.c
```diff
@@ -887,6 +978,78 @@ static void* oneofsubmsg_handler(void* closure, const void* hd) {
...
+
+  return frame;
+}
+
+// Handler for a wrapper submessage field in a oneof.
+static void* wrapper_oneofsubmsg_handler(void* closure, const void* hd) {
+  MessageHeader* msg = closure;
+  const oneof_handlerdata_t *oneofdata = hd;
+  uint32_t oldcase = DEREF(message_data(msg), oneofdata->case_ofs, uint32_t);
+  TSRMLS_FETCH();
```
stanley-cheung: Ditto. Please confirm if this is needed.

php/ext/google/protobuf/encode_decode.c
```diff
@@ -432,6 +449,40 @@ static void *appendsubmsg_handler(void *closure, const void *hd) {
...
+  zval* array = (zval*)closure;
+  TSRMLS_FETCH();
+  RepeatedField* intern = UNBOX(RepeatedField, array);
+
+  const submsg_handlerdata_t *submsgdata = hd;
+  Descriptor* subdesc =
+      UNBOX_HASHTABLE_VALUE(Descriptor, get_def_obj((void*)submsgdata->md));
+  zend_class_entry* subklass = subdesc->klass;
+  MessageHeader* submsg;
+  wrapperfields_parseframe_t* frame =
```
stanley-cheung: So looks like this is the only new structure that allocates new memory in this PR? I see the corresponding free handlers, but just want to confirm.

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -175,6 +175,42 @@ private function initWithDescriptor(Descriptor $desc)
         }
     }
 
+    protected function readWrapperValue($member)
```
stanley-cheung: Thanks for the comprehensive tests!

php/ext/google/protobuf/upb.c
```diff
@@ -7153,7 +7154,8 @@ size_t run_decoder_vm(upb_pbdecoder *d, const mgroup *group,
         CHECK_SUSPEND(upb_sink_startsubmsg(outer->sink, arg, &d->top->sink));
       )
       VMCASE(OP_ENDSUBMSG,
-        CHECK_SUSPEND(upb_sink_endsubmsg(d->top->sink, arg));
+        upb_sink subsink = (d->top + 1)->sink;
```
stanley-cheung: Is this `+ 1` safe? I am not sure about the implication here.

php/ext/google/protobuf/encode_decode.c
TeBoring: UNBOX_HASHTABLE_VALUE needs it

php/ext/google/protobuf/encode_decode.c
TeBoring: UNBOX_HASHTABLE_VALUE needs it

php/ext/google/protobuf/encode_decode.c
TeBoring: Yes.

php/ext/google/protobuf/upb.c
TeBoring: I think it's safe. It's moving pointer from parent's sink to child's sink.
Because, it's handling sub message here, the sub sink should have already be created.

PR #6841, opened 2019-11-04
Ruby: Create a git version file with rake as part of the build process.

penelopezone:
This helps debuggability for dev builds, by allowing us to print the git
version the gem was built against.


penelopezone:
@googlebot I signed it!

acozzette:
@penelopezone Thank you for the pull request and sorry for taking so long to review this. Our main concern is that we would expect the version number to usually be sufficient for debugging--can you say more about why you need the Git commit? Another thing is that we heard in #9219 that if you use bundler with the git source option it will not invoke the Rakefile at all, which means that it helps if we can avoid generating any required files through rake.

penelopezone:
I opened this pull request when I worked at google and don't recall the context, apologies (you can reach out to mcloonan@ who might know)

acozzette:
Oh ok, no problem! Let me just go ahead and close this then.

PR #6856, opened 2019-11-07
Change _Extensions property to normal body rather than expression

ObsidianMinor:
Fixes issues with moving test codegen in #6832

cc: @jtattermusch


ObsidianMinor:
Yes, I cherry-picked the changes locally to the other branch to make sure it compiles properly.

PR #6861, opened 2019-11-08
Update CONTRIBUTING.md

rafi-kamal:
Add instructions for how to merge a pull request.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,3 +96,11 @@ the final release.
   pull request. This will make it easier to identify which languages the pull
   request affects, allowing us to better identify appropriate reviewer, create
   a better release note, and make it easier to identify issues in the future.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -96,3 +96,11 @@ the final release.
   pull request. This will make it easier to identify which languages the pull
   request affects, allowing us to better identify appropriate reviewer, create
   a better release note, and make it easier to identify issues in the future.
+* When merging pull requests, use the "Squash and merge" option in GitHub. Use
```
dankurka: Is there a way we can disable merge commits and multiple commits that are rebased on the github project.

rafi-kamal: We do, it's in the "Merge Button" section in https://github.com/protocolbuffers/protobuf/settings.

But in that case if someone does want to do a merge (the two cases mentioned here), they need to enable the button and then disable it again after the merge.

Issue #1137, opened 2016-01-13
protoc requires files as if they were gems


kailan:
One of my proto files contains this line: `import "api.proto";`
This gets compiled to Ruby like this: `require 'api'`

This causes an error when requiring those files because it tries to load 'api' as a gem. It would be better for the compiler to generate `require_relative 'api'` instead.

aaomidi:
+1

kblanks:
I'm working on a PR for this right now.

Zhomart:
I ended up using `$:.unshift File.expand_path('../my_proto_files', __FILE__)`.

kblanks:
Yeah, that's also what I've been doing for now.

marcinwyszynski:
I just opened PR #2477 for that.

djudd-stripe:
I'm pro adding a compiler option as discussed in https://github.com/google/protobuf/pull/1138 ("What if someone wants to put generated code into a gem, or install it into a system location?")

I have a use case where I'm importing a bunch of Protos and then generating Ruby for them, and I'd rather not mess with the load path because it's controlled by an autoloader. Currently I'm using a regex to replace the `require`s with `require_relative`s, but an option I can pass protoc would be nicer.

gaffneyc:
I would like this option as well. We're using protobufs with a Rails and placing them in a subdirectory based on the package name (e.g. `lib/dms/rpc/dispatch`) for consistency with other files.

Adding a flag for switching to `require_relative` would work for us but I would like to propose another option:

```
// Use this option to control how Ruby dependencies are required. Default is
// empty which will require dependencies as if they were in the LOAD_PATH.
//
// Options
//   "path" or unset: Require dependencies on the load path
//   "relative": Use require_relative to find dependencies relative to the
//               current file.
//   "none" or false: Don't require dependencies
option ruby_require = "relative";
```

hatstand:
I started work on this in https://github.com/protocolbuffers/protobuf/pull/6879

elharo:
@haberman Is this still relevant?

haberman:
I don't think we should emit `require_relative`. Protos should be given full path names just like any other source file.

It's important that these path names are unique; a filename like `api.proto` will confict if any dependency also has a filename `api.proto`. The filename should be something like `myproject/api.proto`, such that the base directory is in the load path.

nvolker:
[...]

PR #6879, opened 2019-11-13
Adds support for require_relative imports in ruby

hatstand:
Adds the option `require_relative` to the ruby code generator to emit
```
require_relative 'imported_proto'
```
instead of the default
```
require 'imported_proto'
```

It can be enabled with `protoc --ruby_out=require_relative:/output/directory`

This is useful, for instance, when integrating with a Ruby on Rails app.

There are a few issues with this still, but I thought I'd ask for feedback before going too far:
* No tests for the new behaviour.
* No documentation for the new behaviour.
* I'm not sure what the correct way to deal with importing well known protobuf types here would be. Right now they would not work, but we could whitelist them to always be absolute imports, for instance.


hatstand:
Ping?

haberman:
Please see my comments here: https://github.com/protocolbuffers/protobuf/issues/1137#issuecomment-932572001

Protos should refer to each other by full path names, relative to a directory in your `RUBYLIB`. This is important as protobuf requires that all imported file names are unique. Protobuf does not have a concept of relative imports.

PR #6272, opened 2019-06-17
Import MutableSequence from collections.abc on Python 3.

benjaminp:
Fixes https://github.com/protocolbuffers/protobuf/issues/5335.
Fixes https://github.com/protocolbuffers/protobuf/issues/5865.

PR #6881, opened 2019-11-13
Revert "Import MutableSequence from collections.abc on Python 3."

dankurka:
Reverts protocolbuffers/protobuf#6272


benjaminp:
Why?

Issue #6867, opened 2019-11-11
Propose to build and publish protobuf-v3.0.0-javalite for ARM platform to Maven central


liusheng:
**What language does this apply to?**
NA

**Describe the problem you are trying to solve.**
Currently, the `protoc-gen-javalite:3.0` has published in the [maven central repo](https://search.maven.org/artifact/com.google.protobuf/protoc-gen-javalite/3.0.0/pom), it support `linux_x86`, `osx_x86`, `windows_x86` system types, but didn't support the `ARM64` platform. When build some project such as `grpc-java` which depend on the `protoc-gen-javalite:3.0` on ARM64 server, it will raise error:
```
3. Execution failed for task ':grpc-protobuf-lite:generateTestProto'.                                                                                                      
> Could not resolve all files for configuration ':grpc-protobuf-lite:protobufToolsLocator_javalite'.                                                                    
   > Could not find protoc-gen-javalite-linux-aarch_64.exe (com.google.protobuf:protoc-gen-javalite:3.0.0).                                                            
     Searched in the following locations:                                                                                                  
         https://maven-central.storage-download.googleapis.com/repos/central/data/com/google/protobuf/protoc-gen-javalite/3.0.0/protoc-gen-javalite-3.0.0-linux-aarch_64.exe
```
**Describe the solution you'd like**
It doesn't need to modify anything, only need to build on ARM64 server and published to maven central repo.

**Describe alternatives you've considered**
Build manually every time

**Additional context**
NA

rafi-kamal:
@dankurka 

I'm not sure if we want to continue supporting `protoc-gen-javalite:3.0`. @ejona86 can `grpc-java` depend on the latest version of protobuf?

ejona86:
@liusheng, I mentioned that upgrading javalite could help at https://github.com/grpc/grpc-java/issues/6364#issuecomment-552550951 .

@rafi-kamal, I was waiting to upgrade until https://github.com/protocolbuffers/protobuf/issues/6565 had documentation on how to workaround, like we had talked about a few months ago.

Although now I'm also trying to remember if it was appropriate to use the pre-compiled protoc from the master branch with javalite even though javalite is on its own branch, since that's the only pre-built binary available https://search.maven.org/search?q=g:com.google.protobuf . We had discussed that at the same time, but in the months since I've forgotten what was decided.

rafi-kamal:
[...]

PR #6882, opened 2019-11-14
Add a proto_lang_toolchain for javalite

rafi-kamal:
So that javalite can be used from Bazel. Related discussion: https://github.com/protocolbuffers/protobuf/issues/6867


rafi-kamal:
@ejona86 do you know how can I test it? Is there a `java_lite_proto_library` defined somewhere that can depend on this rule?

ejona86:
It looks like your examples include a java_lite_proto_library :) https://github.com/protocolbuffers/protobuf/blob/4e93585e8bb234efeacb7737b8d080968c5ab91e/examples/BUILD#L73-L76
(And I only know this because of the build failure you mentioned in protocolbuffers/protobuf#6898)

That means you should also update it's workspace to use a `local_repository` (just like com_google_protobuf) instead of the http_archive:
https://github.com/protocolbuffers/protobuf/blob/4e93585e8bb234efeacb7737b8d080968c5ab91e/examples/WORKSPACE#L24-L28

rafi-kamal:
Thank you very much :)

PR #6889, opened 2019-11-14
Use gpg instead of gpg2 to verify RVM signature

rafi-kamal:
Kokoro tests in Mac are failing because gpg2 is not getting installed. According to https://rvm.io/rvm/security, using either gpg or gpg2 will be fine.


TeBoring:
Could you also talk to kokoro team? Seems not being able to use gpg2 is a bug.

rafi-kamal:
The issue got automatically fixed when I upgraded the Xcode version in https://github.com/protocolbuffers/protobuf/pull/6890 :) Closing this PR.

PR #6890, opened 2019-11-14
Update Xcode version needed to run Kokoro tests and replace deprecated  rubygem options

rafi-kamal:


rafi-kamal:
MacOS Python C++ is failing, but I think that's an intermittent error because the earlier invocations of the same build succeeded. I reached out to kokoro-users internally about this failure.

PR #6891, opened 2019-11-14
Adds set to import flags generation so each path is only added once

nickgeorge:
Currently, if many protos are included from the same location, the paths will be added once per import, which is redundant.  In extreme cases, such as https://github.com/google/fhir/blob/master/proto/r4/core/resources/bundle_and_contained_resource.proto , this causes the generation to fail entirely on certain platforms, including MacOS


rafi-kamal:
Hi Nick! Thanks for the PR, but we actually stopped accepting PRs to master branch due to a code yellow until our next release (possibly next week). I understand that you need it to get done as soon as possible for the launch, but Is it possible to depend on the local branch for now?

rafi-kamal:
@dankurka since we don't pull in protobuf.bzl to Google3, will it be okay to make an exception here and merge it before the CY ends?

acozzette:
@nickgeorge Thank you for the pull request, but I believe this problem ended up being fixed in #7458.

PR #6893, opened 2019-11-15
Down integrate to GitHub

rafi-kamal:


python/google/protobuf/descriptor.py
```diff
@@ -873,9 +873,14 @@ def __new__(cls, name, package, options=None,
                 syntax=None, pool=None):
       # FileDescriptor() is called from various places, not only from generated
       # files, to register dynamic proto files and messages.
-      if serialized_pb:
-        # TODO(amauryfa): use the pool passed as argument. This will work only
-        # for C++-implemented DescriptorPools.
+      # pylint: disable=g-explicit-bool-comparison
+      if serialized_pb == '':
```
asottile: this produces a `BytesWarning` when run with `-bb` -- in python3

An empty serialized protobuf is `b''` not `''` I believe (a bytestring)

rafi-kamal: @anandolee can you please take a look?

rowillia: @rafi-kamal @anandolee I created a PR here - https://github.com/protocolbuffers/protobuf/pull/6981

anandolee: We use this to distinguish the linked Cpp generated code usage, which serialized_pb is not really needed. If we change this, it requires to change the generator as well:
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/python/python_generator.cc#L470

Also change this will cause compatibility issue: old generated code may not work with new runtime.

asottile: @anandolee the problem is this current comparison is nonsense, it will never be true in python 3 as there's bytes on one side and text on the other -- it might as well be `if False`

PR #6899, opened 2019-11-16
Persistent Descriptor Pool

TeBoring:
Previously, because it's possible that different php requests may have conflicting proto message (different message with the same fully qualified name), it is required that the internal descriptor pool is refreshed after each request, which hurts performance.

This change implements persistent descriptor pool. Users needs to make sure there is no conflicting messages by themselves.

By default, this feature is disabled. To enable it, add ```protobuf.keep_descriptor_pool_after_request=1``` into php.ini. Or on command line, add ```-d protobuf.keep_descriptor_pool_after_request=1```.

### DIFF
--- a/php/ext/google/protobuf/protobuf.c
+++ b/php/ext/google/protobuf/protobuf.c
@@ -427,6 +427,7 @@ static PHP_MINIT_FUNCTION(protobuf) {
   upb_strtable_init(&reserved_names, UPB_CTYPE_UINT64);
 
   reserved_names_init();
+  internal_descriptor_pool_impl_init(&generated_pool_impl TSRMLS_CC);
 
   descriptor_pool_init(TSRMLS_C);
   descriptor_init(TSRMLS_C);
@@ -521,6 +522,8 @@ static PHP_MSHUTDOWN_FUNCTION(protobuf) {
   // Only needs to clean one map out of three (def=>desc, ce=>desc, proto=>desc)
   cleanup_desc_table(&upb_def_to_desc_map_persistent);
   cleanup_enumdesc_table(&upb_def_to_enumdesc_map_persistent);
+
+  internal_descriptor_pool_impl_destroy(&generated_pool_impl TSRMLS_CC);
 
   upb_inttable_uninit(&upb_def_to_desc_map_persistent);
   upb_inttable_uninit(&upb_def_to_enumdesc_map_persistent);
--- a/php/ext/google/protobuf/protobuf.c
+++ b/php/ext/google/protobuf/protobuf.c
@@ -488,36 +488,39 @@ static PHP_MINIT_FUNCTION(protobuf) {
   return 0;
 }
 
-static void cleanup_inttable(upb_inttable* t) {
+static void cleanup_desc_table(upb_inttable* t) {
   upb_inttable_iter i;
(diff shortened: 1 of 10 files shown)


php/ext/google/protobuf/protobuf.c
```diff
@@ -241,30 +331,21 @@ static void php_proto_hashtable_descriptor_release(zval* value) {
...
-  zend_hash_init(ce_to_php_obj_map, 16, NULL, HASHTABLE_VALUE_DTOR, 0);
+static void test_release(void* value) {
+  void* ptr = value;
+}
+#endif
 
-  ALLOC_HASHTABLE(proto_to_php_obj_map);
-  zend_hash_init(proto_to_php_obj_map, 16, NULL, HASHTABLE_VALUE_DTOR, 0);
+static initialize_persistent_descriptor_pool(TSRMLS_D) {
+  upb_inttable_init(&upb_def_to_desc_map_persistent, UPB_CTYPE_PTR);
```
stanley-cheung: What does `upb_inttable_init` do? This seems central to the theme about creating the persistent pool.

php/ext/google/protobuf/protobuf.c
```diff
@@ -276,23 +357,79 @@ static PHP_RINIT_FUNCTION(protobuf) {
...
+
+  ALLOC_HASHTABLE(ce_to_php_obj_map);
+  zend_hash_init(ce_to_php_obj_map, 16, NULL, HASHTABLE_VALUE_DTOR, 0);
+
+  generated_pool = NULL;
+  generated_pool_php = NULL;
+  internal_generated_pool_php = NULL;
+
+  if (!PROTOBUF_G(keep_descriptor_pool_after_request)) {
+    initialize_persistent_descriptor_pool(TSRMLS_C);
```
stanley-cheung: So just to confirm - the logic here is that: if the new php.ini config entry is not set (old behavior), we do this `initialize_persistent_descriptor_pool` (which I suppose is just current behavior), for every PHP request? (hence this `RINIT` function.

php/ext/google/protobuf/protobuf.c
TeBoring: It initializes a custom hashtable implemented by upb.

php/ext/google/protobuf/protobuf.c
```diff
@@ -113,6 +112,40 @@ PHP_PROTO_HASHTABLE_VALUE get_def_obj(const void* def) {
...
 }
 
+void add_msgdef_desc(const upb_msgdef* m, DescriptorInternal* desc) {
+  upb_inttable_insertptr(&upb_def_to_desc_map_persistent,
+                         m, upb_value_ptr(desc));
+}
+
+DescriptorInternal* get_msgdef_desc(const upb_msgdef* m) {
+  upb_value v;
+#ifndef NDEBUG
```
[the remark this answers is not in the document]
TeBoring: NDEBUG is a standard macro during compiling, which means no debug mode.
This is somehow exposed by upb. We don't actually depend on it. Just to let upb happy.

php/ext/google/protobuf/protobuf.c
TeBoring: Yes

PR #6904, opened 2019-11-19
C#: Allow readonly view of byteString's data with ByteString.Memory property

erikma:
To allow calling e.g. FileStream.WriteAsync() that requires ReadOnlyMemory<byte>.
You can go from a Memory<T> to a Span<T> but not vice versa.


erikma:
Maintainers please add tags:
release note: yes
language: C#

Issue #6865, opened 2019-11-08
Conformance Test Failures for C# 


rafi-kamal:
Down integration of Google3 changes (https://github.com/protocolbuffers/protobuf/pull/6858) is breaking [C# tests](https://source.cloud.google.com/results/invocations/05b870c9-ee14-4945-b46b-988d7e0ea250/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fcsharp%2Fpresubmit/log).

The broken tests have been added recently in cl/276783306. Should I add those to the failure list for now to move forward?

rafi-kamal:
I've whitelisted the failures for now in https://github.com/protocolbuffers/protobuf/pull/6869/files#diff-d4091de53e9f9136bd14511368ac0624 to move forward with the integration.

jtattermusch:
CC @ObsidianMinor

ObsidianMinor:
`ReadBool` is discarding the top 32 bits of varint values. So any varint values with no bits set in the bottom 32 will return false even if they have some bits set in the top 32.

PR #6910, opened 2019-11-19
Fix conformance test failures for Google.Protobuf

ObsidianMinor:
It also fixes the exception message for unsupported output formats in Google.Protobuf.Conformance.

cc: @jtattermusch 

Fixes https://github.com/protocolbuffers/protobuf/issues/6865


jtattermusch:
@anandolee @rafi-kamal  can you review from protobuf perspective?

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -577,7 +577,32 @@ public uint ReadFixed32()
         /// </summary>
         public bool ReadBool()
         {
-            return ReadRawVarint32() != 0;
+            int result = 0; // can't do AND on bytes, it gets implicitly upcasted to int
```
jtattermusch: good catch but the fix seems overly complex, just `ReadRawVarint64() != 0` should be enough.

FTR that is also what Java does.
https://github.com/protocolbuffers/protobuf/blob/3cae867740e409d9b9fbc1224c35357a49ce0ce2/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L782

jtattermusch:
@ObsidianMinor  I simplified the CodeInputStream changes and also fixed the bool wrapper case. Please doublecheck.

csharp/src/Google.Protobuf/CodedInputStream.cs
```diff
@@ -845,7 +845,7 @@ public bool MaybeConsumeTag(uint tag)
 
         internal static bool? ReadBoolWrapper(CodedInputStream input)
         {
-            return ReadUInt32Wrapper(input) != 0;
+            return ReadUInt64Wrapper(input) != 0;
```
jtattermusch: Unrelated to this PR, I think there might be a bug if ReadUInt64Wrapper returns null - I'll try to check that and add a test in a separate PR.

jtattermusch:
@rafi-kamal  can you take a look (and get a permission from @dankurka  to merge this fix).

csharp/src/Google.Protobuf/CodedInputStream.cs
ObsidianMinor: Right, I thought about that. Though I also thought that maybe reading an actual varint would be excessive work. The case 99% of the time is that we have at least one byte left in the buffer and the varint is a 1 or 0. ReadRawVarint64 needs at least 10 bytes and does bit shifts and other things just to ultimately check if the whole value has at least one bit set.

csharp/src/Google.Protobuf/CodedInputStream.cs
ObsidianMinor: Currently it never does so I don't think that'll be a problem.

jtattermusch:
Fine to merge now?

rafi-kamal:
Yes, please go ahead.

PR #6911, opened 2019-11-20
Implement lazy loading of php class for proto messages

TeBoring:
Previously, when creating a message, all messages defined in the same proto file and transitive proto files need to be resolved by php engine, even though some of them are not needed.
This change lazily resolve php classes of proto messages, so that they are only needed if the actual message is created.

### DIFF
--- a/php/ext/google/protobuf/encode_decode.c
+++ b/php/ext/google/protobuf/encode_decode.c
@@ -429,6 +429,7 @@ static void *appendsubmsg_handler(void *closure, const void *hd) {
 
   const submsg_handlerdata_t *submsgdata = hd;
   DescriptorInternal* subdesc = get_msgdef_desc(submsgdata->md);
+  register_class(subdesc, false TSRMLS_CC);
   zend_class_entry* subklass = subdesc->klass;
   MessageHeader* submsg;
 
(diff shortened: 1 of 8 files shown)


php/ext/google/protobuf/encode_decode.c
```diff
@@ -1356,15 +1366,6 @@ void add_handlers_for_message(const void* closure, upb_handlers* h) {
     return;
   }
 
-  // Ensure layout exists. We may be invoked to create handlers for a given
```
stanley-cheung: I don't understand why we can remove this

php/ext/google/protobuf/protobuf.h
```diff
@@ -817,6 +821,7 @@ void internal_add_generated_file(const char* data, PHP_PROTO_SIZE data_len,
                                  bool use_nested_submsg TSRMLS_DC);
 void init_generated_pool_once(TSRMLS_D);
 void add_handlers_for_message(const void* closure, upb_handlers* h);
+void register_class(void *desc, bool is_enum TSRMLS_DC);
```
stanley-cheung: So is `register_class` a function that already existed before this PR?

php/ext/google/protobuf/encode_decode.c
TeBoring: This has been done in register_class

php/ext/google/protobuf/protobuf.h
TeBoring: It existed as a static method in def.c.
In this change, I refactored a little to make it easy to use.

PR #6922, opened 2019-11-21
Update "proto_gen" rule to remove duplicate includes and deps.

gonnet:
This otherwise causes problems with too-long argument lists to "protoc".

Given "foo.proto" and "bar.proto" that both depend on "baz.proto", any proto_gen rule that depends on both "foo.proto" and "bar.proto" will pull in the includes and dependencies from "baz.proto" twice. When depending on proto targets that all depend on some common, large proto target, this leads to an explosion of import flags and causes "Argument list too long" errors on some platforms (Windows has 32k, MacOS hast 256k limits).

The proposed change converts "import_flags" and "deps" from lists to sets, and back to lists again, thus removing any duplicates. I'm not sure what this does with the order of the values, but the way they are aggregated does not make it look like it's important.

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -87,6 +87,8 @@ def _proto_gen_impl(ctx):
     for dep in ctx.attr.deps:
         import_flags += dep.proto.import_flags
         deps += dep.proto.deps
+    import_flags = depset(import_flags).to_list()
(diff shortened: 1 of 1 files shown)


gonnet:
OK, doing this for the first time, no idea how to add "release notes: no" as a label.

PR #6923, opened 2019-11-21
Reserve extension number in options.md

ashleylgeorge:
We'd like to add our extension number here for custom options.


ashleylgeorge:
Hmm.  Maybe I can do this via an issue instead.  Sorry for the noise.

PR #6925, opened 2019-11-21
Implement lazy loading of php class for proto messages (#6911)

rafi-kamal:
* Implement lazy loading of php class for proto messages

* Fix php 7.1

* Fix encode

* Fix memory leak

* Fix enum descriptor


TeBoring:
Please also cherry-pick https://github.com/protocolbuffers/protobuf/pull/6929 when it's done.

PR #6926, opened 2019-11-22
Call upb_strtable_lookup2 instead of upb_strtable_lookupptr (non-exist)

TeBoring:


php/ext/google/protobuf/protobuf.c
```diff
@@ -225,8 +225,8 @@ EnumDescriptorInternal* get_proto_enumdesc(const char* proto) {
 #ifndef NDEBUG
```
haberman: Can we just remove this entire function? It does not seem to be called anywhere.

PR #6932, opened 2019-11-23
Updated readme.md content for Windows section

Omcsesz:
Hello, Team!

I tried to use the commands written for the Windows section, but they were lacking the .\ in the beginning. I added them. Now they work.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -197,11 +197,11 @@ library.
 
 Protobuf and its dependencies can be installed directly by using `vcpkg`:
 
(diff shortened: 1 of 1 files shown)


acozzette:
I haven't used `vcpkg` before but I would expect that most of the time it wouldn't be in the current working directory, right? Shouldn't it instead go in your `$PATH` somewhere so that you don't have to call it by its full path?

elharo:
Closing due to nonresponse. Reopen if there's more info.

PR #6938, opened 2019-11-24
Fix latest ArgumentException for C# extensions

### LINKED ISSUE #6936: 3.11.0-RC2: C# throws ArgumentException when retrieving descriptors with nested extensions
Version: 3.11.0-RC2
Language: C#

MacOSX 

Protoc 3.11.0-RC2

**What did you do?**

Proto file custom_options.proto

```protobuf
syntax = "proto3";
package package1;

extend google.protobuf.MessageOptions
{
    string my_msg_opt = 50000;
}
```

Proto file message_with_custom_options.proto

```protobuf
syntax = "proto3";
package package1;

import "custom_options.proto";

message MessageWithCustomOption
{
    option (my_msg_opt) = "testing";
    string field1 = 1;
}
```

Proto file message_with_custom_option_nested.proto

```protobuf
syntax = "proto3";
package package1;

import "message_with_custom_options.proto";
import "custom_options.proto";

message MessageWithCustomOptionNested
{
    option (my_msg_opt) = "testing_nested";
    MessageWithCustomOption message = 1;
}
```
Generate code:

```bash
protoc -I. -Ipath/to/tools custom_options.proto message_with_custom_options.proto message_with_custom_options_nested.proto --csharp_out=.
```

C# code:
```csharp
using System;

namespace example
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Package1.MessageWithCustomOption.Descriptor.CustomOptions.TryGetString(50000, out string my_msg_opt);
            Console.WriteLine($"my_msg_opt: {my_msg_opt}");

            Package1.MessageWithCustomOptionNested.Descriptor.CustomOptions.TryGetString(50000, out my_msg_opt);
            Console.WriteLine($"my_msg_opt: {my_msg_opt}");
        }
    }
}
```

```
my_msg_opt: testing
my_msg_opt: testing_nested
```

This is the verified output using Protobuf 3.8. On 3.9 this code doesn't compile (#6510), on 3.10.1 it reproduces issue #6935.

**What did you see instead?**

The following output
```
my_msg_opt: testing
````

Followed by the following exception - full paths removed for simplicity.

```
Unhandled exception. System.TypeInitializationException: The type initializer for 'Package1.MessageWithCustomOptionsNestedReflection' threw an exception.
 ---> System.ArgumentException: An item with the same key has already been added. Key: Google.Protobuf.ObjectIntPair`1[System.Type]
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Google.Protobuf.ExtensionRegistry.Add(Extension extension)
   at Google.Protobuf.ExtensionRegistry.AddRange(IEnumerable`1 extensions)
   at Google.Protobuf.Reflection.FileDescriptor.AddAllExtensions(FileDescriptor[] dependencies, GeneratedClrTypeInfo generatedInfo, ExtensionRegistry registry)
   at Google.Protobuf.Reflection.FileDescriptor.FromGeneratedCode(Byte[] descriptorData, FileDescriptor[] dependencies, GeneratedClrTypeInfo generatedCodeInfo)
   at Package1.MessageWithCustomOptionsNestedReflection..cctor() in MessageWithCustomOptionsNested.cs:line 32
   --- End of inner exception stack trace ---
   at Package1.MessageWithCustomOptionsNestedReflection.get_Descriptor() in MessageWithCustomOptionsNested.cs:line 20
   at Package1.MessageWithCustomOptionNested.get_Descriptor() in MessageWithCustomOptionsNested.cs:line 50
   at example.Program.Main(String[] args) in Program.cs:line 12
```
Note the first line which is printed like expected, most likely due to the fix of this issue: #6822 

This is a minimal example reproducing an issue with real code out of the wild. The code in question compiles and works as expected with Protobuf 3.8, doesn't compile with 3.9, has the above mentioned issue with 3.10 and the described issue with 3.11-RC2.

### PR DESCRIPTION
**Pull request by ObsidianMinor**
Adds a Distinct filter for extensions loaded from depended descriptors. Fixes #6936 

cc: @jtattermusch @rafi-kamal

### CODE REVIEW

File: csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs [hunk 1]
```diff
@@ -445,9 +446,9 @@ private void CrossLink()
             }
         }
 
-        private static void AddAllExtensions(FileDescriptor[] dependencies, GeneratedClrTypeInfo generatedInfo, ExtensionRegistry registry)
+        private static IEnumerable<Extension> GetAllExtensions(FileDescriptor[] dependencies, GeneratedClrTypeInfo generatedInfo)
         {
-            registry.AddRange(dependencies.SelectMany(GetAllDependedExtensions).Concat(GetAllGeneratedExtensions(generatedInfo)).ToArray());
+            return dependencies.SelectMany(GetAllDependedExtensions).Distinct().Concat(GetAllGeneratedExtensions(generatedInfo));
```
ObsidianMinor: Should we get an enumerable over the files and run distinct on those so we don't duplicate work by loading the file's extensions multiple times?

rafi-kamal:
The MacOS Ruby test failure is a known one (thanks for fixing it so quickly btw).

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
jtattermusch: it feels that using Distinct() over Extension objects which do not explicitly define any equality / hashcode rules is fragile.

csharp/protos/extensions/extensions_b.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+import "extensions_a.proto";
+
+option csharp_namespace = "Google.Protobuf.TestProtos.Extensions";
+
+// This file is used as part of a unit test for issue 6936
+// We don't need to use it, we just have to import it in c
```
jtattermusch: typo: "import it in c"

jtattermusch: Ah, "c" stands for "extensions_c". Replace by "extensions_c.proto"?

csharp/protos/extensions/extensions_b.proto
```diff
@@ -0,0 +1,12 @@
...
+
+message Foo {
+    option (opt) = "bar";
```
jtattermusch: qq: why does Foo have option value "bar" and vice versa?

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
jtattermusch: In ExtensionRegistry, the identity of an extension is the pair (extension.TargetType, extension.FieldNumber), so we should honor that when de-duplicating. - looks like you could use Distinct overload that takes `IEqualityComparer`?

csharp/protos/extensions/extensions_a.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
```
jtattermusch: why put these files under a new "extensions" directory? they'd be fine in csharp/protos just as the other .proto files for unittests?

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
ObsidianMinor: So should we do Distinct over the file descriptors instead like above? Either way we're relying on reference equality which I think should be fine.

csharp/protos/extensions/extensions_a.proto
jtattermusch: naming: for .proto files that are there to test for previously reported issues, a pattern like this is used:
https://github.com/protocolbuffers/protobuf/blob/master/csharp/protos/unittest_issues.proto

because we need 3 different .proto files, perhaps we could name them `unittest_issue6936_a.proto` or sth like that?

csharp/protos/extensions/extensions_b.proto
ObsidianMinor: Just thought it up on a whim. I can change it if you think it doesn't work.

csharp/protos/extensions/extensions_a.proto
ObsidianMinor: Sure. Should it use the `unittest_issues` package and namespace as well?

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
owt5008137: I have found the same problem in ```GetAllGeneratedExtensions(generatedInfo)``` when two or more depended proto files import the same BASE proto with extensions. So i push another PR(#6954) to distinct the final collection and add a unit test with a more complicated situations.

ObsidianMinor: There's another issue (#6956) that I believe can be fixed by recursively gathering file descriptors from all dependencies. Can I go ahead and add the fix to this PR with tests @jtattermusch?

ObsidianMinor: Actually the issue may be unrelated depending on what James says. But it does describe another issue: extensions don't support dynamic loading.

owt5008137: Thanks, you are right. I modify my patch and just collect and distinct all depended FileDescriptors for my usage. I'm not very familar with C#. If you have better ways to do this I will close my PR(#6954) .

jtattermusch: @ObsidianMinor  we still haven't addressed my original concern (using Distinct without a well defined equality comparer).

csharp/generate_protos.sh
```diff
@@ -49,18 +49,21 @@ $PROTOC -Isrc -Icsharp/protos \
...
-    csharp/protos/unittest_proto3.proto \
-    csharp/protos/unittest_import_proto3.proto \
-    csharp/protos/unittest_import_public_proto3.proto \
-    csharp/protos/unittest.proto \
-    csharp/protos/unittest_import.proto \
-    csharp/protos/unittest_import_public.proto \
-    src/google/protobuf/unittest_well_known_types.proto \
-    src/google/protobuf/test_messages_proto3.proto \
-    src/google/protobuf/test_messages_proto2.proto
+    map_unittest_proto3.proto \
```
jtattermusch: why are you removing the prefix from the path? Looks unrelated to this PR.

jtattermusch:
distcheck is currently failing (see test results).

csharp/protos/extensions_issue6936_a.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
```
jtattermusch: I prefer a name that starts with `unittest_` to make it clear that this proto is used for unit tests. 
`unittest_issue6936_a.proto` or `unittest_issue6936_extensions_a.proto` both work.

csharp/generate_protos.sh
jtattermusch: Please revert unnecessary changes.

PR #6948, opened 2019-11-26
Fix protobuf-config.cmake for files out of source dir

haraldF:
When *.protoc files are not within the ${CMAKE_CURRENT_SOURCE_DIR},
protoc generates them relative to the first import directory.


acozzette:
@haraldF Thank you for the pull request. Could you provide some more details on what this does? I am not too familiar with the CMake setup yet.

haraldF:
@acozzette Thanks Adam for following up. While trying to create a good explanation what my change fixes I figured out that it doesn't handle all use cases. My specific case is (assuming that the protobuf sources are checked out in `$HOME/src/protobuf` and my little test project is in `$HOME/tmp/prototest`, build directory is `$HOME/tmp/prototest/build`):

```
protobuf_generate(
    PROTOS $ENV{HOME}/src/protobuf/src/google/protobuf/unittest_lite.proto
    TARGET test
    APPEND_PATH off
    IMPORT_DIRS $ENV{HOME}/src/protobuf/src
)
```

The files for `unittest_lite.proto` are generated in my build directory, but then compilation fails with:

```
$HOME/tmp/src/protobuf/src/google/protobuf/unittest_lite.pb.cc: No such file or directory
```

e.g. the information about where protoc generates the file is pretty wrong, it stitches together a path outside my build dir.

Unfortunately, my patch fixes this only partially, I'll try to dig in deeper when I have time and provide a complete fix.

PR #6951, opened 2019-11-26
Update Ruby RepeatedField known issues comment

PeterFaiman:
Array#concat accepts multiple arrays:
https://ruby-doc.org/core-2.6.5/Array.html#method-i-concat

In C, RepeatedField#push accepts multiple arguments since #5736 (a8346bc):
https://github.com/protocolbuffers/protobuf/blob/8393d48/ruby/ext/google/protobuf_c/repeated_field.c#L538-L541

But not in JRuby:
https://github.com/protocolbuffers/protobuf/blob/8393d48/ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java#L159-L173


PeterFaiman:
@googlebot

PeterFaiman:
Needs labels: "ruby" "release notes: no"

PR #6952, opened 2019-11-26
protoc: Add --python_opt.

benjaminp:
Now that there's something interesting to pass to the Python generator (cpp_generated_lib_linked), add the option flag to allow it.


benjaminp:
4447fcc16fb99c74138713aab7e7c855cdf79a1a accomplished this goal.

PR #6970, opened 2019-12-02
Migrate from maven_jar to jvm_maven_import_external to prepare for Bazel 2.0

jin:


jin:
cc @acozzette

jin:
It looks like the ruby failure is unrelated.

PR #6974, opened 2019-12-02
Initialize wrapper message during parsing

TeBoring:
In case the internal value is default


PierrickVoulet:
@TeBoring I confirm this fixes the issues we have in our UTs

PR #6976, opened 2019-12-03
Add a proto_lang_toolchain for javalite (3.11.x backport)

ejona86:
So that javalite can be used from Bazel. Related discussion: #6867

----

This is a backport of #6882

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1001,6 +1001,13 @@ proto_lang_toolchain(
     visibility = ["//visibility:public"],
 )
 
+proto_lang_toolchain(
+    name = "javalite_toolchain",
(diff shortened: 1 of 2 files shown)


rafi-kamal:
The 3.11.1 release process has already started (the protoc for 3.11.1 has already been uploaded to Maven), I'm happy to merge the PR *after* 3.11.1 release is finished (today/tomorrow), but we don't have a timeline for the next release yet.

ejona86:
Just being on the release branch is enough. It is easy to point to a commit from Bazel, but pointing to something on master (like #6882) makes it unclear if there will be compatibility issues.

PR #6981, opened 2019-12-03
Fix a bytes vs. strings bug in FileDescriptor

rowillia:
It appears this was broken in:
https://github.com/protocolbuffers/protobuf/pull/6893/files#r351384126

This is an actual bug, not just a warning issue:

```python
In [1]: '' == b''
Out[1]: False
```

This can be caught by adding the `-bb` flag to CI:

```
rwilliams at rwilliams-mbp114 in ~
$ python3 -c '"" == b""'

rwilliams at rwilliams-mbp114 in ~
$ python3 -bb -c '"" == b""'
  File "<string>", line 1, in <module>
BytesWarning: Comparison between bytes and string
```
Lyft runs all Python 3 CI jobs with this flag to catch just such issues.

Lyft has signed the Google CLA.

### DIFF
--- a/python/google/protobuf/descriptor.py
+++ b/python/google/protobuf/descriptor.py
@@ -874,7 +874,7 @@ class FileDescriptor(DescriptorBase):
       # FileDescriptor() is called from various places, not only from generated
       # files, to register dynamic proto files and messages.
       # pylint: disable=g-explicit-bool-comparison
-      if serialized_pb == '':
+      if serialized_pb == b'':
         # Cpp generated code must be linked in if serialized_pb is ''
         try:
           return _message.default_pool.FindFileByName(name)
--- a/python/google/protobuf/descriptor.py
+++ b/python/google/protobuf/descriptor.py
@@ -874,8 +874,10 @@ class FileDescriptor(DescriptorBase):
       # FileDescriptor() is called from various places, not only from generated
       # files, to register dynamic proto files and messages.
       # pylint: disable=g-explicit-bool-comparison
-      if serialized_pb == b'':
-        # Cpp generated code must be linked in if serialized_pb is ''
+      if len(serialized_pb) == 0:
+        # Cpp generated code must be linked in if serialized_pb is ''.
+        # Explicity check length as this can be either of type `bytes` or `str`
+        # and this is the fastest method for checking for either an empty bytes.
         try:
           return _message.default_pool.FindFileByName(name)
         except KeyError:
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -15,10 +15,10 @@ commands =
     python setup.py -q build_py
     python: python setup.py -q build
     cpp: python setup.py -q build --cpp_implementation --warnings_as_errors --compile_static_extension
-    python: python setup.py -q test -q
-    cpp: python setup.py -q test -q --cpp_implementation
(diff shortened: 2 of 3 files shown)


rowillia:
@rafi-kamal @anandolee Anything I can do to help push this through?  We're currently blocked on picking up newer versions of `protobuf` due to this issue.

python/google/protobuf/descriptor.py
```diff
@@ -874,7 +874,7 @@ def __new__(cls, name, package, options=None,
       # FileDescriptor() is called from various places, not only from generated
       # files, to register dynamic proto files and messages.
       # pylint: disable=g-explicit-bool-comparison
-      if serialized_pb == '':
+      if serialized_pb == b'':
```
anandolee: This is used to distinguish the linked Cpp generated code usage, which serialized_pb is not really needed. If we change this, it requires to change the generator as well:
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/python/python_generator.cc#L470

Also change this will cause compatibility issue: old generated code may not work with new runtime.

rowillia: Thanks for the feedback @anandolee !  Updated the generator.

The change introduced in `3.11` already caused compatibility issues - before 3.11 this branch would be take on empty bytes, now it won't:

```
In [1]: if '' == b'':
   ...:     print('Empty Bytes Are Empty Strings')
   ...: if not b'':
   ...:     print('Empty Bytes Are False-y')
   ...:
Empty Bytes Are False-y
```

Updated to check the length as this is the fastest way to determine if it's either `b''` or `''` without hitting a `BytesWarning`.

```python
In [1]: x = b''

In [2]: %timeit len(x) == 0
72.1 ns ± 0.787 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit (isinstance(x, str) and x == '') or (isinstance(x, bytes) and x == b'')
233 ns ± 8.01 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

rowillia:
@anandolee Anything I can do here to get this landed?  We're currently unable to pick up new versions of protobuf due to this issue.

rowillia:
@rafi-kamal @anandolee Anything I can do to help push this through? We're currently blocked on picking up newer versions of protobuf due to this issue.  Just want to reiterate that this is a real bug - as [@asottile said](https://github.com/protocolbuffers/protobuf/pull/6893/files#r358391909) this branch might as well be  `if False:` in Python 3.

rafi-kamal:
CC @dlj-NaN

davidszotten:
hey, is this blocked on the failing tests above or something else?

mxr:
it looks like this was addressed in b99994d994e399174fe688a5efbcb6d91f36952a

```diff
-       if serialized_pb == '':
+       if serialized_pb == b'':
```

rowillia:
cc @anandolee and @rafi-kamal It looks like this isn't needed anymore after https://github.com/protocolbuffers/protobuf/commit/b99994d994e399174fe688a5efbcb6d91f36952a, right?

anandolee:
Yes, I believe if serialized_pb == b'' solved it already and this does not needed

PR #6999, opened 2019-12-09
Add --<lang>_opt flag for all built-in generators

Yannic:
Plugins (and some built-in generators) allow passing parameters
with `--<lang>_opt` one-by-one instead of passing them as
`--<lang>_out=<params>:<out_base>`. This PR changes protoc to
allow using `--<lang>_opt` for all (built-in) generators.

### DIFF
--- a/src/google/protobuf/compiler/main.cc
+++ b/src/google/protobuf/compiler/main.cc
@@ -69,17 +69,17 @@ int ProtobufMain(int argc, char* argv[]) {
 
   // Proto2 Python
   python::Generator py_generator;
-  cli.RegisterGenerator("--python_out", &py_generator,
+  cli.RegisterGenerator("--python_out", "--python_opt", &py_generator,
                         "Generate Python source file.");
 
   // PHP
   php::Generator php_generator;
-  cli.RegisterGenerator("--php_out", &php_generator,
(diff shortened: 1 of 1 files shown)


Yannic:
Thanks! The failures look unrelated to this change.

macOS Python Release:
```
+ sudo installer -pkg downloads/python-3.5.4-macosx10.9.pkg -target /
installer: Error - the package path specified was invalid: 'downloads/python-3.5.4-macosx10.9.pkg'.


[ID: 4240753] Build finished after 622 secs, exit value: 1
```

macOS Ruby 2.3
```
Ruby was built without documentation, to build it run: rvm docs generate-ri
Using /Users/kbuilder/.rvm/gems/ruby-2.3.8
/Users/kbuilder/.rvm/rubies/ruby-2.3.8/bin/ruby
ERROR:  Loading command: install (LoadError)
	cannot load such file -- openssl
ERROR:  While executing gem ... (NoMethodError)
    undefined method `invoke_with_build_args' for nil:NilClass


[ID: 2752210] Build finished after 705 secs, exit value: 1
```

macOS Ruby Release timed out, not sure why but I'm sure not because of this change.

acozzette:
@Yannic Thanks for the pull request.

PR #7003, opened 2019-12-10
Remove use of VLA.

thomasvl:
VLAs complicate static analysis and bloat stack size. Replace VLA allocation
with calls to malloc and free.  This will alos the code to build with -Wvla.


thomasvl:
@dmaclach

PR #7005, opened 2019-12-11
Add backwards compatibility for custom options

### LINKED ISSUE #6956: MethodDescriptor.CustomOptions no longer works in 3.11.0-rc2 (when old generated code is used)
I want to get custom options off a `MethodDescriptor`. Specifically I want to get `HttpRule` which is from Google.Api.CommonProtos 1.7.0. In Google.Protobuf 3.8.0 I can successfully get this option by doing this:

```cs
const int HttpRuleFieldId = 72295728;
var success = methodDescriptor.CustomOptions.TryGetMessage<HttpRule>(HttpRuleFieldId, out httpRule);
```

With Google.Protobuf 3.9.0 and 3.10.0 `CustomOptions.TryGetMessage<T>` errors, I believe related to this issue: https://github.com/protocolbuffers/protobuf/pull/6824

When I updated to 3.11.0-rc2 `CustomOptions.TryGetMessage<T>` no longer errors, but it also doesn't return a value. You should fix that for backwards compatibility.

I see that CustomOptions has been made obsolete. How do I use the new API?

```cs
const int HttpRuleFieldId = 72295728;
const uint TagId = 0; // what is this?

var extension = new Extension<Google.Protobuf.Reflection.MethodOptions, HttpRule>(HttpRuleFieldId, FieldCodec.ForMessage<HttpRule>(TagId, HttpRule.Parser));
var httpRule = methodDescriptor.GetOption<HttpRule>(extension); // returns null
```

ObsidianMinor:
Fixes #6956. Old generated code that used custom options will now be able to continue to use CustomOptions to access the extensions without generated identifiers. I still need to add tests to make sure it properly works.


ObsidianMinor:
Review time @jtattermusch. Lots of stuff in this one, since I basically had to rewrite CustomOptions to try and access fields from both an extension field dictionary and an UnknownFieldSet.

Related question: When, if ever, can everything be broken once and for all (i.e. next major lib version)

csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
```diff
@@ -75,107 +76,128 @@ OptionFetcher<E> EnumFetcher<E>(CustomOptions options)
         public void ScalarOptions()
```
jtattermusch, line 76: would it make more sense to have a separate test case ScalarOptions_OldGeneratedCode() that would test the legacy code? Such test could also have a comment that explains what it is testing (and perhaps link the github issue).

csharp/src/Google.Protobuf/Reflection/DescriptorPartial.cs
```diff
@@ -0,0 +1,50 @@
+namespace Google.Protobuf.Reflection
+{
+    public sealed partial class FileOptions 
+    {
+        internal ExtensionSet<FileOptions> Extensions => _extensions;
```
jtattermusch, line 5: Looks like these two properties are only needed to create an instance of custom options.
If so, why expose the properties at all and instead add a CreateCustomOptions() method which creates the CustomOptions instance for you.

csharp/src/Google.Protobuf/UnknownFieldSet.cs
```diff
@@ -135,6 +136,75 @@ public override int GetHashCode()
             return ret;
         }
 
+        internal static bool TryGetField(UnknownFieldSet set, int field, out UnknownField value)
```
jtattermusch, line 139: It seems that it should be enough to just provide the TryGetField method here. The other methods (such as TryGetLastVarint) are IMHO just syntactic sugar that bloats the suface of this class and they don't really belong here. Extracting the last value should be done by CustomOptions class (and UnknownFieldSet and UnknownField should only provide the minimum necessary functionality to allow that)

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
```diff
@@ -65,10 +65,15 @@ public sealed class CustomOptions
     {
         private static readonly object[] EmptyParameters = new object[0];
         private readonly IDictionary<int, IExtensionValue> values;
+        // since we made the descriptor protos public, we can't really reintroduce the CustomOptions
```
jtattermusch, line 68: qq: before unknown field support was added, how did the custom options work?
Also how did they work after unknownField support was added but before support for extensions was added?

Understanding these seems important for figuring out the right approach to this.

ObsidianMinor: Before CustomOptions worked by acting as a sort of custom UnknownFieldSet with exposed methods to read values. Once unknown field support was added they kept their custom behavior since unknown fields weren't exposed in the message API.

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -144,12 +144,7 @@ void MessageGenerator::Generate(io::Printer* printer) {
       "private pb::UnknownFieldSet _unknownFields;\n");
 
   if (has_extension_ranges_) {
-    if (IsDescriptorProto(descriptor_->file())) {
-      printer->Print(vars, "internal pb::ExtensionSet<$class_name$> _extensions;\n"); // CustomOptions compatibility
```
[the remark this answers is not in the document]
ObsidianMinor: I don't, which is why I removed that code.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -129,7 +129,7 @@ public EnumValueDescriptor FindValueByName(string name)
         /// The (possibly empty) set of custom options for this enum.
         /// </summary>
         [Obsolete("CustomOptions are obsolete. Use GetOption")]
-        public CustomOptions CustomOptions => new CustomOptions(Proto.Options?._extensions?.ValuesByNumber);
+        public CustomOptions CustomOptions => Proto?.Options?.CreateCustomOptions() ?? CustomOptions.Empty;
```
jtattermusch, line 132: AFAICT If `Proto` is null, it would mean that something is badly broken, so I don't quite understand the `Proto?.` operator (it's also something you didn't have in the previous version of the code).
Same for all `*Descriptor` files.

src/google/protobuf/compiler/csharp/csharp_message.cc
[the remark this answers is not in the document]
jtattermusch: ah, sorry, I got confused and thought it's code addition, not removal. My bad. ;-)

csharp/src/Google.Protobuf/UnknownField.cs
```diff
@@ -64,6 +64,8 @@ public UnknownField()
         {
         }
 
+        internal List<ByteString> LengthDelimitedList => lengthDelimitedList;
```
jtattermusch, line 67: I'd like to get away without exposing lengthDelimitedList (it's the only one we're exposing and feels odd).

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
```diff
@@ -205,98 +232,116 @@ internal CustomOptions(IDictionary<int, IExtensionValue> values)
...
+                                    value = (T)values[values.Count - 1];
+                                    return true;
+                                }
+                            }
+                        }
                     }
                 }
             }
 
+            UnknownField f;
```
jtattermusch, line 334: Without further context, from the code itself it's not clear at all why we would look at extensions first and then fall back to trying to read from unknown fields. There should really be a comment that explains why consulting why UnknownFieldSet is important.
Same for TryGetMessage.

Perhaps the refactoring I proposed would help to make this clearer.

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
```diff
@@ -205,98 +232,116 @@ internal CustomOptions(IDictionary<int, IExtensionValue> values)
...
-            if (values == null)
-            {
-                value = default(T);
-                return false;
-            }
+        private delegate bool TryGetLastValue<T>(UnknownField field, out T value);
 
-            IExtensionValue extensionValue;
-            if (values.TryGetValue(field, out extensionValue))
+        private bool TryGetPrimitiveValue<T, U>(int field, TryGetLastValue<U> getValue, Func<U, T> conv, out T value)
```
jtattermusch, line 280: it feels odd that you're passing a further unspecified "getValue" function, but you're only using it only sometimes, when reading from the unknownFieldSet - feels dirty.
Same for the "conv" argument.

jtattermusch: Here's a idea how to refactor for improved clarity

```
public bool TryGetInt32(int field, out int value)
{
     if (TryGetPrimitiveValueFromExtensions())
     {
         return true;
     }
     // the method's summary would have an explanation why it's also important to check the unknown fields
    // when calling this method, the "getValue" and "conv" arguments make more sense.
     if (TryGetPrimitiveValueFromUnknownFields(filed, getValueFunc, conversionFunc))
     {
        // alternatively, you can perform the conversion here...
       return true;
     }
    return false;
}
```

jtattermusch, line 280: The meaning of the `U` generic argument isn't clear at all from the first glance (I'm guessing it's the type of the result for the unknown field, but it's very unobvious).
Perhaps the refactoring I proposed should help with that.

Btw, the `U` argument also looks useless. Why cannot the getValueFunc get the value AND perform the conversion, eliminating the need for `U` completely?

csharp/src/Google.Protobuf/UnknownField.cs
ObsidianMinor: If we expose the rest of the lists would that be ok? I don't know how I feel about adding a function to UnknownField just to create a message value from the length delimited values in it.

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
ObsidianMinor: The way it's designed currently is optimized to remove code duplication. Since most values in the UnknownField are represented different from how they're exposed in `TryGet*` methods, I abstracted the `TryGetLast*` out and made it so you just need to provide the value's custom conversion method. Making it so the `getValueFunc` had to get the value and perform the conversion would increase the code size tremendously. Plus anonymous delegates can't have `out` parameters. So we'd have to use tuples to communicate if a value existed in the unknown field and the converted value, unless we want to make a separate local function just to do it.

jtattermusch:
Even after merging https://github.com/protocolbuffers/protobuf/pull/7434 this is probably still worth doing, if we can make it simple. The conflicts need to be resolved first (rebase?) and there's a bunch of review comments from me.

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
ObsidianMinor: Adding that to *every* method would add a lot of bloat that could probably be handled with a good doc comment and comments throughout the function. For example, something like...
```cs
/// <summary>
/// Tries to get the primitive value from the extension dictionary, otherwise falls-back to the unknown field set.
/// </summary>
/// <typeparam name="T">The returned type</typeparam>
/// <typeparam name="U">The unknown field representation</typeparam>
/// <param name="field">The field number to try and get a value for</param>
/// <param name="getValue">A function that tries to get the last value of a field of a specific type in an unknown field.</param>
/// <param name="conv">A conversion function that converts from an unknown field value to the output value</param>
/// <param name="value">The resulting value</param>
/// <returns><c>true</c> if a field value was found, <c>false</c> otherwise</returns>
private bool TryGetPrimitiveValue<T, U>(int field, TryGetLastValue<U> getValue, Func<U, T> conv, out T value)
```

ObsidianMinor: Though now that I look at it it seems you may still the 2 calls relatively small with an `||`
```
public bool TryGetInt32(int field, out int value)
    => TryGetExtensionValue<int>(field, out value)
    || TryGetLastUnknownValue<int, ulong>(field, UnknownField.TryGetLastVarint, (v) => (int)v, out value);
```
I don't know if it really looks any better. I guess I could see how it makes it more explicit up front...

ObsidianMinor:
Are most of these comments still issues? I still believe it's best to keep code duplication to a minimum, so I'd prefer to keep TryGetPrimitiveValue instead of expending all its uses to do two calls to try and get the extension value and unknown field value separately *for every field type*.

As for renaming the old generated proto files, I think it'd be good to reorganize them by what version they're generated with and go through them *after this PR* since many of them can be removed at this point and we can go back to using the originals in `src`.

jtattermusch (commented):
Since https://github.com/protocolbuffers/protobuf/issues/6956 is now quite old and there's been several protobuf releases since then, I'm not convinced that this PR still has some value ( note that this PR only fixes the behavior of CustomOptions in the relatively rare case when CustomOptions are being access by a new runtime,  but old generated code is being used).

Btw, since the issue has been opened, other big changes went in in https://github.com/protocolbuffers/protobuf/pull/7351 and https://github.com/protocolbuffers/protobuf/pull/7576, so for those who are using the new runtime with old generated code, this is one more reason to regenerate the code (which basically solves https://github.com/protocolbuffers/protobuf/issues/6956).

I'm going to close this PR since we no longer need it (and more work would be needed to finish this PR, it cannot be merged as is).

Issue #6803, opened 2019-10-25
'protobuf/Any.pbobjc.h' file not found


Interfere:
I use CocoaPods to add Protobuf to my project. However when I build with Xcode 11.0 it reports 'file not found' and do not compile.

I believe the issue is somehow related to this [commit](https://github.com/protocolbuffers/protobuf/commit/479ba8226bed9c0986693f240c4d6f72dc2d0368#diff-e074f11306e8dc9854c4bd70e4886086) and therefore is related to the issue #3218.
<img width="673" alt="Screenshot 2019-10-25 at 16 33 28" src="https://user-images.githubusercontent.com/1098419/67575927-3cb86680-f746-11e9-8a15-69d1ec33d81f.png">

Here is an example project: https://github.com/Interfere/MalformedPod

rafi-kamal:
@thomasvl 

Which version of protobuf are you using?

Interfere:
I have tried this with two versions: 3.9.0 and 3.10.0

thomasvl:
@paulb777

thomasvl:
3.9.0 shouldn't have the change to the pod spec, so if you are saying you see it there, something else must be up.

thomasvl:
https://github.com/Interfere/MalformedPod/blob/master/Podfile.lock#L16

```
COCOAPODS: 1.5.3
```

You appear to be using a pretty old version of cocoapods [(May 2018)](https://github.com/CocoaPods/CocoaPods/releases/tag/1.5.3); if you update your install do things work better?  Your example seems to work fine for me with Cocoapods 1.8.4.

Interfere:
I have updated the cocoapods. Still the same issue.

thomasvl:
> I have updated the cocoapods. Still the same issue.

Did you do a clean and pod install again so it could make changes to the projects?

Here's what I tried:

```
thomasvl@thomasvl-macbookpro:~/Desktop/testing$ git clone https://github.com/Interfere/MalformedPod.git
Cloning into 'MalformedPod'...
remote: Enumerating objects: 36, done.
remote: Counting objects: 100% (36/36), done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 36 (delta 8), reused 36 (delta 8), pack-reused 0
Receiving objects: 100% (36/36), 11.53 KiB | 11.53 MiB/s, done.
Resolving deltas: 100% (8/8), done.
[...]

PR #6464, opened 2019-08-02
Override CocoaPods module to lowercase

paulb777:
Fix #3218 

Implement suggestion from @BugsBunnyBR at https://github.com/protocolbuffers/protobuf/issues/3218#issuecomment-517643842 to use `module_name` in podspec to make the module name consistent with the directory structure.

Presumably another change is needed to create the generated files properly?

PR #6190, opened 2019-05-29
Fix CocoaPods case warnings

paulb777:
Fix #3218 

Verified that warnings go away in Protobuf and FirebaseMessaging client build when installed with `use-frameworks!`, `use-modular-headers`, and static libraries.

I realize this might not be a great solution for non-CocoaPods users, but I'm not sure there is a solution that wouldn't be a breaking change for one or the other user set.

PR #7007, opened 2019-12-11
Fix compile errors on case-sensitive APFS

ledyba-z:
Hi. I couldn't add protobuf to iOS App. using CocoaPods with compile errors:

スクリーンショット 2019-12-11 11 43 43

I figured out `Any.pbobjc.h` is located on `Protobuf`, not `protobuf`, and also, my disk is formatted as case-sensitive APFS. So that the objc compiler cannot find the header.

Here is my podfile:

```
# Uncomment this line to define a global platform for your project
platform :ios, '11.0'
use_modular_headers!
inhibit_all_warnings!
# use_frameworks!

target '*****' do
    ....
    pod 'Protobuf'
    ....
end
```

This patch may also fixes: https://github.com/invertase/react-native-firebase/issues/2780

### DIFF
--- a/objectivec/GPBProtocolBuffers.h
+++ b/objectivec/GPBProtocolBuffers.h
@@ -52,16 +52,16 @@
 
 // Well-known proto types
 #if GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS
- #import <protobuf/Any.pbobjc.h>
- #import <protobuf/Api.pbobjc.h>
- #import <protobuf/Duration.pbobjc.h>
- #import <protobuf/Empty.pbobjc.h>
- #import <protobuf/FieldMask.pbobjc.h>
- #import <protobuf/SourceContext.pbobjc.h>
- #import <protobuf/Struct.pbobjc.h>
- #import <protobuf/Timestamp.pbobjc.h>
- #import <protobuf/Type.pbobjc.h>
- #import <protobuf/Wrappers.pbobjc.h>
+ #import <Protobuf/Any.pbobjc.h>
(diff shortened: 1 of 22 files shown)


ledyba-z:
@googlebot I signed it!

ledyba-z:
Oh, I found these closed pull requests...

Override CocoaPods module to lowercase by paulb777 · Pull Request #6464 · protocolbuffers/protobuf
https://github.com/protocolbuffers/protobuf/pull/6464

Fix CocoaPods case warnings by paulb777 · Pull Request #6190 · protocolbuffers/protobuf
https://github.com/protocolbuffers/protobuf/pull/6190

What should I do for this compile error?

I know that this patch solves my compile errors, but is this the correct way?
To be honest, I am not familiar with cocoapods/iOS/ObjC.

ledyba-z:
I'm using Version 11.3 (11C29),

other software versions:

```
% cat .ruby-version
2.6.3
% bundler exec pod --version
1.8.4
```

thomasvl:
As you see from those other issues, there isn't a clear "fix" at the moment.

This PR would cause warnings for some of the other setups because the tree of directories/links created by cocoapods isn't consistent (some times it is the package name, some times it is the module name).

So short of doing what is suggested in some of those where we abandon the _Protobuf_ pod name for _protobuf_ and thus require **everyone** to migrate, there doesn't seem to be a real solution (and during that migration folks could run into double definitions when they get both).

thomasvl:
ps - #6803 is the currently open issue for tracking the problem

ledyba-z:
Thank you reviewing and telling me the current status!
OK, I will track that issue.

PR #7012, opened 2019-12-13
Draft: Python Runtime Codegen

gnossen:
This PR is meant to serve as a basis for discussion and is not yet ready for a thorough review.

Companion to https://github.com/grpc/grpc/pull/21458.

This PR adds a runtime dependency on `libprotoc` and enables users to import `.proto` files without first manually invoking `protoc`.

The only change to the API surface is the addition of a function `google.protobuf.protos`, which accepts the path to a `.proto` file (as well as some optional configuration) and returns the corresponding Python module. The new API is only made available to users running Python 3+ and using the cpp implementation. All other users who attempt to use the new function will get a `NotImplementedError`.

CC @anandolee


gnossen:
Closing in favor of a different implementation approach after an offline discussion with the team.

PR #7025, opened 2019-12-18
Pluggable Runtime Codegen Implementation

gnossen:
This PR is an alternative to https://github.com/protocolbuffers/protobuf/pull/7012 and a sibling to https://github.com/grpc/grpc/pull/21504.

In simple cases, this will allow users of the library to use messages defined in `foo.proto` like so:

```python
from google import protobuf
protos = protobuf.protos("foo.proto")
print(protos.FooMessage())
```
without ever invoking `protoc`.

The difference from https://github.com/protocolbuffers/protobuf/pull/7012 is that one additional API has been added to enable plugin systems like gRPC's to take advantage of the same dynamic import machinery. At the time of writing, the two additions to the API look like this:

```python
def protos(protobuf_path: Text,
           include_paths: Optional[Sequence[Text]] = None) -> module:
  pass

ImportFunction = Callable[[Text], module] # signature of google.protobuf.protos function

def get_import_machinery(module_suffix: Text,
                         code_generator: CodeGenerator) ->
                         Tuple[importlib.abc.MetaPathFinder, ImportFunction]:
  pass
```
where CodeGenerator is a protocol requiring only the `__int__` method to be defined. It is expected that this method return an integer corresponding to a C++ `const ::google::protobuf::compiler::CodeGenerator*` where the underlying object has static storage duration (in order to avoid the tricky issue of managing memory allocated outside of CPython itself).

This code generator is then used to generate the code for the `.proto` file requested, along with its dependency tree.

This mechanism cuts down substantially on the amount of code duplication between the gRPC and Protobuf repos, as well as enabling other systems built on top of Protobuf (e.g. Stubby) to easily do the same in the future.

I am completely flexible on the naming and module placement of this function, as only library authors such as gRPC itself will ever need to use it.

This PR is still a work in-progress:
 - [ ] Port existing detailed codegen tests from gRPC PR
 - [ ] Ensure various versions of CPython are being tested
 - [ ] Format Code
 - [ ] Add documentation
 - [ ] Integrate with the repo's C++-style Python APIs (i.e. ScopedPythonPtr)


gnossen:
Closing in favor of a different implementation approach after an offline discussion with the team.

PR #7026, opened 2019-12-18
[objc] Use references to Obj C classes instead of names in descriptors.

dmaclach:
This switches the compiler and library to use objective-c class references instead of class names (that are resolved using objc_getClass/NSClassFromString). 

This should:
1. Reduce binary size slightly as we no longer need the strings.
2. Add a small performance improvement as we no longer need to look up classes.
3. Help enforce proper linkage by having actual class references for the linker to resolve. In cases where we are linking static archives there were cases where some of the Objective-C classes were not being pulled into the linkage requiring the use of the -ObjC linker flag (or other linker tricks). There are some other places in the code that we need to do this to get full coverage, but this is a good start.

### DIFF
--- a/objectivec/GPBDescriptor.m
+++ b/objectivec/GPBDescriptor.m
@@ -44,7 +44,7 @@
 
 // The addresses of these variables are used as keys for objc_getAssociatedObject.
 static const char kTextFormatExtraValueKey = 0;
-static const char kParentClassNameValueKey = 0;
+static const char kParentClassValueKey = 0;
 static const char kClassNameSuffixKey = 0;
 
 // Utility function to generate selectors on the fly.
@@ -126,6 +126,8 @@ static NSArray *NewFieldsArrayForHasIndex(int hasIndex,
   GPBFileSyntax syntax = file.syntax;
   BOOL fieldsIncludeDefault =
       (flags & GPBDescriptorInitializationFlag_FieldsWithDefault) != 0;
+  BOOL usesClassRefs =
+      (flags & GPBDescriptorInitializationFlag_UsesClassRefs) != 0;
 
   void *desc;
   for (uint32_t i = 0; i < fieldCount; ++i) {
@@ -143,6 +145,7 @@ static NSArray *NewFieldsArrayForHasIndex(int hasIndex,
     GPBFieldDescriptor *fieldDescriptor =
         [[GPBFieldDescriptor alloc] initWithFieldDescription:desc
                                              includesDefault:fieldsIncludeDefault
+                                               usesClassRefs:usesClassRefs
                                                       syntax:syntax];
     [fields addObject:fieldDescriptor];
     [fieldDescriptor release];
@@ -217,15 +220,19 @@ static NSArray *NewFieldsArrayForHasIndex(int hasIndex,
   extensionRangesCount_ = count;
 }
 
+- (void)setupContainingMessageClass:(Class)messageClass {
+  objc_setAssociatedObject(self, &kParentClassValueKey,
+                           messageClass,
+                           OBJC_ASSOCIATION_ASSIGN);
+}
+
 - (void)setupContainingMessageClassName:(const char *)msgClassName {
   // Note: Only fetch the class here, can't send messages to it because
   // that could cause cycles back to this class within +initialize if
   // two messages have each other in fields (i.e. - they build a graph).
-  NSAssert(objc_getClass(msgClassName), @"Class %s not defined", msgClassName);
-  NSValue *parentNameValue = [NSValue valueWithPointer:msgClassName];
-  objc_setAssociatedObject(self, &kParentClassNameValueKey,
-                           parentNameValue,
-                           OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+  Class clazz = objc_getClass(msgClassName);
+  NSAssert(clazz, @"Class %s not defined", msgClassName);
+  [self setupContainingMessageClass:clazz];
 }
 
 - (void)setupMessageClassNameSuffix:(NSString *)suffix {
@@ -241,14 +248,7 @@ static NSArray *NewFieldsArrayForHasIndex(int hasIndex,
 }
 
 - (GPBDescriptor *)containingType {
-  NSValue *parentNameValue =
-      objc_getAssociatedObject(self, &kParentClassNameValueKey);
-  if (!parentNameValue) {
-    return nil;
-  }
-  const char *parentName = [parentNameValue pointerValue];
-  Class parentClass = objc_getClass(parentName);
-  NSAssert(parentClass, @"Class %s not defined", parentName);
+  Class parentClass = objc_getAssociatedObject(self, &kParentClassValueKey);
   return [parentClass descriptor];
 }
 
@@ -487,6 +487,7 @@ uint32_t GPBFieldAlternateTag(GPBFieldDescriptor *self) {
 
 - (instancetype)initWithFieldDescription:(void *)description
                          includesDefault:(BOOL)includesDefault
+                           usesClassRefs:(BOOL)usesClassRefs
                                   syntax:(GPBFileSyntax)syntax {
   if ((self = [super init])) {
     GPBMessageFieldDescription *coreDesc;
@@ -524,12 +525,17 @@ uint32_t GPBFieldAlternateTag(GPBFieldDescriptor *self) {
 
     // Extra type specific data.
     if (isMessage) {
-      const char *className = coreDesc->dataTypeSpecific.className;
       // Note: Only fetch the class here, can't send messages to it because
       // that could cause cycles back to this class within +initialize if
       // two messages have each other in fields (i.e. - they build a graph).
-      msgClass_ = objc_getClass(className);
-      NSAssert(msgClass_, @"Class %s not defined", className);
+      if (usesClassRefs) {
+        msgClass_ = coreDesc->dataTypeSpecific.clazz;
+      } else {
+        // Backwards compatibility for sources generated with older protoc.
+        const char *className = coreDesc->dataTypeSpecific.className;
+        msgClass_ = objc_getClass(className);
+        NSAssert(msgClass_, @"Class %s not defined", className);
+      }
     } else if (dataType == GPBDataTypeEnum) {
       if ((coreDesc->flags & GPBFieldHasEnumDescriptor) != 0) {
         enumHandling_.enumDescriptor_ =
@@ -559,6 +565,15 @@ uint32_t GPBFieldAlternateTag(GPBFieldDescriptor *self) {
     }
   }
   return self;
+}
+
+- (instancetype)initWithFieldDescription:(void *)description
+                         includesDefault:(BOOL)includesDefault
+                                  syntax:(GPBFileSyntax)syntax {
+  return [self initWithFieldDescription:description
+                        includesDefault:includesDefault
+                          usesClassRefs:NO
+                                 syntax:syntax];
 }
 
 - (void)dealloc {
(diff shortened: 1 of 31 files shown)


src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -414,9 +414,28 @@ void FileGenerator::GenerateSource(io::Printer *printer) {
...
+		printer->Print(
+        "#pragma mark - Objective C Class references\n"
+        "// This somewhat arcane code forces linkage of classes from static archives by\n"
+        "// adding a concrete reference to the classes.\n"
+        "// We don't use `[Foo class]` because we need a static value for our initializer.\n"
+        "// This also has the added benefit of reducing size in that we don't have to\n"
+        "// encode the class names and look them up at runtime.\n");
+  }
+  for (const auto& i : fwd_decls) {
+    printer->Print("$value$;\n", "value", i);
```
thomasvl: This flow seems a little odd/cryptic.

You call something to collect a set of things, that the name implies are some sorta forward decl, but when you print them, you know to add the `;` on the end.  It almost seems like they should be complete, i.e. - maybe just add the newline, but the ending semi should be within them, or you should only be collect the names of the classes, and the full formatting of what to do with the name should exist here.

also, by just collecting the names, each leaf doing an override wouldn't have to know the asm magic, they would just have to list the class, and the asm magic would all be in one place.

src/google/protobuf/compiler/objectivec/objectivec_message.cc
```diff
@@ -504,7 +518,7 @@ void MessageGenerator::GenerateSource(io::Printer* printer) {
     printer->Print(
         vars,
         "    GPBDescriptor *localDescriptor =\n"
-        "        [GPBDescriptor allocDescriptorForClass:[$classname$ class]\n"
+        "        [GPBDescriptor allocDescriptorForClass:self\n"
```
thomasvl: Please hold off on this change, if someone subclassed a message, this will latch onto that class if it is the first instance used for a message. 

To do this, we need to first tag everything as not subclass-able, to sure this change doesn't go wrong (and making that a distinct commit with this will be cleaner).

objectivec/GPBDescriptor_PackagePrivate.h
```diff
@@ -80,7 +80,7 @@ typedef struct GPBMessageFieldDescription {
   // Name of ivar.
   const char *name;
   union {
-    const char *className;  // Name for message class.
+    Class clazz;  // Class of the message.
```
thomasvl: Actually, this is a breaking change.

If someone has sources generated and then updates to a runtime with this change, things won't work.  so you likely need to keep the old union member there, add this one, and then figure out a way to set a flag to the startup flows to let the runtime know which union item to expect and react accordingly.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
dmaclach: Fixed somewhat :)

I will centralize the asm magic in a later change that includes the extensions. I fixed it so that the semi is included in them.

src/google/protobuf/compiler/objectivec/objectivec_message.cc
dmaclach: Fixed

objectivec/GPBDescriptor_PackagePrivate.h
dmaclach: Fixed, and a test was added.

dmaclach: Fixed, and added a test.

dmaclach:
Apologies for the dual comments... sigh github.

objectivec/GPBDescriptor_PackagePrivate.h
```diff
@@ -69,6 +69,10 @@ typedef NS_OPTIONS(uint16_t, GPBFieldFlags) {
   GPBFieldMapKeySFixed64 = 10 << 8,
   GPBFieldMapKeyBool     = 11 << 8,
   GPBFieldMapKeyString   = 12 << 8,
+
+  // This is used as a stopgap as we move from using class names in the
+  // GPBMessageFieldDescriptions to using class references.
+  GPBFieldUsesClazz       = 1 << 12
```
thomasvl: Did you look at using GPBDescriptorInitializationFlags so it just gets passed to the MessageDescriptor?  You'd still have to relay it to the fields, but it keeps things out of the field flags for the time being.  Not sure if that's better or not, just figured I'd call it out incase you didn't look at it.

src/google/protobuf/compiler/objectivec/objectivec_map_field.cc
```diff
@@ -170,6 +174,19 @@ void MapFieldGenerator::DetermineForwardDeclarations(
...
 
+void MapFieldGenerator::DetermineStaticForwardDeclarations(
+    std::set<string>* fwd_decls) const {
+  // Class name is already in "storage_type".
+  const FieldDescriptor* value_descriptor =
+      descriptor_->message_type()->FindFieldByName("value");
+  if (GetObjectiveCType(value_descriptor) == OBJECTIVECTYPE_MESSAGE) {
+    fwd_decls->insert("extern const struct {} " +
+                      value_field_generator_->variable("class_reference") +
+                    " __asm__(\"_OBJC_CLASS_$_\" GPBStringify(" +
```
thomasvl: indent is a little funky here.

the value to insert is likely worth a local just to make this a little more readable.

objectivec/google/protobuf/Type.pbobjc.m
```diff
@@ -29,6 +29,22 @@
...
+// This somewhat arcane code forces linkage of classes from static archives by
+// adding a concrete reference to the classes.
+// We don't use `[Foo class]` because we need a static value for our initializer.
+// This also has the added benefit of reducing size in that we don't have to
+// encode the class names and look them up at runtime.
+extern const struct {} GPBEnumValue_GPBOption_Class __asm__("_OBJC_CLASS_$_" GPBStringify(GPBOption));
+extern const struct {} GPBEnum_GPBEnumValue_Class __asm__("_OBJC_CLASS_$_" GPBStringify(GPBEnumValue));
+extern const struct {} GPBEnum_GPBOption_Class __asm__("_OBJC_CLASS_$_" GPBStringify(GPBOption));
+extern const struct {} GPBEnum_GPBSourceContext_Class __asm__("_OBJC_CLASS_$_" GPBStringify(GPBSourceContext));
+extern const struct {} GPBField_GPBOption_Class __asm__("_OBJC_CLASS_$_" GPBStringify(GPBOption));
```
thomasvl: Why extern?  Can we make them static so they are scoped to this file?

So we actually fetch GPBOption three times here.  Rather than names them for the message class and then the field type, if we can scope them just to this file, why not use use the type, then you only need GPBOption once.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -414,9 +414,28 @@ void FileGenerator::GenerateSource(io::Printer *printer) {
     printer->Print(
         "#pragma clang diagnostic ignored \"-Wdirect-ivar-access\"\n");
   }
-
   printer->Print(
-      "\n"
+		"\n");
```
thomasvl: odd indent

objectivec/GPBDescriptor.m
```diff
@@ -524,12 +524,17 @@ - (instancetype)initWithFieldDescription:(void *)description
...
-      const char *className = coreDesc->dataTypeSpecific.className;
       // Note: Only fetch the class here, can't send messages to it because
       // that could cause cycles back to this class within +initialize if
       // two messages have each other in fields (i.e. - they build a graph).
-      msgClass_ = objc_getClass(className);
-      NSAssert(msgClass_, @"Class %s not defined", className);
+      if ((coreDesc->flags & GPBFieldUsesClazz) != 0) {
+        msgClass_ = coreDesc->dataTypeSpecific.clazz;
+      } else {
+        // Backwards compatibility for old runtime.
```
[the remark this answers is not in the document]
dmaclach: Fixed

src/google/protobuf/compiler/objectivec/objectivec_message_field.cc
```diff
@@ -72,6 +74,13 @@ void MessageFieldGenerator::DetermineForwardDeclarations(
   fwd_decls->insert("@class " + variable("storage_type"));
 }
 
+void MessageFieldGenerator::DetermineStaticForwardDeclarations(
+    std::set<string>* fwd_decls) const {
+  fwd_decls->insert("extern const struct {} " + variable("class_reference") +
+                    " __asm__(\"_OBJC_CLASS_$_\" GPBStringify(" +
```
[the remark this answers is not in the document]
dmaclach: Fixed

src/google/protobuf/compiler/objectivec/objectivec_message_field.cc
```diff
@@ -100,6 +109,12 @@ void RepeatedMessageFieldGenerator::DetermineForwardDeclarations(
   fwd_decls->insert("@class " + variable("storage_type"));
 }
 
+void RepeatedMessageFieldGenerator::DetermineStaticForwardDeclarations(
+    std::set<string>* fwd_decls) const {
+  fwd_decls->insert("extern const struct {} " + variable("class_reference") +
+                    " __asm__(\"_OBJC_CLASS_$_\" GPBStringify(" +
```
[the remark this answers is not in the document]
dmaclach: Fixed

objectivec/GPBDescriptor_PackagePrivate.h
dmaclach: I did.. I felt that temporarily hijacking the field flags was clearer and easier to test. I needed to do it for extensions as well in an upcoming CL.

src/google/protobuf/compiler/objectivec/objectivec_map_field.cc
dmaclach: Fixed indent. The actual use of this will get cleaned up in the next CL, so for the sake of keeping merges on different branches easy I would prefer to keep this as is for now.

objectivec/google/protobuf/Type.pbobjc.m
dmaclach: Interestingly they need to be extern or the compiler crashes. I will be filing a bug with Apple. My next CL cleans up the repetition of the types.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
dmaclach: Fixed

objectivec/GPBDescriptor_PackagePrivate.h
thomasvl: While I'd love this to really be _temporarily_, the truth is it will live until we decide to take a breaking change.  🙁

objectivec/google/protobuf/Type.pbobjc.m
dmaclach: This is now fixed.

src/google/protobuf/compiler/objectivec/objectivec_map_field.cc
dmaclach: This is now fixed.

objectivec/GPBDescriptor_PackagePrivate.h
dmaclach: Moved things over to GPBDescriptorInitializationFlags

objectivec/GPBDescriptor.m
```diff
@@ -561,6 +567,15 @@ - (instancetype)initWithFieldDescription:(void *)description
   return self;
 }
 
+- (instancetype)initWithFieldDescription:(void *)description
```
thomasvl: I don't think this is actually needed since all calls here should be from within the runtime?  If they have to copies loaded, other things are likely to go wrong.  But I guess it doesn't hurt to play it safe.

objectivec/google/protobuf/Duration.pbobjc.m
```diff
@@ -92,7 +92,7 @@ + (GPBDescriptor *)descriptor {
                                         fields:fields
                                     fieldCount:(uint32_t)(sizeof(fields) / sizeof(GPBMessageFieldDescription))
                                    storageSize:sizeof(GPBDuration__storage_)
-                                         flags:GPBDescriptorInitializationFlag_None];
+                                         flags:GPBDescriptorInitializationFlag_UsesClassRefs];
```
thomasvl: I guess in theory we could not pass this, but also no harm in doing so.

objectivec/Tests/GPBMessageTests+ClassNames.m
```diff
@@ -0,0 +1,164 @@
+// Protocol Buffers - Google's data interchange format
```
thomasvl: Since this is new, you will need to add it to `Makefile.am` with the others

dmaclach:
> Looks like use of `asm` is going to cause problem, see [google/google-api-objectivec-client-for-rest#322](https://github.com/google/google-api-objectivec-client-for-rest/issues/322) where this just came up.

Check out latest patch. It should clear bitcode validation now as we are not declaring any new identifiers.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -402,21 +402,46 @@ void FileGenerator::GenerateSource(io::Printer *printer) {
...
   //    another file that is deprecated.
+  //  dollar-in-identifier-extension is needed because we use references to
+  //    objc class names that have $ in identifiers.
   printer->Print(
       "// @@protoc_insertion_point(imports)\n"
       "\n"
       "#pragma clang diagnostic push\n"
-      "#pragma clang diagnostic ignored \"-Wdeprecated-declarations\"\n");
+      "#pragma clang diagnostic ignored \"-Wdeprecated-declarations\"\n"
+      "#pragma clang diagnostic ignored \"-Wdollar-in-identifier-extension\"\n");
```
thomasvl: move this under a `if (fwd_decls.size() > 0) {` so it is only added when needed (like the direct-ivar-access one)

dmaclach: Done

objectivec/Tests/GPBMessageTests+ClassNames.m
dmaclach: Done

Issue #7029, opened 2019-12-18
Ruby: Issue decoding wrapped values after upgrading from 3.10.1 to 3.11.1


makushline:
```bash
$ cat repo.rb
require 'google/protobuf'
require 'google/protobuf/wrappers_pb'
Google::Protobuf::DescriptorPool.generated_pool.build do
  add_message "Person" do
    optional :name, :message, 1, "google.protobuf.StringValue"
  end
end
Person  = Google::Protobuf::DescriptorPool.generated_pool.lookup("Person").msgclass
puts "before decoding"
puts hi = Person.new(:name => {:value => "hi"})
puts empty = Person.new(:name => {:value => ""})
puts not_set = Person.new()
puts "after decoding"
puts Person.decode(hi.to_proto)
puts Person.decode(empty.to_proto)
puts Person.decode(not_set.to_proto)

// at google-protobuf 3.11.0
$ ruby repo.rb 
before decoding
<Person: name: <Google::Protobuf::StringValue: value: "hi">>
<Person: name: <Google::Protobuf::StringValue: value: "">>
<Person: name: nil>
after decoding
<Person: name: <Google::Protobuf::StringValue: value: "hi">>
<Person: name: nil>
<Person: name: nil>

// at google-protobuf 3.10.1
$ ruby repo.rb
before decoding
<Person: name: <Google::Protobuf::StringValue: value: "hi">>
<Person: name: <Google::Protobuf::StringValue: value: "">>
<Person: name: nil>
after decoding
<Person: name: <Google::Protobuf::StringValue: value: "hi">>
<Person: name: <Google::Protobuf::StringValue: value: "">>
<Person: name: nil>

$ cat Gemfile.lock
source "https://rubygems.org"
gem "google-protobuf", "3.10.1"
```

I expect the decoded Person with an empty name to look like
`<Person: name: <Google::Protobuf::StringValue: value: "">>`

haberman:
Could you try this PR and verify that it fixes your problem?

https://github.com/protocolbuffers/protobuf/pull/7195

PR #7037, opened 2019-12-19
Ruby - Add failing tests for issue with deserializing wrapped values

djquan:
This is in reference to https://github.com/protocolbuffers/protobuf/issues/7029.

The expected behavior is 
1.) wrapped values with the default value set should deserialize to a wrapped value with the default value set, but instead, they are being set to nil.

The first test fails with:
`Error: test_map_wrappers_with_default_values(BasicTest::MessageContainerTest): NoMethodError: undefined method value for nil:NilClass`, which matches what https://github.com/protocolbuffers/protobuf/issues/7029 is seeing.

2.) The second test I added for completeness.  A test that asserts the correct behavior for wrapped values with no value set.

The test structure followed the test introduced by @haberman  here, https://github.com/protocolbuffers/protobuf/pull/6797.  My understanding of C is limited, but that PR seems to be  one of the PRs that touched the Ruby code between the upgrade from 3.10.1 to 3.11.0, where this issue was introduced.

### DIFF
--- a/ruby/tests/basic.rb
+++ b/ruby/tests/basic.rb
@@ -318,6 +318,43 @@ module BasicTest
       assert_equal m5, m
     end
 
+    def test_map_wrappers_with_no_value
+      run_asserts = ->(m) {
+        assert_equal nil, m.map_double[0]
+        assert_equal nil, m.map_float[0]
+        assert_equal nil, m.map_int32[0]
+        assert_equal nil, m.map_int64[0]
+        assert_equal nil, m.map_uint32[0]
+        assert_equal nil, m.map_uint64[0]
+        assert_equal nil, m.map_bool[0]
+        assert_equal nil, m.map_string[0]
+        assert_equal nil, m.map_bytes[0]
+      }
+
+      m = proto_module::Wrapper.new(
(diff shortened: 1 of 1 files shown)


djquan:
@googlebot I signed it!

haberman:
Thank you for these test cases. I incorporated them into my fix in https://github.com/protocolbuffers/protobuf/pull/7195.

Could you give your consent to have your changes incorporated into that PR?

Some of the test cases weren't quite right though. Map values are always present, so `m.map_double[0]` will never return `nil`. I corrected that in my PR.

Thanks for submitting and sorry the fix was a bit slow.

djquan:
@haberman Looks great, thanks! I consent to having my changes incorporated in #7195 

Thanks again!

haberman:
@djquan for the system to recognize it, you need to put `@googlebot I consent` into a message on https://github.com/protocolbuffers/protobuf/pull/7195.

See: https://github.com/protocolbuffers/protobuf/pull/7195#issuecomment-584464100

djquan:
Ahh @haberman , gotcha, done!

haberman:
@djquan Thanks!

PR #7038, opened 2019-12-19
Fix python release breakages by set MB_PYTHON_OSX_VER

anandolee:


kokoro/release/python/macos/build_artifacts.sh
```diff
@@ -50,8 +50,12 @@ build_artifact_version() {
   mv wheelhouse/* $ARTIFACT_DIR
 }
 
+export MB_PYTHON_OSX_VER=10.9
 build_artifact_version 2.7
```
TeBoring: Move this to below?

PR #7039, opened 2019-12-20
DO NOT MERGE: Public version of FixedDecimal proposal

jtattermusch:
This is basically a copy of the internal proposal (as we currently don't have a open source proposal process in place for protobuf).

See https://github.com/protocolbuffers/protobuf/issues/4406

### DIFF
--- a/docs/proposal_decimal_wkt.md
+++ b/docs/proposal_decimal_wkt.md
@@ -1,0 +1,348 @@
+# Protobuf "decimal" well known type
+
+
+Created: | 2019-11-14
+
+State:   | PROPOSED
+
+Author:  | jtattermusch@
+
+
+## Summary
+
+Proposes a new "decimal" well known type as a standard representation for
+accurate fractional (e.g. monetary) values and provide language bindings for
+languages that have built-in language support for such a type.
+
+## Motivation
+
+"decimal" type is important for many applications (especially financial
+applications) that need to express fractional values (such as monetary amounts)
+accurately. While many programming languages offer a built-in support for
+"decimal" type, protobuf currently doesn't offer an easy way to utilize those
+built-in types and users are forced to resort to workarounds and to reinvent the
+wheel.
+
+Better support for decimal type has been requested at
+https://github.com/protocolbuffers/protobuf/issues/4406 and is heavily upvoted
+(but there isn't much progress on the issue).
+
+## Detailed Explanation
+
+### Requirements for the new decimal well known type
+
+-   introduce the "decimal" type as a "well-known type" (as opposed to extending
+    the protobuf wire protocol, which would be way too complicated). Protobuf
+    library implementations in different languages will then provide
+    language-specific bindings and helpers to integrate with the native
+    "decimal" type supported by given language.
+-   Provide message representation that is reasonably usable if the well known
+    type doesn't have special handling in given language (= not all languages
+    will have a special binding for the new well known type and the type must
+    still be usable in its "raw form" as a protobuf message).
+-   Match semantics of decimal type in popular programming languages (if
+    possible, in different languages the built-in decimal type can have slightly
+    different properties).
+-   Stay consistent with other pre-existing commonly used .proto types so that
+    the new WKT fits well into the existing ecosystem (e.g. there's a
+    money.proto)
+
+### Rollout plan
+
+-   Decide which .proto definition of the "decimal" type approach to use (see
+    below)
+-   Check in the well known type .proto definition of the type
+-   Language owners can provide their own bindings for the WKT
+-   Because the "decimal" WKT will be designed to be usable even without special
+    language bindings, initially the language-specific bindings for the WKT can
+    be marked as "experimental" in each of the languages and they can be later
+    stabilized once they prove themselves.
+
+### Proposal: message with "units" and "nanos"
+
+```
+// file:fixed_decimal.proto
+package google.protobuf;
+
+
+message FixedDecimal {
+
+    // Whole units part of the amount
+    int64 units = 1;
+
+    // Nano units of the amount (10^-9)
+    // Must be same sign as units
+    // Example: The value -1.25 is represented as units=-1 and nanos=-250000000
+    sfixed32 nanos = 2;
+}
+```
+
+Pros
+
+-   simple concept, easy to understand for users
+-   values are human readable and easy to understand without specialized support
+-   Very similar to .proto definitions that have already been used internally at
+    google
+-   Consistent with other widely used .proto definitions (such as money.proto)
+-   range of values can be safely expressed by "decimal" type in all languages
+    that support such type.
+-   representable value range should be good enough to vast majority of
+    applications (e.g. for financial applications)
+-   Currently this is recommended for C# users by Microsoft (= which gives some
+    indication that this approach works well in the .NET ecosystem).
+    https://docs.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protobuf-data-types#decimals
+
+Cons
+
+-   lower range and resolution than "decimal" types in all languages (the good
+    news is all common decimal types can represent the values carried by the
+    proto, but converting from lang-specific decimal values can lose precision /
+    throw exception)
+
+The type of "units" can be either `int64` or `sint64`. `sint64` is more
+efficient if negative values are common, but
+`money.proto`
+uses `int64`. So the choice is either staying more consistent with `money.proto`
+or aiming for the highest efficiency.
+
+The type of "nanos" can be either `int32`, `sfixed32` or `sint32`. `int32` is
+what `money.proto` uses, but the signed types will be more efficient for
+negative values. One advantage of `sfixed32` over `sint32` is that the "nanos"
+value is likely to be a large number (numerical values like 0.5, 0.25 etc. are
+more common than e.g. 0.000000001), and so `sfixed32` has potential to be more
+efficient that `sint32` or `int32` (we'd need to do some measurements).
+
+The sign of "units" and "nanos" is required to be the same (money.proto uses the
+same restriction). Nevertheless, there are other types (e.g. timestamp.proto),
+where the "nanos" part is always positive. The former seems to make more sense
+for representing decimal numbers, because e.g. -7.5 will be represented as
+`units=-7` and `nanos=-500000000` and the represented value can always be
+computed as `units + 0.000000001 * nanos` regardless of the sign.
+
+TODO(jtattermusch): do the signs of units vs nanos influence conversion speed
+between the message fields and the numeric representation? it might for decimal
+-> FixedDecimal direction of conversion
+
+The message name `FixedDecimal`(in `fixed_decimal.proto`) is chosen to make the
+nature of the WKT explicit (and to express that this is not necessarily a 1:1 of
+language's "decimal" type).
+
+### API Changes
+
+For languages that provide a built-in "decimal" type (see overview in Appendix),
+language owners should add an API that allows easy conversion between
+`FixedDecimal` WKT and the built-in "decimal" type in given language.
+
+The API changes in each language should be purely additive. It's fine for
+language implementations to provide the bindings independent of other languages
+(e.g. C# might provide the bindings sooner than some other language).
+
+The value range representable by the proposed FixedDecimal message (at most ~29
+significant digits) is smaller than the representable range for all the
+language-specific "decimal" types. Therefore, there should be no issue
+converting FixedDecimal message to any of the decimal types. In the opposite
+direction (setting a value from a "decimal" type into a FixedDecimal), there can
+be an "out of range" error and language implentations should be consistent in
+how this situation is handled.
+
+Language specific-APIs should generally allow accessing the FixedDecimal values
+in two modes: 1. access the raw FixedDecimal value (the "units" and "nanos"
+fields) 2. access the language-specific value as a "decimal" type
+
+#### C# bindings
+
+```
+// extra methods for FixedDecimal generated class
+public partial class FixedDecimal
+{
+    private const decimal NanoFactor = 1_000_000_000;
+
+    public decimal ToDecimal()
+    {
+        // IsNormalized checks for Units and Nanos having the same sign
+        // and Nanos being from the right range.
+        if (!IsNormalized(Units, Nanos))
+        {
+            throw new InvalidOperationException(@"Fixed decimal contains invalid values: Units={Units}; Nanos={Nanos}");
+        }
+        return Units + Nanos / NanoFactor;
+    }
+
+    public static FixedDecimal FromDecimal(decimal value)
+    {
+        // ToInt64() throws OverflowException if value is out of range of int64
+        var units = decimal.ToInt64(value);
+        var nanos = decimal.ToInt32((value - units) * NanoFactor);
+        return new FixedDecimal { Units = units, Nanos = nanos };
+    }
+}
+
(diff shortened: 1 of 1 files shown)


jtattermusch:
@JamesNK @JunTaoLuo @shirhatti FYI

mgravell:
The `units` being `int64` and `nanos` being `sfixed32` could perhaps make negative values unnecessarily large? Any reason not to go `sint64` for `units`? (given the overhead of a message header, the potential extra byte in some zig-zag edge cases of positive numbers seems marginal)

`nanos` says "Must be same sign as units" - probably needs a "when not zero"? i.e. `-12345.0` would be a negative and non-negative (integers don't usually have both positive and negative zero).

jtattermusch:
> The `units` being `int64` and `nanos` being `sfixed32` could perhaps make negative values unnecessarily large? Any reason not to go `sint64` for `units`? (given the overhead of a message header, the potential extra byte in some zig-zag edge cases of positive numbers seems marginal)
> 

this is discussed in the proposal:
 *The type of "units" can be either `int64` or `sint64`. `sint64` is more efficient if negative values are common, but `money.proto` uses `int64`. So the choice is either staying more consistent with `money.proto` or aiming for the highest efficiency."*

I understand you're leaning more towards using sint64. How much would of an overhead is the zigzag encoding itself in your opinion?
Also, it is hard to tell in advance how often will users want negative values - for financial applications, I assume messages with positive numbers only could be quite common?

> `nanos` says "Must be same sign as units" - probably needs a "when not zero"? i.e. `-12345.0` would be a negative and non-negative (integers don't usually have both positive and negative zero).

You're right but this probably won't lead to any confusion (intuitively integer 0 has "no sign" and it's fair to use it for -123.0)

mgravell:
>  How much would of an overhead is the zigzag encoding itself in your opinion?

My understanding without checking the spec is that this costs *at most one bit*, which will in 1/7th of the cases be enough to cost us an extra byte (with small magnitude numbers still being cheaper than large magnitude). Compared to negative without zig-zag, which costs a hard 10 bytes regardless of size. Is that the question you were asking?

jtattermusch:
> > How much would of an overhead is the zigzag encoding itself in your opinion?
> 
> My understanding without checking the spec is that initially this costs _at most one bit_, which will in 1/7th of the cases be enough to cost us an extra byte (with small magnitude numbers still being cheaper than large magnitude). Compared to negative without zig-zag, which costs a hard 10 bytes regardless of size. Is that the question you were asking?

I meant the computation overhead of performing the zigzag transformation.
I looked it up and it's very likely negligible: DecodeZigZag64 does this `return (long)(n >> 1) ^ -(long)(n & 1);`

mgravell:
oh, indeed: computationally it is very CPU-friendly - no branching, so it should just pipeline beautifully

vivainio:
Since there is no particular need to pessimize negative numbers, and negative numbers are not rare in financial applications - can the proposal be updated to have type (sint64,  sfixed32)?

I'm not sure how much consistence with money.proto matters here. This is generic decimal number support, not specific to money, and money type may not have much need for negative numbers.

Also @jtattermusch are there any updates where the discussion is going within Google?

Fleshgrinder:
100% compatibility with [`google.type.Money`](https://github.com/googleapis/api-common-protos/blob/master/google/type/money.proto) is very important (speaking as someone who is currently working in the banking sector). The reason is that there are often definitions that do not use the money definition because they would basically repeat the currency multiple times. Think of a simple account balance message:

```proto
message AccountBalance {
    string currency = 1;
    google.protobuf.FixedDecimal available = 2;
    google.protobuf.FixedDecimal usable = 3;
}
```

Here _available_ represents the money in the account and _usable_ how much the owner is actually able to use (which can be more than is available in case the owner has overdraft). Using money in this case makes little sense but being able to switch between the decimal and money is very important for writing abstractions and what not.

vivainio:
@Fleshgrinder this makes sense. I forgot that someone may want to read Money wire protocol to FixedDecimal

flozano:
Whatever is the outcome of this, it should be supported in `Value.proto` so that it can be used inside Structs and Lists and conversion of untyped messages to/from JSON is actually possible without loss of information.

jufemaiz:
Interested to hear the likely timeline on acceptance and publication for this PR.

jtattermusch:
This is still under internal review and we're considering alternative representations of the type.

jufemaiz:
Thanks @jtattermusch !

vivainio:
Montly "ping" for the status :)

jufemaiz:
Dumb question perhaps, but the following claim is made:

> nanos must be same sign as units

How much support is there for a "negative zero" value in the units or the nanos (depending upon the number)? Or is zero considered to have "no sign" so to speak?

jtattermusch:
After a lot of internal discussions, there are some new type defintions that we want to evaluate.
Basically there are a few layouts modeled after IEEE754-2008 decimal types (Decimal64 and Decimal128 - these two types seem to be the future standards for decimal types so it make sense to try to reuse them for protobuf's decimal WKT).
and there is also a "DecimalString" message that is perhaps little harder to (de)serialize, but the representation is very human-friendly and it has an arbitrary range of representable values (but the concern is (de)serialization speed and storage efficiency).

```
/*
 * A decimal number represented by a string.
 *
 * The string format must match this syntax:
 * number    = [ "+" | "-" ] float .
 * number    = "nan" | "+inf" | "-inf" .
 * float     = digits [ "." digits ] [ exponent ] .
 * float     = "." digits [ exponent ] .
 * exponent  = "e" [ "+" | "-" ] digits .
 * digits    = digit { digit } .
 * digit     = "0" ... "9" .
 *
 *
 * TODO: add documentation about which value range is representable
 * by all protobuf languages.
 */
message DecimalString {
  string value = 1;
}

message Decimal64 {
  // matches IEEE 754-2008 decimal64

  // interpreting the values:
  // both significand and exponent within allowed range of IEEE decimal64,
  // exponent in range -> the value is significand * 10^exponent. significand
  // exactly [-10^17, 10^17], exponent is 0 (i.e. not present) -> +/-Infinity
  // significand out of range OR exponent out of range -> NaN

  // Allowed range is [-10^17, 10^17]. Exceeding these bounds is an error.
  // A significand of exactly -10^17/10^17 represent +/- Infinity.
  sint64 significand = 1;

  // Allowed range is [−383, 384].
  sint32 exponent = 2;
}

message Decimal128 {
  // matches IEEE 754-2008 decimal128

  // interpreting the values:
  // both significand and exponent within allowed range of IEEE decimal128,
  // exponent in range -> the value is sign * significand * 10^exponent.
  // significand exactly 10^34, exponent is 0 (i.e. not present) -> +/-Infinity
  // significand out of range OR exponent out of range OR sign out of range ->
  // NaN

  sint32 sign = 1;  // -1 for negative, +1 for positive, 0 for zero. If sign is
                    // zero, the significand must also be zero.

  // low and high 64 bits of the significand
  // allowed range is [0, 10^34]. Exceeding these bounds is an error.
  // A significand of exactly 10^34 represents +/- Infinity (based on sign)
  uint64 significand_low = 2;
  uint64 significand_high = 3;

  // Allowed range is [−6143,+6144].
  sint32 exponent = 4;
}

message Decimal64Version2 {
  // matches IEEE 754-2008 decimal64
  // Allowed range is (-10^17, 10^17). Exceeding these bounds is an error.
  sint64 significand = 1;

  // TODO: should we use oneof? the issue is that if special_value is used, both
  // significand and exponent are supposed to be 0 (not just the exponent), at
  // which point there's not that much value in having a oneof.

  // Interpretation.  If none of these are set then it will be
  // interpreted as if "exponent = 0".
  oneof mode {
    sint32 exponent = 2;  // Allowed range is [−383, 384].

    // if special_value is set, both significand and exponent must be zero (not
    // present)
    SpecialEncoding special_value = 3;
  }
}

message Decimal128Version2 {
  // matches IEEE 754-2008 decimal128

  bool negative =
      1;  // true for negative numbers, false for positive numbers and zero

  // low and high 64 bits of the significand
  // allowed range is [0, 10^34). Exceeding these bounds is an error.
  uint64 significand_low = 2;
  uint64 significand_high = 3;

  // Interpretation.  If none of these are set then it will be
  // interpreted as if "exponent = 0"
  oneof mode {
    // Must be zero when special_value is set
    sint32 exponent = 4;  // Allowed range is [−6143,+6144] -- TBD

    // if special_value is set, all of significand, negative and exponent must
    // have their default value (not present)
    SpecialEncoding special_value = 5;
  }
}

enum SpecialEncoding {
  SPECIAL_VALUE_UNSPECIFIED = 0;
  NAN = 1;
  POSITIVE_INFINITY = 2;
  NEGATIVE_INFINITY = 3;
}
```

@JamesNK  would you be interested in trying to prototype the ToDecimal/FromDecimal C# language bindings and benchmarking their performance / memory efficiency using BenchmarkDotNet?

JamesNK:
Sure. I'll try to do it sometime this week.

TrayanZapryanov:
I am sorry if I miss something in the discussion, but currently we are using following format and functions to encode/decode decimals in our software. I didn't invented them - just copied somewhere over the net and they are working till now.
So I am putting here just as an information:
proto schema:
message Decimal 
  {
	uint64 lo = 1; // the first 64 bits of the underlying value
	uint32 hi = 2; // the last 32 bis of the underlying value
	uint32 signScale = 3; // the number of decimal digits (bits 1-16), and the sign (bit 0)
  }

converters:

public static Protobuf.WellKnownTypes.Decimal ToProtoDecimal(this decimal value)
        {
            var bits = decimal.GetBits(value);
            ulong a = ((ulong) bits[1]) << 32, b = ((ulong) bits[0]) & 0xFFFFFFFFL;
            var low = a | b;
            var high = (uint) bits[2];
            var signScale = (uint) (((bits[3] >> 15) & 0x01FE) | ((bits[3] >> 31) & 0x0001));

            return new Protobuf.WellKnownTypes.Decimal
            {
                Lo = low,
                Hi = high,
                SignScale = signScale
            };
        }

        public static decimal ToDecimal(this Protobuf.WellKnownTypes.Decimal dec)
        {
            var low = dec.Lo;
            var high = dec.Hi;
            var signScale = dec.SignScale;

            if (low == 0 && high == 0) return decimal.Zero;

            int lo = (int) (low & 0xFFFFFFFFL),
                mid = (int) ((low >> 32) & 0xFFFFFFFFL),
                hi = (int) high;

            var isNeg = (signScale & 0x0001) == 0x0001;
            var scale = (byte) ((signScale & 0x01FE) >> 1);

            return new decimal(lo, mid, hi, isNeg, scale);
        }

mgravell:
@TrayanZapryanov that is from protobuf-net, and as the person who wrote that, I say officially "oh gawd, burn it with fire, it hurts, it hurts". Or rather: "that is a decision that I regret, and it is horrible". More recently, protobuf-net uses a "compatibility level" concept that allows me to change conventions over time (without breaking existing code), and at the more recent compatibility levels: it does not use that layout.

TrayanZapryanov:
@mgravell Thank you for pointing me this. Yes I can remember now and will check how things developed in protobuf-net repo.
Sorry for extending discussion here - just wanted to help somehow.

mgravell:
@TrayanZapryanov here's the write-up: https://github.com/protobuf-net/protobuf-net/blob/main/docs/compatibilitylevel.md

D3MaxT:
Impatiently waiting on this in C# 👍

vivainio:
Bump

akulich:
Hello from 2021. Did you give up on this?

St333p:
Any updates on this? I'd love to have support for python's decimals. If you're not planning to add it soon I'll go for a workaround, probably casting to strings. If at all possible I'd like to avoid many backwards incompatible changes to my API.

jufemaiz:
@St333p in the meantime we have opted to approach this with a units + nanos approach. It's not perfect but it's reasonable.

flozano:
Is this discarded? There will be no standardized decimal number support?

fowles:
I was cleaning up old branches and this one hasn't seen any motion for over a year.  

We definitely don't have short term plans to build this and are already stretched a bit thin on other large efforts.

flozano:
Thanks for the insight. Handling "money" values is just not possible with existing protobuf well-known types, it's a little sad. #4406 is important and the lack of this causes many people either to (wrongly) use float types for this kind of magnitude, or use custom non-interoperable messages to properly represent decimal values.

fowles:
I am sympathetic to where you are coming from here having worked on fintech in the past.  The best advice I can give here is to have strong internal style guidelines.  You should either create opaque types to handle money and have dedicated message for them with custom adaptors or store everything as an integral number of pips.

The closing of this is not a statement that the suggestion is a bad idea.  It is a simple statement of scare resources and trying to set expectations for what we are/are not planning to pursue right now.

flozano:
Thank you for your answer. It feels a little underwhelming and I was just expressing some frustration with how the protobuf well-known types are (not) evolving. I understand this is just because of lack of resources. This PR by @jtattermusch seemed a very good direction to me.

The problem here is everyone will be defining their own "opaque", non-interoperable message types. Handling numbers is not easy and the lack of this leaves open the door for bad implementations and/or bad decisions. 

This also leaks to the entire ecosystem built around protobuf... for example, automated grpc-gateway translation of those types for JSON exposure won't happen as there's no standard type to support... so there will be no straightforward "money" support for any grpc-defined API exposed as HTTP in a way that is easy to understand by developers consuming that HTTP api.

Fleshgrinder:
Hmmm… we have [Money](https://github.com/googleapis/api-common-protos/blob/master/google/type/money.proto), what's wrong with it? This PR was about a fixed decimal without currency. As I stated above, a fixed decimal definition ideally is simply compatible with Money.

The easiest workaround for the things you mentioned really is a string. JSON can handle it, every language can handle it.

That said, I'm also sad that this gets closed. 😿

flozano:
I didn't know this type existed 🤦 my use-case doesn't involve actual money but other type of fixed-decimal magnitudes, I was arguing for the "money" case because it seems easiest to understand. My bad.

Fleshgrinder:
I think that there are plenty of use cases for it, and nobody questions that, but especially the one you mentioned wouldn't be covered by this. JS(ON) has no support for 64 bit numbers, so anyone trying to do that there will end up with strings or funny surprises. The safest route really is a string here.

flozano:
Sorry for the offtopic. I understand it's JS that doesn't support 64-bit integer, but not JSON itself, both JSON document format and JSON schema are fine with arbitrary precision numbers.

Fleshgrinder:
Absolutely, but when you mention gateway I think of JS. If JS isn't involved, why the gateway and not gRPC? But yeah, getting too off topic. 😉

jufemaiz:
> Hmmm… we have [Money](https://github.com/googleapis/api-common-protos/blob/master/google/type/money.proto), what's wrong with it? This PR was about a fixed decimal without currency. As I stated above, a fixed decimal definition ideally is simply compatible with Money.
> 
> The easiest workaround for the things you mentioned really is a string. JSON can handle it, every language can handle it.
> 
> That said, I'm also sad that this gets closed. 😿

Sadly three decimal places is insufficient. Working in electricity sector, with normalised pricing in the 6-7 decimal places space.

Further, floating point issues for values that need decimal support is sorely needed.

:sadpanda: (I missed the closure of this issue).

JasonPolisAdmin:
Need support for Decimal128, so new fixed width wire type.

Workaround is to use <bytes> with additional validation, but would prefer native scalar Decimal128.

Big Decimal can also be represented with bytes, or else string.

PR #7048, opened 2019-12-25
Fix GCC build with PROTOBUF_USE_DLLS enabled

stha09:
GCC does not allow mixing __attribute__(()) syntax alignas()
syntax. Re-use approach from chromium base/compiler_specific.h
release notes: no


acozzette:
Thanks, @stha09!

PR #7052, opened 2019-12-27
Add Swift Package Manager support (Fixes #7047)

### LINKED ISSUE #7047: Feature Request: Swift Package Manager support
If it's a proto syntax change, is it for proto2 or proto3? N/A
If it's about generated code change, what programming language? N/A

I am trying to convert a repository from CocoaPods to Swift Package Manager, but this dependency doesn't support it.

I would like this repository to have a `Package.swift` file to support Swift Package Manager.

1. I have considered using something instead of Protobuf, but it currently works in the repository.
2. Additionally, I have considered not converting to Swift Package Manager, but it seems more practical than CocoaPods.
3. Finally, I have considered attempting to implement this myself, but picking out the correct source files proved to be difficult.

**Additional context**

turtlemaster19:
Creates `Package.swift` to fix #7047

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,26 @@
+// swift-tools-version:4.1
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+import PackageDescription
+
+let package = Package(
+    name: "Protobuf",
+    products: [
+        // Products define the executables and libraries produced by a package, and make them visible to other packages.
+        .library(name: "Protobuf", targets: ["Protobuf"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
+        .target(
+            name: "Protobuf",
+            dependencies: [],
+            path: "objectivec/",
+            exclude: ["ProtocolBuffers_OSX.xcodeproj", "ProtocolBuffers_iOS.xcodeproj", "ProtocolBuffers_tvOS.xcodeproj", "DevTools/", "Tests/", ".gitignore", "README.md", "generate_well_known_types.sh"],
+            sources: ["google/protobuf/"],
+            publicHeadersPath: "objectivec/"),
+    ]
+)


turtlemaster19:
@googlebot I signed it!

turtlemaster19:
Maintainers, please add cleanup label

Package.swift
```diff
@@ -0,0 +1,26 @@
...
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
+        .target(
+            name: "Protobuf",
+            dependencies: [],
+            path: "objectivec/",
```
thomasvl: How does `path`, `exclude`, and `sources` all work together?  There are sources needed that don't live under that `sources` reference.  I don't suppose there's any way to use globs to help cut down on what has to be explicitly listed?

Package.swift
```diff
@@ -0,0 +1,26 @@
...
+            exclude: ["ProtocolBuffers_OSX.xcodeproj", "ProtocolBuffers_iOS.xcodeproj", "ProtocolBuffers_tvOS.xcodeproj", "DevTools/", "Tests/", ".gitignore", "README.md", "generate_well_known_types.sh"],
+            sources: ["google/protobuf/"],
+            publicHeadersPath: "objectivec/"),
```
thomasvl: Also, how do we test things to be sure things are working?  We're also seeing some issue with CocoaPods around case sensitive/case insensitive file systems (#6803) and how it handles the path names.  So if we are looking to add SwiftPM support, it would be nice to sure things are working for all setups.

turtlemaster19:
One way you can test is by adding it as a package in a dependent Xcode project within Xcode. I don’t know what globs is, and there are a lot of files in this repository, but I just wanted to get something started if I could.

thomasvl:
> One way you can test is by adding it as a package in a dependent Xcode project within Xcode. I don’t know what globs is, and there are a lot of files in this repository, but I just wanted to get something started if I could.

globs are patterns like _*.m_ and some systems allow them to be recursive so you don't have to list paths.  They are used in the [podspec](https://github.com/protocolbuffers/protobuf/blob/master/Protobuf.podspec#L19) to make it easier to list what is needed without having to have something like the excludes.

As far as getting something started, once it lands, it has to be supported.  So we need to be sure it is going to work correctly for all cases, hence seeing what can be done to test it, etc. Otherwise, it might work for some folks but not others, and in some ways that ends up being worse than not having support.

turtlemaster19:
OK, I would love to see support, but I understand the need to make it work for all test cases. Happy New Year!

turtlemaster19:
Maintainers, please add `release notes: no` label

Package.swift
ydnar: @thomasvl unfortunately, no globbing. You can use more explicit `sources`, or explicit `exclude`. For example: https://github.com/getsentry/sentry-cocoa/pull/352/files

turtlemaster19:
Closing because:
1. It doesn't seem to be gaining acceptance
2. https://github.com/apple/swift-protobuf has had SwiftPM support since the beginning

If anyone believes it should be reopened, it's always a possibility!

PR #7056, opened 2019-12-30
[bazel] Be able to specify protobuf plugins in py_proto_library

moshe:
This PR adds the ability to specify custom proto plugins for the `py_proto_library` rule
I wanted to use https://github.com/dropbox/mypy-protobuf but it seems like it not possible with the current rule.
I made the PR backward compatible, let me know what to do think🙂

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -383,6 +383,8 @@ def py_proto_library(
         default_runtime = "@com_google_protobuf//:protobuf_python",
         protoc = "@com_google_protobuf//:protoc",
         use_grpc_plugin = False,
+        plugin = None,
+        plugin_language = None,
         **kargs):
     """Bazel rule to create a Python protobuf library from proto source files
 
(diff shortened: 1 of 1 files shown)


moshe:
@googlebot I signed it!

moshe:
ping?

moshe:
cc @Yannic

moshe:
You are right, I will fix it and better test it. Fix will be submitted in the next days.
Thanks @Yannic

acozzette:
I'm going to close this PR since there's been no activity in a few months. We actually have some work underway to have protoc generate pyi files, so that might address the motivation for this PR anyway.

micahjsmith:
@acozzette to close the loop on this thread - I believe protoc now generates pyi files using `--pyi_out`, but `py_proto_library` macro does not support that and it seems protobuf repo is no longer responsible for that macro (https://github.com/protocolbuffers/protobuf/issues/10366) - can anyone point to a way forward for people who want to use bazel + protoc + pyi_out?

acozzette:
We are going to keep our `py_proto_library` macro around until Bazel releases their own version, so I think we would OK accepting a pull request that adds support for `--pyi_out`, as long as it's compatible with the future Bazel version.

Issue #4165, opened 2018-01-11
ReferenceError: goog is not defined


shihabkaranchery:
i follow the steps in the docs, convert the .proto message to myproto_libs.js file while importing the file

following error occured 
ReferenceError: goog is not defined

crazed:
I also got this with the `es6` output style. Everything works fine with `commonjs` though.

gdm85:
I was also bitten by this bug and created PR #7059 to try to improve the situation a bit.

PR #7059, opened 2020-01-02
Remove es6 and browser import styles, update docs

gdm85:
See https://github.com/protocolbuffers/protobuf/issues/4165

I think that support for the `browser` import style was broken in 3b3c8abb9635eb3ea078a821a99c9ef29d66dff7 and I notice there are no tests for it (nor for `es6`, which is just a dummy option planned to be implemented from my understanding), thus the breakage slipped through unnoticed.

(To see the breakage, search in that commit for `js_generator.cc`, expand the large diff and search for `IMPORT_BROWSER` which is visible only in removed lines.)

I tried to fix this with a patch:
```diff
--- a/src/google/protobuf/compiler/js/js_generator.cc
+++ b/src/google/protobuf/compiler/js/js_generator.cc
@@ -1877,6 +1877,9 @@ void Generator::GenerateRequiresImpl(const GeneratorOptions& options,
                                      std::set<std::string>* provided,
                                      bool require_jspb, bool require_extension,
                                      bool require_map) const {
+  if (options.import_style == GeneratorOptions::kImportBrowser)
+    return;
+
   if (require_jspb) {
     required->insert("jspb.Message");
     required->insert("jspb.BinaryReader");
```
However this fails also because of an undefined `goog` object, so it was a dead end.

In this PR I propose remove the two unused `kImportBrowser` and `kImportEs6` and I update the documentation to not state that it is possible to use google-protobuf without npm.

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -9,7 +9,7 @@ This directory contains the JavaScript Protocol Buffers runtime library.
 
 The library is currently compatible with:
 
-1. CommonJS-style imports (eg. `var protos = require('my-protos');`)
+1. CommonJS-style imports (eg. `var protos = require('my-protos');`), which is the default
 2. Closure-style imports (eg. `goog.require('my.package.MyProto');`)
 
 Support for ES6-style imports is not implemented yet.  Browsers can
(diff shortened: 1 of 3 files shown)


gdm85:
There is an automated failure about labels; I cannot change labels so there is nothing I can do to fix that (would be nice if the bot message specified who is the audience of the error).

acozzette:

PR #7065, opened 2020-01-08
Blacklist .proto source files if Bazel allows us to

Yannic:
This is a partial revert of https://github.com/protocolbuffers/protobuf/commit/7b28278c7d4f4175e70aef2f89d304696eb85ae3 to unblock, e.g., https://github.com/grpc/grpc/pull/21590 or https://github.com/lyft/envoy-mobile/issues/617 until Bazel is fixed.

Note: this is a forward-compatible change that automatically switches to the behavior intended by https://github.com/protocolbuffers/protobuf/commit/7b28278c7d4f4175e70aef2f89d304696eb85ae3 when a compatible Bazel is released without requiring users to upgrade Protobuf. We will revert this change when Bazel is fixed.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -986,8 +986,8 @@ cc_library(
     ],
 )
 
-reject_blacklisted_files = not hasattr(proto_common, "proto_lang_toolchain_rejects_files_do_not_use_or_we_will_break_you_without_mercy"),
-cc_toolchain_blacklisted_protos = [proto + "_proto" for proto in WELL_KNOWN_PROTO_MAP.keys()] if reject_blacklisted_files else [":well_known_protos"],
+reject_blacklisted_files = not hasattr(proto_common, "proto_lang_toolchain_rejects_files_do_not_use_or_we_will_break_you_without_mercy")
+cc_toolchain_blacklisted_protos = [proto + "_proto" for proto in WELL_KNOWN_PROTO_MAP.keys()] if reject_blacklisted_files else [":well_known_protos"]
 proto_lang_toolchain(
     name = "cc_toolchain",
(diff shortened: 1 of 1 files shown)


BUILD
```diff
@@ -986,9 +986,11 @@ cc_library(
     ],
 )
 
+reject_blacklisted_files = not hasattr(proto_common, "proto_lang_toolchain_rejects_files_do_not_use_or_we_will_break_you_without_mercy"),
```
Yannic: @lberki @cushon @hlopko I'd love to get your lgtm on this "feature flag" so we don't have to synchronize flipping a future `--incompatible_blacklisted_protos_requires_proto_info` (or whatever that flag will be called) with protobuf releases. See https://github.com/bazelbuild/bazel/pull/10493 and https://github.com/bazelbuild/bazel/issues/10484 for context.

Yannic:
@rafi-kamal Sorry, I did move some parts into variables after testing it which left trailing `,` behind :/. Please re-trigger CI, thanks.

rafi-kamal:
FYI the Linux Ruby Release and MacOS Ruby 2.3 tests are broken for a while, so the failures are likely not related.

Yannic:
@rafi-kamal Thanks! The failures are indeed unrelated, we're only touching Bazel `BUILD` files here, which aren't used by Ruby.

I think we can merge (& cherry-pick onto `3.11.x`) this now.

BUILD
lberki: Do I understand correctly that the sole reason for this change is http://github.com/bazelbuild/bazel/issues/10484 and thus if that is fixed, this change and the flag becomes unnecessary?

PR #7066, opened 2020-01-08
Fix for Python 4: replace unsafe six.PY3 with PY2

hugovk:
We don't yet know if 3.10 or 4.0 will follow Python 3.9, but whichever it is, it will probably happen in [2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule) and work begins on Python 3.9+1.

There's some code which uses `six.PY3`:

```python
if six.PY3:
    print("Python 3+ code")
else:
    print "Python 2 code"
```

Where:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this will run the Python 2 code!

Instead, use `six.PY2`.

Similarly, use `sys.version_info[0] >= 3` instead of `sys.version_info[0] == 3`.

Found using https://github.com/asottile/flake8-2020.


hugovk:
Closing old PR. Whilst these changes are still valid, we now know Python 3.10 will follow 3.9, and hopefully Protocol Buffers will drop Python 2 (EOL for a year now) before whenever Python 4 is out? :)

PR #7068, opened 2020-01-09
Fix surrogate pattern DecodeError reason

joseotoro:


acozzette:
@joseotoro Sorry for taking so long to get to this. Thank you for the fix, but it looks like it is too late to merge this since this error message was already removed in another change.

Issue #7070, opened 2020-01-09
Ruby: Gem version 3.11.2 doesn't load on Ruby 2.7


threeplanetssoftware:
**What version of protobuf and what language are you using?**
Version: 3.11.2 (released ~December 13, 2019)
Language: Ruby 2.7 (released December 25, 2019)

Windows 8, Linux Ubuntu and Mint

**What runtime / compiler are you using (e.g., python version or gcc version)** 
Ruby 2.7

**What did you do?**
1. gem install google-protobuf
2. Run any Ruby code which has `require 'google-protobuf'` at the start
3. Get Load Errors, see below.

This can also be reproduced by using Travis-CI's 2.7 environment on any of their images.

**What did you expect to see** 
On Ruby 2.3 through 2.6 this results in the expected program execution.

**What did you see instead?** 
On Ruby 2.7, on each platform I've tested on, this results in the following two errors:

> [Snip]
> 1: from /home/travis/.rvm/gems/ruby-2.7.0/gems/google-protobuf-3.11.2-x86_64-linux/lib/google/protobuf.rb:49:in '<top (required)>'
>
> /home/travis/.rvm/gems/ruby-2.7.0/gems/google-protobuf-3.11.2-x86_64-linux/lib/google/protobuf.rb:49:in `require`: cannot load such file -- google/2.7/protobuf_c (LoadError)
>
> [Snip]
>
>  2: from /home/travis/.rvm/gems/ruby-2.7.0/gems/google-protobuf-3.11.2-x86_64-linux/lib/google/protobuf.rb:48:in `<top (required)>`
>
> 1: from /home/travis/.rvm/gems/ruby-2.7.0/gems/google-protobuf-3.11.2-x86_64-linux/lib/google/protobuf.rb:51:in `rescue in <top (required)>`
>
> /home/travis/.rvm/gems/ruby-2.7.0/gems/google-protobuf-3.11.2-x86_64-linux/lib/google/protobuf.rb:51:in `require`: cannot load such file -- google/protobuf_c

**Anything else we should know about your project / environment** 
This appears to be related to Ruby 2.7 releasing a few days after 3.11.2 released as it occurs on each system I try with 2.7 on it (including Travis-ci), but any older Ruby version works fine with 3.11.2. As 27 is a major release for Ruby and new installs will start to incorporate it, suggest either not letting the gem install on it, or putting out a bug fix to support 2.7 prior to the next release.

ewalk153:
[...]

PR #7027, opened 2019-12-18
Build extensions for Ruby 2.7

qnighy:
I've made two groundworks for building protobuf extensions for Ruby 2.7:

- Reporting ruby issue at https://bugs.ruby-lang.org/issues/16384, which was fixed in https://github.com/ruby/ruby/commit/40026a408df5e3576380f6c1d8bf6c119fa2e32b
- Reporting rake-compiler issue at https://github.com/rake-compiler/rake-compiler/pull/161, which was fixed in https://github.com/rake-compiler/rake-compiler/commit/6822686f0c45e752974de02a980365515e843267

Now that 2.7.0-rc2 has been released, can we start building extensions for Ruby 2.7?

PR #7072, opened 2020-01-10
Updated Rakefile to build for Ruby 2.7.0

threeplanetssoftware:
Per issue #7070, google-protobuf 3.11.2 won't load on Ruby 2.7 (just released). The Rakefile wasn't building for Ruby 2.7.0 in lines 73 or 79, added that as a build target to fix the problem moving forward. Only file changed is protobuf/ruby/Rakefile.

After making this change, and building/installing the gem locally, programs which relied on google-protobuf now run.

### DIFF
--- a/ruby/Rakefile
+++ b/ruby/Rakefile
@@ -70,13 +70,13 @@ else
 
   task 'gem:windows' do
     require 'rake_compiler_dock'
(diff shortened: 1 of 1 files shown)


threeplanetssoftware:
I see this is already covered in #7027, closing.

PR #7078, opened 2020-01-10
Optimize Metadata Initialization for PHP [WIP]

TeBoring:


PierrickVoulet:
@TeBoring This PR seems specific to the C implementation, are you planning to do the same in the PHP implementation?

TeBoring:
Yeah. I’ll add php part later. C part should be testable

PierrickVoulet:
@TeBoring it runs fine on my side with some protos generated with version 3.11.2.

This is the only required runtime edition right? If this is the case, what bout creating a new dedicated branch/PR with all the protoc changes for aggregation purpose?

php/ext/google/protobuf/protobuf.c
```diff
@@ -248,6 +250,23 @@ EnumDescriptorInternal* get_class_enumdesc(const char* klass) {
   }
 }
 
+void add_file(const char* file) {
+  printf("File registered: %s\n", file);
```
PierrickVoulet: @TeBoring printing each registered file could take a significant amount of time, is this really necessary?

haberman:
Closing as obsolete.

PR #7085, opened 2020-01-13
Fix the Ruby extensions build

casperisfine:
Based on https://github.com/protocolbuffers/protobuf/pull/7027/

I finally managed to reproduce locally with:

```
docker run -v $PWD:/tmpfs/src/github/protobuf/ruby -e UID\=1000 -e GID\=1000 -e USER\=kbuilder -e GROUP\=kbuilder -e GEM_PRIVATE_KEY_PASSPHRASE -e ftp_proxy -e http_proxy -e https_proxy -e RCD_HOST_RUBY_PLATFORM\=x86_64-linux -e RCD_HOST_RUBY_VERSION\=2.4.1 -e RCD_IMAGE\=protobuftesting/ruby_rake_compiler_52bbc96c6d72ba1b25e99e19bdcfeb240438566c -w /tmpfs/src/github/protobuf/ruby --rm -i protobuftesting/ruby_rake_compiler_52bbc96c6d72ba1b25e99e19bdcfeb240438566c runas sigfw bash -c 'bundle && IN_DOCKER=true rake cross native gem RUBY_CC_VERSION\=2.7.0:2.6.0:2.5.0:2.4.0:2.3.0]'
```

Installing `bundler` just before `bundle install` like this is not super elegant, but given that I don't have a proper  feedback loop, it seems like the most straightforward fix.

@qnighy @haberman

### DIFF
--- a/kokoro/linux/dockerfile/test/ruby/Dockerfile
+++ b/kokoro/linux/dockerfile/test/ruby/Dockerfile
@@ -32,6 +32,7 @@ RUN /bin/bash -l -c "rvm install 2.3.8"
 RUN /bin/bash -l -c "rvm install 2.4.5"
 RUN /bin/bash -l -c "rvm install 2.5.1"
 RUN /bin/bash -l -c "rvm install 2.6.0"
+RUN /bin/bash -l -c "rvm install 2.7.0"
 
 RUN /bin/bash -l -c "echo 'gem: --no-ri --no-rdoc' > ~/.gemrc"
 RUN /bin/bash -l -c "echo 'export PATH=/usr/local/rvm/bin:$PATH' >> ~/.bashrc"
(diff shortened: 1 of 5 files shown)


casperisfine:
@googlebot I consent.

acozzette:
@casperisfine I think the CLA bot is having trouble validating that all CLAs are signed because of the commits by @byroot and @qnighy.  If they leave comments saying `@googlebot I consent` then that should solve the problem.

byroot:
@googlebot I consent.

qnighy:
@googlebot I consent.

haberman:
Kokoro build appears broken still?

acozzette:
I'm going to close this because I assume it's no longer needed since #7027 was merged.

PR #7089, opened 2020-01-13
Correct @return in Any.unpack docblock

andres-valdes:
As per https://docs.phpdoc.org/latest/references/phpdoc/tags/return.html, the first token after the @return tag should be the return type.

### DIFF
--- a/php/src/Google/Protobuf/Any.php
+++ b/php/src/Google/Protobuf/Any.php
@@ -251,7 +251,7 @@ class Any extends \Google\Protobuf\Internal\Message
      * targeted message type. If failed, an error will be thrown. Otherwise,
      * the method will create a message of the targeted type and fill it with
      * the decoded value in Any.
-     * @return unpacked message
+     * @return Message unpacked message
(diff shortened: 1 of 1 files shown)


carloslahrssen:
Gonna bump this as this is also breaking our ci/cd. Would be a huge win for us if we could get this merged!

david-castaneda:
I second this👍🏻

dranes:
I need this

php/src/Google/Protobuf/Any.php
```diff
@@ -251,7 +251,7 @@ public function setValue($var)
      * targeted message type. If failed, an error will be thrown. Otherwise,
      * the method will create a message of the targeted type and fill it with
      * the decoded value in Any.
-     * @return unpacked message
+     * @return Message unpacked message
```
TeBoring: Should this be \Google\Protobuf\Internal\Message

andres-valdes: The class already has a `use \Google\Protobuf\Internal\Message`, so the namespace is unnecessary.

PR #7106, opened 2020-01-20
Change Objective C class references to using macros.

dmaclach:
Allows easy replacing of class names in generated code using #define to redefine class names at compile time.

### DIFF
--- a/objectivec/GPBUtilities_PackagePrivate.h
+++ b/objectivec/GPBUtilities_PackagePrivate.h
@@ -35,13 +35,22 @@
 #import "GPBDescriptor_PackagePrivate.h"
 
 // Macros for stringifying library symbols. These are used in the generated
-// PB descriptor classes wherever a library symbol name is represented as a
-// string. See README.google for more information.
+// GPB descriptor classes wherever a library symbol name is represented as a
+// string.
 #define GPBStringify(S) #S
 #define GPBStringifySymbol(S) GPBStringify(S)
 
 #define GPBNSStringify(S) @#S
 #define GPBNSStringifySymbol(S) GPBNSStringify(S)
+
+// Macros for generating a Class from a class name. These are used in
+// the generated GPB descriptor classes wherever an Objective C class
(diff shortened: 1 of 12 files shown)


objectivec/GPBUtilities_PackagePrivate.h
```diff
@@ -36,13 +36,22 @@
...
-// string. See README.google for more information.
+// string.
 #define GPBStringify(S) #S
 #define GPBStringifySymbol(S) GPBStringify(S)
 
 #define GPBNSStringify(S) @#S
 #define GPBNSStringifySymbol(S) GPBNSStringify(S)
 
+// Macros for generating a Class from a class name. These are used in
+// the generated PB descriptor classes wherever an Objective C class
```
thomasvl: PB -> GPB ?

src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -432,12 +432,9 @@ void FileGenerator::GenerateSource(io::Printer *printer) {
       "\n");
   if (!fwd_decls.empty()) {
     printer->Print(
-        "#pragma mark - Objective C Class references\n"
-        "// This somewhat arcane code forces linkage of classes from static archives by\n"
-        "// adding a concrete reference to the classes.\n"
-        "// We don't use `[Foo class]` because we need a static value for our initializer.\n"
```
thomasvl: These last three lines might still be useful to explain to someone, up to you if you want to keep them.

objectivec/GPBUtilities_PackagePrivate.h
dmaclach: Done (and in the comment above that I probably copied from)

src/google/protobuf/compiler/objectivec/objectivec_file.cc
dmaclach: Added the important clarifying line IMHO.

PR #7111, opened 2020-01-20
Add folder references to the objectivec compiler

dmaclach:
I find it significantly easier to work in Xcode when I have access to the compiler code as well. This just adds folder references to the compiler/objectivec code to all of the projects so they can be searched and edited.


thomasvl:
I'm not sure the common case it to work on the generator vs. say someone using this project to integrate into their own xcode project.

PR #7119, opened 2020-01-22
Enable Runtime Import of ".proto" Files

gnossen:
This PR enables the user to load modules equivalent to a "_pb2.py" file using code like this:

```python
from google import protobuf
protos = protobuf.protos("foo.proto")
```
The `protos` object is an instance of type `types.ModuleType` whose public members are the same as those present in the equivalent `foo_pb2.py` file.

As requested by the Protobuf Team, the machinery behind this function *does not* generate textual Python code. Instead, we create a `FileDescriptor` object and use it to reflectively instantiate message classes, enum classes, etc.

There are three public API changes in this PR:

 - `google.protobuf.protos`: The function outlined above.
 - `google.protobuf.FileDescriptor.FromFile`: This static method enables the instantiation of a python-level `FileDescriptor` object from a ".proto" file.
 - `google.protobuf.FileDescriptor.CopySourceCodeInfoToProto`: This static method enables the extraction of source code info into a `google.protobuf.descriptor_pb2.FileDescriptorProto` object, mirroring [the C++ method](https://github.com/protocolbuffers/protobuf/blob/a104dffcb6b1958a424f5fa6f9e6bdc0ab9b6f9e/src/google/protobuf/descriptor.h#L1427). This is not used in this PR, but I intend to use it to extract docstrings for use in service help text in gRPC.

If the reviewers do not like the addition of either of the latter two methods as public APIs, I'm more than happy to make them private.

This is my first PR to the repo, so it's entirely possible that I've missed some aspect of the development process, such as special formatting/linting. Please let me know if any of that sort of thing is off.


python/google/protobuf/importer.py
```diff
@@ -0,0 +1,237 @@
...
+                                                       message_descriptor,
+                                                       module.__name__)
+          _sym_db.RegisterMessageDescriptor(message_descriptor)
+          _sym_db.RegisterMessage(message_type)
+          setattr(module, name, message_type)
+
+    class ProtoFinder(importlib.abc.MetaPathFinder):
+      """MetaPathFinder that locates .proto files."""
+
+      def find_spec(self, fullname, path, target=None):
```
michaelsafyan: What happens if there is both a *.proto file and a *_pb2.py file in the same directory?

python/google/protobuf/importer.py
```diff
@@ -0,0 +1,237 @@
...
+      Returns:
+        A module object identical to one that would be generated by protoc
+        and loaded with an `import foo_pb2` statement.
+      """
+      with _augmented_syspath(include_paths):
+        module_name = _proto_file_to_module_name(protobuf_path)
+        module = importlib.import_module(module_name)
+        return module
+
+    sys.meta_path.extend([ProtoFinder()])
```
michaelsafyan: During development, you may actually want the ProtoFinder to go ahead of normal *.py resolution (so that a *.proto file with a newer mtime takes precedence over the locally commited generated *.py file).

python/google/protobuf/importer.py
gnossen: As you pointed out in the comment below, the "_pb2.py" file will take precedence. Let's close this discussion and continue it in that thread.

python/google/protobuf/importer.py
gnossen: Conversely, in production, you want "_pb2.py" files to take precedence over ".proto" files. Perhaps a feature flag/environment variable that allows you two choose between the two behaviors? I'd prefer an argument to the function, but Finders unfortunately have to be per-process globals.

python/google/protobuf/pyext/descriptor_pool.cc
```diff
@@ -121,24 +215,43 @@ static PyDescriptorPool* _CreateDescriptorPool() {
...
   PyDescriptorPool* cpool = _CreateDescriptorPool();
   if (cpool == NULL) {
     return NULL;
   }
-  cpool->pool = new DescriptorPool(underlay);
-  cpool->underlay = underlay;
+  cpool->underlay = DescriptorPool::generated_pool();
+  cpool->file_error_collector = new PyErrorCollector;
+  cpool->disk_source_tree = new DiskSourceTree;
+  cpool->disk_database = new SourceTreeDescriptorDatabase(cpool->disk_source_tree);
```
anandolee: Always create disk database to the generated pool will change current  pool behavior I think? For a project that generated some .proto file to _pb2 and not generated some other .proto file, descriptor pool can only find those are generated before this change but can find all  the files after this PR

anandolee:
Use SourceTreeDescriptorDatabase LGTM overall. There is still some APIs and implementation details we need to take care of.

David Finkelstein. Any thoughts on high level of this PR?

python/google/protobuf/pyext/descriptor_pool.cc
gnossen: As far as I can tell, this does not actually change the behavior of the current pool. `InProcessDescriptorDatabase` gives precedence to descriptors added from generated files, so the `SourceTreeDescriptorDatabase` will only ever be used for descriptors that cannot be found in a generated `_pb2.py` file.

Importing a module is an opt-in process. So only applications that are currently expecting an import to fail should be surprised by this change.

This PR is currently passing the suite of unit tests. If any of this PR constitutes a breaking change, shouldn't we add tests to capture said behavior?

anandolee: "applications that are currently expecting an import to fail should be surprised by this change" This is the behavior change I was talking about. The issues for the behavior change:
1, Pure python and cpp extension will have different result when find file in pool
2, For users that did not use the protobuf.protos(), they will able to find the descriptors from pool but not able to use the modules. Which is confusing.

gnossen: 1. Can you expand on your first point? Is the expectation that the pure Python implementation will behave exactly like the C extension?
2. What would you suggest instead? Initialize the `InProcessDescriptorPool` so that it doesn't use a fallback `SourceTreeDescriptorDatabase` and only enable it the first time the user calls `protobuf.protos()`?

anandolee: 1, Not exactly same. But we are always trying to fix the behavior differences (there are some differences unable to be fixed though). We do not want to introduce more differences.

2, How about not use SourceTreeDescriptorDatabase as fallback for the default pool. Instead: Create a static SourceTreeDescriptorDatabase to parse files when protobuf.protos() is called. And then add the parsed FileDescriptor to the default pool.

gnossen: I explored that option, but I couldn't get it work robustly. In order to build `FileDescriptor` objects, you need a `DescriptorPool` for them to live in. So there are two options for which pool to use:
 1. Use a separate pool from the default Python-level pool.
 2. Directly use the default Python-level pool. (essentially the current state of this PR)

If you use a separate pool, there are a few problems. Suppose `B.proto` depends on `A.proto`. The application first does `import A_pb2`. Then it does `B = protobuf.protos("B.proto")`. `B.proto` needs to be linked with the `A.proto` `FileDescriptor`, which lives in the default Python `DescriptorPool`. You could solve this by using a fallback db pointing at the default Python `DescriptorPool`, but then the opposite dependency direction becomes a problem. That is, the following code would no longer work:
```python
A = protobuf.protos("A.proto")
import B_pb2
```

Or, at the very least, it would create a second copy of `A.proto`'s module loaded from an `A_pb2.py` which could lead to pointer equality problems.

By putting both statically generated and runtime-parsed protos into the same pool, both scenarios above work out, which is explained in detail in the docstring for `protobuf.protos()`.

If the concern is that we don't want to change the behavior for users who don't opt in by using the new function, shouldn't it be sufficient to only start using the `SourceTreeDescriptorDatabase` once `protobuf.protos()` has been called for the first time?

anandolee: Would add an internal_set_fallback_database() API to descriptor pool work? Thus we can turn on SourceTreeDescriptorDatabase only in protobuf.protos(), and turn it off before exit protobuf.protos()

gnossen: @anandolee I think that would work if `internal_set_fallback_database` is a method on the new `InProcessDescriptorDatabase`. The whole reason that it exists (instead of just using the method you're suggesting) is because [you can't call `BuildFile` on a `DescriptorPool` with a fallback database set](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.cc#L3559). Other than that, the finer grained control over the usage of the `SourceTreeDescriptorDatabase` sounds like a good idea to me.

gnossen: @anandolee Friendly ping.

anandolee: " you can't call BuildFile on a DescriptorPool with a fallback database set. "  no need call BuildFile on descriptor pool, only FileFileByName() will be called and then invoke DescriptorPool::TryFindFileInFallbackDatabase().   But add internal_set_fallback_database() in InProcessDescriptorDatabase sounds better to me, because we do not need to change protobuf c++ APIs, the change will be in python c extension only. I'd still like to suggest wait for David's review

python/google/protobuf/_importer_impl.py
```diff
@@ -0,0 +1,216 @@
...
+  def exec_module(self, module):
+    """Instantiate a module identical to the generated version."""
+    file_descriptor = _pyext_message.FileDescriptor.FromFile(
+        self._protobuf_path.encode('ascii'),
+        [path.encode('ascii') for path in sys.path])
+    _sym_db.RegisterFileDescriptor(file_descriptor)
+    for dependency in file_descriptor.dependencies:
+      module_name = _proto_file_to_module_name(dependency.name)
+      if module_name not in sys.modules:
+        importlib.import_module(module_name)
```
dlj-NaN: The runtime generation must not affect system libraries. Consider two transitive dependencies, which actually need a different definition for the same (conflicting) name: which one wins?

We cannot say, and we should not try to say. Touching the import machinery is simply too subtle for us to hide behind a function call that can look at arbitrary files.

python/google/protobuf/pyext/descriptor_pool.cc
dlj-NaN: While I appreciate that the DescriptorPool behavior demonstrates the issue, I think this is the same problem with the importer.

This facility already shouldn't be touching the import machinery, which also means it shouldn't be touching the global DescriptorPool.

python/google/protobuf/_importer_impl.py
gnossen: @dlj-NaN Can you expand on the sorts of system libraries you're concerned about? This should only affect "*_pb2" modules. In google3, this module loader will only ever load a module when the user explicitly adds a ".proto" file as a `data` dependency on their `py_binary`. In open source, non-hermetic builds make a conflict more likely, but perhaps also less serious.

> Consider two transitive dependencies, which actually need a different definition for the same (conflicting) name: which one wins?

Can you give an example of such a situation? While I can see this happening for an arbitrary module loader, I'm having a hard time coming up with a situation where this particular module loader would exhibit this behavior. You've have to have a `foo.proto` as well as a `foo_pb2` on your `sys.path`. Then you would need a situation where "(two transitive dependencies)  actually need a different definition for the same (conflicting) name"? I suppose this could happen if you lay down an alternate definition for a ".proto" used in a low-level library? This seems unlikely to happen outside of Google3.

> Touching the import machinery is simply too subtle for us to hide behind a function call that can look at arbitrary files.

Are you against modification of the import machinery *in principle* or just against wrapping it in a function call? Addition of the module loader to `sys.meta_path` could instead be an opt-in process enabled by something like a `protobuf.enable_disk_proto_files()` (bikesheddable name). Note that in the other discussion thread on this PR, @anandolee and I discussed limiting the time that the module loader spends on `sys.meta_path` to the duration of the call to `protobuf.protos()`. How would that affect the situations you're concerned about?

python/google/protobuf/pyext/descriptor_pool.cc
gnossen: @anandolee 

> no need call BuildFile on descriptor pool

This PR may not be calling `BuildFile`, but [the default Python-level descriptor pool uses that method to register the descriptors from the generated files](https://github.com/protocolbuffers/protobuf/blob/4cf5bfee9546101d98754d23ff378ff718ba8438/python/google/protobuf/pyext/descriptor_pool.cc#L608). This is the operation that would fail if we set a fallback db and an underlay pool on the default Python pool.

@dlj-NaN

> This facility already shouldn't be touching the import machinery, which also means it shouldn't be touching the global DescriptorPool.

I can see the argument to be made here. We need the set of ".proto" files loadable as modules to mirror one-to-one the set of FileDescriptors in the global DescriptorPool. Otherwise, dependency issues could arise. So if we do not allow runtime-parsed protos to share module space with statically generated "_pb2.py" files (as this and your other comment suggest), then we also cannot allow them to share the default Python descriptor pool.

That being the case, given that underlay pools and fallback DBs are the only mechanisms available to cross-link descriptors from different pools, there is no way to make this code work:

```python
A = protobuf.protos("A.proto")
import B_pb2
```

where, as in my earlier comment, "B.proto" depends on A.proto.

However, you could still make this code work:

```python
import A_pb2
B = protobuf.protos("B.proto")
```

by setting up a separate descriptor pool for runtime-parsed files with the default Python pool as an underlay. If we're disallowing modification of the import machinery, then we could perhaps instantiate `B` above using [`types.ModuleType("B_pb2")`](https://docs.python.org/3/library/types.html#types.ModuleType) (or, equivalently, [`importlib.util.module_from_spec()`](https://docs.python.org/3/library/importlib.html#importlib.util.module_from_spec)). This appears not to introduce clashes:

```python
>>> B = types.ModuleType("B_pb2")
>>> import B_pb2
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'B_pb2'
```

Thoughts?

python/google/protobuf/_importer_impl.py
gnossen: @dlj-NaN Friendly ping.

acozzette:
I'm going to close this PR since there has been no activity in a long time. @gnossen If you're still interested in pushing for this change then it may be best to set up a meeting to see if we can resolve the reviewers' concerns.

PR #7121, opened 2020-01-22
Make the unittest proto file generation handle additions better.

thomasvl:
Didn't really handle when a file was missing, but all the other were up to date,
this handles that case also.

### DIFF
--- a/objectivec/DevTools/compile_testing_protos.sh
+++ b/objectivec/DevTools/compile_testing_protos.sh
@@ -26,71 +26,9 @@ case "${ACTION}" in
 esac
 
 # -----------------------------------------------------------------------------
-# Ensure the output dir exists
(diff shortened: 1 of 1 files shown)


objectivec/DevTools/compile_testing_protos.sh
```diff
@@ -25,6 +25,64 @@ case "${ACTION}" in
     ;;
 esac
 
+# -----------------------------------------------------------------------------
+# Most of the proto files that exist in the C++ src tree.  Several are used in
+# the tests, but the extra don't hurt in that they ensure ObjC sources can be
```
dmaclach: Could you clarify the second sentence in the comment? For some reason I'm having troubles parsing what you mean.

thomasvl: tweaked.

PR #7124, opened 2020-01-23
Block subclassing

dmaclach:
Add GPB_FINAL macro to block subclassing of generated ObjC classes.

This will prevent subclassing of any of the protobuf classes generated by the compiler.
If you need to be able to do this, please let us know. You can temporarily "reenable" it by
redefining the GPB_FINAL macro.


dmaclach:
Coment added.

PR #7130, opened 2020-01-23
Add null checks to GetOption

### LINKED ISSUE #7127: C#: MethodDescriptor.GetOption throws NullReferenceException
`MethodDescriptor.GetOption<T>(Extension<MethodOptions, T> extension)` throws `NullReferenceException` when the method descriptor does not have the option on it. I would expect it to return null.

My code:

```csharp
private bool TryGetHttpRule(MethodDescriptor methodDescriptor, out HttpRule httpRule)
{
    var extension = AnnotationsExtensions.Http;
    httpRule = methodDescriptor.GetOption(extension);

    return httpRule != null;
}
```

Google.Protobuf 3.11.2
Google.Api.CommonProtos 2.0.0-beta01

ObsidianMinor:
Fixes #7127 

Null continues to erode away at my soul.

Should we add a TryGetOption API in another PR to better match up with the CustomOptions API @jtattermusch?

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
@@ -184,21 +184,21 @@ namespace Google.Protobuf.Test.Reflection
             var fileDescriptor = UnittestProto3Reflection.Descriptor;
             var messageDescriptor = TestAllTypes.Descriptor;
             Assert.NotNull(fileDescriptor.CustomOptions);
-            Assert.DoesNotThrow(() => { fileDescriptor.GetOption(Fileopt); });
+            Assert.AreEqual(fileDescriptor.GetOption(Fileopt), null);
             Assert.NotNull(messageDescriptor.CustomOptions);
-            Assert.DoesNotThrow(() => { messageDescriptor.GetOption(Msgopt); });
+            Assert.AreEqual(messageDescriptor.GetOption(Msgopt), null);
             Assert.NotNull(messageDescriptor.Fields[1].CustomOptions);
-            Assert.DoesNotThrow(() => { messageDescriptor.Fields[1].GetOption(Fieldopt); });
+            Assert.AreEqual(messageDescriptor.Fields[1].GetOption(Fieldopt), null);
             Assert.NotNull(fileDescriptor.Services[0].CustomOptions);
-            Assert.DoesNotThrow(() => { fileDescriptor.Services[0].GetOption(Serviceopt); });
+            Assert.AreEqual(fileDescriptor.Services[0].GetOption(Serviceopt), null);
             Assert.NotNull(fileDescriptor.Services[0].Methods[0].CustomOptions);
-            Assert.DoesNotThrow(() => { fileDescriptor.Services[0].Methods[0].GetOption(Methodopt); });
+            Assert.AreEqual(fileDescriptor.Services[0].Methods[0].GetOption(Methodopt), null);
             Assert.NotNull(fileDescriptor.EnumTypes[0].CustomOptions);
-            Assert.DoesNotThrow(() => { fileDescriptor.EnumTypes[0].GetOption(Enumopt); });
+            Assert.AreEqual(fileDescriptor.EnumTypes[0].GetOption(Enumopt), null);
             Assert.NotNull(fileDescriptor.EnumTypes[0].Values[0].CustomOptions);
-            Assert.DoesNotThrow(() => { fileDescriptor.EnumTypes[0].Values[0].GetOption(EnumValueOpt1); });
+            Assert.AreEqual(fileDescriptor.EnumTypes[0].Values[0].GetOption(EnumValueOpt1), 0);
             Assert.NotNull(TestAllTypes.Descriptor.Oneofs[0].CustomOptions);
-            Assert.DoesNotThrow(() => { TestAllTypes.Descriptor.Oneofs[0].GetOption(OneofOpt1); });
+            Assert.AreEqual(TestAllTypes.Descriptor.Oneofs[0].GetOption(OneofOpt1), 0);
         }
 
         [Test]
--- a/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
@@ -184,13 +184,21 @@ namespace Google.Protobuf.Test.Reflection
             var fileDescriptor = UnittestProto3Reflection.Descriptor;
             var messageDescriptor = TestAllTypes.Descriptor;
             Assert.NotNull(fileDescriptor.CustomOptions);
+            Assert.DoesNotThrow(() => { fileDescriptor.GetOption(Fileopt); });
             Assert.NotNull(messageDescriptor.CustomOptions);
+            Assert.DoesNotThrow(() => { messageDescriptor.GetOption(Msgopt); });
             Assert.NotNull(messageDescriptor.Fields[1].CustomOptions);
+            Assert.DoesNotThrow(() => { messageDescriptor.Fields[1].GetOption(Fieldopt); });
             Assert.NotNull(fileDescriptor.Services[0].CustomOptions);
+            Assert.DoesNotThrow(() => { fileDescriptor.Services[0].GetOption(Serviceopt); });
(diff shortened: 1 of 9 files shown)


csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
```diff
@@ -184,13 +184,21 @@ public void NoOptions()
             var fileDescriptor = UnittestProto3Reflection.Descriptor;
             var messageDescriptor = TestAllTypes.Descriptor;
             Assert.NotNull(fileDescriptor.CustomOptions);
+            Assert.DoesNotThrow(() => { fileDescriptor.GetOption(Fileopt); });
```
JamesNK: What value does `Assert.DoesNotThrow` add? A thrown error would fail the test anyway.

Wouldn't a better check be that these methods return null if there is no option, and not null if there is.

ObsidianMinor: `Assert.DoesNotThrow` is a style that's been enforced in previous PRs. Though, yes, a better check would be to make sure GetOption returns the default value of the extension.

Falco20019:
I also just ran into that problem and would like the idea of having `TryGetOption`. I'm iterating over some fields and have to check if any has the option set. This is not possible with only using `default(T)`.

thesoftwarejedi:
@Falco20019 I suggest we merge this, but also add the `Try` versions of all methods in another PR.  I can try to get to coding that this weekend using https://github.com/protocolbuffers/protobuf/issues/7240 as the basis for it.

I'm not familiar with the release pipeline here or how to push this along.  I think @ObsidianMinor is king.

ObsidianMinor:
@thesoftwarejedi I'm working on a TryGet PR right now. Also I don't work at Google, I'm just a friendly OSS contributor. Ask @jtattermusch.

thesoftwarejedi:
> @thesoftwarejedi I'm working on a TryGet PR right now. 

Perfect, thanks!  Saved my weekend ;)

> Also I don't work at Google, I'm just a friendly OSS contributor.

I figured as much.  Thanks for your contributions!

ozevren:
Is there any way I can help with getting this checked in? This and the `Try` overload our blocking some of my, so I am highly motivated to help out here. :)

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -136,7 +136,13 @@ public EnumValueDescriptor FindValueByName(string name)
         /// </summary>
         public T GetOption<T>(Extension<EnumOptions, T> extension)
         {
-            var value = Proto.Options.GetExtension(extension);
+            var options = Proto.Options;
+            if (options == null)
+            {
+                return default(T);
```
jtattermusch, line 142: I don't think returning `default(T)` is a good idea. It seems that for e.g. extension of type int, you won't be able to differentiate between the option not being present at all (which should return null) and the option having value of 0.

Given that the GetOption API is relatively new (and looks like it has been kind of broken for the very beginning), should we just introduce TryGetOption methods instead, to keep the semantics clean?

the GetOption can either be removed (that should be fine because it was only introduced as part of the proto2 work and we've marked proto2 support as "experimental")  or it can just throw an exception if the value is not present (which renders it not very useful, so we might wanna mark it as obsolete?).

csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
```diff
@@ -184,13 +184,21 @@ public void NoOptions()
             var fileDescriptor = UnittestProto3Reflection.Descriptor;
             var messageDescriptor = TestAllTypes.Descriptor;
             Assert.NotNull(fileDescriptor.CustomOptions);
+            Assert.AreEqual(fileDescriptor.GetOption(Fileopt), null);
```
jtattermusch, line 187: why are we intermixing the GetOption tests with a CustomOptions-related test (and all that in CustomOptionsTest.cs).
I think it would be better if GetOption methods was tested in a separate test case.

ObsidianMinor: I thought that even though the GetOption APIs weren't in a "CustomOptions" type they still worked with "custom options" so it'd be easier to test everything relating to custom options together.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
ObsidianMinor: Returning `default(T)` is a mistake. It should be the default value for the extension.

Falco20019:
Wouldn’t it be nicer if `Proto.Options` would not be null but always defined empty and that `GetExtension` would then just return the `DefaultValue` of the extension requested? This would avoid a lot of code duplication.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
jtattermusch: Still, even if you return `extension.DefaultValue`, it is unclear to me how I would distinguish between the extension not being present at all  and it being present and having the default value.

ObsidianMinor: You can't, which is why #7282 was made. Should we also add a `HasOption` API to go with it?

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
```diff
@@ -555,7 +555,13 @@ public override string ToString()
         /// </summary>
         public T GetOption<T>(Extension<FileOptions, T> extension)
         {
-            var value = Proto.Options.GetExtension(extension);
+            var options = Proto.Options;
+            if (options == null)
```
jtattermusch, line 559: actually, looking at the whole options API in protobuf C# from a more high level perspective, we only need to check whether `options == null` because e.g. the FileDescriptor doesn't provide access to the file options (the property FileDescriptor.Proto is internal, so one cannot access to FileDescriptor.Proto.Options).  

This actually has several downsides:
- we need extra logic such as a GetOption methods, instead of just exposing e.g. the FileOptions msg (or other options) on a descriptor and letting the user use GetExtension methods the same way they would do it for any other message with extensions (which is the way reading custom options is done for e.g. C++, Java and python, as documented here: https://developers.google.com/protocol-buffers/docs/proto#customoptions)
- because the "options" messages are not exposed, one cannot access the standard (=the non-extension ones) options at all.

This makes me think that the whole GetOptions() and TryGetOption() API is a bad idea and we should rather focus on how to expose the descriptor.Options property safely, because then we wouldn't need to worry about the GetOptions/TryGetOptions etc. API's at all (they could be removed and everyone could just rely on standard extension APIs).

ObsidianMinor: The only way we could expose descriptor.Options safely is by deep cloning it (since all the protos are supposed to be immutable). With these APIs the only thing we need to clone is the value, if we need to clone it at all.

jtattermusch: Maybe the deep clone isn't really all that problematic. Most projects would only have a few options in place and it seems fair to require users to cache the results they got from e.g. messageDescriptor.Options if they wanted to process a large number of options.
It seems to me that just exposing description.Options (as a defensive copy) would lead to a more logical and straightforward API (with smaller footprint and fewer duplications, and also more similar to what other languages do) - WDYT?

ObsidianMinor: Could we make a internal wrapper class that exposes the IExtendableMessage<T> API but only implements the pure functions? Then GetOptions would just return an IExtendableMessage implementation.

jtattermusch: But you actually want to be able to access e.g. the FileOptions' properties (the "standard" options, such as "java_package" etc.) 

see https://github.com/protocolbuffers/protobuf/blob/4059c61f27eb1b06c4ee979546a238be792df0a4/csharp/src/Google.Protobuf/Reflection/Descriptor.cs#L3955

jtattermusch:
Superseded by https://github.com/protocolbuffers/protobuf/pull/7434 (GetOption API is now obsolete).

Issue #5995, opened 2019-04-06
Provide an option to remove __FILE__ in generated C++ header


xhwang-chromium:
**What version of protobuf and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?**
All

**What runtime / compiler are you using (e.g., python version or gcc version)**
gcc

**What did you do?**
Check strings exposed by release build of Chrome on Linux.
$ strings /opt/google/chrome/chrome | grep pb.cc

**What did you expect to see**
No paths of pb.cc files

**What did you see instead?**
...
gen/components/policy/proto/cloud_policy.pb.cc
gen/components/dom_distiller/core/proto/adaboost.pb.cc
gen/components/download/internal/background_service/proto/request.pb.cc
gen/third_party/metrics_proto/memory_leak_report.pb.cc
gen/third_party/metrics_proto/ukm/report.pb.cc
...

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).
This is caused by the __FILE__ in GOOGLE_PROTOBUF_VERIFY_VERSION in src/google/protobuf/stubs/common.h, which is in every generated pb.cc file. As a result, it increased Chrome binary size by 12kB without obvious user benefit.

It would be great not to include __FILE__ in release builds, or at least provide an option to disable it.

nico:
(The background is that we control which version of protoc and libprotobuf is used in chrome builds, so the version check doesn't do anything for us.)

xhwang-chromium:
Kindly ping! Or shall I just upload a pull request to solve it? Thanks!

acozzette:
Maybe a good solution would be that we only define `GOOGLE_PROTOBUF_VERIFY_VERSION` in `stubs/common.h` if it's not already defined. That way you could define it to do nothing by passing `-DGOOGLE_PROTOBUF_VERIFY_VERSION`. Do you want to send a pull request to make that change?

xhwang-chromium:
The plan sgtm. I've not worked on this code base before, so if you could quickly fix it it'll be great. Otherwise I can give it a try. Please let me know what you think! Thanks!

acozzette:
If you could give it a try that would be great because I have my hands full with other things right now.

xhwang-chromium:
Sure, I'll give it a try. Thanks for letting me know!

acozzette:
[...]

PR #7142, opened 2020-01-28
Make defining GOOGLE_PROTOBUF_VERIFY_VERSION optional.

koscelansky:
This macro will by default use __FILE__ macro, so in C++ compiled
binary filenames are present. This is not always wanted. Also
when classes from proto files are generated and compiled at the
same time as protobuf runtime is compiled, there is no need to
check the compatibility of the versions.

Now you can define GOOGLE_PROTOBUF_VERIFY_VERSION as empty and thus
disable the version checks altogether.

The solution was proposed as part of #5995.

### DIFF
--- a/src/google/protobuf/stubs/common.h
+++ b/src/google/protobuf/stubs/common.h
@@ -113,11 +113,12 @@ std::string PROTOBUF_EXPORT VersionString(int version);
 // to use the protobuf library) to verify that the version you link against
 // matches the headers you compiled against.  If a version mismatch is
 // detected, the process will abort.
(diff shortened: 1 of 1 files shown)


koscelansky:
@googlebot I signed it!

acozzette:
I know I suggested this idea but I actually think it's no longer necessary. We used to generate a call to `GOOGLE_PROTOBUF_VERIFY_VERSION` in every `pb.cc` file but we no longer do that. Now the macro will not get called unless you call it yourself, so I think the current behavior is fine as it is.

PR #7155, opened 2020-01-30
Aggregate Metadata Files

TeBoring:
Instead of calling initOnce of dependencies, initialize metadata of dependencies in the same file.

Needs to pass aggregate_metadata option to protoc to trigger, e.g.:
     ```--php_out=aggregate_metadata=foo#bar:generated_dir```
   For each input file, transitive dependencies (including itself), whose package name has the prefix of foo or bar, will be aggregated, in which their metadata string will be aggregated in the same internalAddGeneratedFile call. For other dependencies, initOnce is called as before.

This feature is EXPERIMENTAL. DO NOT USE!!!


dankurka:
Can we please squash this into one commit?

TeBoring:
When merging, it will be squashed.

PR #7159, opened 2020-01-30
Add new C++ keywords (C++20, TS TM)

dmaclach:
Adds several C++ keywords to the reserved keyword list.


dmaclach:
Just realized I may have to bump the version number again because this is another change that may break ABI. Thoughts?

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -182,17 +182,19 @@ const char* const kReservedWordList[] = {
   "self", "instancetype", "nullable", "nonnull", "nil", "Nil",
   "YES", "NO", "weak",
 
-  // C/C++ keywords (Incl C++ 0x11)
+  // C/C++ keywords (Incl C++ 20 including TM TS)
```
thomasvl: The TM ones may change, right?

dmaclach: they may, but unlikely at this point

thomasvl:
> Just realized I may have to bump the version number again because this is another change that may break ABI. Thoughts?

This sorta change isn't what those constants cover, as there are no changes to the library apis themselves.  The danger here is this is a potentially breaking change because it is a change in how we generate the ObjC interfaces.

thomasvl:
Closing this out for now while we do some validation on what these new additions really mean as they likely would break a lot of people's code.

Issue #7076, opened 2020-01-10
protoc: c++ windows #defines IN, OUT, and OPTIONAL to empty strings


scotthart:
**What version of protobuf and what language are you using?**
Version: 3.8.0
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Windows

**What runtime / compiler are you using (e.g., python version or gcc version)**
msvc

**What did you do?**
including field_behavior.pb.h generated from https://github.com/googleapis/googleapis/blob/c9dce519127b97e866ca133a01157f4ce27dcceb/google/api/field_behavior.proto#L57 in other files results in a malformed enum definition for google::api::FieldBehavior

**What did you expect to see**
field_behavior.pb.h was emitted containing:
```
enum FieldBehavior {
  FIELD_BEHAVIOR_UNSPECIFIED = 0,
  OPTIONAL = 1,
  REQUIRED = 2,
  OUTPUT_ONLY = 3,
  INPUT_ONLY = 4,
  IMMUTABLE = 5,
  FieldBehavior_INT_MIN_SENTINEL_DO_NOT_USE_ = std::numeric_limits<::google::protobuf::int32>::min(),
  FieldBehavior_INT_MAX_SENTINEL_DO_NOT_USE_ = std::numeric_limits<::google::protobuf::int32>::max()
};
```

**What did you see instead?**
When included after <windows.h> the resulting .i file contains:
```
enum FieldBehavior : int {
  FIELD_BEHAVIOR_UNSPECIFIED = 0,
   = 1,
  REQUIRED = 2,
  OUTPUT_ONLY = 3,
  INPUT_ONLY = 4,
  IMMUTABLE = 5,
  FieldBehavior_INT_MIN_SENTINEL_DO_NOT_USE_ = std::numeric_limits<::google::protobuf::int32>::min(),
  FieldBehavior_INT_MAX_SENTINEL_DO_NOT_USE_ = std::numeric_limits<::google::protobuf::int32>::max()
};
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).
Code example: https://godbolt.org/z/W_zBzC
The offending header is minwindef.h https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.10240.0/shared/minwindef.h#L87
Notice it also #defines IN and OUT to empty strings as well.

**Anything else we should know about your project / environment**

egorpugin:
[...]

PR #7168, opened 2020-02-02
bug: #7076 adds ERROR to windows portability files

egorpugin:
Same as https://github.com/protocolbuffers/protobuf/pull/7088 but for `ERROR`.
Issue: https://github.com/protocolbuffers/protobuf/issues/7076


rafi-kamal:
Do we want it to merge in master instead? We are currently in the middle of the 3.11.3 release process, I'm not sure if there will be 3.11.4 release in future.

egorpugin:
I also thought about master, but created as previous PR.
Maybe after merge you could cherry-pick it into master?

PR #7173, opened 2020-02-04
Revisit how the WKTs are bundled with ObjC.

thomasvl:
There are have been a few issues around people using case sensitive file systems
what Xcode/clang does when looking at the paths. In attempts to solve one set of
warnings, new warnings/errors happened in different setup. So, to hopefully put
these problem away for got, move the WKTs to be at the same level as the other
headers.

- Revert "Override CocoaPods module to lowercase (#6464)"
  This reverts commit 479ba8226bed9c0986693f240c4d6f72dc2d0368.
- Move WKTs to the objectivec directory and make the old headers shim back to
  the new locations.
- Update objectivec/generate_well_known_types.sh to check them one at a time
  and to deal with the new locations for them.


thomasvl:
@paulb777 @dmaclach fyi

objectivec/generate_well_known_types.sh
```diff
@@ -53,24 +54,36 @@ declare -a RUNTIME_PROTO_FILES=( \
...
+  DIR=${PROTO_FILE%/*}
+  BASE_NAME=${PROTO_FILE##*/}
+  # Drop the extension
+  BASE_NAME=${BASE_NAME%.*}
+  OBJC_NAME=$(echo "${BASE_NAME}" | awk -F _ '{for(i=1; i<=NF; i++) printf "%s", toupper(substr($i,1,1)) substr($i,2);}')
+
+  for EXT in "${OBJC_EXTENSIONS[@]}"; do
+    if ! diff "${ObjCDir}/GPB${OBJC_NAME}${EXT}" "${TMP_DIR}/${DIR}/${OBJC_NAME}${EXT}" > /dev/null 2>&1 ; then
+      if [[ "${CHECK_ONLY}" == 1 ]] ; then
+        echo "ERROR: The WKTs need to be regenerated! Run $0"
```
dmaclach: Can you change WKT to WellKnownTypes just so that I don't have to remember the acronym?

objectivec/generate_well_known_types.sh
```diff
@@ -53,24 +54,36 @@ declare -a RUNTIME_PROTO_FILES=( \
...
+        exit 1
+      fi
+
+      echo "Updating GPB${OBJC_NAME}${EXT}"
```
dmaclach: while here could you add "INFO: " as a prefix on the log string?

objectivec/generate_well_known_types.sh
```diff
@@ -53,24 +54,36 @@ declare -a RUNTIME_PROTO_FILES=( \
...
+      cp "${TMP_DIR}/${DIR}/${OBJC_NAME}${EXT}" "${ObjCDir}/GPB${OBJC_NAME}${EXT}"
+      DID_COPY=1
+    fi
+  done
+done
+
+if [[ "${DID_COPY}" == 0 ]]; then
   echo "Generated source for WellKnownTypes is current."
```
dmaclach: and while here could you add "INFO: " as a prefix on the log string?

objectivec/generate_well_known_types.sh
thomasvl: Given the line ends with the script to run, which spells out what is being generated, I don't think it is needed here.

PR #7175, opened 2020-02-05
Deterministic `protoc --descriptor_set_out`

mpartel:
Repeatable protoc output is useful because it's sometimes checked into version control, and rerunning protoc should not generate unnecessary diffs.

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -2228,12 +2228,18 @@ bool CommandLineInterface::WriteDescriptorSet(
   }
 
   io::FileOutputStream out(fd);
-  if (!file_set.SerializeToZeroCopyStream(&out)) {
+
+  io::CodedOutputStream coded_out(&out);
+  // Determinism is useful here because build outputs are sometimes checked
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -2228,12 +2228,18 @@ bool CommandLineInterface::WriteDescriptorSet(
...
+  // Determinism is useful here because build outputs are sometimes checked
+  // into version control.
+  coded_out.SetSerializationDeterministic(true);
+  if (!file_set.SerializeToCodedStream(&coded_out)) {
     std::cerr << descriptor_set_out_name_ << ": " << strerror(out.GetErrno())
               << std::endl;
     out.Close();
     return false;
   }
+  coded_out.Trim();
```
acozzette: We usually just create the CodedOutputStream inside a block with curly braces: that way the destructor will take care of any cleanup and this Trim() call won't be needed.

mpartel: Done

acozzette:
Thanks, @mpartel!

PR #7180, opened 2020-02-06
Fix std::is_pod deprecated in C++20

stac47:
I followed the deprecation message to fix the compilation for C++ 20. It works like a charm with gcc 10 (almost HEAD).


egorpugin:
+1

PR #7190, opened 2020-02-08
[bazel] Move Java runtime/toolchains into //java

Yannic:
This change moves `java_library` targets from the top-level BUILD file
into `//java/{core,lite,util}` and declares `alias` targets to point to
their new locations (hence, this is not a breaking change).

This will allow users that don't use Java to stop depending on
`@rules_java` (e.g. as requested in
https://github.com/bazelbuild/rules_scala/pull/989#issuecomment-583405161).

Note that there is no intention to deprecate + remove the top-level
targets in the foreseeable future.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,7 +1,6 @@
 # Bazel (https://bazel.build/) BUILD file for Protobuf.
 
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test", "objc_library", native_cc_proto_library = "cc_proto_library")
-load("@rules_java//java:defs.bzl", "java_library")
 load("@rules_proto//proto:defs.bzl", "proto_lang_toolchain", "proto_library")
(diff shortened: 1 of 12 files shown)


ittaiz:
Nice! Do you think the same can be done for python?

Yannic:
I haven't looked into doing the same for other languages yet. Will do when this lands.

ittaiz:
Thanks!

Yannic:
@rafi-kamal Thanks! Looks like there were infra issues yesterday. Maybe re-run the tests?

rafi-kamal:
@Yannic can you please add the build files to java_EXTRA_DIST in https://github.com/protocolbuffers/protobuf/blob/master/Makefile.am?

The PHP and Python failures are unrelated.

Yannic:
@rafi-kamal Thanks, done!

PR #7195, opened 2020-02-11
Fix for wrappers with a zero value

haberman:
The previous optimizations for wrapper types had a bug that prevented    
wrappers from registering as "present" if the "value" field was not    
present on the wire.    
    
In practice the "value" field will not be serialized when it is zero,    
according to proto3 semantics, but due to the optimization this    
prevented it from creating a new object to represent the presence of the    
field.    
    
The fix is to ensure that if the wrapper message is present on the wire,    
we always initialize its value to zero.

Fixes https://github.com/protocolbuffers/protobuf/issues/7029


djquan:
@googlebot I consent

PR #7197, opened 2020-02-11
Address Intel compiler problems

chuckatkins:


cmake/CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ else()
   set(CMAKE_CXX_EXTENSIONS OFF)
 endif()
 
+# Quiet some noisy Intel warnings
+if (CMAKE_CXX_COMPILER_ID MATCHES Intel)
+  string(APPEND CMAKE_CXX_FLAGS " -diag-disable=2196")
```
acozzette: Could you add a comment here saying what kind of warning this disables?

src/google/protobuf/arena.h
```diff
@@ -338,8 +338,10 @@ class PROTOBUF_EXPORT alignas(8) Arena final {
   template <typename T>
   PROTOBUF_ALWAYS_INLINE static T* CreateArray(Arena* arena,
                                                size_t num_elements) {
+#ifndef __INTEL_COMPILER
```
acozzette: What goes wrong here with the Intel compiler?

cmake/CMakeLists.txt
chuckatkins: The Intel compiler has difficulty dealing with `noinline` member functions for template classes since it treats all template code as implicitly inlineable.  As an example:

```cxx
// foo.cxx
#include <iostream>

template<typename T>
class foo
{
public:
  void bar1(int arg) { std::cout << arg; }

  __attribute__((noinline)) void bar2(T arg) { std::cout << arg; }

  inline void bar3(T arg) { std::cout << arg; }
};

void baz()
{
  foo<double> f;

  f.bar1(42);    // OK
  f.bar2(42.0);  // Warning
  f.bar3(42.0);  // OK
}
```
```bash
$ icpc -c foo.cxx 
foo.cxx(21): warning #2196: routine is both "inline" and "noinline"
```

This is a pretty common pattern throughout protobuf and causes `warning #2196: routine is both "inline" and "noinline"` to spam the output with >800 warnings originating from over a quarter of the files in the codebase.  The issue itself seems to be more of a compiler problem in that icpc isn't able to produce noinline member functions of template classes so it seems most appropriate to just silence the warning.

src/google/protobuf/arena.h
chuckatkins: With `icpc` the static assert fails for the `google::protobuf::internal::ExtensionSet::KeyValue` type, even though it certainly fits the definition of POD type and trivially construct-able:
```
/home/khq.kitware.com/chuck.atkins/Code/protobuf/source/src/google/protobuf/arena.h(342): error: static assertion failed with "CreateArray requires a trivially constructible type"
      static_assert(std::is_pod<T>::value,
      ^
          detected during instantiation of "T *google::protobuf::Arena::CreateArray<T>(google::protobuf::Arena *, size_t={unsigned long}) [with T=google::protobuf::internal::ExtensionSet::KeyValue]" at line 193 of "/home/khq.kitware.com/chuck.atkins/Code/protobuf/source/src/google/protobuf/extension_set.cc"

compilation aborted for /home/khq.kitware.com/chuck.atkins/Code/protobuf/source/src/google/protobuf/extension_set.cc (code 2)
make[2]: *** [CMakeFiles/libprotobuf.dir/home/khq.kitware.com/chuck.atkins/Code/protobuf/source/src/google/protobuf/extension_set.cc.o] Error 2
make[1]: *** [CMakeFiles/libprotobuf.dir/all] Error 2
make: *** [all] Error 2
```

cmake/CMakeLists.txt
chuckatkins: Added comment explaining the warning and why it's silenced.

src/google/protobuf/arena.h
chuckatkins: Added comment noting that `icc` mis-evaluates some types as non-pod

PR #7202, opened 2020-02-12
Update the Xcode version number in Kokoro

rafi-kamal:
Currently all of our MacOS tests are failing with the error: "autom4te: need GNU m4 1.4 or later: /usr/bin/m4". This is likely to be happening because Kokoro has updated their Xcode version to 11.3.


rafi-kamal:
All the Mac tests are passing now except PHP7.0 and Python. The Python one is a known issue, Bo is looking into the PHP test failure.

PR #7204, opened 2020-02-12
Fix for JSON serialization of 0/empty-valued wrapper types (#7198)

rafi-kamal:
* Fixed Ruby JSON serialization of 0/empty wrapper fields.

* Removed newly-passing conformance tests from the failure list.


rafi-kamal:
Chery pick of https://github.com/protocolbuffers/protobuf/pull/7198

PR #7207, opened 2020-02-13
Downgrade Xcode version to 10.3

rafi-kamal:
Building protoc with Xcode 11.3 gives an error: `ld: library not found for -latomic`

Sponge link: https://sponge.corp.google.com/target?id=c9ac8abf-af0c-453c-a027-18720243aa58&target=protobuf%2Fgithub%2F3.11.x%2Frelease%2Fprotoc%2Fmacos%2Frelease&searchFor=

### DIFF
--- a/kokoro/macos/prepare_build_macos_rc
+++ b/kokoro/macos/prepare_build_macos_rc
@@ -5,11 +5,11 @@
 ##
 # Select Xcode version
 
-# Remember to update the Xcode version when Xcode_11.3.app is not available.
+# Remember to update the Xcode version when Xcode_10.3.app is not available.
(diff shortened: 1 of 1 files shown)


thomasvl:
You're likely going to need to figure out what is going wrong in the autoconf/configure setup, Xcode 10.x is old already (11.4 is currently in beta), so you'll need to ensure things build with Xcode 11.

rafi-kamal:
The underlying problem is Xcode 11.3 no longer supports building for 32 bit architecture. I think dropping the 32 bit support for Mac is a better solution instead of keep using the old Xcode version.

PR #7217, opened 2020-02-15
examples: update Go example

neild:
Update the Go example to use Go modules: Move the example into a
directory containing a go.mod file, change the installation instructions
to use "go install".

Add a go_package option to addressbook.proto.

Update examples to use the Go APIv2 package
(google.golang.org/protobuf/proto).


acozzette:
@neild Thanks! I think we need to make some tweaks to Makefile.am to fix the distcheck failures, but I will just do that in a subsequent PR.

PR #7218, opened 2020-02-15
[bazel] Remove bootstrap hack from cc_proto_library and add interop with proto_library

Yannic:
Bazel had a native `cc_proto_library` for more than 2 years now.
This is the first step towards removing that rule from the Protobuf
repo.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -114,6 +114,7 @@ LINK_OPTS = select({
 
 load(
     ":protobuf.bzl",
+    "adapt_proto_library",
     "cc_proto_library",
     "internal_copied_filegroup",
     "internal_gen_well_known_protos_java",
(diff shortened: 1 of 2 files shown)


acozzette:
@Yannic Thanks for working on this! This looks great but I was wondering if you have a plan for addressing the circular dependencies between protobuf and Bazel's native cc_proto_library() rule. Bazel needs to depend on protobuf for this, but at the same time protobuf presumably needs to depend on cc_proto_library() for its own protos (such as descriptor.proto). The internal_bootstrap_hack flag was an attempt to deal with our internal protos. With the native cc_proto_library() rule, I'm not sure how this will work. Currently it looks to me like you would end up with ODR violations if you depend on :descriptor_cc_proto, since that would duplicate the descriptor.pb.cc that is built into :protobuf.

Yannic:
@acozzette Thanks for your review!

Bazel's `cc_proto_library` realizes excluding WKPs that are compiled into Protobuf through `proto_lang_toolchain`'s `blacklisted_protos` attribute (compare `//:cc_toolchain`). Unfortunately, that attribute isn't perfect, but that weirdness only affects the protobuf repo itself and isn't an issue for consumers (https://github.com/bazelbuild/bazel/issues/10590, TL;DR broken because file `//:foo` != `@com_google_protobuf//:foo`). There are plans to fix that by excluding protos based on import paths instead of file equivalence.

shahms:
What is the expected replacement for the `cc_wkt_protos` target?  This doesn't appear to be documented anywhere that I can find. Is it just for each client to add a `cc_proto_library(name = "my_any", deps = ["@com_google_protobuf//:any_proto"])`?

PR #7230, opened 2020-02-18
Improves performance of json_stream_parser.cc by factor 1000

wsw2016:
JsonStreamParser::GetNextTokenType() uses HasPrefixString a lot on StringPiece as input. For each call two std::strings are constructed, compared and destroyed. Parsing of json-files with 50-60 MB in debug mode takes minutes.

With this modification the speed is comparable with other Json parsers.

### DIFF
--- a/src/google/protobuf/stubs/strutil.h
+++ b/src/google/protobuf/stubs/strutil.h
@@ -35,6 +35,7 @@
 
 #include <stdlib.h>
 #include <vector>
+#include <cstring>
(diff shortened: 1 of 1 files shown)


acozzette:
Thanks, @wsw2016!

yoavmil:
In which release is this fix included?

arhuaco:
> In which release is this fix included?

v3.12.0-rc1

PR #7234, opened 2020-02-21
Fix codeblock formatting

socsieng:


acozzette:
@socsieng Thank you for the fix and sorry for taking so long to get to this. This looks to me like a good change, but the problem appears to have been fixed already in 75553e9ec8a4a18df18208d44bb6e718cb94256a.

PR #7237, opened 2020-02-21
[bazel] Update gtest and deprecate //external:{gtest,gtest_main}

Yannic:
This change updates the gtest-version used by Bazel.
Also, `//external:{gtest,gtest_main}` is deprecated so we can remove some
of the uses of the discouraged `bind` function.

RELNOTES[bazel]: Starting with Protobuf 3.13.0, building and running
Protobuf tests requires `@com_google_googletest//:{gtest,gtest_main}`
instead of `//external:{gtest,gtest_main}`. Use
`--@com_google_protobuf//:incompatible_use_com_google_googletest=true`
to verify your workspace is not affected by this change.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,0 +1,3 @@
+# These are fetched as external repositories.
+third_party/benchmark
+third_party/googletest
--- a/BUILD
+++ b/BUILD
@@ -1,5 +1,6 @@
 # Bazel (https://bazel.build/) BUILD file for Protobuf.
 
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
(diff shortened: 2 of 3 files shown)


Yannic:
Note that this only changes the gtest version for Bazel. Other builds still use the submodule.

Yannic:
@rafi-kamal Thanks for your review!

As usual, `MacOS PHP & Python` failures are unrelated. For `Mergable`, does it make sense to add `bazel` to the list of labels for `release notes: yes` https://github.com/protocolbuffers/protobuf/blob/master/.github/mergeable.yml#L14?

rafi-kamal:
I think that sounds good, I've sent out https://github.com/protocolbuffers/protobuf/pull/7275.

Thetoan1 (commented):
No

PR #7246, opened 2020-02-23
Swap registry from input when parsing from existing input

### LINKED ISSUE #7241: ParseFrom(CodedInputStream) does not properly deserialize proto2 extensions
Version: v3.11.4
Language: C#

N/A

N/A

**What did you do?**
Parsing a proto2 message with extensions from a manually-constructed `CodedInputStream` does not properly deserialize the extensions.

```csharp
        [Test]
        public void TestParseFromCodedInput()
        {
          var message = new TestAllExtensions();
          message.SetExtension(OptionalBoolExtension, true);
          var serialized = message.ToByteArray();

          using (var cis = new CodedInputStream(serialized))
          {
            var other = TestAllExtensions.Parser
              .WithExtensionRegistry(new ExtensionRegistry { OptionalBoolExtension })
              .ParseFrom(cis);

            Assert.True(message.HasExtension(OptionalBoolExtension));

            // ******************************************************
            // FAILS HERE
            Assert.True(other.HasExtension(OptionalBoolExtension));
            // ******************************************************
            
            Assert.AreEqual(message, other);
            Assert.AreEqual(message.CalculateSize(), other.CalculateSize());
          }
        }
```

This seems to be because the internal `MergeFrom` method that gets called with the stream does not set the parser's `ExtensionRegistry` on the stream. Adding it as below seems to fix the issue, though I'm not sure if there are also other places that would need to be fixed as well.
```diff
        internal void MergeFrom(IMessage message, CodedInputStream codedInput)
        {
            bool originalDiscard = codedInput.DiscardUnknownFields;
            try
            {
                codedInput.DiscardUnknownFields = DiscardUnknownFields;
+               codedInput.ExtensionRegistry = Extensions;
                message.MergeFrom(codedInput);
            }
            finally
            {
                codedInput.DiscardUnknownFields = originalDiscard;
            }
        }
```

**Additional Notes**
It would also be nice if an ExtensionRegistry could be passed directly to a `MessageParser`'s `ParseFrom` method, which I would then expect to be used without needing to construct a new `MessageParser` instance (which the current `WithExtensionRegistry` method does). This seems to be available in the Java API https://github.com/protocolbuffers/protobuf/blob/31d6b4e30e8b09eb7c23b24913805f49ddb5f6c9/java/core/src/main/java/com/google/protobuf/Parser.java#L68

This would save allocations and simplify code that might be operating on the base `MessageParser` type but also needs to respect extensions.

The `ExtensionRegistry` cannot be set on the `CodedInputStream` during construction because it is internal and there is no constructor that exposes it https://github.com/protocolbuffers/protobuf/blob/f20be839276cfc1129c12d89924164624ef3796d/csharp/src/Google.Protobuf/CodedInputStream.cs#L278

ObsidianMinor:
Fixes #7241

### DIFF
--- a/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
+++ b/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
@@ -376,5 +376,18 @@ namespace Google.Protobuf
             TestGroupExtension extendable_parsed = TestGroupExtension.Parser.WithExtensionRegistry(new ExtensionRegistry() { TestNestedExtension.Extensions.OptionalGroupExtension }).ParseFrom(bytes);
             Assert.AreEqual(message, extendable_parsed);
         }
+
+        [Test]
+        public void RoundTrip_ParseUsingCodedInput()
+        {
+            var message = new TestAllExtensions();
+            message.SetExtension(UnittestExtensions.OptionalBoolExtension, true);
+            byte[] bytes = message.ToByteArray();
+            using (CodedInputStream input = new CodedInputStream(bytes))
+            {
+                var parsed = TestAllExtensions.Parser.WithExtensionRegistry(new ExtensionRegistry() { UnittestExtensions.OptionalBoolExtension }).ParseFrom(input);
+                Assert.AreEqual(message, parsed);
+            }
+        }
     }
 }
--- a/csharp/src/Google.Protobuf/MessageParser.cs
+++ b/csharp/src/Google.Protobuf/MessageParser.cs
@@ -159,14 +159,17 @@ namespace Google.Protobuf
         internal void MergeFrom(IMessage message, CodedInputStream codedInput)
         {
             bool originalDiscard = codedInput.DiscardUnknownFields;
+            ExtensionRegistry originalRegistry = codedInput.ExtensionRegistry;
             try
             {
                 codedInput.DiscardUnknownFields = DiscardUnknownFields;
+                codedInput.ExtensionRegistry = Extensions;
                 message.MergeFrom(codedInput);
             }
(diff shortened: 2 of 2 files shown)


csharp/src/Google.Protobuf.Test/GeneratedMessageTest.Proto2.cs
```diff
@@ -376,5 +376,18 @@ public void RoundTrip_NestedExtensionGroup()
...
+
+        [Test]
+        public void RoundTrip_ParseUsingCodedInput()
+        {
+            var message = new TestAllExtensions();
+            message.SetExtension(UnittestExtensions.OptionalBoolExtension, true);
+            byte[] bytes = message.ToByteArray();
+            using (CodedInputStream input = new CodedInputStream(bytes))
+            {
+                var parsed = TestAllExtensions.Parser.WithExtensionRegistry(new ExtensionRegistry() { UnittestExtensions.OptionalBoolExtension }).ParseFrom(input);
```
Jensaarai: Thank you!

If at all possible, can there also be a way to pass an extension registry to ParseFrom/MergeFrom? It would be useful to be able to use the static MessageParser that's already present for a message and not have to construct a new one, or to be able to construct a new message and use MergeFrom with some stream and an extension registry. The current behavior of having `WithExtensionRegistry` create a new MessageParser each time is not ideal when parsing many different types of messages very quickly e.g. in an RPC system.

jtattermusch:
It this PR still needed to fix a problem?

ObsidianMinor:
I think so, assuming by the fact that there's still no conflicts. It should've been swapping out the extension registry the whole time like it does with DiscardUnknownFields.

Issue #5493, opened 2018-12-19
CMake support for plugins in generate_protobuf


pgoodman:
**What language does this apply to?**
All languages.

**Describe the problem you are trying to solve.**
Using CMake to build a project that uses gRPC, I'd like to be able to provide a plugin command to `generate_protobuf`. Partial support is there in the form of `LANGUAGE grpc` and `GENERATE_EXTENSIONS .grpc.pb.h .grpc.pb.cc` but there's no way to pass in the plugin information.

**Describe the solution you'd like**
A way to pass in the plugin name and path.

**Describe alternatives you've considered**
Trying to force it through the `LANGUAGE`; it doesn't work.

faaxm:
@pgoodman @acozzette Only found this issue after creating PR #7249 that fixes this. Please have a look. :)

PR #7249, opened 2020-02-24
cmake: add 'plugin' argument to protobuf_generate

faaxm:
Added an optional `PLUGIN` argument to `protobuf_generate` which
will be forwarded to the `--plugin=` argument of `protoc`.

This can be used, among other things, to run `protoc` with the grpc extension, as is done here:
https://github.com/faaxm/exmpl-cmake-grpc/blob/master/proto/CMakeLists.txt

Addresses issue #5493.


faaxm:
@googlebot I signed it!

acozzette:
Thanks, @faaxm.

Issue #7216, opened 2020-02-14
simple json validation against predefined protobuf schema


holyshipt:
**What language does this apply to?**
Ideally language neutral.
No protobuf syntax changes
First iteration targets at Java, then all languages.

**Describe the problem you are trying to solve.**
Provide generic protobuf schema enforcement capability against dynamic json messages to avoid breaking changes.  e.g., structural, types.
The schema enforcement on json will traverse syntax tree and compute mismatch.

**Describe the solution you'd like**
A general util function or an example demonstrate how to validate json message, by traversing both syntax trees. 

**Describe alternatives you've considered**
NA

**Additional context**
Example predefined protobuf
```
syntax = "proto3";
package com.proofpoint.forensics.protobuf;
import "google/protobuf/struct.proto";

option java_package = "com.proofpoint.forensics.protobuf";
option java_outer_classname = "TestMessage";

message Message {
    message M1 {
        message M2 {
            message M3 {
                message M4 {
                    string name = 1;
                }
                M4 m4 = 1;
            }
            M3 m3 = 1;
        }
        M2 m2 = 1;
    }
    M1 m1 = 1;
}
```
Example target json:
```json
{
    "m1": {
        "m2": 0
    }
}
```
Example implementation
```java
ParserResults diff(Value root, Class<? extends Message> clazz) {
      // compute diff and stored in a result buffer
}

// client
diff(root, TestMessage.Message.class)
// prints something like below:
// m2 supposed to be object or map but found number 
```

PR #7272, opened 2020-03-02
[Java] add protobuf util to validate dynamic json data

holyshipt:
#7216 prototyping

Hi, here's a Java util to validate json data, it basically traverses dynamic json tree to examine every possible mismatch, current supported json type includes: `number/string/bool/null/list/struct`.

**Usage**
Use `JsonValidator.validate(Value, Class)` as static function
```java
Value.Builder builder = Value.newBuilder();
mergeFromJson("JSONSTRING", builder);
List<String> result = JsonValidator.validate(builder.build(), TestJsonValidator.class);
```

**Output**
Generate following mismatch warnings:
```
<field> supposed to be object but not.
<field> supposed to be list but not.
<field> supposed to be map but not.
<field> supposed to be string but not.
<field> supposed to be numeric but not.
<field> supposed to be boolean but not.
```
Throw following exceptions
```
UnsupportedOperationException - Java type is not supported
err message: Cannot convert protobuf type: unsupported type <java type>
UnsupportedOperationException - dynamic map key or value is null
err message: Both map key and value are required.
```

Some tunable behavior/engineering opportunities 
- It collects all mismatches(warnings) into a list.
- It does not differentiate fields with same name (e.g., heterogeneous list, nested struct). 
- For null value in json data, it skips further validation on that field. 
- It skips protobuf "optional" field if json data doesn't have it. (commented out)
- Format of warnings

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonValidator.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonValidator.java
@@ -124,6 +124,7 @@ public final class JsonValidator
       for (Entry<String, Value> entry : node.getStructValue().getFieldsMap().entrySet()) {
         helper(++depth, fields.get(1), entry.getValue(), warnings);
       }
+      return;
     }
 
     if (field.isRepeated()) {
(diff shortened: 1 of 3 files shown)


holyshipt:
@googlebot I signed it!

acozzette:
@holyshipt Thank you for the pull request but unfortunately I don't think we want to accept this. Users can already attempt to parse a JSON message to determine whether it's valid. I imagine that validating without parsing could in principle be faster, but I think it's more important to rely on parsing as the single consistent way of validating JSON.

PR #7278, opened 2020-03-04
Fix #6174: Add parentheses around min/max methods

### LINKED ISSUE #6174: std::min() conflict
Version: master
Language: C++
Windows

protobuf-lite
 
generated enum using std::min, but conflict with windows macro, I have to modify the generated code.
QQ图片20190524121014

jozefizso:
Compiling protoc generated code on Windows usually leads to errors because of clash between min/max macros defined in Windows SDK headers and the std:: min/max methods.

Generated protobuf code will include parentheses around min and max methods (which are used in enums) to fix this.

Fixes #6174

PS: It's not so easy for consumers of protobuf code to just define `NOMINMAX` to get rid of this issue as this will cause a lot of Windows code to not compile correctly.

C++
Release Notes: No

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_enum.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_enum.cc
@@ -125,9 +125,9 @@ void EnumGenerator::GenerateDefinition(io::Printer* printer) {
     if (descriptor_->value_count() > 0) format(",\n");
     format(
         "$classname$_$prefix$INT_MIN_SENTINEL_DO_NOT_USE_ = "
-        "std::numeric_limits<$int32$>::min(),\n"
+        "(std::numeric_limits<$int32$>::min)(),\n"
         "$classname$_$prefix$INT_MAX_SENTINEL_DO_NOT_USE_ = "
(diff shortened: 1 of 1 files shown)


jozefizso:
I have no idea what is the issue with Mergeable: **Failing after 1s — Mergeable run returned Status ***FAIL*****

acozzette:
@jozefizso Thank you for sending the fix. I would already to expect this to be fixed by these [lines](https://github.com/protocolbuffers/protobuf/blob/f130827371fc73ab6e24d44313af6577cc2f9d57/src/google/protobuf/port_def.inc#L397) in port_def.inc, though (and the corresponding ones in port_undef.inc). These two textual headers should be present in the generated code to temporarily undefine problematic macros. Are you getting a compiler error despite those headers?

jozefizso:
I wasn't aware of those defines as we are not using the latest protobuf yet.

acozzette:
OK, let me close this then since I think this should be fixed in recent versions. Let me know if it still doesn't work and then we can reopen this.

PR #7280, opened 2020-03-05
Add compile-time tests for RTTI to port_def.inc

Yannic:
This eliminates an ODR violation where users compile and install
Protobuf's C++ runtime without RTTI but do not define
GOOGLE_PROTOBUF_NO_RTTI when compiling their project with
the generated code.

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -308,6 +308,15 @@
 
 #if defined(GOOGLE_PROTOBUF_NO_RTTI) && GOOGLE_PROTOBUF_NO_RTTI
 #define PROTOBUF_RTTI 0
+#elif defined(__has_feature)
+// https://clang.llvm.org/docs/LanguageExtensions.html#has-feature-and-has-extension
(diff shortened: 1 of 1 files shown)


src/google/protobuf/port_def.inc
```diff
@@ -308,6 +308,15 @@
 
 #if defined(GOOGLE_PROTOBUF_NO_RTTI) && GOOGLE_PROTOBUF_NO_RTTI
 #define PROTOBUF_RTTI 0
+#elif defined(__has_feature)
+// https://clang.llvm.org/docs/LanguageExtensions.html#has-feature-and-has-extension
+#define PROTOBUF_RTTI __has_feature(cxx_rtti)
+#elif !defined(__cxx_rtti) && __cplusplus >= 199711L
```
acozzette: I think you can leave out the `__cplusplus` check since we require C++11 anyway.

Yannic: Indeed, done

acozzette:
Thanks, @Yannic.

Issue #2719, opened 2017-02-15
Undefined `__android_log_write` on NDK builds


swansontec:
When I build Protocol Buffers 3.2.0 using the NDK, I get the following linker error:

```
libtool: link: arm-linux-androideabi-clang++ -std=c++11 -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare -isystem/Users/william/tmp/build/prefix/android/arm/include -fPIC -O2 -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16 -mthumb -fPIC -march=armv7-a -Wl,--fix-cortex-a8 -o protoc google/protobuf/compiler/main.o  -L/Users/william/tmp/build/prefix/android/arm/lib ./.libs/libprotobuf.a ./.libs/libprotoc.a /Users/william/tmp/build/protobuf/android-arm/protobuf-3.2.0/src/.libs/libprotobuf.a -lz
/Users/william/tmp/build/protobuf/android-arm/protobuf-3.2.0/src/.libs/libprotobuf.a(common.o):google/protobuf/stubs/common.cc:function google::protobuf::internal::DefaultLogHandler(google::protobuf::LogLevel, char const*, int, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> > const&): error: undefined reference to '__android_log_write'
/Users/william/tmp/build/protobuf/android-arm/protobuf-3.2.0/src/.libs/libprotobuf.a(common.o):google/protobuf/stubs/common.cc:function google::protobuf::internal::DefaultLogHandler(google::protobuf::LogLevel, char const*, int, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> > const&): error: undefined reference to '__android_log_write'
```

There were no errors on 3.1.0 with an identical build setup.

## Workaround
The problem is that `-llog` needs to be passed to the linker to make the `__android_log_write` function available. By temporarily defining `LDFLAGS="-llog"` before running configure, I can work around the problem. The Protocol Buffers configure script should be doing this automatically.

## Build setup
[...]

PR #7288, opened 2020-03-08
link log library on Android

RPG3D:


pthom:
This seems to try to provide a fix for https://github.com/protocolbuffers/protobuf/issues/2719

PetrJarosik:
Merge this fix please. We are not able to build the protobuf for Android target without this fix!

PR #7289, opened 2020-03-09
Remove unnecessary branch from ReadTag

jtattermusch:
I found this when doing some refactorings to how parsing code works.

The purpose of the following snipped at the end of CodedInputStream.ReadTag is unclear.
I don't see a reason why this would be necessary and it actually seems wrong.
- What it does is that if we just read a tag that that is immediately followed by the limit of the stream (but the tag itself is still readable just fine) we will artificially return a tag of 0 and there seems to be no reason we we would want that
- this also potentially affects performance (one extra check after reading a tag).
- if I remove the check, the tests still seem to be passing.
```
if (ReachedLimit)
{
    return 0;
}
```

The snippet was by #5183
https://github.com/protocolbuffers/protobuf/commit/6f73c509365fcb9bca614f31d0d1234ff1323b58#diff-1d948236189c1eb142de5ed1ed631a40R386

### DIFF
--- a/csharp/src/Google.Protobuf.Test/IssuesTest.cs
+++ b/csharp/src/Google.Protobuf.Test/IssuesTest.cs
@@ -33,6 +33,7 @@
 using Google.Protobuf.Reflection;
 using UnitTest.Issues.TestProtos;
 using NUnit.Framework;
+using System.IO;
 using static UnitTest.Issues.TestProtos.OneofMerging.Types;
 
 namespace Google.Protobuf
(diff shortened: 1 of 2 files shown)


jtattermusch:
@ObsidianMinor   can you provide an explanation why the snippet is important? (and provide a test case that demonstrates it).

FYI, the code change is just removing the snippet, the other lines is just adjusting CR-LF line endings (done automatically by the editor).

jtattermusch:
CC @jskeet  @JamesNK

ObsidianMinor:
This seems to be a mistake left over from when the CodedInputStream tracked groups. On my original proto2 prototype it calls ReachedTagLimit in that position but some way or another it got changed and then never got removed when we later moved the check for the end tag into the group message itself.

jtattermusch:
@ObsidianMinor  the development of that snippet is a bit more interesting:
First you added this in https://github.com/protocolbuffers/protobuf/pull/5183/commits/8c13fe77bce990ad70993250223fc2517c9e3c3d#diff-1d948236189c1eb142de5ed1ed631a40R387,
after that you added the ReachedTagLimit condition here:
https://github.com/protocolbuffers/protobuf/pull/5183/commits/1048bf0976675d47e856069810e1013ae1baee1a#diff-1d948236189c1eb142de5ed1ed631a40R387
and finally you removed the ReachedTagLimit, but left the original snippet
https://github.com/protocolbuffers/protobuf/pull/5183/commits/97b7da7143e87a58fa9e921f9fc473b6a0e3dc8f#diff-1d948236189c1eb142de5ed1ed631a40R386

But looks like we can safely remove that, do you agree?

ObsidianMinor:
Yes it's a useless check that should've been remove back then.

csharp/src/Google.Protobuf.Test/IssuesTest.cs
```diff
@@ -90,5 +91,25 @@ public void OneofMerging()
             merged.MergeFrom(message2);
             Assert.AreEqual(expected, merged);
         }
+
+        // See https://github.com/protocolbuffers/protobuf/pull/7289
```
anandolee: Few words comments here will be better to use PR link

jtattermusch:
@anandolee  this should be now ready to merge.

PR #7294, opened 2020-03-10
Generate documentation for internal.container.

dlj-NaN:
This includes some changes to generate_docs.py to explicitly allow named modules.


Hanae1963 (commented):
Thanks

PR #7296, opened 2020-03-10
Fix compile error on GHS from using __has_cpp_attribute with clang::reinitializes

Breakthrough:
clang:: prefixed attributes are not supported on GHS compiler, even when it supports __has_cpp_attribute.  This change only checks for clang::reinitializes if we're compiling on clang.


Breakthrough:
Closing until further clarification regarding CLA.

Issue #6683, opened 2019-09-21
min/max namespace collision with Winsock2.h


dkonik:
Version: 3.9.1 etc.
Language: C++
OS: Windows 18932
Compiler: VS 2019 (c++ 2017, if that helps at all)
Output Type: Static Library

Issue Description:
If you have an enum defined in your proto file, and you include `winsock2.h` prior to including the generated `.h` file, you will get an error during compilation since `WinDef.h` defines a min and max macro. There are a few ways to get around this, but I won't be the one to make any suggestions.

Repro Steps:
1) Create a simple dummy proto file with an enum (I'll include an example that I can use to repro at the bottom)
**What did you expect to see**
2) Generate it
3) Create a simple new static library project which includes the protobuf lib (statically linked), the protobuf headers, and the generated files
4) If, in one of the files of the project, you do something like:
``` cpp
#include "test.pb.h"
#include <WinSock2.h>
``` 
It compiles fine. If, however, you reverse the order and do:
``` cpp
#include <WinSock2.h>
#include "test.pb.h"
```
It fails to compile.

Example Proto:
```
syntax = "proto3";

enum MessageType {
        // Reserved
	InvalidMessage = 0;
	// A header for stream, including the height, width, etc
	StreamHeaderMessage = 1;
}

// The header for a stream
message StreamHeader {
    uint32 Width = 1;
    uint32 Height = 2;
    uint32 FrameRate = 3;
}
```
I've verified that if I remove the enum, there are no errors.

Compiler output (reduced to the errors):
```
1>test.pb.h(68,99): error C2589:  '(': illegal token on right side of '::'
1>test.pb.h(68,99): error C2062:  type 'unknown-type' unexpected
1>test.pb.h(68,99): error C3805:  'type': unexpected token, expected either '}' or a ','
```
Where line 68 is the third line in the enum (the line with `min`):
```
enum MessageType : int {
  InvalidMessage = 0,
  StreamHeaderMessage = 1,
  MessageType_INT_MIN_SENTINEL_DO_NOT_USE_ = std::numeric_limits<::PROTOBUF_NAMESPACE_ID::int32>::min(),
  MessageType_INT_MAX_SENTINEL_DO_NOT_USE_ = std::numeric_limits<::PROTOBUF_NAMESPACE_ID::int32>::max()
};
```

automotiveUser:
Hi @dkonik,

[...]

PR #7297, opened 2020-03-11
Fix generated pc.h files name collision with winsock2.h (#6683)

LevyCory:
Fixed name collisions between `winsock2.h` (and `windows.h`) min and max macros and `std::numeric_limits<uint32_t>::max/min` generated for protobuf enums.

#6683

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_enum.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_enum.cc
@@ -125,9 +125,9 @@ void EnumGenerator::GenerateDefinition(io::Printer* printer) {
     if (descriptor_->value_count() > 0) format(",\n");
     format(
         "$classname$_$prefix$INT_MIN_SENTINEL_DO_NOT_USE_ = "
-        "std::numeric_limits<$int32$>::min(),\n"
+        "(std::numeric_limits<$int32$>::min)(),\n"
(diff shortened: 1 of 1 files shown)


LevyCory:
@googlebot I signed it!

acozzette:
@LevyCory This should already be fixed on master and in recent protobuf versions. Are you still getting a name collision with protobuf 3.11?

LevyCory:
@acozzette Yeah, still happening with 3.11

acozzette:
@LevyCory Hmm, that's strange. The port_def.inc header has logic to temporarily undefine the min and max macros [here](https://github.com/protocolbuffers/protobuf/blob/f57ae5eac31d5322c8fcb350a6cdc13647e5ca53/src/google/protobuf/port_def.inc#L406), and they're later redefined in port_undef.inc at the end of the generated code. Do you know why this isn't preventing the collision?

LevyCory:
@acozzette Weird. I'll check again.

LevyCory:
@acozzette Protobuf 3.11 does solve this issue, I must've tried an older release.
Thank you.

acozzette:
@LevyCory OK great, I'm glad it already works then. Thank you for the pull request but let me close it since we seem to be good to go already.

LevyCory:
Thank you for your time. It might be a good idea to mark issue #6683 as resolved as well.

PR #7306, opened 2020-03-15
Bazel: Discontinue JDK 7 compatibility support

davido:
The primary reason to drop that compatibility is because Bazel is
adding unconditionally -parameters option that is not compatible
with JDK 7. This pollutes the build log with annoying warnings:

  warning: -parameters is not supported for target value 1.7. Use 1.8 or later.

Second reason is that nobody in the wild relies on JDK 7 any more
because it was discontinued years ago.

Test Plan:

Build :protobuf_java and confirm, that there are no warnings any more
and that major byte version 52 is produced, that corresponds to Java 8:

  $ bazel build protobuf_java
  $ javap -v -cp bazel-bin/java/core/libcore.jar com.google.protobuf.Any | grep major
  major version: 52

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -248,7 +248,6 @@ csharp_EXTRA_DIST=                                                           \
   csharp/src/Google.Protobuf/UnknownFieldSet.cs
 
 java_EXTRA_DIST=                                                                   \
-  java/BUILD                                                                       \
   java/README.md                                                                   \
   java/bom/pom.xml                                                                 \
   java/core/BUILD                                                                  \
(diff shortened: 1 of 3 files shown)


gkdn:
@dankurka Could we get this one in?

rules_closure has been historically patching it to hide warnings:
https://github.com/bazelbuild/rules_closure/pull/478/

Yannic:
@haberman @jtattermusch Can you help review this?

Yannic:
@acozzette ptal

acozzette:
@davido Could you remove the `java/BUILD` line from `Makefile.am`?

davido:
> @davido Could you remove the `java/BUILD` line from `Makefile.am`?

Done.

PR #7317, opened 2020-03-19
[c++] [protoc] Force to use U.S. English for Win32 error messages (#4317)

### LINKED ISSUE #4317: Unit tests seems failure on CommandLineInterfaceTest.Win32ErrorMessage
I build and install following the document on windows10, vs2017, with latest protobuf version v3.5.1, one failure "CommandLineInterfaceTest.Win32ErrorMessage" occurs when run unit tests, should i just ignore it, or how can i fix this failure? thanks in advance.

cqjjjzr:
Unit test `CommandLineInterfaceTest.Win32ErrorMessage` fails on non-English Windows due to the hardcoded English in the unit test.

This PR forces the Win32 error message produced to be in U.S. English, so the tests passes. Alternative solution can be to use messages in proper language to compare against the actual string.

Fixes #4317


cqjjjzr:
@googlebot I signed it!

PR #7320, opened 2020-03-20
add the relative path to the protoc out

peterpeterson:
When building on VS2019 and using CMake, the proto generation creates an output directory structure that matches the input side. The MSBuild is given the correct information, but protoc will put it in the current CMAKE binary path without the correct structure.


peterpeterson:
@googlebot I signed it!

PR #7328, opened 2020-03-25
Null Check in UnknownField::AddAll()

CauhxMilloy:
UnknownField::AddAll() is called (multiple times) from UnknownField::MergeFrom(). The `extras` parameter is one of: `varintList`, `fixed32List`, `fixed64List`, `lengthDelimitedList`, or `groupList`. All of these members can be null, and are appropriately checked in other usage. If attempting to parse a proto with unknown extensions, and exception is thrown (NRE).
This adds the appropriate null check inside UnknownField::AddAll().


CauhxMilloy:
@googlebot I signed it!

CauhxMilloy:
Is there anything else needed for this PR? Let me know if there's any necessary action on my part. :)

jskeet:
This looks okay at a quick look, but it would be nice to see a test that currently fails (presumably with an NRE) and that passes with the new code.

Issue #7331, opened 2020-03-27
descriptor.proto missing from protobuf-javalite


ajasmin:
Building an Android project with protobuf-gradle-plugin:0.8.12 and protobufJavaliteVersion:'3.11.4' fails if one of my proto files includes "google/protobuf/descriptor.proto".

Since the fixing of #1889, some proto files such as well-known-types are included in protobuf-javalite but descriptor.proto is still missing.

Even though the java-lite runtime doesn't make use of descriptors, it's possible to have cross-language projects that do.

ejona86:
protobuf-javalite does not include any descriptor support. The earlier missing protos had their generated code available but were lacking the .proto definitions. protobuf-javalite does not even include generated code for descriptor.proto, on purpose. So it isn't a bug that descriptor.proto is missing, but rather a feature request to include generated code for descriptor.proto.

robcos:
Is there a particular reason why the descriptor code is not there? If I remember correctly, it is there in the _nano_ libraries

elharo:
For questions or troubleshooting, please see the protobuf mailing list: https://groups.google.com/forum/#!forum/protobuf or Stack Overflow https://stackoverflow.com/questions/tagged/protocol-buffers

We don't formally monitor stack overflow, but there are a lot of experienced developers there who also use protobuf.

jhump:
@fowles, I meant to bring this up in our chat today. But it's probably minor enough that we can just discuss here in GitHub.

This issue was closed a while ago, but I can't really tell _why_. Is this something that could be added to the lite runtime? We're obviously not looking for real descriptor support (since the whole point of the lite runtime is to omit that in favor of smaller code size). But we do have cases where schemas transitively import `google/protobuf/descriptor.proto` (like an RPC response that includes `google.protobuf.FileDescriptorSet`) and then can't be used at all with the lite runtime: the generated code fails to compile because the classes for this import do not exist.

Would your team entertain the idea of re-opening this?

fowles:
@googleberg can you take a look at this with fresh eyes and make a call?

jhump:
Thanks, @fowles. It's been brought to my attention that `plugin.proto` is in the same boat. Though I can also understand that file being far less likely to appear in an API so maybe an odd thing to put in the Lite runtime. Maybe `descriptor.proto` and `plugin.proto` could be provided in a separate JAR/artifact?

[...]

PR #7332, opened 2020-03-27
Add descriptor.proto to protobuf-javalite

ajasmin:
Add descriptor.proto to javalite #7331


ajasmin:
@googlebot I signed it!

acozzette:
@ajasmin Sorry for taking to long to respond to this. I don't think we want to do this, because one of the distinguishing features of Java lite is that the generated messages don't have descriptors, so there shouldn't be any need for lite to depend on descriptor.proto.

PR #7333, opened 2020-03-27
Down integration from Google-internal development

haberman:


conformance/conformance.proto
```diff
@@ -109,7 +109,7 @@ message ConformanceRequest {
 
   // The full name for the test message to use; for the moment, either:
   // protobuf_test_messages.proto3.TestAllTypesProto3 or
-  // protobuf_test_messages.proto2.TestAllTypesProto2.
+  // protobuf_test_messages.google.protobuf.TestAllTypesProto2.
   string message_type = 4;
```
rafi-kamal: Is this WAI?

js/test_bootstrap.js
```diff
@@ -1,33 +1,3 @@
-// Protocol Buffers - Google's data interchange format
```
rafi-kamal: I think we should have the license - I'll also add it in Google3.

python/google/protobuf/json_format.py
```diff
@@ -104,7 +104,7 @@ def MessageToJson(
     sort_keys=False,
     use_integers_for_enums=False,
     descriptor_pool=None,
-    float_precision=None):
+    float_precision=8):
```
rafi-kamal: I wonder if this can break existing clients. 

CC @anandolee

src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -2055,10 +2085,11 @@ void Generator::GenerateClassConstructor(const GeneratorOptions& options,
       " * valid.\n"
       " * @extends {jspb.Message}\n"
       " * @constructor\n"
+      " * $final$\n"
       " */\n"
       "$classprefix$$classname$ = function(opt_data) {\n",
-      "classprefix", GetMessagePathPrefix(options, desc), "classname",
-      desc->name());
+      "final", final, "classprefix", GetMessagePathPrefix(options, desc),
```
rafi-kamal: I think this is a breaking change, it will make all apps jspb message final. I'm surprised to see the generator changes appearing in the open source, is this because the sync was started after cl/303150636 was submitted?

python/google/protobuf/json_format.py
anandolee: Yes, it may break some clients. It should be in release note 
In google3, this change did not break much (I do not quit remember, like 2-3 breaks)

src/google/protobuf/compiler/js/js_generator.cc
haberman: Fixed.

haberman:
Closed in favor of https://github.com/protocolbuffers/protobuf/pull/7344

PR #7338, opened 2020-03-29
Request ADLINK EdgeSDK global extension

elder4p:


elder4p:
Closing, as I'm not sure we have a corporate CLA. Trying the request via submitting a ticket, instead. If that fails, then we'll get the lawyers involved ;-)

PR #7347, opened 2020-04-01
Set execute bit on files if and only if they begin with (#!).

haberman:
Git only tracks the 'x' (executable) bit on each file. Prior to this
CL, our files were a random mix of executable and non-executable.
This change imposes some order by making files executable if and only
if they have shebang (#!) lines at the beginning.

We don't have any executable binaries checked into the repo, so
we shouldn't need to worry about that case.


RAPTOR1OUS (commented):
Protocol looks up to task.

PR #7381, opened 2020-04-15
Ruby: encode repeated field in packed format if desired

qnighy:
This patch adds `packed: true` / `packed: false` options to `repeated` descriptor declaration, and uses that information when encoding packable fields.

```ruby
    add_message "basic_test.TestMessage" do
      optional :optional_int32, :int32, 1
      optional :optional_int64, :int64, 2
      optional :optional_uint32, :uint32, 3
      optional :optional_uint64, :uint64, 4
      optional :optional_bool, :bool, 5
      optional :optional_float, :float, 6
      optional :optional_double, :double, 7
      optional :optional_string, :string, 8
      optional :optional_bytes, :bytes, 9
      optional :optional_msg, :message, 10, "basic_test.TestMessage2"
      optional :optional_enum, :enum, 11, "basic_test.TestEnum"
      repeated :repeated_int32, :int32, 12, packed: true
      repeated :repeated_int64, :int64, 13, packed: true
      repeated :repeated_uint32, :uint32, 14, packed: true
      repeated :repeated_uint64, :uint64, 15, packed: true
      repeated :repeated_bool, :bool, 16, packed: true
      repeated :repeated_float, :float, 17, packed: true
      repeated :repeated_double, :double, 18, packed: true
      repeated :repeated_string, :string, 19
      repeated :repeated_bytes, :bytes, 20
      repeated :repeated_msg, :message, 21, "basic_test.TestMessage2"
      repeated :repeated_enum, :enum, 22, "basic_test.TestEnum", packed: true
    end
```

TODO: implementation for JRuby

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -1513,6 +1513,17 @@ static void msgdef_add_field(VALUE msgbuilder_rb, upb_label_t label, VALUE name,
           field_proto,
           FileBuilderContext_strdup(self->file_builder, default_value));
     }
+
+    if (rb_funcall(options, rb_intern("key?"), 1,
+                   ID2SYM(rb_intern("packed"))) == Qtrue) {
+      VALUE packed =
+          rb_hash_lookup(options, ID2SYM(rb_intern("packed")));
+      google_protobuf_FieldOptions *options =
+        google_protobuf_FieldDescriptorProto_mutable_options(
+          field_proto,
+          file_context->arena);
+      google_protobuf_FieldOptions_set_packed(options, RTEST(packed));
+    }
   }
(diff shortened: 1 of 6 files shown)


haberman:
I'm somewhat reluctant to keep going down the path of adding more and more complexity to the DSL. Even if we fix this, we already know there are N more things that the DSL still doesn't cover.

Worse, the DSL will never be able to capture 100% of the information in a protobuf descriptor. Custom options can be self-referential, so it in some cases it will not be possible to specify these cleanly in the DSL.

If the generated code simply embedded a binary descriptor, we would have perfect fidelity and we would never have to fix the DSL again. But we'd give up the readability of the DSL. Perhaps the .proto file could be embedded in comments?

PR #7382, opened 2020-04-15
Add support for proto3 optional fields in C#

jskeet:
This is the generator part of the C# proto3 optional fields work.

Still to do:

- Reflection API changes
- Write unit tests against the (now generated) UnittestProto3Optional.cs


src/google/protobuf/compiler/csharp/csharp_generator.h
```diff
@@ -48,13 +48,16 @@ namespace csharp {
...
-      const string& parameter,
-      GeneratorContext* generator_context,
-      string* error) const;
+  CSharpGenerator();
+  ~CSharpGenerator();
+  bool Generate(
+    const FileDescriptor* file,
+    const string& parameter,
+    GeneratorContext* generator_context,
+    string* error) const;
```
haberman, line 59: Add `override` here too.

src/google/protobuf/descriptor.h
```diff
@@ -745,6 +745,9 @@ class PROTOBUF_EXPORT FieldDescriptor {
   // nullptr.
   const OneofDescriptor* containing_oneof() const;
 
+  // Is the field in a real one-of (as opposed to a synthetic oneof)?
+  bool is_in_real_oneof() const;
```
haberman, line 749: You should now be able to call `real_containing_oneof()` (this function may not have existed when you started this change). So `descriptor.h` shouldn't need to change.

src/google/protobuf/compiler/csharp/csharp_helpers.h
```diff
@@ -158,6 +158,34 @@ inline bool IsProto2(const FileDescriptor* descriptor) {
...
 }
 
+inline bool SupportsPresenceApi(const FieldDescriptor* descriptor) {
+  // Not quite the same as descriptor->is_singular_with_presence() in two ways:
+  
+  // In proto3 we add field presence only for explicitly-optional fields. Message fields
+  // don't get field presence by default, because the user can just compare with a null reference.
+  
+  // We generate Has/Clear for oneof fields in C#.
+  // It's possible that we shouldn't, but stopping doing so would be a breaking change for proto2.
```
haberman, line 168: I think it's good to have has (at least) for oneof fields. I'm going to propose that we do this across the board.

src/google/protobuf/compiler/csharp/csharp_helpers.h
```diff
@@ -158,6 +158,34 @@ inline bool IsProto2(const FileDescriptor* descriptor) {
...
+  // Fortunately the decision is moot for proto3: you can't use "optional" inside a oneof anyway.
+
+  if (IsProto2(descriptor->file())) {
+    // Proto2: every singular field has presence. (Even fields in oneofs.)
+    return !descriptor->is_repeated();
+  } else {
+    // Proto3: only for explictly-optional fields. (Repeated fields can never be explicitly optional,
+    // so we don't need to check for that.)
+    // Currently we can't get at proto3_optional directly, but we can use has_optional_keyword as a surrogate check.    
+    return descriptor->has_optional_keyword();
```
haberman, line 178: Ideally I think we should avoid changing the API based on whether the `optional` keyword is present. Is there a reason that message fields without `optional` shouldn't have `has` and `clear` methods?

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -489,8 +500,12 @@ void MessageGenerator::GenerateFrameworkMethods(io::Printer* printer) {
         generator->WriteHash(printer);
     }
     for (int i = 0; i < descriptor_->oneof_decl_count(); i++) {
-        printer->Print("hash ^= (int) $name$Case_;\n",
-            "name", UnderscoresToCamelCase(descriptor_->oneof_decl(i)->name(), false));
+      // Skip synthetic oneofs (optional proto3 fields)
```
haberman, line 503: You can now iterate up to `descriptor_->real_oneof_decl_count()`, to avoid the need for a test inside the loop. Ditto for other places in the code that are iterating over oneofs.

src/google/protobuf/descriptor.h
jskeet: Great - very happy to revert this :)

src/google/protobuf/compiler/csharp/csharp_generator.h
jskeet: Done.

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: I would like to keep this initial PR as *just* introducing optional fields for proto3 - deliberately not changing any other code, beyond the previously-inefficient presence bit allocation (which would be more significant in proto3 than in proto2). I'm happy to discuss other API changes and make those in a separate PR, but I'll be a lot more confident in *this* PR if existing proto3 code doesn't change at all.

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: Yes - it's introducing extra members for no reason. In C# clearing a message field is as simple as saying `foo.SomeField = null;` and likewise checking for presence is either `foo.SomeField != null` or `foo.SomeField is object`. Those are the idiomatic ways of checking for presence - why make the API surface significantly bigger for no benefit?

As above, I'm happy to discuss this more, but would like to make any change in a future PR.

src/google/protobuf/compiler/csharp/csharp_message.cc
jskeet: Aha - that'll make things *much* simpler, thanks. (Not yet implemented, but will do so before I next push the change, so I'll resolve now.)

jskeet:
Changes now back in - it was nice to remove the first commit and the frequent "is_synthetic" checks.

(It looks like some of the rest of the code could be tidied up by using real_containing_oneof and real_oneof_decl_count too, but I haven't tried to do that here.)

src/google/protobuf/compiler/csharp/csharp_helpers.h
haberman: How is this different from proto3 optional? Is this CL returning a default instance instead of `null` for unset proto3 optional fields?

Most languages will not have any API change when you add/remove the `optional` keyword for a message field. So I am worried that people will make this change in `.proto` files assuming that it is both API and wire compatible. It would be unfortunate if this could inadvertently break C# users by changing the API.

src/google/protobuf/compiler/csharp/csharp_helpers.h
haberman: That's totally fine, I just wanted to free you (and this comment) from the feeling of obligation that you "should" be suppressing has/clear for oneof fields.

When I said "I'm going to propose" I meant I was going to make this proposal separately, as a statement of intent for what all languages should aim for. I'm not proposing it for this review. :)

src/google/protobuf/compiler/csharp/csharp_helpers.h
```diff
@@ -158,6 +158,34 @@ inline bool IsProto2(const FileDescriptor* descriptor) {
   return descriptor->syntax() == FileDescriptor::SYNTAX_PROTO2;
 }
 
+inline bool SupportsPresenceApi(const FieldDescriptor* descriptor) {
+  // Not quite the same as descriptor->is_singular_with_presence() in two ways:
+  
+  // In proto3 we add field presence only for explicitly-optional fields. Message fields
+  // don't get field presence by default, because the user can just compare with a null reference.
```
haberman, line 165: If it's not too much trouble, it would be helpful to keep the CL (and the C# code generator generally) to 80 chars to match Google C++ style: https://google.github.io/styleguide/cppguide.html#Line_Length

Otherwise we'll have to reformat it later to pass google3 ClangTidy checks, etc.

src/google/protobuf/compiler/csharp/csharp_message.cc
```diff
@@ -72,15 +72,13 @@ MessageGenerator::MessageGenerator(const Descriptor* descriptor,
...
-  if (IsProto2(descriptor_->file())) {
-    int primitiveCount = 0;
-    for (int i = 0; i < descriptor_->field_count(); i++) {
-      const FieldDescriptor* field = descriptor_->field(i);
-      if (!IsNullable(field)) {
-        primitiveCount++;
-        if (has_bit_field_count_ == 0 || (primitiveCount % 32) == 0) {
-          has_bit_field_count_++;
-        }
+  int presenceBitCount = 0;
```
haberman, line 75: Snake case (`presence_bit_count`) for C++ style.

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
```diff
@@ -63,42 +63,60 @@ PrimitiveFieldGenerator::~PrimitiveFieldGenerator() {
...
-  // (Basically, should null-handling code be in the getter or the setter?)
+  
+  // Note: in multiple places, this code assumes that all fields
+  // that support presence are either nullable, or use a presence field bit.
+  // Fields which are oneof members are not generated here; they're generated in PrimitiveOneofFieldGenerator below.
+  // Extensions are not generated here either.
+
+
+  // Proto2 allows different default values to be specified. These are retained via static fields. They
+  // don't particularly need to be, but we don't need to change that. In Proto3 the default value we don't
```
haberman, line 74: Just wanted to verify that fixing this wouldn't be an API-breaking change?

It looks like these static variables are private, so I suspect no. But if there are any API breaks we'd want to do, it seems like we should do them now, while the proto2 API is still considered experimental.

Perhaps we should review the C# API for proto2 again before we remove its experimental status and document it.

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: Sure, happy to. (That will probably be tomorrow though.)

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: SGTM :)

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: I haven't changed anything about what the property returns. I was just trying to stick to this guideline in the implementation doc:

> For consistency with proto2, proto3 optional fields should use exactly the same API as proto2 optional.

Proto2 message fields have `HasFoo`/`Clear()` members. Personally I don't like that (for the reasons stated earlier) but I was trying to follow the guidelines. Definitely happy to make an exception in this case though.

src/google/protobuf/compiler/csharp/csharp_message.cc
jskeet: Will fix. (Not sure how it snuck through for the previous version...)

src/google/protobuf/compiler/csharp/csharp_primitive_field.cc
jskeet: I haven't thought it through in massive detail, but no, I don't *think* that it would be a breaking change at all. (It certainly wouldn't be a public API surface breaking change - it would be a matter of subtle semantic changes if anything.)

Definitely happy to review the C# API for proto2 again, yes.

haberman:
> I haven't changed anything about what the property returns. I was just trying to stick to this guideline in the implementation doc:
>
> > For consistency with proto2, proto3 optional fields should use exactly the same API as proto2 optional.
>
> Proto2 message fields have HasFoo/Clear() members. Personally I don't like that (for the reasons stated earlier) but I was trying to follow the guidelines. Definitely happy to make an exception in this case though.

Ah, I think I understand the quandry we are in then. It sounds like proto2 and proto3 are currently differing on this point? Proto2 returns a default instance for an unset message field, while proto3 returns `null`?

That means we have to choose our inconsistency. Either:

1. proto2 optional message fields are inconsistent with proto3 optional message fields
2. proto3 *explicit* optional message fields are inconsistent with proto3 *implicit* optional message fields.

I'm not sure yet what the right path is here.

jskeet:
> Proto2 returns a default instance for an unset message field, while proto3 returns null?

No, both return null - it's just that Proto2 still has `HasFoo` and `ClearFoo()` members which just check for null or set null. [Example](https://github.com/protocolbuffers/protobuf/blob/7b8a241b575b756c4dd7a5aec7f7a60854d5f711/csharp/src/Google.Protobuf.Test.TestProtos/TestMessagesProto2.cs#L927-L942)

If we're content to break existing proto2 customers, we could remove those - I don't think they add value.

At that point we could be consistent between proto2, proto3-implicit-optional and proto3-explicit-optional.

jskeet:
It sounds like the right thing to do for *this* PR is to omit presence for message fields. That's very easy to do.

jskeet:
I've added two more commits - one with source code changes addressing your comments (hopefully - I may have missed some long lines, so please let me know if there's anything there), and another with the result of regenerating for comparison.

I would expect to fold these into the existing commits later - let me know when you want me to (or if you'd rather I didn't).

haberman:
> It sounds like the right thing to do for this PR is to omit presence for message fields. That's very easy to do.

By "omit presence" do you just mean omitting the `HasFoo()` and `ClearFoo()` methods, and a hasbit?

That seems like a fine first step. :)

We would generally say that message fields are still tracking presence, since `null` is distinct from an empty sub-message. You could write `HasFoo()` and `ClearFoo()` methods that do nothing but test against `null` and set to `null`. But omitting them is fine for now too.

jskeet:
Yes, I just mean removing the `HasFoo` property and `ClearFoo()` method for a `foo` field. (We already didn't allocate a presence bit for messages, given the use of null.) The "methods that do nothing but test against null and set to null" is precisely what happens in the proto2 generation at the moment - and personally I'd love to remove those, as they make the API surface much larger for no benefit IMO.

Sounds like we're on the same page - so I suspect the next step is writing unit tests for the code generated from unittest_proto3_optional.proto, and doing whatever's required for the reflection API. At that point I believe *this* PR would be ready to merge, and then we can consider the other suggestions that have come up in the review.

haberman:
> At that point I believe this PR would be ready to merge, and then we can consider the other suggestions that have come up in the review.

Sounds great, thanks! :)

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: Done.

src/google/protobuf/compiler/csharp/csharp_helpers.h
jskeet: Done (so a proto3 optional message field does *not* have Has/Clear)

jskeet:
Okay, I believe this is now ready for review with an eye to merging.

ObsidianMinor:
What I was thinking with HasFoo and ClearFoo for nullable types was I wanted the API to stay the same for all fields no matter how their presence is stored to keep everything consistent with both proto2 non-nullable fields *and* proto3 nullable fields, since proto3 nullable fields don't accept null. I thought it may be confusing if some messages had nullable fields that took null and other messages had nullable fields that did not. Plus if consumers moved from proto2 to proto3 those field setters would silently change behavior rather than just HasFoo and ClearFoo no longer existing and consumers having to adjust accordingly.

jskeet:
> since proto3 nullable fields don't accept null

I don't know what you mean by that. The wrapper type fields are represented by nullable value types, and you can set them to null. Message type fields are represented by references, and you can set them to null.

It's only `bytes` and `string` fields that don't accept null (except for wrapper type fields).

Given that the Has/Clear members for message type fields *just* check/set null references, and that's easy to do in C# already, I'd rather not have them in either proto2 or proto3. It's already quite confusing looking through the reference documentation for a generated C# class given how many members there are - having three members for every message field, without providing any usability benefit IMO, seems like a bad idea.

I don't see what you mean about field setters silently changing behavior - could you clarify what you mean by that?

ObsidianMinor:
> It's only bytes and string fields that don't accept null (except for wrapper type fields).

Ah, it's been a while. You're right if messages don't already do null-checks then that was a waste of API space and a mistake on my part.

I believe when proto2 support was released we mentioned that all the API was experimental so I think it might be ok to remove those properties and methods.

jskeet:
> I believe when proto2 support was released we mentioned that all the API was experimental so I think it might be ok to remove those properties and methods.

Great. I've filed https://github.com/protocolbuffers/protobuf/issues/7395

I'd suggest only looking at implementing it after this is in, as then it'll just be a couple of lines of code, I believe.

csharp/src/Google.Protobuf/Reflection/SingleFieldAccessor.cs
```diff
@@ -57,10 +57,11 @@ internal SingleFieldAccessor(PropertyInfo property, FieldDescriptor descriptor)
...
             }
             setValueDelegate = ReflectionUtil.CreateActionIMessageObject(property.GetSetMethod());
-            if (descriptor.File.Syntax == Syntax.Proto3)
+            if (descriptor.File.Syntax == Syntax.Proto3 && !descriptor.IsProto3Optional)
             {
-                hasDelegate = message => {
-                  throw new InvalidOperationException("HasValue is not implemented for proto3 fields");
+                hasDelegate = message =>
+                {
+                    throw new InvalidOperationException("HasValue is not implemented for non-optional proto3 fields");
```
jtattermusch, line 64: nit: s/implemented/supported/   -  "not implemented" sounds like something that should be implemented, but isn't yet. "Not supported" sounds like this is by design.  Leaving up to you whether you want to change this or not.

src/google/protobuf/compiler/csharp/csharp_generator.h
```diff
@@ -48,13 +48,16 @@ namespace csharp {
 // header.  If you create your own protocol compiler binary and you want
 // it to support C# output, you can do so by registering an instance of this
 // CodeGenerator with the CommandLineInterface in your main() function.
-class PROTOC_EXPORT Generator : public CodeGenerator {
+class PROTOC_EXPORT CSharpGenerator : public CodeGenerator {
```
jtattermusch, line 51: just curious: why rename the class? It's under the csharp namespace already, so strictly speaking this shouldn't be necessary.

jskeet: I did it for consistency with the other classes I was looking at - namely JavaGenerator and CppGenerator. Looking at PHP, Ruby, and Python though, they just have `Generator`.

@haberman any preference on this?

csharp/src/Google.Protobuf/Reflection/SingleFieldAccessor.cs
jskeet: Hmm. I really don't have strong feelings one way or another - I tend to lean towards "keep things as they are" when I'm on the fence.

jskeet:
@haberman: Other than the question about renaming the generator class, do you have any further requested changes? Looks like the rename might have been the cause of the Kokoro build failing - I'll wait for comments about that before changing anything, as either I'll revert the rename, or update the tests.

haberman:
@jskeet Naming of the code generator class is totally up to you. :) There aren't really any compatibility implications there to worry about.

What was the final outcome of the discussion about API?

- does the API of an `optional` message field change between proto2 and proto3?
- does the API of a message field change in proto3 if you add or remove the `optional` keyword?

Those are the two issues I'm the most focused on. I want the answers to both of these to be "no." :)

It sounds like your plan is to merge this PR, then remove hazzers from message fields in proto2 per https://github.com/protocolbuffers/protobuf/issues/7395?

That seems like a fine resolution. If you are always representing non-present messages with `null` (instead of a default instance), then I can see why a separate generated `HasFoo()` method would seem superfluous.

csharp/src/Google.Protobuf/Reflection/FieldDescriptor.cs
```diff
@@ -196,6 +196,11 @@ private static FieldType GetFieldTypeFromProtoType(FieldDescriptorProto.Types.Ty
         /// </summary>
         public bool IsRequired => Proto.Label == FieldDescriptorProto.Types.Label.Required;
 
+        /// <summary>
+        /// Returns <c>true</c> if this field is a proto3 optional field; <c>false</c> otherwise.
+        /// </summary>
+        public bool IsProto3Optional => Proto.Proto3Optional;
```
haberman, line 202: I don't want to expose this publicly. I think we paint ourselves into a corner if code is written to think about the difference between proto2 and proto3.

Could you follow the pattern of the C++ API? C++ is hiding this accessor, and instead exposing a more semantically meaningful accessor: `has_presence()`.

jskeet:
> @jskeet Naming of the code generator class is totally up to you. :) There aren't really any compatibility implications there to worry about.

In that case I'll revert that part of the change then. Given that there's no *need* for the class name to change, and it's as negative as it's positive, let's avoid the churn :)

> does the API of an optional message field change between proto2 and proto3?

Yes, at the moment.

> does the API of a message field change in proto3 if you add or remove the optional keyword?

No.

> It sounds like your plan is to merge this PR, then remove hazzers from message fields in proto2 per #7395?

Exactly - assuming we're willing to make a breaking change to the proto2 generated code.

csharp/src/Google.Protobuf/Reflection/IFieldAccessor.cs
```diff
@@ -59,7 +59,8 @@ public interface IFieldAccessor
         object GetValue(IMessage message);
 
         /// <summary>
-        /// Indicates whether the field in the specified message is set. For proto3 fields, this throws an <see cref="InvalidOperationException"/>
+        /// Indicates whether the field in the specified message is set.
+        /// For proto3 fields that aren't explicitly optional, this throws an <see cref="InvalidOperationException"/>
```
haberman, line 63: I would expect this to work for oneof fields also?

haberman:
> I'm happy to remove it from the public API for now, but I'll need to check what exact semantics we want for HasPresence.

Yep that is fair. I think the best semantics for `HasPresence` are: `Reflection.HasField` returns a semantically meaningful result.

> It could do, but it didn't before this PR - I'd prefer to treat that as a separate feature to implement. Likewise we can do it for message fields whether or not they're explicitly optional.

Yep, agree on both counts (message fields should always work here, and it seems find to implement this separately in a follow-up PR).

> It's probably worth mentioning this as a requirement in the internal implementation guide as well.

Good point, I will add this to: https://github.com/protocolbuffers/protobuf/blob/9ae5203712eb80f89261d6df8d5674efa5a0edb8/docs/implementing_proto3_presence.md

jskeet:
@haberman: Please see the last commit for reflection changes. I'm pretty comfortable with those. Just to confirm, I'm then expecting to perform follow-up PRs of:

- Breaking change for proto2 generation: don't generate Has/Clear for message fields
- Add HasPresence for FieldDescriptor, and make it work wherever feasible (everywhere for proto2; oneofs + message + optional for proto3)

Does that all sound correct to you?

jskeet:
Thanks - will fix up the commits to be a bit more pleasant, and investigate the Kokoro build failures, then merge on green.

haberman:
> Does that all sound correct to you?

Yep, sounds great. :)

PR #7406, opened 2020-04-21
Implemented proto3 presence for Ruby.

haberman:


ruby/ext/google/protobuf_c/defs.c
```diff
@@ -1758,6 +1795,56 @@ VALUE MessageBuilderContext_oneof(VALUE _self, VALUE name) {
...
+  for (i = 0; i < field_count; i++) {
+    google_protobuf_OneofDescriptorProto* oneof_proto;
+    VALUE oneof_name;
+    upb_strview field_name;
+
+    if (!google_protobuf_FieldDescriptorProto_proto3_optional(fields[i])) {
+      continue;
+    }
+
+    // Prepend '_' until we are no longer conflicting.
```
TeBoring, line 1831: Will this renaming suddenly change when users reorder fields in their proto definition?

ruby/ext/google/protobuf_c/message.c
```diff
@@ -242,9 +242,14 @@ static int extract_method_call(VALUE method_name, MessageHeader* self,
   // Method calls like 'has_foo?' are not allowed if field "foo" does not have
   // a hasbit (e.g. repeated fields or non-message type fields for proto3
   // syntax).
-  if (accessor_type == METHOD_PRESENCE && test_f != NULL &&
-      !upb_fielddef_haspresence(test_f)) {
-    return METHOD_UNKNOWN;
+  if (accessor_type == METHOD_PRESENCE && test_f != NULL) {
+    if (!upb_fielddef_haspresence(test_f)) return METHOD_UNKNOWN;
+
+    // TODO(haberman): remove this case, allow for proto3 oneofs.
```
TeBoring, line 248: Do we previously support this?

ruby/ext/google/protobuf_c/storage.c
```diff
@@ -725,20 +729,20 @@ static void slot_clear_hasbit(MessageLayout* layout,
 static bool slot_is_hasbit_set(MessageLayout* layout,
                             const void* storage,
                             const upb_fielddef* field) {
-  assert(field_contains_hasbit(layout, field));
   size_t hasbit = layout->fields[upb_fielddef_index(field)].hasbit;
+  assert(field_contains_hasbit(layout, field));
```
TeBoring, line 733: Do you need to move this assert?

ruby/ext/google/protobuf_c/defs.c
haberman: If there are conflicts it could. But the synthetic oneof name was not created by the user, so the user cannot assume what its name is.

ruby/ext/google/protobuf_c/message.c
haberman: No, we have never supported this. The unit tests were failing until I added this check, see:

https://github.com/protocolbuffers/protobuf/blob/18950451c7bfb1497f1deaf1c3868e79e332d80a/ruby/tests/basic.rb#L45-L50

ruby/ext/google/protobuf_c/storage.c
haberman: Yes, because it is a statement and cannot go before the declaration of `hasbit`.

This worked previously because we had not been testing with asserts enabled.

haberman:
Re: synthetic oneof (and for more background information in general), see this doc which is currently in review: https://github.com/protocolbuffers/protobuf/blob/9ae5203712eb80f89261d6df8d5674efa5a0edb8/docs/implementing_proto3_presence.md

TeBoring:
Did ruby 2.6 fail on mac?

haberman:
@TeBoring fixed the crash that was happening on mac.

Issue #5530, opened 2019-01-03
Carthage support


jozefVrana:
Is there any plan for Carhage support in near future?
Is there some issues with this, or is it straightforward?

PR #7417, opened 2020-04-24
Add support for carthage (iOS build dependency manager)

mdyson:
This PR is meant to add support for carthage when building in iOS (from issue #5530).

For some background, the main idea of how carthage works is that it will search the root for a xcprojfile or xcworkspace file, verify it has a buildable candidate and run through the build script. 

In this case here, since there is no project file associated with this I used the Protobuf.podspec to generate a xcworkspace (and associated xcprojfiles). This is so I didn't have to do any guess work in setup and keep it 1:1 with the podspec file but it does add a lot more files than needed. 

These are the steps I took to get here: (on a mac with Xcode installed)

1.) install both cocoapods and cocoapods-generate
gem 'cocoapods'
gem 'cocoapods-generate'
2.) install carthage (not needed for setup but used just for verifying this all works)
3.) Run the following commands:
`pod gen Protobuf.podspec --local-sources=./ --platforms=ios`
`ln -s gen/Protobuf/Protobuf.xcworkspace Protobuf.xcworkspace` # this is a requirement for carthage, it needs to find an xcproj or xcworkspace in the root
4.) Done! Verify it works by running `carthage build --archive` this will make a Carthage directory that should not be checked in as well as a zip file with the build products

Once again, this might be more complicated in this way rather than simply making a xcprojfile and adding the source code there, this just removes all that work by having cocoapods do it for you. I'm open to suggestions and other ideas.


thomasvl:
> In this case here, since there is no project file associated with this I used the Protobuf.podspec to generate a xcworkspace (and associated xcprojfiles). This is so I didn't have to do any guess work in setup and keep it 1:1 with the podspec file but it does add a lot more files than needed.

There are Xcode projects already in the _objectivec/_ sub directory.

mdyson:
Thanks for pointing that out, I missed that there are xcprojfiles already existing in here. That will make the solution for this a bit simpler, in either case we need a reference to the project in the root directory. I will come back soon with a better solution to address this.

PR #7418, opened 2020-04-24
Suppress gcc override complaint in generated code

poelmanc:
Unfortunately gcc and clang at high warning levels have incompatible requirements about "override" and "final". Gcc -Wsuggest-override issues warnings on members that have only "final", demanding to also see "override". Clang at high warning levels demands at most one of "virtual", "override", and "final". PROTOBUF_FINAL expands to only "final", so compiling protoc-generated files with gcc at high warning levels results in numerous warnings (or warnings-as-errors): "virtual Something* Something::New() const’ can be marked override [-Werror=suggest-override]".

This change follows the example in FileGenerator::GenerateInlineFunctionDefinitions() for suppressing spurious strict-aliasing warnings by adding a #pragma GCC push/pop within which the suggest-override warning is suppressed.


acozzette:

PR #7424, opened 2020-04-24
Add support for carthage (iOS build dependency manager)

mdyson:
Redo of PR #7417 

This PR is meant to add support for carthage when building in iOS (from issue #5530).

For some background, the main idea of how carthage works is that it will search the root for a xcprojfile or xcworkspace file, verify it has a buildable candidate and run through the build script.

Carthage also requires that the exposed build targets are frameworks and not static libraries. In this change I've added new targets that mirror over the static targets and added a xcworkspace file to the root of this repo that points to the xcworkspace files. This allows carthage to find the files and generate the frameworks as expected.

This can be tested by running:
`carthage build --archive` 
from either the root of this repo of in the objectivec directory, another test can be to make an empty project with a Cartfile, adding this repo and running `carthage bootstrap`

### DIFF
--- a/objectivec/Protobuf/Protobuf.h
+++ b/objectivec/Protobuf/Protobuf.h
@@ -1,0 +1,42 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// This header is meant to only be used by the generated source, it should not
+// be included in code using protocol buffers.
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for ProtocolBuffers_dynamic.
+FOUNDATION_EXPORT double ProtocolBuffers_dynamicVersionNumber;
+
+//! Project version string for ProtocolBuffers_dynamic.
+FOUNDATION_EXPORT const unsigned char ProtocolBuffers_dynamicVersionString[];
+
+#import <Protobuf/GPBProtocolBuffers.h>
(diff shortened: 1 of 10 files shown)


objectivec/ProtocolBuffers_iOS.xcodeproj/project.pbxproj
```diff
@@ -1090,6 +1281,95 @@
 			};
 			name = Release;
 		};
+		FF1C791D245340E40099AB0D /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
```
thomasvl, line 1286: Please remove the local settings from the targets that aren't explicitly needed.  The project level should be setting most of the warning related flags, we don't want different warnings between the different targets.

objectivec/Protobuf/Protobuf.h
```diff
@@ -0,0 +1,42 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// This header is meant to only be used by the generated source, it should not
+// be included in code using protocol buffers.
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for ProtocolBuffers_dynamic.
+FOUNDATION_EXPORT double ProtocolBuffers_dynamicVersionNumber;
```
thomasvl, line 37: What is this?  What's the "dynamic" part?  All version info should likely not be apple's versioning tool but instead be tied to the versioning of the protobuf library.

mdyson: Dynamic part is removed, that was a mistake in the naming when I first put this together. This header file is needed for framework module support. If module support is removed this file can get removed. The version numbers in here are derived from the Info.plist file which is mandatory in a framework.

objectivec/ProtocolBuffers_iOS.xcodeproj/project.pbxproj
mdyson: Removed all unnecessary local settings from here, only ones left should be the needed ones for the framework to build correctly.

mdyson:
Changes are made to the two code comments made.

As long as this builds along with all the standard scripts that should be enough guarantee that this wont accidentally break in the future.

I will work on addressing your higher level comments of adding this to the Makefile and full_mac_build script to address that concern. I will also double check through the other project file to make sure I didn't miss any other stray build settings.

objectivec/ProtocolBuffers_iOS.xcodeproj/project.pbxproj
thomasvl: GCC_PREPROCESSOR_DEFINITIONS repeats `DEBUG`, probably should be $(inherited) and the framework imports one. GCC_C_LANGUAGE_STANDARD should be the same for all targets.  Is CLANG_ENABLE_MODULE really needed?

objectivec/Protobuf/Protobuf.h
thomasvl: What actually creates the symbols?  Looks like apple's versioning might be enabled?  Shouldn't it somehow be wired up to the releases of this library instead?  (git tags or something)?

thomasvl:
btw - does the project file/framework name have to be the same as the repo name?  Or would ProtocolBuffers be a better name to use for this?

objectivec/ProtocolBuffers_iOS.xcodeproj/project.pbxproj
mdyson: Fixed! I found a few other things I missed in the tvOS and MacOS projfiles as well.
Short answer for CLANG_ENABLE_MODULE, its not needed. TLDR: It's used in frameworks for compiler optimization to make importing cheap but also extensively import all headers in one shot.

Here is some more information on what it does: https://clang.llvm.org/docs/Modules.html#id10

When the project is built with Cocoapods this feature is supported, it's a standard now at this point, I'd also argue that it's required for swift files to be able to import this but I think in most cases the apple/swift-protobuf library is the better option.

objectivec/Protobuf/Protobuf.h
mdyson: I need to double check to make sure this is right but I believe it come from here: 
https://github.com/mdyson/protobuf/blob/master/objectivec/Protobuf/Info.plist#L20
Which links back to the projfile here: https://github.com/mdyson/protobuf/blob/master/objectivec/ProtocolBuffers_iOS.xcodeproj/project.pbxproj#L1289

You are right, this should be linked up to the library release versions, I will add this in an upcoming commit.

mdyson:
Nope, the naming can be different. I can make the actual framework name to be ProtocolBuffers instead. I think that might break consistency with the Cocoapods project though since the built framework from there is called Protobuf. 
You can see from this example of how this library get configured from: https://github.com/googlecast/CastVideos-ios

objectivec/Protobuf/Protobuf.h
thomasvl: These are decl for exported symbols, so it won't link back to the info.plist unless some code loads it and sets the globals to the same value.

thomasvl:
The name can stay Protobuf, I was just wondering if the existing target(s) in the project files might work, but I forgot it is a static lib, so that probably doesn't work if you really need a framework.

mdyson:
I've made the changes to the Makefile.am but I am a bit unsure of what the best way to proceed with updating the full_mac_build script is. 

Would it be better to add something like as a flag `--dynamic-framework` which will build using the dynamic scheme? Or should it always build and test both? I think the latter might be better for a more comprehensive run of the build script.

thomasvl:
> Would it be better to add something like as a flag `--dynamic-framework` which will build using the dynamic scheme? Or should it always build and test both? I think the latter might be better for a more comprehensive run of the build script.

Best bet is probably to build both, so maybe just add a `--skip-dynamic-framework` and `--skip-static-framework` so either scheme can be turned off for an invocation.  (and fail if someone passes both)

objectivec/Protobuf/Protobuf.h
thomasvl: Did you get a chance to sort out where these symbols come from?

acozzette:
@mdyson I'm going to tentatively close this pull request since there has been no activity in a while, but feel free to reopen it if you would like to keep working on this.

PR #7428, opened 2020-04-27
Fix when compiling without RTTI

mwerle:
no-rtti build is broken due to forced use of typeid() introduced in commit b99994d994e399174fe688a5efbcb6d91f36952a

### DIFF
--- a/src/google/protobuf/util/field_mask_util.h
+++ b/src/google/protobuf/util/field_mask_util.h
@@ -46,12 +46,6 @@ namespace google {
 namespace protobuf {
 namespace util {
 
-#if GTEST_HAS_RTTI
-#define TYPENAME(T) typeid(T).name()
-#else
-#define TYPENAME(T) #T
-#endif
-
 class PROTOBUF_EXPORT FieldMaskUtil {
   typedef google::protobuf::FieldMask FieldMask;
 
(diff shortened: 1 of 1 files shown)


mwerle:
@googlebot I signed it!

src/google/protobuf/util/field_mask_util.h
```diff
@@ -46,6 +46,12 @@ namespace google {
 namespace protobuf {
 namespace util {
 
+#if GTEST_HAS_RTTI
```
acozzette, line 49: Could you do `#if PROTOBUF_RTTI` instead? I assume the gtest macro will work in tests but not for real production binaries. Could you also make sure to undefine the macro after it's used, just to make sure it can't conflict with other macros? It might also be a good idea to call it `PROTOBUF_TYPENAME()` just to be safe.

mwerle: Thank you for your comments.

I've cleaned up the patch as per your suggestions.

acozzette:
Thanks, @mwerle.

haberman:
I think we should be able to just use the name from the descriptor, instead of using RTTI.

I'll send another pull request for this.

haberman:
See: https://github.com/protocolbuffers/protobuf/pull/7465

PR #7429, opened 2020-04-27
Remove Has/Clear members for C# message fields in proto2

jskeet:
cc @ObsidianMinor

### DIFF
--- a/csharp/src/Google.Protobuf.Test.TestProtos/TestMessagesProto2.cs
+++ b/csharp/src/Google.Protobuf.Test.TestProtos/TestMessagesProto2.cs
@@ -290,13 +290,13 @@ namespace ProtobufTestMessages.Proto2 {
       optionalBool_ = other.optionalBool_;
       optionalString_ = other.optionalString_;
       optionalBytes_ = other.optionalBytes_;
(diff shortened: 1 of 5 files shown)


jskeet:
@haberman Ping? Just checking you've seen this...

haberman:
Sorry for the slow reply. I think this is fine from a correctness and API design perspective. My only possible worry is around the sorts of concerns you raised on https://github.com/protocolbuffers/protobuf/pull/7434#issuecomment-620590606. It's true though that you only get the change when you regenerate your code.

jskeet:
> My only possible worry is around the sorts of concerns you raised on #7434 (comment). It's true though that you only get the change when you regenerate your code.

Definitely understand the concern. I'm less worried here though, for a few reasons:

- You're very unlikely (although it's not impossible) to end up "accidentally" upgrading your version of protoc due to some other dependency, whereas transitive dependencies mean it's quite likely that the version of the support library you end up with may not be the one you specify directly.
- Any problems are found at build time rather than execution time
- The proto2 generation has previously been labeled as experimental, whereas option fetching wasn't. Even though they were done in the same set of changes, I think there's a higher compatibility bar for anything in the support library.

If you're happy with those justifications, I'll merge. If not, I'm definitely happy to chat about it more and see whether there are alternative options (although I can't think of any right now).

cc @jtattermusch just for the sake of interest in compatibility.

PR #7430, opened 2020-04-27
Update shell format in installation instructions.

nickpresta:
The Markdown formatting for these installation instructions is (currently) broken, specifically in the section for Mac users.

Since the entire file uses indentation, instead of fenced code blocks, I updated the other two references to be consistent (and remove extra indentation from the code snippet).


acozzette:
@nickpresta Thank you for the fix. This looks great but I'm just trying to figure out if everything is good to go with the CLA. There is a CLA bot which is supposed to verify it but I'm not sure why it's not working. Let me just try to sort this out and then I'll merge this PR.

acozzette:
@googlebot

Issue #7043, opened 2019-12-21
C#: CustomOptions broken and deprecated without working alternatives


ptr1120:
**What version of protobuf and what language are you using?**
Version: v3.11.2
Language: C#

**What operating system (Linux, Windows, ...) and version?**
Win

**What runtime / compiler are you using (e.g., python version or gcc version)**
Generated C# classes from proto files using Grpc.Tools v2.26.0

**What did you do?**

ServiceName.proto
```
syntax = "proto3";
import "google/protobuf/descriptor.proto";

extend google.protobuf.MessageOptions {
    string service_name = 50001; 
}
```
SumRequestDto.proto
```
syntax = "proto3";
import "ServiceName.proto";

message SumRequestDto {
   option (service_name) = "Producer";
}
```
Program.cs
```
nternal class Program
    {
        private static void Main(string[] args)
        {
            SumRequestDto.Descriptor.CustomOptions.TryGetString(50001, out var messageName);
        }
    }
```

Until Google.Protobuf 3.8.0 everything worked fine with the above code. 

I also tried the suggestion to use  `SumRequestDtoV1Reflection.Descriptor.GetOption()` which requires a `Extension<FileOptions, T> extension` argument.

My Questions:
- How can I generate these required Extensions in order the get my custom option value again?
- Will there ever be a working version with custom options?

**What did you expect to see**
The variable `messageName` assigned to  "Producer".

**What did you see instead?**
The variable `messageName` assigned to null.

ObsidianMinor:
Options currently: 

* Downgrade back to 3.8.0 since protoc in `Grpc.Tools` v2.26.0 is 3.8.0
* Wait for #7005
* Generate code manually without Grpc.Tools until Grpc.Tools updates to 3.11

zeinali0:
I have same issue too.
And the problem is if we downgrade to 3.8.0 then we should downgrade Grpc.AspNetCore to 0.1.22-pre3 which broke many things

jtattermusch:
@ObsidianMinor    Grpc.Tools 2.28.x are already using protobuf 3.11.  Does that mean the problem is solved now?

ObsidianMinor:
Should be solved, yes.

PR #7434, opened 2020-04-28
C#: Get rid of broken GetOption API and expose the "GetOptions()" method on descriptors instead

jtattermusch:
The currently existing new API for getting custom options (introduced in https://github.com/protocolbuffers/protobuf/pull/5350 as part of proto2 extensions support) has some serious issues
- it is incomplete: it doesn't allow distinguishing between the custom option not being present at all VS being present and having the default value.
- the GetOption API basically just mimics the existing GetExtension API (with repeated / non-repeated overloads), creating unnecessary duplication (which then leads to problems such as https://github.com/protocolbuffers/protobuf/issues/7127)
- the new API offers no way of accessing the "standard" (non-custom) options at all.
- the GetOption API is not consistent with what other protobuf implementations (C++, java, python) do to access custom options.

This PR removes the unnecessary (and partially broken) GetOption API  and replaces it by a more straightforward approach:
- expose Options property for each descriptor that allows one to access the corresponding "options" message (but use a defensive copy because C# messages are inherently mutable).
- the custom options are retrieved just as any other extensions, using the standard extension API (HasExtension, GetExtension). This is the same approach that is used in C++, Java and Python

(the GetOption API has been marked as "experimental and subject to change" when releasing the C# proto2 support, so removing it if fine to do).

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
@@ -128,24 +128,14 @@ namespace Google.Protobuf.Reflection
         /// <summary>
         /// The (possibly empty) set of custom options for this enum.
         /// </summary>
-        [Obsolete("CustomOptions are obsolete. Use GetOption")]
+        [Obsolete("CustomOptions are obsolete. Use the Options property.")]
         public CustomOptions CustomOptions => new CustomOptions(Proto.Options?._extensions?.ValuesByNumber);
 
         /// <summary>
-        /// Gets a single value enum option for this descriptor
+        /// The <c>EnumOptions</c>, defined in <c>descriptor.proto</c>.
+        /// Custom options can be retrieved as extensions of the returned message.
+        /// NOTE: A defensive copy is created each time this property is retrieved.
         /// </summary>
-        public T GetOption<T>(Extension<EnumOptions, T> extension)
-        {
-            var value = Proto.Options.GetExtension(extension);
-            return value is IDeepCloneable<T> ? (value as IDeepCloneable<T>).Clone() : value;
-        }
-
-        /// <summary>
-        /// Gets a repeated value enum option for this descriptor
-        /// </summary>
-        public RepeatedField<T> GetOption<T>(RepeatedExtension<EnumOptions, T> extension)
-        {
-            return Proto.Options.GetExtension(extension).Clone();
-        }
+        public EnumOptions Options => (Proto.Options as IDeepCloneable<EnumOptions>)?.Clone();
     }
 }
--- a/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
@@ -133,6 +133,7 @@ namespace Google.Protobuf.Reflection
 
         /// <summary>
         /// The <c>EnumOptions</c>, defined in <c>descriptor.proto</c>.
+        /// If the options message is not present (=there are no options), <c>null</c> is returned.
         /// Custom options can be retrieved as extensions of the returned message.
         /// NOTE: A defensive copy is created each time this property is retrieved.
         /// </summary>
--- a/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
@@ -137,5 +137,24 @@ namespace Google.Protobuf.Reflection
         /// NOTE: A defensive copy is created each time this property is retrieved.
         /// </summary>
         public EnumOptions Options => (Proto.Options as IDeepCloneable<EnumOptions>)?.Clone();
+
+        /// <summary>
+        /// Gets a single value enum option for this descriptor
+        /// </summary>
+        [Obsolete("GetOption is obsolete. Use the Options property.")]
+        public T GetOption<T>(Extension<EnumOptions, T> extension)
+        {
+            var value = Proto.Options.GetExtension(extension);
+            return value is IDeepCloneable<T> ? (value as IDeepCloneable<T>).Clone() : value;
+        }
+
+        /// <summary>
+        /// Gets a repeated value enum option for this descriptor
+        /// </summary>
+        [Obsolete("GetOption is obsolete. Use the Options property.")]
+        public RepeatedField<T> GetOption<T>(RepeatedExtension<EnumOptions, T> extension)
+        {
+            return Proto.Options.GetExtension(extension).Clone();
+        }
     }
 }
--- a/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
@@ -128,21 +128,21 @@ namespace Google.Protobuf.Reflection
         /// <summary>
         /// The (possibly empty) set of custom options for this enum.
         /// </summary>
-        [Obsolete("CustomOptions are obsolete. Use the Options property.")]
+        [Obsolete("CustomOptions are obsolete. Use the GetOptions() method.")]
         public CustomOptions CustomOptions => new CustomOptions(Proto.Options?._extensions?.ValuesByNumber);
 
         /// <summary>
         /// The <c>EnumOptions</c>, defined in <c>descriptor.proto</c>.
-        /// If the options message is not present (=there are no options), <c>null</c> is returned.
+        /// If the options message is not present (i.e. there are no options), <c>null</c> is returned.
         /// Custom options can be retrieved as extensions of the returned message.
         /// NOTE: A defensive copy is created each time this property is retrieved.
         /// </summary>
-        public EnumOptions Options => Proto.Options?.Clone();
+        public EnumOptions GetOptions() => Proto.Options?.Clone();
 
         /// <summary>
         /// Gets a single value enum option for this descriptor
         /// </summary>
-        [Obsolete("GetOption is obsolete. Use the Options property.")]
+        [Obsolete("GetOption is obsolete. Use the GetOptions() method.")]
         public T GetOption<T>(Extension<EnumOptions, T> extension)
         {
             var value = Proto.Options.GetExtension(extension);
@@ -152,7 +152,7 @@ namespace Google.Protobuf.Reflection
         /// <summary>
         /// Gets a repeated value enum option for this descriptor
         /// </summary>
-        [Obsolete("GetOption is obsolete. Use the Options property.")]
+        [Obsolete("GetOption is obsolete. Use the GetOptions() method.")]
         public RepeatedField<T> GetOption<T>(RepeatedExtension<EnumOptions, T> extension)
         {
             return Proto.Options.GetExtension(extension).Clone();
--- a/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
@@ -136,7 +136,7 @@ namespace Google.Protobuf.Reflection
         /// Custom options can be retrieved as extensions of the returned message.
         /// NOTE: A defensive copy is created each time this property is retrieved.
         /// </summary>
-        public EnumOptions Options => (Proto.Options as IDeepCloneable<EnumOptions>)?.Clone();
+        public EnumOptions Options => Proto.Options?.Clone();
 
         /// <summary>
         /// Gets a single value enum option for this descriptor
--- a/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Reflection/CustomOptionsTest.cs
@@ -100,12 +100,12 @@ namespace Google.Protobuf.Test.Reflection
         {
             var d = CustomOptionOtherValues.Descriptor;
             var customOptions = d.CustomOptions;
-            AssertOption(-100, customOptions.TryGetInt32, Int32Opt, d.Options.GetExtension);
-            AssertOption(12.3456789f, customOptions.TryGetFloat, FloatOpt, d.Options.GetExtension);
-            AssertOption(1.234567890123456789d, customOptions.TryGetDouble, DoubleOpt, d.Options.GetExtension);
-            AssertOption("Hello, \"World\"", customOptions.TryGetString, StringOpt, d.Options.GetExtension);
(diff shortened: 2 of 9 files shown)


jtattermusch:
CC @ObsidianMinor  @JamesNK

jtattermusch:
Alternative approaches are in PRs https://github.com/protocolbuffers/protobuf/pull/7130/files 
https://github.com/protocolbuffers/protobuf/pull/7282/files  (this PR tries to find a simple solution to the problems encountered when reviewing those PRs).

jskeet:
I'm fine with deprecating the existing API, but we shouldn't make breaking changes by removing the existing one unless you're planning on creating a major version bump (which I'd strongly discourage).

I'm uncomfortable with the idea of APIs being labelled as "experimental" in general - particularly when that designation is only visible if you happen to read the right documentation.

I'm similarly nervous about my own #7429 PR, but at least that's in the *generated code* rather than the support library. (You only get broken when you regenerate, and then it's at compile time.) This change could easily mean that a minor version bump in which support library is consumed ends up breaking a deployed application.

jtattermusch:
> I'm fine with deprecating the existing API, but we shouldn't make breaking changes by removing the existing one unless you're planning on creating a major version bump (which I'd strongly discourage).
> 
> I'm uncomfortable with the idea of APIs being labelled as "experimental" in general - particularly when that designation is only visible if you happen to read the right documentation.
> 
> I'm similarly nervous about my own #7429 PR, but at least that's in the _generated code_ rather than the support library. (You only get broken when you regenerate, and then it's at compile time.) This change could easily mean that a minor version bump in which support library is consumed ends up breaking a deployed application.

I'm fine with both approaches (removing the GetOption method or only marking it as deprecated).
Some arguments why just removing GetOption API might be fine:
- The API only existed in 3.11.0 release and higher (and the whole proto2 support was marked as experimental in the release notes: https://github.com/protocolbuffers/protobuf/releases/tag/v3.11.0).
- Proto2 support is still pretty new so likely it doesn't have many users and those who tried the new GetOption API quickly found that it's broken can only be used with workarounds (so changing/fixing the API doesn't seem very surprising). See e.g. https://github.com/protocolbuffers/protobuf/issues/7043 FTR

jskeet:
If the option had only been available for Proto2-based code, I'd be at least *more* comfortable - but we've been using it (in API generation) for proto3-based options. Again, I don't think marking something as experimental just in release notes is a good way of excusing breaking changes. If we want to be able to make breaking changes in the support library itself, I really think we need a better way of doing so. (Marking it as deprecated would, somewhat oddly, be a reasonable starting point - *something* that brings it to the attention of users.)

And yes, it needed fixing to start with - but it *was* fixed. It's not great in terms of API surface, but it works for at least some scenarios (e.g. how we're using it in https://github.com/googleapis/gapic-generator-csharp)

I tend to think that the benefits of violating semver have to be really, really huge to counter the downsides, which can include lack of trust in future releases. Sometimes it's the best-of-a-bad-bunch option, but in this case I think using deprecation is better.

jtattermusch:
> If the option had only been available for Proto2-based code, I'd be at least _more_ comfortable - but we've been using it (in API generation) for proto3-based options. Again, I don't think marking something as experimental just in release notes is a good way of excusing breaking changes. If we want to be able to make breaking changes in the support library itself, I really think we need a better way of doing so. (Marking it as deprecated would, somewhat oddly, be a reasonable starting point - _something_ that brings it to the attention of users.)
> 
> And yes, it needed fixing to start with - but it _was_ fixed. It's not great in terms of API surface, but it works for at least some scenarios (e.g. how we're using it in https://github.com/googleapis/gapic-generator-csharp)
> 
> I tend to think that the benefits of violating semver have to be really, really huge to counter the downsides, which can include lack of trust in future releases. Sometimes it's the best-of-a-bad-bunch option, but in this case I think using deprecation is better.

Fair enough, I'll add back the GetOption methods and mark them as obsolete.

jskeet:
Fab, thanks. I can then have a look at the "new" API without being distracted ;)

jtattermusch:
> > If the option had only been available for Proto2-based code, I'd be at least _more_ comfortable - but we've been using it (in API generation) for proto3-based options. Again, I don't think marking something as experimental just in release notes is a good way of excusing breaking changes. If we want to be able to make breaking changes in the support library itself, I really think we need a better way of doing so. (Marking it as deprecated would, somewhat oddly, be a reasonable starting point - _something_ that brings it to the attention of users.)
> > And yes, it needed fixing to start with - but it _was_ fixed. It's not great in terms of API surface, but it works for at least some scenarios (e.g. how we're using it in https://github.com/googleapis/gapic-generator-csharp)
> > I tend to think that the benefits of violating semver have to be really, really huge to counter the downsides, which can include lack of trust in future releases. Sometimes it's the best-of-a-bad-bunch option, but in this case I think using deprecation is better.
> 
> Fair enough, I'll add back the GetOption methods and mark them as obsolete.

Done.

haberman:
Note that custom options can be self-referential. Here is valid .proto file that we should ensure works properly in C#:

```protobuf
syntax = "proto2";

message FooOptions {
  // Custom field option used in definition of the custom option's message.
  optional int32 foo = 1 [(foo_options) = {foo: 1234}];
}

  // Custom field option used on the definition of that field option.
  optional int32 bar_options = 1000 [(bar_options) = 1234];

  optional FooOptions foo_options = 1001;
}
```

ObsidianMinor:
I'm still unsure how I feel about doing defensive copies but that may just be the Rust developer in me that's worried and not the C# developer. It will make the API simpler in the long run and will be easier to maintain so it should be ok.

jtattermusch:
I discussed the defensive copy with @haberman and we agreed that at least at the first glance it seemed to be a reasonable solution. I think @haberman  is the ultimate authority here so I'm gonna go with his opinion.

jskeet:
Yes, I'm pretty sure we need the defensive copy. It would be awful if reflection could modify the descriptors globally.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -128,12 +128,20 @@ public EnumValueDescriptor FindValueByName(string name)
...
-        [Obsolete("CustomOptions are obsolete. Use GetOption")]
+        [Obsolete("CustomOptions are obsolete. Use the Options property.")]
         public CustomOptions CustomOptions => new CustomOptions(Proto.Options?._extensions?.ValuesByNumber);
 
+        /// <summary>
+        /// The <c>EnumOptions</c>, defined in <c>descriptor.proto</c>.
+        /// Custom options can be retrieved as extensions of the returned message.
+        /// NOTE: A defensive copy is created each time this property is retrieved.
+        /// </summary>
+        public EnumOptions Options => (Proto.Options as IDeepCloneable<EnumOptions>)?.Clone();
```
JamesNK, line 139: Can `Proto.Options` ever not be `IDeepCloneable<EnumOptions>`? Same question for all the other `as` casts. I prefer to do a regular cast unless there is a reason.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -128,12 +128,20 @@ public EnumValueDescriptor FindValueByName(string name)
         /// <summary>
         /// The (possibly empty) set of custom options for this enum.
         /// </summary>
-        [Obsolete("CustomOptions are obsolete. Use GetOption")]
+        [Obsolete("CustomOptions are obsolete. Use the Options property.")]
         public CustomOptions CustomOptions => new CustomOptions(Proto.Options?._extensions?.ValuesByNumber);
 
+        /// <summary>
```
JamesNK, line 134: Document in what scenario that null will be returned?

haberman:
Yes I agree that it would be better to avoid the copy if possible. But without `const` message instances like we have in C++ I don't see a way to do it.

haberman:
By the way, here's an even more diabolical example of custom options.

```protobuf
syntax = "proto2";

message FooOptions {
  // Custom field option used in definition of the extension message.
  optional int32 int_opt = 1 [(foo_options) = {
             int_opt: 1
             [foo_int_opt]: 2
             [foo_foo_opt]: {
               int_opt: 3
             }
  }];
  extensions 1000 to max;
}

  // Custom field option used on the definition of that field option.
  optional int32 bar_options = 1000 [(bar_options) = 1234];

  optional FooOptions foo_options = 1001;
}

extend FooOptions {
  optional int32 foo_int_opt = 1000;
  optional FooOptions foo_foo_opt = 1001;
}
```

JamesNK:
So to use this:

```cs
var httpRule = methodDescriptor.Options.GetExtension(AnnotationsExtensions.Http);
```

If `HttpRule` is not present on the method will it return null or throw an error? If it errors then I'm guessing this will work:

```cs
var httpRule = methodDescriptor.Options.HasExtension(AnnotationsExtensions.Http) ? methodDescriptor.Options.GetExtension(AnnotationsExtensions.Http) : null;
```

ObsidianMinor:
`GetExtension` will return the default value for the extension if it's not present which for messages is null.

jtattermusch:
> So to use this:
> 
> ```cs
> var httpRule = methodDescriptor.Options.GetExtension(AnnotationsExtensions.Http);
> ```
> 
> If `HttpRule` is not present on the proto method will `GetExtension` return null or throw an error? If it errors then I'm guessing this will work:
> 
> ```cs
> var httpRule = methodDescriptor.Options.HasExtension(AnnotationsExtensions.Http)
>     ? methodDescriptor.Options.GetExtension(AnnotationsExtensions.Http)
>     : null;
> ```

Answer to this question is pretty much documented by this test: https://github.com/protocolbuffers/protobuf/pull/7434/files#diff-605100c634db140d708f7112fc7ab7bcR83

- GetExtension returns the default value, it doesn't throw
- Options messages can be null (so you need to use the `?.` operator) 
- because Options creates a defensive copy each time you request it, it might be better to do `var options = methodDescriptor.Options;` first and then use that instance repeatedly (but it doesn't affect correctness, only efficiency).

JamesNK:
> because Options creates a defensive copy each time you request it, it might be better to do var options = methodDescriptor.Options; first and then use that instance repeatedly (but it doesn't affect correctness, only efficiency).

You should consider a method instead of a property. Properties are assumed to be free/cheap to access. Methods indicate there is a cost and the value should be assigned to a local variable. I don't know what name you would choose. GetOptions is already taken.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
jtattermusch: Improved the docs.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
jtattermusch: Good point, removed the cast.

jtattermusch:
> > because Options creates a defensive copy each time you request it, it might be better to do var options = methodDescriptor.Options; first and then use that instance repeatedly (but it doesn't affect correctness, only efficiency).
> 
> You should consider a method instead of a property. Properties are assumed to be free/cheap to access. Methods indicate there is a cost and the value should be assigned to a local variable. I don't know what name you would choose. GetOptions is already taken.

looks like GetOption() is taken, GetOptions() is not.

Not sure if Options property or GetOptions method is better.

jtattermusch:
> By the way, here's an even more diabolical example of custom options.
> 
> ```protobuf
> syntax = "proto2";
> 
> import "google/protobuf/descriptor.proto";
> 
> message FooOptions {
>   // Custom field option used in definition of the extension message.
>   optional int32 int_opt = 1 [(foo_options) = {
>              int_opt: 1
>              [foo_int_opt]: 2
>              [foo_foo_opt]: {
>                int_opt: 3
>              }
>   }];
>   extensions 1000 to max;
> }
> 
> extend google.protobuf.FieldOptions {
>   // Custom field option used on the definition of that field option.
>   optional int32 bar_options = 1000 [(bar_options) = 1234];
> 
>   optional FooOptions foo_options = 1001;
> }
> 
> extend FooOptions {
>   optional int32 foo_int_opt = 1000;
>   optional FooOptions foo_foo_opt = 1001;
> }
> ```

Do these proto definitions already exist somewhere in the protobuf codebase? I'd expect so because presumably other languages already have tests for these cases?

jtattermusch:
Friendly ping, can we get this reviewed and merged?

jskeet:
It wasn't clear to me that it was ready for another review given the nature of the last few comments. I'll have another look tomorrow.

jtattermusch:
> It wasn't clear to me that it was ready for another review given the nature of the last few comments. I'll have another look tomorrow.

I think the only undecided topic is whether to expose `Options` property (even though it actually created a copy of the options) or whether to have `GetOptions()` method instead (as @JamesNK has suggested)
I think I'm fine with the current approach (a property) because the copy will be pretty lightweight in most normal cases (options are handwritten in the .proto file so usually there will only be a few of them present) and IMHO using a property feels more natural. That said, I don't feel strongly about this and I'd be happy to be convinced otherwise.

csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -128,12 +128,21 @@ public EnumValueDescriptor FindValueByName(string name)
...
+        /// The <c>EnumOptions</c>, defined in <c>descriptor.proto</c>.
+        /// If the options message is not present (=there are no options), <c>null</c> is returned.
```
jskeet, line 136:
```suggestion
        /// If the options message is not present (i.e. there are no options), <c>null</c> is returned.
```

jskeet: (Ditto in all options, of course.)

haberman:
If it is making a copy, putting it behind a method like `GetOptions()` seems less surprising to me.

jtattermusch:
> If it is making a copy, putting it behind a method like `GetOptions()` seems less surprising to me.

@jskeet  do you have a preference? I still think using a property is fine.

jskeet:
I'd just about prefer a method here - it really does make it clearer - but I'm not massively bothered either way.

jtattermusch:
> I'd just about prefer a method here - it really does make it clearer - but I'm not massively bothered either way.

Ok, looks like I've been outvoted then. I'll turn this into a GetOptions() method.

jtattermusch:
@haberman  all feedback addressed.

jtattermusch:
Test failures seem unrelated, merging.

jtattermusch:
@haberman  I think we should cherry-pick this for 3.12.x branch (It's a fix).

PR #7443, opened 2020-04-29
bazel: remove workaround of redundant zlib source

lizan:
It is no longer the case so we don't need to add headers into src. With the srcs added it break if zlib is used as rules_foreign_cc dependency.

Signed-off-by: Lizan Zhou <lizan@tetrate.io>


acozzette:
Let me close this pull request since it has been in the draft state for a while with no activity. If you're still interested in working on it, though, feel free to comment here and we can reopen it.

PR #7448, opened 2020-04-30
C# Add tests for self-referential options

jtattermusch:
Based on https://github.com/protocolbuffers/protobuf/pull/7434

The new C# test is passing, so I think https://github.com/protocolbuffers/protobuf/pull/7434 can be merged separately and this PR can go in as a followup (as it's not directly related to the Options API, it just makes sure that we are correctly handling some unusual cases).


jtattermusch:
Test failures seem unrelated to C#.

PR #7451, opened 2020-05-01
Support cross-compiling when using the cmake build system.

mwerle:
Adds the "WITH_PROTOC" definition when configuring a build with cmake, which specifies an existing 'protoc' binary to use.  This is required when building for another target.

"WITH_PROTOC" was chosen to specifically mimick the "--with-protoc" option used when using the standard 'automake' based build system.

Furthermore, 'libprotoc' is now only built if it is actually required, which can save some porting effort.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -37,6 +37,9 @@ if (CMAKE_CXX_COMPILER_ID MATCHES Intel)
 endif()
 
 # Options
+if(WITH_PROTOC)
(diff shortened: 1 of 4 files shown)


acozzette:
Thanks, @mwerle.

mwerle:
Hi @acozzette ,

Darn, just realised that I forgot to update this pull request.

The actual pull request is fine, but the description is wrong.  It should have said "WITH_PROTOC", and not "WITH_CMAKE".

Suppose too late to amend that now?

acozzette:
Oh, I missed that, too. I'm afraid it's too late to fix it now, but I wouldn't worry about it.

PR #7453, opened 2020-05-01
Stop building binary gems for ruby <2.5.

haberman:
It is no longer possible to "bundle install" these versions:

$ bundle install
Fetching gem metadata from https://rubygems.org/.................
reline-0.1.3 requires ruby version >= 2.5, which is incompatible with the current version, ruby 2.4.9p362


uyu423:
@haberman 

We are using Amazon Linux 4.14.1 for Node.js and yum repository. For that OS image, it only supports installation up to ruby 2.4. All environments are using similar versions of Amazon Linux images and this will be affected to all our server. So this feature appears to be critical issues.

PR #7458, opened 2020-05-04
Fix CLI "too long" issues in py_proto_library.

htuch:
Envoy was experiencing failed builds due to lots of duplicate -I in the
CLI. The solution is to generally use depset() as per Starlark best
practice for this sort of thing.

Might fix other issues beyond py_proto_library, anything that depends on
proto_gen.

release notes: no

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -85,8 +85,8 @@ def _proto_gen_impl(ctx):
         import_flags = depset(direct=["-I."])
 
     for dep in ctx.attr.deps:
-        import_flags = depset(transitive=[import_flags, dep.proto.import_flags])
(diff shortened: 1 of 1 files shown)


acozzette:
@htuch I see a Bazel error in the CI test, do you know what that is about?

htuch:
@acozzette should be fixed now, can you run CI again? Thanks.

acozzette:
Thanks, @htuch.

Yannic:
@htuch Suspecting this as the root-cause for the failure on Bazel-CI. https://buildkite.com/bazel/protobuf/builds/3316

htuch:
@Yannic thanks, I think https://github.com/protocolbuffers/protobuf/pull/7554 should fix this.

PR #7460, opened 2020-05-04
Implement HasPresence for C#

jskeet:
FieldAccessor.HasPresence returns true if both ClearValue and HasValue can be expected to work. Some fields have a working ClearValue, but no HasValue; HasPresence returns false for those fields.

Generally:
- Repeated fields return false
- Map fields return false
- Message fields return true
- Oneof member fields return true
- Singular primitive proto2 fields return true
- Singular explicitly-optional proto3 fields return true
- Singular not-explicitly-optional proto3 fields return false
- Repeated extension fields return false
- Singular extension fields return true

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Reflection/FieldAccessTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Reflection/FieldAccessTest.cs
@@ -98,7 +98,48 @@ namespace Google.Protobuf.Reflection
         }
 
         [Test]
-        public void HasValue_Proto3()
+        public void HasValue_Proto3_Message()
+        {
+            var message = new TestAllTypes();
+            var accessor = ((IMessage) message).Descriptor.Fields[TestProtos.TestAllTypes.SingleForeignMessageFieldNumber].Accessor;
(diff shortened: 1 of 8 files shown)


haberman:
Hmm, this seems kind of complicated. In the other languages, `hasPresence()` is just a few-lines-long method on `FieldDescriptor`. While it has implications for `Reflection::HasField()`, it isn't implemented on Reflection.

What is it buying us to introduce the extra interface, and implement this on `FieldAccessor`?

jskeet:
> What is it buying us to introduce the extra interface, and implement this on FieldAccessor?

Well, the field access is what works out what to *do* with Clear/Has requests... it felt appropriate to keep the logic there, where we're already having to make the checks for whether it's a message field, a proto3 with optional etc. If you'd prefer I just effectively put all those rules into the property, I could definitely do that (and the tests shouldn't need to change).

jskeet:
I've implemented an alternative design in #7466 - I do like the simplicity of the new design, but I still find it unfortunate that `FieldDescriptor` effectively has centralized logic. It's probably better overall though.
Will close this one for now.

PR #7468, opened 2020-05-05
Fix include path to reflection.h in documentation

asp437:
There is a note in the documentation to `Reflection::GetRepeatedFieldRef()`
about file `net/proto2/public/reflection.h` which contains a definition of
`RepeatedFieldRef`. Since there are few places in code with rewrite of
`net...` path to `google/protobuf/...` it looks like the first one is a
legacy path and should be renamed to `google/protobuf` in documentation
as well.

### DIFF
--- a/src/google/protobuf/message.h
+++ b/src/google/protobuf/message.h
@@ -740,8 +740,7 @@ class PROTOBUF_EXPORT Reflection final {
   // long as the message is not destroyed.
   //
   // Note that to use this method users need to include the header file
-  // "google/protobuf/reflection.h" (which defines the RepeatedFieldRef
-  // class templates).
+  // "reflection.h" (which defines the RepeatedFieldRef class templates).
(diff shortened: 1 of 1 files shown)


asp437:
@googlebot I signed it!

src/google/protobuf/message.h
```diff
@@ -740,7 +740,7 @@ class PROTOBUF_EXPORT Reflection final {
   // long as the message is not destroyed.
   //
   // Note that to use this method users need to include the header file
-  // "net/proto2/public/reflection.h" (which defines the RepeatedFieldRef
```
acozzette, line 743: Could you perhaps change it to just reflection.h without the full path? For historical reasons we have a somewhat different path structure within Google, which is why there are some comments like this referring to net/proto2/public.

acozzette:
Thanks, @asp437.

PR #7472, opened 2020-05-06
Use namespace macros everywhere

keith:
Previously the custom `PROTOBUF_NAMESPACE_*` macros were used in some
places but not others. With this change they're used everywhere which
allows consumers to replace these to namespace protobuf differently, in
order to link multiple versions of it into the same binary.

Related: https://groups.google.com/forum/#!topic/protobuf/hyOliJg4_J8


haberman:
We think this is a good direction, thanks for sending. We had recently been talking about doing this very thing.

We're going to try this inside Google first to make sure we can land it.

keith:
Awesome! Let me know how it goes in case I need to update anything here!

HanShelby:
Hi, keith. I met the problem same as you. Can I use your branch to compile a lib using custom namespace? And why your merge did not pass?

keith:
At this point it looks to be quite outdated, and updating it is a huge pain as you might imagine. Even with this building it didn't fully work for us since some dynamically generated symbols still had the wrong prefix. If the protobuf folks are interested in taking this I might be able to rebase, but otherwise I'll probably drop it.

keith:
@haberman anything come from internal discussions?

HanShelby:
Yes, I tried your branch and the result same as you said, it didn't fully work. I already gave up, it seems like there is not a better way to avoid symbol conflict in addition to link with static libraries.

PR #7487, opened 2020-05-11
Format default values for "present" fields regardless of value

### LINKED ISSUE #7486: C# conformance test failure for proto2 JSON formatting
JSON formatting is ignoring field presence for proto2 fields. Here's a failing test in C#:

```csharp
[Test]
public void Proto2_DefaultValuesWritten()
{
    var value = new ProtobufTestMessages.Proto2.TestAllTypesProto2() { FieldName13 = 0 };
    AssertWriteValue(value, "{ 'FieldName13': 0 }");
}
```

I have some code that fixes this by using HasPresence, but it then breaks for message types when specifying "format default values".

I've managed to get the failing test to pass without breaking anything else, but there's significant impact in terms of the complexity of documenting `JsonFormatter.Settings.FormatDefaultValues`.
Effectively, I've got rules of:

- Fields that support presence use that... except when the field is a message field that's not within a oneof
- Otherwise, for default values, use the FormatDefaultValues setting

I'll create a PR with that change in, but we need to be sure it's the behavior we want.

jskeet:
Fixes #7486.

The semantics are relatively complex to explain, even though the
code isn't particularly complex. We need to be certain that these
are the semantics we want.

### DIFF
--- a/csharp/src/Google.Protobuf/JsonFormatter.cs
+++ b/csharp/src/Google.Protobuf/JsonFormatter.cs
@@ -221,19 +221,12 @@ namespace Google.Protobuf
             foreach (var field in fields.InFieldNumberOrder())
             {
                 var accessor = field.Accessor;
-                if (field.ContainingOneof != null && field.ContainingOneof.Accessor.GetCaseFieldDescriptor(message) != field)
-                {
-                    continue;
-                }
-                // Omit default values unless we're asked to format them, or they're oneofs (where the default
-                // value is still formatted regardless, because that's how we preserve the oneof case).
-                object value = accessor.GetValue(message);
-                if (field.ContainingOneof == null && !settings.FormatDefaultValues && IsDefaultValue(accessor, value))
+                var value = accessor.GetValue(message);
+                if (!ShouldFormatFieldValue(message, field, value))
                 {
                     continue;
                 }
 
-                // Okay, all tests complete: let's write the field value...
                 if (!first)
                 {
                     writer.Write(PropertySeparator);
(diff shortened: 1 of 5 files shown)


csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -221,16 +221,27 @@ private bool WriteMessageFields(TextWriter writer, IMessage message, bool assume
...
             {
                 var accessor = field.Accessor;
-                if (field.ContainingOneof != null && field.ContainingOneof.Accessor.GetCaseFieldDescriptor(message) != field)
+                // Populated lazily, so we don't end up boxing a load of values that
+                // we'll just discard due to field presence.
+                object value;
+                // If we know about presence for the field, use that to determine whether
+                // or not to write the value. We *don't* use this for (non-oneof) message fields,
+                // so that we can control whether null is written or not via FormatDefaultValues.
+                if (field.HasPresence && !(field.FieldType == FieldType.Message && field.ContainingOneof == null))
```
haberman, line 230: Why can't this just be:

```
  if (field.HasPresence) {
```

Why exclude messages and oneofs from this? Any field with presence should be skipped if it is not present.

jskeet: It's for message fields that *aren't* in a oneof - so that if you've specified "format default values" you still get null. Otherwise this is a breaking change.

jskeet:
@haberman: I've rewritten this now, in a way that I *think* is what we discussed. PTAL. We really need to get this specified in a cross-language way in some sort of flow chart...

jskeet:
@haberman: Ping?

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -248,6 +241,31 @@ private bool WriteMessageFields(TextWriter writer, IMessage message, bool assume
             return !first;
         }
 
+        /// <summary>
+        /// Determines whether or not a field value should be serialized according to the field,
+        /// its value in the message, and the settings of this formatter.
+        /// </summary>
+        private bool ShouldFormatFieldValue(IMessage message, FieldDescriptor field, object value)
+        {
+            bool formatDefaultValue = settings.FormatDefaultValues &&
```
haberman, line 250: Can this be simply:

```
bool formatDefaultValue = settings.FormatDefaultValues && !field.HasPresence
```

jskeet: No. If FormatDefaultValues is set to true, then I'd expect a singular `int32` field which is either in a proto2 message or explicitly optional in a proto3 message to be formatted regardless of presence... although I can see arguments in favour of making field presence trump FormatDefaultValues.

I think we should probably decide that point, and then decide the implementation separately.

jskeet:
@haberman: I've added an extra commit along the lines you suggest. PTAL, and if you're happy I'll rebase to squash the two commits together (and reword).

Note that currently if you set `FormatDefaultValues` it *does* format repeated fields as an empty array - if that's not what you'd expect, we should fix that at the same time.

PR #7351, opened 2020-04-02
New Span-based C# parsing logic

jtattermusch:
A big refactor of C# parsing logic to allow parsing from ReadOnlySequence<byte> (which would allow parsing directly from the data as received on the wire without requiring to copy the buffers and thus reducing the GC pressure a lot).

Main changes:
- refactor out the parsing logic to helper classes (to avoid any duplication of the parsing logic)
- store all the parsing state into a struct (currently called ParserInternalState)
- introduce a ParseContext ref struct that has public methods, but they are only meant to be invoked by the generated code
- generate a InternalMergeFrom(ref ParseContext ctx) method, which has the actual parsing code (and is only supposed to be called by the generated code, never directly by the users.
- allow easy and fast conversion from CodedInputStream into a ParseContext (because the parsing state is encapsulated in the ParserInternalState, so bringing the state between the two only requires copying a struct.
- the original IMessage.MergeFrom(CodedInputStream cis) methods redirects to InternalMergeFrom(ref ParseContext). This transition has some (very small = copying a struct) overhead, but is basically only done once in the entire parsing tree, so that seem acceptable.
- all the public APIs that take CodedInputStream still behave exactly the same.
- compatibility with old generated code: if one uses old generated code against a newer runtime, everything will still work, but there might be some performance penalty (see benchmarks). Regenerating the code will speed things up.
- the public API increase is kept as small as possible. Basically the only new method we provide is Parse.ParseFrom(ReadOnlySequence<byte> byte) as entry point for the more efficient parsing. There are a few other additions (parsing maps, repeated fields etc.), but they are only meant to be invoked by the generated code.
- this PR only addresses memory-efficient parsing. Memory efficient serialization can be added in a followup PR.

This PR drops netstandard1.0 support (which doesn't support the Span type we need) and adds netstandard1.1 instead. The impact on supported platforms should be minimal (almost everyone supports netstandard2.0 these days anyway).

Basically this PR a new and simpler take on https://github.com/protocolbuffers/protobuf/pull/5888 (which is much bigger, introduces much more code duplication and leads to bigger public API increase).

PR #5888, opened 2019-03-13
Add CodedInputReader and CodedOutputWriter

JamesNK:
This is very WIP/rough. I had to create a new `ISpanMessage<T>` interface, and some areas do not have full support for it yet (they are throwing NotSupportException right now).

Nothing has been done on the code gen side. The way the reader and writer are used compared to the streams is almost identitical so code gen changes should be minimal.

### Perf numbers:

WriteToByteArray and ParseFromByteArray both allocate the byte array. This matches how serialization from gRPC works today.

WriteToBufferWriter writes to an `IBufferWriter<byte>` and doesn't allocate at all. For a small message CodedOutputWriter is slower than CodedOutputStream. It is faster for medium/large messages. Still some tuning to be done.

ParseFromReadOnlySequence reads from a sequence. Its allocations are related to creating the message and whatever strings exist on the message. Using CodedInputReader is consistently faster than CodedInputStream.

```
|                    Method |     Mean |     Error |    StdDev |   Median | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|-------------------------- |---------:|----------:|----------:|---------:|------------:|------------:|------------:|--------------------:|
|          WriteToByteArray | 377.5 ns | 11.210 ns | 32.876 ns | 370.3 ns |      0.0682 |           - |           - |               288 B |
|        ParseFromByteArray | 385.2 ns |  8.280 ns | 23.624 ns | 377.5 ns |      0.2666 |           - |           - |              1120 B |
|       WriteToBufferWriter | 294.0 ns |  3.658 ns |  3.421 ns | 293.6 ns |           - |           - |           - |                   - |
| ParseFromReadOnlySequence | 338.6 ns |  6.696 ns | 10.621 ns | 336.0 ns |      0.1903 |           - |           - |               800 B |
```

@jskeet @davidfowl @JunTaoLuo @jtattermusch

PR #7490, opened 2020-05-12
Overcome the `ref struct` limitation for pre-roslyn compilers by introducing GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE for generated code

jtattermusch:
Followup for https://github.com/protocolbuffers/protobuf/pull/7351.

Attempts to fix the issue mentioned in https://github.com/protocolbuffers/protobuf/pull/7351#issuecomment-627160415

Introduces the `GOOGLE_PROTOBUF_COMPATIBILITY_MODE` define which allows to avoid referencing the `ParseContext` ref struct in the generated code.
The consequences are basically:
- if the user doesn't have a roslyn-generation compiler, they can still build the generated code (after setting GOOGLE_PROTOBUF_COMPATIBILITY_MODE the generated code ends up being pretty much identical as pre #7351). Another option for the users on older compilers is to simply not regenerate their generated code.
- there is some performance overhead (because CodedInputStream's parsing methods are now considered legacy). The overhead is removed by getting a newer compiler and recompiling.
- Folks with older compiler won't be able to use new `ParseFrom(ReadOnlySequence<byte> data)`, because that one requires that the messages implement `IBufferMessage` ( = support parsing from ParseContext).
- One important realization is that if the generated code is shipped in a pre-compiled form (e.g. in a nuget package), users with an old compiler will still be able to use the pre-compiled generated code even without the compatibility_mode enabled, as long as they don't touch the APIs utilizing `ref struct` directly (which they generally don't need to do). E.g. someone with an old compiler can still use the `ParseFrom(ReadOnlySequence<byte> data)` method as long as the generated code was compiled using a newer compiler in advance. That is a good news for the client libraries.

The differences compared to the old approach https://github.com/protocolbuffers/protobuf/pull/5888:
- we still avoid duplicating the parsing logic in the runtime
- the Span based parsing is now supported by default, users with legacy compilers will need to explicitly set the `GOOGLE_PROTOBUF_COMPATIBILITY_MODE` define to be able to compile the regenerated code.


jtattermusch:
CC @haberman @jskeet @JamesNK

jtattermusch:
For now I've manually tested the generated code with `GOOGLE_PROTOBUF_COMPATIBILITY_MODE` enabled and everything seems to be working (except the `ParseFrom(ReadOnlySequence<byte> data)` API of course), but still need to figure out some way to test it automatically.

jskeet:
Looking at the code, I think we can do better in terms of avoiding duplication, by being sneaky with a sort of "bait and switch" approach. Basically we're including the same merging body, but in one method or another based on the compatibility. We can do that while only providing that merging code once.

Current (pre-PR) code:

```csharp
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public void MergeFrom(pb::CodedInputStream input) {
      input.ReadRawMessage(this);
    }

    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {

    MERGING CODE HERE

    }
```

Code in this PR:

```csharp
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public void MergeFrom(pb::CodedInputStream input) {
    #if !GOOGLE_PROTOBUF_COMPATIBILITY_MODE
      input.ReadRawMessage(this);
    #else
        MERGING CODE HERE
    #endif
    }

    #if !GOOGLE_PROTOBUF_COMPATIBILITY_MODE
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {

        MERGING CODE HERE

    }
    #endif
```

How about instead, we do:

```csharp
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public void MergeFrom(pb::CodedInputStream input) {
    #if !GOOGLE_PROTOBUF_COMPATIBILITY_MODE
      input.ReadRawMessage(this);
    }
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
    #endif        

       MERGING CODE HERE
    }
```

In other words:

- We always start the MergeFrom method
- In "regular" mode, we provide a method body, finish that method, then start InternalMergeFrom
- Then there's the merging code
- Then we finish whichever method is actually present at this point

This assumes that the merging code is *identical* between compatibility and non-compatibility mode - that assumption may not be accurate.

jskeet:
Slightly less ugly alternative (a little duplication, but probably clearer):

```csharp
    #if GOOGLE_PROTOBUF_COMPATIBILITY_MODE
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public void MergeFrom(pb::CodedInputStream input) {
    #else
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public void MergeFrom(pb::CodedInputStream input) {
      input.ReadRawMessage(this);
    }

    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
    #endif        

       MERGING CODE HERE
    }
```

jtattermusch:
> Looking at the code, I think we can do better in terms of avoiding duplication, by being sneaky with a sort of "bait and switch" approach. Basically we're including the same merging body, but in one method or another based on the compatibility. We can do that while only providing that merging code once.
> 
> Current (pre-PR) code:
> 
> ```cs
>     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
>     public void MergeFrom(pb::CodedInputStream input) {
>       input.ReadRawMessage(this);
>     }
> 
>     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
>     void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
> 
>     MERGING CODE HERE
> 
>     }
> ```
> 
> Code in this PR:
> 
> ```cs
>     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
>     public void MergeFrom(pb::CodedInputStream input) {
>     #if !GOOGLE_PROTOBUF_COMPATIBILITY_MODE
>       input.ReadRawMessage(this);
>     #else
>         MERGING CODE HERE
>     #endif
>     }
> 
>     #if !GOOGLE_PROTOBUF_COMPATIBILITY_MODE
>     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
>     void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
> 
>         MERGING CODE HERE
> 
>     }
>     #endif
> ```
> 
> How about instead, we do:
> 
> ```cs
>     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
>     public void MergeFrom(pb::CodedInputStream input) {
>     #if !GOOGLE_PROTOBUF_COMPATIBILITY_MODE
>       input.ReadRawMessage(this);
>     }
>     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
>     void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
>     #endif        
> 
>        MERGING CODE HERE
>     }
> ```
> 
> In other words:
> 
> * We always start the MergeFrom method
> * In "regular" mode, we provide a method body, finish that method, then start InternalMergeFrom
> * Then there's the merging code
> * Then we finish whichever method is actually present at this point
> 
> This assumes that the merging code is _identical_ between compatibility and non-compatibility mode - that assumption may not be accurate.

That's a good point, unfortunately the merging code is very similar but now quite the same. For repeated field, map field, wrappers and for unknown fields and extensions, we need to pass `ref input` instead of `input` when using `ParseContext`.
The good news in terms of compilation speed is that we are always only including one copy of parsing logic (the legacy OR the new version, never both). Assuming the compiler is good at skipping the unimportant `#if` branches, the compilation speed (and the resulting code size) should remain almost the same. Only the size of the source .cs file is going to be larger.

jskeet:
Rats, that's what I get for not looking at a large enough sample. Thanks for the rapid evaluation :)

src/google/protobuf/compiler/csharp/csharp_map_field.cc
```diff
@@ -94,9 +94,19 @@ void MapFieldGenerator::GenerateMergingCode(io::Printer* printer) {
...
 
 void MapFieldGenerator::GenerateParsingCode(io::Printer* printer) {
-  printer->Print(
-    variables_,
-    "$name$_.AddEntriesFrom(ref input, _map_$name$_codec);\n");
+  GenerateParsingCode(printer, true);
+}
+
+void MapFieldGenerator::GenerateParsingCode(io::Printer* printer, bool use_parse_context) {
+  if (use_parse_context) {
```
jskeet, line 101: Possibly use a conditional operator here instead?

```
printer->Print(
  variables_,
  use_parse_context
  ? "$name$_.AddEntriesFrom(ref input, _map_$name$_codec);\n"
  : "$name$_.AddEntriesFrom(input, _map_$name$_codec);\n");
```

That has the benefit that the two versions are on consecutive lines, making the difference obvious.

jtattermusch:
Alternatively, the define could be named `GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE` to make its name more specific.

src/google/protobuf/compiler/csharp/csharp_map_field.cc
jskeet: (There are a few others, too.)

JamesNK:
This seems like a good solution. The cost to anyone working with modern .NET Core is bigger generated code files, but the unused implementation will be removed via `#ifdef` during compile so there will be no runtime impact.

haberman:
Some really nice problem solving here!

My main feedback is:
- Is it possible to have tests for this?
- How will we document this for users?

jskeet:
I believe we should be able to write a test for this, only on Windows. I'm happy to try to do that as a separate PR when this is merged. Basically it would try to build the test proto project using the old native C# compiler.

For documentation: I think a combination of release notes and https://developers.google.com/protocol-buffers/docs/csharptutorial is likely to be enough.

jtattermusch:
Ok, I believe I addressed all the feedback and this is now ready for final review.
- the define is not called GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE
- I added a nunit test that calls an old C# compiler and makes sure it can build the generated code in compatibility mode.

csharp/src/Google.Protobuf.Test/RefStructCompatibilityTest.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <summary>
+        /// Invoke an old C# compiler in a subprocess and check it finished successful.
+        /// </summary>
+        /// <param name="args"></param>
+        /// <param name="workingDirectory"></param>
+        private void RunOldCsharpCompilerAndCheckSuccess(string args, string workingDirectory)
+        {  
+            using (var process = new Process())
+            {
+                process.StartInfo.FileName = "c:\\Windows\\Microsoft.NET\\Framework\\v4.0.30319\\csc.exe";  // Old C# 5 compiler from .NET framework
```
JamesNK, line 83: Hard coded path is very brittle. Windows could be installed to a different path or drive.

Should look this up from an environmental variable.

JamesNK: https://stackoverflow.com/questions/6660512/how-to-get-csc-exe-path

I think this would work if the test only runs with the net45 target:
```cs
using System.Runtime.InteropServices;
var frameworkPath = RuntimeEnvironment.GetRuntimeDirectory();
var cscPath = Path.Combine(frameworkPath, "csc.exe");
```

Alternatively the %WINDIR% environment variable would be an improvement. To be more exact and lookup the exact path would probably require accessing Windows registry.

jtattermusch: Good point. I updated the test to use the WINDIR env. It's not 100% but seems good enough for our purposes.

jtattermusch:
@jskeet  @haberman  is there any more feedback? I'd like to merge soon to fix the problem, I have some followup PRs in the queue.

jskeet:
I was fine with it last time I looked; I don't have time to take another careful look at the moment I'm afraid - but if @haberman is happy, that's fine for me.

haberman:
This looks great. Could you possibly add something to CHANGES.txt in the "Unreleased Changes" section? I've started doing this so that there is less work to do at release time.

jtattermusch:
@haberman  
I updated CHANGES.txt with this change and #7351. I think we can merge this PR now?

vuqtran88:
Out of curiosity, how do you enable GOOGLE_PROTOBUF_COMPATIBILITY_MODE ? did you enable it in the `.proto` file ?

jtattermusch:
> GOOGLE_PROTOBUF_COMPATIBILITY_MODE

It's a compile define. You can enable it in your .csproj files. E.g. a similar example here: https://github.com/protocolbuffers/protobuf/blob/d5079ed7db6da57bd8552d65dbfe6df7c902dc52/csharp/src/Google.Protobuf/Google.Protobuf.csproj#L27

Btw, the correct name of the define is `GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE`  - see history of this PR.

lirivrod:
Hello was able to successfully compile my generated code using #define GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE
on top of the script however I'm not able to parse data using the Parse.ParseFrom method I know this is already stated here however I wanted to check with you if there are any workarounds or any other method that could be used when using GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE

jskeet:
@lirivrod: Please file a new issue with more details - that's more appropriate than using a merged PR. The expectation isn't that you define the symbol in individual files though, but instead do it for the whole project.

Issue #7374, opened 2020-04-13
generated C++ code has clang-tidy warnings: bugprone-unhandled-self-assignment and google-explicit-constructor


chih-hung:
**What version of protobuf and what language are you using?**
Version: latest ported to Android
Language: C++

**What operating system (Linux, Windows, ...) and version?**

Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**

clang-tidy

**What did you do?**

Compile Android Open Source frameworks source files with aprotoc and clang-tidy.

compiler/cpp/cpp_message.cc emits assignment operators like:

  inline DrmFrameworkMetrics_Attributes& operator=(const DrmFrameworkMetrics_Attributes& from) {
    CopyFrom(from);
    return *this;
  }
  inline DrmFrameworkMetrics_Attributes& operator=(DrmFrameworkMetrics_Attributes&& from) noexcept {
    if (GetArenaNoVirtual() == from.GetArenaNoVirtual()) {
      if (this != &from) InternalSwap(&from);
    } else {
      CopyFrom(from);
    }
    return *this;
  }

When the above code is compiled with clang-tidy, the first operator gets the following warning:
   warning: operator=() does not handle self-assignment properly [bugprone-unhandled-self-assignment]

Although CopyFrom does compare 'from' with this before assignment,
it is not inlined and invisible to clang-tidy.

To suppress this warning, could you change compiler/cpp/cpp_message.cc to emit a NOLINTNEXTLINE like the following?

  // NOLINTNEXTLINE(bugprone-unhandled-self-assignment)
  inline DrmFrameworkMetrics_Attributes& operator=(const DrmFrameworkMetrics_Attributes& from) {
    CopyFrom(from);
    return *this;
  }

In MessageGenerator::GenerateClassDefinition, it needs to emit
     "// NOLINTNEXTLINE(bugprone-unhandled-self-assignment)\n"
before the assignment operator.

Another clang-tidy warning is google-explicit-constructor for constructors with one argument like this:

 DrmFrameworkMetrics_SessionLifetimesEntry_DoNotUse(::PROTOBUF_NAMESPACE_ID::Arena* arena);

warning: single-argument constructors must be marked explicit to avoid unintentional implicit conversions [google-explicit-constructor]

Could you add another NOLINTNEXTLINE output in MessageGenerator::GenerateClassDefinition:

        "  // NOLINTNEXTLINE(google-explicit-constructor)\n"
[...]

PR #7492, opened 2020-05-12
Make Arena constructor explicit

eoger:
I believe this constructor is not intended to be a conversion one.

Related to https://github.com/protocolbuffers/protobuf/issues/7374.


haberman:
Thanks for the PR. You reminded me that I had a dormant change inside Google to do exactly the same thing. I landed it in Google first to make sure there were no breakages there, and then it was synced to GitHub in https://github.com/protocolbuffers/protobuf/pull/7502.

eoger:
That’s great to hear, thanks!

PR #7495, opened 2020-05-12
Fix integration tests for Go examples

dsnet:
A proper integration test should be hermetic such that its operation
is agnostic to whatever happens outside of its controlled environment.
The script is not so since it uses a fixed (and very old) version of Go,
but continues to build protoc-gen-go from head, leading to an eventual
breakage when protoc-gen-go at head no longer operates on such an old
version of Go.

Fix the script to pin to an older version of protoc-gen-go (i.e., v1.3.5).

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -163,10 +163,13 @@ build_golang() {
 
   export GOPATH="$HOME/gocode"
   mkdir -p "$GOPATH/src/github.com/protocolbuffers"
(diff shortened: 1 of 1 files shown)


dsnet:
I didn't manually test this. I'm hoping to have Kokoro do it. How do I kick off a build?

haberman:
In the future we may want to use a newer version of Go instead, but this fixes the build for now. Thanks Joe!

PR #7498, opened 2020-05-13
Add GetDebugString to Python FileDescriptor interface.

### LINKED ISSUE #7484: Descriptor's DebugString method(s) not exposed in Python API
Python

The C++ API provides [a `DebugString` method](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.h#L268-L274) that will serialize a given `Descriptor` into re-parseable Protobuf syntax. However, this function is not exposed in the Python API, and there seems to be no way to get similar functionality without reimplementing [the ~500 lines of logic done in C++](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.cc#L2459-L2974).

Python bindings for the `DebugString` method on `Descriptor`.

Reimplementing `DebugString` in Python natively.

psobot:
Closes #7484.

This PR exposes the C++ `FileDescriptor` `DebugString` method through to the Python C++ implementation, and adds a test for it.

### DIFF
--- a/python/google/protobuf/internal/descriptor_test.py
+++ b/python/google/protobuf/internal/descriptor_test.py
@@ -58,6 +58,28 @@ TEST_EMPTY_MESSAGE_DESCRIPTOR_ASCII = """
 name: 'TestEmptyMessage'
 """
 
+TEST_FILE_DESCRIPTOR_DEBUG = """syntax = "proto2";
+
+package protobuf_unittest;
+
+message NestedMessage {
+  enum ForeignEnum {
+    FOREIGN_FOO = 4;
+    FOREIGN_BAR = 5;
+    FOREIGN_BAZ = 6;
+  }
+  optional int32 bb = 1;
+}
+
+message ResponseMessage {
+}
+
+service Service {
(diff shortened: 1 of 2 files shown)


python/google/protobuf/internal/descriptor_test.py
```diff
@@ -121,6 +143,13 @@ def testContainingTypeFixups(self):
   def testContainingServiceFixups(self):
     self.assertEqual(self.my_service, self.my_method.containing_service)
 
+  @unittest.skipIf(
+    api_implementation.Type() != 'cpp',
+    'GetDebugString is only available with the cpp implementation',
+  )
+  def testGetDebugString(self):
+    self.assertEqual(self.my_file.GetDebugString(), TEST_FILE_DESCRIPTOR_DEBUG)
+
```
gebi, line 152: shouldn't this test be executed in python too as this PR merged that feature?
(sry just needed that feature and debugged on an older python-protobuf and stumbled over this PR, pls just ignore if everything is as it should be :)!

PR #7499, opened 2020-05-13
Conditionally include zlib.h when HAVE_ZLIB is set

jiuqiant:


acozzette:
@jiuqiant I thought the idea was that gzip_stream.h doesn't need to check HAVE_ZLIB, since this header shouldn't even be compiled if HAVE_ZLIB is unset. Are you getting an error? Which build system are you using (e.g. CMake, Automake, Bazel)?

jiuqiant:
It's Bazel with MSVC. 
In [the protobuf  build rule](https://github.com/protocolbuffers/protobuf/blob/master/BUILD#L209), `gzip_stream.h` is never excluded from the `hdrs` field for msvc. However,  the ZLIB_DEPS is not a dependency of protobuf when MSVC is used: https://github.com/protocolbuffers/protobuf/blob/master/BUILD#L204. We have a mismatch here.

Another solution is to make a select section in the protobuf cc_library rule for msvc. Since HAVE_ZLIB macros are already used in `gzip_stream.cc`, I think it's reasonable to have the macros in the header file as well.

acozzette:
@jiuqiant Thanks for the explanation, that makes sense. I think it would be slightly better though to keep gzip_stream.h out of the headers for the cc_library, since that would cut down on the number of ifdefs and be consistent with what the autotools build does.

PR #7501, opened 2020-05-13
Added background information about proto3 presence.

haberman:


docs/implementing_proto3_presence.md
```diff
@@ -14,13 +16,60 @@ independently by their authors. This includes:
...
+were previously the only supported presence mechanism for proto3. Users have
+pointed to both efficiency and usability issues with the wrapper types.
+
+Presence in proto3 uses exactly the same syntax and semantics as in proto2.
+Proto3 Fields marked `optional` will track presence like proto2, while fields
+without any label (known as "singular fields"), will continue to omit presence
+information.  The `optional` keyword was chosen to minimize differences with
+proto2.
+
+Unfortunately, for descriptors it is not possible to use the same representation
```
dlj-NaN, line 43: Maybe:
"Unfortunately, for the current descriptor protos and `Descriptor` API (as of v3.12.4), it is not possible ... "

(This would provide some anchoring against future changes to the Descriptor API or protos.)

docs/implementing_proto3_presence.md
```diff
@@ -14,13 +16,60 @@ independently by their authors. This includes:
...
+Since oneof fields in proto3 already track presence, existing proto3
+reflection-based algorithms should correctly preserve presence for proto3
+optional fields with no code changes. For example, the JSON and TextFormat
+parsers/serializers in C++ and Java did not require any changes to support
+proto3 presence. This is the major benefit of synthetic oneofs.
+
+This design does leave some cruft in descriptors. Synthetic oneofs are a
+compatibility measure that we can hopefully clean up in the future. For now
+though, it is important to preserve them across different descriptor formats
+and APIs. It is never safe to drop synthetic oneofs from a descriptor.  You
```
dlj-NaN, line 64: "It is never safe to drop synthetic oneofs from a descriptor."

Does this mean "a descriptor proto?" I guess it's a bit surprising, since the Descriptor API objects are immutable.

dlj-NaN, line 64: (Nit) Could "You" be "Code generators?" (And similarly below.)

docs/implementing_proto3_presence.md
haberman: Good point, done.

docs/implementing_proto3_presence.md
haberman: I'm using "descriptor" in a slightly more generic way that encompasses both the proto and API. Sometimes systems are converting proto schemas between various formats and have asked if they can drop synthetic oneofs.

I changed this to "proto schema" to make the more general meaning more clear.

haberman: Done.

PR #7503, opened 2020-05-13
Fixed the PHP Kokoro build.

haberman:


tests.sh
```diff
@@ -716,8 +717,11 @@ build_php7.0_mac() {
 build_php7.4_mac() {
```
TeBoring, line 717: Can we change this to build_php7.3_mac?

haberman: Done.

PR #7511, opened 2020-05-15
Removed the unnecessary setuptools package dependency for Python package

mishas:
The setuptools package was added to allow definition of namespaces using
the now outdated (and discouraged from use) pkg_resources-style.

The code here, for a long while now, uses a try/except (`ImportError`)
protection around the setuptools code, and falls back to the more
encoraged pkgutil-style.

Removing this library won't affect any current workflow, and in the case
of setuptools not found, it'll actually use a more modern (and
encouraged) flow.

More info: https://packaging.python.org/guides/packaging-namespace-packages/

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -241,7 +241,7 @@ if __name__ == '__main__':
     os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'
 
   # Keep this list of dependencies in sync with tox.ini.
-  install_requires = ['six>=1.9', 'setuptools']
+  install_requires = ['six>=1.9']
   if sys.version_info <= (2,7):
     install_requires.append('ordereddict')
     install_requires.append('unittest2')


python/setup.py
```diff
@@ -241,7 +241,7 @@ def get_option_from_sys_argv(option_str):
     os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'
 
   # Keep this list of dependencies in sync with tox.ini.
-  install_requires = ['six>=1.9', 'setuptools']
+  install_requires = ['six>=1.9']
```
haberman, line 244: Per the comment, can you verify whether any change to `tox.ini` is required?

mishas: Funnily enough - tox.ini was NOT in sync with this file, in the one way that it was missing setuptools :).

mishas:
@haberman, and update on this?

mishas:
@haberman , Friendly ping?

mishas:
Friendly ping again?

haberman:
Sorry for letting this slip. This looks fine to me, but I'd feel a little better if someone from our team who knows Python packaging stuff better than me took a look.

@dlj-NaN could you take a look?

mishas:
@dlj-NaN, friendly ping?

mishas:
@dlj-NaN , @haberman - monthly ping?

PR #7524, opened 2020-05-17
Use lexical-binding in protobuf-mode.el

eklitzke:
This is a very minor improvement to the Emacs `protobuf-mode.el` file, that just declares the file as using lexical binding. This allows some improvements in how the file is byte-compiled by Emacs, including letting it be optimized in Andrea Corallo's upstream `feature/native-comp` branch, which can only optimize elisp files that use lexical binding (which is how I discovered that the file wasn't using lexical binding in the first place).

The protobuf-mode.el file already requires Emacs >= 24.4, and lexical-binding was added in Emacs 24.1, so there's no chance this will break protobuf-mode for anyone using older Emacs versions.

I have already signed a Google individual CLA.


acozzette:
Thanks, @eklitzke.

Issue #6719, opened 2019-10-02
String field processing in getSerializedSize/writeTo looks suspicious


hpple:
Hey!

Can someone from protobuf team explain to me the rationale behind the Java code for string fields processing produced by protoc?

Protoc generates Java **Object** field for each proto string field and protobuf uses it to store either **ByteString** or **java.util.String** (dunno why).
Also each string field gets two accessor methods: **getXxx** for **String** and **getXxxBytes** for **ByteString.**

The suspicious part begins inside generated methods **getSerializedSize** and **writeTo.**

For both of these methods **getXxxBytes** will be called disregarding the current type of an instance stored in **Object** field which leads to unnecessary **ByteString** allocation when actual type is **java.util.String** (and this is the common case from my perspective).

It looks even more peculiar if we will take into account the fact that in both methods we may see a usage of dual-type-aware **GeneratedMessageV3.writeString**/**GeneratedMessageV3.computeStringSize** literally in the next line after **getXxxBytes** calls.

Is it a bug or an intentional decision?

**What version of protobuf and what language are you using?**
Version: v3.9.2 
Language: Java

**What did you expect to see**
getSerializedSize/writeTo have specialized versions for ByteString/String

**What did you see instead?**
getXxxBytes is always used for isEmpty check

hpple:
Is there any chance to get an answer?

PR #7526, opened 2020-05-17
[Java] Remove unnecessary allocation during String field emptiness checks

hpple:
Hi!

I've opened an issue about suspicious String field processing earlier:
#6719 

This patch is trying to fix code generation in case this is actually a bug.

### DIFF
--- a/src/google/protobuf/compiler/java/java_string_field.cc
+++ b/src/google/protobuf/compiler/java/java_string_field.cc
@@ -78,6 +78,8 @@ void SetPrimitiveVariables(const FieldDescriptor* descriptor,
       "  if (value == null) {\n"
       "    throw new NullPointerException();\n"
       "  }\n";
+  (*variables)["isStringEmpty"] = "com.google.protobuf.GeneratedMessage" +
+                                GeneratedCodeVersionSuffix() + ".isStringEmpty";
   (*variables)["writeString"] = "com.google.protobuf.GeneratedMessage" +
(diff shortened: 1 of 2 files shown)


hpple:
@googlebot I signed it!

hpple:
Can you take a look please? The change list is relatively small.

hpple:
Any chance to get a code review?

src/google/protobuf/compiler/java/java_string_field.cc
```diff
@@ -110,7 +112,7 @@ void SetPrimitiveVariables(const FieldDescriptor* descriptor,
     (*variables)["clear_has_field_bit_builder"] = "";
 
     (*variables)["is_field_present_message"] =
-        "!get" + (*variables)["capitalized_name"] + "Bytes().isEmpty()";
+       "!" + (*variables)["isStringEmpty"] + "(" + (*variables)["name"] + "_)";
```
fowles, line 115: Don't we know the type in the generated code such that we could call `(*variables)["name"] + "_.isEmpty()"` here directly?

hpple: Unfortunately, no.

As I initially described it in https://github.com/protocolbuffers/protobuf/issues/6719, the problem is that (for some reason) fields for **string** type in proto are generated as **java.lang.Object** in Java. So, such fields may contain either **ByteString** or **java.util.String** depending on which representation was requested last time.

PR #7548, opened 2020-05-25
check null pointer dereference for map_entry

YourButterfly:
null pointer dereference
```
syntax = "proto3";

package tutorial;

message Person0 {
	option map_entry = true;
}
// ...
```

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -253,6 +253,10 @@ void CollectMapInfo(const Options& options, const Descriptor* descriptor,
   std::map<std::string, std::string>& vars = *variables;
   const FieldDescriptor* key = descriptor->FindFieldByName("key");
   const FieldDescriptor* val = descriptor->FindFieldByName("value");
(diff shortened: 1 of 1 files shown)


acozzette:
In principle this should never happen, because the `map_entry` option is only intended to be used internally and shouldn't be set in .proto files. But I guess it can't hurt to add this extra input validation.

src/google/protobuf/compiler/cpp/cpp_message.cc
```diff
@@ -253,6 +253,10 @@ void CollectMapInfo(const Options& options, const Descriptor* descriptor,
   std::map<std::string, std::string>& vars = *variables;
   const FieldDescriptor* key = descriptor->FindFieldByName("key");
   const FieldDescriptor* val = descriptor->FindFieldByName("value");
+  // Can`t find the 'key'
```
acozzette, line 256: Could you change these to use `GOOGLE_CHECK_NOTNULL()`? I think you can also remove the comments since the checks are fairly self-explanatory.

YourButterfly:
https://github.com/protocolbuffers/protobuf/pull/7581

PR #7552, opened 2020-05-26
Update rules_cc

aherrmann:
The previous version does not mark the `toolchain_type` as public,
meaning that users of a CC toolchain using
`@rules_cc//cc:toolchain_type` will get build errors along the lines of
```
ERROR:
.../local_config_cc_toolchains/BUILD.bazel:18:1:
in toolchain rule
@local_config_cc_toolchains//:cc-toolchain-armeabi-v7a: alias
'@rules_cc//cc:toolchain_type' referring to target
'@bazel_tools//tools/cpp:toolchain_type' is not visible from target
'@local_config_cc_toolchains//:cc-toolchain-armeabi-v7a'. Check the
visibility declaration of the former target if you think the dependency
is legitimate
```
See https://github.com/digital-asset/daml/pull/6063#issuecomment-633580469

### DIFF
--- a/protobuf_deps.bzl
+++ b/protobuf_deps.bzl
@@ -35,9 +35,9 @@ def protobuf_deps():
     if not native.existing_rule("rules_cc"):
         http_archive(
             name = "rules_cc",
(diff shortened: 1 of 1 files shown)


acozzette:
Thanks, @aherrmann.

PR #7561, opened 2020-05-27
C# project cleanup for dotnet SDK 3.1

jtattermusch:
- allow building with newer dotnet SDK than just 3.0.100
- upgrade Microsoft.SourceLink.GitHub dependency (current version causes trouble with dotnet SDK 3.1.x)

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 ﻿{
   "sdk": {
-    "version": "3.0.100"
+    "version": "3.0.100",
+    "rollForward": "latestMinor"
   }
 }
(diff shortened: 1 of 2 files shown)


jtattermusch:
CC @JamesNK

global.json
```diff
@@ -1,5 +1,6 @@
 ﻿{
   "sdk": {
-    "version": "3.0.100"
+    "version": "3.0.100",
```
jskeet, line 3: Is it worth going to 3.1 (LTS) in the same PR? I don't know how difficult that is.

jtattermusch: Agreed that we should upgrade the dotnet SDK, but I'd rather do that in a separate PR.

Issue #6148, opened 2019-05-17
[PHP] Incompatibility with Xdebug


lhecker:
**What version of protobuf and what language are you using?**
Version: 3.7.1
Language: PHP 7.3 or later

**What operating system (Linux, Windows, ...) and version?**
Linux, macOS

**What runtime / compiler are you using (e.g., python version or gcc version)**
GCC, LLVM

**What did you do?**
1. Install PHP
2. Install protobuf and xdebug using pecl: `pecl install protobuf xdebug`
3. Make sure both are enabled (xdebug should show in `php --version`)
4. Run this: `php -r 'var_dump(new \Google\Protobuf\Timestamp());'`

**What did you expect to see**

```
$ php -r 'var_dump(new \Google\Protobuf\Timestamp());'
object(Google\Protobuf\Timestamp)#1 (0) {
}
```

**What did you see instead?**

```
$ php -r 'var_dump(new \Google\Protobuf\Timestamp());'
Segmentation fault
```

Running GDB yields:
```
$ gdb -ex=r --args php -r 'var_dump(new \Google\Protobuf\Timestamp());'
[…]
Program received signal SIGSEGV, Segmentation fault.
xdebug_zend_hash_is_recursive (ht=ht@entry=0x0) at /tmp/pear/temp/xdebug/xdebug_compat.c:450
450	/tmp/pear/temp/xdebug/xdebug_compat.c: No such file or directory.
(gdb) bt
#0  xdebug_zend_hash_is_recursive (ht=ht@entry=0x0) at /tmp/pear/temp/xdebug/xdebug_compat.c:450
#1  0x00007ffff2de922a in xdebug_var_export_text_ansi (struc=<optimized out>, str=0x555555dd2bc0, mode=0, level=1, debug_zval=0, options=0x555555ddd240)
    at /tmp/pear/temp/xdebug/xdebug_var.c:1435
#2  0x00007ffff2dea025 in xdebug_get_zval_value_text_ansi (val=<optimized out>, val@entry=0x7ffff307f040, mode=0, debug_zval=<optimized out>, 
    options=0x555555ddd240) at /tmp/pear/temp/xdebug/xdebug_var.c:1507
#3  0x00007ffff2dcde95 in zif_xdebug_var_dump (execute_data=<optimized out>, return_value=<optimized out>) at /tmp/pear/temp/xdebug/xdebug.c:2243
[...]

PR #7562, opened 2020-05-27
Return empty PHP debug info instead of null

dcloues:
This addresses #6148 which describes an incompatibility between Xdebug and classes derived from the Protobuf extension's internal `\Google\Protobuf\Internal\Message` class. This incompatibility causes a segfault due to a null pointer dereference when using Xdebug's overridden `var_dump()` function or whenever a step debugger inspects one of these instances. The following code segfaults when Xdebug and Protobuf extensions are both installed:

```
<?php

var_dump(new \Google\Protobuf\Timestamp());
```

PHP internal provide two different handlers for accessing an object's properties:
[`get_properties`](https://wiki.php.net/internals/engine/objects#get_properties) and [`get_debug_info`](https://wiki.php.net/internals/engine/objects#get_debug_info).

These handlers offer different guidance: `get_debug_info` "Should not return NULL.", but for `get_properties`, returning `NULL` is explicitly supported in order to support cases such as invocation by the garbage collector.

PHP's helpers for accessing object properties (`zend_get_properties_for()` in PHP 7.4, and the various `Z_OBJDEBUG` macros) fall back to `get_properties` if the `get_debug_info` handler isn't defined, and this is where we get into trouble: Xdebug tries to access debug info for these classes but receives `NULL`, unexpectedly.

As a fix, I added a separate `get_debug_info` handler to the base `Message` class. This handler initializes an empty array and returns it, setting `is_tmp` to indicate that the caller should take ownership. While it's wasteful to allocate a new array every time, `get_debug_info` is accessed infrequently.

This change seemed easier and less invasive than modifying `get_properties()` to return an array. The docs for `get_properties()` indicate that the handler, not the caller, owns the array. This means `get_properties()` cannot allocate a temporary array on the fly. `get_debug_info()`, on the other hand, is designed for this case.

With this change in place, the example code above prints out Xdebug's extended `var_dump` output, as expected:

```
/home/vagrant/proto.php:3:
class Google\Protobuf\Timestamp#1 (0) {
}
```

While it would be nice to get _real_ debug info out of these message classes, that would be a much more involved change. This change only prevents the segfault, without trying to make debug info more useful.

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -89,6 +89,7 @@ static zval* message_get_property_ptr_ptr(zval* object, zval* member, int type,
 static HashTable* message_get_gc(zval* object, zval** table, int* n);
 #endif
 static HashTable* message_get_properties(zval* object TSRMLS_DC);
+static HashTable* message_get_debug_info(zval *object, int *is_temp TSRMLS_DC);
 
 // -----------------------------------------------------------------------------
 // PHP Message Handlers
@@ -125,6 +126,7 @@ PHP_PROTO_INIT_CLASS_START("Google\\Protobuf\\Internal\\Message",
   message_handlers->read_property = message_get_property;
   message_handlers->get_property_ptr_ptr = message_get_property_ptr_ptr;
   message_handlers->get_properties = message_get_properties;
+  message_handlers->get_debug_info = message_get_debug_info;
   message_handlers->get_gc = message_get_gc;
 PHP_PROTO_INIT_CLASS_END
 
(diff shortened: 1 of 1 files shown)


dcloues:
@googlebot I signed it!

dcloues:
@googlebot I signed it!

TeBoring:
Did you sign with the same email with which you did the commit?

TeBoring:
This one: dustin.shieldscloues@mailchimp.com

lucasantarella:
+1 please merge this. These seg faults are disrupting my development flow....Please merge!!!

TeBoring:
@dcloues how's your cla going?

dcloues:
@TeBoring Sorry for lagging on getting back to you - I'm trying to get our corporate CLA sorted out, and I'll check back here when we make progress.

thicolares:
Hi @dcloues! How's your CLA going? Any updates? I hope you could merge it soon :)

acozzette:

PR #7570, opened 2020-05-28
Fix #5858:provide clear message when specifying a directory as an input file

### LINKED ISSUE #5858: Misleading warning message when specifying a directory as an input file
Version: master
Language: N/A

Linux

Latest protoc build from source using `gcc (GCC) 8.2.1 20181127`

**What did you do?**
I mistakenly passed a directory as an input file to protoc.
For example:
```bash
cd /tmp
mkdir test
protoc -I . ./test --go_out=/tmp/out --java_out=/tmp/out
```

I'd expect an error message that there are no valid input files.

**What did you see instead?**
protoc outputs a warning about missing syntax version but still generates java and go code (actually happens for all languages) with no message definitions.

The warning message was:
```
[libprotobuf WARNING google/protobuf/compiler/parser.cc:564] No syntax specified for the proto file: test. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defaulted to proto2 syntax.)
```
From a quick investigation, the problem is that `OpenDiskFile` [1] does not check if the given input file path is actually a regular file. Since `open(3)` is capable of opening special files (like directories), protoc continues as if it has opened an empty file.

[1] https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/importer.cc#L489

junekhan:
Fix: [5858](https://github.com/protocolbuffers/protobuf/issues/5858)
Language: C++

### DIFF
--- a/src/google/protobuf/compiler/importer.cc
+++ b/src/google/protobuf/compiler/importer.cc
@@ -490,6 +490,15 @@ io::ZeroCopyInputStream* DiskSourceTree::OpenVirtualFile(
 
 io::ZeroCopyInputStream* DiskSourceTree::OpenDiskFile(
     const std::string& filename) {
+  struct stat sb;
+  int ret = 0;
+  do {
+    ret = stat(filename.c_str(), &sb);
+  } while (ret != 0 && errno == EINTR);
+  if (!S_ISREG(sb.st_mode)) {
+    last_error_message_ = "Input file is not a regular file.";
+    return NULL;
+  }
   int file_descriptor;
   do {
     file_descriptor = open(filename.c_str(), O_RDONLY);
--- a/src/google/protobuf/compiler/importer.cc
+++ b/src/google/protobuf/compiler/importer.cc
@@ -495,12 +495,8 @@ io::ZeroCopyInputStream* DiskSourceTree::OpenDiskFile(
   do {
     ret = stat(filename.c_str(), &sb);
   } while (ret != 0 && errno == EINTR);
-#ifdef _WIN32
-  if ((sb.st_mode & _S_IFMT) == _S_IFREG) {
-#else
-  if (!S_ISREG(sb.st_mode)) {
-#endif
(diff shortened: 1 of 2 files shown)


acozzette:
This seems to fail to build on Windows: `  T:\src\github\protobuf\protobuf\src\google\protobuf\compiler\importer.cc(498): error C3861: 'S_ISREG': identifier not found [T:\src\github\protobuf\protobuf\vcprojects\libprotobuf.vcxproj]`

junekhan:
@acozzette Hi, I'm sorry I don't have a Windows machine to test the building thoroughly. 
I've fixed it and tested it on an online VS C++ compiler. Please check my newest commit. If any problem, get back to me, please.

src/google/protobuf/compiler/importer.cc
```diff
@@ -490,6 +490,19 @@ io::ZeroCopyInputStream* DiskSourceTree::OpenVirtualFile(
 
 io::ZeroCopyInputStream* DiskSourceTree::OpenDiskFile(
     const std::string& filename) {
+  struct stat sb;
+  int ret = 0;
+  do {
+    ret = stat(filename.c_str(), &sb);
+  } while (ret != 0 && errno == EINTR);
+#ifdef _WIN32
```
acozzette, line 498: @junekhan Sorry for taking so long to get back to this, but I have just one comment. I was thinking you might be able to check whether the path is a directory by doing something like this:
`if (sb.st_mode & S_IFDIR)`
I believe that should work the same way on Linux, Mac, and Windows. Also that should allow things to work even with files like `/dev/stdin`, which could be useful even though they're not regular files.

junekhan: Good to know this universal statement. Thanks.
I checked the references of the function and was not sure of whether it's correct to open files other than regular files. But now I see your point and know it's useful.

acozzette:
Thanks @junekhan!

PR #7575, opened 2020-05-29
Cpp export include

dbeaudoin:
Added the capability for the generated cpp code to include the header file that define the export symbol. See this related google group post for reference: https://groups.google.com/forum/#!topic/protobuf/PDR1bqRazts

Also added #GOOGLE_PROTOBUF_NO_FINAL to generate cpp class that are not final. In certain controlled cases it is wanted to derive classes from generated protobuf classes to add system specific logic.

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -505,7 +505,11 @@
 #define PROTOBUF_ALIGNAS(byte_alignment) alignas(byte_alignment)
 #endif
 
+#ifndef GOOGLE_PROTOBUF_NO_FINAL
 #define PROTOBUF_FINAL final
+#else
+#define PROTOBUF_FINAL
+#endif
 
 #if defined(_MSC_VER)
 #define PROTOBUF_THREAD_LOCAL __declspec(thread)
--- a/src/google/protobuf/compiler/cpp/cpp_file.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_file.cc
@@ -1354,6 +1354,10 @@ void FileGenerator::GenerateLibraryIncludes(io::Printer* printer) {
   if (UseUnknownFieldSet(file_, options_) && !message_generators_.empty()) {
     IncludeFile("net/proto2/public/unknown_field_set.h", printer);
   }
+
+  if (!options_.dllexport_include.empty()) {
+    format("#include \"$1$\"\n", options_.dllexport_include);
+  }  
 }
 
 void FileGenerator::GenerateMetadataPragma(io::Printer* printer,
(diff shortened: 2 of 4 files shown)


dbeaudoin:
@googlebot I signed it!

src/google/protobuf/port_def.inc
```diff
@@ -505,7 +505,11 @@
 #define PROTOBUF_ALIGNAS(byte_alignment) alignas(byte_alignment)
 #endif
 
+#ifndef GOOGLE_PROTOBUF_NO_FINAL
```
acozzette, line 508: Unfortunately I don't think we will want to accept this change relating to the `final` keyword. It's important for generated messages to be final, because we need to be able to change things without worrying about breaking derived classes. Marking messages final also allows the compiler to generate better code by devirtualizing some member function calls.

dbeaudoin: Both valid points but this flag could be documented in a way that let the users know that they can use it at their own risk and that Google does not guarantee the backward / forward compatibility between releases. As this flag would not be defined by default, only users that would choose to do use derivation would do it knowing the potential impacts and in this case we let the final decision in the user hands. 

In our case the from the protobuf generated object because we add a layer that adds the capability to create observers to be notified when protobuf objects are received from the network. We could also use composition instead of derivation but derivation was giving us a cleaner / more direct  access to the underlying protobuf object and its attributes.

acozzette: I'm sorry but I checked with some other members of the protobuf team and the consensus is that we do not want to support inheriting from generated messages. If you really need to inherit from generated messages and can't easily refactor to avoid that, you can always make a small modification to port_def.inc in your local copy of the source.

dbeaudoin: No problem, thanks for looking into this with the rest of the team. I understand your reasons, they are valid ones! Modifying the .h is what we were doing so far but we were trying to avoid doing this minor change. We'll see what we do with this, maybe we continue this way or maybe I will simply change the object to use composition of derivation. I will roll back this change.

acozzette:
@dbeaudoin What kind of macro definition do you need to pull in from the separate header? The reason I ask is that we are usually somewhat hesitant to add new options since we then have to commit to maintaining them. Is there any way to make it work without the extra header?

dbeaudoin:
@acozzette Sorry for the delay before answering, I missed your last reply. Basically there is already the option dllexport_decl that lets specify what should be the export symbol to use in the generated class but there is not a clean way to specify in which file this export symbol is defined which makes the compilation of the generated code fail by default. It's almost like if the solution was only half implemented, we let the user define an external symbol to be injected in the code generator but we do not let the user specify where the generated code should look for that symbol. The work around is to force include a header file on the whole project using something like /FI on Visual Studio but it also means that this must be applied to any dependent projects.  This problem was described in an old protobuf forum entry: https://groups.google.com/forum/#!topic/protobuf/PDR1bqRazts this changes gives a solution to this issue.

acozzette:
@dbeaudoin Sorry for taking so long to get back to you. Is there any reason you can't just specify a macro in `dllexport_decl` and then define that macro on the compiler command line?

PR #7576, opened 2020-05-29
New Span-based serialization logic (followup for #7351)

jtattermusch:
Span-based parsing has been added by https://github.com/protocolbuffers/protobuf/pull/7351 , this PR adds the serialization counterpart, which allows serialization to `IBufferWriter` (which is much more GC friendly).

Based on https://github.com/protocolbuffers/protobuf/pull/7490

Update 2020-06-29:
The new benchmarks (as of commit 4f0afc7) are here: https://gist.github.com/jtattermusch/c9e5389a398f7c79710a9bf12aee215f
For both writing primitives and writing messages the performance is comparable (there are some significant improvement and some slight regressions) to the original state, which seems good enough (as this PR offers an API to serialize directly into a Span and/or to IBufferWriter, which are modern APIs allowing for much better GC efficiency).


JamesNK:
Is `IBufferWriter<byte>` supported yet? It is referenced but at the moment all tests write to `byte[]` or `Stream`.

jtattermusch:
> Is `IBufferWriter<byte>` supported yet? It is referenced but at the moment all tests write to `byte[]` or `Stream`.

I still need to expose the API and add tests, but vast majority of the internal implementation is already there.

jtattermusch:
I added two MessageExtensions:
`WriteTo(this IMessage message, IBufferWriter<byte> output)` and `WriteTo(this IMessage message, Span<byte> output)`. 

Also added test coverage for the WriteContext based serialization.

I still need to add some benchmarks.

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -145,6 +146,39 @@ public static ByteString ToByteString(this IMessage message)
             return ByteString.AttachBytes(message.ToByteArray());
         }
 
+        /// <summary>
+        /// Writes the given message data to the given buffer writer in protobuf encoding.
+        /// </summary>
+        /// <param name="message">The message to write to the stream.</param>
+        /// <param name="output">The stream to write to.</param>
+        public static void WriteTo(this IMessage message, IBufferWriter<byte> output)
```
JamesNK, line 154: [SecuritySafeCritical]

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -145,6 +146,39 @@ public static ByteString ToByteString(this IMessage message)
...
+        }
+
+        /// <summary>
+        /// Writes the given message data to the given span in protobuf encoding.
+        /// The size of the destination span needs to fit the serialized size
+        /// of the message exactly, otherwise an exception is thrown.
+        /// </summary>
+        /// <param name="message">The message to write to the stream.</param>
+        /// <param name="output">The span to write to. Size must match size of the message exactly.</param>
+        public static void WriteTo(this IMessage message, Span<byte> output)
```
JamesNK, line 173: [SecuritySafeCritical]

csharp/src/Google.Protobuf.Test/MessageParsingHelpers.cs
```diff
@@ -96,5 +102,26 @@ public static void AssertReadingMessage(MessageParser parser, byte[] bytes, Acti
...
+            int messageSize = message.CalculateSize(); 
+            Assert.AreEqual(message.CalculateSize(), bytes.Length);
+
+            // serialize using IBufferWriter and check it leads to the same output
+            var bufferWriter = new ArrayBufferWriter<byte>();
+            message.WriteTo(bufferWriter);
+            Assert.AreEqual(bytes, bufferWriter.WrittenSpan.ToArray());
+
+            // serialize into a single span and check it leads to the same output
+            var singleSpan = new Span<byte>(new byte[messageSize]);
```
JamesNK, line 120: You could replace this with `Span<byte> singleSpan = stackalloc byte[messageSize];` and test writing to a stack allocated array.

Probably better to test that out by itself. This method is used all over the place and there is a limit on how large you can stackalloc.

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -0,0 +1,556 @@
...
+            if (state.position == state.limit)
+            {
+                WriteBufferHelper.RefreshBuffer(ref buffer, ref state);
+            }
+
+            buffer[state.position++] = value;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void WriteRawByte(ref Span<byte> buffer, ref WriterInternalState state, uint value)
```
JamesNK, line 395: Can be removed. Not used

csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -469,128 +481,52 @@ public void WriteRawTag(byte b1, byte b2, byte b3, byte b4)
...
-
-            buffer[position++] = value;
+            var span = new Span<byte>(buffer);
+            WritingPrimitives.WriteRawByte(ref span, ref state, value);
         }
 
         internal void WriteRawByte(uint value)
         {
-            WriteRawByte((byte) value);
+            var span = new Span<byte>(buffer);
```
JamesNK, line 528: WriteRawByte can be removed. Not used

JamesNK:
Looks good as far!

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -145,6 +146,39 @@ public static ByteString ToByteString(this IMessage message)
...
+        public static void WriteTo(this IMessage message, IBufferWriter<byte> output)
+        {
+            ProtoPreconditions.CheckNotNull(message, nameof(message));
+            ProtoPreconditions.CheckNotNull(output, nameof(output));
+
+            WriteContext.Initialize(output, out WriteContext ctx);
+            WritingPrimitivesMessages.WriteRawMessage(ref ctx, message);
+            ctx.Flush();
+
+            // TODO: handling errors when IBufferWriter is used?
```
jtattermusch, line 163: @JamesNK  what is the best pattern for handling serialization exceptions while writing to an IBufferWriter?
I'm talking about a situation where I've got a buffer using writer.GetSpan() and then in the middle of the serialization I get a protobuf serialization error. It sounds like I might need a `finally` statement that invalidates the current buffer segment by calling `Advance(current position)`(otherwise there's a buffer leak potentially?). Or is this supposed to be solved on the higher level (i.e. protobuf serialization fails, so the owner of the IBufferWriter invalidates the entire IBufferWriter which also releases the current buffer segment
)?:. I'd expect there would be a best practice to follow in this case.

JamesNK: You don't need to do anything at this level.

For example:
1. You call GetSpan
2. An error occurs and is caught while serialization is still happening
3. Serialization resumes and GetSpan is called again

The second GetSpan call means the previous `Span<byte>` buffer is no longer active. (the second GetSpan could have returned the same segment of buffer memory, or it might not)

Or for example:
1. You call GetSpan
2. An error occurs and is not caught
3. The PipeWriter (which implements `IBufferWriter<byte>`) is marked as complete without Advance ever being called.

The buffer reserved by GetSpan is thrown away.

csharp/src/Google.Protobuf/WritingPrimitives.cs
jtattermusch: Done

csharp/src/Google.Protobuf.Test/MessageParsingHelpers.cs
jtattermusch: I made it test serialization to a stackalloc whenever the size is <256 bytes.

csharp/src/Google.Protobuf/CodedOutputStream.cs
jtattermusch: Done.

csharp/src/Google.Protobuf/MessageExtensions.cs
jtattermusch: Done.

csharp/src/Google.Protobuf/MessageExtensions.cs
jtattermusch: Done.

jtattermusch:
I addressed the comments, resolved several TODOs and I think this is very close to being mergeable.
I've also run most of the benchmarks and compared before and after performance and so far things are looking pretty good (there are basically no regressions and there are some improvements) - I will publish the full comparison later.

jtattermusch:
@haberman  this is now ready for review. @JamesNK will review the code from C# perspective, we just need a high-level protobuf review. Most of the ideas are analogous to what we've done in #7351, but this time it's a bit simpler (serialization is easier than parsing).

JamesNK:
It would be useful to have the equivalent of ParseMessagesBenchmark but for writing. That would show the end-to-end perf on complex objects.

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -0,0 +1,568 @@
...
+                WriteRawByte(ref buffer, ref state, (byte)value);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void WriteRawLittleEndian32(ref Span<byte> buffer, ref WriterInternalState state, uint value)
+        {
+            if (state.position + 4 > state.limit)
+            {
+                WriteRawByte(ref buffer, ref state, (byte)value);
```
JamesNK, line 356: If there is not enough room available, instead of writing individual bytes and checking for the size for each byte, have you considered refreshing the buffer so that there is room available. For example:
 https://github.com/JamesNK/protobuf/blob/155774996be69135c05b30960e6ef016c530894c/csharp/src/Google.Protobuf/CodedOutputWriter.cs#L478-L489

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -0,0 +1,568 @@
...
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 8));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 16));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 24));
+            }
+            else
+            {
+                buffer[state.position++] = ((byte)value);
```
JamesNK, line 363: There is no explicit check in this method that the buffer length is greater than the content being written. I wonder whether there are optimizations for bounds checking that could be made.

Low level IL optimization isn't an area I'm experienced with.

JamesNK, line 363: This might be faster:

`BinaryPrimitives.WriteUInt32LittleEndian(buffer.Slice(state.position), value);`

If it isn't we should at least comment on why it isn't used.

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -0,0 +1,568 @@
...
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 16));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 24));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 32));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 40));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 48));
+                WriteRawByte(ref buffer, ref state, (byte)(value >> 56));
+            }
+            else
+            {
+                buffer[state.position++] = ((byte)value);
```
JamesNK, line 386: BinaryPrimitives.WriteUInt64LittleEndian?

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -0,0 +1,568 @@
...
+#endif
+                }
+            }
+            else
+            {
+                // Opportunity for future optimization:
+                // Large strings that don't fit into the current buffer segment
+                // can probably be optimized by using Utf8Encoding.GetEncoder()
+                // but more benchmarks would need to be added as evidence.
+                byte[] bytes = Utf8Encoding.GetBytes(value);
```
JamesNK, line 207: https://github.com/JamesNK/protobuf/blob/155774996be69135c05b30960e6ef016c530894c/csharp/src/Google.Protobuf/CodedOutputWriter.cs#L264

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -0,0 +1,568 @@
...
+            WriteRawByte(ref buffer, ref state, b1);
+        }
+
+        /// <summary>
+        /// Writes the given two-byte tag directly to the stream.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void WriteRawTag(ref Span<byte> buffer, ref WriterInternalState state, byte b1, byte b2)
+        {
+            WriteRawByte(ref buffer, ref state, b1);
```
JamesNK, line 498: WriteRawTag methods can be optimized to write to the span directly if room is available.

https://github.com/JamesNK/protobuf/blob/155774996be69135c05b30960e6ef016c530894c/csharp/src/Google.Protobuf/CodedOutputWriter.cs#L478-L489

csharp/src/Google.Protobuf/WritingPrimitives.cs
jtattermusch: Thanks for the link. Yes, that's what I was referring to. 
I'd prefer to do that in a separate PR (it requires more testing and benchmarks).

csharp/src/Google.Protobuf/WritingPrimitives.cs
jtattermusch: Good point. Let's address/investigate in a followup PR? In this PR I'm happy with what we have as long as the benchmark don't show a performance regression.

csharp/src/Google.Protobuf/WritingPrimitives.cs
jtattermusch: I can look into that. Initially I tried to be conservative and I wanted to preserve the CodedOutputStream behavior (which always finishes the buffer all the way to its end before refreshing) as much as possible.
Maybe this could be a followup PR?

csharp/src/Google.Protobuf/WritingPrimitives.cs
jtattermusch: switched to WriteUInt32LittleEndian

csharp/src/Google.Protobuf/WritingPrimitives.cs
jtattermusch: benchmarks show it's actually slower, so added a comment.

jtattermusch:
Before and after benchmark results is here:
https://gist.github.com/jtattermusch/2d8cc4eab54a53de1079132495d3f55d

Summary 
- for WriteRawPrimitivesBenchmark, there are some primitives that are slighly slower, but most of them are faster and some of the much are much faster.
- for WriteMessageBenchmark, unfortunately it seems that writing of messages is actually quite a bit slower, so that's an area that still needs to be investigated (not sure what the problem is).

jtattermusch:
@JamesNK  can you please take a look at the WriteMessagesBenchmark results from my comment https://github.com/protocolbuffers/protobuf/pull/7576#issuecomment-642698060 and see if:
- you're getting the same results
- there are some optimizations I've missed and that would speed up WriteContext-based writing to an acceptable level.

JamesNK:
Hmmm, interesting. In my PR it was writing that ended up faster, and parsing was the same speed.

My WriteMessagesBenchmark results: https://gist.github.com/JamesNK/86d8707398326904dfbccd3744022960

One optimization is range checks - https://github.com/protocolbuffers/protobuf/pull/7576#discussion_r437119673. I don't think that explains the difference. My guess is it is probably something to do with copying structs.

jtattermusch:
@JamesNK  thanks for the results. It seems that on windows the ManyWrapperFields and Empty serialization has ~similar speed as before, but there's a big regression in terms of serializing ManyPrimitiveFieldsMessage.
On Linux the results have a similar trend but it isn't as clear-cut.

I guess it might have to do with passing the `ref buffer, ref state` to each of the writing primitive method (and each of the writing operations is very fast as it doesn't really do much) and the overhead builds up.

jtattermusch:
It turns out that the comparably bad performance with ManyPrimitiveFields message is mostly caused by using AggresiveInlining heavily for all the primitives. What I think happens is:
- the ManyPrimitiveFieldsMessage.InternalWriteTo method has a lot of invocations to primitives, and they all get inlined.
- the InternalWriteToMethod becomes too complex, which prevents the JIT from noticing the fact that most of the message fields are unused and only the few existing ones are going to be used. 
- as a result, checking whether the message fields are set takes much longer, which results in much slower serialization (where most time is spent checking whether the fields are set or not).

The solution is removing some of the AggresiveInlining annotations, but on the other hand that needs to be done carefully, because if too many are removed, the WriteRawPrimitivesBenchmarks results deteriorate.
So I think the resolution is finding the right balance between which AggresiveInlining annotations to keep and which ones to remove. Based on my experiments, we might not be able to reach 100% of the original performance in all cases unfortunately (but altogether this PR probably still worth accepting because it allows GC friendly serialization directly to the IBufferWriter with no additional copies).

JamesNK:
Good find.

I think the best improvement here is reducing the amount of code in the primitive methods, and the best way to do that is optimize for the fast path of there being enough data in the buffer.

See my comment here about removing the WriteRawByte calls and just having a single call to refresh the buffer if there is not enough room - https://github.com/protocolbuffers/protobuf/pull/7576#discussion_r437118669

What `CodedOutputWriter` does:
```cs
        internal void WriteRawLittleEndian32(uint value)
        {
            const int length = 4;

            Ensure(length);

            BinaryPrimitives.WriteUInt32LittleEndian(span.Slice(buffered), value);

            buffered += length;
        }
```

What WritingPrimitives.cs does today:
```cs
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static void WriteRawLittleEndian32(ref Span<byte> buffer, ref WriterInternalState state, uint value)
        {
            const int length = sizeof(uint);
            if (state.position + length > state.limit)
            {
                WriteRawByte(ref buffer, ref state, (byte)value);
                WriteRawByte(ref buffer, ref state, (byte)(value >> 8));
                WriteRawByte(ref buffer, ref state, (byte)(value >> 16));
                WriteRawByte(ref buffer, ref state, (byte)(value >> 24));
            }
            else
            {
                BinaryPrimitives.WriteUInt32LittleEndian(buffer.Slice(state.position), value);
                state.position += length;
            }
        }
```

Improving WritingPrimitives.cs:
```cs
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static void WriteRawLittleEndian32(ref Span<byte> buffer, ref WriterInternalState state, uint value)
        {
            const int length = sizeof(uint);
            if (state.position + length > state.limit)
            {
                // Rarely called. RefreshBuffer should have NoInlining on it.
                RefreshBuffer(ref buffer, ref state, requireLength: length);
            }
            
            BinaryPrimitives.WriteUInt32LittleEndian(buffer.Slice(state.position), value);
            state.position += length;
        }
```

You don't need to worry about using up the entire buffer. It shouldn't matter to the Stream or IBufferWriter that sometimes you are writing/advancing 4096 bytes, and other times writing/advancing 4091 bytes.

JamesNK:
Useful code from CodedOutputWriter - https://github.com/JamesNK/protobuf/blob/155774996be69135c05b30960e6ef016c530894c/csharp/src/Google.Protobuf/CodedOutputWriter.cs#L626-L655

JamesNK:
Restores pre-Span perf: https://github.com/protocolbuffers/protobuf/commit/337780c0e098185d4df76d498c8f435e105f8a56

Changes of note:
* Removed many inlining attributes. This was the primary performance boost
* Changed writer to refresh the buffer if there is not enough room, rather than write a byte at a time. To support this change a minimum buffer size on `CodedOutputStream` is required. I chose 128 bytes because this is the [minimum size in StreamWriter](https://github.com/dotnet/runtime/blob/b8632bb8e85e9c7a6f978cca49bc625b6f9444f1/src/libraries/System.Private.CoreLib/src/System/IO/StreamWriter.cs#L28).
* Benchmarks are run on netcoreapp3.1. It seemed strange that `BinaryPrimitives.WriteUInt64LittleEndian` was slower, and that appears fixed on 3.1.
* Added `IBufferWriter` benchmarks. Results are about the same as Span benchmarks which is expected.

New results:

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.900 (1909/November2018Update/19H2)
AMD Ryzen 7 1700, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=3.1.400-preview-015151
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


|                                                               Method | messageCount |          Mean |      Error |     StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------------------------------------------- |------------- |--------------:|-----------:|-----------:|-------:|------:|------:|----------:|
|                                 ManyWrapperFieldsMessage_ToByteArray |            ? |   1,133.82 ns |   6.206 ns |   5.805 ns | 0.0420 |     - |     - |     184 B |
|                    ManyWrapperFieldsMessage_WriteToCodedOutputStream |            ? |     628.18 ns |   5.790 ns |   5.133 ns | 0.0153 |     - |     - |      64 B |
|                                 ManyWrapperFieldsMessage_WriteToSpan |            ? |     589.05 ns |   2.809 ns |   2.627 ns |      - |     - |     - |         - |
|                         ManyWrapperFieldsMessage_WriteToBufferWriter |            ? |     603.41 ns |   2.488 ns |   2.078 ns |      - |     - |     - |         - |
|                               ManyPrimitiveFieldsMessage_ToByteArray |            ? |     487.09 ns |   2.783 ns |   2.467 ns | 0.0401 |     - |     - |     168 B |
|                  ManyPrimitiveFieldsMessage_WriteToCodedOutputStream |            ? |     251.36 ns |   3.167 ns |   2.807 ns | 0.0153 |     - |     - |      64 B |
|                               ManyPrimitiveFieldsMessage_WriteToSpan |            ? |     205.37 ns |   4.034 ns |   3.576 ns |      - |     - |     - |         - |
|                       ManyPrimitiveFieldsMessage_WriteToBufferWriter |            ? |     256.14 ns |   1.343 ns |   1.191 ns |      - |     - |     - |         - |
|                                             EmptyMessage_ToByteArray |            ? |      52.90 ns |   1.044 ns |   1.594 ns | 0.0210 |     - |     - |      88 B |
|                                EmptyMessage_WriteToCodedOutputStream |            ? |      45.48 ns |   0.634 ns |   0.593 ns | 0.0153 |     - |     - |      64 B |
|                                             EmptyMessage_WriteToSpan |            ? |      23.10 ns |   0.252 ns |   0.236 ns |      - |     - |     - |         - |
|   ManyWrapperFieldsMessage_WriteDelimitedMessagesToCodedOutputStream |           10 |  10,298.57 ns |  41.767 ns |  37.026 ns | 0.0153 |     - |     - |      64 B |
|                ManyWrapperFieldsMessage_WriteDelimitedMessagesToSpan |           10 |  10,020.72 ns |  44.361 ns |  41.495 ns |      - |     - |     - |         - |
| ManyPrimitiveFieldsMessage_WriteDelimitedMessagesToCodedOutputStream |           10 |   4,424.32 ns |  22.742 ns |  20.160 ns | 0.0153 |     - |     - |      64 B |
|              ManyPrimitiveFieldsMessage_WriteDelimitedMessagesToSpan |           10 |   3,969.32 ns |  34.041 ns |  31.842 ns |      - |     - |     - |         - |
|      ManyPrimitiveFieldsMessage_WriteDelimitedMessagesToBufferWriter |           10 |   3,955.47 ns |  18.120 ns |  15.131 ns |      - |     - |     - |         - |
|   ManyWrapperFieldsMessage_WriteDelimitedMessagesToCodedOutputStream |          100 | 102,826.84 ns | 515.522 ns | 482.220 ns |      - |     - |     - |      64 B |
|                ManyWrapperFieldsMessage_WriteDelimitedMessagesToSpan |          100 | 101,233.79 ns | 257.638 ns | 215.140 ns |      - |     - |     - |         - |
| ManyPrimitiveFieldsMessage_WriteDelimitedMessagesToCodedOutputStream |          100 |  44,027.71 ns | 212.396 ns | 188.283 ns |      - |     - |     - |      64 B |
|              ManyPrimitiveFieldsMessage_WriteDelimitedMessagesToSpan |          100 |  40,416.43 ns | 166.844 ns | 130.261 ns |      - |     - |     - |         - |
|      ManyPrimitiveFieldsMessage_WriteDelimitedMessagesToBufferWriter |          100 |  37,822.49 ns | 110.558 ns |  92.321 ns |      - |     - |     - |         - |
```

JamesNK:
Optimize writing strings across buffers: https://github.com/JamesNK/protobuf/commit/99e830bf6f020da98620c46beffa052295c08504

Before:
```
|                                               Method | BytesToWrite | encodedSize |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------------------- |------------- |------------ |-----------:|----------:|----------:|-------:|------:|------:|----------:|
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |           1 | 150.550 us | 1.6307 us | 1.5254 us |      - |     - |     - |       2 B |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |           4 |  49.011 us | 0.6282 us | 0.5569 us |      - |     - |     - |         - |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |          10 |  26.207 us | 0.4905 us | 0.4348 us | 0.0610 |     - |     - |     320 B |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |         105 |  11.951 us | 0.0635 us | 0.0530 us | 0.2747 |     - |     - |    1152 B |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |       10080 |   3.178 us | 0.0628 us | 0.0616 us | 2.4071 |     - |     - |   10104 B |
```
After:
```
|                                               Method | BytesToWrite | encodedSize |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------------------- |------------- |------------ |-----------:|----------:|----------:|-------:|------:|------:|----------:|
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |           1 | 161.463 us | 1.2570 us | 1.1758 us |      - |     - |     - |         - |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |           4 |  49.170 us | 0.7292 us | 0.9223 us |      - |     - |     - |         - |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |          10 |  23.225 us | 0.1782 us | 0.1667 us |      - |     - |     - |         - |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |         105 |   9.763 us | 0.0765 us | 0.0679 us |      - |     - |     - |         - |
| WriteString_WriteContextBufferWriter_LimitBufferSize |        10080 |       10080 |   2.658 us | 0.0195 us | 0.0173 us | 0.0381 |     - |     - |     160 B |
```

jtattermusch:
With my few last commits (thanks @JamesNK for the ideas), the writing performance for messages now seems ~at par with the previous state (= no regressions) so I think this PR is ready for final review.
@haberman  can you please review?

haberman:
@gerben-s Can you review please?

jtattermusch:
The new benchmarks (as of commit 4f0afc7) are here: https://gist.github.com/jtattermusch/c9e5389a398f7c79710a9bf12aee215f
For both writing primitives and writing messages the performance is comparable (there are some significant improvement and some slight regressions) to the original state, which seems good enough (as this PR offers an API to serialize directly into a Span and/or to IBufferWriter, which are modern APIs allowing for much better GC efficiency).

MaximGurschi:
Is there a timeline, for when the nuget package for C# will be released? Eagerly waiting to be able to use Span with the serialization/deserialization methods you have developed :)

jtattermusch:
> Is there a timeline, for when the nuget package for C# will be released? Eagerly waiting to be able to use Span with the serialization/deserialization methods you have developed :)

If everything goes as planned, it should be available over the course of the next few weeks.

PR #7581, opened 2020-06-01
chk nullptr

YourButterfly:


haberman:
It seems it would be better to put the check earlier:

1. we should refuse to load any descriptor that has `map_entry=true` but doesn't have proper `key` and `value` fields. This check could live in `descriptor.cc`.
2. we should error out if anybody sets this option in a `.proto` file directly. We already have docs stating that proto files should not set this option directly:

https://github.com/protocolbuffers/protobuf/blob/e492e5a4ef16f59010283befbde6112f1995fa0f/src/google/protobuf/descriptor.proto#L517-L519

Also there are new accessors for this. We should migrate the code to use `map_key()` and `map_value()`. These accessors already do a check in debug mode:

https://github.com/protocolbuffers/protobuf/blob/e492e5a4ef16f59010283befbde6112f1995fa0f/src/google/protobuf/descriptor.h#L493-L499

https://github.com/protocolbuffers/protobuf/blob/e492e5a4ef16f59010283befbde6112f1995fa0f/src/google/protobuf/descriptor.cc#L1712-L1722

YourButterfly:
It seems that you are going to use new accessors for this. Please tell me, when you do with this issue.

acozzette:
I just sent out PR #9099 to adopt the new `map_key()` and `map_value()` accessors, so I will close this pull requeset.

PR #7588, opened 2020-06-02
Use packed encoding by default for proto3 repeated primitive fields

### LINKED ISSUE #6045: Repeated Enum is serialized differently in PHP serializeToString() as compared to other languages.
Version: 3.6.1
Language: PHP

Mac OS Mojave, version: 10.14.1

PHP5.6 runtime

**What did you do?**
I was trying to serialize a message containing "repeated enum type" field using the serializeToString() method and it gave me a different encoding in PHP when compared to other languages like Java and C#.

Example 

```
message Example {
   repeated Filed field = 1
   enum Field {
       Value1 = 1;
       Value2 = 2;
   }
} 
```
Now if I use the protobuf message generated class

```
$example = new Example();
$fields =array();
$fields[] = Example/Field::Value1;
$fields[] = Example/Field::Value2;
$example->setField($fields);
echo $example->serializeToString();

```
`10 2 1 2 `
For C# and Java it was serialized this way.

**What did you see instead?**
`8 1 8 2`
For PHP

TeBoring:
Fixes #6045


haberman:
Sorry I missed this before. It looks like this was subsequently applied in https://github.com/protocolbuffers/protobuf/pull/7740.

Issue #7591, opened 2020-06-02
v3.12 field presence does not support Javascript


skearnes:
The [documentation](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md#javascript-example) for field presence includes a Javascript example, and there are no indications in the [announcement](https://github.com/protocolbuffers/protobuf/issues/1606#issuecomment-618687169) or [release notes](https://github.com/protocolbuffers/protobuf/releases/tag/v3.12.0) that suggest Javascript is not supported.

However, Javascript field presence is not _actually_ supported in v3.12. Since this is blocking for our use of proto3 (see https://github.com/Open-Reaction-Database/ord-schema/pull/176), it would be nice to have an ETA so we can decide whether to downgrade to proto2 instead of waiting.

**What version of protobuf and what language are you using?**
Version: v3.12.2
Language: Javascript

**What did you do?**
1. Download the protoc-3.12.2-linux-x86_64.zip release and unzip
2. Run `protoc` with `--experimental_allow_proto3_optional` and `--js_out=...`

**What did you expect to see?**

Javascript proto wrappers that support `optional` fields.

**What did you see instead?**

```
proto/reaction.proto: is a proto3 file that contains optional fields, but code 
generator --js_out hasn't been updated to support optional fields in proto3. 
Please ask the owner of this code generator to support proto3 optional.
```

PR #7592, opened 2020-06-03
Enable experimental presence detection in JS.

antonkast-google:
Migrated from cl/303184408. "make check" passes. Fixes Issue #7591.


skearnes:
Note that this fixes #7591

PR #7596, opened 2020-06-03
annotate ByteString.Span and ByteString.Memory as SecuritySafeCritical

### LINKED ISSUE #7139: Can't call ByteString.Span on net45
Version: 3.11.2
Language: C#

Windows

net45

**What did you do?**
Steps to reproduce the behavior: run 32-bit console application, target net45.
```cs
using System;
using Google.Protobuf;

namespace test1
{
    class Program
    {
        static void Main(string[] args)
        {
            byte[] a = new byte[] { 1, 2, 3 };
            var b = ByteString.CopyFrom(a);
            var s = b.Span;
            Console.WriteLine("length " + s.Length);
        }
    }
}
```

success

**What did you see instead?**

```
Exception thrown: 'System.TypeAccessException' in test1.exe
An unhandled exception of type 'System.TypeAccessException' occurred in test1.exe
Attempt by security transparent method 'Google.Protobuf.ByteString.get_Span()' to access security critical type 'System.ReadOnlySpan`1<System.Byte>' failed.

Assembly 'Google.Protobuf, Version=3.11.2.0, Culture=neutral, PublicKeyToken=a7d26565bac4d604' is marked with the AllowPartiallyTrustedCallersAttribute, and uses the level 2 security transparency model.  Level 2 transparency causes all methods in AllowPartiallyTrustedCallers assemblies to become security transparent by default, which may be the cause of this exception.
```

jtattermusch:
the `.Span` and `.Memory` properties weren't properly annotated as `[SecuritySafeCritical]`.
Also adding tests to verify the fix.

Fixes #7139

### DIFF
--- a/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
+++ b/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
@@ -233,5 +233,21 @@ namespace Google.Protobuf
             ByteString b2 = ByteString.CopyFrom(200, 1, 2, 3, 4);
             Assert.AreNotEqual(b1.GetHashCode(), b2.GetHashCode());
         }
+
+        [Test]
+        public void GetContentsAsReadOnlySpan()
+        {
+            var byteString = ByteString.CopyFrom(1, 2, 3, 4, 5);
+            var copied = byteString.Span.ToArray();
+            CollectionAssert.AreEqual(byteString, copied);
+        }
+
+        [Test]
+        public void GetContentsAsReadOnlyMemory()
+        {
+            var byteString = ByteString.CopyFrom(1, 2, 3, 4, 5);
+            var copied = byteString.Memory.ToArray();
+            CollectionAssert.AreEqual(byteString, copied);
+        }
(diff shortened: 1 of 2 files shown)


jtattermusch:
@JamesNK  can you please review?
 
CC @jskeet

jskeet:
I'm not familiar with SecuritySafeCritical - although I suspect I've looked at it before. What was the problem before adding the attribute to these properties? Does it have to be added to the get specifically rather than annotating the whole (get-only) property?

jtattermusch:
> I'm not familiar with SecuritySafeCritical - although I suspect I've looked at it before. What was the problem before adding the attribute to these properties? Does it have to be added to the get specifically rather than annotating the whole (get-only) property?

The problem was that user code on net45+ wasn't able to call those properties unless it itself added these annotations.
As a demonstration, if you remove the annotations from the properties, the tests I added will fail on desktop .NET. (with .NET core these annotations no longer have any effect).

anon17:
The bug is #7139

PR #7601, opened 2020-06-05
Fix JSON parsing and formatting of NullValue in C#

jskeet:
The NullValue enum is typically used as part of the Value well-known type, but can be used elsewhere. In previous code, a NullValue field (other than in Value) would result in a JSON value of the string "NULL_VALUE"; it should instead be the null literal.
When parsing, we still accept "NULL_VALUE" as a valid value for NullValue - this new code being unable to parse old data, and comes for free anyway.


jskeet:
@haberman: While new code will be able to parse the old JSON, the reverse will not be true. I still think it's worth doing this, as other languages (at least some other languages!) are already serializing NullValue as the null literal.
Happy to chat more about the compatibility concerns.

haberman:
Wow, I don't know if the spec really says what (if anything) should happen when a `google.protobuf.NullValue` is used outside of a `google.protobuf.Value`. I think we should add a conformance test for this and make sure we're on the same page about the intended behavior.

I have no idea what most implementations will do...

jskeet:
Completely agree that we should have a conformance test. It'll be interesting to see what different languages do at the moment.

The reason I've seen this is due to the JSON provided in Google Cloud Function events for Firestore. [Firestore's Value message](https://github.com/googleapis/googleapis/blob/cd804bab06e46dd1a4f16c32155fd3cddb931b52/google/firestore/v1/document.proto#L82) uses NullValue, and the JSON I'm getting includes `{ "nullValue": null }`.

Can I leave adding the conformance test to you? I'm unfamiliar with the infrastructure there. I'm happy to wait on this PR until the conformance test shows the lay of the land.

anandolee:
Though it is not in our proto3 spec document, but the struct.proto has mentioned JSON format for NullValue is null:
https://github.com/protocolbuffers/protobuf/blob/5b232b8ecbce13286be09e703997e887ae0d464d/src/google/protobuf/struct.proto#L83

I think this is enough to change the behavior to follow the struct.proto.  Java and cpp is already print null when used in other oneofs other than Value. Python print "NULL_VALUE", I will fix that.

Also I'd like to mention that when google.protobuf.NullValue is a normal field not oneof, it will not be presented in proto3 json because it equals to default. I will add conformance test for both oneof and normal field for proto2/proto3.

jskeet:
@anandolee: Great. Am I right in saying it's best for me to wait for you to add the conformance test, then we can see whether this PR passes it?

anandolee:
The conformance test's CL is pending. I added more people for the decision. I think once everyone agreed on the decision, you can go ahead submit this PR as long as it matches the decision and have similar tests.

jskeet:
Will double-check the tests on Monday, and merge if it looks like it's doing everything it should.

csharp/protos/unittest_issues.proto
```diff
@@ -137,4 +139,11 @@ message OneofMerging {
...
-}
\ No newline at end of file
+}
+
+message NullValueOutsideStruct {
+  oneof value {
+    string string_value = 1;
+    google.protobuf.NullValue null_value = 2;
+  }
+}
```
anandolee, line 149: Also add the test when google.protobuf.NullValue is a normal field(not oneof field), and make sure nothing is printed in this situation

jskeet: Will do - although presumably it *should* be formatted when FormatDefaultValues is set to true... at which point should it format as `null` or `"NULL_VALUE"`? I'd assume `null`.

jskeet: Done - see the second commit in the PR.

jskeet:
Merging - Jie has reviewed from Protobuf perspective.

PR #7614, opened 2020-06-11
New PHP C Extension based on upb_msg

haberman:
This is a rewrite of the C extension for PHP. A bit of code is reused in `names.c`/`names.h`, but most of the rest of the code is new.

Here is a summary of the design (a more extensive design doc is forthcoming):

- All protobuf data is stored exclusively in the upb layer, through the types `upb_msg`, `upb_array`, and `upb_map`.
- Each upb object (`upb_msg`, `upb_array`, or `upb_map`) is owned by a `upb_arena`. When the user creates a link between objects in different arenas, we fuse the arenas together so they have the same lifetime.
- All encode/decode operations to binary/JSON are implemented exclusively in upb, without touching any PHP objects.
- PHP wrapper objects for message/map/arena are created on-demand.
- Wrapper objects are cached in a global map so that repeatedly accessing the same message will consistently return the same wrapper object. This map is weak so wrapper objects can be deleted when they are no longer used (which will release the reference on the arena object).
- PHP code for well-known types (`Any`, `Struct`, etc) is bundled inside the extension using `make-preload.py` and registered via autoload. This means we don't have to implement these types again in C.


haberman:
Note that this adds the new extension alongside the existing one for the moment. Once it is validated, we can remove the old extension and replace it with the new one.

haberman:
@dlj-NaN @TeBoring I think this is ready for review.

There are a few Kokoro issues I am still working out, but the code is working in Ads tests.

php/ext/google/protobuf2/message.c
```diff
@@ -0,0 +1,841 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "message.h"
+
+#include <inttypes.h>
+#include <php.h>
+#include <stdlib.h>
+
+// Must be last.
```
dlj-NaN, line 37: (As above: confusing comment?)

php/ext/google/protobuf2/names.c
```diff
@@ -0,0 +1,211 @@
...
+    "for",        "foreach",    "function",     "global",       "goto",
+    "if",         "implements", "include",      "include_once", "instanceof",
+    "insteadof",  "interface",  "isset",        "list",         "namespace",
+    "new",        "or",         "print",        "private",      "protected",
+    "public",     "require",    "require_once", "return",       "static",
+    "switch",     "throw",      "trait",        "try",          "unset",
+    "use",        "var",        "while",        "xor",          "int",
+    "float",      "bool",       "string",       "true",         "false",
+    "null",       "void",       "iterable"};
+const int kReservedNamesSize = 73;
```
dlj-NaN, line 89: [Forward-compat] What would you think about:
```
const int kReservedNamesSize = (sizeof(kReservedNamesSize)) / (sizeof(*kReservedNamesSize));
```
(extra parens to work around the strictest warning flags I can think of...)

php/ext/google/protobuf2/names.c
```diff
@@ -0,0 +1,211 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+static bool is_reserved(const char *segment, int length) {
+  bool result;
+  char* lower = malloc(length + 1);
+  memset(lower, 0, length + 1);
```
dlj-NaN, line 103: [Consistency] The PHP docs mention `calloc` quite often... do you think switching to `calloc` would be more idiomatic? (I personally think what's here is fine, though.)

php/ext/google/protobuf2/names.c
```diff
@@ -0,0 +1,211 @@
...
+      } else {
+        stringsink_string(classname, "PB", 2);
+      }
+    }
+  }
+}
+
+static void fill_segment(const char *segment, int length,
+                         stringsink *classname, bool use_camel) {
+  if (use_camel && (segment[0] < 'A' || segment[0] > 'Z')) {
```
dlj-NaN, line 136: [Consistency] Since the code above uses `tolower`, would a `toupper` call be reasonable?
```
if (use_camel /* && islower(segment[0]) */) {
  char first = (char)toupper(segment[0]);
  ...
```

php/ext/google/protobuf2/php-upb.h
```diff
@@ -0,0 +1,3885 @@
...
+*
+*   // Can be omitted for .c files, required for .h.
+*   #include "upb/port_undef.inc"
+*
+* This file is private and must not be included by users!
+*/
+#include <stdint.h>
+#include <stddef.h>
+
+#if UINTPTR_MAX == 0xffffffff
```
dlj-NaN, line 27: [Portability] It may be cleaner to check `sizeof(uintptr_t) == sizeof(uint32_t)`. Alternatively, it might be good to add `ul` to the constant.

php/ext/google/protobuf2/php-upb.h
```diff
@@ -0,0 +1,3885 @@
...
+#endif  /* UPB_ENCODE_H_ */
+/* This file was generated by upbc (the upb compiler) from the input
+ * file:
+ *
+ *     google/protobuf/descriptor.proto
+ *
+ * Do not edit -- your changes will be discarded when the file is
+ * regenerated. */
+
+#ifndef GOOGLE_PROTOBUF_DESCRIPTOR_PROTO_UPB_H_
```
dlj-NaN, line 1445: Would it be safe to drop these extra guard macros, or maybe leave these as separate files? (Or is there a reason they should stay?)

(I'm only mentioning because multiple guards are slower, e.g. on Clang, as compared to multiple files. It probably doesn't matter here, but I just wanted to double-check the reasoning.)

php/ext/google/protobuf2/protobuf.c
```diff
@@ -0,0 +1,346 @@
...
+  HashTable name_enum_cache;
+ZEND_END_MODULE_GLOBALS(protobuf)
+
+ZEND_DECLARE_MODULE_GLOBALS(protobuf)
+
+const zval *get_generated_pool() {
+  return &PROTOBUF_G(generated_pool);
+}
+
+//  * PHP_GINIT_FUNCTION(protobuf) / PHP_GSHUTDOWN_FUNCTION(protobuf)
```
dlj-NaN, line 86: [Minor request] Would it make sense to explicitly state this is a PHP module, and not a Zend extension? (It does follow from the lifetime sequence below, though.)

php/ext/google/protobuf2/array.c
```diff
@@ -0,0 +1,597 @@
...
+  }
+
+  if (!Convert_PhpToUpb(val, &msgval, intern->type, intern->desc, arena)) {
+    return;
+  }
+
+  if (index == size) {
+    upb_array_append(intern->array, msgval, Arena_Get(&intern->arena));
+  } else {
+    upb_array_set(intern->array, index, msgval);
```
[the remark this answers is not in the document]
haberman: Yes, good catch. I added a test.

php/ext/google/protobuf2/bundled_php.h
```diff
@@ -0,0 +1,41 @@
...
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef PHP_PROTOBUF_BUNDLED_PHP_H_
+#define PHP_PROTOBUF_BUNDLED_PHP_H_
+
+typedef struct {
```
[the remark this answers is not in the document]
haberman: Done.

php/ext/google/protobuf2/convert.h
```diff
@@ -0,0 +1,73 @@
...
+// Converts |php_val| to an int64_t. Returns false if the value cannot be
+// converted.
+bool Convert_PhpToInt64(zval *php_val, int64_t *i64);
+
+// Converts |php_val| to a upb_msgval according to |type|. If type is
+// UPB_TYPE_MESSAGE, then |desc| must be the Descriptor for this message type.
+// If type is string, message, or bytes, then |arena| will be used to copy
+// string data or fuse this arena to the given message's arena.
+bool Convert_PhpToUpb(zval *php_val, upb_msgval *upb_val, upb_fieldtype_t type,
+                    const Descriptor *desc, upb_arena *arena);
```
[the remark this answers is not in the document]
haberman: Done.

php/ext/google/protobuf2/convert.c
```diff
@@ -0,0 +1,468 @@
...
+  CASE(UINT32,   UINT32);
+  CASE(UINT64,   UINT64);
+  CASE(SINT32,   INT32);
+  CASE(SINT64,   INT64);
+  CASE(FIXED32,  UINT32);
+  CASE(FIXED64,  UINT64);
+  CASE(SFIXED32, INT32);
+  CASE(SFIXED64, INT64);
+
+#undef CONVERT
```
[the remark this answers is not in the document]
haberman: Done.

php/ext/google/protobuf2/convert.c
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  zend_error(E_ERROR, "Unknown field type.");
+  return 0;
+}
+
+static bool buftouint64(const char *ptr, const char *end, uint64_t *val) {
+  uint64_t u64 = 0;
+  while (ptr < end) {
+    unsigned ch = *ptr - '0';
```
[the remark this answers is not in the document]
haberman: Done.

php/ext/google/protobuf2/names.c
haberman: Done.

php/ext/google/protobuf2/names.c
haberman: I think I'm more inclined to change the other to *not* use `tolower()`, since its behavior is potentially affected by locales. I changed `tolower()` instead.

php/ext/google/protobuf2/php-upb.h
haberman: > It may be cleaner to check sizeof(uintptr_t) == sizeof(uint32_t).

That doesn't work because we need to resolve this in the preprocessor, and `sizeof()` isn't evaluated until later.

> Alternatively, it might be good to add ul to the constant.

I don't think this is necessary, the preprocessor will treat all integer constants as either `intmax_t` or `uintmax_t`, depending on whether they are signed or unsigned.

php/ext/google/protobuf2/php-upb.h
haberman: The main reasoning is convenience. The `upb.h`/`upb.c` files are copied from the upb repo, and copying one file is simpler than copying N (and keeping `config.m4` up to date as the list of files in upb changes).

It would be possible to change [the amalgamation script](https://github.com/protocolbuffers/upb/blob/master/tools/amalgamate.py) to drop the extra guard macros at amalgamation time, but it doesn't seem worth the complexity and effort.

php/ext/google/protobuf2/protobuf.c
haberman: Done.

AFAIK the terms "PHP Module" and "PHP Extension" are interchangeable, as are "Zend Module" and "Zend Extension." I used "PHP Extension" and "Zend Extension".

php/ext/google/protobuf2/convert.c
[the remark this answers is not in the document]
haberman: That is a cool tool!

I'm reluctant to add that check, as the code is currently correct and the rewritten version generates notably worse code on Clang: https://godbolt.org/z/ezAYiA

The cast through int is intended. I'm happy to make the casts explicit if it helps clarify the algorithm.

php/ext/google/protobuf2/def.h
```diff
@@ -0,0 +1,68 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef PHP_PROTOBUF_DEF_H_
+#define PHP_PROTOBUF_DEF_H_
+
+#include <php.h>
+
+#include "php-upb.h"
+
+void Def_ModuleInit();
```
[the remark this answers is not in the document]
haberman: I added a comment.

The source files and modules are named after the upb concepts (array, map, def) instead of the protobuf concepts (RepeatedField, MapField, Descriptor), mainly just because the upb names are shorter.

But I guess file/module names are not typed too often, so I'd be happy to rename these if it helps understanding. Maybe in a follow-up PR?

php/ext/google/protobuf2/convert.c
```diff
@@ -0,0 +1,468 @@
...
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "convert.h"
+
+#include <php.h>
+
+// Must be last.
```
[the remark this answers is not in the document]
haberman: I added extra comments for this.

php/ext/google/protobuf2/def.c
```diff
@@ -0,0 +1,1088 @@
...
+
+    // Prevent this from ever being collected (within a request).
+    GC_ADDREF(&ret->std);
+
+    ZVAL_OBJ(val, &ret->std);
+  }
+}
+
+void EnumDescriptor_FromEnumDef(zval *val, const upb_enumdef *m) {
+  if (m) {
```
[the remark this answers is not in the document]
haberman: I turned around the conditions so the null check is first.

I still left the other part inside an `else` because I'm coding in C style where declarations have to be at the top of a block. I want to avoid declaring the variables once and then initializing them after the NULL check.

php/ext/google/protobuf2/message.c
haberman: Done.

php/ext/google/protobuf2/names.c
haberman: Looking again it seemed even easier to just NULL-terminate the list instead.

php/ext/google/protobuf2/protobuf.c
dlj-NaN: Ah, interesting. I was basing my comment on this:
http://www.phpinternalsbook.com/php7/extensions_design/zend_extensions.html#on-differences-between-php-and-zend-extensions

php/ext/google/protobuf2/php-upb.h
dlj-NaN: (That actually does seem like a good fix... but fine to defer.)

php/ext/google/protobuf2/names.c
dlj-NaN: Even better, IMO.

php/ext/google/protobuf2/php-upb.h
dlj-NaN: As usual, I'm blind. :-)

I was mostly trying to preserve signedness to match the internals (and clarify there is no signed extension), e.g.:
https://github.com/llvm/llvm-project/blob/38b39c3/clang/test/Preprocessor/init.c#L6174
(which is also in clang -E -dM)

But this should be fine.

php/ext/google/protobuf2/arena.h
```diff
@@ -0,0 +1,47 @@
...
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef PHP_PROTOBUF_ARENA_H_
```
TeBoring, line 31: Does generated code need Arena?

haberman: No, arenas are not exposed to generated code. They are totally internal to the C extension.

php/ext/google/protobuf2/make-preload.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/python
```
TeBoring, line 1: For distributing on pecl, should the bundle_php.c also be included?

php/ext/google/protobuf2/protobuf.h
```diff
@@ -0,0 +1,89 @@
...
+
+#if PHP_VERSION_ID < 70300
+#define GC_ADDREF(h) ++GC_REFCOUNT(h)
+#define GC_DELREF(h) --GC_REFCOUNT(h)
+#endif
+
+// ptr -> PHP object cache. This is a weak map that caches lazily-created
+// wrapper objects around upb types:
+//  * upb_msg* -> Message
+//  * upb_array* -> RepeatedFIeld
```
TeBoring, line 49: RepeatedField

php/ext/google/protobuf2/arena.h
TeBoring: Then why a php wrapper for Arena is needed?

php/ext/google/protobuf2/protobuf.c
```diff
@@ -0,0 +1,349 @@
...
+//      GINIT(<This Thread Globals>)
+//      # A non-main thread will only receive requests when using a threaded
+//      # MPM with Apache
+//      foreach request:
+//        RINIT
+//          # Request is processed here.
+//        RSHUTDOWN
+//      GSHUTDOWN(<This Thread Globals>)
+//
+// That said, it appears that few people use threads with PHP:
```
TeBoring, line 130: Nice investigation!

php/ext/google/protobuf2/protobuf.c
```diff
@@ -0,0 +1,349 @@
...
+  NULL,
+  protobuf_deps,
+  "protobuf",               // extension name
+  protobuf_functions,       // function list
+  PHP_MINIT(protobuf),      // process startup
+  PHP_MSHUTDOWN(protobuf),  // process shutdown
+  PHP_RINIT(protobuf),      // request shutdown
+  PHP_RSHUTDOWN(protobuf),  // request shutdown
+  NULL,                     // extension info
+  "3.13.0",                 // extension version
```
TeBoring, line 341: Our script to update version number may need to update.

php/ext/google/protobuf2/arena.h
haberman: Because the Arena wrapper needs to participate in PHP memory management. Every PHP object that references the upb data takes a PHP ref on the Arena wrapper object. When no PHP wrappers are left to reference the Arena, the PHP arena object deletes the `upb_arena`, which frees the underlying data from `upb_msg`, `upb_array`, etc.

php/ext/google/protobuf2/protobuf.h
haberman: Done.

php/ext/google/protobuf2/protobuf.c
haberman: Thanks!

php/ext/google/protobuf2/make-preload.py
haberman: Yes. Where is the code that builds the pecl distribution?

php/ext/google/protobuf2/convert.h
```diff
@@ -0,0 +1,73 @@
...
+#include <php.h>
+
+#include "php-upb.h"
+#include "def.h"
+
+upb_fieldtype_t pbphp_dtype_to_type(upb_descriptortype_t type);
+
+// Converts |php_val| to an int64_t. Returns false if the value cannot be
+// converted.
+bool Convert_PhpToInt64(zval *php_val, int64_t *i64);
```
TeBoring, line 43: Should this be a const pointer?

php/ext/google/protobuf2/arena.h
TeBoring: Because it borrows php's refcount system?

php/ext/google/protobuf2/message.c
```diff
@@ -0,0 +1,841 @@
...
+
+  if (Z_TYPE_P(val) == IS_NULL) {
+    *msg = NULL;
+    return true;
+  }
+
+  if (Z_TYPE_P(val) == IS_OBJECT &&
+      instanceof_function(Z_OBJCE_P(val), desc->class_entry)) {
+    Message *intern = (Message*)Z_OBJ_P(val);
+    upb_arena_fuse(arena, Arena_Get(&intern->arena));
```
TeBoring, line 261: 1) arena1 and arena2 fuse and arena 1 is the lead
2) arena3 and arena4 fuse and arena3 is the lead
3) arena1 and arena3 fuse and arena 1 is the lead
Does arena4 need 2 hop (to find arena1) every time it looks for lead?

haberman: Discussed offline, I think the existing code is correct and efficient.

php/ext/google/protobuf2/convert.h
haberman: Done.

PR #7620, opened 2020-06-11
Add imenu index support

yuezhu:
This adds ability for Emacs imenu to index types defined in a protobuf file. For example, [addressbook.proto](https://github.com/yuezhu/protobuf/blob/de6cde0c92f8f7da47d62dcfb39258fe2bf55c27/examples/addressbook.proto) in the example directory would have these imenu items:

image


yuezhu:
@googlebot I signed it!

acozzette:
Thanks, @yuezhu.

Issue #7447, opened 2020-04-30
PHP messages should not use deprecated message types for internal checks


jdpedrie:
For historical reasons, nested messages generate two files in PHP.

For instance, given the following proto:

```
message foo {
	message bar {}
}
```

Protoc will generate the following files, the last of which contains a deprecation warning and alias for backwards compatibility. Users are directed to use `\Foo\Bar`.

```
./Foo.php
./Foo/Bar.php
./Foo_Bar.php
```

Messages internally still use the deprecated class name for type checking (see [here](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/php/php_generator.cc#L745-L751) for template source).

For instance, given the example above, the method `Foo::setBar($var)` would contain the following:

```php
public function setBar($var)
{
	GPBUtil::checkMessage($var, \Foo_Bar::class);
	$this->bar = $var;

	return $this;
}
```

This means that users have no means of preventing the library from raising `E_USER_DEPRECATED` errors. When users have a custom error handler configured, unless `E_USER_DEPRECATED` is explicitly excluded, those deprecation warnings are transformed to exceptions. This has proven to be a point of friction for a large number of users.

Since `\Foo_Bar` and `\Foo\Bar` are interchangable, there should be no regression or user impact in replacing the deprecated message name in type check calls with the new name.

haberman:
This was fixed in https://github.com/protocolbuffers/protobuf/pull/7695/commits/77edcb3f3d0a1a7fa57fc293b1b91a2435981b95 as part of https://github.com/protocolbuffers/protobuf/pull/7695.

PR #4741, opened 2018-06-07
Use legacy name in php runtime

TeBoring:
Old generated code cannot work with new runtime, because the new runtime
assumes new class name for nested message. For details see #4738.

Issue #7525, opened 2020-05-17
PHP: Next version of protobuf will drop C extension support for PHP 5.


haberman:
This is a heads-up to anybody still on PHP 5 that Protobuf 3.13.0 will drop support for PHP 5 for the C extension.

The pure-PHP implementation of protobuf will be unaffected, and PHP 5 users can continue to use this version.

PR #7629, opened 2020-06-15
[PHP] Fix GPBUtil::getFullClassName() argument order

davidgrayston:
### Fixed
- `Google\Protobuf\Internal\Descriptor::buildFromProto()` now passes arguments to `Google\Protobuf\Internal\GPBUtil::getFullClassName()` in the correct order, preventing deprecation notices for legacy classes.

#### Notes:
- Issue was discovered due to unexpected deprecation notices for legacy classes.
- #7447 is a similar issue, but unrelated to this PR.
- [`Google\Protobuf\Internal\EnumDescriptor::buildFromProto()`](https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/EnumDescriptor.php#L97) already passes arguments in the correct order.
- Introduced in #4741

### DIFF
--- a/php/src/Google/Protobuf/Internal/Descriptor.php
+++ b/php/src/Google/Protobuf/Internal/Descriptor.php
@@ -185,8 +185,8 @@ class Descriptor
             $containing,
             $file_proto,
             $message_name_without_package,
-            $legacy_classname,
             $classname,
+            $legacy_classname,
             $fullname);
         $desc->setFullName($fullname);
         $desc->setClass($classname);


davidgrayston:
@googlebot I signed it!

davidgrayston:
I've opened #7677 in case you want test coverage for deprecations.

I can close this PR if #7677 is preferable.

haberman:
I think this is a good bugfix. I would like to see if we can make it on the 3.12.x branch instead of master so we can release it in 3.12.3.

The next release of protobuf will be 4.0 which will drop PHP5 support for the C extension (see https://github.com/protocolbuffers/protobuf/issues/7525). Ideally we could get these patch fixes into the 3.12.x branch, in case any PHP5 users of the C extension could still benefit from these changes.

haberman:
I guess another option is to cherry-pick it to 3.12.x.

haberman:
Cherry-picking might be better, the tests seem busted on the release branch.

davidgrayston:
Opened #7690 for `3.12.x`

PR #7631, opened 2020-06-15
gRPC-Web-Ts

behunin:
[gRPC-Web-Ts client](https://github.com/behunin/grpc-web-ts)

Generator Options are as follows:

  Name: this is used with the A~1 ratio option, it is the file name used for the output file.

example - protoc -I=$SRC_DIR –ts_out=name=YourNameSpaceHere:$OUT_DIR test.proto test1.proto test2.proto

  Ratio:  input to output ratio – if not used the default ratio is one output to each input file.
   ```
 S~1 – print one file for each service definition.
 A~1 – print one file from all input files.
```

example - protoc -I=$SRC_DIR --ts_out=ratio=S~1:$OUT_DIR test.proto

  Deps: currently only used with the default one to one output.
    Print – prints each dependency in its own file. Then prints import statements for each dependency.
    Import – just prints the import statements for each dependency.
    Include – prints dependencies in the output file.

example - protoc -I=$SRC_DIR --ts_out=deps=print:$OUT_DIR test.proto

  Services: print the generic service definitions or not – can be used with any output ratio.

example - protoc -I=$SRC_DIR --ts_out=services:$OUT_DIR test.proto

I updated cmake to 3.17.3 and CXX flags to c++2017 so that I could use map::insert_or_assign
I am guessing that is a deal breaker, but hopefully that can be dealt with in a separate PR


behunin:
@googlebot I signed it!

acozzette:
@behunin Thank you for the pull request, but I think this should go into one of the gRPC repos rather than directly into protobuf itself. You can write this code generator as a protoc plugin so that protoc can delegate to it without it being directly linked into protobuf itself. You might also want to start a conversation with the gRPC maintainers, because they might have their own plans for Typescript.

behunin:
Thank you! I will find out what they think and close this after a decision is made.

PR #7644, opened 2020-06-23
Optimize reading strings across segments

JamesNK:
Before
```
|                                    Method | BytesToParse | encodedSize |       Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------------ |------------- |------------ |-----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|              ParseString_CodedInputStream |        10080 |           1 |  64.871 us | 0.2090 us | 0.1745 us |      0.1221 |           - |           - |               176 B |
|                  ParseString_ParseContext |        10080 |           1 |  66.493 us | 0.8300 us | 0.6931 us |           - |           - |           - |                   - |
| ParseString_ParseContext_MultipleSegments |        10080 |           1 |  80.660 us | 1.5452 us | 1.7175 us |           - |           - |           - |                   - |
|              ParseString_CodedInputStream |        10080 |           4 |  80.335 us | 1.2552 us | 1.1127 us |    102.6611 |           - |           - |             80816 B |
|                  ParseString_ParseContext |        10080 |           4 | 106.574 us | 0.9521 us | 0.8906 us |    102.4170 |           - |           - |             80640 B |
| ParseString_ParseContext_MultipleSegments |        10080 |           4 | 110.729 us | 2.6566 us | 2.4850 us |    102.4170 |           - |           - |             80640 B |
|              ParseString_CodedInputStream |        10080 |          10 |  42.858 us | 0.6439 us | 0.6023 us |     61.7065 |           - |           - |             48560 B |
|                  ParseString_ParseContext |        10080 |          10 |  54.975 us | 0.1663 us | 0.1474 us |     61.4624 |           - |           - |             48384 B |
| ParseString_ParseContext_MultipleSegments |        10080 |          10 |  70.892 us | 0.3835 us | 0.3399 us |     76.7822 |           - |           - |             60480 B |
|              ParseString_CodedInputStream |        10080 |         105 |  10.244 us | 0.0523 us | 0.0490 us |     29.5105 |           - |           - |             23216 B |
|                  ParseString_ParseContext |        10080 |         105 |  10.740 us | 0.1339 us | 0.1187 us |     29.2816 |           - |           - |             23040 B |
| ParseString_ParseContext_MultipleSegments |        10080 |         105 |  29.147 us | 0.1317 us | 0.1100 us |     51.7273 |           - |           - |             40704 B |
|              ParseString_CodedInputStream |        10080 |       10080 |   6.136 us | 0.1177 us | 0.1309 us |     25.6348 |           - |           - |             20360 B |
|                  ParseString_ParseContext |        10080 |       10080 |   5.732 us | 0.0272 us | 0.0241 us |     25.6348 |           - |           - |             20184 B |
| ParseString_ParseContext_MultipleSegments |        10080 |       10080 |  15.753 us | 0.1460 us | 0.1295 us |     38.4521 |           - |           - |             30344 B |
```

After
```
|                                    Method | BytesToParse | encodedSize |       Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------------ |------------- |------------ |-----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|              ParseString_CodedInputStream |        10080 |           1 |  65.621 us | 0.3992 us | 0.3539 us |      0.1221 |           - |           - |               184 B |
|                  ParseString_ParseContext |        10080 |           1 |  65.119 us | 1.1548 us | 1.0237 us |           - |           - |           - |                   - |
| ParseString_ParseContext_MultipleSegments |        10080 |           1 |  76.722 us | 0.6850 us | 0.6073 us |           - |           - |           - |                   - |
|              ParseString_CodedInputStream |        10080 |           4 |  84.192 us | 0.7777 us | 0.7274 us |    102.6611 |           - |           - |             80824 B |
|                  ParseString_ParseContext |        10080 |           4 | 106.247 us | 1.1588 us | 1.0840 us |    102.4170 |           - |           - |             80640 B |
| ParseString_ParseContext_MultipleSegments |        10080 |           4 | 109.838 us | 0.4112 us | 0.3645 us |    102.4170 |           - |           - |             80640 B |
|              ParseString_CodedInputStream |        10080 |          10 |  43.997 us | 0.6887 us | 0.6105 us |     61.7065 |           - |           - |             48568 B |
|                  ParseString_ParseContext |        10080 |          10 |  55.625 us | 0.3236 us | 0.2868 us |     61.4624 |           - |           - |             48384 B |
| ParseString_ParseContext_MultipleSegments |        10080 |          10 |  78.260 us | 1.0150 us | 0.8476 us |     61.5234 |           - |           - |             48440 B |
|              ParseString_CodedInputStream |        10080 |         105 |  10.687 us | 0.2087 us | 0.2786 us |     29.5258 |           - |           - |             23224 B |
|                  ParseString_ParseContext |        10080 |         105 |  10.966 us | 0.2136 us | 0.1894 us |     29.2816 |           - |           - |             23040 B |
| ParseString_ParseContext_MultipleSegments |        10080 |         105 |  35.011 us | 0.4347 us | 0.4066 us |     29.3579 |           - |           - |             23096 B |
|              ParseString_CodedInputStream |        10080 |       10080 |   6.284 us | 0.1229 us | 0.1463 us |     25.6348 |           - |           - |             20368 B |
|                  ParseString_ParseContext |        10080 |       10080 |   5.795 us | 0.0716 us | 0.0670 us |     25.6348 |           - |           - |             20184 B |
| ParseString_ParseContext_MultipleSegments |        10080 |       10080 |  19.453 us | 0.2476 us | 0.2316 us |     25.6348 |           - |           - |             20240 B |
```

### DIFF
--- a/csharp/src/Google.Protobuf.Benchmarks/ParseRawPrimitivesBenchmark.cs
+++ b/csharp/src/Google.Protobuf.Benchmarks/ParseRawPrimitivesBenchmark.cs
@@ -54,10 +54,12 @@ namespace Google.Protobuf.Benchmarks
 
         // key is the encodedSize of string values
         Dictionary<int, byte[]> stringInputBuffers;
+        Dictionary<int, ReadOnlySequence<byte>> stringInputBuffersSegmented;
 
         Random random = new Random(417384220);  // random but deterministic seed
 
         public IEnumerable<int> StringEncodedSizes => new[] { 1, 4, 10, 105, 10080 };
+        public IEnumerable<int> StringSegmentedEncodedSizes => new[] { 105, 10080 };
 
         [GlobalSetup]
         public void GlobalSetup()
@@ -78,10 +80,17 @@ namespace Google.Protobuf.Benchmarks
             fixedIntInputBuffer = CreateBufferWithRandomData(random, BytesToParse / sizeof(long), sizeof(long), paddingValueCount);
 
             stringInputBuffers = new Dictionary<int, byte[]>();
-            foreach(var encodedSize in StringEncodedSizes)
+            foreach (var encodedSize in StringEncodedSizes)
             {
                 byte[] buffer = CreateBufferWithStrings(BytesToParse / encodedSize, encodedSize, encodedSize < 10 ? 10 : 1 );
                 stringInputBuffers.Add(encodedSize, buffer);
+            }
+
+            stringInputBuffersSegmented = new Dictionary<int, ReadOnlySequence<byte>>();
+            foreach (var encodedSize in StringSegmentedEncodedSizes)
+            {
+                byte[] buffer = CreateBufferWithStrings(BytesToParse / encodedSize, encodedSize, encodedSize < 10 ? 10 : 1);
+                stringInputBuffersSegmented.Add(encodedSize, ReadOnlySequenceFactory.CreateWithContent(buffer, segmentSize: 128, addEmptySegmentDelimiters: false));
             }
         }
 
@@ -301,6 +310,19 @@ namespace Google.Protobuf.Benchmarks
         }
 
         [Benchmark]
+        [ArgumentsSource(nameof(StringSegmentedEncodedSizes))]
+        public int ParseString_ParseContext_MultipleSegments(int encodedSize)
+        {
+            InitializeParseContext(stringInputBuffersSegmented[encodedSize], out ParseContext ctx);
+            int sum = 0;
+            for (int i = 0; i < BytesToParse / encodedSize; i++)
+            {
+                sum += ctx.ReadString().Length;
+            }
+            return sum;
+        }
+
+        [Benchmark]
         [ArgumentsSource(nameof(StringEncodedSizes))]
         public int ParseBytes_CodedInputStream(int encodedSize)
         {
@@ -326,9 +348,27 @@ namespace Google.Protobuf.Benchmarks
             return sum;
         }
 
+        [Benchmark]
+        [ArgumentsSource(nameof(StringSegmentedEncodedSizes))]
+        public int ParseBytes_ParseContext_MultipleSegments(int encodedSize)
+        {
+            InitializeParseContext(stringInputBuffersSegmented[encodedSize], out ParseContext ctx);
+            int sum = 0;
+            for (int i = 0; i < BytesToParse / encodedSize; i++)
+            {
+                sum += ctx.ReadBytes().Length;
+            }
+            return sum;
+        }
+
         private static void InitializeParseContext(byte[] buffer, out ParseContext ctx)
         {
             ParseContext.Initialize(new ReadOnlySequence<byte>(buffer), out ctx);
+        }
+
+        private static void InitializeParseContext(ReadOnlySequence<byte> buffer, out ParseContext ctx)
+        {
+            ParseContext.Initialize(buffer, out ctx);
         }
 
         private static byte[] CreateBufferWithRandomVarints(Random random, int valueCount, int encodedSize, int paddingValueCount)
--- a/csharp/src/Google.Protobuf/ParsingPrimitives.cs
+++ b/csharp/src/Google.Protobuf/ParsingPrimitives.cs
@@ -34,6 +34,7 @@ using System;
 using System.Buffers;
 using System.Buffers.Binary;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
@@ -49,6 +50,7 @@ namespace Google.Protobuf
     [SecuritySafeCritical]
     internal static class ParsingPrimitives
     {
+        private const int StackallocThreshold = 256;
 
         /// <summary>
         /// Reads a length for length-delimited data.
@@ -58,7 +60,6 @@ namespace Google.Protobuf
         /// to make the calling code clearer.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-
         public static int ParseLength(ref ReadOnlySpan<byte> buffer, ref ParserInternalState state)
         {
             return (int)ParseRawVarint32(ref buffer, ref state);
@@ -437,13 +438,7 @@ namespace Google.Protobuf
                 throw InvalidProtocolBufferException.NegativeSize();
             }
 
-            if (state.totalBytesRetired + state.bufferPos + size > state.currentLimit)
-            {
-                // Read to the end of the stream (up to the current limit) anyway.
-                SkipRawBytes(ref buffer, ref state, state.currentLimit - state.totalBytesRetired - state.bufferPos);
-                // Then fail.
-                throw InvalidProtocolBufferException.TruncatedMessage();
-            }
+            ValidateCurrentLimit(ref buffer, ref state, size);
 
             if (size <= state.bufferSize - state.bufferPos)
             {
@@ -453,36 +448,13 @@ namespace Google.Protobuf
                 state.bufferPos += size;
                 return bytes;
             }
-            else if (size < buffer.Length || size < state.segmentedBufferHelper.TotalLength)
+            else if (IsDataAvailableInSource(ref state, size))
             {
                 // Reading more bytes than are in the buffer, but not an excessive number
                 // of bytes.  We can safely allocate the resulting array ahead of time.
 
-                // First copy what we have.
                 byte[] bytes = new byte[size];
-                var bytesSpan = new Span<byte>(bytes);
-                int pos = state.bufferSize - state.bufferPos;
-                buffer.Slice(state.bufferPos, pos).CopyTo(bytesSpan.Slice(0, pos));
-                state.bufferPos = state.bufferSize;
-
-                // We want to use RefillBuffer() and then copy from the buffer into our
-                // byte array rather than reading directly into our byte array because
-                // the input may be unbuffered.
-                state.segmentedBufferHelper.RefillBuffer(ref buffer, ref state, true);
-
-                while (size - pos > state.bufferSize)
-                {
-                    buffer.Slice(0, state.bufferSize)
-                        .CopyTo(bytesSpan.Slice(pos, state.bufferSize));
-                    pos += state.bufferSize;
-                    state.bufferPos = state.bufferSize;
-                    state.segmentedBufferHelper.RefillBuffer(ref buffer, ref state, true);
-                }
-
-                buffer.Slice(0, size - pos)
-                        .CopyTo(bytesSpan.Slice(pos, size - pos));
-                state.bufferPos = size - pos;
-
+                ReadRawBytesIntoSpan(ref buffer, ref state, size, bytes);
                 return bytes;
             }
             else
@@ -543,13 +515,7 @@ namespace Google.Protobuf
                 throw InvalidProtocolBufferException.NegativeSize();
             }
 
-            if (state.totalBytesRetired + state.bufferPos + size > state.currentLimit)
-            {
-                // Read to the end of the stream anyway.
-                SkipRawBytes(ref buffer, ref state, state.currentLimit - state.totalBytesRetired - state.bufferPos);
-                // Then fail.
-                throw InvalidProtocolBufferException.TruncatedMessage();
-            }
+            ValidateCurrentLimit(ref buffer, ref state, size);
 
             if (size <= state.bufferSize - state.bufferPos)
             {
@@ -619,7 +585,7 @@ namespace Google.Protobuf
             }
 
 #if GOOGLE_PROTOBUF_SUPPORT_FAST_STRING
-            if (length <= state.bufferSize - state.bufferPos && length > 0)
+            if (length <= state.bufferSize - state.bufferPos)
             {
                 // Fast path: all bytes to decode appear in the same span.
                 ReadOnlySpan<byte> data = buffer.Slice(state.bufferPos, length);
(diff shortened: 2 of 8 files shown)


JamesNK:
@jtattermusch

csharp/src/Google.Protobuf/ParsingPrimitives.cs
```diff
@@ -638,17 +636,68 @@ public static string ReadRawString(ref ReadOnlySpan<byte> buffer, ref ParserInte
...
+            if (IsDataAvailable(ref state, length))
+            {
+                if (state.stringDecoder == null)
+                {
+                    state.stringDecoder = CodedOutputStream.Utf8Encoding.GetDecoder();
+                }
+
+                // We need to decode bytes incrementally across multiple spans.
+                int maxCharLength = CodedOutputStream.Utf8Encoding.GetMaxCharCount(length);
+                char[] charArray = ArrayPool<char>.Shared.Rent(maxCharLength);
```
jtattermusch, line 657: I'm not convinced that using the array pool here is always going to what the user wants. Is there a precedent of using array pools for such a basic functionality (like parsing a string in what is a general purpose library).

My worries is that using the array pool here can mess with e.g. user app's concurrency model or their own uses of the arrayPool (we don't know what the user app does and user does not have an option to opt-out). I'm fine with this if there is sufficient precedent that this is ok to do.

csharp/src/Google.Protobuf/ParsingPrimitives.cs
```diff
@@ -638,17 +636,68 @@ public static string ReadRawString(ref ReadOnlySpan<byte> buffer, ref ParserInte
...
-            // TODO: even if GOOGLE_PROTOBUF_SUPPORT_FAST_STRING is not supported,
-            // we could still create a string efficiently by using Utf8Encoding.GetString(byte[] bytes, int index, int count)
-            // whenever the buffer is backed by a byte array (and avoid creating a new byte array), but the problem is
-            // there is no way to get the underlying byte array from a span.
+                        state.bufferPos += bytesRead;
+                    }
 
-            // TODO: in case the string spans multiple buffer segments, creating a char[] and decoding into it and then
-            // creating a string from that array might be more efficient than creating a string from the copied bytes.
+                    string value = new string(charArray, 0, initializedChars);
```
jtattermusch, line 687: Based on the documentation, this constructor still creates a new char array internally (which is also why you're returning the rented char array).
That also means
- you're renting a charArray (which is approx. 2*length bytes long) and then decode into the charArray with not-so-trivial code
- then you copy the decoded chars into a new array anyway.

So compared to the current approach, the only improvement seems to be the use of arrayPool (vs just allocating new byte[] array). (Benchmark numbers seem to confirm that).
What you could do is to just rent a byte[] array, read raw bytes into it (much simpler than using a Decoder) and then create a new string by decoding that byte array (and finally return the rented byte array). You'd even need to rent less memory in most cases (because you only need 1byte vs 2bytes for some common characters).

csharp/src/Google.Protobuf.Benchmarks/ParseRawPrimitivesBenchmark.cs
```diff
@@ -78,10 +79,13 @@ public void GlobalSetup()
...
 
             stringInputBuffers = new Dictionary<int, byte[]>();
-            foreach(var encodedSize in StringEncodedSizes)
+            stringInputBuffersSegmented = new Dictionary<int, ReadOnlySequence<byte>>();
+            foreach (var encodedSize in StringEncodedSizes)
             {
                 byte[] buffer = CreateBufferWithStrings(BytesToParse / encodedSize, encodedSize, encodedSize < 10 ? 10 : 1 );
                 stringInputBuffers.Add(encodedSize, buffer);
+
+                stringInputBuffersSegmented.Add(encodedSize, ReadOnlySequenceFactory.CreateWithContent(buffer, segmentSize: 64));
```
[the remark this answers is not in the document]
jtattermusch: Also, if you choose segmentSize=64, there is no value in testing encodedSizes where (64 % encodedSize == 0) - those cases actually lead to a bit misleading results. 
Testing "multisegment" only encodedSizes 105 and 10080 should be sufficient.

csharp/src/Google.Protobuf/ParsingPrimitives.cs
JamesNK: Yes, this pattern is commonly used in .NET Core runtime when you temporarily need an array.

https://github.com/dotnet/runtime/blob/365c91674a32b6b54181c8248c32cf38a85069b2/src/libraries/System.Text.Json/src/System/Text/Json/Reader/JsonReaderHelper.Unescaping.cs#L14-L36

Using a stackalloc array for small strings like they do here is a good improvement.

csharp/src/Google.Protobuf/ParsingPrimitives.cs
JamesNK: Good point. Changed

csharp/src/Google.Protobuf.Test.TestProtos/UnittestIssues.cs
```diff
@@ -2892,9 +2892,6 @@ public sealed partial class Nested : pb::IMessage<Nested>
   }
 
   public sealed partial class NullValueOutsideStruct : pb::IMessage<NullValueOutsideStruct>
-  #if !GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE
```
JamesNK, line 2895: Fix broken code from merge

JamesNK:
Updated numbers:

```
|                                    Method | BytesToParse | encodedSize |      Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------------ |------------- |------------ |----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|              ParseString_CodedInputStream |        10080 |           1 | 65.246 us | 0.7903 us | 0.6599 us |      0.1221 |           - |           - |               176 B |
|                  ParseString_ParseContext |        10080 |           1 | 65.650 us | 1.4109 us | 1.9313 us |           - |           - |           - |                   - |
|              ParseString_CodedInputStream |        10080 |           4 | 83.324 us | 1.4563 us | 1.2161 us |    102.6611 |           - |           - |             80816 B |
|                  ParseString_ParseContext |        10080 |           4 | 81.609 us | 1.6136 us | 2.9506 us |    102.4170 |           - |           - |             80640 B |
|              ParseString_CodedInputStream |        10080 |          10 | 41.770 us | 0.2447 us | 0.2169 us |     61.7065 |           - |           - |             48560 B |
|                  ParseString_ParseContext |        10080 |          10 | 40.760 us | 0.7884 us | 0.6989 us |     61.4624 |           - |           - |             48384 B |
|              ParseString_CodedInputStream |        10080 |         105 | 10.801 us | 0.2133 us | 0.2773 us |     29.5105 |           - |           - |             23216 B |
|                  ParseString_ParseContext |        10080 |         105 | 10.385 us | 0.1335 us | 0.1249 us |     29.2816 |           - |           - |             23040 B |
| ParseString_ParseContext_MultipleSegments |        10080 |         105 | 19.579 us | 0.0907 us | 0.0849 us |     29.2664 |           - |           - |             23040 B |
|              ParseString_CodedInputStream |        10080 |       10080 |  6.538 us | 0.1290 us | 0.1380 us |     25.6348 |           - |           - |             20360 B |
|                  ParseString_ParseContext |        10080 |       10080 |  6.112 us | 0.0757 us | 0.0671 us |     25.6348 |           - |           - |             20184 B |
| ParseString_ParseContext_MultipleSegments |        10080 |       10080 |  9.524 us | 0.1890 us | 0.2457 us |     25.6348 |           - |           - |             20184 B |
```

csharp/src/Google.Protobuf/ParsingPrimitives.cs
```diff
@@ -638,17 +638,64 @@ public static string ReadRawString(ref ReadOnlySpan<byte> buffer, ref ParserInte
...
+                byte[] byteArray = null;
+                Span<byte> byteSpan = length <= StackallocThreshold ?
+                    stackalloc byte[length] :
+                    (byteArray = ArrayPool<byte>.Shared.Rent(length));
 
-            // TODO: in case the string spans multiple buffer segments, creating a char[] and decoding into it and then
-            // creating a string from that array might be more efficient than creating a string from the copied bytes.
+                try
+                {
+                    int remainingByteLength = length;
```
jtattermusch, line 662: the code that reads possibly multiple segments (refreshing when needed) into a span is doing exactly the same thing as one of the branches of ReadRawBytes (after we've checked that the data is available and that is safe to allocate the span) - I think we should refactor that code out into a private method.
Doing that should also help fix some inconsistencies
- e.g. currently for ReadStringSlow/ReadRawString we don't seem to perform a check for crossing currentLimit (while ReadRawBytes performs such check) - it would be good to unify.
- also the exact "IsDataAvailable" check details are slightly different in ReadRawBytes (it checks `size < buffer.Length || size < state.segmentedBufferHelper.TotalLength`  which works but IsDataAvailable arguably does a better job at that - at least in terms of the `size < state.segmentedBufferHelper.TotalLength` (as opposed to `size < state.segmentedBufferHelper.TotalLength - state.totalBytesRetired - state.bufferPos` from IsDataAvailable) part of the condition.

csharp/src/Google.Protobuf/ParsingPrimitives.cs
```diff
@@ -731,5 +753,56 @@ public static long DecodeZigZag64(ulong n)
...
+        /// Checks whether there is known data available of the specified size remaining to parse
+        /// in the underlying data source.
+        /// When parsing from a Stream this will return false because we have no knowledge of the amount
+        /// of data remaining in the stream until it is read.
+        /// </summary>
+        private static bool IsDataAvailableInDataSource(ref ParserInternalState state, int size)
+        {
+            // Data fits in remaining source data.
+            // Note that this will never be true when reading from a stream as the total length is unknown.
+            return size <= state.segmentedBufferHelper.TotalLength - state.totalBytesRetired - state.bufferPos;
```
JamesNK, line 783: Can you double check for me that this check can be `<=`?

csharp/src/Google.Protobuf/ParsingPrimitives.cs
```diff
@@ -638,20 +604,76 @@ public static string ReadRawString(ref ReadOnlySpan<byte> buffer, ref ParserInte
...
-            var decoder = WritingPrimitives.Utf8Encoding.GetDecoder();
+            return ReadStringSlow(ref buffer, ref state, length);
+        }
+
+        /// <summary>
+        /// Reads a string assuming that it is spread across multiple spans in a <see cref="ReadOnlySequence{T}"/>.
+        /// </summary>
+        private static string ReadStringSlow(ref ReadOnlySpan<byte> buffer, ref ParserInternalState state, int length)
+        {
+            ValidateCurrentLimit(ref buffer, ref state, length);
```
jtattermusch, line 615: I think ValidateCurrentLimit invocation should be in ReadRawString (in the intial checking of preconditions) - at least that would make it consistent with ReadRawBytes and SkipRawBytes - and I also think like this we're still missing one possible scenario (entire string fits into the remaining current buffer but the length is still over the currentLimit) and I think that's a bug (not introduced by this PR, it existed even before this PR it seems, but still).

Let's move `ValidateCurrentLimit(ref buffer, ref state, length);` to line ~586 (just after the negative size check)?

csharp/src/Google.Protobuf/ParsingPrimitives.cs
jtattermusch: Yes, that seems correct.

csharp/src/Google.Protobuf/ParsingPrimitives.cs
```diff
@@ -453,36 +448,13 @@ public static byte[] ReadRawBytes(ref ReadOnlySpan<byte> buffer, ref ParserInter
                 state.bufferPos += size;
                 return bytes;
             }
-            else if (size < buffer.Length || size < state.segmentedBufferHelper.TotalLength)
+            else if (IsDataAvailableInSource(ref state, size))
```
jtattermusch, line 451: It makes sense to change `size < state.segmentedBufferHelper.TotalLength` part of the condition into the more strict `IsDataAvailableInSource(ref state, size)` (which also takes into account the amount of data we've read) because we should never be reading anything over the known TotalLength.

the `size < buffer.Length` part of the condition is missing now  and I think that was there to express that if we have CodedInputStream that reads from stream, the buffer being used is considered small enough so that allocating a new byte array up to its size is still safe.

Because we want to avoid taking the code path of allocating chunk-by-chunk and then concatenating those,
it think we can make this condition:
`(!state.segmentedBufferHelper.TotalLength.HasValue &&  size < buffer.Length) || IsDataAvailableInSource(ref state, size)`
that ways we will still be conservative when totalLength is known but we will allow allocating a reasonable sized array when reading from Stream with an intermediate buffer. (this also preserves the original behavior in that case).

jtattermusch:
Ok, now there's also a conflict after merging another PR.

csharp/src/Google.Protobuf/ParsingPrimitives.cs
JamesNK: > and I also think like this we're still missing one possible scenario (entire string fits into the remaining current buffer but the length is still over the currentLimit) and I think that's a bug

bufferPos is limited by currentLimit. If content fits inside bufferPos then it will also be inside the limit.

PR #7646, opened 2020-06-24
Support JsType in Java JSON serialization

rockwotj:
FieldOptions already has a enum for specifying how numbers are serialized
over the wire - this just hooks up the JSON printer to look at this value.

This is both backwards compatible and parsing already supports reading longs
as strings or numbers.


rockwotj:
@haberman is there any chance of merging this?

rockwotj:
Maybe @acozzette knows what it takes to move forward with this?

acozzette:
@rockwotj Thank you for sending us the pull request, but unfortunately I don't think we want to go in this direction. The `jstype` option is intended to have an effect on the Javascript implementation but not on the JSON format. We are also trying to avoid adding new configuration knobs on the JSON format as much as possible, so that the format stays simple and consistent between different language implementations.

rockwotj:
Thanks for a response!

PR #7652, opened 2020-06-26
[bazel] Use proto_library targets to exclude WKPs from code-gen

Yannic:
This prepares protobuf for a future release of Bazel which will make
`ProtoInfo` mandatory for `proto_lang_toolchain.blacklisted_protos`.


acozzette:
Thanks, @Yannic.

Yannic:
@acozzette Thanks for merging! Do you know when we can expect the next release? There's an associated incompatible change in Bazel and I'd like to see when that flag can be flipped.

Also, could we cherry-pick this into `3.12.x` in case there's a patch release?

acozzette:
There should be another release later this week, so I think it would be best to just wait for that one rather than cherry-picking into the 3.12.x branch.

PR #7656, opened 2020-06-28
[README] change go's runtime

3ks:
I found follow description at repository [golang/protobuf](https://github.com/golang/protobuf):

```bash
It has been superseded by the google.golang.org/protobuf module, 

which contains an updated and simplified API, support for protobuf reflection, and many other improvements. We recommend 

that new code use the google.golang.org/protobuf module.
```

So I think we can replace the recommended runtime with [protocolbuffers/protobuf-go](https://github.com/protocolbuffers/protobuf-go) for go.


acozzette:
Thanks, @3ks.

Issue #2397, opened 2016-11-20
Make Linux downloads of protoc available as tarballs


lorenz:
Currently the Linux downloads for protoc are only avaliable as zip files, which means that to use them in a container it is necessary to install unzip and download the file in the container itself instead of an add instruction which automatically decompresses the downloaded file.
Could you make the Linux releases available as tarballs too?

taylorjacklespriggs:
It is also necessary to apply the required permissions to the files extracted from the zip (`chmod -R a+rx`) so that they can be used by other users and groups. A tarball with correct file permissions would be way easier to use.

deannagarcia:
This was added, but is being removed for 22.x as documented [here](https://developers.google.com/protocol-buffers/docs/news/2022-08-03).

PR #7662, opened 2020-06-29
Create tarballs for protobuf releases

taylorjacklespriggs:
This creates a tar archive for each binary, the same way that a zip file is created for each binary already.

Tar archives preserve file permissions, which simplifies installation on *nix systems.

A couple issues arise here:
 * The script name is now inappropriate. I am refraining from renaming it here because I suspect it is probably used in some script external to this repository.
 * The `*win*.tar.gz` releases may be useless, although `cygwin` users may be able to use them?

Addresses the following issue https://github.com/protocolbuffers/protobuf/issues/2397


taylorjacklespriggs:
@googlebot I signed it!

taylorjacklespriggs:
@googlebot I signed it!

PR #7663, opened 2020-06-30
Optimize writing strings across multiple buffers

JamesNK:
Cherry picked from https://github.com/protocolbuffers/protobuf/pull/7576#issuecomment-647272336


csharp/src/Google.Protobuf.Test.TestProtos/UnittestIssues.cs
```diff
@@ -2893,7 +2893,6 @@ public sealed partial class Nested : pb::IMessage<Nested>
 
   public sealed partial class NullValueOutsideStruct : pb::IMessage<NullValueOutsideStruct>
   #if !GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE
-      , pb::IBufferMessage
```
JamesNK, line 2896: Temporary fix of broken code from merge

JamesNK:
Added to this PR is initializing the initial buffer with IBufferWriter.

Before:
```
|                                               Method | BytesToWrite | encodedSize |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------------------------------------- |------------- |------------ |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           1 | 39.56 ns | 0.6020 ns | 0.5631 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           2 | 43.09 ns | 0.7069 ns | 0.6612 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           3 | 45.48 ns | 0.2256 ns | 0.1884 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           4 | 49.91 ns | 0.6544 ns | 0.5801 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           5 | 53.81 ns | 0.9889 ns | 0.8766 ns |           - |           - |           - |                   - |
```

After:
```
|                                               Method | BytesToWrite | encodedSize |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|----------------------------------------------------- |------------- |------------ |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           1 | 33.21 ns | 0.6996 ns | 0.6544 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           2 | 33.31 ns | 0.2451 ns | 0.2047 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           3 | 35.42 ns | 0.7321 ns | 0.7518 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           4 | 38.52 ns | 0.8109 ns | 0.7585 ns |           - |           - |           - |                   - |
| WriteRawVarint32_WriteContextBufferWriter_FirstWrite |        10080 |           5 | 40.58 ns | 0.6540 ns | 0.5798 ns |           - |           - |           - |                   - |
```

JamesNK:
Rebased to fix merge conflict

@jtattermusch Could you please take a look

csharp/src/Google.Protobuf/WriteBufferHelper.cs
```diff
@@ -118,10 +119,12 @@ public static int GetSpaceLeft(ref WriterInternalState state)
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        public static void RefreshBuffer(ref Span<byte> buffer, ref WriterInternalState state)
+        public static void RefreshBuffer(ref Span<byte> buffer, ref WriterInternalState state, int sizeHint = 0)
         {
             if (state.writeBufferHelper.codedOutputStream?.InternalOutputStream != null)
             {
+                Debug.Assert(sizeHint == 0, "CodedOutputStream does not support sizeHint.");
```
JamesNK, line 126: We (.NET team) use Debug.Assert to check for things that shouldn't happen, but aren't worth checking at runtime. I couldn't find any other uses of Debug.Assert outside of test code.

What do you want to do here:
1. Remove this?
2. Verify are runtime (check and throw an exception)
3. Ok to leave as is

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -163,50 +164,122 @@ public static void WriteBool(ref Span<byte> buffer, ref WriterInternalState stat
...
                 byte[] bytes = Utf8Encoding.GetBytes(value);
                 WriteRawBytes(ref buffer, ref state, bytes);
+                return;
+            }
+
+            // The destination byte array is not be large enough so multiple writes are required.
+            // Encoder will keep state of unwritten data.
+            if (state.stringEncoder == null)
+            {
+                state.stringEncoder = Encoding.UTF8.GetEncoder();
```
jtattermusch, line 255: same problem with caching the stringEncoder as in the other PR. Not seeing stringEncoder.Reset() anywhere seems suspicious.

JamesNK: I've changed it so that the convert loop will stop when `completed` is true. That ensures there is no remaining state.

See remarks - https://docs.microsoft.com/en-us/dotnet/api/system.text.encoder.convert?view=netcore-3.1#System_Text_Encoder_Convert_System_Char__System_Int32_System_Byte__System_Int32_System_Boolean_System_Int32__System_Int32__System_Boolean__

jtattermusch: I think it would be safer to just call Reset() proactively because it would make the reasoning much easier.
Like this if an exception is thrown, the internal state of encoder will be maintained. I know normally throwing would invalidate the entire write context, so this should be fine in theory, but it's hard to predict all the possible patterns of use so I think relying on this is fragile.
How much more expensive would be to call Reset() proactively each time? (sounds like it should be cheap and if done so there would be absolutely no question if the encoder could be polluted or not).

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -163,50 +164,126 @@ public static void WriteBool(ref Span<byte> buffer, ref WriterInternalState stat
...
+            bool completed = false;
+
+            // When completed is true then all chars from the source string have been consumed and
+            // the encoder has no state left to write as bytes.
+            while (!completed)
+            {
+                // Refresh the buffer with a minimum size of the maximum unicode char byte size.
+                // A minimum buffer size is required because at least one unicode character must
+                // be written when Encoder.Convert is called. 
+                WriteBufferHelper.RefreshBuffer(ref buffer, ref state, sizeHint: MaximumBytesPerUtf8Char);
```
jtattermusch, line 268: I'm not a fan of introducing the need for having a sizeHint in refreshbuffer for just this specific corner case.

It sounds like having  less than MaximumBytesPerUtf8Char bytes left would be a pretty rare condition, so perhaps if that happens, we could write those bytes into a stackalloc'd buffer and then write the resulting bytes into the destination using WriteRawBytes(span)?

csharp/src/Google.Protobuf.Benchmarks/WriteRawPrimitivesBenchmark.cs
```diff
@@ -150,6 +153,22 @@ public void WriteRawVarint32_WriteContext(int encodedSize)
             ctx.CheckNoSpaceLeft();
         }
 
+        [Benchmark]
+        [Arguments(1)]
+        [Arguments(2)]
+        [Arguments(3)]
+        [Arguments(4)]
+        [Arguments(5)]
+        public void WriteRawVarint32_WriteContextBufferWriter_FirstWrite(int encodedSize)
```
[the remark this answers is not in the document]
jtattermusch: Found it, looks like this is related to the WriteBufferHelper.Initialize change below. Ideally would be a separate PR, but since it's already here, let's leave it.

csharp/src/Google.Protobuf/WritingPrimitives.cs
JamesNK: > Like this if an exception is thrown, the internal state of encoder will be maintained.

Is CodedOutputStream suppose to be safe to use after it throws an exception? We could put it in a try/catch and reset on an error, but it doesn't seem useful if other state (buffer content, position, written output, etc) is in invalid states.

> How much more expensive would be to call Reset() proactively each time? (sounds like it should be cheap and if done so there would be absolutely no question if the encoder could be polluted or not).

I can test, but the `completed` result says that the internal buffer has been emptied so Reset would never do anything.

https://docs.microsoft.com/en-us/dotnet/api/system.text.encoder.convert?view=netcore-3.1#System_Text_Encoder_Convert_System_Char__System_Int32_System_Byte__System_Int32_System_Boolean_System_Int32__System_Int32__System_Boolean

> The `completed` output parameter indicates whether all the data in the input buffer was converted and stored in the output buffer.

deannagarcia:
I'm going to close this request given that there was no response to the last round of comments, but feel free to open again and respond!

PR #7658, opened 2020-06-29
C# write benchmark improvements

jtattermusch:
- target netcoreapp3.1 (the newest LTS of .NET core)
- add a few more useful write benchmarks
- use `BinaryPrimitives.WriteUInt64LittleEndian` for writing fixed64 values as starting from .NET Core 3.x, this is much faster than the current workaround.

Changes are mostly taken from https://github.com/JamesNK/protobuf/commit/337780c0e098185d4df76d498c8f435e105f8a56 (as recommended here https://github.com/protocolbuffers/protobuf/pull/7576#issuecomment-647069227 for more context). Thanks @JamesNK  for the improvements!

PR #7669, opened 2020-06-30
upgrade dotnet SDK in the CI to 3.1

jtattermusch:
the 3.0 SDK is no longer supported and dotnet SDK 3.1 is the LTS version.

This is a prerequsite for https://github.com/protocolbuffers/protobuf/pull/7658


jtattermusch:
@JamesNK  can you please review

jtattermusch:
> https://github.com/protocolbuffers/protobuf/blob/master/global.json hasn't been updated. You don't _have_ to because it has `"rollForward": "latestMinor"`.

I left global.json as is to avoid disturbing developers that already have SDK 3.0.x installed (they can update their sdk at their own pace, the build and tests still work fine).

PR #7673, opened 2020-07-01
csharp compiler support deprecated enum options

dunston:
The deprecated option was only missing on enums


dunston:
@googlebot I fixed it.

dunston:
@googlebot I fixed it.

deannagarcia:
Thanks for this contribution and sorry for the delay. I think we should use the "AddDeprecatedFlag" method that is defined instead of having the copy. Would you mind making that fix?

deannagarcia:
Closing this due to age/lack of response. If you still want to make this change feel free to reopen or create a new PR!

PR #7674, opened 2020-07-01
Check UTF-8 when setting string fields.

haberman:


php/tests/generated_class_test.php
```diff
@@ -375,8 +375,21 @@ public function testStringField()
         // Set bool.
         $m->setOptionalString(true);
         $this->assertSame('1', $m->getOptionalString());
+
     }
 
+    /* TODO(haberman): find a way to enable this test that works across
+     * PHP/PHPUnit versions. This test as written throws the following error for
```
TeBoring, line 382: Which versions doesn't work? Older version?

acozzette:
Closing this since it looks like it's probably obsolete given that UPB has UTF-8 checking.

PR #7678, opened 2020-07-03
Added additional overload in MapField

Soon5:
Hello,

I created an additional Method overload so other IEnumerable<KeyValuePair> collections can be added. This is especially required as e.g. IReadOnlyDictionary is otherwise not possible to Add.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Collections/MapFieldTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Collections/MapFieldTest.cs
@@ -179,6 +179,35 @@ namespace Google.Protobuf.Collections
             CollectionAssert.AreEqual(new[] { "before", "x", "a", "after" }, map2.Keys);
         }
 
+        [Test]
+        public void Add_IEnumerable()
+        {
+            var map1 = new MapField<string, string>
+            {
+                { "x", "y" },
+                { "a", "b" }
+            };
+
+            IReadOnlyDictionary<string, string> map2 = new Dictionary<string, string>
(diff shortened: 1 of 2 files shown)


Soon5:
@googlebot I signed it!

acozzette:
I don't know any C# but I asked some C# folks for advice and someone told me this: "Would not recommend this implementation exactly. IEnumerables are fairly general and technically have no guarantee that they will ever complete. If more overloads are desired for collection types, would recommend using ICollection<T> and IReadOnlyCollection<T> instead to guarantee fully-materialized collections at a type level."

Soon5:
Well, then two additional methods would be required, one with ICollection<KeyValuePair....> and another one with IReadOnlyCollection<...>. The Problem is, with both Methods, lot's of collections will run into an ambiguity issue as they implement ICollection and IReadOnlyCollection and therefore IEnumerable is the only common Interface they both share, so the ambiguity is circumvented. Otherwise the compiler will request an explicit cast to one or the other Interface to know which method to use.

Also the argument that IEnumerables can be endless is kind of strange to me. Within the codebase of the Add(IDictionary) Method in the foreach loop, the IEnumerable Interface is used. Technically if someone want's to write stupid code, there can be an unlimited Size dictionary. Sure the Interface makes it possible to have unlimited size collections, but that's a rather theoretical issue.

Almost all Microsoft Provided Collections have an "AddRange" Method taking an IEnumerable as argument, so MapField being a collection type should as well.

Issue #7682, opened 2020-07-06
Compilation failure on HPPA


Arfrever:
Protobuf 3.12.2 fails to compile on HPPA.
This error does not occur on x86_32, x86_64.
Original report: https://bugs.gentoo.org/731002

```
/bin/sh ../libtool  --tag=CXX   --mode=compile hppa2.0-unknown-linux-gnu-g++ -DHAVE_CONFIG_H -I. -I/var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src -I..   -DGOOGLE_PROTOBUF_NO_RTTI -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare  -mschedule=8000 -march=2.0 -ggdb -Wall -O2 -pipe -c -o google/protobuf/struct.pb.lo /var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src/google/protobuf/struct.pb.cc
libtool: compile:  hppa2.0-unknown-linux-gnu-g++ -DHAVE_CONFIG_H -I. -I/var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src -I.. -DGOOGLE_PROTOBUF_NO_RTTI -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare -mschedule=8000 -march=2.0 -ggdb -Wall -O2 -pipe -c /var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src/google/protobuf/struct.pb.cc  -fPIC -DPIC -o google/protobuf/.libs/struct.pb.o
In file included from /var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src/google/protobuf/struct.pb.h:24,
                 from /var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src/google/protobuf/struct.pb.cc:4:
/var/tmp/portage/dev-libs/protobuf-3.12.2/work/protobuf-3.12.2/src/google/protobuf/arena.h: In instantiation of ‘void* google::protobuf::Arena::AllocateInternal(bool) [with T = google::protobuf::Struct]’:
[...]

PR #7684, opened 2020-07-06
Remove "T is overaligned" assertion

jer-gentoo:
On some targets, the assertion fails without actually informing why this
is the case:

   .../protobuf-3.12.2/src/google/protobuf/arena.h:531:30: error: static
   assertion failed: T is overaligned, see b/151247138

Rather than finding out what "b/151247138" is or where it fails to guard
or change the assertion test, just remove it to see the full test suite
succeed.

Bug: https://bugs.gentoo.org/731002
Closes: https://github.com/protocolbuffers/protobuf/issues/7682
Signed-off-by: Jeroen Roovers <jer@gentoo.org>

### DIFF
--- a/src/google/protobuf/arena.h
+++ b/src/google/protobuf/arena.h
@@ -528,7 +528,6 @@ class PROTOBUF_EXPORT PROTOBUF_ALIGNAS(8) Arena final {
   // the cookie is not null.
   template <typename T>
   PROTOBUF_ALWAYS_INLINE void* AllocateInternal(bool skip_explicit_ownership) {
-    static_assert(alignof(T) <= 8, "T is overaligned, see b/151247138");
     const size_t n = internal::AlignUpTo8(sizeof(T));
     AllocHook(RTTI_TYPE_ID(T), n);
     // Monitor allocation if needed.


haberman:
This is not a safe change. Can you give more details?

- What platform? (From the linked bug report it looks like hppa?)
- What type `T` was in this context?

The `Arena` class rounds all allocations up to size 8. That means that the memory returned from `Create` is only guaranteed to be aligned to a multiple of 8. If you try to allocate a type that requires greater alignment (16 or more) then it may return a memory address that is not safely aligned. This could lead to a `SIGBUS` runtime error indicating an unaligned load or store.

Arfrever:
Yes, at least HPPA platform.
From his build log:
```
In file included from src/google/protobuf/struct.pb.h:24,
                 from src/google/protobuf/struct.pb.cc:4:
src/google/protobuf/arena.h: In instantiation of ‘void* google::protobuf::Arena::AllocateInternal(bool) [with T = google::protobuf::Struct]’:
src/google/protobuf/arena.h:609:28:   required from ‘T* google::protobuf::Arena::DoCreateMessage(Args&& ...) [with T = google::protobuf::Struct; Args = {}]’
src/google/protobuf/arena.h:503:39:   required from ‘static T* google::protobuf::Arena::CreateMessageInternal(google::protobuf::Arena*) [with T = google::protobuf::Struct]’
src/google/protobuf/struct.pb.cc:1097:77:   required from here
src/google/protobuf/arena.h:531:30: error: static assertion failed: T is overaligned, see b/151247138
  531 |     static_assert(alignof(T) <= 8, "T is overaligned, see b/151247138");
      |                   ~~~~~~~~~~~^~~~
```
So `T = google::protobuf::Struct`.

haberman:
I see. I am curious which types need 16 byte alignment on hppa. But regardless, it sounds like we need 16 byte alignment on some architectures.

I see two main possibilities here:

1. We can attempt to calculate the max alignment as a `constexpr` function, [something like this](https://godbolt.org/z/fZGt5K). We can use that instead of hard-coding `8` anywhere.
2. We can let users override the hard-coded `8` with a macro like `-DPROTOBUF_ARENA_MAXALIGN=16`.

The first option seems better if we can implement it.

Arfrever:
This `static_assert(alignof(T) <= 8, "T is overaligned, see b/151247138")` line in `src/google/protobuf/arena.h` was deleted in commit b70695dd68488e0a9bc847de69dadc6b47e251e9.

jorgbrown:
Closing because this static_assert was deleted in commit b70695d.

PR #7689, opened 2020-07-07
PHP: Removed references from generated code to legacy names.

haberman:


haberman:
Fixes: https://github.com/protocolbuffers/protobuf/issues/7447

TeBoring:
Should this change be in a major release? This probably will break backward compatibility of old generated code.

haberman:
@TeBoring For this to break, you would have to be mixing generated code from 3.12.3 (July 2020) with generated code from before 3.6.0 (June 2018). While this is theoretically possible, it seems unlikely.

That said, I agree there is a risk. @bshaffer asked if we could put this in 3.12.x so that users could benefit from getting rid of the deprecation warnings even if they are still on PHP5, since protobuf 4.0 will drop C extension support for PHP5.

TeBoring:
I see. SGTM

haberman:
This was already done in https://github.com/protocolbuffers/protobuf/pull/7695/commits/77edcb3f3d0a1a7fa57fc293b1b91a2435981b95 as part of #7695.

PR #7690, opened 2020-07-07
[PHP] [3.12.x] Fix GPBUtil::getFullClassName() argument order

davidgrayston:
#7629 for `3.12.x`

### DIFF
--- a/php/src/Google/Protobuf/Internal/Descriptor.php
+++ b/php/src/Google/Protobuf/Internal/Descriptor.php
@@ -185,8 +185,8 @@ class Descriptor
             $containing,
             $file_proto,
             $message_name_without_package,
-            $legacy_classname,
             $classname,
+            $legacy_classname,
             $fullname);
         $desc->setFullName($fullname);
         $desc->setClass($classname);


davidgrayston:
@haberman before porting this to `3.12.x`, is it worth double checking that the arguments weren't intentionally swapped for backwards compatibility? It was added in  #4741

davidgrayston:
I think swapping the arguments to the "correct" order will break older generated code (that uses underscores instead of namespace), as `getClass()` is used without falling back on `getLegacyClass()`

davidgrayston:
You could fallback on the legacy class like this https://github.com/davidgrayston/protobuf/commit/233e5e0c63d817fd882d77fa8acc6ef5c11680c2 - or perhaps it would be safest to leave it as is and close this PR?

haberman:
> I think swapping the arguments to the "correct" order will break older generated code (that uses underscores instead of namespace), as getClass() is used without falling back on getLegacyClass()

Can you be more specific? What is the failure scenario?

It's hard to imagine that the old code would have intentionally called `setLegacyClass()` with a non-legacy class name.

@TeBoring do you have any comment?

davidgrayston:
Old code would have class `Foo\TestMessage_Sub`, but not `Foo\TestMessage\Sub`, so code such as the following would fail:

```
$klass = $field->getMessageType()->getClass();
$value = new $klass;
```
(from https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L418)

haberman:
The new class names have been generated since Protobuf 3.6.0 in June 2018.

But I agree it may make sense to cut a new minor release for this.

davidgrayston:
ok - adding a fallback (something like this https://github.com/davidgrayston/protobuf/commit/233e5e0c63d817fd882d77fa8acc6ef5c11680c2) should make it safe enough as a bug fix, as it would use the new class if available, and fallback on the legacy class (which would trigger the current deprecation notice).

If this change isn't compatible with old generated code that only has `Message_Sub` classes (without newer `Message\Sub` classes), wouldn't this be too risky even as a new minor?

davidgrayston:
Apologies, I've just spotted your comment on https://github.com/protocolbuffers/protobuf/pull/7689#issuecomment-655106081 - I suppose if there's little chance of this breaking existing code, then it should be alright.

If you did want to add a fallback on the legacy class (to be backwards compatible), I'd be happy to include that in this PR.

davidgrayston:
Closing this PR as #7629 ended up in `3.13.0`

PR #7694, opened 2020-07-08
[bazel] Move Objective-C runtime into //objectivec

Yannic:
Removing language-specific targets from the top-level BUILD file
will allow users to keep their workspaces smaller and easier to
maintain by not depending on language rules they don't need.

The top-level targets are kept as an alias, so this is not a breaking
change. There is currently no intention to deprecate and remove
the top-level targets.

Similar work was done for Java in #7190.


Yannic:
@acozzette As usual I forgot to add the new file to the `*_EXTRA_DIST` files. Tests should pass now.

PTAL, thanks!

Yannic:
@acozzette Thanks! The remaining failures look unrelated to the change.

acozzette:
@thomasvl Do you have any thoughts on this change?

thomasvl:
> @thomasvl Do you have any thoughts on this change?

nope.  The bazel support hasn't been something I've really been looped into.

PR #7696, opened 2020-07-08
Create tarballs for protobuf releases

taylorjacklespriggs:
This creates a tar archive for each binary, the same way that a zip file is created for each binary already.

Tar archives preserve file permissions, which simplifies installation on *nix systems.

A couple issues arise here:

 * The script name is now inappropriate. I am refraining from renaming it here because I suspect it is probably used in some script external to this repository.
 * The *win*.tar.gz releases may be useless, although it might be helpful for cygwin installs?

Addresses the following issues:
 * #2397


acozzette:

Issue #7700, opened 2020-07-12
Can't call ByteString.CopyFrom(ReadOnlySpan<byte> bytes) on net45


basvdlinden:
**What version of protobuf and what language are you using?**
Version: v3.12.3 / v4.0.0-rc1
Language: C#

**What operating system (Linux, Windows, ...) and version?**
Windows 10 Pro 2004 (OS build 19041.330)

**What runtime / compiler are you using (e.g., python version or gcc version)**
.NET 4.7.2

**What did you do?**
```csharp
using System;
using Google.Protobuf;

namespace test1
{
    class Program
    {
        static void Main(string[] args)
        {
            Span<byte> a = new byte[] { 1, 2, 3 };
            var b = ByteString.CopyFrom(a);
        }
    }
}
```

**What did you expect to see**
No exception

**What did you see instead?**
```
Unhandled Exception: System.TypeAccessException: Attempt by security transparent method 'Google.Protobuf.ByteString.CopyFrom(System.ReadOnlySpan`1<Byte>)' to access security critical type 'System.ReadOnlySpan`1<System.Byte>' failed.

Assembly 'Google.Protobuf, Version=3.12.3.0, Culture=neutral, PublicKeyToken=a7d26565bac4d604' is marked with the AllowPartiallyTrustedCallersAttribute, and uses the level 2 security transparency model.  Level 2 transparency causes all methods in AllowPartiallyTrustedCallers assemblies to become security transparent by default, which may be the cause of this exception.
   at Google.Protobuf.ByteString.CopyFrom(ReadOnlySpan`1 bytes)
   at test1.Program.NewMethod() in D:\Dev\test1\Program.cs:line 11
```

**Anything else we should know about your project / environment**
Same issue as "Can't call ByteString.Span on net45 #7139".
Tested putting the [SecuritySafeCritical] on CopyFrom. This solved the problem:
```csharp
#if GOOGLE_PROTOBUF_SUPPORT_SYSTEM_MEMORY
        /// <summary>
        /// Constructs a <see cref="ByteString" /> from a read only span. The contents
        /// are copied, so further modifications to the span will not
        /// be reflected in the returned <see cref="ByteString" />.
        /// </summary>
        [SecuritySafeCritical]
        public static ByteString CopyFrom(ReadOnlySpan<byte> bytes)
        {
            return new ByteString(bytes.ToArray());
        }
#endif
```

PR #7701, opened 2020-07-12
annotate ByteString.CopyFrom(ReadOnlySpan<byte>) as SecuritySafeCritical

basvdlinden:
The change to test solution for issue: 
> Can't call ByteString.CopyFrom(ReadOnlySpan<byte> bytes) on net45 #7700

Fixes https://github.com/protocolbuffers/protobuf/issues/7700


basvdlinden:
@googlebot I signed it!

jtattermusch (approved):
LGTM, thanks for the fix.

PR #7705, opened 2020-07-14
Fix C# optional field reflection when there are regular fields too

jskeet:
Previous tests didn't spot this as each message was either "all
optional" or "all non-optional", at which point there isn't a problem.


jskeet:
Note that almost all of this PR is adding the test - the production code change is just in `OneofDescriptor.cs`

PR #7709, opened 2020-07-14
fix: toDict should obey the integer type

melonstupid:
In python/ json_format.py, I'm confused about this MessagToDict method, if this is going to convert message into dict, it should obey the type claimed in Message other than str() that integer.


melonstupid:
Someone can help me with this ？

Issue #6401, opened 2019-07-18
protoc for c# generates slightly inconsistent code: Direct use of "new"  instead of "parser.CreateTemplate()" to create new instances for singular message fields


TheDraguun:
**What version of protobuf and what language are you using?**
Version: master/v3.7.0 but looks like 3.9.0 is the same
Language: C#

**What operating system and version?**
Windows 10

**What runtime / compiler are you using**
VS2017

**What did you do?**
1. use protoc to generate c# code
2. look at code

**What did you expect to see**
use of "parser.CreateTemplate()" to create new instances for singular message fields

**What did you see instead?**
direct use of "new T()"  
- inconsistent with RepeatedFields codecs
- limits customizability through factory method and message parser

**Anything else we should know about your project / environment**
Implemented pooling of message objects by setting a new static MessageParser<T> with a pool aware factory method. Now running mostly garbage collection free. Code could be cleaner if this issue was addressed.

owt5008137:
We meet the same problem. Is there any plan to support custom factory?

ManuelKugelmann:
@owt5008137 Apply the patch from my pull request and then add a partial class for each message that should have a custom factory:
```
   partial class MyMessage
   {
      static MyMessage factory()
      {
         <...>
      }
      static MyMessage()
      {
         _parser = new MessageParser<MyMessage>(factory);
      }
   }
```

jtattermusch:
Does the trick from https://github.com/protocolbuffers/protobuf/issues/6401#issuecomment-659395435 really work?
The `_parser` field is readonly as far as I can tell.  Also it's not immediatelly clear that the `static` intitializer beats the generated `_parser` assignment.
If so, I think the trick works, but feels a bit hacky as you're forcibly overwriting a field that intentionally private.

On the other hand, being able to support pooling for messages sounds interesting, if done right.
The problems I see
- you have no easy way to return rented messages to the pool (e.g. there is not Dispose() or Return() method)
- I think a big portion of allocations actually comes from allocating fields like string, ByteString etc., where you can't simply avoid the allocations (and if the pooling only works in very special cases, it's less useful).

jtattermusch:
CC @JamesNK  as this sounds like an interesting idea. It's a long haul but maybe this could be the start of being able to do high-performance allocation-free gRPC services.

JamesNK:
For pooling to be generally useful there needs to be a way to return a hierarchy of messages, and reset all the message properties while you're at it.

[...]

PR #7722, opened 2020-07-16
fix issue #6401 - replace new by Parser.CreateTemplate

ManuelKugelmann:
Moves all message allocations to CreateTemplate(), making it easier to e.g. use pools to prevent GC load.

### DIFF
--- a/csharp/src/Google.Protobuf/MessageParser.cs
+++ b/csharp/src/Google.Protobuf/MessageParser.cs
@@ -60,7 +60,7 @@ namespace Google.Protobuf
         /// Creates a template instance ready for population.
         /// </summary>
         /// <returns>An empty message.</returns>
-        internal IMessage CreateTemplate()
+        public IMessage CreateTemplate()
         {
             return factory();
         }
@@ -247,7 +247,7 @@ namespace Google.Protobuf
         /// Creates a template instance ready for population.
         /// </summary>
         /// <returns>An empty message.</returns>
-        internal new T CreateTemplate()
+        public new T CreateTemplate()
         {
             return factory();
         }
--- a/src/google/protobuf/compiler/csharp/csharp_message_field.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_message_field.cc
@@ -104,7 +104,7 @@ void MessageFieldGenerator::GenerateMergingCode(io::Printer* printer) {
     variables_,
     "if (other.$has_property_check$) {\n"
     "  if ($has_not_property_check$) {\n"
-    "    $property_name$ = new $type_name$();\n"
+    "    $property_name$ = $type_name$.Parser.CreateTemplate();\n"
     "  }\n"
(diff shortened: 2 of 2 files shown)


ManuelKugelmann:
CLA signed.

src/google/protobuf/compiler/csharp/csharp_message_field.cc
```diff
@@ -104,7 +104,7 @@ void MessageFieldGenerator::GenerateMergingCode(io::Printer* printer) {
     variables_,
     "if (other.$has_property_check$) {\n"
     "  if ($has_not_property_check$) {\n"
-    "    $property_name$ = new $type_name$();\n"
+    "    $property_name$ = $type_name$.Parser.CreateTemplate();\n"
```
jtattermusch, line 107: after changing the codegen, you'd also need to regenerate the code in this repository.

jtattermusch:
CC @JamesNK

csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -60,7 +60,7 @@ internal MessageParser(Func<IMessage> factory, bool discardUnknownFields, Extens
         /// Creates a template instance ready for population.
         /// </summary>
         /// <returns>An empty message.</returns>
-        internal IMessage CreateTemplate()
+        public IMessage CreateTemplate()
```
JamesNK, line 63: 👍

src/google/protobuf/compiler/csharp/csharp_message_field.cc
JamesNK, line 107: What is the overhead of using Parser.CreateTemplate?

ManuelKugelmann: = doing a build ?

jtattermusch:
Is there a good way of checking that we haven't missed any occurrences of `new XYZMessage()` in the code?

src/google/protobuf/compiler/csharp/csharp_message_field.cc
jtattermusch: Build and then run https://github.com/protocolbuffers/protobuf/blob/master/csharp/generate_protos.sh.

The updates to the generated files need to be commited (preferrably as a separate commit)

jtattermusch:
One potential concern is that if this gets merged, the newly generated code won't be compatible with the older runtimes anymore (there is no public CreateTemplate() method). 
While that is fine and understandable limitation, it would still be better if this change is bundled with other changes that also require runtime upgrade once code is regenerated, to amortize the cost of potentially disrupting user's workflow over multiple features. This feature in isolation is quite a niche case so I'm not sure the benefits outweigh the costs.

ahmednfwela:
@jtattermusch I have issued https://github.com/protocolbuffers/protobuf/pull/19776 to only change the visibility of the CreateTemplate() method.

PR #7723, opened 2020-07-16
Temporarily disabled ExtensionRegistryFactoryTest

acozzette:
This test has started failing, and I believe the cause is a change in
class loading behavior in Java 11. As far as I know, that breaks the
test logic but there is nothing wrong with the non-test code. This
change temporarily disables the test so that we can unblock the 4.0
release while I work on a more permanent fix.

### DIFF
--- a/java/core/src/test/java/com/google/protobuf/ExtensionRegistryFactoryTest.java
+++ b/java/core/src/test/java/com/google/protobuf/ExtensionRegistryFactoryTest.java
@@ -30,6 +30,7 @@
 
 package com.google.protobuf;
 
+import org.junit.Ignore;
 import protobuf_unittest.NonNestedExtension;
 import protobuf_unittest.NonNestedExtensionLite;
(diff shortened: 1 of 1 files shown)


java/core/src/test/java/com/google/protobuf/ExtensionRegistryFactoryTest.java
```diff
@@ -52,7 +53,12 @@
  *
  * <p>The test mechanism employed here is based on the pattern in {@code
  * com.google.common.util.concurrent.AbstractFutureFallbackAtomicHelperTest}
+ *
+ * <p> This test is temporarily disabled due to what appears to be a subtle * change to class
```
haberman, line 57: Stray `*`?

acozzette: Good catch, I fixed it.

PR #7724, opened 2020-07-16
Added proto3 presence support for PHP

haberman:
This PR adds support for proto3 optional fields to PHP, per the guidelines at https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md

# New Public APIs:

- Optional fields get `hasFoo()` and `clearFoo()` methods.
- All oneof fields now have `hasFoo()` methods also.

# Primary Implementation Changes

- Optional fields are initialized to `null` instead of their default value.
- `hasFoo()` tests `isset($this->field)`
- `clearFoo()` calls `unset($this->field)`
- C extension implements object handlers to handle `isset()` and `unset()` calls.
- PHP serializer calls `hasFoo()` if present to decide whether to serialize a field.

# Other Implementation Changes

- We need to regenerate some well-known types to give oneof fields proper `hasFoo()` methods.
- (Longer term, it seems better if well-known types are always regenerated instead of being checked in with modifications).
- Added Valgrind suppression file to silence warnings coming from PHP itself.


php/ext/google/protobuf/message.c
```diff
@@ -108,6 +108,76 @@ static const upb_fielddef *get_field(Message *msg, zval *member) {
...
+ */
+static int Message_has_property(zval *obj, zval *member, int has_set_exists,
+                                void **cache_slot) {
+  Message* intern = (Message*)Z_OBJ_P(obj);
+  const upb_fielddef *f = get_field(intern, member);
+
+  if (!f) return 0;
+
+  if (!upb_fielddef_haspresence(f)) {
+    zend_throw_exception_ex(
```
TeBoring, line 137: Should this be an internal error instead of exception?

haberman: I spoke with @bshaffer and it sounds like it's better to use exceptions than errors in PHP. In fact we were talking about replacing all errors with exceptions for 4.0, since the major version will let us break backward compatibility for this.

haberman: I think I might file an issue for this.

TeBoring: Only our generated code assess this method. Seems user should never see this exception.

stof:
@haberman `unset($this->field);` will unset the field itself from the object (so future writes to it will make it create a dynamic property). `$this->field = null` should be used instead.

Issue #7732, opened 2020-07-18
[PHP] Intermittent conformance failure with new C extension


haberman:
The following error has been showing up intermittently on Kokoro. So far it has only appeared on Mac:

```
CONFORMANCE TEST BEGIN ====================================

ERROR, test=Required.Proto3.JsonInput.FloatFieldQuotedValue.JsonOutput: Output was not equivalent to reference message: modified: optional_float: 1 -> 113
. request=json_payload: "{\"optionalFloat\": \"1\"}" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=json_payload: "{\"optionalFloat\":113}"

These tests failed.  If they can't be fixed right now, you can add them to the failure list so the overall suite can succeed.  Add them to the failure list by running:
  ./update_failure_list.py failure_list_php_c.txt --add failing_tests.txt

  Required.Proto3.JsonInput.FloatFieldQuotedValue.JsonOutput
```

Another instance.

```
CONFORMANCE TEST BEGIN ====================================

ERROR, test=Required.Proto3.JsonInput.FloatFieldQuotedValue.ProtobufOutput: Output was not equivalent to reference message: modified: optional_float: 1 -> 11
. request=json_payload: "{\"optionalFloat\": \"1\"}" requested_output_format: PROTOBUF message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=protobuf_payload: "]\000\0000A"
ERROR, test=Required.Proto3.JsonInput.FloatFieldQuotedValue.JsonOutput: Output was not equivalent to reference message: modified: optional_float: 1 -> 11
. request=json_payload: "{\"optionalFloat\": \"1\"}" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=json_payload: "{\"optionalFloat\":11}"

These tests failed.  If they can't be fixed right now, you can add them to the failure list so the overall suite can succeed.  Add them to the failure list by running:
  ./update_failure_list.py failure_list_php_c.txt --add failing_tests.txt

  Required.Proto3.JsonInput.FloatFieldQuotedValue.JsonOutput
  Required.Proto3.JsonInput.FloatFieldQuotedValue.ProtobufOutput
```

haberman:
This appears to be a bug in the upb JSON parser when a float field is encoded inside a string.

[...]

PR #7734, opened 2020-07-19
[PHP] Updated upb to pick up several bugfixes.

haberman:
This fixes most of PHP's remaining conformance failures.


haberman:
Fixes: https://github.com/protocolbuffers/protobuf/issues/7732

haberman:
There is a goldenfile test failure.

I am guessing this is because upb now properly respects the proto3 semantics of packed serialization by default.

TeBoring:
https://github.com/protocolbuffers/protobuf/pull/7588/files
I updated the golden file in this PR. Feel free to copy any change you need.

haberman:
Hmm, I updated the golden file test, but the golden file compatibility test is broken. The old test expects that the field is encoded as not packed.

@TeBoring could you figure out a solution for this? It seems we will need to change or patch the compatibility tests.

TeBoring:
Continued in #7740

PR #7740, opened 2020-07-20
Updated upb to pick up several bugfixes

TeBoring:
Copied from #7734 

Temporarily disable encode_decode_test.php.

In 4.0.0-rc1, repeated primitive field encoding is changed to packed, which is a bug fix. However, this fails the compatibility test which hard coded old encoding. Will reenable the test after making a release. After the version bump, the compatibility test will use the updated test code.


acozzette:
LGTM

PR #7741, opened 2020-07-21
[python] ParseDict merge dictionaries

mory91:
https://github.com/protocolbuffers/protobuf/pull/5211 broke old behavior which was merging the dictionaries given to the ParseDict function as the documentation mentions in [here](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.ParseDict):

> **message** – A protocol buffer message to merge into. 

Here I tried to reach the desired behavior without breaking the old one.

### DIFF
--- a/python/google/protobuf/internal/json_format_test.py
+++ b/python/google/protobuf/internal/json_format_test.py
@@ -1213,6 +1213,15 @@ class JsonFormatTest(JsonFormatBase):
         {'value': UnknownClass()},
         message)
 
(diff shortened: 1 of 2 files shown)


mory91:
Any idea?

anandolee:
The old behavior that merges dict was wrong. ParseDict should only do parse. If your project need to merge the message, you can use two messages to call the json_format.ParseDict() and then call message.MergeFrom() api to do the merge

PR #7744, opened 2020-07-21
[bazel] Add BUILD files for benchmark protos

Yannic:
This change adds BUILD files to //benchmarks that allow
users to consume the .proto files in that directory
to run benchmarks (e.g. for custom generators).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -17,6 +17,15 @@ http_archive(
     ],
 )
 
+http_archive(
+    name = "com_github_google_benchmark",
+    sha256 = "2a778d821997df7d8646c9c59b8edb9a573a6e04c534c01892a40aa524a7b68c",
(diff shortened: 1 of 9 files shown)


Yannic:
Thanks! The failures look unrelated to this change.

acozzette:
@Yannic Thank you for this change, but the one thing I'm nervous about is the idea of making all these protos publicly visible. If we do that then I'm afraid we will be on the hook for keeping these targets around indefinitely and being careful not to break other projects that depend on them. CC @haberman who has thoughts on this as well.

Yannic:
I understand your concerns. What do you think about restricting the visibility of the (currently) public targets to `//benchmarks:__subpackages__`?

acozzette:
@Yannic That sounds good to me.

Yannic:
@acozzette Sorry for the super long delay here, PTAL

PR #7749, opened 2020-07-23
Support codepoints outside the BMP

seishun:


src/google/protobuf/compiler/js/js_generator.cc
```diff
@@ -593,8 +593,8 @@ std::string JSOneofIndex(const OneofDescriptor* oneof) {
   return StrCat(index);
 }
 
-// Decodes a codepoint in \x0000 -- \xFFFF.
-uint16 DecodeUTF8Codepoint(uint8* bytes, size_t* length) {
+// Decodes a codepoint in \x0000 -- \x10FFFF.
+uint32 DecodeUTF8Codepoint(uint8* bytes, size_t* length) {
```
elharo, line 597: I'm not a JavaScript expert but this seems unlikely to be backwards compatible. Can it be made so?

elharo (changes requested):

src/google/protobuf/compiler/js/js_generator.cc
seishun: Backwards compatible in what way specifically?

elharo: This changes the return type. Code that was expecting a uint16 is now getting a uint32 instead. Is that a problem in context? I'm  not sure, but I'm no JavaScript expert.

seishun: This is C++, not JavaScript. This function is called only from `EscapeJSString`, which is updated accordingly.

elharo: should be OK then

elharo (commented):
Change looks good, but it needs some tests

seishun:
> 
> 
> Change looks good, but it needs some tests

As far as I can tell, there are no tests for compiler output.

acozzette:

PR #7761, opened 2020-07-29
Incorrect use of CMake's newer features

afshinpir:
I incorrectly used 2 features of newer versions of CMake:

1. using `NAME_WLE` for `get_filename_component()` which is supported from CMake 3.14.
2. using character escape `\c` is regular expression which is supported from CMake 3.13.

Since these parts of code has created problems for users of CMake functions, I updated this part of code to support previous versions of CMake too.


afshinpir:
Hi,

I cannot add labels. I guess this needs `cmake` label alone,  because I have only updated cmake files.

Croydon:
Protobuf seems to support CMake >= 3.1.3

https://github.com/protocolbuffers/protobuf/blob/92e3d3e02a886839c75aca74dab4d0ea4cef8eb8/cmake/CMakeLists.txt#L2

So it is probably not okay to use features from newer CMake versions?

PR #7763, opened 2020-07-29
Check realoneof in json parsing

TeBoring:


TeBoring:
@haberman needs to copy this change to upb.

PR #7768, opened 2020-07-31
Optionally ignore unknown enum string values in the JSON parser

jskeet:
Note the current (tested) behavior for repeated fields: an unknown string value becomes the unknown enum value, rather than being omitted entirely.
If that's not the behavior we want, it'll be slightly awkward to change, but better to do that now than decide later. (This should be one of the conformance tests, too.)

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
@@ -938,6 +938,38 @@ namespace Google.Protobuf
         }
 
         [Test]
+        public void Enum_InvalidString_IgnoreUnknownFields()
+        {
+            // When ignoring unknown fields, invalid enum value strings are ignored too.
+            var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(true));
+            string json = "{ \"singleForeignEnum\": \"NOT_A_VALID_VALUE\" }";
+            var parsed = parser.Parse<TestAllTypes>(json);
+            Assert.AreEqual(ForeignEnum.ForeignUnspecified, parsed.SingleForeignEnum);
+        }
+
+        [Test]
+        public void RepeatedEnum_InvalidString_IgnoreUnknownFields()
+        {
+            // When ignoring unknown fields, invalid enum value strings are ignored too.
+            // For a repeated field, the presence of the value is preserved using the 0 value.
+            var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(true));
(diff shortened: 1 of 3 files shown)


jtattermusch:
I can review the changes from C# perspective, but would like input on the protobuf part from @haberman or @anandolee first

jskeet:
Ping @haberman and @anandolee on the plan for this.

jskeet:
Ping @haberman and @anandolee again. I'd really like to get this merged.
(I've rebased the branch.)

anandolee:
I think in proto3 json spec, it has mentioned unknown enum string is not accept.

But in Java, UNSPECIFIED is used in proto3 semantics. So this change looks reasonable ( similar with java).

Add David who is working on JSON spec to confirm

dlj-NaN:
I worry a little bit about overloading the concept of unknown fields.

Personally, I think this change strikes a very reasonable tradeoff, but it has some subtlety. I could imagine cases where preserving the repeated field _value_ invariant is important, but failing on unknown _fields_ is still desired.

Also, this would be different for proto2, where unknown enum values are be preserved differently.

On balance, I think this is fine, as long as it is clearly stated in the spec and validated in conformance tests. Enums tend to be a forced special case, though, so adding a separate option might be better.

jskeet:
For the record, I'd be perfectly happy to have a separate option for this. An internal discussion originally advised using the same option.

(As a separate matter, I would like to consider the idea of preserving unknown field values parsed from JSON too, but that's a different issue...)

dlj-NaN:
I think the motivating distinction for me is handling repeated fields. We just don't have anywhere else that matches the semantics of "set the value of certain repeated field element(s) to 0." I don't necessarily think that's the wrong thing to do; it is novel, though, and doesn't quite follow obviously from saying "ignore unknown fields."

A new option is safer, especially for folks depending on unknown enumerators to fail. But this also seems like a hair so fine that it may never actually be split in practice.

> (As a separate matter, I would like to consider the idea of preserving unknown field values parsed from JSON too, but that's a different issue...)

+1 I support this idea.

jskeet:
Okay - I suggest we leave this PR unmerged then, and discuss the new option name internally so we can roll it out consistently across multiple languages. I'd love to see this resolved this quarter though :)

jskeet:
Please don't - there's currently discussion going on. I expect we'll need to change this a little bit, but I'm expecting to use this as the starting point.

muradovak:
Hello everyone. 
In my company we develop backend services using Go, Java and C# with async communication organised via Kafka. Json transcoding is a very handy feature, which allows us to distribute proto files and have strictly typed topics without sacrificing human readability. However for both Go and Java services introducing new enum values seems not to be a breaking change while it's able to completely break C# consumers.
Could you please speed up the discussion and merge it? It's been almost 3 years already

jskeet:
Closing in favor of https://github.com/protocolbuffers/protobuf/pull/15758.

Issue #2335, opened 2016-11-07
Does not work on ruby:2.3-alpine docker image


sullerandras:
It throws `__memcpy_chk: symbol not found` error when trying to `require 'google/protobuf'`.
As i understand the issue is that Apline is using `musl` instead of `glibc`. Would be nice if `google-protobuf` could work on Alpine ruby since other ruby images are significantly bigger.

Environment: ruby:2.3-alpine docker image
To reproduce it:
```
$ docker run -it --rm ruby:2.3-alpine sh
/ # gem install google-protobuf
Fetching: google-protobuf-3.1.0-x86_64-linux.gem (100%)
Successfully installed google-protobuf-3.1.0-x86_64-linux
1 gem installed
/ # irb
irb(main):001:0> require 'google/protobuf'
LoadError: Error relocating /usr/local/bundle/gems/google-protobuf-3.1.0-x86_64-linux/lib/google/protobuf_c.so: __memcpy_chk: symbol not found - /usr/local/bundle/gems/google-protobuf-3.1.0-x86_64-linux/lib/google/protobuf_c.so
	from /usr/local/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/bundle/gems/google-protobuf-3.1.0-x86_64-linux/lib/google/protobuf.rb:50:in `rescue in <top (required)>'
	from /usr/local/bundle/gems/google-protobuf-3.1.0-x86_64-linux/lib/google/protobuf.rb:47:in `<top (required)>'
	from /usr/local/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:133:in `require'
	from /usr/local/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:133:in `rescue in require'
	from /usr/local/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:40:in `require'
	from (irb):1
	from /usr/local/bin/irb:11:in `<main>'
irb(main):002:0>
```

Workaround is to change the docker image to `ruby:2.3-slim`, which is about 150 megabytes bigger.

apstndb:
I am also faced with a problem. I have tried to solve it by some workaround.

[...]

PR #7773, opened 2020-08-02
Fix missing symbol __va_copy on musl-based systems

stanhu:
As described in https://www.gnu.org/software/libc/manual/html_node/Argument-Macros.html
and https://github.com/protocolbuffers/protobuf/issues/2335#issuecomment-579913357:

1. `va_copy` is not available when building to ISO C90.
2. When `RUBY_PLATFORM` is `linux` or `darwin`, ISO C90 mode is used.
3. Prior to GCC 3.0, GCC provided `__va_copy` as an extension in any
standard mode. After GCC 3.0, the compiler builtin is available.

Prior to this change, attempting to load the Ruby google-protobuf gem on
a musl-based Linux system (e.g. Alpine Linux) would result in
`__va_copy: symbol not found`. The only workaround would be to build the
Ruby gem with `--with-cflags=-D__va_copy=va_copy`.

To fix this, we only use the `__va_copy` macro in scenarios where
`__GNUC__ < 3`.

This mirrors the change in musl:
https://git.musl-libc.org/cgit/musl/commit/?id=def0af189871a499efdc9bc37438d8b20eb702ab.


stanhu:
@googlebot I signed it!

stanhu:
It seems https://github.com/protocolbuffers/upb/pull/73 is a related change.

To reproduce the error prior to this change:

```shell
docker run -it alpine
apk add musl ruby-full ruby-dev gcc make musl-dev openssl openssl-dev g++ linux-headers
gem install google-protobuf -v 3.12.4 --platform=ruby
ruby -e "require 'google/protobuf'"
```

php/ext/google/protobuf/php-upb.h
```diff
@@ -115,7 +115,11 @@ int msvc_vsnprintf(char* s, size_t n, const char* format, va_list arg);
  */
 #define _upb_snprintf __builtin_snprintf
 #define _upb_vsnprintf __builtin_vsnprintf
+#if __GNUC__ >= 3
```
stanhu, line 118: I'm not sure if protobuf supports GCC < 3.0, but if it doesn't, we can drop this.

PR #7775, opened 2020-08-03
Allow CodedOutputStream to be Extended

Maithem:
Allow for alternative implementations of CodedOutputStream. One such
implementation reduces garbage by using a pool byte buffer internally.


Maithem:
> @googlebot I signed it!

@googlebot I signed it!

acozzette:
@Maithem Sorry but I think we want to prevent other classes from extending this one so that we can preserve the most flexibility to change it in the future. I have not worked much with Java protobuf but I would guess that you could implement some of these optimizations by leaving CodedOutputStream unchanged and instead controlling the array or object that the stream writes to.

PR #7781, opened 2020-08-05
Rebased 4.0.x onto 3.13.x branch

acozzette:
After doing a couple of 4.0 release candidates, we decided to go ahead with a 3.13 minor release instead. I rebased the 4.0.x branch onto a new 3.13.x branch, starting from the same branch cut. I think this is the best way to transition cleanly to 3.13 while still keeping the history from the 4.0.x branch.


acozzette:
The CLA is fine because this PR just includes rebased commits that already went through the CLA check.

PR #7783, opened 2020-08-07
Add a setuptools extension for generating Python protobufs

dlj-NaN:
This setuptools extension generates Python protobuf code from .proto files, so they
can be created as part of a normal Python build. The extension uses an installed
protoc binary, which can be explicitly specified if needed.


python/protobuf_distutils/setup.py
```diff
@@ -0,0 +1,122 @@
...
+        # These Python versions should match the protobuf package:
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.3",
+        "Programming Language :: Python :: 3.4",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
```
bobhancock, line 112: We should also include 3.8.

dlj-NaN: Added.

python/protobuf_distutils/README.md
```diff
@@ -0,0 +1,99 @@
...
+### Example setup.py configuration
+
+```python
+from setuptools import setup
+setup(
+    # ...
+    name='example_project',
+
+    # Require this package, but only for setup (not installation):
+    setup_requires=['protobuf_distutils'],
```
bobhancock, line 30: The package installs as protobuf-disutils, with a hyphen not an underbar.

python/protobuf_distutils/README.md
```diff
@@ -0,0 +1,99 @@
...
+  will be placed under `output_dir` according to the relative source paths
+  under `source_dir`.
+
+  For example, the source file `${source_dir}/subdir/message.proto`
+  will be generated as the Python file `${output_dir}/subdir/message_pb2.py`.
+
+- `proto_files`:
+
+  Specific .proto files can be specified for generating code, instead of
+  searching for all .proto files under `source_path`.
```
bobhancock, line 90: Must be a list of strings.

dlj-NaN: Done.

PR #7786, opened 2020-08-07
Update third_party.md

seishun:
Add a library that helps develop protoc plugins in Node.js, and an example plugin that can serve as a template for alternative generators.

Let me know if I have to include Google's license file in these projects.


acozzette:
Thanks, @seishun.

PR #7788, opened 2020-08-08
Change some SplitStringUsing calls to Split.

suertreus:
This is done to match an internal Google change necessary to deprecate an old internal API.  Nothing needs to change here, however making the change allows the internal code to continue to match the Github code, which is desirable.


suertreus:
@acozzette 

This corresponds to my recent internal change; PTAL.

hiyadads33-ctrl:
### 

> _****_

Issue #7721, opened 2020-07-16
PHP extension 3.13 - missing ARGINFO


remicollet:
ARGINFO are missing for lot of methods

These are mandatory for 8.x

```
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\RepeatedField::__construct() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\RepeatedField::__construct() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\RepeatedField::append() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\RepeatedField::append() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkInt32() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkInt32() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkUint32() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkUint32() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkInt64() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkInt64() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkUint64() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkUint64() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkEnum() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkEnum() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkFloat() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkFloat() in Unknown on line 0
PHP Warning:  Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkDouble() in Unknown on line 0

Warning: Missing arginfo for Google\Protobuf\Internal\GPBUtil::checkDouble() in Unknown on line 0
[...]

Issue #7720, opened 2020-07-16
PHP extension  4.0 - cast from zend_obj


remicollet:
Casting zend_object to Message * (or any other internal struct) is not the proper way

example:
```
+/* Message from zend_object */
+static inline Message *fetch_message(zend_object *obj) {
+       return (Message*)((char*)(obj) - XtOffsetOf(Message, std));
+}
+
+/* Message from zval */
+#define Z_MESSAGE_P(zv) fetch_message(Z_OBJ_P((zv)))
+

-  Message* intern = (Message*)obj;
+  Message* intern = fetch_message(obj);

-  Message* intern = (Message*)Z_OBJ_P(obj);
+  Message* intern = Z_MESSAGE_P(obj);

```

remicollet:
P.S.  no PR because of damned CLA.

haberman:
Thanks for taking a look at the new extension!

Did this lead to an actual issue for you (eg. a SIGSEGV) or is this something you just saw from inspecting the code?

I know extensions are supposed to use the pattern above. But we force the property count to zero, which prevents PHP from writing properties after the `zend_object` structure. So in practice it seems to work, at least in the cases I have tried.

haberman:
To leave a bit more info: in my testing, adding this line prevented PHP from writing to any memory after the `zend_object` member of `struct Message`: https://github.com/protocolbuffers/protobuf/blob/ffb2b538347d3a1e1f63bf85db8260dc4d514aa9/php/ext/google/protobuf/message.c#L71-L72

This is probably not a totally kosher approach. I could imagine that it might break on a different major version of PHP, like PHP 8. If so, we should implement your fix.

But if not, does it cause any harm to leave in for now?

haberman:
Also a note that if we do this, we also need to move the `zend_object std` member of the `Message` struct to be at the end instead of at the beginning.

remicollet:
> we also need to move the zend_object std member of the Message struct to be at the end instead of at the beginning.

Yes, the zend_object must be first member in 5.x and last in 7.x

haberman:
I would rather avoid `#ifdef`.

I'm reluctant to change the current code if it works in practice. But if we do need to change it, I'd rather change it for all PHP versions, rather than use `#ifdef`.

remicollet:
The #if is mandatory, Message_read_property protocol must be different in php 7 (accepting a zval) and php 8 (accepting a zend_object).

TeBoring:
[...]

PR #7793, opened 2020-08-11
Port php c extension to php8

TeBoring:
* Only ported c extension to php8.
* Didn't fixed the issue of throwing warnings for missing arginfo in bundled files.
* Tests not fixed, because syntax of phpunit (<7 vs >9.3) are not compatible.
* In next release, needs to drop php5 and php7.0 support (in order to use phpunit > 7)


TeBoring:
Fixed #7720, hasn't fully fixed #7721

tests.sh
```diff
@@ -701,6 +701,70 @@ build_php7.4_zts_c() {
...
+  pushd php
+  rm -rf vendor
+  composer update
+  tests/compile_extension.sh
+  tests/generate_protos.sh
+  php -dextension=./ext/google/protobuf/modules/protobuf.so ./vendor/bin/phpunit
+  popd
+}
+
+build_php8.0_zts_c() {
```
haberman, line 743: Do we need to test the ZTS build separately? I thought that the API changes for ZTS had disappeared as of PHP 7 (the TSRM stuff) and a ZTS build won't actually exercise any parallelism I don't think.

php/ext/google/protobuf/message.c
```diff
@@ -240,29 +242,41 @@ static zval *Message_read_property(zval *obj, zval *member, int type,
...
   } else {
     upb_fieldtype_t type = upb_fielddef_type(f);
     const Descriptor *subdesc = Descriptor_GetFromFieldDef(f);
     bool ok = Convert_PhpToUpb(val, &msgval, type, subdesc, arena);
-    if (!ok) return;
+    if (!ok) goto error;
   }
 
   upb_msg_set(intern->msg, f, msgval, arena);
+#if PHP_VERSION_ID < 80000
```
haberman, line 268: I think this might be a PHP 7.4 change, not 8.0:
https://github.com/php/php-src/blob/PHP-7.4.0/UPGRADING.INTERNALS#L171-L173

Do we have any tests for 7.4?

php/ext/google/protobuf/protobuf.h
```diff
@@ -43,6 +43,24 @@ const zval *get_generated_pool();
 #define GC_DELREF(h) --GC_REFCOUNT(h)
 #endif
 
+#if PHP_VERSION_ID < 80000
+#define PROTO_VAL zval 
+#define PROTO_STR zval
+#define PROTO_MSG_P(obj) (Message*)Z_OBJ_P(obj)
+#define PROTO_STRVAL_P(obj) Z_STRVAL_P(obj)
+#define PROTO_STRLEN_P(obj) Z_STRLEN_P(obj)
+#define PROTO_RETURN_VAL void
```
haberman, line 52: I think this is for 7.4, not 8.0:
https://github.com/php/php-src/blob/PHP-7.4.0/UPGRADING.INTERNALS#L171-L173

php/ext/google/protobuf/protobuf.h
```diff
@@ -43,6 +43,24 @@ const zval *get_generated_pool();
 #define GC_DELREF(h) --GC_REFCOUNT(h)
 #endif
 
+#if PHP_VERSION_ID < 80000
```
haberman, line 46: Please add a link to this, to help explain these changes:
https://github.com/php/php-src/blob/898bb97706326311a367aaef35b3f95510100f3c/UPGRADING.INTERNALS#L37-L39

tests.sh
TeBoring: Removed

php/ext/google/protobuf/message.c
TeBoring: We have tests https://github.com/protocolbuffers/protobuf/blob/master/tests.sh#L665
This change just causes warning so far

php/ext/google/protobuf/protobuf.h
TeBoring: done

Issue #1639, opened 2016-06-01
--version should indicate beta version


cwelton:
Currently `protoc --version` produces `libprotoc 3.0.0` for beta-1, beta-2, and beta-3 builds, which makes it very difficult to tell which beta version of protobuf you have installed.

This is a problem when the beta builds are not compatible and issues such as  https://github.com/google/protobuf/issues/1605 make it impractical to use beta-3 with java, while beta-2 is  viable.

mkruskal-google:
@deannagarcia This sounds reasonable to me.  Is there any reason not to include rc numbers in the version?  Feel free to assign back to me if you agree

deannagarcia:
This has already been implemented, but we are now using release candidates.

PR #7799, opened 2020-08-12
Display beta version in "protoc --version"

saurabhs2501:
Display the beta version when "protoc --version" is executed. Use
PROTOBUF_VERSION_SUFFIX for this purpose. #1639

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -1715,7 +1715,7 @@ CommandLineInterface::InterpretArgument(const std::string& name,
       std::cout << version_info_ << std::endl;
     }
     std::cout << "libprotoc " << internal::VersionString(PROTOBUF_VERSION)
-              << std::endl;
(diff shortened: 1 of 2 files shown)


acozzette:
@saurabhs2501 Thank you for the pull request. The one problem I see is that we have not updated `PROTOBUF_VERSION_SUFFIX` in a long time. I think `update_version.py` needs to update it, because we rely on that script for updating the version numbers everywhere they appear. Could you edit that script to make it update `PROTOBUF_VERSION_SUFFIX`?

saurabhs2501:
@acozzette I have updated `update_version.py` to update `PROTOBUF_VERSION_SUFFIX`

saurabhs2501:
@acozzette I am looking at the build failures. One particular failure(MacOS Python C++) fails with this:

```
sudo: unable to execute /Library/Frameworks/Python.framework/Versions/3.7/bin/pip: No such file or directory
```

Not sure what I need to do about this.

acozzette:
@saurabhs2501 Thanks!

PR #7802, opened 2020-08-12
No need to generate addtional has method for proto2

TeBoring:


php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -20,7 +20,7 @@ class DescriptorProto extends \Google\Protobuf\Internal\Message
     /**
      * Generated from protobuf field <code>optional string name = 1;</code>
      */
-    protected $name = '';
+    protected $name = null;
     private $has_name = false;
```
haberman, line 24: Can we remove this also?

php/src/Google/Protobuf/Internal/DescriptorProto.php
```diff
@@ -119,11 +129,6 @@ public function setName($var)
         return $this;
     }
 
-    public function hasName()
```
haberman, line 122: Removing this is technical a backward incompatible change. However since the method appears to be useless (has_name is never true AFAICS?) this will hopefully not be disruptive.

TeBoring: The situation is that has method has already been generated generally.
Generating has methods for descriptors is a duplicate.

php/src/Google/Protobuf/Internal/DescriptorProto.php
TeBoring: done

php/src/Google/Protobuf/Internal/DescriptorProto.php
haberman: I think has methods are not generated for repeated fields, and should not be.

So this is backward incompatible, but hopefully rarely used. Anyone who was using it should check the repeated field size instead.

PR #7808, opened 2020-08-14
Fix the python builds on mac kokoro.

thomasvl:


thomasvl:
@haberman – I think this gets the kokoro side going again correctly, but since there is a conformance tests also failing, I'm not sure.  I'm hoping you (or someone else) can take a look to confirm if the failure is real or if something about the python env still isn't right for kokoro on Macs.

PR #7814, opened 2020-08-15
Make MSVC treats the source files as UTF-8

kumar8600:
You have to specify explicitly to build UTF-8 encoded files in some locales.
If you do not do so, you will get a compile error.


kumar8600:
I'm withdrawing this PR because it's too much trouble to check to see if I can agree to the CLA.

Issue #7827, opened 2020-08-18
Java compile version issue when upgrading to 3.12.4


kolea2:
**What version of protobuf and what language are you using?**
Version: protobuf-java 3.12.4
Language: Java

**What did you do?**

When executing this line of code (https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsAcceptanceTest.java#L184), we are seeing this error message:
```
[ERROR] test[twoRowsEmptyValue](com.google.cloud.bigtable.grpc.scanner.ReadRowsAcceptanceTest)  Time elapsed: 0 s  <<< ERROR!
java.lang.NoSuchMethodError: java.nio.CharBuffer.flip()Ljava/nio/CharBuffer;
	at com.google.protobuf.TextFormat$Parser.toStringBuilder(TextFormat.java:1686)
	at com.google.protobuf.TextFormat$Parser.merge(TextFormat.java:1670)
	at com.google.protobuf.TextFormat$Parser.merge(TextFormat.java:1642)
	at com.google.protobuf.TextFormat.merge(TextFormat.java:1448)
	at com.google.cloud.bigtable.grpc.scanner.ReadRowsAcceptanceTest.addResponses(ReadRowsAcceptanceTest.java:184)
	at com.google.cloud.bigtable.grpc.scanner.ReadRowsAcceptanceTest.test(ReadRowsAcceptanceTest.java:173)
```

This started when upgrading from protobuf-java 3.12.2 to 3.12.4. I can confirm that testing locally, the test fails with a Java 8 build and passes with Java 11 (`CharBuffer.flip()` changed between those two versions of java). Were there any changes to the language compile level in protobuf-java recently that would explain this?

Thanks for your help!

suztomo:
I'm able to reproduce the issue in a simple Maven project https://github.com/suztomo/protobuf-i7827.

cpovirk:
It sounds like the linkage checker may soon help detect this, but FYI, another option is to run Animal Sniffer, as [we do in Guava](https://github.com/google/guava/blob/d8e921fd2df52184708d6eb9516e4869f548d841/pom.xml#L139-L160).

I've heard mixed things about the `--release` flag. It may be simplest to just build with JDK8. But as long as Animal Sniffer or an updated linkage checker is in place, I would assume that anything that works is fine.

[...]

PR #7830, opened 2020-08-19
Update maven-surefire-plugin to 3.0.0-M5.

cpovirk:
This avoids https://issues.apache.org/jira/browse/SUREFIRE-1439,
which I hit while investigating
https://github.com/protocolbuffers/protobuf/issues/7827

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -166,6 +166,10 @@
           <artifactId>maven-antrun-plugin</artifactId>
           <version>1.8</version>
         </plugin>
+        <plugin>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <version>3.0.0-M5</version>
+        </plugin>
       </plugins>
     </pluginManagement>
   </build>


haberman:
This appears to make the Java Kokoro tests fail.

cpovirk:
Sounds like it may be from using an older version of Maven (3.0.5, which looks to be [a couple years old](https://apache.osuosl.org/maven/maven-3/)). The current version is 3.6.3. Any idea whether it makes sense to update it and, if so, where that's configured?

cpovirk:
Hmm, looks like maybe you just get whatever is in the Debian VM you launch?

https://github.com/protocolbuffers/protobuf/blob/2558c5deefffa3f01c206d5b132b999a18da7573/kokoro/linux/dockerfile/test/java_stretch/Dockerfile#L28
https://github.com/protocolbuffers/protobuf/blob/2558c5deefffa3f01c206d5b132b999a18da7573/kokoro/linux/64-bit/Dockerfile#L62

One is on Debian "stretch," the other on "latest":

https://github.com/protocolbuffers/protobuf/blob/2558c5deefffa3f01c206d5b132b999a18da7573/kokoro/linux/dockerfile/test/java_stretch/Dockerfile#L1
https://github.com/protocolbuffers/protobuf/blob/2558c5deefffa3f01c206d5b132b999a18da7573/kokoro/linux/64-bit/Dockerfile#L10

Whatever Debian I have on my laptop has Maven 3.6.3, but maybe that's newer than those two?

From the failure output, I get the impression that the relevant Dockerfile is kokoro/linux/64-bit/Dockerfile. I'm not sure what "latest" means, though.

elharo:
still seem to be using Maven 3.0.5, need to fix that somehow

elharo:
The problem is we use apt-get to install maven. Debian is way, way out of date. We need to figure out an alternate means of installing a more current Maven, either a different image, a different repo, or a manual install of some kind.

acozzette:
We ended up dropping our Java 7 tests, which were the ones using the old Debian version, so I believe this is now OK to merge.

Issue #7813, opened 2020-08-15
PHP ext - please fix package.xml with real changes


remicollet:
See: 
https://pecl.php.net/package-changelog.php?package=protobuf

All version (including RC) have "GA release" which have no value for users.

Please fix your release process to provide something useful.

TeBoring:
If we have a new release to correct those release log, does that work?

haberman:
Yes I would echo Bo's question, is it too late to fix all of the existing releases?

If we fix it now, will that fix past releases or only future releases?

remicollet:
> If we fix it now, will that fix past releases or only future releases?

On pecl, only fututre release will have new entry
Old changelog cannot be changed (excepted deleting all tarball and uploading thme again, but loosing stat)

BTW, this resquest is mostly for the future.

remicollet:
Please reopen, this is obviously NOT fixed

Last notification received

```
The new PECL package protobuf-3.14.0RC1 (beta) has been released at https://pecl.php.net/.

Release notes
-------------
GA release.
```

So wrong...

remicollet:
In **0.14.0RC1**
```

 <notes>
GA release.
 </notes>

```
So wrong notes...

And in **changelog**
```
  <release>
   <version>
    <release>3.14.0RC1</release>
    <api>3.14.0</api>
   </version>
   <stability>
    <release>beta</release>
    <api>beta</api>
   </stability>
   <date>2020-11-05</date>
   <time>13:39:47</time>
   <license uri="https://opensource.org/licenses/BSD-3-Clause">3-Clause BSD License</license>
   <notes>

   </notes>
  </release>
```

Empty notes...

acozzette:
#8025 should hopefully fix the incorrect "GA release" notes line, but our release process is not yet set up to populate the release notes with actual updates.

remicollet:
For RC2

```
The new PECL package protobuf-3.14.0RC2 (beta) has been released at https://pecl.php.net/.

Release notes
-------------
PHP protobuf

```

Which still have no value ;)

haberman:
Looks like the code we want to fix is here: https://github.com/protocolbuffers/protobuf/blob/1b18833f4f2a2f681f4e4a25cdf3b0a43115ec26/update_version.py#L278-L323

remicollet:
For 3.15.0RC1, see https://pecl.php.net/package-changelog.php?package=protobuf

So we have only " PHP protobuf "

[...]

PR #7832, opened 2020-08-19
Fixed VERSION_SUFFIX logic in update_version.py

acozzette:
This commit makes a couple of fixes:
- Make sure we always update the suffix, since even for a non-RC release
  we need to clear the suffix.
- Make sure the suffix is properly quoted and begins with -rc


TeBoring:
Would this also fix https://github.com/protocolbuffers/protobuf/issues/7813?
The release logs for rc releases should not be GA.

PR #7834, opened 2020-08-20
Update Gradle dependency reference format

paplorinc:
See https://docs.gradle.org/current/userguide/java_library_plugin.html
> The compile, testCompile, runtime and testRuntime configurations inherited from the Java plugin are still available but are deprecated. You should avoid using them, as they are only kept for backwards compatibility.


acozzette:
Thanks, @paplorinc.

Issue #4616, opened 2018-05-11
Null pointer dereference in Arena.


FSMaxB:
If the allocator from the `ArenaOptions` returns `nullptr`, the placement new after that will dereference a null pointer.

https://github.com/google/protobuf/blob/b61dd9d9a229c5a82788fdeefa344aa65affae09/src/google/protobuf/arena.cc#L119-L120

I'm currently dealing with this by throwing an exception from my allocator instead of returning `nullptr`, but I'm not really sure if your code is exception safe.

acozzette:
Thanks for reporting this, but I believe it is working as intended because we do not use exceptions in C++, as per the Google C++ style guide (see [here](https://google.github.io/styleguide/cppguide.html#Exceptions)).

FSMaxB:
I am aware that you aren't using any exceptions at Google, but that doesn't have anything to do with this null pointer dereference.

In `google/protobuf/arena.h` it says:
https://github.com/google/protobuf/blob/b61dd9d9a229c5a82788fdeefa344aa65affae09/src/google/protobuf/arena.h#L130-L135

Given that `malloc` may return `nullptr` if allocation fails, I would expect from that statement, that my custom allocator should do the same.

In any case if a `nullptr` is returned, it is dereferenced when doing the placement new. This has nothing to do with exceptions being enabled/disabled.

You might argue that `malloc` never returns `nullptr` because of overcommit, but:
1. This isn't always the case e.g.:
    * Overcommit could be disabled
    * Some platforms don't do overcommit in the first place.
    * Even with overcommit, `malloc` might still return `nullptr` if the requested size was too large.
2. Given that this pointer can be supplied by any user of the library using `ArenaOptions`, there can be many reasons why a `nullptr` might be returned, that don't relate to how standard `malloc` works.

> I'm currently dealing with this by throwing an exception from my allocator instead of returning nullptr, but I'm not really sure if your code is exception safe.

The above statement was just a way of telling you what my workaround is. And I'm concerned about this workaround being bad, exactly **because** Google doesn't use exceptions. Because this means that Protobuf is written completely without exception safety in mind, making me question if my workaround is safe.

I'm not sure what code clang++/g++ emit for `new` if the allocation fails. I would guess that they call `std::terminate`, because breaking the guarantee that `new` never returns `nullptr` would be really bad!

[...]

PR #7837, opened 2020-08-21
arena: Terminate the program if block_alloc() returns nullptr

saurabhs2501:
Fire an assert if `block_alloc()` fails to allocate memory(returns
`nullptr`). This makes sure the program terminates right there instead of
continuing to use the memory on page 0(on systems with no MMU) which could be dangerous! #4616

### DIFF
--- a/src/google/protobuf/arena.cc
+++ b/src/google/protobuf/arena.cc
@@ -32,6 +32,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cassert>
 #include <limits>
 
 #include <google/protobuf/stubs/mutex.h>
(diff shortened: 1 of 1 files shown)


acozzette:
I don't think this is the right direction to go in. If the memory allocation fails, there is nothing useful we can do except crash. But if that happens, `block_alloc` should handle that itself instead of returning `nullptr`. If we use an `assert`, that doesn't really solve anything since asserts are disabled in optimized builds. If we use a `CHECK`, that is enabled for optimized builds but it seems like a small but unnecessary runtime cost. I suspect this problem is mostly theoretical anyway, since as I understand, modern OSes usually overcommit memory--that means that allocations never fail, and you only get an out-of-memory crash when you actually try to access the memory.

saurabhs2501:
> I don't think this is the right direction to go in. If the memory allocation fails, there is nothing useful we can do except crash. But if that happens, `block_alloc` should handle that itself instead of returning `nullptr`. If we use an `assert`, that doesn't really solve anything since asserts are disabled in optimized builds. If we use a `CHECK`, that is enabled for optimized builds but it seems like a small but unnecessary runtime cost. I suspect this problem is mostly theoretical anyway, since as I understand, modern OSes usually overcommit memory--that means that allocations never fail, and you only get an out-of-memory crash when you actually try to access the memory.

Ok. What do you suggest I should do?

PR #7845, opened 2020-08-25

haberman:


python/setup.py
```diff
@@ -292,7 +292,6 @@ def get_option_from_sys_argv(option_str):
           'build_py': build_py,
           'test_conformance': test_conformance,
       },
-      setup_requires = ['wheel'],
```
acozzette, line 295: I just added this line recently, so I think something must be wrong if it's getting removed again.

haberman: Thanks for catching that, fixed.

There was one other small change that was somehow lost in the last integrate back to Piper. I'll keep a closer eye on the next integrate to Piper.

PR #7847, opened 2020-08-26
Prototype decimal message types

JamesNK:
Prototype, tests, benchmarks for https://github.com/protocolbuffers/protobuf/pull/7039

@jtattermusch

```
|                 Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|     DecimalString_Zero | 536.4 ns |  3.70 ns |  3.28 ns | 0.0267 |     - |     - |     112 B |
| DecimalString_MaxValue | 911.4 ns | 14.96 ns | 13.26 ns | 0.0534 |     - |     - |     224 B |
|         Decimal64_Zero | 166.3 ns |  1.99 ns |  1.77 ns | 0.0191 |     - |     - |      80 B |
|     Decimal64_MaxValue | 265.9 ns |  2.65 ns |  2.48 ns | 0.0286 |     - |     - |     120 B |
|        Decimal128_Zero | 171.7 ns |  1.58 ns |  1.48 ns | 0.0229 |     - |     - |      96 B |
|    Decimal128_MaxValue | 327.2 ns |  3.92 ns |  3.27 ns | 0.0324 |     - |     - |     136 B |
```

Decimal64Version2 and Decimal128Version2 are pretty similar to the other versions. I didn't do anything with them.


jtattermusch:
Thanks for the benchmarks and the measurements!

billpoole-mi:
@JamesNK, I think the logic implemented in the conversions between `System.Decimal` and `Decimal64` has some issues (which I believe are also present in the conversions to/from `Decimal128`):
- The "scale" component of the `System.Decimal` is being written to the `Decimal64`'s Exponent as +`scale` if the decimal value is +ve and -`scale` if the decimal value is -ve. However, I believe the exponent as defined by IEEE 754-2008 does not depend on the sign of the decimal. I believe the `exponent` of a Decimal64 should always be -`scale` if the `significand` exactly equals the integer component of the `System.Decimal`.
- There are circumstances where the conversion from `System.Decimal` to `Decimal64` fails due to overflow on the significand, but _is_ strictly speaking possible. For example, 1.00000000000000000000m fails to convert, but is easily representable with `Decimal64`.
- I don't believe the conversion from `System.Decimal` to `Decimal64` should fail where `System.Decimal` has a value with a higher precision than that supported by `Decimal64`. Rather, the value should just be rounded - in the same way that converting a `float64` to a `float32` doesn't cause an exception if the `float64`'s value cannot be exactly represented in the `float32`.
- The conversion from `Decimal64` to `System.Decimal` fails if the `exponent` is outside the 0->28 range of the scale parameter of the `decimal(int, int, int, bool, byte)` constructor. However, there are ranges of `significand`/`exponent` values that have out-of-range corresponding `scale` values, but are within the range able to be represented by `System.Decimal`. For example, `significand`=1 and `exponent`=10 would directly translate to `new decimal(1, 0, 0, false, -10)`, which would fail because the `scale` parameter cannot be negative with this constructor. However, this `significand`/`exponent` combination equates to 10,000,000,000, which is within the range supported by `System.Decimal`.

If these issues are legitimate, then I suspect the fixes will impact the performance benchmarks of the conversions.

JamesNK:
Thanks for investigating. A format for protobuf decimals hasn't been decided yet. If this one is chosen then I'll look closer at the issues you raised.

billpoole-mi:
No worries. I thought perhaps these issues might weigh into the decision as to what format to use.

billpoole-mi:
By the way @JamesNK , if the solution ends up needing to access a `System.Decimal` as a 4-element array of 32-bit integers, it is possible to do so in .NET Core 3.1 without the heap allocation of the `Decimal.GetBits` method. That is, although we have the `Decimal.TryGetBits` method in .NET 5 (allowing us write into a `Span<byte>` and thus do so with a stack allocation), we can achieve the same in .NET Core 3.1 using a `DecodedDecimal` struct as follows.

```csharp
[StructLayout(LayoutKind.Explicit)]
public struct DecodedDecimal
{
	[FieldOffset(0)]
	private readonly decimal _value;

	public int Flags => _flags;

	[FieldOffset(0)]
	private readonly int _flags;

	public int High32 => _high32;

	[FieldOffset(4)]
	private readonly int _high32;

	public int Mid32 => _mid32;

	[FieldOffset(12)]
	private readonly int _mid32;

	public int Low32 => _low32;

	[FieldOffset(8)]
	private readonly int _low32;

	public DecodedDecimal(decimal value)
		: this()
	{
		_value = value;
	}
}
```

This is then used as follows.

```csharp
var d = -1_000_000_000_000_000_000_000_000_000_0m;
var bits = decimal.GetBits(d);
var decoded = new DecodedDecimal(d);
Console.WriteLine($"{bits[0]} {decoded.Low32}");
Console.WriteLine($"{bits[1]} {decoded.Mid32}");
Console.WriteLine($"{bits[2]} {decoded.High32}");
Console.WriteLine($"{bits[3]} {decoded.Flags}");
```

Which produces the following output.

Low32: 268435456 268435456
Mid32: 1042612833 1042612833
High32: 542101086 542101086
Flags: -2147483648 -2147483648

alrz:
@JamesNK @acozzette 

There is already `google/type/decimal.proto` and `google/type/money.proto` both of which has helpers in [https://github.com/googleapis/gax-dotnet](https://github.com/googleapis/gax-dotnet/blob/3a37991c73b6ca1e0cfa8e093834a26cd0a849ba/Google.Api.CommonProtos/Type), I think this just need to generate `decimal` props and handle conversion. As a bonous swagger/transcoding would also work as expected.


What is blocking this?

acozzette:
Our team (the protobuf team) doesn't own the types in `google/type/`, but you could try asking the owners of those types if they would be interested in a decimal type. I don't think we are prepared to prioritize a decimal type as part of the core protobuf project, though.

PR #7850, opened 2020-08-26
Fix default value of enum

springhack:
while using proto2 with optional and default keyword, it always return string, so fix it


springhack:
@googlebot I signed it!

acozzette:
@springhack Thank you for the pull request and sorry for taking so long to get back to you. Would you mind adding a test case for this?

PR #7869, opened 2020-09-02
Fix -Wmissing-field-initializers warning

keith:


keith:
looks like the spelling error is from different code than what I changed here, I can fix it if desired

src/google/protobuf/descriptor_database.cc
```diff
@@ -628,7 +628,7 @@ bool EncodedDescriptorDatabase::DescriptorIndex::AddFile(const FileProto& file,
                                                          Value value) {
   // We push `value` into the array first. This is important because the AddXXX
   // functions below will expect it to be there.
-  all_values_.push_back({value.first, value.second});
+  all_values_.push_back({value.first, value.second, nullptr});
```
acozzette, line 631: I haven't looked very closely but I don't think it is valid to pass nullptr here, since that will result in passing nullptr to the std::string constructor. Replacing nullptr with `{}` might work.

keith: updated

src/google/protobuf/descriptor_database.cc
```diff
@@ -628,7 +628,7 @@ bool EncodedDescriptorDatabase::DescriptorIndex::AddFile(const FileProto& file,
                                                          Value value) {
   // We push `value` into the array first. This is important because the AddXXX
   // functions below will expect it to be there.
-  all_values_.push_back({value.first, value.second});
+  all_values_.push_back({value.first, value.second, {}});
```
thomasvl, line 631: @acozzette - should this be moved after the `ValidateSymbolName(file.package())` check and simply pass `file.package()`?  i.e. - why do we add the entry, then potentially return false and then updated what was added?

keith: I thought about that too but didn't want to also move it given that the mutation wouldn't happen in that case, but happy to if that's safe!

acozzette: Good question, I can't tell why it is done this way, but I hesitate to move the logic around given that the ordering does seem to be deliberate based on the comment.

keith:
thanks!

PR #7874, opened 2020-09-05
remove GetArena in _InternalParse

czlhs:
arena is not used in _InternalParse, So remove it for better performance


acozzette:
Thanks, @czlhs.

PR #7878, opened 2020-09-08
Fix more -Wunused-parameter warnings

georgthegreat:
While some of the warnings have been fixed in 76f9074c1c947c6f6773c5095e977572de514821, more remains.


acozzette:
Thanks, @georgthegreat.

PR #7879, opened 2020-09-08
Expose python Encode/Decode APIs

### LINKED ISSUE #54: Protobuf lacks framed stream of messages
Lots of applications want a stream of protobuf messages in a file or a network stream.

It could be as simple as exposing the internal utility functions to write a varint to a stream. An application could then write a varint length prefix and then the blob of serialized protobuf.

saurabhs2501:
Fixes #54

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -964,9 +964,13 @@ python_EXTRA_DIST=                                                           \
   python/google/__init__.py                                                  \
   python/google/protobuf/__init__.py                                         \
   python/google/protobuf/compiler/__init__.py                                \
+  python/google/protobuf/decoder.py
+    \
   python/google/protobuf/descriptor.py                                       \
   python/google/protobuf/descriptor_database.py                              \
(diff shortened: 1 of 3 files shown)


saurabhs2501:
This needs somebody's attention. Please review.

saurabhs2501:
Is this going to be reviewed or should I just discard it?

acozzette:
@saurabhs2501 Sorry for the delay; our usual Python reviewer is on vacation at the moment. @dlj-NaN Do you want to review this one?

saurabhs2501:
No problem @acozzette . I didn't know about the reviewer being on vacation. Would be good to get it reviewed by someone else. Thanks!

acozzette:
Thinking about this some more, I'm afraid we don't want to accept this change. The varint encoding and decoding is an internal implementation detail. You can always fork that code if you need it, but it's not something we want to maintain as a public API.

PR #7882, opened 2020-09-10
Fix double loading same message definitions in python, when custom FileOptions are used.

vchigrin:
Before fix it might fail if between load attempts native library, that use same FileOptions
definitions, is loaded into process address space.

Steps to reproduce: On Linux machine run
```
git clone https://github.com/vchigrin/protobuf_bug_proof.git
cd protobuf_bug_proof/bug_3.12.3.2/
./build.sh
./test.py
```

Expected output: 
```
Success.
```

Actual output:
```
  File "./test.py", line 6, in <module>
    import test_pb2
  File "sub/test_pb2.py", line 25, in <module>
    dependencies=[test__options__pb2.DESCRIPTOR,])
  File "/usr/local/lib/python2.7/dist-packages/protobuf-3.13.0-py2.7-linux-x86_64.egg/google/protobuf/descriptor.py", line 942, in __new__
    return _message.default_pool.AddSerializedFile(serialized_pb)
TypeError: Couldn't build proto file into descriptor pool!
Invalid proto descriptor for file "test.proto":
  test.proto: A file with this name is already in the pool.
```

### DIFF
--- a/src/google/protobuf/descriptor.cc
+++ b/src/google/protobuf/descriptor.cc
@@ -4252,7 +4252,20 @@ static bool ExistingFileMatchesProto(const FileDescriptor* existing_file,
         existing_file->SyntaxName(existing_file->syntax()));
   }
 
-  return existing_proto.SerializeAsString() == proto.SerializeAsString();
+  const std::string existing_proto_str = existing_proto.SerializeAsString();
+  const std::string proto_str = proto.SerializeAsString();
+  if (existing_proto_str == proto_str) {
+    return true;
+  }
+  // Second attempt. It is possible that after loading initial descriptor we've registered some
+  // extensions, and result of parsing the same .proto file will be different (initially unknown
+  // extension options was saved as unknown fields).
+  if (existing_proto.ParseFromArray(existing_proto_str.data(), existing_proto_str.length())) {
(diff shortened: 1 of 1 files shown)


vchigrin:
Small gentle ping. 
If I can help you reviewing this patch in any way I will be glad to do anything.

acozzette:
@anandolee Maybe you know the most about this code since it has to do with Python-C++; do you want to review this PR?

anandolee:
The example reproduce code has:

import sub.test_pb2
import test_pb2

They are indeed from same .proto file. I do not think this is a use case we want to support

vchigrin:
OK, just in case I should remind that code above works in almost all scenarios, and fails only very rare, when custom file options are used, and placed in some specific order in .proto file. 
That seems weird since users may build their infrastructure in assumption that this approach of importing python files is fully supported, and will be surprised when it will fail after adding some FileOptions.
May be it is worth to add more checks, so this import approach will fail always, so people will not rely on it.

acozzette:
I'm going to close this since as @anandolee said, we don't support the use case of having the same .proto file appear in multiple places. I agree that ideally we should have better checks to catch this problem, though.

Issue #7650, opened 2020-06-25
PHP equality comparison between messages always returns true


haberman:
Comparing messages (or repeated fields, etc) in PHP always returns true, regardless of the contents:

```php
        $m1 = new TestMessage();
        $m2 = new TestMessage();
        $m3 = new TestMessage();
        $rf1 = new RepeatedField(GPBType::INT32);
        $rf2 = new RepeatedField(GPBType::INT32);

        $rf1[] = 5;

        $m1->setOptionalInt32(1);
        $m2->setOptionalInt32(1);
        $m3->setOptionalInt32(3);

        // These succeed.
        $this->assertTrue($m1 == $m2);
        $this->assertFalse($m1 == $rf1);  // Mixing types

        // These fail.
        $this->assertFalse($m1 == $m3);
        $this->assertFalse($rf1 == $rf2);
```

This is true for the C extension (haven't tested pure-PHP).

Should we change this? Should equality compare contents? Should it be a deep comparison? We should provide answers to these questions and possibly change/fix the existing behavior.

jtattermusch:
AFAIK, using `==` is an antipattern in PHP? ("Always use ==="?)

haberman:
@jtattermusch I think that `==` (comparison) and `===` (object identity) implement different operations? Why would object comparison be an anti-pattern?

We support deep comparison in other languages, [for example in Ruby](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#message):

* `Message#==`: Performs a deep equality comparison between two messages.

haberman:
I'm leaning towards having `==` in PHP implement a deep comparison between two objects of the same type. For objects of different types it could either return false or signal an error.

bshaffer:
In addition to it being intuitive, the PHP `stdClass` objects behave in the way you suggest the objects behaving:

```php
$s1 = new stdClass();
$s2 = new stdClass();
var_dump($s1 == $s2);
// bool(true)
var_dump($s1 === $s2);
// bool(false)

$s1->foo = 'bar';
var_dump($s1 == $s2);
// bool(false)
```

fiboknacky:
I'm aligned with Brent (incidentally again). :)

jtattermusch:
> @jtattermusch I think that `==` (comparison) and `===` (object identity) implement different operations? Why would object comparison be an anti-pattern?

[...]

PR #7883, opened 2020-09-11
Added support for "==" to the PHP C extension

haberman:
Prior to this CL, the `==` operator in PHP worked as expected for the pure-PHP library, but not for the C extension.

This PR fixes the C extension to have proper `==` semantics. It also adds tests that pass for both pure-PHP and the C extension.

Fixes: https://github.com/protocolbuffers/protobuf/issues/7650

### DIFF
--- a/php/tests/GeneratedClassTest.php
+++ b/php/tests/GeneratedClassTest.php
@@ -1530,6 +1530,148 @@ class GeneratedClassTest extends TestBase
     }
 
     #########################################################
+    # Test equality
+    #########################################################
+
+    public function testShallowEquality()
+    {
+        $m1 = new TestMessage([
+            'optional_int32' => -42,
+            'optional_int64' => -43,
+            'optional_uint32' => 42,
+            'optional_uint64' => 43,
+            'optional_sint32' => -44,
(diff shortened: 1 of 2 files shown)


php/tests/GeneratedClassTest.php
```diff
@@ -1529,6 +1529,148 @@ public function testValueIsReference()
...
+        $m2->getMapInt32Int32()[1234] = 1234;
+        $this->assertTrue($m1 != $m2);
+
+        # SubMessage inside map field is checked.
+        $m2 = new TestMessage();
+        $m2->mergeFromString($data);
+        $m2->getMapInt32Message()[1]->setA(1234);
+        $this->assertTrue($m1 != $m2);
+
+        # TODO: what about unknown fields?
```
TeBoring, line 1671: Question, is equal only recommend for test? If so, I think we don't need to worry about it.
Otherwise, I think not comparing equal ensures backward compatibility.

TeBoring (approved):

haberman:
I think I'll submit this for now and address the unknown field issue in a follow-up.

PR #7888, opened 2020-09-11
Ruby File Generation Frozen String Literals

noahmatisoff:
Adds a comment to the top of the files that are generated to enforce frozen string literals.

A pragma comment is an instruction for Ruby that when reading the file, it should treat it a certain way. They only impact that specific file, not any other files that it in turn loads. Seeking guidance from the team on if this should be an option passed when doing the code generation.

This only works for Ruby 2.3.0 and up, however it shouldn't have any adverse impact on versions prior to 2.3.0.


noahmatisoff:
@googlebot I signed it!

haberman:
What is the motivation for this? What will it help?

noahmatisoff:
It improves memory performance of strings by making them immutable by default, more can be read [here](https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/).

ryanseys:
Any update here?

acozzette (changes requested):
@noahmatisoff Thank you for the pull request. This change looks great but we just need an update to the test [here](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/ruby/ruby_generator_unittest.cc). It looks like it's just failing because there are some golden files that need to be updated.

acozzette:
@noahmatisoff I'm going to close this since it's been a few weeks with no action, but feel free to comment here if you would like to reopen this.

PR #7893, opened 2020-09-17
Add unit test for issue reproduction

PierrickVoulet:
PR with additional unit tests to reproduce https://github.com/protocolbuffers/protobuf/issues/7650#issuecomment-692921151


haberman:
I just tried to reproduce this, and for me it is passing for both pure-PHP and for the C extension!

Perhaps it was fixed by https://github.com/protocolbuffers/protobuf/pull/7900

PR #7901, opened 2020-09-21
Expose python Encode/Decode APIs

### LINKED ISSUE #54: Protobuf lacks framed stream of messages

eme-p:
Fixes #54


acozzette:

PR #1298, opened 2016-03-05
Add back the namespace_packages arg in setup.py.

craigcitro:
Improves #1296.

The problem: in the previous patch, we tweaked the `__init__.py` files to use namespaces, but no longer declared ourselves as a namespace package. The second half was unwise.

Note that this only comes up when installing `protobuf` alongside another package that also installs into the google namespace; as of right now, the only PyPI package that does is `googleapis-common-protos`, though the GAE SDK also uses `google.appengine`. However, this package is in fact a *dependency* of `googleapis-common-protos`, so that's an extremely likely combination. :wink:

With this PR, installing either or both of those alongside this package now works.

The case that still remains is the upgrade path, which is also what worried me in #713. It seems that if protobuf 2.6.1 is installed, there's no way to safely upgrade that to work with a newer protobuf. However, `pip uninstall` && `pip install` does the trick.

PTAL @haberman -- in particular, if we *really* want this issue to go away forever, we should also make sure to add a tests that (1) deal with the upgrade cases and (2) also install another package that uses the google namespace, eg googleapis-common-protos.

/cc @jonparrott @tseaver and @dhermes for some Python wisdom. @jonparrott, with this change, the cases in your gist now pass (since we've fixed the protobuf vs. googleapis conflict). However, I'm worried that we can't seem to upgrade from 2.6.1; is there any hope there? Have other python packages dealt with this?

I'd also be curious about an easy way to trigger the gcloud tests with this PR; could I just switch [this line](https://github.com/GoogleCloudPlatform/gcloud-python/blob/40332e17391369106aa4380eec780cd5c5deccf7/setup.py#L17) to point directly to the commit in my repo and send a PR, thereby letting Travis do its thing?

Issue #713, opened 2015-08-12
PyPI installation lacks __init__.py in top-level google directory


msaniscalchi:
This is a weird one, but I couldn't import google.protobuf for a while because the PyPI installation didn't include an __init__.py file in the top-level google directory. Doing a reinstall produced the same results; the top-level google directory still lacked an __init__.py file.

By looking at this repo, I would expect to see it there. Perhaps you're doing something in the setup.py that is stripping it out unintentionally?

dano:
setup.py installs the "google" package as a setuptools [namespace package](https://pythonhosted.org/setuptools/setuptools.html#namespace-packages), which means it gets installed without an `__init__.py`.

destijl:
+1 on getting this fixed. It's a common enough problem that there is a stack overflow answer:
http://stackoverflow.com/questions/13862562/google-protocol-buffers-not-found-when-trying-to-freeze-python-app

And I've had to work around this in three different build scripts, e.g:
https://github.com/google/grr/blob/master/vagrant/install_linux.sh#L166

haberman:
A PR would be most welcome! None of us are Python packaging experts, so fixing this ourselves often requires a fair amount of background research first. If anyone has a simple fix, I'm all ears.

xfxyjwf:
Seems related to: https://github.com/google/protobuf/issues/1153

craigcitro:
So adding the bits I know:

* The namespace packages PEP suggests using the [variant form](https://www.python.org/dev/peps/pep-0420/#namespace-packages-today) via `pkgutil`. 
* I suspect that we should also make a related change to switch from a namespace package to regular package, i.e. switch from `namespace_packages` to an explicit `packages` list or [just use `setuptools.find_packages()`](https://python-packaging-user-guide.readthedocs.org/en/latest/distributing/#packages). 

Now, the real question is this: what happens as we switch? If someone has one of the current versions installed, and we switch over, will everything _just work_? (Does that ever happen?)

haberman:
Closing in favor of https://github.com/google/protobuf/issues/1296

PR #7902, opened 2020-09-21
Remove uses of pkg_resources in non-namespace packages.

dlj-NaN:
In #713 and #1296, the `google` package in protobuf sources was found
to cause conflicts with other Google projects, because it was not
properly configured as a [namespace package]. The initial fix in
786f80f addressed part of the issue, and #1298 fixed the rest.

However, 786f80f (the initial fix) also made `google.protobuf` and
`google.protobuf.pyext` into namespace packages. This was not correct:
they are both regular, non-namespace, sub-subpackages.

However (still), the follow-up #1298 only nominated `google` as a
namespace packages, leaving the namespace registration behavior
for the other packages, without benefit (or need).

This change removes the unnecessary namespace registration, which has
substantial overhead, thus reducing startup time substantially when
using protobufs.

Because this change affects the import internals, quantifying the
overhead requires a full tear-down/start-up of the Python interpreter.
So, to capture the full cost for every run, I measured the time to
launching a _fresh_ Python instance in a subprocess, varying the
imports and code under test. In other words, I used `timeit` to
measure the time to launch a _fresh_ Python subprocess which actually
performs the imports.

* Reference: normal Python startup (i.e., don't import protobuf at all).
  ```
   % python3 -m timeit -s 'import subprocess' -r 3 -n 10 'subprocess.call(["python3", "-c", "pass"])'
  10 loops, best of 3: 27.1 msec per loop
  ```

* Baseline: cost to import `google.protobuf.descriptor`, with
  extraneous namespace packages.
  ```
  % python3 -m timeit -s 'import subprocess' -r 3 -n 10 'subprocess.call(["python3", "-c", "import google.protobuf.descriptor"])'
  10 loops, best of 3: 133 msec per loop
  ```

* This change: cost to import `google.protobuf.descriptor`, without
  extraneous namespace packages.
  ```
  % python3 -m timeit -s 'import subprocess' -r 3 -n 10 'subprocess.call(["python3", "-c", "import google.protobuf.descriptor"])'
  10 loops, best of 3: 43.1 msec per loop
  ```

[namespace package]:  https://packaging.python.org/guides/packaging-namespace-packages/

### DIFF
--- a/python/google/protobuf/__init__.py
+++ b/python/google/protobuf/__init__.py
@@ -31,9 +31,3 @@
 # Copyright 2007 Google Inc. All Rights Reserved.
 
 __version__ = '3.13.0'
-
-if __name__ != '__main__':
-  try:
-    __import__('pkg_resources').declare_namespace(__name__)
-  except ImportError:
-    __path__ = __import__('pkgutil').extend_path(__path__, __name__)
--- a/python/google/protobuf/pyext/__init__.py
+++ b/python/google/protobuf/pyext/__init__.py
@@ -1,4 +1,0 @@
-try:
-  __import__('pkg_resources').declare_namespace(__name__)
-except ImportError:
-  __path__ = __import__('pkgutil').extend_path(__path__, __name__)


haberman (approved):

acozzette:
I don't quite understand why, but I believe this change may have reintroduced the conflict with different packages sharing the `google` package. This came up in a gRPC test when I tested gRPC against the protobuf 3.14.0 release candidate here (see Bazel the test called `Basic Tests for Python (Local)`): https://github.com/grpc/grpc/pull/24723

dlj-NaN:
> I don't quite understand why, but I believe this change may have reintroduced the conflict with different packages sharing the `google` package.

Strictly speaking, this PR removed such a conflict, so that protobuf doesn't need to be the first thing imported under `google`.

> This came up in a gRPC test when I tested gRPC against the protobuf 3.14.0 release candidate here (see Bazel the test called `Basic Tests for Python (Local)`): [grpc/grpc#24723](https://github.com/grpc/grpc/pull/24723)

Those logs indicate that `google.auth` is now the order-dependent/conflicting package. You should be able to verify by swapping the `from google import auth` line with `import grpc` here:
https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests_aio/interop/methods.py#L28-L29

If that's the case, then google-auth can be fixed... but it would also indicate that the Bazel runner may be installing the package incorrectly.

vam-google:
(The same comment applies to https://github.com/protocolbuffers/protobuf/pull/7908)

@dlj-NaN Why was this migrated to [native-namespace-packages](https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages) instead of [pkgutil-style-namespace-packages](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages)? 
I understand the idea behind getting rid from the essentially deprecated [pkg-resources-style-namespace-packages](https://packaging.python.org/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages), but jumping directly to the `native namespace packages` looks too aggressive, while `pkgutil-style namespace packages` allows for a more  calm transition from a legacy package.

It seems like doing so would not break as many people as it did (including [googleapis/googleapis](https://github.com/googleapis/googleapis) build, which is why I'm here).

Grpc got broken by this: https://github.com/grpc/grpc/pull/25131#issuecomment-760544577
Envoy got broken by this: https://github.com/envoyproxy/envoy/pull/14253#discussion_r541350400
googleapis got broken by this as well: https://github.com/googleapis/gapic-generator/issues/3334

It looks like, at least for the bazel builds, the teams are "fixing" it by adding patches to the protobuf imports, which is not sustainable. I've tried converting the protobuf package to [pkgutil-style-namespace-packages](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages) instead of a native package and it seems fixing at least the googleapis' problem (there is a good chance that it will make patching unnecessary for other teams as well).

I understand that the ultimate solution would be migrating the other google-namespaced packages (like google-auth) but untill that  happens there will be a lot of friction and patches, which would be nice to avoid is possible.

dlj-NaN:
@vam-google : per the PR text, there were other packages were broken by the pkgutil-style namespacing (and not using Bazel). That was my overriding concern.

Looking at googleapis/gapic-generator-python#753, I don't see any details of the error you saw, so I can really only guess at specifics. It looks like the upshot is that protobuf was, at one point, winning a global-transitive-import-order race, but only by chance, and not by right -- the race depends on the global ordering of imports, transitively. (Python imports are usually notionally topological, but in this case the import machinery is being hooked _somewhere_, so ordering matters.)

Incidentally, between `pip` and `wheel`, the pkgutil-style namespace `__init__.py` files are usually dropped altogether. Switching to `pip_install` doesn't make pkgutil logic reappear, but it certainly does affect search paths (specifically, it _additionally_ changed the order in which paths are searched). I think your PR just made some other package win the global import ordering race for the `google` namespace.

PR #7904, opened 2020-09-22
Add support for new compilers when generating c++ message classes

vesavlad:
This pull request includes fixes for generated c++ message classes that refer to:
- mark methods that override functionality from base class as "override"
- fix of pedantic warnings generated by declaring of arrays with 0 length - there is a language extension for this but


vesavlad:
@googlebot I signed it!

PR #7905, opened 2020-09-22
Add support for new compilers when generating c++ message classes

vesavlad:
This pull request includes fixes for generated c++ message classes that refer to:
 - mark methods that override functionality from base class as "override"
 - fix of pedantic warnings generated by declaring of arrays with 0 length - there is a language extension for this but having the pedantic warnings enabled together with -Werror will result in compilation errors in projects that depend on this library & the generated code

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -296,6 +296,26 @@ bool ShouldMarkClearAsFinal(const Descriptor* descriptor,
          options.opensource_runtime;
 }
 
+bool ShouldMarkClearAsOverride(const Descriptor* descriptor,
+                            const Options& options) {
+  static std::set<std::string> exclusions{
+  };
+
+  const std::string name = ClassName(descriptor, true);
+  return exclusions.find(name) == exclusions.end() ||
+         options.opensource_runtime;
+}
+
+bool ShouldMarkIsInitializedAsOverride(const Descriptor* descriptor,
+                                    const Options& options) {
+  static std::set<std::string> exclusions{
+  };
+
+  const std::string name = ClassName(descriptor, true);
+  return exclusions.find(name) == exclusions.end() ||
+         options.opensource_runtime;
+}
+
 bool ShouldMarkIsInitializedAsFinal(const Descriptor* descriptor,
                                     const Options& options) {
   static std::set<std::string> exclusions{
@@ -1318,11 +1338,11 @@ void MessageGenerator::GenerateClassDefinition(io::Printer* printer) {
       "\n"
       "// implements Message ----------------------------------------------\n"
       "\n"
-      "inline $classname$* New() const$ new_final$ {\n"
+      "inline $classname$* New() const$ new_final$ override {\n"
       "  return CreateMaybeMessage<$classname$>(nullptr);\n"
       "}\n"
       "\n"
-      "$classname$* New(::$proto_ns$::Arena* arena) const$ new_final$ {\n"
+      "$classname$* New(::$proto_ns$::Arena* arena) const$ new_final$ override {\n"
       "  return CreateMaybeMessage<$classname$>(arena);\n"
       "}\n");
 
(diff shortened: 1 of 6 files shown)


src/google/protobuf/compiler/cpp/cpp_message.cc
```diff
@@ -296,6 +296,26 @@ bool ShouldMarkClearAsFinal(const Descriptor* descriptor,
          options.opensource_runtime;
 }
 
+bool ShouldMarkClearAsOverride(const Descriptor* descriptor,
```
acozzette, line 299: Our style guide says to annotate overrides as `final` or `override` but not both: https://google.github.io/styleguide/cppguide.html#Inheritance So I'm afraid that we don't want to mark methods as `override` given that they're already marked `final`.

vesavlad: Clang wants the override keyword, for clang "final" is not enough --> as both needed
 - next final means, this method can no longer be overridden, it does not mean that this method is actually an override
 - so the use of final does not necessarily mean that on the method i stick it too, that this one itself is overriding something

Would an acceptable solution be to remove the **final** keyword if anything else is accepted.

> Explicitly annotate overrides of virtual functions or virtual destructors with exactly one of an override or (less frequently) final specifier. Do not use virtual when declaring an override. Rationale: A function or destructor marked override or final that is not an override of a base class virtual function will not compile, and this helps catch common errors. The specifiers serve as documentation; if no specifier is present, the reader has to check all ancestors of the class in question to determine if the function or destructor is virtual or not.

As this states: _with exactly one of an override or **(less frequently)** final specifier._

acozzette: I think we need them to stay `final` because there's an important performance win from giving that information to the compiler. I believe the rationale behind the style guide rule is that if you leave out the `virtual` keyword then `final` does mean it has to be an override (without the `virtual` keyword, it would be an error to use `final` if it's not overriding something).

What error flags do you pass to make clang complain about `final` without `override`? That seems surprising to me that clang would warn about this. I see that by default clang-tidy will actually complain if you use both: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html#cmdoption-arg-allowoverrideandfinal

vesavlad: The complain is not about `final` without `override`, it is related to overriding a method without specifying the `override` keyword and only the `final` one.

Also when compiling the generated code with the GCC compiler in this case the `-Wsuggest-override` is not so smart as clang to handle also the final keyword as valid option for specifying the override. Compiler flags used in the referred project are:
```cpp
    -Wall 
    -Wextra 
    -Werror
    -ftemplate-depth=225 
    -Wno-unknown-warning-option 
    -Wnon-virtual-dtor
    -Wsuggest-override
    -Wno-unused-local-typedefs
    -Wno-maybe-uninitialized
    -Wno-implicit-fallthrough
    -Wno-noexcept-type
    -pedantic-errors
```

Taking this for example:
```cpp
class A {
public:
    virtual ~A() = default;
    virtual void f() = 0;
    virtual void g() = 0;
};

class B : public A {
public:
    virtual ~B() = default;
    void f() final override;
    void g() final;
};

void foo(B& b) {
    b.f();
    b.g();
}
```

And compiling it with clang this will work but when compiling with GCC with the specified flags this will no longer work because of the `override` that is needed due to `-Wsuggest-override`.
You can check the output here: https://gcc.godbolt.org/#compilers:!((compiler:g6,options:'-O2+-Wall++-xc%2B%2B+-std%3Dc%2B%2B11',source:'class+A+%7B%0Apublic:%0A++++virtual+void+f()+%3D+0%3B%0A%7D%3B%0A%0Aclass+B+:+public+A+%7B%0Apublic:%0A++++void+f()+final+override%3B%0A%7D%3B%0A%0Avoid+foo(B%26+b)+%7B%0A++++b.f()%3B%0A%7D%0A')),filterAsm:(commentOnly:!t,directives:!t,intel:!t,labels:!t),version:3

The support of final as an overriding keyword for virtual functions seems to be added starting with GCC-9.2 but we need to stick to an older version for now. Also the code build is cross-compiled with and arm-none compiler which behaves the same as GCC prior to 9.2 when it comes to overriding.

acozzette: Thanks for the detailed explanation. I think we want to stick with just `final`, but if you need to you can always make a local modification to use `override` instead.

PR #7907, opened 2020-09-23
Update default dependency for gtest.

dlj-NaN:
The gtest source was changed in #7237 on an opt-in basis, and has been
released since 3.12.0. The comments state that the default should change
in 3.13.0, but that didn't quite happen.

This change does flip the default, and updates comments to say 3.14.0.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -14,10 +14,10 @@ exports_files(["LICENSE"])
 # build configuration
 ################################################################################
 
+# TODO(yannic): Remove in 3.14.0.
 string_flag(
     name = "incompatible_use_com_google_googletest",
-    # TODO(yannic): Flip to `true` for `3.13.0`.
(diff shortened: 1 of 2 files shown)


BUILD
```diff
@@ -14,10 +14,10 @@ exports_files(["LICENSE"])
 # build configuration
 ################################################################################
 
+# TODO(yannic): Remove in 3.14.0.
```
acozzette, line 17: Should we just remove all these now since 3.14.0 will be the next release?

acozzette (approved):

BUILD
dlj-NaN: Seems like a good idea... I'll do that separately, though, in case anyone is relying on them. That will at least give a couple of commits to find the problem and flip behavior with just a flag.

Issue #7877, opened 2020-09-08
Python: with the generate_py_protobufs extension installed, cannot find google.api_core on import


bobhancock:
**What version of protobuf and what language are you using?**
Version: master/v3.13.0.
Language: Python
OS: Linux
Python 3.7.0 and 3.8.5
tl;dr
If you import google.protobuf or proto prior to google.api_core, the module is found and loaded..
If you import google.api_core before importing these two if you attempt to import google.api_core the module is not found.
Why isn’t google.api_core found with the PathFinder in sys.meta_path find it?
I can duplicate this on my desktop, on a fresh cloudtop, and on our kokoro-ubuntu machines.

```
rwh@rhw8:python_cpp_pyext$ source /tmp/cpp-pyext-O2Xfh0IBIS/cpp_pyext/bin/activate
(cpp_pyext) rwh@rhw8:python_cpp_pyext$ python
Python 3.8.5 (default, Sep  5 2020, 16:13:02) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> import google.api_core
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'google.api_core'
>>> sys.meta_path
[<class '_frozen_importlib.BuiltinImporter'>, <class '_frozen_importlib.FrozenImporter'>, <class '_frozen_importlib_external.PathFinder'>]
>>> sys.path
[...]

PR #7908, opened 2020-09-23
[bazel/py] Omit google/__init__.py from the Protobuf runtime.

dlj-NaN:
Since `google` is a Python namespace package, the `google/__init__.py`
file should be omitted to avoid collisions. For example, its presence
may cause other Pip packages in the google namespace not to be found.

This change narrows which `__init__.py` files are included in the
`//:protobuf_runtime` target to include only those under
`google/protobuf`. It also moves their canonical inclusion to the
internal `//:python_srcs` rule, eliminating the dual specification in
`python_protobuf.extra_srcs`.

This is the difference in files, lexicographically ordered:

```
$ bazel cquery 'labels(srcs, kind(py_library, deps(:protobuf_python)))' | sort > /tmp/deps-{before,after}.txt
$ diff /tmp/deps-{before,after}.txt
1,5d0
< //:python/compatibility_tests/v2.5.0/tests/__init__.py (null)
< //:python/compatibility_tests/v2.5.0/tests/google/__init__.py (null)
< //:python/compatibility_tests/v2.5.0/tests/google/protobuf/__init__.py (null)
< //:python/compatibility_tests/v2.5.0/tests/google/protobuf/internal/__init__.py (null)
< //:python/google/__init__.py (null)
51d45
< //:python/protobuf_distutils/protobuf_distutils/__init__.py (null)
```

This seems like a desirable change, since it avoids polluting other
packages, too.

This is conceptually similar to #7877, but for Bazel.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -739,10 +739,9 @@ py_library(
     name = "python_srcs",
     srcs = glob(
         [
-            "python/google/**/*.py",
+            "python/google/protobuf/**/*.py",
         ],
         exclude = [
-            "python/google/protobuf/**/__init__.py",
             "python/google/protobuf/internal/*_test.py",
             "python/google/protobuf/internal/test_util.py",
         ],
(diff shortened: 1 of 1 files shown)


dlj-NaN:
I'm not sure if this actually warrants release notes, but we could say something like "we don't break the google package namespace anymore in Python builds under Bazel."

acozzette (approved):

PR #7910, opened 2020-09-23
fix: dynamic field generation in python respects json_name

software-dov:
This bug is visible if a user creates a FileDescriptorProto,
DescriptorProto, and FieldDescriptorProto dynamically and adds the
file descriptor to a descriptor pool.
The field descriptor has a custom json_name.
The user then retreives the
message descriptor from the pool via FindMessageTypeByName and queries
the Descriptor's fields.
The FieldDescriptor should preserve the custom json_name.

### DIFF
--- a/python/google/protobuf/descriptor_pool.py
+++ b/python/google/protobuf/descriptor_pool.py
@@ -1003,6 +1003,7 @@ class DescriptorPool(object):
         is_extension=is_extension,
         extension_scope=None,
         options=_OptionsOrNone(field_proto),
+        json_name=field_proto.json_name,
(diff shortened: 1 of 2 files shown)


dlj-NaN:
Adding Josh... I'm not exactly sure where the json name transformation should happen (out of FieldOptions and into FieldDescriptorProto). I think this change correctly reflects the C++ side, but this behavior may be wrong in both cases... thoughts, Josh?

software-dov:
Open review bump.

software-dov:
Open review bump.

Issue #7140, opened 2020-01-27
gcc 9 -Warray-bounds warnings about memset() in generated code


dkl:
**What version of protobuf and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Ubuntu 19.10 on x86_64

**What runtime / compiler are you using (e.g., python version or gcc version)**
g++ (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008

**What did you do?**
Used a proto file such as the following:
```
syntax = "proto2";

message Message1 {
    required int32 a = 1;
    required int32 b = 2;
}
```
and compiled it with `protoc foo.proto --cpp_out=. && g++ -O3 -Wall foo.pb.cc -c`

**What did you expect to see**
compiles without warnings

**What did you see instead?**
```
In file included from /usr/include/string.h:494,
                 from /usr/include/c++/9/cstring:42,
                 from /usr/local/include/google/protobuf/io/coded_stream.h:118,
                 from foo.pb.h:23,
                 from foo.pb.cc:4:
In function ‘void* memset(void*, int, size_t)’,
    inlined from ‘void Message1::SharedCtor()’ at foo.pb.cc:111:11,
    inlined from ‘Message1::Message1()’ at foo.pb.cc:96:13,
    inlined from ‘void InitDefaultsscc_info_Message1_foo_2eproto()’ at foo.pb.cc:94:1:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:71:33: warning: ‘void* __builtin_memset(void*, int, long unsigned int)’ offset [29, 32] from the object at ‘_Message1_default_instance_’ is out of the bounds of referenced subobject ‘Message1::a_’ with type ‘int’ at offset 24 [-Warray-bounds]
   71 |   return __builtin___memset_chk (__dest, __ch, __len, __bos0 (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It warns about memset() being called on a field, but writing not just to that field, but also to following fields. See also: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90376

The warning doesn't appear when using -O2 instead of -O3, but -O3 is common in CMake release builds (which is how it showed up here).

**Anything else we should know about your project / environment**

clementperon:
Hi,

[...]

PR #7913, opened 2020-09-24
Fix: gcc 9 -Warray-bounds warnings about memset() in generated code

vesavlad:
Starting with gcc9 the compilation of generated message classes throws the following error:
` warning: ‘void* __builtin_memset(void*, int, long unsigned int)’ offset [29, 32] from the object at ‘_Message1_default_instance_’ is out of the bounds of referenced subobject ‘Message1::a_’ with type ‘int’ at offset 24 [-Warray-bounds]
   71 |   return __builtin___memset_chk (__dest, __ch, __len, __bos0 (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~`

This fix aims to remove this warning and also maintain backwards compatibility with older compilers, therefore fixing the #7140 issue


acozzette:
Thanks, @vesavlad.

PR #7916, opened 2020-09-25
Reserve 1088 for Confluent Schema Registry

rayokota:


rayokota:
@googlebot I signed it!

PR #7927, opened 2020-09-29
Change Python kokoro configs to test only their specified version.

dlj-NaN:
Our versioned Python tests have been using the common "linux/64-bit" runner
with the same `kokoro/linux/python_cpp/build.sh` build script. Unfortunately,
using non-versioned `build_file` means every Python version is tested (under
Tox), multiplied across each Python version-specific build.

For example, this means that when the `python38` config is used to run Tox, the
Tox runner actually uses the default Python from `kokoro/linux/64-bit`, which
then tests all of the Python versions individually, via Tox.

This change fixes the `build_file` paths to point to the version-specific build
scripts. Because all of the builds were running on the `linux/64-bit` image, this
change also updates the versioned build containers to install tox.

Some tox.ini changes are needed to avoid breakage (see also #7930).

### DIFF
--- a/kokoro/linux/dockerfile/test/python27/Dockerfile
+++ b/kokoro/linux/dockerfile/test/python27/Dockerfile
@@ -20,4 +20,12 @@ RUN apt-get update && apt-get install -y \
   parallel \
   time \
   wget \
-  && apt-get clean
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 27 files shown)


dlj-NaN:
FYI, the latest Kokoro tests used the Docker images with the changes in this patch. Only the Python images were rebuilt (the macOS failures in particular are pre-existing).

haberman (approved):

PR #7938, opened 2020-10-02
Override Map.clone to use Map's dup method

alusco-scratch:
Message and Repeated field override clone so that it uses the internal implementation of dup but Map is missing this and only implements dup.  This can lead to unexpected behavior since two out of three complex types behave correctly.


alusco-scratch:
Signed a CLA

alusco-scratch:
@googlebot I signed it!

PR #7954, opened 2020-10-09
Fix finding google/protobuf/descriptor.proto on FreeBSD

### LINKED ISSUE #7948: protoc cannot find descriptor.proto on FreeBSD
Version: v3.13.0
Language: C++

FreeBSD 12.1-RELEASE

**What did you do?**
I try to compile proto file which imports descriptor.proto

On Linux I can make it without errors

**What did you see instead?**
```
user@freebsd12:~% cat t.proto
syntax = "proto2";
message t {}
user@freebsd12:~% protoc --cpp_out=. t.proto
google/protobuf/descriptor.proto: File not found.
t.proto:2:1: Import "google/protobuf/descriptor.proto" was not found or had errors.
```

protoc is in /usr/local/bin/protoc, descriptor.proto is in /usr/local/include/google/protobuf/descriptor.proto.
It seems that the problem is connected with GetProtocAbsolutePath which is not implemented really for FreeBSD: https://github.com/protocolbuffers/protobuf/blob/93d8e89374947aa6c05c290b08efd211c6703087/src/google/protobuf/compiler/command_line_interface.cc#L190-L215

alxrmorozov:
`size_t` is used for `len` because otherwise I get:
`error: cannot convert 'int*' to 'size_t*' {aka 'long unsigned int*'}`

Fixes #7948

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -58,8 +58,10 @@
 
 #include <memory>
 
-#ifdef __APPLE__
+#if defined(__APPLE__)
 #include <mach-o/dyld.h>
+#elif defined(__FreeBSD__)
+#include <sys/sysctl.h>
 #endif
 
 #include <google/protobuf/stubs/common.h>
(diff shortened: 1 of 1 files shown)


alxrmorozov:
@googlebot I signed it!

PR #7965, opened 2020-10-14
Upgrade six dependency to 1.15.0.

JoshuaCrestone:
Six 1.12.0 conflicts with the latest versions of google-cloud-python: https://github.com/googleapis/google-cloud-python/issues/9965

Is there any reason not to upgrade?


acozzette:
Is this change still needed? We no longer support Python 2, so at this point we might as well drop our dependency on six.

PR #7969, opened 2020-10-14
Add a timeout field to MethodOptions

swankjesse:
Specifying timeouts in the service declaration allows client and
server teams to agree on expectations, and allows software to
implement these expectations.

### DIFF
--- a/src/google/protobuf/descriptor.proto
+++ b/src/google/protobuf/descriptor.proto
@@ -710,6 +710,19 @@ message MethodOptions {
   optional IdempotencyLevel idempotency_level = 34
       [default = IDEMPOTENCY_UNKNOWN];
 
+  // The default duration a client should wait for an RPC call to complete. This may be overridden
+  // by clients at runtime. If no timeout is specified the effective timeout is
+  // implementation-dependent.
+  //
+  // If a response is not received within the client's timeout, the client may treat the call as
+  // failed. Timeouts are enforced by clients and include the transport time of messages over the
+  // network. If a client has a particularly high-latency connection to the server, the amount of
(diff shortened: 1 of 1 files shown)


oldergod:
@acozzette would you have feedback on that? Does that seem good?

acozzette:
Thank you for the pull request but I don't think we want to go in this direction. RPC systems like gRPC that are built on top of protobuf already have their own ways of handling timeouts, so I don't think it is a good idea to add a competing mechanism. If you want your .proto file to include timeouts, you could always do this for an individual project by defining your own [custom option](https://developers.google.com/protocol-buffers/docs/proto#customoptions).

swankjesse:
I was hoping gRPC could honor these timeouts. Rather than clients and servers having to agree on appropriate timeouts in code, they should agree in the schema.

acozzette:
You could try to propose the idea to the gRPC team, but I would guess that they want the timeouts to be configurable on a per-request basis.

swankjesse:
Proposing this to the gRPC team is what I was hoping to do with this PR. Is there another repo that'd be more appropriate?

haberman:
I think the gRPC repo at https://github.com/grpc/grpc would be a better place for gRPC-specific questions.

If the gRPC team wanted to do this, they could add their own options that are separate from `descriptor.proto`. So there is no need for a change in our repo.

togiles:
@swankjesse curious if you took this any further with the gRPC team, considering defining our own custom option for this purpose + adding a custom interceptor for our services to enforce the default timeout if one is not set by the client already.

swankjesse:
@togiles I made no further attempt after this change.

Issue #3012, opened 2017-04-25
JsonFormat.Parser throws InvalidProtocolBufferException when parsing unknown enum constants


veerenmandalia:
I am using [JsonFormat.Parser](https://developers.google.com/protocol-buffers/docs/reference/java/) to convert JSON to Java protobuf models.

When it encounters an enum constant it doesn't recognize, the parser throws an InvalidProtocolBufferException. I would expect it to serialize the `UNRECOGNIZED` enum constant defined by protobuf. This is seen in the [JsonFormat source](https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L1670)

Example:

**Generated Java model**
```
public enum Vehicle implements com.google.protobuf.ProtocolMessageEnum {
  BICYCLE(0),
  CAR(1),
  UNRECOGNIZED(-1);
}
```

If server then sends the JSON `{ vehicle = "TRUCK" }` an exception is thrown rather than deserializing it as UNRECOGNIZED:
```
com.google.protobuf.InvalidProtocolBufferException: Invalid enum value: TRUCK for enum type: messages.company.Vehicle
```

**JSON parser configuration**
```
try {
    JsonFormat.parser()
        .ignoringUnknownFields()
        .merge(json, builder);
} catch (InvalidProtocolBufferException e) {
    throw new IOException(e);
}
```

BSBandme:
Any updates on this? Will your PR be merged?

veerenmandalia:
Thanks @vijay-bhatt for completing this in https://github.com/protocolbuffers/protobuf/pull/4825

The change did not make 3.6.1. This can be closed either now or once the change is released (not sure the process).

BSBandme:
Closed as the change should have been released.

PR #7974, opened 2020-10-17
JsonParser.withIgnoreUnknownFields true will now ignore unknown enums.

doom-goober:
There has been some debate in the Protobuf community about whether ignoreUnknownFields should ignore unknown JSON enum representations. Here's the Java unknown enum value issue: https://github.com/protocolbuffers/protobuf/issues/3012 And here's the Swift JSON Protobuf team debating about unknown enums: https://github.com/apple/swift-protobuf/issues/972

The Protobuf spec leaves it up to each language to decide the behavior https://developers.google.com/protocol-buffers/docs/proto3:
> During deserialization, unrecognized enum values will be preserved in the message, though how this is represented when the message is deserialized is language-dependent. In languages that support open enum types with values outside the range of specified symbols, such as C++ and Go, the unknown enum value is simply stored as its underlying integer representation. In languages with closed enum types such as Java, a case in the enum is used to represent an unrecognized value, and the underlying integer can be accessed with special accessors. In either case, if the message is serialized the unrecognized value will still be serialized with the message.

The justification for handling unrecognized enums is for backwards compatibility. For example, a client written against data v1 encounters a new data v2 that has newly added enum values that didn't exist in v1. The client will fail to parse, breaking backward compatibility. Instead, the unknown string could be replaced with some known enum value.

In my pull request, I have simply replaced the unknown enum value with enum value 0. Obviously, this is just a proof of concept as it somewhat breaks backwards compatibility with people who expect an InvalidProtocolBufferException to throw with unrecognized enums. So, to maintain backward compatibility, here are some proposals:

1) Add a JsonParser.Setting called "ignoreUnknownEnums" and set unknown enums to value 0. This is backwards compatible but any protos that use the enum value 0 can be confused as to whether they are unrecognized or simply the default value.
2) Allow an option on enums that specifies the "unrecognized" value for that enum type. If ignoreUnknownFields = true and that option is present, use the enum value specified in the option as the "unrecognized" value.
3) Introduce the concept of "unrecognized" value in enums. This is what Java does, however, it would also break backwards compatibility since Java's unrecognized is -1, and any C# clients that use -1 enum values will conflict.

Solution 1, 2, or 3 would solve the problem for my team. I believe #2 is the best solution but I don't know how to get an option allocated for this use. (#1 is the easiest solution, #3 is nice and similar to Java but causes problems and is quite involved.)

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
@@ -976,6 +976,26 @@ namespace Google.Protobuf
         }
 
         [Test]
+        public void UnknownEnum_Ignored()
+        {
+            var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(true));
+            string json = "{ \"single_nested_enum\": \"NOT_VALID_ENUM\"" + ", \"singleString\": \"x\" }";
+            var actual = parser.Parse<TestAllTypes>(json);
+            var expected = new TestAllTypes { SingleString = "x" };
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void KnownEnum_NotIgnored()
+        {
+            var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(true));
+            string json = "{ \"single_nested_enum\": \"FOO\"" + ", \"singleString\": \"x\" }";
+            var actual = parser.Parse<TestAllTypes>(json);
+            var expected = new TestAllTypes { SingleNestedEnum = TestAllTypes.Types.NestedEnum.Foo, SingleString = "x" };
+            Assert.AreEqual(expected, actual);
+        }
+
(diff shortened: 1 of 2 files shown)


doom-goober:
@googlebot I signed it!

doom-goober:
@googlebot I fixed it.

alex-roboto:
@googlebot I consent.

smohring:
Is there a way to accelerate this?

ifabijanovic:
Hello, trying to give some visibility here, this would be a very useful addition

acozzette:
This change looks like a good direction, but I think we need to hold off for now because there are some active discussions going on about how we want to handle unknown enums in JSON across all languages.

acozzette:
#7768 is another possible solution to the unknown enum problem.

jskeet:
I've got an internal change progressing which will tackle this, in a slightly different way. I'm hoping we can get that in soon and close this and #7768.

PR #7988, opened 2020-10-27
Spell check only comments and strings in vim

lakshayg:
The vim syntax file allows specifying which regions to spell
check and which ones to not. This patch limits spell checking
to comments and strings


acozzette:
Thanks, @lakshayg.

Issue #8001, opened 2020-11-02
PHP C Extension doesn't implement hasOneof()


haberman:
The generated code contains calls to `$this->hasOneof()`, but that function is not implemented in the C extension.

This causes errors if a user calls the `hasFoo()` method for any oneof field with the C extension.

PierrickVoulet:
This seems to be a duplicate of #7961

haberman:
This was fixed in #8003.

Issue #7961, opened 2020-10-13
v3.13.0.1 from PECL fails due to missing `Message::hasOneOf()` method


PierrickVoulet:
Following up on #7926, initially reported by https://github.com/googleads/google-ads-php/issues/432#issuecomment-707673553 

**What version of protobuf and what language are you using?**
Version: v3.13.0.1
Language: PHP
Implementation: C

**What operating system (Linux, Windows, ...) and version?**
Linux

**What did you do?**
Run some code that uses a `hasXXX` generated method which calls `Message::hasOneOf()`:
1. It runs fine when using the pure-PHP implementation
2. It fails when using the C implementation with the following message (see related [source](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/V5/Resources/Ad.php#L771)):

```
PHP Fatal error:  Uncaught Error: Call to undefined method Google\Ads\GoogleAds\V5\Resources\Ad::hasOneof() in /usr/local/google/home/pierrick/git/gob/ga/google-ads-php/src/Google/Ads/GoogleAds/V5/Resources/Ad.php:771
Stack trace:
#0 /usr/local/google/home/pierrick/git/gob/ga/google-ads-php/examples/BasicOperations/GetExpandedTextAds.php(131): Google\Ads\GoogleAds\V5\Resources\Ad->hasExpandedTextAd()
```

**What did you expect to see**
A successful run when using the C implementation like it is the case when using the pure-PHP implementation.

PierrickVoulet:
This seems to be fixed, I do not reproduce this issue with the version 3.14.0 (tested with both the PHP and C implementations).

Could you confirm?

leongersen:
I can confirm this is resolved with the updated extension. Thanks!

PR #8003, opened 2020-11-03
add missing has<Field>() method for oneof fields

anight:
Possible solution for #8001 and #7961

test.proto:
```proto
syntax = "proto3";
package Test;
message Test {
        oneof O {
                int32 A = 1;
                int32 B = 2;
        }
}
```

test.php:
```php
<?php
        dl('protobuf.so');
        require_once './vendor/autoload.php';
        $t = new Test\Test();
        print_r($t->hasB() === false);
?>
```

Output:
```shell
$ php ./test.php
Fatal error: Uncaught Error: Call to undefined method Test\Test::hasOneof() ...
```

Expected output:
```shell
$ php ./test.php
1
```

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -905,6 +905,39 @@ PHP_METHOD(Message, whichOneof) {
 }
 
 /**
+ * Message::hasOneof()
+ *
+ * Returns the presense of the given oneof field, given a field number. Called
+ * from generated code methods such as:
+ *
+ *    public function hasDoubleValueOneof()
+ *    {
+ *        return $this->hasOneof(10);
+ *    }
+ *
+ * @return boolean
+ */
+PHP_METHOD(Message, hasOneof) {
+  Message* intern = (Message*)Z_OBJ_P(getThis());
+  zend_long field_num;
+  const upb_fielddef* f;
+
+  if (zend_parse_parameters(ZEND_NUM_ARGS(), "l", &field_num) == FAILURE) {
+    return;
+  }
+
+  f = upb_msgdef_itof(intern->desc->msgdef, field_num);
+
+  if (!f || !upb_fielddef_realcontainingoneof(f)) {
+    php_error_docref(NULL, E_USER_ERROR,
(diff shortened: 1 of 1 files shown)


php/ext/google/protobuf/message.c
```diff
@@ -904,6 +904,39 @@ PHP_METHOD(Message, whichOneof) {
   RETURN_STRING(field ? upb_fielddef_name(field) : "");
 }
 
+/**
+ * Message::hasOneof()
+ *
+ * Returns the presense of the given oneof field, given a field number. Called
```
haberman, line 910: Can you fix the spelling error here? `presense ==> presence`

php/ext/google/protobuf/message.c
```diff
@@ -904,6 +904,39 @@ PHP_METHOD(Message, whichOneof) {
...
+ * from generated code methods such as:
+ *
+ *    public function hasDoubleValueOneof()
+ *    {
+ *        return $this->hasOneof(10);
+ *    }
+ *
+ * @return boolean
+ */
+PHP_METHOD(Message, hasOneof) {
```
haberman, line 920: Can you add a test for this?

haberman (commented):
This is great, thanks! Just a few small comments.

php/ext/google/protobuf/message.c
haberman: Test looks great, but it might be better to move it into one of the existing test files (I think `GeneratedClassTest.php` would be a good option). Adding new files is a bit more involved because then the new file has to get added to the Makefile:

https://github.com/protocolbuffers/protobuf/blob/master/Makefile.am#L767-L961

haberman:
I kicked off the tests. If you move your new test into an existing file, it should fix the "distcheck" test failure.

haberman:
We'd like to cut a branch for our upcoming release today and this fix should definitely go in. Are you planning to push again in the next few hours? If not I may need to adopt this change in a different PR to get it merged.

haberman:
Actually let me merge this now and the send a follow-up PR to fix the test failures.

PR #8013, opened 2020-11-06
Deduplicate string suffix predicates

georgthegreat:


acozzette:
Thanks, @georgthegreat.

PR #8014, opened 2020-11-06
Deduplicate StripProto methods

georgthegreat:
There was a long TODO entry for this deduplication (see diff)

### DIFF
--- a/src/google/protobuf/compiler/code_generator.h
+++ b/src/google/protobuf/compiler/code_generator.h
@@ -193,6 +193,9 @@ typedef GeneratorContext OutputDirectory;
 PROTOC_EXPORT void ParseGeneratorParameter(
     const std::string&, std::vector<std::pair<std::string, std::string> >*);
 
+// Strips ".proto" or ".protodevel" from the end of a filename.
+PROTOC_EXPORT std::string StripProto(const std::string& filename);
+
 }  // namespace compiler
 }  // namespace protobuf
 }  // namespace google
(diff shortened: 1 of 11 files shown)


georgthegreat:
This PR will conflict with (much simpler #8013).
I think #8013 should be merged first (I will update this PR upon commit landing).

src/google/protobuf/compiler/code_generator.h
```diff
@@ -193,6 +193,9 @@ typedef GeneratorContext OutputDirectory;
 PROTOC_EXPORT void ParseGeneratorParameter(
     const std::string&, std::vector<std::pair<std::string, std::string> >*);
 
+// Strips ".proto" or ".protodevel" from the end of a filename.
```
georgthegreat, line 196: Method placement might be somewhat strange, as compiler part does not have generic utility file (e. g. to common.h or util.h).

Suggestions are welcome.

georgthegreat:
Rebased onto new master branch thus resolving the conflicts with #8013

src/google/protobuf/compiler/cpp/cpp_helpers.h
```diff
@@ -205,9 +205,6 @@ inline const Descriptor* FieldScope(const FieldDescriptor* field) {
 std::string FieldMessageTypeName(const FieldDescriptor* field,
                                  const Options& options);
 
-// Strips ".proto" or ".protodevel" from the end of a filename.
-PROTOC_EXPORT std::string StripProto(const std::string& filename);
```
acozzette, line 209: I think we still need to keep this one around unfortunately, because within Google we have a bunch of code that calls this particular one. Could you perhaps make this one just delegate to the new version from code_generator.h?

georgthegreat: I have added a proxy method which is easy to remove once internal codebase will be adjusted.
Will this work?

Did not bother about ABI compatibility (i. e. PROTOC_EXPORT / placement the method into .cc file) as entire patch will break ABI anyway.

acozzette:
@georgthegreat Thanks for this cleanup!

PR #8020, opened 2020-11-10
Fix pecl distro

tony2001:
Latest protobuf version from PECL fails to build because of missing file wkt.inc.
This patch adds it to the tar archive.


tony2001:
@googlebot I signed it!

acozzette:
@tony2001 Thank you for the fix but I believe this issue has already been fixed in #8022.

tony2001:
Looks like it was, indeed.
Ok, waiting for the new version then.

PR #8031, opened 2020-11-12
Fix bug in parsing delimited messages at the end of an input stream

rwkeane:
Currently, there is a bug in parsing of delimited messages near the end of an input stream. For example:

Assume that an input stream has 3 remaining bytes:
  `7 4 12`
Calling  ParseDelimitedFromCodedStream() on this input will return true if `4 12` parses to a valid message, even though the input isn't of length 7, because <= 7 bytes are parsed

This PR adds validation that the length of the message matches what is denoted by the length byte(s)

### DIFF
--- a/src/google/protobuf/util/delimited_message_util.cc
+++ b/src/google/protobuf/util/delimited_message_util.cc
@@ -74,6 +74,8 @@ bool ParseDelimitedFromCodedStream(MessageLite* message,
     return false;
   }
 
+  // Get the position after any size bytes have been read (and only the message
+  // itself remains).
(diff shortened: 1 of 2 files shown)


gerben-s:
Can you add a unittest showing this.

rwkeane:
> Can you add a unittest showing this.

Test added. PTAL

rwkeane:
Ping for next steps

FWIW, the same CL for other protobuf versions merged in already: cl/342360226

acozzette:
It is probably safe to close this because this fix is going to be included in #8076.

rwkeane:
Great, thank you!

Issue #6782, opened 2019-10-22
Support using python memoryview for setting bytes data


wmdrichards:
**What language does this apply to?**
Python

**Describe the problem you are trying to solve.**
Protobuf serialization of large byte data from a numpy array requires seemingly unnecessary copies (calling `ndarray.tobytes()`)

**Describe the solution you'd like**
Support setting bytes data using a `memoryview`

**Additional context**
Passing `ndarray.data` to a protobuf class raises `TypeError: <memory at 0x119b1f588> has type memoryview, but expected one of: bytes`

fandreuz:
Hi @acozzette @haberman @zhangskz, I'm reaching out since I see you were active under #8047. I propose to implement `memoryview` support for `bytes` fields only, it should be much more straightforward than having it for repeated scalars.

I have a draft implementation at https://github.com/protocolbuffers/protobuf/commit/821695fa7f57f93032bac821c00762a90fe57d6f, let me know if you feel like such a change would be accepted.

PR #8047, opened 2020-11-13
[python] implement buffer protocol for repeated primitive fields

tgolsson:
This PR adds support for the buffer protocol (`memoryview`) and a `resize` method to the Python APIs, allowing much faster deserialization when interacting with other modules able to consume and produce raw memory. This gives moderate to massive speed up for deserialization (8x at 200 items, 80x! at 2000), with only moderate gains for serialization (~2.2 ~ 3.2).

The [buffer protocol](https://www.python.org/dev/peps/pep-3118/) is a Python API allowing objects to expose internal buffers, intended to improve interoperability between different C-level modules. These buffers are passed as `Py_buffer` objects, which annotate a {ptr, count} pair with information about the type and shape of the pointed-at data. In this PR I've implemented this based on CPP_TYPE, mapping each type I felt safe to expose to the corresponding Python buffer type. 

Unfortunately, this uses a deprecated API (`MutableRepeatedField(...)`), but it looks like the newer APIs are more restrictive and don't let you retrieve a pointer, or even access data when it's inside the buffer. I think the gains here speak for themselves in terms of usefulness - for us, we'd reduce the GRPC overhead by several milliseconds, so if this deprecated usage is a blocker we should try to forge another way forward.

This is potentially a resolution for #6782.

This is my benchmarking script if you'd like to validate the results.

```python3
import numpy as np
import timeit
from itertools import chain
from benchmark_pb2 import Collection, Data

# benchmark.proto
# syntax = "proto3";
# package benchmark;
# message Data {
#     repeated float values = 1;
# }
# message Collection {
#     repeated Data sets = 2;
# }


ITER_COUNT = 1000


def simple(collection):
    return np.array([list(d.values) for d in collection.sets])


def array_from_iter(collection):
    value_count = len(collection.sets[0].values)
    return np.array([np.fromiter(d.values, dtype=np.float32, count=value_count) for d in collection.sets])


def array_from_slice(collection):
    return np.array([d.values[:] for d in collection.sets])


def from_iter_reshape(collection):
    set_count = len(collection.sets)
    value_count = len(collection.sets[0].values)
    total = set_count * value_count
    return np.fromiter(chain.from_iterable(d.values for d in collection.sets), dtype=np.float32, count=total).reshape(set_count, value_count)


def from_buffer_array(collection):
    value_count = len(collection.sets[0].values)
    return np.array([np.frombuffer(memoryview(d.values), dtype=np.float32, count=value_count) for d in collection.sets])


def eval(func, collection):
    average_time = timeit.timeit(lambda: func(collection), number=ITER_COUNT)
    print(f'{func.__name__}: {average_time / ITER_COUNT * 1000} ms')
    return average_time


def simple_write(values):
    return Collection(sets=[Data(values=value.tolist()) for value in values])


def from_memory_view(values):
    sets = []
    value_count = len(values[0])
    for val in values:
        d = Data()
        d.values.Resize(value_count)  # n.b. this is type-sensitive
        memoryview(d.values)[:] = val.data
        sets.append(d)

    return Collection(sets=sets)


def eval_write(func, collection):
    average_time = timeit.timeit(lambda: func(collection), number=ITER_COUNT)
    print(f'{func.__name__}: {average_time / ITER_COUNT * 1000} ms')
    return average_time


def eval_count(n):
    collection = Collection(sets=[Data(values=list(range(n))) for _ in range(32)])
    values = [np.array(list(range(n)), dtype=np.float32) for _ in range(32)]
    print(f'Count == {n}')
    print('To numpy')
    t0 = eval(simple, collection)
    t1 = eval(array_from_iter, collection)
    t2 = eval(array_from_slice, collection)
    t3 = eval(from_iter_reshape, collection)
    t4 = eval(from_buffer_array, collection)

    maxv = max(t0, t1, t2, t3, t4)
    print(f'Perf gain vs best: {maxv/t4}')

    print('\nFrom numpy')
    t1 = eval_write(simple_write, values)
    t2 = eval_write(from_memory_view, values)
    print(f'Perf gain: {t1/t2}')
    print('\n\n')


eval_count(200)
eval_count(500)
eval_count(1000)
eval_count(2000)
```

### DIFF
--- a/python/google/protobuf/internal/containers.py
+++ b/python/google/protobuf/internal/containers.py
@@ -291,6 +291,20 @@ class RepeatedScalarFieldContainer(BaseContainer):
     self._values.extend(other._values)
     self._message_listener.Modified()
 
+  def Resize(self, count):
+    """Resizes to the specified number of elements in the container. Only valid for primtive types."""
+    if len(self._values) >= count:
+      return
+
+    if count < 0:
+      raise ValueError("resize count was negative %d" % count)
+
+    if not hasattr(self._type_checker, 'DefaultValue'):
+      raise TypeError("resize not supported for this type")
+
+    while len(self._values) < count:
+      self.append(self._type_checker.DefaultValue())
+
   def remove(self, elem):
     """Removes an item from the list. Similar to list.remove()."""
     self._values.remove(elem)
--- a/python/google/protobuf/internal/message_test.py
+++ b/python/google/protobuf/internal/message_test.py
@@ -795,6 +795,77 @@ class MessageTest(unittest.TestCase):
     with self.assertRaises(TypeError):
       hash(m.repeated_nested_message)
 
+  def testRepeatedFieldResize(self, message_module):
+    m = message_module.TestAllTypes()
+    m.repeated_int32.Resize(10)
+    self.assertEqual(len(m.repeated_int32), 10)
+    self.assertEqual(list(m.repeated_int32), [0] * 10)
+
+  def testRepeatedFieldAssertsNegative(self, message_module):
+    m = message_module.TestAllTypes()
+
+    with self.assertRaises(ValueError):
+        m.repeated_int32.Resize(-1)
+
+  def testRepeatedFieldResizeAssertsType(self, message_module):
+    m = message_module.TestAllTypes()
+
+    with self.assertRaises(AttributeError):
+        m.repeated_nested_message.Resize(10)
+
+  def testRepeatedFieldMemoryviewAttrs(self, message_module):
+    if (api_implementation.Type() != 'cpp'):
+        return
+    m = message_module.TestAllTypes()
+    m.repeated_int32.Resize(10)
+    view = memoryview(m.repeated_int32)
+
+    self.assertEqual(len(view), 10)
+    self.assertEqual(len(view.tobytes()), 40)
+    self.assertEqual(view.shape, (10, ))
+    self.assertEqual(view.format, "i")
+    self.assertEqual(view.itemsize, 4)
+    self.assertEqual(view.obj, m.repeated_int32)
+
+  def testRepeatedFieldMemoryviewAppendWrite(self, message_module):
+    if (api_implementation.Type() != 'cpp'):
+        return
+
+    m = message_module.TestAllTypes()
+    m.repeated_int32.append(2)
+    m.repeated_int32.append(4)
+    m.repeated_int32.append(6)
+    m.repeated_int32.append(8)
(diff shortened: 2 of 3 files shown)


tgolsson:
Hello! Do I need to do something more to get this looked at?

tgolsson:
Sorry to be a bother but what's the status here? If you want me to rebase please tell me, I'm not doing work to have it rot again without any communication.

Tabrizian:
This PR is very useful. It would be great if Protobuf can support `memoryviews` too.

haberman:
Hi @tgolsson, I'm sorry for the long delay on this.

Is the underlying motivation to get fast NumPy <-> protobuf conversions?

We recently had an internal design review in which we decided to implement `__array__` instead of the buffer protocol for fast protobuf -> NumPy conversion.

Will that address your needs, or are you also looking for a faster NumPy -> protobuf path?

We don't want to implement the buffer protocol because it exposes more of our internal implementation, and may cause problems if the underlying repeated field is modified while a view to it exists.

tgolsson:
Hello @haberman! 

Since my original code didn't address serialization specifically, I'm not too concerned about it. I'd obviously like if I could also provide a numpy array to protobuffers and have it use a fastpath if the "source" supports buffer-protocol or __array__. However, most of my workloads have ingress payloads at least one order of magnitude larger than egress, often multiple, so for my specific case it won't have huge impact. 

Either way, glad to have something happen here, looking forward to using it! Is there an ETA/roadmap for the work involved?

acozzette:
@tgolsson Sorry for the delay on this. We haven't forgotten about this but we're just trying to figure out the status of the internal proposal involving `__array__`. That proposal came from another team and we're trying to find out what their timeline is.

haberman:
Would you get any benefit from adding resize without the `__array__` changes? We may be able to separate that, to realize some of the win while we sort out our plans around `__array__`.

tgolsson:
Keeping in mind it's soon two years since I opened this PR, my recollection is that Resize was only implemented to support the "numpy to protobuffer" use-case (to ensure enough space exists). Whether it'd make sense on its own with the 2022 version of the code I cannot say, but with the code as it was in 2020 I don't think it'd add any measurable performance gains for arrays.

zhangskz:
Hi @tgolsson, closing this PR and tracking it as an issue in https://github.com/protocolbuffers/protobuf/issues/10930 instead since we're not pursuing this specific implementation path.

I've followed up with the owner of the internal proposal for the __array__ solution, and will update the issue once we have an update on timeline and what else it will take for that to land. Apologies for the delay!

Issue #8051, opened 2020-11-16
Wrong sys/stat.h usage resulted in detecting file as directory on some UNIX (zOS)


mahdi-hm:
Hi,
I tried to compile and run protobuffer on zOS. the compilation went smoothly without any issue however while I was trying to use the binary using: ` protoc --go_out=. *.proto` I ran into below probelm:
```
Could not map to virtual file: addr.proto: Input file is a directory.

```

PR #8052, opened 2020-11-16
Fix detecting file as directory on zOS issue #8051

mahdi-hm:
sys/stat.h provides macros like S_ISDIR to test for a directory. The macro S_IFDIR should not be used in this way since it has different result on zOS(Always detecting file as directory as a consequence).

### DIFF
--- a/src/google/protobuf/compiler/importer.cc
+++ b/src/google/protobuf/compiler/importer.cc
@@ -495,10 +495,17 @@ io::ZeroCopyInputStream* DiskSourceTree::OpenDiskFile(
   do {
     ret = stat(filename.c_str(), &sb);
   } while (ret != 0 && errno == EINTR);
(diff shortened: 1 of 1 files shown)


acozzette:
@mahdi-hosseini-moghaddam Thank you for sending this fix, but I think the one problem is that Windows doesn't seem to support the `S_ISDIR` macro. Maybe you could put in an `ifdef` to keep the old if condition for Windows builds?

mahdi-hosseini-moghaddam:
> 
> 
> @mahdi-hosseini-moghaddam Thank you for sending this fix, but I think the one problem is that Windows doesn't seem to support the `S_ISDIR` macro. Maybe you could put in an `ifdef` to keep the old if condition for Windows builds?

I didn't know that, Thank you for your comment. I will update that.

PR #8053, opened 2020-11-17
Fix some -Wunused-parameter warnings

georgthegreat:
With every release new unused parameters appear in the codebase (the previous one was fixed in #7878).

@acozzette, is there any option to ensure the absence of such parameters via CI / testing?


acozzette:
@georgthegreat Thanks for the fix. We could potentially change this [line](https://github.com/protocolbuffers/protobuf/blob/0321358204bf7676a365b48b76a23e41259a56c7/tests.sh#L60) to `CXXFLAGS="-Werror=unused-parameter" ./configure` to make this an error, but I tried it just now and it won't build successfully without fixing other cases in protoc.

Issue #8061, opened 2020-11-18
Add support for zOS


mahdi-hm:
I faced with numerous problem while I was trying to compile and run protocol buffer on zOS since tools like autoconf, cmake and configure are not fully available on zOS as in most unix. Having zOS support would be a nice feature.

TeBoring:
Could you send us a fix.

mahdi-hm:
> Could you send us a fix.

I created a PR for that https://github.com/protocolbuffers/protobuf/pull/8087
I removed the Makefile and one source file from it as requested btw

elharo:
Seems google is not able to maintain this patch. A fork would be required.

mahdi-hm:
> Seems google is not able to maintain this patch. A fork would be required.

Thanks, A fork will be created for IBM zOS (mainframe) users

PR #8063, opened 2020-11-18
 Add support for zOS #8061

mahdi-hm:
This PR will add zOS support for protocol buffer with required files and Makefile. Simply users on zOS can enter zosbuild directory and make it.


mahdi-hm:
Needs C++ and zos label I believe.

mahdi-hm:
@googlebot I fixed it.

PR #8066, opened 2020-11-18
Add zOS build support to protocolbuffer

mahdi-hm:
This PR provides fixes and mechanism to build protobuf in zOS. Besides source add/edit, a Makefile added for build since cmake, configure and some other tools are not fully functional in zos.

### DIFF
--- a/src/google/protobuf/compiler/importer.cc
+++ b/src/google/protobuf/compiler/importer.cc
@@ -500,7 +500,7 @@ io::ZeroCopyInputStream* DiskSourceTree::OpenDiskFile(
       last_error_message_ = "Input file is a directory.";
       return NULL;
   }
-#elif
+#else
   if (ret == 0 && S_ISDIR(sb.st_mode)) {
     last_error_message_ = "Input file is a directory.";
     return NULL;


zosbuild/Makefile
```diff
@@ -0,0 +1,312 @@
+all: check protoc
```
acozzette, line 1: Sorry but I don't think we can take on another build system. We have too many build systems for C++ already (autotools, CMake, and Bazel). What prevents the autotools and CMake builds from working with zOS?

mahdi-hm: Actually zOS do not have cmake, configure and tools like automake and autotools. But if that's the case I can remove makefile from the PR.

mahdi-hm: Also the compiler on zOS is different and needed different FLAGS. We faced with numerous compile and link problems using configure and cmake. But since make is the most basic option and also available on zos for long time and is stable we turned to use that and we were able to compile the protobuf.

acozzette: Are you saying that CMake will work? Actually the source tar.gz files we publish at https://github.com/protocolbuffers/protobuf/releases also include a `configure` script, so you should be able to run `./configure && make` without even needing autotools.

src/google/protobuf/atomic_store.cc
```diff
@@ -0,0 +1,82 @@
+// --- start __atomic_store
+#if defined (__MVS__)
```
acozzette, line 2: I don't think we can accept this much special code for just z/OS. Does z/OS not support C++11 atomics?

zosbuild/Makefile
mahdi-hm: Actually CMake is not available on zOS at all. Configure also does not work as intended. The only way for mainframe users to compile the project is using pure make. That's way an old fashioned makefile was created.

src/google/protobuf/atomic_store.cc
mahdi-hm: The compiler on zOS machine lacks this feature, we had to add it manually since the build requires that and we've faced link error.

zosbuild/Makefile
mahdi-hm: in order to have configure, first there is a need to run `autogen.sh` which failed on zOS machine since it has dependencies not available or fully functional(for instance `autoreconf`) on zOS as well.

acozzette: There's no need to run autogen.sh or autoreconf if you start with one of our packed tar.gz files, which already contains the configure script.

acozzette:
Sorry but I'm afraid this is just too much special code for us to accept for z/OS. We can't take on another build system and special atomics code.

mahdi-hm:
> Sorry but I'm afraid this is just too much special code for us to accept for z/OS. We can't take on another build system and special atomics code.

should I create a new PR without Makefile and atomic-store or just change this one?

PR #8068, opened 2020-11-19
Add alias to Map#length

miyucy:
e.g.
https://github.com/ruby/ruby/blob/master/array.c#L8053-L8054
https://github.com/ruby/ruby/blob/master/hash.c#L6955-L6956


miyucy:
@googlebot I signed it!

acozzette:
Thanks, @miyucy.

Issue #7770, opened 2020-07-31
Duplicate symbols with Unity (Jumbo) builds


greenkalx:
Hello, 

We have multiple projects in Visual Studio that compile with 'protoc.exe' from .proto files to .h and .cc.

When compiling with Unity (Jumbo) builds,
is there a recommended way to work with the duplicate symbols (One Definition Rule) that collide when compiling the merged .cc files?

The errors that are produced:
```
1>C:\app\ProtoBuf\generated\Pelican.pb.cc(284,65): error C2374: 'schemas': redefinition; multiple initialization
1>C:\app\ProtoBuf\generated\Pelican.Ipc.pb.cc(393): message : see declaration of 'schemas'
1>C:\app\ProtoBuf\generated\Pelican.pb.cc(295,55): error C2374: 'file_default_instances': redefinition; multiple initialization
1>C:\app\ProtoBuf\generated\Pelican.Ipc.pb.cc(409): message : see declaration of 'file_default_instances'
```

References:
https://onqtam.com/programming/2019-12-20-pch-unity-cmake-3-16/
https://onqtam.com/programming/2018-07-07-unity-builds/
https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html
https://devblogs.microsoft.com/cppblog/support-for-unity-jumbo-files-in-visual-studio-2017-15-8-experimental/

Running v3.9.1:
c:\>C:\protobuf-compiler\bin\protoc.exe --version
libprotoc 3.9.1

Regards

acozzette:
So far we have not made any attempt to make protobuf generated code compatible with unity builds. I think you might need to use a CMake option such as `SKIP_UNITY_BUILD_INCLUSION` to exclude protobuf generated code from being compiled this way.

greenkalx:
Thanks for the quick reply.
Can I make this a feature request? I would think a lot of teams would benefit from it.

Meanwhile, I can exclude our protobuf projects from unity builds, yes.

The Visual Studio unity builds allow for exclude at the project level, not sure if cmake excludes only at the file level. Looks like SKIP_UNITY_BUILD_INCLUSION is file level only.

acozzette:
[...]

PR #8070, opened 2020-11-19
Avoid duplicate variable name when unity build

mkkim1129:
Fix https://github.com/protocolbuffers/protobuf/issues/7770


mkkim1129:
@googlebot I signed it!

mkkim1129:
@googlebot I fixed it.

mkkim1129:
@googlebot I fixed it.

acozzette:
Sorry but I don't think we can accept this pull request. Making pb.cc files safe to concatenate together would create a new maintenance burden that we don't want to take on. I actually think including pb.cc files in unity builds would probably make incremental rebuilds slower most of the time, since if you're just editing C++ source then there's no need to rebuild the pb.cc code. If you want to do a unity build I would suggest excluding the pb.cc files from it.

PR #8076, opened 2020-11-20

haberman:

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -254,7 +254,7 @@ if __name__ == '__main__':
     os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'
 
   # Keep this list of dependencies in sync with tox.ini.
-  install_requires = ['six>=1.9', 'setuptools']
+  install_requires = ['six>=1.9']
   if sys.version_info <= (2,7):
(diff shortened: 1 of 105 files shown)


python/setup.py
```diff
@@ -254,7 +254,7 @@ def get_option_from_sys_argv(option_str):
     os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'
 
   # Keep this list of dependencies in sync with tox.ini.
-  install_requires = ['six>=1.9']
+  install_requires = ['six>=1.9', 'setuptools']
```
acozzette, line 257: This setuptools dependency was just recently deleted in #7511 so I think something must be going wrong if it's getting reverted.

acozzette (approved):

src/google/protobuf/arena.h
```diff
@@ -348,7 +348,7 @@ class PROTOBUF_EXPORT PROTOBUF_ALIGNAS(8) Arena final {
     }
   }
 
-  // The following are routines are for monitoring. They will aproximate the
```
Heresh83: CHANGES.txt

haberman: Are you suggesting that something needs to be added to `CHANGES.txt` here?

Issue #8073, opened 2020-11-20
Examples should not be installed without namespacing


xkszltl:
**What version of protobuf and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?**

CentOS 7/Ubuntu 18.04

**What runtime / compiler are you using (e.g., python version or gcc version)**

gcc-9/gcc-8

**What did you do?**

Build by cmake with `-Dprotobuf_BUILD_EXAMPLES=ON -Dprotobuf_INSTALL_EXAMPLES=ON` and install to `/usr/local`.
Full script below:
https://github.com/xkszltl/Roaster/blob/3ed8358199b4a08088525ce994b70a5b05b5908c/pkgs/protobuf.sh#L48-L66

**What did you expect to see**

Examples in `/usr/local/examples/protobuf/...` or `/usr/local/share/protobuf/example`

**What did you see instead?**

`examples` dir was directly installed as `/usr/local/examples`, with general files like `/usr/local/examples/CMakeLists.txt` which is not supposed to be there without encapsulating with a protobuf dir.

acozzette:
Thanks for the bug report. We would be glad to accept a pull request if you have a fix for this.

xkszltl:
@acozzette 
I can send one if wanted, most likely as simple as appending `/protobuf` to this line:
https://github.com/protocolbuffers/protobuf/blob/d7a26435fb36b61d742c053407e3a61fb1012706/cmake/install.cmake#L148

xkszltl:
PR sent: https://github.com/protocolbuffers/protobuf/pull/8080
CI build Mergeable failed asking for label.
Please help granting permission needed for adding that.

PR #8080, opened 2020-11-21
Install examples in `/usr/local/share/protobuf/examples`.

xkszltl:
`/usr/local/example` should not be used without protobuf subdir.
Resolve https://github.com/protocolbuffers/protobuf/issues/8073


acozzette:
Thanks, @xkszltl.

PR #8081, opened 2020-11-21
Add `use_field_number` arg to MessageToJson/MessageToDict

X-Wei:
Add an option to use field numbers rather than field names in `MessageToJson`/`MessageToDict`, similar to [text_format.MessageToString()](https://googleapis.dev/python/protobuf/latest/google/protobuf/text_format.html)

This is sometimes more useful, e.g. dart-protobuf only expects **field numbers** as json keys, not **field names**. (https://github.com/dart-lang/protobuf/pull/83)

Moreover, the filed name could change over time, but the proto filed numbers rarely change.

### DIFF
--- a/python/google/protobuf/json_format.py
+++ b/python/google/protobuf/json_format.py
@@ -101,6 +101,7 @@ def MessageToJson(
     message,
     including_default_value_fields=False,
     preserving_proto_field_name=False,
+    use_field_number=False,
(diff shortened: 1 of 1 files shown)


acozzette:
I'm going to close this since there has been no activity in a while, but feel free to reopen this if you would like to keep working on it.

PR #8087, opened 2020-11-24
Add zOS build support to protocolbuffer

mahdi-hm:
This change is to enable building protocolbuffer on zOS. The change is not introducing any new features and build system . This means it would not break many people's projects and is safe.

### DIFF
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -131,7 +131,9 @@
 #pragma runtime_checks("c", off)
 #endif
 #else
+#if !defined(__MVS__)
 #include <sys/param.h>  // __BYTE_ORDER
+#endif
 #if ((defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)) ||    \
      (defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN)) && \
     !defined(PROTOBUF_DISABLE_LITTLE_ENDIAN_OPT_FOR_TEST)
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -276,12 +276,14 @@
 #define PROTOBUF_COLD
 
 // Copied from ABSL.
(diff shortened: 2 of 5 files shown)


mahdi-hm:
@acozzette Sorry I accidentally  did some changes resulted in the previous PR to be removed. 
As suggested, in this PR, I didn't add new build system and new major implementation, just minor changes for header files and build flags which are not available on zOS.
`param.h` is one of them which does not exist on zOS. As for the compiler flag I added zOS specific change using __MVS__ .
Thanks for the time you put on these changes.

acozzette:
@mahdi-hm Thanks for the simplified pull request, but actually I am reluctant to accept even this version. The problem I see is just that we have no way of testing the build on z/OS and as a result I suspect that the #ifdef logic is going to constantly break and need to be fixed. I think the best solution would be if you could maintain a z/OS-specific fork of protobuf with these changes applied.

mahdi-hm:
> @mahdi-hm Thanks for the simplified pull request, but actually I am reluctant to accept even this version. The problem I see is just that we have no way of testing the build on z/OS and as a result I suspect that the #ifdef logic is going to constantly break and need to be fixed. I think the best solution would be if you could maintain a z/OS-specific fork of protobuf with these changes applied.

Thanks for your comment and all of the time you put on this.

mahdi-hm:
> > @mahdi-hm Thanks for the simplified pull request, but actually I am reluctant to accept even this version. The problem I see is just that we have no way of testing the build on z/OS and as a result I suspect that the #ifdef logic is going to constantly break and need to be fixed. I think the best solution would be if you could maintain a z/OS-specific fork of protobuf with these changes applied.
> 
> Thanks for your comment and all of the time you put on this.

As for testing the build on z/OS, can we move forward with this PR once you were able to test it? How one can provide the environment for testing? is there any test bots that we can integrate with and provide the necessary environment for testing z/OS specific changes? I would appreciate if you could guide me in that regards as well.

elharo:
Seems google is not able to maintain this patch.

PR #8089, opened 2020-11-24
Fixed some minor issues

acozzette:
This pull request fixes some minor issues that came up when I tried to integrate changes from Git into Google's internal codebase:
  - Include stubs/strutil.h for access to Split() helper function
  - Use emplace() instead of insert() so that the code still works if we're working with a string_view instead of a string


src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -2255,7 +2255,7 @@ bool Generator::GenerateAll(const std::vector<const FileDescriptor*>& files,
     if (HasPrefixString(option_pair[0], "aggregate_metadata")) {
```
jorgbrown, line 2255: Could you also change HasPrefixString to StartsWith while you're here?

jorgbrown: Ah, never mind, I see you're not using true Abseil here.

TeBoring (approved):

PR #8091, opened 2020-11-25
Register mark objects so Protobuf is compaction friendly

jhawthorn:
This is a rebased version of #6073

Since that PR was originally sent GC compaction has shipped with Ruby 2.7 and many tools will use it by default (for example `puma` and `sidekiq`). This gem is now unsafe and has potential for memory errors because this has not been fixed.

This commit removes an unused reference and registers globals with the GC so that they will never die.  Ruby is getting a compacting GC, and it means that these references can move.  Registering them with `rb_gc_register_mark_object` will ensure the constants don't move and will not be collected.

### DIFF
--- a/ruby/ext/google/protobuf_c/protobuf.c
+++ b/ruby/ext/google/protobuf_c/protobuf.c
@@ -30,7 +30,6 @@
 
 #include "protobuf.h"
 
-VALUE cError;
 VALUE cParseError;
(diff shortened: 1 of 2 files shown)


jhawthorn:
@googlebot I consent.

TeBoring:
Please sign the CLA.

acozzette:

PR #8092, opened 2020-11-25
Work around -Werror=type-limits under gcc 10.2

jorgbrown:
This is an error when tag is greater than 128 under gcc 10.2:   if (tag < 128) return *ptr == tag;
It's an error even though the comparison occurs in a branch of code we know won't be taken.  See https://godbolt.org/z/1eaP86
This works around the problem by clipping tag to within bounds.

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -428,7 +428,7 @@ class PROTOBUF_EXPORT ParseContext : public EpsCopyInputStream {
 template <uint32 tag>
 bool ExpectTag(const char* ptr) {
   if (tag < 128) {
-    return *ptr == tag;
+    return *ptr == tag % 128;
   } else {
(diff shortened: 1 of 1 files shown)


src/google/protobuf/parse_context.h
```diff
@@ -428,7 +428,7 @@ class PROTOBUF_EXPORT ParseContext : public EpsCopyInputStream {
 template <uint32 tag>
 bool ExpectTag(const char* ptr) {
   if (tag < 128) {
-    return *ptr == tag;
+    return *ptr == tag % 128;
```
dlj-NaN, line 431: This seems a bit over-subtle... how about just a `static_cast<char>(tag)`?

dlj-NaN (changes requested):

src/google/protobuf/parse_context.h
jorgbrown: Sure, whatever.

dlj-NaN (approved):

PR #8098, opened 2020-11-28
stubs/mutex: Fix compilation by initializing variable in WrappedMutex class.

Biswa96:
This solves the following error in gcc:
explicitly defaulted function cannot be declared 'constexpr' because the implicit declaration is not 'constexpr'


acozzette:
Thanks, @Biswa96.

PR #8105, opened 2020-11-30
Port for php8

TeBoring:
1. Fixed tests to be compatible with phpunit 9
2. Added missing arginfo to c extension
3. Add port for `compare_objects`, which was changed to `compare` in php runtime.


haberman:
Very nice change.

I don't see any Kokoro results for PHP 8.0?

TeBoring:
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fphp80%2Fpresubmit?search_pattern=protobuf&include_inactive_projects=false

It doesn't show up, maybe because I opened this PR before I added the CI.

haberman (approved):

php/composer.json
```diff
@@ -6,10 +6,10 @@
   "homepage": "https://developers.google.com/protocol-buffers/",
   "license": "BSD-3-Clause",
   "require": {
-    "php": ">=5.5.0"
+    "php": ">=7.0.0"
```
bshaffer, line 9: This drops support for 5.5 and 5.6 in the native PHP library. Was this intentional? I see no mention of this anywhere in the PR or README.

PR #8106, opened 2020-11-30
Don't include sys/param.h for _BYTE_ORDER

giannitedesco:
It includes a bunch of other files including a lot of macros which can
reduce the namespace available for actual protobuf.

For example, create a protobuf with a member called SIGSEGV. Since
macros cannot be namespaced in C++ this results in code which cannot be
compiled.

Fix this by just directly including endian.h

### DIFF
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -131,7 +131,11 @@
 #pragma runtime_checks("c", off)
 #endif
 #else
-#include <sys/param.h>  // __BYTE_ORDER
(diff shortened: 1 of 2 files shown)


giannitedesco:
@googlebot I signed it!

acozzette:
It looks like this endian.h header doesn't exist on Mac OS unfortunately.

giannitedesco:
Ah thanks. Apparently `machine/endian.h` is the thing to include on apple implementations.

acozzette:
Thanks, @giannitedesco.

PR #8115, opened 2020-12-02
Remove PROTOBUF_MAYBE_CONSTEXPR and make the constructor constexpr ev…

acozzette:
…en in MSVC

This solves a conflict when using Clang for Windows.

This is just a cherry-pick of the internal change cl/341698875.


acozzette:
This should fix an issue that is blocking gRPC from upgrading to 3.14.0.

TeBoring (approved):

PR #8122, opened 2020-12-04
Make MessageToDict convert map keys to strings

mamapanda:
According to the [language guide](https://developers.google.com/protocol-buffers/docs/proto3#json), the JSON map keys should be in string form.


deannagarcia:
Thank you for this contribution, merging now!

PR #8126, opened 2020-12-06
[python] Add macOS big sur compatibility

atomiix:
In macOS big sur, `sysconfig.get_config_var('MACOSX_DEPLOYMENT_TARGET')` is returning `11` as an integer which is incompatible with `pkg_resources.parse_version()`.


atomiix:
@googlebot I signed it!

lumaxis:
Running into this too and the fix looks good to me 👍🏼
@atomiix Looks like this PR needs the `python` label to make the bot happy.

atomiix:
@lumaxis I'd like too but it seems that I have no right to add labels 😕

acozzette:
Thanks, @atomiix

PR #8149, opened 2020-12-16
Optimize writing small strings

JamesNK:
If a string is 42 characters or less then its length will always fit in 1 byte. If there is space available in the buffer, write directly to the buffer, then write the length afterwards based on the number of bytes written. Avoids having to calculate the size.

Slightly slower for 1 byte ASCII strings, faster for all other strings 42 chars or less.

@jtattermusch 

Before
```
|                           Method | BytesToWrite | encodedSize |       Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |------------- |------------ |-----------:|----------:|----------:|------:|------:|------:|----------:|
|         WriteString_WriteContext |        10080 |           1 | 149.235 us | 1.5862 us | 1.4061 us |     - |     - |     - |       2 B |
|         WriteString_WriteContext |        10080 |           4 |  48.668 us | 0.4787 us | 0.4478 us |     - |     - |     - |       1 B |
| WriteNonAsciiString_WriteContext |        10080 |           4 |  99.273 us | 1.4526 us | 1.2877 us |     - |     - |     - |       1 B |
|         WriteString_WriteContext |        10080 |          10 |  25.254 us | 0.2668 us | 0.2365 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |          10 |  61.336 us | 0.3819 us | 0.3573 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |         105 |  11.158 us | 0.1050 us | 0.0931 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |         105 |   9.676 us | 0.1001 us | 0.0937 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |       10080 |   8.997 us | 0.0433 us | 0.0405 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |       10080 |   4.459 us | 0.0446 us | 0.0418 us |     - |     - |     - |         - |
```

After
```
|                           Method | BytesToWrite | encodedSize |       Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |------------- |------------ |-----------:|----------:|----------:|------:|------:|------:|----------:|
|         WriteString_WriteContext |        10080 |           1 | 173.719 us | 2.6596 us | 2.4878 us |     - |     - |     - |       2 B |
|         WriteString_WriteContext |        10080 |           4 |  44.960 us | 0.6183 us | 0.5481 us |     - |     - |     - |       1 B |
| WriteNonAsciiString_WriteContext |        10080 |           4 |  56.988 us | 0.2900 us | 0.2571 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |          10 |  19.008 us | 0.3009 us | 0.2349 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |          10 |  24.667 us | 0.2239 us | 0.2094 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |         105 |  11.563 us | 0.0841 us | 0.0745 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |         105 |   9.856 us | 0.0580 us | 0.0543 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |       10080 |   9.143 us | 0.0960 us | 0.0851 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |       10080 |   4.458 us | 0.0306 us | 0.0256 us |     - |     - |     - |         - |
```

### DIFF
--- a/csharp/src/Google.Protobuf/WritingPrimitives.cs
+++ b/csharp/src/Google.Protobuf/WritingPrimitives.cs
@@ -163,10 +163,26 @@ namespace Google.Protobuf
         /// </summary>
         public static void WriteString(ref Span<byte> buffer, ref WriterInternalState state, string value)
         {
-            // Optimise the case where we have enough space to write
-            // the string directly to the buffer, which should be common.
+            const int MaxBytesPerChar = 3;
+            const int MaxSmallStringLength = 128 / MaxBytesPerChar;
+
+            // The string is small enough that the length will always be a 1 byte varint.
+            // Also there is enough space to write length + bytes to buffer.
+            // Write string directly to the buffer, and then write length.
+            // This saves calling GetByteCount on the string. We get the string length from GetBytes.
+            if (value.Length <= MaxSmallStringLength && buffer.Length - state.position - 1 >= value.Length * MaxBytesPerChar)
+            {
+                // Get the original position, then increment it on state by 1, then write string to buffer.
+                // Method will return byte length, which is then set to the original position.
+                buffer[state.position++] = (byte)WriteStringToBuffer(buffer, ref state, value);
+                return;
+            }
+
             int length = Utf8Encoding.GetByteCount(value);
             WriteLength(ref buffer, ref state, length);
+
+            // Optimise the case where we have enough space to write
+            // the string directly to the buffer, which should be common.
             if (buffer.Length - state.position >= length)
             {
                 if (length == value.Length) // Must be all ASCII...
@@ -179,23 +195,7 @@ namespace Google.Protobuf
                 }
                 else
                 {
-#if NETSTANDARD1_1
-                    // slowpath when Encoding.GetBytes(Char*, Int32, Byte*, Int32) is not available
-                    byte[] bytes = Utf8Encoding.GetBytes(value);
-                    WriteRawBytes(ref buffer, ref state, bytes);
-#else
-                    ReadOnlySpan<char> source = value.AsSpan();
-                    int bytesUsed;
-                    unsafe
-                    {
-                        fixed (char* sourceChars = &MemoryMarshal.GetReference(source))
-                        fixed (byte* destinationBytes = &MemoryMarshal.GetReference(buffer.Slice(state.position)))
-                        {
-                            bytesUsed = Utf8Encoding.GetBytes(sourceChars, source.Length, destinationBytes, buffer.Length);
-                        }
-                    }
-                    state.position += bytesUsed;
-#endif
+                    WriteStringToBuffer(buffer, ref state, value);
                 }
             }
             else
@@ -207,6 +207,33 @@ namespace Google.Protobuf
                 byte[] bytes = Utf8Encoding.GetBytes(value);
                 WriteRawBytes(ref buffer, ref state, bytes);
             }
+        }
+
+        private static int WriteStringToBuffer(Span<byte> buffer, ref WriterInternalState state, string value)
+        {
+#if NETSTANDARD1_1
+            // slowpath when Encoding.GetBytes(Char*, Int32, Byte*, Int32) is not available
+            byte[] bytes = Utf8Encoding.GetBytes(value);
+            WriteRawBytes(ref buffer, ref state, bytes);
+            return bytes.Length;
+#else
+            ReadOnlySpan<char> source = value.AsSpan();
+            int bytesUsed;
+            unsafe
+            {
+                fixed (char* sourceChars = &MemoryMarshal.GetReference(source))
+                fixed (byte* destinationBytes = &MemoryMarshal.GetReference(buffer))
+                {
+                    bytesUsed = Utf8Encoding.GetBytes(
+                        sourceChars,
+                        source.Length,
+                        destinationBytes + state.position,
+                        buffer.Length - state.position);
+                }
+            }
+            state.position += bytesUsed;
+            return bytesUsed;
+#endif
         }
 
         /// <summary>
--- a/csharp/src/Google.Protobuf/WritingPrimitives.cs
+++ b/csharp/src/Google.Protobuf/WritingPrimitives.cs
@@ -172,9 +172,8 @@ namespace Google.Protobuf
             // This saves calling GetByteCount on the string. We get the string length from GetBytes.
             if (value.Length <= MaxSmallStringLength && buffer.Length - state.position - 1 >= value.Length * MaxBytesPerChar)
             {
-                // Get the original position, then increment it on state by 1, then write string to buffer.
-                // Method will return byte length, which is then set to the original position.
-                buffer[state.position++] = (byte)WriteStringToBuffer(buffer, ref state, value);
+                int indexOfLengthDelimiter = state.position++;
+                buffer[indexOfLengthDelimiter] = (byte)WriteStringToBuffer(buffer, ref state, value);
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -163,6 +164,38 @@ public static void WriteBool(ref Span<byte> buffer, ref WriterInternalState stat
         /// </summary>
         public static void WriteString(ref Span<byte> buffer, ref WriterInternalState state, string value)
         {
+#if !NETSTANDARD1_1
+            const int MaxBytesPerChar = 3;
```
thomasvl, line 168: drive by - why is the max 3?

JamesNK: Most Chars are represented by one or two bytes. The maximum is 3. That's just how UTF8 works.

thomasvl: That's what I thought you were going for, but then wouldn't the answer be 4 bytes not 3? https://tools.ietf.org/html/rfc3629 -

```
Char. number range  |        UTF-8 octet sequence
      (hexadecimal)    |              (binary)
   --------------------+---------------------------------------------
   0000 0000-0000 007F | 0xxxxxxx
   0000 0080-0000 07FF | 110xxxxx 10xxxxxx
   0000 0800-0000 FFFF | 1110xxxx 10xxxxxx 10xxxxxx
   0001 0000-0010 FFFF | 11110xxx 10xxxxxx 10xxxxxx 10xxxxxx
```

Depending on what the initial string encoding/length is (UTF16?), you might also have to look out for the surrogate pairs that confuse the issue.

JamesNK: https://source.dot.net/#System.Private.CoreLib/UTF8Encoding.cs,49

thomasvl: Thanks!

jtattermusch (commented):
I think the idea for the optimization is neat and the savings for some string lengths seem significant.
On the other hand, I don't like that the complexity of the methods has increased (several cases to handle + some ifdefs) and that we now have 2 copies of basically the same code under the unsafe block - isn't there a way to refactor so that we end have just one copy of that snippet? (reorganize the method, introduce a static method, ...).

Alternative idea (not sure if its good or not):  the method could be modified so that you simply skip `Utf8Encoding.GetByteCount(value);` and `WriteLength` if you know the length is going to fit into a single byte - this might result in a smaller method without code duplication and might be both faster and easier to read.

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -189,9 +222,13 @@ public static void WriteString(ref Span<byte> buffer, ref WriterInternalState st
                     unsafe
                     {
                         fixed (char* sourceChars = &MemoryMarshal.GetReference(source))
-                        fixed (byte* destinationBytes = &MemoryMarshal.GetReference(buffer.Slice(state.position)))
+                        fixed (byte* destinationBytes = &MemoryMarshal.GetReference(buffer))
```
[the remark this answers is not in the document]
JamesNK: It has a tiny effect on performance. There is no need to do a slice because we can jump to the right place with GetBytes using `destinationBytes + state.position`.

JamesNK:
@jtattermusch Updated. Code is a lot smaller. There is some cleverness going on with position but I put a comment on it to explain what is going on.

JamesNK:
After:
```
|                           Method | BytesToWrite | encodedSize |       Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |------------- |------------ |-----------:|----------:|----------:|------:|------:|------:|----------:|
|         WriteString_WriteContext |        10080 |           1 | 197.933 us | 3.4468 us | 3.2242 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |           4 |  51.337 us | 0.2081 us | 0.1845 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |           4 |  63.139 us | 0.3039 us | 0.2843 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |          10 |  21.122 us | 0.2113 us | 0.1977 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |          10 |  27.255 us | 0.5020 us | 0.4450 us |     - |     - |     - |         - |
```


Before:
```
|                           Method | BytesToWrite | encodedSize |       Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |------------- |------------ |-----------:|----------:|----------:|------:|------:|------:|----------:|
|         WriteString_WriteContext |        10080 |           1 | 147.735 us | 0.7897 us | 0.7001 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |           4 |  48.229 us | 0.7844 us | 0.6953 us |     - |     - |     - |       1 B |
| WriteNonAsciiString_WriteContext |        10080 |           4 | 117.600 us | 0.6227 us | 0.5520 us |     - |     - |     - |         - |
|         WriteString_WriteContext |        10080 |          10 |  25.140 us | 0.1824 us | 0.1706 us |     - |     - |     - |         - |
| WriteNonAsciiString_WriteContext |        10080 |          10 |  42.219 us | 0.6843 us | 0.6401 us |     - |     - |     - |         - |
```

csharp/src/Google.Protobuf/WritingPrimitives.cs
```diff
@@ -163,10 +163,26 @@ public static void WriteBool(ref Span<byte> buffer, ref WriterInternalState stat
...
+
+            // The string is small enough that the length will always be a 1 byte varint.
+            // Also there is enough space to write length + bytes to buffer.
+            // Write string directly to the buffer, and then write length.
+            // This saves calling GetByteCount on the string. We get the string length from GetBytes.
+            if (value.Length <= MaxSmallStringLength && buffer.Length - state.position - 1 >= value.Length * MaxBytesPerChar)
+            {
+                // Get the original position, then increment it on state by 1, then write string to buffer.
+                // Method will return byte length, which is then set to the original position.
+                buffer[state.position++] = (byte)WriteStringToBuffer(buffer, ref state, value);
```
jtattermusch, line 177: I think  `int indexOfLengthDelimiter = state.position++;  buffer[indexOfLengthDelimiter] = WriteStringToBuffer(...)` would be easier to grasp and wouldn't require a comment?

jtattermusch (commented):
Looking good with one minor comment. 
Is the new version (with the new WriteStringToBuffer method) faster, slower or the same as the previous version of this PR?

JamesNK:
> Is the new version (with the new WriteStringToBuffer method) faster, slower or the same as the previous version of this PR?

The same - https://github.com/protocolbuffers/protobuf/pull/8149#issuecomment-767901299

jtattermusch (approved):
LGTM.

jtattermusch:
the windows C# build error seems unrelated: https://fusion2.corp.google.com/invocations/fa716574-5e3f-4a95-a360-2ae376505e58 (look like an infrastructure problem with the windows workers). Other test failures are also unrelated.

PR #8154, opened 2020-12-18
remove CMAKE_THREAD_LIBS_INIT from pkgconfig CFLAGS

neheb:
This is a linker flag and does not belong in CFLAGS.

Fixes an issue with ola and protobuf.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


acozzette:
Thanks, @neheb.

Issue #8168, opened 2020-12-28
Allow for Duration deconstruction into constituent member components


mwpowellhtx:
For use with _CSharp_, allow for [Duration](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/duration.proto) deconstruction into constituent member components.

Stemming from a [`Google.OrTools 7.5.7466`](https://nuget.org/packages/Google.OrTools/7.5.7466) dependency, which has dependency on [`Google.Protobuf 3.11.2`](https://nuget.org/packages/Google.Protobuf/3.11.2).

What do I mean by that? For instance:

```C#
namespace Google.Protobuf.WellKnownTypes
{
    public partial class Duration
    {
        // ...

        public void Deconstruct(long seconds, int? nanos) => (seconds, nanos) = (
            this.Seconds
            , this.Nanos == default ? (int?)null : this.Nanos
        );

        // ...
    }
}
```

Today I am doing this via extension method, which is adequate, however, I think [a better factoring](https://github.com/protocolbuffers/protobuf/blob/master/csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs#L41) would be in the source assembly:

```C#
namespace Google.Protobuf.WellKnownTypes
{
    public static void Deconstruct(this Duration value, out long seconds, out int? nanos) =>
        (seconds, nanos) = (value.Seconds, value.Nanos == default ? (int?)null : value.Nanos);
}
```

`nanos` being `int?` due to the fact that, as far as I can tell, it can be an optional component. When it is `0` or `default`, I think it can be safely ignored? Anyway... as appropriate to do so.

jskeet:
It feels quite odd to me to have `nanos` as `int?` instead of `int` - if I *were* to add a `Deconstruct` method, it would just be with `long` and `int` parameters. However, this still feels like quite a niche requirement - I believe it's better to stick with it being an extension method in libraries that particularly need it, until we see a more widespread requirement for this.

mwpowellhtx:
@jskeet I didn't design the original class; was simply looking for a better way to incorporate it into our own bases. However, we can continue with our internal extensions. Fair enough, but that being said, won't know the adoption until it is "there".

mwpowellhtx:
[...]

PR #8173, opened 2020-12-30
C#: Added Duration deconstruction capability.

mwpowellhtx:
Pursuant to the [issue](https://github.com/protocolbuffers/protobuf/issues/8168).

Thank you.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -100,3 +100,5 @@ Patch contributors:
   Andrew Paprocki <andrew@ishiboo.com>
     * Fixed minor IBM xlC compiler build issues
     * Added atomicops for AIX (POWER)
+  Michael W. Powell <mwpowelllde@gmail.com>
+    * Added Duration deconstruction capability
--- a/csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs
+++ b/csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs
@@ -183,6 +183,25 @@ namespace Google.Protobuf.WellKnownTypes
         }
 
         /// <summary>
+        /// Allows for tuple-like deconstruction of the constituent <see cref="Seconds"/> and,
+        /// optionally, when there is a quantitative value, <see cref="Nanos"/> components.
+        /// </summary>
+        /// <param name="seconds">The <see cref="Seconds"/> component.</param>
+        /// <param name="nanoseconds">Optionally, the <see cref="Nanos"/> component. Nullable
+        /// because the component may or may not be specified by subscribers.</param>
+        /// <remarks>Allows for tuple-like deconstruction of the <see cref="Seconds"/> and,
+        /// optionally, <see cref="Nanos"/>, when there is a quantitative value other than
+        /// <em>zero</em>.</remarks>
+        /// <see cref="Seconds"/>
+        /// <see cref="Nanos"/>
+        /// <see cref="!:https://github.com/protocolbuffers/protobuf/issues/8169"/>
+        public void Deconstruct(out long seconds, out int? nanoseconds)
+        {
+            seconds = this.Seconds;
+            nanoseconds = this.Nanos == 0 ? (int?)null : this.Nanos;
+        }
+
+        /// <summary>
         /// Converts a duration specified in seconds/nanoseconds to a string.
         /// </summary>
         /// <remarks>


mwpowellhtx:
@Mizux Any ideas?

> Please include release notes: yes AND Please include at least a language label (e.g., c++, java, python). Or apply one of the following labels: bazel, cmake, cleanup, conformance tests, integration, protoc.) OR Please include release notes: no

Release notes not required?

AFAIK, I included the `C#` language label, however. That is case sensitive, perhaps?

mwpowellhtx:
@Mizux AFAIK, could be falling over on this:

> `{"must_include":{"regex":"^(c#|c\\+\\+|cleanup|conformance tests|integration|java|javascript|go|objective-c|php|python|ruby|bazel|cmake|protoc)","message":"Please include at least a language label (e.g., c++, java, python). Or apply one of the following labels: bazel, cmake, cleanup, conformance tests, integration, protoc."}}]}`

I am not 100% positive, `c#` shouldn't be `c\#`, and/or allow the full name, i.e. `csharp`.

Why else would it be falling over?

mwpowellhtx:
@Mizux Would like to put this to bed and move on. Do you happen to know of a `protobuf` POC I can mention here?

mwpowellhtx:
Anyone at all? Why is `Mergeable` breaking? The current theory is that a commit comment needs to be addressed. Assuming that is the case, did so, i.e. `C#: Added Duration deconstruction capability.`. Is still falling over there. Is it a `regex` issue? `c#` needs to be `c\\#`, perhaps, for instance? Or even allow for `csharp`? Or `dotnet` for that matter.

jskeet:
@anandolee please could you have a look at the CI check failure?

jskeet:
Drive-by review comment (not a full review): I would definitely *not* use `int?` for the `nanoseconds` parameter. If the value is 0, that doesn't indicate that it wasn't specified or should be ignored in some way. A value of "5 seconds and 0 nanoseconds" is precisely that, it's not "5 seconds and some indeterminate number of nanoseconds".

mwpowellhtx:
@jskeet Appreciate the review; the intent there was "unspecified" not "indeterminate". Which is precisely what a `Nullable<T>` `null` is. In keeping with the minimum requirement when instantiating a `Duration`, i.e.

```C#
new Duration { Seconds = seconds }
```

Or,

```C#
new Duration { Seconds = seconds, Nanos = nanoseconds }
```

Or for that matter,

```C#
new Duration { Nanos = nanoseconds }
```

I suppose  the latter would even be possible.

jskeet:
But the point is that you *don't* know it's unspecified. You can't tell the difference between `new Duration { Seconds = seconds }` and `new Duration { Seconds = seconds, Nanos = 0 }` so why pretend that there *is* a meaningful difference? All that happens is 0 becomes an impossible value to return from the `Deconstruct` method, even if it was *explicitly* specified. I don't think that's a good idea - *and* it's inconsistent with just fetching the `Nanos` property directly.

Perhaps you should indicate what benefit you believe is derived from this inconsistency. (Note that if the designers of the Duration message had wished there to be a meaningful distinction between 0 and "not specified", we'd have used `Int32Value` instead of just `int32`, at which point the property would have been of type `int?`, and then it *would* make sense to deconstruct it to `int?` as well.)

mwpowellhtx:
Okay so the assumption is that "unspecified" is rather "default" `0`. Fair enough. I am not attached to the `int?` notion by any means.

Establishing that as the assumption, then, yes, of course, it would make better sense to _decon_ to the actual type, i.e.

```C#
void Deconstruct(out long seconds, out int nanoseconds) { ... }
```

mwpowellhtx:
Cleaned up branch forthcoming. In the meantime, are we talking about also adding a note in the `CHANGES.txt` file?

Mizux:
looking at https://github.com/protocolbuffers/protobuf/blob/master/.github/mergeable.yml I would try to add "release notes: yes" in the body of the commit message, my guess the `CHANGE.txt` is autogenerated using commit message "tags" etc...

jskeet:
@mwpowellhtx: Is this PR still required? It's now over two years since the last comment, and the code still has `int?` instead of `int` (and no tests).

I'm happy to help review and get it in if you really think it's needed, but personally I'd recommend just living without it if it hasn't been *too* painful for the last two years.

PR #8177, opened 2021-01-05
default

xheke:
protobuf  **libdir** and **includedir** should support prefix, just like setting in grpc pc file


xheke:
@googlebot I signed it!

xheke (commented):
libdir and includedir should depend on prefix

xheke (commented):
libdir and includedir should depend on prefix

PR #8182, opened 2021-01-05
Provide a type alias V on EnumTypeWrapper

### LINKED ISSUE #8175: Better mypy typing for enum values in python
Hi - I'm the maintainer of https://github.com/dropbox/mypy-protobuf - a protoc plugin for autogenerating mypy stubs compatible with the built in python generator.

Python

Generated enum values are pure ints (per the spec). We prefer type them with a `NewType` wrapper type around int - specific to the enum. See the typeshed stub
https://github.com/python/typeshed/blob/master/third_party/2and3/google/protobuf/internal/enum_type_wrapper.pyi#L9
```python
class _EnumTypeWrapper(Generic[_V]):
    DESCRIPTOR: EnumDescriptor
    def __init__(self, enum_type: EnumDescriptor) -> None: ...
    def Name(self, number: _V) -> str: ...
    def Value(self, name: str) -> _V: ...
    def keys(self) -> List[str]: ...
    def values(self) -> List[_V]: ...
    def items(self) -> List[Tuple[str, _V]]: ...
```

I've been cooking up a strategy inside mypy-protobuf where a proto
```proto
enum MyEnum {
   FOO = 0;
   BAR = 1;
}
```
Autogenerates pyi stubs
```python
class _MyEnum(google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper[OuterEnum.V], builtins.type):
    DESCRIPTOR: google.protobuf.descriptor.EnumDescriptor = ...
    FOO = OuterEnum.V(0)
    BAR = OuterEnum.V(1)

class MyEnum(metaclass=_MyEnum):
    V = typing.NewType('V', builtins.int)
```

This allows callsites to type their code as
```python
x: MyEnum.V = MyEnum.Value("FOO")
```

Unfortunately - because `.V` does not exist at runtime (only in the pyi stubs) - this does not work.
Fortunately, you can work around this with:
```python
x: "MyEnum.V" = MyEnum.Value("FOO")
```

I'd like to propose adding `V = int` runtime type alias to the EnumTypeWrapper (in enum_type_wrapper.py) - to avoid the need for this workaround.
```python
class EnumTypeWrapper(object):
  """A utility for finding the names of enum values."""
  # This is a type alias, which mypy typing stubs can type as
  # a genericized parameter constrained to an int, allowing subclasses
  # to be typed with more constraint
  # Eg.
  # def Name(self, number: MyGeneratedEnum.V) -> str
  V = int
  ... [rest of impl omitted]
  ```

This would mean that enum variants named "V" would need to jump through a hoop using getattr to be accessed, but this is already the case with "Name", "Value", "keys", etc - so it seems feasible as an extension. I'm open to other naming ideas for this.

Could do something like this for more clarity - though it does add a dependency on typing module for python <= 3.5
```python
from typing import NewType
V = NewType('V', int)
```

I would also be open to discussing other strategies we could employ across mypy-protobuf, protoc, and typeshed in order to get better mypy typing for enum values (better than just typing as an int) - other than this `V` inner type alias.

I've also tried mypy typing V at a sibling scope to `MyEnum` (eg `MyEnumValue = NewType('MyEnumValue', int)`, but this is no improvement - as you still need to play tricks to import it from real code.

Thanks friends! Happy to open discussion.

nipunn1313:
This should provide a runtime alias, which can be
used in mypy stubs to provide better typing for enum values

This is one potential solution to #8175 - see discussion there.
Fixes #8175


nipunn1313:
@googlebot I signed it!

2m:
This would allow us to cleanup some code as well. :)

StefPtrs:
@nipunn1313 can you please add a label? It seems to be the only missing thing before this can be merged.

perezd:
> @nipunn1313 can you please add a label? It seems to be the only missing thing before this can be merged.

This is quite old, so I'm gonna run the tests to make sure it's still viable.

nipunn1313:
Hi! I'm not an official contributor on the project, so it was not clear to me how to add a label. I can rebase it to make sure things still work! Thanks for taking a look.

perezd:
> Hi! I'm not an official contributor on the project, so it was not clear to me how to add a label. I can rebase it to make sure things still work! Thanks for taking a look.

No prob! I am :)

Looks like all the builds failed, if you wanna fix up this PR I'll have someone take a look at it.

nipunn1313:
I believe it needs a label "release notes: yes" - so that it's automatically added to release notes. Can you run the tests once again (I rebased onto master and updated the PR).

I don't have the ability to approve running workflows:

```
First-time contributors need a maintainer to approve running workflows. Learn more.
```

nipunn1313:
hi - it appears that the Kokoro tests are failing on master. I spun up a dev environment on linux and confirmed that
```
bazel test //:build_files_updated_unittest
```
is failing

I submitted a separate patch to address that in #8789
As far as I can tell, I don't think the failures we're seeing here are related to my patch.

nipunn1313:
as far as I can tell from poking around there are at least a handful of other unrelated build issues that exist on master.

python/google/protobuf/internal/enum_type_wrapper.py
```diff
@@ -45,6 +45,13 @@ class EnumTypeWrapper(object):
 
   DESCRIPTOR = None
 
+  # This is a type alias, which mypy typing stubs can type as
+  # a genericized parameter constrained to an int, allowing subclasses
+  # to be typed with more constraint
+  # Eg.
+  # def Name(self, number: MyGeneratedEnum.V) -> str
+  V = int
```
amauryfa, line 53: I'd prefer a more explicit name like `ValueType`

amauryfa (changes requested):

python/google/protobuf/internal/enum_type_wrapper.py
nipunn1313: Hi - since this is for compatibility with https://github.com/dropbox/mypy-protobuf - this is a bit of a large non-backward compatible suggestion unfortunately.

It has the ergonomic downside of having very lengthy type names that don't end up being much more useful (IMO). In either case, you get an error from mypy of the form:
```
incompatible type "MyGeneratedEnum"; expected "ValueType"
incompatible type "MyGeneratedEnum"; expected "V"
```
both of which require some amount of looking inside the generated mypy-protobuf `.pyi` to understand. I believe there is a common confusion amongst protobuf users that `MyGeneratedEnum` is also an `int`, when protobuf generates it as an `EnumTypeWrapper`.

Here's an example `.proto` enum (OuterEnum) https://github.com/dropbox/mypy-protobuf/blob/master/proto/testproto/test.proto#L11
And the corresponding generated `.pyi` in mypy-protobuf `OuterEnum.V` https://github.com/dropbox/mypy-protobuf/blob/master/test/generated/testproto/test_pb2.pyi.expected#L25

nipunn1313: mypy-protobuf has this documented here https://github.com/dropbox/mypy-protobuf#types-enum-int-values-more-strongly.

If we do want to change the `V` to something else, it could be possible via a compatibility version supporting both `V` and `ValueType` types and eventually dropping support for `V`. I know Dropbox alone has thousands of instances of `V` already, as do other companies/codebases using mypy+protobuf. I'm hesitant to do such a change without some strong feedback + consensus - to avoid churning on name changes here.

amauryfa: This is an unfortunate situation... and the approach looks a bit backwards to me: mypy should not require changes to Python or common libraries.

But if I got this correctly, all current usages need quotes, as in: `def f(x: 'MyEnum.V')` and this `V` is provided by mypy-protobuf generated code.
Since removing the quotes is a change anyway, the recommended end state could be `def f(x: MyEnum.ValueType)` (no quotes, but longer name), and the `V` would be supported by mypy as an alias (shorter, but with quotes)

nipunn1313: > This is an unfortunate situation... and the approach looks a bit backwards to me: mypy should not require changes to Python or common libraries.

Agreed.

The reason this comes up is because `mypy-protobuf` is attempting to get *stronger* typing than what protobuf itself generates at runtime. Protobuf generates all enum values as `int`. The stronger typing has proven to catch bugs and is one of the value adds of mypy-protobuf, but in order to namespace the `NewType("ValueType", int)` into `MyEnum`, it requires this change.

Starting with python 3.10 there is a new feature for postponed evaluation of annotations
https://www.python.org/dev/peps/pep-0563/
(or starting in python 3.7 with `from __future__ import annotations`)
However, there are still cases where this is insufficient to avoid evaluation - notably
```
from typing import cast
cast(MyEnum.ValueType, x)
```

> Since removing the quotes is a change anyway, the recommended end state could be def f(x: MyEnum.ValueType) (no quotes, but longer name), and the V would be supported by mypy as an alias (shorter, but with quotes)

I'm happy with your suggestion for a backward-compatible way to move toward a more verbose name. I can work with that in `mypy-protobuf`. @amauryfa - are you a review authority here? I can make the change. Just want to make sure I avoid discussion churn on these sorts of points.

Side note: There is an alternate python generated code option that sidestepped this problem entirely by generating enum values as enums https://github.com/danielgtaylor/python-betterproto. I've never tried it.

nipunn1313: Updated the diff with this idea of `s/V/ValueType`

StefPtrs: @perezd can you approve running to workflow again?

nipunn1313:
Cool. Just wanted to double check before this releases - are you happy with the name `MyEnum.ValueType` - as it would be challenging to change this name in the future. (eg mypy-protobuf currently uses `MyEnum.V` and I'll have to make that change). I'm happy with this naming scheme for the type.

If people have preexisting proto enums with a field named `ValueType`, their existing python calling code might need to be modified - but I think that's a reasonable change to ask customers to make. Technically not backward compatible, but it's a pretty rare scenario.

Once the next major release of protobuf comes out with this, I'll modify mypy-protobuf to work with this!
Appreciate it.

Issue #7922, opened 2020-09-28
Add ruby 3.0.0 support


sriedel:
**What language does this apply to?**
ruby

**Describe the problem you are trying to solve.**
Ruby 3.0.0-preview1 was released over the weekend. We'd like to check our ruby services before the official ruby 3.0 release (probably on christmas, as this is the usual ruby release date for new versions). However the google-protobuf gem explicitly prohibits bumping the ruby version to beyond 2.7 (the current stable version):

    Gem::RuntimeRequirementNotMetError: google-protobuf requires Ruby version >= 2.3, < 2.8.dev. The current ruby version is 3.0.0.preview1.
     An error occurred while installing google-protobuf (3.13.0), and Bundler cannot continue.    
     Make sure that `gem install google-protobuf -v '3.13.0' --source 'https://gem.givve.io/'` succeeds before bundling.

**Describe the solution you'd like**
The google-protobuf gem is installable under ruby 3.0.0. I'd be happy with a solution that is locked to the preview without any guarantee of functionality in production environments.

acozzette:
Do you know where this requirement is coming from? I see that our gemspec requires `>= 2.3` but I don't see the `< 2.8.dev` anywhere (is that coming from one of our dependencies?)

sriedel:
After hours of being stumped and studying the code for both bundler and gem (since if it isn't provided by you, it must be injected somewhere, right?), it dawned on me that the google-protobuf gem is being supplied in different versions:
- the plain version which we've been looking at, which apparently doesn't have the restriction
- a precompiled version, which has the restriction in place, see https://rubygems.org/gems/google-protobuf/versions/3.13.0-x86_64-linux 

So this seems to be injected by the build process or build system. Which makes sense in a why, since the ABI has changed.
I'll have to figure out how to tell bundler that I don't want the precompiled version. Adding a ruby-3.0.0 version to your build system would be great too, though :)

acozzette:
@sriedel Thanks for tracking that down, that is a little surprising but I guess it makes sense. I think in principle we should be able to make our next release add a ruby-3.0 compatible version. @haberman Does that sound doable?

bilby91:
Is there any ETA for supporting ruby 3.0 ?

KapilSachdev:
xref: #8134

haberman:
This should be fixed as of #8184, released in 3.15.0-rc1. Please let us know if you run into problems.

PR #8184, opened 2021-01-05
Ported Ruby extension to upb_msg

haberman:
This PR implements a significant refactoring of the Ruby C extension. The internal representation of messages is changing from a custom Ruby-specific raw memory layout to using `upb_msg`, a representation that is defined in the upb library. Upb defines a [reflection interface](https://github.com/protocolbuffers/upb/blob/master/upb/reflection.h) that we can use to access message data from C.

This refactoring should lead to much better performance, particularly when parsing large messages. The old code would eagerly create a Ruby object for every single sub-message parsed from the wire, which meant tons of objects would participate in garbage collection. The new extension only creates C objects at parse time, and Ruby wrapper objects are created on demand, only when accessed.

This fixes a number of conformance errors in the old extension.

There are a few slight behavior changes:

1. The old code threw `Google::Protobuf::TypeError` in some cases but plain `TypeError` in others. The new code standardizes on using `Google::Protobuf::TypeError` everywhere.
2. The old JSON and `#inspect` serializers were emitting empty optional fields, whereas they should only be emitted when they are set.
3. We no longer prohibit maps for proto2 protos.
4. If you try to compile a proto2 proto containing extensions, we emit a warning an ignore it instead of rejecting the proto completely.


qnighy:
@googlebot I consent.

KapilSachdev:
Is it ready to be shipped?

xref: #7922

gerben-s (approved):

haberman:
@KapilSachdev the change is looking great in unit tests. I'm hoping to get some feedback about whether it is working for users before I merge.

A team at Google is planning to try it out sometime this week, but any other feedback from open-source users would be very helpful too.

ryanseys:
Oooh, I'm very interested in this. When do you anticipate this will land?

brettfishman:
+1

haberman:
This will be in the next release, which should land in the next few weeks.

hatstand:
I suspect this is the cause of https://github.com/protocolbuffers/protobuf/issues/8311 but only due to the size of the changeset.

Issue #8165, opened 2020-12-24
PHP "float" serialization loses precision


bshaffer:
In both the protobuf native PHP library and PHP C-extension, I am seeing a loss of precision with floats:

```
message TestMessage {
  float foo = 1;
}
```

```php
$initialMessage = new TestMessage();
$initialMessage->setFoo(-8.6107481E7);
echo "Initial message: ";
var_dump($initialMessage->getFoo());

echo "Serialized message: ";
$serializedMessage = new TestMessage();
$serializedMessage->mergeFromString($initialMessage->serializeToString());
var_dump($serializedMessage->getFoo());
```

When the C-Extension is enabled, the loss of precision happens immediately. When the C-extension is not enabled, the loss of precision happens when the message is serialized and deserialized:

```
// Without protobuf C-Extension
Initial message: float(-86107481)
Serialized message: float(-86107480)

// With protobuf C-Extension
Initial message: float(-86107480)
Serialized message: float(-86107480)
```

acozzette:
I haven't done the math carefully but this sounds like it could be just caused by the fact that the protobuf `float` type is only 32 bits.  As I understand, PHP always uses double-precision floats and that would explain why the plain PHP implementation doesn't lose precision by itself but only after a serialize and parse. The C extension presumably distinguishes between `float` and `double` and therefore converts immediately to the 32-bit `float`, resulting in the immediate precision loss.

haberman:
Yes, Adam is correct. The bug here is with pure-PHP, which should truncate this to float precision in the setter.

bshaffer:
@haberman So are we saying this is an issue in the Pure-PHP protobuf library, and the way to fix it is to update float setters to truncate to single precision?

bshaffer:
I added a fix in https://github.com/protocolbuffers/protobuf/pull/8187 that passes for my test suites

haberman:
This was fixed in #8187.

PR #8134, opened 2020-12-11
Test and build for Ruby 3.0

qnighy:
Ruby 3.0 has been released. This PR tests protobuf with Ruby 3.0.0 and builds extensions for Ruby 3.0.x.

Waiting for the release of rake-compiler-dock 1.1.0 https://github.com/rake-compiler/rake-compiler-dock/pull/40

Issue #7854, opened 2020-08-28
ES3 protobuf - getRepeatedWrapperField & wrapRepeatedField_


prk68:
I am trying to run protbuf serialization for JScript app that runs in a window scripting host runtime (WSH) on window. The JScript seems to based on a very old version of Ecmascript.

I webpacked my app code and ran into following error

**"length is null or not an object"**

for this line

`for (var d = jspb.Message.getRepeatedField(a, c), e = [], f = 0; f < d.length; f++)`

This seeems to be happening inside wrapRepeatedField_ function from a getRepeatedWrapperField call.

The same webpacked application code runs perfectly fine on chrome browser. So I am suspecting an issue either with missing polyfills because of which the message was not initialized properly.

I would appreciate any hints on how I could make this work on this old JScript runtime.

prk68:
Further logging tells me that this happens because jspb.Message.getField returns undefined. Both the following conditions do not match
`
if (b < a.pivot_)`

`if (a.extensionObject_)`

jelleklaver:
Hi @prk68,

I seem to have a similar issue, but it occurs to me running in both Chrome and Firefox. I've been using protobuf with grpc-web for a while now, and it suddenly occured for a particular definition. Both errors point to the same line as you mentioned. 

```js
// Chrome console
TypeError: Cannot read property 'length' of undefined
    at Function.jspb.Message.wrapRepeatedField_ (google-protobuf.js:432)
    at Function.jspb.Message.getRepeatedWrapperField (google-protobuf.js:432)
    at proto.domain.Question.getAnswersList (domain_pb.js:4651)
    ...
```

```js
// Firefox console
TypeError: d is undefined
    wrapRepeatedField_ google-protobuf.js:432
    getRepeatedWrapperField google-protobuf.js:432
    getAnswersList domain_pb.js:4651
    ...
```

I tried to regenerate the javascript files using 3.12.4 and 3.13.0, but both give the same results. The strange thing is, that it seems that sometimes setting the repeated field as an empty array (`.setAnswersList([])`) helps, and sometimes it doesn't. 

I'm pretty clueless in how to debug this further. Did you find a solution?

Elojah:
Any news on this bug ? I have exactly the same as @jelleklaver, i can reproduce to give more informations if necessary.
Proto definition:
```proto
message FooResp {
    repeated foo.Bar MyField = 1;
}
```
(with `foo.Bar` containing only native strings)

[...]

PR #8187, opened 2021-01-06
fix: convert native PHP floats to single precision

bshaffer:
See https://github.com/protocolbuffers/protobuf/issues/8165

Converts PHP double-precision floats to single-precision for consistency with the protobuf C-extension

### DIFF
--- a/php/src/Google/Protobuf/Internal/GPBUtil.php
+++ b/php/src/Google/Protobuf/Internal/GPBUtil.php
@@ -168,7 +168,7 @@ class GPBUtil
     public static function checkFloat(&$var)
     {
         if (is_float($var) || is_numeric($var)) {
-            $var = floatval($var);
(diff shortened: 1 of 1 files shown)


haberman:
This PR looks reasonable to me on first glance, but appears to cause some test failures. @bshaffer want to investigate?

bshaffer:
@haberman I unfortunately can't see the details of the failing tests - getting 404s on both of them when I click "Details". Running the tests locally now to investigate...

bshaffer:
The failing tests are for Ruby, which was not effected by this change, so I think they're just flakey.

haberman:
The Ruby 3.0 tests are brand new, we can disregard them until https://github.com/protocolbuffers/protobuf/pull/8134 is merged.

Eubanks688:
#8343

Eubanks688:
#7854

Issue #8191, opened 2021-01-07
[Ruby] JRuby - problems with nested messages


tcruys:
**What version of protobuf and what language are you using?**
Version: 3.14.0 (and the commits upto 6-Jan-2021)
Language: (J)Ruby

**What operating system (Linux, Windows, ...) and version?**
Linux Ubuntu 16.x and 18.x

**What runtime / compiler are you using (e.g., python version or gcc version)**
Ruby 2.6.5 and JRuby 9.2.9.0.

**What did you do?**
Consider the following Ruby script `function_call_pb.rb`
```
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: function_call.proto

require 'google/protobuf'

Google::Protobuf::DescriptorPool.generated_pool.build do
  add_file("function_call.proto", :syntax => :proto3) do
    add_message "Function" do
      optional :name, :string, 1
      repeated :parameters, :message, 2, "Function.Parameter"
      optional :return_type, :string, 3
    end
    add_message "Function.Parameter" do
      optional :name, :string, 1
      optional :value, :message, 2, "Function.Parameter.Value"
    end
    add_message "Function.Parameter.Value" do
      oneof :type do
        optional :string, :string, 1
        optional :integer, :int64, 2
      end
    end
  end
end

Function = ::Google::Protobuf::DescriptorPool.generated_pool.lookup("Function").msgclass
Function::Parameter = ::Google::Protobuf::DescriptorPool.generated_pool.lookup("Function.Parameter").msgclass
Function::Parameter::Value = ::Google::Protobuf::DescriptorPool.generated_pool.lookup("Function.Parameter.Value").msgclass
```
**What did you expect to see**
Under plain Ruby this script does not report any problem, as expected.
**What did you see instead?**
Under JRuby, this script fails with:
```
RuntimeError: Function.Parameter.value: ".Function.Parameter.Value" is not defined.
             add_file at com/google/protobuf/jruby/RubyBuilder.java:125
  function_call_pb.rb at function_call_pb.rb:7
        instance_eval at org/jruby/RubyBasicObject.java:2608
                build at com/google/protobuf/jruby/RubyDescriptorPool.java:78
               <main> at function_call_pb.rb:6
```
Please note that this Ruby script has been generated using `protoc`, version `3.14.0`.

tcruys:
@acozzette thanks for reinstating the JRuby support for protobuf.
[...]

PR #8194, opened 2021-01-08
Fix jruby support to handle messages nested more than 1 level deep

rdubya:
This fixes https://github.com/protocolbuffers/protobuf/issues/8191


hafixo:
Duplicate of 
Merge pull requested #8194

acozzette:
Thanks, @rdubya! I believe you will need to add the new test files [here](https://github.com/protocolbuffers/protobuf/blob/468bc193ec155756bccf2b2d9eeb8e003d95ce31/Makefile.am#L1135) to fix the test failures.

rdubya:
Hey @acozzette, sorry about missing that step again. I also updated the test structure to match the other ruby tests better.

rdubya:
@acozzette are those failures something else I need to change? The details links don't give me any info.

acozzette:
@rdubya Sorry about that, I can't see any details either. Let me rerun the tests and see if we get more informative results. If that doesn't fix it then there might be another problem because we have been having some issues with access to the test logs.

PR #8198, opened 2021-01-10
Resolve more java field accessor name conflicts

### LINKED ISSUE #8142: java output for message with field named "all_fields" cannot be compiled with javac
Version: 3.14.0
Language: Java

javac 11.0.9

**What did you do?**

```
$ cat <<EOF > test.proto
syntax = "proto2";

message AllFieldsMessage {
    optional bool all_fields = 1;
}
EOF

$ ./protoc --java_out=. test.proto

$ javac -cp protobuf-java-3.14.0.jar Test.java
Test.java:28: error: getAllFields() in AllFieldsMessageOrBuilder clashes with getAllFields() in MessageOrBuilder
    boolean getAllFields();
            ^
  return type boolean is not compatible with Map<FieldDescriptor,Object>
Test.java:129: error: getAllFields() in AllFieldsMessage cannot implement getAllFields() in MessageOrBuilder
    public boolean getAllFields() {
                   ^
  return type boolean is not compatible with Map<FieldDescriptor,Object>
Test.java:462: error: getAllFields() in Builder cannot implement getAllFields() in MessageOrBuilder
      public boolean getAllFields() {
                     ^
  return type boolean is not compatible with Map<FieldDescriptor,Object>
3 errors
```

javac successfully compiles the .java file generated by protoc

**What did you see instead?**

javac fails to compile the .java file due to a conflict with the `MessageOrBuilder.getAllFields()` method

philsttr:
Previously, some proto field names would cause the java code generator to generate accessor names that conflict with method names from the message super classes/interfaces, leading to java code that would not compile.
A list of field names that cause such conflicts previously existed, but the list did not contain every field name that would cause a conflict.
Additionally, only snake_case field names would be detected. If the field name was in camelCase, the conflict would not be detected.

This change adds the complete set of field names that will cause assessor name conflicts, and detects conflicts in both snake_case and camelCase field names.

Fixes #8142

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
+++ b/java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
@@ -61,8 +61,31 @@ import java.util.concurrent.ConcurrentHashMap;
 final class DescriptorMessageInfoFactory implements MessageInfoFactory {
   private static final String GET_DEFAULT_INSTANCE_METHOD_NAME = "getDefaultInstance";
   private static final DescriptorMessageInfoFactory instance = new DescriptorMessageInfoFactory();
+
+  /**
+   * Names that should be avoided as field names (in lowerCamelCase format).
+   * Using them will cause the compiler to generate accessors whose names are
+   * colliding with methods defined in base classes.
+   *
+   * Keep this list in sync with kForbiddenWordList in
+   * src/google/protobuf/compiler/java/java_helpers.cc
+   */
   private static final Set<String> specialFieldNames =
-      new HashSet<>(Arrays.asList("cached_size", "serialized_size", "class"));
+      new HashSet<>(Arrays.asList(
+        // java.lang.Object:
+        "class",
+        // com.google.protobuf.MessageLiteOrBuilder:
+        "defaultInstanceForType",
+        // com.google.protobuf.MessageLite:
+        "parserForType",
+        "serializedSize",
+        // com.google.protobuf.MessageOrBuilder:
+        "allFields",
+        "descriptorForType",
+        "initializationErrorString",
+        "unknownFields",
+        // obsolete. kept for backwards compatibility of generated code
+        "cachedSize"));
 
   // Disallow construction - it's a singleton.
   private DescriptorMessageInfoFactory() {}
@@ -593,8 +616,9 @@ final class DescriptorMessageInfoFactory implements MessageInfoFactory {
     String name = (fd.getType() == FieldDescriptor.Type.GROUP)
                   ? fd.getMessageType().getName()
                   : fd.getName();
-    String suffix = specialFieldNames.contains(name) ? "__" : "_";
-    return snakeCaseToCamelCase(name) + suffix;
+    String nameInCamelCase = snakeCaseToCamelCase(name);
+    String suffix = specialFieldNames.contains(nameInCamelCase) ? "__" : "_";
+    return nameInCamelCase + suffix;
   }
 
   private static String getCachedSizeFieldName(FieldDescriptor fd) {
--- a/.gitignore
+++ b/.gitignore
@@ -84,7 +84,7 @@ src/**/*.trs
 # JavaBuild output.
 java/core/target
 java/util/target
-javanano/target
+java/lite/target
 java/.idea
 java/**/*.iml
 
--- a/java/core/src/test/proto/com/google/protobuf/test_bad_identifiers.proto
+++ b/java/core/src/test/proto/com/google/protobuf/test_bad_identifiers.proto
@@ -43,10 +43,42 @@ option java_generic_services = true;  // auto-added
 option java_package = "com.google.protobuf";
 option java_outer_classname = "TestBadIdentifiersProto";
 
-message TestMessage {
-  optional string cached_size = 1;
-  optional string serialized_size = 2;
-  optional string class = 3;
+// Message with field names using underscores that conflict with accessors in the base message class in java.
+// See kForbiddenWordList in src/google/protobuf/compiler/java/java_helpers.cc
+message ForbiddenWordsUnderscoreMessage {
+  // java.lang.Object
+  optional bool class = 1;
+  // com.google.protobuf.MessageLiteOrBuilder
+  optional bool default_instance_for_type = 2;
+  // com.google.protobuf.MessageLite
+  optional bool parser_for_type = 3;
+  optional bool serialized_size = 4;
+  // com.google.protobuf.MessageOrBuilder
+  optional bool all_fields = 5;
+  optional bool descriptor_for_type = 6;
+  optional bool initialization_error_string = 7;
+  optional bool unknown_fields = 8;
+  // obsolete. kept for backwards compatibility of generated code
+  optional bool cached_size = 9;
+}
+
+// Message with field names in camel case that conflict with accessors in the base message class in java.
+// See kForbiddenWordList in src/google/protobuf/compiler/java/java_helpers.cc
+message ForbiddenWordsCamelMessage {
+  // java.lang.Object
+  optional bool class = 1;
+  // com.google.protobuf.MessageLiteOrBuilder
+  optional bool defaultInstanceForType = 2;
+  // com.google.protobuf.MessageLite
+  optional bool serializedSize = 3;
+  optional bool parserForType = 4;
+  // com.google.protobuf.MessageOrBuilder:
+  optional bool initializationErrorString = 5;
+  optional bool descriptorForType = 6;
+  optional bool allFields = 7;
+  optional bool unknownFields = 8;
+  // obsolete. kept for backwards compatibility of generated code
+  optional bool cachedSize = 9;
 }
 
 message Descriptor {
@@ -84,7 +116,7 @@ message Deprecated {
 
   optional int32 field1 = 1 [deprecated = true];
   optional TestEnum field2 = 2 [deprecated = true];
-  optional TestMessage field3 = 3 [deprecated = true];
+  optional ForbiddenWordsUnderscoreMessage field3 = 3 [deprecated = true];
 }
 
 message Override {
@@ -117,7 +149,7 @@ message Double {
 }
 
 service TestConflictingMethodNames {
-  rpc Override(TestMessage) returns (TestMessage);
+  rpc Override(ForbiddenWordsUnderscoreMessage) returns (ForbiddenWordsUnderscoreMessage);
 }
 
 message TestConflictingFieldNames {
@@ -125,24 +157,24 @@ message TestConflictingFieldNames {
     UNKNOWN = 0;
     FOO = 1;
   }
-  message TestMessage {}
+  message ForbiddenWordsUnderscoreMessage {}
   repeated int32 int32_field = 1;
   repeated TestEnum enum_field = 2;
   repeated string string_field = 3;
   repeated bytes bytes_field = 4;
-  repeated TestMessage message_field = 5;
+  repeated ForbiddenWordsUnderscoreMessage message_field = 5;
 
   optional int32 int32_field_count = 11;
   optional TestEnum enum_field_count = 12;
   optional string string_field_count = 13;
   optional bytes bytes_field_count = 14;
-  optional TestMessage message_field_count = 15;
+  optional ForbiddenWordsUnderscoreMessage message_field_count = 15;
 
   repeated int32 Int32Field = 21;          // NO_PROTO3
   repeated TestEnum EnumField = 22;        // NO_PROTO3
   repeated string StringField = 23;        // NO_PROTO3
   repeated bytes BytesField = 24;          // NO_PROTO3
-  repeated TestMessage MessageField = 25;  // NO_PROTO3
+  repeated ForbiddenWordsUnderscoreMessage MessageField = 25;  // NO_PROTO3
 
   // This field conflicts with "int32_field" as they both generate
   // the method getInt32FieldList().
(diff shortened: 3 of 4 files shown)


philsttr:
Can someone review this PR? 

Side note, I can't add the required labels (`release notes: yes`, `java`) since I do not have permission.

JonathanLeech:
Does this fix a field named _class?

philsttr:
@JonathanLeech It didn't before, but I just pushed an update to this PR so that it now handles fields named `_class` (and also for other conflicting names beginning with a leading underscore)

shaod2:
Thanks for the fix!

Since we're changing the field names somehow, can you also add a log warning somewhere (e.g. conflicting field names/will be changed)? I know there isn't one before, but I think it'd be good to add one here.

philsttr:
That's a good idea.

I might not be able to get to it for a little while.  Would you consider merging this bug-fix PR as-is?  And create an enhancement feature request for a warning message?  I can submit a new PR for that when I get a chance.

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,92 @@ static String getFieldName(FieldDescriptor fd) {
...
+   *     snakeCaseToLowerCamelCase("foo_bar") => "fooBar"
+   *     snakeCaseToLowerCamelCase("foo") => "foo"
+   * </pre>
+   * </p>
+   *
+   * @param snakeCase the string in snake case to convert
+   * @return the string converted to camel case, with a lowercase first character
+   */
+  private static String snakeCaseToLowerCamelCase(String snakeCase) {
+  	return snakeCaseToCamelCase(snakeCase, false);
```
elharo, line 654: nit: two space indents

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,92 @@ static String getFieldName(FieldDescriptor fd) {
...
   }
 
   /**
-   * This method must match exactly with the corresponding function in protocol compiler. See:
-   * https://github.com/google/protobuf/blob/v3.0.0/src/google/protobuf/compiler/java/java_helpers.cc#L153
+   * Converts a snake case string into upper camel case.
+   *
+   * <p>
+   * Some examples:
+   * <pre>
```
elharo, line 662: pre should be outside the p

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,92 @@ static String getFieldName(FieldDescriptor fd) {
...
+  }
+
+  /**
+   * Converts a snake case string into camel case.
+   *
+   * <p>For better readability, prefer calling either
+   * {@link #snakeCaseToLowerCamelCase(String)} or {@link #snakeCaseToUpperCamelCase(String)}.</p>
+   *
+   * <p>Some examples:
+   * <pre>
```
elharo, line 682: pre outside the p

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,92 @@ static String getFieldName(FieldDescriptor fd) {
...
+   * <p>This implementation of this method must exactly match the corresponding
+   * function in the protocol compiler.  Specifically, the
+   * {@code UnderscoresToCamelCase} function in
+   * {@code src/google/protobuf/compiler/java/java_helpers.cc}.</p>
+   *
+   * @param snakeCase the string in snake case to convert
+   * @param capFirst true if the first letter of the returned string should be uppercase.
+   *                 false if the first letter of the returned string should be lowercase.
+   * @return the string converted to camel case, with an uppercase or lowercase first
+   *         character depending on if {@code capFirst} is true of false, respectively
```
elharo, line 700: of --> or

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: why would you put \<pre> outside the \<p>   everything inside \<pre> is pre rendered stuff (which \<p> is not)

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: Fixed

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: Fixed

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: ah, I think I understand what you meant now.  (I thought you meant nest the p inside the pre, but you just meant separate the two).
Fixed

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: Fixed

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -61,8 +61,31 @@
 final class DescriptorMessageInfoFactory implements MessageInfoFactory {
   private static final String GET_DEFAULT_INSTANCE_METHOD_NAME = "getDefaultInstance";
   private static final DescriptorMessageInfoFactory instance = new DescriptorMessageInfoFactory();
+
+  /**
+   * Names that should be avoided (in UpperCamelCase format).
+   * Using them causes the compiler to generate accessors whose names are
```
elharo, line 67: are colliding --> collide

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,88 @@ static String getFieldName(FieldDescriptor fd) {
...
-    String suffix = specialFieldNames.contains(name) ? "__" : "_";
-    return snakeCaseToCamelCase(name) + suffix;
+
+    // convert to UpperCamelCase for comparison to the specialFieldNames
+    // (which are in UpperCamelCase)
+    String upperCamelCaseName = snakeCaseToUpperCamelCase(name);
+
+    String suffix = specialFieldNames.contains(upperCamelCaseName)
+                  // For field names that match the specialFieldNames,
+                  // append "__" to prevent field accessor method names from
```
elharo, line 626: Why both __ and _? Is there a conflict if both cases uses a single underscore?

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,88 @@ static String getFieldName(FieldDescriptor fd) {
...
+                  // clashing with other methods.
+                  ? "__"
+                  // For other field names, append "_" to prevent field names
```
elharo, line 629: Should field names here be field method accessor names as well?

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,88 @@ static String getFieldName(FieldDescriptor fd) {
...
+                  // from clashing with java keywords.
+                  : "_";
```
elharo, line 631: Don't split the ternary ?: operator across these lines. I found this really hard to read. Possibly this method should be split up, one for handling method names and one for handling field names. As is, I'm finding it very hard to grok. I did not initially understand what it was doing.

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: The single and double underscores referenced here are simply to match the java code that is currently generated by the protoc java compiler.  Notice the previous code used the single and double underscore.  In other words, my change does not change the existing semantics for single and double underscores.

The protoc java compiler currently (prior to my change) uses a single or double underscore depending on the proto field name.

The protoc java compiler always appends at least one underscore to java field names.  This prevents those field names from clashing with java keywords.  For example, if the proto field name was `int`, then the protoc java compiler would generate a java field name of 'int_'.

The protoc java compiler does not include the last trailing underscore in accessor method names for these fields.  For example, the getter method for the proto field named `int` would be `getInt()`.

In addition, the protoc java compiler adds _another_ underscore (for a total of two underscores) to the java field name for names that would result in accessor method names that clash with other method names.  For example, if the proto field name was `class`, the java field name would be `class__` (double underscore).  And again, the last trailing underscore is not included in the accessor method names.  So the getter method for the proto field named `class` would be `getClass_()`

This behavior existed prior to my change.   My change simply adds _more_ detection of field names that result in accessor method name clashes.  Previously, only `cached_size`, `serialized_size`, and `class` were detected.  Unfortunately, those are only a subset of the field names that would actually cause assessor method name conflicts.  My change adds the complete set of field names that cause accessor method name conflicts.

Again, I'm not changing the semantics of single or double underscores.  I am also not going to refactor the protoc compiler to change its current semantics for single and double underscores.  I'm simply fixing a bug in the protoc compiler where it was generating non-compilable java code due to missing detection of field names that cause conflicts.

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: No, the protoc compiler appends a single underscore specifically to the java field names to prevent clashes with java keywords.  The protoc compiler does not include this last undescore in accessor method names.

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
philsttr: Fixed

elharo (approved):

zhangskz:
Thanks for the contribution! This change seems alright, but does differ from similar existing behavior (https://github.com/protocolbuffers/protobuf/blob/d37dcf9e34809a7def529e591860eb6faa653f29/src/google/protobuf/compiler/java/java_context.cc#L163) which instead appends field numbers to handle clashes rather than underscores.

I'd like to hold off a bit before approving to take a closer look at how best to handle this as there are some pros/cons to each solution, but ultimately we'd probably want a consistent approach for this type of scenario.

philsttr:
I would like to reiterate that the underscores are existing behavior as well.  For example, the existing code (prior to my change) will generate a java field named `class_` and a getter method named `getClass_()` for a proto field named `class`.  And similarly for proto fields named `cached_size` and `serialized_size`.

This change just adds more detection of field names that cause conflicts, and utilizes the existing behavior to resolve those conflicts.

I understand the desire for collapsing the two previously existing behaviors though.   But note, you will not be able to change the existing behavior for fields named `class`, `cached_size`, and `serialized_size` without breaking compatibility.

philsttr:
Hi @zhangskz  What were the results of your closer look?

zhangskz:
I haven't had the opportunity to get through a complete audit of how we handle escaping across conflict types and languages. This would actually be a large undertaking that would need to be approached carefully since modifying this would indeed break compatibility. In any case, this shouldn't affect your fix since it doesn't really make the incompatibilities much worse -- we already use both underscores to escape in some cases. 

I'll review your CL shortly (aka tomorrow). Thanks for the patience!

java/core/src/main/java/com/google/protobuf/DescriptorMessageInfoFactory.java
```diff
@@ -593,21 +616,101 @@ static String getFieldName(FieldDescriptor fd) {
     String name = (fd.getType() == FieldDescriptor.Type.GROUP)
                   ? fd.getMessageType().getName()
                   : fd.getName();
-    String suffix = specialFieldNames.contains(name) ? "__" : "_";
-    return snakeCaseToCamelCase(name) + suffix;
+
+    // convert to UpperCamelCase for comparison to the specialFieldNames
+    // (which are in UpperCamelCase)
+    String upperCamelCaseName = snakeCaseToUpperCamelCase(name);
```
zhangskz, line 622: I think snakeCaseToUpperCamelCase is only ever used for checking against specialFieldNames. Can we just keep specialFieldNames in lower camel case instead? This would simplify our code and IMO is a bit more intuitive anyways.

philsttr: The specialFieldNames are in UpperCamelCase to handle proto fields named `_my_field`, since `_my_field` will be `MyField` when it is converted to camel case (either upper or lower).  It seemed more straightforward to keep the special field names in UpperCamelCase for comparison (since it handles proto fields named both `my_field`, `_my_field`, and `__my_field`), rather than adding additional logic to handle proto fields starting with `_`. 

Note the tests that use [`ForbiddenWordsLeadingUnderscoreMessage`](https://github.com/protocolbuffers/protobuf/pull/8198/files#diff-c20ad952face0ceea3afab022d736d584b0edf8065f378f052c72a77b0fd3c3fR67) test this case (pun intended ;)).

zhangskz (approved):

PR #8202, opened 2021-01-13
c++: let destructor in the Message class be Virtual

JingWoo:
let destructor in the Message class be Virtual, because when deleting a derived class object using a pointer of base class type that has a non-virtual destructor results in undefined behavior

Signed-off-by: wujing <jing.woo@outlook.com>


JingWoo:
@googlebot I signed it!

JingWoo:
Excuse me， Can you tell me how to include release notes: yes AND Please include at least a language label (e.g., c++, java, python)?

acozzette:
@JingWoo Thank you for the pull request but I actually don't think this change is necessary, since Message already inherits the virtual destructor from MessageLite.

PR #8219, opened 2021-01-20
Include the size of values (and tags) for extensions, even if the value is the default

### LINKED ISSUE #8218: ExtensionSet is inconsistent for default values
Version: master
Language: C#

Irrelevant (testing on Windows, but...)

n/a

**What did you do?**

The following test fails in the call to `ToByteArray()`:

```csharp
[Test]
public void TestDefaultValueRoundTrip()
{
    var message = new TestAllExtensions();
    message.SetExtension(OptionalBoolExtension, false);
    Assert.IsFalse(message.GetExtension(OptionalBoolExtension));
    Assert.IsTrue(message.HasExtension(OptionalBoolExtension));

    var bytes = message.ToByteArray();
    var registry = new ExtensionRegistry { OptionalBoolExtension };
    var parsed = TestAllExtensions.Parser.WithExtensionRegistry(registry).ParseFrom(bytes);
    Assert.IsFalse(parsed.GetExtension(OptionalBoolExtension));
    Assert.IsTrue(parsed.HasExtension(OptionalBoolExtension));
}
```

A passing test.

**What did you see instead?**

The test throws during `ToByteArray()`, because the extension value has a calculated size of 0, due to being the default value. We want to actually write the extension to the output, and we try to... but at that point, not enough space has been allocated.

I have a fix for this, but wanted to file it as an issue first. This is pretty serious in terms of not being able to serialize descriptors that happen to include a default value for an extension - although that's quite rare.

jskeet:
Fixes #8218.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/ExtensionSetTest.cs
+++ b/csharp/src/Google.Protobuf.Test/ExtensionSetTest.cs
@@ -116,7 +116,22 @@ namespace Google.Protobuf
             var other = message.Clone();
 
             Assert.AreEqual(message, other);
-            Assert.AreEqual(message.CalculateSize(), message.CalculateSize());
+            Assert.AreEqual(message.CalculateSize(), other.CalculateSize());
+        }
+
+        [Test]
+        public void TestDefaultValueRoundTrip()
+        {
+            var message = new TestAllExtensions();
+            message.SetExtension(OptionalBoolExtension, false);
+            Assert.IsFalse(message.GetExtension(OptionalBoolExtension));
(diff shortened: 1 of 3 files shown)


jskeet:
Ping @haberman @jtattermusch

jtattermusch (approved):
LGTM from the C# perspective.
Also LGTM from the protobuf perspective but I'd rather defer to @haberman 's opinion there.

haberman:
Yes this LGTM. Extensions should always be serialized if present, and should never have the proto3 behavior of comparing with the default.

PR #8223, opened 2021-01-21
feat: add support for deprecated fields to PHP compiler

### LINKED ISSUE #8211: [PHP] add support for field option "[deprecated=true]"
PHP (using proto3 with protoc-gen-php)

As a developer of a service client, I would like to be able to identify deprecated message fields instantly in order to not use them anymore and replace them correctly.

This could be implemented by adding a `@deprecated` annotation to the field accessors of the generated code, as already stated in the [documentation](https://developers.google.com/protocol-buffers/docs/overview#options).

Alternatively, triggering `error_log()` with a `$message_type = E_USER_DEPRECATED` would have a similar effect.

**Additional context**
-

cawolf:
fixes #8211 

Please feel free to leave comments for improvements, especially for the compiler part, as I am not fluent in C.

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -645,57 +645,49 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
   GenerateFieldDocComment(printer, field, options, kFieldGetter);
 
   // deprecation
-  std::string deprecation_std_string = (field->options().deprecated()) ? "@trigger_error('" +
-      field->name() +
-      " is deprecated and will be removed in the next major release', E_USER_DEPRECATED);" : "";
-  const char* deprecation_string = deprecation_std_string.c_str();
+  std::string deprecation_trigger = (field->options().deprecated()) ? "@trigger_error('" +
+      field->name() + " is deprecated.', E_USER_DEPRECATED);\n        " : "";
 
   if (oneof != NULL) {
     printer->Print(
         "public function get^camel_name^()\n"
         "{\n"
-        "    ^deprecation_trigger^\n"
-        "    return $this->readOneof(^number^);\n"
+        "    ^deprecation_trigger^return $this->readOneof(^number^);\n"
         "}\n\n"
         "public function has^camel_name^()\n"
         "{\n"
-        "    ^deprecation_trigger^\n"
-        "    return $this->hasOneof(^number^);\n"
+        "    ^deprecation_trigger^return $this->hasOneof(^number^);\n"
         "}\n\n",
         "camel_name", UnderscoresToCamelCase(field->name(), true),
         "number", IntToString(field->number()),
-        "deprecation_trigger", deprecation_string);
+        "deprecation_trigger", deprecation_trigger);
   } else if (field->has_presence()) {
     printer->Print(
         "public function get^camel_name^()\n"
         "{\n"
-        "    ^deprecation_trigger^\n"
-        "    return isset($this->^name^) ? $this->^name^ : ^default_value^;\n"
+        "    ^deprecation_trigger^return isset($this->^name^) ? $this->^name^ : ^default_value^;\n"
         "}\n\n"
         "public function has^camel_name^()\n"
         "{\n"
-        "    ^deprecation_trigger^\n"
-        "    return isset($this->^name^);\n"
+        "    ^deprecation_trigger^return isset($this->^name^);\n"
         "}\n\n"
         "public function clear^camel_name^()\n"
         "{\n"
-        "    ^deprecation_trigger^\n"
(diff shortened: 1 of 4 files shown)


src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -644,43 +644,58 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
...
+  std::string deprecation_std_string = (field->options().deprecated()) ? "@trigger_error('" +
+      field->name() +
+      " is deprecated and will be removed in the next major release', E_USER_DEPRECATED);" : "";
+  const char* deprecation_string = deprecation_std_string.c_str();
+
   if (oneof != NULL) {
     printer->Print(
         "public function get^camel_name^()\n"
         "{\n"
+        "    ^deprecation_trigger^\n"
```
haberman, line 657: Please collapse the newlines so that a non-deprecated field has the same generated code as today.

You can run `generated_descriptor_proto.sh` in the base directory to regenerate the PHP files with the new compiler.

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -644,43 +644,58 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
   // Generate getter.
   GenerateFieldDocComment(printer, field, options, kFieldGetter);
 
+  // deprecation
+  std::string deprecation_std_string = (field->options().deprecated()) ? "@trigger_error('" +
```
haberman, line 648: Can you confirm that this outputs a warning only, and won't halt execution of the script?

haberman (changes requested):

src/google/protobuf/compiler/php/php_generator.cc
cawolf: Yes, I can confirm it. That's why the `@` character is added. This way, you can opt in to handle deprecations, compare the [current Symfony LTS documentation](https://symfony.com/doc/4.4/contributing/code/conventions.html#deprecating-code).

src/google/protobuf/compiler/php/php_generator.cc
cawolf: I get what you mean, but I do not know how. I ended up with this solution to have a properly formatted output file, whether or not a field is deprecated.

Can you please give me a hint?

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -644,43 +644,58 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
...
+      field->name() +
+      " is deprecated and will be removed in the next major release', E_USER_DEPRECATED);" : "";
```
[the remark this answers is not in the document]
cawolf: Can you please explain that to me? What is `'foo'` in your example?

If `'foo'` was a message, then you would be right, but this PR is about field deprecation.
If `'foo'` was a field, as a consuming developer, I would like to know which field is deprecated, not that there is a deprecation in the containing message.
Or did you mean something else?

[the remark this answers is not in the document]
cawolf: Or do you even meant to shorten the whole output message too `field->name() + " is deprecated."`?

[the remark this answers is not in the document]
haberman: Yes exactly, just `field->name() + " is deprecated."`.  We shouldn't presume to say when it will be removed.

src/google/protobuf/compiler/php/php_generator.cc
haberman: Maybe try:

```c++
    std::string deprecation_warning;
    if (field->options().deprecated()) {
        // Note trailing newline and indenting spaces.
        deprecation_warning = "@trigger_error('" +
             field->name() + " is deprecated', E_USER_DEPRECATED);\n    ";
    }
    printer->Print(
        "public function get^camel_name^()\n"
        "{\n"
        "    ^deprecation_warning^return $this->readOneof(^number^);\n"
        "}\n\n",
        "camel_name", UnderscoresToCamelCase(field->name(), true),
        "number", IntToString(field->number()),
        "deprecation_warning", deprecation_warning);
```

src/google/protobuf/compiler/php/php_generator.cc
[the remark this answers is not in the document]
cawolf: Ok understood, I just wanted to keep the messages concise, but I will change it.

src/google/protobuf/compiler/php/php_generator.cc
cawolf: Thanks, this helped.

PR #8224, opened 2021-01-22

deannagarcia:


src/google/protobuf/generated_message_table_driven_lite.h
```diff
@@ -87,7 +87,7 @@ inline ExtensionSet* GetExtensionSet(MessageLite* msg, int64 extension_offset) {
 
 template <typename Type>
 inline Type* AddField(MessageLite* msg, int64 offset) {
-  static_assert(std::is_standard_layout<Type>::value && std::is_trivial<Type>::value,
+  static_assert(std::is_pod<Type>::value,
```
haberman, line 90: Ah, I think this is a mistake. Our extract script is currently rewriting this to `is_pod`, but we need to stop doing that, otherwise we will revert https://github.com/protocolbuffers/protobuf/pull/7180.

deannagarcia: Fixed!

haberman: I should have mentioned that this applies to all cases of std::is_pod in this CL. There are several instances across several files.

deannagarcia: Ok, I think I fixed all instances of this!

haberman (approved):

PR #8231, opened 2021-01-25
Use rake-compiler-dock for x86_64-darwin and arm64-darwin

qnighy:
rake-compiler-dock 1.1 added support for cross-compilation targeting x86_64-darwin and arm64-darwin. This PR removes native darwin compilation and instead use the cross-compilation functionality.

One concern is that I'm not sure whether this is a sustainable solution regarding Mac OS SDK used in the docker image. At least nokogiri seems to be using the darwin cross-compilation.


dlj-NaN:
This PR does not seem suitable. The `kokoro` tree is tied quite closely to our internal CI configuration, so this PR would break things. At a higher level, though, I don't see how cross-building would be beneficial.

PR #8236, opened 2021-01-27
fix: README supported PHP version for C extension

bshaffer:


php/README.md
```diff
@@ -16,7 +16,7 @@ generation functionality.
 
 To use PHP runtime library requires:
 
-- C extension: PHP 5.5, 5.6, or 7.
+- C extension: PHP 7
```
haberman, line 19: Also PHP8 now (it's not released yet, but the code is in GitHub and will be released in the next few weeks).

Issue #8233, opened 2021-01-26
PHP protobuf extension crashes after cloning a protobuf object


tony2001:
**What version of protobuf and what language are you using?**
PHP extension version: 3.14.0
Language: PHP

**What operating system (Linux, Windows, ...) and version?**
Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**
GCC 5.3, GCC 7.5

**What did you do?**
```
<?php
$t = new \Google\Protobuf\Timestamp();
$tNew = clone $t;
$tNew->serializeToJsonString();
```

**What did you expect to see**
No crash

**What did you see instead?**
Segmentation fault

```
Thread 1 "php" received signal SIGSEGV, Segmentation fault.
0x00007ffff49aa1c6 in zim_Message_serializeToJsonString (execute_data=0x7ffff4e140f0, return_value=0x7fffffffa2b0)
    at /build/protobuf-3.14.0/message.c:750
750       size = upb_json_encode(intern->msg, intern->desc->msgdef,
(gdb) bt
#0  0x00007ffff49aa1c6 in zim_Message_serializeToJsonString (execute_data=0x7ffff4e140f0, return_value=0x7fffffffa2b0)
    at /build/protobuf-3.14.0/message.c:750
#1  0x0000000000741b62 in execute_ex ()
#2  0x0000000000743f13 in zend_execute ()
#3  0x00000000006ab6fd in zend_execute_scripts ()
#4  0x000000000063f48c in php_execute_script ()
#5  0x0000000000746495 in do_cli ()
#6  0x0000000000441cd5 in main ()
#7  0x00007ffff569aa35 in __libc_start_main () from /lib64/libc.so.6
#8  0x0000000000442442 in _start ()
(gdb) p *intern
$1 = {std = {gc = {refcount = 2, u = {type_info = 24}}, handle = 5, ce = 0x7ffff4e043b0, handlers = 0xbff7c0 <std_object_handlers>, properties = 0x0, 
    properties_table = {{value = {lval = 140737302307656, dval = 6.9533466158586819e-310, counted = 0x7ffff4e92348, str = 0x7ffff4e92348, 
          arr = 0x7ffff4e92348, obj = 0x7ffff4e92348, res = 0x7ffff4e92348, ref = 0x7ffff4e92348, ast = 0x7ffff4e92348, zv = 0x7ffff4e92348, 
[...]

PR #8245, opened 2021-02-01
Fixed clone for Message, RepeatedField, and MapField.

haberman:
Also updated the code to use a TypeInfo struct for convenient
passing of type and desc together. This simplified a lot of code
and made this change easier to write.

Fixes https://github.com/protocolbuffers/protobuf/issues/8233

### DIFF
--- a/php/ext/google/protobuf/array.c
+++ b/php/ext/google/protobuf/array.c
@@ -55,8 +55,7 @@ typedef struct {
   zend_object std;
   zval arena;
   upb_array *array;
-  upb_fieldtype_t type;
-  const Descriptor* desc;  // When values are messages.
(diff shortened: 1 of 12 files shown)


gerben-s (commented):
What exactly went wrong and how does this fix it?

haberman:
@gerben-s see the attached bug: https://github.com/protocolbuffers/protobuf/issues/8233

This PR fixes the problem by implementing the `clone_obj` handler for `Message`, `RepeatedField`, and `MapField`. `clone_obj` is a function pointer that we give to the PHP interpreter, PHP will call it when a user writes `clone $msg`.

gerben-s (approved):

PR #8254, opened 2021-02-03
Improvement: better ruby from_* support in well_known_types.rb

jufemaiz:
* Added capability to support from_* requests properly by adding class methods and returning self for instance methods
    * `Timestamp.from_time`
    * `Value.from_ruby`


DerekStride (approved):
Looks like a great addition to me!

jufemaiz:
@haberman you seem to be the lead for Ruby, would you mind reviewing and indicating any changes that Google would like to see?

jufemaiz:
@deannagarcia can you please review regarding the addition of a class method and the return of `self` on the instance method (it's not a `!` method so I would expect to receive self back a opposed to `value` of or similar).

PR #8256, opened 2021-02-03
Reserve extension id for protoc-gen-psql & protoc-gen-sanitize

upils:
We have created 2 projects we expect might be used by others so we would like to avoid extension id collision.


upils:
@googlebot I signed it!

PR #8270, opened 2021-02-08
Expand reserved IDs for protoc-gen-psql and protoc-gen-sanititize

upils:
Sorry, I forgot that I should reserve a range of IDs and not a single one for our projects.


dlj-NaN (approved):
This is just updating the change from 59b3d97f6d7e3ff67bc5d89f030017c60b483549.

PR #8271, opened 2021-02-08
Support to generate gRPC stubs from setup.py

Sathvik777:
These changes will run `grpc_python_out` when  `grpc_proto_files` are provided.
The changes will also not affect the current implementation as the new fields are optional and the `grpc_python_out` command runs after the original `spawn` (python_out) is done.

### DIFF
--- a/python/protobuf_distutils/README.md
+++ b/python/protobuf_distutils/README.md
@@ -35,7 +35,9 @@ setup(
             'source_dir':        'path/to/protos',
             'extra_proto_paths': ['path/to/other/project/protos'],
             'output_dir':        'path/to/project/sources',  # default '.'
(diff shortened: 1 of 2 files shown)


dlj-NaN:
Hmm, it looks like this would be more appropriate to live in gRPC. The protobuf suite doesn't include the gRPC code generator, so it seems quite odd to invoke something that would fail.

Sathvik777:
> Hmm, it looks like this would be more appropriate to live in gRPC.

This makes sense. Do you happen to know if they have a `distutils` for python that I can contribute to? 
Any reference is much appreciated.  :) 

>The protobuf suite doesn't include the gRPC code generator, so it seems quite odd to invoke something that would fail.

But in we spawn a `protoc` subprocess that has a gRPC code generator. Shouldn't this work?

PR #8272, opened 2021-02-09
Opensourcing kotlin protos

deannagarcia:
Adding kotlin syntax to java protobufs.


.github/workflows/codespell.yml
```diff
@@ -13,4 +13,4 @@ jobs:
         with:
           check_filenames: true
           skip: ./.git,./conformance/third_party,*.snk,*.pb,*.pb.cc,*.pb.h,./src/google/protobuf/testdata,./objectivec/Tests,./python/compatibility_tests/v2.5.0/tests/google/protobuf/internal
-          ignore_words_list: "alow,alse,ba,cleare,copyable,cloneable,dedup,dur,errorprone,files',fo,fundementals,hel,importd,inout,leapyear,nd,nin,ois,ons,parseable,process',te,testof,ue,unparseable,wasn,wee,gae,keyserver,objext,od,OptIn"
+          ignore_words_list: "alow,alse,ba,cleare,copyable,cloneable,dedup,dur,errorprone,files',fo,fundementals,hel,importd,inout,leapyear,nd,nin,ois,ons,parseable,process',te,testof,ue,unparseable,wasn,wee,gae,keyserver,objext,od,optin"
```
perezd, line 16: Should this be removed? I think this made its way into a different PR.

java/kotlin-lite/pom.xml
```diff
@@ -0,0 +1,383 @@
...
+  <parent>
+    <groupId>com.google.protobuf</groupId>
+    <artifactId>protobuf-parent</artifactId>
+    <version>3.15.6</version>
+  </parent>
+
+  <artifactId>protobuf-kotlin-lite</artifactId>
+  <packaging>bundle</packaging>
+
+  <name>Protocol Buffers [Lite]</name>
```
perezd, line 13: nit: Do we want Kotlin in the name somewhere?

java/kotlin/generate-sources-build.xml
```diff
@@ -0,0 +1,20 @@
+<project name="generate-sources">
+    <echo message="Running protoc ..."/>
+    <mkdir dir="${generated.sources.dir}"/>
+    <exec executable="${protoc}">
+        <arg value="--java_out=${generated.sources.dir}"/>
+        <arg value="--proto_path=${protobuf.source.dir}"/>
```
perezd, line 6: We should likely expose a kotlin version of all of these proto definitions as they are normally part of the public API.

java/kotlin-lite/pom.xml
deannagarcia: I'm not sure how these naming conventions are supposed to work but I copied this from Java because it wasn't java specific either.

.github/workflows/codespell.yml
deannagarcia: It turns out that the spell check has to have the lowercase version, so this is the fix to the other PR.

java/kotlin/generate-sources-build.xml
deannagarcia: Done!

PR #8274, opened 2021-02-09
add string setter from std::string_view (c++17 and above)

lavv17:
c++

release notes: allow setting string fields from std::string_view in c++17 and above.

Signed-off-by: Alexander V. Lukyanov <lavv17f@gmail.com>


lavv17:
@googlebot I signed it!

acozzette:
@lavv17 Thanks so much for this pull request. Unfortunately we can't accept this change, but the good news is that this is because @jorgbrown has already implemented a separate refactoring of the string setters which will enable use of `std::string_view`. This will be available soon and we just need to sync it from our internal codebase.

jorgbrown:
@lavv17 Just curious: Have you tried editing src/google/protobuf/stubs/port.h, and changing:
`using ConstStringParam = const std::string &;`
to
`using ConstStringParam =std::string_view;`
?

lavv17:
> @lavv17 Just curious: Have you tried editing src/google/protobuf/stubs/port.h, and changing:
> `using ConstStringParam = const std::string &;`
> to
> `using ConstStringParam =std::string_view;`
> ?

No, I have not. BTW, I need string_view reference, because I have a derived class. Without reference it will lead to object slicing (at least warning).

PR #8275, opened 2021-02-09
Added PHP changes for 3.15.0 into package.xml

acozzette:


haberman:
Thanks for doing this.

We don't have an automated solution for https://github.com/protocolbuffers/protobuf/issues/7813 yet, but manual changes like this can hold us over until we do.

PR #8280, opened 2021-02-10
Enable crosscompiling aarch64 python wheels under dockcross manylinux docker image

jtattermusch:
Build crosscompiled aarch64 python wheels under dockcross manylinux2014 image.

Supersedes https://github.com/protocolbuffers/protobuf/pull/8196 (some of the challenges I'm solving in this PR are already described there).

Key points:
- manylinux2014 should be good enough for the aarch wheels (manylinux2014 seems to be the first manylinux version that has official aarch64 support).
- a few workarounds are applied but overall the diff is pretty small and change changes integrate pretty well with the existing multibuild-based workflow for building python wheels.
- the build time for an aarch64 wheel is comparable to the build of x86_64 wheel (which is expected, as this is doing crosscompilation).
- the advantage of the setup in this PR is that all the linux wheels can be built on a single kokoro x64 machine and there are no changes needed to the release process and / or the way the wheels are being collected after being built (which seems quite important due to the complexity of the release process). Another obvious advantage is that no arm64 hardware is needed.

I smoke tested the resulting aarch64 wheel manually on a real arm64 machine and I also ran "auditwheel show" and no issues were reported - so I'm quite confident that they work just fine.

### DIFF
--- a/kokoro/release/python/linux/config.sh
+++ b/kokoro/release/python/linux/config.sh
@@ -6,14 +6,37 @@ function pre_build {
     # Runs in the root directory of this repository.
     pushd protobuf
 
-    yum install -y devtoolset-2-libatomic-devel
+    if [ "$PLAT" == "aarch64" ]
+    then
+      local configure_host_flag="--host=aarch64"
+    else
+      yum install -y devtoolset-2-libatomic-devel
+    fi
 
-    # Build protoc
+    # Build protoc and libprotobuf
     ./autogen.sh
-    ./configure
-
-    CXXFLAGS="-fPIC -g -O2" ./configure
+    CXXFLAGS="-fPIC -g -O2" ./configure $configure_host_flag
     make -j8
+
+    if [ "$PLAT" == "aarch64" ]
+    then
+      # we are crosscompiling for aarch64 while running on x64
+      # the simplest way for build_py command to be able to generate
+      # the protos is by running the protoc process under
+      # an emulator. That way we don't have to build a x64 version
+      # of protoc. The qemu-arm emulator is already included
+      # in the dockcross docker image.
+      # Running protoc under an emulator is fast as protoc doesn't
+      # really do much.
+
+      # create a simple shell wrapper that runs crosscompiled protoc under qemu
+      echo '#!/bin/bash' >protoc_qemu_wrapper.sh
+      echo 'exec qemu-aarch64 "../src/protoc" "$@"' >>protoc_qemu_wrapper.sh
+      chmod ugo+x protoc_qemu_wrapper.sh
+
+      # PROTOC variable is by build_py step that runs under ./python directory
+      export PROTOC=../protoc_qemu_wrapper.sh
+    fi
 
     # Generate python dependencies.
     pushd python
@@ -35,7 +58,20 @@ function bdist_wheel_cmd {
     # Modify build version
     pwd
     ls
-    python setup.py bdist_wheel --cpp_implementation --compile_static_extension
+
+    if [ "$PLAT" == "aarch64" ]
+    then
+      # when crosscompiling for aarch64, --plat-name needs to be set explicitly
+      # to end up with correctly named wheel file
+      # the value should be manylinuxABC_ARCH and dockcross docker image
+      # conveniently provides the value in the AUDITWHEEL_PLAT env
+      local plat_name_flag="--plat-name=$AUDITWHEEL_PLAT"
+
+      # override the value of EXT_SUFFIX to make sure the crosscompiled .so files in the wheel have the correct filename suffix
+      export PROTOCOL_BUFFERS_OVERRIDE_EXT_SUFFIX="$(python -c 'import sysconfig; print(sysconfig.get_config_var("EXT_SUFFIX").replace("-x86_64-linux-gnu.so", "-aarch64-linux-gnu.so"))')"
+    fi
+
+    python setup.py bdist_wheel --cpp_implementation --compile_static_extension $plat_name_flag
     cp dist/*.whl $abs_wheelhouse
 }
 
@@ -48,3 +84,12 @@ function run_tests {
     python --version
     python -c "from google.protobuf.pyext import _message;"
 }
+
+if [ "$PLAT" == "aarch64" ]
+then
+  # when crosscompiling for aarch64, override the default multibuild's repair_wheelhouse logic
+  # since "auditwheel repair" doesn't work for crosscompiled wheels
+  function repair_wheelhouse {
+      echo "Skipping repair_wheelhouse since auditwheel requires build architecture to match wheel architecture."
+  }
+fi
--- a/kokoro/release/python/linux/build_artifacts.sh
+++ b/kokoro/release/python/linux/build_artifacts.sh
@@ -49,9 +49,23 @@ build_artifact_version() {
   mv wheelhouse/* $ARTIFACT_DIR
 }
 
+build_crosscompiled_aarch64_artifact_version() {
+  # crosscompilation is only supported with the dockcross manylinux2014 image
+  DOCKER_IMAGE=dockcross/manylinux2014-aarch64
+  PLAT=aarch64
+
+  # TODO(jtatermusch): currently when crosscompiling, "auditwheel repair" will be disabled
+  # since auditwheel doesn't work for crosscomiled wheels.
+  build_artifact_version $@
+}
+
 build_artifact_version 2.7
 build_artifact_version 3.5
 build_artifact_version 3.6
 build_artifact_version 3.7
 build_artifact_version 3.8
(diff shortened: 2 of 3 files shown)


jtattermusch:
@haberman  @janaknat  please also take a look.

kokoro/release/python/linux/config.sh
```diff
@@ -6,15 +6,38 @@ function pre_build {
...
+      # Running protoc under an emulator is fast as protoc doesn't
+      # really do much.
+
+      # create a simple shell wrapper that runs crosscompiled protoc under qemu
+      echo '#!/bin/bash' >protoc_qemu_wrapper.sh
+      echo 'exec qemu-aarch64 "../src/protoc" "$@"' >>protoc_qemu_wrapper.sh
+      chmod ugo+x protoc_qemu_wrapper.sh
+
+      # PROTOC variable is by build_py step that runs under ./python directory
+      export PROTOC=../protoc_qemu_wrapper.sh
```
janaknat, line 38: You can export a function in bash:
```
PROTOC() { qemu-aarch64 ../src/protoc; }
export -f PROTOC
```
It might be a bit simpler.

kokoro/release/python/linux/config.sh
```diff
@@ -48,3 +84,12 @@ function run_tests {
     python --version
     python -c "from google.protobuf.pyext import _message;"
 }
+
+if [ "$PLAT" == "aarch64" ]
+then
+  # when crosscompiling for aarch64, override the default multibuild's repair_wheelhouse logic
+  # since "auditwheel repair" doesn't work for crosscompiled wheels
+  function repair_wheelhouse {
+      echo "Skipping repair_wheelhouse since auditwheel requires build architecture to match wheel architecture."
```
janaknat, line 93: build_wheel() can be modified to use a qemu-aarch64 wrapper with auditwheel:
```
# Based on multibuild build_wheel_cmd(). The only change is calling the repair_wheelhouse with qemu-aarch64.
function build_wheel_aarch64 {
    local cmd=${1:-pip_wheel_cmd}
    local repo_dir=${2:-$REPO_DIR}
    [ -z "$repo_dir" ] && echo "repo_dir not defined" && exit 1
    local wheelhouse=$(abspath ${WHEEL_SDIR:-wheelhouse})
    start_spinner
    if [ -n "$(is_function "pre_build")" ]; then pre_build; fi
    stop_spinner
    if [ -n "$BUILD_DEPENDS" ]; then
        pip install $(pip_opts) $BUILD_DEPENDS
    fi
    (cd $repo_dir && $cmd $wheelhouse)
    qemu-aarch64 repair_wheelhouse $wheelhouse
}

function build_wheel {
    if [ $ARCH == "arm64" ]; then
        build_wheel_aarch64 "bdist_wheel_cmd" $@
    else
        build_wheel_cmd "bdist_wheel_cmd" $@
    fi
}
```
This way you can keep the build and repair process within the container.

jtattermusch: Interesting idea, but are you sure this is actually something that works? It seemed to me that you'd need to be in a aarch64 linux image (with arm-based python and binaries auditwheel uses, e.g. patchelf) for this to work.

hanxiao:
only one broken test left? way to go & looking forward to the merge!

jtattermusch:
@dlj-NaN  @haberman friendly ping. I'd really like to hear feedback on whether this is an approach we could use (I believe so).

I'm aware that before publishing "official" version of the aarch64 wheels we'd need to have some aarch64 tests in place first (I have some ideas around how to set that up aarch64 tests under an emulator), but that is something that should be done in a separate PR. The wheel crosscompilation itself (as presented in this PR) seems quite reasonable to me. If this looks good, we could look into setting up emulated tests as the next step.

jtattermusch:
> only one broken test left? way to go & looking forward to the merge!

The "Linux C++ Distcheck" failure is very likely unrelated to this PR.

kokoro/release/python/linux/config.sh
dlj-NaN: The var is not passed to a shell, so it has to be an executable.

kokoro/release/python/linux/config.sh
```diff
@@ -35,7 +58,20 @@ function bdist_wheel_cmd {
...
+    if [ "$PLAT" == "aarch64" ]
+    then
+      # when crosscompiling for aarch64, --plat-name needs to be set explicitly
+      # to end up with correctly named wheel file
+      # the value should be manylinuxABC_ARCH and dockcross docker image
+      # conveniently provides the value in the AUDITWHEEL_PLAT env
+      local plat_name_flag="--plat-name=$AUDITWHEEL_PLAT"
+
+      # override the value of EXT_SUFFIX to make sure the crosscompiled .so files in the wheel have the correct filename suffix
+      export PROTOCOL_BUFFERS_OVERRIDE_EXT_SUFFIX="$(python -c 'import sysconfig; print(sysconfig.get_config_var("EXT_SUFFIX").replace("-x86_64-linux-gnu.so", "-aarch64-linux-gnu.so"))')"
```
dlj-NaN, line 71: It would be neat if you could do something like this:

`$(qemu-aarch64 python -c 'import sysconfig; print(sysconfig.get_config_var("EXT_SUFFIX")')`

I don't think that would currently work, since qemu won't necessarily have access to its own python... unless I missed something?

dlj-NaN (approved):
One question before I merge this...

kokoro/release/python/linux/config.sh
jtattermusch: Yes, currently this won't work as there is no arm-specific installation of python in the docker image.
We can revisit things like that in the next iteration (along with figuring out how to run auditwheel checks e.g. under an emulator). I am going to continue looking into improvements like this while I'm experimenting with running protobuf tests under an emulator.

jtattermusch:
@dlj-NaN  thanks for the review! Note that once we merge this, the python release builds will start generating the aarch64 wheels automatically (which I think according to your release process will end up uploading aarch64 wheels to pypi once the next release happens). If you want to avoid publishing the aarch64 wheels (until we have more tests in place), we need to explicitly remove them from the list of wheels that get published - let me know if you want me to look into that.

jtattermusch:
> @dlj-NaN thanks for the review! Note that once we merge this, the python release builds will start generating the aarch64 wheels automatically (which I think according to your release process will end up uploading aarch64 wheels to pypi once the next release happens). If you want to avoid publishing the aarch64 wheels (until we have more tests in place), we need to explicitly remove them from the list of wheels that get published - let me know if you want me to look into that.

@dlj-NaN   please let me know if I should look into removing the aarch64 wheels from the published set or if it's ok to leave as is (and merge this).

dlj-NaN:
Yes, let's remove the wheels from the distribution until we can test them.

jtattermusch:
@dlj-NaN   I just added a PR that provides a sufficient level of testing (at least IMHO) under aarch64 emulator: https://github.com/protocolbuffers/protobuf/pull/8392. The auditwheel check passes, the wheels are installable and all the python unit tests (with cpp extension enabled) are passing.

dlj-NaN:
> @dlj-NaN I just added a PR that provides a sufficient level of testing (at least IMHO) under aarch64 emulator: #8392. The auditwheel check passes, the wheels are installable and all the python unit tests (with cpp extension enabled) are passing.

Until we have a chance to actually verify behavior and stability, it still seems prudent not to include release wheels yet.

jtattermusch:
> > @dlj-NaN I just added a PR that provides a sufficient level of testing (at least IMHO) under aarch64 emulator: #8392. The auditwheel check passes, the wheels are installable and all the python unit tests (with cpp extension enabled) are passing.
> 
> Until we have a chance to actually verify behavior and stability, it still seems prudent not to include release wheels yet.

Sounds fair. I have made a small change that puts the aarch64 manylinux artifacts into artifacts/unofficial directory, so that they are still being stored (for manual testing or unofficial releases), but they won't be uploaded to PyPI since the protobuf python release process only considers wheels that are directly under the artifacts/ directory.

With this change, the PR should be good to merge.

hanxiao:
> so that they are still being stored (for manual testing or unofficial releases)

is there a link where one can fetch those unofficial releases/wheels?

jtattermusch:
> > so that they are still being stored (for manual testing or unofficial releases)
> 
> is there a link where one can fetch those unofficial releases/wheels?

They are accessible through the "Linux Python Release" job's artifact. Unfortunately at this point, the results for those jobs are internal-only. Let's look into making the aarch64 wheels externally accessible as a follow up step.

hanxiao:
> Let's look into making the aarch64 wheels externally accessible as a follow up step.

👍

iemejia:
Any ETA for the aarch64 wheels into PyPI ?

PR #8285, opened 2021-02-11
Fix missing -llog when compiling for android

s1341:
When cross-compiling for android, the android logging code is activated, but liblog.so is not linked in.

This patch fixes that.


s1341:
@googlebot I signed it!

s1341:
What are the checks which failed? Clicking details gives me an 'oops' page.

acozzette:
I think the test failures are probably caused by an unrelated issue that we've since fixed, let me rerun them and see if they pass now.

s1341:
Awesome. This seems ready to merge!

s1341:
Thanks!

PR #8290, opened 2021-02-12
Add Android stlport and default toolchains to BUILD.

blueshiftlabs:
These are additional possibilities for --crosstool_top for Android NDK
compilation. Since the NDK doesn't have -lpthread, getting these flags
wrong leads to a linker error.

Fixes: 180084220


blueshiftlabs:
Done, PTAL.

PR #8294, opened 2021-02-13
Some doc on AOT compilation and protobuf

huntc:
Some initial text pertaining to the use of protobuf with AOT compilation tools such as the one provided by GraalVM native-image.

### DIFF
--- a/docs/jvm_native_images.md
+++ b/docs/jvm_native_images.md
@@ -1,0 +1,18 @@
+# Native images for the Java Virtual Machine (JVM)
+
+Ahead Of Time (AOT) compilation build tools such as those provided by [GraalVM's `native-image`](https://www.graalvm.org/reference-manual/native-image/) can require some configuration when using protobuf.
+Protobuf for the JVM uses reflection and some of its target classes are not possible to determine in advance.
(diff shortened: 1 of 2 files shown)


huntc:
@googlebot I signed it!

docs/jvm_native_images.md
```diff
@@ -0,0 +1,18 @@
+# Native images for the Java Virtual Machine (JVM)
```
perezd, line 1: nit: add GraalVM to the start of this header to avoid confusion.

huntc: Ah - I was referring to native images in general - not just those of the GraalVM flavour. Right now, Android produces some native-esk (right?), and then there's Scala Native and perhaps others in future.

perezd: OK, lets title it "Ahead of Time (AOT) compilation for the Java Virtual Machine (JVM)". This will represent that intention more generically I think. Also, renaming the doc to match jvm_aot.md or something of that nature?

huntc: Done

perezd: Thank you!

PR #8301, opened 2021-02-16
Fix cmake install on iOS

anton-danielsson:
Fixes:
```
INSTALL TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
```

The usefulness of these binaries on iOS can be discussed but it helps if the build "just works".
In my case this change will help to exclude special handling when using protobuf trough
the Conan package manager.


anton-danielsson:
Ping?

I know the checks are failing but I assume I can't add labels myself? Or am I missing something?

PR #8304, opened 2021-02-17
Add support for custom well-known handled types in JSON

Falco20019:
We need custom well-known types in a project since we need to handle some oneOfs differently based on the type. The current code base already support well-known types (for types well-known to protobuf) but doesn't allow to add custom well-known types. We first thought about using `ICustomDiagnosticMessage`, but this changes the behavior of the `JsonFormatter` and also doesn't allow parsing.

- Cleanup of unused code
- Unification of `JsonFormatter` and `JsonParser` code for well-known type handling
- Support for `AddWellKnownTypeHandlers` in `JsonFormatter` and `JsonParser` to allow custom handling
- Made `JsonTokenizer` and `JsonToken` public as they are needed for `JsonParser` handlers
- Added tests for new code

### DIFF
--- a/csharp/src/Google.Protobuf.Benchmarks/ParseMessagesBenchmark.cs
+++ b/csharp/src/Google.Protobuf.Benchmarks/ParseMessagesBenchmark.cs
@@ -34,7 +34,6 @@ using BenchmarkDotNet.Attributes;
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Buffers;
 using Google.Protobuf.WellKnownTypes;
 using Benchmarks.Proto3;
--- a/csharp/src/Google.Protobuf.Benchmarks/WriteMessagesBenchmark.cs
+++ b/csharp/src/Google.Protobuf.Benchmarks/WriteMessagesBenchmark.cs
@@ -33,9 +33,6 @@
 using BenchmarkDotNet.Attributes;
 using System;
 using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Buffers;
 using Google.Protobuf.WellKnownTypes;
 
 namespace Google.Protobuf.Benchmarks
--- a/csharp/src/Google.Protobuf.Benchmarks/WriteRawPrimitivesBenchmark.cs
+++ b/csharp/src/Google.Protobuf.Benchmarks/WriteRawPrimitivesBenchmark.cs
@@ -32,10 +32,7 @@
 
 using BenchmarkDotNet.Attributes;
 using System;
-using System.Buffers.Binary;
 using System.Collections.Generic;
-using System.IO;
-using System.Buffers;
 using System.Text;
 
 namespace Google.Protobuf.Benchmarks
--- a/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
+++ b/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
@@ -39,8 +39,6 @@ using System.Collections;
 using System.Linq;
 using System.Buffers;
 using System.Runtime.InteropServices;
-using System.Threading;
-using System.Runtime.CompilerServices;
 #if !NET35
 using System.Threading.Tasks;
 #endif
--- a/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
+++ b/csharp/src/Google.Protobuf.Test/Collections/RepeatedFieldTest.cs
@@ -33,7 +33,6 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Text;
--- a/csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
+++ b/csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
@@ -30,7 +30,6 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endregion
 
-using System.Collections.Generic;
 using Google.Protobuf.Collections;
 using Google.Protobuf.TestProtos;
 using NUnit.Framework;
--- a/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
+++ b/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
@@ -33,10 +33,7 @@
 using System;
 using System.IO;
 using Google.Protobuf.TestProtos;
-using Proto2 = Google.Protobuf.TestProtos.Proto2;
 using NUnit.Framework;
-using System.Collections;
-using System.Collections.Generic;
 using System.Linq;
 using Google.Protobuf.WellKnownTypes;
 
--- a/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
@@ -676,6 +676,80 @@ namespace Google.Protobuf
             AssertWriteValue(value, "{ 'FieldName13': 0 }");
         }
 
+        [Test]
+        public void CustomHandler_Message()
+        {
+            var formatter = new JsonFormatter(JsonFormatter.Settings.Default);
+            formatter.AddWellKnownTypeHandlers<ForeignMessage>(ForeignMessageCustomHandler);
+            var message = new ForeignMessage {C = 10};
+
+            AssertJson("10", formatter.Format(message));
+            // Ensure that the formatter has no effect on any other instance:
+            AssertJson("{ 'c': 10 }", JsonFormatter.Default.Format(message));
+        }
+
+        [Test]
+        public void CustomHandler_RepeatedMessage()
+        {
+            var formatter = new JsonFormatter(JsonFormatter.Settings.Default);
+            formatter.AddWellKnownTypeHandlers<ForeignMessage>(ForeignMessageCustomHandler);
+            var message = new TestAllTypes
+                {RepeatedForeignMessage = {new ForeignMessage {C = 10}, new ForeignMessage {C = 20}}};
+
+            AssertJson("{ 'repeatedForeignMessage': [ 10, 20 ] }",
+                formatter.Format(message));
+            // Ensure that the formatter has no effect on any other instance:
+            AssertJson("{ 'repeatedForeignMessage': [ { 'c': 10 }, { 'c': 20 } ] }",
+                JsonFormatter.Default.Format(message));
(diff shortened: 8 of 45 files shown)


dlj-NaN:
Thanks for the suggestion.

While I understand that your change may make perfect sense in context of your particular use, this change is not suitable for the broader Protobuf project. The overarching purpose of the Protobuf project is cross-consistency, so we only support the WKTs that are understood project-wide. Dealing in JSON that can only be parsed by a single client would pretty directly contravene this goal. Unfortunately, this means we will have to decline this PR (at least, the major functionality -- I would be happy to accept most of the non-functional cleanups, though).

> The only current work-around for JsonParser is to copy the complete code of JsonParser, JsonTokenizer and JsonToken into our own code base to add the handlers directly into the code.

I do understand that JSON is attractive as a presentational format, but the Protobuf JSON format takes a fairly constrained view of what is acceptable, rather than treating JSON as user-styleable presentation. If you want to provide special behavior, then defining your own handling does seem like the pragmatic choice. I would slightly warn against forking our JSON handling, though, simply because it sounds like you need something quite different... it sounds like the JSON you're dealing with won't be interoperable with other protobuf users, anyhow.

Falco20019:
> > The only current work-around for JsonParser is to copy the complete code of JsonParser, JsonTokenizer and JsonToken into our own code base to add the handlers directly into the code.
> 
> I do understand that JSON is attractive as a presentational format, but the Protobuf JSON format takes a fairly constrained view of what is acceptable, rather than treating JSON as user-styleable presentation. If you want to provide special behavior, then defining your own handling does seem like the pragmatic choice. I would slightly warn against forking our JSON handling, though, simply because it sounds like you need something quite different... it sounds like the JSON you're dealing with won't be interoperable with other protobuf users, anyhow.

To give you more context, the ground truth for the protocol that we need to work with comes from a JSON Schema and uses anyOf. Since our application internally uses gRPC and ProtoBuf, we don’t want to pass around the JSON data but convert it to protobuf already when receiving. For type safety, we would prefer to have an oneOf where the anyOf is defined. But since the notation is not 100% matching the expected input for the `JsonParser`, this will fail. So this client IS the only one that needs to understand the JSON since it‘s the gateway before the proto will also be processed by other clients (C++ in our case) which is also why we need to have it in proto instead of just passing the JSON around as a string.

So we don‘t intend to have that JSON to be compatible with any other protobuf implementation but quite the other way around. We need protobuf to be compatible to the JSON that already exists. And to achieve that, we would need to either clone the existing formatter and add custom logic for the oneOf handling, or preferably, have the existing protobuf code already offer the possibility to be flexible enough to handle custom JSON.

I understand your point that the default code must be cross-compatible between all implementations. But this should not mean that you must not allow the user of the library the freedom to change the behavior when he knows what he does. My PR does not change anything comparability wise with other clients by default. If now custom handlers are registered, it behaves exactly like before as the unit tests proof.

To sum it up, I don‘t request to have JSON as a user-friendly Interpretation that only one client understands (this is what `ICustomDiagnosticMessage` is for anyway). I just expect protobuf to not block use-cases that seem to have not been thought about (since as it seem, you also thought about protobuf -> JSON -> protobuf as the only use-case of the code base, which does currently block the JSON -> protobuf -> JSON use-case).

Falco20019:
@dlj-NaN Have you had a chance to check the additional context to get a better picture? Because from your response, I felt, that I didn't make the use case clear in the first post and that you had a wrong assumption about why I made this PR.

dlj-NaN:
This is still not a use case we would want to consider for well-known types.

I'm not quite following your description of why you can't parse the JSON. If you can provide an example, then maybe there might be another way to represent it structurally.

Falco20019:
The `instantActions.schema.json` is given and can't be changed by us. It's part of a automatin protocol called VDA5050. `instant_actions.proto` is controllable by us, but needs to have all information from the JSON Schema mapped onto ProtoBuf. `actionParameters` is dependant on `actionType`. This is easily doable with a custom well-known type for `Action` but not with the parser as-is. Since the type is in the middle, we can't just parse the full data structure and have to parse all messages above the oneof ourselves. All below can just be parsed as sub-message using the regular parser logic. That's necessary, because there is no hook-point on `Action` that we could use. If there is a way to tell the parser that there is complex logic (that is expressible in JSON Schema but not in ProtoBuf), then our problem would also be solved.

I understand, that the parser and formatter are only intended for simple cases, that's why I also never asked about complex stuff to be included in the default implementation (especially since this would be a lot of effort to get into all implementations). So the only thing that we need at minimum is a hook-in point to steer the parsing/formatting.


`instantActions.schema.json`
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "instantActions",
  "description": "JSON Schema for publishing instantActions that the AGV is to execute as soon as they arrive.",
  "documentVersion": "0.1.1",
  "subtopic": "/instantActions",
  "type": "object",
  "properties": {
    "header": {
      "$ref": "#/definitions/header"
    },
    "instantActions": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/action"
      }
    }
  },
  "definitions": {
    "header": {
      "$id": "#header",
      "type": "object",
      "properties": {
        "headerId": {
          "type": "integer",
          "description": "headerId of the message. The headerId is defined per topic and incremented by 1 with each sent (but not necessarily received) message."
        },
        "timestamp": {
          "type": "string",
          "format": "date-time",
          "description": "Timestamp in ISO8601 format.",
          "examples": [
            "1991-03-11T11:40:03.12Z"
          ]
        },
        "version": {
          "type": "string",
          "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$",
          "description": "Version of the protocol [Major].[Minor].[Patch]",
          "examples": [
            "1.3.2"
          ]
        },
        "manufacturer": {
          "type": "string",
          "description": "Manufacturer of the AGV"
        },
        "serialNumber": {
          "type": "string",
          "description": "Serial number of the AGV."
        }
      },
      "required": ["headerId", "timestamp", "version", "manufacturer", "serialNumber"]
    },
    "action": {
      "$id": "#action",
      "type": "object",
      "if": {
        "not": {
          "properties": {
            "actionType": {
              "enum": ["noParamAction"]
            }
          }
        }
      },
      "then": {
        "required": ["actionType", "actionId", "blockingType", "actionParameters"]
      },
      "else": {
        "required": ["actionType", "actionId", "blockingType"]
      },
      "properties": {
        "actionType": {
          "type": "string",
          "description": "The name/type of the action (e.g. setFlashingMode)",
          "enum": ["setFlashingMode", "twoParamAction", "noParamAction"]
        },
        "actionId": {
          "type": "string",
          "description": "ID to distinguish between multiple actions."
        },
        "actionDescription": {
          "type": "string",
          "description": "a verbose description of the action"
        },
        "blockingType": {
          "type": "string",
          "description": "Regulates if the action is allowed to be executed during movement and/or parallel to other actions.\nnone: action can happen in parallel with others, including movement.\nsoft: action can happen simultaneously with others, but not while moving.\nhard: no other actions can be performed while this action is running.",
          "enum": ["NONE", "SOFT", "HARD"]
        },
        "actionParameters": {
          "anyOf": [
            {
              "if": {
                "properties": {
                  "actionType": {
                    "const": "setFlashingMode"
                  }
                }
              },
              "then": {
                "$ref": "#/definitions/apSetFlashingMode"
              }
            },
            {
              "if": {
                "properties": {
                  "actionType": {
                    "const": "twoParamAction"
                  }
                }
              },
              "then": {
                "$ref": "#/definitions/apTwoParamAction"
              }
            },
            {
              "if": {
                "properties": {
                  "actionType": {
                    "const": "noParamAction"
                  }
                }
              },
              "then": {
                "$ref": "#/definitions/apNoParamAction"
              }
            }
          ]
        }
      }
    },
    "apNoParamAction": {
      "type": "array",
      "maxItems": 0
    },
    "apSetFlashingMode": {
      "type": "array",
      "items": [
        {
          "type": "object",
          "required": ["key", "value"],
          "properties": {
            "key": {
              "const": "signalPattern"
            },
            "value": {
              "type": "string",
              "description": "The value of the action parameter",
              "enum": ["Off", "WarningFast", "WarningSlow"]
            }
          }
        }
      ]
    },
    "apTwoParamAction": {
      "type": "array",
      "minItems": 2,
      "maxItems": 2,
      "items": [
        {
          "type": "object",
          "required": ["key", "value"],
          "properties": {
            "key": {
              "const": "param1ForTwoParamAction"
            },
            "value": {
              "type": "integer"
            }
          }
        },
        {
          "type": "object",
          "required": ["key", "value"],
          "properties": {
            "key": {
              "const": "param2ForTwoParamAction"
            },
            "value": {
              "type": "string"
            }
          }
        }
      ]
    }
  }
}
```

`instant_actions.proto`
```
syntax = "proto3";

import "jungheinrich/li_v3/vda/vda5050plus/header.proto";

package jungheinrich.li_v3.vda.vda5050plus;

message InstantActions {
    Header header = 1;
    repeated Action instant_actions = 2;
}

message Action {
    string action_id = 1;
    string action_description = 2;
    BlockingType blocking_type = 3;
    oneof action_type {
        ApSetFlashingMode set_flashing_mode = 10;
        ApTwoParamAction two_param_action = 11;
        ApNoParamAction no_param_action = 12;
    }
}

message ApSetFlashingMode {
    enum SignalPattern {
        OFF = 0;
        WARNING_FAST = 1;
        WARNING_SLOW = 2;
    }
	
    SignalPattern signal_pattern = 1;
}

message ApTwoParamAction {
    int32 param1 = 1;
    string param2 = 2;
}

message ApNoParamAction {
}

enum BlockingType {
    NONE = 0;
    SOFT = 1;
    HARD = 2;
}
```

`instantActions.json`
```
{
  "header": {
    "headerId": 847,
    "timestamp": "2020-10-14T12:21:35.124Z",
    "version": "0.2.0",
    "manufacturer": "JH",
    "serialNumber": "0001"
  },
  "instantActions": [
    {
      "actionId": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
      "actionType": "setFlashingMode",
      "actionDescription": "Sets the flashing modes",
      "blockingType": "NONE",
      "actionParameters": [
        {
          "key": "signalPattern",
          "value": "WarningFast"
        }
      ]
    },
    {
      "actionId": "0815",
      "actionType": "twoParamAction",
      "actionDescription": "Action with two parameters",
      "blockingType": "HARD",
      "actionParameters": [
        {
          "key": "param1ForTwoParamAction",
          "value": 42
        },
        {
          "key": "param2ForTwoParamAction",
          "value": "43"
        }
      ]
    },
    {
      "actionId": "0816",
      "actionType": "noParamAction",
      "actionDescription": "Action without parameters",
      "blockingType": "SOFT",
      "actionParameters": []
    }
  ]
}
```

PR #8322, opened 2021-02-20
add lazy support for c++

gongyiling:
1. support [lazy = true] options for c++ message field.
2. fix unit test bug.


gongyiling:
1. support [lazy = true] options for C++ message field.
2. fix unit test bug.

gongyiling:
@googlebot I signed it!

gongyiling:
@googlebot I signed it!

dlj-NaN:
Thanks for sending this PR.

Unfortunately, we already have a conflicting implementation for C++, it just happens to be one of the last pieces we are trying to reconcile from Google's original proto1 API. That hasn't been a high priority for us, especially since C++ handling of lazy fields is quite surprising to most users: there isn't a consistent way to decide whether a lazy field is valid or not without parsing it -- this indeterminate state tends to be surprising to folks.

Unfortunately, even though the reason is probably very unsatisfying, this means I cannot accept this PR. However, I will look into how we might prioritize releasing our lazy field handling publicly. (The longest pole, directly related to validity checking, is handling `required` fields. I expect we would have to address `required` fields on their own before we could provide a sensible C++ implementation of lazy fields.)

PR #8323, opened 2021-02-20
Fix PROTOBUF_CONSTINIT macro redefinition

georgthegreat:
The #ifdef block is duplicated with the above (which was fixed in 3.15.1).

It would be great to have this fix backported to 3.15.1 (and released as 3.15.2).


acozzette:
@georgthegreat Thanks for this fix!

PR #8328, opened 2021-02-22
Move the class map to a CFDictionary.

thomasvl:
Since the keys are `Class`-s, there's no need to hash/copy/etc. them. This
avoids causing `+initialize` on the classes just when building up a registry.


thomasvl:
@dmaclach fyi

dmaclach (approved):

PR #8329, opened 2021-02-22
Merge 3.15.x into master

acozzette:


acozzette:
The CLA is OK because these are just commits that already went through the CLA process.

haberman (approved):

PR #8335, opened 2021-02-23
python max float fails to deserialize and reserialize

jmphilli:
This pull request is primarily to exemplify an issue with floats.
If you have a proto message with the maximum float value you can successfully decode the proto message into a python dictionary.
If you attempt to reencode that value into a new proto message the encoding will fail, raising
`google.protobuf.json_format.ParseError: Failed to parse floatValue field: Float value too large.`

See [issue](https://github.com/protocolbuffers/protobuf/issues/8336)


acozzette:
@jmphilli Thank you for the bug report. I am going to close this pull request since it's not intended to be merged, but of course we can still refer to it to understand the bug.

Issue #8337, opened 2021-02-23
[Ruby] Increase in memory footprint in 3.15


robertlaurin:
**What version of protobuf and what language are you using?**
Version: v3.15.1
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**
macOS Big Sur Version 11.2.1 

**What runtime / compiler are you using (e.g., python version or gcc version)**
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin19]

**What did you do?**

Used a minimal script with [opentelemetry-ruby](https://github.com/open-telemetry/opentelemetry-ruby) to recreate what was observed in production, this script does require a [collector](https://github.com/open-telemetry/opentelemetry-collector) running to demonstrate the issue.
```rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'google-protobuf', ENV['GOOGLE_PROTOBUF_VERSION']
  gem 'heap-profiler'
  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-exporter-otlp'
end

require 'heap-profiler'
require 'opentelemetry-api'
require 'opentelemetry-sdk'
require 'opentelemetry-exporter-otlp'

HeapProfiler.report("heap_output-protobuf-v-#{Gem.loaded_specs['google-protobuf'].version}") do
  span_processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(
    OpenTelemetry::Exporter::OTLP::Exporter.new(
      endpoint: 'http://localhost:55681/v1/trace'
    ),
  )

  OpenTelemetry::SDK.configure do |c|
    c.service_name = 'protobuf-mem-usage'
    c.add_span_processor(span_processor)
  end

  TestTracer = OpenTelemetry.tracer_provider.tracer('protobuf-mem-usage')

  1..10_000.times do
    TestTracer.in_span('parent') do
      1..100.times { TestTracer.in_span('child') {} }
    end
  end
end
```

**What did you expect to see**
Using google-protobuf 3.14
```
heap-profiler heap_output-protobuf-v-3.14.0/
Total allocated: 2.32 GB (22915749 objects)
Total retained: 2.54 MB (17341 objects)
```

```
retained memory by class
-----------------------------------
   1.05 MB  Thread
 399.55 kB  OpenTelemetry::SDK::Trace::Span
 358.10 kB  Time
 203.60 kB  String
[...]

PR #8341, opened 2021-02-25
Ruby <2.7now uses WeakMap too, which prevents memory leaks.

haberman:
Ruby <2.7 does not allow non-finalizable objects to be WeakMap
keys: https://bugs.ruby-lang.org/issues/16035

We work around this by using a secondary map for Ruby <2.7 which
maps the non-finalizable integer to a distinct object.

For now we accept that the entries in the secondary map wil never
be collected.  If this becomes a problem we can perform a GC pass
every so often that looks at the contents of the object cache to
decide what can be deleted from the secondary map.

This also changes the way that object pinning works. We pin objects
when they are frozen so they do not forget their frozen state.
Now we implement this with a simple array in Arena, instead of
making it part of the object cache.

Fixes: #8337

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -295,7 +295,7 @@ static VALUE DescriptorPool_alloc(VALUE klass) {
 
   self->def_to_descriptor = rb_hash_new();
   self->symtab = upb_symtab_new();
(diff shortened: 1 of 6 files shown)


haberman:
Besides not leaking memory any more, this PR also runs the unit tests faster. So I expect improved performance.

gerben-s (approved):

PR #8344, opened 2021-02-25
Fix some constness / char literal issues being found by MSVC standard conforming modea

georgthegreat:
This one is somewhat tricky.

According to C Language Standard `"string"` has type `char[7]` (that is, 6 letters + null terminator).

According to C++ Standard, however, it has its natural type of `const char[7]` (see [cppreference](https://en.cppreference.com/w/cpp/language/string_literal)).

An attempt to write `char* ptr = "string";`, however, does not cause any errors when compiled by clang or gcc.
MSVC has `/permissive-` flag which is off by default, unless `/std:c++latest` is used.

This PR fixes issues caused by this type mismatch.

### DIFF
--- a/python/google/protobuf/pyext/descriptor_pool.cc
+++ b/python/google/protobuf/pyext/descriptor_pool.cc
@@ -176,9 +176,9 @@ static PyDescriptorPool* PyDescriptorPool_NewWithDatabase(
 // The public DescriptorPool constructor.
 static PyObject* New(PyTypeObject* type,
                      PyObject* args, PyObject* kwargs) {
-  static char* kwlist[] = {"descriptor_db", 0};
+  static const char* kwlist[] = {"descriptor_db", 0};
   PyObject* py_database = NULL;
-  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|O", kwlist, &py_database)) {
+  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|O", (char**)kwlist, &py_database)) {
     return NULL;
   }
   DescriptorDatabase* database = NULL;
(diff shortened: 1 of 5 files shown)


python/google/protobuf/pyext/descriptor_pool.cc
```diff
@@ -176,9 +176,9 @@ static PyDescriptorPool* PyDescriptorPool_NewWithDatabase(
 // The public DescriptorPool constructor.
 static PyObject* New(PyTypeObject* type,
                      PyObject* args, PyObject* kwargs) {
-  static char* kwlist[] = {"descriptor_db", 0};
+  static const char* kwlist[] = {"descriptor_db", 0};
   PyObject* py_database = NULL;
-  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|O", kwlist, &py_database)) {
+  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|O", (char**)kwlist, &py_database)) {
```
acozzette, line 181: Thanks for fixing this! But would you mind changing these casts to `const_cast<char**>`? We try to avoid the plain C-style casts in favor of the more specific casts like `const_cast`, `static_cast`, etc.

georgthegreat:
@acozzette, I am fine with adjusting the code according to the style guide, but is there any chance that Google, being a visionary sponsor of PSF, could make something to fix this issue right in the Python itself?

These const_cast's are nasty. Removing them in the future would require adding `#ifdef` checking for current Python version.

acozzette:
@georgthegreat This would definitely be nice to fix within Python itself, but unfortunately I don't think I can help much there because I'm not really connected to anyone would be in a good position to work on that.

georgthegreat (commented):
Ok. Fixing this in python will require changes in many opensource projects anyway. It's better not to block on this.

I have changed to const_cast, as requested.

georgthegreat:
@acozzette, none of the above links claiming that _Kokoro build failed_ work for me.

Could you, please, help me understanding the failure reasons?

acozzette:
Sorry about that, those failures were caused by an unrelated issue that should be fixed now. Let me try running the tests again.

PR #8346, opened 2021-02-25
Make libprotobuf symbols local on OSX

rowillia:
@gnossen gave a great overview in https://github.com/grpc/grpc/pull/24992 of the overall problem.

If a python process using both protobuf _and_ another native library linking in libprotobuf
frequently can cause crashes.  This seems to frequently affect tensorflow as well:

tensorflow/tensorflow#8394,
tensorflow/tensorflow#9525 (comment)
tensorflow/tensorflow#24976,
tensorflow/tensorflow#35573,
https://github.com/tensorflow/tensorflow/blob/v2.0.0/tensorflow/contrib/makefile/rename_protobuf.sh,
tensorflow/tensorflow#16104

This fixes both crashes when linking in multiple versions of protobuf
and fixes `DescriptorPool` clashes as well (e.g. Python and Native code import different versions of the same message).

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -192,6 +192,18 @@ if __name__ == '__main__':
 
     extra_compile_args = []
 
+    message_extra_link_args = None
+    api_implementation_link_args = None
+    if "darwin" in sys.platform:
+      if sys.version_info[0] == 2:
+          message_init_symbol = 'init_message'
+          api_implementation_init_symbol = 'init_api_implementation'
+      else:
+          message_init_symbol = 'PyInit__message'
+          api_implementation_init_symbol = 'PyInit__api_implementation'
+      message_extra_link_args = ['-Wl,-exported_symbol,_%s' % message_init_symbol]
+      api_implementation_link_args = ['-Wl,-exported_symbol,_%s' % api_implementation_init_symbol]
+
     if sys.platform != 'win32':
         extra_compile_args.append('-Wno-write-strings')
         extra_compile_args.append('-Wno-invalid-offsetof')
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -192,6 +192,18 @@ def get_option_from_sys_argv(option_str):
 
     extra_compile_args = []
 
+    message_extra_link_args = None
```
TeBoring, line 195: https://github.com/grpc/grpc/pull/24992/files
Could you add a comment like the one in grpc's PR to explain the reason?

TeBoring:
1) Could you explain a little more about, before this change, how does mixture of different versions of symbols caused the problem? Seems to me, previously, each versions protobuf symbol was local weak symbols, which is just used in the module locally. If so, seems nothing bad would happen.
2) Seems your fix makes the local weak symbol defined by protobuf public. How does that fix the problem?
3) what will happen if another library links protobuf with this fix? And what will happen if another library links protobuf without this fix?

TeBoring:
Reply to my own questions:
1) https://anadoxin.org/blog/control-over-symbol-exports-in-gcc.html/
    This gives a pretty good explanation. By default, all symbols are exposed. If both B1 and B2 statically link A and expose A's
    symbols, on the other hand, C dynamically links A will get different versions of symbols from B1 and B2, which mess things 
    up.
2) This fix restrict exported symbol from all to those specified.
3) Other libraries need to do the same. Otherwise, those other libraries still create conflicts.

TeBoring (approved):

peteboothroyd:
@TeBoring what happened to these changes? I can't see them in the current branch, nor in the git history for the setup.py file.

peteboothroyd:
@TeBoring @rowillia @jtattermusch these changes were [removed here](https://github.com/protocolbuffers/protobuf/commit/636132bd77c4afc890a294b2a751ae7b73f58622), and then some of the changes to `setup.py` were [reverted here](https://github.com/protocolbuffers/protobuf/pull/8746/files). However the changes in this PR were missed. I believe that this is probably still an ongoing issue with the protobuf package, unless some other change has obviated the need for this?

PR #8347, opened 2021-02-25
Create a CMake option to control whether or not RTTI is enabled

### LINKED ISSUE #5541: Add CMake option to define GOOGLE_PROTOBUF_NO_RTTI
C++

I am integrating Protobuf into a C++ project that does not use RTTI. I am building Protobuf from source (with CMake) and I'd like to be able to set the `GOOGLE_PROTOBUF_NO_RTTI` flag when I configure the build.

I would like a CMake option to control whether or not `GOOGLE_PROTOBUF_NO_RTTI` gets defined.

floriansimon1:
This is useful for Conan recipes that build Protobuf, in which
whatever we want to enable has to be enabled in the initial command line.
Without this, the people maintaining the recipe have to patch the CMake
setup of Protobuf before building the binaries.

Closes #5541

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -40,6 +40,7 @@ endif()
 if(WITH_PROTOC)
   set(protobuf_PROTOC_EXE ${WITH_PROTOC} CACHE FILEPATH "Protocol Buffer Compiler executable" FORCE)
 endif()
+option(protobuf_DISABLE_RTTI "Remove runtime type information in the binaries" ON)
 option(protobuf_BUILD_TESTS "Build tests" ON)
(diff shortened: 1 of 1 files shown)


acozzette:
Thanks, @floriansimon1.

floriansimon1:
My pleasure!

Do we know when the next release is planned?

acozzette:
We don't have a set release schedule but we seem to average once every quarter or so. If you want to have this in a release version sooner then feel free to send me another PR cherry-picking this to the 3.15.x branch because we are probably going to do a 3.15.4 release this week.

floriansimon1:
Will do, thanks!

-------- Message d'origine --------

floriansimon1:
@acozzette > There! https://github.com/protocolbuffers/protobuf/pull/8361

PR #8353, opened 2021-02-27
[C++] Delete StringPiecePod

Yannic:
Protobuf no longer supports C++ < 11, so this type is no longer
required.


acozzette:
@Yannic Thank you for the cleanup!

PR #8354, opened 2021-02-28
[C++] Refactor util::Status

Yannic:
This change refactors util::Status to have a similar shape as the
recently open-sourced absl::Status. This will allow Protobuf to
eventually use absl::Status and reduce the codesize.

Note that there is more work required before absl::Status can be used.


src/google/protobuf/stubs/internal/status.cc
```diff
@@ -37,61 +37,64 @@
 namespace google {
 namespace protobuf {
 namespace util {
-namespace error {
```
acozzette, line 40: What's the rationale for moving this to a new `internal` subdirectory?

Yannic: Just stylistic preference. We can keep the implementation in this file if you'd like, with or without namespace `status_internal` (although I'd prefer keeping the namespace to already give a hint that the symbols will come from a different namespace eventually)

Yannic (commented):
Thanks for running presubmit!

I tried looking at the logs but it seems like I don't have permissions.

src/google/protobuf/stubs/internal/status.cc
acozzette: OK, I think it would be good to keep the code in the same files but moving them into a `status_internal` namespace sounds good.

Yannic: Alright, done!

acozzette:
The CI run is showing an error like this:
```
src/google/protobuf/util/internal/json_stream_parser_test.cc:153:38: error: 'class google::protobuf::util::status_internal::Status' has no member named 'error_message'
     StringPiece error_message(result.error_message());
                                      ^
```

acozzette:
Thanks @Yannic, it looks like there are a couple more call sites to fix in the conformance tests:
```
conformance_cpp.cc:137:54: error: no member named 'error_message' in 'google::protobuf::util::status_internal::Status'
                                  std::string(status.error_message()));
                                              ~~~~~~ ^
conformance_cpp.cc:192:32: error: no member named 'error_message' in 'google::protobuf::util::status_internal::Status'
            std::string(status.error_message()));
```

Yannic:
> Thanks @Yannic, it looks like there are a couple more call sites to fix in the conformance tests:
> 
> ```
> conformance_cpp.cc:137:54: error: no member named 'error_message' in 'google::protobuf::util::status_internal::Status'
>                                   std::string(status.error_message()));
>                                               ~~~~~~ ^
> conformance_cpp.cc:192:32: error: no member named 'error_message' in 'google::protobuf::util::status_internal::Status'
>             std::string(status.error_message()));
> ```

Thanks, fixed

PR #8361, opened 2021-03-02
Create a CMake option to control whether or not RTTI is enabled

floriansimon1:
This is useful for Conan recipes that build Protobuf, in which
whatever we want to enable has to be enabled in the initial command line.
Without this, the people maintaining the recipe have to patch the CMake
setup of Protobuf before building the binaries.


dkorchynskyi-axcient:
@floriansimon1 why the RTTI is now disabled by default?

floriansimon1:
This is a mistake. I flipped the logic incompletely. I will submit a PR to address this ASAP

PR #8362, opened 2021-03-02
read_property() handler is not supposed to return NULL

tony2001:
This handler is called throughout the Zend engine code and in none of the cases the result is checked for NULL because the engine expects it to be a valid zval.
If there is no such property, the object has to return &EG(uninitialized_zval).

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -334,7 +334,7 @@ static zval *Message_read_property(PROTO_VAL *obj, PROTO_STR *member,
   Message* intern = PROTO_VAL_P(obj);
   const upb_fielddef *f = get_field(intern, member);
 
-  if (!f) return NULL;
+  if (!f) return &EG(uninitialized_zval);
   Message_get(intern, f, rv);
   return rv;
 }


haberman:
I agree this makes sense to fix. But can I ask how you encountered this? User code should never be accessing message properties directly.

tony2001:
It's not that easy. 
We have some debugging code incorporated into several million lines of code in PHP. This debugging code tries to provide as much info as possible and dumps all the objects with all their properties (I think. It's a lot of code and I'm not really familiar with it).

I believe that's actually the problem with protobuf extension - always assumes that user should not be doing this or that. There are no safety checks, it just crashes if user does something it was not designed to do (like instantiating an internal class manually). 
Unfortunately, there's no way to prevent that - users (i.e. PHP programmers in this case) sometimes do strange things, sometimes their code is used not the way it was initially designed to. And they expect to get errors, not crashes.

haberman:
> There are no safety checks, it just crashes if user does something it was not designed to do (like instantiating an internal class manually).

I agree we should enforce these rules whenever possible. For example there is https://github.com/protocolbuffers/protobuf/pull/8002 which might have been able to prevent this.

PR #8364, opened 2021-03-02
[Ruby] Fixed quadratic memory usage when appending to arrays.

haberman:
The code mistakenly called realloc() instead of resize() on every
array append, causing quadratic memory usage.

This fixes the problem reported in https://github.com/protocolbuffers/protobuf/issues/8337#issuecomment-789119303


acozzette (approved):

hatstand:
@haberman @acozzette I don't think this made it into 3.15.4 AFAICT and it's mistakenly in the release notes there. https://github.com/protocolbuffers/protobuf/blob/v3.15.4/ruby/ext/google/protobuf_c/ruby-upb.c#L6666

haberman:
@hatstand thanks for the heads-up, I mistakenly merged it to the wrong branch.

Here is a follow-up that will go out in 3.15.5: https://github.com/protocolbuffers/protobuf/pull/8379

PR #8375, opened 2021-03-04
Fix a bug that would disable RTTI in the CMake scripts by default

floriansimon1:
This is supposed to be an option, not something we use in general


floriansimon1:

There's already a version that is mergeable to [3.15.x](https://github.com/protocolbuffers/protobuf/pull/8375).

PR #8377, opened 2021-03-04
Do not disable RTTI by default in the CMake build

floriansimon1:
This is supposed to be optional, not enabled by default.


floriansimon1:
There's also a version that is mergeable to the [master branch](https://github.com/protocolbuffers/protobuf/pull/8378).

dkorchynskyi-axcient (approved):
👍

PR #8378, opened 2021-03-04
Do not disable RTTI by default in the CMake build

floriansimon1:
This is supposed to be optional, not enabled by default.


floriansimon1:
There's also a version that is mergeable to the [3.15.x branch](https://github.com/protocolbuffers/protobuf/pull/8377).

acozzette:
Thanks for the quick fix, @floriansimon1. I think I will probably just merge the other one to the 3.15.x branch and merge 3.15.x into master later today, but let me keep this pull request open for now just in case.

PR #8387, opened 2021-03-06
Correcting arch classifier for s390x

### LINKED ISSUE #8340: s390x protoc artifact on Maven Central uses the wrong classifier
protoc artifacts on Maven Central should be using names agreed upon with the os-detector plugin so that the correct binary can be automatically downloaded when available by the Maven and Gradle protoc plugins. The current classifier "s390x" disagrees with os-detector's "s390_64" so can't be found by the plugins. See https://github.com/google/protobuf-gradle-plugin/issues/472

It's not clear how the current s390x classifier name was chosen. I see https://github.com/protocolbuffers/protobuf/pull/7073 explicitly remove the _64 suffix, but there's no background or reasoning provided. I see they were added in https://github.com/protocolbuffers/protobuf/pull/6558, but I didn't see discussion about the classifier name. I have to assume that people just weren't aware that the names were important and not arbitrary.

CC @salamaniibm, @Hendrik-H, @voidzcy

salamani:
Fixes #8340

### DIFF
--- a/kokoro/release/protoc/linux/build.sh
+++ b/kokoro/release/protoc/linux/build.sh
@@ -23,7 +23,7 @@ sudo apt install -y g++-powerpc64le-linux-gnu
 protoc-artifacts/build-protoc.sh linux ppcle_64 protoc
 
 sudo apt install -y g++-s390x-linux-gnu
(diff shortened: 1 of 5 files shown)


acozzette:
Thanks, @salamani.

PR #8391, opened 2021-03-08
run basic C++ aarch64 tests under qemu emulator

jtattermusch:
Add a script that crosscompiles C++ under a crosscompiler  (and uses dockcross docker image for that) and then runs C++ test suite under QEMU emulator.

- the tests pass and the the whole build + tests only takes a few minutes (= faster than if we were compiled everything under and emulator, which would be quite slow).
- for now there's no kokoro job to actually run the tests, but setting up one is easy (I plan to do that as a followup PR)
- dockcross docker image is used since it makes everything much easier (see comments in the code)
- for manual run, just run `kokoro/linux/aarch64/test_cpp_aarch64.sh`

What the kokoro test output would look like if the kokoro job was already created (I used a hack to obtain an adhoc run):
sponge/988880f9-7e8f-465b-8ff9-9483d6ce2f0d

### DIFF
--- a/kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
+++ b/kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+set -e
+
+# go to the repo root
+cd $(dirname $0)/../../../..
+
+if [[ -t 0 ]]; then
+  DOCKER_TTY_ARGS="-it"
+else
+  # The input device on kokoro is not a TTY, so -it does not work.
+  DOCKER_TTY_ARGS=
+fi
+
+# running dockcross image without any arguments generates a wrapper
+# scripts that can be used to run commands under the dockcross image
+# easily.
+# See https://github.com/dockcross/dockcross#usage for details
+docker run $DOCKER_TTY_ARGS --rm dockcross/linux-arm64 >dockcross-linux-arm64.sh
+chmod +x dockcross-linux-arm64.sh
+
+# the wrapper script has CRLF line endings and bash doesn't like that
+# so we change CRLF line endings into LF.
+sed -i 's/\r//g' dockcross-linux-arm64.sh
+
+# The dockcross wrapper script runs arbitrary commands under the selected dockcross
+# image with the following properties which make its use very convenient:
+# * the current working directory is mounted under /work so the container can easily
+#   access the current workspace
+# * the processes in the container run under the same UID and GID as the host process so unlike
+#   vanilla "docker run" invocations, the workspace doesn't get polluted with files
+#   owned by root.
+./dockcross-linux-arm64.sh "$@"
--- a/kokoro/linux/cpp_aarch64/continuous.cfg
+++ b/kokoro/linux/cpp_aarch64/continuous.cfg
@@ -1,0 +1,5 @@
+# Config file for running tests in Kokoro
+
+# Location of the build script in repository
+build_file: "protobuf/kokoro/linux/cpp_aarch64/build.sh"
+timeout_mins: 120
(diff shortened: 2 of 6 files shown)


jtattermusch:
CC @janaknat

jtattermusch:
@dlj-NaN  can you please reassign as needed?

jtattermusch:
@dlj-NaN  I verified that the tests are passing if run on kokoro CI. Please review.

jtattermusch:
@fowles  please also review this PR.

kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
```diff
@@ -0,0 +1,33 @@
...
+# running dockcross image without any arguments generates a wrapper
+# scripts that can be used to run commands under the dockcross image
+# easily.
+# See https://github.com/dockcross/dockcross#usage for details
+docker run $DOCKER_TTY_ARGS --rm dockcross/linux-arm64 >dockcross-linux-arm64.sh
+chmod +x dockcross-linux-arm64.sh
+
+# the wrapper script has CRLF line endings and bash doesn't like that
+# so we change CRLF line endings into LF.
+sed -i 's/\r//g' dockcross-linux-arm64.sh
```
fowles, line 24: does windows require this, or can we buy the gnome and fix the wrapper script?

kokoro/linux/cpp_aarch64/continuous.cfg
```diff
@@ -0,0 +1,5 @@
+# Config file for running tests in Kokoro
+
+# Location of the build script in repository
+build_file: "protobuf/kokoro/linux/cpp_aarch64/build.sh"
+timeout_mins: 120
```
fowles, line 5: 2 hours is a really long time... Can we tighten this?

kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
jtattermusch: the "dockcross-linux-arm64.sh" is distributed as part of the dockcross/linux-arm64 image, so "fixing" the wrapper script would required opening a pull request against https://github.com/dockcross/dockcross, getting it accepted (maybe it does, maybe it doesn't) and then waiting for a new release of dockcross/linux-arm64  to actually get a fix. That seems as a way to lengthy process (and with uncertain outcome) to be able to rely on it.
Let's keep the sed command as is. I'll look into filing an issue with https://github.com/dockcross/dockcross so that hopefully this will get fixed in the future. But I really don't want to delay this PR because of it.

kokoro/linux/cpp_aarch64/continuous.cfg
jtattermusch: Actually, 120 is the lowest value that protobuf's kokoro .cfg files use (and it's used by many jobs including python, ruby etc.) and it's the most commonly used timeout value, so I simply reused it. (Btw, for some jobs, the timeout value is even much higher)

I can reduce this to e.g. 60 (in reality it takes only about 10mins, but it's useful to have some buffer) if you want, but basically I feel this should be done in a separate effort to set all the protobuf's kokoro jobs to a reasonable timeout (which is unrelated to this PR).

kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
jtattermusch: the "dockcross-linux-arm64.sh" is distributed as part of the dockcross/linux-arm64 image, so "fixing" the wrapper script would required opening a pull request against https://github.com/dockcross/dockcross, getting it accepted (maybe it does, maybe it doesn't) and then waiting for a new release of dockcross/linux-arm64  to actually get a fix. That seems as a way to lengthy process (and with uncertain outcome) to be able to rely on it.
Let's keep the sed command as is. I'll look into filing an issue with https://github.com/dockcross/dockcross so that hopefully this will get fixed in the future. But I really don't want to delay this PR because of it.

jtattermusch (commented):
Responded to all the concerns.

fowles (approved):

PR #8397, opened 2021-03-10
Allocate extension 1108 to Kratos API Errors

tonybase:


tonybase:
@googlebot I signed it!

PR #8408, opened 2021-03-16
Fix -Wsuggest-destructor-override for generated C++ proto classes.

PatriosTheGreat:
Since proto objects inherits Message class with virtual destructor, they need to override it.

This warning was introduced by clang-11 compiler.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -1092,7 +1092,7 @@ void MessageGenerator::GenerateClassDefinition(io::Printer* printer) {
 
   format(
       "inline $classname$() : $classname$(nullptr) {}\n"
-      "virtual ~$classname$();\n"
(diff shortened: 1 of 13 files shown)


src/google/protobuf/compiler/cpp/cpp_message.cc
```diff
@@ -1092,7 +1092,7 @@ void MessageGenerator::GenerateClassDefinition(io::Printer* printer) {
 
   format(
       "inline $classname$() : $classname$(nullptr) {}\n"
-      "virtual ~$classname$();\n"
+      "virtual ~$classname$() override;\n"
```
acozzette, line 1095: I think as long as we're adding `override` we can drop the `virtual` keyword since that is implied. One other thing is would you mind running `./generate_descriptor_proto.sh`? That will rebuild the checked-in pb.h and pb.cc files.

acozzette:
Thanks, @PatriosTheGreat.

PR #8409, opened 2021-03-17
add error returns missing from protoc to prevent it from exiting with…

srago:
… a successful return value when writing the zip file fails


acozzette:
Thanks, @srago.

srago:
Thanks for the prompt merge.

PR #8411, opened 2021-03-18
Fix detection of existing Uint8Array

### LINKED ISSUE #1319: JS: Type not convertible to Uint8Array
Hi!

I've been trying to get protobuf deserialization working in JavaScript but hitting a wall. I have a small sample message just to test everything is working:

```
message Sample {
    string text = 1;
    uint32 number = 2;
    bool truthiness = 3;
}
```

When I create a new Sample message (without deserializing it from anything), the message works fine.
Serializing it to bytes yields me a 24 byte `Uint8Array`.

However, if I try to create a Sample message from existing bytes, I get an assertion error: `Type not convertible to Uint8Array`. This is what I've tried:
```
var Sample = // loaded in from require;
var dataBytes = // some existing uint8array;
var sample = new Sample(dataBytes);
var sample2 = Sample.deserializeBytes(dataBytes);
```
The first one (`sample`) will silently fail, but the message is broken from there out (`sample.getText()` returns `10` instead of a string, as expected.. other fields return garbage as well).

The second one throws the assertion error. I've verified that the `Uint8Array` I'm receiving and testing with is the same one I get from manually building one via `serializeBytes` in the console.

Here's my data:
screen shot 2016-03-13 at 4 28 28 am

Am I perhaps just using this wrong?
Thanks in advance!

beeman:
It looks like [this patch](https://github.com/protocolbuffers/protobuf/issues/1319#issuecomment-440120601) fixes #1319, a 5 year old issue I just ran into. It would be awesome if it could get merged!

### DIFF
--- a/js/binary/utils.js
+++ b/js/binary/utils.js
@@ -1038,7 +1038,7 @@ jspb.utils.stringToByteArray = function(str) {
  * @suppress {invalidCasts}
  */
 jspb.utils.byteSourceToUint8Array = function(data) {
-  if (data.constructor === Uint8Array) {
+  if (data.constructor === Uint8Array || data instanceof Uint8Array) {
     return /** @type {!Uint8Array} */(data);
   }
 
--- a/js/binary/utils_test.js
+++ b/js/binary/utils_test.js
@@ -715,6 +715,7 @@ describe('binaryUtilsTest', function() {
     var sourceBuffer = sourceBytes.buffer;
     var sourceBase64 = goog.crypt.base64.encodeByteArray(sourceData);
     var sourceString = goog.crypt.byteArrayToString(sourceData);
+    var sourceUint8Array = new Uint8Array(sourceData)
 
     function check(result) {
       expect(result.constructor).toEqual(Uint8Array);
@@ -735,5 +736,8 @@ describe('binaryUtilsTest', function() {
 
     // Converting base64-encoded strings into Uint8Arrays should work.
     check(convert(sourceBase64));
+
+    // Existing Uint8Array into Uint8Arrays should work.
+    check(convert(sourceUint8Array));
   });
 });


js/binary/utils.js
```diff
@@ -1038,7 +1038,7 @@ jspb.utils.stringToByteArray = function(str) {
  * @suppress {invalidCasts}
  */
 jspb.utils.byteSourceToUint8Array = function(data) {
-  if (data.constructor === Uint8Array) {
+  if (data.constructor === Uint8Array || data instanceof Uint8Array) {
```
lukesandberg, line 1041: thanks!

instanceof operators are somewhat substantially more expensive than these direct constructor tests.  so it would be best to only use this if we have to

also, this is a bit redundant with the Buffer test on line 1058, in fact i think Buffer will match here now instead of line 1050

finally, due to Buffer now matching up here we have a slight behavior change

I think the best solution would be to add something like

```
 if (data instanceof Uint8Array) {
    // Support types like nodejs Buffer (a subclass of Uint8Array).
    data = /** @type {!Uint8Array} */ (data);
    // Make a shallow copy to ensure we only ever deals with Uint8Array
    // exactly to ensure monomorphism.
    return /** @type {!Uint8Array} */ (
        new Uint8Array(data.buffer, data.byteOffset, data.byteLength));
  }
```

at the end of the method, so we only execute instanceof if we were going to otherwise fail

beeman: @lukesandberg thanks a lot for the review and your comments. Makes a lot of sense.

I will update the PR and push it up again.

I'm excited to see that this fix is being considered!

lukesandberg: one thing im not sure i made clear, Given this change, i think we should drop the special case for Buffer and let it be handled by the instanceof test.

Curious what you think about that?

js/binary/utils_test.js
```diff
@@ -715,6 +715,7 @@ describe('binaryUtilsTest', function() {
     var sourceBuffer = sourceBytes.buffer;
     var sourceBase64 = goog.crypt.base64.encodeByteArray(sourceData);
     var sourceString = goog.crypt.byteArrayToString(sourceData);
+    var sourceUint8Array = new Uint8Array(sourceData)
```
jkrems, line 718: nit: Is this actually testing something new? We already test `sourceBytes` which looks to be identical (in other words: I assume this test change would pass even without the implementation change above).

It should be enough to use a subclass here, e.g. `new (class extends Uint8Array {})(sourceData)`.

nickwelp:
This bug still exists

acozzette:

beeman:
> We are closing all open JavaScript pull requests because the JavaScript implementation is moving to a new repository: https://github.com/protocolbuffers/protobuf-javascript Sorry for the inconvenience but please feel free to reopen the pull request in the new repo.

Thanks for the heads-up. 

I'm moving away from using protobuf altogether so I won't be needing this anymore - in case anyone else want's to fix it. :)

PR #8416, opened 2021-03-20
[C++] Refactor uint128

Yannic:
This makes it compatible with absl::uint128

### DIFF
--- a/src/google/protobuf/stubs/int128.h
+++ b/src/google/protobuf/stubs/int128.h
@@ -33,37 +33,32 @@
 #include <google/protobuf/stubs/common.h>
 
 #include <iosfwd>
+#include <limits>
+#include <string>
 
 #include <google/protobuf/port_def.inc>
 
 namespace google {
 namespace protobuf {
-
-struct uint128_pod;
-
-// TODO(xiaofeng): Define GOOGLE_PROTOBUF_HAS_CONSTEXPR when constexpr is
-// available.
-#ifdef GOOGLE_PROTOBUF_HAS_CONSTEXPR
-# define UINT128_CONSTEXPR constexpr
-#else
-# define UINT128_CONSTEXPR
-#endif
+namespace int128_internal {
 
 // An unsigned 128-bit integer type. Thread-compatible.
 class PROTOBUF_EXPORT uint128 {
  public:
-  UINT128_CONSTEXPR uint128();  // Sets to 0, but don't trust on this behavior.
-  UINT128_CONSTEXPR uint128(uint64 top, uint64 bottom);
+  uint128() = default;
+
+ private:
+  // Use `MakeUint128` instead.
+  constexpr uint128(uint64 top, uint64 bottom);
+
+ public:
 #ifndef SWIG
-  UINT128_CONSTEXPR uint128(int bottom);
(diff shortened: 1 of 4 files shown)


src/google/protobuf/stubs/int128.h
```diff
@@ -379,6 +357,17 @@ inline uint128& uint128::operator--() {
...
 }
 
+constexpr uint128 Uint128Max() {
+  return MakeUint128((std::numeric_limits<uint64>::max)(),
+                     (std::numeric_limits<uint64>::max)());
+}
+
+}  // namespace int128_internal
+
+using absl::uint128;
```
acozzette, line 367: I think these lines must have been added by mistake?

Yannic: I added these when testing compatibility with absl::uint128 and forgot to remove them. Sorry, fixed

src/google/protobuf/stubs/int128_unittest.cc
```diff
@@ -44,16 +44,18 @@ namespace protobuf {
 
 TEST(Int128, AllTests) {
```
acozzette, line 45: The test run is showing some errors like this:
```
google/protobuf/stubs/int128_unittest.cc: In member function ‘virtual void google::protobuf::Int128_Traits_Test::TestBody()’:
google/protobuf/stubs/int128_unittest.cc:370:15: error: ‘is_trivially_copy_constructible’ is not a member of ‘std’
   EXPECT_TRUE(std::is_trivially_copy_constructible<uint128>::value);
```
I think you might just need to include `<type_traits>` at the top of this file.

Yannic: Thanks, done.

acozzette: I'm still seeing the same error but I think it's because GCC 4.x apparently doesn't support `std::is_trivially_copy_constructible` or `std::is_trivially_copy_assignable`. Would you mind disabling this test case for GCC versions before 5.0? I think something like this should work:
```
#if !defined(__GNUC__) || __GNUC__ > 4
...
#endif
```

Yannic: done

acozzette:
Thanks, @Yannic.

lipnitsk:
This broke external code depending on `descriptor.pb.h`.
```
Step #4: In file included from /src/protobuf-install/include/google/protobuf/descriptor.pb.h:23:
Step #4: In file included from /src/protobuf-install/include/google/protobuf/io/coded_stream.h:148:
Step #4: /src/protobuf-install/include/google/protobuf/stubs/logging.h:34:10: fatal error: 'google/protobuf/stubs/int128.h' file not found
Step #4: #include <google/protobuf/stubs/int128.h>
Step #4:          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Link: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32763
Link: https://oss-fuzz-build-logs.storage.googleapis.com/log-b9d0c99a-00b5-4af7-862c-f16a5a11c457.txt

acozzette:
@lipnitsk Thanks for the heads-up, let's roll back this change until we have a chance to come up with a fix. I will sent out a pull request in just a second.

Issue #8421, opened 2021-03-22
Memory Leak in Ruby gem as of 3.15.x


splittingred:
**What version of protobuf and what language are you using?**
Version: 3.15.6
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**

MacOS Catalina, Linux (Debian Buster).

**What runtime / compiler are you using (e.g., python version or gcc version)**

N/A

**What did you do?**

We have noticed a memory leak in the google-protobuf gem as of 3.15, when used in a gRPC server. Rolling back to 3.14 fixes the issue. See an example graph:

![Screen Shot 2021-03-22 at 5 11 58 PM](https://user-images.githubusercontent.com/233334/112065249-57a52b00-8b32-11eb-8e61-231c59b427a4.png)

The 2nd section there shows when google-protobuf 3.15 was deployed; there is a small section after showing normalization of memory after reverting back to 3.14. 

In order to eliminate potential other libraries, I wrote a quick benchmarking test doing a small client->server unary call. This runs a benchmark using [memory-benchmark](https://github.com/michaelherold/benchmark-memory) to run against Protobuf 3.14 and Protobuf 3.15. The results showed a significant number of objects being retained (and increased memory usage) when run with 1,000 requests on protobuf 3.15, but no such leakage on protobuf 3.14:

```
$ ./run.sh
Running Protobuf 3.14 test
------------------
Installing gems...
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Using memory_profiler 0.9.14
Using benchmark-memory 0.1.2
Using bundler 1.17.3
Using google-protobuf 3.14.0 (universal-darwin)
Using googleapis-common-protos-types 1.0.6
Using grpc 1.36.0 (universal-darwin)
Bundle updated!
Starting grpc server...
Letting server boot...
Beginning benchmark...
Calculating -------------------------------------
              client     3.976M memsize (     0.000  retained)
                        30.000k objects (     0.000  retained)
                         4.000  strings (     0.000  retained)
Calculating -------------------------------------
              client     3.976M memsize (     0.000  retained)
                        30.000k objects (     0.000  retained)
                         4.000  strings (     0.000  retained)
Running Protobuf 3.15 test
------------------
Installing gems...
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Using memory_profiler 0.9.14
Using benchmark-memory 0.1.2
Using bundler 1.17.3
[...]

PR #8429, opened 2021-03-25
Fix unbounded memory growth for Ruby <2.7.

haberman:
Previously our secondary map (needed only on Ruby <2.7) was growing without bound.

This PR adds a periodic GC pass of the secondary map to avoid this unbounded memory growth.

Fixes: https://github.com/protocolbuffers/protobuf/issues/8421

### DIFF
--- a/ruby/ext/google/protobuf_c/protobuf.c
+++ b/ruby/ext/google/protobuf_c/protobuf.c
@@ -251,14 +251,36 @@ void Arena_register(VALUE module) {
 // The object is used only for its identity; it does not contain any data.
 VALUE secondary_map = Qnil;
 
+// Lambda that will GC entries from the secondary map that are no longer present
+// in the primary map.
+VALUE gc_secondary_map = Qnil;
+
+extern VALUE weak_obj_cache;
+
 static void SecondaryMap_Init() {
   rb_gc_register_address(&secondary_map);
+  rb_gc_register_address(&gc_secondary_map);
   secondary_map = rb_hash_new();
+  gc_secondary_map = rb_eval_string(
+      "->(secondary, weak) {\n"
+      "  secondary.delete_if { |k, v| !weak.key?(v) }\n"
+      "}\n");
+}
+
+static void SecondaryMap_MaybeGC() {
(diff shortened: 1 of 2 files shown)


ruby/ext/google/protobuf_c/protobuf.c
```diff
@@ -251,16 +251,65 @@ void Arena_register(VALUE module) {
...
+ID length;
+
+extern VALUE weak_obj_cache;
+
 static void SecondaryMap_Init() {
   rb_gc_register_address(&secondary_map);
+  rb_gc_register_address(&gc_secondary_map);
+  rb_gc_register_address(&secondary_map_mutex);
   secondary_map = rb_hash_new();
+  gc_secondary_map = rb_eval_string(
```
fowles, line 272: I didn't realize this was a functor that did the gc, I thought it was a map.  Can we name it something more evocative:
gc_secondary_map_functor or something?

ruby/ext/google/protobuf_c/protobuf.c
```diff
@@ -251,16 +251,65 @@ void Arena_register(VALUE module) {
...
+// The main object cache is a WeakMap that will automatically remove entries
+// when the target object is no longer reachable, but unless we manually
+// remove the corresponding entries from the secondary map, it will grow
+// without bound.
+//
+// To avoid this unbounded growth we periodically remove entries from the
+// secondary map that are no longer present in the WeakMap. The logic of
+// how often to perform this GC is an artbirary tuning parameter that
+// represents a straightforward CPU/memory tradeoff.
+static void SecondaryMap_MaybeGC() {
```
fowles, line 290: I believe that one must hold the mutex to call this function.  Can you add that as a comment and some kind of assertion as well?

haberman: Done.

ruby/ext/google/protobuf_c/protobuf.c
haberman: Done.

fowles (approved):

PR #8432, opened 2021-03-29
Construct CodedInputStream with ReadOnlyMemory<byte>

thinkbeforecoding:
This is a proposition to use ReadOnlyMemory<byte> inside CodedInputStream. This is useful to parse buffers returned as ReadOnlyMemory<byte> from some APIs without copying back to a byte array.
The rest of the code is already using Span<byte> so the change is of low impact.

When a stream is provided, a byte[] is still necessary to read the stream, and is stored in the internalBuffer field. When given directly a byte[], it is also stored in this field. In this case, the buffer field is just the .AsMemory() of this field.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -895,6 +895,13 @@ namespace Google.Protobuf
         }
 
         [Test]
+        public void Dispose_FromReadOnlyMemory()
+        {
+            var stream = new CodedInputStream(new byte[10].AsMemory());
+            stream.Dispose();
+        }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


thinkbeforecoding:
@googlebot I signed it!

thinkbeforecoding:
The tags should be:
include release notes: yes
c#

JamesNK (commented):
Seems ok, but I wonder if it is better to encourage people to use the new API which will be more efficent than CodedInputStream.

Also, if this change is being made then should CodedOutputStream support writing to `Memory<byte>`?

kiwidev:
If you are doing this, should you also change `ByteString.CreateCodedInput` (which is converting from `ReadOnlyMemory<byte>` to `byte[]`)?

Issue #3688, opened 2017-09-26
Migrate protobuf to use Abseil: Google C++ common libraries.


xfxyjwf:
For a long time protobuf has been keeping its own fork of Google C++ common libraries under the stubs directory:
https://github.com/google/protobuf/tree/master/src/google/protobuf/stubs

They have to be forked because Google never released these common libraries as an opensource project, util now:
https://opensource.googleblog.com/2017/09/introducing-abseil-new-common-libraries.html

With the release of Abseil, we should be able to deprecate code in the stubs directory and migrate protobuf to use Abseil instead.

mcy:
This is already fixed in trunk, and will be part of our next major release!

PR #8438, opened 2021-03-30
[bazel] Add option to depend on Abseil

Yannic:
For now, this is experimental and only used to ensure we don't regress
the parts that we already made compatible with the Abseil types.

Progress on #3688

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -14,6 +14,19 @@ exports_files(["LICENSE"])
 # build configuration
 ################################################################################
 
+string_flag(
+    name = "with_abseil",
+    build_setting_default = "false",
+    values = ["true", "false"]
+)
+
+config_setting(
+    name = "use_abseil",
+    flag_values = {
+        "//:with_abseil": "true"
+    },
+)
+
 ################################################################################
 # ZLIB configuration
 ################################################################################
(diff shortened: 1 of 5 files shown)


dlj-NaN:
Hey, Yannic... thanks for sending this PR. We definitely have Abseil in our sights. In fact, our internal code has always used the "true" implementations instead of the the port library implementations, and at this point, it's mostly Abseil.

Unfortunately, this brings me to the bad news: that conversion process is a very real technical debt, but it has to keep working. This means that depending upon Abseil will require some careful coordination with our internal tooling, which is really the worst case I can imagine: we need to carefully change something super complex that you can't even see, just to make a change that seems obvious without that context. To set a realistic timeframe, I would expect to be able to commit the team's time to this in Q3 (July or later), but this is just a forecast.

So, in the interest of unblocking your needs, what is the nearest-term need?

TeBoring:
@Yannic it seems this PR is one of a large process. Did you have some doc mentioning what's the overall plan?

acozzette:
I'm going to go ahead and close this pull request for now. We appreciate the change but we're not yet ready to depend on Abseil.

Issue #8455, opened 2021-04-06
Memory Leak in ruby gem version 3.15.3 and above (including the latest 3.15.7)


bruno-:
**What version of protobuf and what language are you using?**
Version: 3.15.7
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**
Tried this on both Mac and Linux. The issue can be reproduced on both operating systems.

- Mac: Catalina 10.15.7 (19H2)
- Linux: Linux ubuntu2004.localdomain 5.4.0-65-generic #73-Ubuntu SMP Mon Jan 18 17:25:17 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**

Ruby 3.0.0.

Specific versions:
- Mac version: `ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin19]`
- Linux version: `ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-linux]`

**What did you do?**

Use this gemfile:
```ruby
# filename: gems.rb
source "https://rubygems.org"

gem "google-protobuf", "= 3.15.7"
```

Use this script:

```ruby
# filename: test.rb
require "google/protobuf/well_known_types"

loop do
  Google::Protobuf::Value.new
end
```

- Install the `google-protobuf` gem with: `bundle`
- Run the script with: `bundle exec ruby test.rb`
- Monitor memory usage in `htop` in another terminal window.

**What did you see?**

I can observe that program memory usage grows really fast. I reaches 1Gb within a minute and continues to grow fast.

**What did you expect to see**

I expected the program's memory growth to stop at around 50Mb of memory use.

This happens (memory stops growing) with gem versions `3.15.2` and below.

**Anything else we should know about your project / environment**

- I originally reported this to `grpc` repo (link to issue https://github.com/grpc/grpc/issues/25877) but then realized that is probably not the right place.
- This problem seems to be very similar to #8455 . There's a comment in that issue claiming the problem was fixed in latest ruby gem version `3.15.7`. Unfortunately I was not able to confirm that: the problem described here does happen for me with `3.15.7`.
- I tested various gem versions. This issue does **not** happen with `google-protobuf` gem versions `3.15.2` and below.
- I'm responsive, please ping me and I can provide additional help/clarification etc.

MuhammetDilmac:
When will take new release? @haberman

bruno-:
[...]

PR #8461, opened 2021-04-07
Fixed memory leak of Ruby arena objects.

haberman:
In our `Arena_free()` function, we were freeing the memory from the
upb arena but we were failing to free the memory for the
Ruby arena object. This was causing every Ruby arena object
to leak: even though the objects were getting GC'd, the
underlying memory was not getting released.

Fixes: https://github.com/protocolbuffers/protobuf/issues/8455 https://github.com/protocolbuffers/protobuf/issues/8421


acozzette (approved):

Visva92 (approved):
.

PR #8469, opened 2021-04-08
Use a newer version of rules_proto, with the new rule `proto_descriptor_set`

dorfire:


dorfire:
@googlebot I signed it!

acozzette:
Thanks, @dorfire.

Issue #8457, opened 2021-04-07
PHP Segmentation Fault by calling serializeToJsonString()


Zarathustra101:
**What version of protobuf and what language are you using?**
Version: v3.15.7 (and earlier)
Language: PHP

**What operating system (Linux, Windows, ...) and version?**
Ubuntu 20.04.2 LTS or Debian GNU/Linux 10 (buster)

**What runtime / compiler are you using (e.g., python version or gcc version)**
PHP 7.3.27

**What did you do?**
Write the following to _Example.proto_:
```
syntax = "proto3";

package MyNamespace;

message TestCaseMessage {
  SubMessage subMessage = 1;
}

message SubMessage {
  string something = 1;
}
```

Compile with protoc: 
`protoc --proto_path="$SRC" --php_out="$DESTINATION" Example.proto`

auto generated PHP File:
```
<?php
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: Example.proto

namespace MyNamespace;

use Google\Protobuf\Internal\GPBType;
use Google\Protobuf\Internal\RepeatedField;
use Google\Protobuf\Internal\GPBUtil;

/**
 * Generated from protobuf message <code>MyNamespace.TestCaseMessage</code>
 */
class TestCaseMessage extends \Google\Protobuf\Internal\Message
{
    /**
     * Generated from protobuf field <code>.MyNamespace.SubMessage subMessage = 1;</code>
     */
    protected $subMessage = null;

    /**
     * Constructor.
     *
     * @param array $data {
     *     Optional. Data for populating the Message object.
     *
     *     @type \MyNamespace\SubMessage $subMessage
     * }
     */
    public function __construct($data = NULL) {
        \GPBMetadata\Example::initOnce();
        parent::__construct($data);
    }

    /**
     * Generated from protobuf field <code>.MyNamespace.SubMessage subMessage = 1;</code>
     * @return \MyNamespace\SubMessage|null
     */
    public function getSubMessage()
    {
        return isset($this->subMessage) ? $this->subMessage : null;
    }

    public function hasSubMessage()
    {
        return isset($this->subMessage);
    }

    public function clearSubMessage()
    {
        unset($this->subMessage);
    }

    /**
     * Generated from protobuf field <code>.MyNamespace.SubMessage subMessage = 1;</code>
     * @param \MyNamespace\SubMessage $var
     * @return $this
     */
    public function setSubMessage($var)
    {
[...]

PR #8472, opened 2021-04-09
[PHP] Fixed $msg->setMessage(null) to properly clear the message.

haberman:
Fixes: https://github.com/protocolbuffers/protobuf/issues/8457


TeBoring:
Shall we add a test for ```set_msg(null)```?

haberman:
> Shall we add a test for set_msg(null)?

`testJsonEncodeNullSubMessage()` is doing this.

This change is also causing an exception to be thrown if a user does `$repeated_field[] = null`. I tried adding a test for that, but I could figure out a way to make the test work for both pure PHP (which uses `trigger_error()` for PHP 5 compatibility) and the C extension (which throws a `TypeError` which is better in PHP 7+).

TeBoring (approved):

Issue #7885, opened 2020-09-11
C# MergeFrom ReadOnlySpan with no allocations


MaximGurschi:
Hi,

With the latest release of Grpc.Tools, we are now able to execute code that writes the serialized message directly to Span<byte> without incurring allocations. This is based on IBufferMessage. Thank you for that release!

My questions is what is the way to achieve the reverse process? Specifically without allocating a CodedInputStream or any other heap based objects. Is there (or are the plans for) an API that can help us achieve the following:
1. Given an instantiated (new allocated) Message - M,
2. Execute M.MergeFrom(ReadOnlySpan)
3. The above to complete without allocating heap memory (for things like CodedInputStream, etc.)

I have been using a custom solution to achieve the above steps but it involves subclassing **Stream** and a buffer copy to a preallocated buffer. If there were to be a native solution to achieve the equivalent of IMessage.MergeFrom(ReadOnlySpan) without additional allocation then that would help get rid of the workaround code.

MaximGurschi:
The issue with my solution is that the CodedInputStream eventually hits the state.sizelimit, so i have to allocate a new one.

A native solution that does M.MergeFrom(ReadOnlySpan) without allocating any CodedInputStream would be great.

jtattermusch:
Currently there is 

`YourMessage.Parser.ParseFrom(ReadOnlySequence<byte>)` which allows you to parse without needing a CodedInputStream (and getting a ReadOnlySequence without requiring allocations is relatively easy). 
https://github.com/protocolbuffers/protobuf/blob/022078d3ee642816688228e9434775f2a137536d/csharp/src/Google.Protobuf/MessageParser.cs#L124

There is also a MergeFrom variant of this method, but that one is internal.
https://github.com/protocolbuffers/protobuf/blob/022078d3ee642816688228e9434775f2a137536d/csharp/src/Google.Protobuf/MessageExtensions.cs#L288

Does that satisfy your needs?

MaximGurschi:
@jtattermusch Thank you for that. Unfortunately not it does not as that API allocates the holding proto. My goal is to reuse the proto and also not allocate the CIS.

jtattermusch:
@MaximGurschi  alright, so you'd basically need a public API for msg.MergeFrom(ReadOnlySequence<byte> data) (something along the lines of the latter internal method I posted), is that correct?

MaximGurschi:
@jtattermusch Precisely :)

[...]

PR #8473, opened 2021-04-10
C#: Add ParseFrom/MergeFrom ReadOnlySpan<byte>

Jensaarai:
Address #7885

### DIFF
--- a/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -926,39 +926,6 @@ namespace Google.Protobuf
             }
         }
 
-        [Test]
-        public void TestParseFromSpan()
-        {
-            var testMessage = new TestAllTypes
-            {
-                RepeatedBool = { true, false },
-                SingleDouble = 42.0d,
-                SingleString = "testing"
-            };
-
-            ReadOnlySpan<byte> messagesBytes = testMessage.ToByteArray().AsSpan();
-            TestAllTypes parsedMessage = TestAllTypes.Parser.ParseFrom(messagesBytes);
-
-            Assert.AreEqual(testMessage, parsedMessage);
-        }
-
-        [Test]
-        public void TestMergeFromSpan()
-        {
-            var testMessage = new TestAllTypes
-            {
-                RepeatedBool = { true, false },
-                SingleDouble = 42.0d,
-                SingleString = "testing"
-            };
-
-            ReadOnlySpan<byte> messagesBytes = testMessage.ToByteArray().AsSpan();
-            var mergedMessage = new TestAllTypes();
-            mergedMessage.MergeFrom(messagesBytes);
-
-            Assert.AreEqual(testMessage, mergedMessage);
-        }
-
         /// <returns>A serialized big message</returns>
         private static byte[] GenerateBigSerializedMessage()
         {
--- a/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -161,11 +161,20 @@ namespace Google.Protobuf
 
         private static void AssertReadFromParseContext(ReadOnlySequence<byte> input, ParseContextAssertAction assertAction, bool assertIsAtEnd)
         {
+            // Check as ReadOnlySequence<byte>
             ParseContext.Initialize(input, out ParseContext parseCtx);
             assertAction(ref parseCtx);
             if (assertIsAtEnd)
             {
                 Assert.IsTrue(SegmentedBufferHelper.IsAtEnd(ref parseCtx.buffer, ref parseCtx.state));
+            }
+
+            // Check as ReadOnlySpan<byte>
+            ParseContext.Initialize(input.ToArray().AsSpan(), out ParseContext spanParseContext);
+            assertAction(ref spanParseContext);
+            if (assertIsAtEnd)
+            {
+                Assert.IsTrue(SegmentedBufferHelper.IsAtEnd(ref spanParseContext.buffer, ref spanParseContext.state));
             }
         }
 
--- a/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -926,6 +926,39 @@ namespace Google.Protobuf
             }
         }
 
+        [Test]
+        public void TestParseFromSpan()
+        {
+            var testMessage = new TestAllTypes
+            {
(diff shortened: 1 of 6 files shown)


csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
```diff
@@ -926,6 +926,39 @@ public void TestParseMessagesOver2G()
             }
         }
 
+        [Test]
```
jtattermusch, line 929: I'm not sure these test really belong under CodedInputStreamTest.  There's probably a better place where to put them?

jtattermusch: Also, by adding only a simple test like this you will only get a very basic testing. 

If you added a ReadOnlySpan codepath here: https://github.com/protocolbuffers/protobuf/blob/5e84a6169cf0f9716c9285c95c860bcb355dbdc1/csharp/src/Google.Protobuf.Test/MessageParsingHelpers.cs#L53 you'd get a much better coverage and then the Span-based overloads would automatically get a lot of test excercising them.

jtattermusch:
CC @JamesNK  in case he want to review as well.

csharp/src/Google.Protobuf/ParseContext.cs
```diff
@@ -104,6 +104,29 @@ internal static void Initialize(ReadOnlySequence<byte> input, int recursionLimit
             ctx.state.ExtensionRegistry = null;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void Initialize(ReadOnlySpan<byte> input, out ParseContext ctx)
```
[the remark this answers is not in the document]
Jensaarai: Done!

csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
Jensaarai: Done!

jtattermusch (approved):
Thanks for addressing my comments.

I actually removed the "ref" from "ref ReadOnlySpan<byte>" for public methods since what I had in mind is to only add "ref" in the internal APIs. It seems that in other public APIs in .NET,  only "ReadOnlySpan<byte>" is usually passed (I guess that's because of readability and safety), we should probably to the same.

Besides that, I made a few readability improvements.

I'm marking the PR as "approved".- please let me know if the changes looks good to you.

jtattermusch:
@JamesNK  can you please take a look as well? Since this is adding a few new public methods, I wanted to get a second opinion.

Jensaarai:
Changes are fine with me :)

jtattermusch:
@JamesNK  please let me know if you're planning to review this or not.

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -294,6 +303,16 @@ internal static void MergeFrom(this IMessage message, ReadOnlySequence<byte> dat
             ParsingPrimitivesMessages.CheckReadEndOfStreamTag(ref ctx.state);
         }
 
+        [SecuritySafeCritical]
+        internal static void MergeFrom(this IMessage message, ref ReadOnlySpan<byte> data, bool discardUnknownFields, ExtensionRegistry registry)
```
JamesNK, line 307: Does data need to be ref here?

csharp/src/Google.Protobuf/ParseContext.cs
```diff
@@ -65,6 +65,29 @@ internal static void Initialize(ref ReadOnlySpan<byte> buffer, ref ParserInterna
...
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void Initialize(ref ReadOnlySpan<byte> input, out ParseContext ctx)
+        {
+            Initialize(ref input, DefaultRecursionLimit, out ctx);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void Initialize(ref ReadOnlySpan<byte> input, int recursionLimit, out ParseContext ctx)
+        {
```
JamesNK, line 76: Does input need to be ref here?

I think comments explaining how and why these Initialize overloads are different from the existing overload that takes a ReadOnlySpan on line 62.

`internal static void Initialize(ref ReadOnlySpan<byte> buffer, ref ParserInternalState state, out ParseContext ctx)`

JamesNK (commented):
Is there a test of what happens if the input ReadOnlySpan doesn't contain enough data? For example, with a ReadOnlySequence we attempt to get more data from the sequence. Does a ReadOnlySpan fail with an expected error in that situation?

Jensaarai:
> Is there a test of what happens if the input ReadOnlySpan doesn't contain enough data? For example, with a ReadOnlySequence we attempt to get more data from the sequence.
> 
> Is there the correct code to handle that situation when there isn't a sequence backing up the span? Does a ReadOnlySpan fail with an expected error in that situation?

I updated the tests around `ReadOnlySequence` to also test the sequence as a `ReadOnlySpan` (after converting to a `byte[]` with `ToArray()`). Do you think this satisfies these cases?

csharp/src/Google.Protobuf/MessageExtensions.cs
Jensaarai: I think you're right, I removed passing `ReadOnlySpan` by ref in these methods and in `ParseContext`...there are more throughout the codebase, e.g. in `ParsingPrimitives` that probably should also change, but I didn't do that here.

I believe `ReadOnlySpan` is not a big enough struct to benefit from any potential performance increase from passing by `ref` or `in` (which is probably why the Microsoft APIs using it don't use `in`)?

https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#apply-the-in-modifier-to-readonly-struct-parameters-larger-than-systemintptrsize

csharp/src/Google.Protobuf/ParseContext.cs
Jensaarai: As above, I removed passing by `ref` and added some comments with my best understanding of how they were being used!

Jensaarai:
@JamesNK @jtattermusch big enough changes in the last commit probably warrant a re-check?

jtattermusch (approved):
Still LGTM.

jtattermusch:
@JamesNK  if you're happy with the current version, I will merge.

Jensaarai:
@JamesNK I'm sure you're busy, feel free to tell me to bug off if you queue these up :)

JamesNK (approved):
LGTM!

Jensaarai:
Ready @jtattermusch ? I'm not sure what to do about the Bagel failure...

jtattermusch:
The bazel failure is unrelated to this PR.

PR #8479, opened 2021-04-12
Run python aarch64 tests on the CI

jtattermusch:
Followup for https://github.com/protocolbuffers/protobuf/pull/8392
- add a few more fixes (some of them come from https://github.com/protocolbuffers/protobuf/pull/8280)
- add kokoro jobs and scripts to run the python aarch emulated tests with the CI

once this is merged, I'll create the corresponding kokoro jobs internally.

Example of how the jobs's output would look like once created (I used a hack to get an "adhoc" run from kokoro):
sponge/1e2bbf08-5ee0-4428-b905-1786b83edc32

### DIFF
--- a/kokoro/linux/aarch64/qemu_helpers/prepare_qemu.sh
+++ b/kokoro/linux/aarch64/qemu_helpers/prepare_qemu.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+#
+# Setup and configure qemu userspace emulator on kokoro worker so that we can seamlessly emulate processes running
+# inside docker containers.
+
+set -ex
+
+# show pre-existing qemu registration
+cat /proc/sys/fs/binfmt_misc/qemu-aarch64
+
+# Kokoro ubuntu1604 workers have already qemu-user and qemu-user-static packages installed, but it's and old version that:
+# * prints warning about some syscalls (e.g "qemu: Unsupported syscall: 278")
+# * doesn't register with binfmt_misc with the "F" flag we need (see below)
+# To avoid the "unsupported syscall" warnings, we download a recent version of qemu-aarch64-static.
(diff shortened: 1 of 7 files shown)


kokoro/linux/aarch64/qemu_helpers/prepare_qemu.sh
```diff
@@ -0,0 +1,30 @@
...
+chmod ugo+x qemu-aarch64-static
+sudo mv qemu-aarch64-static /usr/local/bin
+/usr/local/bin/qemu-aarch64-static --version
+
+# register qemu-aarch64-static with binfmt_misc to use it as emulator for aarch64 binaries (based on their magic and mask)
+# The format of the configuration string is ":name:type:offset:magic:mask:interpreter:flags"
+# The most important flag is "F", which allows the qemu-aarch64-static binary to be loaded regardless of chroot and namespaces.
+# That way, we can emulate aarch64 binaries running inside docker containers transparently, without needing the emulator
+# to be accessible from the docker image we're emulating.
+echo -1 | sudo tee /proc/sys/fs/binfmt_misc/qemu-aarch64  # delete the existing registration
```
dlj-NaN, line 26: Can this registration be done with the multiarch/qemu-user-static:register container instead?

Alternatively, can you use qemu-binfmt-conf.sh from qemu directly?

dlj-NaN (changes requested):

kokoro/linux/aarch64/qemu_helpers/prepare_qemu.sh
jtattermusch: Good idea! I updated the script to use the multiarch/qemu-user-static docker image to take care of the registration.
Now the script is much simpler and clearer.

fowles (approved):

PR #8490, opened 2021-04-14
Use C++11 enum class for Utf8CheckMode

georgthegreat:
At the time this enum includes members named STRICT  and NONE, both of which are known to have problems under Windows (at least `port_undef.inc` has workaround for this macro).

As protobuf now builds in C++11 mode, `enum class` is a better solution to avoid name conflicts.

This is part 1 / x in #8494 fix.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_helpers.h
+++ b/src/google/protobuf/compiler/cpp/cpp_helpers.h
@@ -774,10 +774,10 @@ class PROTOC_EXPORT NamespaceOpener {
   std::vector<std::string> name_stack_;
 };
 
-enum Utf8CheckMode {
-  STRICT = 0,  // Parsing will fail if non UTF-8 data is in string fields.
-  VERIFY = 1,  // Only log an error but parsing will succeed.
-  NONE = 2,    // No UTF-8 check.
+enum class Utf8CheckMode {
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/cpp/cpp_helpers.h
```diff
@@ -774,10 +774,10 @@ class PROTOC_EXPORT NamespaceOpener {
   std::vector<std::string> name_stack_;
 };
 
-enum Utf8CheckMode {
-  STRICT = 0,  // Parsing will fail if non UTF-8 data is in string fields.
-  VERIFY = 1,  // Only log an error but parsing will succeed.
-  NONE = 2,    // No UTF-8 check.
+enum class Utf8CheckMode {
+  Strict = 0,  // Parsing will fail if non UTF-8 data is in string fields.
```
acozzette, line 778: Looks great to me but one small thing: our [style](https://google.github.io/styleguide/cppguide.html#Enumerator_Names) for enum names is to name them `kStrict`, `kVerify`, `kNone`. (Sometimes we don't use that convention in our existing code because our previous convention was to use `ALL_CAPS` names.)

georgthegreat: Fixed!

PR #8485, opened 2021-04-13
Run java aarch64 tests on the CI (with an emulator)

jtattermusch:
Based on https://github.com/protocolbuffers/protobuf/pull/8479 (only the last few commits need to be reviewed).

- build protoc under crosscompiler
- then build & tests java code under qemu emulator.

Example passing adhoc run: http://sponge2.corp.google.com/95eaabd8-a929-4dfd-839a-4b4bef078391

PR #8495, opened 2021-04-14
Run ruby aarch64 tests on the CI (with an emulator)

jtattermusch:
Based on https://github.com/protocolbuffers/protobuf/pull/8485 (only the few last commits are ruby related).

Ruby's native build is very simple (it only compiles a few .c files), so running ruby's build directly under an emulator is the simplest solution and it's quick enough. Only the protoc binaries get crosscompiled since building protoc under an emulator would be slow.

### DIFF
--- a/kokoro/linux/aarch64/protoc_crosscompile_aarch64.sh
+++ b/kokoro/linux/aarch64/protoc_crosscompile_aarch64.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+#
+# Builds protobuf C++ with aarch64 crosscompiler.
+
+set -ex
+
+./autogen.sh
+CXXFLAGS="-fPIC -g -O2" ./configure --host=aarch64
+make -j8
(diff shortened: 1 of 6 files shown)


jtattermusch:
@dlj-NaN  please review or reassign to protobuf ruby owner.

fowles (commented):
this appears to include both java and python as part of this.  Please update these to be separate or commit earlier ones before pushing follow ons

fowles:
Josh, you can wait for jtattermusch to separate the change a bit before reviewing

jtattermusch:
@fowles  I rebased the PR. The original reason for depending on the other changes was that is it reuses `kokoro/linux/aarch64/protoc_crosscompile_aarch64.sh` introduced by the java PR. 
To make the review simpler, I simply added an identical file into this PR.

@haberman  @fowles  please review.

fowles (approved):

jtattermusch:
@haberman  is this ok to merge?

haberman (approved):

Issue #8494, opened 2021-04-14
Generated map<string, string> code might trigger unused parameter warning


georgthegreat:
**What version of protobuf and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?** Linux, any will do

**What runtime / compiler are you using (e.g., python version or gcc version)**
clang or gcc, with -Wunused-parameter switch enabled

**What did you do?**
1. Use map<string, string> field in a message
2. Use syntax="proto2"; (default behavior)
3. Don't use optimize_for="lite_runtime"; (default behavior)
3. Get
```
error: unused parameter 'p' [-Werror,-Wunused-parameter]
      static void Check(ConstPtr p) {
```
in utf8 key / value validation.

**Anything else we should know about your project / environment**

The issue is triggered by incorrect check. At the time key / value validation method is invoked if either key or value is string.
However, the Utf8CheckMode must be taken into account and check generation / invokation should be omitted.

The fix will result in a number of pull requests. I am goind to attach them all to this issue.

Issue #4293, opened 2018-02-08
Missing extension in protobuf-3.5.1 + proto2


WindLeeWT:
### We're trying to migrate from protobuf-2.6.1 to protobuf-3.5.1. 

After reading through official documents, we decide to use **protobuf-3.5.1 + proto2** which should be fully compatible with the existing protos within code base. It built and passed unit tests with few minor changes. However, when we put newly built services online, they kept showing "extensions not found" errors. After some investigation I wrote a simple example below to illustrate :

Example code was put under a temporary directory(common/rpc/temp)：

#### example_options.proto
```protobuf
package example;

import "google/protobuf/descriptor.proto";

extend google.protobuf.ServiceOptions {
  optional int32 example_version = 20000;
}
```
#### example.proto
```protobuf
package example;

import "common/rpc/temp/example_options.proto";

option cc_generic_services = true;

message ShowRequest {
  optional string name = 1;
}

message ShowResponse {
}

service ExampleService {
  option (example_version) = 100;
  rpc Show(ShowRequest) returns (ShowResponse);
}
```
#### employee.h
```c++
#include <string>

#include "google/protobuf/descriptor.h"

namespace example {

class Employee {
 public:
  explicit Employee(const std::string& name);
  void Report();

 private:
  std::string name_;
  static const google::protobuf::MethodDescriptor* method_descriptor;
};

}
```
#### employee.cc
```c++
#include "common/rpc/temp/employee.h"

#include <iostream>
#include "common/rpc/temp/example.pb.h"

namespace example {

const google::protobuf::MethodDescriptor* Employee::method_descriptor =
    nullptr;
    // Assign to nullptr: example_version found
    // Assign to "Show" method: example_version not found
    // ExampleService::descriptor()->FindMethodByName("Show");

Employee::Employee(const std::string& name) : name_(name) { }

void Employee::Report() {
  std::cout << "Employee::Report" << std::endl
            << "    Name: " << name_ << std::endl
            << "    MethodDescriptor: " << method_descriptor << std::endl;
}

}
```
#### main.cc
```c++
#include <iostream>
#include <memory>

#include "google/protobuf/descriptor.pb.h"
#include "common/rpc/temp/example.pb.h"
#include "common/rpc/temp/employee.h"

using example::example_version;
using example::ExampleService;
using example::Employee;

int main() {
  std::cout << "Enter main" << std::endl;

  const auto* descriptor = ExampleService::descriptor();
[...]

PR #8498, opened 2021-04-15
Remove trivial methods from the codebase

georgthegreat:
This simplifies the code in GetUtf8CheckMode dramatically.

This is part 2 / x in #8494 fix.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_helpers.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_helpers.cc
@@ -1011,29 +1011,16 @@ bool IsWellKnownMessage(const FileDescriptor* file) {
   return well_known_files.find(file->name()) != well_known_files.end();
 }
 
-static bool FieldEnforceUtf8(const FieldDescriptor* field,
-                             const Options& options) {
-  return true;
-}
-
-static bool FileUtf8Verification(const FileDescriptor* file,
-                                 const Options& options) {
-  return true;
-}
-
 // Which level of UTF-8 enforcemant is placed on this file.
 Utf8CheckMode GetUtf8CheckMode(const FieldDescriptor* field,
                                const Options& options) {
-  if (field->file()->syntax() == FileDescriptor::SYNTAX_PROTO3 &&
-      FieldEnforceUtf8(field, options)) {
+  if (field->file()->syntax() == FileDescriptor::SYNTAX_PROTO3) {
     return Utf8CheckMode::kStrict;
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/cpp/cpp_helpers.cc
```diff
@@ -1011,29 +1011,16 @@ bool IsWellKnownMessage(const FileDescriptor* file) {
   return well_known_files.find(file->name()) != well_known_files.end();
 }
 
-static bool FieldEnforceUtf8(const FieldDescriptor* field,
```
acozzette, line 1014: Sorry but unfortunately we need to keep these functions around. Within our internal version of the codebase we have some additional code in these functions for legacy reasons, to allow the UTF-8 enforcement to be disabled for particular fields. Even though the functions look pointless, it helps us to have them around so that we can easily apply our internal changes.

If you just want to fix an unused-parameter warning, it would be fine to remove the parameter names (i.e. just make it `(const FieldDescriptor*, const Options&)`) or add lines like `(void) field;`.

georgthegreat:
@acozzette, @haberman could you, please, take a look onto GetOptimizeFor implelementation in cpp_helpers.cc?

Internally it invokes `HasBootstrapProblem()` which looks like a workaround for #4293, which was fixed by constinit introduction in 3.15. Could you confirm my speculations? If `HasBootstrapProblem()` could be removed, interface of `GetUtf8CheckMode()` could be simplified.

The method was introduced during sync from Google's internal codebase in b99994d, so there may be additional details available internally.

src/google/protobuf/compiler/cpp/cpp_helpers.cc
georgthegreat: I am not working on unused parameter in protoc itself, but rather in the generated code.

It is ok if these methods would not be removed, though this will make interfaces a bit clumsier.

Could you, please, take a look on the question below?

acozzette:
I believe the calls to `HasBootstrapProblem()` in `GetOptimizeFor()` were added a while ago when we were looking into the possibility of using `CODE_SIZE` for (almost) all protos in the build. That turned up some weird edge cases where it did not work, and we try to detect those with the `HasBootstrapProblem()` helper function so that we can fall back on the standard `SPEED` mode. I could be wrong but I don't think that logic was intended as a workaround for #4293.

georgthegreat:
Ok, thanks for the clarification.

I am closing thing PR due to the reasoning from above,

PR #8500, opened 2021-04-16
Fix -Wunused-parameter in map<string, int> fields (fixes #8494)

### LINKED ISSUE #8494: Generated map<string, string> code might trigger unused parameter warning

georgthegreat:
This is a clumsy fix for #8494.
MWE is as follows:

```
syntax = "proto2";
optimize_for = "lite_runtime";
message M {
   map<string, int> field = 1;
}
```

During code generation, protoc will check for key / value types (corresponds to string_key / string_value variables all around).
If Utf8 validation will be disabled due to evaluation of `GetUtf8CheckMode()` to `kNone`, the `Check()` method will be generated as follows:
```
static void Check(ConstPtr p) {
}
```

This will trigger `-Wunused-parameter`.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_map_field.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_map_field.cc
@@ -230,9 +230,9 @@ void MapFieldGenerator::GenerateSerializeWithCachedSizesToArray(
     format(
         "struct Utf8Check {\n"
         "  static void Check(ConstPtr p) {\n"
-		// p may be unused when GetUtf8CheckMode evaluates to kNone,
-		// thus disabling the validation.
-		"    (void)p;\n");
+        // p may be unused when GetUtf8CheckMode evaluates to kNone,
+        // thus disabling the validation.
+        "    (void)p;\n");
     format.Indent();
     format.Indent();
     if (string_key) {
--- a/src/google/protobuf/compiler/cpp/cpp_map_field.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_map_field.cc
@@ -229,7 +229,10 @@ void MapFieldGenerator::GenerateSerializeWithCachedSizesToArray(
   if (utf8_check) {
     format(
         "struct Utf8Check {\n"
-        "  static void Check(ConstPtr p) {\n");
+        "  static void Check(ConstPtr p) {\n"
+		// p may be unused when GetUtf8CheckMode evaluates to kNone,
+		// thus disabling the validation.
+		"    (void)p;\n");
     format.Indent();
     format.Indent();
     if (string_key) {


src/google/protobuf/compiler/cpp/cpp_map_field.cc
```diff
@@ -229,7 +229,10 @@ void MapFieldGenerator::GenerateSerializeWithCachedSizesToArray(
   if (utf8_check) {
     format(
         "struct Utf8Check {\n"
-        "  static void Check(ConstPtr p) {\n");
+        "  static void Check(ConstPtr p) {\n"
+		// p may be unused when GetUtf8CheckMode evaluates to kNone,
```
acozzette, line 233: Looks good but would you mind fixing the indentation? (I think your editor must have inserted tabs whereas we always use spaces.)

georgthegreat: Should be fixed now.

Btw, have you considered providing [.editorconfig](https://editorconfig.org/) with the basic indentation rules?

acozzette: The formatting doesn't look quite right yet, maybe you just need to do a git push?

I hadn't heard of editorconfig before. I wonder if someone has already written a config to match the Google C++ style guide.

georgthegreat: Sorry, something went wrong with git pushing.
Should be fixed now.

georgthegreat:
@acozzette, could you, please, clarify the release timeline of protobuf=3.16 and protobuf=3.17 (as I see, 3.17.0-rc1 has just appeared in tags).

I have most of the patches submitted as PRs applied locally, and I am waiting for release to remove them. As I see, this PR is not included into 3.16.0-rc2 though it landed in the master some time ago.

Does protobuf follow the trunk based development practice?

acozzette:
We ordinarily do what one would call trunk-based development, but this is just an unusual case where we need to do two releases in quick succession. 3.16 is going to be a long-term support release for Java and will get fixes for at least a year, but we have other changes we want to release in 3.17 which we don't want to commit to supporting for that long. 3.17 includes everything merged to master up to this morning and I expect to release it either by tomorrow or early next week.

georgthegreat:
Thanks for the clarification!
I will wait to 3.17 then.

PR #8501, opened 2021-04-16
Run csharp aarch64 tests on the CI (with an emulator)

jtattermusch:
Based on #8495 (only the few last commits are csharp related).

- protobuf C# is written in pure C# (no native components), so we can simply build the test project on x64 (for better build speed) and then use aarch64 docker imager to run all the tests under an emulator.


fowles (commented):
appears to include all earlier changes, please separate them

jtattermusch:
Rebased. It's now basically just 1 script (and a few trivial kokoro configs).

fowles (approved):

PR #8503, opened 2021-04-16
Use byteswap.h when building against musl libc

georgthegreat:
Though manpages say bswap_ _macros are GNU extensions_, both bionic libc and musl libc provide them from the very first releases.

As musl intentionally does not define any specific macro, I suggest changing ifdefs to be based on operating systems rather than on libc's.

### DIFF
--- a/src/google/protobuf/stubs/port.h
+++ b/src/google/protobuf/stubs/port.h
@@ -78,7 +78,7 @@
 #include <intrin.h>
 #elif defined(__APPLE__)
 #include <libkern/OSByteOrder.h>
-#elif defined(__GLIBC__) || defined(__BIONIC__) || defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


acozzette:
Thanks, @georgthegreat.

georgthegreat:
Thanks, @acozzette.
As a follow up, I'd like to add a code snippet illustrating the problem being fixed.
Whenever a source is dependent on both byteswap.h and protobuf, the following error will occur:
```
byteswap.h:22:9: error: 'bswap_16' macro redefined [-Werror,-Wmacro-redefined]
#define bswap_16(x) __bswap_16(x)
        ^
google/protobuf/stubs/port.h:321:9: note: previous definition is here
#define bswap_16(x) bswap_16(x)
        ^
```

PR #8506, opened 2021-04-16
Bazel powered Java testing

perezd:
This PR makes it possible to run all build, unit, and conformance tests for Java, Java Lite, and Util protobuf subprojects. Testing locally, all tests pass when running `bazel test //java:tests` which is the new parent test suite target responsible for exercising all the related tests.

This PR also modernizes the way we deal with inbound maven dependencies by migrating us over to [rules_jvm_external](https://github.com/bazelbuild/rules_jvm_external).


BUILD
```diff
@@ -458,6 +459,12 @@ cc_binary(
 # Tests
 ################################################################################
 
+filegroup(
+    name = "testdata",
+    visibility = ["//:__subpackages__"],
+    srcs = glob(["src/google/protobuf/testdata/**/*"]),
```
haberman, line 465: Can we list these explicitly? My recent experiences have led me to hate `glob()`.

BUILD
```diff
@@ -519,6 +526,57 @@ RELATIVE_TEST_PROTOS = [
...
+    "google/protobuf/unittest_preserve_unknown_enum2.proto",
+    "google/protobuf/unittest_proto3.proto",
+    "google/protobuf/unittest_proto3_arena.proto",
+    "google/protobuf/unittest_proto3_arena_lite.proto",
+    "google/protobuf/unittest_proto3_lite.proto",
+    "google/protobuf/unittest_proto3_optional.proto",
+    "google/protobuf/unittest_well_known_types.proto",
+]
+
+GENERIC_TEST_PROTOS = ["src/" + s for s in GENERIC_RELATIVE_TEST_PROTOS]
```
haberman, line 558: Flatten to just put `src/` in the filenames themselves? Or even better can we just list the sources directly in the `proto_library()` rule? `GENERIC_TEST_PROTOS` only seems to be used in one place.

BUILD
```diff
@@ -519,6 +526,57 @@ RELATIVE_TEST_PROTOS = [
...
+        "//:struct_proto",
+        "//:timestamp_proto",
+        "//:type_proto",
+        "//:wrappers_proto",
+    ],
+)
+
 cc_proto_library(
     name = "cc_test_protos",
     srcs = LITE_TEST_PROTOS + TEST_PROTOS,
```
haberman, line 582: It doesn't have to be in this PR, but I'd really like us to use normal aspect-based `xx_proto_library()` rules where `proto_library()` rules are deps, instead of listing proto sources here.

Actually, if @com_google_protobuf is us within this package, I almost wonder if we can just use the regular vanilla `cc_proto_library()` rule for anything that doesn't need to be bootstrapped.

BUILD
```diff
@@ -1111,3 +1178,75 @@ sh_test(
...
+  deps = [":conformance_proto"],
+)
+
+java_lite_proto_library(
+    name = "test_messages_proto3_java_proto_lite",
+    deps = [":test_messages_proto3_proto"],
+)
+
+java_binary(
+  name = "conformance_java",
```
haberman, line 1213: It feels like the conformance stuff should go in a separate `conformance/BUILD` file. I guess we have some conformance stuff already in the main BUILD file, so maybe moving it all in one go should be a follow-up CL.

java/core/BUILD
```diff
@@ -126,3 +136,190 @@ proto_lang_toolchain(
...
+    "src/test/java/com/google/protobuf/UnknownEnumValueTest.java",
+    "src/test/java/com/google/protobuf/UnknownFieldSetLiteTest.java",
+    "src/test/java/com/google/protobuf/UnknownFieldSetTest.java",
+    "src/test/java/com/google/protobuf/WellKnownTypesTest.java",
+    "src/test/java/com/google/protobuf/WireFormatTest.java",
+]
+
+junit_tests(
+    name = "lite_tests",
+    srcs = glob(["src/test/java/**/*.java"], exclude = LITE_TEST_EXCLUSIONS),
```
haberman, line 311: Avoid glob()?

protobuf.bzl
```diff
@@ -506,3 +514,27 @@ def check_protobuf_required_bazel_version():
     copied filegroup. (Fixed in bazel 0.5.4)
     """
     versions.check(minimum_bazel_version = "0.5.4")
+
+
+def conformance_test(name, testee, language):
```
haberman, line 519: Could we have it pass failure list files directly?  Passing language and inferring filenames from that feels a bit too indirect/magical.

haberman (changes requested):

BUILD
fowles: +1

BUILD
fowles: +1.  Generally speaking build files should be as dumb and straightforward as possible so tooling can manipulate it more easily

java/core/BUILD
```diff
@@ -126,3 +136,190 @@ proto_lang_toolchain(
...
+)
+
+genrule(
+    name = "rewrite_javalite_test_util",
+    srcs = [
+        "//java/lite:lite.awk",
+        "src/test/java/com/google/protobuf/TestUtil.java"
+    ],
+    outs = ["TestUtil.java"],
+    cmd = "awk -f $(location //java/lite:lite.awk) $(location src/test/java/com/google/protobuf/TestUtil.java) > $@"
```
fowles, line 245: ugh, awk?  What is this thing doing and can we do it with some more direct means?

BUILD
perezd: We should do this 100%, and it would work, I've verified this. We should do it separately, however.

BUILD
perezd: Yeah, I was hoping to rectify the grossness in a followup, right now I am just augmenting what we have to at least be consistent for future refactors. I am not sure why this is not in its own folder.

java/core/BUILD
perezd: this is a port of the pom.xml logic. I believe it's pretending to be a preprocessor and stripping out sections of code that don't belong.

BUILD
perezd: glob isn't an antipattern, and given the fact that at this moment in time our story for keeping lists of explicit files in sync is sort of a mess, I'm in favor of this causing loud failures vs silent failures for the time being.

BUILD
perezd: Yeah I have no idea why this is this way, I believe it has to do with a proto rule expecting one thing and the filesystem being another, but I did this to stay consistent within the file.

java/core/BUILD
perezd: Given this is a direct port of the logic we have in place and it needs to consistently work for maven as well, I'm gonna leave this as is.

protobuf.bzl
perezd: Yeah, I prefer your suggestion and reimplemented it. I mistakenly thought I couldn't pass args to sh_test, but turns out you can and it's much easier this way.

BUILD
perezd: Discussed offline, once we have file_list.bzl imported, we can refactor this.

protobuf.bzl
```diff
@@ -506,3 +514,32 @@ def check_protobuf_required_bazel_version():
     copied filegroup. (Fixed in bazel 0.5.4)
     """
     versions.check(minimum_bazel_version = "0.5.4")
+
+
+def conformance_test(name, testee, failure_list = None, text_format_failure_list = None):
```
haberman, line 519: I think we unfortunately have users importing `protobuf.bzl` right now, so it's in a pseudo-public state. Are we intending to make this public for users? If not can we move it to a separate file that is more clearly internal? (Ideally with an actual visibility list, but I'm not sure if Bazel supports this right now).

java/testing.bzl
```diff
@@ -0,0 +1,72 @@
+"""
```
haberman, line 1: Do we have a way of keeping .bzl files private through enforcement? If not, can we more clearly mark this as internal-only in the file or directory name?

BUILD
```diff
@@ -1111,3 +1178,74 @@ sh_test(
         "update_file_lists.sh",
     ],
 )
+
+java_proto_library(
```
haberman, line 1182: Visibility private on these, unless we mean for this BUILD file to be their ultimate location?

haberman (changes requested):

BUILD
perezd: Done! also fixed some inconsistent formatting.

java/testing.bzl
perezd: I thought they were but they aren't https://docs.bazel.build/versions/master/visibility.html#visibility-of-bzl-files I'll move to an "internal" folder.

protobuf.bzl
perezd: honestly, I think we should make this public eventually.

haberman: If we mean to make this function public, I think it makes more sense to put in `conformance/`. This file is full of cruft that we want to burn down, I don't think we should add things to it.

If we're trying to keep this function private, I think we should move it to a filename that makes that clear, and not `protobuf.bzl` which we know a lot of users already include directly.

perezd: Talked offline, moved to internal.bzl

haberman (approved):

PR #8510, opened 2021-04-19
Fix python include path

georgthegreat:
All other includes in `protobuf/python` are relative to `protobuf/python/src` while this one is not.

### DIFF
--- a/src/google/protobuf/stubs/port.h
+++ b/src/google/protobuf/stubs/port.h
@@ -78,7 +78,7 @@
 #include <intrin.h>
 #elif defined(__APPLE__)
 #include <libkern/OSByteOrder.h>
-#elif defined(__GLIBC__) || defined(__BIONIC__) || defined(__CYGWIN__)
+#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__)
 #include <byteswap.h>  // IWYU pragma: export
 #endif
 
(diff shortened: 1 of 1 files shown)


python/google/protobuf/internal/python_protobuf.cc
```diff
@@ -30,7 +30,7 @@
 
 // Author: qrczak@google.com (Marcin Kowalczyk)
 
-#include <google/protobuf/python/python_protobuf.h>
```
acozzette, line 33: @georgthegreat I'm trying to understand how this even worked before. It seems like it should have been a compile error. @anandolee Do you know if this file is still used? I think this file might not even be getting built.

georgthegreat: I do not know.
We have had this patch for many years now.

georgthegreat: From I very brief check I can say that this file is not needed any more.
If you could confirm it, I can change this PR to remove the file (and corresponding header) instead. Or you can do this by yourself.

acozzette: Sounds good to me to just delete the files, but @anandolee can probably say for sure whether we still need them.

georgthegreat:
Sorry, but I have failed to attach a git branch to this PR.
The continuation is in #8513

PR #8512, opened 2021-04-19
Add use_integers_for_number_values option

vtitor:


TeBoring:
https://developers.google.com/protocol-buffers/docs/proto3#json
 I don't think our json spec have this feature. To be consistent for all languages, we only support flag mentioned in the spec.

PR #8513, opened 2021-04-19
Remove unused python_protobuf.{cc,h}

georgthegreat:
This is a continuation of #8510.


acozzette:
I think this [line](https://github.com/protocolbuffers/protobuf/blob/29152fbc064921ca982d64a3a9eae1daa8f979bb/Makefile.am#L1011) probably just needs to be deleted.

georgthegreat:
@acozzette, I see that [google/protobuf/util/__init__.py](https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/util/__init__.py) does not makes anything useful either.

Could you check if it can be removed?

acozzette:
I'm not sure. @anandolee Do you know if that file can be removed?

PR #8516, opened 2021-04-20
Fix some -Wsign-compare issues

georgthegreat:
Switch to range-based for loops where possible, use size_t for offsets to avoid conversions between signed and unsigned types where it is not.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_helpers.h
+++ b/src/google/protobuf/compiler/cpp/cpp_helpers.h
@@ -752,11 +752,11 @@ class PROTOC_EXPORT NamespaceOpener {
       if (name_stack_[common_idx] != new_stack_[common_idx]) break;
       common_idx++;
     }
-    for (auto oldNameIt = name_stack_.crbegin(); oldNameIt != name_stack_.crend() - common_idx; ++oldNameIt) {
-      if (*oldNameIt == "PROTOBUF_NAMESPACE_ID") {
+    for (auto it = name_stack_.crbegin(); it != name_stack_.crend() - common_idx; ++it) {
+      if (*it == "PROTOBUF_NAMESPACE_ID") {
         printer_->Print("PROTOBUF_NAMESPACE_CLOSE\n");
       } else {
-        printer_->Print("}  // namespace $ns$\n", "ns", *oldNameIt);
+        printer_->Print("}  // namespace $ns$\n", "ns", *it);
       }
     }
     name_stack_.swap(new_stack_);
--- a/src/google/protobuf/compiler/cpp/cpp_helpers.h
+++ b/src/google/protobuf/compiler/cpp/cpp_helpers.h
@@ -713,8 +713,8 @@ class PROTOC_EXPORT Formatter {
     std::vector<int> path;
     descriptor->GetLocationPath(&path);
     GeneratedCodeInfo::Annotation annotation;
-    for (int i = 0; i < path.size(); ++i) {
-      annotation.add_path(path[i]);
+    for (int index: path) {
+      annotation.add_path(index);
     }
     annotation.set_source_file(descriptor->file()->name());
     return annotation.SerializeAsString();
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/cpp/cpp_helpers.h
```diff
@@ -746,25 +746,19 @@ class PROTOC_EXPORT NamespaceOpener {
...
         Split(name, "::", true);
-    int len = std::min(name_stack_.size(), new_stack_.size());
-    int common_idx = 0;
-    while (common_idx < len) {
-      if (name_stack_[common_idx] != new_stack_[common_idx]) break;
-      common_idx++;
-    }
-    for (int i = name_stack_.size() - 1; i >= common_idx; i--) {
-      if (name_stack_[i] == "PROTOBUF_NAMESPACE_ID") {
+    for (auto oldNameIt = name_stack_.crbegin(); oldNameIt != name_stack_.crend(); ++oldNameIt) {
```
georgthegreat, line 749: The index-based algorithm version used before turned out to be unstable to replacing int with size_t.
Hence I have changed it to range-based loops (C++ does not have reversed range-based for yet, hence iterators were used).

The new algorithm fully closes nested namespaces and fully opens them.

As of C++17 we have nested namespaces, so none of this will be required soon.

acozzette: Won't this new version result in more namespace openings and closings than necessary? I thought a goal of the old version was to figure out the longest common namespace prefix so that we don't have to close and reopen the same namespaces unnecessarily. E.g. if we're changing from `foo::bar` to `foo::baz` then we don't need to close and reopen `foo`.

acozzette: Oh, I see you already mentioned this in your comment. I think we probably need to keep the old behavior unfortunately. It might be a while before we can start using the C++17 syntax because we don't have any plans to require C++14 or C++17 yet.

georgthegreat: Indeed, the number of namespace operations was increaced in this PR, but I do not think it will lead to any observable change (please, tell if you know of any).

I have attempted to fix -Wsign-compare without simplifying the code around, but proper fix would require more obfuscation: it looks like the current version attempts to operate negative indices in 
`for (int i = name_stack_.size() - 1; i >= common_idx; i--) { ` 

If I change i to being unsigned, an overflow will be triggered.

acozzette: There wouldn't be an observable semantic change, but I think there could be a build performance regression just from the increase in the generated source code size. I believe within Google we have some proto files that could have pb.cc files thousands of lines longer if we fully closed and reopened the namespaces.

I'm not sure what the best solution for the compiler warnings would be, but maybe one easy fix would be to just keep everything signed and immediately `static_cast` to `int` all the `size()` calls.

georgthegreat: Ok, I will try to find a better solution that would not cause additional namespace operations.

georgthegreat: I have restored the old behavior, but I keep using iterators for reversed iteration during namespace closure.

src/google/protobuf/compiler/cpp/cpp_helpers.h
```diff
@@ -746,21 +746,21 @@ class PROTOC_EXPORT NamespaceOpener {
...
-    int common_idx = 0;
+    size_t len = std::min(name_stack_.size(), new_stack_.size());
+    size_t common_idx = 0;
     while (common_idx < len) {
       if (name_stack_[common_idx] != new_stack_[common_idx]) break;
       common_idx++;
     }
-    for (int i = name_stack_.size() - 1; i >= common_idx; i--) {
-      if (name_stack_[i] == "PROTOBUF_NAMESPACE_ID") {
+    for (auto oldNameIt = name_stack_.crbegin(); oldNameIt != name_stack_.crend() - common_idx; ++oldNameIt) {
```
acozzette, line 755: Nice, this looks like a good solution. One more small thing, though: `oldNameIt` should be in `snake_case`, maybe rename it to `old_name_iter` or even just `iter`?

georgthegreat: I have changed it to simple `it`, which is common in the code I write. I hope such naming will fit.

acozzette:
@georgthegreat Thanks!

PR #8519, opened 2021-04-21
fix(splitFloat32): Fix incorrect exponent when converting double to float

### LINKED ISSUE #6887: JavaScript: some float values are serialized into IEEE 754 binary representation incorrectly
Version: v3.9.0 (also reproduces on master).
Language: Javascript / Typescript

Windows 10 (1709).

Typescript 3.6.2, Python 3.7.2 

**What did you do?**
1. Create a message with a float field:
```
message Test
{
  float a = 1;
}
```
2. Generate Javascript code for the message definition.
3. Create an instance of the message, set a value to 0.9999999762949594.
4. Serialize message into buffer - IEEE 754 binary representation of the value will be wrong.

In actual application I experienced this issue as value 0.9999999762949594 was first serialized and then deserialized as 0.5 (!).

The value is serialized as 0x00, 0x00, 0x80, 0x3F

**What did you see instead?**
The value is serialized as 0x00, 0x00, **0x00**, 0x3F

I made investigation of the issue and seems that there is a bug with values close to power of 2.
Function with the bug: [jspb.utils.splitFloat32](https://github.com/protocolbuffers/protobuf/blob/master/js/binary/utils.js#L159).
Line of code with the bug: `mant = Math.round(mant * jspb.BinaryConstants.TWO_TO_23) & 0x7FFFFF` [LoC](https://github.com/protocolbuffers/protobuf/blob/master/js/binary/utils.js#L204)

Several test cases were added by me into protobuf/js/binary/utils_test.js and tests started failing.
Reason: for some input values mantissa is rounded to a value 0x1000000 and then truncated with a mask 0x7FFFFF => resulting in zero mantissa. In this case we probably have to increase exponent value by 1.

I tried two potential fixes - first is based on the current implementation, second is based on ArrayBuffer/DataView to check my first implementation.

Diff with my tests & suggested fix is attached to this issue: [Fix.zip](https://github.com/protocolbuffers/protobuf/files/3847232/Fix.zip)

zakhenry:
Submitted on behalf of a third-party: Ihor Darkov
See https://github.com/protocolbuffers/protobuf/issues/6887#issue-522941516 for original source

Closes #6887

### DIFF
--- a/js/binary/utils.js
+++ b/js/binary/utils.js
@@ -202,7 +202,11 @@ jspb.utils.splitFloat32 = function(value) {
 
   exp = Math.floor(Math.log(value) / Math.LN2);
   mant = value * Math.pow(2, -exp);
-  mant = Math.round(mant * jspb.BinaryConstants.TWO_TO_23) & 0x7FFFFF;
+  mant = Math.round(mant * jspb.BinaryConstants.TWO_TO_23);
+  if (mant >= 0x1000000) {
+    ++exp;
+  }
+  mant = mant & 0x7FFFFF;
 
   jspb.utils.split64High = 0;
   jspb.utils.split64Low = ((sign << 31) | ((exp + 127) << 23) | mant) >>> 0;
--- a/js/binary/utils_test.js
+++ b/js/binary/utils_test.js
@@ -391,6 +391,11 @@ describe('binaryUtilsTest', function() {
     // Pi.
     test(f32_pi, 0x40490fdb);
 
+    // corner cases
+    test(0.9999999762949594, 0x3f800000);
+    test(7.99999999999999, 0x41000000);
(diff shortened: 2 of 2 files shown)


jorgbrown:
I can vouch for the JavaScript code being a good fix for the bug.

There are some 64-bit floating-point values very close to 2^128 which in floating-point math would round to FLOAT32_MAX if JavaScript supported 32-bit floating-point, but which this code considers to be infinity because they are greater than the FLOAT32_MAX constant.  I don't think that's an issue worth worrying about.

zakhenry:
Is there any update on the review of this MR? It's been stale for nearly 3 months now

zakhenry:
@TeBoring any chance of getting a more available reviewer for this pull request? It's approaching the 6 month milestone now..

lukesandberg:
Thanks Zak!

I have confirmed the fix along with several of the other incorrect serializations listed on the bug.

PR #8523, opened 2021-04-21
ensure maven deps are pinned for reproducibility.

perezd:


fowles (approved):

WORKSPACE
```diff
@@ -50,8 +50,12 @@ maven_install(
         "https://jcenter.bintray.com/",
         "https://repo1.maven.org/maven2",
     ],
+    maven_install_json = "//:maven_install.json",
```
dlj-NaN, line 53: Can you add a comment with instructions to update this?

dlj-NaN (changes requested):

WORKSPACE
perezd: Done.

PR #8532, opened 2021-04-26
Fix -Wundefined-inline error when using SharedCtor() or SharedDtor()

georgthegreat:
protoc marks SharedCtor and SharedDtor as inline in `.h` file being generated, but does not define them in this header.
An attempt to use these symbols causes compiler to issue `-Wundefined-inline` warning / error if `-Werror` is enabled.


georgthegreat:
NB: I do not know if there is any procedure behind pregenerated .pb.h files update for well known types.
I just use `sed` to fix them.

acozzette:
Thanks @georgthegreat

PR #8539, opened 2021-04-27
Stop using GG_ prefixed macros

georgthegreat:
`GG_ULONGLONG` was renamed to `PROTOBUF_ULONGLONG` some time ago.

This PR fixes the generated code for `uint64 field = 1 [default = 0];` stanza.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_helpers.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_helpers.cc
@@ -664,9 +664,9 @@ std::string UInt64ToString(const std::string& macro_prefix, uint64_t number) {
 std::string DefaultValue(const FieldDescriptor* field) {
   switch (field->cpp_type()) {
     case FieldDescriptor::CPPTYPE_INT64:
-      return Int64ToString("GG", field->default_value_int64());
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/cpp/cpp_helpers.cc
```diff
@@ -664,9 +664,9 @@ std::string UInt64ToString(const std::string& macro_prefix, uint64_t number) {
 std::string DefaultValue(const FieldDescriptor* field) {
   switch (field->cpp_type()) {
```
acozzette, line 665: Actually I think this helper function could be simplified even further and just be a one-liner that returns `DefaultValue(Options(), field)`.

georgthegreat: Agreed.

It looks like this method is not used by the protoc itself, hence the issue was not discovered earlues. I believe that some of our plugins triggered the issue.

I suggest merge this PR to fix the typo while I am looking for the proper fix in the plugin.

acozzette:
@georgthegreat Thanks!

PR #8542, opened 2021-04-28
[Java] Implement CodedInputStream::readRawBytes with destination byte buffer

nbauernfeind:
I have a custom marshaller that wraps a CodedInputStream into an InputStream to better process a field while performing fewer copies and generating less java-garbage. This PR adds some direct support to delegate calls from `InputStream::read(byte[] dst, int off, int len)` to the backing CodedInputStream.

It seems a bit unrelated, but it would be convenient to also allow `WireFormat::makeTag` to be public. I would prefer to use the existing implementation rather than cloning a version of it. Even [apache arrow could use this](https://github.com/apache/arrow/blob/master/java/flight/flight-core/src/main/java/org/apache/arrow/flight/ArrowMessage.java#L90) if it were not package private.


nbauernfeind:
Can someone please add the labels:
`release notes: yes`
`java`

Thank you =).

nbauernfeind:
> We will try to keep our response time within 7-days but if you don’t get any response in a few days, feel free to comment on the threads to get our attention

Is there any possibility to get this PR in front of a code reviewer? It has been open for almost two weeks now.

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
```diff
@@ -1246,18 +1256,26 @@ public byte readRawByte() throws IOException {
...
+    }
+
+    @Override
+    public void readRawBytes(final byte[] dst, final int offset, final int length)
+        throws IOException {
+      if (length >= 0 && length <= (limit - pos)) {
         final int tempPos = pos;
         pos += length;
-        return Arrays.copyOfRange(buffer, tempPos, pos);
+        System.arraycopy(buffer, tempPos, dst, offset, length);
```
perezd, line 1273: This throws 3 exceptions we should catch and rethrow as a runtime InvalidProtocolBufferException or IOException, whichever makes the most sense: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/lang/System.java#L522

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
```diff
@@ -2864,8 +2924,42 @@ public byte readRawByte() throws IOException {
         pos = tempPos + size;
         return Arrays.copyOfRange(buffer, tempPos, tempPos + size);
```
perezd, line 2925: Should this be using System.arraycopy as well? Also concerning that we're not catching/rethrowing exceptions here.

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
```diff
@@ -2864,8 +2924,42 @@ public byte readRawByte() throws IOException {
...
+      // Verify that the message size so far has not exceeded currentLimit.
+      if (currentMessageSize > currentLimit) {
+        // Read to the end of the stream anyway.
+        skipRawBytes(currentLimit - totalBytesRetired - pos);
+        throw InvalidProtocolBufferException.truncatedMessage();
+      }
+
+      final int sizeBuffered = Math.min(size, bufferSize - pos);
+      if (sizeBuffered > 0) {
+        System.arraycopy(buffer, pos, dest, offset, sizeBuffered);
```
perezd, line 2954: catch/rethrow exceptions related to System.arraycopy here as well.

java/core/src/main/java/com/google/protobuf/CodedInputStream.java
```diff
@@ -3851,40 +3944,16 @@ public byte readRawByte() throws IOException {
...
-     * throw {@code truncatedMessage} exception.
-     *
-     * @param bytes
-     * @param offset
-     * @param length
-     * @throws IOException
-     */
-    private void readRawBytesTo(byte[] bytes, int offset, final int length) throws IOException {
+    @Override
+    public void readRawBytes(byte[] bytes, int offset, final int length) throws IOException {
```
perezd, line 3956: public methods must have javadoc defining what the method is doing for external users.

perezd (changes requested):
Please include tests that demonstrate and exercise these changes that affect the public API. Thanks for the submission!

perezd:
Do you have any benchmarks that demonstrate the improvements created by this PR? I'd love to see something concrete before accepting this.

nbauernfeind:
Thanks for the review @perezd. I hadn't added explicit tests since I diverted all of the existing calls that go through the `public byte[] readRawBytes(final int length)` api's through the new API that allow you to pass a destination buffer. I was hoping that I'd get away with the existing tests. ;)

I will see what I can throw together to utilize and test the new API.

The goal, in general, is for custom marshallers to be able to provide the destination buffers during the un-marshalling phase. This enables a custom marshaller to 1) generate less garbage and 2) create opportunity for copying the payload a single time. (i.e. a custom marshaller similar to apache arrow's which expects a protobuf body followed by an opaque payload.)

Can you quantify what kind of benchmark you would find concrete? It is straight forward to put together a benchmark where the marshaller uses the previous API but is otherwise decoding the same payload (or similar payload) in a loop. With the right JVM settings I can create pressure that exists in the existing API but does not have pressure in the new API. It feels like a silly test, so I just want to make sure the approach is what you are looking for.

perezd:
> Can you quantify what kind of benchmark you would find concrete? It is straight forward to put together a benchmark where the marshaller uses the previous API but is otherwise decoding the same payload (or similar payload) in a loop. With the right JVM settings I can create pressure that exists in the existing API but does not have pressure in the new API. It feels like a silly test, so I just want to make sure the approach is what you are looking for.

I'm open to suggestions here, this is a pretty big change to how bytes are managed, so it'd be nice to demonstrate some microbenchmarks that tell us this change at a minimum does not negatively impact customers once submitted.

perezd:
Also, please ping when you've addressed the comments.

nbauernfeind:
@perezd @deannagarcia Can either of you point me to an example of work that adds a microbenchmark?

PR #8546, opened 2021-04-30

haberman:


benchmarks/download_data.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
```
acozzette, line 1: I think this file was supposed to be deleted, looks like the sync probably reinstated it inadvertently.

haberman: Done.

acozzette (approved):

PR #8558, opened 2021-05-04
Updated and simplified PHP testing structure

haberman:
This PR simplifies a number of issues around the PHP testing setup:

- Before the pure-PHP and C extension testers used different autoloaders. This PR consolidates them into a single autoloader, created by composer.
- We also remove the custom PHPUnit version selection logic in `test.sh` in favor of letting Composer pick the correct dependencies for each PHP version.
- Before the pure-PHP and C extension used different PHPUnit invocation strategies: pure-PHP used a `phpunit.xml` file, while the C extension enumerated the list of testing files in a shell script. Now both use the same strategy: we run PHPUnit on a directory, which automatically tests all of the files matching `*Test.php` in that directory.
- Consolidated on a single autoloader, created by composer.
- We now rely on autoloading to import all test protos. require_once() calls for test protos are removed.
- For now the valgrind tests are removed. A follow-up PR will re-enable them in a more robust way.
- Optimized both test proto generation and the custom extension build to avoid unnecessary work when the files are already up to date.
- Added assertions to verify that the C test doesn't use PHP sources, this should eliminate the need for separate "C" and "mixed" tests.
- Removed obsolete rules from tests.sh.


php/tests/force_c_ext.php
```diff
@@ -0,0 +1,14 @@
...
+// We have to test this because the command-line argument will fail silently
+// if the extension could not be loaded:
+//    php -dextension=ext/google/protobuf/modules/protouf.so
+if (!extension_loaded("protobuf")) {
+    throw new Exception("Protobuf extension not loaded");
+}
+
+spl_autoload_register(function($class) {
+    if (strpos($class, "Google\\Protobuf") === 0) {
+        throw new Exception("When using the C extension, we should not load runtime class: " . $class);
```
TeBoring, line 12: well known types are bundled in c extension?

haberman: Yes, they are as of https://github.com/protocolbuffers/protobuf/pull/7944

TeBoring (approved):

perezd (approved):

PR #8560, opened 2021-05-04
Update Objective-C ProtocolBuffers_iOS.xcodeproj To Skip Install

Twigz:
Fixes an issue with the objective-c iOS .xcodeproj, where archiving a project using this repository as a submodule resulted in a generic Xcode archive being created, rather than an iOS App Archive. Setting the `SKIP_INSTALL` build setting to `YES` results in the iOS App Archive being generated correctly when archiving an application for submission.


Twigz:
@googlebot I signed it!

thomasvl:
Should we do the same thing for the other Apple OSes?

Twigz:
I can change this on the other `.xcodeproj`s if you'd like. `SKIP_INSTALL = YES` is likely the right default for them as well, but I was more sure about iOS since I had a good test case.

thomasvl:
Ok, we'll hold of for now, thanks for the patch!

thomasvl (approved):

PR #8562, opened 2021-05-05
Add class method Timestamp.from_time to ruby well known types

mcclymont:
I was surprised to find out that `from_time` was not a class method, but an instance method.
This limits its usefulness, as to create a new Timestamp you have to first instantiate the object, then call the from_time method, and then you need to return/use the saved instance, because the from_time method returns the self.nanos value, rather than the timestamp itself.

Code example

Before
```ruby
{
  updated_at: Google::Protobuf::Timestamp.new.tap { |t| t.from_time(updated_at) },
}
```

After
```ruby
{
  updated_at: Google::Protobuf::Timestamp.from_time(updated_at),
}
```

### DIFF
--- a/ruby/lib/google/protobuf/well_known_types.rb
+++ b/ruby/lib/google/protobuf/well_known_types.rb
@@ -85,6 +85,11 @@ module Google
       def from_time(time)
         self.seconds = time.to_i
         self.nanos = time.nsec
+        self
+      end
+
+      def self.from_time(time)
+        new.from_time(time)
       end
 
       def to_i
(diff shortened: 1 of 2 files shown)


TeBoring:
@mcclymont, could you sign the cla?

mcclymont:
@googlebot I signed it!

mcclymont:
@TeBoring What's the next step? I don't understand the CI failures

deannagarcia:
Sorry for the delay, I'm rerunning tests now and will let you know if there's anything that needs fixed.

deannagarcia:
Thanks for the addition, merging now!

PR #8571, opened 2021-05-05
Export SerializeInternalToArray

johnduffynh:
Export SerializeInternalToArray which is used by the next function, TableSerializeToArray.


acozzette:
Thanks, @johnduffynh.

PR #8579, opened 2021-05-07
Use template Delete<> for invoking a dependent template

georgthegreat:
This fixes MSVC (tested on MSVC2019 / 16.8) compilation error (line number is different, as I am working with protobuf 3.15.8):
```
google/protobuf/map_entry_lite.h:527: error C2760: syntax error: unexpected token ')', expected 'expression'
```

I believe [this](https://stackoverflow.com/questions/34696351/template-dependent-typename) explains why MSVC considers this to be a dependent template.

### DIFF
--- a/src/google/protobuf/map_entry_lite.h
+++ b/src/google/protobuf/map_entry_lite.h
@@ -527,7 +527,7 @@ class MapEntryLite : public MapEntryImpl<T, MessageLite, Key, Value,
       SuperType;
   constexpr MapEntryLite() {}
   explicit MapEntryLite(Arena* arena) : SuperType(arena) {}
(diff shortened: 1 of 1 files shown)


georgthegreat:
JFY:

I have found another issue with MSVC. By default, it forbids using offsetof in constexpr statements, which is used in `RepeatedPtrFieldBase::InternalSwap()` (see repeated_field.h).

It can be fixed by passing `/D_CRT_USE_BUILTIN_OFFSETOF` to the compiler (this will switch a macro-based implementation to the builtin one), but I do not see this defined being used in protobuf.

I do not know if this could be / should be handled by the library, or if it should be just mentioned in some documentation.

PR #8594, opened 2021-05-10
Mark C# generated code with System.CodeDom.Compiler.GeneratedCode as well

jtattermusch:
Related issue: https://github.com/grpc/grpc/issues/24594
Corresponding gRPC PR https://github.com/grpc/grpc/pull/26164

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
@@ -56,6 +56,9 @@ SourceGeneratorBase::~SourceGeneratorBase() {
 
 void SourceGeneratorBase::WriteGeneratedCodeAttributes(io::Printer* printer) {
   printer->Print("[global::System.Diagnostics.DebuggerNonUserCodeAttribute]\n");
+  // We don't provide protoc info in the [GeneratedCode] attribute the because the
+  // pre-generated code in this repository would then have junk info in it.
+  printer->Print("[global::System.CodeDom.Compiler.GeneratedCode(\"protoc\", null)]\n");
 }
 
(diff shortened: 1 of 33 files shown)


src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
```diff
@@ -56,6 +56,9 @@ SourceGeneratorBase::~SourceGeneratorBase() {
 
 void SourceGeneratorBase::WriteGeneratedCodeAttributes(io::Printer* printer) {
   printer->Print("[global::System.Diagnostics.DebuggerNonUserCodeAttribute]\n");
+  // We don't provide protoc info in the [GeneratedCode] attribute the because the
```
jskeet, line 59: "the because the" => "because the" (for some reason I can't add this as a suggestion).

src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc
```diff
@@ -56,6 +56,9 @@ SourceGeneratorBase::~SourceGeneratorBase() {
...
+  // pre-generated code in this repository would then have junk info in it.
```
jskeet, line 60: I don't really follow what you mean by "junk info". It might be worth clarifying that the second parameter would be the version information, if we wanted to include that... but I agree that it's not worth including it (as it would create a lot of churn every time the version is updated.)

jskeet (approved):
It's a pity we can't do it just on the class itself, but never mind. At least the generator change is tiny :)

Approved, but I've suggested a comment that could be clarified.

jtattermusch:
@jskeet  thanks for the review!
The macos changes seem unrelated, so I'm going to merge soon.

Issue #7131, opened 2020-01-23
C# : Why generated code throw exception on null values(.e.g string) instead of just setting default value?


zeinali0:
Why generated code throw exception on null value assignment?
I know in new version of protobuf(v3) you decide to remove optional feature and all field most have value.
But why just don't set default value for some primitive type like string instead of throwing ArgumantNullException? this behavior has conflict with some tools like AutoMapper.
I mean this behavior would affect other part of a program.
if in grpc-area all null value most be replaced with default value so just keep this constraint in grpc-area. throwing a exception for setting null value would cause this constraint flow in other part of program which i think it is not a good practice and it is against cross-cutting concern.

jtattermusch:
I'm not sure I understand the question. Can you provide an example what is the current behavior of Google.Protobuf (code example) is and how it should be changed?

zeinali0:
I mean instead of throwing exception just use default value for strings.


**This is current behavior(which throw ArgumantNullException):**
```
public const int FirstNameFieldNumber = 2;
    private string firstName_ = "";
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public string FirstName {
      get { return firstName_; }
      set {
        firstName_ = pb::ProtoPreconditions.CheckNotNull(value, "value");
      }
    }
```

**This is my suggestion:**

```
public const int FirstNameFieldNumber = 2;
    private string firstName_ = "";
    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
    public string FirstName {
      get { return firstName_; }
      set {
        firstName_ = value ?? "";
      }
    }
```

anon17:
AIU, your code is generated by grpc, not protobuf.

anon17:
You can use [StringValue](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/test_messages_proto3.proto#L216) that is mapped to nullable string. Though tests have stuff like [optional string field](https://github.com/protocolbuffers/protobuf/blob/master/csharp/src/Google.Protobuf.Test.TestProtos/TestMessagesProto3.cs#L615), I wonder if string is optional by default.

zeinali0:
Yes, But they forward me here. I asked question there at first(https://github.com/grpc/grpc-dotnet/issues/734)

zeinali0:
[...]

PR #8595, opened 2021-05-10
3.15 allow null value for string properties in csharp

arthurvaverko-kaltura:


arthurvaverko-kaltura:
@googlebot I signed it!

arthurvaverko-kaltura:
> What is the motivation for this change?
> 
> Generally we try to limit codegen options as much as possible, to avoid having different variants of the generated API.

The motivation is linked to this issue 
https://github.com/protocolbuffers/protobuf/issues/7131

Falco20019:
As mentioned in the other issue, I would not limit it to string but have it on any nullable value. But also as mentioned over in the other issue, I would go the route of having a helper on `IFieldAccessor` to call it specifically on a field. protoc options would affect all string fields in all compiled proto files (if set), and this might also not be the best option. The helper would be more selective as you would call it only in the situations where you care and know what you do.

IF going with nullables, I would prefer something like `clear_field_on_null` instead of `allow_null_string`.

arthurvaverko-kaltura:
@Falco20019 let me try to understand and maybe clarify our struggle right now.

1. Our .net service has a REST api endpoint, and while using Newtonsoft.Json sterilizer if a given request model has a `string` property and was not sent in the POST body the sterilizer will use null value on that string property 
2. This is an accaptable behaviour while using .net all the way because string having a possible null value is something known for c#
3. I know that having nullable strings changes that in c# 8 but we are looking for a solution for the current situation with previous language versions.
4. So when that REST model is then desterilized and is trying to be mapped to a proto generated model it will throw an exception only at runtime, causing lots of mistakes being made while writing the code (due to the fact that the target property is a string an a typical cshapr developer will never think of null checking a string value)
5. to avoid the above we want to allow the seamless mapping between models so i have created this PR in the hope it will find its way to the official release and we can avoid using a fork
6. Regarding a helper for `IFieldAccessor` if i understood this would make the writer of the proto files to have to declare in each field the helper `clear on null`  and the proto files are written by the service that is owning the data so he doesn't really care if he generates a client in csharp or any other language for that matter. and the problem stated occurs only for the generated csharp models
8. I dont see how adding same behavior for any other nullable type is not  will  benefit as they dont suffer that problem. the csharp client will generate nullable csharp property e.g. (int? float? etc...) only if we use well known types like google.protobuf.Int32Value for example, so in that case we do intend the value to be passed as null
9.  the `--csharp_opt` flag is best solution for this as it provides forward compatibility for c#8 and allow the users who care to generate a null value string safe client for working in csharp.

I know its a very long argument I hope I didn't misunderstood and made the issue and the motivation behind the PR more clear ..

Falco20019:
Thanks for the clarification. `Newtonsoft.Json` has some knobs that can be used to define the behavior on what happens when deserializing. Usually, it will keep the default value of the field and not change it. If I understood you correctly, you have two models, one used for REST and one used for Proto. How get those two mapped? Is there some manual code? Reflection?

Based on how you map and what else is using the REST model, there are some options that I could think of to solve this without needing the option.

If the REST model is only used here, you can change the default value of the field to `string.Empty`. A more generic way would be to use a `ContractResolver` on your deserializer.

It would be good to know which sides are fixed. Is the proto defined by you? Is the REST model defined by you? Or only the mapping code/usage? Depending on that, there also might come up some possibilities to solve the issue.

arthurvaverko-kaltura:
So the situation is more complex .. I tried to simplify it in my previous comment.
The fact is that the Sterilizer settings cannot be changed to make missing JSON string props to become empty string values.
The code that can be changed at the moment is the mapping code and the GRPC client code.

For mapping we are not using reflection and relaying on manually mapping, that is because in many cases there are to many differences and mapping logic to just use reflection.

I really think that making any adjustments as "workaround" wont make sense and would be error prone for lots of other developers as the will required the mapping code to do the same null check I have added to the setter of string properties.

arthurvaverko-kaltura:
one last thing regarding c#8 just realized that nullable strings i.e `string?` will not be an issue .. and will not actually solve the problem.

take the following code example
```csharp
using System;

#nullable enable
namespace cons1
{

    class Data
    {
        public string notNull { get; set; }
        public string? nullable { get; set; }
    }

    class Program
    {



        static void Main(string[] args)
        {

            var data = new Data();


            Console.WriteLine($"is string notNull ? {data.notNull == null} and value is: [{data.notNull}]");
            Console.WriteLine($"is string? nullable ? {data.notNull == null} and value is: [{data.notNull}]");

            try
            {
                Console.WriteLine($"try to call nullable.ToString(): [{data.nullable.ToString()}]");
            }
            catch (Exception e)
            {
                Console.WriteLine($"err while trying to call nullable.ToString() {e.Message}");
            }

            try
            {
                Console.WriteLine($"try to call notNull.ToString(): [{data.notNull.ToString()}]");
            }
            catch (Exception e)
            {
                Console.WriteLine($"err while trying to call notNull.ToString() {e.Message}");
            }
        }
    }
}

```

the output is:
``` shell
PS C:\code\z_archive\cons1> dotnet run
C:\code\z_archive\cons1\Program.cs(9,23): warning CS8618: Non-nullable property 'notNull' must contain a non-null value when exiting constructor. Consider declaring the property as nullable. [C:\code\z_archive\cons1\cons1.csproj]
C:\code\z_archive\cons1\Program.cs(29,72): warning CS8602: Dereference of a possibly null reference. [C:\code\z_archive\cons1\cons1.csproj]
C:\code\z_archive\cons1\Program.cs(38,71): warning CS8602: Dereference of a possibly null reference. [C:\code\z_archive\cons1\cons1.csproj]
is string notNull ? True and value is: []
is string? nullable ? True and value is: []
err while trying to call nullable.ToString() Object reference not set to an instance of an object.
err while trying to call notNull.ToString() Object reference not set to an instance of an object. 
```

the nullable ref types in c#8 will only serve static code analysis and add Warnings when they are note initilized or de-refrenced without a null check .. 


The main issue with current setters for generated proto messages in csharp is that if during runtime for some reason you get a null value an exception is thrown, this causes any user to add null checks before setting string values into proto generated models

I believe that the solution has to be either what I sugested here or to have some *Compile-Time* error to be thrown to a developer that is trying to set a "primitive" string var into a proto model.

Falco20019:
Your main issue is, that the proto file is defined using `string` and not `google.protobuf.StringValue`. For compatibility  with other languages, `string` is non-nullable. So it is required of your mapping logic to make sure your models fit (which they don‘t if the proto defines `string` over `google.protobuf.StringValue`).

arthurvaverko-kaltura:
I see your point. Indeed if we generate from a proto with `google.protobuf.StringValue` there is no null check .. 
But in our case the service (the one who defined the proto file) defined it using `string` and is written in go .. so we do have to make it compatible.

unfortunately in csharp there is no good alternative for a  non nullable string.
so we are left with 3 options:
1. keeping the same runtime null check in the setter (in my opinion its error prone and will cause lots of devs to fall in that pit)
2. assuming that csharp string null value means an empty string. (in 99% of the time I believe that would be the best alternative)
3. create a separate `struct` type that will represent a non-nullable string that will be generated when a `string` type is used in proto file vs a regular csharp string `google.protobuf.StringValue`

option 3 is more complex and require the introduction of a new type, probbably similar in implementation to [Microsoft.Extensions.Primitives.StringValues](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringvalues?view=dotnet-plat-ext-5.0)
but option 3 will give complete compatibility with other languages regarding null values for strings.

Falco20019:
AFAIK go should be one of the languages where `string` can not hold `null` (or `nil`) on a regular `string`. So I assume, that the service defining that proto is then also not able to handle `null` and uses the default for `string` in go, which is the empty string.

So if I understood your case correctly, I would assume that option 2 would fit best here, since you already seem to map an existing REST API to that go service which has an existing proto (gRPC?) API. At least, clearing the field in C# will be deserialized as empty string on the go side.

arthurvaverko-kaltura:
Exactly! 
So the service in go is using strings .. and they are defaulted to an empty string value.

While in csharp we need some way to greacfully  handle the null string issue without throwing an exception .. 

So you believe this is a good candidate for a PR and other might benefit from it ? 

Or should we just keep our fork of protoc ?

Falco20019:
Personally, I think this is what the mapping code needs to take care of and therefore, there you would need to check for `null` and use `string.Empty` as default. But I understand that you see that as error prone while not using C#8 which would warn the user and that you prefer to be able to setup a different fallback behavior.

So I can see that some might benefit from it. But I‘m also just a regular user. So let‘s see what the maintainers think about it. I just tried to see if there are easier solutions and also help everyone involved understand your use case.

arthurvaverko-kaltura:
I see.. many thanks for the insights.

danielgindi:
This should actually support nulls without defaulting to an empty string.
If a field is optional, and null by default, we should be able to set it back to null. By assignment, to adhere to c# standards.

arthurvaverko-kaltura:
> This should actually support nulls without defaulting to an empty string. If a field is optional, and null by default, we should be able to set it back to null. By assignment, to adhere to c# standards.

Not really possible as we want to be consistent with other languages where string is a value type and does not hold null value like go

danielgindi:
> > This should actually support nulls without defaulting to an empty string. If a field is optional, and null by default, we should be able to set it back to null. By assignment, to adhere to c# standards.
> 
> 
> 
> Not really possible as we want to be consistent with other languages where string is a value type and does not hold null value like go
> 
> 
> 
> Nullables are defined using a well know proto type, essentially a struct of nullable


I'd reconsider the requirement for consistency with other languages.
C# != c++.
Every language has its own styles.
And a `oneof` struct is not enough because the generated code still does not allow nulls, it requires calling a "Clear" method.

Falco20019:
In favor of C# 8 and the new nullable standard, I would stick with the current behavior. string = string, StringValue = string?

Also in C#.

arthurvaverko-kaltura:
I completely agree with the flag hence why I added the flag in this PR.

However c#8 indeed support nullable string with the required distinction to better align with other languages, but in prior version or if nullable ref types are but turned on (and they are not by default) we get a wierd behavior were in compile time it's perfectly okay to assign null to a string property of a generated proto message and only during runtime you would get and exception. 

This is opposite to the philosophy of .net "throw you to the pit of success" and basically a pitfal many probably fall into.

danielgindi:
> In favor of C# 8 and the new nullable standard, I would stick with the current behavior. string = string, StringValue = string?

C# 8 is not the only C# out there, and it's not even out there... and it's configurable whether you want refs to be nullable or not.



> You state the assumption "and the field is optional". But using string explicitly makes it NOT optional. It will always come initialized with an empty string for every language. 

Well why not optimize the `oneof` case to allow setting the value to null, which will do what `Clear()` does? At least based on some flag if some people like their users to crash when setting `null`s to optional properties...

Falco20019:
If it's about "only recognizable at runtime", shouldn't the NotNullAttribute help about this? To at least give warnings?

Again, I still don't understand what you mean with optional. Are you talking about proto2 or proto3? Because proto3 doesn't have a notation for optional anymore.

danielgindi:
By optional, I mean nullable, as in C# terms.
And in proto3 we can specify an "optional" by doing a single `oneof` of a string.
That can be a case of None, or contain a value. (An empty string is also a value).
In that case- I expect it to allow setting to null, instead of having to call `Clear`. So users can construct an object in a straightforward C-sharpy way, without a lot of `if`s and method calls.

I understand the proto3 does not currently understand the concept of `null`, and an unspecified field returns a default of `""`. But we can optimize the case of a single string `oneof`.

We can perhaps add a flag for this specific case, to allow null to clear the value.

Falco20019:
> By optional, I mean nullable, as in C# terms. And in proto3 we can specify an "optional" by doing a single `oneof` of a string. That can be a case of None, or contain a value. (An empty string is also a value). In that case- I expect it to allow setting to null, instead of having to call `Clear`. So users can construct an object in a straightforward C-sharpy way, without a lot of `if`s and method calls.
The proper way to do this (according to the documentation), is using the well-known type instead of writing one on your own and expecting library support. If you use `google.protobuf.StringValue` in your proto (version 2 and 3), you will get a nullable message containing only a string value. Doing exactly what you ask for. This also works for all other value types (like int32; using `google.protobuf.Int32Value` is mapped to `int32` and therefore natively usable in C# correctly).

> I understand the proto3 does not currently understand the concept of `null`, and an unspecified field returns a default of `""`. But we can optimize the case of a single string `oneof`.
> 
> We can perhaps add a flag for this specific case, to allow null to clear the value.
Doing a `oneof` still has a different meaning. If you do a `onof` having only the `string`, this means, that it COULD get other fields in the future. So it's absolutely not safe make this assumption of "this is a nullable string".

arthurvaverko-kaltura:
@Falco20019 
> If it's about "only recognizable at runtime", shouldn't the NotNullAttribute help about this? To at least give warnings?

A warning is good but not enough in my opinion as most large project has lots of warning and technical debt, this is much more than a warning since this is for sure something that will cause your system to break and throw an exception.

Falco20019:
> A warning is good but not enough in my opinion as most large project has lots of warning and technical debt, this is much more than a warning since this is for sure something that will cause your system to break and throw an exception.

The main issue is still, that the specification you implement against, is explicitly not expecting and allowing `null` on purpose. So the system breaking, if not conforming to a specification, is exactly what I would expect a system to do.

I still understand your point, that it's not so obvious, since C# was missing a good way to distinguish if a `string` is nullable from the generated code up until C#8 and that there are still lots of existing projects not being able to switch. This is why I said I'm fine with a flag (see https://github.com/protocolbuffers/protobuf/issues/7131#issuecomment-838320156). Basically saying: "I know that this will never become `null` but it's default value and transmitted using this". Which must not be the default behavior. The main point is, for `proto2`, that setting something to default and calling `Clear()` has a different meaning for `optional` fields, since the `HasXXX` getter will behave differently.

To clarify my point, I see different use cases.

- You are the maintainer of the proto and you expect a string to be nullable
  - Make sure to use `google.protobuf.StringValue` instead of `string`
- You are the maintainer of the proto and you expect a string to never be nullable
  - Make sure to use `string` to make this clear for your users
- You are the user of a existing proto using `google.protobuf.StringValue` and you expect it to be nullable
  - Great, both sides used the types correctly
- You are the user of a existing proto using `string` and you expect it to be nullable
  - There are two possible sub-cases here
    - You want it to be nullable to reset it to it's default and just want to use it with object initialization in a nice way
      - This is the use-case that the flag would help with
    - You want it to be nullable because you thing a `string` is nullable and should be able to store `null` at runtime
      - You have a wrong assumption. `string` is not nullable and not optional. Even with the flag, setting it to `null` and reading from it will give you an empty string and not null and that's correct behavior

danielgindi:
@Falco20019 Now I feel like I have wasted your time for nothing.
I did not know about `google.protobuf.StringValue`, and it seems to put in a simple nullable string, which also does not default to an empty string. Which is exactly the default behavior in C# (and java).

I was point at your last presented case, using `oneof` to make the string "optional", and to not have to use `Clear` over there. But `google.protobuf.StringValue` is doing exactly that, I guess that the C# transpiler is optimizing this case of a single field.

Thank you!

Falco20019:
@danielgindi Glad to help! I had a feeling that there was something off in communication :) You might want to have a look at https://docs.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protobuf-data-types to find out about even more well-known types common to all languages with special meaning (especially `google.protobuf.Timestamp`, `google.protobuf.Duration` and `google.protobuf.Empty` are also note-worthy).

There are also other pages like that and this one is a bit misleading since it also talks about WCF, but it's the best I found for C#. These types are embedded into the `Google.Protobuf.dll` and referencing them, will ensure to use those and not even generate own code and add even more helpers (like with `Timestamp.FromDateTime(...)`, `timestamp.ToDateTime()`, `Any.Pack(...)` and `Any.TryUnpack<T>(...)`).

Since you also mentioned Java, you will find, that also other languages have those helpers for these types.

acozzette:
@arthurvaverko-kaltura Thank you for the pull request but unfortunately I don't think we will accept this because we try hard to avoid adding special knobs for changing code generation behavior.

One thing to note is that we reintroduced the `optional` keyword in proto3, so you can now mark a field `optional` if you want to distinguish between a field being absent versus being an empty string (or 0 in the case of numeric fields).

arthurvaverko-kaltura:
@acozzette thanks!
any docs regarding the new `optional` feature ? 
I mean proto3 defines all properties of a message as optional .. so does `optional` in the context of proto3 really means i allow this field to be absent from the request ? 

so if i do .. and then change it back to being not optional means I'm breaking compatibility of previous version of my proto?

acozzette:
There is some documentation about it [here](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md#how-to-enable-explicit-presence-in-proto3), but the basic idea is that you can use `optional` in proto3 the same way you would use it in proto2. One thing that is confusing is that we used to consider all non-repeated non-map fields in proto3 as "optional", but we're now trying to refer to those as "singular" fields to be more precise. (So a proto3 field is optional if it has the `optional` keyword, but otherwise is just singular.)

You can safely change between singular and `optional` (i.e. add or remove the `optional` keyword), but the only thing to watch out for is that any binaries which consider the field to be singular (non-optional) are going to drop any presence information.

PR #8609, opened 2021-05-12
Make the dist_install test use Java 8

acozzette:
This is necessary because the Kotlin support requires Java 8 or
higher.


kokoro/linux/dockerfile/test/java_stretch/Dockerfile
```diff
@@ -29,3 +29,10 @@ RUN apt-get install -y \
   # -- For java_linkage_monitor \
   openjdk-8-jdk \
   && apt-get clean
+
+# Python dependencies
+RUN apt-get install -y \
```
dlj-NaN, line 34: You could merge the apt commands together, if you want. That would make the container slightly faster to build.

dlj-NaN (approved):

PR #8615, opened 2021-05-14
Use correct enum values.

thomasvl:
The clang Xcode 12.5 seems to actually detect this long standing mistake.


dmaclach (approved):

thomasvl:
Kokoro is taking a while, so validated all the ObjC builds locally.

PR #8616, opened 2021-05-14
Fix unportable test(1) operator

0-wiz-0:


0-wiz-0:
This is a bug in the configure.ac, so it should have the label 'autotools' and 'release-notes: no'. Not sure how to add those.

PR #8618, opened 2021-05-14
Fix unmatched braces.

benjaminp:


benjaminp:
Taken care of by #8613 I see.

PR #8621, opened 2021-05-17
Updated upb version and fixed PHP to not get unset message field.

haberman:


acozzette (approved):

php/ext/google/protobuf/php-upb.c
```diff
@@ -1721,16 +1778,24 @@ typedef bool eqlfunc_t(upb_tabkey k1, lookupkey_t k2);
 
 /* Base table (shared code) ***************************************************/
 
-/* For when we need to cast away const. */
-static upb_tabent *mutable_entries(upb_table *t) {
-  return (upb_tabent*)t->entries;
+static uint32_t upb_inthash(uintptr_t key) {
+  return (uint32_t)key;
```
fowles, line 1782: at the very least fold the high bits back over the low bits, but you really likely want some basic mixing here

haberman: The code rearrangement makes this look new, but this is the pre-existing algorithm. I'm happy to revisit this decision, and upb even has a benchmark now that should be able to help evaluate this choice, but this PR probably isn't the best place for that.

(upb is usually using field numbers as keys, which are usually small numbers with good uniqueness in the lower bits and little information in the high bits. On the other hand, upb's inttable special-cases small numbers and packs them into an array, so the array members going into the hash part of the table may indeed benefit from a better hash function).

acozzette (approved):

PR #8623, opened 2021-05-17
Kotlin: update pom.xml to include dokka

shaod2:
Add plugin to generate Kotlin docs


java/kotlin/pom.xml
```diff
@@ -204,6 +205,7 @@
               <sourceDirs>
                 <sourceDir>${generated.sources.dir}</sourceDir>
                 <sourceDir>${project.basedir}/src/main/kotlin</sourceDir>
+                <sourceDir>${project.basedir}/src/main/kotlin/com/google/protobuf</sourceDir>
```
deannagarcia, line 208: Do we need both this source directory and the one above?

deannagarcia (approved):

PR #8626, opened 2021-05-18
Use cfg="exec" instead of cfg="host" in internal rules.

c-mita:
Bazel has switched its protobuf rules to make them use the proto
compiler in the exec configuration instead of the host configuration.

However, if rules in protobuf still use the host configuration then
multiple copies of the compiler can end up being built.

cfg="host" is deprecated in any case and being replaced with cfg="exec"
where possible.
https://docs.bazel.build/versions/master/skylark/rules.html#configurations


acozzette:
@c-mita Thanks!

Issue #8588, opened 2021-05-08
Build badges are not rendering correctly


iamrajiv:
Some build badges are not loading correctly in README.

![Screenshot 2021-05-09 at 12 51 17 AM](https://user-images.githubusercontent.com/42106787/117550939-ac5c2100-b060-11eb-8216-85742111f344.jpg)

PR #8634, opened 2021-05-20
Remove build badges, they were consistently out of date and not very useful.

haberman:
Fixes: https://github.com/protocolbuffers/protobuf/issues/8588


dlj-NaN (approved):

dlj-NaN:
Yeah, lots of these are flat-out wrong. Thanks for catching this.

Issue #8593, opened 2021-05-10
Build issues on QNX 7.0


dschopf:
**What version of protobuf and what language are you using?**
Version: v3.16.0
Language: C++

**What operating system (Linux, Windows, ...) and version?**
QNX 7.0 build on Ubuntu 20.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
GCC 5.4.0

**What did you do?**
Check out v3.16.0
Run `CC=qcc CXX=q++ ./configure --host aarch64-qnx-gnu && make`

**What did you expect to see**
Build succeeds

**What did you see instead?**
Build fails

First, there are many errors about `#include <endian.h>`, which is not available on QNX.
This can be fixed with the following patch which also adds support for 64 bit builds:
```
diff --git a/src/google/protobuf/io/coded_stream.h b/src/google/protobuf/io/coded_stream.h
index b6c11d27a..1cf006832 100644
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -136,8 +136,10 @@
 #elif defined(__FreeBSD__)
 #include <sys/endian.h>  // __BYTE_ORDER
 #else
+#if !defined(__QNX__)
 #include <endian.h>  // __BYTE_ORDER
 #endif
+#endif
 #if ((defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)) ||    \
      (defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN)) && \
     !defined(PROTOBUF_DISABLE_LITTLE_ENDIAN_OPT_FOR_TEST)
diff --git a/src/google/protobuf/stubs/platform_macros.h b/src/google/protobuf/stubs/platform_macros.h
index ce1b1e365..6d8135755 100644
--- a/src/google/protobuf/stubs/platform_macros.h
+++ b/src/google/protobuf/stubs/platform_macros.h
@@ -46,7 +46,11 @@
 #define GOOGLE_PROTOBUF_ARCH_32_BIT 1
 #elif defined(__QNX__)
 #define GOOGLE_PROTOBUF_ARCH_ARM_QNX 1
+#if defined(aarch64)
+#define GOOGLE_PROTOBUF_ARCH_64_BIT 1
+#else
 #define GOOGLE_PROTOBUF_ARCH_32_BIT 1
+#endif
 #elif defined(_M_ARM) || defined(__ARMEL__)
[...]

PR #8637, opened 2021-05-21
Fix QNX 7.0 support #8593

dschopf:
These changes resolve all issues described in #8593 except for the linker issue.


dschopf:
@googlebot I signed it!

fowles:
Looks mostly good, but I would like to have Jan take a look because of the aarch64 change in src/google/protobuf/stubs/platform_macros.h

pablorcum:
@dschopf Please include the missing labels `release notes: yes` and `c++` to enable merging.

dschopf:
Sorry, I can't do that, looks like I don't have permission to add or edit labels.

jtattermusch (commented):
I don't have any experience with QNX, but the change in  platform_macros.h seems fair
(and it only kicks in when `__QNX__` is defined, so I don't think it can break the aarch64 linux support in any way).
I'll let @fowles to review the rest of the code and give the final approval.

fowles (approved):

Issue #8559, opened 2021-05-04
Seg fault using Ruby google-protobuf v3.15.8


stanhu:
**What version of protobuf and what language are you using?**
Version: v3.15.8
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**

Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**

```sh
$ docker run -it registry.gitlab.com/gitlab-org/gitlab-build-images:ruby-2.7.2.patched-golang-1.14-git-2.31-lfs-2.9-chrome-89-node-14.15-yarn-1.22-postgresql-11-graphicsmagick-1.3.36 bash
Unable to find image 'registry.gitlab.com/gitlab-org/gitlab-build-images:ruby-2.7.2.patched-golang-1.14-git-2.31-lfs-2.9-chrome-89-node-14.15-yarn-1.22-postgresql-11-graphicsmagick-1.3.36' locally
ruby-2.7.2.patched-golang-1.14-git-2.31-lfs-2.9-chrome-89-node-14.15-yarn-1.22-postgresql-11-graphicsmagick-1.3.36: Pulling from gitlab-org/gitlab-build-images
bd8f6a7501cc: Already exists
750858b04380: Pull complete
3826b530b192: Downloading [======================>                            ]  111.2MB/242.5MB
3826b530b192: Pull complete
714f4683e9a8: Pull complete
f722c9addae9: Pull complete
3b9016f50984: Pull complete
1116e939b23c: Pull complete
da229ef1ac62: Pull complete
67025ec68add: Pull complete
2b071db8eead: Pull complete
04721e1e144e: Pull complete
08081dca5877: Pull complete
e94f33cf5d42: Pull complete
Digest: sha256:eb4fdeb3196481dd022d7d166468834e40a046b46a29e3349fb3bf3657176290
Status: Downloaded newer image for registry.gitlab.com/gitlab-org/gitlab-build-images:ruby-2.7.2.patched-golang-1.14-git-2.31-lfs-2.9-chrome-89-node-14.15-yarn-1.22-postgresql-11-graphicsmagick-1.3.36
gcc --version
root@785c57d8e498:/# gcc --version
gcc (Debian 8.3.0-6) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
[...]

PR #8639, opened 2021-05-22
Fixed memory bug: properly root repeated/map field when assigning.

haberman:
Previously the protobuf extension would not properly root memory from a repeated field or map when assigning to a message field (see the attached test case).  If there is no direct reference to the repeated field from Ruby, this will result in a memory error if there is a GC and the repeated field is subsequently accessed (for example, in `#encode`).

The fix is to do the appropriate `upb_arena_fuse()` operation when this assignment is performed. This will cause the Union/Find machinery to ensure that the memory for the repeated field or map lives as long as the message.

The the new test case does not actually crash on my machine, but it does trigger a Valgrind error that I can confirm is fixed by the code in this PR. In the future we should make sure that our CI tests are running tests under Valgrind so we can catch future memory errors of this kind.

Also added an explicit check that `upb_arena_fuse()` succeeded. This is not related to the bugfix, but it is another measure to help ensure that the memory model of the extension is correct.

Fixes: https://github.com/protocolbuffers/protobuf/issues/8559


fowles (approved):

haberman:
JDK failure is unrelated. Merging.

PR #8642, opened 2021-05-22
Fix MinGW compilation

JoelLinn:
Fixed syntax error in platform independent `bswap_64` implementation

### DIFF
--- a/src/google/protobuf/stubs/port.h
+++ b/src/google/protobuf/stubs/port.h
@@ -263,14 +263,14 @@ static inline uint32 bswap_32(uint32 x) {
 
 #ifndef bswap_64
 static inline uint64 bswap_64(uint64 x) {
-  return (((x & uint64_t{0xFFu)) << 56) |
-          ((x & uint64_t{0xFF00u)) << 40) |
-          ((x & uint64_t{0xFF0000u)) << 24) |
-          ((x & uint64_t{0xFF000000u)) << 8) |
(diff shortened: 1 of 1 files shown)


JoelLinn:
I consider `release notes: no` to be appropriate.

src/google/protobuf/stubs/port.h
```diff
@@ -263,14 +263,14 @@ static inline uint32 bswap_32(uint32 x) {
...
-          ((x & uint64_t{0xFF000000000000u)) >> 40) |
-          ((x & uint64_t{0xFF00000000000000u)) >> 56));
+  return (((x & uint64_t(0xFFu)) << 56) |
+          ((x & uint64_t(0xFF00u)) << 40) |
+          ((x & uint64_t(0xFF0000u)) << 24) |
+          ((x & uint64_t(0xFF000000u)) << 8) |
+          ((x & uint64_t(0xFF00000000u)) >> 8) |
+          ((x & uint64_t(0xFF0000000000u)) >> 24) |
+          ((x & uint64_t(0xFF000000000000u)) >> 40) |
+          ((x & uint64_t(0xFF00000000000000u)) >> 56));
```
fowles, line 273: I assume the other direction (using `}` on the right) would also work.  Assuming it does, I slightly prefer that

JoelLinn: done

fowles (approved):

PR #8654, opened 2021-05-25
[Objective-C] Comment Correcting

panmingyang2009:
There is  a mistake in the comment of property 'mapKeyDataType'.
If the field is a map, the value's type is not fieldType.
Instead, it's dataType.

e.g. The following case is a field of Int32-String Map.


panmingyang2009:
@googlebot I signed it!

fowles (approved):

PR #8663, opened 2021-05-26
Add internal.bzl to EXTRA_DIST.

benjaminp:
internal.bzl is required to load java/core/BUILD and java/lite/BUILD, so it must be distributed.


acozzette:
Thanks for the fix!

PR #8665, opened 2021-05-27
Fix OSS benchmarking script

jtattermusch:
There is a [kokoro job](https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fbenchmark%2Fcontinuous) that runs protobuf OSS benchmarks continuously, but that job has silently been broken for a long time
(it appears green on kokoro because the failures aren't being reported correctly, but many of the benchmarks are actually broken).

- removed go benchmark since it's broken and also uses the legacy implementation of protobuf go (and corresponding benchmarks for protocolbuffer/protobuf-go AFAIK don't exist in OSS).
- removed PHP benchmark since it's broken and PHP was making the entire setup complicated (and arguably PHP benchmarks results aren't as important as results for e.g. C++, java, python etc.).
- did a bunch of cleanup and added some todos
- make the script fail if the benchmarks fail, so that breakages are actually noticeable.

Background: I need the protobuf OSS benchmarks to work so that I can compare the results for ARM64 vs x86.

### DIFF
--- a/kokoro/linux/benchmark/run.sh
+++ b/kokoro/linux/benchmark/run.sh
@@ -58,6 +58,7 @@ make clean && make -j8
 
 # build Java protobuf
 pushd java
+# TODO: avoid unnecessarily running all java tests.
 mvn package
 popd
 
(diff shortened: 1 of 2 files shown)


jtattermusch:
Log from a passing adhoc test run: https://fusion2.corp.google.com/invocations/2ff21c35-5e5e-4c23-b45f-e249c80c9a52/log

kokoro/linux/benchmark/run.sh
```diff
@@ -41,64 +45,59 @@ echo "benchmarking pure python..."
...
 make python_add_init
-env LD_LIBRARY_PATH="$oldpwd/src/.libs" python -m util.result_uploader -php_c="../tmp/php_c_result.json"  \
-	-cpp="../tmp/cpp_result.json" -java="../tmp/java_result.json" -go="../tmp/go_result.txt" -python="../tmp/python_result.json" -node="../tmp/node_result.json"
-cd $oldpwd
+env LD_LIBRARY_PATH="${repo_root}/src/.libs" python -m util.result_uploader \
+	-cpp="../tmp/cpp_result.json" -java="../tmp/java_result.json" -python="../tmp/python_result.json" -node="../tmp/node_result.json"
+popd
+
+# also persist the results in the build job log (for better debuggability)
+cat benchmarks/tmp/cpp_result.json
```
fowles, line 100: can you throw some echo's between these to make them more separate in the log?

fowles (approved):

kokoro/linux/benchmark/run.sh
jtattermusch: I can do that in a followup PR if you insist. But note that since "set -ex" is active, the command `cat benchmarks/tmp/cpp_result.json` will be echoed to the log and that seems sufficient
 for log separation.

PR #8666, opened 2021-05-27
OSS benchmarks: print benchmarks JSON results to build job log

jtattermusch:
- uploading the benchmark results to bigquery is currently broken (I think it's because it's trying to upload to bq table in "grpc-testing" project without having access).
- for now, instead of uploading the results, just print the formatted JSON results to the build log (that way the data remains available).


jtattermusch:
Adhoc run of the benchmarks job: https://fusion2.corp.google.com/invocations/3f55344d-c702-4f9b-bd2d-11bed866c801/targets

jtattermusch:
@haberman  FYI.

fowles (approved):

PR #8673, opened 2021-05-27
Correct ordering of export API for DescriptorTable

### LINKED ISSUE #8672: Export macro order is incorrect for C++ attribute usage
Version: v3.17.1, master
Language: C++

Debian Linux (buster/10, bullseye/11)
Windows 10, MSVC 2019 (16.9.3)

GCC 10.2, 11.1, Clang 12

**What did you do?**
Execute `protoc --cpp_out` with `dllexport_decl` set to an application specific value, which results in being preprocessor expanded to a C++ visibility attribute such as `[[gnu::visibility("default")]]`.

The resulting source files fail to compile with GCC and Clang since the attribute is in an invalid location.

Successful compilation of generated C++ library.

**What did you see instead?**
GCC failure ("attribute ignored") when building generated source with `error: attribute ignored [-Werror=attributes]`.

mrjoel:
This resolves issues with GCC and Clang warning (as error) about
attribute in an unallowed location, in particular when used with a
generated file using a dllexport_decl value of [[gnu::visibility]].

The correct ordering is already used in the compiler in
FileGenerator::ForwardDeclarations() for each of classes_, as well
as for other existing local values in pregenerated files. This
change ensures that all usages can compile without issue.

Fixes #8672

### DIFF
--- a/src/google/protobuf/any.pb.h
+++ b/src/google/protobuf/any.pb.h
@@ -52,7 +52,7 @@ struct PROTOBUF_EXPORT TableStruct_google_2fprotobuf_2fany_2eproto {
   static const ::PROTOBUF_NAMESPACE_ID::internal::SerializationTable serialization_table[];
   static const ::PROTOBUF_NAMESPACE_ID::uint32 offsets[];
 };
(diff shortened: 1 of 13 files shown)


mrjoel:
Appropriate tag is C++, I (reasonably so) don't have access to apply the label.

fowles (approved):

PR #8690, opened 2021-06-02
constinit not working as expected in Visual Studio 2019 Update 16.10

### LINKED ISSUE #8688: vs2019 update 16.10 : illegal initialization of `constinit` entity with a non-constant expression

Version: `2021-04-07 version 3.15.8`
Language: C++

Windows.

Visual Studio 2019 msvc **Update 19.10** . Latest version that just recently added support for `constinit` on May 25th - [announcement here](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes#summary-of-whats-new-in-this-release-of-visual-studio-2019-version-16100).

**What did you do?**

Nothing horribly bad.. Steps to reproduce the behavior:
1. Make sure the language set to C++20, `-std:c++latest`, e.g. for CMake that'd be `set(CMAKE_CXX_STANDARD 20)` .
2. Try to compile some code that uses generated protobufs that include `port_def.inc` header .

Code compiles.

**What did you see instead?**

Code fails to compile with errors:

image

**Patch Proposal**

The most trivial hacky patch would be to apply the same logic as applied for `__GNUC__` , the issue described in #8310 .

Adding `!defined(_MSC_VER)` check - "resolves" the issue. And so `constinit` is not used:

```cpp
// Our use of constinit does not yet work with GCC:
// https://github.com/protocolbuffers/protobuf/issues/8310
// Our use of constinit does not yet work with Visual Studio 2019 Update 16.10:
// https://github.com/protocolbuffers/protobuf/issues/TBD
#if defined(__cpp_constinit) && !defined(__GNUC__) && !defined(_MSC_VER)
#define PROTOBUF_CONSTINIT constinit
```

And the code works great after that even with latest Visual Studio 2019 compiler. This could be used as an interim patch to unblock customers using the latest compiler.

maxgolov:
Fixes #8688

Visual Studio 2019 Update 16.10 introduced support for `constinit` for the very first time on May 25th. Unfortunately the way how protobuf uses constinit does not work with Visual Studio msvc compiler. This is a temporary solution to unblock those who updated to latest release of Visual Studio 2019 compiler.

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -508,9 +508,10 @@
 #ifdef PROTOBUF_CONSTINIT
 #error PROTOBUF_CONSTINIT was previously defined
 #endif
-#if defined(__cpp_constinit) && !PROTOBUF_GNUC_MIN(3, 0)
+#if defined(__cpp_constinit) && !PROTOBUF_GNUC_MIN(3, 0) && !defined(_MSC_VER)
 // Our use of constinit does not yet work with GCC:
 // https://github.com/protocolbuffers/protobuf/issues/8310
+// Does not work yet with Visual Studio 2019 Update 16.10
 #define PROTOBUF_CONSTINIT constinit
 #elif __has_cpp_attribute(clang::require_constant_initialization)
 #define PROTOBUF_CONSTINIT [[clang::require_constant_initialization]]


maxgolov:
@googlebot I signed it!

maxgolov:
I don't have permission to add label `release notes: no`. This change is for `c++`, the scope of it is too small for inclusion in release notes.

fowles (approved):

fowles:
Thanks for the patch!  will merge after CI finishes

PR #8702, opened 2021-06-06
Add isKnownType to php Google\Protobuf\Internal\AnyBase

solodkiy:
This is is a simple method to check is Any type could be unpacked or not without trigger exception.

### DIFF
--- a/php/src/Google/Protobuf/Internal/AnyBase.php
+++ b/php/src/Google/Protobuf/Internal/AnyBase.php
@@ -83,4 +83,27 @@ class AnyBase extends \Google\Protobuf\Internal\Message
         $type_url = GPBUtil::TYPE_URL_PREFIX . $fully_qualifed_name;
         return $this->type_url === $type_url;
     }
+
+    /**
+     * This method returns whether the type_url in any_message is known by pool
+     * and could be unpack
+     * @return bool
+     */
+    public function isKnownType()
+    {
+        // Get fully qualified name from type url.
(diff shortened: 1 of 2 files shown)


solodkiy:
@googlebot I signed it!

TeBoring:
I think users of Any should make sure the descriptor is known. Providing `isKnown` seems to change this assumption, in which user can intentionally handle unknown types in runtime.

solodkiy:
I see something like this as use case:
```
message QueueMessage {
  string from = 1;
  google.protobuf.Timestamp fired_at = 2;
  google.protobuf.Any message_data = 3;
}
```

You could send any message in this format through kafka, but some of consumers could not be update to latest version and don't know some of send messages.
So I assume some code like 
```
if (!$queueMessage->getMessageData()->isKnownType()) {
  // skip unknown message
}
```
could be helpful in this case.

haberman:
Why not just attempt to unpack, and catch the exception if the type is unknown?

```php
try {
    $unpacked = $queueMessage->getMessageData()->unpack();
} catch (Exception $e) {
    // Unknown type, skip.
}
```

haberman:
Why not just attempt to unpack, and catch the exception if the type is unknown?

```php
try {
    $unpacked = $queueMessage->getMessageData()->unpack();
} catch (Exception $e) {
    // Unknown type, skip.
}
```

acozzette:
@solodkiy I'm going to close this since it's been a few weeks with no action, but feel free to comment here if you would like to reopen this.

PR #8703, opened 2021-06-06
fix php Google\Protobuf\Internal\AnyBase::is for not message classes

solodkiy:
Fix `Call to a member function getFullName() on null` bug

```
PHPUnit 8.5.16 by Sebastian Bergmann and contributors.

..E....................                                           23 / 23 (100%)

Time: 84 ms, Memory: 6.00 MB

There was 1 error:

1) WellKnownTest::testAny
Error: Call to a member function getFullName() on null

/home/doc/projects/_forks/protobuf/php/src/Google/Protobuf/Internal/AnyBase.php:82
/home/doc/projects/_forks/protobuf/php/tests/WellKnownTest.php:87

```

### DIFF
--- a/php/src/Google/Protobuf/Internal/AnyBase.php
+++ b/php/src/Google/Protobuf/Internal/AnyBase.php
@@ -79,6 +79,9 @@ class AnyBase extends \Google\Protobuf\Internal\Message
     {
         $pool = \Google\Protobuf\Internal\DescriptorPool::getGeneratedPool();
         $desc = $pool->getDescriptorByClassName($klass);
+        if (is_null($desc)) {
+            return false;
+        }
         $fully_qualifed_name = $desc->getFullName();
         $type_url = GPBUtil::TYPE_URL_PREFIX . $fully_qualifed_name;
         return $this->type_url === $type_url;
--- a/php/tests/WellKnownTest.php
+++ b/php/tests/WellKnownTest.php
@@ -84,6 +84,7 @@ class WellKnownTest extends TestBase {
         // Test is.
         $this->assertTrue($any->is(TestMessage::class));
         $this->assertFalse($any->is(Any::class));
+        $this->assertFalse($any->is(NotMessage::class));
     }
 
     public function testAnyUnpackInvalidTypeUrl()


solodkiy:
Any updates?

TeBoring:
This is an invalid use case. You have to check proto message for Any.

haberman:
If there is an easy fix to avoid a PHP crash (and this PR shows that there is), we should make it gracefully return false. 

I don't see the benefit of forcing a PHP crash here.

php/tests/WellKnownTest.php
```diff
@@ -84,6 +84,7 @@ public function testAny()
         // Test is.
         $this->assertTrue($any->is(TestMessage::class));
         $this->assertFalse($any->is(Any::class));
+        $this->assertFalse($any->is(NotMessage::class));
```
TeBoring, line 87: I think this should throw error, not just returning false here.

TeBoring: Maybe also test a message not in the generated pool.

php/src/Google/Protobuf/Internal/AnyBase.php
```diff
@@ -79,6 +79,9 @@ public function is($klass)
     {
         $pool = \Google\Protobuf\Internal\DescriptorPool::getGeneratedPool();
         $desc = $pool->getDescriptorByClassName($klass);
+        if (is_null($desc)) {
+            return false;
```
TeBoring, line 83: Log "The class is not a proto message or hasn't been instantiated yet. Make sure to instantiate this proto class first." and throw error.

php/tests/WellKnownTest.php
haberman: I agree with @TeBoring, I think both the C extension and the PHP library should add an error message.

If we just return `false`, there is a risk that the function will return `false` when it should have returned `true` if a given message class has not been constructed yet.

I don't think this change will break anybody.

acozzette:
@solodkiy I'm going to close this since it's been a few weeks with no action, but feel free to comment here if you would like to reopen this.

Issue #8316, opened 2021-02-19
[c#] Parser.WithDiscardUnknownFields throws exception with unknown field


acurrens:
**What version of protobuf and what language are you using?**
Version: 3.15.0
Language: C#

**What operating system (Linux, Windows, ...) and version?**
Windows

**What runtime / compiler are you using (e.g., python version or gcc version)**
* Visual Studio 2019
* .net Core 5.0
* Grpc.AspNetCore (v2.34.0)
* Grpc.Tools (2.36.0-pre1)
* Google.Protobuf (3.15.0)

**What did you do?**
Parse json object with GrpcService Parser.
`var parser = CarList.Parser.WithDiscardUnknownFields(true);`
` var cList = parser.ParseJson(newCarJson);`

Knowing that there are extra fields in the json objects that are not defined in the proto file I attempted to use the WithDiscardUnknownFields option set to true.

**What did you expect to see**
Grpc message of repeated objects derived from the parsed json.

**What did you see instead?**
An Exception when the ParseJson() method was called:`"Unknown field: <field name>"`

It seems like the Message parser is always using the default settings and not considering the ignoreUnknown flag:
https://github.com/protocolbuffers/protobuf/blob/133e5e75263be696c06599ab97614a1e1e6d9c66/csharp/src/Google.Protobuf/MessageParser.cs#L353-L358

**Anything else we should know about your project / environment**
I've setup a simple demo project illustrating the issue in my git repo here:
https://github.com/acurrens/GrpcJsonParseIssue

Feel free to clone/review the demo project and let me know if I'm doing something wrong on my end
Thanks,
Andrew

kwaclaw:
I can confirm this

TheFireCookie:
I encounter the same issue too.

shaneqld:
Can reproduce and it's clear looking at the code that this has been forgotten.  If you want a work around in the meantime you can do something like this:

```cs
public static T ParseJson<T>(string json, bool discardUnknownFields = true) where T : IMessage<T>, new()
{
    var jp = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(discardUnknownFields));
    T msg = jp.Parse<T>(json);
    return msg;
}
```

jskeet:
I don't think we'd want to change the actual behavior here (it could be a very significantly breaking change), but we can certainly clarify the documentation. I'll create a PR for that.

jskeet:
[...]

PR #8706, opened 2021-06-07
fix(chsarp): actually discard unknown fields when parsing JSON via MessageParser<T>

shaneqld:
As per #8316, unknown fields are not discarded when using `MessageParser<T>.WithDiscardUnknownFields(true).ParseJson()`

This PR corrects the behavior of JSON parsing using MessageParser when discarding unknown fields is requested.

This also includes some unit tests which are derived from unit tests in JsonParserTest.

Just a note: I created a new file MessageParserTest.cs, as there were no tests for this class. I copied the copyright header from another file - not sure if this is right :)

### DIFF
--- a/csharp/src/Google.Protobuf/MessageParser.cs
+++ b/csharp/src/Google.Protobuf/MessageParser.cs
@@ -379,7 +379,16 @@ namespace Google.Protobuf
         public new T ParseJson(string json)
         {
             T message = factory();
-            JsonParser.Default.Merge(message, json);
+            JsonParser parser;
+            if (DiscardUnknownFields)
+            {
+                parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(true));
+            }
+            else
+            {
+                parser = JsonParser.Default;
+            }
+            parser.Merge(message, json);
             return message;
         }
 
--- a/csharp/src/Google.Protobuf.Test/MessageParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/MessageParserTest.cs
@@ -46,7 +46,7 @@ namespace Google.Protobuf
     public class MessageParserTest
     {
         [Test]
-        public void JsonUnknownField_NotIgnored()
+        public void ParseJsonUnknownField_NotIgnored()
         {
             var parser = new MessageParser<TestAllTypes>(() => new TestAllTypes());
             string json = "{ \"unknownField\": 10, \"singleString\": \"x\" }";
@@ -58,7 +58,7 @@ namespace Google.Protobuf
         [TestCase("\"text\"")]
         [TestCase("[0, 1, 2]")]
         [TestCase("{ \"a\": { \"b\": 10 } }")]
-        public void UnknownField_Ignored(string value)
+        public void ParseJsonUnknownField_Ignored(string value)
         {
             var parser = new MessageParser<TestAllTypes>(() => new TestAllTypes()).WithDiscardUnknownFields(true);
             string json = "{ \"unknownField\": " + value + ", \"singleString\": \"x\" }";
--- a/csharp/src/Google.Protobuf.Test/MessageParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/MessageParserTest.cs
@@ -1,0 +1,70 @@
+﻿#region Copyright notice and license
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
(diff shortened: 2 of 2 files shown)


shaneqld:
@googlebot I fixed it.

kwaclaw:
I would be interested in this fix. I think an additional static instance is a good way to go.

csharp/src/Google.Protobuf/MessageParser.cs
```diff
@@ -379,7 +379,16 @@ public new T ParseFrom(CodedInputStream input)
         public new T ParseJson(string json)
```
acozzette, line 379: Does the other `ParseJson` method on line 179 need this same fix?

jtattermusch:
CC @amanda-tarafa @jskeet

jskeet:
I'm nervous of this as a breaking change. I've always regarded `ParseJson` as a convenience method to provide the relevant descriptor, but to use the default JSON parser.

I *think* I'd prefer a change which simply documented the existing behavior more clearly, both for ParseJson and WithDiscardedFields. We could always add another method to parse the JSON taking any other parser settings into account (without being specific about it being just discarded fields - there may be other settings we want to propagate in the future).

jskeet:
That said, if there's precedent for this in other languages, that would sway me. I don't *think* the C# design here is directly comparable with that in other languages, but that's based on dim and distant memories.

shaneqld:
> I've always regarded ParseJson as a convenience method to provide the relevant descriptor, but to use the default JSON parser.

As a user, I believed the opposite. In protoc generated C# types for messages, each type contains a `MessageParser<T> Parser` property, which is easily discoverable when asking "how do I deserialize this message?"

From reading the proto3 Language Guide, the Google API Design Guide and the AIPs, I had the opinion that backwards compatibility is a key tenant. Which is why I feel comfortable writing the following, as I know it won't throw an exception.

```cs
var fromProtobuf = TestMessage.Parser.ParseFrom(protobufWithUnknownField);
```

But when working with JSON, the following code does, in fact, throw an exception:

```cs
var fromJson = TestMessage.Parser.ParseJson(jsonWithUnknownField);
```

This became an issue in our system as we were sending JSON-serialized webhooks of our proto messages. We introduced additional "non-breaking" fields into those messages, but customers were deserializing with the above ParseJson code, causing their code to break.

It would be nice if both methods above were safe to use when encountering messages serialized with a newer, wire-compatible version of the type, perhaps by changing `JsonParser.Settings.Default` to ignore unknown fields.

> I think I'd prefer a change which simply documented the existing behavior more clearly, both for ParseJson and WithDiscardedFields.

I think this is definitely a good idea if the plan is not to change the behaviour. But in terms of the behaviour, it just seems odd that you can write:

```cs
var fromJson = TestMessage.Parser.WithDiscardUnknownFields(true).ParseJson(jsonWithUnknownField);
```

and receive the exception: `Google.Protobuf.InvalidProtocolBufferException: 'Unknown field: xx'`

> That said, if there's precedent for this in other languages, that would sway me.

I set up a simple test Go program to see how this works.

```protobuf
syntax = "proto3";
package testing.tests.testmessagespb;
option go_package = "github.com/testing/tests/testmessagespb";

message TestMessage {
  int32 first = 1;
}
```

```go
package main

import "fmt"
import "github.com/testing/tests/testmessagespb"
import "google.golang.org/protobuf/encoding/protojson"

func main() {
    json := "{\"first\":20, \"second\": 50}"
    fmt.Println("Input JSON:    " + json)

    deserialized := &testmessagespb.TestMessage{}
    protojson.Unmarshal([]byte(json), deserialized)

    reserialized, _ := protojson.Marshal(deserialized)
    fmt.Println("Re-serialized: " + string(reserialized))
}
```

This deserializes a JSON-serialized instance of `TestMessage`, but the JSON has an unknown field `second`. It deserializes the JSON without failing, by default.  The output of the program is:

```
Input JSON:    {"first":20, "second": 50}
Re-serialized: {"first":20}
```

csharp/src/Google.Protobuf/MessageParser.cs
shaneqld: It looks like it does. Will wait for the outcome of the discussion and then fix as required.

jskeet:
> It would be nice if both methods above were safe to use when encountering messages serialized with a newer, wire-compatible version of the type, perhaps by changing JsonParser.Settings.Default to ignore unknown fields.

That feels like a *very* big change to me - and one that we'd definitely want to make for all languages at the same time, if at all. Changing the behavior universally feels much less likely than making the targeted change to *just* affect how fields are discarded via MessageParser.ParseJson.

I'm actually surprised at the Go behavior - I *thought* that all languages were "default to being strict, opt in to being lenient". It's worth noting that there's quite a significant difference between parsing JSON and parsing the binary wire format in terms of unknown fields, and I'm somewhat reluctant to mix the two. (In the current JSON parser - at least for C# and I believe all other languages - there's no such concept as "ignore unknown fields but retain them so they can be reproduced later on".

The other thing it's worth being aware of is that even when unknown *fields* are ignored, unknown enum *values* can still cause problems, depending on your language. We're discussing that internally, but I wanted to highlight it now.

jskeet:
Closing this now, as we updated the documentation in #10340 which closed #8316.

PR #8707, opened 2021-06-08
Make StringPiece constructible from std::string_view

georgthegreat:
This improves interop with certain string-like data structures, allowing to save extra allocation.

### DIFF
--- a/src/google/protobuf/stubs/stringpiece.h
+++ b/src/google/protobuf/stubs/stringpiece.h
@@ -214,6 +214,13 @@ class PROTOBUF_EXPORT StringPiece {
       : ptr_(str.data()), length_(0) {
     length_ = CheckSize(str.size());
   }
+ 
+  StringPiece(  // NOLINT(runtime/explicit)
+      std::string_view str)
+      : ptr_(str.data()), length_(0) {
+    length_ = CheckSize(str.size());
+  }
+   
 
   StringPiece(const char* offset, size_type len)
       : ptr_(offset), length_(CheckSize(len)) {}
--- a/src/google/protobuf/stubs/stringpiece.h
+++ b/src/google/protobuf/stubs/stringpiece.h
@@ -148,6 +148,10 @@
 #include <limits>
 #include <string>
 
+#if defined(__cpp_lib_string_view)
+#include <string_view>
+#endif
+
 #include <google/protobuf/stubs/hash.h>
 
 #include <google/protobuf/port_def.inc>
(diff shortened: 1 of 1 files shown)


src/google/protobuf/stubs/stringpiece.h
```diff
@@ -214,6 +214,13 @@ class PROTOBUF_EXPORT StringPiece {
       : ptr_(str.data()), length_(0) {
```
georgthegreat, line 214: I tried to fix this one by replacing const string& with std::string_view, but certain cases were broken.

I. e. HasSuffixString(const string&, const string&) stops working as it requires 2 implicit ctors (and the standard only allows one).

This could be simplified with C++20 starts_with / ends_with, but this is a long story.

src/google/protobuf/stubs/stringpiece.h
```diff
@@ -214,6 +214,13 @@ class PROTOBUF_EXPORT StringPiece {
...
     length_ = CheckSize(str.size());
   }
+ 
+  StringPiece(  // NOLINT(runtime/explicit)
```
acozzette, line 218: I'd have to think some more about whether this change will be OK. Eventually we want to switch to `absl::string_view` but adding a dependency on absl is not entirely trivial and we haven't had time to work on it much yet. I think doing this as a stopgap measure is probably OK, but it needs to be guarded with an ifdef so that it only appears for C++17 and higher (we still support C++11 so we can't unconditionally rely on std::string_view yet).

jorgbrown: I agree.  We do something like this already, but we use a guard that checks the C++ feature macro for string_view.  (See https://github.com/protocolbuffers/protobuf/blob/6051341ab21b4e1f75e91710ab1a8cc8fadda521/src/google/protobuf/map.h )

That is, this should be guarded with
`#if defined(__cpp_lib_string_view)`

And at the top of the file, the string_view header should be included:

```
#if defined(__cpp_lib_string_view)
#include <string_view>
#endif
```

georgthegreat: Sorry, I am using string_view for so long by now.
I though it was introduced in C++11...

Should be fixed now.

georgthegreat: NB: as I see, abseil-cpp [falls back](https://github.com/abseil/abseil-cpp/blob/f72972654b69085b1d58fde04c618b77d3ac6935/absl/strings/string_view.h#L46) to std::string_view if it is available.

I am not sure about the logic behind `ABSL_USES_STD_STRING_VIEW` define, not I am sure on how abseil is going to add starts_with / ends_with from C++20 in the fallback case.

PR #8729, opened 2021-06-11
Add basic KVO test.

dmaclach:
Add a simple KVO test just to verify that KVO continues to function for protobufs.

### DIFF
--- a/objectivec/Tests/GPBMessageTests.m
+++ b/objectivec/Tests/GPBMessageTests.m
@@ -41,6 +41,52 @@
 #import "google/protobuf/Unittest.pbobjc.h"
 #import "google/protobuf/UnittestObjc.pbobjc.h"
 #import "google/protobuf/UnittestObjcOptions.pbobjc.h"
+#import "google/protobuf/UnittestImport.pbobjc.h"
+
+// Helper class to test KVO.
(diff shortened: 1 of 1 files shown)


thomasvl (approved):

objectivec/Tests/GPBMessageTests.m
```diff
@@ -431,6 +474,44 @@ - (void)testKVC_SetValue_ForKey {
...
+ GPBKVOTestObserver *innerFieldObserver =
+     [[[GPBKVOTestObserver alloc] initWithObservee:message
+                                           keyPath:@"optionalImportMessage.d"]
+      autorelease];
+  XCTAssertFalse(autocreateObserver.didObserve);
+  XCTAssertFalse(innerFieldObserver.didObserve);
+  message.optionalImportMessage.d = 2;
+
+  // Autocreation of fields is not observed by KVO.
+  XCTAssertFalse(autocreateObserver.didObserve);
```
thomasvl, line 505: So it sorta feels like this should have fired, after all the field got set.

You should likely also first just fetch the value of `d` (i.e. - get the default value), and ensure that doesn't trigger either of them.

thomasvl (approved):

Issue #7074, opened 2020-01-10
Error in StringPrintfTest.Multibyte with Visual Studio 2019 16.4.1 x64


emmenlau:
When running the tests I get a reproducible assertion in StringPrintfTest.Multibyte on a  Visual Studio 2019 16.4.1 x64 build. All other platforms (Ubuntu 18.04, CentOS 7.7, macOS XCode 9, Windows Visual Studio 2017.9) work fine with the same build instructions.

Version: v3.11.2
Windows 10, Visual Studio 2019 16.4.1 x64
CMake 3.16.1, Google Test 1.10.0

```
[...]
[ RUN      ] StringPrintfTest.Multibyte
1067 minkernel\crts\ucrt\src\appcrt\convert\mbstowcs.cpp(246) : Assertion failed: (pwcs == nullptr && sizeInWords == 0) || (pwcs != nullptr && sizeInWords > 0)
```

emmenlau:
Anyone?

MCESoftwareLT:
I get the same Assertion. 
protobuf-3.11.4
Visual Studio 2019 16.5.2 x86
CMake 3.17

NouberNou:
I get the same assertion in VS 2015 on 3.2.0 and 3.3.0.

Did not get it in 2017.

emmenlau:
I can still reproduce this issue with protobuf 3.13.0 in Visual Studio 2019 16.8 x64:
```

 [ RUN      ] StringPrintfTest.Multibyte
 minkernel\crts\ucrt\src\appcrt\convert\mbstowcs.cpp(246) : Assertion failed: (pwcs == nullptr && sizeInWords == 0) || (pwcs != nullptr && sizeInWords > 0)
```

Could this be looked at? It seems a bit scary if multi-byte character conversions are potentially broken...

PR #8735, opened 2021-06-15
Fix #7047 Safely handle setlocale

### LINKED ISSUE #7047: Feature Request: Swift Package Manager support

florin-crisan:
`setlocale` returns a pointer to a buffer containing the current locale name.
This needs to be copied into a `std::string` or it will be overwritten by the next call.

Trying to call `setlocale` with a non-null, invalid pointer can have unpredictable results, such as https://github.com/protocolbuffers/protobuf/issues/7074
```
[ RUN      ] StringPrintfTest.Multibyte
minkernel\crts\ucrt\src\appcrt\convert\mbstowcs.cpp(246) : Assertion failed: (pwcs == nullptr && sizeInWords == 0) || (pwcs != nullptr && sizeInWords > 0)
```

`setlocale` can also return a `nullptr` if it fails, but we assert against that.

### DIFF
--- a/src/google/protobuf/stubs/stringprintf_unittest.cc
+++ b/src/google/protobuf/stubs/stringprintf_unittest.cc
@@ -91,7 +91,9 @@ TEST(StringPrintfTest, Multibyte) {
   // out of memory while trying to determine destination buffer size.
   // see b/4194543.
 
-  char* old_locale = setlocale(LC_CTYPE, nullptr);
+  char* old_locale_c = setlocale(LC_CTYPE, nullptr);
+  ASSERT_TRUE(old_locale_c != nullptr);
+  std::string old_locale = old_locale_c;
   // Push locale with multibyte mode
   setlocale(LC_CTYPE, "en_US.utf8");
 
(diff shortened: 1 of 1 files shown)


acozzette:
@florin-crisan It looks like your pull request updates the googletest submodule and that is causing some tests to fail. Would you mind updating the PR to avoid changing googletest?

florin-crisan:
> @florin-crisan It looks like your pull request updates the googletest submodule and that is causing some tests to fail. Would you mind updating the PR to avoid changing googletest?

As explained on ee0e1535ec559ffca410c868e9194b3ac2a0bd17, the whole reason for the update was that I couldn't compile with Visual C++ 2019.

But I will try it your way and let somebody else deal with the update, if it's that tricky.

acozzette:
Thanks, @florin-crisan! I think it would be OK to update googletest in another pull request but we would just need to make sure all the tests still pass.

PR #8736, opened 2021-06-15
Add option to use external GTest in CMake

florin-crisan:
If `-Dprotobuf_USE_EXTERNAL_GTEST=ON` is passed to CMake, it will use an external Google Test copy (i.e. by calling `find_package(GTest REQUIRED)`) rather than the one provided as a submodule.

This makes sense for larger projects that might already include Google Test and want to use a more standard CMake approach.

Also updated build instructions with this information, and more idiomatic usage.


acozzette:
Thanks, @florin-crisan!

PR #8737, opened 2021-06-15
Enable testing via CTest

florin-crisan:
Convenience feature: enable users to test via the familiar `ctest` command rather than making the `check` target.

They would be able to use the familiar CMake pattern:
```
cmake -S source/protobuf -B build/protobuf ...
cmake --build build/protobuf
ctest --test-dir build/protobuf
cmake --install build/protobuf
```

This is a follow-up to 9f447fc9d3da93da29b8301f1a8ca57b1ea812d7


acozzette:
Thanks, @florin-crisan. Sorry for the long delay in reviewing this!

PR #8738, opened 2021-06-15
ObjC: Block deprecated warnings when testing on newer OS

dmaclach:
Some of the test methods we use are deprecated on newer OS. This just gets prevents the
warning that we don't actually care about.


thomasvl:
What things get warnings?

thomasvl (approved):

PR #8739, opened 2021-06-15
ObjC: Add option to remove property metadata from message classes to reduce codegen

dmaclach:
Adds an option to protoc `--objc_opt=elide_message_metadata` to remove all the property
metadata from message classes. This significantly reduces the codegen size of the clases.
The downside is that iterating through properties using objective c runtime calls will no
longer function. This is mitigated by the fact that most (all?) of the information that
folks are interested in can be extracted via the message descriptor.

We do this by defining our own classes using the `GPB_MESSAGE_SUBCLASS_IMPL` macro.

The `GPB_MESSAGE_SUBCLASS_IMPL` macro generates code equivalent to:
```
@implementation _name_
+ (GPBDescriptor *)descriptor {
    return _descriptorFunc_(self, _cmd);
}
@end
```
We do this to avoid all of the @property metadata. If we use `@dynamic` for the properties the compiler generates a lot of property data including selectors and encoding strings. For large uses of protobufs, this can add up to several megabytes of unused Objective-C metadata.
This inline class definition avoids the property data generation at the cost of being a little ugly. This has been tested with both 32 and 64 bits on intel and arm with Xcode 11.7 and Xcode 12.4.


objectivec/GPBMessage_PackagePrivate.h
```diff
@@ -42,6 +42,103 @@
...
+    ".section __TEXT, __objc_methname, cstring_literals \n"        \
+    "aDescriptor: .asciz \"descriptor\" \n"                        \
+    ".endif \n"                                                    \
+                                                                   \
+    ".ifndef a1608 \n"                                             \
+    ".section __TEXT, __objc_methtype, cstring_literals \n"        \
+    "a1608: .asciz \"@16@0:8\" \n"                                 \
+    ".endif \n"                                                    \
+  )
+
```
thomasvl, line 141: Should/Can we undef the CPP macros that are impl details inside the main one?  Should we name them in any way to make sure folks don't use them?

objectivec/GPBMessage_PackagePrivate.h
```diff
@@ -42,6 +42,103 @@
...
+#define GTP_ALIGN_SIZE " 3 "
+#else  // __LP64__
+#define GPB_ASM_PTR " .long "
+#define GPB_ASM_PTRSIZE " 4 "
+#define GTP_ASM_ONLY_LP64(x)
+#define GTP_ALIGN_SIZE " 2 "
+#endif  // __LP64__
+
+#define GPB_MESSAGE_CLASS_NAME_RAW GPBMessage
+#define GPB_STRINGIFY_INTERNAL(x) #x
```
thomasvl, line 58: Don't we have two stringify defines already at the top of this file?  Can we use those or do we need unique ones again?

objectivec/GPBMessage_PackagePrivate.h
```diff
@@ -42,6 +42,103 @@
...
+#define GPB_STRINGIFY(x) GPB_STRINGIFY_INTERNAL(x)
+#define GPB_MESSAGE_CLASS_NAME GPB_STRINGIFY(GPB_MESSAGE_CLASS_NAME_RAW)
+
+// This generates code
```
thomasvl, line 62: Likely needs a more completely comment about what all this does and why.

Is there anything we can do detect if this will break in a future tools seed?

thomasvl (changes requested):
Please expand the PR description and commit comment to better capture information for extraction into the release notes.

objectivec/GPBMessage_PackagePrivate.h
dmaclach: We can't undef them because they need to be defined when the macro expands in-place. I figured that the GPB names (and hiding them in Private) was sufficient deterrent.

objectivec/GPBMessage_PackagePrivate.h
dmaclach: Not at the top of this file, but I found them in GPBUtilities. Using those now.

objectivec/GPBMessage_PackagePrivate.h
```diff
@@ -42,6 +44,125 @@
...
+#else
+// 0x80 denotes that the class supports ARC.
+#define GPB_CLASS_FLAGS " 0x80 "
+#define GPB_METACLASS_FLAGS " 0x81 "
+#endif
+
+// This generates code equivalent to:
+// ```
+// @implementation _name_
+// + (GPBMessage*)descriptor {
```
thomasvl, line 76: `+descriptor` returns `GPBDescriptor*`, not `GPBMessage*`

objectivec/GPBMessage_PackagePrivate.h
```diff
@@ -42,6 +44,125 @@
...
+//      return _descriptorFunc_(self, _cmd);
+// }
+// @end
+// ```
+// We do this to avoid all of the @property metadata.
+// Note that for Xcode < 12 having the cstring_literals be local definitions
+// is required to avoid a "malformed method list" assertion from the linker.
+// We do this by starting them with "L":
+// https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_5.html#SEC48
+#define GPB_MESSAGE_SUBCLASS_IMPL(name, descriptorFunc)                        \
```
thomasvl, line 86: Is it worth listed all the places explicitly to allow for cleanup eventually?

objectivec/GPBMessage_PackagePrivate.h
dmaclach: Fixed. Thanks

objectivec/GPBMessage_PackagePrivate.h
dmaclach: Not sure what cleanup you mean? Xcode 12.5 declares them local as well (i.e. this is correct). The linker with 11.7 is just pickier about it. I'll update the comment.

dmaclach: There.. see if that is clearer.

thomasvl: Does 13 declare them as local also?

The current comment is written such that it sorta implies they only have to be local because of the assertion, not from a correctness pov.  If correctness is local, then I wouldn't both with the version comment at all.

dmaclach:
> Please expand the PR description and commit comment to better capture information for extraction into the release notes.

Done. Let me know if you think we need more. Happy to do so.

dmaclach:
Made it an option. iOS and macOS projects now generate and test both code paths.

src/google/protobuf/compiler/objectivec/objectivec_message.cc
```diff
@@ -597,18 +611,25 @@ void MessageGenerator::GenerateSource(io::Printer* printer) {
...
 
-    if (!deprecated_attribute_.empty()) {
+    if (elide_message_metadata_) {
       printer->Print(
-          "#pragma clang diagnostic pop\n"
-          "\n");
+          "GPB_MESSAGE_SUBCLASS_IMPL($classname$, $classname$_descriptor);\n\n",
+          "classname", class_name_);
+    } else {
+    printer->Print("@end\n\n");
```
thomasvl, line 627: fix index.

dmaclach: Done

thomasvl (approved):

PR #8740, opened 2021-06-16
[csharp] ByteString.CreateCodedInput should use ArraySegment offset and count

kiwidev:
CreateCodedInput is created from the underlying array behind the ByteString.
If this was created from a larger array (via Memory<byte> or ArrayPool etc)
then the CodedInput refers to the wrong section of memory.

Change is to add the offset and count like the other methods that use the
ArraySegment (ToString, ToBase64, WriteTo).

Fixup for https://github.com/protocolbuffers/protobuf/pull/7645

### DIFF
--- a/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
+++ b/csharp/src/Google.Protobuf.Test/ByteStringTest.cs
@@ -211,6 +211,22 @@ namespace Google.Protobuf
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


JamesNK (approved):

JamesNK:
@jtattermusch

jtattermusch (approved):
LGTM.

kiwidev:
Thanks! 👍

PR #8741, opened 2021-06-17
Stop using std::iterator (deprecated in C++17).

### LINKED ISSUE #5416: vs2017 15.9.3 c++ 17 标准编译不通过
protobuf 版本：3.6.1

错误提示： C4996: 'std::iterator<std::random_access_iterator_tag,Element,ptrdiff_t,_Ty*,_Ty&>': warning STL4015: The std::iterator class template (used as a base class to provide typedefs) is deprecated in C++17. (The <iterator> header is NOT deprecated.) The C++ Standard has never required user-defined iterators to derive from std::iterator. To fix this warning, stop deriving from std::iterator and start providing publicly accessible typedefs named iterator_category, value_type, difference_type, pointer, and reference. Note that value_type is required to be non-const, even for constant iterators. You can define _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.

### LINKED ISSUE #7120: Do not inherit from std::iterator
C++

When including protocol buffers in my project and compiling with Microsoft's compiler (Microsoft (R) C/C++ Optimizing Compiler Version 19.22.27905 for x64), the protocol buffers headers cause the compiler to emit warning `C4996: Your code uses a function, class member, variable, or typedef that's marked deprecated`.

The origins of this diagnostic are the classes  in `repeated_field.h` that inherit from `std::iterator`, a class which is deprecated in c++17.  There are several other files which appear to contain classes that inherit from `std::iterator` as well.  

Do not use the deprecated feature, so that the diagnostic is not emitted.

The warning can be silenced by defining `_SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING`, but it would be nice to keep that diagnostic on.  

**Additional context**
Originally I was working with protocol buffers 3.7.0, but this problem exists with version 3.11.2

Quuxplusone:
Fixes #7120
Fixes #5416 (maybe?)

Weirdly, the changelog claims that this was already fixed in
```
2019-04-29 version 3.8.0
[...]
  * [c++17] Changed proto2::RepeatedPtrField iterators to no longer derive from the deprecated std::iterator class.
```

### DIFF
--- a/src/google/protobuf/reflection.h
+++ b/src/google/protobuf/reflection.h
@@ -392,13 +392,18 @@ class PROTOBUF_EXPORT RepeatedFieldAccessor {
 
 // Implement (Mutable)RepeatedFieldRef::iterator
 template <typename T>
-class RepeatedFieldRefIterator
-    : public std::iterator<std::forward_iterator_tag, T> {
+class RepeatedFieldRefIterator {
   typedef typename RefTypeTraits<T>::AccessorValueType AccessorValueType;
   typedef typename RefTypeTraits<T>::IteratorValueType IteratorValueType;
   typedef typename RefTypeTraits<T>::IteratorPointerType IteratorPointerType;
 
  public:
+  using iterator_category = std::forward_iterator_tag;
+  using value_type = T;
+  using pointer = T*;
+  using reference = T&;
+  using difference_type = std::ptrdiff_t;
+
   // Constructor for non-message fields.
   RepeatedFieldRefIterator(const void* data,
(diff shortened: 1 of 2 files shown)


Quuxplusone:
@fowles looks like you're the one to poke about C++ reviews? PTAL :)

perezd:
They are not available for the moment, so I've assigned @dlj-NaN who can also review this.

acozzette:
Thanks, @Quuxplusone.

PR #8752, opened 2021-06-21
Create java_export targets

deannagarcia:
Creating java_export targets to all java packages to export the bazel builds as maven consumable packages.


WORKSPACE
```diff
@@ -35,16 +42,24 @@ bind(
     actual = "//util/python:python_headers",
 )
 
+load("@rules_jvm_external//:repositories.bzl", "rules_jvm_external_deps")
```
perezd, line 45: I'm not sure this is needed if we aren't changing the list of standard maven sources, which this addition doesn't appear to be doing. Did you read otherwise somewhere?

WORKSPACE
```diff
@@ -26,6 +26,13 @@ http_archive(
     ],
 )
 
+http_archive(
```
perezd, line 29: What introduced the need for this dependency? If it is actually required, lets move it to proto_deps.bzl and follow the pattern we're using there.

version.bzl
```diff
@@ -0,0 +1 @@
+VERSION = '3.17.3'
```
perezd, line 1: for clarity, can we rename this file protobuf_version.bzl and PROTOBUF_VERSION as the var name? This makes it unambiguous in any context.

perezd, line 1: Also, are there any runbooks/instructions we need to update to ensure this is updated properly when the current release process is running? @acozzette might have some ideas here.

WORKSPACE
deannagarcia: Removed this dependency since we won't be using the publish command.

WORKSPACE
deannagarcia: Removed this dependency.

version.bzl
deannagarcia: Fixed!

perezd (approved):
LGTM!

version.bzl
deannagarcia: I figured out how to do this and will create another PR to edit the update version script.

PR #8758, opened 2021-06-23
Support arena allocation of google::protobuf::AnyMetadata

### LINKED ISSUE #8264: (C++) memory leak in PackFrom() on arena-allocated Any objects
Version: 3.14.0, 3.15.8
Language: C++
The protobuf library has been compiled with the following command line:
```
cmake -G "Visual Studio 15 2017 Win64" -Dprotobuf_BUILD_SHARED_LIBS=ON ..
```

Windows 10 64bit

Microsoft Visual Studio Professional 2017 15.9.24

**What did you do?**
With the following proto definition (my_message.proto):

```
syntax = "proto3";

message MyMessage {
}
```

and the following test program:

```
#include "my_message.pb.h"
#include <google/protobuf/arena.h>
#include <google/protobuf/message.h>
#include <google/protobuf/any.pb.h>
#include <iostream>

std::uint64_t leak()
{
  google::protobuf::Arena arena;

  google::protobuf::Any* any_msg = google::protobuf::Arena::CreateMessage<google::protobuf::Any>(&arena);
  MyMessage my_msg;

  any_msg->PackFrom(my_msg);    //  <--  this causes the memory leak

  return arena.SpaceUsed();   // [1]
}

int main()
{
  std::uint64_t max_size = 0;
  std::uint16_t n = 0;

  while (true) {
    const std::uint64_t size = leak();

    max_size = std::max(size, max_size);
    ++n;
               
    if (n == 0) {
      std::cout << "max = " << max_size << std::endl;
    }
  }

  return 0;
}
```

Memory usage of the program grows indefinitely (observed with Process Explorer and Visual Studio diagnostic tool).  When however the Any object in leak() is allocated without an arena (i.e just with plain `google::protobuf::Any any_msg;`) the memory usage stays constant. 
It makes no difference whether the MyMessage object "my_msg" is arena-allocated or not.

[1] arena.SpaceUsed() always returns the same value, even when the memory usage shown in the diagnostic tool is growing.

georgthegreat:
Fixes #8264


src/google/protobuf/any.pb.cc
```diff
@@ -99,7 +99,7 @@ class Any::_Internal {
 Any::Any(::PROTOBUF_NAMESPACE_ID::Arena* arena,
                          bool is_message_owned)
   : ::PROTOBUF_NAMESPACE_ID::Message(arena, is_message_owned),
-  _any_metadata_(&type_url_, &value_) {
+  _any_metadata_(arena, &type_url_, &value_) {
```
georgthegreat, line 102: NB:

Local regeneration of any.pb.cc from any.proto removes some `PROTOBUF_EXPORT` statements from any.pb.h.
I did not included these changes into PR.

acozzette:
@georgthegreat Thanks for the fix! Looks pretty good but I have just a couple comments:
1. I think we might as well pass `Arena*` as an argument to `AnyMetadata::PackFrom()` instead of storing it as a member variable. `google::protobuf::Any` has an arena member already, so this way we don't have to store it in two places.
2. The fix has to appear in `any_lite.cc` in addition to `any.cc`.

georgthegreat:
@acozzette, this PR should fix the problem in a proper way.

Unfortunately, running `protoc --cpp_out . any.proto` turns any.pb.cc into incompilable code, hence I am unable to check if code generation is ever correct.

Patches in protobuf/compiler might be incorrect.

PR #8765, opened 2021-06-25
Raise DecodeError on attemt to parse invalid bytestream from Python

georgthegreat:
At the time an attempt to parse `proto3` message from `b"data"` will silently exit reporting success.

At least in protobuf=3.4.x this resulted in DecodeError.

### DIFF
--- a/python/google/protobuf/pyext/message.cc
+++ b/python/google/protobuf/pyext/message.cc
@@ -1976,15 +1976,13 @@ static PyObject* MergeFromString(CMessage* self, PyObject* arg) {
   if (ptr == nullptr || ctx.BytesUntilLimit(ptr) < 0) {
     // Parse error or the parser overshoot the limit.
     PyErr_Format(DecodeError_class, "Error parsing message");
-    return NULL;
+    return nullptr;
   }
   // ctx has an explicit limit set (length of string_view), so we have to
   // check we ended at that limit.
   if (!ctx.EndedAtLimit()) {
-    // TODO(jieluo): Raise error and return NULL instead.
-    // b/27494216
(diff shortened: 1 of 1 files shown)


acozzette:
Thank you for the fix but unfortunately I don't think we will be able to accept it. The problem is that we have a large internal customer that relies on being lenient in this case even when the data is technically malformed. We've looked into fixing it in the past, but it's non-trivial because it would require finding the malformed data and re-encoding it.

georgthegreat:
Ok, I got your point.

I am going to perform some tests and will return with some better solution.

georgthegreat:
@acozzette, I carried out some testing. At the time there is no way to get parsing status if the parsing was invoked via static `Message.FromString()`:

```
msg = Message.FromString(raw_input)
assert msg.IsInitialized()
```

It raises when Message is invalid due to the lack of required fields, but silently returns upon attempt to parse invalid / incomplete bytestream.

This leads to a necessity to perform double checks every time when message is being parsed (assume that you are not patching a trusted bytestream which is quite common):

Due to this being totally un-pythonic (also, due to lack of `[[nodiscard]]` attribute in Python), I believe there are tons of open-source code which does not properly perform this check.

I assume that it is impossible to change behavior inside Google. Is it possible to introduce some preprocessor variable responsible to distinct wrether we are running an opensource version of protobuf?

acozzette:
Actually it might even be OK to go ahead and make this change while within Google we can perhaps maintain our internal patch. @anandolee What do you think?

anandolee:
I will check with our internal customer to see if they also depend on our opensource version.

acozzette:
Let me go ahead and close this. @anandolee has a plan to fix the bug while safely migrating our internal customer.

PR #8778, opened 2021-06-29
Allow suppression of objc prefix checks.

thomasvl:
- Accept "-" as the expected prefixes file and turn off all validations of
  prefixes (even the Apple conventions).
- Reflow some of the prefix checks to hopefully make things a little easier
  to follow.
- Don't print warnings about updates to the expected prefix file until there is
  a file.

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -1234,6 +1234,8 @@ bool ValidateObjCClassPrefix(
   // without any prefix at all (for legacy reasons).
 
   bool has_prefix = file->options().has_objc_class_prefix();
+  bool have_expected_prefix_file = !expected_prefixes_path.empty();
+
   const std::string prefix = file->options().objc_class_prefix();
   const std::string package = file->package();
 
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -1267,6 +1269,61 @@ bool ValidateObjCClassPrefix(
     return true;
   }
 
+  // When the prefix is non empty, check it against the expected entries.
+  if (!prefix.empty() && have_expected_prefix_file) {
+    // For a non empty prefix, look for any other package that uses the prefix.
+    std::string other_package_for_prefix;
+    for (std::map<std::string, std::string>::const_iterator i =
```
dmaclach, line 1276: std::string other_package_for_prefix;
auto iter = expected_package_prefixes.find(prefix);
if (iter != expected_package_prefixes.end()) {
  other_package_for_prefix = iter->second;
}

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -1267,6 +1269,61 @@ bool ValidateObjCClassPrefix(
...
+    if (package.empty()) {
+      // The file does not have a package and ...
+      if (other_package_for_prefix.empty()) {
+        // ... no other package has declared that prefix.
+        std::cerr
+             << "protoc:0: warning: File '" << file->name() << "' has no "
+             << "package. Consider adding a new package to the proto and adding '"
+             << "new.package = " << prefix << "' to the expected prefixes file ("
+             << expected_prefixes_path << ")." << std::endl;
+        std::cerr.flush();
```
dmaclach, line 1297: pull this outside the if/else since both branches do it?

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: Doesn't `find` look up by key?  The map is _package_ -> _prefix_, so I think the only option is what I'm already doing, iterate to find the _prefix_ and make note of the _package_ that uses it?

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: Keeping it with each warning logging line at the moment to maintain the practice of ensuring nothing else can interleave with it by accident.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dmaclach: ouch.. got you. Totally missed that. Thanks.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dmaclach: ack

dmaclach (approved):

PR #8783, opened 2021-07-01
Make glob recursive if option is enabled

### LINKED ISSUE #8247: Incorrect Recursing in protobuf_distutils for python
Version: v3.11.4
Language: Python

OSX Big Sur
Python 3.7
**What did you do?**
Assume dir called protos with following structure, protobuf files can be trivial.
Screen Shot 2021-02-01 at 3 11 48 PM

Call python setup.py generate_py_protobufs with the following options:
{
    'generate_py_protobufs': {. 
        # Shortest prefix directory of proto_files. 
        'source_dir':        'protos',  
        'output_dir': '.',    
    } 
}
This should compile each protobuf file, including baz.proto. 
**What did you see instead?**
This fails to compile baz.proto. This is due to https://github.com/protocolbuffers/protobuf/blob/94c3916dcfaf87cc1b4fdc7270b603636f7a70fd/python/protobuf_distutils/protobuf_distutils/generate_py_protobufs.py#L123. Python's glob function has a flag called recursive, which by default is set to false. Hence, even with a double star glob, the line will only match up to 1 directory depth, instead of being arbitrary.

Should be very trivial to fix and repackage to PYPI.

raulbocanegra:
This PR fixes #8247.
The documentation says `The default behavior is to generate sources for all .proto files found under source_dir, recursively` but this is not working as the issue describes.

### DIFF
--- a/python/protobuf_distutils/protobuf_distutils/generate_py_protobufs.py
+++ b/python/protobuf_distutils/protobuf_distutils/generate_py_protobufs.py
@@ -120,7 +120,7 @@ class generate_py_protobufs(Command):
         if self.proto_files is None:
             files = glob.glob(os.path.join(self.source_dir, '*.proto'))
             if self.recurse:
-                files.extend(glob.glob(os.path.join(self.source_dir, '**', '*.proto')))
+                files.extend(glob.glob(os.path.join(self.source_dir, '**', '*.proto'), recursive=True))
             self.proto_files = [f.partition(self.proto_root_path + os.path.sep)[-1] for f in files]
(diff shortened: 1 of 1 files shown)


raulbocanegra:
I signed it!

raulbocanegra:
@googlebot I signed it!

raulbocanegra:
Hi! could someone have a look at this before 3.18 release please? It's just a one line fix.

TeBoring:
Could you send the fix to 3.18.x branch?

raulbocanegra:
Done! #8950

PR #8789, opened 2021-07-03
Run update_file_lists.sh

nipunn1313:
Confirmed it worked via (previously failing)
```
bazel test //:build_files_updated_unittest
```


nipunn1313:
Please apply labels "Release notes: no" and "cmake" for the relevant area (I think)

Likely something slipped through the cracks allowing the master branch to be red - so it may be wise to follow up on what allowed that through.

nipunn1313:
I'm seeing
image

may want to consider updating the repository settings to allow workflow runs for outside contributors

https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#configuring-required-approval-for-workflows-from-public-forks

There appears to be settings for "require approval for first time contributors" and a looser setting for "require approval for first time contributors new to github". The latter might avoid this kind of headache for folks like me - so I can proactively make sure tests pass.

nipunn1313:
going to abandon this - feel free to reopen if it seems useful

PR #8792, opened 2021-07-05
Fix comment syntax in any.proto

chrisbainbridge:
Remove extra spaces which cause incorrect indentation in godoc.
Remove the "====" style title which is not rendered by godoc.

### DIFF
--- a/src/google/protobuf/any.proto
+++ b/src/google/protobuf/any.proto
@@ -64,7 +64,7 @@ option objc_class_prefix = "GPB";
 //       foo = any.unpack(Foo.class);
 //     }
 //
-//  Example 3: Pack and unpack a message in Python.
+// Example 3: Pack and unpack a message in Python.
 //
 //     foo = Foo(...)
 //     any = Any()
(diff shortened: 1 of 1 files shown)


chrisbainbridge:
@googlebot I signed it!

src/google/protobuf/any.proto
```diff
@@ -95,7 +95,7 @@ option objc_class_prefix = "GPB";
 //
 //
 // JSON
-// ====
+//
```
TeBoring, line 98: Can you use some other separator instead of totally removing the separator? e.g. #

chrisbainbridge: > Can you use some other separator instead of totally removing the separator? e.g. #

Not as far as I know - godoc doesn't use an explicit markup for headings - but "godoc recognises a line that starts with a capital letter and does not end in a full stop as a heading."

btw You can see the issue that this commit is fixing [here](https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb)

  - "Example 3" and "Example 4" are nested inside "Example 2"
  - the JSON heading is interpreted as a paragraph and rendered as `JSON ==== The JSON representation of an `Any` value...`

PR #8793, opened 2021-07-06
add null check to equals

PartialDragster:
added a null check to equals implementation so that it matches the method's contract

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/UnknownFieldSet.java
+++ b/java/core/src/main/java/com/google/protobuf/UnknownFieldSet.java
@@ -759,6 +759,9 @@ public final class UnknownFieldSet implements MessageLite {
 
     @Override
     public boolean equals(final Object other) {
+      if (other == null) {
+        return null;
+      }
       if (this == other) {
         return true;
       }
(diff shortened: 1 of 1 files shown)


PartialDragster:
@googlebot I signed it!

PartialDragster:
Okay, says it needs a label but I have NO idea how to add one...

perezd:
What contract are you referencing? I don't think this change immediately makes sense to me, if you pass `null` to equals it should return a boolean and `false`.

PartialDragster:
Ugh sorry that was a stupid mistake. Yes supposed to return `false' not `null`. 

I was submitting based on the result of static checks we got back on work but I just double checked and see that if `null` is provided as the LHS of `instanceof` it returns `false` (TIL) as you'd expect so it looks like their check is wrong. I'll feed that back to them. 

Sorry for the noise!

Hmmm although it looks like it might be possible to get an NRE a little later in the method though. It looks like it's possible for `fields` to be `null` so the expression `fields.equals(((UnknownFieldSet)other).fields)` will error if the `fields` on the given object is `null`. I'll submit a PR for that sometime today.

PR #8794, opened 2021-07-06
Remove deprecated version of SetTotalBytesLimit()

georgthegreat:
It has been deprecated for at least 3 years.
Worse backward incompatible things have happened during this period.

I think this method could be safely removed, as the client code fix is trivial.


acozzette:
It looks like coded_stream_unittest.cc needs to be updated because it still exercises the deprecated function.

georgthegreat:
Thanks, Adam.
This should be fixed now.

acozzette:
@georgthegreat Thanks for the cleanup!

MohamadHalwani:
`error: no matching function for call to ‘google::protobuf::io::CodedInputStream::SetTotalBytesLimit(const int&, const int&)’
`
any idea ?

georgthegreat:
You should use single parameter version instead.

PR #8806, opened 2021-07-09
Revert "Add option for eliding property metadata from messages"

thomasvl:
This reverts commit ca3674b7d5f76299138ca635fa9c02e01e0fdf76.

While there are savings, it ends up being to easy/common to run into issues with
AppStore validation since the selector usage now appears to be a possible match
for private apis vs. for selectors/properties in the generated code.


dmaclach (approved):
Sob... it was good while it lasted :)

PR #8808, opened 2021-07-12
Optimize ToByteArray()

xqrzd:
Update convenience method ToByteArray() to use the new API.

Sample benchmark serializing a tiny protobuf message,
|                        Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------ |---------:|---------:|---------:|-------:|------:|------:|----------:|
| ToByteArray_CodedOutputStream | 76.82 ns | 1.079 ns | 1.009 ns | 0.0279 |     - |     - |      88 B |
|           ToByteArray_WriteTo | 53.73 ns | 0.833 ns | 0.779 ns | 0.0076 |     - |     - |      24 B |

### DIFF
--- a/csharp/src/Google.Protobuf/MessageExtensions.cs
+++ b/csharp/src/Google.Protobuf/MessageExtensions.cs
@@ -109,9 +109,7 @@ namespace Google.Protobuf
         {
             ProtoPreconditions.CheckNotNull(message, "message");
             byte[] result = new byte[message.CalculateSize()];
-            CodedOutputStream output = new CodedOutputStream(result);
-            message.WriteTo(output);
-            output.CheckNoSpaceLeft();
+            message.WriteTo(result);
             return result;
         }
 


xqrzd:
@googlebot I signed it!

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -109,9 +109,7 @@ public static byte[] ToByteArray(this IMessage message)
         {
             ProtoPreconditions.CheckNotNull(message, "message");
             byte[] result = new byte[message.CalculateSize()];
-            CodedOutputStream output = new CodedOutputStream(result);
-            message.WriteTo(output);
-            output.CheckNoSpaceLeft();
```
jtattermusch, line 114: what happens with the CheckNoSpaceLeft() check here?

jtattermusch:
FTR, Can you link to the implementation you're referring to as "new API"?

jtattermusch:
> FTR, Can you link to the implementation you're referring to as "new API"?

Is it this one?
https://github.com/protocolbuffers/protobuf/blob/98b6ead7756237cfbad9006c6032d96dbc82fae3/csharp/src/Google.Protobuf/MessageExtensions.cs#L182

csharp/src/Google.Protobuf/MessageExtensions.cs
xqrzd: The span based WriteTo() extension also performs that check,
https://github.com/protocolbuffers/protobuf/blob/949caa7d66e4c1f0f4afe7bddabe5416eaaf4285/csharp/src/Google.Protobuf/MessageExtensions.cs#L180-L187

xqrzd:
Yep, basically this PR makes it so if you were already doing,
```csharp
var size = messagePb.CalculateSize();
var buffer = new byte[size];
messagePb.WriteTo(buffer);
```
you can now do,
```csharp
var buffer = messagePb.ToByteArray();
```
and get the same performance.

jtattermusch:
Tests are actually failing.
```
A total of 1 test files matched the specified pattern.
  Failed IntermixingOfNewAndLegacyGeneratedCodeWorksWithCodedInputStream [19 ms]
  Error Message:
   Google.Protobuf.InvalidProtocolBufferException : Message LegacyGeneratedCodeMessageA doesn't provide the generated method that enables WriteContext-based serialization. You might need to regenerate the generated protobuf code.
  Stack Trace:
     at Google.Protobuf.WritingPrimitivesMessages.WriteRawMessage(WriteContext& ctx, IMessage message) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf/WritingPrimitivesMessages.cs:line 96
   at Google.Protobuf.LegacyGeneratedCodeTest.ParseContextEnabledMessageB.pb::Google.Protobuf.IBufferMessage.InternalWriteTo(WriteContext& output) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf.Test/LegacyGeneratedCodeTest.cs:line 242
   at Google.Protobuf.MessageExtensions.WriteTo(IMessage message, Span`1 output) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf/MessageExtensions.cs:line 186
   at Google.Protobuf.MessageExtensions.ToByteArray(IMessage message) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf/MessageExtensions.cs:line 113
   at Google.Protobuf.LegacyGeneratedCodeTest.IntermixingOfNewAndLegacyGeneratedCodeWorksWithCodedInputStream() in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf.Test/LegacyGeneratedCodeTest.cs:line 49
  Failed LegacyGeneratedCodeThrowsWithReadOnlySequence [1 ms]
  Error Message:
   Google.Protobuf.InvalidProtocolBufferException : Message LegacyGeneratedCodeMessageA doesn't provide the generated method that enables WriteContext-based serialization. You might need to regenerate the generated protobuf code.
  Stack Trace:
     at Google.Protobuf.WritingPrimitivesMessages.WriteRawMessage(WriteContext& ctx, IMessage message) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf/WritingPrimitivesMessages.cs:line 96
   at Google.Protobuf.LegacyGeneratedCodeTest.ParseContextEnabledMessageB.pb::Google.Protobuf.IBufferMessage.InternalWriteTo(WriteContext& output) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf.Test/LegacyGeneratedCodeTest.cs:line 242
   at Google.Protobuf.MessageExtensions.WriteTo(IMessage message, Span`1 output) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf/MessageExtensions.cs:line 186
   at Google.Protobuf.MessageExtensions.ToByteArray(IMessage message) in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf/MessageExtensions.cs:line 113
   at Google.Protobuf.LegacyGeneratedCodeTest.LegacyGeneratedCodeThrowsWithReadOnlySequence() in /tmp/protobuf/protobuf/csharp/src/Google.Protobuf.Test/LegacyGeneratedCodeTest.cs:line 74

Failed!  - Failed:     2, Passed:  1140, Skipped:     0, Total:  1142, Duration: 7 s - /tmp/protobuf/protobuf/csharp/src/Google.Protobuf.Test/bin/Release/netcoreapp2.1/Google.Protobuf.Test.dll (netcoreapp2.1)
```
https://source.cloud.google.com/results/invocations/f9718809-b44b-43db-bb96-22b8a23370ba/log

I think the issue is that the Span-based codepath will not exist for messages generated by old codegen (which is what the test is complaining about) - but the library still needs to stay compatible with such messages (especially since ToByteArray() method has been in the public API basically since the library's inception).
In order to make this work, you'd need to intelligently detect if using the `WriteTo(this IMessage message, Span<byte> output)` method is safe for given message.

jtattermusch (changes requested):
The tests are failing - see my previous comment.

xqrzd:
Ah I see. For intelligently checking I came up with this,
```csharp
[SecuritySafeCritical]
public static byte[] ToByteArray(this IMessage message)
{
    ProtoPreconditions.CheckNotNull(message, "message");
    byte[] result = new byte[message.CalculateSize()];

    if (message is IBufferMessage)
    {
        message.WriteTo(result);
    }
    else
    {
        CodedOutputStream output = new CodedOutputStream(result);
        message.WriteTo(output);
        output.CheckNoSpaceLeft();
    }

    return result;
}
```

which gets the failures down to 2. Unfortunately, they're mixed cases, which would require checking more than just the argument. I don't think it's worth it, given how minor this is, and that there's an alternative that achieves the same performance.

Maybe a new extension method could be added that does,
```csharp
var size = messagePb.CalculateSize();
var buffer = new byte[size];
messagePb.WriteTo(buffer);
```

PR #8814, opened 2021-07-14
Fix json name oneof in ruby implementation

chatgris:
Hi there, I add the case where with a proto like the following, the ruby implementation would breal with the following exception: `optional': wrong argument type Hash (expected String) (TypeError)`.

``` proto
message TestJsonName {
  int32 value = 1 [json_name = "CustomJsonName"];
  oneof jsonname_oneof {
    string name = 2 [json_name = "OneOfCustomJsonName"];
  }
 }
```

This pull-request should fix this issue.

Cheers!

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -2190,6 +2190,13 @@ static VALUE OneofBuilderContext_optional(int argc, VALUE* argv, VALUE _self) {
 
   rb_scan_args(argc, argv, "32", &name, &type, &number, &type_class, &options);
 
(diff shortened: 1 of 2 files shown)


chatgris:
@googlebot I signed it!

chatgris:
@googlebot I signed it!

acozzette:
@chatgris Thank you for the pull request and I'm sorry for taking so long to review this. It looks like in the meantime the code has changed so much that the `OneofBuilderContext_optional` function no longer exists, so I will go ahead and close this PR. If the bug still exists, though, please let us know and we can try to figure out another fix.

PR #8823, opened 2021-07-16
Update the automake and Bazel configs to be consistent about field_ac…

acozzette:
…cess_listener.cc

In an earlier change I moved field_access_listener.cc from the lite
runtime to the full runtime in the CMake configuration, but this caused
//:build_files_updated_unittest to start failing because it expects all
three build systems to be consistent. To fix that, I updated the
Automake configuration and then ran ./update_file_lists.sh to propagate
that change to the other configs.


perezd (approved):
LGTM

PR #8824, opened 2021-07-17
Remove unused rubygems-tasks development dependency

postmodern:
Cannot find any occurrence of `require 'rubygems/tasks` or `Gem::Tasks` in the source code, so I am confident that this development dependency can safely be removed (even though I may be a little biased in favor of using [rubygems-tasks](https://github.com/postmodern/rubygems-tasks)).


postmodern:
@googlebot I signed it!

PR #8827, opened 2021-07-19
Fix undefined symbol error around SharedCtor()

georgthegreat:
Apparently, #8532 was incorrect if applied to 3.17.x branch.

3.17.x changed code generation to mark `SharedCtor()` and `SharedDtor()` as inline in .pb.cc.

It looks like we have a compile-time undefined behavior in C++ now. As cppreference.com [says](https://en.cppreference.com/w/cpp/language/inline):

_The definition of an inline function <...> must be reachable in the translation unit where it is accessed (not necessarily before the point of access)._

As protobuf allows custom plugins to generate custom code, there is no limitation on where SharedCtor couble be possible referenced from. In our case we have SharedCtor invoked from corresponding `.pb.h` code, thus triggering:
```
ld: error: undefined symbol: package::Message::SharedCtor()`
>>> referenced by file.pb.h:$$$$
```

If this patch is not applicable, I can take a look into changing the code generation, but doing this will be harder than removing _inline_.

### DIFF
--- a/src/google/protobuf/any.pb.h
+++ b/src/google/protobuf/any.pb.h
@@ -129,12 +129,12 @@ class PROTOBUF_EXPORT Any final :
       const ::PROTOBUF_NAMESPACE_ID::FieldDescriptor** value_field);
   template <typename T, class = typename std::enable_if<!std::is_convertible<T, const ::PROTOBUF_NAMESPACE_ID::Message&>::value>::type>
   bool PackFrom(const T& message) {
-    return _any_metadata_.PackFrom<T>(message);
+    return _any_metadata_.PackFrom<T>(GetArena(), message);
   }
   template <typename T, class = typename std::enable_if<!std::is_convertible<T, const ::PROTOBUF_NAMESPACE_ID::Message&>::value>::type>
   bool PackFrom(const T& message,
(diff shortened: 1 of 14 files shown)


georgthegreat:
NB: the same problem applies to SharedDtor() though it is not broken in our case.

acozzette:
@georgthegreat Would you mind running `./generate_descriptor_proto.sh`? That will rebuild the checked-in generated code and should fix the failing test runs.

src/google/protobuf/any.pb.h
```diff
@@ -129,12 +129,12 @@ class PROTOBUF_EXPORT Any final :
       const ::PROTOBUF_NAMESPACE_ID::FieldDescriptor** value_field);
   template <typename T, class = typename std::enable_if<!std::is_convertible<T, const ::PROTOBUF_NAMESPACE_ID::Message&>::value>::type>
   bool PackFrom(const T& message) {
-    return _any_metadata_.PackFrom<T>(message);
```
georgthegreat, line 132: NB: this also includes fixes for #8758. At the time I did not knew about this script for checked-in files regeneration.

georgthegreat:
@deannagarcia, @acozzette, can we proceed with merging of this PR?

PR #8829, opened 2021-07-19
Fixes #8828.

### LINKED ISSUE #8828: Definition of CMAKE_CXX_EXTENSIONS is not consistent with CYGWIN and causes build errors on QNX
Version: master
Language: C++

Windows 10

QCC v8.3

**What did you do?**
1. Crosscompile Protobuf for QNX.

Successful build and consistent behavior with CYGWIN: CMAKE_CXX_EXTENSIONS must be set to ON to match -std=gnu++11 as  defined in [CXX_EXTENSIONS](https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html#prop_tgt:CXX_EXTENSIONS).

**What did you see instead?**
Build errors due to undeclared functions and missing headers.

pablorcum:
According to the [CMake doc](https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html#prop_tgt:CXX_EXTENSIONS), CMAKE_CXX_EXTENSIONS must be set to ON to match -std=gnu++11 instead of -std=c++11.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -25,7 +25,7 @@ if (CYGWIN)
 else()
   set(CMAKE_CXX_STANDARD 11)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
-  set(CMAKE_CXX_EXTENSIONS OFF)
+  set(CMAKE_CXX_EXTENSIONS ON)
 endif()
 
 # The Intel compiler isn't able to deal with noinline member functions of


pablorcum:
@googlebot I signed it!

cmake/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if (CYGWIN)
 else()
```
deannagarcia, line 25: Why doesn't the if statement above take care of the case where we need -std=gnu++11?

PR #8833, opened 2021-07-22
Improve SFINAE for RepeatedPtrField сtor

georgthegreat:
NB: There is ONE BIG IF in this PR. I will understand if this PR will not be accepted.

At the time RepeatedPtrField<string> is implicitly constructible in the following case if string is constructible from char:
```
RepeatedPtrField<string> vector({"first", "second"});
```

Though std::string is not constructible from single char, we use custom string type internally and this code gives us problem. I think there are also more valid cases where such syntax will break.

I have copied the idea from libc++ implementation of std::vector.

### DIFF
--- a/src/google/protobuf/repeated_field.h
+++ b/src/google/protobuf/repeated_field.h
@@ -938,7 +938,7 @@ class RepeatedPtrField final : private internal::RepeatedPtrFieldBase {
 
   template <typename Iter,
             typename = typename std::enable_if<std::is_constructible<
-                Element, decltype(*std::declval<Iter>())>::value>::type>
+                Element, typename std::iterator_traits<Iter>::value_type>::value>::type>
   RepeatedPtrField(Iter begin, Iter end);
 
   ~RepeatedPtrField();


jorgbrown:
Hi Yuriy!

Before we can accept this, I need to know more about the situation where the alternate form will prevent a problem.

Specifically, I thought this code would be prevented with your change:

```
template <typename Element>
struct RepeatedPtrField {
  template <typename Iter,
            typename = typename std::enable_if<std::is_constructible<
                Element, typename std::iterator_traits<Iter>::value_type>::value>::type>
  RepeatedPtrField(Iter begin, Iter end);
};

template <typename T = char>
struct basic_string {
  basic_string();
  basic_string(const basic_string &);
  basic_string(const T *);
  basic_string(T);
};

auto fn() {
  using string = basic_string<char>;
  RepeatedPtrField<string> to_test("one", "tw");
}
```

But alas, it still compiles.  See https://godbolt.org/z/T4nYoecnP .

What am I doing wrong?  Can you share more about your custom string class?

-- Jorg

georgthegreat:
> What am I doing wrong? Can you share more about your custom string class?

Out custom string implementation can be found in [catboost repository](https://github.com/catboost/catboost/blob/master/util/generic/string.h). The problematic ctor starts [here](https://github.com/catboost/catboost/blob/e2ad9464d2e0574cdcd7f974c35b3546178e0329/util/generic/string.h#L506). std::string lacks this ctor because there is no TExplicitType helper and without it the code is quite unsafe.

Your code is incorrect, as it would not work as expected: 
```
Pull8833::RepeatedPtrField<custom::string> to_test("one", "two");
```

The constructor being fixed is expecting iterators, not string literals. 
It will increase them and compare for equality.

My code causes ambiguity in the following case:
```
// TString as referenced above
class Klass {
    /* implicit */ Klass(std::vector<TString>);
    /* implicit */ Klass(RepeatedPtrField<TString>);
}
```

See this for more extensive demonstation: https://godbolt.org/z/hsxG5hWfz.
The problem is, this PR does not fix the ambiguousness: std::is_constructible still gives the problems.

georgthegreat:
Explicitly checking for iterator_category solves the issue, though the code is becoming mostrous: https://godbolt.org/z/4j8c9essc

(still need to add desired iterator cateries into the list).

See [libc++ vector source](https://github.com/llvm/llvm-project/blob/acbc0c5f0ebd8b7ebfa2eb3ae77428eb83c428c5/libcxx/include/vector#L523) for the reference.

jorgbrown:
The monstrous code you cite works for blocking the case we care about, but it also blocks cases we want to support, such as:
```
    custom::string custom_ar[2] = {"one", "two"};
    Pull::RepeatedPtrField<custom::string> pull_good_rpf(&custom_ar[0], &custom_ar[2]);
```

One reason for this being blocked is the mistaken spelling of `iterator_categary`.

---

Separately, note that, for your custom string type, this compiles (though of course it does the wrong thing):
```
    std::vector<custom::string> svcs("one", "two");
```
Whereas this does not:
```
    std::vector<std::string> svcs("one", "two");
```
This is one of at least two reasons I would strongly suggest you do NOT allow your custom string class to be constructible from a single one of its elements.  The implicit conversion of integral types to `char` is another big reason.

You may point out that std::string can't easily be constructed from just a char:
```
auto MakeAString(char c) {
  return std::string(c);  // won't compile
}
```
But thanks to C++11, there's an easy fix:
```
auto MakeAString(char c) {
  return std::string({c});  // Horray for initializer lists!
}
```

georgthegreat:
@jorgbrown, thanks for a thorough review.
I will discard this PR and try to remove single element ctor from out TString.

PR #8834, opened 2021-07-25
Update protokt options link

andrewparmet:
We've open-sourced the protokt code generator under a different GitHub organization than originally planned.


andrewparmet:
@googlebot I signed it!

Issue #7408, opened 2020-04-22
MergeFrom error when using pure python message implementation 


monouno:
**What version of protobuf and what language are you using?**
Version: 3.6.1
Language: Python 3.7

**What operating system (Linux, Windows, ...) and version?**
* Ubuntu 19.04 (GNU/Linux 5.0.0-13-generic aarch64)
* macOS 10.15.4

**What runtime / compiler are you using (e.g., python version or gcc version)**
Python 3.7

**What did you do?**
I have two proto file. In our actual project, these two files have a lot of content, only two message are pasted to illustrate the problem.

One is route.proto:
```
syntax = "proto3";
package a.b.c;

message Route {
    string id = 1; // uuid, generated by minion
    string dst = 2;
    string via = 3;
    string link_name = 4;
    int32 table = 5;
}
```
Another is network.proto:
```
syntax = "proto3";
package a.b.c;

message GetDefaultGatewayResponse {
    Route route = 1;
}
```

I run the following command in Interactive Python:
```
from rpc.network.network_pb2 import *
from rpc.network.route_pb2 import *
GetDefaultGatewayResponse(route=Route())
```
Then the following exception was thrown:
```
~/venv/lib/python3.7/site-packages/google/protobuf/internal/python_message.py in MergeFrom(self, msg)
   1229       raise TypeError(
   1230           "Parameter to MergeFrom() must be instance of same class: "
-> 1231           'expected %s got %s.' % (cls.__name__, msg.__class__.__name__))
   1232 
   1233     assert msg is not self
```
But I add `print(type(msg), cls)` got same value is `<class 'route_pb2.Route'>` but id of that different.

However, there is no such problem with cpp message implementation.
**What did you expect to see**
```
In [1]: from rpc.network.network_pb2 import * 
   ...: from rpc.network.route_pb2 import * 
   ...: GetDefaultGatewayResponse(route=Route())                                                                                                                                                 
Out[1]: 
route {
}
```
**What did you see instead?**
```
In [1]: from rpc.network.network_pb2 import * 
   ...: from rpc.network.route_pb2 import * 
   ...: GetDefaultGatewayResponse(route=Route())                                                                                                                                                 
<class 'route_pb2.Route'> <class 'route_pb2.Route'>
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/venv/lib/python3.7/site-packages/google/protobuf/internal/python_message.py in init(self, **kwargs)
    517         try:
[...]

PR #8843, opened 2021-07-29
[#7408] Fix for unexpected behavior of python's isinstance() in MergeFrom()

samsetegne:
Different versions of python (confirmed 3.7 & 3.8) can have unpredictable behavior for the builtin `isinstance()` method. Particularly when the instance was initialized using a class definition with subtle differences in how it was imported when compared to the class definition used at the time of the `isinstance()` check.

I believe this is because the subsequent import of the same class actually overwrites the previous import. This change can be illustrated by seeing the change in id when overwriting a class in the same file.
```py
class fooclass:
    pass

a = fooclass()

class fooclass:
    pass

b = fooclass()

print(id(a.__class__))
# 140479596517648

print(id(b.__class__))
# 140479596518592
```

This PR softens that validation by instead only checking that the class name matches. This will resolve cases where users ran into this failure and received a confusing exception such as in #7408 which saw `TypeError: Parameter to MergeFrom() must be instance of same class: expected Route got Route.` or in my case where I saw `TypeError: Parameter to MergeFrom() must be instance of same class: expected google.protobuf.Timestamp got google.protobuf.Timestamp.`.


samsetegne:
@googlebot I signed it!

TeBoring:
I think this is an error that should be fixed in python.
If any version of python has this error, avoid using that version of python.

PR #8847, opened 2021-07-30
bazel: make `*.bzl` files publicly visible targets

UebelAndre:
This allows these files to be used in [stardoc](https://github.com/bazelbuild/stardoc) docs generation.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1301,3 +1301,9 @@ exports_files([
     "conformance/text_format_failure_list_java.txt",
     "conformance/text_format_failure_list_java_lite.txt",
 ])
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//visibility:public"],
+)


perezd:
I don't think we want these all to be publicly accessible wholesale. This would allow anyone to depend on internal bzl rules.

UebelAndre:
@perezd This doesn't change whether or not the contents of the `.bzl` files can be accessed by other projects. The visibility of these files is only relevant to the files themselves as targets which is required when generating documentation via `stardoc`. I would kindly ask you to reconsider closing this PR.

BUILD
```diff
@@ -1301,3 +1301,14 @@ exports_files([
     "conformance/text_format_failure_list_java.txt",
     "conformance/text_format_failure_list_java_lite.txt",
 ])
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
```
UebelAndre, line 1307: One thing that might be misleading about the globs here is that they do not actually include all `*.bzl` files in the project. This would only include all `*.bzl` files within the current [package](https://docs.bazel.build/versions/main/build-ref.html#packages).

Again, this change just means I can use the protobuf macros in files I'm generating documentation for

perezd: **/* is a recursive glob, it's everything from that definition point and on that ends in .bzl at any tree level.

BUILD
```diff
@@ -1301,3 +1301,14 @@ exports_files([
...
+    visibility = ["//visibility:public"],
+)
+
+exports_files(
```
perezd, line 1311: This would allow anyone to depend on them in `load` statements.

UebelAndre: No, [exports_files](https://docs.bazel.build/versions/main/be/functions.html#exports_files) allows files in a package to be treated as targets. Nothing prevents loading a public symbol from any `.bzl` file.

BUILD
UebelAndre: No, [glob](https://docs.bazel.build/versions/main/be/functions.html#glob) does not traverse package boundaries. And all `.bzl` files are accompanied by a `BUILD` file
```
protobuf % find . -name "BUILD" -or -name "*.bzl"
./cc_proto_blacklist_test.bzl
./util/python/BUILD
./protobuf_version.bzl
./internal.bzl
./java/core/BUILD
./java/util/BUILD
./java/lite/BUILD
./java/internal/testing.bzl
./java/internal/BUILD
./java/BUILD
./protobuf.bzl
./third_party/BUILD
./protobuf_deps.bzl
./examples/BUILD
./benchmarks/datasets/google_message1/proto2/BUILD
./benchmarks/datasets/google_message1/proto3/BUILD
./benchmarks/datasets/google_message4/BUILD
./benchmarks/datasets/google_message3/BUILD
./benchmarks/datasets/google_message2/BUILD
./benchmarks/datasets/BUILD
./benchmarks/cpp/BUILD
./benchmarks/BUILD
./compiler_config_setting.bzl
./BUILD
./objectivec/BUILD
```

UebelAndre:
@perezd I think there's some confusion around the use of `load` and `exports_files`. This PR is only intended to add `stardoc` support to already available `.bzl` files. This does not change the behavior of how any of these files are loaded (via a `load` statement in a `BUILD` or `*.bzl` file) in any way.

BUILD
perezd: Right, I should say, our BUILD/package boundaries are not consistent/what we want them to be long term, so I'm afraid that will be overly eager and allow non-stardoc dependencies on these files as file inputs to other targets potentially without realizing/being careful to exclude ones we don't ever intend to make accessible.

BUILD
perezd: Ah right, this is a Bazel specific thing I always forget.

perezd: https://docs.bazel.build/versions/main/visibility.html#visibility-of-bzl-files

perezd:
I'd like @dlj-NaN to double check this and make sure we're OK with these targets being added.

BUILD
UebelAndre: The reason these globs are broad is to hopefully allow them to be invisible and conform to whatever structure you end up with.

Concretely, the problem I'm facing is that a project I work on introduced a new `proto` rule which loads [protobuf.bzl](https://github.com/protocolbuffers/protobuf/blob/79370f1ffa81a8394c41a99a32ab1452f923242e/protobuf.bzl) for the implementation and [protobuf_deps.bzl](https://github.com/protocolbuffers/protobuf/blob/79370f1ffa81a8394c41a99a32ab1452f923242e/protobuf_deps.bzl) to expose dependencies. This change broke our ability to generate documentation because `stardoc` requires access to these files as targets. Unless [stardoc](https://github.com/bazelbuild/stardoc) is refactored, any user of it will require all `.bzl` files they load to be accessible as Bazel targets to pass to the `stardoc` rule. Many rules expose similar targets ([rules_nodejs](https://github.com/bazelbuild/rules_nodejs/blob/stable/BUILD.bazel#L31-L59), [rules_python](https://github.com/bazelbuild/rules_python/blob/main/BUILD#L18-L52), [rules_rust](https://github.com/bazelbuild/rules_rust/blob/main/rust/BUILD.bazel#L5-L11), [rules_go](https://github.com/bazelbuild/rules_go/blob/master/go/BUILD.bazel#L16-L24)). The thought in keeping this pattern broad was that the structure hasn't changed in some time and and if it should, the pattern would account for that.

BUILD
```diff
@@ -1301,3 +1301,14 @@ exports_files([
     "conformance/text_format_failure_list_java.txt",
     "conformance/text_format_failure_list_java_lite.txt",
 ])
+
+filegroup(
```
dlj-NaN, line 1305: We should stick to _either_ `filegroup` or `exports_files`. (Regardless of current behavior, we should not add ambiguity in mapping between files and rules.)

dlj-NaN (changes requested):

BUILD
UebelAndre: I feel this is a common pattern even in Bazel itself but maybe I don't think I understand the ambiguity you're referring to.

Either way, the filegroup is now deleted.

UebelAndre: Sorry, after doing more digging it seems the filegroup is what's used in [Bazel itself](https://github.com/bazelbuild/bazel/blob/7ce19e4128f7a5db219623aadf448d5690097140/tools/build_defs/BUILD#L29-L39). I've removed the exports and restored the filegroup.

perezd (approved):

UebelAndre:
@dlj-NaN is there anything else I should address here?

UebelAndre:
@dlj-NaN @perezd friendly ping here. Is there anything more for me to do on this PR?

Issue #8842, opened 2021-07-29
[Urgent] Seg fault using Ruby google-protobuf v3.17.3


nishihara-tm:
**What version of protobuf and what language are you using?**
protobuf: 3.17.3
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**

**What runtime / compiler are you using (e.g., python version or gcc version)**
gcc (Debian 8.3.0-6) 8.3.0

**What did you do?**
Still trying to reproduce error

**What did you expect to see**
No segmentation fault

**What did you see instead?**
Segmentation fault

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

```
/usr/local/bundle/gems/google-ads-googleads-11.0.0/lib/google/ads/google_ads/v7/resources/account_budget_pb.rb:15: [BUG] Segmentation fault at 0x0000000000000010 | /usr/local/bundle/gems/google-ads-googleads-11.0.0/lib/google/ads/google_ads/v7/resources/account_budget_pb.rb:15: [BUG] Segmentation fault at 0x0000000000000010
-- | --
  | 2021-07-29T02:32:13.214+09:00 | ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux]
  | 2021-07-29T02:32:13.214+09:00 | -- Control frame information -----------------------------------------------
  | 2021-07-29T02:32:13.214+09:00 | c:0076 p:---- s:0443 e:000442 CFUNC :add_message
  | 2021-07-29T02:32:13.214+09:00 | c:0075 p:0007 s:0438 E:001c10 BLOCK /usr/local/bundle/gems/google-ads-googleads-11.0.0/lib/google/ads/google_ads/v7/resources/account_budget_pb.rb:15 [FINISH]
  | 2021-07-29T02:32:13.214+09:00 | c:0074 p:---- s:0435 e:000434 CFUNC :instance_eval
  | 2021-07-29T02:32:13.214+09:00 | c:0073 p:---- s:0432 e:000431 CFUNC :add_file
  | 2021-07-29T02:32:13.214+09:00 | c:0072 p:0009 s:0426 E:0017f0 BLOCK /usr/local/bundle/gems/google-ads-googleads-11.0.0/lib/google/ads/google_ads/v7/resources/account_budget_pb.rb:14 [FINISH]
  | 2021-07-29T02:32:13.214+09:00 | c:0071 p:---- s:0423 e:000422 CFUNC :instance_eval
[...]

Issue #1198, opened 2016-01-29
No way to use custom options in Ruby


kailan:
I created a custom option extension that I use in my proto files, like this:

``` proto
service Fishtank {
  rpc GetFish(FishQuery) returns (Fish) {
    option (protapi.http).get = "/fish/{id}";
  }
}
```

It doesn't seem that once it is compiled into Ruby code, I can access the value of the `protapi.http` option. I believe this can be done in other languages by accessing the descriptor.

Am I missing something, or is this not possible right now?

haberman:
You are correct that Ruby does not yet support custom options. Hopefully we can add this before too long.

kblanks:
@haberman Awesome! It'll really help out a project of mine.

Any way I can help out with implementing this?

skipperguy12:
This would be helpful for me as well

haberman:
This is becoming a more and more common request. I have been thinking through how the implementation ought to work on this. It's slightly tricky to modify the DSL for this, because custom options let you declare the option and use it both in the same file. This is a challenge to accommodate in a clean way.

kblanks:
Any progress on this? @haberman

haberman:
Ok there are a couple parts to making this work.

First of all we need to support proto2 field presence. Right now the code generator fails immediately if it sees a file with `syntax="proto2";`. Step 1 would be a PR that supports proto2 files that don't contain any extensions or custom options. Proto2 "optional" fields should be supported by adding hasbits to the in-memory representation to indicate whether a field is present or not. I'm happy to explain this in more detail for anyone who wants to implement this.

Secondly we need to support proto2 extensions. This means extending the DSL with something like `add_extension` and making it so you can read and write extension fields of extended messages with something like `msg[my_extension] = 5`.

Finally we need a way of extending the DSL to support custom options. We can talk more about this part once #1 and #2 are done.

I'd be happy to see pull requests for any/all of these, and I'd be happy to help guide people who want to implement this to the right places in the code.

kblanks:
I'd love to work on this but I have one question: Why would proto2 need to be supported? Surely implementing this for proto3 would be enough.

haberman:
Custom options are defined as extensions of messages `descriptor.proto`, which is a `syntax="proto2";` file. Extensions don't exist at all in proto3.

So even if your messages are only proto3, we need proto2 support to be able to represent the options themselves.

kblanks:
[...]

PR #8850, opened 2021-07-31
Move DSL implementation from C to pure Ruby

haberman:
The Ruby DSL for defining a protobuf schema has always been implemented in C. Originally this was necessary, because upb could not consume descriptors directly. However for several years now the ultimate output of the DSL is just a protobuf (`FileDescriptorProto`), so the DSL could just as easily be implemented in Ruby.

This PR removes 1,352 lines of code from C and reimplements them as ~400 lines of Ruby. This will have the following benefits:

- Reduces our overall C footprint, reducing the chance of SIGSEGV bugs (such as https://github.com/protocolbuffers/protobuf/issues/8842 which is a recent SEGV crash bug in the DSL).
- Reduces our overall line count and maintenance burden.
- Makes it possible to share the DSL code between MRI and JRuby.

It is possible that this PR will have some amount of performance regression in the startup time required to load generated `foo_pb.rb` files, since the DSL is evaluated using Ruby instead of C. However, the DSL was already calling back and forth between C and Ruby a lot, so hopefully the impact will not be too large. In any case, there are ways of mitigating this if it becomes an issue (see below).

This change requires us to add a new method `DescriptorPool#add_serialized_file`, which allows for defining messages using a serialized descriptor proto instead of the DSL. This is necessary to bootstrap `google/protobuf/descriptor_pb.rb`, which cannot use the DSL as these protos are used to implement the DSL.

## Future Directions

In the future we may want to change the code generator to use `DescriptorPool#add_serialized_file` for all generated files, instead of the DSL. This would have the following benefits:

- **Performance:** loading from a serialized descriptor will avoid the CPU cost of evaluating the DSL. For large schemas with many proto files, this could have a noticeable impact on app startup time.
- **Correctness/completeness:** as we continue to implement missing features in Ruby protobuf, particularly [custom options](https://github.com/protocolbuffers/protobuf/issues/1198), the DSL will become more and more of a hindrance. Custom options have some very difficult edge cases that would be nearly impossible to accommodate in the DSL unless we resort to serialized descriptors at some level. Using binary descriptors avoids all of this complication.

There are various options for how we could keep the generated code readable, even if we move to binary descriptors. For example, we could do something like:

```ruby
if false
  # Code for IDEs and human readers.
  class FooMessage
    attr_accessor :bar_field,
    attr_accessor :baz_field
  end
else
  # Code that is actually executed, implements semantics equivalent to the above.
  Google::Protobuf::DescriptorPool.generated_pool.add_serialized_file("<unreadable binary data>")
  FooMessage = Google::Protobuf::DescriptorPool.generated_pool.lookup("FooMessage").msgclass
end
```

Since this will require some input and dialogue about what is the best fit for the Ruby community, I'm leaving this out of the current PR.

The DSL will continue to be available no matter what, since we've exposed it already as a public API.


haberman:
Hmm, looks like this is having problems on Ruby 3.0. Both the Linux and MacOS builds are getting ("argument stack underflow") on Ruby 3.0:

MacOS:
```
 <L010> [sp: 2]
   0081 getlocal_WC_0        3                                           ( 335)
*  0083 opt_send_without_block <calldata:false, -1326341328>             ( 334)
   0085 opt_ltlt             <calldata:<<, 1>                            ( 334)
   0087 leave                                                            ( 338)
 <L003> [sp: 0]
   0088 putnil                                                           ( 334)
 <L001> [sp: -1]
   0089 nop                                                              (   0)
   trace: 200
   0090 leave                                                            ( 338)
---------------------
/Volumes/BuildData/tmpfs/src/github/protobuf/ruby/lib/google/protobuf/descriptor_dsl.rb: /Volumes/BuildData/tmpfs/src/github/protobuf/ruby/lib/google/protobuf/descriptor_dsl.rb:334: argument stack underflow (-208732189) (SyntaxError)
```

Linux:

```
 <L010> [sp: 2]
   0081 getlocal_WC_0        3                                           ( 335)
*  0083 opt_send_without_block <calldata:false, 22188128>                ( 334)
   0085 opt_ltlt             <calldata:<<, 1>                            ( 334)
   0087 leave                                                            ( 338)
 <L003> [sp: 0]
   0088 putnil                                                           ( 334)
 <L001> [sp: -1]
   0089 nop                                                              (   0)
   trace: 200
   0090 leave                                                            ( 338)
---------------------
/tmp/protobuf/protobuf/ruby/lib/google/protobuf/descriptor_dsl.rb: /tmp/protobuf/protobuf/ruby/lib/google/protobuf/descriptor_dsl.rb:334: argument stack underflow (-679739421) (SyntaxError)
```

This is a very odd error, especially that it occurs on Ruby 3.0 only. When running locally I've also seen:

```
<OBJ_INFO:gc_mark_ptr@gc.c:6106> 0x000055afeb295360 [2 M   ] T_NONE                                                                                                                                                
/usr/local/google/home/haberman/code/protobuf/ruby/lib/google/protobuf.rb:54: [BUG] try to mark T_NONE object 
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-linux]
```

haberman:
The Ruby 3.0 errors appear to *not* reproduce under Ruby 3.0.2, only Ruby 3.0.0.

This leads me to believe that these are bugs in Ruby 3.0.0 that have since been fixed.

I'll see if I can update the Kokoro jobs to use Ruby 3.0.2 instead.

ruby/lib/google/protobuf/descriptor_dsl.rb
```diff
@@ -0,0 +1,440 @@
...
+        #   end
+        #   add_message "foo.Bar.Baz" do
+        #   end
+        # end
+        #
+        # Here we need to do a translation where we turn this generated code into the
+        # above descriptor.  We need to infer that "foo" is the package name, and not
+        # a message itself. */
+
+        def get_parent_name(msg_or_enum)
```
fowles, line 196: a getter that mutates it argument and returns  a value is fairly surprising.  Can we instead have this be

```
def split_parent_name(msg_or_enum)
  full_name = msg_or_enum.name
  idx = full.rindex(?.)
  if idx
    return (name[0...idx], name[idz+1..-1])
  else
    return nil, full_name
  end
end
```

src/google/protobuf/compiler/ruby/ruby_generator.cc
```diff
@@ -527,6 +527,42 @@ bool MaybeEmitDependency(const FileDescriptor* import,
   }
 }
 
+bool GenerateDSLDescriptor(const FileDescriptor* file, io::Printer* printer,
```
fowles, line 530: GenerateDslDescriptor for the google style guide

ruby/lib/google/protobuf/descriptor_dsl.rb
```diff
@@ -0,0 +1,440 @@
...
+        # The DSL can omit a package name; here we infer what the package is if
+        # was not specified.
+        def infer_package(names)
+          # Package is longest common prefix ending in '.', if any.
+          if not names.empty?
+            min, max = names.minmax
+            last_common_dot = nil
+            min.size.times { |i|
+              if min[i] != max[i] then break end
+              if min[i] == ?. then last_common_dot = i end
```
fowles, line 82: let's switch to using strings over character literals, the syntax is confusing to non-ruby folks

ruby/lib/google/protobuf/descriptor_dsl.rb
haberman: Done.

src/google/protobuf/compiler/ruby/ruby_generator.cc
haberman: Done.

ruby/lib/google/protobuf/descriptor_dsl.rb
haberman: Done.

fowles (approved):

Peredery:
so...

/usr/local/bundle/gems/google-protobuf-3.18.0/lib/google/protobuf/descriptor_dsl.rb:58:in `add_serialized_file': Unable to build file to DescriptorPool: duplicate file name (groups.proto) (Google::Protobuf::TypeError)

PR #8851, opened 2021-07-31
cmake: support `MSVC_RUNTIME_LIBRARY` property

compnerd:
When using a new enough CMake (3.15+) prefer to use the
`MSVC_RUNTIME_LIBRARY` property on targets to select the runtime
library variant.  This property is automatically set to the value
specified by `CMAKE_MSVC_RUNTIME_LIBRARY`.  This proper requires that
the CMake Policy 91 is set to new (see CMP0091).

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -15,6 +15,10 @@ endif ()
 if(POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif()
+# MSVC runtime library flags are selected by an abstraction.
+if(POLICY CMP0091)
+  cmake_policy(SET CMP0091 NEW)
+endif()
 
 # Project
 project(protobuf C CXX)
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @haberman

cmake/CMakeLists.txt
```diff
@@ -176,20 +176,24 @@ if (protobuf_BUILD_SHARED_LIBS)
...
-    foreach(flag_var
-        CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-        CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-      if(${flag_var} MATCHES "/MD")
-        string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-      endif(${flag_var} MATCHES "/MD")
-    endforeach(flag_var)
-  endif (MSVC AND protobuf_MSVC_STATIC_RUNTIME)
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.15)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded$<$<CONFIG:Debug>:Debug>)
```
haberman, line 180: Maybe add a comment here about the intent behind this line. In particular, how does it differ from the <3.15 code?

Also do we need the `cmake_policy(SET CMP0091 NEW)` that you added in Bloaty? If not, why not?

compnerd: I'll test and verify that tomorrow.  I _think_ that we should potentially do that here as well so good catch!

compnerd: I've updated this to add the check and a comment.  Hopefully the comment is good enough, if not, we can wordsmith it.

compnerd:
Anything else left before we could merge this @haberman?

jtattermusch:
This PR seems to be breaking gRPC build in a test job that tried to continuously test against protobuf's head.

https://source.cloud.google.com/results/invocations/7ea8aefa-6622-47c6-9c2f-092b0cc87f7c/targets/github%2Fgrpc%2Ftoplevel_run_tests_invocations%2Frun_tests_c_linux_dbg_native_buildonly/tests

internal b/198323179

```
-- 3.17.3.0
CMake Error at third_party/protobuf/cmake/CMakeLists.txt:187 (if):
  if given arguments:

    "CMAKE_VERSION" "VERSION_GREATER_EQUAL" "3.15"

  Unknown arguments specified
```

Looks like this change might be broken? Should we revert?

compnerd:
I don't think that this change is broken.  The problem is likely the CMake version that is being used to build.

snnn:
If I understand correctly, when protobuf_BUILD_SHARED_LIBS is OFF and the latest cmake is in-use, this change will force users using static CRT? Previously at least they have a protobuf_MSVC_STATIC_RUNTIME for us to choose. The line 
```
    set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded$<$<CONFIG:Debug>:Debug>)
```
shouldn't exist at all.

PR #8856, opened 2021-08-03
Skip exports if not available by CommonJS

### LINKED ISSUE #8855: ReferenceError: exports is not defined
Version: v3.17.1
Language: Javascript

Windows 10 Enterprise 2004

Chrome 92

**What did you do?**
1. Include on a website via Script Tag google-protobuf.js 
2. Check the developer tools for errors. 

No errors. 

**What did you see instead?**
> google-protobuf.js:523 Uncaught ReferenceError: exports is not defined at google-protobuf.js:523

I think the current implementation of the JavaScript version is not properly portable because it only considers CommonJS and actually expects it. The [exports.js](https://github.com/protocolbuffers/protobuf/blob/master/js/commonjs/export.js) just fills some variables into a global exports variable which is not always available. Just defining one to eliminate the error can lead to unexpected side effects. 

To resolve this issue, it would be better if the library follows a more common approach for libraries like defining the exports as UMD (Universal Module Definition) to reach maximum compatibility across environments like Node, ES6, Browsers and other module systems. A replacement coud look like this: 

```js
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
    typeof define === 'function' && define.amd ? define(['exports'], factory) :
    (global = typeof globalThis !== 'undefined' ? globalThis : global || self, factory(global));
}(this, (function (exports) {

    exports.Map = jspb.Map;
    exports.Message = jspb.Message;
    exports.BinaryReader = jspb.BinaryReader;
    exports.BinaryWriter = jspb.BinaryWriter;
    exports.ExtensionFieldInfo = jspb.ExtensionFieldInfo;
    exports.ExtensionFieldBinaryInfo = jspb.ExtensionFieldBinaryInfo;

    // These are used by generated code but should not be used directly by clients.
    exports.exportSymbol = goog.exportSymbol;
    exports.inherits = goog.inherits;
    exports.object = {extend: goog.object.extend};
    exports.typeOf = goog.typeOf;
})));
```

The minimum version would be to check the existance of CommonJS to avoid any JS error: 
```js
if( typeof exports === 'object' ) {
    exports.Map = jspb.Map;
    exports.Message = jspb.Message;
    exports.BinaryReader = jspb.BinaryReader;
    exports.BinaryWriter = jspb.BinaryWriter;
    exports.ExtensionFieldInfo = jspb.ExtensionFieldInfo;
    exports.ExtensionFieldBinaryInfo = jspb.ExtensionFieldBinaryInfo;

    // These are used by generated code but should not be used directly by clients.
    exports.exportSymbol = goog.exportSymbol;
    exports.inherits = goog.inherits;
    exports.object = {extend: goog.object.extend};
    exports.typeOf = goog.typeOf;
}
```

Danielku15:
Resolves #8855 

This is a minimal fix for the problem reported in the above issue. I added a simplistic check whether the exports variable is an object to be filled. This eliminates the error without making big changes to really extend to an UMD. Proper UMD support (also in the generated code) would likely need a bigger change in code consisting of changes in the file structure, code generator (protoc), tests etc. 

My primary goal is to eliminate the browser error.


Danielku15:
@googlebot I signed it!

Danielku15:
@lukesandberg @acozzette Any chance this will get reviewed and integrated in the near future. Otherwise I we will need likely to either patch locally the library or fully replace it with some other lib. 😢

Danielku15:
> We will try to keep our response time within 7-days but if you don’t get any response in a few days, feel free to comment on the threads to get our attention. (https://github.com/protocolbuffers/protobuf/blob/master/CONTRIBUTING.md)

@lukesandberg  @acozzette 
It has been now quite a while since I reported this issue and provided a PR to fix it. I saw some activities on the labels but beside that the replies are rather slient 😢  Since quite some sprints were are now shifting an open bug report on our side related to this page error. I would appreciate any feedback on whether you plan to integrate this change or whether we are better set with a different framework.

acozzette:
@Danielku15 Sorry for the delay and thank you for the fix!

PR #8859, opened 2021-08-03
Disallow missing field initializers

tamird:
This warning is enabled in Fuchsia, and appeasing it is quite trivial.

@haberman @fowles

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_helpers.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_helpers.cc
@@ -1130,7 +1130,7 @@ bool IsImplicitWeakField(const FieldDescriptor* field, const Options& options,
 
 MessageAnalysis MessageSCCAnalyzer::GetSCCAnalysis(const SCC* scc) {
   if (analysis_cache_.count(scc)) return analysis_cache_[scc];
-  MessageAnalysis result{};
+  MessageAnalysis result;
   if (UsingImplicitWeakFields(scc->GetFile(), options_)) {
     result.contains_weak = true;
   }
--- a/BUILD
+++ b/BUILD
@@ -48,6 +48,7 @@ COPTS = select({
     "//conditions:default": [
         "-DHAVE_PTHREAD",
         "-DHAVE_ZLIB",
+        "-Wmissing-field-initializers",
         "-Woverloaded-virtual",
         "-Wno-sign-compare",
     ],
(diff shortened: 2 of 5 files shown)


src/google/protobuf/arena.cc
```diff
@@ -115,7 +115,11 @@ SerialArena::SerialArena(Block* b, void* owner) : space_allocated_(b->size) {
 SerialArena* SerialArena::New(Memory mem, void* owner) {
   GOOGLE_DCHECK_LE(kBlockHeaderSize + ThreadSafeArena::kSerialArenaSize, mem.size);
 
-  auto b = new (mem.ptr) Block{nullptr, mem.size};
+  auto b = new (mem.ptr) Block{
+      .next = nullptr,
+      .size = mem.size,
+      .start = nullptr,
+  };
```
fowles, line 122: Designated initializers are a c++20 extension. We support c++11

src/google/protobuf/compiler/cpp/cpp_helpers.cc
```diff
@@ -1130,7 +1130,7 @@ bool IsImplicitWeakField(const FieldDescriptor* field, const Options& options,
 
 MessageAnalysis MessageSCCAnalyzer::GetSCCAnalysis(const SCC* scc) {
   if (analysis_cache_.count(scc)) return analysis_cache_[scc];
-  MessageAnalysis result{};
+  MessageAnalysis result = {};
```
fowles, line 1133: This one should be unneeded

src/google/protobuf/arena.cc
tamird: Done.

src/google/protobuf/compiler/cpp/cpp_helpers.cc
tamird: You're right - not sure what I was seeing, but it doesn't seem to be needed.

src/google/protobuf/compiler/cpp/cpp_helpers.cc
```diff
@@ -1130,7 +1130,7 @@ bool IsImplicitWeakField(const FieldDescriptor* field, const Options& options,
 
 MessageAnalysis MessageSCCAnalyzer::GetSCCAnalysis(const SCC* scc) {
   if (analysis_cache_.count(scc)) return analysis_cache_[scc];
-  MessageAnalysis result{};
+  MessageAnalysis result{false, false, false, false, false};
```
fowles, line 1133: blech, I kinda prefer adding defaults to this struct instead

tamird: Done.

fowles (approved):

PR #8862, opened 2021-08-05
Added Java JsonFormat.merge() overload for JsonElement

TheStormN:
JsonFormat.merge() already supports plain text and Reader, but there are cases which I've faced where I already have parsed the JSON and I just need to put the `JsonElement` to the parser which is currently impossible with the public API, although it is well supported by the private one.

What I did here is just a very simple new function to expose the `JsonElement` to the public API.

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -490,6 +490,16 @@ public class JsonFormat {
           .merge(json, builder);
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


TheStormN:
It seems that I cannot add labels...

perezd:
We appreciate the PR! However, we're hesitant to expose a public API that is directly reliant on a GSON type. We consider GSON an internal implementation detail and subject to be replaceable, which creating this API would make hard to do in the future. We're not going to accept this one but, thank you for the contribution nonetheless.

Issue #8867, opened 2021-08-09
Bazel Rule Support for Kotlin


jamesward:
To support generating Kotlin protos via Bazel, we need some support for the classes which the generated Kotlin protos use:
https://github.com/protocolbuffers/protobuf/tree/master/java/kotlin

Can be inspired by: https://github.com/protocolbuffers/protobuf/blob/master/java/core/BUILD

This will be used by new rules which will have examples:
https://github.com/grpc/grpc-kotlin/pull/266

Please assign to: @deannagarcia

SanjayVas:
I think this should stay open, as what's in grpc-kotlin is a workaround. Note that this issue isn't for Kotlin protobuf Bazel rules themselves, but rather the fact that such rules depend on protobuf-kotlin. Currently, the way to get this dependency is as a pre-built Maven artifact (`com.google.protobuf:protobuf-kotlin`), whereas it would be cleaner for Bazel projects to be able to build this artifact from source.

PR #8868, opened 2021-08-09
Adding a build file to java/kotlin package

deannagarcia:
Adding a bazel build file to the java/kotlin package so that it can be imported, see #8867 for more information on the request.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -26,6 +26,20 @@ http_archive(
     ],
 )
 
+rules_kotlin_version = "v1.5.0-beta-3"
+rules_kotlin_sha = "58edd86f0f3c5b959c54e656b8e7eb0b0becabd412465c37a2078693c2571f7f"
+http_archive(
+    name = "io_bazel_rules_kotlin",
+    urls = ["https://github.com/bazelbuild/rules_kotlin/releases/download/%s/rules_kotlin_release.tgz" % rules_kotlin_version],
+    sha256 = rules_kotlin_sha,
+)
+
+load("@io_bazel_rules_kotlin//kotlin:repositories.bzl", "kotlin_repositories")
+kotlin_repositories()
+
+load("@io_bazel_rules_kotlin//kotlin:core.bzl", "kt_register_toolchains")
+kt_register_toolchains()
+
 # Load common dependencies.
 load("//:protobuf_deps.bzl", "protobuf_deps")
 protobuf_deps()
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -26,20 +26,6 @@ http_archive(
     ],
 )
 
-rules_kotlin_version = "v1.5.0-beta-3"
-rules_kotlin_sha = "58edd86f0f3c5b959c54e656b8e7eb0b0becabd412465c37a2078693c2571f7f"
-http_archive(
-    name = "io_bazel_rules_kotlin",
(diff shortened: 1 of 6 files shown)


WORKSPACE
```diff
@@ -26,6 +26,20 @@ http_archive(
     ],
 )
 
+rules_kotlin_version = "v1.5.0-beta-3"
+rules_kotlin_sha = "58edd86f0f3c5b959c54e656b8e7eb0b0becabd412465c37a2078693c2571f7f"
+http_archive(
```
perezd, line 31: Can some of this live in protobuf_deps() ? This is primarily to make usage of these rules less difficult in consuming bazel workspaces.

java/kotlin/BUILD
```diff
@@ -0,0 +1,21 @@
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_library")
+
+KT_SRCS = [
```
perezd, line 3: unless you need to reference this/filter it multiple ways, prefer to just have this list directly in the rule `srcs` field.

java/core/BUILD
```diff
@@ -118,6 +118,18 @@ java_library(
     srcs = LITE_SRCS,
 )
 
+java_library(
```
perezd, line 121: This shouldn't be needed, kt rules should be able to rely on `core` and `lite`, I think this is identical, and generally its a bad practice to have srcs in multiple targets as this proposes. If you can't depend on java_export in a kt_ that is likely a bug we need to fix.

perezd (changes requested):

java/core/BUILD
perezd: Internally java_export uses java_library, so it should be OK: https://github.com/bazelbuild/rules_jvm_external/blob/master/private/rules/java_export.bzl#L66

perezd:
Should we also have kotlin lite targets? Might be good to land them at the same time to ensure they are working as expected with the exclusions etc.

WORKSPACE
deannagarcia: Yep, moved the http_archive rule into protobuf_deps

java/kotlin/BUILD
deannagarcia: I changed it a little and ended up using it in a couple different places

java/core/BUILD
deannagarcia: Ok changed to now rely on existing rules and removed this new rule

deannagarcia:
> Should we also have kotlin lite targets? Might be good to land them at the same time to ensure they are working as expected with the exclusions etc.

Added kotlin lite targets!

java/kotlin/BUILD
```diff
@@ -0,0 +1,36 @@
...
+    "src/main/kotlin/com/google/protobuf/DslProxy.kt",
+    "src/main/kotlin/com/google/protobuf/ExtendableMessageLiteExtensions.kt",
+    "src/main/kotlin/com/google/protobuf/ExtensionList.kt",
+    "src/main/kotlin/com/google/protobuf/OnlyForUseByGeneratedProtoCode.kt",
+    "src/main/kotlin/com/google/protobuf/ProtoDslMarker.kt",
+    "src/main/kotlin/com/google/protobuf/UnmodifiableCollections.kt",
+]
+
+kt_jvm_library(
+    name = "kt_lite",
```
perezd, line 15: unless there's a stylistic reason to do this, I'd drop the kt_ prefix, since the target path contains "kotlin" in the name already.

PR #8874, opened 2021-08-12
Remove .bazelrc file containing the -std=c++11 flag

acozzette:
MSVC warns about that flag since it's not in the format it expects, and
the flag seems to be unnecessary anyway.


perezd (approved):
LGTM!

PR #8875, opened 2021-08-12
Remove workaround for symlink

deannagarcia:


deannagarcia:
> So, I wasn't suggesting removing the entire thing. Just providing the exclusion for the `_build` folder:
> https://docs.bazel.build/versions/main/be/functions.html#glob

Yep, I know just playing around in this draft PR to figure out how that works and then I will send a PR your way when I have it figured out.

perezd:
OH OK, yes, sorry, I kept reading this as in master for some reason,
disregard!

PR #8880, opened 2021-08-15
files ...tar.gz or ..tar.gz cause trouble

picobyte:
Probably pedantic, but with a file named '...tar.gz' this script seems to do a

```
chmod -R u+w ..
cd ..

# etc..
```
Which I think should be prevented.

### DIFF
--- a/post_process_dist.sh
+++ b/post_process_dist.sh
@@ -25,6 +25,12 @@ if [ ! -e $1 ]; then
   exit 1
 fi
 
+if [ "${1:0:2}" == ".." ]; then
+  echo "Illegal filename" >&2
+  exit 1
+fi
+
+
 set -ex
 
 LANGUAGES="cpp csharp java js objectivec python ruby php all"


picobyte:
I don't care about your policies, this is a simple fix, take it or leave it.

acozzette:
Sorry but unfortunately we do have to follow the policy and so we can't accept any pull requests without a CLA.

picobyte:
I signed your CLA

acozzette:
@picobyte Did you sign with a personal CLA or a corporate one? Our CLA bot still seems to think it hasn't been signed, but there could be various reasons for that which I can try to troubleshoot.

picobyte:
personal. I work at a company, but this is a private account

acozzette:
I think the problem might be that the email address you used to sign the CLA has to be attached to your GitHub account.

picobyte:
I believe my e-mail is associated, but I may have used a different case or
added dots to the gmail adress, though. I do that to thwart spam, the email
is delivered, but due to the particular format I can resolve the origin of
the email. In this case it may be a problem though, if the e-mail doesn't
exactly match.

If you consider my patch, is it worthwhile? It's pretty simple and should
provide a small safeguard, in case someone does something stupid like
processing a file with this particular fabricated name which could
theoreticly do a little damage in the parent directory. Maybe anyone with
some knowledge can review and pick it up or not and sign it him or herself.
I don't  mind. I just came across this because I googled this error that I
came across in my own code, and considered it was slightly unsafe.
Best regards, Roel

acozzette:
I think I understand the problem you're trying to prevent, but preventing that shouldn't be strictly necessary because the script doesn't need to handle untrusted inputs.

PR #8881, opened 2021-08-16
Implement RepeatedPtrField<T>::Truncate

georgthegreat:
This makes RepeatedPtrField behavior similar to RepeatedField.

### DIFF
--- a/src/google/protobuf/repeated_field.h
+++ b/src/google/protobuf/repeated_field.h
@@ -652,6 +652,15 @@ class PROTOBUF_EXPORT RepeatedPtrFieldBase {
 
   void Reserve(int new_size);
 
+  template<typename TypeHandler>
+  void Truncate(int new_size) {
+    GOOGLE_DCHECK_LE(new_size, current_size_);
+    for (int i = new_size; i < current_size_; i++) {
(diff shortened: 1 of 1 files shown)


georgthegreat:
I also have a basic test for the newly added routine.
The test uses `RepeatedPtrField<std::string>` and runs using gtest framework.

Could you, please, advide me where such test should be added?

acozzette:
@georgthegreat Thank you for the pull request, but unfortunately I don't think we will want to accept this change. Your code looks great, but the thing is just that we tend to be reluctant to expand the API surface on a widely-used type like `RepeatedPtrField` unless there is a highly compelling reason. You can already perform a truncation like this by calling `field.DeleteSubrange(new_size, size - new_size)`, so that is relatively easy to do without needing a new member function. Internally we have a `repeated_field_util.h` header that contains a bunch of standalone helper functions for working with repeated fields and includes a `Truncate` function that works this way. If there is enough demand for it then we could potentially open source that header in the future.

georgthegreat:
BTW, I am not consent with the above statement.
I fully understand that adding new API methods to basic classes could be a bad decision, but this PR does not add any.
Instead, it makes two RepeatedField's interface symmetrical.

E. g. std::vector provides same API no matter wrether it is instantiated with trivial structure or not.

PR #8889, opened 2021-08-19
Updtae version to 3.17.3

elharo:
@TeBoring


TeBoring:
What issue does 3.15.3 have?

elharo:
It probably has issues, but mostly it's just not the latest by several versions. We should look into making version updates in the docs part of our release process.

PR #8905, opened 2021-08-26
add PROTOBUF_ERROR_CODE_CHECK_WARNING to prevent missusing parse/serialize methods with not checked error-codes

ilnurkh:
A lot of methods in the library (for example: Message interface) return bool to identify existing of errors during method processing. Ignoring return value of this methods is very common and hard detectable error. The added macro add standertized attribute for functions, that lead to 'ignoring return value' warning. Very useful with -Werror. To diable this check (in case of upgrade to new release of protocolbufers during vendoring) just change port_def.inc file or add -Wno-unused-result to your c++ compiling flags.


ilnurkh:
@acozzette can you please refresh the status?

ilnurkh:
Sorry for this misspell, i've fixed it (and one more).

ilnurkh:
@acozzette is there any other issues that I should fix before merging ?

ilnurkh:
=(

acozzette:
@ilnurkh Sorry for the delay, but I need to do some more research before I can determine whether we can accept this change. I need to figure out if it's feasible for us to get our large internal codebase compliant with the warnings.

ilnurkh:
@acozzette in the current version no any warnings will be produced, I've changed id (additional patch or compile comands change required)

ilnurkh:
@acozzette @jorgbrown  sorry for annoying.
it would be very useful to have any estimations or requirements (i remember: it is actually noop commit for now)

ilnurkh:
up?

acozzette:
@ilnurkh Unfortunately I don't think we will be able to accept the change at this time. Before adding these warnings we want to have some confidence that we can migrate our monorepo to comply with the warnings, but no one on the team has time right now to work on this investigation. Even if the warning is optional and controlled by a flag, we don't want to add it until we're sure that it is feasible for existing large projects to enable the warning.

ilnurkh:
> it is feasible for existing large projects to enable the warning.

I would like to notice, that project I have in mind (yandex monorepo) is large enough, and it is look possible to us to enable it (even with out policy warning-as-error build policy), of course after large preparation work (but i think it can be done in several weeks, because the problem is easily detectable by build erros).

------------------

Do you have any ideas how to have both:
1. do not detach protobuf-lib in our repo with upstream
2. solve the problem of not checked parsings calls

-------------

would it help if the `[[nodiscard]] flag` will be set for only one method ? (i think it is acceptable to add one method per month, for example)

fowles:
We already have

PROTOBUF_MUST_USE_RESULT
PROTOBUF_FUTURE_BREAKING_CHANGES

Externally, we mark MUST_USE_RESULT as a no-op to avoid breaking folks, but we could change it to be guarded by FUTURE_BREAKING_CHANGES.

Once that is done, we could then incrementally add MUST_USE_RESULT to the apis suggested.  This will likely be slightly slow on our end as each time we do this we have to clean up Google's codebase to be clean of the new errors.

As a path forward would that satisfy you?  It is likely easier if we implement the first part ourselves as you don't have access to the internal parts of MUST_USE_RESULTS that we have.

ilnurkh:
1. don't creating a new macro and use PROTOBUF_MUST_USE_RESULT - of course is a good point, i should change the patch to use this one
2. what if i change current patch 
from

```
-bool SomeApiCall(...)
+PROTOBUF_ERROR_CODE_CHECK_WARNING bool SomeApiCall(...)
```

to

```
+#ifdef FUTURE_BREAKING_CHANGES
+PROTOBUF_MUST_USE_RESULT
+#endif
bool SomeApiCall(...)
```
?

(or just change definition of proposed PROTOBUF_MUST_USE_RESULT to
```
#ifdef FUTURE_BREAKING_CHANGES
define PROTOBUF_ERROR_CODE_CHECK_WARNING PROTOBUF_MUST_USE_RESULT
#endif
```
)

Will it make the patch to be okay? (as the change is guarded by FUTURE_BREAKING_CHANGES )

I am not quite caught what kind of build in google's codebase do you firstly care of:
1. looks like PROTOBUF_MUST_USE_RESULT  is defined to something not empty (the same should do I in my codebase)
2. but I am not shure is FUTURE_BREAKING_CHANGES set to 1 or not ?
--------
If your ask, is redefining FUTURE_BREAKING_CHANGES and PROTOBUF_MUST_USE_RESULT  in our projects is acceptable - yes, of course.

ilnurkh:
I'have found small problem with PROTOBUF_MUST_USE_RESULT : current usage is set after return type in functions, so it is illegal to define this macro as [[nodiscard]], but there is a simple workaround. https://godbolt.org/z/YM3dGe6vf
Did any one have in mind this kind of problems, then some additional work  in internal google's codebase was mentioned?

fowles:
We support a variety of toolchains, some of which are a bit persnickity about ordering of attributes.  We need to test alternate formulations a bit internally.

ilnurkh:
Any updates? Does any one have an opinion about reality of adding ~~PROTOBUF_MUST_USE_RESULT~~ PROTOBUF_NODISCARD  to this common methods?

(In our codebase we've succeeded to suppress warning in old code => and we would like in future to have not diff with the upstream)

fowles:
Sorry this got lost in my inbox.  There are two issues that make this particularly difficult:
1) the variety of toolchains supported are a bit persnickety about placement of these attributes/macros
2) we need to cleanup Google's code base to be clean as part of these efforts.

Given these two, I think that we won't move forward with this patch, but will start the cleanups on our end with the hope of getting something like it released in principle.

georgthegreat:
We have these changes applied internally.
In order to migrate the entire monorepo within a reasonable timespan, we have added a helper macro, whose implementation can be found e. g. [here](https://github.com/catboost/catboost/blob/7656df832c81e1ae200325c8368c284c84b12b85/contrib/libs/protobuf/src/google/protobuf/stubs/common.h#L61).

We can upstream monorepo independent version of this macro if it will be found useful.

As for the number of toolchains supported, I suggest to just switch off the warning for the toolchains which do no support it or require incompatible wording.

acozzette:
I'm going to go ahead and close this--this is a nice change and something we need to do eventually, but we have to figure out the cleanups on our end before we can make this change.

pg83:
>but we have to figure out the cleanups on our end before we can make this change.
Why you can not just accept this proposal, without using it at "large internal code base" (e.g. Google)? We have to maintain our patchset against vanilla protobuf, and it is not really convenient.

pg83:
Can you clarify - protobuf is open source project, driven by users needs, or Google projects, which accept only what can be applied to Google monorepo?

Issue #6159, opened 2019-05-21
google/protobuf/arena.h(692): error C3857: 'Map': multiple template parameter lists are not allowed


limnax:
I get this error when i try to build my project that uses tensorflow libraries.My code is in c++11 and the environment is windows 10.

I built tensorflow from source and completed,i also built protobuf v3.7.0 from source but it gives me the same error.I also installed protobuf using vcpkg but it returns the same error.

This is the error iam getting mocs_compilation.cpp **C:\libs\vcpkg\installed\x64-windows\include\google/protobuf/arena.h(692): error C3857: 'Map': multiple template parameter lists are not allowed** 
C:\libs\vcpkg\installed\x64-windows\include\google/protobuf/stubs/strutil.h(359): warning C4127: conditional expression is constant 
C:\libs\vcpkg\installed\x64-windows\include\google/protobuf/stubs/strutil.h(359): note: consider using 'if constexpr' statement instead 
C:\libs\vcpkg\installed\x64-windows\include\google/protobuf/stubs/strutil.h(366): warning C4127: conditional expression is constant 
C:\libs\vcpkg\installed\x64-windows\include\google/protobuf/stubs/strutil.h(366): note: consider using 'if constexpr' statement instead
 c:\libs\vcpkg\installed\x64-windows\include\unsupported\eigen\cxx11\src/Tensor/Tensor.h(76): warning C4554: '&': check operator precedence for possible error; use parentheses to clarify precedence
 c:\libs\vcpkg\installed\x64-windows\include\unsupported\eigen\cxx11\src/Tensor/TensorMap.h(33): note: see reference to class template instantiation 'Eigen::Tensor' being compiled

BSBandme:
I think `if constexpr` only available after c++ 17, and it might not be fixed to that since we need to be compatible with c++ 11

Sil3ntStorm:
While `if constexpr` is only available starting with C++17 and you may want to remain compatible with C++11, it would be good if this issue could be fixed.
Especially considering that this issue forces users of the library to either globally disable a useful warning or litter the code with 10+ lines of conditional warning disabling around **every inclusion** of problematic protobuf headers.

Preferably by conditionally marking things that can be `constexpr`, something along the lines of
```cpp
#if defined(__cpp_if_constexpr) && __cpp_if_constexpr >= 201606L
#define CEXPR constexpr
#else
#define CEXPR
#endif
```
[...]

PR #8917, opened 2021-08-27
Use if constexpr in template functions

Sil3ntStorm:
This allows users to not have to disable certain warnings (e.g. 4127 on MSVC)

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -766,6 +766,8 @@
 #pragma warning(disable : 4073)
 // To silence the fact that we will pop this push from another file
 #pragma warning(disable : 5031)
+// Conditional expression is constant
+#pragma warning(disable: 4127)
 #endif
 
 // We don't want code outside port_def doing complex testing, so
--- a/src/google/protobuf/stubs/stringprintf.cc
+++ b/src/google/protobuf/stubs/stringprintf.cc
@@ -44,14 +44,11 @@ namespace google {
 namespace protobuf {
 
 #ifdef _MSC_VER
-enum { IS_COMPILER_MSVC = 1 };
 #ifndef va_copy
 // Define va_copy for MSVC. This is a hack, assuming va_list is simply a
 // pointer into the stack and is safe to copy.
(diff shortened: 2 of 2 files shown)


Sil3ntStorm:
Any particular reason for not wanting to eliminate the cause of the warning?
While it seems like this is something other compilers ignore for some reason, shouldn't it generally be beneficial to do things at compile time rather than runtime?

If we go with disabling/ignoring warnings, can we agree to include all warnings (currently) emitted by protobuf code?

haberman:
This will already be evaluated at compile time if optimizations are enabled: https://godbolt.org/z/EfP95oPY1

`#ifdef` is best to avoid when possible, because it requires multiple builds with different build configurations to fully test.

`#ifdef` conditioned on C++ version is especially problematic when it affects header files. It's technically an ODR violation if the function definition changes when the build switches between C++11 and C++17, for example if protobuf is compiled as C++11 but a user application is compiled as C++17.

Protobuf primarily targets C++11. While we will occasionally opt into certain features of newer C++ versions through `#ifdef`, we generally only do this when it achieves some functional goal that is impossible in C++11. In this case it's just forcing the compiler to do something it would have done anyway if optimizations were enabled.

> If we go with disabling/ignoring warnings, can we agree to include all warnings (currently) emitted by protobuf code?

I think we should evaluate this on a case-by-case basis. If the warning can be fixed without `#ifdef`, it may be better to fix the code instead.

Sil3ntStorm:
As it turns out, it seems as though doing it there is not at all inclusive. Even adding every single warning in protobuf code to the list specified, there are 360 warnings (that are disabled in proto_def.inc) compiling protobuf.

Sil3ntStorm:
>Protobuf primarily targets C++11. While we will occasionally opt into certain features of newer C++ versions through #ifdef, we generally only do this when it achieves some functional goal that is impossible in C++11. In this case it's just forcing the compiler to do something it would have done anyway if optimizations were enabled.

Okay fair enough. I will check and update this PR accordingly in a while if this works out for the constexpr warning.

Sil3ntStorm:
This should take care of (at least part of) #6159

PR #8927, opened 2021-08-29
Fix unreachable code return in a case label

Sil3ntStorm:


TeBoring:
Please improve the PR description. Currently it looks like an overflow from the title.

Sil3ntStorm:
I was just editing the text, but I see you removed it already.

TeBoring (approved):

PR #8929, opened 2021-08-29
Fix warning about unary minus operator

Sil3ntStorm:


src/google/protobuf/compiler/parser.cc
```diff
@@ -1549,7 +1549,7 @@ bool Parser::ParseOption(Message* options,
           value_location.AddPath(
               UninterpretedOption::kNegativeIntValueFieldNumber);
           uninterpreted_option->set_negative_int_value(
-              static_cast<int64_t>(-value));
+              static_cast<int64_t>(0 - value));
```
elharo, line 1552: what warning? The original code seems clearer to me but I'm no C++ expert.

elharo (changes requested):

src/google/protobuf/compiler/parser.cc
fowles: Since value is a `uint64_t`, the idea of negating it doesn't have intrinsic meaning.  unsigned underflow is well defined in C++, so they defined it to be `0 -`, but some toolchains warn because of the first property.

Sil3ntStorm: As @fowles mentioned, the value is unsigned so it cannot be below 0. The exact warning text is `unary minus operator applied to unsigned type, result still unsigned`.
See also the [Microsoft documentation](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=msvc-160).
As for clarity, personally I find `0 - x` to be more readable and intuitive than having `-x`, but that might just be me.

elharo: For me, now that you've pointed out it's an unsigned int I find both equally confusing and surprising, maybe just because I'm not used to C++ conventions.

PR #8941, opened 2021-08-31
Addresses duplicates identified in #8925 (#8939)

TeBoring:
* Addresses duplicates identified in #8925

java_export aliases the rule name to the full maven_project_jar which consolidates all deps into a merged jar. This is not what we want for Bazel users.

* Update :util to match


perezd (approved):
LGTM!

PR #8944, opened 2021-09-01
fix arm64 tests broken by dockcross image update

jtattermusch:
Currently ARM64 tests are broken for C++, java, javascript, python, php and ruby (e.g. https://source.cloud.google.com/results/invocations/33696bce-f922-4572-bc3e-9a369ce3d721)
Pinning dockcross image to a version that's known to work fixes them and also avoid issues with the tests suddenly breaking in the future.

### DIFF
--- a/kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
+++ b/kokoro/linux/aarch64/dockcross_helpers/run_dockcross_linux_aarch64.sh
@@ -12,11 +12,14 @@ else
   DOCKER_TTY_ARGS=
 fi
 
+# Pin the dockcross image since newer versions of the image break the build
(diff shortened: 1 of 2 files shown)


jtattermusch:
Adhoc test results: 
https://fusion2.corp.google.com/invocations/d94b1b54-c6b3-47a0-b426-2a904e89c374/targets
https://sponge2.corp.google.com/ab566eac-4cdc-425b-bfaa-7d79cf09c6ba

elharo (approved):
Do we know why the newer versions broke these?

jtattermusch:
> Do we know why the newer versions broke these?

I haven't investigated yet, but answer can probably be found in https://github.com/dockcross/dockcross changes and the dockerfile tags history: https://hub.docker.com/r/dockcross/manylinux2014-aarch64/tags?page=1&ordering=last_updated 

The priority for now was to make the tests green again.

PR #8945, opened 2021-09-01
Fix unused variable warnings

Sil3ntStorm:


acozzette:
@Sil3ntStorm Thanks for the cleanup! Unfortunately I don't think we can accept all the changes in the `compiler` subdirectory, but the other changes should be OK. The problem is that often an unused parameter appears because we have some internal-only feature that relies on it (we carry a bunch of internal-only diffs). We can try to clean them up over time, but it would require a lot of work on our end to fix all of them in the compiler all at once. If you can restrict the change to everything outside the `compiler` directory then that should be doable, though. This way you can also turn on as many warnings as you want for your own project, as long as you build `protoc` with more lenient warning flags.

Sil3ntStorm:
Quite unfortunate, as that is most of the changes. I will need to look into this some more next week. Unfortunately `[[maybe_unused]]` is also not an option, since protobuf needs to support C++ 11 iirc.

acozzette:
We could potentially try to do the `compiler` changes in a later step, but I think it's just too much to do all at once. If you want to save the `compiler` changes and send them in another PR then I can try to work on that later.

Sil3ntStorm:
@acozzette this should now only include the non compiler changes.
Unrelated curiosity question: When can we expect merges into master to land in a public release?

acozzette:
Thanks, @Sil3ntStorm. We don't have a set schedule for releases, but we usually do one about every three months or so.

PR #8948, opened 2021-09-01
build: fix the support for CMake[<3.7]

compnerd:
The `VERSION_GREATER_EQUAL` operator was introduced in CMake 3.7.0
(released in 2016).  Use the long form of the check by splitting the
operation into two comparisons to support the older CMake.


compnerd:
CC: @jtattermusch

jtattermusch:
I can confirm that the grpc_build_protobuf_at_head has been fixed by this and we're back to green
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:grpc%2Fcore%2Fmaster%2Flinux%2Fgrpc_build_protobuf_at_head

PR #7923, opened 2020-09-28
Fill out JRuby support

rdubya:
This gets the Ruby gem to be able to be used on JRuby. Almost all Ruby tests are passing with a couple minor changes. There are a significant number of compatibility tests that are failing, but I figured this would at least get something in place that could be expanded upon by others in the community.

This is related to the following issues:

- https://github.com/protocolbuffers/protobuf/issues/1594
- https://github.com/protocolbuffers/protobuf/issues/1820
- https://github.com/protocolbuffers/protobuf/issues/3090
- https://github.com/protocolbuffers/protobuf/issues/3650
- https://github.com/protocolbuffers/protobuf/issues/7831

PR #8954, opened 2021-09-03
Include JRuby gem when building gem:native target

JasonLunn:
Since this is the rake target used by kokoro, this will close the gem release gap for JRuby

Should help #7923 by making sure JRuby is built and released as part of the normal cadence.

### DIFF
--- a/ruby/Rakefile
+++ b/ruby/Rakefile
@@ -67,16 +67,18 @@ unless ENV['IN_DOCKER'] == 'true'
 end
 
 if RUBY_PLATFORM == "java"
-  if `which mvn` == ''
-    raise ArgumentError, "maven needs to be installed"
-  end
-  task :clean do
+  task :clean => :require_mvn do
     system("mvn --batch-mode clean")
   end
 
-  task :compile do
+  task :compile => :require_mvn do
     system("mvn --batch-mode package")
   end
+
+  task :require_mvn do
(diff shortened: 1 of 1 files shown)


JasonLunn:
@acozzette - could you rerun the kokoro tests with the second commit included?

ruby/Rakefile
```diff
@@ -95,7 +97,18 @@ else
     ]
   end
 
+  task 'gem:java' do
```
perezd, line 100: ulta nit: should this be jruby to be more clear?

perezd (approved):
LGTM

ruby/Rakefile
JasonLunn: I was inclined to align with the terminology used in the [`Gemfile`](https://github.com/protocolbuffers/protobuf/blame/master/ruby/google-protobuf.gemspec#L14) but I don't feel strongly about it.

perezd: sure thing, was just thinking about future travelers that are reading this, what does "java" mean vs jruby is a clearer target, at least for me.

JasonLunn: @haberman / @acozzette - any preference between `gem:java` vs `gem:jruby`?

acozzette: Either one sounds OK to me.

JasonLunn:
Obviated by #9047

PR #8959, opened 2021-09-09
Update Field Presence doc

VeryStrongFingers:
Updated information surrounding the usage of `optional` and the `--experimental_allow_proto3_optional` flag with relevant information between v3.12 & 3.15.

Since release v3.15:
>  Optional fields for proto3 are enabled by default, and no longer require
>  the --experimental_allow_proto3_optional flag.

Now the doc infers the feature is implicitly enabled, rather than an experimental flag which was only true up until v3.15.0

see notes on optional/`--experimental_allow_proto3_optional` here:
- https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0
- https://github.com/protocolbuffers/protobuf/releases/tag/v3.12.0

### DIFF
--- a/docs/field_presence.md
+++ b/docs/field_presence.md
@@ -198,7 +198,7 @@ If client A depends on _explicit presence_ for `foo`, then a "round trip" throug
 
 ## How to enable _explicit presence_ in proto3
 
-These are the general steps to u field tracking support for proto3:
+These are the general steps to use field tracking support for proto3:
 
 1.  Add an `optional` field to a `.proto` file.
 1.  Run `protoc` (at least v3.15, or v3.12 using `--experimental_allow_proto3_optional` flag).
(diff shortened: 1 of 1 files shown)


VeryStrongFingers:
@googlebot I signed it!

docs/field_presence.md
```diff
@@ -198,10 +198,10 @@ If client A depends on _explicit presence_ for `foo`, then a "round trip" throug
 
 ## How to enable _explicit presence_ in proto3
 
-These are the general steps to use the experimental field tracking support for proto3:
+These are the general steps to u field tracking support for proto3:
```
haberman, line 201: I think this should be "use"?

haberman (changes requested):

docs/field_presence.md
VeryStrongFingers: Good catch. Fixed that up

haberman (approved):

PR #8960, opened 2021-09-09

haberman:


java/core/src/main/java/com/google/protobuf/TextFormat.java
```diff
@@ -445,9 +445,12 @@ private void printField(
     }
 
     /**
-     * An adapter class that can take a MapEntry or a MutableMapEntry and returns its key and entry.
-     * This class is created solely for the purpose of sorting map entries by its key and prevent
-     * duplicated logic by having a separate comparator for MapEntry and MutableMapEntry.
+     * An adapter class that can take a {@link MapEntry} and returns its key and entry.
+     * BEGIN GOOGLE-INTERNAL
```
acozzette, line 449: Something must have gone wrong with the sync here. I'm not sure how these annotations work but maybe they only take effect in a `//` style comment.

acozzette (approved):

java/core/src/main/java/com/google/protobuf/TextFormat.java
perezd: Agreed this shouldn't be here.

PR #8962, opened 2021-09-10
Added "object" as a reserved name for PHP

J0sh0nat0r:
"Object" is a reserved word as of PHP 7.2, as stated [here](https://www.php.net/manual/en/reserved.other-reserved-words.php)


J0sh0nat0r:
@googlebot I signed it!

PR #8746, opened 2021-06-18
Reintroduce setup.py changes from #8280 erased by piper import #8617

jtattermusch:
Fixes https://github.com/protocolbuffers/protobuf/issues/8667.

For an unknown reason (am I missing something?), setup.py changes from #8280 have been erased by piper import #8617 (see https://github.com/protocolbuffers/protobuf/issues/8667#issuecomment-863936224).

These changes are actually necessary for correct build of aarch64 python wheels.
without them, the aarch64 wheels will have python extension .so files with wrong arch suffix `x86_64-linux-gnu.so`,
which results in failing to load the python C++ extension.
This is also basically what the broken tests reported by #8667 have detected.

This PR fixes the problem - as proven by a green adhoc run of the kokoro job (https://fusion2.corp.google.com/invocations/c1e5e71e-e779-48e7-8f0a-23b1d33a06b0/targets/protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fpython_aarch64%2Fcontinuous/log)

One thing I haven't been able to explain is:
- why did the piper import erase the changes made to setup.py (was that intentional or perhaps a bad merge?)
- was there any other important changes that ended up being deleted?

Issue #8667, opened 2021-05-27
Python aarch64 linux tests broken by recent piper import


jtattermusch:
After this diff: https://github.com/protocolbuffers/protobuf/compare/998d3a343a87d53e1a4a4917bb01eb87943a7999...636132bd77c4afc890a294b2a751ae7b73f58622 from
https://github.com/protocolbuffers/protobuf/pull/8617

the linux/python_aarch64 test got broken (and has been failing since)
https://source.cloud.google.com/results/invocations/53e39cd8-0d19-4f6e-a06c-d81f72a0c8e2/log

jtattermusch:
The diff from piper import  seems to be reverting some of the python's setup.py changes introduced by #8280 (and those changes are necessary for generating the aarch64 wheels).

added by the aarch64 PR:
https://github.com/protocolbuffers/protobuf/pull/8280/files#diff-eb8b42d9346d0a5d371facf21a8bfa2d16fb49e213ae7c21f03863accebe0fcf

and seems to be removed by the piper import diff:
https://github.com/protocolbuffers/protobuf/compare/998d3a343a87d53e1a4a4917bb01eb87943a7999...636132bd77c4afc890a294b2a751ae7b73f58622#diff-eb8b42d9346d0a5d371facf21a8bfa2d16fb49e213ae7c21f03863accebe0fcf

PR #8966, opened 2021-09-10
Reintroduce setup.py changes from #8280 erased by piper import #8902

jtattermusch:
This fixes the broken python aarch64 linux tests (see https://source.cloud.google.com/results/invocations/861e900c-12ad-4998-9489-18f5cfbf76ee)

For some reason (bad merge?), the changes needed for crosscompiling aarch64 python wheels have been removed in piper import https://github.com/protocolbuffers/protobuf/pull/8902/files#diff-eb8b42d9346d0a5d371facf21a8bfa2d16fb49e213ae7c21f03863accebe0fcfL161.

This is actually the second time this happened (see https://github.com/protocolbuffers/protobuf/issues/8667 and the fix https://github.com/protocolbuffers/protobuf/pull/8746). What can we do to make sure this won't happen again?

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -16,6 +16,7 @@ import platform
 # namespace_packages option for the "google" package.
 from setuptools import setup, Extension, find_packages
 
+from distutils.command.build_ext import build_ext as _build_ext
 from distutils.command.build_py import build_py as _build_py
 from distutils.command.clean import clean as _clean
 from distutils.spawn import find_executable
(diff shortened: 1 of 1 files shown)


haberman:
Yes there does appear to be a bug in the tooling where setup.py changes aren't getting merged properly. We should take a look at at that.

I do wonder though, why wasn't this caught in Kokoro testing? Shouldn't there be a test verifying this behavior?

jtattermusch:
There is a test job and the failure was caught by it: https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:protobuf%2Fgithub%2Fmaster%2Fubuntu%2Fpython_aarch64%2Fcontinuous

the test doesn't run on PRs though (I didn't want to add a large number of new PR test jobs) - we can add those jobs to protobuf PRs if you think it's worthwhile.

jtattermusch:
Btw it seems that the 3.17.x release has this snippet missing:

https://github.com/protocolbuffers/protobuf/blob/3.17.x/python/setup.py

while 3.16.x and 3.18.x have it:
https://github.com/protocolbuffers/protobuf/blob/3.16.x/python/setup.py
https://github.com/protocolbuffers/protobuf/blob/3.18.x/python/setup.py

jtattermusch:
FTR I filed internal b/213024295 to track the piper import problem.

PR #8971, opened 2021-09-13
enforce android API level 14

elharo:
Add animal-sniffer plugin to check for android compatibility and revert bad commit that required Objects.nonNull.

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/ByteString.java
+++ b/java/core/src/main/java/com/google/protobuf/ByteString.java
@@ -55,7 +55,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.NoSuchElementException;
-import java.util.Objects;
 
 /**
  * Immutable sequence of bytes. Provides conversions to and from {@code byte[]}, {@link
@@ -1329,7 +1328,10 @@ public abstract class ByteString implements Iterable<Byte>, Serializable {
      * @param bytes array to wrap
      */
     LiteralByteString(byte[] bytes) {
-      this.bytes = Objects.requireNonNull(bytes);
+      if (bytes == null) {
+        throw new NullPointerException();
+      }
+      this.bytes = bytes;
     }
 
     @Override
--- a/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
+++ b/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
@@ -42,7 +42,6 @@ import java.io.OutputStream;
 import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -1155,7 +1154,9 @@ public abstract class CodedOutputStream extends ByteOutput {
     private int position;
 
     ArrayEncoder(byte[] buffer, int offset, int length) {
-      Objects.requireNonNull(buffer, "buffer");
+      if (buffer == null) {
+        throw new NullPointerException("buffer");
+      }
       if ((offset | length | (buffer.length - (offset + length))) < 0) {
         throw new IllegalArgumentException(
             String.format(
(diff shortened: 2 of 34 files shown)


java/pom.xml
```diff
@@ -166,6 +166,27 @@
           <artifactId>maven-antrun-plugin</artifactId>
           <version>1.8</version>
         </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
```
perezd, line 170: This is great! Do you think there's any way we can enforce this with Bazel checks as well?

java/core/src/main/java/com/google/protobuf/ByteString.java
```diff
@@ -1329,7 +1328,10 @@ private String truncateAndEscapeForDisplay() {
      * @param bytes array to wrap
      */
     LiteralByteString(byte[] bytes) {
-      this.bytes = Objects.requireNonNull(bytes);
```
perezd, line 1332: should this use  Internal.CheckNotNull?

java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
```diff
@@ -1155,7 +1154,9 @@ public final void writeRawLittleEndian64(final long value) throws IOException {
     private int position;
 
     ArrayEncoder(byte[] buffer, int offset, int length) {
-      Objects.requireNonNull(buffer, "buffer");
+      if (buffer == null) {
```
perezd, line 1157: should this use internal.checkNotNulll?

java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
```diff
@@ -2392,7 +2396,10 @@ final void bufferFixed64NoTag(long value) {
 
     ByteOutputEncoder(ByteOutput out, int bufferSize) {
       super(bufferSize);
-      this.out = Objects.requireNonNull(out, "out");
+      if (out == null) {
```
perezd, line 2399: should this use internal.checkNotNull?

java/core/src/main/java/com/google/protobuf/LazyFieldLite.java
```diff
@@ -428,4 +429,13 @@ protected void ensureInitialized(MessageLite defaultInstance) {
       }
     }
   }
+
+  private static void checkArguments(ExtensionRegistryLite extensionRegistry, ByteString bytes) {
```
perezd, line 433: should these use Internal.checkNotNull?

java/core/src/main/java/com/google/protobuf/MessageSchema.java
```diff
@@ -1177,7 +1176,9 @@ public int hashCode(T message) {
 
   @Override
   public void mergeFrom(T message, T other) {
-    Objects.requireNonNull(other);
+    if (other == null) {
```
perezd, line 1179: should this use internal.checkNotNull?

java/core/src/main/java/com/google/protobuf/MessageSchema.java
```diff
@@ -3849,7 +3850,9 @@ private void writeFieldsInDescendingOrder(T message, Writer writer) throws IOExc
   @Override
   public void mergeFrom(T message, Reader reader, ExtensionRegistryLite extensionRegistry)
       throws IOException {
-    Objects.requireNonNull(extensionRegistry);
+    if (extensionRegistry == null) {
```
perezd, line 3853: should this use internal.checkNotNull?

perezd (approved):
Overall LGTM, since we're touching these files it might make sense to correct some invalid usages I pointed out, but also OK leaving them as is.

elharo (commented):
I think it's safer and cleaner to make this  a relatively pure rollback, aside from adding the animalsniffer rule to catch this problem in the future. The details cam be cleared up later if someone cares to.

perezd:
SGTM

java/pom.xml
elharo: I don't think there's an existing animalsniffer plugin for bazel. It should be possible to write one, though I doubt it's worth the effort.  Bazel can compile android libraries? Perhaps we could create a sample project whose compile would fail if the api level went too high.

elharo (commented):
Hmm, why did none of the continuous builds run?

java/pom.xml
perezd: Yes there is android_(library|binary) and it is actively a use case people use, if we're verifying this in Maven I'd like to see parity in Bazel given we have 1:1 parity with all other tests.

PR #8974, opened 2021-09-13
Update change log

TeBoring:


acozzette (approved):

CHANGES.txt
```diff
@@ -50,9 +50,12 @@
```
haberman, line 50: I think this should be:

> Drops support for Python <3.5.

We still support Python 3.5, though we may drop 3.5 soon also.

PR #8979, opened 2021-09-14
Protobuf: Fix for C5041 for VC++ >= 15.5

### LINKED ISSUE #8978: Generates C5041 warning as of VC++ 15.5
As of VC++ 15.5, protobuf generates build warning C5041. 15.5 made static constexpr variables implicitly inline: https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements-2017?view=msvc-160#inline-variables

I have already fixed this and will be submitting the change in a PR.

Version: master/v3.6.0/v3.5.0 etc.
Language: C++/Java/Python/C#/Ruby/PHP/Objective-C/Javascript

**What did you do?**

**What did you see instead?**

sparist:
Introduced in 15.5: https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements-2017?view=msvc-160#inline-variables
15.5 is 1912: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160

fixes  #8978

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_enum.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_enum.cc
@@ -409,7 +409,7 @@ void EnumGenerator::GenerateMethods(int idx, io::Printer* printer) {
     std::string parent = ClassName(descriptor_->containing_type(), false);
     // Before C++17, we must define the static constants which were
     // declared in the header, to give the linker a place to put them.
-    // But MSVC++ from pre-2015 to 2017 version 15.5 insists that we not.
+    // But MSVC++ pre-2015 and post-2017 (version 15.5+) insists that we not.
     format(
         "#if (__cplusplus < 201703) && "
         "(!defined(_MSC_VER) || (_MSC_VER >= 1900 && _MSC_VER < 1912))\n");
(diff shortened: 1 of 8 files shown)


Eddie-cz:
@googlebot I signed it!

src/google/protobuf/compiler/cpp/cpp_enum.cc
```diff
@@ -409,10 +409,10 @@ void EnumGenerator::GenerateMethods(int idx, io::Printer* printer) {
     std::string parent = ClassName(descriptor_->containing_type(), false);
     // Before C++17, we must define the static constants which were
     // declared in the header, to give the linker a place to put them.
-    // But pre-2015 MSVC++ insists that we not.
+    // But MSVC++ from pre-2015 to 2017 version 15.5 insists that we not.
```
acozzette, line 412: It sounds to me like the problem exists until 2015 *and* starting from 2017 version 15.5 and later, right? I guess MSVC fixed the bug in 2015 but reintroduced it in 2017. So maybe the comment should be more like: "But MSVC++ pre-2015 and post-2017 (version 15.5+) insists that we not."

acozzette:
@sparist Thanks for fixing this!

PR #8980, opened 2021-09-15
move PROTOBUF_MUST_USE_RESULT to first place in attr list

ilnurkh:
this change allow to define PROTOBUF_MUST_USE_RESULT as [[nodiscard]], without this change compilations fail. [[nodiscard]] is standertized variant of __attribute__((warn_unused_result))

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_message.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_message.cc
@@ -974,7 +974,7 @@ $annotate_extension_set$
 template <typename _proto_TypeTraits,
           ::PROTOBUF_NAMESPACE_ID::internal::FieldType _field_type,
           bool _is_packed>
-inline PROTOBUF_MUST_USE_RESULT
+PROTOBUF_MUST_USE_RESULT inline
     typename _proto_TypeTraits::Singular::MutableType
     ReleaseExtension(
         const ::PROTOBUF_NAMESPACE_ID::internal::ExtensionIdentifier<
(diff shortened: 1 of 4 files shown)


ilnurkh:
@googlebot I consent.

ilnurkh:
this pr is inspired by conversation in https://github.com/protocolbuffers/protobuf/pull/8905

ilnurkh:
about tags
1. c++
2. release notes required only if you would like to mention about ability to define this great macro: In my codebase I've already found at least 3 memory leaks (in minor targets and brahches)

fowles:
Internally, we have been implementing this as `__attribute__ ((warn_unused_result))`.  I need to do a bit of diligence to make sure all the toolchains we use except this alternate attribute ordering (yes, they should, but whether they do is another question).

georgthegreat:
@fowles, can we merge this PR?

fowles:
I am sorry, I have been distracted with other things and have not yet run the internal experiment.

fowles:
I started to explore this and realized that PROTOBUF_NODISCARD got externalized recently and is causing conflicts with this CL.  Can you integrate against latest version and see if your underlying toolchain is happy with the current configuration?

ilnurkh:
You have renamed the macro?
Looks like this PR already done 

For example: this line https://github.com/protocolbuffers/protobuf/blob/ec494a07f4e2cd9c7eac82debed289a204512769/src/google/protobuf/parse_context.h#L791
already changed due to this one

fowles:
Yeah, someone else on the team did a bunch of work here and we failed to communicate it well.  Sorry for the level of conflicts that will cause you.

ilnurkh:
no problems

acozzette:
Let me go ahead and close this since it appears that the necessary change has already been made as part of introducing `PROTOBUF_NODISCARD`.

PR #8981, opened 2021-09-15
Backport #8966 into 3.17.x

jtattermusch:
Backports https://github.com/protocolbuffers/protobuf/pull/8966.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -16,6 +16,7 @@ import platform
 # namespace_packages option for the "google" package.
 from setuptools import setup, Extension, find_packages
 
+from distutils.command.build_ext import build_ext as _build_ext
 from distutils.command.build_py import build_py as _build_py
 from distutils.command.clean import clean as _clean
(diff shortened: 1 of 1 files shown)


acozzette:
@jtattermusch Are you sure you need this fix on the 3.17.x branch? We're probably going to be doing any more 3.17.x releases.

jtattermusch:
> @jtattermusch Are you sure you need this fix on the 3.17.x branch? We're probably going to be doing any more 3.17.x releases.

I'd say let's merge it anyway 
- Even in case we're not going to make any more 3.17.x release, the PR on the branch AFTER the release tag makes it clear that this fix in not included.
- In case we decide to do a patch release for 3.17.x, this is going to be ready.

Btw, without this fix, the python aarch64 linux wheels we published for 3.17.x release are broken (their python_cpp layer won't work, only the pure python version will be usable) - just mentioning that to make the impact clear.

PR #8988, opened 2021-09-15
Add suggested field numbers on conflicts. [release notes: yes] [protoc]

karol11:
Make `protoc` output the list of suggested field numbers in the form of:
```
Suggested field numbers for cello.fs.FeatureConfig: 6, 34, 35, 36, 37
```
when it detects bad field numbers in proto files such as conflicts, use of reserved ranges, protocol specific ranges etc.

It suggests the minimal not occupied numbers available in the message.

This output can be used to simplify the creation of new fields especially when a large message has the fields grouped semantically, and these fields aren't go in ascent number order. 
This output can be grabbed by IDE or automated tools to resolve conflicts automatically,
This CL eliminated the necessity of `// next ID: 42` comments in messages.

### DIFF
--- a/src/google/protobuf/descriptor.cc
+++ b/src/google/protobuf/descriptor.cc
@@ -41,6 +41,7 @@
 #include <map>
 #include <memory>
 #include <set>
+#include <sstream>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


karol11:
Unfortunately I have no way to add labels. [release notes: yes] [protoc]

Issue #8984, opened 2021-09-15
3.18.0 isn't compatible with Python 2 despite being installable


convoliution:
**What version of protobuf and what language are you using?**
Version: v3.18.0
Language: Python

**What operating system (Linux, Windows, ...) and version?**
macOS Mojave 10.14.6

**What runtime / compiler are you using (e.g., python version or gcc version)**
```bash
Python 2.7.18 (default, Mar 30 2021, 14:20:09) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)] on darwin
```

**What did you do?**
1. Activate a Python 2 environment
2. `pip install protobuf==3.18.0`
3. `python -c "from google.protobuf import descriptor"`

**What did you expect to see**
Either:
- `ERROR: Could not find a version that satisfies the requirement protobuf==3.18.0`
  in (2)
- no errors in (3)

**What did you see instead?**
```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/convoliution/.pyenv/versions/dev2/lib/python2.7/site-packages/google/protobuf/descriptor.py", line 113
    class DescriptorBase(metaclass=DescriptorMetaclass):
                                  ^
SyntaxError: invalid syntax
```
in (3)

srenauld:
This breaks [nrfutil](https://www.nordicsemi.com/Products/Development-tools/nRF-Command-Line-Tools/Download#infotabs), rendering a major provider's CLI tool for DFU upgrades inoperable with no way to pin this dependency.

odelrio:
https://github.com/googleapis/python-api-core depends on protobuf and this version breaks Python2.7-based builds. My workaround was to force v3.17.3 in `requirements.txt`.

busunkim96:
@odelrio I just released a 1.x version of google-api-core that pins to protobuf <3.18.0. https://pypi.org/project/google-api-core/1.31.3/

alanyee:
https://github.com/protocolbuffers/protobuf/issues/9045

convoliution:
> Ok, I went ahead and yanked 3.18.0, so hopefully that should fix the problem. Users who still need Python 2 support can keep using 3.17.3.

—https://github.com/protocolbuffers/protobuf/issues/9045#issuecomment-943640239

Perfect!

PR #8989, opened 2021-09-16
Add python-requires in setup.py

aisk:
I saw protobuf have some commits which stop testing and building wheels for python2:

https://github.com/protocolbuffers/protobuf/commit/2fbc07b2439c2518a4985972c9e78d07b44603c2
https://github.com/protocolbuffers/protobuf/pull/8891

And the codes does not support python2 now: https://github.com/protocolbuffers/protobuf/issues/8984

So I think I can assume protobuf does not support python2 for now. I think we should add a `python_requires` field in setup.py, pip and poetry will check this field and fallback to older version if it in python2 environment.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -308,5 +308,5 @@ if __name__ == '__main__':
       },
       install_requires=install_requires,
       ext_modules=ext_module_list,
-      python_requires='>=3.3',
+      python_requires='>=3.5',
   )
--- a/python/setup.py
+++ b/python/setup.py
@@ -308,4 +308,5 @@ if __name__ == '__main__':
       },
       install_requires=install_requires,
       ext_modules=ext_module_list,
+      python_requires='>=3.3',
   )


Lothiraldan:
According to https://packaging.python.org/guides/dropping-older-python-versions/#dropping-a-python-release and adding such field on other project, I think we need a 3.17.4 release with the python_requires that includes Python 2 and then a release 3.18.1 that drops Python 2 support.

aisk:
@Lothiraldan As the document you linked above, and the [semver](https://semver.org/) said, I think publish a major version (4.0.0) is the right way.

busunkim96:
@acozzette @teboring @haberman Please take a look.

Lothiraldan:
@aisk Sorry to ask, but where is the 4.0.0 version? I don't see it on Pypi nor see a tag or branch for 4.0.0.

busunkim96:
From past releases of protobuf ,it looks like the version numbers need to match the version of the runtime and protoc. 

I think what @Lothiraldan suggested should work well:
> According to https://packaging.python.org/guides/dropping-older-python-versions/#dropping-a-python-release and adding such field on other project, I think we need a 3.17.4 release with the python_requires that includes Python 2 and then a release 3.18.1 that drops Python 2 support.

aisk:
This change will break old codes so it's better to have a major release. But if protobuf have it's own versioning strategy I think it's ok, the `python_requires` field should solve most problem (except people who have old version pip / setuptools).

Ark-kun:
> I think we need a 3.17.4 release with the python_requires that includes Python 2 and then a release 3.18.1 that drops Python 2 support.

They should probably just release 3.18.1 and **yank** or delete 3.18.0 from PyPI

acozzette:
Thanks, @aisk. I am getting ready to do a 3.18.1 release that will include this fix.

PR #8991, opened 2021-09-16
Include python/google/__init__.py in python sources.

mbabinski-at-google:
This file is critical for resolving imports. While python scripts run
from within bazel will run correctly without this file being declared,
packaging python sources with transitive deps currently omits this,
leading to broken packages.

Note that __init__.py is the only file in python/google other than the "protobuf" directory which is already globbed over.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -834,7 +834,7 @@ py_library(
     name = "python_srcs",
     srcs = glob(
         [
-            "python/google/protobuf/**/*.py",
+            "python/google/**/*.py",
         ],
         exclude = [
             "python/google/protobuf/internal/*_test.py",


acozzette:
It looks like this file is actually being deliberately omitted starting from 02955622cce1e4cbd8337a76269e2b81f39bee09. @dlj-NaN might be able to explain more about the rationale for this.

dlj-NaN:
Adam is correct... including a `google/__init__.py` file will generally interfere with other packages under the `google` namespace. Skipping that file ensures that the Python import machinery understands that `google` is a namespace package, as opposed to our package.

Issue #8484, opened 2021-04-13
cannot compile c++ generated code


Eddie-cz:
Version: 3.15.8
Language: C++
Windows 10, MSVC 16.9.3 with clang-cl compiler (LLVM 11.1)

Message definition:
syntax = "proto3";
message StringMsg 
{    		
	string value = 1;
}

Compilation of generated pb.cc fails for this message with "error : variable does not have a constant initializer".
Compiles with repeated modifier.

Variable is:
PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT StringMsgDefaultTypeInternal _StringMsg_default_instance_;

TeBoring:
Do you see if there is a constexpr constructor for StringMsgDefaultTypeInternal in .pb.cc?

TeBoring:
Could you also paste the full error log?

Eddie-cz:
There is constexpr constructor, code compiles fine with msvc compiler, only clang-cl thinks this message variable doesn't satisfy constinit restrictions introduced by PROTOBUF_CONSTINIT (guess it has problems with string type).
Should I put repeated modifier for that string param, or manualy remove PROTOBUF_CONSTINIT from pb.cc, it compiles fine.
Log:
etherCATMessages.pb.cc(114,79): error : variable does not have a constant initializer
etherCATMessages.pb.cc(114,31): message : required by 'require_constant_initialization' attribute here
protobuf\3.15.8\include\google/protobuf/port_def.inc(571,30): message : expanded from macro 'PROTOBUF_CONSTINIT'

TeBoring:
https://godbolt.org/z/75rfhqY4E
I cannot reproduce it in the godbolt. Could you help advice a reproduce?

TeBoring:
https://godbolt.org/z/j9nv47n5P
Still no error

Eddie-cz:
Issue is isolated to 
PROTOBUF_CONSTINIT google::protobuf::internal::ArenaStringPtr sptr(&google::protobuf::internal::fixed_address_empty_string);
Same error as described above. To reproduce, you must define your ExplicitlyConstructed<std::string> empty_string variable in different compilation unit, in this case in different dll, same as it's for protobuf variables. When I test your code in one compilation unit, it's ok, problem is with protobuf dll usage.

ngg:
We ran into the same problem.
[...]

PR #8993, opened 2021-09-17
Fix for issue #8484, constant initialization doesn't compile in msvc clang-cl environment

Eddie-cz:


Eddie-cz:
@googlebot I signed it!

acozzette:
@Eddie-cz Thank you for the bug fix.

PR #8994, opened 2021-09-17
Fix cmake install for protoc component

johanmoe:
Fixes:
  * RUNTIME DESTINATION for protoc TARGET is not given a COMPONENT option


johanmoe:
@googlebot I signed it!

johanmoe:
@googlebot I fixed it.

acozzette:
Thanks, @johanmoe.

PR #8997, opened 2021-09-18
Adopt pure ruby DSL implementation for JRuby

JasonLunn:
Also squashes all current JRuby unit test failures and resets expectations about known conformance test failures to be the super set of Java + Ruby known failures.

Includes all changes separately posted under #8954


JasonLunn:
Closing in favor of #9047

PR #9002, opened 2021-09-20
docs: 3.18.0 is released

elharo:
@acozzette


acozzette (approved):

acozzette:
If we want to keep these version numbers up to date then you might want to just update them in `update_version.py`, which updates all the other versions.

PR #9003, opened 2021-09-20
Update ruby_generator.cc to allow proto2 imports in proto3

zhangskz:


haberman:
This looks great. Could you also add a test in one of these protos, to ensure that the generated code can be successfully loaded into a Ruby process and used?

https://github.com/protocolbuffers/protobuf/tree/master/ruby/tests

I think this is a good candidate; you could make this file import one of the proto2 protos from this directory: https://github.com/protocolbuffers/protobuf/blob/master/ruby/tests/basic_test.proto

Then add a test in https://github.com/protocolbuffers/protobuf/blob/master/ruby/tests/basic.rb

haberman (approved):

PR #9005, opened 2021-09-21
deps: add j2objc annotations

elharo:
@perezd this will be needed on the next sync with google3. Until then it's harmless.


perezd (approved):
LGTM!

PR #9010, opened 2021-09-22
Creating bazel toolchain to crosscompile with

deannagarcia:
Creating a bazel toolchain to cross compile to all architectures supported with Linux.

### DIFF
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -1,0 +1,90 @@
+load(":cc_toolchain_config.bzl", "cc_toolchain_config")
+package(default_visibility = ["//visibility:public"])
+
+filegroup(name = "empty")
+
+
+cc_toolchain_suite(
+    name = "clang_suite",
+    toolchains = {
+        "aarch64": ":cc-compiler-aarch64",
+        "ppcle": "cc-compiler-ppcle",
+        "s390x": ":cc-compiler-s390x",
+        "x86_32": ":cc-compiler-x86_32",
+        "x86_64": ":cc-compiler-x86_64",
+    },
+)
+
+CC_TOOLCHAIN_CPUS = [
+    "aarch64",
+    "ppcle",
+    "s390x",
+    "x86_32",
+    "x86_64",
+]
+
+[
+    cc_toolchain(
+        name = "cc-compiler-" + cpu,
+        all_files = ":empty",
+        compiler_files = ":empty",
+        dwp_files = ":empty",
+        dynamic_runtime_lib = ":empty",
+        linker_files = ":empty",
+        objcopy_files = ":empty",
+        output_licenses = ["restricted"],
+        static_runtime_lib = ":empty",
+        strip_files = ":empty",
+        toolchain_config = ":" + cpu + "-config",
+        toolchain_identifier = "linux_" + cpu,
+    )
+    for cpu in CC_TOOLCHAIN_CPUS
+]
+
+cc_toolchain_config(
+    name = "aarch64-config",
+    bit_flag = "-m64",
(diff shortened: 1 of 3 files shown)


toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,151 @@
...
+
+  return cc_common.create_cc_toolchain_config_info(
+      abi_libc_version = ctx.attr.target_cpu,
+      abi_version = ctx.attr.target_cpu,
+      ctx = ctx,
+      compiler = "clang",
+      cxx_builtin_include_directories = [
+          ctx.attr.toolchain_dir,
+          "/usr/include",
+          "/usr/lib/clang/11.0.1/include",
```
perezd, line 129: Is there potentially a symlink somewhere that lets us avoid using explicit versions?

toolchain/BUILD
```diff
@@ -0,0 +1,90 @@
...
+    toolchains = {
+        "aarch64": ":cc-compiler-aarch64",
+        "ppcle": "cc-compiler-ppcle",
+        "s390x": ":cc-compiler-s390x",
+        "x86_32": ":cc-compiler-x86_32",
+        "x86_64": ":cc-compiler-x86_64",
+    },
+)
+
+CC_TOOLCHAIN_CPUS = [
```
perezd, line 18: Would it be possible to condense the toolchains dictionary on L9 and this into a single variable that can be reused in the two places? That way you'd be able to use cpu + name from the dictionary directly in the for loop of cc_toolchain, and you'd have less duplication here.

toolchain/BUILD
```diff
@@ -0,0 +1,90 @@
...
+        output_licenses = ["restricted"],
+        static_runtime_lib = ":empty",
+        strip_files = ":empty",
+        toolchain_config = ":" + cpu + "-config",
+        toolchain_identifier = "linux_" + cpu,
+    )
+    for cpu in CC_TOOLCHAIN_CPUS
+]
+
+cc_toolchain_config(
```
perezd, line 44: It'd be good to prefix all the name fields with the OS here as well so they don't collide potentially with other OSes.

perezd (changes requested):
Looking good! Nice work.

toolchain/cc_toolchain_config.bzl
deannagarcia: I removed the version and it still works.

toolchain/BUILD
deannagarcia: Done!

toolchain/BUILD
deannagarcia: Done!

toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,151 @@
...
+                              "--target=" + ctx.attr.target_full_name,
+                          ],
+                      ),
+                  ]),
+              ),
+          ],
+      ),
+  ]
+
+  features.append(feature(
```
perezd, line 98: any reason to not just inline this as part of the array literal? vs calling append?

toolchain/BUILD
```diff
@@ -0,0 +1,83 @@
+load(":cc_toolchain_config.bzl", "cc_toolchain_config")
+package(default_visibility = ["//visibility:public"])
```
perezd, line 2: ultra nit: add space between load and package.

perezd (approved):
LGTM!

toolchain/cc_toolchain_config.bzl
deannagarcia: Nope, added it to part of the array literal.

toolchain/BUILD
deannagarcia: Done!

PR #9013, opened 2021-09-23
Add two new generation options around objc prefix usage.

thomasvl:
- `require_prefixes`: Fail generation if a file doesn't have the option.
- `prefixes_must_be_registered`: If a prefix is in a file, it must be registered
  in the expected file. There is a grey space here for files without packages as
  there is no way to register it, so these case will still warn at the moment.

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
@@ -101,6 +101,41 @@ bool ObjectiveCGenerator::GenerateAll(
         generation_options.expected_prefixes_suppressions.push_back(
             std::string(split_piece));
       }
+    } else if (options[i].first == "prefixes_must_be_registered") {
+      // If objc prefix file option value must be registered to be used. This
+      // option has no meaning if an "expected_prefixes_path" isn't set. The
+      // available options are:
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/objectivec/objectivec_generator.cc
```diff
@@ -101,6 +101,39 @@ bool ObjectiveCGenerator::GenerateAll(
         generation_options.expected_prefixes_suppressions.push_back(
             std::string(split_piece));
       }
+    } else if (options[i].first == "prefixes_must_be_registered") {
+      // If objc prefix file option value must be registered to be used. This
+      // option has no meaning if an "expected_prefixes_path" isn't set. The
+      // available options are:
+      //   "no": They don't have to be registered.
```
dmaclach, line 108: May just want to mention in both cases what the default state is just to be clear.

dmaclach (approved):

thomasvl:
The kokoro failures seem to be things broken on trunk and not related to this change.

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
thomasvl: Done.

PR #9018, opened 2021-09-24
Add type definitions to php

mattgrande:
Much of the generated PHP code is missing return types, and the type definitions for the expected parameters. This PR adds those types to provide a more pleasant and secure experience for PHP developers.

    // Before
    public function getField()
    {
        return $this->field;
    }

    public function hasField()
    {
        return isset($this->field);
    }

    public function clearField()
    {
        unset($this->field);
    }

    public function setField($var)
    {
        GPBUtil::checkMessage($var, \Google\Protobuf\BoolValue::class);
        $this->field = $var;

        return $this;
    }

    // After
    public function getField(): ?\Google\Protobuf\BoolValue
    {
        return $this->field;
    }

    public function hasField(): bool
    {
        return isset($this->field);
    }

    public function clearField(): void
    {
        unset($this->field);
    }

    public function getFieldUnwrapped(): ?bool
    {
        return $this->readWrapperValue("field");
    }

    public function setField(\Google\Protobuf\BoolValue $var): TestBoolValue
    {
        GPBUtil::checkMessage($var, \Google\Protobuf\BoolValue::class);
        $this->field = $var;

        return $this;
    }

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,6 +147,7 @@ php/tests/multirequest.result
 php/tests/nohup.out
 php/tests/.phpunit.result.cache
 php/tests/phpunit-*
+php/tmp
 php/ext/google/protobuf/.libs/
 php/ext/google/protobuf/Makefile.fragments
 php/ext/google/protobuf/Makefile.global
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -969,7 +969,7 @@ void GenerateAddFileToPool(const FileDescriptor* file, const Options& options,
                            io::Printer* printer) {
   printer->Print(
       "public static $is_initialized = false;\n\n"
-      "public static function initOnce() {\n");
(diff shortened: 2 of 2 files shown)


mattgrande:
@googlebot I signed it!

haberman:
I am sorry for the very slow review on this.

It would be great to have more type annotations. Our main constraint here is that we need to be compatible with PHP all the way back to 7.0.

I notice that the tests are failing. I see both syntax errors and type checking errors. The type checking errors in the tests worry me more, because they suggest that this change might be backward-incompatible for existing code. If that's true, we might need to temporarily offer an option to suppress the checks.

Would you be able to:

1. Fix the syntax errors in the tests.
2. Confirm whether these changes could break existing (working) code. If so we should talk about potentially having a switch for this.

acozzette:
@mattgrande I'm going to close this since it's been a few weeks with no action, but feel free to comment here if you would like to reopen this.

PR #9019, opened 2021-09-24
Update some types to be more current.

thomasvl:


fowles:
this was reviewed internally before merging externally.

PR #9021, opened 2021-09-24
Cherry-pick fixes for 3.18.1

acozzette:
This PR cherry-picks a couple fixes for the 3.18.1 release. The main one is from #8989 and ensures that we explicitly require Python 3 in setup.py. I also made some other small tweaks to get all the tests passing.


acozzette:
The CLA is OK because the cherry-picked commits already went through the CLA check on the master branch.

haberman (approved):

PR #9025, opened 2021-09-26
Update json_format.py

windprog:
proto3 case:
if enum provides value "string"
will raise a "number" NameError exception


windprog:
@googlebot I signed it!

elharo (commented):
Is it possible to add a test for this case?

acozzette:
I'm going to close this since there has been no activity in a while, but feel free to reopen this if you would like to keep working on it.

PR #9028, opened 2021-09-28
Changing clang path in toolchain

deannagarcia:
Change clang path in toolchain to match where it is in the docker container.


perezd (approved):
LGTM!

PR #9033, opened 2021-09-29
Specify bits of architectures in toolchain names

deannagarcia:
Changing toolchain names to exactly match the zip files we want to make so that the platform variable can be easily reused for zip file name and bazel toolchain name


perezd (approved):
LGTM!

PR #9043, opened 2021-10-01
Avoid installing anything from Homebrew in test runs

acozzette:
Installing and upgrading Homebrew packages is taking quite a lot of time
(around 20-30 minutes) for each run. This commit removes all Homebrew
usage from the test runs. Homebrew may have been necessary at some point
in the past, but now it appears that everything works without it. The
preinstalled build tools seem to be sufficient for building protoc, and
Ruby is something we already get from rvm.


thomasvl (approved):
LG for the cocoapods tests.

acozzette:
I believe the Ruby 3.0 failure is an issue with RVM and isn't related to this change, so it appears that everything does still work without us needing Homebrew.

PR #9044, opened 2021-10-01
Merge 3.18.x into master

acozzette:
This should fix the Ruby test failures and speed up all the Mac OS tests.


acozzette:
The CLA is OK because these commits have all already gone through the CLA check.

haberman (approved):

PR #9052, opened 2021-10-04
Change include messages to match docker

deannagarcia:
The c++ version in docker is 8, changing this to match the docker packages. Hopefully in the future we can figure out a way to not hardcode this version, but first want to get it working.


perezd (approved):
LGTM!

PR #9053, opened 2021-10-05
Speed up Java UTF-8 validation tests

acozzette:
These are by far our slowest tests, so this commit makes some changes to
speed them up. The reason they're slow is that they exhaustively iterate
over all 2**24 possible 3-byte sequences.
- In DecodeUtf8Test, I changed the test to just count the number of
  valid sequences instead of asserting on the outcome of each one. This
  way, the number of loop iterations is the same but each iteration is
  much faster. If we introduce a bug, it should still be pretty likely
  to make the test fail.
- In IsValidUtf8Test, I kept the same basic test structure but updated
  the test to exercise just one ByteStringFactory instead of all three.
  DIRECT_NIO_FACTORY sounded like the most subtle one, so I chose that
  one to test.

This makes "bazel test --nocache_test_results //java/core:tests" take 2
minutes instead of 6. The time for "mvn test" is down to about 7
minutes. I didn't time it at the baseline but I think it was upwards of
10 minutes.

I suspect we could make the Maven test run comparable to Bazel if we
could configure Maven to parallelize its test cases, but I'm not sure
how to do this.

### DIFF
--- a/java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
+++ b/java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
@@ -55,29 +55,24 @@ public class DecodeUtf8Test extends TestCase {
   }
 
   public void testThreeBytes() throws Exception {
-    // Travis' OOM killer doesn't like this test
-    if (System.getenv("TRAVIS") == null) {
-      int count = 0;
-      int valid = 0;
-      for (int i = Byte.MIN_VALUE; i <= Byte.MAX_VALUE; i++) {
-        for (int j = Byte.MIN_VALUE; j <= Byte.MAX_VALUE; j++) {
-          for (int k = Byte.MIN_VALUE; k <= Byte.MAX_VALUE; k++) {
-            byte[] bytes = new byte[]{(byte) i, (byte) j, (byte) k};
-            ByteString bs = ByteString.copyFrom(bytes);
-            if (!bs.isValidUtf8()) {
-              assertInvalid(bytes);
-            } else {
-              valid++;
-            }
-            count++;
-            if (count % 1000000L == 0) {
-              logger.info("Processed " + (count / 1000000L) + " million characters");
-            }
+    int count = 0;
+    int valid = 0;
+    for (int i = Byte.MIN_VALUE; i <= Byte.MAX_VALUE; i++) {
+      for (int j = Byte.MIN_VALUE; j <= Byte.MAX_VALUE; j++) {
+        for (int k = Byte.MIN_VALUE; k <= Byte.MAX_VALUE; k++) {
+          byte[] bytes = new byte[]{(byte) i, (byte) j, (byte) k};
+          ByteString bs = ByteString.copyFrom(bytes);
+          if (bs.isValidUtf8()) {
+            valid++;
+          }
+          count++;
+          if (count % 1000000L == 0) {
+            logger.info("Processed " + (count / 1000000L) + " million characters");
           }
         }
       }
-      assertEquals(IsValidUtf8TestUtil.EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT, valid);
     }
+    assertEquals(IsValidUtf8TestUtil.EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT, valid);
   }
 
   /**
--- a/java/core/src/test/java/com/google/protobuf/IsValidUtf8Test.java
+++ b/java/core/src/test/java/com/google/protobuf/IsValidUtf8Test.java
@@ -74,12 +74,8 @@ public class IsValidUtf8Test {
   /** Tests that round tripping of all three byte permutations work. */
   @Test
   public void testIsValidUtf8_3Bytes() {
-    // Travis' OOM killer doesn't like this test
-    if (System.getenv("TRAVIS") == null) {
-      testBytes(LITERAL_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
-      testBytes(HEAP_NIO_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
-      testBytes(DIRECT_NIO_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
-    }
+    // This test is slow, so we test just one factory instead of all three.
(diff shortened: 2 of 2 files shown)


java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
```diff
@@ -55,29 +55,24 @@ public void testTwoBytes() throws Exception {
...
+          }
+          count++;
+          if (count % 1000000L == 0) {
+            logger.info("Processed " + (count / 1000000L) + " million characters");
           }
         }
       }
-      assertEquals(IsValidUtf8TestUtil.EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT, valid);
     }
+    assertEquals(IsValidUtf8TestUtil.EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT, valid);
```
elharo, line 75: Is this basically a characterization test?

java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
```diff
@@ -55,29 +55,24 @@ public void testTwoBytes() throws Exception {
...
-              logger.info("Processed " + (count / 1000000L) + " million characters");
-            }
+    int count = 0;
+    int valid = 0;
+    for (int i = Byte.MIN_VALUE; i <= Byte.MAX_VALUE; i++) {
+      for (int j = Byte.MIN_VALUE; j <= Byte.MAX_VALUE; j++) {
+        for (int k = Byte.MIN_VALUE; k <= Byte.MAX_VALUE; k++) {
+          byte[] bytes = new byte[]{(byte) i, (byte) j, (byte) k};
+          ByteString bs = ByteString.copyFrom(bytes);
+          if (bs.isValidUtf8()) {
```
elharo, line 65: I think it's worth adding some sample three byte tests along the lines of testInvalid_4BytesSamples

elharo (commented):
Why does this test not seem to exist in google3?

elharo:
Coincidentally I just saw this while running tests in another PR:

```
[INFO] Running com.google.protobuf.IsValidUtf8Test
Oct 05, 2021 6:39:22 AM com.google.protobuf.IsValidUtf8TestUtil testBytes
INFO: Round tripped 18304 of 65536
Oct 05, 2021 6:39:23 AM com.google.protobuf.IsValidUtf8TestUtil testBytes
INFO: Round tripped 18304 of 65536
Oct 05, 2021 6:39:23 AM com.google.protobuf.IsValidUtf8TestUtil testBytes
INFO: Round tripped 18304 of 65536
```

Why is it logging the same value repeatedly?

java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
```diff
@@ -55,29 +55,24 @@ public void testTwoBytes() throws Exception {
...
+            valid++;
+          }
+          count++;
+          if (count % 1000000L == 0) {
```
elharo, line 69: Random idea: do we also want to get rid of 16 million modulus and compare operations here? I don't know. What does profiling suggest? These log messages don't seem hugely useful except in the rare case when someone is watching the test output in real time. 

Also, why is the int count being modded with a long?

java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
```diff
@@ -55,29 +55,24 @@ public void testTwoBytes() throws Exception {
...
-            count++;
-            if (count % 1000000L == 0) {
-              logger.info("Processed " + (count / 1000000L) + " million characters");
-            }
+    int count = 0;
+    int valid = 0;
+    for (int i = Byte.MIN_VALUE; i <= Byte.MAX_VALUE; i++) {
+      for (int j = Byte.MIN_VALUE; j <= Byte.MAX_VALUE; j++) {
+        for (int k = Byte.MIN_VALUE; k <= Byte.MAX_VALUE; k++) {
+          byte[] bytes = new byte[]{(byte) i, (byte) j, (byte) k};
```
elharo, line 63: No idea if this is measurable, but instead of a new array each time you could create the array before the loop. Then set i in the outer loop, j in the middle loop, and k in the inner loop.

java/core/src/test/java/com/google/protobuf/IsValidUtf8Test.java
```diff
@@ -74,12 +74,8 @@ public void testIsValidUtf8_2Bytes() {
...
   @Test
   public void testIsValidUtf8_3Bytes() {
-    // Travis' OOM killer doesn't like this test
-    if (System.getenv("TRAVIS") == null) {
-      testBytes(LITERAL_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
-      testBytes(HEAP_NIO_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
-      testBytes(DIRECT_NIO_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
-    }
+    // This test is slow, so we test just one factory instead of all three.
+    testBytes(DIRECT_NIO_FACTORY, 3, EXPECTED_THREE_BYTE_ROUNDTRIPPABLE_COUNT);
```
elharo, line 78: This one feels unnecessarily risky. Instead of removing two tests how about splitting them into three separate test methods that could execute in parallel?

java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
acozzette: Yes, I think that is fair to say. In this case I think it is appropriate since as I understand, this aspect of the UTF-8 encoding is pretty much set in stone and not subject to change.

acozzette:
> Why does this test not seem to exist in google3?

Internally we rely on a separate utility for validating UTF-8.

java/core/src/test/java/com/google/protobuf/DecodeUtf8Test.java
acozzette: Good point, this is probably OK to remove. I think it was useful for tracking the progress since the test was pretty slow, but now it should be fast enough to make this unnecessary. I'm not sure why the int was being modded with a long.

acozzette:
> > Why does this test not seem to exist in google3?
> 
> Internally we rely on a separate utility for validating UTF-8.

Actually I think I was wrong about this. I think it might just be an oversight that we don't have an internal version of this test.

PR #9060, opened 2021-10-05
Changing toolchain names

deannagarcia:
Change toolchain/config names to all be consistent with the names of the artifacts we want to create, also changing the imports of the x86_32 build.


perezd (approved):
LGTM!

Issue #8373, opened 2021-03-04
3.15.x doesn't build on FreeBSD cause of missing endian.h


dmilith:
**What version of protobuf and what language are you using?**
Version: 3.15.4
Language: C++

**What operating system (Linux, Windows, ...) and version?**
FreeBSD x86_64, v12.2

**What runtime / compiler are you using (e.g., python version or gcc version)**
clang 10.0.1

**What did you do?**
Either autotools build with ./configure && make or cd cmake && cmake ../ … && make

**What did you expect to see**
Properly built library

**What did you see instead?**

A lot of errors like this in multiple files, example:

```
src/google/protobuf/stubs/port.h:63:10: fatal error: 'endian.h' file not found
#include <endian.h>  // __BYTE_ORDER
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**
No. It's a pretty standard environment.

dmilith:
This can be hacked by:

```
sed -i '' -e 's|endian.h|machine/endian.h|g' src/google/protobuf/*.cc src/google/protobuf/*.h src/google/protobuf/stubs/*.h src/google/protobuf/io/*.h
```

So it's just wrong include path on FreeBSD. After that hack project builds w/o issues.

Mizux:
Strange FreeBSD seems already supported...
https://github.com/protocolbuffers/protobuf/blob/87dd07b4367b7676af42105d0d102f4e536c248b/src/google/protobuf/stubs/port.h#L60-L66

FYI in google/or-tools we are using (and we test using github macos runner (ed provide virtualbox support), and a Vagrant FreeBSD box)
```cpp
#if (defined(__APPLE__) || defined(__FreeBSD__)) && defined(__GNUC__)
#include <machine/endian.h>
#elif !defined(_MSC_VER)
#include <endian.h>
#endif
```

scddev:
Same behavior also on AIX and SunOS:

```
-bash-4.4$ cmake -S cmake -B _build -Dprotobuf_BUILD_TESTS=OFF
-- 
-- 3.17.0.0
-- Configuring done
-- Generating done
-- Build files have been written to: /home/foo/protobuf-3.17.0/_build
-bash-4.4$ cmake --build _build
gmake[1]: Entering directory '/home/foo/protobuf-3.17.0/_build'
gmake[2]: Entering directory '/home/foo/protobuf-3.17.0/_build'
[...]

PR #9065, opened 2021-10-06
Fix build on AIX and SunOS (#8373)

scddev:
* fix includes for AIX and SunOS


scddev:
@googlebot I signed it!

elharo:
Are people still using pre-Solaris SunOS? Or is this for Solaris?

I wonder if we could set up a Github action on one or both of these platforms to test this?

scddev:
@googlebot I signed it!

gdharley:
Thanks Dietmar, I thought I had tried static linking, will try again.
I guess it is all these tweaks that need to be fixed to properly support gcc on AIX.

PR #9075, opened 2021-10-06
Express JRuby's dependency on Java through a Maven submodule

JasonLunn:
An alternative approach to #9071 that makes JRuby a submodule of the protobuf-parent artifact.


JasonLunn:
Rerunning tests - seems like the failed one is due to a sync issue from the mirror

PR #9094, opened 2021-10-13
supporting nanos and second in timestamp merge

imtripathi:
This Commit adds in the ablity of mergeTimestamp method to parse the  timestamp formate of type "{\"seconds\":1800,\"nanos\":0}" as if now only this format was supported "1970-01-01T00:00:00Z".


imtripathi:
> @googlebot I signed it!

PR #9100, opened 2021-10-14
Remove references to HAVE_PTHREAD

acozzette:
This is based on @haberman's changes in #8257. Now that we're using
std::mutex we no longer need to check whether pthreads are available, so
this commit removes references to HAVE_PTHREAD. I left the autotools
build alone, though, since we are likely to drop support for it soon
anyway.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -26,7 +26,6 @@ ZLIB_DEPS = ["@zlib//:zlib"]
 ################################################################################
 
 MSVC_COPTS = [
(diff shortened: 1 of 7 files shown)


redbaron:
@acozzette , cmake still has 

```
find_package(Threads REQUIRED)
```

is it intentional?

acozzette:
I think that line is probably OK to remove.

PR #9108, opened 2021-10-15
Add failing Java test

acozzette:
This is just to double check that we have good enough Java CI coverage after turning off our Java 7 tests.


elharo:
looks like install does fail:

[INFO]
[ERROR] Failures:
[ERROR]   AnyTest.testCustomTypeUrls:84 expected: 5
but was : 2

Issue #6035, opened 2019-04-13
How to set JSON options to emit fields with default values in proto3


ganl:
**PHP**

https://developers.google.cn/protocol-buffers/docs/proto3#json

> JSON options
> A proto3 JSON implementation may provide the following options:
> 
> Emit fields with default values: Fields with default values are omitted by default in proto3 JSON output. An implementation may provide an option to override this behavior and output fields with their default values.

how to set ？like：`EmitDefaults` or `defaults_json`

```
\Google\Protobuf\Internal\Message::existField 
……
            return $values !== $this->defaultValue($field);
```
Fields with default values are always omitted

ganl:
@TeBoring

TeBoring:
We haven't implemented that option yet. Would you like to contribute a PR? Thanks!

patramsey:
@TeBoring could you point me in the right direction where to hook this in? I'm assuming 
`Google/Protobuf/Internal/Message.php` ?

This would be a great addition in my implementation and I'm happy to help.

csuntrane:
Any idea when this might be implemented. It seems proto2 doesn't have this issue.

panol:
plus one

deannagarcia:
While this is a reasonable request, the protobuf team will not be prioritizing this in the near future. If you send us a PR to add it, we would love to help get it merged!

maciek-szn:
Well, this is great. Google's GTFS-RT specification [requires field `incrementality` in the output file](https://gtfs.org/realtime/reference/#message-feedheader), with its default value (FULL_DATASET). Google's Protobuf compiler cuts off the default values 🤷‍♂️

SpencerMalone:
@deannagarcia I opened a PR for y'all when you have time!

PR #9124, opened 2021-10-20
[PHP] Add includeDefaultValueFields to serializeToJsonString

yulia-0990:
Fix for #6035


haberman:
Sorry for the slow review.

Can we use keyword arguments instead of positional arguments? eg. `serializeToJsonString(["foo" => true])`

Let's keep the `preserve_proto_fieldnames` option. It seems possible that people were using it even if it was unregistered. Even if it was not already being used, it should be available via keyword argument.

Let's call the new option `emit_defaults` to match Ruby: https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/message.c#L1038-L1077

yulia-0990:
Thank you for your review!

I hope I understood what you meant correctly. I made JSON options an array, also in C, and kept the `preserve_proto_fieldnames` option there (although it will not be compatible with its previous undocumented C implementation, as before it was a standalone boolean argument, not an array).

Unfortunately I could not use the new `Google\Protobuf\JsonSerializeOptions::EMIT_DEFAULTS` const in tests - it works for php, but for C it gives me `When using the C extension, we should not load runtime class` and I couldn't find how to fix it.

haberman:
Sorry for the slow review.

It looks like you did not implement `preserve_proto_fieldnames` in pure PHP. Would you be able to do that (and add tests), so the two implementations respect the same options?

upb has also changed its API names, so the change will require a bit of renaming.

php/ext/google/protobuf/message.c
```diff
@@ -764,6 +764,11 @@ PHP_METHOD(Message, mergeFromJsonString) {
   }
 }
 
+// JSON options /////////////////////////////////////////////////////////////////////
+
```
haberman, line 768: Please mention that this must match JsonSerializeOptions.php.

php/src/Google/Protobuf/JsonSerializeOptions.php
```diff
@@ -0,0 +1,38 @@
...
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+namespace Google\Protobuf;
+
+class JsonSerializeOptions
+{
+    const EMIT_DEFAULTS = "emit_defaults";
```
haberman, line 37: Add preserve_proto_fieldnames? Also please add a comment that this must match the C file.

acozzette:
@yulia-0990 I am going to close this since there has been no activity in 30 days, but if you are still interested in working on this change then feel free to comment here and I can reopen it.

s2x:
Hey @acozzette. Can I finish this task instead of @yulia-0990. I currently have a version build with PRESERVE_PROTO_FIELD_NAMES support. Additionally, I have written tests for both EMIT_DEFAULTS and PRESERVE_PROTO_FIELD_NAMES parameters.

My changes were created based on the current main branch. I just have a question, do I have to create a new pull request?

acozzette:
Great! Yes, please create a new pull request.

PR #9127, opened 2021-10-20
Make IsDescriptor* helpers lenient about descriptor name

acozzette:
Our internal version of the codebase has a different path and package
name for descriptor.proto, so this change updates IsDescriptorProto()
and IsDescriptorOptionMessage() to be able to handle both the internal
and external descriptor types.


acozzette:
> This could use a test or two.

Good idea, done.

PR #9136, opened 2021-10-21
Merge 3.19.x into master

acozzette:


acozzette:
The CLA is OK because these commits have already gone through the CLA check on 3.19.x.

PR #9139, opened 2021-10-22
Implement unambiguous comparisons for RepeatedPtrIterator in C++20

mumbleskates:
In C++20, code that compares `RepeatedPtrIterator<Foo>` and `RepeatedPtrIterator<const Foo>` can be ambiguous because the reversed operator is equally viable (`[-Wambiguous-reversed-operator]`). If we define viable operators for both combinations of constness there is always a best match when comparing const and non-const iterators.

Without this either warnings will be omitted or compilation failures will occur when using this header in C++20. Notably this causes the `//:protobuf_test` target to no longer fail under `--cxxopt='-std=c++20'`.

### DIFF
--- a/src/google/protobuf/repeated_ptr_field.h
+++ b/src/google/protobuf/repeated_ptr_field.h
@@ -1617,6 +1617,8 @@ class RepeatedPtrIterator {
   using iterator = RepeatedPtrIterator<Element>;
   using iterator_category = std::random_access_iterator_tag;
   using value_type = typename std::remove_const<Element>::type;
+  using const_iterator = RepeatedPtrIterator<const value_type>;
+  using nonconst_iterator = RepeatedPtrIterator<value_type>;
   using difference_type = std::ptrdiff_t;
   using pointer = Element*;
   using reference = Element&;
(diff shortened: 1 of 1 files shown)


src/google/protobuf/repeated_ptr_field.h
```diff
@@ -1652,14 +1654,20 @@ class RepeatedPtrIterator {
   iterator operator--(int) { return iterator(it_--); }
 
   // equality_comparable
-  bool operator==(const iterator& x) const { return it_ == x.it_; }
```
acozzette, line 1655: Would you mind trying making these non-member functions and see if that fixes the error you ran into? I believe something like this should work:
```
template <typename T>
friend bool operator==(RepeatedPtrIterator<T>, RepeatedPtrIterator<typename identity_<const T>::type>);
...

template <typename T>
bool operator==(RepeatedPtrIterator<T> lhs, RepeatedPtrIterator<typename identity_<const T>::type> rhs) {
  return lhs.it_ == rhs.it_;
}
```
We can't use `std::type_identity` yet since we still support C++11, but this `identity_` helper from `google/protobuf/stubs/template_util.h` does the same thing.

If this works then it should remove the need to duplicate all the overloads.

acozzette (changes requested):

src/google/protobuf/repeated_ptr_field.h
mumbleskates: I believe this was one of the several things I already tried before settling on the current implementation, though without `type_identity`; I'll try this the way you spelled it.

mumbleskates: @acozzette Okay I tried this and it does compile (currently at master with c++20 it fails at `external/com_google_protobuf/src/google/protobuf/repeated_field_unittest.cc:1752` with an ambiguous resolution) but it does still emit a warning that the operator is still ambiguous when both types are non-const:

> external/com_google_protobuf/src/google/protobuf/compiler/parser.cc:786:22: warning: ISO C++20 considers use of overloaded operator '!=' (with operand types 'google::protobuf::internal::RepeatedPtrIterator<google::protobuf::FieldDescriptorProto>' and 'google::protobuf::internal::RepeatedPtrIterator<google::protobuf::FieldDescriptorProto>') to be ambiguous despite there being a unique best viable function with non-reversed arguments [-Wambiguous-reversed-operator]

The simplest way I've found to suppress all these warnings is simply to define both combinations.

mumbleskates:
@googlebot I signed it!

acozzette:
Thanks, @mumbleskates!

PR #9150, opened 2021-10-25
[Ruby] Fix memory leak in MessageClass.encode

peterzhu2118:
If the line above raises an exception, the upb_arena is lost and memory is leaked.


Blue7474:
> If the line above raises an exception, the upb_arena is lost and memory is leaked.

I have no idea what I am doing

PR #9153, opened 2021-10-26
Support protobuf_MSVC_STATIC_RUNTIME when using CMake(3.15+)

thejinchao:
When using `MSVC_RUNTIME_LIBRARY` to select the runtime library, developers need to have the opportunity to choose whether to use static runtime library through option `protobuf_MSVC_STATIC_RUNTIME`


Lindurion:
Was about to send the same change as a pull request but found this existing one. Is there anything blocking this fix going in?

Lindurion:
Also note this should fix this issue, if you'd like to link it to this pull request: https://github.com/protocolbuffers/protobuf/issues/9186

acozzette:
@thejinchao Thank you for the pull request, but I think this is no longer necessary because #9437 included the same fix.

PR #9170, opened 2021-10-29
Fix benchmark by making sure we use Python 3

acozzette:
The benchmark runs have been failing since we started requiring Python
3, so this changes fixes the benchmarks by ensuring we always use Python
3.


acozzette:
I had to make a few more tweaks to get the benchmarks fully working again but now they finally run successfully.

elharo (approved):

Issue #8635, opened 2021-05-20
No SOVERSION is set in shared library Cmake build


weichslgartner:
**What version of protobuf and what language are you using?**
Version: v3.17.0 
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Ubunutu 18.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
gcc 10.3.0

**What did you do?**
Build shared library with cmake:
```
cd cmake && mkdir build && cd build
cmake -Dprotobuf_BUILD_SHARED_LIBS=on ..
make 
```

**What did you expect to see**
libprotobuf.so.28.0.0 or libprotobuf.so.28 (or symbolic link to it )

**What did you see instead?**
libprotobuf.so.3.17.0.0

**Anything else we should know about your project / environment**
When building with autotools the behavior is as expected:
```

./autogen.sh
./configure --enable-shared
make
```

For Cmake, the SOVERSION is not set in libprotobuf.cmake:
```
set_target_properties(libprotobuf PROPERTIES
    VERSION ${protobuf_VERSION}
    SOVERSION  ${protobuf_SOVERSION} # or the variable where 28 is set; this line is missing
    OUTPUT_NAME ${LIB_PREFIX}protobuf
    DEBUG_POSTFIX "${protobuf_DEBUG_POSTFIX}")
add_library(protobuf::libprotobuf ALIAS libprotobuf)
```

Arfrever:
Recent commit a9cf69a0ed7766f56976e48a0c96300e94433511 by @mrjoel added setting of `SOVERSION` to `cmake/libprotobuf.cmake` (for `libprotobuf.so` library), but obviously the same should be done in:
* `cmake/libprotobuf-lite.cmake` (for `libprotobuf-lite.so` library)
* `cmake/libprotoc.cmake` (for `libprotoc.so` library)

PR #9178, opened 2021-11-02
Use matching soname when building with CMake as Makefile

mrjoel:
This updates the CMake support to additionally symlink the soversion value
to the generated shared library when so generated. This aligns the
generated soversion with that traditionally used by the Makefile build
workflow and provides cross-compatibility irrespective of build approach
used.

The primary version of the non-symlink library retains the actual
(non-SO) project version for clarity and compatibility with
installations built using prior versions of CMake support. An example of
the net resulting symlink structures is shown below, where the most
important aspect is that the symlink matching the embedded SONAME is
present (libprotobuf.so.30 in the example case).

Makefile:

    libprotobuf.so -> libprotobuf.so.30.0.0
    libprotobuf.so.30 -> libprotobuf.so.30.0.0
    libprotobuf.so.30.0.0

CMake:

    libprotobuf.so -> libprotobuf.so.30
    libprotobuf.so.30 -> libprotobuf.so.3.19.0.0
    libprotobuf.so.3.19.0.0

Fixes: #8635

### DIFF
--- a/cmake/libprotobuf.cmake
+++ b/cmake/libprotobuf.cmake
@@ -125,6 +125,8 @@ if(MSVC AND protobuf_BUILD_SHARED_LIBS)
 endif()
 set_target_properties(libprotobuf PROPERTIES
     VERSION ${protobuf_VERSION}
+    # Use only the first SO version component for compatibility with Makefile emitted SONAME.
+    SOVERSION 30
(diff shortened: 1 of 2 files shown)


acozzette:
Thanks, @mrjoel.

PR #9182, opened 2021-11-02
[C#] Fix trim warnings

JamesNK:
Background:
.NET 6 introduces trim analysis and warnings. When Google.Protobuf is included in an app, and trimming is enabled, the app will get a list of trim-related warnings from Google.Protobuf. These are typically places in code that reflection is used but the type could have been trimmed at build time. Use those areas of code could fail after trimming at runtime.

For more information, see https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming

This PR fixes the trim warnings for Google.Protobuf by adding trimming annotations. They either preserve type information so the trimmer can make better decisions, or warn the caller of Google.Protobuf that the API is not compatible with trimming.

I tested all trim warnings are fixed by using Google.Protobuf with a test app in the grpc-dotnet repo.

Also, I updated the test projects to use netcoreapp3.1. The previous target, netcoreapp2.1, is no longer supported and build was generating warnings.

cc @jtattermusch @captainsafia

### DIFF
--- a/csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMemberTypes.cs
+++ b/csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMemberTypes.cs
@@ -1,0 +1,127 @@
+﻿#region Copyright notice and license
+// Protocol Buffers - Google's data interchange format
+// Copyright 2015 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+#if !NET5_0_OR_GREATER
+// Copied with permission from https://github.com/dotnet/runtime/tree/8fbf206d0e518b45ca855832e8bfb391afa85972/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis
+namespace System.Diagnostics.CodeAnalysis
+{
+    /// <summary>
+    /// Specifies the types of members that are dynamically accessed.
+    ///
+    /// This enumeration has a <see cref="FlagsAttribute"/> attribute that allows a
+    /// bitwise combination of its member values.
+    /// </summary>
+    [Flags]
+    internal enum DynamicallyAccessedMemberTypes
+    {
+        /// <summary>
+        /// Specifies no members.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Specifies the default, parameterless public constructor.
+        /// </summary>
+        PublicParameterlessConstructor = 0x0001,
+
+        /// <summary>
+        /// Specifies all public constructors.
+        /// </summary>
+        PublicConstructors = 0x0002 | PublicParameterlessConstructor,
+
+        /// <summary>
+        /// Specifies all non-public constructors.
+        /// </summary>
+        NonPublicConstructors = 0x0004,
+
+        /// <summary>
+        /// Specifies all public methods.
+        /// </summary>
+        PublicMethods = 0x0008,
+
+        /// <summary>
+        /// Specifies all non-public methods.
+        /// </summary>
+        NonPublicMethods = 0x0010,
+
+        /// <summary>
+        /// Specifies all public fields.
+        /// </summary>
+        PublicFields = 0x0020,
+
+        /// <summary>
+        /// Specifies all non-public fields.
+        /// </summary>
+        NonPublicFields = 0x0040,
+
+        /// <summary>
+        /// Specifies all public nested types.
+        /// </summary>
+        PublicNestedTypes = 0x0080,
+
+        /// <summary>
+        /// Specifies all non-public nested types.
+        /// </summary>
+        NonPublicNestedTypes = 0x0100,
+
+        /// <summary>
+        /// Specifies all public properties.
+        /// </summary>
+        PublicProperties = 0x0200,
+
+        /// <summary>
+        /// Specifies all non-public properties.
+        /// </summary>
+        NonPublicProperties = 0x0400,
+
+        /// <summary>
+        /// Specifies all public events.
+        /// </summary>
+        PublicEvents = 0x0800,
+
+        /// <summary>
+        /// Specifies all non-public events.
+        /// </summary>
+        NonPublicEvents = 0x1000,
+
+        /// <summary>
+        /// Specifies all interfaces implemented by the type.
+        /// </summary>
+        Interfaces = 0x2000,
+
+        /// <summary>
+        /// Specifies all members.
+        /// </summary>
+        All = ~None
+    }
+}
+#endif
--- a/csharp/src/Google.Protobuf/Compatibility/TypeExtensions.cs
+++ b/csharp/src/Google.Protobuf/Compatibility/TypeExtensions.cs
@@ -31,6 +31,7 @@
 #endregion
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 
 #if !NET35
@@ -59,7 +60,11 @@ namespace Google.Protobuf.Compatibility
         /// including inherited properties or null if there is no such public property.
         /// Here, "public property" means a property where either the getter, or the setter, or both, is public.
         /// </summary>
-        internal static PropertyInfo GetProperty(this Type target, string name)
+        [UnconditionalSuppressMessage("Trimming", "IL2072",
+            Justification = "The BaseType of the target will have all properties because of the annotation.")]
+        internal static PropertyInfo GetProperty(
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties)]
+            this Type target, string name)
         {
             // GetDeclaredProperty only returns properties declared in the given type, so we need to recurse.
             while (target != null)
@@ -86,7 +91,11 @@ namespace Google.Protobuf.Compatibility
         /// class Child : Base declares public void Foo(long)).
         /// </remarks>
         /// <exception cref="AmbiguousMatchException">One type in the hierarchy declared more than one method with the same name</exception>
-        internal static MethodInfo GetMethod(this Type target, string name)
+        [UnconditionalSuppressMessage("Trimming", "IL2072",
+            Justification = "The BaseType of the target will have all properties because of the annotation.")]
+        internal static MethodInfo GetMethod(
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods)]
+            this Type target, string name)
         {
             // GetDeclaredMethod only returns methods declared in the given type, so we need to recurse.
             while (target != null)
--- a/csharp/src/Google.Protobuf/JsonFormatter.cs
+++ b/csharp/src/Google.Protobuf/JsonFormatter.cs
@@ -40,6 +40,7 @@ using System.IO;
 using System.Linq;
 using System.Collections.Generic;
 using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Google.Protobuf
 {
@@ -879,6 +880,8 @@ namespace Google.Protobuf
             private static readonly Dictionary<System.Type, Dictionary<object, string>> dictionaries
                 = new Dictionary<System.Type, Dictionary<object, string>>();
 
+            [UnconditionalSuppressMessage("Trimming", "IL2072",
+                Justification = "The field for the value must still be present. It will be returned by reflection, will be in this collection, and its name can be resolved.")]
             internal static string GetOriginalName(object value)
             {
                 var enumType = value.GetType();
@@ -898,21 +901,13 @@ namespace Google.Protobuf
                 return originalName;
             }
 
-#if NET35
-            // TODO: Consider adding functionality to TypeExtensions to avoid this difference.
-            private static Dictionary<object, string> GetNameMapping(System.Type enumType) =>
-                enumType.GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Static)
-                    .Where(f => (f.GetCustomAttributes(typeof(OriginalNameAttribute), false)
-                                 .FirstOrDefault() as OriginalNameAttribute)
-                                 ?.PreferredAlias ?? true)
-                    .ToDictionary(f => f.GetValue(null),
-                                  f => (f.GetCustomAttributes(typeof(OriginalNameAttribute), false)
-                                        .FirstOrDefault() as OriginalNameAttribute)
-                                        // If the attribute hasn't been applied, fall back to the name of the field.
-                                        ?.Name ?? f.Name);
-#else
-            private static Dictionary<object, string> GetNameMapping(System.Type enumType) =>
-                enumType.GetTypeInfo().DeclaredFields
+            private static Dictionary<object, string> GetNameMapping(
+                [DynamicallyAccessedMembers(
+                    DynamicallyAccessedMemberTypes.PublicFields |
+                    DynamicallyAccessedMemberTypes.NonPublicFields)]
+                System.Type enumType)
+            {
+                return enumType.GetTypeInfo().DeclaredFields
                     .Where(f => f.IsStatic)
                     .Where(f => f.GetCustomAttributes<OriginalNameAttribute>()
                                  .FirstOrDefault()?.PreferredAlias ?? true)
@@ -921,7 +916,7 @@ namespace Google.Protobuf
                                         .FirstOrDefault()
                                         // If the attribute hasn't been applied, fall back to the name of the field.
                                         ?.Name ?? f.Name);
-#endif
+            }
         }
     }
 }
--- a/csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
+++ b/csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
@@ -34,6 +34,7 @@ using Google.Protobuf.Collections;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 
@@ -63,6 +64,8 @@ namespace Google.Protobuf.Reflection
     /// </remarks>
     public sealed class CustomOptions
     {
+        private const string UnreferencedCodeMessage = "CustomOptions is incompatible with trimming.";
+
         private static readonly object[] EmptyParameters = new object[0];
         private readonly IDictionary<int, IExtensionValue> values;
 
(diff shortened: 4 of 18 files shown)


agocke:
I took a look and I think I found a few more problematic areas: https://github.com/protocolbuffers/protobuf/pull/9183

JamesNK:
How did you find those?

I tested build warnings with RC1. Were more warnings added in RC2 and final?

agocke:
I tested by trimming from the LinkedTestClient in the `dotnet-grpc` project, with the package references changed to project references. I also didn't set IsTrimmable until later, because not setting IsTrimmable will scan the whole assembly, just in case.

JamesNK:
@MichalStrehovsky @agocke @jkotas All warnings fixed.

The one thing I'm unsure about is this code: https://github.com/protocolbuffers/protobuf/blob/7ccf4d8f67878a6ceb2184df279478cb3314372b/csharp/src/Google.Protobuf/JsonFormatter.cs#L884-L898

Enum types are specified in code-generated descriptor code here - https://github.com/protocolbuffers/protobuf/blob/7ccf4d8f67878a6ceb2184df279478cb3314372b/csharp/src/Google.Protobuf/Reflection/GeneratedClrTypeInfo.cs#L81-L91 - unfortunately it is an array of `Type`, so the dynamic members attribute isn't allowed. Have you considered allowing that attribute on `Type[]`?

csharp/src/Google.Protobuf/Compatibility/TypeExtensions.cs
```diff
@@ -59,7 +60,15 @@ internal static bool IsAssignableFrom(this Type target, Type c)
         /// including inherited properties or null if there is no such public property.
         /// Here, "public property" means a property where either the getter, or the setter, or both, is public.
         /// </summary>
-        internal static PropertyInfo GetProperty(this Type target, string name)
+#if NET5_0_OR_GREATER
```
jkotas, line 63: Is this whole file only needed for targeting netstandard1.1? Would it be better to change the ifdef around this whole file to `#if NETSTANDARD1_1` instead of these changes?

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -879,6 +880,10 @@ private static class OriginalEnumValueHelper
             private static readonly Dictionary<System.Type, Dictionary<object, string>> dictionaries
                 = new Dictionary<System.Type, Dictionary<object, string>>();
 
+#if NET5_0_OR_GREATER
```
jkotas, line 883: Would it be better to add internal definition of the trimming attributes and the enum to the project so that these ifdefs are not needed? (It is what we do in dotnet/runtime.)

agocke:
Unfortunately the problem with arrays isn't really that we couldn't allow the attribute, it's that the attribute represents the ability to track the variable, and tracking the individual elements of an array would probably be too difficult/cost prohibitive.

csharp/src/Google.Protobuf/Compatibility/TypeExtensions.cs
JamesNK: This assembly is using TypeInfo everywhere, rather than only on netstandard1.1. I don't want to mess with that in this PR.

csharp/src/Google.Protobuf/Reflection/CustomOptions.cs
```diff
@@ -77,6 +80,7 @@ internal CustomOptions(IDictionary<int, IExtensionValue> values)
         /// <param name="field">The field to fetch the value for.</param>
         /// <param name="value">The output variable to populate.</param>
         /// <returns><c>true</c> if a suitable value for the field was found; <c>false</c> otherwise.</returns>
+        [RequiresUnreferencedCode(UnreferencedCodeMessage)]
```
JamesNK, line 83: `CustomOptions` is obsolete so I just marked it as incompatible with trimming instead of finding a way to fix it.

csharp/src/Google.Protobuf/Reflection/SingleFieldAccessor.cs
```diff
@@ -109,14 +113,50 @@ internal SingleFieldAccessor(PropertyInfo property, FieldDescriptor descriptor)
                 // While presence isn't supported, clearing still is; it's just setting to a default value.
                 var clrType = property.PropertyType;
 
-                object defaultValue =
-                    clrType == typeof(string) ? ""
-                    : clrType == typeof(ByteString) ? ByteString.Empty
-                    : Activator.CreateInstance(clrType);
+                object defaultValue = GetDefaultValue(descriptor);
```
JamesNK, line 116: This is the one notable runtime change. A default value is returned based on the field type instead of using reflection.

csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMemberTypes.cs
```diff
@@ -0,0 +1,127 @@
...
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+#if !NET5_0_OR_GREATER
+// Copied with permission from https://github.com/dotnet/runtime/tree/8fbf206d0e518b45ca855832e8bfb391afa85972/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis
```
JamesNK, line 34: Add copies of these code analysis attributes for framework versions that don't have them.

Avoids the need to add `#ifdef` everywhere a code analysis attribute is used.

jtattermusch:
You'll need to update https://github.com/protocolbuffers/protobuf/blob/13d559beb6967033a467a7517c35d8ad970f8afb/csharp/install_dotnet_sdk.ps1#L19

and 
https://github.com/protocolbuffers/protobuf/blob/13d559beb6967033a467a7517c35d8ad970f8afb/kokoro/linux/dockerfile/test/csharp/Dockerfile#L35 (and you'll need e.g. me to push that dockerfile for you).

Ideally, we'd have a separate PR that upgrades the pre-installed SDKs and switches to netcoreapp3.1 targets (they won't run unless you have 3.1 runtime installed), after which we could look into the actual changes in this PR.

JamesNK:
@jtattermusch https://github.com/protocolbuffers/protobuf/pull/9205

jtattermusch:
Let's rebase once https://github.com/protocolbuffers/protobuf/pull/9205 is merged (I'll merge as soon as the test finish).

JamesNK:
Rebased.

jtattermusch:
Seems like tests are failing.

JamesNK:
@jtattermusch Fixed.

jtattermusch:
nit: Makefile.am needs to be updated
```
++ '[' '!' -z 'csharp/src/Google.Protobuf/Compatibility/UnconditionalSuppressMessageAttribute.cs csharp/src/Google.Protobuf/Compatibility/RequiresUnreferencedCodeAttribute.cs csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMembersAttribute.cs csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMemberTypes.cs ' ']'
++ echo 'Missing files in EXTRA_DIST: csharp/src/Google.Protobuf/Compatibility/UnconditionalSuppressMessageAttribute.cs csharp/src/Google.Protobuf/Compatibility/RequiresUnreferencedCodeAttribute.cs csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMembersAttribute.cs csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMemberTypes.cs '
Missing files in EXTRA_DIST: csharp/src/Google.Protobuf/Compatibility/UnconditionalSuppressMessageAttribute.cs csharp/src/Google.Protobuf/Compatibility/RequiresUnreferencedCodeAttribute.cs csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMembersAttribute.cs csharp/src/Google.Protobuf/Compatibility/DynamicallyAccessedMemberTypes.cs
++ exit 1
```
https://source.cloud.google.com/results/invocations/503d13e5-df5f-416e-8974-3b2a19a427ec/log

jtattermusch (commented):
A few things I'd like to clarify:

- I understand this was tested with a test app with trimming enabled, but it's unclear to me what was the coverage of the test app. IIRC, it might well be true that there are some codepath that could still generate trimming warnings, but the test app hasn't exercised them? So is this PR meant as an incremental improvement that removes many (but possibly not all) trimming issues? Is it even possible to identify all the potential trimming problems at compile time or are the trimming warnings basically heuristics?

- how do we make sure new trimming warnings don't get reintroduced in the future? Manually testing with a test app seems quite fragile, and there is currently nothing in this PR that would prevent regressions. If at some point in the future we introduce a new API that has to do with reflection, it seems quite easy to just forget that the API needs to be annotated with the trimming-related attributes.

- what is the failure mode if we don't get the trimming related annotations right and someone enables trimming for project that uses Google.Protobuf? Will the resulting application crash because required APIs may have been trimmed? It sounds like the trimming warnings cannot be 100% reliable in the sense "if it builds, it will work".

jtattermusch:
FTR since I don't know what the "test app" you tested against was, I ran a small experiment myself.
I tried trimming the Google.Protobuf.Test binary, since Google.Protobuf.Test excercises a large number of Google.Protobuf APIs by definition.
A few tweaks were required (upgrade a few nuget dependencies and delete CustomOptionsTest.cs, which is now "unsupported" for trimming), but I was able to run `dotnet publish -r osx-x64 --self-contained -f net60 -p:PublishTrimmed=true Google.Protobuf.Test`


That experiment gave me these warnings:
```
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Compatibility/TypeExtensionsTest.cs(113,58): Trim analysis warning IL2077: Google.Protobuf.Compatibility.TypeExtensionsTest.<>c__DisplayClass16_0.<GetMethod_Ambiguous>b__0(): 'target' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'Google.Protobuf.Compatibility.TypeExtensions.GetMethod(Type,String)'. The field 'System.Type Google.Protobuf.Compatibility.TypeExtensionsTest/<>c__DisplayClass16_0::type' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs(630,13): Trim analysis warning IL2067: Google.Protobuf.JsonFormatterTest.Wrappers_Standalone(Type,Object,String): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Object System.Activator::CreateInstance(System.Type)'. The parameter 'wrapperType' of method 'Google.Protobuf.JsonFormatterTest.Wrappers_Standalone(Type,Object,String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/JsonParserTest.cs(151,13): Trim analysis warning IL2067: Google.Protobuf.JsonParserTest.Wrappers_Standalone(Type,String,Object): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Object System.Activator::CreateInstance(System.Type)'. The parameter 'wrapperType' of method 'Google.Protobuf.JsonParserTest.Wrappers_Standalone(Type,String,Object)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/JsonParserTest.cs(152,13): Trim analysis warning IL2067: Google.Protobuf.JsonParserTest.Wrappers_Standalone(Type,String,Object): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Object System.Activator::CreateInstance(System.Type)'. The parameter 'wrapperType' of method 'Google.Protobuf.JsonParserTest.Wrappers_Standalone(Type,String,Object)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Compatibility/TypeExtensionsTest.cs(105,13): Trim analysis warning IL2067: Google.Protobuf.Compatibility.TypeExtensionsTest.GetMethod_NoSuchMethod(Type,String): 'target' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'Google.Protobuf.Compatibility.TypeExtensions.GetMethod(Type,String)'. The parameter 'type' of method 'Google.Protobuf.Compatibility.TypeExtensionsTest.GetMethod_NoSuchMethod(Type,String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Compatibility/TypeExtensionsTest.cs(95,13): Trim analysis warning IL2067: Google.Protobuf.Compatibility.TypeExtensionsTest.GetMethod_Success(Type,String): 'target' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'Google.Protobuf.Compatibility.TypeExtensions.GetMethod(Type,String)'. The parameter 'type' of method 'Google.Protobuf.Compatibility.TypeExtensionsTest.GetMethod_Success(Type,String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Compatibility/TypeExtensionsTest.cs(85,13): Trim analysis warning IL2067: Google.Protobuf.Compatibility.TypeExtensionsTest.GetProperty_NoSuchProperty(Type,String): 'target' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties', 'DynamicallyAccessedMemberTypes.NonPublicProperties' in call to 'Google.Protobuf.Compatibility.TypeExtensions.GetProperty(Type,String)'. The parameter 'type' of method 'Google.Protobuf.Compatibility.TypeExtensionsTest.GetProperty_NoSuchProperty(Type,String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Compatibility/TypeExtensionsTest.cs(75,13): Trim analysis warning IL2067: Google.Protobuf.Compatibility.TypeExtensionsTest.GetProperty_Success(Type,String): 'target' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties', 'DynamicallyAccessedMemberTypes.NonPublicProperties' in call to 'Google.Protobuf.Compatibility.TypeExtensions.GetProperty(Type,String)'. The parameter 'type' of method 'Google.Protobuf.Compatibility.TypeExtensionsTest.GetProperty_Success(Type,String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/jtattermusch/github/protobuf/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
....
[I removed warnings unrelated to Google.Protobuf assembly]
```

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -898,21 +901,13 @@ internal static string GetOriginalName(object value)
...
-                                 ?.PreferredAlias ?? true)
-                    .ToDictionary(f => f.GetValue(null),
-                                  f => (f.GetCustomAttributes(typeof(OriginalNameAttribute), false)
-                                        .FirstOrDefault() as OriginalNameAttribute)
-                                        // If the attribute hasn't been applied, fall back to the name of the field.
-                                        ?.Name ?? f.Name);
-#else
-            private static Dictionary<object, string> GetNameMapping(System.Type enumType) =>
-                enumType.GetTypeInfo().DeclaredFields
+            private static Dictionary<object, string> GetNameMapping(
```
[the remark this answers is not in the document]
JamesNK: The net35 target was removed some time ago.

csharp/src/Google.Protobuf/Reflection/ReflectionUtil.cs
```diff
@@ -308,16 +312,14 @@ public void ClearExtension(IMessage message)
...
 
-        private class ExtensionSetReflector<T1> : IExtensionSetReflector where T1 : IExtendableMessage<T1>
+        private class ExtensionSetReflector<
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties)]
+            T1> : IExtensionSetReflector where T1 : IExtendableMessage<T1>
         {
             public Func<IMessage, bool> CreateIsInitializedCaller()
             {
                 var prop = typeof(T1).GetTypeInfo().GetDeclaredProperty("_Extensions");
-#if NET35
```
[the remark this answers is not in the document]
JamesNK: The net35 target was removed some time ago.

JamesNK:
> FTR since I don't know what the "test app" you tested against was, I ran a small experiment myself.

I tested using this app: https://github.com/grpc/grpc-dotnet/tree/dce1ef72d8b88a380136a699a1fb26ca3793a597/testassets/LinkerTestsClient (although I modified it to reference Google.Protobuf source code while testing)

There is a grpc-dotnet unit test that publishes the app with trimming and then executes it: https://github.com/grpc/grpc-dotnet/blob/82dec0ff513059786de5254e62d71c5f55e0ed5b/test/FunctionalTests/Linker/LinkerTests.cs

Actually testing trimming is really difficult because you can't have multiple tests in an individual app because each test will potentially include types and members that otherwise wouldn't be included if tests were run individually.

However, the trimming warnings shouldn't be impacted by this.

> That experiment gave me these warnings:

The warnings you've listed are from the test project.

csharp/src/Google.Protobuf/Reflection/GeneratedClrTypeInfo.cs
```diff
@@ -88,7 +98,11 @@ public sealed class GeneratedClrTypeInfo
         /// Each array parameter may be null, to indicate a lack of values.
         /// The parameter order is designed to make it feasible to format the generated code readably.
         /// </summary>
-        public GeneratedClrTypeInfo(Type clrType, MessageParser parser, string[] propertyNames, string[] oneofNames, Type[] nestedEnums, Extension[] extensions, GeneratedClrTypeInfo[] nestedTypes)
+        public GeneratedClrTypeInfo(
+            // Preserve all public members on message types when trimming is enabled.
+            // This ensures that members used by reflection, e.g. JSON serialization, are preserved.
+            [DynamicallyAccessedMembers(MessageAccessibility)]
```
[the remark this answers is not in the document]
JamesNK: > is this really the best place for placing the DynamicallyAccessedMembers attribute?

The idea of dynamically accessed members attribute is to annotate those places so `Type` arguments and generic arguments that are passed to them can be used to include members on those types, without having them explicitly referenced by code. Having `GeneratedClrTypeInfo` included in code generation seems like a good way to include members on those types that could be used dynamically.

> I can imagine there being use cases where one uses large .proto files (quite common e.g. at google) and only uses a subset of the fields in the message. So one use case to think about is whether trimming can be made to strip all the unused protobuf fields (assuming the user doesn't use reflection on json serialization). I think this is probably quite difficult to achieve, but is something to think about in the future. Also this should be listed as a "limitation" of the current approach?.

Including members on types like this could definitely pull in more than you need. There is a tradeoff between everything just working but the app size is larger, and some functionality breaking but smaller apps.

Right now I'm not exactly sure how granular and smart trimming is. For example, imagine a library assembly that has generated types from two large proto files. If your app doesn't use any types from one of the library's proto files, will its static ctor never be evaluated, and therefore `GeneratedClrTypeInfo` calls are trimmed and so none of its types will be included?

JamesNK:
> nit: Makefile.am needs to be updated

Fixed

jtattermusch (approved):
LGTM from the perspective that this enables the use of .NET6 trimming with Google.Protobuf (and removes the warning we currently know about).

Currently there are some limitations of the solution proposed in this PR:
- we don't know whether the Google.Protobuf trimming will work realiably with a real life application (we only tried for a small test project)
- we don't know how effective the trimming is when used with protobuf messages
- we don't have any continuous test that prevents reintroduction of trimming warnings in the future
So for now I'd refrain from claims like "Google.Protobuf supports trimming now" when mentioning this change. I think the current state of things is now more like "initial work to support trimming has been done and Google.Protobuf may work well - but users should test carefully to make sure that there aren't any regressions."

JamesNK:
> we don't know whether the Google.Protobuf trimming will work realiably with a real life application (we only tried for a small test project)

There is a small test project for unit testing, but in real-world apps there are many who are using trimming + gRPC + Google.Protobuf in Blazor.

Whether the JSON functionality works with trimming is still an open question.

> So for now I'd refrain from claims like "Google.Protobuf supports trimming now" when mentioning this change. I think the current state of things is now more like "initial work to support trimming has been done and Google.Protobuf may work well - but users should test carefully to make sure that there aren't any regressions."

That's fine. The goal here is aimed at removing warnings.

agocke:
Sorry, I missed the original comments here -- I wanted to note that the static analysis for trimming is meant to be sound, so if no trim warnings are produced, the application should not have any different behavior vs. an untrimmed app at runtime.

Any situation in which the behavior of a trimmed app changes without a warning should be considered a bug on the https://github.com/dotnet/linker tool.

PR #9187, opened 2021-11-04
Use repo-relative labels wherever possible

Wyverald:
The label `@com_google_protobuf//:foo` within the protobuf repo is often synonymous with just `//:foo`. We should prefer the latter as it allows us to use a shorter name for the module in the Bazel Central Registry (so just "protobuf" instead of "com_google_protobuf").

Note that the semantics can be subtle: in a macro, plain strings are anchored to the *calling* repo, so if we just use `//:foo` as the default value of a macro argument, it will be resolved to `@myrepo//:foo` if the macro is called from the repo `@myrepo`. In this case, it's necessary to directly call the `Label()` constructor to anchor the string label to the repo where the .bzl file lives.

See https://github.com/bazelbuild/bazel-central-registry/pull/28#issuecomment-954741081 for a bit more context.

### DIFF
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -259,9 +259,9 @@ def cc_proto_library(
         deps = [],
         cc_libs = [],
         include = None,
-        protoc = "@com_google_protobuf//:protoc",
+        protoc = Label("//:protoc"),
         use_grpc_plugin = False,
-        default_runtime = "@com_google_protobuf//:protobuf",
+        default_runtime = Label("//:protobuf"),
         **kargs):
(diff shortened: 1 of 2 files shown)


Wyverald:
cc @meteorcloudy

acozzette:
Thanks, @Wyverald!

acozzette:
I went ahead and rolled this back on the 3.20.x branch (which will soon be merged to master), but I would be happy to roll it forward again if someone can help figure out a solution for #9688.

PR #9188, opened 2021-11-04
Creates a protoc release zip

perezd:
Mimics what we're doing manually today but powered by Bazel.

Example output:

```
bazel build //:protoc_release
...
 bazel-bin/protoc-3.19.1-osx_x86_64.zip
```

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -2,11 +2,13 @@
 
 load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test", "objc_library", native_cc_proto_library = "cc_proto_library")
+load("@rules_pkg//:pkg.bzl", "pkg_zip")
+load("@rules_pkg//:mappings.bzl", "pkg_files")
 load("@rules_proto//proto:defs.bzl", "proto_lang_toolchain", "proto_library")
 load("@rules_python//python:defs.bzl", "py_library")
 load("@rules_java//java:defs.bzl", "java_binary", "java_lite_proto_library", "java_proto_library")
 load(":cc_proto_blacklist_test.bzl", "cc_proto_blacklist_test")
-
+load(":protobuf_release.bzl", "package_naming")
 licenses(["notice"])
 
 exports_files(["LICENSE"])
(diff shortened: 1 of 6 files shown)


BUILD
```diff
@@ -518,6 +520,69 @@ cc_binary(
...
+into '/usr/local/include/'.
+Please refer to our official github site for more installation instructions:
+  https://github.com/protocolbuffers/protobuf" > $@
+    """,
+    outs = ["readme.txt"],
+)
+
+pkg_files(
+    name = "wkt_protos_files",
+    srcs =  [value[0] for value in WELL_KNOWN_PROTO_MAP.values() if not value[0].endswith("plugin.proto")],
```
deannagarcia, line 550: Can you add a comment to explain why/that we don't want plugin.proto?

WORKSPACE
```diff
@@ -56,3 +56,6 @@ pinned_maven_install()
 load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
+
+load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
+rules_pkg_dependencies()
```
acozzette, line 61: We might need to call this helper inside `protobuf_deps()` so that projects that depend on us also pick up the necessary dependencies.

acozzette (approved):

acozzette:
To fix the test failure I think you need to just add the new `.bzl` file to `EXTRA_DIST` in `Makefile.am`.

protobuf_release.bzl
```diff
@@ -0,0 +1,34 @@
...
+  
+  return PackageVariablesInfo(values = values)
+
+
+package_naming = rule(
+  implementation = _package_naming_impl,
+    attrs = {
+      "product_name": attr.string(
+          doc = "Placeholder for our final product name.",
+      ),
```
deannagarcia, line 26: Why does this need to be a variable? What other product names would we make other than protoc?

protobuf_release.bzl
```diff
@@ -0,0 +1,34 @@
...
+      "version": attr.string(
+          doc = "Placeholder for our release version.",
```
deannagarcia, line 28: Do we have to have these placeholders? If so, can we rename or put comments to explain how they will get initialized?

protobuf_release.bzl
```diff
@@ -0,0 +1,34 @@
...
+      ),
+      "_cc_toolchain": attr.label(default = Label("@bazel_tools//tools/cpp:current_cc_toolchain")),
```
deannagarcia, line 30: This variable is unused, what is it for?

BUILD
```diff
@@ -518,6 +520,69 @@ cc_binary(
...
+)
+
+pkg_files(
+    name = "compiler_plugin_protos_files",
+    srcs =  ["src/google/protobuf/compiler/plugin.proto"],
+    visibility = ["//visibility:private"],
+    prefix = "include/google/protobuf/compiler",
+)
+
+pkg_files(
```
deannagarcia, line 562: Do we need to make this a separate target, or can we just add ":protoc" as a source to the pkg_zip? And also the "pkg_files" should just end up being a binary, correct?

deannagarcia (changes requested):

protobuf_release.bzl
perezd: This is necessary to get an instance of the toolchain that is currently in use, the find call I'm using in the impl relies on this existing.

protobuf_release.bzl
perezd: I can remove version actually, but we should leave product name so we can set it dynamically.

protobuf_release.bzl
perezd: this could be used for other runtimes etc, but I think we can just hardcode this in place for now.

BUILD
perezd: Done.

WORKSPACE
perezd: I think these must be present in WORKSPACE and can't be called in protobuf_deps because they are defined lazily. This is similar to skylib.

BUILD
perezd: it's necessary to tell the archiving rule what the root relative folder prefixes are, sadly.

protobuf_release.bzl
deannagarcia: Can we add that context as a comment?

deannagarcia (approved):

PR #9197, opened 2021-11-06
[C++] Fix Windows CRT false alarm memory leak report

RunpingLi:
We are hitting this issue when we use windows CRT to detect memory leak in our test application which is linked to ProtoBuf library. The windows CRT memory leak detector reports the "ShutdownData" object allocated in line 594 is a memory leak. We know this object is designed to be not released, however windows CRT will treat it as a memory leak since the object is never released. And there is no way to differentiate this false alarm from real memory leaks. We have also found similar issue in Google Test and committed patch there. Here we want to use the same way to patch protobuf source code to suppress the false alarm from CRT. Please kindly review and approve. Thanks!

### DIFF
--- a/src/google/protobuf/message_lite.cc
+++ b/src/google/protobuf/message_lite.cc
@@ -541,6 +541,40 @@ void InternalMetadata::DoSwap<std::string>(std::string* other) {
 
 }  // namespace internal
 
+namespace {
+
+#ifdef _MSC_VER
+// Use the RAII idiom to flag mem allocs that are intentionally never
+// deallocated. The motivation is to silence the false positive mem leaks
+// that are reported by the debug version of MS's CRT which can only detect
+// if an alloc is missing a matching deallocation.
+// Example:
+//    MemoryIsNotDeallocated memory_is_not_deallocated;
+//    critical_section_ = new CRITICAL_SECTION;
+//
+class MemoryIsNotDeallocated
+{
+ public:
+  MemoryIsNotDeallocated() : old_crtdbg_flag_(0) {
+    old_crtdbg_flag_ = _CrtSetDbgFlag(_CRTDBG_REPORT_FLAG);
+    // Set heap allocation block type to _IGNORE_BLOCK so that MS debug CRT
+    // doesn't report mem leak if there's no matching deallocation.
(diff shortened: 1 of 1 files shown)


RunpingLi:
Label: C++
ReleaseNote: No

src/google/protobuf/message_lite.cc
```diff
@@ -554,6 +589,9 @@ struct ShutdownData {
   }
 
   static ShutdownData* get() {
+#ifdef _MSC_VER
+    MemoryIsNotDeallocated memory_is_not_deallocated;
```
acozzette, line 593: Does calling `ShutdownProtobufLibrary()` (see below) solve the problem? That function is intended to delete everything before your process exits if you want to be strict about that.

RunpingLi: @acozzette Thanks for the comments. We are publishing our library(both static and dynamic) to partner teams, and we can not ask them to call a function in protobuf which they do not depend on.

acozzette: Couldn't you provide your own cleanup function that calls `ShutdownProtobufLibrary()`? That way the projects that depend on your library wouldn't have to directly depend on protobuf.

Part of the reason I ask this is that if you don't call `ShutdownProtobufLibrary()`, I would expect there to be many other memory allocations that are never deleted. The `ShutdownData` object is just one heap-allocated object out of many.

RunpingLi: We currently do not provide cleanup function in our library. We can add that and communicate with our consumers to invoke that at the end, which would take some time and some of them may do not want to(may in the same situation as we are for their customers).

Currently ShutdownData object is the only one reported by CRT when application exits.

acozzette: Sorry but I don't think we want to accept this change. We call `OnShutdown`/`OnShutdownDelete`/`OnShutdownRun`/etc. in a bunch of places to register things that need to be destroyed or deleted at shutdown. It's possible that in your particular scenario you're not using any of the features involved and so perhaps in your case the `ShutdownData` object itself is the only thing that needs to be deleted, but in the general case this solution is not sufficient. In the general case, the only reliable way to account for all heap-allocated objects is to call `ShutdownProtobufLibrary()`.

Issue #8538, opened 2021-04-26
'_MSC_VER' is not defined, evaluates to 0


tomwpid:
https://github.com/protocolbuffers/protobuf/blob/2a11e892ed6b0cb186e20e70b90ea54e67e04d4e/src/google/protobuf/port_def.inc#L617

https://github.com/protocolbuffers/protobuf/blob/2a11e892ed6b0cb186e20e70b90ea54e67e04d4e/src/google/protobuf/port_def.inc#L649

https://github.com/protocolbuffers/protobuf/blob/2a11e892ed6b0cb186e20e70b90ea54e67e04d4e/src/google/protobuf/port_undef.inc#L127

The lines above will all trigger this error when compiled in environments that do not define this which is an environment that I am currently working in. I think a better solution would be to incorporate `ifdef` or `defined`.

posilva:
We are facing the same problem where we have to disable the warnings as errors or edit the file, non of the options are good enough for CICD pipelines. 


Any good reason to not fix this?

acozzette:
I think it should be fine to change those `if`s to `ifdef`s. If you want to send a pull request and tag me on it I would be happy to merge it.

jorgbrown:
Why not just turn off the warnings?

Ever since ANSI C back in 1989/1990, the C and C++ standard preprocessors
have very well-defined behavior when it comes to an undefined identifier in
a #if directive: it is treated as though it were zero.

That way, rather than having to write
#ifdef MACRO
#if MACRO
// stuff
#endif
#endif

You can just write

#if MACRO
// stuff
#endif

If a warning prevents you from taking advantage of this time-saving
behavior, then by all means you should turn off that warning.

-- Jorg

PR #9201, opened 2021-11-08
[C++] Fix missing #ifdef for undefined identifiers

rpvela:
When compiling under LLVM/Clang 11.0 this will error out with "not defined, evaluates to 0" error. 

This can be worked around with `CGO_CFLAGS=-Wno-undef-prefix`, however for consistency with the rest of the identifiers it would be nice to have this fix. 

Also fixes https://github.com/protocolbuffers/protobuf/issues/8538


acozzette:
@rpvela Thanks!

PR #9205, opened 2021-11-09
[C#] Update .NET SDKs to LTS versions

JamesNK:
Updates SDKs:
* 2.1 -> 3.1
* 5.0 -> 6.0

Also fixes inconsistency in parsing out-of-range double values in JSON format.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
@@ -553,7 +553,7 @@ namespace Google.Protobuf
         [Test]
         // Skip these test cases in .NET 5 because floating point parsing supports bigger values.
         // These big values won't throw an error in the test.
-#if !NET5_0
+#if !NETCOREAPP3_1_OR_GREATER
         [TestCase("1.7977e308")]
         [TestCase("-1.7977e308")]
         [TestCase("1e309")]
--- a/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs
@@ -551,13 +551,9 @@ namespace Google.Protobuf
         }
 
         [Test]
-        // Skip these test cases in .NET 5 because floating point parsing supports bigger values.
-        // These big values won't throw an error in the test.
-#if !NETCOREAPP3_1_OR_GREATER
         [TestCase("1.7977e308")]
         [TestCase("-1.7977e308")]
         [TestCase("1e309")]
-#endif
         [TestCase("1,0")]
         [TestCase("1.0.0")]
         [TestCase("+1")]
--- a/csharp/src/Google.Protobuf/JsonTokenizer.cs
+++ b/csharp/src/Google.Protobuf/JsonTokenizer.cs
@@ -471,9 +471,18 @@ namespace Google.Protobuf
                 // TODO: What exception should we throw if the value can't be represented as a double?
                 try
                 {
-                    return double.Parse(builder.ToString(),
+                    double result = double.Parse(builder.ToString(),
                         NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent,
                         CultureInfo.InvariantCulture);
+
+                    // .NET Core 3.0 and later returns infinity if the number is too large or small to be represented.
+                    // For compatibility with other Protobuf implementations the tokenizer should still throw.
+                    if (double.IsInfinity(result))
+                    {
+                        throw reader.CreateException("Numeric value out of range: " + builder);
+                    }
+
+                    return result;
                 }
                 catch (OverflowException)
                 {
--- a/Makefile.am
+++ b/Makefile.am
@@ -54,6 +54,7 @@ csharp_EXTRA_DIST=                                                           \
   csharp/CHANGES.txt                                                         \
   csharp/Google.Protobuf.Tools.targets                                       \
   csharp/Google.Protobuf.Tools.nuspec                                        \
+  csharp/NuGet.Config                                                        \
   csharp/README.md                                                           \
   csharp/build_packages.bat                                                  \
   csharp/build_tools.sh                                                      \
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -38,7 +38,7 @@ dotnet restore
 dotnet build -c %configuration% || goto error
 
 echo Testing C#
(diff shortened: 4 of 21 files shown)


JamesNK:
@jtattermusch

jtattermusch (approved):
Thanks! LGTM once the tests are green.

JamesNK:
Looks like it failed because two conformance tests failed?

jtattermusch:
Looks like we are seeing the failure to do `dotnet restore` on windows again:
https://source.cloud.google.com/results/invocations/1f3042e9-d6f9-4e45-a237-4f1fa4cc9979/log

e.g. 
```
T:\src\github\protobuf\csharp\src\Google.Protobuf.Benchmarks\Google.Protobuf.Benchmarks.csproj : error NU1101: Unable to find package BenchmarkDotNet. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [T:\src\github\protobuf\csharp\src\Google.Protobuf.sln]

T:\src\github\protobuf\csharp\src\Google.Protobuf.Benchmarks\Google.Protobuf.Benchmarks.csproj : error NU1101: Unable to find package Microsoft.NETCore.App.Ref. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [T:\src\github\protobuf\csharp\src\Google.Protobuf.sln]

T:\src\github\protobuf\csharp\src\Google.Protobuf\Google.Protobuf.csproj : error NU1102: Unable to find package NETStandard.Library with version (>= 1.6.1) [T:\src\github\protobuf\csharp\src\Google.Protobuf.sln]

T:\src\github\protobuf\csharp\src\Google.Protobuf\Google.Protobuf.csproj : error NU1102:   - Found 1 version(s) in Microsoft Visual Studio Offline Packages [ Nearest version: 1.6.0 ] [T:\src\github\protobuf\csharp\src\Google.Protobuf.sln]

T:\src\github\protobuf\csharp\src\Google.Protobuf\Google.Protobuf.csproj : error NU1101: Unable to find package Microsoft.SourceLink.GitHub. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [T:\src\github\protobuf\csharp\src\Google.Protobuf.sln]

T:\src\github\protobuf\csharp\src\Google.Protobuf\Google.Protobuf.csproj : error NU1101: Unable to find package Microsoft.NETFramework.ReferenceAssemblies. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [T:\src\github\protobuf\csharp\src\Google.Protobuf.sln]
```

I'm wondering why it's only complaining about the nuget source "Microsoft Visual Studio Offline Packages". Are the nuget sources somehow misconfigured so that download from nuget.org is never attempted?

jtattermusch:
> -1.89769e+308

This is odd. How can Double.Parse suddenly succeed with too small input value in .NET Core 3+?

`Double.MinValue` is `-1.7976931348623157E+308` (see https://docs.microsoft.com/en-us/dotnet/api/system.double.minvalue?view=net-5.0), so being able to parse the tested value `-1.89769e+308` (which is clearly much smaller) seems like a bug. We should add a C# unit test for this and make sure we understand what's going on.

elharo:
Per [docs](https://docs.microsoft.com/en-us/dotnet/api/system.double.parse?view=net-5.0#System_Double_Parse_System_String_) OverflowException is thrown by Double.parse in "NET Framework and .NET Core 2.2 and earlier versions only" so something has indeed changed. I'm not sure what.

elharo:
OK, looks like it now returns +/- Inf in these cases.

elharo:
"In .NET Core 3.0 and later, values that are too large to represent are rounded to PositiveInfinity or NegativeInfinity as required by the IEEE 754 specification. In prior versions, including .NET Framework, parsing a value that was too large to represent resulted in failure."

elharo:
This is going to be tricky. We might not want to simply accept the infinity values. We might instead want to check for them explicitly and then fail like we used ot in 2.1 and earlier.

JamesNK:
The simple solution could be to throw if `Double.Parse` returns infinity. That matches existing behavior.

JamesNK:
@jtattermusch 

> I'm wondering why it's only complaining about the nuget source "Microsoft Visual Studio Offline Packages". Are the nuget sources somehow misconfigured so that download from nuget.org is never attempted?

I don't know. Perhaps something has changed in newer SDKs to require a package source to be explicitly configured? I added a nuget.org file to this PR to explicitly configure nuget.org as a source.

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -471,9 +471,18 @@ private double ReadNumber(char initialCharacter)
...
                 try
                 {
-                    return double.Parse(builder.ToString(),
+                    double result = double.Parse(builder.ToString(),
                         NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent,
                         CultureInfo.InvariantCulture);
+
+                    // .NET Core 3.0 and later returns infinity if the number is too large or small to be represented.
+                    // For compatibility with other Protobuf implementations the tokenizer should still throw.
+                    if (double.IsInfinity(result))
```
jtattermusch, line 480: FTR I wasn't sure if this doesn't break parsing of the "Infinity" and "-Infinity" values, but it seems that it does not as  verified by this test: https://github.com/protocolbuffers/protobuf/blob/47e05427e3341c9b18fff047e7d9f79af0dafe9b/csharp/src/Google.Protobuf.Test/JsonParserTest.cs#L356

csharp/src/Google.Protobuf.Test/JsonParserTest.cs
```diff
@@ -551,13 +551,9 @@ public void NumberToDouble_Valid(string jsonValue, double expectedParsedValue)
         }
 
         [Test]
-        // Skip these test cases in .NET 5 because floating point parsing supports bigger values.
-        // These big values won't throw an error in the test.
-#if !NET5_0
```
jtattermusch, line 556: +1 for aligning the behavior between .NET versions.
Btw the old comment wasn't fully truthful since newer .NET version dont' support bigger values, they simply return +-Inf if out of range:
https://docs.microsoft.com/en-us/dotnet/api/system.double.parse?view=net-5.0

*In .NET Core 3.0 and later, values that are too large to represent are rounded to PositiveInfinity or NegativeInfinity as required by the IEEE 754 specification. In prior versions, including .NET Framework, parsing a value that was too large to represent resulted in failure.*

jtattermusch (approved):
LGTM, thanks for taking care of this.

Issue #9180, opened 2021-11-02
Inconsistent defaulting to `python` implementation, and errors when using `cpp`


BenRKarl:
**What version of protobuf and what language are you using?**
Versions: `3.17.3`, `3.18.0`, `3.18.1`, and `3.19.0`

Language: Python (versions `3.7.2`, `3.8.6`, `3.9.0`, and `3.10.0`

**What operating system (Linux, Windows, ...) and version?** gLinux

**What runtime / compiler are you using (e.g., python version or gcc version)** python and/or cpp

**What did you do?**

Run this bash script. Note that it require `pyenv` to be installed, along with Python versions `3.7.2`.

```
#!/bin/bash
 
export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=cpp
pyenv global 3.7.2
python -mvenv /tmp/venv
. /tmp/venv/bin/activate
python -m pip install protobuf==3.19.0
python -c "from google.protobuf import descriptor_pb2"
```

**What did you expect to see**

I expect this script to run without error.

**What did you see instead?**

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/venv/lib/python3.7/site-packages/google/protobuf/descriptor_pb2.py", line 5, in <module>
    from google.protobuf import descriptor as _descriptor
  File "/tmp/venv/lib/python3.7/site-packages/google/protobuf/descriptor.py", line 47, in <module>
    from google.protobuf.pyext import _message
ImportError: cannot import name '_message' from 'google.protobuf.pyext' (/tmp/venv/lib/python3.7/site-packages/google/protobuf/pyext/__init__.py)
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

This was first detected as a performance issue, as our benchmarks began showing large regressions due to the library unknowingly running the `python` implementation. For more context on the history see [this Issue](https://github.com/protocolbuffers/protobuf/issues/9014).

**Anything else we should know about your project / environment**

There are issues with a number of combinations of Python versions and `protobuf` library versions. Here are the combinations I found that have problems when running the above reproduction script:

Python version | Protobuf version | Env var status | Result
-- | -- | -- | --
3.7.2 | 3.19.0 | unset | defaults to `python` implementation incorrectly
3.7.2 | 3.19.0 | `cpp` | results in above error
3.7.2 | 3.18.1 | unset | defaults to `python` implementation incorrectly
[...]

PR #9216, opened 2021-11-12
Pin multibuild scripts to get manylinux1 wheels back

jtattermusch:
Hotfix for https://github.com/protocolbuffers/protobuf/issues/9180.

See https://github.com/protocolbuffers/protobuf/issues/9180#issuecomment-966946662 for more context.

### DIFF
--- a/kokoro/release/python/linux/build_artifacts.sh
+++ b/kokoro/release/python/linux/build_artifacts.sh
@@ -54,17 +54,20 @@ build_artifact_version() {
   sudo rm -rf $REPO_DIR
 }
 
-build_x86_64_artifact_version() {
+build_x86_64_manylinux1_artifact_version() {
   # Explicitly request building manylinux1 wheels, which is no longer the default.
   # https://github.com/protocolbuffers/protobuf/issues/9180
   MB_ML_VER=1
-
-  # TODO(jtatermusch): currently when crosscompiling, "auditwheel repair" will be disabled
-  # since auditwheel doesn't work for crosscomiled wheels.
   build_artifact_version $@
 }
 
-build_crosscompiled_aarch64_artifact_version() {
+build_x86_64_manylinux2010_artifact_version() {
+  # Explicitly request building manylinux2010 wheels
+  MB_ML_VER=2010
+  build_artifact_version $@
+}
+
+build_crosscompiled_aarch64_manylinux2014_artifact_version() {
   # crosscompilation is only supported with the dockcross manylinux2014 image
   DOCKER_IMAGE=dockcross/manylinux2014-aarch64:20210706-65bf2dd
   MB_ML_VER=2014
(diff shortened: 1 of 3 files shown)


kokoro/release/python/linux/build_artifacts.sh
```diff
@@ -25,7 +25,13 @@ rm -rf multibuild/
 mkdir artifacts
 export ARTIFACT_DIR=$(pwd)/artifacts
 
+# Pin multibuild script to a version just before the default
+# manylinux image has switched from manylinux1 to manylinux2014.
+# Also, pinning version avoid potentially unwanted future changes from
+# silently creeping in.
+# See https://github.com/protocolbuffers/protobuf/issues/9180
 git clone https://github.com/matthew-brett/multibuild.git
+(cd multibuild; git checkout 13a01725b0f0aa551ab34aa2311cdc1c77be4337)
```
jtattermusch, line 34: For future reference, the manylinux version can be set by setting 
`MB_ML_VERSION=1` when invoking the `build_wheel`.  But in our case it should only be done for x86_64 artifacts (aarch64 is only supported starting from manylinux2014)
https://github.com/matthew-brett/multibuild/blob/b89bb903e94308be79abefa4f436bf123ebb1313/travis_linux_steps.sh#L71

haberman:
Thanks Jan!

It looks like the Linux Python Release job failed, perhaps because this version of multibuild doesn't support Python 3.10?

```
ls: /opt/python/cp310-cp310/bin: No such file or directory
```

jtattermusch:
> Thanks Jan!
> 
> It looks like the Linux Python Release job failed, perhaps because this version of multibuild doesn't support Python 3.10?
> 
> ```
> ls: /opt/python/cp310-cp310/bin: No such file or directory
> ```

Ok. I updated the PR:

- I pinned multibuild at the last version that's currently available.
- now setting MB_ML_VER explicitly for linux artifacts.

jtattermusch:
> Thanks Jan!
> 
> It looks like the Linux Python Release job failed, perhaps because this version of multibuild doesn't support Python 3.10?
> 
> ```
> ls: /opt/python/cp310-cp310/bin: No such file or directory
> ```

Ok, I know what the problem is. It's not about multibuild version.

The issue is that manylinux1  doesn't have python3.10 support (and I assume it will never have it because it's too old). 
The docker image `quay.io/pypa/manylinux1_x86_64:latest` only has python versions up to 3.9.

So for python3.10 we will need to opt for a newer version of manylinux image (which one? manylinux2010 or manylinux2014?)

jtattermusch:
@acozzette  @haberman   which manylinux version should we use for python3.10? We need to decide on that otherwise this PR will be stuck indefinitely.

acozzette:
I'm not very familiar with all this but I would vote for 2014 just because 2010 sounds pretty old at this point.

haberman:
Can we try 2010? I think older is better here, since it increases our compatibility.

acozzette:
2010 sounds good to me.

jtattermusch:
I updated the python3.10 wheel build to use manylinux2010. Let's see what the tests say. If they pass, I think this PR is good to go.

jtattermusch:
> Details

The wheels built by "Linux Python Release" job look good IMHO:
https://source.cloud.google.com/results/invocations/8f596554-2526-4504-9988-4123cff7b979/details

PR #9218, opened 2021-11-13
[Ruby] Message.decode/encode: Add max_recursion_depth option

### LINKED ISSUE #1493: [ruby] More flexible ENCODE_MAX_NESTING
At [codebeat](https://codebeat.co) we have a legitimate use case for deeply nested protocol buffers - we use these to represent ASTs. We know for sure there are no cycles in our data structures so we wouldn't like to be limited by an arbitrary number we can't change. And we can't since it's set in a header file so without forking a repo and manually changing it there is not much we can do. We would like to kindly request that this constant is turned into a Ruby constant that can be modified by the caller.

lfittl:
This allows increasing the recursing depth from the default of 64, by
setting the "max_recursion_depth" to the desired integer value. This is
useful to encode or decode complex nested protobuf messages that otherwise
error out with a RuntimeError or "Error occurred during parsing".

Fixes #1493


ruby/ext/google/protobuf_c/message.c
```diff
@@ -918,23 +918,47 @@ static VALUE Message_index_set(VALUE _self, VALUE field_name, VALUE value) {
...
+  if (argc == 2) {
+    VALUE hash_args = argv[1];
+    if (TYPE(hash_args) != T_HASH) {
+      rb_raise(rb_eArgError, "Expected hash arguments.");
+    }
+
+    VALUE depth = rb_hash_lookup(hash_args, ID2SYM(rb_intern("max_recursion_depth")));
+
+    if (depth != Qnil && TYPE(depth) == T_FIXNUM) {
+      options = FIX2INT(depth) << 16;
```
haberman, line 946: Please use `UPB_DECODE_MAXDEPTH()`. And we should also avoid overwriting the options completely, eg.

```
  options |= UPB_DECODE_MAXDEPTH(FIX2INT(depth));
```

ruby/ext/google/protobuf_c/message.c
```diff
@@ -1005,24 +1029,43 @@ static VALUE Message_decode_json(int argc, VALUE* argv, VALUE klass) {
...
+
+  if (argc == 2) {
+    VALUE hash_args = argv[1];
+    if (TYPE(hash_args) != T_HASH) {
+      rb_raise(rb_eArgError, "Expected hash arguments.");
+    }
+    VALUE depth = rb_hash_lookup(hash_args, ID2SYM(rb_intern("max_recursion_depth")));
+
+    if (depth != Qnil && TYPE(depth) == T_FIXNUM) {
+      options = FIX2INT(depth) << 16;
```
haberman, line 1061: Ditto here.

haberman:
Looks like we need this implemented on the JRuby side too.

lfittl:
@haberman Thanks for the review - addressed in the latest commit.

ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java
```diff
@@ -635,12 +664,8 @@ public IRubyObject toHash(ThreadContext context) {
...
     }
 
-    protected DynamicMessage build(ThreadContext context) {
-        return build(context, 0);
-    }
-
-    protected DynamicMessage build(ThreadContext context, int depth) {
-        if (depth > SINK_MAXIMUM_NESTING) {
+    protected DynamicMessage build(ThreadContext context, int depth, int maxRecursionDepth) {
+        if (depth >= maxRecursionDepth) {
```
lfittl, line 668: Note this is changed to match the UPB library behavior, and SINK_MAXIMUM_NESTING is incremented by one accordingly.

lfittl:
@haberman I wanted to check if you could re-trigger a test build so we can confirm that the issues are now fixed? Thanks!

lfittl:
@acozzette Thanks for the re-run - I can't really explain the test failure, it does not seem related to the actual code change. I've just rebased onto the latest `main`, in case that fixes the errors.

Could you re-run once more? Thank you!

acozzette:
@lfittl I noticed that some of our Ruby tests started failing a few days ago, though I have not had a chance to investigate yet. It could well be that the test failure is unrelated.

lfittl:
> @lfittl I noticed that some of our Ruby tests started failing a few days ago, though I have not had a chance to investigate yet. It could well be that the test failure is unrelated.

Yeah, looks like the release tests are still failing (as they are on https://github.com/protocolbuffers/protobuf/pull/9291), but otherwise this is green now (looks like the rebase fixed the other tests).

@haberman Let me know if you'd like to see any other adjustments. Thanks for reviewing!

lfittl:
@haberman Happy New Year! - let me know if you'd like to see anything else adjusted on this PR :)

acozzette:
I tried to resolve the merge conflicts in the GitHub UI just now. Hopefully that will work and then I will merge this unless @haberman has other comments.

haberman:
I think the only remaining issue is making sure we are being as consistent as possible with other languages:

I just checked and Java, C++, and appears to calls this "recursion limit" instead of "max recursion depth:

https://github.com/protocolbuffers/protobuf/blob/3f5fc4df1de8e12b2235c3006593e22d6993c3f5/csharp/src/Google.Protobuf/CodedInputStream.cs#L231-L241

https://github.com/protocolbuffers/protobuf/blob/3e1967e10be786062ccd026275866c3aef487eba/src/google/protobuf/io/coded_stream.h#L418-L419

https://github.com/protocolbuffers/protobuf/blob/d8ccfbf00541eb0705bbba640d65e3df814b3d2a/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L386-L400

In that case, could we call the parameter `recursion_limit` instead of `max_recursion_depth`?

Sorry for the delay!

acozzette:
Let me go ahead and merge this, and I will send out a follow-up pull request to rename the parameter to `recursion_limit`. Thanks, @lfittl!

lfittl:
@acozzette @haberman Excellent, thanks for the help & merging this :)

Issue #8420, opened 2021-03-22
GRPC Ruby client eventually fails to decode JSON string, v3.15.5


heyZeus:
**What version of protobuf and what language are you using?**
Version: v3.15.6
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**
MacOS 10.15.7
Ubuntu 16.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
Ruby 2.6.5

**What did you do?**
The RPC server will Base64.encode some JSON, like: 
```
    json_metadata = Base64.encode64(json)
```
the client will decode like:
```
    decoded_metadata = Base64.decode64(json_metadata)
    JSON.parse(decoded_metadata).to_h
```
The proto type is definted as `string json_metadata = 7; // base64 encoded`

If I go into a REPL, as a GPRC client, and make a call in a loop to fetch the metadata then decode, eventually parsing the metadata fails.

 **What did you expect to see**
No JSON::ParserError, no matter how many times the code is executed.

**What did you see instead?**
It will work for awhile in a loop, then eventually fail (at random tries) with an error:
```
 JSON.parse(decoded_metadata).to_h  # the line that throws a JSON::ParserError
 
 JSON::ParserError:  784: unexpected token at (some garbled JSON string sent from the server)
``` 
The loop will return the same string every time, but will eventually fail to parse the string and throw the JSON::ParserError.

The GRPC server is using a different of Ruby (v2.5.8) and protobuf (v3.9.0) than the Ruby client. 

If I downgrade the Ruby client's protobuf version to `3.14.0` this error goes away. All I need to do to reproduce this error is to upgrade the Ruby client to protobuf version `3.15.5`.

haberman:
If I'm understanding correctly, you are saying that the protobuf library is corrupting the `json_metadata` field, which then causes Ruby's `JSON.parse()` function to fail?

You said that the loop will return the same string every time, but parsing sporadically fails. Can you reproduce this if the loop does not use GRPC at all?

```ruby
payload = "..."

loop do
  proto = YourProto.decode(payload)
  decoded_metadata = Base64.decode64(proto.json_metadata)
  JSON.parse(decoded_metadata)  # Does this eventually fail?
end
```

If you can get that to sporadically fail, please send the payload and .proto definition and I can definitely debug that.

heyZeus:
@haberman The GRPC server always sends the same metadata string value to the client. That metadata value is:

```
[...]

PR #9219, opened 2021-11-13
[Ruby] Include "descriptor_pb.rb" well-known protobuf rb in repository

lfittl:
Since https://github.com/protocolbuffers/protobuf/pull/8850 there is a hard requirement for `descriptor_pb.rb` to exist in order for the gem to load. Unfortunately that file only gets created when running `rake build`.

An installation through bundler where this gem is sourced through the "git" source option won't call the Rakefile at all, causing a LoadError for google/protobuf/descriptor_pb. See also someone running into this problem when testing a different PR: https://github.com/protocolbuffers/protobuf/issues/8420#issuecomment-918558228

This change fixes this by including the `descriptor_pb.rb` file directly in the repository. It also adds the other well-known `lib/google/*_pb.rb` files to the `.gitignore` list to make it explicit these are not intended to be committed. Afterwards this gem can again be included via a git source, like this:

```
gem 'google-protobuf', github: 'lfittl/protobuf', branch: 'ruby-support-bundle-git-source', glob: 'ruby/*.gemspec'
```


acozzette (changes requested):
Thanks for the pull request! This looks good but I think we would need two additional changes before we can accept this:
1. Add the new file to `ruby_EXTRA_DIST` around [here](https://github.com/protocolbuffers/protobuf/blob/fe87de31bf4a7623bb4beb6e4890ac2d33a4a402/Makefile.am#L1164). This will fix the failing distcheck tests.
2. Update `generate_descriptor_proto.sh` to keep `descriptor_pb.rb` up to date. That script is responsible for updating other checked-in generated code files like this one.

acozzette:
@lfittl I'm going to go ahead and close this since it's been a few weeks with no activity, but feel free to comment here if you would like to reopen this.

Issue #9114, opened 2021-10-18
[JAVA]supporting nanos and second in timestamp merge


imtripathi:
**[JAVA]supporting nanos and second in timestamp merge**

In process of converting sink record (Avro serialized) to protobuf format using kafka connect 
we encountered the parsing issue in Timestamp data using the protobuf library

**Problem Description**
When we are trying to deserialize the Timestamp value from Avro format it  is getting converted to nanos and seconds which when we are trying to merge this converted  json object using  JsonFormat.parser().merge(json, builder)
of protobuf util it is throwing an exception .
Example -
```
GenericRecord record = genericRecord(timestamp);


System.out.println(JsonFormat.printer().print(timestamp));
System.out.println(record.toString());//This prints the record format //we receive of timestamp which we need to parse that is {"seconds": 10, "nanos": 100}

Timestamp.Builder builder = Timestamp.newBuilder();
JsonFormat.parser().merge(JsonFormat.printer().print(timestamp), builder);
System.out.println(builder.build());//this prints the record after parsing in new code seconds: 10 nanos: 100 it is as expected.

```
**Problem explanation**
When I am sending Timstamp in json format with the existing code in JsonFormat.parser().merge(json, builder)

Example –
{"seconds": 10, "nanos": 100}
I am getting below Exception –

```
Exception in thread "main" com.google.protobuf.InvalidProtocolBufferException: Failed to parse timestamp: {"seconds":10,"nanos":100}
	at com.proto.test.ParserImpl.mergeTimestamp(ParserImpl.java:300)
	at com.proto.test.ParserImpl.access$200(ParserImpl.java:23)
	at com.proto.test.ParserImpl$3.merge(ParserImpl.java:128)
	at com.proto.test.ParserImpl.merge(ParserImpl.java:189)
	at com.proto.test.ParserImpl.merge(ParserImpl.java:74)
	at com.proto.test.Test.main(Test.java:31)

```

 As observed JsonFormat.parser().merge(json, builder)
uses underlying mergeTimestamp method that only supports parsing data in  this format "1970-01-01T00:00:10.000000100Z"

Timestamp value = Timestamps.parse(json.getAsString());
**Solution--**
So to resolve this I have added the ability in this mergeTimestamp method code to check the instance of JsonElement  is JsonObject which  will fetch the value of nanos and second and will create the Time stamp value 
Timestamp.newBuilder().setSeconds(seconds).setNanos(nanos).build();
[...]

PR #9228, opened 2021-11-17
[JAVA]supporting nanos and second in timestamp merge

imtripathi:
Add lenient toggle so that those with any existing behavior today reliant on strict only would not break
Add test case for seconds/nanos
Note no existing Test cases altered ensuring no change in behavior for any existing cases.

Please Take this to Tier 1 or Tier to confirmation if possible .
issue Id -
https://github.com/protocolbuffers/protobuf/issues/9114

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -398,6 +398,7 @@ public class JsonFormat {
         com.google.protobuf.TypeRegistry.getEmptyTypeRegistry(),
         TypeRegistry.getEmptyTypeRegistry(),
         false,
+        false,
         Parser.DEFAULT_RECURSION_LIMIT);
   }
 
@@ -408,6 +409,7 @@ public class JsonFormat {
     private final com.google.protobuf.TypeRegistry registry;
     private final TypeRegistry oldRegistry;
     private final boolean ignoringUnknownFields;
+    private final boolean lenient;
     private final int recursionLimit;
 
     // The default parsing recursion limit is aligned with the proto binary parser.
@@ -417,10 +419,12 @@ public class JsonFormat {
         com.google.protobuf.TypeRegistry registry,
         TypeRegistry oldRegistry,
         boolean ignoreUnknownFields,
+        boolean lenient,
         int recursionLimit) {
       this.registry = registry;
       this.oldRegistry = oldRegistry;
       this.ignoringUnknownFields = ignoreUnknownFields;
+      this.lenient = lenient;
       this.recursionLimit = recursionLimit;
     }
 
@@ -438,6 +442,7 @@ public class JsonFormat {
       return new Parser(
           com.google.protobuf.TypeRegistry.getEmptyTypeRegistry(),
           oldRegistry,
+          lenient,
           ignoringUnknownFields,
           recursionLimit);
     }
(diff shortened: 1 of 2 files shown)


java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -453,17 +458,26 @@ public Parser usingTypeRegistry(com.google.protobuf.TypeRegistry registry) {
           || this.registry != com.google.protobuf.TypeRegistry.getEmptyTypeRegistry()) {
         throw new IllegalArgumentException("Only one registry is allowed.");
       }
-      return new Parser(registry, oldRegistry, ignoringUnknownFields, recursionLimit);
+      return new Parser(registry, oldRegistry, ignoringUnknownFields,lenient,recursionLimit);
```
elharo, line 461: space after comma, per google style

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -453,17 +458,26 @@ public Parser usingTypeRegistry(com.google.protobuf.TypeRegistry registry) {
...
     public Parser ignoringUnknownFields() {
-      return new Parser(this.registry, oldRegistry, true, recursionLimit);
+      return new Parser(this.registry, oldRegistry, true,lenient, recursionLimit);
+    }
+
+    /**
+     * Creates a new {@link Parser} configured to handle lenient parsing e.g. Timestamp represented as seconds and nanos instead of String format.
+     * The new Parser clones all other configurations from this Parser.
+     */
+    public Parser lenient() {
```
elharo, line 476: The name is not sufficiently descriptive. Lenient about what?

java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
```diff
@@ -408,6 +409,7 @@ public static Parser parser() {
     private final com.google.protobuf.TypeRegistry registry;
     private final TypeRegistry oldRegistry;
     private final boolean ignoringUnknownFields;
+    private final boolean lenient;
```
elharo, line 412: You've made this a boolean, but this is actually just two of an infinite possible number of formats it might want to parse.

elharo:
I don't think my opinion on this has changed. This repo and library is not the place to supply converters for arbitrary JSON formats. Rather proto supports a single imperfectly documented JSON format, mostly for purposes of debugging and human readability. 

Any production ingestion of JSON needs an intermediate step that understands the specific JSON format being converted and the specific protobuf schema being converted to.

imtripathi:
> I don't think my opinion on this has changed. This repo and library is not the place to supply converters for arbitrary JSON formats. Rather proto supports a single imperfectly documented JSON format, mostly for purposes of debugging and human readability.
> 
> Any production ingestion of JSON needs an intermediate step that understands the specific JSON format being converted and the specific protobuf schema being converted to.

@elharo  I was trying to extend this JsonFormat.class in my own  class and was trying to override the behaviour as per my need ,even we are not able to do that due to private constructor and private method. cant be this functionality be provide to us atleast it can be made protected so that we can extend it and .. override the functionality ,or we can create our plugin on top of it as per our behaviour.

imtripathi:
> Let me look into that possibility, but I suspect this would be better served by using composition (or even copy-paste; it is all open source) rather than inheritance. We need to be very cautious about expanding the API surface and our implicit promises to support that.

@elharo Any possibility for making the classes public or protected as discussed above.

PR #9233, opened 2021-11-19
Add a envvar to provide a prefix on all proto package prefixed symbols.

thomasvl:
While at it, fix a header comment.

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -102,9 +102,14 @@ class PrefixModeStorage {
 
   bool is_package_exempted(const std::string& package);
 
+  // When using a proto package as the prefix, this should be added as the
+  // prefix in front of it.
+  const std::string& forced_package_prefix() const { return forced_prefix_; }
+
  private:
   bool use_package_name_;
   std::string exception_path_;
+  std::string forced_prefix_;
   std::unordered_set<std::string> exceptions_;
 };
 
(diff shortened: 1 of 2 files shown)


src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -102,9 +102,14 @@ class PrefixModeStorage {
 
   bool is_package_exempted(const std::string& package);
 
+  // When using a proto package as the prefix, this should be added as the
+  // prefix in front of it.
+  const std::string forced_package_prefix() const { return forced_prefix_; }
```
fowles, line 107: const& if you are trying to avoid the copy, or no const if you want to have the copy

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -120,6 +125,13 @@ PrefixModeStorage::PrefixModeStorage() {
   if (exception_path) {
     exception_path_ = exception_path;
   }
+
+  // This one is a not expected to be common, so it doesn't get a generation
+  // option, just the env var.
+  const char* prefix = getenv("GPB_OBJC_USE_PACKAGE_AS_PREFIX_PREFIX");
```
fowles, line 131: PREFIX_PREFIX?

thomasvl: Yea, turning it back into words it make a little more sense: *GPB ObjC UsePackageAsPrefix Prefix*.  A few lines above it one using `GPB_OBJC_USE_PACKAGE_AS_PREFIX`, it conceptional is an addition to that: a prefix to use when that is enabled.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: Added.

PR #9234, opened 2021-11-19
Support no_package:FILE in the expected prefixes file.

thomasvl:
When a file doesn't have a proto package, allow validation to still be supported
via entries with that consist of 'no_package:FILE'. This allows someone using an
expected prefixes file to also list these exceptional cases, and allows them to
becomes error cases with requiring all prefixes to be registered.

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -352,9 +352,9 @@ bool IsReservedCIdentifier(const std::string& input) {
 }
 
 std::string SanitizeNameForObjC(const std::string& prefix,
-                           const std::string& input,
-                           const std::string& extension,
-                           std::string* out_suffix_added) {
+                                const std::string& input,
+                                const std::string& extension,
+                                std::string* out_suffix_added) {
   static const std::unordered_set<std::string> kReservedWords =
       MakeWordsMap(kReservedWordList, GOOGLE_ARRAYSIZE(kReservedWordList));
   static const std::unordered_set<std::string> kNSObjectMethods =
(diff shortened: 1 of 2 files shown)


src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -1286,35 +1292,12 @@ bool ValidateObjCClassPrefix(
...
-        // ... another package has declared the same prefix.
-        std::cerr
-             << "protoc:0: warning: File '" << file->name() << "' has no package "
-             << "and package '" << other_package_for_prefix << "' already uses '"
-             << prefix << "' as its prefix. Consider either adding a new package "
-             << "to the proto, or reusing one of the packages already using this "
-             << "prefix in the expected prefixes file ("
-             << expected_prefixes_path << ")." << std::endl;
-        std::cerr.flush();
+        // Stop on the first real package listing, it if was a no_package file
```
dmaclach, line 1295: if it was a

dmaclach (approved):

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -1240,14 +1240,17 @@ bool ValidateObjCClassPrefix(
 
   const std::string prefix = file->options().objc_class_prefix();
   const std::string package = file->package();
+  static const std::string no_package_prefix("no_package:");
+  const std::string lookup_key =
```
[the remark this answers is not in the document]
thomasvl: Done, also updated the docs for the generator options.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: done

PR #9235, opened 2021-11-19
Change sysroot flags

deannagarcia:
Take away the sysroot flag for the compiler as it isn't necessary, but ensure the sysroot flag is at the top level directory for linker flag.


perezd (approved):
LGTM!

PR #9248, opened 2021-11-24
[C#] added new csharp options for protoc

recih:
## `custom_base_class`
Custom base class to use for generated classes. Defaults to `""`.

If specified, the base class must be a public class in the current namespace.

There is a special support for generic classes, which can be specified by `"MyBaseClass<>"`, `"<>"` is a placeholder, which will be replaced by the current message class name.

For example, for a message named `"MyMessage"`, if `"MyBaseClass<>"` is specified, the base class will be `"MyBaseClass<MyMessage>"`.

## `keep_original_file_name`
Whether to keep the original file name. Defaults to `false`.
If `true`, the CamelCase name transformation is skiped, and the original file name of proto file is preserved

## `keep_original_field_name`
Whether to keep the original field name. Defaults to `false`.
If `true`, the CamelCase name transformation is skiped, and the original field name defined in proto file is preserved

## `disable_nested_types_container`
Whether to generate a nested `Types` container class for sub messages.
If true, the nested `Types` container class will not be generated. Be ware, name collision may occurs.

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_generator.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_generator.cc
@@ -80,11 +80,21 @@ bool Generator::Generate(const FileDescriptor* file,
       cli_options.internal_access = true;
     } else if (options[i].first == "serializable") {
       cli_options.serializable = true;
+    } else if (options[i].first == "custom_base_class") {
+      cli_options.custom_base_class = options[i].second;
+    } else if (options[i].first == "keep_original_file_name") {
+      cli_options.keep_original_file_name = true;
+    } else if (options[i].first == "keep_original_field_name") {
+      cli_options.keep_original_field_name = true;
+    } else if (options[i].first == "disable_nested_types_container") {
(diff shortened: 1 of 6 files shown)


elharo (commented):
Interesting but I'm really not sure if we'd be willing to commit to any of these. This probably needs a detailed proposal and justification for adding these options.

elharo (changes requested):
These features don't appear to have any tests. That's mandatory for considering anything of this scope.

recih:
> These features don't appear to have any tests. That's mandatory for considering anything of this scope.

I'd like to add some tests, but I didn't find any test code for existed c# options, like `internal_access`, `base_namespace`, etc. I'm not very clear how to test these options and need some help.

acozzette:
@recih Sorry but unfortunately I don't think we will be accepting this pull request. We generally try hard to avoid adding any knobs for customization if at all possible, because they tend to greatly increase complexity and make it more difficult to make future changes.

PR #9250, opened 2021-11-24
Adding windows toolchains

deannagarcia:
Adding windows toolchains to the BUILD file and adding some new fields (one to hold extra includes and one for another linker flag) in order to be able to make the windows toolchains work.


perezd (approved):
LGTM!

PR #9263, opened 2021-11-29
[cleanup] Msvc warnings

stephenatwork:
Removes the need for disabling two warnings. Context: Our org considers it best practice to always enable
 4018;4146;4244;4267;4996. The last 3 relate to type conversions and deprecation - they'll require some more effort to root out.

First, by fixing the offending code for
/wd4018 # 'expression' : signed/unsigned mismatch

And second by moving this warning down to the place it is needed (tests)
/wd4146 # unary minus operator applied to unsigned type, result still unsigned

### DIFF
--- a/src/google/protobuf/io/io_win32_unittest.cc
+++ b/src/google/protobuf/io/io_win32_unittest.cc
@@ -164,8 +164,8 @@ void IoWin32Test::SetUp() {
   test_tmpdir.clear();
   wtest_tmpdir.clear();
   DWORD size = ::GetCurrentDirectoryW(MAX_PATH, working_directory);
-  EXPECT_GT(size, 0);
-  EXPECT_LT(size, MAX_PATH);
+  EXPECT_GT(size, 0U);
+  EXPECT_LT(size, static_cast<DWORD>(MAX_PATH));
 
   string tmp;
   bool ok = false;
@@ -581,7 +581,7 @@ TEST_F(IoWin32Test, ExpandWildcardsFailsIfNoFileMatchesTest) {
 TEST_F(IoWin32Test, AsWindowsPathTest) {
   DWORD size = GetCurrentDirectoryW(0, nullptr);
   std::unique_ptr<wchar_t[]> cwd_str(new wchar_t[size]);
-  EXPECT_GT(GetCurrentDirectoryW(size, cwd_str.get()), 0);
+  EXPECT_GT(GetCurrentDirectoryW(size, cwd_str.get()), 0U);
   wstring cwd = wstring(L"\\\\?\\") + cwd_str.get();
 
   ASSERT_EQ(testonly_utf8_to_winpath("relative_mkdirtest"),
--- a/BUILD
+++ b/BUILD
@@ -28,7 +28,6 @@ ZLIB_DEPS = ["@zlib//:zlib"]
 ################################################################################
 
 MSVC_COPTS = [
-    "/wd4018",  # -Wno-sign-compare
     "/wd4065",  # switch statement contains 'default' but no 'case' labels
     "/wd4146",  # unary minus operator applied to unsigned type, result still unsigned
     "/wd4244",  # 'conversion' conversion from 'type1' to 'type2', possible loss of data
--- a/BUILD
+++ b/BUILD
@@ -29,7 +29,6 @@ ZLIB_DEPS = ["@zlib//:zlib"]
 
 MSVC_COPTS = [
     "/wd4065",  # switch statement contains 'default' but no 'case' labels
-    "/wd4146",  # unary minus operator applied to unsigned type, result still unsigned
     "/wd4244",  # 'conversion' conversion from 'type1' to 'type2', possible loss of data
     "/wd4251",  # 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
     "/wd4267",  # 'var' : conversion from 'size_t' to 'type', possible loss of data
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -211,7 +211,6 @@ if (MSVC)
   # MSVC warning suppressions
   add_definitions(
     /wd4065 # switch statement contains 'default' but no 'case' labels
-    /wd4146 # unary minus operator applied to unsigned type, result still unsigned
     /wd4244 # 'conversion' conversion from 'type1' to 'type2', possible loss of data
     /wd4251 # 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
     /wd4267 # 'var' : conversion from 'size_t' to 'type', possible loss of data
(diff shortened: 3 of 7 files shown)


src/google/protobuf/io/io_win32_unittest.cc
```diff
@@ -164,8 +164,8 @@ void IoWin32Test::SetUp() {
   test_tmpdir.clear();
   wtest_tmpdir.clear();
   DWORD size = ::GetCurrentDirectoryW(MAX_PATH, working_directory);
-  EXPECT_GT(size, 0);
-  EXPECT_LT(size, MAX_PATH);
+  EXPECT_GT(size, 0U);
+  EXPECT_LT(size, (DWORD)MAX_PATH);
```
elharo, line 168: per google C++ style guide "Use C++-style casts like static_cast<float>(double_value), or brace initialization for conversion of arithmetic types like int64_t y = int64_t{1} << 42. Do not use cast formats like (int)x unless the cast is to void. You may use cast formats like `T(x)` only when `T` is a class type."

stephenatwork: Thanks, updated

stephenatwork:
FWIW I rebased to make the history cleaner.

acozzette:
Thanks, @stephenatwork.

tusharb86:
Hi there - I'm looking to continue the work to remove the need to disable some of the remaining MSVC warnings that @stephenatwork mentioned originally: 4244;4267;4996.

While investigating this work, I noticed that the [`/WX` flag](https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level) is not enabled, meaning that warnings are *not* being treated as errors currently. Therefore, any new regressions would be missed by the build. Ideally - we would enable the flag so warnings are elevated to errors.

Before moving forward with the work, I wanted to check with @acozzette and @elharo to see if there is a desire to treat warnings as errors with the MSVC toolchain, specifically? Further, is it reasonable to assume that it is OK to fix the existing warnings related to type conversions (e.g. `uint64_t` to `int`) and use of deprecated items:
- [4244](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244): 'conversion' conversion from 'type1' to 'type2', possible loss of data
- [4267](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267): 'var' : conversion from 'size_t' to 'type', possible loss of data
- [4996](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996): use of deprecated item

acozzette:
That would be great if you want to send us changes to treat warnings as errors and fix up the code to make that possible. There are two things that you might want to watch out for, though:
- I wouldn't be surprised if some of our tests deliberately call deprecated functions to exercise protobuf fields that are marked `deprecated = true`. So it might be tricky to make that particular warning into an error.
- The Google style guide takes a somewhat controversial position on unsigned integers: https://google.github.io/styleguide/cppguide.html#Integer_Types Basically we tend to use signed types even if we expect the value to always be non-negative. This way the compiler can sometimes generate better code (can assume overflow will never happen) and sanitizers like UBSAN can detect bugs in the case where overflow does occur. I'm not sure if this will matter but it might make a difference depending on what kinds of warnings come up.

tusharb86:
Thanks for the quick response @acozzette as well as the two things to look out for.

For the first one, I'll likely move the warning suppression 4996 related to usage of deprecated items down to the test code, similar to what @stephenatwork did for 4146.

For the second one, this will involve adding lots of explicit `static_cast`s to the codebase. Some examples of the warnings and their respective fixes are below. If these seem like acceptable changes, let me know and I'll move forward with the changes across the codebase. Thanks for your consideration!

```
coded_stream.cc(484): warning C4244: 'return': conversion from '_Ty1' to 'int', possible loss of data

@@ -481,7 +481,7 @@ int CodedInputStream::ReadVarintSizeAsIntSlow() {
   // for one-byte varints.
   std::pair<uint64_t, bool> p = ReadVarint64Fallback();
   if (!p.second || p.first > static_cast<uint64_t>(INT_MAX)) return -1;
-  return p.first;
+  return static_cast<int>(p.first);
 }
 
 int CodedInputStream::ReadVarintSizeAsIntFallback() {

---

coded_stream.cc(496): warning C4244: 'return': conversion from 'uint64_t' to 'int', possible loss of data

@@ -493,7 +493,7 @@ int CodedInputStream::ReadVarintSizeAsIntFallback() {
     ::std::pair<bool, const uint8_t*> p = ReadVarint64FromArray(buffer_, &temp);
     if (!p.first || temp > static_cast<uint64_t>(INT_MAX)) return -1;
     buffer_ = p.second;
-    return temp;
+    return static_cast<int>(temp);
   } else {
     // Really slow case: we will incur the cost of an extra function call here,
     // but moving this out of line reduces the size of this function, which

---

coded_stream.cc(675): warning C4244: 'initializing': conversion from '__int64' to 'int', possible loss of data

@@ -672,7 +672,7 @@ void EpsCopyOutputStream::EnableAliasing(bool enabled) {
 
 int64_t EpsCopyOutputStream::ByteCount(uint8_t* ptr) const {
   // Calculate the current offset relative to the end of the stream buffer.
-  int delta = (end_ - ptr) + (buffer_end_ ? 0 : kSlopBytes);
+  int delta = static_cast<int>((end_ - ptr) + (buffer_end_ ? 0 : kSlopBytes));
   return stream_->ByteCount() - delta;
 }
 
```

acozzette:
@tusharb86 Those changes look good to me!

tusharb86:
@acozzette While fixing up many of the conversions/casts in the codebase, I noticed the following functions in `message_lite.h`. Should I be using these instead of plain static_casts across the codebase? If so, is there a more common place I could move these functions?

```
// We mainly calculate sizes in terms of size_t, but some functions that
// compute sizes return "int".  These int sizes are expected to always be
// positive. This function is more efficient than casting an int to size_t
// directly on 64-bit platforms because it avoids making the compiler emit a
// sign extending instruction, which we don't want and don't want to pay for.
inline size_t FromIntSize(int size) {
  // Convert to unsigned before widening so sign extension is not necessary.
  return static_cast<unsigned int>(size);
}

// For cases where a legacy function returns an integer size.  We GOOGLE_DCHECK()
// that the conversion will fit within an integer; if this is false then we
// are losing information.
inline int ToIntSize(size_t size) {
  GOOGLE_DCHECK_LE(size, static_cast<size_t>(INT_MAX));
  return static_cast<int>(size);
}
```

acozzette:
I think it would probably be good to use those helper functions where they make sense. Do you have an example of a place where they would be useful but where message_lite.h isn't available (e.g. because it would cause a header cycle)?

tusharb86:
Got it - thanks! I'm close to finishing up all the changes. I'll look at what it would take to switch to using these helpers and whether there are any cases where they can't be used. I may send out a PR without using these first and we can discuss where it would make sense to use the helpers in that PR.

tusharb86:
For those following, a PR on the type conversions has been opened #10042.

PR #9267, opened 2021-11-29
No error throwing for unknown OS

deannagarcia:
Remove error throwing if we can't tell what OS we are trying to run for and just name the artifact something with "unknown". This is mainly to fix the case without a --config where the cpu defaults to local and the script can't tell what OS we are using.


perezd (approved):
LGTM!

PR #9273, opened 2021-12-02
Fix broken max macro on Windows

Kerisa:
Hi there,

I found that if I include the auto generated XX.pb.h in my code then the "max" macro defined by Windows will be broken.

And this is caused by a direct undef in `src/google/protobuf/arena.h`
```cpp
#ifdef max
#undef max  // Visual Studio defines this macro
#endif
```

I think this should be done by using `port_def.inc` and `port_undef.inc` pairs so I removed this undef and add a protect on `src/google/protobuf/stubs/stl_util.h`.

Thank you

### DIFF
--- a/src/google/protobuf/arena.h
+++ b/src/google/protobuf/arena.h
@@ -37,9 +37,6 @@
 #include <limits>
 #include <type_traits>
 #include <utility>
-#ifdef max
-#undef max  // Visual Studio defines this macro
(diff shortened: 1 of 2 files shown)


Kerisa:
I have sign the Google Individual CLA agreement just now, is there any one who can help to trigger a rescan? Thanks.

PR #9276, opened 2021-12-03
Fixing windows builds

deannagarcia:
Fix windows builds by:
- Creating an artifact pattern to match protoc even with an .exe extension
- Changing version of mingw libraries 
- Adding a compiler flag to use fsjlj exceptions to match mingw library exceptions for win32
- Use -pthread for win32


perezd (approved):
LGTM!

PR #9277, opened 2021-12-03
Change include directories

deannagarcia:
Make /usr/include an extra included directory so that windows builds don't have access to /usr/include/stdlib.h and will instead need to use their custom stdlib.


perezd (approved):
LGTM!

PR #9298, opened 2021-12-10
remove unused test proto

elharo:
@perezd


perezd:
This appears to be breaking Maven, compilation targets looking for `TestMap`

elharo:
Yes, and I don't understand why. I need to figure it out.

elharo:
Latest changes appear to have fixed this though the ruby CI is still borked for unrelated reasons, and several of the other builds seem to be having timeout or network issues. I'll revisit once #9303 is resolved.

PR #9309, opened 2021-12-14
show less log junk from Maven

elharo:
@perezd turn off debugging messages such as

```
[DEBUG]   Imported: org.codehaus.plexus.configuration < plexus.core
```


perezd:
LGTM, is there anything more you can silence? I really only want failures.

perezd (approved):

elharo:
Yes, there's more we can silence but we'll need Maven 3.6.1 first. I'm looking into it.

PR #9312, opened 2021-12-15
Fix protobuf_MSVC_STATIC_RUNTIME flag for cmake 3.15.0+

FedyuninV:
Fix for the following config:
- MSVC
- cmake 3.15.0+
- static protobuf
- dynamic runtime (`protobuf_MSVC_STATIC_RUNTIME=OFF`)

Previous behavior:
Ignore `protobuf_MSVC_STATIC_RUNTIME` value and set runtime to static.

New behavior:
Set static runtime if and only if `protobuf_MSVC_STATIC_RUNTIME=ON`.

Signed-off-by: Valeriy Fedyunin <valery.fedyunin@abbyy.com>


FedyuninV:
I'm not a contributor, that's why I can't add reviewers nor labels :(

cmake/CMakeLists.txt
```diff
@@ -177,27 +177,29 @@ if (protobuf_BUILD_SHARED_LIBS)
...
-    if (MSVC AND protobuf_MSVC_STATIC_RUNTIME)
-      foreach(flag_var
-          CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-          CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-        if(${flag_var} MATCHES "/MD")
-          string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-        endif(${flag_var} MATCHES "/MD")
-      endforeach(flag_var)
-    endif (MSVC AND protobuf_MSVC_STATIC_RUNTIME)
+  if(protobuf_MSVC_STATIC_RUNTIME)
```
elharo, line 180: how does this variable affect other platforms?

FedyuninV: It doesn't affect at all. All of the flags here are MSVC-only.

E.g. from `CMAKE_MSVC_RUNTIME_LIBRARY` docs:
```
The value is ignored on non-MSVC compilers but an unsupported value will be rejected as an error when using a compiler targeting the MSVC ABI.
```

FedyuninV: I suppose the reason of tests failing is missing fix from master for ruby. I added them, but [current master](https://github.com/protocolbuffers/protobuf/commit/de5eb0a3e3c0bf6f9fd6b4fa896178b23f4cc020) still fails on Kokoro.

FedyuninV: Can you run tests one more time plz? Master build has been stabilized and I've added the latest master to the branch.

deannagarcia:
Can you go through the process to sign a CLA? Unfortunately we cannot accept pull requests without it. If you get that done, I would be happy to run the tests for you!

FedyuninV:
@deannagarcia Wow, I've never met CLA before! Fixed.

FedyuninV:
Rebase didn't work as intended( I will reopen it one more time

FedyuninV:
I've reopened it as #9337

chronoxor:
Please also see my fix for the issue - https://github.com/protocolbuffers/protobuf/pull/9348

PR #9318, opened 2021-12-16
Change toolchain includes

deannagarcia:
Change toolchain includes to use -isystem instead of -I

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -129,7 +129,7 @@ def _impl(ctx):
               actions = all_compile_actions,
               flag_groups = [
                   flag_group(
-                      flags = [
+                      flags = ctx.attr.extra_compiler_flags + [
                           ctx.attr.bit_flag,
                           "-Wall",
                           "-no-canonical-prefixes",
@@ -137,7 +137,7 @@ def _impl(ctx):
                           "-isystem",
                           ctx.attr.toolchain_dir,
                           "-fvisibility=hidden",
-                      ] + ctx.attr.extra_compiler_flags,
+                      ],
                   ),
               ],
           ),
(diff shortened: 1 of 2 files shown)


dlj-NaN:
I think this flag is still going to be problematic:
https://github.com/protocolbuffers/protobuf/blob/master/toolchain/cc_toolchain_config.bzl#L137-L138

deannagarcia:
> I think this flag is still going to be problematic: https://github.com/protocolbuffers/protobuf/blob/master/toolchain/cc_toolchain_config.bzl#L137-L138

Why? I took out the repeat of that `-isystem` and all the other toolchains need it.

dlj-NaN:
> > I think this flag is still going to be problematic: https://github.com/protocolbuffers/protobuf/blob/master/toolchain/cc_toolchain_config.bzl#L137-L138
> 
> Why? I took out the repeat of that `-isystem` and all the other toolchains need it.

Placing it at the start means it won't be found by the subsequent `#include_next`. The "next" has to fall through to the end, but if this is first, then it can't do that. (That means the `-I` to `-isystem` change is actually doing something else now that the duplication is gone, but it's the right thing to do in any case.)

toolchain/cc_toolchain_config.bzl
```diff
@@ -129,15 +129,15 @@ def _impl(ctx):
               actions = all_compile_actions,
               flag_groups = [
                   flag_group(
-                      flags = [
+                      flags = ctx.attr.extra_compiler_flags + [
```
deannagarcia, line 132: @dlj-NaN Does this fix the ordering problem?

dlj-NaN: Yeah, that should solve the include path search issue.

We might still want the extra_compiler_flags to be ordered after the the warning/functional defaults (`-Wall` and `-fvisibility`), though. That way, we can control warnings/visibility via `extra_compiler_flags` as well as include path injection. It might make sense just to split into three flag_groups:

1. everything else,
2. `ctx.attr.extra_compiler_flags`, and
3. the toolchain include path.

deannagarcia: Ok I think I fixed it, can you take another look?

dlj-NaN (approved):

PR #9337, opened 2021-12-22
Fix protobuf_MSVC_STATIC_RUNTIME flag for cmake 3.15.0+

FedyuninV:
Fix for the following config:
- MSVC
- cmake 3.15.0+
- static protobuf
- dynamic runtime (`protobuf_MSVC_STATIC_RUNTIME=OFF`)

Previous behavior:
Ignore `protobuf_MSVC_STATIC_RUNTIME` value and set runtime to static.

New behavior:
Set static runtime if and only if `protobuf_MSVC_STATIC_RUNTIME=ON`.

(it's the fixed version of #9312 )

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -177,27 +177,29 @@ if (protobuf_BUILD_SHARED_LIBS)
   set(protobuf_SHARED_OR_STATIC "SHARED")
 else (protobuf_BUILD_SHARED_LIBS)
   set(protobuf_SHARED_OR_STATIC "STATIC")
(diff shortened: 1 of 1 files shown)


FedyuninV:
I can't add labels myself. It's CMake Windows-only problem.

chronoxor:
https://github.com/protocolbuffers/protobuf/pull/9348

acozzette:
Let me go ahead and close this because it sounds like it should be fixed by #9437. Please let me know if I'm mistaken and we can reopen it, though.

PR #9345, opened 2021-12-26
avoid allocating memory if all extension are cleared

liuzhijiang:
avoid allocating memory if all extension are cleared

### DIFF
--- a/src/google/protobuf/extension_set.cc
+++ b/src/google/protobuf/extension_set.cc
@@ -971,6 +971,26 @@ size_t SizeOfUnion(ItX it_xs, ItX end_xs, ItY it_ys, ItY end_ys) {
 
 void ExtensionSet::MergeFrom(const MessageLite* extendee,
                              const ExtensionSet& other) {
+  // check if the other is cleared
+  bool is_other_all_cleared = true;
+  if (PROTOBUF_PREDICT_TRUE(!other.is_large())){
+    for (const KeyValue* it = other.flat_begin(); it != other.flat_end(); it++){
+      is_other_all_cleared = is_other_all_cleared && it->second.is_cleared;
+      if (!is_other_all_cleared){
+        break;
+      }
+    }
+  } else {
+    for (LargeMap::iterator it = other.map_.large->begin(); it != other.map_.large->end(); it++){
+      is_other_all_cleared = is_other_all_cleared && it->second.is_cleared;
+      if (!is_other_all_cleared){
+        break;
+      }
+    }
+  }
+  if (is_other_all_cleared){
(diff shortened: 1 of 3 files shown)


sbenzaquen:
We should test that the capacity grows as expected with the new changes for `is_cleared`.

src/google/protobuf/extension_set.cc
```diff
@@ -1950,7 +1970,7 @@ void ExtensionSet::GrowCapacity(size_t minimum_new_capacity) {
 
   auto new_flat_capacity = flat_capacity_;
   do {
-    new_flat_capacity = new_flat_capacity == 0 ? 1 : new_flat_capacity * 4;
```
acozzette, line 1953: I think we should stick with the 4x growth, because as I understand this was a deliberate choice to optimize performance compared to the usual 2x growth.

acozzette (changes requested):

src/google/protobuf/extension_set.cc
liuzhijiang: understood.

acozzette:
@liuzhijiang This change looks good to me but I think we just need a unit test as @sbenzaquen suggested to verify the new behavior. I think the test could just look at `SpaceUsedExcludingSelfLong()` and verify that this number doesn't go up when we merge an ExtensionSet that has been cleared.

liuzhijiang:
The function Reflection::GetExtensionSet() is private, so the object ExtensionSet cannot be obtained and the function ExtensionSet::SpaceUsedExcludingSelfLong() cannot be called. Is there any other way to call the function SpaceUsedExcludingSelfLong() ?
If there is no other way to call this function, is the following unit test OK? 

`TEST(ExtensionSetTest, ExtensionSetSpaceUsed) {
  unittest::TestAllExtensions msg;
  long l = msg.SpaceUsedLong();
  msg.SetExtension(unittest::optional_int32_extension, 100);
  unittest::TestAllExtensions msg2(msg);
  long l2 = msg2.SpaceUsedLong();
  msg.ClearExtension(unittest::optional_int32_extension);
  unittest::TestAllExtensions msg3(msg);
  long l3 = msg3.SpaceUsedLong();
  EXPECT_TRUE((l2 - l) > (l3 - l));
}`

acozzette:
@liuzhijiang That kind of unit test looks good to me. I think you are right that the best way to do it is to rely on `Message::SpaceUsedLong()` since `ExtensionSet::SpaceUsedLong()` is not easily visible for testing.

acozzette:
@liuzhijiang Thanks!

PR #9356, opened 2021-12-30
Updates TimeUtils::GetCurrentTime() to provide nanosecond resolution

### LINKED ISSUE #8930: TimeUtils::GetCurrentTime() and fractions of second
Version: 3.16
Language: C++
Linux
GCC 7.2

I do "message.mutable_timestamp->CopyFrom(util::TimeUtil::GetCurrentTime());"

I expect resulting timestamp field to contain current time up to milliseconds at least because Linux has gettimeofday() supporting milliseconds.
I see the time rounded to whole seconds.

In protobuf sources I see
```
void GetCurrentTime(int64* seconds, int32* nanos) {
  // TODO(xiaofeng): Improve the accuracy of this implementation (or just
  // remove this method from protobuf).
  *seconds = time(nullptr);
  *nanos = 0;
}
```

Dig-Doug:
Updates TimeUtils::GetCurrentTime() to provide nanosecond resolution for c++11

Fixes #8930.

### DIFF
--- a/src/google/protobuf/stubs/time.cc
+++ b/src/google/protobuf/stubs/time.cc
@@ -1,6 +1,6 @@
 #include <google/protobuf/stubs/time.h>
 
-#include <ctime>
+#include <chrono>
 
 #include <google/protobuf/stubs/stringprintf.h>
 #include <google/protobuf/stubs/strutil.h>
@@ -262,10 +262,10 @@ bool DateTimeToSeconds(const DateTime& time, int64* seconds) {
 }
 
 void GetCurrentTime(int64* seconds, int32* nanos) {
-  // TODO(xiaofeng): Improve the accuracy of this implementation (or just
-  // remove this method from protobuf).
-  *seconds = time(nullptr);
-  *nanos = 0;
+  auto now = std::chrono::high_resolution_clock::now().time_since_epoch();
+  auto secs = std::chrono::duration_cast<std::chrono::seconds>(now);
+  *seconds = secs.count();
+  *nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(now - secs).count();
 }
 
 std::string FormatTime(int64 seconds, int32 nanos) {


src/google/protobuf/stubs/time.cc
```diff
@@ -1,6 +1,10 @@
 #include <google/protobuf/stubs/time.h>
 
+#if __cplusplus > 199711L
```
acozzette, line 3: I don't think there is any need for this conditional (and the one below) because starting from version 3.5.0 we require C++11 or newer.

Dig-Doug: Removed.

src/google/protobuf/stubs/time.cc
```diff
@@ -262,10 +262,10 @@ bool DateTimeToSeconds(const DateTime& time, int64* seconds) {
...
 
 void GetCurrentTime(int64* seconds, int32* nanos) {
-  // TODO(xiaofeng): Improve the accuracy of this implementation (or just
-  // remove this method from protobuf).
-  *seconds = time(nullptr);
-  *nanos = 0;
+  auto now = std::chrono::high_resolution_clock::now().time_since_epoch();
+  auto secs = std::chrono::duration_cast<std::chrono::seconds>(now);
+  *seconds = secs.count();
+  *nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(now - secs).count();
```
treehaqr, line 268: Somehow this returns wildly different values for me. It works on GCC with x86, but when compiled on Clang for Apple Silicon it shows something wildly different.
TimeUtil::TimestampToMilliseconds(TimeUtil::GetCurrentTime());
Clang (Apple Si) = 573427268 ms
GCC (X86) = 1642781354185 ms (Received over GRPC)

The delta should be < 100 ms, but it seems pretty big

acozzette:
@Dig-Doug Thank you for the pull request but I'm afraid that we probably don't want to merge this change. I asked @jorgbrown and @sbenzaquen about this and they pointed out two interesting things:
1. Before C++20, `system_clock` is not guaranteed to be based on the usual 1970 Unix epoch: https://en.cppreference.com/w/cpp/chrono/system_clock
2. `high_resolution_clock` is not even guaranteed to be `system_clock` (Clang for example seems to make it use `steady_clock` instead), so if I understand right it could be almost any monotonically increasing clock. https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

It seems to me that either of those two things could explain the behavior that @treehaqr ran into, so more work would be required to figure out an implementation that does the right thing in a portable way. However, we are planning to start using Abseil soon, and Abseil provides its own [time](https://abseil.io/docs/cpp/guides/time) library that should make this easy. That makes me think the best thing to do is hold off on this until we can just rely on Abseil to get a more accurate current time.

Issue #9293, opened 2021-12-09
PHP: protobuf 3.19.1 causes several Deprecation notices


caugner:
**What version of protobuf and what language are you using?**
Version: 3.19.1
Language: PHP

**What operating system (Linux, Windows, ...) and version?**

Docker 4.3.0 inside Ubuntu 20.04 LTS inside WSL2 on Windows 10

**What runtime / compiler are you using (e.g., python version or gcc version)**

```dockerfile
# Dockerfile
FROM php:8.1-cli-alpine

RUN apk add --no-cache git && \
    curl -sSL -o /usr/local/bin/install-php-extensions https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions && \
    chmod uga+x /usr/local/bin/install-php-extensions && \
    sync && \
    install-php-extensions protobuf && \
    rm /usr/local/bin/install-php-extensions
```

**What did you do?**
1. Build the Dockerfile (`docker build -t php-with-protobuf .`)
2. Run the built Docker image (`docker run --rm -it php-with-protobuf`)

**What did you expect to see**

```
$ docker build -t php-with-protobuf .
$ docker run --rm -it php-with-protobuf
Interactive shell

php >
```

**What did you see instead?**

```
$ docker build -t php-with-protobuf .
$ docker run --rm -it php-with-protobuf
Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetExists($index) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0                 
                                                                                                                                                                                                                                                                                                          
Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetGet($index) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0                      
                                                                                                                                                                                                                                                                                                          
Deprecated: Return type of Google\Protobuf\Internal\RepeatedField::offsetSet($index, $newval) should either be compatible with ArrayAccess::offsetSet(mixed $offset, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Unknown on line 0
                                                                                                                                                                                                                                                                                                          
[...]

PR #9359, opened 2022-01-01
Fix PHP 8.1 compatibility

Yurunsoft:
Fix Deprecated such as:
```
Deprecated: Return type of Google\Protobuf\Internal\RepeatedFieldIter::current() should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in
```


caugner:
Fixes #9293.

Dragomeat:
Does anyone know when this pr will be merged and released?

caugner:
@Dragomeat It looks like #9370 is more promising, since the tests are already passing.

haberman:
Agreed, I am closing this in favor of #9370.

PR #9366, opened 2022-01-04
Update GSON to 2.8.9

pallavsingh:
Security Issue: Updating gson version to 2.8.9


elharo (changes requested):
There are some other locations this change also needs to be made.

elharo:
I'm going to close this in favor of a slightly more complete cl, though thanks for bringing ti to my attention.

PR #9370, opened 2022-01-05
fixing php 8.1 deprecation warnings

brettmc:
php 8.1 is more strict, and raises some deprecation notices with existing protobuf code. Not all of the
deprecations can be fixed without dropping support for php7.x (eg, the 'mixed' type doesn't appear until
8.1, and union types until 8.0, but as an interim solution the 'ReturnTypeWillChange' attribute can be
used to suppress the notices. In passing, also be explicit about a cast from float to int in 'zigZagEncode64'
which 8.1 also complains about when running tests.


brettmc:
have signed CLA now. I can't add labels, but I think `php` and `release notes: no`.
Also, this is a duplicate of #9359 but this PR goes a little further and adds some more type hints etc

zeriyoshi:
Looks good.
I think that the PHP Extension needs to support these as well.
If there are no problems with the PHP Library, can you the PHP Extension be modified in the same way?

brettmc:
@zeriyoshi - part one is done (adding return type hints), now I just need to work out how to add annotations

zeriyoshi:
@brettmc 
Very good job!

However, it seems that the minimum supported version of PHP has been changed to 7.1 or higher.
This needs to be mentioned in the release notes and Google needs to agree to this.

Perhaps You can lower the minimum required version of PHP, is that possible?

MrMage:
For what it's worth, PHP 7.0 seems to be used on less than 1% of installs these days, anyway, so an upgrade shouldn't be too risky: https://stitcher.io/blog/php-version-stats-january-2022

Given that protobuf is a but esoterical in the PHP community, I'd suspect that most users of this library are actually using even more recent versions, anyway.

zeriyoshi:
That's right. but, Google still continues to support PHP 5.5 in GAE.

https://cloud.google.com/appengine/docs/standard/php

The owner has the right to decide the lifetime of the software. It is up to Google to decide this.

In either case, I think it is essential to include it in the release notes.


Personally, I think that the PHP Library and Extension should drop support for PHP 7.0.

brettmc:
The reason for php 7.1 is because that's where the `void` return type was introduced (I agree that whether that's acceptable would be the project owner's decision). It might be possible to make the C extension work with both 8.1 and 7.1 (by conditionally compiling in the void return type), but I don't think that's true of the library.
I'm still struggling along with how to make the extension work in 8.1 whilst still being backwards-compatible, but I've just asked on the php-internals list for guidance.
As far at 8.1+ is concerned, it might be ok for users to ignore the deprecation warnings (in the project I'm working on, they cause tests to fail but that's a phpunit feature we could possibly disable). But, in php 9.0 they will all be a hard fail, so whenever that comes out, it will definitely not work without dropping support for <= 8.1

Anyway, still work-in-progress, so I'll convert this to a draft while I continue to find a fix for the extension.

brettmc:
OK all, thanks for your patience. I think I have this working now, with help from the php-internals list. I've run the tests against 8.1, 7.4 and 7.2 with/without the extension, and all is green. I couldn't get the extension to be compatible with 7.1, so now the supported versions are:
library: 7.1+
extension: 7.2+

How do we seek approval to drop 7.0/7.1 support? (and, could somebody please add `release notes: yes`)

zeriyoshi:
Perhaps @haberman  knows about it?

haberman:
Thanks all for your work on this PR.

When considering what versions of PHP to support, we have to coordinate with other projects, particularly gRPC and Google Cloud APIs.

gRPC appears to still be supporting 7.0+ still: https://grpc.io/docs/languages/php/quickstart/

If we were going to drop 7.0 we would need to coordinate with gRPC.

brettmc:
Asked on https://groups.google.com/g/grpc-io about my proposed changes.

haberman:
It looks like there are some broken tests in the CI related to PHP 7.0.

Isn't `ReturnTypeWillChange` compatible with 7.0? I'm not seeing why we need to drop 7.0 support.

I agree we will want to drop 7.0 before long, but if we can make this PR support it we could avoid blocking the PR on that decision.

brettmc:
@haberman I've updated for 7.0 support (by exchanging void return types for `ReturnTypeWillChange` annotations). The php lib should work for 7.0 now, but needs somebody to kick off the workflow.

brettmc:
That moved the php7.0 error along a little bit...the tests passed for the library, but not the extension. I have so far been unable to make 8.1 changes that are BC to 7.0, the best I have done is 7.2. At this point, library is 7.0+ and extension is 7.2+

haberman:
@brettmc what is the issue with supporting 7.0 in the extension?

brettmc:
@haberman I'm struggling to handle this macro signature change between 7.1 and 7.2:
```
#7.1
#define ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(name, return_reference, required_num_args, type, class_name, allow_null)
#7.2+
#define ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(name, return_reference, required_num_args, type, allow_null)
```
I'm sure there's come combination of ifdef's etc I can use to alias that macro so that it works for 7.1+7.2 (and perhaps somebody on the internals mailing list can help). Leave it with me for another attempt, I won't get to it for a couple of days, though.

haberman:
@brettmc I would recommend defining a protobuf-specific macro here: https://github.com/protocolbuffers/protobuf/blob/master/php/ext/google/protobuf/protobuf.h#L46-L85

Often we will polyfill the new interface into old PHPs. But since PHP changed the number of args without changing the symbol name, we'll probably just want to have our own macro like `PROTOBUF_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX`

brettmc:
@haberman I think it's working (builds and tests pass with 7.0 and 8.1)

brettmc:
@haberman - the (hopefully) last issue with 32bit tests is that I added `ext-bcmath` as a required extension in composer.json (since it's used extensively, eg in `Internal/GPBUtil`), but it appears to not be enabled in the test setup for `7.1-zts`. The path of least resistance would be for me to just remove requirement and leave everything as it was. Wdyt?

MrMage:
FYI, I think there's a list of "recommended" instead of "required" extensions; maybe that would be appropriate?

brettmc:
@haberman - update on the ext-bcmath dependency. The 32bit `php-7.x-zts` images used in testing do not have the bcmath extension. The tests currently pass because only the extension, and not the PHP library, are tested with `-zts`. That makes sense, you do not need bcmath for the extension.
I've added testing against 8.1, but note that some build process needs to publish a newer version of `protobuftesting/php80_<hash>` based on my Dockerfile changes - until that happens it falls back to php 8.0 which happens to be in the PATH. Lastly, I needed up update the php80 image from jessie to stretch since 8.1 would not compile with the openssl version which jessie provides.

kokoro/linux/dockerfile/test/php80/Dockerfile
```diff
@@ -90,6 +90,34 @@ RUN wget -O phpunit https://phar.phpunit.de/phpunit-9.phar \
   && cp phpunit /usr/local/php-8.0/bin \
   && mv phpunit /usr/local/php-8.0-zts/bin
 
+# php 8.1
```
brettmc, line 93: I wasn't sure whether the idea was to have one dockerfile per php version moving forward, or an all-in-one like is done for php7.x

php/src/Google/Protobuf/Internal/MapFieldIter.php
```diff
@@ -117,17 +120,18 @@ public function key()
      *
      * @return void
      */
+    #[\ReturnTypeWillChange]
     public function next()
```
Tony-Sol, line 124:
```suggestion
    public function next(): void
```
According to doc, any return value will be ignored, so it seems to be safe to use "void" return type hint

php/src/Google/Protobuf/Internal/RepeatedField.php
```diff
@@ -138,6 +140,7 @@ public function offsetGet($offset)
      * @throws \ErrorException Non-existing index.
      * @throws \ErrorException Incorrect type of the element.
      */
+    #[\ReturnTypeWillChange]
     public function offsetSet($offset, $value)
```
Tony-Sol, line 144:
```suggestion
    public function offsetSet($offset, $value): void
```
According to doc, any return value will be ignored, so it seems to be safe to use "void" return type hint even without `#[\ReturnTypeWillChange]` attribute as like `Iterator::next`

php/src/Google/Protobuf/Internal/RepeatedField.php
```diff
@@ -209,6 +212,7 @@ public function offsetSet($offset, $value)
      * @throws \ErrorException The element to be removed is not at the end of the
      * RepeatedField.
      */
+    #[\ReturnTypeWillChange]
     public function offsetUnset($offset)
```
Tony-Sol, line 216:
```suggestion
    public function offsetUnset($offset): void
```

php/src/Google/Protobuf/Internal/RepeatedFieldIter.php
```diff
@@ -71,6 +71,7 @@ public function __construct($container)
      *
      * @return void
      */
+    #[\ReturnTypeWillChange]
     public function rewind()
```
Tony-Sol, line 75:
```suggestion
    public function rewind(): void
```

php/src/Google/Protobuf/Internal/RepeatedFieldIter.php
```diff
@@ -101,6 +104,7 @@ public function key()
      *
      * @return void
      */
+    #[\ReturnTypeWillChange]
     public function next()
```
Tony-Sol, line 108:
```suggestion
    public function next(): void
```

Tony-Sol (commented):
maybe added some safe, 8.0 and 7.x compatible typehints

UPD: suggested `: void` hints is not safe with 7.0, my fault for not checked this before comment

php/src/Google/Protobuf/Internal/RepeatedField.php
brettmc: @Tony-Sol - I did have have void return types originally, but removed them to retain BC to php 7.0 (some discussion above on the subject). The void return type only [arrived in 7.1](https://wiki.php.net/rfc/void_return_type), and the tests were failing (from memory, searching for and not finding a void class in the current namespace)

JimmyBanks:
@haberman Is there a timeline on this being reviewed/pulled?

haberman:
@brettmc I've updated the new Docker images based on your branch and kicked off the tests. We'll see what happens. Thanks for all your work on this PR!

brettmc:
@haberman I see all green for the checks, but I can't find the kokoro job that ran 8.x tests to confirm that all is well

haberman:
Hmm, it looks like only some of the presubmits ran. Let me try kicking it again.

haberman:
All tests have passed, but it looks like 8.1 tests did not run. I am realizing that your edits to `test.sh` are not taking effect because the current Linux PHP build is using [this script](https://github.com/protocolbuffers/protobuf/blob/master/kokoro/linux/php_all/build.sh), which does not invoke tests.sh.

https://github.com/protocolbuffers/protobuf/blob/master/kokoro/linux/php_all/build.sh is using using Docker images that are not in this repository. I'm going to merge this PR as-is, since it's clearly not breaking any of our existing use cases. Then I'll add an 8.1 Docker image in the other (private) repo and send another PR to test it. If I run into any problems I'll let you know.

MrMage:
Thanks for merging this, @haberman! Are there any plans to tag a patch release that includes this fix?

PR #9376, opened 2022-01-05
Default two more of the options from env variables.

thomasvl:
For build systems that don't allow someone to easily add in generator args,
support env variables for the values.

Most the bool env var support to a helper since it is used in three places now.

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -73,6 +73,14 @@ using ::open;
 
 namespace {
 
+bool BoolFromEnvVar(const char* env_var, bool default_value) {
+  const char* value = getenv(env_var);
+  if (value) {
+    return std::string("YES") == ToUpper(value);
+  }
+  return default_value;
+}
+
 class SimpleLineCollector : public LineConsumer {
  public:
   SimpleLineCollector(std::unordered_set<std::string>* inout_set)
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -73,6 +73,14 @@ using ::open;
 
 namespace {
 
+bool BoolFromEnvVar(const char* env_var, bool default_value) {
+  const char* value = getenv(env_var);
+  if (value) {
+    return std::string("YES") == ToUpper(value);
```
dmaclach, line 79: Could simplify reading this to `return ToUpper(value) == "YES";`

dmaclach: Is 'yes' the only case we handle? 

https://developer.apple.com/documentation/foundation/nsstring/1409420-boolvalue?language=objc as an example

dmaclach (approved):

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: > Could simplify reading this to `return ToUpper(value) == "YES";`

It still has to build the `std::string`, so I'm not sure it really counts as simplier.

> Is 'yes' the only case we handle?
> 
> https://developer.apple.com/documentation/foundation/nsstring/1409420-boolvalue?language=objc as an example

We don't need to match everything, just going for the basics.

PR #9412, opened 2022-01-14
[DO NOT MERGE YET] add native Typescript support

chris-kruining:
Hi, I've run into the issue that my setup with remix (remix.run) is not capable of handling the esports from the generated js files (even with the support of grpc-web).

At first I started working on a patch for grpc-web in order to fix it, however the deeper I dove into that the more it seems that the whole extension is a workaround.

js now supports client-to-server streaming natively with the newly launched fetch upload streaming.

Therefor I want to make this patch to update protoc to generate proper modern files 

To do this I want to:
- generate typescript files for the messages and client stubs.
- update the npm library to be fitting a more modern standard and implement the client side streaming.

my goal is to just do the typescript part as not to blow up the scope for me to do personally since I am way out of my depth in c++. For this reason I am hoping for support to properly implement this


perezd:
Thanks for opening the PR, I wanted to let you know were not planning to support TypeScript in this way with this plugin so I wanted to set your expectations about merging this. Feel free to fork and maintain a plugin that meets your needs, but it won't be something we can devote resources to supporting at the moment.

chris-kruining:
Is updating the javascript that is produced to modern syntax still an option?

chris-kruining:
I'm closing this as I have found success with the protoc-gen-ts plugin

PR #9415, opened 2022-01-15
Implement Utf8.decodeUtf8 by using String constructor

### LINKED ISSUE #9414: Utf8.decodeUtf8: Simplify and optimize for ascii strings
`Utf8.decodeUtf8` can be replaced with a call to `String(byte[] bytes, int offset, int length, Charset charset)` and then a search for the replacement string `'\uFFFD'` in order to decide whether to throw `InvalidProtocolBufferException`.

If this idea is not ruled out, I can provide benchmark results that compare different strings (ascii / non ascii). Spoiler: for ascii strings of 50 characters this method is ~30% faster than the current implementation on modern JVMs.

I created a PR that makes this idea concrete.

amirhadadi:
Implement Utf8.decodeUtf8 by using String constructor and searching for the replacement string "\uFFFD". This greatly simplifies the implementation, speeds it up for ascii and saves in memory allocations for non ascii strings.

Fixes #9414.


elharo (commented):
You need to update to master first.

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -614,8 +615,35 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
      *
      * @throws InvalidProtocolBufferException if the byte array slice is not valid UTF-8.
      */
-    abstract String decodeUtf8(byte[] bytes, int index, int size)
-        throws InvalidProtocolBufferException;
+    String decodeUtf8(byte[] bytes, int index, int size)
+        throws InvalidProtocolBufferException {
+      final String s;
```
elharo, line 620: no final per google style

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -614,8 +615,35 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
...
+      try {
+        s = new String(bytes, index, size, StandardCharsets.UTF_8);
```
elharo, line 622: variable declaration needs to be inside the try block

elharo, line 622: we can't use StandardCharsets yet due to android compatiblity

elharo (changes requested):
It's an interesting idea. You'd need to show benchmarks for proto itself though, not just this technique outside of protos. Also it needs to run on android API level 14.

java/core/src/main/java/com/google/protobuf/Utf8.java
amirhadadi: I made the change as you requested, but it means that it's no longer clear (without analysis of all code inside the try block) which call threw the exception that is caught in `catch (final IndexOutOfBoundsException e)`.

amirhadadi: Sure

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -613,8 +614,34 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
...
+
+        // TODO(amirhadadi): When targeting Java 9, replace with Arrays.equals that accepts from and to indexes
+        // to avoid copying the byte array slice.
+        if (Arrays.equals(s.getBytes(Internal.UTF_8), Arrays.copyOfRange(bytes, index, index + size))) {
+          return s;
+        }
+
+        throw InvalidProtocolBufferException.invalidUtf8();
+      } catch (final IndexOutOfBoundsException e) {
+        throw new ArrayIndexOutOfBoundsException(
```
elharo, line 641: This should be an InvalidProtocolBufferException here, not a runtime exception.

elharo (changes requested):

java/core/src/main/java/com/google/protobuf/Utf8.java
amirhadadi: The catch clause is catching the `StringIndexOutOfBoundsException` thrown from `String` constructor.
`InvalidProtocolBufferException` is not caught.
I think I'm missing your point.

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -613,8 +614,34 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
...
+          return s;
+        }
+
+        // Since s contains "\uFFFD" there are 2 options:
+        // 1) The byte array slice is invalid UTF-8.
+        // 2) The byte array slice is valid UTF-8 and contains encodings for "\uFFFD".
+        // To rule out (1), we encode s and compare it to the byte array slice.
+        // If the byte array slice was invalid UTF-8, then we would get a different sequence of bytes.
+
+        // TODO(amirhadadi): When targeting Java 9, replace with Arrays.equals that accepts from and to indexes
```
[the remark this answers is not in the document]
amirhadadi: Done

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -613,8 +614,34 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
...
+        // to avoid copying the byte array slice.
+        if (Arrays.equals(s.getBytes(Internal.UTF_8), Arrays.copyOfRange(bytes, index, index + size))) {
+          return s;
+        }
+
+        throw InvalidProtocolBufferException.invalidUtf8();
+      } catch (final IndexOutOfBoundsException e) {
```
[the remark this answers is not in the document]
amirhadadi: Done

java/core/src/main/java/com/google/protobuf/Utf8.java
```diff
@@ -613,8 +614,31 @@ private static int partialIsValidUtf8(final ByteBuffer buffer, int index, final
      *
      * @throws InvalidProtocolBufferException if the byte array slice is not valid UTF-8.
      */
-    abstract String decodeUtf8(byte[] bytes, int index, int size)
-        throws InvalidProtocolBufferException;
+    String decodeUtf8(byte[] bytes, int index, int size)
```
amirhadadi, line 617: @elharo should I replace the body of the existing

```
  static String decodeUtf8(byte[] bytes, int index, int size)
      throws InvalidProtocolBufferException {
    return processor.decodeUtf8(bytes, index, size);
  }

```

with this implementation and remove

`String decodeUtf8(byte[] bytes, int index, int size)`

from `Utf8::Processor`?

java/core/src/main/java/com/google/protobuf/Utf8.java
amirhadadi: I'm preserving the [current behavior](https://github.com/protocolbuffers/protobuf/blob/d8ccfbf00541eb0705bbba640d65e3df814b3d2a/java/core/src/main/java/com/google/protobuf/Utf8.java#L1370) of throwing `ArrayIndexOutOfBoundsException` in case the bytes slice is not within the bounds of the array.

java/core/src/main/java/com/google/protobuf/Utf8.java
elharo: I think you can simply call this without copying and pasting

amirhadadi: @elharo not only copying and pasting, also removing `String decodeUtf8(byte[] bytes, int index, int size)` from `Utf8::Processor'.

There's no longer a point in keeping the implementation in `Processor` since there are no safe and unsafe implementations, only a safe one.

amirhadadi: I pushed this change to make the intent clear.

acozzette:
@amirhadadi Have you signed the CLA? The CLA bot says it's unable to find a CLA for the change, and we won't be able to merge the pull request without that. Let us know if you need helping sorting this out (sometimes the corporate CLA is tricky if you're doing this for an employer).

java/core/src/main/java/com/google/protobuf/Utf8.java
ahadadi: @elharo anything else should be done here?

ahadadi:
@acozzette I now signed the CLA.

acozzette:
One of our main concerns about merging this is that we want to have good performance even on old Android phones. Based on the benchmark results on #9414, it sounds like the change is not a net benefit except with fairly recent JVMs, so that could be a problem for Android. One possible solution would be to implement the optimization only in `UnsafeProcessor`, since Android appears to always use `SafeProcessor` at the moment.

I'm also wondering whether this code path is exercised by proto2 or if it's only proto3. This could make a big difference because I suspect that (for us at least) our workloads are still dominated by proto2 messages.

amirhadadi:
@acozzette should I move this implementation to `UnsafeProcessor` then?

acozzette:
@amirhadadi Yes, I think moving it to `UnsafeProcessor` would be good because then we wouldn't have to worry about potentially regressing performance on Android phones. But even after that, it may take us some time to figure out if we want to merge this, because we would like to benchmark it carefully. Is there any chance that your benchmark is published somewhere that we could try it out?

amirhadadi:
@acozzette you can see the benchmark [here](https://github.com/amirhadadi/string-benchmarks).

amirhadadi:
@acozzette did you get a chance to look at the benchmarks?

acozzette:
@amirhadadi Thanks for the benchmark link. Looks promising to me but I believe @elharo is taking a closer look at our Java benchmarks.

acozzette:
@amirhadadi Would you mind fixing the merge conflicts and making this change only in the `UnsafeProcessor`? I think if we can confirm that there's a positive performance benefit or even just no performance regression, this change will be a nice improvement just from the code simplification.

amirhadadi:
@acozzette merged and applied the change only for UnsafeProcessor.

acozzette:
@amirhadadi Thanks!

PR #9418, opened 2022-01-18
Match service argument names to abstract interface

nipunn1313:
The base class/documentation suggest that the argument
names are `self` and `done`, while the runtime used
argument names `srvc` and `callback`.

`mypy.stubtest` was able to identify this - as it compares
the types (autogenerated by
[`mypy-protobuf`](https://github.com/dropbox/mypy-protobuf/))
to the actual code generated by protoc at runtime.

Since the stubs assume the generated code matches the abstract
interface in service.py - it saw these differences.

### DIFF
--- a/python/google/protobuf/service_reflection.py
+++ b/python/google/protobuf/service_reflection.py
@@ -133,7 +133,7 @@ class _ServiceBuilder(object):
     """
     self.descriptor = service_descriptor
 
-  def BuildService(self, cls):
+  def BuildService(builder, cls):
     """Constructs the service class.
 
     Args:
@@ -143,18 +143,25 @@ class _ServiceBuilder(object):
     # CallMethod needs to operate with an instance of the Service class. This
     # internal wrapper function exists only to be able to pass the service
     # instance to the method that does the real CallMethod work.
-    def _WrapCallMethod(srvc, method_descriptor,
-                        rpc_controller, request, callback):
-      return self._CallMethod(srvc, method_descriptor,
-                       rpc_controller, request, callback)
-    self.cls = cls
+    # Making sure to use exact argument names from the abstract interface in
+    # service.py to match the type signature
+    def _WrapCallMethod(self, method_descriptor,
+                        rpc_controller, request, done):
+      return builder._CallMethod(self, method_descriptor,
+                       rpc_controller, request, done)
+    def _WrapGetRequestClass(self, method_descriptor):
+      return builder._GetRequestClass(method_descriptor)
+    def _WrapGetResponseClass(self, method_descriptor):
+      return builder._GetResponseClass(method_descriptor)
+
+    builder.cls = cls
     cls.CallMethod = _WrapCallMethod
-    cls.GetDescriptor = staticmethod(lambda: self.descriptor)
+    cls.GetDescriptor = staticmethod(lambda: builder.descriptor)
     cls.GetDescriptor.__doc__ = "Returns the service descriptor."
-    cls.GetRequestClass = self._GetRequestClass
-    cls.GetResponseClass = self._GetResponseClass
-    for method in self.descriptor.methods:
-      setattr(cls, method.name, self._GenerateNonImplementedMethod(method))
+    cls.GetRequestClass = _WrapGetRequestClass
+    cls.GetResponseClass = _WrapGetResponseClass
+    for method in builder.descriptor.methods:
+      setattr(cls, method.name, builder._GenerateNonImplementedMethod(method))
 
   def _CallMethod(self, srvc, method_descriptor,
                   rpc_controller, request, callback):
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -100,3 +100,8 @@ Patch contributors:
   Andrew Paprocki <andrew@ishiboo.com>
     * Fixed minor IBM xlC compiler build issues
     * Added atomicops for AIX (POWER)
+  Nipunn Koorapati <nipunn1313@gmail.com>
+    * Provide a type alias field ValueType on EnumTypeWrapper
+    * Match service argument names to abstract interface
+
+


nipunn1313:
The stubtest mismatches can be seen in this PR: https://github.com/dropbox/mypy-protobuf/pull/337  (logs [here](https://github.com/dropbox/mypy-protobuf/runs/4850205874?check_suite_focus=true))

Example error (pulled from the logs)
```
error: testproto.test_pb2.ATestService_Stub.GetResponseClass is inconsistent, runtime does not have argument "method_descriptor"
Stub: at line 20
def (self: google.protobuf.service.Service, method_descriptor: google.protobuf.descriptor.MethodDescriptor) -> Type[google.protobuf.message.Message]
Runtime: at line 192 in file /home/runner/work/mypy-protobuf/mypy-protobuf/venv_3.8.11/lib/python3.8/site-packages/google/protobuf/service_reflection.py
def (method_descriptor)
```

With this diff, things match again!

nipunn1313:
Main benefit is that the runtime behavior will match the documentation.
For example if someone calls `CallMethod(done=mycallback)`, [per the documentation](https://googleapis.dev/python/protobuf/latest/google/protobuf/service.html#google.protobuf.service.RpcChannel.CallMethod) - now it would match

python/google/protobuf/service_reflection.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, service_descriptor):
     """
     self.descriptor = service_descriptor
 
-  def BuildService(self, cls):
+  def BuildService(builder, cls):
```
elharo, line 136: Not a Python expert, but isn't self a keyword to refer to the current object? That is, I don't think you can just change the name on this.

elharo (changes requested):

python/google/protobuf/service_reflection.py
nipunn1313: `self` is actually a convention, not a keyword
```
>>> import keyword
>>> keyword.kwlist
['False', 'None', 'True', '__peg_parser__', 'and', 'as', 'assert', 'async', 'await', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'nonlocal', 'not', 'or', 'pass', 'raise', 'return', 'try', 'while', 'with', 'yield']
```

In this specific case, I think it's actually clarifying, since inside the function we're adding methods to the object being built - and need to refer to those argument as `self`. Rather than aliasing the outer `self` of the builder, we're able to just name it `builder`.

Broadly, I don't think this would be the modern way of writing `service_reflection.py` - as sticking methods as properties is a bit convoluted (and there's no enforcement that it matches the abstract base class) - but it's probably safer to do any major changes to `service_reflection.py`. Using something like https://docs.python.org/3/library/abc.html would help enforce that the subclasses actually implement the interface.

nipunn1313: The tests in `python/google/protobuf/internal/service_reflection_test.py` are actually great for testing this code.
```
cd python
python3 setup.py test
```

zhangskz (commented):
(Sorry, looks like I had a pending comment I forgot to actually submit)

python/google/protobuf/service_reflection.py
```diff
@@ -143,18 +143,25 @@ def BuildService(self, cls):
...
+    builder.cls = cls
     cls.CallMethod = _WrapCallMethod
-    cls.GetDescriptor = staticmethod(lambda: self.descriptor)
+    cls.GetDescriptor = staticmethod(lambda: builder.descriptor)
     cls.GetDescriptor.__doc__ = "Returns the service descriptor."
-    cls.GetRequestClass = self._GetRequestClass
-    cls.GetResponseClass = self._GetResponseClass
-    for method in self.descriptor.methods:
-      setattr(cls, method.name, self._GenerateNonImplementedMethod(method))
+    cls.GetRequestClass = _WrapGetRequestClass
```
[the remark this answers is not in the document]
nipunn1313: It works at runtime, but the types are wrong when being compared to the base class (google.protobuf.service.Service) [doc](https://googleapis.dev/python/protobuf/latest/google/protobuf/service.html#google.protobuf.service.Service.GetRequestClass)

Here's an example error from stubtest from before this change
```
error: testproto.test_pb2.ATestService_Stub.GetResponseClass is inconsistent, runtime does not have argument "method_descriptor"
Stub: at line 20
def (self: google.protobuf.service.Service, method_descriptor: google.protobuf.descriptor.MethodDescriptor) -> Type[google.protobuf.message.Message]
Runtime: at line 192 in file /home/runner/work/mypy-protobuf/mypy-protobuf/venv_3.8.11/lib/python3.8/site-packages/google/protobuf/service_reflection.py
def (method_descriptor)
```

Specifically, the `self` being passed in to `GetRequestClass` is the `self` of the builder class, but the abstract base class (and documentation) expects it to be the `self` of the built class. With this change, the `self` being passed into `builder.GetRequestClass` is the `builder`, not the built class.

This test still works confirming that everything works at runtime
https://github.com/protocolbuffers/protobuf/blob/c87b64f992a13140745a44644550a0d59bc3d1fc/python/google/protobuf/internal/service_reflection_test.py#L80

zhangskz (approved):

acozzette:
Thanks, @nipunn1313.

neo:
Hi all,

I recently encountered a new error for python from 3.19 to 3.20, which I have a feeling that it might be related to this:
```
TypeError: _WrapGetResponseClass() missing 1 required positional argument: 'method_descriptor'
```

It points to this line that was working before in a lib I'm working with:
```
service.GetResponseClass(method)
```

I tried a couple things, and this making it this somehow made it work
```diff
- service.GetResponseClass(method)
+ service.GetResponseClass(service, method)
```

It would be great if someone can let me know what you think, thanks a lot!! ❤️

nipunn1313:
There are some tests here that show it working with the one arg. https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/internal/service_reflection_test.py#L82

Can you provide more context from your example? What is `service` in your example?
In these tests, they appear to be an instance of the service.
This matches what the docs show: https://googleapis.dev/python/protobuf/latest/google/protobuf/service.html#google.protobuf.service.Service.GetRequestClass

neo:
Thank you for your speedy reply!

With my limited knowledge to the lib, I would guess the `service` in my example is either an instance of a service, or a `ServiceDescriptor`?

Here the line right above:
```
method = service.GetDescriptor().FindMethodByName(method)
```

PR #9426, opened 2022-01-20
Add ToProto() method to all C# descriptor classes

### LINKED ISSUE #9425: Provide (safe) access to reflection protos in C#

C#

Some APIs (e.g. [BigQuery Storage](https://github.com/googleapis/googleapis/blob/ab61c52324b39d1c8619a96ab19cc6c173b13f2c/google/cloud/bigquery/storage/v1/protobuf.proto#L39) use `DescriptorProto` for schema descriptions.

In C# we have access to reflection via the `MessageDescriptor` type, but there's no access to the underlying proto.

`ToProto()` methods in all descriptor types, returning a clone of the underlying proto.

Access via .NET reflection :(

**Additional context**

I'll create a PR to implement this.

jskeet:
Fixes #9425.

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
@@ -69,6 +69,14 @@ namespace Google.Protobuf.Reflection
         internal EnumDescriptorProto Proto { get { return proto; } }
 
         /// <summary>
+        /// Returns a clone of the underlying <see cref="EnumDescriptorProto"/> describing this enum.
+        /// Note that a copy is taken every time this method is called, so clients using it frequently
+        /// (and not modifying it) may want to cache the returned value.
+        /// </summary>
+        /// <returns>A protobuf representation of this enum descriptor.</returns>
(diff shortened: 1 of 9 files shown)


csharp/src/Google.Protobuf/Reflection/EnumDescriptor.cs
```diff
@@ -68,6 +68,14 @@ internal EnumDescriptor(EnumDescriptorProto proto, FileDescriptor file, MessageD
 
         internal EnumDescriptorProto Proto { get { return proto; } }
 
+        /// <summary>
+        /// Returns a clone of the underlying <see cref="EnumDescriptorProto"/> describing this enum.
+        /// Note that a copy is taken every time this method is called, so clients using it frequently
```
jtattermusch, line 73: nit: worth mentioning that a deep clone is taken? (which also explains why it's safe to to .Clone() internally since it provides a deep clone).
Same below.
Leaving resolution up to you - I don't feel strongly about this at all.

jtattermusch (approved):
LGTM with one very minor nit.

PR #9428, opened 2022-01-20
Migrate kotlin bazel tests from kt_jvm_test

deannagarcia:
We were previously using kt_jvm_test to run Kotlin tests with bazel. However, that library is not compatible with windows (see bug here: https://github.com/bazelbuild/rules_kotlin/issues/599). This PR migrates the kt_jvm_test calls to separate calls: a kt_jvm_library and a java_test.


perezd (approved):
LGTM!

PR #9430, opened 2022-01-21
Fixes NullReferenceException when accessing FieldDescriptor.IsPacked

xoofx:
Hey,

While developing a small service [grpc-curl](https://github.com/xoofx/grpc-curl), I found a small bug in `FieldDescriptor.IsPacked` that was throwing a `NullReferenceException` that I had to workaround here:

https://github.com/xoofx/grpc-curl/blob/f7bbe35af79713f278150b96ce51817807a06581/src/DynamicGrpc/DynamicMessageSerializer.cs#L434-L455

So this PR is a small bug fix. Though, I haven't found a test to update about it, not sure if we have any?

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/FieldDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/FieldDescriptor.cs
@@ -239,7 +239,8 @@ namespace Google.Protobuf.Reflection
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


jskeet (commented):
LGTM, but I'll let a member of the protobuf team actually approve.

elharo (commented):
I don't know of a test for this, but it would help to add one that covers this bug.

MarcelRaschke (approved):

PR #9434, opened 2022-01-21
Use constexpr for FieldTypeClass

gjasny:
Otherwise in C++20 / VisualStudio 2022 the following warning is emitted:

```
warning C5054: operator '*': deprecated between enumerations of different types
```


acozzette:
Thanks, @gjasny!

PR #9436, opened 2022-01-22
Updated Ruby to the current version of upb

haberman:
This has a few behavior changes and bugfixes. From the release notes:

  * JSON will now output shorter strings for double and float fields when possible without losing precision.
  * Encoding and decoding of binary format will now work properly on big-endian systems.
  * UTF-8 verification was fixed to properly reject surrogate code points.
  * Unknown enums for proto2 protos now properly implement proto2's behavior of putting such values in unknown fields.

The new version of upb fully supports extensions and custom options, opening the door to (finally) fixing https://github.com/protocolbuffers/protobuf/issues/1198.

A significant number of upb APIs are renamed due to https://github.com/protocolbuffers/upb/pull/485

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,14 @@
+Unreleased Changes (C++/Java/Python/PHP/Objective-C/C#/Ruby/JavaScript)
+
+  Ruby
+  * JSON will now output shorter strings for double and float fields when possible
+    without losing precision.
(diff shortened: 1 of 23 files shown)


fowles (commented):
Please separate the semantic and non-semantic portions of this into separate changes.

fowles (approved):

PR #9437, opened 2022-01-24
link with correct C runtime on MSVC

r2d3:
when using CMake >= 3.15, we use CMAKE_MSVC_RUNTIME_LIBRARY
to reproduce the /MT vs /MD when protobuf_MSVC_STATIC_RUNTIME
is TRUE or FALSE
MultiThreaded is for /MT
MultiThreadedDebug is for /MTd
MultiThreadedDLL is for /MD
MultiThreadedDebugDLL is for /MDd


acozzette:
Thanks, @r2d3.

PR #9438, opened 2022-01-25
Selectively add source or gen dir to includes.

comius:
When both directories are added this results in protoc emitting a
"warning: directory does not exist.". This makes sense because when
there are no inputs from the other directory, it is also not present
n the sandbox where protoc is executed.

Fixes https://github.com/protocolbuffers/protobuf/issues/6049
Tested manually with envoy project.


acozzette:
Thanks, @comius.

PR #9441, opened 2022-01-26
Use proper SPDX license identifier

Shortfinga:
The previously used term "3-Clause BSD License" is not properly
standarized. A common standard is SPDX, therefore "3-Clause BSD License"
is substituted with "BSD-3-Clause" which is a SPDX identifier.

`grep -rl "3-Clause BSD License" | xargs -n1 sed -i "s/3-Clause BSD
License/BSD-3-Clause/g"`


elharo:
Thanks!

Issue #19, opened 2014-09-10
Proposal: Divide ProtoBuf into a formal specification, and the separate language implementations


protobufel:
I think it is long overdue to have the separate 
1. the formal ProtoBuf specification, including the grammar, the serialization spec, the formal language independent API definitions, and the TCK;
2. the separate concrete language  implementations based on the above, with their own maintenance and release schedule.
3. (optional, and desired) language specific ProtoBuf generators, at least for Java, in addition to the existing protoc.

This would leverage the language specific tools, libraries, and domains, allowing for better, easier, and faster maintenance and evolution.

In addition, regarding Java, it is time to get rid of Java 5, and, maybe, Java 6 support, making a really portable Maven project, with modern tests, and the updated source.

It is very easy to do - the separation could be done within hours! The benefits, IMHO, are huge.

Cheers,
David

xfxyjwf:
FYI. It has been decided protobuf implementations will live in this same repo. We don't have plans to separate the implementations for the time being.

PR #9458, opened 2022-01-31
fix: [PHP] add missing reserved classnames

bshaffer:
Protobuf generates class with the name `Parent` in PHP (see [this PR](https://github.com/googleapis/google-cloud-php/pull/5073/files#diff-2a1d1b0f9829bac3607be2bf330a07e5b136ae8f945571dbdb90755ff9daca75)). This is an invalid class name, and results in a fatal error. The same is true for the word `self`:

https://www.php.net/manual/en/reserved.classes.php

```
Following identifiers may not be used as a class name as they have special purpose.
```

Our current list represents PHP "keywords", which have almost the same restrictions in PHP, but are slightly more restricted:

https://www.php.net/manual/en/reserved.keywords.php
```
The following words cannot be used as constants, class names, function or method names. They are, however, allowed as property, constant, and method names of classes, interfaces and traits, except that class may not be used as constant name.
```

The difference here is keywords cannot be used in **global** functions or constants. As protobuf for PHP does not generate any global functions or constants, they can be treated as the same, and do not need special treatment (such as `kReservedClasses`). 

This is **not** a backwards-compatibility breaking change, as any protobuf objects with the names `self` or `parent` would be throwing PHP fatal errors if generated today.


bshaffer:
@acozzette I fixed the tests - I missed adding them to the `kValidConstantNames` array. This PR should be good to go now 🤞

bshaffer:
@acozzette okay they're really fixed this time I promise!

acozzette:
Thanks, @bshaffer.

bshaffer:
@acozzette my pleasure! Thanks for the quick review.

Do you know when this fix will be available in a release?

andreladocruz:
Got this error:

#19 12.77 /tmp/pear/temp/pear-build-defaultuserjglifJ/protobuf-3.20.0/libtool: line 1290: can't create third_party/utf8_range/naive.loT: nonexistent directory
#19 12.77 mkdir third_party/utf8_range/.libs
#19 12.77 mkdir: can't create directory 'third_party/utf8_range/.libs': No such file or directory
#19 12.77 make: *** [Makefile:226: third_party/utf8_range/naive.lo] Error 1
#19 12.77 ERROR: `make' failed

icatalina:
👋 I just noticed `Object` is not listed as a reserved class name in the php documentation, but it is:

```
PHP Fatal error:  Cannot use 'Object' as class name as it is reserved in Object.php on line 14

Fatal error: Cannot use 'Object' as class name as it is reserved in Object.php on line 14

Errors parsing Object.php
```

I've already raised an issue with them: https://github.com/php/doc-en/issues/1518

damianwadley:
Object it isn't a reserved class name. It is a reserved word and comes with broader restrictions than just as a class name.

icatalina:
Hey @damianwadley, sorry for the confusion. I suppose being a reserved word makes the `Object` class "reserved" because you can't name a class like that. It sure encompasses more than just class names as you pointed out.

In this particular case, protobuf is generating a class called `Object` which isn't valid php after `7.2`.

PR #9460, opened 2022-02-01
Fix checking Windows platform with preprocessor

Biswa96:
Check _WIN32 instead of _MSC_VER because the later one is for MSVC only.
This fixes building with mingw-w64 toolchain which uses GCC or Clang.


acozzette:
Thanks, @Biswa96.

PR #9462, opened 2022-02-02
fix: use E_USER_ERROR for trigger_error calls

dwsupplee:
According to [the manual for trigger_error](https://www.php.net/manual/en/function.trigger-error.php) only constants from the E_USER family can be used. Without this, a cryptic error is presented.


elharo (commented):
Is there a test for this?

acozzette:
Thanks, @dwsupplee.

PR #9466, opened 2022-02-03
Change gcc version

deannagarcia:
Change Dockerfile gcc version since we no longer support 4.8


acozzette:
Let me go ahead and close this since we ended up using another workaround to keep gcc 4.8 working for now.

PR #9468, opened 2022-02-04
Support forward declaration of dependent messages via opt-in option in protoc

Izaron:
Suppose that we have `x.proto` and `y.proto` files. If `x.proto` imports `y.proto`, the generated `x.pb.h` file will include `y.pb.h` file.

In our C++ project protobuf is widely used. It contains hundreds of `.proto` files that actively import each other. Since some of the protobuf messages are huge, each additional dependency results in a further slowdown of the project to compilation.

When project's source file depends on the messages from `x.proto` (and includes `x.pb.h` file), the compiler is forced to transitively parse `y.pb.h`, which in most cases is not necessary.

In fact `x.pb.h` doesn't need to include `y.pb.h` to be usable. Instead of including `y.pb.h`, `x.pb.h` can forward-declare used classes and enums that are defined in `y.pb.h`. It will reduce the cost to parse `y.pb.h`. If a source file needs both `x.pb.h` and `y.pb.h`, it shall now include both of them.

We've measured compile time on Intel Xeon E5-2650 cores clang12 and SSD.

**Build from scratch BEFORE the patch**:
```
real    19m23.164s
user    8m59.026s
sys     5m12.126s
```
**Build from scratch AFTER the patch and codebase fixup**:
```
real    16m0.855s
user    8m29.934s
sys     4m55.299s
```
**Build after changing an essential .proto file BEFORE the patch**:
```
real    15m35.174s
user    7m19.943s
sys     4m27.037s
```
**Build after changing an essential .proto file AFTER the patch**:
```
real    11m45.793s
user    6m58.235s
sys     4m32.752s
```

Though applying this option requires tidying codebase according to include-what-you-use principle, it saves about 25% of the build time which looks like a major improvement to us.

### DIFF
--- a/src/google/protobuf/compiler/cpp/cpp_file.cc
+++ b/src/google/protobuf/compiler/cpp/cpp_file.cc
@@ -37,6 +37,7 @@
 #include <iostream>
 #include <map>
 #include <memory>
+#include <queue>
 #include <set>
 #include <unordered_map>
 #include <unordered_set>
@@ -385,6 +386,10 @@ void FileGenerator::GenerateSourceIncludes(io::Printer* printer) {
       "#include <algorithm>\n"  // for swap()
       "\n",
       CreateHeaderInclude(target_basename, file_));
+
+  if (!options_.transitive_pb_h) {
+    GenerateDependencyIncludes(printer);
+  }
 
   IncludeFile("net/proto2/io/public/coded_stream.h", printer);
   // TODO(gerbens) This is to include parse_context.h, we need a better way
@@ -1066,10 +1071,36 @@ void FileGenerator::GenerateForwardDeclarations(io::Printer* printer) {
 
   FlattenMessagesInFile(file_, &classes);  // All messages need forward decls.
 
+  std::vector<const FieldDescriptor*> fields;
+  if (!options_.transitive_pb_h || options_.proto_h) {
+    ListAllFields(file_, &fields);
+  }
+
+  if (!options_.transitive_pb_h) {
+    // Add forward declaration for all messages, enums, and extended messages
+    // defined outside the file
+    for (int i = 0; i < fields.size(); i++) {
+      const Descriptor* message_type = fields[i]->message_type();
+      if (message_type && message_type->file() != file_) {
+        classes.push_back(message_type);
+      }
+
+      const EnumDescriptor* enum_type = fields[i]->enum_type();
+      if (enum_type && enum_type->file() != file_) {
+        enums.push_back(enum_type);
+      }
+
+      if (fields[i]->is_extension()) {
+        const Descriptor* message_type = fields[i]->containing_type();
+        if (message_type && message_type->file() != file_) {
+          classes.push_back(message_type);
+        }
+      }
+    }
+  }
+
   if (options_.proto_h) {  // proto.h needs extra forward declarations.
     // All classes / enums referred to as field members
-    std::vector<const FieldDescriptor*> fields;
-    ListAllFields(file_, &fields);
     for (int i = 0; i < fields.size(); i++) {
       classes.push_back(fields[i]->containing_type());
       classes.push_back(fields[i]->message_type());
@@ -1081,7 +1112,9 @@ void FileGenerator::GenerateForwardDeclarations(io::Printer* printer) {
   // Calculate the set of files whose definitions we get through include.
   // No need to forward declare types that are defined in these.
   std::unordered_set<const FileDescriptor*> public_set;
-  PublicImportDFS(file_, &public_set);
+  if (options_.transitive_pb_h) {
+    PublicImportDFS(file_, &public_set);
+  }
 
   std::map<std::string, ForwardDeclarations> decls;
   for (int i = 0; i < classes.size(); i++) {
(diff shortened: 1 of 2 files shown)


haberman:
We have a feature for this internally: non-transitive headers are emitted into separate `foo.proto.h` files. I think the real solution here will be to open-source that feature, not tweak the output of `.pb.h` files.

acozzette:
It looks like we open sourced just about all the proto.h functionality except for the part that actually recognizes the flag to turn it on. @Izaron If you want to try this out, you can just add a couple lines around [here](https://github.com/protocolbuffers/protobuf/blob/867436c0b848d11f3b55eda3c97f1ca4db998ae8/src/google/protobuf/compiler/cpp/cpp_generator.cc#L91) to set `file_options.proto_h` to `true` when the `proto_h` flag is seen.

Izaron:
@haberman, @acozzette, thank you.
We will change protoc code as you suggested see if existing behavior will work for us.

acozzette:
Let me go ahead and close this. This looks like a great idea but it's so similar to our `proto.h` feature that I think we want to stick with `proto.h`. We will open source the last bit of `proto.h` functionality shortly so that no manual edits will be necessary to turn it on.

georgthegreat:
@acozzette, @haberman, I would like to further discuss this proposition.

I have conducted a series of experiments using protoc=3.17.3 with a 4-line patch making `transitive_pb_h` and `proto_h` options configurable from command line. I believe that there is no significant change in any version through the latest release.

It turned out that  `compiler::cpp::Options` indeed has two options regarding the subject being discussed: `proto_h` which defaults to false and `transitive_pb_h` which defaults to true. However the options are tightly bound and `--cpp_out=transitive_pb_h` alone does not work resulting in non-compilable source code.

It also turned out that `--cpp_out=proto_h` is the only option that works as intended. I believe that it is the variant internally used inside Google.

Unfortunately, I do not think it can be applied to open source version of protobuf: there are extreme amounts of `.proto` files in various projects and these projects rely on functional version of `.pb.h`. They can not be switched to `.proto.h` as they have to continue to compile using legacy versions of protoc. This migration would take years under the constant pressure, and I do not think there is a force capable of maintaining such pressure.

While `--cpp_out=proto_h,transitive_pb_h` also looks like a working solution, we can not apply it internally: it would require adding extra output file to each protoc compilation. As we have thousands of .proto files, this would increase the complexity of build graph too much.

I suggest to work on making `--cpp_out=transitive_pb_h` fully functional: if the option to make `.pb.h` non-transitive existed, it should be functional disregarding the value of `proto_h` option.

Please let's continue this discussion to work out a solution that suits everyone.

acozzette:
> Unfortunately, I do not think it can be applied to open source version of protobuf: there are extreme amounts of `.proto` files in various projects and these projects rely on functional version of `.pb.h`. They can not be switched to `.proto.h` as they have to continue to compile using legacy versions of protoc. This migration would take years under the constant pressure, and I do not think there is a force capable of maintaining such pressure.

I agree, but I think this is true of just about any variation of this change. To fix the transitivity problem, code ultimately has to directly include all the proto headers it needs (it's no longer possible to rely on code that is pulled in transitively), so this would be a breaking change for many projects. I think `proto.h` headers will end up being a feature that power users can optionally enable, but we will probably never be able to turn it on unconditionally in open source.

> While `--cpp_out=proto_h,transitive_pb_h` also looks like a working solution, we can not apply it internally: it would require adding extra output file to each protoc compilation. As we have thousands of .proto files, this would increase the complexity of build graph too much.
> 
> I suggest to work on making `--cpp_out=transitive_pb_h` fully functional: if the option to make `.pb.h` non-transitive existed, it should be functional disregarding the value of `proto_h` option.

Are you saying you would like a way to essentially use `proto.h` files but make them still keep the old `pb.h` file name (so no new header files)? It seems to me like it would probably be reasonable to make `--cpp_out=nontransitive_pb_h` enable this behavior, but I would be curious to know what @haberman thinks.

haberman:
We should not emit `.pb.h` files that are non-transitive. Suppose you have a file:

```c++
#include "foo.pb.h"

void Func() {
  TransitivelyIncludedProto proto;
}
```

This file would compile successfully under `--cpp_out=transitive_pb_h`, but would fail to compile under `--cpp_out=nontransitive_pb_h`.

I don't think we should fork our API surface in this way. We shouldn't have codegen options that will break existing code.

> While --cpp_out=proto_h,transitive_pb_h also looks like a working solution, we can not apply it internally: it would require adding extra output file to each protoc compilation. As we have thousands of .proto files, this would increase the complexity of build graph too much.

Are you concerned about complexity in terms of mental overload, or complexity in terms of machine resource costs?

georgthegreat:
Thanks for the comments, @acozzette, @haberman.

I will try to elaborate on the problems highlighted by Joshua.

Speaking of the open-source users of protobuf:

I would not expect such projects to switch to the new option.
My experience says that they

* might want to maintain backward compatibility with unspecified range of protobuf versions,
* do not really care about optimizing compilation time: they do compile entire repository from scratch for every changeset being tested,
* do not have codebases which need such optimization (with a few exceptions).

Anyway, backward incompatible changes may be introduced with every new version of the OSS library being released. Protobuf compilation option is hardly the biggest problem.

While proposing the new option, we were mostly worried about scaling non-transitive headers to the scale of our monorepo.
We use our own build system. While it is basically built on the same principles as Google Bazel, we have two features to highlight:

* build configuration must be as simple as possible, the user neither needs nor has to tweak it,
* the build system must properly invalidate build graph by automatically predicting dependencies between commands (e. g. it parses header dependencies from .cpp files automatically).

We use [trunk based development](https://trunkbaseddevelopment.com/) quite close to its original form: we merge to trunk often while keeping it correct by the means of distributed CI system.

The `--cpp_out=proto_h` approach does not match these principles:

* If we generate two headers for each `.proto` file, the library client can deliberately choose whether he wants to use transitive or non-transitive one.
  Each user has to be familiar with the meaning of `.proto.h` and its difference from `.pb.h`.
  
* If the project owner wants to enforce using non-transitive headers, he will have to implement some style check and ensure it is applied to the entire monorepo.
  This check has to take into account `.proto` files affiliations and the projects configuration.
  
  While the `--cpp_out=nontransitive_pb_h` approach allows to perform per-project migration by the means of fixing usages with a short series of commits, there is no specific point which allows us to universally enable non-transitive protos if we choose `--cpp_out=proto_h` way.

Thus the proposed variant allows us to switch one project at a time, fix the desired behavior in trunk and then continue with the next project. All this is achieved without any additional heavy style checks.

acozzette:
@georgthegreat What you're trying to do makes sense to me. I think you should probably just carry a small local patch to enable `--cpp_opt=nontransitive_pb_h` in protoc. That way you can get exactly the behavior you want but without us having to make the `nontransitive_pb_h` option public (so this would address @haberman's concern about forking the API service).

georgthegreat:
This is what we already did.
Unfortunately, our diff with upstream is not that small, as it could be. Without the patch proposed by @lzaron in this PR, `--cppout=transitive_pb_h=false` (this is the syntax we have chosen at the end) generates incorrect protos, which do have neither #include nor forward-declarations for nested types.

While maintaining this patch by ourselves is fine, we are trying to upstream features and fixes in order to mitigate the difference from upstream.

Issue #9469, opened 2022-02-06
[Objective-C] Add ability to pass `objc_class_prefix` as an argument vs file option


dnkoutso:
**What language does this apply to?**
Objective-C

If it's a proto syntax change, is it for proto2 or proto3?
N/A

If it's about generated code change, what programming language?
Objective-C

**Describe the problem you are trying to solve.**
While the `objc_class_prefix` file option works for most cases and for a single iOS app, it does not scale well when multiple iOS apps want consume the same `.proto` file but use a different class prefix throughout their codebase.

**Describe the solution you'd like**
It would be nice if the `objc_class_prefix` can be passed as an `--objc_opt` argument, similar to `use_package_as_prefix`. This would allow iOS apps to invoke `protoc` and pass a different prefix to use. Example invocation:

```
protoc --objc_out=my_proto.proto --objc_opt=objc_class_prefix=SMPL
```

This would allow different iOS apps to consume the same `.proto` file and avoid having to `sed` or somehow edit the `.proto` file to update the `objc_class_prefix` prior to `protoc` invocation.

This option should be honored _after_ the file option but _before_ the `use_package_as_prefix` option and documentation should state that. We could also error out if both options are passed.

**Describe alternatives you've considered**
Not many right now, as mentioned above, the current workaround is to edit the `.proto` file right before the `protoc` invocation to workaround this issue.

**Additional context**
N/A

thomasvl:
Fixed via #9476

PR #9476, opened 2022-02-07
objective-c: Add `objc_class_prefix` generation option.

dnkoutso:
For https://github.com/protocolbuffers/protobuf/issues/9469

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
@@ -190,10 +190,20 @@ bool ObjectiveCGenerator::GenerateAll(
       // header search path since the generate #import will be more complete.
       generation_options.runtime_import_prefix =
           StripSuffixString(options[i].second, "/");
+    } else if (options[i].first == "default_objc_class_prefix") {
+      // The default objc class prefix to use if specified by the command line 
+      // invocation. The file option is always honored first if one is present.
+      std::string value = options[i].second;
+      if (value.empty()) {
+        *error = "error: default_objc_class_prefix cannot be empty.";
+        return false;
+      }
+      SetDefaultObjcClassPrefix(value);
     } else if (options[i].first == "use_package_as_prefix") {
       // Controls how the symbols should be prefixed to avoid symbols
-      // collisions. The objc_class_prefix file option is always honored, this
-      // is just what to do if that isn't set. The available options are:
+      // collisions. The objc_class_prefix file option is always honored first
+      // followed by the default_objc_class_prefix generator option. This is is just 
+      // what to do if either of these options are not set. The available options are:
       //   "no": Not prefixed (the existing mode).
       //   "yes": Make a prefix out of the proto package.
       bool value = false;
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -99,6 +99,9 @@ class PrefixModeStorage {
  public:
   PrefixModeStorage();
 
+  const std::string default_objc_class_prefix() const { return default_objc_class_prefix_; }
+  void set_default_objc_class_prefix(const std::string& default_objc_class_prefix) { default_objc_class_prefix_ = default_objc_class_prefix; }
+
   bool use_package_name() const { return use_package_name_; }
   void set_use_package_name(bool on_or_off) { use_package_name_ = on_or_off; }
 
@@ -116,13 +119,14 @@ class PrefixModeStorage {
 
  private:
   bool use_package_name_;
+  std::string default_objc_class_prefix_;
   std::string exception_path_;
   std::string forced_prefix_;
   std::unordered_set<std::string> exceptions_;
 };
 
 PrefixModeStorage::PrefixModeStorage() {
-  // Even thought there are generation options, have an env back door since some
+  // Even though there are generation options, have an env back door since some
   // of these helpers could be used in other plugins.
 
   use_package_name_ = BoolFromEnvVar("GPB_OBJC_USE_PACKAGE_AS_PREFIX", false);
@@ -168,6 +172,10 @@ bool PrefixModeStorage::is_package_exempted(const std::string& package) {
 PrefixModeStorage g_prefix_mode;
 
 }  // namespace
+
+void SetDefaultObjcClassPrefix(const std::string& default_objc_class_prefix) {
+  g_prefix_mode.set_default_objc_class_prefix(default_objc_class_prefix);
+}
 
 bool UseProtoPackageAsDefaultPrefix() {
   return g_prefix_mode.use_package_name();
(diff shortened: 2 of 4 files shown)


src/google/protobuf/compiler/objectivec/objectivec_generator.cc
```diff
@@ -190,10 +190,15 @@ bool ObjectiveCGenerator::GenerateAll(
...
+      // The objc class prefix to use if specified by the command line invocation.
+      // The file option is always honored first if one is present.
+      SetObjcClassPrefix(options[i].second);
     } else if (options[i].first == "use_package_as_prefix") {
       // Controls how the symbols should be prefixed to avoid symbols
-      // collisions. The objc_class_prefix file option is always honored, this
-      // is just what to do if that isn't set. The available options are:
+      // collisions. The objc_class_prefix file option is always honored first and then
+      // the objc_class_prefix passed through objc_opt. This is is just what to do if either
+      // of these options are not set. set. The available options are:
```
thomasvl, line 201: s/set. set./set./

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -531,6 +539,11 @@ std::string FileClassPrefix(const FileDescriptor* file) {
     return file->options().objc_class_prefix();
   }
 
+  // If prefix is passed through objc_opt then accept that first.
```
thomasvl, line 542: I'd probably drop the `first` on this sentence, since it isn't _first_, the file option is _first_, this is just accepted next before the prefix mode.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -531,6 +539,11 @@ std::string FileClassPrefix(const FileDescriptor* file) {
...
+  if (!g_prefix_mode.objc_class_prefix().empty()) {
```
thomasvl, line 543: So this test means there's no way to use the generation option set to the empty string, as it would appear to always be unset.  The on case that would come up is if someones build setup also ends up setting the support for use package prefix, so I'm not sure how critical it is, but figured I should mention in.

Maybe we should tweak where the option is collected and have that error if an empty string was passed?

thomasvl (commented):
The fact that developers have to edit the files for the prefix has always sorta bugged me, so I'm not opposed to this.  Just want to make sure it is something we spec that can be supported long term.

I wonder if maybe we should tweak your proposed generation option name to be `default_objc_class_prefix` to better indicate when/how it would be used?

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
dnkoutso: will fix!

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dnkoutso: sg, will fix!

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dnkoutso: I agree. I can update to throw an error if the option is set with an empty string.

dnkoutso:
> I wonder if maybe we should tweak your proposed generation option name to be default_objc_class_prefix to better indicate when/how it would be used?

Yeah I did not like it myself either and I appreciate this take. I think we should change it to `default_objc_class_prefix` as it can easily differentiate it from the file option.

dnkoutso:
@thomasvl here's a thought:

Should we actually make the `objc_opt` option be honored _first_ and before the file option, if it is specified? I assume if we ship this we expect folks to be able to use it and will have to update their proto files to remove the file option in order to be able to use this new one.

thomasvl:
> @thomasvl here's a thought:
> 
> Should we actually make the `objc_opt` option be honored _first_ and before the file option, if it is specified? I assume if we ship this we expect folks to be able to use it and will have to update their proto files to remove the file option in order to be able to use this new one.

I was debating that after reading the issue and starting to look at your PR.  We could continue with the current PR and making it a _default_, and that doesn't prevent us from later adding a `forced_objc_class_prefix` (or maybe `override_objc_class_prefix`), but I'm not sure it has to be done at the same time.

I can see values for both, again, once it is officially provided, it has to be supported, so we'll need to think through if there are down sides to this…

Since the prefixes can be needed to avoid type collisions, I can see something that is an override becoming a _foot gun_ when some .proto author has two packages with a _User_ message, and all languages using proto packages keep them unique, and that author might have even included ObjC prefixes to keep them unique, but this generation option can't be used because it completely defeats the proto authors work to help developers out.

So yes, there are likely lots of cases it is safe and someone could use it, but then suddenly one day they get a new proto dependency or one gets updated, and they suddenly have to revise everything because we've lead them down a path that is sorta asking for this issue in the future.

Again, not completely opposed, but as an override, it likely need to be thought about a bit longer compared to a _default_ one.

dnkoutso:
> I was debating that after reading the issue and starting to look at your PR. We could continue with the current PR and making it a default, and that doesn't prevent us from later adding a forced_objc_class_prefix (or maybe override_objc_class_prefix), but I'm not sure it has to be done at the same time.

Sounds good! I will proceed with renaming it to default and update README etc.

dnkoutso:
@thomasvl updated to use `default_objc_class_prefix` generator option along with all other comments.

Please let me know if i should update anything else and thank you for the review and guidance!

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
```diff
@@ -190,10 +190,20 @@ bool ObjectiveCGenerator::GenerateAll(
...
           StripSuffixString(options[i].second, "/");
+    } else if (options[i].first == "default_objc_class_prefix") {
+      // The default objc class prefix to use if specified by the command line 
+      // invocation. The file option is always honored first if one is present.
+      std::string value = options[i].second;
+      if (value.empty()) {
+        *error = "error: default_objc_class_prefix cannot be empty.";
+        return false;
+      }
+      SetDefaultObjcClassPrefix(options[i].second);
```
thomasvl, line 201: s/options[i].second/value/

since you already fetched it.

dnkoutso: argh fixing.

dnkoutso: fixed

dnkoutso:
@thomasvl I looked into adding some tests...it seems it should be possible, let me know if its a requirement

dnkoutso:
Looks like one shard flaked?

thomasvl (approved):

thomasvl:
> @thomasvl I looked into adding some tests...it seems it should be possible, let me know if its a requirement

I'm ok without a specific test on this given the simplicity of it.

> Looks like one shard flaked?

Yea, some issue not related to your PR.

dnkoutso:
thank you!

dnkoutso:
@thomasvl this is actually causing issues.

It turns out that `g_prefix_mode` is a "global" mode for deriving objc class prefix. This means that when a proto depends on on another proto we are outputting the incorrect prefix for dependencies.

Do you have a place to store this for each proto file options (but not real proto file options)

dnkoutso:
I guess I can build it into `g_prefix_mode` but keep track of each file similarly to `is_package_exempted`

dnkoutso:
updating this now.

dnkoutso:
maybe this can be rearchitected into a file like `require_prefixes` because we can pass multiple proto files for generation and default_objc_class_prefix does not make sense?

dnkoutso:
Going to update this to be based from a file instead. The file can be 

<prefix>:proto

If one wants to use a "default" one they can just output the same prefix for all files.

dnkoutso:
This was reverted here https://github.com/protocolbuffers/protobuf/pull/9532

PR #9461, opened 2022-02-01
Integrate from Piper for C++, Java, and Python

deannagarcia:
Important: this integration must be merged, not squashed!

PR #9479, opened 2022-02-08

haberman:
Important: this integration must be merged, not squashed!

This is a continuation of https://github.com/protocolbuffers/protobuf/pull/9461


python/setup.py
```diff
@@ -301,12 +301,11 @@ def get_option_from_sys_argv(option_str):
       classifiers=[
           "Programming Language :: Python",
           "Programming Language :: Python :: 3",
```
acozzette, line 303: I think some setup.py changes got lost again here. The good news is that after this PR is merged, the problem should be solved going forward.

java/core/src/main/java/com/google/protobuf/ByteString.java
```diff
@@ -279,8 +316,13 @@ public int compare(ByteString former, ByteString latter) {
       };
 
   /**
+<<<<<<< HEAD
    * Returns a {@link Comparator} which compares {@link ByteString}-s lexicographically
    * as sequences of unsigned byte values between 0 and 255, inclusive.
+=======
+   * Returns a {@link Comparator} which compares {@link ByteString}-s lexicographically as sequences
```
acozzette, line 323: I think the resolution this conflict should be something like this:
```
   * Returns a {@link Comparator} which compares {@link ByteString}-s lexicographically as sequences
   * of unsigned byte values between 0 and 255, inclusive.
```

acozzette:
I saw that there are a few more merge conflicts that need to be resolved. If you just go to https://github.com/protocolbuffers/protobuf/pull/9479.patch you can find them by searching for `<<<<`.

python/setup.py
haberman: I fixed this.

java/core/src/main/java/com/google/protobuf/ByteString.java
haberman: Fixed.

acozzette (approved):

PR #9486, opened 2022-02-09
Ruby: rename max_recursion_depth to recursion_limit

acozzette:
This will help keep the terminology consistent with the other language
implementations.


haberman (approved):

acozzette:
Let me see about adding a conformance test, since that seems like it would be the most reliable way to make sure languages are enforcing the same recursion limit.

aleksandarabas:
NO NO NO, some people will mistake it as a rushed abbreviation for recursion limit case. 
But the stack depth limit which is imposed on the recursion process must be clearly distinguished from whatever custom property that doesn't require privilege escalation. This will help avoid ww3. thank you

aleksandarabas:
> Let me see about adding a conformance test, since that seems like it would be the most reliable way to make sure languages are enforcing the same recursion limit.

This is very plainly why you struggle with the phenomena of small government republicans. :D

PR #9488, opened 2022-02-10
Add osx aarch_64 to protoc-artifacts

yuzawa-san:
This is my attempt to add a real osx aarch_64 (aka M1, Apple Silicon) artifact. I can run locally on a recent x86_64 mac which is one of the environments specified in the protoc-artifacts/README.md.

It was not super clear how the build automation runs this or moves stuff around. I see the linux CI jobs in kokoro seem to call the build-protoc.sh, but the mac job seems to build it without this script. Per build-zip.sh it should already pick up the artifact, but is currently skipped since it does not exist. It also looks like it should appear on the github release artifacts. I assume in this PR (possibly incorrectly) that the artifact will get to the "right" place, so I updated the pom.xml, but I can hold back on that if you advise. I don't feel comfortable (and the README discourages) making kokoro modifications, so I'm not going to touch that stuff. And building out the CI to assert that these new artifacts in fact work on this new macs is more of a maintainer task.

There seems to be a few issues tracking this and if its accepted, I'll link to those.

Signed-off-by: James Yuzawa <jtyuzawa@gmail.com>

### DIFF
--- a/protoc-artifacts/README.md
+++ b/protoc-artifacts/README.md
@@ -128,6 +128,8 @@ target directory layout:
       + osx
         + x86_64
           protoc.exe
+        + aarch_64
+          protoc.exe
         + x86_32
           protoc.exe
       + windows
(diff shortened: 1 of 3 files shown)


acozzette:
@yuzawa-san Thank you for the pull request, but I don't think we will be able to accept this. The good news is that @deannagarcia is already working on this and has a solution in progress that doesn't rely on the build-protoc.sh script.

PR #9498, opened 2022-02-12
Add prefix_to_proto_package_mappings_path ObjC option.

dnkoutso:
Hey @thomasvl in the [previous PR](https://github.com/protocolbuffers/protobuf/pull/9476) we introduced the `default_objc_class_prefix` objc_opt.

However, upon using it it had a major flaw. We have protos that we generate this prefix for for namespacing and it is not always the same. Upon using this, we realized that objc class prefix passed was also used when generating code for dependencies which was causing compilation issues. In other words, it was applied globally. This might be OK if anyone decides to use the same prefix for all protos so maybe we can keep that option as well?

I figured anyone can generate a map with the same prefix across all proto files if they desire to do so.

This is basically a different solution to https://github.com/protocolbuffers/protobuf/issues/9469

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
@@ -190,6 +190,21 @@ bool ObjectiveCGenerator::GenerateAll(
       // header search path since the generate #import will be more complete.
       generation_options.runtime_import_prefix =
           StripSuffixString(options[i].second, "/");
+    } else if (options[i].first == "prefix_to_proto_package_mappings_path") {
+      // Path to use for when loading the objc class prefix mappings to use.
+      // The `objc_class_prefix` file option is always honored first if one is present.
+      // This option also has precedent over the use_package_as_prefix option.
+      //
+      // The format of the file is:
+      //   - An entry is a line of "package=prefix".
+      //   - Comments start with "#".
+      //   - A comment can go on a line after a expected package/prefix pair.
+      //     (i.e. - "package=prefix # comment")
+      //   - For files that do NOT have a proto package (not recommended), an
+      //     entry can be made as "no_package:PATH=prefix", where PATH is the
+      //     path for the .proto file.
+      //
+      SetPrefixToProtoPackageMappingsPath(options[i].second);
     } else if (options[i].first == "use_package_as_prefix") {
       // Controls how the symbols should be prefixed to avoid symbols
       // collisions. The objc_class_prefix file option is always honored, this
--- a/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
@@ -95,9 +95,28 @@ class SimpleLineCollector : public LineConsumer {
   std::unordered_set<std::string>* set_;
 };
 
+class ExpectedPrefixesCollector : public LineConsumer {
+ public:
+  ExpectedPrefixesCollector(std::map<std::string, std::string>* inout_package_to_prefix_map)
+      : prefix_map_(inout_package_to_prefix_map) {}
+
+  virtual bool ConsumeLine(const StringPiece& line, std::string* out_error) override;
+
+ private:
+  std::map<std::string, std::string>* prefix_map_;
+};
+
 class PrefixModeStorage {
  public:
   PrefixModeStorage();
+
+  const std::string prefix_to_proto_package_mappings_path() const { return prefix_to_proto_package_mappings_path_; }
+  void set_prefix_to_proto_package_mappings_path(const std::string& path) {
+    prefix_to_proto_package_mappings_path_ = path;
+    prefix_to_proto_package_map_.clear();
+  }
+
+  std::string prefix_from_proto_package_mappings(const FileDescriptor* file);
 
   bool use_package_name() const { return use_package_name_; }
   void set_use_package_name(bool on_or_off) { use_package_name_ = on_or_off; }
@@ -116,6 +135,8 @@ class PrefixModeStorage {
 
  private:
   bool use_package_name_;
+  std::map<std::string, std::string> prefix_to_proto_package_map_;
+  std::string prefix_to_proto_package_mappings_path_;
   std::string exception_path_;
   std::string forced_prefix_;
   std::unordered_set<std::string> exceptions_;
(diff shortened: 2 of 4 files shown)


src/google/protobuf/compiler/objectivec/objectivec_generator.cc
```diff
@@ -190,19 +190,21 @@ bool ObjectiveCGenerator::GenerateAll(
       // header search path since the generate #import will be more complete.
       generation_options.runtime_import_prefix =
           StripSuffixString(options[i].second, "/");
-    } else if (options[i].first == "default_objc_class_prefix") {
```
dnkoutso, line 193: I can also keep this option.

I figured we can delete it because I dont think a protobuf version has shipped with it.

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
```diff
@@ -315,7 +316,6 @@ class PROTOC_EXPORT ImportWriter {
                                   const std::string& runtime_import_prefix,
                                   bool default_cpp_symbol = false);
 
- private:
```
dnkoutso, line 318: I wanted to reuse this collector since the format of the `objc_class_prefix_mappings` file is the same as the one for framework mappings.

dnkoutso:
no worries, thank you so much for your time @thomasvl !

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
thomasvl: If we drop it, I'd suggesting doin that in a clean PR/commit so git history is a little more clear.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -144,6 +150,38 @@ PrefixModeStorage::PrefixModeStorage() {
   }
 }
 
+std::string PrefixModeStorage::objc_class_prefix(const FileDescriptor* file) {
```
thomasvl, line 153: should this take a reference so it don't have to worry about getting a nullptr?

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
```diff
@@ -190,19 +190,21 @@ bool ObjectiveCGenerator::GenerateAll(
...
-      if (value.empty()) {
-        *error = "error: default_objc_class_prefix cannot be empty.";
-        return false;
-      }
-      SetDefaultObjcClassPrefix(value);
+    } else if (options[i].first == "objc_class_prefix_mappings_path") {
+      // Path to use for when loading the objc class prefix mappings to use.
+      // The `objc_class_prefix` option is always honored first if one is present.
+      //
+      // The format of the file is:
```
thomasvl, line 197: Since you are keying this by files, it means you are explicitly allowing a proto package to be split between multiple prefixes, which is the opposite direction from what the library was generally trying to encourage with the prefix validation support.  Is there a reason you went this directions vs. trying to keep package on a single prefix?

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
dnkoutso: I can update it, this was mostly copy/pasta from other invocations.

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
dnkoutso: Our problem is that we use sub-directories to organize protos and in some cases the proto filenames are the same.

e.g. company/featureA/service.proto and company/featureB/service.proto.

The .proto files have a different `package` declaration to separate those but using a single/global prefix causes duplicate class names to be generated. So we generate our prefix based on the `package` declared in order to avoid duplicates.

thomasvl: Sorry, I'm not following.  I was saying you'd use a file that is the same as the prefix validation one, so _package_ = _prefix_ lines.  For what you describe, that sounds like it would work just fine.

If the two files were in the _same_ proto package, then it wouldn't work.

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
dnkoutso: sounds good

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
dnkoutso: ah I see what you are saying yes this seems like it would for our case as well. I can update this.

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
dnkoutso: Reverted here in a clean PR https://github.com/protocolbuffers/protobuf/pull/9532

src/google/protobuf/compiler/objectivec/objectivec_generator.cc
dnkoutso: done!

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -95,10 +95,29 @@ class SimpleLineCollector : public LineConsumer {
   std::unordered_set<std::string>* set_;
 };
 
+class ExpectedPrefixesCollector : public LineConsumer {
```
dnkoutso, line 98: This was just moved at the top for visibility.

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
```diff
@@ -140,6 +161,44 @@ PrefixModeStorage::PrefixModeStorage() {
   }
 }
 
+std::string PrefixModeStorage::prefix_from_proto_package_mappings(const FileDescriptor* file) {
+  if (!file) {
```
dnkoutso, line 165: We actually do need the `file` here as we perform some logic below to figure out the package to use.

dnkoutso:
@thomasvl I updated all logic! Let me know if there is anything else for me to change or address!

src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
thomasvl: I think I had commented not about the passing of `file`, but the passing of it by reference so it can't be `nullptr`.

dnkoutso: ah

thomasvl (approved):

PR #9506, opened 2022-02-15
Revert "Add an option to preserve proto names in JsonFormatter"

jtattermusch:
Reverts protocolbuffers/protobuf#6307

- there is a large white-space only diff that obscures the real changes, which is really bad for maintenance. Let's reintroduce the PR without the whitespace changes so that the correct diff is shown in github UI.
- there's been some other concerns with the PR (see comment thread on https://github.com/protocolbuffers/protobuf/pull/6307).

We'll be happy to accept the PR again once the concerns from the review are addressed and once the diff looks fine.

@ObsidianMinor

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
@@ -663,13 +663,6 @@ namespace Google.Protobuf
         }
 
         [Test]
-        public void WriteValue_Message_PreserveNames()
-        {
-            var value = new TestAllTypes { SingleInt32 = 100, SingleInt64 = 3210987654321L };
-            AssertWriteValue(value, "{ 'single_int32': 100, 'single_int64': '3210987654321' }", JsonFormatter.Settings.Default.WithPreserveProtoFieldNames(true));
-        }
-
-        [Test]
         public void WriteValue_List()
         {
             var value = new RepeatedField<int> { 1, 2, 3 };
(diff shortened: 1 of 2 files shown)


jskeet:
Hmm... creating a new commit to revert will only help fix some of the issues by merging the giant whitespace change. `git annotate` will still show "this was all changed in a single commit". Basically it's hard to get back to a nice diff without rewriting history at this point :(

I would suggest that we should probably:

- Check what the actual whitespace changes are (e.g. was it LF -> CRLF, CRLF -> LF or something different?)
- Work out what we *want* the whitespace to be (check other files)
- Make other changes to `JsonFormatter` as desired. (I don't think it's anything semantic, just some formatting/description changes.)

jtattermusch:
> 

Fair enough, I'll close the PR. Filed https://github.com/protocolbuffers/protobuf/issues/9526 to address the CRLF inconsistencies.

> Hmm... creating a new commit to revert will only help fix some of the issues by merging the giant whitespace change. `git annotate` will still show "this was all changed in a single commit". Basically it's hard to get back to a nice diff without rewriting history at this point :(
> 
> I would suggest that we should probably:
> 
> * Check what the actual whitespace changes are (e.g. was it LF -> CRLF, CRLF -> LF or something different?)
> * Work out what we _want_ the whitespace to be (check other files)
> * Make other changes to `JsonFormatter` as desired. (I don't think it's anything semantic, just some formatting/description changes.)

PR #9513, opened 2022-02-16
[Ruby] allow encode json options to be an object that responds to to_hash

### LINKED ISSUE #9500: Calling `to_json` on a Hash that contains a protobuf message value raises an ArgumentError
Version: 3.19.4
Language: Ruby

Macos 12.0.1

Ruby 3.1.0

**What did you do?**

When calling `to_json` on a Hash or Array that contains a protobuf message an `ArgumentError` is raised (`Expected hash arguments. (ArgumentError)`).

I've created this minimal reproducable example:

``` ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'google-protobuf', '= 3.19.4'
end

require 'json'

  add_file('msg.proto', syntax: :proto3) do
    add_message 'Msg' do
      optional :some_field, :string, 1, json_name: 'some_field'
    end
  end
end

Msg = ::Google::Protobuf::DescriptorPool.generated_pool.lookup('Msg').msgclass
msg = Msg.new(some_field: 'some_value')

puts msg.to_json
puts({ msg: msg }.to_json)
```

```
{"some_field":"some_value"}
{"msg": {"some_field": "some_value"}}
```

**What did you see instead?**

```
{"some_field":"some_value"}
/Users/luka/.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/google-protobuf-3.19.4/lib/google/protobuf/message_exts.rb:44:in `encode_json': Expected hash arguments. (ArgumentError)
	from /Users/luka/.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/google-protobuf-3.19.4/lib/google/protobuf/message_exts.rb:44:in `to_json'
	from bug.rb:24:in `to_json'
	from bug.rb:24:in `<main>'
```

lukad:
fixes #9500

### DIFF
--- a/ruby/tests/common_tests.rb
+++ b/ruby/tests/common_tests.rb
@@ -870,6 +870,9 @@ module CommonTests
 
     decoded_msg = Google::Protobuf.decode_json(proto_module::TestMessage, encoded_msg)
     assert_equal proto_module::TestMessage.decode_json(m.to_json), decoded_msg
+
+    assert_equal [m].to_json, Google::Protobuf.encode_json([m])
+    assert_equal proto_module::TestMessage.decode_json([m.to_json].first), decoded_msg
   end
 
   def test_def_errors
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -1141,7 +1141,11 @@ static VALUE Message_encode_json(int argc, VALUE* argv, VALUE klass) {
   if (argc == 2) {
     VALUE hash_args = argv[1];
     if (TYPE(hash_args) != T_HASH) {
-      rb_raise(rb_eArgError, "Expected hash arguments.");
+      if (RTEST(rb_funcall(hash_args, rb_intern("respond_to?"), 1, rb_str_new2("to_h")))) {
+        hash_args = rb_funcall(hash_args, rb_intern("to_h"), 0);
+      } else {
+        rb_raise(rb_eArgError, "Expected hash arguments.");
+      }
     }
 
     if (RTEST(rb_hash_lookup2(hash_args,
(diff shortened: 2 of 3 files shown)


haberman:
What is an example of how this arises in real code?

What is a `json::Ext::Generator::State` object, and why is your code passing that in instead of `Hash`?

lukad:
We ran into this in a Rails application that serves a JSON api and calls out to a GRPC Service. The JSON response is supposed to include the protobuf message.

``` ruby
msg = protobuf_message # returned from a GRPC service
{
  msg: protobuf_message,
  # and other fields
}.to_json
```

[`json::Ext::Generator::State`](https://ruby-doc.org/stdlib-3.1.1/libdoc/json/rdoc/JSON/Ext/Generator/State.html) holds Ruby's json generator options (like identation or space). This will always be created when calling `to_json` on a `Hash` and it will be passed down as opts to all subsequent `to_json` calls on the Hash's values values values as `opts`.

It is not safe for protobuf to assume that the `opts` passed to a message's `to_json` are always a hash.

Links:
* https://github.com/ruby/ruby/blob/v3_1_1/ext/json/generator/generator.c#L414
* https://github.com/ruby/ruby/blob/v3_1_1/ext/json/generator/generator.h#L87
* https://github.com/ruby/ruby/blob/v3_1_1/ext/json/generator/generator.c#L1031

ruby/tests/common_tests.rb
```diff
@@ -870,6 +870,9 @@ def test_protobuf_encode_decode_json_helpers
 
     decoded_msg = Google::Protobuf.decode_json(proto_module::TestMessage, encoded_msg)
     assert_equal proto_module::TestMessage.decode_json(m.to_json), decoded_msg
+
+    assert_equal [m].to_json, Google::Protobuf.encode_json([m])
```
haberman, line 874: Maybe I am missing something, but this test doesn't seem to exercise the new code. We need a test that calls `encode_json` with a second parameter that response to `to_h` but is not itself a `Hash`.

Issue #9492, opened 2022-02-10
arenaz_sampler.h not included in install list


LikeSmith:
**What version of protobuf and what language are you using?**
Version: master (3.19.4)
Language: C++

**What operating system (Linux, Windows, ...) and version?**
linux 5.15.15-76051515-generic

**What runtime / compiler are you using (e.g., python version or gcc version)**
clang version 13.0.1

**What did you do?**
1. clone protobuf from git
2. make build directory
3. call cmake ../cmake from build directory
4. call make in build directory
5. call make install in build directory
6. compile my .proto files
7. try to compile generated .cc files

**What did you expect to see**
successful compilation

**What did you see instead?**
protobuf header failed to include another protobuf header

Error:
install location/google/protobuf/arena_impl.h:48:10: fatal error: google/protobuf/arenaz_sampler.h: No such file or directory

**Anything else we should know about your project / environment**
copying arenaz_sampler.h from src/google/protobuf to the install location allows for successful compilation of my code.  Is this file somehow missed when marking headers for installation?

vitsensei:
I also have this problem, any update on this?

PR #9528, opened 2022-02-18

dlj-NaN:


dlj-NaN:
This will address #9492.

PR #9529, opened 2022-02-18
cmake: Set correct sonames for libprotobuf-lite.so and libprotoc.so (#8635)

Arfrever:
Soname was set for libprotobuf.so in commit a9cf69a0ed7766f56976e48a0c96300e94433511,
but similar changes for libprotobuf-lite.so and libprotoc.so were missed.

Fixes: #8635


acozzette:
Thanks, @Arfrever.

PR #9530, opened 2022-02-19
Add protoc-gen-validate to third-party doc

Lunkentuss:
Details about the plugin is found [here](https://github.com/envoyproxy/protoc-gen-validate).


elharo:
Interesting. This seems like it breaks existing tooling by changing the protobuf schema syntax. Let me know if I'm wrong about that.

PR #9543, opened 2022-02-24
Make the ObjC support conditional to darwin.

### LINKED ISSUE #9392: Disable check for Objective C compiler (AC_PROG_OBJC) on non-macOS systems
Autotools build system currently unconditionally checks for Objective C compiler:
https://github.com/protocolbuffers/protobuf/blob/0ac74b8126b76498075f43c37603d67a15d8d205/configure.ac#L83

However later there is:
https://github.com/protocolbuffers/protobuf/blob/0ac74b8126b76498075f43c37603d67a15d8d205/configure.ac#L217-L224

`OBJC_CONFORMANCE_TEST` is used in `conformance/Makefile.am`.
The only usage of Objective C compiler is for compiling source files which are used only when `OBJC_CONFORMANCE_TEST` is true.

This means that aforementioned check for Objective C compiler is unnecessary on non-macOS systems (when `$target_os` != `darwin*`).

Fix:
```
--- configure.ac
+++ configure.ac
@@ -80,7 +80,7 @@
 ACX_USE_SYSTEM_EXTENSIONS
 m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 AM_CONDITIONAL(GCC, test "$GCC" = yes)   # let the Makefile know if we're gcc
-AC_PROG_OBJC
+AS_CASE([$target_os], [darwin*], [AC_PROG_OBJC], [AM_CONDITIONAL([am__fastdepOBJC], [false])])
 
 # test_util.cc takes forever to compile with GCC and optimization turned on.
 AC_MSG_CHECKING([C++ compiler flags...])
```

thomasvl:
Fixes #9392

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,7 +80,7 @@ AC_LANG([C++])
 ACX_USE_SYSTEM_EXTENSIONS
 m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 AM_CONDITIONAL(GCC, test "$GCC" = yes)   # let the Makefile know if we're gcc
-AC_PROG_OBJC
+AS_CASE([$target_os], [darwin*], [AC_PROG_OBJC], [AM_CONDITIONAL([am__fastdepOBJC], [false])])
 
 # test_util.cc takes forever to compile with GCC and optimization turned on.
 AC_MSG_CHECKING([C++ compiler flags...])


thomasvl:
Clearly _draft_ incase that's why kokoro wasn't testing this yet.

thomasvl:
Looks like the snippet from the original issue works (passes kokoro, and my local mac build still has the objc support).

@Arfrever fyi

acozzette (approved):
Looks good to me, but we will likely be deleting the autotools build soon.

Arfrever:
@acozzette

If you are planning to recommend using cmake build system, why is cmake build system still documented as being only for Windows?

acozzette:
Our documentation for the CMake build is way out of date. We need to update it to say that it is for all platforms.

thomasvl:
src/README.md also seems to steer away from cmake.

Issue #6113, opened 2019-05-08
cmake build system does not make linker use version scripts


Arfrever:
Only Autotools build system uses version scripts:
```
$ grep -r -- --version-script *
configure.ac:LDFLAGS="$LDFLAGS -Wl,--version-script=conftest.map"
src/Makefile.am:libprotobuf_lite_la_LDFLAGS += -Wl,--version-script=$(srcdir)/libprotobuf-lite.map
src/Makefile.am:libprotobuf_la_LDFLAGS += -Wl,--version-script=$(srcdir)/libprotobuf.map
src/Makefile.am:libprotoc_la_LDFLAGS += -Wl,--version-script=$(srcdir)/libprotoc.map
```
Each of 3 version scripts (`src/libprotobuf.map`, `src/libprotobuf-lite.map`, `src/libprotoc.map`) is currently the same:
```
{
  global:
    extern "C++" {
      *google*;
    };

  local:
    *;
};
```
Not passing `-Wl,--version-script=...` option causes that symbols other than `*google*` are not hidden. See commit 13d165de9e737cec750a2ab3d1ef68bb3f276c48 by @afrantzis.

PR #9545, opened 2022-02-24
cmake: Use linker version scripts

Arfrever:
Autotools build system already uses linker version scripts since
commit 13d165de9e737cec750a2ab3d1ef68bb3f276c48.

Fixes: #6113

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Minimum CMake required
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.10)
 
 if(protobuf_VERBOSE)
   message(STATUS "Protocol Buffers Configuring...")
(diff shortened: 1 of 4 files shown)


cmake/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # Minimum CMake required
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.18)
```
Arfrever, line 2: In case anybody is curious, `CheckLinkerFlag` cmake module (used below) is documented in https://cmake.org/cmake/help/latest/module/CheckLinkerFlag.html that:
> New in version 3.18.

acozzette: One of our CI builds is breaking because it's still using CMake 3.10.1. Is there any easy workaround that wouldn't require the `CheckLinkerFlag` module?

acozzette: I just checked and it doesn't appear to be easy to upgrade our CMake version, because we're using a Windows VM maintained by another team. I think we should probably just enable the behavior conditionally, only if `CMAKE_VERSION` is greater than or equal to 3.18 (we already do something like this below).

acozzette: Actually would you mind setting the minimum version to 3.10.1? Let's do that and see if anyone complains. Maybe we can try to gradually increase the minimum version over time so that we can start to rely on newer CMake features.

Arfrever: I added fallback code using `check_cxx_source_compiles` function from `CheckCXXSourceCompiles` module.

acozzette:
Thanks, @Arfrever!

veblush:
Is it really required to bump the cmake version to 3.10? FYI, gRPC still supports 3.5.1.

acozzette:
Let me send out a pull request to lower it to 3.5.1.

PR #9552, opened 2022-02-25
[ObjC] Rename `proto_package_to_prefix_mappings_path` to `package_to_prefix_mappings_path`.

dnkoutso:
The file format is actually the reverse of the initial option so we now rename the option to match the file format to avoid confusion.


thomasvl (approved):
Thanks, sorry I didn't notice when we revised the the PR initially.

PR #9555, opened 2022-02-26
chore: gitignore ephemeral files that clangd uses

devjgm:
This PR tells git to ignore ephemeral files that are created when
developers use clangd.

clangd needs a `compile_commands.json` file at the root of the repo (or
in a `build/` directory) to tell it how to compile each file. CMake can
generate this file with something like the following:

```console
$ cmake -Scmake -B.build \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
  -Dprotobuf_BUILD_TESTS=OFF
$ ln -sf .build/compile_commands.json .
```

Then clangd will compile and index files into the `.cache` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -222,3 +222,7 @@ _build/
 
 # BenchmarkDotNet
 BenchmarkDotNet.Artifacts/
+
+# Clangd uses these common ephemeral files
+.cache
+compile_commands.json


acozzette (approved):

acozzette:
I kicked off the tests but just remembered that they're probably unnecessary since none of them are likely to be affected by the `.gitignore` file.

devjgm:
> I kicked off the tests but just remembered that they're probably unnecessary since none of them are likely to be affected by the `.gitignore` file.

That's OK. I've been surprised by comment-only changes breaking builds before, so let 'em run ;-)

devjgm:
Thanks @acozzette. Should I merge or do you do that? What's the preferred process?

acozzette:
@devjgm Sure, feel free to merge it. It doesn't matter that much but usually the author does the merge if the author is on the team.

PR #9565, opened 2022-03-01
Simplify imports of the WKTs within the library itself.

thomasvl:
Overdue followup to https://github.com/protocolbuffers/protobuf/pull/7173

Since all the files are at the same level as the runtime headers, there is no
need for things to be framework based imports, they should all just work like
the other headers do.

- Directly generate the bundled header imports into the preamble section when
  generating for a bundled proto.
- Update the preamble generation to skip the CPP wrapper when generating for a
  bundled proto file.
- Regenerate the WKTs.
- Update GPBProtocolBuffer.h to also skip the CPP wrapping.

GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS in the podspec and non bundled files still
has to exist because that comes into play for those files to find the runtime
headers.


dmaclach (approved):

thomasvl:
Updated to include GPBWellKnownTypes.h being simplified also.

Issue #9502, opened 2022-02-14
To get the descriptor/field name in OneOf struct which has been set in CPP


Guancheng-Li:
**What language does this apply to?**
proto3 + CPP

**Describe the problem you are trying to solve.**
In python, there is an api (`WhichOneOf`) to get the initialized field name: https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/message.py#L295
The is a similar request in CPP too.

**Describe the solution you'd like**
We can use the reflection doing this in CPP.
I locally solved it and want to add it to the official codebase.

**Describe alternatives you've considered**
Return the descriptor is more useful than the field name itself especially in cpp, so we return the ptr of descriptor

**Additional context**

PR #9567, opened 2022-03-02
Add WhichOneof util in cpp to get the set oneof field descriptor

Guancheng-Li:
**Description**
Address this issue: https://github.com/protocolbuffers/protobuf/issues/9502

In python, there is an API (WhichOneOf) to get the initialized field name: https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/message.py#L295

This function is very useful for us, but it does not implement in CPP.
The API in python implemented in the file `message.py`, I've read the message related code in CPP,
and get a conclusion that this function may not proper to put in the following paths:
```
src/google/protobuf/message.h
src/google/protobuf/message_lite.h
src/google/protobuf/descriptor.h
```

The current way may be a better choice for a "helper function".

### DIFF
--- a/src/google/protobuf/util/oneof_util.cc
+++ b/src/google/protobuf/util/oneof_util.cc
@@ -1,0 +1,58 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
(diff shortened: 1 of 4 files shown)


Guancheng-Li:
@googlebot I signed it!

Guancheng-Li:
Thank you for the quick and detailed reply, creating this PR just want to make it more friendly to multilingual users.

I wonder if the Python code and CPP code(or other languages) maintained by one team?

I'm asking this because in diff language(I guess they should be the most common used) the APIs are not totally the same. And it looks we have diff standard for them.

Of course, sometimes it make sense, considering the language specific issues, or maybe some APIs are developing in diff time.

But if the issues are not language specified, I guess an unified API is much friendly to users.

Still be happy to provide Pull-Request in the future if I can help.

acozzette:
Our team maintains both C++ and Python protobuf (and a bunch of other languages as well). We do usually try to keep things as consistent as possible between language implementations. In this case I think the functionality is fairly similar although the naming is different. The Python `WhichOneof` is roughly equivalent to `Reflection::GetOneofFieldDescriptor()` in C++, though the Python one operates on string field names and C++ uses descriptors instead.

Guancheng-Li:
> Our team maintains both C++ and Python protobuf (and a bunch of other languages as well). We do usually try to keep things as consistent as possible between language implementations. In this case I think the functionality is fairly similar although the naming is different. The Python `WhichOneof` is roughly equivalent to `Reflection::GetOneofFieldDescriptor()` in C++, though the Python one operates on string field names and C++ uses descriptors instead.

Thanks @acozzette , I understand what you said in the quote, shall we close this PR and related issue?

PR #9568, opened 2022-03-02
Add a generation option to control use of forward declarations in headers.

thomasvl:
Swift importing ObjC drops methods/properties if the type is only a forward
declaration since the type is incomplete. Historically the generator has always
use forward declarations to reduce how much will have rebuild when a proto file
does change; but that puts it at odds with Swift. If ObjC Protos end up spanning
Swift modules, the Swift import behavior could become a problem; so this option
provides a control for the behavior. The current behavior is to continue forward
declarations, but eventually the default will be changed.

Generate the WKTs using imports instead of forward decls.


src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -239,16 +247,26 @@ void FileGenerator::GenerateHeader(io::Printer* printer) {
...
     const std::string header_extension(kHeaderExtension);
-    for (int i = 0; i < file_->public_dependency_count(); i++) {
-      import_writer.AddFile(file_->public_dependency(i), header_extension);
+    if (headers_use_forward_declarations) {
+      // #import any headers for "public imports" in the proto file.
+      for (int i = 0; i < file_->public_dependency_count(); i++) {
+        import_writer.AddFile(file_->public_dependency(i), header_extension);
+      }
+    } else {
+      for (int i = 0; i < file_->dependency_count(); i++) {
```
thomasvl, line 267: So there are some edge cases here.

If the proto import is only for an enum and the enum is the value of a repeated field or a map, then we don't really need the types because there isn't covariance for those.

Similarly, if the imported types are for extensions, then the ObjC api doesn't need the types here, as they don't end up in the generated code.

So this code path could be much more complex and figure out what headers are really needed and only import those, but the complexity doesn't seem worth it.

thomasvl:
@dmaclach fyi

thomasvl:
> Can you add me as a reviewer? I can't seem to do it for some reason...

I think this repo/project locks down reviewers.

objectivec/README.md
```diff
@@ -218,6 +218,21 @@ supported keys are:
     helps prepare folks before they end up using a lot of protos and getting a
     lot of collisions.
 
+  * `headers_use_forward_declarations`: The `value` for this can be `yes` or
+    `no`, and indicates if the generated headers for use forward declarations
```
dmaclach, line 222: generated headers use forward declarations

objectivec/README.md
```diff
@@ -218,6 +218,21 @@ supported keys are:
...
+    for Message and Enum types from other .proto files or if they files should
```
dmaclach, line 223: if the files

objectivec/README.md
```diff
@@ -218,6 +218,21 @@ supported keys are:
...
+    be imported into the generated header instead.
+
+    By using forward declarations, less code is likely to recompile when the
+    files do change, but Swift generally doesn't like forward declarations and
+    will fail to include properties when that concrete definition of the type is
```
dmaclach, line 228: when the concrete definition

objectivec/README.md
```diff
@@ -218,6 +218,21 @@ supported keys are:
...
+    know at import time. If your proto usages span modules, this can be a
+    problem.
+
+    `headers_use_forward_declarations` current defaults to `yes` (existing
```
dmaclach, line 232: currently defaults

objectivec/README.md
```diff
@@ -218,6 +218,21 @@ supported keys are:
...
 
+  * `headers_use_forward_declarations`: The `value` for this can be `yes` or
+    `no`, and indicates if the generated headers for use forward declarations
+    for Message and Enum types from other .proto files or if they files should
+    be imported into the generated header instead.
+
+    By using forward declarations, less code is likely to recompile when the
+    files do change, but Swift generally doesn't like forward declarations and
+    will fail to include properties when that concrete definition of the type is
+    know at import time. If your proto usages span modules, this can be a
```
dmaclach, line 229: known at

src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -239,16 +247,26 @@ void FileGenerator::GenerateHeader(io::Printer* printer) {
       "\n",
       "google_protobuf_objc_version", StrCat(GOOGLE_PROTOBUF_OBJC_VERSION));
 
-  // #import any headers for "public imports" in the proto file.
+  // The bundled protos (WKTs) force off the use of forward declarations.
```
dmaclach, line 250: remove "off"

src/google/protobuf/compiler/objectivec/objectivec_file.cc
dmaclach: agreed.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -363,27 +377,34 @@ void FileGenerator::GenerateSource(io::Printer* printer) {
   std::vector<const FileDescriptor*> deps_with_extensions;
   CollectMinimalFileDepsContainingExtensions(file_, &deps_with_extensions);
 
+  // The bundled protos (WKTs) force off the use of forward declarations.
```
dmaclach, line 380: remove "off"

dmaclach (approved):
Approval.. all documentation nits I believe.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
thomasvl: rephrased, but it is off, they force imports instead of fwd decls.

src/google/protobuf/compiler/objectivec/objectivec_file.cc
thomasvl: rephrased.

PR #9571, opened 2022-03-03
[PHP] fix PHP build system

zeriyoshi:
protobuf is not compatible with the PHP core build system. If you try to build PHP built-in protobuf, you will get the following error

```shell
$ docker run --rm -it alpine:latest
# apk add --no-cache ca-certificates curl tar xz openssl git curl autoconf automake libtool make g++ unzip
# mkdir "/php"
# cd "/php"
# curl -fsSL "https://www.php.net/distributions/php-8.0.16.tar.xz" | tar Jxvf - --strip-components=1
# git clone --depth=1 --branch=v3.19.4 --recursive "https://github.com/protocolbuffers/protobuf" "/protobuf"
# cd "/protobuf"
# ./autogen.sh
# ./configure
# make -j$(nproc)
# cp -R "./php/ext/google/protobuf" "/php/ext"
# cd "/php"
# ./buildconf --force
# ./configure --disable-all --enable-cli --enable-protobuf
# make -j$(nproc)
main/internal_functions_cli.c:40:2: error: 'phpext_protobuf_ptr' undeclared here (not in a function); did you mean 'phpext_date_ptr'?
   40 |  phpext_protobuf_ptr,
      |  ^~~~~~~~~~~~~~~~~~~
      |  phpext_date_ptr
```

This PR will be minimally supported and adapted to the PHP build system.

### DIFF
--- a/php/ext/google/protobuf/php_protobuf.h
+++ b/php/ext/google/protobuf/php_protobuf.h
@@ -1,0 +1,40 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#ifndef PHP_PROTOBUF_H
+# define PHP_PROTOBUF_H
+
+# ifdef HAVE_CONFIG_H
+#  include "config.h"
+# endif
+
+extern zend_module_entry protobuf_module_entry;
+# define phpext_protobuf_ptr &protobuf_module_entry
+
+#endif	/* PHP_PROTOBUF_H */
--- a/php/ext/google/protobuf/php_protobuf.h
+++ b/php/ext/google/protobuf/php_protobuf.h
@@ -2,6 +2,11 @@
 // Copyright 2022 Google Inc.  All rights reserved.
 // https://developers.google.com/protocol-buffers/
 //
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+// ! THIS FILE ONLY APPROACHING IN-TREE PHP EXTENSION BUILD !
+// ! DOES NOT USE NORMALLY.                                 !
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+//
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
 // met:
--- a/php/ext/google/protobuf/php_protobuf.h
+++ b/php/ext/google/protobuf/php_protobuf.h
@@ -1,5 +1,5 @@
 // Protocol Buffers - Google's data interchange format
-// Copyright 2008 Google Inc.  All rights reserved.
+// Copyright 2022 Google Inc.  All rights reserved.
 // https://developers.google.com/protocol-buffers/
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


php/ext/google/protobuf/php_protobuf.h
```diff
@@ -0,0 +1,40 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
```
elharo, line 2: // Copyright 2022 Google LLC

is the new form we use here

haberman:
> This PR will be minimally supported and adapted to the PHP build system.

Could you explain more what you mean by this?

I'm reluctant to add something to our repo that is only minimally supported. It seems like the kind of thing that could work differently for different PHP versions, and would need more tweaking in the future.

Do other third-party PHP extensions support this? Can you point to an example?

Since it's such a small amount of code, I'm inclined to think it would be better to patch via a `.patch` file instead.

zeriyoshi:
@haberman 
PHP supports building extensions into the body. However, it requires the appropriate header files.
https://www.php.net/manual/en/install.pecl.static.php

This PR adds that header file.

Normally, header files would include declarations such as MINIT, but we did not add them because they are not needed in this case.

See the PHP extension skeleton for more information:
https://github.com/php/php-src/tree/master/ext/skeleton

You may be right that patches are better.

php/ext/google/protobuf/php_protobuf.h
zeriyoshi: fixed thanks

zeriyoshi:
@acozzette I fixed EXTRA_DIST, please readd `kokoro:run`

haberman:
Can you point to any other popular PHP extensions that are not distributed with PHP but include this `zend_module_entry` line to allow for in-tree builds?

php/ext/google/protobuf/php_protobuf.h
```diff
@@ -0,0 +1,40 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#ifndef PHP_PROTOBUF_H
+# define PHP_PROTOBUF_H
+
+# ifdef HAVE_CONFIG_H
+#  include "config.h"
+# endif
+
+extern zend_module_entry protobuf_module_entry;
```
haberman, line 37: Will this file get included by PHP directly? Is that why it needs to be its own file?

Is the name of this file significant? Does it need to be named `php_protobuf.h`?

php/ext/google/protobuf/protobuf.c
```diff
@@ -33,6 +33,8 @@
 #include <php.h>
 #include <Zend/zend_interfaces.h>
 
+#include "php_protobuf.h"
```
haberman, line 36: Why does this need to be included here? Will one of the other macros reference `phpext_protobuf_ptr`?

php/ext/google/protobuf/php_protobuf.h
```diff
@@ -0,0 +1,40 @@
...
+# define phpext_protobuf_ptr &protobuf_module_entry
```
haberman, line 38: Does this work across a wide variety of PHP versions, or will it need to be changed from version to version?

php/ext/google/protobuf/php_protobuf.h
```diff
@@ -0,0 +1,40 @@
...
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
haberman, line 29: Please add a note that this file exists only for the purpose of in-tree builds, and is unused by normal builds of protobuf. Also it is user-supported, and we do not have tests for in-tree builds.

php/ext/google/protobuf/php_protobuf.h
zeriyoshi: No, not included. This file always exists only in the protobuf repository.

Names are important. The PHP build system automatically references header files according to the naming conventions.

php/ext/google/protobuf/php_protobuf.h
zeriyoshi: This mechanism has been around for a long time and should work with all supported PHP versions.

php/ext/google/protobuf/protobuf.c
zeriyoshi: Oops, Deleted. Thanks

zeriyoshi:
> Can you point to any other popular PHP extensions that are not distributed with PHP but include this zend_module_entry line to allow for in-tree builds?

Most PECL extensions meet that requirement.
I think it's a good thing that in-tree builds are available, don't you?

zeriyoshi:
@acozzette sorry, i fixed Makefile, please `kokoro:run` again.

php/ext/google/protobuf/php_protobuf.h
zeriyoshi: fixed, thanks

zeriyoshi:
@acozzette 
Re-fixing and rebased, please add `kokoro:run` again, really sorry.

zeriyoshi:
@acozzette 
It appears that the mirror acquisition has failed.
I don't know the detailed behavior of kokoro, but how can I solve this problem?

```
+ docker run --rm -e 'BUILD_COMMANDS=build_wheel protobuf/python' -e PYTHON_VERSION=3.6 -e MB_PYTHON_VERSION=3.6 -e UNICODE_WIDTH=32 -e BUILD_COMMIT=e5f37e5bddefe04829f4c80edb4d19cc2690647a -e CONFIG_PATH= -e ENV_VARS_PATH= -e WHEEL_SDIR=wheelhouse -e MANYLINUX_URL= -e BUILD_DEPENDS= -e USE_CCACHE= -e REPO_DIR=protobuf/python -e PLAT=x86_64 -e MB_ML_VER=1 -v /tmpfs/src/github/protobuf:/io -v /home/kbuilder:/parent-home quay.io/pypa/manylinux1_x86_64 /io/multibuild/docker_build_wrap.sh
/usr/bin/python
http://archive.kernel.org/centos-vault/5.11/os/x86_64/repodata/primary.xml.gz: [Errno 12] Timeout: <urlopen error timed out>
Trying other mirror.
http://archive.kernel.org/centos-vault/5.11/os/x86_64/repodata/primary.xml.gz: [Errno 7] HTTP Exception (BadStatusLine):
Trying other mirror.
Error: failure: repodata/primary.xml.gz from base: [Errno 256] No more mirrors to try.


[ID: 9921750] Build finished after 104 secs, exit value: 1```

acozzette:
@zeriyoshi I reran the test and it passed, so I think it must have been a temporary network timeout.

Issue #7316, opened 2020-03-19
protobuf-mode.el should derive from prog-mode


sfllaw:
**What version of protobuf and what language are you using?**
Version: 0.3
Language: Emacs

**What operating system (Linux, Windows, ...) and version?**
N/A

**What runtime / compiler are you using (e.g., python version or gcc version)**
N/A

**What did you do?**
1. In Emacs, turn on protobuf-mode
2. `M-x eval-expression RET (derived-mode-p 'prog-mode)`

**What did you expect to see**
`prog-mode`

**What did you see instead?**
`nil`

**Anything else we should know about your project / environment**
`protobuf-mode` is a major mode for a programming language, albeit a restricted one. In fact, it claims to be [derived from CC Mode](https://github.com/protocolbuffers/protobuf/blob/4567a21004db641e974200f20715a2c9ab1f2378/editors/protobuf-mode.el#L41), which it actually isn’t.

That’s tangential though. In order to derive from `prog-mode`, the `defun prog-mode` at: https://github.com/protocolbuffers/protobuf/blob/4567a21004db641e974200f20715a2c9ab1f2378/editors/protobuf-mode.el#L195-L219 should instead read:
``` elisp
;;;###autoload
(define-derived-mode protobuf-mode prog-mode "Protocol-Buffers"  ;; Replace defun with define-derived-mode
  "Major mode for editing Protocol Buffers description language.

The hook `c-mode-common-hook' is run with no argument at mode
initialization, then `protobuf-mode-hook'.

Key bindings:
\\{protobuf-mode-map}"
  (kill-all-local-variables) ;; Remove spurious interactive call
  (set-syntax-table protobuf-mode-syntax-table)
  (setq major-mode 'protobuf-mode
        mode-name "Protocol-Buffers"
        local-abbrev-table protobuf-mode-abbrev-table
        abbrev-mode t)
  (use-local-map protobuf-mode-map)
  (c-initialize-cc-mode t)
  (if (fboundp 'c-make-emacs-variables-local)
      (c-make-emacs-variables-local))
  (c-init-language-vars protobuf-mode)
  (c-common-init 'protobuf-mode)
  (easy-menu-add protobuf-menu)
  (c-run-mode-hooks 'prog-mode-hook 'c-mode-common-hook 'protobuf-mode-hook) ;; Add prog-mode-hook
  (c-update-modeline))
```

Alex-David-Hu:
you must use windows system




[...]

PR #9572, opened 2022-03-03
[editors/emacs] Fix protobuf-mode definition

kunhtkun:
Fix a few issues with the Emacs mode definition.

First, in 1ab7789f3 (2021-10-13, Emacs: Protobuf mode should be derived from
prog-mode) we made `protobuf-mode` a derived mode of `prog-mode` using the
`define-derived-mode` macro ([doc](https://www.gnu.org/software/emacs/manual/html_node/elisp/Derived-Modes.html)). However, the definition body was not updated
accordingly. So in this commit, we:

- Remove the superfluous `(interactive)` form;
- Remove the unnecessary call of `kill-all-local-variables`, which is already
  handled by `define-derived-mode` and could cause a few issues, for example,
  it prevents `prog-mode-hook` from being run;
- Remove forms that set `major-mode` and `mode-name`, which are automatically
  set to the first and third arguments respectively;
- Remove forms that set key map, syntax table, and abbrev table, which are
  already handled automatically;
- Do not run `protobuf-mode-hook` explicitly in the body. It is already arranged
  to be run after the body.

Second, the call to `c-make-emacs-variables-local` is removed. It is called
inside `c-init-language-vars` already. Calling it again should do no harm now,
but to be future-proof it might be better to just remove it.

Finally, we move the `c-update-modeline` form to the :after-hook argument to
ensure it is run at the very end, so that the mode line will reflect all user
customizations done in various mode hooks. Similarly, we run
`c-mode-common-hook` also at the very end to leave a place for user
customizations (for example, set `imenu-generic-expression` in
`c-mode-common-hook`).

Re: https://github.com/protocolbuffers/protobuf/issues/7316
Re: https://github.com/protocolbuffers/protobuf/pull/9076

Cc @xuhdev.

Thanks!

### DIFF
--- a/editors/protobuf-mode.el
+++ b/editors/protobuf-mode.el
@@ -193,7 +193,7 @@
 ;;;###autoload (add-to-list 'auto-mode-alist '("\\.proto\\'" . protobuf-mode))
 
 ;;;###autoload
-(define-derived-mode protobuf-mode prog-mode "Protobuf"
+(define-derived-mode protobuf-mode prog-mode "Protocol-Buffers"
   "Major mode for editing Protocol Buffers description language.
 
 The hook `c-mode-common-hook' is run with no argument at mode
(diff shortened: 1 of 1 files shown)


acozzette:
@kunhtkun Thanks!

Issue #6977, opened 2019-12-03
Java Performance regression since 3.8.0


thesamet:
**What version of protobuf and what language are you using?**
Version: 3.8.0-3.11.0
Language: Java

**What operating system (Linux, Windows, ...) and version?**
Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**
protoc and jvm 1.8

Between version 3.7.1 and version 3.8.0, the implementation of `CodedOutputStream.writeInt32NoTag` (and possibly other functions) has changed. The new implementation is 6x slower according to two microbenchmarks:

| Function     | 3.5.1    | 3.7.1    | 3.8.0     | 3.9.2     | 3.10.0      | 3.11.0      |
| :----------- | -------: |-------:  | --------: | --------: | ----------: | ----------: |
| message3     | 15.48 ns | 15.55 ns | 30.91 ns  | 30.81 ns  | 30.95 ns    | 30.97 ns    |
| int32s       | 46.23 ns | 46.32 ns | 298.87 ns | 299.07 ns | 299.27ns    | 299.05 ns   |

The function `message3` calls `toByteArray` on a message with 3 `int32s`. It becomes 2x slower after 3.7.1

The function `int32s`  serializes 100 integers into a CodedOutputStream. It exercises directly the impacted code path and therefore gets ~6.5 times slower after 3.7.1.

Benchmarks project: https://github.com/thesamet/protobuf-issue-6977
Code: https://github.com/thesamet/protobuf-issue-6977/blob/master/src/main/scala/Test.scala

ennerf:
I only had a few minutes to look at this, but it seems like there is a problem with the unrolled unsafe loop in [CodedOutputStream#L1340-L1364](https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L1340-L1364).

I tried changing `position` to be a local variable, and that made the performance match the in-loop case (`290ns -> 40ns` for the int32s benchmark), i.e.,

```Java
int position = this.position;
try {
  // unrolled-loop
} finally {
  this.position = position;
}
```

There is probably an optimization that it's not picking up on. I haven't had a chance to look at the assembly or confirm it for other cases though.

elharo:
Thanks for the benchmark. We don't use or support Scala. I'm experimenting with reimplementing this in Java so our build system can run it.

[...]

PR #9578, opened 2022-03-04
Java: fix for performance regression introduced in 3.8.0

ennerf:
This PR serves as a sample for the #6977 discussion. I don't have the toolchains setup anymore to run the benchmarks myself, but this minor change fixed it ~2 years ago.

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
+++ b/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
@@ -1348,35 +1348,14 @@ public abstract class CodedOutputStream extends ByteOutput {
       if (HAS_UNSAFE_ARRAY_OPERATIONS
           && !Android.isOnAndroidDevice()
           && spaceLeft() >= MAX_VARINT32_SIZE) {
-        int position = this.position;
-        try {
+        while (true) {
           if ((value & ~0x7F) == 0) {
             UnsafeUtil.putByte(buffer, position++, (byte) value);
             return;
+          } else {
+            UnsafeUtil.putByte(buffer, position++, (byte) ((value & 0x7F) | 0x80));
+            value >>>= 7;
           }
-          UnsafeUtil.putByte(buffer, position++, (byte) (value | 0x80));
-          value >>>= 7;
-          if ((value & ~0x7F) == 0) {
-            UnsafeUtil.putByte(buffer, position++, (byte) value);
-            return;
-          }
-          UnsafeUtil.putByte(buffer, position++, (byte) (value | 0x80));
-          value >>>= 7;
-          if ((value & ~0x7F) == 0) {
-            UnsafeUtil.putByte(buffer, position++, (byte) value);
-            return;
-          }
-          UnsafeUtil.putByte(buffer, position++, (byte) (value | 0x80));
-          value >>>= 7;
-          if ((value & ~0x7F) == 0) {
-            UnsafeUtil.putByte(buffer, position++, (byte) value);
-            return;
-          }
-          UnsafeUtil.putByte(buffer, position++, (byte) (value | 0x80));
-          value >>>= 7;
-          UnsafeUtil.putByte(buffer, position++, (byte) value);
-        } finally {
-          this.position = position;
         }
       } else {
         try {
--- a/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
+++ b/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
@@ -1345,33 +1345,19 @@ public abstract class CodedOutputStream extends ByteOutput {
 
     @Override
     public final void writeUInt32NoTag(int value) throws IOException {
-      if (HAS_UNSAFE_ARRAY_OPERATIONS
-          && !Android.isOnAndroidDevice()
-          && spaceLeft() >= MAX_VARINT32_SIZE) {
+      try {
         while (true) {
           if ((value & ~0x7F) == 0) {
-            UnsafeUtil.putByte(buffer, position++, (byte) value);
+            buffer[position++] = (byte) value;
             return;
           } else {
-            UnsafeUtil.putByte(buffer, position++, (byte) ((value & 0x7F) | 0x80));
+            buffer[position++] = (byte) ((value & 0x7F) | 0x80);
             value >>>= 7;
           }
         }
-      } else {
-        try {
-          while (true) {
-            if ((value & ~0x7F) == 0) {
-              buffer[position++] = (byte) value;
-              return;
-            } else {
-              buffer[position++] = (byte) ((value & 0x7F) | 0x80);
-              value >>>= 7;
-            }
-          }
-        } catch (IndexOutOfBoundsException e) {
-          throw new OutOfSpaceException(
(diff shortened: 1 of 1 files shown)


java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
```diff
@@ -1348,31 +1348,36 @@ public final void writeUInt32NoTag(int value) throws IOException {
...
-        UnsafeUtil.putByte(buffer, position++, (byte) (value | 0x80));
-        value >>>= 7;
-        if ((value & ~0x7F) == 0) {
-          UnsafeUtil.putByte(buffer, position++, (byte) value);
-          return;
-        }
-        UnsafeUtil.putByte(buffer, position++, (byte) (value | 0x80));
-        value >>>= 7;
-        if ((value & ~0x7F) == 0) {
+        int position = this.position;
```
elharo, line 1351: Interesting. Two things of note:

1. I assume CodedOutputStream is not expected to be thread-safe, and it doesn't look like it is now, but I need to verify that and think about whether position might be updated asynchronously.

2. This whole block doesn't run on android so the relevant benchmarks are for server.

'Thanks!

ennerf:
I ended up going through the benchmark setup, but I unfortunately couldn't reproduce the behavior I was seeing. There may have been other changes since then.

Here is a gist for a Java version of the Scala benchmark: [WriteUint32Bench](https://gist.github.com/ennerf/5a86471138ecedd7e04d7e7a87f4575a)

java/core/src/main/java/com/google/protobuf/CodedOutputStream.java
ennerf: The existing behavior is definitely not thread-safe, and it wouldn't make much sense to use simultaneous writes on a single output stream. Any extra information written while serializing a message would break at least the length delimiters and add fields that the parser may not be expecting.

ennerf:
reverting to the 3.7.0 version fixes it

```Java
/**
 * === 3.7.0
 * Benchmark                              Mode  Cnt   Score   Error   Units
 * WriteUint32Bench.testWriteInt32NoTag  thrpt   10  29,559 ± 0,499  ops/us
 * WriteUint32Bench.testWriteMessage     thrpt   10  71,854 ± 0,623  ops/us
 *
 * === 3.8.0
 * Benchmark                              Mode  Cnt   Score   Error   Units
 * WriteUint32Bench.testWriteInt32NoTag  thrpt   10   2,178 ± 0,016  ops/us
 * WriteUint32Bench.testWriteMessage     thrpt   10  33,218 ± 0,285  ops/us
 *
 * === 3.19.4
 * Benchmark                              Mode  Cnt   Score   Error   Units
 * WriteUint32Bench.testWriteInt32NoTag  thrpt   10   2,164 ± 0,027  ops/us
 * WriteUint32Bench.testWriteMessage     thrpt   10  32,970 ± 0,217  ops/us
 *
 * === 3.19.4 (reverted to 3.7.0 code)
 * Benchmark                              Mode  Cnt   Score   Error   Units
 * WriteUint32Bench.testWriteInt32NoTag  thrpt   10  29,605 ± 0,501  ops/us
 * WriteUint32Bench.testWriteMessage     thrpt   10  44,623 ± 1,239  ops/us
 *
 */
```

elharo (commented):
Why did you take out all the Unsafe paths? This does not look like the same code any more.

ennerf:
Because it's faster. The Unsafe code was added as an optimization, but it only slows things down. The path that actually requires Unsafe (`UnsafeDirectNioEncoder`) is still using it.

I created separate commits so that you can cherry pick the version you want.

ennerf:
I added the benchmark results to the previous comment. I only took out the "fast path", so the code is exactly the same as what was in the `else` statement before.

elharo:
FYI, GWP can detect this function and shows it costing about 14,000 GCU so it is a plausible optimization target

elharo:
Important to note: the optimization is on ArrayEncoder.writeUInt32NoTag rather than the method(s) a little higher in the call stack being benchmarked. That's not wrong, but it is important for following this proposal.

elharo:
With change that eliminates the unsafe branch:

```
# Warmup Iteration   1: 7842926.924 ops/s
# Warmup Iteration   2: 12228825.174 ops/s
# Warmup Iteration   3: 12240813.322 ops/s
# Warmup Iteration   4: 12264911.947 ops/s
# Warmup Iteration   5: 12317264.162 ops/s
Iteration   1: 12227318.361 ops/s
Iteration   2: 12228675.497 ops/s
Iteration   3: 12227465.847 ops/s
Iteration   4: 12340207.375 ops/s
Iteration   5: 12325522.243 ops/s
```

Without change:

```
Duplicate Google agent found - skipping google3agentlib.
# Warmup Iteration   1: 2952518.421 ops/s
# Warmup Iteration   2: 2997562.659 ops/s
# Warmup Iteration   3: 3249345.823 ops/s
# Warmup Iteration   4: 3248366.059 ops/s
# Warmup Iteration   5: 3248051.117 ops/s
Iteration   1: 3248173.974 ops/s
Iteration   2: 3216751.094 ops/s
Iteration   3: 3237493.224 ops/s
Iteration   4: 3195381.124 ops/s
Iteration   5: 3225395.411 ops/s
```

As always with this sort of microbenchmark, the results can be highly dependent on the exact OS, hardware, and JDK. They could even reverse. However these results are from the OS  and JDK Google uses in production, so they are the ones Google cares most about. In that environment, this change is almost three times slower.

ennerf:
> In that environment, this change is almost three times slower.

Removing unsafe branch: `12M ops/sec`
Current version: `3M ops/sec`

I'd interpret this as a 4x speedup. Am I misunderstanding something?

elharo:
The only thing you're misunderstanding is that I can't read big numbers without commas. You're right. That is a 4x speedup which is worth pursuing here. Let me pass this around the team and see if anyone can poke holes in the benchmark, but otherwise this does look good. Thanks.

elharo:
I went back about six years in the history and I think I found where the original "optimization" was introduced, cl/119195498 for the record. The benchmark results at the time showed a small improvement, less than 20%. The results were much more noticeable for some of the other encoders, but it's easy to believe that changes in the JVM or the native OS in the last six years could easily have turned the optimization negative along this code path.

What I plan to do is implement this change in google3 where we have much stronger tools for measuring performance and a very large sample of production code in which to test. If it works there it should save on the order of 1K GCUs, which is noticeable. If that works out, it will get synced back into open source.

ennerf:
Sounds good. Feel free to close the PR if you don't need it anymore.

FWIW if you're looking for an easy performance win, you can change the generated merge statement to allow fallthrough. I've seen quite a few cases where this doubled the total decoding throughput, and I've yet to see anything that became slower.

```Java
int tag = input.readTag();
while(true) {
    switch(tag) {
        case FIELD_1_TAG: {
            // <parse field 1>
            if ((tag = input.readTag) != FIELD_2_TAG) {
                break;
            }
        }
        case FIELD_2_TAG: {
            // <parse field 2>
            if ((tag = input.readTag) != FIELD_3_TAG) {
                break;
            }
        }
        case FIELD_3_TAG: {
            // <parse field 3>
            // etc.
        }
    }
}
```

PR #9584, opened 2022-03-04
Catch non-strings in BinaryEncoder.writeString

dweitzman:
The javascript serializeBinary() function silently tosses away any non-string placed into what should be a string field, writing the equivalent of an empty string instead

Not sure if or why anyone would be relying on that existing arbitrary behavior today, but silent data loss seems very undesirable and worth defending against

### DIFF
--- a/js/binary/decoder_test.js
+++ b/js/binary/decoder_test.js
@@ -377,6 +377,16 @@ describe('binaryDecoderTest', function() {
    });
 
   /**
+   * Verifies that passing a non-string to writeString raises an error.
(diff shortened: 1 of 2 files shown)


acozzette:
Sorry but we can't review the pull request until the CLA is signed. Let me know if you need any help troubleshooting the CLA, though. Sometimes the corporate CLA can be tricky.

acozzette:
Oh sorry, I just realized the CLA is good to go already! Never mind what I said before.

PR #9585, opened 2022-03-04
Updated PHP upb, and utf8_range for Ruby/PHP

haberman:

PR #9588, opened 2022-03-05
Merge master into 3.20.x

acozzette:
This way we will make sure 3.20.0 includes the change from #9585.


haberman (approved):

haberman:
FYI one more (much smaller) change is coming to update Ruby to the latest upb.

PR #9591, opened 2022-03-06
[C#] fix parse failure for extensions with large field numbers

homuler:
Currently, it is failing to parse messages with an extension whose field number is large (> 0x10000000), e.g. https://github.com/google/mediapipe/blob/e6c19885c6d3c6f410c730952aeed2852790d306/mediapipe/calculators/tensor/tensors_to_detections_calculator.proto#L25) because the tag value is cast to `int` before shifting right.

### DIFF
--- a/csharp/compatibility_tests/v3.0.0/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/compatibility_tests/v3.0.0/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -416,6 +416,25 @@ namespace Google.Protobuf
         }
 
         [Test]
+        public void MaximumFieldNumber()
+        {
+            MemoryStream ms = new MemoryStream();
+            CodedOutputStream output = new CodedOutputStream(ms);
+
+            int fieldNumber = 0x1FFFFFFF;
+            uint tag = WireFormat.MakeTag(fieldNumber, WireFormat.WireType.LengthDelimited);
+            output.WriteRawVarint32(tag);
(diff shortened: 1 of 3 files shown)


elharo:
Interesting. Per spec the mac extension number is max is 2^29 - 1, or 536,870,911.
Your example is  268,435,456 in decimal so it should be OK and this does look like a bug.

csharp/compatibility_tests/v3.0.0/src/Google.Protobuf.Test/CodedInputStreamTest.cs
```diff
@@ -415,6 +415,25 @@ public void TestSlowPathAvoidance()
             }
         }
 
+        [Test]
```
jskeet, line 418: Is there any reason you've added this test in here as well? We don't normally do that.

csharp/src/Google.Protobuf/WireFormat.cs
```diff
@@ -90,7 +90,7 @@ public static WireType GetTagWireType(uint tag)
         /// </summary>
         public static int GetTagFieldNumber(uint tag)
         {
-            return (int) tag >> TagTypeBits;
+            return (int) (tag >> TagTypeBits);
```
jskeet, line 93: Ouch, that was a subtle bug. Thanks for catching it.

jskeet:
Great, thanks. I'll squash and merge on green.

jskeet:
This shouldn't affect the MacOS Ruby Release, which is the only target still running. Merging.

PR #9594, opened 2022-03-07
Create .tar.gz archives instead of zip for linux and macos

### LINKED ISSUE #9569: Linux arch releases should be tar.gz
Github release process

Linux distro's like Ubuntu don't include `unzip` by default. The releases already include `tar.gz`. Why would the linux binary builds not use `tar.gz`? Might make sense for Windows, but not Linux.

Archive with `tar.gz` instead of `zip`

Installing `unzip` which is an unnecessary step if the archive was not `zip`.

ntkme:
This PR changes github release artifacts on linux and macos to be tar.gz instead of zip files. 

I don't know how the actual upload happens so there might be other things in kokoro need to be updated.

- Fixes #9569.

### DIFF
--- a/protoc-artifacts/build-zip.sh
+++ b/protoc-artifacts/build-zip.sh
@@ -11,18 +11,18 @@ Example:
   $ $0 protoc-gen-javalite 3.0.0
 
 This script will download pre-built protoc or protoc plugin binaries from maven
-repository and create .zip packages suitable to be included in the github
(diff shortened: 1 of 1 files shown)


perezd:
There's more work to make this happen consistently across our release management platform and it's unlikely we'll be accepting this PR at this time, as we have other critical areas we're focused on at the moment. Thanks for understanding!

PR #9600, opened 2022-03-08
[PHP] Remove unnecessary zval initialization

zeriyoshi:
protobuf extension creates a userland variable (zval) upon loading.

This is an unnecessary process and compromises userland compatibility.

```shell
# stdClass is initialized as the first element.
$ php -r 'var_dump(new \stdClass());'
object(stdClass)#1 (0) {
}
# but, stdClass is the second element
$ php -dextension=modules/protobuf.so -r 'var_dump(new \stdClass());'
object(stdClass)#2 (0) {
}
```

Since this behavior does not seem to be effective at this time, i have changed the implementation and added a test.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -840,6 +840,7 @@ php_EXTRA_DIST=                                                       \
   php/ext/google/protobuf/protobuf.c                                  \
   php/ext/google/protobuf/protobuf.h                                  \
   php/ext/google/protobuf/wkt.inc                                     \
+  php/ext/google/protobuf/tests/unnecessary_zval.phpt                 \
   php/generate_descriptor_protos.sh                                   \
   php/generate_test_protos.sh                                         \
   php/release.sh                                                      \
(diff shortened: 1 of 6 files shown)


zeriyoshi:
@acozzette 
many problems fixed, please re-run kokoro

haberman:
> This is an unnecessary process and compromises userland compatibility.

Can you explain a bit more what is wrong with the current behavior?

What is the harm of the `object(stdClass)#2` output?

Your change causes `InternalDescriptorPool::getGeneratedPool()` to return a new/different object every time. I would prefer not to do this.

zeriyoshi:
@haberman 

> Can you explain a bit more what is wrong with the current behavior?
> What is the harm of the object(stdClass)#2 output?

Defining variables by extension is generally not done with PHP extensions, and many tests that rely on output results such as var_dump will fail. For example, the PHPT testing cases bundled with PHP.

> Your change causes InternalDescriptorPool::getGeneratedPool() to return a new/different object every time. I would prefer not to do this.

Is this for performance reasons? I think it should be treated as a singleton on the PHP side implementations.

As far as I can see it appears to have already been done:

https://github.com/protocolbuffers/protobuf/blob/92cdf87f1a276356657ca1b426c115bff0ad2798/php/src/Google/Protobuf/DescriptorPool.php#L47

haberman:
> As far as I can see it appears to have already been done:

Sounds good, thanks for checking.

I don't love that this produces a new testing entry point (`.php` and `make test`), but we can adapt that later if necessary.

PR #9602, opened 2022-03-08
Fix failing FieldMask.Merge for well-known wrapper field types

### LINKED ISSUE #9283: FieldMask.Merge not worked with WellKnownTypes wrappers
Version: 3.19.1
Language: C#

Any OS (Try on Ubuntu 20.04, Windows 7, 10)

.net5.0, .net6.0

**What did you do?**
Example code

```proto
syntax = "proto3";

import "google/protobuf/wrappers.proto";
import "google/protobuf/field_mask.proto";

option csharp_namespace = "Test.Proto";

message MergeDataRequest {
    google.protobuf.StringValue test = 1;
}
```

```csharp
var mask = new Google.Protobuf.WellKnownTypes.FieldMask();
mask.Paths.Add("test");

var m1 = new Test.Proto.MergeDataRequest {
    Test = "testString",
}; 

var m2 = new Test.Proto.MergeDataRequest();
mask.Merge(m1, m2);
```

Successful merge from source to destination.

**What did you see instead?**

Last line got

```
Unhandled exception. System.InvalidCastException: Unable to cast object of type 'System.String' to type 'Google.Protobuf.IMessage'.
   at Google.Protobuf.FieldMaskTree.Merge(Node node, String path, IMessage source, IMessage destination, MergeOptions options) in /home/heggi/Projects/Google.Protobuf/FieldMaskTree.cs:line 336
   at Google.Protobuf.FieldMaskTree.Merge(IMessage source, IMessage destination, MergeOptions options) in /home/heggi/Projects/Google.Protobuf/FieldMaskTree.cs:line 243
   at Google.Protobuf.WellKnownTypes.FieldMask.Merge(IMessage source, IMessage destination, MergeOptions options) in /home/heggi/Projects/Google.Protobuf/WellKnownTypes/FieldMaskPartial.cs:line 324
   at Google.Protobuf.WellKnownTypes.FieldMask.Merge(IMessage source, IMessage destination) in /home/heggi/Projects/Google.Protobuf/WellKnownTypes/FieldMaskPartial.cs:line 332
   at Program.<Main>$(String[] args) in /home/heggi/Projects/test/Program.cs:line 9
```

Same trouble with any wrappers types

vassilvk:
Closes #9283.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
+++ b/csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
@@ -432,5 +432,89 @@ namespace Google.Protobuf
             Assert.IsNotNull(destination.Payload);
         }
 
+        [Test]
+        public void MergeWrapperFieldsWithNonNullFieldsInSource()
+        {
+            // Instantiate a destination with wrapper-based field types.
+            var destination = new TestWellKnownTypes()
+            {
+                StringField = "Hello",
+                Int32Field = 12,
+                Int64Field = 24,
+                BoolField = true,
+            };
+
+            // Set up a targeted update.
+            var source = new TestWellKnownTypes()
+            {
+                StringField = "Hi",
+                Int64Field = 240
+            };
+
+            Merge(new FieldMaskTree().AddFieldPath("string_field").AddFieldPath("int64_field"),
+                source,
+                destination,
+                new FieldMask.MergeOptions(),
+                false);
+
+            // Make sure the targeted fields changed.
+            Assert.AreEqual("Hi", destination.StringField);
+            Assert.AreEqual(240, destination.Int64Field);
+
+            // Prove that non-targeted fields stay intact...
+            Assert.AreEqual(12, destination.Int32Field);
+            Assert.IsTrue(destination.BoolField);
+
+            // ...including default values which were not explicitly set in the destination object.
+            Assert.IsNull(destination.FloatField);
+        }
+
+        [Test]
+        [TestCase(false, "Hello", 24)]
+        [TestCase(true, null, null)]
+        public void MergeWrapperFieldsWithNullFieldsInSource(
+            bool replaceMessageFields,
+            string expectedStringValue,
+            long? expectedInt64Value)
+        {
+            // Instantiate a destination with wrapper-based field types.
+            var destination = new TestWellKnownTypes()
+            {
+                StringField = "Hello",
+                Int32Field = 12,
+                Int64Field = 24,
+                BoolField = true,
+            };
+
+            // Set up a targeted update with null valued fields.
+            var source = new TestWellKnownTypes()
+            {
+                StringField = null,
+                Int64Field = null
+            };
+
+            Merge(new FieldMaskTree().AddFieldPath("string_field").AddFieldPath("int64_field"),
+                source,
+                destination,
+                new FieldMask.MergeOptions()
+                {
+                    ReplaceMessageFields = replaceMessageFields
+                },
+                false);
+
+            // Make sure the targeted fields changed according to our expectations, depending on the value of ReplaceMessageFields.
+            // When ReplaceMessageFields is false, the null values are not applied to the destination, because, although wrapped types
+            // are semantically primitives, FieldMaskTree.Merge still treats them as message types in order to maintain consistency with other Protobuf
+            // libraries such as Java and C++.
+            Assert.AreEqual(expectedStringValue, destination.StringField);
+            Assert.AreEqual(expectedInt64Value, destination.Int64Field);
+
+            // Prove that non-targeted fields stay intact...
+            Assert.AreEqual(12, destination.Int32Field);
+            Assert.IsTrue(destination.BoolField);
+
+            // ...including default values which were not explicitly set in the destination object.
+            Assert.IsNull(destination.FloatField);
+        }
     }
 }
--- a/csharp/src/Google.Protobuf/FieldMaskTree.cs
+++ b/csharp/src/Google.Protobuf/FieldMaskTree.cs
@@ -333,15 +333,24 @@ namespace Google.Protobuf
                         {
                             if (sourceField != null)
                             {
-                                var sourceByteString = ((IMessage)sourceField).ToByteString();
-                                var destinationValue = (IMessage)field.Accessor.GetValue(destination);
-                                if (destinationValue != null)
+                                // Well-known wrapper types are represented as nullable primitive types, so we do not "merge" them.
+                                // Instead, any non-null value just overwrites the previous value directly.
+                                if (field.MessageType.IsWrapperType)
                                 {
-                                    destinationValue.MergeFrom(sourceByteString);
+                                    field.Accessor.SetValue(destination, sourceField);
                                 }
                                 else
                                 {
-                                    field.Accessor.SetValue(destination, field.MessageType.Parser.ParseFrom(sourceByteString));
+                                    var sourceByteString = ((IMessage)sourceField).ToByteString();
+                                    var destinationValue = (IMessage)field.Accessor.GetValue(destination);
+                                    if (destinationValue != null)
+                                    {
+                                        destinationValue.MergeFrom(sourceByteString);
+                                    }
+                                    else
+                                    {
+                                        field.Accessor.SetValue(destination, field.MessageType.Parser.ParseFrom(sourceByteString));
+                                    }
                                 }
                             }
                         }


vassilvk:
Thank you @acozzette.

We are trying to get through the process of establishing a corporate contributor CLA for my employer (and it is taking some time).
What is the best way to reach you in case we have questions?

acozzette:
You can either ask here or feel free to email me at acozzette@google.com.

vassilvk:
@acozzette, FYI, the CLA is signed.

jskeet:
Ah, I suspect that in Java it's okay because the reflection value for a wrapper field is still the wrapper, whereas in .NET it's the boxed primitive.

However, this leads to an issue in this fix - because it means that the for wrapper fields, `options.ReplacePrimitiveFields` is used instead of `options.ReplaceMessageFields` - when fundamentally it's still a message field. I'd like to see the options behave consistently.

vassilvk:
Hi @jskeet,

Please see cef9f677922068663e29465cd339c5032c60ded7.

Please note that the well known wrappers are semantically primitive values in a sense that the Replace* options would have no logical bearing on the outcome of them being merged. In other words, for those fields "merge" is equivalent to "replace".

That said, the above commit moves the check into the message branch of the merge operation as per your suggestion.

jskeet:
I'll have a look at this tomorrow - I'm afraid I won't be able to get to it today.

jskeet:
Okay, I think I understand now (this whole piece of code is new to me). I think the new behavior is reasonable, but I can imagine it *possibly* behaving differently to the equivalent [Java code](https://github.com/protocolbuffers/protobuf/blob/main/java/util/src/main/java/com/google/protobuf/util/FieldMaskTree.java#L309). I suspect that in the Java code, the wrapper types are treated like any other message types - so a wrapper type with a value of 0 won't actually end up replacing the original value, because it will be merged field-wise, with the absent value field being ignored.

Given the way reflection for wrapper types works in C#, I *suspect* we don't want to replicate that behavior, but I'd like @acozzette's opinion before merging.

vassilvk:
Hello @acozzette. Any update?
Thanks.

acozzette:
Sorry I missed this! I think Jon brings up an interesting point about how in Java, a zero-valued wrapper type can't override the original value. I glanced at the C++ implementation and I suspect this is true in C++ as well, and probably in just about every language, since as far as I know we didn't take any special care to handle this situation.

So unfortunately I think we probably want to imitate the Java behavior--even though it's less intuitive, I think it's best to aim for consistency across languages. Hopefully this is an easy tweak to make: I think we just want to ignore the source field in the special case where a wrapper field is present in both the source and destination but the contents of the source field are empty.

vassilvk:
Thanks @acozzette.
Suggested changes are complete.

jskeet:
I'll have a look later in the week, thanks.

csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
```diff
@@ -432,5 +432,44 @@ public void Merge(bool useDynamicMessage)
...
+                Int32Field = 12,
+                Int64Field = 24,
+                BoolField = null,
+            };
+
+            // Set up a targeted update.
+            var source = new TestWellKnownTypes()
+            {
+                StringField = "Hi",
+                Int64Field = null
```
jskeet, line 453: I think we should possibly have two tests - one for "setting null values in wrapper fields" (for both StringField and Int64Field) and one for "setting non-null values in wrapped fields) (again, for both). We could potentially parameterize it all to have one test, but I suspect it's cleaner not to.

csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
```diff
@@ -432,5 +432,44 @@ public void Merge(bool useDynamicMessage)
...
+            Merge(new FieldMaskTree().AddFieldPath("string_field").AddFieldPath("int64_field"),
+                source,
+                destination,
+                new FieldMask.MergeOptions()
+                {
+                    ReplaceMessageFields = replaceMessageFields
+                },
+                false);
+
+            // Make sure only the targeted fields were updated.
```
jskeet, line 465: I would expect this to look at Int32Field and BoolField... given that StringField *is* targeted, I wouldn't expect it to be asserted in this section. I suspect we effectively want two commented sections:

- Make sure only the targeted fields were updated (Int32Field and BoolField haven't changed)
- "Prove that ReplaceMessageFields works for wrapper types" - I'd expect the same test for both StringField and Int64Field here (see earlier comment about mixing null and non-null behavior in one test).

csharp/src/Google.Protobuf/FieldMaskTree.cs
```diff
@@ -333,15 +333,23 @@ public void Merge(IMessage source, IMessage destination, FieldMask.MergeOptions
                         {
                             if (sourceField != null)
                             {
-                                var sourceByteString = ((IMessage)sourceField).ToByteString();
-                                var destinationValue = (IMessage)field.Accessor.GetValue(destination);
-                                if (destinationValue != null)
+                                // Well-known wrapper types are represented as nullable primitive types, so we do not "merge" them.
```
jskeet, line 336:
```suggestion
                                // Well-known wrapper types are represented as nullable primitive types, so we do not "merge" them.
                                // Instead, any non-null value just overwrites the previous value directly.
```

(Just trying to avoid others having to go through the same mental process I did!)

csharp/src/Google.Protobuf/FieldMaskTree.cs
```diff
@@ -333,15 +333,23 @@ public void Merge(IMessage source, IMessage destination, FieldMask.MergeOptions
...
+                                if (field.MessageType.IsWrapperType)
                                 {
-                                    destinationValue.MergeFrom(sourceByteString);
+                                    field.Accessor.SetValue(destination, sourceField);
                                 }
                                 else
                                 {
-                                    field.Accessor.SetValue(destination, field.MessageType.Parser.ParseFrom(sourceByteString));
+                                    var sourceByteString = ((IMessage)sourceField).ToByteString();
```
jskeet, line 343: Note for anyone else reviewing this: the code here really hasn't changed at all other than indentation - the diff is just confusing. (It's slightly better in split mode than in unified mode, but either way it's not clear that really we're just conditionalizing the existing behavior.)

jskeet (commented):
Production code looks good to me, although it's possible that I've missed some subtlety. I think it would be good to separate out a "wrapper field with null value" test (with both ReplaceMessageFields true and false) and "wrapper field with non-null value" (where we probably don't need parameterization).

csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
vassilvk: Please see https://github.com/protocolbuffers/protobuf/pull/9602/commits/c4e7df18593b613ca08884fc5b955adae438ca3a

Splitting into different tests ended up producing too much repetition, specifically around the "make-sure-non-targeted-fields-remain-intact" part -- a part which I very much want to cover in all cases.

Went for a single test with a little more extensive test matrix instead.
In my opinion it seems clean enough without much unnecessary crud.

csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
vassilvk: Please see comment above and https://github.com/protocolbuffers/protobuf/pull/9602/commits/c4e7df18593b613ca08884fc5b955adae438ca3a

csharp/src/Google.Protobuf.Test/FieldMaskTreeTest.cs
jskeet: Well, it certainly took me longer to review than a larger set of more focused tests would have :(

jskeet:
Ah, the CLA check is still failing - I'll need to check that manually.

jskeet:
Hmm - if you could email me (jonskeet@google.com) with details about the CLA signing (e.g. whether it's been done in a corporate way rather than for your GitHub account) that would be appreciated.

vassilvk:
> I believe the production code is correct - but the test really is quite hard to understand at the moment, from the perspective of someone who has come into this "cold" as it were.

Hi @jskeet - I understand.
Please see commit 9baab1f33cbaa11d7c3b438ae35f7ddfb78895a7 with split tests.

I'll take a look at the SLA.

vassilvk:
I see what the problem is with the CLA.
I believe it trips on your `@pobox.com` email identity:

image

jskeet:
Ah, right - thanks. I'll manually override that when I've reviewed the new tests.

jskeet (approved):

jskeet:
Okay, the CLA manual override process has changed over time.

I suspect the simplest approach now is for you to:

- Soft-reset to 92cdf87f1a276356657ca1b426c115bff0ad2798 (the parent of the first commit)
- Commit all the changes in a single commit (we don't need all the history)
- Force push

That should end up with a single commit just by you, which should avoid all the CLA problems. Sorry for the bureaucracy :(

vassilvk:
> Thanks for your patience :)
Thank you for yours too :)

olivierl:
Hi, do you know when this fix will be released? Thx

RogerioWagner:
Hi, the issue of FieldMask.Merge with Wrappers has been fixed, or is there any workaround to create path's for wrappers? I just experienced same behavior reported here: https://github.com/protocolbuffers/protobuf/issues/9283#issue-1072438360

jskeet:
@RogerioWagner: I've just tried the *exact* repro in #9283, and it's fine using the latest NuGet package. So either you're not actually experiencing the same behavior (i.e. your situation is different in a way we can't know) or you're using an old version where this was still broken. I suggest you file a new issue with a minimal repro.

RogerioWagner:
@jskeet, I were right, it's a different behavior after I bumped the version to 3.26.0, now instead of an exception, it failed silently with a warning at MaskFilter ({ "@warning": "Invalid FieldMask", "paths": [ "description", "usage", "templateId" ] }), where description is an optional string, usage is an enum and templateId is an optional google.protobuf.Int32Value.

jskeet:
@RogerioWagner: Rather than half-describing the problem in a merged PR, please file a new issue with a minimal but complete repro.

Issue #9596, opened 2022-03-08
CMakeLists.txt should be in top-level directory


Arfrever:
The standard practice is to put main `CMakeLists.txt` in top-level directory, not in subdirectory called `cmake` or similar.
Some examples of projects:
[abseil-cpp](https://github.com/abseil/abseil-cpp)
[gflags](https://github.com/gflags/gflags)
[glog](https://github.com/google/glog)

I would suggest to move Protobuf's `cmake/CMakeLists.txt` to top-level directory and adjust appropriate paths.
If there is agreement with this suggestion, then I can try to work on it.

acozzette:
Is there a way to do this without breaking scripts that rely on the existing location? Maybe we could leave a symlink at the old location if that works.

bsergean:
Old versions of Windows don't support symlinks, but newer might. It would be great if the Cmakefile was usable on linux/mac as well.

Arfrever:
`cmake/CMakeLists.txt` can become a small regular file with few lines:
```
cmake_minimum_required(VERSION 3.5)

message(WARNING "Calling of cmake with Protocol Buffers' \"cmake\" subdirectory as source directory is deprecated. Top-level directory should be used instead.")

project(protobuf C CXX)

set(protobuf_DEPRECATED_CMAKE_SUBDIRECTORY_USAGE TRUE)

include(../CMakeLists.txt)
```

bsergean:
Nice, that’s much in my opinion to the symlink approach.

acozzette:
Yeah, that looks like a good way to do it!

PR #9603, opened 2022-03-08
Rename main cmake/CMakeLists.txt to CMakeLists.txt

Arfrever:
cmake/CMakeLists.txt is kept for compatibility and it includes CMakeLists.txt.

Fixes: #9596


CMakeLists.txt
```diff
@@ -0,0 +1,351 @@
+# Minimum CMake required
```
acozzette, line 1: Could you do a `git mv` from the old location so that we can preserve the history of this file?

Arfrever: I actually did `git mv cmake/CMakeLists.txt CMakeLists.txt` and later separate `git add cmake/CMakeLists.txt` for new file.
The only way to do what you want seems to be to have 2 separate commits.
Should I do it?

acozzette: Actually never mind what I said. From reading some more I think I misunderstood how Git handles file moves and copies. If I understand correctly now, Git doesn't try to express file moves through metadata but instead just tries to detect them after the fact. It's too bad the GitHub UI doesn't show the diff better, though.

Arfrever: I have split those changes into 2 commits, and added a clean-up commit (`protobuf_source_dir` -> `protobuf_SOURCE_DIR`), so if you merge these changes as 3 separate commits (not squashed), then history should be better shown.

acozzette: Awesome! I will take a closer look soon, but my only major concern so far is I think this change is a bit too big to try to get into the 3.20.0 release given that the release is already in progress. Would you mind rebasing this pull request to be against the master branch instead? I can first merge 3.20.x into master if that will help.

Arfrever: OK. Please merge `3.20.x` into `master`, and I will re-base these changes against `master`...

acozzette: Ok, 3.20.x has now been merged into master.

acozzette:
Thanks, @Arfrever!

PR #9610, opened 2022-03-09
Changing toolchain build file

deannagarcia:
We are moving to use manylinux toolchains in order to guarantee backwards compatibility to old linux distributions. This toolchain file moves us to use manylinux as the sysroot to find c++ header files. There were also modifications to the Docker image that allows us to remove the extra_compiler_flags for all linux architectures.

### DIFF
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -43,32 +43,24 @@ cc_toolchain_config(
     name = "linux-aarch_64-config",
     bit_flag = "-m64",
     cpp_flag = "-lstdc++",
-    extra_compiler_flags = [
-        "-I/opt/manylinux/2014/aarch64/usr/include/c++/10/aarch64-redhat-linux",
-        "-I/opt/manylinux/2014/aarch64/usr/include/c++/10"
-    ],
     sysroot = "/opt/manylinux/2014/aarch64",
     linker_path = "/usr/bin/ld",
     target_cpu = "aarch64",
     target_full_name = "aarch64-linux-gnu",
     toolchain_name = "linux_aarch_64",
     # Don't really need this, setting it because it's required.
-    toolchain_dir = "/opt/manylinux/2014/aarch64/usr/include",
+    toolchain_dir = "/opt/manylinux/2014/aarch64",
 )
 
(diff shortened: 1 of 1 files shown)


toolchain/BUILD
```diff
@@ -43,72 +43,60 @@ cc_toolchain_config(
...
-        "-I/usr/powerpc64le-linux-gnu/include/c++/8/powerpc64le-linux-gnu/",
-        "-I/usr/powerpc64le-linux-gnu/include/c++/8/"
-    ],
-    extra_include = "/usr/include",
     linker_path = "/usr/bin/ld",
+    sysroot = "/opt/manylinux/2014/ppc64le",
     target_cpu = "ppc64",
     target_full_name = "powerpc64le-linux-gnu",
-    toolchain_dir = "/usr/powerpc64le-linux-gnu/include",
+    toolchain_dir = "/opt/manylinux/2014/ppc64le",
```
haberman, line 63: Is there any remaining case in which `sysroot` and `toolchain_dir` are different?

If not it would be nice to collapse them into a single flag.

deannagarcia: Windows doesn't have a `sysroot` but does have a `toolchain_dir` so we need to have separate flags for now. I think it might be possible to do some work so that windows won't be a special case (and would hopefully only need the `--sysroot` flag) but that change won't be trivial.

haberman: Could we at least make `toolchain_dir` default to `sysroot`, so we only need to pass both when they are different?

perezd: +1 that's a nice enhancement.

deannagarcia: I talked with Josh and we are going to defer this change for another PR when I have more time to investigate how to change windows implementation to be able to delete this variable altogether. 

Any way to have `toolchain_dir` default to `sysroot` removes the mandatory flag from either of them, but if both aren't populated the toolchain would fail.

haberman (approved):

haberman:
This is great work! Reducing each sysroot to be totally self contained, without needing extra include directories, is a really nice simplification.

PR #9611, opened 2022-03-10
Replace gradle plugin

oneiros-de:
The currently linked gradle plugin project is dead; use one which seems to be alive.


acozzette:
Thanks, @oneiros-de.

PR #9614, opened 2022-03-10
cmake: Call get_filename_component() with DIRECTORY mode instead of PATH mode

Arfrever:
`PATH` mode is legacy alias for `DIRECTORY` mode (needed only for CMake <= 2.8.11):
https://cmake.org/cmake/help/latest/command/get_filename_component.html

By the way, `get_filename_component()` command itself is deprecated in favor of [`cmake_path()` command](https://cmake.org/cmake/help/latest/command/cmake_path.html), but `cmake_path()` command was introduced in CMake 3.20, which is too new for Protobuf, which currently requires CMake >= 3.5.


acozzette:
@Arfrever Thanks!

PR #9617, opened 2022-03-10
Enable GeneratedDescriptorTest.EnormousDescriptor test for GCC/Clang CMake builds

Arfrever:


Arfrever:
So apparently commit ff7bdad231d037802fd457ee4fd65e0291d366d7 added exclusion of `google/protobuf/unittest_enormous_descriptor.proto` to `update_file_lists.sh`, but `#ifndef _MSC_VER` in C++ code should suffice.

acozzette:
We should probably actually delete `update_file_lists.sh`, since we're not even relying on that file anymore if I'm not mistaken.

Arfrever:
After running `./update_file_lists.sh`, I see several other unrelated changes in `git diff`...

Arfrever:
In this pull-request, I will include only changes related to `unittest_enormous_descriptor`, and I will leave potential synchronization of those lists of files to another pull-request.

acozzette:
@Arfrever Thanks for the cleanup!

Issue #9593, opened 2022-03-07
[PHP] option exclude backwards compatible proto files


bshaffer:
For PHP all nested messages generate two classes, the current namespaced class and a file to support backwards compatibility with the previous non-namespaced files. There's also a `class_alias` call at the bottom of every nested message. See [this PR from almost 4 years ago](https://github.com/protocolbuffers/protobuf/pull/4536). 

As been almost 4 years, these files are now more of a nuisance. I see three potential options here, ranked from least to most risky:

 - Add a flag to tell `protoc` to not generate these files for PHP (opt-out, most conservative)
 - Add a flag to tell `protoc` to generate these files for PHP (opt-in, could potentially break some very slow to upgrade users)
 - Remove the backwards compatibility logic all together (as it's been throwing deprecation errors for almost four years, I feel comfortable with this solution)

I'm happy to give a PR a try any of these options are possible!

haberman:
I think removing the backward compat logic is the way to go at this point.

Four years with deprecation warnings seems sufficient.

bshaffer:
The only issue I can think of is for libraries like `google-cloud-php`, we wouldn't want to _remove_ the `class_alias` code from proto files which currently have it. Not because we think it's needed, but because it would constitute a backwards-compatibility change. So what we _could_ do is:

1. Only remove the old alias file (e.g. `Nested_Message.php`) but keep the class alias (e.g. the call to `class_alias` at the bottom of `Nested\Message.php`) for now
2. Add a flag (once again) to remove the call to `class_alias`
3. Not care at all and we can handle it on our end :)

terax6669:
since the latest release still outputs deprecated files...

```bash
grep 'is deprecated and will be removed in the next major release' -R path/to/proto -l | xargs rm'
```

haberman:
We're making a breaking change release now. This is the perfect time to remove these deprecated files.

PR #9621, opened 2022-03-11
feat: [PHP] remove legacy generate class file

bshaffer:
See #9593

Remove the generation of the legacy class files for nested messages. Leave the call to `class_alias` at the bottom of namespaced nested message classes alone for now.


haberman:
What is the point of leaving the alias if the old classes have been removed?

bshaffer:
@haberman my thought was that, since _not_ generating a file doesn't _remove_ the files someone would already have, this wouldn't be destructive. However, removing the `class_alias` call would be destructive.

Therefore, it seemed like a better idea to do this in phases. Remove the legacy file generation, _then_ remove the `class_alias` call. I'm happy to remove both if you prefer. It just technically would break Backwards Compatibility with anyone generating PHP classes.

haberman (approved):

PR #9628, opened 2022-03-14
Removing toolchain_dir variable

deannagarcia:
Removing the toolchain_dir variable and replacing usages of it to directly using the sysroot. This change required setting a sysroot for windows. 

Unfortunately, I wasn't able to remove the windows additional compiler flags because gcc defaults to using the non-posix libraries. In the future, we could look into how to change the gcc settings to automatically use the posix directory.


haberman (approved):

perezd (approved):
LGTM!

Issue #9629, opened 2022-03-14
[PHP] ReadOnly is now a keyword for PHP 8.1


bshaffer:
`ReadOnly` is [now a keyword in PHP 8.1](https://php.watch/versions/8.1/readonly), meaning classes previously generated as `ReadOnly` now throw a PHP fatal error in PHP 8.1:

```php
class ReadOnly {}
// PHP Parse error:  syntax error, unexpected token "readonly", expecting identifier in php shell code on line 1
```

To fix this, we can add `readonly` to the list of PHP Reserved Keywords. However, we will need to take into account the previously generated classes which had this name. To do that, we can add a class alias at the bottom:

```php
class_alias(PBReadOnly::class, __NAMESPACE__ . '\ReadOnly');
```

And generate the old `ReadOnly.php` file as an alias, similar to what we have now for nested messages:

```php
<?php
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: example/v1/readonly.proto

namespace Example\V1;

class_exists(PBReadOnly::class);
@trigger_error('Example\V1\ReadOnly is deprecated and will be removed in the next major release. Use Example\V1\\PBReadOnly instead', E_USER_DEPRECATED);
```

This ensures that previous implementations won't break, but fixes the issue in implementations of PHP 8.1. 

Thoughts on this approach? Do you think it'll work?

PR #9633, opened 2022-03-15
fix: reserve "ReadOnly" keyword for PHP 8.1 and add compatibility

bshaffer:
see https://github.com/protocolbuffers/protobuf/issues/9629

### DIFF
--- a/php/ext/google/protobuf/names.c
+++ b/php/ext/google/protobuf/names.c
@@ -82,7 +82,7 @@ const char *const kReservedNames[] = {
     "global",     "goto",        "insteadof",    "interface",    "isset",
     "list",       "match",       "namespace",    "new",          "object",
     "or",         "parent",      "print",        "private",      "protected",
-    "public",     "readme",      "require",      "require_once", "return",
+    "public",     "readonly",    "require",      "require_once", "return",
     "self",       "static",      "switch",       "throw",        "trait",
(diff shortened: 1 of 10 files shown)


bshaffer:
This is bizarre because I get the following error (only when `protobuf.so` is enabled): 

```
1) GeneratedClassTest::testPrefixForReservedWords
Exception: Couldn't find descriptor. Note only generated code may derive from \Google\Protobuf\Internal\Message

/workspace/php/tmp/Lower/PBreadonly.php:27
/workspace/php/tests/GeneratedClassTest.php:954
```

But I am receiving the _same_ error for `PBParent` and `PBSelf` tests. That is strange, as those keywords were added relatively recently. And also strange because the same tests passed when they were merged. So what could be going on here?

bshaffer:
This is bizarre because I get the following error (only when `protobuf.so` is enabled): 

```
1) GeneratedClassTest::testPrefixForReservedWords
Exception: Couldn't find descriptor. Note only generated code may derive from \Google\Protobuf\Internal\Message

/workspace/php/tmp/Lower/PBreadonly.php:27
/workspace/php/tests/GeneratedClassTest.php:954
```

But I am receiving the _same_ error for `PBParent` and `PBSelf` tests. That is strange, as those keywords were added relatively recently. And also strange because the same tests passed when they were merged. So what could be going on here?

haberman (approved):

PR #9634, opened 2022-03-15
Update release version to RC2 (3.20.x 202203152103)

esorot:
Update version to RC2


acozzette (approved):
It would be good to update the pull request description but otherwise LGTM!

deannagarcia:
Is it possible to separate out the ruby changes so the PR isn't so big?

Issue #9626, opened 2022-03-13
[C#] MethodOptions.GetExtension cast exception


JamesNK:
I get a cast exception from `GetExtension` when trying to get `HttpRule`.

Imagine we have two projects:

Project 1 has a gRPC service that uses http extension. It passes http.proto to protoc: 

```xml
<Protobuf Include="..\Proto\greet.proto" GrpcServices="Server"  ProtoRoot="..\Proto\" Link="greet.proto"/>
<Protobuf Include="..\Proto\google\api\annotations.proto" GrpcServices="Server"  ProtoRoot="..\Proto\" Link="google\api\annotations.proto"/>
<Protobuf Include="..\Proto\google\api\http.proto" GrpcServices="Server"  ProtoRoot="..\Proto\" Link="google\api\http.proto"/>
```

Project 2 tries to get `HttpRule` from gRPC service in Project 1. Project 1 references the `Google.Api.CommonProtos` package which contains a different copy of `HttpRule`:

```xml
<PackageReference Include="Google.Api.CommonProtos" Version="2.2.0" />
```

Code:

```cs
var options = methodDescriptor.GetOptions();
var http = options?.GetExtension(AnnotationsExtensions.Http);
```

Error message:

> [A]Google.Protobuf.ExtensionValue1[Google.Api.HttpRule] cannot be cast to [B]Google.Protobuf.ExtensionValue1[Google.Api.HttpRule].
> Type A originates from 'Google.Protobuf, Version=3.18.1.0, Culture=neutral, PublicKeyToken=a7d26565bac4d604' in the context 'Default' at location 'C:\Development\Temp\GrpcHttpApiIssueSample\Service_Api_Reference\bin\Debug\net6.0\Google.Protobuf.dll'.
> Type B originates from 'Google.Protobuf, Version=3.18.1.0, Culture=neutral, PublicKeyToken=a7d26565bac4d604' in the context 'Default' at location 'C:\Development\Temp\GrpcHttpApiIssueSample\Service_Api_Reference\bin\Debug\net6.0\Google.Protobuf.dll'.

I understand why this happens. There are two copies of `AnnotationsExtensions.Http` and the copy being used to get `HttpRule` in Project 2 isn't the one that the gRPC service was compiled with in Project 1.

And a fix is for Project 1 to not compile its own copy of http.proto and instead reference the `Google.Api.CommonProtos` package.

[...]

PR #9635, opened 2022-03-16
Proof of concept for accessing semi-raw extension values

jskeet:
The extension still needs to be registered in the extension
registry, but you don't need to know the exact type of the value in
order to access is.

This is a prototype for #9626.


jskeet:
This is superceded by #9655.

PR #9640, opened 2022-03-16
Toolchain cleanups

deannagarcia:
Make a variety of cleanups to the bazel toolchain including:
- Remove `cpp_flag` variable and instead always use `-lstdc++` 
- Remove `bit_flag` variable and instead set it based on CPU
- Remove `toolchain_name` variable and instead use `target_full_name` as a unique identifier
- Remove osx `extra_includes` since `/usr/include` should be included by default
- Change all single quotes to double quotes to be consistent

### DIFF
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -41,95 +41,70 @@ cc_toolchain_suite(
 
 cc_toolchain_config(
     name = "linux-aarch_64-config",
-    bit_flag = "-m64",
-    cpp_flag = "-lstdc++",
     sysroot = "/opt/manylinux/2014/aarch64",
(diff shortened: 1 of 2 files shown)


haberman (approved):
I love the direction you are going with this!

I'm wondering if `extra_compiler_flags` can go too if the Dockerfile sets up the sysroots right.

deannagarcia:
> I'm wondering if `extra_compiler_flags` can go too if the Dockerfile sets up the sysroots right.

I would love to get that working! I tried a bit on this PR to do it, but Windows is a bit more complicated since there are 2 directories where tools are coming from. I'm planning on playing around with this more in the future.

PR #9645, opened 2022-03-17
Fixed Ruby 3.1 tests by marking intersect? as unimplemented.

haberman:


acozzette:
It looks like the compatibility tests need the same fix. (Or should we delete those? We haven't really been maintaining the compatibility tests.)

PR #9647, opened 2022-03-18
Fix invalid dependency manifest when using `descriptor_set_out`

TrentHouliston:
Currently when using `descriptor_set_out` with `dependency_out` an invalid dependency manifest file is generated.

This is because the code path that generates descriptor sets seems to be entirely separate from the rest of the generators.
This change patches in the `descriptor_set_out` file path into the dependency file when it is used.


acozzette:
@TrentHouliston Thanks!

PR #9648, opened 2022-03-18
Fix uint8array check (Update of #8411)

### LINKED ISSUE #1319: JS: Type not convertible to Uint8Array

graup:
Since #8411 was stale, I incorporated the suggestions by @lukesandberg and @jkrems.

This fixes #1319

### DIFF
--- a/js/binary/utils.js
+++ b/js/binary/utils.js
@@ -1051,11 +1051,6 @@ jspb.utils.byteSourceToUint8Array = function(data) {
     return /** @type {!Uint8Array} */(new Uint8Array(data));
   }
 
-  if (typeof Buffer != 'undefined' && data.constructor === Buffer) {
-    return /** @type {!Uint8Array} */ (
-        new Uint8Array(/** @type {?} */ (data)));
-  }
-
   if (data.constructor === Array) {
     data = /** @type {!Array<number>} */(data);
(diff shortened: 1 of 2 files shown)


acozzette:

PR #9649, opened 2022-03-18
update php release script to not fail when tag exist

esorot:
Current behavior causes the script to fail if there was a previous attempt to create the PHP Release. This change will check for the existence of the tag, and if it doesn't exist will create it making the script slightly more idempotent.


php/release.sh
```diff
@@ -30,7 +30,11 @@ mv ../protobuf/composer.json composer.json
 sed -i 's|php/src|src|g' composer.json
 git add .
 git commit -m "$VERSION"
-git tag "$VERSION"
+if [ $(git tag -l "$VERSION") ]; then
+    echo "tag $VERSION already exists"
```
acozzette, line 34: Nit: looks like this line is indented 2 spaces too much.

acozzette (approved):
LGTM but I think you probably want to base this pull request on 3.20.x so that it will take immediate effect with the release.

PR #9655, opened 2022-03-21
[C#] Update GetExtension to support getting typed value

JamesNK:
Fixes https://github.com/protocolbuffers/protobuf/issues/9626

* Improves GetExtension support for unexpected extension values. The error message for getting an extension value with the wrong type is improved.
* Adds GetExtension support for Extensions that get the value typed as a base type.

@jskeet

### DIFF
--- a/csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
+++ b/csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
@@ -6,7 +6,7 @@
     and without the internal visibility from the test project (all of which have caused issues in the past).
   -->
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard1.1;netstandard2.0</TargetFrameworks>
     <LangVersion>3.0</LangVersion>
     <AssemblyOriginatorKeyFile>../../keys/Google.Protobuf.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
--- a/csharp/src/Google.Protobuf/ExtensionSet.cs
+++ b/csharp/src/Google.Protobuf/ExtensionSet.cs
@@ -112,7 +112,25 @@ namespace Google.Protobuf
             IExtensionValue value;
             if (TryGetValue(ref set, extension, out value))
             {
-                return ((RepeatedExtensionValue<TValue>)value).GetValue();
+                if (value is RepeatedExtensionValue<TValue> extensionValue)
+                {
+                    return extensionValue.GetValue();
+                }
+                else
+                {
+                    var valueType = value.GetType().GetTypeInfo();
+                    if (valueType.IsGenericType && valueType.GetGenericTypeDefinition() == typeof(RepeatedExtensionValue<>))
+                    {
+                        var storedType = valueType.GenericTypeArguments[0];
+                        throw new InvalidOperationException(
+                            "The stored extension value has a type of '" + storedType.AssemblyQualifiedName + "'. " +
+                            "This a different from the requested type of '" + typeof(TValue).AssemblyQualifiedName + "'.");
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("Unexpected extension value type: " + valueType.AssemblyQualifiedName);
+                    }
+                }
             }
             else 
             {
--- a/csharp/src/Google.Protobuf/ExtensionSet.cs
+++ b/csharp/src/Google.Protobuf/ExtensionSet.cs
@@ -84,10 +84,18 @@ namespace Google.Protobuf
                 }
                 else
                 {
-                    var storedType = value.GetType().GetTypeInfo().GenericTypeArguments[0];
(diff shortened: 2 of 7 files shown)


csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
```diff
@@ -6,7 +6,7 @@
     and without the internal visibility from the test project (all of which have caused issues in the past).
   -->
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard1.1;netstandard2.0</TargetFrameworks>
```
JamesNK, line 9: Updated to net462 because VS2022 doesn't support net45.

JamesNK:
Test GeneratedCodeCompilesWithOldCsharpCompiler failed on my machine. Not sure why.

csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
jskeet: Hmm... I haven't had that problem. I've been able to run tests from VS2022 without issues, as far as I'm aware. I'm okay with it though, given that .NET 4.5 (as opposed to 4.5.2) is long out of support, and even 4.5.2 isn't supported for much longer.

csharp/src/Google.Protobuf/ExtensionSet.cs
```diff
@@ -63,7 +64,31 @@ public static class ExtensionSet
...
+                {
+                    return extensionValue.GetValue();
+                }
+                else if (value.GetValue() is TValue underlyingValue)
+                {
+                    return underlyingValue;
+                }
+                else
+                {
+                    var storedType = value.GetType().GetTypeInfo().GenericTypeArguments[0];
```
jskeet, line 87: I think it would be worth leaving a comment here explaining why we believe all `IExtensionValue` values will have generic type arguments - and add a comment to `IExtensionValue` to advise that anyone creating a non-generic `IExtensionValue` implementation should change this code. Alternatively, make this code more defensive.

jskeet (approved):
I'm okay with this barring a small matter of assuming generic-ness... but I'm going to ask for another pair of eyes on it before merging.

jskeet:
> Test GeneratedCodeCompilesWithOldCsharpCompiler failed on my machine. Not sure why.

Hmm... I won't be able to look at that right now, but I'll check it tomorrow.

jskeet:
Please could you add some more details in the PR description that's suitable for release notes?

csharp/src/Google.Protobuf.Test.TestProtos/Google.Protobuf.Test.TestProtos.csproj
JamesNK: It's not a problem if you install VS2022 and VS2019 side-by-side. VS2019 lets you install the bits and VS2022 can then use them.

csharp/src/Google.Protobuf/ExtensionSet.cs
JamesNK: Made it more defensive.

JamesNK:
> Please could you add some more details in the PR description that's suitable for release notes?

Done

> > Test GeneratedCodeCompilesWithOldCsharpCompiler failed on my machine. Not sure why.
> 
> Hmm... I won't be able to look at that right now, but I'll check it tomorrow.

It also fails without any changes (other than updating net451 to net462). Perhaps it is because I only have VS2022 on this machine.

amanda-tarafa (commented):
For completeness maybe improve the method for repeated extensions as well?

1. At least the error message can be improved in a similar fashion.
2. Unless I'm missing something, which very well may be: Maybe even check that the actual type of the elements on the RepeatedField is assignable to TValue and return a new `RepeatedField<TValue>`. This will allow users to get the `RepeatedField<object>` and do the parsing from bytes here as well.

JamesNK:
I updated the get method for repeated values with the improved error message. Creating a new collection doesn't seem right. For example, if a new collection was returned then items added to it wouldn't be in the real collection.

If someone wants to work around this issue they can use the extension method for non-repeated values with a type of `IList`, then process values as desired.

amanda-tarafa (approved):
LGTM, thanks!

amanda-tarafa:
Kokoro is broken in Windows.
There are some specific tests for old compilers: `Google.Protobuf.RefStructCompatibilityTest.RunOldCsharpCompilerAndCheckSuccess`

and also `error CS0006: Metadata file 'bin\Debug\net45\Google.Protobuf.dll' could not be found`

Guessing, I think the net45 target being on the tests projects was forcing the Google.Protobuf.dll to be build in net45, and when you removed that, it's not beeing build anymore so the test is failing?

@jskeet or @JamesNK let me know if you want me to look at it.

JamesNK:
Test fixed by changing the path.

jskeet:
Aha, well spotted. Right, given that everything's passing, I'll squash and merge this now.

jskeet:
Done - thanks @JamesNK!

PR #9657, opened 2022-03-21
Cherry-pick to get Ruby 3.1 support in protobuf 3.20.0

haberman:
@acozzette do we need to update our release script to invoke the new scripts?


acozzette (approved):

acozzette:
I believe the release script should be OK, but we will need to create new Kokoro configs for the 3.20.x branch (can copy over the ones for master and just edit the branch name).

PR #9661, opened 2022-03-22
download_data.sh no longer exists

elharo:
@haberman download_data.sh doesn't seem to exist any more and I can't find it in the git log


Arfrever:
> I can't find it in the git log

https://github.com/protocolbuffers/protobuf/commits/master/benchmarks/download_data.sh

Issue #9620, opened 2022-03-11
Build instructions


elharo:
We are missing a file that explains how to build the whole project. That is,

0. What platforms can you build on?
1. Install bazel.
2. git checkout
3. git clean -xfd .
4. bazel build/test what?

Commit something sooner rather than later, even if it's just scattered notes, then flesh it out. Don't let the perfect be the enemy of the good here.

deannagarcia:
Closing this for now. We will reupdate build instructions as we transition to bazel.

PR #9664, opened 2022-03-22
start work on open source build instructions

elharo:
@deannagarcia @Logofile 

#9620

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -13,10 +13,14 @@ Likely incomplete
 * bazel
 * clang
 
+At the top level, this project is built with bazel. Java parts 
+are built with Maven. All languages assume that protoc is built and
+installed first. 
+
 
 ## Detailed instructions
 
-1. `$ git clone git@github.com:protocolbuffers/protobuf.git`
+1. `$ git clone git@github.com:protocolbuffers/protobuf.git`. You can also clone the repo over https.
 1. `$ cd protobuf`
 1. `$ git clean -xfd .` (Important for cleaning up after previous builds.)
 1. `$ bazel build //...`
@@ -25,3 +29,11 @@ Now run the tests:
 
 
 1. `$ bazel test //...`
+
+
+
+## Building subprojects
+
+After the initial build, it's not always necessary to build everything.
+For instance, you can build and test the Java runtime without
+rebuilding protoc as long as the build artifacts are still available. 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -13,8 +13,7 @@ Likely incomplete
 * bazel
 * clang
 
-At the top level, this project is built with bazel. Java parts 
-are built with Maven. All languages assume that protoc is built and
+This project is built with bazel. Individual language builds assume that protoc is built and
 installed first. 
 
 
@@ -34,6 +33,6 @@ Now run the tests:
 
 ## Building subprojects
 
-After the initial build, it's not always necessary to build everything.
-For instance, you can build and test the Java runtime without
-rebuilding protoc as long as the build artifacts are still available. 
+After the initial build, it's not always necessary to rebuild everything after a change.
+For instance, you can build and test changes to the Java runtime without
+rebuilding protoc as long as you've built it once and the build artifacts are still available. 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,27 @@
+# Building Protocol Buffers
+
+These are instructions for building the open source release of protobuf from source.
+
+Given the right tools, this project can be built on Linux, Mac OS X, or Windows though Linux is by far the most commonly used and best tested platform.
+
+## Prerequisites
+
+Likely incomplete
+
+* git
+* JDK 8 or JDK 11. JDK 17 probably doesn't work.
+* bazel
+* clang
+
+
+## Detailed instructions
+
+1. `$ git clone git@github.com:protocolbuffers/protobuf.git`
+1. `$ cd protobuf`
+1. `$ git clean -xfd .` (Important for cleaning up after previous builds.)
+1. `$ bazel build //...`
+
+Now run the tests:
+
+
+1. `$ bazel test //...`
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -9,7 +9,7 @@ Given the right tools, this project can be built on Linux, Mac OS X, or Windows 
 Likely incomplete
 
 * git
-* JDK 8 or JDK 11. JDK 17 probably doesn't work.
+* JDK 8 or JDK 11.
 * bazel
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+These are instructions for building the open source release of protobuf from source.
+
+Given the right tools, this project can be built on Linux, Mac OS X, or Windows though Linux is by far the most commonly used and best tested platform.
+
+## Prerequisites
+
+Likely incomplete
+
+* git
+* JDK 8 or JDK 11. JDK 17 probably doesn't work.
```
perezd, line 12: rm JDK17 line, Bazel works with JDK11 as the host JDK

BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+* bazel
+* clang
+
+At the top level, this project is built with bazel. Java parts 
+are built with Maven. All languages assume that protoc is built and
```
perezd, line 17: This is about to no longer be true. We build maven artifacts using bazel via rules_jvm_external.

perezd (changes requested):

BUILDING.md
elharo: done

BUILDING.md
elharo: Let's update when that's no longer true. Or is https://github.com/protocolbuffers/protobuf/blob/master/java/BUILD already incorrect?

Logofile (approved):

elharo (commented):
@pzd PTAL

deannagarcia:
There are some open source build instructions at https://github.com/protocolbuffers/protobuf/blob/master/src/README.md. Can we combine these?

BUILDING.md
deannagarcia: That file is the bazel releases. But I think the bigger question here is who is this documentation for? Is this for users building the library from scratch? They can use bazel already. But if you want to mention Maven here I think you should also provide instructions on how to build with maven.

BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+
+These are instructions for building the open source release of protobuf from source.
+
+Given the right tools, this project can be built on Linux, Mac OS X, or Windows though Linux is by far the most commonly used and best tested platform.
+
+## Prerequisites
+
+Likely incomplete
+
+* git
```
deannagarcia, line 11: It's been a while since I completely built from scratch. Maybe you could get a member of the team to do this to make sure the pre-requisites are correct?

BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+are built with Maven. All languages assume that protoc is built and
+installed first. 
+
+
+## Detailed instructions
+
+1. `$ git clone git@github.com:protocolbuffers/protobuf.git`. You can also clone the repo over https.
+1. `$ cd protobuf`
+1. `$ git clean -xfd .` (Important for cleaning up after previous builds.)
+1. `$ bazel build //...`
```
deannagarcia, line 26: I think if we want these explicit instructions we should provide targets for all the languages to build

BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+
+Now run the tests:
+
+
+1. `$ bazel test //...`
```
deannagarcia, line 31: Again, I think we should provide targets here.

BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+
+## Building subprojects
+
+After the initial build, it's not always necessary to build everything.
```
deannagarcia, line 37: Change "necessary to build everything" to "necessary to rebuild everything after a change"

BUILDING.md
perezd: That is the _corrected_ BUILD file that makes use of rules_jvm_external.

BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+For instance, you can build and test the Java runtime without
+rebuilding protoc as long as the build artifacts are still available.
```
deannagarcia, line 39: I think this is confusing because you can't build the Java runtime without building protoc. I think what you're going for is rebuilding? So maybe "you can build and test changes only to to the Java runtime without rebuilding protoc"?

BUILDING.md
```diff
@@ -0,0 +1,39 @@
+# Building Protocol Buffers
```
deannagarcia, line 1: You should be explicit here that these are only instructions for building with bazel, as there are many other ways currently to build protobufs.

BUILDING.md
elharo: This is for developers building this repo from scratch. We need complete instructions for checkout and build. Right now we don't have that. I'll remove the part about Maven unless I discover we're still using it.

BUILDING.md
elharo: Ideally, there's a single command that encompasses all subcommands. If that proves not to be true, then yes, we need to add more commands here or even better improve the top-level build to run all the necessary subcommands.

BUILDING.md
elharo: As above, ideally this isn't needed. I'm trying to find out if it is by writing this and documenting my experience of building.

BUILDING.md
elharo: done

BUILDING.md
perezd: We also have this version:
https://github.com/protocolbuffers/protobuf/blob/main/src/README.md

deannagarcia: There isn't an "official way to build protobufs". People choose how to use protobufs based on their project's needs. Most of the time the best way to use protobufs is to depend on a project that imports protobufs or download an already built executable. If someone directly wants to build protobufs they do it based on the tools they are currently/would like to use, which is why we need to support so many platforms (cmake, autotools, maven, bazel)

BUILDING.md
deannagarcia: Ok, just responding to "Likely incomplete"

BUILDING.md
deannagarcia: It is very rare that a person would want to build the whole protobuf library. More likely, they want to pick a single language and either full or lite and then maybe protoc. So this should have specific build targets for those common workflows, instead of a generic all-encompassing build.

deannagarcia:
I thought we had agreed offline that this documentation should be combined with https://github.com/protocolbuffers/protobuf/blob/main/src/README.md. Can you please combine them?

perezd:
I'd like to suggest we close this for now as we're actively redefining our open source build system to be Bazel source of truth and the interplay between CMake is gonna be different, and also autotools is going away.

PR #9665, opened 2022-03-22

darly:


acozzette (approved):
To fix the test failures, I think all you need to do is add these two new files to `python_EXTRA_DIST` in `Makefile.am`: python/google/protobuf/pyext/unknown_field_set.h python/google/protobuf/pyext/unknown_field_set.cc

Issue #9526, opened 2022-02-18
Normalize line endings for C# sources


jtattermusch:
Followup for incident in https://github.com/protocolbuffers/protobuf/pull/6307/files#r806572011

Run this command to see the current status:
```
git ls-files --eol csharp/src
```

Currently most files use LF, some files use CRLF and some files have mixed line endings.

Also see https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

jtattermusch:
CC @jskeet

jskeet:
We should probably work out any advice we need to give (or repo settings) to avoid users getting into problems in the future.
It would be good to add a test for this too.

jskeet:
I'll try to get to this next week. (This week is pretty busy.)

jskeet:
Okay, (not having read the first comment that shows `git ls-files --eol`, doh) I've written some code to categorize text files (from `src` and `csharp`, recursively) into "empty", "CRLF", "LF" or "mixed". Attaching the results at the bottom of this comment. 

In summary:

- LF: 774
- CRLF: 38
- Mixed: 4

It shouldn't be too hard to normalize on LF - I'm currently not sure of the best place to put a test for that though (to make sure we don't get regressions). I suspect `git ls-files` is the way forward though.

We *might* want to make an exception for .sln files as they're not usually hand-written - it's entirely possible that Visual Studio would clobber them anyway.

[output.txt](https://github.com/protocolbuffers/protobuf/files/8197054/output.txt)

PR #9666, opened 2022-03-23
Add .NET 6 target to Google.Protobuf

Jensaarai:
* .NET 5 reaches end-of-support in May, so add a `net6.0` target to the main project in anticipation of a removal of the `net5.0` target in May.
* Update `PrivateAssets` NuGet packages to latest.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
+++ b/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net60</TargetFrameworks>
+    <TargetFrameworks>net462;netcoreapp3.1;net6.0</TargetFrameworks>
     <AssemblyOriginatorKeyFile>../../keys/Google.Protobuf.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <IsPackable>False</IsPackable>
--- a/csharp/src/Google.Protobuf/Google.Protobuf.csproj
+++ b/csharp/src/Google.Protobuf/Google.Protobuf.csproj
@@ -8,7 +8,7 @@
     <!-- C# 7.2 is required for Span/BufferWriter/ReadOnlySequence -->
     <LangVersion>7.2</LangVersion>
     <Authors>Google Inc.</Authors>
-    <TargetFrameworks>netstandard1.1;netstandard2.0;net45;net50</TargetFrameworks>
+    <TargetFrameworks>netstandard1.1;netstandard2.0;net45;net5.0;net6.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyOriginatorKeyFile>../../keys/Google.Protobuf.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
@@ -29,14 +29,14 @@
     <DefineConstants>$(DefineConstants);GOOGLE_PROTOBUF_SUPPORT_FAST_STRING</DefineConstants>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net50' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net5.0' OR '$(TargetFramework)' == 'net6.0' ">
     <DefineConstants>$(DefineConstants);GOOGLE_PROTOBUF_SUPPORT_FAST_STRING;GOOGLE_PROTOBUF_SIMD</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" PrivateAssets="All" Version="1.0.0"/>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" PrivateAssets="All" Version="1.1.1"/>
     <!-- Needed for the net45 build to work on Unix. See https://github.com/dotnet/designs/pull/33 -->
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.0"/>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.2"/>
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net45' OR '$(TargetFramework)' == 'netstandard1.1' ">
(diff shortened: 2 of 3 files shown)


jskeet:
Hmm... I don't really want to end up adding new targets to main project every time there's a new version out. The package will grow for pretty much no reason. On the other hand, targeting .NET 5.0 forever isn't appealing either.

I think it's worth discussing what we want the long term plan to be here. @jtattermusch any thoughts?

I wonder whether it's worth splitting this PR into two: one for the test targets, and one for the production targets.

The editorconfig should be removed IMO, as it's not related to the targets and the preference for CRLF conflicts with the direction of #9526. (It may well be that a separate PR introducing an editorconfig is the simplest way of avoiding further issues like that, but let's separate the two aspects.)

Jensaarai:
Yeah, my thought was that targeting .NET 5 after May probably won't make much sense since it'll be out of support, and will probably start throwing build warnings at some point like .NET Core 2.1 now does when you build with the .NET 6 SDK.

I didn't outright remove it in this PR since it's still technically supported :)

I split the PR into three
* This one now just updating the main Google.Protobuf project (and updating `net60` to `net6.0` in the Tests project)
* https://github.com/protocolbuffers/protobuf/pull/9674 for .NET 6 and NuGet updates for the various test projects
* https://github.com/protocolbuffers/protobuf/pull/9675 to add the `.editorconfig`, with the `end_of_line ` option commented out for now and referencing https://github.com/protocolbuffers/protobuf/issues/9526

jskeet:
Note: this increases the .nupkg size by about 300K (until we can remove the .NET 5.0 target). That seems acceptable to me.

JamesNK:
A NuGet target doesn't mean a platform is supported, it just provides a minimum baseline required for the NuGet package to run. Even if net50 was removed it would still be possible to run Google.Protobuf on .NET 5 using the netstandard20 target. So this seems unnecessary.

If someone reports a problem on .NET 5 then ask them to update to .NET 6 to see whether that fixes there issue.

However, it does make sense to change unit tests to target net60 instead of net50 at some point.

jskeet:
> Even if net50 was removed it would still be possible to run Google.Protobuf on .NET 5 using the netstandard20 target. 

Yes, although without the SIMD optimization.

> And net60 happily uses net50 assemblies. So this is unnecessary. 

I agree it's not strictly necessary - but I do empathise with users who might look at the set of TFMs and be slightly surprised by net5.0 being the most recent one. (Aside from anything else, it might give the mistaken impression that the project isn't under active maintenance/development.)

So if we add net6.0 now, and then remove net5.0 in (say) 6 months, we end up with:

- .NET 5.0 users (who would be out of support by then) still being able to run, just with slightly less optimization
- .NET 6.0 users observing a net6.0 TFM and being confident in both compatibility and activity
- .NET 7.0 users observing a net6.0 TFM, which probably isn't a surprise as the latest LTS
- All other users unaffected

Additionally, it means that we don't need to have .NET 5.0 (even reference assemblies) anywhere on build machines, which is probably a good thing.

The cost is a slightly larger package until we remove the net5.0 target.

> However, it does make sense to change unit tests to target net60 instead of net50 at some point.

Definitely agree on that one, which is handled in #9674.

jtattermusch:
Since adding net6.0 target to Google.Protobuf.csproj isn't strictly necessary and seems to be only cosmetic (no extra features, no change in the way it is used), I'd be slightly leaning against it (let's add net60 once it actually adds some value?) (we could add a comment about the policy of not adding target frameworks only to create the impression of things being up to date when it doesn't really add anything).

But I don't feel strongly about this, so feel free to use your own judgement. Besides this philosophical question, the PR looks good to me.

Jensaarai:
Yeah, there are .NET 5+ features being used, maybe that could be made clearer by using the new `NET5_0_OR_GREATER` define instead of `GOOGLE_PROTOBUF_SUPPORT_FAST_STRING` and `GOOGLE_PROTOBUF_SIMD`?

https://github.com/protocolbuffers/protobuf/blob/master/csharp/src/Google.Protobuf/Google.Protobuf.csproj#L32-L34

jskeet:
It's worth noting #9692, where the target had to be changed from net6.0 to net60. I haven't looked into why yet.

Jensaarai:
> #9692

Strange...yeah I wondered why `net50` was used instead of `net5.0` etc. Do you need me to change to `net60` here? Is there a way to run those checks first in the PR?

jskeet:
@Jensaarai: I wouldn't worry for now - let's wait until I've had time to actually consider whether we want to do this. (Sorry, this is somewhat low down my priority list. I'll get there eventually, but it may not be for a while.)

jskeet:
Sorry for the delay in looking at this. I've been discussing this with colleagues, and I'd propose dropping this PR - or possibly creating a new one which just updates the PrivateAssets dependencies.

Adding a net6.0 target adds no *technical* value - it just increases the size of the NuGet package.
While it could be a stepping stone to removing net5.0, I'd actually prefer *not* to remove old targets unless there's a very compelling reason to do so.

Basically, my approach is "let's keep providing minimal support for the old targets, even if they're for out-of-support frameworks, as that does no harm" - at least within a major version. If customers report issues that *only* affect out-of-support frameworks, I think we would only put minimal time into addressing those - but it means we don't break anyone arbitrarily.

If a time comes when we're severely limited in terms of future evolution due to targeting an old framework, *then* it might be reasonable to remove support. And certainly if we take a new major version of the library, at that point it's probably worth reconsidering targets - dropping support for out-of-date targets definitely feels reasonable at that point.

Of course, this should all actually be documented as a policy somewhere, and that's a separate piece of work - but in the meantime, does it sound reasonable to you, @Jensaarai?

googleberg:
As @jskeet suggests, closing. Consider a PR that updates the PrivateAssets dependencies.

Jensaarai:
No worries, and my turn to apologize for the delay. That all makes sense to me! My only concern was the `NETSDK1138` warnings that I imagine may be triggered in the future unless `<CheckEolTargetFramework>false</CheckEolTargetFramework>` is used. Maybe just using that will be the better option anyway!

https://andrewlock.net/fixing-build-warning-netsdk1138-when-building-end-of-life-projects-with-dotnet-5/

jskeet:
@Jensaari: Thanks for the link. I'm not sure whether CheckEolTargetFramework would need to be in consuming projects, or just in protobuf. I'll do a bit of investigation - although as we've already had other changes to avoid .NET 7 warning, I *suspect* it's unnecessary.

jskeet:
I've just checked this with an entirely separate pair of projects, where the library targets `netcoreapp2.0` (which generates NETSDK1138 already).

The consuming project never shows warnings, and we can add `<CheckEolTargetFramework>false</CheckEolTargetFramework>` just to Google.Protobuf when net5.0 starts to generate warnings, *if* we decide not to update to net6.0 at that point.

(We may well try to time everything with a new major version, and overhaul the set of supported platforms at the same time.)

I think for now I'd like to *not* add the warning suppression.

PR #9669, opened 2022-03-24

dlj-NaN:


darly (approved):

jtattermusch:
Why were the XYZ_generator.h header files renamed?
It breaks grpc's build with the latest version of protobuf (e.g. see here: https://github.com/grpc/grpc/pull/29341#issuecomment-1091211391)

CC @lidizheng

dlj-NaN:
> Why were the XYZ_generator.h header files renamed? It breaks grpc's build with the latest version of protobuf (e.g. see here: [grpc/grpc#29341 (comment)](https://github.com/grpc/grpc/pull/29341#issuecomment-1091211391))
> 
> CC @lidizheng

The goal is to make our sync process simpler, and also to reduce stuttering in filenames. I'm a bit surprised that this affected grpc, since the code generators themselves are not really intended to be public. In any case, this change affects cpp, java, and python. The other code generators might change in the future.

jtattermusch:
@dlj-NaN  
AFAICT protobuf OSS doesn't have a concept of public/private headers, so all the headers can currently be considered public.
That also matches the behavior when I try to install protobuf from source. All the headers, including e.g. `google/protobuf/compiler/python/python_generator.h` end up in the `./include` directory, where everyone can reference them just as if they were part of public API.

So from gRPC's perspective, renaming a header that isn't distinguishable from other headers that are clearly part of protobuf's public API and that we've been using in the grpc codegen plugins for the past 6 years or so (basically since the initial release of protobuf 3.0.0) is a breaking change.

nicolasnoble:
Are we talking about these files ?

https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.compiler.cpp_generator
https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.compiler.python_generator
etc...

If yes, then they are listed as public APIs from the website.

devjgm:
I agree that renaming headers that are (or appear to be) public should be considered a breaking change. Protobuf should probably leave a forwarding header in the old location to avoid breaking users.

PR #9671, opened 2022-03-24
Make update_version.py accommodate per-language major versions

acozzette:
This change introduces versions.json to track a separate version for
each language and makes update_version.py read from this file to
determine what version it should set for each language.

I also had to update protobuf_version.bzl to introduce a distinction
between PROTOBUF_JAVA_VERSION and PROTOC_VERSION, since those versions
will no longer be the same.

Ideally the update_version.py script should update versions.json itself,
but for now I just set it up to read from that file only. It assumes
that the first set of versions in the list is the most recent one, and
updates all the affected files accordingly. Whenever we do a new
release, we will need to manually update versions.json so that we can
specify which major versions we want to increment.


versions.json
```diff
@@ -0,0 +1,21 @@
...
+  "versions": [
+    {
+      "date": "2022-01-27",
+      "protoc_version": "20.0-rc1",
+      "lts": false,
+      "languages": {
+        "cpp": "3.20.0-rc1",
+        "csharp": "3.20.0-rc1",
+        "java": "3.20.0-rc1",
+        "javascript": "3.20.0-rc1",
```
esorot, line 11: If we're moving towards a word, where the submodules can ideally be owned by their own team, would it make sense to have a versions.json file within each language (directory/repo) which we support?

haberman: Good question. I think `versions.json` is specifically to accommodate the current world, where multiple languages are bundled into `protoc`.

If/when we "unbundle" certain languages, so they can be in their own repo and their own codegen plugin, they will no longer need a `versions.json`. There can be a single release version for the whole repo, stored wherever it's most convenient/native for that repo/language. There will be no need to correlate a version across languages, so `versions.json` will not be necessary.

java/kotlin-lite/BUILD
```diff
@@ -1,7 +1,7 @@
 load("@io_bazel_rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library")
 load("@rules_java//java:defs.bzl", "java_lite_proto_library")
 load("@rules_jvm_external//:kt_defs.bzl", "kt_jvm_export")
-load("//:protobuf_version.bzl", "PROTOBUF_VERSION")
+load("//:protobuf_version.bzl", "PROTOBUF_KOTLIN_VERSION")
```
deannagarcia, line 4: I don't see us decoupling PROTOBUF_JAVA_VERSION and PROTOBUF_KOTLIN_VERSION since kotlin is dependent on java. I think we should just use a single PROTOBUF_JAVA_VERSION

update_version.py
```diff
@@ -1,43 +1,21 @@
 #!/usr/bin/env python
-# Usage: ./update_version.py <MAJOR>.<MINOR>.<MICRO> [<RC version>]
-#
-# Example:
-# ./update_version.py 3.7.1 2
-#   => Version will become 3.7.1-rc-2 (beta)
-# ./update_version.py 3.7.1
-#   => Version will become 3.7.1 (stable)
+# Usage: ./update_version.py
```
deannagarcia, line 2: Can you add an example of how versions will change based on what's in the file?

update_version.py
```diff
@@ -1,43 +1,21 @@
...
-Example:
-./update_version.py 3.7.1 2
-""")
-  exit(1)
-
-NEW_VERSION = sys.argv[1]
+versions = json.load(open('versions.json', 'r'))
+protoc_version = versions['versions'][0]['protoc_version'].split('-')
+RC_VERSION = len(protoc_version) > 1 and int(protoc_version[1][2:]) or -1
+NEW_VERSION = protoc_version[0]
```
deannagarcia, line 13: Can we name this PROTOC_VERSION instead to keep it clear that it's not the NEW_VERSION for all languages?

update_version.py
```diff
@@ -222,77 +192,77 @@ def UpdateCsharp():
...
     lambda document : ReplaceText(
       Find(Find(document.documentElement, 'parent'), 'version'),
-      GetFullVersion()))
+      GetFullVersion('java')))
 
   RewriteXml('java/kotlin/pom.xml',
     lambda document : ReplaceText(
       Find(Find(document.documentElement, 'parent'), 'version'),
-      GetFullVersion()))
+      GetFullVersion('kotlin')))
```
deannagarcia, line 231: Again, I think we should just use java here

deannagarcia (commented):
Is it possible to add tests to this?

PR #9675, opened 2022-03-25
Add .editorconfig for csharp directory

Jensaarai:
Nicer experience for e.g. VS Code

Split from https://github.com/protocolbuffers/protobuf/pull/9666

### DIFF
--- a/csharp/.editorconfig
+++ b/csharp/.editorconfig
@@ -1,0 +1,18 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
+root = true
+
+# C# files
+[*.cs]
+
+#### Core EditorConfig Options ####
+
+# Indentation and spacing
+indent_size = 4
+indent_style = space
+tab_width = 4
+
+# New line preferences
+# See https://github.com/protocolbuffers/protobuf/issues/9526
+#end_of_line = lf
+insert_final_newline = false
+trim_trailing_whitespace = true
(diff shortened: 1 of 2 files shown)


csharp/.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+#### Core EditorConfig Options ####
+
+# Indentation and spacing
+indent_size = 4
+indent_style = space
+tab_width = 4
+
+# New line preferences
+# See https://github.com/protocolbuffers/protobuf/issues/9526
+#end_of_line = lf
```
amanda-tarafa, line 16: Shouldn't this be uncommented?

jskeet:
Please add the file to `Makefile.am` (along with the other C# files in `csharp_EXTRA_DIST`) - that will fix the Kokoro errors.

csharp/.editorconfig
jskeet: Not yet - we may well do that as part of fixing #9526, but this PR just adds the editor without trying to address that.

Jensaarai:
Done...for my own curiosity, what is that file for? I'm guessing it's for packaging up the e.g. [protobuf-csharp-3.20.0.zip](https://github.com/protocolbuffers/protobuf/releases/download/v3.20.0/protobuf-csharp-3.20.0.zip) zip/tar files on the releases, for users who want to consume as source?

jskeet:
@Jensaarai: That's my understanding, yes. It was already in place when I started contributing to the project, and I haven't gone into great depths to check it :)

jskeet (approved):
Merging this now - it'll help with #9526.

PR #9683, opened 2022-03-27
Add support for generating C# struct for a message

### LINKED ISSUE #9682: Add support for generating a struct for a message type in C#
Follow-up of #9658

This is important when passing messages that would be better represented as plain struct in C#, to avoid lots of allocations (e.g a `Point { float x, float y }` with a `repeated Point points` on it)

The idea is to support generating struct for message types by having an option on the message:

```
message MyMessage {
    option csharp_generate_struct = true;
    // fields...
}
```

The generated code for serialization/deserialization should generate optimal code using struct with by ref/in.

xoofx:
Fixes #9682

This is a draft PR to evaluate the changes that would be required to support generating struct for a message in C#.

The PR is adding:

- A new field in a Message Options in descriptor.proto
 ```proto
  // Generates the message as a struct
  optional bool csharp_generate_struct = 10 [default = false];
 ```
- Modifies `csharp_message.cc` and `csharp_message_field.cc` to handle the serialization by ref/in.
- Add the C# runtime infrastructure to handle serializing/deserializing these structs by ref/in

Missing work:

- [x] ~Try to find a way if repeatable fields with a struct couldn't be further optimized (passing by value for now)~
  - Probably not possible without implementing a different FieldCodec
- [x] Add tests and benchmarks
- [ ] Handle properly `optional` for such field
- [ ] Handle `oneof`
- [ ] Run protoc.exe on descriptor code for all supported language (only cpp and C# for this draft)
- [ ] Add documentation about this C# feature

I might have forgotten other corner cases.

cc: @jskeet


csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -316,6 +360,24 @@ public void WriteGroup(IMessage value)
                 ctx.CopyStateTo(this);
             }
         }
+        
+        /// <summary>
+        /// Writes a group, without a tag, to the stream.
+        /// </summary>
+        /// <param name="value">The value to write</param>
+        public void WriteGroup<T>(in T value) where T: struct, IBufferMessage
```
xoofx, line 368: Probably not necessary to support WriteGroup

xoofx:
Note that the code is not compatible with `GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE`, not sure what to do with this?

Also I had to drop support for `netstandard1.1` because it doesn't support basic things like `typeof(T).IsEnum`, so can we drop support for it? It such an antique profile that I'm wondering who is using it these days?

jskeet:
> Note that the code is not compatible with GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE, not sure what to do with this? (The test checking the compilation for this mode is failing)

I'd need to look in detail, but I believe that was added for a performance improvement - it might be okay for generated structs to not support that, but I wouldn't want it to be the 

> Also I had to drop support for netstandard1.1 because it doesn't support basic things like typeof(T).IsEnum, so can we drop support for it? It such an antique profile that I'm wondering who is using it these days?

I don't know, but I wouldn't be entirely surprised. If we'd need to drop it, that will at least require much more careful consideration.

Overall even when it's working, I'm still going to want to play with this, particularly in terms of what happens if you have a struct message in a struct message, or a struct message in a repeated field... fundamentally mutable structs (particularly with properties instead of public fields) have so many sharp edges that I'd consider this a pretty dangerous feature. Yes, I can see the reason for wanting it - but I think a lot of very careful consideration will be required.

xoofx:
> Overall even when it's working, I'm still going to want to play with this, particularly in terms of what happens if you have a struct message in a struct message, or a struct message in a repeated field... fundamentally mutable structs (particularly with properties instead of public fields) have so many sharp edges that I'd consider this a pretty dangerous feature. Yes, I can see the reason for wanting it - but I think a lot of very careful consideration will be required.

Agreed. I'm starting to add tests. Repeated fields seem to work now as discovered a few bugs when testing them.

jskeet:
> btw, I'm considering that if a field in a message is a struct, I would actually generate public fields instead of property for these to allow more efficient access, so that it would use proper by ref instead of copying in and out. That would be important for bigger nested structs.

> Thoughts?

That could potentially work, yes. I'm still pretty skeptical about the whole thing right now, but I'm happy that you've got enough time to experiment :)

xoofx:
> That could potentially work, yes.

Ok, I tried but the descriptor+reflection is failing because it is expecting to go through a method to interact with such fields, so we have to keep properties.

> I'm still pretty skeptical about the whole thing right now, but I'm happy that you've got enough time to experiment :)

It's definitely touching a few pieces here and there, so it will have to go through further testing. I'm also happy to try that out, even better if it can make it to master! 😅

xoofx:
I have added a small benchmark `StructMessagesBenchmark`

```
|         Method |     Mean |     Error |    StdDev | Ratio |  Gen 0 | Allocated |
|--------------- |---------:|----------:|----------:|------:|-------:|----------:|
| ManagedMessage | 8.406 us | 0.0803 us | 0.0712 us |  1.00 | 0.3967 |   6,888 B |
|  StructMessage | 6.485 us | 0.0461 us | 0.0431 us |  0.77 | 0.0076 |     232 B |
```

xoofx:
Updated with using `ParseFrom` instead of raw reading back to an existing object:

```
|         Method |     Mean |     Error |    StdDev | Ratio |  Gen 0 | Allocated |
|--------------- |---------:|----------:|----------:|------:|-------:|----------:|
| ManagedMessage | 9.647 us | 0.0829 us | 0.0775 us |  1.00 | 0.4883 |      8 KB |
|  StructMessage | 7.025 us | 0.0620 us | 0.0580 us |  0.73 | 0.3128 |      5 KB |
```

There are more allocations because the bulk of the data is in the repeatable fields, so the struct version is now bigger (now taking into account the resizing, while before it was not, hence why it was only ~230 bytes)

Performance improvement is around 30% in that case, but there are cases where struct could be less efficient (e.g case of very large struct)

googleberg:
@jskeet @jtattermusch can you comment or close out this PR?

jskeet:
I'll close the PR for now; this definitely isn't something we'd want to do without a *lot* of research and consideration which I don't have time to do at the moment.

PR #9692, opened 2022-03-28
Reverting change to target frameworks that breaks kokoro

mkruskal-google:
Kokoro tests were broken by https://github.com/protocolbuffers/protobuf/commit/10df21ed5e06df20385f1ee8ca01ecd22e0e4087

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
+++ b/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net6.0</TargetFrameworks>
+    <TargetFrameworks>net462;netcoreapp3.1;net60</TargetFrameworks>
     <AssemblyOriginatorKeyFile>../../keys/Google.Protobuf.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <IsPackable>False</IsPackable>


dlj-NaN:
Unfortunately, we don't run aarch64 in presubmit, so I guess we'll see what happens...

@jskeet FYI. I'm not sure if this is the right change, but we did end up seeing errors in the continuous build.

dlj-NaN (approved):

mkruskal-google:
Yea I noticed that.  I took a closer look at the failure though and I'm now fairly confident this is the issue.  *crosses fingers*

jskeet:
It's somewhat strange if net6.0 works but net60 doesn't... but there's no harm in terms of the testing itself. (Basically they should mean the same thing.)

mkruskal-google:
@jskeet It was the other way around.  "net6.0" produced errors like this in our continuous testing:

`/usr/share/dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1005: Assets file '/work/csharp/src/Google.Protobuf.Test/obj/project.assets.json' doesn't have a target for 'net60'. Ensure that restore has run and that you have included 'net60' in the TargetFrameworks for your project. [/work/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]`

jskeet:
Sorry, yes - that's what I meant. Basically I'd *expect* it to work for both or neither...

mkruskal-google:
IIUC the issue was only for csharp_aarch64 which runs the tests with `dotnet publish ... -f net60 .../Google.Protobuf.Test.csproj`.  Other architectures use `netcoreapp3.1`, so they didn't hit this issue.  As we mentioned above though, we weren't able to confirm this in github so I'm not 100% sure that was the issue

mkruskal-google:
Fair enough, I wasn't sure where that was specified so this was the quicker fix.  Either way sgtm

PR #9693, opened 2022-03-29
Temporary Kokoro build for macos-next.

dlj-NaN:
This build will be used to troubleshoot building on a newer macOS
image. Once it is stable, this config will replace the "macos" build.

### DIFF
--- a/kokoro/macos-next/README.md
+++ b/kokoro/macos-next/README.md
@@ -1,0 +1,4 @@
+# macOS-next
+
+This builder is temporary for developing and testing builds using the "next"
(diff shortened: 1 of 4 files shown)


jorgbrown (commented):
The new .cfg files in kokoro/macos-next/cpp/ reference `kokoro/macos/cpp/build.sh`.  Shouldn't they be referencing `kokoro/macos-next/cpp/build.sh`?

dlj-NaN:
Good catch. Fixed.

jorgbrown (approved):

PR #9695, opened 2022-03-29
Helper method on Any to allow an any to be unpacked more easily

jskeet:
We already have the TypeRegistry abstraction for JSON parsing, so it lends itself well to this.

Note that this is much more useful than it would have been before C# gained pattern matching support: it's easy to imagine a switch statement/expression using pattern matching with the result of this, with cases for a set of known message types, for example.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/WellKnownTypes/AnyTest.cs
+++ b/csharp/src/Google.Protobuf.Test/WellKnownTypes/AnyTest.cs
@@ -30,8 +30,11 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endregion
 
+using Google.Protobuf.Reflection;
 using Google.Protobuf.TestProtos;
 using NUnit.Framework;
+using System.Linq;
+using UnitTest.Issues.TestProtos;
 
 namespace Google.Protobuf.WellKnownTypes
 {
@@ -148,10 +151,34 @@ namespace Google.Protobuf.WellKnownTypes
             Assert.False(any.Is(TestOneof.Descriptor));
         }
 
+        [Test]
         public void IsRightType()
         {
             var any = Any.Pack(SampleMessages.CreateFullTestAllTypes());
             Assert.True(any.Is(TestAllTypes.Descriptor));
+        }
+
+        [Test]
+        public void Unpack_TypeRegistry()
+        {
+            var messages = new IMessage[]
+            {
+                SampleMessages.CreateFullTestAllTypes(),
+                new TestWellKnownTypes { BoolField = true },
+                new MoreString { Data = { "x" } },
+                new MoreBytes { Data = ByteString.CopyFromUtf8("xyz") },
+                new ReservedNames { Descriptor_ = 10 }
+            };
+            var anyMessages = messages.Select(Any.Pack);
+
+            // The type registry handles the first four of the packed messages, but not the final one.
+            var registry = TypeRegistry.FromFiles(
+                UnittestWellKnownTypesReflection.Descriptor,
+                UnittestProto3Reflection.Descriptor);
+            var unpacked = anyMessages.Select(any => any.Unpack(registry)).ToList();
+            var expected = (IMessage[]) messages.Clone();
+            expected[4] = null;
+            Assert.AreEqual(expected, unpacked);
         }
     }
 }
--- a/csharp/src/Google.Protobuf/WellKnownTypes/AnyPartial.cs
+++ b/csharp/src/Google.Protobuf/WellKnownTypes/AnyPartial.cs
@@ -121,6 +121,26 @@ namespace Google.Protobuf.WellKnownTypes
         }
 
         /// <summary>
+        /// Attempts to unpack the content of this Any message into one of the message types
+        /// in the given type registry, based on the type URL.
+        /// </summary>
+        /// <param name="registry">The type registry to consult for messages.</param>
+        /// <returns>The unpacked message, or <c>null</c> if no matching message was found.</returns>
+        public IMessage Unpack(TypeRegistry registry)
+        {
+            string typeName = GetTypeName(TypeUrl);
+            MessageDescriptor descriptor = registry.Find(typeName);
+            if (descriptor == null)
+            {
+                return null;
+            }
+
+            var message = descriptor.Parser.CreateTemplate();
+            message.MergeFrom(Value);
+            return message;
+        }
+
+        /// <summary>
         /// Packs the specified message into an Any message using a type URL prefix of "type.googleapis.com".
         /// </summary>
         /// <param name="message">The message to pack.</param>


jskeet:
Note: this is very tentative at the moment. It works, but I'll need to check with the protobuf team whether this is a pattern used elsewhere, and whether there's a different name I should use for consistency, etc.

jtattermusch:
Is this ready for review? Please let me know once it is. I think an example to demonstrate the use case would be useful.

jskeet:
Yes, it's ready for review, in terms of the code - just not ready in terms of needing to check it with Protobuf team members. Will write some sample code here as a comment after a meeting...

jskeet:
Okay, here's a concrete example, based on [error details](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto( and [status](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto):

```csharp
var status = /* obtain the status from an RpcException */;
var typeRegistry = /* a type registry including error_details.proto */;
foreach (var detail in status.Details)
{
    switch (detail.UnpackMessage(typeRegistry))
    {
        case ErrorInfo errorInfo:
            Console.WriteLine($"Error info: {errorInfo.Domain} / {errorInfo.Reason}");
            Console.WriteLine($"Metadata keys: {string.Join(", ", errorInfo.Metadata.Keys)}");
            break;
        case LocalizedMessage lm:
            Console.WriteLine($"Localized message ({lm.locale}): {lm.Message}");
            break;
        case Help help:
            Console.WriteLine($"Help, first link: {help.Links.FirstOrDefault().Url}");
            break;
        case BadRequest badReq:
            Console.WriteLine($"Bad request; broken fields: {string.Join(", ", badReq.FieldViolations.Select(fv => fv.Field))}");
            break;
        case IMessage unhandled:
            Console.WriteLine($"Known but unhandled message: {unhandled}");
            break;
        default
            Console.WriteLine($"Unknown detail type URL: {detail.TypeUrl}");
            break;
    }
}
```

(This code hasn't been compiled or tested, so may have some typos, but should get across the utility of not having to call `Any.TryUnpack` lots of times.)

jtattermusch (approved):
Seems useful. LGTM for the C# code. Please check the general idea with the protobuf team.

csharp/src/Google.Protobuf/WellKnownTypes/AnyPartial.cs
```diff
@@ -120,6 +120,26 @@ public bool TryUnpack<T>(out T result) where T : IMessage, new()
             return true;
         }
 
+        /// <summary>
+        /// Attempts to unpack the content of this Any message into one of the message types
+        /// in the given type registry, based on the type URL.
+        /// </summary>
+        /// <param name="registry">The type registry to consult for messages.</param>
+        /// <returns>The unpacked message, or <c>null</c> if no matching message was found.</returns>
+        public IMessage UnpackMessage(TypeRegistry registry)
```
JamesNK, line 129: Why UnpackMessage instead of Unpack?

JamesNK: (I'm guessing having multiple Unpack overloads with different return types could be confusing)

jskeet: Precisely your guess. Although that doesn't mean I'm unwilling to discuss it :)

amanda-tarafa, line 129: Another overload for this could be parameter-less, and could attempt to build the TypeRegistry from all dependencies. A little heavy on reflection but a type registry built like that could be lazy and cached.

But this of course, could not be done at all, or in another PR.

amanda-tarafa (approved):
LGTM

csharp/src/Google.Protobuf/WellKnownTypes/AnyPartial.cs
JamesNK: Do you mean assembly scanning to find all IMessage types?

JamesNK: I don't know how much president there is for different return types in overloads from .NET APIs or ASP.NET Core APIs. Is there is an example of a popular API that has different return types? I can't think of one off the top of my head.

jskeet: There's some precedent for this, in terms of `Enum.Parse` vs `Enum.Parse<TEnum>`. But it's relatively rare.

amanda-tarafa: Yep, that's what I meant, assembly scanning to find all descriptors from which to build a TypeRegistry. As I said heavy on the reflection :).

Chatting to Jon this morning, he mentioned maybe adding a `TypeRegistry.LoadTheWorld()` and then use that `TypeRegistry` here, which would make it super explicit.

But again, either variant is just an extra feature on top of this one, so that calling code doesn't even have to know how to build the appropiate `TypeRegistry` that will contain the potential actual types of `Any`.

JamesNK: IMO I'd name them the same. They both unpack a message, just the arguments to do it (generic vs TypeRegistry) are different.

JamesNK: Ok. We (.NET team) try to avoid assembly scanning these days. It isn't trimmer/AOT friendly. `TypeRegistry.LoadTheWorld()` is ok-ish because it isolates the assembly scanning to one API.

mkruskal-google: +1 to Unpack.  Both of these methods "unpack a message" as James said, so I find it more confusing for only one of them to be named "UnpackMessage".  I supposed something like "UnpackFromTypeRegistry" would work too, but that seems a bit verbose

csharp/src/Google.Protobuf.Test/WellKnownTypes/AnyTest.cs
```diff
@@ -148,10 +151,35 @@ public void IsWrongType()
...
+                new MoreString { Data = { "x" } },
+                new MoreBytes { Data = ByteString.CopyFromUtf8("xyz") },
+                new ReservedNames { Descriptor_ = 10 }
+            };
+            var anyMessages = messages.Select(Any.Pack);
+
+            // The type registry handles the first four of the packed messages, but not the final one.
+            var registry = TypeRegistry.FromFiles(
+                UnittestWellKnownTypesReflection.Descriptor,
+                UnittestWellKnownTypesReflection.Descriptor,
```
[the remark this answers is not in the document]
mkruskal-google: +1, isn't this redundant?

mkruskal-google (changes requested):

csharp/src/Google.Protobuf/WellKnownTypes/AnyPartial.cs
jskeet: Renamed to Unpack.

csharp/src/Google.Protobuf.Test/WellKnownTypes/AnyTest.cs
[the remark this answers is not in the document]
jskeet: Yup. I think when I originally wrote the test I did have some reason in mind, but it's gone now :)

mkruskal-google:
This LGTM, but let's go through a lightweight Tier 2 review first since this is a public API change

jskeet:
> This LGTM, but let's go through a lightweight Tier 2 review first since this is a public API change

Yup. Will do that on Tuesday. (UK has public holidays Friday/Monday.)

mkruskal-google (approved):

PR #9697, opened 2022-03-29
Add JRuby 9.3.4.0 to Dockerfile and use it for 9.3.x testing.

JasonLunn:


kokoro/linux/dockerfile/test/ruby/Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN /bin/bash -l -c "rvm install 3.0.0"
 RUN /bin/bash -l -c "rvm install 3.1.0"
 RUN /bin/bash -l -c "rvm install jruby-9.2.20.1"
 RUN /bin/bash -l -c "rvm install jruby-9.3.3.0"
```
perezd, line 40: Is this one used elsewhere?

JasonLunn: It is used currently, and it wasn't clear to me if I removed it immediately if it would be disruptive for other PRs.

perezd (approved):

PR #9708, opened 2022-03-30
Remove libpthread CMake dependency

redbaron:
libpthread is not used by the code, yet CMake continued to look for it.

@acozzette , as discussed in https://github.com/protocolbuffers/protobuf/pull/9100#issuecomment-1080533383

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,8 +160,6 @@ else()
 endif()
 file(REMOVE ${CMAKE_CURRENT_BINARY_DIR}/cmaketest.map)
 
-find_package(Threads REQUIRED)
-
 set(_protobuf_FIND_ZLIB)
 if (protobuf_WITH_ZLIB)
   find_package(ZLIB)


acozzette:
@redbaron Thanks! Looks good to me. We're having a temporary problem with our CI builds, but once that is fixed I will go ahead and merge this, assuming the tests pass.

acozzette:
It looks like I was mistaken and apparently we still do need pthreads for thread-local storage: https://github.com/protocolbuffers/protobuf/blob/520c601c99012101c816b6ccc89e8d6fc28fdbb8/src/google/protobuf/stubs/mutex.h#L177 So I'm afraid we will probably have to keep this CMake dependency on pthreads.

redbaron:
@acozzette , this file seems to be gone and there no reference to `pthread_` functions anymore

acozzette:
@redbaron I think you are right, we can probably drop the pthread dependency at this point. I see that we still have an `#include` of pthread.h in src/google/protobuf/stubs/common.h, but it looks like it is unnecessary.

PR #9722, opened 2022-04-01
[Ruby] Fix RepeatedField#last, #first inconsistencies

lucthev:
`RepeatedField#last(n)` has an off-by-one error resulting in one more element being returned than requested.

While I was at it, I also fixed the following inconsistencies with `Array`:

- `last(n)` where n > size should return all elements; the current implementation wraps around such that e.g. `last(size + 2) == last(2)`
- `last(n)` where n < 0 should raise an `ArgumentError`

### DIFF
--- a/ruby/lib/google/protobuf/repeated_field.rb
+++ b/ruby/lib/google/protobuf/repeated_field.rb
@@ -79,12 +79,25 @@ module Google
 
 
       def first(n=nil)
-        n ? self[0...n] : self[0]
+        if n.nil?
+          return self[0]
+        elsif n < 0
+          raise ArgumentError, "negative array size"
+        else
+          return self[0...n]
+        end
       end
 
 
       def last(n=nil)
-        n ? self[(self.size-n-1)..-1] : self[-1]
+        if n.nil?
+          return self[-1]
+        elsif n < 0
+          raise ArgumentError, "negative array size"
+        else
+          start = [self.size-n, 0].max
+          return self[start..-1]
+        end
       end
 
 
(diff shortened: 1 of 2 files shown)


fowles (approved):

fowles:
This looks good.  I will merge it after our test suite finishes

fowles:
Looks like the failure is in jruby.  Jason?

JasonLunn:
It appears that the fix for this issue has uncovered a separate bug in the JRuby implementation of the `at` (aka `[]`) method of [`RubyRepeatedField`](https://github.com/protocolbuffers/protobuf/blob/main/ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java#L131). It appears that the implementation is making assumptions [here](https://github.com/protocolbuffers/protobuf/blob/main/ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java#L145) about the range values that don't hold if you use negative indices to count back from the end of the sequence.

@lucthev - if you're comfortable working in Java, you can fix both issues in the same PR. If not, please open a separate ticket for the JRuby-specific issue and I can work up a patch.

fowles:
rerunning test suite.  Will come back tomorrow to see how it went

lucthev:
Thanks for the reviews! I had a stab at fixing the issue in `RubyRepeatedField#[]` but there's a lot of odd edge-cases (and I'm not very familiar with Java) so I didn't get very far. I've opened https://github.com/protocolbuffers/protobuf/issues/9751 to track that issue just for completeness (it's pretty niche anyways).

In the meantime, I've worked around that in this PR by not using a negative index.

PR #9734, opened 2022-04-04

mkruskal-google:


dlj-NaN (approved):

guodongxiaren:
Why remove 'virtual' keyword from CopyFrom……

mkruskal-google:
This was part of an internal optimization effort I believe.  Marking methods virtual isn't free, and there were no supported cases overriding it anyway

PR #9739, opened 2022-04-05
Escape GetObject macro inside protoc-generated code

smertnik3sh:
GetObject macro is defined in wingdi.h (included transitively in windows.h):

`#ifdef UNICODE
#define GetObject  GetObjectW
#else
#define GetObject  GetObjectA
#endif`

This can cause unexpected problems when compiling "GetObject" contains proto for windows.

Also "GetObject" looks like common method name for grpc. [Here](https://github.com/grpc/grpc/pull/29311) I want to wrap grpc-generated code with port_def/port_undef

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -779,6 +779,8 @@
 #undef GetClassName
 #pragma push_macro("GetMessage")
 #undef GetMessage
+#pragma push_macro("GetObject")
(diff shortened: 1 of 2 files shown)


fowles (approved):

fowles:
Thanks for the contribution.  I will merge after the tests finish

smertnik3sh:
thnx

Issue #7273, opened 2020-03-02
protobuf-java-util JsonFormat does not allow gson builder override


fsilberstein:
Until version v3.6.1 of protobuf-java-util, the gson builder was calling `disableHtmlEscaping()`. After this version, it was removed and now the serialisation encodes everything.
This sounds like a breaking change.
Would it be possible to override the default gson builder in the JsonFormat implementation?

**What version of protobuf and what language are you using?**
Version: v3.6.1 -> v3.7.0
Language: Java

**What did you do?**
We used v3.6.1 of protobuf and used to test the JSON serialisation like this:

```protobuf
// with the following proto schema generated
message Address {
    string area_code = 1;
}
```

```Java
  @Test
  public void shouldSerialiseTest() throws InvalidProtocolBufferException {
    String expected = "{\"area_code\":\"my url with an = in the text\"}";

    UserEvent.AddressOrBuilder address = UserEvent.Address.newBuilder()
         .setAreaCode("my url with an = in the text")
         .build();

    String serialised = JsonFormat.printer()
        .omittingInsignificantWhitespace()
        .preservingProtoFieldNames()
        .print(address);

    assertEquals(expected, serialised);
  }
```
**What did you expect to see**
Test was passing in v3.6.1 and was expected to pass in v3.7.0 and the following versions. 

**What did you see instead?**
This error came up:
![image](https://user-images.githubusercontent.com/8374472/75664277-7f4cfb00-5c72-11ea-9cd2-335fc553e9ec.png)
The equal sign was replaced by a unicode character.

**Anything else we should know about your project / environment**
I searched the root cause and found the change in this commit  https://github.com/protocolbuffers/protobuf/commit/6bbe197e9c1b6fc38cbdc45e3bf83fa7ced792a3
and more specifically https://github.com/protocolbuffers/protobuf/commit/6bbe197e9c1b6fc38cbdc45e3bf83fa7ced792a3#diff-b1fbf341ec74164b260e02f8b4002d4dL613

PR #9740, opened 2022-04-05
Add the `withoutHtmlEscaping` option to `JsonFormat` printer

azurvii:
Related: #7273 (closed but not fixed / discussed).

Since 3.7.0, `JsonFormat.printer()` would only use the default `Gson` instance, which would escape characters like `=`. Not all use cases of `JsonFormat.printer()` is related to HTML. An option to not escape these characters would be great.

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -115,7 +115,8 @@ public class JsonFormat {
         /* preservingProtoFieldNames */ false,
         /* omittingInsignificantWhitespace */ false,
         /* printingEnumsAsInts */ false,
-        /* sortingMapKeys */ false);
+        /* sortingMapKeys */ false,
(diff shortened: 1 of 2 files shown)


elharo:
Thanks for the suggestion, but this isn't going to be accepted. There are security reasons not to provide this as an option. In addition we don't want to clutter the API with semantically optional syntax choices. You can post-process the JSON in your own code if for some reason this is really important in your application.

HuanjieGuo:
How about the current situation? it is totally unreadable and could the team pls tell us how to convert the message without escape the html?

devinrsmith:
I'm curious more about the security reasons not to provide the option. At least in the context where I'm operating, there is no HTML.

> Thanks for the suggestion, but this isn't going to be accepted. There are security reasons not to provide this as an option.

I think this argument implies that Gson itself should also not be providing the option. Should we be looking to file a security issue for the removal of `com.google.gson.GsonBuilder#disableHtmlEscaping`?

PR #9741, opened 2022-04-05
LAST_RELEASED=3.20.0

elharo:
@acozzette We might want to add this to the release process since it seems this hasn't been updated for quite a while.

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -6,7 +6,7 @@
 
 # For when some other test needs the C++ main build, including protoc and
 # libprotobuf.
-LAST_RELEASED=3.9.0
+LAST_RELEASED=3.20.0
 
 internal_build_cpp() {
   if [ -f src/protoc ]; then


elharo:
Possibly. The variable is used though for C# and PHP.

elharo:
This fails though it's not immediately obvious how this change leads to this failure:

```
========================================================================
    Checking on the ObjC Runtime Code
========================================================================
objectivec/DevTools/full_mac_build.sh: objectivec/DevTools/pddm_tests.py: /usr/bin/python3: bad interpreter: No such file or directory
```

fowles:
that is a pre-existing failure

acozzette:
I think things get complicated when protobuf generated code is published to package managers like Maven Central and PyPI. If we required all generated code to have been generated by the same protoc version matching the runtime, then dependency management would become difficult because any packages containing protobuf generated code would all have to be upgraded in lockstep with the protobuf runtime. Also if we release a new runtime without a major version increment, then we're implicitly promising that it's safe to upgrade, so I think we do need to tolerate a certain degree of mismatch between protoc version and runtime version.

However, I'm not sure we should keep the compatibility tests, because if we're not actively updating them then they're not necessarily adding much value.

fowles:
I am nearly certain that this workflow doesn't work for C++ and is unsupported for Java.

elharo:
The tests are passing now, which is good. I'm still not sure whether we want to keep doing this or drop the compatibility tests.

elharo:
So how do we want to move forward with this?

Option A: Update the variable and add this to the release process
Option B: Delete the variable and everything that depends on it

fowles:
I lean towards B

fowles:
if you want to merge this so it is at least unbroken first, we can do that

fowles (approved):

elharo:
OK, I'll merge and then try a PR to reomve it completely and see what happens.

PR #9745, opened 2022-04-05
Tweak how python/python3 is used.

thomasvl:
The kokoro Mac configs seem to be pretty old without python3, tweak how things
are invoked to try and better deal with those configs.


thomasvl:
This is followup to #9737

fowles (commented):
LGTM

PR #9749, opened 2022-04-06
Fix off-by-one in RepeatedMessage#last

bouk:
It was returning one item too many because of the extra `-1`


bouk:
closed in favor of https://github.com/protocolbuffers/protobuf/pull/9722

PR #9750, opened 2022-04-06
Add k8 toolchain

deannagarcia:
Add a toolchain for k8 that is the same as x86-64 to prevent some compiler errors.
Add a feature to support dynamic toolchains for linux/macos.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -157,6 +157,21 @@ def _impl(ctx):
       ],
   )
 
+  features = [linker_flags, compiler_flags, sysroot_flags]
+
+  if "mingw" in ctx.attr.target_full_name:
+      features.append(
+          feature(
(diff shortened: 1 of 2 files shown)


toolchain/cc_toolchain_config.bzl
```diff
@@ -157,6 +157,20 @@ def _impl(ctx):
...
   )
 
+  features = [linker_flags, compiler_flags, sysroot_flags]
+
+  if "mingw" in ctx.attr.target_full_name:
+      features.append(
+          feature(
+              name = "targets_windows",
+              enabled = True,
+          )
```
fowles, line 167: please be consistent with multi-line vs single line formatting for this

fowles (approved):

toolchain/cc_toolchain_config.bzl
deannagarcia: Fixed

PR #9754, opened 2022-04-07
Add Google Gnostic Registered Extension Number

jeffsawatzky:
google/gnostic's protoc-gen-openapi generator has extension support for additional openapi options.

It is [currently using `1042`](https://github.com/google/gnostic/blob/master/openapiv3/annotations.proto) which is registered with grpc-gateway's protoc-gen-openapiv2 extension which probably isn't a good thing.

See [here](https://github.com/google/gnostic/blob/master/cmd/protoc-gen-openapi/examples/tests/openapiv3annotations/message.proto) for an example of the extenstion in action.


jeffsawatzky:
@acozzette / @deannagarcia / @haberman

acozzette (approved):

PR #9769, opened 2022-04-08
Do not use range based UTF-8 validation in truffleruby

ntkme:
This PR fixes the `missing LLVM builtin: llvm.x86.ssse3.pshuf.b.128` in https://github.com/oracle/truffleruby/issues/2640 by forcing it to compile with old `utf8_naive` when [the macro `TRUFFLERUBY`](https://github.com/oracle/truffleruby/blob/master/doc/user/compatibility.md#identification) is defined.


haberman (approved):

ntkme:
Upstream has fixed the bug, but it won't be able to make the fix into the next release of truffleruby later this month. So it will be another 3~4 months for the upstream fix to land into a stable release.

eregon:
Both fixes are compatible (they both disable the SSE4_1 path on TruffleRuby), so I think it makes sense to merge this.

PR #9777, opened 2022-04-12
Replace master with main in URLs & docs

noahdietz:
Since the default branch of the repo has changed from `master` to `main` (as well as in some other repos referenced in docs), all such references in docs/comments should be changed. 

Note: I did not regenerate anything i.e. using `generate_docs.py`.


fowles (approved):

fowles:
Thanks for the cleanup!

PR #9780, opened 2022-04-12
Change enum string name for reserved words

### LINKED ISSUE #9763: PHP: string form of enum contains "PB" prefix used to differentiate the constant from reserved words
In #3612 the step was taken to protect generated PHP code for Enums from use of PHP reserved words that would cause interpretation exceptions. e.g. an Enum value named `CLASS = 1` would be generated as `PBCLASS` in PHP. This did the job.

However, as part of #5342, the string value that the PHP constant maps to _also_ includes the `PB` prefix (which was prepended to the constant identifier to disambiguate it). Use of the `name` helper to map from Enum value to the string form retrieves a name that doesn't represent the proto. The same example as above would result in `"PBCLASS"` as the name, while the name for the value in the proto is `CLASS`. Here is a specific [example in Google Ads](https://github.com/googleads/google-ads-php/blob/5d8c5837ccbb9284b42182e1260376609ba2e6f3/src/Google/Ads/GoogleAds/V10/Enums/ConversionActionCategoryEnum/ConversionActionCategory.php#L168).

Is this ideal? In my opinion it seems like a misrepresentation of the enum value names.

Could the `PB` prefix be excluded from the generated string form? I could see this being perceived as a breaking change if users were depending on the emergent behavior of names including the prefix.

I believe this is only really a concern for users of the generated `name` helper in enum classes. The JSON serialization code uses the proper string form I believe. 

cc: @fiboknacky @bshaffer @dwsupplee

noahdietz:
Changes the `string` name associated with an enum that uses a reserved word to be the enum name _without_ the `PB` prefix that the constant gets. For example the enum value `CLASS` would be generated as `const PBCLASS` and the `string` name returned by the `name()` helper for it will be `'CLASS'` instead of `'PBCLASS'`.

The `value` helper will accept either `'CLASS'` or `'PBCLASS'` because it will attempt to prefix the string with `'PB'` before looking it up again. This "extra check" logic will only be generated for those enums that use a reserved word as a value.

This is a **breaking change** to generated PHP code that depended on the value returned by `name()` being prefixed with `PB`. This is a rare use case. Furthermore, the `name()` function exists for two primary reasons:

1. The REST API accepts the string names, not the constants
2. To make the enum values more easily human-readable in debugging

For the 1st case, the `"PB"` prefix would throw an error in the REST API, and for the 2nd case, there is no reliant code on the prefix behavior. Worth noting, `serializeToJsonString` **does not** use the `name()` function to serialize the enum into the JSON blob. 

Thank you @bshaffer for the concise words describing the scenarios in question.

Fixes #9763

### DIFF
--- a/php/src/Google/Protobuf/Field/Cardinality.php
+++ b/php/src/Google/Protobuf/Field/Cardinality.php
@@ -59,12 +59,8 @@ class Cardinality
     {
         $const = __CLASS__ . '::' . strtoupper($name);
         if (!defined($const)) {
-            $pbconst =  __CLASS__. '::PB' . strtoupper($name);
-            if (!defined($pbconst)) {
-                throw new UnexpectedValueException(sprintf(
-                        'Enum %s has no value defined for name %s', __CLASS__, $name));
-                }
-                return constant($pbconst);
+            throw new UnexpectedValueException(sprintf(
+                    'Enum %s has no value defined for name %s', __CLASS__, $name));
         }
         return constant($const);
     }
--- a/php/src/Google/Protobuf/Field/Cardinality.php
+++ b/php/src/Google/Protobuf/Field/Cardinality.php
@@ -59,8 +59,12 @@ class Cardinality
     {
         $const = __CLASS__ . '::' . strtoupper($name);
         if (!defined($const)) {
-            throw new UnexpectedValueException(sprintf(
-                    'Enum %s has no value defined for name %s', __CLASS__, $name));
+            $pbconst =  __CLASS__. '::PB' . strtoupper($name);
+            if (!defined($pbconst)) {
+                throw new UnexpectedValueException(sprintf(
+                        'Enum %s has no value defined for name %s', __CLASS__, $name));
+                }
+                return constant($pbconst);
         }
         return constant($const);
     }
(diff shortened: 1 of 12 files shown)


noahdietz:
cc @bshaffer

noahdietz:
I understand the breaking nature of this change, while hyper focused and could be considered fixing already broken generated code, might be controversial. I opened this to get the ball rolling after a bit of discussion in the issue.

bshaffer (approved):

noahdietz:
cc @haberman

haberman:
Please be more specific in the PR description about exactly what cases will break backward compatibility. As I understand it, the cases of breakage are rare.

noahdietz:
> Please be more specific in the PR description about exactly what cases will break backward compatibility. As I understand it, the cases of breakage are rare.

Done!

noahdietz:
Hi @haberman what can we expect on timing here? Will this be accepted for the upcoming release or afterwards? Thanks!

haberman:
We will merge this soon and it will go into the .21.0 release early next month.

I think the description somewhat overstates the breakage. As I understand it we expect it to be **very rare** that anyone will actually experience breakage, because:

1. If anyone was using this for JSON, it would have already been broken, since JSON won't accept names like `PBCLASS`.
2. If anyone is using the string for human-readable output, this change will not break them.

So we don't really know of any cases were we expect this to be truly breaking.

haberman (approved):

PR #9784, opened 2022-04-13
fix: phpdoc syntax for repeatedfield parameters

bshaffer:
The current way we document `RepeatedField` parameters is [ambiguous in PHPDoc](https://phpstan.org/writing-php-code/phpdoc-types#iterables). For `Collection|Foo[]`, two possible paths are taken:

 1. If `Collection` implements `Traversable` (which `RepeatedField` does), `Collection|Foo[]` is interpreted as a `Collection` object that iterates over `Foo`. The array part isn’t applied. This is what is currently happening for Protobuf.
 2. If Collection|Foo[] means “Collection or array” (which is what we mean in our case) we need to disambiguate the type by using `Collection|array<Foo>` instead.

This PR replaces all instance of `ProtoClass[]|RepeatedField` with the less-ambiguous `array<ProtoClass>|RepeatedField`

### DIFF
--- a/php/src/Google/Protobuf/Api.php
+++ b/php/src/Google/Protobuf/Api.php
@@ -92,9 +92,9 @@ class Api extends \Google\Protobuf\Internal\Message
      *     @type string $name
      *           The fully qualified name of this interface, including package name
      *           followed by the interface's simple name.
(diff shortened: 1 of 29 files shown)


perezd (approved):
LGTM!

Issue #9782, opened 2022-04-13
Undocumented breaking change in 3.20.0 for Python - numpy singleton arrays


tommilligan:
**What version of protobuf and what language are you using?**
Version: v3.20.0
Language: Python

**What operating system (Linux, Windows, ...) and version?**
Ubuntu 20.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
libprotoc 3.19.4, installed from https://github.com/protocolbuffers/protobuf/releases/download/v3.19.4/protoc-3.19.4-linux-x86_64.zip
Python 3.8.10
numpy 1.19.5

**What did you do?**

In version 3.19.4, it was valid to assign a numpy ndarray of shape `(1,)` to a scalar field, such as `float`:

```python
In [1]: from google.protobuf.wrappers_pb2 import FloatValue

In [2]: import numpy as np

In [3]: m = FloatValue(value=np.array([1.0]))

In [4]: m
Out[4]: value: 1.0
```

In version 3.20.0, this is now a `TypeError`:

```python
In [3]: from google.protobuf.wrappers_pb2 import FloatValue

In [4]: import numpy as np

In [5]: m = FloatValue(value=np.array([1.0]))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [5], in <cell line: 1>()
----> 1 m = FloatValue(value=np.array([1.0]))

TypeError: array([1.]) has type numpy.ndarray, but expected one of: int, float
```

This looks to have been introduced deliberately in this commit: https://github.com/protocolbuffers/protobuf/commit/3ab97ce8301183b46a4412b28171512f00f8edb9#diff-b5df04532fab1fea63308b960106e1a5ff4318875722e6bbd6200c25c56c94d7R162-R164

I think the new behaviour is correct (it should be the caller's responsibility to take the first element of the array), but I just hit this in the wild, and expect others will do the same.

**What did you expect to see**

I expected this behaviour to still be valid in 3.20.0, or for there to a breaking change documented in the changelog.

**What did you see instead?**

`TypeError`

**Anything else we should know about your project / environment**

N/A

zhangskz:
Updated release notes for 3.20.0 and 3.20.0-rc1 to include this behavior change. Also have a PR in progress to update CHANGES.txt. Looks like this was somehow missed in the sync to github, even though release notes were set in the internal CL.

tommilligan:
That sounds perfect, thanks very much.

PR #9788, opened 2022-04-14
Add disallowment of setting numpy singleton arrays and multi-dimensio…

zhangskz:
Add disallowment of setting numpy singleton arrays and multi-dimensional arrays to fields/repeated fields to CHANGES.txt, which was missed in a previous git sync and 3.20.0 release (release notes have been updated). This is a minor breaking change and should be documented. (#9782)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -73,8 +73,8 @@
     JSON serializations with UTF-8 or other non-ASCII encodings.
   * Added experimental support for directly assigning numpy scalars and array.
   * Improve the calculation of public_dependencies in DescriptorPool.
-  * Disallow setting fields to numpy singleton arrays or repeated fields to numpy
-    multi-dimensional arrays.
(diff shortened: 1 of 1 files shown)


CHANGES.txt
```diff
@@ -73,6 +73,8 @@
     JSON serializations with UTF-8 or other non-ASCII encodings.
   * Added experimental support for directly assigning numpy scalars and array.
   * Improve the calculation of public_dependencies in DescriptorPool.
+  * Disallow setting fields to numpy singleton arrays or repeated fields to numpy
+    multi-dimensional arrays.
```
fowles, line 77: can you call out more clearly that this is a breaking change and what the solution is?

zhangskz: Done. Also fixed in release notes.

fowles (approved):

PR #9791, opened 2022-04-15
Move private utf8 helpers to anonymous namespace

cyb70289:
This trivial change hints compiler to inline the utf8 helpers and
generate code with better performance.

Tested on skylake, with clang-12 and gcc-9.3. Observed about 10%
improvement from `google_message1_proto3_serialize` benchmark,
with default ./configure parameters. Similar improvement on arm.


acozzette (approved):

acozzette:
@cyb70289 Thanks!

PR #9793, opened 2022-04-15
Update CMake configuration to add a dependency on Abseil

mkruskal-google:
This is the most recent Abseil LTS branch, from November 2021. We do not yet use Abseil for anything, but this change will make it possible to start using it in CMake when we're ready.  This PR was tested by confirming that using absl::string_view from protobuf::Message doesn't break anything (under CMake builds only).


Makefile.am
```diff
@@ -1403,6 +1403,7 @@ EXTRA_DIST = $(@DIST_LANG@_EXTRA_DIST)              \
   BUILD                                             \
   WORKSPACE                                         \
   CMakeLists.txt                                    \
+  cmake/abseil-cpp.cmake                           \
```
fowles, line 1406: off by one error in whitespace!!

fowles (approved):

Makefile.am
mkruskal-google: damn, I've gotten too used to cider's auto format :(

mkruskal-google:
What does?  The Windows Python Release kokoro build is the only one failing, mainly because it's using CMake 3.10.1 which doesn't support submodules (according to gRPC at least)

cmake/libprotobuf.cmake
```diff
@@ -116,6 +116,9 @@ if(protobuf_HAVE_LD_VERSION_SCRIPT)
     LINK_DEPENDS ${protobuf_SOURCE_DIR}/src/libprotobuf.map)
 endif()
 target_link_libraries(libprotobuf PRIVATE ${CMAKE_THREAD_LIBS_INIT})
+target_include_directories(libprotobuf
```
acozzette, line 119: I think we probably need a similar change in libprotobuf-lite.cmake.

kokoro/release/python/linux/build_artifacts.sh
```diff
@@ -30,7 +30,7 @@ git clone https://github.com/matthew-brett/multibuild.git
 # silently creeping in (see https://github.com/protocolbuffers/protobuf/issues/9180).
 # IMPORTANT: always pin multibuild at the same commit for:
 # - linux/build_artifacts.sh
-# - linux/build_artifacts.sh
+# - macos/build_artifacts.sh
```
acozzette, line 33: Were these changes meant to be included? I think this PR must have been based on the 3.20.x branch because it seems to be including some commits that are on 3.20.x but haven't been merged to main yet.

mkruskal-google: No this was based on main.  I just came across these and got confused, and thought I should update them

cmake/libprotobuf.cmake
mkruskal-google: Ah ofc, I was testing with protobuf::Message so that didn't get picked up

kokoro/release/python/linux/build_artifacts.sh
acozzette: Ah ok, I see. It's still strange that a bunch of unrelated commits show up here: https://github.com/protocolbuffers/protobuf/pull/9793/commits But I guess if you do "squash and merge" then it won't matter.

acozzette (approved):

kokoro/release/python/linux/build_artifacts.sh
mkruskal-google: Ah yea I think I was originally off of 3.20.x and switched to main in commit 72c6de816143fcae34e24c4a2b1ae3d8b70b05de.  Squash and merge should keep it hidden though :)

fowles (approved):

haberman:
This PR appears to have broken the Kokoro release artifacts job (eg. https://fusion.corp.google.com/projectanalysis/summary/kokoro/prod%3Aprotobuf%2Fgithub%2Fmaster%2Frelease%2Fprotoc%2Fwindows%2Frelease)

```
CMake Error at protoc.cmake:11 (add_executable):

  Target "protoc" links to target "absl::strings" but the target was not

  found.  Perhaps a find_package() call is missing for an IMPORTED target, or

  an ALIAS target is missing?

Call Stack (most recent call first):

  T:/[src/github/protobuf/CMakeLists.txt:323](https://cs.corp.google.com/piper///depot/google3/src/github/protobuf/CMakeLists.txt?l=323) (include)

  CMakeLists.txt:9 (include)

[...]

LINK : fatal error LNK1104: cannot open file 'absl::strings.lib' [T:\src\github\protobuf\build32\protoc.vcxproj]
```

PR #9799, opened 2022-04-18
Improve range handling logic of `RepeatedField`

### LINKED ISSUE #9751: [JRuby] RepeatedField#[] inconsistencies with negative ranges
This is a follow-up to https://github.com/protocolbuffers/protobuf/pull/9722#issuecomment-1088937852.

The `RubyRepeatedField#index` implementation assumes that ranges have a positive size:

https://github.com/protocolbuffers/protobuf/blob/d54cafb7c8dffab9cb54d23ed3997207760773dc/ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java#L145-L147

However that doesn't always hold for ranges involving negative indexes:

```rb
a = [1,2,3]
a[0..-1] # => [1, 2, 3]
(0..-1).size # => 0
```

https://github.com/protocolbuffers/protobuf/compare/main...lucthev:range-access shows some failing cases (the first one, `0..5`, works fine, but the others fail). There's a lot of odd edge-cases here, and it might not be worth supporting them all.

JasonLunn:
Adds support for negative start or end and exclusive ends.

Includes regression tests contributed mostly by @lucthev. Fixes #9751.

### DIFF
--- a/ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java
+++ b/ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java
@@ -141,12 +141,12 @@ public class RubyRepeatedField extends RubyObject {
       } else if (arg instanceof RubyRange) {
         RubyRange range = ((RubyRange) arg);
 
-        int beg = RubyNumeric.num2int(range.first(context));
-        int len = RubyNumeric.num2int(range.size(context));
+        int beg = normalizeArrayIndex(range.first(context));
+        int last = normalizeArrayIndex(range.last(context));
 
-        if (len == 0) return context.runtime.newEmptyArray();
+        if (last-beg < 0) return context.runtime.newEmptyArray();
 
-        return this.storage.subseq(beg, len);
+        return this.storage.subseq(beg, last - beg + (range.isExcludeEnd() ? 0 : 1));
       }
     }
     /* assume 2 arguments */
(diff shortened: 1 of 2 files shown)


ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java
```diff
@@ -141,12 +141,12 @@ public IRubyObject index(ThreadContext context, IRubyObject[] args) {
       } else if (arg instanceof RubyRange) {
         RubyRange range = ((RubyRange) arg);
 
-        int beg = RubyNumeric.num2int(range.first(context));
-        int len = RubyNumeric.num2int(range.size(context));
+        int beg = normalizeArrayIndex(range.first(context));
+        int last = normalizeArrayIndex(range.last(context));
 
-        if (len == 0) return context.runtime.newEmptyArray();
+        if (last-beg < 0) return context.runtime.newEmptyArray();
```
perezd, line 147: style nit: can you add braces for this?

perezd (approved):
LGTM!

ruby/src/main/java/com/google/protobuf/jruby/RubyRepeatedField.java
JasonLunn: Done.

PR #9800, opened 2022-04-18
Add a github action workflow to run CocoaPods validation.

thomasvl:
- Run for all Apple configurations/platforms.
- Limit the action to only run when the ObjC source change.
- Document why the CocoaPods doesn't run the tests during the validation.

The kokoro objectivec_cocoapods_integration configuration can likely be retired
by retiring this.


perezd (approved):
LGTM!

PR #9819, opened 2022-04-20
fix: move composer.json out of root to avoid confusion

bshaffer:


haberman:
I'll note that per offline discussion, we are keeping two copies of `composer.json` because the `scripts` in our main copy will not work in the https://github.com/protocolbuffers/protobuf-php repo.

PR #9821, opened 2022-04-20
include python/google/__init__.py to python_srcs for bazel rules

Zaspire:
google.* python's name space is shared between packages(e.g. google.oauth2, google.cloud ...)
pkg_resources magic required in all google/__init__.py files for it to work.

if src/google/__init__.py is not included, bazel will generate an empty __init__.py.
which breaks import of other modules, in case protobuf included as bazel dependency

### DIFF
--- a/python/BUILD.bazel
+++ b/python/BUILD.bazel
@@ -121,7 +121,7 @@ py_library(
     name = "python_srcs",
     srcs = glob(
         [
-            "google/protobuf/**/*.py",
+            "google/**/*.py",
         ],
         exclude = [
             "google/protobuf/internal/*_test.py",


haberman:
> pkg_resources magic required in all google/init.py files for it to work.

Since [PEP 420](https://peps.python.org/pep-0420/), it is no longer necessary to use `pkg_resources` magic to create a namespace package.  Instead we can just omit `__init__.py` to get namespace package behavior.

Could we instead:
1. Delete `google/__init__.py` entirely.
2. Set [legacy_create_init](https://docs.bazel.build/versions/main/be/python.html#py_binary.legacy_create_init) to `0`, so an `__init__.py` is not implicitly created.

Zaspire:
> Delete google/__init__.py entirely

this does work.

> Set [legacy_create_init](https://docs.bazel.build/versions/main/be/python.html#py_binary.legacy_create_init) to 0

legacy_create_init is not available for [py_library](https://docs.bazel.build/versions/main/be/python.html#py_library)

Only for py_binary and py_test.
This way, legacy_create_init will need to be set on each binary and test that transitively depends on protobuf

Kiyo5hi:
This should work:

```diff
--- a/BUILD
+++ b/BUILD
@@ -864,7 +864,7 @@ py_library(
     name = "python_srcs",
     srcs = glob(
         [
-            "python/google/protobuf/**/*.py",
+            "python/google/**/*.py",
         ],
         exclude = [
             "python/google/protobuf/internal/*_test.py",
```

Zaspire:
> stripping the path by one level should not cause any issue right?

It should not.

But since it just one file, ```__init__.py``` can be explicitly listed in srcs instead of modifying glob().

 WDYT?

Kiyo5hi:
> 

Sure, that should work. Have you tried that though?

Kiyo5hi:
> > stripping the path by one level should not cause any issue right?
> 
> It should not.
> 
> But since it just one file, `__init__.py` can be explicitly listed in srcs instead of modifying glob().
> 
> WDYT?

Sure, that should work. Have you tried that though?

Zaspire:
> Have you tried that though?

yes. Both works fine

> Sorry for the slow review on this. Is it still an issue?

yes.  I updated PR to latest HEAD

haberman (approved):

Zaspire:
Rebased PR on top of HEAD

deannagarcia:
I'm really sorry about the delays on this. I think you need a rebase once more. If you can do that and the tests pass I can get this in this week.

haberman:
I would still prefer to just delete `__init__.py` if that will work.

Issue #9809, opened 2022-04-20
Cmake incorrectly installs config files when passed absolute values


jonringer:
https://github.com/protocolbuffers/protobuf/blob/654c25352e831b991f668a2edd97ce6ea4e1f4c3/cmake/install.cmake#L145-L149
Is the offending  offending section.

`CMAKE_CURRENT_BINARY_DIR` will be some build directory (e.g. `/build/source/build/`)
`CMAKE_INSTALL_CMAKEDIR` inherits from `CMAKE_INSTALL_LIBDIR` (e.g. `/nix/store/x4bq4v508n8pxazh9137hcd64i40fvb0-protobuf-3.15.8/lib/cmake/protobuf`)

The resultant `/build/source/build//nix/store/x4bq4v508n8pxazh9137hcd64i40fvb0-protobuf-3.15.8/lib/cmake/protobuf` is not a path which can install files. 

**What version of protobuf and what language are you using?**
Version: 3.15, but still an issue in master
Language: C++

**What operating system (Linux, Windows, ...) and version?**
```
$ uname -a
Linux jon-desktop 5.4.188 #1-NixOS SMP Mon Mar 28 06:46:49 UTC 2022 x86_64 GNU/Linux
```

**What runtime / compiler are you using (e.g., python version or gcc version)**
```
$ gcc --version
gcc (GCC) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

**What did you do?**
```
cd cmake
mkdir build
cd build
cmake --build ..
```

**What did you expect to see**
It not fail to find directory

**What did you see instead?**
Failure around incorrect path

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).
```
       > CMake Error at cmake_install.cmake:1693 (file):
       >   file INSTALL cannot find
       >   "/build/source/build//nix/store/x4bq4v508n8pxazh9137hcd64i40fvb0-protobuf-3.15.8/lib/cmake/protobuf":
       >   No such file or directory.
```

**Anything else we should know about your project / environment**
Using Nix

PR #9822, opened 2022-04-20
Fix cmake install targets

jonringer:
Tried to capture most of the context in the commit messages, so I'll just paste them here:

> [Avoid exporting build tree targets to installation directory](https://github.com/protocolbuffers/protobuf/commit/da040de1eb646cfec2b41ba239b5f29bb808b041)
`export(TARGETS ...)` is meant to allow for build trees to be found
from other projects, however, writing the file to an installation
directory causes buildtree paths to be hardcoded in the installed
protobuf-targets.cmake

> [Install protobuf-targets.cmake from target, not build tree](https://github.com/protocolbuffers/protobuf/commit/66ca89b1c3745ecafb7649f19061481f6d4ed2b9)
This allows for installation prefixes to match the intended
installation directories, instead of being copied from
the build tree.

Fixes: #9809

TL;DR:
`export(TARGETS ...)` isn't meant to be used for installation, only local development. We should be using `install(EXPORT ...)` for providing relocatable target locations.

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -128,12 +128,12 @@ configure_file(${protobuf_SOURCE_DIR}/cmake/protobuf-options.cmake
 if (protobuf_BUILD_PROTOC_BINARIES)
   export(TARGETS libprotobuf-lite libprotobuf libprotoc protoc
     NAMESPACE protobuf::
-    FILE ${CMAKE_INSTALL_CMAKEDIR}/protobuf-targets.cmake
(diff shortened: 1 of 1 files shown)


jonringer:
I can't apply labels, but they should be just be `cmake`

acozzette (approved):

acozzette:
@jonringer Thanks!

jonringer:
@acozzette of course :)

Thanks for the speedy review

PR #9824, opened 2022-04-20
Start reducing our reliance on autotools

acozzette:
This commit updates a few of our CI scripts to build protoc using Bazel
instead of autotools. I also updated the setuptools and Rake build
configs to look for protoc under bazel-bin/ so that they will
preferentially use a Bazel-built protoc binary. The goal is to start
chipping away at our use of autotools in preparation for eventually
dropping autotools support.


kokoro/release/python/macos/config.sh
```diff
@@ -25,12 +25,13 @@ function pre_build {
...
-
-    CXXFLAGS="-std=c++14 -fPIC -g -O2" ./configure
-    make -j8
+    # Build protoc and protobuf libraries
+    use_bazel.sh 5.1.1
+    bazel build //:protoc
+    export PROTOC=$PWD/bazel-bin/protoc
+    mkdir src/.libs
+    ln -s $PWD/bazel-bin/libprotobuf.a src/.libs/libprotobuf.a
+    ln -s $PWD/bazel-bin/libprotobuf_lite.a src/.libs/libprotobuf-lite.a
```
dlj-NaN, line 34: On a side note, there is a flag that you can pass to setup.py to specify the libs directory:

https://github.com/pypa/setuptools/blob/main/setuptools/_distutils/command/build_ext.py#L76

However, that would have to be passed to (I think) the bdist_wheel command, not the build_py command.

Either way, though, this should work.

ruby/Rakefile
```diff
@@ -37,10 +37,12 @@ proto2_protos = %w[
   google/protobuf/compiler/plugin.proto
 ]
 
-if system('../src/protoc --version')
-  protoc_command = '../src/protoc' 
+if system('../bazel-bin/protoc --version')
```
dlj-NaN, line 40: This should be fine, but it might be more consistent overall to check the environment, maybe like:

if ENV['PROTOC'] != '':
  protoc_command = ENV['PROTOC']

(This seems fine, too, though.)

dlj-NaN (approved):

mkruskal-google (approved):

kokoro/release/python/macos/config.sh
acozzette: Thanks for the pointer! I think I will leave it how it is now though, only because these lines are likely going to be deleted very soon anyway when Python-UPB is released.

ruby/Rakefile
acozzette: Good idea, done.

Issue #9825, opened 2022-04-20
Java's memoized hash code can fail to be memoized.


shawjef3:
**What version of protobuf and what language are you using?**
Version: 3.20.0
Language: Java

**What operating system (Linux, Windows, ...) and version?**

MacOS 12.1

**What runtime / compiler are you using (e.g., python version or gcc version)**

gcc

**What did you do?**

The generated hash code for

```
message JustI {
  int32 i = 1;
}
```

is

```
    @java.lang.Override
    public int hashCode() {
      if (memoizedHashCode != 0) {
        return memoizedHashCode;
      }
      int hash = 41;
      hash = (19 * hash) + getDescriptor().hashCode();
      hash = (37 * hash) + I_FIELD_NUMBER;
      hash = (53 * hash) + getI();
      hash = (29 * hash) + unknownFields.hashCode();
      memoizedHashCode = hash;
      return hash;
    }
```

It's trivial to find that setting `i` to `308060953` makes `memoizedHashCode` 0. But this means that it will never be recognized as memoized. `memoizedHashCode` will be recomputed on every call of `hashCode()`.

This is a small example to demonstrate the principle. Any message could potentially have values that result in a hash code of 0.

**What did you expect to see**

`memoizedHashCode`, once set, should not be computed again on the second call.

**What did you see instead?**

It is recomputed on each call.

**Anything else we should know about your project / environment**

**Fix**

`AbstractMessageLite#memoizedHashCode` should be an `Integer` set to `null`. Then in the generated `hashCode()` method, instead of checking for `0`, check for `null`. Alternatively there could be a boolean field in `AbstractMessageLite` that is set to `true` when `memoizedHashCode` is set.

shawjef3:
Actually it'd have to be the `boolean` fix to make the code backwards compatible.

elharo:
neat discovery

elharo:
The current code still returns the correct result. It just does more work than it would have to otherwise. The question I need to ponder is what costs more in performance: the occasional recomputation of a hashCode or the extra code to deal with it.

shawjef3:
The condition is only checking the boolean, rather than the integer equality, so that is a simplification. It more than makes up for assigning the boolean IMO. A more important overhead in my mind is the extra `boolean` in memory for each protobuf object.

elharo:
Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
[...]

PR #9828, opened 2022-04-21

shawjef3:


perezd:
We'll rebase this automatically into our next release, no need to do this.

PR #9829, opened 2022-04-21

shawjef3:


perezd:
This I see #9830, #9829 and #9828 are basically identical, can you close all but one?

shawjef3:
If this is accepted as a bug, I'd think all of the branches should be fixed? Or at least certainly protobuf 4.

perezd:
This branch doesn't really exist. So lets close it.

shawjef3:
Yes, although I don't know which to keep. I figure all the different
versions should be fixed.

PR #9830, opened 2022-04-21

shawjef3:

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/AbstractMessage.java
+++ b/java/core/src/main/java/com/google/protobuf/AbstractMessage.java
@@ -158,12 +158,13 @@ public abstract class AbstractMessage
   @Override
   public int hashCode() {
     int hash = memoizedHashCode;
-    if (hash == 0) {
+    if (!memoizedHashCodeSet) {
       hash = 41;
       hash = (19 * hash) + getDescriptorForType().hashCode();
       hash = hashFields(hash, getAllFields());
       hash = (29 * hash) + getUnknownFields().hashCode();
       memoizedHashCode = hash;
+      memoizedHashCodeSet = true;
     }
     return hash;
   }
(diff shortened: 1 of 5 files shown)


elharo:
Another possible fix occurs to me. I'm not sure this is actually better, but let me rough it out. The problem is that any flag value like 0 or 989798782 we pick to mean "unmemoized" can also be a valid hash value. Here's my idea:

1. Pick a number such as 5003231 that is slightly less likely to show up by accident than 0 to indicate unmemoized.
2. If it does show up in the calculation, then add 1 and use that as the hash code.

This means that we will be deliberately combining two buckets. That is all the protos that now hash to 5003231 get mixed up with protos that hash to 5003232. However hash collisions are expected and handled. There are more protos than ints after all. 

This feels simpler than carrying around an extra non-final protected field, which is an API level change that technically would require a Tier 2 review. Thoughts?

shawjef3:
I suppose 5003231 is better than 0 as a guard value. @elharo your idea has merit.

I picked adding a boolean with the reasoning that past generated code will be compatible with the new library. They'll continue with the bug. Only newly generated code will require the new library.

For 4.0.x at least, changing the API should be okay since it's not released yet.

perezd:
Let's use this one as the PR of record.

elharo:
@shawjef3 I too am concerned about the extra weigh of an additional boolean field on every message. Memory size of proto messages is a large and growing concern for us. 

I think a modification of the hashCode method would still be forwards and backwards compatible. I'm not sure whether adding a non-private, non-final boolean field is or is not compatible. I'd need to think more about that.

shawjef3:
@elharo an incompatibility would arise if someone used a new protoc compiler version with an AbstractMessageLite that was older, so it didn't have boolean flag.

shaod2:
I was a little lost here.

From the discussion above, we have 2 options. (1) modification of hashCode (2) adding a boolean field.

I agreed (1) is forwards and backwards compatible, and from last comment, (2) will break if new gencode uses old runtime.

Then, isn't the only option to be non-breaking both ways is to modify the hashCode?

deannagarcia:
Closing due to no response, feel free to reopen with the comments addressed.

PR #9835, opened 2022-04-22
Add source distribution packaging rules

dlj-NaN:
This change adds `rules_pkg`-based targets that will produce source distribution archives, similar to `make dist`.

These rules produce nearly the same outputs as `make dist`. However, there are some differences and caveats:
1. The outputs do not contain vendored googletest sources.
2. You have to run `autogen.sh` before `blaze build pkg:all`. This produces several autotools-related files directly into the source tree.
3. The output .zip files do not have a directory prefix like `protobuf-3.20.1-rc-1` (this will be addressed after bazelbuild/rules_pkg#577; the tar files do have this prefix, though.)
4. One file is missing from the archives, which is produced during the `make` build: benchmarks/gogo/cpp_no_group/cpp_benchmark.cc
5. In several places, I have explicitly excluded some files that are not in the autotools distribution outputs. I think most of those files should probably be included, but for now, I'm aiming for parity with `make dist`. These are marked with comments, so it should be easy to clean them up later.


acozzette:
Awesome!

I wonder how hard it would be to bundle GoogleTest. Maybe that could be fairly straightforward if we could send the GoogleTest owners a PR adding a `pkg_files()` target that we can depend on.

BUILD
```diff
@@ -1445,3 +1446,139 @@ internal_gen_kt_protos(
...
+    srcs = glob([
+        "*.bzl",
+        "cmake/*.cmake",
+        "cmake/*.in",
+        "editors/*",
+
+        # Several of these files are generated by autogen.sh, so using
+        # glob() lets us ignore them if they are missing. (This is not good
+        # practice, though.)
+        "Makefile.*",  # Makefile.am is always present, though
```
deannagarcia, line 1466: Since Makefile.am is always present, can we put it in the exact matches below?

csharp/BUILD
```diff
@@ -0,0 +1,33 @@
...
+load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
+
+pkg_files(
+    name = "dist_files",
+    srcs = glob([
+        "keys/*",
+        "protos/*",
+        "src/**/*.cs*",  # .cs and .csproj
+    ]) + [
+        ".editorconfig",
```
deannagarcia, line 14: Is there a plan in place for how we will know whether to add new files to these packages?

pkg/README.md
```diff
@@ -0,0 +1,30 @@
+# Protobuf packaging
+
+This directory contains Bazel rules for building packaging and distribution
+artifacts.
+
+*Everything in this directory should be considered experimental.*
```
deannagarcia, line 6: Since we are using the java package rules to build, I'm not sure this is true.

pkg/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Source distribution packaging
+
+Protobuf releases include source distributions, sliced by target language (C++,
+Java, etc.). There are rules in this package to define those source archives.
+These depend upon `pkg_files` rules elsewhere in the repo to get the contents.
+
+The source distribution files should include the outputs from `autogen.sh`, but
+this isn't something we can reliably do from Bazel. To produce fully functioning
+source distributions, run `autogen.sh` before building the archives (this
```
deannagarcia, line 21: Is this a long-term necessity or a short-term workaround?

pkg/BUILD
```diff
@@ -83,7 +88,247 @@ pkg_zip(
...
     package_file_name = "protoc-{version}-{platform}.zip",
-    package_variables = ":protoc_pkg_naming",
+    package_variables = ":protobuf_pkg_naming",
+)
+
+################################################################################
+# Source distribution packaging
+################################################################################
+
+pkg_filegroup(
```
deannagarcia, line 98: Can we make a TODO to delete the other way we package and use this instead?

deannagarcia (approved):

BUILD
dlj-NaN: Done.

csharp/BUILD
dlj-NaN: > Is there a plan in place for how we will know whether to add new files to these packages?

I don't have a complete answer right now. I think there are probably two considerations, which are related but separate:

1. What do we use to produce the release artifacts?
2. What do we want to validate about them in general?

I think the answer for (1) ultimately is that we want to use these rules. However, we need to find a strategy for vendoring googletest before that can happen.

For (2), we have only partial coverage today (C++ distcheck). It seems reasonable to me to change the non-Bazel-driven builds to unzip the distribution artifacts and run inside, though, which I think is the ground truth we would want.

So, for example, this:
https://github.com/protocolbuffers/protobuf/blob/main/tests.sh#L229-L234
could change to something like:
```
bazel build :protoc pkg:dist_java_tar
BAZEL_BIN=$(bazel info bazel-bin)
export PROTOC=$BAZEL_BIN/protoc
tar -C $dir -axf $BAZEL_BIN/pkg/dist_java_tar.tar.gz
cd $dir && $MVN ...
```

pkg/README.md
dlj-NaN: Yeah, I'm more writing for users who might be tempted to depend on behavior. I'll tweak to say it's internal and subject to change.

pkg/README.md
dlj-NaN: Probably long term... at least, as long as we want the distributions to be buildable with autotools.

Having toyed in the past with trying to integrate autotools with Starlark rules, and given the complexity of our current autotools logic, I don't think it's currently worth trying to do as part of the build. There are a lot of ways to change that assessment, though.

pkg/BUILD
dlj-NaN: Done.

mkruskal-google (approved):

csharp/BUILD
deannagarcia: Ok that makes sense. My main concern here is that someone adds a new file that is needed across distributions and it doesn't get included. For now, we should just make sure this is properly documented.

dlj-NaN:
FYI, I wrote up issue #9842 to explain the overall change.

csharp/BUILD
dlj-NaN: Added comments to `Makefile.am`.

PR #9837, opened 2022-04-22
add Kokoro config for OpenJDK11

elharo:
@acozzette Please consider whether this is the correct Dockerfile for this job. For JDK17 (coming up soon) I will need to either modify this one or add a new one that includes OpenJDK 17. 

Also, I'm assuming we put this in Github first and then add the configs in piper.

### DIFF
--- a/kokoro/linux/java_jdk11/continuous.cfg
+++ b/kokoro/linux/java_jdk11/continuous.cfg
@@ -1,0 +1,11 @@
+# Config file for running tests in Kokoro
+
+# Location of the build script in repository
+build_file: "protobuf/kokoro/linux/java_linkage_monitor/build.sh"
+timeout_mins: 120
+
+action {
+  define_artifacts {
+    regex: "**/sponge_log.xml"
+  }
+}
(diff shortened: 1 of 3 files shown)


kokoro/linux/java_jdk11/continuous.cfg
```diff
@@ -0,0 +1,11 @@
+# Config file for running tests in Kokoro
+
+# Location of the build script in repository
+build_file: "protobuf/kokoro/linux/java_linkage_monitor/build.sh"
```
acozzette, line 4: I think this should be: `protobuf/kokoro/linux/java_jdk11/build.sh`

kokoro/linux/java_jdk11/presubmit.cfg
```diff
@@ -0,0 +1,12 @@
+# Config file for running Linkage Monitor in Kokoro
+# https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/linkage-monitor
+
+# Location of the build script in repository
```
acozzette, line 4: This path needs an update, too.

acozzette (approved):

kokoro/linux/java_jdk11/continuous.cfg
elharo: done

kokoro/linux/java_jdk11/presubmit.cfg
elharo: done

PR #9844, opened 2022-04-25
chore(typo): add line break in clear map method

nnnnoel:


nnnnoel:
move to https://github.com/protocolbuffers/protobuf-javascript/pull/2

PR #9856, opened 2022-04-26
make sure JDK17 tests fail as needed

elharo:
do not merge. Checking that this build breakage is detected


elharo:
and they don't, bleah

PR #9868, opened 2022-04-27
more decriptive error messages and exceptions

elharo:
This seems to be failing now. This change should help debug that.

### DIFF
--- a/conformance/ConformanceJava.java
+++ b/conformance/ConformanceJava.java
@@ -229,11 +229,11 @@ class ConformanceJava {
   }
 
   private Conformance.ConformanceResponse doTest(Conformance.ConformanceRequest request) {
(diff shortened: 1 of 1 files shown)


haberman:
I think you should be able to run the conformance tests locally to debug, using commands like this: https://github.com/protocolbuffers/protobuf/blob/main/tests.sh#L252-L265

Hopefully this will be Bazelified soon, so it's easier to reproduce locally...

elharo:
Thanks. That works. Looks like the problem is:

./conformance-test-runner --enforce_recommended --failure_list failure_list_java.txt --text_format_failure_list text_format_failure_list_java.txt ./conformance-java
Exception in thread "main" java.lang.IllegalArgumentException: Protobuf request has unexpected payload type: conformance.FailureSet

PR #9874, opened 2022-04-27
Remove the JavaScript implementation

acozzette:
This implementation is about to be split out into a separate Git repo.


haberman (approved):

zepatrik:
I assume the new repo is https://github.com/protocolbuffers/protobuf-javascript? Or where did this go? How can I use it now?

zepatrik:
Hm ok after two hours of research it seems this was just removed and after 3.20.1 there is no way to generate js. How is this not a breaking change??

mayankav:
why was it removed from this repo?

nocive:
what an absolute sh*t show

alexeagle:
@zepatrik yes, https://github.com/protocolbuffers/protobuf-javascript is the new repo.

szigetics:
A possible solution is shared here : https://github.com/protocolbuffers/protobuf-javascript/issues/127#issuecomment-1204202870

pietrodn:
There is an official first-party solution: if you `npm install grpc-tools`, then you simply use `grpc_tools_node_protoc` (which is inside `node_modules/.bin`) instead of `protoc`. See [documentation](https://www.npmjs.com/package/grpc-tools).

This way I managed to do without [protoc-gen-js](https://www.npmjs.com/package/protoc-gen-js).

PR #9878, opened 2022-04-28
Fail JDK 17 build when unit tests fail

elharo:
@acozzette

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -287,7 +287,7 @@ build_java_jdk11() {
 
 build_java_jdk17() {
   use_java jdk17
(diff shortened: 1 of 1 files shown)


haberman:
I'm not quite following, this appears to be disabling the conformance tests?

elharo:
The conformance tests were effectively disabled already. They fail during setup in a way that doesn't fail the build. This way a failure in the unit tests will be reported as an error and fail the presubmit check, which is not not currently happening.

acozzette (approved):

PR #9882, opened 2022-04-28
Add Starlark utilities to generate files with lists of files

dlj-NaN:
For now, these are simply generating files in CMake and Automake syntax. The generated files aren't actually used (yet) in build definitions.


acozzette (approved):

dlj-NaN:
OK, after a few more tweaks I can now use the output in our cmake and autotools builds for libprotobuf, libprotobuf-lite, and libprotoc.

I'm not 100% happy with how the extra_dist lists work, but that's more of a structural issue I can address separately.

I think this is about ready to land, but I'll leave it open for just a bit longer, in case there are any comments.

PR #9885, opened 2022-04-29
Use PROTOBUF_CONSTEXPR instead of constexpr for kRepHeaderSize

rvandermeulen:
Fixes "no matching function for call to 'CalculateReserveSize'" compiler errors with older clang versions.

### DIFF
--- a/src/google/protobuf/repeated_field.h
+++ b/src/google/protobuf/repeated_field.h
@@ -346,7 +346,7 @@ class RepeatedField final {
   int total_size_;
   // Pad the Rep after arena allow for power-of-two byte sizes when
   // sizeof(Element) > sizeof(Arena*). eg for 16-byte objects.
-  static constexpr size_t kRepHeaderSize =
+  static PROTOBUF_CONSTEXPR const size_t kRepHeaderSize =
       sizeof(Arena*) < sizeof(Element) ? sizeof(Element) : sizeof(Arena*);
   struct Rep {
     Arena* arena;
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -655,7 +655,7 @@
 #define PROTOBUF_CONSTEXPR constexpr
 #else
 #define PROTOBUF_CONSTINIT
-#define PROTOBUF_CONSTEXPR inline
+#define PROTOBUF_CONSTEXPR
 #endif
 
 // Some globals with an empty non-trivial destructor are annotated with


rvandermeulen:
For context, I've recently been looking at updating the protobuf import used by Firefox to version 3.20.1. While it works pretty much out of the box for the more recent toolchains we use for our official builds, it fails to compile in our "base toolchain" job which still uses clang5 (the oldest version we still support).

```
 INFO -  In file included from /builds/worker/checkouts/gecko/toolkit/components/protobuf/src/google/protobuf/repeated_field.cc:35:
ERROR -  /builds/worker/workspace/obj-build/dist/include/google/protobuf/repeated_field.h:933:14: error: no matching function for call to 'CalculateReserveSize'
 INFO -    new_size = internal::CalculateReserveSize<Element, kRepHeaderSize>(
 INFO -               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 INFO -  /builds/worker/checkouts/gecko/toolkit/components/protobuf/src/google/protobuf/repeated_field.cc:49:48: note: in instantiation of member function 'google::protobuf::RepeatedField<bool>::Reserve' requested here
 INFO -  template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<bool>;
 INFO -                                                 ^
 INFO -  /builds/worker/workspace/obj-build/dist/include/google/protobuf/repeated_field.h:905:12: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'kRepHeaderSize'
 INFO -  inline int CalculateReserveSize(int total_size, int new_size) {
 INFO -             ^
```

Changing kRepHeaderSize from constexpr to const appears to successfully resolve the build failures. If this isn't an acceptable option, I'm happy to try out any other proposed solutions.

rvandermeulen:
Thanks, I've been added to our CLA group and run a re-scan. Let me know if there's anything else you need!

src/google/protobuf/repeated_field.h
```diff
@@ -346,7 +346,7 @@ class RepeatedField final {
   int total_size_;
   // Pad the Rep after arena allow for power-of-two byte sizes when
   // sizeof(Element) > sizeof(Arena*). eg for 16-byte objects.
-  static constexpr size_t kRepHeaderSize =
+  static const size_t kRepHeaderSize =
```
acozzette, line 349: Actually could you try changing this to `PROTOBUF_CONSTEXPR const`? I believe that should make this constexpr only for recent enough compilers. (And then we want `const` added on there so that it's always const no matter what.)

acozzette (changes requested):

rvandermeulen:
I tried using the below and it still fails.
```
-  static constexpr size_t kRepHeaderSize =
+  static PROTOBUF_CONSTEXPR const size_t kRepHeaderSize =
```

If I misunderstood, let me know and I'll revise.

acozzette:
Hmm, that is surprising. Is `PROTOBUF_CONSTEXPR` evaluating to `constexpr` even with Clang 5? Or maybe adding `const` is still triggering the problem?

rvandermeulen:
It appears to be evaluating to `inline` for those specific builds.

rvandermeulen:
It was the same error as before:
https://treeherder.mozilla.org/logviewer?job_id=376400012&repo=try&lineNumber=24484

acozzette:
Could you try removing the `inline` keyword from the `PROTOBUF_CONSTEXPR` definition in `port_def.inc`? I don't think we actually need that `inline` and it should be fine to remove.

rvandermeulen:
That works!

acozzette (approved):

PR #9890, opened 2022-04-29
Be explicit with empty globs

FaBrand:
There are empty globs that prevent to use com_google_protobuf with
--incompatible_disallow_empty_glob

Introduced in https://github.com/bazelbuild/bazel/issues/8195


acozzette:
@FaBrand Thanks!

PR #9892, opened 2022-04-29
Rename all BUILD files to BUILD.bazel

dlj-NaN:


haberman:
Just preserving for the record that the intent is for `BUILD.bazel` to be the filename for any `BUILD` files that are not usable with Blaze.

If/when we can share BUILD files between Blaze and Bazel (eg. with suitable transforms), then we could rename the files to just `BUILD`.

haberman (approved):

BosskurrMH:
test

PR #9905, opened 2022-05-03
ObjC: Revise the minimal extension deps algorithm.

thomasvl:
When generating, it isn't uncommon to have generate >1 file at a time, and it is
likely that one file will include another. So cache the results as the
calculation is done so the work isn't repeated.

The previous pruning method didn't have any concept of tracking already done
work, this changes the algorithm to avoid the repeated work to make things more
minimal on the way up.

Some extremely deep proto graphs, this takes the generation time from around 15
min to under 45 seconds.

### DIFF
--- a/src/google/protobuf/compiler/objectivec/objectivec_file.cc
+++ b/src/google/protobuf/compiler/objectivec/objectivec_file.cc
@@ -117,46 +117,77 @@ bool FileContainsExtensions(const FileDescriptor* file) {
   return false;
 }
 
-// Helper for CollectMinimalFileDepsContainingExtensionsWorker that marks all
-// deps as visited and prunes them from the needed files list.
-void PruneFileAndDepsMarkingAsVisited(
-    const FileDescriptor* file,
-    std::vector<const FileDescriptor*>* files,
-    std::set<const FileDescriptor*>* files_visited) {
-  std::vector<const FileDescriptor*>::iterator iter =
-      std::find(files->begin(), files->end(), file);
-  if (iter != files->end()) {
-    files->erase(iter);
-  }
-  files_visited->insert(file);
+bool IsDirectDependency(const FileDescriptor* dep, const FileDescriptor* file) {
   for (int i = 0; i < file->dependency_count(); i++) {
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/objectivec/objectivec_file.cc
```diff
@@ -168,32 +199,23 @@ void CollectMinimalFileDepsContainingExtensionsWorker(
...
-  std::set<const FileDescriptor*> files_visited;
-  for (int i = 0; i < file->dependency_count(); i++) {
-    const FileDescriptor* dep = file->dependency(i);
-    CollectMinimalFileDepsContainingExtensionsWorker(dep, files,
-                                                     &files_visited);
-  }
+const std::vector<const FileDescriptor*>
+FileGenerator::CommonState::CollectMinimalFileDepsContainingExtensions(
+    const FileDescriptor* file) {
+  std::set<const FileDescriptor*> min_deps =
```
allevato, line 205: Are all the file descriptor names unique? Could you place the comparator on the sets, which would remove the need to sort the list afterwards? (You might even be able to return the set directly instead of copying it into a vector?)

thomasvl: They should be unique or the other includes would quickly get confused.

I debated tweaking the compare, but in the case where we're only generating the top most file(s), we end up only having to do the sort(s) at the end when collecting the final result and generating vs. having to sort all the intermediate values also, so this seemed better from a performance pov.

thomasvl:
I'm assuming the _Linux 32-bit_ and _MacOS PHP7.3_ failures are infrastructure flake issues.

@dlj-NaN @perezd can one you look?

dlj-NaN:
> I'm assuming the _Linux 32-bit_ and _MacOS PHP7.3_ failures are infrastructure flake issues.
> 
> @dlj-NaN @perezd can one you look? (edit: at this PR, not at the failures) smile

Yup, those look like infrastructural issues.

dlj-NaN (approved):

PR #9928, opened 2022-05-06
Added cmake abseil include guard

veblush:
This fix is needed to prevent gRPC from having duplicate definitions of abseil targets. Once protobuf started depending on Abseil (by #9793), both gRPC and Protobuf are trying to include abseil project, resulting in calling add_subdirectory toward abseil-cpp throwing a bunch of errors.

To avoid this error, an easy solution is to introduce an inclusion guard, we can think of a better solution in future but this is good enough as a short-term solution. The same fix will be added to gRPC as well.


veblush:
@mkruskal-google

mkruskal-google (approved):

jtattermusch (approved):

PR #9929, opened 2022-05-06
Add a test for bazel-built dist archives

dlj-NaN:
This adds a Kokoro build definition to run Bazel tests inside of a bazel-built distribution tree.

Notably, the distribution archives use the WORKSPACE file to retrieve deps. Sources such as googletest are not vendored in the archive.


deannagarcia (approved):
Can we also test that cmake builds work? And any other build system we support?

acozzette (approved):

dlj-NaN:
> Can we also test that cmake builds work? And any other build system we support?

I don't think the Bazel-built archive will work for cmake, since it doesn't include abseil or googletest sources. The closest existing build is cpp_distcheck, which validates automake.

deannagarcia:
Can we use these dist archives if they don't compile for cmake? I thought that was a requirement?

If they are going to need to support cmake but just don't yet can we leave a TODO to test a cmake build?

dlj-NaN:
> Can we use these dist archives if they don't compile for cmake? I thought that was a requirement?
> 
> If they are going to need to support cmake but just don't yet can we leave a TODO to test a cmake build?

That's a good idea, but I'm not sure where to record that. We wouldn't want to do the cmake build here, since that would just ~double the execution time... we probably want a separate build for that.

I think there is probably some refactoring that could be done as well, though, at which point there should be a natural cut point. I will send you a couple of follow-ups for that.

Issue #9446, opened 2022-01-27
Segmentation Fault caused on PHP


JimmyBanks:
**What version of protobuf and what language are you using?**
Protobuf 3.19.3 installed using pecl
PHP8.1.2 using PHP-FPM

**What operating system (Linux, Windows, ...) and version?**
Ubuntu 18.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
c version installed with pecl

**What did you do?**
1. Install:

```
Steps to reproduce the behavior:
sudo pecl install grpc
sudo pecl install protobuf
echo "extension=grpc.so" | sudo tee -a /etc/php/8.1/fpm/conf.d/grpc.ini
echo "extension=protobuf.so" | sudo tee -a /etc/php/8.1/fpm/conf.d/protobuf.ini
```

2. Release to production and higher traffic causes the segmentation fault. All appeared functional on staging with limited traffic

**What did you expect to see**
No segfault as this breaks the request

**What did you see instead?**
Seg faults listed in php8.1-fpm.log

```
[21-Jan-2022 16:17:11] WARNING: [pool www] child 3379 exited on signal 11 (SIGSEGV - core dumped) after 430.745632 seconds from start
[21-Jan-2022 16:17:11] NOTICE: [pool www] child 3603 started
[21-Jan-2022 16:17:17] WARNING: [pool www] child 3120 exited on signal 11 (SIGSEGV - core dumped) after 706.382257 seconds from start
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**
Stack: Ubuntu + Nginx + PHP-FPM 

Backtraces:

```
Core was generated by `php-fpm: pool www                                                            '.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f72134fa743 in Arena_Get (val=val@entry=0x7f7226552c18)
    at /tmp/pear/temp/protobuf/arena.c:72
72      /tmp/pear/temp/protobuf/arena.c: No such file or directory.
[Current thread is 1 (Thread 0x7f722c618500 (LWP 1938))]
(gdb) bt
#0  0x00007f72134fa743 in Arena_Get (val=val@entry=0x7f7226552c18)
    at /tmp/pear/temp/protobuf/arena.c:72
#1  0x00007f7213506064 in zim_Message___construct (
    execute_data=0x7f72264148c0, return_value=<optimized out>)
[...]

PR #9930, opened 2022-05-06
[do not merge] Remove the special-case code that suppresses pure-PHP properties.

haberman:
There is a chance this may fix some SEGVs seen on PHP 8.1 when
OPCache is enabled:
  https://github.com/protocolbuffers/protobuf/issues/9446#issuecomment-1119517310

However this also will increase memory usage of PHP protos, and
will likely introduce CPU overhead also. Therefore we probably don't
want to release the code in this state.

If this indeed fixes the SEGVs being seen, we should redesign the generated
code to no longer create properties unless pure-PHP is being used.

cc: @bshaffer

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -51,28 +51,21 @@
 // -----------------------------------------------------------------------------
 
 typedef struct {
-  zend_object std;
   zval arena;
   const Descriptor* desc;
   upb_Message *msg;
+  zend_object std;
 } Message;
 
 zend_class_entry *message_ce;
 static zend_object_handlers message_object_handlers;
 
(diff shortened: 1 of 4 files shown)


php/ext/google/protobuf/message.c
```diff
@@ -51,28 +51,21 @@
 // -----------------------------------------------------------------------------
 
 typedef struct {
-  zend_object std;
   zval arena;
   const Descriptor* desc;
   upb_Message *msg;
+  zend_object std;
```
fowles, line 57: please add a comment explaining that this must come last because PHP uses a tail allocation to do more fun stuff

fowles (approved):

PR #9937, opened 2022-05-09
Update version for 21.0-rc-1 release

deannagarcia:


protoc-artifacts/pom.xml
```diff
@@ -8,7 +8,7 @@
   </parent>
   <groupId>com.google.protobuf</groupId>
   <artifactId>protoc</artifactId>
-  <version>3.20.1</version>
+  <version>21.0-rc-1</version>
```
acozzette, line 11: I believe this version should actually match the Java version according to the major version proposal. @haberman Can you confirm if that is right?

haberman: Yes, for protoc packages in a language-specific package manager, I believe we did decide to make the package version match the runtime version.

PR #9939, opened 2022-05-09
[Kokoro] Add dist check build for linux+cmake

dlj-NaN:
This factors out some cmake build logic as well as the file test from cpp_distcheck.

### DIFF
--- a/kokoro/common/check_missing_dist_files.sh
+++ b/kokoro/common/check_missing_dist_files.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+#
+# Script to compare a distribution archive for expected files based on git.
+#
+# Usage:
+#   check_missing_dist_files.sh path/to/dist_archive.tar.gz
+
+set -eux
+set -o pipefail
+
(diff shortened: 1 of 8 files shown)


deannagarcia (approved):

kokoro/common/check_missing_dist_files.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Check for missing files.
+MISSING_FILES=( $(cd ${_workdir} && comm -13 archive.lst expected.lst) )
+if (( ${#MISSING_FILES[@]} == 0 )); then
+  exit 0
+fi
+
+(
+  set +x
+  echo -e "\n\nMissing files from archive:"
```
haberman, line 49: If this error fires, what should a user do to fix it?

haberman: (Can the message include the action that should be taken?)

dlj-NaN: Good idea, done.

PR #9944, opened 2022-05-10
Update CMakeLists.txt to handle new version scheme

acozzette:
Our version regex expected something like 1.2.3-rc-1, but we are
changing the version scheme so that protoc will have a version more like
1.2-rc-1. This commit fixes the regex to handle the new scheme.


deannagarcia (approved):

dlj-NaN:
This looks like it would make the C++ major version 21... I don't think that's intended, is it?

acozzette:
Good point, maybe I should keep the regex roughly the same but just read it from version.json so that we're getting the C++ protobuf version instead of the protoc version.

PR #9946, opened 2022-05-11
Add endian.h to includes list

boscosiu:
Client code is breaking due to missing include.  (endian.h was added in 9e09343 but didn't make it into this list, so it wasn't getting installed)


deannagarcia:
Thanks for this fix, merging now!

PR #9951, opened 2022-05-11
Change how proto_gen rule exposes sources

dlj-NaN:
* Switch to new provider syntax instead of structs.
* Expose files via `DefaultInfo`, and use that for `py_proto_library`.
* Several changes from buildifier.


haberman (approved):

dlj-NaN:
* Switch to new provider syntax instead of structs.
* Expose files via `DefaultInfo`, and use that for `py_proto_library`.
* Several changes from buildifier.

PR #9952, opened 2022-05-12
[Bazel] Move Python rules to //python

dlj-NaN:
* This PR moves Python rules to //python, but leaves aliases for publicly-visible targets.
* There is some internal cleanup for how we copy .proto sources so they are usable from Python.
* python/BUILD.bazel is now more in line with Bazel's BUILD style guide.

### DIFF
--- a/python/internal.bzl
+++ b/python/internal.bzl
@@ -1,20 +1,5 @@
 # Internal helpers for building the Python protobuf runtime.
 
-def _RelativeOutputPath(path, include, dest = ""):
-    if include == None:
-        return path
-
-    if not path.startswith(include):
-        fail("Include path %s isn't part of the path %s." % (include, path))
-
-    if include and include[-1] != "/":
-        include = include + "/"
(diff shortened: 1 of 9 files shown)


python/internal.bzl
```diff
@@ -0,0 +1,81 @@
...
+            mnemonic = "InternalCopyFile",
+            progress_message = "Copying files",
+            use_default_shell_env = True,
+        )
+
+    return [
+        DefaultInfo(files = depset([dest for src, dest in src_dests])),
+    ]
+
+internal_copy_files_impl = rule(
```
haberman, line 63: Could you add some more comments for the semantics of this rule and the attributes?

It's not immediately clear to me what this rule does at a high level, or why it needs to make a batch file.

haberman (approved):

python/internal.bzl
dlj-NaN: Done.

PR #9977, opened 2022-05-16
Add new C++ 20 keywords

keith:
Discovered in https://github.com/envoyproxy/envoy/issues/21010


keith:
> unfortunately, this is a breaking change for folks compiling with older C++ standards. As such, we will have to be fairly careful in how we do this.

makes sense, is there a normal process that's followed for this or has it been too infrequent?

fowles:
We are working on this internally and would like to put this into the next release, but we have to fix things inside google as we do it, so we are not going to move on this pull request.  That said, we do intend to fix this.

PR #9979, opened 2022-05-16
Fix protoc_release target

deannagarcia:
Edit the protoc_release target to package the proto files themselves instead of the proto descriptors.

Tested manually by building the artifact and inspecting the zip file. All contents were as expected.

### DIFF
--- a/pkg/BUILD.bazel
+++ b/pkg/BUILD.bazel
@@ -17,25 +17,17 @@ package_naming(
 pkg_files(
     name = "wkt_protos_files",
     srcs = [
-        "//:any_proto",
-        "//:api_proto",
-        "//:duration_proto",
-        "//:empty_proto",
-        "//:field_mask_proto",
-        "//:source_context_proto",
-        "//:struct_proto",
-        "//:timestamp_proto",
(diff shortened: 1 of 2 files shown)


perezd (approved):
LGTM!

pkg/BUILD.bazel
```diff
@@ -17,33 +17,25 @@ package_naming(
...
-    ],
-    prefix = "include/google/protobuf",
-    visibility = ["//visibility:private"],
-)
-
-pkg_files(
-    name = "descriptor_protos_files",
-    srcs = [
-        "//:descriptor_proto",
+        "//:src/google/protobuf/any.proto",
```
dlj-NaN, line 20: You could also use //:well_known_type_protos and //:built_in_runtime_protos instead of listing each file.

deannagarcia: I can't use `//:built_in_runtime_protos` because the prefixes of the two files in the group need to be different. 

Similarly, I don't really want to use `//:well_known_type_protos` because I would then have to add in `descriptor.proto` which I worry will be less intuitive than just listing them out like this.

PR #9980, opened 2022-05-17
[Bazel/C++] Factor out stubs, io, and testing libraries.

dlj-NaN:
This change creates packages under `src/google/protobuf/{io,stubs,testing}` and moves build definitions there. Future changes will handle `.../util` and `.../compiler`, and finally `src/google/protobuf`.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -656,6 +656,7 @@ cc_binary(
         "src/google/protobuf/compiler/mock_code_generator.cc",
         "src/google/protobuf/compiler/test_plugin.cc",
     ],
+    copts = COPTS,
     deps = [
         ":protobuf",
         ":protoc_lib",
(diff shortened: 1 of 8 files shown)


BUILD.bazel
```diff
@@ -20,12 +20,36 @@ exports_files(["LICENSE"])
...
+    ],
+    visibility = ["//:__subpackages__"],
+)
+
+cc_library(
+    name = "arena",
+    srcs = ["src/google/protobuf/arena.cc"],
+    hdrs = [
+        "src/google/protobuf/arena.h",
+        "src/google/protobuf/arena_impl.h",
```
fowles, line 39: if _impl is a private header should it go in the srcs dir?

fowles (approved):

BUILD.bazel
dlj-NaN: Done.

haberman (approved):

PR #9988, opened 2022-05-18
[Bazel/C++] Move core C++ implementation to //src/google/protobuf package.

dlj-NaN:
This is somewhat of a rough cut, since it doesn't split apart the lite and full targets, or unit tests.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -6,252 +6,11 @@ load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("@rules_proto//proto:defs.bzl", "proto_lang_toolchain", "proto_library")
 load("@rules_java//java:defs.bzl", "java_binary", "java_lite_proto_library", "java_proto_library")
 load("//build_defs:cpp_opts.bzl", "COPTS", "LINK_OPTS")
-load(
-    ":protobuf.bzl",
-    "adapt_proto_library",
-    "py_proto_library",
-)
+load(":protobuf.bzl", "py_proto_library")
 
 licenses(["notice"])
 
 exports_files(["LICENSE"])
-
-################################################################################
-# Protobuf Runtime Library
-################################################################################
-
-cc_library(
-    name = "port_def",
-    hdrs = ["src/google/protobuf/port.h"],
-    includes = ["src"],
-    textual_hdrs = [
-        "src/google/protobuf/port_def.inc",
-        "src/google/protobuf/port_undef.inc",
-    ],
-    visibility = ["//:__subpackages__"],
-)
-
-cc_library(
-    name = "arena",
-    srcs = [
-        "src/google/protobuf/arena.cc",
-        "src/google/protobuf/arena_impl.h",
-        "src/google/protobuf/arenaz_sampler.h",
-    ],
-    hdrs = ["src/google/protobuf/arena.h"],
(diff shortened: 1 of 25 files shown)


BUILD.bazel
```diff
@@ -267,96 +26,124 @@ cc_library(
...
-    srcs = ["src/google/protobuf/wrappers.proto"],
-    strip_import_prefix = "src",
+    actual = "//src/google/protobuf:wrappers_proto",
     visibility = ["//visibility:public"],
 )
 
-# Built-in runtime types
+# Source files: these are aliases to a filegroup, not a `proto_library`.
+#
+# (This is _probably_ not what you want.)
```
haberman, line 91: Is this in place just for compatibility? Do we have specific users that we know of? Can we remove this once we burn them down?

BUILD.bazel
```diff
@@ -267,96 +26,124 @@ cc_library(
...
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "compiler_plugin_proto",
+    actual = "//src/google/protobuf/compiler:plugin_proto",  # proto_library
+    visibility = ["//visibility:public"],
+)
+
+# adapt_proto_library(
```
haberman, line 125: Commented out code?

haberman (approved):

fowles (approved):
LGTM modulo the commented out bit

BUILD.bazel
dlj-NaN: For the old name `lite_well_known_protos`, it's compatibility and that name can probably go away soon-ish. But I think we will probably need to provide access to the sources for the foreseeable future (so `well_known_type_protos` probably ought to stick around), and I sort of think that the root package is the right place to put it. But that can be a future change.

BUILD.bazel
dlj-NaN: Oops, thanks. Removed it.

johanbrandhorst:
I think this PR broke existing Bazel users depending on the well known types in this repository. The error I'm seeing in my repository is
```
no such target '@com_google_protobuf//:well_known_protos': target 'well_known_protos' not declared in package '' defined by /home/vscode/.cache/_grpc_gateway_bazel/external/com_google_protobuf/BUILD.bazel
```
. I fixed it by changing the reference to `@com_google_protobuf//:well_known_type_protos`. It would be nice if this change was added as a note to the release notes, so users can triage these issues themselves.

dlj-NaN:
> I think this PR broke existing Bazel users depending on the well known types in this repository. The error I'm seeing in my repository is
> 
> ```
> no such target '@com_google_protobuf//:well_known_protos': target 'well_known_protos' not declared in package '' defined by /home/vscode/.cache/_grpc_gateway_bazel/external/com_google_protobuf/BUILD.bazel
> ```
> 
> . I fixed it by changing the reference to `@com_google_protobuf//:well_known_type_protos`. It would be nice if this change was added as a note to the release notes, so users can triage these issues themselves.

Sorry about that... the "well_known_protos" target went away in #9915, to distinguish between the Well-Known Types and the built-in runtime protos. Your fix sounds correct, if the WKTs are what you want. Depending on how you use the WKTs, though, you might be able to use the finer-grained targets, too.

I'll go ahead and add this target back with a deprecation warning (see #10061).

PR #9992, opened 2022-05-18
adds setTotalBytesLimit() method

### LINKED ISSUE #9990: PHP version of protobuf has hardcoded total_byte_limit equal to 32MB
Protobuf v3.x, PHP all versions, all systems and evironments.

There is a hardcoded limit which prevents reading ONNX files which are typically very large (hundreds of MBs):
https://github.com/protocolbuffers/protobuf/blob/07303d6f54e4768700925119a003493aacdb70ff/php/src/Google/Protobuf/Internal/CodedInputStream.php#L53

Raeching the limit starts chain of failures that finally throws Exception with misleading name and description:
> Fatal error: Uncaught Google\Protobuf\Internal\GPBDecodeException: Error occurred during parsing: Unexpected EOF inside bytes field.

For other languages this limit is not only much, much higher (like, 2GB for C++ instead of 32MB for PHP):
https://github.com/protocolbuffers/protobuf/blob/07303d6f54e4768700925119a003493aacdb70ff/src/google/protobuf/io/coded_stream.h#L623

...but also there are methods to change it programatically.:
https://github.com/protocolbuffers/protobuf/blob/07303d6f54e4768700925119a003493aacdb70ff/src/google/protobuf/io/coded_stream.h#L367-L384

Either add a method to raise the limit or increase the hardcoded value. And make some better exception message for this.

gakowalski:
fixes #9990 

Adds a method to raise total read bytes limit. Similar methods are present in code for other languages, this one was modelled after a C++ version. Without this method it is imporssible to parse large files like ONNX models as there is no way to change hardcoded read limits.

Proposed usage, instead of:
```php
$model = new \Onnx\ModelProto;
$model->mergeFromStream($data);
```

Do this (raises limit to 64MB):
```php
$model = new \Onnx\ModelProto;
$input_stream = new \Google\Protobuf\Internal\CodedInputStream($data);
$input_stream->setTotalBytesLimit(64 << 20);
$model->parseFromStream($input_stream);
```

### DIFF
--- a/php/src/Google/Protobuf/Internal/CodedInputStream.php
+++ b/php/src/Google/Protobuf/Internal/CodedInputStream.php
@@ -89,6 +89,29 @@ class CodedInputStream
     {
         return substr($this->buffer, $start, $end - $start);
     }
+    
+    /**
+    * Total Bytes Limit -----------------------------------------------
+    * To prevent malicious users from sending excessively large messages
+    * and causing memory exhaustion, CodedInputStream imposes a hard limit on
+    * the total number of bytes it will read.
+    *
+    * Sets the maximum number of bytes that this CodedInputStream will read
+    * before refusing to continue.  To prevent servers from allocating enormous
+    * amounts of memory to hold parsed messages, the maximum message length
+    * should be limited to the shortest length that will not harm usability.
+    *
+    * Note: setting a limit less than the current read position is interpreted
+    * as a limit on the current position.
+    *
+    * This is unrelated to pushLimit() / popLimit().
+    *
+    * @param $total_bytes_limit
+    */
+    public function setTotalBytesLimit($total_bytes_limit) {
+        $this->total_bytes_limit = max($this->current(), $total_bytes_limit);
+        $this->recomputeBufferLimits();
+    }
 
     private function recomputeBufferLimits()
     {


haberman:
I think we could just remove the limit from `CodedInputStream`. In C++ it makes more sense since it supports actual streaming, but in PHP we only support parsing from a string, and the user can just check the limit themself:

```php
$model = new \Onnx\ModelProto;
if (strlen($data) > DEFAULT_TOTAL_BYTES_LIMIT) {
  // Too big.
}
$model->mergeFromStream($data);
```

I'm pretty sure the C extension is not currently enforcing any kind of limit, so removing the limit from PHP would give the two implementations parity.

haberman:
> This solution would work for me, but from other people's perspective it would deepen differences between implementations.

Which implementations do you mean?

C++ has a bytes limit, but I believe it only applies when the user explicitly constructs a `CodedInputStream`.

The pure-PHP implementation uses a `CodedInputStream` internally, but this is not exposed to the user, and no CodedInputStream APIs are public in PHP. So the same does not apply here.

I don't think the PHP C extension has a bytes limit of any kind. So we already have a difference between pure-PHP and PHP/C. Removing the bytes limit from pure-PHP would make it more similar to PHP/C.

deannagarcia:
Closing this due to no response, feel free to reopen.

PR #10000, opened 2022-05-19
Cherry-pick PR #9981 into 21.x branch

jskeet:
.NET 7 (currently in preview) introduces a warning (CS8981) which is triggered by the use of namespace aliases within generated code.
We don't expect the aliases chosen to ever cause conflicts with C# keywords, so the generated code now suppresses this warning.


jskeet:
(Note: we *could* add this into the release-notes; the original PR has "relnotes: no" but we could do so if we wanted to. Let me know if you'd like me to write up a description.)

deannagarcia (approved):

deannagarcia:
If you add a description to this PR I can go ahead and add that to the release notes!

jskeet:
@deannagarcia: Done - hope the description makes sense!

PR #10008, opened 2022-05-20
[Kokoro/Bazel] Send Bazel build results to BES.

dlj-NaN:


kokoro/common/bazel_wrapper.sh
```diff
@@ -0,0 +1,67 @@
+# Wrapper for invoking bazel on Kokoro.
+#
+# This file should be `source`d. It provides a function `bazel` which adds flags
```
haberman, line 3: This seems slightly too clever -- a reader will not easily be able to see that a given call site is invoking a wrapper function instead of running bazel directly.

Could we make this explicit by changing call sites to directly invoke:

```
kokoro/common/bazel_wrapper build ...
```
```

dlj-NaN: Done.

This is a tiny bit complicated because we don't package Kokoro scripts, so I kept bazel_wrapper.sh source'able (in which case it provides the `bazel_wrapper` function).

Let me know whether you prefer the direct call tracking the repo root, or if you think the source+function approach would be better for the distcheck builds.

haberman (approved):

PR #10010, opened 2022-05-20
[Bazel] Make the protobuf workspace (nearly) wildcard-buildable

dlj-NaN:
- Add back benchmarks/datasets/BUILD.bazel to dist archive
- Change dist/build_systems.bzl to be compatible with Bazel 4.0.0.

After this, most builds can use `//...`, with one exception: non-Mac environments must also exclude objectivec:
`bazel build -- //... -//objectivec/...`


dlj-NaN:
Note this has some changes from #10009.

haberman (approved):

PR #10014, opened 2022-05-20
Adding a DEPENDENCIES option to the protobuf_generate cmake function

cjtallman:
Adds a way to pipe additional dependencies into the custom CMake code generation function. This was done to make it easier to run the generation using a custom protoc plugin within the CMake project. This allowed me to work on the plugin and cause the downstream dependencies to be regenerated.

If there's a better way to do this, please advise.


acozzette (approved):

acozzette:
@cjtallman Thanks!

PR #10019, opened 2022-05-20
Fix podspecs

deannagarcia:
Change cocoapod specs to use the version as the tag number. We will be tagging each release with the protoc version number, the c++ version number and the cocoapod version number.


acozzette (approved):
I think we will eventually need for both the C++ and Objective-C versions to be tagged since they might not always be the same version.

deannagarcia:
> I think we will eventually need for both the C++ and Objective-C versions to be tagged since they might not always be the same version.

Good point! Will add that to the description and my TODO list.

PR #10020, opened 2022-05-21
Enable RBE using bazel's rbe_default toolchain.

dlj-NaN:


googleberg:
@dlj-NaN Should we do something with this or close it out?

PR #10031, opened 2022-05-24
[Bazel] Fix headers for protobuf and protobuf_lite

dlj-NaN:
This change makes the headers into actual lists of files, rather than globs.


dlj-NaN:
Unfortunately, in order for the `strip_include_prefix` work properly, I had to move some `.inc` files from `textual_hdrs` to `hdrs`. These were previously being provided through the globs, so even though this doesn't quite model the textual split anymore, I think the overall state is better. I'm happy to tweak if you'd like, as long as these still work.

PR #10036, opened 2022-05-24
Revert "fix: reserve "ReadOnly" keyword for PHP 8.1 and add compatibi…lity (#9633)"

haberman:
This reverts commit eb27c201f121b02c990c3665edce5171a8c70192.

We are rolling back for now because the change breaks if you are using old generated code with the new runtime. We're rolling back for 3.21.0, but will hopefully have a fix to release in 3.21.1.

cc: @bshaffer


bshaffer:
LGTM

deannagarcia (approved):

PR #9344, opened 2021-12-25
Various fixups to warnings flags

benesch:
Two commits in here that fix up some glitches in how libprotobuf/protoc handle warnings, described below.

----

**Don't drop parser warnings on the floor**

Fix #9343.

----

**Convert "missing syntax" warning to an actual warning**

For some reason this warning was emitted as a log message rather than a
structured warning. Convert it to use the AddWarning API so that it gets
emitted with a file and line number by protoc, and is visible via the
error collection interface during programmatic use.

PR #10037, opened 2022-05-24
Revert "Various fixups to warnings flags (#9344)"

acozzette:
This reverts commit 448d421250569811a3e656cd4c87fe42e857ed6f.

Unfortunately we have to revert this because we're finding that it
introduces too much new build log spam for existing proto files that are
out of compliance with the warnings. We might be able to roll it forward
again if we can figure out a way to do so without so many new log
messages.


zhangskz:
I think we don't actually need to revert the whole thing, mostly just the AddWarning() in importer.cc right? I think we can revert the whole thing + re-add the other parts later though

zhangskz (approved):

PR #10038, opened 2022-05-24
feat: [PHP] add previous classname to descriptor pool

bshaffer:
addresses https://github.com/protocolbuffers/protobuf/pull/10036 by adding previously unreserved names to the descriptor pool. This preserves forwards compatibility.

cc @dwsupplee @haberman

### DIFF
--- a/php/src/Google/Protobuf/Internal/Descriptor.php
+++ b/php/src/Google/Protobuf/Internal/Descriptor.php
@@ -45,6 +45,7 @@ class Descriptor
     private $enum_type = [];
     private $klass;
     private $legacy_klass;
+    private $previous_klass;
     private $options;
     private $oneof_decl = [];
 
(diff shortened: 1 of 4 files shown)


php/src/Google/Protobuf/Internal/Descriptor.php
```diff
@@ -162,6 +163,16 @@ public function getLegacyClass()
         return $this->legacy_klass;
     }
 
+    public function setPreviousClass($klass)
```
dwsupplee, line 166: I lean a bit towards this being too verbose, but what about `setPreviouslyReservedClass`? If that's too much, maybe we can just add a note to the docblock about the intent, otherwise the intent isn't too clear.

bshaffer: Done in https://github.com/protocolbuffers/protobuf/pull/10041

bshaffer:
Closing this in favor of https://github.com/protocolbuffers/protobuf/pull/10041 which is submitted to the `21.x` branch

PR #10039, opened 2022-05-25
[Bazel/C++] Split apart protobuf_test rule

dlj-NaN:
This splits the `protobuf_test` target into several other test targets, and adds `lite_unittest` and `lite_arena_unittest`, which were missing previously.

I have left the `protobuf_test` target in place but empty for now.


acozzette:
Nice! The only thing I'm wondering is whether some of the CI jobs might need to be updated? I'm a bit confused because the Bazel run looks like it has already been updated, but on the other hand the log output makes it look like it's not actually running any tests.

acozzette (approved):

dlj-NaN:
> Nice! The only thing I'm wondering is whether some of the CI jobs might need to be updated? I'm a bit confused because the Bazel run looks like it has already been updated, but on the other hand the log output makes it look like it's not actually running any tests.

Sent you a couple of google-internal links offline... there are two separate bazel invocations in some of the Kokoro tests, and it's not totally obvious how to find both of them. But we did run all the tests for both kokoro/linux/bazel and kokoro/linux/bazel_distcheck.

PR #10042, opened 2022-05-25
[cleanup] MSVC Warnings - Fix Type Conversions (4244 and 4267), Enable Warnings as Errors

tusharb86:
This change fixes the following MSVC warnings:
- [C4244 – 'conversion' conversion from 'type1' to 'type2', possible loss of data](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244?view=msvc-170)
- [C4267 – 'var' : conversion from 'size_t' to 'type', possible loss of data](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267?view=msvc-170)

To prevent further instances from making it into the codebase, the change also adds the [`/WX` flag](https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170) to treat all compiler warnings as errors.

This change also disables the following MSVC warning (to be addressed in a follow-up PR):
- [C4018 - 'token' : signed/unsigned mismatch](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4018?view=msvc-170)

For more context on these changes, please see #9263 and the discussion there.

### DIFF
--- a/cmake/tests.cmake
+++ b/cmake/tests.cmake
@@ -6,6 +6,10 @@ option(protobuf_ABSOLUTE_TEST_PLUGIN_PATH
   "Using absolute test_plugin path in tests" ON)
 mark_as_advanced(protobuf_ABSOLUTE_TEST_PLUGIN_PATH)
 
+# The compile_proto_file macro places files in the source tree, so this option
+# must be turned off.
+set(CMAKE_DISABLE_SOURCE_CHANGES OFF)
+
 if (protobuf_USE_EXTERNAL_GTEST)
   find_package(GTest REQUIRED)
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,10 +257,9 @@ if (MSVC)
   add_definitions(/utf-8)
   # MSVC warning suppressions
   add_definitions(
+    /WX # Warnings as errors
     /wd4065 # switch statement contains 'default' but no 'case' labels
-    /wd4244 # 'conversion' conversion from 'type1' to 'type2', possible loss of data
     /wd4251 # 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
-    /wd4267 # 'var' : conversion from 'size_t' to 'type', possible loss of data
     /wd4305 # 'identifier' : truncation from 'type1' to 'type2'
     /wd4307 # 'operator' : integral constant overflow
     /wd4309 # 'conversion' : truncation of constant value
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,7 @@ if (MSVC)
   # MSVC warning suppressions
   add_definitions(
     /WX # Warnings as errors
+    /wd4018 # 'token' : signed/unsigned mismatch
     /wd4065 # switch statement contains 'default' but no 'case' labels
     /wd4251 # 'identifier' : class 'type' needs to have dll-interface to be used by clients of class 'type2'
     /wd4305 # 'identifier' : truncation from 'type1' to 'type2'
--- a/src/google/protobuf/arena_test_util.h
+++ b/src/google/protobuf/arena_test_util.h
@@ -53,7 +53,7 @@ void TestParseCorruptedString(const T& message) {
     message.SerializePartialToCodedStream(&out);
   }
   const int kMaxIters = 900;
-  const int stride = s.size() <= kMaxIters ? 1 : s.size() / kMaxIters;
+  const int stride = s.size() <= kMaxIters ? 1 : static_cast<int>(s.size() / kMaxIters);
   const int start = stride == 1 || use_arena ? 0 : (stride + 1) / 2;
   for (int i = start; i < s.size(); i += stride) {
     for (int c = 1 + (i % 17); c < 256; c += 2 * c + (i & 3)) {
--- a/src/google/protobuf/compiler/annotation_test_util.cc
+++ b/src/google/protobuf/compiler/annotation_test_util.cc
@@ -95,7 +95,7 @@ bool RunProtoCompiler(const std::string& filename,
 bool DecodeMetadata(const std::string& path, GeneratedCodeInfo* info) {
   std::string data;
   GOOGLE_CHECK_OK(File::GetContents(path, &data, true));
-  io::ArrayInputStream input(data.data(), data.size());
+  io::ArrayInputStream input(data.data(), static_cast<int>(data.size()));
   return info->ParseFromZeroCopyStream(&input);
 }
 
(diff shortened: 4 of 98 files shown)


tusharb86:
@acozzette From our chat in the related PR (#9263), I had found out about the helpers (`FromIntSize` and `ToIntSize`) when I was almost done with these changes. I'm happy to make the changes to using them, but it would be great to get more guidance as to when they should/shouldn't be used. Further, do we want to introduce a more generic version (e.g. `ToSmallerType<uint64_t, int>`) to prevent data loss/overflow when going from any larger type to any smaller type? (I'm open to naming suggestions here 🙂)

I had something like this in mind:
```c++
// For cases where an integer is being downcast to a smaller type (e.g. uint64_t
// to uint32_t).  We GOOGLE_DCHECK() that the conversion will fit within the
// target type; if this is false then we are losing information.
template<typename From, typename To>
inline To ToSmallerType(From from) {
  GOOGLE_DCHECK_LE(from, static_cast<From>(std::numeric_limits<To>::max()));
  return static_cast<To>(from);
}
```

My plan would be to use this helper (along with the existing ones) in places that I have modified in this change. For places that already have casts in place, I would consider those out of scope. Let me know what you think - I'm happy to chat about this further over a call, if needed. Thanks!

acozzette:
@tusharb86 Thanks for the change! We need to make sure the CLA check bot is happy before I can review or merge this, though. I see you're at Microsoft and we have a corporate CLA on file, so you likely need to just ask someone at Microsoft to add you to the necessary group. About the helper function, I need to think about this some more and ask the rest of the protobuf team.

tusharb86:
Thanks @acozzette! I've signed the CLA and the bot is now happy 🤖. Thanks again for considering this change!

It looks like there is an issue in the build for Python that I'll need to look into further, likely tomorrow.

cmake/tests.cmake
```diff
@@ -6,6 +6,10 @@ option(protobuf_ABSOLUTE_TEST_PLUGIN_PATH
   "Using absolute test_plugin path in tests" ON)
 mark_as_advanced(protobuf_ABSOLUTE_TEST_PLUGIN_PATH)
 
+# The compile_proto_file macro places files in the source tree, so this option
+# must be turned off.
+set(CMAKE_DISABLE_SOURCE_CHANGES OFF)
```
acozzette, line 11: Can you say more about why this change is needed here?

tusharb86: If I comment out this line, I get the following error during configuration from CMake:
```
CMake Error at cmake/tests.cmake:104 (add_custom_command):
  OUTPUT path

    G:/...../src/google/protobuf/map_lite_unittest.pb.cc

  in a source directory as an output of custom command.
```

This is because the `compile_proto_file` macro adds a custom command that generates output directly into the source tree (which those files expect). My guess is that this has to do with a new check in newer versions of CMake. The failure occurs with CMake version 3.22.2.

I can add more details to the comment block, if that would help.

tusharb86:
@acozzette I wanted to check in on this PR and what kinds of changes I should plan on making. Thanks!

acozzette:
@tusharb86 Sorry for the delay in replying! I think I underestimated the number of static_casts that would be required, so I am just trying to ask around on the protobuf team to make sure we are OK with this.

If we are going to enable these stricter flags then one other thing I think we need to do is enable the equivalent flags for Clang as well, since that is what most of us use for our day-to-day development. We need to make sure the Clang flags are at least as strict, because we wouldn't want to do a bunch of development against Clang and then find out later that it doesn't build with MSVC. I need to do some more research to figure out where to set these flags. Abseil has an interesting setup here that I might try to copy: https://github.com/abseil/abseil-cpp/blob/master/absl/copts/copts.py

acozzette:
I asked some C++ experts about this and unfortunately I don't think we will be able to accept this change as it is now. The main concern is that all the static_casts are silencing the warnings by making the truncation explicit, but this doesn't really solve the underlying problem that the warning is complaining about. So I think if we want to enable these warnings and make them errors, we need to do it in a way that keeps the casting to a minimum, probably by changing a lot of types we use to remove the need for explicit casting. One other thing to be aware of is that we are going to take a dependency on [Abseil](https://github.com/abseil/abseil-cpp) soon, but I am not sure if they are willing to commit to the same level of compiler strictness.

If you're still interested in working on this then maybe one good way to make some progress would to be to focus on error 4244, since it sounds like perhaps that one requires less cleanup than the `size_t`-related errors (though that is just a guess as I haven't checked for sure).

tusharb86:
@acozzette Thanks for the update! (Sorry for the delay - I've been out of the office a bit) I understand the reasoning and the desire to fix the underlying issues and move to using different types. I think that using the helper methods (i.e. `ToIntSize`, `FromIntSize`, or a more generic one) would be another way to solve the problem. However, that doesn't seem to align with the [Google C++ Style Guide link](https://google.github.io/styleguide/cppguide.html#Integer_Types) that you had shared previously, so I'm not sure that route would make sense either. At this time, I will plan on closing this PR without merging it.

On that topic, I'm curious to learn more about what kinds of analyzers or sanitizers are run on the codebase to ensure there are no overflow/underflow issues that could lead to security defects. Do you happen to have a pointer to tests that are running to catch these types of defects?

Separately, in another PR, I would like to make a change to protobuf/compiler/cpp/parse_function_generator.cc on [line 1433](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/cpp/parse_function_generator.cc#L1433). Currently, a `uint64_t` is being converted implicitly to an `int` in the call to `*_IsValid`. I believe that fixing this would result in generated files no longer needing to be compiled with these warnings disabled. What is the best way to go about fixing this? Should we check that the value can fit within an integer before making the explicit cast (similar to [`ToIntSize`](https://github.com/protocolbuffers/protobuf/blob/c38281dd20e562bac239bc77ab2fa10f71661708/src/google/protobuf/message_lite.h#L122)) using a `GOOGLE_DCHECK`?

Thanks for your help with all of this!

acozzette:
We use the LLVM sanitizers (e.g. asan/msan/ubsan) quite a bit and also do some fuzzing, but I don't have any pointers since we do all this internally only.

For the `_IsValid()` issue, the `ToIntSize`-style approach sounds good to me. It would probably even be fine to directly put in the `static_cast` without a helper function.

tusharb86:
> We use the LLVM sanitizers (e.g. asan/msan/ubsan) quite a bit and also do some fuzzing, but I don't have any pointers since we do all this internally only.
> 
> For the `_IsValid()` issue, the `ToIntSize`-style approach sounds good to me. It would probably even be fine to directly put in the `static_cast` without a helper function.

@acozzette - I finally got a chance to get this change ready for PR and opened #10353. Let me know if any changes are needed. Thanks!

PR #10043, opened 2022-05-25
[Bazel] Extract sources from test_suite in cc_dist_library

dlj-NaN:
This allows `cc_dist_library` to get the tests from a `test_suite`.
Unlike the behavior of `cc_dist_library` with `cc_library`, this is
done recursively, by propagating the aspect along the `test_suite`
structure (which is different from `deps`).

The upshot is that we can now use `cc_dist_library` to build lists of
input sources for tests, similar to how it already does this for
cc_library targets.


dlj-NaN:
Huh, well I guess the combination of wildcards is still problematic. I'm going to close this PR, and I'll reformulate (probably around filegroups instead).

PR #10056, opened 2022-05-26
Remove Abseil dependency from CMake build

acozzette:
We will reintroduce this soon, but we do not want it to take effect yet in the current 3.21 release.


mkruskal-google (approved):

kokoro/release/python/linux/build_artifacts.sh
```diff
@@ -30,7 +30,7 @@ git clone https://github.com/matthew-brett/multibuild.git
 # silently creeping in (see https://github.com/protocolbuffers/protobuf/issues/9180).
 # IMPORTANT: always pin multibuild at the same commit for:
 # - linux/build_artifacts.sh
-# - macos/build_artifacts.sh
+# - linux/build_artifacts.sh
```
Mizux, line 33: should NOT have been reverted same below

yoavmil:
note that when I tried to install protobuf via vcpkg it gave an error message.
the install command:
```
PS ...\vcpkg> vcpkg upgrade protobuf:x64-windows --no-dry-run
The following packages will be rebuilt:
  * grpc[codegen,core]:x64-windows -> 1.49.0
  * protobuf[core]:x64-windows -> 3.21.6
```
the error message:
```
Error: Building package grpc:x64-windows failed with: BUILD_FAILED
```
and inside the log file:
```
CMake Error at CMakeLists.txt:2407 (target_link_libraries):
  Target "grpc" links to:

    absl::any_invocable
```

```
 vcpkg --version
Vcpkg package management program version 2022-03-09-1affd32f93b299d5a907816c328ca3ededb73a7e
```

mkruskal-google:
protobuf doesn't depend on grpc, so there should be no reason we're causing an upgrade of it.  It sounds like you have a more recent version of gRPC installed that depends on Abseil, but don't have Abseil installed

PR #10080, opened 2022-05-30
Include endianness header on QNX targets

### LINKED ISSUE #9920: Bug: Big endian assumed on little endian QNX targets

Version: v3.15 and later
Language: C++

QNX SDP 7.0 for aarch64 (little endian)

gcc 5.4 (supplied with QNX SDP 7.0)

**What did you do?**

Compile protobuf for a little endian aarch64 QNX SDP 7.0 target.

`PROTOBUF_LITTLE_ENDIAN` should be defined.

**What did you see instead?**

`PROTOBUF_LITTLE_ENDIAN` is not defined.  We have root-caused this to be why we have observed byte-swapped `fixed64` fields (!!).

The method to determine the endianess of the target changed in v3.15.  Prior to this, the header `sys/param.h` was used.  It was replaced by a method that includes `endian.h` for several different platforms.  A later change (bd9d47db41) skipped over skipped over the inclusion of `endian.h` for QNX targets, since the file does not exist on the QNX platform.  However, this leaves neither `sys/param.h` nor `endian.h` included for QNX targets, resulting in `PROTOBUF_LITTLE_ENDIAN` _not_ being defined on little endian QNX targets.

We suggest small change like this could be applied to `src/google/protobuf/stubs/port.h` and `src/google/protobuf/io/coded_stream.h`
```
 #include <sys/endian.h>  // __BYTE_ORDER
+#elif defined(__QNX__)
+#include <sys/param.h>  // __BYTE_ORDER
 #else
```

dsche:
Fix #9920 in the proposed way.

### DIFF
--- a/src/google/protobuf/stubs/port.h
+++ b/src/google/protobuf/stubs/port.h
@@ -61,6 +61,8 @@
 #include <machine/endian.h>  // __BYTE_ORDER
 #elif defined(__FreeBSD__)
 #include <sys/endian.h>  // __BYTE_ORDER
+#elif defined(__QNX__)
+#include <sys/param.h>  // BYTE_ORDER
 #elif (defined(sun) || defined(__sun)) && (defined(__SVR4) || defined(__svr4__))
 #include <sys/isa_defs.h>  // __BYTE_ORDER
 #elif defined(_AIX) || defined(__TOS_AIX__)


acozzette:
@dsche Thank you for the pull request, but before I can review this we will need to have a CLA signed. Let me know if you have any trouble with the CLA.

PR #10089, opened 2022-05-31
Remove expected cmake failure due to missing abseil sources, since ab…

zhangskz:
, since absl dep was rolled back

### DIFF
--- a/kokoro/linux/cmake_distcheck/build.sh
+++ b/kokoro/linux/cmake_distcheck/build.sh
@@ -49,18 +49,4 @@ fi
 
 tar -C ${DIST_WORK_ROOT} --strip-components=1 -axf ${DIST_ARCHIVE}
 
-#
-# Run tests using extracted sources
-#
-if SOURCE_DIR=${DIST_WORK_ROOT} \
-  CMAKE_GENERATOR=Ninja \
-  CTEST_PARALLEL_LEVEL=$(nproc) \
-  kokoro/common/cmake.sh; then
-  # TODO: remove this conditional.
-  # The cmake build is expected to fail due to missing abseil sources.
-  echo "$0: Expected failure, but build passed." >&2
-  echo "Please update $(basename $0) to remove failure expectation." >&2
-  echo "FAIL" >&2
-  exit 1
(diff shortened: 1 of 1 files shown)


kokoro/linux/cmake_distcheck/build.sh
```diff
@@ -49,18 +49,4 @@ fi
 
 tar -C ${DIST_WORK_ROOT} --strip-components=1 -axf ${DIST_ARCHIVE}
 
-#
-# Run tests using extracted sources
-#
-if SOURCE_DIR=${DIST_WORK_ROOT} \
-  CMAKE_GENERATOR=Ninja \
-  CTEST_PARALLEL_LEVEL=$(nproc) \
-  kokoro/common/cmake.sh; then
```
mkruskal-google, line 58: I think we still want to do all of this, we just expect cmake.sh to return 0 I think?

zhangskz: Done...I think. Based on the line for check_missing_dist_files.sh, it looks like we may not need to explicitly check the status for cmake.sh right?

mkruskal-google: Hmm idk it makes me nervous to assume that (it could just as easily be that check_missing_dist_files is setup to silently fail).  Can you try intentionally breaking the cmake build and see if this test breaks?

acozzette (approved):

kokoro/linux/cmake_distcheck/build.sh
mkruskal-google: I think you alluded to this, but I just want to clarify: we don't need to explicitly check the return value of `cmake.sh`, if it fails the error will propagate up and cause a test failure?

zhangskz:
MacOS Ruby 2.7 failure should be an unrelated flake.

kokoro/linux/cmake_distcheck/build.sh
dlj-NaN: Yeah, the standard `set -e` handles this (line 5).

PR #10095, opened 2022-06-01
.NET tidy-up: remove .NET 3.5, add .cr to gitignore

### LINKED ISSUE #9888: .NET 3.5 support is now infeasible for C# - remove references to it
The [C# README](https://github.com/protocolbuffers/protobuf/blob/main/csharp/README.md) suggests that .NET 3.5 is supported, and indeed there's quite a lot of code that is conditional on .NET 3.5.

However, since all of the serialization changes to use `Span<T>` in the middle of 2020, it's basically infeasible to make the current codebase work with .NET 3.5 without an entirely separate path just for that platform.

IMO we should remove all mention of .NET 3.5 in both the README and the code.

cc @jtattermusch

jskeet:
This is effectively a no-op from a customer perspective - it's just getting rid of unused code.

Fixes #9888.

cc @amanda-tarafa

### DIFF
--- a/csharp/.gitignore
+++ b/csharp/.gitignore
@@ -18,6 +18,7 @@ lib/NUnit
 # 	Untracked files
 #
 .vs
+.cr
 *.user
 *.suo
 *.nupkg


amanda-tarafa (approved):
LGTM

Issue #10098, opened 2022-06-01
Remove const from function return type if returning const value


nglee:
**What language does this apply to?**
c++ codebase

**Describe the problem you are trying to solve.**
It is better to avoid the `const` qualifier on function return type if the function returns by value. The `const` qualifier may prevent move semantics if the type is movable. https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/#return-types-never-const

From [Scott Meyer](https://www.aristeia.com/BookErrata/ec++3e-errata.html):
> Declaring by-value function return values const will prevent their being bound to [rvalue references in C++0x](http://www.artima.com/cppsource/rvalue.html). Because rvalue references are designed to help improve the efficiency of C++ code, it's important to take into account the interaction of const return values and the initialization of rvalue references when specifying function signatures.

**Describe the solution you'd like**
Remove `const` from function return type if returning const value.

**Describe alternatives you've considered**
Leave it as it is if the `const` qualifiers are there for some reasons.

**Additional context**
I would like to merge [this commit](https://github.com/nglee/protobuf/commit/ab5a485c2959aef027881972d21fd1e6f1b13763) if accepted.

mkruskal-google:
Seems reasonable to me, can you send us a PR?

PR #10101, opened 2022-06-02
remove const from function return type if returning const value

nglee:
https://github.com/protocolbuffers/protobuf/issues/10098


mkruskal-google (approved):
Thanks for the cleanup!

PR #10107, opened 2022-06-04
Update inlined_string_field.h

secup:
Modify to #ifdef to prevent certain projects to report warnings are errors. thx u

PR #10111, opened 2022-06-06
Add -Wundef to our Bazel build flags

acozzette:
This should help prevent future warnings like the one being fixed in #10107.

### DIFF
--- a/kokoro/linux/bazel/build.sh
+++ b/kokoro/linux/bazel/build.sh
@@ -21,6 +21,7 @@ bazel_args=(
   test
   --keep_going
   --copt=-Werror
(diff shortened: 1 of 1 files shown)


mkruskal-google (approved):

acozzette:
Note: the Bazel test is currently failing but that is just because it is surfacing the error as intended.

mkruskal-google:
FWIW when I made a similar change to this I ended up just duplicating it in github

acozzette:
Hmm, unfortunately this causes a bunch of errors in googletest. I think I will give up on this for now until I have time to figure out a good way to apply the stricter flags to our own code only.

PR #10112, opened 2022-06-06
Fix usage of sprintf in strutil.cc

strangewiz:
sprintf has been explicitly marked as deprecated in Xcode 14.

### DIFF
--- a/src/google/protobuf/stubs/strutil.cc
+++ b/src/google/protobuf/stubs/strutil.cc
@@ -503,7 +503,7 @@ int CEscapeInternal(const char* src, int src_len, char* dest,
              (last_hex_escape && isxdigit(*src)))) {
           if (dest_len - used < 4) // need space for 4 letter escape
             return -1;
-          sprintf(dest + used, (use_hex ? "\\x%02x" : "\\%03o"),
+          snprintf(dest + used, 5, (use_hex ? "\\x%02x" : "\\%03o"),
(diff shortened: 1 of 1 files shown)


src/google/protobuf/stubs/strutil.cc
```diff
@@ -503,7 +503,7 @@ int CEscapeInternal(const char* src, int src_len, char* dest,
              (last_hex_escape && isxdigit(*src)))) {
           if (dest_len - used < 4) // need space for 4 letter escape
             return -1;
-          sprintf(dest + used, (use_hex ? "\\x%02x" : "\\%03o"),
+          snprintf(dest + used, 4, (use_hex ? "\\x%02x" : "\\%03o"),
```
acozzette, line 506: Some of the tests are failing and I suspect it's because 4 is not enough characters when you include the trailing null byte. So I think this should be a 5 instead of a 4. Fortunately this logic will all be replaced by Abseil fairly soon.

acozzette (changes requested):

src/google/protobuf/stubs/strutil.cc
strangewiz: Makes sense.  Updated to 5.  Will the tests automatically again?

strangewiz:
I can't tell if that JRuby failure is related.  Any ideas?

acozzette (approved):

PR #10121, opened 2022-06-08
Protobuf bytes field decoding allocates extra memory bytes field is …

sureshmbabu1:
Protobuf decoding allocates extra memory when bytes field is decoded to c++ string

### DIFF
--- a/src/google/protobuf/io/coded_stream.cc
+++ b/src/google/protobuf/io/coded_stream.cc
@@ -286,6 +286,8 @@ bool CodedInputStream::ReadStringFallback(std::string* buffer, int size) {
     if (bytes_to_limit > 0 && size > 0 && size <= bytes_to_limit) {
       buffer->reserve(size);
     }
+  } else {
+    buffer->reserve(size);
   }
 
   int current_buffer_size;


amandeepgautam:
@acozzette Can you please review the change?

acozzette:
@sureshmbabu1 Thank you for the pull request, but unfortunately I don't think we can accept this one. While the memory allocation behavior here may seem suboptimal from a performance perspective, it was set up this way intentionally. The thing is that if we are parsing untrusted inputs, we don't want to allocate an unbounded amount of memory upfront. If we did this then an attacker could use a very small payload to trigger almost 2 GiB of allocations.

amandeepgautam:
@acozzette We understand that concern, but we are looking for a solution. What happens with repeated use of string::append is that we end up with twice the amount of memory usage. So, a 4Mb memory usage can end up consuming 8MB memory.

acozzette:
Have you tried setting a limit via CodedInputStream::SetTotalBytesLimit()? That should allow the memory to be reserved as long as it falls below your provided limit. Out of curiosity, what kind of parsing are you doing that is having this performance problem? The reason I ask is that due to some refactoring in the past few years, we no longer use CodedInputStream in the main binary format parser (though it is still used in various places such as text format parsing if I'm not mistaken).

amandeepgautam:
We have not tried using CodedInputStream::SetTotalBytesLimit. We are not familiar with this code so much. We have multiple unary RPCs with 4MB or 8MB payload in our use case. There can be a few 10s of such RPCs together. If all RPCs happen to land on grpc server together, we will happen to have a memory usage that's twice what we need. Because of this, we have some monitors which are fired causing false alarms.

acozzette:
OK, I see. You might want to try asking the gRPC maintainers about this because this limit would likely have to be set somewhere in gRPC's internals.

PR #10124, opened 2022-06-09
DO NOT MERGE! Integrate from Piper for C++, Java, and Python

elharo:
DO NOT MERGE!


CHANGES.txt
```diff
@@ -1,3 +1,7 @@
+
+2022-06-?? version 21.?
```
acozzette, line 2: We usually just put "Unreleased Changes" here and then whoever does the release will update that with the actual date and version number.

src/Makefile.am
```diff
@@ -18,7 +18,7 @@ else
 PTHREAD_DEF =
 endif
 
-PROTOBUF_VERSION = 32:1:0
+PROTOBUF_VERSION = 32:0:0
```
acozzette, line 21: I think the internal template needs to be updated so that the changes to this file are not synced out.

All the changes in the cmake/ directory can also be reverted because those should not be using the internal templates anymore. Let me send out a change to delete those templates and add the cmake/ files to the exclusion list.

elharo: @dlj-NaN said something about Kokoro needing those Makefiles so we might have to copy them into Piper instead of simply deleting them.

Issue #7826, opened 2020-08-18
FileOptions with dynamic libraries: Multiple extension registrations


gudvinr:
**What version of protobuf and what language are you using?**
Version: v3.6.1
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Linux: Ubuntu 18.04 (KDE Neon)

**What runtime / compiler are you using (e.g., python version or gcc version)**
GCC: 7.5.0
gRPC: v1.16.1

**What did you do?**
1. Add custom option to `.proto`
2. Compile `.proto` files to static lib
3. Link this library to executable (compiled with `-rdynamic`)
4. Link this library to dynamic libraries
5. Open dynamic libraries using `dlopen` (with `RTLD_GLOBAL`)
6. Executable not loading and `libprobobuf` producing an error

**What did you expect to see**
Correctly running application which uses protobuf extensions

**What did you see instead?**
Application not working and `libprotobuf` throwing exception:
```
[libprotobuf FATAL …/protobuf/src/google/protobuf/extension_set.cc:93] Multiple extension registrations for type "google.protobuf.FileOptions", field number 50000.
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  Multiple extension registrations for type "google.protobuf.FileOptions", field number 50000.
```

**Anything else we should know about your project / environment**

There are no extensions with same field number (and no extensions with same name either).  
It is not possible to link static library only with executable because symbols inside shared libraries can't be found.  
There are usage of protobuf types inside executable so you can't link library to dynamic libs too.  
Using shared library is not possible without significant changes to build process.

acozzette:
It sounds like this custom option is either getting linked in more than once place, or getting dynamically loaded multiple times. To fix the problem you will have to make sure the option is only registered once.

gudvinr:
@acozzette 

> It sounds like this custom option is either getting linked in more than once place, or getting dynamically loaded multiple times.

It is exactly what I said:
2. Compile `.proto` files to static lib
3. Link this library to executable (compiled with `-rdynamic`)
4. Link this library to dynamic libraries

`libprotobuf` then then calls `Register` function for extensions every time it loads. And this causing errors.

Dlougach:
Is there any workaround for this?

fmvin:
We use "MessageOptions" to check a type which should be de-serialized in several dynamic libraries in one app. And the same issue occurs. Please provide solution for this issue or any workaround.

[...]

Issue #6029, opened 2019-04-12
FileOptions multiple extension registration


huta4u:
I'm trying to utilize file level extension using FileOptions to version proto files.

I created following files:

**Version.proto**
```
import "google/protobuf/descriptor.proto";
extend google.protobuf.FileOptions {
  string version = 50001;
}
```

I have two additional protos that import Version.proto and use option(version) extension:

**A.proto**
```
import "Version.proto";
option (version) = "1.0.1";
```

**B.proto**

```
import "Version.proto";
option (version) = "1.0.1";
```

My setup: Windows 10, C++ VS 2017, protobuf ver 3.4.0. All three protos are compiled by protoc.exe, the output headers & cc files are included into the project. Protobuf.dll linked dynamically.

When I run the application that uses mentioned proto files I get following error just upon startup:

> [libprotobuf FATAL google\protobuf\extension_set.cc:102] Multiple extension registrations for type google.protobuf.FileOptions, field number 50001.

Duplicated from https://stackoverflow.com/questions/55627296/fileoptions-extension-in-multiple-protos

acozzette:
I suspect that the generated code for version.proto is somehow getting linked more than once, which would explain why the problem persists even if you choose a different extension number. Maybe there are multiple DLL's that all include version.proto's generated code?

huta4u:
@acozzette Good recognition. The setup looks this way:
Static lib cotains proto h&cc files. It is linked to DLL and EXEC. The DLL is also linked to EXEC.

acozzette:
OK, I think the important thing then is just to restructure the app so that the version.proto generated code is only linked once.

gmalara:
The thing is the proto files are used both in DLL and EXEC. The DLL is shared by other parts of the system also. If I remove static lib containing h&cc files from EXEC or DLL they won't compile.

Dlougach:
Is there a workaround, I am experiencing the same issue when a dynamically linked binary crashes (while the same binary linked statically works). I am hoping that linking libprotobuf dynamically everywhere would fix this, but perhaps someone can confirm?

fahhem:
Related to #7826

tobegit3hub:
We have the similar issue and load the dynamic library twice which static link to the same protobuf generated library. It works in Linux but crashes in MacOS. It would be great to support in MacOS or we can do something for workaround.

runlilong:
Does someone have a solution for this situation?
@fahhem @tobegit3hub

PR #10127, opened 2022-06-09
Don't fail for identical ExtensionInfo registration

Luabee:
Fixes https://github.com/protocolbuffers/protobuf/issues/7826 and https://github.com/protocolbuffers/protobuf/issues/6029

Linking against a protobuf translation unit (`.pb.o`) more than once can cause this error. Currently there is no real workaround. 

See https://github.com/protocolbuffers/protobuf/issues/6029 for an example of how this problem can happen

### DIFF
--- a/src/google/protobuf/extension_set.cc
+++ b/src/google/protobuf/extension_set.cc
@@ -121,7 +121,7 @@ void Register(const ExtensionInfo& info) {
   static auto local_static_registry = OnShutdownDelete(new ExtensionRegistry);
   global_registry = local_static_registry;
   auto existing = local_static_registry->insert(info);
-  if (!existing.second && IsDifferent(existing.first, info)) {
+  if (!existing.second && IsDifferent(*existing.first, info)) {
       GOOGLE_LOG(FATAL) << "Multiple extension registrations for type \""
         << info.message->GetTypeName() << "\", field number "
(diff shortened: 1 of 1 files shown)


acozzette:
I think I understand the problem here, but I'm a bit skeptical of this kind of solution. It seems to me that even if this change succeeds as a workaround, linking the same `pb.o` file more than once is still an ODR violation that could lead to other problems. I would also expect conflicts from having the same descriptors registered in the global DescriptorPool more than once (though I believe some projects work around that by setting `option optimize_for = LITE_RUNTIME;` in proto files).

The only 2 ways I know of to reliably avoid ODR violations and other problems across dynamic libraries are these:
- Make sure each `pb.o` file appears in only one dynamic library.
- Alternatively, you can have multiple dynamic libraries using the same protos but make sure each library has its own independent protobuf world. This would mean that each dynamic library has its own copy of libprotobuf (or libprotobuf-lite), protos can't be passed across library boundaries, and symbols would have to be hidden carefully so that they don't conflict with each other.

Luabee:
I was working at similar `DescriptorPool` conflict resolution for a separate PR. My dev environment can handle the ODR worries just fine. Given the (hard to describe) constraints of my environment, it's these runtime complaints which are causing me problems.

PR #10128, opened 2022-06-09
Length check accommodates NUL emitted by sprintf

dneto0:
Update the length check in google::protobuf::CEscapeInternal to account
for the extra NUL character emitted by snprintf when escaping hex and
octal sequences.

That function is an internal detail, not exported via any header.
Internally, it is used in two places, and both calls make buffers
that do have space for the extra NUL.
So in the actual usage, the check is redundant.

### DIFF
--- a/src/google/protobuf/stubs/strutil.cc
+++ b/src/google/protobuf/stubs/strutil.cc
@@ -501,7 +501,9 @@ int CEscapeInternal(const char* src, int src_len, char* dest,
         if ((!utf8_safe || static_cast<uint8_t>(*src) < 0x80) &&
             (!isprint(*src) ||
              (last_hex_escape && isxdigit(*src)))) {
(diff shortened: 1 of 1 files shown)


dneto0:
I could not find unit tests for CEscapeInternal or for any of the CEscape* API methods.

I would be happy to add them if requested.

acozzette (approved):

acozzette:
@dneto0 Thanks for the fix! Ordinarily I would be worried about our lack of test coverage here, but this particular code will be going away soon once we start depending on Abseil for this functionality.

dneto0:
Sounds good to me.  I already checked Abseil and it does this bit differently.
https://github.com/abseil/abseil-cpp/blob/master/absl/strings/escaping.cc#L330 

Thanks very much.

PR #10129, opened 2022-06-10

elharo:


cmake/abseil-cpp.cmake
```diff
@@ -0,0 +1,31 @@
+if(TARGET absl::strings)
```
acozzette, line 1: We deleted this file but it looks like the sync is trying to resurrect it. I think this is one more file we need to remove from the sync list.

acozzette (approved):

cmake/abseil-cpp.cmake
elharo: If we deleted it, then we should just need to delete it in piper too.

PR #10135, opened 2022-06-13
Add package_as_prefix_forced_prefix generation option.

thomasvl:
This is a generation option that provides serves the same purpose as the
existing GPB_OBJC_USE_PACKAGE_AS_PREFIX_PREFIX environment variable; just
providing a different way to set/enable it.

### DIFF
--- a/objectivec/README.md
+++ b/objectivec/README.md
@@ -200,12 +200,13 @@ supported keys are:
         entry can be made as "no_package:PATH=prefix", where PATH is the
       path for the .proto file.
 
-  * `use_package_as_prefix` and `proto_package_prefix_exceptions_path`: The
-    `value` for `use_package_as_prefix` can be `yes` or `no`, and indicates
-    if a prefix should be derived from the proto package for all the symbols
-    for files that don't have the `objc_class_prefix` file option (mentioned
-    above). This helps ensure the symbols are more unique and means there is
-    less chance of ObjC class name collisions.
+  * `use_package_as_prefix`, `package_as_prefix_forced_prefix` and
+    `proto_package_prefix_exceptions_path`: The `value` for
+    `use_package_as_prefix` can be `yes` or `no`, and indicates if a prefix
+    should be derived from the proto package for all the symbols for files that
+    don't have the `objc_class_prefix` file option (mentioned above). This helps
+    ensure the symbols are more unique and means there is less chance of ObjC
+    class name collisions.
 
     To help in migrating code to using this support,
     `proto_package_prefix_exceptions_path` can be used to provide the path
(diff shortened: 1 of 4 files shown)


objectivec/README.md
```diff
@@ -200,23 +200,29 @@ supported keys are:
...
+    ensure the symbols are more unique and means there is less chance of ObjC
+    class name collisions.
 
     To help in migrating code to using this support,
     `proto_package_prefix_exceptions_path` can be used to provide the path
     to a file that contains proto package names (one per line, comments allowed
     if prefixed with `#`). These package won't get the derived prefix, allowing
     migrations to the behavior one proto package at a time across a code base.
 
+    `package_as_prefix_forced_prefix` can be used to provide a value that will
```
dmaclach, line 217: Hmm... I honestly don't understand what this means. Perhaps an example would help?

thomasvl: updated

objectivec/README.md
```diff
@@ -200,23 +200,29 @@ supported keys are:
...
+    be used before all prefixes derived from the packages to help group all of
+    these types with a common prefix. Thus is only make sense to use it when
+    `use_package_as_prefix` is also enabled.
+
     `use_package_as_prefix` currently defaults to `no` (existing behavior), but
```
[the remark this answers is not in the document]
thomasvl: done

dmaclach (approved):

thomasvl:
Error seems to be some hiccup with a header file that related to my change.

objectivec/README.md
```diff
@@ -200,23 +200,30 @@ supported keys are:
...
 
     To help in migrating code to using this support,
     `proto_package_prefix_exceptions_path` can be used to provide the path
     to a file that contains proto package names (one per line, comments allowed
     if prefixed with `#`). These package won't get the derived prefix, allowing
     migrations to the behavior one proto package at a time across a code base.
 
+    `package_as_prefix_forced_prefix` can be used to provide a value that will
+    be used before all prefixes derived from the packages to help group all of
+    these types with a common prefix. Thus is only make sense to use it when
```
acozzette, line 219: I think this should read: "Thus it only makes sense..."

objectivec/README.md
```diff
@@ -200,23 +200,30 @@ supported keys are:
...
+    ensure the symbols are more unique and means there is less chance of ObjC
+    class name collisions.
 
     To help in migrating code to using this support,
     `proto_package_prefix_exceptions_path` can be used to provide the path
     to a file that contains proto package names (one per line, comments allowed
     if prefixed with `#`). These package won't get the derived prefix, allowing
     migrations to the behavior one proto package at a time across a code base.
 
+    `package_as_prefix_forced_prefix` can be used to provide a value that will
```
acozzette, line 217: I found this name a little confusing at first. Would it make sense to just call it something like `forced_package_prefix`?

acozzette (approved):

objectivec/README.md
thomasvl: Done

objectivec/README.md
thomasvl: Since it only apples if package_as_prefix is also enable, that's why i left it there.  my expectation is it won't be that common to use, so I'm good with the slightly more verbose name to hopefully better indicate how it is used.

I can also look at adding a validation that this isn't set when the option isn't set; but it still feels like erroring on the side of the longer name is better.

PR #10141, opened 2022-06-15
Removing deleted files from dist rule

mkruskal-google:
These were deleted in https://github.com/protocolbuffers/protobuf/pull/10095 and broke our distribution artifact kokoro tests.


deannagarcia (approved):

jskeet (approved):
Humbug, so sorry about that. (I'd have expected it to show up on the PR that caused it - will check that PR in the morning.)

mkruskal-google:
> Humbug, so sorry about that. (I'd have expected it to show up on the PR that caused it - will check that PR in the morning.)

Yea I was surprised.  I'm not 100% sure but it *looks* like the tests never actually ran even though you added the kokoro:run tag

PR #10156, opened 2022-06-20
Remove useless return from zero_copy_stream_impl_lite.cc

### LINKED ISSUE #10154: clang compiler warning: io/zero_copy_stream_impl_lite.cc:379:10: warning: 'return' will never be executed [-Wunreachable-code-return]
```
protobuf/src/google/protobuf/io/zero_copy_stream_impl_lite.cc:379:10: warning: 'return' will never be executed [-Wunreachable-code-return]
  return true;
         ^~~~
```
Seems like we can just delete the line and I'll submit a PR for doing that.

rvandermeulen:
Fixes #10154.


fowles:
Generally speaking, warnings like this will cause a new warning on some other compiler if you don't have that return statement.  As such, we usually avoid warning like this one.

PR #10105, opened 2022-06-03
Update to C# 10 and upgrade code style

erikma:
Plus update C# README with Visual Studio 2022 info.

PR #10169, opened 2022-06-23
Adjust editor config to reflect cast conventions in C# libraries

jskeet:
Most of our code includes a space after the cast, e.g. `(int) x`
instead of `(int)x`. This .editorconfig change means that will be
the default formatting when it's performed by the editor.

cc @erikma who encountered this in #10105.


jtattermusch (approved):

jskeet:
The three failures are independent of this PR.

PR #10170, opened 2022-06-24
Set all but main project to nullable

erikmav:
Start with the periphery including test code.

### DIFF
--- a/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
+++ b/csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
@@ -119,11 +119,11 @@ namespace Google.Protobuf
         private static void AssertReadVarintFailure(InvalidProtocolBufferException expected, byte[] data)
         {
             CodedInputStream input = new CodedInputStream(data);
-            var exception = Assert.Throws<InvalidProtocolBufferException>(() => input.ReadRawVarint32());
+            var exception = Assert.Throws<InvalidProtocolBufferException>(() => input.ReadRawVarint32())!;
             Assert.AreEqual(expected.Message, exception.Message);
 
             input = new CodedInputStream(data);
-            exception = Assert.Throws<InvalidProtocolBufferException>(() => input.ReadRawVarint64());
+            exception = Assert.Throws<InvalidProtocolBufferException>(() => input.ReadRawVarint64())!;
             Assert.AreEqual(expected.Message, exception.Message);
 
             AssertReadFromParseContext(new ReadOnlySequence<byte>(data), (ref ParseContext ctx) =>
@@ -153,7 +153,7 @@ namespace Google.Protobuf
             }, false);
 
             // Make sure we get the same error when reading directly from a Stream.
-            exception = Assert.Throws<InvalidProtocolBufferException>(() => CodedInputStream.ReadRawVarint32(new MemoryStream(data)));
+            exception = Assert.Throws<InvalidProtocolBufferException>(() => CodedInputStream.ReadRawVarint32(new MemoryStream(data)))!;
             Assert.AreEqual(expected.Message, exception.Message);
         }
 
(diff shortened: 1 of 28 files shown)


erikma:
@jskeet from #9801

erikma:
Started this way simply to reduce later diffs: There are a lot of test code changes that did not need to be ported over from the other PR, but will be introduced incrementally once this goes in. This cuts the initial noise to make those PRs more focused on jus the related API changes.

jskeet:
Okay, I see what you mean. Yup, that's fine. (There are multiple ways of approaching this kind of change.) It's on my to-do list for next week.

csharp/src/Google.Protobuf.Benchmarks/ByteStringBenchmark.cs
```diff
@@ -46,7 +46,7 @@ public class ByteStringBenchmark
         private const int Megabyte = 1024 * 1024;
         private const int _10Megabytes = 1024 * 1024 * 10;
 
-        byte[] byteBuffer;
+        byte[] byteBuffer = new byte[0];
```
jskeet, line 49: Let's make this explicitly private at the same time.

csharp/src/Google.Protobuf.Benchmarks/ParseRawPrimitivesBenchmark.cs
```diff
@@ -46,15 +46,15 @@ namespace Google.Protobuf.Benchmarks
     public class ParseRawPrimitivesBenchmark
     {
         // key is the encodedSize of varint values
-        Dictionary<int, byte[]> varintInputBuffers;
+        Dictionary<int, byte[]> varintInputBuffers = new();
```
jskeet, line 49: Possibly add a comment to say that all of these initial values will be overwritten in GlobalSetup? (And make everything explicitly private.)

csharp/src/Google.Protobuf.Benchmarks/WriteRawPrimitivesBenchmark.cs
```diff
@@ -44,23 +44,23 @@ namespace Google.Protobuf.Benchmarks
     public class WriteRawPrimitivesBenchmark
     {
         // key is the encodedSize of varint values
-        Dictionary<int, uint[]> varint32Values;
-        Dictionary<int, ulong[]> varint64Values;
+        Dictionary<int, uint[]> varint32Values = new();
```
jskeet, line 47: Ditto possible comment. And can we make everything explicitly private at the same time?

csharp/src/Google.Protobuf.Test/CodedInputStreamTest.cs
```diff
@@ -119,12 +119,12 @@ private static void AssertReadVarint(byte[] data, ulong value)
         private static void AssertReadVarintFailure(InvalidProtocolBufferException expected, byte[] data)
         {
             CodedInputStream input = new CodedInputStream(data);
-            var exception = Assert.Throws<InvalidProtocolBufferException>(() => input.ReadRawVarint32());
-            Assert.AreEqual(expected.Message, exception.Message);
+            InvalidProtocolBufferException? exception = Assert.Throws<InvalidProtocolBufferException>(() => input.ReadRawVarint32());
```
jskeet, line 122: I suspect we can change this to be as it was before, just with `!` at the end of the `Throws` call. It really will be non-null, otherwise `Assert.Throws` will itself throw.

jskeet (commented):
Just a few nits - but generally looking good. Sorry this took so long to get round to!

jskeet (approved):
Running Kokoro tests now...

JamesNK:
This is the wrong way around to do this. The main project should be made nullable first, then other projects. Otherwise, you'll need to react to all the nullable changes as main project is annotated.

jskeet:
@JamesNK: I don't think there's an objective "right" or "wrong" here - just different pros and cons.

I completely acknowledge that this way round means multiple iterations of the test projects - but I see that as a *benefit*. By doing it that way, you see the changes required in tests at the same time as the changes required in the main project, *without* any distractions from non-protobuf changes required for the tests.

That, to me, is a good way of getting a "sniff test" about whether the changes to the main project are appropriate - if the required changes to the tests seem reasonable, that's encouraging. It's a lot harder to pick those out when *all* the changes required to the tests come in the same commit.

PR #10175, opened 2022-06-25
Update field_mask.proto

sam2butnot3:
Updat


elharo:
Can you explain in more detail why you want these changes?

PR #10179, opened 2022-06-27
Update the java string comparison for boolean json values.

barbacbd:


barbacbd:
There isn't currently. I can definitely add one in.

elharo:
This change is not OK per the JSON spec. JSON allows "true" and "false" for boolean values. It does not allow TRUE, False, ERROR, etc.

PR #10180, opened 2022-06-27
Remove autotools protoc scripts

deannagarcia:
Delete scripts that build protoc with autotools and remove references of them from docs.


protoc-artifacts/README.md
```diff
@@ -32,130 +22,3 @@ The name of a published ``protoc`` artifact is in the following format:
...
-The above command will produce a `target/linux/x86_64/protoc` binary under the
-protoc-artifacts directory.
-
-For a list of supported platforms, see the comments in the build-protoc.sh
-script. We only use this script to build artifacts on Ubuntu and MacOS (both
-with x86_64, and do cross-compilation for other platforms.
-
-### Tips for building for Linux
-We build on Centos 6.9 to provide a good compatibility for not very new
-systems. We have provided a ``Dockerfile`` under this directory to build the
```
mkruskal-google, line 74: Should we delete the docker file too?

protoc-artifacts/README.md
```diff
@@ -2,20 +2,10 @@
...
-repositories. The artifact can be used by build automation tools so that users
-would not need to compile and install ``protoc`` for their systems.
-
-If you would like us to publish protoc artifact for a new platform, please send
-us a pull request to add support for the new platform. You would need to change
-the following files:
-
-* [build-protoc.sh](build-protoc.sh): script to cross-build the protoc for your
-  platform.
-* [pom.xml](pom.xml): script to upload artifacts to maven.
```
mkruskal-google, line 16: Is this one still used?  It seems weird to delete the documentation for this file but not delete the file

deannagarcia: This is a little in flux right now, but the pom file will only be used by us to upload the artifacts to maven so I don't think we need to document it for external users.

mkruskal-google: Could we move it or document that it's for internal use only?  The existence of this folder seems a little confusing after this PR

deannagarcia: Documented that it's only for internal use. I think after the 21.x releases we may be able to restructure this folder, but I will punt on that for now.

mkruskal-google (approved):

PR #10189, opened 2022-06-29
Fix C# build by upgrading to C# 10

deannagarcia:
PR https://github.com/protocolbuffers/protobuf/pull/10105/ updated the C# version to 10 and made changes that were incompatible with the old version. By mistake, https://github.com/protocolbuffers/protobuf/pull/10183 reverted the version change which in turn broke the build. Updating the C# version will fix. 

Other failures are unrelated and partially fixed by: https://github.com/protocolbuffers/protobuf/pull/10190


deannagarcia:
Closing since the fix was done here: https://github.com/protocolbuffers/protobuf/pull/10195

PR #10192, opened 2022-06-30
Clean up mac PHP kokoro tests

mkruskal-google:
PHP 7.0 is no longer tested or explicitly supported, and we use homebrew to install 7.3 for the remaining test.

### DIFF
--- a/kokoro/macos/php7.0_mac/build.sh
+++ b/kokoro/macos/php7.0_mac/build.sh
@@ -2,10 +2,24 @@
 #
 # Build file to set up and run tests
 
+set -eux
+
 # Change to repo root
 cd $(dirname $0)/../../..
 
 # Prepare worker environment to run tests
 source kokoro/macos/prepare_build_macos_rc
+
+brew tap shivammathur/php
+
+# Install dependencies.
+brew install \
+    shivammathur/php/php@7.3 \
+    composer
+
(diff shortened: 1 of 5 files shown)


haberman:
We should get sign-off from @bshaffer and @dwsupplee before deprecating any old version of PHP.

mkruskal-google:
> We should get sign-off from @bshaffer and @dwsupplee before deprecating any old version of PHP.

Great, thanks Josh!  I'll ping them once this is green

deannagarcia:
Can we prioritize getting a PR out to fix the failures in the tests and then decide whether to deprecate PHP 7.0?

perezd:
Let's separate the two concerns here:

1.  We want to tighten our support window (I am in favor of this)
2. We need to immediately unblock our presubmit tests to ensure we're not introducing critical regressions.

Lets focus on getting point 2 taken care of asap, as we need additional time to get buyin on point 1. It is unacceptable for us be merging code without tests running for a prolonged period of time. Can we focus our efforts there primarily?

mkruskal-google:
I was trying to take a middle ground approach here, since I thought 7.3 would be easy to fix and 7.0 wasn't worth fixing.  It turns out 7.3 was harder than I expected (because we're using such an old mac os, so disabling the tests sounds reasonable)

Issue #7474, opened 2020-05-07
Argument file needs support for multi-char and BOM


Falco20019:
**What version of protobuf and what language are you using?**
libprotoc 3.3.0 / Grpc.Tools 2.28.1

**What operating system (Linux, Windows, ...) and version?**
Windows 10 Build 1803

**What did you do?**
1. Create `test.rsp` containing non-ascii:
```
--plugin=protoc-gen-grpc=ágrpc_csharp_plugin.exe
--grpc_out=grpc_out
test.proto
```
2. Run `protoc.exe @test.rsp`

**What did you expect to see**
`protoc` running and creating the files.

**What did you see instead?**
UTF-8 without BOM: `--grpc_out: protoc-gen-grpc: Das System kann die angegebene Datei nicht finden.`
UTF-8 with BOM: `You seem to have passed an empty string as one of the arguments to C:\Users\dejhbk0l\.nuget\packages\grpc.tools\2.28.1\tools\windows_x64\protoc.exe.  This is actually sort of hard to do.  Congrats.  Unfortunately it is not valid input so the program is going to die now.`

**Anything else we should know about your project / environment**
https://github.com/protocolbuffers/protobuf/blame/master/src/google/protobuf/compiler/command_line_interface.cc#L1326 uses `std::ifstream` instead of `std::wifstream` and therefore does not allow BOM or wide characters.

This is a minimal example. We stumbled over this in https://github.com/grpc/grpc/issues/17995 as we distribute `protoc` and the `protoc-gen-grpc` through NuGet and the default for `dotnet nuget locals global-packages` is `%userprofile%\.nuget\packages\`. For some of our users, the username includes non-ascii characters which leads to `protoc-gen-grpc` containing non-ascii characters and the argument file not being parsable.

Falco20019:
CC:
* @anandolee because it's protoc
* @acozzette as implementer of the method
* @jtattermusch as related to `Grpc.Tools`

jtattermusch:
I'm not sure if making the ExpandArgumentFile function support `std:wifstream` is feasible (the functions returns a std::string which is then processed by the command_line_interface.cc logic, and the entire logic is written in terms of std::string and it doesn't know about wide characters at all).

https://github.com/protocolbuffers/protobuf/blob/3f5fc4df1de8e12b2235c3006593e22d6993c3f5/src/google/protobuf/compiler/command_line_interface.cc#L1425

Falco20019:
[...]

PR #10200, opened 2022-07-01
Fix for grpc.tools #17995 & protobuf #7474 (handle UTF-8 paths in argumentfile)

tonydnewell:
Fix for https://github.com/protocolbuffers/protobuf/issues/7474 (and https://github.com/grpc/grpc/issues/17995)

When calling the plugins in **subprocess.cc** `Subprocess::Start()`  fix to use CreateProcessW instead of CreateProcessA to handle non-ascii paths.

See https://github.com/grpc/grpc/issues/17995#issuecomment-1171385170 for a summary of what the actual problem is and why the fix looks like the way it does.

### DIFF
--- a/src/google/protobuf/compiler/subprocess.cc
+++ b/src/google/protobuf/compiler/subprocess.cc
@@ -148,7 +148,7 @@ void Subprocess::Start(const std::string& program, SearchMode search_mode) {
   // Create the process.
   PROCESS_INFORMATION process_info;
 
-  if (CreateProcessW((search_mode == SEARCH_PATH) ? NULL : wprogram.c_str(),
+  if (CreateProcessW((search_mode == SEARCH_PATH) ? nullptr : wprogram.c_str(),
                      (search_mode == SEARCH_PATH) ? wcommand_line_copy : NULL,
                      nullptr,  // process security attributes
                      nullptr,  // thread security attributes
--- a/src/google/protobuf/compiler/subprocess.cc
+++ b/src/google/protobuf/compiler/subprocess.cc
@@ -47,6 +47,7 @@
 #include <google/protobuf/stubs/common.h>
 #include <google/protobuf/stubs/substitute.h>
 #include <google/protobuf/message.h>
+#include <google/protobuf/io/io_win32.h>
 
 namespace google {
 namespace protobuf {
@@ -113,7 +114,7 @@ void Subprocess::Start(const std::string& program, SearchMode search_mode) {
   }
 
   // Setup STARTUPINFO to redirect handles.
-  STARTUPINFOA startup_info;
+  STARTUPINFOW startup_info;
   ZeroMemory(&startup_info, sizeof(startup_info));
   startup_info.cb = sizeof(startup_info);
   startup_info.dwFlags = STARTF_USESTDHANDLES;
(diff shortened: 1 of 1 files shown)


jtattermusch:
@acozzette  can you please review? This is an important fix for gRPC's codegen support and you seems to be the author of the argumentFile parsing logic in protoc.

jtattermusch:
@tonydnewell  just to clarify:
- based on the analysis https://github.com/grpc/grpc/issues/17995#issuecomment-1171385170 and the changes made here, it seems that the only problematic part is if the protoc plugin's full path (grpc_csharp_plugin in our case) contains non-ASCII characters. IIUC if other arguments (e.g. the .proto file names) contain non-ascii characters in their path, it isn't   currently a problem (did you verify that?), because the logic for handling UTF-8 file names for other params already exists. Only for the protoc plugin arg, things are different since the invocation of the protoc plugin happens through subprocess.cc
- Based on your investigation, can you tell whether the problem is only specific to when using the argument file (`protoc.exe @test.rsp`) as originally reported or if there are other ways of reproducing it (e.g. invoking protoc.exe with args on command line normally)?

src/google/protobuf/compiler/subprocess.cc
```diff
@@ -125,17 +126,30 @@ void Subprocess::Start(const std::string& program, SearchMode search_mode) {
...
-  char* command_line =
-      portable_strdup(("cmd.exe /c \"" + program + "\"").c_str());
+  wchar_t *wcommand_line_copy = _wcsdup(wcommand_line.c_str());
 
   // Create the process.
   PROCESS_INFORMATION process_info;
 
-  if (CreateProcessA((search_mode == SEARCH_PATH) ? nullptr : program.c_str(),
-                     (search_mode == SEARCH_PATH) ? command_line : nullptr,
+  if (CreateProcessW((search_mode == SEARCH_PATH) ? NULL : wprogram.c_str(),
```
jtattermusch, line 151: nit: change NULL -> nullptr everywhere in the diff (as that seems to be the current codestyle requirement in protobuf codebase?)

tonydnewell:
> @tonydnewell just to clarify:
> 
> * based on the analysis [Grpc.Tools: Accented characters in nuget path breaks c# grpc tool compiler grpc/grpc#17995 (comment)](https://github.com/grpc/grpc/issues/17995#issuecomment-1171385170) and the changes made here, it seems that the only problematic part is if the protoc plugin's full path (grpc_csharp_plugin in our case) contains non-ASCII characters. IIUC if other arguments (e.g. the .proto file names) contain non-ascii characters in their path, it isn't   currently a problem (did you verify that?), because the logic for handling UTF-8 file names for other params already exists. Only for the protoc plugin arg, things are different since the invocation of the protoc plugin happens through subprocess.cc

It is just the plugin path that is an issue (these are the paths used in subprocess.cc).  All other paths/names seem to work ok as they use the functions in io_win32.cc which do the utf-8 to wide string conversion correctly.

I've tested with non-ASCII proto_path and non-ASCII .proto file name. Here is one of my test input rsp files:

_--plugin=protoc-gen-grpc=E:\work\grpc\protoctest\test-Dré\tools\grpc_csharp_plugin.exe
--grpc_out=E:\work\grpc\protoctest\test-Dré\grpc_out
--proto_path=E:\work\grpc\protoctest\test-Dré
hélloworld.proto_

> * Based on your investigation, can you tell whether the problem is only specific to when using the argument file (`protoc.exe @test.rsp`) as originally reported or if there are other ways of reproducing it (e.g. invoking protoc.exe with args on command line normally)?

The reported problem was with using an utf-8 encoding in an argument file and this is what I've tested and is what Grpc.Tools uses.

In theory it should be the same with command line arguments as all that reading the argument file does (in ExpandArgumentFile) is push the lines read onto the same vector of command line arguments before the arguments are interpreted.  However this does not seem to work - I think the arguments are getting changed before passing to `main(int argc, char* argv[])`.  I've tried with cmd.exe (with chcp 65001), PowerShell and git bash - all fail to pass the argument correctly.

If we want to fix that too then may need to change to using `wmain(int argc, wchar_t* argv[])` on Windows - but that has bigger impact as wchar_t will need to be passed everywhere.  Maybe just document that if you want to use non-ASCII characters then pass the arguments in a file.

tonydnewell:
As I said above to handle non-ascii arguments on the command line (for Windows) then wmain needs to be used.  The code would need to be something like below. I've not included that in the fix as it is probably outside the scope of the issue being fixed:

In main.cc:
```C++
#ifndef _WIN32
int main(int argc, char* argv[]) {
  return PROTOBUF_NAMESPACE_ID::compiler::ProtobufMain(argc, argv);
}
#else
int wmain( int argc, wchar_t *wargv[ ]) {

  // convert all arguments to utf-8
  char **argv = (char **)malloc(sizeof(char *) * argc);
  for (int i = 0; i < argc; ++i) {
    std::string utf8str;
    if (!google::protobuf::io::win32::strings::wcs_to_utf8(wargv[i], &utf8str)) {
      //GOOGLE_LOG(FATAL) << "wcs_to_utf8: " << Win32ErrorMessage(GetLastError());
    }
    argv[i] = _strdup(utf8str.c_str());
  }

  // call ProtobufMain with the utf-8 arguments
  int sts = PROTOBUF_NAMESPACE_ID::compiler::ProtobufMain(argc, argv);

  // free memory allocated for arguments
  for (int i = 0; i < argc; ++i)
    free(argv[i]);
  free(argv);
  return sts;
}
#endif
```

src/google/protobuf/compiler/subprocess.cc
tonydnewell: Updated the code to use nullptr

src/google/protobuf/compiler/subprocess.cc
```diff
@@ -125,17 +126,30 @@ void Subprocess::Start(const std::string& program, SearchMode search_mode) {
...
+  // get wide string version of command line as the path may contain non-ascii characters
+  std::wstring wcommand_line;
+  if (!io::win32::strings::utf8_to_wcs(command_line.c_str(), &wcommand_line)) {
+    GOOGLE_LOG(FATAL) << "utf8_to_wcs: " << Win32ErrorMessage(GetLastError());
+  }
+
   // Using a malloc'ed string because CreateProcess() can mutate its second
   // parameter.
-  char* command_line =
-      portable_strdup(("cmd.exe /c \"" + program + "\"").c_str());
```
fowles, line 132: There is another call to `portable_strdup` in `SubProccess::Start`.  Does that need to be updated as well?

fowles (approved):

jtattermusch:
Looks like this change didn't make it to protobuf's 21.x release. I'd therefore expect it to become available as part of 22.x release (AFAIKT that release cycle hasn't started yet).

fowles:
If you would like to include this in the next 21.x repease, please send a PR that cherry picks this back

tonydnewell:
@fowles 
> If you would like to include this in the next 21.x repease, please send a PR that cherry picks this back

PR https://github.com/protocolbuffers/protobuf/pull/10721

PR #10202, opened 2022-07-01
Fix maven pom files

deannagarcia:
Add dependencies flag to pom template since java_export doesn't add the tag.

Change the new pom version to match the rest of the repository.

### DIFF
--- a/java/protoc/pom.xml
+++ b/java/protoc/pom.xml
@@ -8,7 +8,7 @@
   </parent>
   <groupId>com.google.protobuf</groupId>
   <artifactId>protoc</artifactId>
-  <version>3.21.2-rc1</version>
+  <version>3.21.2</version>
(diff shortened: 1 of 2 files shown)


java/protoc/pom.xml
```diff
@@ -8,7 +8,7 @@
   </parent>
   <groupId>com.google.protobuf</groupId>
   <artifactId>protoc</artifactId>
-  <version>3.21.2-rc1</version>
+  <version>3.21.2</version>
```
fowles, line 11: seems weird for this to be the specified version on the main branch

deannagarcia: Yeah I left it at that from testing and recently realized I had never changed it back, which is why it's changing now. All of the other files have 3.21.2.

fowles (approved):

PR #10212, opened 2022-07-05
Update rules_kotlin version

Wyverald:


Wyverald:
gentle ping -- this is required to fix protobuf redness on Bazel CI

PR #10213, opened 2022-07-06
Fixed the compilation error of recursive template depth limit exceeding.

junyang-zh:
While we work on real projects in production, here comes a compilation error when field counts tend to be large.

```plain
include/google/protobuf/repeated_field.h:145:1: fatal error: recursive template instantiation exceeded maximum depth of 1024
    PROTO_MEMSWAP_DEF_SIZE(__uint128_t, (1u << 31))
include/google/protobuf/repeated_field.h:135:5: note: expanded from macro 'PROTO_MEMSWAP_DEF_SIZE'
    memswap<kSize - sizeof(reg_type)>(p + sizeof(reg_type), \
[...]
[neutralized].pb.cc:349526:38: note: in instantiation of function template specialization 'google::protobuf::internal::memswap<28136>' requested here
  ::PROTOBUF_NAMESPACE_ID::internal::memswap<
```

toolchain: clang 11

I fixed it by adding a fall back when `kSize` exceeds 512 to avoid the error. Also a unit test is added.


sbenzaquen:
Thanks for the report.
This was fixed with a different approach in commit e8ab276325d9a7598a118022e421cbed581000f6

PR #10220, opened 2022-07-06
Slightly prettier way to express int min

fowles:


mcy (approved):

fowles:
Linux32 bit failure is unrelated

PR #10221, opened 2022-07-06
Adopt FFI

JasonLunn:
Opening this PR in order to get visibility on any potential Kokoro issues introduced by the branch.


JasonLunn:
This PR passes all tests again.

@mkruskal-google for review of changes to Bazel-ization.
@haberman - there are a few TODO messages left in this PR that I'd like to review with you at your convenience.

haberman:
It looks like this PR is proposing to migrate all of JRuby and CRuby to FFI.

We cannot migrate CRuby until we have validated the performance. Do you have performance numbers?

Last we talked my understanding was that this regresses CRuby significantly.

JasonLunn:
It turns out that all performance was significantly negatively impacted by the same issue that was causing memory corruption. Now that has been addressed, I think there is a more compelling case for a general migration. I can post the numbers here or share them with you when I see you later this week.

.bazelrc
```diff
@@ -1 +1,2 @@
-build --cxxopt=-std=c++14
\ No newline at end of file
+build --cxxopt=-std=c++14
+build --action_env=GEM_HOME
```
mkruskal-google, line 2: Ah this is a *really* good idea!  Some suggestions to cover this more thoroughly:
--action_env=PATH --action_env=GEM_PATH --action_env=GEM_HOME

--action_env=DOTNET_CLI_TELEMETRY_OPTOUT=1
--action_env=DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1

The latter are to cover some similar awkwardness around the current state of C#

ruby/BUILD.bazel
```diff
@@ -26,43 +26,31 @@ filegroup(
...
-        ":java_ruby": ["@platforms//:incompatible"],
+        "@platforms//os:linux": ["@platforms//:incompatible"],
         "//conditions:default": ["@platforms//os:osx"],
     }),
 )
 
 internal_ruby_extension(
     name = "protobuf_c",
-    extension = "lib/google/protobuf_c.so",
+    extension = "ext/google/x86_64-linux/libprotobuf_c.so",
```
mkruskal-google, line 49: This will break on other architectures won't it?

ruby/BUILD.bazel
```diff
@@ -116,7 +103,7 @@ inline_sh_test(
     ],
     cmd = """
         pushd `dirname $(location Rakefile)`
-        RUBYLIB=../src:tests:. BAZEL=true rake test
```
mkruskal-google, line 119: Moving this to the kokoro runner doesn't really have the same effect.  If someone were to try to run this manually via Bazel this flag would no longer be set and I think it would break.  The fact that you added it to the kokoro runner just kindof masks that breakage

ruby/BUILD.bazel
JasonLunn: Yes it would, but this is one of only two architectures we use on CI. Per our chat earlier this week, this branch assumes that we're only solving for releasing source gem.

ruby/BUILD.bazel
JasonLunn: Should it go in `.blazerc` as well, then? While this target is unlikely to get much use outside of Kokoro (community Ruby developers are much more likely to just run `rake test` directly) or `travis-test.sh`, I agree we should strive to make it universally runnable.

.bazelrc
JasonLunn: Would you like to see the other places where those environment variables are set migrated here, or do we need them set in multiple places?

ruby/BUILD.bazel
mkruskal-google: Yea that makes sense, I guess my concern is for development and expansion of our cross-architecture CI builds.  We could always add more of these as needed though.

You should definitely mark these as only compatible with the specific architecture tho, so that wildcard bazel expansions don't pick them up on the wrong platform

.bazelrc
mkruskal-google: Yea if we could remove those from our kokoro builds and docker images that would be amazing too!  I think centralizing all of this temporary ugliness here is a better state to be in, and it brings up one step closer to getting `bazel test //...` working :) (see https://github.com/protocolbuffers/protobuf/issues/10427)

ruby/BUILD.bazel
mkruskal-google: Yea that's a good idea

.bazelrc
JasonLunn: Leaving the DOTNET changes out here, can make a follow up PR to address non-Ruby changes.

ruby/BUILD.bazel
JasonLunn: Done

haberman:
Closing this PR in favor of https://github.com/protocolbuffers/protobuf/pull/11483

PR #10222, opened 2022-07-06
Upgrade to Composer 2

deannagarcia:
Composer 1 is now deprecated and we must migrate to use composer 2 instead in our docker files.

### DIFF
--- a/kokoro/linux/dockerfile/test/php_32bit/Dockerfile
+++ b/kokoro/linux/dockerfile/test/php_32bit/Dockerfile
@@ -47,8 +47,10 @@ RUN cd /var/local \
   && make install
 
 # Install composer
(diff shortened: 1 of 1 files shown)


fowles (approved):
We are going to need to backport this to older branches that have the same failure

fowles:
Kokoro failure on this PR:

```
+ docker pull protobuftesting/php_32bit_afb0dc499fbe4de39dc88179628be73f31353bfc
Using default tag: latest
Error response from daemon: pull access denied for protobuftesting/php_32bit_afb0dc499fbe4de39dc88179628be73f31353bfc, repository does not exist or may require 'docker login'
```

deannagarcia:
I have to rebuild and push the docker image, working on that now.

PR #10241, opened 2022-07-11
Add bazel rule to build C# release artifacts

deannagarcia:
The new release script will build both Google.Protobuf and Google.Protobuf.Tools. Installing nuget/dotnet will be done in the docker image and setting up the necessary executable paths for Google.Protobuf.Tools will be done in internal preprocessing.


csharp/build_release.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+cd $(dirname $(readlink $BASH_SOURCE))
+
+#Disable some unwanted dotnet options
```
jskeet, line 5: Tiny nit-pick: I would suggest having a space after the `#` in each comment line, for readability. (Having looked at a couple of shell scripts in the root directory, it looks like that's common in this repo too.)

jskeet (approved):
Just one tiny nit-pick that I've suggested for comments.

csharp/build_release.sh
deannagarcia: Fixed!

csharp/BUILD.bazel
```diff
@@ -51,3 +52,9 @@ pkg_files(
     strip_prefix = strip_prefix.from_root(""),
     visibility = ["//pkg:__pkg__"],
 )
+
+sh_binary(
+  name = "release",
+  srcs = ["build_release.sh"],
+  args = ["$(location build_release.sh)"],
```
perezd, line 59: what is this for? I don't see it used in the script

perezd (approved):
LGTM, Are there already changes in the base docker image for nuget/donet binaries?

deannagarcia:
> LGTM, Are there already changes in the base docker image for nuget/donet binaries?

Yep, they were already added!

csharp/BUILD.bazel
deannagarcia: This is the bazel target to run the bash script so we can just `bazel run //csharp:release`

PR #10244, opened 2022-07-12
Expose plugin protos for C#

jskeet:
The first commit here is just tidying up, effectively, and doesn't require a release note.

The second commit is the important one, exposing the plugin protos under the namespace Google.Protobuf.Compiler. This will be a semi-breaking change (generating a warning) for any project which already generates their own local copy of this proto, but is easily fixed by simply removing the local copy.


amanda-tarafa (approved):
LGTM

fowles (approved):

PR #10245, opened 2022-07-12
Handle long as integer

fjtirado:
Both should be mapped to json number. Right now, for longs, we are
always quouting them, so they are really mapped to json string.


jskeet:
From https://developers.google.com/protocol-buffers/docs/proto3#json for int64, fixed64, uint64:

> JSON value will be a decimal string. Either numbers or strings are accepted.

So it looks like this is behaving as intended and documented. The reason for this is that beyond the range of int32, many integers can't be exactly represented as 64-bit IEEE-754 decimal values.

fjtirado:
> From https://developers.google.com/protocol-buffers/docs/proto3#json for int64, fixed64, uint64:
> 
> > JSON value will be a decimal string. Either numbers or strings are accepted.
> 
> So it looks like this is behaving as intended and documented. The reason for this is that beyond the range of int32, many integers can't be exactly represented as 64-bit IEEE-754 decimal values.
@jskeet  Good, thanks for the explantion (somehow I suspected this was intentional, but I was unable to find out the reason) Closing this.

PR #10243, opened 2022-07-11
Adding Github Action to keep CMake generated file lists up to date

mkruskal-google:

PR #10254, opened 2022-07-14

haberman:


build_files_updated_unittest.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
acozzette, line 1: I think something went wrong with the merge here because in #10243, this file was recently deleted and cmake/update_file_lists.sh was added.

PR #10256, opened 2022-07-15
Update C#-specific README

jskeet:
In particular, this removes the now-inaccurate instructions for building
Google.Protobuf targeting .NET 3.5.


jskeet:
@fowles: If you could merge after review (assuming you're happy) that would be great.

fowles (approved):

PR #10259, opened 2022-07-16
Remove final modifier on private methods. It's useless.

Okapist:


Okapist:
Just small change before big change. Not to be first time committer with all it's problems.

deannagarcia (approved):
Thanks for this change and thinking ahead for future changes :)

Issue #8101, opened 2020-11-30
C#: .proto package name that starts with an underscore followed by a number generates C# code that doesn't compile


gauravsaxenais:
Currently I have a package name like this:

`package _9003aaa09a0c47fc8e24e738499835ef;`

This generates a namespace in C# like `9003aaa09a0c47fc8e24e738499835ef;`

which creates a compile time error. I believe the correct namespace should be `_9003aaa09a0c47fc8e24e738499835ef;`

akjain2052:
Hi, I am not able to reproduce this issue using v3.14.0 and v3.6.0 protoc compilers , checked with java and C# code. Please let me know how to reproduce this issue.

akjain2052:
I have tried this issue with v3.14.0, v3.6.0, v3.5.0 protoc compiler in windows and linux environment. 
I have not observed the issue. Please write the steps to reproduce the issue, and environment details.

**xyz.proto file** 
option csharp_namespace = "_9003aaa09a0c47fc8e24e738499835ef";

**Xyz.cs generated file in C#**
namespace _9003aaa09a0c47fc8e24e738499835ef {

haberman:
I can reproduce this if I use a `package` directive instead of `option csharp_namespace=`.

A workaround would be to use `option csharp_namespace=`.

@jtattermusch any ideas?

jtattermusch:
I haven't tried to reproduce, but this is how the "namespace" is generated in C# codegen:
https://github.com/protocolbuffers/protobuf/blob/de5d1b98c27428450b9a38ab5c2de479f59025af/src/google/protobuf/compiler/csharp/csharp_helpers.cc#L118

The source code also confirms the workaround to set `option csharp_namespace=`. This is probably a bug, but it seems to be quite an odd case (package name starting with and underscore followed by a number) so not a high priority to fix.

Note that the GetFileNamespace function also gets used by gRPC C# codegen:
https://github.com/grpc/grpc/blob/86f93801e5d452cff40d6958d21a8eb211da0654/src/compiler/csharp_generator.cc#L808

fowles:
Jan, is this still an open issue?

tonydnewell:
@jtattermusch I'll add it to my list to look at

tonydnewell:
[...]

PR #10268, opened 2022-07-19
Bugfix/issue 8101

tonydnewell:
Fix for https://github.com/protocolbuffers/protobuf/issues/8101

When converting identifiers for C# camel case, handle the case where the input starts with "_&lt;digit&gt;" and preserve the underscore.

It does not alter existing behaviour for other identifiers:
- e.g. starts with "_&lt;alpha&gt;" - the underscore is consumed, so as to not break anything that relies on existing behaviour.
- e.g. input is already invalid (e.g. identifier starts with a digit) -  it does not try and fix it.

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_generator_unittest.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_generator_unittest.cc
@@ -74,6 +74,22 @@ TEST(DescriptorProtoHelpers, IsDescriptorOptionMessage) {
   EXPECT_FALSE(IsDescriptorOptionMessage(DescriptorProto::descriptor()));
 }
 
+TEST(CSharpIdentifiers, UnderscoresToCamelCase) {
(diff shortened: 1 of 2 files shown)


jskeet (approved):

jskeet (approved):

fowles (approved):

jtattermusch (commented):
LGTM as well.

nickwinger:
@tonydnewell  thank you for this fix, but on our side there is still a similar bug open with this, e.g.: namespace "my.space._3" gets compiled to c# "my.space.3" which again causes an error. So also with "sub" namespaces after dots the numbers with an underscore should be preserved. thank you

tonydnewell:
@nickwinger Thank you for your observation. I can see you've created a new issue https://github.com/protocolbuffers/protobuf/issues/13482
I can't guarantee when it will be fixed but it is on the list.

nickwinger:
here is the PR: https://github.com/protocolbuffers/protobuf/pull/13504

PR #10269, opened 2022-07-19
Disambiguate generated properties in C#

jskeet:


src/google/protobuf/compiler/csharp/csharp_helpers.cc
```diff
@@ -381,13 +381,22 @@ std::string GetFieldConstantName(const FieldDescriptor* field) {
...
+      || property_name == "Equals"
+      || property_name == "ToString"
+      || property_name == "GetHashCode"
+      || property_name == "WriteTo"
+      || property_name == "Clone"
+      || property_name == "CalculateSize"
+      || property_name == "Descriptor"
+      || property_name == "MergeFrom"
+      || property_name == "OnConstruction"
+      || property_name == "Parser") {
```
fowles, line 399: perhaps we have gotten to the point where a hash set or makes sense?

fowles (approved):

src/google/protobuf/compiler/csharp/csharp_helpers.cc
jskeet: I would have done that, but my C++ skills are incredibly poor :) If you have an example elsewhere in the codebase that I could refer to, that would be really handy.

fowles:
```
std::unordered_set<std::string> reserved_names = {
  descriptor->containing_type()->name(),
  "Types",
  ...
};
```

fowles:
```
if (reserved_names->find(name) != reserved_names->end()) {
  ....
}
```

fowles: the invocation for a static one is:

```
static const auto& reserved_names = *new std::unordered_set<std::string>({"hi", "there"});
```

jskeet: Many thanks. I'll give each a try tomorrow, and ping the PR when I've come up with something. Thanks for putting up with my C++ ignorance :)

amanda-tarafa (approved):
LGTM

src/google/protobuf/compiler/csharp/csharp_helpers.cc
```diff
@@ -381,13 +381,22 @@ std::string GetFieldConstantName(const FieldDescriptor* field) {
...
-  // Avoid either our own type name or reserved names. Note that not all names
-  // are reserved - a field called to_string, write_to etc would still cause a problem.
+  // Avoid either our own type name or reserved names.
   // There are various ways of ending up with naming collisions, but we try to avoid obvious
-  // ones.
+  // ones. In particular, we avoid the names of all the members we generate or inherit.
   if (property_name == descriptor->containing_type()->name()
       || property_name == "Types"
-      || property_name == "Descriptor") {
+      || property_name == "Descriptor"
```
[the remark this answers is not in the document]
jskeet: Yes, I forgot GetType and MemberwiseClone for some reason. Will add those tomorrow when I move to a set-based approach.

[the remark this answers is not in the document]
jskeet: It turns out GetType and MemberwiseClone are slightly trickier - they currently only generate warnings, not errors, so adding an underscore *could* be a breaking change in that case. I'm not sure of the best solution for that at the moment. (I'm tempted to go really cautiously here. We haven't heard any reports of those causing problems, so I think I'd prefer to leave them for now.)

src/google/protobuf/compiler/csharp/csharp_helpers.cc
jskeet: Sorted now, thanks.

jskeet:
Changes since preview review:

- Added more members to the test proto
- Used an unordered_set in the implementation

src/google/protobuf/compiler/csharp/csharp_helpers.cc
```diff
@@ -378,16 +378,32 @@ std::string GetFieldConstantName(const FieldDescriptor* field) {
   return GetPropertyName(field) + "FieldNumber";
 }
 
+// Names of members declared or overridden in the message.
+// Used in GetPropertyName below.
+static const auto& reserved_member_names = *new std::unordered_set<std::string>({
```
fowles, line 383: unfortunately, you are doing this at the global scope which leads to issues.  You need to do this at function scope

jskeet: Fixed, I think - but please check carefully! I'm hoping that by declaring it with the `static` modifier it's still only going to be initialized on first use rather than on every call, right?

fowles (approved):

src/google/protobuf/compiler/csharp/csharp_helpers.cc
fowles: correct

fowles:
I agree, but one can hope :)

jskeet:
Looks like it's now just "all the MacOS tests" that are failing - I don't know enough about MacOS to help on that front :(

jskeet:
@fowles: All green now - please merge if you're happy to do so :)

PR #10276, opened 2022-07-20
Update comment generator to apply to maps

dmaclach:
Change comment from "array" to "container" because sometimes it's a map.


fowles (approved):

thomasvl (commented):
LG

PR #10281, opened 2022-07-20
Ruby: Use class inheritance to save memory

casperisfine:
The only difference between all the `Message` classes is just their `descriptor` instance variable.

So rather than create an entirely new class from scratch every time we can simply inherit from an abstract class.

This shink each `Message` class from `1456` bytes to `944` bytes, and the singleton class of each from `960` to `792`, for a total of `680` bytes saved per message class, so a ~28% reduction.

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -37,6 +37,7 @@
 #include "repeated_field.h"
 
 static VALUE cParseError = Qnil;
+static VALUE cAbstractMessage = Qnil;
 static ID descriptor_instancevar_interned;
 
 static VALUE initialize_rb_class_with_no_args(VALUE klass) {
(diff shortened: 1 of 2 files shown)


haberman:
I wish there was a way to do this that didn't affect the public API.

If we do this, developers can start writing:

```ruby
if obj.kind_of? Google::Protobuf::AbstractMessage
  # This is a protobuf
end
```

That might not be the worst thing, but it would commit us to maintaining this inheritance hierarchy.

casperisfine:
I can make that constant private.

casperisfine:
Done: 

```ruby
>> require 'google/protobuf'
=> true
>> Google::Protobuf::AbstractMessage
(irb):2:in `<main>': private constant Google::Protobuf::AbstractMessage referenced (NameError)
```

haberman (approved):

PR #10282, opened 2022-07-20
Ruby: use a valid instance variable name for `descriptor`

casperisfine:
Setting an instance variable without an `@` prefix makes it unacessible from the Ruby side, which prevent introspection by various tools.

I see no reason to prevent this.


haberman:
We want to reduce the public API surface so that we have the freedom to change the internals.

That said, I think it's generally understood that `@`-prefixed instance variables are private and not part of the API contract. So this seems ok to me.

haberman (approved):

casperisfine:
Thanks!

PR #10290, opened 2022-07-21
Upgrade to com.google:google:5 for protoc artifact

liutikas:
Protoc depeneds on com.google:google:1 which is unsigned. Which means users
the validate signatures have to add it to an allow list. Upgrading makes it no longer
necessary.

Compare:
https://repo1.maven.org/maven2/com/google/google/1
vs
https://repo1.maven.org/maven2/com/google/google/5


deannagarcia (approved):
Thanks for the update!

Issue #10280, opened 2022-07-20
Ruby: extension types should define a proper `memsize` function


casperisfine:
Here's an snippet from a Ruby heap dump (generated with `ObjectSpace.dump_all`

```
{"address":"0x7fa2dadd7320", "type":"DATA", "class":"0x7fa3eb7ab408", "struct":"Google::Protobuf::Internal::Arena", "memsize":40}
{"address":"0x7fa2dadd1678", "type":"DATA", "class":"0x7fa3eb7ab408", "struct":"Google::Protobuf::Internal::Arena", "memsize":40}
{"address":"0x7fa2dadd3c70", "type":"DATA", "class":"0x7fa3eb7ab408", "struct":"Google::Protobuf::Internal::Arena", "memsize":40}
{"address":"0x7fa2dadcfaa8", "type":"DATA", "class":"0x7fa3eb7ab408", "struct":"Google::Protobuf::Internal::Arena", "memsize":40}
{"address":"0x7fa2dadc9c98", "type":"DATA", "class":"0x7fa3eb7ab408", "struct":"Google::Protobuf::Internal::Arena", "memsize":40}
{"address":"0x7fa2dadcb390", "type":"DATA", "class":"0x7fa3eb7ab408", "struct":"Google::Protobuf::Internal::Arena", "memsize":40}
```

As you can see, all these `Google::Protobuf::Internal::Arena` instances claim to be 40 bytes, which is obviously a lie.

This is because the `memsize` function is not defined: https://github.com/protocolbuffers/protobuf/blob/520c601c99012101c816b6ccc89e8d6fc28fdbb8/ruby/ext/google/protobuf_c/protobuf.c#L192

**What language does this apply to?**

Ruby

**Describe the problem you are trying to solve.**

I'd like to have a semi-accurate idea of how much memory the `google-protobuf` gem is holding onto.

**Describe the solution you'd like**

I'd like these functions to be defined (`Arena`, `Message`, etc). They don't have to be 100% accurate, not particularly fast. It's a debugging facility.

**Describe alternatives you've considered**

I don't see any.

cc @zhangskz because https://github.com/protocolbuffers/protobuf/issues/9384#issuecomment-1189950068

zhangskz:
This makes sense and would be helpful as a whole for debugging and understanding memory usage.

[...]

PR #10291, opened 2022-07-21
Ruby implement memsize functions for native types

casperisfine:
Fix: https://github.com/protocolbuffers/protobuf/issues/10280

This allows Ruby to report a more correct estimation of the memory used by these objects.

It's useful when running memory profilers against applications.

cc @zhangskz @haberman

### DIFF
--- a/ruby/ext/google/protobuf_c/map.c
+++ b/ruby/ext/google/protobuf_c/map.c
@@ -61,9 +61,13 @@ static void Map_mark(void* _self) {
   rb_gc_mark(self->arena);
 }
 
+static size_t Map_memsize(const void* _self) {
+  return sizeof(Map);
+}
+
 const rb_data_type_t Map_type = {
     "Google::Protobuf::Map",
-    {Map_mark, RUBY_DEFAULT_FREE, NULL},
+    {Map_mark, RUBY_DEFAULT_FREE, Map_memsize},
     .flags = RUBY_TYPED_FREE_IMMEDIATELY,
 };
 
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -66,9 +66,13 @@ static void Message_mark(void* _self) {
   rb_gc_mark(self->arena);
 }
 
+static size_t Message_memsize(const void* _self) {
+  return sizeof(Message);
+}
+
 static rb_data_type_t Message_type = {
     "Google::Protobuf::Message",
-    {Message_mark, RUBY_DEFAULT_FREE, NULL},
+    {Message_mark, RUBY_DEFAULT_FREE, Message_memsize},
     .flags = RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
--- a/ruby/ext/google/protobuf_c/protobuf.c
+++ b/ruby/ext/google/protobuf_c/protobuf.c
@@ -187,11 +187,22 @@ static void Arena_free(void *data) {
   xfree(arena);
 }
 
+static size_t Arena_memsize(const void *data) {
+  const Arena *arena = data;
+  size_t memsize = upb_Arena_SpaceAllocated(arena->arena);
+  if (arena->arena->refcount > 1) {
+      // If other arena were fused we attribute an equal
+      // share of memory usage to each one.
+      memsize /= arena->arena->refcount;
+  }
+  return memsize + sizeof(Arena);
+}
+
 static VALUE cArena;
 
 const rb_data_type_t Arena_type = {
     "Google::Protobuf::Internal::Arena",
-    {Arena_mark, Arena_free, NULL},
+    {Arena_mark, Arena_free, Arena_memsize},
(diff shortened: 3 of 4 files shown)


ruby/ext/google/protobuf_c/map.c
```diff
@@ -61,9 +61,13 @@ static void Map_mark(void* _self) {
   rb_gc_mark(self->arena);
 }
 
+static size_t Map_memsize(void* _self) {
+  return sizeof(Map);
```
casperisfine, line 65: Here I assume that `const upb_Map* map;` is pointing into the `Arena` right?

ruby/ext/google/protobuf_c/message.c
```diff
@@ -63,9 +63,13 @@ static void Message_mark(void* _self) {
   rb_gc_mark(self->arena);
 }
 
+static size_t Message_memsize(void* _self) {
+  return sizeof(Message);
```
casperisfine, line 67: Same here, I'm assuming:

```
  const upb_Message* msg;  // Can get as mutable when non-frozen.
  const upb_MessageDef*
```

Are both pointers into Arena.

ruby/ext/google/protobuf_c/protobuf.c
```diff
@@ -185,11 +185,16 @@ static void Arena_free(void *data) {
   xfree(arena);
 }
 
+static size_t Arena_memsize(void *data) {
+  Arena *arena = data;
+  return sizeof(Arena) + _upb_ArenaHas(arena->arena);
```
casperisfine, line 190: I don't know if `_upb_ArenaHas` is the correct function to get the arena memory footprint.

ruby/ext/google/protobuf_c/map.c
haberman: That is correct.

ruby/ext/google/protobuf_c/message.c
haberman: The message, yes. The MessageDef is a pointer into the `upb_DefPool` (underlying the `DescriptorPool` class).

ruby/ext/google/protobuf_c/protobuf.c
haberman: It is not, `_upb_ArenaHas()` tells you how much space is currently *available* to allocate on the arena.

`upb_Arena` does not currently have a function for reporting how much memory has been allocated. It would be relatively simple to add one, but there is a complication when the arenas are fused:

```ruby
  # Creating a new message creates an associated arena.
  foo_msg = FooMessage()

  # New message, new arena.
  bar_msg = BarMessage()

  # Joining the two messages "fuses" their arenas.
  # The two arenas now share the same underlying memory.
  # If both arenas reported the full allocated size, we would double-count.
  foo_msg.bar = bar_msg
```

haberman: Each arena knows its refcount (the number of arenas that have been fused into it).

So perhaps the simplest solution would be to have:

```c
// Returns the total space allocated by the arena, divided
// by the number of live arenas that have been fused into this one.
size_t upb_Arena_SpaceAllocated(const upb_Arena*)
```

casperisfine: Hum, I need to dig a bit more on how these Arena work. When you fuse an Arena, into another, it stays alive?

ruby/ext/google/protobuf_c/message.c
casperisfine: Right, so I shouldn't consider it either since it should be reported by the `DescriptorPool` `memsize` function.

casperisfine:
@haberman thanks. I updated `Arena_memsize` according to your suggestions.

haberman:
Thanks for the changes. Would you be able to submit the ruby-upb.c/.h changes to https://github.com/protocolbuffers/upb/blob/main/upb/arena.h and https://github.com/protocolbuffers/upb/blob/main/upb/arena.c, which is that code's primary location.

We can create a new function `upb_Arena_SpaceAllocated()` that is then called from here.

casperisfine:
> . Would you be able to submit the ruby-upb.c/.h changes to ...

Done! https://github.com/protocolbuffers/upb/pull/729

haberman:
If you rebase on main it should pick up the `upb_Arena_SpaceAllocated()` function that is there now.

casperisfine:
Rebeased.

haberman:
I think we need one more rebase here (I think this conflicted with your other change).

casperisfine:
Oh wow, I had forgotten about this one. I'll rebase.

casperisfine:
Done.

haberman:
The Ruby 2.6 test failed with:

> ruby/ext/google/protobuf_c/protobuf.c:417: void ObjectCache_Add(const void *, VALUE): Assertion `ObjectCache_Get(key) == val' failed.
Aborted (core dumped)

It's hard to see how this PR would have caused that error. I'll try re-running that test.

haberman (approved):

PR #10293, opened 2022-07-21
CMake: use add_compile_options instead of add_definitions for compile options

Neumann-A:
fixes: 
- using CMake hacks like unnecessary setting `CMAKE_RC_COMPILE_OBJECT`


acozzette (approved):

acozzette:
@Neumann-A Thanks!

PR #10298, opened 2022-07-21
Reduce the required alignment of ArenaString from 8 to 4

yashykt:
The current 21.3 release breaks gRPC android due to this alignment requirement. Verified that reducing it to 4, makes it work again.

### DIFF
--- a/src/google/protobuf/arenastring.cc
+++ b/src/google/protobuf/arenastring.cc
@@ -50,7 +50,8 @@ namespace internal {
 
 namespace  {
 
-// Enforce that allocated data aligns to at least 8 bytes, and that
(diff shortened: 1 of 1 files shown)


zhangskz (approved):

zhangskz (changes requested):
Unapproving for now, though this looks good.

We may opt to instead sync this to github + cp to release branch from the internal change to avoid future sync conflict. However, the conflict isn't too bad anyways so I think this will depend on if we need to sync other changes for the next patch release anyways.

zhangskz (approved):

zhangskz:
Re-approving to get this into 21.x for 21.4 release.

PR #10188, opened 2022-06-29
Add explicit abort() at the end of AlignFail

aladram:
GCC cannot infer that `LogMessage::Finish()` (called indirectly by `GOOGLE_LOG(FAIL)` in `AlignFail`) exits program when `level_ == LOGLEVEL_FATAL`. Compilation will create a warning in projects importing the library for debug builds. The attribute cannot be individually assigned to `LogMessage::Finish()` because it does return conditionally of the log level. Tried with GCC 9.3.0 and GCC 12.1.0.

This change adds an explicit abort() for compiler to clearly understand this does not return, and enables `[[noreturn]]` for MSVC as well.

Closes #9758.

PR #10304, opened 2022-07-22
Add [[noreturn]] to AlignFail

Sil3ntStorm:
This was removed in #10188 however clang emits `-Wmissing-noreturn`
now, as that function cannot ever return due to the explicit
abort call that was added. Adding the `[[noreturn]]` will prevent
this warning

See https://godbolt.org/z/ef9E1G9ah


src/google/protobuf/generated_message_tctable_impl.h
```diff
@@ -269,7 +269,7 @@ enum FieldType : uint16_t {
 
 #ifndef NDEBUG
 template <size_t align>
-void AlignFail(uintptr_t address) {
+[[noreturn]] void AlignFail(uintptr_t address) {
```
fowles, line 272: I don't think all the compilers in our toolchain support a raw [[noreturn]] atribute.  I think you need to do something like

https://github.com/abseil/abseil-cpp/blob/master/absl/base/attributes.h#L195-L204

in port_def.inc and port_undef.inc

fowles (changes requested):

src/google/protobuf/generated_message_tctable_impl.h
Sil3ntStorm: Well it was used for all but msvc before it was removed, so that tells me that every compiler in your suite supported it, except maybe an ancient MSVC version?

fowles: I would feel more comfortable using the same attributes as we do for abseil.  Protobuf is in the (slow) process of rationalizing all its supported platforms and standardizing on the same ones as protobuf and grpc.

Sil3ntStorm: I don't seem to be able to find any information on when Microsoft started to support `[[noreturn]]` in the standard form. It has been in C++11 which means it is likely supported since VS 2017. However I don't have that available to test and the earliest one in compiler explorer is 19.14, which supports it.
Since MSVC doesn't complain about it anyway (and does not seem to generate any different code on older versions), we could just go ahead and only have the attribute on newer MSVC versions.

fowles: I don't recall any bugs filed against absl.  Feels like we can use there version with risk (the licenses are compatible)

src/google/protobuf/port_def.inc
```diff
@@ -413,6 +413,26 @@
 #define PROTOBUF_ATTRIBUTE_REINITIALIZES
 #endif
 
+#ifdef PROTOBUF_NO_RETURN
+#error PROTOBUF_NO_RETURN was previously defined
+#endif
+#ifdef __cpp_has_attribute
```
fowles, line 419: the prelud of pord_def.inc (around linke 55) allows you to skip a lot of this and implement this as

#if __has_cpp_attribute(noreturn)
#elfi ...

src/google/protobuf/port_def.inc
```diff
@@ -413,6 +413,26 @@
...
+#if __cpp_has_attribute(noreturn)
+#define PROTOBUF_NO_RETURN [[noreturn]]
+#else
+#define PROTOBUF_NO_RETURN
+#endif
+#else
+#if defined(__GNUC__) || defined(__clang__)
```
fowles, line 426: here you are probably want to use PROTOBUF_GNUC_MIN(x, y) with whatever version this got added in

src/google/protobuf/port_def.inc
```diff
@@ -413,6 +413,26 @@
...
+#define PROTOBUF_NO_RETURN __attribute__((noreturn))
+#elif defined(_MSC_VER)
```
fowles, line 428: this would be PROTOBUF_MSC_VER_MIN

fowles (changes requested):

src/google/protobuf/port_def.inc
Sil3ntStorm: There does not seem to be any information as to when that became a supported thing. So I would just go with "all of them" and leave it as is.

fowles (approved):

fowles:
Thanks for PR!  I will merge this when the CI finishes.

fowles:
Something is going wrong with our CI.  It will likely not be until Monday when we have a chance to investigate

Sil3ntStorm:
Have you had any chance to work out whatever it is that went wrong with your CI yet?

fowles:
sorry, I lost track of this.  Kokoro has recovered from its issues and is running now.

fowles:
Several builds are failing with

```
bazel-out/k8-opt-exec-2B5CBBC6/bin/src/google/protobuf/_virtual_includes/port_def/google/protobuf/port_def.inc:414:5: error: function-like macro '__cpp_has_attribute' is not defined
```

Looks like you should stub this in the same way that `__has_attribute` is stubbed at the top of the file.

deannagarcia:
All of the builds are now failing with the error Matt posted about earlier. I'm going to close this PR for now, but if you want to fix this error please reopen.

PR #10314, opened 2022-07-26
Add CMake Ninja test

mkruskal-google:
This PR also does some opportunistic cleanup of the CMake build.  Notably:
- ctest should be used for running tests, which runs everything registered with add_test
- newer docker image is used to allow more flexibility in cmake configuration
- .gitignore file updated to include CMake cruft

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,23 @@ m4/ltversion.m4
 m4/lt~obsolete.m4
 autom4te.cache
 
+# CMake-generated files
+.ninja_deps
+.ninja_logs
+cmake/protobuf/*.cmake
+cmake_install.cmake
+CMakeCache.txt
+CTestTestfile.cmake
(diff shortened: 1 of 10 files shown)


acozzette (approved):

mudream4869:
Not sure if I can ask question or not here...
I found that document of cmake changed in this PR:
From (v3.21.12)
> This directory contains CMake files that can be used to build protobuf with MSVC on Windows. You can build the project from Command Prompt and using an Visual Studio IDE.

To

> This directory contains CMake files that can be used to build protobuf.

Does it mean: in v3.21.12 version. The protobuf is still NOT ensure/mean cmake can be used in Linux, through it may be compiled and tested ok?

mkruskal-google:
No the opposite, we support cmake on all operating systems.  That's why I removed some of the windows-specific language

Issue #6218, opened 2019-06-06
Do not include .pdb files to nuget package, create separate .symbols.nupkg package instead


HeikeHofmann:
Consider to NuGet documentation
http://docs.nuget.org/docs/creating-packages/creating-and-publishing-a-symbol-package

Instead of including .pdb files to .nupkg directly, use -Source nuget option to generate special type of package - .symbols.nupkg package that keep pbd files and sources.
 Publish main package to nuget.org and symbols package to symbolsource.org - this can be arrived by one command.

Version 3.6.1 was ok, newer versions (3.7.0 & 3.8.0) are a problem for companies, which take your packages and upload them to validated tools feeds. Then .pdb files will be automatically deleted and signature isn't valid anymore.

See Issue #5785, where this information can be found at the end of the discussion.

jskeet:
Note that instead of a .symbols.nupkg package, we'd probably want to create a .snupkg package: https://docs.microsoft.com/en-gb/nuget/create-packages/symbol-packages-snupkg

NN---:
Is there any update here ?
SourceLink with snupkg makes development much easier.

PR #10319, opened 2022-07-26
Adding snupkg support

deannagarcia:
Create separate .symbols.nupkg package for debugging C# code instead of .pdb files.

Requested in #6218

### DIFF
--- a/csharp/src/Google.Protobuf/Google.Protobuf.csproj
+++ b/csharp/src/Google.Protobuf/Google.Protobuf.csproj
@@ -19,9 +19,8 @@
     <RepositoryUrl>https://github.com/protocolbuffers/protobuf.git</RepositoryUrl>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <!-- Include PDB in the built .nupkg -->
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <IsTrimmable>true</IsTrimmable>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">


jskeet:
I'm on holiday until Monday so don't have a chance to review until then - although personally I tend to prefer having everything in a single package file.

jskeet:
There are definitely pros and cons here. Personally I still prefer having the PDB files in the package - that way all the information you need for debugging is right there, without any connection needed to talk to Microsoft servers. The scenario of company package repositories unpacking and then repacking a package is one I hadn't come across before though. (And my immediate reaction is "ewww" - but that doesn't help developers stuck in that situation, of course.)

It does look like snupkg files are more widely used though, so maybe we should do this. @jtattermusch noted a preference for retaining PDB files before - Jan, is your opinion the same as it was, or has working on Grpc.Net.Client changed your mind?

deannagarcia:
I will definitely go with your opinions here, but thought it was worth at least having the conversation based on requests

jtattermusch:
> There are definitely pros and cons here. Personally I still prefer having the PDB files in the package - that way all the information you need for debugging is right there, without any connection needed to talk to Microsoft servers. The scenario of company package repositories unpacking and then repacking a package is one I hadn't come across before though. (And my immediate reaction is "ewww" - but that doesn't help developers stuck in that situation, of course.)
> 
> It does look like snupkg files are more widely used though, so maybe we should do this. @jtattermusch noted a preference for retaining PDB files before - Jan, is your opinion the same as it was, or has working on Grpc.Net.Client changed your mind?

My opinion hasn't changed.

jskeet:
> My opinion hasn't changed.

In that case, I'd suggest closing this PR and adding a note to #6218. I wouldn't close that issue though - I think it's worth keeping open for occasional re-evaluation, partly to see whether we get additional comments from other customers.

googleberg:
@deannagarcia should we close this?

jskeet:
Not yet. It looks like there are currently some issues debugging into nuget packages. I haven't had time to get to the bottom of it yet, but we may end up wanting to do this after all.

jskeet:
> Not yet. It looks like there are currently some issues debugging into nuget packages. I haven't had time to get to the bottom of it yet, but we may end up wanting to do this after all.

I looked into this for other libraries, and there were some Visual Studio / build issues that were apparently causing the problem.

They're fairly simple to work around in most cases though, and I'd still rather not publish the snupkg files. See https://cloud.google.com/dotnet/docs/reference/help/troubleshooting#how-can-i-debug-into-the-libraries for the workaround - and with that in place, I'm going to close this PR.

PR #10321, opened 2022-07-27
Fix Timestamps fromDate for negative 'exact second' java.sql.Timestamps

q-nathangrand:
(-1000/1000) -1 = -2s // This is what the code used to incorrectly do
(-999/1000) - 1 = -1s // This is correct


deannagarcia (approved):
Thanks for the fix and the added test!!

PR #10326, opened 2022-07-27
Remove distcheck and dist_install Kokoro tests

acozzette:
These tests verify that various things can be built/installed/tested
using our published distribution archives, but starting with 22.0 we
will no longer be publishing those.


mkruskal-google (approved):
Thanks Adam!

PR #10355, opened 2022-08-03
Use release version instead of libtool version in Makefile

### LINKED ISSUE #10323: Release 21.3 broke ABI, ironically because of #10271
Version: v21.2 -> v21.3
Language: n/a

openSUSE Tumbleweed

We updated protobuf from 21.2 to 21.3 (21.4 should be affected the same way) and multiple applications/libraries stopped working due to this error:

```
/usr/sbin/usbguard-daemon: symbol lookup error: /lib64/libusbguard.so.1: undefined symbol: _ZN6google8protobuf8internal16InternalMetadataD1Ev
```

It appears like as a side effect of #10271, the symbol is now completely missing, but libraries/applications build against the previous version still depend on it. This requires a SOVER bump.

CC @pitrou @fowles @coryan

deannagarcia:
Fixes #10323

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -18,7 +18,7 @@ else
 PTHREAD_DEF =
 endif
 
-PROTOBUF_VERSION = 32:4:0
+PROTOBUF_VERSION = 3.21.4
 
 if GCC
 # Turn on all warnings except for sign comparison (we ignore sign comparison
@@ -189,7 +189,7 @@ nobase_include_HEADERS =                                         \
 lib_LTLIBRARIES = libprotobuf-lite.la libprotobuf.la libprotoc.la
 
 libprotobuf_lite_la_LIBADD = $(PTHREAD_LIBS) $(LIBATOMIC_LIBS)
-libprotobuf_lite_la_LDFLAGS = -version-info $(PROTOBUF_VERSION) -export-dynamic -no-undefined
+libprotobuf_lite_la_LDFLAGS = -release $(PROTOBUF_VERSION) -export-dynamic -no-undefined
 if HAVE_LD_VERSION_SCRIPT
 libprotobuf_lite_la_LDFLAGS += -Wl,--version-script=$(srcdir)/libprotobuf-lite.map
 EXTRA_libprotobuf_lite_la_DEPENDENCIES = libprotobuf-lite.map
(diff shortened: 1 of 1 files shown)


fowles (approved):

src/Makefile.am
```diff
@@ -18,7 +18,7 @@ else
 PTHREAD_DEF =
 endif
 
-PROTOBUF_VERSION = 32:4:0
+PROTOBUF_VERSION = 3.21.4
```
fowles, line 21: do we need to udpate the update_version.py to hit this line?

deannagarcia: Yep but needs to happen internally, I have filed a bug and CL will come soon.

acozzette:
I think we also need to make the corresponding change for CMake. libprotobuf.cmake, libprotobuf-lite.cmake, and libprotoc.cmake all set `SOVERSION` and it looks like we could probably update them to set `VERSION` instead: https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html

deannagarcia:
> I think we also need to make the corresponding change for CMake. libprotobuf.cmake, libprotobuf-lite.cmake, and libprotoc.cmake all set `SOVERSION` and it looks like we could probably update them to set `VERSION` instead: https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html

I made this change but I'm going to hold on submitting until we talk about whether protoc should share cpp's major version or have its own.

acozzette (approved):

src/Makefile.am
jengelh: You can just use `-release ${PACKAGE_VERSION}` at which point only configure.ac needs updating.

deannagarcia: Where is PACKAGE_VERSION set?

jengelh: Either of the two answers should help:
```
AC_INIT([Protocol Buffers],[3.21.4],[protobuf@googlegroups.com],[protobuf])
                            ^^^^^^
```
or
```
grep -r PACKAGE_VERSION /usr/share/aut*
/usr/share/autoconf/autoconf/general.m4:AC_SUBST([PACKAGE_VERSION],
```
it's implied.

PR #10360, opened 2022-08-04

fowles:


deannagarcia (approved):

ilatypov:
Why did the previous versions pass all checks?  My (possibly incorrect) compile test of [repeated_field.h v3.20.0](https://github.com/protocolbuffers/protobuf/blob/v3.20.0/src/google/protobuf/repeated_field.h) shows absence of the memswap() definitions for non-regular block sizes that should terminate the compiler's recursive look-ups after reducing the requested block size by the largest regular sizes such as 8, 4 and 2.

ilatypov:
Never mind, the old code did not aim at resolving for block sizes greater than 7 (or 15 for 128-bit compilers).  The `enable_if` condition does not accept the bigger numbers.  For blocks of size 7 and smaller, the hierarchical loop unrolling was resolved based on the present block sizes.

PR #10368, opened 2022-08-05
Fix rule for java target in examples/Makefile

kmizumar:
The steps described in the README.md won't work:

```
### Java
 
Follow instructions in [../README.md](../README.md) to install protoc and then
download protobuf Java runtime .jar file from maven:
 
    https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java
 
Then run the following:
 
    $ export CLASSPATH=/path/to/protobuf-java-[version].jar
    $ make java
 
This will create the add_person_java/list_people_java executables (shell
scripts) and can be used to create/display an address book data file.
```

Change the rules to catch up with the package structure change.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -13,13 +13,14 @@ python: add_person_python list_people_python
 
 clean:
 	rm -f add_person_cpp list_people_cpp add_person_java list_people_java add_person_python list_people_python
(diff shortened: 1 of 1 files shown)


deannagarcia (approved):
Thank you for this fix!

PR #10375, opened 2022-08-08
Add extension for Oclea Service Layer RPCs

samsta:
For the service layer we're currently building for our products we need to add some custom extensions. This will be an SDK that customers can use and extend, so it would make sense to reduce the risk of conflicting IDs by allocating one for our purposes.

Could this change be applied, please, in order to reserve our extension ID?

Thanks,
Sam


deannagarcia (approved):

samsta:
Thank you, @deannagarcia !

PR #10378, opened 2022-08-09
PYI generator: print values as fields of enum

dbeinder:
When generating PYI for python, enums don't have any fields and your IDE shows errors when referring to a value of an enum using `MyEnum.Foo`

Example:
```protobuf
syntax = "proto3";
enum MyEnum {
	Abc = 0;
	Foo = 1;
	Bar = 2;
};
```
`$> protoc --pyi_out=. myenum.proto`

Current PYI Output:
```python
from google.protobuf.internal import enum_type_wrapper as _enum_type_wrapper
from google.protobuf import descriptor as _descriptor
from typing import ClassVar as _ClassVar

DESCRIPTOR: _descriptor.FileDescriptor

class MyEnum(int, metaclass=_enum_type_wrapper.EnumTypeWrapper):
    __slots__ = []
Abc: MyEnum
Foo: MyEnum
Bar: MyEnum

```

With my changes:
```python
from google.protobuf.internal import enum_type_wrapper as _enum_type_wrapper
from google.protobuf import descriptor as _descriptor
from typing import ClassVar as _ClassVar

DESCRIPTOR: _descriptor.FileDescriptor

class MyEnum(int, metaclass=_enum_type_wrapper.EnumTypeWrapper):
    __slots__ = []
    Abc: MyEnum
    Foo: MyEnum
    Bar: MyEnum

Abc: MyEnum
Foo: MyEnum
Bar: MyEnum
```

It's a bit ugly to have all enum values twice in the interface file, but it is an accurate representation of the generated Python interface - all the values are always in the enum class as well as the parent block. The change also works with enums nested inside of classes.

### DIFF
--- a/src/google/protobuf/compiler/python/pyi_generator.cc
+++ b/src/google/protobuf/compiler/python/pyi_generator.cc
@@ -262,6 +262,11 @@ void PyiGenerator::PrintEnum(const EnumDescriptor& enum_descriptor) const {
       "class $enum_name$(int, metaclass=_enum_type_wrapper.EnumTypeWrapper):\n"
       "    __slots__ = []\n",
       "enum_name", enum_name);
+  printer_->Indent();
+  printer_->Indent();
+  PrintEnumValues(enum_descriptor);
+  printer_->Outdent();
+  printer_->Outdent();
 }
 
 void PyiGenerator::PrintEnumValues(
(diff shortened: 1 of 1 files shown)


anandolee:
Thanks for the PR
We have added some tests for stub, I may need to port this change into our internal first and apply to the tests.

anandolee:
An internal change is under review. Close this one for clean up

PR #10386, opened 2022-08-10
Use protoc version for --version

deannagarcia:
Before `protoc --version` returned the C++ version (which includes a major version specific to C++). Changing to have no major version.

### DIFF
--- a/src/google/protobuf/stubs/common.cc
+++ b/src/google/protobuf/stubs/common.cc
@@ -92,7 +92,7 @@ void VerifyVersion(int headerVersion,
   }
 }
 
-std::string VersionString(int version) {
+std::string VersionString(int version, bool cppMajor) {
   int major = version / 1000000;
   int minor = (version / 1000) % 1000;
   int micro = version % 1000;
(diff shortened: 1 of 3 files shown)


src/google/protobuf/stubs/common.cc
```diff
@@ -92,22 +92,30 @@ void VerifyVersion(int headerVersion,
...
+  }
 
   // Guard against broken MSVC snprintf().
   buffer[sizeof(buffer)-1] = '\0';
 
   return buffer;
 }
 
+std::string ProtocVersionString(int version) {
+  return VersionString(version, false);
```
fowles, line 116: instead of a bool with a default argument, simply have two methods.  They do different things and bools are not good for readability from callsites.

fowles (changes requested):

src/google/protobuf/stubs/common.cc
deannagarcia: Changed!

fowles (approved):

PR #10387, opened 2022-08-10
Remove version from makefile

deannagarcia:
We will just use the version that is provided in `configure.ac`. This reduces the number of files that we need to update when versions change.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -18,8 +18,6 @@ else
 PTHREAD_DEF =
 endif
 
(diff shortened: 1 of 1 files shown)


acozzette (approved):
LGTM, though I think this file will be disappearing soon as part of the autotools removal.

deannagarcia:
> LGTM, though I think this file will be disappearing soon as part of the autotools removal.

Didn't realize that, closing this one now.

PR #10390, opened 2022-08-11
avoid unused-parameter warning for c++

marjoleinheyndrickx:
In clang these two functions with an unused parameter give a Wunused-parameter warning. By removing the parametername the warning disappears.


src/google/protobuf/arena.h
```diff
@@ -467,7 +467,7 @@ class PROTOBUF_EXPORT PROTOBUF_ALIGNAS(8) Arena final {
     }
 
     template <typename U>
-    static Arena* GetArenaForAllocation(Rank2, const U* p) {
+    static Arena* GetArenaForAllocation(Rank2, const U*) {
```
fowles, line 470: inside google these are used.  Please switch the suppression technique to

`(void)p;`

marjoleinheyndrickx:
Suppression technique is switched to void cast, as suggested.

fowles (approved):

fowles:
Thanks for the PR.  I will merge this after our CI system finishes

PR #10400, opened 2022-08-12
Dedupe kokoro builds

mkruskal-google:
Replace duplicate kokoro continuous+presubmit configs with a single shared config.  This will avoid accidental desynchronization in the future that can cause continuous build failures.


kokoro/linux/bazel/continuous.cfg
```diff
@@ -1,16 +1 @@
...
-  key: "BAZEL_TARGETS"
-  value: "//src/..."
-}
-
-action {
-  define_artifacts {
-    regex: "**/sponge_log.*"
-  }
-}
+# Keep this empty!
```
fowles, line 1: I like it better when it says "Use common.cfg instead'

fowles (approved):

kokoro/linux/bazel/continuous.cfg
mkruskal-google: Whoops, you managed to find the one file I was using as a POC and got missed by my find/replace :P

PR #10403, opened 2022-08-14
Fix build with Python 3.11

shadchin:
The PyFrameObject structure members have been removed from the public C API.

### DIFF
--- a/python/google/protobuf/pyext/descriptor.cc
+++ b/python/google/protobuf/pyext/descriptor.cc
@@ -58,6 +58,37 @@
               : 0)                                               \
        : PyBytes_AsStringAndSize(ob, (charpp), (sizep)))
 
+#if PY_VERSION_HEX < 0x030900B1 && !defined(PYPY_VERSION)
+static PyCodeObject* PyFrame_GetCode(PyFrameObject *frame)
+{
+    Py_INCREF(frame->f_code);
+    return frame->f_code;
+}
+
+static PyFrameObject* PyFrame_GetBack(PyFrameObject *frame)
+{
+    Py_XINCREF(frame->f_back);
+    return frame->f_back;
+}
+#endif
+
+#if PY_VERSION_HEX < 0x030B00A7 && !defined(PYPY_VERSION)
+static PyObject* PyFrame_GetLocals(PyFrameObject *frame)
+{
+    if (PyFrame_FastToLocalsWithError(frame) < 0) {
+        return NULL;
+    }
+    Py_INCREF(frame->f_locals);
+    return frame->f_locals;
+}
+
+static PyObject* PyFrame_GetGlobals(PyFrameObject *frame)
+{
+    Py_INCREF(frame->f_globals);
+    return frame->f_globals;
+}
+#endif
+
 namespace google {
 namespace protobuf {
 namespace python {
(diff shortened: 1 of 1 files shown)


python/google/protobuf/pyext/descriptor.cc
```diff
@@ -58,6 +58,43 @@
...
+static inline PyObject* PyFrame_GetLocals(PyFrameObject *frame)
+{
+#if PY_VERSION_HEX >= 0x030400B1
+    if (PyFrame_FastToLocalsWithError(frame) < 0) {
+        return NULL;
+    }
+#else
+    PyFrame_FastToLocals(frame);
+#endif
+    Py_INCREF(frame->f_locals);
```
fowles, line 85: I am worried that these increfs don't have corresponding decrefs.  The old code had neither and the new code has only one.

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -100,18 +137,18 @@ bool _CalledFromGeneratedFile(int stacklevel) {
...
   while (stacklevel-- > 0) {
-    frame = frame->f_back;
+    frame = PyFrame_GetBack(frame);
     if (frame == nullptr) {
       return false;
     }
   }
 
-  if (frame->f_code->co_filename == nullptr) {
+  if (PyFrame_GetCode(frame)->co_filename == nullptr) {
```
fowles, line 146: every use of `frame->f_code` is used to access `co_filename`.  Perhaps we should make the helper do that instead.

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -132,7 +169,7 @@ bool _CalledFromGeneratedFile(int stacklevel) {
     return false;
   }
 
-  if (frame->f_globals != frame->f_locals) {
+  if (PyFrame_GetGlobals(frame) != PyFrame_GetLocals(frame)) {
```
fowles, line 172: this appears to be purely testing for at global scope.  How about the helper does that directly

python/google/protobuf/pyext/descriptor.cc
shadchin: Yes, you're right, I'll fix it

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -99,43 +136,64 @@ bool _CalledFromGeneratedFile(int stacklevel) {
...
+  Py_INCREF(frame);
   while (stacklevel-- > 0) {
-    frame = frame->f_back;
+    PyFrameObject* next_frame = PyFrame_GetBack(frame);
+    Py_DECREF(frame);
+    frame = next_frame;
     if (frame == nullptr) {
       return false;
     }
+    Py_INCREF(frame);
```
aronbierbaum, line 147: `PyEval_GetFrame` returns a borrowed reference but `PyFrame_GetBack` returns a strong reference. As a result I think you don't want this last `Py_INCREF(frame)` because `PyFrame_GetBack` will return a string reference with it's refcount already incremented.

shadchin: Valid remark, thanks

aronbierbaum (approved):

shadchin:
Is there anything else I can help with? Or are we waiting for the review to be completed?

shadchin:
gentle ping)

aronbierbaum:
@shadchin Just to verify, you are not waiting on me for anything else, correct?

fowles:
I am waiting for a variety of comments that I made on the code to be addresses

python/google/protobuf/pyext/descriptor.cc
aronbierbaum: I believe the `Py_INCREF` here is to make this function match the version of `PyFrame_GetLocals` in Python 3.11 which returns a string reference. The Python API changed to provide an few functions instead of accessing `frame->f_locals` directly.

shadchin:
> I am waiting for a variety of comments that I made on the code to be addressed

I kind of fixed all the comments. Or is there still something left?

> Just to verify, you are not waiting on me for anything else, correct?

Yes

python/google/protobuf/pyext/descriptor.cc
aronbierbaum: The helper added above is only to provide backwards compatibility. The `PyFrame_GetCode` that was added is meant to match the same behavior as the new function in Python 3.11

python/google/protobuf/pyext/descriptor.cc
aronbierbaum: I believe this falls in the same category, the helpers are for backwards compatibility so we need to keep them consistent with the new API in Python 3.11

aronbierbaum:
@fowles I added a few comments above. I didn't make the changes, so feel free to ignore, but wanted to try to help keep things moving.

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -99,43 +136,63 @@ bool _CalledFromGeneratedFile(int stacklevel) {
...
+  if (frame_globals != frame_locals) {
     // Not at global module scope
+    Py_DECREF(frame_globals);
+    Py_XDECREF(frame_locals);
+    Py_DECREF(frame);
     return false;
   }
+  Py_DECREF(frame_globals);
+  Py_XDECREF(frame_locals);
+  Py_DECREF(frame);
```
fowles, line 195: rather than an ever increasing number of decrefs, please add an error exist that you jump to with a goto that handles all the decrefs

shadchin: Add exit handler

fowles (approved):

aronbierbaum (approved):

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -58,6 +58,43 @@
...
+{
+    Py_XINCREF(frame->f_back);
+    return frame->f_back;
+}
+#endif
+
+#if PY_VERSION_HEX < 0x030B00A7 && !defined(PYPY_VERSION)
+static inline PyObject* PyFrame_GetLocals(PyFrameObject *frame)
+{
+#if PY_VERSION_HEX >= 0x030400B1
```
haberman, line 78: We require Python >=3.7, so this `#if` should be unnecessary.

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -58,6 +58,43 @@
               : 0)                                               \
        : PyBytes_AsStringAndSize(ob, (charpp), (sizep)))
 
+#if PY_VERSION_HEX < 0x030900B1 && !defined(PYPY_VERSION)
+static inline PyCodeObject* PyFrame_GetCode(PyFrameObject *frame)
```
haberman, line 62: We do not generally use `static inline`, just `static` should be sufficient. The compiler will almost certainly inline a function as small as this.

python/google/protobuf/pyext/descriptor.cc
```diff
@@ -58,6 +58,43 @@
...
+    }
+#else
+    PyFrame_FastToLocals(frame);
+#endif
+    Py_INCREF(frame->f_locals);
+    return frame->f_locals;
+}
+#endif
+
+#if PY_VERSION_HEX < 0x030B00A7 && !defined(PYPY_VERSION)
```
haberman, line 90: This looks like the same condition as above? If so please consolidate into a single `#if` block.

haberman (changes requested):

haberman (approved):

fowles:
Thanks for sticking with it through the review!!

xylar:
I believe we're still seeing the issue that this PR supposedly fixed in:
https://github.com/conda-forge/protobuf-feedstock/pull/167

Any patches or recommendations would be appreciated so we don't hold up the python 3.11 roll-out!

chenrui333:
Just curious, which release is gonna to release this change?

PR #10410, opened 2022-08-15
Clean up conformance test documentation

mkruskal-google:


conformance/README.md
```diff
@@ -12,44 +12,38 @@ contains all of the tests themselves.  Then separate programs written
...
+If you're not using Bazel to run these tests, make sure you build the C++
+tester code beforehand, e.g. from the base directory:
 
-    $ make
+    $ cmake . -Dprotobuf_BUILD_CONFORMANCE=ON && cmake --build .
+
+This will produce a `conformance_test_runner` binary that can be used to run
+conformance tests on any executable.  Pass it `--help` for more information.
 
 Running the tests for C++
```
Logofile, line 23: It appears that we're combining the sections for running C++, JavaScript, and Ruby tests, but the heading still says C++. Should the heading be made more generic? 

Or were the JS and Ruby instructions redundant with the "other languages" section, and so this de-duplicates content?

thomasvl (approved):

thomasvl:
fyi - https://github.com/apple/swift-protobuf/pull/1271 is what it seems to take for SwiftProtobuf.

Logofile (approved):

conformance/README.md
mkruskal-google: Sorry, let me clarify:

- The C++ section is still on its own, as we support two out of the box methods for running it.
- Javascript has been moved to another repo, so that section was stale
- Ruby was collapsed into "other languages", since they're all run basically the same now.  We have solid support for running via Bazel, but running through CMake would take some extra steps (they'd have to create a test executable)

Logofile: That makes sense. LGTM.

PR #10412, opened 2022-08-15
[Mac] Support bazelisk and drop -j argument.

thomasvl:
The support for -j was there since make didn't do parallel builds by default.


mkruskal-google:
This LGTM but Kokoro is currently down, so let's wait till that actually runs

mkruskal-google (changes requested):
This LGTM but Kokoro is currently down, so let's wait till that actually runs

thomasvl:
> This LGTM but Kokoro is currently down, so let's wait till that actually runs

Ok, got a full kokoro run, so I think it is likely ready for complete review.

mkruskal-google (approved):

PR #10414, opened 2022-08-16
Sort dependency manifest

TrentHouliston:
Currently the ordering of the files in the dependency manifest file are undefined as the directories are stored in an unordered list.

This has resulted in a bug specifically for me with the combination of `CMake`/`ninja`/`depfiles` as if the first file listed as an output by CMake is not the first file in the list for the dependency manifest file, ninja complains about a missing "primary output" and treats the entire list of dependencies as dirty and rebuilds them.

The outputs for protoc don't really have a "primary" output as outputs of python/cpp etc would all be equal.

This PR sorts the dependency output files before writing them to resolve this.
At least this way so long as both of the CMake list and depfile are sorted it will work as expected.

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface_unittest.cc
+++ b/src/google/protobuf/compiler/command_line_interface_unittest.cc
@@ -1779,6 +1779,39 @@ TEST_F(CommandLineInterfaceTest,
                     "$tmpdir/bar.pb: "
                     "$tmpdir/foo.proto\\\n $tmpdir/bar.proto");
 }
+
+TEST_F(CommandLineInterfaceTest,
+       DependencyManifestFileIsOrderedByFilePath) {
+  CreateTempFile("foo.proto",
+                 "syntax = \"proto2\";\n"
+                 "message Foo {}\n");
+  CreateTempFile("bar.proto",
+                 "syntax = \"proto2\";\n"
+                 "import \"foo.proto\";\n"
+                 "message Bar {\n"
+                 "  optional Foo foo = 1;\n"
+                 "}\n");
+
+  Run("protocol_compiler --dependency_out=$tmpdir/manifest_a "
+      "--test_out=$tmpdir "
+      "--descriptor_set_out=$tmpdir/a.pb --proto_path=$tmpdir bar.proto");
+
+  ExpectNoErrors();
+
+  ExpectFileContent("manifest_a",
+                    "$tmpdir/a.pb \\\n$tmpdir/bar.proto.MockCodeGenerator.test_generator: "
+                    "$tmpdir/foo.proto\\\n $tmpdir/bar.proto");
+
+  Run("protocol_compiler --dependency_out=$tmpdir/manifest_z "
+      "--test_out=$tmpdir "
+      "--descriptor_set_out=$tmpdir/z.pb --proto_path=$tmpdir bar.proto");
+
+  ExpectNoErrors();
+
+  ExpectFileContent("manifest_z",
+                    "$tmpdir/bar.proto.MockCodeGenerator.test_generator \\\n$tmpdir/z.pb: "
+                    "$tmpdir/foo.proto\\\n $tmpdir/bar.proto");
+}
 #endif  // !_WIN32
 
 TEST_F(CommandLineInterfaceTest, TestArgumentFile) {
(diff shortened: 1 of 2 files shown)


src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -1779,6 +1779,39 @@ TEST_F(CommandLineInterfaceTest,
...
+                 "}\n");
+
+  Run("protocol_compiler --dependency_out=$tmpdir/manifest_a "
+      "--test_out=$tmpdir "
+      "--descriptor_set_out=$tmpdir/a.pb --proto_path=$tmpdir bar.proto");
+
+  ExpectNoErrors();
+
+  ExpectFileContent("manifest_a",
+                    "$tmpdir/a.pb \\\n$tmpdir/bar.proto.MockCodeGenerator.test_generator: "
```
mkruskal-google, line 1802: nit: add a newline after "\\\n" to prevent wrapping

src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -1779,6 +1779,39 @@ TEST_F(CommandLineInterfaceTest,
...
+                    "$tmpdir/foo.proto\\\n $tmpdir/bar.proto");
+
+  Run("protocol_compiler --dependency_out=$tmpdir/manifest_z "
+      "--test_out=$tmpdir "
+      "--descriptor_set_out=$tmpdir/z.pb --proto_path=$tmpdir bar.proto");
+
+  ExpectNoErrors();
+
+  ExpectFileContent("manifest_z",
+                    "$tmpdir/bar.proto.MockCodeGenerator.test_generator \\\n$tmpdir/z.pb: "
```
mkruskal-google, line 1812: Same here

src/google/protobuf/compiler/command_line_interface_unittest.cc
```diff
@@ -1779,6 +1779,39 @@ TEST_F(CommandLineInterfaceTest,
...
+
+  Run("protocol_compiler --dependency_out=$tmpdir/manifest_a "
+      "--test_out=$tmpdir "
+      "--descriptor_set_out=$tmpdir/a.pb --proto_path=$tmpdir bar.proto");
+
+  ExpectNoErrors();
+
+  ExpectFileContent("manifest_a",
+                    "$tmpdir/a.pb \\\n$tmpdir/bar.proto.MockCodeGenerator.test_generator: "
+                    "$tmpdir/foo.proto\\\n $tmpdir/bar.proto");
```
mkruskal-google, line 1803: Isn't this the wrong order?  I would expect bar to come before foo...

mkruskal-google (changes requested):

src/google/protobuf/compiler/command_line_interface_unittest.cc
TrentHouliston: I only sorted the outputs as they are the ones that cause the problems.
I could also sort the dependencies list but I don't think it's necessary

mkruskal-google: gotcha, I'm not sure I see how this test locks down the new behavior then.  Am I missing something?

mkruskal-google: Basically, which test assertion makes sure the outputs are sorted?

TrentHouliston: So I do two `protoc` calls, one outputting a descriptor file with the name `a.pb` and the second time with `z.pb` while the test generator output is always `bar.proto.MockCodeGenerator.test_generator`

So one manifest file looks something like this (minus newlines and tempdir)
`a.pb bar.proto.MockCodeGenerator.test_generator: foo.proto bar.proto`
and the other like
`bar.proto.MockCodeGenerator.test_generator z.pb: foo.proto bar.proto`
having sorted the outputs alphabetically

mkruskal-google: Ah thanks for explaining that!

mkruskal-google (approved):

PR #10416, opened 2022-08-16
Add a real dependency on Abseil

mkruskal-google:
This just replaces std::unordered_set with absl::flat_hash_set in a single file, to prove that the Abseil dependency works.

### DIFF
--- a/src/google/protobuf/compiler/cpp/helpers.cc
+++ b/src/google/protobuf/compiler/cpp/helpers.cc
@@ -176,15 +176,17 @@ static const char* const kKeywordList[] = {
 #endif  // !PROTOBUF_FUTURE_BREAKING_CHANGES
 };
 
-static absl::flat_hash_set<std::string>* MakeKeywordsMap() {
-  auto* result = new absl::flat_hash_set<std::string>();
-  for (const auto keyword : kKeywordList) {
-    result->emplace(keyword);
-  }
-  return result;
-}
+const absl::flat_hash_set<std::string>& Keywords() {
+  static const auto* keywords = []{
+    auto* keywords = new absl::flat_hash_set<std::string>();
 
-static absl::flat_hash_set<std::string>& kKeywords = *MakeKeywordsMap();
+    for (const auto keyword : kKeywordList) {
+      keywords->emplace(keyword);
+    }
+    return keywords;
+  }();
+  return *keywords;
+}
 
 std::string IntTypeName(const Options& options, const std::string& type) {
   return type + "_t";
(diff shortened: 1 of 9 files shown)


acozzette (approved):

src/google/protobuf/compiler/cpp/helpers.cc
```diff
@@ -41,9 +41,9 @@
 #include <map>
 #include <memory>
 #include <queue>
-#include <unordered_set>
 #include <vector>
 
+#include "absl/container/flat_hash_set.h"
```
fowles, line 46: feels weird to put it at this point in the include list, but sure

src/google/protobuf/compiler/cpp/helpers.cc
```diff
@@ -176,15 +176,15 @@ static const char* const kKeywordList[] = {
...
+static absl::flat_hash_set<std::string>* MakeKeywordsMap() {
+  auto* result = new absl::flat_hash_set<std::string>();
   for (const auto keyword : kKeywordList) {
     result->emplace(keyword);
   }
   return result;
 }
 
-static std::unordered_set<std::string>& kKeywords = *MakeKeywordsMap();
+static absl::flat_hash_set<std::string>& kKeywords = *MakeKeywordsMap();
```
fowles, line 187: ugh, this is not safe to have at global scope

src/google/protobuf/compiler/cpp/helpers.cc
```diff
@@ -1116,7 +1116,7 @@ bool IsAnyMessage(const Descriptor* descriptor, const Options& options) {
 }
 
 bool IsWellKnownMessage(const FileDescriptor* file) {
-  static const std::unordered_set<std::string> well_known_files{
+  static const absl::flat_hash_set<std::string> well_known_files{
```
fowles, line 1119: global with destructor is also not safe.  Can you file a bug for both of these (preexisting) issues?

fowles (approved):

src/google/protobuf/compiler/cpp/helpers.cc
mkruskal-google: I was aiming for where it won't get reordered in google3.  "third_party/absl" < "third_party/protobuf" :)

src/google/protobuf/compiler/cpp/helpers.cc
mkruskal-google: ew good point, I'll just fix these now

src/google/protobuf/compiler/cpp/helpers.cc
mkruskal-google: Fixed :)

Issue #9175, opened 2021-11-01
protoc allows invalid UTF-8 in source and in option values whose type is string


jhump:
**What version of protobuf and what language are you using?**
The latest:
```
> protoc --version
libprotoc 3.19.1
```
**What operating system (Linux, Windows, ...) and version?**

OS X 11.6.1

**What runtime / compiler are you using (e.g., python version or gcc version)**

N/A

**What did you do?**

I compiled a file with invalid binary data in it. While the specs don't say that the source should be UTF-8 encoded, this issue in GitHub states it to be true: #1418. (Note: the docs should really be updated to explicitly state this, for clarity.)

In order for the source to be valid compilable source, I put it in a string literal for an option whose type is `string`.

```proto
syntax = "proto3";

package foo;

// the question mark is actually binary value 0xbc
option java_outer_classname= "my?value";
```

To play with the binary data of this example file, you can use `xxd -r` with this:
```xxd
00000000: 7379 6e74 6178 203d 2022 7072 6f74 6f33  syntax = "proto3
00000010: 223b 0a0a 7061 636b 6167 6520 666f 6f3b  ";..package foo;
00000020: 0a0a 6f70 7469 6f6e 206a 6176 615f 6f75  ..option java_ou
00000030: 7465 725f 636c 6173 736e 616d 653d 2022  ter_classname= "
00000040: 6d79 bc76 616c 7565 223b 0a              my.value";.
```

I compiled this file just to the descriptor set:
```
protoc -o tmp.protoset tmp.proto
```

**What did you expect to see**

I expected `protoc` to:
1. _Either_ replace the invalid data with the UTF replacement char � (U+FFFD)
2. _Or_ complain that the input was malformed UTF-8.

If I _replaced_ the contents of my source file to properly escape this value -- `"my\xbcvalue"` -- then I would expect the similar behavior since the actual defined scalar type for the option is `string`, not `bytes`.

**What did you see instead?**

Compilation succeeds just fine -- both with the invalid character in the source and with the escape sequence for an invalid character in a string constant. In both cases, if I examine the descriptor, it looks like so:

```
> protoc --decode google.protobuf.FileDescriptorSet < tmp.protoset
file {
  name: "tmp.proto"
  package: "foo"
  options {
    java_outer_classname: "my\274value"
  }
  syntax: "proto3"
}
```

[...]

PR #10430, opened 2022-08-19
validate UTF8 during tokenization

jhump:
This enables validation inside the tokenizer only when used from the proto source parser. The tokenizer is also used by other things (including parsing the protobuf text format). This does not attempt to enable it for other contexts.

Since the parser and tokenizer are inherently byte-based, this was kind of tricky. We basically put a little state machine in the tokenizer to recognize multi-byte codepoints and to decide if it observes valid or invalid ones, as it process the bytes one at a time. The main downside to this approach is that error messages for a naked multi-byte codepoint (not in a comment or string literal, meaning they are invalid syntax) will report the wrong codepoint -- it reports the value of just the first byte, not the whole thing. I think fixing that would require far more intrusive changes, so it's less risk to leave it this way.

This addresses one part of #9175 -- about allowing illegal UTF8 in the file input. (The second part of that bug is about allowing illegal UTF8 in values for string fields, like default values and options. Not touching that at the moment.)

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -617,6 +617,8 @@ bool Parser::Parse(io::Tokenizer* input, FileDescriptorProto* file) {
   input_ = input;
   had_errors_ = false;
   syntax_identifier_.clear();
+  // Proto source code must be in UTF-8 encoding.
+  // TODO: make this conditional somehow: compile-time Or command-line?
   input->set_require_valid_utf8(true);
 
   // Note that |file| could be NULL at this point if
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -617,6 +617,7 @@ bool Parser::Parse(io::Tokenizer* input, FileDescriptorProto* file) {
   input_ = input;
   had_errors_ = false;
   syntax_identifier_.clear();
+  input->set_require_valid_utf8(true);
 
   // Note that |file| could be NULL at this point if
   // stop_after_syntax_identifier_ is true.  So, we conservatively allocate
--- a/src/google/protobuf/io/tokenizer.h
+++ b/src/google/protobuf/io/tokenizer.h
@@ -237,6 +237,15 @@ class PROTOBUF_EXPORT Tokenizer {
   // ignored.
   void set_allow_f_after_float(bool value) { allow_f_after_float_ = value; }
 
+  // Set true to require that input be valid UTF8.
+  void set_require_valid_utf8(bool value) {
+    require_valid_utf8_ = value;
+    if (value) {
+      // initialize book-keeping flags from first char of input
+      CheckUtf8();
+    }
+  }
+
   // Valid values for set_comment_style().
   enum CommentStyle {
     // Line comments begin with "//", block comments are delimited by "/*" and
@@ -307,6 +316,12 @@ class PROTOBUF_EXPORT Tokenizer {
   bool report_whitespace_ = false;
   bool report_newlines_ = false;
 
+  bool require_valid_utf8_ = false;
+  // book-keeping to implement above requirement
+  int expect_utf8_follow_ = 0;
+  int utf8_codepoint_ = 0;
+  bool at_utf8_codepoint_end_ = false;
+
   // Since we count columns we need to interpret tabs somehow.  We'll take
   // the standard 8-character definition for lack of any way to do better.
   // This must match the documentation of ColumnNumber.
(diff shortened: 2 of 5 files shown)


jhump:
cc @amckinney

jhump:
@googleberg, curious for your input on this branch. Validating one byte at a time this way is a bit tricky. If you can think of ways to simplify the logic or additional tests you think I should add, I am all ears.

jhump:
FYI, CI complained about a bunch of pre-existing spelling errors. I pushed a second commit that tries to address them (though I'm not sure it's going to accept `"falsey"` either; probably need to change codespell config to allow it...)

I can remove that commit if desired, since it's not really part of my change. (I'm just hoping for a green build...)

jhump:
@googleberg, any chance I can get some eyes on this PR?

googleberg:
Yes. I will take a look today.

src/google/protobuf/io/tokenizer.h
```diff
@@ -322,6 +337,10 @@ class PROTOBUF_EXPORT Tokenizer {
   // Read a new buffer from the input.
   void Refresh();
 
+  // Check the most recent character read to verify that the input
+  // is correctly encoded UTF8. Does nothing if !require_valid_utf8_.
+  void CheckUtf8();
```
googleberg, line 342: The naming style for this type of call that is controlled by an option seems to be: MaybeCheckUtf8

python/google/protobuf/text_format.py
```diff
@@ -890,7 +890,7 @@ def _MergeField(self, tokenizer, message):
       expanded_any_sub_message = _BuildMessageFromTypeName(packed_type_name,
                                                            self.descriptor_pool)
       # Direct comparison with None is used instead of implicit bool conversion
-      # to avoid false positives with falsy initial values, e.g. for
+      # to avoid false positives with "falsey" initial values, e.g. for
```
googleberg, line 893: Can we put the typo fixes and other cleanups into a separate PR? I'd like to keep the clarity of "just fix the UTF8 problem".

src/google/protobuf/compiler/parser.cc
```diff
@@ -615,6 +615,7 @@ bool Parser::Parse(io::Tokenizer* input, FileDescriptorProto* file) {
   input_ = input;
   had_errors_ = false;
   syntax_identifier_.clear();
+  input->set_require_valid_utf8(true);
```
googleberg, line 618: Is this unconditional change intended?

googleberg (commented):
@jhump Thank you for this thorough PR. I love the cleanup, but for something as fundamental as "validating the characterset of proto files", please move the unrelated cleanup to a separate PR. 

In general, I think this looks reasonable. For changes to the core language, we usually do a serious design review, but I'm not sure that this warrants that full process, but I am going to have another team member that has more historical context on the parse take a look.

src/google/protobuf/compiler/parser.cc
jhump: This is what actually enables enforcement that proto sources have valid UTF8. But I am happy to make it conditional. Do you think this should be behind a compile-time #define flag? Or maybe it could be disabled via protoc command-line flag? I think it's appropriate that the default behavior, at least for open-source builds, be that this does enforce UTF8. But I'm not sure what the best "idiomatic" way to make it conditional would be. Advice?

python/google/protobuf/text_format.py
jhump: Definitely. I was really just trying to get a green build. But looks like that won't happen anyway, so I'll pull this commit out.

jhump: I've dropped this commit.

src/google/protobuf/compiler/parser.cc
jhump: I added a comment. I'm interested in advice about the most appropriate way to make this conditional.

jhump:
@haberman, ping. Any chance you can take a look soon?

fowles:
We are having a team summit this week, so expect extra latency.

jhump:
@googleberg, @haberman: I had a thought of an alternate way to approach this. Instead of having the tokenizer keep a state machine to track the UTF8 encoding, we could instead handle it only in a couple of places after numerous bytes have already been read: string literals and comments.

At least today, the tokenizer would never accept a document that has invalid UTF8 because the initial and follow bytes (which have MSB set) are not valid tokens or punctuation. So the only place they such bytes might "sneak" in are in comments and in string literals. So we could instead do a post-process on comments and string literals to validate that they contain valid UTF8. (For inline/block comments and string literals, this would also need to adjust the column position to account for multi-byte code points.) Then this could re-use existing function [`IsStructurallyValidUTF8`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/stubs/structurally_valid.cc#L556) (though we'd probably want a version that reports the position/offset of the first offending character).

Even with this PR and its state machine approach, because the tokenizer reads single bytes at a time while parsing the input, there is already the issue that error messages on bad bytes complain about byte values (not the unicode code point that might be encoded). That would still be the case with an approach like described above.

WDYT?

jhump:
Closing this pull request. I'll open a different one with the alternate approach mentioned in last comment.

Issue #10435, opened 2022-08-22
protobuf CMake dependency management using add_subdirectory


cblichmann:
I'm having a similar issue to this one reported on gRPC a while back:
https://github.com/grpc/grpc/issues/29608

Namely: I have a C++ project (https://github.com/google/binexport), that already includes Abseil and uses protobuf via CMake's `add_subdirectory()`. \
We don't use the LTS version of Abseil and therefore have no "install" of it (so no `absl-config.cmake`, etc.).

Now, Protobuf only has two modes for using Abseil: "module" (Git submodule) or "package" (system-wide install). Neither works out of the box or without resorting to workaround/hacks.

I suggest a similar solution to the one chosen by the gRPC team: Check if there exists an `absl::strings` target in the project. If there is one, skip any/all special handling for Abseil, as it is already there. If not, continue with `protobug_ABSL_PROVIDER`.

cblichmann:
Note: for PR #10436 to work on a non-LTS Abseil, one needs to set `protobuf_INSTALL` to false before including protobuf.
These are the settings that I just confirmed working:
```CMake
set(protobuf_ABSL_PROVIDER "package" CACHE STRING "" FORCE)
set(protobuf_INSTALL OFF CACHE BOOL "" FORCE)

# Download/check out into `_deps`. This ultimately calls `add_subdirectory()`
FetchContent_MakeAvailable(protobuf)
# Protobuf targets available starting here

# Make protobuf_generate_cpp available by providing necessary info to find_package()
set(Protobuf_INCLUDE_DIR "${protobuf_SOURCE_DIR}/src" CACHE INTERNAL "")
set(Protobuf_LIBRARIES protobuf::libprotobuf CACHE INTERNAL "")
find_package(Protobuf 3.21 REQUIRED)
```

mkruskal-google:
Thanks for the fix Christian!

PR #10436, opened 2022-08-22
CMake: Check whether Abseil is already present in the current project

cblichmann:
If the `absl::strings` already exists at configure time, this means protobuf was likely included via `add_subdirectory()` from a parent project that also uses Abseil. Do not check `protobuf_ABSL_PROVIDER` in that case.


cblichmann:
Should fix #10435

mkruskal-google (approved):

PR #10437, opened 2022-08-22
Upgrade to MSVC 2017, since 2015 is no longer supported

mkruskal-google:
Visual Studio 2015 is no longer supported as per our [support matrix](https://developers.google.com/protocol-buffers/docs/news/2022-08-03#cpp-support-matrix).  Specifically, our dependency on Abseil breaks this old compiler.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,16 +3,16 @@ configuration:
 
 environment:
   matrix:
-    # Only test few combinations: "Visual Studio 2015 (14) + Win64/MinGW + Debug + DLL". We can
+    # Only test few combinations: "Visual Studio 2017 (15) + Win64/MinGW + Debug + DLL". We can
     # test more combinations but AppVeyor just takes too long to finish (each
     # combination takes ~15mins).
     - platform: MinGW
       language: cpp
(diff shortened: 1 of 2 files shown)


fowles (approved):

perezd:
is this a breaking change? Should this be part of the announcement for 22.x?

mkruskal-google:
Yes it's breaking in the sense that 2015 no longer works, but isn't this covered by our new support matrix?

perezd:
> Yes it's breaking in the sense that 2015 no longer works, but isn't this covered by our new support matrix?

perhaps it's covered, but we should ensure people understand this is why, from a notification/awareness perspective. It should be explicitly communicated.

fowles:
> is this a breaking change? Should this be part of the announcement for 22.x?

This was announced specifically in https://developers.google.com/protocol-buffers/docs/news/2022-08-03#cpp-support-matrix

mkruskal-google:
I updated the PR description to be a little clearer here, does that help?

perezd:
Perfect, thanks to you both!

PR #10438, opened 2022-08-22
Fix Bazel 4 support

mkruskal-google:
This fixes Bazel 4 by adding an explicit dependency on the more recent version of bazelbuild/platforms that Abseil uses

### DIFF
--- a/kokoro/linux/java_linkage_monitor/build.sh
+++ b/kokoro/linux/java_linkage_monitor/build.sh
@@ -6,7 +6,7 @@
 # This script selects a specific Dockerfile (for building a Docker image) and
 # a script to run inside that image.
 
-use_bazel.sh 5.1.1
+use_bazel.sh 4.0.0
 
 # Change to repo root
 cd $(dirname $0)/../../..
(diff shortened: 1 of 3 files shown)


kokoro/linux/java_linkage_monitor/build.sh
```diff
@@ -6,7 +6,7 @@
 # This script selects a specific Dockerfile (for building a Docker image) and
 # a script to run inside that image.
 
-use_bazel.sh 5.1.1
+use_bazel.sh 4.0.0
```
fowles, line 9: not 4.2.2?

mkruskal-google: 4.0.0 was the broken one, I'm not 100% certain 4.2.2 requires any fixes

mkruskal-google: Note: it was only broken in our benchmark builds, so I have no way to confirm this fix actually works without downgrading a presubmit

fowles (approved):

kokoro/linux/java_linkage_monitor/build.sh
mkruskal-google: Upgraded us to 4.2.2 after offline chat

PR #10441, opened 2022-08-23
Implement IComparable<Duration> for the Duration type (C#)

### LINKED ISSUE #7628: WellKnownTypes.Duration does not implement IComparable
If it's a proto syntax change, is it for proto2 or proto3?
If it's about generated code change, what programming language? C#

I am trying to perform comparisons on the Duration type for data validation.  I get a runtime exception when I run this code:
```csharp
ComparativeValidation.LessThanOrEqual("x", Duration.FromTimeSpan(_timespanOf10), 
Duration.FromTimeSpan(_timespanOf20), Comparer<Duration>.Default); // At least one object must implement IComparable
```

Can the Duration type implement the IComparable interface?

For our immediate need, we are pursuing a manual comparison method for the seconds and nano properties. 

**Additional context**
This request is similar to the [Timestamp request](https://github.com/protocolbuffers/protobuf/issues/4267).

jskeet:
Note that unlike Timestamp, this does *not* also overload comparison
operators. Adding a `==` overload now would be a breaking change (as
it would change the meaning of existing code from a reference
comparison to a value comparison), and implementing `<`, `<=`, `>=`
and `>` without implementing `==` would be odd.

Implementing `IComparable<T>` makes sorting much easier, however.

Fixes #7628

### DIFF
--- a/csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs
+++ b/csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs
@@ -38,7 +38,7 @@ namespace Google.Protobuf.WellKnownTypes
 {
     // Manually-written partial class for the Duration well-known type,
     // providing a conversion to TimeSpan and convenience operators.
-    public partial class Duration : ICustomDiagnosticMessage
+    public partial class Duration : ICustomDiagnosticMessage, IComparable<Duration>
     {
         /// <summary>
         /// The number of nanoseconds in a second.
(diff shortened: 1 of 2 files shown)


csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs
```diff
@@ -266,5 +266,26 @@ internal static void AppendNanoseconds(StringBuilder builder, int nanos)
...
+        /// Given another duration, returns 0 if the durations are equivalent, -1 if this duration is shorter than the other, and 1 otherwise.
+        /// </summary>
+        /// <remarks>
+        /// This method expects that both durations are normalized; that is, that the values of <see cref="Seconds"/>
+        /// and <see cref="Nanos"/> are within the documented bounds.
+        /// If either value is not normalized, the results of this method are unspecified.
+        /// </remarks>
+        /// <param name="other">The duration to compare with this object.</param>
+        /// <returns>An integer indicating whether this duration is shorter or longer than <paramref name="other"/>.</returns>
+        public int CompareTo(Duration other)
```
amanda-tarafa, line 281: Unknown fields are not being taken into account here but they are being taken into account for equality and hashcode. Not sure how to take them into account but it may happen that something equals here is then not equals with `Equals`.

Maybe a note in the docs is enough.

amanda-tarafa (approved):
LGTM with a note.

csharp/src/Google.Protobuf/WellKnownTypes/DurationPartial.cs
jskeet: Given that this is a well-known type, we really shouldn't be getting any unknown fields - I think it's fine for the comparison to ignore them.

fowles (approved):

PR #10445, opened 2022-08-23
Use --announce_rc to any bug reports are more likely to include $HOME rc info

thomasvl:


fowles (approved):

thomasvl:
Different Kokoro tests getting timeouts between `kokoro:force-run` attempts, so going to merge since they do pass, just not all passing in the same test run.

PR #10446, opened 2022-08-23
Add reserved option range for mypy-protobuf

nipunn1313:
Mypy-protobuf is here
https://github.com/nipunn1313/mypy-protobuf

It currently uses extensions as explained here.
https://github.com/nipunn1313/mypy-protobuf/blob/main/proto/mypy_protobuf/extensions.proto

It's a fairly stable project for generating type stubs for protobuf.
It's been around since ~2015 (open source since 2017). Since open
sourcing it, it would make sense to reserve some extension numbers in
the global range.

See
https://github.com/nipunn1313/mypy-protobuf/issues/396

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -343,4 +343,4 @@ with info about your project (name and website) so we can add an entry for you.
 
 1. mypy-protobuf
    * Website: https://github.com/nipunn1313/mypy-protobuf
-   * Extension: 1151-1159
+   * Extension: 1151-1154
--- a/docs/options.md
+++ b/docs/options.md
@@ -340,3 +340,7 @@ with info about your project (name and website) so we can add an entry for you.
 1. Oclea Service Layer RPC
    * Website: https://oclea.com/
    * Extension: 1150
+
+1. mypy-protobuf
+   * Website: https://github.com/nipunn1313/mypy-protobuf
+   * Extension: 1151-1159


fowles:
It seems like you only need one extension number since you are using a message with subfields.

nipunn1313:
Hi @fowles - appreciate the quick turnaround.

The proto extension (as used today) uses three option extensions (from https://github.com/nipunn1313/mypy-protobuf/blob/main/proto/mypy_protobuf/extensions.proto)
```proto
    // Tells mypy to use a specific newtype rather than the normal type for this field.
    optional string casttype = 60000;

    // Tells mypy to use a specific type for keys; only makes sense on map fields
    optional string keytype = 60002;
    // Tells mypy to use a specific type for values; only makes sense on map fields
    optional string valuetype = 60003;
}
```

Example usage is like this (from https://github.com/nipunn1313/mypy-protobuf/blob/main/proto/testproto/test.proto#L70):
```proto
message Simple1 {
  optional uint32 user_id = 21 [(mypy_protobuf.casttype)="test/test_generated_mypy.UserId"];
  optional string email = 22 [(mypy_protobuf.casttype)="test/test_generated_mypy.Email"];
  map<uint32, string> email_by_uid = 23 [
    (mypy_protobuf.keytype)="test/test_generated_mypy.UserId",
    (mypy_protobuf.valuetype)="test/test_generated_mypy.Email"
  ];
}
```

Which generates `.pyi` files like this (from https://github.com/nipunn1313/mypy-protobuf/blob/main/test/generated/testproto/test_pb2.pyi#L165):
```python
class Simple1:
    user_id: test.test_generated_mypy.UserId
    email: test.test_generated_mypy.Email
    @property
    def email_by_uid(self) -> google.protobuf.internal.containers.ScalarMap[test.test_generated_mypy.UserId, test.test_generated_mypy.Email]: ...
```

As you can see here, there are 3 options used (so far).

I spent some time reading through the docs and found this - https://developers.google.com/protocol-buffers/docs/proto#customoptions - specifically `Usually you only need one extension number. You can declare multiple options with only one extension number by putting them in a sub-message:`.

We could do this, however it would be a non-backward compatible change to mypy-protobuf.
We'd have to convert to 
```proto
message Options {
    // Tells mypy to use a specific newtype rather than the normal type for this field.
    optional string casttype = 1;
    // Tells mypy to use a specific type for keys; only makes sense on map fields
    optional string keytype = 2;
    // Tells mypy to use a specific type for values; only makes sense on map fields
    optional string valuetype = 3;
}

    optional Options options = 1151;
}
```

Callsite would have to switch to:
```proto
message Simple1 {
  optional uint32 user_id = 21 [(mypy_protobuf.options).casttype="test/test_generated_mypy.UserId"];
  optional string email = 22 [(mypy_protobuf.casttype)="test/test_generated_mypy.Email"];
  map<uint32, string> email_by_uid = 23 [
    (mypy_protobuf.options).keytype="test/test_generated_mypy.UserId",
    (mypy_protobuf.options).valuetype="test/test_generated_mypy.Email"
  ];
}
```

I would really prefer to be able to make this switch in a code-compatible way, by reserving at least 3 numbers.

One possibility is that we could reserve 4 numbers. 3 for the legacy options and a 4th one for the `Options` struct approach outlined above.

nipunn1313:
I went ahead and implemented this "reserve 4 numbers" strategy.
It's now visible here https://github.com/nipunn1313/mypy-protobuf/blob/main/proto/mypy_protobuf/extensions.proto

3 legacy numbers (for backward compat) and one main number going forward.

fowles (approved):

PR #10455, opened 2022-08-24

mkruskal-google:


nickdesaulniers:
Did this PR introduce a _different ABI_ between sanitized and non-sanitized builds?

Trying to upgrade from ancient versions to 22.5 (my bisection landed on this merge commit), I hit an issue:
```
<path>/usr/lib/clang/20/include/emmintrin.h:3462:47: runtime error: member access within null pointer of type 'const struct __loadu_si128'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior <path>usr/lib/clang/20/include/emmintrin.h:3462:47
```
Turns out, we were building protobuf with `-fno-sanitize=all`, but then (accidentally) linking against object files compiled with `-fsanitized=undefined` (bad, don't do that).  Turns out, absl also has different ABIs, which was also a recent problem for us.
https://github.com/abseil/abseil-cpp/issues/1979

Simply not setting `-fno-sanitize=all` when building protobuf (when linking against object files built with `-fsanitize=*`) was my solution. Adding this note in case search engines help someone else track down similar failures in the future.

PR #10458, opened 2022-08-24
Changing windows release to VS 2017

mkruskal-google:
We no longer support VS 14, which was still in the path for this kokoro build


perezd:
are these actually in use still?

mkruskal-google:
I suspect they're not, but they're still running in CI.  So it seemed quicker to just fix it and worry about deleting it later

perezd:
SGTM, I'll have Sandy/Deanna prioritize deleting stuff to reduce migration churn in the future.

anandolee (approved):

PR #10460, opened 2022-08-24
Make abseil a public dependency of libprotobuf

haraldF:
descriptor.h includes abseil's mutex.h directly, thus abseil is not a
private dependency any more but must be included by every user of
libprotobuf.

### DIFF
--- a/cmake/libprotobuf.cmake
+++ b/cmake/libprotobuf.cmake
@@ -26,8 +26,7 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
   target_link_libraries(libprotobuf PRIVATE log)
 endif()
 target_include_directories(libprotobuf PUBLIC ${protobuf_SOURCE_DIR}/src)
-target_link_libraries(libprotobuf PRIVATE ${protobuf_ABSL_USED_TARGETS})
-target_include_directories(libprotobuf PRIVATE ${ABSL_ROOT_DIR})
+target_link_libraries(libprotobuf PUBLIC ${protobuf_ABSL_USED_TARGETS})
 if(protobuf_BUILD_SHARED_LIBS)
   target_compile_definitions(libprotobuf
     PUBLIC  PROTOBUF_USE_DLLS
--- a/cmake/libprotobuf-lite.cmake
+++ b/cmake/libprotobuf-lite.cmake
@@ -23,8 +23,7 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
   target_link_libraries(libprotobuf-lite PRIVATE log)
 endif()
 target_include_directories(libprotobuf-lite PUBLIC ${protobuf_SOURCE_DIR}/src)
-target_link_libraries(libprotobuf-lite PRIVATE ${protobuf_ABSL_USED_TARGETS})
-target_include_directories(libprotobuf-lite PRIVATE ${ABSL_ROOT_DIR})
+target_link_libraries(libprotobuf-lite PUBLIC ${protobuf_ABSL_USED_TARGETS})
 if(protobuf_BUILD_SHARED_LIBS)
   target_compile_definitions(libprotobuf-lite
     PUBLIC  PROTOBUF_USE_DLLS
--- a/cmake/libprotoc.cmake
+++ b/cmake/libprotoc.cmake
@@ -16,8 +16,7 @@ if(protobuf_HAVE_LD_VERSION_SCRIPT)
     LINK_DEPENDS ${protobuf_SOURCE_DIR}/src/libprotoc.map)
 endif()
 target_link_libraries(libprotoc PRIVATE libprotobuf)
-target_link_libraries(libprotoc PRIVATE ${protobuf_ABSL_USED_TARGETS})
-target_include_directories(libprotoc PRIVATE ${ABSL_ROOT_DIR})
+target_link_libraries(libprotoc PUBLIC ${protobuf_ABSL_USED_TARGETS})
 if(protobuf_BUILD_SHARED_LIBS)
   target_compile_definitions(libprotoc
     PUBLIC  PROTOBUF_USE_DLLS


haraldF:
Note - change 701dd83594c5d979481bc36ac58efea670a929ba makes abseil a public dependency of `libprotobuf` by including an abseil header directly in an exported header file. If this wasn't intentional, please revert that change.

Otherwise, abseil must be declared as public dependency, otherwise, `cmake` builds will complain that `absl/synchronization/mutex.h` cannot be found.

perezd:
Thanks for this, we do intend for this to be public and required in our next major release:
https://developers.google.com/protocol-buffers/docs/news/2022-08-03#abseil-dep

cmake/libprotobuf.cmake
```diff
@@ -26,7 +26,7 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
   target_link_libraries(libprotobuf PRIVATE log)
 endif()
 target_include_directories(libprotobuf PUBLIC ${protobuf_SOURCE_DIR}/src)
-target_link_libraries(libprotobuf PRIVATE ${protobuf_ABSL_USED_TARGETS})
+target_link_libraries(libprotobuf PUBLIC ${protobuf_ABSL_USED_TARGETS})
```
mkruskal-google, line 29: For consistency, I think we should make all of these public or none of them.  Notably, there's the target_include_directories in the line below, and then all of our other binaries that declare a dependency on Abseil

mkruskal-google:
Can you explain a little more how making it a private dependency breaks you?  When I decided to make it private it seemed like a more conservative first-step, since I assumed anyone using absl directly would declare their own dependency anyway and not need ours.

haraldF:
> Can you explain a little more how making it a private dependency breaks you? 

Thanks for the quick reply :)

I link to `protobuf::libprotobuf` and include a public header from the lib, e.g. . `protobuf/descriptor.h`. I'll immediately get a compilation error because `"absl/synchronization/mutex.h"` cannot be found. Since abseil is a private dependency, CMake doesn't add the abseil include directory to my search path, although it's now required by anyone including protobuf headers.

To reproduce, you can use this little CMake script:

```make
include(FetchContent)

FetchContent_Declare(
  Protobuf
  GIT_REPOSITORY https://github.com/protocolbuffers/protobuf.git
  GIT_TAG        ad55f52fdb4557953593cd096b903b0347b02f25 # or any other SHA close to HEAD
  OVERRIDE_FIND_PACKAGE
)
set(protobuf_BUILD_TESTS OFF CACHE INTERNAL "")
FetchContent_MakeAvailable(Protobuf)
```

(Note - requires CMake 3.24 or higher)

Then create a dummy project, link to `protobuf::libprotobuf` and include descriptor.h in the dummy.cpp:

```cmake
find_package(Protobuf REQUIRED) 
add_executable(dummy dummy.cpp)
target_link_libraries(dummy protobuf::libprotobuf)
```

You should get an immediate compilation error

After making abseil a PUBLIC dependency, CMake understands that it also must add the abseil include dir to projects consuming protobuf, thus the compiler error disappears.

cmake/libprotobuf.cmake
mkruskal-google: Define "legacy" CMake?  We support everything down to 3.5, so we're currently fairly constrained on what we can modernize

mkruskal-google:
Ah that makes sense, thanks for the clarification!  When I first rolled this out I tested it against a dummy CMake project, but at the time we weren't actually using Abseil anywhere so I wouldn't have hit this issue.  Given our rapidly increasing use of Abseil I think a public dependency makes sense

cmake/libprotobuf.cmake
haraldF: `target_include_directories` and scoped projects were introduced during the 2.x series, so we're fairly safe here.

mkruskal-google: ok, as long as it works with 3.10+ I'm fine cleaning up and modernizing this code :). Note, we also unfortunately have some test environments still using 3.6.

mkruskal-google (changes requested):
This lgtm, but let's update all the other private absl dependencies too

cmake/libprotobuf.cmake
```diff
@@ -26,8 +26,8 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
   target_link_libraries(libprotobuf PRIVATE log)
 endif()
 target_include_directories(libprotobuf PUBLIC ${protobuf_SOURCE_DIR}/src)
-target_link_libraries(libprotobuf PRIVATE ${protobuf_ABSL_USED_TARGETS})
-target_include_directories(libprotobuf PRIVATE ${ABSL_ROOT_DIR})
+target_link_libraries(libprotobuf PUBLIC ${protobuf_ABSL_USED_TARGETS})
+target_include_directories(libprotobuf PUBLIC ${ABSL_ROOT_DIR})
```
mkruskal-google, line 30: I thought you were going to remove this line?

cmake/libprotobuf.cmake
```diff
@@ -26,8 +26,8 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
   target_link_libraries(libprotobuf PRIVATE log)
 endif()
 target_include_directories(libprotobuf PUBLIC ${protobuf_SOURCE_DIR}/src)
-target_link_libraries(libprotobuf PRIVATE ${protobuf_ABSL_USED_TARGETS})
-target_include_directories(libprotobuf PRIVATE ${ABSL_ROOT_DIR})
+target_link_libraries(libprotobuf PUBLIC ${protobuf_ABSL_USED_TARGETS})
```
mkruskal-google, line 29: Can you update the other cmake files while you're at it?  e.g. libprotoc, protoc, libprotobuf-lite, conformance

haraldF: Done :)

mkruskal-google (approved):

PR #10464, opened 2022-08-26
CMake: Enable projects to set the C++ version

cblichmann:
This change enables projects that consume protobuf via
`FetchContent_MakeAvailable()` to set the C++ version to be used. This is
necessary, as linking code compiled for different C++ standards is asking for
trouble (and will simply not work in some cases).

Check that any version that might be set in `CMAKE_CXX_STANDARD` is new
enough (C++14 or later). On Cygwin, check if any `-std=gnu++XX` has
already been set. In all cases, default to C++14.


cblichmann:
Friendly ping, is there anything missing?

ericsalo:
Hi Christian. Most of the protobuf team was away at an offsite last week so progress on PRs has been slow. Things should start to speed up again now.

acozzette (approved):

acozzette:
@cblichmann Thanks and sorry for the delay!

PR #10496, opened 2022-09-03
Added staleness test to keep ruby-upb.{c,h} amalgamation files in sync with upb repo

haberman:
# Background

The Ruby C extension embeds an "amalgamation" of the upb sources into `ruby-upb.c` and `ruby-upb.h`. This amalgamation is a .c/.h file pair that concatenates all the sources from upb into a single .c and .h, similar to the SQLite amalgamation: https://www.sqlite.org/amalgamation.html

Prior to this PR, the amalgamation was only updated on an ad hoc basis, whenever someone decided to do it. This meant that the amalgamation was often months behind the canonical sources in https://github.com/protocolbuffers/upb.

# PR Summary

This PR adds a staleness test, so that `ruby-upb.c` and `ruby-upb.h` will always be compared against the amalgamation as generated from the upb repo. If the amalgamation is out of date, the test will fail and the user will need to run:

```
$ bazel-bin/ruby/test_amalgamation_staleness --fix
```

Initially the PR contains a patch file against upb that makes `generated_file_staleness_test()` and `amalgamate.py` work properly across repos. However this will be submitted separately to the upb repo, and `upb.patch` will be reverted before merging this PR.


ruby/BUILD.bazel
```diff
@@ -3,6 +3,7 @@
 # See also code generation logic under /src/google/protobuf/compiler/ruby.
 
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
+load("@upb//cmake:build_defs.bzl", "generated_file_staleness_test")
```
mkruskal-google, line 6: Seems a little weird for this to be in the cmake package, now might be a good time to move it given Adam's coming work to expand its usage?

ruby/BUILD.bazel
```diff
@@ -161,6 +162,29 @@ conformance_test(
...
+)
+
+genrule(
+    name = "copy_ruby_amalgamation_c",
+    srcs = ["@upb//:ruby-upb.c"],
+    outs = ["generated-in/ext/google/protobuf_c/ruby-upb.c"],
+    cmd = "cp $< $@",
+)
+
+generated_file_staleness_test(
```
mkruskal-google, line 179: If this staleness test fails what are the mitigation steps?  If the upb repo is the source of truth and nobody is editing these directly, this makes sense but might warrant a comment

ruby/BUILD.bazel
```diff
@@ -161,6 +162,29 @@ conformance_test(
...
+    name = "copy_ruby_amalgamation_h",
+    srcs = ["@upb//:ruby-upb.h"],
+    outs = ["generated-in/ext/google/protobuf_c/ruby-upb.h"],
+    cmd = "cp $< $@",
+)
+
+genrule(
+    name = "copy_ruby_amalgamation_c",
+    srcs = ["@upb//:ruby-upb.c"],
+    outs = ["generated-in/ext/google/protobuf_c/ruby-upb.c"],
```
mkruskal-google, line 175: It might be nice to allow the staleness test to more compactly support this case (i.e. just pass it the @upb targets).  These genrules seem a bit hacky

ruby/BUILD.bazel
haberman: If the staleness test fails it will tell you how to fix it. It will output a message like:

```
To fix run THIS command:
  bazel-bin/cmake/test_generated_files --fix
```

The expectation is that this will (temporarily) fail whenever we update our protobuf->upb dep to point to a newer commit: https://github.com/protocolbuffers/protobuf/blob/f7188186c52b9e3708112a622f47168e3333cf24/protobuf_deps.bzl#L118

ruby/BUILD.bazel
haberman: Makes sense, I shared a doc with you about this.

ruby/BUILD.bazel
mkruskal-google: I guess my point is more that this strategy only works for 1 directional changes.  If changes are coming from both sides this won't work

mkruskal-google (approved):

PR #10500, opened 2022-09-06
[C#] Use string literals instead of base64

nxtn-staged:
Avoids runtime overheads and reduces code size.

This reduces the size of *Google.Protobuf.Test.TestProtos.dll* by 512 bytes.

### DIFF
--- a/csharp/src/Google.Protobuf.Test.TestProtos/TestMessagesProto2.cs
+++ b/csharp/src/Google.Protobuf.Test.TestProtos/TestMessagesProto2.cs
@@ -2376,7 +2376,7 @@ namespace ProtobufTestMessages.Proto2 {
 
     /// <summary>Field number for the "default_string" field.</summary>
     public const int DefaultStringFieldNumber = 254;
-    private readonly static string DefaultStringDefaultValue = global::System.Text.Encoding.UTF8.GetString(global::System.Convert.FromBase64String("Um9zZWJ1ZA=="), 0, 7);
+    private readonly static string DefaultStringDefaultValue = "Rosebud";
 
     private string defaultString_;
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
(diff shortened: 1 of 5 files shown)


jskeet:
I'll look at this when I get a chance, but:

- There's other work at some point to consider what we do with strings that aren't valid UTF-8
- This only affects proto2 default string values as far as I can see, so I don't expect it to have a very large impact

nxtn-staged:
> There's other work at some point to consider what we do with strings that aren't valid UTF-8

It's actually possible to store invalid Unicode in .NET strings. If we're going to guard against them we should do it at the parsing level to be consistent across languages.

> This only affects proto2 default string values as far as I can see, so I don't expect it to have a very large impact

No, I don't expect a large impact. This is one of a very small set of changes from our internal fork that I think fits here.

jskeet:
> It's actually possible to store invalid Unicode in .NET strings. If we're going to guard against them we should do it at the parsing level to be consistent across languages.

Yes, absolutely. I'm just saying that this is an area we may need to change anyway, so I'd like to do some more research to see whether we can avoid changing it twice.

jskeet:
One alternative thought I've just had: we could reduce the runtime cost by performing a base64-decode operation on the string array rather than a concatenation followed by base64 decoding. We'd still have the string literals (so it wouldn't help the size of the DLL at all) but it would reduce the execution-time memory and CPU cost. (Not planning on doing this soon though.)

nicomoya123:
@jskeet Will this fix the random unicode chars that are being added to <FileName>Proto.Descriptor.File property ?
I'm generatig the class from a proto source file and I want to get the source file back from  it.
I'm trying to use the `<ClassName>Proto.Descriptor.File ` property but the string contians random chars.

If not I will open a bug separately.

jskeet:
@nicomoya123: No, that wouldn't be affected by this. I *suspect* that it's not a factor from the C# library either, but something in the FileDescriptorSet. Please open a new bug with a repro.

PR #10501, opened 2022-09-06
Avoid `String.Concat`

nxtn-staged:
There is no benefit! The compiler (csc) isn't smart enough to do this at compile time.

This reduces the size of *Google.Protobuf.Test.TestProtos.dll* by 15872 bytes.


jskeet:
I believe we *used* to do this, a very long time ago... and there was some reason why it was really bad. I can't remember what the reason was, offhand - I'll need to dig into it.

(It *may* have been that for very large proto files, we ended up with a string literal which was too big for the compiler. See https://github.com/googleapis/googleapis/blob/master/google/cloud/compute/v1/compute.proto for example - which is generated as https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1/Compute.g.cs. The `descriptorData` declaration there is over 12,000 lines of code... I suspect that at least in the past, that blew up the C# compiler.)

jskeet:
(In an ideal world, the descriptor would be a binary file included as a resource in the assembly, but that would be quite painful to get working, I suspect.)

nxtn-staged:
> It _may_ have been that for very large proto files, we ended up with a string literal which was too big for the compiler.

Fair point. A 12k-line concatenated string causes the compiler to crash with a stack overflow, but a one-line string works. Would you like me to change this to use one-line strings? ;)

jskeet:
It's possible that both VS and the compiler itself are okay with the huge concatenation now. I'll try it when I get a chance, but like the default string literal change, this will be low down my priority list I'm afraid. (Working on protobuf is only a small part of my role, and I need to prioritize my time within that.)

PR #10514, opened 2022-09-07
Fix 32-bit floating point JSON parsing of maximal values for C#

### LINKED ISSUE #10509: JSON parsing of large 32-bit floating point values fails.
The proto3 JSON spec doesn't give very much guidance on *exactly* how floating point values are represented.

The maximal finite 32-bit floating point value is precisely 340282346638528859811704183484516925440. Both .NET and (I believe) Go format this as 3.4028235e+38.

Unfortunately that fails to round-trip in .NET, because of the way we handle JSON parsing of 32-bit floating point numbers:

- Parse as a 64-bit float
- Compare with the maximal 32-bit float - if it's less than or equal, proceed, otherwise throw an exception

There are 64-bit floating point values closer to 3.4028235e+38 than the max 32-bit value, so we end up throwing an exception.
We should, instead, try casting to `float` and see whether the result is infinite.

jskeet:
Fixes #10509.


csharp/src/Google.Protobuf.Test/JsonParserTest.cs
```diff
@@ -584,8 +586,8 @@ public void NumberToFloat_Valid(string jsonValue, float expectedParsedValue)
         }
 
         [Test]
-        [TestCase("3.402824e38", typeof(InvalidProtocolBufferException))]
-        [TestCase("-3.402824e38", typeof(InvalidProtocolBufferException))]
+        [TestCase("3.4028236e38", typeof(InvalidProtocolBufferException))]
+        [TestCase("-3.4028236e38", typeof(InvalidProtocolBufferException))]
```
amanda-tarafa, line 590: super tiny nit (and just so that is clear at first sight that the string format is not the problem, use the same format in the valid and invalid tests):
```suggestion
        [TestCase("3.4028236e+38", typeof(InvalidProtocolBufferException))]
        [TestCase("-3.4028236e+38", typeof(InvalidProtocolBufferException))]
```

jskeet: Hmm... both formats should work. I'll update the tests to have both "with +" and without.

mcy (approved):

PR #10525, opened 2022-09-08
Bazelify ruby runtime

mkruskal-google:
This uses https://github.com/protocolbuffers/rules_ruby to fully Bazelify our ruby runtime code.  The Rakefile is left in place for now and is still used by our aarch64 tests.  With the current implementation ruby behaves similarly to our python wrapper, which selects whatever version is installed in the system.  Future enhancements will allow for more hermetic builds via Bazel flags to pin a specific version


.gitignore
```diff
@@ -174,7 +174,6 @@ ruby/tests/generated_code_proto2_pb.rb
 ruby/tests/multi_level_nesting_test_pb.rb
 ruby/tests/test_import_proto2_pb.rb
 ruby/tests/test_ruby_package_proto2_pb.rb
-ruby/Gemfile.lock
```
JasonLunn, line 177: I think of this as a safety mechanism to prevent accidentally committing `Gemfile.lock`. Is there a reason why we have to remove this?

WORKSPACE
```diff
@@ -69,6 +75,18 @@ kotlin_repositories()
...
 kt_register_toolchains()
 
+load("@rules_ruby//ruby:deps.bzl", "rules_ruby_dependencies", "rules_ruby_register_toolchains")
+rules_ruby_dependencies()
+rules_ruby_register_toolchains("system")
+
+load("@rules_ruby//ruby:defs.bzl", "ruby_bundle")
+ruby_bundle(
+    name = "protobuf_bundle",
+    bundler_version = "2.3.17",
```
JasonLunn, line 85: Do we have to pin the version of bundler, or could we let if float up?

maven_install.json
```diff
@@ -651,6 +651,18 @@
                 "sha256": "66fdef91e9739348df7a096aa384a5685f4e875584cce89386a7a47251c4d8e9",
                 "url": "https://repo1.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar"
             },
+            {
```
JasonLunn, line 654: Do we have to include this here, or is there a way it can be supplemented at runtime? We currently run tests in CI against both JRuby 9.2.x and 9.3.x.

protobuf_deps.bzl
```diff
@@ -13,6 +13,7 @@ PROTOBUF_MAVEN_ARTIFACTS = [
     "com.google.truth:truth:1.1.2",
     "junit:junit:4.13.2",
     "org.mockito:mockito-core:4.3.1",
+    "org.jruby:jruby-complete:9.2.20.1"
```
JasonLunn, line 16: Same question as above, with regard to how we can continue to support testing of both JRuby 9.2.x and 9.3.x.

JasonLunn (changes requested):

.gitignore
mkruskal-google: Nope, I can add this back.  I think an earlier iteration *did* check it in

maven_install.json
mkruskal-google: Good call, I'll modify rules_ruby to export the jruby jar

PR #10527, opened 2022-09-09

mkruskal-google:


CHANGES.txt
```diff
@@ -12,6 +12,13 @@
   * Use table-driven parser for reflection based objects.
   * Update Map's InternalSwap() to take a pointer to the other Map.
   * Add ARM-optimized Varint decoding functions.
+  * Minor optimization for parsing groups
+  * Declare ReflectiveProtoHook class
+  * Reduce size of VarintParse code in protocol buffers, by calling the shared routine after handling just one-byte varint encoding inline, rather than handling one-byte and two-byte varints inline.
```
haberman, line 17: Wrap lines at 80 chars.

src/google/protobuf/arena.h
```diff
@@ -49,11 +49,12 @@ using type_info = ::type_info;
 #endif
 
 #include <type_traits>
-#include <google/protobuf/arena_impl.h>
-#include <google/protobuf/port.h>
+#include "google/protobuf/arena_config.h"
```
haberman, line 52: It is intended to switch the include style for all headers?

I thought we had used <google/protobuf/foobar.h> so that it would find system-installed headers if they were present.

If we switch these to `""` includes, I wonder if we will run into issues where headers can't include each other if the user has installed them to /usr/include.

mkruskal-google: Yes, see cl/473142230.  This was based on an on-duty discussion where we noticed that on certain platforms (macos) if an older version of protobuf was installed it was taking precedence over the local files.  The behavior of <> is much more platform-dependent and I think "" is the correct thing to use in these cases.  We do have install tests that install protobuf, delete all local headers, and then build+test against the local installation.  So I think those should cover the cases you're worried about

haberman (approved):

PR #10541, opened 2022-09-12
Update generated C# and PHP code

deannagarcia:
Just started to hit errors related to descriptor.proto changing. These files were auto generated using ./generate_descriptor_proto.sh. Should fix all the continuous failures


deannagarcia:
Yeah this is old, and we found another fix. Closing it.

PR #10555, opened 2022-09-13
protoc: fix consistency with parsing very large decimal numbers

### LINKED ISSUE #10554: protoc handles very large negative integer literals inconsistently
Although the `default` field pseudo-option _looks like_ an option, it is handled differently. In particular, it allows a larger range of values for integer literals when the field's type is a float or double.

Take this example file:
```protobuf
syntax = "proto2";

message Foo {
  optional double a = 1 [default=18446744073709551615];
  optional double b = 2 [default=18446744073709551616];
  optional double c = 3 [default=-18446744073709551615];
  optional double d = 4 [default=-18446744073709551616];

  optional string w = 5 [(d)=18446744073709551615];
  optional string x = 6 [(d)=18446744073709551616];
  optional string y = 7 [(d)=-9223372036854775808];
  optional string z = 8 [(d)=-9223372036854775809];
}

  optional double d = 10001;
}
```

Attempting to compile this (`protoc test.proto -o /dev/null`) results in the following errors:
```
test.proto:5:34: Integer out of range.
test.proto:7:35: Integer out of range.
test.proto:10:30: Integer out of range.
test.proto:12:31: Integer out of range.
```

As can be seen, large integer literal values for normal and custom options are required to fit in the range [-2^63, 2^64). But for the `default` field option, the value is allowed to be (-2^64, 2^64). Values less than -2^63 are reported as erroneous only if the field's actual type is an integer. If the field value is a float or double (like in the example) then the overly large value is accepted.

This is due to the order of operations. [Processing a default value](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/parser.cc#L1234) is a different function from [processing an option value into an uninterpreted option](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/parser.cc#L1461). Processing a default value, when the field is a double or float, effectively applies the sign _last_, which is why it allows parsing such a large value (the maximum unsigned 64-bit int). But for normal and custom options, the field type is not yet known, and the representation in `UninterpretedOption` does not allow storing a negative integer that large (the field is an `int64`). So it applies the sign first, which impacts the maximum allowed range when parsing the integer value.

I think there are two options for how to address this:
1. _Always_ limit the range to the lower-bound of signed 64-bit int and the upper-bound of unsigned 64-bit int. In this case, the above line with a value of -18,446,744,073,709,551,615 would be disallowed. This is a potentially breaking change, however, since it means that existing sources that use integer literals in the range (-2^64, -2^63) for float and double default values would suddenly generate compilation errors.
2. _Always_ allow overly large decimal literals and assume they are doubles if they are out of range. This is more like most programming languages, where the error would happen when trying to assign the excessively large value to a type that cannot accept it (e.g. assigning it to a int field). Unfortunately, this results in a confusing error message for the normal/custom option case, because we have to put the value into the `UninterpretedOption` intermediate representation. This means that we _lose_ the knowledge that it was originally an integer literal (e.g. no floating point, no exponent) and end up with an error message about the field's type requiring an int value, not a float value. We could probably tweak the wording of the error so it is less confusing in the case of an excessively large int literal.

I'm happy to open a PR with either of these choices. In fact, I already implemented option 2 (and realized that the error message, without further tweaks, is a bit confusing). And option 1 is even easier to implement, but has the backwards compatibility issue.

jhump:
Fixes #10554.

I think the most controversial part of this would be the error messages, which I've updated to include the actual allowed range of values for integer literals (since this change makes the parser accept out-of-range integer literals, but they are interpreted as if float/double literals).

I re-generated code from protos in the third commit in order to make tests pass. I think that means the `main` branch is actually broken now because of stale generated code? I can remove that commit, since I understand it may be noise for this CL, but I included it to get a successful run of tests.

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -310,9 +310,14 @@ bool Parser::ConsumeNumber(double* output, const char* error) {
                                      std::numeric_limits<uint64_t>::max(),
                                      &value)) {
       *output = value;
-    } else {
-      // out of int range, treat as double literal
-      *output = io::Tokenizer::ParseFloat(input_->current().text);
+    } else if (input_->current().text[0] == '0') {
+      // octal or hexadecimal; don't bother parsing as float
+      AddError("Integer out of range.");
+      // We still return true because we did, in fact, parse a number.
+    } else if (!io::Tokenizer::TryParseFloat(input_->current().text, output)) {
+      // out of int range, and not valid float? 🤷
+      AddError("Integer out of range.");
+      // We still return true because we did, in fact, parse a number.
     }
     input_->Next();
     return true;
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -288,6 +288,16 @@ bool Parser::ConsumeInteger64(uint64_t max_value, uint64_t* output,
   }
 }
 
+bool Parser::TryConsumeInteger64(uint64_t max_value, uint64_t* output) {
+  if (LookingAtType(io::Tokenizer::TYPE_INTEGER) &&
+      io::Tokenizer::ParseInteger(input_->current().text, max_value,
+                                  output)) {
+    input_->Next();
+    return true;
+  }
+  return false;
+}
+
 bool Parser::ConsumeNumber(double* output, const char* error) {
   if (LookingAtType(io::Tokenizer::TYPE_FLOAT)) {
     *output = io::Tokenizer::ParseFloat(input_->current().text);
@@ -296,13 +306,14 @@ bool Parser::ConsumeNumber(double* output, const char* error) {
   } else if (LookingAtType(io::Tokenizer::TYPE_INTEGER)) {
     // Also accept integers.
     uint64_t value = 0;
-    if (!io::Tokenizer::ParseInteger(input_->current().text,
+    if (io::Tokenizer::ParseInteger(input_->current().text,
                                      std::numeric_limits<uint64_t>::max(),
                                      &value)) {
-      AddError("Integer out of range.");
-      // We still return true because we did, in fact, parse a number.
+      *output = value;
+    } else {
+      // out of int range, treat as double literal
+      *output = io::Tokenizer::ParseFloat(input_->current().text);
     }
-    *output = value;
     input_->Next();
     return true;
   } else if (LookingAt("inf")) {
@@ -1551,18 +1562,20 @@ bool Parser::ParseOption(Message* options,
             is_negative
                 ? static_cast<uint64_t>(std::numeric_limits<int64_t>::max()) + 1
                 : std::numeric_limits<uint64_t>::max();
-        DO(ConsumeInteger64(max_value, &value, "Expected integer."));
-        if (is_negative) {
-          value_location.AddPath(
-              UninterpretedOption::kNegativeIntValueFieldNumber);
-          uninterpreted_option->set_negative_int_value(
-              static_cast<int64_t>(0 - value));
-        } else {
-          value_location.AddPath(
-              UninterpretedOption::kPositiveIntValueFieldNumber);
-          uninterpreted_option->set_positive_int_value(value);
+        if (TryConsumeInteger64(max_value, &value)) {
+          if (is_negative) {
+            value_location.AddPath(
+                UninterpretedOption::kNegativeIntValueFieldNumber);
+            uninterpreted_option->set_negative_int_value(
+                static_cast<int64_t>(0 - value));
+          } else {
+            value_location.AddPath(
+                UninterpretedOption::kPositiveIntValueFieldNumber);
+            uninterpreted_option->set_positive_int_value(value);
+          }
+          break;
         }
-        break;
+        // value too large for an integer; fall through below to treat as floating point
       }
 
       case io::Tokenizer::TYPE_FLOAT: {
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -1941,6 +1941,22 @@ TEST_F(ParserValidationErrorTest, FieldDefaultValueError) {
       "2:32: Enum type \"Baz\" has no value named \"NO_SUCH_VALUE\".\n");
 }
 
+TEST_F(ParserValidationErrorTest, FieldDefaultIntegerOutOfRange) {
+  ExpectHasErrors(
+      "message Foo {\n"
+      "  optional double bar = 1 [default = 0x10000000000000000];\n"
+      "}\n",
+      "1:37: Integer out of range.\n");
+}
+
+TEST_F(ParserValidationErrorTest, FieldOptionOutOfRange) {
+  ExpectHasErrors(
+      "message Foo {\n"
+      "  optional double bar = 1 [foo = 0x10000000000000000];\n"
+      "}\n",
+      "1:33: Integer out of range.\n");
+}
+
 TEST_F(ParserValidationErrorTest, FileOptionNameError) {
   ExpectHasValidationErrors(
       "option foo = 5;",
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -592,6 +592,56 @@ TEST_F(ParseMessageTest, FieldOptions) {
       "}");
 }
 
+TEST_F(ParseMessageTest, FieldOptionsSupportLargeDecimalLiteral) {
+  // decimal integer literal > uint64 max
+  ExpectParsesTo(
+      "import \"google/protobuf/descriptor.proto\";\n"
+      "extend google.protobuf.FieldOptions {\n"
+      "  optional double f = 10101;\n"
+      "}\n"
+      "message TestMessage {\n"
+      "  optional double a = 1 [default = 18446744073709551616];\n"
+      "  optional double b = 2 [default = -18446744073709551616];\n"
+      "  optional double c = 3 [(f) = 18446744073709551616];\n"
+      "  optional double d = 4 [(f) = -18446744073709551616];\n"
+      "}\n",
+
+      "dependency: \"google/protobuf/descriptor.proto\""
+      "extension {"
+      "  name: \"f\" label: LABEL_OPTIONAL type: TYPE_DOUBLE number: 10101"
+      "  extendee: \"google.protobuf.FieldOptions\""
+      "}"
+      "message_type {"
+      "  name: \"TestMessage\""
+      "  field {"
+      "    name: \"a\" label: LABEL_OPTIONAL type: TYPE_DOUBLE number: 1"
+      "    default_value: \"1.8446744073709552e+19\""
+      "  }"
+      "  field {"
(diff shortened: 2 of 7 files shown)


src/google/protobuf/descriptor.cc
```diff
@@ -7683,7 +7683,7 @@ bool DescriptorBuilder::OptionInterpreter::SetOptionValue(
       if (uninterpreted_option_->has_positive_int_value()) {
         if (uninterpreted_option_->positive_int_value() >
             static_cast<uint64_t>(std::numeric_limits<int32_t>::max())) {
-          return AddValueError("Value out of range for int32 option \"" +
+          return AddValueError("Value out of range, -2,147,483,648 to 2,147,483,647, for int32 option \"" +
```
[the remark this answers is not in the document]
jhump: I was a bit worried that parens would make it look like a closed range (e.g. `(` vs `[` for range notation). Building on the other comment, to use formatted numeric constants instead of string literals for these, it looks better when the numbers are formatted without commas. Take a look and latest commit and let me know what you think.

src/google/protobuf/descriptor.cc
```diff
@@ -7693,15 +7693,15 @@ bool DescriptorBuilder::OptionInterpreter::SetOptionValue(
       } else if (uninterpreted_option_->has_negative_int_value()) {
         if (uninterpreted_option_->negative_int_value() <
             static_cast<int64_t>(std::numeric_limits<int32_t>::min())) {
-          return AddValueError("Value out of range for int32 option \"" +
+          return AddValueError("Value out of range, -2,147,483,648 to 2,147,483,647, for int32 option \"" +
```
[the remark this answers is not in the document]
jhump: Create macros for building these strings, to make the messages more DRY.

jhump:
I realized there is no new test to confirm that options and defaults allow very large decimal literals for float/double fields Adding that now.

jhump:
@fowles, I think this is ready for another look. I think I addressed your comments, and I've added a new test, too.

src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -592,6 +592,56 @@ TEST_F(ParseMessageTest, FieldOptions) {
...
+      "  field {"
+      "    name: \"d\" label: LABEL_OPTIONAL type: TYPE_DOUBLE number: 4"
+      "    options{"
+      "      uninterpreted_option{"
+      "        name{ name_part: \"f\" is_extension: true }"
+      "        double_value: -1.8446744073709552e+19"
+      "      }"
+      "    }"
+      "  }"
+      "}");
```
fowles, line 642: you might be able to use a raw string of the form `R"pb( ... )pb"` and get clang-format to do all the formatting here for you.

src/google/protobuf/descriptor.cc
```diff
@@ -7675,6 +7676,14 @@ bool DescriptorBuilder::OptionInterpreter::ExamineIfOptionIsSet(
   return true;
 }
 
+#define VALUE_OUT_OF_RANGE(T, NAME) absl::StrFormat(  \
```
fowles, line 7679: Please us a templated helper function that passes in the type name expicitly.  macros are a thing we usually avoid for maintence reasons (they interfere with static analysis

src/google/protobuf/compiler/parser_unittest.cc
jhump: All of the other tests above and below look like this (though admittedly this result descriptor is longer). If it's all the same to you, I'd rather leave it.

Is that okay, or do you insist on this change?

src/google/protobuf/descriptor.cc
jhump: Roger that.

jhump: Fixed

jhump:
BTW, I rebased to more recent master and was able to remove the commit that was just re-generating files from protos. So the diff is a little leaner/less noisy now.

src/google/protobuf/compiler/parser_unittest.cc
fowles: nah, it is fine to remain consistent.

src/google/protobuf/descriptor.cc
```diff
@@ -7675,6 +7676,22 @@ bool DescriptorBuilder::OptionInterpreter::ExamineIfOptionIsSet(
...
+template <typename T> std::string ValueOutOfRange(
+    std::string type_name, std::string option_name) {
+  return absl::StrFormat(
+    "Value out of range, %d to %d, for %s option \"%s\".", \
+    std::numeric_limits<T>::min(), std::numeric_limits<T>::max(),
+    type_name, option_name);
+}
+
+template <typename T> std::string ValueMustBeInt(
+    std::string type_name, std::string option_name) {
```
fowles, line 7688: please pass both of these as `absl::string_view`

src/google/protobuf/descriptor.cc
```diff
@@ -7722,27 +7735,22 @@ bool DescriptorBuilder::OptionInterpreter::SetOptionValue(
                  uninterpreted_option_->negative_int_value(),
                  option_field->type(), unknown_fields);
       } else {
-        return AddValueError("Value must be integer for int64 option \"" +
-                             option_field->full_name() + "\".");
+        return AddValueError(ValueMustBeInt<int64>("int64", option_field->full_name()));
       }
```
fowles, line 7739: please use the `_t` variant of all the types when naming the C++ type (the error message should still refer to the proto type as you do now)

fowles (commented):
only nits remaining, I will approve and kick the CI systems on the next round

src/google/protobuf/descriptor.cc

src/google/protobuf/descriptor.cc

src/google/protobuf/descriptor.cc
```diff
@@ -7675,6 +7676,22 @@ bool DescriptorBuilder::OptionInterpreter::ExamineIfOptionIsSet(
...
+    type_name, option_name);
+}
+
+template <typename T> std::string ValueMustBeInt(
+    absl::string_view type_name, absl::string_view option_name) {
+  return absl::StrFormat(
+    "Value must be integer, from %d to %d, for %s option \"%s\".", \
+    std::numeric_limits<T>::min(), std::numeric_limits<T>::max(),
+    type_name, option_name);
+}
```
fowles, line 7693: super nit that I cannot determine from context:  These should be in an anonymous namespace if they are not already

jhump: Done.

Thanks for the feedback. I'm learning a bit about modern C++ 👍.
I mostly do Go development and haven't done earnest C++ development in like 20 years (which you can probably tell 😆).

fowles (approved):

jhump:
@fowles, I see you approved this, but I realized one more thing missing from this before it's ready to merge -- there's a code path that could accidentally trigger a fatal error, if the literal is an octal or hex integer that is >2^64-1. I think it's an easy fix as well as easy to test.

jhump:
@fowles, see last commit and let me know what you think.

src/google/protobuf/compiler/parser.cc
```diff
@@ -296,13 +306,19 @@ bool Parser::ConsumeNumber(double* output, const char* error) {
...
                                      &value)) {
+      *output = value;
+    } else if (input_->current().text[0] == '0') {
+      // octal or hexadecimal; don't bother parsing as float
+      AddError("Integer out of range.");
+      // We still return true because we did, in fact, parse a number.
+    } else if (!io::Tokenizer::TryParseFloat(input_->current().text, output)) {
+      // out of int range, and not valid float? 🤷
       AddError("Integer out of range.");
       // We still return true because we did, in fact, parse a number.
```
jhump, line 320: This is "just in case". I can't think of any actual scenario, other octal or hex literal (handled above), where `ParseFloat` would fail. But I also really don't want this inadvertently raising an exception.

src/google/protobuf/compiler/parser.cc
```diff
@@ -296,13 +306,19 @@ bool Parser::ConsumeNumber(double* output, const char* error) {
...
     uint64_t value = 0;
-    if (!io::Tokenizer::ParseInteger(input_->current().text,
+    if (io::Tokenizer::ParseInteger(input_->current().text,
                                      std::numeric_limits<uint64_t>::max(),
                                      &value)) {
+      *output = value;
+    } else if (input_->current().text[0] == '0') {
+      // octal or hexadecimal; don't bother parsing as float
+      AddError("Integer out of range.");
+      // We still return true because we did, in fact, parse a number.
```
jhump, line 316: I put this here, instead of just relying on the branch below, because I think an octal literal would likely be successfully parsed by `ParseFloat`, but incorrectly interpreted as decimal.

src/google/protobuf/io/tokenizer.cc
```diff
@@ -1002,9 +1002,19 @@ bool Tokenizer::ParseInteger(const std::string& text, uint64_t max_value,
 }
 
 double Tokenizer::ParseFloat(const std::string& text) {
+  double result;
+  GOOGLE_LOG_IF(DFATAL,
+         !TryParseFloat(text, &result))
```
fowles, line 1007: I hate side effected in LOG statements, do you mind switching this to the more straight forward:

```
if (!TryParseFloat(...)) {
  LOG(DFATAL) << ...
}
```

jhump: done

src/google/protobuf/io/tokenizer.cc
```diff
@@ -1003,11 +1003,12 @@ bool Tokenizer::ParseInteger(const std::string& text, uint64_t max_value,
 
 double Tokenizer::ParseFloat(const std::string& text) {
   double result;
```
fowles, line 1005: let's initialize this to zero just in case we are in a non-debug build.  I would rather have a well specified behavior then UB

jhump:
```suggestion
  double result = 0;
```

jhump: Done

fowles (approved):

jhump:
> Thanks for the PR. Sorry about the many rounds of back and forth

No worries! I totally understand the desire for high code quality, and I know that I don't write C++ often enough to get everything right the first time :)

src/google/protobuf/io/tokenizer.cc
```diff
@@ -1002,9 +1002,20 @@ bool Tokenizer::ParseInteger(const std::string& text, uint64_t max_value,
 }
 
 double Tokenizer::ParseFloat(const std::string& text) {
+  double result = 0;
+  if (!TryParseFloat(text, &result)) {
+    LOG(DFATAL)
```
fowles, line 1007: blech, `LOG(DFATAL)` is the internal spelling.  You need to make this `GOOGLE_LOG(DFATAL)`

jhump: Weird. I could have sworn I ran tests locally before pushing...

jhump, line 1007:
```suggestion
    GOOGLE_LOG(DFATAL)
```

jhump: done

PR #10561, opened 2022-09-14
fix serialization warnings in generated code when compiling with Java 18 and above

### LINKED ISSUE #9673: compiler serial warnings when compiling generated source with Java 18
Version: 3.20.0-rc-1
Language: Java

macOS 12.3

using `com.google.protobuf:protoc:3.20.0-rc-1:exe:osx-x86_64` compiler artifact
invoked via `org.xolstice.maven.plugins:protobuf-maven-plugin:0.6.1`

```
% java -version
openjdk version "18" 2022-03-22
OpenJDK Runtime Environment (build 18+36-2087)
OpenJDK 64-Bit Server VM (build 18+36-2087, mixed mode, sharing)
```

**What did you do?**
Compile the code generated from the following proto definition:

`example.proto`
```
syntax = "proto3";

package test;

option java_package = "org.example.test";
option java_outer_classname = "TestProto";
option java_multiple_files = true;

message KeyValuePair {
  string key = 1;
  string value = 2;
}
```

No compiler warnings when compiling the generated Java source code.

**What did you see instead?**

As part of https://bugs.openjdk.java.net/browse/JDK-8274336 https://bugs.openjdk.java.net/browse/JDK-8274335 Java 18 added additional linter checks.

Java 18 reports the following warning for generated source files, and fails compilation if `-Werror` is specified:

```
KeyValuePair.java:[19,4] [serial] non-transient instance field of a serializable class declared with a non-serializable type
```

See [KeyValuePair.java.txt](https://github.com/protocolbuffers/protobuf/files/8344770/KeyValuePair.java.txt) for the generated source

Today possible workarounds are to either:
a) not pass -Werror to the Java compiler
b) mangle the source code generated by protoc using something like the maven-replacer-plugin, to add the necessary `@SuppressWarnings` annotations, which is rather ugly and brittle
e.g.
```
    <plugin>
        <groupId>com.google.code.maven-replacer-plugin</groupId>
        <artifactId>maven-replacer-plugin</artifactId>
        <executions>
          <execution>
            <phase>generate-sources</phase>
            <goals>
              <goal>replace</goal>
            </goals>
          </execution>
        </executions>
        <configuration>
          <includes>
            <include>target/generated-sources/**/*.java</include>
          </includes>
          <regex>true</regex>
          <regexFlags>
            <regexFlag>MULTILINE</regexFlag>
          </regexFlags>
          <replacements>
             <replacement> 
               <token>public (final )?class</token> 
               <value>@SuppressWarnings("serial") public final class</value> 
             </replacement> 
          </replacements>
        </configuration>
      </plugin>
```

xvrl:
Java 18 added additional linter checks as part of
https://bugs.openjdk.java.net/browse/JDK-8274336 and https://bugs.openjdk.java.net/browse/JDK-8274335

These additional checks cause generated protobuf code to raise the following compiler warning: "non-transient instance field of a serializable class declared with a non-serializable type"

This change fixes the code generation to annotate the generated fields with the necessary suppressions to avoid false positives.

All the code generated from src/google/protobuf/*.proto now compiles successfully without serialization warnings in Java 18.

fixes #9673


src/google/protobuf/compiler/java/map_field.cc
```diff
@@ -335,6 +335,7 @@ void ImmutableMapFieldGenerator::GenerateMembers(io::Printer* printer) const {
       "              $value_default_value$);\n"
       "}\n");
   printer->Print(variables_,
+                 "@SuppressWarnings(\"serial\")\n"
```
xvrl, line 338: this may not actually be a false-positive, since I could not find any evidence that any MapField types are actually serializable.

googleberg: I believe that our custom serialization format is implemented in such a way that we dump a bunch of details into the serialization that are not actually used in the deserialization (which is really accomplished through proto deserialization). This is probably fine until we can fix the serialization format to drop things that shouldn't be there.

googleberg (approved):

src/google/protobuf/compiler/java/map_field.cc
xvrl: sounds good, thanks for checking

xvrl:
@fowles any details on what the aarch64 failure might be?

fowles:
> @fowles any details on what the aarch64 failure might be?

Most likely a flake or bad since point.  Can you rebase onto tip of tree and I will rerun?

acozzette:
I kicked off a rerun of the failing test, let's see if maybe that will fix it. I believe the test infrastructure will run tests with the latest `main` merged in, so there shouldn't be any need to do a rebase.

xvrl:
@googleberg I didn't see this change included in the latest release. Is there anything else we need to do? I'm not very familiar with the process.

xvrl:
@googleberg @acozzette @fowles I was hoping this would make it into one of the recent releases but I may not have followed the right process. Should I be submitting a new PR to cherry-pick this change into the 21.x branch to make sure it makes it into the next release?

googleberg:
@xvrl this would naturally end up in the next minor-version release (3.22). We don't typically merge non-critical changes into patch releases. We are hoping to establish a more regular cadence of minor releases this year.

xvrl:
👍 thanks for clarifying

PR #10581, opened 2022-09-14
protoc: validate custom json_name configuration

### LINKED ISSUE #5063: protoc: does not verify that json_name does not conflict

jhump:
This makes a few other related changes, for consistency:
* Adds the existing checks, for default JSON names, to run even for proto2 files, but only as warnings (symmetry with similar check for enum value names).
* Tweaks the similar check for enum value names to include "This is not allowed in proto3", for clarity, since it's only a warning in proto2.

When checking custom JSON names, this considers it an error even in proto2 if two `json_name` options conflict. However, if a `json_name` option conflicts with a _default_ JSON name, it is just a warning in proto2 (aligns with the existing check and the similar check for enum value names).

Fixes #5063.

### DIFF
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -1972,12 +1972,14 @@ TEST_F(ParserValidationErrorTest, Proto3Default) {
 
 TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto3';\n"
-      "message TestMessage {\n"
-      "  uint32 foo = 1;\n"
-      "  uint32 Foo = 2;\n"
-      "}\n",
-      "3:9: The default JSON name of field \"Foo\" (\"Foo\") conflicts "
+      R"pb(
+      syntax = 'proto3';
+      message TestMessage {
+        uint32 foo = 1;
+        uint32 Foo = 2;
+      }
+      )pb",
+      "4:15: The default JSON name of field \"Foo\" (\"Foo\") conflicts "
       "with the default JSON name of field \"foo\" (\"foo\"). "
       "This is not allowed in proto3.\n");
 }
@@ -1985,33 +1987,38 @@ TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
 TEST_F(ParserValidationErrorTest, Proto2JsonConflictError) {
   // conflicts with default JSON names are not errors in proto2
   ExpectParsesTo(
-      "syntax = 'proto2';\n"
-      "message TestMessage {\n"
-      "  optional uint32 foo = 1;\n"
-      "  optional uint32 Foo = 2;\n"
-      "}\n",
-
-      "syntax: 'proto2'"
-      "message_type {"
-      "  name: 'TestMessage'"
-      "  field {"
-      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1"
-      "  }"
-      "  field {"
-      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'Foo' number: 2"
-      "  }"
-      "}"
+      R"pb(
+      syntax = "proto2";
+      message TestMessage {
+        optional uint32 foo = 1;
+        optional uint32 Foo = 2;
+      }
+      )pb",
+      R"pb(
+      syntax: 'proto2'
+      message_type {
+        name: 'TestMessage'
+        field {
+          label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1
+        }
+        field {
+          label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'Foo' number: 2
+        }
+      }
+      )pb"
       );
 }
 
 TEST_F(ParserValidationErrorTest, Proto3CustomJsonConflictWithDefaultError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto3';\n"
-      "message TestMessage {\n"
-      "  uint32 foo = 1 [json_name='bar'];\n"
-      "  uint32 bar = 2;\n"
-      "}\n",
-      "3:9: The default JSON name of field \"bar\" (\"bar\") conflicts "
+      R"pb(
+      syntax = 'proto3';
+      message TestMessage {
+        uint32 foo = 1 [json_name='bar'];
+        uint32 bar = 2;
+      }
+      )pb",
+      "4:15: The default JSON name of field \"bar\" (\"bar\") conflicts "
       "with the custom JSON name of field \"foo\". "
       "This is not allowed in proto3.\n");
 }
@@ -2019,45 +2026,52 @@ TEST_F(ParserValidationErrorTest, Proto3CustomJsonConflictWithDefaultError) {
 TEST_F(ParserValidationErrorTest, Proto2CustomJsonConflictWithDefaultError) {
   // conflicts with default JSON names are not errors in proto2
   ExpectParsesTo(
-      "syntax = 'proto2';\n"
-      "message TestMessage {\n"
-      "  optional uint32 foo = 1 [json_name='bar'];\n"
-      "  optional uint32 bar = 2;\n"
-      "}\n",
-
-      "syntax: 'proto2'"
-      "message_type {"
-      "  name: 'TestMessage'"
-      "  field {"
-      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1 json_name: 'bar'"
-      "  }"
-      "  field {"
-      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'bar' number: 2"
-      "  }"
-      "}"
+      R"pb(
+      syntax = 'proto2';
+      message TestMessage {
+        optional uint32 foo = 1 [json_name='bar'];
+        optional uint32 bar = 2;
+      }
+      )pb",
+      R"pb(
+      syntax: 'proto2'
+      message_type {
+        name: 'TestMessage'
+        field {
+          label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1 json_name: 'bar'
+        }
+        field {
+          label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'bar' number: 2
+        }
+      }
+      )pb"
       );
 }
 
 TEST_F(ParserValidationErrorTest, Proto3CustomJsonConflictError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto3';\n"
-      "message TestMessage {\n"
-      "  uint32 foo = 1 [json_name='baz'];\n"
-      "  uint32 bar = 2 [json_name='baz'];\n"
-      "}\n",
-      "3:9: The custom JSON name of field \"bar\" (\"baz\") conflicts "
+      R"pb(
+      syntax = 'proto3';
+      message TestMessage {
+        uint32 foo = 1 [json_name='baz'];
+        uint32 bar = 2 [json_name='baz'];
+      }
+      )pb",
+      "4:15: The custom JSON name of field \"bar\" (\"baz\") conflicts "
       "with the custom JSON name of field \"foo\".\n");
 }
 
 TEST_F(ParserValidationErrorTest, Proto2CustomJsonConflictError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto2';\n"
-      "message TestMessage {\n"
-      "  optional uint32 foo = 1 [json_name='baz'];\n"
-      "  optional uint32 bar = 2 [json_name='baz'];\n"
-      "}\n",
+      R"pb(
+      syntax = 'proto2';
+      message TestMessage {
+        optional uint32 foo = 1 [json_name='baz'];
+        optional uint32 bar = 2 [json_name='baz'];
+      }
+      )pb",
       // fails in proto2 also: can't explicitly configure bad custom JSON names
-      "3:18: The custom JSON name of field \"bar\" (\"baz\") conflicts "
+      "4:24: The custom JSON name of field \"bar\" (\"baz\") conflicts "
       "with the custom JSON name of field \"foo\".\n");
 }
 
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -1977,8 +1977,88 @@ TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
       "  uint32 foo = 1;\n"
       "  uint32 Foo = 2;\n"
       "}\n",
-      "3:9: The JSON camel-case name of field \"Foo\" conflicts with field "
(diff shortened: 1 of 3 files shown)


jhump:
@mcy, I've heard that this PR is in your territory :)

src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -1977,8 +1977,88 @@ TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
...
-      "3:9: The JSON camel-case name of field \"Foo\" conflicts with field "
-      "\"foo\". This is not allowed in proto3.\n");
+      "3:9: The default JSON name of field \"Foo\" (\"Foo\") conflicts "
+      "with the default JSON name of field \"foo\" (\"foo\"). "
+      "This is not allowed in proto3.\n");
+}
+
+TEST_F(ParserValidationErrorTest, Proto2JsonConflictError) {
+  // conflicts with default JSON names are not errors in proto2
+  ExpectParsesTo(
```
mcy, line 1987: Can you use C++ raw strings here?

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
   }
 }
 
+std::string GetJSONName(const FieldDescriptorProto& field, bool use_custom, bool* was_custom) {
```
mcy, line 5492: JSON -> Json. Acronyms should only capitalize the first letter.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+  const FieldDescriptorProto* field;
+  std::string orig_name;
+  bool is_custom;
+};
+
+void DescriptorBuilder::CheckFieldJSONNameUniqueness(
+    std::string message_name,const DescriptorProto& message, bool is_proto2, bool use_custom_names) {
+
+  std::map<std::string, struct JSONNameDetails> name_to_field;
+  for (int i = 0; i < message.field_size(); ++i) {
```
mcy, line 5520: This looks like it could be a range-for loop.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+struct JSONNameDetails {
+  const FieldDescriptorProto* field;
+  std::string orig_name;
+  bool is_custom;
+};
+
+void DescriptorBuilder::CheckFieldJSONNameUniqueness(
+    std::string message_name,const DescriptorProto& message, bool is_proto2, bool use_custom_names) {
+
+  std::map<std::string, struct JSONNameDetails> name_to_field;
```
mcy, line 5519: This is not C; the `struct` here is redundant.

mcy, line 5519: We depend on Abseil now; use absl::flat_hash_map here.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+}
+
+struct JSONNameDetails {
+  const FieldDescriptorProto* field;
+  std::string orig_name;
+  bool is_custom;
+};
+
+void DescriptorBuilder::CheckFieldJSONNameUniqueness(
+    std::string message_name,const DescriptorProto& message, bool is_proto2, bool use_custom_names) {
```
mcy, line 5517: You may need to run clang-format here.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+
+  std::map<std::string, struct JSONNameDetails> name_to_field;
+  for (int i = 0; i < message.field_size(); ++i) {
+    bool is_custom;
+    std::string name = GetJSONName(message.field(i), use_custom_names, &is_custom);
+    std::string lowercase_name = absl::AsciiStrToLower(name);
+    auto existing = name_to_field.find(lowercase_name);
+    if (existing != name_to_field.end()) {
+      auto match = existing->second;
```
mcy, line 5526: Can you give this an explicit type?

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+      if (use_custom_names && !is_custom && !match.is_custom) {
+        // if this pass is considering custom JSON names, but neither of the
+        // names involved in the conflict are custom, don't bother with a message.
+        // That will have been reported from other pass (non-custom JSON names).
+        continue;
+      }
+      std::string this_type = is_custom ? "custom" : "default";
```
mcy, line 5533: Prefer absl::string_view for constants such as these.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+      std::string existing_type = match.is_custom ? "custom" : "default";
+      // If the matched name differs (which it can only differ in case), include it
+      // in the error message, for maximum clarity to user.
+      std::string name_suffix = name == match.orig_name ? "" : " (\"" + match.orig_name + "\")";
```
mcy, line 5537: absl::StrCat here, please.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+      std::string error_message =
```
mcy, line 5538: absl::StrFormat for this complex string.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+          "\" (\"" + name + "\") conflicts with the " + existing_type + " JSON name of field \"" +
+          match.field->name() + "\"" + name_suffix + ".";
+
+      bool involves_default = !is_custom || !match.is_custom;
+      if (is_proto2 && involves_default) {
+        AddWarning(message_name, message.field(i),
+                 DescriptorPool::ErrorCollector::NAME, error_message);
+      } else {
+        if (involves_default) {
+          error_message += " This is not allowed in proto3.";
```
mcy, line 5549: absl::StrAppend.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5489,75 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+        }
+        AddError(message_name, message.field(i),
+                 DescriptorPool::ErrorCollector::NAME, error_message);
+      }
+    } else {
+      struct JSONNameDetails details = { &message.field(i), name, is_custom };
```
mcy, line 5555: This is C++, so `struct` is redundant.

src/google/protobuf/compiler/parser_unittest.cc
jhump: I can. But you'll notice that all of the test cases above and below use this format of string constants. I was going for consistency.

src/google/protobuf/descriptor.cc
jhump: Roger that. (I mostly write Go, which uses different conventions for acronyms than typical C++ and Java norms. Mea culpa.)

src/google/protobuf/descriptor.cc
jhump: Oh, I thought that was only if I used `typedef`. I'll change.

src/google/protobuf/descriptor.cc
jhump: Not sure how to do that. The message descriptor has `field_size()` and `field(int)`. I guess I need to look for some other accessor for a different kind of iteration?

src/google/protobuf/descriptor.cc
jhump: Yep, sorry. I had used `auto` because at one point I had formulated this where it was the iterator (which has umpteen template parameters...). But the type is simple and clear now, so I'll change.

jhump (commented):
Thanks for the feedback! I'll ping here again after I push revisions to address them.

src/google/protobuf/descriptor.cc
jhump: Looks like `message.field()` can be used in a range loop.

src/google/protobuf/compiler/parser_unittest.cc
jhump: Holding off on this, just because it would make these new tests look very different from all the rest. Let me know if you insist on this change or not.

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc

src/google/protobuf/descriptor.cc
jhump: Done.

I applied clang-format changes for all of my new code with one exception: it wanted to move `&` and `*` to be adjacent to the var/param instead of the type (e.g. `DescriptorProto& message` -> `DescriptorProto &message`). But the rest of this file uses the convention as seen here. I can change all of these, too, but figured it might be better (for consistency) to leave them looking like the rest of the file. Let me know what you think.

src/google/protobuf/descriptor.cc
jhump: done

jhump: done

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: done

jhump (commented):
@mcy, I think this is ready for another look.

src/google/protobuf/descriptor.cc
```diff
@@ -5974,8 +5980,7 @@ void DescriptorBuilder::CheckEnumValueUniqueness(
     const EnumValueDescriptor* value = result->value(i);
     std::string stripped =
         EnumValueToPascalCase(remover.MaybeRemove(value->name()));
-    std::pair<absl::flat_hash_map<std::string, const EnumValueDescriptor*>::iterator, bool>
-        insert_result = values.insert(std::make_pair(stripped, value));
+    auto insert_result = values.insert(std::make_pair(stripped, value));
```
jhump, line 5983: clang-format blew this type up considerably, so it seemed far easier to read to just use auto here,. Is that okay?

jhump:
ping @mcy, if you have a chance to make another pass over this

src/google/protobuf/compiler/parser_unittest.cc
mcy: I would prefer that we update this test case, even if it makes it look different from existing ones. I have been gradually burning down old-style strings throughout the codebase.

src/google/protobuf/descriptor.cc
mcy: Yes, please leave them like the rest of the file. It is possible that there are some issues with your clang-format config.

src/google/protobuf/descriptor.cc
mcy: For insert, I think `auto` is very reasonable.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5490,88 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+  // two passes: one looking only at default JSON names, and one that considers
+  // custom JSON names
+  CheckFieldJsonNameUniqueness(message_name, proto, is_proto2, false);
+  CheckFieldJsonNameUniqueness(message_name, proto, is_proto2, true);
+}
+
+namespace {
+// Helpers for function below
+
+std::string GetJsonName(const FieldDescriptorProto& field, bool use_custom,
```
mcy, line 5506: Optional: define a local struct and return that. Out-params are not great.

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5490,88 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+struct JsonNameDetails {
+  const FieldDescriptorProto* field;
+  std::string orig_name;
+  bool is_custom;
+};
+
+} // namespace
+
+void DescriptorBuilder::CheckFieldJsonNameUniqueness(
+    std::string message_name, const DescriptorProto& message, bool is_proto2,
```
fowles, line 5525: const std::string&

mcy, line 5525: Can you thread in a FileDescriptor::Syntax instead of a bool for is_proto2?

src/google/protobuf/descriptor.cc
```diff
@@ -5480,6 +5490,88 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+    bool use_custom_names) {
+
+  absl::flat_hash_map<std::string, JsonNameDetails> name_to_field;
+  for (const FieldDescriptorProto& field : message.field()) {
+    bool is_custom;
+    std::string name = GetJsonName(field, use_custom_names, &is_custom);
+    std::string lowercase_name = absl::AsciiStrToLower(name);
+    auto existing = name_to_field.find(lowercase_name);
+    if (existing != name_to_field.end()) {
```
mcy, line 5534: Switch this condition around:

```cc
if (existing == end) {
  JsonNameDetails details = {&field, name, is_custom};
  name_to_field[lowercase_name] = details;
  continue;
}

this removes some nesting for the larger block.
```

src/google/protobuf/descriptor.cc
jhump: done

jhump: done

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: done (re-used existing `JsonNameDetails` struct)

mcy (approved):

PR #10586, opened 2022-09-15
protoc: validate reserved names are identifiers

### LINKED ISSUE #6335: protoc: reserved names are not validated
Consider the following proto file:
```
message M {
    optional string foo = 1; 
    reserved "foo bar";
}
```

Here, `"foo bar"` is clearly an invalid identifier. The intention of the user is to specify two different identifiers `"foo"` and `"bar"` as reserved. Since the reserved name logic looks for exact matches, it does not warn the user that a reserved name is being used.

I expect protoc to fail if any reserved identifier is invalid.

jhump:
Fixes #6335 -- just the validation of strings part, not the ambiguity issues (which likely calls for a separate issue since it really requires a minor language change).

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -390,11 +390,14 @@ void Parser::AddError(const std::string& error) {
   AddError(input_->current().line, input_->current().column, error);
 }
 
-void Parser::AddWarning(const std::string& warning) {
+void Parser::AddWarning(int line, int column, const std::string& warning) {
   if (error_collector_ != nullptr) {
-    error_collector_->AddWarning(input_->current().line,
-                                 input_->current().column, warning);
+    error_collector_->AddWarning(line, column, warning);
   }
+}
+
+void Parser::AddWarning(const std::string& warning) {
+  AddWarning(input_->current().line, input_->current().column, warning);
 }
 
 // -------------------------------------------------------------------
@@ -1736,7 +1739,7 @@ bool Parser::ParseReservedName(std::string* name, const char* error_message) {
   int col = input_->current().column;
   DO(ConsumeString(name, error_message));
   if (!io::Tokenizer::IsIdentifier(*name)) {
-    AddError(line, col, absl::StrFormat("Reserved name \"%s\" is not a valid identifier.", *name));
+    AddWarning(line, col, absl::StrFormat("Reserved name \"%s\" is not a valid identifier.", *name));
     return false;
   }
   return true;
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -1740,7 +1740,6 @@ bool Parser::ParseReservedName(std::string* name, const char* error_message) {
   DO(ConsumeString(name, error_message));
   if (!io::Tokenizer::IsIdentifier(*name)) {
     AddWarning(line, col, absl::StrFormat("Reserved name \"%s\" is not a valid identifier.", *name));
-    return false;
   }
   return true;
 }
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -1730,7 +1730,8 @@ bool Parser::ParseReserved(DescriptorProto* message,
 }
 
 bool Parser::ParseReservedName(std::string* name, const char* error_message) {
-  // capture position of token
+  // Capture the position of the token, in case we have to report an
+  // error after it is consumed.
   int line = input_->current().line;
   int col = input_->current().column;
   DO(ConsumeString(name, error_message));
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -51,6 +51,7 @@
 #include "absl/strings/ascii.h"
 #include "absl/strings/escaping.h"
 #include "absl/strings/str_cat.h"
+#include "absl/strings/str_format.h"
 #include "google/protobuf/descriptor.h"
 #include "google/protobuf/descriptor.pb.h"
 #include "google/protobuf/io/tokenizer.h"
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/parser.cc
```diff
@@ -1728,11 +1729,23 @@ bool Parser::ParseReserved(DescriptorProto* message,
   }
 }
 
+bool Parser::ParseReservedName(std::string* name, const char* error_message) {
+  // capture position of token
```
jhump, line 1733:
```suggestion
  // capture position of token, in case we have to report an error after it is consumed
```

jhump:
ping @mkruskal-google, mind taking a quick pass?

mkruskal-google:
Hey Joshua, sorry for the delay and thanks for the PR!  This LGTM, but would you mind filling out a change proposal? https://docs.google.com/document/d/1RPgh6engDi319RhZ9cKmLQYLgyF7xscBHpqfHTpkQ9Q/edit#heading=h.f4bilzmezze2

mkruskal-google (changes requested):
Hey Joshua, sorry for the delay and thanks for the PR! This LGTM, but would you mind filling out a change proposal? https://docs.google.com/document/d/1RPgh6engDi319RhZ9cKmLQYLgyF7xscBHpqfHTpkQ9Q/edit#heading=h.f4bilzmezze2

jhump:
@mkruskal-google, I've filled out the proposal for a small language change:
https://docs.google.com/document/d/1aJeMlnK-4t6rYkuRy1fShP_13p5ghjTq4Wne-jvBXpg/edit#
(You'll have to request access to comment or send me your email address so I can pro-actively give you that permission.)

cc @fowles

It's a very minor change -- especially since it requires no changes in the descriptor, which means it has no impact on anything but the parser.

But do you _also_ want a proposal for this PR? That seems a little heavy-weight -- what this PR is doing is trivial. Is the issue that you expect many existing files in google3 to have invalid reserved names, so this change need to a more elaborate flagging/migration process? If so, then it sounds like maybe we should just close this PR and possibly add the check in a future "edition" (where there would be a standard way of controlling this logic via message option).

mkruskal-google:
Thanks Joshua!  Please add protobuf-team@google.com as editors too.

I think it would be reasonable to fold this PR into your proposal.  The most important purpose of these change proposals is just recording the motivation for and discussion around new changes.  However, this PR is also a breaking change so we need to be careful about rollout.  I suspect there will be *some* pre-existing mistakes that this will now break.  For the google3-side that shouldn't be too hard to clean up, and I'm more worried about how we release this to OSS

jhump:
> I'm more worried about how we release this to OSS

FWIW, [the](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec#reserved) [grammars](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#reserved) on the protobuf developer site already state that this must be an identifier. So I think this is just fixing a bug -- the compiler should have never accepted the bad input in the first place. If there are users that have bad sources, they will have to fix their sources before they can upgrade to the new version of `protoc`. (Which, from my experience, is fairly common when upgrading to a new version of a compiler or library -- if I was depending on buggy behavior, I may have to change my code to upgrade.)

Is there something about this particular issue that makes that approach unreasonable?

I'll update the proposal doc to describe this validation step. But I'm not sure what I should say about the migration: as stated above, for open-source land, I think it's reasonable to simply release this as a bug-fix and require users to fix any invalid sources before they upgrade.

But I think the google3/internal migration would be something I can't really write about. Can you just change the compiler, let Tap (or whatever current CI system is called) turn red for some projects, and expect teams to fix their sources to get green again? Or do you use a global presubmit queue to identify all of the problematic sources first, before integrating this change into google3?

mkruskal-google:
No there's nothing unique about this issue, and I agree that this could be viewed as just a bug fix.  But it is a *breaking* bug fix, so we might want to be careful about how it's released.  I'm not sure what the best procedure is and need to discuss it with the team.

As far as google3 goes, I don't think you need to worry about it at all.  I can handle fixing any downstream breakages when we import this

mkruskal-google:
To unblock this PR, what do you think about downgrading this error to a warning?  That way it's *not* a breaking change and is just a harmless new validation check.  Once we meet to decide on your tier 1 proposal we can then just flip it to an error (assuming we decide to)

jhump:
> what do you think about downgrading this error to a warning?

@mkruskal-google, done.

src/google/protobuf/compiler/parser.cc
```diff
@@ -1728,11 +1732,24 @@ bool Parser::ParseReserved(DescriptorProto* message,
...
 
+bool Parser::ParseReservedName(std::string* name, const char* error_message) {
+  // Capture the position of the token, in case we have to report an
+  // error after it is consumed.
+  int line = input_->current().line;
+  int col = input_->current().column;
+  DO(ConsumeString(name, error_message));
+  if (!io::Tokenizer::IsIdentifier(*name)) {
+    AddWarning(line, col, absl::StrFormat("Reserved name \"%s\" is not a valid identifier.", *name));
+    return false;
```
mkruskal-google, line 1743: Does this cause the parser to fail?

jhump: Oops, I bet it does. When I added the `ExpectHasWarnings` test function, looks like I modeled too closely after `ExpectHasErrors`, so it doesn't have the assertion to verify that the operation succeeds 🤦. So even though the test passes, that's likely not enough to verify that the parse succeeded. Fix incoming...

jhump: Pushed a fix.

Interestingly, it does _not_ cause the parser to fail. It causes the current statement to be aborted/skipped, but the parser keeps going (in the interest of reporting all syntax errors instead of failing fast). But, at the end of the parse, it looks at the number of errors reported to decide if the whole operation was a failure. 

So it was not failing... but it was potentially skipping some of the source and producing an incomplete descriptor.

Even if/when this gets changed back to reporting an error, this function can still return true (since a reserved name *has* been consumed/parsed).

mkruskal-google: Hah good to know!  This LGTM otherwise, lemme rerun the tests and merge it

mkruskal-google (approved):

PR #10592, opened 2022-09-16
Auto-generate CMake file lists in GitHub action

acozzette:
This GitHub action will run after each pull request merge and will auto-update the file lists in in src/file_lists.cmake. The action will run as our bot account.

I realized that if a bug somehow made the file generation non-idempotent, this could trigger an infinite loop of commits, so I put in an extra safeguard against that. If the previous commit was by "Protobuf Team Bot", the GitHub action will revert any local changes to ensure that no new commit will be made.

### DIFF
--- a/.github/workflows/generated_cmake.yml
+++ b/.github/workflows/generated_cmake.yml
@@ -1,8 +1,13 @@
-name: Generated CMake File Lists
+name: Generate CMake File Lists
 
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - main
+      - '[0-9]+.x'
+      # The 21.x branch predates support for auto-generation of the CMake file
+      # lists, so we make sure to exclude it.
+      - '!21.x'
 
 jobs:
   cmake:
@@ -12,13 +17,12 @@ jobs:
       fail-fast: false   # Don't cancel all jobs if one fails.
 
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Bazel read-only caching
-        run: echo "BAZEL_CACHE_AUTH=--remote_upload_local_results=false" >> $GITHUB_ENV
-      - name: Generate CMake files
-        run: cd ${{ github.workspace }} && bazel build //pkg:gen_src_file_lists --test_output=errors $BAZEL_CACHE $BAZEL_CACHE_AUTH
-      - name: Compare to Golden file
-        run: diff -du bazel-bin/pkg/src_file_lists.cmake src/file_lists.cmake
-      - name: Report
-        run: echo "::error file=cmake/update_file_lists.sh::CMake files are stale, please run cmake/update_file_lists.sh"
-        if: failure()
+      - uses: actions/checkout@v3
+        with:
+          # Note: this token has an expiration date, so if the workflow starts
+          # failing then you may need to generate a fresh token.
+          token: ${{ secrets.BOT_ACCESS_TOKEN }}
+      - name: Configure name and email address in Git
+        run: cd ${{ github.workspace }} && git config user.name "Protobuf Team Bot" && git config user.email "protobuf-team-bot@google.com"
+      - name: Commit and push update
+        run: cd ${{ github.workspace }} && ./cmake/push_auto_update.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/generated_cmake.yml
```diff
@@ -1,8 +1,10 @@
-name: Generated CMake File Lists
+name: Generate CMake File Lists
 
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - main
+      - '[0-9]+.x'
```
mkruskal-google, line 7: This won't work on 21.x will it?

acozzette: Right, I don't think this will work on 21.x. I was assuming that it would automatically be disabled on that branch since this code will never make it onto 21.x, but now that I think of it I'm not entirely sure how GitHub handles this, so I went ahead and excluded 21.x explicitly.

acozzette:
I did test it out on my personal fork and confirmed that it worked there, but I guess the only way to know for sure that it works in the main repo is to submit it and see what happens. That's a good idea to shuffle the file list, I think I'll try that afterward as a way to confirm that the action works.

.github/workflows/generated_cmake.yml
mkruskal-google: Ah right, it probably wouldn't affect 21.x anyway (the original workflow also would have been broken)

mkruskal-google:
Nice, that's more than anything I thought of!  As long as the functionality works any breakages will likely just be from config/settings

.github/workflows/generated_cmake.yml
```diff
@@ -12,13 +17,17 @@ jobs:
...
+        run: cd ${{ github.workspace }} && git config user.name "Protobuf Team Bot" && git config user.email "protobuf-team-bot@google.com"
+      - name: Generate CMake file lists
+        run: cd ${{ github.workspace }} && ./cmake/update_file_lists.sh
+      - name: Guard against infinite loop
+        # As an extra precaution, we undo any pending changes if the most
+        # recent commit was made by the bot. This reduces the risk of a bug
+        # causing an infinite loop of commits.
+        run: if (git log -1 --pretty=format:'%an' | grep -q "Protobuf Team Bot"); then git reset --hard HEAD; fi
+      - name: Commit and push change
+        run: cd ${{ github.workspace }} && git add -A && (git diff --staged --quiet || git commit -am "Auto-generate CMake file lists") && git push
```
mkruskal-google, line 33: I feel like this could be reworded to make it clearer what the commit did and why.  e.g.
- "Auto-refresh CMake configuration after PR #{pr number}"
- "Auto-generate new CMake configution after PR #{pr number}"

mkruskal-google: I'm leaning towards the second one just so we can unify all our bot-generated commits with the prefix "auto-generated"

acozzette: Ok, let me give that a try. I think this logic is getting complicated enough that I should probably move it into a shell script at this point, and that should make it much easier to do than if I did it all directly in the yml file.

mkruskal-google: sgtm, I think you need to update this to run it?

mkruskal-google (approved):

.github/workflows/generated_cmake.yml
acozzette: Yes, this is now ready to go and has the more detailed commit message.

cmake/push_auto_update.sh
```diff
@@ -0,0 +1,41 @@
...
+  pr="$pr_from_squash"
+fi
+
+if [ ! -z "$pr" ]; then
+  commit_message="Auto-generate CMake file lists after PR #$pr"
+else
+  # If we are unable to determine the pull request number, we fall back on this
+  # default commit message. Typically this should not occur, but could happen
+  # if a pull request was merged via a rebase.
+  commit_message="Auto-generate CMake file lists"
```
mkruskal-google, line 36: I think you can get this more reliably from GITHUB_REF or GITHUB_EVENT_PATH environment variables (e.g. see https://github.com/actions/checkout/issues/58)

acozzette: Unfortunately I don't think that will work in this case, since those seem to be specific to the `pull_request` event whereas we're using the `push` event. I don't think the `push` event necessarily knows anything about the latest PR, since it will run on any update to the branch, even if was updated without a PR.

PR #10594, opened 2022-09-16
[PHP] allow dynamic properties in Message

brettmc:
PHP 8.2 complains about Internal\Message setting dynamic properties:

`Creation of dynamic property xxx is deprecated
/usr/src/myapp/vendor/google/protobuf/src/Google/Protobuf/Internal/Message.php:153`

Since the class uses dynamic properties a bit, enable them explicitly with an attribute.


deannagarcia (approved):

deannagarcia:
Thanks for the fix!

PR #10597, opened 2022-09-16
[ObjC] Provide a protocol for GPBExtensionRegistry's lookup support.

thomasvl:
This allows some to use an alternative registry if they have a different implementation.

This is really just wiring though the change to use the GPBExtensionRegistry protocol vs the concrete GPBExtensionRegistry through the other apis.


objectivec/GPBCodedInputStream.h
```diff
@@ -184,7 +184,7 @@ CF_EXTERN_C_END
  *                          extensions for message.
  **/
 - (void)readMessage:(GPBMessage *)message
-  extensionRegistry:(nullable GPBExtensionRegistry *)extensionRegistry;
+  extensionRegistry:(nullable NSObject<GPBExtensionRegistry> *)extensionRegistry;
```
dmaclach, line 187: I'd love to have these as id as opposed to NSObject. Why do you need them as NSObject?

thomasvl: I think I've had hiccups with that in the past, but will swap it, the change would be: `NSObject<GPBExtensionRegistry> *` → `id<GPBExtensionRegistry>` (i.e. - the `*` goes away also)

thomasvl: Done

dmaclach (approved):

Issue #10593, opened 2022-09-16
Degrade: Kotlin explicit api mode


STAR-ZERO:
**What version of protobuf and what language are you using?**
Version: 3.21.6
Language: Kotlin

**What operating system (Linux, Windows, ...) and version?**
M1 Mac

**What runtime / compiler are you using (e.g., python version or gcc version)**
gcc

**What did you do?**
1. Run protoc command in [the tutorial](https://developers.google.com/protocol-buffers/docs/kotlintutorial).

**What did you expect to see**
Generated Kotlin source code is compatible with [Explicit API mode](https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md).

**What did you see instead?**
Generated code is missing `public` modifier.

**Anything else we should know about your project / environment**

protoc 3.20.2 is working, but 3.21.6 is not working. I found the related [issue](https://github.com/protocolbuffers/protobuf/issues/8857) but this change seems to have been lost from 3.21.x.

deannagarcia:
Sorry that this changed, we received conflicting requests on whether to have the public modifiers. I am working on adding them back in to support explicit API mode.

alexander-yevsyukov:
@deannagarcia, please see the attached project which can help you reproduce the issue with versions starting from `3.19.6`.

[protobuf-kotlin.zip](https://github.com/protocolbuffers/protobuf/files/10083817/protobuf-kotlin.zip)

It still happens with `3.21.9`, at least to me, under M1 Max. It was a little better with `3.20.2.` which only seemed to lost `public` modifiers for `val`s.

Just modify the value of the `protobufVersion` in `build.gradle.kts` and then run `./gradlew clean build`.

@armiol, FYI.

armiol:
@deannagarcia @alexander-yevsyukov let me add my observations on the subject.

After some investigation, I have found that the changes made within this PR are indeed present in the `protobuf/master` branch. And they were present in every changeset since this PR was merged.

However, the source code attached to the [v21.9 release](https://github.com/protocolbuffers/protobuf/releases/tag/v21.9) does NOT have these fixes. As far as I can tell. 
Here is the reference to the source code: https://github.com/protocolbuffers/protobuf/archive/refs/tags/v21.9.zip

What I think happened was that the 21.9 release had not been not made from `master`, but rather from some side branch. Which, for some reason, never got these fixes.

[...]

PR #10598, opened 2022-09-16
Add public modifier to generated kotlin code

STAR-ZERO:
fix https://github.com/protocolbuffers/protobuf/issues/10593
Add `public` modifier to generated kotlin code for [Explicit API mode](https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md) compatible.


STAR-ZERO:
Let me close.
https://github.com/protocolbuffers/protobuf/pull/10599 was merged.

PR #10608, opened 2022-09-16
Open up visibility for some compiler internals

mkruskal-google:
https://github.com/protocolbuffers/protobuf/pull/9985 made all of our language-specific compiler code visibility-restricted in Bazel builds.  While most of this code was intended to be internal, we never explicitly specified it.  Specifically, gRPC (and likely other code generators) needs visibility into language-specific naming helpers.

This PR takes a middle-ground approach for our 22.x breaking release, only exposing what we think is reasonable for downstream codegen to reuse.  Other alternatives open for discussion are:
1) Make all of the compiler internals public again.  This wouldn't break anyone, but is intentionally exposing a lot of code we intended to make private
2) Keep all of the compiler internals private and have downstream teams branch any functionality they need.  This would  block gRPC's upgrade to our main branch, and require additional work from their end


src/google/protobuf/compiler/csharp/BUILD.bazel
```diff
@@ -6,6 +6,15 @@ load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("//build_defs:cpp_opts.bzl", "COPTS")
 
+cc_library(
+    name = "csharp_names",
+    hdrs = ["csharp_names.h"],
```
haberman, line 11: While we're in cleanup mode, can we name this just `names.h`? The stuttering names (`compiler/csharp/csharp_xyz.h`) are legacy I think, other languages have moved away from these names, eg. for C++ in https://github.com/protocolbuffers/protobuf/pull/9669)

src/google/protobuf/compiler/objectivec/objectivec_names.h
```diff
@@ -0,0 +1,143 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
```
haberman, line 5: Ditto for this file, `names.h`?

src/google/protobuf/compiler/php/php_names.h
```diff
@@ -0,0 +1,59 @@
...
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
```
haberman, line 16: Ditto, `names.h`?

haberman (changes requested):

fowles (approved):
This seems like a reasonable compromise to me

src/google/protobuf/compiler/csharp/BUILD.bazel
mkruskal-google: sure, good idea

haberman:
This LGTM. We should get signoff from @jskeet and @thomasvl too since we're publicly exposing symbols in the C#/Obj-C generators. They may want to take a look in case there are any parts of those APIs they would want to clean up before making them officially public.

mkruskal-google:
Good point, I'll add them thanks Josh!

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
```diff
@@ -39,10 +39,10 @@
 #include "google/protobuf/descriptor.h"
 #include "google/protobuf/descriptor.pb.h"
 #include "google/protobuf/io/zero_copy_stream.h"
+#include "google/protobuf/compiler/objectivec/names.h"
 
-// clang-format off
```
thomasvl, line 43: I think you need to keep this wrapping or future clang-format runs might move it.

src/google/protobuf/compiler/objectivec/BUILD.bazel
```diff
@@ -6,6 +6,15 @@ load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("//build_defs:cpp_opts.bzl", "COPTS")
 
+cc_library(
+    name = "names",
+    hdrs = ["names.h"],
+    copts = COPTS,
+    include_prefix = "google/protobuf/compiler/objectivec",
```
thomasvl, line 13: why is this needed?

src/google/protobuf/compiler/objectivec/BUILD.bazel
```diff
@@ -6,6 +6,15 @@ load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("//build_defs:cpp_opts.bzl", "COPTS")
 
+cc_library(
+    name = "names",
+    hdrs = ["names.h"],
```
thomasvl, line 11: This still feels like a breaking change, as grpc/etc. have to update to now import names.h, should we put the private things in names.h and use objectivec_helpers.h as the public one so they have less code to change?

src/google/protobuf/compiler/objectivec/objectivec_helpers.h
mkruskal-google: Adding a trailing comment:
// Must be included last.

is how we get around this in other files.  I just updated this for consistency, but it should work the same

src/google/protobuf/compiler/objectivec/BUILD.bazel
```diff
@@ -6,6 +6,15 @@ load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("//build_defs:cpp_opts.bzl", "COPTS")
 
+cc_library(
```
thomasvl, line 9: This only exposes the header for folks to include it, how do they link against the code?

mkruskal-google: Good point, lemme refactor this with public/private headers

src/google/protobuf/compiler/objectivec/BUILD.bazel
haberman: In other languages we use `names.h` as the public header exposing naming functions only. I would like to keep this convention consistent between languages.

src/google/protobuf/compiler/objectivec/BUILD.bazel
mkruskal-google: This is a better practice for modeling visibility.  Bazel doesn't enforce "private" internal headers today

src/google/protobuf/compiler/objectivec/BUILD.bazel
mkruskal-google: I refactored these so that the "names" target also includes the definitions for linkage.  Thanks for flagging this!

src/google/protobuf/compiler/objectivec/BUILD.bazel
thomasvl: `includes` should get propagated by `cc_library`, so it impacts everything that depends on it.  If you only want it to apply to this target, a `copt` would be the better choice. But I don't think it's actually needed since the compiler should always check in the same directory as the source file doing the `#include`.

src/google/protobuf/compiler/csharp/names.cc
```diff
@@ -0,0 +1,225 @@
...
+    return descriptor->options().csharp_namespace();
+  }
+  return UnderscoresToCamelCase(descriptor->package(), true, true);
+}
+
+std::string GetClassName(const Descriptor* descriptor) {
+  return ToCSharpName(descriptor->full_name(), descriptor->file());
+}
+
+std::string GetClassName(const EnumDescriptor* descriptor) {
```
jskeet, line 97: Arguably this should be GetEnumName, but I propose leaving it as it is.

jskeet (approved):
C# looks fine to me, with a few concerns to be checked.

src/google/protobuf/compiler/csharp/names.cc
```diff
@@ -0,0 +1,225 @@
...
+#include "google/protobuf/port_def.inc"
+
+namespace google {
+namespace protobuf {
+namespace compiler {
+namespace csharp {
+
+namespace {
+
+std::string StripDotProto(const std::string& proto_file) {
```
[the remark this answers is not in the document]
mkruskal-google: Yep, this one is totally private

src/google/protobuf/compiler/csharp/names.cc
mkruskal-google: Yea, let's punt that to a later cleanup.  I'm just moving this right now

src/google/protobuf/compiler/csharp/names.h
```diff
@@ -73,9 +73,35 @@ std::string PROTOC_EXPORT GetClassName(const Descriptor* descriptor);
...
+//   The unqualified name of the C# class that provides access to the file
+//   descriptor. Proto compiler generates
 //   such class for each .proto file processed.
+std::string GetReflectionClassUnqualifiedName(const FileDescriptor* descriptor);
+
+// Gets unqualified name of the extension class
+// Requires:
+//   descriptor != NULL
+//
+// Returns:
```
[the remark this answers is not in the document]
mkruskal-google: Whoops, I misread this.  I'd prefer to get the comments right now, updated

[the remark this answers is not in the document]
jskeet: Okay - I'm away for a couple of days now. I'm happy for you to merge and I'll do a fast-follow on Friday, or I can update this branch on Friday if you'd prefer.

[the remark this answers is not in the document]
mkruskal-google: I did update it to hopefully be more accurate, but please take another look when you get back :)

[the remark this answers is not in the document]
jskeet: I think we can remove "message" from "message extensions" in both places. I haven't double-checked, but I'm 99% sure it will also be used for field extensions etc. Also, there's a trailing "processed." line at the bottom.

Other than that, this is fine - thanks for fixing.

thomasvl (approved):

PR #10626, opened 2022-09-20
[ObjC] Some prep work and then run clang-format over the ObjC runtime sources and tests

thomasvl:


dmaclach (approved):

thomasvl:
The Linux JRuby and Python failures look unrelated. The previous Kokoro runs passed, and the ObjC code won't be compiled on Linux.

PR #10628, opened 2022-09-20
Remove deprecated protoc-artifacts directory

mkruskal-google:
build-protoc.sh and pom.xml are the only files we still use in this directory.  The build script is still used for part of the release process, but will eventually be replaced and can be moved to the legacy release directories.  The pom file is moved to the root directory


deannagarcia (commented):
I don't like the idea of having a pom file in the root directory. I think it's pretty confusing (it would be very reasonable to assume that file also publishes the Java runtime).

I also don't want all the documentation about the Maven protoc package to be deleted.

Why is it a goal to delete this directory?

mkruskal-google:
Putting it in the root directory made the most sense to me, since this is the pom file for the protobuf compiler (rather than any java-specific thing).  I'm open to moving it somewhere else, but I'm not sure where.

What value does this documentation provide?  We're moving all our release code back to google3 so I don't see why any of our users would need this information on top of what we already provide in the other README files.

The main goal for deleting this directory is that it's a remnant of the legacy release process, and reducing the number of top level directories makes our repo less confusing.

deannagarcia:
Can we put this in java/protoc-artifacts? I want it to be explicit that it's just the compiler.

I didn't read all the documentation and I agree a lot is not needed anymore, but maybe a short README with what this publishes (including a list of supported architectures) and where it can be found.

mkruskal-google:
Will this file survive the release migration?  Or will we not need it once that's done?

java/protoc/pom.xml
```diff
@@ -30,4 +30,115 @@
       scm:git:https://github.com/protocolbuffers/protobuf.git
     </connection>
   </scm>
+  <build>
```
deannagarcia, line 33: This all shouldn't be added. I changed the way the pom file works so we upload one at a time and don't list the artifacts like this.

protoc-artifacts/README.md
```diff
@@ -1,161 +0,0 @@
-# Build scripts that publish pre-compiled protoc artifacts
```
deannagarcia, line 1: Can we keep the relevant documentation in here to move to java/protoc?

deannagarcia (approved):

deannagarcia (approved):

PR #10646, opened 2022-09-23
[c++] Optimize Varint paring for c++

zzachimed:
optimize Varint paring by avx instruction. In our funciton level benchmarking, we get about 2.7x  maximum performance boost.
Run on (112 X 3100 MHz CPU s)
CPU Caches:
  L1 Data 48K (x56)
  L1 Instruction 32K (x56)
  L2 Unified 1280K (x56)
  L3 Unified 43008K (x2)
Load Average: 113.09, 103.01, 100.32
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_Parse32_1           5.41 ns         5.41 ns    127614707
BM_ParseAvx32_1        5.41 ns         5.41 ns    131481251
BM_Parse32_2           7.60 ns         7.60 ns    101486965
BM_ParseAvx32_2        7.21 ns         7.21 ns    100662140
BM_Parse32_3           18.6 ns         18.6 ns     37747465
BM_ParseAvx32_3        15.8 ns         15.8 ns     45671267
BM_Parse32_4           21.3 ns         21.3 ns     32821412
BM_ParseAvx32_4        15.4 ns         15.4 ns     45606688
BM_Parse32_5           24.0 ns         24.0 ns     28883162
BM_ParseAvx32_5        14.2 ns         14.2 ns     48919738
BM_Parse32_6           25.3 ns         25.3 ns     28171374
BM_ParseAvx32_6        13.1 ns         13.1 ns     53483401
BM_Parse32_7           26.9 ns         26.9 ns     25914018
BM_ParseAvx32_7        13.1 ns         13.1 ns     53334350
BM_Parse32_8           29.1 ns         29.1 ns     24179552
BM_ParseAvx32_8        13.1 ns         13.1 ns     53368440
BM_Parse32_9           32.8 ns         32.8 ns     21251833
BM_ParseAvx32_9        13.1 ns         13.1 ns     52993618
BM_Parse32_10          35.1 ns         35.1 ns     19933683
BM_ParseAvx32_10       13.1 ns         13.1 ns     52927471
BM_Parse64_1           5.96 ns         5.96 ns    114083173
BM_ParseAvx64_1        5.98 ns         5.98 ns    117008484
BM_Parse64_2           6.74 ns         6.74 ns    106423611
BM_ParseAvx64_2        6.55 ns         6.55 ns    108907164
BM_Parse64_3           18.7 ns         18.7 ns     37244811
BM_ParseAvx64_3        13.1 ns         13.1 ns     54240934
BM_Parse64_4           21.7 ns         21.7 ns     33925661
BM_ParseAvx64_4        12.8 ns         12.8 ns     53840538
BM_Parse64_5           25.2 ns         25.2 ns     28179568
BM_ParseAvx64_5        12.7 ns         12.7 ns     57454421
BM_Parse64_6           28.4 ns         28.4 ns     24626749
BM_ParseAvx64_6        13.1 ns         13.1 ns     54277091
BM_Parse64_7           32.2 ns         32.2 ns     21875611
BM_ParseAvx64_7        12.9 ns         12.9 ns     55145458
BM_Parse64_8           35.0 ns         35.0 ns     19970139
BM_ParseAvx64_8        12.5 ns         12.5 ns     52959753
BM_Parse64_9           38.3 ns         38.3 ns     18326964
BM_ParseAvx64_9        16.6 ns         16.6 ns     42330051
BM_Parse64_10          41.6 ns         41.6 ns     16794702
BM_ParseAvx64_10       16.4 ns         16.4 ns     42469314

in our object level test. End to End benchmarking for object that all are varint fields ,we can get about 20% performance boost

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -35,6 +35,9 @@
 #include <cstring>
 #include <string>
 #include <type_traits>
+#include <immintrin.h>
+#include <emmintrin.h>
+#include <pmmintrin.h>
 
 #include "google/protobuf/io/coded_stream.h"
 #include "google/protobuf/io/zero_copy_stream.h"
@@ -601,9 +604,81 @@ static PROTOBUF_ALWAYS_INLINE inline uint64_t Ubfx7(uint64_t data,
 }
 
 #endif  // __aarch64__
+#if defined(__SSE3__) && defined(__AVX512BW__) && defined(__AVX512VL__) && \
+    defined(__BMI2__)
+
+constexpr static const uint64_t AvxByteMask[] = {0x0,
+                                                   0xff,
+                                                   0xffff,
+                                                   0xffffff,
+                                                   0xffffffff,
+                                                   0xffffffffff,
+                                                   0xffffffffffff,
+                                                   0xffffffffffffff,
+                                                   0xffffffffffffffff};
+static const __m128i Zero128I = _mm_setzero_si128();
 
 template <typename T>
-PROTOBUF_NODISCARD const char* VarintParse(const char* p, T* out) {
+inline PROTOBUF_NODISCARD const char* VarintParseAvx(const char* p, T* out) {
+  auto ptr = reinterpret_cast<const uint8_t*>(p);
+  uint32_t res = ptr[0];
+  if (!(res & 0x80)) {
+    *out = res;
+    return p + 1;
+  }
+  uint32_t byte = ptr[1];
+  res += (byte - 1) << 7;
+  if (!(byte & 0x80)) {
+    *out = res;
+    return p + 2;
+  }
+  __m128i data128i = _mm_lddqu_si128(reinterpret_cast<const __m128i*>(p));
+  const uint8_t* dptr = reinterpret_cast<uint8*>(&data128i);
+  //"0x3ff" = 10 bytes
+  __mmask16 msb =
+      _mm_mask_cmp_epi8_mask(0x3ff, data128i, Zero128I, _MM_CMPINT_NLT);
+  uint32_t cnt = _tzcnt_u32(msb);
+
+  if constexpr (std::is_same<T, uint32_t>::value) {
+    if (PROTOBUF_PREDICT_TRUE(cnt < 10)) {
+      uint32_t maskIndex = cnt < 5 ? cnt + 1 : 5;
+      uint64_t data64 =
+          (*(reinterpret_cast<const uint64_t*>(dptr))) & AvxByteMask[maskIndex];
+      *out = static_cast<uint32>(_pext_u64(data64, 0x7F7F7F7F7F));
+      return p + cnt + 1;
+    }
+    return nullptr;
+  }
+
+  if constexpr (std::is_same<T, uint64_t>::value) {
+    if (cnt < 8) {
+      uint32_t maskIndex = cnt + 1;
+      uint64_t data64 =
+          (*(reinterpret_cast<const uint64_t*>(dptr))) & AvxByteMask[maskIndex];
+      *out = _pext_u64(data64, 0x7F7F7F7F7F7F7F7F);
+      return p + cnt + 1;
+    }
+
+    if (cnt < 10) {
+      uint32_t maskIndex = cnt + 1 - 8;
+      uint64_t tempValueLo = _pext_u64(*(reinterpret_cast<const uint64_t*>(dptr)),
+                                     0x7F7F7F7F7F7F7F7F);
+      uint64_t tempValueHi =
+          _pext_u64(*(reinterpret_cast<const uint64_t*>(dptr + 8)) &
+                        AvxByteMask[maskIndex],
+                    0x7F7F);
+      *out = (tempValueHi << 56) | tempValueLo;
+      return p + cnt + 1;
+    }
+  }
+
+  return nullptr;
+}
+
+#endif  //__SSE3__ && __AVX512BW__ && __AVX512VL__ && __BMI2__
+
+template <typename T>
+PROTOBUF_NODISCARD const char* VarintParseNoneAvx(const char* p, T* out) {
 #if defined(__aarch64__) && defined(PROTOBUF_LITTLE_ENDIAN)
   // This optimization is not supported in big endian mode
   uint64_t first8;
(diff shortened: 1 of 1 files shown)


acozzette:
@zzachimed Thank you for the pull request, but unfortunately we are not in a good position to accept this kind of change right now. Probably the biggest problem is that we are not internally building with the necessary compiler flags to make these instructions available, so if we accepted this change then we wouldn't currently be able to test it. We have also run into problems in the past where an optimization looks great in a microbenchmark but doesn't work out in real binaries in production. This kind of optimization seems promising, but I think we will need our internal performance experts to drive this effort to make sure that we can support it properly.

jorgbrown:
Thanks again for the pull request.  I really appreciate that you took the time to run benchmarks, and include them along with your code.

For what it's worth, if we were to go this route, we would want these changes at a minimum:

1) You read the next 16 bytes with _mm_lddqu_si128.  However, our input stream prefetches only 16 bytes, and that's to be sure we can read 5 bytes of tag plus 10 bytes of varint.  At the time your code runs, we may have already retrieved 5 bytes of tag information, so you may only have 11 bytes available to read... therefore you can't read 16.  Thankfully, at the time you use _mm_lddqu_si128, you've already read two bytes and incorporated them into res... so you only really need to read 8 bytes anyway.

2) _mm_mask_cmp_epi8_mask is an AVX-512 instruction.  That's a deal-breaker right there, because of the awful way that Intel can slow down the clock rate of every core on a chip just because one of them started using AVX-512.  (See https://blog.cloudflare.com/on-the-dangers-of-intels-frequency-scaling/ )  Something like _mm_movemask_epi8 , an SSE2 instruction, would be much better.

3) You're using pext.  While most x86 chips these days support it, some important ones - Zen1 and Zen2 - support it with microcode that is so slow that it's usually better not to use pext.  On a Zen it's at least twice as fast to use this for 64-bit:

// Note: Assumes the high bits of each byte of `in` are zero.
int64_t pext_7F7F7F7F7F7F7F7F(uint64_t in) {
  in += (in & 0x00FF00FF00FF00FF);
  in += (in & 0x0000FFFF0000FFFF) * 3;
  return ((0xFFFFFFFF & in) >> 3) + (in >> 32 << 25);
}

( For your perusal, I have various other algorithms for pext at https://godbolt.org/z/xvn8W4en3 )

= - = - =

Now, regarding your benchmarks, here's the really bad news...

I believe they are repeatedly testing specific widths of incoming integers.  That's very good for any branches which might be in your code, since the CPU will be able to predict them.  It's not so good for real-world parsing, however.  In the real-world, two things ruin these optimizations and others like them:

a) Most integers are single-byte encodings.  Those that are not? They are typically double-byte encodings.
b) In situations where you're dealing with larger encodings, their length is random.

You'll want to add a benchmark whose data reflects the reality of (a) and (b).  Sadly, once you've done so, you'll find it to be incredibly difficult to beat a simple loop.  In particular, if your parser is inline, then its decode will be repeated throughout the binary.  A larger faster decoder can dramatically increase the overall size of an application, and the resulting instruction cache misses can overwhelm the time savings of the faster decoder.

And that's what's happened time and time again: people with great optimization experience have come up with some great code to speed up decoding, only to find that the results of a large real-world program aren't much affected, and sometimes are actually slower.  :-(

zzachimed:
@jorgbrown 👍 
  verry thanks and appreciate for your carefuly comments . sorry for latter response for current busy work. 
1 for your concern about avx512 issue 
    we disucssed with intel senior principal architect  Moga, Adrian C [adrian.c.moga@intel.com]. This is his comment about"
Intel can slow down the clock rate of every core on a chip just because one of them started using AVX-512"
"I don’t this this is true. 
First, the article is old and based on a SKX 12C SKU.

Second, the Cloudflare blog writer is confused by wikichip reference, which refers to N cores using the same license.
Nothing in there says that one core doing AVX-512 brings down the other cores to base frequency if the run SSE.

Also, if I understand his experiment correctly, he is running a fully-loaded system with two deployments.
So, he’s observing a noisy neighbor effect for frequency under TDP constraints.
If some cores use more power with AVX-512, the other cores may also see a frequency degradation.
I think even that may not be true starting with SPR, as the cores have autonomous power management.
But I am not 100% certain. I will ask and let you know.

Thanks!
Adrian 
"
2 for your concern about out range memory issue about "_mm_lddqu_si128":
1) first current stream mechinism can additional bytes for our processing . our decode process varintparse is private function and server for up message level  depend on the buffer from streaming processing. In sepecifct words the buffer is from ParseContext::Done processing(every object decode start it will call this func). Maybe you coulddeepdive in this process you will found some wondeful mechnism to prevent memory run out in some range .In this buffer size check and buffer swap processing ,there has some woderful mechinisim to  processing some visit out of range issue --at most additional 16 bytes (kSlopBytes)for run-out processing(in EpsCopyInputStream class that is base of ParseContext) . It afford additional  in "char buffer_[2 * kSlopBytes] = {};" . The specific processing is comlicated and hard to description very shortly.  So its  better to check that ParseContext::Done code line by line.
2) second read extra bytes didn't impact the algo correctless:
beacause the highest bytes 's msb always zero. even we get the additional undefined bytes ,this will not implact our correctless.

3 for your concern about pext execution effiency issue in AMD platform:
 May we need check if the new AMD platform has same issue. Or we could do some tricky processing for

zzachimed:
@jorgbrown  👍
verry thanks and appreciate for your carefuly comments . sorry for latter response for current busy work.

1 for your concern about avx512 issue "	_mm_mask_cmp_epi8_mask is an AVX-512 instruction. That's a deal-breaker right there, because of the awful way that Intel can slow down the clock rate of every core on a chip just because one of them started using AVX-512."
As _mm_mask_cmp_epi8_mask is only uses 128 bit lane though it is AVX512 instruction set, we are pretty sure it will only runs under SSE power license, you won’t see a frequency drop actually.  And Cloudflare blog is quite old and outdated.  it refers to N cores using the same license which is not true for latest Xeon CPU like Icelake. Nothing in there says that one core doing AVX-512 brings down the other cores to base frequency if the run SSE.

2 for your concern about out range memory issue about "_mm_lddqu_si128":
  1）	first current stream mechinism can additional bytes for our processing . our decode process varintparse is private function and server for up message level depend on the buffer from streaming processing. In sepecifct words the buffer is from ParseContext::Done processing(every object decode start it will call this func). Maybe you could deep dive in this process you will found some wondeful mechnism to prevent memory run out in some range .In this buffer size check and buffer swap processing ,there has some woderful mechinisim to processing some visit out of range issue --at most additional 16 bytes (kSlopBytes)for run-out processing(in EpsCopyInputStream class that is base of ParseContext) . It afford additional space in "char buffer_[2 * kSlopBytes] = {};" . The specific processing is comlicated and hard to description very shortly. So its better to check that ParseContext::Done code line by line.
 2） second read extra bytes didn't impact the algo correctless:
beacause the highest bytes 's msb always zero. even we get the additional undefined bytes ,this will not implact our correctless.
3 for your concern about pext execution effiency issue in AMD platform “PDEP/PEXT can be much slower than the equivalent integer instructions on Zen and Zen2 CPUs.”:
It looks like Zen3 will have some improvement regarding to this.

PR #10650, opened 2022-09-23
Remove the pre Xcode 10.2 support.

thomasvl:
objectivec/README.md lists Xcode 10.2 as the minimum, update things accordingly.

- Remove code paths referencing the older versions.
- Remove support from the testing script.

### DIFF
--- a/objectivec/DevTools/full_mac_build.sh
+++ b/objectivec/DevTools/full_mac_build.sh
@@ -224,27 +224,9 @@ if [[ "${DO_XCODE_IOS_TESTS}" == "yes" ]] ; then
   # just pick a mix of OS Versions and 32/64 bit.
   # NOTE: Different Xcode have different simulated hardware/os support.
   case "${XCODE_VERSION}" in
-    [6-8].* )
-      echo "ERROR: The unittests include Swift code that is now Swift 4.0." 1>&2
-      echo "ERROR: Xcode 9.0 or higher is required to build the test suite, but the library works with Xcode 7.x." 1>&2
+    [6-9].* )
+      echo "ERROR: Xcode 10.2 or higher is required." 1>&2
       exit 11
-      ;;
-    9.[0-2]* )
-      XCODEBUILD_TEST_BASE_IOS+=(
-          -destination "platform=iOS Simulator,name=iPhone 4s,OS=8.1" # 32bit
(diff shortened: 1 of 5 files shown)


objectivec/DevTools/full_mac_build.sh
```diff
@@ -224,28 +224,10 @@ if [[ "${DO_XCODE_IOS_TESTS}" == "yes" ]] ; then
   # just pick a mix of OS Versions and 32/64 bit.
   # NOTE: Different Xcode have different simulated hardware/os support.
   case "${XCODE_VERSION}" in
-    [6-8].* )
-      echo "ERROR: The unittests include Swift code that is now Swift 4.0." 1>&2
-      echo "ERROR: Xcode 9.0 or higher is required to build the test suite, but the library works with Xcode 7.x." 1>&2
+    [6-9].* )
+      echo "ERROR: Xcode 9.0 or higher is required." 1>&2
```
dmaclach, line 228: Don't we want this to be 10.2 or better? Can we push it up to higher than that (given that Apple won't accept anything iOS with less than 13 AFAIK)

thomasvl: Fixed message.

I'll look at cycling back to raising this in the future. But for now, just want to keep it as is, but prune some of the stale things for where the mins were.

objectivec/DevTools/full_mac_build.sh
```diff
@@ -315,15 +296,15 @@ if [[ "${DO_XCODE_TVOS_TESTS}" == "yes" ]] ; then
   )
   case "${XCODE_VERSION}" in
     [6-9].* )
-      echo "ERROR: Xcode 10.0 or higher is required to build the test suite." 1>&2
+      echo "ERROR: Xcode 10.2 or higher is required." 1>&2
       exit 11
       ;;
     10.* | 11.* | 12.*)
```
dmaclach, line 302: How sensitive are we to 10.2... do we need to check the 10.*?

dmaclach (approved):

objectivec/DevTools/full_mac_build.sh
thomasvl: This script is mainly just use for CI and anyone working on the library (me), so I'm good with keeping it simple at the moment.

thomasvl:
Only changes are to macOS specific files, so the JRuby and Python failures are likely Kokoro flake.

PR #10652, opened 2022-09-23
[ObjC] Raise the min OS versions (and required Xcode)

thomasvl:
Xcode min: 13.3.1
iOS min: 10.0
macOS min: 10.12
tvOS min: 12.0
watchOS min: 6.0

Apple's AppStore requirements now require Xcode 13:
  https://developer.apple.com/news/?id=2t1chhp3

Update to the minOS version and Xcode version that also matches what Firebase as done as that seems like a common set for most Apple platforms (https://firebase.google.com/docs/ios/setup).


dmaclach (approved):

thomasvl:
The Ruby test seems like infrastructure flake updating things on the CI machine, and the CL doesn't change anything related to that.

PR #10658, opened 2022-09-26
Revert "Merge pull request #10470 from jhump/jh/source-code-info-pseudo options"

mkruskal-google:
This reverts commit dd052c9dc601f831611ef9b04d70d8210c71053d, reversing changes made to 0d3eaed6eb91f5b7702918586f730e4dd1ceccd2.

This breaks https://github.com/kythe/kythe, as well as some internal code.

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -431,11 +431,6 @@ Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent) {
 }
 
 Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent,
-                                           SourceCodeInfo* source_code_info) {
-  Init(parent, source_code_info);
-}
-
-Parser::LocationRecorder::LocationRecorder(const LocationRecorder& parent,
                                            int path1,
                                            SourceCodeInfo* source_code_info) {
   Init(parent, source_code_info);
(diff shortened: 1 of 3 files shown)


mcy (approved):

mkruskal-google:
@jhump FYI

jhump:
> This breaks https://github.com/kythe/kythe, as well as some internal code.

@mkruskal-google, how did it break things? Are those tools _expecting_ the source code info as it was previously being generated? These changes are very minor, so I'm surprised that anything would break. Is it just a test expectation that went red -- like some date derived from the source code info that is now slightly different?

mkruskal-google:
Hey Joshua, sorry for the delay on this.  @jaysachs can probably fill you in on some details.  I'm not entirely sure what broke on their end.

jaysachs:
So this PR changed what turned up as the text pan for the default value expression. See https://github.com/kythe/kythe/compare/master...jaysachs:kythe:proto-default-field-change-sadness for how we'd have to adapt. Per my (admittedly rudimentary) reading of the code, our only option in the face of this PR as originally presented to restore the original desired behavior is to grovel through the bytes of the buffer, rather than depending on the spans as returned by the parser. I *think* we need the value location in file descriptor, not the no-op that this PR created.

jaysachs:
Kythe is specifically interested in the span of the default value, for the purposes of overlaying decorations on the code. (Note that this is only an issue for enums, as default values for int64, string etc are literal values, so there's no symbol reference to decorate in those cases.)

Option (2) could work, we could adapt the code to choose the proper span.

jhump:
@jaysachs, ping, how hard would it be to have Kythe handle a situation where the descriptor had two spans for the same path for default value? I'd like to revisit this soon and implementing what was mentioned above, two spans for the default value (one with the full declaration that includes the option name, another with _just_ the value), is the fairly straight-forward. But I don't want to land a change like that only to have it reverted again.

The spans should be emitted (IMO) with the full declaration first and then the value second, mainly so they are effectively sorted by span start (since the value's span must begin after the option name in the full declaration). That is also the easiest way to implement it since the parser starts parsing the full declaration before the value portion.

I wonder if this would even require a change to Kythe at all. If it encounters multiple locations for the path to the default value, how would it behave? If it would use the last location (perhaps a long-shot, since keep first seems more likely), then it would "just work" with the proposed descriptor tweaks.

jaysachs:
Do you have proposed PR so we could evaluate it?

@zrlk for context

jhump:
> Do you have proposed PR so we could evaluate it?

@jaysachs, if you mean a PR to this protocolbuffers project, I don't yet. But I can probably get one together this week and ping this thread with a link.

PR #10671, opened 2022-09-29
[C#] Reduce delegate instantiation

nxtn-staged:
Prevents instantiating and storing references to delegates by using virtual methods. This doesn't introduce a breaking change yet.


jskeet:
As with other PRs, I'm afraid it's unlikely that I'll get to this any time soon. I have very limited time available to work on protobufs, and I have to prioritize it carefully.

PR #10673, opened 2022-09-29
Clean up TextFormat parser

mkruskal-google:
Fix TextFormat parser to build up recurring (but supposedly not repeated) sub-messages directly from text rather than building a new sub-message and merging the fully formed message into the existing field.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,7 +7,11 @@
   * Change the Lite runtime to prefer merging from the wireformat into mutable
     messages rather than building up a new immutable object before merging. This
     way results in fewer allocations and copy operations.
-  * Make message-type extensions merge from wire-format instead of building up instances and merging afterwards. This has much better performance.
+  * Make message-type extensions merge from wire-format instead of building up
+    instances and merging afterwards. This has much better performance.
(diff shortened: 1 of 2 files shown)


googleberg (approved):

zhangskz (approved):

ldziedziul:
Do you plan to backport it to 3.19.x ?

zhangskz:
This fix should have already been included in 3.19.x in https://github.com/protocolbuffers/protobuf/commit/b0e8e607dc1ace17c823276d3384eeda775bbd76

iliyaj-loginsoft:
Do we have a fix for the CVE-2022-3509 vulnerability in the 2.x.x series? If not, could you please let me know where this vulnerability is triggered in the 2.x.x version, and in which file?

mkruskal-google:
3.9 (and even more so 2.x) are extremely old and out of support.  You can try to backport this to an old version, but they're considered EOL.  We would recommend using a newer version that has this fix

Ashutosh-Loginsoft:
@mkruskal-google do you have any POC ? I wanted to check correctness of my patch.
if not can you please tell me how this vulnerability get triggered.

PR #10688, opened 2022-09-30
Merge release branch 21.x into main

zhangskz:


mkruskal-google (approved):

zhangskz:
Note, package.xml's changelog only includes the new release from 21.x. May also append the 19.x and 20.x releases as well for completion, but TBD since it looks like historically we haven't bothered and lots of old versions are already missing.

We can either decide this doesn't matter, or commit to always merging these back.

Issue #10696, opened 2022-10-02
C# / Nested Enum / Reflection / EnumDescriptor


abbotware:
I was trying to get the enum descriptors from a nested enum, and it seems it fails due to this old code check:

https://github.com/protocolbuffers/protobuf/blob/488b8b91f81e6f7d50bd5f879ba429d50b0ba901/csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs#L329-L334

(7 years ago @jskeet )

If I step over that check, it works as expected - is there a reason this is even needed?  I would make a PR to remove those few lines, unless there is a good reason to keep them.

jskeet:
That code is *much* more than 7 years old - it goes back to [at least October 2008](https://github.com/jskeet/protobuf-csharp-port/blob/6803686bc06c4d96afd9bd2637f7b37a58596699/src/ProtocolBuffers/Descriptors/FileDescriptor.cs#L158) :)

I'll look into this in more detail later.

abbotware:
@jskeet  for an easy repro of the issue:

https://github.com/abbotware/EnumValueOptions-gRPC-Example/

change: ProfileEnum to a nested enum like

```
message ProfilesEnum {
enum Type  {
	A = 0 [(enum_name) = "AAA"];
	B = 1 [(enum_name) = "BBB"]; 
	C = 2 [(enum_name) = "CCC"];
}
};
```

this line returns null:

`EnumDescriptor ProfilesEnumDescriptor = ProfilesEnumReflection.Descriptor.FindTypeByName<EnumDescriptor>("ProfilesEnum.Type");`

jskeet:
Yup - I need to check what other languages do, and what other impacts this might have.

abbotware:
How is one supposed to get at this metadata at runtime?  I thought that was the whole point of reflection.  If you can demonstrate how to do it, that would be great... otherwise sounds like a bug consistent in 3 implementations :-)

abbotware:
that works for me!  I thought there was no way to get the metadata at all!

PR #10698, opened 2022-10-02
[C#] - Allow Retrieval of Nested Metadata

abbotware:
remove check to allow retrieval of nested enum metadata

fix for issue:
https://github.com/protocolbuffers/protobuf/issues/10696

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
+++ b/csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
@@ -326,7 +326,7 @@ namespace Google.Protobuf.Reflection
         public T FindTypeByName<T>(string name)
             where T : class, IDescriptor
         {
-            if (Package.Length > 0)
+            if (Package.Length > 0) 
             {
                 name = Package + "." + name;
             }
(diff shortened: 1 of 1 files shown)


abbotware:
FYI - All C# unit tests pass with this change

jskeet:
I'm going to have a look at the issue when I get the chance, later in the week. I'll be adding various tests, aside from anything else - but I really need to take some time over it to make sure I don't break anything.

I suggest you close this PR, as I'll create my own when I get the chance.

abbotware:
@jskeet - would creating additional method work as an alternative?  (see new commit) From what I can see, the only places this method is reference are unittests.  An additional method then won't break functionality for any library users

I just want access to the data that's already in there :-)   but alas, all the objects are sealed internal  :-/

jskeet:
That would certainly make sure it's not breaking, but changing the existing behaviour may be a better option if it brings it into line with other languages. Again, please be patient until I have time to look at this properly. (Bear in mind that it won't be released until the next release comes anyway.)

abbotware:
well, no rush - I can get access to the internal DescriptorPool via some dirty reflection tricks :-P 

just willing to help :-)

jskeet:
Closing this as the related issue is closed.

PR #10705, opened 2022-10-03
[java] Add option to JsonFormat for ignoring unresolved Any failures

KevinRamsunder:
This MR adds an additive option (`ignoringUnresolvedAnyTypes: Boolean = false`) to the `JsonFormat` class for ignoring unresolved `Any` types during serialization. Currently, JsonFormat will throw an `InvalidProtocolBufferException` whenever it attempts to serialize an `Any` type that has not been provided by a `TypeRegistry`. This is problematic because it forces consumers of `Any` types to stay updated with all of the proto definitions that upstream providers may produce. The added option allows consumers to choose to omit the serialized output altogether - instead of failing the entire serialization process. The property is set to false by default to preserve backwards compatibility.

Additionally, when set to `false`, the default behavior should, IMO, return a list of all Any fields that we were unable to serialize, so callers of the method can get an exhaustive list of failures instead of one at a time. I have not made these changes in this PR - but wanted to get the conversation started to see if there was any feedback / concerns with any of these ideas.

Thank you!

### DIFF
--- a/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
+++ b/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java
@@ -115,7 +115,8 @@ public class JsonFormat {
         /* preservingProtoFieldNames */ false,
         /* omittingInsignificantWhitespace */ false,
         /* printingEnumsAsInts */ false,
-        /* sortingMapKeys */ false);
+        /* sortingMapKeys */ false,
+        /* ignoringUnresolvedAnyTypes */ false);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


KevinRamsunder:
This should be tagged with a `java` label, but I don't think I have permission to do so unfortunately

mcy:
Hi Kevin,

This seems like a potentially useful option, but I think it would need a bit more design work to be in a good state. In particular I think we'd need to figure out:
- How can I tell that an Any that I couldn't figure out wasn't present?
- Should this be present in other languages, too?

There's a lot of unfortunate corner cases with Any and JSON at the moment that we'll need to sort out before we have a good idea of whether or not this is the correct design.

This intersects with some ongoing efforts to write a normative spec for the JSON codec, and the scope of this change probably means we can't easily merge it from an external contributor at this time, sorry. :(

Mupli:
Any news ? I would like to have the way to ignore or automatically resolve Any. 
Which is possible by getting classes. Sadly not all classes are in classpath and it would be nice to print them as simple Json structure. Or try to resolve with class and if it is not there then print standard Any. 

I need to have a way to see the content nicely formatted.

PR #10706, opened 2022-10-03
Add sanitized builds to Kokoro

mkruskal-google:
This will parametrize the Bazel kokoro presubmit, running all of optimized, debug, asan, msan, tsan, and ubsan.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -16,6 +16,13 @@ http_archive(
     ],
 )
 
+http_archive(
+    name = "com_googlesource_code_re2",
+    sha256 = "906d0df8ff48f8d3a00a808827f009a840190f404559f649cb8e4d7143255ef9",
+    strip_prefix = "re2-a276a8c738735a0fe45a6ee590fe2df69bcf4502",
+    urls = ["https://github.com/google/re2/archive/a276a8c738735a0fe45a6ee590fe2df69bcf4502.zip"],  # 2022-04-08
+)
+
 # Bazel platform rules.
 http_archive(
     name = "platforms",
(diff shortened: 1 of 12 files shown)


WORKSPACE
```diff
@@ -16,6 +16,13 @@ http_archive(
     ],
 )
 
+http_archive(
```
perezd, line 19: This should live in protobuf_deps.bzl under a maybe.

protobuf_deps.bzl
```diff
@@ -49,9 +49,9 @@ def protobuf_deps():
         http_archive(
             name = "zlib",
             build_file = "@com_google_protobuf//:third_party/zlib.BUILD",
-            sha256 = "629380c90a77b964d896ed37163f5c3a34f6e6d897311f1df2a7016355c45eff",
-            strip_prefix = "zlib-1.2.11",
-            urls = ["https://github.com/madler/zlib/archive/v1.2.11.tar.gz"],
+            sha256 = "d8688496ea40fb61787500e863cc63c9afcbc524468cedeb478068924eb54932",
```
perezd, line 52: was this change necessary in this PR? this just is a good thing to keep separate typically RE: rollbacks and cherrypicks etc.

perezd (changes requested):

protobuf_deps.bzl
mkruskal-google: Yep, this was necessary :). There was a UBSAN bug in zlib that was fixed with 1.2.12

WORKSPACE
mkruskal-google: I don't think it should.  This dependency is brought in by googletest, which we don't propagate to clients via protobuf_deps.  Note: *this* one is somewhat unrelated, I hit errors that were likely pre-existing due to this missing dependency.  I suspect the root cause is from when we upgraded googletest

perezd (approved):
LGTM

Issue #10716, opened 2022-10-04
[java] ByteString needs a serialVersionUID


stickfigure:
Version: latest (3.21.7 in particular, but it's in git)
Langauge: java

ByteString.java does not currently specify a serialVersionUID. This is wreaking havoc on countless places that include byte strings in their serialization format. For example, the google-cloud-datastore `Entity` class is serializable and frequently serialized into memcache-like services, but it composes ByteString so serialization breaks with different versions of the protobuf jar.

Some of the subclasses specify serialVersionUID which is great, but the base class needs to as well. Looks like an oversight.

PR #10718, opened 2022-10-04
Add serialVersionUID to ByteString and subclasses

stickfigure:
Addresses issue #10716

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/ByteString.java
+++ b/java/core/src/main/java/com/google/protobuf/ByteString.java
@@ -1606,8 +1606,6 @@ public abstract class ByteString implements Iterable<Byte>, Serializable {
   // Keep this class private to avoid deadlocks in classloading across threads as ByteString's
   // static initializer loads LiteralByteString and another thread loads BoundedByteString.
   private static final class BoundedByteString extends LiteralByteString {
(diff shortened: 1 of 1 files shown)


googleberg (approved):

java/core/src/main/java/com/google/protobuf/ByteString.java
```diff
@@ -1603,6 +1606,7 @@ protected int getOffsetIntoBytes() {
   // Keep this class private to avoid deadlocks in classloading across threads as ByteString's
   // static initializer loads LiteralByteString and another thread loads BoundedByteString.
   private static final class BoundedByteString extends LiteralByteString {
+    private static final long serialVersionUID = 1L;
```
googleberg, line 1609: This is a duplicate definition. I didn't notice in the earlier review that this was already defined below in the Serializable section.

googleberg (changes requested):

java/core/src/main/java/com/google/protobuf/ByteString.java
stickfigure: Ooops! Sorry, I missed that. Fixed.

googleberg (approved):

PR #10736, opened 2022-10-07
Upgrade to Kotlin 1.6

deannagarcia:
Upgrading Kotlin due to [this CVE](https://nvd.nist.gov/vuln/detail/CVE-2022-24329). This shouldn't affect our project, so we will not be pushing out a release immediately.

### DIFF
--- a/java/kotlin-lite/pom.xml
+++ b/java/kotlin-lite/pom.xml
@@ -16,7 +16,7 @@
   </description>
 
   <properties>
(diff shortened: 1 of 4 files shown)


perezd:
Is moving this a breaking change?

deannagarcia:
> Is moving this a breaking change?

Nope, Kotlin versions are backwards compatible as long as we don't add new language features.

perezd (approved):

Issue #7192, opened 2020-02-09
protoc doesn't allow some field names due to bad json_name logic


jhump:
**What version of protobuf and what language are you using?**
Version: 3.11.3
Language: none (using `--descriptorset_out`)

**What operating system (Linux, Windows, ...) and version?**
Mac OS X

**What runtime / compiler are you using (e.g., python version or gcc version)**
No runtime; just using the `protoc` binary

**What did you do?**
I tried to compile the following proto file:
```proto
// test.proto
syntax = "proto3";
message Foo {
  string abc = 1;
  string abc_def = 2;
  string __a__b__c = 3;
  string _abc_def = 4;
}
```
I used the following command:
```
protoc test.proto -otest.protoset
```

**What did you expect to see**
I expected this to work and to produce a `test.protoset` file.

If I remove the conflicting fields and generate this in isolation, the resulting `json_name` fields in the output descriptors are all different -- so there's no actual conflict. For reference:

* `abc` becomes JSON name of `abc`
* `abc_def` becomes JSON name of `abcDef`
* `__a__b__c` becomes JSON name of `ABC`
* `_abc_def` becomes JSON name of `AbcDef`

**What did you see instead?**
The command fails with the following error:
```
test.proto:6:10: The JSON camel-case name of field "__a__b__c" conflicts with field "abc". This is not allowed in proto3.
test.proto:7:10: The JSON camel-case name of field "_abc_def" conflicts with field "abc_def". This is not allowed in proto3.
```
It appears that it is using a case-insensitive comparison to determine if there is a conflict. But I believe that is wrong for a couple of reasons:
1. Key names in JSON are case-sensitive. So it is fine to have both "abc" and "ABC" in the same output, as well as both "abcDef" and "AbcDef".
2. The error message suggests the conflict is on the camel-case JSON name and doesn't say anything about being case insensitive.

jhump:
I think this is related to #5063, at least in the fact that they are both related to how `protoc` tries to check for uniqueness of JSON names.

In fact, adding a `json_name` option does not prevent the above errors. So if we add `json_name = "FOO"` to the `__a__b__c` field in the example above, it still complains with the same message:

```
test.proto:7:10: The JSON camel-case name of field "__a__b__c" conflicts with field "abc". This is not allowed in proto3.
```

jhump:
[...]

Issue #6175, opened 2019-05-24
protoc: code gen is different between using source files and using descriptor set


jhump:
**What version of protobuf and what language are you using?**
Version: v3.5.1

**What operating system (Linux, Windows, ...) and version?**
OS X

**What did you do?**
I created an extremely simple test file:
```proto
syntax = "proto3";

message Foo {
  uint64 id = 1;
  string name = 2;
  Bar bar = 3 [json_name="barre"];
}

enum Bar {
  VAL0 = 0;
  VAL1 = 1;
  VAL2 = 2;
}
```

I generated this to `Test.java` with the following command:

```bash
protoc test.proto --java_out=.
```

I then produced a descriptor set and used that to generate Java code again:

```bash
protoc test.proto -o test.protoset --include_source_info
protoc test.proto --descriptor_set_in test.protoset --java_out=.
```

**What did you expect to see**

I expected the output from the two code gen approaches to be the same. (And having contributed to this repo and knowing some of the code, I'm a bit surprised, since I know the compiler first parses into a descriptor and then does code gen based on the descriptor -- it's weird that the descriptor just parsed would differ from the descriptor de-serialized from file.)

**What did you see instead?**

Instead, it produced different Java output. When using a protoset, it _always_ emits comments that look as if the `json_name` option were always specified. But when using source, it only includes that comment when the `json_name` option _is actually_ specified.

This smells related to #5587.

jhump:
cc @pedgeio

pedgeio:
I believe the issue is that `json_name` is always populated when it is returned from `--descriptor_set_out`, but when `protoc` calls one of it's builtin plugins internally, it doesn't populate `json_name` unless it is explicitly set (this comment hints at this: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.cc#L6024).

While this seems minor, this is important - I should be able to do a roundtrip with `protoc`, ie generate directly using it, or generate from a pre-computed `FileDescriptorSet`, and have the same result.

Possible fixes:

- Always populate `json_name` for `FileDescriptorSets` passed to builtin plugins - this immediately fixes everything.
- Remove `json_name` when passing `FileDescriptorSets` given with `--descriptor_set_in` before passing them to builtin plugins if the value of `json_name` is the same as the default computed value for `json_name` for that field - this also would fix the builtin plugin issue.

[...]

Issue #5587, opened 2019-01-16
protoc should not populate `json_name` in descriptor if no `json_name` was specified


zellyn:
**What version of protobuf and what language are you using?**
Version: v3.6.0
Language: any

**What operating system (Linux, Windows, ...) and version?**
n/a

**What runtime / compiler are you using (e.g., python version or gcc version)**
n/a

**What did you do?**

test.proto:
```proto
syntax = "proto3";

package testing;

message Animal {
  string name = 1;
  bool has_hooves = 2;
  int32 birth_year = 3 [json_name = "year_of_birth"];
}
```

Command:
```shell
protoc -o /dev/stdout test.proto | protoc --decode=google.protobuf.FileDescriptorSet descriptor.proto
```

Output:
```
file {
  name: "test.proto"
  package: "testing"
  message_type {
    name: "Animal"
    field {
      name: "name"
      number: 1
      label: LABEL_OPTIONAL
      type: TYPE_STRING
      json_name: "name"
    }
    field {
      name: "has_hooves"
      number: 2
      label: LABEL_OPTIONAL
      type: TYPE_BOOL
      json_name: "hasHooves"
    }
    field {
      name: "birth_year"
      number: 3
      label: LABEL_OPTIONAL
      type: TYPE_INT32
      json_name: "year_of_birth"
    }
  }
  syntax: "proto3"
}
```

**What did you expect to see**
It should be possible to tell whether `json_name` was specified in the `.proto` file. In particular, there should be a way to distinguish that `json_name` was specified for `birth_year`, and not specified for `has_hooves`.

**What did you see instead?**
It is impossible to tell whether `json_name` was specified in the `.proto` file, because `protoc` invents a `json_name` for each field, regardless.

**Anything else**

1. This makes it impossible to create a json serializer that prefers the vanilla protobuf field names (with underscores, by convention), but lets individual fields be overridden. Instead, the default is the weird java-ish mangled version of field names. I'm fine with that being the default, but other options should be _possible_.  For instance, in Go's `jsonpb`, there is an `OrigName` option. But it's all-or-nothing. If you opt out of the java-ish names, you also opt out of being able to override a name.
2. This bloats the serialized descriptor. The language runtimes can manufacture default json field names just fine.
3. This makes it less possible to recreate something close to the original `.proto` file from the descriptor.

ObsidianMinor:
[...]

PR #10750, opened 2022-10-10
protoc: validate custom json_name configuration (take 2)

### LINKED ISSUE #5063: protoc: does not verify that json_name does not conflict

jhump:
This is a re-do of #10581. Merging this PR previously caused issues when importing the change into Google's internal repo (IIUC). So additional testing is needed before attempting to merge again. cc @mkruskal-google 

----

This makes a few other related changes, for consistency:

* Adds the existing checks, for default JSON names, to run even for proto2 files, but only as warnings (symmetry with similar check for enum value names).
* Tweaks the similar check for enum value names to include "This is not allowed in proto3", for clarity, since it's only a warning in proto2.

When checking custom JSON names, this considers it an error even in proto2 if two `json_name` options conflict. However, if a `json_name` option conflicts with a default JSON name, it is just a warning in proto2 (aligns with the existing check and the similar check for enum value names).

Fixes #5063.

### DIFF
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -2074,8 +2074,88 @@ TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
       "  uint32 foo = 1;\n"
       "  uint32 Foo = 2;\n"
       "}\n",
-      "3:9: The JSON camel-case name of field \"Foo\" conflicts with field "
-      "\"foo\". This is not allowed in proto3.\n");
+      "3:9: The default JSON name of field \"Foo\" (\"Foo\") conflicts "
+      "with the default JSON name of field \"foo\" (\"foo\"). "
+      "This is not allowed in proto3.\n");
+}
+
+TEST_F(ParserValidationErrorTest, Proto2JsonConflictError) {
+  // conflicts with default JSON names are not errors in proto2
+  ExpectParsesTo(
+      "syntax = 'proto2';\n"
+      "message TestMessage {\n"
+      "  optional uint32 foo = 1;\n"
+      "  optional uint32 Foo = 2;\n"
+      "}\n",
+
+      "syntax: 'proto2'"
+      "message_type {"
+      "  name: 'TestMessage'"
+      "  field {"
+      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1"
+      "  }"
+      "  field {"
+      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'Foo' number: 2"
+      "  }"
+      "}"
+      );
+}
+
+TEST_F(ParserValidationErrorTest, Proto3CustomJsonConflictWithDefaultError) {
+  ExpectHasValidationErrors(
+      "syntax = 'proto3';\n"
+      "message TestMessage {\n"
+      "  uint32 foo = 1 [json_name='bar'];\n"
+      "  uint32 bar = 2;\n"
+      "}\n",
+      "3:9: The default JSON name of field \"bar\" (\"bar\") conflicts "
+      "with the custom JSON name of field \"foo\". "
+      "This is not allowed in proto3.\n");
+}
+
+TEST_F(ParserValidationErrorTest, Proto2CustomJsonConflictWithDefaultError) {
+  // conflicts with default JSON names are not errors in proto2
+  ExpectParsesTo(
+      "syntax = 'proto2';\n"
+      "message TestMessage {\n"
+      "  optional uint32 foo = 1 [json_name='bar'];\n"
+      "  optional uint32 bar = 2;\n"
+      "}\n",
+
+      "syntax: 'proto2'"
+      "message_type {"
+      "  name: 'TestMessage'"
+      "  field {"
+      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1 json_name: 'bar'"
+      "  }"
+      "  field {"
+      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'bar' number: 2"
+      "  }"
+      "}"
+      );
+}
+
+TEST_F(ParserValidationErrorTest, Proto3CustomJsonConflictError) {
+  ExpectHasValidationErrors(
+      "syntax = 'proto3';\n"
+      "message TestMessage {\n"
+      "  uint32 foo = 1 [json_name='baz'];\n"
+      "  uint32 bar = 2 [json_name='baz'];\n"
+      "}\n",
+      "3:9: The custom JSON name of field \"bar\" (\"baz\") conflicts "
+      "with the custom JSON name of field \"foo\".\n");
+}
+
+TEST_F(ParserValidationErrorTest, Proto2CustomJsonConflictError) {
+  ExpectHasValidationErrors(
+      "syntax = 'proto2';\n"
+      "message TestMessage {\n"
+      "  optional uint32 foo = 1 [json_name='baz'];\n"
+      "  optional uint32 bar = 2 [json_name='baz'];\n"
+      "}\n",
+      // fails in proto2 also: can't explicitly configure bad custom JSON names
+      "3:18: The custom JSON name of field \"bar\" (\"baz\") conflicts "
+      "with the custom JSON name of field \"foo\".\n");
 }
 
 TEST_F(ParserValidationErrorTest, EnumNameError) {
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -2069,12 +2069,14 @@ TEST_F(ParserValidationErrorTest, Proto3Default) {
 
 TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto3';\n"
-      "message TestMessage {\n"
-      "  uint32 foo = 1;\n"
-      "  uint32 Foo = 2;\n"
-      "}\n",
-      "3:9: The default JSON name of field \"Foo\" (\"Foo\") conflicts "
+      R"pb(
+      syntax = 'proto3';
+      message TestMessage {
+        uint32 foo = 1;
+        uint32 Foo = 2;
+      }
+      )pb",
+      "4:15: The default JSON name of field \"Foo\" (\"Foo\") conflicts "
       "with the default JSON name of field \"foo\" (\"foo\"). "
       "This is not allowed in proto3.\n");
 }
@@ -2082,33 +2084,38 @@ TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
 TEST_F(ParserValidationErrorTest, Proto2JsonConflictError) {
   // conflicts with default JSON names are not errors in proto2
   ExpectParsesTo(
-      "syntax = 'proto2';\n"
-      "message TestMessage {\n"
-      "  optional uint32 foo = 1;\n"
-      "  optional uint32 Foo = 2;\n"
-      "}\n",
-
-      "syntax: 'proto2'"
-      "message_type {"
-      "  name: 'TestMessage'"
-      "  field {"
-      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1"
-      "  }"
-      "  field {"
-      "    label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'Foo' number: 2"
-      "  }"
-      "}"
+      R"pb(
+      syntax = "proto2";
+      message TestMessage {
+        optional uint32 foo = 1;
+        optional uint32 Foo = 2;
+      }
+      )pb",
+      R"pb(
+      syntax: 'proto2'
+      message_type {
+        name: 'TestMessage'
+        field {
+          label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'foo' number: 1
+        }
+        field {
+          label: LABEL_OPTIONAL type: TYPE_UINT32 name: 'Foo' number: 2
+        }
+      }
+      )pb"
       );
 }
 
 TEST_F(ParserValidationErrorTest, Proto3CustomJsonConflictWithDefaultError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto3';\n"
-      "message TestMessage {\n"
-      "  uint32 foo = 1 [json_name='bar'];\n"
-      "  uint32 bar = 2;\n"
-      "}\n",
-      "3:9: The default JSON name of field \"bar\" (\"bar\") conflicts "
+      R"pb(
+      syntax = 'proto3';
+      message TestMessage {
+        uint32 foo = 1 [json_name='bar'];
+        uint32 bar = 2;
+      }
+      )pb",
+      "4:15: The default JSON name of field \"bar\" (\"bar\") conflicts "
       "with the custom JSON name of field \"foo\". "
       "This is not allowed in proto3.\n");
 }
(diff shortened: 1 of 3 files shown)


mkruskal-google:
Ok so on the plus side, this *does* break out tests.  Let's try to get those green and then I can test internally

jhump:
@mkruskal-google, I guess you are referring to the red "Bazel" and various "Windows C++ ..." kokoro builds?

How do I troubleshoot these? The links for the failed checks do not provide me any details whatsoever as to what exactly failed. I get "Unable to retrieve invocation log". If I look at assets and try to download the build log, I get a 403 Forbidden error.

mkruskal-google:
Ah sorry, I guess the results are internal only >.<. I'll look into that, but in the meantime I'll summarize the issues and send them your way

mkruskal-google:
These tests fail on linux under `--config=asan`, and on windows for normal builds:
//src/google/protobuf:descriptor_unittest
//src/google/protobuf/compiler:parser_unittest

jhump:
@mkruskal-google, I don't suppose you could provide any advice on how I'd run linux tests with `--config=asan`? Is that a command-line argument to `bazel`?

Also, I don't suppose you can dig up any more details (like error message details for which assertion failed)?

If you can, thanks in advance.

src/google/protobuf/descriptor.cc
```diff
@@ -5487,6 +5497,84 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+    const std::string& message_name, const DescriptorProto& message,
+    FileDescriptor::Syntax syntax, bool use_custom_names) {
+
+  absl::flat_hash_map<std::string, JsonNameDetails> name_to_field;
+  for (const FieldDescriptorProto& field : message.field()) {
+    JsonNameDetails details = GetJsonNameDetails(&field, use_custom_names);
+    std::string lowercase_name = absl::AsciiStrToLower(details.orig_name);
+    auto existing = name_to_field.find(lowercase_name);
+    if (existing == name_to_field.end()) {
+      name_to_field[lowercase_name] = details;
```
fowles, line 5539: this is a double map lookup.

```
auto it_inserted = named_to_field.try_emplace(lowercase_name, details);
if (it_inserted.second) {
  continue;
}
auto& match = it.first->second;
```

is the pattern to use to avoid it

src/google/protobuf/descriptor.cc
```diff
@@ -5487,6 +5497,84 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+      // JSON names).
+      continue;
+    }
+    absl::string_view this_type = details.is_custom ? "custom" : "default";
+    absl::string_view existing_type = match.is_custom ? "custom" : "default";
+    // If the matched name differs (which it can only differ in case), include
+    // it in the error message, for maximum clarity to user.
+    absl::string_view name_suffix = "";
+    if (details.orig_name != match.orig_name) {
+      name_suffix = absl::StrCat(" (\"", match.orig_name, "\")");
```
fowles, line 5556: this is your bug right here.  Change the type of `name_suffix` to `std::string`.

src/google/protobuf/descriptor.cc
```diff
@@ -5890,13 +5978,12 @@ void DescriptorBuilder::CheckEnumValueUniqueness(
...
   PrefixRemover remover(result->name());
-  std::map<std::string, const EnumValueDescriptor*> values;
+  absl::flat_hash_map<std::string, const EnumValueDescriptor*> values;
   for (int i = 0; i < result->value_count(); i++) {
     const EnumValueDescriptor* value = result->value(i);
     std::string stripped =
         EnumValueToPascalCase(remover.MaybeRemove(value->name()));
-    std::pair<std::map<std::string, const EnumValueDescriptor*>::iterator, bool>
-        insert_result = values.insert(std::make_pair(stripped, value));
+    auto insert_result = values.insert(std::make_pair(stripped, value));
```
fowles, line 5986: `try_emplace(stripped, value)` is slightly nicer then `insert(make_pair`

mkruskal-google:
Here are you errors in windows:
```
[ RUN      ] ValidationErrorTest.ValidateProto3JsonName

src/google/protobuf/descriptor_unittest.cc(3935): error: Expected equality of these values:

  expected_errors

    Which is: "foo.proto: Foo: NAME: The default JSON name of field \"Name\" (\"Name\") conflicts with the default JSON name of field \"name\" (\"name\"). This is not allowed in proto3.\n"

  error_collector.text_

    Which is: "foo.proto: Foo: NAME: The default JSON name of field \"Name\" (\"Name\") conflicts with the default JSON name of field \"name\"\0(\"name\"). This is not allowed in proto3.\n"

src/google/protobuf/descriptor_unittest.cc(3935): error: Expected equality of these values:

  expected_errors

    Which is: "foo.proto: Foo: NAME: The default JSON name of field \"_a__b_\" (\"AB\") conflicts with the default JSON name of field \"ab\" (\"ab\"). This is not allowed in proto3.\n"

  error_collector.text_

    Which is: "foo.proto: Foo: NAME: The default JSON name of field \"_a__b_\" (\"AB\") conflicts with the default JSON name of field \"ab\"\0(\"ab\"). This is not allowed in proto3.\n"

[  FAILED  ] ValidationErrorTest.ValidateProto3JsonName (1 ms)
```
```
[ RUN      ] ParserValidationErrorTest.Proto3JsonConflictError

src/google/protobuf/compiler/parser_unittest.cc(187): error: Expected equality of these values:

  expected_errors

    Which is: "4:15: The default JSON name of field \"Foo\" (\"Foo\") conflicts with the default JSON name of field \"foo\" (\"foo\"). This is not allowed in proto3.\n"

  error_collector_.text_

    Which is: "4:15: The default JSON name of field \"Foo\" (\"Foo\") conflicts with the default JSON name of field \"foo\"\0(\"foo\"). This is not allowed in proto3.\n"

[  FAILED  ] ParserValidationErrorTest.Proto3JsonConflictError (0 ms)
```

As for the ASAN failures, Matt seems to have solved it by inspection :)

jhump:
I am hoping the same issue tickling ASAN errors may also be causing the Windows issues. I really have no idea how a NUL character is sneaking into that string where a space belongs, but it is exactly the same part of code as the ASAN issue, with construction of the `name_suffix` variable. Maybe changing the type will fix it there, too. 🤞

src/google/protobuf/descriptor.cc
jhump: done

src/google/protobuf/descriptor.cc
jhump: Ah, makes sense. Done.

jhump:
@mkruskal-google, looks like everything is green. 🎉 

Anything else I can do to help this get merged safely?

mkruskal-google:
Nope, I will try to import this into google and see if there's any uncaught issues.  Side note: what's the timeline on this look like?  If we can wait a few weeks this would be *much* easier to port than it is today

jhump:
> If we can wait a few weeks this would be much easier to port than it is today

@mkruskal-google, in my ideal world, this would get into the next release. So that question goes back to you: if we wait a few weeks, would it still be able to make it into the next release?

I'm trying to reconcile Buf's compiler with `protoc`. In most cases when there are differences, I update the compiler in Buf to match `protoc`. But in cases that seem like bugs in `protoc` (especially when there are issues filed against `protoc` where the team has suggested they're open to a fix, like this one), I prefer to update/fix in `protoc`.

In this case, we implemented some checks in Buf's compiler since it was indicated in #5063 that this functionality is acceptable for `protoc`, too. So now I want to get this into the next release so that the two compilers will behave the same.

fowles:
Yes, this is not at risk of missing the 22.x release

mkruskal-google:
Rebased this to kick our copybara service, which should allow for automated import of this PR

mkruskal-google:
So pulling this into google3, I immediately hit issues where we have existing protos with fields of the form "a_b_c" and "ab_c", which causes this conflict.  I don't have an estimate for the size of the problem yet, but I suspect it's very large given the speed at which I hit it.

I think the root issue here is that we're not getting the "custom" bit set properly.  The problematic protos *don't* have custom json options, but it's still showing up as an error rather than a warning

src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -2069,13 +2069,107 @@ TEST_F(ParserValidationErrorTest, Proto3Default) {
...
 TEST_F(ParserValidationErrorTest, Proto3JsonConflictError) {
   ExpectHasValidationErrors(
-      "syntax = 'proto3';\n"
-      "message TestMessage {\n"
-      "  uint32 foo = 1;\n"
-      "  uint32 Foo = 2;\n"
-      "}\n",
-      "3:9: The JSON camel-case name of field \"Foo\" conflicts with field "
-      "\"foo\". This is not allowed in proto3.\n");
+      R"pb(
```
mkruskal-google, line 2072: Please don't use R"pb()pb" for these.  We format those literals as *text protos*, not proto files and this is causing test failures due to reformatting.

jhump:
> so your proto2 change ends up being very loud

Okay. I can remove that warning. I had mainly added it for symmetry/consistency with the analogous warnings that are issued for enum value names. But if it's too noisy, I will drop that.

> edit2: There's a second component to the failures where fields w/o any custom json name are still showing up as "custom" in this code. I can't trace down why it's different than in OSS, but AFAICT the json_name field in FieldDescriptorProto will get set even for "default" fields...

Damn. I am willing to bet that it's related to issue #5587 (and more closely to #6175, which was closed as a duplicate of the former). If blaze, internally, generates descriptor sets as intermediate artifacts that get pulled into dependent targets, then the `json_name` field gets set in the emitted descriptor set, and the consumer has no way to distinguish between a custom-defined JSON name vs. a default one.

A possible remedy is to compare the `json_name` value to the default, computed name. Then we'd only consider it a custom name if they do not match. The downside is that the error message may not be quite accurate in the event that a user explicitly sets a custom JSON name that matches the default name for the field.

WDYT?

mkruskal-google:
Let's leave the warning in until the other issues are settled.  It's not entirely clear to me why warnings would be treated as errors, but it's hard to diagnose with all the other real errors showing up.

As for your workaround, I really like it!  I agree it's unfortunate that we'll lose coverage over an edge case, but we can add a TODO and link to that open issue in a comment.

jhump:
> that seems to have fixed all the issues in our tests, so we can keep the warnings

The rebase you pushed? Did it include any changes I should examine other than just merge resolutions? Or were the changes you're referring to just to some problem cases inside of google3 sources? Your prior comment suggested that it was hard to troubleshoot the warnings, so it's unclear to me how they were suddenly resolved.

FYI, I'm hoping to pick this back up later this week, to push the change mentioned above, about distinguishing custom vs. default JSON name.

mkruskal-google:
No sorry, I made the change we discussed on my internal clone of your PR.  I haven't touched your PR outside of the rebase.  I'm currently wrestling with a few cases that explicitly take advantage of our existing behavior (definitely Hyrum's law), but I'll keep you updated

mkruskal-google:
Ok so I've identified ~30 root causes for our internal test breakages.  All of the ones I looked at appear to be intentionally taking advantage of the old behavior.  A simplistic example is something like:
```
message Foo {
    oneof bar {
      string str = 2 [json_name = "bar"];
      int32 int = 3 [json_name = "bar"];
    }
}
```

Most of these appear to be proto3, so converting them to proto2 *might* be a possible mitigation here.  However, there's at least 1 that's already proto2 and is still failing due to the warning (they must have warnings configured as errors somewhere).  I'm thinking the easiest way forward would be to provide an escape hatch that I can migrate these folks to to recover the old behavior.  WDYT?

Note: I've made 2 changes internally for this analysis.  I changed all your `R"pb()pb"` lines to raw strings, and I treated json_names equal to the default as if they were unset (like you suggested)

jhump:
Also, do you mind pushing your local changes to this PR? Or do you want me to reproduce them and push them here? It sounds like maybe I should just hold off to await the outcome of whatever else you are trying in the internal codebase.

jhump:
@mkruskal-google, I just realized another validation check that should probably be done here, and it's one that should probably be an error even in proto2: the custom json name must _not_ be enclosed in square brackets, `[` and `]`.

At least in the Go runtime, a JSON field name that is enclosed in brackets is assumed to be an extension field, with a fully-qualified extension name between the brackets. So if a custom JSON name were formatted that way, it would not be unmarshalable. This should be an error in proto2, too, because only proto2 messages can even have extensions.

What do you think? Should I open another issue for this aspect? Or is it worth trying to work this additional check into this PR? If this new point is a distraction from the other stuff you are investigating on my behalf for this branch, no worries: I can just add it as a new issue and maybe look at adding it in a separate PR. (Hopefully this extra check will not pose any problems for existing sources.)

mkruskal-google:
Hey Joshua,

I believe your assumption is correct that these protos are only ever *serialized* to json, and never deserialized from it.  I can think of two options here:
1) Only error during at runtime during deserialization in these oneof cases
2) Provide an option to silence this error and restore the old behavior

As for the square bracket check, I did a search internally and there are no existing cases that have square brackets enclosing the entire name.  I'm fine with tacking that onto this PR, but note that there *are* cases that *use* square brackets (e.g. `json_name = "foo[bar]"`

jhump:
> Provide an option to silence this error and restore the old behavior

@mkruskal-google, do you mean a command-line option? If so, do you think that is something I can go ahead and add? Or does that change the scope of this PR (since it would impact the command-line interface) such that it needs to go through a more formal change proposal and review process?

What about if we just downgraded the error to warnings, for cases where the name conflicts with another field in the same oneof (maybe with a TODO to make this an error, possibly with some escape hatch, as part of a future edition?) Would that suffice?

mkruskal-google:
I was thinking a field or message option, but yea I think that might require more formal review.  Downgrading it to a warning would almost surely be fine though.  Let me give that a shot and see if there's a more manageable amount of fallout from that

mkruskal-google:
Actually there's still the issue of the old error you replaced too.  Replacing it with a stricter check was good, but if we're just replacing it with a warning we should probably restore it

jhump:
> Actually there's still the issue of the old error you replaced too.

I don't quite follow. What error was replaced? I think I just added new errors (related to custom JSON names) and lots of new warnings for proto2 related to JSON names.

mkruskal-google:
> I don't quite follow. What error was replaced? I think I just added new errors (related to custom JSON names) and lots of new warnings for proto2 related to JSON names.

You removed the "The JSON camel-case name of field..." error, which made sense when you were producing an error (since yours is a much more accurate one).

jhump:
@mkruskal-google, I pushed a change that relaxes it to a warning if the conflict is between two fields in the same oneof. I've also added the check mentioned above, about fields that are enclosed in brackets.

I'm still not quite sure what you meant for me to do re: "if we're just replacing it with a warning we should probably restore it". By "restore", are you requesting that I revert the language change in the message? Sorry if I'm being dense.

mkruskal-google:
Ah you're only downgrading it to a warning in the oneof case.  I like that, and I think this might satisfy all our internal constraints and still be an improvement over the old error you deleted.  Lemme rerun with those changes

src/google/protobuf/descriptor.cc
```diff
@@ -5487,6 +5497,112 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+    JsonNameDetails& match = it_inserted.first->second;
+    if (use_custom_names && !details.is_custom && !match.is_custom) {
+      // if this pass is considering custom JSON names, but neither of the
+      // names involved in the conflict are custom, don't bother with a
+      // message. That will have been reported from other pass (non-custom
+      // JSON names).
+      continue;
+    }
+    bool same_oneof = false;
+    if (use_custom_names && details.is_custom && match.is_custom
```
mkruskal-google, line 5566: I think either the is_custom check should be turned into an `or`.  It's very possible than the newer field has a custom json name to match the old, but they aren't both custom

mkruskal-google:
Ok so that doesn't clear up all the issues, but I think migrating many of them to oneof is a feasible option.  Lemme give that a shot to estimate the scale and see how doable it is

src/google/protobuf/descriptor.cc
jhump:
```suggestion
    if (use_custom_names && (details.is_custom || match.is_custom)
```

mkruskal-google:
Ok so I did an internal audit and I've found 4 classes of protos, with about ~15 files violating this error condition.

Deprecation (singular) - They're in the process of changing the type of a singular field, so they have two different versions mapping to the same json name.  We can either silence this by explicitly marking it `oneof` or we can change the json_name of the deprecated field.

Deprecation (repeated) - Same as the above, but where one of the fields is repeated.  `oneof` isn't an option here, so the only practical option is to change the json_name of the deprecated field.  This *should* be safe in principle, since if they're deserializing json with this field they already have UB today.

Intentional oneof (singular) - They're intentionally overloading the json_name to implicitly model oneof behavior (only enforced via comments).  These can likely be migrated to google.protobuf.Value, or if infeasible marked `oneof`

Intentional oneof (repeated) - Same as the above but where one of the fields is repeated.  Marking these `oneof` isn't an option, but google.protobuf.Value *should* be what they're using in this case.

I'll start migrating these and also see if we can remove the oneof exemption once that's done

jhump:
@mkruskal-google, wow, thank you so much for continuing to investigate and for your effort in resolving these issues on behalf of this change. I really appreciate it!

src/google/protobuf/descriptor.cc
```diff
@@ -5487,6 +5497,112 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
   }
 }
 
+void DescriptorBuilder::CheckFieldJsonNameUniqueness(
+    const DescriptorProto& proto, const Descriptor* result) {
+  FileDescriptor::Syntax syntax = result->file()->syntax();
+  std::string message_name = result->full_name();
+  // two passes: one looking only at default JSON names, and one that considers
+  // custom JSON names
+  CheckFieldJsonNameUniqueness(message_name, proto, syntax, false);
```
mkruskal-google, line 5506: What's the purpose of this check?  If custom json_names are set why do we still warn/error?  I would expect setting an explicit json_name to be a good mitigation to overlapping field names

jhump: I left that in place due to this comment from @acozzette on an issue I filed:
https://github.com/protocolbuffers/protobuf/issues/7192#issuecomment-695024065

And, indeed, the [custom JSON format](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#json-encoding-of-field-masks) for `FieldMask` simply converts the contained field names to camel-case names, without knowledge of the fields' custom names (e.g. field masks always refer to fields' _default_ JSON name).

I would personally _love_ to remove the check, since I agree that a custom JSON name seems like the best way to prevent the ambiguity it is complaining about. 🤷

mkruskal-google: Ah that makes sense, thanks for the explanation!  This might be a good thing to follow up on, since that error case is still fairly confused (it does say "default" vs "default" clashes, but it's not clear why that matters).  The specific case I hit was a single proto2 file that:

1) had two fields with clashing defaults
2) didn't care about JSON at all
3) was compiled in a way where warnings triggered errors

Would it make sense to just disable the warning in the non-custom branch?

jhump: Sure, disabling the warning for that case seems fair.

mkruskal-google: I think this may not be necessary.  I'm nearly green for your original change, which doesn't include the oneof exemption and warns in proto2

src/google/protobuf/descriptor.cc
```diff
@@ -5487,6 +5497,116 @@ void DescriptorBuilder::BuildMessage(const DescriptorProto& proto,
...
+      std::string error_message =
+          absl::StrFormat("The custom JSON name of field \"%s\" (\"%s\") is invalid: "
+                          "JSON names may not start with '[' and end with ']'.",
+                          field.name(), details.orig_name);
+      AddError(message_name, field, DescriptorPool::ErrorCollector::NAME,
+               error_message);
+      continue;
+    }
+
+    std::string lowercase_name = absl::AsciiStrToLower(details.orig_name);
```
mkruskal-google, line 5556: Why do we do a case-insensitive search here?  AFAIK json is case-sensitive, so this is creating clashes where there don't necessarily need to be any right?

jhump: I asked the same question when I opened this issue over two years ago: https://github.com/protocolbuffers/protobuf/issues/7192

I did not get a clear answer, so I didn't try to change that with this PR. The check for enum value names is also weirdly case-insensitive. 🤷 

I had assumed it was just to catch potentially confusing JSON names, that vary only in case. But, TBH (and I think I said this in that issue), that seems like a linter/style issue, not a compile issue. I'd be more than happy to make these checks more intuitive. They definitely feel overly broad as is today.

mkruskal-google: I'll follow up with @acozzette about this, but I think if we remove this case-insensitivity it'll fix most (if not all) of the remaining issues

mkruskal-google: Update: discussed this with @haberman and a case-sensitive comparison should be ok here.  As far as we can tell, this was made case-insensitive purely as a conservative measure.  I'm also investigating the fallout from taking this PR a step further and unconditionally erroring in both proto2 and proto3.  Will keep you updated

mkruskal-google:
I wrote up a [proposal](https://docs.google.com/document/d/1TsKvK9uU7q42-kVyVPzCTSEpJKpQ74rkrQz0UHCibiY/edit?resourcekey=0-XkzVi_BpzVsOCelkscBbpQ#) based on this PR.  The notable changes are that:
1) We do a case-sensitive comparison
2) We unify proto2 and proto3 behavior
3) We do some opportunistic tech debt cleanup that this unlocks

We'll invite you to the proposal meeting.  I'm hoping it can get approved as-is, but I think worst-case we can fallback to your implementation here (which is nearly safe to submit)

mkruskal-google:
Just merged into main!  Currently it's setup to only warn for proto2, I'm going to work on enabling errors there next

PR #10757, opened 2022-10-10
protoc: accept capital X to indicate hex escape in string literals

### LINKED ISSUE #10756: protoc: doesn't accept capital X in hex escapes in string literals
In the [specs on the developer site](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#string_literals), string literals can use `\X` (capital X) to initiate a hexadecimal byte literal:
```
hexEscape = '\' ( "x" | "X" ) hexDigit hexDigit
```

However, attempting to do so results in a syntax error:
```protobuf
// test.proto
syntax = "proto2";

message Foo {
  optional bytes bar = 1 [default = "\XBC"];
}
```
```
$ protoc test.proto -o /dev/null
test.proto:5:39: Invalid escape sequence in string literal.
```
Changing the string literal to `"\xBC"` (lower-case X) fixes the problem.

This appears to be a bug in the tokenizer, not in the documentation. I say this for a couple of reasons:
1. Both lower-case and capital X are accepted for hexadecimal integer literals.
2. Decoding the contents of `default_value` for a bytes field accepts hex escapes that use capital X.

To test the latter, I modified a file descriptor set so it included `"\\X74"` in a default value:
```
protoc test.proto -o /dev/stdout | xxd > test.protoset.txt
```
```
# edit the output of xxd (see third line, third word) to use capital X escape
00000000: 0a2b 0a0a 7465 7374 2e70 726f 746f 221d  .+..test.proto".
00000010: 0a03 466f 6f12 160a 0362 6172 1801 2001  ..Foo....bar.. .
00000020: 280c 3a04 5c58 3734 5203 6261 72         (.:.\X74R.bar
```
It turns out that `protoc` happily accepts this, even though it would complain about the escape if it appeared in a proto source file.
```
$ xxd -r < test.protoset.txt | protoc --descriptor_set_in /dev/stdin -o /dev/stdout test.proto | xxd
00000000: 0a28 0a0a 7465 7374 2e70 726f 746f 221a  .(..test.proto".
00000010: 0a03 466f 6f12 130a 0362 6172 1801 2001  ..Foo....bar.. .
00000020: 280c 3a01 7452 0362 6172                 (.:.tR.bar
```
As can be seen above, the `\X74` escape was accepted and translated to `t`.

jhump:
Fixes #10756.

Allows use of capital `X` to initiate a hex escape in a string literal. This is already documented to work on the developer site, but it didn't _actually_ work.

Note that `protoc` actually supports either one or two digits following the `\X` (or `\x`), but the documentation on the developer site suggests that two hex digits are always required. I think it's reasonable to assert that is a documentation bug and the lenience should be kept (especially since changing it would be a breaking change).

### DIFF
--- a/src/google/protobuf/descriptor_unittest.cc
+++ b/src/google/protobuf/descriptor_unittest.cc
@@ -2775,15 +2775,19 @@ TEST_F(MiscTest, DefaultValues) {
       ->set_default_value("hello");
   AddField(message_proto, "data", 9, label, FD::TYPE_BYTES)
       ->set_default_value("\\001\\002\\003");
+  AddField(message_proto, "data2", 10, label, FD::TYPE_BYTES)
+      ->set_default_value("\\X01\\X2\\X3");
+  AddField(message_proto, "data3", 11, label, FD::TYPE_BYTES)
+      ->set_default_value("\\x01\\x2\\x3");
 
   FieldDescriptorProto* enum_field =
-      AddField(message_proto, "enum", 10, label, FD::TYPE_ENUM);
+      AddField(message_proto, "enum", 12, label, FD::TYPE_ENUM);
   enum_field->set_type_name("DummyEnum");
   enum_field->set_default_value("B");
 
   // Strings are allowed to have empty defaults.  (At one point, due to
   // a bug, empty defaults for strings were rejected.  Oops.)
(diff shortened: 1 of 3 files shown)


jhump:
cc @fowles

fowles (approved):

PR #10764, opened 2022-10-11
[c++] Optimize Varint Codeing for c++

zzachimed:
optimize Varint Codeing by BMI2 instruction. In our funciton level benchmarking, we get about 2.7x  maximum performance
    boost.
2022-10-12T09:50:01+08:00
Running ./zz_bench
Run on (128 X 3300 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x64)
  L1 Instruction 32 KiB (x64)
  L2 Unified 1280 KiB (x64)
  L3 Unified 49152 KiB (x2)
Load Average: 0.20, 0.50, 0.47
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_Code_Varint32_1             3.13 ns         3.13 ns    223111504
BM_Code_BMI2_Varint32_1        3.17 ns         3.17 ns    220073339
BM_Code_Varint32_2             3.98 ns         3.98 ns    176472207
BM_Code_BMI2_Varint32_2        3.58 ns         3.58 ns    196546601
BM_Code_Varint32_3             5.86 ns         5.86 ns    119058237
BM_Code_BMI2_Varint32_3        4.42 ns         4.41 ns    158287666
BM_Code_Varint32_4             8.44 ns         8.43 ns     82981161
BM_Code_BMI2_Varint32_4        8.78 ns         8.77 ns     79598740
BM_Code_Varint32_5             10.7 ns         10.7 ns     65171356
BM_Code_BMI2_Varint32_5        7.73 ns         7.73 ns     90640481
BM_Code_Varint32_6             10.7 ns         10.7 ns     65157143
BM_Code_BMI2_Varint32_6        7.72 ns         7.71 ns     90778985
BM_Code_Varint32_7             10.7 ns         10.7 ns     65259369
BM_Code_BMI2_Varint32_7        7.73 ns         7.73 ns     90555467
BM_Code_Varint32_8             10.7 ns         10.7 ns     65215212
BM_Code_BMI2_Varint32_8        7.72 ns         7.72 ns     90833824
BM_Code_Varint32_9             10.7 ns         10.7 ns     65180833
BM_Code_BMI2_Varint32_9        7.73 ns         7.73 ns     90581863
BM_Code_Varint32_10            10.7 ns         10.7 ns     65177312
BM_Code_BMI2_Varint32_10       7.73 ns         7.72 ns     90671029
BM_Code_Varint64_1             3.20 ns         3.20 ns    218468799
BM_Code_BMI2_Varint64_1        3.18 ns         3.18 ns    220401104
BM_Code_Varint64_2             3.95 ns         3.95 ns    177792984
BM_Code_BMI2_Varint64_2        3.69 ns         3.69 ns    189878208
BM_Code_Varint64_3             6.04 ns         6.04 ns    114832150
BM_Code_BMI2_Varint64_3        4.26 ns         4.26 ns    164892413
BM_Code_Varint64_4             8.43 ns         8.43 ns     83087712
BM_Code_BMI2_Varint64_4        8.71 ns         8.71 ns     80103738
BM_Code_Varint64_5             10.7 ns         10.7 ns     65192875
BM_Code_BMI2_Varint64_5        8.68 ns         8.67 ns     80551592
BM_Code_Varint64_6             13.2 ns         13.2 ns     53063208
BM_Code_BMI2_Varint64_6        8.69 ns         8.68 ns     80423826
BM_Code_Varint64_7             15.8 ns         15.8 ns     44324556
BM_Code_BMI2_Varint64_7        8.55 ns         8.55 ns     82274517
BM_Code_Varint64_8             18.3 ns         18.3 ns     38198496
BM_Code_BMI2_Varint64_8        8.35 ns         8.35 ns     84413733
BM_Code_Varint64_9             20.8 ns         20.8 ns     33624867
BM_Code_BMI2_Varint64_9        7.75 ns         7.75 ns     90452908
BM_Code_Varint64_10            23.3 ns         23.3 ns     30076428
BM_Code_BMI2_Varint64_10       8.33 ns         8.33 ns     83756807

in our object level test. End to End benchmarking for object that all are varint fields ,we can get about 20% performance boost

### DIFF
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -114,6 +114,7 @@
 #include <atomic>
 #include <climits>
 #include <cstddef>
+#include <cstdint>
 #include <cstring>
 #include <limits>
 #include <string>
@@ -1048,15 +1049,9 @@ class PROTOBUF_EXPORT EpsCopyOutputStream {
   }
 #endif  /// __BMI2__ && __LZCNT__
 
-  PROTOBUF_ALWAYS_INLINE static uint8_t* UnsafeWriteSize(uint32_t value,
+  PROTOBUF_ALWAYS_INLINE  uint8_t* UnsafeWriteSize(uint32_t value,
                                                          uint8_t* ptr) {
-    while (PROTOBUF_PREDICT_FALSE(value >= 0x80)) {
-      *ptr = static_cast<uint8_t>(value | 0x80);
-      value >>= 7;
-      ++ptr;
-    }
-    *ptr++ = static_cast<uint8_t>(value);
-    return ptr;
+    return UnsafeVarint(value, ptr);
   }
 
   template <int S>
@@ -1295,6 +1290,10 @@ class PROTOBUF_EXPORT CodedOutputStream {
   void WriteVarint64(uint64_t value);
   // Like WriteVarint64()  but writing directly to the target array.
   static uint8_t* WriteVarint64ToArray(uint64_t value, uint8_t* target);
+  // Like WriteVarint*()  but writting directly to target only for
+  // ZeroCopyOutputStream
+  template<typename T>
+  static uint8_t* WriteVarintToArrayStream(T value, uint8_t* target);
 
   // Equivalent to WriteVarint32() except when the value is negative,
   // in which case it must be sign-extended to a full 10 bytes.
@@ -1772,11 +1771,7 @@ inline void CodedOutputStream::InitEagerly(Stream* stream) {
 
 inline uint8_t* CodedOutputStream::WriteVarint32ToArray(uint32_t value,
                                                         uint8_t* target) {
-#if defined(__BMI2__) && defined(__LZCNT__)
-    return EpsCopyOutputStream::UnsafeVarintBMI2(value, target);
-#else
-    return EpsCopyOutputStream::UnsafeVarintNoneBMI2(value, target);
-#endif
+  return EpsCopyOutputStream::UnsafeVarintNoneBMI2(value, target);
 }
 
 inline uint8_t* CodedOutputStream::WriteVarint32ToArrayOutOfLine(
@@ -1791,6 +1786,13 @@ inline uint8_t* CodedOutputStream::WriteVarint32ToArrayOutOfLine(
 
 inline uint8_t* CodedOutputStream::WriteVarint64ToArray(uint64_t value,
                                                         uint8_t* target) {
+  return EpsCopyOutputStream::UnsafeVarintNoneBMI2(value, target);
+}
+
+template <typename T>
+inline uint8_t* CodedOutputStream::WriteVarintToArrayStream(T value, uint8_t *target) {
+  static_assert(std::is_unsigned<T>::value,
+                  "Varint serialization must be unsigned");
 #if defined(__BMI2__) && defined(__LZCNT__)
     return EpsCopyOutputStream::UnsafeVarintBMI2(value, target);
 #else
@@ -1844,12 +1846,12 @@ inline uint8_t* CodedOutputStream::WriteLittleEndian64ToArray(uint64_t value,
 
 inline void CodedOutputStream::WriteVarint32(uint32_t value) {
   cur_ = impl_.EnsureSpace(cur_);
-  SetCur(WriteVarint32ToArray(value, Cur()));
+  SetCur(WriteVarintToArrayStream(value, Cur()));
 }
 
 inline void CodedOutputStream::WriteVarint64(uint64_t value) {
   cur_ = impl_.EnsureSpace(cur_);
-  SetCur(WriteVarint64ToArray(value, Cur()));
+  SetCur(WriteVarintToArrayStream(value, Cur()));
 }
 
 inline void CodedOutputStream::WriteTag(uint32_t value) {
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -634,12 +634,6 @@ class PROTOBUF_EXPORT CodedInputStream {
 // use of this class, ie. keep ptr a local variable, eliminates the need to
 // for the compiler to sync the ptr value between register and memory.
 class PROTOBUF_EXPORT EpsCopyOutputStream {
-#if defined(__BMI2__) && defined(__LZCNT__)
-#define UnsafeVarint UnsafeVarintBMI2
-#else
-#define UnsafeVarint UnsafeVarintNoneBMI2
-#endif
-
  public:
   enum { kSlopBytes = 16 };
 
@@ -898,6 +892,18 @@ class PROTOBUF_EXPORT EpsCopyOutputStream {
   }
   static uint64_t ZigZagEncode64(int64_t v) {
     return (static_cast<uint64_t>(v) << 1) ^ static_cast<uint64_t>(v >> 63);
+  }
+
+  template <typename T>
+  PROTOBUF_ALWAYS_INLINE uint8_t* UnsafeVarint(T value, uint8_t* ptr) {
+    if(!stream_)
+      return UnsafeVarintNoneBMI2(value, ptr);
+
+#if defined(__BMI2__) && defined(__LZCNT__)
+    return UnsafeVarintBMI2(value, ptr);
+#else
+    return UnsafeVarintNoneBMI2(value, ptr);
+#endif
   }
 
   template <typename T>
@@ -1766,7 +1772,11 @@ inline void CodedOutputStream::InitEagerly(Stream* stream) {
 
 inline uint8_t* CodedOutputStream::WriteVarint32ToArray(uint32_t value,
                                                         uint8_t* target) {
-  return EpsCopyOutputStream::UnsafeVarint(value, target);
+#if defined(__BMI2__) && defined(__LZCNT__)
+    return EpsCopyOutputStream::UnsafeVarintBMI2(value, target);
+#else
+    return EpsCopyOutputStream::UnsafeVarintNoneBMI2(value, target);
+#endif
 }
 
 inline uint8_t* CodedOutputStream::WriteVarint32ToArrayOutOfLine(
@@ -1781,7 +1791,11 @@ inline uint8_t* CodedOutputStream::WriteVarint32ToArrayOutOfLine(
 
 inline uint8_t* CodedOutputStream::WriteVarint64ToArray(uint64_t value,
                                                         uint8_t* target) {
-  return EpsCopyOutputStream::UnsafeVarint(value, target);
+#if defined(__BMI2__) && defined(__LZCNT__)
+    return EpsCopyOutputStream::UnsafeVarintBMI2(value, target);
+#else
+    return EpsCopyOutputStream::UnsafeVarintNoneBMI2(value, target);
+#endif
 }
 
 inline void CodedOutputStream::WriteVarint32SignExtended(int32_t value) {
--- a/src/google/protobuf/io/coded_stream.h
+++ b/src/google/protobuf/io/coded_stream.h
@@ -634,6 +634,12 @@ class PROTOBUF_EXPORT CodedInputStream {
 // use of this class, ie. keep ptr a local variable, eliminates the need to
 // for the compiler to sync the ptr value between register and memory.
 class PROTOBUF_EXPORT EpsCopyOutputStream {
+#if defined(__BMI2__) && defined(__LZCNT__)
+#define UnsafeVarint UnsafeVarintBMI2
+#else
+#define UnsafeVarint UnsafeVarintNoneBMI2
+#endif
+
  public:
   enum { kSlopBytes = 16 };
 
@@ -895,7 +901,7 @@ class PROTOBUF_EXPORT EpsCopyOutputStream {
   }
 
   template <typename T>
-  PROTOBUF_ALWAYS_INLINE static uint8_t* UnsafeVarint(T value, uint8_t* ptr) {
+  PROTOBUF_ALWAYS_INLINE static uint8_t* UnsafeVarintNoneBMI2(T value, uint8_t* ptr) {
     static_assert(std::is_unsigned<T>::value,
                   "Varint serialization must be unsigned");
     ptr[0] = static_cast<uint8_t>(value);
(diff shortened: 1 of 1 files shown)


jorgbrown:
Thanks for taking the time to look into this, and, like last time, special thanks for taking the time to do some benchmarks.

Alas, similar to the effort to make parsing faster, you have similar caveats here: the benchmarks can be misleading because they're not tracking random data, and PDEP/PEXT can be much slower than the equivalent integer instructions on Zen and Zen2 CPUs.

But the far bigger concern with this CL is this part of the code:

>    uint64_t* pdata = reinterpret_cast<uint64_t*>(ptr);
>    if (efftiveBitCounts <= 56) {
>      *pdata = _pdep_u64(value, mask);

Here, you're writing 8 bytes of data to the output stream, even if you need fewer.  This is a problem because unlike the input streams, there is no guaranteed buffer area on the output stream.  We haven't found it to be significantly beneficial to allocate more buffer area at the very end of the output stream, so we don't.  You have to take care not to write any more data than is actually needed; if you write more, you may be overwriting some other buffer!

zzachimed:
@jorgbrown 👍
verry thanks and appreciate for your carefuly comments . sorry for latter response for current busy work.

1 for your concern about avx512 issue
we disucssed with intel senior principal architect Moga, Adrian C [[adrian.c.moga@intel.com](mailto:adrian.c.moga@intel.com)]. This is his comment about"
Intel can slow down the clock rate of every core on a chip just because one of them started using AVX-512"
"I don’t this this is true.
First, the article is old and based on a SKX 12C SKU.

Second, the Cloudflare blog writer is confused by wikichip reference, which refers to N cores using the same license.
Nothing in there says that one core doing AVX-512 brings down the other cores to base frequency if the run SSE.

Also, if I understand his experiment correctly, he is running a fully-loaded system with two deployments.
So, he’s observing a noisy neighbor effect for frequency under TDP constraints.
If some cores use more power with AVX-512, the other cores may also see a frequency degradation.
I think even that may not be true starting with SPR, as the cores have autonomous power management.
But I am not 100% certain. I will ask and let you know.

Thanks!
Adrian
"
2 for your concern about run-out memory issue "
But the far bigger concern with this CL is this part of the code:

uint64_t* pdata = reinterpret_cast<uint64_t*>(ptr);
if (efftiveBitCounts <= 56) {
*pdata = _pdep_u64(value, mask);

Here, you're writing 8 bytes of data to the output stream, 
":
current stream mechinism can additional bytes for our processing . our decode process varintparse is private function and server for up message level depend on the buffer from streaming processing. In sepecifct words the buffer is from EpsCopyOutputStream:: EnsureSpace processing(every object encode start it will call this func). Maybe you could deep dive in this process you will found some wondeful mechnism to prevent memory run out in some range .In this buffer size check and buffer swap processing ,there has some woderful mechinisim to processing some visit out of range issue --at most additional 16 bytes (kSlopBytes)for run-out processing(in EpsCopyOutputStreamclass ) . It afford additional in "char buffer_[2 * kSlopBytes] = {};" . The specific processing is comlicated and hard to description very shortly. So its better to check that ParseContext::Done code line by line.

zzachimed:
@jorgbrown 👍
verry thanks and appreciate for your carefuly comments . sorry for latter response for current busy work.
1     for your concern about pext execution effiency issue in AMD platform “PDEP/PEXT can be much slower than the equivalent integer instructions on Zen and Zen2 CPUs.”:

       It looks like Zen3 will have some improvement regarding to this.


2    for your concern about run-out memory issue ":
But the far bigger concern with this CL is this part of the code:
uint64_t* pdata = reinterpret_cast<uint64_t*>(ptr);
if (efftiveBitCounts <= 56) {
*pdata = _pdep_u64(value, mask);
Here, you're writing 8 bytes of data to the output stream,
":
                current stream mechinism can additional bytes for our processing . our encode process UnsafeVarint is private function and server for up message level depend on the buffer from streaming processing. In sepecifct words the buffer is from EpsCopyOutputStream:: EnsureSpace processing(every object encode start it will call this func). Maybe you could deep dive in this process you will found some wondeful mechnism to prevent memory run out in some range .In this buffer size check and buffer swap processing ,there has some woderful mechinisim to processing some visit out of range issue --at most additional 16 bytes (kSlopBytes)for run-out processing(in EpsCopyOutputStreamclass ) . It afford additional space in "char buffer_[2 * kSlopBytes] = {};" . The specific processing is comlicated and hard to description very shortly. So its better to check that ParseContext::Done code line by line.

jorgbrown:
It's true that EpsCopyOutputStream has the capability to use a separate buffer, which would allow you to use your faster BMI routines.

Unfortunately it also has this constructor:
```
// Only for array serialization. No overflow protection, end_ will be the
// pointed to the end of the array. When using this the total size is already
// known, so no need to maintain the slop region.
EpsCopyOutputStream(void* data, int size, bool deterministic)
    : end_(static_cast<uint8_t*>(data) + size),
      buffer_end_(nullptr),
      stream_(nullptr),
      is_serialization_deterministic_(deterministic) {}
```
This is one we use quite often at Google, and I believe is the one used when serializing directly into a std::string.  And since it does not maintain a slop buffer, you can't write even one byte more than you actually need to.

jorgbrown:
I'd be curious how your benchmarks come out if you use this for your PDEP; it works on all CPUs:
```
inline uint64_t handPDEP_7F7F7F7F7F7F7F7F(uint64_t value) {
  // 1. Reshape value into two halves, each of which have 28 bits
#if defined(__aarch64__)
  uint64_t top28 = value & 0x00FFFFFFF0000000;
  uint64_t bot28 = value & 0x000000000FFFFFFF;
  value = (top28 << 4) + bot28;
#else
  value <<= 4;
  asm("" : "+r"(value));  // Improve clang
  value = (value & 0x0FFFFFFF00000000) | (static_cast<uint32_t>(value) >> 4);
#endif
  // 2. Reshape value into four quarters, each of which have 14 bits
  uint64_t top14s = value & 0xFFFFC000'0FFFC000;
  uint64_t bot14s = value - top14s;
  value = bot14s + (top14s << 2);
  // 3. Reshape value into eight bytes, each of which have 7 bits
  uint64_t top7s = value & 0x3F80'3F80'3F80'3F80;
  value += top7s;
  return value;
}
```

zzachimed:
> Google



> It's true that EpsCopyOutputStream has the capability to use a separate buffer, which would allow you to use your faster BMI routines.
> 
> Unfortunately it also has this constructor:
> 
> ```
> // Only for array serialization. No overflow protection, end_ will be the
> // pointed to the end of the array. When using this the total size is already
> // known, so no need to maintain the slop region.
> EpsCopyOutputStream(void* data, int size, bool deterministic)
>     : end_(static_cast<uint8_t*>(data) + size),
>       buffer_end_(nullptr),
>       stream_(nullptr),
>       is_serialization_deterministic_(deterministic) {}
> ```
> 
> This is one we use quite often at Google, and I believe is the one used when serializing directly into a std::string. And since it does not maintain a slop buffer, you can't write even one byte more than you actually need to.

hi brog: 

> Thanks for taking the time to look into this, and, like last time, special thanks for taking the time to do some benchmarks.
> 
> Alas, similar to the effort to make parsing faster, you have similar caveats here: the benchmarks can be misleading because they're not tracking random data, and PDEP/PEXT can be much slower than the equivalent integer instructions on Zen and Zen2 CPUs.
> 
> But the far bigger concern with this CL is this part of the code:
> 
> > uint64_t* pdata = reinterpret_cast<uint64_t*>(ptr);
> > if (efftiveBitCounts <= 56) {
> > *pdata = _pdep_u64(value, mask);
> 
> Here, you're writing 8 bytes of data to the output stream, even if you need fewer. This is a problem because unlike the input streams, there is no guaranteed buffer area on the output stream. We haven't found it to be significantly beneficial to allocate more buffer area at the very end of the output stream, so we don't. You have to take care not to write any more data than is actually needed; if you write more, you may be overwriting some other buffer!



> I'd be curious how your benchmarks come out if you use this for your PDEP; it works on all CPUs:
> 
> ```
> inline uint64_t handPDEP_7F7F7F7F7F7F7F7F(uint64_t value) {
>   // 1. Reshape value into two halves, each of which have 28 bits
> #if defined(__aarch64__)
>   uint64_t top28 = value & 0x00FFFFFFF0000000;
>   uint64_t bot28 = value & 0x000000000FFFFFFF;
>   value = (top28 << 4) + bot28;
> #else
>   value <<= 4;
>   asm("" : "+r"(value));  // Improve clang
>   value = (value & 0x0FFFFFFF00000000) | (static_cast<uint32_t>(value) >> 4);
> #endif
>   // 2. Reshape value into four quarters, each of which have 14 bits
>   uint64_t top14s = value & 0xFFFFC000'0FFFC000;
>   uint64_t bot14s = value - top14s;
>   value = bot14s + (top14s << 2);
>   // 3. Reshape value into eight bytes, each of which have 7 bits
>   uint64_t top7s = value & 0x3F80'3F80'3F80'3F80;
>   value += top7s;
>   return value;
> }
> ```

okay ,intresting code . I will bench this code and send your result .

zzachimed:
> It's true that EpsCopyOutputStream has the capability to use a separate buffer, which would allow you to use your faster BMI routines.
> 
> Unfortunately it also has this constructor:
> 
> ```
> // Only for array serialization. No overflow protection, end_ will be the
> // pointed to the end of the array. When using this the total size is already
> // known, so no need to maintain the slop region.
> EpsCopyOutputStream(void* data, int size, bool deterministic)
>     : end_(static_cast<uint8_t*>(data) + size),
>       buffer_end_(nullptr),
>       stream_(nullptr),
>       is_serialization_deterministic_(deterministic) {}
> ```
> 
> This is one we use quite often at Google, and I believe is the one used when serializing directly into a std::string. And since it does not maintain a slop buffer, you can't write even one byte more than you actually need to.

First and first: happy new year :) . Maybe we find out some method that  if use this constructor , we can use  none optimization function. I will try it.

zzachimed:
> 
hi :
 i already test this code in our benchmark(for varint64). it seems the maxisum peformance for this manualy pdep is 2.0x peformance boost. our pdep is 3.0x peformace boost . this is benchmark result : 
root@icx-14 /home/zhaozhou/protobuf/debug (avx_serialize_optimize)
$ numactl -C0 ./zz_bench
2023-01-20T01:07:39+08:00
Running ./zz_bench
Run on (128 X 3300 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x64)
  L1 Instruction 32 KiB (x64)
  L2 Unified 1280 KiB (x64)
  L3 Unified 49152 KiB (x2)
Load Average: 0.07, 0.15, 0.30
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
BM_Code_Varint64_1                      3.19 ns         3.19 ns    219087480
BM_Code_HandlePDEP7Fs_Varint64_1        3.18 ns         3.17 ns    220564861
BM_Code_BMI2_Varint64_1                 3.19 ns         3.18 ns    220183632
BM_Code_Varint64_2                      3.95 ns         3.94 ns    179381031
BM_Code_HandlePDEP7Fs_Varint64_2        3.59 ns         3.58 ns    195887374
BM_Code_BMI2_Varint64_2                 3.68 ns         3.68 ns    190235238
BM_Code_Varint64_3                      6.03 ns         6.03 ns    116418546
BM_Code_HandlePDEP7Fs_Varint64_3        4.43 ns         4.43 ns    158354401
BM_Code_BMI2_Varint64_3                 4.32 ns         4.32 ns    161751959
BM_Code_Varint64_4                      8.43 ns         8.42 ns     83102081
BM_Code_HandlePDEP7Fs_Varint64_4        12.8 ns         12.8 ns     54641861
BM_Code_BMI2_Varint64_4                 8.73 ns         8.72 ns     80476584
BM_Code_Varint64_5                      10.7 ns         10.7 ns     65222627
BM_Code_HandlePDEP7Fs_Varint64_5        12.9 ns         12.9 ns     54364384
BM_Code_BMI2_Varint64_5                 8.71 ns         8.70 ns     80458090
BM_Code_Varint64_6                      13.2 ns         13.2 ns     53080268
BM_Code_HandlePDEP7Fs_Varint64_6        12.9 ns         12.9 ns     54340507
BM_Code_BMI2_Varint64_6                 8.61 ns         8.61 ns     80468164
BM_Code_Varint64_7                      15.8 ns         15.8 ns     44278968
BM_Code_HandlePDEP7Fs_Varint64_7        12.9 ns         12.9 ns     54247917
BM_Code_BMI2_Varint64_7                 8.64 ns         8.63 ns     80797400
BM_Code_Varint64_8                      18.3 ns         18.3 ns     38186321
BM_Code_HandlePDEP7Fs_Varint64_8        12.9 ns         12.9 ns     54300176
BM_Code_BMI2_Varint64_8                 8.61 ns         8.60 ns     81317987
BM_Code_Varint64_9                      20.8 ns         20.8 ns     33600812
BM_Code_HandlePDEP7Fs_Varint64_9        13.0 ns         12.9 ns     54084671
BM_Code_BMI2_Varint64_9                 7.76 ns         7.75 ns     90315240
BM_Code_Varint64_10                     23.3 ns         23.3 ns     30065172
BM_Code_HandlePDEP7Fs_Varint64_10       13.8 ns         13.8 ns     50593565
BM_Code_BMI2_Varint64_10                8.44 ns         8.43 ns     82778429

"BM_Code_Varint64" is origin code.  "BM_Code_HandlePDEP7Fs_Varint64" is code include what your wanted . "BM_Code_HandlePDEP7Fs_Varint64" is code that my optimization use intel intrinc pdep.



code snipet:
  PROTOBUF_ALWAYS_INLINE static uint8_t* UnsafeVarint64HandlePDEP7Fs(uint64_t value, uint8_t* ptr){
    size_t efftiveBitCounts = 64 - _lzcnt_u64(value);
    auto numberOf7Bits=Number7Bits[efftiveBitCounts];

    constexpr const uint64_t mask=0x7f7f7f7f7f7f7f7f;
    uint64_t *pdata=reinterpret_cast<uint64_t*>(ptr);
    if( efftiveBitCounts<=56 ){
      //*pdata= _pdep_u64 (value,mask);
      *pdata= handPDEP_7F7F7F7F7F7F7F7F(value);
      *pdata|=ExtendedBitsMask[numberOf7Bits];
      ptr+=numberOf7Bits;
      return ptr;
    }

    //*pdata= _pdep_u64 (value & 0xffffffffffffff,mask);
    *pdata=handPDEP_7F7F7F7F7F7F7F7F(value & 0xffffffffffffff);
    *pdata|=0x8080808080808080;
    ptr+=8;
    if(value< (1ull<<63) ){
      ptr[0]=static_cast<uint8_t>(value>>56);
      return ptr+1;
    }
    ptr[0]=static_cast<uint8_t>(value>>56);
    ptr[0]|=0x80;
    ptr[1]=1;
    return ptr+2;

  }

zzachimed:
> It's true that EpsCopyOutputStream has the capability to use a separate buffer, which would allow you to use your faster BMI routines.
> 
> Unfortunately it also has this constructor:
> 
> ```
> // Only for array serialization. No overflow protection, end_ will be the
> // pointed to the end of the array. When using this the total size is already
> // known, so no need to maintain the slop region.
> EpsCopyOutputStream(void* data, int size, bool deterministic)
>     : end_(static_cast<uint8_t*>(data) + size),
>       buffer_end_(nullptr),
>       stream_(nullptr),
>       is_serialization_deterministic_(deterministic) {}
> ```
> 
> This is one we use quite often at Google, and I believe is the one used when serializing directly into a std::string. And since it does not maintain a slop buffer, you can't write even one byte more than you actually need to.

@jorgbrown  
  hi i aready commit the new changes to deal with this issue. when constructor not use the zerooutputstream for input ,the code will not use our optimized version code. plz check it . verrry thanks for your carefully review !!

danlark1:
Hi, first of all, thanks for the patch.

We decided not to move forward with that patch directly as AMD Zen and Zen 2 have horrible pdep/pext latency. This impacts google a lot as well as industry who adopt new code. We are aware of this issue and what we are going to do on our end is profile driven varint where we fallback to pdep/pext for intel machines and fallback implementation with shld we have now for AMD ones. However, this requires a bit more code generation and integration with our systems.

For now we are going to put this on hold, sorry about that

PR #10767, opened 2022-10-11
Add remote caching to bazel builds

mkruskal-google:
This enables remote caching for all of our Bazel builds.  This will have the largest impact on reruns, but should hopefully speed up all our CI builds.

Note: each job is given its own cache to avoid issues from non-hermetic builds.  For example, if the version of python changes our current setup doesn't give Bazel enough knowledge to know the cache is stale.  Ruby, PHP, C#, and Objective-c likely have this issue as well, since they all use non-hermetic genrules to grab system dependencies.  Once all of these have been fully modified we can likely fall back to a single shared cache


kokoro/common/bazel_flags.sh
```diff
@@ -0,0 +1,46 @@
...
+  if [[ -n ${KOKORO_BAZEL_AUTH_CREDENTIAL:-} ]]; then
+    _flags+=( --google_credentials=${KOKORO_BAZEL_AUTH_CREDENTIAL} )
+  else
+    _flags+=( --google_default_credentials=true )
+  fi
+
+  echo "${_flags[@]}"
+}
+
+echo $(bazel_flags::kokoro_flags) --keep_going --test_output=errors
```
devjgm, line 46: May want to use `echo -- ...` so that none of the kokoro flags accidentally get interpreted as flags to echo.

mkruskal-google: Added quotes instead, that should fix it :)

PR #10770, opened 2022-10-12
Mark default instance as immutable first to avoid race during static initialization of default instances.

googleberg:
Mark default instance as immutable first to avoid race during static initialization of default instances. Previously, the instance could be published before getting marked immutable.


mkruskal-google (approved):
Remove the cl # from the description :)

Also, I think the PR title is what ends up in the changelog, so you might want a little more detail

googleberg:
Done. It's funny, github gave me a "pro-tip" to shorten the PR title.

PR #10773, opened 2022-10-14
Do not force C++14

mumbleskates:
Rather than forcing protobuf to always compile in exactly C++14, overriding all settings in .bazelrc and environment variables and commandline options, ~~instead guard the codebase against versions that are too low~~ simply don't do that. This allows for compiling in higher C++ standards, such as those that have `std::string_view` instead of `absl::string_view` without generating objects that are incompatible.

### DIFF
--- a/src/google/protobuf/stubs/common.h
+++ b/src/google/protobuf/stubs/common.h
@@ -53,9 +53,16 @@
 #error "C++ versions less than C++14 are not supported."
 #endif  // _MSVC_LANG < 201402L
 #elif defined(__cplusplus)
+// Special-case GCC < 5.0, as it has a strange __cplusplus value for C++14
+#if defined(__GNUC__) && __GNUC__ < 5
+#if __cplusplus < 201300L
+#error "C++ versions less than C++14 are not supported."
+#endif  // __cplusplus < 201300L
+#else // defined(__GNUC__) && __GNUC__ < 5
 #if __cplusplus < 201402L
 #error "C++ versions less than C++14 are not supported."
 #endif  // __cplusplus < 201402L
+#endif // defined(__GNUC__) && __GNUC__ < 5
 #endif
 
 #ifndef PROTOBUF_USE_EXCEPTIONS
--- a/src/google/protobuf/stubs/common.h
+++ b/src/google/protobuf/stubs/common.h
@@ -47,10 +47,6 @@
 #include "google/protobuf/stubs/platform_macros.h"
 #include "google/protobuf/stubs/port.h"
 
-#if !defined(PROTOBUF_FORCE_OLD_CPP_VERSION_WE_KNOW_WHAT_WE_ARE_DOING) && __cplusplus < 201402L
-#error "Protobuf requires at least C++14"
-#endif
-
 #ifndef PROTOBUF_USE_EXCEPTIONS
 #if defined(_MSC_VER) && defined(_CPPUNWIND)
   #define PROTOBUF_USE_EXCEPTIONS 1
--- a/src/google/protobuf/stubs/common.h
+++ b/src/google/protobuf/stubs/common.h
@@ -47,6 +47,10 @@
 #include "google/protobuf/stubs/platform_macros.h"
 #include "google/protobuf/stubs/port.h"
 
+#if !defined(PROTOBUF_FORCE_OLD_CPP_VERSION_WE_KNOW_WHAT_WE_ARE_DOING) && __cplusplus < 201402L
+#error "Protobuf requires at least C++14"
+#endif
+
 #ifndef PROTOBUF_USE_EXCEPTIONS
 #if defined(_MSC_VER) && defined(_CPPUNWIND)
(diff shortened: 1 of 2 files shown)


mumbleskates:
Note that we already set C++14 in `.bazelrc`; the only thing these newly added lines in `cpp_opts.bzl` do is force libraries to build this way when protobuf is not the bazel root. unfortunately, any project using C++20 and propagating `--cxxopt=-std=c++20` or similar will also be overridden, leading to missing symbols at link time as absl generates different code in different versions.

src/google/protobuf/stubs/common.h
```diff
@@ -47,6 +47,10 @@
 #include "google/protobuf/stubs/platform_macros.h"
 #include "google/protobuf/stubs/port.h"
 
+#if !defined(PROTOBUF_FORCE_OLD_CPP_VERSION_WE_KNOW_WHAT_WE_ARE_DOING) && __cplusplus < 201402L
```
mkruskal-google, line 50: Can you remove this guard?  Our code is completely broken pre-C++14 right now, so if someone were to want to use it they'd need to make changes anyway (i.e. they could just delete this check)

mkruskal-google (commented):
This is a good point, thanks for the cleanup!

src/google/protobuf/stubs/common.h
mkruskal-google, line 50: This is causing failures in our windows and aarch64 builds.  I suspect you may need some variant of __cplusplus for these platforms

mumbleskates: I see, that's fair. I wanted something to emulate this as a lower-bound guard but if it already explodes below C++14 this is somewhat redundant.

mkruskal-google: Sorry I think you misunderstood.  I love this C++14 guard, it'll give a very clear error message instead of some random compiler error.  The issue I had was just with your `PROTOBUF_FORCE_OLD_CPP_VERSION_WE_KNOW_WHAT_WE_ARE_DOING` short-circuit, which seems unnecessary

mkruskal-google: See above, I like the guard a lot.  It's a good idea, I just think we need to either drop it on certain platforms or use alternatives

mumbleskates: i see! i wanted to offer a pre-built way to bypass the error without patching the header files; like if someone is on `__cplusplus == 201103L` but has `<string_view>` and is defining `ABSL_USES_STD_STRING_VIEW` etc., but i don't know what all possibilities exist. If it's messy enough to do this we can drop it.

mumbleskates:
Great, I eventually found the equivalent guard in abseil. Let me know if you would prefer to squash these commits

src/google/protobuf/stubs/common.h
mkruskal-google: Yea the issue is that we make no guarantees that C++11 will work, and in fact I'm pretty confident it doesn't.  So in this case they'd need hack our code anyway to get it compiling, and could just remove this guard then

mkruskal-google:
Perfect!  Lemme just rerun the tests and I'll squash and merge

src/google/protobuf/stubs/common.h
mumbleskates: out of curiosity, what compiler is the aarch64 build using that it doesn't have __cplusplus?

mkruskal-google: That's a good question.  IIRC our aarch64 are using an ancient version of gcc that's actually well outside our support window.  I have a TODO to upgrade those, but it's non-trivial

src/google/protobuf/stubs/common.h
```diff
@@ -47,6 +47,17 @@
 #include "google/protobuf/stubs/platform_macros.h"
 #include "google/protobuf/stubs/port.h"
 
+// Enforce C++14 as the minimum.
+#if defined(_MSVC_LANG)
+#if _MSVC_LANG < 201402L
+#error "C++ versions less than C++14 are not supported."
+#endif  // _MSVC_LANG < 201402L
+#elif defined(__cplusplus)
```
mkruskal-google, line 55: Could you just special-case gcc < 5.0 here?  We don't support it anyway so it should be safe to add, and will unblock the aarch64 tests.  e.g.
#elif defined(__cplusplus) || (defined(__GNUC__) && __GNUC__ < 5)

mkruskal-google: It looks like this version of gcc sets \_\_cplusplus to 201300L

mumbleskates: okay i set a special case for this; this will have to be backported to the guard in absl for these tests to pass there if we update the absl pin

mkruskal-google: Ah thanks for the heads up!  I'll try to prioritize this issue then, because we definitely want to bump our Abseil version before the next release

mkruskal-google (approved):

mumbleskates:
thanks!

PR #10786, opened 2022-10-18
Update zlib to 1.2.13.

benjaminp:
This pulls in the fix for CVE-2022-37434.

### DIFF
--- a/protobuf_deps.bzl
+++ b/protobuf_deps.bzl
@@ -49,9 +49,12 @@ def protobuf_deps():
         http_archive(
             name = "zlib",
             build_file = "@com_google_protobuf//:third_party/zlib.BUILD",
-            sha256 = "d8688496ea40fb61787500e863cc63c9afcbc524468cedeb478068924eb54932",
-            strip_prefix = "zlib-1.2.12",
-            urls = ["https://github.com/madler/zlib/archive/v1.2.12.tar.gz"],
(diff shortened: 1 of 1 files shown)


protobuf_deps.bzl
```diff
@@ -49,9 +49,12 @@ def protobuf_deps():
...
             name = "zlib",
             build_file = "@com_google_protobuf//:third_party/zlib.BUILD",
-            sha256 = "d8688496ea40fb61787500e863cc63c9afcbc524468cedeb478068924eb54932",
-            strip_prefix = "zlib-1.2.12",
-            urls = ["https://github.com/madler/zlib/archive/v1.2.12.tar.gz"],
+            sha256 = "d14c38e313afc35a9a8760dadf26042f51ea0f5d154b0630a31da0540107fb98",
+            strip_prefix = "zlib-1.2.13",
+            urls = [
+                "https://github.com/madler/zlib/releases/download/v1.2.13/zlib-1.2.13.tar.xz",
+                "https://zlib.net/zlib-1.2.13.tar.xz",
```
deannagarcia, line 56: Why did this URL need to be adde?

benjaminp: It doesn't need to be added, but it does always Bazel to fallback if the github.com url doesn't work.

deannagarcia:
Thanks for the fix!

PR #10789, opened 2022-10-18
Create bazel ruby release target

deannagarcia:
Create `ruby:release` bazel target that builds 2 gems for release (one just source and one for java).

### DIFF
--- a/src/google/protobuf/lite_unittest.cc
+++ b/src/google/protobuf/lite_unittest.cc
@@ -39,6 +39,7 @@
 #include <gtest/gtest.h>
 #include "google/protobuf/stubs/logging.h"
 #include "absl/strings/match.h"
+#include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
 #include "google/protobuf/arena_test_util.h"
 #include "google/protobuf/io/coded_stream.h"
@@ -1170,6 +1171,7 @@ TYPED_TEST(LiteTest, DebugString) {
   // later without breaking any code.
   EXPECT_NE(message1.DebugString(), message2.DebugString());
 }
+
 
 TYPED_TEST(LiteTest, EnumValueToName) {
   EXPECT_EQ("FOREIGN_LITE_FOO", protobuf_unittest::ForeignEnumLite_Name(
--- a/src/google/protobuf/message_lite.h
+++ b/src/google/protobuf/message_lite.h
@@ -222,6 +222,13 @@ class PROTOBUF_EXPORT MessageLite {
   // with Message.
   std::string Utf8DebugString() const { return DebugString(); }
 
+  // Implementation of the `AbslStringify` interface. This adds `DebugString()`
+  // to the sink. Do not rely on exact format.
+  template <typename Sink>
+  friend void AbslStringify(Sink& sink, const google::protobuf::MessageLite& msg) {
+    sink.Append(msg.DebugString());
+  }
+
   // Parsing ---------------------------------------------------------
   // Methods for parsing in protocol buffer format.  Most of these are
   // just simple wrappers around MergeFromCodedStream().  Clear() will be


protobuf_version.bzl
```diff
@@ -1,3 +1,4 @@
 PROTOC_VERSION = "21.7"
 PROTOBUF_JAVA_VERSION = "3.21.7"
 PROTOBUF_PYTHON_VERSION = "4.21.7"
+PROTOBUF_RUBY_VERSION = "3.21.7"
```
mkruskal-google, line 4: This should be 21.8 (post-merge with #10788)

ruby/build_release.sh
```diff
@@ -0,0 +1,25 @@
...
+cp external/utf8_range/range2-neon.c $UTF8_DIR
+cp external/utf8_range/range2-sse.c $UTF8_DIR
+cp external/utf8_range/utf8_range.h $UTF8_DIR
+
+# Make all files global readable/writable/executable
+cd $WORKING_DIR
+chmod -R 777 ./
+
+# Build gem
+bash --login -c "rvm use system && gem build google-protobuf.gemspec"
```
mkruskal-google, line 23: "system" could be jruby couldn't it?

ruby/BUILD.bazel
```diff
@@ -83,6 +85,35 @@ filegroup(
     ],
 )
 
+java_export(
+    name = "protobuf_java_release",
```
mkruskal-google, line 89: nit: name this jruby instead of java?

ruby/build_release.sh
deannagarcia: I guess it could, what should I set this to to get a reasonable ruby version?

ruby/BUILD.bazel
deannagarcia: I was struggling with naming and ended up matching the jar name and the `protobuf_java` rule above.

I agree it would be more clear if this target had "jruby" in it, but I'm worried that makes it less clear what the other "java" means. WDYT?

ruby/build_release.sh
mkruskal-google: I would leave it alone and just require the caller to set the ruby version.  The future plan for Bazel+Ruby is that our ruby rules will use the system ruby by default, with an optional flag to pin a specific version.  So we could just run this twice, once with cruby and once with jruby

ruby/ext/google/protobuf_c/extconf.rb
```diff
@@ -25,4 +25,4 @@
          "repeated_field.c", "map.c", "ruby-upb.c", "wrap_memcpy.c",
          "naive.c", "range2-neon.c", "range2-sse.c"]
 
-create_makefile(ext_name, Dir.pwd+"/../../../../ext/google/protobuf_c")
```
mkruskal-google, line 28: I think this may have been necessary to get this file working correctly in a Bazel sandbox (looking at your failures).  Did your Bazel command fail from this or was it just manual runs?

ruby/build_release.sh
deannagarcia: Leave it alone meaning don't call `rvm use` and just call `gem build google-protobuf.gemspec`? Isn't that also just defaulting to the system version?

ruby/BUILD.bazel
mkruskal-google: Good point, better to keep them consistent.  These will probably all get refactored soon anyway with my Bazel changes

ruby/build_release.sh
mkruskal-google: Exactly.  It defaults to the *current* version (which is distinct from rvm's "system" I believe), but it doesn't make any assumptions about what that version is.  So if we want to produce both c-ruby and java-ruby artifacts, we'd use rvm to set our version and run this target twice

protobuf_version.bzl
deannagarcia: Fixed

ruby/build_release.sh
deannagarcia: Resolved offline. Will wait to submit until we have implemented a flag to specify ruby versions.

ruby/build_jruby_release.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# This file should be executed with jruby
+
+set -ex
+
+WORKING_DIR=$(dirname $(readlink $BASH_SOURCE))
```
mkruskal-google, line 6: If you use inline_sh_binary instead of sh_binary, you can take advantage of Bazel make variables.  Right now, this is highly non-hermetic, and runs over the non-sandboxed source tree

PR #10801, opened 2022-10-20
Rename C# files to indicate generated code

### LINKED ISSUE #10776: Change extension of C# generated source code in Google.Protobuf
We have a fair amount of generated code in https://github.com/protocolbuffers/protobuf/tree/main/csharp/src
It can be useful to differentiate this code based purely on the extension, and .NET typically uses .g.cs as a file extension.

We should change our [generation script](https://github.com/protocolbuffers/protobuf/blob/main/csharp/generate_protos.sh) to explicitly specify .g.cs as the extension while building (using `--csharp_opt=file_extension=.g.cs`) and delete the old code.

Extra stuff to check:

- Conformance test build scripts
- Compatibility tests
- distcheck

jskeet:
Fixes #10776 - at least for the "main/test" protos. Conformance and compatibility still to be done.


mkruskal-google (changes requested):
Looks like you have some real test failures from dangling references to the old files.  Otherwise lgtm tho

jskeet:
Thanks @mkruskal-google - I'd missed a Bazel file. That's currently in a second commit. Kokoro is looking happier now (just waiting for the Ruby build).

mkruskal-google (approved):

PR #10803, opened 2022-10-21
fix: remove extension exception

bshaffer:
- for parity with the native implementation (changing the implementation to match the message would break BC)
 - the return value is already checked everywhere this method is called ([1](https://github.com/protocolbuffers/protobuf/blob/main/php/ext/google/protobuf/message.c#L290), [2](https://github.com/protocolbuffers/protobuf/blob/main/php/ext/google/protobuf/message.c#L324), [3](https://github.com/protocolbuffers/protobuf/blob/main/php/ext/google/protobuf/message.c#L361), [4](https://github.com/protocolbuffers/protobuf/blob/main/php/ext/google/protobuf/message.c#L392))

There's also an issue we've seen here when an API removes fields, and we want to keep the methods in order to preserve BC. The protobuf message throws an exception (only in the extension) because the fields exist, even if they're never used. It would be easier to just removed this unnecessary exception.

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -115,11 +115,6 @@ static const upb_FieldDef *get_field(Message *msg, PROTO_STR *member) {
   const upb_FieldDef *f =
       upb_MessageDef_FindFieldByNameWithSize(m, PROTO_STRVAL_P(member), PROTO_STRLEN_P(member));
 
-  if (!f) {
-    zend_throw_exception_ex(NULL, 0, "No such property %s.",
-                            ZSTR_VAL(msg->desc->class_entry->name));
-  }
-
   return f;
 }
 


haberman:
> There's also an issue we've seen here when an API removes fields, and we want to keep the methods in order to preserve BC.

Can you give an example of what you mean?

If a proto message removes a field, I don't think we should allow API references to that field. Otherwise a user could refer to any field name, even if it never existed.

bshaffer:
> Can you give an example of what you mean?

An API removes a field `fieldA`. We keep the PHP methods `getFieldA` and `setFieldA`, because rather than throw a PHP Fatal error in functioning code, it results in a No-Op. It's extra work (and also another thing to remember) to remove the property also  because otherwise the extension throws an exception. This isn't the main reason to remove this behavior though - the main reason is to have the extension and native implementation function the same.

> If a proto message removes a field, I don't think we should allow API references to that field.

If our libraries are GA, we have to preserve the PHP surface. Otherwise it would be a breaking change. So the no-op is the way to go.

> Otherwise a user could refer to any field name, even if it never existed.

I don't understand how this would be true

haberman:
> An API removes a field fieldA.

It would be helpful to frame this scenario around `.proto` files specifically. Are you saying something like this?

Before:

```proto
message Foo {
  optional int32 field_a = 1;
  optional int32 field_b = 2;
}
```

After:

```proto
message Foo {
  optional int32 field_b = 2;
}
```

What I hear you saying is that you want `$foo->getFieldA()` to work (return `null`?) even when used with the "after" proto. Is that correct?

bshaffer:
@haberman I think the misunderstanding is that we want to keep the _previously generated_ protobuf file in favor of the new one. We don't want any protobuf message to be able to set any property. We've had to modify it currently, and remove the setting of the properties, because otherwise the extension throws an exception (because we are still updating the metadata). 

It's a hack, but it's a necessary one when APIs make breaking changes. Also, again, this isn't the main reason for this change. The main reason is for consistency.

haberman:
I'm just trying to understand the desired behavior. Can you show an example of the `.proto` file input, and some PHP code that uses the resulting generated class, and the behavior you'd like it to have?

If we have the input:

```proto
// After
message Foo {
  optional int32 field_b = 2;
}
```

It is impossible to generate the following output, because `protoc` has no way of knowing that `field_a` ever existed:

```php
// After:
class Foo extends Message
{
    public function setFieldA($var);
    public function getFieldA();

    public function setFieldB($var);
    public function getFieldB();
}
```

PR #10816, opened 2022-10-25
Update cc_dist_library() to include transitive sources

acozzette:
The cc_dist_library() rule originally included only the sources from direct dependencies. This resulted in a less than ideal developer experience, because if you ever added a new cc_library() then you would have to carefully update the necessary cc_dist_library() targets to ensure that the change was correctly reflected in the CMake build.

This commit addresses that issue by making cc_dist_library() include transitive sources. We have to be careful to avoid introducing ODR violations (e.g. from libprotoc duplicating sources from libprotobuf), so we introduce a new dist_deps attribute on cc_dist_library(). Anything in dist_deps is assumed to be covered by a separate cc_dist_library() and is not included. We also make sure to exclude anything that's not part of our repo (i.e. Abseil and zlib).

### DIFF
--- a/pkg/BUILD.bazel
+++ b/pkg/BUILD.bazel
@@ -344,15 +344,8 @@ cc_dist_library(
     }),
     tags = ["manual"],
     deps = [
-        "//src/google/protobuf:arena",
         "//src/google/protobuf:arena_align",
-        "//src/google/protobuf:arena_allocation_policy",
-        "//src/google/protobuf:arena_cleanup",
-        "//src/google/protobuf:arena_config",
         "//src/google/protobuf:protobuf_lite",
-        "//src/google/protobuf/io",
-        "//src/google/protobuf/io:io_win32",
-        "//src/google/protobuf/stubs:lite",
     ],
 )
 
@@ -367,33 +360,11 @@ cc_dist_library(
     }),
     tags = ["manual"],
     deps = [
-        "//src/google/protobuf:arena",
         "//src/google/protobuf:arena_align",
-        "//src/google/protobuf:arena_allocation_policy",
-        "//src/google/protobuf:arena_cleanup",
-        "//src/google/protobuf:arena_config",
-        "//src/google/protobuf:port_def",
-        "//src/google/protobuf:protobuf_lite",
         "//src/google/protobuf:protobuf_nowkt",
         "//src/google/protobuf:wkt_cc_proto",
         "//src/google/protobuf/compiler:importer",
-        "//src/google/protobuf/io",
-        "//src/google/protobuf/io:gzip_stream",
-        "//src/google/protobuf/io:io_win32",
-        "//src/google/protobuf/io:printer",
-        "//src/google/protobuf/io:tokenizer",
-        "//src/google/protobuf/io:zero_copy_sink",
         "//src/google/protobuf/json",
-        "//src/google/protobuf/json:descriptor_traits",
-        "//src/google/protobuf/json:lexer",
-        "//src/google/protobuf/json:message_path",
-        "//src/google/protobuf/json:parser",
-        "//src/google/protobuf/json:unparser",
-        "//src/google/protobuf/json:untyped_message",
-        "//src/google/protobuf/json:writer",
-        "//src/google/protobuf/json:zero_copy_buffered_stream",
-        "//src/google/protobuf/stubs",
-        "//src/google/protobuf/stubs:lite",
         "//src/google/protobuf/util:delimited_message_util",
         "//src/google/protobuf/util:differencer",
         "//src/google/protobuf/util:field_mask_util",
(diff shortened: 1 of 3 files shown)


pkg/BUILD.bazel
```diff
@@ -426,13 +426,18 @@ cc_dist_library(
         "//src/google/protobuf/compiler/python",
         "//src/google/protobuf/compiler/ruby",
```
mkruskal-google, line 427: Can you clean these up now (and update the file_lists)?  It would be a good proof-of-concept to show that it works, and reduce a lot of complexity here

src/file_lists.cmake
```diff
@@ -254,6 +254,8 @@ set(libprotobuf_lite_hdrs
   ${protobuf_SOURCE_DIR}/src/google/protobuf/metadata_lite.h
   ${protobuf_SOURCE_DIR}/src/google/protobuf/parse_context.h
   ${protobuf_SOURCE_DIR}/src/google/protobuf/port.h
+  ${protobuf_SOURCE_DIR}/src/google/protobuf/port_def.inc
```
mkruskal-google, line 257: Any idea why these weren't included before (or how this worked w/o them)?

pkg/cc_dist_library.bzl
```diff
@@ -265,6 +272,27 @@ def _collect_inputs(deps):
...
+    for field in top_level_fields:
+        to_remove = {e: None for e in getattr(b, field)}
+        result_args[field] = [e for e in getattr(a, field) if not e in to_remove]
+
+    cc_file_list_args = {}
+    file_list_fields = ["srcs", "hdrs", "internal_hdrs", "textual_hdrs"]
+    for field in file_list_fields:
+        to_remove = {e: None for e in getattr(b.cc_file_list, field).to_list()}
+        cc_file_list_args[field] = depset(
+            [e for e in getattr(a.cc_file_list, field).to_list() if not e in to_remove and e.path.find("external") == -1],
```
mkruskal-google, line 290: Clever, is this how we're filtering out non-proto dependencies?  It might be worth a comment given the importance here

pkg/BUILD.bazel
acozzette: Done. Some of these still need multiple deps in the list just due to how our CMake targets are set up, but at least the lists can be much shorter now.

pkg/cc_dist_library.bzl
acozzette: Right, it's filtering out non-proto dependencies by checking for "external" in the file name. (There might actually be a better way to do it, but this is the best I've figured out so far.) I added a comment to mention this.

src/file_lists.cmake
acozzette: It looks like the :port_def target was included in the `cc_dist_library` for libprotobuf but not the one for libprotobuf-lite. I'm guessing the reason we didn't notice is that the build still probably worked even without the undeclared dependency (i.e. the C++ compiler would still see `port_def.inc` in the include path even though CMake didn't know about it).

pkg/cc_dist_library.bzl
mkruskal-google: The way I thought you were going to do it after our discussion a while back was to force the cc_dist_library to explicitly exclude these in some way, so this is a huge improvement over that :)

I think one way to handle this a little better is in `_flatten_target_files`, you can use `target.label.workspace_name` to find the repo that the target is in.  I don't think you want to search past external deps, so it's better to stop the transitive search once you hit them anyway

mkruskal-google: See here: https://bazel.build/rules/lib/Target, https://bazel.build/rules/lib/Label

pkg/BUILD.bazel
mkruskal-google: I'm a little confused why we still need some of those extra ones.  For example, why can't protobuf_lite just depend on `//src/google/protobuf:protobuf_lite`?

acozzette: We could consider doing that, but that would result in `arena_align` being excluded from the CMake build. I think it just comes down to the Bazel build being more fine-grained than the CMake build, so a single CMake target can correspond to multiple Bazel targets.

pkg/cc_dist_library.bzl
acozzette: Oh, good idea! Done.

pkg/BUILD.bazel
mkruskal-google: protobuf_lite depends directly on arena_align though..  I think if it's getting excluded there's a bug somewhere

pkg/cc_dist_library.bzl
```diff
@@ -134,12 +134,15 @@ CcFileList = provider(
 )
 
 def _flatten_target_files(targets):
-    return depset(transitive = [target.files for target in targets])
+    return depset(transitive = [
+        target.files
+        for target in targets
+        # Filter out targets from external workspaces
+        if target.label.workspace_name == ""
```
mkruskal-google, line 141: You should probably include "com_google_protobuf" here too?  Does that pick up some unwanted dependencies?

pkg/BUILD.bazel
acozzette: Where do you see that dependency? Unless I'm missing something, it looks like nothing depends on `:arena_align` except for its unit test.

mkruskal-google: O whoops, I was looking at //pkg:protobuf_lite >.<.  Yea this makes sense that we want to package up utilities that we don't use.

mkruskal-google (approved):

pkg/cc_dist_library.bzl
acozzette: Good point, done.

PR #10821, opened 2022-10-26
Clean up the dotnet test output

jskeet:
Prior to this change, running "dotnet test" would result in a warning about csc only supporting C# up to version 5, and warnings of unrecognized warnings being suppressed.

With this change in place, the test output doesn't include anything from csc.


jskeet:
Hmm - I'm surprised that the Windows C# Kokoro build has failed. (Obviously I tested it locally before creating the PR.) I don't have access to the log file, unfortunately :(

acozzette:
It looks like it was just a network failure, so I kicked off another run to see if it passes this time.

jskeet:
@acozzette: Thanks, all green now :)

jskeet:
@mkruskal-google: Please could you merge as well? I don't have write permission...

PR #10826, opened 2022-10-26
Add build rule for pecl php release

deannagarcia:
`bazel build //php:release` should now build a `.tgz` file that we can upload to PECL.


php/BUILD.bazel
```diff
@@ -137,3 +139,48 @@ pkg_files(
     strip_prefix = strip_prefix.from_pkg(""),
     visibility = ["//pkg:__pkg__"],
 )
+
+################################################################################
+# PECL .tgz Release
```
zhangskz, line 144: Do we want to separate this into php/ext/google/protobuf or php/ext? Right now, I have the BUILD rule for the package.xml generation stuff and templates in php/ext/google/protobuf.

php/BUILD.bazel
```diff
@@ -137,3 +139,48 @@ pkg_files(
...
+)
+
+pkg_tar(
+  name = "release",
+  extension = "tgz",
+  srcs = [
+    ":release_files",
+    "ext/google/protobuf/package.xml",
+  ],
+  package_file_name = "protobuf-%s.tgz" % PROTOBUF_PHP_VERSION,
```
zhangskz, line 184: Note that the PHP release version that we write into package.xml (and thus the .tgz) doesn't quite match the format we use for protobuf_version.bzl for other versions. For PHP we use 3.21.0RC1 instead of 3.21.0-rc1. Maybe its ok to keep this in protobuf_version.bzl, but we should ensure it reflects the correct version.

I have logic in the package.xml generation to read the version number directly from version.json right now and reformat, but I think having a PROTOBUF_PHP_VERSION is probably better.

php/BUILD.bazel
deannagarcia: I'm fine depending on your rules in php/ext/google/protobuf but I want the pkg_tar/release rule to be in this file.

php/BUILD.bazel
deannagarcia: I think we should be able to keep PROTOBUF_PHP_VERSION to be the correct formatting, even if that means not all the formats in protobuf_version.bzl match.

PR #10832, opened 2022-10-27
Hide ruby native extension symbols on FreeBSD

dankm:
They probably should be hidden on every ELF platform, but for now only do it on Darwin, Linux, and FreeBSD.

This prevents some load-time errors that have been seen in GitLab's gitaly server, and in a few other places.

### DIFF
--- a/ruby/ext/google/protobuf_c/extconf.rb
+++ b/ruby/ext/google/protobuf_c/extconf.rb
@@ -6,7 +6,7 @@ ext_name = "google/protobuf_c"
 
 dir_config(ext_name)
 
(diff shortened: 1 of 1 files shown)


dankm:
Hm. That commit message needs editing. I wonder if all the build checks will restart if I just update it?

acozzette:
I would need to manually kick off the builds again, but I'm happy to do that if you want to edit the commit message.

dankm:
Sure, now the commit message doesn't look like somebody who got distracted while writing it ;)

dankm:
This relates to [this FreeBSD review](https://reviews.freebsd.org/D37140).

dankm:
Anybody else who should be added to help review this?

haberman (approved):

PR #10839, opened 2022-10-28
Add update php repo github action

deannagarcia:
This Github action will automatically sync to the php repository whenever we push a new release. It will copy over the new files and add proper tags.


.github/workflows/update_php_repo.yml
```diff
@@ -0,0 +1,45 @@
+name: Update protobuf-php Repo
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+
```
zhangskz, line 6: I think this needs to be expanded to include RCs. I also suspect the "."  should also be escaped.

.github/workflows/update_php_repo.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Copy files
+        run: |
+          rm -rf src
+          cp -r protobuf/php/src .
+          cp protobuf/php/composer.json.dist composer.json
+          rm -rf protobuf
+      - name: Push Changes
+        run: |
+          git add .
+          git commit -m "$VERSION sync"
```
zhangskz, line 42: Nit: You can do `git commit -a -m "$VERSION sync" instead.

zhangskz (approved):

.github/workflows/update_php_repo.yml
deannagarcia: Added a new tag to match on RCs.

.github/workflows/update_php_repo.yml
deannagarcia: Fixed.

zhangskz (approved):

zhangskz (approved):

googleberg:
These changes have landed through copybara. Copybara should have autoclosed or automerged this. Closing manually.

PR #10843, opened 2022-10-31
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.


zhangskz:
There are merge conflicts in .github/workflows/generated_cmake.yml that are preventing some integration tests from running.

zhangskz (approved):

.github/workflows/generate_files.yml
```diff
@@ -9,8 +9,12 @@ on:
       # to exclude it.
       - '!21.x'
 
+permissions: {}
 jobs:
   cmake:
+    permissions:
+      contents: write  #  for git push
```
mkruskal-google, line 16: Doesn't this need read permissions as well?  This workflow doesn't run as a presubmit, so it won't fail until after you merge this PR

.github/workflows/generate_files.yml
```diff
@@ -9,8 +9,12 @@ on:
       # to exclude it.
       - '!21.x'
 
+permissions: {}
```
mkruskal-google, line 12: Why not put this in the other workflow files?

.github/workflows/generate_files.yml
sashashura: `write` permission gives read as well, that's the syntax.

.github/workflows/generate_files.yml
sashashura: I see only one workflow without `permissions` - update_php_repo.yml which didn't exist at the moment when the pull-request was created. Or do you mean why it is `read`, but not `{}` in other workflows?

mkruskal-google: Yea I meant why is this the only one where you make the workflow-wide default empty?

Issue #10833, opened 2022-10-27
Protobuf fails to compile on Ubuntu 22.04 when -Dprotobuf_BUILD_SHARED_LIBS=ON flag set.


BDokas:
**What version of protobuf and what language are you using?**
Version: v3.21.4
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Linux, Ubuntu 22.04

**What runtime / compiler are you using (e.g., python version or gcc version)**
cmake 

**What did you do?**
1. Run `git clone https://github.com/protocolbuffers/protobuf.git`
2. Run `cd protobuf`
3. Run `git submodule update --init --recursive`
4. Run `cmake -Dprotobuf_BUILD_SHARED_LIBS=ON .`
5. Run `cmake --build . --parallel 10`

**What did you expect to see**
A successful build with shared objects generated in /usr/local/libs after installation.

**What did you see instead?**
```
/usr/bin/ld: libprotoc.so.3.21.4.0: undefined reference to `google::protobuf::RepeatedPtrField<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::erase(google::protobuf::internal::RepeatedPtrIterator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const>)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [CMakeFiles/protoc.dir/build.make:151: protoc-3.21.4.0] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:889: CMakeFiles/protoc.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
[protobuf_build_failure.txt](https://github.com/protocolbuffers/protobuf/files/9882547/protobuf_build_failure.txt)

```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**

enum-class:
I think your problem is similar to this, please check it :
https://github.com/protocolbuffers/protobuf/issues/10828

enum-class:
The problem is about extern template class in repeated_ptr_field.h line 2029.
The trivial way to solve this problem is comment the extern which will impose a lot of burden on compiler and linker.
You can use this method until the problem truly get solved.
I can fix it if nobody mind

acozzette:
[...]

PR #10879, opened 2022-11-02
Fix protobuf fails to compile when -Dprotobuf_BUILD_SHARED_LIBS=ON

enum-class:
Protobuf fails to compile when -Dprotobuf_BUILD_SHARED_LIBS=ON flag set. #10833

### DIFF
--- a/src/google/protobuf/repeated_field.cc
+++ b/src/google/protobuf/repeated_field.cc
@@ -55,7 +55,6 @@ template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<int64_t>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<uint64_t>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<float>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<double>;
-//template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedPtrField<std::string>;
 
 namespace internal {
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedIterator<bool>;
--- a/src/google/protobuf/repeated_field.cc
+++ b/src/google/protobuf/repeated_field.cc
@@ -46,6 +46,7 @@
 namespace google {
 namespace protobuf {
 
+extern template class PROTOBUF_EXPORT_TEMPLATE_DECLARE RepeatedPtrField<std::string>;
 
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<bool>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<int32_t>;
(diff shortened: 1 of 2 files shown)


enum-class:
Hi @acozzette , could you please add a c++ label here ? :))

enum-class:
Hi @acozzette , could you please help me with `import/copybara` error ? I can not see the details :)

enum-class:
Hi @acozzette, @haberman , could you please help me help me with review of this P.R.?

enum-class:
Hi @acozzette , could you please review this code ? it is less than 5 line :))

acozzette:
@enum-class Sorry for the delay on this! We have been changing our review setup lately and I believe I need to close and reopen this to get out internal tests going, so let me give that a try.

enum-class:
Hi @acozzette , thank you for approval. could you please help me with import/copybara error ?

acozzette:
Sorry for the continuing delay on this. We revamped our infrastructure for syncing changes into our internal codebase, and we have been ironing out out a few bugs in the process. I believe it is finally working and I plan to take a look today.

acozzette:
@enum-class Could you explain more about how what causes the build error and how this fixes it? I don't understand why we need a special case for `RepeatedPtrField<std::string>` and why we should handle that one differently from the other template instantiations.

enum-class:
Hi @acozzette , As you know, `extern template` force the compiler to not instantiate a template when you know that it will be instantiated somewhere else. It is used to reduce compile time and object file size.
When object files will be linked together, the second object file will just use the symbol from the first object file. No need for discard and no wasted compile time and object file size.
This should only be used within a project, like in times when you use a template like `RepeatedPtrField<std::string>` multiple times, you should use extern.

enum-class:
There is another way to solve this problem:

```
diff --git a/src/google/protobuf/repeated_field.cc b/src/google/protobuf/repeated_field.cc
index 9e1d41957..c5dbf66f1 100644
--- a/src/google/protobuf/repeated_field.cc
+++ b/src/google/protobuf/repeated_field.cc
@@ -54,7 +54,7 @@ template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<int64_t>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<uint64_t>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<float>;
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedField<double>;
-template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedPtrField<std::string>;
 
 namespace internal {
 template class PROTOBUF_EXPORT_TEMPLATE_DEFINE RepeatedIterator<bool>;
diff --git a/src/google/protobuf/repeated_ptr_field.cc b/src/google/protobuf/repeated_ptr_field.cc
index 1661474f2..e7e26cce3 100644
--- a/src/google/protobuf/repeated_ptr_field.cc
+++ b/src/google/protobuf/repeated_ptr_field.cc
@@ -46,6 +46,8 @@
 namespace google {
 namespace protobuf {
 
+extern template class PROTOBUF_EXPORT_TEMPLATE_DECLARE RepeatedPtrField<std::string>;
+
 namespace internal {
 
 void** RepeatedPtrFieldBase::InternalExtend(int extend_amount) {
diff --git a/src/google/protobuf/repeated_ptr_field.h b/src/google/protobuf/repeated_ptr_field.h
index 699b6c764..8d3a310c9 100644
--- a/src/google/protobuf/repeated_ptr_field.h
+++ b/src/google/protobuf/repeated_ptr_field.h
@@ -2021,9 +2021,6 @@ UnsafeArenaAllocatedRepeatedPtrFieldBackInserter(
       mutable_field);
 }
 
-extern template class PROTOBUF_EXPORT_TEMPLATE_DECLARE
-    RepeatedPtrField<std::string>;
-
 }  // namespace protobuf
 }  // namespace google
 

```

acozzette:
I ended up doing a slightly different fix [here](https://github.com/protocolbuffers/protobuf/commit/225b936c0183e98b7c5e072d9979c01f952c2d5a), but thank you for bringing this problem to our attention and helping us track down what was going on.

PR #10900, opened 2022-11-03
Fix misdetection of libatomic with strict compiler flags

### LINKED ISSUE #10899: libatomic mis-detected as required with strict compiler flags
Version: main
Language: C++

macOS, but the problem is more general

Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin22.1.0

**What did you do?**
I am including protobuf via CMake into a larger project (https://github.com/shaka-project/shaka-packager).  In that context, the entire project is built with very strict compiler flags and -Werror.

What I see is that this test:

```
  check_cxx_source_compiles("
    #include <atomic>
    int main() {
      return std::atomic<int64_t>{};
    }
  " protobuf_HAVE_BUILTIN_ATOMICS)
```

fails with an error about integer truncation from int64_t to int, even though 64-bit atomics are available without a library.  But this error occurs on macOS, which doesn't have libatomic, leading later to a link error.

I found that a simple solution is to add a static cast to the test, so that it produces the correct results even in a very strict environment:

```
  check_cxx_source_compiles("
    #include <atomic>
    int main() {
      return static_cast<int>(std::atomic<int64_t>{});
    }
  " protobuf_HAVE_BUILTIN_ATOMICS)
```

I'll send a PR, in the hopes that this is an acceptable solution.

joeyparrish:
With strict flags and -Werror, the detection for libatomic breaks, causing it to be included in places where it is not needed.  And on macOS, this leads to a link error when libatomic cannot be found.

Adding a static_cast fixes the problem by making the test compatible with strict compiler flags.

Closes #10899

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,7 +238,7 @@ if (NOT MSVC)
   check_cxx_source_compiles("
     #include <atomic>
     int main() {
-      return std::atomic<int64_t>{};
+      return static_cast<int>(std::atomic<int64_t>{});
     }
   " protobuf_HAVE_BUILTIN_ATOMICS)
   if (NOT protobuf_HAVE_BUILTIN_ATOMICS)


zhangskz (approved):

googleberg:
This PR has been incorporated. Copybara should have autoclosed/automerged it, but didn't. Closing manually.

PR #10901, opened 2022-11-03
Add a staleness test for the well-known types

acozzette:
This commit adds a staleness test in preparation for setting up auto-updating of the well-known type generated code. I also went ahead and updated them manually just to make sure we don't run into any surprises when this starts happening automatically.

I had to arrange things so that the generated any.pb.cc (for example) lives at:
`//src/google/protobuf:wkt/google/protobuf/any.pb.cc`

This is a weird directory structure, but I think it's the only way to allow the checked-in and generated versions of the file to coexist while ensuring that Bazel always uses the generated version.

### DIFF
--- a/src/google/protobuf/BUILD.bazel
+++ b/src/google/protobuf/BUILD.bazel
@@ -5,6 +5,7 @@
 load("@rules_cc//cc:defs.bzl", "cc_library", "cc_proto_library", "cc_test")
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("@rules_proto//proto:defs.bzl", "proto_library")
+load("@upb//cmake:build_defs.bzl", "staleness_test")
 load("//build_defs:cpp_opts.bzl", "COPTS", "LINK_OPTS")
 
 package(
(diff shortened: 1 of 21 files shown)


src/google/protobuf/BUILD.bazel
```diff
@@ -106,28 +107,44 @@ WELL_KNOWN_TYPES = [
...
+        [wkt + ".pb.cc" for wkt in WELL_KNOWN_TYPES],
+    generated_pattern = "wkt/google/protobuf/%s",
+    tags = ["manual"],
+)
+
 cc_library(
     name = "wkt_cc_proto",
-    srcs = [wkt + ".pb.cc" for wkt in WELL_KNOWN_TYPES],
-    hdrs = [wkt + ".pb.h" for wkt in WELL_KNOWN_TYPES],
+    srcs = ["wkt/google/protobuf/" + wkt + ".pb.cc" for wkt in WELL_KNOWN_TYPES],
```
mkruskal-google, line 144: Clever :)

mkruskal-google (approved):

mkruskal-google (changes requested):
copybara will merge this, revoking approval

PR #10978, opened 2022-11-14
Add support for .NET Native AOT

JamesNK:
- [x] Support reflection accessors
- [x] Support extensions
- [ ] Integrate Native AOT app into unit tests **EDIT:** Should these run automatically? How? From CI script or as part of unit test publishing and launching app?

Previously added trim attributes enables most of Google.Protobuf to work with Native AOT. There are some problematic `MakeGenericType` usages around reflection and extensions that need work.

### DIFF
--- a/csharp/src/Google.Protobuf.Test.NativeAot/Program.cs
+++ b/csharp/src/Google.Protobuf.Test.NativeAot/Program.cs
@@ -1,0 +1,68 @@
+﻿#region Copyright notice and license
+// Protocol Buffers - Google's data interchange format
+// Copyright 2016 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+using Google.Protobuf.Test.NativeAot.Reflection;
+
+try
+{
+    // This project tests specific Google.Protobuf features with .NET Native AOT.
+    // It is designed to be published as a native app, and then run.
+    // Return value of 0 is success, 1 is failure.
+    //
+    // Because of trimming, there is a chance a scenario might not work in isolation.
+    // For now, all these tests are compiled and run together. Re-evaluate if this becomes a problem.
+    // Could be isolated by adding #ifdef here and then caller passes compiler flag to run specific tests.
+    Descriptors.GetFileDescriptorMessages();
+    Descriptors.GetMessageFields();
+    Descriptors.GetMessageEnumTypes();
+    Descriptors.GetEnumFields();
+    Descriptors.GetEnumDescriptorFields();
+    Descriptors.AccessorSingleFieldReflection();
+    Descriptors.AccessorOneOfFieldReflection();
+    Descriptors.AccessorRepeatingFieldReflection();
+    Descriptors.GetFileDescriptorServices();
+    Extensions.GetExtensionSingle();
+    Extensions.GetExtensionRepeated();
+    Extensions.GetExtensionEnumSingle();
+    Extensions.GetExtensionValue();
+    Extensions.ExtensionsIsInitialized();
+    Extensions.OptionLocations();
+    Json.FormatJson();
+    Json.ParseJson();
+
+    return 0;
+}
+catch (Exception ex)
+{
+    Console.WriteLine(ex);
+    return 1;
+}
--- a/csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Descriptors.cs
+++ b/csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Descriptors.cs
@@ -1,0 +1,161 @@
+﻿#region Copyright notice and license
+// Protocol Buffers - Google's data interchange format
+// Copyright 2016 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+using System.Collections;
+using System.Reflection;
+using ProtobufTestMessages.Proto3;
+using UnitTest.Issues.TestProtos;
+
+namespace Google.Protobuf.Test.NativeAot.Reflection
+{
+    internal static class Descriptors
+    {
+        public static void GetFileDescriptorMessages()
+        {
+            var messageTypes = TestMessagesProto3Reflection.Descriptor.MessageTypes;
+            Assert.AreNotEqual(0, messageTypes.Count);
+            
+            foreach (var item in messageTypes)
+            {
+                Console.WriteLine($"{item.Name} message");
+            }
+            Console.WriteLine($"{messageTypes.Count} messages");
+        }
+        
+        public static void GetMessageFields()
+        {
+            var fields = TestAllTypesProto3.Descriptor.Fields.InDeclarationOrder();
+            Assert.AreNotEqual(0, fields.Count);
+            
+            foreach (var item in fields)
+            {
+                Console.WriteLine($"{item.Name} field");
+            }
+            Console.WriteLine($"{fields.Count} fields");
+        }
+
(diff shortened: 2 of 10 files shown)


csharp/src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>10.0</LangVersion>
+    <LangVersion>11.0</LangVersion>
```
jskeet, line 4: Is this necessary for the change?

I can see that we may need to have at least one project targeting .NET 7, but I'd prefer not to make that a general requirement, sticking to LTS SDK versions if at all possible.

I think it would be reasonable for CI to install .NET 7 (as you've done above) to make sure that we really solve the AOT  issues, but I'd want everything other than that project to only require .NET 6, if at all possible.

Does that sound like a reasonable proposal that should work out the kinks on, or do you think it's unreasonable for folks to stay on the .NET 6 SDK?

(It would *definitely* make things simpler internally if we could stick to C# 10 features, at least for the main package.)

csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Descriptors.cs
```diff
@@ -0,0 +1,128 @@
...
+            foreach (var item in enumTypes)
+            {
+                Console.WriteLine($"{item.Name} enum type");
+            }
+            Console.WriteLine($"{enumTypes.Count} enum types");
+        }
+
+        public static void GetEnumFields()
+        {
+            // There isn't a descriptor for enum types. Use standard .NET reflection to check fields.
```
jskeet, line 47: Yes there is - there's EnumDescriptor. Use `MessageDescriptor.EnumTypes` in this case to fetch it.
(Checking the actual enum is present with all fields as well probably isn't a bad idea too...)

csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Descriptors.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            bytesfield.Accessor.SetValue(message, ByteString.CopyFrom(new byte[] { 1, 2, 3 }));
+
+            var bytesValue = (ByteString) bytesfield.Accessor.GetValue(message);
+            if (bytesValue.Span is not [1, 2, 3])
+            {
+                throw new Exception($"Expected 1,2,3.");
+            }
+
+            if (!bytesfield.Accessor.HasValue(message))
```
jskeet, line 88: Any reason for not using Assert.AreEqual for the next three assertions? (We could add Assert.True and Assert.False if necessary too, although I'm not bothered by `Assert.AreEqual(true, xyz)`.)

csharp/src/Google.Protobuf.Test.NativeAot/Program.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Google.Protobuf.Test.NativeAot.Reflection;
+
+try
```
jskeet, line 3: I think it would be good to include a descriptive comment, either here or in a README.md, to explain the purpose of all of this code, and the reason for not using a regular test framework. (I'm pretty sure I get it, but clarity's good.)

csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Extensions.cs
```diff
@@ -0,0 +1,185 @@
...
+            var wrongTypedExtension = new Extension<TestAllExtensions, TestAllTypes>(OptionalNestedMessageExtension.FieldNumber, codec: null);
+
+            var message = new TestAllExtensions();
+
+            var value1 = message.GetExtension(untypedExtension);
+            Assert.IsNull(value1);
+
+            message.SetExtension(OptionalNestedMessageExtension, extensionValue);
+            var value2 = message.GetExtension(untypedExtension);
+            Assert.IsNotNull(value2);
```
jskeet, line 25: Can we assert that it's equal to `extensionValue`?

csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Extensions.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            var message = CreateFullTestAllExtensions();
+
+            Assert.AreEqual(true, message.GetExtension(OptionalBoolExtension));
+
+            var ex = Assert.Throws<NotSupportedException>(() => TestProtos.Proto2.TestAllExtensions.Descriptor.FindFieldByNumber(OptionalBoolExtension.FieldNumber).Accessor.GetValue(message));
+            Assert.AreEqual("Extensions reflection is not supported with AOT.", ex.Message);
+        }
+
+        public static void IsExtensionsInitialized()
```
jskeet, line 88: Rename to ExtensionsIsInitialized given that the method we're testing is IsInitialized?

csharp/src/Google.Protobuf.sln
```diff
@@ -12,7 +13,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Protobuf.Conformance
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Protobuf.JsonDump", "Google.Protobuf.JsonDump\Google.Protobuf.JsonDump.csproj", "{9695E08F-9829-497D-B95C-B38F28D48690}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Google.Protobuf.Test.TestProtos", "Google.Protobuf.Test.TestProtos\Google.Protobuf.Test.TestProtos.csproj", "{ADF24BEB-A318-4530-8448-356B72B820EA}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Protobuf.Test.TestProtos", "Google.Protobuf.Test.TestProtos\Google.Protobuf.Test.TestProtos.csproj", "{ADF24BEB-A318-4530-8448-356B72B820EA}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Google.Protobuf.Test.NativeAot", "Google.Protobuf.Test.NativeAot\Google.Protobuf.Test.NativeAot.csproj", "{642232EA-FA89-4E84-9CA0-423EA98F37ED}"
```
jskeet, line 18: Depending on the outcome of discussions around .NET 7, we may not want to include this in the solution.

csharp/src/Google.Protobuf/Google.Protobuf.csproj
```diff
@@ -8,7 +8,7 @@
     <VersionPrefix>3.21.9</VersionPrefix>
     <LangVersion>10.0</LangVersion>
     <Authors>Google Inc.</Authors>
-    <TargetFrameworks>netstandard1.1;netstandard2.0;net45;net50</TargetFrameworks>
+    <TargetFrameworks>netstandard1.1;netstandard2.0;net45;net50;net60</TargetFrameworks>
```
jskeet, line 11: I assume we genuinely *need* this for AOT support? (If we could continue to use the net50 target, that would avoid the package getting bigger.) I do expect to remove net50 at some point in 2023 though (I believe that we'd still have compatibility - just without some optimizations. Let me know if your understanding is different.)

csharp/src/Google.Protobuf/Google.Protobuf.csproj
```diff
@@ -23,13 +23,14 @@
     <!-- Include PDB in the built .nupkg -->
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
     <IsTrimmable>true</IsTrimmable>
+    <EnableAOTAnalyzer>true</EnableAOTAnalyzer>
```
jskeet, line 26: Is this available with the .NET 6 SDK, if we *do* stick with that?

csharp/src/Google.Protobuf/Reflection/ReflectionUtil.cs
```diff
@@ -304,7 +328,28 @@ public void ClearExtension(IMessage message)
...
+        {
+            private static readonly object[] EmptyObjectArray = new object[0];
+            public Action<IMessage> CreateActionIMessage(MethodInfo method) => message => method.Invoke(message, EmptyObjectArray);
+            public Action<IMessage, object> CreateActionIMessageObject(MethodInfo method) => (message, arg) => method.Invoke(message, new object[] { arg });
+            public Func<IMessage, bool> CreateFuncIMessageBool(MethodInfo method) => message => (bool) method.Invoke(message, EmptyObjectArray);
+            public Func<IMessage, int> CreateFuncIMessageInt32(MethodInfo method) => message => (int) method.Invoke(message, EmptyObjectArray);
+            public Func<IMessage, object> CreateFuncIMessageObject(MethodInfo method) => message => method.Invoke(message, EmptyObjectArray);
+        }
+
+        private sealed class AotExtensionReflectionHelper : IExtensionReflectionHelper
```
jskeet, line 342: Please could you add a comment explaining *why* it's not supported? (That way, if things change we could potentially fix it later.)

global.json
```diff
@@ -1,6 +1,6 @@
 ﻿{
   "sdk": {
-    "version": "6.0.100",
+    "version": "7.0.100",
```
jskeet, line 3: If we go ahead with my suggestion of "you only need .NET 7 to run the AOT test project", I assume we'd revert this but add another global.json in the AOT test project directory. Does that sound feasible?

jskeet (commented):
Code seems fine - my main concern is around requiring .NET 7 and C# 11.

csharp/src/Directory.Build.props
JamesNK: I used list pattern matching in some tests. I can revert.

csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Descriptors.cs
JamesNK: I added the slim `Assert` class after I wrote this. Can update.

csharp/src/Directory.Build.props
jskeet: Thank you - much appreciated :)

jskeet: (Although if that's only in the project that relies on .NET 7, it would be fine to specify LangVersion 11 in that specific project file.)

JamesNK: The main Google.Protobuf csproj overrides this setting and forces C# 10 on itself. I think you already have the situation you desire.

By the way, I noticed that updating Google.Protobuf to use C# 11 introduced new compilation errors in Google.Protobuf. It doesn't like how the library uses ref structs. I didn't investigate.

jskeet: > I think you already have the situation you desire.

Well, not really. With this change it's "every project which doesn't specify anything gets C# 11, but Google.Protobuf specifies C# 10" whereas I'd prefer "every project which doesn't specify anything gets C# 10, but the AOT project gets C# 11". (This change would currently still be a problem internally where we'll be able to compile C# 10 but not C# 11 - we definitely want to be able to build Google.Protobuf.Test etc.)

csharp/src/Google.Protobuf/Google.Protobuf.csproj
JamesNK: It's new in .NET 7 SDK - https://github.com/dotnet/sdk/pull/24091

jskeet: Ah, that's interesting. So presumably the project *could* be built with the .NET 6 SDK and the property would be ignored, but with .NET 7 it can sniff out problematic bits? Hmm.

csharp/src/Google.Protobuf.Test.NativeAot/Reflection/Descriptors.cs
JamesNK: Immediately above this is a method that uses enum descriptors 😄

JamesNK:
PR feedback is done.

Is there a problem with updating to the .NET 7 SDK?  A few years ago, the SDK was .NET 5 to support some new features. There wasn't any issue.

jskeet:
> A few years ago, the SDK was .NET 5 to support some new features. There wasn't any issue.

I'm pretty sure I didn't review that PR, or I might have been concerned :)

The problem is that it forces anyone working on the runtime to install a non-LTS version of the SDK. My gut feeling is that a lot of companies and individuals will only want to install LTS versions of the SDK (and depend on LTS versions of .NET). Not all, certainly - but enough to make this at least something to consider.

Within Google, it definitely makes things more difficult if we need C# 11 support. I'm not going to reveal internal details, but keeping to just LTS SDK versions would reduce maintenance.

This isn't a complete blocker - I just want us to think about the feasibility of sticking to LTS SDK versions.

(I'll do a more complete review of the PR later; I haven't really started my working day yet.)

csharp/src/Google.Protobuf.Test.NativeAot/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#endregion
+
+using Google.Protobuf.Test.NativeAot.Reflection;
+
+try
+{
+    // This project tests specific Google.Protobuf features with .NET Native AOT.
+    // It is designed to be published as a native app, and then run.
+    // Return value of 100 is success, 101 is failure.
```
jskeet, line 39: Any reason for returning 100 vs 101, instead of 0 vs 1? The latter would be simpler to handle in shell scripts etc, where 0 is normally "success" and anything else is failure.

(We could potentially have an `Action[]` and test each in turn, catching exceptions, and returning the number of failures.)

Sorry for not spotting this before.

jskeet (commented):
One nit, but other than that I think we're good to go once we've decided on the "make everyone install .NET 7" vs "only need to install .NET 7 to run the AOT tests" part.

Although thinking about it more, presumably we have to be using the .NET 7 SDK in order to the get benefits of the EnableAOTAnalyzer property, which is in Google.Protobuf. Hmm. I'll need to think about that more - and check with the Protobuf team about their release system.

I'll ping the team an email today, but I'm on vacation tomorrow, so it'll probably be Monday before we can get to a decision. If this is blocking and urgent, let me know and I'll see what I can do.

csharp/src/Google.Protobuf.Test.NativeAot/Program.cs
JamesNK: I followed some other tests which used 100 and 101. I've changed it to 0 and 1.

Using `Action[]` would be nice, but something to consider with AOT and trimming is running things in isolation. I think the ideal approach is adding `#ifdef` and then building and running the app for each test, passing in the test name as a compiler constant. I added that to the comment.

jskeet:
@JamesNK: I've spoken to the team, and we think it's okay to require the .NET 7 SDK for building. That resolves a bunch of my comments, basically :) We need to make sure we stick with C# 10 as the language version though - so thanks for amending that already.

I'm just checking whether we need to make our release pipeline changes before merging this PR, or whether we can merge first. Are you basically happy for this to be merged when everything else is good to go, or are there any other changes you want to make?

jskeet:
@JamesNK: Could you rebase this onto the current head - or are you happy for me to do so? I believe everything should be in place now for this to be merged in the new flow, but I suspect it's more likely to go through if it's synced against a recent commit.

JamesNK:
This one is still blocked. I rebased https://github.com/protocolbuffers/protobuf/pull/11128

JamesNK:
Blocked on repo using .NET 7 SDK. Will reopen when SDK is successfully updated.

Issue #10965, opened 2022-11-10
Any reason why protobuf doesn't use the last version of bazel skybuild?


Clement-Jean:
**What version of protobuf and what language are you using?**

`libprotoc 3.21.6`

**What operating system (Linux, Windows, ...) and version?**

MacOS, Apple M1 Max, Ventura 13.0

**What did you do?**

I defined a Bazel workspace from scratch, such as:

```bazel
workspace(name = "my_workspace")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "rules_cc",
    urls = ["https://github.com/bazelbuild/rules_cc/releases/download/0.0.4/rules_cc-0.0.4.tar.gz"],
    sha256 = "af6cc82d87db94585bceeda2561cb8a9d55ad435318ccb4ddfee18a43580fb5d",
    strip_prefix = "rules_cc-0.0.4",
)

http_archive(
    name = "rules_proto",
    sha256 = "80d3a4ec17354cccc898bfe32118edd934f851b03029d63ef3fc7c8663a7415c",
    strip_prefix = "rules_proto-5.3.0-21.5",
    urls = ["https://github.com/bazelbuild/rules_proto/archive/refs/tags/5.3.0-21.5.tar.gz"],
)

load("@rules_cc//cc:repositories.bzl", "rules_cc_dependencies")
rules_cc_dependencies()

load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_proto_toolchains")
rules_proto_dependencies()
rules_proto_toolchains()
```

then, I have an empty BUILD file in my root and the following BUILD file in the proto directory:

```bazel
load("@rules_cc//cc:defs.bzl", "cc_library")
load("@rules_proto//proto:defs.bzl", "proto_library")

cc_proto_library(
    name = "account_cc_proto",
    deps = [":account_proto"],
)

proto_library(
    name = "account_proto",
    srcs = ["account.proto"]
)
```

**What did you expect to see**

I expected it to create my library.

**What did you see instead?**

if I run `bazel build //proto:account_cc_proto`, I have the following error:

```
[...]

PR #2530, opened 2016-12-22
Podsvirov CMake changes backported to 3.0.x

rpavlik:
These are CMake changes, including an essential bugfix for `protobuf_MODULE_COMPATIBLE`, originally found in #1865 by @podsvirov - all I did was a `git rebase --onto` to get them on the 3.0.x branch.

PR #10979, opened 2022-11-14
Changing bazel skylib version from 1.2.1 to 1.3.0

Clement-Jean:
This intends to be a first step solving issue #10965 and #2530. It just updates bazel skylib to version 1.3.0.

### DIFF
--- a/protobuf_deps.bzl
+++ b/protobuf_deps.bzl
@@ -30,10 +30,10 @@ def protobuf_deps():
         http_archive(
             name = "bazel_skylib",
             urls = [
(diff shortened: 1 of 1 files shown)


Clement-Jean:
Is this a problem related to Bazel ? Any way to check ?

Clement-Jean:
@acozzette Anything we can do to solve this? I'm trying to use bazel for my students but telling them that we need to add skylib in the WORKSPACE for no other reason than make it work is kind of not attractive for them.

acozzette (approved):

PR #11011, opened 2022-11-17
Make generated python files compatible with Cython

### LINKED ISSUE #10800: Python generated code is no longer compatible with Cython since 3.20
Version: 4.21.8
Language: Python

Before the 3.20 update, the generated code was statically defining all the messages, and could be given to cython without issues.
Since the 3.20 update, those messages are dynamically generated, leading to a cython error.

For example, given this proto file

```proto
syntax = "proto3";

message Foo {
    bool a = 1;
}
```

The codegen gives this:

`$ protoc --python_out=gen proto/a.proto`

```py
# -*- coding: utf-8 -*-
# source: proto/a.proto
"""Generated protocol buffer code."""
from google.protobuf.internal import builder as _builder
from google.protobuf import descriptor as _descriptor
from google.protobuf import descriptor_pool as _descriptor_pool
from google.protobuf import symbol_database as _symbol_database
# @@protoc_insertion_point(imports)

_sym_db = _symbol_database.Default()




DESCRIPTOR = _descriptor_pool.Default().AddSerializedFile(b'\n\rproto/a.proto\"\x10\n\x03\x46oo\x12\t\n\x01\x61\x18\x01 \x01(\x08\x62\x06proto3')

_builder.BuildMessageAndEnumDescriptors(DESCRIPTOR, globals())
_builder.BuildTopDescriptorsAndMessages(DESCRIPTOR, 'proto.a_pb2', globals())
if _descriptor._USE_C_DESCRIPTORS == False:

  DESCRIPTOR._options = None
  _FOO._serialized_start=17
  _FOO._serialized_end=33
# @@protoc_insertion_point(module_scope)
```

This can no longer be compiled by cython:

```sh
$ cython gen/proto/a_pb2.py
Error compiling Cython file:
------------------------------------------------------------
...
_builder.BuildMessageAndEnumDescriptors(DESCRIPTOR, globals())
_builder.BuildTopDescriptorsAndMessages(DESCRIPTOR, 'proto.a_pb2', globals())
if _descriptor._USE_C_DESCRIPTORS == False:

  DESCRIPTOR._options = None
  _FOO._serialized_start=17
 ^
------------------------------------------------------------

gen/proto/a_pb2.py:23:2: undeclared name not builtin: _FOO
```

Cython does not dynamically check variables from modifications of `globals()`, and thinks the variable is not set.

This isn't really a regression since I suppose you do not support Cython. However, a small change could make this work with native python and cython:

Instead of generating this:

`_FOO._serialized_start=17`

generating this would fix the issue:

`globals()["_FOO"]._serialized_start=17`

This is not an easy fix to do with some post processing of the generated files with regexes and seds and stuff like this, but shouldn't be too hard to do in the code generator I suppose.

Would you be OK with such a change? That would be really useful for cython users. I can try to make this change if needed.

vthib:
Replace access to variables created in globals() by explicit access to the globals() array. This avoids static analysis of the code to error out on those unknown variables, and make the files cythonable.

Fixes #10800

### DIFF
--- a/src/google/protobuf/compiler/python/generator.cc
+++ b/src/google/protobuf/compiler/python/generator.cc
@@ -298,6 +298,7 @@ bool Generator::Generate(const FileDescriptor* file,
   PrintTopBoilerplate();
   PrintImports();
   PrintFileDescriptor();
+  printer_->Print("_globals = globals()\n");
   if (GeneratingDescriptorProto()) {
     printer_->Print("if _descriptor._USE_C_DESCRIPTORS == False:\n");
     printer_->Indent();
@@ -312,7 +313,7 @@ bool Generator::Generate(const FileDescriptor* file,
   // Find the message descriptors first and then use the message
   // descriptor to find enums.
   printer_->Print(
-      "_builder.BuildMessageAndEnumDescriptors(DESCRIPTOR, globals())\n");
+      "_builder.BuildMessageAndEnumDescriptors(DESCRIPTOR, _globals)\n");
   if (GeneratingDescriptorProto()) {
     printer_->Outdent();
   }
@@ -323,7 +324,7 @@ bool Generator::Generate(const FileDescriptor* file,
   }
   printer_->Print(
       "_builder.BuildTopDescriptorsAndMessages(DESCRIPTOR, '$module_name$', "
-      "globals())\n",
+      "_globals)\n",
       "module_name", module_name);
   printer.Print("if _descriptor._USE_C_DESCRIPTORS == False:\n");
   printer_->Indent();
@@ -345,7 +346,7 @@ bool Generator::Generate(const FileDescriptor* file,
   printer_->Outdent();
   if (HasGenericServices(file)) {
     printer_->Print(
-        "_builder.BuildServices(DESCRIPTOR, '$module_name$', globals())\n",
+        "_builder.BuildServices(DESCRIPTOR, '$module_name$', _globals)\n",
         "module_name", module_name);
   }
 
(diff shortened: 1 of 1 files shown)


deannagarcia:
This got closed by a bot because it was submitted separately here: https://github.com/protocolbuffers/protobuf/commit/e3ea0441a328214492b717442887ba63dcc03c7f. We use a tool that sometimes does this where it closes a PR and then submits the commit separately.

On a manual check, I'm pretty sure this got into the 22.x branch (https://github.com/protocolbuffers/protobuf/commit/e3ea0441a328214492b717442887ba63dcc03c7f), so it should be in the 22.0/22.1 release. Are you using that version? If you still see this issue in 22, please open a new issue so we can investigate it separately.

garysferrao:
@deannagarcia thank you for your reply. it works with v22 release!~ 😄

sorry it took so long to revert. i was unable to test it properly in my CI flow, mainly because it's not available in [grpcio-tools](https://pypi.org/project/grpcio-tools/) v1.53.0 yet. i think i have to wait for it to be released here to properly list it in dependencies and CI tools.
```
₹ pip install grpcio-tools # v1.53.0 as of 2023-03-31
₹ python -m grpc_tools.protoc --version
libprotoc 3.21.12
₹ python -m grpc_tools.protoc -I=. --python_out=. --grpc_python_out=. --mypy_out=. ./xxx.proto
₹ python -m flake8 ./
xxx_pb2.py:nnn:5: F821 undefined name '_XXXREQUEST'
…
```

‌

but i generated the `*pb2.py` file with v22.2 from the release page.
```
₹ wget https://github.com/protocolbuffers/protobuf/releases/download/v22.2/protoc-22.2-linux-x86_64.zip -o ~/protoc/protoc-22.2-linux-x86_64.zip
₹ unzip ~/protoc/protoc-22.2-linux-x86_64.zip

₹ ~/protoc/bin/protoc --version
libprotoc 22.2
₹ ~/protoc/bin/protoc -I=. --python_out=. ./*.proto
₹ python -m flake8 ./
# no error related to global variables
```

thanks for your help. 🙇

Issue #10078, opened 2022-05-30
Add compiler version as a comment in _pb2.py files


simoncozens:
**What language does this apply to?**

Python, but probably others too.

**Describe the problem you are trying to solve.**

There are some compiled protobuf files in a library I am using. When trying to use that library, I just got an error of the form

```
 If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
```

So pb2.py files are linked to certain versions of the protobuf library, which is fine. To make that library work, I need to install the correct version of protobuf. But which version was it? The generated files don't tell me.

**Describe the solution you'd like**

The pb2.py files should contain a comment like:

```
# Generated by protoc 3.x.y.  DO NOT EDIT!
```

**Describe alternatives you've considered**

Sure, the library itself should set a dependency on the right version of protobuf, but it doesn't. This shouldn't be too hard to add and would probably help with debugging other version incompatibility problems too.

fowles:
All generators should include the version of the compiler and plugin that generated them.  At minimum as comments, even better would be in a way that the host language can introspect on.

haberman:
Agreed, this would be good to add to the Python generator.

anandolee:
Thanks for the suggestion. I've created a feature request in google internal to make this a higher priority.

PR #11015, opened 2022-11-17
Add compiler version to comment for all generators

simoncozens:
#10078 bit me on the backside again today, so I just implemented it.


jskeet:
#10862 would already add the version number to the generated C# code, in a machine-readable way (and still fairly close to the top of the file, so easy to find). That change would end up with the version number in the file twice (once for protoc version, once for minimum runtime)... I think I'd prefer to avoid having a third change per file when updating protoc if possible.

fowles:
Various generators like Ruby have golden tests that this change breaks.  Can you improve those tests to not be sensitive to the version in the headers?

PR #11029, opened 2022-11-19
Fix segfaults on FreeBSD

mfechner:
If more than one version of rubygem google-protobuf is installed it causes problem. Make sure symbols are also hidden on FreeBSD like it is already done for Linux and Darwin.

This fixes problems reported here:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=266925
https://gitlab.com/gitlab-org/gitlab/-/issues/345693

### DIFF
--- a/ruby/ext/google/protobuf_c/extconf.rb
+++ b/ruby/ext/google/protobuf_c/extconf.rb
@@ -6,7 +6,7 @@ ext_name = "google/protobuf_c"
 
 dir_config(ext_name)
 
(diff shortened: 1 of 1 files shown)


mfechner:
I'm sorry, I think this is a duplicate to the following MR:
https://github.com/protocolbuffers/protobuf/pull/10832

googleberg:
Thanks for the help! https://github.com/protocolbuffers/protobuf/pull/10832 has been incorporated. Closing this duplicate.

Issue #5144, opened 2018-09-13
Need to manually add ./google/protobuf/.libs/descriptor.pb.o to get protoc to link


sweetcw:
Link error: undefined reference to 'scc_info_FileDescriptorProto_google_2fprotobuf_2fdescriptor_2eproto' 

Using an ARM docker to compile for ARM Yocto environment.  Apply patch found [here ](https://raw.githubusercontent.com/openembedded/meta-openembedded/master/meta-oe/recipes-devtools/protobuf/protobuf/0001-protobuf-fix-configure-error.patch) so autogen will work.

bash-4.3# g++ --version
g++ (GCC) 4.9.3
bash-4.3# libtool --version
libtool (GNU libtool) 2.4.6

Commands:
bash-4.3# ./autogen
bash-4.3# ./configure
bash-4.3# make

Get the following error:
/bin/bash ../libtool  --tag=CXX   --mode=link g++ -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare -O2 -g -std=c++11 -DNDEBUG -pthread  -o protoc goo
gle/protobuf/compiler/main.o  libprotobuf.la libprotoc.la -lz 
libtool: link: g++ -pthread -DHAVE_PTHREAD=1 -DHAVE_ZLIB=1 -Wall -Wno-sign-compare -O2 -g -std=c++11 -DNDEBUG -pthread -o .libs/protoc google/protobuf/compiler/main
.o  ./.libs/libprotobuf.so ./.libs/libprotoc.so /home/root/protobuf/src/.libs/libprotobuf.so /usr/lib/libstdc++.so -lm -lz -pthread -Wl,-rpath -Wl,/usr/local/lib
./.libs/libprotoc.so: error: **undefined reference to 'scc_info_FileDescriptorProto_google_2fprotobuf_2fdescriptor_2eproto'**
collect2: error: ld returned 1 exit status
Makefile:3499: recipe for target 'protoc' failed
make[2]: *** [protoc] Error 1
make[2]: Leaving directory '/home/root/protobuf/src'
Makefile:1559: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/root/protobuf'
Makefile:1466: recipe for target 'all' failed
make: *** [all] Error 2

To resolve I call the following:
bash-4.3# cd src
[...]

PR #11032, opened 2022-11-21
Fix shared object ABI exports (#5144)

LourensVeen:
On some linkers, the ABI definitions in src/libprotoc.map and friends do not match all symbols that need to be exported. This causes a linker error when trying to use the resulting shared object library: `undefined reference to 'scc_info_FileDescriptorProto_google_2fprotobuf_2fdescriptor_2eproto'`

Quite a few people seem to be affected by this, see #5144 for discussion. It seems that yocto now applies this patch itself, but that doesn't help people building in other places (I encountered this on some HPC machines), so it would probably be good to apply it upstream as well.

### DIFF
--- a/src/libprotobuf-lite.map
+++ b/src/libprotobuf-lite.map
@@ -3,6 +3,8 @@
     extern "C++" {
       *google*;
     };
+    scc_info_*;
+    descriptor_table_*;
 
   local:
     *;
--- a/src/libprotobuf.map
+++ b/src/libprotobuf.map
@@ -3,6 +3,8 @@
     extern "C++" {
       *google*;
     };
+    scc_info_*;
+    descriptor_table_*;
 
   local:
     *;
--- a/src/libprotoc.map
+++ b/src/libprotoc.map
@@ -3,6 +3,8 @@
     extern "C++" {
       *google*;
     };
+    scc_info_*;
+    descriptor_table_*;
 
   local:
     *;


googleberg:
@LourensVeen Please sign the CLA so that we can assign a reviewer to this PR. More information [here](https://github.com/protocolbuffers/protobuf/pull/11032/checks?check_run_id=9610683591)

LourensVeen:
Hi, I'm working on it, but I'm doing this at work so we have to sign the corporate version, and it has to go through channels. The text shouldn't be an issue, we're just waiting for some very busy people to have a moment to give me the official go-ahead to set it up and then to sign it.

mkruskal-google:
Can you also rebase this PR?  We recently overhauled our CI infrastructure and our tests would fail in the current state

LourensVeen:
Rebased, and I've set the CLA process in motion, that should be a signature away from passing.

LourensVeen:
The CLA is signed now, and I've tried to do a rescan following the documentation, but I'm getting a 400 - That's an error - That's all we know. The URL on that page ends with `expiry=2023-01-05` which if it's the 5th of January doesn't sound right.

I think someone at Google should be able to re-run the CLA check though, could you try?

LourensVeen:
I tried again, and rescanning worked now.

LourensVeen:
I think I've done all I can now. I'm not sure how it's used exactly, but perhaps the "wait for user action"  label can be removed now?

fowles (approved):

PR #11034, opened 2022-11-21
Amend repeated_ptr_fields.h template declearation

feast107:
This causes various continuous lines of compilation error within vs.


googleberg:
@feast107 Please sign the CLA so we can assign a reviewer to this PR. For more information see: https://github.com/protocolbuffers/protobuf/pull/11034/checks?check_run_id=9617744316

feast107:
> @feast107 Please sign the CLA so we can assign a reviewer to this PR. For more information see: https://github.com/protocolbuffers/protobuf/pull/11034/checks?check_run_id=9617744316

Thanks for your reminding, I've now signed the CLA.

feast107:
I would like to know the build situation this time

googleberg:
@feast107 I think you'll need to rebase this PR to get past a testing breakage.

deannagarcia:
@feast107 If you rebase this PR I can rerun the tests for you!

src/google/protobuf/repeated_ptr_field.h
```diff
@@ -693,11 +693,19 @@ class PROTOBUF_EXPORT RepeatedPtrFieldBase {
   Rep* rep_;
 
   template <typename TypeHandler>
-  static inline typename TypeHandler::Type* cast(void* element) {
+  #ifdef _MSC_BUILD 
+  static inline typename TypeHandler::Type* (cast) (void* element){
```
mcy, line 697: This makes me suspect some kind of rude macro is getting pushed into our namespace. Do you know where it comes from? I don't think `windows.h` has a `cast` macro. We should figure out what's causing this so we can fix this globally in `port_def.inc`.

feast107: Thanks for your patience, it derives from my initial problem which is like when I include ```repeated_ptr_field``` will cause compiler error, changing the template declaration like above will solve this problem and successfully build. I m not sure what will happen on other platforms.

haberman: Does it work to `#undef cast` in the list of `push_macro` calls in https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/port_def.inc ?

Also add pop_macro in: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/port_undef.inc

haberman:
btw, you will want to rebase on our main branch, there have been a lot of changes in our testing infrastructure in the last few weeks.

PR #11035, opened 2022-11-21
Update to .NET 7 SDK

deannagarcia:
This is in preparation for https://github.com/protocolbuffers/protobuf/pull/10978. We will now need to build with .NET 7, but will keep other versions installed to test with.

### DIFF
--- a/csharp/install_dotnet_sdk.ps1
+++ b/csharp/install_dotnet_sdk.ps1
@@ -18,3 +18,4 @@ Invoke-WebRequest -Uri $InstallScriptUrl -OutFile $InstallScriptPath
 # installed by kokoro/linux/dockerfile/test/csharp/Dockerfile
 &$InstallScriptPath -Version 3.1.415
 &$InstallScriptPath -Version 6.0.100
+&$InstallScriptPath -Version 7.0.100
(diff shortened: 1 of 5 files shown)


JamesNK:
Something you could try is updating the test dependencies here: https://github.com/protocolbuffers/protobuf/blob/8809a113bc0a00b685b787a03d0698aa8c3e10e8/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj#L17-L19

Try updating Microsoft.NET.Test.Sdk to 17.4.0 first - https://www.nuget.org/packages/Microsoft.NET.Test.SDK

deannagarcia:
> Something you could try is updating the test dependencies here:
> 
> https://github.com/protocolbuffers/protobuf/blob/8809a113bc0a00b685b787a03d0698aa8c3e10e8/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj#L17-L19
> 
> Try updating Microsoft.NET.Test.Sdk to 17.4.0 first - https://www.nuget.org/packages/Microsoft.NET.Test.SDK

Thanks for the tip! I just tried that and unfortunately still seeing the same failures 😞

Issue #658, opened 2015-07-30
Add DynamicMessage support for C#


jtattermusch:
Not a requirement as of now, but it would be nice to have one day in not too distant future.

jskeet:
It would be nice to have specific requirements around this - scenarios we're trying to make simple.
A few questions that arise:

- Do we need reflection support? (I assume so.)
- Will these messages be mutable, just like the rest? (I assume so.)
- Do we need to be able to mix and match, e.g. set a field in a dynamic message to a value which is a strongly-typed message? (Could be tricky.)
- Do we need C# dynamic typing support?

xfxyjwf:
@jskeet @anandolee Is this DynamicMessage implemented in C#?

jskeet:
Not that I'm aware of.

dluc:
The simpler, still very useful, support would be removing the need to compile .proto files to DLLs at build time, so that a .NET application could generate the serializer/deserializer at runtime (caching it in memory) from .proto files provided by users for example.

jskeet:
@dluc: That's *far* from simple - the C# code has no facility for parsing .proto files at all. That's all in protoc. If any dynamic message support is introduced, it would almost certainly work with the descriptors, not the original .proto files.

(And generating the code at execution time is far from trivial too - again, all the code generation is in protoc, not in the Google.Protobuf library.)

jskeet:
I'm actively working on this now. The initial support is likely to be relatively bare, but it'll be there. The main reason for me doing this is for the sake of writing protoc plugins in C#.

mcon-gr:
@jskeet Is this something you're still working on? I'm happy to contribute if you consider this a parallelisable task.

jskeet:
I'm not working on it at the moment, no. Basically I don't have time to work on protobuf other than for features I need for other work, e.g. being able to *load* the descriptors dynamically. I'm afraid I wouldn't have time to review significant chunks of code either, but hopefully @anandolee will be able to help.

mcon-gr:
Thanks Jon - I'm getting the feeling that adding `DynamicMessage` to C# will involve some non-trivial work, once I get some time I'll start by taking my lead from the Java implementation.

jrimig:
@jskeet Thanks for adding `FileDescriptor.BuildFromByteStrings()`. Is this method intended to read the FileDescriptorSet generated by 

> protoc.exe --include_imports --descriptor_set_out=FILE 

[...]

PR #11092, opened 2022-11-30
Protobuf dynamic typing in C#

himkak:
* Added support for Dynamic typing by adding DynamicMessage for C#. The implementation closely follows the implementation done in java.
* Added dotnet workflow for build and test C# code


jskeet:
I'll have a look at this when I get some time, but it may not be for a while I'm afraid.

himkak:
Sure, thanks.
It would be great if you could review it soon.
Ref issue : https://github.com/protocolbuffers/protobuf/issues/658

himkak:
> I strongly suspect we won't be in a position to _merge_ until some time in 2023. Protobuf isn't my main role, so I'll only be able to put time into reviewing periodically. Do you have an urgent need for this for some reason?

Yes, I am using these changes in my project, and looking to use the released version of the library.
Might be, if you can also pull someone who looks into Protobuf, to review these changes.
I would suggest to let's start reviewing the changes and make the changes merge ready.

jskeet:
> Might be, if you can also pull someone who looks into Protobuf, to review these changes.

No, we really don't have enough C#-suitable resources - and close to the end of the year, lots of people are taking vacation time etc.
I'll try to have a quick look ahead of time to point out any large areas that need more work.

csharp/src/Google.Protobuf/Reflection/Dynamic/DynamicMessage.cs
```diff
@@ -0,0 +1,387 @@
+﻿using Google.Protobuf.Collections;
+using System;
+using System.Collections;
+
+namespace Google.Protobuf.Reflection.Dynamic
```
jskeet, line 5: I think we'll be able to use just Google.Protobuf.Reflection for this. I don't think a new sub-namespace is required.

csharp/src/Google.Protobuf/Reflection/Dynamic/DynamicMessage.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        /// <summary>
+        /// Descriptor for this message.
+        /// </summary>
+        public MessageDescriptor Descriptor => type;
+
+        /// <summary>
+        /// Default implementation of Overridden method.
+        /// This specific implementation is never used, mergeFrom is used from the builder class.
```
jskeet, line 31: This prohibits it from being used in code that just uses IMessage.

jskeet (commented):
Overall, this isn't going in a direction I'd want to merge. I'd expect DynamicMessage to support:

- Being handled like any other IMessage
- JSON parsing and formatting
- Reflection access via `message.Descriptor.FindField(...).Accessor.GetValue(message)` etc
- Unknown fields

This all requires quite a bit of work, which is why it hasn't been done yet.

Given that this PR doesn't need to modify any existing code, I suggest that if it satisfies your needs, you keep it within your own codebase for now, and then migrate to whatever full support we end up with eventually.

himkak:
Got your point. Let me try to make the changes in this PR, according to your suggestions.
If it will not look doable, will proceed as you suggested.

jskeet:
You're welcome to try - but having looked at this a few times (and having worked in this codebase for 14 years) it's always felt like a daunting prospect to me, with changes required across the reflection code as well as creating the core `DynamicMessage` type itself.

himkak:
I have created a new PR in my forked repo itself. It tries to follow the same pattern as other IMessage implementations do.
Have a quick look and if it looks close to what is expected, will raise a PR against the main repo branch.
[himkak protobuf PR 2](https://github.com/himkak/protobuf/pull/2)

jskeet:
I'm on vacation now until January. I'll have a look next time I'm working.

himkak:
Sure, Merry Christmas and Happy New Year in advance. Enjoy your vacation.

PR #11110, opened 2022-12-01
Add custom option numbers for use by Buf

jhump:
We are requesting a range of numbers instead of just one since we have some plans for a variety of orthogonal kinds of custom options. For example:
1. Injecting extra metadata into FileDescriptorProtos via custom options. (These options are inserted by the Buf tool when building proto sources, to encode additional metadata about the unit of files compiled.)
2. Configuration options for protoc plugins, that generate extra metadata into sources. (These options may be used by users when writing proto sources.)
 
The suggested approach of reserving just one number and making it a message that contains *all* options is unattractive since users of one category of options is unlikely to want to import types related to another category.

FWIW, there are several other rows above that have also chosen to reserve ranges of 5 or 10 numbers.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -202,7 +202,7 @@ with info about your project (name and website) so we can add an entry for you.
    * Extensions: 1070
 
 1. Protoc-gen-validate
-   * Website: https://github.com/envoyproxy/protoc-gen-validate
+   * Website: https://github.com/bufbuild/protoc-gen-validate
    * Extensions: 1071
 
 1. Protokt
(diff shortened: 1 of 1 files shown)


fowles:
Happy to approve this, but I am curious what "unit of files" means

jhump:
@fowles, by "unit of files", I just meant a compilation unit -- the set of files compiled together. This is often a "module" in the BSR, so we can add metadata about the module from which the files originated, such as its URL (e.g. https://buf.build/envoyproxy/protoc-gen-validate/) and version.

PR #11208, opened 2022-12-08
feat: more readable phpdoc escaping

bshaffer:
The PHPDoc escaping in PHP is aggressive in that it escapes some character sequences that don't need to be escaped (`/*`), and it uses HTML entities to escape others (`*/` and `@`) instead of the recommended PHPDoc escape sequences.

For Example, in [`Google\Api\RoutingParameter`](https://github.com/googleapis/common-protos-php/blob/main/src/Api/RoutingParameter.php#L42): 

```
 * path_template: "projects/&#42;&#47;{table_location=instances/&#42;}/tables/&#42;"
```

Should be escaped as:

```
 * path_template: "projects/{@*}{table_location=instances/*}/tables/*"
```

according to [the PHPDoc guide](https://manual.phpdoc.org/HTMLframesConverter/default/phpDocumentor/tutorial_phpDocumentor.howto.pkg.html#basics.desc):

 - For `@`: "if you need an actual "@" in your DocBlock's description parts, you should be careful to either ensure it is not the first character on a line, or else escape it ("\\@") to avoid it being interpreted as a PhpDocumentor tag marker." 

 - For `*/`: " If you need to use the closing comment "\*/" in a DocBlock, use the special escape sequence "{@*}."

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -1610,38 +1610,37 @@ static std::string EscapePhpdoc(absl::string_view input) {
   std::string result;
   result.reserve(input.size() * 2);
 
-  char prev = '*';
+  char prev = '\0';
 
   for (std::string::size_type i = 0; i < input.size(); i++) {
     char c = input[i];
     switch (c) {
-      case '*':
-        // Avoid "/*".
-        if (prev == '/') {
-          result.append("&#42;");
-        } else {
-          result.push_back(c);
-        }
-        break;
+      // "/*" is allowed, do nothing
+      // case '*':
+      //   if (prev == '/') {
+      //     result.append("&#42;");
+      //   } else {
+      //     result.push_back(c);
+      //   }
+      //   break;
       case '/':
-        // Avoid "*/".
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -1625,38 +1625,37 @@ static std::string EscapePhpdoc(const std::string& input) {
...
     switch (c) {
-      case '*':
-        // Avoid "/*".
-        if (prev == '/') {
-          result.append("&#42;");
-        } else {
-          result.push_back(c);
-        }
-        break;
+      // "/*" is allowed, do nothing
```
haberman, line 1633: Please remove this commented-out code. You could leave a comment that "/*" is allowed so we do not translate it.

bshaffer: done!

NeilMasters:
@haberman is there anything preventing this from being merged?

bshaffer:
@haberman sorry for the delay here. Done!

bshaffer:
This is an important feature, please don't just automatically close this

PR #11243, opened 2022-12-10
Fix PHP 8.2 dynamic property deprecation warning

edsrzf:
PHP 8.2 [deprecates dynamic properties](https://www.php.net/releases/8.2/en.php#deprecate_dynamic_properties). `MapFieldIter` triggers a warning on PHP 8.2 because it doesn't declare the `key_type` property. This PR declares the property to fix the warning.


edsrzf:
Addressed in #11485

PR #11312, opened 2022-12-15
[PHP] Serialize message to json add include default values and preserve fields name

s2x:


haberman:
Also please rebase, the testing infrastructure has changed a lot in the last few weeks. Sorry for the delay on this PR.

knobik:
Piotrze @s2x any updates on this?

haberman:
This PR makes several changes to the semantics of JSON serialization. We have to be careful, because we have to keep these semantics the same across languages.

None of our other languages currently serialize default values when serializing to JSON. So I don't think we can make that change in PHP.

This PR also is changing the behavior of the upb JSON serializer, so that JSON name is always used if it is explicitly provided, even if a user asked for proto field names. We cannot change this behavior, as it would cause divergence from what other languages do.

For example, here is the C++ logic for `preserve_proto_field_names`. Note that it always uses the proto field name if the option is set; it doesn't look to see if a `json_name` was set: https://github.com/protocolbuffers/protobuf/blob/671b61b523dbec8bf8e41a61d5a8f689c3e2bb9f/src/google/protobuf/json/internal/unparser.cc#L409-L410

s2x:
Hey, thanks for the quick reply. I read your comment, but I'm not sure if I understood it correctly.
> This PR makes several changes to the semantics of JSON serialization. We have to be careful, because we have to keep these semantics the same across languages.
> 
> None of our other languages currently serialize default values when serializing to JSON. So I don't think we can make that change in PHP.

Unfortunately, I don't really understand what the problem is. To make the default values serialized, you must provide the parameter `JsonSerializeOptions::EMIT_DEFAULTS` in `serializeToJsonString` function. Without this parameter, only non-empty fields are serialized. It appears that this behavior is already implemented in other languages

* java: https://github.com/protocolbuffers/protobuf/blob/671b61b523dbec8bf8e41a61d5a8f689c3e2bb9f/java/util/src/test/java/com/google/protobuf/util/JsonFormatTest.java#L1486
* ruby: ???https://github.com/protocolbuffers/protobuf/blob/671b61b523dbec8bf8e41a61d5a8f689c3e2bb9f/ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java#L703
* python: https://github.com/protocolbuffers/protobuf/blob/671b61b523dbec8bf8e41a61d5a8f689c3e2bb9f/python/google/protobuf/internal/json_format_test.py#L301

> This PR also is changing the behavior of the upb JSON serializer, so that JSON name is always used if it is explicitly provided, even if a user asked for proto field names. We cannot change this behavior, as it would cause divergence from what other languages do.
> 
> For example, here is the C++ logic for `preserve_proto_field_names`. Note that it always uses the proto field name if the option is set; it doesn't look to see if a `json_name` was set:
> 
> https://github.com/protocolbuffers/protobuf/blob/671b61b523dbec8bf8e41a61d5a8f689c3e2bb9f/src/google/protobuf/json/internal/unparser.cc#L409-L410

Okay, I didn't know. I will change the code to be compatible with C++ and keep `preserve_proto_field_names` priority over `json_names`

haberman:
> It appears that this behavior is already implemented in other languages

You are right, I misremembered. If this is implementing the behavior already supported by the `upb_JsonEncode_EmitDefaults` flag in upb, that should be fine.

s2x:
Hi @haberman,
I hope you had a great weekend. I've been busy with a few tasks lately, but today I finally made the necessary changes.

I've incorporated your suggestions in the latest commit, ensuring that PHP code is now compatible with C++ code. Now the priority is to keep `preserve_proto_field_names` instead of `json_names` as you mentioned earlier.

If you could take a moment from your schedule, I would really appreciate it if you could review my latest commit.

php/src/Google/Protobuf/JsonSerializeOptions.php
```diff
@@ -0,0 +1,40 @@
...
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+namespace Google\Protobuf;
+
+class JsonSerializeOptions
```
haberman, line 35: Could we name this Google\Protobuf\Json\SerializeOptions instead, so we can group all the JSON-related things together?

php/ext/google/protobuf/json_options.h
```diff
@@ -0,0 +1,22 @@
...
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+#ifndef PHP_PROTOBUF_JSON_OPTIONS_H_
+#define PHP_PROTOBUF_JSON_OPTIONS_H_
+
+// Define constants
+#define JSON_SERIALIZE_OPTIONS_EMIT_DEFAULTS "EMIT_DEFAULTS"
```
esrauchg, line 12: This flag was recently made consistently named in C++, Python and Java (also more verbose for clarity about what it affects), so it would be better to match here too:

JSON_SERIALIZE_OPTIONS_ALWAYS_PRINT_FIELDS_WITH_NO_PRESENCE

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -1574,11 +1597,11 @@ private function existField($field)
...
         $values = $this->$getter();
         if ($field->isMap()) {
-            return count($values) !== 0;
+            return $emitDefaults || count($values) !== 0;
         } elseif ($field->isRepeated()) {
-            return count($values) !== 0;
+            return $emitDefaults || count($values) !== 0;
         } else {
-            return $values !== $this->defaultValue($field);
+            return $values !== $this->defaultValue($field) || ($emitDefaults && $this->defaultValue($field) !== null);
```
haberman, line 1604: I think this should just be `$emitDefaults || $values !== $this->defaultValue($field)`.

Even if a default is not specified in the `.proto` file, the "default default" is 0 or an empty string. I don't think `$this->defaultValue($field)` should ever return `null`, except possibly message/repeated/map fields, but those should all have been handled above.

haberman (changes requested):

kiyakkoray:
If this PR is dead, Is there any other way to serialize default fields?

PR #11313, opened 2022-12-15
Make proguard feature a part of protolite library

### LINKED ISSUE #11252: Include valid proguard config in java protolite library
Related to 

https://issuetracker.google.com/issues/144631039
and
https://github.com/protocolbuffers/protobuf/issues/6463

Java Protolite

Include necessary proguard definitions to avoid failures in release builds.

Include https://github.com/protocolbuffers/protobuf/blob/520c601c99012101c816b6ccc89e8d6fc28fdbb8/java/lite/proguard.pgcfg in the library, so it's applied automatically.

See https://github.com/square/okhttp/blob/master/docs/features/r8_proguard.md for how this feature works in other libraries

https://raw.githubusercontent.com/square/okhttp/master/okhttp/src/jvmMain/resources/META-INF/proguard/okhttp3.pro

Currently every app needs to fail initially, search for the error message and include the same fix.

yschimke:
See https://github.com/square/okhttp/blob/master/docs/features/r8_proguard.md for how this feature works in other libraries

Fixes #11252

### DIFF
--- a/java/lite/proguard.pgcfg
+++ b/java/lite/proguard.pgcfg
@@ -1,4 +1,0 @@
-
-# Skip runtime check for isOnAndroidDevice().
-# One line to make it easy to remove with sed.
--assumevalues class com.google.protobuf.Android { static boolean ASSUME_ANDROID return true; }
--- a/java/lite/proguard.pgcfg
+++ b/java/lite/proguard.pgcfg
@@ -2,7 +2,3 @@
 # Skip runtime check for isOnAndroidDevice().
 # One line to make it easy to remove with sed.
 -assumevalues class com.google.protobuf.Android { static boolean ASSUME_ANDROID return true; }
-
--keepclassmembers class * extends com.google.protobuf.GeneratedMessageLite {
-  <fields>;
-}
--- a/java/core/BUILD.bazel
+++ b/java/core/BUILD.bazel
@@ -123,7 +123,8 @@ internal_gen_well_known_protos_java(
 java_library(
     name = "lite",
     resources = [
-        "src/main/resources/META-INF/proguard/protolite.pro"
+        "src/main/resources/META-INF/proguard/protolite.pro",
+        "src/main/resources/META-INF/com.android.tools/r8/protolite.pro",
     ],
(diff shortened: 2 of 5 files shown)


yschimke:
@googleberg What would sufficient tests for this be? Would it be failing if this is not applied, because the current config has those fields removed?

java/lite/proguard.pgcfg
```diff
@@ -2,7 +2,3 @@
 # Skip runtime check for isOnAndroidDevice().
 # One line to make it easy to remove with sed.
 -assumevalues class com.google.protobuf.Android { static boolean ASSUME_ANDROID return true; }
-
--keepclassmembers class * extends com.google.protobuf.GeneratedMessageLite {
```
perezd, line 6: does this also need to be removed from here, other things may be depending on this? Also, the -assumevalues bit, should that exist in the new file?

yschimke: I believe the assumevalues is something related to this build and testing. https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/Android.java#L39

The point of this change is to put the config we want all apps to get into the inherited location. See https://twitter.com/jakewharton/status/1004401938467876865

I hope nothing should rely on the keepclassmembers from here, because hopefully everything is picking up the new proguard file in the right location.  But I'd like to flush out errors to prove this is true.

yschimke: cc @agrieve Not sure if you can provide insight into where the ASSUME_ANDROID proguard rules should live? Are they required to be shipped with the published libraries?

yschimke: I've added that file. Intellij gives a lot of errors for it.

I've removed the original file, because I think we want to see that it's building and passing tests with the new file, not saved by the old file.

But happy to revisit.

yschimke:
Looks like it's better to make this change internally, as source code is not exact 1:1

yschimke:
I tried and failed, I couldn't get it integrated with the internal build, so closing.

PR #11357, opened 2022-12-19
Don't use int as index parameter

2A5F:


mcy (approved):

jorgbrown:
Could you clarify what problem this is solving for you?

2A5F:
Implicit conversions are disabled by lint in my cef project

2A5F:
size_t -> int -> size_t

jorgbrown:
> Implicit conversions are disabled by lint in my cef project

Can you be more specific about where you're seeing these unwanted conversions?

I'd like to avoid adding extra overloads to such a widely-used API; if I can change the callers, that might be preferable.

2A5F:
https://github.com/protocolbuffers/protobuf/blob/fbf491a397e8799e47128063a3aca53ebdf58897/src/google/protobuf/has_bits.h#L81-L83

https://github.com/protocolbuffers/protobuf/blob/fbf491a397e8799e47128063a3aca53ebdf58897/src/google/protobuf/has_bits.h#L48

2A5F:
at line 82

`rhs[i]`

sbenzaquen:
Adding extra overloads is not a good idea. If the old one is not used, then it adds dead code. If the old one is used, then it potentially adds ambiguity on some callers.
Fixing the existing overloads to only take `size_t` would be ok. The template parameter is already `size_t`, so it makes sense to make them the same.

However, imposing your local lint rules on third party project is a losing battle.
This is not the only place where implicit conversions happen in protobuf and we will do not prevent new ones from happening. We can't support arbitrary lint rules that are imposed by users.

2A5F:
It's not actually lint, it just turns on treating warnings as errors, warning C4245

sbenzaquen:
Who is turning on these warnings?

> A third-party library should not have warnings

It is impossible for a library to satisfy all warnings that a user could turn on. Some warnings are actually contradictory with each other.

2A5F:
> Who is turning on these warnings?

https://bitbucket.org/chromiumembedded/cef/src/178a44bd182bbc3c9ccfca70f53807e94b6a9fc7/cmake/cef_variables.cmake.in#lines-408

PR #11405, opened 2022-12-25
Fix Int64Value's jstype

mkvce:
The javascript `number` type capacity is equal to int32 type. So after parsing Int64Value, the result is not necessarily correct. The JS should parse it as an string.

### DIFF
--- a/src/google/protobuf/wrappers.proto
+++ b/src/google/protobuf/wrappers.proto
@@ -71,7 +71,7 @@ message FloatValue {
 // The JSON representation for `Int64Value` is JSON string.
 message Int64Value {
   // The int64 value.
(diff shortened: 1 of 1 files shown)


lukesandberg:
I believe this is a fundamentally good idea, however from a compatibility perspective it is unworkable in my opinion.

This will change the types of generated accessors in the protobuf-javascript implementation and thus will be a breaking change for clients.  Internally at google, this also affects certain wire formats which may lead to dataloss or type errors when clients and servers have different ideas of the js_type of this field.

In the long run protobuf-javascript plans to stop relying on this option altogether and introduce new types instead of `number` for 64 bit fields.

Issue #11085, opened 2022-11-29
Add version to header comment of generated source.


jenstroeger:
**What language does this apply to?**

I’m raising it for Python but this change would probably be useful for all supported languages.

**Describe the problem you are trying to solve.**

Frequent version changes of the gRPC tools require generated code to keep in step with tools changes. That can be difficult to maintain.

**Describe the solution you'd like**

Generated code should contain the version number of the protoc it was built with, e.g. for generated Python code:
```
# Generated by the protocol buffer compiler: libprotoc 3.19.5.  DO NOT EDIT! 
```

**Describe alternatives you've considered**

I currently wrap the `protoc` invocation in a shell script and use `sed` to inject that version number.

**Additional context**

Related issue https://github.com/grpc/grpc/issues/31765

esorot:
Hi @jenstroeger, this is actually something we've been wanting to do. Are you interested in adding this feature as a contributor?

jenstroeger:
@esorot happy to contribute, if you could please point me into the right directtion?

jenstroeger:
@esorot let me know if you have guidance?

jenstroeger:
I’d like to see this addressed.

jenstroeger:
I’d like to see this addressed.

jenstroeger:
I’d like to see this addressed.

PR #11408, opened 2022-12-26
Add version number to generated files

alielbashir:
This adds the version of the protoc compiler that generated the python files to the header of the generated files as specified in #11085

Edit: this PR now includes
- Python
- C++
- Java
- C#
- Objective C
- Ruby
- PHP


alielbashir:
If this feature is also wanted for the other languages should I include that within this PR, or should different PRs be opened for each language?

deannagarcia:
> If this feature is also wanted for the other languages should I include that within this PR, or should different PRs be opened for each language?

It would be great if you could add them to this PR!

alielbashir:
I've noticed there are unit tests in the ruby compiler directory. From what I understand they compare output to the expected files in that dir. This will break that test, and will make it hard to maintain it in the future as the output will change with every new version. 

1. Are there similar tests for the other languages as well? If so how should I move forward with this so that they don't fail, while not making them a maintenance issue?
2. Should I contribute this change to the Go, JavaScript, and Dart compilers too, given that they're in other repos?

deannagarcia (approved):

deannagarcia:
> I've noticed there are unit tests in the ruby compiler directory. From what I understand they compare output to the expected files in that dir. This will break that test, and will make it hard to maintain it in the future as the output will change with every new version.

> 1. Are there similar tests for the other languages as well? If so how should I move forward with this so that they don't fail, while not making them a maintenance issue?

Talked with the team about this and we are going to create an easier way to update the expected output for changes like this to decrease the maintenance burden. Can you make the manual change for the Ruby expected files for this time though?


> 2. Should I contribute this change to the Go, JavaScript, and Dart compilers too, given that they're in other repos?

Up to you, we don't work on those repos and it's not organized the same so it may be a bit of extra work for you but I think those repo managers would appreciate it.

alielbashir:
> Talked with the team about this and we are going to create an easier way to update the expected output for changes like this to decrease the maintenance burden. Can you make the manual change for the Ruby expected files for this time though?

Done

deannagarcia:
I think you also need to run [generate_descriptor_proto.sh](https://github.com/protocolbuffers/protobuf/blob/8870581adf5d95988bcc995a870972caa8d63e8c/generate_descriptor_proto.sh) to update C++ tests that compare the generated output.

deannagarcia:
Sorry I'm just getting back to this, can you update your branch and fix conflicts so I can rerun the tests?

alielbashir:
Done, please let me know if there's anything else I need to do

deannagarcia (approved):

haberman:
Could you rebase this PR against the main branch? Sorry for the trouble, you caught us in the middle of a migration to GitHub Actions.

alielbashir:
Sure thing, doing it now

alielbashir:
Apologies, I accidentally pushed an incorrect rebase, which could be the reason for the protobuf-btr and protobuf-docs teams' review requests

alielbashir:
@deannagarcia it seems that the PR isn't passing the safety check despite having the "safe for tests" label. Is this intentional, if so is there something else I should do on my part? Thanks

alielbashir:
@deannagarcia any updates regarding this PR?

PR #11438, opened 2023-01-03
Support for LoongArch64

Panxuefeng-loongson:


jorgbrown:
We'd prefer not to explicitly list every possible architecture.

I tested using the loongarch64 gcc compiler; see <https://godbolt.org/z/9Er7Yac6o> and it appears that we already do the right thing, with regards to detecting whether it's 64-bit...

What problem are you seeing right now?

ps Are you aware of https://github.com/protocolbuffers/protobuf/pull/11442 ?

Panxuefeng-loongson:
> We'd prefer not to explicitly list every possible architecture.
> 
> I tested using the loongarch64 gcc compiler; see https://godbolt.org/z/9Er7Yac6o and it appears that we already do the right thing, with regards to detecting whether it's 64-bit...
> 
> What problem are you seeing right now?
> 
> ps Are you aware of #11442 ?

Sorry, It's my fault. We use the v3.14.0 version of protobuf in our project, I have made some changes in this version, and want to sync these changes to the latest code. I didn't realize my colleague had already done the work, so this pr is redundant. I will be closing this PR．

PR #11442, opened 2023-01-03
Add support for loongarch64

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

### DIFF
--- a/src/google/protobuf/stubs/platform_macros.h
+++ b/src/google/protobuf/stubs/platform_macros.h
@@ -83,6 +83,9 @@
 #elif defined(__PPC__)
 #define GOOGLE_PROTOBUF_ARCH_PPC 1
 #define GOOGLE_PROTOBUF_ARCH_32_BIT 1
+#elif defined(__loongarch64)
+#define GOOGLE_PROTOBUF_ARCH_LOONGARCH64 1
+#define GOOGLE_PROTOBUF_ARCH_64_BIT 1
 #elif defined(__GNUC__)
 # if (((__GNUC__ == 4) && (__GNUC_MINOR__ >= 7)) || (__GNUC__ > 4))
 // We fallback to the generic Clang/GCC >= 4.7 implementation in atomicops.h


jorgbrown:
We'd prefer not to explicitly list every possible architecture.

I tested using the loongarch64 gcc compiler; see <https://godbolt.org/z/9Er7Yac6o> and it appears that we already do the right thing, with regards to detecting whether it's 64-bit...

What problem are you seeing right now?

ps Are you aware of https://github.com/protocolbuffers/protobuf/pull/11438 ?

zhaixiaojuan:
> We'd prefer not to explicitly list every possible architecture.
> 
> I tested using the loongarch64 gcc compiler; see https://godbolt.org/z/9Er7Yac6o and it appears that we already do the right thing, with regards to detecting whether it's 64-bit...
> 
> What problem are you seeing right now?
> 
> ps Are you aware of #11438 ?

sorry, I will close this PR.

zhaixiaojuan:
Hello, I would like to ask whether the release under the protobuf project can add the binary package of loongarch64. Or what efforts do we need to make this happen?

Panxuefeng-loongson:
> Hello, I would like to ask whether the release under the protobuf project can add the binary package of loongarch64. Or what efforts do we need to make this happen?

It looks like protobuf is important to the LoongArch architecture, but the community may be short of building machines. docker + qume might be a viable option, but it will increase maintenance overheads. I wonder what your opinion is on this. @jorgbrown

zhaixiaojuan:
Is the following approach possible?
We provide build machines locally, set up CI build process, and perform remote builds.

PR #11449, opened 2023-01-03
Use forward slash instead of backslash in nuspec file

deannagarcia:
We are going to pack this on a linux machine now, so we need to use forward slashes.


perezd (approved):

jskeet (approved):
I'm slightly surprised that the tooling doesn't already handle this (as it does for things like project references) but it won't do any harm to use forward slashes.

PR #11468, opened 2023-01-05
Add ruby release targets

deannagarcia:
Add bazel targets to create ruby release artifacts.

Should be run with:
```
bazel run ruby:release
bazel run ruby:jruby_release 
```


maven_install.json
```diff
@@ -651,6 +651,18 @@
                 "sha256": "66fdef91e9739348df7a096aa384a5685f4e875584cce89386a7a47251c4d8e9",
                 "url": "https://repo1.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar"
             },
+            {
+                "coord": "org.jruby:jruby-complete:9.2.20.1",
```
perezd, line 655: does "JRuby support version" become something we have to have an established policy on now as well? cc @devjgm

perezd (approved):
LGTM

maven_install.json
devjgm: > does "JRuby support version" become something we have to have an established policy on now as well? cc @devjgm

Most likely, yes.

mkruskal-google:
> bazel run ruby:release --define=ruby_platform=c
> bazel run ruby:jruby_release --define=ruby_platform=java

These defines aren't necessary anymore, and don't actually do anything.  The new pattern is to use `target_compatible_with` (e.g. https://github.com/protocolbuffers/protobuf/blob/ade256e153ea7bfff85987a7f1bf2e9c9632cb10/ruby/BUILD.bazel#L54) and just allow the toolchain resolution to pick the appropriate target

protobuf_deps.bzl
```diff
@@ -12,6 +12,7 @@ PROTOBUF_MAVEN_ARTIFACTS = [
     "com.google.guava:guava-testlib:31.1-jre",
     "com.google.truth:truth:1.1.2",
     "junit:junit:4.13.2",
+    "org.jruby:jruby-complete:9.2.20.1",
```
mkruskal-google, line 15: This shouldn't be necessary.  This forces all our downstream users to depend on a pinned version of jruby.  rules_ruby can be used to avoid this and always use a consistent version

ruby/BUILD.bazel
```diff
@@ -19,6 +21,46 @@ ruby_library(
     ],
 )
 
+java_export(
+    name = "protobuf_java_release",
+    maven_coordinates = "com.google.protobuf:protobuf-jruby:%s" % PROTOBUF_RUBY_VERSION,
+    srcs = glob(["src/main/java/com/google/protobuf/jruby/*"]),
+    runtime_deps = [
+        "@maven//:org_jruby_jruby_complete",
```
mkruskal-google, line 29: You can change this to "@rules_ruby//ruby/runtime:jars" to select the system jruby jars

ruby/BUILD.bazel
```diff
@@ -19,6 +21,46 @@ ruby_library(
...
+      "//:well_known_ruby_protos",
+    ],
+    srcs = [
+      "build_jruby_release.sh",
+    ],
+)
+
+# Run with the flag: --define=ruby_platform=java
+sh_binary(
+    name = "release",
```
mkruskal-google, line 53: What's the point of this target?  It looks like it should be broken, as it's missing dependencies on a lot of files it uses.  Also, this looks like it's meant for jruby, where is the cruby version?

ruby/BUILD.bazel
```diff
@@ -19,6 +21,46 @@ ruby_library(
     ],
 )
 
+java_export(
+    name = "protobuf_java_release",
```
mkruskal-google, line 25: We already have a really similar target to this in //ruby/lib/google:protobuf_lib.  Do we need this additional jar?

ruby/build_release.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# This file should be executed with ruby
+
+set -ex
+
+WORKING_DIR=$(dirname $(readlink $BASH_SOURCE))
```
mkruskal-google, line 6: Does this dump the results in the source tree or in the sandbox?  If it's the former this seems very un-bazel-like

ruby/build_release.sh
```diff
@@ -0,0 +1,24 @@
...
+
+rvm use ruby-3.0
```
mkruskal-google, line 8: This requires rvm and ignores the ruby version selected by the caller

ruby/BUILD.bazel
deannagarcia: The tags got switched, this is for cruby and I left it as the default because that's how the other targets are named.

When I run it it builds a gem that installs, why do you think it should be broken?

ruby/BUILD.bazel
deannagarcia: That's a ruby library, but I need to use this rule to correctly package the maven jar including creating some template files.

ruby/build_release.sh
deannagarcia: It dumps in the source tree, but I think that's just how `sh_binary` works. I don't think it's a problem for release targets since we will always start with a clean tree and dump everything else.

ruby/build_release.sh
deannagarcia: I thought you told me to use this command? 

What should I do instead here to make sure we are using ruby and not jruby?

deannagarcia:
> > bazel run ruby:release --define=ruby_platform=c
> > bazel run ruby:jruby_release --define=ruby_platform=java
> 
> These defines aren't necessary anymore, and don't actually do anything. The new pattern is to use `target_compatible_with` (e.g.
> 
> https://github.com/protocolbuffers/protobuf/blob/ade256e153ea7bfff85987a7f1bf2e9c9632cb10/ruby/BUILD.bazel#L54
> 
> ) and just allow the toolchain resolution to pick the appropriate target

I implemented that but now when I test I get:

```
Dependency chain:
    //ruby:jruby_release (06aa77)   <-- target platform (@local_config_platform//:host) didn't satisfy constraint @platforms//:incompatible
```

How can I use this and test locally?

ruby/BUILD.bazel
deannagarcia: I see what you're saying, I don't have a formal dependency on the source files since they are in this directory I just paste the things that need to be generated in the folder.

I'm looking at how to get the bazel data files and that's more inconvenient than just using the directory structure that already exists here.

Do you have any better ideas on how to do this?

mkruskal-google: Are you using files outside of the sandbox then?  Ideally you would add them as formal dependencies in the BUILD.bazel file and then use the relative paths to get the sandboxed versions.  inline_sh_binary is a pretty useful tool for getting bazel variable expansion in sh files

ruby/BUILD.bazel
mkruskal-google: `protobuf_lib` is a ruby library, but under jruby it has a data dependency on `//ruby/src/main/java:protobuf_java` which is built using the resolved jruby's jar files.  There is also a jar created as an intermediary at `//ruby/lib/google/protobuf:protobuf_java.jar`

ruby/build_release.sh
mkruskal-google: That's not how `sh_binary` is *supposed* to work IIUC.  This is non-hermetic and could cause issues for us during development/testing.  It would be much better to keep everything inside the Bazel sandbox, as it forces explicit dependency declaration and keeps the source tree isolated.  I could also see this playing badly with Bazel caching, since it's unaware of some of our dependencies

ruby/build_release.sh
mkruskal-google: I said you should use this command from the release script, and specifically not inside a Bazel rule.  Ruby is now bazelified, so Bazel rules should be using whatever version was selected by the user/toolchain resolution. `@rules_ruby//ruby/runtime:interpreter` should resolve to the appropriate binary.

rvm can still be used *outside* Bazel to select a pinned version in our release script.  I was imagining our release script would look something like:
```
rvm use ruby-3.0
bazel build //ruby:release
cp bazel-bin/ruby/<generated release> <dest>
```

mkruskal-google: Note: that assumes our WORKSPACE file selects the system ruby version, similar to our python setup.  After some planned enhancements we would also be able to pass the ruby version as a bazel flag

mkruskal-google:
Our current configuration is to use the system ruby version.  So if you're using c-ruby and you run a bazel target only compatible with jruby, you'll see those constraint errors.  You'll need to switch to jruby with rvm before running the target.

Alternatively, I came up with a strategy that I liked for this kind of thing.  If you have cruby and jruby targets that build the same "kind" of thing, you can add a third target with a conditional dependency on both.  I did this with the protobuf runtime, where `//lib/google/protobuf:protobuf_lib` will always compile, but sometimes refer to the jruby jar and sometimes to the cruby library.

.github/workflows/ruby_install.yml
```diff
@@ -0,0 +1,50 @@
...
+  workflow_dispatch:
+
+jobs:
+  test_ruby_gems:
+    name: Test ruby gems
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.6, 2.7, 3.0, 3.1, jruby-9.2, jruby-9.3]
```
opti, line 21:
```suggestion
        ruby: [2.6, 2.7, 3.0, 3.1, 3.2, jruby-9.2, jruby-9.3]
```

ruby/BUILD.bazel
deannagarcia: Done.

ruby/BUILD.bazel
deannagarcia: I fixed this to depend on all files we need and run in bazel-bin instead of the github repo itself.

ruby/build_release.sh
deannagarcia: As mentioned above, results are no longer dumped in the sandbox.

ruby/build_release.sh
deannagarcia: I removed the lines setting ruby/jruby version and am just using `target_compatible_with` to force the correct version.

ruby/build_jruby_release.sh
```diff
@@ -0,0 +1,63 @@
...
+    source "$(grep -m1 "^bazel_tools/tools/bash/runfiles/runfiles.bash " \
+              "$RUNFILES_MANIFEST_FILE" | cut -d ' ' -f 2-)"
+else
+echo >&2 "ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash"
+exit 1
+fi
+# --- end runfiles.bash initialization ---
+
+# Make a temporary directory and move to it to do all packaging work
+mkdir -p tmp
```
mkruskal-google, line 30: Since we're in the sandbox now do we need a tmp directory?

ruby/build_jruby_release.sh
```diff
@@ -0,0 +1,63 @@
...
+cd tmp
+
+# Move all generated files to lib/google/protobuf
+mkdir -p lib/google/protobuf
+cp "$(rlocation com_google_protobuf/src/google/protobuf/any_pb.rb)" lib/google/protobuf
```
mkruskal-google, line 35: You can use `dirname` to copy these with a wildcard

mkruskal-google, line 35: `ln` or `mv` would be slightly more efficient here

ruby/build_release.sh
```diff
@@ -0,0 +1,88 @@
...
+else
+echo >&2 "ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash"
+exit 1
+fi
+# --- end runfiles.bash initialization ---
+
+# rvm use ruby-3.0
+
+# Make a temporary directory and move to it to do all packaging work
+mkdir -p tmp
```
mkruskal-google, line 32: Same comments here

ruby/build_jruby_release.sh
```diff
@@ -0,0 +1,63 @@
...
+if [[ -f "${RUNFILES_DIR:-/dev/null}/bazel_tools/tools/bash/runfiles/runfiles.bash" ]]; then
+  source "${RUNFILES_DIR}/bazel_tools/tools/bash/runfiles/runfiles.bash"
+elif [[ -f "${RUNFILES_MANIFEST_FILE:-/dev/null}" ]]; then
+    source "$(grep -m1 "^bazel_tools/tools/bash/runfiles/runfiles.bash " \
+              "$RUNFILES_MANIFEST_FILE" | cut -d ' ' -f 2-)"
+else
+echo >&2 "ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash"
+exit 1
+fi
+# --- end runfiles.bash initialization ---
```
mkruskal-google, line 27: optional: This boilerplate is why I created inline_sh_binary.  I don't have a strong preference here, but if you can get this smaller it might be nice to define it inline in the BUILD.bazel file

.github/workflows/ruby_install.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          $HOME/bin/bazel run ruby:release
+          gem install bazel-bin/ruby/release.runfiles/com_google_protobuf/tmp/google-protobuf-*
+        if: ${{ !contains(matrix.ruby, 'jruby') }}
+      - name: Build and install jruby gem
+        run: |
+          $HOME/bin/bazel run ruby:jruby_release
+          gem install bazel-bin/ruby/jruby_release.runfiles/com_google_protobuf/tmp/google-protobuf-*
+        if: ${{ contains(matrix.ruby, 'jruby') }}
+      - name: Test installation
```
mkruskal-google, line 52: Maybe we should add a sanity check to the beginning to make sure these tests *fail* before installation?

ruby/build_jruby_release.sh
deannagarcia: I needed to make a new directory to start with a clean directory since this ends up being run in bazel-bin's runfile directory where a copy of all of the dependency files are.

ruby/build_jruby_release.sh
deannagarcia: I went back and forth on this. rlocation doesn't support wildcards, but I think you are suggesting that I get the dirname and then use wildcards which could be possible.

I ended up manually listing them as more of a precaution. Everywhere else the files are wildcarded in which I think could make it very easy to lose a file. This will fail in that case, which I think ends up being a good check. On the other hand, a new file being added could cause this to fail silently. What do you think about that tradeoff?

ruby/build_jruby_release.sh
deannagarcia: This code seems like a pretty common pattern in many bazel files using `sh_binary` and I don't think we can cut it down much and putting it in the BUILD.bazel file as is seems clunky.

.github/workflows/ruby_install.yml
deannagarcia: Do you know how to run in GHA expecting a failure? I can't find a good way of doing that (but can confirm that I have seen this fail when the gems were not correct).

mkruskal-google: ! bazel test <target> will invert the result

ruby/build_jruby_release.sh
mkruskal-google: This gets run in bazel-bin/ruby/jruby_release.runfiles/com_google_protobuf/, which should *only* contain your dependencies.  Why do we need to copy all of them into a fresh folder?  This seems redundant, we could just run gem build from this directory

ruby/build_jruby_release.sh
mkruskal-google: If we were missing a file I would expect the gem build/test to fail in our CI.  This seems a bit redundant to me, and not worth the maintenance burden.

One alternative that might work nicely is to add an sh_test rule that actually runs the tests using the built gem.  That way we can actually run our tests locally and not have to wait for the CI to be run over a PR

PR #11501, opened 2023-01-09
Dynamic messaging

himkak:
Touched files to show correct user


jskeet:
I'll have a look when I get a chance, although just at a glance it's missing copyright notices, there's a lot of non-useful vertical whitespace, and you seem to be trying to change the version number (which is something the protobuf team does, and shouldn't be in this PR).

PR #11571, opened 2023-01-17
python: make ctypes dependency optional

robbertvanginkel:
This allows the protobuf python library to work on cpython 3.11's new wasm target [1], which does not have ctypes.

This reverts back to the way TruncateToFourByteFloat worked before 75eff10d81233d3d08ce5ded215e526c1fcace88.

  [1] https://docs.python.org/3/whatsnew/3.11.html#build-changes:~:text=CPython%20now%20has%20PEP%2011%20Tier%203%20support%20for%20cross%20compiling%20to%20the%20WebAssembly%20platforms

### DIFF
--- a/python/google/protobuf/internal/type_checkers.py
+++ b/python/google/protobuf/internal/type_checkers.py
@@ -45,11 +45,7 @@ TYPE_TO_DESERIALIZE_METHOD: A dictionary with field types and deserialization
 
 __author__ = 'robinson@google.com (Will Robinson)'
 
-try:
-  import ctypes
-except Exception:  # pylint: disable=broad-except
-  ctypes = None
-  import struct
+import struct
 import numbers
 
 from google.protobuf.internal import decoder
@@ -61,10 +57,7 @@ _FieldDescriptor = descriptor.FieldDescriptor
 
 
 def TruncateToFourByteFloat(original):
-  if ctypes:
-    return ctypes.c_float(original).value
-  else:
-    return struct.unpack('<f', struct.pack('<f', original))[0]
+  return struct.unpack('<f', struct.pack('<f', original))[0]
 
 
 def ToShortestFloat(original):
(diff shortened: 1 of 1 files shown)


python/google/protobuf/internal/type_checkers.py
```diff
@@ -57,7 +61,10 @@
 
 
 def TruncateToFourByteFloat(original):
-  return ctypes.c_float(original).value
+  if ctypes:
+    return ctypes.c_float(original).value
+  else:
+    return struct.unpack('<f', struct.pack('<f', original))[0]
```
haberman, line 67: If this is correct, we might as well use `struct` all of the time.

Otherwise we won't get any CI coverage for this code path.

akshayjshah:
@fowles FYI, this is one of the WASM-related PRs that we discussed last week. @robbertvanginkel's going to update the PR to address CR feedback, of course :)

python/google/protobuf/internal/type_checkers.py
robbertvanginkel: From my understanding they are the same:

```
Python 3.10.9 (main, Dec 15 2022, 10:44:50) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import ctypes, struct
>>> original = 0.12345678901234567890123456789
>>> ctypes.c_float(original).value
0.12345679104328156
>>> struct.unpack('<f', struct.pack('<f', original))[0]
0.12345679104328156
```

I can switch this to always use struct. When putting up the PR I reverted this to what it was in the proto project in the past:

https://github.com/protocolbuffers/protobuf/blob/10d0faef6de038960ec825c003c85581800fba1c/python/google/protobuf/internal/type_checkers.py#L68-L72

It was changed to ctypes only in https://github.com/protocolbuffers/protobuf/commit/75eff10d81233d3d08ce5ded215e526c1fcace88 so I figured there may have been some reasoning as to why, but that's unfortunately not visible to OSS due to the big sync PR.

robbertvanginkel: Done.

haberman: > but that's unfortunately not visible to OSS due to the big sync PR.

I looked back and it was part of a CL to remove Python 2.x compatibility code.

If `ctypes` is available, then the `ctypes` code is simpler. But if it's not always available, we might as well use `struct` all of the time.

Issue #11578, opened 2023-01-18
google-protobuf-3.21.12-x86_64-linux is incompatible with Ruby 3.2.0


smaboshe:
**What version of protobuf and what language are you using?**
Version: `google-protobuf-3.21.12-x86_64-linux`
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**
Linux (`ubuntu-latest` in a GitHub Action)

**What runtime / compiler are you using (e.g., python version or gcc version)**
I do not know at the moment

**What did you do?**
1. Run a GitHub action that includes installation of the latest `google-protobuf` gem
2. See error

**What did you expect to see**
```
Fetching google-protobuf 3.21.12
Installing google-protobuf 3.21.12 with native extensions
```

**What did you see instead?**
```
google-protobuf-3.21.12-x86_64-linux requires ruby version < 3.2.dev, >= 2.5,
which is incompatible with the current version, 3.2.0
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**

casperisfine:
You can fallback to the source version by running `bundle install` with Ruby 3.2. Bundler will notice this incompatibility and remove the precompiled version from the Gemfile.lock.

If you use `dependabot`, you'll need to add `ruby ">= 3.2"` in your `Gemfile` so that it knows to not include incompatible gems either.

casperisfine:
A new ruby release would be nice though.

mrmason:
Think this is a duplicate of this - https://github.com/protocolbuffers/protobuf/issues/11538

smaboshe:
> You can fallback to the source version by running `bundle install` with Ruby 3.2. Bundler will notice this incompatibility and remove the precompiled version from the Gemfile.lock.
> 
> If you use `dependabot`, you'll need to add `ruby ">= 3.2"` in your `Gemfile` so that it knows to not include incompatible gems either.

The fallback suggestion fixes our problem. Happy for you to close this issue.


> Think this is a duplicate of this - #11538

Yes, I believe it is. I did not notice this when searching the issue history.


Thank you all for the help!

zhangskz:
Thanks for the community contribution here. Closing this thread as resolved. Note that as pointed out in #11538, this is mitigated in future releases which will only release source gems.

johnnyshields:
PR raised for supporting fat binaries on Ruby 3.2: https://github.com/protocolbuffers/protobuf/pull/11592

Issue #11538, opened 2023-01-12
Support for Ruby 3.2.0 in prebuilt binaries


frederikspang:
**What version of protobuf and what language are you using?**
Version: main
Language: Ruby

**What operating system (Linux, Windows, ...) and version?**
All

**What runtime / compiler are you using (e.g., python version or gcc version)**
Any

**What did you do?**
Install latest rubygem.

**What did you expect to see**
Prebuilt binary for Ruby 3.2.0 to be installed

**What did you see instead?**
It is installing `with native extensions` (ie. compiling)

**Anything else we should know about your project / environment**
Reopening #11407 with proper template, and because the issue has been closed.

Virtually, the rubygem needs to explicitly support and build for Ruby 3.2.0 as well as the others.
Currently the native prebuilt gems are build for `>= 2.5, < 3.2.DEV`
 (see https://rubygems.org/gems/google-protobuf/versions/3.21.12-x86_64-linux for instance)

We should add support for 3.2.0 as it has been released in December.

deannagarcia:
As stated in the bug you linked to, the workaround for this is to use the ruby source gem instead of the compiled binary. We are working on 22.0 currently, which will only release ruby source gems making them default compatible with ruby 3.2.

Unfortunately I don't think we will be updating the ruby gem binary for 21.x given the impending release.

pond:
@deannagarcia and how long until that is released?

Bundler does not like keeping `ruby` under `PLATFORMS` these days and tends to want to force in specific architectures; and some of the components in our gem stack, including `grpc`, have _extremely_ long build times when built from source - which costs us a fortune in CI time. Compilation can also be fragile. So - we don't want to keep the hack we have in our CI `Dockerfile` forcing Bundler into a `ruby` platform any longer than we must.

deannagarcia:
We are currently aiming to release next week.

When we release only source gems will that solve your problem or will you still have the same issue?

pond:
@deannagarcia Oh cool, that's not long to wait at all, thanks.

Source-only gems should be OK for protobuf since it builds fairly quickly (GRPC is around 10 minutes or more!). There's always a risk of failure when building native components but it sounds like you don't want to have to manage a wide range of pre-built binaries or have to keep the builds on a Ruby treadmill and that definitely makes sense.

mrmason:
[...]

PR #11592, opened 2023-01-19
Add prebuilt binaries for Ruby 3.2

johnnyshields:
This PR adds pre-compiled Ruby binaries for Ruby 3.2. Note that such binaries are already available on other Ruby versions.

Fixes issues:
- https://github.com/protocolbuffers/protobuf/issues/11578
- https://github.com/protocolbuffers/protobuf/issues/11538

(Those issues should not have been closed without a fix...)

The protobufs Ruby maintainers should get in the habit of updating this every year when there is a new Ruby release. Please set a calendar reminder to check in early Jan with the maintainers of `rake-compiler` gem; when they release cross-compilation support for the new Ruby version it's an easy change to add it to protobufs (can basically copy this PR.)

### DIFF
--- a/ruby/Rakefile
+++ b/ruby/Rakefile
@@ -133,7 +133,7 @@ else
     ['x86-mingw32', 'x64-mingw32', 'x64-mingw-ucrt', 'x86_64-linux', 'x86-linux'].each do |plat|
       RakeCompilerDock.sh <<-"EOT", platform: plat
         bundle && \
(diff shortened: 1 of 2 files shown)


johnnyshields:
@apolcyn @zhangskz

johnnyshields:
@zhangskz so the protobufs team is making the decision to abandon the pre-built binaries?! That seems like a big headache for the Ruby community given that they have historically been supported. Binaries help save everyone’s CI resources when bundling.

As you can see this PR is quite trivial and could easily be added to 22.0 or 23.0. If the plan is to continue to support binaries, please leave this PR open.

CC: @apolcyn

PR #11601, opened 2023-01-19
Update Google.Protobuf.Tools.nuspec

deannagarcia:
Find all source files inside the csharp directory. We have to prepopulate this directory anyway, so we can put the .proto files directly into the csharp directory.

The current setup can find the files in linux, but for some reason it isn't copying them correctly unless the files are in the csharp folder

### DIFF
--- a/csharp/Google.Protobuf.Tools.nuspec
+++ b/csharp/Google.Protobuf.Tools.nuspec
@@ -21,17 +21,17 @@
     <file src="protoc/linux_x86/protoc" target="tools/linux_x86/protoc"/>
     <file src="protoc/linux_x64/protoc" target="tools/linux_x64/protoc"/>
     <file src="protoc/macosx_x64/protoc" target="tools/macosx_x64/protoc"/>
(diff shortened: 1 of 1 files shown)


jskeet:
> We have to prepopulate this directory anyway

Is that in some other bit of the release script? I'm not aware of it at the moment.

deannagarcia:
> > We have to prepopulate this directory anyway
> 
> Is that in some other bit of the release script? I'm not aware of it at the moment.

Yep, this is part of our internal release process and is the migration of [this file](https://github.com/protocolbuffers/protobuf/blob/main/kokoro/release/collect_all_artifacts.sh).

jskeet (approved):
Okay - if you're happy this is right, then I'm fine with the change.

zhangskz:
I believe this was submitted already in https://github.com/protocolbuffers/protobuf/commit/196193d65cb111913f1615853b44ff60aa8a533d but not auto-closed since it wasn't imported via copybara.

PR #11608, opened 2023-01-20
Migrate utf8_range dependency to a git subtree

mkruskal-google:
This will make the CMake workflow easier, where users won't need to either install utf8_range or initialize our submodules.  Since this is a relatively small library that's owned by us, we can just bundle it with protobuf as a subtree


zhangskz (approved):
I think this should land internally first or maybe copybara needs to be fixed after-the-fact since utf8_range.cmake isn't github only. Also just sanity checking that this doesn't affect how we use the utf8_range submodule for php release (https://github.com/protocolbuffers/protobuf/blob/main/php/BUILD.bazel#L34)? (pretty sure we have test coverage for this though)

mkruskal-google:
This shouldn't affect bazel builds at all, since we depend on utf8_range through Bazel.  I plan on importing this with copybara, it *should* work?  I might have to force it in tho :P

PR #11622, opened 2023-01-21
Upgrade to Abseil LTS 20230117

mkruskal-google:


zhangskz (approved):
Import as CL? (you'll probably have the same copybara issue as with utf8_range though)

mkruskal-google:
Yep exactly, it's got to be bidirectional

PR #11624, opened 2023-01-21
[cpp] Add initializer_list insertion support for repeated field.

pateldeev:
Add initializer list insertion support for repeated fields. Currently, something like the following compiles but does not [do what you expect](https://godbolt.org/z/zE7K3Yaab)! This is obviously wrong.
```
  RepeatedPtrField<std::string> field;
  field.Add({"Hello", "World"});
```


mkruskal-google:
Hey Deev, would you mind rebasing this PR to main?  We recently migrated to GHA and you'll be able to see the details of the test failures

Issue #9561, opened 2022-03-01
Ruby oneof fields should have `has_foo?`, even for proto3 files


haberman:
Currently Ruby suppresses `has_foo?` methods for proto3: https://github.com/protocolbuffers/protobuf/blob/c578f59fc8db97e2482208e02743ab1306ccd09e/ruby/ext/google/protobuf_c/message.c#L246-L254

As of 2020-09-15, it was decided that all oneof fields should have hazzers, even for proto3.

We should update Ruby to support `has_foo?` for oneof fields in Ruby.

(Note: in this case `foo` is the name of a field in a oneof, `foo` is not a oneof itself).

haberman:
All we need to do is remove this `if` statement, and add a unit test to verify: https://github.com/protocolbuffers/protobuf/blob/7309e8882cb81a92cee4b0be708a61c8f4707224/ruby/ext/google/protobuf_c/message.c#L245-L253

PR #11655, opened 2023-01-25
For Ruby oneof fields, generate hazzers for members

shaldengeki:
When generating Ruby clients for proto3 messages that have a oneof, we generate a hazzer for members of the oneof, not just a hazzer for the oneof itself.

In other words, for a proto like this:

```
syntax = "proto3";
message Foo {
  oneof bar {
    string baz = 1;
  }
}
```

The generated `Foo` will now have a method called `has_baz?`, in addition to the (pre-existing) method `has_bar?`.

I updated the unit tests, and verified that all the tests under `//ruby/...` pass.

Fixes https://github.com/protocolbuffers/protobuf/issues/9561.

### DIFF
--- a/ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java
+++ b/ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java
@@ -455,11 +455,7 @@ public class RubyMessage extends RubyObject {
 
         fieldDescriptor = descriptor.findFieldByName(methodName);
 
-        if (fieldDescriptor != null
-            && (!proto3
-                || fieldDescriptor.getContainingOneof() == null
-                || fieldDescriptor.getContainingOneof().isSynthetic())
-            && fieldDescriptor.hasPresence()) {
+        if (fieldDescriptor != null && fieldDescriptor.hasPresence()) {
           return fields.containsKey(fieldDescriptor) ? runtime.getTrue() : runtime.getFalse();
         }
 
--- a/ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java
+++ b/ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java
@@ -306,11 +306,7 @@ public class RubyMessage extends RubyObject {
     if (methodName.startsWith(HAS_PREFIX) && methodName.endsWith(QUESTION_MARK)) {
       String strippedMethodName = methodName.substring(4, methodName.length() - 1);
       FieldDescriptor fieldDescriptor = descriptor.findFieldByName(strippedMethodName);
-      if (fieldDescriptor != null
-          && (!proto3
-              || fieldDescriptor.getContainingOneof() == null
-              || fieldDescriptor.getContainingOneof().isSynthetic())
-          && fieldDescriptor.hasPresence()) {
+      if (fieldDescriptor != null && fieldDescriptor.hasPresence()) {
         return context.runtime.getTrue();
       }
       oneofDescriptor =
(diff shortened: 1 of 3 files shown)


JasonLunn:
Thanks for your contribution, @shaldengeki. Can you update this PR to include the changes to the JRuby implementation as well?

ruby/src/main/java/com/google/protobuf/jruby/RubyMessage.java
```diff
@@ -306,11 +306,7 @@ public IRubyObject respondTo(ThreadContext context, IRubyObject[] args) {
     if (methodName.startsWith(HAS_PREFIX) && methodName.endsWith(QUESTION_MARK)) {
       String strippedMethodName = methodName.substring(4, methodName.length() - 1);
       FieldDescriptor fieldDescriptor = descriptor.findFieldByName(strippedMethodName);
-      if (fieldDescriptor != null
-          && (!proto3
-              || fieldDescriptor.getContainingOneof() == null
-              || fieldDescriptor.getContainingOneof().isSynthetic())
-          && fieldDescriptor.hasPresence()) {
+      if (fieldDescriptor != null && fieldDescriptor.hasPresence()) {
```
shaldengeki, line 309: @JasonLunn Can you take a look and let me know if I'm on the right track here? I _think_ I've replicated the same logic, but:
- I'm not totally sure if this is the right place to be modifying,
- I could use a sanity-check on the logic here.

JasonLunn: Looks about right to me on its surface. Rerunning tests now to confirm.

shaldengeki:
Hmm. Looks like the kokoro JRuby builds failed. I don't appear to have access to the test logs, either. Let me know if there's something I should fix here -- I looked at `kokoro/linux/bazel.sh` and it _looks_ like I wouldn't be able to replicate this locally.

JasonLunn:
> Hmm. Looks like the kokoro JRuby builds failed. I don't appear to have access to the test logs, either. Let me know if there's something I should fix here -- I looked at `kokoro/linux/bazel.sh` and it _looks_ like I wouldn't be able to replicate this locally.

```
Error: test_has_field(BasicTest::MessageContainerTest): NoMethodError: undefined method `has_a?' for <BasicTest::OneofMessage: >:BasicTest::OneofMessage
org/jruby/RubyBasicObject.java:1715:in `method_missing'
com/google/protobuf/jruby/RubyMessage.java:540:in `method_missing'
/workspace/_build/out/sandbox/processwrapper-sandbox/1188/execroot/com_google_protobuf/bazel-out/k8-fastbuild/bin/ruby/tests/basic.runfiles/com_google_protobuf/ruby/tests/basic.rb:122:in `test_has_field'
     119:
     120:       m = OneofMessage.new
     121:       assert !m.has_my_oneof?
  => 122:       assert !m.has_a?
     123:       m.a = "foo"
     124:       assert m.has_my_oneof?
     125:       assert m.has_a?
```

I think the issue is that the JRuby change you made is only in the context of the implementation of `responds_to?`. There is a very similar change, in the same file, to the implementation of `method_missing` starting at ~ line 462.

shaldengeki:
Ooh. Looks like we have one remaining failing build. I poked more at running the kokoro builds locally, but this one's a MacOS build and I unfortunately don't have a device I can test on.

@JasonLunn Thanks for your help so far; when you have some time, take a look & let me know what's breaking.

JasonLunn:
LGTM, @haberman - WDYT?

haberman (approved):

haberman:
LGTM, thanks for your contribution. This has been on our TODO list for a while.

shaldengeki:
woot - thank y'all for the help here! 🎉

PR #11668, opened 2023-01-25
Update protobuf upb dep

zhangskz:


zhangskz:
Will import once required CI tests pass

mkruskal-google (approved):

Issue #5721, opened 2019-02-13
Insertion points and their future


yuri-sergiichuk:
I've noticed in several other issues (e.g. #3292, #3291, #2684) Protobuf contributors are standing that insertion points should not be used, does this actually mean that in some time custom Protobuf plugins won't be supported at all? 
It's that I'm looking at customizing Protobuf Java output with some extra stuff and e.g. I've noticed an insertion point `@@protoc_insertion_point(message_implements:NAME)` that is actually present in the generated sources but is not documented at the [website](https://developers.google.com/protocol-buffers/docs/reference/java-generated#plugins). 

Could you please provide some statement about the insertion points support and usage at all and about `message_implements` insertion point in particular, is it safe to use it?

BSBandme:
I'm not sure about the plan for that but I think we're trying to deprecate insertion point and we may break you in the future.

yuri-sergiichuk:
@BSBandme thank you for your attention to the question. Is there any change to announce any actual "plans" or "dates" if the functionality is gonna be deprecated?
Any plans to discuss this with the community?

marcoferrer:
I’m also interested in the discussion around this feature. I currently author a library that tags the builder interfaces with a Kotlin specific annotation. It’s purpose is to allow composition of proto message in a type safe dsl like manner, but doesn’t modify the interface in any way. It’s purely meant to provide hints to the ide and compile time safety. 

Is this not an acceptable usage of the insertions api?

haon4:
@@protoc_insertion_point(message_implements:NAME) is inserted at the following point:
```
public final class Foo extends
    com.google.protobuf.GeneratedMessageV3 implements
    // @@protoc_insertion_point(message_implements:package.name.Foo)
    FooOrBuilder {
...
}
```

We have found that some insertion points are too fine grained and allow developers to get into the way of our implementation details, while others are fine to have. e.g., we have a number of insertion points in C++ for serialize/parse, which isn't something that we would like developers to change. We are reevaluating the values of some of these insertion points to see if they should be removed.

Our stand is not to completely kill _all_ insertion points, but to eliminate certain insertion points that are problematic and likely to cause more harms. Most other insertion points at class scope / builder scope are acceptable to have.

yuri-sergiichuk:
@haon4 we are now using `@@protoc_insertion_point(message_implements:NAME)` and `@@protoc_insertion_point(class_scope:NAME)`.

[...]

PR #11669, opened 2023-01-25
Add post_imports protoc_insertion_point for Python

fahhem:
This is part of an incoming series of PR's intended to support what I call "unified_protos" (unifying the C++ runtime with the Python runtime for protobufs). My approach will not be in the way that the code currently seems to require (a statically linked python executable with all C++ protos and C-extensions linked into it), but through building a shared library for every C++ proto (via `cc_shared_library`) and loading them (in the post_imports hook) in the generated Python code and using an aspect to make them dynamic_deps for C extensions. This allows Python code and C/C++ code to pass protobufs back-and-forth without serializing/deserializing them.

### DIFF
--- a/src/google/protobuf/compiler/python/generator.cc
+++ b/src/google/protobuf/compiler/python/generator.cc
@@ -433,6 +433,7 @@ void Generator::PrintImports() const {
     printer_->Print("from $module$ import *\n", "module", module_name);
   }
   printer_->Print("\n");
+  printer_->Print("# @@protoc_insertion_point(post_imports)\n\n");
 }
 
 // Prints the single file descriptor for this file.


zhangskz:
Thanks for the contribution! Unfortunately, protoc insertion points are highly deprecated, so we cannot move forward with this approach. See https://github.com/protocolbuffers/protobuf/issues/5721 for some context.
 
If you would like to explore additional paths forward without protoc insertion points, feel free to open an issue for discussion with the protobuf team.

See [Contributing to Protocol Buffers](https://github.com/protocolbuffers/protobuf/blob/main/CONTRIBUTING.md#pull-request-guidelines): 
"For speculative changes, consider opening an issue and discussing it first. If you are suggesting a behavioral or API change, make sure you get explicit support from a protobuf team member before sending us the pull request."

fahhem:
Strange, the only team member posting in that issue said:

> Our stand is not to completely kill all insertion points, but to eliminate certain insertion points that are problematic and likely to cause more harms. Most other insertion points at class scope / builder scope are acceptable to have.

And a post_imports hook seems not problematic at all, and this hook was used internally to great effect. I'll open the issue for the overall idea of unified_protos though.

Issue #11672, opened 2023-01-26
Visual Studio 2022: protobuf\src\google\protobuf\arena.cc(457,51): error C2127: 'thread_cache_': illegal initialization of 'constinit' entity with a non-constant expression


chronoxor:
Visual Studio 2022 generates compilation error in arena.cc:
```
protobuf\src\google\protobuf\arena.cc(457,51): error C2127: 'thread_cache_': illegal initialization of 'constinit' entity with a non-constant expression
```

chronoxor:
Fixed & Merged

Issue #9698, opened 2022-03-29
error C2127: illegal initialization of 'constinit' entity with a non-constant expression


chronoxor:
**What version of protobuf and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Windows 10

**What runtime / compiler are you using (e.g., python version or gcc version)**
Visual Studio 2022

**What did you do?**

`repeated Order orders` protobuf definition generates compile error in Visual Studio 2022
```
trade.pb.cc(53,111): error C2127: 'Trade::protobuf::_Account_default_instance_': illegal initialization of 'constinit' entity with a non-constant expression
```

for the following .proto file:
```
syntax = "proto3";
package Trade.protobuf;

message Order
{
    int32 id = 1;
    string symbol = 2;
}

message Account
{
    int32 id = 1;
    string name = 2;
    repeated Order orders = 3;
}
```

jorgbrown:
FWIW, I copied the generated code, or at least, enough of it to allow _Account_default_instance_ to be defined, and was unable to reproduce this error.  https://godbolt.org/z/j97bvjq44

chronoxor:
Here is a reproduced issue in godbolt - https://godbolt.org/z/44bsG5733

ReeceSX:
1>C:\Users\Admin\Desktop\Aurora2\Scriptability\Vendor\protobuf\src\google\protobuf\api.pb.cc(42,107): error C2127: 'google::protobuf::_Api_default_instance_': illegal initialization of 'constinit' entity with a non-constant expression
`PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT PROTOBUF_ATTRIBUTE_INIT_PRIORITY1 ApiDefaultTypeInternal _Api_default_instance_;`

1>C:\Users\Admin\Desktop\Aurora2\Scriptability\Vendor\protobuf\src\google\protobuf\api.pb.cc(61,110): error C2127: 'google::protobuf::_Method_default_instance_': illegal initialization of 'constinit' entity with a non-constant expression
1>generator.cc
`
PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT PROTOBUF_ATTRIBUTE_INIT_PRIORITY1 MethodDefaultTypeInternal _Method_default_instance_;`

_fun_

jgcodes2020:
The bug seems like it would be caused by #10159.

[...]

PR #11674, opened 2023-01-26
Fixed Visual Studio 2022: protobuf\src\google\protobuf\arena.cc(457,51): error C2127: 'thread_cache_': illegal initialization of 'constinit' entity with a non-constant expression #11672

chronoxor:


src/google/protobuf/arena.cc
```diff
@@ -453,7 +453,7 @@ ThreadSafeArena::ThreadCache& ThreadSafeArena::thread_cache() {
   return thread_cache;
 }
 #else
-ABSL_CONST_INIT PROTOBUF_THREAD_LOCAL
```
mkruskal-google, line 456: Can we change this to PROTOBUF_CONSTINIT?  That handles this MSVC issue we've encountered before, without affecting linux

mkruskal-google: e.g. #8688 and #9698 are two other cases we hit this

chronoxor: Yes, changing ABSL_CONST_INIT -> PROTOBUF_CONSTINIT fixed the issue! I've updated the PR

mkruskal-google (approved):

PR #11691, opened 2023-01-27
[C#] Optimize reading/writing float and double

yiting2002:
Unsafe.As has better performance than BitConverter in benchmark result.


jskeet:
Do you have benchmarks showing the difference for actual protobuf usage? This is the sort of change that has to be reviewed *very* carefully - I'm happy to do so if we know that it makes a significant difference, but if the difference is marginal I'd prefer not to. (My time available for protobuf work is very limited.)

Issue #7806, opened 2020-08-14
ruby: repeated_field aliases each_index to each_with_index - thus incorrectly returning the each not the index


jufemaiz:
Aliased method for `each_index` is invalid:
https://github.com/protocolbuffers/protobuf/blob/63d2f3bc80a870a09c6aea74025df40aad7dba04/ruby/lib/google/protobuf/repeated_field.rb#L106-L110

Ref:
- array: https://ruby-doc.org/core-2.7.1/Array.html#method-i-each_index
- enumerable: https://ruby-doc.org/core-2.7.1/Enumerable.html#method-i-each_with_index

**What version of protobuf and what language are you using?**
Version: master
Language: ruby

**What operating system (Linux, Windows, ...) and version?**

Docker: ruby:2.7-slim-buster

**What runtime / compiler are you using (e.g., python version or gcc version)**

N/A

**What did you do?**
1. have a proto with a repeated_field
2. build a ruby library
3. make use of `each_index`

**What did you expect to see**

The index of the repeated field.

**What did you see instead?**

The repeated field's value instead.

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**

N/A

PR #11767, opened 2023-02-02
In Ruby repeated fields, each_index actually iterates over the index

shaldengeki:
Currently we're aliasing `each_index` to `each_with_index`, incorrectly passing both the index and the value of a repeated field to the block.

What we want is to just pass the index. Luckily this is a method on Ruby arrays, so we just wrap the native Ruby array method.

Fixes https://github.com/protocolbuffers/protobuf/issues/7806

### DIFF
--- a/ruby/compatibility_tests/v3.0.0/tests/repeated_field_test.rb
+++ b/ruby/compatibility_tests/v3.0.0/tests/repeated_field_test.rb
@@ -119,6 +119,20 @@ class RepeatedFieldTest < Test::Unit::TestCase
     assert_equal ['string'] * 5, result
   end
 
+  
+  def test_each_index
+    m = TestMessage.new
+    5.times{|i| m.repeated_string << 'string' }
+
+    expected = 0
+    m.repeated_string.each_index do |idx|
+      assert_equal expected, idx
+      expected += 1
+      assert_equal 'string', m.repeated_string[idx]
+    end
+    assert_equal 5, expected
+  end
+
 
   def test_empty?
     m = TestMessage.new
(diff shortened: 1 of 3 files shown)


shaldengeki:
Oh, hm. Should I not be working off of my fork? From the failing build:

```
This pull request is from an unsafe fork and hasn't been approved to run tests!
```

I can just branch off this repo, I suppose, if that's more convenient for y'all in the future.

shaldengeki:
@haberman 👋 Hi there! Anything I can do to help move this forward? (No worries if not, just making sure I'm not missing anything.)

haberman:
HI @shaldengeki, this looks good to me. I think the only question is whether we need to have a major version bump for this. Users could technically be relying on the old buggy behavior.

shaldengeki:
It's true! This is definitely a breaking change in the API. What does a
major bump entail? (does it just mean "we'll wait to merge this"? happy to
let this sit.)

shaldengeki:
@haberman just wanted to check in and set expectations -- let me know if I should be doing anything here.

haberman:
Thanks for checking in. No action is required on your part, I think we will plan to merge this in coordination with the next major version bump in Ruby, whenever that is (I think it will happen sometime this year).

shaldengeki:
@haberman apologies if this is noise - I see v24 release candidates are out, should we try to get this in?
https://github.com/protocolbuffers/protobuf/releases/tag/v24.0-rc3

shaldengeki:
@haberman happy Thanksgiving, and touching base once again on this; let me know what I can do to help.

haberman:
We are planning to bump the major version of Ruby in February. So it's time to move forward with a breaking change like this.

Thanks for your patience.

haberman (approved):

zhangskz (approved):

haberman:
This is now announced in https://protobuf.dev/news/2023-12-27/#ruby-breaking-changes

PR #11793, opened 2023-02-03
Ruby: Implement Write Barriers

casperisfine:
Write barrier protected objects are allowed to be promoted to the old generation, which means they only get marked on major GC.

The downside is that the `RB_BJ_WRITE` macro MUST be used to set references, otherwise the referenced object may be garbaged collected.

But the `*Descriptor` classes and `Arena` have very few references and are only set in a few places, so it's relatively easy to implement.

cc @peterzhu2118

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -223,6 +223,8 @@ static void DescriptorPool_register(VALUE module) {
 
 typedef struct {
   const upb_MessageDef* msgdef;
+// IMPORTANT: WB_PROTECTED objects must only use the RB_OBJ_WRITE()
+// macro to update VALUE references, as to trigger write barriers.
   VALUE klass;
   VALUE descriptor_pool;
 } Descriptor;
@@ -238,7 +240,7 @@ static void Descriptor_mark(void* _self) {
 static const rb_data_type_t Descriptor_type = {
     "Google::Protobuf::Descriptor",
     {Descriptor_mark, RUBY_DEFAULT_FREE, NULL},
-    .flags = RUBY_TYPED_FREE_IMMEDIATELY,
+    .flags = RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 static Descriptor* ruby_to_Descriptor(VALUE val) {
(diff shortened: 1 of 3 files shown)


casperisfine:
I just noticed the `"Message"` struct was also from protobuf, and easy to add write barriers for.

For context, here's we have a couple thousands of these instances in our heap that need to be mark on every minor GC:

```
 ["DATA(Google::Protobuf::Internal::Arena)", 1268],
 ["DATA(Message)", 1268],
...
 ["DATA(Google::Protobuf::EnumDescriptor)", 77],
 ["DATA(Google::Protobuf::FileDescriptor)", 68],
```

casperisfine:
@JasonLunn @haberman sorry for the ping, but is this on your radar? We have about 3.5k of these objects on our heap, and we'd love if we didn't have to mark them on every minor GC.

haberman:
It looks like there is an assertion failure on Ruby 2.7:

```
....ruby: ruby/ext/google/protobuf_c/protobuf.c:404: void ObjectCache_Add(const void *, VALUE): Assertion `ObjectCache_Get(key) == val' failed.
Aborted (core dumped)
```

Also could you rebase the PR from main? There have been some fixes to the tests in the meantime.

haberman:
It would be helpful to add comments above the struct members that must only be changed through `RB_OBJ_WRITE()`, to caution future readers/writers of this invariant.

casperisfine:
@haberman done. I rebased and added a comment above each reference to WB_PROTECTED.

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -238,7 +238,9 @@ static void Descriptor_mark(void* _self) {
 static const rb_data_type_t Descriptor_type = {
     "Google::Protobuf::Descriptor",
     {Descriptor_mark, RUBY_DEFAULT_FREE, NULL},
-    .flags = RUBY_TYPED_FREE_IMMEDIATELY,
+    // IMPORTANT: WB_PROTECTED objects must only use the RB_OBJ_WRITE()
```
haberman, line 241: Could we put this comment immediately above the VALUE struct members?

eg.

```
typedef struct {
  const upb_MessageDef* msgdef;
  // IMPORTANT: these VALUE references may only be mutated with
  // RB_OBJ_WRITE() due to our use of RUBY_TYPED_WB_PROTECTED.
  VALUE klass;
  VALUE descriptor_pool;
} Descriptor;
```

casperisfine:
@haberman done

haberman (approved):

casperisfine:
Sorry to be a bother, but is there any other blockers I can help with?

haberman:
This was merged in https://github.com/protocolbuffers/protobuf/commit/d82d8a48f6c50ae6c811dbd6b7383e36a691c6b3

casperisfine:
🙌 

Thank you!

PR #11895, opened 2023-02-09
Merge main to 22.x

zhangskz:


thomasvl (approved):

.github/workflows/test_runner.yml
```diff
@@ -0,0 +1,161 @@
...
+    # Run daily at 10 AM UTC (2 AM PDT)
+    - cron: 0 10 * * *
+
+  # postsubmit
+  push:
+    branches:
+      - main
+      - '[0-9]+.x'
+      # The 21.x and 22.x branches still use Kokoro
+      - '!2[12].x'
```
mkruskal-google, line 24: We should update all of these to not exclude 22.x now (I have no idea why the tests even ran)

zhangskz: Done. I think these were initially triggered because I accidentally created the PR as a merge to main and then switched it to 22.x. 

Also, not sure why but the deleted failing kokoro tests seem to have disappeared after pushing -- no complaints there :)

PR #12026, opened 2023-02-23
Use PROTOBUF_DEPRECATED instead of PROTOBUF_DEPRECATED_ENUM

georgthegreat:
At the time the way of marking some stuff `deprecated` is overcomplicated taken into account that protobuf mandates C++11 to be built.

Particularly, PROTOBUF_DEPRECATED_ENUM references gcc-6 release notes and the compiler is almost 4.5 years old as of today.
I believe it should be possible to simplify the definitions.


georgthegreat:
Why we need this:

We have `-Werror-deprecated` set up in our codebase and every protobuf update mandates some adapation of PROTOBUF_DEPRECATED patch we apply locally.

I would like to simplify patch adapation.

src/google/protobuf/compiler/cpp/enum.cc
```diff
@@ -302,7 +302,7 @@ void EnumGenerator::GenerateSymbolImports(io::Printer* p) const {
         {
             Sub("VALUE", EnumValueName(enum_->value(j))).AnnotatedAs(value),
             {"DEPRECATED",
-             value->options().deprecated() ? "PROTOBUF_DEPRECATED_ENUM" : ""},
+             value->options().deprecated() ? "PROTOBUF_DEPRECATED" : ""},
```
georgthegreat, line 305: I believe that I need to run `generate_descriptor_proto.sh` to regenerate prebuilt `pb{.cc,.h}` files, but running the script gives the changes in php sources only.

georgthegreat: I got it.
There are no deprecated enum values in builtin proto files, hence there is no need to regenerate them.

fowles (approved):

georgthegreat:
@fowles, can we proceed with merging this PR?

fowles:
sorry, a bit of internal tooling fell over.

PR #12052, opened 2023-02-24
Enforce Java 8 Compatibility

googleberg:
Update maven and bazel build files to use -target 8 -source 8 and error prone Java8ApiChecker to enforce Java 8 compatibility.


java/pom.xml
```diff
@@ -111,8 +112,30 @@
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.6.1</version>
           <configuration>
-            <source>1.7</source>
-            <target>1.7</target>
+          <source>8</source>
+          <target>8</target>
+          <encoding>UTF-8</encoding>
+              <compilerArgs combine.children="append">
```
perezd, line 118: Is this the only POM that needs updating? There's many, not sure how this works.

googleberg: I think this would be the only one. But evidently Mike and Deanna have bigger better plans, so they're going to learn from this but do something better.

PR #12055, opened 2023-02-24
Add java8 tests and build flags

deannagarcia:
- Adds a java_opts.bzl file that creates a set of common java options and creates macros for `protobuf_java_library` and `protobuf_java_export` that automatically use those options
- Adds a java8 test
- Updates docker images for other java tests

### DIFF
--- a/java/internal/JavaVersionTest.java
+++ b/java/internal/JavaVersionTest.java
@@ -40,12 +40,14 @@ public class JavaVersionTest {
   @Test
   public void testJavaVersion() throws Exception {
     String exp = System.getenv("KOKORO_JAVA_VERSION");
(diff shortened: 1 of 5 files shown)


java/internal/JavaVersionTest.java
```diff
@@ -44,8 +44,10 @@ public void testJavaVersion() throws Exception {
       System.err.println("No kokoro java version found, skipping check");
       return;
     }
+    // Java 8's version is read as "1.8"
+    if(exp.equals("8")) exp = "1.8";
```
mkruskal-google, line 48: Nice, good to see this infra still works! :)

mkruskal-google (approved):

PR #12064, opened 2023-02-27
Restore using protoc as protobuf compiler binary name

georgthegreat:
This partially reverts changes made in #4579.

At the time protoc binary changes its filename with every new protobuf version being released.

### DIFF
--- a/cmake/protoc.cmake
+++ b/cmake/protoc.cmake
@@ -9,6 +9,3 @@ target_link_libraries(protoc
   ${protobuf_ABSL_USED_TARGETS}
 )
 add_executable(protobuf::protoc ALIAS protoc)
-
-set_target_properties(protoc PROPERTIES
-    VERSION ${protobuf_VERSION})


devjgm:
Hi. Can you please say a little more about the problem that currently exists and how this is fixing the problem? My understanding is that _with_ the VERSIOn property on the protoc binary, it will end up getting installed like:

```shell
lrwxr-xr-x  1 me  group       15 Feb 23 09:53 protoc@ -> protoc-4.22.0.0
-rwxr-xr-x  1 me  group  5521024 Feb 23 09:53 protoc-4.22.0.0*
```

So you can still use `protoc` as the binary name. What's the problem you're seeing?

georgthegreat:
We use a semi-automaic tool for building custom build configurations (i. e. generate internal build system description using original cmake build).
It works as follows:

1. Run original build and trace it by the means of strace system call
2. Taken its results (binaries, static and dynamic libraries) and reconstruct the build graph
3. Convert graph into build configuration

We operate binary names and library names to configure the non-automatic part of the system.
At the time we have to change binary names with every protobif version update.

I am trying to solve this issue by removing version number from the protoc binary being built.
Is it possible to move the versioning manupulation onto install stage of the cmake (out from build stage)?

mkruskal-google:
My point is we *also* produce a `protoc` file, why not just use that?  The binary version does match the library version for the most part.  The `3.` prefix is referring to the C++ runtime version, which actually appears to be a bug

georgthegreat:
> My point is we also produce a protoc file

This is not a file produced by linker, it is just a softlink to the actual output.

Anyway, I believe that fixing versioning schema will be helpful.

mkruskal-google:
Perfect, thanks for catching that bug!

PR #12098, opened 2023-02-28
Fix pom file to compile for java 8

deannagarcia:
Makes java 8 the source/target and adds errorprone warnings to check for java 8 compatibility if not run by java 8.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -29,6 +29,10 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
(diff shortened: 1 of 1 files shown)


java/pom.xml
```diff
@@ -273,6 +288,35 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>run-error-prone</id>
+      <activation>
+        <jdk>[11,)</jdk>
```
googleberg, line 294: Do we want to support jdk 9?
Or does error prone support start at 11?

googleberg (commented):
Should I fix the tests to be Java 8 compatible as well?

java/pom.xml
deannagarcia: Error prone starts at 11, so luckily that means we are getting the 11 & 17 tests and I think that should be enough.

googleberg (approved):

PR #12163, opened 2023-03-07
Use [[deprecated]] attribute from c++14 to mark fields deprecated

georgthegreat:
Using standard C++ attribute to remove compiler-based switch.

This continues #12026.


georgthegreat:
While abseil mandates C++14 it still uses __attribute__ (and thus does not issue a warning undef MSVC)

```
#if ABSL_HAVE_ATTRIBUTE(deprecated)
#define ABSL_DEPRECATED(message) __attribute__((deprecated(message)))
#else
#define ABSL_DEPRECATED(message)
#endif
```

fowles:
Yes, but that will allow all of google's main projects to be consistent and to update to newer things at the same time, which eliminates a bunch of compatibility skew concerns for us

georgthegreat:
I will close this PR then.

PR #12174, opened 2023-03-08
[C#] Replace regex that validates descriptor names

JamesNK:
This PR replaces the descriptor name validation regex with a validation method. This change allows the `System.Text.RegularExpressions` engine to be trimmed away in published apps that do standard protobuf serialization.

There are some other usages of `Regex` in Google.Protobuf, but they in `JsonParser`. They are only included in a published app if `JsonParser` is used.

Another benefit is a slightly faster app startup time. The removed regex was compiled, which has a high-ish fixed cost.

cc @jskeet @jtattermusch

### DIFF
--- a/csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
+++ b/csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
@@ -187,35 +187,24 @@ namespace Google.Protobuf.Reflection
                 throw new DescriptorValidationException(descriptor, "Missing name.");
             }
 
-            if (!IsValidSymbolNameFormat(descriptor.Name))
-            {
-                throw new DescriptorValidationException(descriptor,
-                                                        "\"" + descriptor.Name + "\" is not a valid identifier.");
-            }
-        }
-
-        /// <summary>
-        /// Symbol name must start with a letter or underscore, and it can contain letters, numbers and underscores.
-        /// </summary>
-        private static bool IsValidSymbolNameFormat(string name)
-        {
+            // Symbol name must start with a letter or underscore, and it can contain letters, numbers and underscores.
+            string name = descriptor.Name;
             if (!IsAsciiLetter(name[0]) && name[0] != '_')
             {
-                return false;
+                ThrowInvalidSymbolNameException(descriptor);
             }
-
             for (int i = 1; i < name.Length; i++)
             {
                 if (!IsAsciiLetter(name[i]) && !IsAsciiDigit(name[i]) && name[i] != '_')
                 {
-                    return false;
+                    ThrowInvalidSymbolNameException(descriptor);
                 }
             }
 
-            return true;
-
             static bool IsAsciiLetter(char c) => (uint) ((c | 0x20) - 'a') <= 'z' - 'a';
-            static bool IsAsciiDigit(char c) => (uint) (c - '0') <= ('9' - '0');
+            static bool IsAsciiDigit(char c) => (uint) (c - '0') <= '9' - '0';
+            static void ThrowInvalidSymbolNameException(IDescriptor descriptor) =>
+                throw new DescriptorValidationException(descriptor, "\"" + descriptor.Name + "\" is not a valid identifier.");
         }
 
         /// <summary>
--- a/csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
+++ b/csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
@@ -32,6 +32,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Text;
 using System.Text.RegularExpressions;
 
@@ -176,8 +177,6 @@ namespace Google.Protobuf.Reflection
             descriptorsByName[fullName] = descriptor;
         }
 
-        private static readonly Regex ValidationRegex = new Regex("^[_A-Za-z][_A-Za-z0-9]*$", FrameworkPortability.CompiledRegexWhereAvailable);
-
         /// <summary>
         /// Verifies that the descriptor's name is valid (i.e. it contains
         /// only letters, digits and underscores, and does not start with a digit).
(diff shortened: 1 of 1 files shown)


csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
```diff
@@ -189,13 +186,38 @@ private static void ValidateSymbolName(IDescriptor descriptor)
...
+                if (!IsAsciiLetter(name[i]) && !IsAsciiDigit(name[i]) && name[i] != '_')
+                {
+                    return false;
+                }
+            }
+
+            return true;
+
+            static bool IsAsciiLetter(char c) => (uint) ((c | 0x20) - 'a') <= 'z' - 'a';
+            static bool IsAsciiDigit(char c) => (uint) (c - '0') <= ('9' - '0');
```
jskeet, line 218:
```suggestion
            static bool IsAsciiDigit(char c) => (uint) (c - '0') <= '9' - '0';
```

(Or add the parentheses to IsAsciiLetter.)

csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
```diff
@@ -189,13 +186,38 @@ private static void ValidateSymbolName(IDescriptor descriptor)
...
+            {
+                if (!IsAsciiLetter(name[i]) && !IsAsciiDigit(name[i]) && name[i] != '_')
+                {
+                    return false;
+                }
+            }
+
+            return true;
+
+            static bool IsAsciiLetter(char c) => (uint) ((c | 0x20) - 'a') <= 'z' - 'a';
```
jskeet, line 217: Possibly use `unchecked` to avoid throwing exceptions if this is ever compiled in a checked-by-default context and arithmetic overflows? (I believe we currently don't use checked-by-default, but it's not inconceivable that we might do so in the future, and the overflow here really *wouldn't* be a problem.)

csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
```diff
@@ -189,13 +186,38 @@ private static void ValidateSymbolName(IDescriptor descriptor)
...
                                                         "\"" + descriptor.Name + "\" is not a valid identifier.");
             }
         }
 
+        /// <summary>
+        /// Symbol name must start with a letter or underscore, and it can contain letters, numbers and underscores.
+        /// </summary>
+        private static bool IsValidSymbolNameFormat(string name)
+        {
+            if (!IsAsciiLetter(name[0]) && name[0] != '_')
```
jskeet, line 202: Possibly add a comment indicating that an empty name is already checked before the call to this method?
I'd also probably make it a local method to ValidateSymbolName - otherwise it *could* be called by other code in this class in the future with an empty string.

jskeet (commented):
Definitely happy with the overall aim - a few nits.

csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
JamesNK: Refactored to remove this method.

csharp/src/Google.Protobuf/Reflection/DescriptorPool.cs
JamesNK: Are you sure? Feels a little YAGNI. And there are only 3 `unchecked` statements in the source today.

btw I copied `IsAsciiLetter` directly from https://github.com/dotnet/runtime/blame/faa1d0993afad9d751b2dca6bcfe356369ec6c08/src/libraries/System.Private.CoreLib/src/System/Char.cs#L236 (added in .NET 6 which Google.Protobuf doesn't target)

JamesNK:
All feedback applied.

jskeet (approved):

JamesNK:
Is this waiting on more reviews? I just want to make sure it doesn't get left behind.

For context, this change saves 1 MB from .NET Native gRPC client and server apps. I'd love to show off the smallest gRPC apps possible in demos 😄

jskeet:
Thanks for the prod - I'll chase it up.

jskeet:
@deannagarcia: I think your approval (or at least, someone actually on the Protobuf team) is required before this can proceed.

JamesNK:
This missed 3.23.0.

Could it be merged now to make it into the next protobuf release?

jskeet:
@JamesNK: Aargh, sorry that this fell through the cracks.
@deannagarcia: Please could you merge?

deannagarcia:
Sorry this fell through, working to get it in now.

jskeet:
@JamesNK: This went out in 3.23.1 :)

JamesNK:
Nice. Thanks!

robertodalmonte:
Hi there,
This morning I updated my WPF project from:
<PackageReference Include="Google.Protobuf" Version="3.23.0" />
to
<PackageReference Include="Google.Protobuf" Version="3.23.1" />

This is the only change I made.
I published my project using my usual settings:
<Project>
  <PropertyGroup>
    <Configuration>Release</Configuration>
    <Platform>Any CPU</Platform>
    <PublishDir>D:\Github\OneClick\OneClick.Dental\bin\publish</PublishDir>
    <PublishProtocol>FileSystem</PublishProtocol>
    <_TargetId>Folder</_TargetId>
    <TargetFramework>net7.0-windows</TargetFramework>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
    <SelfContained>false</SelfContained>
    <PublishSingleFile>true</PublishSingleFile>
    <PublishReadyToRun>true</PublishReadyToRun>
  </PropertyGroup>
</Project>
If I try to launch my WPF client it stops as soon as trying to connect to my gRPC server.
It does not happen if I do the same from Visual Studio in debug or if I launch the exe from the release folder after building.
I rolled back to <PackageReference Include="Google.Protobuf" Version="3.23.0" />
and I can publish my WPF project successfully and I'm able to connect as usual to my gRPC Server app.
Is it possible that the last changes might have affected publishing?
Should I look somewhere else?
Best regards
Roberto

robertodalmonte:
from 3.23 to 3.23.1

jskeet:
@robertodalmonte: Please file a new issue in https://github.com/grpc/grpc-dotnet with more details - it's not appropriate to use a closed PR to report an issue, even if the issue was caused by the change.

I'd also *strongly* recommend that in the new issue, you give more detail than "it stops". I expect there's an exception being thrown somewhere - please look for that and give full details of it in the issue.

Issue #12170, opened 2023-03-08
Java 3.22.1: broken pom.xml


xuwei-k:
**What version of protobuf and what language are you using?**

Java 3.22.1

**What operating system (Linux, Windows, ...) and version?**

all

**What runtime / compiler are you using (e.g., python version or gcc version)**

**What did you do?**

use `protobuf-java-util` 3.22.1 with sbt

```scala
libraryDependencies += "com.google.protobuf" % "protobuf-java-util" % "3.22.1"
```

**What did you expect to see**

**What did you see instead?**

```
[error]   not found: https://repo1.maven.org/maven2/$com/google/protobuf/protobuf-java//protobuf-java-.pom
```

https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java-util/3.22.1/protobuf-java-util-3.22.1.pom

<img width="537" alt="broken-pom-xml" src="https://user-images.githubusercontent.com/389787/223585176-64979062-ca25-4263-896c-2279fb02c2b4.png">

**Anything else we should know about your project / environment**

XN137:
seeing same problem on 3.22.1, compare these poms:
https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java-util/3.22.0/protobuf-java-util-3.22.0.pom
https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java-util/3.22.1/protobuf-java-util-3.22.1.pom

@deannagarcia i guess the problem was introduced [here](https://github.com/protocolbuffers/protobuf/commit/9b5a2c6f6fc76da3c93f6793550fa488ae62abaf) by using `${groupId}` instead of `{groupId}`

deannagarcia:
Thanks for the report and diagnosing the issue, I see the error and am working on a fix now and we will get another release out with the fix.

xenomachina:
FWIW, we're also running into this issue when building with gradle.

Both releases since the pom templates were added have broken our build: missing dependencies in 3.22.0, and now this in 3.22.1. Perhaps it would be worthwhile to diff the poms currently being generated against the corresponding poms from 3.21.12 (the last version before the poms started having issues) and confirm that every difference makes sense.

xuwei-k:
👍 

- https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java-util/3.22.2/protobuf-java-util-3.22.2.pom
- https://github.com/protocolbuffers/protobuf/pull/12194
[...]

PR #12177, opened 2023-03-08
Fix pom file errors caused by the extra "$"

deannagarcia:
Fixes https://github.com/protocolbuffers/protobuf/issues/12170

### DIFF
--- a/java/util/pom_template.xml
+++ b/java/util/pom_template.xml
@@ -14,7 +14,7 @@
   <description>Utilities for Protocol Buffers</description>
   <dependencies>
     <dependency>
-      <groupId>${groupId}</groupId>
+      <groupId>{groupId}</groupId>
       <artifactId>protobuf-java</artifactId>
     </dependency>
     {dependencies}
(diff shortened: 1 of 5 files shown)


perezd (approved):

java/util/pom_template.xml
```diff
@@ -32,7 +32,7 @@
             <Automatic-Module-Name>com.google.protobuf.util</Automatic-Module-Name> <!-- Java9+ Jigsaw module name -->
             <Bundle-DocURL>https://developers.google.com/protocol-buffers/</Bundle-DocURL>
             <Bundle-SymbolicName>com.google.protobuf.util</Bundle-SymbolicName>
-            <Export-Package>com.google.protobuf.util;version=${project.version}</Export-Package>
+            <Export-Package>com.google.protobuf.util;version={project.version}</Export-Package>
```
XN137, line 35: since we see `com.google.protobuf.util;version=${project.version}` in the [generated pom](https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java-util/3.22.1/protobuf-java-util-3.22.1.pom) of `3.22.1`

i am thinking this needs to use `{version}` instead (see line 7)

if `{project.version}` would be available, the generated pom would have looked like this:
```
com.google.protobuf.util;version=$3.22.1
```

XN137: i think both variants work, the plugin configuration probably does not even need to be in the published pom

what is important is that the version can be evaluated correctly when the build is generating the bundle, so that in the jar we see the right values in the `META-INF/MANIFEST.MF` file (snippet from `protobuf-java-util-3.21.12.jar`):
```
...
Bundle-SymbolicName: com.google.protobuf.util
Bundle-Version: 3.21.12
Created-By: Apache Maven Bundle Plugin
Export-Package: com.google.protobuf.util;version="3.21.12";uses:="com.go
 ogle.protobuf,javax.annotation"
...
```

this can be achieved by "hardcoding" the version during bazel replacement or by letting the bundle maven plugin evaluate `${project.version}` correctly (as done in the follow-up PR where `$` was added back)

PR #12212, opened 2023-03-12
python: make ctypes dependency optional - continued

twoism:
This branch continues the work from https://github.com/protocolbuffers/protobuf/pull/11571. Starting a new PR seemed like the best way to take this over. Happy to switch if there's a preferred way to do so. 

**Original Description**

This allows the protobuf python library to work on cpython 3.11's new wasm target [1], which does not have ctypes.

This reverts back to the way TruncateToFourByteFloat worked before 75eff10


twoism:
Hi @anandolee! Is there anything I need to do in order to re-run these checks now that my CLA is signed? thanks!

python/google/protobuf/internal/type_checkers.py
```diff
@@ -45,7 +45,7 @@
 
 __author__ = 'robinson@google.com (Will Robinson)'
 
-import ctypes
+import struct
```
anandolee, line 48: Use the original try import struct https://github.com/protocolbuffers/protobuf/commit/75eff10d81233d3d08ce5ded215e526c1fcace88

And add comment that this is for cpython 3.11's new wasm target

twoism: @anandolee Updated! It seems this fork has not been approved to run tests, though. Is there anything else I need to do to get approval. thanks again!

twoism:
Hi! Just bumping this PR @anandolee. Thanks!

python/google/protobuf/internal/type_checkers.py
```diff
@@ -57,7 +62,7 @@
 
 
 def TruncateToFourByteFloat(original):
-  return ctypes.c_float(original).value
+  return struct.unpack('<f', struct.pack('<f', original))[0]
```
anandolee, line 65: If able to import ctypes, there's no struct. This should be:

   
 if ctypes:
      return ctypes.c_float(converted_value).value
    else:
      return struct.unpack('<f', struct.pack('<f', converted_value))[0]

haberman: It appears that `struct` is available even in the most recent versions of Python: https://docs.python.org/3/library/struct.html

Can't we just use `struct` all the time?

Jeremy-Boyle:
What happened to this pr ?

anandolee:
Sorry I've forgot this PR. Just merged the change

Issue #12017, opened 2023-02-22
OSGi manifest headers missing in 3.22.0/3.22.1/3.22.2 [Java]


sjaenick:
Hi there,

I just attempted to upgrade `protobuf-java` from 3.21.12 to 3.22.0 via maven, and noticed that
3.22.0 no longer contains OSGi manifest headers. Could this be fixed?

New dependency is:

```
<dependency>
            <groupId>com.google.protobuf</groupId>
            <artifactId>protobuf-java</artifactId>
            <version>3.22.0</version>
</dependency>
```

Thanks in advance.

deannagarcia:
Thanks for this report! Working on a fix now and it should be out in 22.1 this week or next week.

deannagarcia:
Sorry you're still hitting this issue, I think https://github.com/protocolbuffers/protobuf/issues/12170 might be the root cause. I'm working on that now.

sjaenick:
Sorry, but stiill no luck. Maybe add a regression test to make sure `META-INF/MANIFEST.MF` has appropriate `Export-Package` statements?

deannagarcia:
We are unfortunately aware that this didn't get fixed in 22.2. As linked, we are working on a fix to this now and will keep you updated. In the meantime it's probably best to use 21.12.

googleberg:
@sjaenick I have a fix out for review. We don't have any OSGi experts on our team, would you be interested in contributing a test that would actually make sure that we don't break this in the future?

sjaenick:
> @sjaenick I have a fix out for review. We don't have any OSGi experts on our team, would you be interested in contributing a test that would actually make sure that we don't break this in the future?

Could you take a look at https://github.com/protocolbuffers/protobuf/pull/12235 ?

deannagarcia:
Closing this since the fix is in, we will be releasing this in 3.22.3 in the next few weeks.

deannagarcia:
Update: I'm releasing 3.22.3 without this fix, but it will be in 23.0 which should be out next week.

PR #12235, opened 2023-03-15
Introduce tests for java manifest contents

sjaenick:
As per https://github.com/protocolbuffers/protobuf/issues/12017#issuecomment-1469334478, add basic
testing functionality for whatever is supposed to be present in the generated `META-INF/MANIFEST.MF`;  at
the moment, this currently successfully completes for `main` during a maven build, but doesn't seem to be
picked up by `bazel` at all.

As I'm completely unfamiliar with `bazel`, some guidance would be appreciated.

### DIFF
--- a/java/core/src/test/java/com/google/protobuf/ManifestTest.java
+++ b/java/core/src/test/java/com/google/protobuf/ManifestTest.java
@@ -1,0 +1,87 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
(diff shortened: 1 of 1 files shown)


googleberg (commented):
This looks like a test for ensuring that the content that is there today is at least partially there tomorrow, but is there any way to know if the content is correct? If we add a new dependency that doesn't get included in the Import-Package for example, would that break OSGi users?

sjaenick:
> This looks like a test for ensuring that the content that is there today is at least partially there tomorrow, but is there any way to know if the content is correct? If we add a new dependency that doesn't get included in the Import-Package for example, would that break OSGi users?

Yes, you are correct, these tests are pretty basic. To validate OSGi behavior in its entirety, it would be necessary to

- setup an OSGi container
- deploy the protobuf bundle
- deploy another bundle that *imports* `com.google.protobuf`
- validate both bundles are fully resolved

Not sure how to approach this.

PR #12244, opened 2023-03-16
feat: support riscv64

ernado:
I want to add riscv64 support for protobuf.
The protobuf is part of Kubernetes build pipeline, so this change is required to add riscv64 support to Kubernetes.

I'm not quite familiar with profobuf codebase, please guide me how I can cross-compile protobuf for riscv64, I'll test it on my devboard.

Fix: #12266

### DIFF
--- a/protobuf_release.bzl
+++ b/protobuf_release.bzl
@@ -22,6 +22,8 @@ def _package_naming_impl(ctx):
     cpu = "aarch_64"
   elif cpu == "ppc64":
     cpu = "ppcle_64"
+  elif cpu == "riscv64":
+    cpu = "riscv_64"
 
   # use the system name to determine the os and then create platform names
   if "apple" in system_name:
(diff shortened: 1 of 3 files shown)


zhangskz:
Unfortunately, adding support for riscv64 would also require us to update our internal release process and images accordingly to compile for riscv64, which we are unlikely to be able to staff at the moment especially considering lack of support in manylinux currently: https://github.com/pypa/manylinux/discussions/1426

Per https://github.com/kubernetes/kubernetes/pull/123661#issuecomment-1977297851, it appears Kubernetes itself does not officially support riscv64 either. Closing accordingly, but we can revisit this in the future if demand changes.

luhenry:
Hello! Bump on that PR since https://github.com/pypa/manylinux/discussions/1426 has been fixed with https://github.com/pypa/manylinux/pull/1743 being merged.

PR #12254, opened 2023-03-17
Update deprecated aliases for Python 3.12 compatibility.

tirkarthi:
https://docs.python.org/3.12/whatsnew/3.12.html#removed . `assertRaisesRegexp` was removed in favor of `assertRaisesRegex`


Nooknook88:
> https://docs.python.org/3.12/whatsnew/3.12.html#removed . `assertRaisesRegexp` was removed in favor of `assertRaisesRegex`

fowles (approved):

PR #12261, opened 2023-03-17
Fix compile-time error for Google.Protobuf in C# 11

### LINKED ISSUE #11004: Google.Protobuf has errors when compiled with C# 11
This was noted in #10978:

> By the way, I noticed that updating Google.Protobuf to use C# 11 introduced new compilation errors in Google.Protobuf. It doesn't like how the library uses ref structs. I didn't investigate.

While we're not using C# 11 yet, it would be good to make sure it *will* be able to compile with C# 11...

jskeet:
Note that the error only affects users trying to build Google.Protobuf directly using C# 11. It doesn't affect users who are compiling code against the pre-built library (which is likely to be basically everyone).

Fixes #11004.

(This is code I'm less familiar with than most of the library, and I'd be nervous of making any more wide-ranging changes. This seems pretty safe though.)

cc @JamesNK who first noticed this.

### DIFF
--- a/csharp/src/Google.Protobuf/ParseContext.cs
+++ b/csharp/src/Google.Protobuf/ParseContext.cs
@@ -64,8 +64,9 @@ namespace Google.Protobuf
             state.recursionLimit = DefaultRecursionLimit;
             state.currentLimit = int.MaxValue;
             state.bufferSize = buffer.Length;
-
-            Initialize(buffer, ref state, out ctx);
+            // Equivalent to Initialize(buffer, ref state, out ctx);
+            ctx.buffer = buffer;
+            ctx.state = state;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


csharp/src/Google.Protobuf/ParseContext.cs
```diff
@@ -74,6 +75,7 @@ internal static void Initialize(ReadOnlySpan<byte> buffer, out ParseContext ctx)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal static void Initialize(ReadOnlySpan<byte> buffer, ref ParserInternalState state, out ParseContext ctx)
```
amanda-tarafa, line 76: I guess we are not using this anywhere else? Else we'd likely get the same build errors elsewhere? Can't you just remove this internal method?

amanda-tarafa: Ah yes I see, because the `state` used there is not a local variable in a method, it's and instance variable on the CodedInputStream class so it's already on the heap, not on the stack (the problem seems to be because of leaking a ref to a by value type variable that's local to a method and thus allocated in the stack and not the heap, as I understand it).
Passing the `state` by ref here is just for optimizing, right, so that it's not copied? But I wonder, is that even necessary when the method is marked for aggresive inlining? Anyway, LGTMing, this is more to maybe avoid the nuisance of 
"chenge there if you change here" comment.

amanda-tarafa (approved):
LGTM

JamesNK (approved):

PR #12262, opened 2023-03-17
Make json_name take priority over name (fully) in C# parsing

### LINKED ISSUE #11987: C# possible incorrect implementation of CreateJsonFieldMap causes JSON data corruption during parse
Relevant code: https://github.com/protocolbuffers/protobuf/blob/6d776dfed52e36a58f318c0ba3acc93445ed710e/csharp/src/Google.Protobuf/Reflection/MessageDescriptor.cs#L117-L126

(proposed fix shown at bottom)

Problem: I believe the *intent* is that the json names should take precedence, but the *reality* is that later fields can corrupt this, in particular when names overlap. The following is arguably a pathological case, but consider:

``` protobuf
syntax = "proto3";

package EvilJson;

message Oof {
  int32 a = 1 [json_name = 'b'];
  int32 b = 2 [json_name = 'a'];
  int32 c = 3 [json_name = 'd'];
}
```

with example:

``` c#
using EvilJson;
using Google.Protobuf;

var msg = new Oof { A = 123, B = 456, C = 789 };
var json = JsonFormatter.Default.Format(msg);
Console.WriteLine(json); // { "b": 123, "a": 456, "d": 789 } as expected

var clone = Oof.Parser.ParseJson(json);

Write(msg);
Write(clone);

static void Write(Oof oof)
{
    Console.WriteLine();
    Console.WriteLine($"A={oof.A}");
    Console.WriteLine($"B={oof.B}");
    Console.WriteLine($"C={oof.C}");
}
```

which has output:

``` txt
{ "b": 123, "a": 456, "d": 789 }

A=123
B=456
C=789

A=0
B=456
C=789
```

Here field 1 claims `a` then `b`, then field 2 *also* claims `b` then `a`. This causes parse to not match format, since field 2 gets everything and field 1 gets nothing.

Proposed fix:

``` c#
foreach (var field in fields)
{
    map[field.Name] = field;
}
foreach (var field in fields)
{
    map[field.JsonName] = field;
}
```

this would ensure that `JsonName` claims always take priority over `Name` claims, which I *believe* is the intended implementation.

---

Unrelated: maybe `protoc` should emit a warning if a `JsonName` conflicts with a `Name` / `JsonName` of a **different** field on the same message, or any reserved name.

(I'm happy to contribute a PR with the proposed fix and tests)

jskeet:
Fixes #11987

(testprotos.pb and UnittestIssues.pb.cs are both generated; no need to review.)


csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
```diff
@@ -630,6 +630,18 @@ public void Wrappers_Standalone(System.Type wrapperType, object value, string ex
...
 
+        // See See https://github.com/protocolbuffers/protobuf/issues/11987
+        [Test]
+        public void JsonNamePriority()
+        {
+            // This tests both the formatter and the parser, but the issue was when parsing.
+            var original = new Issue11987Message { A = 10, B = 20, C = 30 };
+            var json = JsonFormatter.Default.Format(original);
+            AssertJson("{ 'b': 10, 'a': 20, 'd': 30 }", json);
+            var parsed = Issue11987Message.Parser.ParseJson(json);
```
amanda-tarafa, line 641: I'm curios, what happens if you try to parse "{ 'b': 10, 'a': 20, 'd': 30, 'c': 40 }"? I'm just wondering if on the map, there should only be a single entry per field with the key being JsonName if any, else Name. Or basically just have an internal EffectiveJsonName in the field descriptor?

amanda-tarafa (approved):
LGTM, with a question

csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
jskeet: Good question - will check on Monday

jskeet: Right... our draft specification says that should be rejected, but currently the later field in the JSON will overwrite the earlier field. (For fields of a message type, it'll actually merge the two values.)

So the C# parser is in strict violation of the spec in that sense... as it happens, we *need* that non-compliant behavior for one CloudEvent parsing scenario at least. (We'll need to work out what to do about either having a "strict" parser setting or change the default in a new major version at some point.)

But I think the behavior with this PR is strictly better than before :)

amanda-tarafa: Yep, that all makes sense. Thanks!

JamesNK (approved):

JamesNK:
Reminder: https://github.com/protocolbuffers/protobuf/issues/11987 effects C# and Java.

Remember to reopen the issue after this is merged, so Java has an open issue to fix it as well.

jskeet:
@JamesNK: I've raised #12290 just to keep the Java and C# cases separate.

PR #12378, opened 2023-03-31
[dotnet, csharp] Run tests for supported platforms .NET 6, .NET Framework 4.6.2

Romfos:
What's changed?
- Update some nuget packages (for test projects)
- Run unit tests for supported platforms: .NET 6 + .NET Framework 4.6.2

Notes:
.NET Framework 4.6.1 is out of support, .NET Framework 4.6.2 is the older version that is still supported
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

.NET 5 is out of support
https://devblogs.microsoft.com/dotnet/dotnet-5-end-of-support-update/

.NET Core 3.1 is out of support
https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/

Visual studio 2012 is out of support
https://devblogs.microsoft.com/visualstudio/support-ends-for-older-versions-of-visual-studio-feb2022

### DIFF
--- a/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
+++ b/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net60</TargetFrameworks>
+    <TargetFrameworks>net462;net6.0</TargetFrameworks>
     <AssemblyOriginatorKeyFile>../../keys/Google.Protobuf.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <IsPackable>False</IsPackable>
@@ -14,12 +14,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.5.0" />
     <PackageReference Include="NUnit" Version="3.13.3" />
-    <PackageReference Include="NUnit3TestAdapter" Version="4.2.1" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.2.2" />
   </ItemGroup>
 
-  <!-- Needed for the net45 build to work on Unix. See https://github.com/dotnet/designs/pull/33 -->
+  <!-- Needed for the netfx build to work on Unix. See https://github.com/dotnet/designs/pull/33 -->
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" PrivateAssets="All" />
   </ItemGroup>
--- a/.github/workflows/test_csharp.yml
+++ b/.github/workflows/test_csharp.yml
@@ -65,7 +65,7 @@ jobs:
             /bin/bash -c '
             DOTNET_CLI_TELEMETRY_OPTOUT=true
             DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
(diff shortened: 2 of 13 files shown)


jskeet:
This would be a breaking change. I'm working on writing out a policy for runtime support - in the meantime I definitely don't want to change the target frameworks for the main project.

I'd be happy with a change that *just* affected the test projects (and the change to use NET5_0_OR_GREATER is fine too) if you want to reduce your PR to that.

Romfos:
> This would be a breaking change, without us giving folks enough warning (IMO) that we're reducing the scope of target frameworks. I'm working on writing out a policy for runtime support - in the meantime I definitely don't want to change the target frameworks for the main project.
> 
> I'd be happy with a change that _just_ affected the test projects (and the change to use NET5_0_OR_GREATER is fine too) if you want to reduce your PR to that.

yes, I agree that it is a breaking change, but at the same time not sure that a lot of peoples uses platforms without support at least net standard 2.0. From my point of view this is reasonable change, because in any case no one test on windows phones, e.t.c

pr was updated

jskeet (approved):
Thanks - will ask a protobuf team member to look as well.

jskeet:
Please edit csharp/compatibility_tests/v3.0.0/test.sh as well - I *think* that's the cause of the test failures here.

deannagarcia:
For the test failures, I think you need to update this file as well: https://github.com/protocolbuffers/protobuf/blob/main/csharp/buildall.sh.

deannagarcia:
I think you also need to update our bazel files (you can just find & replace the version):

- https://github.com/protocolbuffers/protobuf/blob/580678eacd52d5d1d092f5073306f1bf6269996a/csharp/BUILD.bazel#L76
- https://github.com/protocolbuffers/protobuf/blob/580678eacd52d5d1d092f5073306f1bf6269996a/csharp/src/Google.Protobuf.Conformance/BUILD.bazel#L64

JamesNK (approved):

csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net60</TargetFrameworks>
+    <TargetFrameworks>net462;net6.0</TargetFrameworks>
```
jskeet, line 4: This change broke the GitHub action - please edit `.github/workflows/test_csharp.yml` as well to specify net6.0 instead of net60. (It's in two places.)

As far as I can see that should be the last net60 reference... we'll get there.

jskeet:
Looks like there's still an issue in a Docker image - will track it down when I get the chance.

jskeet:
(It looks like it tested a stale commit... we'll see what happens now.)

jskeet:
Ah - it looks like your fork can't modify files under .github, which does make a certain amount of sense. Will ask the protobuf team what the normal process is for this.

Romfos:
> Ah - it looks like your fork can't modify files under .github, which does make a certain amount of sense. Will ask the protobuf team what the normal process is for this.

life is hard :d

deannagarcia:
Once this passes tests I can clone it to get it submitted. I see there is still a failure that looks real: 
```
/usr/share/dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1005: Assets file '/workspace/csharp/src/Google.Protobuf.Test/obj/project.assets.json' doesn't have a target for 'net60'. Ensure that restore has run and that you have included 'net60' in the TargetFrameworks for your project. [/workspace/csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj]
```

Can you add `net60` to the target frameworks for Google.Protobuf.Test?

deannagarcia:
Got it, I'm working on getting this submitted now.

Issue #4335, opened 2018-02-25
[PHP] Merging of Timestamp field from JSON string results in wrong seconds


bitevo:
Merging of Timestamp field from JSON string results in wrong seconds. In my test, it was off by 1 sec for microsecond resolution string and even worse for nanosecond resolution string.

Test output (run on MacOS High Sierra):

```text
PHP Version: 7.1.7

Original as string : 2018-02-25T03:48:17.428086+00:00
Original seconds   : 1519530497.428086

Convert nano string to/from JSON:
Nano input string  : 2018-02-25T03:48:17.428086000Z
JS Decoded seconds : 1519530490
JS Decoded nanos   : 428086000

Convert to and from protobuf:
Pre-enc seconds   : 1519530497
Pre-enc nanos     : 428086
PB Decoded sec    : 1519530497
PB Decoded nsec   : 428086

Convert to and from JSON:
JS Decoded sec    : 1519530496
JS Decoded nsec   : 428086
```

The issue described here is about **correctness**, **not precision**. I know PHP doesn't natively support nano second, but string parsing should not be affected by that. There will be lost of precision during conversion from string to PHP `DateTime` object, but the number of seconds should be always correct.

The proto file I used in the test:

```protobuf
syntax = "proto3";

import "google/protobuf/timestamp.proto";

message Sample {
  google.protobuf.Timestamp timestamp = 1;
}
```

PHP test script (using compiled PHP class):

```php
$timestamp = new \DateTime('2018-02-25T03:48:17.428086', new \DateTimeZone('UTC'));
$gpbTimestamp = new \Google\Protobuf\Timestamp();
$gpbTimestamp->setSeconds((int) $timestamp->format('U'));
$gpbTimestamp->setNanos((int) $timestamp->format('u'));
echo "PHP Version: ", PHP_VERSION, PHP_EOL;
echo PHP_EOL;
echo "Original as string : ", $timestamp->format('Y-m-d\TH:i:s.uP'), PHP_EOL;
echo "Original seconds   : ", $timestamp->format('U.u'), PHP_EOL;
echo PHP_EOL;

echo "Convert nano string to/from JSON:", PHP_EOL;
$nanoArray = [
    'timestamp' => $timestamp->format('Y-m-d\TH:i:s.u') . '000Z',
];
$message = new Sample();
$message->mergeFromJsonString(json_encode($nanoArray));
echo "Nano input string  : ", $nanoArray['timestamp'], PHP_EOL;
[...]

PR #12396, opened 2023-04-03
Fix inconsistent timestamp json encode/decode

kindratmakc:
Protobuf php lib encodes 123_000_000 nano like this: 2000-01-01T00:00:00.**123**Z but then it gets decoded into 123 nanoseconds instead of 123_000_000.

There were issue opened some time ago that also describes this behaviour #4335


haberman (approved):

DimaFens:
Very useful fix. Need to be merged

PR #12431, opened 2023-04-09
restore pypy support broken when adding 3.11 support

tgbugs:
da973aff2adab60a9e516d3202c111dbdde1a50f did not correctly handle endif and else needed when PYPY_VERSION is set


ericsalo (approved):

ericsalo:
Thank you for pointing out this bug. I have duplicated your patch from our side (with a very minor tweak) and the fix has now been pushed to the protobuf main branch so I will close this PR.

tgbugs:
Yep, just wanted to make sure that you all saw it, no point bothering with the CLA for something this small. Thanks!

PR #12476, opened 2023-04-15
Varint: Save a branch in over-serialized case and remove inline-assembly

goldsteinn:
```
if(x & 1 == 0)
    y ^= 1 << 63

```

Is equivilent to:
```
y ^= (~x & 1) << 63
```

And the latter saves a branch at no instruction cost.

LLVM-17 currently doesn't get this (although patchset going there too
to fix that).

The inline-assembly version does save an instruction (if
macrofused). But 1-branch for 2-alu is almost certainly a worthwhile
trade.


goldsteinn:
> I think you are synced to a bad point, do you mind rebasing onto head?

Done, accidentally closed PR when updating my fork.

PR #12496, opened 2023-04-18
Update version.json to 23-dev

zhangskz:
This commit should have been auto-generated as in https://github.com/protocolbuffers/protobuf/commit/96ecd25b9c88d59113cb629505e61ebf4130f3ad, but we missed the step merging this to main from CSR when 22.x was cut.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,17 +1,17 @@
 {
     "main": {
-        "protoc_version": "22-dev",
+        "protoc_version": "23-dev",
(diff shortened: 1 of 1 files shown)


zhangskz:
Note, I left this as `"lts": false` for now, but we may want to update this to true pending BTR discussion.

deannagarcia (approved):

PR #12518, opened 2023-04-20
Fix declared dependencies for pkg-config

### LINKED ISSUE #12439: protobuf.pc pkgconfig file generated by protobuf's cmake build doesn't declare absl and utf8_range dependencies.
Starting from 22.x, protobuf now depends on absl and utf8_range, but the protobuf.pc pkgconfig file doesn't properly declare those dependencies.
As a result, the build is broken if someone installs protobuf and then wants to include it via pkgconfig (since they'd be missing absl and utf8_validity symbols when trying to link).

See https://github.com/protocolbuffers/protobuf/blob/f11079b01f075bc7c5b3286d85a8feca798c09e9/cmake/protobuf.pc.cmake

This issue breaks grpc's build-and-install "distribtest" that tries to build helloworld example with grpc and protobuf:
https://github.com/grpc/grpc/blob/24d83cda9bb1caaff9c03ace6f418e946e88c49d/examples/cpp/helloworld/Makefile#L28
(which runs pkgconfig to get a list of libraries to link against when building a project that depends on grpc++ and protobuf).

mkruskal-google:
Closes #12439


deannagarcia (approved):

darkdragon-001:
Could this be backported to v22?

oliveselow87:
```c#

```

deannagarcia:
It was already backported: https://github.com/protocolbuffers/protobuf/pull/12542

Issue #12524, opened 2023-04-21
FieldMask JSON serialization improvement proposal


antongrbin:
**What language does this apply to?**
This feature request concerns FieldMask JSON serialization.

**Describe the problem you are trying to solve.**
We are using FieldMasks as described in [https://google.aip.dev/161](https://google.aip.dev/161). We understand that the semantics described in AIP are slightly different than the original semantics described in the [field_mask.proto](https://github.com/protocolbuffers/protobuf/blob/983dac24c34f07a278d45fdfd30348775d95d12a/src/google/protobuf/field_mask.proto#L43). Some differences were discussed in https://github.com/aip-dev/google.aip.dev/issues/673 and https://github.com/aip-dev/google.aip.dev/issues/1013.

The differences in semantics are not ideal, but also not a deal-breaker since each implementation can decide on the specifics or give config options.

The wall that we hit is with [https://google.aip.dev/161#map-fields](https://google.aip.dev/161#map-fields) in REST JSON services. The current protobuf FieldMask JSON serialization tries to `camelCase` all fields in the mask and fails if uppercase letter is encountered, for example:
* [FieldMaskUtil.java](https://github.com/protocolbuffers/protobuf/blob/983dac24c34f07a278d45fdfd30348775d95d12a/java/util/src/main/java/com/google/protobuf/util/FieldMaskUtil.java#L176)
* [field_mask.py](https://github.com/protocolbuffers/protobuf/blob/983dac24c34f07a278d45fdfd30348775d95d12a/python/google/protobuf/internal/field_mask.py#L41)

This means that it's not possible to implement [https://google.aip.dev/161#map-fields](https://google.aip.dev/161#map-fields) that extends `FieldMask` semantics to traverse into map keys. For map keys that contain an uppercase letters or special characters, we get an exception during JSON serialization of `google.protobuf.FieldMask`.

**Describe the solution you'd like**
The mentioned AIP suggests a solution:
> Field masks should support string keys that contain characters that are problematic for the field mask syntax, using the backtick character.

We could extend JSON ser/de of FieldMasks to enable backtick quotation. For quoted fields, we wouldn't attempt camelCase/snake_case conversion. I found [one implementation that implements this](https://pkg.go.dev/go.chromium.org/luci/common/proto/mask).

**Describe alternatives you've considered**
Workarounds that we considered are:
* use binary serialization
* since we don't use protobuf field mask utils, we don't need to use `google.protobuf.FieldMask` but a custom similar message

[...]

PR #12551, opened 2023-04-25
[12524] FieldMask JSON serialization improvement proposal conformance tests

antongrbin:
# Motivation

Adding conformance tests to support discussion in issue https://github.com/protocolbuffers/protobuf/issues/12524.

### DIFF
--- a/conformance/binary_json_conformance_suite.cc
+++ b/conformance/binary_json_conformance_suite.cc
@@ -2948,97 +2948,97 @@ void BinaryAndJsonConformanceSuite::RunJsonTestsForFieldMask() {
   // If a path element is quoted, no camelCase conversion will be done when serializing
   // to JSON, it will be copied to JSON as-is, including the quotes.
   RunValidJsonTest("FieldMaskWithBacktickQuotes", RECOMMENDED,
-                   R"json({"optionalFieldMask": "`foo`"})json",
-                   R"(optional_field_mask: {paths: "`foo`"})");
+                   R"json({"optionalFieldMask": "mapField.`foo`"})json",
+                   R"(optional_field_mask: {paths: "map_field.`foo`"})");
 
   RunValidJsonTest("FieldMaskWithBacktickQuotesNoCamelCase", RECOMMENDED,
-                   R"json({"optionalFieldMask": "`foo_bar`"})json",
-                   R"(optional_field_mask: {paths: "`foo_bar`"})");
+                   R"json({"optionalFieldMask": "mapField.`foo_bar`"})json",
+                   R"(optional_field_mask: {paths: "map_field.`foo_bar`"})");
 
   RunValidJsonTest("FieldMaskWithBacktickQuotesTwoPaths", RECOMMENDED,
-                   R"json({"optionalFieldMask": "`foo`,`bar`"})json",
-                   R"(optional_field_mask: {paths: "`foo`" paths: "`bar`"})");
+                   R"json({"optionalFieldMask": "mapField.`foo`,mapField.`bar`"})json",
+                   R"(optional_field_mask: {paths: "map_field.`foo`" paths: "map_field.`bar`"})");
 
   RunValidJsonTest("FieldMaskWithBacktickQuotesTwoElements", RECOMMENDED,
-                   R"json({"optionalFieldMask": "`foo`.`bar`"})json",
-                   R"(optional_field_mask: {paths: "`foo`.`bar`"})");
+                   R"json({"optionalFieldMask": "mapField.`foo`.otherMapField.`bar`"})json",
+                   R"(optional_field_mask: {paths: "map_field.`foo`.other_map_field.`bar`"})");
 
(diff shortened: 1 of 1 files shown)


conformance/binary_json_conformance_suite.cc
```diff
@@ -2926,6 +2926,56 @@ void BinaryAndJsonConformanceSuite::RunJsonTestsForFieldMask() {
...
+  RunValidJsonTest("FieldMaskWithBacktickQuotes", RECOMMENDED,
+                   R"json({"optionalFieldMask": "`foo`"})json",
+                   R"(optional_field_mask: {paths: "`foo`"})");
+
+  RunValidJsonTest("FieldMaskWithBacktickQuotesNoCamelCase", RECOMMENDED,
+                   R"json({"optionalFieldMask": "`foo_bar`"})json",
+                   R"(optional_field_mask: {paths: "`foo_bar`"})");
+
+  RunValidJsonTest("FieldMaskWithBacktickQuotesTwoPaths", RECOMMENDED,
+                   R"json({"optionalFieldMask": "`foo`,`bar`"})json",
```
jskeet, line 2945: One potentially-interesting case which I don't think is directly covered yet:

```text
JSON: "`foo,bar`"
Result: { paths: "`foo,bar`" }
```

I *suspect* the backtick should only occur *either* at the start and end of a path *or* double in an escape sequence. Although I'm not sure about that right now...

So we could have a bunch of test cases for "this can't be converted to JSON" or "this JSON can't be parsed".

antongrbin (commented):
Thank you for your time!

conformance/binary_json_conformance_suite.cc
```diff
@@ -2926,6 +2926,106 @@ void BinaryAndJsonConformanceSuite::RunJsonTestsForFieldMask() {
...
+                   R"(optional_field_mask: {paths: "`foo`" paths: "`bar`"})");
+
+  RunValidJsonTest("FieldMaskWithBacktickQuotesTwoElements", RECOMMENDED,
+                   R"json({"optionalFieldMask": "`foo`.`bar`"})json",
+                   R"(optional_field_mask: {paths: "`foo`.`bar`"})");
```
jcready, line 2950: Should backtick quoted parts of a path be restricted to only occurring in odd numbered (zero-based) path segments/elements? The above could never be a valid field mask for any message since you cannot have a map of maps (e.g. `map<sting, map<string, string>>` isn't allowed).

antongrbin: Great callout, thank you! I fully agree. I fixed the unit tests so that a valid test never uses two consecutive map keys.

jcready: It might be wise to also include a few parse failure tests for these cases. Failure if the path starts with a quoted segment, failure if there are two quoted segments in a row.

antongrbin: I included such cases as well!

I was not 100% confident that a case where a path starts with a quoted segment should be a serialization failure. I think we can confidently say that that is a failure only when we know that a field mask is applied to a regular message. If a field mask for any reason wants to get applied on a map, it might be hurtful to fail it on serialization layer.

TLDR; I agree it's a failure, just not sure on which layer that failure should be surfaced. I added the tests though.

jcready: Agreed. One problem I find with AIP-161 in general is that it is effectively incompatible with existing implementations. AFAIK every existing implementation validates/parses the JSON representation by:

1. splitting on `,`
2. throwing on any encountered `_`
3. Replace every capital letter with `_` plus lowercase letter.

This string splitting strategy fundamentally cannot work with AIP-161. To be fully conformant the implementation needs to write a parser for reading JSON (throwing during JSON parsing), and a separate parser for any or all of these (throwing during serialization):

1. Programmatically creating a new FieldMask instance from a list/array of paths.
2. Adding a new string path to an existing FieldMask instance
3. Serialization of a FieldMask.

Not to mention any of the FieldMask utilities which might (depending on the implementation) require one to parse individual paths into a "tree" of path segments/elements.

The main point being that all existing implementations that would receive an AIP-161 FieldMask (with backticks and underscores and commas inside the backticks) would fail at parsing the JSON. Not to mention that all the FieldMask utilities would also be non-AIP-161-conforming.

zawarkhan1: @antongrbin , @jcready , @jskeet i went through your comments in this PR and this issue: https://github.com/protocolbuffers/protobuf/issues/12524
I am unable to do apply field mask to map example given in https://google.aip.dev/161, am i missing anything ?
i have tried all the patterns mentioned in there. Is it safe to assume that field mask cannot be applied to keys in a map or struct ?

iot2198: @antongrbin , @jcready , @jskeet i went through your comments in this PR and this issue: #12524
I am unable to do apply field mask to map example given in https://google.aip.dev/161, am i missing anything ?
i have tried all the patterns mentioned in there. Is it safe to assume that field mask cannot be applied to keys in a map or struct ?

PR #12568, opened 2023-04-26
deserialization for int64 to int in python instead of string

lightning-dabbler:
Python: deserialization for int64 to int in python instead of string


anandolee:
int64 deserialize as string according to protobuf json mapping. All protobuf languages follow this json mapping:
https://protobuf.dev/programming-guides/proto3/#json

PR #12591, opened 2023-04-27
Cherrypick fixes for 23.0-rc2 into 23.x

deannagarcia:


src/google/protobuf/descriptor.cc
```diff
@@ -2612,10 +2612,12 @@ void FileDescriptor::CopyHeadingTo(FileDescriptorProto* proto) const {
   }
 
   // TODO(liujisi): Also populate when syntax="proto2".
```
mkruskal-google, line 2614: I think you're missing https://github.com/protocolbuffers/protobuf/commit/9b218c1d9aaa8108f65cca0d65325936a8bbeac6

PR #12593, opened 2023-04-27
Cherrypick fixes for 23.0-rc2 into 23.x

deannagarcia:


src/google/protobuf/descriptor.cc
```diff
@@ -2612,10 +2612,12 @@ void FileDescriptor::CopyHeadingTo(FileDescriptorProto* proto) const {
   }
```
mkruskal-google, line 2613: You're missing https://github.com/protocolbuffers/protobuf/commit/9b218c1d9aaa8108f65cca0d65325936a8bbeac6 I think

mkruskal-google:
Should we just pull in every change on main to be safe?  It's only a day, and I'm worried we might be missing dependent commits

src/google/protobuf/descriptor.cc
deannagarcia: Pulled it in, thanks for the catch!

deannagarcia:
> Should we just pull in every change on main to be safe? It's only a day, and I'm worried we might be missing dependent commits

Discussed offline, tools work better with these cherrypicks so we are going to stick with this strategy.

fowles (approved):

mkruskal-google (approved):

PR #12707, opened 2023-05-08
[PHP] allow encoding enums as integers for json

brettmc:
The php extension already had an implementation for encoding enums as integers, but it was not exposed, so add a bitmask param to `Message::serializeToJsonString()` to enable it as well as `preserve_proto_field_names`.

Converted the existing boolean parameter to preferentially be an integer (with boolean handling for BC), which accepts a bitmask of `Google\Protobuf\PrintOptions::*`, eg `PrintOptions::ALWAYS_PRINT_ENUMS_AS_INTS | PrintOptions::PRESERVE_PROTO_FIELD_NAMES`

`PrintOptions` class name and constant names were chosen to align with protobuf-cpp's implementation.

Implemented `preserve_proto_fieldnames` in the native version, to match the existing implementation in the extension (with tests).


brettmc:
Can I get any update on how this is progressing? I believe that it's been moved into google's internal systems and possibly being looked at there, but from the outside there's no sign of any activity.

brettmc:
Hi. This is still active, and awaiting work/feedback/something from a maintainer.

brettmc:
Still active. Please notice me.

brettmc:
Still active/required/wanted. Pretty please notice me.

brettmc:
@anandolee can I get an update on where this PR is at, since there's no visibility of what's happening behind the scenes? Thanks!

googleberg:
@bshaffer can you please take a look?

zhangskz:
Apologies for the delay in getting this reviewed! Looks like most tests are failing -- could you rebase to latest main?

zhangskz:
@bshaffer Friendly ping, are you able to take a look at this?

brettmc:
Still required, please don't close.

php/ext/google/protobuf/message.c
```diff
@@ -814,16 +814,21 @@ PHP_METHOD(Message, serializeToJsonString) {
   int options = 0;
   char buf[1024];
   zend_bool preserve_proto_fieldnames = false;
+  zend_bool format_enums_as_integers = false;
   upb_Status status;
 
-  if (zend_parse_parameters(ZEND_NUM_ARGS(), "|b",
-                            &preserve_proto_fieldnames) == FAILURE) {
+  if (zend_parse_parameters(ZEND_NUM_ARGS(), "|bb",
```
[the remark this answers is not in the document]
bshaffer: I would definitely argue that a bitmask is the optimal way to go here. As for what would be safe, I am not sure. From very basic testing, passing in the bool `true` and performing bitmask operations on it performs the same (in native PHP) as passing the integer `1`:

```php
var_dump(true & 1);
// int(1)
var_dump(true & 2);
// int(0)
```

I've also confirmed that with current implementation of the C-extension, passing an int instead of a boolean does not throw an error:

```php
var_dump($m->serializeToJsonString(1));
// string(2) "{}"
```

However, we would still need to test:

  1. the first example behaves the same in the C-extension (e.g `$m->serializeToJsonString(true)` does not result in a breaking change)
  2. The second example sets the `preserve_proto_fieldnames` flag as expected

bshaffer (commented):
I think this would be a great improvement for this library!

Could you tell us a little bit about your use-case? I imagine that there's a mismatch between how the server interprets enums and how the client is sending them? Or vice-versa?

Either way I agree exposing this behavior is very important. Thanks for taking the time to craft this PR.

php/ext/google/protobuf/message.c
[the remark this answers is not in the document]
brettmc: Thanks for the initial review. I will get started on implementing this as a bitmask, with added tests for backwards compatibility.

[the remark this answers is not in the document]
brettmc: @bshaffer I've implemented the options as a bitmask:
- Added a BC test to ensure that `preserve_proto_fieldnames` is set as expected. In the process, I discovered that the option to preserve proto fieldnames in json output actually *is* implemented in the extension, but not in native. I've gone ahead and implemented it in native so that they behave the same, but FYI this would be a BC break in the native version: `$m->serializeToJsonString(true)`.
- I didn't find any existing place to add public-facing constants outside of `Internal`, so I created `Google\Protobuf\Options` to hold the new ones I've added
- Added some boolean checks in both versions of `serializeToJsonString` to handle BC
- I see that there is a third possible option to `serializeToJsonString`, which is "emit default values". I've added a constant for this but not implemented.

bshaffer (commented):
looks like some of the tests are failing because of the unexpected new file, but I can fix those.

php/src/Google/Protobuf/Internal/CodedOutputStream.php
```diff
@@ -11,25 +11,36 @@
 
 class CodedOutputStream
 {
+    /* @see ext/php-upb.h */
+    const JSON_ENCODE_EMIT_DEFAULTS = 1 << 0;
+    const JSON_ENCODE_FORMAT_ENUMS_AS_INTEGERS = 1 << 1;
+    const JSON_ENCODE_PRESERVE_PROTO_FILENAMES = 1 << 2; //@todo should be JSON_ENCODE_PRESERVE_PROTO_FIELDNAMES
```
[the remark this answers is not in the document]
brettmc: I think I left that there as a question, and you've given me the answer. Since it's an `Internal` field, it seems safe enough to fix it so that it matches what it actually does.

php/tests/EncodeDecodeTest.php
```diff
@@ -1619,4 +1620,64 @@ public function testEmptyAnySerialization()
...
+    public function testJsonStringBCPreserveProto(bool $value, $expected)
+    {
+        $m = new TestMessage();
+
+        $m->setOneofEnum(TestEnum::ONE);
+        $data = $m->serializeToJsonString($value);
+        $this->assertSame($expected, $data);
+    }
+
+    public static function preserveProtoFilenamesProvider(): array
```
bshaffer, line 1676: should be 
```suggestion
    public static function preserveProtoFieldnamesProvider(): array
```

php/src/Google/Protobuf/Internal/Message.php
```diff
@@ -1898,23 +1918,31 @@ private function fieldJsonByteSize($field)
...
             }
         } elseif ($this->existField($field) || GPBUtil::hasJsonValue($this)) {
             if (!GPBUtil::hasSpecialJsonMapping($this)) {
                 $size += 3;                              // size for "\"\":".
-                $size += strlen($field->getJsonName());  // size for field name
+                if ($options & Options::JSON_ENCODE_PRESERVE_PROTO_FIELDNAMES) {
+                    $size += strlen($field->getName());
+                } else {
+                    $size += strlen($field->getJsonName());
+                } // size for field name
```
bshaffer, line 1941: I might be missing something - why is this behavior changing? I thought we were only changing the behavior of allowing enums as ints, but this is doing something different. Is this resolving a separate bug?

bshaffer (commented):
One thing we should check is if this is something we'd rather handle behind a feature flag. I'll ask @haberman

php/src/Google/Protobuf/Internal/Message.php
brettmc: This is a separate bug. When I added a test for BC of current behaviour, I discovered that the existing boolean parameter (preserve proto fieldnames) works in the extension, and does not work in the native. This aligns the two.

php/src/Google/Protobuf/Internal/CodedOutputStream.php
```diff
@@ -11,25 +11,36 @@
 
 class CodedOutputStream
 {
+    /* @see ext/php-upb.h */
+    const JSON_ENCODE_EMIT_DEFAULTS = 1 << 0;
+    const JSON_ENCODE_FORMAT_ENUMS_AS_INTEGERS = 1 << 1;
+    const JSON_ENCODE_PRESERVE_PROTO_FIELD_NAMES = 1 << 2;
```
bshaffer, line 17: It seems like since the options are here, we don't need to add the `Options` class. Was "Options" added because this is in the `Internal` namespace? If so, I think it's okay to keep it as-is (it's not a bad idea to consider these options "internal" for now anyway, in case they behave unexpectedly)

bshaffer: Correction: By "as-is" I meant that we should use the constants defined in `CodedOutputStream` instead of creating a new `Options` class.

php/src/Google/Protobuf/Internal/Message.php
brettmc: I've made a patch to do this. BUT, when using the extension there is no user-facing constant to represent the flags (`Internal\CodedOutputStream` does not exist in the extension). It should have the same FQN in both. If we want to avoid adding new files, perhaps exposing constants in a class like `Message` instead?

php/src/Google/Protobuf/Internal/CodedOutputStream.php
brettmc: I've made a patch to do this. BUT, when using the extension there is no user-facing constant to represent the flags (`Internal\CodedOutputStream` does not exist in the extension). It should have the same FQN in both. If we want to avoid adding new files, perhaps exposing constants in a class like `Message` instead?

bshaffer: Thank you, I understand now why it was done this way! And I agree it should be consistent between extension and native library when possible. 

I will propose this question to the protobuf team and also try to find out how flags like this are implemented in other languages. I think exposing an `Options` class is fine, but I'd like to check first to see if there isn't a precedent already.

Thanks again for your work on this!

bshaffer: After chatting with the rest of the Protobuf team, I've been pointed to the fact that  C++ uses a [PrintOptions](https://github.com/protocolbuffers/protobuf/blob/73a38534796565ba7f9cdc6055f357985936613f/src/google/protobuf/json/json.h#L38-L55) struct. Since PHP doesn't really have structs, I think we can just rename what you have to `PrintOptions` (to be at least a little consistent), and rename the constants to match as well, for example:

```php
class PrintOptions
{
    const PRESERVE_PROTO_FIELD_NAMES = 1 << 0;
    const ALWAYS_PRINT_ENUMS_AS_INTS = 1 << 1;
}
```

I'd like to keep `PRESERVE_PROTO_FIELD_NAMES` as `1`, since that matches existing behavior (`true` gets cast to `1`, so it'll be backwards compatible). And I don't think we really need a constant for the defaults (which would be `0`), since that'll be the default anyway.

What do you think?

php/ext/google/protobuf/message.c
```diff
@@ -754,16 +755,25 @@ PHP_METHOD(Message, serializeToJsonString) {
...
 
-  if (zend_parse_parameters(ZEND_NUM_ARGS(), "|b",
-                            &preserve_proto_fieldnames) == FAILURE) {
+  if (zend_parse_parameters(ZEND_NUM_ARGS(), "|z",
+                            &flags) == FAILURE) {
     return;
   }
 
-  if (preserve_proto_fieldnames) {
+  if (flags == NULL || Z_TYPE_P(flags) == IS_FALSE) {
```
brettmc, line 766: FYI I've left this boolean-handling code in the C code, as `Z_LVAL_P(bool)` causes a segfault (and the param needs to be a zval to accept both int&bool inputs)

php/src/Google/Protobuf/Internal/CodedOutputStream.php
brettmc: Makes sense, and done:
- renamed to `PrintOptions` and aligned consts with c++
- BC of `true`/`1` works in native, but needed some handing in extension since some of the zval-handling macros are more picky and can cause a segfault if operating on the wrong type
- removed the default constant

php/ext/google/protobuf/message.c
bshaffer: I don't know C++ well, but I added a bunch more test cases to ensure that this functions as expected and aligns with the previous behavior (I added tests for boolean `true`, `false`, and calling without parameters). So I feel confident that this behavior is being preserved, and it should be okay.

Thanks for fixing the segfault :)

bshaffer (approved):

bshaffer (approved):

Issue #12714, opened 2023-05-08
PHP: use __debugInfo in Message to fix endless var_dump


bshaffer:
## Debugging a Protobuf Message Object
Calling `var_dump` or `print_r` on a protobuf message with the Protobuf C extension enabled results very limited, not very useful output:

```
// Calling var_dump on a Protobuf message (extension enabled):
php > var_dump(new Google\Protobuf\Timestamp());
object(Google\Protobuf\Timestamp)#1 (0) {
}

// Calling print_r on a Protobuf message (extension enabled):
php > var_dump(new Google\Protobuf\Timestamp());
Google\Protobuf\Timestamp Object
```

When using the Protobuf native library, calls to `var_dump` and `print_r` are hard to read and filled with noise:

```
// Calling var_dump on a Protobuf message (extension disabled):
php > var_dump(new Google\Protobuf\Timestamp());
object(Google\Protobuf\Timestamp)#2 (4) {
  ["desc":"Google\Protobuf\Internal\Message":private]=>
  object(Google\Protobuf\Internal\Descriptor)#424 (13) {
    ["full_name":"Google\Protobuf\Internal\Descriptor":private]=>
    string(25) "google.protobuf.Timestamp"
    ["field":"Google\Protobuf\Internal\Descriptor":private]=>
    array(2) {
      [1]=>
      object(Google\Protobuf\Internal\FieldDescriptor)#428 (14) {
        ["name":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        string(7) "seconds"
        ["json_name":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        string(7) "seconds"
        ["setter":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        string(10) "setSeconds"
        ["getter":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        string(10) "getSeconds"
        ["number":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        int(1)
        ["label":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        int(1)
        ["type":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        int(3)
        ["message_type":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        NULL
        ["enum_type":"Google\Protobuf\Internal\FieldDescriptor":private]=>
        NULL
        ["packed":"Google\Protobuf\Internal\FieldDescriptor":private]=>
[...]

PR #12718, opened 2023-05-08
feat: better debug info for PHP messages and repeated fields

bshaffer:
addresses https://github.com/protocolbuffers/protobuf/issues/12714 by dumping more concise debug info for protobuf messages and repeated fields via the `serializeToJsonString` function. Additionally, message types which serialize into something other than an array (e.g. `Google\Protobuf\Value`, `Google\Protobuf\Timestamp`, etc) are handled in a special way to make their output consistent with other messages.

### DIFF
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -2037,4 +2037,56 @@ class Message
         }
         return $size;
     }
+
+    public function __debugInfo()
+    {
+        if (is_a($this, 'Google\Protobuf\FieldMask')) {
+            return ['paths' => $this->getPaths()->__debugInfo()];
+        }
+
+        if (is_a($this, 'Google\Protobuf\Value')) {
+            switch ($this->getKind()) {
+                case 'null_value':
+                    return ['nullValue' => $this->getNullValue()];
+                case 'number_value':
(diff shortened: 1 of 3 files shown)


bshaffer:
Thanks @haberman. This is a great question. As [__debugInfo](https://www.php.net/manual/en/language.oop5.magic.php#object.debuginfo) has "debug" in the name, I feel like it's self-evident that this behavior shouldn't be relied upon in production systems (and probably isn't being relied upon currently, given the output). However, if a major version bump is coming already, I think it'd make sense to have this be a part of it, for continuity. 

So, while not strictly necessary to put this behind a major version bump, it would be appropriate to do so.

bshaffer:
Ping on this @haberman ! I think this would be a very great improvement for our customers!

shawnlindstrom:
This would have been a huge help when implementing v2 for Cloud Speech a couple weeks ago. @bshaffer thanks for doing this and I hope it gets merged soon.

haberman (approved):

zhangskz (approved):

zhangskz:
(Note, another LGTM was needed after rebase, since review needs to be at head commit to be imported)

bshaffer:
@zhangskz I submitted a fix for the failing tests, could you approve them to run again?

fowles (approved):

PR #12720, opened 2023-05-08
fix(cmake): use `find_dependency()` in config files

coryan:
`find_dependency()` is designed to be used in `*-config.cmake` files. It is a thin wrapper around `find_package()` that propagates any `QUIET` or `REQUIRED` parameters:

  https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html

In general, the error messages from `find_dependency()` are easier to grok.  Without them it may be hard to find out why a `find_package(Protobuf CONFIG REQUIRED)` failed, or to understand why `find_package(Protobuf CONFIG QUIET)` is generating noise Protobuf's dependencies.


acozzette (approved):

coryan:
I need to start this over. Closing for now.

PR #12733, opened 2023-05-09
CMake: Use the correct jsoncpp target name in static builds

semlanik:
When building protobuf with BUILD_SHARED_LIBS disabled, conformance_test_runner should link jsoncpp_static but not jsoncpp_lib.


mkruskal-google (approved):

semlanik:
Any ideas what went wrong?

PR #12762, opened 2023-05-11
fix: avoid warnings with MSVC

coryan:
In both cases a `size_t` was being converted to a `uint32_t`. These headers are used from application code, or at least from generated code, and the application may be compiling with more warnings enabled than normal.


src/google/protobuf/string_block.h
```diff
@@ -145,7 +145,8 @@ inline StringBlock* StringBlock::Emplace(void* p, size_t n, StringBlock* next) {
   ABSL_DCHECK_EQ(n, NextSize(next));
   uint32_t doubled = static_cast<uint32_t>(n) * 2;
```
fowles, line 146: rather than having the `static_cast` twice in this function, can you pull a local?

fowles (approved):

PR #12778, opened 2023-05-12
Update docs for Any type URLs

akshayjshah:
The documentation for the well-known `Any` type spends quite a bit of time describing the behavior of a "type server," which is effectively a schema registry. Readers can easily walk away with the impression that type servers, and the associated `Type`, `Field`, `Enum`, etc. messages, are widely used within Google. From talking to @fowles, this impression is incorrect.

This PR preserves the original documentation, but adds a note that no widely-used type servers (as described on `Any`) exist within Google or in open-source.


fowles (approved):

acozzette:
Thank you for the documentation fix. Unfortunately there is an internal compliance reason that prevents us from being able to accept contributions to this file and other well-known types.  However, I will independently submit a change to this file to clarify that the type server idea has not been widely implemented or used.

PR #12812, opened 2023-05-15
Resolve #12809: Pass key-value pairs into Struct constructor in python

### LINKED ISSUE #12809: Pass key-value pairs into Struct constructor in python
python

When we initialise a `Struct` and want to set key-value pair on the struct, there are two steps:
```py
struct = Struct()
struct.update(key1='value-1', key2=2)
```

Alternatively, we can also use `json_format.ParseDict`.

Hopefully, we can do the following:
```py
struct = Struct(fields={'key1': 'value-1, 'key2: 2})
# or 
struct = Struct({'key1': 'value-1, 'key2: 2})
```

**Additional context**

TommyDew42:
See #12809

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -476,6 +476,10 @@ class Struct(object):
 
   __slots__ = ()
 
+  def __init__(self, dictionary=None):
+    if dictionary:
+      self.update(dictionary)
+
   def __getitem__(self, key):
     return _GetStructValue(self.fields[key])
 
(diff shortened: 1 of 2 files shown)


TommyDew42:
Realised it's actually more complicated than just adding a `__init__` method. The `__init__` method is defined via [_AddInitMethod](https://github.com/TommyDew42/protobuf/blob/e6adc6d3cd80496f3393b0475ffaa52fa71e9365/python/google/protobuf/internal/python_message.py#L464) so we need to do sth there too.

PR #12828, opened 2023-05-16
Add support for descriptor options in ruby interface

jsteinberg:
Add support for descriptor options in Ruby:

spawned off of [discussion here](https://github.com/protocolbuffers/protobuf/issues/1198#issuecomment-1548160127)

Notes:  
- `EnumValueOptions` / `ServiceOptions` / `MethodOptions` were not included.  There are not ruby interfaces for `EnumValueDescriptorProto` / `ServiceDescriptorProto` / `MethodDescriptorProto`.  If we want to add those first I could look into that.
- In the conversation on the original issue, @haberman commented `# We actually need to freeze recursively.`.  I did not implement that yet and am looking for guidance.  
  - Should that be the default behavior in the `Message_freeze` method?
  - What all field types need to be frozen (`SubMessage` / `Repeated` / `Map` ?)

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -907,6 +907,33 @@ static VALUE Message_freeze(VALUE _self) {
 
 /*
  * call-seq:
+ *     Message.deep_freeze => self
+ *
+ * Deeep freezes the message object recursively.
+ */
+static VALUE Message_deep_freeze(VALUE _self) {
+  Message* self = ruby_to_Message(_self);
+  if (!RB_OBJ_FROZEN(_self)) {
+    Message_freeze(_self);
+
+    int n = upb_MessageDef_FieldCount(self->msgdef);
+    for (int i = 0; i < n; i++) {
+      const upb_FieldDef* f = upb_MessageDef_Field(self->msgdef, i);
+      VALUE field = Message_getfield(_self, f);
+
+      if (upb_FieldDef_IsMap(f) || upb_FieldDef_IsRepeated(f)) {
+        rb_funcall(field, rb_intern("freeze"), 0);
+      } else if (upb_FieldDef_IsSubMessage(f)) {
+        Message_deep_freeze(field);
+      }
+    }
+  }
+
+  return _self;
+}
+
+/*
+ * call-seq:
  *     Message.[](index) => value
  *
  * Accesses a field's value by field name. The provided field name should be a
@@ -1407,6 +1434,7 @@ static void Message_define_class(VALUE klass) {
   rb_define_method(klass, "==", Message_eq, 1);
   rb_define_method(klass, "eql?", Message_eq, 1);
   rb_define_method(klass, "freeze", Message_freeze, 0);
+  rb_define_private_method(klass, "deep_freeze", Message_deep_freeze, 0);
   rb_define_method(klass, "hash", Message_hash, 0);
   rb_define_method(klass, "to_h", Message_to_h, 0);
   rb_define_method(klass, "inspect", Message_inspect, 0);
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -909,7 +909,7 @@ static VALUE Message_freeze(VALUE _self) {
  * call-seq:
  *     Message.internal_deep_freeze => self
  *
- * Deeep freezes the message object recursively.
+ * Deep freezes the message object recursively.
  * Internal use only.
  */
 static VALUE Message_internal_deep_freeze(VALUE _self) {
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -907,11 +907,12 @@ static VALUE Message_freeze(VALUE _self) {
 
 /*
  * call-seq:
- *     Message.deep_freeze => self
+ *     Message.internal_deep_freeze => self
  *
  * Deeep freezes the message object recursively.
+ * Internal use only.
  */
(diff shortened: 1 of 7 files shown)


ruby/lib/google/protobuf/descriptor_dsl.rb
```diff
@@ -451,6 +451,10 @@ def value(name, number)
         end
       end
 
+      def self.decode_options(klass, serialized_options)
+        options = klass.decode(serialized_options)
+        options.freeze
```
haberman, line 456: We want this freeze to be recursive. Otherwise a user could write something like:

```ruby
# This works, because only my_descriptor.options itself is frozen.
my_descriptor.options.some_message_options.field = 5
```

We'll want to write a loop that visits all fields and freezes them all. I'm not sure how easy this will be to do in Ruby. We may want to do it in C.

haberman (changes requested):

ruby/lib/google/protobuf/descriptor_dsl.rb
jsteinberg: Thank you for the review.  I've addressed the first 2 concerns (I _think_).

I have a question about the recursive freeze.  The [Message class already has a `freeze` method](https://github.com/protocolbuffers/protobuf/blob/main/ruby/ext/google/protobuf_c/message.c#L892-L906) that I am calling here.  Should that method be updated to be recursive or should I create a separate `deep_freeze` method for this use case?

Additionally, for the recursive freeze, should all fields also be pinned to the same arena?

haberman: `Message#freeze` follows the general Ruby semantics of only doing a shallow freeze. I don't think we should change that.

I think we could add `deep_freeze` as an internal-only API, but I don't think we should make it public. We should keep new public APIs to a minimum, since each one will need to be reviewed and scrutinized.

ruby/ext/google/protobuf_c/message.c
```diff
@@ -905,6 +905,33 @@ static VALUE Message_freeze(VALUE _self) {
...
+  if (!RB_OBJ_FROZEN(_self)) {
+    Message_freeze(_self);
+
+    int n = upb_MessageDef_FieldCount(self->msgdef);
+    for (int i = 0; i < n; i++) {
+      const upb_FieldDef* f = upb_MessageDef_Field(self->msgdef, i);
+      VALUE field = Message_getfield(_self, f);
+
+      if (upb_FieldDef_IsMap(f) || upb_FieldDef_IsRepeated(f)) {
+        rb_funcall(field, rb_intern("freeze"), 0);
```
haberman, line 925: Repeated fields and maps can contain messages, so we will need them to support `deep_freeze` also.

ruby/ext/google/protobuf_c/message.c
```diff
@@ -905,6 +905,33 @@ static VALUE Message_freeze(VALUE _self) {
   return _self;
 }
 
+/*
+ * call-seq:
+ *     Message.deep_freeze => self
+ *
+ * Deeep freezes the message object recursively.
+ */
+static VALUE Message_deep_freeze(VALUE _self) {
```
haberman, line 914: Let's call it `internal_deep_freeze` just so it's unambiguous that it's intended to be internal-only.

It doesn't make it harder to call, but if someone does call it they will know that it's an API that could change.

haberman (changes requested):

bitwise-aiden:
👋🏻 Hey @jsteinberg, is there anything I can help do to get this going again?

jsteinberg:
Yes, I've unfortunately been busy at work so haven't gotten back to this.  It's definitely on my backlog (along with adding extension support to ruby after this).

So, I'm not sure what is needed to move forward here.  I got a little bit of time this afternoon so happy to make any requested changes / rebase this off of main.

jsteinberg:
It looks like I need to figure out how to make this work for JRuby.  Its been a while since I've done anything with that.

bitwise-aiden:
Oh interesting, there isn't an existing freeze method implemented in the Java version. Will be a bit more involved than just mirroring it like I first thought.

JasonLunn:
If you have questions about JRuby I'm happy to help.

haberman:
@JasonLunn is it true that JRuby Protobuf doesn't support `#freeze` right now? If so, what would it take to add that.

jsteinberg:
> @JasonLunn is it true that JRuby Protobuf doesn't support `#freeze` right now? If so, what would it take to add that.

Sorry, I didn't end up getting around to this and then had some vacation around the holiday.  I essentially zero JRuby knowledge and since the CI tests were not enabled for this branch, I didn't realize I had omitted the necessary changes.

I can try to get JRuby running locally so I can run the tests and try to figure out what is missing, but if someone else with much more knowledge wants to do it then I'm not going to stop you.

I have limited time this week but I will try to at least start making progress on this.

jsteinberg:
It _looks_ like I need to implement `serialized_options` for the JRuby classes:

```java
NameError: undefined local variable or method `serialized_options' for #<Google::Protobuf::FileDescriptor:0x644947ee>
```

Additionally, I need to implement the `internal_deep_freeze` method (which makes sense)

```java
NoMethodError: undefined method `internal_deep_freeze' for #<BasicTest::TestDeprecatedMessage:0x65259c53>
```

JasonLunn:
> @JasonLunn is it true that JRuby Protobuf doesn't support `#freeze` right now? If so, what would it take to add that.

JRuby as a language supports `#freeze` and there are tests for Protobuf-specific freeze behaviors in `common_tests.rb` that pass under JRuby.

@jsteinberg - feel free to ping me if you have any JRuby-specific questions.

Issue #12830, opened 2023-05-16
Protobuf address sanitizer settings has to match user app


jiulongw:
**What version of protobuf and what language are you using?**
Version: 23.0 
Language: C++

**What operating system (Linux, Windows, ...) and version?**
- Ubuntu 22.04
- macOS 13.3

**What runtime / compiler are you using (e.g., python version or gcc version)**
- clang 16

**What did you do?**
1. Build protobuf static library, using default CMake settings (ASAN not turned on).
2. Build user app with ASAN turned on, link library produced in step 1.
3. Run basic protobuf serialization and de-serialization.

**What did you expect to see**
No crash

**What did you see instead?**
```
==28429==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000004 (pc 0x000100e6db3c bp 0x00016f034790 sp 0x00016f034770 T0)
==28429==The signal is caused by a READ memory access.
==28429==Hint: address points to the zero page.
==28429==WARNING: Symbolizer buffer too small
    #0 0x100e6db3c in std::__1::__optional_storage_base<int, false>::has_value[abi:v15006]() const optional:344
    #1 0x100e6d6cc in google::protobuf::internal::EpsCopyInputStream::LimitToken::token() && parse_context.h:156
    #2 0x100e6395c in google::protobuf::internal::EpsCopyInputStream::PopLimit(google::protobuf::internal::EpsCopyInputStream::LimitToken) parse_context.h:196
    #3 0x107f9dfe0 in google::protobuf::internal::ParseContext::ParseMessage(google::protobuf::MessageLite*, char const*) parse_context.cc:334
```

**Anything else we should know about your project / environment**
I would like to know if this is by design.  The clear solution is to build two version of libraries: one with ASAN turned on and one without.  However, this would be a mess to manage when there are different combination of different sanitizers needed...

The problem is there is a conditional compiler settings here: https://github.com/protocolbuffers/protobuf/blob/03501ccb49c8a4ed6aaafee32743d906e5b16b9f/src/google/protobuf/parse_context.h#L136-L179

This causes user app see a different version of the `LimitToken` class than the compiled library, which is the exact reason of the crash.

A possible fix: keep `LimitToken` the same memory structure regardless of ASAN config.  Move the conditional logic inside the methods of `LimitToken`.  However, I don't know if this is a common practice that is used over the entire repo that are too hard to fix them all.

jiulongw:
[...]

PR #12831, opened 2023-05-16
fix: keep same ABI in asan and non-asan builds

jiulongw:
fixes: #12830

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -138,19 +138,21 @@ class PROTOBUF_EXPORT EpsCopyInputStream {
   //  - Every non-empty token is moved from and consumed.
   class LimitToken {
    public:
-    LimitToken() { PROTOBUF_POISON_MEMORY_REGION(this, sizeof(LimitToken)); }
-    explicit LimitToken(int token) : token_(token) { PROTOBUF_UNPOISON_MEMORY_REGION(this, sizeof(LimitToken)); }
+    LimitToken() { PROTOBUF_POISON_MEMORY_REGION(&token_, sizeof(token_)); }
+    explicit LimitToken(int token) : token_(token) {
+      PROTOBUF_UNPOISON_MEMORY_REGION(&token_, sizeof(token_));
+    }
     LimitToken(LimitToken&& other) { *this = std::move(other); }
     LimitToken& operator=(LimitToken&& other) {
-      PROTOBUF_UNPOISON_MEMORY_REGION(this, sizeof(LimitToken));
+      PROTOBUF_UNPOISON_MEMORY_REGION(&token_, sizeof(token_));
       token_ = other.token_;
-      PROTOBUF_POISON_MEMORY_REGION(&other, sizeof(LimitToken));
+      PROTOBUF_POISON_MEMORY_REGION(&other.token_, sizeof(token_));
       return *this;
     }
 
     ~LimitToken() {
 #ifdef ADDRESS_SANITIZER
-      ABSL_CHECK(__asan_address_is_poisoned(this));
+      ABSL_CHECK(__asan_address_is_poisoned(&token_));
 #endif
     }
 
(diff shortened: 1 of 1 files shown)


src/google/protobuf/parse_context.h
```diff
@@ -160,23 +159,6 @@ class PROTOBUF_EXPORT EpsCopyInputStream {
    private:
     absl::optional<int> token_;
   };
-#else
```
fowles, line 163: you should switch to this branch of the else

src/google/protobuf/parse_context.h
```diff
@@ -160,23 +159,6 @@ class PROTOBUF_EXPORT EpsCopyInputStream {
...
-  class LimitToken {
-   public:
-    LimitToken() = default;
-    explicit LimitToken(int token) : token_(token) {}
-    LimitToken(LimitToken&&) = default;
-    LimitToken& operator=(LimitToken&&) = default;
```
fowles, line 169: and then add poisoning in the move operations

src/google/protobuf/parse_context.h
```diff
@@ -160,23 +159,6 @@ class PROTOBUF_EXPORT EpsCopyInputStream {
...
-
-    LimitToken(const LimitToken&) = delete;
-    LimitToken& operator=(const LimitToken&) = delete;
-
-    int token() const { return token_; }
```
fowles, line 174: and switch this to be `&&` qualified instead of `const`, and add poisoning here as well

fowles (changes requested):

jiulongw:
> The crux of this change is to modify `parse_context.h` so that it has the same ABI in asan and non-asan builds. Unfortunately, the affected methods are extremely performance sensitive (which is exactly why we have this branch). I would accept a variation of this CL that uses our poisoning macros and leaves the actual ABI as an int.
> 
> https://github.com/protocolbuffers/protobuf/blob/e285d3e307512f24576b56d18cb5a5b655e2f393/src/google/protobuf/port_def.inc#L129-L130

Thanks for the review!  I've updated the PR and made sure tests passed with or without ASAN.

src/google/protobuf/parse_context.h
```diff
@@ -133,50 +133,43 @@ class PROTOBUF_EXPORT EpsCopyInputStream {
...
-  // In sanitizer mode we use an optional<int> to guarantee that:
+  // In sanitizer mode we use memory poisoning to guarantee that:
   //  - We do not read an uninitialized token.
   //  - Every non-empty token is moved from and consumed.
   class LimitToken {
    public:
-    LimitToken() = default;
-    explicit LimitToken(int token) : token_(token) {}
+    LimitToken() { PROTOBUF_POISON_MEMORY_REGION(this, sizeof(LimitToken)); }
+    explicit LimitToken(int token) : token_(token) { PROTOBUF_UNPOISON_MEMORY_REGION(this, sizeof(LimitToken)); }
```
fowles, line 142: Can you mark this as poisoning only the memory of `&token_` and not `this`.  I am pretty sure they are equivalent long run, but it feels better to pick things more carefully because `this` would include vtables if we ever add them

src/google/protobuf/parse_context.h
```diff
@@ -133,50 +133,43 @@ class PROTOBUF_EXPORT EpsCopyInputStream {
...
 
     int token() && {
-      ABSL_CHECK(token_.has_value());
-      return *std::exchange(token_, absl::nullopt);
-    }
-
-   private:
-    absl::optional<int> token_;
-  };
+#ifdef ADDRESS_SANITIZER
```
fowles, line 161: you don't need this `ifdef`, you can implement this directly as:

```
int t = token_;
POISON(&token, sizeof(token_));
return t;
```

fowles (changes requested):

fowles (approved):

fowles (approved):

fowles:
Thanks for the contribution.  If you want this to show up in the next 23.x release, feel free to send a PR cherrypicking it back.  If you are happy to wait for 24.x, you are good to go.

PR #12891, opened 2023-05-24
CMake: Abseil targets referenced with imported namespace

johnwparent:
Abseil targets from an external abseil are properly exported with namespaces. CMake references to those targets should include the namespace or operations such as `target_link_librarie`s will not propagate includes/etc.

Current behavior results in a build error with MSVC + CMake. 

CMake doc for reference: https://cmake.org/cmake/help/latest/policy/CMP0028.html

An example of the failure without this change:
```
FAILED: CMakeFiles/libprotoc.dir/src/google/protobuf/compiler/code_generator.cc.obj
C:\PROGRA~2\MICROS~1\2019\BUILDT~1\VC\Tools\MSVC\1429~1.301\bin\Hostx64\x64\cl.exe  /nologo /TP -DGOOGLE_PROTOBUF_CMAKE_BUILD -DHAVE_ZLIB -DLIBPROTOC_EXPORTS -DPROTOBUF_USE_DLLS -Dlibprotoc_EXPORTS -IC:\spack_win\spack\opt\spack\windows-windows10.0.19041-AMD64\msvc-19.29.30145\zlib-1.2.13-uefnjquewrfzbyoy3p4dqguo4txb4wq7\include -IC:\dev-builds\protobuf\proto-3.22\build -IC:\dev-builds\protobuf\proto-3.22\src /DWIN32 /D_WINDOWS /W3 /GR /EHsc /Zi /Ob0 /Od /RTC1 -MDd /MP /utf-8 /wd4065 /wd4146 /wd4244 /wd4251 /wd4267 /wd4305 /wd4307 /wd4309 /wd4334 /wd4355 /wd4506 /wd4800 /wd4996 /bigobj -std:c++14 /showIncludes /FoCMakeFiles\libprotoc.dir\src\google\protobuf\compiler\code_generator.cc.obj /FdCMakeFiles\libprotoc.dir\ /FS -c C:\dev-builds\protobuf\proto-3.22\src\google\protobuf\compiler\code_generator.cc
C:\dev-builds\protobuf\proto-3.22\src\google/protobuf/compiler/code_generator.h(45): fatal error C1083: Cannot open include file: 'absl/strings/string_view.h': No such file or directory
[85/204] Building CXX object CMakeFiles\libprotoc.dir\src\google\protobuf\compiler\command_line_interface.cc.obj
```

### DIFF
--- a/cmake/abseil-cpp.cmake
+++ b/cmake/abseil-cpp.cmake
@@ -39,7 +39,7 @@ set(_protobuf_FIND_ABSL "if(NOT TARGET absl::strings)\n  find_package(absl CONFI
 
 if (BUILD_SHARED_LIBS AND MSVC)
   # On MSVC Abseil is bundled into a single DLL.
-  set(protobuf_ABSL_USED_TARGETS abseil_dll)
+  set(protobuf_ABSL_USED_TARGETS absl::abseil_dll)
 
   set(protobuf_ABSL_USED_TEST_TARGETS abseil_test_dll)
 else()


cmake/abseil-cpp.cmake
```diff
@@ -39,7 +39,7 @@ set(_protobuf_FIND_ABSL "if(NOT TARGET absl::strings)\n  find_package(absl CONFI
 
 if (BUILD_SHARED_LIBS AND MSVC)
   # On MSVC Abseil is bundled into a single DLL.
-  set(protobuf_ABSL_USED_TARGETS abseil_dll)
+  set(protobuf_ABSL_USED_TARGETS absl::abseil_dll)
 
   set(protobuf_ABSL_USED_TEST_TARGETS abseil_test_dll)
```
mkruskal-google, line 44: Do you need to add it here too?

cmake/abseil-cpp.cmake
```diff
@@ -39,7 +39,7 @@ set(_protobuf_FIND_ABSL "if(NOT TARGET absl::strings)\n  find_package(absl CONFI
 
 if (BUILD_SHARED_LIBS AND MSVC)
   # On MSVC Abseil is bundled into a single DLL.
-  set(protobuf_ABSL_USED_TARGETS abseil_dll)
+  set(protobuf_ABSL_USED_TARGETS absl::abseil_dll)
```
mkruskal-google, line 42: it looks like this breaks out cmake+msvc shared library build.  What version of cmake are you using where this is necessary?

johnwparent: 3.24

johnwparent: Oh I see the issue, this change is only required when using an abseil that is not part of your git submodule system. I'll add the necessary guards.

mkruskal-google: ah that makes a lot of sense, thanks for tracking that down!  Long-term we'd like to remove these submodules, and definitely not use them for testing.  Things like this keep popping up :(

johnwparent: I'll look into that as it has implications for my use case as well! I'd expect abseil just doesn't export those namespaces for MSVC because it only creates that single target or something like that.

Submodules are a bear to maintain to be sure, particularly on top of multiple possible methods of providing the deb the submodule vendors! 
Out of curiosity, what's the blocker for making the switch?

mkruskal-google: There's no real blocker, just prioritization.  We're working with folks from grpc and abseil to try to unify our cmake strategy.  The end result will likely end up with use deleting these submodules, or just keeping them around for developers (and maybe a single dedicated CI test for them)

johnwparent: Ah, very cool! The submodules are kinda nice to "just have" the dependency, but it's a lot more work to be sure.
FWIW I truly appreciate the ability to easily bring in an external abseil/etc with just the `find_package` command, makes everything much simpler, particularly coming from a package manager env like [spack](https://github.com/spack/spack) (which is the driver behind this PR). The find_package interface is one of CMake's most useful features IMO.

mkruskal-google: yea we're likely going to lean more heavily on `find_package` (and test it more thoroughly)

hlopko:
@johnwparent could you please sign the CLA (https://github.com/protocolbuffers/protobuf/pull/12891/checks?check_run_id=13728593608)? We cannot accept your contribution without it.

cmake/abseil-cpp.cmake
Ryanf55: FYI, it's best CMake practice (according to Craig Scott's professional CMake book) to have the target names the same for either a `find_package` or `add_subdirectory` method. 

In the install, the export's namespace `ns` should match any call to `add_librarary(ns::the_targetALIAS the_target)`. Then, consumers will always link to `ns::the_target` regardless of how it the target was found.

PR #12899, opened 2023-05-25
ruby: Don't assume RepeatedField->array is not NULL

casperisfine:
It's initially set as NULL and in corner cases it may not be initialized yet when we check its size.

Ref: https://github.com/protocolbuffers/protobuf/pull/12216#discussion_r1204879670

cc @haberman


casperisfine:
> Do you remember how you triggered this bug?

I don't, it was a bit too long ago, sorry :/

Issue #12896, opened 2023-05-24
Interest in modernizing the python code (use new py3.7+ language features)?


thisch:
I hope that it is fine that this is a feature request .I can post the question also on the mailing-list if desired.

Is there interest in cleaning up the python bindings to make use of new language features that are supported by the oldest python version that is currently supported by protobuf (=python3.7)?

The code could be automatically cleaned up by using tools like e..g `pyupgrade --py3-only --py37-plus` or `ruff`. 

Some things that would be changed by pyupgrade:

* Call `super()` without any args
* Use f-strings instead string-formatting via `%` or `.format()`
* Removal of the unicode literal, which is no longer needed
* Use new-style classes (classes that don't inherit from `object`)
* Use dict comprehension instead of calling `dict(...)` with a iterator (This has a minor performance benefit, since now lookup of the global `dict` is needed)
* Use `yield from`
* ...

haberman:
Yes, I think that modernizing within the range of Python versions we support would be very welcome. Feel free to send a PR.

fowles:
Sorry that this slipped between the cracks.  I have asked @anandolee to take a look

PR #12900, opened 2023-05-25
Modernize python code-base using the pyupgrade tool

thisch:
This change removes code that was needed in older versions of python and it also updates existing code to make use of new language features.

pyupgrade was run with the --py3-only --py37-plus flags to ensure that the code still works on our oldest supported python version.

The IOError was changed to OSError (also by pyupgrade), because the in python3.3 many exception classes have been merged into OSError. See https://github.com/asottile/pyupgrade/issues/141.

Closes: #12896

### DIFF
--- a/conformance/conformance_python.py
+++ b/conformance/conformance_python.py
@@ -175,12 +175,12 @@ def do_test_io():
   if len(length_bytes) == 0:
     return False   # EOF
   elif len(length_bytes) != 4:
-    raise IOError("I/O error")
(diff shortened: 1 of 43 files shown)


anandolee (approved):

thisch:
@anandolee approved this changed a couple of months ago. What is missing for this change to be merged? Thx

I guess I need to solve the merge conflicts, but will it then soon after I did that be merged?

PR #12916, opened 2023-05-26
cleanup(cmake): prefer target over global options

coryan:
In CMake >= 3.0 it is more idiomatic to set per-target compiler options than global compiler settings. I have kept these options and defines as `PRIVATE` so they won't be exported with the target.


fowles (approved):

coryan:
This is fairly close. I sent a separate change  to `utf8_range` that should fix the [JRuby 9.3 Build](https://github.com/protocolbuffers/protobuf/actions/runs/5092963400/jobs/9155013711?pr=12916)

PR #12923, opened 2023-05-30
ci: add java showcase binary compatibility check

burkedavison:
This PR adds a check to regenerate the Showcase Java client, and verify that the proposed changes do not break binary compatibility.

For the generation step, the PR is currently using a fresh bazel cache: https://github.com/protocolbuffers/protobuf/compare/main...burkedavison:protobuf:binary-compatibility-check?expand=1#diff-768f07ae5ae002ce203fc8bafc379c9f97a1ebace521a592f3f9f095f322a527R165-R177

However, we may want to use protobuf's `bazel-docker` instead:
https://github.com/protocolbuffers/protobuf/compare/main...burkedavison:protobuf:binary-compatibility-check?expand=1#diff-768f07ae5ae002ce203fc8bafc379c9f97a1ebace521a592f3f9f095f322a527R152-R164

### DIFF
--- a/.github/workflows/test_java.yml
+++ b/.github/workflows/test_java.yml
@@ -122,3 +122,63 @@ jobs:
       uses: googleapis/java-cloud-bom/tests/validate-bom@v26.13.0
       with:
         bom-path: java/bom/pom.xml
+
+  binary-compatibility:
+    name: Showcase Client Binary Compatibility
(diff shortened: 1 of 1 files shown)


mkruskal-google:
Hey Burke,

Thanks for this PR, but could you send it out as an internal cl?  We can't qualify test changes from forked repos

PR #12933, opened 2023-05-30
Allow overriding USE_SECONDARY_MAP

opsengine:
Make `USE_SECONDARY_MAP` configurable to mitigate https://github.com/protocolbuffers/protobuf/issues/11968

Need to add a way to propagate the macro as a build flag from bundle.


haberman:
Please add the test case https://github.com/protocolbuffers/protobuf/issues/11968 as a unit test, and enable `USE_SECONDARY_MAP` by default.

We can make it conditional again when/if we have a fix that can pass the test.

PR #12971, opened 2023-06-02
Set top level permissions to read only on GitHub Workflows

joycebrum:
Hi, here is Joyce from Google again.

I'd like starting suggesting the [Token-Permission check](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) fix.

Let me know if I might be missing any permission. Thanks!

### Security Reason

This is needed because, by default, github grants write-all permission to all workflows, which could be exploit by an attacker in case of a compromised workflow. Limiting permissions is a simple and effective way to also limit the impact of an eventual compromised workflow.

Thus, it is both a recommendation from [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) to always use credentials that are minimally scoped.

### DIFF
--- a/.github/workflows/test_runner.yml
+++ b/.github/workflows/test_runner.yml
@@ -63,7 +63,7 @@ jobs:
     if: |
       (github.event_name != 'pull_request' &&
        github.event_name != 'pull_request_target' &&
-       github.event.repository.full_name == 'protocolbuffers/protobuf') ||
+       github.event.repository.full_name == 'joycebrum/protobuf') ||
       (github.event_name == 'pull_request' &&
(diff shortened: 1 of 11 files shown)


.github/workflows/test_runner.yml
```diff
@@ -60,7 +63,7 @@ jobs:
     if: |
       (github.event_name != 'pull_request' &&
        github.event_name != 'pull_request_target' &&
-       github.event.repository.full_name == 'protocolbuffers/protobuf') ||
```
mkruskal-google, line 63: Revert this?

PR #12997, opened 2023-06-07
Cherry-pick portability fixes to 23.x line

fowles:


mkruskal-google (approved):

h-vetinari:
https://github.com/protocolbuffers/protobuf/pull/12997/commits/efa1ec75cf152991a64cedf9c4a1ec5cb826ee34 is an ABI-change, which I'm a bit surprised to see in a patch-release... Is there an policy regarding what changes can / should / cannot be backported?

fowles:
Yes, we do not provide any guarantees about ABI stability.  https://protobuf.dev/support/cross-version-runtime-guarantee/#cpp

PR #13054, opened 2023-06-14
Ruby: Only add RepeatedField instance to cache once fully initialized

casperisfine:
Fix: https://github.com/protocolbuffers/protobuf/issues/11968

Since calling a method on WeakMap allow the RubyVM to switch threads, adding a non-fully initialized object into the cache allow another thread from using it.

Note: with this fix, the other two thread may create their own wrapper for the same underlying Array. We don't know if it's a big deal or not.

Co-Authored-By: @peterzhu2118 

cc @haberman

### DIFF
--- a/ruby/ext/google/protobuf_c/repeated_field.c
+++ b/ruby/ext/google/protobuf_c/repeated_field.c
@@ -87,7 +87,6 @@ VALUE RepeatedField_GetRubyWrapper(upb_Array* array, TypeInfo type_info,
   if (val == Qnil) {
     val = RepeatedField_alloc(cRepeatedField);
     RepeatedField* self;
-    ObjectCache_Add(array, val);
     TypedData_Get_Struct(val, RepeatedField, &RepeatedField_type, self);
     self->array = array;
     self->arena = arena;
@@ -95,6 +94,7 @@ VALUE RepeatedField_GetRubyWrapper(upb_Array* array, TypeInfo type_info,
     if (self->type_info.type == kUpb_CType_Message) {
       self->type_class = Descriptor_DefToClass(type_info.def.msgdef);
     }
+    ObjectCache_Add(array, val);
   }
 
   PBRUBY_ASSERT(ruby_to_RepeatedField(val)->type_info.type == type_info.type);


casperisfine:
Ah damn it. Crediting @peterzhu2118 as co-author is making the CLA bot nervous -_-. 

Peter could you sign the CLA?

peterzhu2118:
I signed the CLA.

haberman:
Thanks so much for debugging and fixing this.

This pattern also occurs in the following place, which will also need the same change: https://github.com/protocolbuffers/protobuf/blob/2cf94fafe39eeab44d3ab83898aabf03ff930d7a/ruby/ext/google/protobuf_c/map.c#L96-L105

I wish there was a way to atomically perform a "get or insert" operation on the WeakMap, so we could avoid the duplicate wrapper situation. Duplicate wrappers are mostly benign, but they are observable by the user, eg. with `BasicObject#equal?`. If we could write our own WeakMap it would be possible to implement "get or insert", but it appears that it's not feasible to make a custom WeakMap unless we have `rb_objspace_garbage_object_p`.

Would it be possible to add a "get or insert" operation to `WeakMap` upstream, so we could at least avoid duplicate wrappers going forward?

fowles:
After this fix in debug builds we may start to fail the assert here: https://github.com/protocolbuffers/protobuf/blob/2cf94fafe39eeab44d3ab83898aabf03ff930d7a/ruby/ext/google/protobuf_c/protobuf.c#L406

I think the best solution is to have `ObjectCache_Add` return the `VALUE` that was inserted and then the caller can do a try insert and get back the official one that won the race.

fowles:
> After this fix in debug builds we may start to fail the assert here:
> 
> https://github.com/protocolbuffers/protobuf/blob/2cf94fafe39eeab44d3ab83898aabf03ff930d7a/ruby/ext/google/protobuf_c/protobuf.c#L406
> 
> I think the best solution is to have `ObjectCache_Add` return the `VALUE` that was inserted and then the caller can do a try insert and get back the official one that won the race.

Ok, that won't quite work since a racing thread can call `item_set` and clobber the existing value.  An `insert or return` operation (like @haberman asked for is really the right answer here).  I think we can survive the race condition of duplicate wrappers although it is a leaky abstraction, but we do need to remove the assert at the end

casperisfine:
> the other two thread may create their own wrapper for the same underlying Array. We don't know if it's a big deal or not.

Reading the code again, I think this case was already possible before since we were calling `RepeatedField_alloc(cRepeatedField)` before `ObjectCache_Add()`. 

Even with `getset`, both thread may need to alloc the `VALUE` and eventually one will notice it ran into a race condition and may abandon its own instance and use the other. 

All this to say, I think this is a different concern, and shouldn't prevent the fix from being merged.

fowles:
Closing this in favor of #13075 13075

PR #13075, opened 2023-06-15
Ruby: Refactor the object cache to better account for race conditions

casperisfine:
Superseeds: https://github.com/protocolbuffers/protobuf/pull/13054

The object cache is fundamentally subject to race conditions. Objects must be created before they are registered into the cache, so if two threads try to create the same object, we'll inevitably end up with two instances mapping to the same underlying memory.

To entirely prevent that we'd need a lot of extra locking which I don't think is really worth it compared to a few useless allocations.

Instead we can replace `ObjectCache_Add` by a `getset` type of operation, the extra instance is still created, but the later threads will receive the "canonical" instance and will be able to abandon their duplicated instance.

Additionally, this PR moves the ObjectCache implementation in Ruby, as it's much easier to debug there, and the performance difference is negligible. The `ObjectCache` instance is also exposed as `Google::Protobuf::OBJECT_CACHE` to better allow to debug potential memory issues.

cc @haberman @fowles

### DIFF
--- a/ruby/ext/google/protobuf_c/protobuf.h
+++ b/ruby/ext/google/protobuf_c/protobuf.h
@@ -84,10 +84,9 @@ void Arena_Pin(VALUE arena, VALUE obj);
 // being collected (though in Ruby <2.7 is it effectively strong, due to
 // implementation limitations).
 
-// Adds an entry to the cache. The "arena" parameter must give the arena that
-// "key" was allocated from.  In Ruby <2.7.0, it will be used to remove the key
-// from the cache when the arena is destroyed.
-void ObjectCache_Add(const void* key, VALUE val);
+// Tries to add a new entry to the cache, returning the newly installed value or
+// the pre-existing entry.
+VALUE ObjectCache_GetSet(const void* key, VALUE val);
 
 // Returns the cached object for this key, if any. Otherwise returns Qnil.
 VALUE ObjectCache_Get(const void* key);
--- a/ruby/ext/google/protobuf_c/repeated_field.c
+++ b/ruby/ext/google/protobuf_c/repeated_field.c
@@ -87,7 +87,6 @@ VALUE RepeatedField_GetRubyWrapper(upb_Array* array, TypeInfo type_info,
   if (val == Qnil) {
     val = RepeatedField_alloc(cRepeatedField);
     RepeatedField* self;
-    ObjectCache_Add(array, val);
     TypedData_Get_Struct(val, RepeatedField, &RepeatedField_type, self);
     self->array = array;
     self->arena = arena;
@@ -95,6 +94,7 @@ VALUE RepeatedField_GetRubyWrapper(upb_Array* array, TypeInfo type_info,
     if (self->type_info.type == kUpb_CType_Message) {
       self->type_class = Descriptor_DefToClass(type_info.def.msgdef);
     }
+    val = ObjectCache_GetSet(array, val);
   }
 
   PBRUBY_ASSERT(ruby_to_RepeatedField(val)->type_info.type == type_info.type);
@@ -615,7 +615,8 @@ VALUE RepeatedField_init(int argc, VALUE* argv, VALUE _self) {
 
   self->type_info = TypeInfo_FromClass(argc, argv, 0, &self->type_class, &ary);
   self->array = upb_Array_New(arena, self->type_info.type);
-  ObjectCache_Add(self->array, _self);
+  VALUE stored_val = ObjectCache_GetSet(self->array, _self);
+  PBRUBY_ASSERT(stored_val == _self);
 
   if (ary != Qnil) {
     if (!RB_TYPE_P(ary, T_ARRAY)) {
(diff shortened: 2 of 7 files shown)


ruby/ext/google/protobuf_c/protobuf.h
```diff
@@ -84,10 +84,11 @@ void Arena_Pin(VALUE arena, VALUE obj);
 // being collected (though in Ruby <2.7 is it effectively strong, due to
 // implementation limitations).
 
-// Adds an entry to the cache. The "arena" parameter must give the arena that
+// Adds an entry to the cache. Returns the pre-existing entry if there was one.
```
fowles, line 87: `Tries to add a new entry to the cache, returning the newly installed value or the pre-existing entry.`

feels a bit more accurate to me

ruby/ext/google/protobuf_c/protobuf.h
```diff
@@ -84,10 +84,11 @@ void Arena_Pin(VALUE arena, VALUE obj);
...
+// The "arena" parameter must give the arena that
```
fowles, line 88: what `arena` parameter?

ruby/ext/google/protobuf_c/repeated_field.c
```diff
@@ -615,7 +615,7 @@ VALUE RepeatedField_init(int argc, VALUE* argv, VALUE _self) {
 
   self->type_info = TypeInfo_FromClass(argc, argv, 0, &self->type_class, &ary);
   self->array = upb_Array_New(arena, self->type_info.type);
-  ObjectCache_Add(self->array, _self);
+  ObjectCache_GetSet(self->array, _self);
```
fowles, line 618: should we assert `_self == result` from this GetSet?  alternately, should we assign `_self` to the result?

ruby/lib/google/protobuf.rb
```diff
@@ -35,6 +35,97 @@
...
+        @map[key]
+      end
+
+      def getset(key, value)
+        @map[key] || @mutex.synchronize do
+          @map[key] ||= value
+        end
+      end
+
+      def purge
```
fowles, line 71: why do we have this method if it is empty?

ruby/ext/google/protobuf_c/protobuf.h
casperisfine: Probably an obsolete comment, I mostly just moved them around.

ruby/lib/google/protobuf.rb
casperisfine: So we expose the same interface as `LegacyObjectCache`.

Since people on older Rubies complain a lot about memory usage, I figured we could let them call purge if they want. But I can also make it private.

ruby/ext/google/protobuf_c/repeated_field.c
casperisfine: I'm fine either way. The assumption here is that we allocated our own, so we can't possible have a duplicate, but I may be wrong.

fowles: yeah, do you mind adding the assert?

ruby/lib/google/protobuf.rb
fowles: that makes sense, can you add a comment here?

ruby/ext/google/protobuf_c/repeated_field.c
casperisfine: Done.

ruby/ext/google/protobuf_c/protobuf.h
casperisfine: Done.

haberman:
How do we validate that there is no performance regression?

I think the relevant benchmark would be something like:

```ruby
msg = FooMessage()

# Read benchmark
msg.submsg = BarMessage()
1000000.times {
  msg.submsg
}

# Write benchmark
1000000.times {
  msg.submsg = BarMessage()
  msg.submsg
}
```

casperisfine:
> How do we validate that there is no performance regression?

So, note that there will necessarily be one. We're adding an extra method call on the fast path, and extra locking on the slow path.

That said I'm extremely unfamiliar with protobuf, and I never used it (it's a transitive dependency of various Google Cloud gems), so I'd appreciate if one of you could take this over, as it would be faster than doing back and forth. I mostly opened this PR as to show a proof of concept.

ruby/ext/google/protobuf_c/repeated_field.c
```diff
@@ -615,7 +615,8 @@ VALUE RepeatedField_init(int argc, VALUE* argv, VALUE _self) {
 
   self->type_info = TypeInfo_FromClass(argc, argv, 0, &self->type_class, &ary);
   self->array = upb_Array_New(arena, self->type_info.type);
-  ObjectCache_Add(self->array, _self);
+  VALUE stored_val = ObjectCache_GetSet(self->array, _self)
```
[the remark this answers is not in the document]
casperisfine: Thanks, fixed.

PR #13085, opened 2023-06-16
Add dllexport_decl for generated default instance.

owent:
Fixes #13084

PR #13090, opened 2023-06-16
Add dllexport_decl for generated default instance. (#13085)

### LINKED ISSUE #13084: `$dllexport_decl$` is missing in generated `.pb.cc`
Version: v21/v23
Language: C++

Windows,Linux

GCC 13, GCC 4.8, MSVC 19.36

**What did you do?**

1. Using `--cpp_out=dllexport_decl=XXX_API:<OUTPUT DIR>`.
2. Set the visibility to hidden.
3. Build generated `.pb.cc` as shared libraries.

Linking success.

**What did you see instead?**

Unresolved reference to `PACKAGE::_MESSAGE_default_instance_`.

`$dllexport_decl$` is missing for default instance in generated `.pb.cc`.

fowles:
Fixes #13084

Closes #13085

COPYBARA_INTEGRATE_REVIEW=https://github.com/protocolbuffers/protobuf/pull/13085 from owent:add_dllexport_decl_for_generated_default_instance a46d4e81c206785510c61c1ce7638a11e0d6dc6d PiperOrigin-RevId: 540938351

### DIFF
--- a/src/google/protobuf/compiler/cpp/file.cc
+++ b/src/google/protobuf/compiler/cpp/file.cc
@@ -556,7 +556,7 @@ void FileGenerator::GenerateSourceDefaultInstance(int idx, io::Printer* p) {
             };
           };
 
-          PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT
+          PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT$ dllexport_decl$
               PROTOBUF_ATTRIBUTE_INIT_PRIORITY1 const $type$ $name$;
         )cc");
   }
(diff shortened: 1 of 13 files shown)


src/google/protobuf/compiler/cpp/file.cc
```diff
@@ -556,7 +556,7 @@ void FileGenerator::GenerateSourceDefaultInstance(int idx, io::Printer* p) {
             };
           };
 
-          PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT
+          PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT$ dllexport_decl$
```
mkruskal-google, line 559: shouldn't the space be before the `$`?  I'm a little surprised this works but it seems to...

fowles: ` $foo$` means "add a space and expand foo"
`$ foo$` means "add a leading space if the expansion of foo is non-empty and expand foo"
`$foo $` means "add a trailing space if the expansion of foo is non-empty and expand foo"

mkruskal-google (approved):

mkruskal-google (approved):

PR #13097, opened 2023-06-17
Enable building PHP extension on Windows

Jan-E:
This PR adds a config.w32 for the PHP extension and includes it in the tgz for PECL.
I used this to build php_protobuf.dll for PHP 8.2.7. 
See https://github.com/phalcon/cphalcon/issues/16318#issuecomment-1592534432 for the request to build the extension and the result.

Please review and merge.

### DIFF
--- a/php/BUILD.bazel
+++ b/php/BUILD.bazel
@@ -148,6 +148,7 @@ pkg_files(
   ]) + [
     "//:LICENSE",
     "ext/google/protobuf/config.m4",
+    "ext/google/protobuf/config.w32",
     "ext/google/protobuf/wkt.inc",
   ],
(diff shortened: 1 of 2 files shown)


fowles (approved):

Jan-E:
One of the comments above says I requested a review by @perezd and @anandolee 
I never did that intentionally. Maybe it was somehow implicit. But if they could review the PR as well, they are of course very welcome.

fowles:
The system figures out likely people automatically.  At this point no actions are needed from you.  Odds are strong that we won't deal with the internal portions of landing this PR until Tuesday (Monday is a US holiday)

PR #13158, opened 2023-06-28
X86: optimize parsing var int with pext instruction.

LeiW000:
It speeds up parsing var int for x86 platform.

To enable it, the "-mbmi2" needs to passed to GCC.

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -36,6 +36,8 @@
 #include <string>
 #include <type_traits>
 #include <utility>
+#include <immintrin.h>
+#include <stdint.h>
 
 #include "absl/base/config.h"
 #include "absl/log/absl_check.h"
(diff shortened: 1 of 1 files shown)


LeiW000:
I'm in the process of being an authorized contributor for the protocol buffer project.

fowles:
We have already looked at this a great deal with https://github.com/protocolbuffers/protobuf/pull/10646 and https://github.com/protocolbuffers/protobuf/pull/10764 as well as several internal studies that we have not published.  The short version is that we don't intend to accept patches like this right now because the performance is usually worse on large scale system tests with real world data and/or the performance of instructions varies too much across architectures to be relied upon.

LeiW000:
@fowles , Thanks for your info.

With the PR, I saw +~4% performance gain with the real-word data from a data center. That is why I created it.

I read #10646 and #10764, and one of the concerns is the following.

> In particular, if your parser is inline, then its decode will be repeated throughout the binary. A larger faster decoder can dramatically increase the overall size of an application, and the resulting instruction cache misses can overwhelm the time savings of the faster decoder. 

My code is similar to the optimization for Arm in parse_context.h. It is even less than the code of that Arm optimization. Does it mean that the code of the Arm optimization should be disabled?

Overall, it seems to me that the big concern is that AMD Zen and Zen 2 have horrible pext latency.

martijnvels:
Hi Steven!

Can you clarify how you tested or benchmarked this? 

In the past we found that BMI enabled parsing offered significant speed ups compared to the 'old' parse functions (which are targeted in this proposal), but we never enabled these as supporting all different flavors of x86 is hard, and we found significant regressions on older AMD processors (Zen 2 / Rome)

With the introduction of the table driven parser, we are using a different parsing function `ShiftMixParseVarint` defined in varint_shuffle.h which we have found to be faster than a corresponding BMI implementation. This is somewhat of a complex story with various factor coming into play. First of, the current version is fully inlined, and better utilizes x86 superscalar capabilities. Introducing BMI parsing here would reduce the cycles spend in parsing, but also reduce the throughput. Our analysis shows that BMI gets in the way of the branch predictor / speculative execution: the processor can speculative execute parsing code for the next field while it's still finalizing 'this' field whereas we have seen that BMI prevents that from happening, as the processor can't speculate on BMI parsing outcomes. (note, this is our somewhat naive explanation, not matter of fact / documented)

There may be cases where BMI may still be beneficial, i.e., with large complex messages, there may be hot varint values being processed by the fallback parsing, or repeated varint fields where a tight loop BMI parse function can benefit. We are still exploring options on these cases. But we're intending to deprecate all 'older' use cases and implementations in ParseContext.

It would be great If you have benchmarks and/or example proto definitions with sample data showing the benefits of the BMI version even while using the TC parser. It may well be possible that there is room for BMI, but we need to see the see the actual use case for that, as all testing and internal betas we have explored inside google shave shown either 'no win' or 'regression'

Best,
Martijn

LeiW000:
Hi @martijnvels, appreciate the details you provided. Sorry for my late reply. I just came back from the vacation. 

The benchmark and the data I used are from a public cloud provider. Sorry, I am not allowed to share them here. That benchmark is to parse the real-word protobuf data, which is used in the data center of the cloud provider.

I was not aware of ShiftMixParseVarint. Give me some time to take a look at it and run my BMI version using the TC parser.

Issue #12881, opened 2023-05-23
Add c# serializationDeterministic 


fmg-lydonchandra:
**What language does this apply to?**
C#

**Describe the problem you are trying to solve.**
DeterministicSerialization for c#

**Describe the solution you'd like**
Same API CodedOutputStream.SetSerializationDeterministic for C# binding (same as Java/Ruby/Python API)

**Describe alternatives you've considered**
PInvoke into C++ library

**Additional context**
None

Is there any technical reason this has not been implemented for c# ?

jskeet:
I'd never heard of it before, but I don't see any reason why it can't be implemented for C#.

Looking at the Java implementation, it looks like this is *just* about the order in which map entries are serialized - is that your understanding as well?

I doubt that I'll have the time to work on it as a feature any time soon, but I'd happily review a pull request.

jskeet:
Fixed in #13160

PR #13160, opened 2023-06-28
add C# deterministic serialization

fmg-lydonchandra:
https://github.com/protocolbuffers/protobuf/issues/12881 

@jskeet  review please :-) as this is my first contrib to protobuf, any design / performance feedback is very sought after

### DIFF
--- a/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
+++ b/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
@@ -334,7 +334,7 @@ namespace Google.Protobuf
 
             output.WriteTag(TestMap.MapInt32Int32FieldNumber, WireFormat.WireType.LengthDelimited);
 
-            var key = 10; // Field 1 
+            var key = 10; // Field 1
             var value = 20; // Field 2
             var extra = 30; // Field 3
 
@@ -689,6 +689,57 @@ namespace Google.Protobuf
             {
                 Assert.AreEqual(TestAllTypes.OneofFieldOneofCase.OneofUint32, parsedMessage.OneofFieldCase);
             });
+        }
+
+        [Test]
+        public void MapSerializationDeterministicTrue_ThenBytesIdentical()
+        {
+            var testMap1 = new TestMap();
+            testMap1.MapInt32Bytes.Add(1, ByteString.CopyFromUtf8("test"));
+            testMap1.MapInt32Bytes.Add(2, ByteString.CopyFromUtf8("test2"));
+            var bytes1 = SerializeTestMap(testMap1, true);
+
+            var testMap2 = new TestMap();
+            testMap2.MapInt32Bytes.Add(2, ByteString.CopyFromUtf8("test2"));
+            testMap2.MapInt32Bytes.Add(1, ByteString.CopyFromUtf8("test"));
+            var bytes2 = SerializeTestMap(testMap2, true);
+
+            Assert.IsTrue(bytes1.SequenceEqual(bytes2));
+        }
+
+        [Test]
+        public void MapSerializationDeterministicFalse_ThenBytesDifferent()
+        {
+            var testMap1 = new TestMap();
+            testMap1.MapInt32Bytes.Add(1, ByteString.CopyFromUtf8("test"));
+            testMap1.MapInt32Bytes.Add(2, ByteString.CopyFromUtf8("test2"));
+            var bytes1 = SerializeTestMap(testMap1, false);
+
+            var testMap2 = new TestMap();
+            testMap2.MapInt32Bytes.Add(2, ByteString.CopyFromUtf8("test2"));
+            testMap2.MapInt32Bytes.Add(1, ByteString.CopyFromUtf8("test"));
+            var bytes2 = SerializeTestMap(testMap2, false);
+
+            Assert.IsFalse(bytes1.SequenceEqual(bytes2));
+        }
+
+
+        private byte[] SerializeTestMap(TestMap testMap, bool useDeterministicSerialization)
+        {
+            using (var memoryStream = new MemoryStream())
+            {
+                var codedOutputStream = new CodedOutputStream(memoryStream);
+                if (useDeterministicSerialization)
+                {
+                    codedOutputStream.UseDeterministicSerialization();
+                }
+
+                testMap.WriteTo(codedOutputStream);
+                codedOutputStream.Flush();
+
+                memoryStream.Seek(0, SeekOrigin.Begin);
+                return memoryStream.ToArray();
+            }
         }
 
         [Test]
--- a/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
+++ b/csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
@@ -692,45 +692,70 @@ namespace Google.Protobuf
         }
 
         [Test]
-        public void MapSerializationDeterministicTrue_ThenBytesIdentical()
+        public void MapStringString_DeterministicTrue_ThenBytesIdentical()
+        {
+            // Define three strings consisting of different versions of the letter I.
+            // LATIN CAPITAL LETTER I (U+0049)
+            string capitalLetterI = "I";
+            // LATIN SMALL LETTER I (U+0069)
+            string smallLetterI   = "i";
+            // LATIN SMALL LETTER DOTLESS I (U+0131)
+            string smallLetterDotlessI = "\u0131";
+            var testMap1 = new TestMap();
+
+            testMap1.MapStringString.Add(smallLetterDotlessI, "value_"+smallLetterDotlessI);
+            testMap1.MapStringString.Add(smallLetterI, "value_"+smallLetterI);
+            testMap1.MapStringString.Add(capitalLetterI, "content_"+capitalLetterI);
+            var bytes1 = SerializeTestMap(testMap1, true);
+
+            var testMap2 = new TestMap();
+            testMap2.MapStringString.Add(capitalLetterI, "content_"+capitalLetterI);
+            testMap2.MapStringString.Add(smallLetterI, "value_"+smallLetterI);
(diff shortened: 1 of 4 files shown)


csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -452,7 +452,17 @@ public void WriteTo(CodedOutputStream output, Codec codec)
             WriteContext.Initialize(output, out WriteContext ctx);
             try
             {
-                WriteTo(ref ctx, codec);
+                if (output.IsSerializationDeterministic())
```
fmg-lydonchandra, line 455: How do I unit-test this ? Is this for old proto2 format ?

csharp/src/Google.Protobuf.Test/GeneratedMessageTest.cs
```diff
@@ -691,6 +691,54 @@ public void OneofSerialization_DefaultValue()
             });
         }
 
+        [Test]
+        public void MapSerializationDeterministicTrue_ThenBytesIdentical()
```
fmg-lydonchandra, line 695: Will add more tests for different types of keys, once the general implementation approach is approved.

jskeet:
I'll get to this when I can - it may not be this week though.

csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -160,6 +160,41 @@ public long Position
...
+        /// Once set, the serializer will: (Note this is an implementation detail and may subject to
+        /// change in the future)
+        /// <list type="bullet">
+        /// <item><description>Sort map entries by keys in lexicographical order or numerical order. Note: For string
+        /// keys, the order is based on comparing the Unicode value of each character in the strings.
+        /// The order may be different from the deterministic serialization in other languages where
+        /// maps are sorted on the lexicographical order of the UTF8 encoded keys.</description></item>
+        /// </list>
+        /// </remarks>
+        public void UseDeterministicSerialization() {
```
jskeet, line 188: This should just be a property:

```csharp
public bool DeterministicSerialization { get; set; }
```

I'm not hugely keen on the stream being mutable like this, but it appears to be the design in other languages.

(Naming could be `IsDeterministic` or just `Deterministic`... I'm not hugely bothered on that front.)

csharp/src/Google.Protobuf/Collections/MapField.cs
jskeet: I'm not sure what you mean by that. Is *what* for proto2 format?

csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -160,6 +160,41 @@ public long Position
...
+        /// Note the deterministic serialization is NOT canonical across languages; it is also unstable
+        /// across different builds with schema changes due to unknown fields. Users who need canonical
+        /// serialization, e.g. persistent storage in a canonical form, fingerprinting, etc, should define
+        /// their own canonicalization specification and implement the serializer using reflection APIs
+        /// rather than relying on this API.
+        /// Once set, the serializer will: (Note this is an implementation detail and may subject to
+        /// change in the future)
+        /// <list type="bullet">
+        /// <item><description>Sort map entries by keys in lexicographical order or numerical order. Note: For string
+        /// keys, the order is based on comparing the Unicode value of each character in the strings.
```
jskeet, line 183: Each Unicode character, or each UTF-16 code unit?

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -452,7 +452,17 @@ public void WriteTo(CodedOutputStream output, Codec codec)
...
+                {
+                    var sorted = new List<KeyValuePair<TKey, TValue>>(list);
+                    sorted.Sort((pair1, pair2) => Comparer<TKey>.Default.Compare(pair1.Key, pair2.Key));
+
+                    WriteTo(ref ctx, codec, sorted);
+                }
+                else
+                {
+                    WriteTo(ref ctx, codec, list);
```
jskeet, line 464: I think it would be better not to duplicate the `WriteTo` call here. Something like:

```csharp
IList<KeyValuePair<TKey, TValue>> listToWrite = list;
if (output.IsDeterministic)
{
    listToWrite = new List<KeyValuePair<TKey, TValue>>(list);
    listToWrite.Sort(...);
}
WriteTo(ref ctx, codec, listToWrite);
```

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -469,7 +479,24 @@ public void WriteTo(CodedOutputStream output, Codec codec)
         [SecuritySafeCritical]
         public void WriteTo(ref WriteContext ctx, Codec codec)
         {
-            foreach (var entry in list)
+            if (ctx.state.CodedOutputStream is not null && ctx.state.CodedOutputStream.IsSerializationDeterministic())
```
jskeet, line 482: This can just be:

```csharp
if (ctx.state.CodedOutputStream?.IsDeterministic == true)
```

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -469,7 +479,24 @@ public void WriteTo(CodedOutputStream output, Codec codec)
...
+            {
+                // We can't sort the list in place, as that would invalidate the linked list.
+                // Instead, we create a new list, sort that, and then write it out.
+                var sorted = new List<KeyValuePair<TKey, TValue>>(list);
+                sorted.Sort((pair1, pair2) => Comparer<TKey>.Default.Compare(pair1.Key, pair2.Key));
```
jskeet, line 487: Given the repetition here, I'd probably extract the logic into a method of `GetSortedListCopy` or similar. Otherwise if we tweaked how the comparer is set up, it could end up inconsistent.

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -452,7 +452,17 @@ public void WriteTo(CodedOutputStream output, Codec codec)
             WriteContext.Initialize(output, out WriteContext ctx);
             try
             {
-                WriteTo(ref ctx, codec);
+                if (output.IsSerializationDeterministic())
+                {
+                    var sorted = new List<KeyValuePair<TKey, TValue>>(list);
+                    sorted.Sort((pair1, pair2) => Comparer<TKey>.Default.Compare(pair1.Key, pair2.Key));
```
jskeet, line 458: Using Comparer.Default begs the question as to whether the default string comparison is ordinal or not. I *believe* it is, but I'm not entirely sure.

I'd at least add a comment here - and possibly also note that map keys are guaranteed to be bools, integers or strings; never bytestrings or messages.

jskeet (commented):
Various suggestions, but I'm happy with the overall approach.

jskeet:
Thanks for the additional commit - I'll review on Monday.

csharp/src/Google.Protobuf/CodedOutputStream.cs
fmg-lydonchandra: Updated to  `UTF-16 code unit`

csharp/src/Google.Protobuf/Collections/MapField.cs
fmg-lydonchandra: refactored to `GetSortedListCopy`

csharp/src/Google.Protobuf/Collections/MapField.cs
fmg-lydonchandra: It is NOT ordinal by default,  explicitly using StringComparer.Ordinal now

csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -160,6 +160,33 @@ public long Position
             }
         }
 
+        /// <summary>
+        /// Configures serialization to be deterministic.
```
jskeet, line 164: Perhaps "Configures whether or not serialization is deterministic." (It's a property, but the current text sounds sounds like it's a method.)

csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -160,6 +160,33 @@ public long Position
...
+        /// </summary>
+        /// <remarks>
+        /// The deterministic serialization guarantees that for a given binary, equal messages (defined by the
+        /// equals methods in protos) will always be serialized to the same bytes. This implies:
+        /// <list type="bullet">
+        /// <item><description>Repeated serialization of a message will return the same bytes</description></item>
```
jskeet, line 170: Nit: add period

csharp/src/Google.Protobuf/CodedOutputStream.cs
```diff
@@ -160,6 +160,33 @@ public long Position
             }
         }
 
+        /// <summary>
+        /// Configures serialization to be deterministic.
+        /// </summary>
+        /// <remarks>
+        /// The deterministic serialization guarantees that for a given binary, equal messages (defined by the
```
jskeet, line 167: Nit: "The deterministic serialization" => "Deterministic serialization"

jskeet (approved):
Yup, this looks good to me. Will rerun tests.

csharp/src/Google.Protobuf/Collections/MapField.cs
fmg-lydonchandra: refactored to call `GetSortedListCopy`

csharp/src/Google.Protobuf/Collections/MapField.cs
fmg-lydonchandra: Refactored to `if (ctx.state.CodedOutputStream?.Deterministic ?? false)`

jskeet:
The tests appear to be failing in CI - I'll take a look.

mkruskal-google (approved):

PR #13204, opened 2023-07-04
Refactor the object cache to better account for race conditions

fowles:
Supersedes  #13075

### DIFF
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -108,7 +108,9 @@ void Message_InitPtr(VALUE self_, upb_Message* msg, VALUE arena) {
   Message* self = ruby_to_Message(self_);
   self->msg = msg;
   RB_OBJ_WRITE(self_, &self->arena, arena);
-  ObjectCache_GetSet(msg, self_);
+  VALUE stored = ObjectCache_GetSet(msg, self_);
+  (void)stored;
+  PBRUBY_ASSERT(stored == self_);
 }
 
 VALUE Message_GetArena(VALUE msg_rb) {
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -108,7 +108,7 @@ void Message_InitPtr(VALUE self_, upb_Message* msg, VALUE arena) {
   Message* self = ruby_to_Message(self_);
   self->msg = msg;
   RB_OBJ_WRITE(self_, &self->arena, arena);
-  ObjectCache_Add(msg, self_);
+  ObjectCache_GetSet(msg, self_);
 }
 
 VALUE Message_GetArena(VALUE msg_rb) {
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -108,7 +108,7 @@ void Message_InitPtr(VALUE self_, upb_Message* msg, VALUE arena) {
   Message* self = ruby_to_Message(self_);
   self->msg = msg;
   RB_OBJ_WRITE(self_, &self->arena, arena);
-  VALUE stored = ObjectCache_GetSet(msg, self_);
+  VALUE stored = ObjectCache_TryAdd(msg, self_);
   (void)stored;
   PBRUBY_ASSERT(stored == self_);
 }
--- a/ruby/ext/google/protobuf_c/protobuf.c
+++ b/ruby/ext/google/protobuf_c/protobuf.c
@@ -272,6 +272,8 @@ static void ObjectCache_Init(VALUE protobuf) {
 
   weak_obj_cache = rb_class_new_instance(0, NULL, cache_class);
   rb_const_set(protobuf, rb_intern("OBJECT_CACHE"), weak_obj_cache);
+  rb_const_set(protobuf, rb_intern("SIZEOF_LONG"), INT2NUM(SIZEOF_LONG));
+  rb_const_set(protobuf, rb_intern("SIZEOF_VALUE"), INT2NUM(SIZEOF_VALUE));
 }
 
 VALUE ObjectCache_TryAdd(const void *key, VALUE val) {
--- a/ruby/ext/google/protobuf_c/protobuf.c
+++ b/ruby/ext/google/protobuf_c/protobuf.c
@@ -252,164 +252,39 @@ void Arena_register(VALUE module) {
 // Object Cache
 // -----------------------------------------------------------------------------
 
-// A pointer -> Ruby Object cache that keeps references to Ruby wrapper
-// objects.  This allows us to look up any Ruby wrapper object by the address
-// of the object it is wrapping. That way we can avoid ever creating two
-// different wrapper objects for the same C object, which saves memory and
-// preserves object identity.
-//
-// We use WeakMap for the cache. For Ruby <2.7 we also need a secondary Hash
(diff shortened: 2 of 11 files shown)


ruby/lib/google/protobuf.rb
```diff
@@ -35,6 +35,98 @@
 # That way the module init can grab references to these.
 module Google
   module Protobuf
+    # A pointer -> Ruby Object cache that keeps references to Ruby wrapper
```
JasonLunn, line 38: Can we move these class definitions to their own file?

fowles: Done!

ruby/ext/google/protobuf_c/protobuf.h
```diff
@@ -84,10 +84,9 @@ void Arena_Pin(VALUE arena, VALUE obj);
 // being collected (though in Ruby <2.7 is it effectively strong, due to
 // implementation limitations).
 
-// Adds an entry to the cache. The "arena" parameter must give the arena that
-// "key" was allocated from.  In Ruby <2.7.0, it will be used to remove the key
-// from the cache when the arena is destroyed.
-void ObjectCache_Add(const void* key, VALUE val);
+// Tries to add a new entry to the cache, returning the newly installed value or
+// the pre-existing entry.
+VALUE ObjectCache_GetSet(const void* key, VALUE val);
```
mkruskal-google, line 89: "Add" still seems like a better fitting name than "GetSet".

fowles: I don't like `Add` since it can fail.  What about `TryAdd`?

ruby/lib/google/protobuf/object_cache.rb
```diff
@@ -0,0 +1,126 @@
...
+          end
+        end
+
+        @mutex.synchronize do
+          secondary_key ||= (@secondary_map[key] ||= Object.new)
+          @map[secondary_key] ||= value
+        end
+      end
+
+      def purge
```
JasonLunn, line 113: Recommend making `purge()` `private` since its only caller is `get()`.

ruby/lib/google/protobuf/object_cache.rb
```diff
@@ -0,0 +1,126 @@
...
+        @map[key]
+      end
+
+      def getset(key, value)
+        @map[key] || @mutex.synchronize do
+          @map[key] ||= value
+        end
+      end
+
+      def purge
```
JasonLunn, line 66: This can be omitted. Since `purge()` is never called by any code outside this class, so it doesn't need to part of the public API.

ruby/ext/google/protobuf_c/protobuf.c
```diff
@@ -252,164 +252,39 @@ void Arena_register(VALUE module) {
...
+  PBRUBY_ASSERT((key_val & 3) == 0);
+  return rb_funcall(weak_obj_cache, item_getset, 2, LL2NUM(key_val), val);
 }
 
 // Returns the cached object for this key, if any. Otherwise returns Qnil.
 VALUE ObjectCache_Get(const void *key) {
-  VALUE key_rb = ObjectCache_GetKey(key, false);
-  return rb_funcall(weak_obj_cache, item_get, 1, key_rb);
+  VALUE key_val = (VALUE)key;
+  PBRUBY_ASSERT((key_val & 3) == 0);
```
JasonLunn, line 286: What's the significance of the magic bitmask `3`? Should 3 be a named constant or should this assert be given a documentary comment?

ruby/ext/google/protobuf_c/repeated_field.c
```diff
@@ -87,14 +87,14 @@ VALUE RepeatedField_GetRubyWrapper(upb_Array* array, TypeInfo type_info,
...
     RepeatedField* self;
-    ObjectCache_Add(array, val);
     TypedData_Get_Struct(val, RepeatedField, &RepeatedField_type, self);
     self->array = array;
     self->arena = arena;
     self->type_info = type_info;
     if (self->type_info.type == kUpb_CType_Message) {
       self->type_class = Descriptor_DefToClass(type_info.def.msgdef);
     }
+    val = ObjectCache_GetSet(array, val);
```
JasonLunn, line 97: Does this need call need to be followed by an assert?

JasonLunn (changes requested):

ruby/ext/google/protobuf_c/message.c
```diff
@@ -108,7 +108,9 @@ void Message_InitPtr(VALUE self_, upb_Message* msg, VALUE arena) {
   Message* self = ruby_to_Message(self_);
   self->msg = msg;
   RB_OBJ_WRITE(self_, &self->arena, arena);
-  ObjectCache_Add(msg, self_);
+  VALUE stored = ObjectCache_GetSet(msg, self_);
+  (void)stored;
```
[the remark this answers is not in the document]
fowles: it prevents a compiler warning when assertions are disabled.

ruby/ext/google/protobuf_c/repeated_field.c
fowles: no, in this case we allow the race to change `val` and use that through the rest of the method

JasonLunn:
Should unit tests be added to verify any of the following:

1. Selection of the cache implementation on the current platform
2. Standalone behavior of both `ObjectCache` implementations outside of the side effects / asserts elsewhere in the runtime
3. Verification of the synchronization / safety properties of both caches
4. Verification of the garbage collection / memory consumption properties of the legacy cache

ruby/ext/google/protobuf_c/protobuf.c
fowles: I genuinely don't know.  This was a carry forward from the code in other places.  I would prefer deferring this until later when Josh returns from vacation.

ruby/ext/google/protobuf_c/message.c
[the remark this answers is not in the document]
JasonLunn: Is the assert always valid? Could `ObjectCache_GetSet()` assert that its return value is always its second argument just before returning it? That would obviate the need to have this weird construction to avoid the warning (since `ObjectCache_GetSet()` would actually use the value to return it) and simplify all the call sites by relieving callers of the duty to assert.

ruby/lib/google/protobuf/object_cache.rb
fowles: Done

ruby/lib/google/protobuf/object_cache.rb
fowles: Done

ruby/ext/google/protobuf_c/protobuf.h
fowles: Using `TryAdd`

ruby/ext/google/protobuf_c/message.c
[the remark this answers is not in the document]
fowles: not in all contexts.  When there are racing calls to `TryAdd` (formerly `GetSet`) with a particular key one of them is guaranteed to win and all of them will return the `VALUE` of that winner.  When other parts of the code knows there cannot be racing calls (as in new object construction), the assert is valid.  When there can be racing calls it is not valid.

mkruskal-google (approved):

mkruskal-google (approved):

Issue #6632, opened 2019-09-10
Add option to generate C# nullable annotations


YairHalberstadt:
**Language** : C#

**Feature Request**

C# 8.0 adds support for [nullability annotations](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references).

It would be helpful if the the generated protobuf objects added these annotations. This would provide warnings if people tried to dereference a possibly null value, or for example, tried to assign null to a string property of a protobuf object.

 At first sight it looks like the rules for which types can be nullable and which can't should be quite simple.

I think the first stage would be annotating the Protobuf Nuget Packages.

Of course nullable annotaions are a compile error in earlier versions of C# than c# 8.0. Therefore this has to be off by default, and triggered by a command line option.

It would be helpful if this was added for GRPC as well.

jtattermusch:
I'm not generally opposed to having this, but here are the challenges:
1. we would need a clear design - feel free to provide more thoughts here if you want
2. currently this is not a priority so there's no ETA and implementation would need to be contributed by someone from the community (but we'd need a design for this feature first).
3. C# 8 nullable annotations are a brand new C# 8 feature, but we definitely need to maintain the backward compatibility - which probably also means that it's going to take a while before this can actually be used. 

There are two separate areas:
1. annotations in the Google.Protobuf  runtime  (=the nuget) - then concern is backward-compatibility of the APIs (you cannot have an "option" for enabling/disabling this so it would need to work for everyone - and it's going to take a very long time until everyone can use C# 8).
2. annotations in the generated code (you can have a codegen option to enable the default-off feature), the question is how useful these checks would be - it's generated code and most of it is just field accessors, so I'm not sure we'd gain much here (a more detailed design would help).

YairHalberstadt:
> annotations in the Google.Protobuf runtime (=the nuget) - then concern is backward-compatibility of the APIs (you cannot have an "option" for enabling/disabling this so it would need to work for everyone - and it's going to take a very long time until everyone can use C# 8).

The annotations are backwards compatible, so I don't think this should be an issue.

> annotations in the generated code (you can have a codegen option to enable the default-off feature), the question is how useful these checks would be - it's generated code and most of it is just field accessors, so I'm not sure we'd gain much here (a more detailed design would help).

A very common mistake I make is to set a string to null. That would be the single most useful gain.

With GRPC specifically, I often by mistake return null from the server implementation. That would be another gain but of course that is for a different repository.

[...]

PR #13218, opened 2023-07-05
Add option to enable csharp nullable reference types

ogxd:
Related to:
- https://github.com/protocolbuffers/protobuf/issues/6632
- https://github.com/grpc/grpc/issues/20729
- https://github.com/grpc/grpc/pull/33507

# Context

Nullable reference types introduce a nullability context. Currently, the C# code generated via protobuf does not allow null for reference types and will intentionally throw at runtime when a null is assigned to such field.     
Because generated files are annotated with the `// <auto-generated>` attribute on top, nullable reference types won't work, even if NRT is explicitly enabled at the project level for instance.    
The consequence of this is that there is no way to benefit from NRT static code analysis to know in advance where fields could be assigned with a null (and thus it would throw at runtime)

# Solution

A first solution would be to allow removing the `// <auto-generated>` attribute, however, it has the unwanted side effects of enabling code analysis on the generated file itself, which we want to avoid.    
Instead, it turns out that marking generated files with `#nullable enable` works.    

The following PR is about this: Adding a CLI option to add `#nullable enable` on top of generated files.
The unit tests show that the code generated with that option now carries the nullability information so that we have a warning for any consuming code that may assign nulls ([CS8601](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings#possible-null-assigned-to-a-nonnullable-reference)). 

image

# Compatibility

Files annotated with `#nullable enable` can still compile fine under versions of dotnet prior to 6.0, so it is a backward-compatible change. 
Also, fields are still annotated as non-nullable, so I believe this feature does not break compatibility.

# Todo

I haven't made any changes to the documentation yet as I would like feedback first.    
CLA in progress.


jskeet:
Thank you for all your work on this, but we're still discussing whether or not we want to include nullable reference types in code generation at all. Please see https://github.com/grpc/grpc/pull/33507 where there is considerable discussion.
If you're happy to keep working on this PR knowing that the end result *may* be that we don't want to do anything, that's fine - but I won't be reviewing it in any detail until we've decided what we actually want to do.

(In general we *try* to avoid adding language-specific command line options - when to use NRTs is one of the key design discussions here. It feels like a new option *might* be the only way to go as different consumers may want to generate code for the same proto differently, but it's not certain yet.)

ogxd:
No problem, I understand.    
I just noticed that my PR is actually about what @JamesNK just proposed yesterday [here](https://github.com/grpc/grpc/pull/33507/files#r1252761833) and [here](https://github.com/grpc/grpc/pull/33507/files#r1252778711). As a user of protobuf for C# I confirm that simply having warnings in our code fulfills my most important need which is to avoid runtime null exceptions.

(I don't think I need to push anything else to that PR anyway until a decision has been taken, the implementation and the tests already shows the benefits what is involved with this solution)

tonydnewell:
The code generation needs further changing to fully support handline NRT.
See https://github.com/grpc/grpc/pull/33507#discussion_r1252792691
e.g. allowing message fields to be set to null, `Equals`, `MergeFrom`, and removing redundant null checks elsewhere.

ogxd:
> The code generation needs further changing to fully support handline NRT.
See https://github.com/grpc/grpc/pull/33507#discussion_r1252792691
e.g. allowing message fields to be set to null, Equals, MergeFrom, and removing redundant null checks elsewhere.

I don't think it does. As the [MS documentations says](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types#nullable-references-and-static-analysis):

> Nullable reference types aren't new class types, but rather annotations on existing reference types. The compiler uses those annotations to help you find potential null reference errors in your code.

This means that, given that the reference types are currently considered non-nullable, enabling NRT is about adding annotations so that a user knows that he might assign a null and trigger NREs.

What you describe with "allowing message fields to be set to null, Equals, MergeFrom, and removing redundant null checks elsewhere" is not about NRT but simply about allowing fields to be set to null. NRT annotations can be added on top, but again, it is different.

This PR is about adding NRT annotations so that users can consume this API while avoiding NREs, without any change to the generated code, unlike allowing fields to be set to null. Personally, I am not against also allowing reference fields to be set to null, but given that it would be an important API change, I think it's worth tackling the NRT subject first (because it's not a breaking change)

tonydnewell:
@ogxd Fields within a message that themselves are messages can be assigned null.
For example, proto file:
```proto
message Person {
  string name = 1;
  int32 id = 2;  // Unique ID number for this person.
  string email = 3;

  enum PhoneType {
    PHONE_TYPE_UNSPECIFIED = 0;
    PHONE_TYPE_MOBILE = 1;
    PHONE_TYPE_HOME = 2;
    PHONE_TYPE_WORK = 3;
  }

  message PhoneNumber {
    string number = 1;
    PhoneType type = 2;
  }

  PhoneNumber phone = 4; // this is a message and can be assigned null
}
```
with code:
```c#
Person person = new Person()
{
     Id = 1234,
     Name = "Fred",
     Email = "fred@example.com",
     Phone = null
 };
```
compiles today without your changes. Setting `Phone` to null effectively removes it from the message.

But with your changes:
```
Warning	CS8625	Cannot convert null literal to non-nullable reference type.
```

There is more involved than just adding `#nullable enable` to the generated code. Other changes to the generated code are necessary too.

ogxd:
Yes, you're right, I thought they were treated the same as strings. But it means adding the `?` annotation for these types would be enough to annotate these fields as nullable. This is also about adding annotations, the compiled code remains the same, so it is still a safe change.    
image

ogxd:
I've just updated the PR to use `#nullable enable annotations` instead of `#nullable enable` and to take into account [remark](https://github.com/protocolbuffers/protobuf/pull/13218#issuecomment-1636740241) from @tonydnewell about message fields. Generated code now also has nullable annotations for message fields (case covered by unit test)

Falco20019:
One special case that I think might have been missed (only gave it a quick glance) is `google.protobuf.StringValue` which then needs to be mapped onto `string?` instead of `string`.

src/google/protobuf/compiler/csharp/csharp_options.h
```diff
@@ -68,9 +69,12 @@ struct Options {
   // Whether the generated classes should have accessibility level of "internal".
   // Defaults to false that generates "public" classes.
   bool internal_access;
-  // Whether the generated classes should have a global::System.Serializable attribute added
-  // Defaults to false
+  // Whether the generated classes should have a global::System.Serializable attribute added.
+  // Defaults to false.
   bool serializable;
+  // Whether the generated files are annotated with "#nullable enable" to display warnings on potential null assignments.
```
Falco20019, line 75: The comment still only talks about `#nullable enable`, which is not true anymore.

src/google/protobuf/compiler/csharp/csharp_generator.cc
```diff
@@ -79,6 +79,8 @@ bool Generator::Generate(const FileDescriptor* file,
       cli_options.internal_access = true;
     } else if (options[i].first == "serializable") {
       cli_options.serializable = true;
+    } else if (options[i].first == "enable_nullable_reference_types") {
```
tonydnewell, line 82: My preference would be for a shorter option name. Microsoft just uses "nullable" (`<Nullable>` and `#nullable`) so maybe we should? Or "nullable_reference_types"?  Whatever is chosen I would be good if both the protocol buffers compiler and gRPC plugin use the same name.

jskeet: Given the requests that we've had for generating `int?`-backed properties for optional fields, I think just "nullable" could lead to confusion.

How about `enable_nrt`? I think NRT is a *reasonably* commonly-used abbrevation.

ogxd:
> I think there is a lot more work to be done. Adding `#nullable enable annotations` rather than `#nullable enable` hides a lot of potential problems.
> 
> As an experiment I took the code generated from [unittest_well_known_types.proto](https://github.com/protocolbuffers/protobuf/blob/main/csharp/compatibility_tests/v3.0.0/protos/src/google/protobuf/unittest_well_known_types.proto) and added `#nullable enable` and it produced 80 warnings.

As [the docs](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) says about NRT, its purpose is to "minimize the likelihood that your code causes the runtime to throw System.NullReferenceException". I think the point here is that the autogenerated code is not "your" code (you don't edit it, you don't maintain it, it is simply an external API generated by an external tool). This is why I think there shouldn't be warnings from NRT within the generated code itself, and one of the reasons why `#nullable enable annotations` exists.   

image

src/google/protobuf/compiler/csharp/csharp_generator.cc
ogxd: Indeed it is too long and can be confusing, `enable_nrt` seems more appropriate

tonydnewell:
> > I think there is a lot more work to be done. Adding `#nullable enable annotations` rather than `#nullable enable` hides a lot of potential problems.
> > As an experiment I took the code generated from [unittest_well_known_types.proto](https://github.com/protocolbuffers/protobuf/blob/main/csharp/compatibility_tests/v3.0.0/protos/src/google/protobuf/unittest_well_known_types.proto) and added `#nullable enable` and it produced 80 warnings.
> 
> As [the docs](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) says about NRT, its purpose is to "minimize the likelihood that your code causes the runtime to throw System.NullReferenceException". I think the point here is that the autogenerated code is not "your" code (you don't edit it, you don't maintain it, it is simply an external API generated by an external tool). This is why I think there shouldn't be warnings from NRT within the generated code itself, and one of the reasons why `#nullable enable annotations` exists.

Just my opinion (I'm not part of the protocol buffers team) but getting the generated code to be "clean" will probably help in not missing any corner cases, such as those arising with well-known types such as `StringValue` and `BytesValue`, and methods such as `Equals`.

JamesNK:
What are the warnings? Nullable warnings shouldn't come from generated code.

I'm guessing the problem is Google.Protobuf itself isn't annotated. `#nullable enable annotations` could be a short term fix and a longer term fix is annotating Google.Protobuf (not a small task).

tonydnewell:
> What are the warnings? Nullable warnings shouldn't come from generated code.
> 
> I'm guessing the problem is Google.Protobuf itself isn't annotated. `#nullable enable annotations` could be a short term fix and a longer term fix is annotating Google.Protobuf (not a small task. I see someone attempted to do that here - #9801).

A lot of warnings about setting null in reflection methods and classes such as `GeneratedClrTypeInfo` - these could wait for Google.Protobuf to be updated, or the generated code could add the null forgiving operator.

Also warnings in constructors when fields contain null values, e.g. `CS8618 Non-nullable field '_unknownFields' must contain a non-null value when exiting constructor. Consider declaring the field as nullable.`

Also warnings in `Equals` - e.g. `CS8604 Possible null reference argument for parameter 'other' in 'bool OneofWellKnownTypes.Equals(OneofWellKnownTypes other)'.`

I know that we could use `#nullable enable annotations` but by using `#nullable enable` instead it might help spot the areas of the code that really need fixing.

JamesNK:
It sounds like these a nullable warnings related to the generated code. In that case you need to update the generated code to allow things to be null where necessary.

> Also warnings in constructors when fields contain null values, e.g. `CS8618 Non-nullable field '_unknownFields' must contain a non-null value when exiting constructor. Consider declaring the field as nullable.`

`_unknownFields` field should be nullable.

> Also warnings in `Equals` - e.g. `CS8604 Possible null reference argument for parameter 'other' in 'bool OneofWellKnownTypes.Equals(OneofWellKnownTypes other)'.`

Messages implement `IMessage` which implements `IEquatable<T>`. Its equals method accepts a nullable argument. The generated code needs to match: https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1.equals?view=net-7.0#system-iequatable-1-equals(-0)

jskeet:
Just to chime in, I agree that the generated code should build warning-free. It sounds like some more work is needed - but thanks for what you've already done, and I look forward to looking more closely when it's fully "done".
I suspect that when Google.Protobuf *eventually* becomes NRT-aware, we *may* need a second pass anyway - but that's okay.

ogxd:
Thanks for the feedback! I just became dad on the 4th of August so my days were a little busy since but if the nullable topic hasn't addressed in parallel since I'll update the PR accordingly to your remarks when I have some time 😄

stebet:
> Thanks for the feedback! I just became dad on the 4th of August so my days were a little busy since but if the nullable topic hasn't addressed in parallel since I'll update the PR accordingly to your remarks when I have some time 😄

I'm very excited about getting this into the tooling so if there is anything I can do to help or speed things along, please let me know :)

csharp/generate_protos.sh
```diff
@@ -75,6 +75,16 @@ $PROTOC -Isrc -I. \
     src/google/protobuf/test_messages_proto2.proto \
     src/google/protobuf/unittest_proto3_optional.proto \
     src/google/protobuf/unittest_retention.proto
+# Test proto for nullable reference types tests
+$PROTOC -Isrc -I. \
+    --experimental_allow_proto3_optional \
+    --csharp_out=csharp/src/Google.Protobuf.Test.TestProtos \
+    --csharp_opt=file_extension=.pb.cs \
+    --csharp_opt=enable_nullable_reference_types \
+    --descriptor_set_out=csharp/src/Google.Protobuf.Test/testprotos.pb \
```
tonydnewell, line 84: The overwriting of testprotos.pb here causes the tests in DescriptorDeclarationTest.cs to fail

csharp/protos/unittest_nrt_proto3.proto
```diff
@@ -0,0 +1,46 @@
...
+package protobuf_unittest3;
+
+message TestNrtMessage {
+  message TestNrtNestedMessage {
+  }
+
+  TestNrtNestedMessage message_field = 1;
+  string string_field = 2;
+  int32 int32_field = 3;
+}
```
tonydnewell, line 46: Code generated for unittest_nrt_proto3.proto causes RefStructCompatibilityTest.cs tests to fail

csharp/src/Google.Protobuf.Test/testprotos.pb
tonydnewell, line 1: This version of the files causes tests in DescriptorDeclarationTest.cs to fail

PR #13255, opened 2023-07-10
Parsing single varint optimizations

palade:
This PR has the changes for parsing variable-length integers. These changes are based on the BMI2 instruction. This instruction may not be available on all the CPUs so a check is added to consider the implementation based on the available instruction set.

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -37,6 +37,10 @@
 #include <type_traits>
 #include <utility>
 
+#if defined(__BMI2__) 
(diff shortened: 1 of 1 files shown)


zhangskz:
@acozzette Does this PR have the same issues as https://github.com/protocolbuffers/protobuf/pull/13256 limiting our ability to adopt these changes? If so, should these PRs be closed?

acozzette:
@zhangskz Yes, unfortunately I think we can't accept these changes for the same reason as for the other PR.

PR #13256, opened 2023-07-10
Parsing array of varints optimizations

palade:
This PR has the changes for parsing array of varints. Most of the changes originate in the `ReadPackedVarint` and `ReadPackedVarintArray` functions. The core of optimizations is to break the one-at-a-time parsing in the original code into a series of steps to leverage SIMD instructions.

### DIFF
--- a/src/google/protobuf/parse_context.cc
+++ b/src/google/protobuf/parse_context.cc
@@ -465,27 +465,33 @@ const char* VarintParser(void* object, const char* ptr, ParseContext* ctx) {
 
 const char* PackedInt32Parser(void* object, const char* ptr,
                               ParseContext* ctx) {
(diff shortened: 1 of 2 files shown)


acozzette:
@martijnvels

acozzette:
@palade Thank you for the pull request, but unfortunately I don't think we will be able to accept it at this time. We are not yet set up to use AVX 512 BMI, so we're not in a good position to maintain code that uses it. This might be something we could accept in the future though, especially if we can quantify the performance benefits in production.

acozzette:
Actually I heard from @fowles that AVX 512 instructions cause modern CPUs to downclock, so that problem will likely prevent us from ever using these instructions for varint parsing.

PR #13257, opened 2023-07-10
AVX-based UTF-8 Validation optimizations

palade:
This PR has changes made to the UTF-8 validation function. These were mainly done in the IsStructurallyValidUTF8 function, which previously lived in `src/google/protobuf/stubs/structurally_valid.cc`. I noticed that the code around this functionality has been moved to `utf8-range` module. Based on the length of the buffer and also based on the availability of certain instructions in the CPU, a certain implementation will be selected. To take advantage of the optimizations, the code should be compiled "-march=native".

### DIFF
--- a/third_party/utf8_range/utils.h
+++ b/third_party/utf8_range/utils.h
@@ -1,0 +1,88 @@
+#pragma once
+
+
+#ifdef _MSC_VER
+
+#include <intrin.h>
+
+class uint128_t {
+public:
+    uint64_t lo, hi;
+    uint128_t(int v) { lo = v; hi = 0; }
+    uint128_t(uint32_t v) { lo = v; hi = 0; }
+    uint128_t(uint64_t v) { lo = v; hi = 0; }
+    uint128_t(uint64_t p_hi, uint64_t p_lo) { lo = p_lo; hi = p_hi; }
+
+    inline uint128_t &operator+=(const uint64_t &other) {
+        lo += other;
+        hi += (lo < other);
+        return *this;
+    }
+    inline uint128_t &operator+=(const uint128_t &other) {
+        lo += other.lo;
+        hi += (lo < other.lo);
+        hi += other.hi;
+        return *this;
+    }
+    inline uint128_t &operator&=(const uint128_t &other) {
+        lo &= other.lo;
+        hi &= other.hi;
+        return *this;
+    }
+    inline uint128_t &operator|=(const uint128_t &other) {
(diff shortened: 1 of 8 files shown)


mkruskal-google:
hey Andrei, thanks for the contributions!  These changes need to be made upstream in the utf8_range repo though.  Can you port these there?  We can then bump the pin here if they get accepted

third_party/utf8_range/utils.h
```diff
@@ -0,0 +1,88 @@
...
+};
+
+inline bool operator==(const uint128_t &lhs, const int &rhs) { return lhs.lo == rhs && lhs.hi == 0; }
+inline bool operator!=(const uint128_t &lhs, const int &rhs) { return !(lhs == rhs); }
+
+#else
+
+#include <x86intrin.h>
+
+typedef unsigned __int128 uint128_t;
```
mkruskal-google, line 86: We're seeing a build failure in our 32-bit linux tests:
```
In file included from utf8vec.cpp:12:
utils.h:86:18: error: expected unqualified-id before '__int128'
 typedef unsigned __int128 uint128_t;
                  ^~~~~~~~
```

danlark1:
Thanks for the contribution. Several notes:

1. For now having bigger throughput for UTF-8 validation is not a big priority. 80% of CPU cycles we spend in skipping ascii code. Only 20% fall through
2. Strings are not big, median is around 15 bytes, 90th percentile is 64

That means we do care about latency of the execution as well. Say, ascii or small strings are more important. Can we have benchmarks for them as well? We don't want to waste 3 additional branches for empty strings which might be costly

palade:
> hey Andrei, thanks for the contributions! These changes need to be made upstream in the utf8_range repo though. Can you port these there? We can then bump the pin here if they get accepted

Hey Mike, I now added the changes to the `utf8_range` repo.  

https://github.com/protocolbuffers/utf8_range/pull/25

third_party/utf8_range/utils.h
palade: I refactored the code here to address this error https://github.com/protocolbuffers/utf8_range/pull/25

mkruskal-google:
Can we get some timing benchmarks as well?  I think Danila's point was that we wouldn't want to sacrifice execution time longer for small strings just to get high throughput on big strings

palade:
Adding a comment to maintain this PR active. Additional benchmarking data requested.

PR #13293, opened 2023-07-12
Use `omit` to skip ObjectCache tests when it is not in use.

JasonLunn:
Unblocks execution of these tests on JRuby FFI.

Automated reformatting of WORKSPACE with documentary comment to ease future debugging of JRuby tests locally.

### DIFF
--- a/ruby/tests/object_cache_test.rb
+++ b/ruby/tests/object_cache_test.rb
@@ -3,10 +3,7 @@ require 'test/unit'
 
 class PlatformTest < Test::Unit::TestCase
   def test_correct_implementation_for_platform
-    if RUBY_PLATFORM == "java"
-      return
-    end
-
+    omit('OBJECT_CACHE not defined') unless defined? Google::Protobuf::OBJECT_CACHE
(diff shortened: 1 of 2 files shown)


ruby/tests/object_cache_test.rb
```diff
@@ -48,12 +45,9 @@ def test_multithreaded_access
   end
 end
 
-if RUBY_PLATFORM == "java"
-  return
-end
-
 class ObjectCacheTest < Test::Unit::TestCase
   def create
+    omit('OBJECT_CACHE not defined') unless defined? Google::Protobuf::OBJECT_CACHE
```
fowles, line 50: feels weird that you put `omit` on the factory method.  You sure that is correct?

JasonLunn: Under the hood, `omit` raises an exception that test unit catches and puts in a different category than other unchecked exceptions. I put it here just to avoid code echo, but I don't feel strongly about keeping the test DRY.

fowles (approved):

PR #13308, opened 2023-07-13
Add standalone protoc plugin binaries

jchadwick-buf:
This PR adds targets to build `protoc` plugins for all the built-in plugins. This will result in the following binaries being available:

- `protoc-gen-cpp`
- `protoc-gen-csharp`
- `protoc-gen-java`
- `protoc-gen-kotlin`
- `protoc-gen-objectivec`
- `protoc-gen-php`
- `protoc-gen-python`
- `protoc-gen-pyi`
- `protoc-gen-ruby`
- `protoc-gen-rust`

The goal of this PR is for these to be added as assets to each GitHub release, although we are unable to find the code that builds the assets for releases - if it is in this repository, let us know and we're happy to do any work required!

There's multiple motivations for this PR:

- Optimally, people want to version their plugins separately from their compiler. That is, the version of `protoc` should be decoupled from the version of each plugin, as is the case with the rest of the Protobuf ecosystem outside of the built-in plugins. You should be able to, for example, take advantage of fixes in `protoc` v24.0 whilst continuing to use the Java generated code from v23.0, in case there are breaking changes you're unable to adopt yet.
- This partially addresses https://github.com/protocolbuffers/protobuf/issues/5587 - the core motivation behind this issue is that when piping a `FileDescriptorSet` into `protoc` with `--descriptor_set_in`, and then invoking one of the built-in plugins, different code can result, as the value for `json_name` will be subtly different when taking an external `FileDescriptorSet` vs `protoc` producing one internally and then invoking one of the built-in plugins.
- Selfishly, it makes life a lot easier for alternative Protobuf compilers. For example, for us with `buf`, to use the built-in plugins, we have a convoluted [mechanism to proxy to protoc](https://github.com/bufbuild/buf/blob/main/private/bufpkg/bufpluginexec/protoc_proxy_handler.go) that cannot deal with issue #5587, and requires people to install both `protoc` and `buf` when they may just want one of the built-in plugins. Having separate, independently-versionable plugins would make our lives a lot easier.

The majority of this PR is something [we have already been maintaining](https://github.com/bufbuild/plugins/tree/main/plugins/protocolbuffers/cpp/v23.4) for our remote plugins feature within Buf, we're just hoping we can get this upstreamed, as we think it would benefit the Protobuf community.

Follow-ups that would be nice:

- Adding to GitHub releases, as mentioned.
- Get these plugins installed as part of the Homebrew `protobuf` formula. There's precedent for this with i.e. the `grpc` formula, which installs each language-specific plugin.
- Change `protoc` to default to using an on-disk plugin vs a built-in plugin, that is if `protoc-gen-cpp` exists, use `protoc-gen-cpp` over the builtin `cpp` plugin. This allows people to independently version plugins as mentioned, and mirrors the behavior of the Well-Known Types (if a WKT is in the include path, it is used instead of the WKTs included as part of the `protoc` distribution). We're happy to put up a PR for this as well.


jchadwick-buf:
CLA has been signed.

src/google/protobuf/compiler/java/kotlin_main.cc
```diff
@@ -0,0 +1,37 @@
...
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "google/protobuf/compiler/java/kotlin_generator.h"
+#include "google/protobuf/compiler/plugin.h"
+
+int main(int argc, char *argv[]) {
+  google::protobuf::compiler::java::KotlinGenerator generator;
```
fowles, line 35: doesn't this also need the JavaGenerator to be useful?

jchadwick-buf: Sorry, could you please elaborate if possible? Right now I have a separate entrypoint for Java in `src/google/protobuf/compiler/java/main.cc`, and this entrypoint here using only KotlinGenerator matches what we already do to make an unofficial `protoc-gen-kotlin`.

fowles: I have two distinct directions of concern/thought here.

1) I am imagining a hypothetical user that has installed `protoc-gen-kotlin`.  They are able to generate kotlin code, but that code depends on a java library that they are unable to generate.  Seems like not a great situation.

2) A user may have different versions of `protoc-gen-kotlin` and `protoc-gen-java`.  I literally don't know if their generated kotlin stubs will work.  I am not certain how tight the coupling is between these two and don't relish finding out the hard way.

bufdev: > A user may have different versions of protoc-gen-kotlin and protoc-gen-java. I literally don't know if their generated kotlin stubs will work. I am not certain how tight the coupling is between these two and don't relish finding out the hard way.

I don't think this is any different than the rest of the Protobuf ecosystem though - you could have different versions of dependent plugins, say a Protobuf and grpc plugin, and it may or may not cause an issue. I can see the integration here may be tighter, but I'd think that it's an orthogonal issue. I'd love to have a way to declare library and plugin dependencies for plugins, but it's not part of the plugin API at the moment

fowles: (1) is genuinely different, we don't have a way to end up in this state right now.

(2) has existing issues, but they require having two distinct full copies of `protoc` installed, which is not nearly as easy as having two different binaries installed.

bufdev: Perhaps we're talking about two different things here - my argument is that neither (1) or (2) is any different from the rest of the plugin ecosystem, and I don't know if there's a reason to special case kotlin and Java. As an example, if you generate code with protoc-gen-go-grpc, you need generated code from protoc-gen-go, but if you don't have protoc-gen-go installed, you can't generate it.

If there's a kotlin/Java-specific solution to this though, that's great. But I think my argument is: massively in favor of dependencies between plugins (we actually represent the dependencies in BSR), but not sure builtins are any different than the rest of the ecosystem, and not sure if this is the place to have special-case guardrails. But I may be missing something massive here (wouldn't be the first time), apologies if so.

fowles:
> Optimally, people want to version their plugins separately from their compiler. That is, the version of `protoc` should be decoupled from the version of each plugin, as is the case with the rest of the Protobuf ecosystem outside of the built-in plugins. You should be able to, for example, take advantage of fixes in `protoc` v24.0 whilst continuing to use the Java generated code from v23.0, in case there are breaking changes you're unable to adopt yet.

Can you clarify if you are referring to the same language in v23 and v24 here, or do you mean something like "you want v24 java, but v23 C++"?

bufdev:
Sorry, should have clarified here:

- I don't think this PR is intending to make it so that v23 java and v24 java can be generated into the same codebase, that's a separate issue entirely heh/
- It would be nice to easily be able to do v23 java and v24 c++ in the same monorepo as an example, something that is more difficult now if you have a single point of entry for `protoc`.
- It would also be nice if the core compiler functionality of `protoc` would be able to work with older plugins, so for example there's a bug fix in v24, but I continue to use `protoc-gen-java` v23.

Does that make sense?

fowles:
Sorry this has lingered for so long.  There are some internal things that are moving around that make this direction look really promising, so odds are strong we will take this eventually, but moving the behemoth takes time :)

jchadwick-buf:
Oops, this PR probably shouldn't be closed. I could of course just open a new PR later, but the relevant discussion is probably worth keeping.

googleberg:
This is still coming along internally.

jchadwick-buf:
Bump to unmark inactive.

jchadwick-buf:
Bumping again. I should probably rebase/update this too.

jchadwick-buf:
Still relevant, needs to be rebased though.

JasonLunn:
@jchadwick-buf - can you rebase this PR?

jchadwick-buf:
Sorry for the delay. I have been a bit busy.

I've rebased. I see the Kotlin generator moved but otherwise not much needed to be changed, seems to be building. Please let me know if there's anything else that needs to be done.

JasonLunn:
Looks close! The last test run looks like it only failed due to comparison with goldens:

```
Installed files do not match goldens!
If this is expected, please update the golden file: cmake/installed_bin_golden.txt
The following diffs were found:
--- cmake/installed_bin_golden.txt	2025-03-26 23:42:40.086630065 +0000
+++ static/bin.txt	2025-03-26 23:42:41.302627899 +0000
@@ -1,3 +1,13 @@
 protoc
+protoc-gen-cpp
+protoc-gen-csharp
+protoc-gen-java
+protoc-gen-kotlin
+protoc-gen-objectivec
+protoc-gen-php
+protoc-gen-pyi
+protoc-gen-python
+protoc-gen-ruby
+protoc-gen-rust
 protoc-gen-upb
 protoc-gen-upbdefs
```

jchadwick-buf:
Thanks, I've applied those changes.

googleberg (approved):

jchadwick-buf:
I see that the CL failed internally. Let me know if there's anything I should do to help make this pass in google3; otherwise, I'll cross my fingers that it isn't anything too ridiculous and leave it to y'all.

PR #13320, opened 2023-07-14
Fix gcc compilation error about non virtual destructor

yanavlasov:
gcc is complaining that MapFieldBase has virtual methods but non-virtual destructor.

### DIFF
--- a/src/google/protobuf/map_field.h
+++ b/src/google/protobuf/map_field.h
@@ -324,7 +324,7 @@ class PROTOBUF_EXPORT MapFieldBase : public MapFieldBaseForParse {
 
  protected:
   // "protected" stops users from deleting a `MapFieldBase *`
-  ~MapFieldBase();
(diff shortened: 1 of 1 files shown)


fowles:
This was deliberately done to avoid the overhead of having an additional entry in the vtable.  These destructors are protected, so the warning is suppressed on most compilers automatically.

yanavlasov:
> This was deliberately done to avoid the overhead of having an additional entry in the vtable. These destructors are protected, so the warning is suppressed on most compilers automatically.

Hmm, it is not suppressed on gcc 9. Is there minimum gcc version requirement to compile protobuf?

fowles:
https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

GCC >=7.3.1

that said, we don't support compiling our headers with arbitrary warnings turned on.

PR #13343, opened 2023-07-18
Ruby FFI implementation

JasonLunn:
Supersedes #11483.


haberman:
Looks like CI tests are failing?

JasonLunn:
> Looks like CI tests are failing?

Yep! The new test `implementation.rb` failed because FFI was not being properly loaded in these configs and was falling back to the native implementation. So yay, the test works!

Just pushed a fix which I expect will fix the test failure. 🤞

.github/workflows/test_ruby.yml
```diff
@@ -102,23 +112,26 @@ jobs:
         with:
           credentials: ${{ secrets.GAR_SERVICE_ACCOUNT }}
           bazel-cache: ruby_macos/${{ matrix.version }}
-          bazel: test //ruby/... --test_env=KOKORO_RUBY_VERSION=${{ matrix.version }}
+          bazel: test //ruby/... --test_env=KOKORO_RUBY_VERSION=${{ matrix.version }} --test_env=BAZEL=true ${{ matrix.ffi == 'FFI' && '--//ruby:ffi=enabled --test_env=PROTOCOL_BUFFERS_RUBY_IMPLEMENTATION=FFI' || '' }}
```
mkruskal-google, line 115: Same comment here, you could add a `flags` and a `suffix` argument to capture this and make it more readable

.github/workflows/test_ruby.yml
```diff
@@ -134,8 +147,9 @@ jobs:
           bash: >
             bazel --version;
             ruby --version;
-            bazel build //ruby:release //:protoc $BAZEL_FLAGS;
+            bazel build //ruby:release //:protoc ${{ matrix.ffi == 'FFI' && '--//ruby:ffi=enabled' || '' }} $BAZEL_FLAGS;
```
mkruskal-google, line 150: Same comment here

.gitignore
```diff
@@ -166,6 +166,17 @@ vendor/
 /bazel-*
 
 # ruby test output
+ruby/lib/google/protobuf/any_pb.rb
```
mkruskal-google, line 169: Can we exclude all `*_pb.rb` files?

ruby/.gitignore
```diff
@@ -6,4 +6,9 @@ protobuf-jruby.iml
 target/
 pkg/
 tmp/
-tests/google/
\ No newline at end of file
+tests/google/
```
mkruskal-google, line 9: I wasn't even aware this file existed.  Why not collapse it into the top-level one?

ruby/BUILD.bazel
```diff
@@ -13,12 +15,83 @@ load("//:protobuf_version.bzl", "PROTOBUF_RUBY_VERSION")
...
+selects.config_setting_group(
+    name = "macos_ffi_enabled",
+    match_all = [
+        ":ffi_enabled",
+        "@platforms//os:osx",
+    ],
+)
+
+selects.config_setting_group(
+    name = "linux_ffi_enabled",
```
mkruskal-google, line 82: What about windows?

ruby/README.md
```diff
@@ -93,6 +109,12 @@ From the project root (rather than the `ruby` directory):
 $ bazel test //ruby/tests/...
 ```
 
+To run tests against the FFI implementation:
+
+```
+$ bazel test //ruby/tests/... //ruby:ffi=enabled --test_env=PROTOCOL_BUFFERS_RUBY_IMPLEMENTATION=FFI
```
mkruskal-google, line 115: Why do we need to specify both?  Couldn't we make the `protobuf` target set this env flag based on the config setting?

.github/workflows/test_ruby.yml
```diff
@@ -40,7 +43,7 @@ jobs:
           image: ${{ matrix.image || format('us-docker.pkg.dev/protobuf-build/containers/test/linux/ruby:{0}-{1}-508417e5215994ade7585d28ba3aad681a25fa5d', matrix.ruby, matrix.bazel) }}
           credentials: ${{ secrets.GAR_SERVICE_ACCOUNT }}
           bazel-cache: ruby_linux/${{ matrix.ruby }}_${{ matrix.bazel }}
-          bazel: test //ruby/... //ruby/tests:ruby_version --test_env=KOKORO_RUBY_VERSION
+          bazel: test //ruby/... //ruby/tests:ruby_version --test_env=KOKORO_RUBY_VERSION --test_env=BAZEL=true ${{ matrix.ffi == 'FFI' && '--//ruby:ffi=enabled --test_env=PROTOCOL_BUFFERS_RUBY_IMPLEMENTATION=FFI' || '' }}
```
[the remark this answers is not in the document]
JasonLunn: I am new to Github actions and the syntax they support - if there is a simpler expression that would do the same thing, I'm all for it. The choice to do this once on this line with substitution logic rather than introduce a `flags` attribute into the testing matrix intended to 1) reduce echo and 2) give myself one place to change things when I iterated.

.github/workflows/test_ruby.yml
JasonLunn: I'm not sure I'm visualizing the same change you have in your mind - can you give an example? From a readability (and a reducing opportunity to introduce a typo) standpoint, I wouldn't want to repeat the same value of the `flags` attribute for every config that needs it.

ruby/.gitignore
JasonLunn: Great question. I don't know what sort of exclusions belong in what level of the hierarchy. I assume that having multiple files avoids triggering overly broad reviews? If that's a desirable attribute to maintain, perhaps I should move all the `.gitignore` changes in this PR here?

ruby/BUILD.bazel
JasonLunn: Good question. We don't have a Windows test configuration for CI today. Are there plans to introduce one?

ruby/README.md
mkruskal-google: We could change that :), but fair enough

mkruskal-google: Does ruby_library allow it?

ruby/BUILD.bazel
mkruskal-google: Not that I'm aware of, our Windows support is pretty limited.

ruby/.gitignore
mkruskal-google: Yea that sgtm

JasonLunn:
> > Looks like CI tests are failing?
> 
> Yep! The new test `implementation.rb` failed because FFI was not being properly loaded in these configs and was falling back to the native implementation. So yay, the test works!
> 
> Just pushed a fix which I expect will fix the test failure. 🤞

That look a lot more experimentation and fixing than anticipated, but all Ruby tests pass now. The breakage of the Java tests is unrelated - they're currently broken at HEAD, too.

ruby/README.md
JasonLunn: No - https://github.com/bazelruby/rules_ruby#ruby_library

ruby/.gitignore
JasonLunn: Done.

ruby/BUILD.bazel
JasonLunn: I'm happy to revisit this once we determine a Ruby + Windows test plan. Right now I'd have no way to confirm it works.

.gitignore
JasonLunn: Done.

mkruskal-google (approved):

ruby/README.md
```diff
@@ -51,6 +52,17 @@ puts MyTestMessage.encode_json(mymessage)
...
 ---------------------------------------
 
+Protocol Buffers uses the [ffi](https://github.com/ffi/ffi) gem to provide a
+new unified C-based implementation across Ruby interpreters based on
+[UPB](https://github.com/protocolbuffers/upb). For now, use of the FFI
+implementation is opt-in. If any of the following are true, the traditional
+platform-native implementations (MRI-ruby based on CRuby, Java based on JRuby)
+are used instead of the new FFI-based implementation:
+1. `ffi` and `ffi-compiler` gems are not installed
+2. `PROTOCOL_BUFFERS_RUBY_IMPLEMENTATION` environment variable has a value
```
haberman, line 62: Does this environment variable take effect at build time, runtime, or both?

ruby/README.md
```diff
@@ -51,6 +52,17 @@ puts MyTestMessage.encode_json(mymessage)
 Installation from Source (Building Gem)
 ---------------------------------------
 
+Protocol Buffers uses the [ffi](https://github.com/ffi/ffi) gem to provide a
```
haberman, line 55: Could we instead say "Protocol Buffers has an experimental backend which uses the ffi gem to provide a unified C-based implementation..."

"Protocol Buffers uses..." makes it sound like FFI is the default.

haberman (changes requested):

ruby/README.md
JasonLunn: Just at runtime.

ruby/README.md
JasonLunn: Updated.

ruby/README.md
JasonLunn: The build time flag (for Bazel only) is `//ruby:ffi=enabled`.

haberman: If the flag doesn't take effect at build time, it seems somewhat confusing to document it in a section about building gems.

What happens if you build a source gem with Rake instead of Bazel? Do FFI libraries get built?

JasonLunn: `gem build` includes FFI support unconditionally.

JasonLunn: The reason it is discussed here despite not being a build time variable is that I've added references to it in the test examples below. I'm attempting to reinforce the idea that just because it is built with FFI support doesn't mean that it is used all of the time.

haberman (approved):

mkruskal-google (approved):

protobuf_deps.bzl
```diff
@@ -110,8 +110,8 @@ def protobuf_deps():
         _github_archive(
             name = "rules_ruby",
             repo = "https://github.com/protocolbuffers/rules_ruby",
-            commit = "8fca842a3006c3d637114aba4f6bf9695bb3a432",
-            sha256 = "2619f9a23cee6f6a198d9ef284b6f6cbc901545ee9a9aac9ffa6b83dbf17cf0c",
+            commit = "b7f3e9756f3c45527be27bc38840d5a1ba690436",
```
CareF, line 113: https://github.com/protocolbuffers/rules_ruby/commit/b7f3e9756f3c45527be27bc38840d5a1ba690436 shows 
> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

It seems this commit is not part of the `main` branch is has been deleted

PR #13401, opened 2023-07-27
bazel: Get rid of exec_tools.

benjaminp:
Bazel has removed this attribute in bazelbuild/bazel@c061e57a7004a88eeb2f84d094d9a88b56c146b6.


perezd:
@mkruskal-google can you triage if/when this is safe to accept given our Bazel support requirements?

mkruskal-google:
If this passes tests today it should be safe

mkruskal-google (approved):

PR #13424, opened 2023-08-01
Fix pyi top level messages/enums in public imports

avelychko12:
According to PEP 484 https://peps.python.org/pep-0484/#stub-files
> Modules and variables imported into the stub are not considered exported from the stub unless the import uses the import ... as ... form or the equivalent from ... import ... as ... form.


anandolee:
Thank you to raise it up, I will port this change in our internal code base to prevent breaks of some internal tests that compare the generated stub code.

anandolee:
The change has been ported by https://github.com/protocolbuffers/protobuf/commit/b09aa967a65543c81b7b600e13f096d74fb64081

PR #13439, opened 2023-08-02
Breaking Change: protoc: require that reserved names are identifiers

### LINKED ISSUE #6335: protoc: reserved names are not validated

jhump:
This changes the current warnings into errors. In the previous PR (#10586), we downgraded these to warnings with the decision that they could be upgraded back to errors in a future major version.

Now that v24.0 is on the cusp of release, this seemed like a reasonable time to change it to error.

Resolves #6335 and #4558.

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -1793,10 +1793,11 @@ bool Parser::ParseReservedName(std::string* name,
   int col = input_->current().column;
   DO(ConsumeString(name, error_message));
   if (!io::Tokenizer::IsIdentifier(*name)) {
-    RecordWarning(
+    RecordError(
         line, col,
         absl::StrFormat("Reserved name \"%s\" is not a valid identifier.",
                         *name));
+    return false;
   }
   return true;
 }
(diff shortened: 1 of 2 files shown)


jhump:
@mkruskal-google, is now an appropriate time to make this change?

fowles:
We are planning some breaking changes for the January release, so I suspect the best time for this is immediately after the October release.

zhangskz:
I believe the grammar changes described in https://github.com/protocolbuffers/protobuf/issues/13440 isn't a breaking change and thus wouldn't need to wait until January to be released.

zhangskz:
Perhaps this change could be added to the PROTOBUF_FUTURE_BEAKING_CHANGES queue for the next breaking release? https://github.com/protocolbuffers/protobuf/blob/0681fe6278c3597a925986c19ff3dc5767f5f2fb/src/google/protobuf/port_def.inc#L231

mkruskal-google:
IIRC this is the grammar change we bundled into edition 2023, so closing this :)

PR #13443, opened 2023-08-02
Conformance test runner big endian fixes

jonathan-albrecht-ibm:
Add some missing endian conversions so that the conformance tests can be run on big endian platforms.

The message length value created by the conformance test runner is little endian according to the comments in the file but actually was sent in the native endianness of the host. I was able to run the java, python, ruby, php and csharp test executables and they all expect little endian length values so those tests would hang on big endian machines. Only the cpp test executable was using native endian so it has been changed to expect little endian too.

Also change the fixed32 and fixed64 functions in binary_json_conformance_test_suite.cc to send the data as little endian which fixes some failures in the python conformance tests on big endian platforms.

### DIFF
--- a/conformance/binary_json_conformance_suite.cc
+++ b/conformance/binary_json_conformance_suite.cc
@@ -35,6 +35,7 @@
 #include <utility>
 #include <vector>
 
+#include "google/protobuf/endian.h"
 #include "google/protobuf/text_format.h"
 #include "google/protobuf/util/json_util.h"
 #include "google/protobuf/util/type_resolver_util.h"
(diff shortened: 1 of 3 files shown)


zhangskz:
You'll need to rebase your change for the current test failures

jonathan-albrecht-ibm:
Thanks @zhangskz for the info. The changes rebased cleanly

jonathan-albrecht-ibm:
@zhangskz could you remove the `wait for user action` label when you get a chance?

acozzette (approved):

jonathan-albrecht-ibm:
Thanks @acozzette!

PR #13444, opened 2023-08-02
PHP: Fix read and write of float and double on big endian platforms

jonathan-albrecht-ibm:
Change the pack/unpack format codes 'f' -> 'g' and 'd' -> 'e' which ensure little endian format is used on the wire regardless of machine endianness. This allows the php composer and conformance tests to pass on big endian platforms and should not change any behavior on little endian platforms.

According to the PHP documentation, 'g' and 'e' were added in PHP versions 7.0.15 and 7.1.1.

### DIFF
--- a/php/src/Google/Protobuf/Internal/GPBWire.php
+++ b/php/src/Google/Protobuf/Internal/GPBWire.php
@@ -241,7 +241,7 @@ class GPBWire
         if (!$input->readRaw(4, $data)) {
             return false;
         }
-        $value = unpack('f', $data)[1];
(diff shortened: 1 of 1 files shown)


zhangskz:
This change also needs to be rebased to clear the test failures

jonathan-albrecht-ibm:
Thanks @zhangskz. The changes rebased cleanly

zhangskz (approved):

jonathan-albrecht-ibm:
Thanks @zhangskz!

Issue #13440, opened 2023-08-02
protoc: reserved names should be identifiers in the grammar, not strings


jhump:
As a follow on from #6335, the grammar for reserved names should be updated to support identifiers as values, not just strings:

```diff
- reserved = "reserved" ( ranges | strFieldNames ) ";"
+ reserved = "reserved" ( ranges | strFieldNames | fieldNames ) ";"
  ranges = range { "," range }
  range =  intLit [ "to" ( intLit | "max" ) ]
  strFieldNames = strFieldName { "," strFieldName }
  strFieldName = "'" fieldName "'" | '"' fieldName '"'
+ fieldNames = fieldName { "," fieldName }
```

Bare identifiers aren't as error-prone since there's no risk of forgetting a comma causing the parser to concatenate adjacent tokens into an unintended reserved name.

Eventually, it would be nice to deprecate or remove support for using string values, but removal would be a backwards-incompatible change.

fowles:
@Logofile for the doc part of this

jhump:
@fowles, @Logofile, just to be clear, this isn't a doc bug for the grammar, but a proposal to change the actual parser to accept identifiers (instead of string literals).

This was discussed in a tier 1 review on October 7th, 2022. IIUC, this part of the proposal (the grammar/parser change) was to be delayed until "edition zero". I opened this issue since it looks like "edition zero" is getting close and also for posterity, so this proposal is documented in GitHub.

Logofile:
Ah! That's the piece I was missing. I will add this to the list of things that need to be documented for Editions.

mkruskal-google:
This has been bundled into edition 2023, where we will enforce this new grammar and use prototiller to migrate

PR #13471, opened 2023-08-08
protoc: support identifiers as reserved names in addition to string literals (only in editions)

jhump:
This addresses #13440.

@mkruskal-google, I saw you are assigned the above issue. I hope it's okay that I took a stab at it.

### DIFF
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -869,6 +869,22 @@ TEST_F(ParseMessageTest, ReservedNames) {
       "}");
 }
 
+TEST_F(ParseMessageTest, ReservedIdentifiers) {
+  ExpectParsesTo(
+      "edition = \"2023\";\n"
+      "message TestMessage {\n"
+      "  reserved foo, bar;\n"
+      "}\n",
+
+      "syntax: \"editions\" "
+      "edition: \"2023\" "
+      "message_type {"
+      "  name: \"TestMessage\""
+      "  reserved_name: \"foo\""
+      "  reserved_name: \"bar\""
+      "}");
+}
+
 TEST_F(ParseMessageTest, ExtensionRange) {
   ExpectParsesTo(
       "message TestMessage {\n"
@@ -1227,6 +1243,24 @@ TEST_F(ParseEnumTest, ReservedNames) {
       "}");
 }
 
+TEST_F(ParseEnumTest, ReservedIdentifiers) {
+  ExpectParsesTo(
+      "edition = \"2023\";\n"
+      "enum TestEnum {\n"
+      "  FOO = 0;\n"
+      "  reserved foo, bar;\n"
+      "}\n",
+
+      "syntax: \"editions\" "
+      "edition: \"2023\" "
+      "enum_type {"
+      "  name: \"TestEnum\""
+      "  value { name:\"FOO\" number:0 }"
+      "  reserved_name: \"foo\""
+      "  reserved_name: \"bar\""
+      "}");
+}
+
 // ===================================================================
 
 typedef ParserTest ParseServiceTest;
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -869,6 +869,22 @@ TEST_F(ParseMessageTest, ReservedNames) {
       "}");
 }
 
+TEST_F(ParseMessageTest, ReservedIdentifiers) {
+  ExpectParsesTo(
+      "edition = \"2023\";\n"
+      "message TestMessage {\n"
+      "  reserved foo, bar;\n"
```
mkruskal-google, line 876: Can you add test points that combinations of identifier, string, and field number still work?

src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -1500,6 +1534,42 @@ TEST_F(ParseErrorTest, DuplicateJsonName) {
...
+TEST_F(ParseErrorTest, EnumReservedIdentifierCantMixWithStrings) {
+  ExpectHasErrors(
+      "edition = \"2023\";\n"
+      "enum TestEnum {\n"
+      "  FOO = 0;\n"
+      "  reserved foo, \"bar\";\n"
+      "}\n",
+      "3:16: Expected enum value identifier.\n");
+}
+
```
mkruskal-google, line 1572: These "CantMixWithStrings" tests seem redundant.  They're actually just testing the same thing as the tests above, since the identifier is invalid.

mkruskal-google:
Np, thanks Joshua!

mkruskal-google:
After discussing this a bit with the team, I think we can make this a stronger condition.  Under editions, we can *require* the new syntax, since there's a simple no-op transformation for prototiller to upgrade to during the edition migration

src/google/protobuf/compiler/parser_unittest.cc
mkruskal-google: Whoops, misunderstood what this test was doing.  Can you add tests for mixing field numbers as well?  IIUC those won't work

src/google/protobuf/compiler/parser_unittest.cc
mkruskal-google: Whoops, these *won't* work by design

src/google/protobuf/compiler/parser_unittest.cc
jhump: The error message is different: the identifier is not invalid here because these include `edition = "2023";` at the top.

src/google/protobuf/compiler/parser_unittest.cc
jhump: I'm not sure I understand. You are not allowed to combine numbers and names currently. Do you mean a test that makes sure this is still disallowed?

Or do you mean a test file that has separate statements, one with numbers and one with names?

mkruskal-google: I wasn't familiar enough with our reserved syntax, disregard :P

src/google/protobuf/compiler/parser_unittest.cc
mkruskal-google: Is this error because you mixed identifiers and strings or because you used string at all?  If it's the latter, we should remove the identifier from this test.  If it's the former, we should add a test that mixes identifiers and field numbers and a dedicated test for the latter

jhump: It's because it has a mix: since the first item was an _identifier_, it wants all subsequent items to be identifiers. It actually does this whenever there is a comma but the next token is not an identifier. (The existing functionality for numbers and strings is very similar: if the the next token is not a number of string literal, depending on context, it emits a similar message.)

I added the word "identifier" here, hoping to make it clear that an identifier is needed (in case they instead put a string, like in this example). I can also update existing messages for string literals to clarify they want a string literal.

jhump:
> After discussing this a bit with the team, I think we can make this a stronger condition. Under editions, we can require the new syntax, since there's a simple no-op transformation for prototiller to upgrade to during the edition migration

Ah, very good idea!

jhump:
@mkruskal-google, I think this is ready for another look. Editions now requires identifiers. The error message for when a user gets it backwards (e.g. tries to use an identifier in proto2 or proto3 or a string literal in editions) should be very clear now, too.

mkruskal-google (approved):

mkruskal-google:
Awesome lgtm, thanks for the contribution!

Issue #13442, opened 2023-08-02
protoc: message reserved tags allowed to be out of order


jhump:
The following file successfully compiles, even though the start and end tags for the reserved declaration are backwards:
```protobuf
// test.proto
message Foo { reserved 10 to 1; }
```
Running `protoc test.proto -o /dev/null` succeeds.

If I try to generate code, some plugins complain about this when processing the descriptors in the code generation request.
```
> protoc test.proto --go_out ./tmpgen --go_opt Mtest.proto=foo.bar/baz
protoc-gen-go: invalid FileDescriptorProto "test.proto": proto: message "Foo" reserved ranges has invalid range: 10 to 1
--go_out: protoc-gen-go: Plugin failed with status code 1.
```

Note that it is only reserved tags for messaged that exhibit this issue. If I do the same thing (swap start and end) in reserved numbers for enums _or_ in extension ranges, the compiler correctly complains:
```protobuf
// test.proto
message Foo { reserved 10 to 1; }
message Bar { extensions 10 to 1; }
enum Baz { BUZZ = 0; reserved 10 to 1; }
```
```
> protoc test.proto -o /dev/null
test.proto:3:26: Extension range end number must be greater than start number.
test.proto: Reserved range end number must be greater than start number.
```
(Interestingly, there is no line and column reported for that second error, which suggests the error is happening when the descriptor is added to the compiler's pool instead of during parse. Whereas the error for the extension range does include line and column details.)

fowles:
happy to call this a bug.  If you want to send a PR for this we will accept it

PR #13474, opened 2023-08-08
protoc: validate that reserved range start is before end

jhump:
This addresses #13442. This also backfills the tests to add tests for the two checks that were already implemented as well as one for the newly added check.

Finally, this fixes the location information so that positions for reserved ranges are correctly reported. (The previous check that already existed, for enums, failed to show line and column info.)


fowles (approved):

fowles:
this is going to take a few days, I need to cleanup google3.

Issue #13441, opened 2023-08-02
protoc: explicit map_entry option is sometimes allowed


jhump:
In the following example source file, `protoc` correctly produces an error:
```protobuf
// test.proto
message Foo { option map_entry = true; reserved 10 to 1; }
message Bar { optional Foo foo = 1; }
```
```
> protoc test.proto -o /dev/null
test.proto:3:24: map_entry should not be set explicitly. Use map<KeyType, ValueType> instead.
```
Interestingly, the reported position in the error message is where the message is _used_, not where it is defined.

**However**, if the file does not actually refer to the offending message type (`Foo`), the compiler _accepts_ the invalid input:
```protobuf
// test.proto
message Foo { option map_entry = true; reserved 10 to 1; }
```
With the above, running `protoc test.proto -o /dev/null` produces no error.

This is inconsistent, but it also leads to "errors at a distance". If I define a file elsewhere that uses this message, I get an error in that other compile operation, instead of in the original:
```protobuf
// test2.proto
import "test.proto";
message Bar { optional Foo foo = 1; }
```
```
> protoc test2.proto -o /dev/null
test2.proto:3:24: map_entry should not be set explicitly. Use map<KeyType, ValueType> instead.
```

PR #13479, opened 2023-08-09
protoc: parser rejects explicit use of map_entry option

jhump:
This addresses #13441.

This preserves the similar check at the _point of use_ of invalid messages in `DescriptorBuilder` (and there's an existing test that verifies that check still works).

But it adds another check in the parser, to catch this error at the _point of definition_ of an invalid message. And the corresponding test is updated: we no longer need a usage of the message to catch the error, and the reported position is the definition of the option, not the usage site of the message.

The way this works feels kinda gross, but I wasn't sure of a better way to do it. The only place we know for certain that it was an explicit option (vs. auto-added by the parser when synthesizing a map entry message) is when after processing the message body, we can look at the uninterpreted options. So that's what this does. If you have ideas on better/cleaner approaches, I'd be happy to revise.

### DIFF
--- a/src/google/protobuf/compiler/parser.cc
+++ b/src/google/protobuf/compiler/parser.cc
@@ -931,6 +931,26 @@ bool Parser::ParseMessageBlock(DescriptorProto* message,
   if (message->reserved_range_size() > 0) {
     AdjustReservedRangesWithMaxEndNumber(message);
   }
+
+  // check if map_entry option was explicitly set
+  if (message->has_options()) {
+    const MessageOptions& opts = message->options();
+    for (int i = 0; i < opts.uninterpreted_option_size(); i++) {
+      const UninterpretedOption& opt = opts.uninterpreted_option(i);
+      if (opt.name_size() > 0 && !opt.name(0).is_extension()
+          && opt.name(0).name_part() == "map_entry") {
+        int line = -1, col = 0;
+        if (source_location_table_ != nullptr) {
+           source_location_table_->Find(&opt,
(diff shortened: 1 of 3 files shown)


src/google/protobuf/compiler/parser.cc
```diff
@@ -931,6 +931,26 @@ bool Parser::ParseMessageBlock(DescriptorProto* message,
   if (message->reserved_range_size() > 0) {
     AdjustReservedRangesWithMaxEndNumber(message);
   }
+
+  // check if map_entry option was explicitly set
```
fowles, line 935: rather than a comment to section off part of the message, please use a well named helper

jhump: Done. I've added a `ValidateMessage` helper, analogous to existing `ValidateEnum`. The only check it performs is the check for an explicit `map_entry` option.

WDYT? Too generic?

fowles (approved):

src/google/protobuf/compiler/parser.cc
fowles: nah, looks about right.

jhump:
@fowles, so does that mean that there are places in the google3 codebase that were using explicit `map_entry` options? And the other failing build means reserved ranges where start and end were reversed?

BTW, sorry if this makes more work for your team. These are obviously not urgent fixes. And thanks for your help in shepherding these fixes into the codebase!

fowles:
yes, the cleanups are simple so I should be able to slice them in this week

PR #13492, opened 2023-08-09
port_def: Ignore -Wpedantic warnings

jonaski:
Fixes `extra ';' [-Wpedantic]` warnings in generated `.pb` headers.

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -1087,6 +1087,8 @@ static_assert(PROTOBUF_ABSL_MIN(20230125, 3),
 #pragma GCC diagnostic ignored "-Wstringop-overflow"
 // This causes spurious warnings in GCC 13.
 #pragma GCC diagnostic ignored "-Wself-move"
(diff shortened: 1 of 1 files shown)


src/google/protobuf/port_def.inc
```diff
@@ -851,7 +851,7 @@ static_assert(PROTOBUF_ABSL_MIN(20230125, 3),
 #endif  // PROTOBUF_TSAN
 
 #ifdef PROTOBUF_TSAN
-#define PROTOBUF_TSAN_DECLARE_MEMBER ::uint32_t _tsan_detect_race = 0
+#define PROTOBUF_TSAN_DECLARE_MEMBER ::uint32_t _tsan_detect_race = 0;
 #else
 #define PROTOBUF_TSAN_DECLARE_MEMBER
```
fowles, line 856: instead of pushing this complexity into the macro and code base, can you simply add -Wpedantic to the set of warnings we suppress in port_def.inc?

jonaski: Probably

fowles (approved):

Issue #13482, opened 2023-08-09
Protoc generates faulty C# namespace


nickwinger:
Hello,

i'm not sure if this is a bug or how to fix this, haven't found anything on the internet.
We have a package like "my.name.space._3".
This works in protobuf and java generated code.
However in C# generated code the unserscore gets removed, and then the C# Compiler tells an error about that namespace "my.name.space.3".
Seems like only having a number in the namespace in C# is not working.

So how to handle this issue ? And we cannot change the protobuf files, since they are delivered by 3rd party.

C:\protoc --version
libprotoc 24.0

actually it should have been fixed: https://github.com/protocolbuffers/protobuf/pull/10268

Thanks for help,
Nick

jskeet:
Please could you post your specific .proto file? (It's not clear what you mean by "namespace" here - protobuf package, or csharp_namespace option. If you want a specific C# namespace, the simple fix would be to specify it explicitly.)

nickwinger:
Hello,

i mean package in proto file, e.g.:

package my.scope._3

in C#:

namespace my.scope.3

the underscore gets removed which does not work in C#.
The bugfix i linked seem to only preserve the underscore before numbers at the beginning, but the routine has to check and preserve the underscore before numbers on every part of the package/namespace string separated by dots.

nickwinger:
Thank you for the suggestion, i will try it out as a workaround. However it’s not the best solution as we have to modify and scan the delivered 3rd party proto files

nickwinger:
I will try to do a PR based on the fix that has already been made

nickwinger:
here is the PR: https://github.com/protocolbuffers/protobuf/pull/13504

jskeet:
We still intend to address this, but it's unlikely to be in 2023 due to higher-priority issues.

jskeet:
(#13504 is merged, so presumably this is at least better than it was - but I believe there are still improvements we want to make.)

PR #13504, opened 2023-08-10
BugFix for issue 13482

nickwinger:
https://github.com/protocolbuffers/protobuf/issues/13482

### DIFF
--- a/src/google/protobuf/compiler/csharp/names.cc
+++ b/src/google/protobuf/compiler/csharp/names.cc
@@ -184,6 +184,15 @@ std::string UnderscoresToCamelCase(absl::string_view input,
     } else if ('0' <= input[i] && input[i] <= '9') {
       result += input[i];
       cap_next_letter = true;
+    // https://github.com/protocolbuffers/protobuf/issues/13482
+    // Preserve underscores before numbers if the come right after a dot
+    } else if (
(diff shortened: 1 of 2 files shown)


nickwinger:
@jskeet , @tonydnewell here is the PR for the Bug

src/google/protobuf/compiler/csharp/names.cc
```diff
@@ -184,6 +184,15 @@ std::string UnderscoresToCamelCase(absl::string_view input,
     } else if ('0' <= input[i] && input[i] <= '9') {
       result += input[i];
       cap_next_letter = true;
+    // https://github.com/protocolbuffers/protobuf/issues/13482
+    // Preserve underscores before numbers if the come right after a dot
+    } else if (
```
jskeet, line 189: Drive-by comment - I haven't analysed this fully - what would this do with "my.scope.__3"? I *suspect* we'd still end up with "my.scope.3", which presumably isn't intended.

nickwinger: yes you are right, there is room for improvement, however my fix doesn't break anything and it still is an improvement. double underscore handling should be another PR as it makes things even more complicated. however most of the time when people only use one underscore it is an improvement

PR #13554, opened 2023-08-15
Update collect_all_artifacts.sh to remove C# Google.Protobuf.Tools nuget build

zhangskz:
CI shows this part of legacy release infrastructure is no longer working as expected due to failing keyserver request.

C# has already exited support window for 21.x so we are removing this artifact from releases on this line.


mkruskal-google (approved):

zhangskz:
@mkruskal-google Seems to be waiting on a "Linux C++ TC Malloc" Kokoro test which still seems to exist as a presubmit. Don't think it affects this PR, but is this something we should 1) revive if it should be running but isn't, or 2) remove from branch protection?

zhangskz:
(NVM actually, I think it still works but is just slow. I forgot that Kokoro jobs take a while)

PR #13608, opened 2023-08-19
pyproject.toml

rinarakaki:


python/setup.py
```diff
@@ -429,44 +393,12 @@ def HasLibraryDirsOpt():
...
-      ],
-      namespace_packages=['google'],
-      packages=find_packages(
-          exclude=[
-              'import_test_package',
-              'protobuf_distutils',
-          ],),
       test_suite='google.protobuf.internal',
       cmdclass={
-          'clean': CleanCmd,
```
anandolee, line 464: why this is removed?

rnarkk: Hi Jie, this setup by default will clean up some type of build-time only artefacts, especially applicable to those defined in the CleanCmd by their file extension. I’d be happy to hear your feedback further.

rnarkk:
Could someone at least approve this to let me see if all the tests pass and integration works?

PR #13805, opened 2023-08-31
examples: fix deprecated ioutil package

profnandaa:
This commit updates the deprecated ioutil functions (Write/ReadFile) to their os equivalents.

Additionally, also sets to ignore `*.pb.go` files that are generated by the Make commands (protoc compilations).

### DIFF
--- a/examples/.gitignore
+++ b/examples/.gitignore
@@ -1,2 +1,4 @@
 # Ignore the bazel symlinks
 /bazel-*
+# ignore protoc generated files
+*.pb.go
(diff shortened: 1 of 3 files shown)


examples/.gitignore
```diff
@@ -1,2 +1,4 @@
 # Ignore the bazel symlinks
 /bazel-*
+# ignore protoc generated files
+*.pb.go
```
anandolee, line 4: why this need to be ignored?

profnandaa: Since it's auto-generated from the  .proto file. This is only as far as the /examples directory is concerned.

PR #13812, opened 2023-08-31
Add versions files to hold Protobuf implementations versions.

shaod2:
This will prepare for adding version numbers to Protobuf gencodes.

As a followup, update_version.py will be enhanced to catch the version updates to versions.h. After that, code generators can adopt them into the gencode.

### DIFF
--- a/src/google/protobuf/versions.h
+++ b/src/google/protobuf/versions.h
@@ -36,21 +36,29 @@
 #include "absl/strings/string_view.h"
 
 // Macros to be used by language generators to insert version information.
+// Although "main" is not a normal version suffix, it would be helpful for
+// code generators to discern the main version.
+//
+// Currently, it holds language versions only for C++, Java and Python,
+// but can be extended to other Protobuf languages.
+//
+// Usually, they should be updated automatically by Protobuf release
+// process.
 #define PROTOBUF_CPP_VERSION 4024000
-#define PROTOBUF_CPP_VERSION_SUFFIX "main"
+#define PROTOBUF_CPP_VERSION_SUFFIX "-main"
 
 #define PROTOBUF_JAVA_VERSION 3024000
-#define PROTOBUF_JAVA_VERSION_SUFFIX "main"
+#define PROTOBUF_JAVA_VERSION_SUFFIX "-main"
 
 #define PROTOBUF_PYTHON_VERSION 4024000
-#define PROTOBUF_PYTHON_VERSION_SUFFIX "main"
+#define PROTOBUF_PYTHON_VERSION_SUFFIX "-main"
 
 namespace google {
 namespace protobuf {
 namespace internal {
 
-// Gets the version string with suffix. For example, "4.24.0-dev".
-// version must be an integer in the form of
+// Gets the version string with version number and suffix.
+// For example, "4.24.0-dev". version must be an integer in the form of
 // {1 digit major}0{2 digits minor}0{2 digits micro}.
 // suffix can be empty or prefixed by "-".
 std::string VersionString(int version, absl::string_view suffix);
--- a/src/google/protobuf/versions.h
+++ b/src/google/protobuf/versions.h
@@ -1,0 +1,62 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2023 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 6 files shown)


zhangskz:
This will need to land in Piper first anyways so you could always develop / review there and let Copybara sync it back to Github?

src/google/protobuf/versions.h
```diff
@@ -0,0 +1,70 @@
...
+// Macros to be used by language generators to insert version information.
+// Although "main" is not a normal version suffix, it would be helpful for
+// code generators to discern the main version.
+//
+// Currently, it holds language versions only for C++, Java and Python,
+// but can be extended to other Protobuf languages.
+//
+// Usually, they should be updated automatically by Protobuf release
+// process.
+#define PROTOBUF_CPP_VERSION 4024000
```
zhangskz, line 47: Could this replace the cpp version in port_def.inc?

src/google/protobuf/versions.h
```diff
@@ -0,0 +1,70 @@
...
+#define PROTOBUF_CPP_VERSION_SUFFIX "-main"
```
zhangskz, line 48: Will these ever diverge between languages, or could a single suffix be sufficient?

shaod2:
> This will need to land in Piper first anyways so you could always develop / review there and let Copybara sync it back to Github?

I think it would make more sense to me for the versions macros defined here

src/google/protobuf/versions.h
mkruskal-google: +1, I don't think we want three of these lying around

shaod2: Yes we have PROTOBUF_VERSION in port_def.inc and GOOGLE_PROTOBUF_VERSION in common.h. For simplicity, I'd like to avoid them as they're complicatedly used interchangeably as for protoc version and c++ version. We might want to clean them up later.

shaod2: Okay, maybe taking a step back while not blocking on Java and Python; how about removing CPP and keeping Java and Python. I'll come up with another change to clean them up

mkruskal-google, line 47: We should probably do this from piper so we can update the release script at the same time

src/google/protobuf/versions_test.cc
```diff
@@ -0,0 +1,50 @@
...
+
+#include <gtest/gtest.h>
+
+
+namespace google {
+namespace protobuf {
+namespace internal {
+
+TEST(VersionsTest, VersionStringWithEmptySuffix) {
+  EXPECT_EQ(VersionString(4024000, ""), "4.24.0");
```
mkruskal-google, line 41: Some other test cases that might show issues:
10.1.1
4.24.100
4.100.0

shaod2:
Let's close this PR.

It sounds like we're leaning towards starting a CL from Piper and also updating update_version.py

src/google/protobuf/versions.h
```diff
@@ -0,0 +1,70 @@
...
+
+#define PROTOBUF_JAVA_VERSION 3024000
```
googleberg, line 50: Should these all be in the same versions.h file? Or should they be in cpp/version.h, java/version.h, python/version.h?

The answer might be "in the same file for now, different files later, because..."

I'm thinking in terms of seeing each generator plugin as a separate entity.

src/google/protobuf/versions.h
```diff
@@ -0,0 +1,70 @@
...
+
+namespace google {
+namespace protobuf {
+namespace internal {
+
+// Gets the version string with version number and suffix.
+// For example, "4.24.0-dev". version must be an integer in the form of
+// {1 digit major}0{2 digits minor}0{2 digits micro}.
+// suffix can be empty or prefixed by "-".
+std::string VersionString(int version, absl::string_view suffix);
```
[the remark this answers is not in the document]
shaod2: yes

I see benefits of defining constant version string, but if so that will be inconsistent with PROTOBUF_VERSION and GOOGLE_PROTOBUF_VERSION and the migration (if needed) off of them will look more complex. For one thing, these 2 existing macros are used for comparisons directly (in ignorance of suffix string).

src/google/protobuf/versions.h
shaod2: I'm totally fine with either way, but I don't really know the end state of the C++ built-in code generators (i.e. Java Python C++). If they will be totally separated plugins, do they still share common proto compiler utilities? If so, defining in a central file seems to be fine and align with version.json.

src/google/protobuf/versions.h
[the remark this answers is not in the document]
googleberg: You would keep the compile-time macros.
I was just suggesting the string constants instead of the function to "build" the version string.

src/google/protobuf/versions.h
googleberg: I think you have a good point. Let's keep them in one file for now.

PR #13818, opened 2023-09-01
Fix TypeError when passing an instance of a subclass of String to a string field

NotFounds:
## Issue
When an object that is an instance of a string-derived class is passed to a string field in a protobuf message in Ruby, it results in a ﻿`Google::Protobuf::TypeError`.

### Steps to reproduce
```rb
~/src/github.com/protocolbuffers/protobuf/ruby/tests ❯❯❯ irb -I .
irb(main):001:0> require 'basic_test_pb'
=> true
irb(main):002:0> myString = Class.new(String)
=> #<Class:0x00000001531540d8>
irb(main):003:0> str = myString.new("foo")
=> "foo"
irb(main):004:0> BasicTest::TestMessage.new(optional_string: "foo")
=> <BasicTest::TestMessage: optional_string: "foo", repeated_int32: [], repeated_int64: [], repeated_uint32: [], repeated_uint64: [], repeated_bool: [], repeated_float: [], repeated_double: [], repeated_string: [], repeated_bytes: [], repeated_msg: [], repeated_enum: []>
irb(main):005:0> BasicTest::TestMessage.new(optional_string: str)
(irb):5:in `initialize': Invalid argument for string field 'optional_string' (given #<Class:0x00000001531540d8>). (Google::Protobuf::TypeError)
irb(main):006:0>
```

## Fix
The issue appears to be caused by the field checking mechanism not properly handling instances of classes that inherit from basic types like String. My proposed solution is to improve the type checking for string fields to consider not just ﻿String instances but also instances of subclasses of ﻿String.

## Impact
The changes will allow instances of classes derived from ﻿String to be passed to string fields without any error.
This is a backwards-compatible change and will not affect the existing behaviour with standard ﻿String instances.


JasonLunn:
@anandolee - are you able to complete the CLA?

ruby/ext/google/protobuf_c/convert.c
```diff
@@ -164,7 +164,7 @@ upb_MessageValue Convert_RubyToUpb(VALUE value, const char* name,
       VALUE utf8 = rb_enc_from_encoding(rb_utf8_encoding());
       if (rb_obj_class(value) == rb_cSymbol) {
         value = rb_funcall(value, rb_intern("to_s"), 0);
-      } else if (rb_obj_class(value) != rb_cString) {
+      } else if (rb_obj_class(value) != rb_cString && !rb_obj_is_kind_of(value, rb_cString)) {
```
JasonLunn, line 167: Are both terms of the conditional check needed? `'foo'.kind_of?(String)` is `true`, so doesn't it cover both the case where `value` is a `String` and cases where it is is a subclass?

ruby/tests/basic.rb
```diff
@@ -711,4 +711,15 @@ def test_oneof_fields_respond_to? # regression test for issue 9202
     assert_respond_to msg, :has_d?
     refute msg.has_d?
   end
+
+  def test_string_subclass
```
JasonLunn, line 715: This test doesn't exercise the same case as the description - `str` is a normal `String` and it doesn't look like any instances of `myString` are every created.

JasonLunn (changes requested):

ruby/ext/google/protobuf_c/convert.c
NotFounds: You're right. 
✅ [2a6b513](https://github.com/protocolbuffers/protobuf/pull/13818/commits/2a6b513d81ff1d028987e98871ff0204e5bbea31)

ruby/tests/basic.rb
NotFounds: Oops! I was wrong.
Fixed it in [7dad6cf](https://github.com/protocolbuffers/protobuf/pull/13818/commits/7dad6cfd187bceacc53625a6a8698aec97f1c14b)

JasonLunn:
I think the errors observed in the test runs would be resolved if this branch is rebased or if you merge from `main`.

JasonLunn (approved):

NotFounds:
@JasonLunn Thanks for the quick review! Really appreciate your help.

NotFounds:
@zhangskz 
I've merged the main branch. Could you please check?

zhangskz (approved):

PR #13838, opened 2023-09-05
Improve .NET debugging of Protobuf messages

JamesNK:
Generated .NET protobuf messages override `ToString` and return JSON. By default, the .NET debugger displays the result from `ToString` and wraps it in curly braces, e.g. `{{ "Message": "Hello world" }}`. The double curly braces is an ugly result. People expecting JSON here are confused.

This PR adds a [DebuggerDisplayAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggerdisplayattribute?view=net-7.0) to generated messages. The attribute tells the debugger to not quote the ToString value with `nq`.

Before:
image

After:
image

I haven't touched the protoc compiler in a long time. I'm guessing I need to build it and then run it to regenerate the checked-in code. I don't suppose someone else could do that? It would save me a lot of time 😬

### DIFF
--- a/src/google/protobuf/compiler/csharp/csharp_message.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_message.cc
@@ -123,6 +123,7 @@ void MessageGenerator::Generate(io::Printer* printer) {
   AddDeprecatedFlag(printer);
   AddSerializableAttribute(printer);
 
+  printer->Print("[global::System.Diagnostics.DebuggerDisplay(\"{ToString(),nq}\")]\n");
   printer->Print(
     vars,
     "$access_level$ sealed partial class $class_name$ : ");
--- a/src/google/protobuf/compiler/csharp/csharp_message.cc
+++ b/src/google/protobuf/compiler/csharp/csharp_message.cc
@@ -123,7 +123,7 @@ void MessageGenerator::Generate(io::Printer* printer) {
   AddDeprecatedFlag(printer);
   AddSerializableAttribute(printer);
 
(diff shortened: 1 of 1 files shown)


JamesNK:
cc @jtattermusch @jskeet

jskeet (commented):
I'm in favor of this in general (although I'll need to read up on the attribute) - I'm going to check what impact it has on code size though. Last time I did some analysis, just DebuggerNonUserCodeAttribute and GeneratedCodeAttribute accounted for a surprisingly large proportion of the code size (in Google.Cloud.Compute.V1 as a sample).

JamesNK:
This attribute is only on messages. The impact should be much less than attributes placed on all members.

Off topic: Is `DebuggerNonUserCodeAttribute` and `GeneratedCodeAttribute` required on every property? I would have thought that putting them on the class would infer that everything in the class is generated.

jskeet:
> This attribute is only on messages. The impact should be much less than attributes placed on all members.

Ah, whoops - I'd missed that, thinking it was on all the properties. Doh. Yes, that should be fine then.

> Off topic: Is DebuggerNonUserCodeAttribute and GeneratedCodeAttribute required on every property? I would have thought that putting them on the class would infer that everything in the class is generated. Edit: I noticed the message type doesn't have those attributes. Is that because the type can be extended with partial?

I honestly can't remember, to be honest. I suspect you're right that it's due to them being partial - you wouldn't want to not be able to debug into the user code.

jskeet (approved):
Tested locally (not the actual generator, but manually adding the attribute to observe the difference) - looks good.

jskeet:
Ping @deannagarcia

mkruskal-google (approved):

PR #13882, opened 2023-09-06
Python: stop using backward compatiblity timezones

aurel32:
The 'Japan' and 'US/Pacific' timezones used in Python tests are backward compatibility timezones (defined in the backward file of tzdata) and might not be available on all systems. For instance Debian recently split those timezones in a tzdata-legacy package.

Update the tests to the use the canonical name of the timezones instead, that is respectively 'Asia/Tokyo' and 'America/Los_Angeles'.


fowles (approved):

fowles:
@aurel32 there are conflicts blocking this PR

PR #10404, opened 2022-08-14
Use hex representation for binary data inside generated php files

Grebenschikov:
For the sake of performance, the hex2bin call was removed from the generated PHP code in #8006
However, after this PR all autogenerated files contain binary data, which makes it hard or even impossible to see using common tools (git and github included (i.e. [this file](https://github.com/protocolbuffers/protobuf/blob/main/php/src/GPBMetadata/Google/Protobuf/Struct.php)), as well as some code editors) since most software considers such files as binary ones and not as code.

At the same, there is a possibility to use hex representation of binary data inside PHP files, which doesn't affect the performance (because it compiles to the same OP codes by the interpreter).

```
package ~/Desktop/proto $ cat test1.php
<?php

$str = "teststring";
echo $str;%
package ~/Desktop/proto $ phpdbg -p\* test1.php

$_main:
     ; (lines=3, args=0, vars=1, tmps=1)
     ; /Users/package/Desktop/proto/test1.php:1-4
L0003 0000 ASSIGN CV0($str) string("teststring")
L0004 0001 ECHO CV0($str)
L0004 0002 RETURN int(1)
[Script ended normally]
package ~/Desktop/proto $ cat test2.php
<?php

$str = "\x74\x65\x73\x74\x73\x74\x72\x69\x6e\x67";
echo $str;%
package ~/Desktop/proto $ phpdbg -p\* test2.php

$_main:
     ; (lines=3, args=0, vars=1, tmps=1)
     ; /Users/package/Desktop/proto/test2.php:1-4
L0003 0000 ASSIGN CV0($str) string("teststring")
L0004 0001 ECHO CV0($str)
L0004 0002 RETURN int(1)
[Script ended normally]
```

So, this PR replaces binary data with hex representation to fix issues mentioned above

PR #13911, opened 2023-09-08
Do not store binary data inside php files

mikhainin:
Follow up on #10404, the initial description:

> For the sake of performance, the hex2bin call was removed from the generated PHP code in https://github.com/protocolbuffers/protobuf/pull/8006
However, after this PR all autogenerated files contain binary data, which makes it hard or even impossible to see using common tools (git and github included (i.e. [this file](https://github.com/protocolbuffers/protobuf/blob/main/php/src/GPBMetadata/Google/Protobuf/Struct.php)), as well as some code editors) since most software considers such files as binary ones and not as code.

Alexander suggested using a hex representation of string literals and I updated the original patch a bit, not hex-encoding printable characters. 

Benchmarks from [#10404#issuecomment-1635939062](https://github.com/protocolbuffers/protobuf/pull/10404#issuecomment-1635939062)
```
The percent of printable chars: 0.83869 (80112 vs 15408)
hex parsing: 3.22371, length=382090
hex2Bin parsing: 1.18489, length=191059
hexAscii parsing: 0.89524, length=142306 (suggested option)
binary parsing: 0.26437, length=95542
```


haberman:
Some tests are failing, and they look unrelated to your change. Could you merge with main and re-push?

mikhainin:
Apologies for this. I rebased from `main` and force-pushed the branch

mikhainin:
Found one more issue with binary data.

When the repository has settings in `.gitattributes`:
```
*.php text eol=lf
```

git will break protobuf descriptors and will give developers hard debugging times. The same effect will be achieved if `git config core.autocrlf` is set to **true**.

Not sure if we need to make it a separate issue or no: it definitely worsens developer experience but probably not directly relates to this PR.

haberman (approved):

PR #14025, opened 2023-09-11
cmake: don't consider include paths on different drives

### LINKED ISSUE #9290: Wrong relative path in protobuf_generate
https://github.com/protocolbuffers/protobuf/blob/88384c99ec9621c0054e24bc80cab7709f550357/cmake/protobuf-config.cmake.in#L111
Under Windows, the command `file(RELATIVE_PATH _rel_dir ${DIR} ${_abs_dir})` returns an absolute path when the include path and the file path are on different drives.
Sample: called with $DIR=C:/include_path and _abs_dir = D:/current_repository/, the command returns an absolute path without any `..` in it.
The `_is_in_parent_folder` is set to false and therefore a wrong directory is accepted as relative path.
The result is that `_rel_dir` points to an absolute path which generates bad pathnames line 126

Tested with cmake 3.21

cgudrian:
The current logic of finding a suitable include path relies on CMake to always return a meaningful relative path. On Windows, however, if the two paths reside on different drives (e.g. `c:/Users/Me` and `d:/project/`) the absolute path gets returned unaltered. This yields the invalid final path `c:/Users/Me/d:/project/`.

This case is now detected and the corresponding include path is deemed unsuitable.

(fixes #9290)

### DIFF
--- a/cmake/protobuf-generate.cmake
+++ b/cmake/protobuf-generate.cmake
@@ -108,6 +108,12 @@ function(protobuf_generate)
     foreach(DIR ${_protobuf_include_path})
       if(NOT DIR STREQUAL "-I")
         file(RELATIVE_PATH _rel_dir ${DIR} ${_abs_dir})
+        if(_rel_dir STREQUAL _abs_dir)
+          # When there is no relative path from DIR to _abs_dir (e.g. due to
+          # different drive letters on Windows), _rel_dir is equal to _abs_dir.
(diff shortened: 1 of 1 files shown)


fowles:
@cgudrian this PR is blocked on signing the CLA

cgudrian:
I've already filed the Google Individual CLA on September 11. Is there anything else I need to do?

cgudrian:
Ok. Found the _rescan_ button.

fowles (approved):

PR #14097, opened 2023-09-15
Add .NET debugging attributes

JamesNK:
I've been working with a large Protobuf model and noticed debugging issues that can easily be improved.

Add debugging attributes to collections, `ByteString`, and descriptors. No impact on runtime behavior. Debugger type proxies are what the debugger displays by default, but full data is available by selecting "Raw View".

`RepeatedField` and `MapField` now display their item count. This is standard across .NET collections. For example, .NET's list and dictionary both display `Count = {Count}`.

Note that previously, Protobuf collections displayed the result of `ToString`, which returned JSON. The JSON debug view isn't useful when there is too much content to display at once because it's truncated. That experience will be fairly common.

Count benefits:
* Always useful
* Makes debugging Protobuf collections feel more like regular .NET collections
* Collection contents is now easier to access with debugger type proxies

No dependency between this PR and https://github.com/protocolbuffers/protobuf/pull/13838. Each can be merged independently.

**MapField before:**

image

**MapField after:**

image

**RepeatedField before:**

image

**RepeatedField after:**

image

**ByteString before:**

image

**ByteString after:**

image

**Descriptor before:**

image

**Descriptor after:**

image

**TypeRegistry before:**

image

**TypeRegistry after:**

image

### DIFF
--- a/csharp/src/Google.Protobuf/ByteString.cs
+++ b/csharp/src/Google.Protobuf/ByteString.cs
@@ -10,6 +10,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Security;
@@ -23,6 +24,8 @@ namespace Google.Protobuf
     /// Immutable array of bytes.
     /// </summary>
     [SecuritySafeCritical]
+    [DebuggerDisplay("Length = {Length}")]
+    [DebuggerTypeProxy(typeof(ByteStringDebugView))]
     public sealed class ByteString : IEnumerable<byte>, IEquatable<ByteString>
     {
         private static readonly ByteString empty = new ByteString(new byte[0]);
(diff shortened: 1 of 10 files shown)


JamesNK:
cc @jtattermusch

jskeet:
Will look at this tomorrow - I've got Covid so I'm working more slowly in general. Will get to it eventually...

csharp/src/Google.Protobuf/ByteString.cs
```diff
@@ -23,6 +24,8 @@ namespace Google.Protobuf
     /// Immutable array of bytes.
     /// </summary>
     [SecuritySafeCritical]
+    [DebuggerDisplay("Length = {Length}")]
```
JamesNK, line 27: Note: `ByteString` displays length rather than count to act more like array/span/memory.

jskeet (approved):
Love it! I particularly like the fact that there's no need to change the generated code at all :)

mkruskal-google (approved):

PR #14174, opened 2023-09-22
Remove <iostream> from the common.h header

serge-sans-paille:
<iostream> embeds a global constructor (to initialize std::cout and such), typically `static ios_base::Init __ioinit;` in libstdc++).

This header is not directly necessary in the common header and has a small impact on every compilation unit that includes it.

As an example, removing that header dependency in turns suppresses 33 global constructors from the firefox core library libxul.


serge-sans-paille:
@fowles I've been added to Mozilla's allowlist for google contribution, that part should be ok now.

mkruskal-google (approved):

PR #14197, opened 2023-09-25
Support non ascii commandline args

hknielsen:
Using none ASCII chars for protoc on Windows are not handled well.
This adresses argument file at a location where either directory, or filename contains Unicode chars.

Specifically because MsBuild tool saves argument file in with the UserName appended to the .rsp file.
https://github.com/dotnet/msbuild/pull/9232

But in general we should really handle if none ascii chars are in the path.

### DIFF
--- a/build_defs/cpp_opts.bzl
+++ b/build_defs/cpp_opts.bzl
@@ -35,6 +35,7 @@ LINK_OPTS = select({
     "//build_defs:config_msvc": [
         # Suppress linker warnings about files with no symbols defined.
         "-ignore:4221",
+        "/utf-8",
     ],
     "@platforms//os:macos": [
         "-lpthread",


hknielsen:
@haberman im not sure how to fix the "unsafe fork" issue, do I need to do anything?

fowles:
Instead of shifting the code to wchar's can we shift the environment to utf-8 as outlined in https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page ?

hknielsen:
@fowles thanks for the question! That let me on to this:
https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170
I agree its much better than starting to go to `wchar`. Changed so we on windows compile

PR #14230, opened 2023-09-27
Adding DLL export/import tags to generated public upb API

dawidcha:
I have been collaborating with the grpc developers to make it possible to build that library as a Windows DLL - a couple of PRs were already merged, more like [this one](https://github.com/grpc/grpc/pull/34345) are pending.

The grpc library incorporates some upb-generated, and upbdefs-generated code into grpc.dll, which is referenced by other code that consumes the library. Since this is now a DLL, that code doesn't know how to link to these generated symbols because they are not annotated with __declspec(dllimport).

This PR aims to fix that by introducing a parameter 'dllexport_tag' to the upb and upbdefs plugins. That parameter should be a string e.g. MYAPP_DLL and when set, the extern symbols are annotated with a macro with that name. This can either be set externally to __declspec(dllimport) or, as is usual practice, when compiling code into a DLL, the macro <dllexport_tag>_EXPORT (i.e. MYAPP_DLL_EXPORT) is defined, and when consuming the DLL <dllexport_tag>_IMPORT is defined if neither are defined then the MYAPP_DLL macro becomes empty string which is what you want for building a static library.

### DIFF
--- a/upb_generator/bootstrap_compiler.bzl
+++ b/upb_generator/bootstrap_compiler.bzl
@@ -67,7 +67,7 @@ def _stage0_proto_staleness_test(name, base_dir, src_files, src_rules, strip_pre
             "$(location " + _protoc + ") " +
             "-I$(GENDIR)/" + strip_prefix + " " + _extra_proto_path +
             "--plugin=protoc-gen-upb=$(location " + _upbc("upb", 0) + ") " +
-	    "--upb_out=dllexport_decl=TEST_DLL,bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
+            "--upb_out=bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
             " ".join(src_files),
     )
 
--- a/upb_generator/bootstrap_compiler.bzl
+++ b/upb_generator/bootstrap_compiler.bzl
@@ -67,7 +67,7 @@ def _stage0_proto_staleness_test(name, base_dir, src_files, src_rules, strip_pre
             "$(location " + _protoc + ") " +
             "-I$(GENDIR)/" + strip_prefix + " " + _extra_proto_path +
             "--plugin=protoc-gen-upb=$(location " + _upbc("upb", 0) + ") " +
-	    "--upb_out=dllexport_tag=TEST_DLL,bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
+	    "--upb_out=dllexport_decl=TEST_DLL,bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
             " ".join(src_files),
     )
 
--- a/upb_generator/bootstrap_compiler.bzl
+++ b/upb_generator/bootstrap_compiler.bzl
@@ -67,7 +67,7 @@ def _stage0_proto_staleness_test(name, base_dir, src_files, src_rules, strip_pre
             "$(location " + _protoc + ") " +
             "-I$(GENDIR)/" + strip_prefix + " " + _extra_proto_path +
             "--plugin=protoc-gen-upb=$(location " + _upbc("upb", 0) + ") " +
-            "--upb_out=bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
+	    "--upb_out=dllexport_tag=TEST_DLL,bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
             " ".join(src_files),
     )
 
--- a/upb_generator/common.cc
+++ b/upb_generator/common.cc
@@ -60,6 +60,26 @@ std::string MessageInitName(upb::MessageDefPtr descriptor) {
   return MessageInit(descriptor.full_name());
 }
 
+void EmitDllExportMacros(absl::string_view dllexport_tag, Output& output) {
+  if (dllexport_tag.length()) {
+    output(
+        "#if !defined $0\n"
+        "#if defined $0_EXPORT\n"
+        "#define $0 __declspec(dll_export)\n"
+        "#elif defined $0_IMPORT\n"
+        "#define $0 __declspec(dll_import)\n"
+        "#else\n"
+        "#define $0\n"
+        "#endif\n"
+        "#endif\n\n",
+        dllexport_tag);    
+  }
+}
+
+std::string PadSuffix(absl::string_view tag) {
+  return tag.length() ? absl::StrCat(tag, " ") : std::string(tag);
+}
+
 std::string MessageName(upb::MessageDefPtr descriptor) {
   return ToCIdent(descriptor.full_name());
 }
--- a/upb_generator/common.cc
+++ b/upb_generator/common.cc
@@ -60,8 +60,8 @@ std::string MessageInitName(upb::MessageDefPtr descriptor) {
   return MessageInit(descriptor.full_name());
 }
 
-std::string PadSuffix(absl::string_view tag) {
-  return tag.length() ? absl::StrCat(tag, " ") : std::string(tag);
+std::string PadPrefix(absl::string_view tag) {
+  return tag.empty() ? "": absl::StrCat(" ", tag);
 }
 
 std::string MessageName(upb::MessageDefPtr descriptor) {
--- a/upb_generator/common.cc
+++ b/upb_generator/common.cc
@@ -60,8 +60,8 @@ std::string MessageInitName(upb::MessageDefPtr descriptor) {
   return MessageInit(descriptor.full_name());
 }
 
-void EmitDllExportMacros(absl::string_view dllexport_tag, Output& output) {
-  if (dllexport_tag.length()) {
+void EmitDllExportMacros(absl::string_view dllexport_decl, Output& output) {
+  if (dllexport_decl.length()) {
     output(
         "#if !defined $0\n"
         "#if defined $0_EXPORT\n"
@@ -72,7 +72,7 @@ void EmitDllExportMacros(absl::string_view dllexport_tag, Output& output) {
         "#define $0\n"
         "#endif\n"
         "#endif\n\n",
-        dllexport_tag);    
+        dllexport_decl);    
   }
 }
 
--- a/upb_generator/common.cc
+++ b/upb_generator/common.cc
@@ -60,22 +60,6 @@ std::string MessageInitName(upb::MessageDefPtr descriptor) {
   return MessageInit(descriptor.full_name());
 }
 
-void EmitDllExportMacros(absl::string_view dllexport_decl, Output& output) {
-  if (dllexport_decl.length()) {
-    output(
-        "#if !defined $0\n"
-        "#if defined $0_EXPORT\n"
-        "#define $0 __declspec(dll_export)\n"
-        "#elif defined $0_IMPORT\n"
-        "#define $0 __declspec(dll_import)\n"
-        "#else\n"
-        "#define $0\n"
-        "#endif\n"
-        "#endif\n\n",
-        dllexport_decl);    
-  }
-}
-
 std::string PadSuffix(absl::string_view tag) {
   return tag.length() ? absl::StrCat(tag, " ") : std::string(tag);
 }
--- a/upb_generator/protoc-gen-upbdefs.cc
+++ b/upb_generator/protoc-gen-upbdefs.cc
@@ -43,7 +43,7 @@ void GenerateMessageDefAccessor(upb::MessageDefPtr d, Output& output) {
   output("\n");
 }
 
-void WriteDefHeader(upb::FileDefPtr file, Options const& options,
+void WriteDefHeader(upb::FileDefPtr file, const Options& options,
                     Output& output) {
   EmitFileWarning(file.name(), output);
 
@@ -80,7 +80,7 @@ void WriteDefHeader(upb::FileDefPtr file, Options const& options,
       ToPreproc(file.name()));
 }
 
-void WriteDefSource(upb::FileDefPtr file, Options const& options, Output& output) {
+void WriteDefSource(upb::FileDefPtr file, const Options& options, Output& output) {
   EmitFileWarning(file.name(), output);
 
   output("#include \"upb/reflection/def.h\"\n");
@@ -132,7 +132,7 @@ void WriteDefSource(upb::FileDefPtr file, Options const& options, Output& output
   output("};\n");
 }
 
-void GenerateFile(upb::FileDefPtr file, Options const& options,
+void GenerateFile(upb::FileDefPtr file, const Options& options,
                   Plugin* plugin) {
   Output h_def_output;
   WriteDefHeader(file, options, h_def_output);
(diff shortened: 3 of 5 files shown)


dawidcha:
Following an exchange with @veblush I changed the name of the option that controls dllimport/export annotations to be the same as protobuf c++ (dllexport_decl)

The approach proposed in this PR is compatible with the protoc option, but I’m going to argue that it’s easier to use – and more in line with how this is implemented elsewhere, including in abseil.

To recap, both approaches append a macro to declarations in the header files declaring symbols that need to be visible to consumers of the DLL e.g. with dllexport_decl=MYLIB_DLL, we might get:

void MYLIB_DLL some_public_function();

When compiling the library code into the DLL, we set MYLIB_DLL=__declspec(dllexport). When compiling some other code which consumes the DLL we set MYLIB_DLL=__declspec(dllimport). For non-Windows or static libs, MYLIB_DLL can be set to empty string.

In most implementations of this, there is another layer of indirection: rather than having to know what to set MYLIB_DLL to, you define other macros and directives in some common include file ensure that MYLIB_DLL is set appropriately. For example, this is from [abseil-cpp](https://github.com/abseil/abseil-cpp/blob/b272bcbd0b3e89eae347cf71f3f7601e418761b6/absl/base/config.h#L712C1-L720C17), which uses ABSL_BUILD_DLL and ABSL_CONSUME_DLL to indicate we are building or consuming the DLL respectively:

```text
#if defined(ABSL_BUILD_DLL)
#define ABSL_DLL __declspec(dllexport)
#elif defined(ABSL_CONSUME_DLL)
#define ABSL_DLL __declspec(dllimport)
#else
#define ABSL_DLL
#endif
```
The protoc c++ generated code doesn’t do this, so you have to know to set MYLIB_DLL to __declspec(dllexport) or __declspec(dllimport), or empty string, as appropriate.

One reason it might have been done that way is that the approach used by abseil requires that there be a common ‘config.h’ which is included in all compilation units and contains the above snippet – which complicates the generated code. However, I’m going to argue that since we are code-generating, and these are purely pre-processor directives, we can duplicate this code in every header file that contains exported functions.

The pull request as implemented inserts the following code into those headers:

```text
#if !defined MYLIB_DLL
#if defined MYLIB_DLL_EXPORT
#define MYLIB_DLL __declspec(dll_export)
#elif defined MYLIB_DLL_IMPORT
#define MYLIB_DLL __declspec(dll_import)
#else
#define MYLIB_DLL
#endif
#endif
```
This can work the same way as the protoc c++ approach – if you set MYLIB_DLL=__declspec(dllimport), then it has the same effect as this code won’t try to define it if it’s already defined. However, you can also simply define MYLIB_DLL_IMPORT to achieve the same effect.

upb_generator/protoc-gen-upbdefs.cc
```diff
@@ -62,7 +66,8 @@ void GenerateMessageDefAccessor(upb::MessageDefPtr d, Output& output) {
   output("\n");
 }
 
-void WriteDefHeader(upb::FileDefPtr file, Output& output) {
+void WriteDefHeader(upb::FileDefPtr file, Options const& options,
```
fowles, line 69: west const please (applies throughout the file)

dawidcha: > west const please (applies throughout the file)

@fowles This is now fixed

dawidcha:
fyi, I am working on getting the CLA set up. I thought it was the same one as for grpc/grpc, but apparently not.

veblush:
@dawidcha Thank you for the PR with detailed explanation!

I agree with you that having `#if blocks` makes it easier for users to use generated code in this DLL setting. I suggested the same idea before to the Protobuf C++ (not upb). The idea to try to sell was having an option to add #Include custom header file statement in generate source code. And this customer header file defines `#if blocks` which technically does the same thing as your code does.

However, I'd prefer to start with the same thing that upb has first so that both can stay consistent. If you can manage to use the DLL of upb-generated file with the simple dllexport_decl, I think we can live with this although `dllexport_decl` needs to be defined differently based on whether it's for being built or consumed.

WDTY?

dawidcha:
> @dawidcha Thank you for the PR with detailed explanation!
> 
> I agree with you that having `#if blocks` makes it easier for users to use generated code in this DLL setting. I suggested the same idea before to the Protobuf C++ (not upb). The idea to try to sell was having an option to add #Include custom header file statement in generate source code. And this customer header file defines `#if blocks` which technically does the same thing as your code does.
> 
> However, I'd prefer to start with the same thing that upb has first so that both can stay consistent. If you can manage to use the DLL of upb-generated file with the simple dllexport_decl, I think we can live with this although `dllexport_decl` needs to be defined differently based on whether it's for being built or consumed.
> 
> WDTY?

I agree that the most important thing is to stay consistent between upb and protoc. I also considered adding an option for naming a 'config.h' (for example) header to be included in every generated file which has the necessary #if directives

So three options are:
1. What protoc does: no #if directives. Advantages are that it's the simplest in concept, but it's a bit ugly because it requires that the macro be defined explicitly via a /DMACRO=__declspec(dllexport) switch to the compiler when building the DLL which exports the symbols You'd also need to ensure that it is set to empty string if compiling the same sources as a static lib. For consuming that DLL, we can always include a common 'config.h' before including any of the generated headers, so it's only ugly when building the library.
2. Repeat the #if directives in every generated header (what this PR currently does). The user of the generated source doesn't need to do anything special, just follow the well understood convention of defining one symbol for exporting and a different symbol for consuming. I still hold out that duplicating code in generated code shouldn't be viewed as a red flag: replace this code with a #include and the compiler will literally end up parsing the exact same code after the preprocessor has done its stuff. In other words this is just pre-processor inlining. Assuming the DLL exports other symbols, the developer will still need a 'config.h' but can use the generated #if directives as a template for what to write in that. Note that this approach is compatible with option 1 - the developer can choose to set the macro to __declspec(dllexport/import) externally and that setting will be respected.
3. Nominate a named 'config.h' file to be written by the developer and including the necessary directives via another property of the plugin. Then write a #include line for this header into every generated header file. This ends up with the neatest solution - and what you would write if the generated code had been hand written. The main downside is, I think, ensuring that developers know what to do. It's going to need some careful documentation to ensure that developers understand how this works how it follows convention, and what they need to write in their 'config.h'.

In summary, my preference is still option 2 above, as it seems to me to be the most developer-friendly. However, I am happy to change the PR to use a different approach as directed - and update the protoc generator as needed (in a separate PR) so that both generators are consistent.

fowles:
We definitely do not want option 3 above.  I think we should stick with option 1.  You can use the existing upb/port/def.inc as a config file if you need to

dawidcha:
> We definitely do not want option 3 above. I think we should stick with option 1. You can use the existing upb/port/def.inc as a config file if you need to

I don't think we can use upb/port/def.inc as the 'annotation macro' that gets added to make symbols public is going to be different for each lib.

dawidcha:
In the interests of expediting approval on this PR, and absent any other opinion, by tomorrow, I'll go ahead and make the change to conform with option 1.

veblush:
@fowles 

> You can use the existing upb/port/def.inc as a config file if you need to

This doesn't work as this needs to be specific to certain targets. e.g. I should be able to have different __declspec for different upb targets. Eventually we need to have the option 2 or option 3 that David explained above for both `protoc` and `upbc` if we want to meet Windows Developers' expectation. 

@coryan, fyi.

veblush:
@dawidcha I haven't seen changes you mentioned in your comment. Is there something you're waiting for?

dawidcha:
@fowles @veblush I've removed the bits of code that define the DLL export macros making the submission follow option 1 above - and be entirely consistent with what the built-in protobuf c++ generator is doing.

Are there tests that need to run to validate this? I'm still chasing the CLA at my end.

Assuming those are resolved is there anything else blocking this being merged?

veblush:
@dawidcha LGTM to me. This needs reviews from Protobuf team. @fowles

ericsalo (commented):
Please resync with HEAD - I see a few changes in this PR that appear to erase recent CLs.

dawidcha:
> Please resync with HEAD - I see a few changes in this PR that appear to erase recent CLs.

Hi @ericsalo, thanks for taking a look.

I've rebased the add_dll_tags branch onto upstream HEAD. However, I'm not sure that doing that affected the diffs in any way, and I can't see any instances of erasure of recent CLs. Can you be more specific?

ericsalo:
Unless I'm missing something obvious, this PR appears to revert both https://github.com/protocolbuffers/protobuf/commit/9a610f02f8a50395852033a5f1f28084cfa29841 and https://github.com/protocolbuffers/protobuf/commit/c20b337221fb1c8fb02b4ea3d9363adbc2ef0ad3

dawidcha:
> Unless I'm missing something obvious, this PR appears to revert both [9a610f0](https://github.com/protocolbuffers/protobuf/commit/9a610f02f8a50395852033a5f1f28084cfa29841) and [c20b337](https://github.com/protocolbuffers/protobuf/commit/c20b337221fb1c8fb02b4ea3d9363adbc2ef0ad3)

You are not wrong (bad conflict resolution) - Let me fix that.

upb_generator/protoc-gen-upbdefs.cc
```diff
@@ -77,7 +82,8 @@ void WriteDefHeader(upb::FileDefPtr file, Output& output) {
       "#endif\n\n",
       ToPreproc(file.name()));
 
-  output("extern _upb_DefPool_Init $0;\n", DefInitSymbol(file));
+  output("extern $1_upb_DefPool_Init $0;\n", DefInitSymbol(file),
```
esrauchg, line 85: Nit: While it is a nice property that the codegen doesn't have 2 spaces when the param isn't passed, having the $1 attached to the symbol names looks too much to me like its actually trying to concatenate there.

I think if we want to avoid double-spaces it might be clearer if you attach the $1 to the extern (and PadPrefix instead of PadSuffix), since it'll be more clear that its not actually concatenating to make a type name. What do you think?

upb_generator/bootstrap_compiler.bzl
```diff
@@ -67,7 +67,7 @@ def _stage0_proto_staleness_test(name, base_dir, src_files, src_rules, strip_pre
             "$(location " + _protoc + ") " +
             "-I$(GENDIR)/" + strip_prefix + " " + _extra_proto_path +
             "--plugin=protoc-gen-upb=$(location " + _upbc("upb", 0) + ") " +
-            "--upb_out=bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
+	    "--upb_out=dllexport_decl=TEST_DLL,bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
```
esrauchg, line 70: Hm, can you explain this part of the change to me? Or is this just lingering from testing the flag?

esrauchg (changes requested):
Just a few stylistic comments, thanks for this!

upb_generator/common.cc
```diff
@@ -91,6 +91,10 @@ std::string MessageInitName(upb::MessageDefPtr descriptor) {
   return MessageInit(descriptor.full_name());
 }
 
+std::string PadSuffix(absl::string_view tag) {
+  return tag.length() ? absl::StrCat(tag, " ") : std::string(tag);
```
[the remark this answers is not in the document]
dawidcha: Agreed, I'll make the change

upb_generator/protoc-gen-upbdefs.cc
dawidcha: Yes, I think that is less confusing - will change.

upb_generator/bootstrap_compiler.bzl
dawidcha: This was put in as a way of smoke-testing the dllexport_decl parameter - by invoking it as part of the build we at least get a failure if it's totally broken. However, I now realize that since removing the 'export declarations' from the generated files (reverting to 'option 1' from 'option 2'), this will likely break the build because the TEST_DLL symbol is no longer being explicitly defined.

Can you suggest a place I can add a proper test for this?

upb_generator/bootstrap_compiler.bzl
```diff
@@ -67,7 +67,7 @@ def _stage0_proto_staleness_test(name, base_dir, src_files, src_rules, strip_pre
             "$(location " + _protoc + ") " +
             "-I$(GENDIR)/" + strip_prefix + " " + _extra_proto_path +
             "--plugin=protoc-gen-upb=$(location " + _upbc("upb", 0) + ") " +
-            "--upb_out=bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
+	    "--upb_out=bootstrap_upb:$(@D)/bootstrap_generated_sources/" + base_dir + "stage0 " +
```
esrauchg, line 70: Looks like you accidentally replaced the tabs with spaces in this file when intending to revert, can you revert?

Once you resolve this the PR LGTM and I'll unblock it to move forward. Thank you for iterating on this!

upb_generator/bootstrap_compiler.bzl
esrauchg: We unfortunately don't have a good place for this shape of codegen integration test, in the interest of moving forward we can leave this without a test and hopefully get coverage later. Thanks!

esrauchg:
FYI it looks like the CLA check is still failing. Can you follow up on the steps to get your CLA confirmed so that we can merge the PR? Thanks!

dawidcha:
Finally. Managed to run the bazel build, though I had to make a change to src/google/protobuf/message.cc to get it to compile - I submitted that change as a separate PR [here](https://github.com/protocolbuffers/protobuf/pull/14720) in case it's helpful.

_build command_
bazel build //upb_generator:protoc-gen-upb //upb_generator:protoc-gen-upbdefs

_test protoc-gen-upb_
protoc --upb_out=dllexport_decl=TEST_DLL,bootstrap_upb:upb_out examples/addressbook.proto --proto_path src --proto_path . --plugin ./bazel-bin/upb_generator/protoc-gen-upb

_test proco-gen-upbdefs_
protoc --upbdefs_out=dllexport_decl=TEST_DLL:upbdefs_out examples/addressbook.proto --proto_path src --proto_path . --plugin ./bazel-bin/upb_generator/protoc-gen-upbdefs

Generated files were checked manually for the existence of the 'TEST_DLL' annotation on external symbols.

Still battling with internal protocols to get me added to the list of CLA approved people.

hlopko:
(please note that we'll need you to sign the CLA in order to be able to accept your contribution: https://github.com/protocolbuffers/protobuf/pull/14230/checks?check_run_id=18529039461)

dawidcha:
Just rebased, but CLA is still a problem. Our internal processes for open source contributions changed recently and the team that deals with CLAs is working through the repercussions. Unfortunately this is the cost of working in a highly regulated industry :-(

dawidcha:
Hi @esrauchg 
 
I am told by our CLA folks that a request to "sync up google groups" is pending and once that is done, the CLA test should pass. Does that sound right? Is this something that can be expedited?

esrauchg:
I forced the CLA rescan and it is still showing it as not signed (and any Google Groups delays or anything would be long since passed). Can you double check the suggested steps on the CLA to confirm the email/github address matches? Thanks!

esrauchg:
Unfortunately the CLA check is still red. Can you check the steps suggested on "Help! I've signed the CLA, but it's still showing me as unsigned." on the checks page to proceed? Thanks!

dawidcha:
Recreating this PR as suggested to try to fix the missing CLA issue

PR #14272, opened 2023-10-03
Use ConcurrentDictionary in JsonFormatter

JamesNK:
I have a theory that the lock in `JsonFormatter` periodically causes this VS debugging error for Protobuf types:

image

Hopefully `ConcurrentDictionary` resolves this issue. Even if it doesn't, this change still resolves the TODO so this is a good improvement.

### DIFF
--- a/csharp/src/Google.Protobuf/JsonFormatter.cs
+++ b/csharp/src/Google.Protobuf/JsonFormatter.cs
@@ -970,10 +970,7 @@ namespace Google.Protobuf
                 Justification = "The field for the value must still be present. It will be returned by reflection, will be in this collection, and its name can be resolved.")]
             internal static string GetOriginalName(object value)
             {
-                Dictionary<object, string> nameMapping = dictionaries.GetOrAdd(value.GetType(), static t =>
-                {
-                    return GetNameMapping(t);
-                });
+                Dictionary<object, string> nameMapping = dictionaries.GetOrAdd(value.GetType(), static t => GetNameMapping(t));
 
                 // If this returns false, originalName will be null, which is what we want.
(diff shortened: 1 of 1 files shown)


JamesNK:
cc @jskeet

jskeet:
I'd certainly be happy to use ConcurrentDictionary. I think basically this code predates the availability of it in all the needed target platforms.

JamesNK:
.NET 3.5 doesn't have it. No longer a blocker.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -962,25 +963,17 @@ public Settings(bool formatDefaultValues, TypeRegistry typeRegistry) : this(form
...
+                = new ConcurrentDictionary<System.Type, Dictionary<object, string>>();
 
             [UnconditionalSuppressMessage("Trimming", "IL2072",
                 Justification = "The field for the value must still be present. It will be returned by reflection, will be in this collection, and its name can be resolved.")]
             internal static string GetOriginalName(object value)
             {
-                var enumType = value.GetType();
-                Dictionary<object, string> nameMapping;
-                lock (dictionaries)
+                Dictionary<object, string> nameMapping = dictionaries.GetOrAdd(value.GetType(), static t =>
```
jskeet, line 973: Any reason for not just using `static t => GetNameMapping(t)` here rather than a block body? (I *would* suggest `GetNameMapping` with a method group conversion, but I think that still creates a delegate instance per call with most compilers, annoyingly.)

jskeet (approved):

mkruskal-google (approved):

PR #14284, opened 2023-10-04
Add AsStructEnumerable() method on RepeatedField and MapField (C#)

### LINKED ISSUE #13503: Add API to prevent boxing of `Enumerator`s

C#

When iterating over `MapField`s and `RepeatedField`s using the `GetEnumerator` method, the
enumerator struct is boxed, as the signature of `GetEnumerator` is:

```csharp
public IEnumerator<T> GetEnumerator()
```

The interface return type causes the enumerator to always be boxed. This is not only slower, but
also allocates some memory. The amount of allocated memory is usually not a problem, but if you're
doing it inside of a loop, this might end up being an issue. See the benchmark results at the bottom
of this issue for more details.

I'd like an API which exposes the enumerator without boxing it. Collections in the BCL also do this,
see
https://github.com/dotnet/runtime/blob/b68382b60349fece947020d234d4a590e9bd5fe8/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L365.

Unfortunately it isn't possible to change the return type of `GetEnumerator`, as that method is
required by the `IEnumerable` interface. One option would be to make the implementation explicit, in
which case adding a new `GetEnumerator` method with a different return type becomes possible. This
has the benefit that `foreach` loops will also use this overload, increasing their performance.
However, this would be a breaking change.

An alternative to that might be to simply expose a new method `GetEnumeratorWithoutBoxing`, which
could be consumed manually. For `MapField` this is as simple as exposing the `LinkedList`
enumerator. For `RepeatedField` this would be slightly more involved, as arrays in C# don't have a
struct enumerator, but it would still be a very straight forward implementation. I included the
implementations I used for the benchmarks at the end of this issue.

I'd be more than happy to create a PR for this, if this is something you're interested in.

As shown in the benchmarks below, boxing of the `RepeatedField` enumerator can be avoided by using a
`for` loop, which performs comparably to an enumerator which isn't boxed. This workaround is not
possible for the `MapField` type, as you can't index it with an `int`.

**Benchmarks.**

<details>
<summary>Benchmark results MapField</summary>

```

BenchmarkDotNet v0.13.7, Windows 11 (10.0.22621.2134/22H2/2022Update/SunValley2)
11th Gen Intel Core i7-11800H 2.30GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 7.0.400
  [Host]             : .NET 7.0.10 (7.0.1023.36312), X64 RyuJIT AVX2
  .NET 7.0           : .NET 7.0.10 (7.0.1023.36312), X64 RyuJIT AVX2
  .NET Framework 4.8 : .NET Framework 4.8.1 (4.8.9167.0), X64 RyuJIT VectorSize=256


```

| Method           | Job                | Runtime            |     Mean |    Error |   StdDev |   Gen0 | Allocated |
| ---------------- | ------------------ | ------------------ | -------: | -------: | -------: | -----: | --------: |
| DefaultForeach   | .NET 7.0           | .NET 7.0           | 53.59 ns | 1.085 ns | 1.372 ns | 0.0038 |      48 B |
| CustomEnumerator | .NET 7.0           | .NET 7.0           | 40.98 ns | 0.606 ns | 0.567 ns |      - |         - |
| DefaultForeach   | .NET Framework 4.8 | .NET Framework 4.8 | 90.82 ns | 1.738 ns | 1.626 ns | 0.0088 |      56 B |
| CustomEnumerator | .NET Framework 4.8 | .NET Framework 4.8 | 43.02 ns | 0.465 ns | 0.435 ns |      - |         - |

</details>

<details>
<summary>Benchmark results RepeatedField</summary>

```

BenchmarkDotNet v0.13.7, Windows 11 (10.0.22621.2134/22H2/2022Update/SunValley2)
11th Gen Intel Core i7-11800H 2.30GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 7.0.400
  [Host]             : .NET 7.0.10 (7.0.1023.36312), X64 RyuJIT AVX2
  .NET 7.0           : .NET 7.0.10 (7.0.1023.36312), X64 RyuJIT AVX2
  .NET Framework 4.8 : .NET Framework 4.8.1 (4.8.9167.0), X64 RyuJIT VectorSize=256


```

| Method           | Job                | Runtime            |      Mean |     Error |    StdDev |   Gen0 | Allocated |
| ---------------- | ------------------ | ------------------ | --------: | --------: | --------: | -----: | --------: |
| DefaultForeach   | .NET 7.0           | .NET 7.0           | 33.161 ns | 0.3042 ns | 0.2846 ns | 0.0032 |      40 B |
| CustomEnumerator | .NET 7.0           | .NET 7.0           |  5.833 ns | 0.1348 ns | 0.1195 ns |      - |         - |
| DefaultFor       | .NET 7.0           | .NET 7.0           |  5.375 ns | 0.0696 ns | 0.0651 ns |      - |         - |
| DefaultForeach   | .NET Framework 4.8 | .NET Framework 4.8 | 46.320 ns | 0.7284 ns | 0.6457 ns | 0.0063 |      40 B |
| CustomEnumerator | .NET Framework 4.8 | .NET Framework 4.8 | 14.018 ns | 0.1128 ns | 0.1000 ns |      - |         - |
| DefaultFor       | .NET Framework 4.8 | .NET Framework 4.8 | 13.320 ns | 0.0847 ns | 0.0793 ns |      - |         - |

</details>

<details>
<summary>MapField custom enumerator implementation</summary>

```csharp
public LinkedList< KeyValuePair< TKey, TValue > >.Enumerator GetCustomEnumerator( ) => list.GetEnumerator( );
```

</details>

<details>
<summary>RepeatedField custom enumerator implementation</summary>

```csharp
public Enumerator GetCustomEnumerator( ) => new( this );

public struct Enumerator : IEnumerator< T >
{
    private const int UNSET_IDX = -1;
    private readonly RepeatedField< T > repeatedField;

    private int idx;

    internal Enumerator(
        RepeatedField< T > repeatedField )
    {
        this.repeatedField = repeatedField;
        idx = UNSET_IDX;
    }

    public readonly void Dispose( )
    {
    }

    public bool MoveNext( ) => ++idx < repeatedField.count;

    public void Reset( ) => idx = UNSET_IDX;

    readonly object IEnumerator.Current => repeatedField[idx];

    public readonly T Current => repeatedField[idx];
}
```

</details>

<details>
<summary>Benchmark implementations</summary>

```csharp
[MemoryDiagnoser]
[SimpleJob( RuntimeMoniker.Net70 )]
[SimpleJob( RuntimeMoniker.Net48 )]
public class RepeatedFieldEnumeratorBenchmark
{
    private RepeatedField< int > repeatedField;

    [GlobalSetup]
    public void Setup( )
    {
        repeatedField = new RepeatedField< int >( );
        for ( int i = 0;
              i < 10;
              i++ )
        {
            repeatedField.Add( i );
        }
    }

    [Benchmark]
    public int DefaultForeach( )
    {
        int result = 0;
        foreach ( int i in repeatedField )
        {
            result = i;
        }
        return result;
    }

    [Benchmark]
    public int CustomEnumerator( )
    {
        int result = 0;
        using( RepeatedField< int >.Enumerator enumerator = repeatedField.GetCustomEnumerator( ) )
        {
            while ( enumerator.MoveNext( ) )
            {
                result = enumerator.Current;
            }

            return result;
        }
    }

    [Benchmark]
    public int DefaultFor( )
    {
        int result = 0;
        var repeated = repeatedField;
        for ( int i = 0;
              i < repeated.Count;
              i++ )
        {
            result = repeated[ i ];
        }
        return result;
    }
}

[MemoryDiagnoser]
[SimpleJob( RuntimeMoniker.Net70 )]
[SimpleJob( RuntimeMoniker.Net48 )]
public class MapFieldEnumeratorBenchmark
{
    private MapField< int, int > mapField;

    [GlobalSetup]
    public void Setup( )
    {
        mapField = new MapField< int, int >( );
        for ( int i = 0;
              i < 10;
              i++ )
        {
            mapField.Add(
                i,
                i );
        }
    }

    [Benchmark]
    public int DefaultForeach( )
    {
        int result = 0;
        foreach ( KeyValuePair< int, int > keyValuePair in mapField )
        {
            result = keyValuePair.Value;
        }
        return result;
    }

    [Benchmark]
    public int CustomEnumerator( )
    {
        int result = 0;
        using( LinkedList< KeyValuePair< int, int > >.Enumerator enumerator = mapField.GetCustomEnumerator( ) )
        {
            while ( enumerator.MoveNext( ) )
            {
                result = enumerator.Current.Value;
            }

            return result;
        }
    }
}
```

</details>

jskeet:
Code should basically be okay, but I'd like feedback before writing XML docs.

Fixes #13503.

### DIFF
--- a/csharp/src/Google.Protobuf/Collections/MapField.cs
+++ b/csharp/src/Google.Protobuf/Collections/MapField.cs
@@ -531,6 +531,26 @@ namespace Google.Protobuf.Collections
             return writer.ToString();
         }
 
+        /// <summary>
+        /// Returns an <see cref="Enumerable"/> view over this map, enabling
+        /// iteration over the entries without any boxing operations.
+        /// The returned value contains a reference to the instance this method
+        /// is called on; it doesn't create a snapshot of the content.
+        /// </summary>
+        /// <remarks>
+        /// Use this method to avoid allocating any new objects when iterating over
+        /// the map. In most applications this will not have any significant
+        /// performance impact, and most developers should just iterate over the map
+        /// directly. However, in some cases (particularly when iterating over
+        /// many empty maps) the boxing operation required to iterate can be
+        /// significant.
+        /// </remarks>
+        /// <returns>
+        /// An <see cref="Enumerable"/> allowing for efficient iteration over
+        /// the content of the map.
+        /// </returns>
+        public Enumerable AsStructEnumerable() => new Enumerable(this);
+
         #region IDictionary explicit interface implementation
 
         void IDictionary.Add(object key, object value) => Add((TKey)key, (TValue)value);
@@ -719,6 +739,70 @@ namespace Google.Protobuf.Collections
                     array.SetValue(item, index++);
                 }
             }
+        }
+
+        /// <summary>
+        /// A value type wrapping a <see cref="MapField{TKey, TValue}"/> to enable
+        /// efficient iteration without boxing.
+        /// </summary>
+        /// <seealso cref="MapField{TKey, TValue}.AsStructEnumerable"/>
+        public readonly struct Enumerable : IEnumerable<KeyValuePair<TKey, TValue>>
+        {
+            private readonly MapField<TKey, TValue> field;
+
+            internal Enumerable(MapField<TKey, TValue> field)
+            {
+                this.field = field;
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="Enumerator"/> to iterate over the map.
+            /// </summary>
+            /// <returns>An <see cref="Enumerator"/> to iterate over the map.</returns>
+            public Enumerator GetEnumerator() => new Enumerator(field);
+
+            IEnumerator<KeyValuePair<TKey, TValue>> IEnumerable<KeyValuePair<TKey, TValue>>.GetEnumerator() => GetEnumerator();
+
+            IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+        }
+
+        /// <summary>
+        /// A value type implementation of <see cref="IEnumerator{T}"/>,
+        /// to enable efficient iteration over a <see cref="MapField{TKey, TValue}"/> without boxing.
+        /// </summary>
+        /// <seealso cref="MapField{TKey, TValue}.AsStructEnumerable"/>
+        public struct Enumerator : IEnumerator<KeyValuePair<TKey, TValue>>
+        {
+            private LinkedList<KeyValuePair<TKey, TValue>>.Enumerator enumerator;
+
+            internal Enumerator(MapField<TKey, TValue> field)
+            {
+                this.enumerator = field.list.GetEnumerator();
+            }
+
+            /// <summary>
+            /// Gets the element at the current position of the enumerator.
+            /// </summary>
+            public KeyValuePair<TKey, TValue> Current => enumerator.Current;
+
+            object IEnumerator.Current => Current;
+
+            /// <summary>
+            /// Releases all resources used by the enumerator.
+            /// </summary>
+            public void Dispose() => enumerator.Dispose();
+
+            /// <summary>
+            /// Advances the enumerator to the next element of the map.
+            /// </summary>
+            /// <returns><c>true</c> if the enumerator was successfully advanced to the next element;
+            /// <c>false</c> if the enumerator has passed the end of the collection.</returns>
+            public bool MoveNext() => enumerator.MoveNext();
+
+            /// <summary>
+            /// Always throws <see cref="NotSupportedException"/>.
+            /// </summary>
+            void IEnumerator.Reset() => throw new NotSupportedException();
         }
 
         private sealed class MapFieldDebugView
--- a/csharp/src/Google.Protobuf/Collections/RepeatedField.cs
+++ b/csharp/src/Google.Protobuf/Collections/RepeatedField.cs
@@ -645,7 +645,105 @@ namespace Google.Protobuf.Collections
                 Remove(t);
             }
         }
-        #endregion        
+        #endregion
+
+        /// <summary>
+        /// Returns an <see cref="Enumerable"/> view over this repeated field, enabling
+        /// iteration over the entries without any boxing operations.
+        /// The returned value contains a reference to the instance this method
+        /// is called on; it doesn't create a snapshot of the content.
+        /// </summary>
+        /// <remarks>
+        /// Use this method to avoid allocating any new objects when iterating over
+        /// the repeated field. In most applications this will not have any significant
+        /// performance impact, and most developers should just iterate over the repeated
+        /// field directly. However, in some cases (particularly when iterating over
+        /// many empty repeated fields) the boxing operation required to iterate can be
+        /// significant.
+        /// </remarks>
+        /// <returns>
+        /// An <see cref="Enumerable"/> allowing for efficient iteration over
+        /// the content of the repeated field.
+        /// </returns>
+        public Enumerable AsStructEnumerable() => new Enumerable(this);
+
+        /// <summary>
+        /// A value type wrapping a <see cref="RepeatedField{T}"/> to enable
+        /// efficient iteration without boxing.
+        /// </summary>
+        /// <seealso cref="RepeatedField{T}.AsStructEnumerable"/>
+        public readonly struct Enumerable : IEnumerable<T>
+        {
+            private readonly RepeatedField<T> field;
+
+            internal Enumerable(RepeatedField<T> field)
+            {
+                this.field = field;
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="Enumerator"/> to iterate over the repeated field.
+            /// </summary>
+            /// <returns>An <see cref="Enumerator"/> to iterate over the repeated field.</returns>
+            public Enumerator GetEnumerator() => new Enumerator(field);
+
+            IEnumerator<T> IEnumerable<T>.GetEnumerator() => GetEnumerator();
+
+            IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+        }
+
+        /// <summary>
+        /// A value type implementation of <see cref="IEnumerator{T}"/>,
+        /// to enable efficient iteration over a <see cref="RepeatedField{T}"/> without boxing.
+        /// </summary>
+        /// <seealso cref="MapField{TKey, TValue}.AsStructEnumerable"/>
+        public struct Enumerator : IEnumerator<T>
+        {
+            private readonly RepeatedField<T> field;
+            private int index;
+
+            internal Enumerator(RepeatedField<T> field)
+            {
+                this.field = field;
+                this.index = -1;
+            }
+
+            /// <summary>
+            /// Gets the element at the current position of the enumerator.
+            /// </summary>
+            public T Current => index == -1 || index >= field.Count
+                ? throw new InvalidOperationException()
+                : field[index];
+
+            object IEnumerator.Current => Current;
+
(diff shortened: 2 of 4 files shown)


csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -721,6 +727,65 @@ public void CopyTo(Array array, int index)
...
+            /// <summary>
+            /// 
+            /// </summary>
+            /// <returns></returns>
+            public bool MoveNext() => enumerator.MoveNext();
+
+            /// <summary>
+            /// 
+            /// </summary>
+            void IEnumerator.Reset() => throw new NotSupportedException();
```
amanda-tarafa, line 786: Wondering why you don't keep the list around so you can implement reset?

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -645,7 +645,86 @@ void IList.Remove(object value)
...
+            }
+
+            /// <summary>
+            /// 
+            /// </summary>
+            /// <returns></returns>
+            public bool MoveNext()
+            {
+                var count = field.Count;
+                index = Math.Min(index + 1, count);
```
amanda-tarafa, line 716: Why this extra check, instead of incrementing index always? Why do we care what is index value once it's equal to or greater than count? If the collection gets modified while iterating we don't need to guarantee any behaviour I guess.

amanda-tarafa (commented):
A couple of comments only.

csharp/src/Google.Protobuf/Collections/MapField.cs
jskeet: I wondered about that. My gut feeling is that Reset is basically used by *no-one*, and was an API mistake, so I didn't want to make anything even slightly more complex to support it.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
jskeet: Overflow - calling MoveNext 2^32 times shouldn't make it valid again. Maybe that's slightly silly, but...

KnapSac: I don't think the check is going to make much of a difference. If this turns out to be a performance problem, that would probably also go for the bounds checks on the array access. So it seems fine to keep it for now, for correctness sake.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -645,7 +645,86 @@ void IList.Remove(object value)
...
+                this.field = field;
+                this.index = -1;
+            }
+
+            /// <summary>
+            /// 
+            /// </summary>
+            public T Current => index == -1 || index >= field.Count
+                ? throw new InvalidOperationException()
+                : field[index];
```
KnapSac, line 698: Is this check really necessary? The indexer does the exact same check, but throws a slightly different exception. Skipping the check here would create an implicit dependency on that implementation, but that seems fine given that it belongs to the same class.

KnapSac (approved):
LGTM.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
jskeet: I think it's worth making sure we throw the right exception, personally.

jskeet:
(XML comments have now been added, and I consider this ready for final review.)

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -531,6 +531,26 @@ public override string ToString()
             return writer.ToString();
         }
 
+        /// <summary>
+        /// Returns an <see cref="Enumerable"/> view over this map, enabling
+        /// iteration over the entries without any boxing operations.
+        /// The returned value contains a reference to the instance this method
+        /// is called on; it doesn't not create a snapshot of the content.
```
KnapSac, line 538:
```suggestion
        /// is called on; it doesn't create a snapshot of the content.
```

csharp/src/Google.Protobuf/Collections/MapField.cs
```diff
@@ -721,6 +741,69 @@ public void CopyTo(Array array, int index)
...
+            private LinkedList<KeyValuePair<TKey, TValue>>.Enumerator enumerator;
+
+            internal Enumerator(MapField<TKey, TValue> field)
+            {
+                this.enumerator = field.list.GetEnumerator();
+            }
+
+            /// <summary>
+            /// 
+            /// </summary>
```
KnapSac, line 785: Are these `<summary>`s empty on purpose?

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -645,7 +645,105 @@ void IList.Remove(object value)
...
             }
         }
-        #endregion        
+        #endregion
+
+        /// <summary>
+        /// Returns an <see cref="Enumerable"/> view over this repeated field, enabling
+        /// iteration over the entries without any boxing operations.
+        /// The returned value contains a reference to the instance this method
+        /// is called on; it doesn't not create a snapshot of the content.
```
KnapSac, line 654:
```suggestion
        /// is called on; it doesn't create a snapshot of the content.
```

KnapSac (approved):
Couple of typos in the docs, but nothing worth blocking over.

csharp/src/Google.Protobuf/Collections/MapField.cs
jskeet: Fixed, thanks.

csharp/src/Google.Protobuf/Collections/MapField.cs
jskeet: Nope - had failed to save the file in Visual Studio! Should be fine now...

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
jskeet: Fixed.

KnapSac:
@jskeet any thoughts on when this might get merged?

jskeet:
I don't know I'm afraid - right now I'm on a rather urgent project which isn't giving me much time for protobuf.

jskeet:
@amanda-tarafa if you could review this, I'll look at getting it in. I *think* it's ready.

JamesNK:
I like the feature. The name is a bit 🤢 but I understand that `GetEnumerator` API is already taken on these types.

I wonder if there is prior art in .NET APIs for this situation.

jskeet:
(Removed inactive label as I want to get back to this when I have time.)

JamesNK:
This might not be needed after https://github.com/protocolbuffers/protobuf/pull/16772

Getting the span from the collection and iterating over the span should be fast.

jskeet:
I'd be okay to drop this given #16772, yes. @KnapSac: does that work for you? @amanda-tarafa any thoughts?

jskeet:
@KnapSac: Hmm, probably not. My experience is that map fields are used *vastly* more rarely than repeated fields. If we were doing `AsStructEnumerable` for repeated fields, it would make sense to have it for map fields as well... I'm not sure whether it's really worth it *just* for map fields. It's entirely possible that at this point you'd be the only user... I'll think about it more.

jskeet:
Reopening

jskeet:
Reopening as we at least *potentially* still want to do this.

Issue #13736, opened 2023-05-11
Build fails on Fedora 38 with clang 16 (-Wgnu-offsetof-extensions)


airlockgithubci:
The build fails on Fedora 38 with clang 16.

`.bazelrc` file to use clang:
```
build --config=clang
build --cxxopt=-std=c++17

build:clang --action_env=BAZEL_COMPILER=clang
build:clang --action_env=CC=clang --action_env=CXX=clang++
build:clang --linkopt=-fuse-ld=lld
```

Build error:
```
ERROR: /home/user/upb/BUILD:102:11: Compiling upb/arena.c failed: (Exit 1): clang-16 failed: error executing command (from target //:upb) /usr/bin/clang-16 -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer -MD -MF ... (remaining 22 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
upb/arena.c:177:25: error: defining a type within '__builtin_offsetof' is a Clang extension [-Werror,-Wgnu-offsetof-extensions]
  n = UPB_ALIGN_DOWN(n, UPB_ALIGN_OF(upb_Arena));
                        ^~~~~~~~~~~~~~~~~~~~~~~
./upb/port_def.inc:95:38: note: expanded from macro 'UPB_ALIGN_OF'
#define UPB_ALIGN_OF(type) offsetof (struct { char c; type member; }, member)
                                     ^~~~~~
/usr/lib64/clang/16/include/stddef.h:111:43: note: expanded from macro 'offsetof'
#define offsetof(t, d) __builtin_offsetof(t, d)
                                          ^
./upb/port_def.inc:93:48: note: expanded from macro 'UPB_ALIGN_DOWN'
#define UPB_ALIGN_DOWN(size, align) ((size) / (align) * (align))
                                               ^~~~~
upb/arena.c:177:25: error: defining a type within '__builtin_offsetof' is a Clang extension [-Werror,-Wgnu-offsetof-extensions]
  n = UPB_ALIGN_DOWN(n, UPB_ALIGN_OF(upb_Arena));
                        ^~~~~~~~~~~~~~~~~~~~~~~
./upb/port_def.inc:95:38: note: expanded from macro 'UPB_ALIGN_OF'
[...]

PR #14481, opened 2023-10-19
Fix 13736 error on upb compilation with Clang

RmStorm:
When compiling upb Clang will emit a warning like this: `error: defining a type within '__builtin_offsetof' is a Clang extension [-Werror,-Wgnu-offsetof-extensions]` as shown in issue #13736.

This commit fixes that error by adding the compiler flag '-Wno-gnu-offsetof-extensions' which suppresses the warning/error.

### DIFF
--- a/upb/BUILD
+++ b/upb/BUILD
@@ -91,7 +91,7 @@ cc_library(
     hdrs = [
         "upb.hpp",
     ],
-    copts = UPB_DEFAULT_COPTS,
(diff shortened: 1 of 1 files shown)


acozzette (approved):

esrauchg:
Per reply on #13736 this appears to be something else going on with the build configuration (the warning point at here is on a path that is already in a #ifndef __clang__) so the proposed PR isn't the fix we want. If there's something else going on wrong on our side which makes the wrong definition of the macro being used on clang which has a corresponding different fix please let us know on the issue.

Thanks!

PR #14562, opened 2023-10-31
Add Automatic-Module-Name

### LINKED ISSUE #12639: Automatic-Module-Name entry missing since release 3.21.12
**Issue**
Any newer release of protobuf-java after 3.21.12 is missing the Automatic-Module-Name entry in the META-INF/MANIFEST.MF file. This prevents using the protobuf module with a stable name instead. Instead, we are forced to the bad practice of using a filename-inferred module name.

Please restore the Automatic-Module-Name entry in the manifest or (alternatively) provide a proper module-info.class in the library.

Versions containing the issue are 3.22.0, 3.22.1, 3.22.2, 3.22.3 and 3.23.0-RC2
Last version to be correct: 3.21.12
Language: Java

Irrelevant as issue pertains to the Java Platform Module System.

Java 17

**What did you do?**
1. Add "requires transitive com.google.protobuf;" to the module-info.java of a project using protobuf-java.
2. Find compilation error due to "com.google.protobuf" cannot be resolved to a module.
3. Find that the protobuf manifest contains no Automatic-Module-Name  since version 3.21.12

I expected an Automatic-Module-Name entry in META-INF/MANIFEST.MF. This was last seen in version 3.21.12.

**What did you see instead?**
I found no Automatic-Module-Name entry in META-INF/MANIFEST.MF.

**Note**
I reported this before in #12175, which was closed because it was suggested the fix for #12017 would also restore the Automatic-Module-Name entry. However with #12017 fixed the OSGi entries are restored but the Automatic-Module-Name entry is still missing from the manifest.

Sineaggi:
Adds Automatic-Module-Name after it was lost during the maven-bazel migration (and subsequent re-addition of osgi bundle support).

Updates OsgiWrapper to support adding the Automatic-Module-Name header to the list of properties supported by the `osgi_java_library` rule.

Fixes #12639


deannagarcia (approved):

Sineaggi:
https://github.com/protocolbuffers/protobuf/actions/runs/6724762122/job/18277738222?pr=14562 one check failed, unrelated to my changes.

deannagarcia:
Agreed this failure is unrelated, a team member is working on getting it fixed.

PR #14570, opened 2023-10-31
refactor: relocate ruby targets under the //ruby package

### LINKED ISSUE #14569: Fork of rules_ruby
Every Bazel user of this repo depends on ruby, due to an eager `load` statement evaluated during Bazel's loading phase
https://github.com/protocolbuffers/protobuf/blob/30f157529d660bdb36ec3ed5abeaa50cec2d4e5f/protobuf.bzl#L5

However, this doesn't depend on a canonical rules_ruby which could be placed in the Bazel Central Registry, rather it's a protocolbuffers-team specific fork:

https://github.com/protocolbuffers/rules_ruby

which all bazel users are getting in their workspace due to
https://github.com/protocolbuffers/protobuf/blob/30f157529d660bdb36ec3ed5abeaa50cec2d4e5f/protobuf_deps.bzl#L115

This seems like a pretty major tech debt in this repo, and inhibits the task of providing protocol buffers on Bazel's registry, as I don't think we want your fork to be registered as the canonical rules_ruby, since you probably aren't offering to take the maintenance burden.

Note that the current entry https://registry.bazel.build/modules/protobuf/21.7 doesn't have this problem since the rules_ruby load statement didn't appear until later, introduced by https://github.com/protocolbuffers/protobuf/pull/10525

I'll try cutting this load by moving everything ruby-related under the //ruby package.  It still means bzlmod users won't be able to use the ruby well-known-types, but at least it makes the dependency graph satisfiable.

alexeagle:
Fixes #14569

Note, this is a breaking change for Bazel users who depended on the labels of the `ruby_library` targets of the well-known-protos

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -5,7 +5,7 @@ load("@rules_java//java:defs.bzl", "java_lite_proto_library", "java_proto_librar
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
 load("@rules_proto//proto:defs.bzl", "proto_lang_toolchain", "proto_library")
 load("//build_defs:cpp_opts.bzl", "COPTS", "LINK_OPTS")
-load(":protobuf.bzl", "internal_objc_proto_library", "internal_php_proto_library", "internal_py_proto_library", "internal_ruby_proto_library")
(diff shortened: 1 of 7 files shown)


alexeagle:
Apparently @mkruskal-google picked this up

PR #14572, opened 2023-10-31
Add an explicit self-reference `local_repository`

Wyverald:
See https://github.com/bazelbuild/bazel/issues/19973#issuecomment-1787814450 for an explanation.

cc @meteorcloudy


meteorcloudy:
To get this merged faster, you can consider submit this as a CL internally.

Wyverald:
> To get this merged faster, you can consider submit this as a CL internally.

I'll do that instead!

PR #14590, opened 2023-11-01
refactor:  toolchainize language runtimes

thesayyn:
Working towards https://github.com/bazelbuild/rules_proto/issues/179


.bazelversion
```diff
@@ -0,0 +1 @@
+5a24c8a3c6cd336ff69ce14fd1c95d50589ec27a
```
thesayyn, line 1: @comius this commit seems to fix issue where ToolchainInfo  is absent in proto_lang_toolchain in the earlier versions.

thesayyn: Isn't this possibly a breaking change?

BUILD.bazel
```diff
@@ -194,6 +194,18 @@ cc_binary(
     deps = ["//src/google/protobuf/compiler:protoc_lib"],
 )
 
+proto_toolchain(
```
comius, line 197: `proto_toolchain` is a macro that already creates a `toolchain` target for you.
https://github.com/bazelbuild/rules_proto/blob/main/proto/private/rules/proto_toolchain.bzl#L34

BUILD.bazel
```diff
@@ -194,6 +194,18 @@ cc_binary(
...
+    name = "protoc_source_toolchain",
```
comius, line 198: The `proto_toolchain` macro defined in rules_proto is causing problems with Bazel 6, because `allowlist_different_package` wasn't defined yet.

I marked https://github.com/bazelbuild/bazel/commit/6ede08f4a4243e4813b04169f24e07ba7354ef63 for cherry-picking.

What we could do in the mean time, is use https://github.com/bazel-contrib/bazel_features in rules_proto to make https://github.com/bazelbuild/rules_proto/blob/main/proto/private/rules/proto_toolchain_rule.bzl#L38 only be added on Bazel versions > 6.4.0.

.bazelversion
comius: After some really quick investigation, not even `--incompatible_enable_proto_toolchain_resolution` exists in a released bazel version. So we'll need to cherry-pick that set of changes first, before we can use (or flip the flag).

This change can go forward with other fixes and without the using the `--incompatible_enable_proto_toolchain_resolution`.

I've also tested the change with Bazel 7.0.0, with fixes to rules_java, mentioned in the other comment and I got one java_proto_library to build. So it looks promising.

protobuf_deps.bzl
```diff
@@ -178,3 +179,12 @@ def protobuf_deps():
         name = "nuget_python_x86-64_3.10.0",
         sha256 = "4474c83c25625d93e772e926f95f4cd398a0abbb52793625fa30f39af3d2cc00",
     )
+
+
+def protobuf_register_toolchains():
+    "registers source toolchain"
+    native.register_toolchains("//:protoc_toolchain")
```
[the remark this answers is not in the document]
thesayyn: I don't think using `:all` is good as we have had problems with in the past. It is implicit and hides a lot of information, like targets being sorted before being registered. also it might register stuff we didn't meant to. 

We usually create two .bzl files that users need to load in their WORKSPACE. one for bringing in the dependencies and then performing toolchain registration. Wouldn't that be preferable instead of a new package?

thesayyn:
@comius this should be ready now.

.bazelignore
```diff
@@ -1,4 +1,5 @@
 # These are fetched as external repositories.
 third_party/abseil-cpp
 third_party/googletest
+third_party/utf8_range
```
acozzette, line 4: We recently moved utf8_range into this repo, so third_party/utf8_range needs to be visible to Bazel.

acozzette (changes requested):

.bazelignore
thesayyn: done. PTAL @acozzette

PR #14594, opened 2023-11-01
Add support for options in CRuby, JRuby and FFI

### LINKED ISSUE #1198: No way to use custom options in Ruby

JasonLunn:
Rewrrte and extension of #12828, with additional work for JRuby. Partially fixes #1198 by adding support for custom options. Handling of extensions will be handled in a follow up.

Also includes these unrelated fixes:
* Removes code echo between `google/protobuf/repeated_field.rb` and `google/protobuf/ffi/repeated_field.rb` by `require`'ing the former in the latter.
* Adds missing calles to `testFrozen()` from methods of `RepeatedField` under JRuby that mutate.
* Various typos in comments.

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -44,7 +44,7 @@ static VALUE rb_str_maybe_null(const char* s) {
   }
   return rb_str_new2(s);
 }
-
+static ID options_instancevar_interned;
 // -----------------------------------------------------------------------------
 // DescriptorPool.
 // -----------------------------------------------------------------------------
@@ -192,6 +192,7 @@ static void DescriptorPool_register(VALUE module) {
 
   rb_gc_register_address(&generated_pool);
   generated_pool = rb_class_new_instance(0, NULL, klass);
+  options_instancevar_interned = rb_intern("options");
 }
 
 // -----------------------------------------------------------------------------
@@ -227,9 +228,15 @@ static Descriptor* ruby_to_Descriptor(VALUE val) {
 }
 
 // Decode and return a frozen instance of a Descriptor Option for the given pool
-static VALUE decode_options(const char *option_type, int size, const char *bytes, VALUE descriptor_pool) {
+static VALUE decode_options(VALUE self, const char *option_type, int size, const char *bytes, VALUE descriptor_pool) {
+  VALUE options_rb = rb_ivar_get(self, options_instancevar_interned);
+  if (options_rb!= Qnil) {
+    return options_rb;
+  }
+
   static const char * prefix = "google.protobuf.";
-  char fullname[strlen(prefix) + strlen(option_type) + 1];
+  char fullname[/*strlen(prefix)*/ 16 + /*strln(longest option type supported e.g. "MessageOptions")*/ 14 + /*null terminator*/1];
+
   snprintf(fullname, sizeof(fullname), "%s%s", prefix, option_type);
   const upb_MessageDef* msgdef = upb_DefPool_FindMessageByName(
     ruby_to_DescriptorPool(descriptor_pool)->symtab,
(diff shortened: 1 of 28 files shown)


ruby/ext/google/protobuf_c/glue.c
```diff
@@ -19,3 +19,35 @@ google_protobuf_FileDescriptorProto* FileDescriptorProto_parse(
   return google_protobuf_FileDescriptorProto_parse(serialized_file_proto,
                                                    length, arena);
 }
+
+char* EnumDescriptor_serialized_options(const upb_EnumDef* enumdef, size_t *size) {
+  const google_protobuf_EnumOptions* opts = upb_EnumDef_Options(enumdef);
```
JasonLunn, line 24: @haberman - is this pattern safe? Is it OK to return from this method the return value from `google_protobuf_EnumOptions_serialize()`? Is it a pointer into memory that was potentially just freed by the call to `upb_Arena_Free()`?

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -226,6 +226,25 @@ static Descriptor* ruby_to_Descriptor(VALUE val) {
   return ret;
 }
 
+// Decode and return a frozen instance of a Descriptor Option for the given pool
+static VALUE decode_options(const char *option_type, int size, const char *bytes, VALUE descriptor_pool) {
+  static const char * prefix = "google.protobuf.";
+  char fullname[strlen(prefix) + strlen(option_type) + 1];
```
haberman, line 232: We generally try to avoid VLAs. You could use `VALUE fullname = rb_sprintf("%s%s", prefix, option_type)`.

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -374,6 +393,23 @@ static VALUE Descriptor_msgclass(VALUE _self) {
...
+ *
+ * Returns the `MessageOptions` for this `Descriptor`.
+ */
+static VALUE Descriptor_options(VALUE _self) {
+  Descriptor* self = ruby_to_Descriptor(_self);
+  const google_protobuf_MessageOptions* opts = upb_MessageDef_Options(self->msgdef);
+  upb_Arena* arena = upb_Arena_New();
+  size_t size;
+  char* serialized = google_protobuf_MessageOptions_serialize(opts, arena, &size);
+  VALUE message_options = decode_options("MessageOptions", size, serialized, self->descriptor_pool);
```
haberman, line 408: We lost the caching. Could you use `rb_ivar_get()` and `rb_ivar_set()` to add back the caching we had before? This could go into `decode_options()` if you pass `self` as a parameter.

ruby/ext/google/protobuf_c/glue.c
haberman: It is unsafe as written. You'd want to pass the arena in as a parameter

haberman (changes requested):

ruby/ext/google/protobuf_c/defs.c
JasonLunn: Rewritten using `snprintf` - the string in question is just a C string, so avoiding using `rb_*` here.

haberman (approved):

ruby/ext/google/protobuf_c/map.c
```diff
@@ -572,6 +572,30 @@ static VALUE Map_freeze(VALUE _self) {
...
 }
 
+/*
+ * Deep freezes the map and values recursively.
+ * Internal use only.
+ */
+VALUE Map_internal_deep_freeze(VALUE _self) {
+  Map* self = ruby_to_Map(_self);
+
+  if (!RB_OBJ_FROZEN(_self)) {
```
haberman, line 582: We probably should remove this `if`. There is no guarantee that our children are already frozen. And in the limited places where we are using this, we already know that it is not frozen.

haberman: Ditto in the other places where we have `deep_freeze`

haberman (changes requested):

haberman (approved):

PR #14600, opened 2023-11-02
Fix `setup.py` build

ZhongRuoyu:
In 501ececd39f the `python/BUILD` Bazel build file was introduced. On a case-insensitive filesystem (default on e.g. macOS), it conflicts with `setup.py`'s default behaviour, which creates a `build` directory for build artifacts.

Fix that by renaming the `BUILD` file to `BUILD.bazel`.


acozzette:
Thank you for the pull request. There's a change we need to make internally to make this possible, so let me do that myself and I will send out a change for this.

PR #14647, opened 2023-11-05
Add file extension to bazel file in python subfolder

ottmar-zittlau:
The extension of the BUILD-bazel-file in the python subfolder has been removed recently (between tag v24.4 and v25.0-rc1).
If I now run `python setup.py build` in this folder on Windows, an error occurs because - as far as I can tell - the name of the build directory (in lowercases) and the name of this file coincide.


zhangskz:
Thanks for the PR! Looks like this was already addressed in https://github.com/protocolbuffers/protobuf/commit/7e0e8ef385538b074a895a8d824075f7e96c20d9 as discussed in https://github.com/protocolbuffers/protobuf/pull/14600.

PR #14670, opened 2023-11-07
Recover version.json to main/26-dev

shaod2:


shaod2:
Seems like https://github.com/protocolbuffers/protobuf/actions/runs/6788246701/job/18452841139?pr=14670 is failing on finding a 26 version file

zhangskz:
Thanks! I think this might've been an issue with our accidental late Copybara import since the internal MergeToMain PR actually looked correct here: https://github.com/protocolbuffers/protobuf/pull/14617/files#diff-093664cc0c7cf4a76165141e93265eeb139f655bbee8105ed3ee2534a612354c

This still needs to land in Piper first though (so have Copybara import as a CL to land, or import this manually).

zhangskz (approved):

shaod2:
Done separately

PR #14703, opened 2023-11-09
Add support for extensions in CRuby, JRuby, and FFI Ruby

### LINKED ISSUE #1198: No way to use custom options in Ruby

JasonLunn:
Follow up to #14594, which added support for custom options, this PR implements extensions support, which should fully resolve #1198.


src/google/protobuf/compiler/ruby/ruby_generator.cc
```diff
@@ -323,6 +330,9 @@ bool GenerateFile(const FileDescriptor* file, io::Printer* printer,
   for (int i = 0; i < file->enum_type_count(); i++) {
     GenerateEnumAssignment("", file->enum_type(i), printer);
   }
+  for (int i = 0; i < file->extension_count(); i++) {
```
haberman, line 333: This handles the case of top-level extensions, but we also need to handle the case of extensions nested in other messages.

Maybe add a test for this case, and then change the code generator accordingly?

JasonLunn: Added, plus tests for MessageSet.

haberman (approved):

ruby/ext/google/protobuf_c/defs.h
```diff
@@ -79,4 +79,7 @@ extern VALUE generated_pool;
 // Call at startup to register all types in this module.
 void Defs_register(VALUE module);
 
+// FieldDescriptor
+VALUE FieldDescriptor_get(VALUE _self, VALUE msg_rb);
```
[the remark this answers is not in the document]
JasonLunn: Reverted

haberman (approved):

PR #14720, opened 2023-11-10
constexpr causes undefined references

dawidcha:
It looks to me like the build is currently failing with undefined references to kDescriptorMethods?

This change fixes the issue for me at least.


hlopko:
Hi, could you please sign the CLA? We can't accept your contribution without it. Thanks!

PR #14789, opened 2023-11-17
[C#] Fix trimming warning in JSON formatter enum handling

JamesNK:
There is a new trimming/AOT warning in JSON formatter enum handling. I have fixed it by suppressing the value.
I also tested the solution with the .NET 8 SDK and suppressed some other warnings that came up (they're already handled).

It would be great to include this fix in a 25.x release.


jskeet:
I'll try to get to this on Monday, but I make no promises.

jskeet:
(Obviously it's great that it's small, but I need to dig into AOT/trimming in more detail to understand things.)

jskeet (approved):
Okay, I've looked through the docs, and these look reasonable to me.

@deannagarcia: Please could you add appropriate labels/whatever so that this would be included in 25.2, assuming such a release happens?

JamesNK:
ping @deannagarcia

Just want to make sure this doesn't get left behind.

PR #14792, opened 2023-11-17
Remove inappropriate executable permissions from a header

musicinmybrain:
This file is not script-like, so having the executable bit set makes no sense.


hlopko (approved):

musicinmybrain:
This PR is still valid.

ericsalo (approved):

JasonLunn:
@musicinmybrain - would you mind merging the main branch back into this? All the current tests are passing but a few new configurations were added recently and they seem to be what is keeping this from merging automagically.

musicinmybrain:
> @musicinmybrain - would you mind merging the main branch back into this? All the current tests are passing but a few new configurations were added recently and they seem to be what is keeping this from merging automagically.

Done. Hopefully that takes care of it.

PR #14809, opened 2023-11-20
ruby: return 0 from shared_convert.c in failure cases

dcermak:
Functions declared as returning a value must return a value, otherwise the return value is undefined.
This fixes building protobuf with -Werror=return-type


hlopko:
Hi, could you please sign the CLA? We can't accept your contribution without it. Thank you!

hlopko:
I see, thank you. Therefore I'll have to close this PR.

PR #14842, opened 2023-11-24
Added ability to specify a root namespace as protoc option in CSharp

loicmorvan:


jskeet:
Please edit your pull request description to provide a lot more context here - the purpose of the feature, why the existing features are inadequate etc. We don't take new options lightly, and personally it's not clear to me that this is really necessary.

PR #14847, opened 2023-11-27
python: support passing `memoryview` to `bytes` fields

### LINKED ISSUE #6782: Support using python memoryview for setting bytes data

fandreuz:
`memoryview` is not immutable, but a similar functionality is offered for similar types in other languages (e.g. Java's [UnsafeByteOperations](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/UnsafeByteOperations)). An alternative way to implement this would be to provide a custom type (something like `UnsafeBytes`) to wrap buffer-like objects, and to let `bytes` field accept this object; so the user demonstrates to be well aware of what they are doing.

Fixes #6782


hlopko:
I've just ran the CI and it seems many tests are broken, for example with `TypeError: '<' not supported between instances of 'memoryview' and 'memoryview'`, could you please take a look?

Are you having issues with your local bazel build? Could you try rebasing your PR, and using the latest supported Bazel version?

fandreuz:
Hi @hlopko, thanks for your reply. I've managed to run the tests locally and fixed the problems you mentioned in [a6a1b26](https://github.com/protocolbuffers/protobuf/pull/14847/commits/a6a1b26bf510faf1cb4dec9bcb40fde76315ec51). However, I still can't figure out how to run the tests with `upb` backend, for now I've run only with `python`.

python/google/protobuf/internal/byte_buffer.proto
```diff
@@ -0,0 +1,10 @@
+// TODO
+
+syntax = "proto3";
+
+package google.protobuf.internal;
+
+message BytesBufferMsg {
+  bytes single_bytes = 1;
+  repeated bytes repeated_bytes = 2;
+}
```
fandreuz, line 10:
```suggestion
}

```

fandreuz:
Hi @hlopko, any chance we could resume the review of this PR? Thanks!

Issue #13222, opened 2023-07-06
DiscardUnknownFields unnecessarily triggers SyncRepeatedWithMap


PandaBlueYellow:
**What version of protobuf and what language are you using?**
Version: main
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**
clang 11

**What did you do?**
1. use generated api to insert a value to a map<int32, int32> field
2. Call DiscardUnknown on parent message
3. gdb into DiscardUnknownFields

**What did you expect to see**
no SyncRepeatedWithMap is called

**What did you see instead?**
SyncRepeatedWithMap is called

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**

PandaBlueYellow:
In 3.8, https://github.com/protocolbuffers/protobuf/blob/3.8.x/src/google/protobuf/reflection_ops.cc#L249C9-L268C9
Looks like if map is int, int, L266 was executed, which doesn't cause SyncRepeatedWithMap.

However, in main, https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/reflection_ops.cc#L353C5-L364C74
If the map is int, int, it will go into the branch for handling repeated field

shaod2:
Hi,

Map fields should be both is_repeated and is_map, so I'm not sure why L266 was executed for Map<int, int> in 3.8. It the main branch, it is because is_map is now cheaper than is_repeated so that's why it's called first instead.

PandaBlueYellow:
For 3.8, L251, the check is false because the value field of map<int, int> is int instead of message.

PandaBlueYellow:
@shaod2 for main, the point is that, for map<int, int>, it shouldn't go to L366, which triggers SyncRepeatedWithMap.

shaod2:
Yes I see what you meant. I'm checking the intention behind this.

PandaBlueYellow:
Let me know if this could work
```
    for (const FieldDescriptor* field : fields) {
      // Skip over non-message fields.
      if (field->cpp_type() != FieldDescriptor::CPPTYPE_MESSAGE) {
        continue;
      }   
      // Discard the unknown fields in maps that contain message values.
      const MapFieldBase* map_field = field->is_map() ?
          reflection->MutableMapData(message, field) : nullptr;
[...]

PR #14879, opened 2023-11-29
Fix unnecessary switching of map representation for DiscardUnknownFields

gerben-stavenga:
When unknown fields are discarded using reflection in reflection_ops.cc, unnecessary switching of the internal map representation is triggered.

A map<int,int> field cannot have unknown fields. However, in the current code the check used is

if (field->IsMap() && IsMapValueMessageTyped(field)) {
  // Discard map field 
} else {
  // Discard repeated message field
}

This means that the code proceeds discarding unknown fields for each mapentry using the RepeatedPtrField of MapEntry representation of the map. Causing unnecessary work and unnecessary, potentially big, memory allocation. This PR fixes this.

See also issue https://github.com/protocolbuffers/protobuf/issues/13222

This is also a bug fix. When a map<int, MyProto> is in a dirty state, previously it would be skipped. After this change it will correctly use the repeated path to clean the MyProto sub-entries

### DIFF
--- a/src/google/protobuf/reflection_ops.cc
+++ b/src/google/protobuf/reflection_ops.cc
@@ -327,14 +327,14 @@ void ReflectionOps::DiscardUnknownFields(Message* message) {
       continue;
     }
     // Discard the unknown fields in maps that contain message values.
-    if (field->is_map() && IsMapValueMessageTyped(field)) {
-      const MapFieldBase* map_field =
-          reflection->MutableMapData(message, field);
-      if (map_field->IsMapValid()) {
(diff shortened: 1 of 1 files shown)


fowles (approved):

gerben-stavenga:
It fails on internal google tests, which is hard for me to debug. It might be tests relying on the current wrong behavior?

fowles:
that isn't a failure, it is waiting for internal review by someone.  The FFI failure looks unrelated.  I would give the internal reviewer a day or two to review it.  No action needed from you unless this goes quiet for a week.

PR #14924, opened 2023-12-01
Table driven parser

gerben-stavenga:
At bytedance we don't rely solely on clang and deal with other compilers. Hence we cannot in all cases rely on [[clang::musttail]] attribute. Unfortunately, the non-tail call version has substantial overhead. This is a proposal for a table driven parser implementation not relying on tail-calls. The design tries to optimize for parsing in an environment of a server, not in the environment of simple micro-benchmarks. Ie. the branch predictor, after processing 100k-10^6 other branches in the code while serving a request, has lost a lot of information when returning to a parse. I used fleetbench / BM_Protogen as a benchmark as it tries to mimic the production execution environment somewhat more realistic. 

Cascade lake
```
Comparing ../fleetbench_old/bazel-bin/fleetbench/proto/proto_benchmark to bazel-bin/fleetbench/proto/proto_benchmark
Benchmark                           Time             CPU      Time Old      Time New       CPU Old       CPU New
----------------------------------------------------------------------------------------------------------------
BM_Protogen_Arena                -0.0776         -0.0776      13636598      12578904      13636333      12578701
OVERALL_GEOMEAN                  -0.0776         -0.0776             0             0             0             0
```

Sapphire rapids
```
Comparing ../fleetbench-old/bazel-bin/fleetbench/proto/proto_benchmark to bazel-bin/fleetbench/proto/proto_benchmark
Benchmark                           Time             CPU      Time Old      Time New       CPU Old       CPU New
----------------------------------------------------------------------------------------------------------------
BM_Protogen_Arena                -0.0995         -0.0995       9023751       8126207       9023419       8125952
OVERALL_GEOMEAN                  -0.0995         -0.0995             0             0             0             0
```
So it seems that it could be around 7-10% improvement (note the benchmark is not just parsing but also serializing and building protos so the improvement on the parser alone would be bigger). 

Overview of the design. The speed of iterating of the tag/value pairs of the input buffer is limited by 1) the time to figure out where the next tag/value pair starts (the number of bytes depend on content) and 2) branch misses, which tends to adds ~15 cycles latency. The old generated code parser and the tail-call parser dispatch on field number to code dedicated to parsing the field and storing the data as described by the proto definition. The design of the tail call parser was inspired due to the ability to embed function pointers directly into the tables, just like the old generated code effectively had. Using a regular switch would have necessitated an extra indirection with the table containing an enum describing the field and an additional load from a branch table for the switch. If the branch predictor is 100% accurate OoO execution would largely hide this extra load, which typically happens in micro-benchmarks, but alas in actual use cases this is often not the case.

Here I made an alternative design choice. The wireformat is mostly self-describing (length-delimited submessages being the exception). We use the self-describing property of the wireformat to address issue 1) and we use tables for correctly storing the data. Note, that storing the data is not part of the critical path. Therefore latency is not so important as long as storing the data doesn't suffer frequent branch misses. A branch miss would put the latency immediately on the critical path.

It turns out that one can parse varint, fixed32 and fixed64 fields branchless. For simple singular/optional scalar fields we can also store the fields without branches. The difficult to predict indirect branch of the tail call parser with a lot of potential targets is thus replaced, with a simple conditional branch 
```
if (wiretype == 2) { 
   parse string 
} else { 
   parse scalar
}
```
This has two benefits. By the sheer reduction in options the probability of correct guess is higher (WIN), and there is no need to do a branch target lookup so in case of a mispredict where the branch is on the critical path the latency is 5 cycles less (WIN).

This all based on assumptions that groups/oneofs/maps etc.. are all infrequent.

One last trick is that, for performance we have wiretype 2 vs non-2. The serializer can sort non-2, followed by 2, which greatly simplifies the job of the branch predictor. I saw that support for non-increasing order was added. Note that this sorting of wiretype 2/non-2 also unlocks fast table driven serialization as an added bonus. But that's a different story.

I'm not certain this is a win. It's definitely not a win in ALL cases. Batch jobs that parse rows of databases with the same fields in the same order, would most certainly be better off with the tail call parser. If it does turn out to be a good, I'm sure that there is more opportunity in optimizing the loop further.


hlopko:
We've started reviewing this PR, but it will take a lot of time because of the size of the contribution, thank you for understanding.

Issue #14255, opened 2023-09-29
upb-generator to emit code that can be used in a Windows shared library (DLL)


dawidcha:
**What language does this apply to?**
upb generated code

**Describe the problem you are trying to solve.**
Make upb-generated symbols accessible when compiled into Windows shared libraries (i.e. DLLs)

**Describe the solution you'd like**
Add a parameter to protoc-gen-upb and protoc-gen-upbdefs which causes the necessary annotations to be added

**Describe alternatives you've considered**
I can export all symbols by including a def file with an EXPORTS section, but there is no way since the IMPORTS section was deprecated for clients to consume those symbols

**Additional context**
This is part of an ongoing effort to compile the gRPC libraries as a DLL.
I have added a PR with what I believe are the necessary code changes [here](https://github.com/protocolbuffers/protobuf/pull/14230)

haberman:
I'm unassigning myself from this issue because I'm not working on it myself. But https://github.com/protocolbuffers/protobuf/pull/14981 will fix this issue if/when it is submitted.

PR #14981, opened 2023-12-06
Adding DLL export/import tags to generated public upb API

dawidcha:
I have been collaborating with the grpc developers to make it possible to build that library as a Windows DLL - a couple of PRs were already merged, more like https://github.com/grpc/grpc/pull/34345 are pending.

The grpc library incorporates some upb-generated, and upbdefs-generated code into grpc.dll, which is referenced by other code that consumes the library. Since this is now a DLL, that code doesn't know how to link to these generated symbols because they are not annotated with __declspec(dllimport).

This PR aims to fix that by introducing a parameter 'dllexport_tag' to the upb and upbdefs plugins. That parameter should be a string e.g. MYAPP_DLL and when set, the extern symbols are annotated with a macro with that name. This can either be set externally to __declspec(dllimport) or, as is usual practice, when compiling code into a DLL, the macro <dllexport_tag>_EXPORT (i.e. MYAPP_DLL_EXPORT) is defined, and when consuming the DLL <dllexport_tag>_IMPORT is defined if neither are defined then the MYAPP_DLL macro becomes empty string which is what you want for building a static library.

This is a continuation of https://github.com/protocolbuffers/protobuf/pull/14230

Fixes: https://github.com/protocolbuffers/protobuf/issues/14255

### DIFF
--- a/upb_generator/protoc-gen-upb.cc
+++ b/upb_generator/protoc-gen-upb.cc
@@ -890,16 +890,16 @@ void WriteHeader(const DefPoolPair& pools, upb::FileDefPtr file,
 
   if (options.bootstrap) {
     for (auto message : this_file_messages) {
-      output("extern const $1upb_MiniTable* $0();\n", MessageInitName(message),
-             PadSuffix(options.dllexport_decl));
+      output("extern const$1 upb_MiniTable* $0();\n", MessageInitName(message),
+             PadPrefix(options.dllexport_decl));
     }
     for (auto message : forward_messages) {
-      output("extern const $1upb_MiniTable* $0();\n", MessageInitName(message),
-             PadSuffix(options.dllexport_decl));
+      output("extern const$1 upb_MiniTable* $0();\n", MessageInitName(message),
+             PadPrefix(options.dllexport_decl));
     }
     for (auto enumdesc : this_file_enums) {
-      output("extern const $1upb_MiniTableEnum* $0();\n", EnumInit(enumdesc),
-             PadSuffix(options.dllexport_decl));
+      output("extern const$1 upb_MiniTableEnum* $0();\n", EnumInit(enumdesc),
+             PadPrefix(options.dllexport_decl));
     }
     output("\n");
   }
--- a/upb_generator/protoc-gen-upb.cc
+++ b/upb_generator/protoc-gen-upb.cc
@@ -43,7 +43,7 @@ namespace {
 
 struct Options {
   bool bootstrap = false;
-  std::string dllexport_tag;
+  std::string dllexport_decl;
 };
 
 std::string SourceFilename(upb::FileDefPtr file) {
@@ -888,20 +888,20 @@ void WriteHeader(const DefPoolPair& pools, upb::FileDefPtr file,
       "#endif\n"
       "\n");
 
-  EmitDllExportMacros(options.dllexport_tag, output);
+  EmitDllExportMacros(options.dllexport_decl, output);
 
   if (options.bootstrap) {
     for (auto message : this_file_messages) {
       output("extern const $1upb_MiniTable* $0();\n", MessageInitName(message),
-             PadSuffix(options.dllexport_tag));
+             PadSuffix(options.dllexport_decl));
     }
     for (auto message : forward_messages) {
       output("extern const $1upb_MiniTable* $0();\n", MessageInitName(message),
-             PadSuffix(options.dllexport_tag));
+             PadSuffix(options.dllexport_decl));
     }
     for (auto enumdesc : this_file_enums) {
       output("extern const $1upb_MiniTableEnum* $0();\n", EnumInit(enumdesc),
-             PadSuffix(options.dllexport_tag));
+             PadSuffix(options.dllexport_decl));
     }
     output("\n");
   }
@@ -1112,8 +1112,8 @@ bool ParseOptions(Plugin* plugin, Options* options) {
   for (const auto& pair : ParseGeneratorParameter(plugin->parameter())) {
     if (pair.first == "bootstrap_upb") {
       options->bootstrap = true;
-    } else if (pair.first == "dllexport_tag") {
-      options->dllexport_tag = pair.second;
+    } else if (pair.first == "dllexport_decl") {
+      options->dllexport_decl = pair.second;
     } else {
       plugin->SetError(absl::Substitute("Unknown parameter: $0", pair.first));
       return false;
--- a/upb_generator/protoc-gen-upb.cc
+++ b/upb_generator/protoc-gen-upb.cc
@@ -888,8 +888,6 @@ void WriteHeader(const DefPoolPair& pools, upb::FileDefPtr file,
       "#endif\n"
       "\n");
 
-  EmitDllExportMacros(options.dllexport_decl, output);
-
   if (options.bootstrap) {
     for (auto message : this_file_messages) {
       output("extern const $1upb_MiniTable* $0();\n", MessageInitName(message),
--- a/upb_generator/protoc-gen-upb.cc
+++ b/upb_generator/protoc-gen-upb.cc
@@ -43,6 +43,7 @@ namespace {
 
 struct Options {
   bool bootstrap = false;
+  std::string dllexport_tag;
 };
 
 std::string SourceFilename(upb::FileDefPtr file) {
@@ -732,7 +733,7 @@ void GenerateNonRepeatedSetters(upb::FieldDefPtr field,
 
   if (field == field.containing_type().map_value()) {
     output(R"cc(
-             UPB_INLINE void $0_set_$1($0 *msg, $2 value) {
+             UPB_INLINE void $0_set_$1($0* msg, $2 value) {
                _upb_msg_map_set_value(msg, &value, $3);
              }
            )cc",
@@ -741,7 +742,7 @@ void GenerateNonRepeatedSetters(upb::FieldDefPtr field,
                                               : "sizeof(" + CType(field) + ")");
   } else {
     output(R"cc(
-             UPB_INLINE void $0_set_$1($0 *msg, $2 value) {
+             UPB_INLINE void $0_set_$1($0* msg, $2 value) {
                const upb_MiniTableField field = $3;
                _upb_Message_SetNonExtensionField((upb_Message *)msg, &field, &value);
              }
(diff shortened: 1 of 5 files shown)


esrauchg (approved):

veblush:
LGTM

upb_generator/protoc-gen-upb.cc
```diff
@@ -888,13 +889,16 @@ void WriteHeader(const DefPoolPair& pools, upb::FileDefPtr file,
 
   if (options.bootstrap) {
     for (auto message : this_file_messages) {
-      output("extern const upb_MiniTable* $0();\n", MessageInitName(message));
+      output("extern const$1 upb_MiniTable* $0();\n", MessageInitName(message),
```
haberman, line 892: The "bootstrap" output is only used during bootstrapping and should never be built as part of a DLL. It should only be linked into the `stage0` code generators themselves, which should only be statically linked.

hlopko:
@dawidcha could you resolve the comment from @haberman? We'll then look into the internl failure (which seems to be an issue with some pre-generated files that just need to be regenerated together with your PR).

dawidcha:
Looks like the CLA blocker is finally unblocked
I've rebased and I think we should be ready to take this over the line

upb_generator/protoc-gen-upb.cc
haberman: It looks like this comment has not been addressed. Nothing in the bootstrap stage should be affected by `dllexport_decl`.

h-vetinari:
> @dawidcha could you resolve the comment from @haberman?

That still seems to be open, but assuming this is not that hard to fix, it'd be awesome if this could still make v26.0! 🙃

It's causing [issues](https://github.com/grpc/grpc/issues/35794) in grpc, for example.

upb_generator/protoc-gen-upb.cc
dawidcha: Sorry missed this - looking now to see...

h-vetinari: @haberman, a comment from the peanut gallery here[^1]

> [...] should never be built as part of a DLL.

Building grpc as a shared library did run into exactly these missing symbols though... As an example (taken from 1.57), it required:
```diff
diff --git a/src/core/ext/upb-generated/src/proto/grpc/gcp/altscontext.upb.h b/src/core/ext/upb-generated/src/proto/grpc/gcp/altscontext.upb.h
index 28c2185afd..68b940b360 100644
--- a/src/core/ext/upb-generated/src/proto/grpc/gcp/altscontext.upb.h
+++ b/src/core/ext/upb-generated/src/proto/grpc/gcp/altscontext.upb.h
@@ -27,7 +27,7 @@ extern "C" {
 
 typedef struct grpc_gcp_AltsContext grpc_gcp_AltsContext;
 typedef struct grpc_gcp_AltsContext_PeerAttributesEntry grpc_gcp_AltsContext_PeerAttributesEntry;
-extern const upb_MiniTable grpc_gcp_AltsContext_msg_init;
+extern const GRPC_DLL upb_MiniTable grpc_gcp_AltsContext_msg_init;
 extern const upb_MiniTable grpc_gcp_AltsContext_PeerAttributesEntry_msg_init;
 struct grpc_gcp_RpcProtocolVersions;
 extern const upb_MiniTable grpc_gcp_RpcProtocolVersions_msg_init;
diff --git a/src/core/ext/upb-generated/src/proto/grpc/health/v1/health.upb.h b/src/core/ext/upb-generated/src/proto/grpc/health/v1/health.upb.h
index 9217ea2312..e8a4b2b0c5 100644
--- a/src/core/ext/upb-generated/src/proto/grpc/health/v1/health.upb.h
+++ b/src/core/ext/upb-generated/src/proto/grpc/health/v1/health.upb.h
@@ -27,8 +27,8 @@ extern "C" {
 
 typedef struct grpc_health_v1_HealthCheckRequest grpc_health_v1_HealthCheckRequest;
 typedef struct grpc_health_v1_HealthCheckResponse grpc_health_v1_HealthCheckResponse;
-extern const upb_MiniTable grpc_health_v1_HealthCheckRequest_msg_init;
-extern const upb_MiniTable grpc_health_v1_HealthCheckResponse_msg_init;
+extern const GRPC_DLL upb_MiniTable grpc_health_v1_HealthCheckRequest_msg_init;
+extern const GRPC_DLL upb_MiniTable grpc_health_v1_HealthCheckResponse_msg_init;
 
 typedef enum {
   grpc_health_v1_HealthCheckResponse_UNKNOWN = 0,
diff --git a/src/core/ext/upb-generated/xds/data/orca/v3/orca_load_report.upb.h b/src/core/ext/upb-generated/xds/data/orca/v3/orca_load_report.upb.h
index 163152c592..3fb692e896 100644
--- a/src/core/ext/upb-generated/xds/data/orca/v3/orca_load_report.upb.h
+++ b/src/core/ext/upb-generated/xds/data/orca/v3/orca_load_report.upb.h
@@ -29,7 +29,7 @@ typedef struct xds_data_orca_v3_OrcaLoadReport xds_data_orca_v3_OrcaLoadReport;
 typedef struct xds_data_orca_v3_OrcaLoadReport_RequestCostEntry xds_data_orca_v3_OrcaLoadReport_RequestCostEntry;
 typedef struct xds_data_orca_v3_OrcaLoadReport_UtilizationEntry xds_data_orca_v3_OrcaLoadReport_UtilizationEntry;
 typedef struct xds_data_orca_v3_OrcaLoadReport_NamedMetricsEntry xds_data_orca_v3_OrcaLoadReport_NamedMetricsEntry;
-extern const upb_MiniTable xds_data_orca_v3_OrcaLoadReport_msg_init;
+extern const GRPC_DLL upb_MiniTable xds_data_orca_v3_OrcaLoadReport_msg_init;
 extern const upb_MiniTable xds_data_orca_v3_OrcaLoadReport_RequestCostEntry_msg_init;
 extern const upb_MiniTable xds_data_orca_v3_OrcaLoadReport_UtilizationEntry_msg_init;
 extern const upb_MiniTable xds_data_orca_v3_OrcaLoadReport_NamedMetricsEntry_msg_init;
```

So not all `upb_MiniTable` were affected (see surrounding lines), but it does at least seem to happen quite naturally that these symbols end up in the final DLL. The affected files correspond to the [workaround](https://github.com/grpc/grpc/blob/v1.59.0/CMakeLists.txt#L430-L444) Dawid had to introduce in the initial support for shared builds in grpc.

Note that at the time, this was happening in the checked-in, upb-generated sources. For [grpc 1.61](https://github.com/grpc/grpc/issues/35794), it wasn't necessary to patch stuff in [`upb-gen`](https://github.com/grpc/grpc/tree/master/src/core/ext/upb-gen), but rather I needed to patch the grpc-vendored upb itself (again for `upb_MiniTable` instances) to get shared builds to compile at all:
```diff
diff --git a/third_party/upb/upb/mini_table/internal/message.c b/third_party/upb/upb/mini_table/internal/message.c
index b25edacca7..05da188fc7 100644
--- a/third_party/upb/upb/mini_table/internal/message.c
+++ b/third_party/upb/upb/mini_table/internal/message.c
@@ -7,7 +7,7 @@
 
 #include "upb/mini_table/internal/message.h"
 
-const struct upb_MiniTable _kUpb_MiniTable_Empty = {
+UPB_DLL const struct upb_MiniTable _kUpb_MiniTable_Empty = {
     .subs = NULL,
     .fields = NULL,
     .size = 0,
diff --git a/third_party/upb/upb/mini_table/internal/message.h b/third_party/upb/upb/mini_table/internal/message.h
index 3457ea2cd6..8fa3325911 100644
--- a/third_party/upb/upb/mini_table/internal/message.h
+++ b/third_party/upb/upb/mini_table/internal/message.h
@@ -65,7 +65,7 @@ extern "C" {
 #endif
 
 // A MiniTable for an empty message, used for unlinked sub-messages.
-extern const struct upb_MiniTable _kUpb_MiniTable_Empty;
+UPB_DLL extern const struct upb_MiniTable _kUpb_MiniTable_Empty;
 
 // Computes a bitmask in which the |l->required_count| lowest bits are set,
 // except that we skip the lowest bit (because upb never uses hasbit 0).
```

If I understand you correctly, you seem to be saying that this is a bug in the way grpc does shared windows builds? 
Seeing that this always seems to involve `upb_MiniTable`, perhaps it's "too easy" to make that mistake though? (not that very many people are doing shared builds on windows...)

[^1]: I help maintain protobuf&grpc builds in the conda-forge distribution, we've been doing shared grpc builds on windows since 1.55, which was the jumping off point for the work that @dawidcha has been driving forward

haberman: > As an example (taken from 1.57), it required:

All of the symbols you quoted (`extern const struct`) should indeed get the DLL export macro, I agree. My comment was not about them.

My comment was about a different set of generated files, a special and very specific kind of generated code that is only used when building upb itself and its core internal protos `descriptor.proto` and `plugin.proto`.

Because these protos are used internally by upb, these protos have to go through a special bootstrapping process when building upb from scratch in the GitHub repo. The bootstrapped version of these protos does not use `extern const struct`; it actually emits *functions* rather that const structs for the MiniTables.  You can see this in the very line we are discussing on this comment thread: https://github.com/protocolbuffers/protobuf/blob/fe6edca15b2d80df79ab31117dc49621ef461202/upb_generator/protoc-gen-upb.cc#L892

You can also see what the bootstrap generated code looks like, as we check it in to the source repo: https://github.com/protocolbuffers/protobuf/blob/fe6edca15b2d80df79ab31117dc49621ef461202/upb/reflection/stage0/google/protobuf/descriptor.upb.h#L20-L68

As you can see, this is clearly a different set of generated code than the samples you quoted, which should indeed get the macro, I agree.

It should be very easy to resolve my comment on this thread by simply reverting the change on this line, while still providing the export macros needed by end users.

h-vetinari: Thanks a lot for the response!

> As you can see, this is clearly a different set of generated code than the samples you quoted, which should indeed get the macro, I agree.

Yeah, I'm not surprised that there are other ways where `upb_MiniTable` things get generated, even though I had tried to double-checked whether I had missed other occurrences in `protoc-gen-upb.cc`. Wonderful news if it's just a one-line revert that's necessary!

h-vetinari: > It should be very easy to resolve my comment on this thread by simply reverting the change on this line, while still providing the export macros needed by end users.

@dawidcha, are you still planning to work on this PR? Otherwise I can try my hand at carrying this effort into another PR (or you invite me to your fork so I can push into this PR directly).

bebuch: @h-vetinari It would be great if you could finish this in a new PR!

@dawidcha Thanks for the awesome work so far!

h-vetinari: I've continued this PR in https://github.com/protocolbuffers/protobuf/pull/17079 and removed the changes to the bootstrap case.

PR #15017, opened 2023-12-08
protoc: support inf, -inf, nan, and -nan in option values

### LINKED ISSUE #15010: protoc: parsing of float/double numeric values for options is inconsistent
The parser has a special grammar for parsing default values that allows infinity, -infinity, and NaN. However, this grammar is not used when parsing other numeric option values.

Take the following example:
```proto
syntax = "proto2";

extend google.protobuf.MessageOptions {
  optional Foo foo = 10101;
}

message Foo {
  optional double val = 1 [default = -nan]; // "nan", "-nan", "inf", and "-inf" allowed here
  repeated float vals = 2;

  option (foo) = { val: -infINITY }; // aside: inside a message literal, anything goes 😭

  option (foo).vals = inf; // "nan", "-nan", "inf", and "-inf" NOT allowed here (??)
  // The above line results in the following error from protoc:
  //    test.proto:15:23: Value must be number for float option "Foo.vals".
  // If the value were -inf (negative), the message changes:
  //    test.proto:15:24: Invalid '-' symbol before identifier.
}
```
I assert that it's a bug that infinity and NaN can only be used in default values and not other option values. It seems like a simple overlooked thing that is easy to fix.

(As shown above, the contents of message literals, which are parsed using the text format, are _very_ lenient. I'm certainly not suggesting that anything else in `protoc` should be that forgiving -- just pointing it out in the example since it's... interesting.)

jhump:
The parser already supports these values for the special "default" field option. But, inconsistently, does not support them for other options whose type is float or double. This addresses that inconsistency.

Fixes #15010.

### DIFF
--- a/src/google/protobuf/compiler/parser_unittest.cc
+++ b/src/google/protobuf/compiler/parser_unittest.cc
@@ -12,6 +12,7 @@
 #include "google/protobuf/compiler/parser.h"
 
 #include <algorithm>
+#include <cmath>
 #include <memory>
 #include <string>
 #include <utility>
@@ -19,6 +20,7 @@
 
 #include "google/protobuf/any.pb.h"
 #include "google/protobuf/descriptor.pb.h"
+#include "google/protobuf/dynamic_message.h"
 #include <gmock/gmock.h>
 #include "google/protobuf/testing/googletest.h"
 #include <gtest/gtest.h>
@@ -460,6 +462,7 @@ TEST_F(ParseMessageTest, FieldDefaults) {
       "  required double foo = 1 [default= inf ];\n"
       "  required double foo = 1 [default=-inf ];\n"
       "  required double foo = 1 [default= nan ];\n"
+      "  required double foo = 1 [default= -nan ];\n"
       "  required string foo = 1 [default='13\\001'];\n"
       "  required string foo = 1 [default='a' \"b\" \n \"c\"];\n"
       "  required bytes  foo = 1 [default='14\\002'];\n"
@@ -508,6 +511,8 @@ TEST_F(ParseMessageTest, FieldDefaults) {
       "  field { type:TYPE_DOUBLE  default_value:\"-inf\"      " ETC
       " }"
       "  field { type:TYPE_DOUBLE  default_value:\"nan\"       " ETC
+      " }"
+      "  field { type:TYPE_DOUBLE  default_value:\"-nan\"      " ETC
       " }"
       "  field { type:TYPE_STRING  default_value:\"13\\001\"   " ETC
       " }"
@@ -659,7 +664,6 @@ TEST_F(ParseMessageTest, FieldOptionsSupportLargeDecimalLiteral) {
 }
 
 TEST_F(ParseMessageTest, FieldOptionsSupportInfAndNan) {
-  // decimal integer literal > uint64 max
   ExpectParsesTo(
       "import \"google/protobuf/descriptor.proto\";\n"
       "extend google.protobuf.FieldOptions {\n"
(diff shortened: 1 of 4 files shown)


acozzette (changes requested):
Thank you for the fix and sorry for taking so long to get to this. This looks great to me but would you mind also adding some examples to unittest_custom_options.proto and another short unit test case in parser_unittest.cc that checks that we can read the correct values for those options? I think it would be good to just have test coverage that validates the options after they're interpreted.

jhump:
> I think it would be good to just have test coverage that validates the options after they're interpreted.

@acozzette, sorry for the long hiatus from the branch!

This was a good idea since I totally messed up my original change and forgot to actually update the options interpreting step to allow identifiers "inf" and "nan" for double and float values 🤦.

I've done as requested... but the new test case in parser_unittest.cc is _really_ complicated. We can't easily extract values from a parsed/linked descriptor since interpreting options stores them in the unrecognized field set. So we have to serialize and de-serialize them. I'm using a dynamic message to avoid issues where the `FileOptions` descriptor from the compiler output != the descriptor returned by the generated type. (Though I think I could also get around that by instead creating a new pool that uses the generated pool as its fallback database, so it wouldn't try to rebuild a descriptor from the `FileDescriptorProto` for `google/protobuf/descriptor.proto`.)

Anyhow. I wonder if this new test is really necessary? There is an existing test (`ParseDescriptorDebugTestTestCustomOptions`) that indirectly exercises much of the same, by effectively compiling `unittest_custom_options.proto` and making sure the output matches the embedded descriptor of the generated code for that file. Is that sufficient? If not, is there a way to at least make the new test a little more readable? I'm not as familiar with all of the helpers in this test file, but it doesn't look like much else is actually going through the full `pool.BuildFile` step to link the descriptors and interpret options 🤷.

src/google/protobuf/compiler/parser_unittest.cc
```diff
@@ -509,6 +512,8 @@ TEST_F(ParseMessageTest, FieldDefaults) {
       " }"
       "  field { type:TYPE_DOUBLE  default_value:\"nan\"       " ETC
       " }"
+      "  field { type:TYPE_DOUBLE  default_value:\"-nan\"      " ETC
```
jhump, line 515: For the heck of it, I decided to make this test case a little more thorough and add the `-nan` case. And I was slightly surprised that this assertion insisted that there be a minus sign in front of this NaN value. I think maybe this helper is generating the compiled output to the text format, too, and then checking the output matches? Or maybe it's doing some other equality test on the output compared to the result of unmarshaling this string? In any event, perhaps something under the hood is directly examining the sign bit to decide whether this needs to have a sign.

(I didn't dig into the code, but I'd be a little surprised if `< 0` returned true for any NaN, regardless of the sign bit 🤷. And I'd also be surprised if any comparisons of two NaN values _ever_ returned "equal", whether they have sign bits set or not, unless something were actually comparing their bits directly?)

acozzette: I think this makes sense, because the `default_value` field is documented [here](https://github.com/protocolbuffers/protobuf/blob/7dc243c70a870198b80ba4923e8bb8b251e87564/src/google/protobuf/descriptor.proto#L283) to contain "the original text representation of the value." So it sounds like this field is expected to contain the original text without any interpretation. I haven't tested this but I assume if the default value were specified as 1.000 for example then it wouldn't be simplified down to 1 here.

acozzette:
@jhump How about a test that just looks at the generated descriptors from the protos you added in unittest_custom_options.proto and verifies that the options have the expected values? That should be a lot simpler because then you can just rely on the generated descriptor pool instead of building your own.

src/google/protobuf/compiler/parser_unittest.cc
jhump: Ah. Interestingly, that comment is a bit misleading since it's not _quite_ true: `string` and `bytes` values are decoded (and then, for `bytes` fields, [re-encoded](https://github.com/protocolbuffers/protobuf/blob/v26.1/src/google/protobuf/compiler/parser.cc#L1433)). And, in the example you give with "1.000" as a value, the precision would not actually be preserved because it is just [formatting the numeric value to string](https://github.com/protocolbuffers/protobuf/blob/v26.1/src/google/protobuf/compiler/parser.cc#L1407-L1409). It just happens to capture the negative sign independently and [store that in the value first](https://github.com/protocolbuffers/protobuf/blob/v26.1/src/google/protobuf/compiler/parser.cc#L1399-L1402).

Anyhow, I guess it makes sense now why this value needs to look this way.

jhump:
> How about a test that just looks at the generated descriptors from the protos you added in unittest_custom_options.proto and verifies that the options have the expected values?

@acozzette, okay, I've replaced the complicated test with that. Definitely simpler.

acozzette (approved):

jhump:
@acozzette, where does this PR go from here? Does it need to be tested internally before it can actually be merged?

acozzette:
@jhump Sorry for the delay, I was out for a few days last week and lost track of this. Would you mind merging main into your PR (or rebasing)? I'm hitting an issue syncing it into our internal system which I think should be fixed by that. After that it should be straightforward to test it internally and land the change.

jhump:
> Would you mind merging main into your PR (or rebasing)?

@acozzette, done.

acozzette (approved):

acozzette (approved):

acozzette (approved):

acozzette:
@jhump The only test failure I ran into is [this one](https://github.com/bufbuild/protocompile/blob/792268afce75e99f4a1fd98c3597529032154ffe/parser/validate_test.go#L701), which makes sense because it expects to see the current protoc behavior of rejecting values like `-inf`. I assume the right fix is just to delete the line with `expectedDiffWithProtoc: true`?

jhump:
@acozzette, yes, that would be the right fix (and what I will do in that repo as soon as this fix is released and that repo updated to use  that latest `protoc` release).

PR #15113, opened 2023-12-15
Fix compiling problems with lua 5.4

### LINKED ISSUE #13750: Lua binding compatibility for lua 5.3+
1. `luaL_checkint` is replaced by `luaL_checkinteger` and is removed from lua 5.3
2. `lua_isinteger` is already defined in lua 5.3+
3. `else if (isprint(ch))` in `upb/bindings/lua/upbc.cc` is UB when ch is negative, and will crash when compiled by MSVC.

### LINKED ISSUE #15112: [UPB] Support lua 5.1 or upper
Version: v25.1
Language: C++/lua

Linux,Windows

MSVC,gcc

**What did you do?**

Just compiling.

Some functions changed in lua 5.1 or upper. And `isprint(ch)` is UB when `ch` is negative and it may crash when compiling with MSVC.

owent:
Fixes #15112
Fixes #13750

### DIFF
--- a/lua/def.c
+++ b/lua/def.c
@@ -597,7 +597,7 @@ const upb_FileDef* lupb_FileDef_check(lua_State* L, int narg) {
 
 static int lupb_FileDef_Dependency(lua_State* L) {
   const upb_FileDef* f = lupb_FileDef_check(L, 1);
-  int index = luaL_checkint(L, 2);
+  int index = lupb_checkint32(L, 2);
   const upb_FileDef* dep = upb_FileDef_Dependency(f, index);
   lupb_wrapper_pushwrapper(L, 1, dep, LUPB_FILEDEF);
   return 1;
@@ -611,7 +611,7 @@ static int lupb_FileDef_DependencyCount(lua_State* L) {
 
 static int lupb_FileDef_enum(lua_State* L) {
   const upb_FileDef* f = lupb_FileDef_check(L, 1);
-  int index = luaL_checkint(L, 2);
+  int index = lupb_checkint32(L, 2);
   const upb_EnumDef* e = upb_FileDef_TopLevelEnum(f, index);
   lupb_wrapper_pushwrapper(L, 1, e, LUPB_ENUMDEF);
   return 1;
@@ -625,7 +625,7 @@ static int lupb_FileDef_enumcount(lua_State* L) {
 
 static int lupb_FileDef_msg(lua_State* L) {
   const upb_FileDef* f = lupb_FileDef_check(L, 1);
-  int index = luaL_checkint(L, 2);
+  int index = lupb_checkint32(L, 2);
   const upb_MessageDef* m = upb_FileDef_TopLevelMessage(f, index);
   lupb_wrapper_pushwrapper(L, 1, m, LUPB_MSGDEF);
   return 1;
--- a/lua/upbc.cc
+++ b/lua/upbc.cc
@@ -5,13 +5,13 @@
 // license that can be found in the LICENSE file or at
 // https://developers.google.com/open-source/licenses/bsd
 
-#include "google/protobuf/descriptor.pb.h"
 #include "absl/strings/str_replace.h"
 #include "absl/strings/string_view.h"
 #include "absl/strings/substitute.h"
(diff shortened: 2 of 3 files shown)


lua/def.c
```diff
@@ -597,7 +597,11 @@ const upb_FileDef* lupb_FileDef_check(lua_State* L, int narg) {
 
 static int lupb_FileDef_Dependency(lua_State* L) {
   const upb_FileDef* f = lupb_FileDef_check(L, 1);
+#if defined(LUA_VERSION_NUM) && LUA_VERSION_NUM >= 501
+  int index = luaL_checkinteger(L, 2);
```
esrauchg, line 601: Maybe these lines should all be changed to be `lupb_checkint32` instead? checkinteger() is a ptrdiff_t, so on 64-bit targets that means it'll be an int64 and 'int' is an int32 so this line is an implicit cast as you wrote it.

If _checkint32 is not appropriate for some reason, can you make a lupb_checkinteger() in upb.h/.c which just internally does the #ifdefined and uses check_int or check_integer based on which lua version supports it? It'll be cleaner than repeating the #ifs each time in this file (and have a place to have a comment about the large-ints overflow behavior)

lua/upbc.cc
```diff
@@ -61,7 +61,7 @@ static void PrintString(int max_cols, absl::string_view* str,
     } else if (ch == '\'') {
       printer->PrintRaw("\\'");
       max_cols--;
-    } else if (isprint(ch)) {
+    } else if (isprint(static_cast<int>(static_cast<unsigned char>(ch)))) {
```
esrauchg, line 64: Can you make a helper for this, and do the <0 check explicitly instead of casting through:

```
static void IsPrint(int ch) {
  // isprint(ch) with negative values is UB.
  return ch < 0 ? false : isprint(ch);
}
```

esrauchg (commented):
Thanks for the contribution!

lua/def.c
```diff
@@ -597,7 +597,11 @@ const upb_FileDef* lupb_FileDef_check(lua_State* L, int narg) {
 
 static int lupb_FileDef_Dependency(lua_State* L) {
   const upb_FileDef* f = lupb_FileDef_check(L, 1);
+#if defined(LUA_VERSION_NUM) && LUA_VERSION_NUM >= 501
```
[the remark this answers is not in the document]
owent: Lua 5.0 do not declare `LUA_VERSION_NUM`.Do we support lua 5.0 and before?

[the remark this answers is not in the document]
esrauchg: Yep: lua/upb.h already uses LUA_VERSION_NUMBER without a defined() check and explicitly says it only supports 5.1-5.4.

So I would match that preexisting behavior and not worry about trying to support 5.0 in this PR. Thanks!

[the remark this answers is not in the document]
owent: Thanks, `defined(LUA_VERSION_NUMBER) &&` is removed.

lua/upbc.cc
owent: Thanks and fixed.

lua/def.c
```diff
@@ -597,7 +597,11 @@ const upb_FileDef* lupb_FileDef_check(lua_State* L, int narg) {
 
 static int lupb_FileDef_Dependency(lua_State* L) {
   const upb_FileDef* f = lupb_FileDef_check(L, 1);
+#if LUA_VERSION_NUM >= 501
```
esrauchg, line 600: Since we only support lua_version > 501, you can just drop this #if and only keep the corresponding case, and then the same below on lines 618 and 636 (and similarly in those cases use lupb_checkint32 since _checkinteger won't be defined on 5.1)

owent: Thanks and fixed.

esrauchg (approved):

owent:
Is there anything I can do to let this PR continue?

googleberg:
For some reason tests never ran.

owent:
> Apologies for the dead air. For some reason tests never ran. Unfortunately, it looks like this PR will need a rebase and some merge conflict resolution.

Done

PR #15114, opened 2023-12-15
Add bootstrap build for upb

### LINKED ISSUE #13751: Serveral problem of cmake users(output name, including path and installation config)
Hi, I'm a cmake user and I tried to use `cmake/make_cmakelists.py` to generate `CMakeLists.txt` and integrate upb into my project and find some problems.

1. The generated dependency of `utf8_range` is invalid because it has a `/third_party/` prefix.
2. CMake CONFIG files and is missing and also nothing to install when I try to create a prebuilt package.
3. Should all exported header be under `upb/` ? The `third_party/utf8_range/utf8_range.h` is a different path from others.
4. Could all output name of libraries in upb startswith `upb` ? There are a lot of targets (`extension_registry`, `mini_table`, `fastdecode` and etc) now , and if we install them into any path, we can not tell which files are belong to upb.

Could we have a disscuss these topics and can I create a PR for them?

### LINKED ISSUE #15111: [Feature Request] Export cmake CONFIG package for upb.
C++

Using cmake to build generator and libraries of upb and lua bindings

**Additional context**

It will be easier to use for cmake users when we export cmake CONFIG package.This package file will help users to resolve the dependencies and link order.
All the dependencies of upb(abesil-cpp, utf_range) support cmake CONFIG package now and gRPC which depends upb also support it.It will be nice and easier to maintain if upb also support it.

owent:
Fixes #15111 
Fixes #13751

### DIFF
--- a/upb/cmake/CMakeLists.txt
+++ b/upb/cmake/CMakeLists.txt
@@ -107,6 +107,21 @@ if (UPB_ENABLE_CODEGEN)
       set(UPB_HOST_INCLUDE_DIR "${PROTOBUF_INCLUDE_DIR}")
     endif()
   endif()
+  find_package(Lua QUIET)
+  if (NOT TARGET lua::liblua AND (Lua_FOUND OR LUA_FOUND))
+    foreach(TEST_LUA_TARGET_NAME lua::liblua-dynamic lua::liblua-static unofficial-lua::lua)
+      if(TARGET ${TEST_LUA_TARGET_NAME})
+        add_library(lua::liblua ALIAS ${TEST_LUA_TARGET_NAME})
+        break()
+      endif()
+    endforeach()
+    if (NOT TARGET lua::liblua AND LUA_FOUND)
+      add_library(lua::liblua INTERFACE IMPORTED)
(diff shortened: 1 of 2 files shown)


ericsalo:
Thank you for this submission, it's clear you put a fair amount of effort into this and we appreciate it. Unfortunately we are already working on fixing the cmake support for upb in a way which ties it much more strongly to the existing protobuf cmake support.

owent:
> Thank you for this submission, it's clear you put a fair amount of effort into this and we appreciate it. Unfortunately we are already working on fixing the cmake support for upb in a way which ties it much more strongly to the existing protobuf cmake support.

Thank you, I have been eagerly awaiting this for months. Prior to the completion of this implementation, we have an internal patch to bootstrap build upb with Lua binding, and we also use gRPC (which depends on upb). However, the `make_cmakelists.py` has undergone significant changes in recent months, requiring us to repeatedly rework the patch. Therefore, I am wondering if we could use a temporary solution until the final implementation is completed.

calvin2021y:
@ericsalo  any news ?

PR #15193, opened 2023-12-26
ruby: Fix format-security warning when compiling the Ruby Gem

nathantalewis:
When compiling the ruby gem with `-Werror=format-security` there are some strings that might contain formatting characters. This commit makes sure those calls are formatted with "%s" first so that any unintentional formatting characters are successfully passed through without being interpreted.


esrauchg (approved):

nathantalewis:
Looks like this was fixed in `main` in the mean time. Closing.

PR #15236, opened 2023-12-29
Do no longer depend on deprecated //external:python_headers

mering:
`bind()` targets are deprecated and unsupported with Bzlmod. Specifying the dependency directly as a mitigation.

Alternative: Define an `alias()` within `/third_party/BUILD`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -191,11 +191,6 @@ load("@fuzzing_py_deps//:requirements.bzl", fuzzing_py_deps_install_deps = "inst
 
 fuzzing_py_deps_install_deps()
 
-bind(
-    name = "python_headers",
-    actual = "@system_python//:python_headers",
-)
-
 http_archive(
     name = "rules_rust",
     sha256 = "9ecd0f2144f0a24e6bc71ebcc50a1ee5128cedeceb32187004532c9710cb2334",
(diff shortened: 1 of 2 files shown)


mering:
Any idea why some tests are skipped and therefore failing the jobs?

acozzette:
I don't think there's a problem with skipped tests but I do see some real errors like this one:
```
ERROR: /workspace/_build/out/external/rules_python/python/cc/BUILD.bazel:14:22: While resolving toolchains for target @rules_python//python/cc:current_py_cc_headers: No matching toolchains found for types @rules_python//python/cc:toolchain_type.
```

mering:
@acozzette I now changed it to use exactly the same as the original `bind()` rule. Can you run the tests again?

mering:
All tests are passing now.

acozzette:
@mering Thank you for the fix, this looks great to me except could you update it to drop the alias and have us just refer directly to `@system_python//:python_headers`?

mering:
> @mering Thank you for the fix, this looks great to me except could you update it to drop the alias and have us just refer directly to `@system_python//:python_headers`?

@acozzette Sure. I updated the PR description to reflect his.

acozzette (approved):

PR #15265, opened 2024-01-03
Fix separate issues in JRuby's "native" `dup` and `inspect` methods.

JasonLunn:
Update the stress test to exercise `proto3_optional` fields.


haberman:
I suspect there are more bugs like this lurking. It appears that we have two parallel data structures where neither one is the source of truth.

If we move towards the direction of FFI, we may be able to sidestep facing this issue head-on.

haberman (approved):

PR #15296, opened 2024-01-05
Add missing header file for CSharp

gichan-jang:
Add missing header file "cshap_names.h" for CSharp.

Signed-off-by: gichan2-jang <gichan2.jang@samsung.com>

### DIFF
--- a/src/google/protobuf/compiler/csharp/BUILD.bazel
+++ b/src/google/protobuf/compiler/csharp/BUILD.bazel
@@ -58,6 +58,7 @@ cc_library(
         "csharp_repeated_primitive_field.h",
         "csharp_source_generator_base.h",
         "csharp_wrapper_field.h",
+        "csharp_names.h",
     ],
     copts = COPTS + select({
         "//build_defs:config_msvc": [],


src/google/protobuf/compiler/csharp/BUILD.bazel
```diff
@@ -58,6 +58,7 @@ cc_library(
         "csharp_repeated_primitive_field.h",
         "csharp_source_generator_base.h",
         "csharp_wrapper_field.h",
+        "csharp_names.h",
```
zhangskz, line 61: I don't think csharp_names.h exists as a file anymore, as it was renamed to names.h in https://github.com/protocolbuffers/protobuf/commit/01fe22219a0312b178a265e75fe35422ea6afbb1#diff-587b2ebc8febdd2143ad0791c27371a5152f5ce50851a9bb2b059634f68c925d.

Can you provide more context on why you believe this header is missing and what this is meant to fix?

gichan-jang: An error occurred while building the previous version of grpc.
I think I need to update the version of grpc.
Thanks. Close this PR.

Issue #14790, opened 2023-11-17
[C#] Update .NET solution to use .NET 8 SDK


JamesNK:
The .NET solution currently uses an old .NET 6 SDK:

https://github.com/protocolbuffers/protobuf/blob/4864e43b7d62a30bf335f1b5ba56d07515b3f959/global.json#L3

It should be updated to .NET 8 to keep tooling modern.

jskeet:
This is definitely something to aspire to, but it may take longer than expected due to complexities in terms of *internal* tooling.

embeddetech:
Wow, @JamesNK wrote Newtonsoft.Json??? 

https://images.app.goo.gl/yjNLXn18L9A7XqyX8

jskeet:
> I'm also needing to use protobuf with .NET 8 soon...

There's nothing stopping you doing that now. You don't need the Google.Protobuf library to explicitly target .NET 8, or for it to be built with the .NET 8 SDK, for you to use it from .NET 8.

jskeet:
Reopening, as we do want to do this eventually.

JamesNK:
FYI .NET 6 is end-of-life November 12, 2024

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

jskeet:
Yup, we definitely need to get on this - if only I could clone myself...

JamesNK:
Still need

jskeet:
cc @amanda-tarafa - we might want to see what our team can do to help the Protobuf team on this one.

Issue #14523, opened 2023-10-25
Add NET8 support


patrick-dmxc:
**What language does this apply to?**
C#

**Describe the problem you are trying to solve.**
Hi, there is a Preview awailable for NET8 and so i wold migrate soon to NET8 but as long Protubuf is not capable of NET8 i cant migrate

**Describe the solution you'd like**
Compile as NET8 and publish on nuget

**Describe alternatives you've considered**
none

**Additional context**

jskeet:
Any library targeting .NET 6 (or earlier, or .NET Standard) can already be used from .NET 8. What makes you think it can't? If you're having a specific problem, please provide a minimal repro and details of the error you're receiving.

patrick-dmxc:
Currently the Package is only Targeting Net 5 or Net Standart2.0.
When i use Protobuff in my Net8.0 Applikation, the Compolet Target the the Protobuff-Library to Net Standart 2.0.
In the Dev-Environment this works.
But if i whant to Publish my application as Self-contained Single file or Ready to Run EXE, Net-Standart is to old and this leeds to an non-functional Compiled EXE.

Allso .NET 6 LTS is gone reach end of LTS

jskeet:
No, .NET 6 is still supported unt November 2024.

If you're having problems with a specific publication model, you should give details of that rather than making a blanket statement that protobuf doesn't work with .NET 8. It's unlikely that *just* targeting .NET 7 would fix that, but until we have details of what's not working we can't really test.

patrick-dmxc:
What ar whe talking abaut,
Currently Protobuf is not Targeted to Net6.0 at all
And nobody is hurt if this is addes

jskeet:
> And nobody is hurt if this is addes

It makes the package larger for everyone, and adds additional testing requirements. I'm not going to add a target just for the sake of it, without any evidence that it actually addresses any problems. Please put more details about the actual problem you're facing and we can look into it (although time is limited right now, so it may take a while before we can get to it).

EnemyArea:
This should be updated because its using an deprecated .net version. .net 5 is not longer supported: https://www.nuget.org/packages/Microsoft.NETCore.Targets/ and this goes for some of the dependencies.

jskeet:
@EnemyArea: It's not "using" a deprecated version - it's *supporting* a deprecated version. We do have plans to update the targets at some point, but I don't regard it as hugely urgent.

[...]

PR #15321, opened 2024-01-06
Add Net8.0 support

patrick-dmxc:
#14523 #14790


jskeet:
I don't want to add a target just for the sake of it. You can use the existing library from .NET 8 without any problems. We may well want to target .NET 8 in the future for features that are specific to it, but there's no need to do so yet.

patrick-dmxc:
See #14523

googleberg:
Since this target isn't really needed right now, closing until it is.

Issue #10889, opened 2022-11-02
main branch `//:protobuf` target doesn't provide any headers (glob cross package boundary)


shahms:
**What version of protobuf and what language are you using?**
Version: main
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Debian GNU/Linux rodete

**What runtime / compiler are you using (e.g., python version or gcc version)**
bazel 5.3.2

**What did you do?**
Attempt to depend on `@com_google_protobuf//:protobuf` from another cc_library target.

**What did you expect to see**
A successful build and import of that target.

**What did you see instead?**
```
Error in glob: glob pattern 'src/**/*.h' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
```
Using `--noincompatible_disallow_empty_glob` yields an `#include` error on some of the headers which would've been found by the glob if it didn't cross a package boundary.
Depending on both `@com_google_protobuf//:protobuf` and `@com_google_protobuf//:protobuf_headers` appears to work (but is ugly).

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**
It's particularly obstinate if you've enabled `layering_check` in bazel.

shahms:
Note the work-around with `:protobuf_headers` only works for *some* things, it's still missing headers from at least the visibility-restricted targets:
`@com_google_protobuf//src/google/protobuf/util:json_util`
`@com_google_protobuf//src/google/protobuf/util:type_resolver_util`
`@com_google_protobuf//src/google/protobuf/util:delimited_message_util`
`@com_google_protobuf//src/google/protobuf/util:differencer`

It's also missing a lot of headers from `@com_google_protobuf//src/google/protobuf/io` but that target is public, so easier/better to use directly anyway :-)

acozzette:
Thank you for the bug report. I guess we should probably start listing the headers explicitly since recursive globs are discouraged anyway as I understand.

mkruskal-google:
Hey Shahms,

Bazel doesn't actually enforce that you directly depend on the headers you include (like Blaze does).  So the simplest solution here is just to delete the useless globs and keep everything else the same.  Is there another reason you need a direct dependency on the headers?  I could try to fix protobuf_headers while I'm at it, but if we're not actually fixing a reproducible bug it's almost guaranteed to regress in the future

shahms:
[...]

PR #15332, opened 2024-01-08
Disable unsupported `layering_check` Bazel feature

fmeum:
This allows downstream projects to use `layering_check` without having to patch Protobuf to disable the feature for the repository.

Related to https://github.com/protocolbuffers/protobuf/issues/10889

### DIFF
--- a/REPO.bazel
+++ b/REPO.bazel
@@ -1,0 +1,11 @@
+# This file is read by Bazel 7 and newer, both if Protobuf is the main
+# repository and if it is an external repository.
+repo(
+    features = [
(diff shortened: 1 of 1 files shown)


acozzette:
@fmeum I actually submitted https://github.com/protocolbuffers/protobuf/commit/a7b0421c78412baa48704d727601a17ac0f451d1 last week which makes protobuf compliant with the layering check. Hopefully that removes the need for this change?

fmeum:
It does, thank you @acozzette for resolving this! I think that you can close https://github.com/protocolbuffers/protobuf/issues/10889 as done now.

PR #15358, opened 2024-01-09
25.x 202401092212

sbenzaquen:


mkruskal-google (changes requested):
This looks off, why are there so many code changes?  These are usually just version bumps

zhangskz:
For a patch release, you should be PR'ing the 25.x-202401092212 branch into 25.x (not main). You'll need to create a new PR instead of switching this one over to merge into 25.x (since the tests will be incorrect). Note, you also have merge conflicts that should be due to wrong target branch.

PR #15481, opened 2024-01-18
deps: Bump `rules_pkg` -> 0.10.1

phlax:


BUILD.bazel
```diff
@@ -2,7 +2,7 @@
 
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_proto_library")
 load("@rules_java//java:defs.bzl", "java_lite_proto_library", "java_proto_library")
-load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
+load("@rules_pkg//:pkg:mappings.bzl", "pkg_files", "strip_prefix")
```
acozzette, line 5: Should this be `//pkg:mappings.bzl` instead of `//:pkg:mappings.bzl`?

acozzette (changes requested):
Thank you for the PR, looks good to me except for the comment I made about the path to `mappings.bzl`.

One other thing, I think you pushed an update to the `third_party/abseil-cpp` submodule by mistake.

BUILD.bazel
phlax: yep - bad sed-fu - ill update now

phlax:
@acozzette hopefully that fixes it - i generally avoid submodules so have weak git-fu there

phlax:
nope - trying again ...

phlax:
k - that worked - apologies for force push

acozzette (approved):

phlax:
seems there are more moved files upstream

```console
ERROR: Error computing the main repository mapping: at /workspace/_build/out/external/rules_ruby/ruby/defs.bzl:29:5: at /workspace/_build/out/external/rules_ruby/ruby/private/gemspec/def.bzl:11:5: cannot load '@@rules_pkg//:pkg.bzl': no such file
```

i will look again later today

phlax:
@acozzette `rules_pkg` now has a workaround for the moved bazel targets which allowed us to update - this was my original motivation for raising this PR

for completeness i have reverted all the target changes and updated the dependency here to 0.10.1 (which has the workaround)

Issue #9495, opened 2022-02-11
Improve IDE support for Ruby


acozzette:
The generated Ruby code is not very IDE-friendly, and could be improved to support things like auto-completion. For Python, we're going in the direction of making the main generated code small and efficient but not human-readable, and generating separate `.pyi` files that are more human-readable and IDE-friendly. We need to figure out a plan for Ruby, though.

#7593 suggests one solution, which would be to embed code in `if false` blocks so that it's visible to the IDE and to human readers without changing behavior. @haberman also suggesting doing something like this:
```
Google::Protobuf::DescriptorPool.generated_pool.add_serialized_descriptor(
  "<serialized_descriptor>"
)

# Google::Protobuf::Message defines #inherited to hook in the correct behaviors.
# This fails if #add_serialized_descriptor has not been called for "Foo."
class Foo < Google::Protobuf::Message
  # The base class overrides attr_accessor to do the right thing (maybe even
  # define it as a no-op if #inherited is handling the hard part.
  attr_accessor :foo
end
```

haberman:
I think RBS is a promising direction here: https://blog.appsignal.com/2021/01/27/rbs-the-new-ruby-3-typing-language-in-action.html

We could make the main generated code `foo_pb.rb` nothing but a serialized descriptor. Then we could have a separate `foo_pb.rbs` file that contains full static typing information, for IDE-friendliness.

zhangskz:
This issue is still active per pending PR https://github.com/protocolbuffers/protobuf/pull/15633

PR #15633, opened 2024-01-29
draft: ruby codegen: support generation of rbs files

HoneyryderChuck:
this introduces support for a new protoc option, `--rbs_out`, which points to a directory where ruby type definition files, defined in the RBS format, are stored.

[rbs](https://github.com/ruby/rbs) is the type signature syntax blessed by the ruby core team, used by static analysis tools such as [steep](https://github.com/soutaro/steep), which integrates with VS Code, the `irb` console (for features such as autocompletion). and [typeprof](https://github.com/ruby/typeprof).

It relies on type definitions written into `.rbs` files.

The `protobuf` library already exposes type definitions in [gem_rbs_collection](https://github.com/ruby/gem_rbs_collection/tree/main/gems/google-protobuf/3.22), which is used to source type definitions for libraries which do not want, or can't maintain type definitions themselves.

(`protobuf` could arguably import these into this repository, lmk if you're interested).

This should fix gaps such as better IDE integration, such as the ones described
[here](https://github.com/protocolbuffers/protobuf/issues/9495).

The plan is to do roughly the same type of integration as was done for `.pyi` annotations, which also write to separate files: add the cli option and the rbs generator.

protobuf classes in ruby rely on dynamic attribution of the base class, which makes what I'm trying to achieve a bit difficult. Ideally the type hierarchy could be specified statically in the ruby source code.

```ruby
class Bar < AbstractMessage
class Bar <
Google::Protobuf::DescriptorPool.generated_pool.lookup("Bar").msgclass
```

```

### DIFF
--- a/src/google/protobuf/compiler/ruby/rbs_generator.cc
+++ b/src/google/protobuf/compiler/ruby/rbs_generator.cc
@@ -1,0 +1,43 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+#include "google/protobuf/compiler/ruby/rbs_generator.h"
+
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/log/absl_check.h"
+#include "absl/log/absl_log.h"
+#include "absl/strings/ascii.h"
+#include "absl/strings/match.h"
+#include "absl/strings/str_split.h"
+#include "absl/strings/string_view.h"
+#include "google/protobuf/compiler/code_generator.h"
+#include "google/protobuf/descriptor.h"
+#include "google/protobuf/descriptor.pb.h"
+#include "google/protobuf/io/printer.h"
+#include "google/protobuf/io/zero_copy_stream.h"
+
+namespace google {
+namespace protobuf {
+namespace compiler {
+namespace ruby {
+
+bool RbsGenerator::Generate(const FileDescriptor* file,
+                            const std::string& parameter,
+                            GeneratorContext* context,
+                            std::string* error) const {
+
+  return true;
+}
+
+}  // namespace ruby
+}  // namespace compiler
+}  // namespace protobuf
+}  // namespace google
--- a/src/google/protobuf/compiler/main.cc
+++ b/src/google/protobuf/compiler/main.cc
@@ -16,6 +16,7 @@
 #include "google/protobuf/compiler/python/generator.h"
 #include "google/protobuf/compiler/python/pyi_generator.h"
 #include "google/protobuf/compiler/ruby/ruby_generator.h"
+#include "google/protobuf/compiler/ruby/rbs_generator.h"
 #include "google/protobuf/compiler/rust/generator.h"
 
 // Must be included last.
@@ -83,6 +84,11 @@ int ProtobufMain(int argc, char* argv[]) {
   ruby::Generator rb_generator;
   cli.RegisterGenerator("--ruby_out", "--ruby_opt", &rb_generator,
                         "Generate Ruby source file.");
+
+  // Ruby
+  ruby::RbsGenerator rbs_generator;
+  cli.RegisterGenerator("--rbs_out", &rb_generator,
+                        "Generate Ruby rbs stub.");
 
   // CSharp
   csharp::Generator csharp_generator;
--- a/src/google/protobuf/compiler/ruby/rbs_generator.h
+++ b/src/google/protobuf/compiler/ruby/rbs_generator.h
@@ -1,0 +1,50 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+// Author: jieluo@google.com (Jie Luo)
+//
+// Generates Ruby type definition (.rbs) for a given .proto file.
+
+#ifndef GOOGLE_PROTOBUF_COMPILER_RUBY_RBS_GENERATOR_H__
+#define GOOGLE_PROTOBUF_COMPILER_RUBY_RBS_GENERATOR_H__
+
+#include <string>
+
+#include "google/protobuf/compiler/code_generator.h"
+
+#include "google/protobuf/port_def.inc"
+
+namespace google {
+namespace protobuf {
+namespace compiler {
+namespace ruby {
+
+class PROTOC_EXPORT RbsGenerator : public CodeGenerator {
+ public:
+  RbsGenerator();
+  RbsGenerator(const RbsGenerator&) = delete;
+  RbsGenerator& operator=(const RbsGenerator&) = delete;
+  ~RbsGenerator() override;
+
+  // CodeGenerator methods.
+  uint64_t GetSupportedFeatures() const override {
+    return FEATURE_PROTO3_OPTIONAL;
+  }
+  bool Generate(const FileDescriptor* file, const std::string& parameter,
+                GeneratorContext* generator_context,
+                std::string* error) const override;
+
+};
+
+}  // namespace ruby
+}  // namespace compiler
+}  // namespace protobuf
+}  // namespace google
+
+#include "google/protobuf/port_undef.inc"
+
+#endif  // GOOGLE_PROTOBUF_COMPILER_RUBY_RBS_GENERATOR_H__

### CODE REVIEW

**Comment by HoneyryderChuck** [comment]
@acozzette can I ask for your input on directions here, given your thoughts described in https://github.com/protocolbuffers/protobuf/issues/9495 ?

**Comment by acozzette** [comment]
@HoneyryderChuck This looks like a great direction to me, but I will let @haberman review this since he is much more knowledgeable about Ruby than I am.

**Comment by HoneyryderChuck** [comment]
@haberman any feedback? waiting for comments about the direction, before coming back to this.

**Comment by haberman** [comment]
Sorry for the delay. I think this is a good direction, and I appreciate the small and incremental PR. I think we should move forward with this.

It looks like the CLA check is failing.

**Comment by HoneyryderChuck** [comment]
thx for the feedback @haberman . Did you have a look at https://github.com/protocolbuffers/protobuf/pull/15881 , which may be already the endgame?

**Comment by haberman** [comment]
https://github.com/protocolbuffers/protobuf/pull/15881 has a much greater scope, and I think we'd need to individually evaluate its many parts.

I'm convinced that we want a `.rbs` code generator that has static types for all of the generated APIs. I'm less certain about:

- Adding .rbs files for our public core APIs
- Adding .rbs files for our internal core APIs
- Adding .rbs files for our unit tests
- Adding Steep type checking to our build

**Comment by HoneyryderChuck** [comment]
makes sense 👍 will deal with CLA and move on from there.

**Comment by HoneyryderChuck** [comment]
@haberman went with cherry-picking @qnighy 's commit, and removing the parts not related with sole generation of rbs files on target protos. There was a lot done there that I'd struggle to replicate with the same quality, as I'm still fighting with the build system to generate binaries I can test with locally.

FWIW RBS definitions to protobuf itself are already available [here](https://github.com/ruby/gem_rbs_collection); while it's possible to have the definitions within this repo itself as per the other MR, it's a choice of the maintainers.

**Comment by qnighy** [comment]
Honestly, from the experience working on the PR, I agree that it is probably too early to introduce a typechecker to an existing large project like google-protobuf, which naturally have idioms incompatible with the current type system.

That said, I think it is better to bundle the public type definition here.

The reason: firstly, just similarly to other ecosystems like TypeScript, type definition files are not mere derivative of the runtime code; writing type definitions inherently involves additional design decision. For example, whether to give a type an alias or not, and its name, depends on the author of the type definitions.

Secondly, in google-protobuf, the runtime and the generated code are expected to work in cooperation, and this would also be true in type definitions as well. Therefore, design decision made in the runtime type definition naturally depends on the decisions in the generator.

So, to consistently maintain type definitions, I think it should bundle the one for the runtime as well.

**Comment by haberman** [comment]
Those arguments make sense to me, but we also have to balance the benefits against the maintenance cost. The primary motivation for introducing type signatures for generated code is to make generated interfaces more discoverable and to help IDEs, because the Ruby generated code is currently very opaque, especially since we merged https://github.com/protocolbuffers/protobuf/pull/12462.

The public APIs from the core runtime will also have this issue to some extent, since many interfaces are implemented in C and are probably not discoverable to IDEs. We have Rubydoc embedded in comments, like this example, but that won't help IDEs: https://github.com/protocolbuffers/protobuf/blob/f2d8c2bdd86f34b953c7bae9075daa7648025d3c/ruby/ext/google/protobuf_c/defs.c#L146-L152

But once we're talking about internal-only APIs (like everything in `ffi/`) and especially unit tests, I think we start to get diminishing returns. The cost is having to write everything twice (once in `.rb` and once in `.rbs`), using an unfamiliar language/tool that nobody on the team currently knows, and we don't get any benefits unless we perform type checking in our own project. I think the maintenance burden would outweigh the benefit.

Steep describes itself as "Gradual Typing", which I assume means that type signatures are not required to be complete. Anything that has type signatures is checked, but everything else can continue to work even if it's not typed.

I think the sweet spot would be if we could generate `.rbs` for generated code, and possibly for the *public* interfaces in the core runtime, and add just enough type checking in our CI tests to ensure that those type signatures are correct with respect to the actual code. But we wouldn't bother with typing the internal-only interfaces or unit tests.

How does that sound?

**Comment by HoneyryderChuck** [comment]
@haberman that sounds reasonable, but I don't think we need to do this as part of this PR. FWIW anyone forcibly generating rbs defs for their protobufs will very likely know about rbs-collection already, and get typedefs for core runtime.

There'd still be value in adding typedefs to core runtime; I could see `steep` integration in CI via the added rbs-based suite catching basic bugs, and could be gradually adopted by the core team (or not).

**Comment by haberman** [comment]
I totally agree. If anything, I'm arguing to limit the scope of each individual PR and make things as incremental as possible.

**Comment by HoneyryderChuck** [comment]
@haberman thx for the feedback. Would you mind adding the safe for tests label, so we can continue moving forward?

**Comment by haberman** [comment]
I ran the tests, looks like the build failed.

**Comment by HoneyryderChuck** [comment]
@haberman is it possible for the bot not to remove the "safe for tests" label?

**Comment by haberman** [comment]
> @haberman is it possible for the bot not to remove the "safe for tests" label?

Unfortunately not, the purpose of the tag is to visually verify each commit to ensure that our CI infrastructure isn't being exploited either for resources (eg. bitcoin mining) or to exfiltrate repository secrets.

**Comment by HoneyryderChuck** [comment]
@haberman would you mind having a look? The CI failures seem unrelated to this change (windows bazel build failing on objective C / java, and some security check around workflow files).

**Comment by googleberg** [comment]
@HoneyryderChuck can you rebase and we'll try the tests again?

**Comment by HoneyryderChuck** [comment]
@googleberg done 👍

**Comment by haberman** [comment]
Could we add a disclaimer in the generated `.rbs` files?

```ruby
# This RBS interface is provided for convenience, on a best-effort basis.
# The library is the definitive source for the API contract; if the RBS file
# and the library's behavior differ, the library behavior is authoritative.
# We welcome fixes to change the RBS file to match.
```

That will reduce the risk of accepting this change. We don't want people to take the RBS definition as definitive, and expect the library to change to match.

haberman:
I'm not seeing this warning in the `.rbs` files or in the code generator. Am I missing something?

HoneyryderChuck:
@haberman the change was in the `rbs_generator.cc` file (which shows a "large diff, click to expand" message). I did forget to rerun the generator on the test files, fixed that already. Wondering whether the "generating the stubs" part shouldn't be part of the CI to avoid that.

HoneyryderChuck:
@haberman any updates?

JasonLunn:
Apologies that this PR fell through the cracks and hasn't been updated recently. Could I ask you to rebase it so that I can re-run the latest tests on it, @HoneyryderChuck ?

HoneyryderChuck:
@JasonLunn done.

JasonLunn:
@HoneyryderChuck - There are widespread failures in CI from this change, even in non-Ruby tests, suggesting that the changes that touch protoc have an issue. One example from the test logs:

```
external/protobuf~/src/google/protobuf/compiler/ruby/rbs_generator.cc: In function 'void google::protobuf::compiler::ruby::GenerateEnumValueTypeDefinition(const google::protobuf::EnumValueDescriptor*, google::protobuf::io::Printer*)':
external/protobuf~/src/google/protobuf/compiler/ruby/rbs_generator.cc:332:33: error: conversion from 'absl::lts_20250127::string_view' ***aka 'std::basic_string_view<char>'*** to non-scalar type 'std::string' ***aka 'std::__cxx11::basic_string<char>'*** requested
  332 |   std::string name = value->name();
      |                      ~~~~~~~~~~~^~
```

HoneyryderChuck:
@JasonLunn fix the issues with protoc compilation. FWIW it'd be great if the CI configuration would allow running the CI suite on first approval, that'd help having an overview of the impact of the change without replicating the whole suite locally (which I had some trouble with as well, bazel isn't an easy setup for constrained macOS).

JasonLunn:
> @JasonLunn fix the issues with protoc compilation. FWIW it'd be great if the CI configuration would allow running the CI suite on first approval, that'd help having an overview of the impact of the change without replicating the whole suite locally (which I had some trouble with as well, bazel isn't an easy setup for constrained macOS).

It is a frequent request, but the current configuration is working as intended. A human must apply the `Safe for tests` tag before third party contributions are allowed to run on our CI infrastructure.

JasonLunn:
Looks very close!

In the test logs I see two things:

1. The bazel test target //src/google/protobuf/compiler/ruby:generator_unittest fails from minor text differences, of the form:

```
With diff:
@@ -3,5 +3,5 @@
 # The library is the definitive source for the API contract; if the RBS file
 # and the library's behavior differ, the library behavior is authoritative.
-# We welcome fixes to change the RBS file to match.\\n\"
+# We welcome fixes to change the RBS file to match.
 # source: ruby_generated_pkg_explicit_legacy.proto
```

This failure mode appears to repeat for several tests cases:
```
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] RubyGeneratorTest.Proto3GeneratorTest
[  FAILED  ] RubyGeneratorTest.Proto2GeneratorTest
[  FAILED  ] RubyGeneratorTest.Proto3ImplicitPackageTest
[  FAILED  ] RubyGeneratorTest.Proto3ExplictPackageTest
[  FAILED  ] RubyGeneratorTest.Proto3ExplictLegacyPackageTest
```

2. Lots of warning messages of the form ` ruby_generator.cc:172] ruby_package option should be in the form of: 'A::B::C' and not 'A.B.C'`

JasonLunn:
`//src:cmake_lists_staleness_test` is still failing...

JasonLunn:
Do these links work for you?

https://github.com/protocolbuffers/protobuf/actions/runs/14174126025/job/39704847071#step:4:4075
https://github.com/protocolbuffers/protobuf/actions/runs/14174126025/job/39704787099#step:4:1242

HoneyryderChuck:
they do, thank you 🙏

src/google/protobuf/compiler/ruby/rbs_generator.cc
```diff
@@ -0,0 +1,806 @@
...
+  bool need_change_to_module = true;
+  std::string package_name;
+  if (file->options().has_ruby_package()) {
+    package_name = file->options().ruby_package();
+
+    if (package_name.find("::") != std::string::npos) {
+      need_change_to_module = false;
+    }
+  } else {
+    package_name = file->package();
```
JasonLunn, line 91: The line is getting flagged with an error on some builds and not others. Here's one from a [failing GCC-based build](https://github.com/protocolbuffers/protobuf/actions/runs/14271858302/job/40006682513?pr=15633#step:4:4025).

Elsewhere, I see similar lines constructed as:
`package_name = std::string(file->package())`

HoneyryderChuck: @JasonLunn done.

HoneyryderChuck:
@JasonLunn I see that the tests are still failing, but it seems unrelated, right? Is this something that may be fixed upstream and just requires rebasing?

JasonLunn:
> @JasonLunn I see that the tests are still failing, but it seems unrelated, right? Is this something that may be fixed upstream and just requires rebasing?

Yes, I tried sending you https://github.com/HoneyryderChuck/protobuf/pull/1 to do just that but a rebase of any flavor should do the trick.

HoneyryderChuck:
@JasonLunn rebased

HoneyryderChuck:
@JasonLunn any updates? it seems to pass the tests this time.

JasonLunn:
Working on this internally to get it working with our infra.

googleberg:
@HoneyryderChuck it looks like this needs another rebase so we can kick the tests again.

HoneyryderChuck:
@googleberg done

zhangskz:
Manually imported this since this change touches our workflow files, to allow new test workflow to run. Please do bump this PR if not submitted, or updated in the next few days.

zhangskz:
Note this change required some changes on our side to handle merge conflicts / transforms / linters when importing.

Workflow changes in `test_ruby.yml` currently fail with:
```
ERROR: Skipping 'ruby': no such target '//ruby:ruby': target 'ruby' not declared in package 'ruby' defined by /workspace/ruby/BUILD.bazel
ERROR: no such target '//ruby:ruby': target 'ruby' not declared in package 'ruby' defined by /workspace/ruby/BUILD.bazel
```
https://github.com/protocolbuffers/protobuf/actions/runs/18666531950/job/53218943443?pr=24064

You'll likely want to at least use a more up-to-date Bazel 7 or 8 image e.g. `us-docker.pkg.dev/protobuf-build/containers/test/linux/ruby:ruby-8.0.1-3.4-b77fdae6d4771789dfc66a56bf8d806354e8011a` since Bazel 6 is out of support which might explain the current error.

Also noticing that the `typecheck.gemfile.lock` being copied in this image doesn't seem to exist, and `bundle` isn't actually available in this image (neither is `gem install bundler`). We use `bundle` in a few other GHA tests in this file but using other images. I'm not as familiar with ruby type checking and this test, but looks like the setup for typecheck action needs some work to include the right dependencies.

qnighy:
I hope it lands some day. Thank you for your efforts.

HoneyryderChuck:
@qnighy thx, I hope so too. Something for @JasonLunn to answer. Thx for the initial work.

.github/workflows/test_ruby.yml
```diff
@@ -210,3 +210,23 @@ jobs:
...
+      - name: Run type checker
+        uses: protocolbuffers/protobuf-ci/bazel-docker@v2
+        with:
+          image: us-docker.pkg.dev/protobuf-build/containers/test/linux/ruby:ruby-3.3.0-6.3.0-904cad5249547845454998ca3837a34c71fabf96
+          credentials: ${{ secrets.GAR_SERVICE_ACCOUNT }}
+          bazel-cache: ruby_typecheck
+          bash:
+            bazel build //:protoc
+            cd ruby
+            cp typecheck.gemfile.lock Gemfile.lock
```
JasonLunn, line 231: Where does `typecheck.gemfile.lock` come from?

JasonLunn:
I took a swing at getting this landed back in August, but got derailed and then had to focus on other projects that have since been completed.

There is an internal process behind-the-scenes that has to be performed for any change that touches `protoc`, and that's why this has been stuck in the interim.

The only challenge I see right now is the new type checking test, which appears to require files that don't appear in the PR.

.github/workflows/test_ruby.yml
HoneyryderChuck: this change was imported 2y ago from [the original PR](https://github.com/protocolbuffers/protobuf/pull/15881/files#diff-bdc24a46e5d0ab448eb849a76773a9546bbd896317b6eb1882d162b380110a97) this was based on. Since then, it's been iterated to the point where the whole directory disappeared, and this step is still around. Could you delete it? 🙏

HoneyryderChuck:
@JasonLunn appreciate all the effort that you have put in helping move forward with this. This would probably have not been so close to be merged without your help in the last few months.

From my part, it's been 2y since this PR has been created (and it was itself based on a prior PR which had been lingering for a while). A lot changed since then, and I no longer work at the company that would have benefited from this, and with it, I have lost access to the setup that allowed me to troubleshoot this locally. It was a multi-day time-consuming effort that I no longer have the will to go through again (I only ever touched bazel for this project, and beyond the learning curve that I never quite got over, it was quite dev-unfriendly in a Mac).

Moreover, feedback has also been sparse through the whole timeline, and waiting multiple days for someone to manually unlock CI after every change only added to the frustration. I'm sure there were justifiable reasons for this due to the internal processes google enforces, so not looking at blaming individuals here, just that, all of this made my contribution way harder than it could have been. Hope this feedback helps you create internal processes more friendly to outside contributions like this one.

.github/workflows/test_ruby.yml
JasonLunn: Will do

JasonLunn:
@HoneyryderChuck this has been merged, and should be included in the next RC candidate. Thank you for your contributions, your patience, and your feedback. We will strive to improve!

PR #15677, opened 2024-02-01
reserve memory space for packed repeated field to avoid unnecessary memory overhead

firedtoad:
avoid bunch memory waste when parse packed repeated field


fowles:
Thanks for the contribution.  Review of this is blocked on acceptance of the Contributor License Agreement.

firedtoad:
signed

src/google/protobuf/parse_context.h
```diff
@@ -1196,7 +1196,14 @@ template <typename Add, typename SizeCb>
...
   int size = ReadSize(&ptr);
-  size_callback(size);
+  auto element_size = 0;
+  const uint8_t* pData = (uint8_t*)ptr;
+  for (auto i = 0; i < size; i++) {
+    if (pData[i] < 0x80) {
+      element_size++;
+    }
+  }
+  size_callback(element_size);
```
fowles, line 1206: unfortunately, the type of size_callback is specified as bytes and we depend on this fact elsewhere.  

https://github.com/protocolbuffers/protobuf/blob/461b50b18097122f8a5ab3c3686f8254827a1716/src/google/protobuf/generated_message_tctable_lite.cc#L1344-L1366

src/google/protobuf/parse_context.h
```diff
@@ -1196,7 +1196,14 @@ template <typename Add, typename SizeCb>
 const char* EpsCopyInputStream::ReadPackedVarint(const char* ptr, Add add,
                                                  SizeCb size_callback) {
   int size = ReadSize(&ptr);
-  size_callback(size);
+  auto element_size = 0;
+  const uint8_t* pData = (uint8_t*)ptr;
+  for (auto i = 0; i < size; i++) {
```
fowles, line 1201: this loop is not safe, data corruption (or a truncated input buffer) can cause size to be much larger than pData.

src/google/protobuf/parse_context.cc
```diff
@@ -437,6 +437,8 @@ const char* VarintParser(void* object, const char* ptr, ParseContext* ctx) {
       val = varint;
     }
     static_cast<RepeatedField<T>*>(object)->Add(val);
+  },[object](int size){
+    static_cast<RepeatedField<T>*>(object)->Reserve(size);
```
fowles, line 441: instead of changing the callback, you should add some heuristic here for getting an estimate of reserve size from size.

firedtoad: I knew this, But gogoproto use this method to calculate the real element count,and widely used for many years
https://github.com/gogo/protobuf/blob/master/test/casttype/combos/both/casttype.pb.go#L1933.
If loop to the size end is unsafe, ReadPackedVarintArray is also unsafe https://github.com/protocolbuffers/protobuf/blob/461b50b18097122f8a5ab3c3686f8254827a1716/src/google/protobuf/parse_context.h#L1185

firedtoad: I've use this modify version in release produdct and have not found crital issue yet.

fowles: At minimum the existing usage would need to be updated, but this also changes the loop to scan across the data twice instead of once.  The barrier for accepting that is much higher and you can get most of the impact by mimicking the heuristic used where it is already called.

`ReadPackedVarintArray` is safe because its loop is expressed in terms of the end pointer and it validates the varint at every step returning `nullptr` on failure and breaking out of the loop.  This loop on the other hand iterates [0, size from the byte steam) and dereferences `ptr[idx]` without ever checking if that is inside the loop bounds.

firedtoad: Do you mean something like this:
`
  const uint8_t* pData = (uint8_t*)ptr;
  while (pData && pData < (pData + size)) {
    if (*pData < 0x80) {
      element_size++;
    }
  }
`
And why not write the real element size for packed repeated when serialize? Is is the history tech debt?

fowles: `EpsCopyInputStream` is not guaranteed to have a contiguous buffer, so the way to safely iterate it is [substantially more complicated](https://github.com/protocolbuffers/protobuf/blob/d67f921f90e3aa03f65c3e1e507ca7017c8327a6/src/google/protobuf/parse_context.h#L1202-L1228).

Additionally, the cost of scanning the data twice vs scanning once is a complex trade off that I don't have time to evaluate for all of Google's production fleet right now.

As such, we won't accept this patch in its current form.

If you would like to do something like the [existing heuristic](https://github.com/protocolbuffers/protobuf/blob/461b50b18097122f8a5ab3c3686f8254827a1716/src/google/protobuf/generated_message_tctable_lite.cc#L1344-L1366) in the reserve call, we would be willing to accept a patch for that.


As to why does it write the byte size and not the element size in the serialized form.  I am not sure, that choice predates me.  Most likely it was done so that you could easily skip all the data when you are scanning the bytes and are not interested in the field.  Otherwise you would have to parse every varint to find the end point.

Issue #12810, opened 2023-05-15
protobuf-bom does not include protobuf-kotlin


gavvvr:
**What did you do?**

I create a `java-platform` module in Gradle and consume `protobuf-bom` to specify version of protobuf-java stack only once:

```kotlin
plugins {
    `java-platform`
}

javaPlatform {
    allowDependencies()
}

dependencies {
    api(platform("com.google.protobuf:protobuf-bom:3.23.0"))
}
```

Then in my application's Gradle build I consume my `java-platform` module and declare dependency on `protobuf-kotlin` without version (hoping that the version will come from BOM):
```kotlin
dependencies {
    implementation(platform("com.my-project:my-gradle-platform:0.0.1-SNAPSHOT"))
    implementation("com.google.protobuf:protobuf-kotlin")
}
```

**What did you expect to see**

Working build

**What did you see instead?**

```
Could not determine the dependencies of task ':my-app:compileJava'.
> Could not resolve all task dependencies for configuration ':my-app:compileClasspath'.
   > Could not find com.google.protobuf:protobuf-kotlin:.
```

PR #15775, opened 2024-02-08
Add protobuf-kotlin-lite into protobuf-bom

NearTox:


zhangskz:
This does seem like it should've been included here -- do you have more context on what this fixes / what you're hitting right now?

NearTox:
Hello, this change resolves the missing protobuf-kotlin-lite version definition in the protobuf bom for java-platform, which is similar to the one described in #12810

zhangskz (approved):

PR #15738, opened 2024-02-06
Set exit status to 1 when fatal_warnings flag is on

junekhan:
This fixes #10486

PR #15792, opened 2024-02-10
Set exit status to 1 when finding warnings with fatal_warnings being on

### LINKED ISSUE #10486: protoc cli returns success with --fatal_warnings switch when warnings are detected
Version: libprotoc 3.19.4
Language: NA
OS: OSX, installed via brew

1. open a terminal (I use zsh)
2. decode a protobuf file for which a warning will be emitted using the `--fatal_warnings` switch
3. check the command return value

```
$ cat temp.in.pbuf | protoc  --decode job_definition.job_t --fatal_warnings --proto_path ~/workspace/engine/proto/ job.proto > /dev/null
warning:  Input message is missing required fields:  country[0].id, country[1].id
$ echo $?
0
```

notice the contract include 
```
message country_t {
	required int32 id = 1;
	required string country = 2;
}
```

and the encode file does not indeed have an `id` field, i.e. the warning is correct.

protoc shall return a non-zero error

**What did you see instead?**
protoc returns 0

junekhan:
This is a resubmission of [PR#15738](https://github.com/protocolbuffers/protobuf/pull/15738) and fixes #10486.

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -1398,11 +1398,11 @@ int CommandLineInterface::Run(int argc, const char* const argv[]) {
       file.add_message_type()->set_name("EmptyMessage");
       ABSL_CHECK(pool.BuildFile(file) != nullptr);
       codec_type_ = "EmptyMessage";
-      if (!EncodeOrDecode(&pool)) {
+      if (!EncodeOrDecode(&pool, error_collector.get())) {
         return 1;
       }
     } else {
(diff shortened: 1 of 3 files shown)


fowles (approved):

fowles:
Looks like the unit tests need updating:

```
src/google/protobuf/compiler/command_line_interface_unittest.cc:4032: Failure
  StripCR(expected_text)
    Which is: "warning:  Input message is missing required fields:  a, b, c\n"
  StripCR(captured_stderr_)
    Which is: "input: warning: Input message is missing required fields:  a, b, c\n"
```

fowles (approved):

PR #15794, opened 2024-02-11
Add [ThreadStatic] StringBuilder in JsonTextTokenizer

TrayanZapryanov:
While profiling allocations in our code, I've found a lot of StringBuilder instances.
image

Changes add similar functionality to the dotnet StringBuilderCache class in  JsonTextTokenizer.
They are not modifying existing logic and only will reduce relatively small text/number allocations by using cached instance.

### DIFF
--- a/csharp/src/Google.Protobuf/JsonTokenizer.cs
+++ b/csharp/src/Google.Protobuf/JsonTokenizer.cs
@@ -772,7 +772,7 @@ namespace Google.Protobuf
                 {
                     if (sb.Capacity <= MaxCachedStringBuilderSize)
                     {
-                        cachedInstance = sb;
+                        cachedInstance = cachedInstance?.Capacity >= sb.Capacity ? cachedInstance : sb;
                     }
                 }
 
--- a/csharp/src/Google.Protobuf/JsonTokenizer.cs
+++ b/csharp/src/Google.Protobuf/JsonTokenizer.cs
@@ -732,7 +732,10 @@ namespace Google.Protobuf
                 }
             }
 
-            /// <summary>Provide a cached reusable instance of stringbuilder per thread.</summary>
+            /// <summary>
+            /// Provide a cached reusable instance of stringbuilder per thread.
+            /// Copied from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/StringBuilderCache.cs
+            /// </summary>
             private static class StringBuilderCache
             {
                 private const int MaxCachedStringBuilderSize = 360;
--- a/csharp/src/Google.Protobuf/JsonTokenizer.cs
+++ b/csharp/src/Google.Protobuf/JsonTokenizer.cs
@@ -177,12 +177,6 @@ namespace Google.Protobuf
             // The set of states in which a value is valid next token.
             private static readonly State ValueStates = State.ArrayStart | State.ArrayAfterComma | State.ObjectAfterColon | State.StartOfDocument;
 
-            private const int MAX_BUILDER_SIZE = 360;
-            private const int DEFAULT_CAPACITY = 16;
-
-            [ThreadStatic]
-            private static StringBuilder? t_cachedInstance;
-
             private readonly Stack<ContainerType> containerStack = new Stack<ContainerType>();
             private readonly PushBackReader reader;
             private State state;
@@ -307,7 +301,8 @@ namespace Google.Protobuf
             /// </summary>
             private string ReadString()
             {
-                var builder = AcquireStringBuilder();
+                //builder will not be released in case of an exception, but this is not a problem and we will create new on next Acquire
+                var builder = StringBuilderCache.Acquire();
                 bool haveHighSurrogate = false;
                 while (true)
                 {
@@ -322,7 +317,7 @@ namespace Google.Protobuf
                         {
                             throw reader.CreateException("Invalid use of surrogate pair code units");
                         }
-                        return GetStringAndRelease(builder);
+                        return StringBuilderCache.GetStringAndRelease(builder);
                     }
                     if (c == '\\')
                     {
@@ -414,7 +409,8 @@ namespace Google.Protobuf
 
             private double ReadNumber(char initialCharacter)
             {
-                var builder = AcquireStringBuilder();
+                //builder will not be released in case of an exception, but this is not a problem and we will create new on next Acquire
+                var builder = StringBuilderCache.Acquire();
                 if (initialCharacter == '-')
                 {
                     builder.Append("-");
@@ -443,7 +439,7 @@ namespace Google.Protobuf
                 }
 
                 // TODO: What exception should we throw if the value can't be represented as a double?
-                var builderValue = GetStringAndRelease(builder);
+                var builderValue = StringBuilderCache.GetStringAndRelease(builder);
                 try
                 {
                     double result = double.Parse(builderValue,
@@ -570,38 +566,6 @@ namespace Google.Protobuf
                     ContainerType.Document => State.ExpectedEndOfDocument,
                     _ => throw new InvalidOperationException("Unexpected container type: " + parent),
                 };
-            }
-
-            private static StringBuilder AcquireStringBuilder(int capacity = DEFAULT_CAPACITY)
-            {
-                if (capacity <= MAX_BUILDER_SIZE)
-                {
-                    var sb = t_cachedInstance;
-                    // Avoid stringbuilder block fragmentation by getting a new StringBuilder
-                    // when the requested size is larger than the current capacity
-                    if (capacity <= sb?.Capacity)
-                    {
-                        t_cachedInstance = null;
-                        sb.Clear();
-                        return sb;
-                    }
-                }
-                return new StringBuilder(capacity);
-            }
-
-            private static void Release(StringBuilder sb)
-            {
-                if (sb.Capacity <= MAX_BUILDER_SIZE)
-                {
-                    t_cachedInstance = sb;
-                }
-            }
-
-            private static string GetStringAndRelease(StringBuilder sb)
-            {
-                var result = sb.ToString();
-                Release(sb);
-                return result;
             }
 
             private enum ContainerType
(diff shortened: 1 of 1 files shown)


TrayanZapryanov:
CLA signed

TrayanZapryanov:
Is it okay if I push another PR for changing JsonToken to readonly struct?
It is also quite high in memory allocations and is prepared to be immutable.

jskeet:
> Is it okay if I push another PR for changing JsonToken to readonly struct?
> It is also quite high in memory allocations and is prepared to be immutable.

Certainly happy to see a PR - the fact that it's internal is helpful here.

(I'll try to review *this* PR some time next week. Looking *very* briefly though, I'd encourage you to consider ThreadLocal vs ThreadStatic; I was under the impression that these days ThreadLocal was generally preferred, but I could be wrong.)

TrayanZapryanov:
@jskeet I've mimic code from here [StringBuilderCache](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/StringBuilderCache.cs) and it is using [ThreadStatic]. If you prefer I can try with ThreadLocal, but this code is "battle tested" in runtime and also in our software :).

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -177,6 +177,12 @@ private sealed class JsonTextTokenizer : JsonTokenizer
             // The set of states in which a value is valid next token.
             private static readonly State ValueStates = State.ArrayStart | State.ArrayAfterComma | State.ObjectAfterColon | State.StartOfDocument;
 
+            private const int MAX_BUILDER_SIZE = 360;
```
jskeet, line 180: Please rename to MaxBuilderSize and DefaultCapacity.

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -177,6 +177,12 @@ private sealed class JsonTextTokenizer : JsonTokenizer
...
+            private const int DEFAULT_CAPACITY = 16;
+
+            [ThreadStatic]
+            private static StringBuilder? t_cachedInstance;
```
jskeet, line 184: Please rename to just `cachedInstance` - we don't have any `t_` prefixes anywhere as far as I'm aware.

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -565,6 +572,38 @@ private void PopContainer()
                 };
             }
 
+            private static StringBuilder AcquireStringBuilder(int capacity = DEFAULT_CAPACITY)
```
jskeet, line 575: Remove the optional parameter as we're not using it?

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -565,6 +572,38 @@ private void PopContainer()
...
+                    {
+                        t_cachedInstance = null;
+                        sb.Clear();
+                        return sb;
+                    }
+                }
+                return new StringBuilder(capacity);
+            }
+
+            private static void Release(StringBuilder sb)
```
jskeet, line 592: I don't think we need this as a separate method, given that we only ever call it from GetStringAndRelease.

csharp/src/Google.Protobuf/JsonTokenizer.cs
jskeet: Or better, MaxCachedStringBuilderSize and DefaultStringBuilderCapacity.

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -301,7 +307,7 @@ private void ValidateState(State validStates, string errorPrefix)
             /// </summary>
             private string ReadString()
             {
-                var value = new StringBuilder();
+                var builder = AcquireStringBuilder();
```
jskeet, line 310: We don't release this if we throw an exception. Presumably that's okay, it just means we won't reuse it - and given that an exception's being thrown, that's the least of our concerns. Maybe add a comment to that effect?

csharp/src/Google.Protobuf/JsonTokenizer.cs
jskeet: Or better, `cachedStringBuilder` (unless we introduce a new class)

jskeet (commented):
Another alternative to consider would be to have a private nested StringBuilderCache class instead of adding these methods directly to the JsonTextTokenizer - that we wouldn't need to add "StringBuilder" into all the names.

TrayanZapryanov:
> Another alternative to consider would be to have a private nested StringBuilderCache class instead of adding these methods directly to the JsonTextTokenizer - that we wouldn't need to add "StringBuilder" into all the names.

In general I also prefer this to be extracted as a general class and used in on every place with new StringBuilder().
If you like I can expand changes in this direction and apply all the other comments.

jskeet:
> In general I also prefer this to be extracted as a general class and used in on every place with new StringBuilder().
> If you like I can expand changes in this direction and apply all the other comments.

I think it's worth extracting as a general class, but I'd *start* with it in JsonTextTokenizer rather than applying it to *all* uses of StringBuilder in the project. We can easily expand its usage further later.

csharp/src/Google.Protobuf/JsonTokenizer.cs
TrayanZapryanov: As we have private class with chance to make it internal - I would like to stay with it, but if you prefer I will submit new commit

TrayanZapryanov: I see usages in project where it can be useful : [here](https://github.com/protocolbuffers/protobuf/blob/73769a030d127107be84e052b4afa3cbc62e77ee/csharp/src/Google.Protobuf/JsonParser.cs#L946C30-L946C31)

jskeet:
Right, I'll fetch and try to do some benchmarking when I can. It may not be this week though.

jskeet:
Benchmark results:

Before optimizations:

| Method                    | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|-------------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| TestAllTypesProto3_Medium | 84.52 us | 1.639 us | 2.013 us | 8.0566 | 0.3662 |  49.39 KB |

After stringbuilder:

| Method                    | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|-------------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| TestAllTypesProto3_Medium | 78.62 us | 1.509 us | 2.066 us | 5.3711 | 0.2441 |  33.14 KB |

So it looks like this is probably worth doing (and the benefit will very much depend on the JSON being parsed).
The current code introduces new warnings though, which should be fixed.

TrayanZapryanov:
@jskeet Where do you see "new warnings"?
As I cannot compile it - it is hard for me to see if compiler warns.
Also tests are somehow paused :
image

Maybe there is somewhere a check that should be set in order to be executed....(Check for Safety)

jskeet:
I've triggered the tests.

jskeet (approved):
I'd still like @JamesNK to have a look at this before merging, in case he has thoughts. But otherwise, this looks fine to me.

TrayanZapryanov:
@JamesNK I've added note from where I've copied code.

Here it is a result with allocations once I've build this PS + JsonToken structure of the same benchmarked app.
image

In general it is 120MB less

JamesNK:
I mean the steady state memory usage of the app will be higher. Cached StringBuilders stick around.

TrayanZapryanov:
@jskeet Do you need something more from my side? Or this is enough to be merged?

jskeet:
> @jskeet Do you need something more from my side? Or this is enough to be merged?

Just time to get round to final checks and things. Sorry - I'm really busy and Protobuf is only one small part of my work, so it's hard to get round to things.

csharp/src/Google.Protobuf/JsonTokenizer.cs
```diff
@@ -728,6 +731,59 @@ internal InvalidJsonException CreateException(string message)
...
+
+                    return new StringBuilder(capacity);
+                }
+
+                /// <summary>Place the specified builder in the cache if it is not too big.</summary>
+                private static void Release(StringBuilder sb)
+                {
+                    if (sb.Capacity <= MaxCachedStringBuilderSize)
+                    {
+                        cachedInstance = sb;
```
amanda-tarafa, line 775: Should we keep the bigger of the two here? There's a chance the cached one has bigger capacity than the one being released and we might want to keep the bigger one to reduce the chances of line 758 evaluating to true?

TrayanZapryanov: The idea is to not keep too big StringBuilder instances as they are static per thread and can increase total memory consumption of the app. At least this are considerations that people in .net runtime had. As written above the class - this is a copy of internal class used there.

amanda-tarafa: Yes, I'm not saying we should keep anything that's above `MaxCachedStringBuilderSize`. I'm saying we should do something like the following code, to make sure that we keep cache the biggest string builder possible of the ones we have already allocated, within the `MaxCachedStringBuilderSize` boundaries.

```
cachedInstance = cachedInstance?.Capacity >= sb.Capacity ? cachedInstance : sb;
```

TrayanZapryanov: If you check the code in Acquire - it clears cachedInstance and set it to null in happy path.
So when releasing **cachedInstance** will be null and there is no need to check it's capacity.

amanda-tarafa: I'm refering to this scenario:

- Acquiere sb of 50 capacity
- Release sb of 50 capacity => sb of 50 is now cached
- Acquire sb of 100 capacity => new sb of 100 is created, sb of 50 continues to be cached
- Release sb of 100 capacity => sb of 100 is discarded, even though it's better to discard the 50 capacity and keep the 100 capacity
- Acquire sb of 70 capacity => new sb of 70 capacity is created, sb of 50 continues to be cached, we could have reused the sb of 100 capacity

It's probably easy to write a test that demontrates this, unless I'm missunderstanding something. This seems like a simple change that would maximize the benefits of caching.

amanda-tarafa (approved):
Thanks for the changes!

TrayanZapryanov:
Same benchmark as in [here](https://github.com/protocolbuffers/protobuf/pull/15795)

main:
// * Summary *

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3235/23H2/2023Update/SunValley3)
11th Gen Intel Core i9-11900K 3.50GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.200
  [Host]     : .NET 8.0.2 (8.0.224.6711), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.2 (8.0.224.6711), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI

Job=DefaultJob  Error=3.84 ns  StdDev=3.21 ns

| Method | Mean         | Allocated |
|------- |-------------:|----------:|
| Small  |     281.1 ns |   1.01 KB |
| Big    | 256,221.4 ns |  642.7 KB |

PR:
Stringbuilder cache

| Method | Mean         | Allocated |
|------- |-------------:|----------:|
| Small  |     292.6 ns |     512 B |
| Big    | 231,857.0 ns |  260792 B |

jskeet:
@mkruskal-google: This has been reviewed from a C# standpoint, and I've just marked it safe to run tests again. If you could review it to kick off the merge process, that'd be great.

jskeet:
Implemented in https://github.com/protocolbuffers/protobuf/commit/fac929d9aa607ccdfd078344ac12cee7d2b3fdda

PR #15817, opened 2024-02-13
Implement service & method descriptor lookup in Ruby

KJTsanaktsidis:
This PR implements lookup of service descriptor and method descriptor objects in Ruby as described in issue https://github.com/protocolbuffers/protobuf/issues/14891. 

It contains three implementations - one for the CRuby extension API, one for JRuby, and one for FFI.

With this patch,

* `DescriptorPool#lookup('fully.qualified.service.name')` works and returns a `Google::Protobuf::ServiceDescriptor` object
* You can call `#options` on that to get the service options
* You can call `#methods` on that to get the services' methods as `Google::Protobuf::MethodDescriptor` objects,
* You can call `MethodDescriptor#options` to get method options
* You can also get the streaming flags & input/output types of the method with `#input_type`, `#output_type`, `#client_streaming`, and `#server_streaming`.

In order to make the FFI implementation work, I had to mark some more methods in the UPB header as exported - I guess that's something which will have to be done on the UPB side, like this https://github.com/protocolbuffers/upb/commit/01fed1cc1ba255bf22b49393ba054b8d270e6ba3

CC @dazuma & @haberman from the original issue, and @JasonLunn (since you work on protobuf it seems - small world!)

I apologies for the large volume of copy-pasta'd code from the existing descriptor class implementations into the new ones - I felt this was probably better than designing new abstractions to reduce it off the bat though; this feels like it "fits in" with the existing implementation.


ruby/tests/service_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    refute @test_service.options.deprecated
+  end
+
+  def test_service_options_extensions
+    # The ruby API doesn't actually support extensions it seems yet, but we can bodge it for this
+    # test by abusing to_json (but only for the UPB implementation of to_json it seems - the java
+    # one doesn't do this).
+    omit('JRuby implementation cannot display option extensions') if defined?(::JRUBY_VERSION)
+
+    options_hash = JSON.parse(@test_service.options.to_json)
```
KJTsanaktsidis, line 41: So the actual utility of this for the original use-case in https://github.com/protocolbuffers/protobuf/issues/14891 (getting RPC options) might be somewhat dubious without support for being able to get message extensions (without resorting to `#to_json`). I guess that can be tackled in a subsequent PR

haberman:
This sounds great; I haven't taken a look at the code yet, but from the description this sounds like just the right approach.

> that's something which will have to be done on the UPB side

The upb code is [now in this repo](https://github.com/protocolbuffers/protobuf/tree/main/upb), and the old repo is defunct.

We should probably make the banner at the top of [the upb readme](https://github.com/protocolbuffers/upb/blob/main/README.md) bigger and more emphatic about this point.

ruby/tests/service_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    puts "JSON: #{@test_service.options.to_json}"
```
JasonLunn, line 42: Can this output be removed?

ruby/src/main/java/com/google/protobuf/jruby/RubyServiceDescriptor.java
```diff
@@ -0,0 +1,168 @@
+/*
+ * Protocol Buffers - Google's data interchange format
+ * Copyright 2014 Google Inc.  All rights reserved.
```
JasonLunn, line 3: Nit: 2014 -> 2024.

ruby/src/main/java/com/google/protobuf/jruby/RubyMethodDescriptor.java
```diff
@@ -0,0 +1,156 @@
+/*
+ * Protocol Buffers - Google's data interchange format
+ * Copyright 2014 Google Inc.  All rights reserved.
```
JasonLunn, line 3: Nit: 2014 -> 2024

ruby/lib/google/protobuf/ffi/service_descriptor.rb
```diff
@@ -0,0 +1,110 @@
+# Protocol Buffers - Google's data interchange format
+# Copyright 2022 Google Inc.  All rights reserved.
```
JasonLunn, line 2: Nit: 2022 -> 2024

ruby/lib/google/protobuf/ffi/method_descriptor.rb
```diff
@@ -0,0 +1,114 @@
+# Protocol Buffers - Google's data interchange format
+# Copyright 2022 Google Inc.  All rights reserved.
```
JasonLunn, line 2: Nit: 2022 -> 2024

JasonLunn (changes requested):

JasonLunn:
The FFI tests are currently failing with errors like:

`Caught exception `Function 'upb_MethodDef_Service' not found in [/usr/local/rvm/gems/ruby-2.7.0/gems/google-protobuf-4.27.0/ext/google/protobuf_c/x86_64-linux/libprotobuf_c_ffi.so]` while loading FFI implementation of google/protobuf.
Falling back to native implementation.`

Similar changes as you've already made in `ruby-upb.h` to apply the `UPB_API` macro may also needed in `ruby-upb.c`.

haberman:
> Similar changes as you've already made in ruby-upb.h to apply the UPB_API macro may also needed in ruby-upb.c.

I think they should be made in the `upb` tree instead. Those changes should then automatically propagate to `ruby-upb.c`.

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -1294,6 +1302,308 @@ static void EnumDescriptor_register(VALUE module) {
...
+
+static void ServiceDescriptor_register(VALUE module) {
+  VALUE klass = rb_define_class_under(module, "ServiceDescriptor", rb_cObject);
+  rb_define_alloc_func(klass, ServiceDescriptor_alloc);
+  rb_define_method(klass, "initialize", ServiceDescriptor_initialize, 3);
+  rb_define_method(klass, "name", ServiceDescriptor_name, 0);
+  rb_define_method(klass, "each_method", ServiceDescriptor_each_method, 0);
+  rb_define_method(klass, "file_descriptor", ServiceDescriptor_file_descriptor, 0);
+  // n.b. this shadows Object#methods, which seems less than ideal?
+  rb_define_method(klass, "methods", ServiceDescriptor_methods, 0);
```
KJTsanaktsidis, line 1447: I do want to point out that this shadows `Object#methods`, which is a bit icky. Perhaps we should rename this `#rpc_methods` or `#service_methods` or `#method_descriptors` or some such?

ruby/tests/service_test.rb
KJTsanaktsidis: Sorry, forgot about this and didn't spot it when I was committing.

ruby/src/main/java/com/google/protobuf/jruby/RubyServiceDescriptor.java
KJTsanaktsidis: I wasn't actually sure if I was supposed to change those - I've gone and done that now.

KJTsanaktsidis:
> I think they should be made in the upb tree instead. Those changes should then automatically propagate to ruby-upb.c.

How does this happen? I made the changes in the `upb/` tree and regenerated the almagamation with `bazel build //upb:gen_ruby_amalgamation`; however, that just updates the amalgamation in the `bazel-bin/` build directory:

```
% bazel build //upb:gen_ruby_amalgamation
INFO: Analyzed target //upb:gen_ruby_amalgamation (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //upb:gen_ruby_amalgamation up-to-date:
  bazel-bin/upb/ruby-upb.c
  bazel-bin/upb/ruby-upb.h
INFO: Elapsed time: 0.288s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

For now I regenerated the amalgamation, then copied it into `ruby/ext` and checked it in to this PR, but if there's some automatic process which is supposed to address that I can drop https://github.com/protocolbuffers/protobuf/pull/15817/commits/c93913c140f76cb8cb51afa4c3c44efa243ee716

JasonLunn (approved):

haberman:
> but if there's some automatic process which is supposed to address that I can drop c93913c

Yes, there is an automated process that will regenerate the files; for example see: https://github.com/protocolbuffers/protobuf/commit/a3c33a87c1f24a7aa082317257a9ee297b4d883d

haberman (changes requested):

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -1294,6 +1302,308 @@ static void EnumDescriptor_register(VALUE module) {
...
+    rb_yield(obj);
+  }
+  return Qnil;
+}
+
+/*
+ * call-seq:
+ *    ServiceDescriptor.methods => Enumerator
+ *
+ * Returns an enumerator over all the methods in this service
```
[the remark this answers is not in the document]
KJTsanaktsidis: > If we want to add methods returning enumerators later, let's do it consistently across all the descriptor types.

So some of our other descriptor types (including plain message `Descriptor`) _are_ enumerables though (they include the Enumerable mixin), which is more or less the same thing. So you can do

```
descriptor.each { |field| puts "have field #{field}" }
```

but also things like

```
descriptor.to_a
# or
descriptor.select { |field| field.type == 'int32' }
```

I didn't implement `ServiceDescriptor#each`, however, because I don't think a service is-a collection of RPC methods in the same way that a message type is-a collection of fields. So instead, I implemented `#each_method`, and thus implemented `#methods` to return an enumerable in the same way that `self` is an enumerator for a message descriptor.

If you like I could just implement `#each_method` as `ServiecDescriptor#each` instead, if you think service descriptor is-a list of RPC methods. (as a bonus that would solve the `#methods` method shadowing `Object#methods` too, since we wouldn't need that since `self` _is_ the enumerable then.

ruby/ext/google/protobuf_c/ruby-upb.h
```diff
@@ -11420,7 +11420,7 @@ const upb_FieldDef* upb_DefPool_FindExtensionByNumber(const upb_DefPool* s,
                                                       const upb_MessageDef* m,
                                                       int32_t fieldnum);
 
-const upb_ServiceDef* upb_DefPool_FindServiceByName(const upb_DefPool* s,
+UPB_API const upb_ServiceDef* upb_DefPool_FindServiceByName(const upb_DefPool* s,
```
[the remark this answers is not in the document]
KJTsanaktsidis: I did that _as well_; i'll drop the commit which updates these vendored copies.

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -1294,6 +1302,308 @@ static void EnumDescriptor_register(VALUE module) {
...
+}
+
+static void ServiceDescriptor_register(VALUE module) {
+  VALUE klass = rb_define_class_under(module, "ServiceDescriptor", rb_cObject);
+  rb_define_alloc_func(klass, ServiceDescriptor_alloc);
+  rb_define_method(klass, "initialize", ServiceDescriptor_initialize, 3);
+  rb_define_method(klass, "name", ServiceDescriptor_name, 0);
+  rb_define_method(klass, "each_method", ServiceDescriptor_each_method, 0);
+  rb_define_method(klass, "file_descriptor", ServiceDescriptor_file_descriptor, 0);
+  // n.b. this shadows Object#methods, which seems less than ideal?
```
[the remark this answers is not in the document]
KJTsanaktsidis: I'll rename `#each_method` to `#each` and include `Enumerable` then, which is how the other descriptors list their children. Then I don't need this method at all.

KJTsanaktsidis:
@haberman OK, I dropped the each_method/to_enum stuff in favour of just implementing `#each` and including `Enumerable` like the other descriptor types with children. WDYT?

ruby/ext/google/protobuf_c/defs.c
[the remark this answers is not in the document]
haberman: I see your point, that "service is-a collection of methods" is a bit of a stretch. But methods *are* the only thing you can list inside a `service {}` block in a `.proto` file, other than options, which have their own accessor. The fact that this conveniently resolves the `#methods` shadowing is another argument in its favor.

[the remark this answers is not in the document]
KJTsanaktsidis: alright 👍 the current version of my PR just implements `#each`/`Enumerable` directly on `ServiceDescriptor`. Thanks for talking through that with me.

KJTsanaktsidis:
@haberman anything i can do to move this along?

dorner:
This is exactly what I need :) I'm looking at generating Rails routes from proto options mirroring what grpc-gateway does (so basically allowing us to have two different processes, one running grpc, one running REST, both funneling to the same handler). I'm trying to write a protoc plugin in Ruby (I have one working in Go, but it means that whoever uses it would need a cumbersome "download binary and put in your path" workflow - I'd rather do it in pure Ruby). Right now this is impossible because a) I can't get the original method protos since I can't call `.methods`on the service descriptor, and b) I can't access the options I need even if I could.

KJTsanaktsidis:
Oh, apologies, I didn't see that. Thanks for the tip about how to use extensions from Ruby. It should be good to go now.

haberman (approved):

JasonLunn (approved):

JasonLunn:
@KJTsanaktsidis - would you mind merging HEAD into your branch? There was an internal issue that kept this from merging previously and I think the new test failures will disappear if the branch is updated.

dorner:
@KJTsanaktsidis @haberman wondering if you can help me here. I'm trying to implement a protoc plugin in Ruby where I want to inspect the methods in the input services. However, the fix in this PR doesn't seem to work. The code generation request gives me a `ServiceDescriptionProto`, not a `ServiceDescription`. I also can't look it up in the generated pool because that would require me to first generate the Ruby code, then somehow require it in the plugin itself, which is almost impossible because of how the generated files don't match the expected format (i.e. they don't work with standard Rails autoloading using [zeitwerk)](https://github.com/fxn/zeitwerk/)).

If I inspect the `ServiceDescriptionProto`, I can absolutely see the method descriptors inside it. I just can't access them via `method` or `methods`, because those are reserved methods applicable to all Ruby objects. I can't use `each` on the proto object either.

Any idea for options here? I'm stumped.

JasonLunn:
> @KJTsanaktsidis @haberman wondering if you can help me here. I'm trying to implement a protoc plugin in Ruby where I want to inspect the methods in the input services. However, the fix in this PR doesn't seem to work. The code generation request gives me a `ServiceDescriptionProto`, not a `ServiceDescription`. I also can't look it up in the generated pool because that would require me to first generate the Ruby code, then somehow require it in the plugin itself, which is almost impossible because of how the generated files don't match the expected format (i.e. they don't work with standard Rails autoloading using [zeitwerk)](https://github.com/fxn/zeitwerk/)).
> 
> If I inspect the `ServiceDescriptionProto`, I can absolutely see the method descriptors inside it. I just can't access them via `method` or `methods`, because those are reserved methods applicable to all Ruby objects. I can't use `each` on the proto object either.
> 
> Any idea for options here? I'm stumped.

Can you open a new issue for this use case?

dorner:
Done! #21091

PR #15828, opened 2024-02-13
Fix --maximum_edition usage text

rgov:
For the conformance test runner, `--maximum` is not an option, but `--maximum_edition` is.


mkruskal-google (approved):

mkruskal-google:
whoops, nice catch :)

PR #15836, opened 2024-02-14
[csharp] Fix indentation for MapField with nested Messages

q42jaap:
WriteDictionary does not pass indentationLevel downwards to WriteValue. The existing test for `MapField` only tests `MapField<string, string>`, which isn't affected by the issue.
Tests are added for this issue.

Note: I also added `WriteValueWithIndentation_MapWithEmptyNested` to ensure this change would break things, but I doubt if it's necessary. Let me know whether to remove it.


jskeet (approved):

jskeet:
Thanks, this looks good to me. @mkruskal-google please could you approve after the tests have completed, to kick off the import process?

PR #15846, opened 2024-02-15
Avoid a 'may be used uninitialized' warning when built with '-c opt'.

bcsgh:
The motivation for this is first that under some combinations of flags it breaks builds and second it's trivial (and possibly zero runtime cost) to fix.

It looks like the possibly bad case "should never happen", but converting this to a null pointer in the error case (which should at least make the failure less problematic) is cheap.

### DIFF
--- a/src/google/protobuf/arena.cc
+++ b/src/google/protobuf/arena.cc
@@ -193,7 +193,7 @@ SizedPtr SerialArena::Free(Deallocator deallocator) {
 PROTOBUF_NOINLINE
 void* SerialArena::AllocateAlignedFallback(size_t n) {
   AllocateNewBlock(n);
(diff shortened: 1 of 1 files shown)


bcsgh:
Because the CI refuses to run (and refuses to tell me what to do about that #15847) I have no idea if this is ready for review yet so I'll just kick it out and see what happens.

fowles (approved):

bcsgh:
@fowles so which presubmit did I break internally?

fowles:
> @fowles so which presubmit did I break internally?

It got hung up on something silly and I had to give it kick.  Hopefully it goes through smoothly in the next couple hours.

PR #15849, opened 2024-02-15
Update rules_python to 0.28.0

mering:
Impacting upstream changes:
1. Migrate from `requirements` to `requirements_lock` attribute: bazelbuild/rules_python#1514
1. Replace the removed `pip_install` with `pip_parse` with identical functionality: bazelbuild/rules_python#1570
1. Update `rules_fuzzing` to include bazelbuild/rules_fuzzing#240 and get rid of the downstream patch.


mering:
@acozzette could you please trigger the tests?

zhangskz:
@mering Looks like most tests are failing due to following error in WORKSPACE

```
ERROR: /workspace/WORKSPACE:164:10: //external:pip_deps: no such attribute 'requirements' in 'pip_repository' rule
ERROR: Error computing the main repository mapping: Encountered error while reading extension file 'requirements.bzl': no such package '@pip_deps//': error loading package 'external': Could not load //external package
```

mering:
@zhangskz Thanks for digging this out. I rebased and noticed the same problem. This is because of bazelbuild/rules_python#1514.

So probably we cannot easily split this up and merge #15850 directly. @acozzette @googleberg @zhangskz could someone trigger the CI there?

mering:
@zhangskz I updated a dependency to be compatible with the new version of rules_python and I think I found a way to allow upgrade with minimal changes. Could you please trigger again?

PR #15881, opened 2024-02-20
Add support for RBS, Ruby type signature

qnighy:
This PR adds support for RBS, a standard type signature format for Ruby. With this, Ruby programmers working on google-protobuf get better typechecking/completion/documentation experience with IDEs they are using (e.g. RubyMine and Steep).

"Support" here includes:

- Support for RBS in `protoc`
- RBS type definitions generated by `protoc` for built-in types
- RBS type definitions for the runtime
- RBS type definitions for the private implementation, which can then be partially checked by the Steep type checker

Here is the rationale of the all-in-one change:

- The generated types and the runtime types should work in unison, so it is best to have them in the same repository. Note that the types are not just _a priori_ reflection of the runtime, as they include type aliases, interfaces, and type parameters.
- The generated types describe how the protobuf runtime works, so it is reasonable to write the generator in the same repository as the runtime.

But I'm willing to split the PR into smaller pieces if that's better.

This PR adds a new language `--rbs_out` rather than extending the `--ruby_out` generator. This is to generate these files in a different place: we need to put the type definitions under `./sig`, whereas we usually put the implementation under `./lib`.

The FFI implementation is checked by Steep to some extent, but there are a lot of suppressed errors, many of which require refactoring of the implementation. In this PR, only the trivial fixes to grammatical mistakes are applied.


zhangskz:
There is another issue https://github.com/protocolbuffers/protobuf/issues/9495 and PR https://github.com/protocolbuffers/protobuf/pull/15633 open for this as well -- @haberman can you take a look at this PR and how RBS should best be landed?

HoneyryderChuck:
FWIW I approve the direction of this MR 👍 looking forward to some feedback.

PR #15917, opened 2024-02-22
feat: add visionOS deployment target for Cocoapods

YoloMao:
add visionOS deployment target for Cocoapods

### DIFF
--- a/Protobuf.podspec
+++ b/Protobuf.podspec
@@ -45,6 +45,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.12'
   s.tvos.deployment_target = '12.0'
   s.watchos.deployment_target = '6.0'
+  s.visionos.deployment_target = '1.0'
   s.requires_arc = false
 
   # The unittest need the generate sources from the testing related .proto
--- a/Protobuf.podspec
+++ b/Protobuf.podspec
@@ -46,6 +46,7 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '12.0'
   s.watchos.deployment_target = '6.0'
   s.visionos.deployment_target = '1.0'
+  
   s.requires_arc = false
 
   # The unittest need the generate sources from the testing related .proto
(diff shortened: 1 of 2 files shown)


Protobuf.podspec
```diff
@@ -45,6 +45,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.12'
   s.tvos.deployment_target = '12.0'
   s.watchos.deployment_target = '6.0'
+  s.visionos.deployment_target = '1.0'
```
thomasvl, line 48: When was visionOS support added to CocoaPods?  i.e. - do we need to raise the `s.cocoapods_version = '>= 1.12.0'` line or will versions that old be ok with this directive?

ps - can you add back the blank line that was here to keep the deployment_target info grouped?

thomasvl (approved):
Change for the ObjC podspec is fine

Protobuf.podspec
YoloMao: added blank line, and I saw [Kingfisher](https://github.com/onevcat/Kingfisher/blob/5c21dc6a151367d23e8cc8ae624b35e7182011b2/Kingfisher.podspec#L34) and [SDWebImage](https://github.com/SDWebImage/SDWebImage/blob/80c8b2023a5efb4415a2c615acfec075e5c243d2/SDWebImage.podspec#L9) have not limit this. Maybe it is not nessary, I will take a check about it if I have free time.

thomasvl: @paulb777 do you happen to know if one should to raise the `cocoapods_version` when adding a `visionos.deployment_target` value?

paulb777:
CocoaPods Support was introduced in 1.13.0 and there were fixes in 1.15.0 - so at least 1.13.0 and maybe 1.15.0.

https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md

and there are still outstanding issues that may or may not be solved:

https://github.com/CocoaPods/CocoaPods/issues?q=is%3Aissue+is%3Aopen+visionos

Firebase requires Swift PM for visionOS support.

PR #15983, opened 2024-02-27
Actually fail if staleness fail

mkruskal-google:
PiperOrigin-RevId: 610833812

### DIFF
--- a/.github/workflows/staleness_check.yml
+++ b/.github/workflows/staleness_check.yml
@@ -55,6 +55,7 @@ jobs:
           bazel-cache: staleness
           bash: >
             set -ex;
+            echo "Please run ./regenerate_stale_files.sh to regenerate stale files";
             if [[ -z $COMMIT_TRIGGERED_RUN || -z $MAIN_RUN ]]; then
             bazel query 'attr(tags, "staleness_test", //...)' | xargs bazel test $BAZEL_FLAGS;
             else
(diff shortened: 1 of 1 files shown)


mkruskal-google:
Skipping presubmits, since this workflow doesn't even run

.github/workflows/staleness_check.yml
```diff
@@ -56,8 +56,7 @@ jobs:
           bash: >
             set -ex;
             if [[ -z $COMMIT_TRIGGERED_RUN || -z $MAIN_RUN ]]; then
-            bazel query 'attr(tags, "staleness_test", //...)' | xargs bazel test $BAZEL_FLAGS ||
-            echo "Please run ./regenerate_stale_files.sh to regenerate stale files";
+            bazel query 'attr(tags, "staleness_test", //...)' | xargs bazel test $BAZEL_FLAGS;
```
haberman, line 59: The message seems useful. I didn't know the right command to run to fix this.

Would this work?

```
bazel query 'attr(tags, "staleness_test", //...)' | xargs bazel test $BAZEL_FLAGS || {
  echo "Please run ./regenerate_stale_files.sh to regenerate stale files";
  exit 1;
}
```

mkruskal-google: I think so, but these tests are so fragile I wanted to keep it as simple as possible :(

haberman (approved):

.github/workflows/staleness_check.yml
mkruskal-google: Yea that sgtm

PR #15984, opened 2024-02-27
Actually fail if staleness fail

mkruskal-google:
PiperOrigin-RevId: 610833812


zhangskz (approved):

zhangskz:
Maybe push to dev branch in protocolbuffers so test can actually run? (although I think you tested this already though)

Sidenote: I thought we supposedly exempted ourselves from the "safe for tests" requirement?

mkruskal-google:
Nope, our forks aren't exempt.  And this actually won't run tests from a fork because I touched workflows, changed the base

PR #14565, opened 2023-10-31
feat: automate publishing releases to Bazel Central Registry

alexeagle:
Note, it's also required that someone with sufficient permissions "installs" the Publish-to-BCR GitHub app to this repo.
FYI @thesayyn 

Fixes #14564

Issue #14564, opened 2023-10-31
[bazel] Automate publishing releases to Bazel's central registry


alexeagle:
Currently https://github.com/protocolbuffers/protobuf/releases has many releases beyond what currently appears on https://registry.bazel.build/modules/protobuf, which is currently 10mo old.

We could attempt to keep this updated manually, but like many processes that live in OSS, it will be easy for Googler maintainers to miss this.

Fortunately the Bazel community already created a GitHub app, https://github.com/bazel-contrib/publish-to-bcr/ . With a bit of configuration in this repo, and after "installing" that app to the repo, then every release will be automatically published as a pull request to https://github.com/bazelbuild/bazel-central-registry/pulls

I'm happy to take on this work, just filing the issue first to have a place for discussion. Note that we may need a couple of extra releases when we "launch" in case of mistakes in the automation which are hard to test ahead of time.
FYI @comius @thesayyn

malt3:
Still relevant

erikkerber:
Just hit us (lightly) as well—hoping to pull in a fix for a build warning but BCR is still in 23.1 (current release 25.2)

david-crouse:
Relevant to pybind11_protobuf, we could use latest version in BCR as well.

wffurr:
Why is this closed?  The latest version in the Bazel Central Registry is still 23.1.  bzlmod support is only part of the answer, the rest is publishing to BCR.

alexeagle:
This work is complete, actually. It was a few hours too late to make the last protobuf release, but the next one should include a current MODULE.bazel file and an automation for creating the BCR entry.

wffurr:
Fantastic news - thank you!

PR #16013, opened 2024-03-01
refactor: add MODULE.bazel file for bzlmod

alexeagle:
Step towards #14564. The strategy is:

1. Burndown the set of patches which are currently applied only when protobuf is published to BCR, by applying those patches to the repo.
2. In cases where a patch can't be accepted to the repo, it can be in the .bcr folder.
3. Finish #14565 by configuring the Publish-to-BCR github app.

In this PR, we remove the first patch from the stack of the most recent release: https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/protobuf/23.1/patches

If the maintainers prefer, I could apply a number of these patches in one PR.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,2 +1,51 @@
-# TODO: migrate all dependencies from WORKSPACE to MODULE.bazel
-# https://github.com/protocolbuffers/protobuf/issues/14313
+"""Bazel dependencies
+
+Currently only used by external users via the Bazel Central Registry.
+TODO(alexeagle): empty the WORKSPACE file and enable bzlmod for this repo.
+
+Note, originally this file existed as a patch:
+https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/patches/0001-Add-MODULE.bazel.patch
+"""
+module(
+    name = "protobuf",
+    compatibility_level = 1,
+    version = "23.1",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "rules_python", version = "0.10.2")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "rules_java", version = "4.0.0")
+bazel_dep(name = "rules_pkg", version = "0.7.0")
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "abseil-cpp", repo_name = "com_google_absl", version = "20230802.0.bcr.1")
+bazel_dep(name = "zlib", version = "1.2.11")
+bazel_dep(name = "upb", version = "0.0.0-20230516-61a97ef")
+
+# TODO: Add missing rules_kotlin
+
+# Maven dependencies
+bazel_dep(name = "rules_jvm_external", version = "4.4.2")
+
+maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
+
+maven.install(
+    name = "maven",
+    artifacts = [
+        "com.google.code.findbugs:jsr305:3.0.2",
+        "com.google.code.gson:gson:2.8.9",
+        "com.google.errorprone:error_prone_annotations:2.3.2",
+        "com.google.j2objc:j2objc-annotations:1.3",
+        "com.google.guava:guava:31.1-jre",
+        "com.google.guava:guava-testlib:31.1-jre",
+        "com.google.truth:truth:1.1.2",
+        "junit:junit:4.13.2",
+        "org.mockito:mockito-core:4.3.1",
+    ],
+)
+
+use_repo(maven, "maven")
+
+# Dependencies needed in tests
+bazel_dep(name = "googletest", repo_name = "com_google_googletest", version = "1.11.0")


mmorel-35:
I also started this #15976 earlier ;)
I believe you intend to make it one patch at a time which is going to be easier to review.

alexeagle:
Yes and also to test the waters to see what the maintainers plan to do

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+TODO(alexeagle): empty the WORKSPACE file and enable bzlmod for this repo.
+
+Note, originally this file existed as a patch:
+https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/patches/0001-Add-MODULE.bazel.patch
+"""
+module(
+    name = "protobuf",
+    compatibility_level = 1,
+    version = "23.1",
+)
```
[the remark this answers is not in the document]
alexeagle: My intent is to start by un-forking, which means we literally take the existing content that is in the last published protobuf version on BCR: 
https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/patches/0001-Add-MODULE.bazel.patch
So I think as a matter of principle we shouldn't consider changing it as part of this PR, first we should converge the content and then propose changes separately.

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+        "com.google.truth:truth:1.1.2",
+        "junit:junit:4.13.2",
+        "org.mockito:mockito-core:4.3.1",
+    ],
+)
+
+use_repo(maven, "maven")
+
+# Dependencies needed in tests
+bazel_dep(name = "googletest", repo_name = "com_google_googletest", version = "1.11.0")
```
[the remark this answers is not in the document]
alexeagle: I do think that's a mistake on https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/MODULE.bazel but like the previous comment, I'd like to start by un-forking and then consider changes separately.

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+        "com.google.code.findbugs:jsr305:3.0.2",
+        "com.google.code.gson:gson:2.8.9",
+        "com.google.errorprone:error_prone_annotations:2.3.2",
+        "com.google.j2objc:j2objc-annotations:1.3",
+        "com.google.guava:guava:31.1-jre",
+        "com.google.guava:guava-testlib:31.1-jre",
+        "com.google.truth:truth:1.1.2",
+        "junit:junit:4.13.2",
+        "org.mockito:mockito-core:4.3.1",
+    ],
```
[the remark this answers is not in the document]
alexeagle: Again, intentionally unforking https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/MODULE.bazel first, then changing separately. Want to have a clear git history.

alexeagle:
Hey @haberman could we get some feedback from maintainers?

mkruskal-google:
Hey Alex, Sandy's OOO until next week, but I can try digging into this in the meantime.

QQ: does any of this take effect immediately or does it only apply after our next release?

alexeagle:
Oh, you're going to have a lot of questions since Bazel is so different from Blaze, and you have no repository rules or module extensions or MODULE.bazel files...
Anyhow this only would have an effect if a Bazel user with `--enable_bzlmod` used a git SHA to fetch `protobuf` rather than getting a release, like with [git_override](https://bazel.build/rules/lib/globals/module#git_override). That wouldn't work for them anyhow, so the simple answer is no, it doesn't have an effect until we release and also publish that release to the BCR

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "rules_python", version = "0.10.2")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "rules_java", version = "4.0.0")
+bazel_dep(name = "rules_pkg", version = "0.7.0")
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "abseil-cpp", repo_name = "com_google_absl", version = "20230802.0.bcr.1")
+bazel_dep(name = "zlib", version = "1.2.11")
+bazel_dep(name = "upb", version = "0.0.0-20230516-61a97ef")
```
mkruskal-google, line 24: upb has been merged into the protobuf repo, do we still need this?

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "rules_python", version = "0.10.2")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "rules_java", version = "4.0.0")
+bazel_dep(name = "rules_pkg", version = "0.7.0")
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "abseil-cpp", repo_name = "com_google_absl", version = "20230802.0.bcr.1")
```
mkruskal-google, line 22: Is there a newer version on BCR?  We're currently using 20240116.0

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+    compatibility_level = 1,
+    version = "23.1",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "rules_python", version = "0.10.2")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "rules_java", version = "4.0.0")
+bazel_dep(name = "rules_pkg", version = "0.7.0")
```
mkruskal-google, line 20: All of these dependencies seem older than what we're using now.  Why are we downgrading?

MODULE.bazel
```diff
@@ -1,2 +1,51 @@
...
+Currently only used by external users via the Bazel Central Registry.
+TODO(alexeagle): empty the WORKSPACE file and enable bzlmod for this repo.
+
+Note, originally this file existed as a patch:
+https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/patches/0001-Add-MODULE.bazel.patch
+"""
+module(
+    name = "protobuf",
+    compatibility_level = 1,
+    version = "23.1",
```
mkruskal-google, line 12: We're currently working on 26.0 (you could say main is on 27.0), it seems like all of these dependencies are stale

mkruskal-google:
IIUC, this is a direct port from the patches of 23.1 from BCR.  While adding these changes incrementally makes sense, I don't see the value in submitting *broken* configs that we can't even test.  Ideally we would have bzlmod CI setup and make these changes correctly from the start

alexeagle:
You'd need to add a new test fixture in this repo to verify the next release will work in bzlmod. The BCR presubmit test only runs after the release is cut and the artifact is published.

mkruskal-google:
Yea I'd like to add new tests either with this PR or right after, but my point is more that this config is incorrect and won't work IIUC.  It's based on a very old version, so I don't see the value in submitting this as-is

alexeagle:
> I don't see the value in submitting this as-is

Okay, I'll close this then. I'll reach out to you on chat so we can figure out what shape you want these new tests to be.

MODULE.bazel
alexeagle: Yes, there probably is, but these are lower bounds. We could artificially raise them, but we shouldn't worry about that.

alexeagle:
Replaced by https://github.com/protocolbuffers/protobuf/pull/16319

PR #16014, opened 2024-03-01
feat: automate publishing releases to Bazel Central Registry

alexeagle:
Note, recent releases on the BCR have a patch set applied, and it seems these patches are developed independently to "fix" each protobuf release, rather than make changes to protobuf repo.

The effect of this PR will be to create a *broken* publish to BCR for each protobuf release. At least this red PR on BCR will be our indication that the patches need to be manually replayed there.

In parallel, starting with #16013 I'll apply as many of those patches to the protobuf repo as possible. That will reduce the manual effort for each release.

Replaces https://github.com/protocolbuffers/protobuf/pull/14565 which originated from my fork so the tests wouldn't run.

### DIFF
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,0 +1,4 @@
+# see https://github.com/bazel-contrib/publish-to-bcr/#a-note-on-release-automation
+fixedReleaser:
+  login: thesayyn
+  email: sahin@aspect.dev
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,4 +1,0 @@
-# see https://github.com/bazel-contrib/publish-to-bcr/#a-note-on-release-automation
-fixedReleaser:
-  login: thesayyn
-  email: sahin@aspect.dev
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -2,18 +2,19 @@
   "homepage": "https://github.com/protocolbuffers/protobuf",
   "maintainers": [
     {
-      "github": "comius",
-      "name": "Ivo List"
+      "email": "sandyzhang@google.com",
+      "github": "zhangskz",
+      "name": "Sandy Zhang"
     },
     {
-      "email": "alex@aspect.dev",
-      "github": "alexeagle",
-      "name": "Alex Eagle"
+      "email": "mkruskal@google.com",
+      "github": "mkruskal-google",
+      "name": "Mike Kruskal"
     },
     {
-      "email": "sahin@aspect.dev",
-      "github": "thesayyn",
-      "name": "Şahin Yort"
+      "email": "gberg@google.com",
+      "github": "googleberg",
+      "name": "Jerry Berg"
     }
   ],
   "repository": ["github:protocolbuffers/protobuf"],
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -1,0 +1,22 @@
+{
+  "homepage": "https://github.com/protocolbuffers/protobuf",
+  "maintainers": [
+    {
+      "github": "comius",
+      "name": "Ivo List"
+    },
+    {
+      "email": "alex@aspect.dev",
+      "github": "alexeagle",
+      "name": "Alex Eagle"
+    },
+    {
+      "email": "sahin@aspect.dev",
+      "github": "thesayyn",
+      "name": "Şahin Yort"
+    }
+  ],
+  "repository": ["github:protocolbuffers/protobuf"],
+  "versions": [],
+  "yanked_versions": {}
+}
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,0 +1,24 @@
+matrix:
+  platform: ["debian10", "macos", "ubuntu2004"]
+
+tasks:
+  verify_targets:
+    name: "Verify build targets"
+    platform: ${{ platform }}
+    build_targets:
+    - '@protobuf//:all'
+    - '@protobuf//java/core/...'
+    - '@protobuf//java/lite/...'
+    - '-@protobuf//:common_dist_files'
+    - '-@protobuf//:proto_api'
+
+bcr_test_module:
+  module_path: "examples"
+  matrix:
+    platform: ["centos7", "debian10", "macos", "ubuntu2004"]
+  tasks:
+    run_test_module:
+      name: "Run test module"
+      platform: ${{ platform }}
+      build_targets:
+      - "//..."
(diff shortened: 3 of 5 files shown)


.bcr/config.yml
```diff
@@ -0,0 +1,4 @@
+# see https://github.com/bazel-contrib/publish-to-bcr/#a-note-on-release-automation
+fixedReleaser:
+  login: thesayyn
+  email: sahin@aspect.dev
```
mkruskal-google, line 4: Can these be set to some team email?  At the very least it should be one of us

.bcr/metadata.template.json
```diff
@@ -0,0 +1,22 @@
...
+    {
+      "email": "alex@aspect.dev",
+      "github": "alexeagle",
+      "name": "Alex Eagle"
+    },
+    {
+      "email": "sahin@aspect.dev",
+      "github": "thesayyn",
+      "name": "Şahin Yort"
+    }
```
mkruskal-google, line 17: Given that we're moving this to our repo, shouldn't we be the maintainers?

.bcr/presubmit.yml
```diff
@@ -0,0 +1,24 @@
+matrix:
+  platform: ["debian10", "macos", "ubuntu2004"]
```
mkruskal-google, line 2: Can we add windows here too?

.bcr/presubmit.yml
```diff
@@ -0,0 +1,24 @@
...
+tasks:
+  verify_targets:
+    name: "Verify build targets"
+    platform: ${{ platform }}
+    build_targets:
+    - '@protobuf//:all'
+    - '@protobuf//java/core/...'
+    - '@protobuf//java/lite/...'
+    - '-@protobuf//:common_dist_files'
+    - '-@protobuf//:proto_api'
```
mkruskal-google, line 13: We should probably run some tests too

.bcr/config.yml
alexeagle: Sure, can you propose an edit of who on your team plans to respond to emails about publishing failures to the BCR?

.bcr/metadata.template.json
alexeagle: sure, I can't see a list of who is on the team, please propose an edit?

.bcr/presubmit.yml
alexeagle: yes, updated to match the content of the last time the module was published: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/presubmit.yml

.bcr/presubmit.yml
alexeagle: I'm not sure that any tests will pass on the BCR, and this PR doesn't include that testing. I think we should stick to what was last published in https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/presubmit.yml first, just to un-fork the situation and get the protobuf repo at HEAD to represent the state that's currently published to BCR, then edit after.

.bcr/config.yml
mkruskal-google: unless we can put a group here, let's just remove this file?  I don't think we want a fixed author if it has to be an individual

.bcr/metadata.template.json
mkruskal-google: {
      "email": "sandyzhang@google.com",
      "github": "zhangskz",
      "name": "Sandy Zhang"
    },
    {
      "email": "mkruskal@google.com",
      "github": "mkruskal-google",
      "name": "Mike Kruskal"
    },
    {
      "email": "gberg@google.com",
      "github": "googleberg",
      "name": "Jerry Berg"
    }

would be a decent placeholder until we setup a group email for this

.bcr/config.yml
alexeagle: Sure, I can remove this file, but depending on how you create releases, it may fail to publish.

Publish to BCR uses information about the GitHub author of a release in order to tag that author in the commit, push an entry to their BCR fork, or send error notifications. If you use a GitHub action to automate your release and the author is the `github-actions` bot (likely because you used the `GITHUB_TOKEN` secret to authorize the action), then the app won't know who cut the release and may not function properly.

.bcr/metadata.template.json
alexeagle: thanks - FYI github has a feature where you can just "pencil" on my file and make the edit yourself, or, you can make changes inside a code suggestion like critique

.bcr/config.yml
alexeagle: see https://github.com/bazel-contrib/publish-to-bcr/blob/262c9e15a3c556cb6ecc7ce67e31956acf91e201/README.md#a-note-on-release-automation

mkruskal-google: fair enough, I guess it's an arbitrary choice anyway.  Let's just go with one of the names below?

mkruskal-google (approved):

.bcr/metadata.template.json
mkruskal-google: yea I tried that, but the only option seemed to be to commit the changes directly to your branch.  That seems much more heavy-handed than a "suggestion"

.bcr/config.yml
alexeagle: I'll just remove the file for this pass, it's possible if you're creating manual GH releases that the releaser identity will propagate correctly.

.bcr/metadata.template.json
alexeagle: There's a subtle button right above the textarea where you type a comment, it looks like a page with a plus/minus on it, that's where you can make suggested edits. But yeah this is fine too :)

.bcr/metadata.template.json
```diff
@@ -0,0 +1,23 @@
...
+      "name": "Mike Kruskal"
+    },
+    {
+      "email": "gberg@google.com",
+      "github": "googleberg",
+      "name": "Jerry Berg"
+    }
+  ],
+  "repository": ["github:protocolbuffers/protobuf"],
+  "versions": [],
```
zhangskz, line 21: Should this be propagated with the current versions published to BCR?

  "versions": [
    "3.19.0",
    "3.19.2",
    "3.19.6",
    "21.7",
    "23.1"
  ],
  "yanked_versions": {
    "3.19.0": "CVE-2022-3171 (https://github.com/advisories/GHSA-h4h5-3hr4-j3g2)",
    "3.19.2": "CVE-2022-3171 (https://github.com/advisories/GHSA-h4h5-3hr4-j3g2)"
  }

alexeagle: no, this file is the template to be filled in by the Publish to BCR app

zhangskz (approved):

PR #16033, opened 2024-03-04
[C#] Eliminate string concatenation when looking for enum values descriptor

TrayanZapryanov:
While deserializing our definitions I've found following concatenation:
image

Using same pattern already used for enum lookup by number, I've created a struct key and used it in a new Dictionary.
This helps eliminating this concatenation:
image

### DIFF
--- a/csharp/src/Google.Protobuf/ObjectFullNamePair.cs
+++ b/csharp/src/Google.Protobuf/ObjectFullNamePair.cs
@@ -1,0 +1,51 @@
+#region Copyright notice and license
+// Protocol Buffers - Google's data interchange format
+// Copyright 2015 Google Inc.  All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+#endregion
+
+using System;
+using System.Collections.Generic;
+
+namespace Google.Protobuf
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


jskeet:
I won't have time to look at this for a while - I'm still trying to find sufficient time to get your previous PRs evaluated appropriately.

jskeet (commented):
I'm fine with the idea of this, but I've created a branch in my fork with some suggested changes. (I'd really like to get rid of ObjectIntPair, but that's not quite as simple as I'd hoped at the moment.)

Please have a look at https://github.com/jskeet/protobuf/commit/f4fe37cce1e71c0e5fec626e5c8b910a29530314 - if you're happy with that, I'll create a new change (probably internally, as it's easier to push out that way) based on the current code (rather than worrying about rebasing this PR).

TrayanZapryanov:
@jskeet I like your code more.
Should I close this PR?

jskeet:
Yup, that's fine. I'll start the process of getting my change through - you won't see a PR, because it'll all be internal until it gets merged and published.

jskeet:
(I've now accidentally force-pushed a squashed version to your fork instead of mine. I'm glad we were already closing the PR...)

jskeet:
(I've re-pushed your original commit, just so it's all clearer...)

jskeet:
This was implemented in https://github.com/protocolbuffers/protobuf/commit/e6684ac037940c962c7fc3bd7b247b0e583bae2e

PR #16094, opened 2024-03-07
[C#] Optimize reading in PushBackReader by removing caching into nullable char.

TrayanZapryanov:
I've found that PushBackReader is storing next character in nullable char field. And accessing it is visible in CPU profiler:
image

This PR is converting it to int and checks for -1 instead of .HasValue.
image

Using [benchmark](https://github.com/protocolbuffers/protobuf/pull/15795#issuecomment-1980222331) gives following results
Main:
// * Summary *

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3235/23H2/2023Update/SunValley3)
11th Gen Intel Core i9-11900K 3.50GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.200
  [Host]     : .NET 8.0.2 (8.0.224.6711), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.2 (8.0.224.6711), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI

Job=DefaultJob
| Method | Mean         | Allocated |
|------- |-------------:|----------:|
| Small  |     295.7 ns |   1.01 KB |
| Big    | 265,443.3 ns |  642.7 KB |

PR
| Method | Mean         | Allocated |
|------- |-------------:|----------:|
| Small  |     241.6 ns |   1.01 KB |
| Big    | 207,593.5 ns |  642.7 KB |

### DIFF
--- a/csharp/src/Google.Protobuf/JsonTokenizer.cs
+++ b/csharp/src/Google.Protobuf/JsonTokenizer.cs
@@ -204,13 +204,14 @@ namespace Google.Protobuf
                 while (true)
                 {
                     var next = reader.Read();
-                    if (next == null)
+                    if (next == -1)
                     {
                         ValidateState(State.ExpectedEndOfDocument, "Unexpected end of document in state: ");
                         state = State.ReaderExhausted;
                         return JsonToken.EndDocument;
                     }
-                    switch (next.Value)
+                    char nextChar = (char)next;
+                    switch (nextChar)
                     {
                         // Skip whitespace between tokens
                         case ' ':
@@ -279,11 +280,11 @@ namespace Google.Protobuf
                         case '7':
                         case '8':
                         case '9':
-                            double number = ReadNumber(next.Value);
+                            double number = ReadNumber(nextChar);
                             ValidateAndModifyStateForValue("Invalid state to read a number token: ");
                             return JsonToken.Value(number);
                         default:
-                            throw new InvalidJsonException("Invalid first character of token: " + next.Value);
+                            throw new InvalidJsonException("Invalid first character of token: " + nextChar);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


jskeet:
I'll have a closer look, but this really feels like it's hurting the readability of the code.

I don't know when I'll have time to look at any of this though. I don't think there's very much point in creating an ever-growing backlog of micro-optimisations for JSON parsing, to be honest.

TrayanZapryanov:
I understand that JSON parsing is not in the focus and already start a change in our code to migrate to binary serialization.
But still see a value in all these PRs as they slowly closing the gap between both kinds of deserializations.
I haven't saw somewhere a guideline that is advising people to prefer one over the other and maybe this is what is missing and should be documented. Please fill free in closing all PRs which you don't like - I am really satisfied that found/suggested them and once combined they lead to following results:
Start:
| Method | Mean         | Allocated |
|------- |-------------:|----------:|
| Small  |     295.7 ns |   1.01 KB |
| Big    | 265,443.3 ns |  642.7 KB |

All in one:
| Method | Mean         | Allocated |
|------- |-------------:|----------:|
| Small  |     261.2 ns |     456 B |
| Big    | 190,585.2 ns |  118008 B |

Thanks for review and guidance.

jskeet:
It's not that there won't eventually be a benefit - it's that I don't see much point in stacking up an ever-growing number of them which will potentially end up causing merge conflicts with one another. I would suggest staggering your effort on this - work out which *one* PR you most want to land next, I can try to find time to review that (but it'll take a while - Protobuf isn't my main project, and I have other Protobuf work to do as well) and get it merged, then we can move on to the next one.

jskeet:
As ever, I'll have a look when I get time, but there's only so much time I can put into these PRs. (I'm expecting to basically put time into one per week.)

jskeet:
I've had a closer look, and given that it's purely internal within the class (and nested classes) *and* that the performance difference is significant, I think it's worth going with the `int` version.

I've created an alternative for the PR [here](https://github.com/protocolbuffers/protobuf/commit/454abe07ce1250767b80777309a3f1c124e4684e) - please have a look and see if that looks okay to you. If it does, I'll push it through internally like the last PR.

There are a few changes in there compared with your commit, to take further advantage of just having a single value - as well as more comments to make up for the "odd" choice of type.

jskeet:
Closing now that 68e6e3e8eecd31b0078422f688a3f01fb881d367 is merged.

PR #16123, opened 2024-03-11
[python] Remove semicolon after `PyObject_HEAD`

goffrie:
`PyObject_HEAD` already expands to a field with a semicolon; the extra semicolon is unnecessary, and makes some compilers unhappy.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,4 @@
 BasedOnStyle: Google
 DerivePointerAlignment: false
 PointerAlignment: Left
+StatementMacros: [PyObject_HEAD]
--- a/python/descriptor.c
+++ b/python/descriptor.c
@@ -23,7 +23,7 @@
 // This representation is used by all concrete descriptors.
 
 typedef struct {
-  PyObject_HEAD;
+  PyObject_HEAD
   PyObject* pool;          // We own a ref.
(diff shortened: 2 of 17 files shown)


haberman:
The only downside of this is that clang-format doesn't know about this convention, and will reformat it as:

```c
typedef struct {
  PyObject_HEAD PyObject* pool;          // We own a ref.
  const void* def;         // Type depends on the class. Kept alive by "pool".
  PyObject* options;       // NULL if not present or not cached.
  PyObject* features;      // NULL if not present or not cached.
  PyObject* message_meta;  // We own a ref.
} PyUpb_DescriptorBase;
```

Is there a flag we can add to suppress the warning instead?

goffrie:
Ah, I see. I believe that can be worked around by adding PyObject_HEAD to `.clang-format`'s `StatementMacros`. Let me see.

goffrie:
With the change to the `.clang-format` config, clang-format now leaves PyObject_HEAD on its own line.

haberman:
I haven't (yet) been able to find a way to get the `.clang-format` config file to be respected in the google3 source tree.

googleberg:
@ericsalo is going to take a look at whether the clang formatter can be satisfied.

zhangskz:
@ericsalo have you taken a look at this yet?

PR #16125, opened 2024-03-11
[python] Remove explicit visibility("default") on PyInit__message

goffrie:
This is unnecessary as PyMODINIT_FUNC already implies Py_EXPORTED_SYMBOL.


goffrie:
ping

googleberg:
@haberman this looks like a fairly simple review

haberman (approved):

JasonLunn:
@goffrie - I tried to rerunning tests on this today, and found two clusters of issues with upb:
1. Did `PyMODINIT_FUNC` have a significant change in 3.8? I ask because I see the `//python:conformance_test_upb` target fails just on Python 3.8. We could potentially punt on those if you're content to let this sit for a while longer until 3.8 support is dropped.
1. On macos and ubuntu, upb Python wheel tests failed on both Python 3.8 and 3.12, but passed on windows, suggesting a platform-specific behavior for this macro. It may be the case that we still need to be explicit in order to get consistent behavior across the Protobuf support matrix.

acozzette (approved):

PR #16171, opened 2024-03-14
fix: pin bazel → `6.3.2`

sgammon:
Fixes and closes protocolbuffers/protobuf#16169 by pinning the build to Bazel `6.3.2`.

PR #16172, opened 2024-03-14
fix: support for bazel `7.x`

sgammon:
## Summary

Fixes and closes protocolbuffers#16169 by amending the build to support the latest version of Bazel (`7.1.0`). This is an alternative PR to #16171.

Tested against:
- Bazel `7.1.0`
- Bazel `7.0.0`
- Bazel `6.3.2`

On:
```
macOS 14.3.1 (aarch64)

Local Java:
openjdk version "11.0.19" 2023-04-18 LTS
OpenJDK Runtime Environment Zulu11.64+19-CA (build 11.0.19+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.64+19-CA (build 11.0.19+7-LTS, mixed mode)
```

## Changelog

- fix: use `to_list()[0]` for depset
- fix: upgrade `rules_jvm_external` → latest
- fix: upgrade `rules_kotlin` → `1.9.0`, to match `rules_jvm_external`
- fix: `io_bazel_rules_kotlin` → `rules_kotlin`
- fix: target for `kotlin-stdlib`
- chore: re-generate maven dependency lockfile


java/kotlin/BUILD.bazel
```diff
@@ -50,10 +50,23 @@ kt_jvm_library(
     deps = ["//java/core"],
 )
 
+java_import(
+    name = "kotlin-stdlib-jvm",
```
haberman, line 54: Why is this indirection necessary?

sgammon: It's not necessary, and I could remove it; the target changed within `@com_github_jetbrains_kotlin`, and I didn't want to expose the `java_import` etc functionality. I felt it was cleaner to have a single alias used in both `//java/kotlin` and `//java/kotlin-lite`.

But this is a matter of taste and I would defer to Google and/or other advice :)

sgammon: Unless you mean the `java_import`? That is necessary because the target is now a JAR file, there is no `kotlin-stdlib` in that dependency anymore. Rules Kotlin was upgraded because Rules JVM was upgraded, which happened because the earlier version depends on toolchain internals which changed with Bazel 7.x.

sgammon: Sorry, @haberman; maybe you didn't get these because I forgot to tag. My mentions is where I'm usually looking.

java/osgi/osgi.bzl
```diff
@@ -174,7 +174,7 @@ def _osgi_jar_impl(ctx):
         source_jars = source_jars.to_list()
     if len(source_jars) > 1:
         fail("osgi_jar rule doesn't know how to deal with more than one source jar.")
-    source_jar = target_java_output.source_jars[0]
+    source_jar = target_java_output.source_jars.to_list()[0]
```
zhangskz, line 177: You may need to rebase -- I believe this one was already fixed in https://github.com/protocolbuffers/protobuf/commit/925de189ecb8aa9b7dcd733ad46c651c12ca42d2

sgammon: Thank you, will do.

zhangskz:
This seems to be failing CI tests with the following:

```
ERROR: /workspace/_build/out/external/local_jdk/BUILD.bazel:82:19: no such target '@bazel_tools//tools/jdk:bootstrap_runtime_toolchain_type': target 'bootstrap_runtime_toolchain_type' not declared in package 'tools/jdk' defined by /workspace/_build/out/external/bazel_tools/tools/jdk/BUILD (Tip: use `query "@bazel_tools//tools/jdk:*"` to see all the targets in that package) and referenced by '@local_jdk//:bootstrap_runtime_toolchain_definition'
```

sgammon:
@zhangskz Ah, that surfaces because `rules_java` needs to be upgraded. I will handle it in my rebase

haberman:
I think Bazel 7.x is working now. Can we close this PR?

sgammon:
I think so! Thanks @haberman

PR #16194, opened 2024-03-15
Table parser upstream

gerben-stavenga:
This is similar to earlier PR, but with a smaller diff and removal of changes orthogonal to the table parser.

To facilitate syncing generated pb.cc files you can temporarily disable the "FastParseLoop" and add a dummy constructor for FastFieldEntry that allows compiling against legacy pb.cc files, relying solely on MiniParseLoop. Then you can regenerate all checked-in generated code.


src/google/protobuf/generated_message_tctable_gen.cc
```diff
@@ -251,170 +99,33 @@ bool IsFieldEligibleForFastParsing(
...
-      uint8_t rmax_value;
-      if (!message_options.uses_codegen &&
-          GetEnumRangeInfo(field, rmax_value) == EnumRangeInfo::kNone) {
-        // We can't use fast parsing for these entries because we can't specify
-        // the validator.
-        // TODO: Implement a fast parser for these enums.
-        return false;
-      }
+    case FieldDescriptor::TYPE_ENUM:
+      if (!cpp::HasPreservingUnknownEnumSemantics(field)) return false;
```
seongkim75, line 111: Why did you make this change?

gerben-stavenga: The fast loop path of enums / varints is the same for proto3. For proto2 we have to do additional stuff so it needs a fallback

src/google/protobuf/generated_message_tctable_gen.cc
```diff
@@ -251,170 +99,33 @@ bool IsFieldEligibleForFastParsing(
...
-      if (field->options().ctype() == FieldOptions::STRING) {
-        // strings are fine...
-      } else if (field->options().ctype() == FieldOptions::CORD) {
-        // Cords are worth putting into the fast table, if they're not repeated
-        if (field->is_repeated()) return false;
-      } else {
+      if (field->options().ctype() != FieldOptions::STRING) {
         return false;
       }
-      if (options.is_string_inlined) {
```
gerben-stavenga, line 270: I guess we need to exclude inlined strings for now as well

src/google/protobuf/generated_message_tctable_full.cc
```diff
@@ -58,15 +59,22 @@ const char* TcParser::ReflectionFallback(PROTOBUF_TC_PARAM_DECL) {
   }
 
   SyncHasbits(msg, hasbits, table);
+
+  auto* full_msg = DownCast<Message*>(msg);
+
+  if ((table->extension_offset & 1) != 0) {
```
seongkim75, line 65: Also curious about the rationale for this change.

seongkim75, line 65: Also curious about this change as well.

gerben-stavenga: Something changed when merging with the latest versions where message set has been moved to table driven parsing as well. I incorporated this by adding this to indicate message set

src/google/protobuf/arenastring.cc
```diff
@@ -264,14 +264,12 @@ void ArenaStringPtr::ClearToDefault(const LazyString& default_value,
 }
 
 const char* EpsCopyInputStream::ReadArenaString(const char* ptr,
+                                                int size,
```
seongkim75, line 267: Is this change required?

gerben-stavenga: yes the fast parse loop already reads the size.

src/google/protobuf/generated_message_tctable_impl.h
```diff
@@ -795,17 +803,27 @@ class PROTOBUF_EXPORT TcParser final {
...
+            ptr, ctx);
+      }
     }
   }
 
   template <class MessageBaseT>
   static const char* MessageSetWireFormatParseLoopImpl(
       PROTOBUF_TC_PARAM_NO_DATA_DECL) {
-    return RefAt<ExtensionSet>(msg, table->extension_offset)
+    return RefAt<ExtensionSet>(msg, table->extension_offset & -2)
```
seongkim75, line 826: Why do we need this bit masking? Would it work with and without new parser?

gerben-stavenga: In order to support message set natively (which was introduced in main line recently). This was introduced by a special 1 entry fast path handler that would branch to MessageSetWireFormatParseLoopImpl. The new table doesn't store function pointers so I added message set in the fallback parse function. I use the lowest bit of the offset of the extension set (which due to alignment is always 0), to indicate if this proto is a message set. In which case the parser just calls the fallback, which will call this function.

src/google/protobuf/compiler/cpp/parse_function_generator.cc
```diff
@@ -412,8 +416,7 @@ ::_pbi::TcParser::GetTable<$hottest_type_name$>(),  // to_prefetch
       if (line_entries) format("\n");
       format("65535, 65535\n");
     }
-    if (ordered_fields_.empty() &&
-        !descriptor_->options().message_set_wire_format()) {
```
seongkim75, line 416: This might have caused issues internally. We still need aux table for message set. Why did you drop this?

src/google/protobuf/generated_message_tctable_gen.cc
```diff
@@ -753,34 +464,6 @@ TailCallTableInfo::TailCallTableInfo(
...
-    ABSL_DCHECK(ordered_fields.empty());
-    ABSL_DCHECK(inlined_string_indices.empty());
-    if (message_options.uses_codegen) {
-      fast_path_fields = {{TailCallTableInfo::FastFieldInfo::NonField{
-          message_options.is_lite
-              ? TcParseFunction::kMessageSetWireFormatParseLoopLite
-              : TcParseFunction::kMessageSetWireFormatParseLoop,
-          0, 0}}};
-
-      aux_entries = {{kSelfVerifyFunc}};
```
seongkim75, line 766: Removing this is problematic as we internally need verify function in aux table.

src/google/protobuf/generated_message_tctable_gen.cc
gerben-stavenga: Perhaps if you should add "return false" in this case for pdproto

src/google/protobuf/generated_message_tctable_decl.h
```diff
@@ -391,12 +401,21 @@ struct alignas(uint64_t) TcParseTableBase {
...
+      if (h >= (1 << 16)) NotConstexpr();
+      if (h < -1) NotConstexpr();
+      if (a >= (1 << 14)) NotConstexpr();
+    }
+    uint64_t type_card : 16;  // `FieldType` and `Cardinality` (see _impl.h)
+    uint64_t has_idx : 16;     // has-bit index, relative to the message object
+    uint64_t aux_idx : 14;    // index for `field_aux`.
+    uint64_t offset : 18;     // offset in the message object
 
     static constexpr uint16_t kNoAuxIdx = 0xFFFF;
```
seongkim75, line 417: If aux_idx becomes 14 bits instead of 16bits, kNoAuxIdx needs to be changed as well, otherwise, internal check against kNoAuxIdx fails. A temp fix is to make it 0x3FFFF.

gerben-stavenga: That makes a lot of sense

src/google/protobuf/generated_message_tctable_lite.cc
```diff
@@ -2816,6 +2848,438 @@ PROTOBUF_NOINLINE const char* TcParser::MpMap(PROTOBUF_TC_PARAM_DECL) {
...
+  static_cast<char*>(x)[idx / 8] |= 1 << (idx & 7);
+}
+
+inline void Store(uint64_t value, void* out, uint64_t fd, void* dummy) {
+    using FFE = TcParseTableBase::FastFieldEntry;
+    unsigned offset = fd >> FFE::kOffsetShift;
+    out = static_cast<char*>(out) + offset;
+    *static_cast<bool*>(out) = value;
+    auto dst = (fd & FFE::kRep32Bit) ? out : dummy;
+    asm volatile (""::"r"(dst));
```
seongkim75, line 2865: Why do you need this?

src/google/protobuf/generated_message_tctable_lite.cc
```diff
@@ -2816,6 +2848,438 @@ PROTOBUF_NOINLINE const char* TcParser::MpMap(PROTOBUF_TC_PARAM_DECL) {
...
+  x = (fd & FFE::kCardMask) == FFE::kSingular ? dummy : x;
+  static_cast<char*>(x)[idx / 8] |= 1 << (idx & 7);
+}
+
+inline void Store(uint64_t value, void* out, uint64_t fd, void* dummy) {
+    using FFE = TcParseTableBase::FastFieldEntry;
+    unsigned offset = fd >> FFE::kOffsetShift;
+    out = static_cast<char*>(out) + offset;
+    *static_cast<bool*>(out) = value;
+    auto dst = (fd & FFE::kRep32Bit) ? out : dummy;
```
seongkim75, line 2864: This adds branch so this code isn't exactly branch'less, right? Is it still better than switch due to at most two branches (compared to 3 in switch)?

PR #16308, opened 2024-03-26
Set label to REQUIRED for descriptors with LEGACY_REQUIRED feature.

zhangskz:
Ensures isOptional() does not return true for LEGACY_REQUIRED fields which would otherwise get the optional label applied by default (non-optional fields still get the optional label).

Adds validation to feature resolution instead of cross link, which is too early to have FieldPresence.LEGACY_REQUIRED resolved.

PiperOrigin-RevId: 618857590

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/Descriptors.java
+++ b/java/core/src/main/java/com/google/protobuf/Descriptors.java
@@ -639,7 +639,7 @@ public final class Descriptors {
         if (this.features != null) {
           return;
         }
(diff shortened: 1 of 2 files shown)


ericsalo (approved):

mkruskal-google:
The test failure here looks real.  I tried rerunning and it still fails, and it's from JRuby so it does depend on the Java runtime

ericsalo (approved):

ericsalo (approved):

mkruskal-google (approved):

PR #16319, opened 2024-03-26
feat: test bzlmod usage on CI

alexeagle:


MODULE.bazel
```diff
@@ -1,2 +1,22 @@
 # TODO: migrate all dependencies from WORKSPACE to MODULE.bazel
 # https://github.com/protocolbuffers/protobuf/issues/14313
+module(
+    name = "protobuf",
+    compatibility_level = 1,
+    version = "23.1",
```
mkruskal-google, line 6: Could we bump this up to 27.0 to match protobuf_version.bzl?  Or ideally hook it up to that file if it's possible

MODULE.bazel
```diff
@@ -1,2 +1,22 @@
...
+bazel_dep(name = "abseil-cpp", repo_name = "com_google_absl", version = "20230802.0.bcr.1")
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "rules_java", version = "4.0.0")
+bazel_dep(name = "rules_jvm_external", version = "5.1")
+bazel_dep(name = "rules_pkg", version = "0.7.0")
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "rules_python", version = "0.10.2")
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "zlib", version = "1.2.11")
```
mkruskal-google, line 22: These seem out of date.  Could we bump these up based on the current state of protobuf_deps.bzl?  I'd expect that the examples directory wouldn't need to override any deps

MODULE.bazel
alexeagle: Sadly you cannot have load statements in MODULE.bazel so we can't reuse the constant. I copied the value.

Note in most rulesets we just leave the version in code as a placeholder, to avoid having to create "version bump commits" - I can show you more about it if you're curious.

mkruskal-google: Yea I'd be curious about possible better solutions, but we have *many* places we need to bump every release.  Our tooling handles this automatically

mkruskal-google (approved):

alexeagle:
@mkruskal-google anything in the TAP results that I could address from external so you don't have to become the "google3 ghost author" of my change?

mkruskal-google:
Nope, we only test these files in GH.  Just waiting on a second reviewer

MODULE.bazel
```diff
@@ -1,2 +1,23 @@
 # TODO: migrate all dependencies from WORKSPACE to MODULE.bazel
 # https://github.com/protocolbuffers/protobuf/issues/14313
+module(
+    name = "protobuf",
+    compatibility_level = 1,
+    version = "27.0",
+)
+
+# LOWER BOUND dependency versions.
```
zhangskz, line 9: Is it possible to set upper bounds too or pin at a specific version if necessary?

PR #16378, opened 2024-04-02
docs: add protoc-gen-openapi custom options to registry

shashankram:
I am introducing custom field options, and subsequently message options as extensions to a new options.proto file. I would like to reserve extension number for the same.


shashankram:
@acozzette do you review these type of changes?

acozzette (approved):

shashankram:
@acozzette I might need to add more options, will revisit in a separate PR.

PR #16379, opened 2024-04-02
Add bigdecimal to dependencies

JasonLunn:
In Ruby 3.4, `bigdecimal` is no longer built in. See announcement in [Ruby 3.3 release notes](https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/#:~:text=base64-,bigdecimal,-csv).

### DIFF
--- a/ruby/google-protobuf.gemspec
+++ b/ruby/google-protobuf.gemspec
@@ -32,6 +32,8 @@ Gem::Specification.new do |s|
     s.add_development_dependency "rake-compiler-dock", "= 1.2.1"
   end
   s.required_ruby_version = '>= 3.0'
+  # bigdecimal must be used as a non-built in gem as of ruby-3.4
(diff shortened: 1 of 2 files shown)


zhangskz (approved):
Will this be disruptive for users? We probably want to make sure to include this in the release notes at least

JasonLunn:
It should not be disruptive for users, but I agree it deserves to make it to the release notes. https://rubygems.org/gems/bigdecimal is a mature gem with tens of millions of installs and is something we already have a dependency on, we're just used to getting said dependency from the ruby installation itself, rather than having to explicitly treat it as a gem. It has a platform specific implementation for JRuby but is otherwise a pure-Ruby gem that is architecture independent. This is part of a larger effort on the part of the Ruby language maintainers to push more of the current standard library into optional gems.

PR #16381, opened 2024-04-02
Update JRuby from 9.4.3.0 to 9.4.6.0.

JasonLunn:


zhangskz:
Is this just to bump what we are testing against, or is there something specific motivating this?

JasonLunn:
Mostly just to stay up to date with what we're testing against... I'm working separately on some Ruby ASAN stuff that's leading me toward bumping up the highest version we test on under CRuby, so I thought I'd bump this while I was in the area.

zhangskz (approved):

PR #16454, opened 2024-04-09
Enable Unity(Jumbo) builds

jslap-ubi:
**Description**
Added a cmake option to enable Unity builds for libprotobuf-lite.
Added missing symmetric include for port_def.inc
Added preproc guards for PROTOBUF_PRAGMA_INIT_SEG so it does not get included multiple times in unity build.

**Motivation and Context**
For large code base, especially in the game dev industry, Unity builds (https://en.wikipedia.org/wiki/Unity_build) are used to speed up the compilations.
Special care must be taken to enable Unity builds for a project, since many translation units will be merged togheter.


jslap-ubi:
Hello,

I just want to make sure my PR does not fall into the cracks.

CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ option(protobuf_BUILD_LIBUPB "Build libupb" ON)
 option(protobuf_DISABLE_RTTI "Remove runtime type information in the binaries" OFF)
 option(protobuf_TEST_XML_OUTDIR "Output directory for XML logs from tests." "")
 option(protobuf_ALLOW_CCACHE "Adjust build flags to allow for ccache support." OFF)
+option(protobuf_USE_UNITY_BUILD "Enable Unity (Jumbo) build for" ON)
```
zhangskz, line 38: Our .NET support policy states that we explicitly don't support Unity (https://opensource.google/documentation/policies/dotnet-support) and thus won't have tests, etc for these environments. We may be able to accept a contribution here anyways, but note that support would be best-effort / community-contributed, and this should at least be noted in a comment here accordingly.

jslap-ubi: Sorry, I think there is a confusion. I'm not talking about building for the Unity game engine.

This is about having unity-build: [Unity-Jumbo Build](https://en.wikipedia.org/wiki/Unity_build) A build in which all the cpp files are merged together into a big translation unit.

zhangskz: Ah that makes more sense and seems more feasible then -- thanks for clarifying!

jslap-ubi:
Bump. Just to make sure we don't forget about this one.

mkruskal-google:
Sorry for the delays here.  I think this could be reasonable to accept as best-effort support (especially since most of these changes are unobjectionable cleanups), but given the changes you had to make to get it to work I'm worried we'll quickly break it again.  What kind of support do you expect for this?

jslap-ubi:
Did you had the oportunity to have a look at this?

mkruskal-google:
Hey, so we discussed this a bit and we can offer best-effort support here (i.e. we'll accept small PRs but not actively test that unity builds work).  All of the necessary changes look beneficial, and I can add some checks to try to limit backslide on port_def/port_undef mismatches on our end.

mkruskal-google (approved):

Issue #7392, opened 2020-04-16
Need specification for how to parse unrecognized enums in JSON


srcreigh:
I was forwarded to here by @tbkka over in the Swift protobuf library from this issue: https://github.com/apple/swift-protobuf/issues/972

As it stands the JSON serialization option in the swift-protobuf library is unusable because when it sees an unrecognized enum value that has been serialized to a JSON string, it throws an exception. There is no option to ignore the unrecognized string enum value, the only behaviour that swift-protobuf offers is to throw an exception.

To me it is abundantly obvious that any protobuf library has to ignore information that was generated using a newer schema.

For example, if a new enum field is shipped on a proto with v12.5 of an iOS app but is totally ignored (maybe it's a new feature under a shared schema that is implemented on a web app and is not implemented on the mobile app for a while). Then when this field, which is totally ignored, gets a new enum value, the v12.5 version of the app breaks because the only thing this that swift-protobuf can do is throw an exception when it sees the unrecognized enum value.

Of course this scenario is the core problem that protocol buffers as a whole set out to solve.

> Protocol buffers were designed to solve many of these problems:
> - New fields could be easily introduced, and intermediate servers that didn't need to inspect the data could simply parse it and pass through the data without needing to know about all the fields.
> https://developers.google.com/protocol-buffers/docs/overview#a-bit-of-history

(Of course an iOS app is not a server but it's the same thing as far as this is concerned because it participates in a distributed protocol.)

**Could someone please provide a specification for what a protobuf library should do when it sees an unrecognized Enum value serialized in a JSON string?**

I can't imagine that every other protobuf client that supports JSON serialized string enum values here crashes when it sees something unrecognized. Can someone point me to the behaviour of some of the other client libraries so that the folks over at swift-protobuf can have some precedence for how to handle this situation?

Thank you.

acozzette:
I'm not sure about other languages, but at least for C++ you can set an option to ignore unknown fields during JSON parsing, and this includes known fields with unknown enum values. The documentation [here](https://developers.google.com/protocol-buffers/docs/proto3#json) says: "Proto3 JSON parser should reject unknown fields by default but may provide an option to ignore unknown fields in parsing."

tbkka:
Could you please update the spec to explicitly call out enums as falling into this case?  It would also help if you could clarify expectations here for repeated enum fields.

And of course, expanding the conformance tester to verify these cases would be a nice treat!

thomasvl:
[...]

PR #16479, opened 2024-04-11
[7392] [cpp] Fix IgnoreUnknownEnumStringValueInMap conformance tests

antongrbin:
## Motivation

This PR fixes failing JSON conformance tests for cpp with name `IgnoreUnknownEnumStringValueInMap`. Specifically, for the following input of type [TestMapOfEnums](https://github.com/protocolbuffers/protobuf/blob/2f3242c576504459621fd7d78bbccf39bfaa49c5/src/google/protobuf/util/json_format.proto#L119):

```json
{
  "enum_map": {
    "key1": "PROTOCOL",
    "key2": "UNKNOWN_ENUM_STRING_VALUE",
    "key3": "BUFFER"
  }
}
```

The JSON parser with `ignore_unknown_fields=true` should produce a map with 2 entries, skipping the unknown enum value under `key2`. The implementation before this PR outputs `key2` with 0-th enum value (`PROTOCOL`) instead.

The JSON parsing spec was discussed in https://github.com/protocolbuffers/protobuf/issues/7392.

Recent similar changes in other languages:
- Python: https://github.com/protocolbuffers/protobuf/commit/86abf35ef5ee5b1004ec11bebb36d84c2ef6645e
- Swift: https://github.com/apple/swift-protobuf/pull/1345
- C#: https://github.com/protocolbuffers/protobuf/pull/15758

## Changes

The main difficulty in implementing the fix is the fact that parser will first allocate a new map entry message and then consume the value. In this special case (unknown enum string value), we need to do it the other way around: first consume the value, then allocate the map entry if needed.

The PR initially contains 5 commits, with first 4 being noop unit tests and refactors:

1. unit test that documents the mainline behavior (this test will be changed in the 5th commit)
2. noop refactoring that simplifies the signature of ParseEnumFromStr
3. noop refactoring that extract map key parsing into a separate function (we'll call it twice in 5th commit)
4. noop refactoring that extracts parse map entry into a separate function (since it will get more logic)
5. the commit that actually implements the fix and adjusts tests

### DIFF
--- a/src/google/protobuf/json/internal/parser.cc
+++ b/src/google/protobuf/json/internal/parser.cc
@@ -691,12 +691,105 @@ absl::Status ParseMapKey(const Desc<Traits>& type, Msg<Traits>& entry,
   return absl::OkStatus();
 }
 
+// Parses map entry when value is an enum name (string) and ignoring unknown
+// fields flag is set.
+// Specification says that in this case, if the enum name is unknown, we should
+// skip the map entry.
+// For example {"key1": "UNKNOWN_ENUM_VALUE"} should parse into an empty map if
+// the ignore_unknown_fields is set.
+//
+// The default implementation in ParseMapEntry will first allocate a new map
+// entry message (by calling Traits::NewMsg) and then consume the value from
+// the lexer.
+//
+// In this special case, we must do it the other way around:
+// - first consume the value from the lexer,
+// - check that this is a known enum value,
+// - and only then proceed with allocating a new map entry message.
+//
+// Note that in both cases the key is already consumed and passed in through
+// 'LocationWith<MaybeOwnedString>& key' param.
+template <typename Traits>
+absl::Status ParseMapEntryWithEnumStringValueWhenIgnoringUnknownFields(
+    JsonLexer& lex, Field<Traits> map_field, Msg<Traits>& parent_msg,
+    LocationWith<MaybeOwnedString>& key) {
+  // Parse the enum value from string, advancing the lexer.
+  absl::StatusOr<absl::optional<int32_t>> enum_value;
+  (void)Traits::WithFieldType(
+      map_field, [&lex, &enum_value](const Desc<Traits>& map_entry_desc) {
+        enum_value = ParseEnum<Traits>(lex, Traits::ValueField(map_entry_desc));
+        return absl::OkStatus();
+      });
+  RETURN_IF_ERROR(enum_value.status());
+
+  // If enum value is unknown, we'll stop here and avoid allocating a new map
+  // entry message.
+  if (!enum_value->has_value()) {
+    return absl::OkStatus();
+  }
+
+  // If the enum value is known, output the map entry with already parsed
+  // value.
+  return Traits::NewMsg(
+      map_field, parent_msg,
+      [&](const Desc<Traits>& type, Msg<Traits>& entry) -> absl::Status {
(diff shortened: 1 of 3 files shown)


antongrbin:
@haberman friendly ping to take a look :)

antongrbin:
@sbenzaquen can you help me ping @haberman through internal channels :) I bet this is a GitHub notifications recall miss.

antongrbin:
@jskeet, since you worked with me on the linked issue https://github.com/protocolbuffers/protobuf/issues/7392, can you help me get reviewer's attention here through internal channels? I suspect that they are not seeing GitHub notifications.

I am not blocked on this, just wanted to eventually remove the failing tests that I added.

jskeet:
I've pinged internally.

googleberg:
@esrauchg Can you take a look at this?

src/google/protobuf/json/internal/parser.cc
```diff
@@ -625,6 +625,180 @@ absl::Status ParseArray(JsonLexer& lex, Field<Traits> field, Msg<Traits>& msg) {
...
+  return Traits::NewMsg(
+      map_field, parent_msg,
+      [&](const Desc<Traits>& type, Msg<Traits>& entry) -> absl::Status {
+        RETURN_IF_ERROR(ParseMapKey<Traits>(lex.options(), type, entry, key));
+
+        // This is different from the default implementation in 'ParseMapEntry'
+        // since we already parsed this value (we must not attempt to advance
+        // the lexer to parse the value here again).
+        Traits::SetEnum(Traits::ValueField(type), entry,
+                        enum_value->value_or(0));
```
esrauchg, line 742: Nit: this should just be `*enum_value` (or `enum_value->value()`), you already checked that has_value is true above so it will be guaranteed to succeed. 

It's academic in this case because the 0 value would never be used, but under proto2 semantics (where enums are closed, which is the case you are handling here) the default value for an enum is whatever is listed first in the definition, which isn't necessarily 0 value for the enum; there could either be a 0 value that is not listed first (then 0 is a legal value but not the default) or no listed 0 value at all (then its not a legal value at all).

Only under proto3 / open enums is it mandatory that the enum has a 0 value and that it is listed first (exactly to allow every enum to have a 0 default in proto3).

src/google/protobuf/json/internal/parser.cc
```diff
@@ -625,6 +625,180 @@ absl::Status ParseArray(JsonLexer& lex, Field<Traits> field, Msg<Traits>& msg) {
...
+// 'key'.
+template <typename Traits>
+absl::Status ParseMapEntry(JsonLexer& lex, Field<Traits> map_field,
+                           Msg<Traits>& parent_msg,
+                           LocationWith<MaybeOwnedString>& key) {
+  auto needs_special_case =
+      MapEntryNeedsEnumStringValueSpecialCase<Traits>(lex, map_field);
+  RETURN_IF_ERROR(needs_special_case.status());
+  if (*needs_special_case) {
+    return ParseMapEntryWithEnumStringValueWhenIgnoringUnknownFields<Traits>(
```
esrauchg, line 789: Can we structure this to be less drastically special cased? I'd rather it 'infect' the happy case if that avoids such a large diversion to handle the edge case.

Is it possible to structure this to always parse map key, parse map value, decide whether to err or ok if the value can't be parsed due to it being unknown enum, then construct the message?

esrauchg (changes requested):
Thanks for the contribution! This is a great start, just one structuring comment and a few nits.

PR #16489, opened 2024-04-12
Updated Resource bundle path.

pavan-chingari:


haberman:
This PR looks like it may have been targeted to the wrong branch. It originally had a huge delta, something like -250k lines, +113k lines.

I tried changing the target branch to 22.x to match the v3.22.5 branch name, which reduced the size of the diff, but it still appears to touch far more files than are intended based on the PR description.

I'm going to close this for now; please feel free to submit a new PR that has the appropriate base.

PR #16491, opened 2024-04-12
Fix the bug of DefaultValueObjectWriter class

HUSTzhangbo:
Fix the bug when the DefaultValueObjectWriter class handles the element whose key is an empty string in the map.


haberman:
This PR is attempting to patch the 3.5.x branch. Protobuf 3.5.x is over six years old, and is far outside our support window.

If this bug is still occurring in currently supported versions of protobuf, feel free to send another PR.

PR #16495, opened 2024-04-12
[bzlmod] Set repo_name to com_google_protobuf

keith:
This is currently required by https://github.com/protocolbuffers/protobuf/blob/8d025c08ff9c4e63f37f62b64f3046183ef7ab20/upb_generator/bootstrap_compiler.bzl#L30

We could try to change it but this is fine to do for now

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,18 +1,19 @@
 # TODO: migrate all dependencies from WORKSPACE to MODULE.bazel
 # https://github.com/protocolbuffers/protobuf/issues/14313
-PROTOBUF_VERSION  = "27.0"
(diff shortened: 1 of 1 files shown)


keith:
this also fixes the buildifier formatting issues

zhangskz (approved):

keith:
@zhangskz looks like something failed on the import, anything I can do to fix it?

zhangskz:
Thanks for pointing this out! Looks like we had an unrelated issue that prevented this from importing correctly which should be unblocked shortly. Feel free to re-ping if you don't see this submitted within the next ~day or so.

zhangskz (approved):

PR #16525, opened 2024-04-16
Don't use return statement for void-returning function.

phst:
With Visual C++, this otherwise generates lots of warnings of type C4098.


ericsalo (approved):

phst:
I think this is still current.

phst:
Never mind, looks like commit 5fb5b871dc9e9d9119374121b7cb4576a057470b does the same.

PR #16592, opened 2024-04-22
enable edition flags for php

bshaffer:


bshaffer:
I received a list of errors here: https://gist.github.com/bshaffer/128662df8ef3e6d4d6bec33f81899498

PR #16655, opened 2024-04-26
Allow builders to return default instance in Java

yuzawa-san:
rationale: I was investigating allocations during parsing where a lot of empty messages were present and realized that each of these empty messages was a newly allocated instance, when it would be more memory efficient to reuse the default instance each time.

solution: add a default flag to the builder which is invalidated onChange. then conditionally return the default instance if that flag is true.

note: I had to bring back some logic removed in c0d08bdcade9d44698015b85c9c23343ca1319b7 in order for the schema flow to work correctly since it was using toBuilder and expecting a new instance, not the default instance.


googleberg:
@yuzawa-san Unfortunately, while we like the idea, we can't absorb this change. We've actually attempted this internally and there are currently a bunch of tests that rely on `.newBuilder().build()` returning unique instances. We'd need to address those first.

PR #16712, opened 2024-05-02
feat: enable edition for php

bshaffer:
Expose editions for PHP

 - Skips `TextFormatOutput` tests, as text payload is not supported for PHP
 - Skips `TestAllTypesEdition2023` tests, as editions-specific features are not yet supported for PHP
 - Skips `TestAllTypesProto2` tests for Editions, as proto2 is not supported in PHP

Additionally, refactors `conformance_php.php` by using `switch` statements for readability, and adding consistent whitespace formatting.


BUILD.bazel
```diff
@@ -615,6 +616,37 @@ internal_php_proto_library(
         "//php:__subpackages__",
     ],
 )
+internal_copy_files(
```
mkruskal-google, line 619: I think this might be unnecessary now.  In the ruby/python case we also needed to copy the WKT, which wasn't required here

src/google/protobuf/compiler/php/php_generator.h
```diff
@@ -37,7 +37,13 @@ class PROTOC_EXPORT Generator : public CodeGenerator {
                    std::string* error) const override;
 
   uint64_t GetSupportedFeatures() const override {
-    return FEATURE_PROTO3_OPTIONAL;
+    return Feature::FEATURE_PROTO3_OPTIONAL | Feature::FEATURE_SUPPORTS_EDITIONS;
```
mkruskal-google, line 40: We probably should hold off on advertising support until it's ready.  Any protos under google/protobuf are allowlisted, and we could tweak php_proto_library to pass --experimental editions if that doesn't work

conformance/conformance_php.php
```diff
@@ -27,57 +36,89 @@
...
       } catch (Exception $e) {
           $response->setParseError($e->getMessage());
           return $response;
       }
-	} elseif ($request->getPayload() == "text_payload") {
-		$response->setSkipped("PHP doesn't support text format yet");
-        return $response;
-	} else {
+      break;
+	  case "text_payload":
```
mkruskal-google, line 99: whitespace here is still off

conformance/conformance_php.php
```diff
@@ -27,57 +36,89 @@
...
-    if ($request->getRequestedOutputFormat() == WireFormat::UNSPECIFIED) {
+  switch ($request->getRequestedOutputFormat()) {
+    case WireFormat::UNSPECIFIED:
       trigger_error("Unspecified output format.", E_USER_ERROR);
-    } elseif ($request->getRequestedOutputFormat() == WireFormat::PROTOBUF) {
+    case WireFormat::PROTOBUF:
       $response->setProtobufPayload($test_message->serializeToString());
-    } elseif ($request->getRequestedOutputFormat() == WireFormat::JSON) {
+      break;
+    case WireFormat::JSON:
```
mkruskal-google, line 112: Are you missing a a TEXT case?

bshaffer: Added explicit ignoring of text output cases in [6ccd522](https://github.com/protocolbuffers/protobuf/pull/16712/commits/6ccd522c16415aa55d6b094582b95c0866854235)

conformance/conformance_php.php
bshaffer: I moved everything to 4 spaces and made a few other adjustments to bring this file up to community code standards for PHP in [b53e15a](https://github.com/protocolbuffers/protobuf/pull/16712/commits/b53e15aa00b908afc0d1d1016a674e4642bc94f2)

src/google/protobuf/compiler/php/php_generator.h
bshaffer: Maybe I'm doing something wrong, but when I remove `FEATURE_SUPPORT_EDITIONS`, I get the following error:

```
test_messages_proto3_editions.proto: is an editions file, but code generator --php_out hasn't been updated to support editions yet.  Please ask the owner of this code generator to add support or switch back to proto2/proto3.
```

BUILD.bazel
bshaffer: I cleaned up the bazel files, and I believe it's much cleaner / better than before: e039fa0335bfdb4b4c91d1d9282fb7b5715b6cca

src/google/protobuf/compiler/php/php_generator.h
mkruskal-google: protoc will skip this check for any proto path under "google/protobuf/" and "upb/".  We should probably add "editions." to [that list](https://github.com/protocolbuffers/protobuf/blob/d3b2fc52a4aec2dc9711c8235cbcdb71d4c10295/src/google/protobuf/compiler/code_generator.cc#L147) too

mkruskal-google: Alternatively, you could modify `internal_php_proto_library` to accept an experimental_editions argument and have it set `--experimental_editions` on protoc

bshaffer: I looked into both of these and, to do the first, `editions/` does not bypass the check because the filename being passed in is `test_messages_proto3_editions.proto` (the `editions/` path is not part of the filename, I am not sure why). As for the second, I cannot find a way to set custom flags on the `protoc` binary, so it seems like I'd have to modify `_proto_gen`, which I am not super excited about doing.

protobuf.bzl
```diff
@@ -655,6 +660,7 @@ def _source_proto_library(
         protoc = Label("//:protoc"),
         testonly = None,
         visibility = ["//visibility:public"],
+        enable_editions = False,
```
mkruskal-google, line 663: did you mean to add this to the php macro?

src/google/protobuf/compiler/php/php_generator.h
bshaffer: Okay! Using the same flags added by Objective C in 6c7c5a5d2a6befdc4117242671612146efa50679, I removed the editions flag from PHP and added the `experimental_editions` flag for `protoc` (3b212ea2c0378a9f3ceace19d8acae7c5cf9898e)

protobuf.bzl
bshaffer: Without this, I get `name 'enable_editions' is not defined`. Is there a better way?

mkruskal-google: whoops nvm

mkruskal-google (approved):

haberman:
It looks like these changes were submitted in https://github.com/protocolbuffers/protobuf/commit/01744cccae57ca934ab925e13d2eaaf93761b1f4#diff-5ebd764f133e72efb1415d478675b92c7b091b3bf007045c2256aa8742267c75, so I'm closing this PR now.

Issue #16745, opened 2024-05-05
Span<T> in C# on RepeatedField<T>


PascalSenn:
**What language does this apply to?**
c# protobuf 3

**Describe the problem you are trying to solve.**

The issue #6217 was recently marked as stale, but the problem it addresses is still relevant. The goal is to work with a `﻿Span<T>` of a ﻿`RepeatedField<T>`  in C# Protobuf, which would allow for working with apis like vectors and reuse of already allocated memory. Accessing the internal array of a ﻿RepeatedField<T> would make this possible.

I have read through the discussions on the issue and the PR https://github.com/protocolbuffers/protobuf/pull/6538. 

**Describe the solution you'd like**
Ideally, provide access to a ﻿`ReadOnlySpan<T>` of the internal array of a `﻿RepeatedField<T>`. I understand that modifying the array after creating a span from it could lead to unexpected behavior, but in most cases, incoming messages are not mutated.

**Describe alternatives you've considered**
If providing access to the internal data structures through ReadOnlySpan<T> is not feasible, an alternative solution would be to introduce ﻿`CopyTo `methods that accept a ﻿`Span<T>` as the destination. The proposed method signatures are:
```csharp
    public void CopyTo(Span<T> destination) {..}
    public void CopyTo(Span<T> destination, int length) {..}
    public void CopyTo(Span<T> destination, int sourceIndex, int destinationIndex, int length) {..}
```

**Additional context**

jskeet:
I'd want the input of @jamesnk on this (for consistency with other .NET APIs), but I don't *think* I object to exposing a `ReadOnlySpan<T>`. Can you create a PR for us to review?

PascalSenn:
@jskeet  The objection on `ReadOnlySpan<T>` in the initial proposal was about the possibility of resizing the underlying array. 

https://github.com/protocolbuffers/protobuf/pull/6538#issuecomment-630114989

> Also note that exposing the data as read-only (ReadOnlySpan<T>) addresses some concerns over the safety, but there are still some potential issues - e.g. the returned span is bound to the current array, but if a repeated field gets resized, a new internal array is allocated, so the span will point to a dead array and all the changes will be lost.

If exposing `ReadOnlySpan<T>` is an option. Something simple as this could work:
https://github.com/protocolbuffers/protobuf/pull/16772

If you prefer a `CopyTo` approach, let me know.

jskeet:
[...]

PR #16772, opened 2024-05-07
Adds UnsafeCollectionOperations for unsafe access to RepeatedField<T>

PascalSenn:
This is a proposal to add `UnsafeCollectionOperations `  for fast access on  `RepeatedField<T>`
  https://github.com/protocolbuffers/protobuf/issues/16745


jskeet:
@JamesNK We're aware that if the repeated field is modified while the span is "active", it may or may not reflect the changes (e.g. add/remove operations will not be seen in terms of the length, and if the array is reallocated any further updates won't be visible at all) - but do you think that's reasonable behavior, so long as we document that appropriately? (I'd just say that the state of the span is undefined after any modifications.) Any other thoughts about this PR?

JamesNK:
Modification is the problem. Sometimes the span reflects changes, sometimes it doesn't.

I like the idea, but I don't think it should be a property on `RepeatedField<T>`. It makes the dangerous behavior too easy to access. I suggest doing what the BCL does with `List<T>` + span and make it a static method on an "advanced" type.

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.asspan?view=net-8.0

PascalSenn:
@JamesNK 

I changed the implementation to include your suggestions. There is already a similar class `UnsafeByteOperations`, so I aligned the naming with it -> `UnsafeCollectionOperations`. The implementation of the class is very close to the [CollectionsMarshal.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CollectionsMarshal.cs) from `System.Runtime.InteropServices`. I implemented `SetCount` and `AsSpan`. I also ported the tests over from [CollectionsMarshalTests.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/CollectionsMarshalTests.cs).

csharp/src/Google.Protobuf.Test/Collections/UnsafeCollectionOperationsTest.cs
```diff
@@ -0,0 +1,210 @@
+using System;
```
jskeet, line 1: Please add a copyright header (as per the other files) here.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -611,6 +600,17 @@ public override string ToString()
             }
         }
 
+        // May increase the size of the internal array, but will never shrink it.
```
jskeet, line 603: Any reason this has been moved rather than just changed from private to internal?

csharp/src/Google.Protobuf.Test/Collections/UnsafeCollectionOperationsTest.cs
```diff
@@ -0,0 +1,210 @@
...
+    public void NullFieldAsSpanClass()
+    {
+        RepeatedField<object> field = null;
+        var span = UnsafeCollectionOperations.AsSpan(field);
+
+        Assert.AreEqual(0, span.Length);
+    }
+
+    [Test]
+    public void AsSpanOnConcurrentFieldAcccessThrows()
```
jskeet, line 35: Acccess => Access

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -34,8 +34,8 @@ public sealed class RepeatedField<T> : IList<T>, IList, IDeepCloneable<RepeatedF
         private static readonly T[] EmptyArray = new T[0];
         private const int MinArraySize = 8;
 
-        private T[] array = EmptyArray;
-        private int count = 0;
+        internal T[] array = EmptyArray;
```
jskeet, line 37: Please don't make fields internal. I was already nervous about having the SetCount method, but I *definitely* don't want internal access to fields.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,145 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Span<T> AsSpan<T>(RepeatedField<T> field)
+        {
+            Span<T> span = default;
+            if (field is not null)
+            {
+                var size = field.count;
+                var items = field.array;
+
+                Debug.Assert(items is not null, "Implementation depends on RepeatedField<T> always having an array.");
```
jskeet, line 85: Please use a normal if/throw rather than Debug.Assert. (We don't use Debug.Assert in any of the production code in protobuf.)

However, I don't think we actually need to check this anyway. I don't think the array can ever be null, and if it ever *were* null it would cause other issues.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
jskeet: Instead, I'd recommend adding internal SetCount and AsSpan methods, and calling those from the public methods in UnsafeCollectionOperations.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// The element type of the <see cref="RepeatedField{T}"/>.
+        /// </typeparam>
+        /// <param name="field">
+        /// The <see cref="RepeatedField{T}"/> to wrap.
+        /// </param>
+        /// <returns>
+        /// A <see cref="Span{T}"/> that wraps the backing array of the
+        /// <see cref="RepeatedField{T}"/>.
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jskeet, line 76: I don't think we need aggressive inlining for this.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <exception cref="NullReferenceException">
+        /// <paramref name="field"/> is <see langword="null"/>.
+        /// </exception>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// <paramref name="count"/> is negative.
+        /// </exception>
+        /// <remarks>
+        /// When increasing the count, uninitialized data is being exposed.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jskeet, line 117: I don't think we need aggressive inlining for this.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
jskeet: (With my suggested approach of a new internal SetCount method, this could stay unchanged.)

jskeet (commented):
I haven't reviewed the tests in detail as I'm suggesting quite significant implementation changes.

csharp/src/Google.Protobuf.Test/Collections/UnsafeCollectionOperationsTest.cs
```diff
@@ -0,0 +1,210 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using NUnit.Framework;
+
+namespace Google.Protobuf.Collections;
+
+public class UnsafeCollectionOperationsTest
+{
+    [Test]
+    public unsafe void NullFieldAsSpanValueType()
```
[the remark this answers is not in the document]
PascalSenn: I don't mind either change. The reasoning behind this is to follow the same patterns as the `CollectionsMarshal`.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
PascalSenn: Yes, I understand. I follow the same patterns as in

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L25-L27

It appears to be a performance optimization in the `List<T>` class to expose these fields. 

I can revert this change.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
PascalSenn: I moved it down as it appeared to be in the wrong place for an internal method. However, after reviewing the code again, I'm not sure if there even is any specific "ordering by visibility" of members in this class.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
PascalSenn: I believe a performance improvement from `CollectionsMarshal.cs`.
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CollectionsMarshal.cs#L21

I removed it

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
PascalSenn: resolved

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,145 @@
...
+                Debug.Assert(items is not null, "Implementation depends on RepeatedField<T> always having an array.");
+
+                if ((uint) size > (uint) items.Length)
+                {
+                    // RepeatedField<T> was erroneously mutated concurrently with this call, leading
+                    // to a count larger than its array.
+                    throw new InvalidOperationException("Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.");
+                }
+
+                Debug.Assert(typeof(T[]) == field.array.GetType(), "Implementation depends on RepeatedField<T> always using a T[] and not U[] where U : T.");
```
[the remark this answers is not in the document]
PascalSenn: I will remove the check as `RepeatedField<T>` is sealed anyway

csharp/src/Google.Protobuf.Test/Collections/UnsafeCollectionOperationsTest.cs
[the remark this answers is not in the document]
PascalSenn: I think it's more consistent to throw in case of null, as `SetCount` does the same. 
I changed it to this behaviour.

PascalSenn:
@jskeet 
Most of this implementation closely aligned with `CollectionsMarshal.cs` from the runtime. The goal was to keep the implementation and behavior relatively consistent.

I also believe a few things were performance optimizations in the `CollectionsMarshal`. For example, the aggressive inlining and the direct field access.

I incorporated your feedback into this pull request to better align with the behavior of Protobuf. I left a couple of comments open to provide some context. Feel free to close them as needed after review, as everything should be resolved.

PascalSenn:
@jskeet One last remark. The tests are ported from `CollectionsMarshalTests.cs`   to ensure consistent behavior:
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/CollectionsMarshalTests.cs

jskeet:
I'll have a look tomorrow. I would say that I'm not as interested in alignment with CollectionsMarshal as with consistency with the rest of *this* codebase, and clarity of code and usage.

csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;net6.0</TargetFrameworks>
```
JamesNK, line 4: How come net462 is removed? I think it's valuable to run tests there. If some features aren't supported on .NET Framework then they can be selectively `#ifdef`ed out.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,118 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Security;
+using Google.Protobuf.Collections;
+
+namespace Google.Protobuf
+{
+    /// <summary>
```
JamesNK, line 20: I think the info here should be merged with the comment on `AsSpan`.

The overall comment for the type could be similar to CollectionMarshal: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal?view=net-8.0

> An unsafe class that provides a set of methods to access the underlying data representations of collections.

The reason to simplify it is because in the further there could be methods that are specific to `MapField`, and then the current class comment wouldn't make sense.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// cref="RepeatedField{T}"/>.
+    /// </description>
+    /// </item>
+    /// </list>
+    /// </remarks>
+    [SecuritySafeCritical]
+    public static class UnsafeCollectionOperations
+    {
+        /// <summary>
+        /// Returns a <see cref="Span{T}"/> that wraps the backing array of the specified
```
JamesNK, line 62: The comment should mention:

Values in the `Span<T>` shouldn't be set to `null`. Items should be removed from a RepeatedField with Remove or RemoveAt.

JamesNK (commented):
I think its good! Some minor comment and test improvements in feedback

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,118 @@
...
+
+#endregion
+
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Security;
+using Google.Protobuf.Collections;
+
+namespace Google.Protobuf
```
jskeet, line 18: Let's use a file-scoped namespace for simplicity.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
jskeet: +1

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
jskeet, line 62: Possibly change "the backing array" to "the current backing array" to emphasize that it can change?

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// <typeparam name="T">The type of the elements in the field.</typeparam>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="field"/> is <see langword="null"/>.
+        /// </exception>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// <paramref name="count"/> is negative.
+        /// </exception>
+        public static void SetCount<T>(RepeatedField<T> field, int count)
+        {
+            if (field is null)
```
jskeet, line 105: Use `ProtoPreconditions.CheckNotNull(field, nameof(field))` to simplify this and the check above.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                throw new ArgumentNullException(nameof(field));
+            }
+
+            if (count < 0)
```
jskeet, line 110: I'd expect this test to be in RepeatedField.SetCount instead. Basically this method can just be a proxy.

jskeet (commented):
Down to just nits now - it feels like we're very close, thanks very much for this!

jskeet:
cc @amanda-tarafa for any additional thoughts

csharp/src/Google.Protobuf.Test/Google.Protobuf.Test.csproj
PascalSenn: Oops. I removed it to be able to run the test on my unix machine

amanda-tarafa (commented):
A couple of comments, but looks good otherwise.

jskeet:
Thanks - I'm afraid at this point I won't get to this until Monday, but hopefully I'll be able to look then.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// Returns a <see cref="Span{T}"/> that wraps the current backing array of the given
+    /// <see cref="RepeatedField{T}"/>.
+    /// </para>
+    /// <para>
+    /// Values in the <see cref="Span{T}"/> should not be set to null. Use
+    /// <see cref="RepeatedField{T}.Remove(T)"/> or <see cref="RepeatedField{T}.RemoveAt(int)"/> to
+    /// remove items instead.
+    /// </para>
+    /// <para>
+    /// The returned <see cref="Span{T}"/> is only valid until the <see cref="RepeatedField{T}"/> is
```
jskeet, line 36: This isn't quite right - it's fine for the field to be modified so long as the size doesn't change. (So modifying existing elements preserves the link.)

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// If count is less than <see cref="RepeatedField{T}.Count"/> the collection is effectively
+    /// trimmed down to the first count elements. <see cref="RepeatedField{T}.Capacity"/>
+    /// is unchanged, meaning the underlying array remains allocated.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of elements in the <see cref="RepeatedField{T}"/>.
+    /// </typeparam>
+    /// <param name="field">The field to set the count of.</param>
+    /// <param name="count">The value to set the field's count to.</param>
```
jskeet, line 78: Specify "Must be non-negative."

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <para>
+    /// If count is less than <see cref="RepeatedField{T}.Count"/> the collection is effectively
+    /// trimmed down to the first count elements. <see cref="RepeatedField{T}.Capacity"/>
+    /// is unchanged, meaning the underlying array remains allocated.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of elements in the <see cref="RepeatedField{T}"/>.
+    /// </typeparam>
+    /// <param name="field">The field to set the count of.</param>
```
jskeet, line 77: Specify "Must not be null."

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <para>
+    /// The returned <see cref="Span{T}"/> is only valid until the <see cref="RepeatedField{T}"/> is
+    /// modified, after which its state becomes undefined.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of elements in the <see cref="RepeatedField{T}"/>.
+    /// </typeparam>
+    /// <param name="field">
+    /// The <see cref="RepeatedField{T}"/> for which to wrap the current backing array.
```
jskeet, line 44: Specify "Must not be null."

jskeet: (We haven't historically been terribly good at this in the proto library, but I'd like to try to make sure that at least new code is clearer...)

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// is unchanged, meaning the underlying array remains allocated.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of elements in the <see cref="RepeatedField{T}"/>.
+    /// </typeparam>
+    /// <param name="field">The field to set the count of.</param>
+    /// <param name="count">The value to set the field's count to.</param>
+    /// <exception cref="ArgumentNullException">
+    /// Thrown if <paramref name="field"/> is <see langword="null"/>.
```
jskeet, line 80: Specify ArgumentOutOfRangeException as well? (If we're doing one, we might as well do both.)

jskeet (commented):
Just down to comment nits from my side.

csharp/src/Google.Protobuf/UnsafeCollectionOperations.cs
PascalSenn: I added the comment, though the precondition check was moved to `SetCount` on `RepeatedField`.

PascalSenn:
@jskeet  Comments are resolved 🚀

jskeet (approved):

jskeet:
@JamesNK: Could you take one final pass at this? It looks good to me, but due to the unsafe aspects of it, I'd appreciate a double-check. After that, I can get it reviewed by a member of the protobuf team and merged.

JamesNK (approved):

jskeet:
@mkruskal-google: We've reviewed this from a .NET perspective. Let me know if you'd like any more background. (But basically this allows very efficient access to a repeated field, with caveats around anything that might resize it.)

mkruskal-google:
This lgtm, @PascalSenn we need you to sign the CLA though

PascalSenn:
@mkruskal-google CLA Is signed.
The failed test is somehow related to the `[AllowPartiallyTrustedCallers]`. It's a attribute from .NET Framework. I work on a unix machine and cannot run the windows tests. @jskeet / @JamesNK  Do i just have to add `[AllowPartiallyTrustedCallers]` to the `UnsafeCollectionOperations` aswell?

jskeet:
@PascalSenn: It's always been an area I really find hard to understand, to be honest. @JamesNK could you advise, please?

JamesNK:
You need to put `[SecuritySafeCritical]` attribute on methods that return span. See https://github.com/protocolbuffers/protobuf/issues/7700 as an example

JamesNK:
I copied your branch locally and ran net462 tests. Here are the changes you need to make:

https://github.com/JamesNK/protobuf/commit/55dda6293ff82291288d630158499b9fec017d00

PascalSenn:
Thanks @jskeet  & @JamesNK, tests should pass now.

jskeet:
Ping @mkruskal-google for approval.

verdie-g:
Hi, I would be very interested by that feature too. Is there any chance to go forward with that PR?

JasonLunn:
@JamesNK - you're being flagged for not having signed the CLA but having contributed to the last commit. Can you resolve that?

JamesNK:
I'm pretty sure I already have. When I click on the CLA details link I see this:

image

JasonLunn:
> I'm pretty sure I already have. When I click on the CLA details link I see this:
> 
> image

Can you check that the contact info matches `j***s​@newtonking.com` - that's the email associated with commit https://github.com/protocolbuffers/protobuf/commit/46ed223b2c3aceece6d2128c258bb94bedf8fd26 which is the one in question.

JamesNK:
Ok, the signed in google account was different. I signed it with a different email address.

JasonLunn:
@PascalSenn - could you push a rebase? All the test failures are outside of the C# directory, I suspect their all due to drift with test targets given this PR's longevity.

prezaei:
@JamesNK , while you are at it, can you please add an overload for AddRange that accepts ReadOnlyMemory<T> or similar?

PR #16785, opened 2024-05-08
Fix indention formatter for google.protobuf.Any

q42jaap:
`google.protobuf.Any` formatting with indentation was somewhat off.

Formatting an `Any` as root object:
```json
{"@type": "type.googleapis.com/protobuf_unittest3.ForeignMessage",
"c": 1
}
```
changes to
```json
{
  "@type": "type.googleapis.com/protobuf_unittest3.ForeignMessage",
  "c": 1
}
```

For messages were `Any` is in a nested field the change makes more of a visual impact.
The `c` field seems to be at the same level as the `anyField`, but it's nested so it should be indented:
```json
{
  "anyField": {"@type": "type.googleapis.com/protobuf_unittest3.ForeignMessage",
  "c": 1
  }
}
```
changes to:
```json
{
  "anyField": {
    "@type": "type.googleapis.com/protobuf_unittest3.ForeignMessage",
    "c": 1
  }
}
```

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonFormatterTest.cs
@@ -901,6 +901,42 @@ namespace Google.Protobuf
         }
 
         [Test]
+        public void WriteValueWithIndentation_Any()
+        {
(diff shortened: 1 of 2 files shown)


jskeet (approved):
This is great, thanks. (I'm assuming all the existing tests work as-is...)

PR #16787, opened 2024-05-08
Add temporary alias for system_python.bzl at previous location .

zhangskz:
Cherry-pick of b77343aa2c6e81d81d5c8c627ab8ef8856fac2f5

This was added to protobuf_deps.bzl in cl/629786458 and no longer needs to be specified directly in WORKSPACE. However, an alias is needed to prevent users from breaking when upgrading from 26.x -> 27.x due to the obsolete load pointing to the wrong path.

This alias will be removed in a future release, tentatively 4.30.x.

PiperOrigin-RevId: 631821637


zhangskz:
Needs a cherry pick of cl/631918306 as well once submitted

haberman (approved):

PR #16807, opened 2024-05-09
Floats in ruby/protobuf are not in fact equal

shawnfrench:
The existing test is misleading:
https://github.com/protocolbuffers/protobuf/blob/82e83ddc95330bbd988ba874ea5aa37d343d4490/ruby/tests/common_tests.rb#L39-L40

This change adds a new test making the difference between ruby and protobuf floats explicit.

Something else to consider is currently the assignment of a ruby float returns the float with ruby precision. Perhaps the setter should return the float with protobuf precision?

### DIFF
--- a/ruby/tests/common_tests.rb
+++ b/ruby/tests/common_tests.rb
@@ -56,6 +56,16 @@ module CommonTests
     assert_equal :C, m.optional_enum
   end
 
+  def test_float_precision
+    # See https://github.com/protocolbuffers/protobuf/issues/6966
+    # A float field in protobuf is a single-precision floating point value.
+    # The Ruby Float type uses double-precision floating point.
+    # When you convert a double-precision floating point value to single precision, there is going to be some loss of precision.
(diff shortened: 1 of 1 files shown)


JasonLunn:
Interesting; this test exposes a behavior gap between the JRuby and JRuby FFI variants; the FFI-based implementation uses the same C-based implementation under the hood, but the legacy JRuby implementation is built on top of the Java Protobuf jar.

@haberman - does Java Protobuf handle float precision differently, or is this specific to JRuby's use of it?

haberman:
> @haberman - does Java Protobuf handle float precision differently, or is this specific to JRuby's use of it?

I suspect this is specific to JRuby's use.

I suspect that JRuby is caching the Ruby object in `this.fields` when the user calls `m.optional_float = 0.5`. Then `m.optional_float` is returning the user's cached object, which will have the original precision since it was not actually derived from the Java message's value.

PR #16825, opened 2024-05-10
    Work around windows path length limitations by moving immutable -> full in open source (new generator paths) and shortening long file names.     

zhangskz:
Re-enable windows bazel test.
    
PiperOrigin-RevId: 630443328

Cherry-pick of cd83a75315e7b564f8204630d86932d2791963a4


mkruskal-google (approved):

src/google/protobuf/compiler/objectivec/tf_decode_data.h
thomasvl, line 1: Shouldn't the full name .h file get updated to import this one?

zhangskz: It did initially, but was meant to be deleted since its not publicly exposed.

PR #16884, opened 2024-05-18
support bzlmod for//:protobuf_java and //:protobuf_javalite targets

mmorel-35:
* support /:protobuf_java and //:protobuf_javalite with bzlmod 
* upgrade rules_java to 6.0.0 (in MODULE.bazel) 
* upgrade rules_jvm_external to 6.0.0 (in MODULE.bazel) 
* upgrade com.google.errorprone:error_prone_annotations to 2.18.0 (in MODULE.bazel) 

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -3,7 +3,7 @@
 
 module(
     name = "protobuf",
-    version = "28.0-dev", # Automatically updated on release
+    version = "28.0-dev",  # Automatically updated on release
     compatibility_level = 1,
     repo_name = "com_google_protobuf",
 )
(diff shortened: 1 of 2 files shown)


keith:
nice! this will unblock bazel itself upgrading protobuf / grpc

MODULE.bazel
```diff
@@ -15,13 +15,39 @@ module(
...
+        "com.google.caliper:caliper:1.0-beta-3",
+        "com.google.code.findbugs:jsr305:3.0.2",
+        "com.google.code.gson:gson:2.8.9",
+        "com.google.errorprone:error_prone_annotations:2.18.0",
+        "com.google.j2objc:j2objc-annotations:2.8",
+        "com.google.guava:guava:32.0.1-jre",
+        "com.google.guava:guava-testlib:32.0.1-jre",
+        "com.google.truth:truth:1.1.2",
+        "junit:junit:4.13.2",
+        "org.mockito:mockito-core:4.3.1",
```
fmeum, line 41: This is problematic: All these deps leak into the default dependency closure. Only actual runtime deps should.

keith: are these really just dev deps? https://github.com/protocolbuffers/protobuf/blob/ee98ba2c18486938952a7748765f2e8e119e7fc4/java/util/BUILD.bazel#L16

keith: https://github.com/protocolbuffers/protobuf/blob/ee98ba2c18486938952a7748765f2e8e119e7fc4/BUILD.bazel#L353-L357

when i try to integrate w/ this patch into grpc-java this target is used which requires protobuf have a `@maven` with this stuff

fmeum: I think that we need to split this into a dev and a prod list of maven deps, with two different hub repo names

keith: I just found another repo where you needed this in the WORKSPACE too so yea I think that's just an issue that folks are probably avoiding by not updating protobuf

keith:
looks like the maven list is good now?

MODULE.bazel
```diff
@@ -15,13 +15,36 @@ module(
...
 bazel_dep(name = "zlib", version = "1.2.11")
 
 # TODO: remove after toolchain types are moved to protobuf
-bazel_dep(name = "rules_proto", version = "4.0.0")
\ No newline at end of file
+bazel_dep(name = "rules_proto", version = "4.0.0")
+
+maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
+maven.install(
+    artifacts = [
```
zhangskz, line 31: Can you explain why these artifacts are needed?

keith: This mirrors the WORKSPACE https://github.com/protocolbuffers/protobuf/blob/a978b75794a6ce4547c9db08a115c458d9190934/WORKSPACE#L61-L70, seems like the java targets just need these as typical third party deps? https://github.com/protocolbuffers/protobuf/blob/a978b75794a6ce4547c9db08a115c458d9190934/java/util/BUILD.bazel#L16-L20

https://github.com/protocolbuffers/protobuf/blob/a978b75794a6ce4547c9db08a115c458d9190934/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L14

zhangskz: thanks!

MODULE.bazel
```diff
@@ -15,13 +15,36 @@ module(
 bazel_dep(name = "abseil-cpp", version = "20230802.0.bcr.1", repo_name = "com_google_absl")
 bazel_dep(name = "bazel_skylib", version = "1.4.1")
 bazel_dep(name = "jsoncpp", version = "1.9.5")
+bazel_dep(name = "platforms", version = "0.0.8")
 bazel_dep(name = "rules_cc", version = "0.0.9")
-bazel_dep(name = "rules_java", version = "5.3.5")
-bazel_dep(name = "rules_jvm_external", version = "5.1")
+bazel_dep(name = "rules_java", version = "6.0.0")
+bazel_dep(name = "rules_jvm_external", version = "6.0")
```
zhangskz, line 21: Looks like this is failing for bazel 6 due to use_repo_rule which wasn't added until bazel 7 https://github.com/protocolbuffers/protobuf/actions/runs/9292936143/job/25574974352?pr=16884

We still support bazel 6 for a few more months: https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

keith: I think one option is to continue supporting bazel 6.x with the WORKSPACE, and say that bzlmod support is 7.x only. see also https://github.com/bazelbuild/rules_jvm_external/pull/1151

Issue #15957, opened 2024-02-26
`CLOCK_UPTIME_RAW` isn't always available on `__APPLE__`


adamnovak:
Since https://github.com/protocolbuffers/protobuf/commit/76d05d4cb9200c371c8894df21f37ba4060bdc8a and #15554, Protobuf tries to always use `CLOCK_UPTIME_RAW` if compiled for an `__APPLE__` system:

https://github.com/protocolbuffers/protobuf/blob/9c0d130168936120b4ed1953d041db86d7f362de/src/google/protobuf/map.h#L710-L715

But that symbol isn't always defined. On my machine, the MacOSX 14.2 SDK `time.h` only defines it if `!defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)` evaluates to true:

```
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
_CLOCK_MONOTONIC_RAW __CLOCK_AVAILABILITY = 4,
#define CLOCK_MONOTONIC_RAW _CLOCK_MONOTONIC_RAW
_CLOCK_MONOTONIC_RAW_APPROX __CLOCK_AVAILABILITY = 5,
#define CLOCK_MONOTONIC_RAW_APPROX _CLOCK_MONOTONIC_RAW_APPROX
_CLOCK_UPTIME_RAW __CLOCK_AVAILABILITY = 8,
#define CLOCK_UPTIME_RAW _CLOCK_UPTIME_RAW
_CLOCK_UPTIME_RAW_APPROX __CLOCK_AVAILABILITY = 9,
#define CLOCK_UPTIME_RAW_APPROX _CLOCK_UPTIME_RAW_APPROX
#endif
```

And Protobuf doesn't seem to define the appropriate macros to enable non-POSIX Darwin features before it includes `time.h`:
https://github.com/protocolbuffers/protobuf/blob/9c0d130168936120b4ed1953d041db86d7f362de/src/google/protobuf/map.h#L28-L30

This leads to a build failure in my project, which defines `_POSIX_C_SOURCE` to get Posix 2008 features, but does not define `_DARWIN_C_SOURCE` because it does not directly use Darwin extensions.

```
    /usr/local/Cellar/protobuf/25.3/include/google/protobuf/map.h:694:31: error: use of undeclared identifier 'CLOCK_UPTIME_RAW'
        s = clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
[...]

PR #16951, opened 2024-05-28
Don't use CLOCK_UPTIME_RAW if it won't be defined

adamnovak:
This will fix https://github.com/protocolbuffers/protobuf/issues/15957 by not trying to use CLOCK_UPTIME_RAW if the feature test macros are not currently set to make it available.

Since the build system is in charge of the feature test macros, and since this header might be included under any end user program's build system, we can't just #define the macros we would like.


adamnovak:
I've had copyright for this contribution assigned from my employer to me personally, so I can sign the CLA for this contribution as an individual.

adamnovak:
I tried to do the CLA "rescan", and after completing the CAPTCHA I got an error:

>400. That’s an error.
>
>That’s all we know.

adamnovak:
@deannagarcia It looks like the inactivity bot has closed the associated issue #15957 due to there not being any progress on getting it fixed. When might you be available to review this PR?

The CLA rescan still doesn't work, I think because the time to get the copyright assigned to me exceeded the time for which the "rescan" link in the CLA check was valid for (it apparently expired on 2024-07-12).

acozzette (approved):

PR #16956, opened 2024-05-28
bazel: Update debian version on bcr CI

keith:
Debian buster is EOL next month, it also defaults to gcc 8.x, which no longer builds cleanly because of this issue fixed in 9.x https://stackoverflow.com/a/73511549/902968


zhangskz (approved):

honglooker:
Looks like there's a merge conflict; please resolve if you'd like to continue with this merge.

keith:
rebased!

PR #16976, opened 2024-05-29
csharp: use span-based ToBase64 in ByteString

verdie-g:
With .NET 5+, a copy can be avoided when the memory is not an array.


jskeet (approved):
This looks fine to me, thanks.

@mkruskal-google I don't think we'll need release notes for this.

mkruskal-google (approved):

Issue #16944, opened 2024-05-25
Build from main branch fail on macos 14.5 x86


qinghui-xu:
**What version of protobuf and what language are you using?**
Version: main branch (commit 2f3242c57)
Language: C++/Java/Python/C#/Ruby/PHP/Objective-C/Javascript

**What operating system (Linux, Windows, ...) and version?**
MacOS Sonoma (14.5)
**What runtime / compiler are you using (e.g., python version or gcc version)**
Apple clang version 15.0.0 (clang-1500.3.9.4)
**What did you do?**
Run `bazel build :protoc :protobuf`

**What did you see instead?**
```
ERROR: /Sources/thirdparty/protobuf/src/google/protobuf/io/BUILD.bazel:98:11: Compiling src/google/protobuf/io/printer.cc failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target //src/google/protobuf/io:printer) external/local_config_cc/cc_wrapper.sh -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics ... (remaining 42 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from src/google/protobuf/io/printer.cc:12:
bazel-out/darwin_x86_64-fastbuild/bin/src/google/protobuf/io/_virtual_includes/printer/google/protobuf/io/printer.h:904:19: error: 'get<std::function<bool ()>, std::string, std::function<bool ()>>' is unavailable: introduced in macOS 10.13
    value = absl::get<Callback>(that.value);
                  ^
bazel-out/darwin_x86_64-fastbuild/bin/src/google/protobuf/io/_virtual_includes/printer/google/protobuf/io/printer.h:849:11: note: in instantiation of function template specialization 'google::protobuf::io::Printer::ValueImpl<false>::operator=<true>' requested here
    *this = that;
          ^
bazel-out/darwin_x86_64-fastbuild/bin/src/google/protobuf/io/_virtual_includes/printer/google/protobuf/io/printer.h:1136:12: note: in instantiation of function template specialization 'google::protobuf::io::Printer::ValueImpl<false>::ValueImpl<true>' requested here
    return ValueView(it->second);
           ^
[...]

PR #16993, opened 2024-05-31
Fix segmentation faults with enabled keep_descriptor_pool_after_request

### LINKED ISSUE #16894: [PHP] Heap Corruption when keep_descriptor_pool_after_request is active
When protobuf messages are created in PHP scripts executed in CLI, `zend_mm_heap corrupted` errors and memory leaks are reported.

I created a reproducer repository here. 
https://github.com/simonberger/protobuf-reproducer

The only thing needed is initializing any protobuf message. I picked the available`Google\Protobuf\Api` message. With codeception the output is

```
Tests.Protobuf Tests (1) ------------------------------------------------------------------------------------------------------------------------------------------------------------
✔ ProtobufTest: Protobuf(0.01s)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Time: 00:00.110, Memory: 10.00 MB

OK (1 test, 0 assertions)
zend_mm_heap corrupted
```

PHP and Extension versions:
- The docker image is using newest debian with PHP 8.3.7 and protobuf extension 3.25.3
- I reproduced it also on a Mac M1 with PHP 8.3.7 and protobuf extension 4.26.1

I started the repository with executing it in a codeception test, because this is what we are using.
The problem can be reproduced much easier. A call to a PHP script just with the message creation is enough.
The script is also added to the repository.

```php
<?php
use Google\Protobuf\Api;

$message = new Api();
// Not necessary for the zend_mm_heap corrupted error to appear.
$message->setName('TestApi');
```
 `php public/protobuf_init.php`

These problems where already reported in the old bug https://github.com/protocolbuffers/protobuf/issues/9215 and not only in commands. We do not yet use the protobuf extension in production and just noticed the errors while executing tests.

simonberger:
Fixes #16894 

It is tested on command execution only, where it fixes the issues of reported memory leaks, heap corruption and segmentation fault.

I could not run tests on my Mac because of #16944

### DIFF
--- a/php/ext/google/protobuf/protobuf.c
+++ b/php/ext/google/protobuf/protobuf.c
@@ -143,11 +143,11 @@ static PHP_GINIT_FUNCTION(protobuf) { protobuf_globals->global_symtab = NULL; }
 static PHP_RINIT_FUNCTION(protobuf) {
   // Create the global generated pool.
   // Reuse the symtab (if any) left to us by the last request.
-  upb_DefPool* symtab = PROTOBUF_G(global_symtab);
-  if (!symtab) {
-    PROTOBUF_G(global_symtab) = symtab = upb_DefPool_New();
-    zend_hash_init(&PROTOBUF_G(name_msg_cache), 64, NULL, NULL, 0);
-    zend_hash_init(&PROTOBUF_G(name_enum_cache), 64, NULL, NULL, 0);
+  if (!PROTOBUF_G(global_symtab)) {
+    zend_bool persistent = PROTOBUF_G(keep_descriptor_pool_after_request);
+    PROTOBUF_G(global_symtab) = upb_DefPool_New();
+    zend_hash_init(&PROTOBUF_G(name_msg_cache), 64, NULL, NULL, persistent);
+    zend_hash_init(&PROTOBUF_G(name_enum_cache), 64, NULL, NULL, persistent);
   }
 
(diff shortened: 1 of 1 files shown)


hakimio:
@haberman take a look?

haberman:
Hmm, looks like there are some test failures unrelated to your change. Want to rebase and push again?

simonberger:
> Hmm, looks like there are some test failures unrelated to your change. Want to rebase and push again?

@haberman Sure, done

haberman (approved):

MrMage:
Looks like only some internal "copybara" check failed. Is that a blocker, or can this be merged?

@bshaffer would it be possible for you to merge this in your function as an owner?

haberman:
All PRs must be committed internally first, because the internal repo is the source of truth. If we tried to merge the PR directly in GitHub I think it might get reverted by the next CL that syncs to GitHub.

The Copybara failure should be temporary. Internally the CL is in @honglooker's court to review and resolve any issues with the internal CI.

simonberger:
I already asked in the related issue about if a backport to 25.x should be done what I personally do not really care about, but I just realized this fix currently would not do it in a  potential 27.2 release.
Is this something you do internally or shall I create a backport fix for `27.x` and possibly 25.x. Some more clear expectations for the different types of pull requests would be good. I read in the guide everything shall target main first which is unlike most other repositories I know. What shall follow after is a bit uncertain to me.

PR #17013, opened 2024-06-03
Removes msvc warning in map

TinyTinni:
Hi,
We get a compiler warning in MSVC that a signed/unsigned comparison is executed.

The warning is at around line 1073 in src/google/protobuf/map.h
```cpp
ABSL_DCHECK_GE(new_num_buckets, kMinTableSize); // <-- signed/unsigned comparison
const auto old_table = table_;
const map_index_t old_table_size = num_buckets_;
num_buckets_ = new_num_buckets;
```

This is due to `kMinTableSize` is a enum value. Probably a C++98 way to enforce constexpr.

This PR changes `kMinTableSize` to be a constexpr unsigned `map_index_t` and removes the casts instead of fixing the warning with a cast at ne 1073. 
It seems like `kMinTableSize` is correlated with the `TableSize` of this class, which is map_index_t. Therefore the min value can at maximum the max value of map_index_t. And the min value of `kMinTableSize` can also not be negative, as this makes no sense.

It is also nicer to have fewer casts.

### DIFF
--- a/src/google/protobuf/map.h
+++ b/src/google/protobuf/map.h
@@ -559,7 +559,7 @@ class PROTOBUF_EXPORT UntypedMapBase {
 
  protected:
   // 16 bytes is the minimum useful size for the array cache in the arena.
-  constexpr static map_index_t kMinTableSize = 16 / sizeof(void*);
+  enum : map_index_t { kMinTableSize = 16 / sizeof(void*) };
 
  public:
   Arena* arena() const { return this->alloc_.arena(); }
--- a/src/google/protobuf/map.h
+++ b/src/google/protobuf/map.h
@@ -29,7 +29,6 @@
 #include <time.h>
 #endif
 
-#include "google/protobuf/stubs/common.h"
 #include "absl/base/attributes.h"
 #include "absl/container/btree_map.h"
 #include "absl/hash/hash.h"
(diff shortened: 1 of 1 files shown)


acozzette:
@TinyTinni Thank you for the fix, but we need to have you agree to the CLA before we can review this.

acozzette:
Oh sorry, I missed that the CLA check is passing now.

TinyTinni:
Hi,
not sure why GCC8.1 with linux 32 bit build does not like to link the static constexpr member.  
I can only guess here, try a typed enum and hope that it will fix the compile error or do you have any advice.

Can I just commit it on my branch and the buildpipeline tries it again?
Otherwise, I will look for a local gcc8.1 and a 32 bit build.

TinyTinni:
added a commit to hopefully fix gcc 32 bit 8.1. If it fails, I will try to set it up locally.

TinyTinni:
Looking at the failed jobs, I am not really sure what is missing.

1 Ruby test failed, I do not understand how this has anything to do with my changes and in the run before, it didn't fail.

micropb job failing, but the tests run through. Does some after_test_step fails there?

acozzette:
@TinyTinni Sorry for the trouble, would you mind rebasing your branch and pushing it again? I believe that will fix the test failures.

TinyTinni:
Sorry, first rebase failed a bit and the bot triggered more reviewers.

But the branch is now up-to-date with the main branch.

TinyTinni:
Thanks for the rebase advice. The pipeline passed this time.

acozzette (approved):

PR #17052, opened 2024-06-06
Fix a warning from uninitialized variables

guitar24t:
Currently, when a project that includes protobuf is compiled using -Werror, the build will fail due to these variables not being initialized. This fixes this warning allowing the build to complete successfully


acozzette:
@guitar24t What kind of error are you getting?

guitar24t:
The error I received was a "Warning: variable may be used uninitialized"

PR #17054, opened 2024-06-06
Export PrivacyInfo.xcprivacy

mjburghard:
Export `PrivacyInfo.xcprivacy` file to allow users of protobuf to access the file from Bazel, e.g. to bundle it with an iOS application.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -12,7 +12,10 @@ load(":protobuf.bzl", "internal_objc_proto_library", "internal_php_proto_library
 
 licenses(["notice"])
 
-exports_files(["LICENSE"])
+exports_files([
+    "LICENSE",
+    "PrivacyInfo.xcprivacy",
+])
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


acozzette:
@mjburghard Could you rebase your branch against main and push it again? I believe that will fix the test failures.

thomasvl (commented):
I don't actually know what the bazel rules are doing for CC or ObjC libraries to collect privacy manifest info. I would think something has to merge the data as other targets might need the info also.

Is exporting the file required, or do the rules use an aspect to collect the data from the whole graph an merge it?

thomasvl (approved):
Approving, but just copying the file likely isn't right since you'll need to merge it with the information for the rest of your code.

mjburghard:
Yeah, this is part of a more elaborate setup.

mjburghard:
Are there any steps I can do to resolve the CI failure?

acozzette (approved):

PR #17057, opened 2024-06-06
Fixes unitialized variables causing linker error with gRPC

kongen:
Variables left unitialized causes linker warnings when linked statically with gRPC


acozzette:
I'm surprised that this would be necessary. What kind of linker warnings are you getting?

kongen:
Hi Adam,
We're getting uninitialized variable warnings, which due to our settings becomes errors.

acozzette:
I'm reluctant to add this initialization, because I believe the code is already correct in the sense that we don't try to read the variable without first initializing it.

Issue #13726, opened 2023-08-24
Add annotations for exporting generated functions from a Windows DLL


dawidcha:
I'm currently engaged in some work with gRPC aiming at merging a PR which permits gRPC to be built as a dynamic library under Windows (i.e. a DLL). [here](https://github.com/grpc/grpc/pull/34103). gRPC makes use of μpb and needs to expose symbols generated by it from a DLL. That requires that the functions marked as extern be annotated with '__declspec(dllexport)' when compiled and '__declspec(dllimport)' when consumed.

For the protoc compiler, this is achieved by prefixing the file output location on the command line with [dllexport_decl=MY_EXPORT_MACRO](https://stackoverflow.com/questions/64449006/generate-a-dll-which-uses-protocol-buffers), and then defining MY_EXPORT_MACRO as either __declspec(dllimport) or __declspec(dllexport); or empty string, as appropriate, but this feature doesn't appear to be available in μpb.

Can something similar be added?

h-vetinari:
Not stale.

h-vetinari:
This is not stale, and there's an on open PR for this as well, at least in the context of upb (for now): https://github.com/protocolbuffers/protobuf/pull/14981

h-vetinari:
Not stale, and has a new PR now too: #17079

h-vetinari:
Not stale

h-vetinari:
Not stale (c.f. also #19422)

PR #17079, opened 2024-06-08
Adding DLL export/import tags to generated public upb API (redux)

h-vetinari:
Picking up #14981 from @dawidcha after several months of radio silence.

Quoting the OP of that PR:
> I have been collaborating with the grpc developers to make it possible to build that library as a Windows DLL - a couple of PRs were already merged, more like [grpc/grpc#34345](https://github.com/grpc/grpc/pull/34345) are pending.
> 
> The grpc library incorporates some upb-generated, and upbdefs-generated code into grpc.dll, which is referenced by other code that consumes the library. Since this is now a DLL, that code doesn't know how to link to these generated symbols because they are not annotated with __declspec(dllimport).
> 
> This PR aims to fix that by introducing a parameter 'dllexport_tag' to the upb and upbdefs plugins. That parameter should be a string e.g. MYAPP_DLL and when set, the extern symbols are annotated with a macro with that name. This can either be set externally to __declspec(dllimport) or, as is usual practice, when compiling code into a DLL, the macro <dllexport_tag>_EXPORT (i.e. MYAPP_DLL_EXPORT) is defined, and when consuming the DLL <dllexport_tag>_IMPORT is defined if neither are defined then the MYAPP_DLL macro becomes empty string which is what you want for building a static library.
> 
> This is a continuation of #14230
> 
> Fixes: #14255

Towards #13726
Closes #14981


haberman (approved):

upb_generator/protoc-gen-upb.cc
```diff
@@ -1128,6 +1129,8 @@ bool ParseOptions(Plugin* plugin, Options* options) {
       options->bootstrap = true;
     } else if (pair.first == "experimental_strip_nonfunctional_codegen") {
       continue;
+    } else if (pair.first == "dllexport_decl") {
```
haberman, line 1132: Sorry, I just noticed: why do we need this in the protoc-gen-upb code generator?

This generator only generates inline methods, so should not require dllexport.

Can we remove the changes in this file, and only change protoc-gen-upbdefs??

h-vetinari: I just took the changes from #14981, so not sure about the intention behind, but I'll remove it for now.

Issue #17068, opened 2024-06-07
Ruby gem fails to compile under x86_64-linux-gnux32 on bookworm


haileys:
```
linking shared-object google/protobuf_c.so
/usr/bin/ld: protobuf_c.so: no symbol version section for versioned symbol `memcpy@GLIBC_2.2.5'
collect2: error: ld returned 1 exit status
make: *** [Makefile:265: protobuf_c.so] Error 1
```

This is due to the [`wrap_memcpy.c`](https://github.com/protocolbuffers/protobuf/blob/08a6c36f5ace9c5d41eaa3df6d718e3f7acdcf47/ruby/ext/google/protobuf_c/wrap_memcpy.c), which appears to force the `memcpy` symbol to link against a specific version exported by glibc.

Removing the wrap_memcpy code fixes the issue.

haileys:
It appears that Debian bookworm's [libc6-x32](https://packages.debian.org/bookworm/libc6-x32) only provides the `GLIBC_2.16` version of this symbol

```
$ objdump -T /usr/libx32/libc.so.6 | grep memcpy
000aae50  w   DF .text	00000008  GLIBC_2.16  wmemcpy
00110730 g    DF .text	00000018  GLIBC_2.16  __wmemcpy_chk
0010f470 g   iD  .text	00000107  GLIBC_2.16  __memcpy_chk
00095fc0 g   iD  .text	00000107  GLIBC_2.16  memcpy
```

haileys:
This can also be fixed by adding `&& !defined(__ILP32__)` to the `#if defined(__x86_64__) && defined(__GNU_LIBRARY__)` preprocessor condition in `wrap_memcpy.c`

acozzette:
Thank you for the bug report. Your suggested fix sounds good to me if you would like to send us a pull request.

PR #17098, opened 2024-06-11
Don't use versioned memcpy symbol on linux-x32

haileys:
Debian bookworm's x32 libc does not provide the version required here:

```
$ objdump -T /usr/libx32/libc.so.6 | grep memcpy
000aae50  w   DF .text	00000008  GLIBC_2.16  wmemcpy
00110730 g    DF .text	00000018  GLIBC_2.16  __wmemcpy_chk
0010f470 g   iD  .text	00000107  GLIBC_2.16  __memcpy_chk
00095fc0 g   iD  .text	00000107  GLIBC_2.16  memcpy
```

Fixes https://github.com/protocolbuffers/protobuf/issues/17068

cc @acozzette

### DIFF
--- a/ruby/ext/google/protobuf_c/wrap_memcpy.c
+++ b/ruby/ext/google/protobuf_c/wrap_memcpy.c
@@ -16,7 +16,7 @@
 // This wrapper is enabled by passing the linker flags -Wl,-wrap,memcpy in
 // extconf.rb.
 #ifdef __linux__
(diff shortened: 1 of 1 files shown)


haileys:
Sorry, I don't sign CLAs. Since this is a trivial change, I'm happy if someone just commits it directly.

acozzette:
Thank you for the pull request, but unfortunately we can't accept PRs without a CLA.

PR #17171, opened 2024-06-18
Use is_a? instead of kind_of?

ixti:
If it walks like a duck, quacks like a duck - it might be a duck.

Rails ActiveSupport::TimeWithZone is a Time alike and can de serialized correctly, that's why when you do: `Google::Protobuf::Timestamp.new(Time.current)` it works, but as this code checks for `kind_of?` it prevents assigning `Time.current` to a field directly without type-casting it into Ruby native Time.


JasonLunn:
@ixti - Thanks for your contribution. Since the [canonical definition](https://ruby-doc.org/3.2.2/Object.html#method-i-kind_of-3F) of `is_a?` is just an alias of `kind_of?`m I'm having trouble understanding what problem this PR solves as it doesn't reference an open issue. Perhaps you can you include a test that fails without this change?

ixti:
> @ixti - Thanks for your contribution. Since the [canonical definition](https://ruby-doc.org/3.2.2/Object.html#method-i-kind_of-3F) of `is_a?` is just an alias of `kind_of?` I'm having trouble understanding what problem this PR solves as it doesn't reference an open issue. Perhaps you can you include a test that fails without this change?

Oops. Sorry, will try avoid late-night PR openning. Somehow I read `kind_of?` as `instance_of?` LOL

ixti:
Would be nice if message.c would check if object responds to `#to_time` (Rails extension to both TimeWithZone and Time) and called it before checking the type :D

JasonLunn:
It should be possible to swap the direct call to `rb_obj_is_kind_of(value, rb_cTime)` for `rb_funcall(value, rb_intern("is_a?"), 1, rb_cTime)` or similar in order to invoke the overloaded method. We'd still need a test. You might get that behavior for free if you switch to using the FFI implementation which does not use `message.c`: https://github.com/protocolbuffers/protobuf/blob/84e75a5940909cc0e881f65e62e539ede33c6ebe/ruby/lib/google/protobuf/ffi/internal/convert.rb#L72

ixti:
> It should be possible to swap the direct call to rb_obj_is_kind_of(value, rb_cTime) for rb_funcall(value, rb_intern("is_a?"), 1, rb_cTime) or similar in order to invoke the overloaded method. We'd still need a test.

If it's fine to add active_support into testing, then the test will be pretty simple:

``` ruby
require "active_support/all"

Google::Protobuf::Timestamp.from_time(Time.current)
# works fine

MyMessage.new(created_at: Time.current)
# fails
```

Message for this one would look like:

``` protobuf
message MyMessage {
  google.protobuf.Timestamp created_at = 1;
}
```

Without ActiveSupport, test might look like:

``` ruby
TimeAlike = Struct.new(:tv_sec, :tv_nsec) do
  def self.current
    time = Time.new.utc

    new(time.tv_sec, time.tv_nsec)
  end

  alias_method :to_i, :tv_sec
  alias_method :nsec, :tv_nsec

  def is_a?(klass)
    klass == ::Time || super
  end

  alias_method :kind_of?, :is_a?
end

Google::Protobuf::Timestamp.from_time(TimeAlike.current)
# works

MyMessage.new(created_at: Time.current)
# fails
```

> You might get that behavior for free if you switch to using the FFI implementation which does not use message.c

That I did not know. Thank you! Will give it a try.

Issue #17177, opened 2024-06-18
Request to change access level of ByteString.AttachBytes() method to public


VAllens:
I suggest changing the access level of the static method [AttachBytes](https://github.com/protocolbuffers/protobuf/blob/2088bc6e66850aa02cceaaae6096f39d20f1c677/csharp/src/Google.Protobuf/ByteString.cs#L38) of ByteString from internal to public, in order to make this static method accessible to users.

In some situations, if I already have a ReadOnlyMemory<byte> containing data, and I know that this ReadOnlyMemory<byte> remains valid throughout the lifetime of ByteString, I would like to directly create a ByteString object based on this existing ReadOnlyMemory<byte>, thus avoiding data copying operations.

The current static method [ByteString.AttachBytes()](https://github.com/protocolbuffers/protobuf/blob/2088bc6e66850aa02cceaaae6096f39d20f1c677/csharp/src/Google.Protobuf/ByteString.cs#L38) does exactly this. However, it is currently defined with internal access, preventing direct invocation.

VAllens:
The [UnsafeByteOperations.UnsafeWrap](https://github.com/protocolbuffers/protobuf/blob/2088bc6e66850aa02cceaaae6096f39d20f1c677/csharp/src/Google.Protobuf/UnsafeByteOperations.cs#L53) method meets my requirements, so I will close this issue.

PR #17178, opened 2024-06-18
Change access level of ByteString.AttachBytes() method to public

VAllens:
In some situations, if I already have a ReadOnlyMemory containing data, and I know that this ReadOnlyMemory remains valid throughout the lifetime of ByteString, I would like to directly create a ByteString object based on this existing ReadOnlyMemory, thus avoiding data copying operations.

https://github.com/protocolbuffers/protobuf/issues/17177

### DIFF
--- a/csharp/src/Google.Protobuf/ByteString.cs
+++ b/csharp/src/Google.Protobuf/ByteString.cs
@@ -33,9 +33,9 @@ namespace Google.Protobuf
         private readonly ReadOnlyMemory<byte> bytes;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


VAllens:
> I _definitely_ don't want to do this directly in `ByteString` - that's the equivalent of asking Microsoft to create a method that creates a `System.String` backed by a `ReadOnlyMemory<char>`.
> 
> There's already `UnsafeByteOperations.UnsafeWrap` which is public - does that not do exactly what you want?

Oh, yes. It meets my needs. 
Apparently, there has been this public UnsafeWrap method before, but I never noticed it.

Issue #17117, opened 2024-06-12
[Bazel] Stop using rules_jvm_external default maven repository name


joca-bt:
The `protobuf` Bazel module depends on `rules_jvm_external` and it's including the artifacts it needs under the default maven repository name (`@maven`, https://github.com/protocolbuffers/protobuf/blob/main/WORKSPACE#L65). This can create some issues for `rules_jvm_external` users, as described in https://github.com/bazelbuild/rules_jvm_external/issues/1168.

Is it necessary for these artifacts to be under the default maven repository name? Otherwise, could we use a different name for the maven repository so it won't conflict with other modules or users' projects (https://github.com/bazelbuild/rules_jvm_external?tab=readme-ov-file#multiple-maven_installjson-files)? This should also lead to a better isolation of this module.

For example:

```
maven_install(
    name = "maven_protobuf",
    ...
)
```

PR #17190, opened 2024-06-20
Use a custom name for the maven repository (maven_protobuf)

joca-bt:
Addresses https://github.com/protocolbuffers/protobuf/issues/17117.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -63,6 +63,7 @@ rules_jvm_external_setup()
 load("@rules_jvm_external//:defs.bzl", "maven_install")
 
 maven_install(
+    name = "maven_protobuf",
     artifacts = PROTOBUF_MAVEN_ARTIFACTS,
     # For updating instructions, see:
     # https://github.com/bazelbuild/rules_jvm_external#updating-maven_installjson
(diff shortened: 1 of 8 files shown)


joca-bt:
Could someone review this PR? cc @haberman

zhangskz:
My understanding is that the contents of this WORKSPACE file shouldn't leak the maven repository -- this is a dev dependency in WORKSPACE for packaging protobuf but not actually in protobuf_deps.bzl (that users should transitively depend on).

The rules_jvm_external issue linked seems to use BCR [v21.7](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/protobuf/21.7). I think this would actually need to be patched to MODULE.bazel 21.7.bcr.1 for rules_jvm_external. I think a `dev_dependency=True` in the use_extension for maven: https://bazel.build/external/migration#specify_dev_dependencies might address this without having to rename.

joca-bt:
If I manually fix the protobuf version to a more recent version, then the problem seems to goes away. I think we can close it for now, will try updating the protobuf dependency version in the relevant projects.

PR #17191, opened 2024-06-20
[C#] Use input Stream.Length if seekable

### LINKED ISSUE #17111: [C#] Large repeated field doesn't pre-allocate the array on deserialization
Version: 3.27.0
Language: C#
.NET: 8.0

**What did you do?**

```proto
syntax = "proto2";

message Test {
  repeated float values = 1 [packed=true];
}
```

```csharp
using Google.Protobuf;

MemoryStream ms = new();
new Test { Values = { Enumerable.Range(0, 100_000).Select(x => (float)x) }}.WriteTo(ms);

ms.Position = 0;
var x = Test.Parser.ParseFrom(ms);
```

When deserializing the 100K values, I would expect the RepeatedField's array to be allocated once.

**What did you see instead?**

https://github.com/protocolbuffers/protobuf/blob/a6d482748271f36fff5afe1abea3889789d0e3b5/csharp/src/Google.Protobuf/Collections/RepeatedField.cs#L112-L131

In this case `codec.FixedSize > 0 && length % codec.FixedSize == 0` is true, but `ParsingPrimitives.IsDataAvailable(ref ctx.state, length)` is false. So the code branches to the variable size part which makes the array grow as the data is read. That doubles the amount of memory allocated and in production, my array is several MB large so most temporary arrays reach LOH :(

In both branches, the code seems to be almost the same, that is,
```csharp
while (!SegmentedBufferHelper.IsReachedLimit(ref ctx.state))
{
    var val = reader(ref ctx);
    // do something with val
}
```
Could you help me understand the purpose of `IsDataAvailable` here? :thinking:

verdie-g:
Closes #17111.

A large repeated packed field of floats is considered unpacked when the input is a Stream to avoiding allocating more than the input size. Though, if the Stream is a MemoryStream or a FileStream, the Length is known. I'm using Stream.CanSeek to detect that case.

The feature already existing for ROSequence. I'm just extending it for seekable streams.

### DIFF
--- a/csharp/src/Google.Protobuf/SegmentedBufferHelper.cs
+++ b/csharp/src/Google.Protobuf/SegmentedBufferHelper.cs
@@ -277,10 +277,10 @@ namespace Google.Protobuf
                 return stream.InternalBuffer.Length;
             }
 
-            // By convention, if CanSeek is true, then the Length should be available.
+            // By convention, if CanSeek is true, then the Length and Position should be available.
             if (stream.InternalInputStream.CanSeek)
             {
-                return stream.InternalInputStream.Length;
+                return stream.InternalInputStream.Length - stream.InternalInputStream.Position;
             }
 
             return null;
--- a/csharp/src/Google.Protobuf/SegmentedBufferHelper.cs
+++ b/csharp/src/Google.Protobuf/SegmentedBufferHelper.cs
@@ -22,7 +22,7 @@ namespace Google.Protobuf
     [SecuritySafeCritical]
     internal struct SegmentedBufferHelper
     {
-        private int? totalLength;
+        private long? totalLength;
         private ReadOnlySequence<byte>.Enumerator readOnlySequenceEnumerator;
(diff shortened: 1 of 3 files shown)


csharp/src/Google.Protobuf/SegmentedBufferHelper.cs
```diff
@@ -269,5 +269,22 @@ private static void CheckCurrentBufferIsEmpty(ref ParserInternalState state)
...
+        {
+            if (stream.InternalInputStream == null)
+            {
+                return stream.InternalBuffer.Length;
+            }
+
+            // By convention, if CanSeek is true, then the Length should be available.
+            if (stream.InternalInputStream.CanSeek)
+            {
+                return stream.InternalInputStream.Length;
```
verdie-g, line 283: For non-BCL Stream, this could still throw which would be a small breaking change. Is that acceptable?

jskeet:
Closing as discussed in #17111.

PR #17198, opened 2024-06-20
Update GHA tests to return single pass/fail signal at the end

deannagarcia:
Before this PR, we stored a list internally of tests that must pass on presubmit and tried to keep it up to date.

This PR moves that information keeping into GitHub by adding a 'continuous-only' variable to most testing matrices to allow authors to specify which of their tests should be skipped on presubmit. During presubmit, tests that were specified to not run on presubmit will not be run and their names will be prefixed with "[SKIPPED]". All continuous only tests will be suffixed with "(Continuous)".

At the end of running all the tests, we have a single "All Blocking Tests" signal that will tell us whether all of the necessary tests have passed (either for presubmit or continuous based on how the test was triggered).

I've tested this from a different branch [here](https://github.com/protocolbuffers/protobuf/actions/runs/9602443750?pr=17151) and from a different fork [here](https://github.com/protocolbuffers/protobuf/actions/runs/9602554500?pr=17192). These should be the same and are as far as I can tell.

I also have a continuous test run [here](https://github.com/protocolbuffers/protobuf/actions/runs/9603824200) which runs the entire test suite.


.github/workflows/test_cpp.yml
```diff
@@ -224,27 +239,33 @@ jobs:
...
+            run-on-presubmit: true
           - name: C++17
             flags: -DCMAKE_CXX_STANDARD=17
+            run-on-presubmit: false
           - name: C++20
             flags: -DCMAKE_CXX_STANDARD=20
+            run-on-presubmit: true
 
-    name: Linux CMake GCC ${{ matrix.name }}
+    name: ${{ (!matrix.run-on-presubmit && inputs.test-type == 'presubmit') && '[SKIPPED]' || '' }} Linux CMake GCC ${{ matrix.name }} ${{ !matrix.run-on-presubmit && '(Continuous)' || '' }}
```
mkruskal-google, line 250: This seems like a lot of sensitive logic we're duplicating many times.  Is it possible to reduce this a little to make it harder to mess up?  Notably, we might want to centralize the spelling of `'[SKIPPED]'` and `'(Continuous)'`.  If we put them both on the same side we might be able to reduce the number of branches too.

e.g. something like `${{ matrix.skip-on-presubmit && inputs.continuous-prefix }}`, and then the runner could compute the prefix in 1 place

.github/workflows/test_cpp.yml
```diff
@@ -224,27 +239,33 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - name: Checkout pending changes
+        if: ${{ matrix.run-on-presubmit || inputs.test-type == 'continuous' }}
```
mkruskal-google, line 254: If someone mispelled "continuous" this could just silently do nothing.  Making a new boolean input might make this less error-prone, but would unfortunately still have its own overhead (I'm not convinced which is better)

.github/workflows/test_runner.yml
```diff
@@ -77,6 +77,9 @@ jobs:
       # Store the sha for checkout so we can easily use it later.  For safe
       # events, this will be blank and use the defaults.
       checkout-sha: ${{ steps.safe-checkout.outputs.sha }}
+      # Stores a string denoting whether this is a presubmit or continuous
+      # run. This helps us determine which tests to block on.
+      test-type: ${{ (github.event_name == 'pull_request' || github.event_name == 'pull_request_target') && 'presubmit' || 'continuous' }}
```
mkruskal-google, line 82: It feels a bit awkward to put this in a "Check for Safety" job.  Should we rename this?

.github/workflows/test_runner.yml
```diff
@@ -206,3 +218,15 @@ jobs:
     with:
       safe-checkout: ${{ needs.check-tag.outputs.checkout-sha }}
```
mkruskal-google, line 219: Missing input?  Or did you skip it because it's not needed?

.github/workflows/test_runner.yml
```diff
@@ -206,3 +218,15 @@ jobs:
...
     secrets: inherit
+
+  # This test depends on all blocking tests and indicates whether they all suceeded.
+  all_blocking_tests:
+    name: All Blocking Tests
+    needs: [bazel, cpp, java, python, ruby, php, php-ext, csharp, objectivec, rust, upb, staleness]
```
mkruskal-google, line 225: We should probably add IFTTT comments even if it's unenforceable from GH.  I can't think of any other way to notify people when this list goes stale

.github/workflows/test_runner.yml
```diff
@@ -206,3 +218,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check test results
+        run: "${{ !contains(join(needs.*.result, ' '), 'failure') && !contains(join(needs.*.result, ' '), 'cancelled') && !contains(join(needs.*.result, ' '), 'skipped') }}"
```
mkruskal-google, line 229: Won't staleness tests be skipped on 'scheduled' runs?  Unfortunately we don't have any way to truly test that, but you could update the `if:` above to include `workflow_dispatch` and see what happens when you manually trigger this

.github/workflows/test_upb.yml
```diff
@@ -171,47 +175,53 @@ jobs:
...
-          - { os: macos-12, python-version: "3.12", architecture: x64, type: 'binary' }
-          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'source' }
-          - { os: macos-11, python-version: "3.8", architecture: x64, type: 'source' }
-          - { os: ubuntu-latest, python-version: "3.12", architecture: x64, type: 'source' }
-          - { os: macos-12, python-version: "3.12", architecture: x64, type: 'source' }
+          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'binary', run-on-presubmit: true }
+          - { os: macos-11, python-version: "3.8", architecture: x64, type: 'binary', run-on-presubmit: true }
+          - { os: ubuntu-latest, python-version: "3.12", architecture: x64, type: 'binary', run-on-presubmit: true }
+          - { os: macos-12, python-version: "3.12", architecture: x64, type: 'binary', run-on-presubmit: true }
+          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'source', run-on-presubmit: false }
```
mkruskal-google, line 182: I think leaving these unset for the false case might help reduce clutter

.github/workflows/test_upb.yml
```diff
@@ -171,47 +175,53 @@ jobs:
...
           # coverage.
-          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'binary' }
-          - { os: macos-11, python-version: "3.8", architecture: x64, type: 'binary' }
-          - { os: ubuntu-latest, python-version: "3.12", architecture: x64, type: 'binary' }
-          - { os: macos-12, python-version: "3.12", architecture: x64, type: 'binary' }
-          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'source' }
-          - { os: macos-11, python-version: "3.8", architecture: x64, type: 'source' }
-          - { os: ubuntu-latest, python-version: "3.12", architecture: x64, type: 'source' }
-          - { os: macos-12, python-version: "3.12", architecture: x64, type: 'source' }
+          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'binary', run-on-presubmit: true }
```
mkruskal-google, line 178: Is there a shorter name that would work to limit the clutter a bit?

mkruskal-google:
Note: we should make sure the dashboard and copybara are ready for this before we merge.  You'll probably need to update the copybara configs atomically too

deannagarcia:
> Note: we should make sure the dashboard and copybara are ready for this before we merge. You'll probably need to update the copybara configs atomically too

Agreed and was planning on discussing more with you after approval to decide the best way to get this in. For copybara, I'm thinking it might be best to submit this as a CL so we can include the copybara config update with it.

deannagarcia (commented):
I still want to run more tests but figured I would get feedback on some of the more difficult questions so I can just run those once at the end.

mkruskal-google:
> > Note: we should make sure the dashboard and copybara are ready for this before we merge. You'll probably need to update the copybara configs atomically too
> 
> Agreed and was planning on discussing more with you after approval to decide the best way to get this in. For copybara, I'm thinking it might be best to submit this as a CL so we can include the copybara config update with it. Does that work?
> 
> And does anything specific need to be done for the dashboard?

Unfortunately copybara uses the config from HEAD for blocking tests, so you'll still have to force it in and won't be able to validate it until after submission.  However we should still probably do that atomically to avoid breaking people in the interim.

For the dashboard we probably have to tweak the GCP setup to filter using the new criteria instead of the old hardcoded list.  We might be able to get away with not updating the dashboard itself at all though

.github/workflows/test_cpp.yml
mkruskal-google: Makes sense, this is pretty minimal still

.github/workflows/test_cpp.yml
mkruskal-google: You could make it a string that you treat purely as a boolean.   An empty string will evaluate to false, and if nobody cares *what* the value is for "true" it works out fine

.github/workflows/test_runner.yml
mkruskal-google: Yea that's a good point.  I don't have any great ideas..  Maybe we could parse this file in a google3 test to validate this condition?  There must be an internal yaml parser somewhere..

.github/workflows/test_runner.yml
mkruskal-google: Yea I can't think of a case where a skip would be a failure, especially at the top level like this.  We could probably just ignore skipped test suites

.github/workflows/test_runner.yml
mkruskal-google: They actually could, maybe that would clean them up a bit.  These staleness tests are kindof a nightmare piece of tech debt I have no idea how to clean up :(. But on presubmit they run the same `blaze query` check N times, where N is the number of release branches.  They only checkout release branches and run tests on them during continuous runs

.github/workflows/test_upb.yml
```diff
@@ -171,10 +180,10 @@ jobs:
           # a single wheel. As a result we can just test the oldest and newest
           # supported Python versions and assume this gives us sufficient test
           # coverage.
-          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'binary' }
-          - { os: macos-12, python-version: "3.8", architecture: x64, type: 'binary' }
-          - { os: ubuntu-latest, python-version: "3.12", architecture: x64, type: 'binary' }
-          - { os: macos-13, python-version: "3.12", architecture: x64, type: 'binary' }
+          - { os: ubuntu-latest, python-version: "3.8", architecture: x64, type: 'binary', presubmit: true }
```
mkruskal-google, line 183: Can we flip the polarity of these?  I think it's safer to default to "always run" and have to explicitly opt out "continuous only" tests

.github/workflows/test_upb.yml
```diff
@@ -17,24 +26,26 @@ jobs:
...
         config:
-          - { name: "Bazel 7", bazel_version: "7.1.1" }
-          - { name: "Fastbuild" }
-          - { name: "Optimized", flags: "-c opt" }
-          - { name: "ASAN", flags: "--config=asan -c dbg", exclude-targets: "-//benchmarks:benchmark -//python/...", runner: ubuntu-20-large }
-          - { name: "UBSAN", flags: "--config=ubsan -c dbg", exclude-targets: "-//benchmarks:benchmark -//python/... -//lua/..." }
-          - { name: "32-bit", flags: "--copt=-m32 --linkopt=-m32", exclude-targets: "-//benchmarks:benchmark -//python/..." }
+          - { name: "Bazel 7", bazel_version: "7.1.1", presubmit: false }
+          - { name: "Fastbuild", presubmit: true }
+          - { name: "Optimized", flags: "-c opt", presubmit: false }
```
mkruskal-google, line 31: Similar comment as before, I think it's more concise to just omit the value in the "false" case (here and in other files).  Although as mentioned later, we should probably also flip the polarity to explicitly mark the more dangerous case

.github/workflows/test_objectivec.yml
```diff
@@ -27,24 +36,32 @@ jobs:
...
       DEVELOPER_DIR: /Applications/Xcode_14.1.app/Contents/Developer
     steps:
       - name: Checkout pending changes
+        if: ${{ !matrix.presubmit || inputs.test-type == 'continuous' }}
         uses: protocolbuffers/protobuf-ci/checkout@v3
         with:
           ref: ${{ inputs.safe-checkout }}
 
       - name: Setup ccache
+        if: ${{ matrix.presubmit || inputs.test-type == 'continuous' }}
```
mkruskal-google, line 57: My previous comment about parsing and validating these yaml files got me thinking.  We could probably enforce that every step has this line in it :). Similar thing with the naming scheme

mkruskal-google (commented):
This is looking great!  I think a yaml validation layer is a valuable follow-up to this though, as despite all the improvements this still looks a bit brittle to maintain (although I'm out of ideas to reduce duplication :P)

.github/workflows/test_runner.yml
```diff
@@ -93,9 +100,20 @@ jobs:
           ${{ github.event_name != 'pull_request_target' }} ||
           echo "sha=${{ github.event.pull_request.head.sha  }}"  >> $GITHUB_OUTPUT
 
+      - name: Set Test Type Variables
+        id: set-test-type-vars
+        run: |
+          if [ "${{ github.event_name }}" == 'pull_request' ] || [ "${{ github.event_name }}" == 'pull_request_target' ]; then
+            echo "test-type=presubmit" >> "$GITHUB_OUTPUT"
+            echo "continuous-prefix=[SKIPPED](Continuous)" >> "$GITHUB_OUTPUT"
```
mkruskal-google, line 108: nit: a space between these would make them easier to read

.github/workflows/test_runner.yml
deannagarcia: Since our tests are already language specific, the only way that this goes stale is if someone adds a new language to test. I think instead we should just document how to start tests for a language and include updating this. 

I agree it's not ideal I just don't think it's worth the work to parse this for the very few times I think we will get an addition here.

jskeet (approved):
LGTM for C#

.github/workflows/staleness_check.yml
```diff
@@ -20,15 +25,20 @@ jobs:
...
-    name: Test staleness ${{ matrix.os.name }} ${{ github.head_ref && 'PR' || matrix.branch }}
+    name: Test staleness ${{ matrix.os.name }} ${{ matrix.branch }}
     runs-on: ${{ matrix.os.value }}
     if: ${{ github.event.repository.full_name == 'protocolbuffers/protobuf' }}
     steps:
-      - name: Checkout ${{ github.head_ref && 'PR' || matrix.branch }}
+      - name: Checkout ${{ matrix.branch }}
         uses: protocolbuffers/protobuf-ci/checkout@v3
         with:
-          ref: ${{ inputs.safe-checkout || github.head_ref || matrix.branch }}
```
mkruskal-google, line 31: I think we need github.head_ref here.  Otherwise it'll test staleness on main for every PR.  The way this is setup now PRs only get `bazel query ...` run to make sure we don't submit invalid bazel files

.github/workflows/staleness_check.yml
```diff
@@ -20,15 +25,20 @@ jobs:
       matrix:
         branch: [main, 25.x, 27.x, 28.x]
         os: [{ name: Linux, value: ubuntu-latest}]
+        exclude:
+          # If we are in a presubmit run, only test main
+          - branch: ${{ !inputs.continuous-run && '25.x' }}
```
mkruskal-google, line 30: Can we flip this to make it more maintainable?

.github/workflows/test_ruby.yml
```diff
@@ -20,22 +30,24 @@ jobs:
...
-          - { name: Ruby 3.2, ruby: ruby-3.2.0 }
+          - { name: Ruby 3.0, ruby: ruby-3.0.2, ffi: FFI, continuous-only: true }
+          - { name: Ruby 3.1, ruby: ruby-3.1.0, continuous-only: true }
+          - { name: Ruby 3.2, ruby: ruby-3.2.0, continuous-only: true }
           - { name: Ruby 3.3, ruby: ruby-3.3.0, ffi: NATIVE }
-          - { name: Ruby 3.3, ruby: ruby-3.3.0, ffi: FFI }
+          - { name: Ruby 3.3, ruby: ruby-3.3.0, ffi: FFI, continuous-only: true }
           - { name: JRuby 9.4, ruby: jruby-9.4.6.0, ffi: NATIVE }
-          - { name: JRuby 9.4, ruby: jruby-9.4.6.0, ffi: FFI }
+          - { name: JRuby 9.4, ruby: jruby-9.4.6.0, ffi: FFI, continuous-only: true }
```
JasonLunn, line 39: I'd like one FFI test configuration kept as presubmit blocking on each interpreter. I'm fine if that's on the highest version number we support only, as that will be the newest and as they age out the likelihood of introducing a breakage goes down.

.github/workflows/staleness_check.yml
deannagarcia: That's what I originally wanted, but the include syntax unfortunately doesn't work the same as the exclude so I can't have a condition like this. All the other options I considered seemed more confusing than this one, but I'm definitely open to other ideas!

.github/workflows/test_ruby.yml
deannagarcia: Fixed here and made one FFI test run on presubmit for macos as well.

.github/workflows/staleness_check.yml
deannagarcia: I don't really understand that. This variable is just for the checkout step, right? Why do we need to pass the string "Pull Request" to that instead of the branch that we want to check out?

.github/workflows/test_objectivec.yml
deannagarcia: Discussed offline, yml parsing will be a follow up project.

.github/workflows/staleness_check.yml
mkruskal-google: What if we removed `main` from the `branch` list and just exclude *all* of them on presubmits.  We should be able to add `main` back in with `include:` then to always run

.github/workflows/staleness_check.yml
mkruskal-google: Ah yea I was mixing this up with the behavior below, but I think this is still wrong.    Your changes to `name:` should be fine, since you've also removed the confusing duplication (that's why I made them all say "PR").  However, we still need to check out `github.head_ref` when it's set (it only gets set for pull requests).  Otherwise we'd be testing against main instead of pending changes

.github/workflows/staleness_check.yml
deannagarcia: GHA doesn't support syntax to exclude all branches unless we explicitly list them, which leaves us in the same spot.

.github/workflows/staleness_check.yml
deannagarcia: Got it, reverted this change.

.github/workflows/staleness_check.yml
```diff
@@ -20,12 +25,17 @@ jobs:
...
+          - branch: ${{ !inputs.continuous-run && '27.x' }}
+          - branch: ${{ !inputs.continuous-run && '28.x' }}
 
-    name: Test staleness ${{ matrix.os.name }} ${{ github.head_ref && 'PR' || matrix.branch }}
+    name: Test staleness ${{ matrix.os.name }} ${{ matrix.branch }}
     runs-on: ${{ matrix.os.value }}
     if: ${{ github.event.repository.full_name == 'protocolbuffers/protobuf' }}
     steps:
-      - name: Checkout ${{ github.head_ref && 'PR' || matrix.branch }}
```
mkruskal-google, line 28: I think the old naming scheme was a little clearer.  For PR's we're checking out the changes not "main"

.github/workflows/test_cpp.yml
```diff
@@ -395,20 +430,23 @@ jobs:
...
             cache-prefix: windows-2022-win32-cmake
+            continuous-only: true
           - name: Windows CMake Shared
             os: windows-2022
             flags: >-
               -G Ninja -Dprotobuf_WITH_ZLIB=OFF -Dprotobuf_BUILD_CONFORMANCE=OFF
               -Dprotobuf_BUILD_SHARED_LIBS=ON
             vsversion: '2022'
             cache-prefix: windows-2022-cmake
+            continuous-only: false
```
mkruskal-google, line 449: We should probably remove all "continuous-only: false" just to minimize clutter

.github/workflows/test_php.yml
```diff
@@ -22,37 +32,45 @@ jobs:
             version: "8.1.14"
             version-short: "8.1"
             command: composer test \&\& composer test_c
+            presubmit: true
```
mkruskal-google, line 35: We should keep this consistent across test files and use `continuous-only`

.github/workflows/staleness_check.yml
deannagarcia: Changed back here and in the naming above.

.github/workflows/test_cpp.yml
deannagarcia: Deleted here and in all other files.

.github/workflows/test_php.yml
deannagarcia: Definitely agreed. Not sure how I missed this file, but it's now updated to `continuous-only`.

PR #17212, opened 2024-06-21
Fix various unsigned to signed comparison warnings.

benjaminp:
GCC aggressively emit warnings when comparing unsigned and signed integer types, which causes failures under protobuf's -Werror default. We can fix these often by switching to iterators, but sometimes it's easiest to add a cast or switch a variable type.

### DIFF
--- a/src/google/protobuf/compiler/java/full/enum.cc
+++ b/src/google/protobuf/compiler/java/full/enum.cc
@@ -82,14 +82,13 @@ void EnumNonLiteGenerator::Generate(io::Printer* printer) {
     }
   }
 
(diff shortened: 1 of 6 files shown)


tonyliaoss:
Hi Benjamin,

Apologies for getting to this PR relatively late.

We're okay with accepting the range-based for loop changes. For anything else, we're probably not going to hold ourselves to maintain signed/unsigned casting -- internally we build with those warnings turned off.

benjaminp:
Okay, I've rebased and removed any lines that added `static_cast<size_t>`.

tonyliaoss (approved):

PR #17225, opened 2024-06-24
Fix windows 2019 staleness issue

mkruskal-google:
Bazel on windows doesn't sandbox very well, and our old strategy for bootstrap staleness test was causing some non-deterministic failures.  Before, we were creating multiple genrules copying the same batch of files, but only marking a specific one as the output of each.  That works fine with proper sandboxing, but on windows can cause conflicts.  With this change, each genrule limits its scope to only copy the 1 file it's meant to generate


zhangskz (approved):

zhangskz:
Can you update the PR description to mention the windows sandboxing and making the copy mutually exclusive for future reference?

PR #17290, opened 2024-07-01
fix pkg-config protobuf library not dependes from utf8_range

mdionisio:
the .pc file generated insert a dependency to a utf8_range.pc file that is not required

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -7,7 +7,6 @@ foreach(_target IN LISTS protobuf_ABSL_USED_TARGETS)
   string(REPLACE :: _ _modified_target ${_modified_target})
   list(APPEND _pc_targets ${_modified_target})
 endforeach()
-list(APPEND _pc_targets "utf8_range")
 
 set(_protobuf_PC_REQUIRES "")
 set(_sep "")
--- a/cmake/libprotobuf.cmake
+++ b/cmake/libprotobuf.cmake
@@ -49,3 +49,4 @@ set_target_properties(libprotobuf PROPERTIES
 add_library(protobuf::libprotobuf ALIAS libprotobuf)
 
 target_link_libraries(libprotobuf PRIVATE utf8_validity)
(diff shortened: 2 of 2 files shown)


acozzette:
@mdionisio If I'm not mistaken, we do need this because we rely on utf8_range, like here for example: https://github.com/protocolbuffers/protobuf/blob/47f633eae687939fbb68af4661503685a8efd239/src/google/protobuf/generated_message_tctable_lite.cc#L1395

mdionisio:
@acozzette I don't say that utf8_range is not used inside protobuf, I say only that If I use protobuf I don't neet to link my program with utf8_range because all the dependency to utf8_range was statically linked directly inside protobuf library. My change is only on ".pc" that is not used from protobuf but it is used from code that use protobuf.

I test the fix on protobuf 4.25 in yocto environment building and testing code that only used protobuf library and in more complicated example that use grpc.

mdionisio:
@acozzette you are probably right. There is also another issue in cmake that doesn't configure protobuf to explicitly link utf8_range. I don't understand exactly why utf8_range was linked anyway.

I update my pull request to fix also that.

acozzette:
@mdionisio Maybe I'm missing something but I think this is still not going to work because it will cause `utf8_range` to be removed from the .pc file even though it's a required dependency. With this change, I find that `pkgconfig --libs protobuf` no longer includes the linker flag `-lutf8_validity`, so I would expect this to break builds.

mdionisio:
@acozzette in my opinion the problem is that `utf8_range` and `utf8_validity` are configured in .pc to be a required dependency but is not true. In any of .so library there is any required symbol from `utf8_range` or `utf8_validity`. In any case I will double check the yocto recipes that build protobuf for me if there is any strange trick for that.

zhangskz:
It sounds like this PR may indeed by inactive/obsolete based on the prior comment thread. Marking this as inactive again accordingly, which will cause it to get closed in another ~14 days per the above comment. Please comment / reopen as needed if this change is still needed.

PR #17323, opened 2024-07-03
java: Use `!=` to compare oneof enums in `equals`

Juneezee:
This PR updates the `GenerateEqualsAndHashCode` function to use `!=` for comparing oneof enums instead of the `!equals()` method. This follows the recommendation by static analysis tools like SonarSource [^1] and IntelliJ IDEA [^2].

In addition, Java's enum `equals()` implementation itself uses `==` internally to compare enum values, see https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.base/share/classes/java/lang/Enum.java#L154-L164.

Before:

```java
if (!getChoiceCase().equals(other.getChoiceCase())) return false;
```

After:

```java
if (getChoiceCase() != other.getChoiceCase()) return false;
```

[^1]: https://rules.sonarsource.com/java/RSPEC-4551/
[^2]: https://www.jetbrains.com/help/inspectopedia/EqualsCalledOnEnumConstant.html

### DIFF
--- a/src/google/protobuf/compiler/java/full/message.cc
+++ b/src/google/protobuf/compiler/java/full/message.cc
@@ -993,8 +993,8 @@ void ImmutableMessageGenerator::GenerateEqualsAndHashCode(
   for (auto& kv : oneofs_) {
     const OneofDescriptor* oneof = kv.second;
     printer->Print(
-        "if (!get$oneof_capitalized_name$Case().equals("
-        "other.get$oneof_capitalized_name$Case())) return false;\n",
+        "if (get$oneof_capitalized_name$Case() != "
+        "other.get$oneof_capitalized_name$Case()) return false;\n",
         "oneof_capitalized_name",
         context_->GetOneofGeneratorInfo(oneof)->capitalized_name);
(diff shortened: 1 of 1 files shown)


zhangskz:
Thanks for the contribution! Though == does have some benefits, unfortunately Google has conflicting guidance internally to prefer `equals` over `==`

Juneezee:
> Thanks for the contribution! Though == does have some benefits, unfortunately Google has conflicting guidance internally to prefer `equals` over `==`

@zhangskz No worries. Thanks for taking the time to review this PR :smile: . Sorry I wasn't aware of the internal preference because I see that both `.equals()` and `==` are used in the codebase.

https://github.com/protocolbuffers/protobuf/blob/fba8024963b8884cfa0d3903b3b0cbed27587f02/java/core/src/main/java/com/google/protobuf/MessageSchema.java#L473

https://github.com/protocolbuffers/protobuf/blob/fba8024963b8884cfa0d3903b3b0cbed27587f02/java/core/src/main/java/com/google/protobuf/MessageSchema.java#L519

Issue #12932, opened 2023-05-30
Some symbols in `libprotobuf`'s public API are hidden


coryan:
**What version of protobuf and what language are you using?**

Version: 23.2
Language: C++

**What operating system (Linux, Windows, ...) and version?**

Linux, Fedora 37.  Probably applies to all Linux variants and 

**What runtime / compiler are you using (e.g., python version or gcc version)**

GCC 12.2

**What did you do?**

This was motivated by https://github.com/conda-forge/protobuf-feedstock/pull/191#issuecomment-1569020333.  It seems the "c++ backend" for Python is not loadable when compiled as a .so.  Maybe because some symbols are hidden:

https://github.com/conda-forge/protobuf-feedstock/pull/191#issuecomment-1569108909

**What did you expect to see**

I expect that public symbols are not hidden.  `google::protobuf::io::SafeDoubleToFloat(double)` seems like a public symbol.  I would have expected it to be exported.

**What did you see instead?**

The symbol is hidden:

```
nm /usr/local/lib64/libprotobuf.so.23.2.0 | c++filt | grep SafeDoubleToFloat
000000000018f350 t google::protobuf::io::SafeDoubleToFloat(double)
```
**Anything else we should know about your project / environment**

N/A

This reminds me of: #12618

h-vetinari:
Thanks for opening this issue @coryan. One other aspect about having the protobuf-bindings built on top of shared libs is that it's missing the now-required linkage to abseil. In particular, our link check infrastructure shows (at least) `absl_spinlock_wait` and `absl_strings` should be there:
```
   INFO (protobuf,lib/python3.10/site-packages/google/protobuf/pyext/_message.cpython-310-x86_64-linux-gnu.so): Needed DSO lib/libprotobuf.so.23.2.0 found in conda-forge::libprotobuf-4.23.2-hd1fb520_1
   INFO (protobuf,lib/python3.10/site-packages/google/protobuf/pyext/_message.cpython-310-x86_64-linux-gnu.so): Needed DSO lib/libabsl_spinlock_wait.so.2301.0.0 found in conda-forge::libabseil-20230125.2-cxx17_h59595ed_2
   INFO (protobuf,lib/python3.10/site-packages/google/protobuf/pyext/_message.cpython-310-x86_64-linux-gnu.so): Needed DSO lib/libabsl_strings.so.2301.0.0 found in conda-forge::libabseil-20230125.2-cxx17_h59595ed_2
```

[...]

PR #17378, opened 2024-07-09
export EnumName for c++

zwkno1:


zhangskz:
Is this for the same issue as https://github.com/protocolbuffers/protobuf/issues/12932?

zhangskz (approved):

zwkno1:
> Is this for the same issue as #12932?

mybe not. i was writing a protoc plugin  and want to use this function.

zhangskz:
Apologies, looks like this got stuck behind a failing test after approval and the logs don't seem to be present anymore (not sure if legitimate failure or a flake).
 
I've triggered a rerun but I suspect this change will need to be rebased or manually imported.

zhangskz:
You should be able to use ClassName(const EnumDescriptor* enum_descriptor) instead for enums: https://github.com/protocolbuffers/protobuf/blob/e26cb4864db312fce543f2a58b7822f090428124/src/google/protobuf/compiler/cpp/helpers.cc#L412

We don't actually have a separate EnumName function to expose here.

PR #17433, opened 2024-07-12
Improve support for Clang inside of UPB

Vadmeme:
After using UPB I noticed that field accessors were not getting inlined properly using Clang. The source appears to be the UPB_FORCEINLINE macro falling back to just being `static` when underneath clang despite clang having full support of the required attributes.

### DIFF
--- a/upb/port/def.inc
+++ b/upb/port/def.inc
@@ -116,7 +116,7 @@ Error, UINTPTR_MAX is undefined
 #endif
 
 // Macros for function attributes on compilers that support them.
-#ifdef __GNUC__
+#if defined(__GNUC__) || defined(__clang__)
 #define UPB_FORCEINLINE __inline__ __attribute__((always_inline)) static
 #define UPB_NOINLINE __attribute__((noinline))
 #define UPB_NORETURN __attribute__((__noreturn__))


upb/port/def.inc
```diff
@@ -116,7 +116,7 @@ Error, UINTPTR_MAX is undefined
 #endif
 
 // Macros for function attributes on compilers that support them.
-#ifdef __GNUC__
+#if defined(__GNUC__) || defined(__clang__)
```
haberman, line 119: I am pretty sure that Clang defines `__GNUC__`, so I think this PR is a no-op: https://godbolt.org/z/bcMTaWc57

Vadmeme: It used to but in recent versions it has stopped. https://clang.llvm.org/docs/UsersManual.html 

```
-fgnuc-version=
This flag controls the value of __GNUC__ and related macros. This flag does not enable or disable any GCC extensions implemented in Clang. Setting the version to zero causes Clang to leave __GNUC__ and other GNU-namespaced macros, such as __GXX_WEAK__, undefined.
```

On a default windows-x64 configuration with clang from MSVC build tools

```
clang version 17.0.3
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin
```

```cpp
// clang test.cpp -o test.exe && test

#include <iostream>

int main() {
#ifdef __GNUC__
std::cout << "Gnu!" << std::endl;
#else
std::cout << "Not Gnu!" << std::endl;
#endif

return 0;
}
```

Outputs `Not Gnu!`

Vadmeme: Running as `clang test.cpp -o test.exe -fgnuc-version=1234 && test` does resolve this and the test program properly prints `Gnu!` however having to add an option specific to make UPB generate optimal code feels rather hacky

haberman (approved):

Vadmeme:
Any updates on this?

tonyliaoss:
I saw some CI failures but they look unrelated to the CL.

Just did a force-pull with Copybara and the CI tests looked good: https://github.com/protocolbuffers/protobuf/pull/19358

PR #17513, opened 2024-07-17
Add breaking change detection for well-known types

jchadwick-buf:
Adds new tests in `//compatibility` for detecting undesired breaking changes in the schema of the well-known types and `descriptor.proto` files, using the Buf breaking change detector, via [`rules_buf`](https://github.com/bufbuild/rules_buf/).

In order to keep things light-touch as far as maintenance goes, I have chosen to keep the integration as small and simple as possible. Some notes:

- Breaking change behavior can be granularly controlled via [`buf.yaml`](https://buf.build/docs/configuration/v1/buf-yaml#breaking).

- Bazel sandboxes us away from meaningful VCS information, so in order to pick a target to check for breaking changes against, a new variable is added to `protobuf_versions.bzl` that needs to be updated with changes in the release version.

- Breaking change detection is performed on a file-level, not a package-level, so migrating types across WKT files would constitute a breaking change. If this is not desired the behavior can be made to work on a package-level, though we need to do some more work as `buf_breaking_test` currently only accepts a single file descriptor set target for the `against` attribute.


compatibility/BUILD.bazel
```diff
@@ -17,3 +18,98 @@ java_runtime_conformance(
...
+
+buf_breaking_test(
+    name = "api_proto_breaking",
+    targets = ["//:api_proto"],
+    config = ":buf.yaml",
+    against = "@com_google_protobuf_previous_release//:api_proto",
+)
+
+buf_breaking_test(
+    name = "descriptor_proto_breaking",
```
mkruskal-google, line 38: We should add plugin.proto, cpp_features.proto, and java_features.proto too

protobuf_version.bzl
```diff
@@ -4,3 +4,4 @@ PROTOBUF_JAVA_VERSION = "4.29.0"
 PROTOBUF_PYTHON_VERSION = "5.29.0"
 PROTOBUF_PHP_VERSION = "4.29.0"
 PROTOBUF_RUBY_VERSION = "4.29.0"
+PROTOBUF_PREVIOUS_RELEASE = "28.0-rc1"
```
mkruskal-google, line 7: This should probably be 27.0, but I'm not sure what's best for our update script @zhangskz

jchadwick-buf: Do you mean 27.2, or specifically 27.0? 

The way I see it, setting the "previous release" value too new is never really a big problem; it will just lead to catching breaking changes for things that may have not hit a stable release yet. Setting it too old, though, is a problem, since then it can miss changes that would break from, in this case, anything added in 27.1 and 27.2.

compatibility/BUILD.bazel
jchadwick-buf: Thanks, should be added now.

protobuf_version.bzl
zhangskz: 28.0-rc1 seems correct here. For our update script, the limitation is that we don't want to have to update this for other branches when cutting releases (e.g. we don't want to update previous versions on main/28.x when releasing 27.2). PROTOBUF_PREVIOUS_RELEASE should stay on the prior release *when the branch was cut* (aka 28.0-rc1).

Once 29.x is cut with this state I imagine PROTOBUF_PREVIOUS_RELEASE can either stay on the first 28.0-rc1 release, or get upgraded as we make further releases on 29.x, depending on what is more valuable.

PROTOC_RELEASE = 29.0 (dev)
PROTOBUF_PREVIOUS_RELEASE =  28.0-rc1
  |
  | 29.0-rc1 release (releases to 28.x do not change PROTOBUF_PREVIOUS_RELEASE for main or 29.x)
 V
PROTOC_RELEASE = 29.0 (dev)
PROTOBUF_PREVIOUS_RELEASE = 29.0-rc1
 |
 | 29.0 release 
V
PROTOC_RELEASE = 29.1 (dev)
PROTOBUF_PREVIOUS_RELEASE = 29.0

mkruskal-google: Ah yea that makes sense, and we really shouldn't be introducing new things to these protos in a patch release anyway

WORKSPACE
```diff
@@ -254,3 +254,27 @@ http_archive(
...
+load("//:protobuf_version.bzl", "PROTOBUF_PREVIOUS_RELEASE")
+
+http_archive(
+    name = "com_google_protobuf_previous_release",
+    strip_prefix = "protobuf-" + PROTOBUF_PREVIOUS_RELEASE,
+    url = "https://github.com/protocolbuffers/protobuf/releases/download/v{0}/protobuf-{0}.tar.gz".format(PROTOBUF_PREVIOUS_RELEASE),
+)
+
+http_archive(
+    name = "rules_buf",
```
mkruskal-google, line 268: One of our bzlmod tests are failing, I think you need to add a dev_dependency to our MODULE file

protobuf_version.bzl
zhangskz: Alternatively, we could try something like the following which I was looking into for gencode runtime compat tests, but I'm not sure if this might hit problems w caching.

http_archive(
    name = "com_google_protobuf_previous_release",
    strip_prefix = "protobuf-28.x",
    url = "https://github.com/protocolbuffers/protobuf/archive/28.x.zip",
)

mkruskal-google: I like that, then it'll be checking per-commit too.  It shouldn't be a problem for caching since the release branches don't get updated that frequently, but I think we'd need to be careful to update the branch *before* cutting a new one.  That way main is always pointing at the latest branch and each branch is pointing back at itself

jchadwick-buf: Regarding the alternative approach of using the release branches:

I think this works. It might run into caching problems, it's not 100% clear; I know `http_archive` is implemented in terms of [`repository_ctx.download_and_extract`](https://bazel.build/rules/lib/builtins/repository_ctx#download_and_extract) which says:

> If provided, the repository cache will first be checked for a file with the given hash; a download will only be attempted if the file was not found in the cache. After a successful download, the file will be added to the cache.

This doesn't tell us what the behavior is if the hash _isn't_ specified, though. Experimentally, I can at least confirm that it doesn't just simply always download, which I think leaves two logical possibilities: it is able to check for staleness by using standard HTTP caching techniques, or it is simply not bothering to check and treating the repository as up-to-date so as long as it is populated. I don't know enough about Bazel internals to know which is more likely.

---

With that having been said, I am totally happy to implement it this way if it is desired, but I do personally have some doubts as to whether or not it is the overall best approach. It will mean that retroactively, running tests on old versions from a given release branch would fail whenever a normal forward-compatible change happens, because you would always be testing against the latest release branch version. Adding a field in a later version and then running `bazel test //compatibility:proto_breaking` would see the field as having been removed, which seems suboptimal.

Outside of Bazel, this normally _would_ be the approach that you'd take to test, that is, testing against the branch that changes would be merged into, but in those cases it is typically done as part of a change request CI test (e.g. GitHub actions PR workflows) so it doesn't have the same concerns about reproducibility; those CI tests don't need to pass later, they are _inherently_ only relevant during the review of a change set. I guess it comes down to sensibility for as to whether or not this should be likewise considered to only be relevant in similar conditions, but my feeling is that it feels like a bit of a smell if `bazel test //...` could start failing spontaneously on old commits.

On the flip-side, though, more granular checking does make it easier to deal with intentional breaking changes, which is part of why it is the "defacto" way of using buf breaking.

WORKSPACE
jchadwick-buf: Sorry about that, I've updated `MODULE.bazel` and `WORKSPACE.bzlmod` such that the proto breaking tests should work with bzlmod enabled or disabled. I'm pretty new to bzlmod, let me know if I need to do anything else here.

protobuf_version.bzl
mkruskal-google: O I missed the hashing issue, yea that sounds like it would be problem (in addition to the caching concern we'd also have warnings and security issues).  The coarser-grained approach seems simpler for now then

mkruskal-google (approved):

protobuf_version.bzl
zhangskz: Setting to 28.0-rc1 sgtm given the hash and possible cache issues. i think there isn't too much benefit between testing 29.x against 28.0-rc1 versus later patches.

zhangskz (approved):

PR #17530, opened 2024-07-18
Introduce FieldDescriptor::cpp_string_type() API to replace direct ct…

mkruskal-google:
…ype inspection which will be removed in the next breaking change

This should provide the roughly same result as ctype, except that it reflects actual behavior rather than the specification in the proto file.  VIEW will now be visible, and some subtleties around CORD and PIECE will change.

PiperOrigin-RevId: 653677655


mkruskal-google:
Redundant with https://github.com/protocolbuffers/protobuf/pull/17789

PR #17587, opened 2024-07-24
CherryPick the newer work for Objective-C over the the 28 branch.

thomasvl:


zhangskz (approved):

conformance/failure_list_objc.txt
```diff
@@ -1,7 +1,2 @@
 # JSON input or output tests are skipped (in conformance_objc.m) as mobile
 # platforms don't support JSON wire format to avoid code bloat.
-
-Required.Editions_Proto2.ProtobufInput.UnknownOrdering.ProtobufOutput
```
zhangskz, line 4: Users are not likely to be dependent on previous unknown field ordering before this fix right?

thomasvl: Failing the test is the whole reason for the change, objc was out of spec.

And i think in general we tell folks not to rely on binary orderings because some languages randomize order.

zhangskz: Yep agreed, I think its fine to change this in 28.x accordingly, thanks for confirming!

Issue #13428, opened 2023-08-01
PHP: Deprecation message for "return_immediately" is thrown which is impossible to fix


bshaffer:
See https://github.com/googleapis/google-cloud-php/issues/5350

@pkruithof has reported the following, which seems to be an error in the Protobuf package. The  function `getReturnImmediately` is being called internally without any way (as far as we can tell) to prevent it from happening:

Because the `return_immediately` option has been deprecated some time ago, this produces a deprecation log on every pull request:

```
[2022-06-17T06:57:33.459907+00:00] deprecation.INFO: User Deprecated: return_immediately is deprecated. {"exception":"[object] (ErrorException(code: 0): User Deprecated: return_immediately is deprecated. at /app/vendor/google/cloud-pubsub/src/V1/PullRequest.php:138)"} []
[2022-06-17T06:57:33.460273+00:00] deprecation.INFO: User Deprecated: return_immediately is deprecated. {"exception":"[object] (ErrorException(code: 0): User Deprecated: return_immediately is deprecated. at /app/vendor/google/cloud-pubsub/src/V1/PullRequest.php:118)"} []
[2022-06-17T06:57:33.460353+00:00] deprecation.INFO: User Deprecated: return_immediately is deprecated. {"exception":"[object] (ErrorException(code: 0): User Deprecated: return_immediately is deprecated. at /app/vendor/google/cloud-pubsub/src/V1/PullRequest.php:118)"} []
```

At first I thought to submit a workaround for this, so that the deprecation is only shown when the option is set to `true` (`false` is the default):

```php
    public function getReturnImmediately()
    {
        if ($this->return_immediately === true) {
            @trigger_error('return_immediately is deprecated.', E_USER_DEPRECATED);
        }
        return $this->return_immediately;
    }
```

The setter should still always trigger the deprecation, so you still see it when you use the option and set it to `false`.

However, then I noticed that this code is generated. So I don't think this is possible, unless there's a way to customize generated classes like this?

In any case, I would very much like to be able to remove this deprecation from our processes, as it's polluting our logs, making them harder to read. Is there another way to go here?

Here is the stack trace

```
PHP Fatal error:  Uncaught ErrorException: return_immediately is deprecated. in /app/vendor/google/cloud-pubsub/src/V1/PullRequest.php:167
Stack trace:
#0 [internal function]: {closure}(16384, 'return_immediat...', '/app/vendor/goo...', 167)
[...]

PR #17607, opened 2024-07-25
Deprecation in getter is triggered only when field is actually used

pkruithof:
See #13428

This is my first commit here, having no experience with this project or language. I think I got it right, but I cannot compile `protoc` to actually test it. Nor do I know if there's an automated test here I need to write/improve. So any help is appreciated to get this through.

This is the build error I'm getting:

```
❯ bazel build :protoc
INFO: Analyzed target //:protoc (0 packages loaded, 0 targets configured).
ERROR: /Users/peter/projects/protobuf/src/google/protobuf/io/BUILD.bazel:98:11: Compiling src/google/protobuf/io/printer.cc failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target //src/google/protobuf/io:printer) external/local_config_cc/cc_wrapper.sh -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics ... (remaining 42 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from src/google/protobuf/io/printer.cc:12:
bazel-out/darwin_arm64-fastbuild/bin/src/google/protobuf/io/_virtual_includes/printer/google/protobuf/io/printer.h:918:19: error: 'get<std::function<bool ()>, std::string, std::function<bool ()>>' is unavailable: introduced in macOS 10.13
    value = absl::get<Callback>(that.value);
                  ^
bazel-out/darwin_arm64-fastbuild/bin/src/google/protobuf/io/_virtual_includes/printer/google/protobuf/io/printer.h:863:11: note: in instantiation of function template specialization 'google::protobuf::io::Printer::ValueImpl<false>::operator=<true>' requested here
    *this = that;
          ^
bazel-out/darwin_arm64-fastbuild/bin/src/google/protobuf/io/_virtual_includes/printer/google/protobuf/io/printer.h:1150:12: note: in instantiation of function template specialization 'google::protobuf::io::Printer::ValueImpl<false>::ValueImpl<true>' requested here
    return ValueView(it->second);
           ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/variant:1581:22: note: 'get<std::function<bool ()>, std::string, std::function<bool ()>>' has been explicitly marked unavailable here
constexpr const _Tp& get(const variant<_Types...>& __v) {
```

Looks like it's the same as in #16944 but I don't know how to apply the workaround for it.

/cc @bshaffer

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -575,6 +575,9 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
                          " is deprecated.', E_USER_DEPRECATED);\n        ")
           : "";
 
+  std::string deprecation_trigger_getter =
+      "if ($this->^name^ !== null) {\n    ^deprecation_trigger^\n}";
+
   // Emit getter.
   if (oneof != nullptr) {
     printer->Print(
@@ -584,7 +587,7 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
         "}\n\n",
         "camel_name", UnderscoresToCamelCase(field->name(), true), "number",
         IntToString(field->number()), "deprecation_trigger",
-        deprecation_trigger);
+        deprecation_trigger_getter);
   } else if (field->has_presence() && !field->message_type()) {
     printer->Print(
         "public function get^camel_name^()\n"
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -575,6 +575,9 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
                          " is deprecated.', E_USER_DEPRECATED);\n        ")
           : "";
 
+  std::string deprecation_trigger_getter =
+      "if ($this->^name^ !== null) {\n    ^deprecation_trigger^\n}";
+
```
bshaffer, line 580: This looks good, but I'd just update `deprecation_trigger` to include the conditional, rather than create a separate variable

bshaffer (commented):
Thanks for your contribution! 

I imagine you'll also need to update the tests for them to pass.

bshaffer:
I've written https://github.com/protocolbuffers/protobuf/pull/17788 to address the issues in this PR

Issue #17665, opened 2024-07-30
MSVC ARM64 Compilation Error Due to Unsupported Inline Assembly


dsmith111:
**What version of protobuf and what language are you using?**
Version: main
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Windows 11, ARM64

**What runtime / compiler are you using (e.g., python version or gcc version)**
MSVC
Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33139

**What did you do?**
1. Build with MSVC for a Windows ARM64 device.

**What did you expect to see**
Successful build.

**What did you see instead?**
`google\protobuf\parse_context.h(721,10): error C2059: syntax error: ':'`

`google\protobuf\parse_context.h(898,14): error C3861: 'Ubfx7': identifier not found`

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/parse_context.h
Line 638 for potential patch.
Lines 721, 728 for breaking inline assembly.

I have noticed that other usages of inline assembly occur when either x86 architecture is defined, or when __GNUC__ is defined. This line seems to be the only case that does not follow this. I'm not sure if this was intentional, or if `#if defined(__aarch64__) && defined(__GNUC__)` just happened to be missed during the addition. Could also use `!defined(_MSC_VER)`, if it was intentional for clang and gnu.

Additionally, ARM-specific assembly instructions, such as "UBFX" also fails with MSVC.

dsmith111:
Fix merged in here: https://github.com/protocolbuffers/protobuf/commit/c5f623184dc91e5c5bc58adb91b93f24e4467282

PR #17671, opened 2024-07-30
Add Compiler Condition to use inline assembly optimizations with ARM64 for Compatibility with MSVC

dsmith111:
https://github.com/protocolbuffers/protobuf/issues/17665

**Problem**
`google\protobuf\parse_context.h(721,10): error C2059: syntax error: ':'`
`google\protobuf\parse_context.h(898,14): error C3861: 'Ubfx7': identifier not found`

The latest protobuf versions fail to compile with MSVC on Windows ARM64 architecture.

> Inline assembly is only available for x86 targets. For similar functionality in x64 or ARM64 code, use [compiler intrinsics](https://learn.microsoft.com/en-us/cpp/intrinsics/compiler-intrinsics?view=msvc-170).

https://learn.microsoft.com/en-us/cpp/assembler/inline/writing-functions-with-inline-assembly?view=msvc-170

**Solution**
As inline assembly for win ARM64 with MSVC as a compiler is not supported (even ARM specific assembly functions), a term has been added to the conditions in parse_context.h to prevent the assembly optimizations for arm64 from executing if the compiler is MSVC.

Tested by compiling protobuf with MSVC on Windows ARM64 and AMD64 architecture. Compilation was a success.

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -866,7 +866,7 @@ static const char* VarintParseSlowArm(const char* p, uint64_t* out,
 // The caller must ensure that p points to at least 10 valid bytes.
 template <typename T>
 PROTOBUF_NODISCARD const char* VarintParse(const char* p, T* out) {
-#if defined(__aarch64__) && defined(ABSL_IS_LITTLE_ENDIAN)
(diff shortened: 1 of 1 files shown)


dsmith111:
Hey @fowles, just a habit when contributing to open-source, want to make sure you're still an active maintainer here!

acozzette (approved):

PR #17714, opened 2024-08-04
make hasOptionalKeyword function public

mscheong01:
Hi, I am currently migrating my third party Kotlin protobuf code generation libaray, krotoDC from protobuf v3 to v4.
However, I noticed that `FieldDescriptor::hasOptionalKeyword()` has become package-private. My library relies on this api to check if the field is private or not. can we please make this function public again?

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/Descriptors.java
+++ b/java/core/src/main/java/com/google/protobuf/Descriptors.java
@@ -1443,7 +1443,7 @@ public final class Descriptors {
      * Returns true if this field was syntactically written with "optional" in the .proto file.
      * Excludes singular proto3 fields that do not have a label.
      */
-    boolean hasOptionalKeyword() {
+    public boolean hasOptionalKeyword() {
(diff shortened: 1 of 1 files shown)


mscheong01:
ref: https://github.com/mscheong01/krotoDC/issues/32

zhangskz:
hasOptionalKeyword() was intentionally removed as part of https://github.com/protocolbuffers/protobuf/commit/1aeacd4f4eb4e0aa05d6336e2988a565e475e9a0 since these API's are proto2/proto3 specific and break under Editions (https://protobuf.dev/editions/overview/).

Most users of hasOptionalKeyword() likely want to move to use FieldDescriptor.hasPresence() instead. Can you elaborate on what you mean by "to check if the field is private or not"?

zhangskz:
Closing per my prior response. Feel free to open a new Github issue if you'd like further assistance with how to proceed, without a public hasOptionalKeyword() API.

PR #17788, opened 2024-08-12
fix: do not throw deprecated warning on field getters for default values

bshaffer:
fixes https://github.com/protocolbuffers/protobuf/issues/13428

Wraps deprecated field getters in a conditional to see if the value has been set before throwing the warning. This is because this method is used internally, so users can get deprecated warnings even when they are not using the field.


src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -642,7 +654,8 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
         "$this->readWrapperValue(\"^field_name^\");\n"
         "}\n\n",
         "camel_name", UnderscoresToCamelCase(field->name(), true), "field_name",
-        field->name(), "deprecation_trigger", deprecation_trigger);
+        field->name(), "deprecation_trigger",
+        deprecation_trigger_with_conditional);
```
bshaffer, line 658: the field getter here may throw an error for wrapped fields since we aren't calling `readWrapperValue`

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -584,7 +595,7 @@ void GenerateFieldAccessor(const FieldDescriptor* field, const Options& options,
         "}\n\n",
         "camel_name", UnderscoresToCamelCase(field->name(), true), "number",
         IntToString(field->number()), "deprecation_trigger",
-        deprecation_trigger);
+        deprecation_trigger_with_conditional);
```
bshaffer, line 598: The field getter in the deprecation trigger conditional here may throw an error since we're not calling `readOneof`

bshaffer (commented):
there's enough complexity around getting the field to even check if it's default... for instance, `$this->readOneof` should be called instead of the property name for "OneOf fields", and `$this->readWrapperValue` should be called for "wrapped" fields

src/google/protobuf/compiler/php/php_generator.cc
bshaffer: fixed in 3d3e299

src/google/protobuf/compiler/php/php_generator.cc
bshaffer: fixed in 3d3e299

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -219,6 +219,20 @@ std::string DefaultForField(const FieldDescriptor* field) {
   }
 }
 
+std::string DeprecatedConditionalForField(const FieldDescriptor* field) {
+  if (field->is_repeated() || field->is_map()) {
```
mkruskal-google, line 223: nit: is_map is redundant, since is_repeated will always be true for maps

mkruskal-google (approved):

mkruskal-google (approved):

Issue #16863, opened 2024-05-15
C++ generator `proto_h` feature does not compile


nikhilkalige:
I have been unable to compile c++ code generated by passing `--cpp_out=proto_h` to the compiler. I tried compiling just the generated `pb.cc` file and I get the following errors on v24.4.

```
gcc -I/nix/store/wbg3pd2jb1p9xc8208bs7hl1s55ya3ic-protobuf-24.4/include  -I/nix/store/pd328i3rxal7kr0j0sq2kf2hmyk2ss5d-abseil-cpp-20240116.1/include -std=c++17 -c mymsg.pb.cc -o mymsg.o

Data.proto.h:152:7: note: forward declaration of ‘class MyData’
  152 | class MyData;
      |       ^~~~~~~~~~~~
In file included from /usr/include/c++/11/bits/move.h:57,
                 from /usr/include/c++/11/bits/stl_pair.h:59,
                 from /usr/include/c++/11/bits/stl_algobase.h:64,
                 from /usr/include/c++/11/bits/char_traits.h:39,
                 from /usr/include/c++/11/string:40,
                 from MyMsg.proto.h:8,
                 from MyMsg.pb.cc:4:
/usr/include/c++/11/type_traits:1341:52: error: static assertion failed: template argument must be a complete class or an unbounded array
 1341 |       static_assert(std::__is_complete_or_unbounded(__type_identity<_Tp>{}),
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/11/type_traits:1341:52: note: ‘std::__is_complete_or_unbounded<std::__type_identity<MyData> >((std::__type_identity<MyData>{}, std::__type_identity<MyData>()))’ evaluates to false
In file included from MyMsg.proto.h:25,
                 from MyMsg.pb.cc:4:
/nix/store/wbg3pd2jb1p9xc8208bs7hl1s55ya3ic-protobuf-24.4/include/google/protobuf/arena.h: In instantiation of ‘class google::protobuf::Arena::InternalHelper<MyData>’:
/nix/store/wbg3pd2jb1p9xc8208bs7hl1s55ya3ic-protobuf-24.4/include/google/protobuf/arena.h:527:10:   required from ‘struct google::protobuf::Arena::is_arena_constructable<MyData>’
[...]

PR #17809, opened 2024-08-14
Fix .cc file compilation when using "proto_h" option

jerry-skydio:
My understanding is that the .proto.h header contains
the full source for that class but only the fwd declarations
of any transitive classes. When compiling the .cc file, we need
the full type info of the entire transitive tree.

Fixes: https://github.com/protocolbuffers/protobuf/issues/16863

### DIFF
--- a/src/google/protobuf/compiler/cpp/file.cc
+++ b/src/google/protobuf/compiler/cpp/file.cc
@@ -657,7 +657,7 @@ void FileGenerator::GenerateSourceIncludes(io::Printer* p) {
         GetBootstrapBasename(options_, basename, &basename);
       }
       p->Emit({{"name", basename}}, R"(
-        #include "$name$.proto.h"
+        #include "$name$.pb.h"
       )");
     }
   }


jerry-skydio:
Reviews in this chain:
└**https://github.com/protocolbuffers/protobuf/pull/17809 Fix .cc file compilation when using "proto_h" option**

jerry-skydio:
| # | head | base | diff | date | summary |
| - | - | - | - | - | - |
| 0 | [ef544155](/jerry-skydio/protobuf/commit/ef544155c9773a8af9b2219f4a75a8eb9ee27966) | [6424bcad](/jerry-skydio/protobuf/commit/6424bcad0af3a6e00f607eee37d38af60aaa74ae) | [diff](/jerry-skydio/protobuf/compare/6424bcad0af3a6e00f607eee37d38af60aaa74ae..ef544155c9773a8af9b2219f4a75a8eb9ee27966) | Aug 13 17:49 PM |  1 file changed, 1 insertion(+), 1 deletion(-) |

acozzette:
@jerry-skydio This shouldn't be necessary, as the pb.cc file should be able to compile successfully with just proto.h includes. Could you provide example .proto files to reproduce the error?

jerry-skydio:
I think I'm missing an important factor somewhere, because although this fixes the compilation of the .cc files, I get the exact same errors when trying to include a .proto.h file and using a forward declared type. The issue happens here for any message that has another message as a field:
```
/src/google/protobuf/_virtual_includes/protobuf_lite/google/protobuf/arena.h:687:57: error: invalid application of 'sizeof' to incomplete type 'Foo'
  687 |   void* mem = arena != nullptr ? arena->AllocateAligned(sizeof(T))
```
```
         inline $Submsg$* $Msg$::_internal_mutable_$name_internal$() {
           $TsanDetectConcurrentMutation$;
           $StrongRef$;
           if ($field_$ == nullptr) {
             auto* p = $superclass$::DefaultConstruct<$Submsg$>(GetArena());
             $field_$ = reinterpret_cast<$MemberType$*>(p);
           }
           return $cast_field_$;
         }
```
DefaultConstruct <T> uses sizeof(T) 
```
 template <typename T>
 PROTOBUF_NOINLINE void* Arena::DefaultConstruct(Arena* arena) {
   static_assert(is_destructor_skippable<T>::value, "");
   void* mem = arena != nullptr ? arena->AllocateAligned(sizeof(T))
                                : ::operator new(sizeof(T));
   return new (mem) T(arena);
 }
```
which isn't available because the nested message type is only fwd declared in the .proto.h file.

This is using v26.0 of protobuf and gcc 9

acozzette:
I suspect this line is the culprit for that error: https://github.com/protocolbuffers/protobuf/blob/b2a38a7b9764a0b55887ea1fa3cd958df9d6a111/src/google/protobuf/compiler/cpp/helpers.h#L1108

I think we need that function to return true so that we generate the necessary specializations of `Arena::DefaultConstruct`. Does that fix the problem for you?

Ideally we should add some test coverage of the proto_h option to prevent this kind of breakage. We rely heavily on it internally but don't really exercise this code in open source.

jerry-skydio:
ok making this change fixes it, thanks!

```
diff --git a/src/google/protobuf/compiler/cpp/helpers.h b/src/google/protobuf/compiler/cpp/helpers.h
index 0fe8de99c..bb5df56f7 100644
--- a/src/google/protobuf/compiler/cpp/helpers.h
+++ b/src/google/protobuf/compiler/cpp/helpers.h
@@ -1085,11 +1085,11 @@ void GenerateUtf8CheckCodeForCord(io::Printer* p, const FieldDescriptor* field,
 inline bool ShouldGenerateExternSpecializations(const Options& options) {
   // For OSS we omit the specializations to reduce codegen size.
   // Some compilers can't handle that much input in a single translation unit.
   // These specializations are just a link size optimization and do not affect
   // correctness or performance, so it is ok to omit them.
-  return !options.opensource_runtime;
+  return !options.opensource_runtime || options.proto_h;
 }
 
 struct OneOfRangeImpl {
   struct Iterator {
     using iterator_category = std::forward_iterator_tag;

```

mitchallain:
@acozzette Is there any guidance available from the protobuf authors/maintainers on how to handle well-known types when using the `proto_h` option? Generated `*.proto.h` headers appear to expect WKT proto headers with the same format, e.g., `any.proto.h`. Should we recompile the WKT protos with this option?

acozzette:
Yes, I think you would probably want to rebuild the well-known types with the `proto_h` option.

PR #17819, opened 2024-08-14
Add in patchable _INT64_CONVERTER

mikedh:
An alternative to https://github.com/protocolbuffers/protobuf/pull/16130, which I presume was rejected because you would also have to contribute the option to the [mapping specification](https://protobuf.dev/programming-guides/proto3/#json-options) for every language, add a bunch of tests, etc. 

The pitch here: outputting `int64` as a string produces results that are defensible as a mapping, but not great for a consumer ("why are my numbers, strings?"). This PR modifies no existing logic, but allows an upstream user who knows what they want to in-place patch the `json_format.py` without vendoring the entire `json_format` submodule:

```
from google.protobuf import json_format

assert json_format._INT64_CONVERTER == str
json_format._INT64_CONVERTER = int
```

### DIFF
--- a/python/google/protobuf/json_format.py
+++ b/python/google/protobuf/json_format.py
@@ -43,6 +43,16 @@ _INT64_TYPES = frozenset([
     descriptor.FieldDescriptor.CPPTYPE_INT64,
     descriptor.FieldDescriptor.CPPTYPE_UINT64,
 ])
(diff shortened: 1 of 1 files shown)


mikedh:
Hey @anandolee the CLA check is now passing thanks!

PR #17845, opened 2024-08-16
Build with -Wundef

cramertj:
This allows downstream consumers of protobuf to enable Wundef in their code which consumes protobuf APIs.

### DIFF
--- a/src/google/protobuf/port_def.inc
+++ b/src/google/protobuf/port_def.inc
@@ -940,10 +940,12 @@ static_assert(PROTOBUF_ABSL_MIN(20230125, 3),
 #pragma warning(disable: 4125)
 #endif
 
+#ifdef PROTOBUF_ENABLE_DEBUG_LOGGING_MAY_LEAK_PII
 #if PROTOBUF_ENABLE_DEBUG_LOGGING_MAY_LEAK_PII
 #define PROTOBUF_DEBUG true
 #else
 #define PROTOBUF_DEBUG false
+#endif
 #endif
 
 #ifdef PROTOBUF_NO_THREADLOCAL
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -146,6 +146,7 @@ def _impl(ctx):
                         flags = [
                             bit_flag,
                             "-Wall",
+                            "-Wundef",
                             "-no-canonical-prefixes",
                             "--target=" + ctx.attr.target_full_name,
                             "-fvisibility=hidden",


ericsalo (approved):

toolchain/cc_toolchain_config.bzl
```diff
@@ -146,6 +146,7 @@ def _impl(ctx):
                         flags = [
                             bit_flag,
                             "-Wall",
+                            "-Wundef",
```
mkruskal-google, line 149: I don't think this does what you think it does.  These flags only apply during our release process, and only affect our precompiled protoc artifacts

src/google/protobuf/port_def.inc
```diff
@@ -940,11 +940,13 @@ static_assert(PROTOBUF_ABSL_MIN(20230125, 3),
 #pragma warning(disable: 4125)
 #endif
 
+#ifdef PROTOBUF_ENABLE_DEBUG_LOGGING_MAY_LEAK_PII
 #if PROTOBUF_ENABLE_DEBUG_LOGGING_MAY_LEAK_PII
```
mkruskal-google, line 944: Let's just get rid of this one, we never define this anyway when we aren't using it

mkruskal-google (changes requested):

toolchain/cc_toolchain_config.bzl
cramertj: Is there another place where flags are applied such that they are checked in CI/CQ? It would be great to have some enforcement so that the warning doesn't pop back up in the future.

mkruskal-google: You should add it here: https://github.com/protocolbuffers/protobuf/blob/main/ci/common.bazelrc to apply it to CI

cramertj:
Are there further changes needed in order to get this in?

toolchain/cc_toolchain_config.bzl
mkruskal-google: This isn't resolved, this won't actually do anything

mkruskal-google:
Yes, the bzl changes aren't actually doing anything

zhangskz:
@cramertj Are you still interested in this change? Per @mkruskal-google's comment in https://github.com/protocolbuffers/protobuf/pull/17845#discussion_r1722592892, this will need to be done in https://github.com/protocolbuffers/protobuf/blob/main/ci/common.bazelrc instead.

cramertj:
@zhangskz I've pushed changes to ci/common.bazelrc, but I don't know how to test locally-- let me know if there is a process I should use here.

mkruskal-google:
I kicked off tests, but you'll probably need to rebase as it's been a while

cramertj:
Looks like there are a lot of other test failures with -Wall -Wextra or the other warnings, so I've pulled it down to a more reduced set. It's difficult to know when it's working, however, without the ability to check locally.

acozzette:
I see from the test runs that we have failures with both `shadow` and `redundant-decls`, so we probably need to limit this to `undef` and `pointer-arith` for now.

cramertj:
@acozzette Updated!

acozzette:
@cramertj Thank you for the fix. Unfortunately I don't think we can easily enforce `-Wundef` on our build, because I found that one of our dependencies is not compatible with it: https://github.com/google/benchmark If you feel up to sending another pull request there to fix them first, then that would make this possible, though.

For now I plan to submit a change that at least makes our code compatible with `-Wundef`. There won't be anything enforcing that it stays that way, however.

PR #17882, opened 2024-08-20
Fix -Wdeprecated-this-capture warning

dotnwat:
`this` needs to be explicitly captured to avoid the warning.


dotnwat:
>This seems to cause some compiler errors like this one:

🤦 thanks. fixed

PR #17888, opened 2024-08-20
[CMake] Allow for protoc executable to be configured

jonringer:
For cross-compilation, we may not want to re-use the same protoc which would be used during a CMake build (these would differ in hostPlatform).

Instead, allow for invocations of `protobuf_generate` pass their preferred programs, similar to how you can do this for grpc plugins.

Optional proposal:
Use `find_program`, however, this is a pretty big behavior change, this PR allows you to "opt-out" of the legacy behavior, but users shouldn't see a difference unless they opt-in to passing their own protoc-exe command/path

### DIFF
--- a/cmake/protobuf-generate.cmake
+++ b/cmake/protobuf-generate.cmake
@@ -2,7 +2,7 @@ function(protobuf_generate)
   include(CMakeParseArguments)
 
   set(_options APPEND_PATH)
-  set(_singleargs LANGUAGE OUT_VAR EXPORT_MACRO PROTOC_OUT_DIR PLUGIN PLUGIN_OPTIONS DEPENDENCIES)
+  set(_singleargs LANGUAGE OUT_VAR EXPORT_MACRO PROTOC_OUT_DIR PLUGIN PLUGIN_OPTIONS DEPENDENCIES PROTOC_EXE)
   if(COMMAND target_sources)
     list(APPEND _singleargs TARGET)
   endif()
@@ -83,6 +83,11 @@ function(protobuf_generate)
     endforeach()
   endif()
 
+  if(NOT protobuf_generate_PROTOC_EXE)
+    # Default to using the CMake executable
+    set(protobuf_generate_PROTOC_EXE protobuf::protoc)
+  endif()
+
   foreach(DIR ${protobuf_generate_IMPORT_DIRS})
     get_filename_component(ABS_PATH ${DIR} ABSOLUTE)
     list(FIND _protobuf_include_path ${ABS_PATH} _contains_already)
@@ -143,7 +148,7 @@ function(protobuf_generate)
 
     add_custom_command(
       OUTPUT ${_generated_srcs}
-      COMMAND protobuf::protoc
+      COMMAND ${protobuf_generate_PROTOC_EXE}
       ARGS ${protobuf_generate_PROTOC_OPTIONS} --${protobuf_generate_LANGUAGE}_out ${_plugin_options}:${protobuf_generate_PROTOC_OUT_DIR} ${_plugin} ${_protobuf_include_path} ${_abs_file}
       DEPENDS ${_abs_file} ${protobuf_PROTOC_EXE} ${protobuf_generate_DEPENDENCIES}
       COMMENT ${_comment}
--- a/docs/cmake_protobuf_generate.md
+++ b/docs/cmake_protobuf_generate.md
@@ -121,6 +121,7 @@ Single-value arguments:
 - `LANGUAGE` — A single value: cpp or python. Determines what kind of source files are being generated.
 - `OUT_VAR` — Name of a CMake variable that will be filled with the paths to the generated source files.
 - `EXPORT_MACRO` — Name of a macro that is applied to all generated Protobuf message classes and extern variables. It can, for example, be used to declare DLL exports.
+- `PROTOC_EXE` — Command name, path, or CMake executable used to run protoc commands. Defaults to `protobuf::protoc`.
 - `PROTOC_OUT_DIR` — Output directory of generated source files. Defaults to `CMAKE_CURRENT_BINARY_DIR`.
 - `PLUGIN` — An optional plugin executable. This could, for example, be the path to `grpc_cpp_plugin`.
 - `PLUGIN_OPTIONS` — Additional options provided to the plugin, such as `generate_mock_code=true` for the gRPC cpp plugin.


jonringer:
Created as draft until I properly test this

jonringer:
Was able to test this in-house, and successfully build protos while passing this option

mkruskal-google:
This looks reasonable, but I'm a little confused about the use-case.  Presumably even under cross-compilation there's only a single version of protoc that *can* execute during a cmake build right?  Why would you end up with the wrong one hooked up to `protobuf::protoc`?  Is it that you need the cross-compiled `libprotobuf` and `protoc` comes along for the ride?

mkruskal-google:
Also do you mind rebasing this PR?

jonringer:
> presumably even under cross-compilation there's only a single version of protoc that can execute during a cmake build right?

Yes, but is it the one which is valid for the platform? `protobuf::protoc` is the target which will be populated when doing `find_package`, but that will likely be targeting the target platform, which will be different from the build platform. This is the error I'm seeing:

```
/nix/store/1xhds5s320nfp2022yjah1h7dpv8qqns-bash-5.2p32/bin/bash: line 1: /nix/store/irnwnib0g9g29p8b27hnbw83gl57f6bz-protobuf-aarch64-unknown-linux-gnu-24.4/bin/protoc: cannot execute binary file: Exec format error
```

So `protobuf::protoc` translates to already built protobuf package for the target platform (GNU naming), however, I want to bring in the `protoc` variant for x86 (which is on PATH).

Essentially this change allows me to do:
```
find_program(protoc_exe protoc REQUIRED)

protobuf_generate(
  ...
  PROTOC_EXE ${protoc_exe}
)
```

jonringer:
>  Is it that you need the cross-compiled libprotobuf and protoc comes along for the ride?

I need to link against `protobuf::libprotobuf` of the host/target platform, but need `protoc` to be of the build platform.

mkruskal-google:
Yea that makes sense.  I'm surprised cmake doesn't have support for this kind of thing tho..  If we had two different packages (one for protoc and one for the C++ runtime) I don't think this problem would exist right?  Are we out of model by providing our library and build tools together in the same package?

jonringer:
You may be able to do something toolchain files, but I'm kind of just using nix to invoke "vanilla" CMake builds.

> If we had two different packages (one for protoc and one for the C++ runtime) I don't think this problem would exist right?

Not sure, when invoking CMake, I'm passing it target specific C/C++ compilers
```
-- The C compiler identification is GNU 13.2.0
-- The CXX compiler identification is GNU 13.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /nix/store/fk52fiazhyw9nwb015kbn4wd0xp0psdp-aarch64-unknown-linux-gnu-gcc-wrapper-13.2.0/bin/aarch64-unknown-linux-gnu-gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /nix/store/fk52fiazhyw9nwb015kbn4wd0xp0psdp-aarch64-unknown-linux-gnu-gcc-wrapper-13.2.0/bin/aarch64-unknown-linux-gnu-g++ - skipped
```

> Are we out of model by providing our library and build tools together in the same package?

No, I don't think so. It's just cross compiling seems to throw a wrench in most package workflows. In this example, I need to bring in two different versions of protobuf (one x86, one aarch64) to do this workflow. So I can't really "re-use" one single variant.

With Nix, I can introduce both variants, just need a way to be opinionated when one variant is used. Generally the assumption is that I want stuff for hostTarget (e.g. libaries, and aarch64 in this case), but I still need something for build platform (x86) to run an executable at build time.

mkruskal-google (approved):

mkruskal-google:
IIUC you don't actually need two different versions of protobuf though.  You need a `libprotobuf` for the target and a `protoc` for the host right?  If those were separate packages you'd be able to pull them in separately w/o worrying about contamination I'd think

jonringer:
> IIUC you don't actually need two different versions of protobuf though. You need a libprotobuf for the target and a protoc for the host right? If those were separate packages you'd be able to pull them in separately w/o worrying about contamination I'd think

Then I would need to bring a `protobuf` package for aarch64 and a `protoc` package for x86. Still the same amount of work for me :upside_down_face:

mkruskal-google:
Kindof, you wouldn't need to worry about which binary gets used for protobuf_generate though (assuming it shipped with the protoc package).

Anyway, this lgtm for now.  We can discuss packaging internally, this is a good data point :)

jonringer:
> Kindof, you wouldn't need to worry about which binary gets used for protobuf_generate though (assuming it shipped with the protoc package).

Sure, but I'm not sure if the pros will out weigh the cons when most package repos logically thing of protobuf and protoc as coming from the same source.

I'm kind of indifferent... but maybe it would simplify distribution on your side to think about protoc and protobuf as separate.

justusranvier:
> IIUC you don't actually need two different versions of protobuf though. You need a `libprotobuf` for the target and a `protoc` for the host right? If those were separate packages you'd be able to pull them in separately w/o worrying about contamination I'd think

If `protoc` and `libprotobuf` where different cmake packages then projects could do:

```
find_package(libprotobuf CONFIG REQUIRED)
find_package(protoc CONFIG REQUIRED NO_CMAKE_FIND_ROOT_PATH)
```

This would cause libprotobuf to be searched for in the paths for the target architecture and protoc to be searched for in the paths for the host architecture which is what everybody wants.

jonringer:
> If protoc and libprotobuf where different cmake packages then projects could do:

You would still need package those two things separately, and introduce them in a way so that they get found in the correct order and for the correct platforms.

Not trying to dismiss your suggestion, rather, trying to point out that someone who would already have to do a fair amount of construction to find the right cmake config files.

I think `find_program(` would be a better tool for finding protoc. It aligns better with POSIX PATH being used for finding executables/commands.

justusranvier:
> I think `find_program(` would be a better tool for finding protoc. It aligns better with POSIX PATH being used for finding executables/commands.

The case for that would be stronger if the only resource consumed was the protoc executable itself and not package-specific CMake functions related to that binary.

In an alternate history where the packaging was perfect from day one CMake would not ship special case handling for protobuf at all. `find_package(libprotobuf CONFIG)` would bring in targets for the library only, and `find_package(protoc CONFIG)` would not add new targets but would make an implementation of the `protobuf_generate` function available that would call the `protoc` executable corresponding to the package file which was found, and that search path could be controlled just like any other package (`<PackageName>_DIR`, etc.).

justusranvier:
A theoretical `protoc` package would only need to be three files at most:

```
${CMAKE_INSTALL_PREFIX}/*/protoc-config.cmake
${CMAKE_INSTALL_PREFIX}/*/protobuf-generate.cmake
${CMAKE_INSTALL_PREFIX}/bin/protoc
```

`protoc-config.cmake` appends the correct directory to `CMAKE_MODULE_PATH` and calls `include(protobuf-generate)`

`protobuf-generate.cmake` needs to be written to not call any random `protoc` which happens to be on the path but the specific `protoc` which was installed at a known location relative to `protobuf-generate.cmake`.

Then any number of versions can exist at the same time as long as they were installed to different locations and users can control which one gets found and used during a build.

Issue #17852, opened 2024-08-18
Protobufs use of <sys/time.h> prevents building of protobuf php extension on Windows


gwharton:
https://github.com/protocolbuffers/protobuf/blob/025909cb539359c8001cf69707229587d430adff/third_party/utf8_range/main.c#L7

Microsoft Windows 11
Microsoft Visual Studio Community 2022
Protocol Buffers v27.3

Can anything be done to get compatability with Windows on this dependent library
```
c:\Users\GrahamWharton\php-sdk-binary-tools\php8.3.10\vs17\x64\php-src
$ "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.41.34120\bin\HostX64\x64\nmake.exe" /I /nologo "php_protobuf.dll"
[...]

PR #17903, opened 2024-08-21
Updated Build Configuration for PHP Extension to fix issue building on Windows

gwharton:
Protobuf PHP Extension

When building the module under Linux, the config.m4 file compiles the module files "arena.c array.c convert.c def.c map.c message.c names.c php-upb.c protobuf.c" and in addition it compiles the third party file "third_party/utf8_range/utf8_range.c"

When building under Windows, a separate config.w32 file is used which compiles the same "arena.c array.c convert.c def.c map.c message.c names.c php-upb.c protobuf.c" module files, but in addition, it compiles *.c from the third_party/utf8_range subfolder. Some of these files are not compatible with windows due to including <sys/time.h>. Only the utf8_range.c file is required, as shown under the config.m4 linux build configuration.

This PR modifies the build script config.w32 so that builds under windows only compile the same files that is compiled under linux. It removes the wildcard compile of the entire third party include, and thus resolves the build error under Windows.

Here is the linux build config.m4 file for reference. https://github.com/protocolbuffers/protobuf/blob/72707c38f3580d5ce76eda1d736ddad468219b57/php/ext/google/protobuf/config.m4#L1-L11

Verified working under :

Microsoft Windows 11 Pro 24H2 x64
Micrsoft Visual Studio Community 2022  (17.0 - 14.41.34120) x64 Build Environment
Builds against PHP 8.3.10 from source in full PHP recompile


gwharton:
Fixes https://github.com/protocolbuffers/protobuf/issues/17852

JasonLunn:
@gwharton - would you mind rebasing this PR? The test failures don't look related to the change.

acozzette (approved):

PR #17959, opened 2024-08-27
Fix signature of `PyUpb_MessageMeta_Clear`

hoodmane:
A `tp_clear` function should have signature `int f(PyObject*)`. The presence of erroneous extra parameters leads to undefined behavior as indicated in the C specification 6.3.2.3.8. In WebAssembly builds, this causes crashes. https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf#page=60

### DIFF
--- a/python/message.c
+++ b/python/message.c
@@ -2025,7 +2025,7 @@ static int PyUpb_MessageMeta_Traverse(PyObject* self, visitproc visit,
   return cpython_bits.type_traverse(self, visit, arg);
 }
 
-static int PyUpb_MessageMeta_Clear(PyObject* self, visitproc visit, void* arg) {
+static int PyUpb_MessageMeta_Clear(PyObject* self) {
   PyUpb_MessageMeta* meta = PyUpb_GetMessageMeta(self);
   Py_CLEAR(meta->py_message_descriptor);
   return cpython_bits.type_clear(self);


haberman (approved):

hoodmane:
Thanks for the review @haberman. Can this be merged?

haberman (approved):

haberman:
We have an automated system that should take over now and submit the change internally first, with it propagating to GitHub after.

acozzette (approved):

hoodmane:
Why did the bot close this? The patch is clearly correct...

haberman:
It was closed because the bot committed the change in https://github.com/protocolbuffers/protobuf/commit/b915e9f44e9d979617d7738f7ddd7dd02ccd7c76

haberman:
Out of curiosity, what's your use case for compiling this to WebAssembly? I'd love to hear how that works out.

hoodmane:
Ah, the workflow here is a bit confusing, thanks for the explanation! It might be helpful for new contributors if the bot posted a comment explaining that the PR was merged. It does say "closed in commit hash" but I got confused because github marks the status as `Closed` and not `Merged`.

>  what's your use case

@bartbreore could perhaps clarify what he uses it for. He added a wasm protobuf build to Pyodide in May of 2023. I think he was mostly concerned about other packages downstream of protobuf but I'm not sure which ones.
https://github.com/pyodide/pyodide/pull/3813

PR #17971, opened 2024-08-27
Fixed some shadowing between local variables

mikael-s-persson:
Fixed some shadowing between local variables

Pretty simple fix, just renaming a couple local variables so they don't shadow other variables in scope. Just trying to reduce verbosity of warnings in users' own builds with higher warning levels (as is my case ;) ).

Obviously, no change in behavior, and hopefully the new names are reasonable.

### DIFF
--- a/src/google/protobuf/descriptor.cc
+++ b/src/google/protobuf/descriptor.cc
@@ -9656,8 +9656,8 @@ void FieldDescriptor::InternalTypeOnceInit() const {
       } else {
         name = lazy_default_value_enum_name;
       }
-      Symbol result = file()->pool()->CrossLinkOnDemandHelper(name, true);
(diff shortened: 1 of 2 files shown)


acozzette (approved):

acozzette:
Thanks for cleaning this up!

acozzette (approved):

mikael-s-persson:
FYI, I guess there's a copybara error I can't see the details of from this side (outside Google, that is).

acozzette (approved):

mikael-s-persson:
@acozzette I don't know if it's an automated system that keeps re-approving this.
I cannot click the "merge" button, it's grayed out, because I'm not an authorized user. So, maybe merge it if you can, or ping benza or whoever.

PR #18005, opened 2024-08-28
Remove `--copt="-Werror"` from `.bazelrc`

acozzette:
This appears to be causing a bunch of errors during our release process, but I don't think it's critical to have this flag in `.bazelrc`. We should still be able to get about the same test coverage since we also set `-Werror` in the `bazelrc` files in the `ci/` directory.


JasonLunn (approved):

rockwotj:
@acozzette speaking of warnings, would it be possible to take a look at https://github.com/protocolbuffers/protobuf/pull/17882? Thanks in advance!

PR #18103, opened 2024-09-04
make sure getJsonName returns a valid string

lekoala:
this ensure compatibility with recent version of php expecting a string and not null

I've encountered this issue when using the https://github.com/googleapis/php-analytics-data sdk

### DIFF
--- a/php/src/Google/Protobuf/Internal/FieldDescriptor.php
+++ b/php/src/Google/Protobuf/Internal/FieldDescriptor.php
@@ -61,7 +61,7 @@ class FieldDescriptor
 
     public function getJsonName()
     {
-        return $this->json_name;
+        return isset($this->json_name) ? $this->json_name : '';
     }
 
     public function setSetter($setter)


JasonLunn:
@lekoala - the CLA must be completed before this can be reviewed.

lekoala:
> @lekoala - the CLA must be completed before this can be reviewed.

i did that already as far as I'm aware

=> It looks like you've already signed this CLA. If you'd like to edit your contact information, you may do so below.

lekoala:
@bshaffer yes the missing types are a bit of a shame really but if you look how its used then its obvious

in java, it's typed

https://github.com/protocolbuffers/protobuf/blob/627a691347b3aaf47d63263c0d6ed5855a9a8701/java/core/src/main/java/com/google/protobuf/Descriptors.java#L1263

in php, strlen is called on it and in modern php, it cannot be null

https://github.com/protocolbuffers/protobuf/blob/627a691347b3aaf47d63263c0d6ed5855a9a8701/php/src/Google/Protobuf/Internal/Message.php#L1873

lekoala:
to be honest, maybe there should be a value there and the missing value indicates a broader issue

i don't have the stack trace unfortunately so i cannot reproduce the error

PR #18130, opened 2024-09-05
only print out protobuf_VERSION_PRERELEASE if it is set

imrichardcole:
Building from main in the source repo, without this you always get a blank output i.e

`--`

This will only print out this value if it is set.

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -68,7 +68,7 @@ class Any(object):
   def TypeName(self):
     """Returns the protobuf type name of the inner message."""
     # Only last part is to be used: b/25630112
-    return self.type_url.split('/')[-1]
+    return self.type_url.rpartition('/')[2]
 
   def Is(self, descriptor):
     """Checks if this Any represents the given protobuf type."""


imrichardcole:
Hi - any idea how/when this might get reviewed?  If I need to make changes, please let me know

imrichardcole:
Hi - do I need to do anything to get this PR merged?  Don't imagine the above test failures are due to my change?

CMakeLists.txt
```diff
@@ -106,7 +106,9 @@ string(REGEX REPLACE     "${protobuf_VERSION_REGEX}" "\\3"
 string(REGEX REPLACE     "${protobuf_VERSION_REGEX}" "\\5"
   protobuf_VERSION_PRERELEASE "${protobuf_VERSION_STRING}")
 
-message(STATUS "${protobuf_VERSION_PRERELEASE}")
+if(protobuf_VERSION_PRERELEASE)
```
zhangskz, line 109: Could we just move this to the existing if(protobuf_VERSION_PRERELEASE) block on line 17 instead?

zhangskz (commented):
Sorry for the delay in review! Minor nit, but otherwise looks reasonable. Can you rebase as well?

CMakeLists.txt
imrichardcole: Sure, let me create a new PR as I think the codebase has moved on quite a bit since I created this one.

PR #18193, opened 2024-09-09
fix raw_ptr.cc on exotic architectures

jagub2:
I was working with an exotic architecture where ABSL_CACHELINE_SIZE of abseil was less than 64, so I got a compilation error about redefined symbols. I think the cc file should be adapted to the header file, so here is my change.

### DIFF
--- a/src/google/protobuf/raw_ptr.cc
+++ b/src/google/protobuf/raw_ptr.cc
@@ -18,7 +18,7 @@ namespace protobuf {
 namespace internal {
 
 ABSL_CONST_INIT PROTOBUF_EXPORT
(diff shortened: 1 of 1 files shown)


sbenzaquen (approved):

tonyliaoss:
Integration tests are failing due to version skew... I'm going to force-pull this one.

jagub2:
I could rebase this change on top of any commit, if you provide it. However, I'd be able to rebase it in the middle of next week.

tonyliaoss:
Thanks for the offer Jakub! I think the force-pull went well :)

PR #18231, opened 2024-09-11
fix: extern cause clang link failure

reitowo:
Use clang-cl 17.0.6 / 18.1.8 failed to compile the code using CMake, with link error:

```
lld-link: error: undefined symbol: void __cdecl google::protobuf::internal::memswap<12>(char *__restrict, char *__restrict)
>>> referenced by xxxxx\cmake-build-debug-win-x86-md-17\stackwalker.pb.cc:1179
>>>               xxx.lib(stackwalker.pb.cc.obj):(private: void __thiscall xxx::protobuf::OnDbgHelpErrRequest::InternalSwap(class xxx::protobuf::OnDbgHelpErrRequest *))
>>> referenced by xxxxx\cmake-build-debug-win-x86-md-17\vcpkg_installed\x86-windows-static-md\include\google\protobuf\repeated_ptr_field.h:330
>>>               xxx.lib(stackwalker.pb.cc.obj):(public: void __thiscall google::protobuf::internal::RepeatedPtrFieldBase::InternalSwap(class google::protobuf::internal::RepeatedPtrFieldBase *__restrict))
ninja: build stopped: subcommand failed.
```

It seems LLVM failed to keep the symbol as extern, or it simply a bug to let the code compile using MSVC.
Either use `extern "C"` or remove `extern` solves the problem. I use `extern "C"` to keep the original behavior.

### DIFF
--- a/src/google/protobuf/repeated_ptr_field.h
+++ b/src/google/protobuf/repeated_ptr_field.h
@@ -2041,7 +2041,7 @@ UnsafeArenaAllocatedRepeatedPtrFieldBackInserter(
 namespace internal {
 // Size optimization for `memswap<N>` - supplied below N is used by every
 // `RepeatedPtrField<T>`.
-template PROTOBUF_EXPORT_TEMPLATE_DECLARE void
+extern "C" template PROTOBUF_EXPORT_TEMPLATE_DECLARE void
 memswap<ArenaOffsetHelper<RepeatedPtrFieldBase>::value>(
     char* PROTOBUF_RESTRICT, char* PROTOBUF_RESTRICT);
 }  // namespace internal
--- a/src/google/protobuf/repeated_ptr_field.h
+++ b/src/google/protobuf/repeated_ptr_field.h
@@ -2041,7 +2041,7 @@ UnsafeArenaAllocatedRepeatedPtrFieldBackInserter(
 namespace internal {
 // Size optimization for `memswap<N>` - supplied below N is used by every
 // `RepeatedPtrField<T>`.
-extern template PROTOBUF_EXPORT_TEMPLATE_DECLARE void
+template PROTOBUF_EXPORT_TEMPLATE_DECLARE void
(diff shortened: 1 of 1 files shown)


JasonLunn:
@reitowo - this seems to be failing tests. Example error:

```
src/google/protobuf/repeated_ptr_field.cc:129:1: error: duplicate explicit instantiation of 'memswap<16UL>'
  129 | memswap<ArenaOffsetHelper<RepeatedPtrFieldBase>::value>(
      | ^
bazel-out/k8-fastbuild/bin/src/google/protobuf/_virtual_includes/protobuf_lite/google/protobuf/repeated_ptr_field.h:291:15: note: previous explicit instantiation is here
  291 |     internal::memswap<ArenaOffsetHelper<RepeatedPtrFieldBase>::value>(
      |               ^
1 error generated.
```

reitowo:
Which compiler did it use?

JasonLunn:
I see variations of the same error message under both GCC and Clang.

reitowo:
After I changed my compiler to MSVC this issue was gone. 

But I still hope someone can take a look about this link failure under Clang/ClangCL. This probably a MSVC/LLVM bug.

JasonLunn:
> After I changed my compiler to MSVC this issue was gone.
> 
> But I still hope someone can take a look about this link failure under Clang/ClangCL. This probably a MSVC/LLVM bug.

I'm not clear what you mean by "this issue is gone" - do you mean the original issue (that prompted the PR in the first place) or the test failures on the PR?

reitowo:
The original issue.

acozzette:
@reitowo Thank you for the pull request, but I'm going to close this since it sounds like you have a workaround, and since this fix unfortunately doesn't work for all the compilers we test.

If we could find a way to reproduce the issue in Godbolt then that would make it a lot easier to find a solution, but I gave it a try just now and wasn't able to reproduce it.

PR #18233, opened 2024-09-11
Fix ruby lib paths in `rules_ruby`

tempoz:
See protocolbuffers/rules_ruby#38. This fixes the ruby portion of the protocolbuffers build for me.


JasonLunn:
I've left review comments on https://github.com/protocolbuffers/rules_ruby/pull/38. I'm not yet clear on what makes this change necessary.

PR #18234, opened 2024-09-11
Fix bash ternary in `//benchmarks:size_data`

tempoz:
Replaces pseudocode bash ternary with working bash ternary.


tonyliaoss:
This looks upb related. Let me add @haberman.

tonyliaoss:
Same version skew issue as PR #18241 affecting integration tests.

I've checked with @haberman  and sounds like he doesn't have a problem with merging this. Gonna force-pull.

PR #18235, opened 2024-09-11
Fix deprecated use of `namespace_packages` in `python/dist/setup.py`

tempoz:
Uses implicit namespaces instead of explicit, as per https://peps.python.org/pep-0420/

Fixes a warning otherwise output during build.

### DIFF
--- a/python/dist/setup.py
+++ b/python/dist/setup.py
@@ -12,7 +12,7 @@ import glob
 import os
 import sys
 
-from setuptools import setup, Extension, find_packages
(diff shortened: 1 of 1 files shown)


urbas:
You will also have to remove the file `python/google/__init__.py` otherwise it will interfere with Python's namespace logic and Python will not treat it as a namespace package (I think :D ).

zhangskz:
Can you include the warning you hit without this?

acozzette (approved):

PR #18239, opened 2024-09-11
Fix use of `tmpnam`

tempoz:
`c++17` introduces `std::filesystem::temp_directory_path`, which allows us to resolve the warning regarding th euse of `tmpnam` here.

### DIFF
--- a/src/google/protobuf/testing/googletest.cc
+++ b/src/google/protobuf/testing/googletest.cc
@@ -31,6 +31,7 @@
 #include <fcntl.h>
 #include <iostream>
 #include <fstream>
+#include <filesystem>
 
 namespace google {
 namespace protobuf {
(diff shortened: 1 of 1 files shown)


tonyliaoss:
So I don't think we will accept this PR because we still need to support C++14 until December 15 of this year:
https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

Bifurcating our code for this test function that "mostly works" doesn't seem worth it.

In the long term, we want to use [`testing::TempDir()`](https://github.com/google/googletest/blob/0953a17a4281fc26831da647ad3fcd5e21e6473b/googletest/src/gtest.cc#L6904-L6914) from GoogleTest that pretty much does this, but migrating would still be hard due to things depending on the environment variable `$TEST_TMPDIR` being set...

tempoz:
sounds good; thanks for your time considering this.

PR #18240, opened 2024-09-11
Initialize variables to address warning regarding possible use of uninitialized variables

tempoz:


haberman:
Which compiler is raising these warnings? What compiler version and what compilation flags?

tempoz:
The compiler I am using is `gcc (Debian 14.2.0-3) 14.2.0`. I am not using any special flags, I just run `bazel build //...`.

zhangskz:
Could you rebase? I suspect the failing tests were at head.

tempoz:
yeah, no problem

tempoz:
Rebased this again since it's been a bit; feel free to re-run tests and review whenever you find time!

haberman:
I don't think this is the correct fix.

The real issue here is that both `upb_WireReader_ReadTag()` and `upb_WireReader_ReadSize()` can return `NULL` to indicate error, and in that case the variables will not be initialized.

However, in this case the data is coming from the unknown fields of a `upb_Message`. We should be guaranteed that this is valid wire format.

I would suggest the following fix. This *should* make the warnings go away:

Add `UPB_ASSUME(ptr != NULL)` after the `upb_WireReader_ReadTag()` and `upb_WireReader_ReadSize()` calls. As a bonus, this will give us loud warnings in debug mode if these assumptions turn out not to be true.

We should probably do the same for every WireReader call in this function that returns `ptr`.

haberman:
It's possible that we may need to add `UPB_ASSUME(ret != NULL)` into the success paths of the WireReader functions also.

PR #18242, opened 2024-09-11
Fix warnings about comparisons of integers of different sizes/signedness

tempoz:


JasonLunn:
@tempoz - Closing this issue because many of these changes may have already landed via #17212.

PR #18244, opened 2024-09-11
Fix broken namespaces in `hpb_generator/`

tempoz:
The wrong namespaces were being used, and the wrong number of curly braces (!!!) terminated namespaced code. Presumably this is an artifact of a refactor; here is a fix.

### DIFF
--- a/hpb_generator/context.h
+++ b/hpb_generator/context.h
@@ -107,7 +107,6 @@ inline void EmitFileWarning(const google::protobuf::FileDescriptor* file, Contex
       file->name());
   ctx.Emit("\n");
 }
-}  // namespace protobuf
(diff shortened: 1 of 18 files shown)


JasonLunn:
@tempoz - thanks for the fix. It seems like there have been some updates to some of these files in the interval since you opened this PR. Can you rebase and see if the fix is still needed, and if so, resolve the merge conflicts?

tempoz:
Done! I think these changes are just as necessary; I don't really understand how these files were supposed to compile before? They all have extra unmatched closing curly braces.

honglooker:
This'll need internal transformations to complete - I'll take of that. Thank you for spotting nonetheless!

PR #18246, opened 2024-09-11
Fix deprecated usage of `MutableRepeatedPtrField` by implementing iterator for `MutableRepeatedFieldRef`

tempoz:
Couldn't replace the deprecated usage of `MutableRepeatedPtrField` in `src/google/protobuf/reflection_visit_field_info.h` with `MutableRepeatedFieldRef` as the warning suggested because it did not implemet an iterator, so I implemented the iterator.

### DIFF
--- a/src/google/protobuf/reflection.h
+++ b/src/google/protobuf/reflection.h
@@ -46,8 +46,8 @@ class MutableRepeatedFieldRef;
 // RepeatedFieldRef definition for non-message types.
 template <typename T>
 class RepeatedFieldRef<
-    T, typename std::enable_if<!std::is_base_of<Message, T>::value>::type> {
-  typedef typename internal::RefTypeTraits<T>::iterator IteratorType;
+    T, typename std::enable_if<!std::is_base_of<Message, typename std::remove_const<T>::type>::value>::type> {
+  typedef typename internal::RefTypeTraits<const T>::iterator IteratorType;
   typedef typename internal::RefTypeTraits<T>::AccessorType AccessorType;
 
  public:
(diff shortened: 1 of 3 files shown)


zhangskz:
There are some tests failing, but at an initial glance I suspect these are unrelated to the current change. Can you try rebasing?

tempoz:
Done!

tempoz:
Rebased it again, since it's been a while. Feel free to take another look; I can't re-run the tests myself.

acozzette:
@tempoz Thank you for the pull request. I looked at this yesterday and unfortunately there is a big blocker preventing us from merging this, which is that in our internal code we support repeated strings of type `absl::Cord`. We have open sourced partial support for Cords but only for singular `bytes` fields. We also support repeated `StringPiece` fields--this is a weird legacy type that is similar to `std::string_view` but also supports aliasing to reduce copying.

These special types are tricky to handle, because while const iteration treats them as if they were plain string fields, for mutable iteration we would need to expose the actual type (e.g. `absl::Cord`). I think in principle we could figure out a way to make this work, but it would probably be a big project.

Unfortunately I think the best we could do for now would probably be to silence the deprecation warnings by just updating the code in reflection_visit_field_info.h to call `MutableRepeatedFieldInternal` and `MutableRepeatedPtrFieldInternal`, which is what the deprecated methods call internally.

acozzette:
I updated this code in https://github.com/protocolbuffers/protobuf/commit/07faf360a3cdfcce79be26d6ed7561195f759f81 to work around the deprecated methods, so this will at least silence the compiler warnings. I'm going to close this pull request because unfortunately I don't think there is a path to merging it for the reasons I mentioned in my previous comment. Sorry for the bad news and thank you for the pull request nonetheless.

PR #18247, opened 2024-09-11
Add missing `google/protobuf` import prefix to protos in `editions/codegen_tests`

tempoz:
This was causing build failures.


mkruskal-google:
What build failures?  None of these have any dependencies right?

mkruskal-google:
Actually I see what happened here.  This is just a bug in the transformations from our internal code.  https://github.com/protocolbuffers/protobuf/pull/18408 should fix our internal setup

tempoz:
Awesome! Happy it was an easy fix for you guys ^-^

PR #18248, opened 2024-09-11
Add missing dep to `//rust:protobuf_cpp_test`

tempoz:


esrauchg:
I'm surprised by this proposed change; I would have expected that any deps that are used by the crate-under-test aren't added to the rust_test(crate=""), only deps that are solely in the test configuration.

Did you add this because it gives a clear error when you are building it under bazel test?

tempoz:
This fixed an error I hit early on when trying to get things to build, but appears not to be necessary after my other changes, so I think this fix masked another underlying problem that was fixed by something else. I did get an error that suggested this change, but I couldn't tell you what it was anymore. Closing.

PR #18249, opened 2024-09-11
Fix visibility for `//upb_generator:keywords`

tempoz:
THis was causing build failures.


tempoz:
This target was removed since my pull.

PR #18250, opened 2024-09-11
Fix mis-named dep for `//hpb:protoc-gen-upb-protos`

tempoz:


tonyliaoss:
Hi tempoz,

What is the OS, build environment, and toolchain that you are using that runs into this build issue?

We'd like to understand if we have any gaps in our CI coverage.

Thanks!
Tony

tempoz:
Hi Tony,
My environment is always follows:

OS: debian trixie
CC: gcc (Debian 14.2.0-3) 14.2.0
Bazel: 7.3.1

In this case, though, AFAICT, the named dep simply doesn't exist. There is no target named `plugin` named in `src/google/protobuf/compiler/BUILD.bazel`.

haberman:
We've removed these BUILD files for the time being, as we are in a state of flux with hpb: https://github.com/protocolbuffers/protobuf/pull/18363

PR #18253, opened 2024-09-11
Fix `rust_library` targets in `//rust`

tempoz:
`rust/shared.rs` also requires the `--cfg=bzl` config flag to trigger `cpp_kernel` builds.

### DIFF
--- a/rust/BUILD
+++ b/rust/BUILD
@@ -114,6 +114,7 @@ rust_library(
     rustc_flags = ["--cfg=upb_kernel"],
     visibility = [
         "//src/google/protobuf:__subpackages__",
+        "//rust/test:__subpackages__",
     ],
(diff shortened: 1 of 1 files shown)


esrauchg:
Thanks for flagging this breakage! Unfortunately the PR would improperly merge with some internal lines that we have in the file which are stripped before we push to open source.

I think these PRs being merged will solve the problem you have identified here:
https://github.com/protocolbuffers/protobuf/pull/18383
https://github.com/protocolbuffers/protobuf/pull/18282

Appreciate your legwork on these breakages. I also have enabled //rust:all on our GHA CI and I am working towards fixing the other breakages so we can turn on the rest of //rust/..., hopefully within the next few work days.

PR #18254, opened 2024-09-11
Make rust includes point to existing proto header files

tempoz:
The files that were being pointed to did not exist.

### DIFF
--- a/src/google/protobuf/compiler/rust/naming.cc
+++ b/src/google/protobuf/compiler/rust/naming.cc
@@ -62,7 +62,7 @@ std::string GetThunkCcFile(Context& ctx, const FileDescriptor& file) {
 
 std::string GetHeaderFile(Context& ctx, const FileDescriptor& file) {
   auto basename = StripProto(file.name());
-  return absl::StrCat(basename, ".proto.h");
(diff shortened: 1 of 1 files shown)


src/google/protobuf/compiler/rust/naming.cc
```diff
@@ -62,7 +62,7 @@ std::string GetThunkCcFile(Context& ctx, const FileDescriptor& file) {
 
 std::string GetHeaderFile(Context& ctx, const FileDescriptor& file) {
   auto basename = StripProto(file.name());
-  return absl::StrCat(basename, ".proto.h");
+  return absl::StrCat(basename, ".pb.h");
```
haberman, line 65: Internally we use `.proto.h` files (this option is open-sourced by apparently not enabled by default: https://github.com/protocolbuffers/protobuf/blob/ebbb1a465fedf69f7cb450c9133d9b1a49678a68/src/google/protobuf/compiler/cpp/generator.cc#L157)

So we probably cannot land this PR as-is. If we need to do `.pb.h` for open-source, we'd need to internally land a change that conditions it on internal vs external.

esrauchg:
This issue was separately fixed at tip of main, thanks!

PR #18255, opened 2024-09-11
Fix `plugin_output` for `_generate_rust_gencode`

tempoz:
Used the python rules as an example to fix this. Bazel now recognizes that the output files have been produced, where before there were build failures regarding the outputs not being produced.

### DIFF
--- a/rust/aspects.bzl
+++ b/rust/aspects.bzl
@@ -123,6 +123,12 @@ def _generate_rust_gencode(
         ),
     )
 
+    proto_root = proto_info.proto_source_root
+    # Handles virtual import cases
+    if proto_root.startswith(ctx.bin_dir.path):
+        proto_root = proto_root[len(ctx.bin_dir.path) + 1:]
+
+
     proto_common.compile(
         actions = ctx.actions,
         proto_info = proto_info,
@@ -130,7 +136,7 @@ def _generate_rust_gencode(
         additional_args = additional_args,
         generated_files = rs_outputs + cc_outputs,
         proto_lang_toolchain_info = proto_lang_toolchain,
-        plugin_output = ctx.bin_dir.path,
+        plugin_output = ctx.bin_dir.path + "/" + proto_root
     )
     return (rs_outputs, cc_outputs)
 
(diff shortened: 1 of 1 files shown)


rust/aspects.bzl
```diff
@@ -125,14 +125,20 @@ def _generate_rust_gencode(
...
+
     proto_common.compile(
         actions = ctx.actions,
         proto_info = proto_info,
         additional_inputs = depset(direct = [crate_mapping]),
         additional_args = additional_args,
         generated_files = rs_outputs + cc_outputs,
         proto_lang_toolchain_info = proto_lang_toolchain,
-        plugin_output = ctx.bin_dir.path,
+        plugin_output = ctx.bin_dir.path + "/" + proto_root
```
haberman, line 141: `plugin_output` is deprecated: https://github.com/protocolbuffers/protobuf/blob/ebbb1a465fedf69f7cb450c9133d9b1a49678a68/bazel/common/proto_common.bzl#L142-L145

I think it would be better to just remove it if possible. Does that work?

tempoz: Sounds good to me! Do you want to handle it or would you like me to?

JasonLunn: Please include the change in the PR

tempoz: Done!

JasonLunn:
Seeing failures on the Rust tests of the form:

```
	File "/workspace/rust/aspects.bzl", line 284, column 37, in _rust_cc_proto_aspect_impl
		return _rust_proto_aspect_common(target, ctx, is_upb = False)
	File "/workspace/rust/aspects.bzl", line 321, column 47, in _rust_proto_aspect_common
		(gencode, thunks) = _generate_rust_gencode(
	File "/workspace/rust/aspects.bzl", line 131, column 25, in _generate_rust_gencode
		proto_lang_toolchain.output_files = ctx.bin_dir.path + "/" + proto_root
Error: struct value does not support field assignment
```

tempoz:
Sorry, I must have lost track of things and not pushed my local changes; this was a failed intermediate state

tempoz:
Should be corrected now.

acozzette (approved):

PR #18261, opened 2024-09-11
Fix includes for `hpb` and `hpb_generator`

tempoz:
These include paths are pointing to files that do not exist


googleberg:
@honglooker how do you want to address unbuildable hpb bits?

honglooker:
Hey @tempoz, thanks for taking a stab! This is likely better tackled internally after our migration is complete. I plan on tackling this; if you're curious, please feel free to ask about bug `366171162` later on -- thanks!

PR #18262, opened 2024-09-11
Update absl to get rid of OOB warning

tempoz:
Without this, there are many instances of a warning about OOB indices inside of string_view in absl.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -14,7 +14,7 @@ module(
 # Thus the highest version in their module graph is resolved.
 bazel_dep(
     name = "abseil-cpp",
-    version = "20230802.0.bcr.1",
+    version = "20240722.0",
     repo_name = "com_google_absl",
 )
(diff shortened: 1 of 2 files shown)


tempoz:
https://github.com/abseil/abseil-cpp/commit/50d39219dc5a63f26d3871cfb3dbe198091eadad

This is the fix I'm trying to pick up. Up to you!

zhangskz:
I don't think it makes sense for protobuf to update its absl *lower-bound* to avoid a spurious warning if its otherwise still compatible, as this could cause other users to fail to find a valid absl version. IIUC, minimal version selection means bzlmod should typically still resolve to a higher version of absl if possible.

It should be sufficient to handle this in the root module via [overrides](https://bazel.build/versions/6.0.0/rules/lib/globals#single_version_override) to pin the absl version or by silencing warnings if needed.

PR #18264, opened 2024-09-11
Add `unsafe` keyword to unsafe block in rust code

tempoz:


tempoz:
Done!
Hopefully I added the `allow_unused_unsafe` correctly; I'm not really very familiar with Rust.

PR #18265, opened 2024-09-11
Comment out non-building rust targets

tempoz:
Comment out all non-building rust targets. This change is in furtherance of my effort to make the `protocolbuffers/protobuf` repo build free of errors or warnings to make it easier to detect regressions.

### DIFF
--- a/rust/test/BUILD
+++ b/rust/test/BUILD
@@ -18,73 +18,75 @@ UNITTEST_PROTO3_OPTIONAL_TARGET = "//src/google/protobuf:test_protos"
 
 UNITTEST_EDITION_TARGET = "//src/google/protobuf:test_protos"
 
-rust_upb_proto_library(
-    name = "unittest_upb_rust_proto",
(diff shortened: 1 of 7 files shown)


googleberg:
@esrauchg How are we addressing the unbuildability of the rusty bits?

esrauchg:
Thanks for this report / PR showing the issue @tempoz.

It is intended that these targets should work, but due to historical reasons these targets weren't configured to run on GHA and so many of them naturally broke without us noticing.

Unfortunately we can't accept this PR in its current shape as it would merge badly with our internal copy (where these targets do all work).

I can prioritize getting the corresponding fix on our side (and fixing our GHA Rust config so we would catch such problems in the future).

tempoz:
Of course, close whichever of these aren't helpful. I just was trying to get the repo to build without warnings, and when I eventually succeeded, I figured I'd submit my changes to you guys and you could accept whatever was useful.

PR #18405, opened 2024-09-18
Fix packed reflection handling bug in edition 2023.

mkruskal-google:
This is a very narrow edge case where touching a packed extension via generated APIs first, and then doing so reflectively will trigger a DCHECK.  Otherwise, reflective APIs will work but not use packed encoding for the extension.  This was likely a pre-existing bug dating back to proto3, where it would only be visible on custom options (the only extensions allowed in proto3).

To help qualify this and uncover similar issues, unittest.proto was migrated to editions.  This turned up some other minor issues in DebugString and python.

PiperOrigin-RevId: 675785611


tonyliaoss (approved):

Joj501 (commented):
Submit changes to update

PR #18507, opened 2024-09-26
docs: add options for `protoc-gen-env` and `protoc-gen-default` to registry

MarnixBouhuis:
Reserved field options used for `protoc-gen-env` and `protoc-gen-default`, implemented here: https://github.com/MarnixBouhuis/confpb

I've recently released a protoc plugin for generating defaults / binding environment variables to go protobuf structs. For this I would like to reserve the field options used.


esrauchg (approved):

acozzette:
@MarnixBouhuis In the time since you created this PR, we seem to have already allocated these numbers to someone else unfortunately. Is it OK if we give you different numbers instead? I see that you're already using 1200 and 1201, so you would have to change that. That could technically be a breaking wire format change for anyone working with serialized descriptors that refer to the affected options, but in practice I think this would be unlikely to break anything since your project was just released two weeks ago.

MarnixBouhuis:
Thanks everyone for looking at this PR this and approving it!

@acozzette yeah just saw the merge conflict. I'm fine with changing the allocated field options numbers, I can just do a major semver release.
Do you want me to update the PR or do you want to do it?

MarnixBouhuis:
Was able to quickly fix it in the github UI, claimed a different range.

PR #18610, opened 2024-10-04
Clean up PHP CI

mkruskal-google:
This PR does the following:
* Special-case descriptor.proto to allow for codegen despite being proto2
* Fix a pre-existing bug that gets exercised now during descriptor builds
* Expand test coverage of conformance tests, and fix pre-existing issues
* Properly hook up gencode to staleness infrastructure for automated regen


bshaffer (commented):
This looks great, but shouldn't this also update the stale PHP gencode?

mkruskal-google:
Nope, as discussed off line stale file regen is handled automatically post-submit.  Tests don't run against the stale checked-in code

bshaffer (approved):

PR #18641, opened 2024-10-07
Rename maven to protobuf_maven in MODULE.bazel

### LINKED ISSUE #16839: Using `maven` as the repo name causes duplicate warnings when using bzlmod
Downstream bug in rules_jvm_external: https://github.com/bazelbuild/rules_jvm_external/issues/916

v23.1

A clean build when using bzlmod.

**What did you see instead?**

```
DEBUG: $TMP/external/rules_jvm_external~/private/extensions/maven.bzl:154:14:
The maven repository 'maven' is used in two different bazel modules,
originally in '<my_workspace>' and now in 'protobuf'
```

Summarizing @shs96c in https://github.com/bazelbuild/rules_jvm_external/issues/916#issuecomment-1645527584

The common `maven` repo name allows rulesets to contribute to the user's JARs. However, this implies that `maven` is for the end user, not for transitive dependencies. If a ruleset needs private dependencies, it should use a custom namespace rather than the `maven` namespace.

- rules_jvm_external uses [rules_jvm_external_deps](https://github.com/bazelbuild/rules_jvm_external/blob/0dfbe26a16f78d4669c8c11b2f0d79da8c086062/MODULE.bazel#L48) instead of `maven`.
- rules_kotlin uses [kotlin_rules_maven](https://github.com/bazelbuild/rules_kotlin/blob/42ea2e5539ec52ef93cf29607aa889b7d57ea320/MODULE.bazel#L54) instead of `maven`.

**Proposed fixes**

There are a few ways to fix the logline:

1. Patch protobuf in the Bazel Central Registry rewrite `maven` to `protobuf_maven`.
2. Modify the protobuf repo to rewrite `maven` to `protobuf_maven`.
3. Fetch the JARs manually without using Maven or rules_jvm_external. Trims the dependency graph a fair bit.

I think 2 is the best short-term option. If you're open to it, I can send a PR.

jschaf:
The default name for the rules_jvm_external.maven rule is "maven". When not set, it defaults to "maven". For root modules also using rules_jvm_external, the name clash causes a warning:

```
DEBUG: $TMP/external/rules_jvm_external~/private/extensions/maven.bzl:154:14:
The maven repository 'maven' is used in two different bazel modules,
originally in '<my_workspace>' and now in 'protobuf'
```

Summarizing @shs96c in [1]:

> The common maven repo name allows rulesets to contribute to the user's JARs.
> However, this implies that maven is for the end user, not for transitive
> dependencies. If a ruleset needs private dependencies, it should use a custom
> namespace rather than the maven namespace.

Since protobuf is not contributing to user's JARs, we'll use a custom namespace. There's precedent for using a custom namespace for library modules:

- rules_jvm_external uses `rules_jvm_external_deps` instead of `maven`.
- rules_kotlin uses `kotlin_rules_maven` instead of `maven`.

[1]: https://github.com/bazel-contrib/rules_jvm_external/issues/916#issuecomment-1645527584

Fixes #16839.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -63,6 +63,7 @@ rules_jvm_external_setup()
 load("@rules_jvm_external//:defs.bzl", "maven_install")
 
 maven_install(
+    name = "protobuf_maven",
     artifacts = PROTOBUF_MAVEN_ARTIFACTS,
     # For updating instructions, see:
     # https://github.com/bazelbuild/rules_jvm_external#updating-maven_installjson
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -73,7 +73,7 @@ maven_install(
     ],
 )
 
-load("@maven//:defs.bzl", "pinned_maven_install")
+load("@protobuf_maven//:defs.bzl", "pinned_maven_install")
 
 pinned_maven_install()
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -158,6 +158,7 @@ use_repo(
 maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
 
 maven.install(
+    name = "protobuf_maven",
     artifacts = [
         "com.google.caliper:caliper:1.0-beta-3",
         "com.google.code.findbugs:jsr305:3.0.2",
@@ -178,7 +179,7 @@ maven.install(
     ],
 )
 
-use_repo(maven, "maven")
+use_repo(maven, "protobuf_maven")
 
 # Development dependencies
 bazel_dep(
--- a/java/core/BUILD.bazel
+++ b/java/core/BUILD.bazel
@@ -390,8 +390,8 @@ protobuf_java_library(
         ":generic_test_protos_java_proto",
         ":java_test_protos_java_proto",
         ":lite_test_protos_java_proto",
-        "@maven//:com_google_guava_guava",
-        "@maven//:junit_junit",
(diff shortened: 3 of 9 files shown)


deannagarcia (approved):
Thanks for the fix!

WORKSPACE
```diff
@@ -73,7 +73,7 @@ maven_install(
     ],
 )
 
-load("@maven//:defs.bzl", "pinned_maven_install")
+load("@protobuf_maven//:defs.bzl", "pinned_maven_install")
```
JasonLunn, line 76: Is this change correct? In the error logs for the failing tests we see:

```
ERROR: Failed to load Starlark extension '@@protobuf_maven//:defs.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
```

jschaf: I think the problem was I forgot to add the `name = "protobuf_maven"` in the WORKSPACE. I only did it in MODULE.bazel for the first commit.

acozzette (approved):

ejona86:
FYI: this looks broken for protobuf-util, as that does contribute to user's JARs. See my comment on a PR to do something similar in gRPC: https://github.com/grpc/grpc-java/pull/11614#issuecomment-2414518707

JasonLunn:
@ejona86 - can you open a new issue for the protobuf-util problem?

PR #18768, opened 2024-10-10
Cherry-pick: Fix Bazel dependency cycle issue affecting Kythe

acozzette:
The fact that our `:protobuf_nowkt` target actually does depend on the well-known types is causing a dependency cycle for Kythe. This fixes that so that `:protobuf_nowkt` no longer depends on the well-known types.

PiperOrigin-RevId: 684160567

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -266,7 +266,7 @@ alias(
 
 alias(
     name = "protobuf_nowkt",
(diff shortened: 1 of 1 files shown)


AdnanJutt007:
exustaadbolty94@gmail.com

AdnanJutt007:
Mobile data sim card activit krde redmi10c please 🥺 network Sim service on

PR #18820, opened 2024-10-15
Cherry-picks for Bazel 8

comius:
Commits:
9c4f3b817ed0b1cfcfdca66c2c29874a5bbb51fd Prepare supporting targets for testing 
b00dca3665b872f280702283be831bad6607d86c Load Python rules everywhere in protobuf 
5aa7abc659f727de9ca90ca37a3a3187d88a790f Upgrade rules_cc to 0.0.13 
8ec13ede1f4e07ca88a71d02aca004e3c281700a  Load Shell rules everywhere in protobuf
a0eb24061f3850824eaba4b9e8c627aa2b4257ff Replace use of C++17 with disabling a warning

Pending commits:
a03850971866aa61f3e2966eaca3db3ee2ed19d9 Convert proto toolchain string to Label (cl/684391633)
f484e9caf77e1550ca65cfe67a674cb979dd62c1 Use rules_cc everywhere in protobuf (cl/674352147)

### DIFF
--- a/.github/workflows/test_cpp.yml
+++ b/.github/workflows/test_cpp.yml
@@ -45,7 +45,7 @@ jobs:
             cache_key: Bazel7
             image: "us-docker.pkg.dev/protobuf-build/containers/common/linux/bazel:7.1.2-cf84e92285ca133b9c8104ad7b14d70e953cbb8e"
             targets: "//src/... //third_party/utf8_range/..."
-          - config: { name: "Bazel7 with Bzlmod", flags: --enable_bzlmod --enable_workspace }
+          - config: { name: "Bazel7 with Bzlmod", flags: --enable_bzlmod --enable_workspace --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --cxxopt="-Wno-self-assign-overloaded" }
             cache_key: Bazel7bzlmod
             image: "us-docker.pkg.dev/protobuf-build/containers/common/linux/bazel:7.1.2-cf84e92285ca133b9c8104ad7b14d70e953cbb8e"
             targets: "//src/... //third_party/utf8_range/..."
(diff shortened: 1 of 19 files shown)


.github/workflows/test_cpp.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             cache_key: Bazel7
             image: "us-docker.pkg.dev/protobuf-build/containers/common/linux/bazel:7.1.2-cf84e92285ca133b9c8104ad7b14d70e953cbb8e"
             targets: "//src/... //third_party/utf8_range/..."
-          - config: { name: "Bazel7 with Bzlmod", flags: --enable_bzlmod --enable_workspace }
+          - config: { name: "Bazel7 with Bzlmod", flags: --enable_bzlmod --enable_workspace --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --cxxopt="-Wno-self-assign-overloaded" }
```
zhangskz, line 48: This should ideally stay on c++14 in 29.x, if silencing the warnings without bumping to Bazel 17 is sufficient here. We're not raising our minimum C++ version to 17 until 30.x

zhangskz:
We also still need another change to remove `--experimental_google_legacy_api` right?

.github/workflows/test_cpp.yml
comius: I can't silence the warning, because it's enabled on abseil:
https://github.com/abseil/abseil-cpp/blob/master/absl/strings/BUILD.bazel#L91

comius:
> We also still need another change to remove `--experimental_google_legacy_api` right?

Perhaps let's do this as a separate PR? I still need to prepare a patch release of rules_java, it will take a couple of days.

.github/workflows/test_cpp.yml
comius: Done. Disable using --per_file_copt

zhangskz (approved):

PR #18825, opened 2024-10-15
Prevent Bazel lockfile non reproducibility for end users

cerisier:
Using `crate.from_specs` results in non reproducible `MODULE.bazel.lock` between platforms for end users of the `protobuf` bazel module.

Using `crate.from_cargo` instead (which marks the crate module extension as reproducible) makes bazel skips the entry in the lockfile altogether and allows reproducible lockfiles for end users of the `protobuf` bazel module.

cc @fmeum

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -68,9 +68,10 @@ pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
 use_repo(pip, "pip_deps")
 
 crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
-# As of rules_rust@0.52.2 crate.from_specs specifies the crate module extension
-# as non reproducible and non os/arch dependent, which can cause bazel lockfile
-# non reproducibility between platforms for users of the `protobuf` module.
+# As of rules_rust@0.52.2 crate.from_specs still specifies the crate
+# module extension as non reproducible and non os/arch dependent, which can
+# cause bazel lockfile non reproducibility between platforms for users of the
+#`protobuf` module.
 #
 # crate.from_cargo is specified reproducible since using Cargo.lock mechanism
 # directly and does not result in a Bazel lockfile entry for the crate module
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -68,15 +68,20 @@ pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
 use_repo(pip, "pip_deps")
 
 crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
-crate.spec(
-    package = "googletest",
-    version = ">0.0.0",
+# As of rules_rust@0.52.2 crate.from_specs specifies the crate module extension
+# as non reproducible and non os/arch dependent, which can cause bazel lockfile
+# non reproducibility between platforms for users of the `protobuf` module.
+#
+# crate.from_cargo is specified reproducible since using Cargo.lock mechanism
+# directly and does not result in a Bazel lockfile entry for the crate module
+# extension.
+crate.from_cargo(
+    name = "crates",
+    cargo_lockfile = "Cargo.lock",
+    manifests = [
+        "//rust/cargo:Cargo.toml",
+    ],
 )
-crate.spec(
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -68,15 +68,20 @@ pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
 use_repo(pip, "pip_deps")
 
 crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
```
dzbarsky, line 70: isn't this code gone? https://github.com/protocolbuffers/protobuf/pull/18650/files

dzbarsky: oh I see it's removed in 29.x but not main...

MODULE.bazel
```diff
@@ -68,15 +68,20 @@ pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
...
-crate.spec(
-    package = "googletest",
-    version = ">0.0.0",
-)
-crate.spec(
-    package = "paste",
-    version = ">=1",
+# As of rules_rust@0.52.2 crate.from_specs specifies the crate module extension
```
dzbarsky, line 71: nitL: I believe crate.from_specs *always* specified it as non-reproducible (I wrote it but my memory is fuzzy at times :) )

MODULE.bazel
cerisier: Sorry I'm not familiar with the branching / PR process here ! I guess this PR doesn't make sence if it's going to be removed in 29.x !

MODULE.bazel
cerisier: Right sorry, I meant that, as of 0.52.2, it is "still" non-reproducible !

MODULE.bazel
dzbarsky: I'm not sure, I've never worked in this repo, I was also confused that the commit landed to a release branch but not main :)

esrauchg: Yeah, sorry that part is confusing: the code is temporarily removed for the 29 release towards debloating the release for people who don't use Rust. It was made directly on the 29 release branch and not made on main since there's ongoing work that will decide how to avoid some of the bloat concerns for our 30+ releases.

So its worth still looking at improvements here, though know that its not being included for 29 and things made also change for 30+

rust/cargo/BUILD
```diff
@@ -0,0 +1,3 @@
+exports_files([
+    "Cargo.toml",
```
esrauchg, line 2: This Cargo.toml file is part of our under-development work where we intend to build a crates.io release from bazel; I don't think we expected it to be used for the Bazel flow when it was introduced.

Looking at the Cargo.toml though the only maybe-problematic thing is that it declares a version number for the `protobuf` crate and its going to ~always be arbitrary and wrong on main. But I suppose that is effectively true anyway for all Cargo.toml on main branches?

dzbarsky: Does that mean it's only supposed to be used by this module? if end users are not supposed to use this stuff can it be marked `dev_dependency = True` in MODULE.bazel?

cerisier: You

> This Cargo.toml file is part of our under-development work where we intend to build a crates.io release from bazel; I don't think we expected it to be used for the Bazel flow when it was introduced.
> 
> Looking at the Cargo.toml though the only maybe-problematic thing is that it declares a version number for the `protobuf` crate and its going to ~always be arbitrary and wrong on main. But I suppose that is effectively true anyway for all Cargo.toml on main branches?

Correct, I think the right way is to create a workspace Cargo.toml and delcare dependencies thhere and have the sub crates depends on the dependencies from the workspace. I'll amend.

zhangskz: If this is only used to build the crates.io as part of our release process, marking as a dev dependency seems appropriate.

The version being a tentative -dev version at head is always the case, but we can add that file to our release process's version updaters so that the version number is updated to the proper version on release (and stays at a sensible "-dev" version).

cerisier:
I'll wait until rust is reintroduced after 29.x before fixing this then ! Thanks.

PR #18873, opened 2024-10-16
Delete repo's submodules

mkruskal-google:
These have been a devex pain for a long time, and have only stuck around to support cmake workflows.  Our cmake setup now uses FetchContent instead, and they're not needed.


zhangskz:
I like the branch name :)

zhangskz (approved):

PR #18880, opened 2024-10-17
Fix order of `inline` qualifiers for MSVC

fmeum:
This gets rid of this very noisy warning when compiling with MSVC:

```
C:\b\z46qjgck\execroot\_main\bazel-out\x64_windows-fastbuild\bin\external\protobuf+\src\google\protobuf\_virtual_includes\protobuf_lite\google/protobuf/message_lite.h(1191): warning C4141: 'inline': used more than once
C:\b\z46qjgck\execroot\_main\bazel-out\x64_windows-fastbuild\bin\external\protobuf+\src\google\protobuf\_virtual_includes\protobuf_lite\google/protobuf/message_lite.h(1297): warning C4141: 'inline': used more than once
```

### DIFF
--- a/src/google/protobuf/message_lite.h
+++ b/src/google/protobuf/message_lite.h
@@ -1183,7 +1183,7 @@ inline void AssertDownCast(const MessageLite& from, const MessageLite& to) {
 }
 
 template <bool test_call, typename MessageLite>
-PROTOBUF_ALWAYS_INLINE MessageLite* MessageCreator::PlacementNew(
+inline PROTOBUF_ALWAYS_INLINE MessageLite* MessageCreator::PlacementNew(
     const MessageLite* prototype_for_func,
     const MessageLite* prototype_for_copy, void* mem, Arena* arena) const {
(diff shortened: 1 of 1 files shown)


fmeum:
@meteorcloudy FYI, this causes Windows logs in Bazel CI to initially overflow the Buildkite limit for streamable logs.

zhangskz (approved):

zhangskz (changes requested):
@sbenzaquen pointed out in review that these are not duplicate since we can't mark a function always_inline if it is not inline, and tht they likely have the be in the opposite order to make MSVC happier.

fmeum:
@zhangskz Changed

zhangskz:
@sbenzaquen Can you take a look?

mkruskal-google (approved):

tonyliaoss:
@fmeum Looks like there is a merge conflict. Could you rebase? I can kick off a new round of tests afterwards.

tonyliaoss:
Actually, I think this is addressed by https://github.com/protocolbuffers/protobuf/commit/8a33e2b5bc5efc4db9c9e0a93b95bcd02f502f27. Please reopen that commit didn't resolve your issue.

PR #18882, opened 2024-10-17
GitHub Actions:  Add Python 3.13 to the testing

cclauss:
* https://www.python.org/downloads/release/python-3130/
* https://pythoninsider.blogspot.com/2024/10/python-3130-final-released.html
* https://devguide.python.org/versions/


zhangskz:
Thanks for the contribution! This is actually on our radar / in progress already and requires a few additional changes to our docker images and other setup.

See https://github.com/protocolbuffers/protobuf/pull/18270 for a draft of our PR testing out -rc2 prior to 3.13.0 final release. I believe this will be replaced by a separate PR when we update it for 3.13.0.

PR #18890, opened 2024-10-17
Fix uses of __cpp_lib_constexpr_string

keith:
The intent of these checks was that if the C++20 version of this feature
are enabled to execute the top codepath. libstdc++ 12 defines this
feature as:

```cpp
// in: /usr/include/c++/12/bits/basic_string.h

#ifdef __cpp_lib_is_constant_evaluated
// Support P0980R1 in C++20.
# define __cpp_lib_constexpr_string 201907L
#elif __cplusplus >= 201703L && _GLIBCXX_HAVE_IS_CONSTANT_EVALUATED
// Support P0426R1 changes to char_traits in C++17.
# define __cpp_lib_constexpr_string 201611L
#endif
```

So this codepath was always being hit even with -std=c++17 and then
resulted in a failure because the string wasn't actually constinit. This
matches the other use of this feature check in `inlined_string_field.h`

### DIFF
--- a/src/google/protobuf/port.cc
+++ b/src/google/protobuf/port.cc
@@ -97,7 +97,7 @@ void RealDebugCounter::Register(absl::string_view name) {
   }
 }
 
-#if defined(__cpp_lib_constexpr_string)
+#if defined(__cpp_lib_constexpr_string) && __cpp_lib_constexpr_string >= 201907L
 PROTOBUF_ATTRIBUTE_NO_DESTROY PROTOBUF_CONSTINIT const GlobalEmptyString
     fixed_address_empty_string{};
 #else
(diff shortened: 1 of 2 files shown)


zhangskz:
Looks like there's a conflict -- can you rebase / resolve? `upb / No System Python` test also seems to be failing (incl. rerun) but seems unrelated afaict. @haberman thoughts?

keith:
looks like the commit that introduced this was rolled back https://github.com/protocolbuffers/protobuf/commit/f58849b60344fea170103ce7ae83d34d5a489c5b

maybe a googler can comment on if that's going to come back out, or if this patch is irrelevant

tonyliaoss:
Change https://github.com/protocolbuffers/protobuf/commit/bd03560d9f17487737efc71dd03f21141a942f29 got rolled back in https://github.com/protocolbuffers/protobuf/commit/f58849b60344fea170103ce7ae83d34d5a489c5b due to some Google internal code relying on UB and breaking as a result :) It's definitely something we want to eventually ship out again once we fix all the internal breakages.

sbenzaquen:
The change was submitted again.

sbenzaquen (approved):

keith:
reabsed!@

tonyliaoss (approved):
Approving for tests

PR #18934, opened 2024-10-22
dotnet: Dispose StringWriters. Microoptimizations

Henr1k80:
Comparing against empty string will first do a reference comparison, then value comparison.

Whitespace changes are caused by `trim_trailing_whitespace = true` in  `.editorconfig`

### DIFF
--- a/csharp/src/Google.Protobuf/ByteString.cs
+++ b/csharp/src/Google.Protobuf/ByteString.cs
@@ -140,7 +140,7 @@ namespace Google.Protobuf
         {
             // By handling the empty string explicitly, we not only optimize but we fix a
             // problem on CF 2.0. See issue 61 for details.
-            return bytes == "" ? Empty : new ByteString(Convert.FromBase64String(bytes));
+            return bytes.Length == 0 ? Empty : new ByteString(Convert.FromBase64String(bytes));
         }
 
         /// <summary>
--- a/csharp/src/Google.Protobuf/JsonFormatter.cs
+++ b/csharp/src/Google.Protobuf/JsonFormatter.cs
@@ -85,7 +85,7 @@ namespace Google.Protobuf {
 
     static JsonFormatter() {
       for (int i = 0; i < CommonRepresentations.Length; i++) {
-        if (CommonRepresentations[i] == "") {
+        if (CommonRepresentations[i].Length == 0) {
           CommonRepresentations[i] = ((char)i).ToString();
         }
       }
@@ -123,7 +123,7 @@ namespace Google.Protobuf {
     /// </remarks>
     /// <returns>The formatted message.</returns>
     public string Format(IMessage message, int indentationLevel) {
-      var writer = new StringWriter();
+      using var writer = new StringWriter();
       Format(message, writer, indentationLevel);
       return writer.ToString();
     }
@@ -291,7 +291,7 @@ namespace Google.Protobuf {
       return descriptor.FieldType switch {
         FieldType.Bool => (bool)value == false,
         FieldType.Bytes => (ByteString)value == ByteString.Empty,
-        FieldType.String => (string)value == "",
+        FieldType.String => ((string)value).Length == 0,
         FieldType.Double => (double)value == 0.0,
         FieldType.SInt32 or FieldType.Int32 or FieldType.SFixed32 or FieldType.Enum =>
             (int)value == 0,
@@ -497,7 +497,7 @@ namespace Google.Protobuf {
       WriteBracketClose(writer, ObjectCloseBracket, true, indentationLevel);
     }
 
-    private void WriteDiagnosticOnlyAny(TextWriter writer, IMessage value) {
+    private static void WriteDiagnosticOnlyAny(TextWriter writer, IMessage value) {
       string typeUrl =
           (string)value.Descriptor.Fields[Any.TypeUrlFieldNumber].Accessor.GetValue(value);
       ByteString data =
--- a/csharp/src/Google.Protobuf/JsonParser.cs
+++ b/csharp/src/Google.Protobuf/JsonParser.cs
@@ -830,7 +830,7 @@ namespace Google.Protobuf
                 // TODO: It would be nice not to have to create all these objects... easy to optimize later though.
                 Timestamp timestamp = Timestamp.FromDateTime(parsed);
                 int nanosToAdd = 0;
-                if (subseconds != "")
+                if (subseconds.Length != 0)
                 {
                     // This should always work, as we've got 1-9 digits.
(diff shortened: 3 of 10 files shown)


csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
```diff
@@ -63,15 +63,15 @@ static FileDescriptor()
...
+        private FileDescriptor(ByteString descriptorData, FileDescriptorProto proto, IList<FileDescriptor> dependencies, DescriptorPool pool, bool allowUnknownDependencies, GeneratedClrTypeInfo generatedCodeInfo)
         {
             SerializedData = descriptorData;
             DescriptorPool = pool;
             Proto = proto;
             // Note: the Edition property relies on the proto being set first, so this line
             // has to come after Proto = proto.
             Features = FeatureSetDescriptor.GetEditionDefaults(Edition).MergedWith(proto.Options?.Features);
-            Dependencies = new ReadOnlyCollection<FileDescriptor>(dependencies.ToList());
+            Dependencies = new ReadOnlyCollection<FileDescriptor>(dependencies);
```
jskeet, line 74: I disagree with this change. It means that if the dependencies have been passed in externally (as they are from FromGeneratedCode) we don't end up making a defensive copy, and the collection could be modified. Yes, often this means we make a defensive copy unnecessarily - but I'd personally rather have that than have weird potential effects due to mutation.

jskeet (commented):
I'm not really sure it's worth disposing of StringWriter (as that will do nothing) but I doubt that it'll do any harm, either.

I'm not sure about all the length checks - these look less readable to me. Do we have any benchmarks that show this actually makes any difference at all? (I wouldn't be at all surprised to find that the JIT can optimize this anyway.) If there *is* a performance improvement, that's fine - I'd just rather not do it without evidence that it's worth what I consider to be a readability hit.

csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs
Henr1k80: Yeah, I can see that it is obsoleted as well.
I have reverted the removal of the copy

Henr1k80:
Oh, I had not checked the actual implementation of StringWriter, it just inherits IDisposable from TextReader.
I will revert disposal on these, no reason to spend time calling dispose.

Henr1k80:
JIT assembly difference for empty string check can be [seen here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKBuIGYACMxgYUYG8bGfGBLAHYYYUAdgA2zckkbAIESQCUAFMXIAGRgDN5ASk7deRo8QDs2+YwC8VxgCI7AbkM8Avi/5CRYyWplyFRgAZVQ0LCF0uamNeM3CAOiCYAQBzDAALa1t1Z2jed2pXIA==)

jskeet:
Thanks for the JIT difference. I do wonder whether progressive JITting might improve that to just the length check, mind you. Or maybe the JIT expects the reference comparison to be quicker, on the grounds that it doesn't need to access the data within the string object - in which case we might want to consider that too. I might do some benchmarking myself to see if I can record any differences one way or the other.

It's only just struck me - and it really should have done so before - that's there's another huge difference between `x == ""` and `x.Length == 0` in terms of what happens in `x` is null. I suspect that in all of these cases we *do* genuinely know that the LHS is non-null, but I'll need to review that more carefully when I get a chance. (That may well not even be this week I'm afraid.)

csharp/src/Google.Protobuf/ByteString.cs
```diff
@@ -140,7 +140,7 @@ public static ByteString FromBase64(string bytes)
         {
             // By handling the empty string explicitly, we not only optimize but we fix a
             // problem on CF 2.0. See issue 61 for details.
-            return bytes == "" ? Empty : new ByteString(Convert.FromBase64String(bytes));
+            return bytes.Length == 0 ? Empty : new ByteString(Convert.FromBase64String(bytes));
```
jskeet, line 143: This will change from ArgumentNullException (thrown by Convert.FromBase64String) to NullReferenceException if `bytes` is null.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -85,7 +85,7 @@ public sealed class JsonFormatter {
 
     static JsonFormatter() {
       for (int i = 0; i < CommonRepresentations.Length; i++) {
-        if (CommonRepresentations[i] == "") {
+        if (CommonRepresentations[i].Length == 0) {
```
jskeet, line 88: This will be fine.

csharp/src/Google.Protobuf/JsonFormatter.cs
```diff
@@ -291,7 +291,7 @@ private static bool IsDefaultValue(FieldDescriptor descriptor, object value) {
       return descriptor.FieldType switch {
         FieldType.Bool => (bool)value == false,
         FieldType.Bytes => (ByteString)value == ByteString.Empty,
-        FieldType.String => (string)value == "",
+        FieldType.String => ((string)value).Length == 0,
```
jskeet, line 294: This *should* be fine, as a field with a FieldType of string should never have a null value.

csharp/src/Google.Protobuf/JsonParser.cs
```diff
@@ -830,7 +830,7 @@ private static void MergeTimestamp(IMessage message, JsonToken token)
                 // TODO: It would be nice not to have to create all these objects... easy to optimize later though.
                 Timestamp timestamp = Timestamp.FromDateTime(parsed);
                 int nanosToAdd = 0;
-                if (subseconds != "")
+                if (subseconds.Length != 0)
```
jskeet, line 833: This is fine - it's from a matched regex. Same at line 909.

csharp/src/Google.Protobuf/Reflection/FieldDescriptor.cs
```diff
@@ -120,7 +120,7 @@ internal FieldDescriptor(FieldDescriptorProto proto, FileDescriptor file,
             // a MapField, but that feels a tad nasty.
             PropertyName = propertyName;
             Extension = extension;
-            JsonName =  Proto.JsonName == "" ? JsonFormatter.ToJsonName(Proto.Name) : Proto.JsonName;
+            JsonName =  Proto.JsonName.Length == 0 ? JsonFormatter.ToJsonName(Proto.Name) : Proto.JsonName;
```
jskeet, line 123: This should be fine (because the value should never be null; it's a string proto field)

jskeet (commented):
I'd suggest reverting the `.Length` check in ByteString. It looks like the rest are okay.

PR #18944, opened 2024-10-22
Add package only if not empty

### LINKED ISSUE #18943: [Python] Cannot use reflection to find a service by name if service does not have package
Version: main
Language: Python

All

Not applicable

**What did you do?**
1. Deploy gRPC service that does not have a package
2. Use reflection and `DescriptorPool` to find service by name
3. KeyError is being raised

It should be possible to find a service even though it does not have a pckage

I have debugged through the code and the culprit is [this](https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/descriptor_database.py#L50-L62) code that assumes that package name is not empty. I happened to have an example where this was not the case.

The fix I think is trivial - just check if package is empty or not before concatenating it with an object name.

inikolaev:
Fixes #18943 

Changes:
* Added a helper function `_AddPackage` that checks whether package is not empty before prepending it to the name
* Changed affected block of code to use new function

### DIFF
--- a/python/google/protobuf/descriptor_database.py
+++ b/python/google/protobuf/descriptor_database.py
@@ -52,14 +52,14 @@ class DescriptorDatabase(object):
       for name in _ExtractSymbols(message, package):
         self._AddSymbol(name, file_desc_proto)
     for enum in file_desc_proto.enum_type:
-      self._AddSymbol(('.'.join((package, enum.name))), file_desc_proto)
+      self._AddSymbol((_AddPackage((package, enum.name))), file_desc_proto)
       for enum_value in enum.value:
         self._file_desc_protos_by_symbol[
-            '.'.join((package, enum_value.name))] = file_desc_proto
+            _AddPackage((package, enum_value.name))] = file_desc_proto
     for extension in file_desc_proto.extension:
-      self._AddSymbol(('.'.join((package, extension.name))), file_desc_proto)
+      self._AddSymbol((_AddPackage((package, extension.name))), file_desc_proto)
     for service in file_desc_proto.service:
-      self._AddSymbol(('.'.join((package, service.name))), file_desc_proto)
+      self._AddSymbol(_AddPackage((package, service.name)), file_desc_proto)
 
   def FindFileByName(self, name):
     """Finds the file descriptor proto by file name.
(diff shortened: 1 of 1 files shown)


anandolee:
Can you help to add a test al well?

inikolaev:
> Can you help to add a test al well?

There seems to be some existing test-cases for no-package protobuf, so I modified it a bit. I haven't yet figured out how to run tests locally though.

inikolaev:
Alright, I figured out the tests:

```bash
$ bazel test //python:descriptor_database_test
INFO: Analyzed target //python:descriptor_database_test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Target //python:descriptor_database_test up-to-date:
  bazel-bin/python/descriptor_database_test
INFO: Elapsed time: 0.825s, Critical Path: 0.49s
INFO: 2 processes: 2 darwin-sandbox.
INFO: Build completed successfully, 2 total actions
//python:descriptor_database_test                                        PASSED in 0.4s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

anandolee:
As there's existing test and check prefix ".", I guess this behavior change may break some of our internal projects. I will double check with all google internal projects to see if this change is safe to make

inikolaev:
> As there's existing test and check prefix ".", I guess this behavior change may break some of our internal projects. I will double check with all google internal projects to see if this change is safe to make

Please do check, I just wanted to add that without this fix the reflection feature is broken:
* it's impossible to find service with a name without dot, such as `MyService`, because the library breaks
* it's impossible to find a service with a name with dot, such as `.MyService`, because remote gRPC service will not find matching service definition

Alternative solution could be to try both combinations in this case, but that adds a special case, which might be suboptimal as well.

anandolee:
Hi @inikolaev  change the behavior directly breaks a lot of our internal projects. I've made change internally to accept both, but raise warnings for prefix ".".  It has merged to github by https://github.com/protocolbuffers/protobuf/commit/610702ed18d4323e44b9741102ed90377243470e

We will keep accept prefix "." for one more year and remove the support around 2026 Jan (Will clean up google internal project before the breaking change)

Thanks for raise this issue up

anandolee:
Thank you again @inikolaev  
Will close this PR as the need is now supported

PR #18958, opened 2024-10-23
Only print pre-release version if it makes sense to do so

imrichardcole:
The pre-release version was printed even if it was blank.  Move the message inside a pre-existing if condition so it'll only get printed if it's set.


imrichardcole:
@zhangskz reraised a new PR which is rebased and updated based on the previous comment.

zhangskz (approved):

PR #19009, opened 2024-10-28
utf8_range: add version marker to library

jengelh:
Unversioned libraries / libraries without due ABI indicators are not allowed in certain Linux distributions because it precludes the concurrent presence of multiple versions.

If you have both /usr/lib/libprotobuf-lite.so.28.3.0 and /usr/lib/libprotobuf-lite.so.29.0.0, both of them want libutf8_validity.so, but if the ABI is different between utf8_range 28 and utf8_range 29, that's a problem.

### DIFF
--- a/third_party/utf8_range/CMakeLists.txt
+++ b/third_party/utf8_range/CMakeLists.txt
@@ -19,6 +19,15 @@ add_library (utf8_range
 # A heavier-weight C++ wrapper that supports Abseil.
 add_library (utf8_validity utf8_validity.cc utf8_range.c)
 
+set_target_properties(utf8_range PROPERTIES
+	VERSION ${protobuf_VERSION}
+	OUTPUT_NAME ${LIB_PREFIX}utf8_range
+)
+set_target_properties(utf8_validity PROPERTIES
+	VERSION ${protobuf_VERSION}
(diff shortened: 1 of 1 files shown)


acozzette:
@jengelh Thank you for this fix, but unfortunately we can't review it until you sign the CLA.

jengelh:
I've long signed it, what's the hold-up?

tonyliaoss:
> not allowed in certain Linux distributions

@jengelh Could you elaborate which platforms you are seeing this on?

The idea behind this change SGTM. It looks like the PR doesn't pass our GHA right now though -- I'm seeing some failures in our Python tests.

jengelh:
By definition, the lack of a cmake `VERSION` property affects all platforms.

Manifestion happens on all ELF targets (and probably *-pc-cygwin) with shared library configuration (`cmake -DBUILD_SHARED_LIBS=BOOL:ON`), because that's where VERSION ends up in the filename in one form or another.

Observation: protobuf produces/installs `libutf8_validity.so`.

Expected to see instead: install as `libutf8_validity.so.29`

acozzette:
@jengelh Would you mind rebasing your branch (or just merging main into it)? I suspect that will fix the failing test runs.

acozzette (approved):

PR #19013, opened 2024-10-28
Update protoc release to include editions language features proto for Go

mkruskal-google:
related to https://github.com/protocolbuffers/protobuf/issues/16211

PiperOrigin-RevId: 684366194

PR #19016, opened 2024-10-28
[29.x] Add missing java load

ted-xie:
(cherry-pick of 89533c743fc7cbec48c1298a043451caada061f1)

For most _native_ loads of Java rules (e.g. java_library), these can be automatically handled by Bazel 8's autoload feature.

However, this is not the case when the main repository (IOW, the repository currently being built) is a rule repo itself. In those cases, module dependencies are not automatically loaded.

This PR fixes a compatibility issue of OSS protobuf Bazel rules with other Bazel rule repositories such as bazelbuild/rules_android.

Without this PR, builds within rules repositories with Bazel 8 will fail like so:

```
external/protobuf+/java/core/BUILD.bazel:142:13: @@protobuf+//java/core:lite: no such attribute 'srcs' in 'java_library' rule
```

PiperOrigin-RevId: 690735533

### DIFF
--- a/java/core/BUILD.bazel
+++ b/java/core/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
+load("@rules_java//java:java_library.bzl", "java_library")
 load("@rules_pkg//pkg:mappings.bzl", "pkg_files", "strip_prefix")
 load("//:protobuf.bzl", "internal_gen_well_known_protos_java")
 load("//:protobuf_version.bzl", "PROTOBUF_JAVA_VERSION")
--- a/java/core/BUILD.bazel
+++ b/java/core/BUILD.bazel
@@ -706,3 +706,4 @@ pkg_files(
     strip_prefix = strip_prefix.from_root(""),
     visibility = ["//java:__pkg__"],
 )
+


zhangskz (approved):

ted-xie:
The failures in presubmit seem unrelated to this PR:

From upb/MacOS:
```
  File "/private/var/tmp/_bazel_runner/88067fd1245450c0f2a417255e46b72b/external/pypi__pip/pip/_internal/resolution/resolvelib/resolver.py", line 92, in resolve
    result = self._result = resolver.resolve(
                            ~~~~~~~~~~~~~~~~^
        collected.requirements, max_rounds=limit_how_complex_resolution_can_be
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
```

From Annotations/All Blocking Tests:
```
Run false
  false
  shell: /usr/bin/bash -e {0}
Error: Process completed with exit code 1.
```

It looks like the first one is some kind of python toolchain error, and the second is a test set up specifically to fail?

ted-xie:
The github workflows are still failing with the numpy error. @zhangskz any idea about the root cause? I still don't think my change should have caused this, and it looks like the recent other commits in the 29.x branch have been green.

I do see that https://github.com/protocolbuffers/protobuf/commit/7060d3d7232096eae2188c61ad6ddfdf87a06f31 bumped the numpy version, and also bumped the MacOS version number from 12 to 13. Would that be a relevant commit to cherry-pick into 29.x?

zhangskz:
Went ahead and merged this -- the error you saw should've been fixed in main and backported in https://github.com/protocolbuffers/protobuf/pull/19013 (due to upgrade of github to python 3.13)

PR #19028, opened 2024-10-29
Fixed parsing bool value from query string

Teclor:
When boolean value comes from query it is represented as 'true' or 'false string. However, it is not parsed until it is not a key for a map field. As a result, error message is thrown when boolean values are passed in GET requests (e.g. when some rest service is mapped to grpc one).
My suggestion: as long as field is defined as boolean, let's accept string 'true' or 'false' values too.

### DIFF
--- a/php/src/Google/Protobuf/Internal/Message.php
+++ b/php/src/Google/Protobuf/Internal/Message.php
@@ -795,8 +795,7 @@ class Message
     private function convertJsonValueToProtoValue(
         $value,
         $field,
-        $ignore_unknown,
-        $is_map_key = false)
+        $ignore_unknown)
     {
         switch ($field->getType()) {
             case GPBType::MESSAGE:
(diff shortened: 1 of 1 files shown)


anandolee:
@Teclor Thank you for this PR, but unfortunately we can't review it until you sign the CLA.

Teclor:
> @Teclor Thank you for this PR, but unfortunately we can't review it until you sign the CLA.

Hello, I've signed CLA the day I made this MR, then I restarted check and now you can see it succeeded 3 days ago.
Here it is:
https://github.com/protocolbuffers/protobuf/pull/19028/checks

Also it looks passed in the check list here in MR
image

Teclor:
@anandolee could you remove cla: no tag from pull request, please? Or do I have to proceed some extra steps?

tonyliaoss:
I don't think we want to accept this. This doesn't conform to the JSON or ProtoJSON spec, or Javascript, for that matter.

PR #19032, opened 2024-10-29
Add a deprecation warning to py_proto_library

mkruskal-google:


protobuf.bzl
```diff
@@ -647,6 +647,7 @@ def py_proto_library(
       This is provided for backwards compatibility only.  Bazel 5.3 will
       introduce support for py_proto_library, which should be used instead.
     """
+    print("The py_proto_library macro is deprecated, switch to the rule defined by rules_python and protobuf (after 29.0).")
```
zhangskz, line 650: "This py_proto_library macro is deprecated and will be removed in the next release. Switch to the official rule defined in protobuf's bazel/py_proto_library.bzl or rules_python instead."

Not sure if the v29.0 is needed if this warning only starts showing up in v29.0 anyways -- but also doesn't really hurt if you think its unclear.

mkruskal-google: o yea, good point lol

zhangskz: can you add the bit about removing in next release and where the official py_proto_library rule is? Might be confusing for users since technically this is also the rule defined by protobuf :P

zhangskz (approved):

Issue #15057, opened 2023-12-12
New Tag: is_sensitive_message


Mahes2:
**What language does this apply to?**
If it's a proto syntax change, is it for proto2 or proto3?
-> proto3

If it's about generated code change, what programming language?
-> I think it's useful for every programming language, but in my case it's Golang.

**Describe the problem you are trying to solve.**
I have a project which has grpc interceptor that log every request and response. Meanwhile in current changes, i added a new sensitive field which i do not want to print that field into the log. I do some research and found that proto3 has a tag called `json_name`. But unfortunately it only generates the proto3 json tag that can only be used by jsonpb / protojson. Also it's still not working as json: `-,omitempty` , it's still printing the field as `"-"`.
Example:
```
proto file:
message Response {
    string phone_number= 1 [json_name="-"];
}

pb.go file:
type Response struct {
     PhoneNumber string `protobuf:"bytes,1,opt,name=phone_number,json=-,proto3" json:"phone_number,omitempty"`
}

json:
{
    "-": "3463625"
}
```

**Describe the solution you'd like**
I suggest we add new tag, which is `is_sensitive_message`, and by reading that tag in message encoder like protojson we can skip the message to hide sensitive data.
```
proto file:
message Response {
    string phone_number= 1 [is_sensitive_message=true];
}

pb.go file:
type Response struct {
     PhoneNumber string `protobuf:"bytes,1,opt,name=phone_number,json=phone_number,is_sensitive_message=true,proto3" json:"phone_number,omitempty"`
}

json:
{}
```

**Describe alternatives you've considered**
-

**Additional context**
Currently i'm in progress updating `protocolbuffers/protobuf` to add the tag, as well as the protojson in `golang/protobuf` in my local. If this feature is approved, then i will submit PR ASAP.

Issue #5577, opened 2019-01-14
Ability to hide some sensitive fields in JsonFormat.Printer


wangjunwei87:
**What language does this apply to?**

proto3 with java

**Describe the problem you are trying to solve.**

Some messages include sensitive data, eg: LoginRequest may contains a password.

**Describe the solution you'd like**

```
{
"password": "****"
}
```
instead of 

```
{
"password": "actual use input"
}
```

haon4:
Today, JsonFormat.parser().parse(JsonFormat.printer().print(message)) would return an equivalent proto. However, this feature request would cause the parse() to return a different proto. I don't think we would want to introduce such behavior into JsonFormat. Furthermore, obfuscation logics is application specific. Some may want to completely obfuscate a field, while others may want to show some hints.

I think it is better to do this at the application level. Here is what you can do:

1. Define a proto field option like this:
```
message ObfuscationMode {
  bool obfuscate = 1;
}

extend proto2.FieldOptions {
  ObfuscationMode obfuscation_mode = 123456789;
}
```

2. Use proto reflection to determine to obfuscate a field based on the option:
```
  public void obfuscate(Builder builder) {
    for (FieldDescriptor field : builder.getAllFields().keySet()) {
      if (field.isRepeated()) {
        if (field.getJavaType() == JavaType.MESSAGE) {
          for (int i = 0; i < builder.getRepeatedFieldCount(field); ++i) {
            obfuscate(builder.getRepeatedFieldBuilder(field, i));
          }
        } else if (shouldObfuscate(field)) {
          for (int i = 0; i < builder.getRepeatedFieldCount(field); ++i) {
            // Actually obfuscating the field.
            builder.setRepeatedField(field, i, "****");
          }
        }
      } else if (builder.hasField(field)) {
        if (field.getJavaType() == JavaType.MESSAGE) {
          obfuscate(builder.getFieldBuilder(field));
        } else if (shouldObfuscate(field)) {
          // Actually obfuscating the field.
          builder.setField(field, "****");
        }
      }
    }
  }

  private static boolean shouldObfuscate(FieldDescriptor field) {
    // Determine whether to obfuscate a field. Change this as needed.
    if (field.getJavaType() != JavaType.STRING) {
      return false;
    }
    ObfuscationMode obfuscationMode =
        field.getOptions().getExtension(ObfuscationModeOuterClass.obfuscationMode);
    if (obfuscationMode == null) {
      return false;
    }
[...]

Issue #1160, opened 2016-01-20
Support data redaction, so that sensitive data in protos doesn't leak into places like log files


jhump:
In every server language, it is trivial for someone to include a proto in an unstructured log line:
`logger.info("something happened: %s", proto);`

It is very common for such log output to be ingested into various systems (like exception tracking / alerting, full-text indexing for search) which suddenly makes it possible to accidentally leak sensitive or secret data into these systems (PII, account numbers and PANs, passwords, etc). Scraping all such sensitive data after it is accidentally leaked can be a huge pain, depending on the tools used. And identifying what should be redacted in the path of ingestion is error-prone. Pattern marching could identify things that look like social security numbers and credit card numbers, but phone numbers can get tricky (especially if supporting international phone number formats) and addresses moreso. Passwords? Forget about it...

So having a way to redact fields from operations that convert messages into strings (`Object#toString` in Java, for example) is extremely useful to prevent accidental logging of sensitive data. Even if this functionality doesn't belong in the core protobuf runtime libraries, I think there should at least be a hook point in the library so that redaction can be "plugged in" for a given deployment.

This problem is simpler when all logs are structured (e.g. not unstructured text output, log processors can then use custom message and field options to filter/modify the logs to prevent leaking sensitive data). But even when using structured logging for some things, there is likely always a need for unstructured log output from a server, even if just for debugging.

---

Strawman solution (using Java runtime as an example, but pattern can be applied to other runtimes/languages):
- Add `redacted` fields to `FieldOptions` and `MessageOptions` in `descriptor.proto`.
- `TextFormat.Parser.Builder` gets a new setting: `setRedacted(boolean)` (defaults to false, existing behavior). When true, redacted messages result in `"{ <redacted> }"` and redacted fields, if present, are shown as `fieldname: <redacted>`.
- `AbstractMessage#toString()` updated to use redaction feature of `TextFormat`.

If adding `redacted` options into `descriptor.proto` is going too far, then `TextFormat.Parser.Builder#setRedacted` could instead take a `GeneratedExtension<FieldOptions, Boolean>` -- a custom field option that is queried to determine if field values should be skipped. This would require a way to configure a default option that can then be used from `AbstractMessage#toString`.

jhump:
@lukaszx0 @cconroy @zellyn

toddlipcon:
[...]

PR #19084, opened 2024-11-01
Extend JsonFormatter with an optional ShouldSerializeField delegate function to determine when a field is serialized

juandemanjon:
JsonFormatter is very handy to serialize the request or response payload in logs but it serializes the whole message including sensitive fields.

This PR adds an optional ```ShouldSerializeField``` delegate function to the Settings class to indicate when it should be done

For example
```csharp
static String RemoveSensitiveData(IMessage message)
{
    JsonFormatter.Settings.ShouldSerializeFieldDelegate shouldSerializeField =
        (IMessage message, FieldDescriptor field, object value) => field.Name != "password";

    var formatter = new JsonFormatter(JsonFormatter.Settings.Default.WithShouldSerializeField(shouldSerializeField));

    return formatter.Format(message);
}
```

### DIFF
--- a/csharp/src/Google.Protobuf.Test/JsonFormatterSettingsTest.cs
+++ b/csharp/src/Google.Protobuf.Test/JsonFormatterSettingsTest.cs
@@ -18,12 +18,22 @@ namespace Google.Protobuf
 {
     public class JsonFormatterSettingsTest
     {
+        static JsonFormatter.Settings.ShouldSerializeFieldDelegate shouldSerializeField = (IMessage message, FieldDescriptor field, object value) => true;
+
         [Test]
         public void WithIndentation()
         {
             var settings = JsonFormatter.Settings.Default.WithIndentation("\t");
             Assert.AreEqual("\t", settings.Indentation);
         }
+
+        [Test]
+        public void WithFormatField()
+        {
+            var settings = JsonFormatter.Settings.Default.WithShouldSerializeField(shouldSerializeField);
+            Assert.AreEqual(shouldSerializeField, settings.ShouldSerializeField);
+        }
+
 
         [Test]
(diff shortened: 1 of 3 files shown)


jskeet:
Is this functionality which exists in a similar form in other languages? I'd rather avoid adding new features which are *generally* applicable, only to C#.

juandemanjon:
Related issues:
* #15057 
* #5577 
* #1160

jskeet:
Right, it's good to see those other bugs - but I don't see any sign that there's been an effort to address them in this way. We really don't want multiple completely different ways of addressing the same issue.

@mkruskal-google Do you know if JSON redaction is a planned feature across the board?

anandolee:
@juandemanjon you may also need to sign the CLA to for forward the contribution

juandemanjon:
> @juandemanjon you may also need to sign the CLA to for forward the contribution

Done!

mkruskal-google:
I'm not aware of any JSON redaction planned or in place.  We have plans to implement redaction of sensitive fields in debug strings, which we'll be releasing for C++ in 6.30.0.  But that's limited to DebugString, and text-format still won't redact

acozzette:
I appreciate the pull request but unfortunately I don't think we want to merge this one, because we don't want C# to have a redaction feature that's inconsistent with the way it's done in other languages. I think the general principle is that when we're serializing to a wire format we want to serialize everything, and we only want to redact data when we're just stringifying a proto for a developer to look at directly.

If you would like to redact fields from JSON format then I think the best way to do it would be to just clear fields from the proto that you don't want to serialize, and then serialize the whole proto. You could use reflection to apply arbitrary criteria to decide whether to remove a field.

PR #19115, opened 2024-11-02
Performance optimization extension for `RepeatedField<T>`

prezaei:
.NET 8 added support for `AddRange` using the `ReadOnlySpan<T>` data structure. This can significantly improve the performance of `RepeatedField<T>` when adding multiple elements and can avoid the potential need to create extra arrays when using spans.

Here is the code that was added to .NET to support `AddRange` on `List<T>`: https://github.com/dotnet/runtime/blob/85f6d2d9449fb438cd8a5ca41d4beb4c657594e9/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/CollectionExtensions.cs#L85-L103

In the current implementation of `RepeatedField<T>`, the `AddRange` method only operates on `IEnumerable<T>`. Unfortunately, this means that the callers using `Span<T>` have to copy the content of the span to an array or to some other `IEnumerbale<T>` before calling `AddRange`. This is an expensive operation and can be avoided by adding support for `AddRange` that takes a `ReadOnlySpan<T>` directly.

In this proposal, `AddRange` is added as an extension method instead of directly adding it to `RepeatedField<T>` because we want to avoid ambiguity between `RepeatedField<T>.AddRange(IEnumerbale<T>)` and `RepeatedField<T>.AddRange(Span<T>)` for types such as `T[]`. This is the same exact approach taken by .NET runtime itself.


csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// A set of extension methods on RepeatedField
+    /// </summary>
+    public static class RepeatedFieldExtensions
+    {
+        /// <summary>Adds the elements of the specified span to the end of the <see cref="RepeatedField{T}"/>.</summary>
+        /// <typeparam name="T">The type of elements in the <see cref="RepeatedField{T}"/>.</typeparam>
+        /// <param name="repeatedField">The list to which the elements should be added.</param>
+        /// <param name="source">The span whose elements should be added to the end of the <see cref="RepeatedField{T}"/>.</param>
+        /// <exception cref="ArgumentNullException">The <paramref name="repeatedField"/> is null.</exception>
+        public static void AddRange<T>(this RepeatedField<T> repeatedField, ReadOnlySpan<T> source)
```
JamesNK, line 24: Change this to an overload on RepeatedField.

.NET added `AddRange` as an extension method to `List<T>` to avoid AOT file growth. RepeatedField has much narrower usage and that isn't a concern.

Using an overload also eliminates need to make some fields internal on RepeatedField.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (default(T) == null)
+            {
+                foreach (var value in source)
+                {
+                    if (value == null)
+                    {
+                        throw new ArgumentException("ReadOnlySpan contained null element", nameof(source));
+                    }
+                }
+            }
```
JamesNK, line 46: I'm sure there must be a faster way to check if there is a null in a span.

Does this work? IndexOf must have SIMD optimizations to quickly search the span.
```suggestion
            if (default(T) == null)
            {
                if (source.IndexOf(null) != -1)
                {
                    throw new ArgumentException("ReadOnlySpan contained null element", nameof(source));
                }
            }
```

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
prezaei: AddRange is added as an extension method instead of directly adding it to RepeatedField<T> because we want to avoid ambiguity between RepeatedField<T>.AddRange(IEnumerbale<T>) and RepeatedField<T>.AddRange(Span<T>) for types such as T[]. This is the same exact approach taken by .NET runtime itself.
I agree that AOT is not a concern here.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
prezaei: The latest code uses a for loop:

```csharp
for (int i = 0; i < source.Length; i++)
                {
                    if (source[i] == null)
                    {
                        throw new ArgumentException("ReadOnlySpan contained null element", nameof(source));
                    }
                }
```

prezaei: More context:

source here is a ReadOnlySpan, so just foreach'ing it will be lowered by the C# compiler into the equivalent for loop.
 
IndexOf won't work here as the available overloads for IndexOf on a span require the T to be IEquatable<T>, and in here T isn't constrained.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
```diff
@@ -34,8 +34,8 @@ public sealed class RepeatedField<T> : IList<T>, IList, IDeepCloneable<RepeatedF
         private static readonly T[] EmptyArray = new T[0];
         private const int MinArraySize = 8;
 
-        private T[] array = EmptyArray;
-        private int count = 0;
+        internal T[] array = EmptyArray;
```
jskeet, line 37: Please keep all fields private. I have an alternative suggestion that doesn't require exposing any existing members.

I can see that we need a new type in order to "house" the extension method - but that can just call a new internal instance method on RepeatedField. That's more in the spirit of "really, we'd like it to be a normal instance method, but we don't want overload resolution issues." Keep all the actual *implementation* in RepeatedField, with an internal method of `AddSpan` or similar (it could even be a public method, potentially) and then just make the extension method delegate to that.

JamesNK: Yeah, I agree. An internal `RepeatedField.AddRange(ReadOnlySpan<T>)` (or a different name if AddRange causes problems) extension method calls keeps things encapsulated.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+#endregion
+
+using System;
+
+namespace Google.Protobuf.Collections
+{
+    /// <summary>
+    /// A set of extension methods on RepeatedField
```
JamesNK, line 15: nit
```suggestion
    /// A set of extension methods on <see cref="RepeatedField{T}" />.
```
I don't remember if this works or not when there isn't the generic type in the context.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+// Copyright 2015 Google Inc.  All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+#endregion
+
+using System;
+
+namespace Google.Protobuf.Collections
```
JamesNK, line 12: Should the namespace just be `Google.Protobuf`? It's rare that people will directly reference `Google.Protobuf.Collections` in their app because they're just modifying collections created by generated code.

Having the extension method in a rarely used namespace makes it hard to discover/use.

csharp/src/Google.Protobuf/Collections/RepeatedField.cs
prezaei: Addressed.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
prezaei: Makes sense to me.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>Adds the elements of the specified span to the end of the <see cref="RepeatedField{T}"/>.</summary>
+        /// <typeparam name="T">The type of elements in the <see cref="RepeatedField{T}"/>.</typeparam>
+        /// <param name="repeatedField">The list to which the elements should be added.</param>
+        /// <param name="source">The span whose elements should be added to the end of the <see cref="RepeatedField{T}"/>.</param>
+        /// <exception cref="ArgumentNullException">The <paramref name="repeatedField"/> is null.</exception>
+        public static void AddRange<T>(this RepeatedField<T> repeatedField, ReadOnlySpan<T> source)
+        {
+            ProtoPreconditions.CheckNotNull(repeatedField, nameof(repeatedField));
+            repeatedField.AddRange(source);
```
JamesNK, line 28:
```suggestion
            repeatedField.AddRangeSpan(source);
```

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using Google.Protobuf.Collections;
+
+namespace Google.Protobuf
+{
+    /// <summary>
+    /// A set of extension methods on <see cref="RepeatedField{T}"/>
+    /// </summary>
+    public static class RepeatedFieldExtensions
+    {
+        /// <summary>Adds the elements of the specified span to the end of the <see cref="RepeatedField{T}"/>.</summary>
```
JamesNK, line 20:
```suggestion
        // Note: This method is an extension method to avoid ambigious overload conflict when the source is an array.
        /// <summary>Adds the elements of the specified span to the end of the <see cref="RepeatedField{T}"/>.</summary>
```

prezaei:
@jskeet , I am new to the process here to get PRs approved. What is next here? Thank you for the help.

jskeet:
I will have another look when I get a chance. (protobuf isn't my main project)

jskeet (commented):
Just one final change, then I think we'll be good to go.

csharp/src/Google.Protobuf/Collections/RepeatedFieldExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+#endregion
+
+using System;
+using Google.Protobuf.Collections;
+
+namespace Google.Protobuf
```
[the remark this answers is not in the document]
prezaei: Thank you and done, @jskeet .

jskeet:
@mkruskal-google I believe this is now ready to merge.

mkruskal-google:
Looks like C# tests are failing on windows

jskeet:
Will investigate when I get a chance, but I'm away next week.

prezaei:
> Will investigate when I get a chance, but I'm away next week.

Thank you, @jskeet . Looks like some legacy Patterns and Practices rule being used for the old .NET Framework that is no longer relevant. Please let me know if you'd like to disable this rule.

jskeet:
I doubt that we want to disable the rule. It looks like we *may* want to add `[AllowPartiallyTrustedCallers]` to the extension method. I'll try to look later today or later this week - I've just got back from a week of travelling, so my inbox is currently huge, but I'll see what I can do.

JamesNK:
Add `[SecuritySafeCritical]` to `RepeatedFieldExtensions.AddRange` and `RepeatedField.AddRangeSpan`.

prezaei:
> I doubt that we want to disable the rule. It looks like we _may_ want to add `[AllowPartiallyTrustedCallers]` to the extension method. I'll try to look later today or later this week - I've just got back from a week of travelling, so my inbox is currently huge, but I'll see what I can do.

All tests except for `Rust` seem to be passing now. Can you please do one final check and let me know what the next step here is? Thanks.

jskeet:
Looks fine to me - and I'd be surprised if the Rust failures were actually relevant here.

@mkruskal-google Please could you kick off the internal process for this change?

prezaei:
> Looks fine to me - and I'd be surprised if the Rust failures were actually relevant here.
> 
> @mkruskal-google Please could you kick off the internal process for this change?

Thank you, @jskeet  and @mkruskal-google . Looking forward to the next steps here. Cheers

PR #19152, opened 2024-11-06
Add `preserve_descriptor_field_order` option to PrintOptions

same-id:
This is useful to control the order of the fields in output json.

Default false - so does not affect current users of the library.

### DIFF
--- a/src/google/protobuf/json/internal/unparser.cc
+++ b/src/google/protobuf/json/internal/unparser.cc
@@ -469,10 +469,12 @@ absl::Status WriteFields(JsonWriter& writer, const Msg<Traits>& msg,
   // Add extensions *before* sorting.
   Traits::FindAndAppendExtensions(msg, fields);
 
-  // Fields are guaranteed to be serialized in field number order.
-  absl::c_sort(fields, [](const auto& a, const auto& b) {
-    return Traits::FieldNumber(a) < Traits::FieldNumber(b);
-  });
+  if (!writer.options().preserve_descriptor_field_order) {
+    // Serialize in field number order.
+    absl::c_sort(fields, [](const auto& a, const auto& b) {
+      return Traits::FieldNumber(a) < Traits::FieldNumber(b);
+    });
+  }
 
   for (auto field : fields) {
     RETURN_IF_ERROR(WriteField<Traits>(writer, msg, field, first));
--- a/src/google/protobuf/json/internal/writer.h
+++ b/src/google/protobuf/json/internal/writer.h
@@ -56,6 +56,10 @@ struct WriterOptions {
   // in the unit tests; we intend to remove this setting eventually. See
   // b/234868512.
   bool allow_legacy_syntax = false;
+  // If set, the fields are ordered by their index in the descriptor.
+  // Otherwise, the fields are ordered by their tag number.
+  // This is useful to control the order of fields in the output.
+  bool preserve_descriptor_field_order = false;
 };
 
 template <typename Tuple, typename F, size_t... i>
--- a/src/google/protobuf/json/json.cc
+++ b/src/google/protobuf/json/json.cc
@@ -36,6 +36,7 @@ absl::Status BinaryToJsonStream(google::protobuf::util::TypeResolver* resolver,
   opts.always_print_fields_with_no_presence =
       options.always_print_fields_with_no_presence;
   opts.unquote_int64_if_possible = options.unquote_int64_if_possible;
+  opts.preserve_descriptor_field_order = options.preserve_descriptor_field_order;
 
   // TODO: Drop this setting.
   opts.allow_legacy_syntax = true;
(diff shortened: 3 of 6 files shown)


sirlupinwatson:
Hi there! I think this is a good step forward, and those output could also be even better in the future with improvement and space as each element, class, etc, should be easy to read, but not easy to get and decode... There is room for much improvements still. 

Great idea!

esrauchg:
Unfortunately we've decided not to accept this change back to the protobuf library.

The code change itself is good, but we have a strong aversion to adding more options to the JSON parser/serializers for more corner cases; this is actually the V2 implementation of the C++ JSON library and the V1 had 50+ problematic options on it. Its somewhat current pain point that the different languages have protobuf implementations each have an uneven set of different one-off options (which makes it harder to have logic/code be portable between languages).

It's true that the ProtoJSON format is intended to be human-readable compared to binary format, but really its still primarily intended as a machine-consumed format (either parsed back by a protobuf impl, or simply JSON.parse()'d for some hand-written logic), and the JSON spec itself doesn't ascribe any meaning to the ordering of keys.

Sorry about that!

same-id:
Hi @esrauchg,

Thank you for the detailed answer.
I understand why we don't like more options, but do not agree regarding the necessity of the ordering change

My reasoning is the following:

* If ordering doesn't matter, why do we have an explicit `sort` in the C json serializer? Seems like it is just wasteful computation.

Are we fine with removing it? This way we get what this PR is trying to achieve without adding another option.

My guess is that the answer will be "no" for backwards compatibility with existing consumers of this library (not to mention the tests).
(Probably would have worked in Go's protobuf implementation, where the output is a bit randomized anyway - spacing wise, not order wise)

It is true that the JSON spec does not talk about order of keys, but we do see the following in the broader ecosystem:
1. Javascript - New ECMA standard defines that order is preserved on JSON keys (https://262.ecma-international.org/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys) at least for non numeric keys `JSON.stringify({options: {"key0":"0", "key4":"4", "key2":"2", "3": "3", "2": "2"}})`
2. Python's `json.dump` since version 3.x default encoder preserves the order of the python fields
3. Even protobuf's Go implementation, uses descriptor order.

So languages do try and make an effort to preserve order when possible (even if previously they did not)

I would argue that this should have been the default behavior and no option is necessary - however it conflicts with the more important rule of `Changes to existing APIs are almost never accepted. Stability and backwards compatibility are paramount. In the unlikely event a breaking change is required, it must usually be implemented in google3 first`.

WDYT?

esrauchg:
> If ordering doesn't matter, why do we have an explicit sort in the C json serializer?
I think the situation is more that:
- The APIs intended to commit to no specific ordering
- When they were written, they accidentally had some arbitrary, deterministic and stable ordering 
- Once you have an enormous number of usages, too many people rely on the incidental behavior as it happens to exist, rather than the claimed behaviors, even if you loudly officially claim "the order is arbitrary and not stable", you don't really have the freedom to do so in the implementation (see: https://www.hyrumslaw.com/ ).

> Seems like it is just wasteful computation.

I checked the change history of this specific sort, and what happened is a prior version of this library _accidentally_ sorted the fields in field number order, then when a new version was rewritten that explicit sort was added to make it possible to migrate usages of the old API that had tests which implicitly entrenched to that order even though it shouldn't really be a reliable invariant.

> It is true that the JSON spec does not talk about order of keys, but we do see the following in the broader ecosystem:

The examples are interesting (I was unaware) but I think 1+2 are a bit different than this case; they seem to be about _preserving_ the order given to you, so eg if you iterate over it you'll see the same order and if you round trip it you may be more stable. Those properties won't be true here regardless; iteration and round-trip will still reorder the fields regardless here.

The Go Protobuf example is also relevant in the sense that I suspect that's an example of a different accidentally stable order and not intended to be reliable behavior of the API; I don't see order mentioned in the Go Protobuf JSON documentatio, instead it explicitly says "Do not depend on the output being stable." (and it actually slightly randomizes the JSON output, since active randomization of subtle properties you aren't committed to maintaining forever is one of the few ways to avoid Hyrum's Law effects).

> I would argue that this should have been the default behavior and no option is necessary

I think there's a valid argument there, but exactly for the same reason that sort was added to begin with its very costly to change now (anyone who wrote a simple unit test that has a hardcoded JSON string to compare to could break by the change, which is ... a lot of people).

I think despite your good points, it's just too costly to change the current behavior and we're just really that adverse to having more toggle options on the serializers that we unfortunately don't want to take an option on the officially supported API even though the behavior that toggle would enable is sensible.

We definitely can keep an eye on this and revisit if there are clear business need usecases to justify overcoming either the cost of changing the default behavior or the cost of having one more toggle option.

I appreciate the thoughtful discussion and sorry that our priorities for the API don't quite line up here.

same-id:
So I would mainly say we agree on all points here beside:

"Is it really THAT necessary to justify yet-another-toggle"

I'm fine with the verdict, although I was excited to see this change (for our own API's sake, when using envoy) - let's see if it will be worth revisiting down the line.

PR #19187, opened 2024-11-07
Begin adding extension numbers to SourceCodeInfo and FileDescriptorSet

googleberg:
for tooling purposes.

Allocate extension number 536000000 to Buf

Extension numbers will be allocated one at a time by request. Once extension declarations are open-sourced for extensions in descriptor.proto, we will move these declarations into a separate config file to declutter descriptor.proto.

PiperOrigin-RevId: 692315472


googleberg:
This is the more correct version of the CP/PR that I tried to make earlier.

mkruskal-google (approved):

PR #19232, opened 2024-11-12
Building python binding with bazel on s390x

gongsu832:
Changes needed for building python binding with bazel on s390x.

### DIFF
--- a/python/dist/BUILD.bazel
+++ b/python/dist/BUILD.bazel
@@ -479,7 +479,6 @@ py_dist(
         "win64": "310",
         "linux-x86_64": "38",
         "linux-aarch_64": "38",
-        "linux-s390x": "38",
         "osx-universal2": "38",
     },
     # LINT.ThenChange(:python_tag)
--- a/python/dist/BUILD.bazel
+++ b/python/dist/BUILD.bazel
@@ -85,6 +85,25 @@ config_setting(
 )
 
 config_setting(
+    name = "linux_s390x_release",
+    flag_values = {
+        "//toolchain:release": "True",
+    },
+    values = {"cpu": "linux-s390x"},
+)
+
+config_setting(
+    name = "linux_s390x_local",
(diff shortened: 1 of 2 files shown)


zhangskz:
We don't currently release or have plans to release python binary wheels for s390x. Is this a request for us to add s390x to our release, or is the purpose of this PR to add these config_settings to enable building for s390x yourself?

gongsu832:
Hi @zhangskz this is just so that we can build on s390x ourselves. Because protobuf is a dependency for one of our open source project [onnx-mlir](https://github.com/onnx/onnx-mlir). Thanks.

acozzette (approved):

python/dist/BUILD.bazel
```diff
@@ -458,6 +479,7 @@ py_dist(
         "win64": "310",
         "linux-x86_64": "38",
         "linux-aarch_64": "38",
+        "linux-s390x": "38",
```
acozzette, line 482: @gongsu832 Unfortunately I haven't been able to figure out any way to merge this change without breaking our CI, unless I delete this line. Otherwise our build fails because we don't have an appropriate toolchain installed for s390x. Is is still useful to merge the PR without this line? I guess that might still allow you to write your own `py_dist()` target in a separate repo.

gongsu832: Hi @acozzette thank you very much for trying to merge this PR. Yes it would still be very helpful for us if you could merge without that line.

acozzette: Sounds good, I submitted the change as 15e7d71eacd7df66531aae2bb7c9b8b8a1a81483. I made a slight tweak to add an `_unused` suffix on the `config_setting` targets just as a reminder to our team that we are not using them ourselves.

gongsu832: Thank you @acozzette for your help. Very much appreicated.

PR #19243, opened 2024-11-13
Fix an UBSAN error

zhihongliuus:
external/com_google_protobuf/src/google/protobuf/stubs/stringpiece.h:362:23: runtime error: null pointer passed as argument 1, which is declared to never be null
external/com_google_protobuf/src/google/protobuf/stubs/stringpiece.h:362:23: runtime error: null pointer passed as argument 2, which is declared to never be null


acozzette:
Sorry but unfortunately we can't review this pull request until you sign the CLA.

Also I suspect the fix might not address the error you ran into, because we deleted our StringPiece code two years ago. You might find that the problem goes away on its own if you upgrade to a more recent protobuf version.

acozzette:
I'm going to go ahead and close this, but feel free to open a new PR if you're OK with signing the CLA.

PR #19265, opened 2024-11-14
Make μpb descriptor library targets public.

phst:
Otherwise there’s no public target for descriptor.proto after commit eab1fa2765cc5fe21128fe92f1acbb6c2f8f73c2.


phst:
Or is the idea that users should write their own upb_c_proto_library targets for this?

haberman:
Since the upb API is not generally public, and rules_elisp's use of it is by special arrangement, I'd prefer not to make these public. We could potentially add rules_elisp to an allowlist here.

PR #19266, opened 2024-11-14
Refer to upb_proto_c_library rather than upb_proto_library.

phst:
The former seems to be preferred nowadays.


haberman (approved):

haberman:
I'm sorry, I read this too quickly. This should refer to `upb_minitable_proto_library()`, which generates MiniTables. The `upb_c_proto_library()` target generates a C generated API, and is layered on top of the MiniTables.

I'll update the CL accordingly.

PR #19292, opened 2024-11-18
fix(example): c++14 limitation

Kaikaikaifang:
fix the bug: 

```shell
‘std::string_view’ is only available from C++17 onwards
```

The error occurs because the code is being compiled using the C++14 standard, which does not include the std::string_view type introduced in C++17. The Abseil library (absl), used by Protocol Buffers, includes absl/strings/string_view.h, which uses std::string_view. This causes the compilation to fail in environments configured with C++14 or earlier.

In the provided build system, the file add_person.cc is being compiled using the -std=c++14 flag, which is incompatible with code depending on features from C++17 (e.g., std::string_view).

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -46,11 +46,11 @@ protoc_middleman_ruby: addressbook.proto
 
 add_person_cpp: add_person.cc protoc_middleman
 	pkg-config --cflags protobuf  # fails if protobuf is not installed
-	c++ -std=c++14 add_person.cc addressbook.pb.cc -o add_person_cpp `pkg-config --cflags --libs protobuf`
+	c++ add_person.cc addressbook.pb.cc -o add_person_cpp `pkg-config --cflags --libs protobuf`
 
 list_people_cpp: list_people.cc protoc_middleman
 	pkg-config --cflags protobuf  # fails if protobuf is not installed
-	c++ -std=c++14 list_people.cc addressbook.pb.cc -o list_people_cpp `pkg-config --cflags --libs protobuf`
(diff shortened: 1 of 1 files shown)


tonyliaoss:
Hello,

Unfortunately we won't be able to proceed until the CLA has been signed.

Also, while we have migrated to C++17 since https://github.com/protocolbuffers/protobuf/commit/fe535930d3183b46d7e088683cbe8c49285715f8, the current PR description doesn't make much sense. `examples/add_person.cc` and `examples/list_people.cc` don't use `std::string_view` as you described.

Kaikaikaifang:
> Hello,
> 
> Unfortunately we won't be able to proceed until the CLA has been signed.
> 
> Also, while we have migrated to C++17 since [fe53593](https://github.com/protocolbuffers/protobuf/commit/fe535930d3183b46d7e088683cbe8c49285715f8), the current PR description doesn't make much sense. `examples/add_person.cc` and `examples/list_people.cc` don't use `std::string_view` as you described.

Hello,

Thank you for pointing that out. I’ve updated the PR description to clarify the issue. The build problem arises from dependencies (absl) requiring C++17 features such as std::string_view, even though examples/add_person.cc and examples/list_people.cc do not directly use std::string_view.

Meanwhile, the CLA has signed. This pr's tag has been changed to cla:yes, but the other #19293  pr's tag is still cla:no even though the cla checksum passes, and the cla bot didn't remove the tag and add cla:yes.

tonyliaoss:
SGTM. Thanks Kaikai! I'll kick of a round of CI tests then approve if tests look good.

PR #19293, opened 2024-11-18
fix(example): go version is too low

Kaikaikaifang:
[bug] tutorialpb/addressbook.pb.go:317:40: predeclared any requires go1.18 or later (-lang was set to go1.14; check go.mod)


tonyliaoss (approved):

tonyliaoss (approved):
Approving again after Copybara fixes.

Issue #19408, opened 2024-11-26
Compiling from source fails on windows due to bash requirement.


freeform-andre:
I'm trying to upgrade protobuf and ran into a new build error when building with Bazel I believe is caused by https://github.com/protocolbuffers/protobuf/commit/898d8fa1fd6ff07149b283b3bbff00a6a3d8c52b introducing an action which uses [ctx.actions.run_shell](https://bazel.build/rules/lib/builtins/actions#run_shell). This will fail on systems that don't bash setup and appears to fail after `v25.3` (which is what I'm upgrading from).

**What version of protobuf and what language are you using?**
Upgrading from 25.3 to 28.3.

Language: C++

**What operating system (Linux, Windows, ...) and version?**
Windows 10

**What runtime / compiler are you using (e.g., python version or gcc version)**
MSVC

**What did you do?**
1. Setup a Windows machine without Bash installed or WSL configured.
2. Install Bazel
3. Clone the Protobuf repo
4. run `bazel build //editions:embed_test_defaults`

**What did you expect to see**

A successful build

**What did you see instead?**

The specific error may vary but for me I hit this compliant about WSL when Bazel tries to invoke Bash.
```
ERROR: C:/users/user/code/com_google_protobuf/editions/BUILD:58:23: Action editions/defaults_test_embedded_base64.h failed: (Exit 1): bash.exe failed: error executing Action command (from target //editions:embed_test_defaults_base64) C:\Windows\system32\bash.exe -c ... (remaining 1 argument skipped)
Windows Subsystem for Linux has no installed distributions.

Use 'wsl.exe --list --online' to list available distributions
and 'wsl.exe --install <Distro>' to install.

Distributions can also be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
Error code: Bash/Service/CreateInstance/GetDefaultDistro/WSL_E_DEFAULT_DISTRO_NOT_FOUND
Target //editions:embed_test_defaults_base64 failed to build
Use --verbose_failures to see the command lines of failed build steps. 
```

**Anything else we should know about your project / environment**

PR #19409, opened 2024-11-26
Avoid unnecessary use of run_shell

freeform-andre:
Replaces the use of `ctx.actions.run_shell` with `ctx.actions.run` by rewriting the shell script logic in the existing C++ tool being used in `run_shell`.

closes https://github.com/protocolbuffers/protobuf/issues/19408


mkruskal-google (approved):

mkruskal-google:
Thanks for cleaning this up!

Issue #19430, opened 2024-11-27
Python release for 5.29.0 has indent problem


thatch:
If you run `python -m compileall` on any of the wheels for the latest release ~40 minutes ago, it will fail with:

```
Compiling 'google/protobuf/service.py'...
*** Sorry: IndentationError: unindent does not match any outer indentation level (service.py, line 78)
```

I can confirm there appears to be a missing space, but I don't see that file (under that name or contents) checked in.

piotlinski:
Confirming that, it seems to be visible on the 29.x branch:
https://github.com/protocolbuffers/protobuf/blob/e0a3978b678be0f3aab8689770b64d1465d2479e/python/google/protobuf/service.py#L78

ejona86:
That file [has been deleted on `main`](https://github.com/protocolbuffers/protobuf/commit/5ba74b11e8d2bd5e9b22e972beb572668bf6191c), but [it is in the 29.x branch](https://github.com/protocolbuffers/protobuf/blob/29.x/python/google/protobuf/service.py).

SebKleiner:
Hi, I still face this issue while installing 5.29.0

miguelgfierro:
Confirming as well, it's breaking all our tests: https://github.com/recommenders-team/recommenders/issues/2191

**Crazy idea: it world be nice to test your library before making a release.**

miguelgfierro:
@anandolee @ejona86 you need to yank 5.29.0 https://pypi.org/project/protobuf/5.29.0/

dionhaefner:
lmao @ the fact that this is still live 5 days later

```bash
$ pip install protobuf
$ python -c "import google.protobuf.service"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/dion/.virtualenvs/tempenv-2d0a294330027/lib/python3.10/site-packages/google/protobuf/service.py", line 78
    raise NotImplementedError
                             ^
IndentationError: unindent does not match any outer indentation level
```

zhangskz:
This should be patched in v29.1 ~tomorrow. We are not planning on yanking v29.0 currently, since the issue is isolated to a long-deprecated API slated for removal. For now, we recommend not picking up v29.0 yet if you need `compileall` or if using the `service.py` (or ideally moving off of that since it'll be removed soon in v30.x).

[...]

PR #19434, opened 2024-11-27
Revert "Remove deprecated service.py usages from test". For 29.x only

anandolee:
This reverts commit 471ff7727031d7ca17c62f34c38314758142393e.

Add the service test back and fix indent issue only for 29.x for https://github.com/protocolbuffers/protobuf/issues/19430

### DIFF
--- a/python/google/protobuf/service.py
+++ b/python/google/protobuf/service.py
@@ -70,12 +70,12 @@ class Service(object):
     Postconditions:
 
     * "done" will be called when the method is complete.  This may be
-     before CallMethod() returns or it may be at some point in the future.
+      before CallMethod() returns or it may be at some point in the future.
     * If the RPC failed, the response value passed to "done" will be None.
-     Further details about the failure can be found by querying the
-     RpcController.
+      Further details about the failure can be found by querying the
+      RpcController.
     """
-   raise NotImplementedError
+    raise NotImplementedError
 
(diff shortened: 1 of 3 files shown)


acozzette (approved):

shaod2 (commented):
We should probably make this change in Piper and let copybara do the job

mkruskal-google:
This file is deleted in piper :P

mkruskal-google (approved):

zhangskz:
> This file is deleted in piper :P

Clarification: This is deleted in main (for 30.x), but we're rolling this back for just the 29.x branch.

python/google/protobuf/service.py
```diff
@@ -70,12 +70,12 @@ def CallMethod(self, method_descriptor, rpc_controller,
...
-     before CallMethod() returns or it may be at some point in the future.
+      before CallMethod() returns or it may be at some point in the future.
     * If the RPC failed, the response value passed to "done" will be None.
-     Further details about the failure can be found by querying the
-     RpcController.
+      Further details about the failure can be found by querying the
+      RpcController.
     """
-   raise NotImplementedError
+    raise NotImplementedError
```
jorenham, line 78: This is causing the following fatal error in `protobuf==5.29.0`:

```
IndentationError: unindent does not match any outer indentation level (service.py, line 78)
```

PR #19445, opened 2024-11-28
patched GetCurrentTime to use correct timezone-aware datetime

### LINKED ISSUE #19161: DeprecationWarning: datetime.datetime.utcnow()
python

Remove DeprecationWarning for `datetime.datetime.utcnow()`

Change `datetime.datetime.utcnow()` in _python/google/protobuf/internal/well_known_types.py_ to
- `datetime.datetime.now(datetime.UTC)` if the object should be timezone aware or
- `datetime.datetime.now(datetime.UTC).replace(tzinfo=None)` if it should be like now

no alternatives

**Additional context**
no context

b-d-e:
Resolves #19161 by addressing the deprecation warning.

See python 3.12 [release notes](https://docs.python.org/3/whatsnew/3.12.html) for motivating python change:
> [datetime](https://docs.python.org/3/library/datetime.html#module-datetime): [datetime.datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime)’s [utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) and [utcfromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) are deprecated and will be removed in a future version. Instead, use timezone-aware objects to represent datetimes in UTC: respectively, call [now()](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) and [fromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) with the tz parameter set to [datetime.UTC](https://docs.python.org/3/library/datetime.html#datetime.UTC). (Contributed by Paul Ganssle in [gh-103857](https://github.com/python/cpython/issues/103857).)

Existing unit tests should handle these changes.

Without this, this warning appears when using protobuf:
```
.venv/lib/python3.12/site-packages/google/protobuf/internal/well_known_types.py:178: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  self.FromDatetime(datetime.datetime.utcnow())
```


b-d-e:
This won't work as is under Python `3.8.x`, but this has not passed end-of-life support from Python. If desired, I can add a conditional check that change the behaviour based on if version <`3.9`.

JasonLunn:
@b-d-e - Can you take a look at the failing tests?

JasonLunn:
I see `AttributeError: module 'datetime' has no attribute 'UTC'` [here](https://github.com/protocolbuffers/protobuf/actions/runs/12655207524/job/35265009148#step:3:976), which seems directly related to your change.

You also probably need to rebase your changes to pick up test fixes that have landed since this PR was opened.

JasonLunn:
@b-d-e - could you rebase as well? The tests that are failing look like they're unrelated to this change.

b-d-e:
@JasonLunn should be up to date with `main` now

JasonLunn (approved):

haberman:
This was fixed separately in https://github.com/protocolbuffers/protobuf/pull/21341, so this PR should no longer be required.

Thanks for your contribution -- sorry we didn't get to it earlier.

PR #1293, opened 2016-03-03
Define all atomic functions for Atomic64 in atomicops_internals_generic_gcc.h

gongsu832:
Hi, I was compiling on Linux s390x (IBM mainframe) and was getting undefined references to some atomic functions. It turns out that the #ifdef **LP64** part of atomicops_internals_generic_gcc.h doesn't have all the atomic functions defined for Atomic64. I added them and now the compilation finishes successfully and passes all 6 tests in "make check".

PR #19528, opened 2024-12-05
feat(python_pyi): add generated pyi to py_proto_library

josephrubin:
Since protoc supports `pyi_out`, allow py_proto_library to generate the pyi files for editor support (otherwise, most editors cannot autocomplete with proto python output).

References: https://github.com/protocolbuffers/protobuf/pull/1293

Testing:

There are no tests specifically for py_proto_library, but when I run:

`~/protobuf/examples$ bazel build //:addressbook_py_pb2`

I get:

```
INFO: Analyzed target //:addressbook_py_pb2 (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:addressbook_py_pb2 up-to-date:
  bazel-bin/external/protobuf~/src/google/protobuf/_virtual_imports/timestamp_proto/google/protobuf/timestamp_pb2.py
  bazel-bin/external/protobuf~/src/google/protobuf/_virtual_imports/timestamp_proto/google/protobuf/timestamp_pb2.pyi
  bazel-bin/addressbook_pb2.py
  bazel-bin/addressbook_pb2.pyi
```

### DIFF
--- a/bazel/common/proto_common.bzl
+++ b/bazel/common/proto_common.bzl
@@ -186,6 +186,8 @@ def _compile(
 
     if plugin_output:
         args.add(plugin_output, format = proto_lang_toolchain_info.out_replacement_format_flag)
+        if proto_lang_toolchain_info.out_replacement_format_flag.startswith("--python"):
+            args.add(plugin_output, format = "--pyi_out=%s")
     if proto_lang_toolchain_info.plugin:
         tools.append(proto_lang_toolchain_info.plugin)
         args.add(proto_lang_toolchain_info.plugin.executable, format = proto_lang_toolchain_info.plugin_format_flag)
--- a/bazel/py_proto_library.bzl
+++ b/bazel/py_proto_library.bzl
@@ -64,12 +64,16 @@ def _py_proto_aspect_impl(target, ctx):
     proto_root = proto_info.proto_source_root
     if proto_info.direct_sources:
         # Generate py files
-        generated_sources = proto_common.declare_generated_files(
-            actions = ctx.actions,
-            proto_info = proto_info,
-            extension = "_pb2.py",
-            name_mapper = lambda name: name.replace("-", "_").replace(".", "/"),
-        )
+        generated_sources = []
+        for extension in ["_pb2.py", "_pb2.pyi"]:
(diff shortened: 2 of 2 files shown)


bazel/common/proto_common.bzl
```diff
@@ -186,6 +186,8 @@ def _compile(
 
     if plugin_output:
         args.add(plugin_output, format = proto_lang_toolchain_info.out_replacement_format_flag)
+        if proto_lang_toolchain_info.out_replacement_format_flag.startswith("--python"):
+            args.add(plugin_output, format = "--pyi_out=%s")
```
josephrubin, line 190: I'm not entirely sure the best way to do this. I would prefer to just add the `--pyi_out` to python's `proto_lang_toolchain` instantiation but the `command_line = "--python_out=%s",` can only contain one `%s` string.

This is my first time in this code base so I would appreciate any suggestions.

shaod2:
Why relying on the output of py_proto_library is required? I would think that it should be consumed by other bazel rules. Could you just run protoc to get the desired result?

josephrubin:
I think that `bazel build` should be the canonical way to generate build artifacts in a bazel repo. You could run `protoc` directly but bazel in theory has a better understanding of the packaging, build directory, and so forth for a Python project  (as well as uses the hermetic `protoc`.

The resulting pyi files _could_ be used as input to, for example, another rule that generates documentation based on build output.

But I also understand where you are coming from.

bazel/common/proto_common.bzl
haberman: I think we should have you define another toolchain here for pyi: https://github.com/protocolbuffers/protobuf/blob/73a5afe278170c1771993cedea6694a2700569b8/python/build_targets.bzl#L553-L561

Then it can be another dependency to the rule, and we can invoke `proto_common.compile()` twice, once for each toolchain.

haberman:
Sorry for the slow review. I've just added some suggestions for how to do this without changing `proto_common.bzl`.

mkruskal-google:
https://github.com/protocolbuffers/protobuf/pull/21567 seems to do this the way Josh was suggesting, closing this to follow up on that PR

PR #19576, opened 2024-12-10
bazel 8 compile error fix

20083017:


shaod2:
Hi thanks for the contribution.

Could you provide any context of this PR?

zhangskz:
I believe these were added to ensure protobuf continues to work for older Bazel versions i.e. Bazel 7: https://github.com/protocolbuffers/protobuf/commit/41be4091d580f32e3a7c2bb0304af087bfbd9626, so this change would break those.

Do you have more details on the error you hit and why you believe this would help?

comius:
Problem is the reference to top-level cc_proto_aspect, which needs to be done via bazel_features/proto_bazel_features, so that on Bazel 8, there's no reference to in, while it needs to be present on older Bazels.

20083017:
> Problem is the reference to top-level cc_proto_aspect, which needs to be done via bazel_features/proto_bazel_features, so that on Bazel 8, there's no reference to in, while it needs to be present on older Bazels.
> 
> I also believe this problem is only on main branch, not on 29.x releases.

yes, that's the reason! so, what should I do to fix this ^_^

20083017:
> Hi thanks for the contribution.
> 
> Could you provide any context of this PR?

image

20083017:
seems like the pr is working with bazel 7.3.0, @shaod2 @zhangskz @comius

20083017:
image
image
or maybe something like this, which is also works for both 7.3.0 and 8 @shaod2 @zhangskz @comius

20083017:
@shaod2  please remove the "wait for user action" label and help pushing the pr review moving on

comius:
Hey all, I'll prepare a change internally that addresses this problem. PR doesn't help, because the code is ATM forked and I need to defork it.

comius:
First piece of the puzzle: https://github.com/bazel-contrib/bazel_features/pull/88

20083017:
> First piece of the puzzle: [bazel-contrib/bazel_features#88](https://github.com/bazel-contrib/bazel_features/pull/88)

good to know that~

PR #19670, opened 2024-12-15
Ignore files that are generated files by the protocol buffer compiler

y-yagi:
It seems that some files that are generated by the compiler for tests aren't ignored correctly. So I always get the untracked files when I run the tests.
https://github.com/protocolbuffers/protobuf/blob/59a8de6610b6adbd49211a7805c505877f3ace0c/ruby/Rakefile#L23-L39

This PR ignores those files to avoid treating them as untracked files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,6 +176,11 @@ ruby/tests/multi_level_nesting_test_pb.rb
 ruby/tests/service_test_pb.rb
 ruby/tests/test_import_proto2_pb.rb
 ruby/tests/test_ruby_package_proto2_pb.rb
+ruby/tests/basic_test_features_pb.rb
+ruby/tests/generated_code_editions_pb.rb
+ruby/tests/repeated_field_test_pb.rb
+ruby/tests/stress_pb.rb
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -176,6 +176,11 @@ ruby/tests/multi_level_nesting_test_pb.rb
 ruby/tests/service_test_pb.rb
 ruby/tests/test_import_proto2_pb.rb
 ruby/tests/test_ruby_package_proto2_pb.rb
+ruby/tests/basic_test_features_pb.rb
+ruby/tests/generated_code_editions_pb.rb
+ruby/tests/repeated_field_test_pb.rb
+ruby/tests/stress_pb.rb
+ruby/tests/utf8_pb.rb
```
y-yagi, line 183: `protobuf/ruby` also has `.gitignore`. So I'm wondering whether it's better to move these files under that file.
 https://github.com/protocolbuffers/protobuf/blob/main/ruby/.gitignore

zhangskz (approved):

Issue #12947, opened 2023-05-31
src build on windows not working


gregorsoll:
**What version of protobuf and what language are you using?**
Version: main/v3.6.0/v3.5.0 etc. (NOTE: please try updating to the latest version of protoc/runtime possible beforehand to attempt to resolve your problem)
Language: C++/Java/Python/C#/Ruby/PHP/Objective-C/Javascript

**What operating system (Linux, Windows, ...) and version?**
Windows 

**What runtime / compiler are you using (e.g., python version or gcc version)**
VC 14.16.27023

**What did you do?**
1. cloned src and submodules
2. installed bazel, c++ buildtools 
3. bazel build :protoc :protobuf

**What did you expect to see**
build completed

**What did you see instead?**
 C:/users/xyz/_bazel_xyz/y7fcwcs2/external/com_google_protobuf/src/google/protobuf/compiler/csharp/BUILD.bazel:22:11: Compiling src/google/protobuf/compiler/csharp/csharp_repeated_enum_field.cc [for tool] failed: (Exit 2): cl.exe failed: error executing command (from target @com_google_protobuf//src/google/protobuf/compiler/csharp:csharp) C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe /nologo /DCOMPILER_MSVC /DNOMINMAX /D_WIN32_WINNT=0x0601 /D_CRT_SECURE_NO_DEPRECATE ... (remaining 65 arguments skipped)
cl : Command line warning D9002 : ignoring unknown option '-std=c++14'
external/com_google_protobuf/src/google/protobuf/compiler/csharp/csharp_repeated_enum_field.cc(31): fatal error C1083: Cannot open include file: 'google/protobuf/compiler/csharp/csharp_repeated_enum_field.h': No such file or directory
[...]

PR #19679, opened 2024-12-16
bazel: Remove hardcoded dependency on `//:protoc` from language runtimes

fmeum:
Without this change, language runtimes still result in a build of `//:protoc` even with a prebuilt `proto_toolchain` registered or `--proto_compiler` set to a precompiled protoc.

### DIFF
--- a/bazel/private/BUILD
+++ b/bazel/private/BUILD
@@ -1,4 +1,5 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load(":current_protoc.bzl", "current_protoc")
 load(":native_bool_flag.bzl", "native_bool_flag")
 
 package(default_applicable_licenses = ["//:license"])
@@ -26,6 +27,11 @@ toolchain_type(
 toolchain_type(
     name = "python_toolchain_type",
     visibility = ["//visibility:public"],
+)
+
+current_protoc(
+    name = "current_protoc",
+    visibility = ["//:__subpackages__"],
 )
 
 bzl_library(
--- a/protobuf.bzl
+++ b/protobuf.bzl
@@ -2,6 +2,7 @@ load("@bazel_skylib//lib:versions.bzl", "versions")
 load("@rules_cc//cc:defs.bzl", "objc_library")
 load("@rules_python//python:defs.bzl", "py_library")
 load("//bazel/common:proto_info.bzl", "ProtoInfo")
+load("//bazel/private:current_protoc.bzl", "ProtocFilesToRun")
 
 def _GetPath(ctx, path):
     if ctx.label.workspace_root:
@@ -310,7 +311,7 @@ def _internal_gen_well_known_protos_java_impl(ctx):
             args.add_all([src.path[offset:] for src in dep.direct_sources])
 
     ctx.actions.run(
-        executable = ctx.executable._protoc,
+        executable = ctx.attr._protoc[ProtocFilesToRun].files_to_run,
         inputs = descriptors,
         outputs = [srcjar],
         arguments = [args],
@@ -335,9 +336,7 @@ internal_gen_well_known_protos_java = rule(
             default = False,
         ),
         "_protoc": attr.label(
-            executable = True,
-            cfg = "exec",
-            default = "//:protoc",
+            default = "//bazel/private:current_protoc",
         ),
     },
 )
@@ -373,7 +372,7 @@ def _internal_gen_kt_protos(ctx):
             args.add_all([src.path[offset:] for src in dep.direct_sources])
 
     ctx.actions.run(
-        executable = ctx.executable._protoc,
+        executable = ctx.attr._protoc[ProtocFilesToRun].files_to_run,
         inputs = descriptors,
         outputs = [srcjar],
         arguments = [args],
@@ -398,9 +397,7 @@ internal_gen_kt_protos = rule(
             default = False,
         ),
         "_protoc": attr.label(
-            executable = True,
-            cfg = "exec",
-            default = "//:protoc",
+            default = "//bazel/private:current_protoc",
         ),
     },
 )


fmeum:
CC @comius @alexeagle

fmeum:
@shaod2 Would you be available to review this?

mkruskal-google:
Hey Fabian,

So IIUC, this PR is trying to make the protoc binary used by proto_library customizable.  While that might be a possible long-term solution, it opens the door to a lot of potential misuse.  Notably, mixing alternate implementations or different versions of protoc is typically unsupported and can have unexpected results.  We have poison pills to avoid some of these situations outside of Bazel, but they haven't been added to all our languages yet.  A more conservative change would be to just have some kind of flag that toggles between build-from-source and the prebuilt releases we publish to github, without allowing arbitrary binaries to be injected.

Additionally, this PR is touching protobuf.bzl, which just contains our legacy internal macros/rules (which we want to delete).  These have been replaced by the rules in //bazel, which were recently moved from other Bazel repos.

fmeum:
> Additionally, this PR is touching protobuf.bzl, which just contains our legacy internal macros/rules (which we want to delete). These have been replaced by the rules in //bazel, which were recently moved from other Bazel repos.

These macros are still used to bootstrap the Java well-known protos for the default Java `proto_lang_toolchain`, which is why `protoc` ends up being compiled from source even when using a precompiled `protoc` binary.

> So IIUC, this PR is trying to make the protoc binary used by proto_library customizable. While that might be a possible long-term solution, it opens the door to a lot of potential misuse. Notably, mixing alternate implementations or different versions of protoc is typically unsupported and can have unexpected results.

Isn't the `protoc` binary used by `proto_library` already configurable via `--proto_compiler` or toolchain registration (the latter with `--incompatible_enable_proto_toolchain_resolution`)? The aim of this PR is specifically to ensure that if users use custom `protoc` binary, it's the only `protoc` used in the entire build. Right now, when configuring a non-default `protoc` binary in any of the existing ways, the `//:protoc` from this repo still ends up being used to bootstrap protos.

adincebic:
@fmeum do you know what's required for this to land?

fmeum:
@mkruskal-google Friendly ping

mbland:
As another data point, I made a [patch from this pull request for bazel-contrib/rules_scala](https://github.com/bazel-contrib/rules_scala/blob/38391c13fef4e32c0347b6be6702a7ab690784f5/protoc/0001-protobuf-19679-rm-protoc-dep.patch). I created it while working towards the upcoming Bzlmod-enabled `rules_scala` release (bazel-contrib/rules_scala#1482), when our Windows builds failed after upgrading past `protobuf` v21.7 in bazel-contrib/rules_scala#1710 .

The breakage was due to later `protobuf` versions pushing some file paths past the cursed 260 character limit, breaking Windows MSVC source builds. Since [protobuf support for MSVC is officially dead for precisely this reason](https://github.com/bazel-contrib/rules_scala/tree/38391c13fef4e32c0347b6be6702a7ab690784f5#windows-msvc-builds-of-protobuf-broken-by-default) (references below), this patch became essential to fixing our Windows builds via [--incompatible_enable_proto_toolchain_resolution](https://bazel.build/reference/command-line-reference#common_options-flag--incompatible_enable_proto_toolchain_resolution).

- https://github.com/protocolbuffers/protobuf/issues/12947
- [protobuf.dev News Announcements for Version 30.x:Poison MSVC + Bazel](https://protobuf.dev/news/v30/#poison-msvc--bazel)
- https://github.com/protocolbuffers/protobuf/issues/20085

[The precompiled `protoc` toolchain mechanism](https://github.com/bazel-contrib/rules_scala/tree/38391c13fef4e32c0347b6be6702a7ab690784f5?tab=readme-ov-file#protoc) is only mandatory for Windows builds, but also shaves minutes off the build for Linux and macOS. Though we'll have to maintain the patch as long as current `protobuf` releases are in widespread use, it would be nice to not have to require it for future releases.

mkruskal-google:
Sorry for the delay in responding here.  While we do understand that this is a strict improvement to the current situation, it only further endorses a deprecated set of Bazel features we want to turn down completely.  We would prefer to focus on providing a future-facing solution to the root problem, and are working towards providing a _supported_ mechanism to use prebuilt binaries with the protobuf rules.

Note: we have no intention of killing MSVC+Bazel support before this solution is in place.  There is currently an opt-out flag to allow building from source with MSVC that will be in place until at least 34.0

fmeum:
@mkruskal-google Could you elaborate on what these deprecated Bazel features are? As far as I can tell, proto toolchainization is a much more robust and modern approach than `--proto_compiler` (which should definitely just go away).

Please also keep in mind that prebuilt binaries are not the end of the story. You won't be able to supply prebuilt binaries for all OS/architecture pairs, but users on such platforms can't be limited to from source builds.

mbland:
> Note: we have no intention of killing MSVC+Bazel support before this solution is in place. There is currently an opt-out flag to allow building from source with MSVC that will be in place until at least 34.0. Setting `--output_user_root=C:/` should get you past the 255 character limit issue, as we've have CI to prevent further accidental breakages for a while now

`rules_scala` uses Bazel's Buildkite CI system, [bazelbuild/continuous-integration](https://github.com/bazelbuild/continuous-integration). https://github.com/bazel-contrib/rules_scala/pull/1710#issuecomment-2691560698 shows [the Windows MSVC build breaking on `protobuf` v29.3](https://buildkite.com/bazel/rules-scala-scala/builds/5394#01954e5a-8f5b-4880-befa-1bce0d21d512/75-179):

```txt
ERROR: C:/tools/msys64/home/b/_bazel_b/xknd5zlq/external/com_google_protobuf/src/google/protobuf/compiler/java/BUILD.bazel:87:11:
  Compiling src/google/protobuf/compiler/java/java_features.pb.cc [for tool] failed: (Exit 2): cl.exe failed:
  error executing CppCompile command
  (from target @@com_google_protobuf//src/google/protobuf/compiler/java:java_features_bootstrap)
  C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\cl.exe ...
  (remaining 1 argument skipped)
--
  | external/com_google_protobuf/src/google/protobuf/compiler/java/java_features.pb.cc(6): fatal error C1083:
    Cannot open include file: 'google/protobuf/compiler/java/java_features.pb.h': No such file or directory
```

The [full build results](https://buildkite.com/bazel/rules-scala-scala/builds/5394#01954e5a-8f5b-4880-befa-1bce0d21d512/75-179) show `bazel --output_user_root=C:/b`. Perhaps we could try setting it to `C:/` to see if that one character makes a difference. Trying to run under `cmd.exe` instead of MSYS2 may also help (since `C:/tools/msys64/home/b/...` appears in the `ERROR:` message), and be good for other reasons. That might take some work, since the bulk of `rules_scala` tests are Bash scripts invoking `bazel` in various ways. But both potential experiments are effectively moot, at least for the moment. Right now we're using the precompiled `protoc` to make progress, and it's making both CI and local development faster.

FWIW, I'm totally on board with replacing the old and busted with the new hotness. (This explains my commitment to completing bazel-contrib/rules_scala#1482.) So if there's a better solution around the corner, I'll be amongst the first to adopt it. But until then, or until this pull request lands, we _have_ to keep patching `protobuf` to unblock Windows builds of `rules_scala`. We'll also keep patching it for the overall benefit of avoiding `protoc` recompiles on all platforms.

mkruskal-google:
O yea msys2 is definitely the problem there.  You're getting extra characters from `tools/msys64/home`, which is going to push you over the limit for sure.  We're *very* close to the limit and I think `C:\tmp` is the extent of what Bazel+MSVC can handle today.

alexeagle:
@mkruskal-google I think we've confirmed that the toolchain flag isn't being removed in the near future, could we consider merging this one? Will be needed to get pre-built protoc working as expected.

vibbix:
Whats the current status on this patch?

mkruskal-google (approved):

PR #19731, opened 2024-12-18
Fix ruby `has_...?` method

### LINKED ISSUE #18807: [Ruby] optional field's presence checker method `has_...?` is broken on JRuby and has cosmetic issue on MRI
Version: 4.28.2
Language: Ruby

All operating systems.

Ruby 3.3.5 / JRuby 9.4.8.0

**What did you do?**

Protobuf definition: https://github.com/sass/sass/blob/eba0a58b46788b893983c2886c1be3b9c2d260e1/spec/embedded_sass.proto#L807-L834

Test code:
``` ruby
require_relative 'embedded_sass_pb'

value = Sass::EmbeddedProtocol::Value::Color.new(
  space: 'rgb',
  channel1: 0, # this is an optional field
  channel2: 0, # this is an optional field
  channel3: 0, # this is an optional field
  alpha: nil # this is an optional field
)

# Note, the JRuby bug only occurs on a deserialized object. `.new` created objects have no issue. 
value = Sass::EmbeddedProtocol::Value::Color.decode(value.to_proto)

puts "  value.has_channel1? = #{value.has_channel1?}" # <- should return true
puts "  value.has_alpha?    = #{value.has_alpha?}" # <- should return false
puts "==="
puts value.to_s # <- note that the behavior changes on JRuby before and after this call
puts "==="
puts "  value.has_channel1? = #{value.has_channel1?}"  # <- should return true
puts "  value.has_alpha?    = #{value.has_alpha?}" # <- should return false
```

```
  value.has_channel1? = true
  value.has_alpha?    = false
===
<Sass::EmbeddedProtocol::Value::Color: space: "rgb", channel1: 0.0, channel2: 0.0, channel3: 0.0>
===
  value.has_channel1? = true
  value.has_alpha?    = false
```

**What did you see instead?**

### MRI

`.has_...?` returns `0` instead of `true`. Note: `0` in ruby is considered as truthy, so this is just a cosmetic issue in MRI. 

```
  value.has_channel1? = 0
  value.has_alpha?    = false
===
<Sass::EmbeddedProtocol::Value::Color: space: "rgb", channel1: 0.0, channel2: 0.0, channel3: 0.0>
===
  value.has_channel1? = 0
  value.has_alpha?    = false
```

### JRuby

`.has_...?` always return `false`, until `.to_s` is called on the object. Somehow calling `.to_s` changes behavior of `.has_...?`, that after the call `.has_...?` returns the expected value.

```
  value.has_channel1? = false
  value.has_alpha?    = false
===
<Sass::EmbeddedProtocol::Value::Color: space: "rgb", channel1: 0.0, channel2: 0.0, channel3: 0.0>
===
  value.has_channel1? = true
  value.has_alpha?    = false
```

https://github.com/sass-contrib/sass-embedded-host-ruby/pull/246

ntkme:
- Fixes #18807

### DIFF
--- a/ruby/tests/basic_proto2.rb
+++ b/ruby/tests/basic_proto2.rb
@@ -43,6 +43,15 @@ module BasicTestProto2
 
       m = TestMessage.new(:optional_int32 => nil)
       refute m.has_optional_int32?
+      refute TestMessage.descriptor.lookup('optional_int32').has?(m)
+
+      m = TestMessage.new(:optional_int32 => 0)
+      assert m.has_optional_int32?
+      assert TestMessage.descriptor.lookup('optional_int32').has?(m)
+
+      m = TestMessage.decode(TestMessage.encode(m))
+      assert m.has_optional_int32?
+      assert TestMessage.descriptor.lookup('optional_int32').has?(m)
 
       assert_raises NoMethodError do
         m.has_repeated_msg?
--- a/ruby/ext/google/protobuf_c/message.c
+++ b/ruby/ext/google/protobuf_c/message.c
@@ -362,7 +362,8 @@ static VALUE Message_field_accessor(VALUE _self, const upb_FieldDef* f,
       if (!upb_FieldDef_HasPresence(f)) {
         rb_raise(rb_eRuntimeError, "Field does not have presence.");
       }
-      return upb_Message_HasFieldByDef(Message_Get(_self, NULL), f);
+      return upb_Message_HasFieldByDef(Message_Get(_self, NULL), f) ? Qtrue
+                                                                    : Qfalse;
     case METHOD_WRAPPER_GETTER: {
       Message* self = ruby_to_Message(_self);
       if (upb_Message_HasFieldByDef(self->msg, f)) {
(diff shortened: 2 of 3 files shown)


JasonLunn:
Thanks for the PR! Can you add a regression test that fails without this fix?

ruby/tests/basic_proto2.rb
```diff
@@ -43,6 +43,15 @@ def test_has_field
...
       refute m.has_optional_int32?
+      refute TestMessage.descriptor.lookup('optional_int32').has?(m)
+
+      m = TestMessage.new(:optional_int32 => 0)
+      assert m.has_optional_int32?
+      assert TestMessage.descriptor.lookup('optional_int32').has?(m)
+
+      m = TestMessage.decode(TestMessage.encode(m))
+      assert m.has_optional_int32?
+      assert TestMessage.descriptor.lookup('optional_int32').has?(m)
```
ntkme, line 54: @JasonLunn Regression test added, specifically, the highlighted lines here are failing on without the patch.

colinbendell:
Thanks for putting the PR together! I had it on my "holiday TODO" list!

JasonLunn (approved):

PR #19776, opened 2024-12-23
feat(csharp/Reflection): change MessageParser.CreateTemplate to be public

ahmednfwela:
this is a subset of the changes in #7722 where `MessageParser.CreateTemplate` is made public, this removes the need to use reflection to create protobuf messages given a descriptor/parser.

future work: maybe for the next major release we can change the generated code to use `MessageParser.CreateTemplate`, fully solving #6401.


JasonLunn:
Thank you for your contribution, @ahmednfwela. Although this PR is small, the impact of expanding the API surface could be significant and we need additional time to review it to make sure it aligns with our long term goals across the languages that Protobuf supports.

esrauchg:
Unfortunately our C# expertise is somewhat bandwidth limited right now, but as a lower-c#-context protobuf team member it seems a bit odd to me that this method should be a public method on `Parser`, since this is not really a parse related operation.

A similar shape of method is commonly an instance method on Message type, eg in Java `var builder = msg.newBuilderForType()` and in C++ the `Message* m = msg.New()` method (where there is a common pattern to pass around a SomeMsg.defaultInstance() as a prototype and then newBuilderForType()/New() on that in code that wants to create more instances of the same type).

Just to understand the possible design scope better, would a method on IMessage that is parallel to those precedents also solve for the need that you have?

ahmednfwela:
@esrauchg 
The whole point of this PR is to enable the creation of an arbitrary message given its parser.

This opens the door for something like https://github.com/protocolbuffers/protobuf/pull/7722 to unify the creation interface, so  that the consumer can modify the factory logic to their needs (e.g. introducing reusable message pools).

I would personally not recommend moving the method on `IMessage`, because it forces us to have an instance of `IMessage` to be able to create an empty message .

But having it on the parser is similar to having an `IMessageFactory` which is a much better design

esrauchg:
Unfortunately we need to be very conservative with API changes in C#, because we likely can never make a breaking change in the C# API ever. And this has too high of a chance of being a regretted API at some future date, since its making something on "MessageParser" and it isn't a Parsing topic.

Especially for something that is not clearly and directly precedented in the other languages, we unfortunately can't take API changes of this kind today, even though we understand that the new API provides value to you.

Note that I originally asked about exposing it on IMessage (which would match Java, where you have to pass around a default instance as the 'prototype' to stamp out new instances), but with updated view we actually couldn't make that change either because IMessage adding any methods would be a breaking change for old codegen, and since we can't ever do a breaking change in C# we likely can never add any methods to IMessage either.

I understand that this is locally annoying, and you may wish to use a vendored copy of Protobuf with your local patches, but it's the unfortunate reality of maintaining a library that can't make any breaking changes.

ahmednfwela:
@esrauchg 

- Other official protobuf runtimes (C++, Java, Python, Go) support constructing new message instances at runtime given a descriptor (via DynamicMessage, reflection factories, or generated parser + new instances).
- The protobuf data model inherently assumes you can 
  1. discover a message via descriptors and 
  2. create an empty instance to populate/merge/parse. This is not an optional convenience; it is the basis for schema‑driven tooling.
- All this PR does is expose this "required" functionality to library users, just as other protobuf runtimes are doing!

Also I do disagree with the notion that this library can't have breaking changes, as the protobuf spec itself is evolving in a major way, this is specifically documented in [protobuf itself](https://protobuf.dev/editions/overview/#lifecycles)

esrauchg:
> Other official protobuf runtimes [...]

There's a long standing open feature request for DynamicMessage in C# here which would provide this behavior: https://github.com/protocolbuffers/protobuf/issues/658

You can chime in there about the importance of this behavior. That behavior could be added as a non-breaking change to C#, the only topic is about staffing the change. In current staffing we're unfortunately not likely to be able to staff adding major new API behaviors to C# Proto today, even where it is a "parity with other Java/Go/Python supported behaviors" topic.

> Also I do disagree with the notion that this library can't have breaking changes, as the protobuf spec itself is evolving in a major way, this is specifically documented in [protobuf itself](https://protobuf.dev/editions/overview/#lifecycles)

By breaking changes I mean "someone has a project which compiles when used on Protobuf 3.25 but does not compile when used with a published release whose number is something higher than 3.25 (eg 3.26 or 4.x)

The proposed PR to only add behavior isn't a breaking change; and the relevance of "no breaking changes" is relating to necessary conservatism because we likely wouldn't be able to rollback this PR if we decide we didn't like it in even in 2035. And for adding something to IMessage it would immediately be a breaking change that would be highly disruptive for current users, that case is basically a type of change that we just can't make at all in any language.

The idea of Editions is to let us evolve what features exist in .proto files without having any breaking changes against files people have already written; the idea is that for any given .proto file the behaviors that you observe on that will be the same forever, only when someone makes a change to the .proto file will it change behavior. Any breaking-change on gencode, no matter how benign-seeming, would only happen if your .proto files bump to use the new Edition number, and so if you're still using Proto3 syntax inputs you can reliability continue without any observable breaking changes on your gencode api (and in fact Editions is also designed so that you can safely bump to a new Edition and set features to perfectly keep the bahavior you had, even where default behavior on that new Edition is different).

It's also worth being clear that we also will make major version bumps on runtime APIs (= semver breaking changes on runtime API) in some languages, and in those langugaes we wouldn't need to be as paranoid because we _could_ rollback a PR like this in less than a decade. There's a few languages where the ecosystem basically cannot handle major version bumps because of ecosystem expectations about diamond dependency problems. C# happens to be an example of a language where we know it would be extremely disruptive to users if we had to do a C# 4.x release and so this API could never be removed, Python and C++ are examples of languages where we could do a major version bump and so an API like this could be removed and then we don't need to be quite as vigilantly terrified of additional public APIs.

ahmednfwela:
The only reason you would ever need to roll back this PR is if you decide to completely change the parser, which would be a breaking change itself.

Also @JamesNK himself said he preferred this approach:

https://github.com/protocolbuffers/protobuf/issues/6401#issuecomment-659935786

Not to mention that other libraries are already using this method (even though it's internal) via reflection, that's just how important is it in the protobuf model, it's honestly a wonder how it stayed internal all those years.

here is some proof: 
- https://github.com/channeldorg/channeld-unity-mirror/blob/f11103c98f6146f1cfeadb3ecba7dd57d17a0588/Assets/channeld/ChannelDataView.cs#L37
- https://github.com/SciSharp/protobuf.Text/blob/1daa9d75d72c2fc02c20270bc488d8b50b80561e/src/protobuf.Text/ProtobufAdapter.cs#L153-L161

PR #19808, opened 2024-12-27
Switch `rules_ruby` to the supported version from `bazel-contrib`

JasonLunn:


conformance/conformance_ruby.rb
```diff
@@ -1,12 +1,14 @@
-#!/usr/bin/env ruby
-#
 # Protocol Buffers - Google's data interchange format
 # Copyright 2008 Google Inc.  All rights reserved.
 #
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
+$LOAD_PATH.unshift ENV['PWD']
```
mkruskal-google, line 8: I like the `-I` hashbang better :P. Either way we should add a TODO and file an issue with rules_ruby.  The root problem here is that there's no way to strip the `ruby/lib` prefix from the ruby runtime code, so importing `google/protobuf` doesn't work in Bazel.  This will affect any downstream Ruby/Bazel users

Disable_bundle_install.patch
```diff
@@ -0,0 +1,37 @@
+Subject: [PATCH] Disable bundle install
```
mkruskal-google, line 1: We shouldn't submit patches, is this going to be upstreamed?  Can we wait for it?

ruby/Gemfile
```diff
@@ -1,3 +1,24 @@
...
+
+group :test do
+  gem "test-unit", '~> 3.0', '>= 3.0.9'
+end
+
+group :development do
+  gem "rake", ">= 13"
+  gem "rake-compiler", "~> 1.1.0"
+  gem "rake-compiler-dock", "= 1.2.1"
+end
```
mkruskal-google, line 24: Why does this need to duplicate the gemspec file?  This seems fragile

JasonLunn: I agree that it is fragile, but I don't have a better alternative right now that fits the constraints.

Context: `rules_ruby` is opinionated in that it requires a `Gemfile.lock` to be checked in, and it goes the extra mile to tell `bundler` to freeze the lock file, meaning that if `bundle install` cannot satisfy what's in the lock file as-is, it fails.

In an earlier commit before adopting `rules_ruby`, I was able to check in a `Gemfile.lock`, but we don't tell bundler to freeze, and it turns out that this was allowing subtle divergence across interpreters, which they silently allowed and wallpapered over by updating the `Gemfile.lock` in-place in CI.

Using explicit `gem` statements here gives us more control than the all-in-one `gemspec` statement, and it results in a `Gemfile.lock` that both CRuby and JRuby can successfully perform a `bundle install` on while frozen.

Disable_bundle_install.patch
JasonLunn: There is some troubleshooting to do here before we get to a fix, I owe the maintainers of `rules_ruby` a Dockerfile that can repro the issue. I'm inclined to submit this for now so that we can part with our fork and come back to clean this up.

conformance/conformance_ruby.rb
JasonLunn: I changed this because local testing showed that once we changed the `rb_binary` rule to pass a `main` attribute, the hashbang line wasn't used, presumably because the file is getting read by a ruby parser directly rather than a shell. I will add a TODO pointing to the feature request.

ruby/Gemfile
mkruskal-google: Checking in the lock file seems fine to me, but why do we need the gemspec *and* Gemfile?

JasonLunn: They're used for different purposes. The gemspec file contains a bunch of information need for publishing gems, not just the information needed for resolving dependencies.

conformance/conformance_ruby.rb
mkruskal-google: That's strange.. my local tests definitely read the hashbang line after adding the `main` attribute

Disable_bundle_install.patch
mkruskal-google: yea I guess this is fine for a dev dependency.  It'll just block us making it available as a real dep

ruby/Gemfile
mkruskal-google: Do we have test coverage of both?

JasonLunn: Yes. Regular tests will fail if there is a problem with the `Gemfile`. If there are problems with the Gemspec, the [Ruby Install](https://github.com/protocolbuffers/protobuf/blob/main/.github/workflows/test_ruby.yml#L176-L191) tests should fail. There is nothing checking that they're in sync with one another.

mkruskal-google (approved):

PR #19971, opened 2025-01-12
Add `to_proto` to descriptor classes

### LINKED ISSUE #12044: Ruby: Convert Descriptor to DescriptorProto
Ruby, proto2 in particular.

Exactly the same as #9425 for C#, we need to send a DescriptorProto to use the BigQuery Storage write streaming API. The Ruby bindings allow us to generate the message class and get it as a Descriptor, but there doesn't seem to be an API to convert this to a DescriptorProto as required for the proto2 writer_schema argument.

Just like #9426 for C#.

Manually constructing a Descriptor that is identical to the message class's fields.

**Additional context**

y-yagi:
Some APIs (e.g. [BigQuery Storage](https://github.com/googleapis/googleapis/blob/ab61c52324b39d1c8619a96ab19cc6c173b13f2c/google/cloud/bigquery/storage/v1/protobuf.proto#L39)) use DescriptorProto for schema descriptions. But,  in Ruby, there are no methods for that so far.

This PR implements the method for that(`to_proto`) to descriptor classes.

Fixes #12044.

### DIFF
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -979,6 +979,24 @@ static VALUE FieldDescriptor_options(VALUE _self) {
   return field_options;
 }
 
+/*
+ * call-seq:
+ *     FieldDescriptor.to_proto => FieldDescriptorProto
+ *
+ * Returns the `FieldDescriptorProto` of this `FieldDescriptor`.
+ */
+static VALUE FieldDescriptor_to_proto(VALUE _self) {
+  FieldDescriptor* self = ruby_to_FieldDescriptor(_self);
+  upb_Arena* arena = upb_Arena_New();
+  google_protobuf_FieldDescriptorProto* proto = upb_FieldDef_ToProto(
+    self->fielddef, arena);
+  size_t size;
+  const char* serialized = google_protobuf_FieldDescriptorProto_serialize(
+    proto, arena, &size);
+  VALUE proto_class = rb_path2class("Google::Protobuf::FieldDescriptorProto");
+  return Message_decode_bytes(size, serialized, 0, proto_class, false);
+}
+
 static void FieldDescriptor_register(VALUE module) {
   VALUE klass = rb_define_class_under(module, "FieldDescriptor", rb_cObject);
   rb_define_alloc_func(klass, FieldDescriptor_alloc);
@@ -998,6 +1016,7 @@ static void FieldDescriptor_register(VALUE module) {
   rb_define_method(klass, "get", FieldDescriptor_get, 1);
   rb_define_method(klass, "set", FieldDescriptor_set, 2);
   rb_define_method(klass, "options", FieldDescriptor_options, 0);
+  rb_define_method(klass, "to_proto", FieldDescriptor_to_proto, 0);
   rb_gc_register_address(&cFieldDescriptor);
   cFieldDescriptor = klass;
 }
--- a/ruby/ext/google/protobuf_c/defs.c
+++ b/ruby/ext/google/protobuf_c/defs.c
@@ -1461,6 +1461,24 @@ static VALUE ServiceDescriptor_options(VALUE _self) {
   return service_options;
 }
 
+/*
+ * call-seq:
+ *     ServiceDescriptor.to_proto => ServiceDescriptorProto
+ *
+ * Returns the `ServiceDescriptorProto` of this `ServiceDescriptor`.
+ */
+static VALUE ServiceDescriptor_to_proto(VALUE _self) {
+  ServiceDescriptor* self = ruby_to_ServiceDescriptor(_self);
+  upb_Arena* arena = upb_Arena_New();
+  google_protobuf_ServiceDescriptorProto* proto = upb_ServiceDef_ToProto(
+    self->servicedef, arena);
+  size_t size;
+  const char* serialized = google_protobuf_ServiceDescriptorProto_serialize(
+    proto, arena, &size);
+  VALUE proto_class = rb_path2class("Google::Protobuf::ServiceDescriptorProto");
+  return Message_decode_bytes(size, serialized, 0, proto_class, false);
+}
+
 static void ServiceDescriptor_register(VALUE module) {
   VALUE klass = rb_define_class_under(module, "ServiceDescriptor", rb_cObject);
   rb_define_alloc_func(klass, ServiceDescriptor_alloc);
@@ -1470,6 +1488,7 @@ static void ServiceDescriptor_register(VALUE module) {
   rb_define_method(klass, "file_descriptor", ServiceDescriptor_file_descriptor,
                    0);
   rb_define_method(klass, "options", ServiceDescriptor_options, 0);
+  rb_define_method(klass, "to_proto", ServiceDescriptor_to_proto, 0);
   rb_include_module(klass, rb_mEnumerable);
   rb_gc_register_address(&cServiceDescriptor);
   cServiceDescriptor = klass;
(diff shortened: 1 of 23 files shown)


upb/reflection/file_def.c
```diff
@@ -181,7 +181,7 @@ bool upb_FileDef_Resolves(const upb_FileDef* f, const char* path) {
   return false;
 }
 
-static char* strviewdup(upb_DefBuilder* ctx, upb_StringView view) {
+static char* _strviewdup(upb_DefBuilder* ctx, upb_StringView view) {
```
y-yagi, line 184: I renamed this because there is the same name function in `upb/util/def_to_proto.c`. 
https://github.com/protocolbuffers/protobuf/blob/c8ad77475634f77468cd237dd10bcff978a2372a/upb/util/def_to_proto.c#L57

The amalgamation merges `upb` files into one file. So, it causes a compile error when the functions have the same name.

ruby/lib/google/protobuf/ffi/descriptor.rb
```diff
@@ -106,6 +106,15 @@ def options
         end
       end
 
+      def to_proto
```
mkruskal-google, line 109: I think we should follow the C++ convention of naming these something like "copy_to".  I think it's an important distinction that these aren't necessarily 100% preserved in a roundtrip.  They also unfortunately reach into a lot of our internals (that we usually want to hide), so I think this alternate naming will help discourage use a bit.

y-yagi: Thanks for your review! 
You mean, `#copy_to_proto` or just `#copy_to`?

In my personal opinion, using `#to_xx` is natural for generating another object from an object in Ruby's context(e.g. `Hash#to_a`). Sometimes those methods are costable. But, at Ruby, I think understandable naming is preferred.
But I would like to follow the Protocol Buffers's rule. So let me clarify the name.

JasonLunn:
Thanks for contributing this PR, @y-yagi!

The tests are really close to passing, but there seems to be an issue on macOS. Both the Ruby 3.0 and Ruby 3.4 tests are failing on macOS with a failure like this:

```
Error: test_enum_descriptor_to_proto(BasicTest::MessageContainerTest): Google::Protobuf::ParseError: Error occurred during parsing
/private/var/tmp/_bazel_runner/88067fd1245450c0f2a417255e46b72b/sandbox/darwin-sandbox/1336/execroot/com_google_protobuf/bazel-out/darwin_x86_64-fastbuild/bin/ruby/tests/basic.runfiles/com_google_protobuf/ruby/tests/basic.rb:701:in 'Google::Protobuf::EnumDescriptor#to_proto'
/private/var/tmp/_bazel_runner/88067fd1245450c0f2a417255e46b72b/sandbox/darwin-sandbox/1336/execroot/com_google_protobuf/bazel-out/darwin_x86_64-fastbuild/bin/ruby/tests/basic.runfiles/com_google_protobuf/ruby/tests/basic.rb:701:in 'BasicTest::MessageContainerTest#test_enum_descriptor_to_proto'
     698:     def test_enum_descriptor_to_proto
     699:       enum_descriptor = TestDeprecatedEnum.descriptor
     700: 
  => 701:       assert_instance_of Google::Protobuf::EnumDescriptorProto, enum_descriptor.to_proto
     702:     end
     703: 
```

The same tests are passing on Linux and under FFI on the macOS, and the other descriptor tests are working as expected - it is just enum descriptor test without FFI that is having an issue. I'm going to re-run the tests now to see if the test failures are stable or if the tests are flaky.

Do you have a macOS machine available to troubleshoot this on in the meantime?

JasonLunn (changes requested):

ruby/lib/google/protobuf/ffi/descriptor.rb
mkruskal-google: Well tbh we haven't standardized this yet.  In C++ we use `copyTo(DescriptorProto* out)`, and in Java we use `DescriptorProto toProto()` (and python doesn't offer this at all).  The Java implementation doesn't copy, so the name seems accurate but it makes it *really* easy to reach into our internals.  For example, we try to hide raw editions features because they're easy to misuse, but in Java you can simply call `toProto().getOptions().getFeatures()` whenever you want because we need to preserve round-trip behavior.

If `to_proto` is really more ergonomic for Ruby I guess it's fine, it's just kindof unfortunate

y-yagi: Thanks for your detailed explanation! 
I understand your concern. But, I still `#to_proto` is better for Ruby. And, we already have `#to_proto` for the Message class in Ruby.  
https://github.com/protocolbuffers/protobuf/blob/ae3015c78a0ea3baa76f3ad556c0ecd575f2e618/ruby/lib/google/protobuf/message_exts.rb#L24-L26

So if we choose `#copy_to`,  it may cause confusion. So if it's OK, I would like to keep this name :pray:.

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -558,12 +577,35 @@ static VALUE FileDescriptor_options(VALUE _self) {
...
+  FileDescriptor* self = ruby_to_FileDescriptor(_self);
+  upb_Arena* arena = upb_Arena_New();
+  google_protobuf_FileDescriptorProto* file_proto = upb_FileDef_ToProto(
+    self->filedef, arena);
+
+  size_t size;
+  const char* serialized = google_protobuf_FileDescriptorProto_serialize(
+    file_proto, arena, &size);
+
+  upb_Arena_Free(arena);
```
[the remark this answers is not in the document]
y-yagi: This was just my miss :pray:. I fixed to free an Area in all functions.
a35588ea2a0f1d3cbcc976f38207a09f7ab3b923

ruby/ext/google/protobuf_c/defs.c
```diff
@@ -1298,6 +1378,27 @@ static VALUE EnumDescriptor_options(VALUE _self) {
...
+ * Returns the `EnumDescriptorProto` of this `EnumDescriptor`.
+ */
+static VALUE EnumDescriptor_to_proto(VALUE _self) {
+  EnumDescriptor* self = ruby_to_EnumDescriptor(_self);
+  upb_Arena* arena = upb_Arena_New();
+  google_protobuf_EnumDescriptorProto* proto = upb_EnumDef_ToProto(
+    self->enumdef, arena);
+
+  size_t size;
+  const char* serialized = google_protobuf_EnumDescriptorProto_serialize(
```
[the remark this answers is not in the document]
y-yagi: Thanks for your suggestion!

I don't have a macOS so I can't reproduce this issue on my local...

I have added the validation for `size` and `serialized` for confirmation. 51e54e3d5e014536cdfb6cf89e09ed43d80dd609 
Could you rerun the CI?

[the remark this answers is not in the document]
y-yagi: Hmm. The enum descriptor test is still getting `Google::Protobuf::ParseError`. And, it seems that other tests are getting the same error. I will check.

y-yagi:
I'm doubting I released an arena too early. So I fixed the timing of it.  1afd9e8f6ff76b9b0b937b6117c0bce7091c8fa4

y-yagi:
Ruby tests are passed. But, unrelated tests(Bazel, C++, PHP Extension) are failed. I will rebase with the latest code.

y-yagi:
I've rebased with the latest code.

JasonLunn (approved):

JasonLunn (approved):

Issue #12887, opened 2023-05-23
cc_library `protoc_lib` defines `main()`


scotthart:
**What version of protobuf and what language are you using?**
Version: v23.1
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Linux ... 6.1.20-2rodete1-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.20-2rodete1 (2023-04-21) x86_64 GNU/Linux

**What runtime / compiler are you using (e.g., python version or gcc version)**
Debian clang version 14.0.6

**What did you do?**
Ran the tests defined in 
https://github.com/googleapis/google-cloud-cpp/blob/main/generator/internal/predicate_utils_test.cc
using bazel test

**What did you expect to see**
Tests executing as they did in protobuf v21.12

**What did you see instead?**
The usage error from invoking protoc without any arguments.

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).
```
gdb home/username/.cache/bazel/_bazel_username/f564c2f714c6826c87550e7c5eec6cf0/execroot/google_cloud_cpp/bazel-out/k8-dbg/bin/generator/internal_predicate_utils_test
GNU gdb (GDB) 10.0-gg5
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-grtev4-linux-gnu".
Type "show configuration" for configuration details.

<http://go/gdb-home FAQ: http://go/gdb-faq Email: c-toolchain-team>
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
Reading symbols from home/username/.cache/bazel/_bazel_username/f564c2f714c6826c87550e7c5eec6cf0/execroot/google_cloud_cpp/bazel-out/k8-dbg/bin/generator/internal_predicate_utils_test...
Unable to determine compiler version.
Skipping loading of libstdc++ pretty-printers for now.
Loading libc++ pretty-printers.
Non-google3 binary detected.
(gdb) break CommandLineInterface::PrintHelpText
Function "CommandLineInterface::PrintHelpText" not defined.
[...]

PR #20016, opened 2025-01-16
Fix cc_library protoc_lib defines main() #12887

tongke6:
Fix issue #12887

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -234,27 +234,16 @@ alias(
     visibility = ["//visibility:public"],
 )
 
-cc_binary(
+alias(
     name = "protoc",
-    copts = COPTS,
-    linkopts = LINK_OPTS,
+    actual = "//src/google/protobuf/compiler:protoc",
     visibility = ["//visibility:public"],
-    deps = ["//src/google/protobuf/compiler:protoc_lib"],
 )
 
-cc_binary(
+alias(
     name = "protoc_static",
-    copts = COPTS,
-    features = select({
-        # This isn't possible on mac because there is no static library for lcrt0.o
-        "@platforms//os:osx": [],
-        "//build_defs:config_osx": [],
-        # When cross-compiling we need to statically link all C++ libraries.
-        "//conditions:default": ["fully_static_link"],
-    }),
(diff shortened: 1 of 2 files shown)


mkruskal-google (approved):

src/google/protobuf/compiler/BUILD.bazel
```diff
@@ -250,12 +247,29 @@ cc_binary(
...
+    features = select({
+        # This isn't possible on mac because there is no static library for lcrt0.o
+        "@platforms//os:osx": [],
+        "//build_defs:config_osx": [],
+        # When cross-compiling we need to statically link all C++ libraries.
+        "//conditions:default": ["fully_static_link"],
+    }),
+    linkopts = LINK_OPTS,
     visibility = ["//visibility:public"],
+    deps = ["//src/google/protobuf/compiler:protoc_lib"],
```
mkruskal-google, line 272: You're gonna have to duplicate the deps here because we enable strict layering in our repo.  We should probably create a new variable to avoid duplication

tongke6: How about use relative label here，have changed it to `:protoc_lib`

mkruskal-google: That doesn't change anything.  Strict layering requires you explicitly declare all your direct dependencies.  So this fails to build because a bunch of headers are missing

tongke6: Got it, thank you for your explanation.

BUILD.bazel
```diff
@@ -234,27 +233,16 @@ alias(
     visibility = ["//visibility:public"],
 )
 
-cc_binary(
```
mkruskal-google, line 237: Looks like there's still a bunch of test failures related to the alias/cc_binary switch.  There are a pretty large number of places where we assume protoc and protoc_static get dumped directly into bazel-bin, and this change moves the binaries somewhere else.

I agree with the direction, and it would be nice to keep as much as possible in //src/google/protobuf/compiler, but I think the cc_binary targets need to live here for now (to avoid a major refactor)

PR #20033, opened 2025-01-17
Security fix to stop initializing a class before checking that it is assignable to MessageLite

yn33:
I have submitted a related bug report to OSS VRP. This is a fairly simple fix, which stops arbitrary classes from being initialized. Only classes which implement MessageLite are now allowed. If there is a reason that classes should be deserialized even when they do not implement MessageLite, then this fix will break them, but I do not see how that would ever be the case.

### DIFF
--- a/java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java
+++ b/java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java
@@ -1451,7 +1451,7 @@ public abstract class GeneratedMessageLite<
           throw new ClassNotFoundException();
         }
       }
-      return messageClass
+      return messageClass;
     }
   }
 
--- a/java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java
+++ b/java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java
@@ -1445,7 +1445,13 @@ public abstract class GeneratedMessageLite<
     }
 
     private Class<?> resolveMessageClass() throws ClassNotFoundException {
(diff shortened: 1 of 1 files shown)


java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java
```diff
@@ -1445,7 +1445,13 @@ protected Object readResolve() throws ObjectStreamException {
...
 
     private Class<?> resolveMessageClass() throws ClassNotFoundException {
-      return messageClass != null ? messageClass : Class.forName(messageClassName);
+      if (messageClass == null) {
+        Class messageClass = Class.forName(messageClass, false, getClass().getClassLoader()); 
+        if (!MessageLite.class.isAssignableFrom(c)) {
+          throw new ClassNotFoundException();
+        }
+      }
+      return messageClass
```
shaod2, line 1454: missing a ";"

shaod2 (changes requested):

java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java
yn33: fixed

mkruskal-google:
This is still failing tests

acozzette:
Thank you for the fix (and reporting the bug), but we actually just landed a similar fix for this in b9971a8be237a1fee521041cf276e6f8ccd2ecd2.

yn33:
My bug report related to this was closed as not reproducible: https://issuetracker.google.com/issues/390233266

This must be a mistake since a fix was clearly made.

PR #20052, opened 2025-01-20
Fix missing port_undef

wilstoff:
unparser.cc, arenaz_sampler_test.cc, and message_module.cc included port_def.inc but forgot to include the undef file.  This can cause some issues on which absl macros are defined:

In file included from
/opt/protobuf-29/include/google/protobuf/arena.h:36: In file included from
/opt/protobuf-29/include/google/protobuf/arena_align.h:63: /opt/protobuf-29/include/google/protobuf/port_def.inc:542:5: error: function-like macro 'ABSL_HAVE_FEATURE' is not defined \#if ABSL_HAVE_FEATURE(address_sanitizer)

### DIFF
--- a/python/google/protobuf/pyext/message_module.cc
+++ b/python/google/protobuf/pyext/message_module.cc
@@ -347,3 +347,5 @@ PyMODINIT_FUNC PyInit__message() {
 
   return m;
 }
+
+#include "google/protobuf/port_undef.inc"
--- a/src/google/protobuf/arenaz_sampler_test.cc
+++ b/src/google/protobuf/arenaz_sampler_test.cc
@@ -598,3 +598,5 @@ TEST(ThreadSafeArenazSamplerTest, UsedAndWasted) {
 }  // namespace internal
 }  // namespace protobuf
 }  // namespace google
+
+#include "google/protobuf/port_undef.inc"
--- a/src/google/protobuf/json/internal/unparser.cc
+++ b/src/google/protobuf/json/internal/unparser.cc
@@ -905,3 +905,5 @@ absl::Status BinaryToJsonStream(google::protobuf::util::TypeResolver* resolver,
 }  // namespace json_internal
 }  // namespace protobuf
 }  // namespace google
+
+#include "google/protobuf/port_undef.inc"


src/google/protobuf/arenaz_sampler_test.cc
```diff
@@ -592,6 +592,9 @@ TEST(ThreadSafeArenazSamplerTest, UsedAndWasted) {
   });
   SetThreadSafeArenazSampleParameter(oldparam);
 }
+
+#include "google/protobuf/port_undef.inc"
```
mkruskal-google, line 596: This should be at the bottom of the file

wilstoff: my bad, got confused thinking this was an h file and it needing to be before the endif like any.h

#include "google/protobuf/port_undef.inc"

#endif  // GOOGLE_PROTOBUF_ANY_H__

wilstoff: fixed and signed.

mkruskal-google (approved):

src/google/protobuf/arenaz_sampler_test.cc
wilstoff: Will look at the tests failing tomorrow as well as changing my signature email to match the commits

python/google/protobuf/pyext/message_module.cc
```diff
@@ -347,3 +347,5 @@ PyMODINIT_FUNC PyInit__message() {
 
   return m;
 }
+
+#include "google/protobuf/port_undef.inc
```
mkruskal-google, line 351: Missing end-quote here

wilstoff: Sorry for delay, been busy at work upgrading protobuf and coming back to this side fix.  Comedy of errors you can tell i'm not actually building locally before pushing.   Will reach out again when i consider this ready.

wilstoff: I believe i should have fixed the author and the silly typos.  Tried to do some local building, but our work machines have some clashes with our version of ccache etc.  But i believe the changes to be relatively safe to run on your build nodes.

`ERROR: /home/bstoffel/.cache/bazel/_bazel_bstoffel/d6c92e1c5e534d31833b4c3ac5594187/external/com_google_absl/absl/base/BUILD.bazel:60:11: Compiling absl/base/log_severity.cc failed: (Exit 1): ccache failed: error executing CppCompile command (from target @@com_google_absl//absl/base:log_severity) /usr/bin/ccache -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer '-std=c++14' -MD ... (remaining 34 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
/usr/bin/ccache: invalid option -- 'U'
Target //python:google/protobuf/pyext/_message.so failed to build`

acozzette:
@wilstoff I actually suspect that this is not the right fix for your compiler error. The main purpose of `port_undef.inc` is to prevent our internal macros from leaking out into non-protobuf code. While it is important to include `port_undef.inc` at the end of our header files, it doesn't do anything useful in .cc files.

mkruskal-google:
I assumed this was a unity build situation 🤷‍♂️

python/google/protobuf/pyext/message_module.cc
mkruskal-google: How are you running tests?  You shouldn't need ccache

mkruskal-google (approved):

PR #20195, opened 2025-02-01
Fix the C++ compiler version check

y-yagi:
Since fe535930d3183b46d7e088683cbe8c49285715f8, the minimum C++ version is C++17.


JasonLunn:
CC: @tonyliaoss

JasonLunn (approved):

PR #16240, opened 2024-03-20
C#: add option and code generation for null reference types

tonydnewell:
This replaces the closed PR https://github.com/protocolbuffers/protobuf/pull/13218

It adds an option for generating C# code that handles null reference types.
See https://github.com/grpc/proposal/blob/master/L110-csharp-nullable-reference-types.md

Further work could be done to:
- add tests to check the generated code is as expected - there doesn't appear to be these kind of tests for other code generation
- change `#nullable enable annotations` to `#nullable enable` - but this requires further codegen changes
  - to annotate handling calling the `Google.Protobuf.Reflection` API that doesn't yet handle null reference types
  - to handle types such as `string` and `bytes` that use null to indicate presence

PR #20242, opened 2025-02-05
C#: add option and code generation for null reference types

Falco20019:
This replaces the abandoned https://github.com/protocolbuffers/protobuf/pull/16240 and should address the open points mentioned.


jskeet:
> `bytes` will stay `ByteString` as it's ensured to use `ByteString.Empty` as default and is not allowing to be set to null.

Completely agree about regular `bytes` fields - my comment on #16240 was around `BytesValue` as the wrapper message for `bytes` fields (e.g. `TestWellKnownTypes.bytes_field`). That's what I'd expect to become a `ByteString?`-typed property, in the same way that a `StringValue` becomes `string?`. (The default value of `TestWellKnownTypes.bytes_field` really is null.)

It's possible you already cover this - at a *very* quick glance, it looks like you haven't, but I could be wrong. If you don't cover this already, but you agree with my analysis, if you could make that change, that would be great.

Falco20019:
Thanks for the clarification and quick response.. I will definetly check and fix it if it's off.

Falco20019:
I fixed the `BytesValue`, it behaves now as follows:
* `bytes` -> `ByteString`
* `google.protobuf.BytesValue` -> `ByteString?`

While checking the generated code, I remembered that in the past I had the assumption that repeated and map fields could have `null` as value for wrapper types. This is due to i.e. `repeated google.protobuf.Int32Value` being generated as `RepeatedField<int?>` instead of `RepeatedField<int>`. Accordingly, `repeated google.protobuf.StringValue` would now generate to `RepeatedField<string?>` as well.

Technically it's correct and `RepeatedField` mentions that it is not allowing `null` even though it's containing a nullable type. As you mentioned they are runtime-checked against `null` on `Add`, which makes it a bit strange API-wise. It would be possible to change it for NRT generation, but this might force users to check/cast their nullables first when switching to NRT generation. So while they should have checked it in the first place, it would be a breaking change API wise in case the users didn't pass `.Value`. So I would tend to keep it as is and have them generated `RepeatedField<string?>` and `RepeatedField<ByteString?>` for the nullables with `null` not being allowed.

jskeet:
I'll need to go back and revisit why we end up with `RepeatedField<int?>` for `repeated Int32Value`. It could well just be a mistake, I'm not sure. Likewise I'll see about repeated fields for `StringValue`. If it *was* a mistake for `Int32Value`, I think it would be better not to propagate it for string/bytes. Chances are I'll need to write all of this up in a doc somewhere, otherwise I'll just end up chasing my own tail. Thanks for flagging it up though.

Falco20019:
I already found the corresponding location in `RepeatedMessageFieldGenerator::GenerateMembers` and will either suppress it for only the `StringValue` and `BytesValue`, or for all wrapper types.

Will fix it to whatever you declare correct behavior once you find time to figure it out. Only leaving this as a reminder for myself by then ;)

Falco20019:
@JasonLunn I merged back `protocolbuffers:main` in the hope that this fixes the failing tests :)

JasonLunn:
@Falco20019 - can you add some tests that cover the new behavior?

Falco20019:
I can try to. The tests would need to check if a generated field is marked as nullable or not I assume. I could also duplicate the existing tests to check if they behave the same for both variants (which they should). But maybe doing it for a subset would suffice. I will try to think about it over the weekend.

Falco20019 (commented):
@JasonLunn I added the tests as requested. Please have them run. I needed to add `net8.0` targets as it's otherwise not fully possible to get NRTs tested. I therefore also changed the `global.json`. Please make sure that the .NET 8 SDK is available to the CI (not sure how to check it).

I tried to create regression tests also for the nullability on repeated and map value types, but it's sadly not possible as the runtime doesn't have this information anymore by design. If someone else would know how to do it, we can add them later on. It's just about a point that came up here in the discussion that Jon is not sure himself if they should be nullable or not. They had been in the past, but probably just as an oversight as adding `null` will throw an exception.

Falco20019:
Thanks, I will have a look at the workflows as the failing tests use .NET 6 SDK which just has no way to test the .NET 8 features. Will try to update them tomorrow at work.

Falco20019:
@JasonLunn You might want to replace Jon with yourself until tests are really working and the code becomes reviewable.

As I sadly have no permissions to run the tests myself, we need to do trial-and-errors here. I updated from `main` to ensure the C++ test failures should be gone. I adjusted the `test_csharp.yml` as much as possible to use the .NET 8 SDK.

It's using `us-docker.pkg.dev/protobuf-build/containers/test/linux/csharp:7.1.2-3.1.415-6.0.100-d9624f2aa83cba3eaf906f751d75b36aacb9aa82` which is based on .NET 6.0.100 which is sadly not able to use .NET 8 in the `csproj`. I sadly don't see how I could create a new image here to use.

I'm also not aware of how BAZEL works and therefore not able to adjust `csharp/src/Google.Protobuf.Conformance\BUILD.bazel` myself to use .NET 6 + .NET 8 for testing. I saw that line 61 has the `net6.0` build, but don't know how to ensure both will be used.

I therefore would ask for help of the team on complete those two missing links.

Falco20019:
@JasonLunn It seems that the tests are always using `protocolbuffers/protobuf/.github/workflows/test_csharp.yml@refs/heads/main` and therefore it will not be possible to see the change of the workflows here. A separate rule is also ensuring I'm not allowed to adjust the workflow. And adjusting the workflow would break all other builds until this is in.

Cyclic dependency :) The Windows build is the only one that doesn't use `us-docker.pkg.dev/protobuf-build` and therefore succeeds. It's also the only one that tests ALL test cases (including `net462`) and therefore needs no adjustment to the pipeline. But even then, the containers are still only including `6.0.100` and not `8.0.100` which would be needed. Still not sure where that is maintained.

JasonLunn:
To work around the test workflow changes, I manually imported this PR into our internal systems. With minor manual tweaks, it ran cleaner as https://github.com/protocolbuffers/protobuf/pull/20561, but still has a couple of issues, some of which may need to be addressed internally because they'll mean changes to the images used to run CI.

1. In the vanilla linux test config, the `8.0.100` SDK is not installed. It sees the configuration select it, but it only has `3.1.415` and `6.0.100` available to choose from.
2. On the linux aarch64 test config, where the `8.0.100` SDK is available, we see a different error: `error NETSDK1129: The 'Publish' target is not supported without specifying a target framework. The current project targets multiple frameworks, you must specify one of the following frameworks in order to publish: net462, net6.0, net8.0`
1. I see a lot of errors coming out of an internal-only test that have to do with collisions in the 'Google.Protobuf.TestProtos` namespace. Is it straightforward to put the new test protos in the `Nrt` directory into a distinct namespace so that they don't clash with their non-NRT equivalents?

Falco20019:
1. As the image name suggests, only those two were selected similar to the ps1 script. I assume a new image is needed. Also not sure why the 3.1 SDK is still in there, as I didn't see it being used anymore. The `global.json` in the root pins the used SDK for everything. Will give it another look if there is anything modifying the SDK used.

2. I messed this up as I hoped to be able to have it as one publish. I will fix it to have both run as distinct ones.

3. It should be possible to have different namespaces using an option. I tried to avoid it to keep the changes to the tests simpler. I will have to change the `using` then on each test to import the correct namespace depending on the target framework. But if it helps your internals, I will look into it.

I have some other topics today on my plate but hope to still find time to adjust the PR late today (latest by tomorrow).

JasonLunn:
I'm checking on what I can do to with regard to details of the internal test

global.json
```diff
@@ -1,6 +1,6 @@
 ﻿{
   "sdk": {
-    "version": "6.0.100",
```
jskeet, line 3: We should work out some ordering here. I would like to separate "using the .NET 8 SDK" from "generating NRTs" - just updating to use .NET 8 may cause some significant difficulties internally, for reasons I'd rather not go into.

I'd personally be okay with this repo not having any NRT-based C# code in - or possibly explicitly keep it present but not compiled for now. (I think it's good that we've generated it and can check that it *does* compile, but we might not want to check it in at the moment.)

jskeet (approved):
Okay, I've now found time to review this (at least - thanks for your patience) - focusing on the generator.

I'd be fine with the generator changes going in on their own, so long as we make sure it's labeled as an experimental feature to start with. We might even want to open a fresh issue  for feedback and include a link to that in release notes.

Protobuf folks: I'm confident that this won't affect generated code when the flag isn't explicitly enabled, which is the most important thing.

My personal preference (as noted on global.json) is to avoid switching SDKs within the same change as adding the experimental feature - but I defer to the protobuf team on that. (In particular, *if* it already just works on the internal build, maybe it's just fine...)

haberman:
Could you expand the PR description to clearly state what the PR is doing?

As someone who hasn't been following this discussion closely, it's not clear to me what the approach is, or how it differs from the PR that preceded it (https://github.com/protocolbuffers/protobuf/pull/16240).

Here are some specific questions the PR description could answer:

1. Is the PR adding any new public API(s), or is it merely adding annotations that codify the semantics of existing APIs?
2. This PR appears to add a new codegen flag `enable_nrt`. What is the intention for this flag? Who do we expect will be using it? Is it intended to eventually become the default?

jskeet:
@haberman: Nullable Reference Types (NRT) are a feature from C# 8 onwards, where you can annotate that a reference is expected to potentially be null, or should not be null (either accepted as a parameter or returned by a method/property). The compiler is then able to warn if it thinks the code *using* that annotated code is doing the wrong thing. For example:

```csharp
public static string ReturnNonNullString() { ... }
public static string? ReturnMaybeNullString() { ... }
...
string x = ReturnNonNullString();
string? y = ReturnMaybeNullString();
Console.WriteLine(x.Length); // No warning
Console.WriteLine(y.Length); // Warning that you might be dereferencing a null value
```

It doesn't change *execution-time* semantics at all. (Indeed, the annotations can be "wrong" in that it's *possible* for a null value to be returned even a method is annotated not to do so, etc.)

Now, back to protobuf: the `enable_nrt` flag will start using these annotations on the generated code, so that users of that generated code gain more compile-time safety around nullness. It doesn't add any APIs, it just adds an annotation for some parameters and return types in the generated code. (There's the possibility of a separate change - which is a lot of work - to add nullable reference types within the protobuf runtime library itself. This is definitely *not* that change.)

I would expect users who use NRT within the rest of the code base to *possibly* specify `enable_nrt` for existing proto compilation (but they'll then potentially need to change the consuming code, because there will be warnings) and more likely specify it for *new* proto compilation.

Defaulting: I wouldn't expect the flag to ever become the default within protoc - support for nullable reference types still isn't even the default in the C# compiler, although the flag to enable it is present in the default project template for new projects. The MSBuild target for using protoc could change whether *it* uses enable_nrt by default - potentially based on whether the rest of the project does.

That's a fairly brief summary - the exact details of NRT are *very* complex, unfortunately. (We're still working through exactly how to standardize it in the C# specification...) https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references gives more details if you want to delve further though.

global.json
Falco20019: @JasonLunn Still waiting for feedback if I should mark it as experimental and how to proceed with the tests.

Falco20019:
> Could you expand the PR description to clearly state what the PR is doing?
> 
> As someone who hasn't been following this discussion closely, it's not clear to me what the approach is, or how it differs from the PR that preceded it (#16240).
It doesn't differ from the approach, it only picks up where Tony left off. As https://github.com/protocolbuffers/protobuf/issues/6632 is already very detailed on what the intention is, I avoided duplicating it. Jon gave a pretty good picture and if you prefer to have some of it in the initial post, I can copy parts of it or link to it.

global.json
JasonLunn: I'm aligned with checking in as little generated code as possible, and I think separating the SDK update from the NRT feature is a great suggestion. I don't feel strongly about the naming question, and would defer to @haberman and / or @jskeet on that point.

Falco20019:
As there hasn't been any clear preference for the last weeks, I will now continue with using `enable_nrt` and revert the .NET 8 stuff to the extend that we have the stuff around but that it's not run automatically or being generated by the pipeline.

Falco20019:
@JasonLunn Can you please let the tests run? It should be now according to the requested state. Your internal tests should hopefully also work now again :) If the tests fail, I will make sure to merge `main` in and resolve any conflicts.

Falco20019:
@JasonLunn Merged in `main`, can you run the tests again to see if the failing ones work now? Seemed unrealted to my changes for me.

Falco20019:
@JasonLunn I see a lot of `well_known_types_staleness_test` failings and errors that `starts_with` was not found. Nothing I did touch. Is this normal?

Falco20019:
Just had another look and saw that some commits after the merge should fix it. Did another merge, hoping that the tests run now.

Falco20019:
@JasonLunn @jskeet Friendly ping. This is waiting for a week now to have the tests run :) If I should trigger a team handle instead, please let me know. Those errors should not have been caused by me but some other changes on the branch, so I'm sorry for the additional work we all have to go through.

jskeet:
I'm afraid I have zero time to do anything on Protobuf at the moment. I do understand this is frustrating.

Falco20019:
@jskeet I only tagged you hoping you might know a team member/handle to trigger instead. I fully understand you don't have time to label the issue over and over again 👍 I will refrain from tagging you and wait for Jason (or any other member) to aid me on this.

jlumsden-mts:
I've just come across this after updating some C# projects to enable NRT and this would be great to enable for generated protobuf code especially for preventing errors when accessing null submessages.

It looks like the C# tests and most of the tests are passing here?

Falco20019:
I would think so. It's up to the team to decide if/how/when to merge it. Sadly nothing more I can do myself here... It had been approved by Jon, so there shouldn't be any blockers around.

@JasonLunn Could you try to push this along? Thanks.

stebet:
I would love to see this change merged soon. NRT support in the generated code (even if experimental) could be immensely useful for newer .NET client code.

sstockton-auvesy-mdt:
We would also love to see this change merged. It would allow us to use Mapperly with protobuf generated classes which would save us a lot of time and effort.

Falco20019:
@JasonLunn Friendly ping. Any chance we can make progress on this? It's done since June (4 months) and awaiting just to be merged by the team :(

Falco20019:
@protocolbuffers/protobuf-csharp Not sure if I can ping the group, but would be great if someone responsible for C# could merge this.

ngbrown:
Based on other changes in the csharp folders, since @jskeet has already approved it, the next steps seem to be that @googleberg does something to make the `copybara-service` do its thing to copy the code into a new pull request and actually merge the changes.

iqexdev02:
It would be great if someone could poke the right person to get this moving forward. This is a gap in the language support that sorely needs closing, given that the .NET ecosystem as whole has already adopted nullable annotations.

juwens:
@Falco20019 i see there are 4 failing tests. Do these tests need to be fixed before merging, or are they false positives and can be ignored?

image

Falco20019:
Should be unrelated but feel free to rebase/merge in again to see if they would fail on latest.

esrauchg:
Sorry about slow updates here. This is still actively (slowly) being discussed within Google.

I can give an update about why this is not straightforward and where we are here.

The main topic is regarding .NET Framework users who cannot compile the code which has nullable annotations in it. Broadly Protobuf's standard stance is that we set our minimum supported language/compiler/platform/whatever, and we constrain our feature set to that version. For example in Java we support Java 8 minimum, we don't do anything which is not Java 8 compatible. Inside our runtimes we can sometimes have conditional branches for newer runtimes to get better behavior in new environment while maintaining our compatibility, with only one set of source files.

Unfortunately this is not really possible with C# nullable annotations: there is no way we can have the gencode have those annotations while still having it be source compatible with our complete support matrix.

The second topic is, given this is sufficiently important, how can we give people a lever to opt into "not compatible with our complete support matrix". Since doing so is unprecedented across all languages, we don't have a good playbook of precedents to follow for how to do it.

The PR proposes a protoc flag; this functionally works, but protoc flags are something that the Protobuf team would like to avoid essentially at all costs. It has many problems, but one major one is that the protoc design is that people are expected to run protoc many times in isolation (each time with a -I for its includes). Each invocation will create gencode which references _other_ gencode produced independently by a different invocation. The gencode basically has to just 'know' what the other invocation is going to produce for the final gencode to successfully compile, and any flag that influences gencode creates compatibility problems based on the combination. And its an extremely difficult property to test because you get into combinatorial support matrix problems of "one set of gencode was made with protoc 26 with flags [x, y, z] enabled and has outgoing references to other imports, the imports are created with protoc 33 which was run with flags [a, b, c] enabled, does that work or not?".

Two other alternatives that have been discussed internally are:

1) Options in .proto files. These days it would probably mean an Edition Feature, where each .proto file could then opt in to the new behavior and the corresponding gencode only works on correspondingly supported versions.

2) #ifdefs PROTOBUF_ENABLE_NRT in the gencode which you can turn on in the csproj, effectively making it so the gencode itself is the same no matter what and then the entire project is guaranteed to have a consistent state of it all on-or-off.

Broadly speaking the current discussion seems to be leaning towards that this should be supported as an option in the .proto file as a better solution compared to a protoc-flag or #ifdefs, but there's not clear consensus yet.

We are open to more feedback about those or other possible options here to help inform next steps.

Thanks!

Falco20019:
Not sure if the flags are that much of a problem for C# users as basically everyone uses Grpc.Tools which should apply the same flags all the time. So MSBuild would be able to supply the correct flag for the correct TFM.

But I do understand your points and that the other options would avoid problems. I like the idea of the feature edition as it would generate cleaner code than the ifdefs but both sound fine to me to circumvent the issues.

I sadly won't have time myself to supply another PR for it as our team stopped using ProtoBuf and gRPC completely a couple months ago. I would just have invested the time necessary to bring it to an end looking at how long this took already.

stebet:
Starting to feel like death-by-committee :(

jskeet:
Genuinely active - hoping to put this into a feature flag.

diegofrata:
@jskeet @esrauchg I refactored @Falco20019 branch to remove the protoc flag and use an Edition Feature instead. I avoided touching the implementation itself, the goal was to get the work which has already been reviewed and approved across the finish line. See #25789

csharp/src/AddressBook/AddressBook.csproj
```diff
@@ -11,4 +11,17 @@
...
+    <None Include="**/*.pb*.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' != 'net8.0' ">
+    <Compile Include="**/*.pb.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
+    <Compile Include="**/*.pb.nrt.cs" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
leonclara685-lab: Tienes la razón

jskeet:
@diegofrata Thanks for that - I'm talking internally about the best process for introducing a C# features proto, as there may well be internal changes made as well. Your PR will be very useful, but I may well need to start internally and it will appear externally in due course. (And yes, hopefully the implementation itself can stay as-is.)

diegofrata:
@jskeet no worries. I am still working out a few details out and bits I missed, but I will leave it ready in case there's a desire to merge the support as an Edition Feature.

stebet:
> Genuinely active - hoping to put this into a feature flag.

Thank you and @diegofrata for keeping this alive :)

Falco20019:
@diegofrata Thanks for taking over. Due to this and some other topics, our company decided to switch from gRPC + ProtocolBuffers to REST APIs. Therefore I sadly was not able to continue pushing this any further. I am thankful since it would make the time spent feel less wasted and would finally help others as-well.

jguamie:
We established we will follow up with a different approach. We will provide updates on the issue itself.

PR #20287, opened 2025-02-10
Ruby: Allow to get a file descriptor by a file name

y-yagi:
In Python, users can get a file descriptor by a file name. https://googleapis.dev/python/protobuf/latest/google/protobuf/descriptor_pool.html#google.protobuf.descriptor_pool.DescriptorPool.FindFileByName

This would be useful for sometimes. For example, GRPC Reflection allows to get a proto by file name.
https://github.com/grpc/grpc-proto/blob/483f11e91a30de9c11d65d872c09bc529e94e2ee/grpc/reflection/v1/reflection.proto#L45-L46

This feature would be useful to implement such a case.


y-yagi:
I'm not sure why, but it seems that `safe for tests` label has not worked correctly.

tonyliaoss:
Sorry, I'm giving it another kick. The tag was a bit flakey last week but it seemed to still work on other PRs recently.

JasonLunn (approved):

PR #20335, opened 2025-02-13
Load py_runtime_pair from rules_python for Bazel 8

justinwon777:
Bazel 8 requires loading `py_runtime_pair` from `rules_python` since it's deprecated in `bazel_tools` and causes failure. See https://github.com/bazelbuild/bazel/blob/8.0.0/tools/python/toolchain.bzl


justinwon777:
Hi @tonyliaoss. Do I need to do anything with the failing checks or is it fine?

PR #20392, opened 2025-02-19
Fix for performance penalty in deep dependencies

### LINKED ISSUE #20371: Runtime performance degradation for highly nested depending csharp files
Version: v3.21.9
Language: C#

Windows 11

csharp

**What did you do?**
[unittest_deep_dependencies.zip](https://github.com/user-attachments/files/18825472/unittest_deep_dependencies.zip)

1. [Optional] Take a structure of highly nested of .proto files that we have anonymized and flattened for testing purposes (see attached example: unittest_deep_dependencies folder)
2. [Optional] Generate the classes for the .proto files (also included in the attached example: csharp folder)
3. Copy the `csharp\UnittestDeepDependencies` folder into `\csharp\src\Google.Protobuf.Test`
4. Open solution in csharp\src\Google.Protobuf.sln
5. Create an additional Unit Test with the following content in `Google.Protobuf.Test\Reflection\DescriptorsTest.cs`:
```
        [Test]
        public void FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString()
        {
            Console.WriteLine("Running performance test for extensions");

            var stopwatchCached = Stopwatch.StartNew();

            var example = new UnittestDeepDependencies.Example();
            Console.WriteLine(example.ToString());

            stopwatchCached.Stop();

            Console.WriteLine($"Time elapsed: {stopwatchCached.Elapsed}");

            // actually test the elapsed time
            Assert.Less(stopwatchCached.Elapsed, TimeSpan.FromSeconds(2), "Extension performance was not as low as expected - please improve performance of extension resolution");
        }
```

6. Run the unit test. The unit test takes, in my case more than 1 minute.

A similar behavior as in v3.5.2, where printing a message would take milliseconds.

**What did you see instead?**
The print of a message takes > 1 minute
Image

- The performance is affected by the change introduced to support `extend` in the runtime: [Pull 8220](https://github.com/protocolbuffers/protobuf/pull/8220)
- The recursive iteration through the classes to search for extensions creates a runtime performance penalty.[See FileDescriptor.cs](https://github.com/protocolbuffers/protobuf/blob/main/csharp/src/Google.Protobuf/Reflection/FileDescriptor.cs#L421)
- This performance impact increases with the increased depth of the hierarchy.
- The test set provided is based on an actual data set where the following processing has been done:
  - Data has been anonymized.
  - Folder layout has been flattened. 
  - Hierarchy has been reduced to have a considerable runtime delay during unit testing.
  - Stripped of additional data, leaving only dependencies between .proto files (as this seems to be the major source of the delay). This may trigger `warning` during `protoc` execution due to unused dependencies.
  - Observe that the only existing Message is contained in the File117.cs

We have devised a proposal that we will bring as a PR.

fgarciacorona:
The goal of this PR is to fix #20371 (closed) #24267 and #24330 and improve the performance of csharp runtime when dealing with deep nested dependencies of .proto files.

- The fix consists in caching the recursive calls for the search of extensions.

- Two identical datasets (nested proto files) are created automatically using Bazel: one to test it with caching enabled, and another one to test it with no caching. Note: the same dataset cannot be used because once it is loaded inside the test, it cannot be unloaded.

- The created datasets have a width of 6 and a depth of 6, enough to showcase the dependency impact and to not impact the testing time for unit tests. This could be configured.

- In the dataset there is only one proto file with a message `Example` whose descriptor is loaded, and by doing so all the dependencies (and sub dependencies) that it has. 

- A set of benchmark metrics have been created to evaluate the performance impact before and after the fix.

- The assertion inside the unit test makes sure that when using metrics the number of traversed extensions (dependencies of the dependencies) when using caching, is reduced by a factor of 1000x. 

- The second assertion also makes sure that the time used to load the descriptor when using cache is less than the one without using cache.


tonyliaoss:
Hello @fgarciacorona, we won't be able to start reviewing your PR until the CLA has been signed.

Also, I suspect that we won't accept the newly added test protos, but we can discuss that after we start our review process.

fgarciacorona:
> Hello @fgarciacorona, we won't be able to start reviewing your PR until the CLA has been signed.
> 
> Also, I suspect that we won't accept the newly added test protos, but we can discuss that after we start our review process.

@tonyliaoss CLA has been passed! I cannot modify the `wait for user action` tag

fgarciacorona:
@tonyliaoss Is there any additional action required from my side to have the PR marked as safe for test and proceed with the testing of the PR? Additionally, I think it is missing the `C#` label.

JasonLunn:
Could you rebase this PR? Github thinks it is changing almost 800 files which makes review impractical.

fgarciacorona:
> Could you rebase this PR? Github thinks it is changing almost 800 files which makes review impractical.
> 
> Edit, nevermind about the rebase. I see what @tonyliaoss means about the test protos, though. Is there no other way to excite the issue that this PR aims to fix?

Right, there are 3 files that have actually been changed and the rest are half of them .proto and the other half the .cs generated classes from the protos.

I could only think of dynamically generating .proto files highly nested from within the testcase (if you want to reduce the number of existing files in the PR) and then somehow try to run `protoc` to generate the corresponding classes.

fgarciacorona:
@JasonLunn how could I have the repo marked as "safe for tests" without having to request it for every commit?

JasonLunn:
> @JasonLunn how could I have the repo marked as "safe for tests" without having to request it for every commit?

Humans have to apply the tag by policy for third party contributions to make sure they're safe to run on our CI infrastructure.

JasonLunn:
```
Failed FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString [263 ms]
  Error Message:
     Expected: 402
  But was:  419

  Stack Trace:
     at Google.Protobuf.Reflection.DescriptorsTest.FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString() in D:\a\protobuf\protobuf\csharp\src\Google.Protobuf.Test\Reflection\DescriptorsTest.cs:line 59

1)    at Google.Protobuf.Reflection.DescriptorsTest.FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString() in D:\a\protobuf\protobuf\csharp\src\Google.Protobuf.Test\Reflection\DescriptorsTest.cs:line 59
```

Interestingly, this only fails on the Windows test configuration, but it passed on Linux...

fgarciacorona:
> ```
> Failed FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString [263 ms]
>   Error Message:
>      Expected: 402
>   But was:  419
> 
>   Stack Trace:
>      at Google.Protobuf.Reflection.DescriptorsTest.FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString() in D:\a\protobuf\protobuf\csharp\src\Google.Protobuf.Test\Reflection\DescriptorsTest.cs:line 59
> 
> 1)    at Google.Protobuf.Reflection.DescriptorsTest.FileDescriptor_CreateMessageWithDeepDependencies_BuildFromByteString() in D:\a\protobuf\protobuf\csharp\src\Google.Protobuf.Test\Reflection\DescriptorsTest.cs:line 59
> ```
> 
> Interestingly, this only fails on the Windows test configuration, but it passed on Linux...

@JasonLunn the reason why it was not failing in linux and arm64 is because the metrics for benchmarking are only generated for Debug (windows) but not for Release (linux and arm64). 

Additionally we were storing the benchmark metrics in Debug mode in a static variable that was impacted by all unit tests. This needs to be reset before the specific test that looked at them.

fgarciacorona:
@JasonLunn would it be possible to remove the assigned reviewers and I will do a rebase and hopefully based on that the right reviewers will be assigned? I think because I didn't rebase it properly the first time a lot of commits from main triggered so many people.

fgarciacorona:
> It would expedite review if you could refactor the PR so that there are not ~800 files. I assume you have a script that generated the deeply nested test `.proto` files in the first place - could you use that in a `genrule` to generate the test input files at test time so that they don't have to be individually reviewed?

The script that we have anonymized our data, so the proto files are based on actual data. But thanks for the feedback, knowing that a generator could also be used, I will focus on that. 

Question though: do the generated `.cs` files need to be also committed? because I see them committed in the repo for other test `.proto` files, and that would eventually defeat the purpose (we would have ~400 only). Additionally if we don't commit the generated `.cs` files I am not sure whether it would be an accepted approach to have the test project not able to build if the test classes are not committed.

googleberg:
Anything that needs to be generated should be generated from a Bazel genrule rather than checked in.

fgarciacorona:
@googleberg as requested I have created a generator using Bazel genrule.

fgarciacorona:
@googleberg could we add the "safe for tests" tag so that I can check if anything else is missing in the branch?

fgarciacorona:
I could reproduce the problem locally and I have fixed it. The path for the output of the generated .pb.cs files did not exist. I have added these paths to .gitignore as well as they are dynamically generated. Let me know if this approach is good. 
@googleberg could you add the "safe for tests" tag again?

JasonLunn:
I suspect that the test failures would be fixed with a rebase.

fgarciacorona:
@JasonLunn no, unfortunately it was an issue on my side. The command `find` supports different parameters for linux and MacOS. 
I have used this commit to avoid the problematic parameter and simplify more the code. 

Could you mark them as `safe for tests`, let's see if we get them finally running now.

fgarciacorona:
@googleberg could we get the "safe for tests" tag after you review the changes? is it needed to have all the changes in one single commit for that?

fgarciacorona:
@googleberg I don't think that the failures on MacOs were related to my changes. I have rebased could you mark the commits as "safe for tests"?

JasonLunn (approved):

PR #20418, opened 2025-02-20
Python: Option to validate duplicate fields during json deserialization

sersh4nt:
## The issue

Proto:
```proto
syntax = "proto3";

package test;

message User {
  string first_name = 1 [json_name = "first_name"];
  string last_name = 2;
}
```

Code:
```python
from google.protobuf.json_format import ParseDict
import test_pb2

data = {"first_name": "Aaron", "last_name": "Surname", "lastName": "Fake"}

message = ParseDict(data, test_pb2.User())

print(message)
```
Output:
```
first_name: "Aaron"
last_name: "Fake"
```

## Why is this a problem?

This is a pretty unexpected behavior. Even using json_name cannot fully prevent one from using duplicate fields in json.

## What about implementations in other languages?

protobuf-go features a specific [check](https://github.com/protocolbuffers/protobuf-go/blob/259e665f26b1019a88c9ed6c7f16f01242838720/encoding/protojson/decode.go#L212) for duplicate fields. 
This PR is created similarly to [C# PR](https://github.com/protocolbuffers/protobuf/pull/20288)

## What does this PR do?

This PR improves checks for fields already used during JSON parsing. Back to the example, the result will be as follows:

Code:

```python
from google.protobuf.json_format import ParseDict
import test_pb2

data = {"first_name": "Aaron", "last_name": "Surname", "lastName": "Fake"}

message = ParseDict(data, test_pb2.User())

print(message)
```

Output:

```
  File "/Users/sersh4nt/Projects/protobuf/venv/lib/python3.12/site-packages/google/protobuf/json_format.py", line 597, in _ConvertFieldValuePair
    raise ParseError(
google.protobuf.json_format.ParseError: Message type "test.User" should not have multiple "lastName" fields at "User".
```

### DIFF
--- a/python/google/protobuf/internal/json_format_test.py
+++ b/python/google/protobuf/internal/json_format_test.py
@@ -1739,5 +1739,21 @@ class JsonFormatTest(JsonFormatBase):
     with self.assertRaises(json_format.ParseError):
       json_format.Parse(text, json_format_proto3_pb2.TestMessage())
 
+  def testParseDuplicateJsonFields(self):
+    data = json.dumps({"self_": 1, "self_underscore": 2})
+
+    with self.assertRaises():
+      json_format.Parse(
+          data, test_proto2_pb2.MessageWithSelfAndSelfUnderscoreField()
+      )
+
+  def testParseDictDuplicateJsonFields(self):
+    data = {"self_": 1, "self_underscore": 2}
+
+    with self.assertRaises():
+      json_format.ParseDict(
+          data, test_proto2_pb2.MessageWithSelfAndSelfUnderscoreField()
+      )
+
 if __name__ == '__main__':
   unittest.main()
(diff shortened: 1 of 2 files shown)


zhangskz:
This seems like the right idea, but unfortunately I think this would break users dependent on the current behavior.

We may want to land this stricter enforcement as an opt-in, and wait until a breaking change to enforce this by default. And same goes for other languages.

sersh4nt:
This looks legit, although it's a serious bug. Should I improve the validation myself? If so, how'd you like it to be? For now I can see 2 variants: first using some environment variable, and the second - using additional kwarg, which had to be passed to the Parse/ParseDict method

zhangskz:
Upon further consideration / discussion, it appears the JSON spec does not actually prohibit duplicate keys and we likely want to update protobuf to be consistent by loosening this check instead.

The conformant behavior for protobuf should actually be to accept duplicate keys and retain the last value (like with our binary format parsing). Note that in some languages, we seem to already do this today and it may not actually be feasible to reject those anyways if using standard JSON parsers.

We'll plan on updating https://protobuf.dev/programming-guides/json/ accordingly.

PR #20470, opened 2025-02-25
Remove reference to rules_rust from .bazelrc

acozzette:
I missed this when I removed the rules_rust dependency from MODULE.bazel in b8248f687f07d78a4baedd1581843883544ae239. We need to drop this flag because it's an error to reference it when rules_rust is not available.


zhangskz (approved):

zhangskz:
@googleberg FYI

PR #20531, opened 2025-02-28
Optimize varint decoding without intrinsics

davidzeng0:
This PR builds on https://github.com/protocolbuffers/protobuf/pull/10646 and https://github.com/protocolbuffers/protobuf/pull/13158 to provide a more optimized varint decoder.

This code does not use any intrinsics and is concise in order to fit in a single cache line for maximum efficiency.

See https://godbolt.org/z/rPvKePY9a for more details.

### Explanation

We read a uint32 and extracts relevant bits via bit manipulation, taking the cold path only if the varint is longer than 4 bytes. This integer size is picked because it strikes a reasonable balance bit manip overhead and branch probability. 

This gives us a range of `0..268,435,456` on the hot path, or `0..33,554,432` if it's tagged for a field.

### Benchmarking details

libprotobuf compiler: gcc 11.4
allocator: jemalloc
cpu: amd 5800x
linking: static
lto: no

input message: ~40kb, varies in message types, varints, and bytes
results: average of 10 runs, each run is 500,000 iterations
op: construct, parse from memory array, destruct
control: 23.2 microsecs/op
with changes: 22.4 microsecs/op

microbench: 1.7-2ns/call

About a 3-4% improvement overall. Not much but for servers which may parse millions of protobuf messages, it's not insignificant

### DIFF
--- a/src/google/protobuf/parse_context.h
+++ b/src/google/protobuf/parse_context.h
@@ -873,6 +873,16 @@ static const char* VarintParseSlowArm(const char* p, uint64_t* out,
 }
 #endif
 
+// Helper function for preventing inlining
+template <typename T>
+ABSL_ATTRIBUTE_COLD
+PROTOBUF_NOINLINE
+static const char* VarintParseCold(const char* p, T* out) {
+  auto ptr = reinterpret_cast<const uint8_t*>(p);
+
+  return VarintParseSlow(p, ptr[0], out);
+}
+
 // The caller must ensure that p points to at least 10 valid bytes.
 template <typename T>
 [[nodiscard]] const char* VarintParse(const char* p, T* out) {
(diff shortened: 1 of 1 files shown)


davidzeng0:
CLA form isn't working

davidzeng0:
I haven't looked much into the TailCallTable parsing mechanism much, but replacing some code in the non-fast table could also improve performance.

I'm not sure how feasible it is to replace the whole `ParseLoop` and `TagDispatch` mechanism with this as well, my understanding is that the fast table has to be somewhat small to be worth it

davidzeng0:
Sorry had to force push to change details about the commit author

davidzeng0:
Could I have someone else take a look at this? The process seems to be stalled.

davidzeng0:
Hello, @acozzette, I've seen you around the previous varint opt PRs so maybe you'd be willing to take look? Thanks in advance.

davidzeng0:
Woops, did something wrong in the rebase haha. Sorry everyone

mkruskal-google:
Looks like it's still not building, more merge issues?

src/google/protobuf/parse_context.h
```diff
@@ -929,15 +939,60 @@ template <typename T>
     return p + 2;
   }
   return VarintParseSlowArm(p, out, first8);
-#else   // __aarch64__
+#elif defined(__x86_64__) // __aarch64__
```
mkruskal-google, line 942: Is this optimization meant for x86_64 or arm64?  If it's x86_64, the build error is because you're using ValueBarrier, which we only define for arm64 builds.  If it's for arm64, you need to fix this line and probably call that out in the PR description.  We tend to be willing to accept a lot of ugly code for x86_64 optimization that deliver, but for arm64 the bar would likely be higher

src/google/protobuf/parse_context.h
```diff
@@ -929,15 +939,60 @@ template <typename T>
...
+  // Hot path: try to parse a 28 bit varint
+  // Range: 0..268,435,456
+  // Tagged: 0..33,554,432
+  //
+  // Messages longer than 128 bytes are not uncommon, so this
+  // should be faster in the general case
+
+  // Input is guaranteed atleast 10 bytes
+  uint32_t value = *reinterpret_cast<const uint32_t*>(p);
```
mkruskal-google, line 951: We're seeing crashes here from alignment issues, does using memcpy affect the results?

src/google/protobuf/parse_context.h
davidzeng0: Aha. This is an optimization for x86_64. I'll fix that soon.

src/google/protobuf/parse_context.h
davidzeng0: I couldn't find any good resources on the alignment requirements of C++ pointers. For x86_64, a memcpy should optimize to the same but I did notice `UnalignedLoad` having poor optimization on LLVM (though I could not remember if I chose a big endian target for this, but I opted out of everything but x86_64 since it's the only platform I've tested and which I have confidence in the results).

Are the crashes from the address sanitizer? Theres no reason why it should crash during normal runs.

mkruskal-google: I'm seeing SIGILLs from clang, with and without sanitizers.  From what I found this is strictly UB in C++ due to the alignment requirements of uint32_t

src/google/protobuf/parse_context.h
mkruskal-google: Cool, also note that this varint parser is not particularly load-bearing anymore.  ShiftMixParseVarint is our table-driven parser's implementation that's used for parsing most protobuf data on the wire.

davidzeng0: I'm aware. I tried to look into if this would he helpful for table parsing but concluded it wasn't. I was still surprised it managed an extra 4%.

A little unrelated but I was writing a protobuf library in Rust which uses this in a load-bearing manner :)

davidzeng0:
For the mean time I've moved the `ValueBarrier`s outside of the guards and added a memcpy. The `memcpy` adds an additional `mov` in godbolt which is suspicious but I don't have access to my work computer right now so I'm a little limited in what I can do.

I did remember to run the tests this time, they seem to have passed.

davidzeng0:
Oh great, thanks Jason. The integrations also have passed.

mkruskal-google:
Sorry for the delay, I'll see how this performs internally

PR #20549, opened 2025-03-03
Split maven dependencies into dev vs local

mkruskal-google:


.github/workflows/test_bazel.yml
```diff
@@ -39,6 +39,13 @@ jobs:
         with:
           ref: ${{ inputs.safe-checkout }}
 
+      - name: Pin to Java 11
```
zhangskz, line 42: Add comment indicating why we pinned this?

zhangskz (approved):

zhangskz (approved):

PR #20596, opened 2025-03-06
[PHP] - phpunit upgrade

FabioBatSilva:
Hi.

This is meant to reduce the noise when running php tests and make the test a bit more friendly.

There is a ton of [warnings](https://github.com/protocolbuffers/protobuf/actions/runs/13707746863/job/38337186415#step:8:1683) happening currently due to deprecation on phpunit..  

What is included :
* Upgrade to [phpunit 10.5
](https://phpunit.de/supported-versions.html)* Add generated files to git ignore
* Add phpunit.xml and bootstrap
* Remove Polyfill for phpunit6

I have a few other PRs coming for [issues]( https://github.com/protocolbuffers/protobuf/issues?q=is%3Aissue%20state%3Aopen%20author%3AFabioBatSilva) I have open. 
And this should make the contributing a bit easier for me other future contributors..


php/tests/PreviouslyGeneratedClassTest.php
```diff
@@ -12,7 +12,11 @@ class PreviouslyGeneratedClassTest extends TestBase
     public function testPrefixForReservedWords()
     {
         // In newer versions of PHP, we cannot reference the old class name.
-        if (version_compare(phpversion(), '8.1.0', '>=')) return;
+        if (version_compare(phpversion(), '8.1.0', '>=')) {
+          $this->markTestSkipped('PHP 8.1.0 and later do not allow reserved class name.');
+        };
```
bshaffer, line 17: nit: remove extra semicolon
```suggestion
        }
```

php/tests/PreviouslyGeneratedClassTest.php
```diff
@@ -12,7 +12,11 @@ class PreviouslyGeneratedClassTest extends TestBase
...
+
+        $this->assertTrue(class_exists('Previous\readonly'));
```
bshaffer, line 19: nit: if we add this here, we should probably just remove the `eval` below it

php/phpunit.xml
bshaffer, line 1: it's more standard to name this file `phpunit.xml.dist` for phpunit config files which are version controlled.

php/tests/PreviouslyGeneratedClassTest.php
```diff
@@ -12,7 +12,11 @@ class PreviouslyGeneratedClassTest extends TestBase
     public function testPrefixForReservedWords()
     {
         // In newer versions of PHP, we cannot reference the old class name.
-        if (version_compare(phpversion(), '8.1.0', '>=')) return;
+        if (version_compare(phpversion(), '8.1.0', '>=')) {
```
bshaffer, line 15: If you look at composer, lower versions of PHP are not supported (the minumum supported version is PHP 8.1). So why not just get rid of this statement altogether?

bshaffer (commented):
This looks good but why not upgrade the version of PHPUnit instead of having to suppress the warnings? I think at this point we can just make the minimum phpunit version be `^9.6`, since lower versions of PHP are not supported

php/tests/PreviouslyGeneratedClassTest.php
FabioBatSilva: Great point..

I think that means we can also upgrade to phpunit 10.

googleberg:
Test failures look like this needs a rebase and retest.

FabioBatSilva:
> Test failures look like this needs a rebase and retest.

Done

```bash
composer test
# ...
PHPUnit 10.5.58 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.33
Configuration: /home/fabio/backup/workspace/cpp/protobuf/php/phpunit.xml.dist

D..............................................................  63 / 462 ( 13%)
............................................................... 126 / 462 ( 27%)
............................................................... 189 / 462 ( 40%)
............................................................... 252 / 462 ( 54%)
............................................................... 315 / 462 ( 68%)
............................................................... 378 / 462 ( 81%)
............................................................... 441 / 462 ( 95%)
.....................                                           462 / 462 (100%)

Time: 00:00.229, Memory: 16.00 MB

OK, but there were issues!
Tests: 462, Assertions: 2879, PHPUnit Deprecations: 13.
```

mpesari:
Hi,

Would be nice to have this merged and add CI runs for PHP 8.4 and 8.5.

There are still PHPUnit deprecations (`phpunit --display-phpunit-deprecations`):

```
14 tests triggered 14 PHPUnit deprecations:                                                                                                                                                                

1) EncodeDecodeTest::testWrapperJsonDecodeAndGet                                                     
Data Provider method EncodeDecodeTest::wrappersDataProvider() is not static
...
```

JasonLunn:
@FabioBatSilva - Thank you for contribution and your effort to help get ahead of this issue. I believe the intent of this PR has been addressed another PR (#25531) unintentionally. That was then subsequently backported, so I think this PR can be closed.

Issue #20036, opened 2025-01-18
PHP Extension - Missing Google\Protobuf\Value:has* Methods


FabioBatSilva:
The PHP C extension implementation is missing `has*` methods for the `Google\Protobuf\Value` class that are present in the native PHP implementation. 

This creates inconsistency between the two:

PHP Implementation :
* [hasNullValue](https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/src/Google/Protobuf/Value.php#L61
)
* [hasNumberValue](https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/src/Google/Protobuf/Value.php#L92)
* [hasStringValue](https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/src/Google/Protobuf/Value.php#L123)
* [hasBoolValue](https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/src/Google/Protobuf/Value.php#L154)
* [hasStructValue](https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/src/Google/Protobuf/Value.php#L185)
* [hasListValue](https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/src/Google/Protobuf/Value.php#L216)

C Implementation
https://github.com/protocolbuffers/protobuf/blob/b51d59e193d4301d7e102ef0509787e7a89519f1/php/ext/google/protobuf/wkt.inc#L1356-L1371

----
**What version of protobuf and what language are you using?**
Version: main
Language: php
OS: Linux

**What did you do?**
```php
<?php

require_once (dirname(__DIR__) . '/vendor/autoload.php');

use Google\Protobuf\Value;

$v = new Value();

var_dump($v->hasNullValue());
```

```bash
# Compile the extension 
./php/tests/compile_extension.sh

# Run script
php -d display_errors=on -dextension=../ext/google/protobuf/modules/protobuf.so has_value_test.php
```

**What did you expect to see**
```
bool(false)
```

**What did you see instead?**
```
Uncaught Error: Call to undefined method Google\Protobuf\Value::hasNullValue() 
```

JasonLunn:
[...]

PR #20636, opened 2025-03-10
[php][compiler][ext] - Add missing has* methods and other well known types

FabioBatSilva:
Should fix : https://github.com/protocolbuffers/protobuf/issues/20036

### DIFF
--- a/php/ext/google/protobuf/wkt.inc
+++ b/php/ext/google/protobuf/wkt.inc
@@ -319,6 +319,13 @@ static PHP_METHOD(google_protobuf_Api, setSourceContext) {
   RETURN_COPY(getThis());
 }
 
+static PHP_METHOD(google_protobuf_Api, hasSourceContext) {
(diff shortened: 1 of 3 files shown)


php/ext/google/protobuf/wkt.inc
```diff
@@ -319,6 +319,13 @@ static PHP_METHOD(google_protobuf_Api, setSourceContext) {
   RETURN_COPY(getThis());
 }
 
+static PHP_METHOD(google_protobuf_Api, hasSourceContext) {
```
FabioBatSilva, line 322: `php/ext/google/protobuf/wkt.inc` was generated with `bazel build //php:gen_wkt_sources`.

Unsure why this file is in git..

haberman: This will be automatically updated after the PR is submitted, so you could leave it out of the PR if you wanted. But there's also no harm in updating it eagerly, as you have.

haberman (approved):

PR #20644, opened 2025-03-11
fix(cmake): abseil and protobuf msvc static runtime must have the same value

### LINKED ISSUE #20355: Incompatible MSVC runtime linking between Abseil dependency and Protobuf
Version: v30.0-rc1
Language: C++

Windows 11

Visual Studio 2022 / MSVC 19.43.34808 / Cmake 3.31.1

**What did you do?**
1. Cloned protobuf from repo (tag v30.0-rc1)
2. Opened protobuf in Visual Studio as folder (i.e. using VS cmake integration)
3. Set in CMakePresets.json: "CMAKE_CXX_STANDARD": "17", "protobuf_MSVC_STATIC_RUNTIME":  false
4. Built protobuf (and installed). The Abseil dependency is automatically downloaded and built 

Abseil and consecutively, protobuf built with dynamically linked MSVC runtime as per protobuf_MSVC_STATIC_RUNTIME flag

**What did you see instead?**
Abseil is built with static MSVC runtime linking, Protobuf  is build with dynamic MSVC linking. Build fails

The problem is caused by protobuf/CMakeLists.txt lines 202-212, specifically 206: set(ABSL_MSVC_STATIC_RUNTIME ON)
```cmake
if (protobuf_BUILD_SHARED_LIBS)
  set(protobuf_SHARED_OR_STATIC "SHARED")
else (protobuf_BUILD_SHARED_LIBS)
  set(protobuf_SHARED_OR_STATIC "STATIC")
  set(ABSL_MSVC_STATIC_RUNTIME ON)
  if (protobuf_MSVC_STATIC_RUNTIME)
      set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded$<$<CONFIG:Debug>:Debug>)
  else()
      set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreaded$<$<CONFIG:Debug>:Debug>DLL)
  endif()
endif (protobuf_BUILD_SHARED_LIBS)
```
The logic sets ABSL to use static runtime linking when protobuf is configured to produce a static lib. Those are two different things though, and which kind of a lib is produced shouldn't affect how the MSVC runtime is linked to them.

A simple fix for me to move forward was to fix set ABSL_MSVC_STATIC_RUNTIME to OFF instead since I don't care about static runtime linking. A proper fix would probably be to set ABSL_MSVC_STATIC_RUNTIME only based on protobuf_MSVC_STATIC_RUNTIME, regardless of whether the output is dynamic or static.

carlocorradini:
Fix #20355


acozzette (approved):

acozzette:
@carlocorradini Thanks for fixing this!

Issue #19978, opened 2025-01-13
PHP Extension - Segmentation fault if message constructor is not called.


FabioBatSilva:
We are trying to migrate an application from the PHP Protobuf implementation to the C extension.

A few tests are using the [prophecy](https://github.com/phpspec/prophecy) to mock objects during tests.
and prophecy ends up creating a proxy that looks something like this :

```php
class P5
    extends \UserProfile\V1alpha\Account
    implements \Prophecy\Doubler\DoubleInterface, \Prophecy\Prophecy\ProphecySubjectInterface, \Prophecy\Doubler\Generator\ReflectionInterface
{
    private $objectProphecyClosure;

    public  function __construct( $data = NULL) {
        if (0 < func_num_args()) {
            call_user_func_array(array('parent', '__construct'), func_get_args());
        }
    }

    public  function setProphecy(\Prophecy\Prophecy\ProphecyInterface $prophecy) {
        if (null === $this->objectProphecyClosure) {
            $this->objectProphecyClosure = static function () use ($prophecy) {
                return $prophecy;
            };
        }
    }

    public  function getProphecy() {
        return \call_user_func($this->objectProphecyClosure);
    }
}
```

Since the parent constructor is not always called it seems the message never properly [initialized](https://github.com/protocolbuffers/protobuf/blob/98aab04c0bf846ad6b4af339d9dbc5cf77d6c03c/php/ext/google/protobuf/message.c#L568) and the field [lookup](https://github.com/protocolbuffers/protobuf/blob/98aab04c0bf846ad6b4af339d9dbc5cf77d6c03c/php/ext/google/protobuf/message.c#L348-L349) fails when [get_field](https://github.com/protocolbuffers/protobuf/blob/98aab04c0bf846ad6b4af339d9dbc5cf77d6c03c/php/ext/google/protobuf/message.c#L91) is called.

> **_NOTE:_**  This is a common pattern in many php frameworks.
> And it is debatable weather or not this should be mocked/proxied but the behaviour is inconsistent between the two implementations and I believe no condition should trigger a seg fault. 

----

**What version of protobuf and what language are you using?**
Version: main
Language: php
OS: Linux

**What did you do?**
```php
<?php
# proxy_test.php

require_once('../vendor/autoload.php');
require_once('test_util.php');

use Foo\TestMessage;

class TestMessageProxy extends TestMessage
{
    private $foo;

    public  function __construct($data = NULL) {
        if (0 < func_num_args()) {
[...]

PR #20649, opened 2025-03-11
[php][ext] - Fix isset behaviour

### LINKED ISSUE #20648: PHP Extension - Isset throws an exception
The PHP C extension implementation of `isset` throws an exception.

This happens BC the `has_property` and other handlers all use the same `get_field` function.

https://github.com/ChessCom/protobuf/blob/d62123ed8cbf47d2e5de328d77e62b994d66e2dc/php/ext/google/protobuf/message.c#L97
https://github.com/ChessCom/protobuf/blob/d62123ed8cbf47d2e5de328d77e62b994d66e2dc/php/ext/google/protobuf/message.c#L243

----
Version: main
Language: php
OS: Linux

**What did you do?**
```php
<?php
# isset_test.php

require_once (dirname(__DIR__) . '/vendor/autoload.php');

use Foo\TestMessage;
 
$message = new TestMessage();
 
var_dump(isset($message->foo));
```

```bash
# Compile the extension 
./php/tests/compile_extension.sh

# Run script
php -d display_errors=on -dextension=../ext/google/protobuf/modules/protobuf.so isset_test.php
```

```
bool(false)
```

**What did you see instead?**
```
Uncaught Error: No such property foo.
```

FabioBatSilva:
Should fix : https://github.com/protocolbuffers/protobuf/issues/20648


php/ext/google/protobuf/message.c
```diff
@@ -250,15 +268,11 @@ static int Message_compare_objects(zval* m1, zval* m2) {
...
+  const upb_FieldDef* f = lookup_field(intern, member);
 
   if (!f) return 0;
 
-  if (!upb_FieldDef_HasPresence(f)) {
-    zend_throw_exception_ex(
-        NULL, 0,
-        "Cannot call isset() on field %s which does not have presence.",
-        upb_FieldDef_Name(f));
+  if (upb_FieldDef_IsOptional(f) && !upb_FieldDef_HasPresence(f)) {
```
bshaffer, line 275: Any particular reason why you're removing the error message above? 

I am not sure if it's appropriate or not, but I'd rather not change behavior unless we have a specific reason for doing so (and this isn't mentioned in issue #20648)

php/tests/GeneratedClassTest.php
```diff
@@ -56,6 +56,31 @@ public function testSetterGetter()
...
+        $message->setOptionalString('bar');
+
+        $this->assertTrue(property_exists(TestMessage::class, 'optional_string'));
+        $this->assertFalse(property_exists(TestMessage::class, 'foo'));
+
+        $this->assertTrue(property_exists($message, 'optional_string'));
+        $this->assertFalse(property_exists($message, 'foo'));
+
+        // isset() should return false for attributes that are not public.
+        $this->assertFalse(isset($message->optional_string));
```
bshaffer, line 80: We are already testing this above, no reason to do it here as well

php/ext/google/protobuf/message.c
FabioBatSilva: Sorry I only notice that one while implementing the fix / test. 
I can open another issue if you think that makes sense.  but without it isset is still unusable in many scenarios..

That will prevent using `isset($message->optional_string)` on a existing field. 
```
1) GeneratedClassTest::testIsset
Exception: Cannot call isset() on field optional_string which does not have presence.

protobuf/php/tests/GeneratedClassTest.php:69
```

I don't think `isset` should ever throw..

bshaffer: Hmm that's a good point. There's nowhere in PHP where calling `isset` should throw an exception.

haberman:
Can you give a bit more info about why your code is calling `isset()` on these fields?

The member variables of proto generated classes are private to the class, and should not be accessed by user code.

bshaffer:
@haberman I think that since it's valid in PHP to call `isset` on objects, the behavior should be consistent when calling `isset` on protobuf messages. So rather than throwing an exception, it should just return `false`. 

It also seems like the behavior is inconsistent between the extension and the native library. So if we want to fix the inconsistency, while we _could_ make the native `Message` class throw an exception using the magic [`__isset`](https://www.php.net/manual/en/language.oop5.overloading.php#object.isset) method, it would be better to update the extension to be more consistent with PHP behavior.

FabioBatSilva:
> Can you give a bit more info about why your code is calling `isset()` on these fields?

I first run into this in some scenario where the proto message is used in a [Twig](https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L1785
) template.

Don't think the behaviour here was intentional it was most likely copy and pasta from one of the object property handlers.
As I mentioned earlier can't think of a scenario in PHP where you would expect `isset` to throw an exception.

FabioBatSilva:
PR is still relevant, Waiting for the maintainers.

bshaffer:
+1 this is still relevant. And if we include this, we should try to get it in for the next major release of Protobuf 

cc @haberman

esrauchg:
Can you clarify this a bit more?

The fields are protected, so outside usage isset() should always return false right? Why would it even execute this path at all to decide what we think it should be?

So how does it makes sense; does PHP actually run this path on private fields and then afterwards check if it was private to discard the value and instead give false instead? Or is it intended that overrides are expected to somehow do their own visibility check logic to decide whether the caller actually should get the truth or else "false only because you don't have visibliity"?

In general I think the PR behavior is problematic because of the behavior that it will give to callers which do have visibility to the variables:

- I think If anything for these fields isset() as visible internally if we want this to be correct it needs to always return _true_ not false; the category of fields you are affecting are exactly "non-nullable fields", so it returning false is always wrong if someone does have visibility to them.
 
- The throwing behavior is only defensible because we use isset() in important cases on the nullable strings and having the throw will help us catch programmer errors where we incorrectly did use it on the non-nullable fields (because it would never make sense for us to emit gencode that did it, and no one besides our gencode should have visibility onto them).

IMO it does seems like we should be able to make the choice of this throwing if our gencode was bugged here, without it being a behavior which is visible to any of the users of Protobuf.

FabioBatSilva:
> The fields are protected, so outside usage isset() should always return false right? Why would it even execute this path at all to decide what we think it should be?
> 

I"m not familiar enough with PHP internals to answer the why, but `Message_has_property` does get invoked given the 
exception.

> * I think If anything for arguably for these fields isset() internally should always return _true_ not always return false; the category of fields you are affecting are exactly "non-nullable fields", so it returning false is always wrong if someone does have visibility to them.

Not sure if i understand the suggestion,  if we always return _true_ this would break:
```php
if (isset($message->field_that_doesnt_exist)) {
   print($message->field_that_doesnt_exist);
}
```

> * The throwing behavior is only defensible because we use isset() in important cases on the nullable strings and having the throw will help us catch programmer errors where we incorrectly did use it on the non-nullable fields (because it would never make sense for us to emit gencode that did it, and no one besides our gencode should have visibility onto them).

This is not consistent with how PHP behaves. I don't think there is a case in PHP language where isset should throw.
To the best of my knowledge It only return true/false. 
No developer expects `isset` to throw an exception.

> IMO it does seems like we should be able to make the choice of this throwing if our gencode was bugged here, without it being a behavior which is visible to any of the users of Protobuf.

My goal here is mostly to make the PHP native and C ext versions behave the same. As you can see in the test scenarios.
Most of the choices were already made by the PHP core team in the language design.

Throwing an exception makes the behaviour incompatible with the PHP native version and also breaks the expectations for how `isset` behaves in PHP.

esrauchg:
Lets take as a given that the users of protobuf must never be able to observe an exception when writting isset for a minute here and take it as a followup.

It seems that to match the pure-PHP behavior it must be able to honor the visibility distinction here; the observable pure-PHP behavior is that the same method called with the same value does not return the same value depending on the caller. And if I'm understanding correctly the proposed PR won't match that behavior in both cases because:

The pure-PHP behavior must be:

- Any calls to will always return false from outside the class.
- Any calls to will always return true on implicit presence fields (implicit presence = normal string and int32 fields in proto3)
- Any calls to this will return true IFF the field is set on explicit presence fields (explicit presence = optional fields, message-typed fields, fields in oneofs).

If this method is running on the upb-backed one, for the symptom you are seeing I believe the behavior is actually this:

- Calls both inside and outside the class will throw if called on an implicit presence field
- Calls both inside and outside the class will return true IFF the field is explicit presence and set.

With your proposed PR I believe the change is then:

- Calls both inisde and outside the class will return false if its an implicit presence field
- Calls both inside and outside the claass will return true IFF the field is explicit presence and set.


- In Implicit presence fields, the external behavior will match (always false), but the internal behavior will not (should be always true).
- In explicit presence fields, the observed external behavior will not match because the upb one must still return 'true' some of the time, where the current one never will return true.

esrauchg:
I guess to clarify the problem to be more concrete case, you gave the example of something shaped like this:

```
if (isset($message->field_that_doesnt_exist)) {
   print($message->field_that_doesnt_exist);
}
```

It seems like given the symptom, this example must 'break' only on upb? (both before and after the proposed PR):

`message SomeMsg { optional string does_exist = 1; }`

```
$message->set_optional_does_exist("123")
if (isset($message->does_exist)) {  // return true unrelated to visibility, only on upb backend
   print($message->does_exist);  // illegal attempt at protected field access on all backends
}
```

And (assuming thats true) I believe we also can't fix by just unilaterally changing isset() to always return false, as the gencode itself does use uses `isset($message->does_exist)` and relies on that giving sometimes-true answer, and we can't invalidate that old gencode without it being an enormous breaking change.

FabioBatSilva:
> And (assuming thats true) I believe we also can't fix by just unilaterally changing isset() to always return false, as the gencode itself does use uses `isset($message->does_exist)` and relies on that giving sometimes-true answer, and we can't invalidate that old gencode without it being an enormous breaking change.

I believe the tests cover that scenario:
https://github.com/ChessCom/protobuf/blob/ca205390c9ac2797810a0a1f58d5ca1485323664/php/tests/GeneratedClassTest.php#L69

For the scenarios I could think of the behaviour is now consistent for both PHP native and C ext.

Anything else you think we should test ?

esrauchg:
I'm still kind of struggling to understand how the behavior you're observing could make sense.

On the test line that is checking this:
`$this->assertFalse(isset($message->optional_string));`

If isset is running the 'Message_has_property', then I think this case is effectively be confirming that it doesn't matter if the function returns true or false for what value is observed, because Message_has_property will definitely return true for that case.

"The return value doesn't matter in any way, but whether it throws does matter" seems plausible if the implementation assumed zend has_property() is idempotent and calls it first and then ends up checking visibility _after_ calling it, but it would be really pretty weird if thats true. Definitely the code here believes that these functions shouldn't ever be invoked if the caller doesn't have visibility, per the comment that says "Note that all properties of generated messages are private, so this should only be possible to invoke from generated code", which is also on every other method too.

If that is the case, then it does get back to my prior "if this isn't going to throw it at least should return true, not false on implicit presence fields" feedback. It apparently doesn't matter to outside calls if we return true or false, but it does matter to internal calls.

I think to observe the "internal" calls behavior would need a test helper type that that looks like this:

```
class SubclassToObserveProtected extends TestMessage {
    public function InternalOptionalStringIsSet() { return isset($this->optional_string); }
    public function InternalRepeatedInt32IsSet() { return isset($this->repeated_int32); }
    public function InternalFooIsSet() { return isset($this->foo); }
}
```

I think if the above explanation is right then in pure php protobuf this test would pass:
```
$message = new SubclassToObserveProtected();

// Explicit presence fields begin nullable but become non-null when set.
$this->assertFalse($message->InternalOptionalStringIsSet());
$message->SetOptionalString("hello");
$this->assertTrue($message->InternalOptionalStringIsSet());

// Implicit presence fields are always there and never null, so isset() always returns true even if you haven't touched it.
$this->assertTrue($message->InternalRepeatedInt32IsSet());

// Fields that don't exist at all return false.
$this->assertFalse($message->InternalFooIsSet());
```

And with tip-of-main upb the behavior would be that `InternalRepeatedInt32IsSet` throws instead, and with post-your-PR change it would be that `InternalRepeatedInt32IsSet` returns false instead.

FabioBatSilva:
I think the extends scenario is not possible.
Extending the message should trigger `Couldn't find descriptor. Note only generated code may derive from \Google\Protobuf\Internal\Message` [here](https://github.com/protocolbuffers/protobuf/blob/main/php/ext/google/protobuf/message.c#L561-L567) and bypassing the constructor triggers a segfault as described in https://github.com/protocolbuffers/protobuf/issues/19978

There is also fair amount of assertions around `->has{Field}()` in [EncodeDecodeTest](https://github.com/protocolbuffers/protobuf/blob/main/php/tests/EncodeDecodeTest.php) witch invokes isset. 
So the most common paths are already covered by the existing tests.

For what it's worth I'v been running this and the other patches/PRs I have for a while now.

esrauchg:
I'm not sure how to implement the test, since extending gencode is not supported; we definitely do need to not prevent the parent ctor from running in that case.

Taking another step here though, if the premise about the behavior above is correct then all of the tests that you have would also pass if you change L265 to return 1 not return 0, as in:

```
if (!upb_FieldDef_HasPresence(f)) {
    return 1;
}
```

Can you confirm that is the case?

Thanks for iterating on this, I know its annoying for the PR to be held up, I just want to make sure we're not actually making things broken in a way that will be another footgun in the future.

FabioBatSilva:
No worries, Appreciate the attention to details.. 

Changing to `1` breaks [this](https://github.com/ChessCom/protobuf/blob/php-isset-fix/php/tests/GeneratedClassTest.php#L68-L69)


Patch:
```diff
diff --git a/php/ext/google/protobuf/message.c b/php/ext/google/protobuf/message.c
index 3c063bce5..fed0432a8 100644
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -263,7 +263,7 @@ static int Message_has_property(zend_object* obj, zend_string* member,
   if (!f) return 0;

   if (!upb_FieldDef_HasPresence(f)) {
-    return 0;
+    return 1;
   }
```

Tests Result:
```
1) GeneratedClassTest::testIsset
Failed asserting that true is false.

/workspace/cpp/protobuf/php/tests/GeneratedClassTest.php:69
```

esrauchg:
Ok, I believe I understand whats going on now!

One big confusion from my previous replies is that actually 'optional_string' and its not marked optional, this one is an implicit-presence field not an explicit-presence field:
https://github.com/ChessCom/protobuf/blob/php-isset-fix/php/tests/proto/test.proto#L29

Extremely confusingly this made it so fields 69 and 71 of the test were actually the same case (both implicit presence). Lower in the test proto it has fields prefixed *true_optional_* which are the ones marked optional.

I believe whats happening is more simply that no visibility enforcement is happening here, which makes sense in the lens that this method should be exactly like __isset() which is called on anything, which you can confirm if you add this test case (will pass on pure-php but fail on upb):

```
$message->setTrueOptionalString("hello");
$this->assertFalse(isset($message->true_optional_string));
```

Then we may have a bigger problem that upb effectively has no private visibility here, as in I have to believe that this test will also pass on pure-php and fails on upb:

```
$message->setTrueOptionalString('hello');
unset($message->true_optional_string);
$this->assertSame('hello', $message->getTrueOptionalString());
```

Will pass in php (because of visibility restriction behavior) and they will fail on php-upb both with and without your PR (because its not having any visibility behavior).

It seems like there _should_ be some way for us to implement visibility behavior to actually match pure-PHP behaviors, but I'm not a PHP expert and looking at __isset() for this I wasn't immediately able to find anything on this topic.

esrauchg:
It appears that _maybe_ zend_get_executed_scope() is the function that can be used to implement this check, where every method in this C extension that is on the properties 'should' actually explicitly look at that and handle it as a non-existing field if the scope isn't the right type (the has_property would just always return 0 if not).

It's not immediately obvious to me how trivial such a check is and if there is a performance implication though.

FabioBatSilva:
> It seems like there _should_ be some way for us to implement visibility behavior to actually match pure-PHP behaviors, but I'm not a PHP expert and looking at __isset() for this I wasn't immediately able to find anything on this topic.

Same here, But I suspect the extension is relying on some unusual behaviours that might influence how that works..
https://github.com/protocolbuffers/protobuf/blob/d53c7aa64e82bba6c8656caa8105c14ece4007a2/php/ext/google/protobuf/message.c#L41-L54

esrauchg:
It does look to me like something using zend_get_executed_scope() looks like the 'right' fix here to be able to have all of these properties behave as though they were private instead of public.

FabioBatSilva:
Yes, not very familiar with it that seems like the only alternative given the ext bypasses the default zend engine mechanism.
I suspect that will have some performance side effects given the  additional function call overhead for the scope resolution.


And properly enforcing the scope I believe will also break this :
```php
<?php
$msg = new TestMessage();

$msg->setOptionalString("Hello, World!");

echo $msg->optional_string . PHP_EOL;

# Hello, World!
```
Don't think the behaviour is intentional but the extension currently allows it.


I'd prefer to keep the changes here to minimal if possible.

esrauchg:
Sorry, if I'm understanding the state correctly I really don't think we can accept this patch in the mixed state, you're providing this example:

```
<?php
$msg = new TestMessage();

$msg->setOptionalString("Hello, World!");

echo $msg->optional_string . PHP_EOL;

# Hello, World!
```

This snippet must only work in Php-upb exactly because we failed visibility checks on the getter; in pure-PHP this would fail because $msg->optional_string is protected. And then it looks to me that the proposed PR would make isset($msg->optional_string) be false here though `$msg->optional_string` is a readable and non-null-value (actually its _always_ going to be readable and non-null, so isset() compared to the getter value will always be wrong).

Basically the state here seems to be "Pure-PHP-impl has proper private visibility, PHP-upb-impl the properties are all public, isset() on most of them will return true or false matching the return value, except implicit presence fields it throws".

Having PHP-upb be self-inconsistent in behavior seems like it would worse to me than the properties behaving as public (contrasted with pure-PHP where they are properly protected)

I'd be interested in what @bshaffer and maybe @haberman think, but IMO probably the decision here is one of these:

1) Not touch this until we can get the patch that fixes the fields to be private

2) Accept that the fields are actually public as an unintended behavior and divergence from pure-PHP. We can 'fix' the fact that isset() can throw by accepting the fields are public, then the proper would be to switch to return 1 not return 0 or throw on that path (because those fields are actually readable and non-nullable, hardcoded 1 is the same behavior you would get on pure-php if the fields were public instead of protected):

```
if (!upb_FieldDef_HasPresence(f)) {
    return 1;
 }
```

FabioBatSilva:
> Not touch this until we can get the patch that fixes the fields to be private.

I Hope not, The throw makes isset very much unusable.

haberman:
> I Hope not, The throw makes isset very much unusable.

We do not want to support `isset()` for the internal members of generated classes, because these members are intended to be private.

It looks like `isset()` for private variables always returns false; we should do the same for all of our fields:

```php
<?php
class MyClass {
    private $privateVariable = 'I am private';
}

$obj = new MyClass();
$set = isset($obj->privateVariable);
var_dump($set);
?>
```

```
$ php test.php 
bool(false)
```

esrauchg:
Yeah, really it seems to me we just need to get the patch for lookup_field to do whatever spelling of this that enforces the intended scope:

```
if (Z_OBJCE_P(object) != zend_get_executed_scope()) { return 0; }
```

And probably all of the property has/get/set actually should be guarded by checks like this, since right now if they are actually ~all leaking on the ext case only.

JasonLunn:
@FabioBatSilva - I noticed this PR has develoepd merge conflicts - can you resolve them and tag me when done, and I can kick off tests?

FabioBatSilva:
@JasonLunn done.

esrauchg:
Unfortunately I think the latest version of your PR would still make the C-Ext one and the pure-PHP not behave the same.

I think this version is the one that has the correct semantics of aligning upb onto the pure-PHP observable semantics:
https://github.com/protocolbuffers/protobuf/pull/27614/changes

But we will need to discuss if we can ship this change without it being a breaking change here: notably someone may succesfully using isset() and unset() on the C-Extension one on explicit-presence fields and making those correctly no-op (as they are on pure-PHP) will actually silently make their code wrong.

One middle ground option we have is to align the behavior only on implicit presence fields, but hold the fix for explicit-presence fields until the next breaking change release: since the implicit-presence case actively throws on the C-ext it seems to be a much lower risk that aligning it to silent-noop instead of throw will break someone compared to the explicit-presence case where it actually is doing something and not-throwing.

FabioBatSilva:
> Unfortunately I think the latest version of your PR would still make the C-Ext one and the pure-PHP not behave the same.

@esrauchg can you clarify where you think that is happening ?
The only intent behavior changes are in the path where the extension segfaults today.

[This](https://github.com/protocolbuffers/protobuf/pull/20649#issuecomment-3372489863) was happening on main last time I checked.

I agree aligning the semantics makes the most sense but i'm not convinced #27614 is the right direction.
PHP extensions api already provide a mechanism for achieving that behavior and the extension currently [bypasses](https://github.com/protocolbuffers/protobuf/pull/20649#issuecomment-3285966623) the mechanism for reasons i don't fully understand. That not something I'm willing to tackle now.

bshaffer:
In testing https://github.com/protocolbuffers/protobuf/pull/27614, it seems that these two assertions would still fail for the c-extension (while passing for the native library, as expected):

```php
$message = new TestMessage();
$this->assertFalse(isset($message->foo));
$this->assertFalse(property_exists($message, 'foo'));
```

I agree with @FabioBatSilva that the expected behavior is for the above conditions to pass. @esrauchg could you explain more what your concerns are?

> the latest version of your PR would still make the C-Ext one and the pure-PHP not behave the same.

From what I can tell, these changes will make the C-Ext behave more consistently with the pure-php library. What will be different?

> someone may succesfully using isset() and unset() on the C-Extension one on explicit-presence fields and making those correctly no-op (as they are on pure-PHP) will actually silently make their code wrong.

So is the concern that if someone had code like this, it would break with this change?

```php
public function has_field(Message $m, $fieldName): bool
{
    try {
        isset($m->$fieldName);
        return true;
    } catch (Exception $e) {
        return false;
    }
}
```

This seems like an edge case (and poorly written code at that). And since PHP c-extensions typically do not follow semantic-versioning as strictly as native libraries, I am personally not concerned with this. However, if we feel the need to be extra safe, I am happy to wait for the next major version to include this change.

esrauchg:
> PHP extensions api already provide a mechanism for achieving that behavior and the extension currently https://github.com/protocolbuffers/protobuf/pull/20649#issuecomment-3285966623 the mechanism for reasons i don't fully understand

I looked closer and really the root cause is we have our preexisting gencode which has a property for every field in the .proto: we're reusing gencode which reads/isset() on them and expects it to work, but if the C-Ext case ever _actually_ reads that prop it'll always be wrong because we never modify those props in the C-Ext case at all (instead its all a parallel storage situation and the reads/sets/isset() are all intercepted to go to that instead).

I think if we were willing to have different gencode for the C Ext one versus the pure-PHP one the whole topic would be different: we wouldn't have to try to make it so props are never used in any way despite the fact that they do exist.

The usecase "I am trying to swap out the superclass to be a C Ext and maintaining compatibility with subclasses, those subclasses must call isset() and get the right value but external reads must not" does like a fringe situation, so it doesn't seem weird the vanilla C Ext support affordances likely just don't do what we want.

# Behavior audit

For this message:

```
syntax = "proto3";
message TestMessage {
   int32 implicit = 1;
   optional int32 explicit = 2;
}
```

This is what we're evaluating
```
$m = new TestMessage();
isset($m->implicit);
isset($m->explicit)

property_exists($m, 'implicit');
property_exists($m, 'explicit');

$m->set_implicit(123);
$m->set_explicit(123);

isset($m->implicit);
isset($m->explicit);

property_exists($m, 'implicit');
property_exists($m, 'explicit');

try { unset($m->implicit) } catch {... }
try { unset($m->explicit) } catch {... }

(And then if it didn't throw, what value does get_* return)
```

### Pure-PHP:

Here's the ground truth behavior of how protected fields are 'supposed' to be exposed to outside readers:

| Field Type           | Before isset()  | Before prop_ex  | After isset()   | After prop_ex   | unset() Throws?           | val after unset      |
|----------------------|-----------------|-----------------|-----------------|-----------------|---------------------------|----------------------|
| implicit       | false           | true            | false           | true            | Throws: Error             | N/A (Threw)          |
| explicit  | false           | true            | false           | true            | Throws: Error             | N/A (Threw)          |

(Not in the table, but we do **strongly** rely on at least the 'After => isset(explicit) == true' being a differing outcome internal and external. By contrast, `After => isset(implicit)` from subclasses should never be reached, it'll be a programmer error on our part if our gencode did).

### C-Ext at tip of main

| Field Type           | Before isset()  | Before prop_ex  | After isset()   | After prop_ex   | unset() Throws?           | val after unset      |
|----------------------|-----------------|-----------------|-----------------|-----------------|---------------------------|----------------------|
| implicit       | Throws: Exception | true            | Throws: Exception | true            | Throws: Exception         | N/A (Threw)          |
| explicit  | false           | true            | true            | true            | No                        | 0                    |

The biggest 'breaking change' concern here if we waved a magic wand and made this table the same as the first table is that isset() and unset() actually just fully work on explict presence fields: it's not too crazy to imagine someone is using them instead of has_x() and clear_x() and not realizing it.

### C-Ext with my draft PR that I linked:

| Field Type           | Before isset()  | Before prop_ex  | After isset()   | After prop_ex   | unset() Throws?           | val after unset      |
|----------------------|-----------------|-----------------|-----------------|-----------------|---------------------------|----------------------|
| implicit       | false           | true            | false           | true            | No                        | 123                  |
| explicit  | false           | true            | false           | true            | No                        | 123                  |

This looks to me that the first 4 columns would perfectly match pure-PHP. There's a lingering behavior mismatch that unset() does not throw and instead it does nothing.

### C-Ext with Fabio draft PR

| Field Type           | Before isset()  | Before prop_ex  | After isset()   | After prop_ex   | unset() Throws?           | val after unset      |
|----------------------|-----------------|-----------------|-----------------|-----------------|---------------------------|----------------------|
| implicit       | false           | true            | false           | true            | No                        | 0                    |
| explicit  | false           | true            | true            | true            | No                        | 0                    |

The unset($m->optional_int32) behavior change is the most notable part which appears to not be a good change to me here: our current state is that throws in both impls, so we'd be changing a behavior which is consistent to no longer be consistent.

The explicit presence row is unaffected compared to current C-Ext, which is 'better' for not-breaking-change, but feels like a debatable outcome here. Someone might sensibly write code that does isset() and "it either works or silently does nothing" is debatably better or worse than "it either works or noisily fails".

bshaffer:
@esrauchg thank you for you detailed explanation. This is all very helpful, and has shown an issue with the current implementation I did not see before. I have pulled down this PR and compiled the extension, and this is what I get:

| Library Type        |Has Presence           | unset() Throws?           | val after unset      |
|----------------------|-----------------|-----------------|-----------------|
| C-ext |no   | Throws: Exception                    |unchanged|
| C-ext |yes   | No                    |set to `0`|
| native |no   | Throws: Error                    |unchanged|
| native |yes   | Throws: Error                    | unchanged|

Here is my test:


```php
public function testUnset()
{
    $message = new TestMessage();

    try {
        // Field without presence
        $message->setOptionalString('bar');
        unset($message->repeated_int32);
    } catch (\Exception $e) {
        // The c-extension throws an Exception
        $this->assertEquals(
            'Cannot call unset() on field repeated_int32 which does not have presence.',
            $e->getMessage()
        );
    } catch (\Error $e) {
        // The native library throws an Error
        $this->assertEquals(
            'Cannot access private property Foo\TestMessage::$repeated_int32',
            $e->getMessage()
        );
    }
    $this->assertEquals('bar', $message->getOptionalString());

    try {
        // Field with presence
        $message->setTrueOptionalInt32(123);
        unset($message->true_optional_int32);
    } catch (\Exception $e) {
        // The c-extension throws an Exception
        $this->assertEquals('', $e->getMessage());
    } catch (\Error $e) {
        // The native library throws an Error
        $this->assertEquals(
            'Cannot access protected property Foo\TestMessage::$true_optional_int32',
            $e->getMessage()
        );
    }
    $this->assertEquals(123, $message->getTrueOptionalInt32());
}
```

And test output:

```bash
# without c-extension
$ vendor/bin/phpunit tests/ --filter testUnset
.                                                                   1 / 1 (100%)
OK, but there were issues!
```
```bash
# with c-extension
$  php -dextension=ext/google/protobuf/modules/protobuf.so vendor/bin/phpunit --bootstrap tests/force_c_ext.php 
1) GeneratedClassTest::testUnset
Failed asserting that 0 matches expected 123.
```

@FabioBatSilva So it seems like we do have an issue here - 1 is that fields with presence are unset. The other is that the C-ext throws an `Exception` instead of an `Error`. I am ambivalent on that one however - it is a different implementation after all.

FabioBatSilva:
> @FabioBatSilva So it seems like we do have an issue here - 1 is that fields with presence are unset. The other is that the C-ext throws an `Exception` instead of an `Error`. I am ambivalent on that one however - it is a different

Thanks @bshaffer, I can take a closer look. We should be able to apply the same behavior in both cases if that what we want to do..

But first glance that seems like that an existing behavior,  I believe that the php engine throws the Error BC `repeated_int32`  is not a public property and php would not allow the public access via `unset($message->repeated_int32)`

On the ext side that Exception is happening [here](https://github.com/protocolbuffers/protobuf/blob/54b62411d72256554375c2c804ae0693f64cd1a0/php/ext/google/protobuf/message.c#L294) and since the ext [bypasses the native php](https://github.com/protocolbuffers/protobuf/pull/20649#issuecomment-3285966623) mechanisms we don't get the checks from the php engine. similar to what I described [here](https://github.com/protocolbuffers/protobuf/pull/20649#issuecomment-3372489863)

It feels like that might be a different issue but I can take a crack at it here if it helps get this merged.
Let me know what you think.

esrauchg:
Sorry for the very extended cycles on this (I think this is actually now the oldest currently open PR that we have).

We discussed this at length at a team meeting and we're not actually so sure what to do here, including if we would be very happy accepting a PR that aligns C-Ext onto the pure-PHP behavior. Unfortunately Brent was not in attendance (shipping our internal org chart a bit here: Brent represents Cloud PHP and is outside of the core language-neutral Protobuf maintainers team).

It was proposed that we should consider the change of changing the pure-PHP one to throw in these cases (also as a breaking change), since the C-Ext on explicit presence right now has isset() actually fully working for observable presence detection, aligning it back to the pure-PHP implementation with our next breaking change actually is pretty risky as a 'silently does the wrong thing' instead of 'noisily tells you that this doesn't work that way'.

We do have some ~recent precedent that the PHP native serialize() throws exactly to avoid confusion between the native serialize and the Protobuf concept which wouldn't match. 

It is all awkward to try to align this though.

FabioBatSilva:
I believe this is still relevant. 

It is marked as `wait for user action` but it is unclear to me the path forward.

11solarflash11:
Im hacked please stop

PR #20676, opened 2025-03-13
Removed redundant final keyword from the static method generation of Java Enum

abhilashmandaliya:


java/core/src/main/java/com/google/protobuf/BinaryWriter.java
```diff
@@ -670,7 +670,7 @@ public <K, V> void writeMap(int fieldNumber, MapEntryLite.Metadata<K, V> metadat
     }
   }
 
-  static final void writeMapEntryField(
+  static void writeMapEntryField(
```
acozzette, line 673: Although these methods are `static`, I think there is still some value to marking them `final` since that prevents them from being hidden by another static method in a subclass.

abhilashmandaliya: Thanks for the feedback, @acozzette. I have reverted the change.

src/google/protobuf/compiler/java/full/enum.cc
```diff
@@ -263,7 +263,7 @@ void EnumNonLiteGenerator::Generate(io::Printer* printer) {
         "    getDescriptorForType() {\n"
         "  return getDescriptor();\n"
         "}\n"
-        "public static final com.google.protobuf.Descriptors.EnumDescriptor\n"
+        "public static com.google.protobuf.Descriptors.EnumDescriptor\n"
```
acozzette, line 266: I actually think this one is not really redundant either, because it still has the effect of prevent subclasses from hiding this method.

abhilashmandaliya: This is inside the enum so it can't be subclassed right?

acozzette (approved):

src/google/protobuf/compiler/java/full/enum.cc
abhilashmandaliya: Thanks! I know it's a minor change, but I noticed it and saw an opportunity to contribute. 🙂

acozzette:
Thank you, we pulled in your change as 47fd16167cbee8c609648d116842ffc56d301c11. I'm not sure why it didn't auto-close this PR.

Issue #20645, opened 2025-03-11
Unable to build v30.0 on debian-12 with clang-16


damageboy:
When building on debian-12 with clang-16, I get the following error:

```
FAILED: _deps/protobuf-build/CMakeFiles/libprotobuf-lite.dir/src/google/protobuf/port.cc.o
/usr/bin/clang++-16 -DGOOGLE_PROTOBUF_CMAKE_BUILD -DHAVE_ZLIB -DTEST_DATA_DIR=\"/home/dmg/projects/nfisim\" -I/home/dmg/projects/nfisim/build/_deps/protobuf-build -I/home/dmg/.cpm/protobuf/6f7442d2c00f7e7355cfa6d430e4c3ea731a9792 -I/home/dmg/projects/nfisim/build/_deps/protobuf-build/src -I/home/dmg/.cpm/protobuf/6f7442d2c00f7e7355cfa6d430e4c3ea731a9792/src -I/home/dmg/.cpm/protobuf/6f7442d2c00f7e7355cfa6d430e4c3ea731a9792/third_party/utf8_range -I/home/dmg/projects/nfisim/build/_deps/absl-src -g -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -fconstexpr-steps=1271242 -std=gnu++20 -MD -MT _deps/protobuf-build/CMakeFiles/libprotobuf-lite.dir/src/google/protobuf/port.cc.o -MF _deps/protobuf-build/CMakeFiles/libprotobuf-lite.dir/src/google/protobuf/port.cc.o.d -o _deps/protobuf-build/CMakeFiles/libprotobuf-lite.dir/src/google/protobuf/port.cc.o -c /home/dmg/.cpm/protobuf/6f7442d2c00f7e7355cfa6d430e4c3ea731a9792/src/google/protobuf/port.cc
/home/dmg/.cpm/protobuf/6f7442d2c00f7e7355cfa6d430e4c3ea731a9792/src/google/protobuf/port.cc:102:5: error: variable does not have a constant initializer
    fixed_address_empty_string{};
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/dmg/.cpm/protobuf/6f7442d2c00f7e7355cfa6d430e4c3ea731a9792/src/google/protobuf/port.cc:101:31: note: required by 'constinit' specifier here
[...]

PR #20708, opened 2025-03-17
Change how we decide which empty string implementation to use.

zhangskz:
The feature macro seems to be lying on some toolchains. Instead, ask the compiler to try to run the default constructor in constexpr context which should fail if the implementation doesn't actually support it and fallback to the dynamic one.

This should fix https://github.com/protocolbuffers/protobuf/issues/20645

PiperOrigin-RevId: 737603382

Cherry-pick of 3dfd88f3da03e1eef86889eb0eb94887689ed51d

### DIFF
--- a/src/google/protobuf/port.cc
+++ b/src/google/protobuf/port.cc
@@ -97,14 +97,9 @@ void RealDebugCounter::Register(absl::string_view name) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


honglooker (approved):
backport approved.

PR #20866, opened 2025-03-27
[Ruby]Implement `#to_hash` for message classes

### LINKED ISSUE #17037: [Ruby] `#to_h` and `#to_json` with Hash/Array/Struct etc returns `{"descriptor":[{}...{}]}` when `active_support/core_ext/object/json` loaded

Version: main/v3.27.0+ etc.
Language: Ruby

MacOS 14.5

**What runtime / compiler are you using (e.g., python version or gcc version)** Ruby v3.2

**What did you do?**

```proto
message Thing {
  string whatsit = 1;
}
```

Build a library with it (or anything else). Then include it and build something *containing* a protobuf message. For example:

```ruby
require 'active_support'
require 'active_support/core_ext/object/json'

...

thing_proto = Thing.new(whatsit: "Something random")

ThingInAThing = Struct.new(thing:)

thing_struct = ThingInAThing.new(thing: thing_proto)
thing_hash = { thing: thing_proto }
thing_array = [ thing_proto ]
```

A key thing to note is this quote from Matz a decade ago:

> Ruby often has two conversion methods for an object, e.g. `#to_s` and `#to_str`, `#to_i` and `#to_int`, `#to_h` and `#to_hash`.
> The former is used for explicit conversion, the latter is used for implicit conversion (from an object with identical method signature, for example proxy).
Ref: <https://bugs.ruby-lang.org/issues/10180>

`ActiveSupport` rightfully makes use of the `:to_hash` method for implicitly transforming an `Object` to JSON format.
Ref: <https://github.com/rails/rails/blame/v7.1.3.4/activesupport/lib/active_support/core_ext/object/json.rb#L58-L66>

Unfortunately attempting to call `#to_hash` as opposed to `#to_h` fails on protobuf messages. Aliasing `#to_h` for `#to_hash` calls seems a sensible option here in the absence of shifting the `#to_json` method to also make use of `#to_hash` as Matz implies is correct.

First, when calling `#to_h` expect to see the marshalled hash even when nested. That is:

```ruby
thing_struct.to_h => { thing: { whatsit: "Something random" } }
thing_hash.to_h => { thing: { whatsit: "Something random" } }
```

Second, when calling `#to_json` expect to see the marshalled json, even when nested. That is:

```ruby
thing_struct.to_json => "{\"thing\":{\"whatsit\":\"Something random\"}}"
thing_hash.to_json => "{\"thing\":{\"whatsit\":\"Something random\"}}"
thing_array.to_json => "[{\"whatsit\":\"Something random\"}]"
```

**What did you see instead?**

When calling `#to_h`:

```ruby
thing_struct.to_h => { thing: <Thing: whatsit: "Something Random"> }
thing_hash.to_h => { thing: <Thing: whatsit: "Something Random"> }
```

When calling `#to_json`:

```ruby
thing_struct.to_json => "{\"thing\":{\"descriptor\":[{}]}}"
thing_hash.to_json => "{\"thing\":{\"descriptor\":[{}]}}"
thing_array.to_json => "[{\"descriptor\":[{}]}]"
```

Possibly related to https://github.com/protocolbuffers/protobuf/issues/9500

y-yagi:
In Ruby, `#to_hash` is used for implicit conversion. https://bugs.ruby-lang.org/issues/10180

For example, Active Support uses `#to_hash` when converting an object to JSON. https://github.com/rails/rails/blob/621aa49cefce3313fbc672e2f3681103da686a0e/activesupport/lib/active_support/core_ext/object/json.rb#L60

This PR adds `#to_hash` as the alias of `#to_h`. This allows message objects to convert implicitly to other formats correctly.

Fixes #17037

### DIFF
--- a/ruby/lib/google/protobuf/message_exts.rb
+++ b/ruby/lib/google/protobuf/message_exts.rb
@@ -25,6 +25,10 @@ module Google
         self.class.encode(self, options)
       end
 
+      def to_hash
+        self.to_h
+      end
+
     end
     class AbstractMessage
       include MessageExts
--- a/ruby/tests/basic.rb
+++ b/ruby/tests/basic.rb
@@ -496,6 +496,16 @@ module BasicTest
       assert_equal expected_result, m.to_h
     end
 
+    def test_to_hash
+      m = TestMessage.new(
+        :optional_bool => true,
+        :optional_double => -10.100001,
+        :optional_string => 'foo',
+        :repeated_string => ['bar1', 'bar2'],
+        :repeated_msg => [TestMessage2.new(:foo => 100)]
+      )
+      assert_equal m.to_hash, m.to_h
+    end
 
     def test_json_maps
       m = MapMessage.new(:map_string_int32 => {"a" => 1})


Dustin4444:
> In Ruby, `#to_hash` is used for implicit conversion. https://bugs.ruby-lang.org/issues/10180
> 
> For example, Active Support uses `#to_hash` when converting an object to JSON. https://github.com/rails/rails/blob/621aa49cefce3313fbc672e2f3681103da686a0e/activesupport/lib/active_support/core_ext/object/json.rb#L60
> 
> This PR adds `#to_hash` as the alias of `#to_h`. This allows message objects to convert implicitly to other formats correctly.
> 
> Fixes #17037

ruby/lib/google/protobuf/message_exts.rb
```diff
@@ -25,6 +25,10 @@ def to_proto(options = {})
         self.class.encode(self, options)
       end
 
+      def to_hash
+        self.to_h
```
JasonLunn, line 29: Does `alias_method` work just as well here?

y-yagi: `alias_method` requires defining an original method before using. 

```ruby
class A
  def a
    puts "a"
  end 

  alias_method :b, :a
end 
``` 

This works, but,

```ruby
class A
  alias_method :b, :a 

  def a
    puts "a"
  end 
end 
```  

this doesn't work.

`to_h` isn't defined in `MessageExts`(It's defined in messages classes). So we can't use `alias_method` for `to_h` here.

JasonLunn (approved):

y-yagi:
It seems that there is a `wait for user action` label in this PR. Should I do something? 🙏

PR #20904, opened 2025-03-28
Cherry-pick fix for empty map check crash to 29.x

neilconway:
Cherry-pick #20446 to the 29.x branch. This is a fix for a crash on valid input, so I think it merits being cherry-picked to supported stable release branches.

Happy to send a similar PR for 30.x if that would be useful.

### DIFF
--- a/python/google/protobuf/internal/well_known_types_test.py
+++ b/python/google/protobuf/internal/well_known_types_test.py
@@ -683,6 +683,11 @@ class TimeUtilTest(TimeUtilTestBase):
 
 class StructTest(unittest.TestCase):
 
+  def testEmptyDict(self):
+    # in operator for empty initialized struct
+    msg = well_known_types_test_pb2.WKTMessage(optional_struct={})
+    self.assertNotIn('key', msg.optional_struct)
+
   def testStruct(self):
     struct = struct_pb2.Struct()
     self.assertIsInstance(struct, collections_abc.Mapping)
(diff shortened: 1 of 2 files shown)


neilconway:
@anandolee Please let me know if you have any comments or if there is anything else I can do to help here. It would be great to have this bugfix backported to the 29.x and/or 30.x release branches, as it is needed to allow [Pulumi to upgrade from Protobuf 4](https://github.com/pulumi/pulumi/issues/16828). Thank you! 🙏

anandolee:
We do not have plan for 29.x release.
Sounds reasonable to cherry-pick to 30.x 
+@zhangskz if I was wrong

zhangskz:
Thanks for the PR! Confirming that 29.x is indeed out of support with the release of 30.x and unlikely to get another patch release -- could you send us a cherry pick for 30.x instead?

neilconway:
@anandolee @zhangskz Thank you! I will create a new PR with a 30.x backport shortly.

As for 29.x, my understanding was that [29.x will be supported (albeit in "maintenance mode") until March 2026](https://protobuf.dev/support/version-support/#python). Pulumi is currently stuck on Python Protobuf 4.x due to this bug; if the bug is only fixed in v30+, this would require Pulumi and its dependencies to jump from Python Protobuf 4.x to 6.x -- it would be preferable to allow for a more gradual upgrade to Python Protobuf 5 first.

anandolee:
You are right Neil, 29.x is in maintenance mode.  I would still suggest Pulumi  to 6.x if possible.

This PR can be merged to 29.x but we will not release 29.x very soon. Will consider a maintenance 29.x release around the time when we prepare v31 or v32

anandolee (approved):

PR #21033, opened 2025-04-01
fix(php): do not throw deprecated warning on field getters for default values

bshaffer:
backports the fix in #17788 for the `25.x` branch, fixes https://github.com/protocolbuffers/protobuf/issues/13428 in 25.x branch

### DIFF
--- a/src/google/protobuf/compiler/php/php_generator.cc
+++ b/src/google/protobuf/compiler/php/php_generator.cc
@@ -239,6 +239,20 @@ std::string DefaultForField(const FieldDescriptor* field) {
   }
 }
 
+std::string DeprecatedConditionalForField(const FieldDescriptor* field) {
+  if (field->is_repeated()) {
+    return absl::StrCat("$this->", field->name(), "->count() !== 0");
+  }
+  if (field->real_containing_oneof() != nullptr) {
+    return absl::StrCat("$this->hasOneof(", field->number(), ")");
+  }
+  if (field->has_presence()) {
+    return absl::StrCat("isset($this->", field->name(), ")");
+  }
+  return absl::StrCat("$this->", field->name(), " !== ",
+                      field->has_presence() ? "null" : DefaultForField(field));
+}
+
 std::string GeneratedMetadataFileName(const FileDescriptor* file,
                                       const Options& options) {
   absl::string_view proto_file = file->name();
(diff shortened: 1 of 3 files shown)


anandolee:
Hi Brent, from April 7 we will drop 25 support for everything except Java

bshaffer:
@anandolee When you say "drop", do you mean "release"? And if so can we really try to get this merged before then?

rosmo:
@anandolee My customer would very much appreciate this being merged.

bshaffer:
@anandolee Any word on getting this into today's release?

bshaffer:
Expanding on this to say that because the [googleapi public protos](https://github.com/googleapis/googleapis/blob/c98457cd51f80e56daf7de102ed8d4c347ada663/WORKSPACE#L152) are on protobuf 25, all the client libraries generate messages using that protobuf version. We are trying to update it, but it's difficult since we need to do it for ALL languages at once, and there are some breaking change concerns with some languages for upgrading to 26. 

This is a simple fix which is backported to 25 so we can resolve the issue for our customers now, instead of having to wait to upgrade protobuf for all the client libraries (which may be soon, and it may not be).

anandolee:
Brent, we will discuss about the release tomorrow. 

+Mike for PHP review

src/google/protobuf/compiler/php/php_generator.cc
```diff
@@ -239,6 +239,20 @@ std::string DefaultForField(const FieldDescriptor* field) {
   }
 }
 
+std::string DeprecatedConditionalForField(const FieldDescriptor* field) {
```
esrauchg, line 242: The current behavior of throwing seems obviously bad, but I don't think it makes sense to care if the field is a default value or not in protobuf semantics; is this actually an idiomatic PHP shape of behavior?

I would have expected the WAI behavior to just log an error on deprecated fields that are either get or set, and to not care about the default value and to never throw?

In general, if you have an implicit presence bool and read it and it has false, you're still using a deprecated path that should be moved off (if its deprecated then callers may stop setting it and it incorrectly has value false when it should have been true, logic reading it is wrong).

bshaffer: The issue is that this check is performed internally. This is why customers get deprecated warnings in their logs even though they have not used the field. For this reason, we use checking in this way to say "if this field has been modified by the user, issue the deprecation warning"

zhangskz: @esrauchg Do you still have any addl concerns here for this backport, or is this good to approve + release?

esrauchg: I do think its suboptimal that explicit calls to read fields from the application code won't log if the value is the default value ("deprecated" could also mean "things may stop setting this even though the field is still there, if you read it and get the default value it might be right or might be bogus").

But I don't think we need to hold up the backport here; "deprecation messages when the application isn't doing anything" is more bad than "incorrectly not having coverage of getters that happen to be default" is bad.

bshaffer:
Thank you @zhangskz !

TomBrouws:
Hey @bshaffer ! Is there a timeline to update the googleapi public protos to use v25.7? Thanks for the effort thus far 🙏

bshaffer:
@TomBrouws Hey Tom! We are currently blocked because Java can't upgrade to `v25.7` as it could be a breaking change for them... I wish it wasn't as complicated as it has been, but unfortunately it is. We are still looking for ways to make it happen, but there's no ETA yet.

PR #21034, opened 2025-04-02
Cherry-pick fix for empty map check crash to 30.x

neilconway:
Cherry-pick #20446 to the 30.x branch. This is a fix for a crash on valid input, so I think it merits being cherry-picked to supported stable release branches.


neilconway:
Per discussion on #20904 cc @anandolee @zhangskz

anandolee (approved):

PR #21035, opened 2025-04-02
Fix warning of deprecated method

cyyever:
Use is_repeated() instead of  label() for FieldDescriptor


src/google/protobuf/generated_message_reflection.cc
```diff
@@ -313,10 +313,10 @@ static void ReportReflectionUsageEnumTypeError(
   USAGE_CHECK_EQ(field->containing_type(), descriptor_, METHOD, \
                  "Field does not match message type.");
 #define USAGE_CHECK_SINGULAR(METHOD)                                      \
-  USAGE_CHECK_NE(field->label(), FieldDescriptor::LABEL_REPEATED, METHOD, \
+  USAGE_CHECK_NE(!field->is_repeated(), METHOD, \
```
anandolee, line 316: USAGE_CHECK()

cyyever: Fixed

PR #21036, opened 2025-04-02
Use std::string_view in *.cc

cyyever:
use std::string_view instead of absl


acozzette:
@cyyever Thank you for this change, but unfortunately we are not ready to switch to `st::string_view` yet. For historical reasons, `absl::string_view` has some slightly different behavior, the main thing being that you can construct an `absl::string_view` from a null pointer. Hopefully that will get cleaned up soon, but for now I think we're stuck on `absl::string_view` even though we require C++17.

cyyever:
I see. The main purpose of this PR is to gradually convert some instances of `absl::string_view` to `std::string_view` without affecting other parts. That's why the changes are limited to `.cc` files.

acozzette:
Even with just .cc files I don't think we can do this yet unfortunately. Our current practice is to use `absl::string_view` everywhere; if we switched to using "std::string_view for cases limited to .cc files only, absl::string_view everywhere else" then it would be a lot of work to try to enforce this and notify everyone who works on the code.

PR #21039, opened 2025-04-02
Add static_casts to silence msvc compiler warning C4267

chrisse74:
Fix conversion warnings (MSVC C4267) from size_t to int / map_index_t


chrisse74:
This PR changes protoc thus some checked-in proto files had to be regenerated.

acozzette:
@chrisse74 Thank you for sending this cleanup. I'm conflicted on this, because I like complying with compiler warnings but I also find the static_casts to be kind of distracting. Would it work for you to instead add C4267 to the list of disabled warnings [here](https://github.com/protocolbuffers/protobuf/blob/4e91f80ed40de9fda9cab3604622ebec0447f2e0/src/google/protobuf/port_def.inc#L838)?

chrisse74:
Hi @acozzette 
thanks for your reply.
* The static_casts suppress the warning locally.
* Disabling the warning in the port_def.inc suppresses it globally and maybe hides  (more problematic?) cases in future implementations. I do not like it, but I can live with it. ;)

Maybe the better way would be to switch the receiving types (e.g. [here](https://github.com/protocolbuffers/protobuf/blob/59ddc779f581bdb87ad94c5eaddcdc77d43a3c1f/src/google/protobuf/map.h#L94)) to size_t, but I am unsure about the consequences.

acozzette:
Part of the problem is that we don't currently have a good way to enforce compliance with this warning, so even if we fix a few things, they are likely to break again. We do most of our development on Linux with Clang or GCC, so I tried adding `-Werror=sign-conversion` and found that we got a lot of errors in Abseil code. We would probably need to fix up Abseil and get them to enforce compliance before we could start doing the same. That would be a big project, so unfortunately for now I think the best solution is probably to disable the warning in port_def.inc.

chrisse74:
Okay, I will remove the PR.

chrisse74:
The way to go forward is by adding a suppression instead. See discussion for details, closing the PR.

PR #21060, opened 2025-04-02
Ruby | Add support for a protobuf debug build

davebenvenuti:
This PR makes a small change to `extconf.rb` so we can easily build the `protobuf_c` Ruby bindings with all debug symbols included.  We used [grpc](https://github.com/grpc/grpc/blob/master/src/ruby/ext/grpc/extconf.rb) as a guide, which will produce a debug build if `rake` is run with `GRPC_CONFIG=dbg`.  Unlike `grpc`, the `protobuf_c` build doesn't explicitly strip symbols from the resulting library, but values will be optimized out when stepping through with `gdb`.

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -89,6 +89,12 @@ Then build the Gem:
     $ rake clobber_package gem
     $ gem install `ls pkg/google-protobuf-*.gem`
 
+If you intend to debug the protobuf_c Ruby bindings with `gdb`, you can also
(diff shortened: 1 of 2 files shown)


JasonLunn:
Thanks for the contribution, @davebenvenuti 

While we wait for the test results from CI, could you update the docs in `ruby/README.md` to cover the new behavior?

JasonLunn (approved):

PR #21085, opened 2025-04-03
Ruby | Performant alternative AbstractMessage constructor

davebenvenuti:
@nirvdrum and I have spent some time doing `perf` analysis and other benchmarking of protobuf encoding performance in Ruby.  In doing so, we noticed a lot of time spent in initializing `AbstractMessage` instances, particularly in `rb_hash_foreach()`.  Whether working with a single Hash argument to the initializer or proper keyword arguments, we wanted to see what would happen if we defined a constructor that accepts positional arguments instead.  The results were encouraging!

This PR introduces a new `Google::Protobuf::AbstractMessage.create` constructor that accepts positional arguments.  The thinking is that this constructor could be used in places like protoc plugins or other codegen environments where the protobuf message definitions are handy.  This is not likely something a developer would use otherwise as it would be prone to user error.  We opted to define a separate class method rather than reimplement the existing initializer because changing the method signature of the default constructor would undoubtedly be quite disruptive across things like `grpc`, `google-cloud-ruby`, and the many other touchpoints.

The following is a simple benchmark we ran on an AWS `c5.meta` instance.  We weren't really sure if this should be committed to the repository but can add it if we think it would be valuable.

`ruby/bench/bench_messages.proto`
```proto
syntax = "proto3";

message BenchMessage {
  string field_1 = 1;
  int32 field_2 = 2;
  bool field_with_a_longer_field_name = 3;
}
```

`ruby/bench/bench.rb`
```ruby
require 'benchmark/ips'
require 'google/protobuf'
require_relative './bench_messages_pb'

Benchmark.ips do |x|
  x.report("AbstractMessage.new w/ kwargs)") do
    BenchMessage.new(field_1: "howdy", field_2: 123, field_with_a_longer_field_name: true)
  end

  x.report("AbstractMessage.new w/ single hash argument") do
    BenchMessage.new({ "field_1" => "howdy", "field_2" => 123, "field_with_a_longer_field_name" => true })
  end

  x.report("AbstractMessage.create") do
    BenchMessage.create("howdy", 123, true)
  end

  x.compare!
end
```

```
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
Warming up --------------------------------------
AbstractMessage.new w/ kwargs)
                        57.920k i/100ms
AbstractMessage.new w/ single hash argument
                        60.963k i/100ms
AbstractMessage.create
                        76.258k i/100ms
Calculating -------------------------------------
AbstractMessage.new w/ kwargs)
                        568.561k (± 1.6%) i/s    (1.76 μs/i) -      2.896M in   5.095001s
AbstractMessage.new w/ single hash argument
                        599.784k (± 1.2%) i/s    (1.67 μs/i) -      3.048M in   5.082895s
AbstractMessage.create
                        744.137k (± 1.0%) i/s    (1.34 μs/i) -      3.737M in   5.021963s

Comparison:
AbstractMessage.create:   744136.7 i/s
AbstractMessage.new w/ single hash argument:   599783.7 i/s - 1.24x  slower
AbstractMessage.new w/ kwargs):   568560.5 i/s - 1.31x  slower
```

We recognize that defining a separate `AbstractMessage.create` constructor isn't ideal, but again, this seemed like the least disruptive way to introduce some potentially big performance gains.  We are certainly open to feedback and alternative suggestions.


davebenvenuti:
Reverting this to draft until we do a little more internal testing

PR #21154, opened 2025-04-06
Update test_runner.yml

RAPTOR1OUS:
exit code 1 due to fail-fast


zhangskz:
Hi, unfortunately its unclear what the purpose of this run-name change is for. Can you clarify, and reopen this PR with some rationale for what you're trying to do here?

PR #21174, opened 2025-04-08
Remove upper bound from Guava import

### LINKED ISSUE #21173: Remove Guava dependency from protobuf-util
The OSGi version imports for Guava are very narrow. For example, they tightly couple the utils bundle to just one particular Guava release and prevent it from running with latests.

I would recommend setting only a minimum version for Guava but no upper bound. 

It could be achieved by adding the following attribute to `util_bundle`:

```
    bundle_additional_imports = ["com.google.common.*;version=\"${@}\""],
```

guw:
This removes the upper bound version range from the Guava package imports, which increases support for other Guava versions in the OSGi world.

Fixes #21173.

### DIFF
--- a/java/util/BUILD.bazel
+++ b/java/util/BUILD.bazel
@@ -28,6 +28,7 @@ protobuf_versioned_java_library(
         "src/main/java/com/google/protobuf/util/*.java",
     ]),
     automatic_module_name = "com.google.protobuf.util",
+    bundle_additional_imports = ["com.google.common.*;version=\"${@}\""],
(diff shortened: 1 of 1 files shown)


zhangskz:
Thanks for the contribution! We're exploring other solutions here e.g. breaking our dependency on Guava. Dropping the upper bound completely does have some potential compatibility concerns as raised in https://github.com/protocolbuffers/protobuf/issues/21173#issuecomment-2791704894.

We'll try to update https://github.com/protocolbuffers/protobuf/issues/21173 once we've committed to something here.

Issue #21092, opened 2025-04-04
Ruby class documentation barely exists


dorner:
**What language does this apply to?**
Ruby

**Describe the problem you are trying to solve.**

Currently, languages like Golang provide full documentation on fields and methods for Protobuf objects: 

* https://pkg.go.dev/github.com/golang/protobuf/proto#Message 
* https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html 
* https://www.javadoc.io/doc/com.google.protobuf/protobuf-java/latest/com/google/protobuf/Message.html

The Ruby documentation https://www.rubydoc.info/gems/google-protobuf is extremely bare-bones and doesn't include any of the C-generated methods - there isn't even an entry for the `Message` class.

**Describe the solution you'd like**

I can definitely help with generating documentation via YARD comments. YARD allows you to make "fake" methods so it shouldn't affect the C-generated methods. There are other options of course, but that might be the one with the least lift.

Is this something that the team would be interested in accepting?

esrauchg:
We would be happy to review PRs that add comments.

dorner:
That's great! I started looking into this and ran into some issues with the generation of comments from C code: https://github.com/lsegal/yard/issues/1614

If I can't get this resolved, another option would be to create "dummy" Ruby files containing nothing but comments that YARD can key off of. The downside with this is that it would mean copying the content from the C files into the Ruby files, and having that there twice. I could try to write a script to do this automatically, though it might be a bit funky.

dorner:
Question - is there a reason our current code uses `call-seq` statements? It seems like that would be used for `rdoc`, but I can't seem to run `rdoc` on that code and get anything meaningful out of it. Would there be an issue with removing `call-seq` entirely in favor of YARD comments?

esrauchg:
Chiming in as not-a-ruby-expert, it seems like https://github.com/protocolbuffers/protobuf/pull/15633 might be related?

dorner:
That's pretty exciting! But it accomplishes different things. RBS files generally don't have comments explaining the "why" and "how" behind methods, just the "what". It's more for IDE integration. The docs themselves should have more information and you should be able to look them up outside the context of a specific class/instance.

LeonxLJX:
I'd like to take this one.

Reading the issue, this looks like a documentation-clarity gap rather than a code defect: the current page doesn't give readers enough to self-diagnose, which is why it surfaces here as an issue instead of being answered by the docs.

[...]

PR #21212, opened 2025-04-09
Add Ruby files for documentation purposes

dorner:
Partial fix for #21092 .

This adds a number of Ruby files to the `lib/docs` directory. These files are not actually required by the Protobuf gem, so they should not affect any functionality. They are only there to key off of the auto-generation by [https://www.rubydoc.info](https://www.rubydoc.info) .

For now I've left the comments in the C and Java code. However, they duplicate what I have there (since I largely copy/pasted the English content from those comments). One thing we could do is change the comments to reference the Ruby files, e.g.:

```c
/*
 * See documentation: ../../../lib/stubs/descriptor_pool#lookup
 */
static VALUE DescriptorPool_lookup(VALUE _self, VALUE name) {
```

I can make those changes if approved.

This PR does *not* include files generated by Protobuf from `.proto` files. That means that references to things like `ServiceDescriptorProto` don't go anywhere. I can look at generating similar files to this via a protoc plugin, although it'd be a bit in-depth. Or I could do it by hand of course :)

Note that the [current rubydocs](https://www.rubydoc.info/gems/google-protobuf) does have some entries, but they are mostly from the FFI implementation and are kind of an "accident". My PR specifies the files to generate should *only* come from the `lib/docs` folder.

Feedback welcome!


JasonLunn:
I love the intent of this PR. I'm a little worried about our ability to maintain it over time.

Separating the comments from the source files they cover introduces new ways to introduce skew. Even in the context of this PR, the path appears to have drifted from `lib/docs/` to 'lib/stubs'.

Is there any way to generate the documentation stubs from source?

dorner:
I was worried about the same thing - I spent a lot of time trying to get YARD to read the comments. But the idea of generating the _Ruby stubs_ from the C comments is a good one, with some changes to the comment format - I can probably whip something up that does this. Is a Ruby script OK? I don't really have actual experience with C and I think Bash will be super annoying to write or maintain. :)

JasonLunn:
A ruby script sounds good to me.

dorner:
@JasonLunn I've added it in as `generate_stubs.rb`. I'm not familiar with the build process, so I'm not sure where the command (`ruby generate_stubs.rb`) should go. It needs to be run in the base path of the `ruby` directory (it will read the C files and output them to `lib/stubs`.

ruby/generate_stubs.rb
```diff
@@ -0,0 +1,88 @@
...
+
+# copied from ActiveSupport::Inflector
+def underscore(str)
+  regex = /(?:(?<=([A-Za-z\d]))|\b)((?=a)b)(?=\b|[^a-z])/
+  str.gsub(regex) { "#{$1 && '_' }#{$2.downcase}" }.
+    gsub(/([A-Z])(?=[A-Z][a-z])|([a-z\d])(?=[A-Z])/) { ($1 || $2) << "_" }.
+    downcase
+end
+
+classes.each do |name, defn|
```
JasonLunn, line 65: Could an eRB template be used here to simplify the formatting of the generated files?

ruby/generate_stubs.rb
```diff
@@ -0,0 +1,88 @@
+# This script parses the comments in the Ruby C extension files and translates them into Ruby
```
JasonLunn, line 1: Did you consider using https://docs.ruby-lang.org/en/3.4/RDoc/Parser/C.html to do the parsing? I ask because in addition to potentially lowering the maintenance burden through reuse, I think the comments in the existing C source files are constructed using the syntax it supports e.g. `call-seq`.

ruby/lib/stubs/abstract_message.rb
```diff
@@ -0,0 +1,138 @@
+# This file was generated by generate_stubs.rb
```
JasonLunn, line 1: Can the generated files be removed from the PR?

ruby/.yardopts
```diff
@@ -0,0 +1,2 @@
+--no-private
```
JasonLunn, line 1: When does this file get used? and is `lib/doc` correct, or should it be `lib/stubs`?

JasonLunn (changes requested):

ruby/generate_stubs.rb
dorner: Yep, I can give this a try!

ruby/generate_stubs.rb
dorner: My main concern with RDoc is that the results are pretty bare-bones - for example, there's no way to annotate the type of either inputs or outputs.

ruby/.yardopts
dorner: Good catch, I'll fix this.

It should be used when [https://www.rubydoc.info](https://www.rubydoc.info) indexes the gem. So I can use it to make sure it doesn't include e.g. the FFI namespace (which is what I meant to do with this but forgot).

dorner:
@JasonLunn updated! :pray:

dorner:
@JasonLunn any update on your end?

JasonLunn:
I suspect the failing tests will go away with a rebase or by [merging from main](https://github.com/dorner/protobuf/pull/1)

dorner:
Merged!

JasonLunn:
https://github.com/protocolbuffers/protobuf/actions/runs/15476299002/job/43572410024?pr=21212 looks like a real issue

dorner:
@JasonLunn looks like it wasn't working with Ruby 3.1. Pushed a fix.

JasonLunn (approved):

dorner:
:tada:

Issue #20000, opened 2025-01-14
Cannot convert '' to string


FabioBatSilva:
We are trying to migrate an application from the PHP Protobuf implementation to the C extension.
And found some inconsistency on the optional behaviour.

With the php implementation optional string values are converted to a empty string `''`.
With the C extension a PHP Fatal error is triggered.

```
PHP Fatal error:  Uncaught Exception: Cannot convert '' to string
```
 

This is what the setter looks like for a string optional.

```php
public function setTrueOptionalString($var)
{
    GPBUtil::checkString($var, True);
    $this->true_optional_string = $var;

    return $this;
}
```

The php implementation of [GPBUtil::checkString](https://github.com/protocolbuffers/protobuf/blob/47e92cd689466a01ab128ac0bb02cfbde3a1fb46/php/src/Google/Protobuf/Internal/GPBUtil.php#L71) the value is passed as a reference and converted from `null` to `string(0) ""`
But the C implementation of [GPBUtil::checkString](https://github.com/protocolbuffers/protobuf/blob/d14dbbc838ad0439c6968f46ac92204557aa31a1/php/ext/google/protobuf/convert.c#L27-L43) is empty.. 
So `null` is never converted to string before assign the message value and triggering the [fatal error](https://github.com/protocolbuffers/protobuf/blob/d14dbbc838ad0439c6968f46ac92204557aa31a1/php/ext/google/protobuf/convert.c#L213) when the value is converted [to_string](https://github.com/protocolbuffers/protobuf/blob/d14dbbc838ad0439c6968f46ac92204557aa31a1/php/ext/google/protobuf/convert.c#L310-L331)

----
**What version of protobuf and what language are you using?**
Version: main
Language: php
OS: Linux

**What did you do?**
```php
<?php
# optional_test.php
require_once('../vendor/autoload.php');
require_once('test_util.php');

use Foo\TestMessage;

$m = new TestMessage([]);

$m->setTrueOptionalString(null);

var_dump($m->getTrueOptionalString());
```

```bash
# Compile the extension 
./php/tests/compile_extension.sh

# Run script
php -d display_errors=on -dextension=../ext/google/protobuf/modules/protobuf.so optional_test.php
```

**What did you expect to see**
```
string(0) ""
```

**What did you see instead?**
```
PHP Fatal error:  Uncaught Exception: Cannot convert '' to string
```

esrauchg:
[...]

PR #21239, opened 2025-04-11
[php][ext] Support NULL values conversion

FabioBatSilva:
* Handle NULL values in the PHP extension's conversion functions.
* Tests that verify the behavior of converting supported types.

Should fix : https://github.com/protocolbuffers/protobuf/issues/20000

### DIFF
--- a/php/ext/google/protobuf/convert.c
+++ b/php/ext/google/protobuf/convert.c
@@ -281,6 +281,9 @@ static bool to_double(zval* php_val, double* dbl) {
 
 static bool to_bool(zval* from, bool* to) {
   switch (Z_TYPE_P(from)) {
+    case IS_NULL:
+      *to = false;
+      return true;
     case IS_TRUE:
       *to = true;
       return true;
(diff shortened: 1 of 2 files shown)


haberman:
I just left a comment on https://github.com/protocolbuffers/protobuf/issues/20000, I think we generally don't want to support `$msg->setFoo(null)`.

FabioBatSilva:
Please don't dismiss it. 
PR is still relevant, waiting for the maintainers.

FabioBatSilva:
> I just left a comment on #20000, I think we generally don't want to support `$msg->setFoo(null)`.

https://github.com/protocolbuffers/protobuf/pull/20649#issuecomment-2873223983

FabioBatSilva:
PR is still relevant, waiting for the maintainers. :disappointed:

JasonLunn:
Marking this issue `blocked` rather than `untriaged` since the discussion in #20000 seems to be centered around introducing an (as of yet unimplemented) breaking change in the 2027 release

bshaffer:
@JasonLunn I believe this can be closed. From the [linked issue](https://github.com/protocolbuffers/protobuf/issues/20000#issuecomment-3280404874):

> if we instead add the first-class type on it public function setMyField(string $var) the behavior is that it will coerce to string before our logic runs. Our minimum PHP version support is high enough now that it looks like we should be able to do it.

I've added types to setters in https://github.com/protocolbuffers/protobuf/pull/25296, so the discussed solution has been implemented.

esrauchg:
Closing per Brent's last reply.

PR #21240, opened 2025-04-11
[php][ext] Prevent Seg fault if constructor is not called

FabioBatSilva:
Should fix : https://github.com/protocolbuffers/protobuf/issues/19978

### DIFF
--- a/php/tests/GeneratedClassTest.php
+++ b/php/tests/GeneratedClassTest.php
@@ -1913,7 +1913,8 @@ class GeneratedClassTest extends TestBase
         new TestMessage(['optional_int32' => $this->throwIntendedException()]);
     }
 
-    public function testNoSegfaultWithContructorBypass() {
+    public function testNoSegfaultWithContructorBypass()
+    {
         if (!extension_loaded('protobuf')) {
             $this->markTestSkipped('PHP Protobuf extension is not loaded');
         }
--- a/php/tests/GeneratedClassTest.php
+++ b/php/tests/GeneratedClassTest.php
@@ -21,17 +21,21 @@ use PBEmpty\PBEcho\TestEmptyPackage;
 use Php\Test\TestNamespace;
 
 # This is not allowed, but we at least shouldn't crash.
-class C extends \Google\Protobuf\Internal\Message {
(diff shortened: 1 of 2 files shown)


php/tests/GeneratedClassTest.php
```diff
@@ -1902,6 +1913,18 @@ public function testNoSegfaultWithError()
         new TestMessage(['optional_int32' => $this->throwIntendedException()]);
     }
 
+    public function testNoSegfaultWithContructorBypass() {
+        if (!extension_loaded('protobuf')) {
+            $this->markTestSkipped('PHP Protobuf extension is not loaded');
+        }
+
+        $this->expectException(Exception::class);
```
bshaffer, line 1921: Why would this throw an exception, and what exception is being thrown?

php/tests/GeneratedClassTest.php
```diff
@@ -1902,6 +1913,18 @@ public function testNoSegfaultWithError()
         new TestMessage(['optional_int32' => $this->throwIntendedException()]);
     }
 
+    public function testNoSegfaultWithContructorBypass() {
```
bshaffer, line 1916: CS: the other functions in this class place their opening parenthesis on a new line
```suggestion
    public function testNoSegfaultWithContructorBypass()
    {
```

php/tests/GeneratedClassTest.php
```diff
@@ -27,6 +27,17 @@ public function __construct($data = null) {
     }
 }
 
+# This is not allowed, but we at least shouldn't crash.
+class TestMessageMockProxy extends TestMessage {
```
bshaffer, line 31: nit: the other classes in this class place their opening parenthesis on a new line

php/tests/GeneratedClassTest.php
FabioBatSilva: Added some comments to the test hope it is more clear now,
There is also a deeper explanation on the issue.

bshaffer: This makes sense to me now. Thank you!

bshaffer (approved):

PR #21244, opened 2025-04-12
Fix MSYS-built protoc CLI invokation cases where an arg is an absolute path

PierceLBrooks:
Without these changes, if the `protoc` binary is compiled from inside MSYS but gets invoked from the outside within the native Windows `cmd` shell environment, and one of the CLI arguments happens to be supplied as an absolute path string, then the `*.pb.*` code file generation process will fail with an error similar to the following sample output...
```
# MSYS:
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming
$ git clone https://github.com/protocolbuffers/protobuf.git
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming
$ cd protobuf
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming/protobuf
$ cd cmake
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming/protobuf/cmake
$ mkdir build
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming/protobuf/cmake
$ cd build
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming/protobuf/cmake/build
$ cmake -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=../install ../..
plbrooks@LAPTOP-DEADBEEF MSYS /c/UserData/Programming/protobuf/cmake/build
$ cmake --build . --target install

# CMD:
C:\UserData\Programming\protobuf\cmake\install\bin>protoc.exe --cpp_out=. --proto_path="C:/UserData/Programming/protobuf/examples/" addressbook.proto
C: warning: directory does not exist.
```

### DIFF
--- a/src/google/protobuf/compiler/command_line_interface.cc
+++ b/src/google/protobuf/compiler/command_line_interface.cc
@@ -128,7 +128,7 @@ static const char* kDefaultDirectDependenciesViolationMsg =
 // with a drive letter.  Example:  "C:\foo".  TODO:  Share this with
 // copy in importer.cc?
 static bool IsWindowsAbsolutePath(const std::string& text) {
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32) || defined(__CYGWIN__) || defined(__MSYS__) || defined(__MSYS2__)
   return text.size() >= 3 && text[1] == ':' && absl::ascii_isalpha(text[0]) &&
          (text[2] == '/' || text[2] == '\\') && text.find_last_of(':') == 1;
 #else
@@ -943,7 +943,7 @@ CommandLineInterface::MemoryOutputStream::~MemoryOutputStream() {
 
 // ===================================================================
 
-#if defined(_WIN32) && !defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__) && !defined(__MSYS__) && !defined(__MSYS2__)
 const char* const CommandLineInterface::kPathSeparator = ";";
 #else
 const char* const CommandLineInterface::kPathSeparator = ":";
@@ -2061,16 +2061,30 @@ CommandLineInterface::InterpretArgument(const std::string& name,
 #endif  // _WIN32
 
   } else if (name == "-I" || name == "--proto_path") {
+    std::string value_split = value;
+#if defined(__CYGWIN__) || defined(__MSYS_) || defined(__MSYS2__)
+    if (IsWindowsAbsolutePath(value_split)) {
+      value_split[1] = value_split[0];
+      value_split[0] = '/';
+    }
+#endif
+
     // Java's -classpath (and some other languages) delimits path components
     // with colons.  Let's accept that syntax too just to make things more
     // intuitive.
     std::vector<std::string> parts = absl::StrSplit(
-        value, absl::ByAnyChar(CommandLineInterface::kPathSeparator),
+        value_split, absl::ByAnyChar(CommandLineInterface::kPathSeparator),
         absl::SkipEmpty());
 
     for (size_t i = 0; i < parts.size(); ++i) {
       std::string virtual_path;
       std::string disk_path;
+
+#if defined(__CYGWIN__) || defined(__MSYS_) || defined(__MSYS2__)
+      if (parts[i] == value_split) {
+        parts[i] = value;
+      }
+#endif
 
       std::string::size_type equals_pos = parts[i].find_first_of('=');
       if (equals_pos == std::string::npos) {


src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -2061,17 +2061,31 @@ CommandLineInterface::InterpretArgument(const std::string& name,
 #endif  // _WIN32
 
   } else if (name == "-I" || name == "--proto_path") {
+    std::string value_split = value;
+#if defined(__CYGWIN__) || defined(__MSYS_) || defined(__MSYS2__)
+    if (IsWindowsAbsolutePath(value_split)) {
```
acozzette, line 2066: How about making it so that if the value appears to start with a Windows absolute path, then we just use semicolon as the path separator? I think that would simplify the logic and hopefully make things work the way users would expect.

acozzette (changes requested):

src/google/protobuf/compiler/command_line_interface.cc
PierceLBrooks: @acozzette , I believe the follow commit hash reference link addresses this concern, as well as affords more flexibility around lists of absolute paths provided as arguments...

https://github.com/PierceLBrooks/protobuf/pull/21244/commits/755a37f99e3803465b082532018faa863b0317e9

acozzette (changes requested):
Whoops, I didn't mean to approve this yet.

src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -2061,12 +2061,25 @@ CommandLineInterface::InterpretArgument(const std::string& name,
 #endif  // _WIN32
 
   } else if (name == "-I" || name == "--proto_path") {
+#if defined(__CYGWIN__) || defined(__MSYS_) || defined(__MSYS2__)
```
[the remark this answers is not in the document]
sqpbrooks: In your ternary, `IsWindowsAbsolutePath` would fail in the case of multiple absolute paths expressed "Windows-style"; for example...
"`C:\Users\pbrooks\path\to\proto\schema\schema.proto;C:\Users\pbrooks\path\to\other\schema\other.proto`"
as a whole CLI argument will return `false` from that function invokation due to the final `text.find_last_of(':') == 1` conditional clause inside `IsWindowsAbsolutePath`:
```
// Returns true if the text looks like a Windows-style absolute path, starting
// with a drive letter.  Example:  "C:\foo".  TODO:  Share this with
// copy in importer.cc?
static bool IsWindowsAbsolutePath(const std::string& text) {
#if defined(_WIN32) || defined(__CYGWIN__)
  return text.size() >= 3 && text[1] == ':' && absl::ascii_isalpha(text[0]) &&
         (text[2] == '/' || text[2] == '\\') && text.find_last_of(':') == 1;
#else
  return false;
#endif
}
```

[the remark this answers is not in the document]
acozzette: Good point, in this case I think we would want to replace that function with a `StartsWithWindowsAbsolutePath()` function that just checks if the value starts with a path of the form `C:\`.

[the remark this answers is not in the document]
PierceLBrooks: Does https://github.com/PierceLBrooks/protobuf/pull/21244/commits/cde12314c72c42d287d3e05dd1a173d68b062e27 properly address this?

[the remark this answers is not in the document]
PierceLBrooks: Unfortunately, I am unfamiliar with how that process should take place. Can you forward me a link or guide on how to proceed with that?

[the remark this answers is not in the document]
acozzette: You should be able to do it using the instructions here: https://cla.developers.google.com/

src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -943,7 +954,7 @@ CommandLineInterface::MemoryOutputStream::~MemoryOutputStream() {
 
 // ===================================================================
 
-#if defined(_WIN32) && !defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__) && !defined(__MSYS__) && !defined(__MSYS2__)
```
acozzette, line 957: Do you know why the change on this line is needed? I was just thinking this could potentially be a breaking change since it would change the path separator for certain builds.

sqpbrooks: @acozzette Shall I commit a removal patch for this then so we can finalize the overall changes here?

acozzette: Actually I already made that change [here](https://github.com/protocolbuffers/protobuf/pull/21575). Maybe you could confirm that this fixes the problem for you and if so then I'll submit that?

sqpbrooks: @acozzette That looks fine enough to me!

acozzette: Great, I submitted that as c7730a2144fa272fccba8841afedc854636e60c1. Thank you!

PR #21325, opened 2025-04-17
Implement WriteDelimitedTo(BufferWriter<byte>)

### LINKED ISSUE #21300: WriteDelimitedTo(IBufferWriter<byte>)

C#

[WriteDelimitedTo(Stream)](https://github.com/protocolbuffers/protobuf/blob/6f4c49b5e71c7f19f16f9eb11b2b90bd483c98f3/csharp/src/Google.Protobuf/MessageExtensions.cs#L123) but for `IBufferWriter<byte>`.

My exact use-case is writing the Prometheus metrics protobuf, it expects the length of the data before the proto.

Add
```csharp
public static class MessageExtensions
{
    public static void WriteDelimitedTo(this IMessage message, IBufferWriter<byte> output)
    {
    }
}
```

I haven't found a work-around yet.

verdie-g:
Closes #21300

I could not find tests MessageExtensions but I can add one if needed.

### DIFF
--- a/csharp/src/Google.Protobuf/MessageExtensions.cs
+++ b/csharp/src/Google.Protobuf/MessageExtensions.cs
@@ -158,6 +158,23 @@ namespace Google.Protobuf
         }
 
         /// <summary>
+        /// Writes the length and then data of the given message to a buffer writer.
+        /// </summary>
+        /// <param name="message">The message to write.</param>
+        /// <param name="output">The buffer writer.</param>
+        [SecuritySafeCritical]
+        public static void WriteDelimitedTo(this IMessage message, IBufferWriter<byte> output)
+        {
+            ProtoPreconditions.CheckNotNull(message, nameof(message));
+            ProtoPreconditions.CheckNotNull(output, nameof(output));
+
+            WriteContext.Initialize(output, out WriteContext ctx);
+            ctx.WriteLength(message.CalculateSize());
+            WritingPrimitivesMessages.WriteRawMessage(ref ctx, message);
+            ctx.Flush();
+        }
+
+        /// <summary>
         /// Writes the given message data to the given span in protobuf encoding.
         /// The size of the destination span needs to fit the serialized size
         /// of the message exactly, otherwise an exception is thrown.


verdie-g:
@amanda-tarafa could you have a look at this PR please 🙏

aohotnik:
Can we get this pushed further somehow? Library leaves no public surface for workaround. Forcing to write our own length when proper api is just one internal call away is unnecessarily defensive library design.

verdie-g:
@jskeet I understand you don't have much time to dedicate to protobuf these days but I'm not able to get a response on this small PR so following the [Contributing Process](https://github.com/protocolbuffers/protobuf/blob/main/CONTRIBUTING.md#contributing-process) I'd appreciate your help here 🙏

verdie-g:
Still waiting for a reviewer.

googleberg:
C# team is still underwater.

amanda-tarafa:
I'm sorry, but my focus is elsewhere at the moment.

csharp/src/Google.Protobuf.Test/MessageExtensionsTest.cs
```diff
@@ -0,0 +1,22 @@
+using System.IO;
+using Google.Protobuf.Buffers;
+using Google.Protobuf.TestProtos;
+using NUnit.Framework;
+
+namespace Google.Protobuf;
+
+public class MessageExtensionsTest
+{
```
esrauchg, line 9: Do you mind doing a serialize in a different implementation and having a test that parses that byte[] here, just for the most superficial of check that the format actually does match the intended format (just round tripping only confirms that C# matches itself)

csharp/src/Google.Protobuf/MessageExtensions.cs
```diff
@@ -157,6 +157,23 @@ public static void WriteTo(this IMessage message, IBufferWriter<byte> output)
             ctx.Flush();
         }
 
+        /// <summary>
+        /// Writes the length and then data of the given message to a buffer writer.
+        /// </summary>
+        /// <param name="message">The message to write.</param>
+        /// <param name="output">The buffer writer.</param>
+        [SecuritySafeCritical]
+        public static void WriteDelimitedTo(this IMessage message, IBufferWriter<byte> output)
```
esrauchg: This name does match our older implementations, but we've recently decided this should be called 'LengthPrefixed' format (the name was actually very confusing because the group wire format for messages has a start/end delimiter and we call that delimited, contrasted with this format).

Can you change the name to WriteLengthPrefixedTo() and add a comment that explains that this is the format referred to as 'Delimited' format in JavaProto and C++Proto

For context, here is the name used in upb which is the freshest case that we had this topic:
https://github.com/protocolbuffers/protobuf/blob/main/upb/wire/decode.h#L106

csharp/src/Google.Protobuf.Test/MessageExtensionsTest.cs
verdie-g: > in a different implementation 

What do you mean by that?

> having a test that parses that byte[] here

Should I use the CodedInputStream API for that?

esrauchg: I only mean it would be good to grab some byte[] from any other preexisting path for this format; exactly checking the round trip is a good unit test but it only confirms its self consistent and not aligned with the format that anyone else would parse.

verdie-g: If I use a byte[], that would mean I would test the delimited read right?

verdie-g: I can compare the WriteLengthPrefixedTo output with a byte array if you'd like

esrauchg: Thanks!

verdie-g:
@esrauchg anything missing to get this merged?

esrauchg (approved):

esrauchg:
Sorry, this got hung up on CI failures. I kicked the import and it should resolve now.

Since we've already cut our first Q1 release's branch for the RC, we would plan to only release this with Q2 release.

Thanks for the contribution!

PR #21355, opened 2025-04-18
Automated rollback of commit f9863dfd3c98278e703be0e3c19ded2d09e60fde.

zhangskz:
PiperOrigin-RevId: 749053436

### DIFF
--- a/php/ext/google/protobuf/message.c
+++ b/php/ext/google/protobuf/message.c
@@ -65,7 +65,6 @@ static zend_object* Message_create(zend_class_entry* class_type) {
   Message_SuppressDefaultProperties(class_type);
   zend_object_std_init(&intern->std, class_type);
   intern->std.handlers = &message_object_handlers;
-  intern->desc = NULL;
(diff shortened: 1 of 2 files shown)


bshaffer:
How was this missed when the PR was approved?

zhangskz:
The test that caught this was a continuous-only test, so didn't run on presubmit. Adding #test-continuous to the PR description would force continuous-only tests to run on presubmit anyways.

On our side, we noticed this was failing a bit late, but these run hourly so we def could've caught this earlier.

haberman (approved):

bshaffer:
@zhangskz so would the next steps be for me to try and debug the failing continuous tests and then resubmit the PR?

zhangskz:
> @zhangskz so would the next steps be for me to try and debug the failing continuous tests and then resubmit the PR?

Yes that would be great -- I also updated and reopened https://github.com/protocolbuffers/protobuf/issues/19978 to let the original PR author know.

PR #21363, opened 2025-04-19
C# code generation optimization

crmarsh:
I've been looking into the C# code generation for a Unity game I'm working on, and ended up making a bunch of changes

- strip out features we don't need
- simplify the generated code
- add in Unity serialize field attributes
- improve the hash functions
- support enabling nullable fields

These changes are all off by default, and many/most of them break the unit tests and such if you enable them, because of course, the unit tests test all the features and we’re deleting them. With all of these things enabled, though, we save about 20% on the generated code size in our project.

I can split this out into smaller pull requests to make it easier to review, but before I put effort into that, I wanted to put this out there to ask if this is even in the direction of the kind of improvements you might accept?


zhangskz:
Thanks for the contribution! Unfortunately stripping out features, simplifications to gencode, and Unity serialize field attributes are changes we're unlikely to be able to accept.

Improvements to our hash function is something we could potentially accept, but our preference there would likely be to update hashing to be consistent with our Java hash implementation. 

Nullability is something we've considered in the past but have intentionally chosen not to support due to the intended behavior of default values. This is further described in https://protobuf.dev/design-decisions/nullable-getters-setters/.

crmarsh:
Thanks for getting back to me -- I'll close this out.

It looks like the Java hash function is basically the same as what I did, but has some different numbers, I'll see about making a PR with just that change.

I read that article, but we only use proto3, so the problem with nullability outlined there doesn't make sense for us.

Issue #18782, opened 2024-10-10
Add C# codegen for Clear message to reuse message objects in a pool


batesj:
**What language does this apply to?**
C# codegen.

**Describe the problem you are trying to solve.**
Using protobufs in Unity can cause excessive garbage collection overhead because of all the new class instances created while parsing a proto into a message instance.

**Describe the solution you'd like**
I would like to be able to reuse a pool of message instances repeatedly for parsing protos.

**Describe alternatives you've considered**
I could not find examples, but perhaps there is a way to do it with reflection? Of course, if that's expensive then it is replacing garbage with CPU overhead.

ifshuaishuai:
In game developing area, especially in unity, recycle message object from pool is a must-have. I'm current looking for same Clear() method like PB did for Java for recycling.

danielhedren:
This is a much-wanted feature in the context of real time bidding as well since we're normally parsing and creating thousands of messages per second.

PR #21472, opened 2025-04-26
[C#] Adding Clear mechanism to IMessage

ifshuaishuai:
# Motivation
In C# game development or real-time bidding, object pooling is a common way to optimize. However, the C# runtime of Protobuf does not support Clear message. See issue: #18782.

This pull request add Clear method to IMessage.cs, makes Protobuf C# message can be reused.

# What's added or changed
* Protoc: csharp codegen add Clear() method, inspired of java clear() implementation.
* WellKnown protos: Regenerated by Protoc tools above.
* IMessage: interface add Clear() method, few old proto generated file or C# file was correct by hand.
* Tests: base unit tests are included.

If this pull request has any breaking changes or improvement, please let me know. And I will try my best to improve it.

Thanks for your time.

### DIFF
--- a/csharp/src/AddressBook/Addressbook.pb.cs
+++ b/csharp/src/AddressBook/Addressbook.pb.cs
@@ -99,6 +99,21 @@ namespace Google.Protobuf.Examples.AddressBook {
       return new Person(this);
     }
 
+    [global::System.Diagnostics.DebuggerNonUserCodeAttribute]
+    [global::System.CodeDom.Compiler.GeneratedCode("protoc", null)]
+    public void Clear() {
+      name_ = "";
+      id_ = 0;
+      email_ = "";
+      phones_.Clear();
+      if (lastUpdated_ != null) {
+        lastUpdated_.Clear();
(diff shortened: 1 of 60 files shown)


jskeet:
Removed myself as a reviewer as I don't have time to consider this thoroughly, but if I *were* reviewing it I would reject it in its current state as adding a method to an interface is a breaking change.

esrauchg:
I think Jon's comment is the decision here; we unfortunately cannot accept a change that adds a method to IMessage because it would be a breaking change.

It would be possible for us to accept the change to add the Clear method on generated classes though as that shouldn't be a breaking change, but its not clear from the context if that would satisfy your usecase.

verdie-g:
@ifshuaishuai maybe you could introduce a new interface to avoid the breaking change
```csharp
public interface IClearableBufferMessage : IBufferMessage
{
    void Clear();
}
```

googleberg:
As decided: adding a method to an interface is a breaking change.

PR #21497, opened 2025-04-29
feat: Reserve RepeatedFields when read packed elements

chenyahui:


acozzette:
I don't think it is safe for us to do this optimization unfortunately. If we did this then a very small message would be able to trick the parser into allocating up to 2 GiB of memory.

chenyahui:
If security is the concern, we could at least provide an ​option​ to enable this optimization when the user can guarantee safety. After all, this feature can significantly improve performance when dealing with very long repeated arrays.

acozzette:
Sorry but I don't think we want to go in that direction. Generally we are very reluctant to add any new options because they increase our support burden and the number of configurations we have to test and maintain.

Actually the current parser does not call `WireFormatLite::ReadPackedPrimitive()` anymore, so I believe this is just a vestige of the way the parser used to work.

PR #21416, opened 2025-04-22
Patch `rules_ruby` to apply `neverlink = True` to the `jars` rule

JasonLunn:
Fixes #21369

PR #21505, opened 2025-04-29
Patch `rules_ruby` to apply `neverlink = True` to the `jars` rule (#21416)

JasonLunn:


MODULE.bazel
```diff
@@ -67,6 +67,7 @@ single_version_override(
     patch_strip = 1,
     patches = [
         "@com_google_protobuf//:Disable_bundle_install.patch",
+        "@com_google_protobuf//:Neverlink_jruby_jars.patch",
```
zhangskz, line 70: Nit: Can we add a TODO to clean this up once https://github.com/bazel-contrib/rules_ruby/pull/226 is released?

JasonLunn: Ack; I'll add that into my PR that moves these patches into the `patches` directory

PR #21567, opened 2025-05-02
Generate .pyi files in py_proto_library (#10366)

### LINKED ISSUE #10366: Support pyi_out in py_proto_library
Recently, the `--pyi_out` flag was added to protoc, which is really great. I have a project where protos are compiled using bazel/py_proto_library. From everything that I have seen, there is no ability to additionally generate the `.pyi` files using this mode.

* is it currently possible to generate `.pyi` files using `py_proto_library`? if so, how?
* if not, what would be a path to add this support? I'm happy to make a PR (with maybe a few pointers)

thanks!

dws:
We here adjust py_proto_library to produce `.pyi` files along with the `.py` files it already generates.  This achieves the same effect as the gRPC py_proto_library.

Closes #10366.

### DIFF
--- a/bazel/py_proto_library.bzl
+++ b/bazel/py_proto_library.bzl
@@ -61,6 +61,7 @@ def _py_proto_aspect_impl(target, ctx):
     api_deps = [proto_lang_toolchain_info.runtime]
 
     generated_sources = []
+    generated_stubs = []
     proto_info = target[ProtoInfo]
     proto_root = proto_info.proto_source_root
     if proto_info.direct_sources:
@@ -69,6 +70,14 @@ def _py_proto_aspect_impl(target, ctx):
             actions = ctx.actions,
             proto_info = proto_info,
             extension = "_pb2.py",
+            name_mapper = lambda name: name.replace("-", "_").replace(".", "/"),
+        )
+
+        # Generate pyi files
+        generated_stubs = proto_common.declare_generated_files(
+            actions = ctx.actions,
+            proto_info = proto_info,
+            extension = "_pb2.pyi",
             name_mapper = lambda name: name.replace("-", "_").replace(".", "/"),
         )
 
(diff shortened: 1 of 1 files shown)


jasonkuster:
+1; with `grpc` moving to using protobuf as the default for py_proto_library it regresses the creation of pyi files; this is the approach that repo takes and it would be very appreciated to get something like this merged.

dws:
@anandolee Thanks for adding the tag!  It looks like my PR passes CI, so I look forward to your review.

bazel/py_proto_library.bzl
```diff
@@ -84,16 +93,21 @@ def _py_proto_aspect_impl(target, ctx):
...
             proto_lang_toolchain_info = proto_lang_toolchain_info,
-            generated_files = generated_sources,
+            generated_files = generated_sources + generated_stubs,
             plugin_output = plugin_output,
+            additional_args = additional_args,
         )
 
     # Generated sources == Python sources
-    python_sources = generated_sources
+    python_sources = generated_sources + generated_stubs
```
mkruskal-google, line 110: We discussed this with some python folks, and while we agree this PR looks like a good approach, it needs a few tweaks.  Specifically, the pyi files should end up in the new `direct_pyi_files` and `transitive_pyi_files` fields of `PyInfo` added in rules_python 1.1, instead of getting globbed together into `transitive_sources` with the py files like is currently happening

I think this is just a matter of piping the information through _PyProtoInfo, and should be a relatively minor tweak to your PR though

dws: @mkruskal-google Thank you for your feedback.  At present, it looks like protobuf requires rules_python 1.0.0.  Are you asking for me to bump that to rules_python 1.1.0 as part of this PR or possibly a separate PR?  Or should I attempt to be compatible with rules_python 1.0.0 but take advantage of the new `PyInfo` fields if they are available?

mkruskal-google: Those are the two options :). I think bumping to 1.1 should be fine as it's a minor version release, and it would be a lot easier.  We could go the extra mile to support 1.0.0 by checking for the presence of the new fields, but I don't think it's necessary here

dws: @mkruskal-google I've put up https://github.com/protocolbuffers/protobuf/pull/22079 to bump the minimum `rules_python` version to 1.1.0.

dws: @mkruskal-google I've added a second commit here which I believe does what you requested. I'm assuming here that part of the intent was that the `.pyi` files not be automatically part of the runfiles.  This differs from the behavior of the gRPC version of `py_proto_library`.  For drop-in compatibility with the gRPC version of `py_proto_library` it would be easy enough to also make the `.pyi` files automatically part of the runfiles.  If you want that, please let me know and I'll add it.

Since the rules_python version bump is in a separate PR (#22079), I'd suggest landing that before trying to do CI on this PR.

mkruskal-google: This lgtm as a starting point.  If grpc requires further tweaks we can make those separately I think

dws:
@mkruskal-google Now that the rules_python bump has landed, please try another round of CI here.

dws:
@mkruskal-google ping?

dws:
@mkruskal-google ping?

dws:
@anandolee perhaps you could enable a new round of CI testing?  I think that's all that should be needed to land this now that it is approved.

Issue #21590, opened 2025-05-05
Rust protobuf 4.31.0-rc.2 passing incorrect path to compiler


windows-fryer:
**What version of protobuf and what language are you using?**
Version: v31.0-rc2
Language: Rust

**What operating system (Linux, Windows, ...) and version?**
Windows 11 24H2 (OS Build 26100.3775)

**What runtime / compiler are you using (e.g., python version or gcc version)**
rustc 1.88.0-nightly (13e879094 2025-05-04)

target x86_64-pc-windows-gnu:
gcc.exe (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders, r3) 14.2.0

target x86_64-pc-windows-msvc:
Microsoft (R) C/C++ Optimizing Compiler Version 19.43.34810 for x64

**What did you do?**
1. Setup a project with protobuf(-...) = "4.31.0-rc.2"
2. Attempt to compile
3. "upb/generated_code_support.h: No such file or directory"

**What did you expect to see**
A correct path being passed to GCC.

**What did you see instead?**
"\\\\?\\C:\\Users\\user\\.cargo\\registry\\src\\index.crates.io-1949cf8c6b5b557f\\protobuf-4.31.0-rc.2\\libupb"

Upon manually copying the command and removing the incorrect path, it will execute successfully. Additionally, modifying the output path of the codegen and inserting the dependency there will also allow for a successful, yet ugly, compilation.

**Anything else we should know about your project / environment**
I am working in a workspace on a nightly version of rust. I believe the issue stems [from here although](https://github.com/protocolbuffers/protobuf/blob/f63ed55d4b85ad3cd558057e178c035e70c02c77/src/google/protobuf/io/io_win32.cc#L227)

Additionally, when using both MSVC and GNU targets the compilation will still fail.

windows-fryer:
@acozzette Looks like I believe I've traced the source of the issue. When compiling on Windows it attempts to use extended path syntax due to the usage of [std::fs::canonicalize](https://github.com/protocolbuffers/protobuf/blob/main/rust/release_crates/protobuf/build.rs#L23). The official documentation of canonicalize states: "... and it may be incompatible with other applications (if passed to the application on the command-line, or written to a file another application may read)."

Here's a quick diff for a patch to it:
```diff
diff --git a/rust/release_crates/protobuf/build.rs b/rust/release_crates/protobuf/build.rs
index 5351880e9..99aada3ed 100644
[...]

PR #21606, opened 2025-05-05
Fix Windows path handling in build script

windows-fryer:
PR for patch mentioned in #21590

### DIFF
--- a/rust/release_crates/protobuf/build.rs
+++ b/rust/release_crates/protobuf/build.rs
@@ -5,18 +5,7 @@
 // license that can be found in the LICENSE file or at
 // https://developers.google.com/open-source/licenses/bsd
 
-use std::path::Path;
-
 const VERSION: &str = env!("CARGO_PKG_VERSION");
-
-fn clean_path(path: &Path) -> String {
-    let canonical = path.canonicalize().unwrap();
-    let path_str = canonical.to_string_lossy();
-    path_str
-        .strip_prefix(r"\\?\")
-        .unwrap_or(&path_str)
-        .to_string()
-}
 
 fn main() {
     cc::Build::new()
@@ -30,7 +19,9 @@ fn main() {
         .file("libupb/third_party/utf8_range/utf8_range.c")
         .define("UPB_BUILD_API", Some("1"))
         .compile("libupb");
-    let path = std::path::Path::new("libupb");
-    println!("cargo:include={}", clean_path(&path));
+
+    let path = std::path::absolute("libupb").expect("Failed to get resolve path libupb");
+
+    println!("cargo:include={}", path.display());
     println!("cargo:version={}", VERSION);
 }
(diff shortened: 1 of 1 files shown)


rust/release_crates/protobuf/build.rs
```diff
@@ -20,6 +31,6 @@ fn main() {
         .define("UPB_BUILD_API", Some("1"))
         .compile("libupb");
     let path = std::path::Path::new("libupb");
-    println!("cargo:include={}", path.canonicalize().unwrap().display());
+    println!("cargo:include={}", clean_path(&path));
```
acozzette, line 34: Would [std::path::absolute](https://doc.rust-lang.org/std/path/fn.absolute.html) work instead of canonicalize? That might be a simpler solution.

acozzette (changes requested):

rust/release_crates/protobuf/build.rs
windows-fryer: Just tested absolute and it looks like it would work much better. I'll make a commit reflecting this.

windows-fryer: Didn't realize until now, but it seems the version of Rust being used for CI is quite old (1.74). For a quick workaround to make it pass, I've added `#![feature(cfg_version)]`.

windows-fryer: Looks like I forgot stable can't use feature flags. I'll look into a better solution in a little bit, I'll more than likely revert to canonicalize and strip the path. I've read that stripping from canonicalize is improper and can cause issues due to sim-links being resolved. 

Another approach is to use an [external dependency](https://crates.io/crates/path-absolutize) that works on Rust v1.60 although I'd rather find a solution in the Rust STD.

acozzette: I think we might actually be OK with increasing our minimum supported Rust version, so why don't I see if I can do that and upgrade the version we're using on CI.

windows-fryer: Sounds good, thank you very much!

acozzette: I upgraded us to 1.79 and I'm now testing your change [here](https://github.com/protocolbuffers/protobuf/pull/21616), so hopefully this will work.

windows-fryer: Hopefully! I am a little bit suspicious that the feature flag might break it but I am hopeful.

acozzette: Actually I removed the feature flag since we shouldn't need that on 1.79.

windows-fryer: Awesome, thank you very much 🤝

acozzette:
@windows-fryer Thank you for this fix. I just submitted it and will cherry-pick it to 31.x so that it makes it into the 31.0 release.

PR #21743, opened 2025-05-14
Cherry-pick fixes to 31.x

shaod2:


zhangskz (approved):

zhangskz:
You should merge without squashing for these cherry-picks since there are multiple to ensure these end up as separate entries in the commit history / release notes

PR #21754, opened 2025-05-15
Change pre-22 poison pill to only log once per affected message type.

esrauchg:


shaod2 (approved):

java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
```diff
@@ -329,12 +331,23 @@ public int getSerializedSize() {
...
 
-  static void warnPre22Gencode() {
+  private static final Set<String> loggedPre22TypeNames
+      = Collections.synchronizedSet(new HashSet<String>());
+  static void warnPre22Gencode(Class<?> messageClass) {
     if (System.getProperty(PRE22_GENCODE_SILENCE_PROPERTY) != null) {
       return;
     }
+    String messageName = messageClass.getName();
+    if (loggedPre22TypeNames.contains(messageName)) {
```
zhangskz, line 341: I think technically contains + add should normally be synchronized even though the individual operations are synchronized already (although extra adds should be ok in practice)
https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedSet-java.util.Set-
https://stackoverflow.com/questions/2158931/synchronizedcollection-and-contains-do-i-need-to-synchronize-manually

Also, can we avoid checking this for the PRE22_GENCODE_ERROR_PROPERTY case?

esrauchg: > I think technically contains + add should normally be synchronized even though the individual operations are synchronized already (although extra adds should be ok in practice)

add() actually returns whether it existed already or not, so I updated this to just use add() and then check the return value to avoid the split.

> Also, can we avoid checking this for the PRE22_GENCODE_ERROR_PROPERTY case?

Done.

I also updated it to be clearer that the log case doesn't actually have a stacktrace in it (exception.toString() doesn't include the stack trace), which I think is OK now that we're explicitly putting the type name in there. Can explicitly build the stack trace for the log if we think its important enough.

java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
```diff
@@ -329,21 +331,30 @@ public int getSerializedSize() {
...
+  private static final Set<String> loggedPre22TypeNames
+      = Collections.synchronizedSet(new HashSet<String>());
+  static void warnPre22Gencode(Class<?> messageClass) {
     if (System.getProperty(PRE22_GENCODE_SILENCE_PROPERTY) != null) {
       return;
     }
-    UnsupportedOperationException exception =
-        new UnsupportedOperationException(PRE22_GENCODE_VULNERABILITY_MESSAGE);
+    String vulnerabilityMessage =
+          "Vulnerable Protobuf Generated type in use: " + messageName + "\n" +
```
zhangskz, line 341: Supernit: Vulnerable protobuf generated type in use?

(idk why but the current capitalization bothers me :P)

java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
zhangskz: nice :)

zhangskz (approved):

java/core/src/main/java/com/google/protobuf/GeneratedMessage.java
esrauchg: Fixed!

noelportillo (commented):
I don't think

Issue #20540, opened 2025-03-02
Suggested improvement for CMake: Specify C++ version


OgreTransporter:
Abseil requires a C++17 compiler, so protobuf can only be compiled with C++17 or higher.

Therefore my suggestion: After line 15
https://github.com/protocolbuffers/protobuf/blob/24830cf07e559e912fa2622d3487a679ef0df9a9/CMakeLists.txt#L15

insert the following two lines:
```
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

acozzette:
Thank you for the suggestion. Abseil actually [recommends](https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md#incorporating-abseil-into-a-cmake-project) that libraries should leave `CMAKE_CXX_STANDARD` unset and instead use `target_compile_features()` to specify a minimum C++ standard. But now that I look at this, I see we are still setting C++14 as our minimum, so we need to update this. I will send out a change to fix this shortly.

PR #21803, opened 2025-05-17
Enable C++17 on all platforms

cyyever:


mkruskal-google:
IIUC this will force all downstream projects that include our cmake config to C++17, which is not what we want.  Support a *minimum* of C++17, and AFAIK CMake doesn't support specifying a constraint like that

cyyever:
@mkruskal-google Fixed by defaulting to 17.

mkruskal-google:
hmm I guess that works, thanks!

CMakeLists.txt
```diff
@@ -196,7 +202,6 @@ if (NOT MSVC)
   if (NOT protobuf_HAVE_BUILTIN_ATOMICS)
     set(protobuf_LINK_LIBATOMIC true)
   endif (NOT protobuf_HAVE_BUILTIN_ATOMICS)
-  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
 endif (NOT MSVC)
```
mkruskal-google, line 205: Looking at the history, I think this entire atomic check comes from when we supported pre-C++11 and is severely outdated.    It seems safe to assume that std::atomic is present in C++17 and newer, feel like some opportunistic cleanup?

mkruskal-google: Otherwise I'd ask that you leave the tech-debt to make it clear that this can be removed.  The lines you removed were what made me think of it

CMakeLists.txt
```diff
@@ -13,6 +13,14 @@ endif()
 
 # Project
 project(protobuf C CXX)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_EXTENSIONS OFF)
```
mkruskal-google, line 18: We probably shouldn't blindly override this one if someone has set it

CMakeLists.txt
```diff
@@ -13,6 +13,14 @@ endif()
 
 # Project
 project(protobuf C CXX)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
mkruskal-google: This came up previously in https://github.com/protocolbuffers/protobuf/issues/20540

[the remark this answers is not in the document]
cyyever: OK, let's avoid the global variables and set target properties.

cyyever:
@mkruskal-google Now only libprotobuf* is guaranteed C++17.

CMakeLists.txt
cyyever: I have removed this check.

CMakeLists.txt
cyyever: Reverted

cmake/libprotobuf.cmake
```diff
@@ -46,6 +43,10 @@ set_target_properties(libprotobuf PROPERTIES
     CXX_VISIBILITY_PRESET hidden
     VISIBILITY_INLINES_HIDDEN ON
 )
+get_target_property(OLD_CXX_STANDARD libprotobuf CXX_STANDARD)
+if(OLD_CXX_STANDARD LESS 17)
+  set_target_properties(libprotobuf PROPERTIES CXX_STANDARD 17)
```
mkruskal-google, line 48: I think this might be redundant.  We already set `target_compile_features(* PUBLIC cxx_std_17)` [here](https://github.com/protocolbuffers/protobuf/blob/47183273b6fc5ce931888f852dd2a02a7dfaeaef/cmake/protobuf-configure-target.cmake#L10)

If it's not redundant, you could reduce duplication by putting it there.  But I'm not sure what the difference between these things is.  And how would someone else set the CXX_STANDARD property on *our* targets?

CMakeLists.txt
```diff
@@ -180,25 +180,6 @@ if (protobuf_WITH_ZLIB)
   endif (ZLIB_FOUND)
 endif (protobuf_WITH_ZLIB)
```
mkruskal-google, line 182: I think the error message you added to this file made a lot of sense.  If someone tries to run this config with <17 we should error quickly and early with a clear error message

cmake/libprotobuf.cmake
mkruskal-google: https://cliutils.gitlab.io/modern-cmake/chapters/features/cpp11.html seems to explain the difference between these 3 options pretty well.  I think our current setup is what we want, except maybe we'd want to mark it `PRIVATE`

CMakeLists.txt
cyyever: Added

cmake/libprotobuf.cmake
cyyever: They shouldn't be PRIVATE because C++17 syntax is leaked in headers.

cmake/protobuf-configure-target.cmake
```diff
@@ -7,7 +7,7 @@
 
 # Refactors configuration options set on all Protobuf targets
 function(protobuf_configure_target target)
-    target_compile_features("${target}" PUBLIC cxx_std_17)
+    target_compile_features("${target}" PRIVATE cxx_std_17)
```
mkruskal-google, line 10: I actually think this needs to be PUBLIC for now.  We make no effort to hide C++17 features in our internals, so our headers almost for sure require C++17.  This means anyone who uses protobuf libraries should have this feature too.  We could remove this from binaries, but I'm not sure private vs public even really matters there

mkruskal-google: Ah nvm I saw your other comment, good point

cmake/libprotobuf.cmake
mkruskal-google: https://cmake.org/cmake/help/v3.20/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#high-level-meta-features-indicating-c-standard-support - It appears these are actually *minimums*, not strict settings.  So I think it does make more sense to keep these PUBLIC to make it transitive (see my last comment I resolved)

cmake/protobuf-configure-target.cmake
cyyever: Fine, revert it.

mkruskal-google:
Thanks for the contribution and iteration!

PR #21805, opened 2025-05-17
Update .NET 6 to .NET 8

henric-root:
.NET 6 has been out of support since October 2024.  This changes the build to use .NET 8 (the current LTS).

The .NET 5 target (not supported since May 2022) was left in.  I'm not sure what the support policy is, but it seemed reasonable to include at least one .NET target that is supported by Microsoft.  (One might also observe that the .NET Standard 2.0 target could be considered enough to let those using older versions of .NET consume Google.Protobuf.)

### DIFF
--- a/csharp/src/AddressBook/AddressBook.csproj
+++ b/csharp/src/AddressBook/AddressBook.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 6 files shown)


jskeet:
We do want to do this, but we'll need to check that our internal build systems are able to cope with it - and I personally don't have the bandwidth for this at the moment.

googleberg:
Will need internal testing and team bandwidth.

PR #21811, opened 2025-05-17
fix(php): use count instead of ->count() to avoid bug in c-extension

bshaffer:
see https://github.com/protocolbuffers/protobuf/issues/21812

Calling `setX($val)` on a deprecated repeated field or map when the php extension is enabled results in a PHP error:

```
PHP Error: Call to a member function count() on array
```

This is because the c-extension seems to have different behavior than the PHP native library - the return type of `GPBUtil::checkRepeatedField` is of type `RepeatedField` in the native library, but in the extension it is `array`.

This should be fixed in the PHP extension as well, but this PR only fixes the issue in the gencode.

### DIFF
--- a/php/tests/GeneratedClassTest.php
+++ b/php/tests/GeneratedClassTest.php
@@ -130,6 +130,34 @@ class GeneratedClassTest extends TestBase
         $this->assertEquals(0, $deprecationCount);
     }
 
+    public function testDeprecatedFieldSetterThrowsWarning()
+    {
+        // temporarily change error handler to capture the deprecated errors
+        $deprecationCount = 0;
+        set_error_handler(function ($errno, $errstr) use (&$deprecationCount) {
+            if (false !== strpos($errstr, ' is deprecated.')) {
+                $deprecationCount++;
+            }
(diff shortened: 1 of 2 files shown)


mkruskal-google (approved):

php/tests/GeneratedClassTest.php
```diff
@@ -130,6 +130,27 @@ public function testDeprecatedFieldGetterDoesNotThrowWarning()
...
+
+
+        // This behavior exists because otherwise the deprecation is thrown on serializeToJsonString
+        $message = new TestMessage();
+        $message->setDeprecatedRepeatedInt32([]); // repeated field
+        $message->setDeprecatedMapInt32Int32([]); // map field
+
+        restore_error_handler();
+
+        $this->assertEquals(2, $deprecationCount);
```
mkruskal-google, line 151: This is failing some tests, not sure why this would be platform dependent tho...

mkruskal-google (approved):

bshaffer:
This was handled in Piper ID 761147122

PR #21839, opened 2025-05-20
Use utf-8 by default

cyyever:


cyyever:
@mkruskal-google Just found it.

cmake/protobuf-configure-target.cmake
```diff
@@ -32,9 +32,7 @@ function(protobuf_configure_target target)
         # Allow big object
         target_compile_options("${target}" PRIVATE /bigobj)
     endif ()
-    if (protobuf_UNICODE)
-        target_compile_definitions("${target}" PRIVATE -DUNICODE -D_UNICODE)
-    endif ()
+    target_compile_definitions("${target}" PRIVATE -DUNICODE -D_UNICODE)
```
mkruskal-google, line 35: This will affect all platforms, is this useful anywhere except MSVC?

appveyor.bat
```diff
@@ -12,7 +12,7 @@ echo Building MinGW
 set PATH=C:\mingw-w64\x86_64-7.2.0-posix-seh-rt_v5-rev1\mingw64\bin;%PATH:C:\Program Files\Git\usr\bin;=%
 mkdir build_mingw
 cd build_mingw
-cmake -G "%generator%" -Dprotobuf_BUILD_SHARED_LIBS=%BUILD_DLL% -Dprotobuf_UNICODE=%UNICODE% -Dprotobuf_BUILD_TESTS=0 ../cmake
+cmake -G "%generator%" -Dprotobuf_BUILD_SHARED_LIBS=%BUILD_DLL% -Dprotobuf_BUILD_TESTS=0 ../cmake
```
mkruskal-google, line 15: Is this just cleanup?  Or is this a meaningful change?

cyyever: cleanup

cmake/protobuf-configure-target.cmake
cyyever: This two macros are used in Windows C runtime. See 
```
https://learn.microsoft.com/en-us/cpp/text/unicode-programming-summary?view=msvc-170
```

appveyor.bat
mkruskal-google: I didn't even know this file existed tbh, and I'm not familiar with what it's meant for.  When is %UNICODE% set?  I assume that's some environment variable users might have unset?

cmake/protobuf-configure-target.cmake
mkruskal-google: makes sense, we should guard this on MSVC is we're not going to guard it on protobuf_UNICODE anymore then.  Otherwise this will be a newly added macro for linux and macos that could cause issues

appveyor.bat
cyyever: If appveyor is not activated in CI, it is safe to remove it.

cyyever: And `protobuf_UNICODE` is not documented, so it should be safe to ignore.

mkruskal-google: Our cmake documentation is pretty poor, *a lot* of people use our cmake configs as "documentation" and figure out flags from there.  So if this is a behavior change it's not quite "safe"

PR #21858, opened 2025-05-20
Cherrypick Pure Python recursion limit enforcement to 29.x

shaod2:


zhangskz (approved):

zhangskz:
Merge commit is fine since the set of commits is clean, but if there ends up being other commits (e.g. merging main in to "rebase", fixes to the PR) please update the PR description and squash instead to keep the release notes sane :)

PR #21864, opened 2025-05-21
Use std::optional

cyyever:
Use std::optional since latest ABSL has declared that abs::optional is an alias of it.

### DIFF
--- a/src/google/protobuf/io/test_zero_copy_stream_test.cc
+++ b/src/google/protobuf/io/test_zero_copy_stream_test.cc
@@ -7,14 +7,14 @@
 
 #include "google/protobuf/io/test_zero_copy_stream.h"
 
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
-
-#include <gmock/gmock.h>
-#include <gtest/gtest.h>
-#include "absl/types/optional.h"
 
 namespace google {
 namespace protobuf {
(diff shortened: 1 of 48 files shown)


src/google/protobuf/io/test_zero_copy_stream_test.cc
```diff
@@ -7,15 +7,15 @@
 
 #include "google/protobuf/io/test_zero_copy_stream.h"
 
+#include <gmock/gmock.h>
```
mkruskal-google, line 10: Note: these header reorderings will cause merge conflicts and make our import super painful, try to avoid them if you can :)

src/google/protobuf/json/internal/parser.cc
```diff
@@ -7,12 +7,11 @@
 
 #include "google/protobuf/json/internal/parser.h"
 
-#include <cfloat>
```
mkruskal-google, line 10: Similarly, header removals might *seem* like cleanups, but we have hidden internal code that could be using this.  So this might just end up making our merges messier

mkruskal-google (approved):

cyyever:
@mkruskal-google Any guideline for sorting headers?

mkruskal-google:
Don't worry about it for this PR, but in the future just try to avoid touching headers as much as possible.  Our sync script doesn't handle them well at all and they usually require manual intervention if you make any changes to them.  For this PR I'd have limited it to simply adding `<optional>` and removing `"absl/types/options.h"`

src/google/protobuf/message.h
```diff
@@ -632,7 +632,7 @@ class PROTOBUF_EXPORT Reflection final {
     friend class Reflection;
 
     absl::string_view CopyFromCord(const absl::Cord& cord) {
-      if (absl::optional<absl::string_view> flat = cord.TryFlat()) {
+      if (std::optional<absl::string_view> flat = cord.TryFlat()) {
```
mkruskal-google, line 635: hmm our cmake tests are failing here because absl::Cord::TryFlat is returning an absl::optional.  We're likely using an older version of Abseil in our CI, so I'd have to double check what our compatibility window is.  Either we can upgrade that and push this through, or we'll have to wait until we drop support for that version of Abseil

mkruskal-google: clever, if this is the only instance I think I can live with `auto` :)

mkruskal-google (approved):

PR #21867, opened 2025-05-21
Remove absl::memory

cyyever:


src/google/protobuf/compiler/python/generator.cc
```diff
@@ -32,7 +32,6 @@
 #include "absl/container/flat_hash_set.h"
 #include "absl/log/absl_check.h"
 #include "absl/log/absl_log.h"
-#include "absl/memory/memory.h"
```
mkruskal-google, line 35: We use both absl::WrapUnique and absl::make_unique in a bunch of places, I'm surprised this builds...

cyyever: It builds

src/google/protobuf/feature_resolver.cc
```diff
@@ -465,7 +464,7 @@ absl::StatusOr<FeatureSetDefaults> FeatureResolver::CompileDefaults(
   FeatureSetDefaults defaults;
   defaults.set_minimum_edition(minimum_edition);
   defaults.set_maximum_edition(maximum_edition);
-  auto message_factory = absl::make_unique<DynamicMessageFactory>();
+  auto message_factory = std::make_unique<DynamicMessageFactory>();
   for (const auto& edition : editions) {
     auto fixed_defaults_dynamic =
         absl::WrapUnique(message_factory->GetPrototype(feature_set)->New());
```
mkruskal-google, line 470: This one still uses memory.h

PR #21873, opened 2025-05-21
Fix expected parse failures that may fail for the wrong reasons

hudlow:
These expected JSON parse failures will currently fail for reasons _other_ than the tested behavior.

Note that I'm being a little cute here by fixing the tests that [now apparently contradict the spec](https://github.com/protocolbuffers/protocolbuffers.github.io/commit/c18c8a72de22b99eaf536eb1f961371510159714#diff-98c94e5ab37db71fd002eff0d9cb2cfae9b0058ea00b412b466312ad819029d7). I happen to think the spec is misguided and these tests are better behavior, but I _expect_ the tests will end up being updated to match the spec.

### DIFF
--- a/conformance/binary_json_conformance_suite.cc
+++ b/conformance/binary_json_conformance_suite.cc
@@ -2289,17 +2289,17 @@ void BinaryAndJsonConformanceSuiteImpl<
   // Duplicated field names are not allowed.
   ExpectParseFailureForJson("FieldNameDuplicate", RECOMMENDED,
                             R"({
-        "optionalNestedMessage": {a: 1},
+        "optionalNestedMessage": {\"a\": 1},
         "optionalNestedMessage": {}
       })");
   ExpectParseFailureForJson("FieldNameDuplicateDifferentCasing1", RECOMMENDED,
                             R"({
-        "optional_nested_message": {a: 1},
+        "optional_nested_message": {\"a\": 1},
         "optionalNestedMessage": {}
       })");
   ExpectParseFailureForJson("FieldNameDuplicateDifferentCasing2", RECOMMENDED,
                             R"({
-        "optionalNestedMessage": {a: 1},
+        "optionalNestedMessage": {\"a\": 1},
         "optional_nested_message": {}
       })");
   // Serializers should use lowerCamelCase by default.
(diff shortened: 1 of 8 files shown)


mkruskal-google:
Thanks for catching this!  It appears we missed these conformance tests when we updated the spec, or may have misinterpreted the fact that we were "non-conformant".

Can you update the failure_list files for C++, Java and PHP?  It looks like you fixed conformance issues in C++ and Java (they follow the old spec of rejecting duplicate names), and caught a new issue in pure PHP

conformance/binary_json_conformance_suite.cc
```diff
@@ -2289,17 +2289,17 @@ void BinaryAndJsonConformanceSuiteImpl<
   // Duplicated field names are not allowed.
   ExpectParseFailureForJson("FieldNameDuplicate", RECOMMENDED,
                             R"({
-        "optionalNestedMessage": {a: 1},
+        "optionalNestedMessage": {\"a\": 1},
```
mkruskal-google, line 2292: nit: this is a raw string, you don't need to escape quotes

hudlow: Not a nit at all, as it turns out.

hudlow:
I've taken another pass at this, but I'm finding this repo to be incredibly difficult to work in.

hudlow:
@mkruskal-google I've been unable to find any clear instructions on how to build this project locally. I've put a significant amount of effort fixing various things that were broken or incomplete as I ran into them, but it seems I am not much closer to passing the builds.

Can you help me understand how I can be successful here?

mkruskal-google:
Don't worry about, I can fix these up on import.  We're still discussing what direction we want to go here, hopefully we'll have an update later this week

mkruskal-google:
Ok so we discussed this quite extensively.  We will likely add a bunch of conformance tests to cover new edge cases uncovered, and at the very least clarify the spec.

The core problem with the behavior you *want* is that for languages that leverage a builtin `JSON.parse` function, information about duplicate keys is largely lost.  These typically take the last instance of any key and drop all others, meaning that those languages can only realistically error in the case where the JSON key is spelled differently for the same field or for different fields in oneofs.  So to truly error on all duplicated keys we would need custom JSON parsers in every language, which can be expensive and error-prone (only go and python get close today).  Retaining the error in the cases where we can might still make sense though (e.g. oneofs and different spellings of the same field).

I'll accept this PR as a bug fix for now though, and we'll follow up with further changes

mkruskal-google (approved):

hudlow:
Thanks @mkruskal-google!

> The core problem with the behavior you want is that for languages that leverage a builtin JSON.parse function, information about duplicate keys is largely lost. These typically take the last instance of any key and drop all others, meaning that those languages can only realistically error in the case where the JSON key is spelled differently for the same field or for different fields in oneofs. So to truly error on all duplicated keys we would need custom JSON parsers in every language, which can be expensive and error-prone (only go and python get close today). Retaining the error in the cases where we can might still make sense though (e.g. oneofs and different spellings of the same field).

Yes, I am well aware; I'm trying to build a war chest of evidence that those languages need to provide stricter options in their native parsers. (I am rarely accused of fatalism.)

Issue #21938, opened 2025-05-24
Warning: comparison of integer expressions of different signedness


iksrandeb:
**What version of protobuf and what language are you using?**

Version: v31.0 (release) and main (0b411d068d52c1fedebb02671b3516ac9608ae1a)
Language: C++

**What operating system (Linux, Windows, ...) and version?**
OS Linux Ubuntu 24.04.2
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.2 LTS"
```

**What runtime / compiler are you using (e.g., python version or gcc version)**
- GCC
```
gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
- Bazelisk
```
Bazelisk version: v1.26.0
Build label: 8.2.1
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Thu Apr 17 18:22:55 2025 (1744914175)
Build timestamp: 1744914175
Build timestamp as int: 1744914175
```

**What did you do?**
1. Create MODULE.bazel with content:
```
module(
    name = "bug_report_v31",
    version = "0.0.1",
)

bazel_dep(name = "protobuf", version = "31.0")
```
2. Run the command:
`bazelisk run @protobuf//:protoc` or `bazel run @protobuf//:protoc`

**What did you expect to see**

No warnings during build.

**What did you see instead?**

```
bazelisk run @protobuf//:protoc
2025/05/24 09:25:07 Downloading https://releases.bazel.build/8.2.1/release/bazel-8.2.1-linux-x86_64...
Downloading: 60 MB out of 60 MB (100%) 
Extracting Bazel installation...
Starting local Bazel server (8.2.1) and connecting to it...
INFO: Analyzed target @@protobuf+//:protoc (125 packages loaded, 2483 targets configured).
INFO: From Compiling src/google/protobuf/compiler/java/internal_helpers.cc:
external/protobuf+/src/google/protobuf/compiler/java/internal_helpers.cc: In lambda function:
[...]

PR #21944, opened 2025-05-24
Resolve different signedness warning

iksrandeb:
The size() by default returns an unsigned integer datatype. In other places in this code a C cast style is used. Follow the same resolution, and cast the unsigned integer type to signed integer.

### DIFF
--- a/src/google/protobuf/compiler/java/internal_helpers.cc
+++ b/src/google/protobuf/compiler/java/internal_helpers.cc
@@ -177,7 +177,7 @@ void GenerateLarge(
         }},
        {"aliases",
         [&] {
-          for (int i = 0; i < (int)aliases.size(); i++) {
+          for (size_t i = 0; i < aliases.size(); i++) {
             WriteEnumValueDocComment(printer, aliases[i].first,
                                      context->options());
             printer->Emit({{"name", aliases[i].first->name()},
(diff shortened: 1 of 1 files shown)


iksrandeb:
Proposal for solving #21938

src/google/protobuf/compiler/java/internal_helpers.cc
```diff
@@ -177,7 +177,7 @@ void GenerateLarge(
         }},
        {"aliases",
         [&] {
-          for (int i = 0; i < aliases.size(); i++) {
+          for (int i = 0; i < (int)aliases.size(); i++) {
```
honglooker: how about `size_t i = 0;` without the cast?

iksrandeb: I totally agree with the proposal of `size_t`. However the rest of the code in this file uses a C cast and I just wanted to be consistent.

honglooker: @sbenzaquen wdyt?

iksrandeb: Using the suggested solution `size_t i = 0;` without the cast.

googleberg (approved):

PR #22010, opened 2025-05-29
Use absl 20250512

cyyever:
Update absl dependency to 20250512

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -12,7 +12,7 @@ endif()
 
 set(rules_apple-version "3.16.0")
 set(apple_support-version "1.15.1")
-set(abseil-cpp-version "20250127.0")
+set(abseil-cpp-version "20250512.0")
 set(rules_cc-version "0.0.17")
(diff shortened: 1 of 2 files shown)


honglooker:
Hi @cyyever could you explain why you'd like this version bump? Thanks

cyyever:
@honglooker Test the integration with lastest ABSL

mkruskal-google:
I believe the latest abseil actually doesn't work with gcc 7.5, and we can't bump our dependency.  They're working on a patch to fix the issue, but we'll see what the CI says

cmake/dependencies.cmake
```diff
@@ -12,7 +12,7 @@ endif()
 
 set(rules_apple-version "3.16.0")
 set(apple_support-version "1.15.1")
-set(abseil-cpp-version "20250127.0")
```
mkruskal-google, line 15: This file is generated from our MODULE.bazel file, our GHA just overwrites it before every test and will revert your commit immediately :)

PR #22171, opened 2025-06-09
[numpy] Fix breakage in users of Python protobufs under NumPy 2.3rc1.

mkruskal-google:
As of NumPy 2.3.0rc1, numpy.bool scalars can no longer be interpreted as index values (https://github.com/numpy/numpy/releases/tag/v2.3.0rc1). This causes protobuf no longer to accept a np.bool scalar as a legal value for a boolean field.

We have two options:
a) either we can change protobuf so that it continues to accept NumPy boolean scalars (this change), or b) decide that protobuf should reject NumPy boolean scalars and that users must update their code to cast to a Python bool explicitly.

I have no strong opinion as to which, but option (a) seems less disruptive.

No test updates are needed: the existing tests fail under NumPy 2.3.

PiperOrigin-RevId: 766629310

### DIFF
--- a/python/convert.c
+++ b/python/convert.c
@@ -208,6 +208,35 @@ bool PyUpb_IsNumpyNdarray(PyObject* obj, const upb_FieldDef* f) {
   return is_ndarray;
 }
 
+bool PyUpb_IsNumpyBoolScalar(PyObject* obj) {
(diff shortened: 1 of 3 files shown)


noelportillo (commented):
It's the bests

PR #22187, opened 2025-06-10
Drop windows-2019 runners from CI

mkruskal-google:
#test-continuous

PiperOrigin-RevId: 769765346


zhangskz:
I think 30.x is out of support now and we shouldn't need this one, but doesn't hurt

PR #22218, opened 2025-06-12
Remove redundant `break`

r-barnes:
`Py_RETURN_NONE` returns, so the `break` is unreachable.

Identified with `-Wunreachable-code-break` in LLVM.


r-barnes:
Tests need to be approved to run.

PR #22296, opened 2025-06-17
try hermetic python for 3.12

honglooker:


python/build_targets.bzl
```diff
@@ -97,7 +97,7 @@ def build_targets(name):
         ],
         deps = select({
             "//conditions:default": [],
-            ":use_fast_cpp_protos": ["@system_python//:python_headers"],
+            ":use_fast_cpp_protos": ["@python_3_12//cc:current_py_cc_headers"],
```
honglooker, line 100: `@rules_python//python/cc:current_py_cc_headers`

MODULE.bazel
```diff
@@ -118,42 +118,21 @@ register_toolchains(
...
-    python,
-    system_python = "python_{}".format(SUPPORTED_PYTHON_VERSIONS[-1].replace(".", "_")),
+python.toolchain(
+    ignore_root_user_error = True,
+    is_default = True,
+    python_version = "3.12",
 )
 
-pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip", dev_dependency = True)
+use_repo(python, "python_3_12")
```
honglooker, line 128: rm

MODULE.bazel
```diff
@@ -118,42 +118,21 @@ register_toolchains(
...
-        # Disable root warning for .pyc cache misses since CI runs as root.
-        # See https://github.com/bazelbuild/rules_python/pull/713
-        ignore_root_user_error = True,
-        is_default = python_version == SUPPORTED_PYTHON_VERSIONS[-1],
-        python_version = python_version,
-    )
-    for python_version in SUPPORTED_PYTHON_VERSIONS
-]
-
-use_repo(
```
honglooker, line 142: import sys

print(sys.executable)

(in a testO

honglooker: usr/bin/python or some bzl path cache etc

python/dist/dist.bzl
```diff
@@ -17,7 +16,7 @@ def _get_suffix(limited_api, python_version, cpu):
         return ".cp{}-{}.{}".format(python_version, abi, "pyd")
 
     if python_version == "system":
-        python_version = SYSTEM_PYTHON_VERSION
```
honglooker, line 20: if rule can get from toolchain

if macro: might explain aliasing going on

python/dist/dist.bzl
```diff
@@ -17,7 +16,7 @@ def _get_suffix(limited_api, python_version, cpu):
...
+        python_version = "312"
```
rickeylev, line 19: Can replace with something like:

```
runtime = ctx.toolchains[<python toolchain type>].py3_runtime
print(runtime.python_version)
```

Also add the python toolchain type `//python:tooclhain_type` to the toolchains attribute of the rule

honglooker: dist.bzl ~line 115
toolchains = @rulespython/...

PR #22302, opened 2025-06-18
Faster datetime parsing

senhalil:
We observed that replacing `date_object = datetime.datetime.strptime(second_value, _TIMESTAMPFORMAT)` with `date_object = datetime.datetime.fromisoformat(second_value)` cuts the total loading time by about 50% (we parse many jsons and it is a significant savings for us).

This change seems like it should be transparent since `_TIMESTAMPFORMAT` is practically the same as ISO format. 

A possible behavioral difference I can think of is `fromisoformat` would accept fractional/decimal seconds but I am not sure if `_TIMESTAMPFORMAT` would do so. If it doesn't, the code can be amended with a `td.seconds == int(td.seconds)` check if needed.

Would you be interested in this change?

### DIFF
--- a/python/google/protobuf/internal/well_known_types.py
+++ b/python/google/protobuf/internal/well_known_types.py
@@ -26,7 +26,6 @@ from google.protobuf.internal import field_mask
 
 FieldMask = field_mask.FieldMask
 
-_TIMESTAMPFORMAT = '%Y-%m-%dT%H:%M:%S'
 _NANOS_PER_SECOND = 1000000000
 _NANOS_PER_MILLISECOND = 1000000
 _NANOS_PER_MICROSECOND = 1000
@@ -145,7 +144,7 @@ class Timestamp(object):
           "time data '{0}' does not match format '%Y-%m-%dT%H:%M:%S', "
           "lowercase 't' is not accepted".format(second_value)
       )
-    date_object = datetime.datetime.strptime(second_value, _TIMESTAMPFORMAT)
+    date_object = datetime.datetime.fromisoformat(second_value)
     td = date_object - datetime.datetime(1970, 1, 1)
     seconds = td.seconds + td.days * _SECONDS_PER_DAY
     if len(nano_value) > 9:


esrauchg:
Sorry, but we unfortunately cannot make this particular change.

The context ProtoJSON timestamps are spec'd to be RFC 3339 format (actually _even more technically_ they are "RFC 3339 as a profile of ISO-8601-2019" which is only different in the sense that it wasn't clear before that whether lowercase letters should be accepted and only under that specific combination is it clear they shouldn't be, and ProtoJSON spec follows that spec).

RFC 3339 accepts only a very small subset of all ISO-8601 timestamps, and its considered a nonconformant Protobuf implementation if it accepts ISO timestamps which are not in that RFC 3339 subset (this PR would actually show up as a failure on our conformance test suite).

It may seem harmless for a parser to accept _more_ inputs than it is supposed to, but we seen users of protobuf have very major pain points when due to e.g. JavaProto's JSON parse being too lax on what it accepts, and then when they intend to rewrite their server to Go but when they go to turn on the Go server they find it would start "correctly" 400'ing many external customer requests which have been working for years. And then they end up stuck between a choice to either abandon their language rewrite, or breaking preexisting customers with a change just due to something that is intended to be an internal technical choice which should be a no-op to customers (its not realistic to play the rat race game of relaxing all language implementations to be lax-bug-compatible with all others).

For that reason we do really have to try to keep the canonical Protobuf implementations consistent in only success-parsing the same things, even when there is some performance overhead like this example.

senhalil:
Thanks for the crystal clear explanation @esrauchg, very much appreciated!

senhalil:
@esrauchg thanks again for the comprehensive response.  I read https://www.rfc-editor.org/rfc/rfc3339.html (and related references) and `ToJsonString` code carefully.  I think I have a proposal that makes sense in the "RFC 3339 as a profile of ISO-8601-2019" context.

`ToJsonString` strips the time offset and the optional fractional second parts away and handles them separately. Then `strptime` is used only for
1. ensuring that the remaining part of the string is in fact in the `%Y-%m-%dT%H:%M:%S` format 
2. and to convert the string into a `datetime` object

The following regex ensures 1. and then `fromisoformat` converts the string into `datetime`.  I believe this keeps the canonical behavior and it is faster because `strptime` is designed to handle arbitrary formats and comes with a significant overhead.

```python
...
if re.fullmatch(r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}", second_value) is None:
  raise ValueError(
      "time data '{0}' does not match format '%Y-%m-%dT%H:%M:%S'".format(second_value)
  )
date_object = datetime.datetime.fromisoformat(second_value)
```

I tried to run the tests but `bazel test //python:well_known_types_test --test_env=3.12 --noenable_bzlmod --enable_workspace` fails with some rust and android rule errors and I couldn't manage to make it work 🤷 

I pushed this latest version to the branch of this PR `patch-1` which has the following diff https://github.com/protocolbuffers/protobuf/compare/main...senhalil:protobuf:patch-1

PR #22316, opened 2025-06-19
Convert list to string for COMMENT argument

stima:
The issue occurs because CMake expects a single string for the COMMENT argument, but when a list is passed, it tries to expand the list into multiple arguments. To resolve this convert lists to strings using string(REPLACE ...) to join list items into a single string.

```
CMake Warning (dev) at <path>/protobuf-generate.cmake:150 (add_custom_command):
  COMMENT requires exactly one argument, but multiple values or COMMENT
  keywords have been given.
```

### DIFF
--- a/cmake/protobuf-generate.cmake
+++ b/cmake/protobuf-generate.cmake
@@ -140,10 +140,12 @@ function(protobuf_generate)
 
     set(_comment "Running ${protobuf_generate_LANGUAGE} protocol buffer compiler on ${_proto}")
     if(protobuf_generate_PROTOC_OPTIONS)
-      set(_comment "${_comment}, protoc-options: ${protobuf_generate_PROTOC_OPTIONS}")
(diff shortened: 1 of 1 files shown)


esrauchg:
Thanks for raising this proposal. It looks like the cmake tests are broken with the proposed PR, can you take a look at that?

stima:
Hey @esrauchg, I doubt the failures are related to the proposed changes:

- Windows CMake and Windows CMake Shared: internal compiler error
- Linux CMake C++20: static assertion failure

I double-checked with a local "user" build on both Windows and Linux - everything passed successfully. Would appreciate any guidance on how I could narrow this down further if needed.

acozzette:
@stima Could you try rebasing this onto the current `main`? I suspect that will fix the test failures, which I agree look unrelated.

stima:
@acozzette rebased.

PR #22317, opened 2025-06-20
[PHP] add missing reserved words: mixed, never and object

fernandocarletti:
A few reserved keywords are missing in the protobuf implementation: 
- `mixed` (PHP 8.0)
-  `never` (PHP 8.1)
-  `object` (PHP 7.2)

They are described in this page: https://www.php.net/manual/en/reserved.other-reserved-words.php.

There should be no breaking changes to the current active versions considering that PHP 7.2 and 8.0 already reached EOL and PHP 8.1 is under security fixes only according to its website: https://www.php.net/supported-versions.php. Any attempt to use those words generates a fatal error in PHP 8.1+.


php/src/Google/Protobuf/Internal/GPBUtil.php
```diff
@@ -251,23 +251,23 @@ public static function getClassNamePrefix(
...
+            "for"=>0, "foreach"=>0, "function"=>0, "global"=>0, "goto"=>0,
+            "if"=>0, "implements"=>0, "include"=>0, "include_once"=>0, "instanceof"=>0,
+            "insteadof"=>0, "int"=>0, "interface"=>0, "isset"=>0, "iterable"=>0,
+            "list"=>0, "match"=>0, "mixed"=>0, "namespace"=>0, "never"=>0,
+            "new"=>0, "null"=>0, "object"=>0, "or"=>0, "parent"=>0,
+            "print"=>0, "private"=>0, "protected"=>0, "public"=>0, "readonly" => 0,
+            "require"=>0, "require_once"=>0, "return"=>0, "self"=>0, "static"=>0,
+            "string"=>0, "switch"=>0, "throw"=>0, "trait"=>0, "true"=>0,
+            "try"=>0, "unset"=>0, "use"=>0, "var"=>0, "void"=>0,
+            "while"=>0, "xor"=>0, "yield"=>0,
```
fernandocarletti, line 270: For whoever is reviewing it, I noticed that some keys were not alphabetically ordered, being more difficult to find the keywords when checking for it.

Since I was already changing it, I ended up refactoring it to reorder them, hence the bigger change set. Let me know if this is a problem, I can revert it and add the keys to the end only.

The same was done in [names.cc](https://github.com/protocolbuffers/protobuf/pull/22317/files#diff-21e0cdb53aa203c0854e1ab855d186b72abe928ca2b8a7b5e3769687783b3181R18).

esrauchg: This is ok, thanks!

esrauchg (approved):

fernandocarletti:
@esrauchg FYI I just rebased it with the current main branch.

acozzette:
This test seems to be failing:
```
1) GeneratedClassTest::testPrefixForReservedWords
Exception: Couldn't find descriptor. Note only generated code may derive from \Google\Protobuf\Internal\Message
```
I don't know enough about PHP to comment on why that's happening, though.

Issue #22403, opened 2025-06-24
`GPBDictionary` is incorrectly marking `objectForKey:` as `nonnull ` due to `NS_ASSUME_NONNULL_BEGIN` Macro


luoshuaiqing:
**What version of protobuf and what language are you using?**
Version: 28.3
Language: Objective-C

**What operating system (Linux, Windows, ...) and version?**
MacOS

**What did you do?**
Root Cause: 
- `GPBDictionary` is incorrectly marking `objectForKey:` as `nonnull ` due to `NS_ASSUME_NONNULL_BEGIN` Macro

Problem it caused for me:
- When I call this method from swift, it will crash the App if the key doesn't exist in the `GPBDictionary` because objc marks it as nonnull while it is actual returning null, and therefore Swift recognizes as a non-optional object while it is actually a nil.

thomasvl:
Thank you for the report, you are right, this is a long standing bug.

Since the fix changes the Swift signature, it's likely going to count as a breaking change.

You reported this in 28.3, but likely goes back further and is in the current releases also.  Because it is breaking, I don't think we'll be fixing it on any release branches as it probably means a major version number bump.

It's going to take a time to sort out the right way to deal with this since it is a breaking change and we didn't have any breaking changes planned for ObjC. But once we get a plan together we'll update things.

thomasvl:
Still working on messaging to get this fixed.

thomasvl:
Looks like the messaging went out a little while ago, and I hadn't noticed it: https://protobuf.dev/news/2025-09-19/

thomasvl:
Per that timeline, we'll wait for the _v33_ release, before landing the change to fix this.

thomasvl:
This is fixed on _main_, but won't be a release until Q1 2026.

PR #22415, opened 2025-06-26
Fix objc `objectForKey` nullability

luoshuaiqing:
This is for fixing issue https://github.com/protocolbuffers/protobuf/issues/22403

`GPBDictionary` is incorrectly marking `objectForKey:` as nonnull due to `NS_ASSUME_NONNULL_BEGIN` Macro


thomasvl:
Thanks for putting up the PR, but as mentioned on the issue, this will be a breaking change, so we're still working through that messaging and this can't land without being in sync with that messaging.

luoshuaiqing:
Thank you @thomasvl ! I will close this PR for now

PR #22440, opened 2025-06-29
Fix warning C4141: 'inline': used more than once

cyyever:


cyyever:
@acozzette GCC builds are fixed.

acozzette:
It looks like there may be some changes needed in utf8_range to get it to build successfully:
```
third_party/utf8_range/utf8_range_neon.inc:9:26: error: duplicate 'inline' declaration specifier [-Werror,-Wduplicate-decl-specifier]
```

cyyever:
@acozzette Fixed

Issue #18750, opened 2024-10-10
Cannot build `@com_google_protobuf//python/dist:binary_wheel`


AtoZdevelopment:
**What version of protobuf and what language are you using?**
Version: 28.2
Language: Bazel

**What operating system (Linux, Windows, ...) and version?**
Linux, Ubuntu 22.4

**What runtime / compiler are you using (e.g., python version or gcc version)**
gcc (GCC) 13.2.0
bazel 7.3.2

**What did you do?**
1. Create folder with the following content:
MODULE.bazel
```
bazel_dep(name = "protobuf", version = "28.2", repo_name = "com_google_protobuf")
```
**What did you expect to see**
successfull build of the wheel

**What did you see instead?**
```
$ bazel build @com_google_protobuf//python/dist:binary_wheel
DEBUG: /home/az/.cache/bazel/_bazel_az/be14e48a27bba7024ecb06f57ad2a794/external/rules_python~/python/private/bzlmod/python.bzl:44:10: WARNING: Ignoring toolchain 'python_3_8' from module 'rules_fuzzing': Toolchain 'python_3_8' from module 'protobuf' already registered Python version 3.8 and has precedence
DEBUG: /home/az/.cache/bazel/_bazel_az/be14e48a27bba7024ecb06f57ad2a794/external/rules_python~/python/private/bzlmod/python.bzl:44:10: WARNING: Ignoring toolchain 'python_3_9' from module 'rules_fuzzing': Toolchain 'python_3_9' from module 'protobuf' already registered Python version 3.9 and has precedence
DEBUG: /home/az/.cache/bazel/_bazel_az/be14e48a27bba7024ecb06f57ad2a794/external/rules_python~/python/private/bzlmod/python.bzl:44:10: WARNING: Ignoring toolchain 'python_3_10' from module 'rules_fuzzing': Toolchain 'python_3_10' from module 'protobuf' already registered Python version 3.10 and has precedence
DEBUG: /home/az/.cache/bazel/_bazel_az/be14e48a27bba7024ecb06f57ad2a794/external/rules_python~/python/private/bzlmod/python.bzl:44:10: WARNING: Ignoring toolchain 'python_3_11' from module 'rules_fuzzing': Toolchain 'python_3_11' from module 'protobuf' already registered Python version 3.11 and has precedence
[...]

Issue #8520, opened 2021-04-21
MessageToDict converts uint64 to string in python


chris-from-the-coast:
**What version of protobuf and what language are you using?**
Version: protobuf 3.12.4 and 3.15.8
syntax = "proto3"

Language: Python

**What operating system (Linux, Windows, ...) and version?**
Macos: 10.15.7 64-bit

**What runtime / compiler are you using (e.g., python version or gcc version)**
Python 3.7.7
`import struct;print( 8 * struct.calcsize("P"))`
Output:
> 64

**What did you do?**

1. Create a source proto file with
```
message transmitter {
  ...
  uint64 transmitter_ticks = 8;
  ...
}
```

2. Compile
`protoc -I=$SRC --python_out=./ $SRC/transmitter.proto

3. Parse a transmitter message with the ticks set to 1618188465. (Note ticks is of the type int)

4.  Pass the message to `MessageToDict(parsed_transmitter_message)`

**What did you expect to see**
type(transmitter_ticks) should be an int

**What did you see instead?**
type(transmitter_ticks) is a str

**Anything else we should know about your project / environment**
Example code:
Source protobuf code:
```
...
syntax = "proto3";
message transmitter {
  string xxx = 1;
  optional Units xxx = 2;
  TimestampWithOffset xxx = 3;
  TimestampWithOffset xxx = 4;
  uint32 xxx = 5;
  string xxx = 6;
  string xxx = 7;
  uint64 transmitter_ticks = 8;
  string xxx = 9;
  string xxx = 10;
}
```
Client code:
```
MessageToDict(transmitter_message)
Output:
{
  ...
  'transmitterTicks' = '1618188465', # Should be an int
  ...
}
```

chris-from-the-coast:
Hey guys. I just wanted to check up on this issue.

anandolee:
The behavior is correct according to our json mapping:

https://developers.google.com/protocol-buffers/docs/proto3#json

hxtk:
I wonder how you intend that behavior to interact with the `jstype` pseudo-option. From my reading, the `jstype` option would allow a user to annotate a field as preferring a numeric type with a message definition such as the following:

```proto
message Foo {
   int64 bar = 1 [jstype = JS_NUMBER];
}
```

[...]

PR #22448, opened 2025-06-30
Support the JSType FieldOption in Python's JSON format

hxtk:
In Python code for marshalling protobuf message types into JSON strings or Python dicts suitable for conversion to the canonical Protobuf JSON format of that message, inspect the JSType option value on 64-bit integer fields and render them as numeric types if the annotation is present and indicates that preference. Similarly, marshal 32-bit integer types as strings when the JSType option value on those fields indicates that preferences.

-----

I commented about this change [on a long-dead issue related to the topic](https://github.com/protocolbuffers/protobuf/issues/8520#issuecomment-3006816080), and I ended up polishing the change enough that I thought it was suitable for a pull request to make discussion more visible if nothing else, since I doubt that old issue is still being watched by anyone.

I hope that I have interpreted the maintainers' intent for the `jstype` FieldOption, because if this interpretation is canonical, it will allow my team to use Protobuf to migrate to a schema-first API design methodology without simultaneously making breaking changes to our existing JSON APIs. If I have misunderstood it, I do not know what the alternative interpretation might be and would welcome a correction (and suggest documentation additions).

I believe this change solves the problem demonstrated in the tests, and I have tested it by monkey-patching the proposed `google.protobuf.json_format._Printer._FieldToJsonObject` method in at runtime after installing `protobuf` from `pypi`, but I have not been able to run the tests myself because #18750 impacts me. I will update this description if I succeed in getting tests to run locally.

I have misgivings about my inclusion of `descriptor_pb2` as a top-level import, because in other places where I see it used throughout the codebase, its import tends to be deferred until its first use. I'd be willing to make that change if a maintainer can suggest a more appropriate scope for the import.


acozzette:
Thank you for the pull request, but unfortunately I don't think we want to go in this direction. The `jstype` option is only meant to affect JavaScript generated code, not JSON parsing or serialization behavior.

If you really need to represent 64-bit integers as numbers, you might want to consider just forking json_format.py so that you can customize it to do this. This wouldn't be the standard protobuf JSON format, but there's nothing to stop you from using your own customized format with protos.

hxtk:
> Thank you for the pull request, but unfortunately I don't think we want to go in this direction. The `jstype` option is only meant to affect JavaScript generated code, not JSON parsing or serialization behavior.
> 
> If you really need to represent 64-bit integers as numbers, you might want to consider just forking json_format.py so that you can customize it to do this. This wouldn't be the standard protobuf JSON format, but there's nothing to stop you from using your own customized format with protos.

Ah, I hadn't been considering different types for JS and JSON so I had a hard time imagining that use case when I thought about the intended purpose of the `jstype` option. I'd hoped this was a bug in the Python JSON format because it would be very convenient for me if it were, but I can understand the value in making serialization formats as semantically inflexible as possible so that it's easier to maintain conformant parsers across a multitude of languages.

PR #22511, opened 2025-07-04
Fix several Windows warnings

cyyever:


src/google/protobuf/arena_unittest.cc
```diff
@@ -541,6 +541,7 @@ class DispatcherTestProto : public Message {
   }
   const internal::ClassData* GetClassData() const PROTOBUF_FINAL {
     ABSL_LOG(FATAL);
+    return nullptr;
```
acozzette, line 544: Can you say more about the warning that this fixes? My understanding is that `ABSL_LOG(FATAL)` expands to a constructor of a class whose destructor is marked `[[noreturn]]`, so this should tell the compiler that we will never return from this function.

src/google/protobuf/map_field.h
```diff
@@ -837,9 +837,6 @@ class PROTOBUF_EXPORT MapIteratorBase {
   ValueRef value_;
 };
 
-extern template class MapIteratorBase</*kIsMutable=*/false>;
```
acozzette, line 840: I don't think we want to delete these lines, because as I understand, these are here to tell the compiler to look for their implementation elsewhere (it's in map_field.cc) instead of generating them again.

Does adding `PROTOBUF_EXPORT` here fix the problem? That is just a guess, but I see that we do something similar [here](https://github.com/protocolbuffers/protobuf/blob/23649b0c8c124b2d3b1fe46752df5107cd2a5322/src/google/protobuf/metadata_lite.h#L214).

src/google/protobuf/arena_unittest.cc
cyyever: It is warning `C4715 'function' : not all control paths return a value`. I hadn't investigated the reason, it could be a MSVC bug..

src/google/protobuf/map_field.h
cyyever: Because `MapIteratorBase` is used below as parent of `PROTOBUF_EXPORT ConstMapIterator`, `MapIteratorBase` is implicitly exported in this context, that is why MSVC thought export and extern are conflicting.

src/google/protobuf/arena_unittest.cc
acozzette: Ok, that does sound to me like a bug in MSVC or maybe Abseil. In that case I don't think we should add this extra return because it makes the code less clear.

src/google/protobuf/map_field.h
acozzette: I'm not sure what the right solution is but I think we do need to keep these two lines, because I believe they're intentionally preventing the compiler from generating unnecessary instantiations of this template.

Issue #22514, opened 2025-07-04
Python: Drop usage of pkg_resources


parthea:
**What version of protobuf and what language are you using?**
Version: main/v3.6.0/v3.5.0 etc. (NOTE: please try updating to the latest version of protoc/runtime possible beforehand to attempt to resolve your problem)
Language: C++/Java/Python/C#/Ruby/PHP/Objective-C/Javascript

Version: main
Language: Python

**What operating system (Linux, Windows, ...) and version?**
Linux
6.12.27-1rodete1-amd64

**What runtime / compiler are you using (e.g., python version or gcc version)**
3.12.9
gcc (Debian 14.2.0-19+build4) 14.2.0

**What did you do?**
1. Search for usage of [pkg_resources](https://github.com/search?q=repo%3Aprotocolbuffers%2Fprotobuf%20pkg_resources&type=code) in protobuf
2. Observe setuptools release notes that `pkg_resources` is slated for removal by EOY https://setuptools.pypa.io/en/latest/history.html#v80-9-0

**What did you expect to see**

No usage of `pkg_resources` in `protobuf`

**What did you see instead?**

`pkg_resources` is used here:
https://github.com/search?q=repo%3Aprotocolbuffers%2Fprotobuf%20pkg_resources&type=code

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment**

jguamie:
We have a fix in progress.

acozzette:
This should be fixed by 8351926380c7cc91aae6df5695c91426e209f958.

PR #22515, opened 2025-07-04
fix: remove `google/__init__.py`

parthea:
Fixes https://github.com/protocolbuffers/protobuf/issues/22514
Fixes https://github.com/googleapis/google-cloud-python/issues/13974


acozzette:
@parthea Thank you for the PR, but we just landed a separate fix for this in 8351926380c7cc91aae6df5695c91426e209f958 so I don't think this is necessary anymore.

Issue #21177, opened 2025-04-08
Clarification on protobuf_maven namespace change and transitive dependency resolution


alfredomusumeci:
Hi all,

We’ve recently upgraded to the latest version of protobuf, and we noticed an unexpected side effect caused by the change in the Maven repository name from maven to protobuf_maven. Specifically, this seems to have impacted our dependency resolution when using rules_jvm_external, and I wanted to better understand the rationale behind this change and whether we might be approaching our setup incorrectly.

To give some context: rules_jvm_external allows projects to declare and pin their own Maven dependencies centrally in a maven_install block, which can then be referenced across targets. This ensures consistency in versions and avoids surprises from transitive dependencies.
However, since protobuf now declares its own maven_install under the protobuf_maven namespace, we no longer have visibility or control over overlapping dependencies declared there. As a result, even if we explicitly define and version certain dependencies in our own maven_install, they can be silently overridden by those declared by protobuf_maven, especially for targets that transitively use protobuf.

To make this more concrete: suppose we are building target A, which depends on:

- our own dependency B (declared via our maven_install)
- and protobuf (directly or transitively, i.e. dependency C)

If dependency B in turn depends on dependency X, and dependency X is also brought in (perhaps at a different version) by protobuf_maven, then Bazel appears to resolve it using the version from protobuf_maven, even though we’ve declared a different version ourselves. In our case, this led to hard-to-diagnose issues where transitive dependencies like gson were pulled in by protobuf at incompatible versions, silently causing breakages.

I understand that there are already several tickets advocating for protobuf to not use the generic maven namespace, and instead adopt a dedicated one (which it now does). I see the reasoning behind this: shared use of a generic namespace like maven can lead to collisions when different modules pull in their own artifacts. However, in our case, the opposite seems true: by having protobuf isolate its dependencies in a separate namespace, we’ve lost the ability to unify and control dependency resolution across our workspace.

So I’m hoping to understand:

1. Is there a recommended way to handle scenarios like the above which share overlapping dependencies, yet they are managed by two different namespaces?
2. Can one "alias" the deps from protobuf_maven to the ones in our own maven namespace?

I’m genuinely trying to understand the tradeoffs here, so I’m happy to be corrected if we’re missing a better way to align dependency versions.

Thanks.

Related issues in favour of the namespace change:
- https://github.com/protocolbuffers/protobuf/issues/17117
- https://github.com/protocolbuffers/protobuf/issues/16839
[...]

PR #22544, opened 2025-07-09
Rename maven install back to `maven` in MODULE.bazel

meteorcloudy:
Fixes https://github.com/protocolbuffers/protobuf/issues/21177

This PR essentially reverts https://github.com/protocolbuffers/protobuf/pull/18641, which claimed

> Since protobuf is not contributing to user's JARs

This is not true since targets like `@com_google_protobuf//:protobuf_java` are meant to be consumed by other projects, therefore protobuf should not use a private maven install namespace. Otherwise, it leads to duplicated maven jars and classpath conflicts. See https://github.com/protocolbuffers/protobuf/issues/21177 and https://github.com/bazel-contrib/rules_jvm_external/issues/916#issuecomment-3045506487

The original warning message caused by multiple modules contributing to `maven` can be suppressed with https://github.com/bazel-contrib/rules_jvm_external/pull/1393, which will be available in rules_jvm_external 6.8.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -177,6 +177,7 @@ crate.spec(
 crate.from_specs()
 use_repo(crate, crate_index = "crates")
 
+# Keep this list minimal; these dependencies will be part of the common `maven` install and could affect end-user projects.
 PROTOBUF_MAVEN_ARTIFACTS = [
     "com.google.code.findbugs:jsr305:3.0.2",
     "com.google.code.gson:gson:2.8.9",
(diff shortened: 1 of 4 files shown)


meteorcloudy:
This change is implemented internally at cl/781031722

jschaf:
> This is not true since targets like @com_google_protobuf//:protobuf_java are meant to be consumed by other projects

That target is aliased to `//java/core` which is plain [java_library](https://github.com/protocolbuffers/protobuf/blob/main/java/core/BUILD.bazel#L261).

I figured the change was safe since it only impacted test-only targets: https://github.com/protocolbuffers/protobuf/pull/18641/files

That said, I don't use Java directly—my only goal was to silence the warning. Since that's now possible, I no longer have skin in the game.

Issue #20304, opened 2025-02-11
Support Python 3.13t (upb)


liujianwish:
Hi,

There is an performance issue in protobuf when I switching python version from 3.10 to 3.13. Performance droped 20x comparing with 3.10. After some investigating, I find the root cause is due to the backend implmentation of the protobuf. In 3.13, the backend implementation is python, compatible version I guess. 

Is there any release plan for supporting python 3.13 (upb ver) or any temp solution to build a upb for python 3.13?

Frank

shaod2:
Hi, thanks for reporting the performance issue.

Do you have examples / evidence of the performance drop, considering Protobuf backend, in details?

liujianwish:
> Hi, thanks for reporting the performance issue.
> 
> Do you have examples / evidence of the performance drop, considering Protobuf backend, in details?

Hi @shaod2 , thanks for your rapid reply. 

I have made a demo test as below:

Lib dependency:  protobuf==5.29.3

Environment: Linux 5.15.0-78-generic #85-Ubuntu SMP Fri Jul 7 15:25:09 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

Python version: 3.13.12 vs 3.13.1t

Command: `python test.py`

Test results:

![Image](https://github.com/user-attachments/assets/b9b81607-8725-4997-a591-093f50cbd3be)

![Image](https://github.com/user-attachments/assets/bbf69243-cae5-4f3b-89c6-3a8963632a11)


I have attached all the test code as below:

[demo.tar.gz](https://github.com/user-attachments/files/18760965/demo.tar.gz)


Frank

davidfritzsche:
@liujianwish Do you use the regular 3.13.1 or the experimental free-threaded 3.13.1**t** (see https://docs.python.org/3/whatsnew/3.13.html#free-threaded-cpython)?
It seems to me that the upb-based protobuf package does not support free threaded Python at the moment, at least there are no wheels for free-threaded Python on pypi.org, i.e., on a free-threaded Python you would get the Python-only wheel.

liujianwish:
[...]

PR #22736, opened 2025-07-23
Use strong reference CPython C APIs under free-threading

lysnikolaou:
Hey folks! 👋 

For the last few weeks, I've been looking into support for free-threading. This is the first in a series of PRs related to that. It does the necessary changes to only use strong reference APIs under the free-threaded build of Python.

Related to #20304.

### DIFF
--- a/python/google/protobuf/pyext/message.cc
+++ b/python/google/protobuf/pyext/message.cc
@@ -292,6 +292,12 @@ static PyObject* New(PyTypeObject* type, PyObject* args, PyObject* kwargs) {
   }
   CMessageClass* newtype = reinterpret_cast<CMessageClass*>(result.get());
 
+  // Cache the descriptor, both as Python object and as C++ pointer.
+  const Descriptor* descriptor =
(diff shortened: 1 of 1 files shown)


python/message.c
```diff
@@ -1865,6 +1871,9 @@ PyObject* PyUpb_MessageMeta_DoCreateClass(PyObject* py_descriptor,
     args = Py_BuildValue("s(OOO)O", name, state->cmessage_type,
                          state->message_class, wkt_base, dict);
   }
+#ifdef Py_GIL_DISABLED
+  Py_DECREF(wkt_base);
```
nascheme, line 1875: This needs to be `Py_XDECREF()` since the argument could be NULL.

googleberg:
Looks like this PR might need a rebase so we can kick the tests again.

nascheme:
The changes from this PR that affect that UPB backend are included in GH-23202 and so this PR can be closed.

Note that the changes to the CPP backend are not included in GH-23202.  If we want to fix that backend to also be free-threaded compatible, changes from this PR might be useful.  However, I think it's better to close this.

PR #22740, opened 2025-07-23
Fix a typo in ruby ffi backend

ntkme:
Fix a typo that would cause crash when converting hash-like object to hash.

### DIFF
--- a/ruby/lib/google/protobuf/ffi/message.rb
+++ b/ruby/lib/google/protobuf/ffi/message.rb
@@ -254,7 +254,7 @@ module Google
             decoding_options = 0
             unless options.is_a? Hash
               if options.respond_to? :to_h
-                options options.to_h
+                options = options.to_h
               else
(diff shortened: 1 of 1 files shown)


JasonLunn:
Thanks for the patch, @ntkme. How did you discover the issue? Is there a straightforward way to reproduce the crash in a test that we can confirm now passes?

ntkme:
> How did you discover the issue?

I was just walking through this code and found it purely by coincidence.

> Is there a straightforward way to reproduce the crash in a test that we can confirm now passes?

You can pass an empty array `[]` as options. Array is not a Hash, but it implements `.to_h`. Or it can be something like `[['key1', 'value1'], ['key2', 'value2']]`.

https://docs.ruby-lang.org/en/3.4/Array.html#method-i-to_h

mkruskal-google (approved):

PR #22763, opened 2025-07-24
Remove all usages of std::is_trivial.

Romain-Geissler-1A:
std::is_trivial is deprecated in C++ >= 26. The standard advises to use more specific checks depending on the use case. In practice, gcc >= 15 with libstdc++ will emit deprecation warnings when using C++ >= 26.

I will comment inline on each case about the different changes, I have some doubts for some of them.


src/google/protobuf/arena.h
```diff
@@ -258,7 +258,7 @@ class PROTOBUF_EXPORT PROTOBUF_ALIGNAS(8) Arena final {
   template <typename T>
   PROTOBUF_NDEBUG_INLINE static T* PROTOBUF_NONNULL
   CreateArray(Arena* PROTOBUF_NULLABLE arena, size_t num_elements) {
-    static_assert(std::is_trivial<T>::value,
+    static_assert(std::is_trivially_default_constructible<T>::value,
```
Romain-Geissler-1A, line 266: For this one, this is effectively a revert of a 4 years old commit citing "old libstdc++". I hope it's ok now: https://github.com/protocolbuffers/protobuf/commit/b74892e9c368ee3517c404b569b3b447006518b6

src/google/protobuf/arenastring.h
```diff
@@ -200,8 +200,12 @@ class PROTOBUF_EXPORT TaggedStringPtr {
   void* ptr_;
 };
 
-static_assert(std::is_trivial<TaggedStringPtr>::value,
-              "TaggedStringPtr must be trivial");
+static_assert(std::is_trivially_default_constructible<TaggedStringPtr>::value,
```
Romain-Geissler-1A, line 203: I added these 3 checks, however I am unsure if we also want to have any check about being trivially copyable. I see the class implements some specific `Copy` methods, so I guess we are not meant to copy it via any copy operator/constructor, that's why I omitted the trivial copy check.

src/google/protobuf/map.h
```diff
@@ -224,8 +224,8 @@ class UntypedMapIterator {
 };
 
 // These properties are depended upon by Rust FFI.
-static_assert(std::is_trivial<UntypedMapIterator>::value,
-              "UntypedMapIterator must be a trivial type.");
+static_assert(std::is_trivially_default_constructible<UntypedMapIterator>::value,
```
Romain-Geissler-1A, line 227: Here since we already had quite other checks written already, I added the only "missing" one: trivial default construction.

src/google/protobuf/repeated_field.h
```diff
@@ -535,7 +535,7 @@ class ABSL_ATTRIBUTE_WARN_UNUSED RepeatedField final
   // Destroys all elements in [begin, end).
   // This function does nothing if `Element` is trivial.
   static void Destroy(const Element* begin, const Element* end) {
-    if (!std::is_trivial<Element>::value) {
+    if constexpr (!std::is_trivially_destructible<Element>::value) {
```
Romain-Geissler-1A, line 539: Since now protobuf requires C++17, I added an `if constexpr`. Since this is only about calling a destructor, I guess it's ok to only check `is_trivially_destructible`.

src/google/protobuf/repeated_field.h
```diff
@@ -1244,8 +1244,8 @@ PROTOBUF_NOINLINE void RepeatedField<Element>::GrowNoAnnotate(bool was_soo,
   if (old_size > 0) {
     Element* pnew = static_cast<Element*>(new_rep->elements());
     Element* pold = elements(was_soo);
-    // TODO: add absl::is_trivially_relocatable<Element>
-    if (std::is_trivial<Element>::value) {
+    if constexpr (std::is_trivially_copyable<Element>::value ||
```
Romain-Geissler-1A, line 1248: Same thing here for C++ 17 and the usage of `if constexpr`.

I checked the minimum abseil version, right now it's Abseil 20230125.3 and newer (https://github.com/protocolbuffers/protobuf/blob/07a81cb2d26c21275313b13b6158a34f6fc8ae78/src/google/protobuf/port_def.inc#L134). This version of Abseil already had absl::is_trivially_relocatable: https://github.com/abseil/abseil-cpp/blob/lts_2023_01_25/absl/meta/type_traits.h#L838

honglooker:
Looks like CI is failing since `end` isn't used if the `constexpr` expression evals to false? (repeated_field for example)

Romain-Geissler-1A:
> Looks like CI is failing since `end` isn't used if the `constexpr` expression evals to false? (repeated_field for example)

I have just rebased over the `main` branch and applied this change compared to the previous patch:

```
--- a/src/google/protobuf/repeated_field.h
+++ b/src/google/protobuf/repeated_field.h
@@ -534,7 +534,8 @@ class ABSL_ATTRIBUTE_WARN_UNUSED RepeatedField final

   // Destroys all elements in [begin, end).
   // This function does nothing if `Element` is trivial.
-  static void Destroy(const Element* begin, const Element* end) {
+  static void Destroy([[maybe_unused]] const Element* begin,
+                      [[maybe_unused]] const Element* end) {
     if constexpr (!std::is_trivially_destructible<Element>::value) {
       std::for_each(begin, end, [&](const Element& e) { e.~Element(); });
     }
```

honglooker:
@sbenzaquen, wdyt?

src/google/protobuf/arena.h
mkruskal-google: Yea this seems reasonable to me, especially since we never even updated the error message :P. I think this effectively allows types that aren't trivially copyable now, but I don't think that's actually a requirement for arena allocation

src/google/protobuf/arenastring.h
mkruskal-google: Based on the code history, I think trivially destructible is the important one here, but the others can't hurt.  I agree that being trivially copyable doesn't seem strictly necessary for how we use it

src/google/protobuf/repeated_field.h
mkruskal-google: Agreed

src/google/protobuf/repeated_field.h
mkruskal-google: Nice, thanks for the cleanup :)

mkruskal-google (approved):

PR #23024, opened 2025-08-06
Hold a per-object mutex when modifying PyUpb_WeakMaps

lysnikolaou:
Because `upb_inttable` is not thread-safe, we need to hold a per-object mutex when calling into those APIs. This makes sure that's true, while also adding a very thin compatibility layer in `python/python_api.h`, so we don't have to wrap all the usages of the `Py_BEGIN_CRITICAL_SECTION` APIs in `#ifdef`s.


nascheme:
An alternative version of these fixes are included in GH-23202 and so this PR can be closed.

PR #23056, opened 2025-08-08
Fix thread safety issues when sharing `Message` objects across threads

lysnikolaou:
Because `Message` objects may be shared across threads, we need to lock when accessing its internal fiels as well as the `upb` backend.

Here, I'm using `Py_BEGIN_CRITICAL_SECTION` to use the per-object mutex to lock the message when using it.


nascheme:
These fixes are included in GH-23202 and so this PR can be closed.

PR #23195, opened 2025-08-19
Add missing `load` for `sh_binary` in `python/dist/BUILD.bazel`

tempoz:
This was causing a build failure, since `sh_binary` is referenced in `python/dist/BUILD.bazel`, but it was never initialized.


zhangskz (approved):

zhangskz:
Thanks! I believe we normally build these with bazel 7 still which doesn't hit the missing load.

PR #23202, opened 2025-08-20
Add support for free-threaded CPython build

nascheme:
This branch adds support for the free-threaded (nogil) build of CPython.  It includes the work done by Lysandros (use strong references, critical sections for `Message` objects).  I excluded the changes to the "cpp" backend since I think we should focus on support for upb only at this time.  These changes are designed such that they shouldn't affect the behavior or performance of the default (non-free-threaded) build.

Lysandraos's PRs superseded by this one:

- GH-23056
- GH-23024
- GH-22736

Additional changes I made:

- Add recursive mutex implementation, essentially the same as `threading.RLock()`
- Add locking for the ObjCache structure.
- Add a critical section for the DescriptorPool object.
- Allocate the `c_descriptor_symtab` data on module init, avoiding thread-safety issue.
- Change PyUpb_WeakMap to use strong references. 
- Modify the `py_wheel` function in `python/dist/BUILD.bazel` so that free-threaded wheel files and .so files have to correct names (free-threaded build does not yet have a stable ABI).

I have some limited multi-threaded testing that I've run with the TSAN build.  These changes fix all TSAN warnings and the tests complete without crashing (previously they caused a SEGV).

I've also tested the multi-threaded scaling of the free-threading build of the modified library.  Unfortunately, it doesn't scale well when used from multiple parallel threads.  The ObjCache structure is a bottleneck since it is updated on every `Message` creation and deletion.  I think that could be fixed by having some lock-free hash map operatations for ObjCache (i.e. make the fast path where the object already exists not need to acquire the lock).  It's not clear to me if `upb_inttable_*` is thread-safe without locking.  I think that scaling work could come later, after we have it working without crashing.

Instructions on testing this:

- Install bazelisk
- Have a Python 3.14t build somewhere, preferably with TSAN and `--with-pydebug` options on.
- Add the "bin" folder for that Python to your PATH
- I use the following shell script to build with bazel, using Clang as the compiler:

```
#!/bin/sh
export PATH=$HOME/tmp/py-3.14t-tsan/bin:/usr/local/bin:/usr/bin:/bin
CC=clang CXX=clang++ bazelisk build  --config=tsan --compilation_mode=dbg --noenable_bzlmod //python/dist:binary_wheel
```
- Install the package: `python3 -m pip install --force-reinstall bazel-bin/python/dist/protobuf-6.33.0-cp314-cp314t-linux_x86_64.whl`

- Run tests, e.g. `python3 pb_threaded.py`

Attached is the `pb_threaded.py` script I was using to test.  It is too simple yet and we need some better testing.  Perhaps we can use pytest-run-parallel and run the existing unit tests.

[pb_threaded.py.txt](https://github.com/user-attachments/files/21906621/pb_threaded.py.txt)

### DIFF
--- a/python/protobuf.c
+++ b/python/protobuf.c
@@ -181,17 +181,6 @@ PyUpb_WeakMap* PyUpb_WeakMap_New(void) {
 }
 
 void PyUpb_WeakMap_Free(PyUpb_WeakMap* map) {
-#ifdef Py_GIL_DISABLED
-  // for the free-threaded build, the map is not weak and we need to decref
-  // all map entries
-  intptr_t iter = UPB_INTTABLE_BEGIN;
-  uintptr_t key;
-  upb_value val;
-  while (upb_inttable_next(&map->table, &key, &val, &iter)) {
-    PyObject* obj = upb_value_getptr(val);
-    Py_DECREF(obj);
-  }
-#endif
   upb_Arena_Free(map->arena);
 }
 
@@ -206,28 +195,12 @@ uintptr_t PyUpb_WeakMap_GetKey(const void* key) {
 }
 
 void PyUpb_WeakMap_Add(PyUpb_WeakMap* map, const void* key, PyObject* py_obj) {
-#ifdef Py_GIL_DISABLED
-  // For the free-threaded build, this is a strong referenced mapping, despite
-  // the API name.  That avoids quite a bit of complexity (like the need to
-  // use PyUnstable_TryIncRef().
-  Py_INCREF(py_obj);
-#endif
+  PyUnstable_EnableTryIncRef(py_obj);
   upb_inttable_insert(&map->table, PyUpb_WeakMap_GetKey(key),
                       upb_value_ptr(py_obj), map->arena);
 }
 
-static void weakmap_maybe_decref(PyUpb_WeakMap* map, const void* key) {
-#ifdef Py_GIL_DISABLED
-  upb_value val;
-  if (upb_inttable_lookup(&map->table, PyUpb_WeakMap_GetKey(key), &val)) {
-    PyObject* obj = upb_value_getptr(val);
-    Py_DECREF(obj);
-  }
-#endif
-}
-
 void PyUpb_WeakMap_Delete(PyUpb_WeakMap* map, const void* key) {
-  weakmap_maybe_decref(map, key);
   upb_value val;
   bool removed =
       upb_inttable_remove(&map->table, PyUpb_WeakMap_GetKey(key), &val);
@@ -236,7 +209,6 @@ void PyUpb_WeakMap_Delete(PyUpb_WeakMap* map, const void* key) {
 }
 
 void PyUpb_WeakMap_TryDelete(PyUpb_WeakMap* map, const void* key) {
-  weakmap_maybe_decref(map, key);
   upb_inttable_remove(&map->table, PyUpb_WeakMap_GetKey(key), NULL);
 }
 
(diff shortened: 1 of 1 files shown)


googleberg:
@nascheme can you please check that you've signed the [CLA](https://cla.developers.google.com/)?

nascheme:
> @nascheme can you please check that you've signed the [CLA](https://cla.developers.google.com/)?

As of today, I believe it should be.  My employer (Quansight) has already signed the CLA and it was a matter of adding my github email address to the list of employees.  Hopefully that takes care of it.

nascheme:
Just a heads up, I've been testing `protobuf` with a free-threaded version of grpc and found some issues.  Specifically, we need to be more careful about the weak-map in order to avoid reference counting issues in the free-threaded build.  I don't have a working fix yet but hopefully soon.  We will likely need to use `PyUnstable_TryIncRef()` since `PyUpb_ObjCache_Get()` is able to resurrect objects with refcnt == 0.

ngoldbaum:
Probably worth noting that this includes the commits from https://github.com/protocolbuffers/protobuf/pull/22736. @nascheme maybe you should clarify what the maintainers should do with that PR, given that this has diverged a little bit.

nascheme:
Latest push fixes some crashes found when running a multi-threaded version of the grpc route_guide example.  I avoided the need for `PyUnstable_TryIncRef()` by using strong references for the object map.

nascheme:
Any feedback on what might need to be done to move this PR along?  Based on my (admittedly somewhat limited) testing, it seems to work.  It shouldn't impact the non-free-threaded build since the alternative logic is either conditional on free-threading being enabled or is a no-op for the default build (like the locking functions).

anandolee:
The free threading support is under discussion. We will also document what will be thread safe and what are not thread safe. Sharing messages are not thread safe usages we will support.

We will submit some experimental free threading PRs for python cpp soon. UPB may take longer time for design, because message is under global object map. Add locks to each message is not what we want.

esrauchg:
Sorry to keep this is limbo longer but we have an extended internal discussion about what the shape of this is going to be.

We have a lot of concern about trying to make the current impl FT compatible with locks not just because the lock contention might not actually deliver what people need (where if people aren't concerned about single-threaded performance they can already use the Pure-Python impl) but also because we actually already have some 'unfixable' deadlock concerns preexisting internally between the GIL and other locks potentially being held in opposite order, and more locks is only more prone to deadlocks as well.

In the near term we intend to make changes so that both Py-C++ and Py-upb are compatible with FT in the sense of "you cannot use any one PyProto object concurrently from two threads". That unfortunately actually means no concurrent _reads_ of the same thing from two threads even (worse than standard C++ thread compatibility behavior where you can read on N threads or write on 1 thread). But its the most local stepping stone that we're comfortable with to try to get customers able to use it.

Long term we currently expect for FT to be both safe and efficient we will need to introduce the concept of immutable messages to PyProto. Concretely, how this topic works in e.g. Java is that the Builders are not at all threadsafe, but the built-messages are immutable and are safe to share freely between threads without any locks. But, as a new API shape instead of an implementation detail change we have to be very careful about how we do it (once we add any new nontrivial API we're realistically stuck with it forever).

Thank you for the high quality PR which is helping inform our discussion, and for creating sosmething that may already help users who may want to vendor Protobuf and patch this solid proposal, and apologies that we're not simply validating and accepting it as-is while we're figuring out how we want to support this important usecase.

esrauchg:
We have an ongoing higher touch engagement internally right now, and have made significant incremental progress towards compatibility.

We have an EOY delivery target for being free threading compatible with total memory safety (as in: racing writes of the same object should also not be unsound).

Thanks again for the PR that helped inform some of the topics here, but as we're not merging the PR as is I'm going to go ahead and close the PR.

Thanks!

PR #23220, opened 2025-08-24
Update raw_ptr_test.cc

cctv130:
error: constexpr variable 'raw' must be initialized by a constant expression
[build]    47 |   constexpr RawPtr<Obj> raw;
[build]       |                         ^~~
[build] 1 error generated.

### DIFF
--- a/src/google/protobuf/raw_ptr_test.cc
+++ b/src/google/protobuf/raw_ptr_test.cc
@@ -44,7 +44,11 @@ TEST(RawPtr, Basic) {
 }
 
 TEST(RawPtr, Constexpr) {
+#ifdef __windows__
   constexpr RawPtr<Obj> raw(kZeroBuffer);
+#else
+  constexpr RawPtr<Obj> raw;
+#endif
   EXPECT_EQ(raw->i, 0);
   EXPECT_EQ((*raw).i, 0);
   EXPECT_EQ(static_cast<void*>(&raw->i), kZeroBuffer);


cctv130:
@googleberg

cctv130:
Please wait a couple of days. I need to switch computers during this time, and the code needs to be tested on both Windows and Linux. I haven’t fully tested it yet, sorry.

cctv130:
@mkruskal-google 
5142a74a776a6981fdb9fdd1b3b8cb64

mkruskal-google:
I don't understand what problem this is fixing.  RawPtr has a constexpr default constructor, so the existing code should be valid

cctv130:
> I don't understand what problem this is fixing. RawPtr has a constexpr default constructor, so the existing code should be valid
When compiling protobuf on Windows, I encountered the error:
`error: constexpr variable 'raw' must be initialized by a constant expression`.
Of course, this error doesn’t occur if I disable the tests. Although `RawPtr` has a `constexpr` default constructor and it seems fine at first glance, the Windows version of Clang doesn’t automatically invoke the default constructor — perhaps it’s an MSVC-related issue. To ensure compatibility, I had to write it this way.

mkruskal-google:
This is fundamentally changing the test though.  What version of MSVC and C++ are you trying to build with?  If this is in our support window I'd say we should just add `#ifdef` to the test, otherwise this is WAI

src/google/protobuf/raw_ptr_test.cc
```diff
@@ -44,7 +44,11 @@ TEST(RawPtr, Basic) {
 }
 
 TEST(RawPtr, Constexpr) {
+#ifdef __windows__
```
mkruskal-google, line 47: We test on windows with MSVC and clang-cl, so this definitely isn't a universal issue.  What version of MSVC are you using?

mkruskal-google: I wasn't able to reproduce this going all the way back to msvc 19.20 (which we don't support) https://godbolt.org/z/ExqrYvrcs

cctv130: 91ae1ed4-b167-4859-b3ef-143369ee30e9

cctv130: When I compiled your demo using vs2022 enterprise version, it was obviously an error

PR #23335, opened 2025-09-03
Add missing `strip_import_prefix` parameter to `addressbook_proto` in `examples/BUILD`

tempoz:
Without this, the example does not build.


mkruskal-google:
This actually breaks the examples build according to our CI...

mkruskal-google:
Note: examples is its own Bazel workspace, and isn't meant to be buildable from the root directory

PR #23351, opened 2025-09-03
add missing `field_mask` upb proto targets to `src/google/protobuf/BUILD.bazel`

tempoz:
These targets are specified in the `BUILD.bazel` file in the root directory, but strangely did not exist.


googleberg:
We actually don't want people using upb-codegen directly because it is not guaranteed to be stable. Instead, we have a C++ wrapper that is going to eventually be stable but isn't just yet:
https://github.com/protocolbuffers/protobuf/tree/main/hpb

We should remove the dangling dead references in the root.

PR #23366, opened 2025-09-04
Remove unused import in `rust/cpp.rs`

tempoz:


esrauchg:
Unfortunately the PR won't merge cleanly with our internal state (the context of this is that there's some additional code related to https://github.com/google/crubit support on the internal version of this file which we have to strip from open source that uses this import, which is why the unused import is there only inside google3).

I'm going to recreate your PR with the internal change needed as https://github.com/protocolbuffers/protobuf/pull/23455 and will close this PR

Thanks for raising this!

PR #23369, opened 2025-09-04
Migrate `proto_descriptor_set`

### LINKED ISSUE #19647: proto_descriptor_set Bazel rule missing
`rules_proto` is deprecated and points here for Bazel rules, but this module is missing the `proto_descriptor_set` rule that was defined there.

SpencerC:
Closes #19647

### DIFF
--- a/bazel/BUILD.bazel
+++ b/bazel/BUILD.bazel
@@ -64,6 +64,12 @@ bzl_library(
     deps = ["//bazel/private:upb_proto_library_internal_bzl"],
 )
 
+bzl_library(
+    name = "proto_descriptor_set_bzl",
+    srcs = ["proto_descriptor_set.bzl.bzl"],
+    visibility = ["//visibility:public"],
+)
+
 # The data in this target is exposed in //bazel/private:for_bazel_tests
 filegroup(
     name = "for_bazel_tests",
--- a/tools/file_concat/BUILD.bazel
+++ b/tools/file_concat/BUILD.bazel
@@ -1,0 +1,7 @@
+cc_binary(
+    name = "file_concat",
+    srcs = [
+        "main.cc",
+    ],
+    visibility = ["//visibility:public"],
+)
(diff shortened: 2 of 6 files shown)


bazel/BUILD.bazel
```diff
@@ -64,6 +64,12 @@ bzl_library(
     deps = ["//bazel/private:upb_proto_library_internal_bzl"],
 )
 
+bzl_library(
+    name = "proto_descriptor_set_bzl",
+    srcs = ["proto_descriptor_set.bzl.bzl"],
```
turmanticant:
```suggestion
    srcs = ["proto_descriptor_set.bzl"],
```

neilconway:
+1 -- we also use this, and the lack of `proto_descriptor_set` makes it impossible to migrate from `rules_proto` to the built-in protobuf rules.

bazel/private/file_concat/main.cc
mkruskal-google, line 1: Can you move this tool to //bazel/private?  This isn't something we'd want to make public, and I think this could be simplified a lot for this purpose (simply concatenating the serialized bytes would be sufficient)

tools/file_concat/BUILD.bazel
```diff
@@ -0,0 +1,7 @@
+cc_binary(
+    name = "file_concat",
+    srcs = [
+        "main.cc",
+    ],
+    visibility = ["//visibility:public"],
```
mkruskal-google: Does this need to be public if it's a hidden attr?

bazel/private/file_concat/main.cc
SpencerC: Moved to //bazel/private. Do you want me to try to simplify it now, or could this be done in a follow-up PR? This is exactly as it was implemented in `rules_proto` (and not by me), so I hesitate to monkey with the implementation here.

tools/file_concat/BUILD.bazel
SpencerC: Changed.

bazel/private/file_concat/main.cc
mkruskal-google: Yea as long as it's private we can do that later

Issue #23236, opened 2025-08-26
[PHP] Cannot install 4.32.0 alongside google/gax (and therefore most google packages)


acoulton:
**What version of protobuf and what language are you using?**
Version: Attempting to update from 4.31.1 to 4.32.0
Language: PHP

**What operating system and version?**
* Debian 12
* PHP 8.4
* composer 2.8.10
* native ext-protobuf installed and enabled at version 4.31.1

**What runtime / compiler are you using (e.g., python version or gcc version)**
PHP

**What did you do?**

Our automated tooling attempted to update the google/protobuf composer package from 4.31.1 to 4.32.0. This failed with an error during composer install, so I attempted to perform the update manually.

1. Start a runtime with PHP 8.4 and the native ext-protobuf installed.
2. Run `composer show --platform` and observe that the output includes `ext-protobuf 4.31.1`, confirming that composer can detect the presence and version of the native extension.
3. Navigate to an empty directory.
4. `composer init -n --require google/gax:^1.36` to create a new composer.json requiring only the most recent version of google/gax.
5. `composer require google/protobuf:^4.31.1`
6. `composer require --with-all-dependencies google/protobuf:^4.32.0`
**What did you expect to see**

I expected the final command to succeed and install google/gax at 1.36.0 and google/protobuf at 4.32.0, with these requirements written to the composer.json and composer.lock files.

**What did you see instead?**

The final command failed with this output:

```
./composer.json has been updated
Running composer update google/protobuf --with-all-dependencies
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - google/gax is locked to version v1.36.1 and an update of this package was not requested.
    - Root composer.json requires google/protobuf ^4.32.0 -> satisfiable by google/protobuf[v4.32.0].
    - google/gax v1.36.1 conflicts with ext-protobuf <3.7.0 (google/protobuf v4.32.0 provides ext-protobuf *).


Installation failed, reverting ./composer.json and ./composer.lock to their original content.

```

**Anything else we should know**

It appears that this issue was introduced as a result of changes by @bshaffer in #21217, which was merging in changes from @brettmc in #13141 that had not made it into the distribution version of the package until now.

As far as I can tell from debugging and from the composer message output, the problem is:

* This package now specifies that it provides **any** version of ext-protobuf (`*`).
[...]

PR #13141, opened 2023-06-25
adding provide to php composer.json

brettmc:
Adding composer config to allow the native protobuf extension to provide ext-protobuf.

This allows libraries to require at least one protobuf implementation. If the extension is not available, it can be provided by the native package. If the extension is available but the native package is required, the native will be installed.

Importantly, for libraries which require at least one of them to be installed, composer will complain if neither is available.

PR #23373, opened 2025-09-04
fix(php): remove "provide" from composer.json

bshaffer:
fixes https://github.com/protocolbuffers/protobuf/issues/23236

see https://github.com/protocolbuffers/protobuf/issues/23236#issuecomment-3254793639

TLDR: `provide` is for virtual packages, and we are setting this to `ext-protobuf: *`, which is not virtual. As a result its conflicting with our latest versions of GAX

### DIFF
--- a/php/composer.json
+++ b/php/composer.json
@@ -8,9 +8,6 @@
   "require": {
     "php": ">=8.1.0"
   },
-  "provide": {
-    "ext-protobuf": "*"
-  },
   "require-dev": {
     "phpunit/phpunit": ">=5.0.0 <8.5.27"
   },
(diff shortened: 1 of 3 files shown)


haberman:
Here is the original PR that added this `provide` line: https://github.com/protocolbuffers/protobuf/pull/13141

Will this PR break that use case?

bshaffer:
@haberman 
The previous version `4.31.0` will remain marked as "providing" the virtual package. We are only making sure the next version of `4.31.1` does _not_ provide the virutual package.

It will break the use case in the sense that someone requiring the virtual package will not get `4.31.1` when they update.  However, I don't think that's an important use-case (as I don't think anyone's using it this way). And it is only supported in version `4.31.0` as the addition of the `provides` didn't actually make it to packagist until 2 years after that PR was submitted, so I don't think anyone is using it or will care when it's no longer supported.

PR #23390, opened 2025-09-05
fix(strtod): use size_t instead of int for strlen() result #1

lyriccoder:
`strlen()` returns `size_t`, but the code was storing its value in an `int`.

Why this is important:
- `strlen()` can return values larger than `INT_MAX`, which may lead to 
  truncation or undefined behavior.
- Using the wrong type makes the code non-compliant with C standards and 
  triggers MISRA / static analysis warnings.
- This could potentially break parsing in edge cases with very long strings.

The fix:
- Changed `int size_diff` to `size_t size_diff` when computing 
  `strlen(localized) - strlen(str)`.

Additional note:
`size_diff` is only used to account for cases where the localized radix 
character occupies multiple bytes (e.g., in exotic locales).  
Since localization replaces '.' with a longer symbol at most, `localized` 
is guaranteed to be at least as long as `str`.  
This means the difference cannot be negative, so `size_t` is safe and 
semantically correct (see the original comment in the code).

### DIFF
--- a/upb/lex/strtod.c
+++ b/upb/lex/strtod.c
@@ -68,7 +68,7 @@ double _upb_NoLocaleStrtod(const char *str, char **endptr) {
     // Update endptr to point at the right location.
     if (endptr != NULL) {
       // size_diff is non-zero if the localized radix has multiple bytes.
-      int size_diff = strlen(localized) - strlen(str);
+      size_t size_diff = (size_t)strlen(localized) - (size_t)strlen(str);
       *endptr = (char *)str + (localized_endptr - &localized[0] - size_diff);
     }
   }


anandolee:
Thanks for your pull request. Can you please sign the CLA before we can go forward?

lyriccoder:
> Thanks for your pull request. Can you please sign the CLA before we can go forward?

Hi @anandolee, I thought I had signed the CLA already.

lyriccoder:
Hi @anandolee, what should I do next?

lyriccoder:
Hi @anandolee, could you please approve and make the merge? Seems the test have been passed

upb/lex/strtod.c
```diff
@@ -68,7 +68,7 @@ double _upb_NoLocaleStrtod(const char *str, char **endptr) {
     // Update endptr to point at the right location.
     if (endptr != NULL) {
       // size_diff is non-zero if the localized radix has multiple bytes.
-      int size_diff = strlen(localized) - strlen(str);
+      size_t size_diff = (size_t)strlen(localized) - (size_t)strlen(str);
```
googleberg, line 71: I think this should be ptrdiff_t rather than size_t, no?

googleberg:
Actually... after reviewing this function: int vs size_t vs ptrdiff_t is the least of the problems with this code. We're putting this in the bin to be significantly reworked.

lyriccoder:
@googleberg Understood. Should I work on the refactor in this PR or should we create a new issue for tracking?

PR #23429, opened 2025-09-09
Cherrypick editions fixes for 32.1

mkruskal-google:


java/core/src/test/java/com/google/protobuf/ImportOptionTest.java
```diff
@@ -29,26 +26,15 @@ public void testImportOption() throws Exception {
...
+        .isEqualTo(1);
+    assertThat(messageDescriptor.getOptions().getExtension(UnittestCustomOptions.messageOpt1))
+        .isEqualTo(2);
+    assertThat(fieldDescriptor.getOptions().getExtension(UnittestCustomOptions.fieldOpt1))
+        .isEqualTo(3);
 
-    assertThat(unknownFieldsFile.asMap()).hasSize(2);
-    assertThat(unknownFieldsMessage.asMap()).hasSize(2);
-    assertThat(unknownFieldsField.asMap()).hasSize(2);
+    // TODO: Since `option_deps` are treated as `deps` in Bazel 7, the unknown
```
zhangskz, line 37: Nit: Would it make sense to check that unknown fields has size <= 1 instead for now? I think the comment as-is is a tad misleading since iirc option_deps being treated as deps means the unknown fields *are* empty when they should have 1 entry in it otherwise right?

mkruskal-google: Good point, let me add you on the actual cl since it got held up by wald anyway :P

zhangskz (approved):

src/google/protobuf/compiler/command_line_interface.cc
```diff
@@ -28,6 +28,8 @@
 #include <string>
 #include <utility>
 #include <vector>
+
+#include "google/protobuf/compiler/plugin.h"
```
jeuenfiwlf-gif, line 32: 3

src/google/protobuf/compiler/java/file.cc
```diff
@@ -139,18 +141,27 @@ void CollectExtensions(const FileDescriptor& file,
...
-  // TODO: Remove descriptor pool pollution from protoc full.
-  // Check against dependencies to handle option dependencies polluting pool
-  // from using protoc_full with built-in generators instead of plugins.
-  // Option dependencies and transitive dependencies are not allowed, except in
-  // the case of import public.
+  if (options.strip_nonfunctional_codegen) {
+    // Skip feature extensions, which are a visible (but non-functional)
+    // deviation between editions and legacy syntax.
+    absl::erase_if(*extensions, [](const FieldDescriptor* field) {
+      return field->containing_type()->full_name() == "google.protobuf.FeatureSet";
```
jeuenfiwlf-gif, line 148: https://github.com/protocolbuffers/protobuf/commit/d365736f4f6e4e496deca7051b5fa574c52cc45f

PR #23465, opened 2025-09-11
feat: update bazel central registry publish workflow

alexeagle:
Includes pre-built protoc integrity hashes in the release artifact, in a new file `bazel/private/prebuilt_tool_integrity.bzl`

tested on my fork of protobuf: https://github.com/alexeagle/protobuf/releases/tag/v0.1000.10

Next steps:
- create toolchain definitions that fetch the prebuilt protoc from github releases
- change logic in Bazel setup to use the prebuilt toolchain when desired
- create a token from @protobuf-team-bot account and add to GHA secrets so the pull request to BCR opens

### DIFF
--- a/.github/workflows/release_bazel_module.yaml
+++ b/.github/workflows/release_bazel_module.yaml
@@ -20,7 +20,7 @@ jobs:
     with:
       release_files: protobuf-*.tar.gz
       prerelease: false
-      tag_name: ${{ inputs.tag_name || github.ref_name }}
+      tag_name: ${{ inputs.tag_name }}
       # The release was already created by Google-internal mechanism,
       # so there aren't any useful assertions to make here.
       bazel_test_command: 'true'


.github/workflows/release_bazel_module.yaml
```diff
@@ -0,0 +1,30 @@
...
+  attestations: write
+  contents: write
+jobs:
+  release:
+    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.2.3
+    with:
+      release_files: protobuf-*.tar.gz
+      prerelease: false
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
+      bazel_test_command: bazel query @com_google_protobuf//:protoc
```
mkruskal-google: Should we be actually building protoc to test this a little more?

.github/workflows/release_bazel_module.yaml
```diff
@@ -0,0 +1,30 @@
...
+  publish:
+    needs: release
+    uses: ./.github/workflows/publish_to_bcr.yaml
+    with:
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
```
mkruskal-google: We probably shouldn't support releasing from anything but a tag, that seems like asking for trouble

.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+# NB: this file must be named release_prep.sh because the attestation generation doesn't trust user control.
```
mkruskal-google, line 2: Ew, this is spooky :(

.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,42 @@
...
+# see https://github.com/bazel-contrib/.github/blob/v7.2.3/.github/workflows/release_ruleset.yaml#L33-L45
+set -o errexit -o nounset -o pipefail
+
+# Argument provided by reusable workflow caller, see
+# https://github.com/bazel-contrib/.github/blob/v7.2.3/.github/workflows/release_ruleset.yaml#L104
+TAG=$1
+# HACK during debugging, to allow us to fetch older release artifacts.
```
mkruskal-google: Do you need this to be checked in or did you forget to remove it?

.github/workflows/release_bazel_module.yaml
```diff
@@ -0,0 +1,30 @@
+# Prepare a release specifically for Bazel users, including a pre-built protoc.
+name: Bazel Release
+on:
+  # Can be triggered from the GitHub Actions ui, using the "Run workflow" button on
+  # https://github.com/protocolbuffers/protobuf/actions/workflows/release_bazel_module.yaml
+  workflow_dispatch:
```
mkruskal-google, line 7: IIUC this currently can *only* be triggered manually right?

.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,42 @@
...
+# Delete the placeholder file
+tar --file $ARCHIVE_TMP --delete $INTEGRITY_FILE
+
+mkdir -p ${PREFIX}/bazel/private
+cat >${INTEGRITY_FILE} <<EOF
+"Generated during release by release_prep.sh"
+
+RELEASED_BINARY_INTEGRITY = $(
+curl -s https://api.github.com/repos/protocolbuffers/protobuf/releases/tags/${RELEASE_ARTIFACT_TAG} \
+  | jq 'reduce .assets[] as $a ({}; . + { ($a.name): ($a.digest | sub("^sha256:"; "")) })'
```
mkruskal-google: Maybe add some comments/documentation about what this is actually doing?  I assume this is duplicating some central logic, and if it ever broke I wouldn't know where to start :P

mkruskal-google (approved):

.github/workflows/release_bazel_module.yaml
alexeagle: I haven't included other triggers here yet, but we could certainly trigger after a new release is created to avoid the manual step. Is there a change you'd like to see here?

.github/workflows/release_bazel_module.yaml
alexeagle: I don't think that's useful, because protoc is already built successfully (and published as an artifact) by the time we run this release. I should have commented here - actually we don't want to do *any* testing but the re-usable workflow I'm calling here requires something as the test command.

.github/workflows/release_bazel_module.yaml
alexeagle: yup, `github.ref_name` is here because other uses of this script could be triggered in ways that didn't have the explicit `tag_name` input. Not needed in this PR.

.github/workflows/release_prep.sh
alexeagle: At least I documented it?

.github/workflows/release_prep.sh
alexeagle: can remove

.github/workflows/release_bazel_module.yaml
alexeagle: Changed to just `true`

.github/workflows/release_prep.sh
alexeagle: extracted, commented, and line-breaks

.github/workflows/release_bazel_module.yaml
mkruskal-google: No this can be a follow up, I was just looking to clarify

alexeagle:
Note @mkruskal-google I just did a fresh test from HEAD on my fork: https://github.com/alexeagle/protobuf/releases/tag/v0.1000.11

good to merge, then I'll start on the next bit

zhangskz:
IIUC, we should also be able to remove the `Publish to BCR` github apps and its configured repo access once we switch to using this workflow for publication to BCR right?

alexeagle:
@zhangskz yes - now that this is merged, let's do those release prep steps? Remove the Publish to BCR app, instead we need to create a token from @protobuf-team-bot account and add to GHA secrets to open the pull request to BCR

PR #23519, opened 2025-09-16
I believed myself strongly.I'm always Safe.I'm always Rights.

pg9835624-cyber:
I can only controls my own datas all.i don't alliw whoever global data teams controls and laws i dont want.So my owns i doings One framework .All my owns internet and my owns products I can only make Laws and Rules.Payments also Cash my bank Acc


pg9835624-cyber:
I change to take back my own API or my displays techanology .I don't like Google frame work and that google global teams using laws and Rules i don't need for me that laws because this techanology is my onw Internet .My own display API All products is my owns somany Techanology So if google and universe this world all country and all presidents want to use my own Program my own products my own data my own Websites my own Created Artificial Intelligents AI .This world International All presidents and clients and Global Data Privacy Teams stay in use Owner Founder Me or PyaePhyoAung or Common Laws Rules ground uses that is fair.I believed and i love my techanology so start today i changing all before they systems.Who am' I ?
I'm the best or I'm only can do i created for me .I'm always Rights .I'm always Safe.A better me is Comings for me.

pg9835624-cyber (commented):
My Techanology is My own Laws My Rules under in use All International or Global.

PR #23547, opened 2025-09-17
Add gencode smoke tests

esrauchg:
This creates a new //compatability/smoke directory which has a simple unit test and checked-in gencode which was produced by a given protoc version.

Included is a script that you can run with a protoc release number, which will fetch the released protoc linux-x64 binary, run it on the test proto, add the code to a subdirectory and hook up the unit test to that generated code.

If this setup looks good we can quickly add a number of additional versions; I have already validated it works with a wider range but this PR only includes one version to show the setup and focus the initial review on the core things that aren't about the (naming, directory structure, etc)


compatibility/smoke/test.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
zhangskz, line 1: proto3_gencode_test.proto? (and similar for proto2, editions to test these in only in branches that supported the syntax)

compatibility/smoke/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+load("@rules_java//java:defs.bzl", "java_test")
+
+[java_test(
```
zhangskz: cool, i didn't know you could do this in build files! i might still prefer pulling this into a blaze rule in bzl but could be convinced: 

```
gencode_compatibility_smoke_test("3.19.0")
...
```

```
def gencode_compatibility_smoke_test(gencode_version):
  java_test(...)


```

compatibility/smoke/SmokeTest.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.protobuf.util.JsonFormat;
+import java.util.Map;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+
+/** Unit test that touches a few behaviors on old versions of generated code. */
+@RunWith(JUnit4.class)
+public class SmokeTest {
```
zhangskz: GencodeSmokeTest?

compatibility/smoke/SmokeTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @Test
+  public void builderWorks() {
```
zhangskz: testBuilder, testReflection, etc?

compatibility/smoke/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+This directory contains a smoke unit tests that runs the core functionality of
+generated code against the latest runtime.
+
+Note that running protoc as a build step rather than checking in gencode is the
+strongly recommended pattern, and any Bazel users should use `proto_library()`
+rules to achieve that rather than checking in generated code.
+
+This directory is only using checked in generated code as the mechanism to test
+tho compatibility with stale gencode created by an older protoc, which is
```
zhangskz: sp: the

compatibility/smoke/SmokeTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.google.protobuf.Descriptors.FieldDescriptor;
+import com.google.protobuf.util.JsonFormat;
+import java.util.Map;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+
+/** Unit test that touches a few behaviors on old versions of generated code. */
```
zhangskz, line 22: IDT this has changed much in practice and these API's in particular should be stable, but we could remove a generated method in a future breaking change s.t. these tests do need to be versioned (theoretically, prob not in practice anytime soon) .

compatibility/smoke/add_gencode.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e -x
```
zhangskz, line 2: Could we have protobuf-team-bot handle this? I imagine we could could configure branches in yml and a GHA that calls this script to check in jars + BUILD file for those branches (similar to staleness_refresh.yml) if test.proto / SmokeTest.java change?

compatibility/smoke/SmokeTest.java
esrauchg: Going with "StaleGencodeSmokeTest"

compatibility/smoke/SmokeTest.java
esrauchg: Yeah, I think its pragmatically this will only cover the core functionality and so shouldn't be sensitive to that.

But I think its also not a problem besides that because:

- I think in Editions strategy we intend to keep the gencode API surface stable per input and so we wouldn't want that (instead at some point we'd maybe add an Edition 2024 test file and then the Edition 2024 smoke unit test against it, and that test only runs above whatever version accepts Edition 2024).

- The runtime methods used (like the JSON, reflection) could change, but its WAI that this test should also change (including those new methods as applied to all prior supported gencode). If we did something stupid like changed the name of `JsonFormat`, what this test wants to check is if the new API works successfully with old gencode, nothing about the old runtime methods.

The only way I can see that not holding up is if we end up doing something like adding a new runtime method _and_ document that it simply throws on old gencode (= non-breaking change that doesn't actually work with ancient gencode). But I think we can handle that if we ever do that.

compatibility/smoke/add_gencode.sh
esrauchg: So I actually like the property of "test the gencode that is produced by literally the prior protoc released .exe", no futzing with building protoc again separately that just 'should' match whatever we have released.

But I think its also an open question if we want to add the gencode here for every point release that we do, or if we just want to leave it as a smaller set of them manually chosen based on knowing the ranges where gencode had any meaningful change to them (which looking backwards appears to have only happened at the 3.7/3.8 boundary, and the 3.21/3.22 boundary, and across the 3.25/4.26 boundary).

If we _do_ want to check in for every single minor release, it makes sense that either:
- The bot handle it and have it poll our releases page and add this after we've done every release.
- Just add a final number to the releaser script which just runs this script with the corresponding number (which is epsilon more overhead for the releaser to hit that number, and one more PR that has to be stamped by someone on every release).

compatibility/smoke/BUILD.bazel
esrauchg: Done

compatibility/smoke/proto3_gencode_test.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package legacy_gencode_test.proto3;
+
```
zhangskz, line 4: nit: java_outer_classname = Proto3GencodeTestProto 

(i mistook the gencode for actual test on my first pass, and this would match the default edition 2024 name too)

compatibility/smoke/add_gencode.sh
zhangskz: (oops drafted this ~last week but forgot to send) 

yeah i think checking in the gencode is sane given this is meant to replicate a non-bazel checked in gencode case anyways. bazel is just helpful where it can help with maintenance burden.

if we wanted to be comprehensive about our coverage of minor release, perhaps we could have the script check if the outputted gencode is actually unique and skip minor versions where gencode hasn't actually changed.

> - The bot handle it and have it poll our releases page and add this after we've done every release.

This potentially seems the easiest to maintain if we were to add fields or make modifications to the test proto / smoke tests and need to update *everything* which may be rare anyways. We could also explicitly set a minimum / maximum gencode version in yml and update just that as part of release.

> - Just add a final number to the releaser script which just runs this script with the corresponding number (which is epsilon more overhead for the releaser to hit that number, and one more PR that has to be stamped by someone on every release).

If we went this route, we could collapse this into the single release PR (which is already effectively 2 PR's collapsed into 1 and merged (without squashing) to minimize the number of PRs that need to be stamped on release). But would need rearranging since we'd need to create the github release prior to creating / merging the PR for that release to exist already for the new smoke test which is technically fine but odd.

mkruskal-google: Do we need to check in the gencode at all?  We could have GHA invoke this script on the fly

esrauchg: We could, but I think I like the characteristics of checking in the gencode for this specific test: mainly that it you can have a local checkout and make a proposed edit and run `bazel test ...` and run your change with all of the old gencode, with it being hooked up as a download in GHA you would have to do something specially more manual or iterate against slow GHA runs, right?

More minor it also seems more easily inspectable, and maybe avoids some amount of flakes from the CI re-fetching all of the corresponding released protoc to regen the code constantly.

zhangskz: I think checked in gencode does mimic the user case we're trying to test anyways, so I think either is fine dep on what's better for us maintenance wise. Esp if we invoke on-the-fly we should probably include some sort of sha check (and same for python if we don't already).

compatibility/smoke/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
+load("@rules_java//java:defs.bzl", "java_test")
+
+load("//compatibility/smoke:stale_gencode_smoke_test.bzl", "stale_gencode_smoke_test");
+
+stale_gencode_smoke_test("3.19.0")
```
mkruskal-google, line 5: Did you intentionally have 3 of these?

compatibility/smoke/add_gencode.sh
```diff
@@ -0,0 +1,27 @@
...
+
+PROTOC="/tmp/protoc-$VER/bin/protoc"
+
+$PROTOC --version
+
+mkdir -p v$VER
+$PROTOC proto3_gencode_test.proto --java_out=v$VER
+
+cp CHECKED_IN_GENCODE_BUILD.bazel.template v$VER/BUILD.bazel
+
```
mkruskal-google, line 28: nit: it would be good to cleanup the temporary files at the end

mkruskal-google (approved):

compatibility/smoke/BUILD.bazel
esrauchg: No, fixed. Accidental side effect of testing changes to the script on the same version and only cleaning up some of it.

PR #23584, opened 2025-09-20
bazel: add missing rules_cc loads

keith:
This is required for use with bazel after this commit https://github.com/bazelbuild/bazel/commit/71ca0ed111ff3d842a0d23bc3a46bd2e6745491d

Many files have these already

### DIFF
--- a/lua/BUILD.bazel
+++ b/lua/BUILD.bazel
@@ -5,6 +5,9 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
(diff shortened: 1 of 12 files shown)


fmeum:
Could this be included in the next minor release?

zhangskz (approved):

zhangskz:
Looks like this will be needed for Bazel 9 support which is released in November so likely won't officially support until at least v34.0. We just cut v33.0-RC1, but I think this seems reasonable to include in v33.0 anyways.

Issue #6641, opened 2019-09-12
Java util for conversions between Timestamp/Duration and java.time.Instant/Duration


sulume:
It's not a big deal but tedious to every time exchange between the types while remembering its internal representations (seconds + nanos). It's nice if the utility methods for the official utility messages are given with the official package to avoid other users have their own versions of the same logic.

elharo:
This can't really be addressed until we drop support for Java 7. That might happen in 2022 but no promises.

elharo:
This amounts to a request to open source google3/java/com/google/protobuf/util/JavaTimeConversions.java, which is completely reasonable. However it does require us to drop Java 7.

googleberg:
Requires Java 8 support.
Can't do at the moment due to internal limitations.

desh-woes:
Any updates as to when this is planned to happen?

googleberg:
As @elharo noted, we have an internal utility that we'd like to open source. However, we are currently constrained by needing to support Android API level 21 as a minimum. `java.time.Instant` was added to Android in API level 26.

PR #23687, opened 2025-09-29
Timestamp <-> Instant

michaldo:
When protobuf use Java 8, nothing can stop complete 6 years old request https://github.com/protocolbuffers/protobuf/issues/6641


googleberg:
We actually already have an internal utility to do this that we would love to open source. However, we are currently constrained by needing to support Android API level 21 as a minimum. `java.time.Instant` was added to Android in API level 26.

